Issue #17433, opened 2020-10-23
PathExists in path unit when paired with ConditionPathExists in triggered service makes pid1 unresponsive


joshuamiller01:
**systemd version the issue has been seen with**
246

**Used distribution**
CentOS7

**Linux kernel version used** (`uname -a`)
5.2.9

**CPU architecture issue was seen on**
x86_64

**Expected behaviour you didn't see**
path unit fires upon the existence of `PathExists`, corresponding service unit runs once, systemd remains responsive

**Unexpected behaviour you saw**
path unit fired after file was touched, service unit also fired, but pid1 CPU goes to 100%, and calls via `systemctl` (both via sudo and as root) hang.

**Steps to reproduce the problem**
[root@box ~]# cat /etc/systemd/system/test123.path
[Path]
PathExists=/tmp/test123

[Install]
WantedBy=multi-user.target

[root@box ~]# cat /etc/systemd/system/test123.service
[Unit]
Description=Test 123
ConditionPathExists=!/tmp/nopenopenope

[Service]
Type=oneshot
ExecStart=/bin/touch /tmp/nopenopenope

[Install]
WantedBy=multi-user.target

[root@box ~]# systemctl daemon-reload
[root@box ~]# systemctl start test123.path
[root@box ~]# touch /tmp/test123

[root@box ~]# systemctl status test123.service
   ... hangs  

/var/log/messages has many thousands of messages about:
"systemd[1]: Condition check resulted in Test 123 being skipped."

Removing the original trigger file (`rm /tmp/test123`) restores systemd responsiveness

mbiebl:
Sounds like a duplicate of #16669

joshuamiller01:
Removing the `ConditionPathExists` property changes the behavior - pid1 never goes unresponsive, and the service and path units are marked as failed with "failed (Result: start-limit-hit)" and " failed (Result: unit-start-limit-hit)" respectively.

anitazha:
Sounds like the rate limiting was fixed properly for unit start as part of #16669, but not condition checking?

keszybz:
https://bugzilla.redhat.com/show_bug.cgi?id=1919538

DaanDeMeyer:
So it turns out we check unit conditions before we check start rate limiting so if conditions fail no rate limiting gets applied and we can enter a busy loop. I'm not too familiar with the code, but intuitively it seems like this could be fixed if we just move the start rate limiting checks before the condition checks. 

[...]

Issue #21603, opened 2021-12-02
TEST-17-UDEV randomly fails on Ubuntu ppc64el


yuwata:
E.g. https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/ppc64el/s/systemd-upstream/20211202_024112_d0130@/log.gz

I am not sure the failure is a simple timeout or not.

mrc0mmand:
```
[  173.340103] testsuite-17.sh[287]: + for t in "${0%.sh}".*.sh
[  173.364285] testsuite-17.sh[287]: + echo 'Running /usr/lib/systemd/tests/testdata/units/testsuite-17.04.sh'
[  173.373284] testsuite-17.sh[287]: Running /usr/lib/systemd/tests/testdata/units/testsuite-17.04.sh
[  173.391867] testsuite-17.sh[287]: + .//usr/lib/systemd/tests/testdata/units/testsuite-17.04.sh
[  173.839042] testsuite-17.sh[486]: + set -o pipefail
[  173.878203] testsuite-17.sh[486]: + mkdir -p /run/udev/rules.d/
[  174.159781] testsuite-17.sh[486]: + test '!' -f /run/udev/tags/added/c1:3
[  174.172940] testsuite-17.sh[486]: + test '!' -f /run/udev/tags/changed/c1:3
[  174.206573] testsuite-17.sh[491]: + grep -E 'E: (TAGS|CURRENT_TAGS)=.*:(added|changed):'
[  174.270381] testsuite-17.sh[490]: + udevadm info /dev/null
[  174.887732] testsuite-17.sh[486]: + cat
[  175.035023] testsuite-17.sh[486]: + udevadm control --reload
[  175.346805] testsuite-17.sh[486]: + SYSTEMD_LOG_LEVEL=debug
[  175.368767] testsuite-17.sh[486]: + udevadm trigger --verbose --settle --action add /dev/null
[  175.573193] testsuite-17.sh[495]: SELinux enabled state cached to: disabled
[  176.333073] testsuite-17.sh[495]: /sys/devices/virtual/mem/null
[  176.333073] testsuite-17.sh[495]: settle /sys/devices/virtual/mem/null
[  176.520292] testsuite-17.sh[486]: + test -f /run/udev/tags/added/c1:3
[  176.782810] systemd[1]: testsuite-17.service: Main process exited, code=exited, status=1/FAILURE
[...]

PR #21808, opened 2021-12-17
core: Add trigger limit for path units

### LINKED ISSUE #21802: v250-rc2: Regression in path unit handling regarding not started service unit
 > v250-rc2

**Used distribution**
 > Own build

 > 5.10.85

 > amd64

 > waiting path unit is waiting until the waited-for file is actually created

 > path unit fails because associated service unit is not started because ConditionFileExists is not fulfilled

2 units:

```
# test.path
[Unit]
Description = path check

[Path]
PathExists = /tmp/foo
``` 
```
# test.service
[Unit]
Description = test
ConditionPathExists=/tmp/foo

[Service]
Type = oneshot
ExecStart = /bin/true
RemainAfterExit = true

```
The following actions are performed
- `systemctl daemon-reload`
- `systemctl start test.path`
- Now test.path is active and waiting
- `systemctl start test.service`
- This fails because the ConditionPathExists is not fulfilled
- This also causes failure in test.path, which is the regression 

When performing the same test with v249 the unit test.path remains active and waiting

```
# Logs with v250-rc2
2021-12-17 11:53:32.888 init.scope: test.path: Trying to enqueue job test.path/start/replace
2021-12-17 11:53:32.889 init.scope: test.path: Installed new job test.path/start as 380
2021-12-17 11:53:32.889 init.scope: test.path: Enqueued job test.path/start as 380
2021-12-17 11:53:32.890 init.scope: test.path: Changed dead -> waiting
2021-12-17 11:53:32.890 init.scope: test.path: Job 380 test.path/start finished, result=done
2021-12-17 11:53:32.890 init.scope: Started path check.
2021-12-17 11:53:57.941 init.scope: test.service: Trying to enqueue job test.service/reload/replace
2021-12-17 11:54:06.631 init.scope: test.service: Trying to enqueue job test.service/restart/replace
2021-12-17 11:54:06.631 init.scope: test.service: Installed new job test.service/restart as 413
2021-12-17 11:54:06.632 init.scope: test.service: Enqueued job test.service/restart as 413
2021-12-17 11:54:06.632 init.scope: test.service: Job 413 test.service/restart finished, result=done
2021-12-17 11:54:06.632 init.scope: test.service: Converting job test.service/restart -> test.service/start
2021-12-17 11:54:06.632 init.scope: test.service: ConditionPathExists=/tmp/foo failed.
2021-12-17 11:54:06.632 init.scope: test.service: Starting requested but condition failed. Not starting unit.
2021-12-17 11:54:06.632 init.scope: test.service: Job 413 test.service/start finished, result=done
2021-12-17 11:54:06.633 init.scope: test was skipped because of a failed condition check (ConditionPathExists=/tmp/foo).
2021-12-17 11:54:06.633 init.scope: test.path: Failed with result 'unit-condition-failed'.
2021-12-17 11:54:06.636 init.scope: test.path: Changed waiting -> failed
2021-12-17 11:54:06.636 init.scope: test.path: Unit entered failed state.
2021-12-17 11:55:13.999 init.scope: test.path: Changed dead -> failed
2021-12-17 11:55:18.731 init.scope: test.path: Changed dead -> failed

```

From the output of `systemctl list-units` in v250-rc2 after the test:
```
● test.path      loaded    failed   failed    path check
  test.service   loaded    inactive dead      test
```

```
# Logs with v249
2021-12-17 12:15:58.362 init.scope: test.path: Trying to enqueue job test.path/start/replace
2021-12-17 12:15:58.362 init.scope: test.path: Installed new job test.path/start as 1098
2021-12-17 12:15:58.362 init.scope: test.path: Enqueued job test.path/start as 1098
2021-12-17 12:15:58.363 init.scope: test.path: Changed dead -> waiting
2021-12-17 12:15:58.363 init.scope: test.path: Job 1098 test.path/start finished, result=done
2021-12-17 12:15:58.363 init.scope: Started path check.
2021-12-17 12:16:06.483 init.scope: test.service: Trying to enqueue job test.service/start/replace
2021-12-17 12:16:06.483 init.scope: test.service: Installed new job test.service/start as 1131
2021-12-17 12:16:06.483 init.scope: test.service: Enqueued job test.service/start as 1131
2021-12-17 12:16:06.484 init.scope: test.service: ConditionPathExists=/tmp/foo failed.
2021-12-17 12:16:06.484 init.scope: test.service: Starting requested but condition failed. Not starting unit.
2021-12-17 12:16:06.484 init.scope: test.service: Job 1131 test.service/start finished, result=done
2021-12-17 12:16:06.484 init.scope: Condition check resulted in test being skipped.
```

From the output of `systemctl list-units` in v249 after the test:
```
  test.path        loaded    active   waiting   path check
  test.service     loaded    inactive dead      test
```

daandemeyer:
When conditions fail on a service unit, a path unit can cause
PID 1 to busy loop as it keeps trying to activate the service unit.
To avoid this from happening, add a trigger limit to the path unit,
identical to the trigger limit we have for socket units.

Initially, let's start with a high limit and not make it configurable.
If needed, we can add properties to configure the rate limit similar
to the ones we have for socket units.

This is an alternative fix for #17433. The PR also reverts the previous fix which fixes #21802.

### DIFF
--- a/src/core/path.c
+++ b/src/core/path.c
@@ -265,6 +265,9 @@ static void path_init(Unit *u) {
         assert(u->load_state == UNIT_STUB);
 
         p->directory_mode = 0755;
+
+        p->trigger_limit.interval = 2 * USEC_PER_SEC;
+        p->trigger_limit.burst = 200;
 }
 
 void path_free_specs(Path *p) {
@@ -493,6 +496,12 @@ static void path_enter_running(Path *p) {
         /* Don't start job if we are supposed to go down */
         if (unit_stop_pending(UNIT(p)))
                 return;
+
+        if (!ratelimit_below(&p->trigger_limit)) {
+                log_unit_warning(UNIT(p), "Trigger limit hit, refusing further activation.");
+                path_enter_dead(p, PATH_FAILURE_TRIGGER_LIMIT_HIT);
+                return;
+        }
 
         trigger = UNIT_TRIGGER(UNIT(p));
         if (!trigger) {
@@ -836,6 +845,7 @@ static const char* const path_result_table[_PATH_RESULT_MAX] = {
         [PATH_FAILURE_RESOURCES]            = "resources",
         [PATH_FAILURE_START_LIMIT_HIT]      = "start-limit-hit",
         [PATH_FAILURE_UNIT_START_LIMIT_HIT] = "unit-start-limit-hit",
+        [PATH_FAILURE_TRIGGER_LIMIT_HIT]    = "trigger-limit-hit",
 };
 
 DEFINE_STRING_TABLE_LOOKUP(path_result, PathResult);
--- a/src/core/path.h
+++ b/src/core/path.h
@@ -46,6 +46,7 @@ typedef enum PathResult {
         PATH_FAILURE_RESOURCES,
         PATH_FAILURE_START_LIMIT_HIT,
         PATH_FAILURE_UNIT_START_LIMIT_HIT,
+        PATH_FAILURE_TRIGGER_LIMIT_HIT,
         _PATH_RESULT_MAX,
         _PATH_RESULT_INVALID = -EINVAL,
 } PathResult;
@@ -61,6 +62,8 @@ struct Path {
         mode_t directory_mode;
 
         PathResult result;
+
+        RateLimit trigger_limit;
 };
 
 void path_free_specs(Path *p);
--- a/test/testsuite-63.units/test63.service
+++ b/test/testsuite-63.units/test63.service
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 [Unit]
-ConditionPathExists=!/tmp/nonexistent
+ConditionPathExists=/tmp/nonexistent
 
 [Service]
 ExecStart=true
--- a/test/units/testsuite-63.service
+++ b/test/units/testsuite-63.service
@@ -5,13 +5,26 @@ Description=TEST-63-ISSUE-17433
 [Service]
 ExecStartPre=rm -f /failed /testok
 Type=oneshot
+
+# Test that a path unit continuously triggering a service that fails condition checks eventually fails with
+# the trigger-limit-hit error.
 ExecStart=rm -f /tmp/nonexistent
 ExecStart=systemctl start test63.path
 ExecStart=touch /tmp/test63
-# Make sure systemd has sufficient time to hit the start limit for test63.service.
+# Make sure systemd has sufficient time to hit the trigger limit for test63.path.
 ExecStart=sleep 2
-ExecStart=sh -x -c 'test "$(systemctl show test63.service -P ActiveState)" = failed'
-ExecStart=sh -x -c 'test "$(systemctl show test63.service -P Result)" = start-limit-hit'
+ExecStart=sh -x -c 'test "$(systemctl show test63.service -P ActiveState)" = inactive'
+ExecStart=sh -x -c 'test "$(systemctl show test63.service -P Result)" = success'
 ExecStart=sh -x -c 'test "$(systemctl show test63.path -P ActiveState)" = failed'
-ExecStart=sh -x -c 'test "$(systemctl show test63.path -P Result)" = unit-start-limit-hit'
+ExecStart=sh -x -c 'test "$(systemctl show test63.path -P Result)" = trigger-limit-hit'
+
+# Test that starting the service manually doesn't affect the path unit.
+ExecStart=rm -f /tmp/test63
+ExecStart=systemctl reset-failed
+ExecStart=systemctl start test63.path
+ExecStart=systemctl start test63.service
+ExecStart=sh -x -c 'test "$(systemctl show test63.service -P ActiveState)" = inactive'
+ExecStart=sh -x -c 'test "$(systemctl show test63.service -P Result)" = success'
+ExecStart=sh -x -c 'test "$(systemctl show test63.path -P ActiveState)" = active'
+ExecStart=sh -x -c 'test "$(systemctl show test63.path -P Result)" = success'
 ExecStart=sh -x -c 'echo OK >/testok'


src/core/path.c
```diff
@@ -265,6 +265,9 @@ static void path_init(Unit *u) {
         assert(u->load_state == UNIT_STUB);
 
         p->directory_mode = 0755;
+
+        p->trigger_limit.interval = 2 * USEC_PER_SEC;
+        p->trigger_limit.burst = 200;
```
daandemeyer, line 270: I went for the same limits as accept socket units. Not sure if we should be even more lenient for path units.

DaanDeMeyer:
Let me know if the rate limit for path units should be configurable as well.

src/core/path.c
```diff
@@ -494,6 +497,12 @@ static void path_enter_running(Path *p) {
         if (unit_stop_pending(UNIT(p)))
                 return;
 
+        if (!ratelimit_below(&p->trigger_limit)) {
+                log_unit_warning(UNIT(p), "Trigger limit hit, refusing further activation.");
+                path_enter_dead(p, PATH_FAILURE_TRIGGER_LIMIT_HIT);
```
bluca, line 502: will it become runnable again automatically later with this?

bluca:
would it be possible to add a unit test for the regression case? it should be simple enough

bluca (approved):

src/core/path.c
daandemeyer: It won't unless the path unit is started again, but if we hit this at all, it very likely means that the path unit is causing pid1 to busy loop so we don't actually want it to trigger again until that's fixed.

bluca: This is an issue though - for example think about a service using the new psi based conditions. Things shouldn't fail forever if a condition is missed. Can't we add a timer event to automatically clean the trigger limit after a while?

daandemeyer: Is it an issue though if you consider that currently, this situation would cause PID1 to busy loop when a condition is missed? The trigger limits for socket units don't have such remediation either unless I missed something.

I figure at the moment, using conditions with path units containing PathExists is completely broken since as soon as conditions fail, you end up busy looping PID 1. There's two problems here, one is that it's possible to cause a busy loop in PID 1 by combined usage of conditions and PathExists. The second problem is figuring out how we should define the interaction between PathExists and failed conditions.

This PR solves the first problem but not the second one. I still think it's an improvement over what we have since we solve the first problem. The second problem is IMO best solved in a second PR and can be delayed to v251.

DaanDeMeyer:
Seems there some issues with UDEV tests. cc @mrc0mmand

mrc0mmand:
> Seems there's some issues with the UDEV tests. cc @mrc0mmand

Both are unrelated - focal-ppc64el is #21603 and TEST-64-UDEV-STORAGE is tracked in https://github.com/systemd/systemd-centos-ci/issues/424.

src/core/path.c
bluca: I agree that this is better than the current situation, however we really need to fix the side effect too, as it makes the combination of event triggered services and conditions way less powerful than it could be

bluca:
> Let me know if the rate limit for path units should be configurable as well.

I think it should yeah - added via: https://github.com/systemd/systemd/pull/21818

mrc0mmand:
> @DaanDeMeyer I've seen TEST-63-ISSUE-17433 fail a couple of times in the past few days, eg:
> 
> https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/arm64/s/systemd-upstream/20211221_161634_86521@/log.gz (logs at https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/arm64/s/systemd-upstream/20211221_161634_86521@/artifacts.tar.gz )

I think this should be "fixed" by #21840 as well, since it happens when the machine is under a heavy load and it misses the tight trigger limit interval. But it's weird that the run is apparently from today, so it still might need some tweaking? Or maybe Ubuntu CI doesn't rebase the PR branch on the latest main? (I can't remember).

```
-- Journal begins at Tue 2021-12-21 17:00:05 CET, ends at Tue 2021-12-21 17:02:03 CET. --
[ 3030.177187] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.185919] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.197937] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.209818] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.221758] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.233733] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.245915] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.257388] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.269134] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.278568] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.291351] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.303220] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.317225] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.328996] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
[ 3030.335592] systemd[1]: test63.service was skipped because of a failed condition check (ConditionPathExists=/tmp/nonexistent).
```

bluca:
the ubuntu builds were not with coverage flags though?

mrc0mmand:
> the ubuntu builds were not with coverage flags though?

Ugh, of course :facepalm: 

So I guess if the `TriggerLimitIntervalSec=` tweak would be done uncoditionally?

bluca:
Yeah, worth a try I think

DaanDeMeyer:
If this is a problem for our tests, won't it turn out to be a problem elsewhere as well? Shouldn't we just relax the default values then?

bluca:
> If this is a problem for our tests, won't it turn out to be a problem elsewhere as well? Shouldn't we just relax the default values then?

The test CI is _really_ slow - but I guess it wouldn't hurt to increase the default timer?

DaanDeMeyer:
I'm not entirely sure. I'm kinda scared about interfering with use cases that trigger path units very often.

mrc0mmand:
> I'm not entirely sure. I'm kinda scared about interfering with use cases that trigger path units very often.

Well, I guess we could tweak the trigger limit only if we're running with coverage or without KVM. With that we'd still have coverage for the "unaltered" test case.

PR #21809, opened 2021-12-17
network: a couple of tweaks suggested by Coccinelle

mrc0mmand:


src/network/networkd-address-generation.c
```diff
@@ -359,8 +359,7 @@ int config_parse_address_generation_type(
                                 addr = addr_alloc;
                         }
                 } else if (*addr == ',') {
```
bluca: remove brackets?

mrc0mmand: Done. I wasn't sure if it wouldn't look too weird, but I guess it could be worse.

bluca (approved):

yuwata (approved):

PR #21814, opened 2021-12-18
network: several fixes for DHCP prefix delegation

yuwata:


src/libsystemd-network/sd-radv.c
```diff
@@ -631,19 +632,21 @@ int sd_radv_add_prefix(sd_radv *ra, sd_radv_prefix *p) {
...
         ra->n_prefixes++;
 
-        if (ra->state == RADV_STATE_IDLE) {
-                log_radv(ra, "Added prefix %s", strna(addr_p));
+        log_radv(ra, "Added prefix %s", strna(addr_p));
+
+        if (ra->state == RADV_STATE_IDLE)
+                return 0;
+
+announce:
```
bluca, line 640: Maybe a small static helper instead of these gotos would look better?

src/libsystemd-network/sd-dhcp-client.c
```diff
@@ -1994,6 +1994,7 @@ static int client_receive_message_udp(
                 return 0;
         }
 
+        log_dhcp_client(client, "Received message from UDP socket, processing.");
```
bluca, line 1997: is this at debug level?

bluca (approved):

src/libsystemd-network/sd-radv.c
yuwata: OK. I will do that in another PR.

src/libsystemd-network/sd-dhcp-client.c
yuwata: Yes.

src/libsystemd-network/sd-radv.c
yuwata: #21820

PR #21818, opened 2021-12-18
path unit: add TriggerLimitBurst= and TriggerLimitIntervalSec= 

bluca:
Given there's now a default for these settings, also allow users to configure
them, matching socket units

Follow-up for https://github.com/systemd/systemd/pull/21808


DaanDeMeyer:
LGTM, thanks for doing this

PR #21823, opened 2021-12-19
repart: avoid hardcoded sector size

tomty89:


src/partition/repart.c
```diff
@@ -1811,40 +1813,42 @@ static int context_load_partition_table(
...
         } else {
                 /* Add space left of first partition */
-                assert(left_boundary >= first_lba);
+                if (sector_size == 512)
+                        left_boundary = round_down_size(left_boundary, 4096);
 
-                first_lba = round_up_size(first_lba, 4096);
-                left_boundary = round_down_size(left_boundary, 4096);
-                last_lba = round_down_size(last_lba, 4096);
+                assert(left_boundary >= first_lba);
```
tomty89, line 1851: It seems right to also perform this assertion check after aligning, but I'm not really sure if it is.

medhefgo:
Imho, this should add a testcase to "./src/partition/test-repart.sh"

src/partition/repart.c
```diff
@@ -215,10 +216,12 @@ DEFINE_PRIVATE_STRING_TABLE_LOOKUP_FROM_STRING_WITH_BOOLEAN(encrypt_mode, Encryp
 
 
 static uint64_t round_down_size(uint64_t v, uint64_t p) {
-        return (v / p) * p;
+        return p == 1 ? v : (v / p) * p;
```
keszybz: Isn't this actually making the code slower? A conditional might be more costly than an unnecessary operation.

I think this whole commit can be dropped. It's doing a microoptimization that either has no effect or might maybe even be a pessimization. The most important part is that it makes the code more complicated, and we don't want that.

keszybz (changes requested):

poettering:
What's the point here? It was a contious decision to align everything to 4K, since we want images to make sense on any backing store, and wasting 7K per partition in the (unlikely) worst case isn't really something we should be concerned about. 

Hence, why bother? Things become a lot simpler and more universal if we always align to multiple of 4K. Moreover we become independent of sector size reporting of backing store in case complex storage is used.

Hence, I am not convinced this should be merged. This removes universiality at the benefit of saving 7K per partition...

src/partition/repart.c
tomty89: Yeah I've been thinking whether da83ab0bc8c4a675b44380283ed2e1a4b77df8a8 makes enough of a point, since devices with logical block size other than 512 and 4096 are probably scarce.

tomty89:
> Moreover we become independent of sector size reporting of backing store in case complex storage is used.

@poettering Unless I missed something, 4K-alignment does not make it work with "AF 4Kn" drive (logical block size = 4096) automatically. (I'm not talking about images on one though, but an actual block device to work on.)

src/partition/repart.c
```diff
@@ -4777,6 +4778,24 @@ static int determine_auto_size(Context *c) {
         return 0;
 }
 
+static void determine_sector_size(const char *node, uint64_t *lbs) {
+        _cleanup_close_ int fd = -1;
+        struct stat st;
+        int r;
+
+        fd = open(node, O_RDONLY|O_CLOEXEC);
+
```
bluca: we do the checks immediately after, ie, remove newline here and after fstat too

src/partition/repart.c
```diff
@@ -4777,6 +4778,24 @@ static int determine_auto_size(Context *c) {
         return 0;
 }
 
+static void determine_sector_size(const char *node, uint64_t *lbs) {
+        _cleanup_close_ int fd = -1;
+        struct stat st;
+        int r;
+
```
bluca: assert on node and lbs

src/partition/repart.c
```diff
@@ -4777,6 +4778,24 @@ static int determine_auto_size(Context *c) {
...
+        fd = open(node, O_RDONLY|O_CLOEXEC);
+
+        if (fd < 0)
+                return;
```
bluca: the errnos should be returned, so that the caller can log

src/partition/repart.c
```diff
@@ -4777,6 +4778,28 @@ static int determine_auto_size(Context *c) {
...
+                return -errno;
+
+        r = fstat(fd, &st);
+        if (r < 0)
+                return -errno;
+
+        if (!S_ISBLK(st.st_mode))
+                return 0;
+
+        ioctl(fd, BLKSSZGET, lbs);
```
bluca: error checking is missing

bluca: also use a local variable, and assign to the caller's ref only if the ioctl succeeds, otherwise leave lbs unoutched

tomty89: > also use a local variable, and assign to the caller's ref only if the ioctl succeeds, otherwise leave lbs unoutched

Hmm, is that really necessary though? I don't think (the) ioctl will touch lbs when it fails.

bluca: it's part of the coding style

tomty89:
I figure it makes slightly more sense to `stat` `node` first instead of `open`ing it unconditionally and then `fstat` (even when we need an fd for the ioctl anyway). Let me know if I'm wrong.

bluca:
> I figure it makes slightly more sense to `stat` `node` first instead of `open`ing it unconditionally and then `fstat` (even when we need an fd for the ioctl anyway). Let me know if I'm wrong.

It was better the other way around, as it's a slight TOCTOU - even better, you should be able to just pass down `backing_fd` instead of the path? That way there's no need to open it

tomty89:
> as it's a slight TOCTOU

Ah right, I forgot about that. I was thinking about the open flags (the fact that which to use more or less depends on what is being opened, though the ones being used probably suits anyway).

> you should be able to just pass down backing_fd instead of the path

To be frank, I have no idea. I don't actually grok all the heuristics / possible scenarios here, e.g. in what case this is expected to happen:
https://github.com/systemd/systemd/blob/af706d5254352b2004afeb037bada451803c023c/src/partition/repart.c#L4637-L4639
Maybe I should do some more tests.

I also wondered if I could just insert the ioctl here:
https://github.com/systemd/systemd/blob/af706d5254352b2004afeb037bada451803c023c/src/partition/repart.c#L4472-L4478

bluca:
> I also wondered if I could just insert the ioctl here:

Yeah, that would probably be fine

src/partition/repart.c
```diff
@@ -4450,6 +4451,7 @@ static int acquire_root_devno(
         _cleanup_close_ int fd = -1;
         struct stat st;
         int r;
+        unsigned ssz;
```
bluca: This declaration can be moved down inside the if block

src/partition/repart.c
```diff
@@ -1591,6 +1592,13 @@ static int context_load_partition_table(
         if (r < 0)
                 return log_error_errno(r, "Failed to open device '%s': %m", node);
 
+        /* FIXME: it's probably better to take this out of its current caller
```
medhefgo: Don't add a FIXME comment when you're gonna remove it some commit later.

tomty89: Probably I should just squash commit 1 and 3.

medhefgo (approved):

poettering:
I don't follow. 512 is the basic unit of the Linux block layer. It's the smallest granularity we ever can calculate with, and it's what libfdisk manages stuff with. Making that configurable makes no sense to me, it's a technical value of the Linux kernel.

512 is not how we align stuff though, we align to multiple of 4K always. This is slightly different from what libfdisk's own logic would do it which uses a "grain" of 1M instead.

Now, you query the native sector size of the backing device, and then use that as multiplicator for the values libfdisk returns (but which are in 512 units, always, according to docs), but that really makes no sense to me.

So let me ask you: what precisely are you "fixing" here? The commits look pretty wrong to me and come with no explanation. What's going on here? Please elaborate?

poettering:
(in case libfdisk docs are wrong and 512 is not necessarily the unit it expects "lba" values to be specified in, then we should query the 512 value via libfdisk apis, and not go directly to the block device, because libfdisk supports backends that are plain files, and not just blcok devices)

bluca:
repart wasn't working for me at all when trying to reproduce this problem, because it was trying to use 512 on a device that is actually 4096, and things just break

poettering:
So I got confirmation from Karel that fdisk_get_sector_size() is the factor, and not necessarily 512. So the patch' background makes sense, but it should not use the ioctl directly. And no global variables for this please.

tomty89:
I'm not willing to put more time or effort into this, at least not until I actually need to use it. (I filed this merely because I noticed the problem while I was fixing yet another one for someone else.) Besides, I don't really see why BLKSSZGET or how global variable should be avoided in this case.

Either way, if you are interested to merge this, feel free to add commit(s) that will polish it to your desire, else you may just close it.

poettering:
> Besides, I don't really see why BLKSSZGET or how global variable should be avoided in this case.

As mentioned libfdisk works with non-block device backends (specifically, regular files), and we shouldn't thus call BLKSSZGET on the fd just like that, but just use what libfdisk figured out for us.

And avoiding global variables is a matter of coding style, see 
https://systemd.io/CODING_STYLE/

We only use global variables for stuff that is truly global, i.e. process properties, or stuff parsed out from process parameters or so. But the sector size is a per-device property, and one day we might refactor repart to operate on multiple devices from the same process, who knows... We enforce this across the codebase: avoid global variables unless something is really really a global concept.

Hope that makes some sense.

poettering:
So my proposal is waiting in #22423. It uses libfdisk's own API, and also makes the 4K grain stuff dependent on the sector size. It adds a test based on losetup's -b switch to ensure this really and actually works.

PR #21825, opened 2021-12-19
boot: Add gdb support and documentation

medhefgo:
This will finally allow debugging issues in systemd without resorting to
Print() calls all over the place. Since this is not exactly trivial
to setup, add a helper script to run QEMU+gdb for us.

(I would appreciate if someone could confirm the instructions works for them too.)

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -330,3 +330,43 @@ To debug systemd components other than PID 1, set "program" to the full path of 
 debug and set "processId" to "${command:pickProcess}". Now, when starting the debugger, VSCode will ask you
 the PID of the process you want to debug. Run `systemctl show --property MainPID --value <component>` in the
 container to figure out the PID and enter it when asked and VSCode will attach to that process instead.
+
+# Debugging systemd-boot
+
+During boot, systemd-boot and the stub loader will output a message like `systemd-boot@0x0A,0x0B`,
+providing the location of the text and data sections. These location can then be used to attach
+to a QEMU session (provided it was run with `-s`) with these gdb commands:
+
+```
+    (gdb) file build/src/boot/efi/systemd-bootx64.efi
+    (gdb) add-symbol-file build/src/boot/efi/systemd_boot.so 0x0A -s .data 0x0B
+    (gdb) set architecture i386:x86-64
+    (gdb) target remote :1234
+```
+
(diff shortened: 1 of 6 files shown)


docs/HACKING.md
```diff
@@ -330,3 +330,71 @@ To debug systemd components other than PID 1, set "program" to the full path of
 debug and set "processId" to "${command:pickProcess}". Now, when starting the debugger, VSCode will ask you
 the PID of the process you want to debug. Run `systemctl show --property MainPID --value <component>` in the
 container to figure out the PID and enter it when asked and VSCode will attach to that process instead.
+
+# Hacking on systemd-boot
+
+To test changes to systemd-boot, the easiest way is to just run QEMU directly with a peroperly prepared
```
topimiettinen: peroperly -> properly

DaanDeMeyer:
I really think we should try to integrate this with mkosi. mkosi already handles building an image with EFI partition and sd-boot and invoking qemu for us, we shouldn't duplicate that here. Let's have mkosi build the image and invoke qemu. The debug script should just invoke mkosi with the extra arguments required to make sd-boot debugging work. It's already possible to pass extra qemu args to mkosi which should get us most of the way there I think.

medhefgo:
You may have a point there. But I honestly have no idea how mkosi works and wouldn't know where to start. Like how/where can I inject stuff into the efi partition it creates or can I supply my own? As far as I can tell, mkosi creates a full blown os image with a userspace to boot into (everytime?) which I would rather not deal with for simple bootloader testing.
Also, I have 0 python skills. :D

If we really don't want the (imho small) duplication, I don't mind keeping this script for myself and just leave a little helper for the gdb-script part. Someone interested in mkosi can pick up from there...

medhefgo:
For now I reduced the script to extraction of the info and handling the gdb/gdbscript.

evverx:
I think if it makes it easier for @medhefgo to test and debug `systemd-boot` it should be good to go.

> I really think we should try to integrate this with mkosi.

@DaanDeMeyer I'm not particularly familiar with mkosi but it seems something like that came up in https://github.com/systemd/systemd/pull/21960#issuecomment-1003587815. It would be great if it was possible to somehow use the mkosi action to test secure boot as well.

Issue #21826, opened 2021-12-19
Build failure with glibc-2.34 and -D_TIME_BITS=64


floppym:
**systemd version the issue has been seen with**
v250-rc2

**Used distribution**
Gentoo with glibc-2.34

**Linux kernel version used** (`uname -a`)
`Linux i686 5.15.7-dist #1 SMP Wed Dec 8 15:10:51 EST 2021 i686 AMD Phenom(tm) II X6 1055T Processor AuthenticAMD GNU/Linux`

**CPU architecture issue was seen on**
i686

**Steps to reproduce the problem**
Build with `CFLAGS="-D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64"`.

**Additional program output to the terminal or log subsystem illustrating the issue**
```text
FAILED: src/timesync/libtimesyncd-core.a.p/timesyncd-manager.c.o 
i686-pc-linux-gnu-gcc -Isrc/timesync/libtimesyncd-core.a.p -Isrc/timesync -I../systemd-250-rc2/src/timesync -Isrc/basic -I../systemd-250-rc2/src/basic -Isrc/fundamental -I../systemd-250-rc2/src/fundamental -Isrc/systemd -I..
/systemd-250-rc2/src/systemd -I. -I../systemd-250-rc2 -I../systemd-250-rc2/src/libsystemd/sd-bus -I../systemd-250-rc2/src/libsystemd/sd-device -I../systemd-250-rc2/src/libsystemd/sd-event -I../systemd-250-rc2/src/libsystemd/
sd-hwdb -I../systemd-250-rc2/src/libsystemd/sd-id128 -I../systemd-250-rc2/src/libsystemd/sd-journal -I../systemd-250-rc2/src/libsystemd/sd-netlink -I../systemd-250-rc2/src/libsystemd/sd-network -I../systemd-250-rc2/src/libsy
stemd/sd-resolve -Isrc/shared -I../systemd-250-rc2/src/shared -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -std=gnu99 -Wno-format-signedness -Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wendif
-labels -Werror=format=2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=overflow -Werror=override-init -Werror=return-type -Werror=shift-count-overflow -Werror=shift-o
[...]

PR #21827, opened 2021-12-19
timesync: assume struct timex member type matches time_t

floppym:
This appears to be enforced by glibc headers in bits/timex.h.

Fixes: https://github.com/systemd/systemd/issues/21826


floppym:
Closing in favor of #21829.

PR #21829, opened 2021-12-20
basic: add a size check to format timex members properly

floppym:
As of glibc-2.34, the size of members in struct timex varies depending on
the _TIME_BITS macro.

Fixes: https://github.com/systemd/systemd/issues/21826

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -468,6 +468,7 @@ conf.set('SIZEOF_DEV_T', cc.sizeof('dev_t', prefix : '#include <sys/types.h>'))
 conf.set('SIZEOF_INO_T', cc.sizeof('ino_t', prefix : '#include <sys/types.h>'))
 conf.set('SIZEOF_TIME_T', cc.sizeof('time_t', prefix : '#include <sys/time.h>'))
 conf.set('SIZEOF_RLIM_T', cc.sizeof('rlim_t', prefix : '#include <sys/resource.h>'))
+conf.set('SIZEOF_TIMEX_MEMBER', cc.sizeof('((struct timex *)0)->freq', prefix : '#include <sys/timex.h>'))
 
 decl_headers = '''
 #include <uchar.h>
--- a/src/basic/format-util.h
+++ b/src/basic/format-util.h
@@ -26,10 +26,12 @@ assert_cc(sizeof(gid_t) == sizeof(uint32_t));
 #  error Unknown time_t size
 #endif
 
-#if defined __x86_64__ && defined __ILP32__
+#if SIZEOF_TIMEX_MEMBER == 8
 #  define PRI_TIMEX PRIi64
-#else
+#elif SIZEOF_TIMEX_MEMBER == 4
 #  define PRI_TIMEX "li"
+#else
+#  error Unknown timex member size
 #endif
 
 #if SIZEOF_RLIM_T == 8


yuwata (approved):
LGTM.

asavah:
This check might not be correct and fails for me

```
Code:
 #include <stdio.h>
        #include <sys/timex.h>
        int main(void) {
            ((struct timex *)0)->freq something;
            return 0;
        }
Compiler stdout:
.
Compiler stderr:
 /home/asavah/kross/tmp/asusb450eg/systemd/meson-private/tmphlzlorxa/testfile.c: In function 'main':
/home/asavah/kross/tmp/asusb450eg/systemd/meson-private/tmphlzlorxa/testfile.c:4:38: error: expected ';' before 'something'
    4 |             ((struct timex *)0)->freq something;
      |                                      ^~~~~~~~~~
      |                                      ;

Checking for size of "((struct timex *)0)->freq" : -1
Running compile:
Working directory:  /home/asavah/kross/tmp/asusb450eg/systemd/meson-private/tmpr_wa5rzi
Command line:  ccache x86_64-asusb450eg-linux-gnu-gcc /home/asavah/kross/tmp/asusb450eg/systemd/meson-private/tmpr_wa5rzi/testfile.c -o /home/asavah/kross/tmp/asusb450eg/systemd/meson-private/tmpr_wa5rzi/output.obj -snipped-for-brevity
```

it's a cross compiling environment with glibc 2.34, gcc 11.2

Build fails afterwards:
```
In file included from ../../../src/systemd/src/basic/process-util.h:15,
                 from ../../../src/systemd/src/libsystemd/sd-bus/bus-container.c:11:
../../../src/systemd/src/basic/format-util.h:34:4: error: #error Unknown timex member size
   34 | #  error Unknown timex member size
      |    ^~~~~
```

floppym:
What version of meson are you using? I am using meson-0.60.2.

floppym:
Oh, I see. It works fine for native builds, but meson does an extra compile check when cross-compiling:

https://github.com/mesonbuild/meson/blob/0.60.2/mesonbuild/compilers/mixins/clike.py#L573

I think this code in meson is broken, but I guess we will need to work around it somehow.

asavah:
I have a (nasty) workaround for gcc, but I'm not sure about clang, I don't have a working clang cross environment to test.

```
cat 007-timex.patch
diff --git a/meson.build b/meson.build
index ff3e066e2c..6eae80608c 100644
--- a/meson.build
+++ b/meson.build
@@ -468,7 +468,7 @@ conf.set('SIZEOF_DEV_T', cc.sizeof('dev_t', prefix : '#include <sys/types.h>'))
 conf.set('SIZEOF_INO_T', cc.sizeof('ino_t', prefix : '#include <sys/types.h>'))
 conf.set('SIZEOF_TIME_T', cc.sizeof('time_t', prefix : '#include <sys/time.h>'))
 conf.set('SIZEOF_RLIM_T', cc.sizeof('rlim_t', prefix : '#include <sys/resource.h>'))
-conf.set('SIZEOF_TIMEX_MEMBER', cc.sizeof('((struct timex *)0)->freq', prefix : '#include <sys/timex.h>'))
+conf.set('SIZEOF_TIMEX_MEMBER', cc.sizeof('typeof(((struct timex *)0)->freq)', prefix : '#include <sys/timex.h>'))

 decl_headers = '''
 #include <uchar.h>
```

`SIZEOF_TIMEX_MEMBER` is 8 after this check.

heitbaum:
With this https://github.com/systemd/systemd/pull/21829#issuecomment-998257503 “patch” my builds succeed too.

PR #21835, opened 2021-12-20
hwdb: update for -rc3

keszybz:
Just a small bunch of additions and a naming updates.


keszybz:
rpm-build:fedora-rawhide-i386: the µhttpd deprecation strikes again…

keszybz:
Let's merge this. The only thing that could realistically fail would be the parse-hwdb test, and it passes here.

PR #21837, opened 2021-12-20
journal: Handle partially read HashItem's when punching holes

daandemeyer:


src/journal/journald-file.c
```diff
@@ -97,6 +97,9 @@ static int journald_file_punch_holes(JournalFile *f) {
                 if (n < 0)
                         return n;
 
+                /* Let's ignore any partial hash items by rounding down to the nearest multiple of HashItem. */
+                n -= n % sizeof(HashItem);
```
keszybz, line 101: I don't understand. Isn't the loop below enough to achieve the same effect? `(size_t) n / sizeof(HashItem)` rounds down too…

yuwata (approved):
LGTM.

src/journal/journald-file.c
yuwata: If `n` is not a multiple of `sizeof(HashItem)`, then the offset of next `pread()` becomes strange.

keszybz: Hmm, doesn't this increase the risk that we end up in an infinite loop here? If the file is strangely truncated, pread() could return something between 0 and sizeof(HashItem). Shouldn't the loop above if `n < 0` or the rounded-down `n` is 0?

yuwata: Indeed.

daandemeyer: Added a n > 0 check to the outer loop.

yuwata (approved):
LGTM.

PR #1227, opened 2015-09-09
systemctl: fix various aspects of polkit authorization in legacy tools.

intelfx:
This deals with multiple concerns raised in the issue #213.

First, disallow interactive polkit auth(orization|entication) if we're being called not as `systemctl` but as one of the legacy tools (`halt`, `poweroff`, `reboot`, `shutdown`, `telinit`).

Then, relax preliminary root permission check in `halt_main()` to allow (non-interactive) polkit auth to kick in the "activate special target" code path (i. e. non-root non-logind non-delayed non-forced mode).

Finally, there are three ad-hoc fixes:
- fix logind bus call error handling in `halt_main()` to make it look similar to the call in `start_special()`,
- improve error messages in the logind call code path (`reboot_with_logind()`) to make it clear for the user which exact operation had failed,
- and (while at it) refactor repeated comparisons into applications of `IN_SET()`.

(Maybe I'm overzealous with the last one.)

Issue #213, opened 2015-06-15
`halt`/`poweroff`/`reboot` are not identical to `systemctl %s` with respect to non-root users


intelfx:
Verbs `systemctl halt`, `systemctl poweroff` and `systemctl reboot`, if invoked under a non-root user and logind is not available, are governed by polkit action `org.freedesktop.systemd1.manage-units`. 

(For `systemctl halt` logind is obviously never available, as there's no corresponding bus method.)

However, `halt`, `poweroff` and `reboot` in the same conditions say "Must be root." and bail out. Shouldn't they be equivalent to `systemctl %s` in this regard?

kaysievers:
They are system V compat, and should not be "smarter" than the original tools.

intelfx:
Hm, but they already divert to logind if it's available on the bus... which itself is an enhancement over sysv semantics, IIRC.

kaysievers:
Hmm, yeah, we did that to cover for Fedora's usermode tool (broken idea) which grants access to some tools for locally logged-in users. Not sure, what's the right answer here. @poettering?

poettering:
We cannot make poweroff/reboot suddenly interactive when they never were before. People call these binaries from daemons and scripts and that would be quite a change of behaviour. For "systemctl poweroff" and friends this is different, because we define the API and can provide --no-ask-password to remove the interactivity (and we do). But we really shouldn't extend the old APIs with new switches or different default behaviour.

poettering:
OK, I guess the checks in systemctl actually should be losened and also allow talking to PID1's StartUnit("reboot.target") now for unprivileged clients. After all we recently opened up StartUnit() to polkit, and hence it might be useful for unprivileged clients too. 

Of course, when we change systemctl like this I think we should only allow non-interactive polkit auth still for these tools, and leave interactive auth for systemctl poweroff and friends.

I hope that makes sense?

i have reopened the bug accordingly.

intelfx:
At least `poweroff` and `reboot` (and `telinit`, btw) _currently_ may end up in interactive polkit auth kicking in. First two may divert to logind, and the last one is straight equivalent to `systemctl isolate`.

So — do you intend to remove existing places where interactive auth is allowed?

poettering:
Yeah, the old tools should _never_ be interactive. If they end up being interactive now, that'd be a bug.

intelfx:
[...]

PR #21838, opened 2021-12-20
Logind shutdown refactor

lnussel:
Did the refactor last to show where this is coming from. The previous commits could be picked separately. Esp the shutdown -c fix makes sense right away IMO.

### DIFF
--- a/test/test-shutdown.py
+++ b/test/test-shutdown.py
@@ -1,0 +1,114 @@
+#!/usr/bin/python3
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+
+import argparse
+import logging
+import pexpect
+import sys
+
+
+def run(args):
+
+    ret = 1
+    logger = logging.getLogger("test-shutdown")
+
+    logger.info("spawning test")
+    console = pexpect.spawn(args.command, args.arg, env={
+            "TERM": "linux",
+        }, encoding='utf-8', timeout=30)
+
+    if args.verbose:
+        console.logfile = sys.stdout
+
+    logger.debug("child pid %d" % console.pid)
+
+    try:
+        logger.info("waiting for login prompt")
+        console.expect('H login: ', 10)
+
+        logger.info("log in and start screen")
+        console.sendline('root')
+        console.expect('bash.*# ', 10)
+        console.sendline('screen')
+        console.expect('screen0 ', 10)
+        console.sendcontrol('a')
+        console.send('c')
+        console.expect('screen1 ', 10)
+
+#        console.interact()
+
+        console.sendline('tty')
+        console.expect(r'/dev/(pts/\d+)')
+        pty = console.match.group(1)
+        logger.info("window 1 at line %s", pty)
+
+        logger.info("schedule reboot")
+        console.sendline('shutdown -r')
+        console.expect("Reboot scheduled for (?P<date>.*), use 'shutdown -c' to cancel", 2)
+        date = console.match.group('date')
+        logger.info("reboot scheduled for %s", date)
+
+        console.sendcontrol('a')
+        console.send('0')
+        logger.info("verify broadcast message")
+        console.expect('Broadcast message from root@H on %s' % pty, 2)
+        console.expect('The system is going down for reboot at %s' % date, 2)
+
+        logger.info("check show output")
+        console.sendline('shutdown --show')
+        console.expect("Reboot scheduled for %s, use 'shutdown -c' to cancel" % date, 2)
+
+        logger.info("cancel shutdown")
+        console.sendline('shutdown -c')
+        console.sendcontrol('a')
+        console.send('1')
+        console.expect('The system shutdown has been cancelled', 2)
+
+        logger.info("call for reboot")
+        console.sendline('sleep 10; shutdown -r now')
+        console.sendcontrol('a')
+        console.send('0')
+        console.expect("The system is going down for reboot NOW!", 12)
+
+        logger.info("waiting for reboot")
+
+        console.expect('H login: ', 10)
+        console.sendline('root')
+        console.expect('bash.*# ', 10)
+
+        console.sendline('> /testok')
+
+        logger.info("power off")
+        console.sendline('poweroff')
+
+        logger.info("expect termination now")
+        console.expect(pexpect.EOF)
+
+        ret = 0
+    except Exception as e:
+        logger.error(e)
+        logger.info("killing child pid %d" % console.pid)
+        console.terminate()
+
+    return ret
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='test logind shutdown feature')
+    parser.add_argument("-v", "--verbose", action="store_true", help="verbose")
+    parser.add_argument("command", help="command to run")
+    parser.add_argument("arg", nargs='*', help="args for command")
+
+    args = parser.parse_args()
+
+    if args.verbose:
+        level = logging.DEBUG
+    else:
+        level = logging.INFO
+
+    logging.basicConfig(level=level)
+
+    sys.exit(run(args))
+
+# vim: sw=4 et
--- a/test/TEST-69-SHUTDOWN/Makefile
+++ b/test/TEST-69-SHUTDOWN/Makefile
@@ -1,0 +1,1 @@
+../TEST-01-BASIC/Makefile
--- a/test/TEST-69-SHUTDOWN/test.sh
+++ b/test/TEST-69-SHUTDOWN/test.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+
+TEST_DESCRIPTION="shutdown testing"
+IMAGE_NAME="shutdown"
+TEST_NO_QEMU=1
+
+# shellcheck source=test/test-functions
+. "${TEST_BASE_DIR:?}/test-functions"
+
+_ORIG_NSPAWN="$SYSTEMD_NSPAWN"
+SYSTEMD_NSPAWN="$STATEDIR/run-nspawn"
+
+setup_nspawn_root_hook() {
+    cat > "$STATEDIR"/run-nspawn <<-EOF
+	#!/bin/bash
+	exec "$TEST_BASE_DIR"/test-shutdown.py -- "$_ORIG_NSPAWN" "\$@"
+	exit 1
+	EOF
+    chmod 755 "$STATEDIR"/run-nspawn
+}
+
+test_append_files() {
+    # prevent shutdown in test suite, the expect script does that manually.
+    rm "$1"/usr/lib/systemd/tests/testdata/units/end.service
+    inst /usr/bin/screen
+    echo "PS1='screen\$WINDOW # '" > "$1"/etc/bash.bashrc
+    echo 'startup_message off' > "$1"/etc/screenrc
+    echo 'bell_msg ""' >> "$1"/etc/screenrc
+}
+
+do_test "$@"
--- a/test/test-functions
+++ b/test/test-functions
@@ -1884,6 +1884,8 @@ has_user_dbus_socket() {
     fi
 }
 
+setup_nspawn_root_hook() { :;}
+
 setup_nspawn_root() {
     if [ -z "${initdir}" ]; then
         dfatal "\$initdir not defined"
@@ -1896,6 +1898,8 @@ setup_nspawn_root() {
         ddebug "cp -ar $initdir $TESTDIR/unprivileged-nspawn-root"
         cp -ar "$initdir" "$TESTDIR/unprivileged-nspawn-root"
     fi
+
+    setup_nspawn_root_hook
 }
 
 setup_basic_dirs() {
--- a/test/units/testsuite-69.service
+++ b/test/units/testsuite-69.service
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=TEST-69-SHUTDOWN
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/systemd/tests/testdata/units/%N.sh


lnussel:
Added  test case. It uses wraps nspawn and uses pexpect to log in on the console. Then spawn screen to have two ptys where the script can check the wall message.

marcosfrm:
https://github.com/systemd/systemd/issues/213
https://github.com/systemd/systemd/pull/1227

lnussel:
Ic. So I shall remove the commit here and potentially have  separate discussion about it I guess

mrc0mmand:
I installed the necessary dependencies for TEST-69 in CentOS CI, but the test still keeps failing on CentOS Stream 8. Also, the deps need to be installed in Ubuntu CI as well (/cc @bluca) - `python-pexpect` and `screen`.

bluca:
added and re-triggered the focal-amd64 run

src/login/logind-utmp.c
```diff
@@ -135,11 +135,11 @@ int manager_setup_wall_message_timer(Manager *m) {
                 return 0;
         }
 
-        if (elapse < n)
+        if (elapse && elapse < n)
```
keszybz: Please use `if (elapse > 0 && elepse < n)`. See CODING_STYLE, we use "boolean operations" only for booleans and pionters (and sometimes for convenience, but only in limited circumstances).

src/login/logind-utmp.c
```diff
@@ -135,11 +135,11 @@ int manager_setup_wall_message_timer(Manager *m) {
...
                 return 0;
 
         /* Warn immediately if less than 15 minutes are left */
-        if (elapse - n < 15 * USEC_PER_MINUTE) {
+        if (!elapse || elapse - n < 15 * USEC_PER_MINUTE) {
```
keszybz: `if (elapse == 0 || …`

src/login/logind-action.c
```diff
@@ -135,11 +220,7 @@ int manager_handle_action(
                                        inhibit_what_to_string(m->action_what),
                                        handle_action_to_string(handle));
 
-        assert_se(target = manager_target_for_action(handle));
-
-        inhibit_operation = IN_SET(handle, HANDLE_SUSPEND, HANDLE_HIBERNATE,
-                                           HANDLE_HYBRID_SLEEP,
-                                           HANDLE_SUSPEND_THEN_HIBERNATE) ? INHIBIT_SLEEP : INHIBIT_SHUTDOWN;
+        inhibit_operation = manager_item_for_handle(handle)->inhibit_what;
```
keszybz, line 229: This assumes `manager_item_for_handle` cannot return NULL. Shouldn't `manager_item_for_handle` just have
```C
assert(handle < (ssize_t) ELEMENTSOF(action_table));
```
?

src/login/logind-dbus.c
```diff
@@ -1488,59 +1493,32 @@ static int have_multiple_sessions(
...
         else
-                p = strjoina(p, " (", m->wall_message, ").");
+                message = "MESSAGE=System is shutting down";
 
-        return log_struct(LOG_NOTICE, d, p, q);
-}
+        if (isempty(m->wall_message))
+                message = strjoina(message, ".");
+        else
+                message = strjoina(message, " (", m->wall_message, ").");
```
keszybz, line 1518: Hmm, this is a user-controlled message (via polkit). So I don't think we should use `strjoina` here, but normal `strjoin` instead. This is not a very pressing issue, but it's a bit sloppy.

src/login/logind-action.c
```diff
@@ -11,29 +11,115 @@
...
 #include "process-util.h"
-#include "sleep-config.h"
+#include "sd-messages.h"
 #include "special.h"
 #include "string-table.h"
 #include "terminal-util.h"
 #include "user-util.h"
 
+static const ActionTableItem action_table[_HANDLE_ACTION_MAX] = {
+        [HANDLE_POWEROFF] = {
```
keszybz, line 22: I wonder if we need HANDLE_POWEROFF and the other enum entries. Wouldn't it be simpler to pass around `const ActionTableItem*` ?

keszybz (commented):
Thanks for working on this. Looks like a refactoring was overdue.

src/login/logind-action.c
```diff
@@ -11,29 +11,115 @@
 #include "logind-dbus.h"
 #include "logind-session-dbus.h"
 #include "process-util.h"
-#include "sleep-config.h"
+#include "sd-messages.h"
```
keszybz: Please put this in a separate "block":
```C
#include <unistd.h>

#include "sd-messages.h"

#include "alloc-util.h"
#include "bus-error.h"
...
```

src/login/logind-utmp.c
lnussel: ack

src/login/logind-utmp.c
lnussel: ack

src/login/logind-action.c
lnussel: indeed. I took that from manager_target_for_action() which weirdly had an assert for negative values, ie _HANDLE_ACTION_INVALID not passed but returned NULL for too large values.

src/login/logind-dbus.c
lnussel: Right. Maybe it makes sense to even impose a length limit in method_set_wall_message()? desktop users have implicit permission to set the wall message. Not sure it's too useful to allow dumping a novel in there :-)

src/login/logind-action.c
lnussel: Probably. There's one or two places where it checks for a specific action but that could also be done with calculating the offset of the pointer from action_table I guess.

src/login/logind-action.c
lnussel: ack

src/login/logind-dbus.c
lnussel: There's actually a limit in plain old sysv shutdown. So I've added the same to logind now. https://git.savannah.nongnu.org/cgit/sysvinit.git/tree/src/shutdown.c

src/login/logind-action.c
lnussel: changed the code to use the pointer where possible. I guess the use could be expanded further for someone who doesn't dare to also touch logind-button etc. :-)

src/login/logind-action.c
lnussel: added assert()

lnussel:
updated based on previous comments

src/login/logind-dbus.c
```diff
@@ -1854,6 +1854,30 @@ static int verify_shutdown_creds(
...
+        if (r >= 0) {
+                const char *tty = NULL;
+
+                (void) sd_bus_creds_get_uid(creds, &m->scheduled_shutdown_uid);
+                (void) sd_bus_creds_get_tty(creds, &tty);
+
+                r = free_and_strdup(&m->scheduled_shutdown_tty, tty);
+                if (r < 0) {
+                        return log_oom();
+                }
```
yuwata: Unnecessary `{}`

src/systemctl/systemctl-compat-halt.c
```diff
@@ -144,44 +144,36 @@ int halt_parse_argv(int argc, char *argv[]) {
...
-                return 0;
-
-        /* No delay, or logind failed or is not at all available */
-        if (geteuid() != 0) {
-                if (arg_dry_run || arg_force > 0) {
-                        (void) must_be_root();
-                        return -EPERM;
-                }
+        /* always try logind first */
+        if (arg_when > 0) {
```
yuwata: Unnecessary '{}'

src/systemctl/systemctl-compat-halt.c
```diff
@@ -144,44 +144,36 @@ int halt_parse_argv(int argc, char *argv[]) {
...
-                                /* Requested operation is not supported on the local system or already in
-                                 * progress */
-                                return r;
-
-                        /* on all other errors, try low-level operation */
-                }
+                r = logind_reboot(arg_action);
         }
 
+        if (r >= 0)
```
yuwata: Not a big matter, but we usually do not add any blank lines between function call and error check.

src/systemctl/systemctl-compat-halt.c
```diff
@@ -144,44 +144,36 @@ int halt_parse_argv(int argc, char *argv[]) {
...
+                return r;
+        if (IN_SET(r, -EOPNOTSUPP, -EINPROGRESS))
+                /* Requested operation is not supported on the local system or already in
+                 * progress */
+                return r;
+        /* on all other errors, try low-level operation */
```
yuwata, line 163: Maybe, better to add a debugging log?

src/login/logind-dbus.c
```diff
@@ -1958,6 +1958,8 @@ static int method_do_shutdown_or_sleep(
 static int method_poweroff(sd_bus_message *message, void *userdata, sd_bus_error *error) {
         Manager *m = userdata;
 
+        m->scheduled_shutdown_type = strdup("poweroff");
```
yuwata: Maybe, add one more argument to `method_do_shutdown_or_sleep()` to take type string?

Also, it is better to use `free_and_strdup()`.

src/login/logind-dbus.c
```diff
@@ -2324,20 +2324,49 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
 static int method_cancel_scheduled_shutdown(sd_bus_message *message, void *userdata, sd_bus_error *error) {
         Manager *m = userdata;
+        const char *action;
         bool cancelled;
+        int r;
 
         assert(m);
         assert(message);
 
         cancelled = m->scheduled_shutdown_type != NULL;
```
yuwata: Could you drop `! = NULL`? Our coding style allows implicit cast from pointer to boolean.

src/login/logind-action.c
```diff
@@ -11,31 +13,122 @@
...
+                NULL
+                },
+};
 
+const char* manager_target_for_action(HandleAction handle) {
         assert(handle >= 0);
-        if (handle < (ssize_t) ELEMENTSOF(target_table))
-                return target_table[handle];
+        assert(handle < (ssize_t) ELEMENTSOF(action_table));
+        if (handle < (ssize_t) ELEMENTSOF(action_table))
```
yuwata: That's redundant...

src/login/logind-action.c
```diff
@@ -11,31 +13,122 @@
...
 
+const ActionTableItem* manager_item_for_handle(HandleAction handle) {
+        assert(handle >= 0);
+        assert(handle < (ssize_t) ELEMENTSOF(action_table));
+
+        return &action_table[handle];
+}
+
+HandleAction manager_handle_for_item(const ActionTableItem* a) {
+        if (a != NULL && a < action_table + ELEMENTSOF(action_table))
```
yuwata: `a != NULL` -> `a`

src/login/logind-action.c
```diff
@@ -129,17 +221,13 @@ int manager_handle_action(
                 return log_warning_errno(SYNTHETIC_ERRNO(EOPNOTSUPP),
                                          "Requested %s operation not supported, ignoring.", handle_action_to_string(handle));
 
-        if (m->action_what > 0)
+        if (m->delayed_action != NULL)
```
yuwata: Please drop `!= NULL`

src/login/logind-dbus.c
```diff
@@ -1488,59 +1495,32 @@ static int have_multiple_sessions(
         return false;
 }
 
-_printf_(2, 0)
-static int log_with_wall_message(Manager *m, const char *d, const char *p, const char *q) {
+static int bus_manager_log_shutdown(
+                Manager *m,
+                const ActionTableItem *a) {
+
+        const char *message = a->message;
```
yuwata: It is better to add an assertion for `a`:
```
const char *message, *log_str;

assert(m);
assert(a);

message = a->message;
log_str = a->log_str;

File: src/login/logind-dbus.c [hunk 17]
```diff
@@ -1488,59 +1495,32 @@ static int have_multiple_sessions(
...
+        const char *log_str = a->log_str;
+
         assert(m);
 
-        if (isempty(m->wall_message))
-                p = strjoina(p, ".");
+        if (message)
```
yuwata, line 1510: `isempty()`?

src/login/logind-dbus.c
```diff
@@ -1660,10 +1636,10 @@ int manager_dispatch_delayed(Manager *manager, bool timeout) {
 
         assert(manager);
 
-        if (manager->action_what == 0 || manager->action_job)
+        if (manager->delayed_action == NULL || manager->action_job)
```
yuwata: `!manager->delayed_action`

src/login/logind-dbus.c
```diff
@@ -1921,34 +1897,41 @@ static int method_do_shutdown_or_sleep(
                 flags = interactive ? SD_LOGIND_INTERACTIVE : 0;
         }
 
-        if ((flags & SD_LOGIND_REBOOT_VIA_KEXEC) && kexec_loaded())
-                unit_name = SPECIAL_KEXEC_TARGET;
+        if ((flags & SD_LOGIND_REBOOT_VIA_KEXEC) && kexec_loaded()) {
+                a = manager_item_for_handle(HANDLE_KEXEC);
+        }
```
yuwata: Please drop `{}`.

src/login/logind-dbus.c
```diff
@@ -1921,34 +1897,41 @@ static int method_do_shutdown_or_sleep(
...
 
         /* Don't allow multiple jobs being executed at the same time */
-        if (m->action_what > 0)
+        if (m->delayed_action != NULL)
```
yuwata: Please drop `!= NULL`

src/login/logind-dbus.c
```diff
@@ -2173,31 +2106,20 @@ static int manager_scheduled_shutdown_handler(
...
-        else if (streq(m->scheduled_shutdown_type, "halt"))
-                target = SPECIAL_HALT_TARGET;
-        else
-                assert_not_reached();
+        a = m->scheduled_shutdown_type;
+        assert(a);
 
         /* Don't allow multiple jobs being executed at the same time */
-        if (m->action_what > 0) {
+        if (m->delayed_action != NULL) {
```
yuwata: Please drop `!= NULL`

src/login/logind-dbus.c
```diff
@@ -3324,7 +3190,7 @@ static int method_inhibit(sd_bus_message *message, void *userdata, sd_bus_error
          * executing the operation. We shouldn't create the impression
          * that the lock was successful if the machine is about to go
          * down/suspend any moment. */
-        if (m->action_what & w)
+        if (m->delayed_action != NULL && m->delayed_action->inhibit_what & w)
```
yuwata: Also here.

src/login/logind-utmp.c
```diff
@@ -130,10 +130,8 @@ int manager_setup_wall_message_timer(Manager *m) {
 
         /* wall message handling */
 
-        if (isempty(m->scheduled_shutdown_type)) {
-                warn_wall(m, n);
+        if (m->scheduled_shutdown_type == NULL)
```
yuwata: Also here.

test/test-shutdown.py
```diff
@@ -0,0 +1,116 @@
...
+#
+# Copyright (c) 2021 SUSE LLC
+#
+
+import argparse
+import logging
+import pexpect
+import sys
+
+
```
yuwata, line 10: doubled blank line.

test/test-shutdown.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/python3
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (c) 2021 SUSE LLC
+#
```
yuwata: We usually do not add Copyright for a new file anymore.

test/test-shutdown.py
```diff
@@ -0,0 +1,116 @@
...
+
+        ret = 0
+    except Exception as e:
+        logger.error(e)
+        logger.info("killing child pid %d" % console.pid)
+        console.terminate()
+
+    return ret
+
+
```
yuwata, line 96: Doubled.

yuwata (changes requested):
Nice. Several coding style nits.

src/systemctl/systemctl-compat-halt.c
lnussel: *shrug* wasn't there before either :) Would be separate task to reproduce hitting that condition and deciding what's best.

src/login/logind-dbus.c
lnussel: I'll rewrite it to free_and_strdup() as that's easier for rebasing if that's ok. Will be removed with later refactoring anyway.

src/login/logind-dbus.c
lnussel: Makes sense for conditions, I overlooked that. Even in the assignment case though? Should I add a cast instead so it's clear we don't try to assign a pointer?

test/test-shutdown.py
lnussel: python coding style likes that though. flake8 complains otherwise :)

test/test-shutdown.py
lnussel: dito. looks like the other python files in the test dir are not flake8's favorite either. I don't mind either way. If you insist I'll remove the blank lines.

src/login/logind-dbus.c
yuwata: I see. Then, `free_and_strdup()` sounds enough.

test/test-shutdown.py
yuwata: OK.

src/login/logind-dbus.c
yuwata: I do not have any strong opinions. But I prefer the current style than explicit cast to `bool`. Let's keep this as is.

lnussel:
Working on it. GH fooled me by hiding he big diff so didn't see all :)

src/login/logind-dbus.c
lnussel: hmm, it's a static built in table, so unless someone intentionally puts "" in the source it's either NULL or something IMO.

yuwata: Ah, I see.

yuwata (approved):
Thank you! LGTM.

src/systemctl/systemctl-compat-halt.c
lnussel: AFAICT the functions above log their error conditions themselves so no need to do so here.

yuwata: Thank you for checking that!

src/login/logind-action.c
```diff
@@ -11,29 +13,119 @@
...
-        return NULL;
+        assert(handle < (ssize_t) ELEMENTSOF(action_table));
+
+        return &action_table[handle];
+}
+
+HandleAction manager_handle_for_item(const ActionTableItem* a) {
+        if (a && a < action_table + ELEMENTSOF(action_table))
+                return a - action_table;
+        return _HANDLE_ACTION_INVALID;
```
dnicolodi, line 128: This looks very hackish to me. I would simply include the enum value in a member of the `ActionTableItem` structure. At least I would add `assert(a >= action_table && a < action_table + ELEMENTSOF(action_table))` here as calling this function with anything not satisfying these constraints is a misuse of the function.

lnussel: Yes, there are tons of improvements that can be applied still and I didn't even get to the thing I actually wanted to do :-) As I wrote elsewhere the rabbit hole is deep. When is a pr considered enough of an improvement to merge it though? If I change this, will someone else chime in and demand yet another change?

yuwata: OK. Let's merge this as is. If necessary, let's apply followups.

dnicolodi: I would agree with you if this would be existing code. However, this is code you are introducing with this PR, thus I would much prefer if it would not add a questionable code pattern that would need yet another PR to clean up. However, I don't have a say in whether this PR should be accepted as it is or not.

yuwata: At least, the function checks NULL, so it satisfies minimum criteria.

We usually introduce functions like the following way.
```
void function(Object *obj) {
    assert(obj);

    ...;
}
```
Here, it avoids only case that `obj == NULL`. But it can easily triggers segfault or so by e.g.
```
Object obj;

function((Object*) ((uint8_t*) &obj + 1));
```

The situation @dnicolodi concerns sounds a kind of such case for me.

yuwata: Note, I do not say @dnicolodi's suggestion is bad. I think it is not a strong requirement for this PR being merged.
If you'd like, please submit a followup PR.

lnussel: will do. Thank you!

src/login/logind-dbus.c
```diff
@@ -2323,20 +2323,49 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
 
         assert(m);
         assert(message);
 
         cancelled = m->scheduled_shutdown_type != NULL;
+
+        if (!cancelled)
+                goto done;
+
+        // mirrors code in method_schedule_shutdown()
```
poettering: as per coding style, please use `/* */` style comments for explanatory stuff, and `//` style comments for "FIXME" style stuff.

See https://systemd.io/CODING_STYLE/#formatting

src/login/logind-dbus.c
```diff
@@ -2323,20 +2323,49 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
+        if (streq(m->scheduled_shutdown_type, "poweroff")) {
```
poettering: as per coding style, use {} for single-line if blocks

src/login/logind-dbus.c
```diff
@@ -2323,20 +2323,49 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
+                action = "org.freedesktop.login1.power-off";
+        } else if (STR_IN_SET(m->scheduled_shutdown_type, "reboot", "kexec")) {
+                action = "org.freedesktop.login1.reboot";
+        } else if (streq(m->scheduled_shutdown_type, "halt")) {
+                action = "org.freedesktop.login1.halt";
```
poettering: given this is shared, please add helper:

```c
static const char *shutdown_type_to_polkit_action(const char *);
```

or so, that can be used at both places and nicely abstracts this away

src/login/logind-dbus.c
```diff
@@ -2323,20 +2323,49 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
         bool cancelled;
+        int r;
 
         assert(m);
         assert(message);
 
         cancelled = m->scheduled_shutdown_type != NULL;
+
+        if (!cancelled)
+                goto done;
```
poettering, line 2245: goto is useful, but I'd always avoid it, hence maybe write this as:

```c
if (!m->scheduled_shutdown_type)
        return sd_bus_reply_method_return(message, "b", false);
```

and then change the last line to:

```c
return sd_bus_reply_method_return(message, "b", true);
```

src/login/logind-action.c
```diff
@@ -11,29 +13,119 @@
...
 #include "process-util.h"
-#include "sleep-config.h"
 #include "special.h"
 #include "string-table.h"
 #include "terminal-util.h"
 #include "user-util.h"
 
+static const ActionTableItem action_table[_HANDLE_ACTION_MAX] = {
+        [HANDLE_POWEROFF] = {
+                SPECIAL_POWEROFF_TARGET,
```
poettering, line 23: we'd usually use C99 structured initialization for things like this, to make this less opaque and make it easy to refactor the structure one day. i.e. i'd prefer if this was written as:

```
…
        [HANDLE_POWEROFF] = {
                .target = …,
                .inhibit_what = …,
        }
…
```

Yes, this will be longer, but I think it still makes it more readable given how many fields this has (and how non-obvious they are). Moreover, we kinda do this in most other similar cases (though admittedly, not in all)

src/login/logind-action.c
```diff
@@ -11,29 +13,119 @@
...
-                [HANDLE_SUSPEND] = SPECIAL_SUSPEND_TARGET,
-                [HANDLE_HIBERNATE] = SPECIAL_HIBERNATE_TARGET,
-                [HANDLE_HYBRID_SLEEP] = SPECIAL_HYBRID_SLEEP_TARGET,
-                [HANDLE_SUSPEND_THEN_HIBERNATE] = SPECIAL_SUSPEND_THEN_HIBERNATE_TARGET,
-                [HANDLE_FACTORY_RESET] = SPECIAL_FACTORY_RESET_TARGET,
-        };
+        assert(handle >= 0);
+        assert(handle < (ssize_t) ELEMENTSOF(action_table));
 
+        return action_table[handle].target;
```
poettering, line 115: simplify to:

```c
const char* manager_target_for_action(HandleAction handle) {
        return manager_item_for_handle(handle)->target;
}
```

(that said, I am not convinced we really need this extra helper...)

src/login/logind-action.c
```diff
@@ -11,29 +13,119 @@
...
-        if (handle < (ssize_t) ELEMENTSOF(target_table))
-                return target_table[handle];
-        return NULL;
+        assert(handle < (ssize_t) ELEMENTSOF(action_table));
+
+        return &action_table[handle];
+}
+
+HandleAction manager_handle_for_item(const ActionTableItem* a) {
+        if (a && a < action_table + ELEMENTSOF(action_table))
```
poettering, line 126: this should also check that `a >= action_table` holds

src/login/logind-action.h
```diff
@@ -19,8 +19,26 @@ typedef enum HandleAction {
...
+        const char *target;
+        InhibitWhat inhibit_what;
+        const char *polkit_action;
+        const char *polkit_action_multiple_sessions;
+        const char *polkit_action_ignore_inhibit;
+        SleepOperation sleep_operation;
+        const char* message_id;
+        const char* message;
+        const char* log_str;
+
```
poettering, line 40: spurious neline

src/login/logind-action.h
```diff
@@ -19,8 +19,26 @@ typedef enum HandleAction {
...
+struct ActionTableItem {
+        const char *target;
+        InhibitWhat inhibit_what;
+        const char *polkit_action;
+        const char *polkit_action_multiple_sessions;
+        const char *polkit_action_ignore_inhibit;
+        SleepOperation sleep_operation;
+        const char* message_id;
+        const char* message;
+        const char* log_str;
```
poettering, line 39: i'd call this `log_message`

we usually don't abbreviate in structure field naming needlessly. We do so in local variables, i.e. in minimal scope. But structs are handed around everywhere, thus the extra verbosity is helpful

src/login/logind-dbus.c
```diff
@@ -3256,6 +3115,15 @@ static int method_set_wall_message(
         if (r < 0)
                 return r;
 
+        /* sysvinit has a 252 (256-(strlen(" \r\n")+1)) character
+         * limit for the wall message. There is no real technical
+         * need for that but doesn't make sense to store arbitrary
+         * armounts either.
+         * https://git.savannah.nongnu.org/cgit/sysvinit.git/tree/src/shutdown.c#n72)
+        */
+        if (strlen(wall_message) > 252)
```
poettering, line 3124: given how arbitrary this number is, please add a macro for it.

Please return a proper dbus error i.e.

```c
return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Wall message too long, maximum permitted length is %u characters.", WALL_MESSAGE_MAX);
```

src/login/logind-dbus.c
```diff
@@ -3256,6 +3115,15 @@ static int method_set_wall_message(
...
+                return -EMSGSIZE;
```
poettering, line 3125: this is kinda a compat breakage: so far we accepted that without complaining, suddenly if someone passes a long string we'll just fail.

it probably makes sense to enforce a limit here, but maybe larger? 252 sounds awfully short. maybe enforce 4k or so? and then document in the man page that it would be wise not to pass more than 252 because this is a sysvinit interface?

poettering (commented):
Looks OK in general, just some minor things.

(Sorry for the post-merge review, been on parental leave the past months)

poettering:
(I understand that some of the points I raised are already fixed in the meantime by later commits, ignor those then)

lnussel:
Sure, no worries. Congrats, hope you had a good time :)

src/login/logind-dbus.c
lnussel: yup, already fixed in follow up

src/login/logind-dbus.c
lnussel: will do

src/login/logind-action.c
lnussel: ok, will do. Wasn't sure about that as the whole thing is quite big.

src/login/logind-action.c
lnussel: right, only one consumer left actually, will remove

src/login/logind-dbus.c
lnussel: The man page of systemctl actually states "set a short message". Doesn't say how short is short though so could be 160 as well :-) 4k bytes is quite a novel: https://lipsum.com/feed/html?amount=4096&what=bytes
Out of all the arbitrary numbers we could pull out of thin air the 252 has some anecdotal reference to old crap at least :-) I don't mind really though. So if you like 4k I'll change it to 4096 :-)

poettering: I think we should pick a limit that is unlikely to interfere with what people actually intend to do. i.e. so if we assume people want to write a word we'll accept a sentence and if they want a sentence we'll accept a paragraph, and if they want a paragraph we'll take a novel. thus, it doesn't really matter if 4k is a novel, what matters if 252 is enough for the usual sentences in the world's many languages. And I doubt it would be, given UTF8, and stuff. Thus I'd suggest 4K precisely for the reason that it's unlikely people will need that much. hope that thinking makes sense?

lnussel: sure

Issue #21842, opened 2021-12-21
FTBFS on armhf in efi


mbiebl:
Version: 250-rc3

The latest version of systemd fails to build on armhf
```
[...]

PR #21849, opened 2021-12-21
boot: Fix armhf build failure

medhefgo:
Fixes: #21842

I've only tested this in theory (using compiler explorer with a minimal test case).


bluca (approved):
can confirm build is fixed by this

PR #21850, opened 2021-12-21
machined: set TTYPath for container shell

lnussel:
TTYPath is needed for proper utmp registration of the shell to
receive wall messages.


lnussel:
not sure what the test suite fail is about. looks unrelated.

PR #21851, opened 2021-12-21
Accept `-Ddefault-user-shell=[/bin/bash]`, use it as the fallback non-root user shell and nspawn interactive session shell

nabijaczleweli:
~~This leaves k-i the only hard bash dep~~ This removes the final hard bash dependency (user-record). No changes are expected in the default configuration.

### DIFF
--- a/src/nspawn/nspawn.c
+++ b/src/nspawn/nspawn.c
@@ -3550,10 +3550,13 @@ static int inner_child(
                         /* If we cannot change the directory, we'll end up in /, that is expected. */
                         (void) chdir(home ?: "/root");
 
-                execle("/bin/bash", "-bash", NULL, env_use);
-                execle("/bin/sh", "-sh", NULL, env_use);
+                execle(DEFAULT_USER_SHELL, "-" DEFAULT_USER_SHELL_NAME, NULL, env_use);
+                if (!streq(DEFAULT_USER_SHELL, "/bin/bash"))
+                        execle("/bin/bash", "-bash", NULL, env_use);
+                if (!streq(DEFAULT_USER_SHELL, "/bin/sh"))
+                        execle("/bin/sh", "-sh", NULL, env_use);
 
-                exec_target = "/bin/bash, /bin/sh";
+                exec_target = DEFAULT_USER_SHELL ", /bin/bash, /bin/sh";
         }
 
         return log_error_errno(errno, "execv(%s) failed: %m", exec_target);
--- a/src/shared/user-record.c
+++ b/src/shared/user-record.c
@@ -1747,7 +1747,7 @@ const char *user_record_shell(UserRecord *h) {
                 return "/bin/sh";
 
         if (user_record_disposition(h) == USER_REGULAR)
-                return "/bin/bash";
+                return DEFAULT_USER_SHELL;
 
         return NOLOGIN;
 }
--- a/meson.build
+++ b/meson.build
@@ -760,6 +760,10 @@ conf.set('TIME_EPOCH', time_epoch)
 
 conf.set('CLOCK_VALID_RANGE_USEC_MAX', get_option('clock-valid-range-usec-max'))
 
+default_user_shell = get_option('default-user-shell')
+conf.set_quoted('DEFAULT_USER_SHELL',      default_user_shell)
+conf.set_quoted('DEFAULT_USER_SHELL_NAME', fs.name(default_user_shell))
+
 foreach tuple : [['system-alloc-uid-min', 'SYS_UID_MIN', 1],  # Also see login.defs(5).
                  ['system-uid-max',       'SYS_UID_MAX', 999],
                  ['system-alloc-gid-min', 'SYS_GID_MIN', 1],
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -220,6 +220,8 @@ option('time-epoch', type : 'integer', value : '-1',
        description : 'time epoch for time clients')
 option('clock-valid-range-usec-max', type : 'integer', value : '473364000000000', # 15 years
        description : 'maximum value in microseconds for the difference between RTC and epoch, exceeding which is considered an RTC error')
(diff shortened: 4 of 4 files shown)


src/shared/user-record.c
```diff
@@ -1747,7 +1747,7 @@ const char *user_record_shell(UserRecord *h) {
                 return "/bin/sh";
 
         if (user_record_disposition(h) == USER_REGULAR)
-                return "/bin/bash";
+                return access("/bin/bash", X_OK) ? "/bin/sh" : "/bin/bash";
```
bluca: This is called to answer D-Bus messages, so it could be running under a different filesystem, so a check like this might not be valid

poettering:
I am not convinced this is really desirable to be dynamic like this. 

I'd be fine making this a compile-time config option, i.e. in meson or so. On distros which really like to excise bash from their world, they could configure systemd to invoke some other shell then, or even call some path that would be a symlink to the actual shell to use, or so.

nabijaczleweli:
Agree, this is also the conclusion I came to, but hadn't had the time to write it.

nabijaczleweli:
Updated: we now allow `-Ddefault-user-shell=/bin/ksh93` and `user_record_shell()` will fall back to `/bin/ksh93` and nspawn interactive sessions `exec(/bin/ksh93, -ksh93, ...)`.

src/nspawn/nspawn.c
```diff
@@ -3550,10 +3550,10 @@ static int inner_child(
                         /* If we cannot change the directory, we'll end up in /, that is expected. */
                         (void) chdir(home ?: "/root");
 
-                execle("/bin/bash", "-bash", NULL, env_use);
+                execle(DEFAULT_USER_SHELL, "-" DEFAULT_USER_SHELL_NAME, NULL, env_use);
```
poettering, line 3553: uh, not convinced about this change here...

you are talking about a container here, i.e. a *different* system from where this systemd build is installed into. Thus, to remain compatible with 99% of the container images this really should invoke bash too. hence, it's OK if you try DEFAULT_USER_SHELL first, but please try bash next.

nabijaczleweli: Yeah, makes sense – updated to do DEFAULT_USER_SHELL, then bash (if DEFAULT_USER_SHELL != /bin/bash), then sh.

Could always try /bin/bash – it'd be less churny than the conditional code:
```c
execle(DEFAULT_USER_SHELL, "-" DEFAULT_USER_SHELL_NAME, NULL, env_use);
execle("/bin/bash", "-bash", NULL, env_use);
execle("/bin/sh", "-sh", NULL, env_use);

exec_target = DEFAULT_USER_SHELL ", /bin/bash, /bin/sh";
```
but in the default configuration we'd try to exec it twice before falling through to sh, so I opted for the less redundant option.

src/nspawn/nspawn.c
```diff
@@ -3550,10 +3550,16 @@ static int inner_child(
                         /* If we cannot change the directory, we'll end up in /, that is expected. */
                         (void) chdir(home ?: "/root");
 
+                execle(DEFAULT_USER_SHELL, "-" DEFAULT_USER_SHELL_NAME, NULL, env_use);
+#ifndef DEFAULT_USER_SHELL_IS_BINBASH
                 execle("/bin/bash", "-bash", NULL, env_use);
+#define BASH_INFIX ", /bin/bash"
+#else
+#define BASH_INFIX
+#endif
```
poettering: umpf, this is a bit too complex for my taste. both gcc and llvm optmize strcmp() of constant strings away, so I'd just do a simple, plain streq() here, i.e:

```c
execle(DEFAULT_USER_SHELL, "-" DEFAULT_USER_SHELL_NAME, NULL, env_use);
if (!streq(DEFAULT_USER_SHELL, "/bin/bash"))
        execle("/bin/bash", "-/bin/bash", NULL, env_use);
if (!streq(DEFAULT_USER_SHELL, "/bin/sh"))
        execle("/bin/sh", "-/bin/sh", NULL, env_use);
```

verify here: 
https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:2,endLineNumber:5,positionColumn:2,positionLineNumber:5,selectionStartColumn:2,selectionStartLineNumber:5,startColumn:2,startLineNumber:5),source:'%23include+%3Cstring.h%3E%0A%0Aint+main(int+argc,+char+*argv%5B%5D)+%7B%0A++++return+!!!!strcmp(%22x%22,+%22y%22)%3B%0A%7D'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cg112,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'',selection:(endColumn:12,endLineNumber:8,positionColumn:12,positionLineNumber:8,selectionStartColumn:12,selectionStartLineNumber:8,startColumn:12,startLineNumber:8),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+11.2+(C,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

nabijaczleweli: Yeah, that's a lot better; applied. My main concern with doing it the roundabout way was `exec_target`, but that's just a constant `DEFAULT_USER_SHELL ", /bin/bash, /bin/sh"` now, which I think is good enough (though I guess we could strlcat() it if this exceedingly rare cosmetic issue is that much of a problem?).

keszybz:
Looks reasonable.

PR #21852, opened 2021-12-21
test: set the TriggerLimitIntervalSec= unconditionally

mrc0mmand:
since it affects slower machines (or machines under a heavy load) as
well.

See: https://github.com/systemd/systemd/pull/21808#issuecomment-998927401

/cc @bluca

### DIFF
--- a/test/TEST-63-ISSUE-17433/test.sh
+++ b/test/TEST-63-ISSUE-17433/test.sh
@@ -9,12 +9,13 @@ TEST_DESCRIPTION="https://github.com/systemd/systemd/issues/17433"
 
 test_append_files() {
     (
-        # Collecting coverage slows this particular test quite a bit, causing
-        # it to fail with the default settings (20 triggers per 2 secs)
-        # to trip over the default limit. Let's help it a bit in such case.
-        if get_bool "$IS_BUILT_WITH_COVERAGE"; then
+        # Tweak the trigger limit interval in case we are collecting coverage
+        # or running without KVM - in both cases we might be slow enough that
+        # we could miss the default rate-limit window and cause the test to fail
+        # unexpectedly.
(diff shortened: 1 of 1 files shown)


test/TEST-63-ISSUE-17433/test.sh
```diff
@@ -9,13 +9,12 @@ TEST_DESCRIPTION="https://github.com/systemd/systemd/issues/17433"
...
-        if get_bool "$IS_BUILT_WITH_COVERAGE"; then
-            mkdir -p "${initdir:?}/etc/systemd/system/test63.path.d"
-            printf "[Path]\nTriggerLimitIntervalSec=10\n" >"${initdir:?}/etc/systemd/system/test63.path.d/coverage-override.conf"
-        fi
+        # Make the test work reliably on slower machines as well, where we might
+        # miss the default TriggerLimitIntervalSec= window and end up
+        # reactivating the service just slow enough to fly under the rate-limit
+        # radar
+        mkdir -p "${initdir:?}/etc/systemd/system/test63.path.d"
+        printf "[Path]\nTriggerLimitIntervalSec=10\n" >"${initdir:?}/etc/systemd/system/test63.path.d/coverage-override.conf"
```
bluca: just change the original unit given it's unconditional?

mrc0mmand: looks like today is not my day, thanks for spotting this

bluca (approved):

mrc0mmand:
I pushed a new version which tweaks the trigger limit interval for in case we're collecting coverage or running without KVM. That should, hopefully, address both the CI fails and @DaanDeMeyer's concerns in https://github.com/systemd/systemd/pull/21808#issuecomment-998997825.

PR #21607, opened 2021-12-02
ci: run build test with BPF-related stuff as well

mrc0mmand:

PR #21859, opened 2021-12-22
meson: make it compatible with AFL and honggfuzz again

evverx:
afl-clang and hufzz-clang try to instrument the code and the
underlying compilers don't like it. It should probably be
fixed in both afl and honggfuzz eventually but until then
let's just use "raw" clang to build bpf-skeletons.

It's a follow-up to https://github.com/systemd/systemd/pull/21607


mrc0mmand (approved):
Looks reasonable, thanks!

evverx:
Thanks for https://github.com/systemd/systemd/pull/21607! I didn't even realize the BPF part wasn't tested until I came across that PR :-)

PR #21863, opened 2021-12-22
analyze: do not connect to DBUS with --offline

bluca:
Co-authored-by: Lucas Werkmeister <mail@lucaswerkmeister.de>


bluca:
I'd like this in 250 as it goes with the new `systemd-analyze security --offline`, it will be useful in build containers without a running dbus session

yuwata (approved):

PR #21878, opened 2021-12-24
resolve: fix building without DoT with openssl

Hello71:
otherwise build fails with undefined reference to symbol 'EVP_sha1@@OPENSSL_3.0.0'. this has overlinking of basic_gcrypt but that's probably fine?


yuwata:
Thank you for the report. Could you test PR #21880?

PR #21880, opened 2021-12-24
meson: fix build with -Dcryptolib=openssl -Ddns-over-tls=false

yuwata:
Previously, when -Ddns-over-tls=false, libopenssl was missing in the
dependency of resolved.
Also, this drops libgpg_error when it is not necessary.

Replaces #21878.

PR #21888, opened 2021-12-24
ci: test build with supported cryptolibs to some degree

mrc0mmand:
Let's assign a specific -Dcryptolib= value to each job to have at least
some coverage for all supported cryptolibs without unnecessarily
multiplying the test matrix.

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -10,7 +10,7 @@ success() { echo >&2 -e "\033[32;1m$1\033[0m"; }
 ARGS=(
     "--optimization=0"
     "--optimization=s"
-    "--optimization=3 -Db_lto=true"
+    "--optimization=3 -Db_lto=true -Ddns-over-tls=false"
     "--optimization=3 -Db_lto=false"
     "--optimization=3 -Ddns-over-tls=openssl"
     "--optimization=3 -Dfexecve=true -Dstandalone-binaries=true -Dstatic-libsystemd=true -Dstatic-libudev=true"
@@ -62,6 +62,7 @@ PACKAGES=(
 COMPILER="${COMPILER:?}"
 COMPILER_VERSION="${COMPILER_VERSION:?}"
 LINKER="${LINKER:?}"
+CRYPTOLIB="${CRYPTOLIB:?}"
 RELEASE="$(lsb_release -cs)"
 
 bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $RELEASE main restricted universe multiverse' >>/etc/apt/sources.list"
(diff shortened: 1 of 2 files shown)


.github/workflows/build_test.sh
```diff
@@ -121,7 +121,7 @@ for args in "${ARGS[@]}"; do
          CC="$CC" CC_LD="$LINKER" CFLAGS="-Werror" \
          CXX="$CXX" CXX_LD="$LINKER" CXXFLAGS="-Werror" \
          meson -Dtests=unsafe -Dslow-tests=true -Dfuzz-tests=true --werror \
-               $args build; then
+               -Dcryptolib="${CRYPTOLIB:?}" $args build; then
```
medhefgo, line 125: The check whether this var is set should be at the top with the others ones (for early fail).

.github/workflows/build_test.yml
```diff
@@ -19,20 +19,20 @@ jobs:
...
+          - { COMPILER: "gcc",   COMPILER_VERSION: "11", LINKER: "gold", CRYPTOLIB: "openssl" }
+          - { COMPILER: "clang", COMPILER_VERSION: "11", LINKER: "bfd",  CRYPTOLIB: "auto"    }
+          - { COMPILER: "clang", COMPILER_VERSION: "12", LINKER: "gold", CRYPTOLIB: "gcrypt"  }
+          - { COMPILER: "clang", COMPILER_VERSION: "13", LINKER: "lld",  CRYPTOLIB: "openssl" }
     env: ${{ matrix.env }}
     steps:
       - name: Repository checkout
         uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
-      - name: Build check (${{ env.COMPILER }}-${{ env.COMPILER_VERSION }}-${{ env.LINKER }})
+      - name: Build check (${{ env.COMPILER }}-${{ env.COMPILER_VERSION }}-${{ env.LINKER }}-${{ env.CRYPTOLIB }})
```
medhefgo: This might be a bit clear now:
```suggestion
      - name: ${{ format('Build check ({0}-{1}-{2}-{3})', env.COMPILER, env.COMPILER_VERSION, env.LINKER, env.CRYPTOLIB) }}
```

.github/workflows/build_test.sh
mrc0mmand: Indeed, thanks.

.github/workflows/build_test.yml
mrc0mmand: I though about this as well, but it still looks pretty ugly. But I guess we could give it a try ;-)

mrc0mmand:
Looks like it reproduces #21880, so I'll wait until it's merged and then rebase this PR.

yuwata:
LGTM.

PR #21890, opened 2021-12-25
random-util: use ssize_t for getrandom return value

floppym:
This matches the prototype provided by glibc.

### DIFF
--- a/src/basic/missing_syscall.h
+++ b/src/basic/missing_syscall.h
@@ -78,7 +78,8 @@ static inline int missing_memfd_create(const char *name, unsigned int flags) {
 /* ======================================================================= */
 
 #if !HAVE_GETRANDOM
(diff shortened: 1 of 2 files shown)


src/basic/missing_syscall.h
```diff
@@ -78,7 +78,7 @@ static inline int missing_memfd_create(const char *name, unsigned int flags) {
 /* ======================================================================= */
 
 #if !HAVE_GETRANDOM
-static inline int missing_getrandom(void *buffer, size_t count, unsigned flags) {
+static inline ssize_t missing_getrandom(void *buffer, size_t count, unsigned flags) {
```
yuwata, line 82: Maybe, it is better to add some comments?

floppym: I'm not certain what kind of comment you are looking for, but I added a note about matching glibc's definition.

PR #21893, opened 2021-12-25
meson: don't try to guess versioned clang/llvm-strip bins for cross compile

jameshilliard:
This should simplify overriding the program locations as the binary names should now not change if cross compiling.

It's likely any attempts at autodetecting these in cross environments will be brittle at best so lets just disable it.


yuwata:
LGTM.

PR #21894, opened 2021-12-25
hwdb: 60-keyboard: Fix volume-button mapping on Asus TF103C

jwrdegoede:
The Asus TF103C misses the home button in its PNP0C40 GPIO resources
causing the button mappings for the volume buttons to be off by one,
leading to the volume-up button sending home button presses and the
volume-down button sending volume-up button presses.

Add a 60-keyboard hwdb entry to correct the mappings. Note this is
split over 2 input devices because the soc_button_array driver
creates separate input devices for power + home and vol up/down.
This is done because power/home act as wakeup buttons where as
the volume buttons do not.

This means that after this fixup the home -> volume-up button
still acts as a wakeup button, there is nothing which can be done
about this without adding a kludge to the kernel which is not
worth the trouble (IMHO).


yuwata:
LGTM.

PR #21733, opened 2021-12-11
Disable exporting D-Bus Introspection XML if cross-compiling

igo95862:
Fixes #21730

This is a soft disable. Passing `dbus-interfaces-dir` build option explicitly
will enable exports again even when cross compiling. (maybe your
environment will allow to execute cross compiled binaries)

PR #21895, opened 2021-12-25
meson: attempt to get dbus interfaces_dir from pkg-config

jameshilliard:
Using pkg-config variables should work more reliably and be compatible with cross compilation by default.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1466,12 +1466,10 @@ if want_dbus != 'false' and not skip_deps
         libdbus = dependency('dbus-1',
                              version : '>= 1.3.2',
                              required : want_dbus == 'true')
-        have = libdbus.found()
 else
-        have = false
-        libdbus = []
+        libdbus = disabler()
 endif
-conf.set10('HAVE_DBUS', have)
+conf.set10('HAVE_DBUS', libdbus.found())
 
 # We support one or the other. If gcrypt is available, we assume it's there to
 # be used, and use it in preference.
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -1732,15 +1730,20 @@ public_programs = []
 # D-Bus introspection XML export
 dbus_programs = []
 dbus_interfaces_dir = get_option('dbus-interfaces-dir')
-if dbus_interfaces_dir == ''
-        if not meson.is_cross_build()
-                dbus_interfaces_dir = datadir / 'dbus-1'
+if dbus_interfaces_dir == '' or dbus_interfaces_dir == 'yes'
+        if meson.is_cross_build() and dbus_interfaces_dir != 'yes'
+                dbus_interfaces_def =  'no'
```
yuwata: doubled white space.

meson.build
```diff
@@ -1732,15 +1730,20 @@ public_programs = []
...
         else
+                dbus_interfaces_def =  datadir / 'dbus-1'
```
yuwata: Also here. White space is doubled.

yuwata (commented):
It seems slightly complicated for me.
- cross compliling -> 'no'
- dbus enabled -> use pkg-config
- dbus disabled -> `datadir / 'dbus-1'`

Is this the logic you intended?

jameshilliard:
> Is this the logic you intended?

Sorta

- cross compiling( with`-Ddbus-interfaces-dir` option not set or set to empty string "") -> dbus_interfaces_dir=`no`
- if `dbus_interfaces_dir` != `no` -> try getting variable from pkg-config -> fallback to `datadir / 'dbus-1'` on failure

So we only use the fallback default `datadir / 'dbus-1'` if pkg-config fails this way.

meson.build
jameshilliard: fixed

meson.build
jameshilliard: fixed

yuwata:
@jameshilliard I've tried to simplify your patch in #21898. Could you take a look?

jameshilliard:
> I've tried to simplify your patch in #21898.

Getting those additional pkg-config variables makes sense, but I think some of the simplifications removed the behavior introduced in #21733 and may be disable more tests than intended.

bluca:
Replaced by https://github.com/systemd/systemd/pull/21898

PR #11294, opened 2018-12-30
meson: use disabler feature to convert configuration errors to warnings

jameshilliard:
There are a number of configure checks which emit errors that really should only emit warnings. Hard errors should be only used when nothing in the project can be built without the dependency or when that dependency is required for an option that was explicitly requested to be enabled.

This is mostly a follow up to #11141.

PR #21898, opened 2021-12-26
meson: obtain dbus related directories from pkg-config

yuwata:
Replaces #21895.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -191,21 +191,6 @@ if docdir == ''
         docdir = datadir / 'doc/systemd'
 endif
 
-dbuspolicydir = get_option('dbuspolicydir')
-if dbuspolicydir == ''
-        dbuspolicydir = datadir / 'dbus-1/system.d'
-endif
-
-dbussessionservicedir = get_option('dbussessionservicedir')
-if dbussessionservicedir == ''
-        dbussessionservicedir = datadir / 'dbus-1/services'
-endif
-
-dbussystemservicedir = get_option('dbussystemservicedir')
-if dbussystemservicedir == ''
-        dbussystemservicedir = datadir / 'dbus-1/system-services'
-endif
-
 pamlibdir = get_option('pamlibdir')
 if pamlibdir == ''
         pamlibdir = rootlibdir / 'security'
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -1466,12 +1451,43 @@ if want_dbus != 'false' and not skip_deps
...
+        if libdbus.found()
+                dbussessionservicedir = libdbus.get_variable(pkgconfig: 'session_bus_services_dir', default_value: dbussessionservicedir)
+        endif
+endif
+
+dbussystemservicedir = get_option('dbussystemservicedir')
+if dbussystemservicedir == ''
+        dbussystemservicedir = datadir / 'dbus-1/system-services'
+        if libdbus.found()
+                dbussystemservicedir = libdbus.get_variable(pkgconfig: 'system_bus_services_dir=', default_value: dbussystemservicedir)
```
jameshilliard: I think should just be `system_bus_services_dir` not `system_bus_services_dir=`.

src/libsystemd/meson.build
```diff
@@ -232,7 +232,8 @@ tests += [
           libglib,
           libgobject,
           libgio,
-          libdbus]],
+          libdbus],
```
jameshilliard: I think you want to use a explicit `if libdbus.found()` check for this code like in #21895 to ensure the disabler doesn't propagate too far and disable all the tests here.

meson.build
```diff
@@ -1731,17 +1747,6 @@ public_programs = []
 
 # D-Bus introspection XML export
 dbus_programs = []
-dbus_interfaces_dir = get_option('dbus-interfaces-dir')
-if dbus_interfaces_dir == ''
-        if not meson.is_cross_build()
-                dbus_interfaces_dir = datadir / 'dbus-1'
-        else
-                message('D-Bus interfaces export is disabled during cross build. Pass path or yes to force enable.')
```
jameshilliard, line 1739: I think we still need to disable this by default for cross builds since it requires execution of native binaries.

meson.build
yuwata: Ouch!

meson.build
yuwata: I see.

src/libsystemd/meson.build
yuwata: If libdbus is not found, then HAVE_DBUS is 0. And we usually control enable or disable by the condition HAVE_XXX.
Or you mean this doesn't work?

jameshilliard: > If libdbus is not found, then HAVE_DBUS is 0. And we usually control enable or disable by the condition HAVE_XXX.

Different issue, [disablers](https://mesonbuild.com/Disabler.html) depending on how they are adding to lists will generally disable the entire list and anything that depends on that list I think. By making the list addition for `libdbus` conditional on `libdbus.found()` which is a boolean we can guard against unintended behavior.

yuwata: > > If libdbus is not found, then HAVE_DBUS is 0. And we usually control enable or disable by the condition HAVE_XXX.
> 
> Different issue, [disablers](https://mesonbuild.com/Disabler.html) depending on how they are adding to lists will generally disable the entire list and anything that depends on that list I think. By making the list addition for `libdbus` conditional on `libdbus.found()` which is a boolean we can guard against unintended behavior.

Sorry, I cannot follow... What problem did you see? I tried to disable libdbus by `meson configure -Ddbus=false`, then test-bus-marshal is not built, and all other build succeeds.

I noticed that setting `HAVE_DBUS` in the condition is even not necessary, as now libdbus is a dependency or disabler object. So, in the current version, nothing changed for the test-bus-marshal.

yuwata:
@jameshilliard Updated. Now, the logic is simpler, but I hope still the feature you want to keep is kept. PTAL.

src/libsystemd/meson.build
yuwata: Hm, I noticed that test-bus-marshal can be built without libdbus. So, even if `-Ddbus=false`, the test should be built. Using disabler is now dropped.

See also @keszybz's comment in #11294.

meson.build
```diff
@@ -1473,6 +1458,59 @@ else
...
+        if conf.get('HAVE_DBUS') == 1
+                dbussystemservicedir = libdbus.get_variable(pkgconfig: 'system_bus_services_dir', default_value: dbussystemservicedir)
+        endif
+endif
+
+dbus_interfaces_dir = get_option('dbus-interfaces-dir')
+if dbus_interfaces_dir == '' or dbus_interfaces_dir == 'yes'
+        if meson.is_cross_build() and dbus_interfaces_dir != 'yes'
+                dbus_interfaces_dir = 'no'
+                message('Exporting D-Bus interface XML files is disabled during cross build. Pass path or "yes" to force enable.')
```
medhefgo: This should be warning imho

meson.build
```diff
@@ -1473,6 +1458,59 @@ else
...
+        elements = dbus_interfaces_dir.split('/')
+        dbus_interfaces_dir_name = elements[-1]
+        dbus_interfaces_dir_parent = ''
+        foreach elem : elements
+                if elem == dbus_interfaces_dir_name
+                        break
+                endif
+                dbus_interfaces_dir_parent = dbus_interfaces_dir_parent / elem
+        endforeach
+endif
```
jameshilliard, line 1512: What's this if else block for?

meson.build
yuwata: Done.

meson.build
yuwata: This splits the dirname and basename.
If the interfaces directory is not located at the default place, we need to manually split it.

jameshilliard:
> Updated. Now, the logic is simpler, and I hope still the feature you want to keep is kept.

Yeah, I think this looks like it should work.

bluca (approved):

PR #19598, opened 2021-05-13
PoC: add Rust submodule as libbasic_rust

bluca:
Proof-of-concept to test integrating Rust submodules.

This adds a (very naive) rewrite of string_extract_line from string-util.h
in Rust, and compiles it in so that the rest of our C code, including the
unit tests, call into it instead.

Given we are only using the standard library, there is no need for a
(complicated) integration with Cargo, and the only additional
requirement is the Rust compiler. Meson abstracts it quite nicely.

Note that this is the first time I've written Rust - so it's most likely horrible and completely wrong. Learning was the main idea of the exercise :-)
Opening a PR as I think there were questions in the past about how easily it would be to add Rust modules in the code base.

PR #16136, opened 2020-06-10
Move the gcc/clang build check to GitHub Actions

mrc0mmand:

PR #21902, opened 2021-12-26
ci: replace apt-key with signed-by

evverx:
to limit the scope of the key to apt.llvm.org only.

This is mostly inspired by https://blog.cloudflare.com/dont-use-apt-key/

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -80,9 +80,10 @@ if [[ "$COMPILER" == clang ]]; then
     # llvm package if available in such cases to avoid that.
     if ! apt show --quiet "llvm-$COMPILER_VERSION" &>/dev/null; then
         # Latest LLVM stack deb packages provided by https://apt.llvm.org/
-        # Following snippet was borrowed from https://apt.llvm.org/llvm.sh
-        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-        add-apt-repository -y "deb http://apt.llvm.org/$RELEASE/   llvm-toolchain-$RELEASE-$COMPILER_VERSION  main"
+        # Following snippet was partly borrowed from https://apt.llvm.org/llvm.sh
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --yes --dearmor --output /usr/share/keyrings/apt-llvm-org.gpg
+        printf "deb [signed-by=/usr/share/keyrings/apt-llvm-org.gpg] http://apt.llvm.org/%s/   llvm-toolchain-%s-%s  main\n" \
+		"$RELEASE" "$RELEASE" "$COMPILER_VERSION" >/etc/apt/sources.list.d/llvm-toolchain.list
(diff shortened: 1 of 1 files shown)


.github/workflows/build_test.sh
```diff
@@ -80,9 +80,10 @@ if [[ "$COMPILER" == clang ]]; then
     # llvm package if available in such cases to avoid that.
     if ! apt show --quiet "llvm-$COMPILER_VERSION" &>/dev/null; then
         # Latest LLVM stack deb packages provided by https://apt.llvm.org/
-        # Following snippet was borrowed from https://apt.llvm.org/llvm.sh
-        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-        add-apt-repository -y "deb http://apt.llvm.org/$RELEASE/   llvm-toolchain-$RELEASE-$COMPILER_VERSION  main"
+        # Following snippet was partly borrowed from https://apt.llvm.org/llvm.sh
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --yes --dearmor --output /usr/share/keyrings/apt-llvm-org.gpg
+        printf "deb [signed-by=/usr/share/keyrings/apt-llvm-org.gpg] http://apt.llvm.org/%s/   llvm-toolchain-%s-%s  main\n" \
```
medhefgo, line 85: If I read the manpage correctly, you can still use add-apt-repository for this.

medhefgo:
I honestly don't see the point of this. CI instanced are too short-lived to bother with this.

evverx:
The idea is mostly to avoid promoting patterns like `curl | bash` and so on that keeps popping up in PRs (where rust, for example, was introduced :-)) But ideally the CI should be ready to sign and release systemd tarballs and using stuff like that shouldn't hurt I think.

> I honestly don't see the point of this

I'm not sure where reviews like that were when several multi-write-heap-buffer-overflows were introduced in `systemd-analyze` :-)

medhefgo:
We're not doing `curl | bash` here, though. Rather than changing patterns to how to properly add an apt repo everywhere, the upstream tools themselves should be changed to support narrowing down the key scope.

evverx:
> We're not doing curl | bash here

That's probably because it is usually caught during review: https://github.com/systemd/systemd/pull/16136#discussion_r438703539 and https://github.com/systemd/systemd/pull/19598#discussion_r635523495

> Rather than changing patterns to how to properly add an apt repo everywhere, the upstream tools themselves should be changed to support narrowing down the key scope

I agree to some extent but I think it will take a few years at best. Until then I don't think there is anything wrong with using kludges to get it around.

.github/workflows/build_test.sh
evverx: I read the code and as far as I can tell `add-apt-repository` calls `apt-key` under the hood sometimes. Other than that removing it saves at least one `apt-get update` that shouldn't be run because the script runs it after it's ready to deal with all the repositories and packages.

mrc0mmand (approved):
The relevant CIs appear to be happy, so LGTM.

The three CI fails are unrelated (and already reported as issues).

Issue #21901, opened 2021-12-26
Cannot set file attributes for '/var/log/journal', maybe due to incompatibility in specified attributes


mbiebl:
Version: 250

Since upgrading from 249.7 to 250, I get the following warning from systemd-tmpfiles:
```
# systemd-tmpfiles --create --prefix /var/log/journal
Cannot set file attributes for '/var/log/journal', maybe due to incompatibility in specified attributes, previous=0x00080000, current=0x00080000, expected=0x00880000, ignoring.
Cannot set file attributes for '/var/log/journal/567a68a5c2672114bcf5192d00000008', maybe due to incompatibility in specified attributes, previous=0x00080000, current=0x00080000, expected=0x00880000, ignoring.
```
The file system in question is a bog standard ext4, but I also noticed it with XFS.

bluca:
easily reproduced with mkosi build/boot

bluca:
```
root@image:~# SYSTEMD_LOG_LEVEL=debug systemd-tmpfiles --create --prefix /var/log/journal
Looking for configuration files in (higher priority first):
	/etc/tmpfiles.d
	/run/tmpfiles.d
	/usr/local/lib/tmpfiles.d
	/usr/lib/tmpfiles.d
SELinux enabled state cached to: disabled
Reading config file "/usr/lib/tmpfiles.d/dbus.conf"…
Entry "/var/lib/dbus" does not match any include prefix, skipping.
Entry "/var/lib/dbus/machine-id" does not match any include prefix, skipping.
Reading config file "/usr/lib/tmpfiles.d/debian.conf"…
Entry "/run/shm" does not match any include prefix, skipping.
Entry "/run/sendsigs.omit.d" does not match any include prefix, skipping.
Entry "/etc/mtab" does not match any include prefix, skipping.
Reading config file "/usr/lib/tmpfiles.d/etc.conf"…
Entry "/etc/os-release" does not match any include prefix, skipping.
Entry "/etc/mtab" does not match any include prefix, skipping.
/usr/lib/tmpfiles.d/etc.conf:12: Ignoring entry C! "/etc/locale.conf" because --boot is not specified.
/usr/lib/tmpfiles.d/etc.conf:13: Ignoring entry C! "/etc/nsswitch.conf" because --boot is not specified.
/usr/lib/tmpfiles.d/etc.conf:14: Ignoring entry C! "/etc/pam.d" because --boot is not specified.
[...]

PR #21906, opened 2021-12-26
chrattr-util: return EOPNOTSUPP from chrattr_full if no other failure was observed

bluca:
When chattr_full tries to apply flags one-by-one, and one fails,
record which errno was returned. But record EOPNOTSUPP(&friends)
only if no other error is observed, and return it only in that case
(otherwise keep returning ENOANO), so that callers can respond
appropriately to EOPNOTSUPP vs more relevant errors.
For example, this lets tmpfiles.d log at debug level when a filesystem
flag cannot be applied because the filesystem does not support it,
but at warning level if something else went wrong when applying it.
Restores logging behaviour of tmpfiles.d to pre-250.

Follow-up for: https://github.com/systemd/systemd/commit/c1631ee124a30abfb9c71e2a1534b8afffc3b6a7

Fixes: https://github.com/systemd/systemd/issues/21901

### DIFF
--- a/src/basic/chattr-util.c
+++ b/src/basic/chattr-util.c
@@ -22,6 +22,7 @@ int chattr_full(const char *path,
 
         _cleanup_close_ int fd_will_close = -1;
         unsigned old_attr, new_attr;
+        int set_flags_errno = 0;
         struct stat st;
 
         assert(path || fd >= 0);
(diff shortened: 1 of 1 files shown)


yuwata:
LGTM.

yuwata:
Needs to be backported to v250-stable.

PR #21908, opened 2021-12-27
man: clarify Environmentfile format

yonran:
Remove incorrect claim that C escapes (such as `\t` and `\n`) are recognized and that control characters are disallowed. Specify the allowed characters and escapes with single quotes, with double quotes, and without quotes.

I am changing this after being misled by the documentation. I also added a [stackoverflow answer](https://stackoverflow.com/a/70491161/471341) to respond to someone else who was misled by the documentation.

### DIFF
--- a/src/test/test-env-file.c
+++ b/src/test/test-env-file.c
@@ -13,11 +13,11 @@
         "a=a\n"                                 \
         "b=b\\\n"                               \
         "c\n"                                   \
-        "d=d\\\n"                               \
-        "e\\\n"                                 \
-        "f\n"                                   \
+        "d= d\\\n"                              \
+        "e  \\\n"                               \
+        "f  \n"                                 \
         "g=g\\ \n"                              \
-        "h=h\n"                                 \
+        "h= ąęół\\ śćńźżµ \n"                   \
         "i=i\\"
 
 #define env_file_2                              \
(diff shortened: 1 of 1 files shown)


keszybz:
The description is written very well. The only question is whether is actually matches what the code does. It seems that we don't have any unit tests for this…

keszybz:
@yonran can you add some unittests for the EnvironmentFile parsing code? I'd be happy to merge the PR once we have that. If not, I or someone else will add them, but it'll take some more time.

keszybz:
I added some tests and a small refactoring. The new docs seem to match behaviour, but please review the new code.

src/core/execute.c
```diff
@@ -5370,20 +5370,18 @@ static int exec_context_named_iofds(
         return targets == 0 ? 0 : -ENOENT;
 }
 
-static int exec_context_load_environment(const Unit *unit, const ExecContext *c, char ***l) {
-        char **i, **r = NULL;
+static int exec_context_load_environment(const Unit *unit, const ExecContext *c, char ***ret) {
+        _cleanup_strv_free_ char **r = NULL;
```
yuwata: I'd like to rename this variable to e.g. `v` or something, and move and replace `int k;` with `int r;` at the most outer scope.

yuwata (approved):
LGTM. One minor coding style suggestion.

Issue #21911, opened 2021-12-27
User session manager access on `/run/systemd/systemd-units-load`


cgzones:
**systemd version the issue has been seen with**
 > 250-1

**Used distribution**
 > Debian experimental

**Linux kernel version used** (`uname -a`)
 > `Linux debianBullseye 5.15.0-2-amd64 #1 SMP Debian 5.15.5-2 (2021-12-18) x86_64 GNU/Linux`

**Unexpected behaviour you saw**
 > The user session manager tries to access the system wide file `/run/systemd/systemd-units-load`

**Steps to reproduce the problem**
 > (re-)start the user session manager

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
type=PROCTITLE msg=audit(12/27/21 16:17:57.167:85) : proctitle=(systemd) 
type=PATH msg=audit(12/27/21 16:17:57.167:85) : item=0 name= inode=235 dev=00:15 mode=file,444 ouid=root ogid=root rdev=00:00 obj=system_u:object_r:systemd_runtime_t:s0 nametype=NORMAL cap_fp=none cap_fi=none cap_fe=0 cap_fver=0 cap_frootid=0 
type=CWD msg=audit(12/27/21 16:17:57.167:85) : cwd=/ 
type=SYSCALL msg=audit(12/27/21 16:17:57.167:85) : arch=x86_64 syscall=newfstatat success=no exit=EACCES(Permission denied) a0=0x16 a1=0x7f2a03f5a75a a2=0x7ffc8ee84f50 a3=0x1000 items=1 ppid=1 pid=1539 auid=root uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=(none) ses=3 comm=systemd exe=/usr/lib/systemd/systemd subj=root:sysadm_r:sysadm_systemd_t:s0-s0:c0.c1023 key=(null) 
type=AVC msg=audit(12/27/21 16:17:57.167:85) : avc:  denied  { getattr } for  pid=1539 comm=systemd path=/run/systemd/systemd-units-load dev="tmpfs" ino=235 scontext=root:sysadm_r:sysadm_systemd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:systemd_runtime_t:s0 tclass=file permissive=0 
----
type=PROCTITLE msg=audit(12/27/21 16:17:57.167:86) : proctitle=(systemd) 
[...]

PR #21913, opened 2021-12-27
core: do not touch /run/systemd/systemd-units-load from user session instances

bluca:
Follow-up for: https://github.com/systemd/systemd/commit/15b9243c0d7f6d1531fa65dbc01bd11e8e6c12ca
Fixes: https://github.com/systemd/systemd/issues/21911


yuwata (approved):
Ouch. LGTM.

PR #21727, opened 2021-12-10
ci: Build test with different linkers

medhefgo:

PR #21915, opened 2021-12-27
tests: add fuzz-bcd

evverx:


evverx:
I'm not sure why CIFuzz thinks that
```==19==ERROR: AddressSanitizer: stack-use-after-return on address 0x7f440dccc5fa at pc 0x0000004da668 bp 0x7ffd3a6ae2a0 sp 0x7ffd3a6ae298
READ of size 1 at 0x7f440dccc5fa thread T0
SCARINESS: 50 (1-byte-read-stack-use-after-return)
    #0 0x4da667 in get_key /work/build/../../src/systemd/src/boot/efi/bcd.c:160:77
    #1 0x4da667 in get_subkey /work/build/../../src/systemd/src/boot/efi/bcd.c:134:34
    #2 0x4d972d in get_bcd_title /work/build/../../src/systemd/src/boot/efi/bcd.c:294:34
```
isn't reproducible though. It seems new fuzz targets somehow confuse it.

src/boot/efi/fuzz-bcd.c
```diff
@@ -0,0 +1,26 @@
...
+        /* This limit was borrowed from src/boot/efi/boot.c
+         * I'm not sure where it came from initially though
+         */
+        if (size > 100*1024)
+                return 0;
+
+        _cleanup_free_ void *p = memdup(data, size);
+        assert_se(p);
+
+        char16_t *title = get_bcd_title(p, size);
```
evverx, line 23: `get_bcd_title` isn't UBSan-friendly because it triggers a lot of "unaligned access" reports. I'm not sure what to do about it. If it's supposed to be run on x86 platforms only probably it shouldn't be even be built on non-x86 architectures: https://github.com/systemd/systemd/pull/21727#issuecomment-1001624144

evverx: It was fixed in https://github.com/systemd/systemd/pull/21922

evverx:
I don't think it's a draft PR anymore. It kind of conflicts with https://github.com/systemd/systemd/pull/21928 though but all in all I think it should be ready. @medhefgo could you take a look?

src/boot/efi/fuzz-bcd.c
```diff
@@ -0,0 +1,37 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "alloc-util.h"
+#include "fd-util.h"
+#include "fuzz.h"
+#include "utf8.h"
+
+#include "bcd.c"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
```
medhefgo, line 10: You're overcomplicating this way too much. I've been running witht his:
```c

#include "alloc-util.h"
#include "fuzz.h"
#include "log.h"
#include "utf8.h"

#include "bcd.c"

int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
        _cleanup_free_ uint8_t *copy = memdup(data, size);
        assert_se(copy);

        char16_t *title = get_bcd_title(copy, size);
        assert_se(!title || char16_strlen(title) >= 0);

        return 0;
}
```

tools/oss-fuzz.sh
```diff
@@ -56,6 +64,17 @@ fi
 
 ninja -v -C "$build" fuzzers
 
+# Compressed BCD files are kept in test/test-bcd so let's unpack them
+# and populate the directory the script uses to generate seed corpora
+rm -rf test/fuzz/fuzz-bcd/
+mkdir -p test/fuzz/fuzz-bcd/
```
medhefgo: You should never touch the source tree like this. The unpacking should be done in a temp folder.

tools/oss-fuzz.sh
```diff
@@ -56,6 +64,17 @@ fi
...
+
+# shellcheck disable=SC2016
+# Single quotes are used here intentionally to let sh expand $
+# when it's run by xargs
+find "test/test-bcd" -type f -name '*.zst' -print0 |
+xargs -0 -I%%FILE%% sh -x -c 'unzstd %%FILE%% -o "test/fuzz/fuzz-bcd/$(basename %%FILE%% .zst)"'
```
medhefgo: Something like this is much nicer:
```shell
bcd=$(mktemp -d)
for i in test/test-bcd/*.zst; do
     unzstd "$i" -o "$bcd/$(basename "${i%.zst}")";
done
zip -jqr "$OUT/fuzz-bcd_seed_corpus.zip" "$bcd"
rm -rf "$bcd"
```

evverx: I don't think `for` can handle directories that public OSS-Fuzz corpora bring with them so I don't think `find` should be replaced with it. Other than that, I agree the script should put those files in a temporary directory.

Ideally, the bcd parser should have been integrated properly into the test suite though to avoid kludges like that. In its current form it isn't compatible with the regression test suite.

src/boot/efi/fuzz-bcd.c
evverx: If it's safe to assume that the parser will never log anything or trigger OOMs by reading data without any limits I agree it can be trimmed significantly. But I generally don't try to make any assumptions about code :-)

evverx: I obviously agree that the if expression where `/dev/null` is opened can be safely replaced with
```
assert_se(!title || char16_strlen(title) >= 0);
```
though :-)

tools/oss-fuzz.sh
medhefgo: > I don't think `for` can handle directories that public OSS-Fuzz corpora bring with them so I don't think `find` should be replaced with it.

What exactly makes you think this for is unsafe?

src/boot/efi/fuzz-bcd.c
medhefgo: The bcd parser neither logs nor allocates memory.

tools/oss-fuzz.sh
evverx: I think initially the idea was to keep those files there to make it easier to test it locally. Instead of unzipping that seed corpus into a directory and passing it to the fuzzer I was possible to simply run `./out/fuzz-bcd test/fuzz/fuzz-bcd/` and be done with it. Then again I agree it isn't perfect.

tools/oss-fuzz.sh
medhefgo: > Ideally, the bcd parser should have been integrated properly into the test suite though to avoid kludges like that. In its current form it isn't compatible with the regression test suite.

What exactly makes it incompatible?

tools/oss-fuzz.sh
medhefgo: The fuzzing infra should really be taught to (optionally) decompress input if needed if you ask me.

tools/oss-fuzz.sh
evverx: > What exactly makes you think this for is unsafe?

I'm not saying it's unsafe. I'm saying that if directories are added to `test/test-bcd` that for loop would just skip them because "test/test-bcd/*.zst" can't expand, for example, "test/test-bcd/regressions/crash-...."

> What exactly makes it incompatible?

The regression testsuite expects files to be kept in `test/fuzz/fuzz-*`

tools/oss-fuzz.sh
evverx: I'm not sure how libFuzzer can tell compressed seed corpora from zip archives that are supposed to be passed to fuzz targets fuzzing zip files apart though.

src/boot/efi/fuzz-bcd.c
evverx: If it receives 4Gb files can it always finish in less than 10 seconds?

> The bcd parser neither logs nor allocates memory.

Currently it doesn't. I have no idea what is going to happen to it in 2 years though.

medhefgo: If you really wanna restrict the size, shouldn't you be telling the fuzzer directly like with fuzz-env-file?

tools/oss-fuzz.sh
medhefgo: > I'm not saying it's unsafe. I'm saying that if directories are added to test/test-bcd that for loop would just skip them because "test/test-bcd/*.zst" can't expand, for example, "test/test-bcd/regressions/crash-...."

bash has `shopt -s globstar` for that.

src/boot/efi/fuzz-bcd.c
evverx: That ".options" file is compatible with libFuzzer only. It has been kind of broken since AFL and honggfuzz started to build systemd on OSS-Fuzz. And as far as I can see to address that `fuzz-env-file` limits its input explicitly by calling
```
        if (size == 0 || size > 65535)
                return 0;
```

tools/oss-fuzz.sh
evverx: I think the problem is that most `for` bash loops in the systemd codebase (and I assume everywhere else) don't expect it to be enabled by default and I certainly wouldn't want to start rewriting those scripts to just replace `find` with `for`.

medhefgo: You don't have to change anything. no shell file has `**` in it. Even if they had, you don't need to change unrelated scripts.

evverx: I'm still not sure why
```
shopt -s globstar
for i in test/test-bcd/**.zst; do
     unzstd "$i" -o "$bcd/$(basename "${i%.zst}")";
done
```
is better than
```
find "test/test-bcd" -type f -name '*.zst' -print0 |
xargs -0 -I%%FILE%% sh -x -c 'unzstd %%FILE%% -o "'"$bcd"'/$(basename %%FILE%% .zst)"'
```
though :-) Since it's maintained by @mrc0mmand mostly @mrc0mmand I wonder what you'd prefer here?

medhefgo: It's more readable (less quote-nesting) and doesn't involve xargs. Imho, shell (bash) features should always be preferred to some crazy cmd pipeline.

Btw, the code as-is doesn't even work if subdirs were involved. Or rather it flattens the subdir structure, making the whole thing moot to begin with (and risk overriding files, which could hide real issues). I don't think subdirs are likely to be ever needed (for bcd tests).

evverx: > the code as-is doesn't even work if subdirs were involved

It doesn't work in general. But files kept in those "regression" directories are named according to their hash sums and collisions there are negligible.

Anyway, since it's mostly my use case I can keep it locally so I'll replace find with the for loop without `shopt` though because `test/test-bcd` is as flat as it can get.

evverx: FWIW @medhefgo if you're interested I can add your email address to https://github.com/google/oss-fuzz/blob/master/projects/systemd/project.yaml. It should be a gmail address though due to https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication

evverx: find was replaced with for

tools/oss-fuzz.sh
evverx: `mktemp` is now used

evverx: There was a typo. `mktemp` is now used

src/boot/efi/fuzz-bcd.c
evverx: I replaced that if expression with `assert_se(!title || char16_strlen(title) >= 0);` and decided to keep the other code mostly because I don't want to receive bogus notifications from OSS-Fuzz about pointless OOMs, timeouts and so on in the foreseeable future.

tools/oss-fuzz.sh
mrc0mmand: > I'm still not sure why
> 
> ```
> shopt -s globstar
> for i in test/test-bcd/**.zst; do
>      unzstd "$i" -o "$bcd/$(basename "${i%.zst}")";
> done
> ```
> 
> is better than
> 
> ```
> find "test/test-bcd" -type f -name '*.zst' -print0 |
> xargs -0 -I%%FILE%% sh -x -c 'unzstd %%FILE%% -o "'"$bcd"'/$(basename %%FILE%% .zst)"'
> ```
> 
> though :-) Since it's maintained by @mrc0mmand mostly @mrc0mmand I wonder what you'd prefer here?

I can throw in a third variant to make things more complicated ;)

```bash
while read -r file; do
    unzstd "$file" -o "$bcd/$(basename "${file%.zst}")"
done < <(find "test/test-bcd" -type f -name '*.zst') 
```

This seems to be somewhere in the middle of both proposed variants. So far I kept avoiding `shopt`s, since they are, let's say, not that common, so it might make the shell code less understandable (at least on the first glance), but in the end I don't really care which variant eventually "wins".

medhefgo (approved):

PR #21923, opened 2021-12-28
network: ndisc: ignore route prefix if the destination is ::/0 and UseGateway=no

### LINKED ISSUE #21912: UseGateway=no does not work on IPv6AcceptRA
[root@localhost ~]# systemctl --version
systemd 250 (250-1-arch)

**Used distribution**
Arch Linux

[root@localhost ~]# uname -a
Linux localhost 5.15.11-arch2-1 #1 SMP PREEMPT Wed, 22 Dec 2021 09:23:54 +0000 x86_64 GNU/Linux

x86-64

When setting UseGateway=no on IPv6AcceptRA, default route advertised by IPv6 RA should be ignored.

Default route is still inserted with UseGateway=no on IPv6AcceptRA section.

Set UseGateway=no on IPv6 network with default route advertised by RA.
Default route should not be inserted for the specified interface.

```text
[root@localhost ~]# cat /etc/systemd/network/ens160.network
[Match]
Name=ens160

[Link]
MTUBytes=1500

[Network]
DHCP=yes
IPv6LinkLocalAddressGenerationMode=eui64

[DHCPv4]
UseGateway=yes

[root@localhost ~]# cat /etc/systemd/network/ens192.network
[Match]
Name=ens192

[Link]
MTUBytes=9000

[Network]
DHCP=yes
IPv6LinkLocalAddressGenerationMode=eui64

[DHCPv4]
UseGateway=no

[IPv6AcceptRA]
UseGateway=no

[root@localhost ~]# ip -6 r
::1 dev lo proto kernel metric 256 pref medium
2001:aaaa:aaaa:aaaa::/64 dev ens160 proto ra metric 1024 expires 2591944sec pref medium
2a01:bbbb:bbbb:bbbb::/64 dev ens192 proto ra metric 1024 expires 86388sec pref medium
fe80::/64 dev ens160 proto kernel metric 256 pref medium
fe80::/64 dev ens192 proto kernel metric 256 pref medium
default proto ra metric 1024 expires 1744sec mtu 1500 pref medium
        nexthop via fe80::aaaa dev ens160 weight 1
        nexthop via fe80::bbbb dev ens192 weight 1
```

yuwata:
Fixes #21912.

### DIFF
--- a/src/network/networkd-ndisc.c
+++ b/src/network/networkd-ndisc.c
@@ -612,6 +612,11 @@ static int ndisc_router_process_route(Link *link, sd_ndisc_router *rt) {
         if (r < 0)
                 return log_link_error_errno(link, r, "Failed to get route prefix length: %m");
 
+        if (in6_addr_is_null(&dst) && prefixlen == 0) {
+                log_link_debug(link, "Route prefix is ::/0, ignoring");
+                return 0;
+        }
+
         if (in6_prefix_is_filtered(&dst, prefixlen, link->network->ndisc_allow_listed_route_prefix, link->network->ndisc_deny_listed_route_prefix)) {
                 if (DEBUG_LOGGING) {
                         _cleanup_free_ char *buf = NULL;


yuwata:

bluca (approved):

src/network/networkd-ndisc.c
```diff
@@ -612,6 +612,11 @@ static int ndisc_router_process_route(Link *link, sd_ndisc_router *rt) {
         if (r < 0)
                 return log_link_error_errno(link, r, "Failed to get route prefix length: %m");
 
+        if (!link->network->ipv6_accept_ra_use_gateway &&
```
tomty89: Hmm, in `ndisc_router_process_default()`:

https://github.com/systemd/systemd/blob/a420d71793bcbc1539a63be60f83cdc14373ea4a/src/network/networkd-ndisc.c#L346-L363

Do we even need `!link->network->ipv6_accept_ra_use_gateway &&` here? Or are they actually for different stuff?

yuwata: When networkd receives an RA, then networkd creates the default gateway via the router address. Also, networkd creates routes defined in route prefixes. If an RA contains a route prefix whose destination is ::/0, then the created route becomes the default gateway. Hence, if `UseGateway=no` is set, we also need to check route prefixes contained in RA.

tomty89: > If an RA contains a route prefix whose destination is ::/0, then the created route becomes the default gateway.

And the nexthop/via of that will always also be the router address? If that's the case, `ndisc_router_process_route()` can return (0) whenever `in6_addr_is_null(&dst) && prefixlen == 0`, regardless of whether `link->network->ipv6_accept_ra_use_gateway` is false? (Because `ndisc_router_process_default()` will do the same thing regardless of whether the RA contains such route prefix?)

yuwata: Indeed. Updated.

millnet-maho:
Wouldn't it be a good idea to distinguish between automatically using the router as the default gateway and accepting an explicit ::/0 route? I'd like to turn off the former but still have networkd process explicit default routes (route info option (24) of ::/0). (Specifically, if you have two or more routers on the same network but only one advertises ::/0.)

yuwata:
@millnet-maho Please file as a new request. Otherwise we can easily forget your request.

PR #21924, opened 2021-12-28
stub: Do not assume having DeviceHandle

ksa678491784:
My bootloader does not load EFI stub from disk, but from memory instead and that leads to "assertion 'handle' failed at disk.c:13" ([here](https://github.com/systemd/systemd/blob/4b3ad81bfafcd97acb06db463495e348d159d8e6/src/boot/efi/disk.c#L13)). After some digging it turned out that DeviceHandle is null, so I added checks for it.

### DIFF
--- a/src/boot/efi/cpio.c
+++ b/src/boot/efi/cpio.c
@@ -330,6 +330,12 @@ EFI_STATUS pack_cpio(
         assert(ret_buffer);
         assert(ret_buffer_size);
 
+        if (!loaded_image->DeviceHandle) {
+                *ret_buffer = NULL;
+                *ret_buffer_size = 0;
+                return EFI_SUCCESS;
+        }
+
         root = LibOpenRoot(loaded_image->DeviceHandle);
         if (!root)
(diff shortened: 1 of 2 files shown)


medhefgo (changes requested):
Please move the check inside of `disk_get_part_uuid/pack_cpio` instead.

ksa678491784:
The one used internally. It loads any EFI application from specific hardware, makes some small modifications, then runs LoadImage and StartImage.

medhefgo:
> The one used internally. It loads any EFI application from specific hardware, makes some small modifications, then runs LoadImage and StartImage.

Is that the firmware one are a custom efi boot loader? It should really pass the file path to LoadImage (assuming it has one to pass of course)...

Commits need squashing.

ksa678491784:
Firmware does not recognize partition (or maybe hardware that holds that partition) that contains an application, so there is no valid path to pass to LoadImage.

medhefgo (approved):
Should probably be backported to v250-stable.

PR #21922, opened 2021-12-28
boot: More BCD parser fixes

medhefgo:
Partially fixes: #21917

PR #10220, opened 2018-09-30
meson: use an array option for efi-cc

floppym:
Fixes: https://github.com/systemd/systemd/issues/10211

PR #21928, opened 2021-12-29
meson: Boot

medhefgo:


src/boot/efi/meson.build
```diff
@@ -3,7 +3,7 @@
 conf.set10('ENABLE_EFI', get_option('efi'))
 conf.set10('HAVE_GNU_EFI', false)
 
-if not get_option('efi') or get_option('gnu-efi') == 'false'
+if skip_deps or not get_option('efi') or get_option('gnu-efi') == 'false'
```
evverx: I think this will break the fuzz target I'm trying to add in https://github.com/systemd/systemd/pull/21915 (because on OSS-Fuzz skip-deps is set to true always). Considering that it's the only place where the fuzz targets are run on a regular basis it would mean that that fuzz target could be effectively thrown away

medhefgo: The fuzzer can easily be added in front of this check.

evverx: as far as I understand it should be wrapped in
```
if efi_arch[1] in ['ia32', 'x86_64', 'arm', 'aarch64']
        systemd_boot_sources += 'bcd.c'
        tests += [
                [['src/boot/efi/test-bcd.c'],
                 [],
                 [libzstd],
                 [],
                 'HAVE_ZSTD'],
        ]
endif
```
and at that point it would be complicated

medhefgo: Requiring gnu-efi to be around for fuzzing when the fuzzer doesn't even need it is complication.

evverx: I agree and I wrote a comment in that PR that `gnu-efi` shouldn't have to be installed there to build and run the fuzzer but I think there should be a better way to decouple the bcd parser from "gnu-efi". I mean, I agree it makes sense to skip all that when skip_deps is true and I generally have no problem with adding the fuzz target at the beginning but it seems to be compatible with OSS-Fuzz only and the unit tests would be kind of broken because the fuzz target would be always built and `test-bcd` is built depending on the architecture.

evverx: Looking at https://github.com/systemd/systemd/pull/21922/files, I think this particular issue could be solved by placing
```
efi_arch = {
        # host_cc_arch: [efi_arch (see Table 3-2 in UEFI spec), gnu_efi_inc_arch]
        'x86':     ['ia32', 'ia32'],
        'x86_64':  ['x64', 'x86_64'],
        'arm':     ['arm', 'arm'],
        'aarch64': ['aa64', 'aarch64'],
        'riscv64': ['riscv64', 'riscv64'],
}.get(host_machine.cpu_family(), [])
```
right before the "if" expression where skip_deps is checked. This way it should be possible to include the fuzz target depending on the architecture by analogy with `test-bcd`.

evverx: One downside would be that people compiling systemd with `-Dgnu-efi=false -Dfuzz-tests=true` would still get the bcd fuzzer they probably don't care about but I doubt anyone uses this combination in practice

evverx: With the "fuzz-bcd" branch rebased on top of this PR I changed `meson.build` and `tools/oss-fuzz` to build the fuzzer depending on the architecture and also dropped `apt-get install gnu-efi`:
```diff
diff --git a/src/boot/efi/meson.build b/src/boot/efi/meson.build
index 9e7fbced67..b5f3596fc6 100644
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -3,13 +3,6 @@
 conf.set10('ENABLE_EFI', get_option('efi'))
 conf.set10('HAVE_GNU_EFI', false)

-if skip_deps or not get_option('efi') or get_option('gnu-efi') == 'false'
-        if get_option('gnu-efi') == 'true'
-                error('gnu-efi support requested, but general efi support is disabled')
-        endif
-        subdir_done()
-endif
-
 efi_arch = {
         # host_cc_arch: [efi_arch (see Table 3-2 in UEFI spec), gnu_efi_inc_arch]
         'x86':     ['ia32', 'ia32'],
@@ -19,6 +12,26 @@ efi_arch = {
         'riscv64': ['riscv64', 'riscv64'],
 }.get(host_machine.cpu_family(), [])

+# The fuzz target is always built regardless of whether gnu-efi is enabled or not
+# because it doesn't depend on gnu-efi. It makes it possible to
+# build and run it when the gnu-efi package isn't installed
+# (like on OSS-Fuzz for example).
+#
+# The fuzzer only makes sense on arches supported by Windows
+# (by analogy with test-bcd).
+if efi_arch[1] in ['ia32', 'x86_64', 'arm', 'aarch64']
+        fuzzers += [
+                [['src/boot/efi/fuzz-bcd.c']],
+        ]
+endif
+
+if skip_deps or not get_option('efi') or get_option('gnu-efi') == 'false'
+        if get_option('gnu-efi') == 'true'
+                error('gnu-efi support requested, but general efi support is disabled')
+        endif
+        subdir_done()
+endif
+
 efi_incdir = get_option('efi-includedir')
 if efi_arch.length() > 0 and not cc.has_header('@0@/@1@/efibind.h'.format(efi_incdir, efi_arch[1]))
         efi_arch = []
diff --git a/tools/oss-fuzz.sh b/tools/oss-fuzz.sh
index 98271dd93b..df9bad9aa1 100755
--- a/tools/oss-fuzz.sh
+++ b/tools/oss-fuzz.sh
@@ -35,13 +35,7 @@ else
     apt-get update
     apt-get install -y gperf m4 gettext python3-pip \
         libcap-dev libmount-dev libkmod-dev \
-        pkg-config wget python3-jinja2
-
-    # gnu-efi is installed here to enable -Dgnu-efi behind which fuzz-bcd
-    # is hidden. It isn't linked against efi. It doesn't
-    # even include "efi.h" because "bcd.c" can work in "unit test" mode
-    # where it isn't necessary.
-    apt-get install -y gnu-efi zstd
+        pkg-config wget python3-jinja2 zstd

     pip3 install -r .github/workflows/requirements.txt --require-hashes
```

evverx: I believe this conversation is resolved. I'm not sure why I didn't press the button

meson_options.txt
```diff
@@ -418,8 +418,6 @@ option('dbus', type : 'combo', choices : ['auto', 'true', 'false'],
 
 option('gnu-efi', type : 'combo', choices : ['auto', 'true', 'false'],
        description : 'gnu-efi support for sd-boot')
-option('efi-cc', type : 'array',
-       description : 'the compiler to use for EFI modules')
```
evverx, line 422: FWIW judging by https://github.com/systemd/systemd/pull/10220 `efi-cc` is used in the wild so it seems it can break build scripts suddenly. I wonder if it would be better to somehow "deprecate" it instead of just removing it?

medhefgo: That one is about (presumably) `CC="ccache cc" meson` failing. It also refers to df7cacae696ae3c1d13d2f9a4db24815e07e29a1, which talks about a proper cross-build. So i tend to disagree on this. Currently, using efi-cc is *very* likely not going to do what one intends to.

evverx: > using efi-cc is very likely not going to do what one intends to.

I agree it makes sense to fail hard then.

One last thing. It doesn't seem to work when systemd is compiled with compiler wrappers like hfuzz-clang/hfuzz-clang++ but it seems it's much easier to just pass -Dgnu-efi=false than to make them work together.

medhefgo: If there's an easy way to detect them that would be best. In the worst case you can always hardcode a denylist for gnu-efi=auto.

medhefgo: Does compilation for this test file break on those wrappers when run like this: `$CC -nodefaultlibs -lc test.c`?
```c
int main(int argc, char **argv) {
        return argc;
}
```

evverx: With hfuzz-gcc/hfuzz-clang it does
```
$ gcc test.c
$ hfuzz-gcc test.c
$ hfuzz-gcc -nodefaultlibs -lc  test.c
/usr/bin/ld: /tmp/libhfuzz.1000.c49a3fb6f63a0022.a(instrument.o): undefined reference to symbol '__tls_get_addr@@GLIBC_2.3'
/usr/bin/ld: /lib64/ld-linux-x86-64.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
With afl-gcc it seems to work
```
$ afl-gcc -nodefaultlibs -lc test.c
afl-cc 2.57b by <lcamtuf@google.com>
afl-as 2.57b by <lcamtuf@google.com>
[+] Instrumented 1 locations (64-bit, non-hardened mode, ratio 100%).
```

evverx: In https://github.com/systemd/systemd/pull/21859 I just looked for "afl-clang/hfuzz-clang" to detect them and replace with clang. But in that case it was easy because BPF skeletons shouldn't be instrumented and they aren't linked against anything. With "boot" it seems it's not that easy as far as I can tell. At least I haven't figured it out yet

src/boot/efi/meson.build
```diff
@@ -43,11 +43,7 @@ if not cc.has_header_symbol('efi.h', 'EFI_IMAGE_MACHINE_X64',
         subdir_done()
 endif
 
-objcopy = find_program('objcopy')
-efi_cc = get_option('efi-cc')
-if efi_cc.length() == 0
-        efi_cc = cc.cmd_array()
-endif
+objcopy = run_command(cc.cmd_array(), '-print-prog-name=objcopy', check: true).stdout().strip()
```
yuwata, line 46: `--print-prog-name`?? In the main meson.build, we use `--print-prog-name` for detecting `llvm-strip`.

yuwata (approved):
LGTM. One question.

src/boot/efi/meson.build
medhefgo: Both gcc and clang list this with one dash in `--help`, so I prefer that. They happily accept either, though.

yuwata: Indeed. Thanks.

src/boot/efi/meson.build
```diff
@@ -43,7 +43,7 @@ if not cc.has_header_symbol('efi.h', 'EFI_IMAGE_MACHINE_X64',
         subdir_done()
 endif
 
-objcopy = find_program('objcopy')
+objcopy = run_command(cc.cmd_array(), '-print-prog-name=objcopy', check: true).stdout().strip()
```
Artturin, line 46: @medhefgo do you remember why this was changed from find_program? now a prefixed `objcopy` can't be specified in a meson cross-file binaries section

i.e.
```meson
[binaries]
objcopy = 'aarch64-unknown-linux-gnu-objcopy'
```

PR #21930, opened 2021-12-29
oss-fuzz: turn on the alignment check

evverx:
just to trigger all the fuzz targets to figure out how safe
it is to roll it out globally in https://github.com/systemd/systemd/pull/21915.


evverx:
Looks like it should be good to go

PR #21937, opened 2021-12-30
sysusers: use filename if /proc is not mounted

yuwata:
During system install, /proc may not be mounted yet.

Fixes RHBZ#2036217 (https://bugzilla.redhat.com/show_bug.cgi?id=2036217).


yuwata:
Needs to be backported to v250-stable.

bluca (approved):

PR #21938, opened 2021-12-30
nss-systemd: fix required buffer size calculation

### LINKED ISSUE #21935: SIGBUS on SPARC caused by NSS lookups
>Known good versions:
>- sys-apps/systemd-249.4-r4 (but not recompiled it yet)
>
>Known bad versions:
>- sys-apps/systemd-249.6-r1
>- sys-apps/systemd-249.7
>- sys-apps/systemd-250-r1

**Used distribution**
 > Gentoo Linux.

 > `5.15.5-gentoo`

 > sparc64

 > Commands which end up involving NSS lookups like `ls -al` succeeding.

 > SIGBUS on `ls -al`, `sudo su`.

 > 1. Install >sys-apps/systemd-249.4 on a `sparc64` machine
 > 2. Have systemd configured for NSS lookups in `/etc/nsswitch.conf`
 > 3. Try to run `ls -al` or some other command which will need to lookup entries via NSS

Originally reported downstream in Gentoo [here](https://bugs.gentoo.org/830275).

```
$ sudo su
Bus error (core dumped)
```

```
$ ls -al
Bus error (core dumped)
```

Backtrace:
```
(gdb) bt
#0  copy_synthesized_group (dest=0xfff800010049bf90 <resbuf>, src=0xfff8000100706b08 <root_group>, buffer=0x100001382a0 "root", buflen=1024, errnop=0xfff800010002fbd8)
    at ../systemd-250/src/nss-systemd/nss-systemd.c:254
#1  0xfff80001005b6730 in _nss_systemd_getgrgid_r (gid=<optimized out>, gr=0xfff800010049bf90 <resbuf>, buffer=0x100001382a0 "root", buflen=1024, errnop=0xfff800010002fbd8)
    at ../systemd-250/src/nss-systemd/nss-systemd.c:498
#2  0xfff80001002f1f54 in __getgrgid_r (gid=<optimized out>, resbuf=0xfff800010049bf90 <resbuf>, buffer=0x100001382a0 "root", buflen=1024, result=0x7fefff207e0) at ../nss/getXXbyYY_r.c:274
#3  0xfff80001002f150c in getgrgid (gid=<optimized out>) at ../nss/getXXbyYY.c:135
#4  0x0000010000012d3c in getgroup (gid=<optimized out>) at lib/idcache.c:167
#5  0x000001000000dc84 in format_group_width (g=0) at src/ls.c:4183
#6  gobble_file (name=0x1000012f623 "glibc-2.30-r9.ebuild", type=<optimized out>, command_line_arg=<optimized out>, dirname=<optimized out>, inode=0) at src/ls.c:3534
#7  0x000001000000e484 in print_dir (name=0x100001276e0 ".", realname=<optimized out>, command_line_arg=<optimized out>) at src/ls.c:2989
#8  0x00000100000046e4 in main (argc=<optimized out>, argv=<optimized out>) at src/ls.c:1778
(gdb)
```

Note that the machine is running glibc-2.33 and the 2.30 reference in the backtrace is just because there's a bunch of glibc ebuilds (=filenames) in the directory I was in.

I don't think there was anything else related/useful in the journal.

yuwata:
This also fixes the pointer assigned to the gr_mem element of struct group.

Fixes a bug introduced by 47fd7fa6c650d7a0ac41bc89747e3b866ffb9534.

Fixes #21935.

### DIFF
--- a/src/nss-systemd/nss-systemd.c
+++ b/src/nss-systemd/nss-systemd.c
@@ -236,7 +236,7 @@ static enum nss_status copy_synthesized_group(
 
         required = strlen(src->gr_name) + 1;
         required += strlen(src->gr_passwd) + 1;
-        required += 1; /* ...but that NULL still needs to be stored into the buffer! */
+        required += sizeof(char*); /* ...but that NULL still needs to be stored into the buffer! */
 
         if (buflen < required) {
                 *errnop = ERANGE;
@@ -250,7 +250,7 @@ static enum nss_status copy_synthesized_group(
         /* String fields point into the user-provided buffer */
         dest->gr_name = buffer;
         dest->gr_passwd = stpcpy(dest->gr_name, src->gr_name) + 1;
-        dest->gr_mem = (char **) strcpy(dest->gr_passwd, src->gr_passwd) + 1;
+        dest->gr_mem = (char **) stpcpy(dest->gr_passwd, src->gr_passwd) + 1;
         *dest->gr_mem = NULL;
 
         return NSS_STATUS_SUCCESS;


yuwata:

Arfrever:
Problematic commit was also [backported](https://github.com/systemd/systemd-stable/commit/5e44156893515e031491346f7cac960c9317ccd0) to branch v248-stable.

src/nss-systemd/nss-systemd.c
```diff
@@ -250,7 +250,7 @@ static enum nss_status copy_synthesized_group(
         /* String fields point into the user-provided buffer */
         dest->gr_name = buffer;
         dest->gr_passwd = stpcpy(dest->gr_name, src->gr_name) + 1;
-        dest->gr_mem = (char **) strcpy(dest->gr_passwd, src->gr_passwd) + 1;
+        dest->gr_mem = (char **) stpcpy(dest->gr_passwd, src->gr_passwd) + 1;
```
bluca, line 253: uh, how did this ever work?

bluca (approved):

yuwata:
> Problematic commit was also [backported](https://github.com/systemd/systemd-stable/commit/5e44156893515e031491346f7cac960c9317ccd0) to branch v248-stable.

Oh. Thanks.

src/nss-systemd/nss-systemd.c
yuwata: I guess this did not work. We do not have enough tests for nss modules...

PR #21939, opened 2021-12-30
fuzz: set maximum data size to avoid timeouts

yuwata:
This also increases the maximum data size.

C.f. https://github.com/systemd/systemd/pull/21915#discussion_r776745240

### DIFF
--- a/src/core/fuzz-unit-file.options
+++ b/src/core/fuzz-unit-file.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/fuzz/fuzz-env-file.options
+++ b/src/fuzz/fuzz-env-file.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65535
--- a/src/journal-remote/fuzz-journal-remote.options
+++ b/src/journal-remote/fuzz-journal-remote.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/journal/fuzz-journald-stream.options
+++ b/src/journal/fuzz-journald-stream.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/libsystemd-network/fuzz-dhcp-server.options
+++ b/src/libsystemd-network/fuzz-dhcp-server.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 600
--- a/src/libsystemd-network/fuzz-dhcp6-client.options
+++ b/src/libsystemd-network/fuzz-dhcp6-client.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/libsystemd-network/fuzz-lldp-rx.options
+++ b/src/libsystemd-network/fuzz-lldp-rx.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 2048
--- a/src/libsystemd-network/fuzz-ndisc-rs.options
+++ b/src/libsystemd-network/fuzz-ndisc-rs.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 2048
--- a/src/network/fuzz-network-parser.options
+++ b/src/network/fuzz-network-parser.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65535
--- a/src/nspawn/fuzz-nspawn-oci.options
+++ b/src/nspawn/fuzz-nspawn-oci.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/nspawn/fuzz-nspawn-settings.options
+++ b/src/nspawn/fuzz-nspawn-settings.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/resolve/fuzz-dns-packet.options
+++ b/src/resolve/fuzz-dns-packet.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65535
--- a/src/udev/fuzz-udev-rules.options
+++ b/src/udev/fuzz-udev-rules.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65536
--- a/src/udev/net/fuzz-link-parser.options
+++ b/src/udev/net/fuzz-link-parser.options
@@ -1,2 +1,0 @@
-[libfuzzer]
-max_len = 65535
--- a/tools/oss-fuzz.sh
+++ b/tools/oss-fuzz.sh
@@ -75,4 +75,3 @@ wget -O "$OUT/fuzz-json.dict" https://raw.githubusercontent.com/rc0r/afl-fuzz/ma
 
 find "$build" -maxdepth 1 -type f -executable -name "fuzz-*" -exec mv {} "$OUT" \;
 find src -type f -name "fuzz-*.dict" -exec cp {} "$OUT" \;
-cp src/fuzz/*.options "$OUT"


yuwata:
cc @evverx.

evverx:
@yuwata thanks! While I agree that something like that should be added eventually it isn't always clear when fuzz targets should reject data. In https://github.com/systemd/systemd/pull/21915 the limit wasn't arbitrary because it was borrowed from the part of the code that calls the bcd parser. As far as I can remember arbitrary limits for example prevented OSS-Fuzz from finding a bug in NSS: https://googleprojectzero.blogspot.com/2021/12/this-shouldnt-have-happened.html

evverx:
FWIW I'd just drop the "*.options" files where `max_len` is set because with AFL and honggfuzz they are kind out of date and somewhat misleading.

evverx:
And honestly I'm not sure where `if (size == 0 ...)` came from :-) Looking at 18d51b45097f51, @mrc0mmand I wonder if that was intentional? It seems that that fuzzer should handle empty files just fine as far as I can tell.

mrc0mmand:
> And honestly I'm not sure where `if (size == 0 ...)` came from :-) Looking at [18d51b4](https://github.com/systemd/systemd/commit/18d51b45097f510676ce0ae90ff8becdd3564f3c), @mrc0mmand I wonder if that was intentional? It seems that that fuzzer should handle empty files just fine as far as I can tell.

To be quite honest, I don't remember either why it's there, so feel free drop it if it's not needed.

evverx:
Looks like it's there to get around a glibc bug probably :-)
> In  versions  of glibc before 2.22, if size is specified as zero, fmemopen() fails with the error EINVAL.  It would be more consistent if this case successfully created a stream that then re‐
       turned end-of-file on the first attempt at reading; since version 2.22, the glibc implementation provides that behavior.

Given that the base builder image on Ubuntu was updated it should be fixed there.

evverx:
I removed "size == 0" in fuzz-env-file and judging by
```
$ ./out/fuzz-env-file EMPTY
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 563569067
INFO: Loaded 2 modules   (74041 inline 8-bit counters): 74030 [0x7f6756edf550, 0x7f6756ef167e), 11 [0x58cd60, 0x58cd6b),
INFO: Loaded 2 PC tables (74041 PCs): 74030 [0x7f6756ef1680,0x7f6757012960), 11 [0x564550,0x564600),
./out/fuzz-env-file: Running 1 inputs 1 time(s) each.
Running: EMPTY
Assertion 'f' failed at src/fuzz/fuzz-env-file.c:20, function int LLVMFuzzerTestOneInput(const uint8_t *, size_t)(). Aborting.
==184768== ERROR: libFuzzer: deadly signal
    #0 0x519e5e in __sanitizer_print_stack_trace (/home/vagrant/systemd/out/fuzz-env-file+0x519e5e)
    #1 0x467519 in fuzzer::PrintStackTrace() (/home/vagrant/systemd/out/fuzz-env-file+0x467519)
    #2 0x448976 in fuzzer::Fuzzer::CrashCallback() (.part.0) FuzzerLoop.cpp.o
    #3 0x448a3b in fuzzer::Fuzzer::StaticCrashSignalCallback() (/home/vagrant/systemd/out/fuzz-env-file+0x448a3b)
    #4 0x7f675605574f  (/lib64/libc.so.6+0x4274f)
    #5 0x7f67560a284b in __pthread_kill_implementation (/lib64/libc.so.6+0x8f84b)
    #6 0x7f67560556a5 in gsignal (/lib64/libc.so.6+0x426a5)
    #7 0x7f675603f7d2 in abort (/lib64/libc.so.6+0x2c7d2)
    #8 0x7f6756999619 in log_assert_failed /home/vagrant/systemd/build/../src/basic/log.c:866:9
    #9 0x54921a in LLVMFuzzerTestOneInput /home/vagrant/systemd/build/../src/fuzz/fuzz-env-file.c:20:9
    #10 0x448ea9 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/vagrant/systemd/out/fuzz-env-file+0x448ea9)
    #11 0x42994f in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) (/home/vagrant/systemd/out/fuzz-env-file+0x42994f)
    #12 0x43160e in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/vagrant/systemd/out/fuzz-env-file+0x43160e)
    #13 0x420ab6 in main (/home/vagrant/systemd/out/fuzz-env-file+0x420ab6)
    #14 0x7f675604055f in __libc_start_call_main (/lib64/libc.so.6+0x2d55f)
    #15 0x7f675604060b in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2d60b)
    #16 0x420b04 in _start (/home/vagrant/systemd/out/fuzz-env-file+0x420b04)

NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
```
it doesn't seem to be working with glibc-2.34 so I'm not sure the fmemopen manpage describes what actually happens :-)

Anyway, it seems `size == 0` shouldn't be used when `fmemopen` isn't used. I think ideally it should be replaced with something that could actually handle empty files because as far as I can remember in the past "empty strings" passed to systemd caused it to either crash or hang and probably it should never happen again :-)

@yuwata let's put this PR on hold at least until 2022 :-)

src/journal/fuzz-journald-audit.c
```diff
@@ -7,6 +7,9 @@
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         Server s;
 
+        if (size > 65536)
```
evverx: I think the journald fuzzers should reject buffers that are a bit larger than the underlying functions are supposed to handle. Looking at server_process_audit_message it seems it receives messages via a socket so this limit should be a bit larger that what can be received via that socket.

The fuzzers fuzzing various argv functions should reject buffers that are much larger than 65536 bytes to be able to cover all the corner cases.

Anyway, I'm not sure I can pick sensible defaults in a reasonable amount of time :-)

yuwata:
@evverx Thank you for the comments and suggestions. Now the maximum data size are set only for fuzzers which often timeouts.

evverx:
@yuwata judging by
```
$ ls -lh test/fuzz/fuzz-unit-file/
total 4.2M
-rw-r--r--. 1 vagrant vagrant  216 Oct 12 12:44 'dev-mapper-fedora_krowka\x2dswap.swap'
-rw-r--r--. 1 vagrant vagrant  15K Dec  1 03:58  directives-all.service
-rw-r--r--. 1 vagrant vagrant   74 Nov 26 15:22  directives.automount
-rw-r--r--. 1 vagrant vagrant 3.4K Nov 11 00:43  directives.mount
-rw-r--r--. 1 vagrant vagrant  166 Dec 21 08:40  directives.path
-rw-r--r--. 1 vagrant vagrant 1.1K Oct 12 12:44  directives.scope
-rw-r--r--. 1 vagrant vagrant 6.1K Dec  1 03:58  directives.service
-rw-r--r--. 1 vagrant vagrant  954 Oct 12 12:44  directives.slice
-rw-r--r--. 1 vagrant vagrant 4.1K Nov 11 00:43  directives.socket
-rw-r--r--. 1 vagrant vagrant 3.3K Nov 11 00:43  directives.swap
-rw-r--r--. 1 vagrant vagrant  233 Oct 12 12:44  directives.timer
-rw-r--r--. 1 vagrant vagrant   14 Oct 12 12:44  empty.scope
-rw-r--r--. 1 vagrant vagrant   61 Oct 12 12:44  github-19178
-rw-r--r--. 1 vagrant vagrant  459 Oct 12 12:44  machine.slice
-rw-r--r--. 1 vagrant vagrant 751K Oct 12 12:44  oss-fuzz-10007
-rw-r--r--. 1 vagrant vagrant 271K Oct 12 12:44  oss-fuzz-11569
-rw-r--r--. 1 vagrant vagrant 591K Oct 12 12:44  oss-fuzz-13125
-rw-r--r--. 1 vagrant vagrant   35 Oct 12 12:44  oss-fuzz-32991
-rw-r--r--. 1 vagrant vagrant 6.9K Oct 12 12:44  oss-fuzz-33270
-rw-r--r--. 1 vagrant vagrant 6.1K Oct 12 12:44  oss-fuzz-33876
-rw-r--r--. 1 vagrant vagrant   31 Oct 12 12:44  oss-fuzz-6884
-rw-r--r--. 1 vagrant vagrant   32 Oct 12 12:44  oss-fuzz-6885
-rw-r--r--. 1 vagrant vagrant   43 Oct 12 12:44  oss-fuzz-6886
-rw-r--r--. 1 vagrant vagrant   38 Oct 12 12:44  oss-fuzz-6892
-rw-r--r--. 1 vagrant vagrant   55 Oct 12 12:44  oss-fuzz-6897
-rw-r--r--. 1 vagrant vagrant   61 Oct 12 12:44  oss-fuzz-6897-evverx
-rw-r--r--. 1 vagrant vagrant   40 Oct 12 12:44  oss-fuzz-6908
-rw-r--r--. 1 vagrant vagrant 131K Oct 12 12:44  oss-fuzz-6917
-rw-r--r--. 1 vagrant vagrant 839K Oct 12 12:44  oss-fuzz-6977
-rw-r--r--. 1 vagrant vagrant 686K Oct 12 12:44  oss-fuzz-6977-unminimized
-rw-r--r--. 1 vagrant vagrant   40 Oct 12 12:44  oss-fuzz-7004
-rw-r--r--. 1 vagrant vagrant   57 Oct 12 12:44  oss-fuzz-8064
-rw-r--r--. 1 vagrant vagrant 791K Oct 12 12:44  oss-fuzz-8827
-rw-r--r--. 1 vagrant vagrant  763 Oct 12 12:44  proc-sys-fs-binfmt_misc.automount
-rw-r--r--. 1 vagrant vagrant 2.7K Oct 12 12:44  syslog.socket
-rw-r--r--. 1 vagrant vagrant  717 Oct 12 12:44  systemd-ask-password-console.path
-rw-r--r--. 1 vagrant vagrant 1.7K Oct 12 12:44  systemd-machined.service
-rw-r--r--. 1 vagrant vagrant 1.7K Oct 12 12:44  systemd-resolved.service
-rw-r--r--. 1 vagrant vagrant  858 Oct 12 12:44  systemd-tmpfiles-clean.timer
-rw-r--r--. 1 vagrant vagrant  460 Oct 12 12:44  timers.target
-rw-r--r--. 1 vagrant vagrant  529 Oct 12 12:44  var-lib-machines.mount
```
some of those issues were triggered by files that were much larger than 64Kb so I'm not sure fuzz-unit-file should start rejecting files like that. I'm also not sure that fuzzers like fuzz-dhcp-server and fuzz-journal-remote that fuzz functions receiving arbitrary data over the network should reject files like that. In those cases I think it's better to get bogus timeouts and keep catching issues that are potentially remotely exploitable.

yuwata:
@evverx OK. I will re-consider that tomorrow or later. Thank you!

keszybz:
I don't think we want to replace the limits in the .options files with the check in the code. The issue is that with the options setting, the runner will never try sizes over the limit. When the check is done in the code, it has to "learn" the size leads to an immediate return. This inelegant, and depending on whether coverage-feedback is available, might not work immediately.

evverx:
> The issue is that with the options setting, the runner will never try sizes over the limit. When the check is done in the code, it has to "learn" the size leads to an immediate return. This inelegant, and depending on whether coverage-feedback is available, might not work immediately.

Corpora are cross-polinated on OSS-Fuzz regularly (which partly can explain why it can take it a couple of hours instead of days or weeks to find new issues) so there that feedback is received pretty fast but I agree that if the fuzzers are run locally those hints could be helpful.

yuwata:
Replaced by #23310. Closing.

PR #21941, opened 2021-12-30
hostname-setup: support kernel with empty CONFIG_DEFAULT_HOSTNAME

### LINKED ISSUE #21896: "Failed to set hostname" if CONFIG_DEFAULT_HOSTNAME is empty
 > 250

**Used distribution**
 > Linux From Scratch

 > `5.16.0-rc6+`

 > mips64

 > systemd should update hostname to the content of /etc/hostname

 > "Failed to set hostname to \<xry111-A1901\>: No such device or address"

 > Build a kernel with empty CONFIG_DEFAULT_HOSTNAME, and boot it.

yuwata:
Fixes #21896.

### DIFF
--- a/src/basic/hostname-util.c
+++ b/src/basic/hostname-util.c
@@ -46,8 +46,7 @@ int gethostname_full(GetHostnameFlags flags, char **ret) {
         assert_se(uname(&u) >= 0);
 
         s = u.nodename;
-        if (isempty(s) ||
-            (!FLAGS_SET(flags, GET_HOSTNAME_ALLOW_NONE) && streq(s, "(none)")) ||
+        if (isempty(s) || streq(s, "(none)") ||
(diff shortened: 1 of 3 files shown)


yuwata:

bluca:
what about adding something like `GET_HOSTNAME_ALLOW_EMPTY` instead, so that we can keep using the same internal helpers everywhere?

yuwata:
> what about adding something like `GET_HOSTNAME_ALLOW_EMPTY` instead, so that we can keep using the same internal helpers everywhere?

In `sethostname_idempotent_full()`, not only empty string, any hostname currently configured should be accepted. So, it is better to just call `uname()` there.

keszybz:
Yeah, this seems to DTRT in a simpler way.

PR #21942, opened 2021-12-30
meson: fix detection of libcryptsetup functions

keszybz:
Meson would generate the following compile test:
```C
  #define crypt_set_metadata_size meson_disable_define_of_crypt_set_metadata_size

  #include <limits.h>
  #undef crypt_set_metadata_size

  #ifdef __cplusplus
  extern "C"
  #endif
  char crypt_set_metadata_size (void);

  #if defined __stub_crypt_set_metadata_size || defined __stub___crypt_set_metadata_size
  fail fail fail this function is not going to work
  #endif

  int main(void) {
    return crypt_set_metadata_size ();
  }
```

This works fine when the identifier being queried is an actual function. But
crypt_token_max() is an inline function, so getting the address would fail,
leading to a false negative result. Complation would fail because the function
would be defined twice.

With this patch, the check is changed to include the header:
```C
  #include <libcryptsetup.h>
  #include <limits.h>

  #if defined __stub_crypt_set_metadata_size || defined __stub___crypt_set_metadata_size
  fail fail fail this function is not going to work
  #endif

  int main(void) {
    void *a = (void*) &crypt_set_metadata_size;
    long long b = (long long) a;
    return (int) b;
  }
```
which seems to work correctly.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1182,12 +1182,15 @@ if want_libcryptsetup != 'false' and not skip_deps
                                    required : want_libcryptsetup == 'true' or want_libcryptsetup_plugins == 'true')
         have = libcryptsetup.found()
 
-        conf.set10('HAVE_CRYPT_SET_METADATA_SIZE',
-                   have and cc.has_function('crypt_set_metadata_size', dependencies : libcryptsetup))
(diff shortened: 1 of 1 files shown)


yuwata (approved):
LGTM.

PR #21943, opened 2021-12-30
udev: fix ID_NET_NAME_MAC= udev property

yuwata:
This fixes a bug introduced by eaba9bb3e69635d2c490c5e1b0d262b763753e1d.

The commit mistakenly drops 'x' in ID_NET_NAME_MAC, and adds colons.
The colons were dropped by the commit dfa4876c417e2a9935d58100d44d94bb41cd5bfb,
but the missing 'x' was not added at that time.

Follow-up for dfa4876c417e2a9935d58100d44d94bb41cd5bfb.


yuwata:

bluca (approved):

PR #21954, opened 2022-01-01
sd-bus/man: document EBUSY error in bus_message_read

oldaccountdeadname:
I had to dig through the source to find the cause of this error, so hopefully the docs introduced in this commit are correct and would be helpful. Thanks for taking a look at the commit, and happy 2022!

Here's the commit message:

```
The EBUSY error can be returned from sd_bus_exit_container(), and, if
that happens, it will be propogated upwards towards bus_message_read. In
terms of documentation, this means that bus_message_read's man page
can't just include the error text for sd_bus_message_read_basic, as
reading basic types exclusively doesn't have the potential for this
error.

sd_bus_message_read_basic's error documentation isn't incorrect when
applied to sd_bus_message_read, it's just incomplete. This patch
replaces the include directive with a manual copy supplemented with
documentation of EBUSY.
```

### DIFF
--- a/man/sd_bus_message_read.xml
+++ b/man/sd_bus_message_read.xml
@@ -178,26 +178,9 @@
       <para>Returned errors may indicate the following problems:</para>
 
       <variablelist>
-        <varlistentry>
-          <term><constant>-EINVAL</constant></term>
-
-          <listitem><para>Specified type string is invalid or the message parameter is
-          <constant>NULL</constant>.</para></listitem>
-        </varlistentry>
-
-        <varlistentry>
-          <term><constant>-ENXIO</constant></term>
-
-          <listitem><para>The message does not contain the specified type at current position.
-          </para></listitem>
-        </varlistentry>
-
-        <varlistentry>
-          <term><constant>-EBADMSG</constant></term>
-
-          <listitem><para>The message cannot be parsed.</para></listitem>
-        </varlistentry>
-
+        <xi:include href="sd_bus_message_read_basic.xml" xpointer="errors-einval"/>
+        <xi:include href="sd_bus_message_read_basic.xml" xpointer="errors-enxio"/>
+        <xi:include href="sd_bus_message_read_basic.xml" xpointer="errors-ebadmsg"/>
         <varlistentry>
           <term><constant>-EBUSY</constant></term>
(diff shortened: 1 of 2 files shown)


lincolnauster:
Ah, looks like I mistyped a tag. Force-pushed an ammended commit, CI might pass now.

man/sd_bus_message_read.xml
```diff
@@ -172,7 +172,40 @@
...
+        </varlistentry>
+
+        <varlistentry>
+          <term><constant>-EBADMSG</constant></term>
+
+          <listitem><para>The message cannot be parsed.</para></listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><constant>-EBUSY</constant></term>
```
bluca, line 185: Why not simply adding this entry to the existing list in man/sd_bus_message_read_basic.xml ?

keszybz:
Instead of copying the text, please include the descriptions of the individual errors:
```xml
    <refsect2 id='errors'>
      <title>Errors</title>

      <para>Returned errors may indicate the following problems:</para>

      <xi:include href="sd_bus_message_read_basic.xml" xpointer="errors-einval" />
      <xi:include href="sd_bus_message_read_basic.xml" xpointer="errors-enxio" />
      <xi:include href="sd_bus_message_read_basic.xml" xpointer="errors-ebadmsg" />
      ...
      <varlistentry>
          <term><constant>-EBUSY</constant></term>
This conversation was marked as resolved by keszybz

          <listitem><para>When reading from a container, this error will be returned if unread elements
          are left in the container.</para></listitem>
     </varlistentry>
...
```

(This means that id's will need to be added the the varlistentries for the individual errors.)

You can look at `common-variables.xml` for inspiration.

lincolnauster:
> Instead of copying the text, please include the descriptions of the individual errors

I believe this is done in 16d34e0e7a. Local rendering checks out, but apparently there's still a CI fail. The 177k-line log file is a bit much for me to parse, any hints? Thanks for the quick feedback and sorry for the slow patching!

yuwata:
Ubuntu CIs are slow and unstable. Please ignore the failures.

bluca (approved):

Issue #21965, opened 2022-01-02
systemd-boot does not properly handle "Initializing" drivers as documented in EDK2


DHowett:
**systemd version the issue has been seen with**
 > 250

**Used distribution**
 > Debian

**Linux kernel version used** (`uname -a`)
 > `5.15` (not relevant)

**CPU architecture issue was seen on**
 > x64

**Expected behaviour you didn't see**
 > The EDK2 Driver Writer's guide specifies a type of EFI driver that initializes hardware and exits without registering any additional platform fixtures. It calls them [Initializing Drivers](https://edk2-docs.gitbook.io/edk-ii-uefi-driver-writer-s-guide/7_driver_entry_point/78_initializing_driver_entry_point).
> Part of being an initializing driver is returning a failing status code, such as `EFI_ABORTED` (_"This type of driver returns an error from the entry point so the driver is unloaded by the UEFI image services."_).
> When systemd-boot loads such a driver from the drop-in driver directory, it should not emit an error.

**Unexpected behaviour you saw**
 > systemd-boot emits an error and pauses before continuing to boot, `xxx.efi failed to load: Aborted`

**Steps to reproduce the problem**
 > I've published a driver binary at https://github.com/DHowett/SampleEfiInitializingDriverPkg/releases/tag/initial; [the source](https://github.com/DHowett/SampleEfiInitializingDriverPkg/blob/main/SampleEfiInitializingDriver/Driver.c) is in that repository as well. All it does is return `EFI_ABORTED`.
> Drop it into `EFI/systemd/drivers`
> Boot!

medhefgo:
I find this practice rather unusual as I cannot find any reference to it in the UEFI spec itself. Please test the PR.

DHowett:
(cc @yuwata)

The fix looks like it'll work, but I agree with you--it feels strange and I also can't find anything in the UEFI Spec that backs this up.

It looks like this section of the EDK2 Driver Writer's Guide was introduced all at once in a conversion to GitBooks: https://github.com/tianocore-docs/edk2-UefiDriverWritersGuide/blob/master/7_driver_entry_point/78_initializing_driver_entry_point.md

@kwshaw1, would you be able to point us in the right direction? We are trying to track down the provenance of the edk2 book recommendation that an "Initializing Driver" return something like `EFI_ABORTED` to signal completion.

PR #21968, opened 2022-01-02
boot: Do not warn if an initializing driver returns EFI_ABORTED

medhefgo:
Fixes: #21965


yuwata (approved):
Superficially LGTM. But not sure about that this really fixes the reported issue. Let's wait for the reporter's comment.

DHowett:
Tested and working on x64. We can drive the discussion to a close over on the original issue about whether this is the correct fix. Given that the EDK2 book is maintained by Intel, I'm inclined to believe this is right (or right enough!)

Thanks all!

PR #21973, opened 2022-01-02
coredump: context->meta[META_EXE] is an optional field and may be NULL

yuwata:
Fixes RHBZ#2036517 (https://bugzilla.redhat.com/show_bug.cgi?id=2036517).

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -813,11 +813,11 @@ static int submit_coredump(
                 return log_error_errno(r, "Failed to drop privileges: %m");
 
         /* Try to get a stack trace if we can */
-        if (coredump_size > arg_process_size_max) {
+        if (coredump_size > arg_process_size_max)
                 log_debug("Not generating stack trace: core size %"PRIu64" is greater "
                           "than %"PRIu64" (the configured maximum)",
                           coredump_size, arg_process_size_max);
-        } else if (coredump_fd >= 0)
+        else if (coredump_fd >= 0 && context->meta[META_EXE])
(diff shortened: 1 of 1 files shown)


yuwata:
cc @bluca.

src/coredump/coredump.c
```diff
@@ -813,11 +813,11 @@ static int submit_coredump(
                 return log_error_errno(r, "Failed to drop privileges: %m");
 
         /* Try to get a stack trace if we can */
-        if (coredump_size > arg_process_size_max) {
+        if (coredump_size > arg_process_size_max)
                 log_debug("Not generating stack trace: core size %"PRIu64" is greater "
                           "than %"PRIu64" (the configured maximum)",
                           coredump_size, arg_process_size_max);
-        } else if (coredump_fd >= 0)
+        else if (coredump_fd >= 0 && context->meta[META_EXE])
```
bluca, line 820: Maybe use `strna` instead? It should extract the core just fine when there's no executable name

bluca: also do you have a backtrace? would be good to figure out where it crashes, as it means there's a missing assert

keszybz:
Oh, I did very similar change in #21976… @yuwata please mention a pull request on the downstream bug too when submitting it, so there's less chance of duplicate work. The changes are essentially equivalent, but my patch has a commit message that actually explains the change, so I'll close this one.

yuwata:
@keszybz Yeah, I sometimes do so, but in this case I forgot that. Sorry for wasting your time.

PR #21976, opened 2022-01-03
coredump: do not crash if we failed to acquire exe path

keszybz:
The COREDUMP_EXE attribute is "optional", i.e. we continue to process the
crash even if we didn't acquire it. The coredump generation code assumed
that it is always available:

 #5 endswith at ../src/fundamental/string-util-fundamental.c:41
 [ endswith() is called with NULL here, and an assertion fails. ]
 #6 submit_coredump at ../src/coredump/coredump.c:823
 #7 process_socket at ../src/coredump/coredump.c:1038
 #8 run at ../src/coredump/coredump.c:1413

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2036517.

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -813,16 +813,19 @@ static int submit_coredump(
                 return log_error_errno(r, "Failed to drop privileges: %m");
 
         /* Try to get a stack trace if we can */
-        if (coredump_size > arg_process_size_max) {
+        if (coredump_size > arg_process_size_max)
                 log_debug("Not generating stack trace: core size %"PRIu64" is greater "
                           "than %"PRIu64" (the configured maximum)",
                           coredump_size, arg_process_size_max);
-        } else if (coredump_fd >= 0)
+        else if (coredump_fd >= 0) {
+                bool skip = startswith(context->meta[META_COMM], "systemd-coredum"); /* COMM is 16 bytes usually */
+
                 (void) parse_elf_object(coredump_fd,
                                         context->meta[META_EXE],
-                                        /* fork_disable_dump= */endswith(context->meta[META_EXE], "systemd-coredump"), /* avoid loops */
+                                        /* fork_disable_dump= */ skip, /* avoid loops */
                                         &stacktrace,
                                         &json_metadata);
+        }
 
 log:
         core_message = strjoina("Process ", context->meta[META_ARGV_PID],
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
This is the same as #21973 no? Should that one be closed?

src/coredump/coredump.c
```diff
@@ -813,16 +813,22 @@ static int submit_coredump(
                 return log_error_errno(r, "Failed to drop privileges: %m");
 
         /* Try to get a stack trace if we can */
+        const char *exe = context->meta[META_EXE];
```
bluca: Use strna here? Things should work without the exe name

keszybz: Hmm, dwfl_core_file_report() seems to be fine will NULL.
I'll rework this.

src/coredump/coredump.c
```diff
@@ -860,8 +863,9 @@ static int submit_coredump(
         JSON_VARIANT_OBJECT_FOREACH(module_name, module_json, json_metadata) {
                 JsonVariant *package_name, *package_version;
 
-                /* We only add structured fields for the 'main' ELF module */
-                if (!path_equal_filename(module_name, context->meta[META_EXE]))
+                /* We only add structured fields for the 'main' ELF module, and only if we can identify it. */
+                if (!context->meta[META_EXE] ||
```
bluca: Can META_COMM be used here too?

keszybz: IIUC, no. When parsing the core file we use file names, and COMM could be anything.

bluca: Is it worth trying at least some best-effort approach, eg `string_contains_word(module_name, context->meta[META_COMM])` that might give us some more hits?

keszybz: I don't think we care too much. The cases where META_EXE is not available should be fairly rare, only when something went wrong when reading the attributes. So I think it's fine to just skip this extra information in those cases.

bluca: ok, in that case I'd do

```
if (!context->meta[META_EXE])
  break;
```

and avoid looping around

bluca (approved):

src/coredump/coredump.c
keszybz: Updated.

yuwata (approved):
LGTM.

yuwata:
I opened #21985 to fix one possible assertion when `META_EXE` is NULL.

PR #21977, opened 2022-01-03
hwdb: Allow end-users root-less access to TL866 EPROM readers

hadess:
As is currently done in the upstream minipro tool:
https://gitlab.com/DavidGriffith/minipro/-/tree/master/udev

### DIFF
--- a/hwdb.d/70-analyzers.hwdb
+++ b/hwdb.d/70-analyzers.hwdb
@@ -31,3 +31,14 @@ usb:v1679p3001*
 usb:v1679p6003*
 usb:v0483pDF11*
  ID_SIGNAL_ANALYZER=1
+
+###########################################################
+# XGecu
+###########################################################
+# TL866A/CS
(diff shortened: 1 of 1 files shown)


hadess:
The failure looks like it was caused by a Python update:
```
/home/runner/work/systemd/systemd/hwdb.d/parse_hwdb.py:178: UserWarning: warn_multiple_tokens_in_named_alternation: setting results name 'VALUE' on Or expression will return a list of all parsed tokens in an And alternative, in prior versions only the first token was returned; enclosecontained argument in Group
  ('!' ^ (Optional('!') - Word(alphanums + '_')))('VALUE')
```

@keszybz ?

keszybz:
The change LGTM.

I'll take a stab at fixing the test later.

hadess:
I repushed your changes with my commit last...

hadess:
Can't remember the number of times I've been bitten by that case problem...

PR #21978, opened 2022-01-03
ci: bump mkosi to v12 with libsolv workaround

mrc0mmand:
Replaces: https://github.com/systemd/systemd/pull/21574
Related:
    * https://github.com/systemd/mkosi/issues/861
    * https://github.com/systemd/mkosi/pull/878

### DIFF
--- a/.github/workflows/mkosi.yml
+++ b/.github/workflows/mkosi.yml
@@ -40,7 +40,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
-    - uses: systemd/mkosi@4d64fc8134f93d87ac584183e7762ac1d0efa0e5
+    - uses: systemd/mkosi@01ea953fd2af738f974b228991c768c12b50db95
 
     - name: Install
       run: sudo apt-get update && sudo apt-get install --no-install-recommends python3-pexpect python3-jinja2


mrc0mmand:
It appears to be working! @evverx PTAL (hopefully I didn't break Dependabot by touching the action's SHA)

evverx:
Thanks! I think last time I pointed Dependabot at a commit that didn't refer to a release it broke in the sense that it couldn't bump the action to the next version when it was released but I'm not sure whether it was a bug in Dependabot. Let's find out :-)

evverx:
To judge from https://github.com/systemd/systemd/runs/4691185759?check_suite_focus=true mkosi failed to build openSUSE differently

mrc0mmand:
*It's always something with these distributions.* 

Judging from the zypper's exit code, it should be just an intermittent fail:

```
       106 - ZYPPER_EXIT_INF_REPOS_SKIPPED
           Some repository had to be disabled temporarily because it failed to refresh. You should check your
           repository configuration (e.g. zypper ref -f).
```

caused by:

```
 Building repository 'repo-oss' cache [....done]
File './repodata/f4fba37aa004f93b3b9f72917e4621a69964f902184252d9cfce7f49399cb2fe-primary.xml.gz' not found on medium 'http://download.opensuse.org/update/tumbleweed/'
Abort, retry, ignore? [a/r/i/...? shows all options] (a): a
Error building the cache:
[repo-update|http://download.opensuse.org/update/tumbleweed/] Valid metadata not found at specified URL
Warning: The metadata cache needs to be built for the 'repo-update' repository. You can run 'zypper refresh' as root to do this.
Disabling repository 'repo-update'.
Some of the repositories have not been refreshed because of an error.
```

DaanDeMeyer:
I also found "ERROR: module /lib64/security/pam_pwquality.so is not installed." but not sure if that's fatal or not.

evverx:
> It's always something with these distributions.

To be fair Ubuntu/Debian/Fedora don't seem to fail there very often :-)

I restarted it in https://github.com/systemd/systemd/actions/runs/1649357573 and it seems to have passed so hopefully it was just a glitch.

Werkov:
The zypper download failure has time coincidence with the repo update, given the later success I'd consider it a random failure due to external reasons (I've raised that on openSUSE Factory ML, awating moderation).

The pam error is a red herring ([won't break](https://bugzilla.suse.com/show_bug.cgi?id=1185439#c1) image build really).

As for the "always something" -- rebuilding images from scratch is indeed not ideal (given objective of this CI), I think some caching of action artifacts was considered at one time. Alternatively, install the compiled artifacts into "regular" distro images from registry.

evverx:
> given the later success I'd consider it a random failure due to external reasons 

It popped up at least one more time in https://github.com/systemd/systemd/runs/4692095421?check_suite_focus=true an hour ago though

> rebuilding images from scratch is indeed not ideal (given objective of this CI),

I agree (to some extent). I'd add that relying on the infrastructure provided by distributions seems to kind of prevents systemd from being tested as much as I'd like :-) The latest example would be `boot` where it would be great to build it with linkers different from bfd but all the distributions seem to default to `bfd` so it is what it is :-)

> I think some caching of action artifacts was considered at one time. Alternatively, install the compiled artifacts into "regular" distro images from registry.

I think caches usually create more problems than they solve but I have to admit I have no idea how the action works

PR #21981, opened 2022-01-03
boot: Cleanup

medhefgo:


src/boot/efi/boot.c
```diff
@@ -104,6 +104,11 @@ enum {
         TIMEOUT_TYPE_MAX    = UINT32_MAX,
 };
 
+enum {
+        IDX_MAX = INT16_MAX,
+        IDX_INVALID,
```
bluca, line 110: we usually initialize the _INVALID enum to `-EINVAL`

bluca (approved):

src/boot/efi/boot.c
keszybz: Yes, please do this so that it can _INVALID can be used in strerror().

keszybz: Hmm, so here IDX_INVALID is positive, and the return type is changed to UINTN. Why do this change? It'd seem to make things more confusing.

medhefgo: This isn't userspace, so no `strerror()`. And UINTN (== size_t) is used everywhere for array indices and lengths. Using a signed type is more confusing and inconsistent. It also creates unsound signed/unsigned comparisons that require a cast.

If you want to insist, I can set this to UINTN_MAX (which was the effective invalid value before). Ultimately, the value doesn't matter as it's just an internal sentinel.

keszybz: OK, the enum here is just to give a name to the constant.

PR #21984, opened 2022-01-03
basic/log: allow errno values higher than 255

keszybz:
When the support for "synthetic errno" was added, we started truncating
the errno value to just the least significant byte. This is generally OK,
because errno values are defined up to ~130.

The docs don't really say what the maximum value is. But at least in principle
higher values could be added in the future. So let's stop truncating
the values needlessly.

The kernel (or libbpf?) have an error where they return 524 as an errno
value (https://bugzilla.redhat.com/show_bug.cgi?id=2036145). We would
confusingly truncate this to 12 (ENOMEM). It seems much nicer to let
strerror() give us "Unknown error 524" rather than to print the bogus
message about ENOMEM.

### DIFF
--- a/src/basic/log.h
+++ b/src/basic/log.h
@@ -31,10 +31,10 @@ typedef enum LogTarget{
  * used a regular log level. */
 #define LOG_NULL (LOG_EMERG - 1)
 
-/* Note to readers: << and >> have lower precedence than & and | */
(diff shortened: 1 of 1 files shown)


yuwata (approved):
LGTM.

bluca (approved):

PR #21985, opened 2022-01-03
elf-util: several cleanups

yuwata:


yuwata:
cc @bluca.

src/shared/elf-util.c
```diff
@@ -198,30 +197,33 @@ static int frame_callback(Dwfl_Frame *frame, void *userdata) {
...
-
-                                                if (symbol)
-                                                        break;
-                                        }
-                                }
+                        for (int i = 0; i < n; i++) {
+                                Dwarf_Die *s = &scopes[i];
+                                Dwarf_Attribute *a, space;
+
+                                if (!IN_SET(sym_dwarf_tag(s), DW_TAG_subprogram, DW_TAG_inlined_subroutine, DW_TAG_entry_point))
```
bluca: `s` can be null

src/shared/elf-util.c
```diff
@@ -624,8 +624,8 @@ static int parse_elf(int fd, const char *executable, char **ret, JsonVariant **r
                 if (r < 0)
                         return log_warning_errno(r, "Failed to parse package metadata of ELF file: %m");
 
-                /* If we found a build-id and nothing else, return at least that. */
-                if (!package_metadata && id_json) {
```
bluca, line 624: I don't think this is right? It's possible nothing is found (it's not an error condition)

src/shared/elf-util.c
```diff
@@ -430,11 +431,9 @@ static int module_callback(Dwfl_Module *mod, void **userdata, const char *name,
         Elf *elf;
 
         assert(mod);
+        assert(name);
```
bluca: this is a callback, and `name` is not guaranteed to be set by libdwfl as far as I understand

src/shared/elf-util.c
```diff
@@ -388,6 +388,7 @@ static int parse_buildid(Dwfl_Module *mod, Elf *elf, const char *name, StackCont
         int r;
 
         assert(mod || elf);
+        assert(name);
```
bluca, line 391: same here, if this comes from the other callback, it might not be set (so changing the use in the log messages to `strna` should be the right fix)

src/shared/elf-util.c
yuwata: Here, `package_metadata` is always NULL. I just dropped the redundant condition.

src/shared/elf-util.c
yuwata: `parse_buildid()` is called from `parse_elf()` and `module_callback()`, and both sets non-NULL `name` now.

src/shared/elf-util.c
bluca: It's not always null? Otherwise `parse_package_metadata` would be doing nothing

src/shared/elf-util.c
bluca: It can still be NULL when it comes from parse_core

yuwata: Ah, `package_metadata` is also passed to StackContext... Ouch.

src/shared/elf-util.c
yuwata: Ah, `package_metadata` is also passed to StackContext... Ouch.

src/shared/elf-util.c
yuwata: You mean parse_core -> module_callback -> parse_buildid ??

bluca: Yes, because it's called from libdwl, not directly, so we don't know for sure

yuwata: But `module_callback()` sets `name` if it is NULL... (I already dropped the change in `module_callback()`)

bluca (approved):

src/shared/elf-util.c
bluca: ah yes, of course

Issue #21983, opened 2022-01-03
random-seed.c should be mindful of reseeding states


zx2c4:
At boot, systemd will read the existing seed file and call `RNDADDENTROPY` on it. It will then extract a new seed and write that to disk. What happens between adding that entropy and extracting a new seed?

- Case A) If you call `RNDADDENTROPY` at `crng_init<2`, and there's enough entropy in your call, it'll switch to `crng_init==2` and reseed everything. The subsequent extraction will then be fine.
- Case B) If you call `RNDADDENTROPY` and you're already at `crng_init==2`, then it won't reseed everything. You could call `RNDRESEEDCRNG` after if you wanted it to have an immediate effect, but it's not super important, since there's already sufficient entropy from before. The subsequent extraction is fine.
- Case C) If you call `RNDADDENTROPY` at `crng_init<2`, and there's **not** enough entropy in your call, it'll just mix it into the pool, but nothing will happen immediately. The subsequent extraction will then use potentially crappy randomness.

Case C might be a bit of a problem. Stemming out of a conversation with @thejh, we were thinking:
- For newer kernels, [this pending commit](https://git.kernel.org/pub/scm/linux/kernel/git/crng/random.git/commit/?id=b1fde14b931d941abfaa91a7ac9d72b2f5c834ab) will mean that calling `RNDRESEEDCRNG` before you extract a new seed to write to disk will mostly do the rightish thing for the `crng_init<2` case.
- For older kernels, it's not a great situation, but you could at least hash together the old seed file and the new seed file before writing it out to disk.

Thoughts on this?

CC @poettering

zx2c4:
Looking at the systemd code a bit now, following the discussion with Jann, I'm not totally sure this matters (or even that the linked pending commit is totally necessary).

- For random-seed.c's "load" mode, there's a synchronous call to `getrandom()` before writing out a new seed file to disk. That means we're in "case B" territory, and things are fine. 
- If the kernel doesn't support `getrandom()`, that pending commit doesn't help anyway. The best we can do in that case is hash it together with the old seed; that still might be a good suggestion for these cases.
- For random-seed.c's "save" mode, it may well fall back to insecure /dev/urandom if called too early, but the concern we should have here is mostly about "load" mode, right? "save" should only be called at shutdown, correct?

I'll wait for @thejh to chime in if he sees things differently.

zx2c4:
[...]

PR #21986, opened 2022-01-03
random-seed: hash together old seed and new seed before writing out file

zx2c4:
If we're consuming an on-disk seed, we usually write out a new one after consuming it. In that case, we might be at early boot and the randomness could be rather poor, and the kernel doesn't guarantee that it'll use the new randomness right away for us. In order to prevent the new entropy from getting any worse, hash together the old seed and the new seed, and replace the final bytes of the new seed with the hash output. This way, entropy strictly increases and never regresses.

Fixes: https://github.com/systemd/systemd/issues/21983

### DIFF
--- a/src/random-seed/random-seed.c
+++ b/src/random-seed/random-seed.c
@@ -26,6 +26,7 @@
 #include "random-util.h"
 #include "string-util.h"
 #include "sync-util.h"
+#include "sha256.h"
 #include "util.h"
 #include "xattr-util.h"
 
@@ -106,9 +107,11 @@ static int run(int argc, char *argv[]) {
         _cleanup_close_ int seed_fd = -1, random_fd = -1;
         bool read_seed_file, write_seed_file, synchronous;
         _cleanup_free_ void* buf = NULL;
+        struct sha256_ctx hash_state;
+        uint8_t hash[32];
         size_t buf_size;
         struct stat st;
-        ssize_t k;
+        ssize_t k, l;
         int r;
 
         log_setup();
@@ -242,6 +245,16 @@ static int run(int argc, char *argv[]) {
                         if (r < 0)
                                 log_error_errno(r, "Failed to write seed to /dev/urandom: %m");
                 }
+                /* If we're going to later write out a seed file, initialize a hash state with
+                 * the contents of the seed file we just read, so that the new one can't regress
+                 * in entropy. */
+                if (write_seed_file) {
+                        sha256_init_ctx(&hash_state);
+                        if (k < 0)
+                                k = 0;
+                        sha256_process_bytes(&k, sizeof(k), &hash_state);
+                        sha256_process_bytes(buf, k, &hash_state);
+                }
         }
 
         if (write_seed_file) {
@@ -275,6 +288,17 @@ static int run(int argc, char *argv[]) {
                         if (k == 0)
                                 return log_error_errno(SYNTHETIC_ERRNO(EIO),
                                                        "Got EOF while reading from /dev/urandom.");
+                }
+
+                /* If we previously read in a seed file, then hash the new seed into the old one,
+                 * and replace the last 32 bytes of the seed with the hash output, so that the
+                 * new seed file can't regress in entropy. */
+                if (read_seed_file) {
+                        sha256_process_bytes(&k, sizeof(k), &hash_state);
+                        sha256_process_bytes(buf, k, &hash_state);
+                        sha256_finish_ctx(&hash_state, hash);
+                        l = MIN(k, 32);
+                        memcpy((uint8_t *)buf + k - l, hash, l);
                 }
 
                 r = loop_write(seed_fd, buf, (size_t) k, false);


src/random-seed/random-seed.c
```diff
@@ -242,6 +245,11 @@ static int run(int argc, char *argv[]) {
                         if (r < 0)
                                 log_error_errno(r, "Failed to write seed to /dev/urandom: %m");
                 }
+                if (write_seed_file) {
+                        sha256_init_ctx(&hash_state);
+                        sha256_process_bytes(&buf_size, sizeof(buf_size), &hash_state);
+                        sha256_process_bytes(buf, buf_size, &hash_state);
```
thejh: shouldn't these two lines use `k` (read result) instead of `buf_size` (allocated size)?

zx2c4: Yes indeed, thanks.

src/random-seed/random-seed.c
```diff
@@ -242,6 +245,13 @@ static int run(int argc, char *argv[]) {
                         if (r < 0)
                                 log_error_errno(r, "Failed to write seed to /dev/urandom: %m");
                 }
+                if (write_seed_file) {
```
bluca, line 251: Please add a comment in the code too, so that we don't have to go back to the commit message to find out why this is being done. A short one is fine, just to give a hint.

bluca (approved):

src/random-seed/random-seed.c
zx2c4: Will do.

zx2c4:
Requested comment added. Should be good to go.

src/random-seed/random-seed.c
```diff
@@ -106,9 +107,11 @@ static int run(int argc, char *argv[]) {
         _cleanup_close_ int seed_fd = -1, random_fd = -1;
         bool read_seed_file, write_seed_file, synchronous;
         _cleanup_free_ void* buf = NULL;
+        struct sha256_ctx hash_state;
+        uint8_t hash[32];
```
poettering, line 111: move to innermost scope?

src/random-seed/random-seed.c
```diff
@@ -242,6 +245,16 @@ static int run(int argc, char *argv[]) {
...
+                /* If we're going to later write out a seed file, initialize a hash state with
+                 * the contents of the seed file we just read, so that the new one can't regress
+                 * in entropy. */
+                if (write_seed_file) {
+                        sha256_init_ctx(&hash_state);
+                        if (k < 0)
+                                k = 0;
+                        sha256_process_bytes(&k, sizeof(k), &hash_state);
+                        sha256_process_bytes(buf, k, &hash_state);
+                }
```
poettering, line 257: this if block is unnecessary unless we actually managed to read a seed off disk, no? hence, move this into the if block immediately preceeding this, i.e. after the successful random_write_entropy()? if so, you can drop the `if (k < 0) k = 0)` thing, because that's guaranteed then.

src/random-seed/random-seed.c
```diff
@@ -242,6 +245,16 @@ static int run(int argc, char *argv[]) {
...
         }
```
poettering, line 258: what's the rationale of including the size of the seed in the hash?

I presume there's a good reason for that. Can you add a comment about it to the line?

src/random-seed/random-seed.c
zx2c4: `hash_state` is already innermost, because it gets updated in two scopes later. But `hash` could be moved to an innerscope.

src/random-seed/random-seed.c
```diff
@@ -277,6 +290,17 @@ static int run(int argc, char *argv[]) {
...
 
+                /* If we previously read in a seed file, then hash the new seed into the old one,
+                 * and replace the last 32 bytes of the seed with the hash output, so that the
+                 * new seed file can't regress in entropy. */
+                if (read_seed_file) {
+                        sha256_process_bytes(&k, sizeof(k), &hash_state);
+                        sha256_process_bytes(buf, k, &hash_state);
+                        sha256_finish_ctx(&hash_state, hash);
+                        l = MIN(k, 32);
+                        memcpy((uint8_t *)buf + k - l, hash, l);
```
poettering, line 301: doesn't matter, but if you'd hash this into the beginning of the buffer, you don't need a temporary variable, i.e.:

```c
memcpy(buf, hash, MIN(32, k));
```

And whether the SHA256 result is placed at the end or the beginning of the seed buffer doesn't really matter, does it?

src/random-seed/random-seed.c
zx2c4: I figured it was cleanest to unconditionally initialize the hash state here, so that regardless of the outcome it'd be in a good state later on down when it's used again. Instead, I guess more code (really, another boolean) could be added to determine whether there is an old seed hashed or not, if you really wanted. Doesn't seem to get us anything though?

poettering: i must say I am typically not a fan of such "implicit dependencies". people move code around, and it is not obvious under which circumstances `hash_state` is initialized or not, hence this might get out of sync by accident, in some error case or so. The whole function is unusually large already after all, hence such implicit deps are not easy to see. Hence usually i'd add an explicit bool here, `hash_state_initialized` or so, and then always set/check that. the compiler is then free to optimize it away if it wants to, but given this is not a hot path, it doesn't really matter... 

i mean, this doesn't matter much i guess, it's just something i noticed. at the very least could use a comment, so that this isn't changed by accident so easily.

src/random-seed/random-seed.c
zx2c4: The idea is just to distinguish "bake race" from "baker ace" type of thing. Probably not vitally important here but seems hygienic to do.

src/random-seed/random-seed.c
zx2c4: It doesn't really matter at all here I suppose. Were this a different sort of thing, maybe we'd have some handwavy concerns about hiding the last bytes that were hashed rather than the first, to make length extension or other attacks harder. But that's very much not what's happening here, so I guess we could do the beginning if you feel strongly about it.

zx2c4:
@poettering luca already merged this. Do you care enough about the feedback here that you want me to submit a new PR, or does it not matter?

src/random-seed/random-seed.c
zx2c4: Will do.

PR #21994, opened 2022-01-04
fuzz: no longer skip empty files

evverx:
Empty files and empty strings seem to have triggered various
issues in the past so it seems they shouldn't be ignored by the
fuzzers just because fmemopen can't handle them.

Prompted by https://github.com/systemd/systemd/pull/21939#issuecomment-1003113669

@mrc0mmand @yuwata could you take a look?

### DIFF
--- a/src/core/fuzz-unit-file.c
+++ b/src/core/fuzz-unit-file.c
@@ -2,7 +2,6 @@
 
 #include "conf-parser.h"
 #include "fd-util.h"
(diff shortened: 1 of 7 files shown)


keszybz:
LGTM.

mrc0mmand (commented):
Belated LGTM, thanks!

yuwata (commented):
LGTM.

PR #21996, opened 2022-01-04
fuzz-bcd: silence warning about always-true comparison

keszybz:
Occurs with gcc-11.2.1-7.fc35.x86_64.

### DIFF
--- a/src/boot/efi/fuzz-bcd.c
+++ b/src/boot/efi/fuzz-bcd.c
@@ -21,6 +21,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         assert_se(p);
 
         char16_t *title = get_bcd_title(p, size);
-        assert_se(!title || char16_strlen(title) >= 0);
+        if (title)
+                (void) char16_strlen(title);
         return 0;
 }


yuwata (approved):

evverx:
Thanks! What's weird is that I used that compiler and missed that somehow. The build workflow was able to find it in https://github.com/systemd/systemd/runs/4697002693?check_suite_focus=true but it didn't fail even though `-Werror` is passed there. I'm puzzled.

mrc0mmand:
> Thanks! What's weird is that I used that compiler and missed that somehow. The build workflow was able to find it in https://github.com/systemd/systemd/runs/4697002693?check_suite_focus=true but it didn't fail even though `-Werror` is passed there. I'm puzzled.

The `-Werror` flag doesn't seem to be passed to the fuzzer build:

```

  User defined options
    optimization                   : s
    werror                         : True
    cryptolib                      : openssl
    fuzz-tests                     : true
    slow-tests                     : true
    tests                          : unsafe

Found ninja-1.10.2.git.kitware.jobserver-1 at /home/runner/.local/bin/ninja
+ meson compile -C build
...
```

```
...
  User defined options
    optimization                   : s
    b_lundef                       : false
    b_sanitize                     : address,undefined
    fuzz-tests                     : true

Found ninja-1.10.2.git.kitware.jobserver-1 at /home/runner/.local/bin/ninja
+ ninja=ninja
+ which ninja-build
+ ninja -C test/fuzz/sanitize-address-undefined-fuzzers fuzzers
ninja: Entering directory `test/fuzz/sanitize-address-undefined-fuzzers'
```

evverx:
> The -Werror flag doesn't seem to be passed to the fuzzer build

Looks like it doesn't but it probably should be passed on because with
```diff
diff --git a/test/fuzz/meson.build b/test/fuzz/meson.build
index 30e26b09cf..5af2dcd562 100644
--- a/test/fuzz/meson.build
+++ b/test/fuzz/meson.build
@@ -7,7 +7,7 @@ sanitize_address_undefined = custom_target(
                    project_source_root,
                    '@OUTPUT@',
                    'fuzzers',
-                   '-Dfuzz-tests=true -Db_lundef=false -Db_sanitize=address,undefined --optimization=@0@'.format(get_option('optimization')),
+                   '-Dfuzz-tests=true -Db_lundef=false -Db_sanitize=address,undefined --werror --optimization=@0@'.format(get_option('optimization')),
                    ' '.join(cc.cmd_array()),
                    cxx_cmd])
```
I can no longer build systemd without this PR.

mrc0mmand:
Agreed, I'll prep a PR.

PR #21999, opened 2022-01-04
homed: stop before stopping dbus

gekmihesg:
Otherwise, systemd-homed-active.service will fail to deactivate all
homes because homectl can no longer talk to homed if dbus stops first.
As a result, /home cannot be umounted.

Doing this on systemd-homed-active.service instead works as well, but
systemd-homed will exit 1 if dbus is already shut down.

### DIFF
--- a/units/systemd-homed.service.in
+++ b/units/systemd-homed.service.in
@@ -12,7 +12,7 @@ Description=Home Area Manager
 Documentation=man:systemd-homed.service(8)
 Documentation=man:org.freedesktop.home1(5)
 
-After=home.mount
+After=home.mount dbus.service
 
 [Service]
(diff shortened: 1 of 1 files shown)


bluca (approved):
Seems reasonable to me. @keszybz ?

keszybz:
Yeah, I think so. I was wondering if this should be After=dbus.socket or dbus.service. But .service seems better, because we want to avoid a scenario where dbus.service is stopped during shutdown and we try to retrigger it by accessing the socket.

gekmihesg:
Thanks! dbus.socket was what I tried first, but it didn't work, homectl deactivate-all had the same connection refused as without any changes.

Issue #22001, opened 2022-01-04
test-watchdog causes system reboot


floppym:
**systemd version the issue has been seen with**
948d085e899273f3a34376665414d1ecd1858978

**Used distribution**
Gentoo

**Linux kernel version used** (`uname -a`)
 `Linux naomi 5.15.12-dist #1 SMP Fri Dec 31 12:43:34 EST 2021 x86_64 AMD Phenom(tm) II X6 1055T Processor AuthenticAMD GNU/Linux`

**CPU architecture issue was seen on**
x86_64

**Expected behaviour you didn't see**
Successful test run.

**Unexpected behaviour you saw**
System immediately reboots.

**Steps to reproduce the problem**
Run `test-watchdog` as root.

floppym:
Note that `test-watchdog` is run automatically by `meson test`, which might not be a great idea.

bluca:
That's weird - what's in the log, with debug level? Sounds like watchdog pinging is not working

floppym:
> what's in the log, with debug level?

The kernel reboots the system before any logs can be stored.
In a terminal, I see `Pinging...` appear twice before the reboot happens.

floppym:
When I set SYSTEMD_SLOW_TESTS=true, the test works fine.

Maybe the watchdog device on this system can't handle the 1 second timeout that the test uses by default.
```
# SYSTEMD_SLOW_TESTS=true ./test-watchdog
Using hardware watchdog 'SP5100 TCO timer', version 0, device /dev/watchdog0
Watchdog running with a timeout of 10s.
Pinging...
Pinging...
Pinging...
Pinging...
Pinging...
```

evverx:
@floppym I wonder if the test passes with v249 without SYSTEMD_SLOW_TESTS=true? If it does it's probably a regression caused by https://github.com/systemd/systemd/pull/20650.

floppym:
I can give it a try with v249 tomorrow.

evverx:
On the other hand, considering that the test seems to have printed "Pinging..." without "Failed to ping watchdog" it could be that the test itself can't handle scenarios where 1s isn't supported because if the timeout is actually set to something else the following loop
```c
        for (i = 0; i < count; i++) {
                log_info("Pinging...");
                r = watchdog_ping();
                if (r < 0)
                        log_warning_errno(r, "Failed to ping watchdog: %m");

                usleep(t/2);
```
seems to be wrong.

evverx:
[...]

PR #22009, opened 2022-01-05
test-watchdog: mark as unsafe

floppym:
If something goes wrong with this test it may result in an unsafe system restart. Let's avoid running it automatically.

See https://github.com/systemd/systemd/issues/22001.

### DIFF
--- a/src/test/meson.build
+++ b/src/test/meson.build
@@ -552,7 +552,8 @@ tests += [
          [],
          core_includes, '', 'manual'],
 
-        [['src/test/test-watchdog.c']],
+        [['src/test/test-watchdog.c'],
+         [], [], [], '', 'unsafe'],
 
         [['src/test/test-sched-prio.c'],
          [libcore,


evverx:
The test is certainly unsafe and shouldn't be run by default but before merging the second commit it would be great to figure out whether it can be triggered with v249.

I think at some point something like `-watchdog i6300esb` should be passed to `qemu` in `test-functions` to really exercise that code with the test. Though it's unlikely that it could have caught https://github.com/systemd/systemd/issues/22001 because 1s is supported by that watchdog as far as I can see.

floppym:
I dropped the second commit for now.

evverx (approved):
LGTM. Thanks!

mrc0mmand:
> @mrc0mmand Looking at https://pagure.io/centos-infra/issue/505 it seems it's run on CentOS CI. Just to make sure it keeps running, are unsafe tests run there too or should something like `-Dtests=unsafe` be additionally passed there?

All CentOS CI jobs run with `-Dslow-tests=true -Dfuzz-tests=true -Dtests=unsafe`, so it shouldn't have any effect there.

PR #22013, opened 2022-01-05
fstab-generator: skip root directory handling when nfsroot is requested

yuwata:
Fixes RHBZ#2037233 (https://bugzilla.redhat.com/show_bug.cgi?id=2037233).


yuwata:
(I have not test this yet.)

src/fstab-generator/fstab-generator.c
```diff
@@ -714,6 +714,12 @@ static int add_sysroot_mount(void) {
                 return 0;
         }
 
+        if (startswith(arg_root_what, "nfs:")) {
```
keszybz: The dracut docs say:

       root=[<server-ip>:]<root-dir>[:<nfs-options>]
           mount nfs share from <server-ip>:/<root-dir>, if no server-ip
           is given, use dhcp next_server. If server-ip is an IPv6
           address it has to be put in brackets, e.g. [2001:DB8::1]. NFS
           options can be appended with the prefix ":" or "," and are
           separated by ",".

       root=nfs:[<server-ip>:]<root-dir>[:<nfs-options>],
       root=nfs4:[<server-ip>:]<root-dir>[:<nfs-options>],
       root={dhcp|dhcp6}

I think we'd want to support all of those. Checking for "nsf:", "nsf4:", "dhcp", "dhcp6" should be easy.
The case where `server-ip` is given with no 'nfs/nfs4' prefix could be detected by trying to parse the IP.
Those docs say that specifying just the `root-dir` alone is also allowed, but I think that this case is impossible to ignore for a normal device path, so we should just ignore it.

src/fstab-generator/fstab-generator.c
```diff
@@ -714,6 +714,12 @@ static int add_sysroot_mount(void) {
...
+                /* This is handled by the kernel or the initrd */
+                log_debug("Skipping root directory handling, as nfsroot was requested.");
```
keszybz: Maybe say "root on NFS"? "nfsroot" is a bit jargony.

keszybz (changes requested):

src/fstab-generator/fstab-generator.c
yuwata: > Those docs say that specifying just the `root-dir` alone is also allowed, but I think that this case is impossible to ignore for a normal device path, so we should just ignore it.

Is it possible to specify non device path without requesting nfsroot, e.g. /foo/bar? I guess we can assume that non device path is NFS root.

src/fstab-generator/fstab-generator.c
```diff
@@ -691,6 +692,63 @@ static int parse_fstab(bool initrd) {
...
+
+                a = strndupa(arg_root_what, sep - arg_root_what - 1);
+
+                if (in_addr_from_string(AF_INET, a, &u) >= 0)
+                        return true;
+        }
+
+        /* root directory without address */
+        return path_is_absolute(arg_root_what) && !path_startswith(arg_root_what, "/dev");
+}
```
keszybz, line 744: This looks very nice… Can you add a test though?

keszybz: (… a unittest for the function itself, nothing too fancy.)

yuwata:
@keszybz Test cases for fstab-generator are added. Upgrading the green label.

keszybz:
focal-ppc64el:
```
[  131.168953] systemd-udevd[424]: null: Worker [427] processing SEQNUM=2977 is taking a long time
[  131.273329] systemd-udevd[427]: null: Spawned process '/bin/sleep 60' [433] is taking longer than 3s to complete
[  138.110264] systemd-udevd[424]: null: Worker [427] processing SEQNUM=2977 killed
[  144.375659] kernel: ext4 filesystem being remounted at /run/systemd/unit-root/var/lib/systemd/coredump supports timestamps until 2038 (0x7fffffff)
[  144.391352] kernel: ext4 filesystem being remounted at /run/systemd/unit-root/tmp supports timestamps until 2038 (0x7fffffff)
[  144.405458] kernel: ext4 filesystem being remounted at /run/systemd/unit-root/var/tmp supports timestamps until 2038 (0x7fffffff)
[  145.219588] kernel: ext4 filesystem being remounted at /run/systemd/unit-root/tmp supports timestamps until 2038 (0x7fffffff)
[  145.288116] kernel: ext4 filesystem being remounted at /run/systemd/unit-root/var/lib/systemd/coredump supports timestamps until 2038 (0x7fffffff)
[  145.324500] kernel: ext4 filesystem being remounted at /run/systemd/unit-root/var/tmp supports timestamps until 2038 (0x7fffffff)
[  147.887683] )[478]: Could not parse core file, dwfl_core_file_attach() failed: Unwinding not supported for this architecture
[  148.061296] )[478]: Failed to inspect core file: Invalid argument
[  148.223642] systemd-coredump[465]: Process 427 (systemd-udevd) of user 0 dumped core.
[  150.257913] systemd-udevd[424]: Worker [427] terminated by signal 6 (ABRT)
[  150.309583] systemd-udevd[424]: null: Worker [427] failed
[  156.878940] systemd[1]: testsuite-17.service: Failed with result 'exit-code'.
[  156.994132] systemd[1]: Failed to start TEST-17-UDEV.
TEST-17-UDEV: (failed; see logs)
```
Hmm, this looks a bit strange. 

focal-arm64 failed in TEST-02-UNITTESTS, but I don't see why. Hopefully both failures are unrelated.

keszybz:
LGTM too.

keszybz:
Ubuntuautopkgci is very slow today… I'll merge without waiting for it to finish. What failed seems unrelated and I also tested this locally.

medhefgo:
> Ubuntuautopkgci is very slow today… I'll merge without waiting for it to finish. What failed seems unrelated and I also tested this locally.

Might've been premature, the fstab generator test seems to fail all over the place.

evverx:
> Might've been premature, the fstab generator test seems to fail all over the place.

While I agree that probably it shouldn't have been merged I agree that Ubuntu CI is kind of annoying. I "fixed" it in https://github.com/systemd/systemd/pull/22053.

evverx:
> focal-ppc64el:

That is a known issues and was reported in https://github.com/systemd/systemd/issues/21603

PR #22014, opened 2022-01-05
Drop detailed error messages for netlink message append operations

keszybz:
See the first commit for detailed justification and the next-to-last for final stats.


yuwata (approved):
LGTM. Thanks. I planned to downgrade the log level of error 'Failed to append XXX attribute: %m' as the error is mostly programming error, that is, type mismatch or type is not defined. But, as you mentioned in the commit message, if it massively decrease the binary size, then it is OK to drop the log message.

PR #22015, opened 2022-01-05
test: use full date & time when checking for coredumps

mrc0mmand:
Otherwise we might hit a window where the coredump happens before
midnight, but we check for it after midnight, which yields no results.

E.g.:

```
$ coredumpctl --no-legend --no-pager --file system.journal
Wed 2022-01-05 01:00:06 CET 359 0 0 SIGABRT journal /usr/bin/udevadm n/a
$ coredumpctl --since 23:59:55 --no-legend --no-pager --file system.journal
No coredumps found.
$ coredumpctl --since "2022-01-04 23:59:59" --no-legend --no-pager --file system.journal
Wed 2022-01-05 01:00:06 CET 359 0 0 SIGABRT journal /usr/bin/udevadm n/a
```

*Based on a true story.*

### DIFF
--- a/test/units/testsuite-17.03.sh
+++ b/test/units/testsuite-17.03.sh
@@ -26,7 +26,7 @@ teardown() {
 }
 
 run_test() {
-    since="$(date +%T)"
+    since="$(date '+%F %T')"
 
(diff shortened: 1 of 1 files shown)


keszybz:
> _Based on a true story._

Queue *appropriate music*

keszybz:
LGTM.

PR #22019, opened 2022-01-05
shutdown command fixes

lnussel:
Fixes embarrassing c&p error in the shutdown --show implementation. Mea culpa. :face_with_spiral_eyes:

### DIFF
--- a/src/systemctl/systemctl-logind.c
+++ b/src/systemctl/systemctl-logind.c
@@ -297,7 +297,6 @@ int logind_schedule_shutdown(void) {
 #if ENABLE_LOGIND
         _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
         const char *action;
-        const char *log_action;
         sd_bus *bus;
         int r;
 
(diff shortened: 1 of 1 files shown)


src/systemctl/systemctl-logind.c
```diff
@@ -339,9 +330,7 @@ int logind_schedule_shutdown(void) {
                 return log_warning_errno(r, "Failed to call ScheduleShutdown in logind, proceeding with immediate shutdown: %s", bus_error_message(&error, r));
 
         if (!arg_quiet)
-                log_info("%s scheduled for %s, use 'shutdown -c' to cancel.",
-                         log_action,
-                         FORMAT_TIMESTAMP_STYLE(arg_when, arg_timestamp_style));
+                logind_show_shutdown();
```
yuwata, line 336: I do not think it is not necessary to retrieve the `ScheduledShutdown` property here.
How about to introduce tiny helper `log_scheduled_shutdown()` instead?

yuwata (changes requested):

src/systemctl/systemctl-logind.c
```diff
@@ -308,24 +307,16 @@ int logind_schedule_shutdown(void) {
...
                 action = "poweroff";
-                log_action = "Shutdown";
                 break;
         case ACTION_KEXEC:
                 action = "kexec";
-                log_action = "Reboot via kexec";
-                break;
-        case ACTION_EXIT:
-                action = "exit";
-                log_action = "Shutdown";
```
yuwata, line 323: Maybe, it is better to drop `ACTION_EXIT` in a separate commit.

src/systemctl/systemctl-logind.c
lnussel: reading back the property makes sure we actually print the server side view though.

src/systemctl/systemctl-logind.c
lnussel: Right. I'll put it back in for now instead. There's more cleanup to do in that area

src/systemctl/systemctl-logind.c
keszybz: Yeah, I think this makes things simpler and easier.

PR #22021, opened 2022-01-05
bpf: enable bpf lsm only on x86 and x86_64

jkartseva:
BPF_LSM_MAC program requires BPF trampoline [0] which is supported
only on x86 and x86_64 architectures.

[0] https://lwn.net/Articles/804937/

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2036145

### DIFF
--- a/src/core/bpf-lsm.c
+++ b/src/core/bpf-lsm.c
@@ -10,6 +10,7 @@
 #include <unistd.h>
 
 #include "alloc-util.h"
+#include "architecture.h"
 #include "bpf-lsm.h"
 #include "cgroup-util.h"
 #include "fd-util.h"
@@ -127,10 +128,21 @@ static int mac_bpf_use(void) {
 int lsm_bpf_supported(void) {
         _cleanup_(restrict_fs_bpf_freep) struct restrict_fs_bpf *obj = NULL;
         static int supported = -1;
-        int r;
+        int a, r;
 
         if (supported >= 0)
                 return supported;
+
+        a = uname_architecture();
+        if (a < 0) {
+                log_info_errno(a, "Can't determine CPU architecture, LSM BPF is not supported: %m");
+                return supported = 0;
+        }
+        if (a != ARCHITECTURE_X86 && a != ARCHITECTURE_X86_64) {
+                log_info_errno(SYNTHETIC_ERRNO(EOPNOTSUPP),
+                               "LSM BPF is supported only on x86 and x86_64 architectures");
+                return supported = 0;
+        }
 
         r = dlopen_bpf();
         if (r < 0) {


bluca (approved):

bluca:
Separately from this issue, the bug report shows:

```
[   10.699809] systemd[1]: Failed to link 'restrict_filesystems' LSM BPF program: Cannot allocate memory
[   10.736605] systemd[1]: Failed to allocate manager object: Cannot allocate memory                
[   10.739118] systemd[1]: Freezing execution.
```

When something goes wrong with the LSM BPF settings we shouldn't fail that hard and freeze the system. Could you please look into that as well? Thanks!

wat-ze-hex:
@bluca



> Separately from this issue, the bug report shows:
> 
> ```
> [   10.699809] systemd[1]: Failed to link 'restrict_filesystems' LSM BPF program: Cannot allocate memory
> [   10.736605] systemd[1]: Failed to allocate manager object: Cannot allocate memory                
> [   10.739118] systemd[1]: Freezing execution.
> ```
`ENOMEM` is misleading, that was already addressed by @keszybz in https://github.com/systemd/systemd/pull/21984
> 
> When something goes wrong with the LSM BPF settings we shouldn't fail that hard and freeze the system. Could you please look into that as well? Thanks!

That is happening because of the following code in core manager:
```
r = lsm_bpf_setup(m);
if (r < 0)
       return r;
```
https://github.com/systemd/systemd/blob/main/src/core/manager.c#L934L936
It seems that instead we should print warning and proceed.

The reason why it was put in manager is that BPF LSM is cgroup unaware and requires global setup.
I'll address in a separate PR.

evverx:
I wonder if it should be done at runtime? As far as I see architecture-specific code like that is usually compiled out when systemd is built.

evverx:
> Is this true for cross-compilation (target arch is different form host)?

I think it should be. `__i386__` and `__x86_64__` should be set correctly by compilers when those are target architectures.

FWIW I'm not saying that it shouldn't be done at runtime. It's just that judging by
```
$ git grep uname_architecture
docs/PORTING_TO_NEW_ARCHITECTURES.md:   `native_architecture()` and `uname_architecture()`.
src/basic/architecture.c:int uname_architecture(void) {
src/basic/architecture.h:int uname_architecture(void);
src/core/dbus-manager.c:static BUS_DEFINE_PROPERTY_GET_GLOBAL(property_get_architecture, "s", architecture_to_string(uname_architecture()));
src/core/main.c:                log_info("Detected architecture %s.", architecture_to_string(uname_architecture()));
src/shared/condition.c:        a = uname_architecture();
src/shared/specifier.c:        t = strdup(architecture_to_string(uname_architecture()));
src/test/test-architecture.c:        a = uname_architecture();
src/test/test-condition.c:        a = uname_architecture();
src/udev/udev-rules.c:                        val = architecture_to_string(uname_architecture());
```
`uname_architecture` isn't usually used for that.

src/core/bpf-lsm.c
```diff
@@ -127,11 +128,22 @@ static int mac_bpf_use(void) {
...
         static int supported = -1;
-        int r;
+        int a, r;
 
         if (supported >= 0)
                 return supported;
 
+        a = uname_architecture();
+        if (a < 0) {
+                log_info_errno(a, "Can't determine CPU architecture, LSM BPF is not supported: %m");
```
yuwata, line 138: I prefer "assuming LSM BPF is not supported".

src/core/bpf-lsm.c
```diff
@@ -127,11 +128,22 @@ static int mac_bpf_use(void) {
...
+                return supported = 0;
+        }
+        if (a != ARCHITECTURE_X86 && a != ARCHITECTURE_X86_64) {
```
yuwata, line 141: Please use `IN_SET()` macro.

yuwata (approved):
LGTM. Two minor suggestions.

bluca:
A runtime check seems the right thing to me, we are checking a property of the running kernel, not of the binary architecture at build time. Ideally there would be a better way to check for the feature support at runtime, rather than hard coding a list of archs, but I don't think there is

bluca:
I do wonder if https://github.com/systemd/systemd/pull/22025 by itself is better as a fix? That way as more architectures are added to the kernel (I believe at least arm64 is in the works?) we get seamless support without having to change the code and rebuild here?

wat-ze-hex:
> I do wonder if #22025 by itself is better as a fix? That way as more architectures are added to the kernel (I believe at least arm64 is in the works?) we get seamless support without having to change the code and rebuild here?

I agree, hard-coded list is not flexible.

evverx:
> we are checking a property of the running kernel, not of the binary architecture at build time

If targets architectures don't match runtime architectures I'm pretty sure systemd won't run at all. There are exceptions like compiling systemd with -m32 on x86_64 machines but in that case the build time check actually works better because binaries compiled with `-m32` would still receive `x86_64` from `uname` on x86_64 machines with `__i386__` defined correctly.

> That way as more architectures are added to the kernel (I believe at least arm64 is in the works?) we get seamless support without having to change the code and rebuild here?

I don't think it's a good idea (at this point at least) because any attempts to load BPF skeletons on some architectures can still crash systemd: https://github.com/libbpf/libbpf/issues/391 (that bug was fixed but there are still issues caused by libelf). I think that if it's known that this works on two architectures only it should be compiled and run on two architectures only. I don't think it's hard to update the code once arm is supported.

bluca:
Nah, we support things like binfmt with nspawn&friends. A runtime check is the right thing here.

evverx:
> A runtime check is the right thing here.

If it is I wonder why it isn't used for anything like that in the systemd codebase?

evverx:
Anyway, in this particular case I don't think it matters how it's done. But I still think that that part should be run on i386/x86_64 only.

bluca:
> > A runtime check is the right thing here.
> 
> If it is I wonder why it isn't used for anything like that in the systemd codebase?

Uh? Checks like those in #22025 are done all over the place - try to use some optional feature, and gracefully fallback if not supported

evverx:
Those checks aren't architecture specific. They just try to do something to make sure it works.

bluca:
Exactly - that's the right fix for this issue

evverx:
I don't think it is because as I said
> I don't think it's a good idea (at this point at least) because any attempts to load BPF skeletons on some architectures can still crash systemd: libbpf/libbpf#391 (that bug was fixed but there are still issues caused by libelf). I think that if it's known that this works on two architectures only it should be compiled and run on two architectures only. I don't think it's hard to update the code once arm is supported.

evverx:
Just to clarify I'm not saying that https://github.com/systemd/systemd/pull/22025 shouldn't be merged. What I'm trying to say that that PR should be combined with this PR

bluca:
No, I want it to work out of the box on arm when the kernel is updated. The bug is fixed by the other PR already.

evverx:
>  I want it to work out of the box on arm when the kernel is updated

I don't think it matters what you want (it doesn't matter what I want either just to be clear). I think what matters is that systemd works correctly.

> The bug is fixed by the other PR already

I'm aware of that

wat-ze-hex:
Summarizing discussion above:
* Fixed list of arches adds tech debt to systemd since it needs to be updated overtime.
* No arch check has potential danger of crashing systemd

The solution could be retrieving that info from libbpf / kernel itself. libbpf supports probing for features, perhaps adding one for BPF trampoline could be feasible solution.
https://github.com/libbpf/libbpf/blob/master/src/libbpf.h#L1048
That needs to be discussed w/ libbpf maintainers.

evverx:
> libbpf supports probing for features, perhaps adding one for BPF trampoline could be feasible solution.

@wat-ze-hex thanks! I think something like that should work but as far as I can remember all the probing functions were kind of deprected at some point. I'm not sure if anything has changed since then though.

evverx:
> No arch check has potential danger of crashing systemd

FWIW @wat-ze-hex if you don't think it's an issue I think it should be safe to just merge  https://github.com/systemd/systemd/pull/22025 without the i386/x86_64 restrictions. I think at the end of the day it's your call. Thanks!

evverx:
@mrc0mmand looks like `test-bpf-lsm` isn't actually run anywhere apart from Arch (which is x86_64). I'm not sure about Ubuntu CI but could it be that it's skipped on Packit because it's run in a sandbox? Last time I checked it wasn't possible to run tests as root there.

evverx:
I looked at https://github.com/systemd/systemd/pull/22027 (where https://github.com/systemd/systemd/pull/22025 wasn't merged yet)

mrc0mmand:
Yeah, in case of Packit it's indeed caused by running under a non-root user (and I can't do anything about that).

As for Ubuntu CI. the test is skipped because of:

```
--- test-bpf-lsm begin ---
test-bpf-lsm: Can't use mlock(), skipping tests.
--- test-bpf-lsm end ---
```

(from aarch64 build).

evverx:
Looks like it's either a bug in `can_memlock` or machines there are somehow constrained too much as well

mrc0mmand:
Ugh, the message I pasted is from TEST-02-UNITTESTS, which makes sense, I guess. However, the "regular" meson test also skips the test, but I can't say why, since there are no verbose logs:

```
501/930 test-bpf-lsm                            SKIP     0.02 s 
```

mrc0mmand:
I managed to run the Packit builds under root, but apart from fewer tests getting skipped, it still doesn't run test-bpf-lsm:

```
BPF LSM hook not enabled in the kernel, LSM BPF not supported
test-bpf-lsm: LSM BPF hooks are not supported, skipping tests.
```

this might be caused either by the fact this runs in an nspawn container or by having an older kernel on the host.

evverx:
> this might be caused either by the fact this runs in an nspawn container or by having an older kernel on the host.

The kernel there is new enough as far as I know so I think it's either nspawn or libseccomp by analogy with https://sourceware.org/pipermail/elfutils-devel/2021q4/004581.html :-)

Regarding Ubuntu CI, the kernel is too old there but other than that apparently it isn't actively maintained and looks like it's probably just patched when it fails. If there had been any attempts to bring the bpf framework there it would have uncovered that `bpftool` isn't enough for it to work because `bpftool` actually has to support the "gen" command also. Without it `ninja` just fails to build `systemd` on Ubuntu Focal with
```
[462/1877] /home/vagrant/systemd/tools/build-bpf-skel.py --clang_exec /usr/bin/clang --llvm_strip_exec /usr/bin/llvm-strip --bpftool_exec /usr/sbin/bpftool --arch x86_64 ../src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c src/core/bpf/restrict_ifaces/restrict-ifaces.skel.h
FAILED: src/core/bpf/restrict_ifaces/restrict-ifaces.skel.h
/home/vagrant/systemd/tools/build-bpf-skel.py --clang_exec /usr/bin/clang --llvm_strip_exec /usr/bin/llvm-strip --bpftool_exec /usr/sbin/bpftool --arch x86_64 ../src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c src/core/bpf/restrict_ifaces/restrict-ifaces.skel.h
Usage: /usr/lib/linux-tools/5.4.0-92-generic/bpftool [OPTIONS] OBJECT { COMMAND | help }
       /usr/lib/linux-tools/5.4.0-92-generic/bpftool batch file FILE
       /usr/lib/linux-tools/5.4.0-92-generic/bpftool version

       OBJECT := { prog | map | cgroup | perf | net | feature | btf }
       OPTIONS := { {-j|--json} [{-p|--pretty}] | {-f|--bpffs} |
	            {-m|--mapcompat} | {-n|--nomount} }
  File "/home/vagrant/systemd/tools/build-bpf-skel.py", line 128, in <module>
    bpf_build(args)
  File "/home/vagrant/systemd/tools/build-bpf-skel.py", line 92, in bpf_build
    gen_bpf_skeleton(bpftool_exec=args.bpftool_exec,
  File "/home/vagrant/systemd/tools/build-bpf-skel.py", line 63, in gen_bpf_skeleton
    skel = subprocess.check_output(bpftool_args, universal_newlines=True)
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/usr/sbin/bpftool', 'g', 's', '../src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.o']' returned non-zero exit status 255.
```

Personally I think it would make sense to get the tests actually running on architectures different from `x86_64` before changing anything

evverx:
> I think it would make sense to get the tests actually running on architectures different from x86_64 before changing anything

Just to clarify, it would make sense in an ideal world :-) It's too late for that at this point. But as far as I can see those tests could have prevented https://bugzilla.redhat.com/show_bug.cgi?id=2036145 and https://github.com/systemd/systemd/pull/22025#issuecomment-1007604974 so it would still be great to get them running.

evverx:
In the meantime I've just opened https://github.com/systemd/systemd/issues/22051

PR #22023, opened 2022-01-05
oomd: move oomctl to bindir

mbiebl:
We don't really need oomctl during early boot, so bindir seems like a
more suitable place for the binary.


yuwata (approved):
LGTM.

PR #22024, opened 2022-01-06
kernel-install: prefer /boot over /boot/efi for $BOOT_ROOT

AdamWill:
This restores the preference order from before 9e82a74. The code
previous to that change 'preferred' /boot over /boot/efi; that
commit changed it to check /boot/efi before checking /boot.
Changing this precedence could (and did, for me) have unexpected
effects - it seems safer to leave it how it was.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/src/kernel-install/kernel-install
+++ b/src/kernel-install/kernel-install
@@ -108,7 +108,7 @@ fi
 [ -z "$MACHINE_ID" ] && MACHINE_ID="Default"
 
 [ -z "$BOOT_ROOT" ] && for suff in "$MACHINE_ID" "loader/entries"; do
-    for pref in "/efi" "/boot/efi" "/boot"; do
+    for pref in "/efi" "/boot" "/boot/efi" ; do
         if [ -d "$pref/$suff" ]; then
             BOOT_ROOT="$pref"
             break 2


AdamWill:
So, what happened here is that, somehow, my Fedora system had *all of the following*:

* `/boot/efi/loader/entries`
* `/boot/efi/$MACHINE_ID`
* `/boot/loader/entries`

What's supposed to happen on Fedora is that it should install the kernel and initramfs to `/boot` and drop a (not-really-proper-BLS) config snippet in `/boot/loader/entries`. I do not know why `/boot/efi/loader/entries` and `/boot/efi/$MACHINE_ID` existed on my system or what created them. But 9e82a74 definitely changed the code such that, while this ambiguous situation was resolved 'correctly' for me before, it no longer is.

Of course, it's possible some other system could be in the opposite situation: it could have both `/boot/loader/entries` and `/boot/efi/loader/entries`, and want to use `/boot/efi`. But that would never have worked up until systemd 249, so far as I can tell. To me, when dealing with an inherently unclear situation like this, the best thing the tool can do is *keep resolving it the same way it always has so far*. So I think it would be better to put the order of precedence back the way it was before, as this PR does. Especially since the commit that changed the order does not explain why it did so, which to me suggests it may have been inadvertent rather than intentional.

AdamWill:
I've sent a call out to other Fedora Rawhide users to ask how many (if any) of them have the same situation I did, it'll be interesting to see what the response is.

keszybz:
Yep, we don't want to change the order. That slipped through inadvertently. Thank you for the patch.

PR #22027, opened 2022-01-06
watchdog: return current timeout

yuwata:
Addresses https://github.com/systemd/systemd/issues/22001#issuecomment-1005896305.

### DIFF
--- a/src/core/manager.c
+++ b/src/core/manager.c
@@ -3234,7 +3234,7 @@ void manager_set_watchdog(Manager *m, WatchdogType t, usec_t timeout) {
 
         if (t == WATCHDOG_RUNTIME)
                 if (!timestamp_is_set(m->watchdog_overridden[WATCHDOG_RUNTIME]))
-                        (void) watchdog_setup(timeout);
+                        (void) watchdog_setup(&timeout);
 
         m->watchdog[t] = timeout;
 }
@@ -3251,7 +3251,7 @@ void manager_override_watchdog(Manager *m, WatchdogType t, usec_t timeout) {
 
         if (t == WATCHDOG_RUNTIME) {
                 timeout = timestamp_is_set(timeout) ? timeout : m->watchdog[t];
-                (void) watchdog_setup(timeout);
+                (void) watchdog_setup(&timeout);
         }
 
         m->watchdog_overridden[t] = timeout;
--- a/src/shared/watchdog.c
+++ b/src/shared/watchdog.c
@@ -163,36 +163,44 @@ int watchdog_set_device(const char *path) {
         return r;
 }
 
-int watchdog_setup(usec_t timeout) {
+int watchdog_setup(usec_t *timeout) {
         usec_t previous_timeout;
         int r;
 
         /* timeout=0 closes the device whereas passing timeout=USEC_INFINITY opens it (if needed)
          * without configuring any particular timeout and thus reuses the programmed value (therefore
-         * it's a nop if the device is already opened). */
+         * it's a nop if the device is already opened).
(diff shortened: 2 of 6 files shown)


yuwata:
cc @evverx, @fbuihuu and @floppym.

src/shared/watchdog.c
```diff
@@ -163,36 +163,42 @@ int watchdog_set_device(const char *path) {
...
         /* timeout=0 closes the device whereas passing timeout=USEC_INFINITY opens it (if needed)
          * without configuring any particular timeout and thus reuses the programmed value (therefore
-         * it's a nop if the device is already opened). */
+         * it's a nop if the device is already opened).
+         *
+         * The timeout argument is used for both input and output. And if the watchdog_timeout is
+         * updated, then the current timeout will be stored into the argument. */
 
-        if (timeout == 0) {
+        if (*timeout == 0) {
```
bluca, line 179: assert or check on timeout

bluca (approved):

src/shared/watchdog.c
yuwata: Indeed. Done. Thanks.

fbuihuu:
Can we wait a bit before merging this ?

 I'd like to give it a look but I can't do it right now.

Thanks.

src/core/manager.c
```diff
@@ -3250,9 +3250,8 @@ int manager_override_watchdog(Manager *m, WatchdogType t, usec_t timeout) {
                 return 0;
 
         if (t == WATCHDOG_RUNTIME) {
-                usec_t usec = timestamp_is_set(timeout) ? timeout : m->watchdog[t];
-
-                (void) watchdog_setup(usec);
+                timeout = timestamp_is_set(timeout) ? timeout : m->watchdog[t];
```
fbuihuu, line 3253: That's wrong, if `timeout` is unset, we want `m->watchdog_overridden[t]` to be unset too and not to be initialized with `m->watchdog[t]`.

fbuihuu:
@yuwata I think your patch  "watchdog: make watchdog_setup() return the current watchdog timeout" is incorrect: IIRC there's a reason why `watchdog_setup()` doesn't update the timeout parameter: we want to store in `m->watchdog_overridden[]` the timeout value set by the user not the one actually used by the driver because if PID1 is reexecuted and the watchdog device was changed (in the config), we want to initialize the new device with the user's timeout value, not with the timeout used by the previous watchdog device which might  have been approximated.

src/shared/watchdog.c
```diff
@@ -54,14 +54,12 @@ static int watchdog_get_timeout(void) {
...
 static int watchdog_set_timeout(void) {
-        usec_t t;
         int sec;
 
         assert(watchdog_fd >= 0);
         assert(timestamp_is_set(watchdog_timeout));
 
-        t = DIV_ROUND_UP(watchdog_timeout, USEC_PER_SEC);
-        sec = MIN(t, (usec_t) INT_MAX); /* Saturate */
+        sec = MIN(DIV_ROUND_UP(watchdog_timeout, USEC_PER_SEC), (usec_t) INT_MAX); /* Saturate */
```
fbuihuu, line 62: You earned 2 lines of code but loosed in readability (at least IMO), not sure if it's worth the trade off ;)

floppym:
I'm not sure that these changes really fix any problems. See #22031 for some simpler changes to test-watchdog.

fbuihuu:
@yuwata could you have a look at https://github.com/systemd/systemd/pull/22037 ?

I think it should replace your PR.

PR #19402, opened 2021-04-22
test: configure swap for TEST-55-OOMD

mrc0mmand:
oomd works way better with swap, so let's make the test less flaky by
configuring a swap device for it. This also allows us to drop the ugly
`cat`s from the load-generating script.

---

@anitazha does this sound reasonable? I mean, we lose the possible coverage for systems without swap, but it should make the test more robust (and actually test the environment we expect oomd to run in).

PR #22029, opened 2022-01-06
test: make sure TEST-55-OOMD runs under qemu

mbiebl:
To be able to run the test, we need to ensure that swap is available.
Case in point, the autopkgtest environment does not setup test images
with a swap partition/file.

So one gets an warning like this:

[  158.188957] systemd-oomd[67]: Swap is currently not detected; memory pressure usage will be degraded

if nspawn is used, e.g. via TEST_PREFER_NSPAWN=1.

Thus ensure QEMU is always used so we can successfully attach a swap
device.

### DIFF
--- a/test/TEST-55-OOMD/test.sh
+++ b/test/TEST-55-OOMD/test.sh
@@ -4,6 +4,7 @@ set -e
 
 TEST_DESCRIPTION="systemd-oomd Memory Pressure Test"
 IMAGE_NAME="oomd"
+TEST_NO_NSPAWN=1
 
 # shellcheck source=test/test-functions
 . "${TEST_BASE_DIR:?}/test-functions"


mbiebl:
unless of course, TEST-55-OOMD is supposed to work without swap being available.

mrc0mmand:
systemd-oomd is supposed to work both with and without swap, so the nspawn part of the test covered the latter scenario (as since https://github.com/systemd/systemd/pull/19402 we explicitly configure swap for the qemu part).

mbiebl:
With #22032 applied, I do get an error when trying to run the test via nspawn.

log and journal are attached.
[journal.txt](https://github.com/systemd/systemd/files/7828652/journal.txt)
[log.txt](https://github.com/systemd/systemd/files/7828653/log.txt)

mrc0mmand:
I see, looks like without swap the pressure increases more rapidly and oomd kills the whole user scope instead of just the "offending" unit:

```
Jan 07 12:58:57 H testsuite-55.sh[164]: ● testsuite-55-testbloat.service - Create a lot of memory pressure
Jan 07 12:58:57 H testsuite-55.sh[164]:      Loaded: loaded (/usr/lib/systemd/tests/testdata/units/testsuite-55-testbloat.service; static)
Jan 07 12:58:57 H testsuite-55.sh[164]:      Active: active (running) since Fri 2022-01-07 11:58:44 UTC; 13s ago
Jan 07 12:58:57 H testsuite-55.sh[164]:    Main PID: 111
Jan 07 12:58:57 H testsuite-55.sh[164]:       Tasks: 2 (limit: 1122)
Jan 07 12:58:57 H testsuite-55.sh[164]:      Memory: 6.4M (high: 5.0M available: 0B)
Jan 07 12:58:57 H testsuite-55.sh[164]:         CPU: 18ms
Jan 07 12:58:57 H testsuite-55.sh[164]:      CGroup: /user.slice/user-4711.slice/user@4711.service/testsuite.slice/testsuite-55.slice/testsuite-55-workload.slice/testsuite-55-testbloat.service
Jan 07 12:58:57 H testsuite-55.sh[164]:              ├─111 bash /usr/lib/systemd/tests/testdata/units/testsuite-55-slowgrowth.sh
Jan 07 12:58:57 H testsuite-55.sh[164]:              └─117 bash /usr/lib/systemd/tests/testdata/units/testsuite-55-slowgrowth.sh
Jan 07 12:58:57 H testsuite-55.sh[42]: + sleep 2
Jan 07 12:58:57 H systemd-oomd[68]: oomd: New incoming message: {"parameters":{"cgroups":[{"mode":"auto","path":"/system.slice/run-u35.service","property":"ManagedOOMSwap"},{"mode":"auto","path":"/system.slice
/run-u35.service","property":"ManagedOOMMemoryPressure"}]},"continues":true}
Jan 07 12:58:57 H systemd-oomd[68]: oomd: Changing state awaiting-reply-more → processing-reply
Jan 07 12:58:57 H systemd-oomd[68]: oomd: Changing state processing-reply → awaiting-reply-more
Jan 07 12:58:57 H systemd-oomd[68]: oomd: New incoming message: {"parameters":{"cgroups":[{"mode":"auto","path":"/user.slice/user-4711.slice/session-12.scope","property":"ManagedOOMSwap"},{"mode":"auto","path"
:"/user.slice/user-4711.slice/session-12.scope","property":"ManagedOOMMemoryPressure"}]},"continues":true}
Jan 07 12:58:57 H systemd-oomd[68]: oomd: Changing state awaiting-reply-more → processing-reply
Jan 07 12:58:57 H systemd-oomd[68]: oomd: Changing state processing-reply → awaiting-reply-more
Jan 07 12:58:58 H systemd-oomd[68]: Memory pressure for /user.slice/user-4711.slice/user@4711.service is 71.94% > 50.00% for > 2s with reclaim activity
Jan 07 12:58:58 H systemd-oomd[68]: Failed to insert context for /user.slice/user-4711.slice/user@4711.service/testsuite.slice/testsuite-55.slice/testsuite-55-workload.slice/testsuite-55-testchill.service, ign
oring: File exists
Jan 07 12:58:58 H systemd-oomd[68]: Failed to insert context for /user.slice/user-4711.slice/user@4711.service/testsuite.slice/testsuite-55.slice/testsuite-55-workload.slice/testsuite-55-testbloat.service, ign
oring: File exists
Jan 07 12:58:58 H systemd-oomd[68]: oomd attempting to kill 111 with KILL
Jan 07 12:58:58 H systemd[34]: testsuite-55-testbloat.service: Main process exited, code=killed, status=9/KILL
Jan 07 12:58:58 H systemd[34]: testsuite-55-testbloat.service: Failed with result 'signal'.
Jan 07 12:58:58 H systemd-oomd[68]: oomd attempting to kill 117 with KILL
Jan 07 12:58:58 H systemd-oomd[68]: oomd attempting to kill 34 with KILL
Jan 07 12:58:58 H systemd-oomd[68]: oomd attempting to kill 35 with KILL
Jan 07 12:58:58 H systemd-oomd[68]: Killed /user.slice/user-4711.slice/user@4711.service/init.scope due to memory pressure for /user.slice/user-4711.slice/user@4711.service being 71.94% > 50.00% for > 2s with 
reclaim activity
```

Given that the swap is highly recommended for oomd, I guess we should merge this to make the test more reliable.

/cc @anitazha just in case I misunderstand what exactly happened in this case.

anitazha:
I'm on the road for another week or so my responses will be spotty.

If Debian is using the same configs that Fedora has, that means there is a drop in for the user@.service, -.slice, and oomd.conf. Likely these 3 are taking precedence over the configs in TEST-55. The test was written to expect no other systemd-oomd configs in the environment. It looks like in the other PR mrc0mmand overrode the one for oomd.conf. We likely need to do the same for the user manager drop-in (which has a memory pressure limit) and maybe the root cgroup drop-in (this one has a swap limit) to get the test environment in the expected state.

slyon:
> If Debian is using the same configs that Fedora has, that means there is a drop in for the [user@.service](mailto:user@.service), -.slice, and oomd.conf.

Yes, indeed. We're shipping the same configuration as Fedora does in Debian.

mrc0mmand:
> I'm on the road for another week or so my responses will be spotty.
> 
> If Debian is using the same configs that Fedora has, that means there is a drop in for the [user@.service](mailto:user@.service), -.slice, and oomd.conf. Likely these 3 are taking precedence over the configs in TEST-55. The test was written to expect no other systemd-oomd configs in the environment. It looks like in the other PR mrc0mmand overrode the one for oomd.conf. We likely need to do the same for the user manager drop-in (which has a memory pressure limit) and maybe the root cgroup drop-in (this one has a swap limit) to get the test environment in the expected state.
> 
> The test itself shouldn't be killing the user manager otherwise since nothing in it is configured with a memory pressure limit. That's why I think there is an external config doing so.

I see, thanks for the explanation! I'll tweak the test further to fix this.

mbiebl:
@anitazha so, @mrc0mmand fixed the overrides in #22032, but I still get test failures when no swap device is available.
Log files are over at https://github.com/systemd/systemd/pull/22032#issuecomment-1007684637

Any ideas?

mbiebl:
In case you want to reproduce the issue, here's my setup:
```
git clone https://salsa.debian.org/systemd-team/systemd.git
```
Switch to `debian/experimental` branch
Build the package
```
apt-get build-dep systemd
dpkg-buildpackage -b
```
Setup an autopkgtest VM
```
autopkgtest-build-qemu sid /path/to/sid.img
```
Edit systemd/debian/tests/upstream to only run `TEST-55-OOMD` by adding `export SELECTED_TESTS=TEST-55-OOMD`

Run the `upstream` test
```
autopkgtest -o logs *.deb systemd/ --test-name=upstream -- qemu /path/to/sid.img
```

I tried various options like `qemu --ram-size=4096 /path/to/sid.img`, bug that didn't help.
Only why I could make it pass reliably is to add a swap file inside sid.img.

mbiebl:
Small update: While playing around with `--ram-size=` didn't really help, using -c (num of virtual CPUs) did.
The default is 1, I've run 5 tests in a row with `-c 2` and all of them succeeded.

mbiebl:
@anitazha Did you have a chance to look into this in the mean time?

anitazha:
@mbiebl Just got back yesterday so I'll look into it sometime today after I catch up on work

anitazha:
Thanks for the commands @mbiebl  those were really helpful. I was able to repro and I've confirmed the theory I commented about in https://github.com/systemd/systemd/pull/22032 by adding some more log lines. I'll try to fix this tomorrow and get a PR out

mbiebl:
> Thanks for the commands @mbiebl those were really helpful. I was able to repro and I've confirmed the theory I commented about in #22032 by adding some more log lines

Awesome. Since you suspect a race condition this could also explain why adding additional CPUs to the VM helped in my case.

PR #22031, opened 2022-01-06
test-watchdog adjustments

floppym:
- Use watchdog_runtime_wait() to determine sleep interval
- Set timeout to 2 seconds by default

See https://github.com/systemd/systemd/issues/22001.

### DIFF
--- a/src/test/test-watchdog.c
+++ b/src/test/test-watchdog.c
@@ -17,7 +17,7 @@ int main(int argc, char *argv[]) {
 
         slow = slow_tests_enabled();
 
-        t = slow ? 10 * USEC_PER_SEC : 1 * USEC_PER_SEC;
+        t = slow ? 10 * USEC_PER_SEC : 2 * USEC_PER_SEC;
         count = slow ? 5 : 3;
 
         r = watchdog_setup(t);


src/test/test-watchdog.c
```diff
@@ -27,12 +27,13 @@ int main(int argc, char *argv[]) {
                 t = 0;
 
         for (i = 0; i < count; i++) {
+                t = watchdog_runtime_wait();
+                log_info("Sleeping " USEC_FMT " microseconds...", t);
+                usleep(t);
```
bluca, line 32: isn't it risky to sleep for the same time (you're at the mercy of the scheduler), any reason to change from t/2?

floppym: watchdog_runtime_wait() returns half the timeout at most, and takes the last ping into account.

https://github.com/systemd/systemd/blob/v250/src/shared/watchdog.c#L208

floppym: Also, watchdog_runtime_wait() does not actually wait/sleep itself. It just performs a calculation and returns the result.

bluca (approved):

src/test/test-watchdog.c
```diff
@@ -27,12 +27,13 @@ int main(int argc, char *argv[]) {
                 t = 0;
 
         for (i = 0; i < count; i++) {
+                t = watchdog_runtime_wait();
+                log_info("Sleeping " USEC_FMT " microseconds...", t);
```
fbuihuu, line 31: Isn't this info already displayed by the previous call to `watchdog_setup()` ?

fbuihuu:
Guys, you seem to be so eagier to merge PRs sometimes ;)

src/test/test-watchdog.c
floppym: watchdog_setup outputs the timeout value. The value from watchdog_runtime_wait is approximately half of that, and varies based on the exact time of the last ping.

The output now looks like this:

```
Using hardware watchdog 'SP5100 TCO timer', version 0, device /dev/watchdog0
Watchdog running with a timeout of 2s.
Sleeping 999999 microseconds...
Pinging...
Sleeping 1000000 microseconds...
Pinging...
Sleeping 1000000 microseconds...
Pinging...
```

Issue #15316, opened 2020-04-02
"Unexpected error response from GetNameOwner(): Connection terminated" messages + boot takes a lot of time


tomgar:
**systemd version the issue has been seen with**
245.4

**Used distribution**
Archlinux

**Expected behaviour you didn't see**
* Not taking more than 5 seconds to get to the graphical target at boot
* No error messages
* Expected systemd-analyze output:
```
% systemd-analyze
Startup finished in 6.791s (firmware) + 3.762s (loader) + 1.088s (kernel) + 25.338s (initrd) + 5.208s (userspace) = 42.189s 
graphical.target reached after 4.606s in userspace

% systemd-analyze critical-chain 
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

graphical.target @4.606s
└─lightdm.service @3.476s +1.129s
  └─systemd-user-sessions.service @3.469s +6ms
    └─nss-user-lookup.target @4.332s

```

**Unexpected behaviour you saw**
 * Odd errors in the journal. Here's the relevant log lines:
```
Apr 02 20:47:23 nix lightdm[1054]: Error getting user list from org.freedesktop.Accounts: Timeout was reached
Apr 02 20:47:23 nix accounts-daemon[1176]: error getting polkit authority: Error initializing authority: Error calling StartServiceByName for org.freedesktop.PolicyKit1: Timeout was reached
Apr 02 20:47:23 nix dbus-daemon[989]: [system] Failed to activate service 'org.freedesktop.Accounts': timed out (service_start_timeout=25000ms)
Apr 02 20:47:23 nix dbus-daemon[989]: [system] Failed to activate service 'org.freedesktop.PolicyKit1': timed out (service_start_timeout=25000ms)
Apr 02 20:47:23 nix systemd[1]: NetworkManager.service: Unexpected error response from GetNameOwner(): Connection terminated
Apr 02 20:47:23 nix systemd[1]: systemd-logind.service: Unexpected error response from GetNameOwner(): Connection terminated
Apr 02 20:47:23 nix systemd[1]: systemd-homed.service: Unexpected error response from GetNameOwner(): Connection terminated
Apr 02 20:47:23 nix systemd[1]: bluetooth.service: Unexpected error response from GetNameOwner(): Connection terminated
Apr 02 20:47:23 nix systemd[1]: lightdm.service: Unexpected error response from GetNameOwner(): Connection terminated
[...]

PR #22038, opened 2022-01-07
Fix deadlock between nss-systemd and dbus (dbus call timeout)

slyon:
The fix from fd63e71 wasn't updated when dynamic user lookup switched away from d-bus.

This solves a long standing issue where dbus calls would hit a timeout that we have been observing randomly since systemd v245 and is really hard to reproduce. Most recently @nowrep came up with a good suggestion that seems to solve the issue!

We're looking for review of this suggested change.

Fixes: #15316

### DIFF
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -1886,7 +1886,7 @@ static int build_environment(
          * users via PID 1, possibly dead-locking the dbus daemon. This way it will not use D-Bus to resolve names, but
          * check the database directly. */
         if (p->flags & EXEC_NSS_BYPASS_BUS) {
-                x = strdup("SYSTEMD_NSS_BYPASS_BUS=1");
+                x = strdup("SYSTEMD_NSS_DYNAMIC_BYPASS=1");
                 if (!x)
                         return -ENOMEM;
                 our_env[n_env++] = x;


bluca:
How is varlink causing a deadlock with dbus exactly? The old env var is no longer used for anything and could be removed yes, but could you please explain why skipping resolution of DynamicUser has anything to do with dbus-daemon, which uses a fixed user everywhere (afaik)?

nowrep:
It's deadlocking because both d-bus and varlink calls are blocking.

bluca:
But how? Which parts are? Dbus is clear as it's dbus-daemon that provides it, so it's a circular dependency. What's blocking the varlink resolution?

nowrep:
systemd does blocking GetConnectionUnixUser and dbus will then (via nss-systemd) do blocking varlink call to systemd.

bluca:
I see, what's the call stack? The commit message is very vague, it would greatly benefit from having all these details

DemiMarie:
> systemd does blocking GetConnectionUnixUser and dbus will then (via nss-systemd) do blocking varlink call to systemd.
> 
> It's exactly the same situation as before, it doesn't matter nss-systemd no longer uses dbus to communicate with systemd daemon. It still does blocking call so it will deadlock just the same.

Why is systemd doing a blocking call?  nss-systemd is stuck having to conform to a blocking interface but systemd is under no such restrictions.

slyon:
Unfortunately I am still unable to reproduce this issue myself.
@nowrep has been spot on with his suggestion for a workaround, I wonder if you know any more about the call stack that leads to this situation?

nowrep:
I don't have a stack trace as it only happens early during the boot (in my case it's triggered when starting user session).

The issue is easy to see however. This is where the systemd does the problematic blocking call: 
https://github.com/systemd/systemd/blob/bfae960e53f6986f1c4d234ea82681d0acad57df/src/libsystemd/sd-bus/bus-control.c#L658-L680

eventually dbus will call into nss-systemd while handling this call and you have a deadlock.

It's a race and it will only deadlock in a rare case where this systemd call is the first time dbus will need to query info about that particular user (dbus has a cache for user database).

poettering:
(btw, my educated guess is that this is an issue with dbus-daemon only, and not with dbus-broker, since the latter resolves names very differently and very specific times only. That's why this was never noticed on Fedora)

slyon:
> (btw, my educated guess is that this is an issue with dbus-daemon only, and not with dbus-broker, since the latter resolves names very differently and very specific times only. That's why this was never noticed on Fedora)

I can confirm that we did not see this issue when using `dbus-broker`

poettering:
I prepped #22552 now, which closely matches this PR, but is a bit more comprehensive, and starts with converting synchronous calls to the dbus broker into asynchronous code. But there's a lot left to do in that area.

Anyway, let's contine this in #22552. I credited @slyon on the patch (and for tracking down the deadlock) given my patch is just a more comprehensive version of his work. 

I hope it's OK if I close this one here.

poettering:
> So you feel like we could merge this?

Well, yes more or less. But the fix should be more comprehensive, to make things easy see → #22552

poettering:
> > (btw, my educated guess is that this is an issue with dbus-daemon only, and not with dbus-broker, since the latter resolves names very differently and very specific times only. That's why this was never noticed on Fedora)
> 
> How does dbus-broker avoid this deadlock?

Iirc dbus-daemon resolves lazily and dbus-broker resolves ahead of time. the lazy resolution means that it might happen when triggered by a blocking PID 1 operation, but in dbus-broker that's not gonna happen then.

> > But doing cascades of async stuff from async handlers is kinda nasty to write in C and not lose track of what happens.
> 
> Would using Rust and async functions help? What about some sort of coroutine abstraction?

Well, the developers of systemd are generally positive on Rust I think, me included. The big issue is build systems: when starting to go Rust with systemd we need some way to build a hybrid codebase effectively, and that for a long time. So far no build system even remotely makes that easy. The projects that mix C and Rust usually have a very clear separation about leaf/stem components and only one side is Rust. In systemd with its large body of interdependent components we need something that can deal with more complex builds, i.e we want to port library functions over, but others not, and some services overs but others not, and that means you have dependencies from C to rust to C to rust and so on. So far the meson and rust/cargo people were not the biggest of friends, and given there aren't really other projects of this scales who started doing such a transition with a codebase arranged that way, we'd be pioneering this, and we really don#t want to be pioneers with this, but just use rust as tool that already works.

or to say this differently: we build a lot of spearate libraries, internally and externally, that consumer each others and a lot of separate ELF binaries that consume them all. This is quite different from the usual area where Rust is used where you basiclly just build one big binary and that's it...

in other words, we are waiting for other people to solve the toolchains problems properly before we bother, because *we* want to spend our time on solving other problems instead.

DemiMarie:
> in other words, we are waiting for other people to solve the toolchains problems properly before we bother, because _we_ want to spend our time on solving other problems instead.

If I recall correctly, Meson has decent Rust support already, but only if you don’t have `crates.io` dependencies.  That would be a blocker for lots of projects, but my understanding is that systemd would not need any such dependencies.

poettering:
if we go rust, we should go rust properly. i.e. we should be able to use crates where it makes sense.

evverx:
FWIW apart from build systems I'd add that rust testing infrastructure is far from what I would call ideal (especially when it comes to mixing two languages). It was discussed in https://github.com/systemd/systemd/pull/19598#discussion_r641902890 and as far as I can tell none of those issues have been addressed.

@DemiMarie another issue is https://dl.acm.org/doi/10.1145/3418898 (the paper was published about a year ago so I'm not sure how relevant it is today. If that was addressed somehow I'd appreciate it if you could point me in the right direction) Thanks!

PR #19600, opened 2021-05-13
Bootctl short-id

kreijack:
The IDs of the boot entries are very long and difficult to insert
correctly when the commands 'set-default' and 'set-oneshot' are used.

So it is introduced a new id (called 'short-id') to indentify a
boot entry. These new IDs are in the form of @1, @2...

```
$ bootctl list
Boot Loader Entries:
        title: Debian GNU/Linux bullseye/sid (5.9.12)
           id: e84907d099904117b355a99c98378dca-5.9.12.conf
     short-id: @1
       source: /boot/efi/loader/entries/e84907d099904117b355a99c98378dca-5.9.12.conf
      version: 5.9.12
   machine-id: e84907d099904117b355a99c98378dca
        linux: /e84907d099904117b355a99c98378dca/5.9.12/linux
       initrd: /e84907d099904117b355a99c98378dca/5.9.12/initrd
      options: root=UUID=d7a06504-cc14-435d-a5df-674da09c2894 ro 

        title: Debian GNU/Linux bullseye/sid (5.10.16)
           id: e84907d099904117b355a99c98378dca-5.10.16.conf
     short-id: @2
       source: /boot/efi/loader/entries/e84907d099904117b>
      version: 5.10.16
 
[...]

```
And the same 'short-id' can be used in the command 'set-default' (and set-oneshot)

```
$ sudo bootctl set-default @1
$ bootctl
System:
     Firmware: UEFI 2.70 (American Megatrends 5.17)
  Secure Boot: disabled
   Setup Mode: setup
[...]
Default Boot Loader Entry:
        title: Debian GNU/Linux bullseye/sid (5.9.12)
           id: e84907d099904117b355a99c98378dca-5.9.12.conf
       source: /boot/efi/loader/entries/e84907d099904117b355a99c98378dca-5.9.12.conf
      version: 5.9.12
```

These IDs are temporary. When **bootctl** see these short-id, it searches the correspondent ID and sets 
the EUFI variables with the latter.

PR #10175, opened 2018-09-25
the bootctl changes from PR #9437 (the boot counting PR)

poettering:
As requested by @keszybz 

This also contains the commits from PR #10173, as they are required by this PR.

PR #10173, opened 2018-09-25
the utf8.c changes from PR #9437 (i.e. the boot counting PR)

poettering:
@keszybz asked for this to be split out into a separate PR

Note that these changes have no real user without #9437 being merged too.

PR #22041, opened 2022-01-07
bootctl: ignore the bootloader boot entries

kreijack:
This topic was discussed previously here https://github.com/systemd/systemd/pull/19600#issuecomment-898277697

Basically, __bootctl list__ shows a list of entries which is a sum of the ones synthesized from */boot/efi/loaders/...* and ones returned by systemd-boot (via the LoaderEntries efi variable). The information returned by __systemd-boot__ are still synthesized from */boot/efi/loaders/...* boot at __boot time__.

The problem arise when (e.g.) after a removal of a kernel package, __bootctl__ still shows the deleted entries because __systemd-boot__ saw it at __boot time__.

Below an example of what happens:
```
$ bootctl list
Boot Loader Entries:
        title: Debian GNU/Linux bookworm/sid (5.15.12) (default)
           id: e84907d099904117b355a99c98378dca-5.15.12.conf
       source: /boot/efi/loader/entries/e84907d099904117b355a99c98378dca-5.15.12.conf
      version: 5.15.12
   machine-id: e84907d099904117b355a99c98378dca
        linux: /e84907d099904117b355a99c98378dca/5.15.12/linux
       initrd: /e84907d099904117b355a99c98378dca/5.15.12/initrd
      options: root=UUID=d7a06504-cc14-435d-a5df-674da09c2894 ro rootflags=subvol=@rootfs btrfsrollback=@rol>

        title: e84907d099904117b355a99c98378dca-5.15.4.conf
           id: e84907d099904117b355a99c98378dca-5.15.4.conf
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

        title: e84907d099904117b355a99c98378dca-5.15.7.conf
           id: e84907d099904117b355a99c98378dca-5.15.7.conf
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

        title: Reboot Into Firmware Interface
           id: auto-reboot-to-firmware-setup
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

```
The kernel 5.15.7 and 5.15.4 are removed but because __systemd-boot__ saw it during the boot, __bootctl__ continue to show them.

This patch change the __bootctl__ behavior, so it ignores the information returned by __systemd-boot__ related to */boot/efi/loaders/...*

```
$ build/bootctl list
Boot Loader Entries:
        title: Debian GNU/Linux bookworm/sid (5.15.12) (default)
           id: e84907d099904117b355a99c98378dca-5.15.12.conf
       source: /boot/efi/loader/entries/e84907d099904117b355a99c98378dca-5.15.12.conf
      version: 5.15.12
   machine-id: e84907d099904117b355a99c98378dca
        linux: /e84907d099904117b355a99c98378dca/5.15.12/linux
       initrd: /e84907d099904117b355a99c98378dca/5.15.12/initrd
      options: root=UUID=d7a06504-cc14-435d-a5df-674da09c2894 ro rootflags=subvol=@rootfs btrfsrollback=@rol>

        title: Reboot Into Firmware Interface
           id: auto-reboot-to-firmware-setup
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

```

In the end no information is lost, because __bootctl__ ignores the information of systemd-boot that it is already capable to obtain autonomously. 

It is added a new option for __bootctl__  (called --entry-from-loader) which restores the old behavior.

-------------------------------------------------------------------------------------------------
When bootctl lists the boot entries, considers also the ones
returned by systemd-boot (via the efi LoaderEntries variable),
created at boot time.

Unfortunately this list may became incorrect if (e.g.) the user remove a
kernel package.

This patch changes this behaviour, so bootctl ignores some the
boot entries returned by systemd-boot.

In any case, bootctl still considers the 'auto-xxx' boot entries
listed below:
```
  Boot entrie name                 Title
  -----------------------------    ------------------------------
  auto-osx                         macOS boot loader
  auto-windows                     Windows Boot Manager
  auto-efi-shell                   EFI Shell
  auto-efi-default                 EFI Default Loader
  auto-reboot-to-firmware-setup    Reboot Into Firmware Interface
```

The other entries that systemd-boot synthetizes (e.g. the ones loaded from
/efi/loader/entries/<uuid>) can be synthetized by bootctl too, so no
information is lost.

The old behviour can be restored using the switch '--entry-from-loader'.

Signed-off-by: Goffredo Baroncelli <kreijack@inwind.it>

### DIFF
--- a/man/bootctl.xml
+++ b/man/bootctl.xml
@@ -277,6 +277,13 @@
         Partitions, or that multiple OS instances are not possible.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--entry-from-loader</option></term>
+        <listitem><para>This option modifies the behaviour of <command>list</command>, taking in consideration
+	also the boot entries returned by the boot loader.
+        </para></listitem>
+      </varlistentry>
+
       <xi:include href="standard-options.xml" xpointer="no-pager"/>
       <xi:include href="standard-options.xml" xpointer="help"/>
       <xi:include href="standard-options.xml" xpointer="version"/>
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -53,6 +53,7 @@ static char *arg_xbootldr_path = NULL;
 static bool arg_print_esp_path = false;
 static bool arg_print_dollar_boot_path = false;
 static bool arg_touch_variables = true;
+static bool arg_entry_from_loader = false;
 static PagerFlags arg_pager_flags = 0;
 static bool arg_graceful = false;
 static int arg_make_machine_id_directory = 0;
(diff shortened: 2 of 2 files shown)


keszybz:
I wanted to figure out **why** we are adding those boot entries, but the reason seems to be lost in the shadows of time. It was added in #10175, which was split out of  #10173, which was split out of #9437, which has too many comments to read. bd2865caf6e730fa26807b749b1c3269943df8e2 does not describe the motivation. @poettering or anyone else, do you remember?

keszybz:
So we show entries from the bootloader that are not "auto-" entries. In the common case those are entries that have been uninstalled since we booted the machine and showing them is not useful. I remember we discussed this in the past, but as mentioned above, I don't remember the outcome of the discussion. Right now I want to say that we should just stop doing that (and not even bother with adding an option), but there must have been some good reason.

keszybz:
@poettering do you remember the reasoning behind showing those entries?

keszybz:
@kreijack Could you remove the option for now? @poettering will not look into it for a while yet, but if it turns out that we want the functionality, we can always dig out this version of the patch.

kreijack:
Ok; it is not a problem to remove the option. I will add a further patch that will remove also the argument from the function boot_entries_augment_from_loader(), because in each invocation the last parameter (called "only_auto") is always "true".
```
$ egrep -r boot_entries_augment_from_loader .
./src/boot/bootctl.c:                (void) boot_entries_augment_from_loader(&config, efi_entries, true);
./src/shared/bootspec.h:int boot_entries_augment_from_loader(BootConfig *config, char **list, bool only_auto);
./src/shared/bootspec.c:int boot_entries_augment_from_loader(
./src/login/logind-dbus.c:                (void) boot_entries_augment_from_loader(&config, m->efi_boot_loader_entries, true);
./src/login/logind-dbus.c:                (void) boot_entries_augment_from_loader(&config, m->efi_boot_loader_entries, true);

```

If we don't want this further step we can drop the last patch.

kreijack:
I reopened this issue

PR #22043, opened 2022-01-07
boot: Add BitLocker TPM key sealing workaround

medhefgo:


medhefgo:
Note that this is ontop of (#22028) for convenience.

@danderson please gives this a try.

man/loader.conf.xml
```diff
@@ -196,6 +196,28 @@
...
+        and Windows Boot Manager is selected in the boot menu, set the <literal>BootNext</literal>
+        EFI variable and reset the system. The firmware will then boot into Windows Boot Manager
+        directly, leaving the TPM PCRs in expected states so that Windows can unseal the encryption
+        key. This allows systemd-boot to be updated without needing to provide the recovery key to
+        BitLocker.</para>
+
+        <para>Note that the PCRs that Windows uses can be configured with the
+        <literal>Configure TPM platform validation profile for native UEFI firmware configurations</literal>
+        group policy under <literal>Computer Configuration\Administrative Templates\Windows Components\BitLocker Drive Encryption</literal>.
+        When secure boot is enabled, changing this to PCRs <literal>0,2,7,11</literal> should be save.
```
bluca: /save/safe/

bluca:
needs a rebase

src/boot/efi/boot.c
```diff
@@ -1943,6 +1948,82 @@ static void config_entry_add_osx(Config *config) {
         }
 }
 
+static EFI_STATUS boot_windows_bitlocker(void) {
+        _cleanup_freepool_ EFI_HANDLE *handles = NULL;
+        UINTN n_handles;
+        EFI_STATUS err;
+
+        /* BitLocker key cannot be sealed without a TPM presetn. */
```
bluca: /presetn/present/

src/boot/efi/boot.c
```diff
@@ -1943,6 +1948,82 @@ static void config_entry_add_osx(Config *config) {
...
+                err = BS->HandleProtocol(handles[i], &BlockIoProtocol, (void **) &block_io);
+                if (EFI_ERROR(err) || block_io->Media->BlockSize < 512)
+                        continue;
+
+                CHAR8 buf[block_io->Media->BlockSize];
+                block_io->ReadBlocks(block_io, block_io->Media->MediaId, 0, sizeof(buf), buf);
+                if (EFI_ERROR(err))
+                        continue;
+
+                if (CompareMem(buf + 3, "-FVE-FS-", STRLEN("-FVE-FS-")) == 0) {
```
bluca, line 1977: Maybe add a `#define` for this? Magic strings in the middle of a function are not great for readability

src/boot/efi/boot.c
```diff
@@ -85,6 +85,7 @@ typedef struct {
         BOOLEAN editor;
         BOOLEAN auto_entries;
         BOOLEAN auto_firmware;
+        BOOLEAN bitlocker;
```
bluca: This (and the corresponding config) could be more descriptive I think - maybe something like bitlocker_boot_via_reset? A bit longer but more self-descriptive

bluca (approved):
just some nits

src/boot/efi/measure.c
```diff
@@ -135,6 +135,10 @@ static EFI_TCG2 * tcg2_interface_check(void) {
         return tcg;
 }
 
+BOOLEAN tpm_present(void) {
+        return !!tcg2_interface_check() || !!tcg1_interface_check();
```
keszybz: !! is not necessary — such a conversion to a boolean happens automatically.

man/loader.conf.xml
```diff
@@ -196,6 +196,28 @@
         by using the <keycap>f</keycap> key.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term>bitlocker</term>
```
keszybz: Please give this a more descriptive name. I'm pretty sure that this commit will not be the last time we have to deal with bitlocker, so a single-word description (and a noun in particular), might give us a headache in the future. Maybe "reboot-for-bitlocker" or something like that.

man/loader.conf.xml
```diff
@@ -196,6 +196,28 @@
...
+
+        <listitem><para>Work around BitLocker requiring a recovery key when the boot loader was
+        updated (enabled by default).</para>
+
+        <para>Try to detect BitLocker encrypted drives along with an active TPM. If both are found
+        and Windows Boot Manager is selected in the boot menu, set the <literal>BootNext</literal>
+        EFI variable and reset the system. The firmware will then boot into Windows Boot Manager
```
keszybz: "reset" or "reboot" ?

src/boot/efi/boot.c
keszybz: Oh, right, I made a similar comment above :)

keszybz (changes requested):

src/boot/efi/boot.c
medhefgo: For single use magic strings, `#define` doesn't really help much, imho.

bluca:
No response yet from the reporter - given it was tested in qemu, let's merge this

poettering:
I think we should not bother with trying to detect BitLocker. BitLocker might be the primary user for this, but not necessarily the only one, i.e. any kind of of Windows concept thta binds stuff to TPM might be subject to this.

Hence, I'd probably drop the BitLocker detction and just do this always if a TPM exists and is enabled, regardless if Bitlocker is also detected or not.

In fact I am kinda tempted to always boot other (i.e. non Linux) OSes this way, except I have two concerns:

1. This means we write to EFI memory on every such boot, and given the low report quality of such chips this might be something to avoid if we can.
2. it's a bit more fragile, since we only can hope the boot entry exists and works, we are not actually doing anything ourselves there...

I wonder if we should synthesize a Boot0XYZ entry automatically if needed (and rewrite it whenever we need this again). By keeping it out of BootOrder it shouldn't be considered for uses outside of our immediate use. we could make the Boot0XYZ entry recognizable as "sd-boot One-Time Boot Entry" or so, and then always find it again with that... Why all that? I am thinking of self-contained removable media here, i.e. where one cannot assume all OSes installed on the media actually have a Boot0XYZ entry on the host already.

poettering:
btw, i woder if there should be a new knob selecting this mode in each type #1 snippet individually. I mean the use case is not specific to Windows really, it could similarly apply to any other OS that uses TPM, including Linux-based ones, hence adding a per-snippet boolean option might be wise that forces this mode? 

(This even more makes me want to drop the BitLocker specificity of the concept, in particular in the naming of the option, it just be called execute-by-nextboot or so, and maybe take it true/false/auto or so, where "auto" means "turn this on for Windows when a TPM exists, otherwise off", and is the default. And we'd expressly take the liberty to extend this further on, i.e. maybe also move macos under that umbrella should that be necessary)

PR #22047, opened 2022-01-07
src/fundamental: list fundamental_source_paths using relative paths

kanavin:
Otherwise, the compiler takes the full path to the source file
and writes it into the binary output, breaking reproducibility.

As these paths are only used in one place (src/boot/efi/meson.build),
it is safe to assume two levels of going up the source tree.

### DIFF
--- a/src/fundamental/meson.build
+++ b/src/fundamental/meson.build
@@ -20,7 +20,7 @@ sources = '''
 # for sd-boot
 fundamental_source_paths = []
 foreach source : sources
-        fundamental_source_paths += meson.current_source_dir() / source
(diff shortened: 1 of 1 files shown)


medhefgo:
After seeing this PR I made #22047, which imho is the correct meson approach (and probably also why libbasic is reproducible). Can you please verify?

kanavin:
Wrong link?

medhefgo:
> Wrong link?

Sorry, fixed.

kanavin:
I verified, it works, and is better indeed - thanks.

PR #22050, opened 2022-01-07
bpf: check if lsm link ptr is libbpf error

jkartseva:
BPF_RAW_TRACEPOINT_OPEN is expected to work only on x86 and x86_64,
since BPF trampoline is implemented only on these architectures.

Attach probing by bpf_program__attach_lsm already happens in
`bpf_lsm_supported`. The resulting pointer can store libbpf error and
that is the case for unsupported architectures.
Add libbpf error check to `bpf_lsm_supported` so execution does not
reach the point where unit startup fails.

### DIFF
--- a/src/core/bpf-lsm.c
+++ b/src/core/bpf-lsm.c
@@ -167,9 +167,9 @@ int lsm_bpf_supported(void) {
         if (r < 0)
                 return supported = 0;
 
-        r = bpf_can_link_lsm_program(obj->progs.restrict_filesystems);
-        if (r < 0) {
-                log_warning_errno(r, "Failed to link BPF program. Assuming BPF is not available: %m");
+        if (!bpf_can_link_lsm_program(obj->progs.restrict_filesystems)) {
+                log_warning_errno(SYNTHETIC_ERRNO(EOPNOTSUPP),
+                                  "Failed to link BPF program. Assuming BPF is not available");
                 return supported = 0;
         }
 


src/core/bpf-lsm.c
```diff
@@ -48,7 +48,7 @@ static bool bpf_can_link_lsm_program(struct bpf_program *prog) {
         if (!link)
                 return -ENOMEM;
 
-        return 1;
+        return sym_libbpf_get_error(link) == 0;
```
bluca, line 52: maybe add a comment explaining why this is done, given it's not very obvious?

bluca (approved):

evverx:
The bpf framework isn't actually tested by the CI on architectures different from x86_64 so I'm not sure why the "waiting-for-ci" label keeps being added to PRs related to it. Looks like @keszybz runs the test on an arm machine so it would probably make sense to wait for @keszybz here

src/core/bpf-lsm.c
keszybz: This still doesn't look right because `libbpf_get_error` already handles NULL:
```C
long libbpf_get_error(const void *ptr)
{
	if (!IS_ERR_OR_NULL(ptr))
		return 0;

	if (IS_ERR(ptr))
		errno = -PTR_ERR(ptr);

	/* If ptr == NULL, then errno should be already set by the failing
	 * API, because libbpf never returns NULL on success and it now always
	 * sets errno on error. So no extra errno handling for ptr == NULL
	 * case.
	 */
	return -errno;
}
```
Thus, we probably want something like this:
```C
link = sym_bpf_program__attach_lsm(prog);
return sym_libbpf_get_error(link) == 0;
```

keszybz: It would be nice to adjust `bpf_link_free` too:
```diff
- if (!link)
+ if (IS_ERR_OR_NULL(link))
```

keszybz: Oh, IS_ERR_OR_NULL is not public. I'll adjust the comment a bit though.

keszybz:
For the sake of time, I force-pushed an update with the suggested change and a small adjustment to `bpf_link_free()`. The spirit of the original change is not changed, so I took the liberty to set the green label.

keszybz:
Arrgh, this fails with assertion error in execute.c:3984, but that error has been fixed in 46004616a1. I'll check if the rebased version works…

wat-ze-hex:
> Arrgh, this fails with assertion error in execute.c:3984, but that error has been fixed in 4600461. I'll check if the rebased version works…

Pre-rebased version fails because the following also requires a fix.
`bpf_can_link_lsm_program` returns `bool` and `r < 0` is never true.
```
        r = bpf_can_link_lsm_program(obj->progs.restrict_filesystems);
        if (r < 0) {
                log_warning_errno(r, "Failed to link BPF program. Assuming BPF is not available: %m");
                return supported = 0;
        }
```

https://github.com/systemd/systemd/blob/main/src/core/bpf-lsm.c#L169L173

Sorry didn't eyeball that when posted this fix.

wat-ze-hex:
> `bpf_can_link_lsm_program` returns `bool` and `r < 0` is never true.

Posted a fix in my fork on top of present two commits: https://github.com/wat-ze-hex/systemd/commit/ccfc534deed2f3873c967851497af10d8a1ee01c

src/core/bpf-lsm.c
jkartseva: And ` return -ENOMEM;` never worked as intended because function return type is boolean...

keszybz:
The CI all passed, except for focal-amd64 and focal-s390x which hasn't even started yet. I'll add the third commit here.

keszybz:
> The bpf framework isn't actually tested by the CI on architectures different from x86_64 so I'm not sure why the "waiting-for-ci" label keeps being added to PRs related to it. Looks like @keszybz runs the test on an arm machine so it would probably make sense to wait for @keszybz here

We do that because the CI still does compilation (so various stupid mistakes would be caught) and static analysis. In this case it's unlikely to catch any non-trivial issues, but it's a good general principle to let any patches that touch code go through the CI.

evverx:
@keszybz sorry. I didn't mean to say that the CI should be ignored here. What I was trying to say is that in addition to it it would be better to actually build and run the test on a machine where the code could be tested before merging it.

mrc0mmand:
> It would be even better if the CI could start covering the bpf framework on various architectures but on Packit `-Dbpf-framework=false` is passed explicitly when systemd is built there and `bpf` is blocked by an nspawn sandbox there (it fails to compile on s390x with [#22067 (comment)](https://github.com/systemd/systemd/pull/22067#issuecomment-1008483248) if it's on FWIW). Ubuntu CI has an out-of-date kernel (in terms of BPF features) and its toolchain fails to build the bpf framework as well: #22051

Yeah, this is really sad. Today I also played with Circle CI, since one of its plans provide aarch64 Linux VMs utilizing AWS Graviton, but they still run Ubuntu Focal, so it wouldn't help us anyway. I'd be willing to play around with any infra which would help us to better cover alternative archs, but so far I can't seem to find any.

keszybz:
> What I was trying to say is that in addition to it it would be better to actually build and run the test on a machine where the code could be tested before merging it.

I've been doing that ;)

evverx:
@keszybz Thanks!  I saw what seems to be a raspberry pi 3 arm machine in one PR and I was sure it could be relied on here too :-)

keszybz:
Yes. The only problem is that it's abysmally slow for any big rebuilds.

Issue #21964, opened 2022-01-01
default route for wireguard interface automatically set when allowedIPs contains ::/0 or 0.0.0.0/0 since v250


freswa:
Last good version: `249.7`
First bad version: `250`

With `249.7` this file doesn't set a default route, with `250` it does.

`wg0.network` config file:
```
[Match]
Name=wg0

[Network]
Address = fd00:dead:beef::1/64
Address = 172.16.42.1/24
IPForward = yes

[Route]
Destination = 172.16.0.0/22
Gateway = 172.16.42.2

[Route]
Destination = 192.168.178.0/24
Gateway = 172.16.42.2

[Route]
Destination = 192.168.80.0/24
Gateway = 172.16.42.2

[Route]
Destination = 2a01:4f8:1c0c:80a0::/64
Gateway = fd00:dead:beef::2
```

yuwata:
Could you provide relevant .netdev file?

yuwata:
I guess, you have `AllowedIPs=0.0.0.0/0` or `AllowedIPs=::/0`. If so, please try to set `RouteTable=off` in [WireGuardPeer] or [WireGuard] section.

freswa:
netdev:
```
[NetDev]
Name = wg0
Kind = wireguard
MTUBytes = 1412

[WireGuard]
PrivateKey = yyy
ListenPort = 51820

[WireGuardPeer]
PublicKey = xxx
AllowedIPs = 0.0.0.0/0
AllowedIPs = ::/0
```

Indeed, `RouteTable = off` fixes the issue. Is this an intentional change in systemd 250?

yuwata:
Since v250, networkd creates routes to `AllowedIPs=` by default. But yeah, I think we should disable that for 0.0.0.0/0 or ::/0.

`wg-quick` does something more complicated and smart way to handle such case.

yuwata:
@zx2c4 Could you give me any suggestions?

zx2c4:
It might not be a bad idea to handle this the way `wg-quick` does. Either that, or handle it normally as is the case now, and expect people to set RouteTable=off who dont want it. Don't special case it to doing nothing, though.

yuwata:
@zx2c4 OK. Thanks.

m29h:
I think this is caused by pull #21553.
[...]

PR #22054, opened 2022-01-08
network: Do NOT add unspecified (::/0, 0.0.0.0/0) wireguard routes

igo95862:
Breaks systems configured before systemd 250.

See #21964.


igo95862:
I have not yet updated documentation but will do soon.

zx2c4:
NACK.

Special casing one route doesn't make sense and isn't the right way to solve this problem, as mentioned in https://github.com/systemd/systemd/issues/21964#issuecomment-1003771183.

zx2c4 (changes requested):
There's no reason to bifurcate the discussion in #21964, where we were discussing this. Please bring things back to there. Strong n'ack on the approach of this PR. Do not merge.

igo95862:
See #22065

PR #22062, opened 2022-01-09
bpf: actually skip RestrictFileSystems= when not supported

keszybz:
Units would fail to start, incl. systemd-journald.service and systemd-udevd.service.
Since unit->manager->restrict_fs will be set if and only if we can use it,
we can just check for that and remove the other checks.
Follow-up for 299d9417238e0727a48ebaabb5a9de0c908ec5c8.

### DIFF
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -1732,21 +1732,6 @@ static int apply_lock_personality(const Unit* u, const ExecContext *c) {
 #endif
 
 #if HAVE_LIBBPF
-static bool skip_lsm_bpf_unsupported(const Unit* u, const char* msg) {
-        assert(u);
-        assert(u->manager);
-
-        if (lsm_bpf_supported())
-                return false;
-
-        /* lsm_bpf_setup succeeded */
-        if (u->manager->restrict_fs)
-                return false;
-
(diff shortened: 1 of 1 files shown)


bluca:
```
FAIL: test-firewall-util20:00
Using nftables as firewall backend.20:00
/* test_v4(backend=nftables) */20:00
/* test_v6(backend=nftables) */20:00
Assertion 'fw_add_masquerade(&ctx, true, AF_INET6, &u3, prefixlen) >= 0' failed at src/test/test-firewall-util.c:34, function test_v6(). Aborting.
```

Does that use the BPF LSM?

anitazha (approved):
So the previous logic was proceeding when LSM BPF was "supported", but failed to check that it was actually set up. This PR makes sense for fixing that.

> Assertion 'fw_add_masquerade(&ctx, true, AF_INET6, &u3, prefixlen) >= 0' failed at src/test/test-firewall-util.c:34, function test_v6(). Aborting.
> 
> Does that use the BPF LSM?

The firewall-util code doesn't use BPF AFAICT. It existed before BPF LSM was added to systemd.

bluca:
I don't remember seeing it fail - but I've restarted it, let's see

bluca:
It passed, so it was a coincidence, let's merge

Issue #22060, opened 2022-01-09
when using sd-boot, systemd-analyze overestimates time in firmware


cmurf:
**systemd version the issue has been seen with**
 > systemd-250.1-1.fc36.x86_64

**Used distribution**
 > Fedora Rawhide

**Linux kernel version used** (`uname -a`)

 > `5.16.0-0.rc8.55.fc36.x86_64`

**CPU architecture issue was seen on**
 > x86_64

**Expected behaviour you didn't see**
 > time in firmware in the milliseconds

**Unexpected behaviour you saw**
 > $ systemd-analyze 
>Startup finished in 28min 31.090s (firmware) + 132ms (loader) + 990ms (kernel) + 1.663s (initrd) + 6.042s (userspace) = 28min 39.918s 
>graphical.target reached after 6.030s in userspace

**Steps to reproduce the problem**
 > boot a VM using sd-boot, then check systemd-analyze

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
…
```

cmurf:
It's definitely always way off
```
# systemd-analyze 
Startup finished in 33min 27.834s (firmware) + 125ms (loader) + 845ms (kernel) + 1.490s (initrd) + 6.038s (userspace) = 33min 36.334s 
# bootctl
System:
     Firmware: UEFI 2.70 (EDK II 1.00)
  Secure Boot: disabled (unsupported)
 TPM2 Support: no
 Boot into FW: supported

Current Boot Loader:
      Product: systemd-boot v250.1-1.fc36
     Features: ✓ Boot counting
               ✓ Menu timeout control
               ✓ One-shot menu timeout control
               ✓ Default entry control
               ✓ One-shot entry control
               ✓ Support for XBOOTLDR partition
               ✓ Support for passing random seed to OS
               ✓ Load drop-in drivers
               ✓ Boot loader sets ESP information
          ESP: /dev/disk/by-partuuid/8013bd90-ede9-425b-8493-5190417d7ed5
         File: └─/EFI/systemd/systemd-bootx64.efi

Random Seed:
 Passed to OS: no
 System Token: not set
       Exists: yes

Available Boot Loaders on ESP:
          ESP: /efi (/dev/disk/by-partuuid/8013bd90-ede9-425b-8493-5190417d7ed5)
         File: └─/EFI/systemd/systemd-bootx64.efi (systemd-boot v250.1-1.fc36)
         File: └─/EFI/BOOT/BOOTIA32.EFI
[...]

PR #22063, opened 2022-01-09
boot-timestamps: Discard firmware init time when running in a VM

medhefgo:
Fixes: #22060

### DIFF
--- a/src/shared/boot-timestamps.c
+++ b/src/shared/boot-timestamps.c
@@ -5,11 +5,13 @@
 #include "efi-loader.h"
 #include "macro.h"
 #include "time-util.h"
+#include "virt.h"
 
 int boot_timestamps(const dual_timestamp *n, dual_timestamp *firmware, dual_timestamp *loader) {
         usec_t x = 0, y = 0, a;
         int r;
         dual_timestamp _n;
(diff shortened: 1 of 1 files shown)


src/shared/boot-timestamps.c
```diff
@@ -33,12 +39,13 @@ int boot_timestamps(const dual_timestamp *n, dual_timestamp *firmware, dual_time
          * the monotonic timestamps here as negative of the actual
          * value. */
 
-        firmware->monotonic = y;
-        loader->monotonic = y - x;
-
-        a = n->monotonic + firmware->monotonic;
-        firmware->realtime = n->realtime > a ? n->realtime - a : 0;
+        if (discard_firmware) {
```
bluca: Did you mean `!discard_firmware`?

medhefgo: You're right. The perils of compile time testing. :D

bluca (approved):

keszybz:
e683878c0f should fix the test failures.

poettering:
I am not sure I follow? Isn't this just a bug in qemu? Why tape over this?

What am I missing?

PR #22065, opened 2022-01-09
networkd: revert rushed wireguard changes

### LINKED ISSUE #21964: default route for wireguard interface automatically set when allowedIPs contains ::/0 or 0.0.0.0/0 since v250

zx2c4:
Right before v250, a bunch of wireguard changes were rushed through without much discussion or consideration. These have caused problems. This PR reverts these changes so that they can all be visited later with more consideration. Feel free to resubmit any of these changes, CC the relevant parties, and we can discuss again. But for now, let's get rid of the half baked stuff.

- WireGuard doesn't support LLv6. Don't try to bolt it on; it won't work. It's been discussed to death on the mailing list and elsewhere. We don't need to have that discussion again here, and I'm not happy the commit forcing it somehow snuck through without me noticing.

- The `RouteTable=` feature not only broke configurations, but it's not even clear how common use cases are that would even benefit from it:
  * Servers with multiple peers don't want multiple /32 routes; they usually want one /24 or /16 route. There's not enough magic in the implementation to handle that.
  * Clients with a 0.0.0.0/0 peer just get a routing loop because there's no fwmark magic.
  * Clients with multiple non-/0 peers might benefit, but nobody's analyzed how common that is yet.

This fixes #21964.

### DIFF
--- a/man/systemd.netdev.xml
+++ b/man/systemd.netdev.xml
@@ -1571,29 +1571,6 @@
           <para>Sets a firewall mark on outgoing WireGuard packets from this interface. Takes a number between 1 and 4294967295.</para>
         </listitem>
       </varlistentry>
-      <varlistentry>
-        <term><varname>RouteTable=</varname></term>
-        <listitem>
-          <para>The table identifier for the routes to the addresses specified in the
-          <varname>AllowedIPs=</varname>. Takes the special value <literal>off</literal>, one of the
-          predefined names <literal>default</literal>, <literal>main</literal>, and
-          <literal>local</literal>, names defined in <varname>RouteTable=</varname> in
-          <citerefentry><refentrytitle>networkd.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry>,
-          or a number in the range 1…4294967295. When <literal>off</literal> the routes to the
-          addresses specified in the <varname>AllowedIPs=</varname> setting will not be configured.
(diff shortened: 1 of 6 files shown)


yuwata:
> WireGuard doesn't support LLv6. Don't try to bolt it on; it won't work. It's been discussed to death on the mailing list and elsewhere. We don't need to have that discussion again here, and I'm not happy the commit forcing it somehow snuck through without me noticing.

Could you provide references about the previous discussions? I only found https://lists.zx2c4.com/pipermail/wireguard/2017-April/001177.html.

keszybz:
As discussed in https://github.com/systemd/systemd/issues/21964#issuecomment-1015228644, those features made it into a release and we'll try to fix them up instead of reverting.

PR #22066, opened 2022-01-09
boot: Change boot entry sorting

medhefgo:
There are a few undesirable properties to how boot entries are
currently sorted.

First, it sorts by entry file name only, which may not correspond
to the title that is shown (for exmaple because it is prefixed by
machine-id). The file ending will also create unexpected ordering
("arch-lts.conf" would come before "arch.conf").

While the list is sorted alphabetically ascending, it is also
lower version/priority first, which is unintuitive. In particular,
a boot-counted entry that is bad (0 tries left) will be at the very
top.

Additionally, the Windows and Mac loaders should be sorted with
the rest of the loaders.

### DIFF
--- a/man/systemd-boot.xml
+++ b/man/systemd-boot.xml
@@ -501,10 +501,10 @@
     considered 'good' from then on.</para>
 
     <para>The boot menu takes the 'tries left' counter into account when sorting the menu entries: entries in 'bad'
-    state are ordered at the beginning of the list, and entries in 'good' or 'indeterminate' at the end. The user can
-    freely choose to boot any entry of the menu, including those already marked 'bad'. If the menu entry to boot is
-    automatically determined, this means that 'good' or 'indeterminate' entries are generally preferred (as the bottom
-    item of the menu is the one booted by default), and 'bad' entries will only be considered if there are no 'good' or
+    state are ordered towards the end of the list, and entries in 'good' or 'indeterminate' towards the beginning.
+    The user can freely choose to boot any entry of the menu, including those already marked 'bad'. If the menu entry
+    to boot is automatically determined, this means that 'good' or 'indeterminate' entries are generally preferred as
+    boot entries are tried in sort order, and 'bad' entries will only be considered if there are no 'good' or
     'indeterminate' entries left.</para>
 
     <para>The <citerefentry><refentrytitle>kernel-install</refentrytitle><manvolnum>8</manvolnum></citerefentry> kernel
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -1492,7 +1492,6 @@ static void config_entry_add_from_file(
 
         entry->device = device;
         entry->id = xstrdup(file);
-        StrLwr(entry->id);
 
         config_add_entry(config, entry);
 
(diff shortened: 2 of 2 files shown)


medhefgo:
Afaik, grub also sorts higher versions first (at least in the fedora vm I quickly checked), so I don't seem to be the only being bothered by how things are currently sorted in sd-boot.

This is a before -> after comparison. 
```
               loader4+0.conf: Loader (4)                     → **AAA
             loader-lts1.conf: Loader (LTS) (1)               → *Loader (3)
             loader-lts2.conf: Loader (LTS) (2)               → Loader (2)
             loader-lts3.conf: Loader (LTS) (3)               → Loader (1)
           loader-other1.conf: Loader (2) (dddddddd)          → Loader (3) (dddddddd)
           loader-other2.conf: Loader (3) (dddddddd)          → Loader (2) (dddddddd)
                 loader1.conf: Loader (1)                     → Loader (LTS) (3)
                 loader2.conf: Loader (2)                     → Loader (LTS) (2)
                 loader3.conf: *Loader (3)                    → Loader (LTS) (1)
                machine1.conf: Machine (22222222)             → Machine (11111111)
                machine2.conf: Machine (11111111)             → Machine (22222222)
                     sss.conf: sss                            → sss
                      tt.conf: ttt                            → ttt
                     ttt.conf: TTT                            → TTT
                     zzz.conf: **AAA                          → Windows 11
                 auto-windows: Windows 11                     → Loader (4)
               auto-efi-shell: EFI Shell                      → EFI Shell
auto-reboot-to-firmware-setup: Reboot Into Firmware Interface → Reboot Into Firmware Interface

 * default loader*
** default ""
```

keszybz:
FWIW, I have been always bothered by the current order. It is unintuitive because the "best" entry is sorted in the middle, between old entries and auto entries. I was considering implementing an option to control the order, but I think it's better to just change it. If people show up who rely on the order, we can consider adding an option…

man/systemd-boot.xml
```diff
@@ -501,11 +501,11 @@
     considered 'good' from then on.</para>
 
     <para>The boot menu takes the 'tries left' counter into account when sorting the menu entries: entries in 'bad'
-    state are ordered at the beginning of the list, and entries in 'good' or 'indeterminate' at the end. The user can
+    state are ordered to the end of the list, and entries in 'good' or 'indeterminate' to the beginning. The user can
```
keszybz: s/to/towards/: "entries are ordered towards the end/begging"

man/systemd-boot.xml
```diff
@@ -501,11 +501,11 @@
...
     freely choose to boot any entry of the menu, including those already marked 'bad'. If the menu entry to boot is
-    automatically determined, this means that 'good' or 'indeterminate' entries are generally preferred (as the bottom
-    item of the menu is the one booted by default), and 'bad' entries will only be considered if there are no 'good' or
-    'indeterminate' entries left.</para>
+    automatically determined, this means that 'good' or 'indeterminate' entries are generally preferred as boot entries
+    are tried in sorting order, and 'bad' entries will only be considered if there are no 'good' or 'indeterminate'
```
keszybz: "in sort order" ?

poettering:
Hmm, I am puzzled by this one? You now compare by title instead of id? But that looks broken?

If you have two entries "quuxos-1.conf" and "quuxos-2.conf" and they have a title of "QuuxOS Wild Rabbit" and "QuuxOS Tamed Tiger", then suddenly you'd assume that quuxos-1.conf was a preferred boot option over quuxos-2.conf? That's just broken.

I am pretty sure we should not order by human strings, but by entry ids, and expect people to pick them sanely.

Issue #11018, opened 2018-12-01
fuzz-compress doesn't seem to fuzz anything


evverx:
Judging by https://storage.googleapis.com/oss-fuzz-coverage/systemd/reports/20181130/linux/src/systemd/src/fuzz/fuzz-compress.c.html#L58, fuzz-compress has never got past `compress` on ClusterFuzz. I suspect that's because both `lz4` and `xz` are excluded when the fuzzers are built. It's possible to link `fuzz-compress` against either or them, but, as far as I know, they should be linked statically as per https://github.com/google/oss-fuzz/blob/9aa11dbd251f393eee0e5d199797d0293e69cd71/docs/fuzzer_environment.md#runtime-dependencies.

evverx:
By the way, built with `lz4` it fails almost immediately with
```sh
#30083	NEW    cov: 36 ft: 38 corp: 7/175b lim: 33 exec/s: 0 rss: 106Mb L: 33/33 MS: 2 ShuffleBytes-CrossOver-
Assertion '*buffer_size == 0 || *buffer' failed at ../../src/systemd/src/journal/compress.c:299, function int decompress_startswith_lz4(const void *, uint64_t, void **, size_t *, const void *, size_t, uint8_t)(). Aborting.
AddressSanitizer:DEADLYSIGNAL
=================================================================
==11==ERROR: AddressSanitizer: ABRT on unknown address 0x00000000000b (pc 0x7f3941d8f428 bp 0x7ffdcd2753d0 sp 0x7ffdcd275268 T0)
SCARINESS: 10 (signal)
    #0 0x7f3941d8f427 in gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x35427)
    #1 0x7f3941d91029 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x37029)
    #2 0x7f394303b40a in log_assert_failed_realm /work/build/../../src/systemd/src/basic/log.c:795:9
    #3 0x7f39431cb0a6 in decompress_startswith_lz4 /work/build/../../src/systemd/src/journal/compress.c:299:9
    #4 0x7f39431cb188 in decompress_startswith /work/build/../../src/systemd/src/journal/compress.c:356:24
    #5 0x530e51 in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/fuzz/fuzz-compress.c:76:13
[...]

PR #22069, opened 2022-01-10
meson: no longer skip dependencies when fuzzers are built locally

evverx:
to make it easier to fuzz code that uses external libraries like libelf/libdw.

The dependencies are skipped on OSS-Fuzz because they aren't available
at runtime if they aren't linked statically. This restriction can safely
be lifted when the fuzzers are built locally with all the dependencies
installed. As far as I know there is at least one fuzz target in the systemd
repository that can benefit from this: https://github.com/systemd/systemd/issues/11018

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -40,7 +40,7 @@ if want_ossfuzz and want_libfuzzer
         error('only one of oss-fuzz or llvm-fuzz can be specified')
 endif
 
-skip_deps = want_ossfuzz or want_libfuzzer
+skip_deps = want_ossfuzz
 fuzzer_build = want_ossfuzz or want_libfuzzer
 
 # Create a title-less summary section early, so it ends up first in the output.


keszybz:
Looks reasonable. I checked the history, and `skip_deps` was added in this form back in 87ac55a12996dbe20ac3f67f2a3b47e650f740e2.

evverx:
> I checked the history, and skip_deps was added in this form back in 87ac55a.

I think `skip_deps` was added in that commit but before that the dependencies had been skipped with `fuzzer_build` set with `fuzzer_build = want_ossfuzz or want_libfuzzer` so as far as I can see that commit didn't change anything in how the fuzzers were build locally on OSS-Fuzz.

evverx:
The idea behind that commit was to turn on the dependencies on FuzzBuzz where it was safe to assume that they were available at runtime.

PR #22070, opened 2022-01-10
Simplify recursive removal code and replace stack recursion with a heap queue

keszybz:


bluca (approved):
Deja-vu :-D

bluca:
focal arm64 and s390x have disappeared from https://autopkgtest.ubuntu.com/running#pkg-systemd-upstream but are still showing as pending on https://api.github.com/repos/systemd/systemd/statuses/5b1cf7a9be37e20133c0208005274ce4a5b5c6a1 so we won't get the result back, no point in waiting

keszybz:
I backported it v249-stable for now.

mbiebl:
Would you consider a backport for v247-stable (as is used in Debian stable)?
This is a non-trivial code change and I feel a bit uneasy doing that myself.

bluca:
> Would you consider a backport for v247-stable (as is used in Debian stable)? This is a non-trivial code change and I feel a bit uneasy doing that myself.

https://github.com/systemd/systemd-stable/pull/155

PR #22072, opened 2022-01-10
xdg-autostart-service: Ignore missing desktop-sepcific condition binary

benzea:
If a desktop specific ExecCondition= binary does not exist, this just
means that the desktop environment is not available. As such, it is not
an error condition that should prevent the service from being installed
in the .wants target.

Fix this by simply returning zero.

### DIFF
--- a/src/xdg-autostart-generator/xdg-autostart-service.c
+++ b/src/xdg-autostart-generator/xdg-autostart-service.c
@@ -485,7 +485,7 @@ static int xdg_autostart_generate_desktop_condition(
                         log_full_errno(r == -ENOENT ? LOG_DEBUG : LOG_WARNING, r,
                                        "%s not found: %m", test_binary);
                         fprintf(f, "# ExecCondition using %s skipped due to missing binary.\n", test_binary);
-                        return r;
+                        return 0;
                 }
 
                 e_autostart_condition = cescape(condition);


keszybz:
Hmm, but as discussed in https://bugzilla.redhat.com/show_bug.cgi?id=2038750, with
```
# /etc/xdg/autostart/ksensors.desktop
[Desktop Entry]
Encoding=UTF-8
Name=KSensors
GenericName=Hardware Monitor
Exec=ksensors --autostart
Icon=ksensors
Type=Application
Comment=Monitor system hardware sensors
Categories=System;Monitor;KDE;
X-KDE-autostart-phase=2
X-KDE-autostart-condition=ksensorsrc:General:AutoStart:true
X-Desktop-File-Install-Version=0.26
```
If I don't have plasma-workspace installed (and thus no `kde-systemd-start-condition`), `app-ksensors@autostart.service` will be started automatically under GNOME. How is this supposed to work in that case?

benzea:
> `app-ksensors@autostart.service` will be started automatically under GNOME. How is this supposed to work in that case?

It'll also (incorrectly) start in the non-systemd startup case. The bug is in ksensors.desktop, which should specify `OnlyShowIn=KDE`, preventing GNOME from picking up the file. And, this check will also work correctly with systemd using the `systemd-xdg-autostart-condition` binary.

kkofler:
Well, in principle, it **should** work if you have an XEmbed system tray host, such as the GNOME Shell TopIcons (Plus) extension (which works only under X11), or a combination of a GNOME Shell AppIndicator extension + the xembedsniproxy from Plasma running. Other X11 desktop environments generally also support XEmbed system tray icons, so I do not see why it should be restricted to KDE/Plasma.

keszybz:
Looking at the patch again, it effectively changes the handling of the autostart conditions from opt-in to opt-out. Before it was "if conditions are defined, they all have to be checked and met", and now it becomes "those conditions that we can actually check, have to be met". One big advantage of the new approach is that it is now easier to support two environments with the same desktop file. You can define AutostartCondition for GNOME and X-KDE-autostart-condition for KDE and on one system the first will be checked and on the other the second. But this is not very reliable, because it only works if the helper binary is not present. If both helper binaries are installed, we'd check both and require both to pass, which doesn't make much sense.

But things will be consistent again if those conditions are changed to triggering conditions. From the man page:

       If multiple conditions are specified, the unit will be executed if all
       of them apply (i.e. a logical AND is applied). Condition checks can use
       a pipe symbol ("|") after the equals sign ("Condition...=|..."), which
       causes the condition to become a triggering condition. If at least one
       triggering condition is defined for a unit, then the unit will be
       started if at least one of the triggering conditions of the unit
       applies and all of the regular (i.e. non-triggering) conditions apply.

So I think we should use (as now) normal conditions for the checks for $XDG_CURRENT_DESKTOP, but one or more triggering conditions for the checks generated from AutostartCondition / X-KDE-autostart-condition.

benzea:
OK, so, I think we have to require that the desktop specific helpers detect whether they are applicable. Now, I think we did not clearly specify how they should behave though.

> But things will be consistent again if those conditions are changed to triggering conditions

As I understand it, `ExecConditon=` works differently. They do not have a "trigger" condition, so that any of the `ExecCondition=` rows failing means that the service will not be started (it works the same as `ExecPreStart=`).

As such, my take is that the utilities *must* detect that they are not applicable and *must* exit successfully then.

Doing this, means we behave in the same way as legacy startup. The `X-KDE-autostart-condition=` and `AutostartCondition=` are desktop specific extensions that are ignored by other desktops. XDG relies on the separate `OnlyShowIn=` (and `NotShowIn=) option to make sure that the service is not started incorrectly on other desktops.

I feel like the solution implemented in this patch is sane. Yes, it is weird and can have unintended side effects (e.g. ksensors starting up in GNOME as it misses `OnlyShowIn=`), but I do believe the behaviour is correct.

keszybz:
You're right, I forgot that ExecCondition is "special".

> As such, my take is that the utilities _must_ detect that they are not applicable and _must_ exit successfully then.

We should document this somewhere. The best place would be systemd-xdg-autostart-generator.8 I think. We should also list what keys are understood and supported.

PR #22075, opened 2022-01-10
build the bpf framework on Packit

evverx:


evverx:
Judging by
```
CONFIGURE_OPTS=(--werror -Dbpf-framework=true ... -Dselinux=true -Dbpf-framework=false -Dapparmor=false
```
it should be appended to CONFIGURE_OPTS instead

mrc0mmand:
The bpf-framework is enabled conditionally in the specfile:

https://src.fedoraproject.org/rpms/systemd/blob/rawhide/f/systemd.spec#_107

so if you'd want to do it unconditionally, using something like

```
sed -i '1i %global want_bpf_framework 1' .packit_rpm/systemd.spec
```

should do the trick.

evverx:
@mrc0mmand I opened this PR to just figure out what's going on on various architectures. Does it make sense to start building systemd with the bpf framework there generally considering the tests are skipped there? I guess it's better than nothing but if Packit is supposed to test the systemd package I'm not sure I should fiddle with it :-)

mrc0mmand:
> @mrc0mmand I opened this PR to just figure out what's going on on various architectures. Does it make sense to start building systemd with the bpf framework there generally considering the tests are skipped there? I guess it's better than nothing but if Packit is supposed to test the systemd package I'm not sure I should fiddle with it :-)

I see, I thought you were onto something which would make the respective tests run :-) I agree that enabling the bpf framework whilst we can't run the tests (so far) is a bit pointless.

evverx:
I think the only way to run those tests there would be to relax the sandbox somehow (or let systemd run its tests without any sandboxes as root). I'm not sure it would make sense for Packit to make an exception in this case though.

mrc0mmand:
> I think the only way to run those tests there would be to relax the sandbox somehow (or let systemd run its tests without any sandboxes as root). I'm not sure it would make sense for Packit to make an exception in this case though.

Yeah, since Copr (the underlying SW Packit is making use of) is meant to be a build system, having an exception for this use case in not very realistic (given the security measures are pretty much hardwired there). For the actual testing with Packit one should utilize the Testing Farm project, but that supports only x86_64.

evverx:
I have to admit I don't even know what the Testing Farm project is. But given that it supports x86_64 only I agree it indeed seems to be a dead end (at this point at least).

evverx:
I had to explicitly add `BuildRequires:  llvm` to install `llvm-strip` on Fedora 34 but other than that and what appears to be a clang bug on Fedora Rawhide s390x it seems the package can be built with the bpf framework.

PR #22078, opened 2022-01-11
Include domain name in canonical name if possible.

iandall:
If hostname does not already include a domain name and /etc/resolv.conv exists
and contains a "domain" or "search" line, then return a canonical name of
"host.domain" otherwise return a canonical name of "host".

The first search domain is assumed to be the principal domain.

### DIFF
--- a/src/nss-myhostname/nss-myhostname.c
+++ b/src/nss-myhostname/nss-myhostname.c
@@ -5,6 +5,9 @@
 #include <netdb.h>
 #include <nss.h>
 #include <stdlib.h>
+#include <sys/stat.h>
+#include <stdio.h>
+#include <pthread.h>
 
 #include "alloc-util.h"
 #include "errno-util.h"
@@ -13,7 +16,9 @@
 #include "macro.h"
 #include "nss-util.h"
 #include "signal-util.h"
-#include "string-util.h"
+#include "stat-util.h"
+#include "fd-util.h"
+#include "fileio.h"
 
 /* We use 127.0.0.2 as IPv4 address. This has the advantage over
  * 127.0.0.1 that it can be translated back to the local hostname. For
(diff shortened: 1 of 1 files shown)


jonesmz:
I'm not completely certain, but I believe that this change would address a fairly annoying problem with using samba / winbind to make a Linux machine an active directory domain member, which is discussed here: https://bugzilla.samba.org/show_bug.cgi?id=14959

The crux of it is that winbindd / samba insist on having the first result from any hostname lookup have the FQDN of the machine, even when the machine might be getting it's IP from DHCP, or a laptop which changes environments.

poettering:
uh, i am not convinced this is the right approach. /etc/resolv.conf might be anything really. And there might not even be "the" fqdn, but multiple. This seems like a very weak approach... i mean, the "search" field in resolv.conf takes a list of domains for a reason...

iandall:
You are right there might not be a single fqdn, none the less "hostname
-f" returns its idea of a "canonical" fqdn and I think this is the
least worst solution. If you have a configuration where there are
multiple fqdns (and you care) then you can't use "hostname -f".

The 'hostname' man page says

   Technically: The FQDN is the name getaddrinfo(3) returns for the host
   name returned by gethostname(2).
   
If systemd-resolved is running, and hostname is a single "label", then
this uses the search option to get the dns part. Ideally, one would
check the domain names in order until a match is found, but this is
just what a resolver does, and the whole point is that this is needed
before systemd-resolved is running.

The use case for this is systems which, for whatever reason, use
dynamic network configuration (either because the systemd image is
shared or because it is an appliance where each instance has an
identical copy of the system image or just because an enterprise wants
to centrally manage its network). In such a situation, the hosts have
to get their domain name from a server somewhere , e.g. by dhcp
"domain" or "domain-search" options. Currently, clients on receiving
this dhcp information, stick it in the search field of /etc/resolv.conf
(or whatever the symbolic link is pointing to), so for the dynamically
configured system, this is as authoritative as it gets. Hypothetically
the dhcp server could be configured with both "domain" and "domain-
search" and the former could be considered the canonical domain and the
latter the resolve search domains. I don't think most dhcp servers are
configured like that though and it would require some other changes in
the client to handle the information. 

It does seem untidy that the "search" option is doing double duty.
 There is no strict requirement that the host be in any of the search
domains, but it is likely to cause problems if it isn't and for
dynamically configured clients, I think it is acceptable to to say the
first listed domain name is the "canonical" one and just accept that as
a limitation. Consider a configuration where the second search domain
name is the canonical one. With the current arrangement, the domain
name goes from null to 2nd depending on the state of the domain
servers. With the proposed patch,  the domain name would go from 1st to
2nd depending on the state of the domain servers. Admittedly, when the
1st domain is used it results in wrong fqdn instead of a truncated
one but I would argue both are broken. If you don't want (or care)
about the domain you use the short hostname. If you ask for a fqdn,
presumably you want a fqdn and not a short hostname.

For statically configured systems, hostname (short and long) would
normally be in /etc/hosts and is is recommended to have "files" ahead
of "myhostname" so the problem wouldn't arise.

Ian

iandall:
> Hypothetically the dhcp server could be configured with both "domain" and "domain- search" and the former could be considered the canonical domain and the latter the resolve search domains.

Looking through the ISC dhclient-script (at least as packaged for Fedora 35), there is the following comment:

                # Note that the DHCP 'Domain Name Option' is really just a domain
                # name, and that this practice of using the domain name option as
                # a search path is both nonstandard and deprecated [in favour of the 'Domain Search' option]

However RFC 2132 describes the 'Domain Name Option' as the domain name that client should use when resolving hostnames via the Domain Name System and I can't find anything in the relevant RFCs which deprecates this usage.

iandall:
Closed by mistake.

poettering:
I am sorry, but I am very convinced that this is not the right appraoch, for the mentioned reasons. This is likely going to break at least as many setups as it fixes. It appears to me, that the Samba issue should be addressed in Samba. You could also work around the whole mess by setting the hostname to an fqdn in the first place, though that sucks for other reasons...

spacefrogg:
Just for reference, because I keep hitting this PR when encountering this problem. The way systemd refuses to deliver at least some FQDN sometimes also breaks all things related to GSSAPI, Kerberos etc. The libraries and applications written against them assume to be able to detect their FQDN via gethostbyaddr*. So, this breaks with current systemd versions.

I do not imply that it is anyone's fault, merely stating that longstanding expectations are no longer met with the way systemd behaves now.

So, while this PR may not be the right approach, a clear interface to define a local notion of an FQDN is important, because, for Kerberos, machines draw their cryptographic identity from it.

iandall:
> Just for reference, because I keep hitting this PR when encountering this problem. The way systemd refuses to deliver at least some FQDN sometimes also breaks all things related to GSSAPI, Kerberos etc. The libraries and applications written against them assume to be able to detect their FQDN via gethostbyaddr*. So, this breaks with current systemd versions.
> 
> EDIT:
> 
> My mitigation for this issue is to change nsswitch.conf to contain:
> 
> ```
> hosts mymachines files resolv dns [!UNAVAIL=return] myhostname
> ```
> 

That is the workaround I discussed in the original pull request and what I am using. Another alternative is to just leave myhostname out altogether, it does nothing useful in my environment* and what it does do it does wrong (but quickly).

\* My environment has network booted hosts which can't work without a reliable network, including a working dns, anyway. 

> Without returning early, the myhostname module actively destroys the earlier result from other modules and returns an > empty domainname although other modules had found one.

I'm not sure about this though. My understanding is that the chain of resolvers terminates when a name is resolved. All the `[!UNAVAILABLE=return]` does is return if a name has NOT been resolved AND dns is available.

PR #22086, opened 2022-01-11
Avoid looking at /prcc/cmdline from nss modules

keszybz:


yuwata (approved):
LGTM.

bluca (approved):

poettering:
Hmm, this still doesn't look right to me: if one of our own programs ends up calling into NSS and ends up with one of our own modules we'll parse the env vars twice...

I have the suspicion the pthread_once() stuff should actually be moved into log.c and shared across the whole tree, so that it also applies to log initializations from our usual main.c functions...

PR #19276, opened 2021-04-10
units: install dbus-org.freedesktop.oom1.service alias

anitazha:
Fixes: #18469

Issue #18469, opened 2021-02-04
TEST-55-OOMD fails sporadically due to non-existent dbus service


mrc0mmand:
**systemd version the issue has been seen with**
 > latest master

**Used distribution**
 > Arch Linux

Several of the recent runs failed because of missing `dbus-org.freedesktop.oom1.service` in `TEST-56-OOMD`:

```
[   14.628829] kernel: audit: type=1130 audit(1612431027.616:98): pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=testsuite-56-testbloat comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
[   14.511995] testsuite-56.sh[232]: + grep /testsuite-56-workload.slice
[   14.515917] testsuite-56.sh[231]: + oomctl
[   14.595242] dbus-daemon[222]: [system] Activating via systemd: service name='org.freedesktop.oom1' unit='dbus-org.freedesktop.oom1.service' requested by ':1.2' (uid=0 pid=231 comm="oomctl ")
[   14.915309] dbus-daemon[222]: [system] Activation via systemd failed for unit 'dbus-org.freedesktop.oom1.service': Unit dbus-org.freedesktop.oom1.service not found.
[   14.915457] testsuite-56.sh[231]: Failed to dump context: Unit dbus-org.freedesktop.oom1.service not found.
```

Logs:
 * console: https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/2521/artifact//systemd-centos-ci/artifacts_LhSnWH/vagrant-logs.gib/vagrant-arch-testsuite.PMn/TEST-56-OOMD_FAIL.log
 * journal: https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/2521/artifact//systemd-centos-ci/artifacts_LhSnWH/vagrant-logs.gib/vagrant-arch-testsuite.PMn/TEST-56-OOMD/system.journal

I'm not quite sure what exactly is going wrong, but it's kind of weird that meson doesn't create the `dbus-*.service` symlink for the oomd service like it does for other services, e.g.:

logind:
```
...
Installing /build/src/login/org.freedesktop.login1.conf to /var/tmp/systemd-test-TEST-01-BASIC/root/usr/share/dbus-1/system.d
[...]

PR #22090, opened 2022-01-12
units: don't install dbus-org.freedesktop.oom1.service alias

mbiebl:
The systemd-oomd.service unit contains
```
[Install]
WantedBy=multi-user.target
Alias=dbus-org.freedesktop.oom1.service
```
which means the symlink is supposed to be created dynamically when the
service is enabled.

### DIFF
--- a/test/units/testsuite-55.sh
+++ b/test/units/testsuite-55.sh
@@ -35,6 +35,10 @@ echo -e "[Service]\nEnvironment=SYSTEMD_LOG_LEVEL=debug" >/etc/systemd/system/sy
 
 systemctl daemon-reload
 
+# enable the service to ensure dbus-org.freedesktop.oom1.service exists
+# and D-Bus activation works
+systemctl enable systemd-oomd.service
+
 # if oomd is already running for some reasons, then restart it to make sure the above settings to be applied
 if systemctl is-active systemd-oomd.service; then
     systemctl restart systemd-oomd.service


mrc0mmand:
Hmm, this goes against the original PR https://github.com/systemd/systemd/pull/19276 - I wonder which one is the correct one in this case.

mbiebl:
> Hmm, this goes against the original PR #19276 - I wonder which one is the correct one in this case.

If systemd-oomd is supposed to be a service which can be enabled/disabled (which judging by the `[Install]` section, it is), then I think this PR is correct.
Maybe #18469 could be fixed by running `systemctl enable systemd-oomd.service` if it needs D-Bus activation?

mbiebl:
```
diff --git a/test/units/testsuite-55.sh b/test/units/testsuite-55.sh
index 379ea9e..d1e740a 100755
--- a/test/units/testsuite-55.sh
+++ b/test/units/testsuite-55.sh
@@ -29,6 +29,10 @@ echo -e "[Service]\nEnvironment=SYSTEMD_LOG_LEVEL=debug" >/etc/systemd/system/sy
 
 systemctl daemon-reload
 
+# enable the service to ensure dbus-org.freedesktop.oom1.service exists
+# and D-Bus activation works
+systemctl enable systemd-oomd.service
+
 # if oomd is already running for some reasons, then restart it to make sure the above settings to be applied
 if systemctl is-active systemd-oomd.service; then
     systemctl restart systemd-oomd.service
```
@mrc0mmand , @anitazha WDYT?

yuwata (approved):
LGTM.

mrc0mmand (approved):
LGTM too.

bluca (approved):

keszybz:
LGTM too.

PR #22093, opened 2022-01-12
meson: require bpftool version >= 5.6

### LINKED ISSUE #22051: The bpf framework requires 'bpftool gen'
For the bpf framework to work bpftool has to support the gen command. Without it `ninja` fails to build `systemd` with
```
[462/1877] /home/vagrant/systemd/tools/build-bpf-skel.py --clang_exec /usr/bin/clang --llvm_strip_exec /usr/bin/llvm-strip --bpftool_exec /usr/sbin/bpftool --arch x86_64 ../src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c src/core/bpf/restrict_ifaces/restrict-ifaces.skel.h
FAILED: src/core/bpf/restrict_ifaces/restrict-ifaces.skel.h
/home/vagrant/systemd/tools/build-bpf-skel.py --clang_exec /usr/bin/clang --llvm_strip_exec /usr/bin/llvm-strip --bpftool_exec /usr/sbin/bpftool --arch x86_64 ../src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c src/core/bpf/restrict_ifaces/restrict-ifaces.skel.h
Usage: /usr/lib/linux-tools/5.4.0-92-generic/bpftool [OPTIONS] OBJECT { COMMAND | help }
       /usr/lib/linux-tools/5.4.0-92-generic/bpftool batch file FILE
       /usr/lib/linux-tools/5.4.0-92-generic/bpftool version

       OBJECT := { prog | map | cgroup | perf | net | feature | btf }
       OPTIONS := { {-j|--json} [{-p|--pretty}] | {-f|--bpffs} |
	            {-m|--mapcompat} | {-n|--nomount} }
  File "/home/vagrant/systemd/tools/build-bpf-skel.py", line 128, in <module>
    bpf_build(args)
  File "/home/vagrant/systemd/tools/build-bpf-skel.py", line 92, in bpf_build
    gen_bpf_skeleton(bpftool_exec=args.bpftool_exec,
  File "/home/vagrant/systemd/tools/build-bpf-skel.py", line 63, in gen_bpf_skeleton
    skel = subprocess.check_output(bpftool_args, universal_newlines=True)
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/usr/sbin/bpftool', 'g', 's', '../src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.o']' returned non-zero exit status 255.
```

It was discovered in https://github.com/systemd/systemd/pull/22021#issuecomment-1007918469

yuwata:
Closes #22051.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -990,20 +990,32 @@ if want_bpf_framework == 'false'
 else
         # Support 'versioned' clang/llvm-strip binaries, as seen on Debian/Ubuntu
         # (like clang-10/llvm-strip-10)
-        clang_bin = cc.get_id() == 'clang' ? cc.cmd_array()[0] : 'clang'
-        if meson.is_cross_build() or clang_bin.contains('afl-clang') or clang_bin.contains('hfuzz-clang')
-                clang_bin = 'clang'
+        if meson.is_cross_build() or cc.get_id() != 'clang' or cc.cmd_array()[0].contains('afl-clang') or cc.cmd_array()[0].contains('hfuzz-clang')
+                r = find_program('clang', required : bpf_framework_required)
+                clang_found = r.found()
+                if clang_found
+                        if meson.version().version_compare('>= 0.55')
+                                clang = [r.full_path()]
+                        else
+                                clang = [r.path()]
+                        endif
+                endif
+                # Assume that the required flags are supported by the found clang.
+                clang_supports_flags = clang_found
+        else
+                clang_found = true
+                clang = cc.cmd_array()
+                clang_supports_flags = cc.has_argument('-Wno-compare-distinct-pointer-types')
         endif
-        clang = find_program(clang_bin, required : bpf_framework_required)
 
-        if clang.found()
+        if clang_found
                 # Check if 'clang -target bpf' is supported.
                 clang_supports_bpf = run_command(clang, '-target', 'bpf', '--print-supported-cpus', check : false).returncode() == 0
         else
                 clang_supports_bpf = false
         endif
 
-        if not meson.is_cross_build() and clang.found()
+        if not meson.is_cross_build() and clang_found
                 llvm_strip_bin = run_command(clang, '--print-prog-name', 'llvm-strip',
                                              check : true).stdout().strip()
         else
@@ -1019,7 +1031,8 @@ else
                                required : bpf_framework_required,
                                version : '>= 5.6')
 
-        deps_found = libbpf.found() and clang.found() and clang_supports_bpf and llvm_strip.found() and bpftool.found()
+        deps_found = libbpf.found() and clang_found and clang_supports_bpf and clang_supports_flags and llvm_strip.found() and bpftool.found()
+
         # Can build BPF program from source code in restricted C
         conf.set10('BPF_FRAMEWORK', deps_found)
 endif


yuwata:
cc @evverx and @wat-ze-hex

keszybz:
LGTM.

evverx:
Thanks! With this PR applied I tried to build systemd in an Ubuntu Focal container and it failed to compile there with
```
Program bpftool found: YES (/usr/sbin/bpftool)

meson.build:1009:8: ERROR: Running ['/usr/sbin/bpftool', '--version'] failed
```
```
Program /usr/bin/llvm-strip found: YES (/usr/bin/llvm-strip)
Program bpftool found: YES (/usr/sbin/bpftool)
Running command: /usr/sbin/bpftool --version
--- stdout ---

--- stderr ---
WARNING: bpftool not found for kernel 5.4.0-92

  You may need to install the following packages for this specific kernel:
    linux-tools-5.4.0-92-generic
    linux-cloud-tools-5.4.0-92-generic

  You may also want to install one of the following packages to keep up to date:
    linux-tools-generic
    linux-cloud-tools-generic



meson.build:1009:8: ERROR: Running ['/usr/sbin/bpftool', '--version'] failed
```
I'm not sure where `bpftool` got lost

evverx:
I've just installed `linux-tools-5.4.0-92-generic` to really install `bpftool` and with it the bpf framework seems to be skipped now. But I'm not sure meson should fail hard if `bpftool` points to a script exiting with 1

evverx:
FWIW looking for `clang` isn't enough either because `-target bpf` can be turned off
```
# clang -target bpf --print-supported-cpus
clang version 14.0.0 (https://github.com/llvm/llvm-project.git aee49255074fd4ef38d97e6e70cbfbf2f9fd0fa7)
Target: bpf
Thread model: posix
InstalledDir: /usr/local/bin
No available targets are compatible with triple "bpf"
```
so I think instead of looking for just `clang` and `bpftool` it would be better to actually try to build a simple bpf skeleton and skip the bpf framework if it fails (by analogy with how `meson` makes sure that the compiler and linker work with certain flags).

yuwata:
@evverx Added more checks for clang. PTAL.

evverx:
@yuwata I rebased this PR on top of https://github.com/systemd/systemd/pull/22103 and according to the "unit tests" workflow failed with
```
2022-01-13T07:58:25.1869514Z --- stderr ---
2022-01-13T07:58:25.1869822Z clang: error: unknown argument: '-print-targets'
2022-01-13T07:58:25.1870065Z clang: error: no input files
2022-01-13T07:58:25.1870200Z 
2022-01-13T07:58:25.1870296Z 
2022-01-13T07:58:25.1870389Z 
2022-01-13T07:58:25.1870747Z meson.build:1004:16: ERROR: Command "/usr/bin/clang -print-targets" failed with status 1.
2022-01-13T07:58:25.1881879Z ##[error]Process completed with exit code 1.
```
and the "build" workflow failed with
```
2022-01-13T07:58:50.7542057Z Run-time dependency libbpf found: YES 0.4.0
2022-01-13T07:58:50.7542499Z Program clang-11 found: YES (/usr/bin/clang-11)
2022-01-13T07:58:50.7542672Z 
2022-01-13T07:58:50.7542922Z meson.build:999:16: ERROR: Function does not take keyword arguments.
2022-01-13T07:58:50.7558067Z ##[error]Process completed with exit code 1.
2022-01-13T07:58:50.7621209Z Post job cleanup.
```

meson.build
```diff
@@ -1005,9 +1016,14 @@ else
 
         # Debian installs this in /usr/sbin/ which is not in $PATH.
         # We check for 'bpftool' first, honouring $PATH, and in /usr/sbin/ for Debian.
-        bpftool = find_program('bpftool', '/usr/sbin/bpftool', required : bpf_framework_required)
+        # We use 'bpftool gen' subcommand, it was added by 985ead416df39d6fe8e89580cc1db6aa273e0175 (v5.6).
+        bpftool = find_program('bpftool',
+                               '/usr/sbin/bpftool',
+                               required : bpf_framework_required,
```
evverx, line 1031: Looks like `find_program` fails when `bpftool` exits with 1 on Ubuntu Focal even if `required` is set to `false` here. It's probably better than https://github.com/systemd/systemd/issues/22051 but I don't think it should happen if `-Dbpf-framework` wasn't set to `true` explicitly. There doesn't seem to be `check: false` that could be passed to `find_program` :-(

yuwata: Could you report that to meson team? It seems that we have no way to workaround the issue.

evverx: I'm not sure whether it's a meson bug or not. It seems to me that it's reasonable to fail when `cmd --version` fails. I have no idea why that script is installed there in the first place and why `meson` should get it around. I mean, if `bpftool` isn't actually there that script shouldn't be there either.

evverx: > It seems that we have no way to workaround the issue

As far as I understand it should be possible to call `find_program('bpftool')` and then pass the result to `run_command` to see if it fails and pass it again to `find_program(..., version)` if it doesn't

yuwata: Ah, I understand the issue. ubuntu installs bpftool as a tiny wrapper... But IIUC, the error is triggered only when the matching real bpftool does not installed. I do not think we should not handle such issue here, as it is a build host issue.

evverx: @yuwata let's wait for @wat-ze-hex. I think if there is a relatively easy way to somehow build a small bpf skeleton with `clang -target bpf`, `llvm-strip` and `bpftool gen` it would be better than looking for all the tools independently to just find out they can't work together.

evverx: That would be similar to compiling and linking `int main() { return 0; }` to make sure the compiler and linker can work together.

yuwata: Yeah.

meson.build
```diff
@@ -990,11 +990,22 @@ if want_bpf_framework == 'false'
...
         # (like clang-10/llvm-strip-10)
-        clang_bin = cc.get_id() == 'clang' ? cc.cmd_array()[0] : 'clang'
-        if meson.is_cross_build() or clang_bin.contains('afl-clang') or clang_bin.contains('hfuzz-clang')
-                clang_bin = 'clang'
+        if meson.is_cross_build() or cc.get_id() != 'clang' or cc.cmd_array()[0].contains('afl-clang') or cc.cmd_array()[0].contains('hfuzz-clang')
+                clang = find_program('clang', required : bpf_framework_required)
+                # Assume that the required flags are supported.
+                clang_supports_flags = true
+        else
+                clang = find_program(cc.cmd_array()[0], required : bpf_framework_required)
```
eli-schwartz: What should this do when cc.cmd_array() is `['ccache', 'clang']`?

evverx: @eli-schwartz it shouldn't rely on `cc.cmd_array()[0]` I think :-) That's a good point. Thanks!

eli-schwartz: Quite so. ;)

Honestly, you can just use the entire cmd_array() directly as the found compiler, if you need to use it directly anywhere as a custom_target or run_target command. A found program and a cmd_array() are going to be interchangeable.

And setting `required: bpf_framework_required` (a bool value) is pointless, since the `meson.get_compiler('c')` found compiler will always exist, and can never not be found. So, you can set `clang_found = clang.found()` in the if, and unconditionally set `clang_found = true` in the else... and key off of that everywhere.

yuwata:
@eli-schwartz Thank you for the comment. Applied your suggestion. PTAL.

keszybz:
Looks reasonable… If it doesn't really work, we'll find out soon enough.

evverx:
Looks like it broke LGTM: https://lgtm.com/projects/g/systemd/systemd/logs/rev/pr-18363806178f819c04e3769f06d232c767d20520/lang:cpp/stage:Build%20merge_276b59423b40f518d21d8bce4826db7d6508bd64
```
[2022-01-18 22:51:20] [build-stdout] Found CMake: /usr/bin/cmake (3.13.4)
[2022-01-18 22:51:20] [build-stdout] Run-time dependency libbpf found: NO (tried pkgconfig and cmake)
[2022-01-18 22:51:20] [build-stdout] Program clang found: YES (/usr/bin/clang)
[2022-01-18 22:51:20] [build-stdout] Program /usr/bin/llvm-strip found: YES (/usr/bin/llvm-strip)
[2022-01-18 22:51:20] [build-stdout] meson.build:1028:8: ERROR: Running ['/usr/sbin/bpftool', '--version'] failed
[2022-01-18 22:51:20] [build-stdout] A full log can be found at /opt/src/build/meson-logs/meson-log.txt
[2022-01-18 22:51:21] [build-stdout] /opt/dist/cpp/tools/configure-wrapper: Configure failed
[2022-01-18 22:51:21] [build-stderr] + cd .
```

evverx:
I opened https://github.com/systemd/systemd/pull/22169 (where the bpf framework is off) to figure out whether it's really the culprit.

yuwata:
Does it mean the installed bpftool does not match with the currently installed (or running?) kernel?

Which distro does LGTM use? Ubuntu?

yuwata:
Are there no way to install the correct version of bpftool?

evverx:
LGTM uses Ubuntu and it installs the build dependencies (including bpftool) automagically but I don't think it's clever enough to figure out that bpftool on Ubuntu should be kind of installed twice to really make it work. But considering that `libbpf` isn't there either I'm not sure why systemd should try to look for bpftool

yuwata:
Indeed.
```
[2022-01-18 22:51:20] [build-stdout] Run-time dependency libbpf found: NO (tried pkgconfig and cmake)
```

PR #22095, opened 2022-01-12
core,cgroup-show: ignore -EOPNOTSUPP in cg_read_pid()

### LINKED ISSUE #21945: systemctl status shows wrong cgroup tree
>systemd 250 (250-4-arch)
>+PAM +AUDIT -SELINUX -APPARMOR -IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK +XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified

**Used distribution**
 > Arch Linux

 > `Linux enterprise 5.15.12-zen1-1-zen #1 ZEN SMP PREEMPT Wed, 29 Dec 2021 12:04:52 +0000 x86_64 GNU/Linux`

 > x86-64

 >  `systemctl status` shows tree of slices/scopes/sessions/services

image

After updating from 249.7, `systemctl status` started showing a flat process list under `net_cls`.  As far as I know, my VPN software [1] uses `net_cls` cgroups to control split tunneling.
```
$ mount | grep cgroup
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot)
net_cls on /sys/fs/cgroup/net_cls type cgroup (rw,relatime,net_cls)
```
If I stop VPN service and unmount `/sys/fs/cgroup/net_cls`, issue is (temporarily) fixed. 
`systemd-cgls` is unaffected. `systemctl status --user` works fine as well.

[1] https://github.com/mullvad/mullvadvpn-app

### LINKED ISSUE #22089: "systemctl status" Regression when running libvirt VM
 > systemd 250 (250.2-2-arch)

**Used distribution**
 > Arch

 > `Linux ryzen 5.15.13-arch1-1 #1 SMP PREEMPT Wed, 05 Jan 2022 16:20:59 +0000 x86_64 GNU/Linux`

 > x86_64

 > normal output from "systemctl status"

 > systemctl status
> Failed to dump process list for 'ryzen', ignoring: Operation not supported

 > fire up any libvirt VM
run systemctl status

Not sure if this is a libvirt or a systemd issue. It only started as of systemd-2.50.

Obvious suspect:
https://github.com/systemd/systemd/pull/21172

Arch Forum report:
https://bbs.archlinux.org/viewtopic.php?id=272962

Thanks

yuwata:
Fixes #22089.

### DIFF
--- a/src/core/dbus-unit.c
+++ b/src/core/dbus-unit.c
@@ -1312,11 +1312,15 @@ static int append_cgroup(sd_bus_message *reply, const char *p, Set *pids) {
         for (;;) {
                 pid_t pid;
 
+                /* libvirt / qemu uses threaded mode and cgroup.procs cannot be read at the lower levels.
+                 * From https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#threads,
+                 * “cgroup.procs” in a threaded domain cgroup contains the PIDs of all processes in
+                 * the subtree and is not readable in the subtree proper. */
                 r = cg_read_pid(f, &pid);
+                if (IN_SET(r, 0, -EOPNOTSUPP))
+                        break;
                 if (r < 0)
                         return r;
-                if (r == 0)
-                        break;
 
                 if (is_kernel_thread(pid) > 0)
                         continue;
--- a/src/shared/cgroup-show.c
+++ b/src/shared/cgroup-show.c
@@ -89,7 +89,6 @@ static int show_cgroup_one_by_path(
         _cleanup_fclose_ FILE *f = NULL;
         _cleanup_free_ char *p = NULL;
         size_t n = 0;
-        pid_t pid;
         char *fn;
         int r;
 
@@ -102,7 +101,18 @@ static int show_cgroup_one_by_path(
         if (!f)
                 return -errno;
 
-        while ((r = cg_read_pid(f, &pid)) > 0) {
+        for (;;) {
+                pid_t pid;
+
+                /* libvirt / qemu uses threaded mode and cgroup.procs cannot be read at the lower levels.
+                 * From https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#threads,
+                 * “cgroup.procs” in a threaded domain cgroup contains the PIDs of all processes in
+                 * the subtree and is not readable in the subtree proper. */
+                r = cg_read_pid(f, &pid);
+                if (IN_SET(r, 0, -EOPNOTSUPP))
+                        break;
+                if (r < 0)
+                        return r;
 
                 if (!(flags & OUTPUT_KERNEL_THREADS) && is_kernel_thread(pid) > 0)
                         continue;
@@ -112,9 +122,6 @@ static int show_cgroup_one_by_path(
 
                 pids[n++] = pid;
         }
-
-        if (r < 0)
-                return r;
 
         show_pid_array(pids, n, prefix, n_columns, false, more, flags);
 


mrc0mmand:
I tried to reproduce both mentioned issues and so far everything works as expected, thanks!

yuwata:
@mrc0mmand Thank you for testing this.

mrc0mmand (approved):

src/basic/cgroup-util.c
```diff
@@ -211,6 +211,17 @@ int cg_read_subgroup(DIR *d, char **fn) {
                 if (dot_or_dot_dot(de->d_name))
                         continue;
 
+                if (ignore_net_cls) {
+                        _cleanup_free_ char *p = NULL;
+
+                        p = path_join(de->d_name, "net_cls.classid");
```
keszybz: This doesn't look right. The bug is about people mount cgroup-v1 controllers in a directory below the cgroup-v2 in a hierarchy. In particular, any v1 controller will behave like this, not just net_cls.

This can never work — as described in the cgroup delegation documents, systemd expects to have exclusive ownership of the hierarchy (except where delegation is enabled). I think we should close this as CANTFIX.

src/core/dbus-unit.c
```diff
@@ -1309,14 +1309,7 @@ static int append_cgroup(sd_bus_message *reply, const char *p, Set *pids) {
...
 
-        for (;;) {
-                pid_t pid;
-
-                r = cg_read_pid(f, &pid);
-                if (r < 0)
-                        return r;
-                if (r == 0)
-                        break;
+        for (pid_t pid; cg_read_pid(f, &pid) > 0;) {
```
keszybz: I think this is removes too much of the check.

libvirt / qemu uses threaded mode (https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#threads), and cgroup.procs cannot be read at the lower levels:

>  “cgroup.procs” in a threaded domain cgroup contains the PIDs of all processes in the subtree and is not readable in the subtree proper.

Please change this to only accept EOPNOTSUPP, and add a comment why.

keszybz (changes requested):

src/basic/cgroup-util.c
yuwata: Ah, I did not know that net_cls is a cgroup v1 controller... Thanks.

yuwata:
@keszybz Thank you for the review. Addressed the comments. PTAL.

keszybz:
LGTM, thanks.

poettering:
I think we should ignore threaded cgroups entirely. They are not relevant for systemd's cgroup logic.

keszybz:
Do you mean that we should just threading mode on cgroups? I guess we could, but threaded mode is very very rarely used, and we'd pay the price for the extra check all the time. (A while back I queried codesearch.debian.net, and libvirt was the *only* user of the API listed there.)

poettering:
threading mode is a bad idea, I would try hard to skip these cgroups early on. I mean, yes, checking this explicitly might be expensive, but maybe we should detect this case explicitly at least, and handle it as such. Specifically:

in cg_read_pid() if we see EOPNOTSUPP() use that as hint that we might be looking at a thread cgroup, and if so, verify that and return some recognizable error (EUCLEAN or so), and then catch tat in the callers, and use it as hint to really fully ignore the whole cgroup.

bobrik:
Could it be more than just `status` regression? We're also seeing failed `reload` for the affected services:

```
Feb 10 23:17:12 nope systemd[1]: heh.service: Control process exited, code=exited, status=219/CGROUP
Feb 10 23:17:12 nope systemd[1]: Reload failed for heh.
```

bobrik:
Syscalls:

```
stat("/sys/fs/cgroup/system.slice/nginx-fl.service", {st_mode=S_IFDIR|0755, st_size=0, ...}) = 0
mkdir("/sys/fs/cgroup/system.slice/nginx-fl.service/.control", 0755) = -1 EEXIST (File exists)
removexattr("/sys/fs/cgroup/system.slice/nginx-fl.service/.control", "user.oomd_avoid") = -1 ENODATA (No data available)
removexattr("/sys/fs/cgroup/system.slice/nginx-fl.service/.control", "user.oomd_omit") = -1 ENODATA (No data available)
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fe46a98ec10) = 317371
openat(AT_FDCWD, "/sys/fs/cgroup/system.slice/nginx-fl.service/.control/cgroup.procs", O_WRONLY|O_NOCTTY|O_CLOEXEC) = 303
fcntl(303, F_GETFL)                     = 0x8001 (flags O_WRONLY|O_LARGEFILE)
write(303, "317371\n", 7)               = -1 EOPNOTSUPP (Operation not supported)
```

I think this is the stack as viewed by perf:

```
systemd     1 [112]  3104.441355: syscalls:sys_enter_write: fd: 0x0000012f, buf: 0x7ffd1d92648a, count: 0x00000007
            7fe46b139f6f write+0x4f (/usr/lib/x86_64-linux-gnu/libc-2.31.so)
            7fe46b0cb664 _IO_file_write+0x24 (/usr/lib/x86_64-linux-gnu/libc-2.31.so)
            7fe46b0ca9d5 [unknown] (/usr/lib/x86_64-linux-gnu/libc-2.31.so)
            7fe46b0cbd5d _IO_file_xsputn+0x15d (/usr/lib/x86_64-linux-gnu/libc-2.31.so)
            7fe46b0bffc6 _IO_fputs+0xa6 (/usr/lib/x86_64-linux-gnu/libc-2.31.so)
            7fe46b49186c write_string_stream_ts+0x7c (/usr/lib/systemd/libsystemd-shared-249.so)
            7fe46b491c31 write_string_file_ts+0x1d1 (/usr/lib/systemd/libsystemd-shared-249.so)
            7fe46b3edd09 cg_attach.localalias+0x99 (inlined)
            55ba458458d0 exec_spawn+0x580 (/usr/lib/systemd/systemd)
            55ba4587f282 service_spawn+0x692 (/usr/lib/systemd/systemd)
            55ba458848b1 service_reload+0xb1 (/usr/lib/systemd/systemd)
            55ba458495e7 job_perform_on_unit+0xd7 (/usr/lib/systemd/systemd)
            55ba4584ca61 job_run_and_invalidate+0x211 (/usr/lib/systemd/systemd)
            55ba4586762f manager_dispatch_run_queue+0x3f (/usr/lib/systemd/systemd)
            7fe46b557d84 source_dispatch+0x244 (/usr/lib/systemd/libsystemd-shared-249.so)
            7fe46b5582ec sd_event_dispatch+0x11c (/usr/lib/systemd/libsystemd-shared-249.so)
            7fe46b558a1f sd_event_run+0xbf (/usr/lib/systemd/libsystemd-shared-249.so)
            55ba4586e641 manager_loop+0xf51 (/usr/lib/systemd/systemd)
            55ba458183de main+0x7de (/usr/lib/systemd/systemd)
            7fe46b071d09 __libc_start_main+0xe9 (/usr/lib/x86_64-linux-gnu/libc-2.31.so)
            55ba4581ae09 _start+0x29 (/usr/lib/systemd/systemd)
```

bobrik:
I filed #22486 with more details.

PR #22096, opened 2022-01-12
Open the bus once in networkctl

keszybz:


src/basic/macro.h
```diff
@@ -254,6 +254,14 @@ static inline int __coverity_check_and_return__(int condition) {
 #define assert(expr) assert_message_se(expr, #expr)
 #endif
 
+/* This passes the argument through after (if asserts are enabled) checking that it is not null. */
+#define ASSERT_PTR(expr)                        \
```
medhefgo: Maybe put that in macro-fundamental. It could replace assert_alloc.

yuwata (approved):
LGTM.

src/basic/macro.h
yuwata: Moved.

yuwata:
The macro is moved to fundamental. Upgrading the green label.

keszybz:
Thanks.

PR #22098, opened 2022-01-12
journal: Fixes for handling of corrupt entry objects

daandemeyer:
Previously, we'd try to handle corruption by bumping the index even
if it was an entry array object that was corrupted (which we can't
deal with).

Now, we only try to deal with corrupted entry objects by moving the
corruption handling into generic_array_get().

On top, we also add an additional check for -EADDRNOTAVAIL which can
also be caused by corrupted journal data.

### DIFF
--- a/src/libsystemd/sd-journal/journal-file.c
+++ b/src/libsystemd/sd-journal/journal-file.c
@@ -2099,14 +2099,35 @@ static void chain_cache_put(
         ci->last_index = last_index;
 }
 
+static int bump_array_index(uint64_t *i, direction_t direction, uint64_t n) {
+        assert(i);
+
+        /* Increase or decrease the specified index, in the right direction. */
+
+        if (direction == DIRECTION_DOWN) {
+                if (*i >= n - 1)
+                        return 0;
+
+                (*i)++;
+        } else {
+                if (*i <= 0)
+                        return 0;
+
+                (*i)--;
+        }
+
+        return 1;
+}
+
 static int generic_array_get(
                 JournalFile *f,
                 uint64_t first,
                 uint64_t i,
+                direction_t direction,
                 Object **ret, uint64_t *ret_offset) {
 
-        Object *o;
-        uint64_t p = 0, a, t = 0;
+        Object *o, *e;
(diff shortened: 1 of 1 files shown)


src/libsystemd/sd-journal/journal-file.c
```diff
@@ -2130,16 +2150,50 @@ static int generic_array_get(
...
+                if (k == 0)
+                        break;
+
+                if (i == UINT64_MAX)
+                        i = direction == DIRECTION_DOWN ? 0 : k - 1;
+
+                do {
+                        p = le64toh(o->entry_array.items[i]);
+
+                        r = journal_file_move_to_object(f, OBJECT_ENTRY, p, &e);
```
yuwata, line 2181: The call of `journal_file_move_to_object()` below is not necessary any more because of the call here?

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -2130,16 +2150,50 @@ static int generic_array_get(
...
+         * direction). */
+
+        while (a > 0) {
+                uint64_t k;
+
+                r = journal_file_move_to_object(f, OBJECT_ENTRY_ARRAY, a, &o);
+                if (r < 0)
+                        return r;
+
+                k = journal_file_entry_array_n_items(o);
```
yuwata: The initial calls of these functions are redundant. But that is OK if these are not so heavy.

daandemeyer: Changed to not execute these on the first iteration of the while loop.

yuwata (approved):
LGTM.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -2099,14 +2099,34 @@ static void chain_cache_put(
         ci->last_index = last_index;
 }
 
+static int bump_array_index(uint64_t *i, direction_t direction, uint64_t n) {
+
```
bluca, line 2104: `assert i`

bluca (approved):

yuwata:
The change from the previous version is trivial, merging.

PR #22100, opened 2022-01-12
Fix TEST-58-REPART on ppc64el

bluca:
TEST-58-REPART started running on ppc64el, but we have hard-coded x86-64 uuids so it fails

### DIFF
--- a/test/units/testsuite-58.sh
+++ b/test/units/testsuite-58.sh
@@ -11,6 +11,60 @@ fi
 export SYSTEMD_LOG_LEVEL=debug
 export PAGER=cat
 
+machine="$(uname -m)"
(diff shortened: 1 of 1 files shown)


yuwata:
The CI still failed, but on TEST-17-UDEV, not related.

yuwata (approved):
LGTM.

test/units/testsuite-58.sh
```diff
@@ -11,6 +11,60 @@ fi
 export SYSTEMD_LOG_LEVEL=debug
 export PAGER=cat
 
+machine="$(uname -m)"
+if [ "${machine}" = "x86_64" ]; then
```
medhefgo, line 15: Blurgh. shell has `case` for this kind of thing. :P

PR #22103, opened 2022-01-13
build-test: show meson-log.txt when meson fails

evverx:
to make it easier to figure out why it fails.

For example in https://github.com/systemd/systemd/runs/4799774735?check_suite_focus=true
it failed with
```

meson.build:1003:8: ERROR: Command "/usr/bin/clang -print-targets" failed with status 1.

A full log can be found at /home/runner/work/systemd/systemd/build/meson-logs/meson-log.txt
Error: Process completed with exit code 1.
```
and it wasn't clear what exactly happened there.

### DIFF
--- a/.github/workflows/unit_tests.sh
+++ b/.github/workflows/unit_tests.sh
@@ -27,6 +27,13 @@ function info() {
     echo -e "\033[33;1m$1\033[0m"
 }
 
+function run_meson() {
+    if ! meson "$@"; then
+        find . -type f -name meson-log.txt -exec cat '{}' +
+        return 1
+    fi
+}
+
 set -ex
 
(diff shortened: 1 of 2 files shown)


evverx:
It should be added to the "unit tests" too

.github/workflows/unit_tests.sh
```diff
@@ -27,6 +27,13 @@ function info() {
     echo -e "\033[33;1m$1\033[0m"
 }
 
+function run_meson() {
+    if ! meson "$@"; then
+        find . -type f -name meson-log.txt -exec cat '{}' \;
```
medhefgo:
```suggestion
        find . -type f -name meson-log.txt -exec cat '{}' +
```

PR #22104, opened 2022-01-13
ci: turn meson warnings into errors

evverx:


.github/workflows/build_test.sh
```diff
@@ -117,12 +117,16 @@ ninja --version
 for args in "${ARGS[@]}"; do
     SECONDS=0
 
+    if [[ "$LINKER" == lld ]]; then
+        additional_meson_args="-Defi-ld=bfd"
```
medhefgo: Why this? Meson should already be skipping lld automatically for efi-ld.

evverx: Without it meson fails with
```
2022-01-13T10:50:24.5821541Z Header <efi.h> has symbol "EFI_IMAGE_MACHINE_X64" : YES 
2022-01-13T10:50:24.5821950Z src/boot/efi/meson.build:52: WARNING: Not using lld as efi-ld, falling back to bfd
2022-01-13T10:50:24.5822273Z 
2022-01-13T10:50:24.5822521Z src/boot/efi/meson.build:52:16: ERROR: Fatal warnings enabled, aborting
```

medhefgo: You should add a comment then…

evverx: @medhefgo I agree but this PR is still a work in progress and it's marked as "draft" :-) At this point I'm trying to figure out what
```
2022-01-13T12:38:44.4250776Z NOTICE: Future-deprecated features used:
2022-01-13T12:38:44.4251302Z  * 0.55.0: {'ExternalProgram.path'}
```
is and why exactly meson-0.61 started printing that.

evverx: FWIW I think that instead of passing `efi_ld` there it would be better to not turn on the warnings when `LINKER` is set to `ldd` to make sure the linker is chosen correctly automatically.

evverx: @medhefgo I turned off the warnings with ldd and added a comment. I hope it makes more sense now. Thanks!

I expected more warnings with meson-0.61 but it turns out all of them were fixed in https://github.com/systemd/systemd/pull/22088 so it seems it should be safe to switch to it in the foreseeable future.

yuwata (approved):
LGTM.

.github/workflows/build_test.sh
eli-schwartz: The notice is a Future notice, and as such meson does not consider it a warning (and it won't ever be a fatal one).

The idea is to let people know when they are using functionality, in your case the `.path()` method of a find_program() result, which:
- is correct, and required for the versions of meson that you support in `meson_version: '>= ...'`
- later versions of meson deprecate in favor of something else

This gives you a heads up that once you *do* upgrade to meson 0.55.0 as a minimum requirement, you will start getting warnings which in combination with --fatal-meson-warnings are fatal.

You may want to upgrade your minimum and fix that deprecation at the same time, or you may want to preserve support for `<0.55.0` and as a result keep on using `.path()`.

evverx: @eli-schwartz thanks! I think what confused me is that meson doesn't seem to show where exactly it's used and I couldn't translate `'ExternalProgram.path'` into `find_program.path`.

eli-schwartz: ExternalProgram is the implementation detail name, `mesonbuild.programs.ExternalProgram`. The meson reference docs call it [`external_program`](https://mesonbuild.com/Reference-manual_returned_external_program.html) instead.

It can be returned by more things than just find_program, though... the only thing I can think of off the top of my head is `python_mod.find_installation()`.

> meson doesn't seem to show where exactly it's used

It will do that for an actual warning warning (emitted immediately when encountered) but not for the summary printed at the end. I guess it's not hugely important since it is really just there to give you an overview of things you can consider upgrading, rather than pointing you exactly where there is something you need to fix "right now".

PR #22109, opened 2022-01-13
hwdb: add HP mobile broadband modules

oneukum:
These entries enable USB autosuspend on some internal Mobile Broadband USB devices shipped by HP


oneukum:
v2: changed all checks to precede setting the attribute

PR #22115, opened 2022-01-13
man: Add more details about desktop file processing

benzea:
In particular, mention the contract the generator has with external
ExecCondition= binaries that may be provided by desktop environments.

But, also mention all the other relevant keys. In particular
X-systemd-skip= is important to be documented.

This is a follow up for #22072.

CC: @davidedmundson, @keszybz


keszybz:
Looks reasonable. I'll wait for more comments though.

davidedmundson:
As for the changes, +1

PR #22117, opened 2022-01-14
ci: switch to requirements.txt in the unit tests workflow

evverx:


evverx:
@mrc0mmand could you take a look? Now that meson warnings are treated as errors there I think meson should be updated predictably (by analogy with the build test workflow) to avoid bogus warnings produced by ancient versions of meson

evverx:
Judging by
```
2022-01-14T00:25:46.9581612Z The Meson build system
2022-01-14T00:25:46.9582568Z Version: 0.60.3
```
it seems to be working.

PR #21454, opened 2021-11-20
analyze: add inspect-elf verb to parse package metadata

bluca:
Parses and prints package metadata from executables, libraries and core files

```
$ systemd-analyze inspect-elf /tmp/core.fsverity.1000.f77dac5dc161402aa44e15b7dd9dcf97.58561.1637106137000000 ~/git/fsverity-utils/libfsverity.so.0
FILE         /TMP/CORE.FSVERITY.1000.F77DAC5DC161402AA44E15B7DD9DCF97.58561.1637106137000000
module name  /home/luca/git/fsverity-utils/fsverity
type         deb
name         fsverity-utils
version      1.3-1
architecture amd64
os           debian
debugInfoUrl https://debuginfod.debian.net
buildId      f70ffd61ea22aee78c33c56aeda61c63309a55e3
FILE         /HOME/LUCA/GIT/FSVERITY-UTILS/LIBFSVERITY.SO.0
type         deb
name         fsverity-utils
version      1.3-1
architecture amd64
os           debian
debugInfoUrl https://debuginfod.debian.net
buildId      bb9b8587ab7d114b0ddd762d8a1c030dee96fe2f
$ systemd-analyze inspect-elf /tmp/core.fsverity.1000.f77dac5dc161402aa44e15b7dd9dcf97.58561.1637106137000000 ~/git/fsverity-utils/libfsverity.so.0  --json=pretty
{
        "/home/luca/git/fsverity-utils/fsverity" : {
                "type" : "deb",
                "name" : "fsverity-utils",
                "version" : "1.3-1",
                "architecture" : "amd64",
                "os" : "debian",
                "debugInfoUrl" : "https://debuginfod.debian.net",
                "buildId" : "f70ffd61ea22aee78c33c56aeda61c63309a55e3"
        }
}
{
        "/home/luca/git/fsverity-utils/libfsverity.so.0" : {
                "type" : "deb",
                "name" : "fsverity-utils",
                "version" : "1.3-1",
                "architecture" : "amd64",
                "os" : "debian",
                "debugInfoUrl" : "https://debuginfod.debian.net",
                "buildId" : "bb9b8587ab7d114b0ddd762d8a1c030dee96fe2f"
        }
}
```

PR #22121, opened 2022-01-14
fuzz parse_elf

evverx:
This is the fuzzer from
https://github.com/systemd/systemd/pull/21454#issuecomment-1012948381.
I finally got it to compile.

The fuzzer isn't supposed to be merged because it's kind of kludgy
due to that part of the code not being test-friendly in general.
But it should be enough to start fuzzing parse_elf. Happy fuzzing!

To get it to work either elfutils has to be compiled with libFuzzer
and passed via LD_LIBRARY_PATH='path-to-libelf;path-to-libdw' or
a corpus of some kind should be prepared to keep it going.

### DIFF
--- a/src/fuzz/fuzz-core.c
+++ b/src/fuzz/fuzz-core.c
@@ -1,0 +1,40 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "fd-util.h"
+#include "fs-util.h"
+#include "fuzz.h"
+#include "elf-util.h"
+#include "tmpfile-util.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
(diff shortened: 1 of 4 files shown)


evverx:
Generally I turn off this part of systemd (along with systemd-coredump) so I'm not particularly interested in fuzzing it any more. I think it can be closed.

`elfutils` itself has been fuzzed and tested since 2021 and its coverage has been extended far from what `systemd` uses so as long as it's >= 0.187 it should no longer include stuff like https://sourceware.org/git/?p=elfutils.git;a=commit;h=3038ee821b29147f2b036372937747f150f2f755 and https://sourceware.org/git/?p=elfutils.git;a=commit;h=db862a11910a5d4c007c549c2b4ce4cad62f242b.

Issue #17508, opened 2020-11-02
systemd-boot: support beep


sthibaul:
**Is your feature request related to a problem? Please describe.**
Blind users cannot know when the boot menu (of e.g. an ISO installation CD) is showing up, and thus they don't know when to press a shortcut to enable speech-driven installation

**Describe the solution you'd like**
We would need a way to make systemd-boot emit a beep when the boot menu comes up. For instance, syslinux implements this by letting people  put control-G `\007` in the menu text, grub implements this by providing a `play` command that can be called from the menu entry.

webczat:
would certainly be useful. actually you can do without a beep, but it is a bit more hit and miss experience at least when doing it for first time, and the beep makes it clear you just successfully booted from cd, instead of say, accidentally booting from disk, and you are good to go.

poettering:
I am not sure UEFI even has apis for a system beep generator. And if it has, I am not sure how many PCs still have a pc beeper hw around?

Anyway, if this can be merged, happy to merge a patch for this!

sthibaul:
grub simply uses the PIT to emit beeps

sthibaul:
(see `include/grub/speaker.h`)

poettering:
we don't do direct hw access in sd-boot. We use UEFI APIs only. Talking to the PIT directly interfering with UEFI is not OK really in sd-boot.

Also, how many PCs actually still have a speaker connected to a PIT? I doubt many have...

alex19EP:
hello.

> 
> 
> we don't do direct hw access in sd-boot. We use UEFI APIs only. Talking to the PIT directly interfering with UEFI is not OK really in sd-boot.

i found [this](https://github.com/fpmurphy/UEFI-Utilities-2018/blob/master/MyApps/Beep/Beep.c)
as i understand u don't want to do exactly that?

> 
> Also, how many PCs actually still have a speaker connected to a PIT? I doubt many have...

i think it's not important hear. example: I am a blind systemd user and i use systemd-boot on some of my machines. i can control loader via bootctl. but for bootable isos where i need to select boot option beep would be much convenient.

I think it's understandable that I will make sure when assembling or purchasing that the computer has a pc-speaker.

webczat:
to be honest, although no laptop has a pc speaker, I believe many (all?) of them have this somehow wired to the soundcard. :) so that actually works, for example here.

poettering:
[...]

PR #22122, opened 2022-01-14
boot: Add PC speaker support

medhefgo:
Fixes: #17508

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -89,6 +89,7 @@ typedef struct {
         BOOLEAN force_menu;
         BOOLEAN use_saved_entry;
         BOOLEAN use_saved_entry_efivar;
+        BOOLEAN beep;
         INT64 console_mode;
         INT64 console_mode_efivar;
         RandomSeedMode random_seed_mode;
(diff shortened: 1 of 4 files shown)


src/boot/efi/boot.c
```diff
@@ -725,6 +727,11 @@ static BOOLEAN menu_run(
                         ST->ConOut->OutputString(ST->ConOut, clearline + 1 + x + len);
                 }
 
+                if(do_beep) {
```
bluca: missing space

bluca (approved):
hackaton project: play the windows 95 startup theme when the windows bootloader is selected

alex19EP (approved):
tested it in qemu. all works fine.

sthibaul:
> > hackaton project: play the windows 95 startup theme when the windows bootloader is selected
> 
> actually we can further improve convenience for the blind by adding the ability to add a different number of beeps to the boot-entry.

Yes, some people find it useful, I had introduced the `play` command in grub to that end, and people do use it to be able to distinguish between boot entries while moving blindly.

medhefgo:
> Yes, some people find it useful, I had introduced the `play` command in grub to that end, and people do use it to be able to distinguish between boot entries while moving blindly.

Would it be enough to beep n times when selecting the nth entry? (I'd rather not make an explosion of config (entry) options unless it's truly needed. If anyone wants full control, they really ought to use grub.)

sthibaul:
> Would it be enough to beep n times when selecting the nth entry?

That would already be plenty, yes!

Thanks a lot for your work on this!

medhefgo:
See https://github.com/systemd/systemd/pull/22140. Let me know if the beep frequency is too fast or slow.

PR #22125, opened 2022-01-14
shared: Copy holes in sparse files in copy_bytes_full()

daandemeyer:
Previously, all holes in sparse files copied with copy_bytes_full()
would be expanded in the target file. Now, we correctly detect holes
in the input file and we replicate them in the target file.

This PR also includes a second commit that enables COPY_HOLES 
when we copy archived btrfs journal files in journald.

### DIFF
--- a/src/shared/copy.c
+++ b/src/shared/copy.c
@@ -211,29 +211,28 @@ int copy_bytes_full(
 
                         /* To see if we're in a hole, we search for the next data offset. */
                         e = lseek(fdf, c, SEEK_DATA);
-                        if (e < 0 && errno == ENXIO) {
+                        if (e < 0 && errno == ENXIO)
                                 /* If errno == ENXIO, that means we've reached the final hole of the file and
                                 * that hole isn't followed by more data. */
                                 e = lseek(fdf, 0, SEEK_END);
-                                if (e < 0)
-                                        return -errno;
-                        } else if (e < 0)
+                        if (e < 0)
                                 return -errno;
 
-                        /* If we're in a hole (current offset is not a data offset), create a hole of the same size
-                        * in the target file. */
+                        /* If we're in a hole (current offset is not a data offset), create a hole of the
+                         * same size in the target file. */
                         if (e > c && lseek(fdt, e - c, SEEK_CUR) < 0)
                                 return -errno;
 
                         c = e; /* Set c to the start of the data segment. */
 
-                        /* After copying a potential hole, find the end of the data segment by looking for the next
-                        * hole. If we get ENXIO, we're at EOF. */
+                        /* After copying a potential hole, find the end of the data segment by looking for
+                         * the next hole. If we get ENXIO, we're at EOF. */
                         e = lseek(fdf, c, SEEK_HOLE);
-                        if (e < 0 && errno == ENXIO)
-                                break;
-                        else if (e < 0)
+                        if (e < 0) {
+                                if (errno == ENXIO)
+                                        break;
                                 return -errno;
+                        }
 
                         /* SEEK_HOLE modifies the file offset so we need to move back to the initial offset. */
                         if (lseek(fdf, c, SEEK_SET) < 0)
--- a/src/test/test-copy.c
+++ b/src/test/test-copy.c
@@ -323,4 +323,48 @@ TEST(copy_proc) {
         assert_se(!isempty(a));
 }
 
+TEST_RET(copy_holes) {
+        char fn[] = "/var/tmp/test-copy-hole-fd-XXXXXX";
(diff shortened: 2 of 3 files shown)


src/shared/copy.c
```diff
@@ -202,6 +202,47 @@ int copy_bytes_full(
...
+                        /* To see if we're in a hole, we search for the next data offset. */
+                        e = lseek(fdf, c, SEEK_DATA);
+                        if (e < 0 && errno == ENXIO) {
+                                /* If errno == ENXIO, that means we've reached the final hole of the file and
+                                * that hole isn't followed by more data. */
+                                e = lseek(fdf, 0, SEEK_END);
+                                if (e < 0)
+                                        return -errno;
+                        } else if (e < 0)
+                                return -errno;
```
yuwata: How about the following?
```
if (e < 0 && errno == ENXIO)
        e = lseek();
if (e < 0)
        return -errno;
```

src/shared/copy.c
```diff
@@ -202,6 +202,47 @@ int copy_bytes_full(
...
+
+                        c = e; /* Set c to the start of the data segment. */
+
+                        /* After copying a potential hole, find the end of the data segment by looking for the next
+                        * hole. If we get ENXIO, we're at EOF. */
+                        e = lseek(fdf, c, SEEK_HOLE);
+                        if (e < 0 && errno == ENXIO)
+                                break;
+                        else if (e < 0)
+                                return -errno;
```
yuwata: `e < 0` is evaluated twice. How about the following?
```
if (e < 0) {
        if (errno == ENXIO)
                break;
        return -errno;
}
```

yuwata (approved):
LGTM. Two super minor suggestions. Please ignore them if you do not like.

I think it is better to add testcases for the new flag.

DaanDeMeyer:
Fixed comments and added a test as well

yuwata (approved):
LGTM.

bluca (approved):

src/test/test-copy.c
```diff
@@ -323,4 +323,48 @@ TEST(copy_proc) {
...
+        char fn[] = "/var/tmp/test-copy-hole-fd-XXXXXX";
+        char fn_copy[] = "/var/tmp/test-copy-hole-fd-XXXXXX";
+        struct stat stat;
+        int r, fd, fd_copy;
+
+        fd = mkostemp_safe(fn);
+        assert_se(fd >= 0);
+
+        fd_copy = mkostemp_safe(fn_copy);
+        assert_se(fd >= 0);
```
martinetd, line 336: (doesn't really matter as this will fail on copy_bytes below anyway, but) this should be `assert_se(fd_copy >= 0)`

yuwata: Thanks. Fix is waiting in #22154.

daandemeyer: Woops, my bad, good catch!

src/shared/copy.c
```diff
@@ -202,6 +202,47 @@ int copy_bytes_full(
...
+                                * that hole isn't followed by more data. */
+                                e = lseek(fdf, 0, SEEK_END);
+                                if (e < 0)
+                                        return -errno;
+                        } else if (e < 0)
+                                return -errno;
+
+                        /* If we're in a hole (current offset is not a data offset), create a hole of the same size
+                        * in the target file. */
+                        if (e > c && lseek(fdt, e - c, SEEK_CUR) < 0)
```
poettering, line 223: Hmm, I am not sure how I feel about this. The thing is that this just skips over the target bytes, regardless what they currently are. This works OK if the holes are in the middle of the files *and* the target file is empty so far. But this will go wrong if the hole is at the end of the file, because no write() or ftruncate() is done, so the seek will have zero effect...

Not sure what the best way to fix this is.

At the very least we need to a new flags COPY_TRUNCATE or so, that causes a truncation of the target file once the copy operation is complete, and most calls of COPY_HOLES also need to specify that (except if they operate piecemeal, and call ftruncate on their own in the end).

We might also use fallocate() with FALLOC_FL_PUNCH_HOLE to punch the hole explicitly. I think this would be much nicer in most cases, since regardless what the target fd points to currently the end result would be correct. i.e. we'd always combine FALLOC_FL_PUNCH_HOLE with lseek(). Doing this requires some careful coding to ensure we have good fallbacks, since FALLOC_FL_PUNCH_HOLE is only supported on some file systems and we should try to make sure we don't need more than two syscalls for this, i.e. we need to correctly track the offset to pass to fallocate().

daandemeyer: See also https://github.com/systemd/systemd/pull/22250 which addresses some of these concerns but not all. I was assuming FALLOC_FL_PUNCH_HOLE was supported everywhere.

PR #22126, opened 2022-01-14
Increase maximum length of unit names

msekletar:


msekletar:
@keszybz something like this might work. This is the test-case that I am using so far

https://gist.github.com/msekletar/4942d920b92fbbb284862eb679b08355

Btw, there is still at least one major thing that needs to be figured out which is stopping such units through systemd. For example, manual umount (as done in the test) would work but `systemctl stop <LONG_UNIT_NAME>` will fail right now (and cause some timeout on shutdown) because cgroup realisation will fail. Should we run umount and swapoff in `init.scope` if realising dedicated cgroup fails? That seems sensible to me as realisation can fail for other reasons than just very long cgroup name.

src/basic/unit-name.c
```diff
@@ -821,3 +835,11 @@ bool unit_name_prefix_equal(const char *a, const char *b) {
...
         return memcmp_nn(a, p - a, b, q - b) == 0;
 }
+
+const UnitNameVTable default_unit_name_vtable = {
+        .max_name = 256,
+};
+
+const UnitNameVTable long_unit_name_vtable = {
+        .max_name = UNIT_NAME_LIMIT,
+};
```
keszybz: Hmm, why define a "vtable" for this? It seems unnecessarily complicated.

```C
static inline size_t unit_name_max_length(UnitType type) {
    assert(type >= 0);
    if (IN_SET(type, UNIT_DEVICE, UNIT_MOUNT, UNIT_AUTOMOUNT, UNIT_SWAP))
            /* Add a comment here why we allow longer names for those types… */
            return UNIT_NAME_MAX_LONG;
    return UNIT_NAME_MAX_DEFAULT;
}
```

(And I think it's necessary to adjust the macro names, because it'd be misleading to have UNIT_NAME_MAX when we allow longer unit names.)

src/test/test-unit-name.c
```diff
@@ -152,7 +152,7 @@ TEST(unit_name_from_path_instance) {
...
         test_unit_name_from_path_instance_one("waldo", "/foo", ".waldi", NULL, -EINVAL);
         test_unit_name_from_path_instance_one("wa--ldo", "/--", ".mount", "wa--ldo@\\x2d\\x2d.mount", 0);
-        test_unit_name_from_path_instance_one("waldoaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", "/waldo", ".mount", NULL, -ENAMETOOLONG);
+        test_unit_name_from_path_instance_one("waldoaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", "/waldo", ".mount", "waldoaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa@waldo.mount", 0);
```
keszybz, line 161: Please also add one one over the limit to check that ENAMETOOLONG is still returned.

keszybz:
> For example, manual umount (as done in the test) would work but `systemctl stop <LONG_UNIT_NAME>` will fail right now (and cause some timeout on shutdown) because cgroup realisation will fail. Should we run umount and swapoff in `init.scope` if realising dedicated cgroup fails? That seems sensible to me as realisation can fail for other reasons than just very long cgroup name.

I think we should to this iff we get ENAMETOOLONG.

keszybz (changes requested):

src/test/test-unit-name.c
medhefgo: Dunno if it makes things easier to read, but the super long lines can be avoided with a bit of gcc extensions: `(char[]){'a', 'b', 'c', [3 ... 15] = 'x', '\0'}` creates a string with however many extra repetitions.

src/basic/unit-name.c
msekletar: Initially I wanted this to become part of already existing vtable structs. But that would create "bad" dependency between translation units (src/basic/unit-name.c would then depend on stuff from src/core/) so I had to separate it to its own struct. Since maximum unit name length is now becoming unit type specific property I sticked to table approach.

Anyway, we only have single new property the function is indeed simpler so let's do it that way.

msekletar:
@keszybz I got rid of vtable based approach and incorporated your suggestion. Also, I've addressed problem with stopping units. Last but not least, I've also added integration test. PTAL.

test/units/testsuite-69.sh
```diff
@@ -0,0 +1,29 @@
...
+
+mkdir -p $LONGPATH
+mount -t tmpfs tmpfs $LONGPATH
+systemctl daemon-reload
+
+# Check that unit is active(mounted)
+systemctl --no-pager show -p SubState --value $LONGPATH | grep -q mounted
+
+# Check that relevant part of journal doesn't contain any errors related to unit
+[ $(journalctl -b --since=$TS --priority=err | grep "$LONGMNT"  | wc -l) = "0" ]
```
mrc0mmand:
```suggestion
[ "$(journalctl -b --since="$TS" --priority=err | grep -c "$LONGMNT")" = "0" ]
```

test/units/testsuite-69.sh
```diff
@@ -0,0 +1,29 @@
...
...3333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333/444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444/555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555/666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666/777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777777"
```
mrc0mmand: We could "abuse" `printf` here and write it as:

```suggestion
# Generate a long path: "/(256 * 'x')" and all that 7 times 
LONGPATH="$(printf "/$(printf "x%0.s" {1..256})%0.s" {1..7})"
```

It doesn't help much in terms of readability, but at least the line is shorter ;)

test/units/testsuite-69.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
```
mrc0mmand:
```suggestion
set -eux
```

mrc0mmand (commented):
Just a couple of test-related nitpicks.

src/core/unit.c
```diff
@@ -5450,7 +5450,13 @@ int unit_prepare_exec(Unit *u) {
 
         /* Prepares everything so that we can fork of a process for this unit */
 
-        (void) unit_realize_cgroup(u);
+        r = unit_realize_cgroup(u);
+        if (r < 0 && IN_SET(u->type, UNIT_MOUNT, UNIT_SWAP)) {
+                /* Realizing the cgroup failed, but as a fallback we will run mount/unmount or swapon/swapoff in "init.scope". */
+
+                (void) unit_set_cgroup_path(u, "/init.scope");
```
Werkov: Macro `SPECIAL_INIT_SCOPE` would be better here.

But I'm worried `unit_realize_cgroup()` is not ready for such (ab)use (it'd unnecessarily enable controllers in regular tree, it'd modify attributes of `init.scope` unexpectedly).
The simpler approach could be to zero `exec_params.cgroup_path` so that the helper commands forked by PID 1 just remain in its own cgroup.

src/basic/unit-name.h
```diff
@@ -2,6 +2,7 @@
 #pragma once
 
 #include <stdbool.h>
+#include <linux/limits.h>
```
keszybz: Headers should be sorted alphabetically.

src/basic/unit-name.h
```diff
@@ -16,6 +17,18 @@ typedef enum UnitNameFlags {
...
+#define UNIT_NAME_MAX_DEFAULT 256
+
+static inline size_t unit_name_max_length(UnitType type) {
+    assert(type != _UNIT_TYPE_INVALID && type < _UNIT_TYPE_MAX);
+    if (IN_SET(type, UNIT_DEVICE, UNIT_MOUNT, UNIT_SWAP))
+            /* We allow longer unit names for types where unit can get created because of external event
+             * outside of our control (e.g. device is connected or admin mounts a filesystem by hand). */
+            return UNIT_NAME_MAX_LONG;
+    return UNIT_NAME_MAX_DEFAULT;
+}
```
keszybz: So this commit should be squashed with the the actual first use. It doesn't do what the commit message says, since the new function is unused.

src/basic/unit-name.h
```diff
@@ -16,6 +17,18 @@ typedef enum UnitNameFlags {
         _UNIT_NAME_INVALID = -EINVAL,
 } UnitNameFlags;
 
+#define UNIT_NAME_MAX_LONG    PATH_MAX-1
+#define UNIT_NAME_MAX_DEFAULT 256
+
+static inline size_t unit_name_max_length(UnitType type) {
+    assert(type != _UNIT_TYPE_INVALID && type < _UNIT_TYPE_MAX);
```
keszybz: `>= 0` maybe?

src/basic/unit-name.c
```diff
@@ -42,14 +43,17 @@ bool unit_name_is_valid(const char *n, UnitNameFlags flags) {
...
         e = strrchr(n, '.');
         if (!e || e == n)
                 return false;
 
-        if (unit_type_from_string(e + 1) < 0)
+        t = unit_type_from_string(e + 1);
+        if (t < 0)
+                return false;
+
+        assert(t != _UNIT_TYPE_INVALID);
```
keszybz: So I think it's better to do `>= 0`, because `_from_string()` functions return all kinds of errors, not just -EINVAL.

src/basic/unit-name.h
```diff
@@ -29,6 +27,8 @@ static inline size_t unit_name_max_length(UnitType type) {
     return UNIT_NAME_MAX_DEFAULT;
 }
 
+#define UNIT_NAME_MAX(t) unit_name_max_length(t)
```
keszybz: This macros doesn't seem useful at all.

src/core/unit.c
```diff
@@ -5450,7 +5450,13 @@ int unit_prepare_exec(Unit *u) {
 
         /* Prepares everything so that we can fork of a process for this unit */
 
-        (void) unit_realize_cgroup(u);
+        r = unit_realize_cgroup(u);
+        if (r < 0 && IN_SET(u->type, UNIT_MOUNT, UNIT_SWAP)) {
```
keszybz, line 5458: This chunk should not be part of this commit.

src/test/test-unit-name.c
```diff
@@ -152,7 +154,8 @@ TEST(unit_name_from_path_instance) {
...
+        test_unit_name_from_path_instance_one(s, "/waldo", ".mount", NULL, -ENAMETOOLONG);
```
keszybz, line 162: And this commit should be squashed together with the commit that changes the maximum allowed lengths…

keszybz (changes requested):

src/core/unit.c
msekletar: I will experiment with this alternative approach, it might be a bit safer.

msekletar: I just had a closer look at `unit_set_cgroup_path()` and you are absolutely right. I can't do this, because otherwise I'd remove the association between `init.scope` and unit object for that unit and I'd replace it with unit object of mount/swap unit.

src/core/unit.c
msekletar: I've changed this a bit based on input from @Werkov and split it into separate commit.

src/core/unit.c
msekletar: In the latest version I am just unsetting `u->cgroup_path` and that seems to work fine. PTAL

msekletar:
I think this is now ready for re-review. @poettering @keszybz PTAL.

poettering:
Let's take a step back here, what's precisely the rationale for doing this? This is a compat break, as we so far communicated unit names fit into filenames. If that's not the case anymore this might have a ripple effect.

poettering:
Clicking through the linked issues I can't really find an explanation why we'd want this? Can you elaborate?

src/basic/unit-name.h
```diff
@@ -16,6 +15,18 @@ typedef enum UnitNameFlags {
...
+#define UNIT_NAME_MAX_LONG    PATH_MAX-1
+#define UNIT_NAME_MAX_DEFAULT 256
+
+static inline size_t unit_name_max_length(UnitType type) {
+    assert(type >= 0 && type < _UNIT_TYPE_MAX);
+    if (IN_SET(type, UNIT_DEVICE, UNIT_MOUNT, UNIT_SWAP))
+            /* We allow longer unit names for types where unit can get created because of external event
+             * outside of our control (e.g. device is connected or admin mounts a filesystem by hand). */
+            return UNIT_NAME_MAX_LONG;
+    return UNIT_NAME_MAX_DEFAULT;
```
poettering, line 27: indetnation...

src/basic/unit-name.h
```diff
@@ -16,6 +15,18 @@ typedef enum UnitNameFlags {
         _UNIT_NAME_INVALID = -EINVAL,
 } UnitNameFlags;
 
+#define UNIT_NAME_MAX_LONG    PATH_MAX-1
```
poettering, line 18: if your intention to be able to stash any file system path into unit names, then PATH_MAX isnot going to work, since paths can basically contain almost any char, and we only allow non-crazy chars and expect the other ones to be escaped.

src/basic/unit-name.h
poettering: why different limits for different types? that's awkward and unnecessary?

src/core/unit.c
```diff
@@ -5454,7 +5454,12 @@ int unit_prepare_exec(Unit *u) {
...
+                /* Realizing the cgroup failed, but as a fallback we will run mount/unmount or swapon/swapoff in init.scope. */
```
poettering, line 5459: no. that's a hack. let's not do that.

poettering (commented):
I am not convinced this is desirable, and this is quite hackish as it doesn't solve any of the issues this creates, i.e. the fact that we'll try to load unit files of disks from paths that are not valid anymore, and we'll try to place unit processes in cgroups with have no valid paths either... And this goes on, i.e. the /run/systemd/units/ thing is not going to work either anymore...

poettering:
anyway, before we cntinue discusson on the details of the implementation: this needs a really strong rationale: why do we want this, and why is that so important that we are willing to accept all the shortcomings this introduces? why would adding this half-baked support be beneficial over just saying mounts with ridiculously long paths are not managable by systemd and exist outside of systemd's management?

src/basic/unit-name.h
keszybz: The rationale for doing this just for mounts and devices: for those types the unit name can be specified externally and we just create a unit to match what the kernel tells us. For other types, the unit name is under our control. I think it makes sense to limit the change to only the specific unit types where we need this, but nowhere else. In particular, we prolly don't want service units with very long names.

keszybz:
We discussed this in a meeting today. The conclusion was that @msekletar will rework the patch to keep the length limit but hash overly long unit names to something like `dev-disk-extender-extender-…hash.device`. This way we can create files for the unit and a cgroup if necessary.

msekletar:
Here is the alternative proposal,

https://github.com/systemd/systemd/pull/22759

poettering:
Let's close this one. I am pretty sure #22759 is the way to go.

PR #22132, opened 2022-01-15
resolved: Fix DoT timeout on multiple answer records (for CloudFlare, Google, etc. DoT servers)

joanbm:
When sending multiple DNS questions to a DNS-over-TLS server (e.g. a question for A and AAAA records, as is typical) on the same session, the server may answer to each question in a separate TLS record, but it may also aggregate multiple answers in a single TLS record.
(Some servers do this very often (e.g. Cloudflare 1.0.0.1), some do it sometimes (e.g. Google 8.8.8.8) and some seem to never do it (e.g. Quad9 9.9.9.10)).

Both cases should be handled equivalently, as the byte stream is the same, but when multiple answers came in a single TLS record, usually the first answer was processed, but the second answer was entirely ignored, which caused a 10s delay until the resolution timed out and the missing question was retried. This can be reproduced by configuring one of the offending server and running `resolvectl query google.com --cache=no` a few times.

To be notified of incoming data, systemd-resolved listens to `EPOLLIN` events on the underlying socket. However, when DNS-over-TLS is used, the TLS library (OpenSSL or GnuTLS) may read and buffer the entire TLS record when reading the first answer, so usually no further `EPOLLIN` events will be generated, and the second answer will never be processed.

To avoid this, if there's buffered TLS data, generate a "fake" EPOLLIN event. This is hacky, but it makes this case transparent to the rest of the IO code.

Fixes: https://bbs.archlinux.org/viewtopic.php?id=262926

--

**Appendix:**

To verify the issue, you can use socat like so to intercept some DoT server: `socat -v -x openssl-listen:853,reuseaddr,fork,cert=127.0.0.1.pem,key=127.0.0.1-key.pem,verify=0 openssl:8.8.8.8:853` (you will need a trusted certificate for localhost).

Sample stream that did not cause systemd-resolved to hang:
```
> 2022/01/15 22:23:48.849930  length=2 from=820 to=821
 00 27                                            .'
--
> 2022/01/15 22:23:48.850041  length=39 from=822 to=860
 f2 43 01 00 00 01 00 00 00 00 00 01 06 67 6f 6f  .C...........goo
 67 6c 65 03 63 6f 6d 00 00 01 00 01 00 00 29 ff  gle.com.......).
 e4 00 00 00 00 00 00                             .......
--
> 2022/01/15 22:23:48.850178  length=2 from=861 to=862
 00 27                                            .'
--
> 2022/01/15 22:23:48.850230  length=39 from=863 to=901
 b0 30 01 00 00 01 00 00 00 00 00 01 06 67 6f 6f  .0...........goo
 67 6c 65 03 63 6f 6d 00 00 1c 00 01 00 00 29 ff  gle.com.......).
 e4 00 00 00 00 00 00                             .......
--
< 2022/01/15 22:23:48.883489  length=69 from=1260 to=1328
 00 43 b0 30 81 80 00 01 00 01 00 00 00 01 06 67  .C.0...........g
 6f 6f 67 6c 65 03 63 6f 6d 00 00 1c 00 01 c0 0c  oogle.com.......
 00 1c 00 01 00 00 00 87 00 10 2a 00 14 50 40 03  ..........*..P@.
 08 06 00 00 00 00 00 00 20 0e 00 00 29 02 00 00  ........ ...)...
 00 00 00 00 00                                   .....
--
< 2022/01/15 22:23:48.883906  length=57 from=1329 to=1385
 00 37 f2 43 81 80 00 01 00 01 00 00 00 01 06 67  .7.C...........g
 6f 6f 67 6c 65 03 63 6f 6d 00 00 01 00 01 c0 0c  oogle.com.......
 00 01 00 01 00 00 01 12 00 04 8e fa b2 ae 00 00  ................
 29 02 00 00 00 00 00 00 00                       )........
```

Sample stream that caused systemd-resolved to hang:
```
> 2022/01/15 22:23:50.113990  length=2 from=902 to=903
 00 27                                            .'
--
> 2022/01/15 22:23:50.114074  length=39 from=904 to=942
 d4 53 01 00 00 01 00 00 00 00 00 01 06 67 6f 6f  .S...........goo
 67 6c 65 03 63 6f 6d 00 00 01 00 01 00 00 29 ff  gle.com.......).
 e4 00 00 00 00 00 00                             .......
--
> 2022/01/15 22:23:50.114174  length=2 from=943 to=944
 00 27                                            .'
--
> 2022/01/15 22:23:50.114216  length=39 from=945 to=983
 35 79 01 00 00 01 00 00 00 00 00 01 06 67 6f 6f  5y...........goo
 67 6c 65 03 63 6f 6d 00 00 1c 00 01 00 00 29 ff  gle.com.......).
 e4 00 00 00 00 00 00                             .......
--
< 2022/01/15 22:23:50.147560  length=126 from=1386 to=1511
 00 37 d4 53 81 80 00 01 00 01 00 00 00 01 06 67  .7.S...........g
 6f 6f 67 6c 65 03 63 6f 6d 00 00 01 00 01 c0 0c  oogle.com.......
 00 01 00 01 00 00 01 10 00 04 8e fa b2 ae 00 00  ................
 29 02 00 00 00 00 00 00 00 00 43 35 79 81 80 00  ).........C5y...
 01 00 01 00 00 00 01 06 67 6f 6f 67 6c 65 03 63  ........google.c
 6f 6d 00 00 1c 00 01 c0 0c 00 1c 00 01 00 00 00  om..............
 85 00 10 2a 00 14 50 40 03 08 06 00 00 00 00 00  ...*..P@........
 00 20 0e 00 00 29 02 00 00 00 00 00 00 00        . ...)........
```


joanbm:
Update: Fixed error handling for `sd_event_source_get_io_events`.

src/resolve/resolved-dnstls-gnutls.c
```diff
@@ -211,6 +211,14 @@ ssize_t dnstls_stream_read(DnsStream *stream, void *buf, size_t count) {
         return ss;
 }
 
+int dnstls_stream_has_buffered_data(DnsStream *stream) {
```
yuwata: `int` -> `bool`

src/resolve/resolved-dnstls-openssl.c
```diff
@@ -367,6 +367,14 @@ ssize_t dnstls_stream_read(DnsStream *stream, void *buf, size_t count) {
         return ss;
 }
 
+int dnstls_stream_has_buffered_data(DnsStream *stream) {
```
yuwata: `int` -> `bool`

src/resolve/resolved-dnstls-openssl.c
```diff
@@ -367,6 +367,14 @@ ssize_t dnstls_stream_read(DnsStream *stream, void *buf, size_t count) {
...
+        assert(stream);
+        assert(stream->encrypted);
+        assert(stream->dnstls_data.ssl);
+
+        return SSL_has_pending(stream->dnstls_data.ssl);
```
yuwata: The man page says it returns 0 or 1, but for safety `return SSL_has_pending() > 0;`

src/resolve/resolved-dns-stream.c
```diff
@@ -441,6 +453,34 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
...
+                        return r;
+
+                /* When using DNS-over-TLS, the underlying TLS library may read the entire TLS record
+                   and buffer it internally. If this happens, we will not receive further EPOLLIN events,
+                   and unless there's some unrelated activity on the socket, we will hang until time out.
+                   To avoid this, if there's buffered TLS data, generate a "fake" EPOLLIN event.
+                   This is hacky, but it makes this case transparent to the rest of the IO code. */
+                if ((events & EPOLLIN) != 0 && dnstls_stream_has_buffered_data(s)) {
+                        if (!s->tls_buffered_data_event_source) {
+                                r = sd_event_add_defer(s->manager->event, &s->tls_buffered_data_event_source,
```
yuwata: Why do not you simply loop this function?

src/resolve/resolved-dns-stream.c
```diff
@@ -441,6 +453,34 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
...
+                r = sd_event_source_get_io_events(s->io_event_source, &events);
+                if (r < 0)
+                        return r;
+
+                /* When using DNS-over-TLS, the underlying TLS library may read the entire TLS record
+                   and buffer it internally. If this happens, we will not receive further EPOLLIN events,
+                   and unless there's some unrelated activity on the socket, we will hang until time out.
+                   To avoid this, if there's buffered TLS data, generate a "fake" EPOLLIN event.
+                   This is hacky, but it makes this case transparent to the rest of the IO code. */
+                if ((events & EPOLLIN) != 0 && dnstls_stream_has_buffered_data(s)) {
```
yuwata: Not a matter, but I prefer `FLAGS_SET(events, EPOLLIN)`.

yuwata (commented):
Sorry for late review.

src/resolve/resolved-dnstls-gnutls.c
joanbm: Done.

src/resolve/resolved-dnstls-openssl.c
joanbm: Done.

src/resolve/resolved-dnstls-openssl.c
joanbm: Done.

src/resolve/resolved-dns-stream.c
joanbm: Agreed and done.

joanbm:
Reviewed according to comments + added an automated test for DnsStream reproducing the bug.

src/resolve/resolved-dns-stream.c
joanbm: Looping is possible, it's more of an style thing:
(1) Besides the declarations the entire handling of the edge case is contained in a single block
(2) Looping/recursing/goto'ing to the top of the function feels ugly (more indentation or stack frames or labels, inside #ifdefs for ENABLE_DNS_OVER_TLS)
(3) The code flow is more unified with the plain TCP case, e.g. an event emits at most one DNS packet

But I understand that emitting a "fake IO" event is also a bit ugly and less performant, so if you feel looping is preferable, I can change that.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,332 @@
...
+        0x20, 0x00, 0x01, 0x02, 0x48, 0x18, 0x93, 0x25, 0xC8, 0x19, 0x46,
+};
+
+/**
+ * A mock TCP DNS server that asserts certain questions are received
+ * and replies with the same answer every time.
+ */
+static void receive_and_check_question(int fd, const uint8_t *expected_question,
+                                       size_t question_size) {
+        uint8_t actual_question[question_size];
```
yuwata: This is a test code, so not strong requirement, but we do not usually use dynamic array in this style.
So, I prefer to use `newa()` or `new()`.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,332 @@
...
+}
+
+/* Sends two answers together in a single write operation,
+ * so they hopefully end up in a single TCP packet / TLS record */
+static void send_answers_together(int fd,
+                                  const uint8_t *answer1, size_t answer1_size,
+                                  const uint8_t *answer2, size_t answer2_size) {
+        size_t answer_size = answer1_size + answer2_size;
+        uint8_t answer[answer_size];
+        memcpy(answer, answer1, answer1_size);
```
yuwata, line 86: Please add a blank line between variable declarations and code. Also, I prefer to use `new()` or `newa()` here.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,332 @@
...
+}
+
+static void *tcp_dns_server(void *p) {
+        _cleanup_close_ int bindfd = -1, acceptfd = -1;
+
+        assert_se((bindfd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0)) != -1);
+        assert_se(setsockopt(bindfd, SOL_SOCKET, SO_REUSEADDR, &(int){1}, sizeof(int)) == 0);
+        assert_se(bind(bindfd, &SERVER_ADDRESS, sizeof(SERVER_ADDRESS)) == 0);
+        assert_se(listen(bindfd, 1) == 0);
+        assert_se((acceptfd = accept(bindfd, NULL, NULL)) != -1);
```
yuwata: We usually use `>= 0` for the error check in such cases, even if the man page says it return -1 on error, and 0 on success.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,332 @@
...
+        pid_t openssl_pid;
+        _cleanup_close_ int fd_server = -1, fd_tls = -1;
+        _cleanup_free_ char *cert_path = NULL, *key_path = NULL;
+        char ip_str[40], bind_str[50];
+
+        assert_se(get_testdata_dir("test-resolve/selfsigned.cert", &cert_path) >= 0);
+        assert_se(get_testdata_dir("test-resolve/selfsigned.key", &key_path) >= 0);
+
+        assert_se(inet_ntop(SERVER_ADDRESS.sin_family, &SERVER_ADDRESS.sin_addr,
+                            ip_str, sizeof(ip_str)) != NULL);
```
yuwata: `in_addr_to_string()`?

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,332 @@
...
+        sprintf(bind_str, "%s:%d", ip_str, be16toh(SERVER_ADDRESS.sin_port));
```
yuwata: It is ok to use `asprintf()`, then you can drop magic number `40` or `50` for the buffer size.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,331 @@
...
+         * stdin/stdout, so we will be able to read and write plaintext
+         * from the other end's file descriptor like an usual TCP server */
+        {
+                int fd[2];
+                assert_se(socketpair(AF_UNIX, SOCK_STREAM, 0, fd) == 0);
+                fd_server = fd[0];
+                fd_tls = fd[1];
+        }
+
+        assert_se((openssl_pid = fork()) >= 0);
```
yuwata: Not a matter, but you can use `safe_fork()`.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,331 @@
...
+static const uint64_t EVENT_TIMEOUT_USEC = 5 * 1000 * 1000;
+
+static void send_simple_question(DnsStream *stream, sd_event *event, uint16_t type) {
+        _cleanup_(dns_packet_unrefp) DnsPacket *p = NULL;
+        _cleanup_(dns_resource_key_unrefp) DnsResourceKey *key = NULL;
+        _cleanup_(dns_question_unrefp) DnsQuestion *question = NULL;
+
+        assert_se(dns_packet_new(&p, DNS_PROTOCOL_DNS, 0, DNS_PACKET_SIZE_MAX) == 0);
+        assert_se((question = dns_question_new(1)) != NULL);
+        assert_se((key = dns_resource_key_new(DNS_CLASS_IN, type, TEST_DOMAIN)) != NULL);
```
yuwata: `!= NULL` is not necessary. Our coding style allows implicit cast from pointer to boolean.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,331 @@
...
+        assert_se(dns_question_add(question, key, 0) == 0);
+        assert_se(dns_packet_append_question(p, question) == 0);
+        DNS_PACKET_HEADER(p)->qdcount = htobe16(dns_question_size(question));
+        assert_se(dns_stream_write_packet(stream, p) == 0);
+}
+
+static const size_t MAX_RECEIVED_PACKETS = 2;
+static DnsPacket *received_packets[2] = { 0 };
```
yuwata: `0` is not necessay.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,331 @@
...
+         _cleanup_(dns_stream_unrefp) DnsStream *stream = NULL;
+        _cleanup_(sd_event_unrefp) sd_event *event = NULL;
+        _cleanup_close_ int clientfd = -1;
+        int r;
+
+        void *(*server_entrypoint)(void *);
+        pthread_t server_thread;
+        pthread_mutex_t server_lock;
+
+        Manager manager = { 0 };
```
yuwata: Also here. Please drop 0.

src/resolve/resolved-dns-stream.c
yuwata: If looping also works, then I prefer to convert the logic you added into loop.

How about to rename `on_stream_io()` to e.g. `on_stream_io_impl()` and then
```
static int on_stream_io_impl(DnsStream *s, uint32_t revents) {
        ...
}

static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *userdata) {
        _cleanup_(dns_stream_unrefp) DnsStream *s = NULL;
        int r;

        assert(userdata);

        s = dns_stream_ref(userdata); /* Protect stream while we process it */

        r = on_stream_io(s, revents);
        if (r < 0)
                return r;

#if ENABLE_DNS_OVER_TLS
        if (!s->encrypted)
                return 0;
        if (!FLAGS_SET(events, EPOLLIN))
                return 0;

        while (dnstls_stream_has_buffered_data(s)) {
                /* When using DNS-over-TLS, the underlying TLS library may read the entire TLS record
                   and buffer it internally. If this happens, we will not receive further EPOLLIN events,
                   and unless there's some unrelated activity on the socket, we will hang until time out.
                   To avoid this, ....... */
                r = on_stream_io_impl(s, revents);
                if (r < 0)
                        return r;
        }
#endif

        return 0;
}
```

yuwata (commented):
Thank you for adding test. That's super nice!

joanbm:
Ready for re-review:
* Converted logic to handle TLS buffered data from defer event to loop
* Fixed style issues in the tests
* Fix CI: Skip the unit test if the 'openssl' command is not available at runtime. @mrc0mmand not sure if you have any better way to go about this. I don't like to use OpenSSL or GnuTLS as a library in the test, I prefer something like openssl/socat/stunnel/etc. since it's simpler, and harder to get wrong.
* Fix a minor memory safety (ASAN) issue in the test

PS: Sorry for launching so many CI runs, I kept finding small things to fix.

src/resolve/test-resolved-stream.c
```diff
@@ -0,0 +1,345 @@
...
+                .sin_port = htobe16(12345),
+                .sin_addr.s_addr = htobe32(INADDR_LOOPBACK)
+        };
+
+        test_setup_logging(LOG_DEBUG);
+
+        test_dns_stream(false);
+#if ENABLE_DNS_OVER_TLS
+        if (system("openssl version >/dev/null 2>&1") != 0) {
+                log_warning("Skipping TLS test since the 'openssl' command does not seem to be available");
```
yuwata: `return log_tests_skipped("...");` can be used.

src/resolve/resolved-dns-stream.c
```diff
@@ -280,13 +281,10 @@ static int on_stream_timeout(sd_event_source *es, usec_t usec, void *userdata) {
         return dns_stream_complete(s, ETIMEDOUT);
 }
 
-static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *userdata) {
-        _cleanup_(dns_stream_unrefp) DnsStream *s = dns_stream_ref(userdata); /* Protect stream while we process it */
+static int on_stream_io_impl(DnsStream *s, uint32_t revents) {
         bool progressed = false;
         int r;
 
-        assert(s);
```
yuwata, line 288: Please do not remove it. Of course, it is non-NULL now, but just for readability.

src/resolve/resolved-dns-stream.c
```diff
@@ -441,6 +439,45 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
...
+
+static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *userdata) {
+        _cleanup_(dns_stream_unrefp) DnsStream *s = dns_stream_ref(userdata); /* Protect stream while we process it */
+        int r;
+
+        assert(s);
+
+        r = on_stream_io_impl(s, revents);
+        if (r <= 0)
+                return r;
```
yuwata, line 458: Could you add a brief comment at the beginning of `on_stream_io_impl()` to mention that it returns negative errno on critical error, 0 when recoverable error or the stream handling is finished. and 1 for possibly remaining stream exists.

src/resolve/resolved-dns-stream.c
```diff
@@ -441,6 +439,45 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
...
+
+#if ENABLE_DNS_OVER_TLS
+        if (s->encrypted) {
```
yuwata: Let's reduce the indent:
```
if (!s->encrypted)
        return 0;
```

src/resolve/resolved-dns-stream.c
```diff
@@ -441,6 +439,45 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
...
+                uint32_t events;
+
+                r = sd_event_source_get_io_events(s->io_event_source, &events);
```
yuwata, line 473: Why this is called?? Why do not you reuse `revents` here?

yuwata (changes requested):

mrc0mmand:
>    * Fix CI: Skip the unit test if the 'openssl' command is not available at runtime. @mrc0mmand not sure if you have any better way to go about this. I don't like to use OpenSSL or GnuTLS as a library in the test, I prefer something like openssl/socat/stunnel/etc. since it's simpler, and harder to get wrong.

I definitely agree, so using the `openssl` utility here is fine by me.

> PS: Sorry for launching so many CI runs, I kept finding small things to fix.

Don't worry about it, most of the CIs automagically cancel the previous runs on force push.

src/resolve/resolved-dns-stream.c
joanbm: From `man 2 poll`, `revents` are the returned events (i.e. which events we received to process at the start of the call), `events` are the requested events (i.e. which events we could process next).

The point of checking whether `FLAGS_SET(events, EPOLLIN)` is to know whether the stream still wants to read more data. It would be wrong to generate our "pseudo-`EPOLLIN` event " on buffered TLS data if the stream is not even expecting to receive further data available events.

Note that in this case `events` is just what was set by the last call to `dns_stream_update_io`. For example, looking at the condition on `dns_stream_update_io`, when we already have a full DNS packet stored on `s->read_packet`, we don't want to process any further data, even if it's available.

PS: BTW, something unrelated to this issue which I noticed when developing the test, is that `dns_stream_take_read_packet`doesn't call `dns_stream_update_io`. What this means is that unless you process the packets on the (supposedly optional) `on_packet` callback, taking the packet doesn't set the `EPOLLIN` flag again, so no further packets will be processed even if they are available... which seems wrong. But all callers use the `on_packet` callback, so this is just a theoretical issue for now.

src/resolve/test-resolved-stream.c
joanbm: Good catch, this is much cleaner.

src/resolve/resolved-dns-stream.c
yuwata: That makes sense for me. Thanks.

yuwata: > PS: BTW, something unrelated to this issue which I noticed when developing the test, is that `dns_stream_take_read_packet`doesn't call `dns_stream_update_io`. What this means is that unless you process the packets on the (supposedly optional) `on_packet` callback, taking the packet doesn't set the `EPOLLIN` flag again, so no further packets will be processed even if they are available... which seems wrong. But all callers use the `on_packet` callback, so this is just a theoretical issue for now.

I cannot catch the issue... There is `dns_stream_update_io()` after `on_packet()`. Is it not enough?

yuwata (approved):
LGTM. Thank you!

(I dropped doubled empty line. And re-ordered the commits, as the test does not pass without the fix.)

src/resolve/resolved-dns-stream.c
joanbm: Imagine I had written the loop of line 300 in `src/resolve/test-resolved-stream.c ` like this:
```c
        stream->on_packet = NULL;
        while (n_received_packets != 2) {
                assert_se(sd_event_run(event, EVENT_TIMEOUT_USEC) >= 1);
                DnsPacket *p = dns_stream_take_read_packet(stream);
                if (p)
                        received_packets[n_received_packets++] = p;
        }
```

IMO this ought to work, but it actually fails for both TCP and TLS, reading only a single packet, then timing out.

Why? Because when I call `sd_event_run`, `stream->read_packet` is filled. When `dns_stream_update_io` is called after reading the packet in `on_stream_io_impl`, the `EPOLLIN` event flag is unset. Later I call `dns_stream_take_read_packet` so `stream->read_packet` is free again, but no one sets the `EPOLLIN` event flag again, so no further packets are read and the test times out.

But all current usages call `dns_stream_take_read_packet` *inside* the `on_packet` callback rather than outside, so the `EPOLLIN` event flag is never unset, and everything works.

yuwata: Thank you for the explanation! Let's add a brief comment in `dns_stream_take_read_packet()`.

yuwata: BTW, while I try to understand your comment previously, I found a potential issue, and filed as #22266. If you'd like, PTAL.

yuwata: > Thank you for the explanation! Let's add a brief comment in `dns_stream_take_read_packet()`.

#22274.

yuwata:
BTW, I noticed that this extremely improves performance (~7x faster :+1: ):

With
```
$ resolvectl status enp0s20f0u2 
Link 4 (enp0s20f0u2)
    Current Scopes: DNS LLMNR/IPv4 LLMNR/IPv6
         Protocols: +DefaultRoute +LLMNR -mDNS +DNSOverTLS DNSSEC=yes/supported
Current DNS Server: 1.1.1.1
       DNS Servers: 1.1.1.1 1.0.0.1
```

Before:
```
$ resolvectl --cache=no query www.youtube.com
www.youtube.com: 172.217.31.142                -- link: enp0s20f0u2
                 172.217.161.46                -- link: enp0s20f0u2
                 142.250.196.142               -- link: enp0s20f0u2
                 172.217.175.14                -- link: enp0s20f0u2
                 172.217.175.46                -- link: enp0s20f0u2
                 172.217.175.78                -- link: enp0s20f0u2
                 172.217.175.110               -- link: enp0s20f0u2
                 172.217.31.174                -- link: enp0s20f0u2
                 216.58.220.142                -- link: enp0s20f0u2
                 172.217.161.78                -- link: enp0s20f0u2
                 172.217.175.238               -- link: enp0s20f0u2
                 216.58.220.110                -- link: enp0s20f0u2
                 172.217.174.110               -- link: enp0s20f0u2
                 142.251.42.142                -- link: enp0s20f0u2
                 142.250.196.110               -- link: enp0s20f0u2
                 142.251.42.174                -- link: enp0s20f0u2
                 2404:6800:4004:813::200e      -- link: enp0s20f0u2
                 2404:6800:4004:80b::200e      -- link: enp0s20f0u2
                 2404:6800:4004:808::200e      -- link: enp0s20f0u2
                 2404:6800:4004:81c::200e      -- link: enp0s20f0u2
                 (youtube-ui.l.google.com)

-- Information acquired via protocol DNS in 635.8ms.
-- Data is authenticated: no; Data was acquired via local or encrypted transport: yes
-- Data from: network
```
```
$ resolvectl --cache=no query www.youtube.com
www.youtube.com: 2404:6800:4004:826::200e      -- link: enp0s20f0u2
                 2404:6800:4004:80f::200e      -- link: enp0s20f0u2
                 2404:6800:4004:801::200e      -- link: enp0s20f0u2
                 2404:6800:4004:827::200e      -- link: enp0s20f0u2
                 142.250.207.46                -- link: enp0s20f0u2
                 216.58.220.110                -- link: enp0s20f0u2
                 142.250.196.110               -- link: enp0s20f0u2
                 172.217.25.78                 -- link: enp0s20f0u2
                 216.58.197.238                -- link: enp0s20f0u2
                 142.250.199.110               -- link: enp0s20f0u2
                 142.251.42.174                -- link: enp0s20f0u2
                 172.217.161.78                -- link: enp0s20f0u2
                 216.58.197.206                -- link: enp0s20f0u2
                 142.250.207.14                -- link: enp0s20f0u2
                 172.217.25.110                -- link: enp0s20f0u2
                 172.217.31.142                -- link: enp0s20f0u2
                 142.251.42.142                -- link: enp0s20f0u2
                 172.217.26.238                -- link: enp0s20f0u2
                 142.250.196.142               -- link: enp0s20f0u2
                 142.251.42.206                -- link: enp0s20f0u2
                 (youtube-ui.l.google.com)

-- Information acquired via protocol DNS in 90.9ms.
-- Data is authenticated: no; Data was acquired via local or encrypted transport: no
-- Data from: network
```

foutrelis:
Does this fix meet the criteria for a stable backport?

Context: I'm using systemd 250.4 and noticed page loads being delayed in Chromium shortly after enabling `DNSOverTLS`. After applying 2aaf6bb6e99b0f2bd73e0c49bef9e11a2844bf1a on top of v250.4, `resolvectl query google.com --cache=no` no longer hangs. 🐭

foutrelis:
Does this fix meet the criteria for a stable backport?

Context: I'm using systemd 250.4 and noticed page loads being delayed in Chromium shortly after enabling `DNSOverTLS`. After applying 2aaf6bb6e99b0f2bd73e0c49bef9e11a2844bf1a on top of v250.4, `resolvectl query google.com --cache=no` no longer hangs. 🐭

PR #22134, opened 2022-01-16
core: add missing dependency DBus properties

### LINKED ISSUE #22133: UpheldBy property does not exist
 > systemd 250 (250.1-1-arch)

**Used distribution**
 > Arch

 > `Linux 5.15.13-arch1-1 #1 SMP PREEMPT Wed, 05 Jan 2022 16:20:59 +0000 x86_64 GNU/Linux`

 > x86_64

 > systemctl show b.service | grep "UpheldBy"
   -> UpheldBy=a.service

 > systemctl show b.service | grep "UpheldBy"
   -> \<empty\>

 > 1.  pick two services _a.service_ and _b.service_
 > 1.  stop both services
 > 1. in _[Unit]_ of _a.service_, add `Upholds=b.service`
 > 1. start _a.service_
 > 1. verify _a.service_ has started
 > 1. verify _b.service_ has started
 > 1. execute `systemctl show b.service |grep "UpheldBy"`

```text
<from systemd.unit>
       Upholds=
           Configures dependencies similar to Wants=, but as long as this unit is up, all units listed in Upholds= are started whenever found to be inactive or failed, and no job is queued for them. While a
           Wants= dependency on another unit has a one-time effect when this units started, a Upholds= dependency on it has a continuous effect, constantly restarting the unit if necessary. This is an
           alternative to the Restart= setting of service units, to ensure they are kept running whatever happens.

           When Upholds=b.service is used on a.service, this dependency will show as UpheldBy=a.service in the property listing of b.service. The UpheldBy= dependency cannot be specified directly.

```

yuwata:
Follow-up for 0bc488c99ab2ed3464237607e381f4d72cd321d5.

Also sort dependency properties to make them match the definition of
`enum UnitDependency` in basic/unit-def.h.

Fixes #22133.


yuwata:
The dependency is introduced since v249. Please backport this to v249-stable and v250-stable.

bluca:
Docs need an update

yuwata:
Docs are also updated.

bluca (approved):

PR #9028, opened 2018-05-18
meson: use run_target for generating tags with ctags

evverx:
In https://github.com/systemd/systemd/pull/6561, `run_target` was changed to `custom_target`, which inadvertently caused relative paths to be passed to ctags due to https://github.com/mesonbuild/meson/issues/3589. The switch to `run_target` causes absolute paths to be passed again and makes it easier to jump from file to file, hopefully delaying the need to exit Vim :-)

PR #22135, opened 2022-01-16
meson: force ctags to use absolute paths

evverx:
Looks like https://github.com/mesonbuild/meson/issues/957 was
reintroduced in meson-0.57.0 (and looking and https://mesonbuild.com/Release-notes-for-0-57-0.html
I'm not sure whether it was intentional or not) so run_command can no
longer be used to get around
https://github.com/mesonbuild/meson/issues/3589. Let's just force
ctags to always use absolute paths to fix it once and for all.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3877,7 +3877,7 @@ if git.found()
                         command : [env, 'etags', '-o', '@0@/TAGS'.format(project_source_root)] + all_files)
                 run_target(
                         'ctags',
-                        command : [env, 'ctags', '-o', '@0@/tags'.format(project_source_root)] + all_files)
+                        command : [env, 'ctags', '--tag-relative=never', '-o', '@0@/tags'.format(project_source_root)] + all_files)
         endif
 endif
 


medhefgo:
No idea what ctags does. But why are relative paths bad?

evverx:
In this particular case they make that `tags` file almost unusable: https://github.com/systemd/systemd/pull/9028#issuecomment-390306378

evverx:
FWIW another option would be to just drop that altogether because `meson` has supported `ctags` natively since 0.53 was released and personally I'd probably prefer that route (because `git ls-tree` there isn't much different from `git ls-files` used by the systemd build system).

medhefgo:
Removing code size is always nice. You should check that the result is the same or better, though. Because one thing that nags me is that the compile_commands.json does not contain sd-boot stuff, because meson tells ninja to create a reduced compdb. I always have to create one manually from time to time.

evverx:
> You should check that the result is the same or better, though

I think it's generally better in the sense that it even works with systemd tarballs because meson switch to a Python alternative of ls when `.git` isn't present. But since it uses `ls-tree HEAD` it kind of ignores new files and that's where the systemd build systemd shines :-) I'm not sure that particular use case warrants that custom `ctags` target though.

> Because one thing that nags me is that the compile_commands.json does not contain sd-boot stuff, because meson tells ninja to create a reduced compdb.

That's kind of bad. Some static analyzers I use rely on compile_commands.json so it seems they completely miss the sd-boot part because of that.

evverx:
I reported that meson issue in https://github.com/mesonbuild/meson/issues/9840. Until it's fixed it seems `ninja -C ./build/ -t compdb` has to be called every time :-(

keszybz:
> FWIW another option would be to just drop that altogether because `meson` has supported `ctags` natively since 0.53 was released and personally I'd probably prefer that route (because `git ls-tree` there isn't much different from `git ls-files` used by the systemd build system).

We looked at the native implementation when it was provided, and it didn't seem to be as nice as what we had with `git ls-files`. But maybe it's worth to look at it again. If we can retire our custom target with no (or little) loss of functionality, then I'd be for it.

PR #21553, opened 2021-11-29
network: wireguard: add routes to AllowedIPs=

yuwata:
Closes #14176.

PR #22136, opened 2022-01-16
network: wireguard: disable adding routes to allowed ips by default

### LINKED ISSUE #21964: default route for wireguard interface automatically set when allowedIPs contains ::/0 or 0.0.0.0/0 since v250

yuwata:
As setting such routes may break existing setups.

Closes #21964.

### DIFF
--- a/man/systemd.netdev.xml
+++ b/man/systemd.netdev.xml
@@ -1575,14 +1575,14 @@
         <term><varname>RouteTable=</varname></term>
         <listitem>
           <para>The table identifier for the routes to the addresses specified in the
-          <varname>AllowedIPs=</varname>. Takes the special value <literal>off</literal>, one of the
-          predefined names <literal>default</literal>, <literal>main</literal>, and
-          <literal>local</literal>, names defined in <varname>RouteTable=</varname> in
+          <varname>AllowedIPs=</varname>. Takes a negative boolean value, one of the predefined names
+          <literal>default</literal>, <literal>main</literal>, and <literal>local</literal>, names
+          defined in <varname>RouteTable=</varname> in
           <citerefentry><refentrytitle>networkd.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry>,
           or a number in the range 1…4294967295. When <literal>off</literal> the routes to the
           addresses specified in the <varname>AllowedIPs=</varname> setting will not be configured.
-          Defaults to <literal>off</literal>. This setting will be ignored when the same setting is
-          specified in the [WireGuardPeer] section.</para>
+          Defaults to false. This setting will be ignored when the same setting is specified in the
+          [WireGuardPeer] section.</para>
         </listitem>
       </varlistentry>
       <varlistentry>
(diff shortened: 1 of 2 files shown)


yuwata:
As reported in #21964, the PR breaks some existing setups. But It is not necessary to revert whole commits of PR #21553. But disabling the feature by default is enough. Let's fix the incompleteness of the PR #21553 later.

bluca (approved):

zx2c4:
The whole feature should have been reverted instead.

man/systemd.netdev.xml
```diff
@@ -1575,14 +1575,14 @@
         <term><varname>RouteTable=</varname></term>
         <listitem>
           <para>The table identifier for the routes to the addresses specified in the
-          <varname>AllowedIPs=</varname>. Takes the special value <literal>off</literal>, one of the
-          predefined names <literal>default</literal>, <literal>main</literal>, and
-          <literal>local</literal>, names defined in <varname>RouteTable=</varname> in
+          <varname>AllowedIPs=</varname>. Takes a negative boolean value, one of the predefined names
```
flokli, line 1578: My brain fails to parse this, sorry. How does a negative boolean value for the `RouteTable=` field look like?

man/systemd.netdev.xml
```diff
@@ -1682,9 +1682,9 @@
...
+          <literal>default</literal>, <literal>main</literal>, and <literal>local</literal>, names
+          defined in <varname>RouteTable=</varname> in
```
flokli, line 1687: same

man/systemd.netdev.xml
yuwata: E.g. `RouteTable=no` or `RouteTable=off`.

yuwata: Actually, it looks spurious comparing with other `RouteTable=` settings in other sections, e.g. `[DHCPv4]`. Maybe, we should add, say `RouteToAllowedIPs=` boolean setting. WDYT?

flokli: I really don't want to further add more knobs until there have been some wider design discussions.

But I agree the field names are not chosen in a very self-explaining fashion. Wireguard interface configuration doesn't have routing tables per se, but this flag will cause wireguard to add routes *derived from allowedIPs* (and specify the name of the routing table to add to).

So I'd probably name it the other way round - `AllowedIPsToRoutingTable`, but again, let's sit down and have these discussions (including @zx2c4, if that bridge isn't burnt down), before adding more knobs. This again would be a change in the interface.

PR #22137, opened 2022-01-16
core: update log message

yuwata:
Fixes CID#1469009.

PR #22138, opened 2022-01-16
ci: install libbpf-dev in the unit_tests workflow

evverx:
Those dependencies are also used by Coverity and Codeql so
it should be installed there to get them to analyze that code.

Judging by https://github.com/systemd/systemd/pull/22137 it seems
to be working.


mrc0mmand (approved):
LGTM, thanks!

PR #22140, opened 2022-01-16
boot: Beep n times for n-th entry

medhefgo:
As suggested by https://github.com/systemd/systemd/pull/22122#issuecomment-1013851061.


src/boot/efi/boot.c
```diff
@@ -727,9 +727,10 @@ static BOOLEAN menu_run(
                         ST->ConOut->OutputString(ST->ConOut, clearline + 1 + x + len);
                 }
 
-                if (do_beep) {
+                /* Beep several times so that the selected entry can be distinguished. */
+                for (UINTN i = 0; config->beep && i <= idx_highlight; i++) {
                         beep();
-                        do_beep = FALSE;
+                        BS->Stall(400 * 1000);
```
bluca: Maybe move this inside `beep()` so that there's no stall on !x86?

medhefgo: Oops. Totally slipped my mind.

bluca (approved):

alex19EP (approved):
tested on qemu. everything works as it should. the beep is well distinguished and easy to count.

Issue #4784, opened 2016-12-01
systemd-stdio-bridge is not documented


fsateler:
### Submission type

  - [x] Bug report

### systemd version the issue has been seen with

> git master

### In case of bug report: Expected behaviour you didn't see

A man page for all binaries installed into `/usr/bin`

### In case of bug report: Unexpected behaviour you saw

`systemd-stdio-bridge` is missing a manpage.

poettering:
it's an internal tool, it's never supposed to be called directly. The only reason it's in $PATH, is that we invoke it on a remote system for the ssh dbus transport, where we don't know whether systemd's private directory is /usr/lib/systemd or /lib/systemd or anything else, and hence prefer relying on the search path.

I don't think this should have a man page. It's really not user-facing, and we generally don't document all internal files of systemd via man pages.

fsateler:
I think it should be moved to `$rootlibdir/systemd` then, and always invoked via `/lib/systemd/systemd-stdio-bridge`. Or does systemd want to support systems with merged-usr but where `/lib` does not point to `/usr/lib` ?

> I don't think this should have a man page. It's really not user-facing, and we generally don't document all internal files of systemd via man pages.

By virtue of placing it in a public path, you make it user-facing.

boucman:
would it be possible to mention in --help that it's an internal systemd tool ?

madscientist42:
Even internal use tools need DOCUMENTATION, guys.

PR #22141, opened 2022-01-16
systemd-stdio-bridge: add manpage

bluca:


medhefgo:
As someone who doesn't know what this prog does: this manpage is not enlightening at all. Maybe an example section would help?

bluca:
Would be nice, problem is, I don't know either besides what's written here XD

mbiebl:
See #4784

bluca:
If it has a `--help` it can have a manpage - I think it's fine to start with a minimal one and then expand it later if we can

yuwata:
At least, this is a good start point. Let's improve it later if necessary. Merging.

PR #22147, opened 2022-01-17
More docs for systemd-stdio-bridge

keszybz:


yuwata (approved):
LGTM.

bluca (approved):

behrmann:
Wow, things I never knew existed. Is 
> -M --machine=CONTAINER Name of local container to connect to

in the help (still?) the case? The man page mentions SSH.

keszybz:
> Wow, things I never knew existed. Is
> 
> > -M --machine=CONTAINER Name of local container to connect to
> 
> in the help (still?) the case? The man page mentions SSH.

Various tools that support remote connections using -H will start ssh, and then call systemd-stdio-bridge remotely. The fact that systemd-stdio-bridge supports -M by itself is confusing, but actually unrelated ;)

PR #22149, opened 2022-01-17
Update systemd.netdev.xml

march1993:
Mind users to avoid use name gre0, gretap0 and erspan0 for GRE tunnels.

### DIFF
--- a/man/systemd.netdev.xml
+++ b/man/systemd.netdev.xml
@@ -94,13 +94,13 @@
           <entry>A dummy device drops all packets sent to it.</entry></row>
 
           <row><entry><varname>gre</varname></entry>
-          <entry>A Level 3 GRE tunnel over IPv4. See <ulink url="https://tools.ietf.org/html/rfc2784">RFC 2784</ulink> for details.</entry></row>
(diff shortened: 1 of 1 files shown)


man/systemd.netdev.xml
```diff
@@ -94,7 +94,7 @@
           <entry>A dummy device drops all packets sent to it.</entry></row>
 
           <row><entry><varname>gre</varname></entry>
-          <entry>A Level 3 GRE tunnel over IPv4. See <ulink url="https://tools.ietf.org/html/rfc2784">RFC 2784</ulink> for details.</entry></row>
+          <entry>A Level 3 GRE tunnel over IPv4. See <ulink url="https://tools.ietf.org/html/rfc2784">RFC 2784</ulink> for details. For GRE tunnel devices, names like gre0, gretap0 and erspan0 should not be used on certain platforms considering they would be created as dummy devices for compatibility reasons.</entry></row>
```
yuwata: Please move gretap0 and erspan0 to the corresponding lines below.
```
should not be used on certain platforms considering they would be created as dummy devices for compatibility reasons.
```
This sounds slightly ambiguous for me. How about the following?
```
should not be used, as the kernel creates the device with the name when the corresponding kernel module is loaded.
```

march1993:
@yuwata Am I doing right now?

keszybz:
It's be nicer to wrap the interface names in quotes. Also "the device with the name" is unclear.

```diff
- Name gretap0 should not be used, as the kernel creates the device with the name
+ Name <literal>gretap0</literal> should not be used, as the kernel creates a device with this name
```

yuwata (approved):
Thanks. LGTM.

PR #22150, opened 2022-01-17
man: add more references for extensions to portablectl

### LINKED ISSUE #22146: systemd-portable (250) --extension documentation
The documentation for --extension=PATH in the portablectl manual page is unclear.

This feature functionality is also inconsistent with `ExtensionImages=` (systemd.exec)

I would argue that it should be interchangeable

ie if ExtensionImages=/var/lib/extensions/myext.raw works for a service then --extension=/var/lib/extensions/myext.raw should also work with a portable service

instead if fails with a confusing error:

```
root@myguest1:/usr/lib# portablectl attach /var/lib/portables/apache2.raw --runtime --extension=/var/lib/extensions/myext.raw --profile=trusted apache2
(Matching unit files with prefixes 'apache2'.)
AttachImageWithExtensions failed: Image '/var/lib/extensions/myext.raw' lacks extension-release data, refusing.
```
and I have no idea what that is trying to tell me....

bluca:
Link in systemd.io and systemd-sysext

Fixes #22146

### DIFF
--- a/man/portablectl.xml
+++ b/man/portablectl.xml
@@ -47,7 +47,8 @@
     <para>Portable service images are an efficient way to bundle multiple related services and other units together,
     and transfer them as a whole between systems. When these images are attached the local system the contained units
     may run in most ways like regular system-provided units, either with full privileges or inside strict sandboxing,
-    depending on the selected configuration.</para>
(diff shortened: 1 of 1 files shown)


yuwata (approved):
LGTM.

PR #22152, opened 2022-01-17
ci: drop debian on mkosi temporarily

evverx:
meson can't be installed there
```
root@ad3e557ad841:/# apt-get install -y meson
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 meson : Depends: python3-distutils but it is not installable
         Recommends: dpkg-dev but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

I'm not sure why Debian Unstable is used there. The systemd CI is
supposed to test systemd after all.

cc @DaanDeMeyer @mrc0mmand could you take a look?

### DIFF
--- a/.github/workflows/mkosi.yml
+++ b/.github/workflows/mkosi.yml
@@ -33,7 +33,6 @@ jobs:
       matrix:
         distro:
           - arch
-          - debian
           - ubuntu
           - fedora
           - opensuse


bluca:
No need to remove, these things happen in unstable, it will fix itself in a few hours - switch it from unstable to testing if you want to avoid transient issues like these

evverx:
>  these things happen in unstable, it will fix itself in a few hours

I receive notifications from GitHub when GHActions fail so glitches like that fill my inbox with emails that have nothing to do with `systemd`.

> switch it from unstable to testing if you want to avoid transient issues like these

I have no idea how the mkosi action works so dropping debian is the only thing I can do. I've just opened https://github.com/systemd/systemd/pull/22153 but I don't know whether it's enough to make it work.

bluca:
#22153 looks about right, let's close this in favour of that

PR #22153, opened 2022-01-17
ci: switch Debian from unstable to testing on mkosi

evverx:


mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -5,7 +5,7 @@
 
 [Distribution]
 Distribution=debian
-Release=unstable
+Release=bullseye
```
bluca: make this `testing` - bullseye is the stable release, but for CIs we want something newer, testing is more or less equivalent to fedora rawhide or archlinux or suse tumbleweed

evverx: Last time I checked the "testing" tag wasn't supported (I'm not sure whether it was mkosi though). I'll change it shortly

evverx: FWIW opensuse is flaky as well (partly because of the way the mkosi action works as far as I understand). It was briefly discussed in https://github.com/systemd/systemd/pull/21978#issuecomment-1004084735

evverx: Anyway, judging by
```
              Distribution: debian
                   Release: testing
                    Mirror: http://deb.debian.org/debian
     Use Host Repositories: no
```
it seems to be working

bluca (approved):

evverx:
It seems to be failing to boot because an infinite loop of some kind is being triggered:
```
systemd-boot@0x3E374000,0x3E390000
    Reboot Into Firmware Interface                                    Boot in 10 s.                                                                   Boot in 9 s.                                                                   Boot in 8 s.                                                                   Boot in 7 s.                                                                   Boot in 6 s.                                                                   Boot in 5 s.                                                                   Boot in 4 s.                                                                   Boot in 3 s.                                                                   Boot in 2 s.                                                                   Boot in 1 s.                                 3h...
BdsDxe: starting Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x3,0x0)
systemd-boot@0x3E374000,0x3E390000
```

evverx:
Judging by https://github.com/systemd/systemd/runs/4846586948?check_suite_focus=true it isn't reproducible with bullsye so I'd switch back to it for the time being

evverx:
I switched back to bullseye for now.

FWIW given that the last version of the PR wasn't approved I think it would make sense to automatically dismiss reviews by flipping the "dismiss approvals" setting
Screenshot 2022-01-18 at 02 18 11

The screenshot is from a repository where it's on. Here it's turned off as far as I can see

bluca:
https://github.com/systemd/mkosi/pull/886

evverx:
I switched back to "testing" and pointed mkosi to the commit where the bug is fixed

PR #18128, opened 2021-01-04
ci: move jobs from Travis CI to GitHub Actions

mrc0mmand:

PR #22159, opened 2022-01-18
boot: xbootldr improvements

medhefgo:


bluca (approved):

src/boot/efi/xbootldr.c
```diff
@@ -139,10 +133,12 @@ static EFI_STATUS try_gpt(
                 if (end < start) /* Bogus? */
                         continue;
 
-                *ret_part_number = i + 1;
-                *ret_part_start = start;
-                *ret_part_size = end - start + 1;
-                CopyMem(ret_part_uuid, &entry->UniquePartitionGUID, sizeof(*ret_part_uuid));
+                ret_hd->PartitionNumber = i + 1;
```
poettering, line 136: looks like a job for structured initialization, i.e.:

```c
*ret_hd = (HARDDRIVE_DEVICE_PATH) {
        …
};
```

Even if you have to `CopyMem()`in the signature field manually after wards

medhefgo: I think I tried something like this and things froze up for some reason.

medhefgo: Just tested this again to make sure. This doesn't work, neither does assigning the `Header` fields manually. It just freezes up and it doesn't make any bloody sense to me why.
```c
*ret_hd = (HARDDRIVE_DEVICE_PATH) {
        .Header = ret_hd->Header,
        .PartitionNumber = i + 1,
        .PartitionStart = start,
        .PartitionSize = end - start + 1,
        .MBRType = MBR_TYPE_EFI_PARTITION_TABLE_HEADER,
        .SignatureType = SIGNATURE_TYPE_GUID,
};
CopyMem(hd.Signature, &entry->UniquePartitionGUID, sizeof(hd.Signature));
´´´

bluca: it's because computers are cursed, and bootloaders doubly so

anitazha: Any relation to https://github.com/systemd/systemd/pull/18128#issuecomment-757041294 ? That's the last time I remember some struct initialization issues (due to non-zero-initialized padding)

medhefgo: That's probably it. If I make an identical `_packed_` struct, it's smaller. The way efi device paths are supposed to work is by looking at the length fields in the header part to find how large a given path node is. So I guess this would clobber over the next node. Good to know.

Issue #6279, opened 2017-07-04
machined: provide more details to polkit auth


lnussel:
### Submission type

 - Request for enhancement (RFE)

### systemd version the issue has been seen with

git master

### description

src/machine/machine-dbus.c doesn't pass any details to
bus_verify_polkit_async(). So custom polkit rules can't e.g. allow
only certain users to operate on certain machines using
org.freedesktop.machine1.shell.

I'd like to suggest to always pass the unit ie machine name for all
actions where it is applicable, like it's done for
org.freedesktop.systemd1.manage-units.

poettering:
A patch adding the props you need would be very welcome!

poettering:
Implemented by 09364a8043a2f9b698e49a172094d658ae289ac6. Closing.

PR #22160, opened 2022-01-18
machined: provide more details to polkit auth

lnussel:
completes #6279


lnussel:
example polkit file to allow users to manage machines following a specific naming scheme https://build.opensuse.org/package/view_file/home:lnussel:systemd/systemd-machined-userpolicy/20-machined.rules?expand=1

bluca (approved):

PR #22161, opened 2022-01-18
bootctl: Ignore boot entries (continue #22041)

kreijack:
NOTE: This is the continuation of the issue #22041, that I closed by mistake.

This topic was also discussed previously here #19600 (comment)

**bootctl list** shows a list of entries which is a sum of the ones synthesized from _/boot/efi/loaders/..._ and ones returned by **systemd-boot** (via the LoaderEntries efi variable). The information returned by **systemd-boot** are still synthesized from _/boot/efi/loaders/..._ boot at boot time.

The problem arise when (e.g.) after a removal of a kernel package, **bootctl** still shows the deleted entries because **systemd-boot** saw it at boot time.

Below an example of what happens:
```
$ bootctl list
Boot Loader Entries:
        title: Debian GNU/Linux bookworm/sid (5.15.12) (default)
           id: e84907d099904117b355a99c98378dca-5.15.12.conf
       source: /boot/efi/loader/entries/e84907d099904117b355a99c98378dca-5.15.12.conf
      version: 5.15.12
   machine-id: e84907d099904117b355a99c98378dca
        linux: /e84907d099904117b355a99c98378dca/5.15.12/linux
       initrd: /e84907d099904117b355a99c98378dca/5.15.12/initrd
      options: root=UUID=d7a06504-cc14-435d-a5df-674da09c2894 ro rootflags=subvol=@rootfs btrfsrollback=@rol>

        title: e84907d099904117b355a99c98378dca-5.15.4.conf
           id: e84907d099904117b355a99c98378dca-5.15.4.conf
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

        title: e84907d099904117b355a99c98378dca-5.15.7.conf
           id: e84907d099904117b355a99c98378dca-5.15.7.conf
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

        title: Reboot Into Firmware Interface
           id: auto-reboot-to-firmware-setup
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
```
The kernel 5.15.7 and 5.15.4 are removed but because **systemd-boot** saw it during the boot, **bootctl** continue to show them.

These patches change the **bootctl** behavior, so it ignores the information returned by **systemd-boot** related to _/boot/efi/loaders/..._
```
$ build/bootctl list
Boot Loader Entries:
        title: Debian GNU/Linux bookworm/sid (5.15.12) (default)
           id: e84907d099904117b355a99c98378dca-5.15.12.conf
       source: /boot/efi/loader/entries/e84907d099904117b355a99c98378dca-5.15.12.conf
      version: 5.15.12
   machine-id: e84907d099904117b355a99c98378dca
        linux: /e84907d099904117b355a99c98378dca/5.15.12/linux
       initrd: /e84907d099904117b355a99c98378dca/5.15.12/initrd
      options: root=UUID=d7a06504-cc14-435d-a5df-674da09c2894 ro rootflags=subvol=@rootfs btrfsrollback=@rol>

        title: Reboot Into Firmware Interface
           id: auto-reboot-to-firmware-setup
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
```

In any case, **bootctl** still considers the 'auto-xxx' boot entries
listed below:
```
  Boot entrie name                 Title
  -----------------------------    ------------------------------
  auto-osx                         macOS boot loader
  auto-windows                     Windows Boot Manager
  auto-efi-shell                   EFI Shell
  auto-efi-default                 EFI Default Loader
  auto-reboot-to-firmware-setup    Reboot Into Firmware Interface
```
The other entries that **systemd-boot** synthesizes (e.g. the ones loaded from
/efi/loader/entries/) can be synthesized by **bootctl** too, so no
information is lost.

This patches set is composed by two patches:
- the first one changes the behavior of **bootctl**, to ignore the some entries returned by systemd-boot
- the second patch, cleanup the code removing the parameter "only_auto" from boot_entries_augment_from_loader() which is always true.

Signed-off-by: Goffredo Baroncelli kreijack@inwind.it

### DIFF
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -1603,7 +1603,7 @@ static int verb_list(int argc, char *argv[], void *userdata) {
         else if (r < 0)
                 log_warning_errno(r, "Failed to determine entries reported by boot loader, ignoring: %m");
         else
-                (void) boot_entries_augment_from_loader(&config, efi_entries, true);
+                (void) boot_entries_augment_from_loader(&config, efi_entries);
 
         if (config.n_entries == 0)
                 log_info("No boot loader entries found.");
--- a/src/login/logind-dbus.c
+++ b/src/login/logind-dbus.c
@@ -3008,7 +3008,7 @@ static int boot_loader_entry_exists(Manager *m, const char *id) {
 
         r = manager_read_efi_boot_loader_entries(m);
         if (r >= 0)
-                (void) boot_entries_augment_from_loader(&config, m->efi_boot_loader_entries, true);
+                (void) boot_entries_augment_from_loader(&config, m->efi_boot_loader_entries);
 
         return boot_config_has_entry(&config, id);
 }
@@ -3166,7 +3166,7 @@ static int property_get_boot_loader_entries(
 
         r = manager_read_efi_boot_loader_entries(m);
         if (r >= 0)
-                (void) boot_entries_augment_from_loader(&config, m->efi_boot_loader_entries, true);
+                (void) boot_entries_augment_from_loader(&config, m->efi_boot_loader_entries);
 
         r = sd_bus_message_open_container(reply, 'a', "s");
         if (r < 0)
--- a/src/shared/bootspec.c
+++ b/src/shared/bootspec.c
@@ -759,8 +759,7 @@ int boot_entries_load_config_auto(
 
 int boot_entries_augment_from_loader(
                 BootConfig *config,
-                char **found_by_loader,
-                bool only_auto) {
+                char **found_by_loader) {
 
         static const char *const title_table[] = {
                 /* Pretty names for a few well-known automatically discovered entries. */
@@ -785,7 +784,12 @@ int boot_entries_augment_from_loader(
                 if (boot_config_has_entry(config, *i))
                         continue;
 
-                if (only_auto && !startswith(*i, "auto-"))
+                /*
+                 * consider the 'auto-' entries only, because the others
+                 * ones are detected scanning the 'esp' and 'xbootldr'
+                 * directories by boot_entries_load_config()
+                 */
+                if (!startswith(*i, "auto-"))
                         continue;
 
                 c = strdup(*i);
--- a/src/shared/bootspec.h
+++ b/src/shared/bootspec.h
@@ -76,7 +76,7 @@ static inline BootEntry* boot_config_default_entry(BootConfig *config) {
 void boot_config_free(BootConfig *config);
 int boot_entries_load_config(const char *esp_path, const char *xbootldr_path, BootConfig *config);
 int boot_entries_load_config_auto(const char *override_esp_path, const char *override_xbootldr_path, BootConfig *config);
-int boot_entries_augment_from_loader(BootConfig *config, char **list, bool only_auto);
+int boot_entries_augment_from_loader(BootConfig *config, char **list);
 
 static inline const char* boot_entry_title(const BootEntry *entry) {
         return entry->show_title ?: entry->title ?: entry->id;


yuwata (approved):
LGTM.

yuwata:
Maybe, this should be backported to stable branches? Feel free to drop the 'needs-stable-backport' label if it should not.

keszybz:
I dropped the tag for now… I think this is a change in behaviour that shouldn't be backported unless somebody asks for it.

yuwata:
OK. Thanks.

poettering:
Uh, I am not sure how I feel about this. The thing is that sd-boot might find different entries than we find from our side. The idea was to show everything we know from all sources. The thing is that sd-boot collects entries from various sources: automatic entries, Type 1 and Type 2 entries, from the ESP and XBOOTLDR partitions. Sooner or later it might show even more from other sources. From our systemd side/bootctl we only can find some of them easily, and there's a good chance things will be out of sync. Thus I think there's benefit if we present to the user *everything* discovered, regardless by who: by us in userspace, or by sd-boot.

Now I understand that it's confusing if we show stuff that isn#t there anymore. But I think that's a presentation issue, not an issue of actually hiding the stuff. I.e. show the entries, but mark them clearly as enrties we didn't discover from userspace (and conversely it might make sense to highlight items we only found from userspace, but not from sdboot).

i.e. maybe add a string "absent" or so to those we cannot find from userspace, and grey them out maybe even. And may be add a string "restricted" or so for those which only we found but sd-boot so far didn't. And then maybe highlight those in bright or so.

bootctl is in particular a tool to help debugging stuff, i.e. finding problems with the boot loader. introspecting the state of the boot loader is everything that "bootctl status" is about really, hence we should show everything we have, and hightlight the discrepancies as well as we can.

Hence, it would suck to hide the stuff discovered, we should just try to present the data better.

kreijack:
> The thing is that sd-boot collects entries from various sources: automatic entries, Type 1 and Type 2 entries, from the ESP and XBOOTLDR partitions. Sooner or later it might show even more from other sources. 

Could you elaborate a bit ? Looking at the code it seems that both "systemd-boot" and "bootctl" get the information from the same sources:
- the boot entries are got from ESP and XBOOTLDR
- the "foreign bootloader" are detected looking at the ESP
- the "reboot to firmware", is detected looking at the appropriate efi variable.

Unfortunately these checks are equal but performed by codes slightly different (due to different code history)

I see only two main differences:
1) sb-boot and bootctl see the same set of information in two different times: the former at boot time, the latter at the "actual time". 
2) bootctl may see a different ESP partition (e.g. if you are looking to look to a foreign hard disk coming from another PC)[what about checking the random seed to be sure that we are looking to the right partition ?]

Regarding the goal of bootctl, I invite you to consider bootctl a way to configure the "next-boot"; and in this case it is important to know which are the boot entries really available. 

Finally, I have a similar opinion about adding a switch to show only one "half" of information. The main differences is that looking at the "current" boot entries should be the default. The switch should enable all the entries highlighting the differencies.

poettering:
> Could you elaborate a bit ? Looking at the code it seems that both "systemd-boot" and "bootctl" get the information from the same sources:
> 
>     * the boot entries are got from ESP and XBOOTLDR

We at least *hope* these are the same sources. But it's a real hard problem to find the right devices from Linux userspace that the uefi firmware used before. We typically go via GPT partition GUID, but once that is not unique we have a problem.

So we *hope* we mean the same sources, but we don't know.

And even if we mean the same device, we can't necessarily be sure firmware picked the same ESP on it, in case there are multiple. And similar, we declare that if there are multiple XBOOTLDR partitions we use the first, but there are always bugs and given the code to search this stuff is quite different there's a good chance things might differ.

>     * the "foreign bootloader" are detected looking at the ESP

Here too, it's not just the ESP, the MacOS detection looks around on other partitions too.

The thing here is really: Linux userspace and sd-boot/firmware use *very* different codepaths, so there's a really good chance they might come to different results, even if we hope that usually don't.

> Unfortunately these checks are equal but performed by codes slightly different (due to different code history)

Not just that. On EFI we use UEFI APIs and parsers/device discovery/drivers/file system implementations and on Linux we user Linux APIs/parsers/discovery/drivers/fs. While they hopefully find the same devices, there's no guarantee… Even if we wanted to use the same code we can't really the worlds are just too different.

So we have to accept that there's a good chance things will differ there, and we should make the best of it.

And it's not just that, there's also a good chance that user configuration will fuck things up for us. THink a system where the admin mounts /efi/ manually (or /boot/efi/ like the fedora installer would mount it), and thus makes us believe that the ESP used is one thing, but then the BIOS actually booted from a different disk and noone noticed. This isn't even an unlikely scenario, there are plenty of reports of people who think simple mirror raid of two HDDs including the HDD should just work (of course it doesn't, and it's a bad idead, but that's besides the point here)

> I see only two main differences:
> 
>     1. sb-boot and bootctl see the same set of information in two different times: the former at boot time, the latter at the "actual time".
> 
>     2. bootctl may see a different ESP partition (e.g. if you are looking to look to a foreign hard disk coming from another PC)[what about checking the random seed to be sure that we are looking to the right partition ?]

This is always risky. I mean, there are even environments like OPAL disk encryption where the first bunch of sectors (which likely contain the ESP) can be virtually swapped out once a disk is decrypted, so that the early EFI boot process sees a different ESP contents then userspace eventually will...

> Regarding the goal of bootctl, I invite you to consider bootctl a way to configure the "next-boot"; and in this case it is important to know which are the boot entries really available.

That usecase makes a ton of sense no doubt. And I am not saying we shouldn't make behaviour nicer for this usecase, but I am definitely against hiding the info we got from the real boot loader.

i.e. there are better ways to address this: show *all* entries but mark the ones that are not accessible from userspace, for example because they might have been deleted, or for other reasons.

> Finally, I have a similar opinion about adding a switch to show only one "half" of information. The main differences is that looking at the "current" boot entries should be the default. The switch should enable all the entries highlighting the differencies.

I disagree: we should show all info by default, both the stuff sd-boot detected and that we detected, but we should highlight the cases where things don't match up, so that the user of bootctl knows about it and knows which items are good candidates are likely good for "next-boot", and which ones are not visible from userspace.

Or to say this differently: I invite you to see things from the perspective of someone who wants to debug their system, and just wants to know what was discovered at all, i.e. don't narrow yourself on the "next-boot" scenarion.

After all "bootctl" is a generic, relatively low-level tool. That's quite different from a UI tool. For example, if i'd be a GNOME hacker adding a reboot menu to gdm I'd only show the entries I am really sure exist, too, of course, but the low-level tool should show the others too.

kreijack:
Ok, let me to summarize:
- sd-boot and bootctl look to the same path but from different environments and in different times. So for the following reasons these might get different results:
1) they look to different path:
    possible mitigation:
         - if bootctl found different potential candidates (e.g. multiple partitions with EFI/XBOOTLDR signature) then warn 
         - check random seed stored in UEFI: if a seed (from sd-boot or bootctl) is found AND if these are different then warn 
         - (optional) if a warning is raised, stop and ask the user for explicit path
2) they look in different times
    possible mitigation:
         - if an entry is not found anymore (in bootctl list), mark it as "potential unavailable" in the list
         - exclude "osx" (and others ?) entry from this check, because these rules are too strange and the risk of false positives are too high (?)

Comments ?

keszybz:
bootctl is not a debugging tool foremost: it is a cli tool with prettified user output to introspect and control boot options. I think it is OK to allow extra output and warn about inconsistencies. In particular, I think we should DTRT in the common case of a normal system. It just isn't useful to show entries that have been removed since boot. The checks suggested by @kreijack above seem quite reasonable.

PR #22162, opened 2022-01-18
path_id: ignore devices with GENHD_FL_HIDDEN flag

ematsumiya:
Block devices that has the GENHD_FL_HIDDEN flag are not registered
as user facing devices, but are available for kernel I/O submission.

For NVMe, /dev/nvmeXcYnZ are NVMe multipath path devices, created
with the GENHD_FL_HIDDEN flag, that are not supposed to be parsed
or used by userspace tools.

The NVMe driver already avoids emitting uevents for these kind of
devices, but the problem is that these devices exists in sysfs,
making systemd/udev throw warnings when it can't parse them.

This commit makes udev path_id ignore NVMe devices that were created
with the GENHD_FL_HIDDEN flag (/sys/block/nvme*/hidden == 1).

Signed-off-by: Enzo Matsumiya <ematsumiya@suse.de>

### DIFF
--- a/src/udev/udev-builtin-path_id.c
+++ b/src/udev/udev-builtin-path_id.c
@@ -630,7 +630,17 @@ static int builtin_path_id(sd_device *dev, sd_netlink **rtnl, int argc, char *ar
                         supported_transport = true;
                         supported_parent = true;
                 } else if (streq(subsys, "nvme")) {
-                        const char *nsid;
+                        const char *nsid, *hidden;
+
+			/*
+			 * NVMe multipath paths (created with the GENHD_FL_HIDDEN
+			 * flag) exist for kernel I/O submission, but they're
+			 * not supposed to be accessed/used by userspace tools.
+			 * Thus, ignore such devices.
+			 */
+                        if (sd_device_get_sysattr_value(dev, "hidden", &hidden) >= 0)
+                                if (!strncmp(hidden, "1", 2))
+                                        break;
 
                         if (sd_device_get_sysattr_value(dev, "nsid", &nsid) >= 0) {
                                 path_prepend(&path, "nvme-%s", nsid);


ematsumiya:
It should be noted that, while the GENHD_FL_HIDDEN flag is generic to the block layer, i.e. this commit could verify this flag for whatever block subsystem device, NVMe is the only driver using it currently.

But, of course, I can adjust the patch if handling any block device is required.

src/udev/udev-builtin-path_id.c
```diff
@@ -630,7 +630,12 @@ static int builtin_path_id(sd_device *dev, sd_netlink **rtnl, int argc, char *ar
                         supported_transport = true;
                         supported_parent = true;
                 } else if (streq(subsys, "nvme")) {
-                        const char *nsid;
+                        const char *nsid, *hidden;
+
+                        if (sd_device_get_sysattr_value(dev, "hidden", &hidden) >= 0)
+                                /* NVMe multipath path (GENHD_FL_HIDDEN); ignore */
```
bluca: maybe add a comment explaining _why_ it is ignored?

ematsumiya: Force-pushed with an explanation that's a lean version of the commit message. Let me know how that works.

bluca (approved):

yuwata:
> The NVMe driver already avoids emitting uevents for these kind of
devices, but the problem is that these devices exists in sysfs,
making systemd/udev throw warnings when it can't parse them.

Does it mean that the sysfs directory for the device has `uevent` file, but does not emit uevents through netlink?

In what path, the sysfs directory is parsed and emit warning??

yuwata:
I mean, udevd process devices only when a uevent is received through netlink. If the device does not emit any uevents, then I have no idea why udevd tries to process it...

yuwata:
Or does the device emits uevents only when an event string is written to the 'uevent' file?

src/udev/udev-builtin-path_id.c
```diff
@@ -630,7 +630,17 @@ static int builtin_path_id(sd_device *dev, sd_netlink **rtnl, int argc, char *ar
...
                 } else if (streq(subsys, "nvme")) {
-                        const char *nsid;
+                        const char *nsid, *hidden;
+
+			/*
+			 * NVMe multipath paths (created with the GENHD_FL_HIDDEN
+			 * flag) exist for kernel I/O submission, but they're
+			 * not supposed to be accessed/used by userspace tools.
+			 * Thus, ignore such devices.
+			 */
```
yuwata, line 640: Not a matter, but I'd like to the lines have the same indentation as the code.

yuwata (commented):
(Tentatively drop the green label, until the questions are answered.)

src/udev/udev-builtin-path_id.c
```diff
@@ -630,7 +630,17 @@ static int builtin_path_id(sd_device *dev, sd_netlink **rtnl, int argc, char *ar
...
+                        if (sd_device_get_sysattr_value(dev, "hidden", &hidden) >= 0)
+                                if (!strncmp(hidden, "1", 2))
```
keszybz, line 642: `if (streq(hidden, "1"))`

ematsumiya:
> Does it mean that the sysfs directory for the device has `uevent` file, but does not emit uevents through netlink?
> 
> In what path, the sysfs directory is parsed and emit warning??

This was originally a customer reported issue, and I only had logs available for assessment. My initial assumption was they had path_id running from some custom udev rule (because 

Now, looking at further data provided, it doesn't seems to be the case (even though I *could* reproduce the issue with some manual tweaks, hence this patch).

So even though I don't think having this patch would hurt, I might be chasing the wrong bug here. Please leave it on hold while I get the required information from our customer.

Thanks.

poettering:
Urks, what a mess. Devices that exist in sysfs but for which there are no uevents? What a half-assed approach is that? Either make them full devices or hide them entirely, but providing a half-broken exposure to userspace is a mess.

poettering:
Quite frankly this looks like kernel oversight, i.e. judging by the comments in this thread: https://patchwork.kernel.org/project/linux-block/patch/20171018165258.23212-7-hch@lst.de/

the device should be gone entirely from userspace view if the flag is set, i.e. GENHD_FL_HIDDEN should also hide the device from sysfs.

Either way, i am pretty sure doing this in path_id seems wrong to me. The kernel should hide the devices fully. (and even if the kernel people disagree, then we should filter the whole set of devices early via udev rules, not late in path_id)

@hreinecke do you know more about this flag?

Anyway for now I'd consider this a kernel issue, not something we should tape over from userspace.

ematsumiya:
@poettering you're correct. I'm verifying things on kernel side now, thanks. Dropping PR.

PR #22163, opened 2022-01-19
Improve ExtensionImages validation

bluca:
Accept hosts without VERSION_ID/SYSEXT_LEVEL (rolling distros), but refuse if there isn't at least an ID.


keszybz:
LGTM.

PR #22167, opened 2022-01-19
Test udev storage debug

yuwata:
cc @mrc0mmand

### DIFF
--- a/test/TEST-17-UDEV/test.sh
+++ b/test/TEST-17-UDEV/test.sh
@@ -5,6 +5,7 @@ set -e
 TEST_DESCRIPTION="UDEV"
 IMAGE_NAME="default"
 TEST_NO_NSPAWN=1
+TEST_SAVE_JOURNAL=yes
 
 # shellcheck source=test/test-functions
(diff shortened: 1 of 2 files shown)


mrc0mmand:
If you want, I could setup a test job which would run only TEST-64, so you don't have to wait for all tests to finish to get the results.

yuwata:
@mrc0mmand If possible, could you give me the journal of TEST-64 in these successful run? E.g. https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-centos8/5883/artifact//systemd-centos-ci/index.html.

mrc0mmand:
> @mrc0mmand If possible, could you give me the journal of TEST-64 in these successful run? E.g. https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-centos8/5883/artifact//systemd-centos-ci/index.html.

If you push a debug-only commit which changes "fail" to "yes" here:
https://github.com/systemd/systemd/blob/b8387d24fefa9ce70142592b34b8a27bcc339448/test/TEST-64-UDEV-STORAGE/test.sh#L15
it should do the trick.

PR #22175, opened 2022-01-19
kernel-install: add support for KERNEL_INSTALL_INITRD_GENERATOR and KERNEL_INSTALL_STAGING_AREA

keszybz:
This should be reviewed together with the corresponding PR to mkosi-initrd, so I'm marking it as a draft for now.

### DIFF
--- a/src/kernel-install/90-loaderentry.install
+++ b/src/kernel-install/90-loaderentry.install
@@ -40,6 +40,8 @@ fi
 
 case "$COMMAND" in
     remove)
+        [ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] && \
+            echo "Removing $BOOT_ROOT/loader/entries/$MACHINE_ID-$KERNEL_VERSION*.conf"
         exec rm -f \
             "$BOOT_ROOT/loader/entries/$MACHINE_ID-$KERNEL_VERSION.conf" \
             "$BOOT_ROOT/loader/entries/$MACHINE_ID-$KERNEL_VERSION+"*".conf"


johannbg:
Sorry I dont follow this change. If the intent is to use other generators dont install dracut.

keszybz:
It should be completely fine to have dracut installed, but not actually used. It may be pulled in through packaging dependencies, or the user might install it just out of curiosity. The fact that a program is installed should be independent of whether it is enabled.

src/kernel-install/kernel-install
```diff
@@ -128,10 +128,18 @@ fi
 
 ENTRY_DIR_ABS="$BOOT_ROOT/$MACHINE_ID/$KERNEL_VERSION"
 
+# Provide a directory where to store generated initrds
+cleanup() {
```
yuwata, line 132: Where is this used??

keszybz: Ooops, there was supposed to be `trap` line for this.
/me checks, and indeed there are 500 snapshot directories in /tmp

johannbg:
> It should be completely fine to have dracut installed, but not actually used. It may be pulled in through packaging dependencies, or the user might install it just out of curiosity. The fact that a program is installed should be independent of whether it is enabled.

I would have to disagree and from my pov that would be a poor OS design if was the case. 

A package dependency pulls in a program then obviously something in that program is depentent on it ( if not the dependency is wrong ) and programs that aren't being used should not be installed in the first place ( if that's the case it's just sitting there wasting space and said program existing on a system could even become a security threat ).

If a program is installed then upstream should make the assumption that, that program is being used/enabled either by the vendor or by the fact that the user shows the intent of using the program by manually installing a program so he should have to disable it if he's not planning on using it, then obviously if the user intends on not using it he could just as well remove it and upstream could also rely on that being the case and even go as far as leave out any option of the program being capable of being disabled  ( as in expect it to be simply removed instead ).

johannbg:
> > 
> 
> That's a blatant nonsense. By that logic, e.g., `systemctl enable/disable/mask/unmask` is not needed, because the user can just uninstall the service...

No it's not " a blantant nonsense" and the ability that users could decouple systemd, uninstall the program ( and related service it added ) and continue to use whatever functionality said part of systemd replaced it with ( or some other program that was installed and provide type units ), was and is still is a sought after feature. 

> It also completely ignores development needs: developers often need a way to quick-test programA even if programB is installed (because it's a part of the system).

So developers writing their own requirements is impassable in this day and age o_O have you tried throwing more hardware at at? :smile: .

> And last but not least, [dracutdevs/dracut#1691](https://github.com/dracutdevs/dracut/pull/1691) changes a kernel-install script. Or let's call it a module. Well, dracut modules _can_ be enabled/disabled on demand. So how are kernel-install modules different?

Here's a news flash the sole purpose of generators to exist in the first place ( be it dracut or some other generator ) is guess what generate!  so comparing disabling it's core reason for being and twist that into a comparision with one of the module a generator ships since it fits the narrative being pushed is arguably "a blatant nonsense". 

If it's just the .install file why is the solution to simply use/ship an ,install  that takes precedence over what dracut and any other generator ship? No let's glitter prefixes all over the places and force/expect upstream generators to adopt that because "my tool/program" needs it!

Why change something in systemd, dracut and potentially any other generators when they just work fine with systemd and systemd work fine with other generators. 

This provides zero benefit to systemd and generators and solves nothing in systemd or for generators that exist out there.

yuwata:
The discussion here sounds slightly abstract for me...

IIUC, this and related PRs are for supporting multiple initrd generators, dracut, mkinitcpio, and under developing mkosi-initrd.
The purpose sounds not bad for me, and sounds it has actual benefit in general for me.

As already commented by @dtardon in the above, The change for dracut https://github.com/dracutdevs/dracut/pull/1691 does not change anything on the basic concept of dracut. It just updates kernel-install modules (plugins?), and such change is always necessary in general when we try to change something on kernel-install.
If the basic dracut functionality needs to be changed with this PR, then that's questionable. But it does not...

keszybz:
Updated to fix the missing `trap` that @yuwata noticed. I think the overall design is OK, so I'm removing the draft status.

bluca (approved):
Code looks reasonable to me, but not a kernel-install user

src/kernel-install/90-loaderentry.install
```diff
@@ -77,14 +77,10 @@ else
...
-        echo "+mkdir -v -p $ENTRY_DIR_ABS"
-        mkdir -v -p "$ENTRY_DIR_ABS"
-    else
-        mkdir -p "$ENTRY_DIR_ABS"
-    fi
-fi
+[ -d "$ENTRY_DIR_ABS" ] || {
+    echo "Error: entry directory '$ENTRY_DIR_ABS' does not exist" >&2
+    exit 1
+}
```
daandemeyer: I don't mind switching to this style of ifs but let's do it in a separate commit and all at once instead of using two different styles?

src/kernel-install/90-loaderentry.install
```diff
@@ -88,7 +90,8 @@ install -g root -o root -m 0644 "$KERNEL_IMAGE" "$ENTRY_DIR_ABS/linux" || {
 }
 
 shift "$INITRD_OPTIONS_SHIFT"
-for initrd; do
+# All files listed as arguments, and staged files called "initrd*" are installed as initrds.
+for initrd in "$@" "${KERNEL_INSTALL_STAGING_AREA}"/initrd*; do
```
daandemeyer, line 96: Isn't the usual name "initramfs-\<version\>.img"? Shouldn't we use that instead of initrd?

keszybz: No, we have a separate directory for the boot loader entry, so the names are usually `linux` and `initrd`. Longer names are used when everything is squished into the same directory, which we are not doing here.

src/kernel-install/90-loaderentry.install
keszybz: I backed out the style change. It doesn't make the code shorter, so we might just as well keep the old style.

PR #22176, opened 2022-01-19
boot: Use -ffile-prefix-map when present

medhefgo:
This should make sure the stub elf binary is reproducible.

Fixes: #22157

### DIFF
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -210,7 +210,17 @@ if get_option('b_lto')
 endif
 
 foreach arg : get_option('c_args')
-        if arg in ['-Werror', '-g', '-ggdb', '-O1', '-O2', '-O3', '-Og', '-Os', '-DNDEBUG', '-flto', '-fno-lto']
+        if arg in [
+                '-DNDEBUG',
+                '-fno-lto',
+                '-g', '-ggdb',
+                '-O1', '-O2', '-O3', '-Og', '-Os',
+                '-Werror',
+           ] or arg.split('=')[0] in [
+                '-ffile-prefix-map',
+                '-flto',
+           ]
+
                 message('Using "@0@" from c_args for EFI compiler'.format(arg))
                 efi_cflags += arg
         endif


medhefgo:
This is my best guess why it's not reproducing. We didn't use to pick up `-g`, so the output didn't have debug info (and hence no file names in it).

I am having a hard time getting reprotest to work, so this is a shot at the dark after having stared way too much at the provided repro logs.

bluca:
Thanks, running here: https://salsa.debian.org/bluca/systemd/-/pipelines/339063

probably the easiest way to test this is to create an account on Salsa and fork the repo, and build on my repro branch - you'll have to add commits as discrete patches in debian/patches but it shouldn't be too much trouble (git format-patch and then list the patchname in debian/patches/series)

medhefgo:
So the build-id is now the same, but the gnu_debuglink is still different. Dunno what debian build system does to cause it to be generated in the first place (or added after the build), because we do not generate one ourselves.

bluca:
Indeed:

```
│ │ ├── ./usr/lib/systemd/boot/efi/linuxx64.elf.stub
│ │ │┄ File has been modified after NT_GNU_BUILD_ID has been applied.
│ │ │ ├── readelf --wide --decompress --hex-dump=.gnu_debuglink {}
│ │ │ │ @@ -1,7 +1,7 @@
│ │ │ │  
│ │ │ │  Hex dump of section '.gnu_debuglink':
│ │ │ │    0x00000000 36376361 36323036 61666638 36613862 67ca6206aff86a8b
│ │ │ │    0x00000010 35396132 33303163 66666561 32623634 59a2301cffea2b64
│ │ │ │    0x00000020 64316236 61652e64 65627567 00000000 d1b6ae.debug....
│ │ │ │ -  0x00000030 36578ec3                            6W..
│ │ │ │ +  0x00000030 a7d00fc6                            ....
```

This is very strange...

bluca (approved):
But this improves things, so let's get it merged once the CI is done

PR #22180, opened 2022-01-19
test-sd-device-monitor: add one more testcase

yuwata:
This is mostly for #21603.

### DIFF
--- a/test/units/testsuite-17.sh
+++ b/test/units/testsuite-17.sh
@@ -5,6 +5,13 @@ set -o pipefail
 
 : >/failed
 
+if [[ -x /usr/lib/systemd/tests/test-sd-device-monitor ]]; then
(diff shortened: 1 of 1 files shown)


yuwata:
Ugh, I expected the test failed on focal-ppc64el, but the test is skipped on it...

yuwata:
Ugrrrr, I noticed that the TEST-02-UNITTESTS run only on nspawn on ubuntu CIs...

yuwata:
BTW, looking at https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/ppc64el/s/systemd-upstream/20220120_095601_3e5d7@/log.gz, the TEST-02-UNITTESTS passes on Ubuntu ppc64el on nspawn. Why does it still disabled?

yuwata:
Ah, maybe deny-list-ubuntu-ci vs deny-list-upstream-ci does a trick?
https://salsa.debian.org/systemd-team/systemd/-/blob/debian/master/debian/tests/upstream

PR #22187, opened 2022-01-20
macro: workaround for false-positive use-after-free warning by GCC-12

yuwata:


evverx:
Looking at https://gcc.gnu.org/gcc-12/changes.html it seems it hasn't been released yet. I wonder if it would make sense to wait for it to be released before adding workarounds?

yuwata:
yeah, and this does not work. Closing.

evverx:
As far as I understand it was triggered on fedora-rawhide-i386 (where systemd keeps failing to build). I think I'd just switch to fedora-35 or something like that on Packit. It's supposed to test `systemd` after all :-) @mrc0mmand what do you think?

PR #22189, opened 2022-01-20
ci: switch to fedora-35 on i386 on Packit

evverx:
systemd seems to be failing to compile there with gcc-12 but considering
that gcc-12 hasn't been released yet it doesn't seem to make sense
to add workarounds to get it to compile there. Until gcc-12 is
stabilized it should be enough to build systemd on fedora-35 to
make sure it's buildable on i386.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -38,6 +38,6 @@ jobs:
   metadata:
     targets:
     - fedora-rawhide-aarch64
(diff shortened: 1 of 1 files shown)


yuwata:
Opened #22191. Let's revert the commit when the issue is resolved.

PR #22195, opened 2022-01-20
Add unit specifiers for fragment path and directory

keszybz:


ams-tschoening:
Let's check if I understand things correctly... For the following example:

```bash
root@potsdam:/etc/systemd/system# ls -lisa
total 92
5288 0 drwxr-xr-x 1 root root 1994 Feb  5  2021 .
 566 0 drwxr-xr-x 1 root root  176 Sep 30 21:57 ..
5297 4 lrwxrwxrwx 1 root root   71 May 21  2018 ams_actiond.service -> /usr/local/AM-SoFT/de.am_soft.actiond/init/systemd/ams_actiond.service
```

* `%y` resolves to `/usr/local/AM-SoFT/de.am_soft.actiond/init/systemd/ams_actiond.service`
* `%Y` resolves to `/usr/local/AM-SoFT/de.am_soft.actiond/init/systemd`

Is that correct?

keszybz:
As long as `ams_actiond.service` is a real file. If `ams_actiond.service` would be a symlink, the symlink would be resolved until we reached a real file.

src/shared/specifier.c
```diff
@@ -212,22 +213,10 @@ int specifier_architecture(char specifier, const void *data, const char *root, c
 }
 
 static int specifier_os_release_common(const char *field, const char *root, char **ret) {
-        char *t = NULL;
-        int r;
+        /* Note: fields in /etc/os-release might quite possibly be missing, even if everything is entirely
+         * valid otherwise. We'll might return an empty value or NULL in that case. */
```
bluca: /We'll might/We'll/ - I think?

src/shared/specifier.c
```diff
@@ -212,22 +213,10 @@ int specifier_architecture(char specifier, const void *data, const char *root, c
 }
 
 static int specifier_os_release_common(const char *field, const char *root, char **ret) {
```
bluca: is there any point in having this wrapper after this change?

man/systemd.unit.xml
```diff
@@ -2098,6 +2098,16 @@ Note that this setting is <emphasis>not</emphasis> influenced by the <varname>Us
...
           <xi:include href="standard-specifiers.xml" xpointer="w"/>
           <xi:include href="standard-specifiers.xml" xpointer="W"/>
+          <row>
+            <entry><literal>%y</literal></entry>
+            <entry>The path to the fragment</entry>
+            <entry>This is the path where the main part of the unit file is located. For linked unit files, the real path outside of the unit search directories is used. For units that don't have a fragment file, this specifier will return an error.</entry>
+          </row>
+          <row>
+            <entry><literal>%Y</literal></entry>
+            <entry>The direcotry of the fragment</entry>
```
bluca: directory

man/systemd.unit.xml
```diff
@@ -2098,6 +2098,16 @@ Note that this setting is <emphasis>not</emphasis> influenced by the <varname>Us
           <xi:include href="standard-specifiers.xml" xpointer="V"/>
           <xi:include href="standard-specifiers.xml" xpointer="w"/>
           <xi:include href="standard-specifiers.xml" xpointer="W"/>
+          <row>
+            <entry><literal>%y</literal></entry>
+            <entry>The path to the fragment</entry>
+            <entry>This is the path where the main part of the unit file is located. For linked unit files, the real path outside of the unit search directories is used. For units that don't have a fragment file, this specifier will return an error.</entry>
```
bluca: What does "return an error" look like for a user with that in the unit? Will it fail to start?

bluca:
```
306/770 test-specifier                                                            FAIL             0.31s   killed by signal 6 SIGABRT
>>> SYSTEMD_LANGUAGE_FALLBACK_MAP=/root/src/src/locale/language-fallback-map PATH=/root/build:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin MALLOC_PERTURB_=120 /root/build/test-specifier
stderr:
/* test_specifier_escape */
/* test_specifier_escape_strv */
/* test_specifier_printf */
/* test_specifier_real_path */
Assertion 'r >= 0' failed at src/test/test-specifier.c:101, function test_specifier_real_path(). Aborting.
```

man/systemd.unit.xml
keszybz: Yes. It fails like with a bad specifier, e.g. %X (which is not defined).
In the first version I made %y and %Y return an empty string, but that seems like a pitfall. I think it's better to fail, then e.g. try to write to `%y/output`, which would then become `/output`.

src/shared/specifier.c
keszybz: That was potentially doubly conditonalized presumptory tense. I'll make it less.

src/shared/specifier.c
keszybz: Yeah, because the function interface is different. field,root,ret vs. root,field,ret.

man/systemd.unit.xml
bluca: Sounds good - can you maybe change this to say "will raise an error"? 'Return' sounded a bit confusing to me, since it's not something you call

src/shared/specifier.c
bluca: it's just used in the functions defined immediately below it - any reason not to call parse_os_release directly switching the parameters around?

keszybz: Indeed. Somehow I managed now to see that.

man/systemd.unit.xml
keszybz: Done.

bluca (approved):

yuwata (approved):
LGTM.

bluca:
Computer says no:

```
495/965 test-specifier                                                            FAIL             0.22s   killed by signal 11 SIGSEGV
>>> MALLOC_PERTURB_=78 PATH=/root/build:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin SYSTEMD_LANGUAGE_FALLBACK_MAP=/root/src/src/locale/language-fallback-map /root/build/test-specifier
stderr:
/* test_specifier_escape */
/* test_specifier_escape_strv */
/* test_specifier_printf */
/* test_specifier_real_path */
```

src/test/test-specifier.c
```diff
@@ -83,6 +84,47 @@ TEST(specifier_printf) {
...
+        _cleanup_free_ char *w = NULL;
+        int r;
+
+        r = specifier_printf("p=%p y=%y Y=%Y w=%w W=%W", SIZE_MAX, table, NULL, NULL, &w);
+        assert_se(r >= 0 || r == -ENOENT);
+        assert_se(w || r == -ENOENT);
+        puts(strnull(w));
+
+        /* /dev/initctl should normally be a symlink to /run/initctl */
+        if (files_same("/dev/initctl", "/run/initctl", 0))
```
yuwata: This may return negative errno. So, `> 0` is necessary.

5310:
Can someone please direct me to how I can make use of fragments in my systemd config? Or even what they _are_? 

I've been searching for many hours over multiple weeks now, and the only mention of "fragments" I can even find related to systemd is this one [DigitalOcean article](https://www.digitalocean.com/community/tutorials/how-to-use-systemctl-to-manage-systemd-services-and-units) that doesn't even explain it, and this PR that shows up in the [specifiers table](https://www.freedesktop.org/software/systemd/man/systemd.unit.htm) on the manual.

What _are_ fragments? I can't any information at all! 

Are they known by some other name in the ecosystem? It doesn't seem like they're drop-ins, or template variables. I don't even know the right words to ask questions elsewhere, I feel like! This is the only place that I can find that mentions fragments related to systemd! :pray:

mvirkkunen:
@5310 I want to know as well, so I'm going to hazard a guess because it's the Internet and somebody is guaranteed to come and correct me if I'm wrong.

From a cursory look at the source code I think "fragment" is just the same as "unit file", i.e. the `.service` or `.socket` or whatnot files that contain the actual unit configuration. All the file finding and parsing code seems to call loading one of those files "loading a fragment", so that would make sense.

Issue #22201, opened 2022-01-21
Boot failure in multipath setup: systemd tries to mount invalid device


mwilck:
**systemd version the issue has been seen with**

249 (very likely happens with current main, too)

**Used distribution**

openSUSE Tumbleweed, openSUSE Leap 15.4 beta

**Linux kernel version used** (`uname -a`)

`5.14.21-150400.3-default`

**architecture**

x86_64 (actually, any)

**Expected behaviour you didn't see**

This is observed with root on dm-multipath, in a VM, btrfs root with multiple subvolumes (standard openSUSE subvolume layout). After initrd processing and switching root, the system should boot successfully, mounting all subvolumes on the  dm-multipath device which was already mounted at `/`.

**Unexpected behaviour you saw**

After switching root, we observe this when `/dev/sdb` is processed by udev during coldplug:
```
[    9.233750] mptest systemd-udevd[621]: sdb: '/sbin/multipath -u sdb'(out) 'DM_MULTIPATH_DEVICE_PATH="1"'
[    9.351661] mptest systemd-udevd[621]: sdb: Device processed (SEQNUM=2023, ACTION=add)
...
[    9.383398] mptest systemd-udevd[625]: sdb2: Found 'b8:18' claiming '/run/udev/links/disk\x2fby-uuid\x2fa13cefb5-72b9-4ca4-abcb-17e970758c61'
[    9.383589] mptest systemd-udevd[625]: sdb2: Found 'b8:2' claiming '/run/udev/links/disk\x2fby-uuid\x2fa13cefb5-72b9-4ca4-abcb-17e970758c61'
[    9.386176] mptest systemd-udevd[625]: sdb2: Atomically replace '/dev/disk/by-uuid/a13cefb5-72b9-4ca4-abcb-17e970758c61'
```
 - `sdb` is correctly classified as multipath member. This implies `SYSTEMD_READY="0"`.
 - Before this happens, the symlink `/dev/disk/by-uuid/a13cefb5-72b9-4ca4-abcb-17e970758c61` points to a dm-multipath device (`dm-2` aka `b254:2`). But there's no message stating `Found 'b254:2' claiming …`.  This causes the udev worker to replace the existing, correct `by-uuid` symlink  with a wrong one pointing to `sdb2`.
  
Shortly later, the partition `sdb2` (which is the btrfs root volume) is processed by udev:
```
[...]

PR #22202, opened 2022-01-21
udevadm info --cleanup-db: don't delete information for kept db entries

mwilck:
Devices with the **db_persist** property won't be deleted during database
cleanup. This applies to dm and md devices in particular.
For such devices, we should also keep the files under `/run/udev/links`,
`/run/udev/tags`, and `/run/udev/watch`, to make sure that after restart,
udevd has the same information about the devices as it did before
the cleanup.

If we don't do this, a lower-priority device that is discovered in
the coldplug phase may take over symlinks from a device that persisted.
Not removing the watches also enables udevd to resume watching a device
after restart.

This fixes the failure to boot mentioned in #22201.

### DIFF
--- a/src/udev/udevadm-info.c
+++ b/src/udev/udevadm-info.c
@@ -256,6 +256,62 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
         }
 }
 
+/*
+ * Assume that dir is a directory with file names matching udev data base
+ * entries for devices in /run/udev/data (such as "b8:16"), and removes
+ * all files except those that haven't been deleted in /run/udev/data
+ * (i.e. they were skipped during db cleanup because of the db_persist flag).
+ * Returns true if the directory is empty after cleanup.
+ */
+static bool cleanup_dir_after_db_cleanup(DIR *dir, DIR *datadir) {
+        unsigned int kept = 0;
+
+        assert(dir && datadir);
+
+        FOREACH_DIRENT_ALL(dent, dir, break) {
+                struct stat data_stats, link_stats;
+
+                if (dot_or_dot_dot(dent->d_name))
+                        continue;
+                if (fstatat(dirfd(dir), dent->d_name, &link_stats, AT_SYMLINK_NOFOLLOW) < 0) {
+                        if (errno != ENOENT)
+                                kept++;
+                        continue;
+                }
+
+                if (fstatat(dirfd(datadir), dent->d_name, &data_stats, 0) < 0)
+                        (void) unlinkat(dirfd(dir), dent->d_name,
+                                        S_ISDIR(link_stats.st_mode) ? AT_REMOVEDIR : 0);
+                else
+                        /* The entry still exists under /run/udev/data */
+                        kept++;
+        }
+
+        return kept == 0;
+}
+
+static void cleanup_dirs_after_db_cleanup(DIR *dir, DIR *datadir) {
+
+        assert(dir && datadir);
+
+        FOREACH_DIRENT_ALL(dent, dir, break) {
+                struct stat stats;
+
+                if (dot_or_dot_dot(dent->d_name))
+                        continue;
+                if (fstatat(dirfd(dir), dent->d_name, &stats, AT_SYMLINK_NOFOLLOW) < 0)
+                        continue;
+                if (S_ISDIR(stats.st_mode)) {
+                        _cleanup_closedir_ DIR *dir2 = NULL;
+
+                        dir2 = fdopendir(openat(dirfd(dir), dent->d_name, O_RDONLY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC));
+                        if (dir2 && cleanup_dir_after_db_cleanup(dir2, datadir))
+                                (void) unlinkat(dirfd(dir), dent->d_name, AT_REMOVEDIR);
+                } else
+                        (void) unlinkat(dirfd(dir), dent->d_name, 0);
+        }
+}
+
 static void cleanup_db(void) {
         _cleanup_closedir_ DIR *dir1 = NULL, *dir2 = NULL, *dir3 = NULL, *dir4 = NULL, *dir5 = NULL;
 
@@ -265,11 +321,11 @@ static void cleanup_db(void) {
 
         dir2 = opendir("/run/udev/links");
         if (dir2)
-                cleanup_dir(dir2, 0, 2);
+                cleanup_dirs_after_db_cleanup(dir2, dir1);
 
         dir3 = opendir("/run/udev/tags");
         if (dir3)
-                cleanup_dir(dir3, 0, 2);
+                cleanup_dirs_after_db_cleanup(dir3, dir1);
 
         dir4 = opendir("/run/udev/static_node-tags");
         if (dir4)
@@ -277,7 +333,7 @@ static void cleanup_db(void) {
 
         dir5 = opendir("/run/udev/watch");
         if (dir5)
-                cleanup_dir(dir5, 0, 1);
+                cleanup_dir_after_db_cleanup(dir5, dir1);
 }
 
 static int query_device(QueryType query, sd_device* device) {


src/udev/udevadm-info.c
```diff
@@ -254,6 +254,53 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+/*
+ * Assume that dir is a directory with file names matching udev data base
+ * entries for devices in /run/udev/data (such as "b8:16"), and removes
+ * all files except those that haven't been deleted in /run/udev/data
+ * (i.e. they were skipped during db cleanup because of the db_persist flag).
+ * Returns true if the directory is empty after cleanup.
+ */
+static bool cleanup_dir_after_db_cleanup(DIR *dir, DIR *datadir) {
+        unsigned int kept = 0;
+
```
bluca, line 268: assert on dir and datadir

src/udev/udevadm-info.c
```diff
@@ -254,6 +254,53 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+                                        S_ISDIR(link_stats.st_mode) ? AT_REMOVEDIR : 0);
+                else
+                        kept++;
+        }
+
+        return kept == 0;
+}
+
+static void cleanup_dirs_after_db_cleanup(DIR *dir, DIR *datadir) {
+        FOREACH_DIRENT_ALL(dent, dir, break) {
```
bluca, line 297: assert on dir

src/udev/udevadm-info.c
```diff
@@ -254,6 +254,53 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+                if (dot_or_dot_dot(dent->d_name))
+                        continue;
+                if (fstatat(dirfd(dir), dent->d_name, &link_stats, AT_SYMLINK_NOFOLLOW) != 0)
+                        continue;
+
+                if (fstatat(dirfd(datadir), dent->d_name, &data_stats, 0) != 0)
+                        (void) unlinkat(dirfd(dir), dent->d_name,
+                                        S_ISDIR(link_stats.st_mode) ? AT_REMOVEDIR : 0);
+                else
+                        kept++;
```
yuwata, line 287: I think the counter should be also increased in the above two `continue` cases.

bluca:
```
../src/basic/fd-util.h: In function â€˜cleanup_dirs_after_db_cleanupâ€™:
../src/basic/fd-util.h:49:40: error: â€˜dir2â€™ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   49 | DEFINE_TRIVIAL_CLEANUP_FUNC_FULL(DIR*, closedir, NULL);
      |                                        ^~~~~~~~
../src/udev/udevadm-info.c:294:49: note: â€˜dir2â€™ was declared here
  294 |                         _cleanup_closedir_ DIR *dir2;
```

src/udev/udevadm-info.c
```diff
@@ -254,6 +254,53 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+static void cleanup_dirs_after_db_cleanup(DIR *dir, DIR *datadir) {
+        FOREACH_DIRENT_ALL(dent, dir, break) {
+                struct stat stats;
+
+                if (dot_or_dot_dot(dent->d_name))
+                        continue;
+                if (fstatat(dirfd(dir), dent->d_name, &stats, AT_SYMLINK_NOFOLLOW) != 0)
+                        continue;
+                if (S_ISDIR(stats.st_mode)) {
+                        _cleanup_closedir_ DIR *dir2;
```
yuwata: missing ` = NULL;`

src/udev/udevadm-info.c
```diff
@@ -254,6 +254,53 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+ */
+static bool cleanup_dir_after_db_cleanup(DIR *dir, DIR *datadir) {
+        unsigned int kept = 0;
+
+        FOREACH_DIRENT_ALL(dent, dir, break) {
+                struct stat data_stats, link_stats;
+
+                if (dot_or_dot_dot(dent->d_name))
+                        continue;
+                if (fstatat(dirfd(dir), dent->d_name, &link_stats, AT_SYMLINK_NOFOLLOW) != 0)
```
yuwata: Not a big matter, but we usually use `< 0` for such case instead of `!= 0`. Of course, the man page says it returns 0 on success. But just for readability. I.e., negative means failure, and non-negative means success.

Also, I prefer a brief comment why we skip the file here. Of course, it is explained in the above, but adding short comment is better.

src/udev/udevadm-info.c
```diff
@@ -263,19 +310,19 @@ static void cleanup_db(void) {
...
+                cleanup_dirs_after_db_cleanup(dir3, dir1);
 
         dir4 = opendir("/run/udev/static_node-tags");
         if (dir4)
                 cleanup_dir(dir4, 0, 2);
 
         dir5 = opendir("/run/udev/watch");
         if (dir5)
-                cleanup_dir(dir5, 0, 1);
+                cleanup_dir_after_db_cleanup(dir5, dir1);
```
yuwata, line 336: This is not your fault, but recent udev also stores symlinks from watch handle to device id in 'watch' directory.
If you'd like, could you support such files?

If not, that's ok in this PR. I will do that later.

src/udev/udevadm-info.c
mwilck: > assert on dir and datadir

well, it's a static function and the caller, which checks NULL, is just a few lines below... but if you deem it important, fine.

src/udev/udevadm-info.c
mwilck: The assignment is right in following line. I could have written

```
                        _cleanup_closedir_ DIR *dir2 = fdopendir(...)
```
but that way the line became over-long

src/udev/udevadm-info.c
yuwata: It may not necessary, but it improves readability of the code.

bluca: Yes - coding style, it's in-line documentation for callers. You know this know, but in 5 years after three refactors when it's called from six different places it won't be so obvious anymore

src/udev/udevadm-info.c
mwilck: That would be wrong - the handles will be invalid after restarting udev. `udev_watch_restore()` removes all "handle" files:
```
                if (safe_atoi(ent->d_name, &wd) < 0)
                        goto unlink;
```

yuwata:
Anyway, the basic concept of this PR looks good. Thank you!

src/udev/udevadm-info.c
mwilck: certainly not for `.` or `..`, or am I missing something? In the case when `fstatat()` errors out, I suppose `keep` should only be incremented for certain error codes. Not for `ENOENT`, for example.

src/udev/udevadm-info.c
mwilck: resolved with b148f0cc26

src/udev/udevadm-info.c
mwilck: resolved with b148f0cc26

mwilck (commented):
Replies to review comments. All should be resolved except the one about the watch handles.

mwilck:
> ```
> ../src/basic/fd-util.h: In function â€˜cleanup_dirs_after_db_cleanupâ€™:
> ../src/basic/fd-util.h:49:40: error: â€˜dir2â€™ may be used uninitialized in this function 
> ```

I didn't see this error message. It was a false positive. Anyway, it's resolved with b148f0c  I think.

src/udev/udevadm-info.c
```diff
@@ -255,6 +255,60 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+ * Assume that dir is a directory with file names matching udev data base
+ * entries for devices in /run/udev/data (such as "b8:16"), and removes
+ * all files except those that haven't been deleted in /run/udev/data
+ * (i.e. they were skipped during db cleanup because of the db_persist flag).
+ * Returns true if the directory is empty after cleanup.
+ */
+static bool cleanup_dir_after_db_cleanup(DIR *dir, DIR *datadir) {
+        unsigned int kept = 0;
+
+        assert(dir != NULL && datadir != NULL);
```
yuwata: Not a big matter, but we accept implicit cast from pointer to boolean, hence `!= NULL` is not necessary.
Also, that's super minor comment, but we usually add one empty line between assertions and other parts:
```
static bool cleanup_dir_after_db_cleanup(DIR *dir, DIR *datadir) {
        unsigned int kept = 0;

        assert(dir);
        assert(datadir);

        FOREACH_DIRENT_ALL(...) {
                ...
        }
```

src/udev/udevadm-info.c
```diff
@@ -255,6 +255,60 @@ static void cleanup_dir(DIR *dir, mode_t mask, int depth) {
...
+                        /* The entry still exists under /run/udev/data */
+                        kept++;
+        }
+
+        return kept == 0;
+}
+
+static void cleanup_dirs_after_db_cleanup(DIR *dir, DIR *datadir) {
+
+        assert(dir != NULL && datadir != NULL);
```
yuwata: Also here:
```
static void cleanup_dirs_after_db_cleanup(DIR *dir, DIR *datadir) {
        assert(dir);
        assert(datadir);

        FOREACH_DIRENT_ALL(...)
```

yuwata:
> > ```
> > ../src/basic/fd-util.h: In function â€˜cleanup_dirs_after_db_cleanupâ€™:
> > ../src/basic/fd-util.h:49:40: error: â€˜dir2â€™ may be used uninitialized in this function 
> > ```
> 
> I didn't see this error message. It was a false positive. Anyway, it's resolved with [b148f0c](https://github.com/systemd/systemd/commit/b148f0cc263a6604461304a7c7ed0f4b1e7a667e) I think.

I guess the error message is from clang? Anyway, adding `= NULL` should solve the warning.

mwilck:
Addressed additional style issues

src/udev/udevadm-info.c
yuwata: It may be. I will take a look. Let's leave the code as is at least in this PR.

yuwata (approved):
LGTM. Thank you!

bluca (approved):

PR #17957, opened 2020-12-14
core/namespace: ignore ENOENT for /proc/sys/kernel/domainname and hostname

yuwata:
If they do not exist, hostname or domainname cannot be modified. So, it is ok.

Fixes #17866, especially https://github.com/systemd/systemd/issues/17866#issuecomment-744118614.

PR #22203, opened 2022-01-21
core/namespace: allow using ProtectSubset=pid and ProtectHostname=tru…

brauner:
…e together

If a service requests both ProtectSubset=pid and ProtectHostname=true
then it will currently fail to start. The ProcSubset=pid option
instructs systemd to mount procfs for the service with subset=pid which
hides all entries other than /proc/<pid>. Consequently trying to
interact with the two files /proc/sys/kernel/{hostname,domainname}
covered by ProtectHostname=true will fail.

Fix this by only performing this check when ProtectSubset=pid is not
requested. Essentially ProtectSubset=pid implies/provides
ProtectHostname=true.

### DIFF
--- a/src/core/namespace.c
+++ b/src/core/namespace.c
@@ -1578,7 +1578,14 @@ static size_t namespace_calculate_mounts(
                 ns_info->private_ipc; /* /dev/mqueue */
 }
 
-static void normalize_mounts(const char *root_directory, MountEntry *mounts, size_t *n_mounts) {
+/* Walk all mount entries and dropping any unused mounts. This affects all
+ * mounts:
+ * - that are implicitly protected by a path that has been rendered inaccessible
+ * - whose immediate parent requests the same protection mode as the mount itself
+ * - that are outside of the relevant root directory
+ * - which are duplicates
+ */
+static void drop_unused_mounts(const char *root_directory, MountEntry *mounts, size_t *n_mounts) {
         assert(root_directory);
         assert(n_mounts);
         assert(mounts || *n_mounts == 0);
@@ -1684,7 +1691,7 @@ static int apply_mounts(
                 if (!again)
                         break;
 
-                normalize_mounts(root, mounts, n_mounts);
+                drop_unused_mounts(root, mounts, n_mounts);
         }
 
         /* Now that all filesystems have been set up, but before the
@@ -2250,7 +2257,7 @@ int setup_namespace(
                 if (r < 0)
                         goto finish;
 
-                normalize_mounts(root, mounts, &n_mounts);
+                drop_unused_mounts(root, mounts, &n_mounts);
         }
 
         /* All above is just preparation, figuring out what to do. Let's now actually start doing something. */


brauner:
See bits of the discussion in https://github.com/systemd/systemd/pull/17957.

src/core/namespace.c
```diff
@@ -2157,7 +2157,10 @@ int setup_namespace(
                                 goto finish;
                 }
 
-                if (ns_info->protect_hostname) {
+                /* If proc is mounted with subset=pid then neither of the two
+                 * paths will exist, i.e. they are implicitly protected by the
+                 * mount option. */
+                if (ns_info->protect_hostname && !ns_info->proc_subset == PROC_SUBSET_PID) {
```
brauner: Oh, one thing I should check is how this is handled if the kernel doesn't support `subset=pid`, i.e. do we at that point already now that `PROC_SUBSET_PID` won't work and will have unset `protect_subset`. If not we need to tweak the patch.

bluca: Maybe use `.ignore = ns_info->proc_subset == PROC_SUBSET_PID` instead? Using proc_subnet_pid can fail (eg, old kernel), in which case this should still apply I think

bluca: see other comment, we gracefully fall back in that case

brauner: Ha, we had the same worry. :)
Changed it to set `.ignore` in case `PROC_SUBSET_PID` is requested.

brauner: fixed

yuwata (approved):
LGTM. If possible, adding a testcase for the original issue is better.

yuwata:
(BTW, I do not have any good idea about that, neither I think we should support such case, but the ENOENT error for hostname or domainname are also ignored on older kernels which do not support ProtectSubset=.)

bluca (approved):

brauner:
> (BTW, I do not have any good idea about that, neither I think we should support such case, but the ENOENT error for hostname or domainname are also ignored on older kernels which do not support ProtectSubset=.)

I'm not sure what you mean by "neither I think we should support such case". Are you saying we shouldn't support `ProcSubset=pid` and `ProtectHostname=true` used together? I find that to be an unlikely reading of what you're trying to say. :)

It seems you worry about the case where your `subset=pid` isn't supported and you still get `ENOENT` for `/proc/sys/kernel/{domainname,hostname}`? That seems impossible and so not worth worrying imho.

brauner:
Also, are there released systemd versions that ignore `ENOENT` from `{domainname,hostname}` that would hide this issue before the revert?

bluca:
> Also, are there released systemd versions that ignore `ENOENT` from `{domainname,hostname}` that would hide this issue before the revert?

No, I don't think so

yuwata:
> > (BTW, I do not have any good idea about that, neither I think we should support such case, but the ENOENT error for hostname or domainname are also ignored on older kernels which do not support ProtectSubset=.)
> 
> I'm not sure what you mean by "neither I think we should support such case". Are you saying we shouldn't support `ProcSubset=pid` and `ProtectHostname=true` used together? I find that to be an unlikely reading of what you're trying to say. :)
> 
> It seems you worry about the case where your `subset=pid` isn't supported and you still get `ENOENT` for `/proc/sys/kernel/{domainname,hostname}`? That seems impossible and so not worth worrying imho.

Sorry for my unclear comment.

I mean, previously, error in mounting domainname or hostname are always handled as critical. But with this change, if `ns_info->proc_subset == PROC_SUBSET_PID`, then an occasional and accidental error in mounting one of them with ENOENT is ignored even for older kernels witch do not support the new procfs mount option.
I do not have any good way to make such error critical for older kernels. But if there exists a simple way to handle them for older kernels, then that's better.

Is it clearer now?

Anyway, I think it is not necessary to be handled in this PR. It's a kind of off-topic here.

brauner:
> > > (BTW, I do not have any good idea about that, neither I think we should support such case, but the ENOENT error for hostname or domainname are also ignored on older kernels which do not support ProtectSubset=.)
> > 
> > 
> > I'm not sure what you mean by "neither I think we should support such case". Are you saying we shouldn't support `ProcSubset=pid` and `ProtectHostname=true` used together? I find that to be an unlikely reading of what you're trying to say. :)
> > It seems you worry about the case where your `subset=pid` isn't supported and you still get `ENOENT` for `/proc/sys/kernel/{domainname,hostname}`? That seems impossible and so not worth worrying imho.
> 
> Sorry for my unclear comment.
> 
> I mean, previously, error in mounting domainname or hostname are always handled as critical. But with this change, if `ns_info->proc_subset == PROC_SUBSET_PID`, then an occasional and accidental error in mounting one of them with ENOENT is ignored even for older kernels witch do not support the new procfs mount option. I do not have any good way to make such error critical for older kernels. But if there exists a simple way to handle them for older kernels, then that's better.
> 
> Is it clearer now?

Yes, thank you!

It shouldn't be super hard to create a dependency between the procfs mount and these two mounts. Once you have that you can e.g. add a callback function `.ingnorable()` that you can call. Of course, you can only create dependencies on mounts that happened earlier.

brauner:
> > > > (BTW, I do not have any good idea about that, neither I think we should support such case, but the ENOENT error for hostname or domainname are also ignored on older kernels which do not support ProtectSubset=.)
> > > 
> > > 
> > > I'm not sure what you mean by "neither I think we should support such case". Are you saying we shouldn't support `ProcSubset=pid` and `ProtectHostname=true` used together? I find that to be an unlikely reading of what you're trying to say. :)
> > > It seems you worry about the case where your `subset=pid` isn't supported and you still get `ENOENT` for `/proc/sys/kernel/{domainname,hostname}`? That seems impossible and so not worth worrying imho.
> > 
> > 
> > Sorry for my unclear comment.
> > I mean, previously, error in mounting domainname or hostname are always handled as critical. But with this change, if `ns_info->proc_subset == PROC_SUBSET_PID`, then an occasional and accidental error in mounting one of them with ENOENT is ignored even for older kernels witch do not support the new procfs mount option. I do not have any good way to make such error critical for older kernels. But if there exists a simple way to handle them for older kernels, then that's better.
> > Is it clearer now?
> 
> Yes, thank you!
> 
> It shouldn't be super hard to create a dependency between the procfs mount and these two mounts. Once you have that you can e.g. add a callback function `.ingnorable()` that you can call. Of course, you can only create dependencies on mounts that happened earlier.

There are other methods of course. :)

yuwata:
@brauner Thank you for the suggestion. It maybe better to handled with ENOENT handling issue in ExtensionImage/ExtensionDIrectory=.

(Feel free to upgrade the green label if you are planning to add a testcase in later PR.)

topimiettinen (changes requested):
Could you reuse `ignore_protect_proc` above? It would also consider `ns_info->ignore_protect_paths`.

PR #22204, opened 2022-01-21
meson: Add missing test dependencies

daandemeyer:
Currently, running "meson build" followed by "meson test -C build"
will result in many failed tests due to missing dependencies. This
commit adds the missing dependencies to make sure no tests fail.


DaanDeMeyer:
cc @medhefgo who ran into this problem as well I think

medhefgo (approved):
Nice. I just noticed that `ninja -C build test` does the right thing, which would explain to me why this didn't always happen.

PR #22208, opened 2022-01-21
boot: Only build with debug symbols in developer mode

medhefgo:
The debug symbols are of very limited use in proper deployments
unlike with regular userspace. Unless someone goes through the pain
of setting up an EFI debugger (assuming their firmware even supports
this in the first place) any provided debug symbols will just be
useless.
Debugging under QEMU is possible, but even then it is non-trivial
to set up, so anyone willing to go that far can just build in
developer mode.

Meanwhile, at least x86 firmware tends to refuse binaries that contain
debug symbols. We do strip the files when converted to PE anyway, but
the elf file needs to stay around on other arches as objcopy does not
support PE as input there.

Also, the generated debug symbols seem to be not reproducible when
building with LTO. Whether this is an issue in tooling or our side
is unclear. This works around this issue.

Fixes: #22157

### DIFF
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -195,7 +195,7 @@ efi_cflags += cc.get_supported_arguments({
 if get_option('werror')
         efi_cflags += ['-Werror']
 endif
-if get_option('debug')
+if get_option('debug') and get_option('mode') == 'developer'
         efi_cflags += ['-ggdb', '-DEFI_DEBUG']
 endif
 if get_option('optimization') != '0'
@@ -213,13 +213,15 @@ foreach arg : get_option('c_args')
         if arg in [
                 '-DNDEBUG',
                 '-fno-lto',
-                '-g', '-ggdb',
                 '-O1', '-O2', '-O3', '-Og', '-Os',
                 '-Werror',
            ] or arg.split('=')[0] in [
                 '-ffile-prefix-map',
                 '-flto',
-           ]
+           ] or (get_option('mode') == 'developer' and arg in [
+                '-DEFI_DEBUG',
+                '-g', '-ggdb',
+           ])
 
                 message('Using "@0@" from c_args for EFI compiler'.format(arg))
                 efi_cflags += arg


evverx:
FWIW even without this patch reprotest says that `linuxx64.elf.stub` is reproducible
```
=======================
Reproduction successful
=======================
No differences in ./build/src/boot/efi/linuxx64.elf.stub
d825d1fc3e7a36f877d982b3b66ed83b2b3cb19804811ce5035070746bc250f4  ./build/src/boot/efi/linuxx64.elf.stub
```
assuming something like `meson --buildtype=release` is used.

medhefgo:
> FWIW even without this patch reprotest says that `linuxx64.elf.stub` is reproducible

Except when it isn't: https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/systemd.html

evverx:
Given that it seems to be debian unstable it isn't clear whether it has anything to do with systemd. I haven't been able to reproduce that manually.

evverx:
With debugging symbols systemd or (any other binary) isn't reproducible either. I'm not sure why only the boot part triggered that.

bluca:
Thank you! I've kicked off another build, but this should fix the issue https://salsa.debian.org/bluca/systemd/-/pipelines/339826

evverx:
Judging by https://salsa.debian.org/bluca/systemd/-/pipelines/339826, whatever it was it is gone apparently :-)

It would be interesting to see what reprotest says there with `-Dbpf-framework=true` (right now it seems to be disabled there)

bluca:
> Thank you! I've kicked off another build, but this should fix the issue https://salsa.debian.org/bluca/systemd/-/pipelines/339826

As expected it works, so let's go ahead - thanks!

bluca (approved):

evverx:
@bluca I'm wondering why those files are installed differently? Looking at 
```
debian/systemd.lintian-overrides:systemd: executable-not-elf-or-script usr/lib/systemd/boot/efi/linuxx64.efi.stub
debian/systemd.lintian-overrides:systemd: executable-not-elf-or-script usr/lib/systemd/boot/efi/systemd-bootx64.efi
```
it seems they're treated differently by the debian packaging tools.

I'm asking because I thought it would be useful to run `reprotest` once a day or so here but it appears the toolchain on Debian is different somehow and it wouldn't help to avoid issues like that.

bluca:
lintian is often based on heuristics, and it can get it wrong

evverx:
It can but given that reprotest didn't complain about anything else it appears only those two files weren't stripped

Issue #7390, opened 2017-11-19
hostnamed needs a quirk database for chassis type


afranke:
### Submission type

 - Bug report

### systemd version the issue has been seen with

systemd 233

### Used distribution

Fedora 27

### In case of bug report: Expected behaviour you didn't see

This machine should be detected as a tablet.

### In case of bug report: Unexpected behaviour you saw

This machine is detected as a server.

### In case of bug report: Steps to reproduce the problem

Type `hostnamectl` on a Microsoft Surface Pro (first generation).

afranke:
CCing @hadess as requested.

hadess:
Can you please attach the output of `dmidecode` run as root, as well as get the decompiled `DSDT.dsl` file:
```
# cat /sys/firmware/acpi/tables/DSDT > DSDT.bin
$ iasl -d DSDT.bin
```

afranke:
Here you go.

[dmidecode.txt](https://github.com/systemd/systemd/files/1485937/dmidecode.txt)
[DSDT.zip](https://github.com/systemd/systemd/files/1485938/DSDT.zip)

Had to zip the file or Github wouldn’t accept it.

hadess:
Shows up as a server in dmidecode, which is our preferred method of checking chassis type:

```
Handle 0x0003, DMI type 3, 22 bytes
Chassis Information
	Manufacturer: Microsoft Corporation
	Type: Main Server Chassis
	Lock: Not Present
	Version: 1
	Serial Number: 0
	Asset Tag: 0
	Boot-up State: Safe
	Power Supply State: Safe
	Thermal State: Safe
	Security Status: None
	OEM Information: 0x00000000
	Height: Unspecified
	Number Of Power Cords: 1
	Contained Elements: 0
	SKU Number: Surface_PRO_1
```

@poettering do we have a quirk file for chassis types?

poettering:
@hadess not yet, but we should totally have one. This would not only mean adding a new hwdb file for this, but also making hostnamed actually check it. Patches very welcome, of course

zdzichu:
@afranke, why not report the issue to the manufacturer to have it fixed?

hadess:
> @afranke, why not report the issue to the manufacturer to have it fixed?

This half sounds like a troll.

1) it's Microsoft and they probably don't use this data, or have it correct in another bit of data such as ACPI or even somewhere in the registry
[...]

PR #22209, opened 2022-01-21
hostname: Allow overriding the chassis type from hwdb

hadess:
Untested (and possibly too late)

If someone has a copy of `udevadm info /sys/class/dmi/id` for their system to add, or from the original Surface Pro...

Closes: #7390

### DIFF
--- a/hwdb.d/20-dmi-id.hwdb
+++ b/hwdb.d/20-dmi-id.hwdb
@@ -4,3 +4,7 @@
 dmi:bvnLENOVO*
     ID_SYSFS_ATTRIBUTE_MODEL=product_version
     ID_VENDOR_FROM_DATABASE=Lenovo
+
+# Microsoft Surface 1's chassis type
+dmi:bvnMicrosoft Corporation*:pvrSurface with Windows 8 Pro*
+    ID_CHASSIS=tablet


src/hostname/hostnamed.c
```diff
@@ -725,16 +725,19 @@ static int property_get_chassis(
...
 
-        if (isempty(c->data[PROP_CHASSIS]))
-                name = fallback_chassis();
-        else
+        if (isempty(c->data[PROP_CHASSIS])) {
+                if (get_dmi_data ("ID_CHASSIS", NULL, &dmi_chassis) <= 0)
+                        name = fallback_chassis();
+        } else {
                 name = c->data[PROP_CHASSIS];
+        }
```
yuwata: `{}` can be dropped.

src/hostname/hostnamed.c
```diff
@@ -725,16 +725,19 @@ static int property_get_chassis(
...
 
-        return sd_bus_message_append(reply, "s", name);
+        return sd_bus_message_append(reply, "s", name ? name : dmi_chassis);
```
medhefgo:
```suggestion
        return sd_bus_message_append(reply, "s", name ?: dmi_chassis);
```

src/hostname/hostnamed.c
```diff
@@ -725,16 +725,19 @@ static int property_get_chassis(
...
+        _cleanup_free_ char *dmi_chassis = NULL;
+        const char *name = NULL;
 
         context_read_machine_info(c);
 
-        if (isempty(c->data[PROP_CHASSIS]))
-                name = fallback_chassis();
-        else
+        if (isempty(c->data[PROP_CHASSIS])) {
+                if (get_dmi_data ("ID_CHASSIS", NULL, &dmi_chassis) <= 0)
```
medhefgo: remove extra whitespace

PR #22210, opened 2022-01-21
Fix journald audit logging with fields > N_IOVEC_AUDIT_FIELDS.

YmrDtnJu:
Changes in 250 introduced a limit on the number of fields created from an audit message.

map_all_fields got a new parameter m to specify the highest index to use for fields from the audit message.

In function process_audit_string an iovec is allocated with N_IOVEC_META_FIELDS for later use, plus 8 elements for fields added in the function itself, plus a maximum of N_IOVEC_AUDIT_FIELDS to be added from the audit message by map_all_fields.

map_all_fields is called with ELEMENTSOF(iovec) which allows it to use all remaining elements in the array including those reserved with N_IOVEC_META_FIELDS. This cases the assertion in journald-server.c line 923 to be triggered and journald to dump core.

The solution is to call map_all_fields with n + N_IOVEC_AUDIT_FIELDS which is the correct highest index to use for audit fields. n is the number of already added fields before the call to map_all_fields (8 for now) and N_IOVEC_AUDIT_FIELDS is the maximum number of fields to add. The recursive call to map_all_fields is correct, because the parameter m is passed unchanged.

### DIFF
--- a/src/journal/journald-audit.c
+++ b/src/journal/journald-audit.c
@@ -399,7 +399,7 @@ void process_audit_string(Server *s, int type, const char *data, size_t size) {
 
         z = n;
 
-        map_all_fields(p, map_fields_kernel, "_AUDIT_FIELD_", true, iovec, &n, ELEMENTSOF(iovec));
+        map_all_fields(p, map_fields_kernel, "_AUDIT_FIELD_", true, iovec, &n, n + N_IOVEC_AUDIT_FIELDS);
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
LGTM, thanks for taking the time to fix the issue.

dcousens:
Thanks for this @YmrDtnJu,  this was dumping my `journald` service repeatedly until it stopped restarting

PR #22216, opened 2022-01-22
sd-event: workaround maybe-uninitalized warning in sd_event_add_inotify()

thom311:
With LTO, the compiler might think that the variable is uninitialized
(from NetworkManager's fork, with gcc-11.2.1-1.fc35):
```
    src/libnm-systemd-core/src/libsystemd/sd-event/sd-event.c: In function 'sd_event_add_inotify':
    src/libnm-systemd-core/src/libsystemd/sd-event/sd-event.c:2120: error: 's' may be used uninitialized in this function [-Werror=maybe-uninitialized]
     2120 |                 *ret = s;
          |
    src/libnm-systemd-core/src/libsystemd/sd-event/sd-event.c:2102: note: 's' was declared here
     2102 |         sd_event_source *s;
          |
    lto1: all warnings being treated as errors
```
In particular, that would happen for codepaths where `event_add_inotify_fd_internal()`
returns `-errno`, and the compiler cannot be sure that the returned value will
be negative. Technically, the compiler is right, but we rely on libc functions
to set errno correctly, so this only happens in code paths, where something
bad already happend.

While LTO is prone to such false warnings, we are largely able to build systemd
without warnings. So it is feasible and we should make the effort of working
around warnings as they appear.

---

I guess, and alternative would be to use [a macro like c-util has](https://github.com/c-util/c-stdaux/blob/main/src/c-stdaux.h#L450). But I think systemd doesn't have that (has it?), so that solution ws not taken. Also, it would be more complicated.

### DIFF
--- a/src/libsystemd/sd-event/sd-event.c
+++ b/src/libsystemd/sd-event/sd-event.c
@@ -2095,7 +2095,7 @@ _public_ int sd_event_add_inotify(
                 sd_event_inotify_handler_t callback,
                 void *userdata) {
 
-        sd_event_source *s;
+        sd_event_source *s = NULL; /* avoid false maybe-uninitialized warning */
         int fd, r;
 
         assert_return(path, -EINVAL);


src/libsystemd/sd-event/sd-event.c
```diff
@@ -2095,7 +2095,7 @@ _public_ int sd_event_add_inotify(
                 sd_event_inotify_handler_t callback,
                 void *userdata) {
 
-        sd_event_source *s;
+        sd_event_source *s = NULL;
```
yuwata: We usually add `/* avoid false maybe-uninitialized warning */` comment in such cases other places. Please also add it here.

yuwata (approved):
LGTM. One suggestion.

src/libsystemd/sd-event/sd-event.c
thom311: done.

PR #22219, opened 2022-01-22
hostname: also expose hardware serial number

### LINKED ISSUE #22119: [hostnamed] Get "Serial Number" of device using hostnamed
Hello , 
Currently , I am working to display "Serial Number" of device in gnome-settings . I am using ThinkPad X1 Yoga G6
We can get "Serial Number" of the device using sysfs "/sys/class/dmi/id/product_serial" . But since this needs root permission , we cannot directly access it .
Is it acceptable to add "serial number" in hostnamed ? Any comment regarding this will be helpful. 
Thank you !1

### LINKED ISSUE #22207: Allow override of HardwareModel value reported to org.freedesktop.hostname1
It is ridiculously common for motherboard vendors to leave DMI information blank on hardware sold to consumers (rather than to Dell and HP), resulting in motherboards reporting their model as "To be filled in by O.E.M." or "System Product Name". GNOME Control Center dutifully reports the value taken from DMI:

FJo_fMLXwAUWq4H

Which mirrors the behaviour where Windows also reports the same value, so thoughtfully filled in by the motherboard manufacturer:

FA8PXI5WQAogvOR

Windows, however, has an override function. Setting the override will cause anything which asks Windows for the hardware model (as opposed to anything which probes DMI directly) to report a user-specified or vendor-specified value instead, as filled in via the Windows Registry key `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\OEMInformation\Model`

There should be some means to override the value hostnamed reports via dbus. The exact implementation of that override doesn't really matter, just that it should exist.

Per-app overrides could exist instead, but fixing it at the lower level seems like a better use of time. Of course, hostnamectl would need a per-app override too in that case, and we'd still be back here in GitHub with a feature request issue.

I looked at 249.5 and didn't see any relevant changes in git master

yuwata:
Closes #22119.

### DIFF
--- a/src/hostname/hostnamed.c
+++ b/src/hostname/hostnamed.c
@@ -205,6 +205,33 @@ static int get_hardware_model(char **ret) {
         return get_dmi_data("ID_MODEL_FROM_DATABASE", "ID_MODEL", ret);
 }
 
+static int get_hardware_serial(char **ret) {
+        _cleanup_(sd_device_unrefp) sd_device *device = NULL;
+        _cleanup_free_ char *b = NULL;
+        const char *s = NULL;
+        int r;
+
+        r = sd_device_new_from_syspath(&device, "/sys/class/dmi/id");
+        if (r < 0)
+                return log_debug_errno(r, "Failed to open /sys/class/dmi/id device, ignoring: %m");
+
+        (void) sd_device_get_sysattr_value(device, "product_serial", &s);
+        if (isempty(s))
+                /* Fallback to board serial */
+                (void) sd_device_get_sysattr_value(device, "board_serial", &s);
+
+        if (!isempty(s)) {
+                b = strdup(s);
+                if (!b)
+                        return -ENOMEM;
+        }
+
+        if (ret)
+                *ret = TAKE_PTR(b);
+
+        return !isempty(s);
+}
+
 static const char* valid_chassis(const char *chassis) {
         assert(chassis);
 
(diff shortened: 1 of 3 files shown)


bluca:
Is showing the serial number something that should require privilege (ie, polkit)? Or is it something that is expected to be available to anything running on a system, like the machine-id?

src/hostname/hostnamed.c
```diff
@@ -473,52 +479,6 @@ static int context_write_data_static_hostname(Context *c) {
         return 0;
 }
 
-static int context_write_data_machine_info(Context *c) {
```
bluca: Why was this removed? It breaks the build as it's used further down

yuwata: Ouch. I have no idea why I did. Maybe mistake when rebase...

yuwata:
> Is showing the serial number something that should require privilege (ie, polkit)? Or is it something that is expected to be available to anything running on a system, like the machine-id?

I do not think it is not necessary. We already have exposed hardware vendor and model without authentication. I do not think serial number contains any security related info.

bluca:
> > Is showing the serial number something that should require privilege (ie, polkit)? Or is it something that is expected to be available to anything running on a system, like the machine-id?
> 
> I do not think it is not necessary. We already have exposed hardware vendor and model without authentication. I do not think serial number contains any security related info.

Those are visible without root permissions though:

```
$ ls -l /sys/class/dmi/id/product_name /sys/class/dmi/id/sys_vendor /sys/class/dmi/id/product_serial
-r--r--r-- 1 root root 4096 Jan 23 12:26 /sys/class/dmi/id/product_name
-r-------- 1 root root 4096 Jan 23 12:26 /sys/class/dmi/id/product_serial
-r--r--r-- 1 root root 4096 Jan 23 12:26 /sys/class/dmi/id/sys_vendor
```

And it seems to be done intentionally:

https://github.com/torvalds/linux/blob/df0cc57e057f18e44dac8e6c18aba47ab53202f9/drivers/firmware/dmi-id.c#L49
https://github.com/torvalds/linux/commit/e0733e975385d25d88c1384dafb8ea4dcf7513c0

I imagine it's because the vendor and board name are generic, but the serial is unique and burnt-in, so it could be used to track a machine, and it can't be changed.
I think the request to be able to read it from the Gnome software is perfectly legitimate, because that's shown to the logged in user - so IMHO we should put this behind a polkit authorization. Then Gnome can ship with the required policy, but other unprivileged processes cannot read it.

yuwata:
OK. Let's wait for other reviewers for a while.

yuwata:
@nitinexclusively What do you think about the above security concern? Is it OK to require polkit authentication?

nitinexclusively:
> @nitinexclusively What do you think about the above security concern? Is it OK to require polkit authentication?

Hello @yuwata , Thank you for working on it . yes , its OK to require polkit authentication . 
JFYI: Target is to display "Serial Number" in Gnome settings (Refence : https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/1543)

yuwata:
@bluca and @nitinexclusively, thank you for the comments. Now the new dbus method GetHardwareSerial with polkit is added. PTAL.

bluca (approved):

PR #22222, opened 2022-01-22
dhcp-server: support packets from non-Ethernet interfaces

### LINKED ISSUE #22217: DHCP Server of systemd-networkd will not work on USB Gadget RNDIS connected to Windows
 249.5

**Used distribution**
 Buildroot

 `Linux smartcross 5.15.10 #8 SMP PREEMPT Sat Jan 22 20:15:20 CST 2022 aarch64 GNU/Linux`

 aarch64

 systemd-networkd respond to DHCP requests sent from Windows

 systemd-networkd does not respond to DHCP requests sent from Windows.

The reason is simple: in https://github.com/systemd/systemd/blob/f939a8984afb33e3242b8d18b40279876230dc61/src/libsystemd-network/sd-dhcp-server.c#L904 the DHCP server will reject anything but with `message->htype == ARPHRD_ETHER`(0x01). But for USB RNDIS interfaces Windows will send htype ARPHRD_ARCNET (0x07), causing networkd to drop this request.

截屏2022-01-22 下午10 51 15

Maybe we can add an exception for ARPHRD_ARCNET here?
 

 1. Grab a RPi Zero or RPi 4 with USB gadget functionality. You can use RPi OS or any OS you like, as this doesn't affect the reproducibility of this issue.
 2. Set up USB Gadget, like
 ```bash
CONFIGFS_ROOT=/sys/kernel/config
cd "${CONFIGFS_ROOT}"/usb_gadget
mkdir g1
cd g1
echo 0x5678 > idVendor
echo 0x1234 > idProduct
mkdir strings/0x409
echo "my vendor" > strings/0x409/manufacturer
echo "my device" > strings/0x409/product
mkdir configs/c.1
mkdir functions/rndis.usb0
ln -s functions/rndis.usb0 configs/c.1
ls /sys/class/udc > UDC
```
3. Add a systemd-networkd network file for the USB interface, like
```
[Match]
Name=usb0

[Network]
DHCPServer=true
DHCP=no
Address=10.222.11.1/24

[DHCPServer]
PoolOffset=128
PoolSize=32
EmitRouter=false
EmitNTP=false
EmitDNS=false
EmitTimezone=false
```
4. Bring up the interface.

systemd-networkd will not respond to DHCP request sent by Windows.

### PR DESCRIPTION
**Pull request by yuwata**
Fixes #22217.

### DIFF
--- a/src/libsystemd-network/sd-dhcp-server.c
+++ b/src/libsystemd-network/sd-dhcp-server.c
@@ -15,6 +15,7 @@
 #include "fd-util.h"
 #include "in-addr-util.h"
 #include "io-util.h"
+#include "memory-util.h"
 #include "network-common.h"
 #include "ordered-set.h"
 #include "siphash24.h"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by yuwata** [comment]
The issue reporter confirmed that this works fine.

File: src/libsystemd-network/sd-dhcp-server.c [hunk 1]
```diff
@@ -1536,18 +1536,13 @@ int sd_dhcp_server_set_static_lease(
...
                 DHCPClientId c;
 
-                data = memdup(client_id, client_id_size);
-                if (!data)
-                        return -ENOMEM;
-
                 c = (DHCPClientId) {
                         .length = client_id_size,
-                        .data = data,
+                        .data = client_id,
```
bluca, line 1585: Can you add a note in the commit message explaining why it's not necessary? It's not immediately obvious without following all the crumbs. Probably doesn't need a code comment, but at least in the commit message it would help to make it clear.

bluca (approved):

src/libsystemd-network/sd-dhcp-server.c
yuwata: Sure. Sorry, I am always lazy about commenting.

yuwata:
Updated the commit message, and all tests are previously green. Skipping the CIs now. Merging.

PR #22231, opened 2022-01-23
resolve: synthesize several filtered name

### LINKED ISSUE #22229: systemd-resolved returns SERVFAIL for reverse DNS lookups of 0.0.0.0/::0
Reverse DNS lookups for 0.0.0.0 and some other addresses listed in [RFC6303](https://datatracker.ietf.org/doc/html/rfc6303) fail with SERVFAIL:
```
$ host 0.0.0.0
Host 0.0.0.0.in-addr.arpa not found: 2(SERVFAIL)
$ resolvectl query 0.0.0.0
0.0.0.0: resolve call failed: No appropriate name servers or networks for name found
```

According to the RFC, NXDOMAIN should be returned for such queries. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=961378 is an example of a program failure that can result from this behavior.

Looking at the code, it looks like reverse lookups for some (but not all) of the addresses in the RFC are prevented from going to global resolvers here:
https://github.com/systemd/systemd/blob/bb995f747afa17055c26b5d69a21f68f7dab0c28/src/resolve/resolved-dns-scope.c#L624-L627

However, they do not appear to be handled in the place where DNS answers are synthesized:
https://github.com/systemd/systemd/blob/bb995f747afa17055c26b5d69a21f68f7dab0c28/src/resolve/resolved-dns-synthesize.c#L373-L470

Here is a [completely untested patch](https://github.com/systemd/systemd/files/7921340/untested_rfc6303.patch.txt) that contains some experiments at fixing this. However, I probably won't have the time to clean up/test it to make a nice pull request, sorry!

yuwata:
Fixes #22229.
Fixes RHBZ#2039854 (https://bugzilla.redhat.com/show_bug.cgi?id=2039854).
Replaces #22197.

### DIFF
--- a/src/resolve/resolved-dns-synthesize.c
+++ b/src/resolve/resolved-dns-synthesize.c
@@ -397,6 +397,14 @@ int dns_synthesize_answer(
                 if (dns_name_is_empty(name)) {
                         /* Do nothing. */
 
+                } else if (dns_name_endswith(name, "0.in-addr.arpa") > 0 ||
+                           dns_name_equal(name, "255.255.255.255.in-addr.arpa") > 0 ||
+                           dns_name_equal(name, "0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa") > 0 ||
+                           dns_name_endswith(name, "invalid") > 0) {
+
+                        nxdomain = true;
+                        continue;
+
                 } else if (is_localhost(name)) {
 
                         r = synthesize_localhost_rr(m, key, ifindex, &answer);


bluca (approved):

poettering:
This doesn't look right, as it likely will break DNSSEC logic. There's a root zone after all, and it carries DNSSEC RRs that we need for validating. With this change we'll never be able to resolve those anymore AFAICS.

poettering:
(I am only concerned about the "empty", i.e. top-level domain handling, the other stuff looks good. i.e. I think it might make sense to synthesize NXDOMAIN artificially for A+AAAA looksup in the root zone, but for all other RRs we should allow things to be handled as it wants)

PR #22235, opened 2022-01-24
user-runtime-dir: error out immediately if mkdir fails

keszybz:
We try to create two directories: /run/user and /run/user/<UID>. For the
first we check the return value and error out if creation fails. But for
the second one we continued based on the assumption that the subsequent
mount will immediately fail anyway. But this has the disadvantage that we
get a somewhat confusing error message:

janv. 23 22:04:31 nsfw systemd-user-runtime-dir[1660]: Failed to mount per-user tmpfs directory /run/user/1000: No such file or directory

Let's instead fail immediately with a precise error message.

For https://bugzilla.redhat.com/show_bug.cgi?id=2044100.


yuwata (approved):
LGTM.

bluca (approved):

PR #21904, opened 2021-12-26
docs: update "fuzzers" a bit

evverx:

PR #22236, opened 2022-01-24
fuzz-dhcp-server: three fixes or improvements

yuwata:


evverx:
> The fuzzer does not send anything to the address, so it should be not
necessary, but just for safety.

FWIW looking at https://bugzilla.redhat.com/show_bug.cgi?id=1639067 I think it would be great if those dhcp fuzz targets could cover those codepaths (where data is sent) as well.

src/libsystemd-network/fuzz-dhcp-server.c
```diff
@@ -21,12 +21,15 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
         struct in_addr address = {.s_addr = htobe32(UINT32_C(10) << 24 | UINT32_C(1))};
         static const uint8_t chaddr[] = {3,3,3,3,3,3,3,3,3,3,3,3,3,3,3,3};
+        _cleanup_free_ uint8_t *duped = NULL;
         uint8_t *client_id;
         DHCPLease *lease;
 
         if (size < sizeof(DHCPMessage))
                 return 0;
 
+        assert_se(duped = memdup(data, size));
```
evverx, line 31: What's weird is that the fuzz target has been running in its current form since 2018 and it seems to be the first time it has hit an issue like that. Looks like https://github.com/systemd/systemd/pull/22222 has changed something there. Even though the function signature doesn't say "const" anywhere I wonder if it's safe to assume that the dhcp server actually expects the data to be changed?

evverx:
> I think it would be great if those dhcp fuzz targets could cover those codepaths (where data is sent) as well.

Looking at https://github.com/systemd/systemd/pull/22222/ where the `dhcp_server_send*` functions were changed, https://github.com/systemd/systemd/pull/22227 where one heap-buffer-overflow was fixed and https://storage.googleapis.com/oss-fuzz-coverage/systemd/reports/20220119/linux/src/systemd/src/libsystemd-network/sd-dhcp-server.c.html#L894 where it's shown that the fuzzer doesn't actually cover much I think that part certainly needs more testing.

src/libsystemd-network/fuzz-dhcp-server.c
yuwata: > I wonder if it's safe to assume that the dhcp server actually expects the data to be changed?

Yes. The data is received by sd_dhcp_server, so it is ok to modify or update by itself. Also, the data is soon freed after dhcp_server_handle_message() is called.

yuwata:
@evverx Added one more commit to cover more code by the fuzzer. I hope it improves the coverage.

evverx:
The latest coverage report can be found at https://oss-fuzz.com/coverage-report/job/libfuzzer_asan_systemd/latest . I'm planning to add it to the documentation in https://github.com/systemd/systemd/pull/21904 :-)

yuwata:
Thanks.

yuwata:
So, I'd like this to be merged and see further reports. PTAL.

evverx:
FWIW @yuwata the link to the latest coverage report is kind of broken due to https://github.com/google/oss-fuzz/issues/7188 but according to https://storage.googleapis.com/oss-fuzz-coverage/systemd/reports/20220126/linux/src/systemd/src/libsystemd-network/report.html the coverage of `dhcp-packet.c` and `sd-dhcp-server.c` went up (from 0% to 53.77%  and from 18.8% to 51.75% accordingly) so the fuzzer seems to have gotten back to normal. Thanks!

PR #22239, opened 2022-01-24
homed: give users larger idmappings uperward of UID_NOBODY

brauner:
When using systemd-homed with idmapped mounts to manage home directories
systemd will allocate a uid for the user in the range of
HOME_UID_MIN/60001, HOME_UID_MAX/60513. Say systemd allocated 60370 as
the users uid. It will then create an idmapped mount with the following
idmapping for the user's home directory:

0 0 60001
60370 60370 1
60513 60513 5021

The problem is that this will prevent users from running containers as
nothing is mapped post UID_NOBODY. Give the users a 10m range starting
from UID_NOBODY + 1 this is enough to run variety of isolated containers.

Issues with the current approach were reported on the systemd-devel
mailing list: https://lists.freedesktop.org/archives/systemd-devel/2022-January/047285.html

### DIFF
--- a/src/home/home-util.h
+++ b/src/home/home-util.h
@@ -11,6 +11,7 @@
 /* See https://systemd.io/UIDS-GIDS for details how this range fits into the rest of the world */
 #define HOME_UID_MIN 60001
 #define HOME_UID_MAX 60513
+#define HOME_UID_RANGE_MAX 10000000
 
 /* Put some limits on disk sizes: not less than 5M, not more than 5T */
 #define USER_DISK_SIZE_MIN (UINT64_C(5)*1024*1024)
--- a/src/home/homework-mount.c
+++ b/src/home/homework-mount.c
@@ -209,6 +209,12 @@ static int make_userns(uid_t stored_uid, uid_t exposed_uid) {
         if (r < 0)
                 return log_oom();
 
+        /* Map HOME_UID_RANGE_MAX ids upwards of UID_NOBODY to let unprivileged
+         * users run containers and other shenanigans. */
+        r = strextendf(&text, UID_FMT " " UID_FMT " " UID_FMT "\n", UID_NOBODY + 1, UID_NOBODY + 1, HOME_UID_RANGE_MAX);
+        if (r < 0)
+                return log_oom();
+
         /* Leave everything else unmapped, starting from UID_NOBODY itself. Specifically, this means the
          * whole space outside of 16bit remains unmapped */
 


brauner:
Alternatives are:
- give them the full remaining range (i.e. every ranges substracted from `(uint32_t)-1`/`UINT_MAX`)
- introduce a command line switch to `homectl create` to allow the homed creator to specify the number of ids upwards of `UID_NOBODY + 1` to map
- introduce a generic command line switch to `homectl create` like `--i-want-to-run-containers` (sorry for the bad name) which allocates a fixed range (kinda like the first option but with opt-in behavior)

bluca:
Are there any downsides in adding these mappings?

brauner:
In terms of security, no.
If we step out of the picture a bit and look at the overall composition then
the question becomes why weren't more mappings added. Either this was a
conscious decision or it was just an oversight of more elaborate use-cases.

I think the design goal is to keep as minimal a range of ids in a systemd-homed
directory with the ultimate ideal that everything is owned by
UID_NOBODY:GID_NOBODY on-disk.

That obviously isn't guaranteed to work since there will be daemons and
programs that will create files with specific {g,u}id in a user's home
directory.

But the assumption is that there's supposed to be only one guiding {g,u}id
which is assigned to the user in the range [HOME_UID_MIN,HOME_UID_MAX]. That
random assigned {g,u}id will always map to UID_NOBODY on-disk.

But users can run containers as an unprivileged users since user namespaces can
be created - for better or worse - unprivileged users. So that's a use-case we
can't get rid of. For example, you can:

[myuser@fed35 ~]$ cat /etc/subuid
myuser:165536:100000

[myuser@fed35 ~]$ cat /etc/subgid
myuser:165536:100000

and then install e.g. LXC or Podman and do:

podman pull ubuntu
podman run -i -t ubuntu bash

or

[myuser@fed35 ~]$ cat .config/lxc/default.conf
lxc.idmap = u 0 165536 100000
lxc.idmap = g 0 165536 100000

[myuser@fed35 ~]$ lxc-create busy1 -t busybox

[myuser@fed35 ~]$ ls -al .local/share/lxc/
total 0
drwxr-xr-x 1 myuser myuser 10 Jan 25 09:17 .
drwx------ 1 myuser myuser 26 Jan 24 13:22 ..
drwxrwx--- 1 165536 myuser 24 Jan 25 09:17 busy1

[myuser@fed35 ~]$ ls -al .local/share/lxc/busy1/
total 4
drwxrwx--- 1 165536 myuser   24 Jan 25 09:17 .
drwxr-xr-x 1 myuser myuser   10 Jan 25 09:17 ..
-rw-r----- 1 myuser myuser 1811 Jan 25 09:17 config
drwxr-xr-x 1 165536 165536  110 Jan 25 09:17 rootfs

In order to run such containers tools like Podman and LXC will want to create
files on-disk as a {g,u}id other than the randomly assigned {g,u}id in order to
run containers. So per the idmapping in /etc/sub{g,u}id in the example above
they will want to create files as {g,u}id 165536 as that is what {g,u}id 0 in
the container will be mapped to on the host.

So the solution I'm proposing is relatively dumb. We simply add a range of 10m
ids upwards of 65534. So now the idmapping used by systemd-homed for idmapped
home directories will look like (assuming 60370 is the randomly assigned
{g,u}id for the homed user):

0 0 60001
60370 60370 1
60513 60513 5021
65535 65535 10000000

A range of 10m is generous and will allow users to run a fair bit of
containers with non-overlapping idmappings if they desire to do so.

The price of this is that home directories may contain {g,u}id other than
UID_NOBODY on-disk. If you have a very conservative stance and want to keep
UID_NOBODY to be the only legitimate {g,u}id on-disk then you can argue this is
a drawback. However, as I've said above that's already a reality today as we
e.g. map everything below the assigned {g,u}id.

Another argument could be to say that this makes home directories less portable
because you now have {g,u}ids on-disk that don't translate between machines.
But in contrast to real systemd {g,u}ids that might not translate between
different machines this isn't a problem for containers. The runtime will use
the {g,u}id specified in the container's config, i.e. the idmapping used by the
runtime will be the same no matter what machine you take it to.

I think in the long run users will likely want control how many ids upwards of
65534 to map because there might be crazy people that will want to have all ids
upwards of 65534 available which is why I mentioned this in the thread. We
don't need to enable this right now and can cross that bridge once users
request that.

An alternative to the approach of enabling users to create {g,u}ids on-disk
post 65534 is to make it a command-line switch. So by default homectl create
would not map anything upwards of UID_NOBODY. But if users specify the new
command-line switch they will get assigned 10m {g,u}id.

So, not real downsides imho, just architectural decisions.

gdonval:
Regarding design,

* Just in case, this gives the existing ID ranges in systemd (there is a summary table by the end) as well as the rational of why at least a full 16bit range should be allocated per user: https://systemd.io/UIDS-GIDS

* As said, the mapping already exists for the first 16-bit UIDs, hurting portability in a very similar way.

Now, would it be technically possible to mark container files as such and idmap them separately? 

If so, `systemd-homed` could read `/etc/sub{u,g}id` at activation time to determine container id ranges and perform the shift to the actual IDs on the fly.

Allowing user to map things down to low IDs (even if it's just on disk) is probably a big no-no but another ID range could be used instead and idmapped appropriately.

> Example:
> 
> Considering this configuration:
> 
> `$ homectl create foo --internal-id-map=100000:65536`
> `$ echo "foo:200000:65536" > /etc/subuid`
>
> Let `/home/foo/container/random_file` be a file marked as a "container file". If its on-disk UID is 100002, it would be idmapped to 200002, following `/etc/subuid` data.
>
> Now if instead we had
> `$ cat /etc/subuid`
> foo:420000:65536
> `/home/foo/container/random_file` with on-disk UID 100002 would be idmapped to 420002.

I have no idea of whether this is possible but if it is, we get our portability back.

anthr76:
Design implementation aside this patch is working well! I hope to see it or something similar included soon. Thanks for your work @brauner

poettering:
So, the point of homed home directories are that they are reasonably independent of local UID assignments, i.e. that they UID mappings are not propagated from the home dir to /etc and vice versa. The mappings we establish so far mean we pick an UID for a user on first login (i.e. the moment when it is bound to the local system) and  map that to "nobody". The home area doesn't store which UID we picked, and the assignment only exists as long as the "binding" of the home dir to the host is left in place (this "binding" info is stored in /var). i.e. the UID mapping is a property of the host, not of the home area, the home area doesn't store that info *at* *all* and still everything works fine.

/etc/subuid means assigning fixed UID ranges persistently to a user afaiu, so that's very much opposite to what homed is supposed to be about. i.e. if you add an entry to /etc/subuid, and use it inside of $HOME and then move the homedir elsewhere and attach it, then those files are not usable: there is no meaning associated to these UIDs in the new system, so they are just UIDs without records.

Now, in homed we currently map the UIDs from the lower 16bit range — with the exception of the homed range — 1:1. This is precisely to allow containers to be stored there in a reasonably safe fashion, so that they can be run with little privileges, by mapping them at invocation time with idmapping to some higher range (always excluding the homed range). i.e. if you really want to put containers in your homed dir (which btw I find strange given quota and stuff falling flat if you do), then do so in the 0…65535 range. Why bother with the higher ranges? if you then move your home area elsewhere it still makes as much sense as where you created it, given it's still mapped 0…65535 there, and you know that that's the basic OS range.

To me it appears /etc/subuid is a mistake in a multitude of ways, it creates major complexities not only for homed, but for anything that needs to manage UIDs (i.e. LDAP stuff, …). And that given that idmapped mounts make the usecase pretty much go away. i.e. with the advent with idmapped mounts you can make a container fs image with low UIDs appear at any UID range, dynamically, you never actually have to "physically" pass ownership of any file to the high UID range. So besides compat with old stuff, why would I ever want to use /etc/subuid? i.e. instead of managing persistent uid allocations in /etc via a SUID binary, why not do that dynamically in /run, and then immediately set up a idmapped mount as well of a specified container rootfs? That's a lot safer operation I am sure, and means the assignments become *runtime* concepts, not persistent concepts, and that's *so* *much* of a benefit. (Should we ever implement an "unpriv" mode to nspawn, this is how I'd do it: have a suid binary that does the heavy lifting, that takes a file system hierarchy, applies idmapping to it, applies the same userns mapping to the caller's process, adds a lock file for the range somewhere to /run/ and exits)
 
I know that podman advertises "rootless" containers (which in fact are very much root-full given they invoke a suid-root binary to do the heavy lifting), but doing so via /etc/subuid is not the only way, the idmapping way is much simpler and robust, and I am sure much more preferable since easier to manage.

Anyway, I understand there are people who believe in /etc/subuid either way, regardless how puzzled I might be about that. While I am pretty sure we should not integrate with the concept we can certainly support it enough so that people can manually make things work for them, i.e. that they copy the ranges around manually.

hence the suggested patch by @brauner looks conceptually OK, but we should carefully select the UID range to map 1:1 from the host. i.e. it needs to fit into https://systemd.io/UIDS-GIDS in some way, and probably be a range that is built around 2-exponent multiples instead of 10-exponent multiples (as UID ranges otherwise are mostly organized that way). Right now we already define the UID range 524288…1879048191 (aka 0x80000…0x6FFFFFFF) for use by containers, so maybe that's the mapping we should pass through 1:1.

gdonval:
> /etc/subuid means assigning fixed UID ranges persistently to a user afaiu

> if you encouter a UID 0…65536 in the homedir it means: "this is part of a container image, whose UID starts at zero"

A fixed pool of `subuid` is attributed to all homed users already then. So what remains is the idmapping at activation time to whatever is in /etc/subuid (or any other system, provided podman moves to dynamic idmappings too)?

What is the security implication of having (potentially SUID) UID 0 files in home that people can alter as they want as pseudo-root in container? 

> if you really want to put containers in your homed dir (which btw I find strange given quota and stuff falling flat if you do)

You find it strange because you've designed it therefore you know most limits and implications of everything and anything homed-related (among other broader stuff). Most people are not exactly there and just wondered why podman wasn't working in their setup (which is clear now).

poettering:
> A fixed pool of `subuid` is attributed to all homed users already then.

Hmm? Not sure I follow?

> So what remains is the idmapping at activation time to whatever is in /etc/subuid (or any other system, provided podman moves to dynamic idmappings too)?

Sorry, can't parse the question?

> What is the security implication of having (potentially SUID) UID 0 files in home that people can alter as they want as pseudo-root in container?

homed mounts $HOME with nosuid, so the SUID stuff shouldn't have an effect. In general: I think putting container dirs/files owned by other users in $HOME is a bad idea. i.e. files/dirs you cannot remove yourself in the general case is just a bad idea. Put container images in /var/ somewhere, $HOME is really not the right place.



> > if you really want to put containers in your homed dir (which btw I find strange given quota and stuff falling flat if you do)
> 
> You find it strange because you've designed it therefore you know most limits and implications of everything and anything homed-related (among other broader stuff). Most people are not exactly there and just wondered why podman wasn't working in their setup (which is clear now).

I did not design UNIX quota. On UNIX quota is traditionally built around UIDs, under the assumption that each user gets one UID and one homedir. This all is just broken if you place files you don't own in $HOME: the stuff belongs to you conceptually apparently, but also not, because it's not acounted to your quota and you cannot delete it. So it's really a broken idea to put that stuff in $HOME. I mean, knock yourself out, but this will always be half-broken and a bad idea. Might be OK for "hacker" type workflows, who have root privs anyway just a "sudo" away, but for clean workflows it's just totally conceptually broken.

This is all entirely independent of homed btw. podman shouldn't place containers in $HOME, regardless if you use homed or not, it's just broken either way.

anthr76:
@rhatdan Are there any input on how podman can change things that would better incorporate with homed?

poettering:
> That obviously isn't guaranteed to work since there will be daemons and
> programs that will create files with specific {g,u}id in a user's home
> directory.

So that's something I am not sure about. Except for container managers (where this is a very questionnable approach, given this is generally not compatible with quota logic, and means you will have files in your home directory you cannot delete or access, and you lose compatibility with a variety of network filesystems often used to back home directories) I am not aware of anything doing that. Are you? The thing is: placing files owned by user X into a directory owned by user Y in a safe way is very hard on UNIX, due to symlink and similar attacks, hence I am really sure its nothing we should ever advertise as being a good idea. It might be OK to do that in local "hacker" style codepaths, but not in clean, documented, advertised codepaths.

It appears to me that outside of a specialist users who have sudo privs anyway on the local machine placing files/directories inside of $HOME that are not owned by the user whose $HOME it is is a bad idea, and nothing we should advertise or suggest people to use. But at the same time I understand people do this anyway, i.e. precisely these specialist users, hence I am OK with the approach suggested here. I just wish podman wouldn't do this by default, and instead just store is containers elsewhere. Given it needs privs to set up the UID mappings anyway it might as well put the containers in /var, and avoid the ownership mess it is doing in $HOME

poettering:
Anyway, I prepped #22526 now, which is inspired/based on @brauner's patch but makes some adjustments, and adds docs. Let's continue discussion there. Hope that's OK.

bluca:
This was implemented via https://github.com/systemd/systemd/pull/22526

poettering:
A user cannot write to container trees dropped into their home directory, the files will not be owned by them and they will not be quota-assigned to them. If you need a suid-root/fcaps tool to access them (newuidmap, wrapped by podman unshare), then they really shouldn't be in the user's home directory, but somewhere else.

There are two independent things I think podman should fix:

1. stop putting files not owned by the user's own UID into $HOME, i.e. drop container OS trees it manages for each user in /var/lib/podman/user-containers/$USER/… or so. given you need specialist privileged (i.e. suidroot/fcaps) tools to add/remove/modify them anyway you can just pick put them pretty much anywhere you want. They really shouldn't be in $HOME where the user is expected to make modifications with any file browser tool of their choice. If you want you can then add a symlink (or maybe even bind mount, but that's icky too) from the home dir to that dir in /var/lib/ if you like.

2. stop persistently allocating uid/gid ranges for containers. Instead put them at UID offset 0 in the dir of choice, and map them to high userns uid/gid ranges *at* *runtime* via uidmapped bind mounts. That makes allocation cheap and something you don't have to think about anymore, and requires zero propagation of "subuid" ranges or so between systems. No static configuration files in /etc/uidmap, admins thinking about allocating ranges and whatnots.

or in other words: it's pretty arbitrary to say that that `newuidmap` being setuid-root/fcaps  is OK (and claiming this was "root-less"), but then not going 2 steps further and saying that it's also OK to slightly extend that privileged tool to set up a uidmapped bind mount, or even to drop files into /var/lib/<somewhere> instead of $HOME where they really don't belong.

if podman would follow that logic then the quota issues, NFS issues, permission issues, suid issues all would just go away...

gdonval:
> stop putting files not owned by the user's own UID into $HOME

The premise is wrong here: my container files are mine as a user. `/home/user` is for `user` data. Period.

I couldn't care less what technicalities makes that inconvenient. If homed had a 4GB limit on file size, I'm sure you'd be arguing home is not the right place of ISOs and videos.

> given you need specialist privileged...

...then you want to put your tree in a (growable) loopback device (or similar) whose UID on disk is `nobody` like everything else in home.

>  put them at UID offset 0 in the dir of choice, and map them to high userns uid/gid ranges at runtime via uidmapped bind mounts

Glad to see https://github.com/systemd/systemd/pull/22239#issuecomment-1021134109 is possible after all!

ljrk0:
I just want to note here that podman is currently moving to support idmapped mounts, hopefully fixing this "properly" soon:
https://github.com/containers/podman/issues/10374

ljrk0:
> stop putting files not owned by the user's own UID into $HOME, i.e. drop container OS trees it manages for each user in /var/lib/podman/user-containers/$USER/… or so. given you need specialist privileged (i.e. suidroot/fcaps) tools to add/remove/modify them anyway you can just pick put them pretty much anywhere you want. They really shouldn't be in $HOME where the user is expected to make modifications with any file browser tool of their choice. If you want you can then add a symlink (or maybe even bind mount, but that's icky too) from the home dir to that dir in /var/lib/ if you like.

This is actually possible:

https://www.redhat.com/sysadmin/nfs-rootless-podman

Put 
```toml
[storage]
driver="overlay"
rootless_storage_path="/var/lib/podman/user-containers/$USER/storage"
```
into your `$XDG_CONFIG_DIR/containers/storage.conf` or `/etc/containers/storage.conf`.

(there's a typo in the blog post that says `$/var`, ignore the superfluous dollar sign).


Although I'm partly with @gdonval as well:  Ideally I would want to carry my containers around (I mean, that's part of the use-case of systemd-homed anyway, right?).  Also, I may be building containers of my "personal" projects which I'd prefer to be stored in my encrypted homed and they should also count against my quota... .

If I understood correctly, the main issue is, that the files are owned by not-the-user.  I frankly don't understand why that couldn't be adjusted, that is, have them stored under whatever uid the user is running and only map them to root within the username space of the running container?  But I don't know much about the depths of neither containers, their images, nor user namespaces or related technologies.  N.B., flatpak doesn't use subuid AFAIK but still manages to have "user" applications run and stored locally.

ljrk0:
Also, podman is currently investigating the use of id-mapping but only for "rootful" containers. rootless likely won't receive support (due to missing overlayfs features if I don't mix up things here).

brauner:
> Also, podman is currently investigating the use of id-mapping but only for "rootful" containers. rootless likely won't receive support (due to missing overlayfs features if I don't mix up things here).

If there are no surprises then kernel 5.19 will support overlayfs being mountable on top of idmapped layers.

behrmann:
> If I understood correctly, the main issue is, that the files are owned by not-the-user. I frankly don't understand why that couldn't be adjusted, that is, have them stored under whatever uid the user is running and only map them to root within the username space of the running container? But I don't know much about the depths of neither containers, their images, nor user namespaces or related technologies. N.B., flatpak doesn't use subuid AFAIK but still manages to have "user" applications run and stored locally.

@ljrk0 If my understanding is correct, the problem here is, that id-mapping is a constant uid-shift. A container does not have all files from a single uid, though, so even if you the root-owned files in the container would belong to a user's uid number, the non-root owned files in the container would still not belong to the user, which plays havoc with quota, since quota is uid number based and not "all files below this path belong to this user".

I too understand the desire to take container images elsewhere, just like a home directory, but I guess the difference is that a container should be throwaway and easily recreatable (e.g. from config in your home), so that it doesn't need to live there as well. This does come with the downside of needing to recreate them, which in some sense is wasteful, but this is hitting against UNIX design walls, unfortunately.

PR #22240, opened 2022-01-24
portable: add GetImageStateWithExtensions method

bluca:
Allow to correctly query a layered portable service for
attached/detached state.

### DIFF
--- a/src/portable/portablectl.c
+++ b/src/portable/portablectl.c
@@ -1034,11 +1034,11 @@ static int set_limit(int argc, char *argv[], void *userdata) {
 }
 
 static int is_image_attached(int argc, char *argv[], void *userdata) {
-        _cleanup_(sd_bus_message_unrefp) sd_bus_message *reply = NULL;
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *m = NULL, *reply = NULL;
         _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
         _cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus = NULL;
         _cleanup_free_ char *image = NULL;
-        const char *state;
+        const char *state, *method;
         int r;
 
         r = determine_image(argv[1], true, &image);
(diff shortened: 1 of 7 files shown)


src/portable/portabled-bus.c
```diff
@@ -235,9 +237,28 @@ static int method_get_image_state(sd_bus_message *message, void *userdata, sd_bu
...
+                r = sd_bus_message_read_strv(message, &extension_images);
+                if (r < 0)
+                        return r;
+
+                r = sd_bus_message_read(message, "t", &input_flags);
+                if (r < 0)
+                        return r;
+
+                /* No flags are supported by this method for now. */
+                if (input_flags!= 0)
```
yuwata: missing white space.

src/portable/portablectl.c
```diff
@@ -1049,9 +1049,29 @@ static int is_image_attached(int argc, char *argv[], void *userdata) {
...
+
+        r = bus_message_new_method_call(bus, &m, bus_portable_mgr, method);
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = sd_bus_message_append(m, "s", image);
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = attach_extensions_to_message(m, arg_extension_images);
```
yuwata, line 1062: Is this OK to be called for `GetImageState`?

yuwata (changes requested):

src/portable/portablectl.c
bluca: yes, this helper is made to make that simpler, it will do nothing if the strv is empty

yuwata: Thanks.

yuwata: Hm, it calls `sd_bus_message_open_container(m, 'a', "s"); sd_bus_message_close_container(m);` even if `arg_extension_images` is empty. I cannot remember right now and not sure, is it really does nothing to the message?

I think it is simpler to move this to the if block below.

bluca: It returns immediately as the first thing if the strv is empty:

https://github.com/systemd/systemd/blob/f338a496e48392929400916ebb0eb9819dec1e2b/src/portable/portablectl.c#L98

yuwata: Ouch. Indeed.

yuwata (approved):

PR #22245, opened 2022-01-25
portable: add flag to return extension-releases in GetImageMetadataWithExtensions

bluca:
Return the name of each extension and the associated extension-release
file, and pretty-print them in 'portablectl inspect', if a new flag
is passed.

```
$ portablectl inspect --extension app2 --extension app0  minimal app0 app1
(Matching unit files with prefixes 'app0', 'app1'.)
Image:
        /run/portables/minimal.raw
Portable Service:
        n/a
Operating System:
        Debian GNU/Linux 10 (buster)
Extension:
        /run/portables/app2.raw
        Extension Scope:
                n/a
        Extension Compatibility Level:
                n/a
        Portable Service:
                n/a
        Portable Prefixes:
                n/a
        Operating System:
                n/a (debian 10)
Extension:
        /run/portables/app0.raw
        Extension Scope:
                n/a
        Extension Compatibility Level:
                n/a
        Portable Service:
                n/a
        Portable Prefixes:
                n/a
        Operating System:
                n/a (debian 10)
Unit files:
        app0.service
```

### DIFF
--- a/man/org.freedesktop.portable1.xml
+++ b/man/org.freedesktop.portable1.xml
@@ -193,7 +193,15 @@ node /org/freedesktop/portable1 {
       This method is a superset of <function>GetImageMetadata()</function> with the addition of
       a list of extensions as input parameter, which were overlaid on top of the main
       image via <function>AttachImageWithExtensions()</function>.
-      The <varname>flag</varname> parameter is currently unused and reserved for future purposes.</para>
+      The <varname>flag</varname> parameter can be used to request that, before the units, the path of
+      each extension and an array of bytes with the content of the respective extension-release file
+      are sent. One such structure will be sent for each extension named in the input arguments. The
+      flag value to enable this functionality is defined as follows:</para>
+
+      <programlisting>
+#define PORTABLE_INSPECT_EXTENSION_RELEASES  (UINT64_C(1) &lt;&lt; 1)
+      </programlisting>
+
 
       <para><function>GetImageState()</function> retrieves the image state as one of the following
       strings:
--- a/src/portable/portable.c
+++ b/src/portable/portable.c
@@ -512,6 +512,7 @@ static int extract_image_and_extensions(
                 bool validate_sysext,
                 Image **ret_image,
                 OrderedHashmap **ret_extension_images,
+                OrderedHashmap **ret_extension_releases,
                 PortableMetadata **ret_os_release,
                 Hashmap **ret_unit_files,
                 char ***ret_valid_prefixes,
@@ -519,7 +520,7 @@ static int extract_image_and_extensions(
 
         _cleanup_free_ char *id = NULL, *version_id = NULL, *sysext_level = NULL;
         _cleanup_(portable_metadata_unrefp) PortableMetadata *os_release = NULL;
-        _cleanup_ordered_hashmap_free_ OrderedHashmap *extension_images = NULL;
+        _cleanup_ordered_hashmap_free_ OrderedHashmap *extension_images = NULL, *extension_releases = NULL;
         _cleanup_hashmap_free_ Hashmap *unit_files = NULL;
         _cleanup_strv_free_ char **valid_prefixes = NULL;
         _cleanup_(image_unrefp) Image *image = NULL;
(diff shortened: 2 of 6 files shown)


src/portable/portablectl.c
```diff
@@ -366,6 +366,74 @@ static int inspect_image(int argc, char *argv[], void *userdata) {
...
+                                return log_error_errno(errno, "Failed to open extension-release buffer: %m");
+
+                        r = parse_env_file(f, name,
+                                                "ID", &id,
+                                                "VERSION_ID", &version_id,
+                                                "SYSEXT_SCOPE", &sysext_scope,
+                                                "SYSEXT_LEVEL", &sysext_level,
+                                                "PORTABLE_PRETTY_NAME", &pretty_portable,
+                                                "PORTABLE_PREFIXES", &portable_prefixes,
+                                                "PRETTY_NAME", &pretty_os);
```
yuwata: Indentation?

src/portable/portablectl.c
```diff
@@ -700,6 +768,15 @@ static int maybe_stop_disable(sd_bus *bus, char *image, char *argv[]) {
         if (r < 0)
                 return bus_log_parse_error(r);
 
+
```
yuwata: Doubled blank line.

yuwata (approved):
LGTM. Two minor points.

src/portable/portable.h
```diff
@@ -21,13 +21,14 @@ typedef struct PortableMetadata {
 #define PORTABLE_METADATA_IS_UNIT(m) (!IN_SET((m)->name[0], 0, '/'))
 
 typedef enum PortableFlags {
-        PORTABLE_RUNTIME        = 1 << 0, /* Public API via DBUS, do not change */
-        PORTABLE_PREFER_COPY    = 1 << 1,
-        PORTABLE_PREFER_SYMLINK = 1 << 2,
-        PORTABLE_REATTACH       = 1 << 3,
-        _PORTABLE_MASK_PUBLIC   = PORTABLE_RUNTIME,
+        PORTABLE_RUNTIME                    = 1 << 0,
```
poettering, line 24: the ` /* Public API via DBUS, do not change */` comment was lost here?

man/org.freedesktop.portable1.xml
```diff
@@ -193,7 +193,15 @@ node /org/freedesktop/portable1 {
       This method is a superset of <function>GetImageMetadata()</function> with the addition of
       a list of extensions as input parameter, which were overlaid on top of the main
       image via <function>AttachImageWithExtensions()</function>.
-      The <varname>flag</varname> parameter is currently unused and reserved for future purposes.</para>
+      The <varname>flag</varname> parameter can be used to request that, before the units, the path of
+      each extension and an array of bytes with the content of the respective extension-release file
+      are sent. One such structure will be sent for each extension named in the input arguments. The
+      flag value to enable this functionality is defined as follows:</para>
```
poettering, line 199: I am not a fan of this I must say. THis is paramete roverloading, no?

it sounds to me as if this should be a new function that has another array with the extensions with their names and then a nested array of units and fields for the extension release data.

i.e. please let's keep the method call parameter list clean and normalized, i.e. don't pass data through return parameters that actually exist for a different kind of thing.

src/portable/portable.c
```diff
@@ -599,10 +607,15 @@ static int extract_image_and_extensions(
                 if (r < 0)
                         return r;
 
-                if (!validate_sysext && !ret_valid_prefixes)
+                if (!validate_sysext && !ret_valid_prefixes && !ret_extension_releases)
                         continue;
 
-                r = take_fdopen_unlocked(&extension_release_meta->fd, "r", &f);
+                /* We need to keep the fd valid, to return the PortableMetadata to the caller. */
+                extension_release_fd = fd_reopen(extension_release_meta->fd, O_CLOEXEC);
```
poettering, line 614: should specify O_RDONLY

poettering (commented):
Looks conceptually OK, but please let's fix the the method parameters: don't shove unrelated data through existing parameters. It sucks that D-Bus method parameter lists are not reasonably extensible, but it is what it is, and let's not resort to tricks like overloading existing parameters with different meanings depending on some flag.

Hence, please add a new superset method for this.

src/portable/portable.h
bluca: Moved down - it's an enum, I don't think it's useful to duplicate it, anything up to the marker is fixed, anything after the marker is free game

bluca:
> Looks conceptually OK, but please let's fix the the method parameters: don't shove unrelated data through existing parameters. It sucks that D-Bus method parameter lists are not reasonably extensible, but it is what it is, and let's not resort to tricks like overloading existing parameters with different meanings depending on some flag.
> 
> Hence, please add a new superset method for this.

The problem is that this is how it should have been from the beginning - so if anything, I'd prefer to break compatibility, but that's nasty. I don't think it makes much sense to have a GetImageMetadataWithExtensionsWithExtensions, it just looks like a copypasta...

src/portable/portable.c
bluca: https://github.com/systemd/systemd/pull/22405

PR #22247, opened 2022-01-26
basic: mac_[selinux,smack]_apply_fd does not work when applying labels

hoiho-amzn:
mac_smack_apply_fd() only works when the file descriptor for the device
node is opened with O_RDWR|O_WRONLY mode, but udev only opens with
read-only by default, so SMACK label apply fails.

Use mac_smack_apply() instead which does not have such restriction.

### DIFF
--- a/src/shared/selinux-util.c
+++ b/src/shared/selinux-util.c
@@ -346,7 +346,7 @@ int mac_selinux_apply_fd(int fd, const char *path, const char *label) {
 
         assert(label);
 
-        if (fsetfilecon(fd, label) < 0)
+        if (setfilecon(FORMAT_PROC_FD_PATH(fd), label) < 0)
                 return log_enforcing_errno(errno, "Failed to set SELinux security context %s on path %s: %m", label, strna(path));
 #endif
         return 0;
--- a/src/shared/smack-util.c
+++ b/src/shared/smack-util.c
@@ -95,9 +95,9 @@ int mac_smack_apply_fd(int fd, SmackAttr attr, const char *label) {
                 return 0;
 
         if (label)
-                r = fsetxattr(fd, smack_attr_to_string(attr), label, strlen(label), 0);
+                r = setxattr(FORMAT_PROC_FD_PATH(fd), smack_attr_to_string(attr), label, strlen(label), 0);
         else
-                r = fremovexattr(fd, smack_attr_to_string(attr));
+                r = removexattr(FORMAT_PROC_FD_PATH(fd), smack_attr_to_string(attr));
         if (r < 0)
                 return -errno;
 


src/udev/udev-node.c
```diff
@@ -690,7 +690,7 @@ int udev_node_apply_permissions(
                         } else if (streq(name, "smack")) {
                                 smack = true;
 
-                                q = mac_smack_apply_fd(node_fd, SMACK_ATTR_ACCESS, label);
+                                q = mac_smack_apply(devnode, SMACK_ATTR_ACCESS, label);
```
yuwata: Not familiar with SMACK, but how about to use `fd_reopen()`?

yuwata (commented):
`mac_smack_apply_fd()` is used at one more place below. Is it not necessary to convert it?

src/udev/udev-node.c
hoiho-amzn: Sure, but why not add `O_RDWR` in the `open` call in line 300 - the end state would be the same.

hoiho-amzn:
> `mac_smack_apply_fd()` is used at one more place below. Is it not necessary to convert it?

On a second thought, we probably should keep using the *_fd() functions but open the file descriptor with write access so the underlying calls to apply smack/selinux label won't fail.

src/udev/udev-node.c
yuwata: Not sure about SELinux, but, at least `fchmod_and_chown()` does not require `O_RDWR`, no? So, I'd like to open files in rw on demand.

yuwata:
> > `mac_smack_apply_fd()` is used at one more place below. Is it not necessary to convert it?
> 
> On a second thought, we probably should keep using the *_fd() functions but open the file descriptor with write access so the underlying calls to apply smack/selinux label won't fail.

That sounds good for me.

src/udev/udev-node.c
hoiho-amzn: Let me give a shot in my internal environment.

hoiho-amzn: After some more digging, I think commit a7fdc6c was the root cause as it opened the fd using the `O_PATH` option. This will cause `fsetxattr()` to fail with a bad file descriptor error according to the man page for open.

I'm trying to see if I can do a `fd_reopen` with the right mode so the labelling will go through.

hoiho-amzn:
Updated with using `fd_reopen` approach.

poettering:
Please change mac_smack_apply_fd() to work on O_PATH and O_RDONLY fds instead, i.e. replace its current use of `fsetxattr(fd, …)` by `setxattr(FORMAT_PROC_FD_PATH(fd), …)`.

i.e. always go via /proc/self/fd/… instead of the fd directly, so that the operation is run on the inode afresh insteda of via our limited fd.

hoiho-amzn:
Updated with remapping the fd into a full path. Much cleaner.

yuwata (approved):
LGTM.

PR #22248, opened 2022-01-26
network: tc: use request queue to configure QDisc or TClass

yuwata:


src/libsystemd/sd-netlink/netlink-message-rtnl.c
```diff
@@ -1072,6 +1072,42 @@ int sd_rtnl_message_new_traffic_control(
         return 0;
 }
 
+int sd_rtnl_message_traffic_control_get_ifindex(sd_netlink_message *m, int *ret) {
+        struct tcmsg *tcm;
+
+        assert_return(rtnl_message_type_is_traffic_control(m->hdr->nlmsg_type), -EINVAL);
```
bluca, line 1080: assert/check on `m`, `m->hdr` and tcm? Same below

bluca (approved):
LGTM, just one nit

bluca:
Also would be good to have some tests for coverage, but not sure how feasible it is

yuwata:
@bluca Thank you for the review. The above point is addressed. Upgrading the green label.

> Also would be good to have some tests for coverage, but not sure how feasible it is

We already have tests for traffic control, and now the sections about traffic control are stored in hashmap (hence the order is randomized). If the existing tests are green, then it means the resolving dependencies between sections works fine.

PR #22250, opened 2022-01-26
shared: Ensure COPY_HOLES copies trailing holes

daandemeyer:
Previously, files with a hole at the end would get silently truncated
which breaks reading journal files. This commit makes sure that holes
are punched in existing space and if no more space is available, that
we grow the file and the hole by using ftruncate().

The corresponding test is extended to put a hole at the end of the file
and we make sure that hole is copied correctly.

### DIFF
--- a/src/shared/copy.c
+++ b/src/shared/copy.c
@@ -107,6 +107,45 @@ static int look_for_signals(CopyFlags copy_flags) {
         return 0;
 }
 
+static int create_hole(int fd, off_t size) {
+        off_t offset;
+        off_t end;
+
+        offset = lseek(fd, 0, SEEK_CUR);
+        if (offset < 0)
+                return -errno;
+
+        end = lseek(fd, 0, SEEK_END);
+        if (end < 0)
+                return -errno;
+
+        /* If we're not at the end of the target file, punch a hole in the existing space using fallocate(). */
+
+        if (offset < end && fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, MIN(size, end - offset)) < 0)
+                return -errno;
+
+        if (end - offset >= size) {
+                /* If we've created the full hole, set the file pointer to the end of the hole we created and exit. */
+                if (lseek(fd, offset + size, SEEK_SET) < 0)
+                        return -errno;
+
+                return 0;
+        }
+
+        /* If we haven't created the full hole, use ftruncate() to grow the file (and the hole) to the
+         * required size and move the file pointer to the end of the file. */
+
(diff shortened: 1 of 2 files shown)


src/shared/copy.c
```diff
@@ -107,6 +107,45 @@ static int look_for_signals(CopyFlags copy_flags) {
         return 0;
 }
 
+static int create_hole(int fd, off_t size) {
+        off_t offset;
+        off_t end;
+
+        offset = lseek(fd, 0, SEEK_CUR);
+        if (offset < 0)
+                return offset;
```
yuwata: `return -errno;`?

src/shared/copy.c
```diff
@@ -107,6 +107,45 @@ static int look_for_signals(CopyFlags copy_flags) {
...
+
+        end = lseek(fd, 0, SEEK_END);
+        if (end < 0)
+                return end;
```
yuwata: Also here.

src/shared/copy.c
```diff
@@ -107,6 +107,45 @@ static int look_for_signals(CopyFlags copy_flags) {
...
+
+        /* If we're not at the end of the target file, punch a hole in the existing space using fallocate(). */
+
+        if (offset < end && fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, MIN(size, end - offset)) < 0)
+                return -errno;
+
+        if (end - offset >= size) {
+                /* If we've created the full hole, set the file pointer to the end of the hole we created and exit. */
+                if (lseek(fd, offset + MIN(size, end - offset), SEEK_SET) < 0)
```
yuwata: Not a matter, here `MIN()` is not necessary, and `offset + size` is enough.

src/test/test-copy.c
```diff
@@ -340,23 +341,32 @@ TEST_RET(copy_holes) {
...
 
-        assert_se(lseek(fd, 1024 * 1024, SEEK_CUR) >= 0);
-        assert_se(write(fd, "abc", strlen("abc")) >= 0);
+        /* We need to make sure to create hole in multiples of the block size, otherwise filesystems (btrfs)
+         * might silently truncate/extend the holes. */
+
+        assert_se(lseek(fd, blksz, SEEK_CUR) >= 0);
+        assert_se(write(fd, buf, blksz) >= 0);
+        assert_se(lseek(fd, 0, SEEK_END) == 2 * blksz);
+        // /* Only ftruncate() can create holes at the end of a file. */
```
yuwata: Unnecessary `//`.

yuwata (approved):
LGTM. Several minor issues.

PR #22251, opened 2022-01-26
boot: Small improvements

medhefgo:


bluca (approved):

src/boot/efi/boot.c
```diff
@@ -1717,88 +1716,83 @@ static void config_default_entry_select(Config *config) {
                 config->timeout_sec = 10;
 }
 
-static BOOLEAN find_nonunique(ConfigEntry **entries, UINTN entry_count) {
-        BOOLEAN non_unique = FALSE;
+static BOOLEAN entries_unique(ConfigEntry **entries, BOOLEAN *unique, UINTN entry_count) {
+        BOOLEAN is_unique = TRUE;
 
         assert(entries);
```
yuwata, line 1724: assert(unique);

src/boot/efi/boot.c
```diff
@@ -1717,88 +1716,83 @@ static void config_default_entry_select(Config *config) {
...
 
-        return non_unique;
+        return is_unique;
 }
 
 /* generate a unique title, avoiding non-distinguishable menu entries */
 static void config_title_generate(Config *config) {
         assert(config);
 
+        BOOLEAN unique[config->entry_count];
```
yuwata, line 1742: Variable-length array? Is it now allowed in our coding style?

yuwata (approved):
LGTM. Two minor points. Feel free to merge this as is if you do not agree with the comments.

src/boot/efi/boot.c
yuwata: Hm, it is allowed since e3bde91293df336c8839a1a7a79ed50f954150db. I did not know that.

PR #22256, opened 2022-01-26
man: Fix paths for user units (transient/generator.early)

Debilski:
transient and generator.early user units were documented with the path for the system unit.


anitazha:
Thanks!

PR #19384, opened 2021-04-21
network: RFC3046 implementation

yegor-alexeyev:
Final changeset which addresses https://github.com/systemd/systemd/issues/7650

PR #22261, opened 2022-01-26
tests: fuzz dhcp_server_relay_message

evverx:
It's a follow-up to https://github.com/systemd/systemd/pull/19384 where
dhcp_server_relay_message was introduced.

This PR was prompted by https://github.com/systemd/systemd/pull/22236#issuecomment-1020113269
for the most part.

@yuwata could you take a look?


yuwata (approved):
LGTM. Thanks.

PR #21183, opened 2021-10-29
journal: Add compact mode 

daandemeyer:
The following improvements are gated behind the new compact mode:

- 32-bit offsets for Entry Array object payloads
- Use a Trie to store Entry object offsets
- Support for inlining specific fields (and their values) pairs into the Entry object (defaults to MESSAGE only)
- Support for not indexing specific fields
- Boot ID deduplication

The Trie commit contains a more detailed explanation of how it works. The changes should be backwards compatible. journalctl can both read new journals written by the updated systemd-journald and the current journals generated by a Fedora 35 system.

All credits go to @mokomull for the trie idea. The original experiment which I started from can be found here: https://github.com/mokomull/systemd/commit/1b1b3d06b55293526de425cb5696b35a70468e27

PR #22262, opened 2022-01-26
Journal fixes

daandemeyer:
A bunch of smaller journal fixes split out from #21183.

The biggest change here is probably the --verify change. Previously, we'd iterate each data object's entry array and for each entry, we'd try to find a reference to the data object in its entry items.

Now, when iterating all entry items, we additionally check if the data object pointed to by that entry item also has that entry's offset in its entry array.

Conceptually, this is the same check, but it turns out doing it this way is faster than doing it the other way (+-10s faster on a 4G journal on my laptop). It's also a lot faster (+-40s on a 4G journal on my laptop) for the compact mode from #21183.


yuwata (approved):
LGTM.

anitazha (commented):
LGTM too

PR #22268, opened 2022-01-27
test: initialize buffer to make Coverity silent

yuwata:
Follow-up for f82f0b993740f4a9a3719c4df7c185411710df2b.

Fixes CID#1469119.

### DIFF
--- a/src/test/test-copy.c
+++ b/src/test/test-copy.c
@@ -329,6 +329,7 @@ TEST_RET(copy_holes) {
         struct stat stat;
         off_t blksz;
         int r, fd, fd_copy;
+        char *buf;
 
         fd = mkostemp_safe(fn);
         assert_se(fd >= 0);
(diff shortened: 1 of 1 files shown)


src/test/test-copy.c
```diff
@@ -343,7 +344,7 @@ TEST_RET(copy_holes) {
 
         assert_se(fstat(fd, &stat) >= 0);
         blksz = stat.st_blksize;
-        char buf[blksz];
+        buf = alloca0(blksz);
```
bluca, line 347: Maybe just initialize with `= {0}`?

yuwata: VLA seems not support `= {0};`
```
../src/test/test-copy.c: In function ‘test_copy_holes’:
../src/test/test-copy.c:347:9: error: variable-sized object may not be initialized
  347 |         char buf[blksz] = {0};
      |         ^~~~
```

bluca: got it

bluca (approved):

Issue #22058, opened 2022-01-08
System V Compatbility Shutdown/Reboot Commands Should Honor Inhibitors and Send PrepareForShutdown DBus Signal When Run As Root


JonnyWhatshisface:
**Description of Behavior**

System V compatibility shutdown/reboot/halt commands (ie via /usr/sbin/{shutdown,reboot}) do not exhibit the same behavior as a systemctl {reboot,halt} when run as root. 

Most notable is that inhibitors appear to be skipped and a PrepareForShutdown DBus signal is not emitted when immediate action via /usr/sbin/{reboot,shutdown,halt,poweroff} are executed. 

This behavior changes when using /usr/sbin/shutdown in a scheduled manner - i.e. /usr/sbin/shutdown -r +1 . In this case, delay inhibitors are honored and the DBus signal goes out, giving an inconsistent behavior in shutdown, reboot and poweroff operations between systemctl and the System V compatibility commands. 

**Example Test Case**
Install required python depedencies:

```
apt install libdbus-1-dev libglib2.0-dev python3-gi pip
pip3 install dbus-python
```

Use the following sample script:

```
#!/usr/bin/python3

import time, os, sys
import dbus
from gi.repository import GLib
from dbus.mainloop.glib import DBusGMainLoop

def handle_shutdown(*args):
    # Note: Default maximum inhibit delay time is 5 seconds.
    # This can be raised in /etc/systemd/login.conf by
    # setting InhibitDelayMaxSec .
    print("Inhibit the shutdown for 30 seconds...")
    time.sleep(30)
    print("Letting go of inhibit lock")
    os.close(inhibitor_fd)

DBusGMainLoop(set_as_default=True)
bus = dbus.SystemBus()

login = bus.get_object("org.freedesktop.login1", "/org/freedesktop/login1")
manager = dbus.Interface(login, dbus_interface='org.freedesktop.login1.Manager')
inhibitor = manager.Inhibit("shutdown", "Shutdown Handler", "Handle events on shutdown notification", "delay")

inhibitor_fd = inhibitor.take()

bus.add_signal_receiver(handle_shutdown, dbus_interface="org.freedesktop.login1.Manager", signal_name="PrepareForShutdown")

loop = GLib.MainLoop()
loop.run()

```

Run the above script. 

As root, issue a systemctl reboot command. The inhibitor is honored, the PrepareForShutdown signal is sent and the inhibitor actions the signal.

[...]

PR #22269, opened 2022-01-27
Add LegacyHaltRootInhibit option

JonnyWhatshisface:
When enabled, executing shutdown, reboot, halt and poweroff actions via legacy system v commands will call logind_reboot(), allowing them to take advantage of inhibitors and the DBus PrepareForShutdown signal.

The default is no which keeps the current behavior as it is.

https://github.com/systemd/systemd/issues/22058


JonnyWhatshisface:
Closing to resubmit; accidentally used personal github instead of work / firm github.

PR #22271, opened 2022-01-27
Freeze manager if reexec fails

keszybz:


src/core/main.c
```diff
@@ -1977,13 +1975,13 @@ static int invoke_main_loop(
         assert(ret_error_message);
 
         for (;;) {
-                r = manager_loop(m);
-                if (r < 0) {
+                ManagerObjective obj = manager_loop(m);
+                if (obj < 0) {
                         *ret_error_message = "Failed to run main loop";
-                        return log_emergency_errno(r, "Failed to run main loop: %m");
+                        return log_emergency_errno(obj, "Failed to run main loop: %m");
```
yuwata: This is out of the scope in this commit, but is it not necessary to set `ret_retval` here?

src/core/main.c
```diff
@@ -3021,16 +3015,23 @@ int main(int argc, char *argv[]) {
...
+                             &switch_root_init,
+                             &error_message);
+        assert(r < 0 || IN_SET(r, MANAGER_EXIT,          /* MANAGER_OK is not expected here. */
+                                  MANAGER_RELOAD,
+                                  MANAGER_REEXECUTE,
+                                  MANAGER_REBOOT,
+                                  MANAGER_POWEROFF,
+                                  MANAGER_HALT,
+                                  MANAGER_KEXEC,
+                                  MANAGER_SWITCH_ROOT));
```
yuwata, line 3041: Or should we set `retval` here if `r < 0`??

src/core/main.c
```diff
@@ -3045,13 +3052,14 @@ int main(int argc, char *argv[]) {
...
-                             switch_root_init,
-                             &error_message); /* This only returns if reexecution failed */
+                r = do_reexecute(r,
+                                 argc, argv,
+                                 &saved_rlimit_nofile,
+                                 &saved_rlimit_memlock,
+                                 fds,
+                                 switch_root_dir,
+                                 switch_root_init,
+                                 &error_message); /* This only returns if reexecution failed */
```
yuwata, line 3062: Same here. Is it not necessary to convert negative `r` into `retval = EXIT_FAILURE` or something?

bluca (approved):
Do we have any tests for this functionality?

src/core/main.c
```diff
@@ -1977,13 +1975,13 @@ static int invoke_main_loop(
         assert(ret_error_message);
 
         for (;;) {
-                r = manager_loop(m);
-                if (r < 0) {
+                ManagerObjective obj = manager_loop(m);
```
poettering: manager_loop() returns `int`, not `ManagerObjective`?

`obj` is a weird name. I am pretty sure most people would assume it refers to an "object" of some form, not an "objective".

Maybe name it "objective"? or "ov"?

src/core/main.c
```diff
@@ -3021,16 +3015,23 @@ int main(int argc, char *argv[]) {
...
-                                &saved_rlimit_nofile,
-                                &saved_rlimit_memlock,
-                                &reexecute,
-                                &retval,
-                                &shutdown_verb,
-                                &fds,
-                                &switch_root_dir,
-                                &switch_root_init,
-                                &error_message);
+        r = invoke_main_loop(m,
```
poettering, line 3025: so up-top you changed invoke_main_loop() to return ManagerObjective, and now you store the result in an `int` anyway?

I'd be more comfortable with returning *any* error numbers as ManagerObjective, if the enum had a line `MANAGER_OBJECTIVE_MIN = -ERRNO_MAX` or so, to underline that not just EINVAL but in fact *any* errno can be stored in it.

poettering: And I'd probably introduce a local variable "objective" here too and not use `r` for this anymore. Otherwise it just seems too likely that we'll accidentally override `r` with some other failure but this failure is too important for that, I think it deserves its own properly named local variable.

src/core/main.c
```diff
@@ -3076,7 +3077,7 @@ int main(int argc, char *argv[]) {
         __lsan_do_leak_check();
 #endif
 
-        if (shutdown_verb) {
+        if (r >= 0 && shutdown_verb) {
```
poettering, line 3090: maybe add an explanatory comment here, what this means

poettering:
looks conceptually good to me, just some nits on implementatio details

src/core/main.c
keszybz: Here `r` is used throughout the function to store the state (with `goto finish`). So if this variable was renamed, we wouldn't have `r` but only this new variable. I think this would be almost the same, but we'd have a longer and unusual variable name.

src/core/main.c
keszybz: With the changes done in patch 1/2, if we return an error, we don't try to shut down, so those parameters are never used. In other words, we follow the usual pattern of "set output variables only on success".

src/core/main.c
keszybz: `retval` is initalized to `EXIT_FAILURE`. If `r < 0`, we never set it to anything else, so we should end up either freezing (for PID1) or returning `EXIT_FAILURE`.

keszybz:
> Do we have any tests for this functionality?

We obviously test the success paths all the time. But not the failure paths. And it's not so easy to add it. We'd need to set up some scaffolding to fail reexecution…

--

Updated with the suggested changes. I made functions return `int`, not ManagerObjective again, this way we avoid any issues with range.

src/core/main.c
```diff
@@ -1938,17 +1946,16 @@ static void do_reexecute(
...
                 args[0] = "/bin/sh";
                 args[1] = NULL;
                 (void) execve(args[0], (char* const*) args, saved_env);
-                log_error_errno(errno, "Failed to execute /bin/sh, giving up: %m");
+                return log_error_errno(errno, "Failed to execute /bin/sh, giving up: %m");
         } else
-                log_warning_errno(r, "Failed to execute /sbin/init, giving up: %m");
-
-        *ret_error_message = "Failed to execute fallback shell";
+                return log_warning_errno(r, "Failed to execute /sbin/init, giving up: %m");
```
poettering: hmm, i figure all error paths out of this functio should log at the same log level, and that's probably LOG_ERR. i.e. maybe upgrade the level here?

poettering:
lgtm, just some minor nitpick on log levels, see above.

src/core/main.c
keszybz: I adjusted this and added ", ignoring" in a few places.

keszybz:
Thank you for the review. I upgraded the label.

mrc0mmand:
TEST-36-NUMAPOLICY seems unhappy and it might be related?

```
Feb 15 16:58:12 H testsuite-36.sh[248]: PID1 NUMAPolicy support - Bind policy w/o mask
Feb 15 16:58:12 H testsuite-36.sh[248]: + writePID1NUMAPolicy bind
Feb 15 16:58:12 H testsuite-36.sh[248]: + cat
Feb 15 16:58:12 H testsuite-36.sh[248]: + pid1ReloadWithJournal
Feb 15 16:58:12 H testsuite-36.sh[248]: + startJournalctl
Feb 15 16:58:12 H testsuite-36.sh[248]: + :
Feb 15 16:58:12 H testsuite-36.sh[248]: + journalctl --no-pager --cursor-file=jounalCursorFile -n0 -ocat
Feb 15 16:58:12 H testsuite-36.sh[248]: + systemctl daemon-reload
Feb 15 16:58:12 H systemd[1]: Bus private-bus-connection: changing state UNSET → OPENING
Feb 15 16:58:12 H systemd[1]: sd-bus: starting bus private-bus-connection on fds 25/25 (socket:[15002], socket:[15002])...
Feb 15 16:58:12 H systemd[1]: Bus private-bus-connection: changing state OPENING → AUTHENTICATING
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1 iface=org.freedesktop.systemd1.Manager
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/job iface=org.freedesktop.systemd1.Job
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Unit
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Automount
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Device
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Mount
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Path
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Scope
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Service
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Slice
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Socket
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Swap
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Target
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Timer
Feb 15 16:58:12 H systemd[1]: Registering bus object implementation for path=/org/freedesktop/LogControl1 iface=org.freedesktop.LogControl1
Feb 15 16:58:12 H systemd[1]: Accepted new private connection.
Feb 15 16:58:12 H systemd[1]: Bus private-bus-connection: changing state AUTHENTICATING → RUNNING
Feb 15 16:58:12 H systemd[1]: Got message type=method_call sender=n/a destination=org.freedesktop.systemd1 path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=Reload cookie=1 reply_cookie=0 signature=n/a error>
Feb 15 16:58:12 H systemd[1]: Reloading.
Feb 15 16:58:12 H systemd[1]: Setting NUMA policy to bind, with nodes .
Feb 15 16:58:12 H systemd[1]: Failed to set NUMA memory policy, ignoring: Invalid argument
```

PR #22274, opened 2022-01-27
resolve: cleanups for on_stream_io()

### LINKED ISSUE #22266: resolve: llmnr: why dns_stream_unref() called in on_llmnr_stream_packet()??
LLMNR stream does not have `complete`  callback, that means the stream is unref()ed when `dns_stream_complete()` is called.

However, unlike the other `on_packet` callbacks, `on_llmnr_stream_packet()` unref() the stream.

In `on_stream_io()`, `on_packet` is called and then `dns_stream_complete()` is called. Does this mean the stream is doubly unref()ed?

Also, I cannot understand the comment added by 31f2a5e2b8c72b95c5fd965c6768755d5cba657f.

yuwata:
Based on the analysis by @joanbm.
See https://github.com/systemd/systemd/pull/22132#discussion_r793951650.

Closes #22266.

### DIFF
--- a/src/resolve/resolved-dns-stream.c
+++ b/src/resolve/resolved-dns-stream.c
@@ -446,17 +446,25 @@ static int on_stream_io_impl(DnsStream *s, uint32_t revents) {
                                 r = dns_stream_update_io(s);
                                 if (r < 0)
                                         return dns_stream_complete(s, -r);
+
+                                s->packet_received = true;
                         }
                 }
         }
 
-        /* Call "complete" callback if finished reading and writing one packet, and there's nothing else left
-         * to write. */
-        if (s->type == DNS_STREAM_LLMNR_SEND &&
-            (s->write_packet && s->n_written >= sizeof(s->write_size) + s->write_packet->size) &&
-            ordered_set_isempty(s->write_queue) &&
-            (s->read_packet && s->n_read >= sizeof(s->read_size) + s->read_packet->size))
-                return dns_stream_complete(s, 0);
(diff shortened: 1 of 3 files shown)


yuwata:
cc @joanbm. PTAL. I hope it explains what you observed.

joanbm:
Certainly the comment helps, but can't we remove the pitfall altogether?

Since everyone is actually using the `on_packet` callback (and even if they weren't, it'd be trivial to convert to it), we could just remove `dns_stream_take_read_packet` altogether, and instead pass the received packet as a new parameter to the `on_packet` callback, consuming it in the process.

joanbm:
As a side note, I haven't tested it and I may be missing something obvious... but since all users of DnsStream register an `on_packet` callback and immediately call

yuwata:
@joanbm Thank you for your help. I hope all the points are addressed.

joanbm:
Looks good to me.

joanbm:
Since once we have a full packet we now immediately call `on_packet` and start reading the next one, those two `s->read_packet || s->n_read < sizeof(s->read_size) + s->read_packet->size` conditions are always true now, and can be removed:

https://github.com/yuwata/systemd/blob/a5e2a488e83fabf6d8ade7621c2fc3574a8faaa7/src/resolve/resolved-dns-stream.c#L46
https://github.com/yuwata/systemd/blob/a5e2a488e83fabf6d8ade7621c2fc3574a8faaa7/src/resolve/resolved-dns-stream.c#L368

yuwata:
These can be dropped, but I'd like to keep them. They are safeguard.

bluca (approved):

PR #22277, opened 2022-01-27
test-network: fixes for test_activation_policy

### LINKED ISSUE #22267: test-network: test_activation_policy is racy
```python
        start_networkd()

        always = test.startswith('always')
        if test == 'manual':
            initial_up = 'UP' in check_output('ip link show test1')    # <--- This is racy
        else:
            initial_up = not test.endswith('down') # note: default is up
        expect_up = initial_up
        next_up = not expect_up
```
See https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/9726/artifact//systemd-centos-ci/index.html

yuwata:
Fixes #22267.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -3101,6 +3101,8 @@ class NetworkdNetworkTests(unittest.TestCase, Utilities):
                 check_output('ip link set dev test1 down')
             expect_up = initial_up if always else next_up
             next_up = not next_up
+            if always:
+                time.sleep(1)
 
     def test_activation_policy(self):
         for test in ['up', 'always-up', 'manual', 'always-down', 'down', '']:


test/test-network/systemd-networkd-tests.py
```diff
@@ -3101,6 +3101,8 @@ def _test_activation_policy(self, test):
                 check_output('ip link set dev test1 down')
             expect_up = initial_up if always else next_up
             next_up = not next_up
+            if always:
+                time.sleep(1)
```
anitazha, line 3105: Just wondering if there is there a more reliable way to tell if the state propagates to networkd? Sometimes the hardcoded sleeps aren't enough on a slow CI.

anitazha (approved):
LGTM one question but not a blocker

test/test-network/systemd-networkd-tests.py
yuwata: Unfortunately, I do not have any good idea... Potentially, the following should work: invoke `journalctl --follow` before `ip link set`, and monitor the log until it contains `"ActivationPolicy is \"always-on\", forcing link up."`. But that sounds quite overkill, and it is had for me as I am not good at writing Python...

I'd like to keep this as is until the hardcoded sleep causes CI failures many times.

yuwata:
Let's merge this, as @anitazha provides positive review, and find a better way later if the hadcoded sleep does not work fine.

PR #22278, opened 2022-01-27
hwdb: add two Elecom trackballs

rhendric:
Some things I'm not 100% sure I'm doing correctly:

* The Relacon is a handheld trackball. I assume that doesn't matter, for purposes of assigning `ID_INPUT_TRACKBALL`?
* These `MOUSE_DPI` values are pulled from documentation, not measured.
* I marked the first of the three DPI settings as default because that's what these trackballs are set to when first powered on. When the Relacon is powered off, it resets back to 500. The DEFT Pro seems to remember the setting between power cycles, but it was 500 initially. I notice that the nearby entry for the HUGE has the middle DPI setting marked as default, and I don't know if that's because the HUGE works differently (I doubt it?) or if there's some other policy at play.

### DIFF
--- a/hwdb.d/70-mouse.hwdb
+++ b/hwdb.d/70-mouse.hwdb
@@ -198,6 +198,17 @@ mouse:usb:v0461p4d46:name:USB Optical Mouse:*
 mouse:usb:v056ep010d:name:ELECOM TrackBall Mouse HUGE TrackBall:*
  MOUSE_DPI=500@125 *1000@125 1500@125
 
+# Elecom DEFT Pro TrackBall (M-DPT1MR)
+mouse:usb:v056ep0131:name:ELECOM TrackBall Mouse DEFT Pro TrackBall Mouse:*
+ MOUSE_DPI=*500 1000 1500
+ MOUSE_WHEEL_CLICK_ANGLE=10
+
+# Elecom Relacon (M-RT1DR)
+mouse:usb:v056ep0155:name:ELECOM ELECOM Relacon:*
+ ID_INPUT_TRACKBALL=1
+ MOUSE_DPI=*500 1000 1500
+ MOUSE_WHEEL_CLICK_ANGLE=30
+
 ##########################################
 # Fujitsu Siemens
 ##########################################


rhendric:
Hmm. Are frequencies required? The documentation at the top of 70-mouse.hwdb says that a list of only resolutions is acceptable.

I can't find documented frequencies for these trackballs; am I supposed to measure them somehow?

yuwata:
cc @whot.

yuwata:
@rhendric I am not sure, how about the following? http://who-t.blogspot.com/2014/12/building-a-dpi-database-for-mice.html

rhendric:
From that blog post:

> Trackballs are a bit hard to measure like this, my suggestion is to check the manufacturer's website first for any resolution data.

And it's true. I tried (awkwardly) running a ruler over the trackball with `mouse-dpi-tool` and the manufacturer-provided DPI values look ballpark correct. The frequencies were pretty variable, though. On the Relacon, I think the tool told me at one point that the mean was so far from the estimate that I needed to manually verify, whatever that entails. So I figured I'd leave those values out—I don't even know what they're used for elsewhere in the desktop stack.

But I'm happy to run more tests or to entirely drop the DPI values from these entries, whatever you all think is best.

whot:
> Hmm. Are frequencies required? The documentation at the top of 70-mouse.hwdb says that a list of only resolutions is acceptable.

They're not required, nothing uses them but when we started with the database it was easier to just include the frequencies on every entry. Ignoring a value is easier than than having to go back and adding them to every existing entry. So they're recommended but not required. The `mouse-dpi-tool` will give you an estimate for the frequency as you probably noticed.

> I don't even know what they're used for elsewhere in the desktop stack.

libinput uses the value to adjust pointer acceleration with the goal of making mice feel the same regardless of the resolution (i.e. granularity changes but not speed). That applies to devices >1000dpi, devices <1000 have a custom acceleration method that adjusts for the lower granularity.

`ID_INPUT_TRACKBALL` is primarily used for features that are unique to trackballs but there's no fixed specification what those are. libinput for example exposes a rotation setting on trackballs (you wouldn't use a mouse upside-down but for trackballs it's possible).

For devices with multiple (fixed) DPIs it's again good to have them but nothing uses them - we can't usually query the device for them anyway so everything just works with the default DPI. But for this glorious day in the future where we *can* do something, then we'll already have all the database entries.

rhendric:
I could use some slightly more concrete direction for this PR, please. 😅

Should I:
* remove the DPI entry?
* add what `mouse-dpi-tool` estimates as the frequency to the DPI entry?
* do the previous but round it to some nicer number than 142?
* hack the Python test that is failing because the DPI entries don't have frequency values?

If the DPI entry stays, should I:
* keep the default asterisk where it is, corresponding to the hardware default, but apparently suggesting to libinput that this trackball should use the custom acceleration method regardless of the hardware's DPI setting?
* move the default asterisk to the middle DPI, contradicting the directions given by the comment at the top of the file but making it more likely that libinput's heuristics match how the hardware is used (assuming more people use one of the two higher DPI settings than the lowest one)?

yuwata:
Fix for the parser is waiting in #22284.

yuwata:
#22284 is approved. I rebased this and force-pushed. Let's wait for the CIs are happy.

yuwata (approved):

whot (approved):

PR #22214, opened 2022-01-22
bpf: name unnamed bpf programs

jkartseva:
bpf-firewall and bpf-devices do not have names. This complicates
debugging with bpftool(8).

Assign names starting with 'sd_' prefix:
* firewall program names are 'sd_fw_ingress' for ingress attach
point and 'sd_fw_egress' for egress.
* 'sd_devices' for devices prog

'sd_' prefix is already used in source-compiled programs, e.g.
sd_restrictif_i, sd_restrictif_e, sd_bind6.

The name must not be longer than 15 characters or BPF_OBJ_NAME_LEN - 1.

Assign names only to programs loaded to kernel by systemd since
programs pinned to bpffs are already loaded.

PR #22279, opened 2022-01-28
bpf: load firewall with name only if supported

jkartseva:
BPF firewall is supported starting from v4.9 kernel where
BPF_PROG_TYPE_SOCKET_FILTER support was added [0].

However, program name support was added to v4.15 [1] and BPF_PROG_LOAD
syscall will fail on older kernels if called with prog_name attribute.
BPF_F_ALLOW_MULTI was also added to v4.15 kernel which allows reusing
BPF_F_ALLOW_MULTI probe to indicate that program name is also supported.

It is no problem for BPF_PROG_TYPE_CGROUP_DEVICE since it was added in
v4.15.

[0] https://elixir.bootlin.com/linux/v4.9/source/include/uapi/linux/bpf.h#L92
[1] https://elixir.bootlin.com/linux/v4.15/source/include/uapi/linux/bpf.h#L191

Follow-up of https://github.com/systemd/systemd/pull/22214

### DIFF
--- a/src/core/bpf-devices.c
+++ b/src/core/bpf-devices.c
@@ -306,7 +306,7 @@ int bpf_devices_supported(void) {
                 return supported = 0;
         }
 
-        r = bpf_program_new(BPF_PROG_TYPE_CGROUP_DEVICE, NULL, &program);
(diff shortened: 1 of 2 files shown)


yuwata (approved):
LGTM.

PR #22285, opened 2022-01-28
sd-dhcp-server: refuse too large packet to send

yuwata:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44134.


yuwata:
cc @evverx

bluca (approved):

evverx:
@yuwata thanks! I agree packets like this should be rejected but I'm not sure they should even reach that point. I wonder if it would be better to reject them earlier somewhere?

yuwata:
The function is called in various path. I think the current place of the check is the simplest and safest.

evverx:
Fair enough. Thanks!

PR #21807, opened 2021-12-17
Use KERNEL_INSTALL_MACHINE_ID and KERNEL_INSTALL_LAYOUT with bootctl install

keszybz:
/cc @nabijaczleweli

PR #22290, opened 2022-01-28
boot: Don't require a machine ID to be available

### LINKED ISSUE #22224: `bootctl install` modifies currently running system's state, won't work without a machine ID
Using systemd v250 on carbonOS 2022.1

It seems like `bootctl install` now assumes that it will only ever be run inside of the system it will be booting. This assumption is wrong in many cases: it is only right if the system is being upgraded from a different bootloader to sd-boot. In most (all?) other cases, `bootctl install` will be invoked from within a _different_ copy of the OS than the one it will be booting. Here's two use cases that will elaborate exactly what I mean by this:

### The OS image builder case

When building an OS image, I use a transient container with the right packages staged into it to generate the ISO image. When `bootctl install` gets invoked, the userspace it's running in was created 30 seconds ago and will be destroyed as soon as the ISO image is created and stored. It doesn't have an /etc/machine-id because it isn't a machine, and creating the file is undesirable (it introduces randomness into the build, which is bad for the purposes of reproducible builds)

Upgrading to systemd v250 broke my ISO image builder. Basically, `bootctl install` now requires that /etc/machine-id exists (regression from #16687) even if I pass `--make-machine-id-directory=no`. Additionally, if I do create the /etc/machine-id file to work around this issue, `bootctl install` will write to `/etc/machine-info` in my build environment! This new file just contains `KERNEL_INSTALL_LAYOUT=bls`. This is definitely unexpected and unwanted. This created file is useless because it will immediately get destroyed once the ISO is generated

### The LiveOS or other OS case

I distribute a LiveOS image of my OS with a installer tool baked into it. At install time, the installer tool invokes `bootctl install` to install the bootloader, set up EFI variables, etc. This also has the unintended side-effect of writing /etc/machine-info into the LiveOS's filesystem. This file will never make it into the installed system, and really shouldn't be created. Luckily, this doesn't have any bad side-effects for the user because any changes to the LiveOS are thrown out on shutdown

Here's a situation where this can backfire for the user: Let's say the user has two machines. A botched bootloader upgrade on machine A caused it to stop booting, so they removed the SSD and connected it to machine B. In machine B, they mount the SSD's ESP to /mnt and delete all of the systemd-boot-related binaries (executables, random-seed, etc; not config). Then, they run `bootctl install --no-variables --esp-path=/mnt`. This will do two unintended things:
- Put machine B's machine-id into machine A's new bootloader config. This doesn't happen in the LiveOS because systemd (correctly) detects that /etc is on a temporary fs. This will probably mess up machine A's boot config and require some manual fixup
- Modify machine B's /etc/machine-info. This could reconfigure machine B's kernel-install, and the next time kernel-install is invoked, it'll misconfigure machine B's bootloader

### Extra note

I build systemd with kernel-install disabled in meson (since kernel-install would break my distro's configuration if used). bootctl should probably avoid doing anything specific for kernel-install if kernel-install cannot and will not ever be used on the system.

daandemeyer:
Regression introduced in https://github.com/systemd/systemd/pull/21807.

Fixes #22224

### DIFF
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -141,7 +141,7 @@ static int load_install_machine_id_and_layout(void) {
 
         if (isempty(s)) {
                 r = sd_id128_get_machine(&arg_machine_id);
-                if (r < 0)
+                if (r < 0 && !IN_SET(r, -ENOENT, -ENOMEDIUM))
                         return log_error_errno(r, "Failed to get machine-id: %m");
         } else {
                 r = sd_id128_from_string(s, &arg_machine_id);
@@ -170,7 +170,7 @@ static int settle_install_machine_id(void) {
 
         bool layout_non_bls = arg_install_layout && !streq(arg_install_layout, "bls");
         if (arg_make_machine_id_directory < 0) {
-                if (layout_non_bls)
+                if (layout_non_bls || sd_id128_is_null(arg_machine_id))
                         arg_make_machine_id_directory = 0;
                 else {
                         r = path_is_temporary_fs("/etc/machine-id");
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
cc @keszybz

AdrianVovk:
I tested it. Yep, seems to work without a machine-id present

bluca (approved):

PR #22291, opened 2022-01-28
shared: Handle filesystems that don't support hole punching in COPY_HOLES

daandemeyer:


bluca (approved):

src/shared/copy.c
```diff
@@ -119,9 +119,11 @@ static int create_hole(int fd, off_t size) {
         if (end < 0)
                 return -errno;
 
-        /* If we're not at the end of the target file, punch a hole in the existing space using fallocate(). */
+        /* If we're not at the end of the target file, try to punch a hole in the existing space using fallocate(). */
 
-        if (offset < end && fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, MIN(size, end - offset)) < 0)
+        if (offset < end &&
+            fallocate(fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, MIN(size, end - offset)) < 0 &&
+            !ERRNO_IS_NOT_SUPPORTED(errno))
```
yuwata, line 126: Hm? If `ERRNO_IS_NOT_SUPPORTED()` is true, then the `lseek()` below fails, no?

daandemeyer: Why would it fail? We're only trying to punch a hole in the existing space here. If it fails, the hole simply isn't created and we continue on as usual.

yuwata: Sorry, please forget the comment. I misunderstand the conditions below.

yuwata (approved):

PR #22295, opened 2022-01-29
ci: use CFLite to test forks (including systemd-stable)

evverx:
It's like CIFuzz but unlike CIFuzz it's compatible with forks and
it should make it possible to run the fuzzers to make sure that
patches backported to them are backported correctly without introducing
new bugs and regressions.

@mrc0mmand could you take a look? I'm going to unleash it on my fork to make sure it works but in the meantime I'm not sure how to turn it off here properly. It should include something like
```
if: github.repository != 'systemd/systemd'
```
to avoid clogging up the GHActions pool here but at the same time it should be run here when Dependabot updates the actions or when `tools/oss-fuzz.sh` is changed.

### DIFF
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -1,0 +1,4 @@
+FROM gcr.io/oss-fuzz-base/base-builder@sha256:14b332de0e18683f37386eaedbf735bc6e8d81f9c0e1138d620f2178e20cd30a
+COPY . $SRC/systemd
+WORKDIR $SRC/systemd
+COPY tools/oss-fuzz.sh $SRC/build.sh
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,8 @@ updates:
     schedule:
       interval: "monthly"
     open-pull-requests-limit: 2
+  - package-ecosystem: "docker"
+    directory: "/.clusterfuzzlite"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 2
--- a/.github/workflows/cflite_build.yml
+++ b/.github/workflows/cflite_build.yml
@@ -1,0 +1,31 @@
+---
+# vi: ts=2 sw=2 et:
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+name: ClusterFuzzLite continuous builds
+on:
+  push:
+    branches:
+      - main
+      - v[0-9]+-stable
+
+permissions: read-all
+
+jobs:
+  Build:
+    runs-on: ubuntu-latest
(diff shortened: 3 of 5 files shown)


mrc0mmand:
> It's like CIFuzz but unlike CIFuzz it's compatible with forks and it should make it possible to run the fuzzers to make sure that patches backported to them are backported correctly without introducing new bugs and regressions.
> 
> @mrc0mmand could you take a look? I'm going to unleash it on my fork to make sure it works but in the meantime I'm not sure how to turn it off here properly. It should include something like
> 
> ```
> if: github.repository != 'systemd/systemd'
> ```
> 
> to avoid clogging up the GHActions pool here but at the same time it should be run here when Dependabot updates the actions or when `tools/oss-fuzz.sh` is changed.

Maybe:

```yml
on:
  push:
    branches:
      - main
      - v[0-9]+-stable
    pull_request:
      paths:
        - tools/oss-fuzz.sh
```

mrc0mmand:
Ah, I see what you mean. I guess in addition to the above, it something like:

```
if: github.repository != 'systemd/systemd' || (github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot[bot]')
```

which is not very elegant, but should work. Another option would be getting the list of changes from the `pull_request` event and checking if `tools/oss-fuzz.sh` is there. Not sure how to do that (yet), unfortunately).

evverx:
Thanks!

> Another option would be getting the list of changes from the pull_request event and checking if tools/oss-fuzz.sh is there. Not sure how to do that (yet), unfortunately).

I'll take a look at how the labeler action does that. As far as I know it gets the list of files changed in PRs and compare them with its config. Then again maybe I'm just overthinking it and it should be fine to just skip it here with
```
if: github.repository != 'systemd/systemd' || (github.event_name == 'pull_request' && github.event.pull_request.user.login == 'dependabot[bot]')
```

evverx:
I'm not sure what's up with Packit :-)

I think it's almost ready. I'm trying to figure out whether Dependabot can really update docker images. Once it's done I'll change `dependabot.yml` here as well.

evverx:
@mrc0mmand I think it should be ready for review. I tested it and judging by https://github.com/evverx/systemd/actions/runs/1766409686 (where it was triggered on a PR) and https://github.com/evverx/systemd/actions/runs/1766402069 (where it was triggered on a commit) it seems to be working.

mrc0mmand (approved):
That's nifty, thanks!

mrc0mmand:
Not sure if it's ready to be merged (given the red label), so feel free to merge it once it is.

mrc0mmand:
image

Understood :-)

PR #22302, opened 2022-01-29
ci: merge seed corpora with public OSS-Fuzz corpora on CFLite

evverx:
to let it cover as much code as possible. It should help to catch most regressions in a few minutes.


evverx:
Judging by https://github.com/systemd/systemd/actions/runs/1767066906 it seems to be working. For example the fuzz-varlink seed corpus was merged with the OSS-Fuzz corpus and the fuzzer used 5088 files (instead of 10)
```
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 1337
INFO: Loaded 2 modules   (74432 inline 8-bit counters): 74344 [0x7f3e542ed4d0, 0x7f3e542ff738), 88 [0x5bea28, 0x5bea80), 
INFO: Loaded 2 PC tables (74432 PCs): 74344 [0x7f3e542ff738,0x7f3e54421db8), 88 [0x57d028,0x57d5a8), 
INFO:     5088 files found in /github/workspace/cifuzz-corpus/fuzz-varlink
```

evverx:
It's blocked by https://github.com/google/clusterfuzzlite/issues/84 :-(

evverx:
@mrc0mmand to get it around I decided to no longer upload artifacts containing the latest build

    ci: no longer upload the latest builds on commits

    The idea behind this action is to make it possible to compare the
    latest fuzz targets with PRs to figure out whether bugs are really
    reproducible in PRs only. Since forks (including systemd-stable) are
    usually based on the upstream repository where almost all the bugs
    are fixed before releases are cut it should be safe to assume that
    if CFLite finds bugs in PRs they are most likely introduced in those
    PRs.

    It should probably be brought back once https://github.com/google/clusterfuzzlite/issues/84
    is fixed.

The action is being tested in https://github.com/evverx/systemd/runs/4997912999?check_suite_focus=true and once it's green it should be good to go I think.

.clusterfuzzlite/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM gcr.io/oss-fuzz-base/base-builder@sha256:14b332de0e18683f37386eaedbf735bc6e8d81f9c0e1138d620f2178e20cd30a
+ENV MERGE_WITH_OSS_FUZZ_CORPORA yes
```
mrc0mmand: Looks like this `ENV` syntax is deprecated and might be removed in the future (https://docs.docker.com/engine/reference/builder/#env "Alternative syntax"). Not sure if it matters here though :-)

mrc0mmand (approved):

mrc0mmand:
Looks like CI is happy with this change (given the job you've posted), so LGTM.

.clusterfuzzlite/Dockerfile
evverx: Thanks! I'll change that

evverx: Done. It's being tested https://github.com/evverx/systemd/runs/4998101837?check_suite_focus=true

mrc0mmand:
Still LGTM, thanks! Let's wait for the job to finish.

PR #22304, opened 2022-01-30
core: check if argc > 0 and argv[0] is set

yuwata:
Follow-up for 1637e757070a61b12b26a2a4ef547330562dc77f.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -2726,6 +2726,8 @@ int main(int argc, char *argv[]) {
         Manager *m = NULL;
         FDSet *fds = NULL;
 
+        assert(argc > 0 && !isempty(argv[0]));
+
         /* SysV compatibility: redirect init → telinit */
         redirect_telinit(argc, argv);
 


mrc0mmand (approved):

mrc0mmand:
focal-amd64 failed due to:
```
[  195.656398] systemd[860]: Failed to create destination mount point node '/run/systemd/unit-extensions/1': File exists
[  195.656825] systemd[860]: Failed to mount /tmp/app1 to /run/systemd/unit-extensions/1: No such file or directory
```

which is already being resolved in https://github.com/systemd/systemd/pull/22293.

src/core/main.c
```diff
@@ -2726,6 +2726,8 @@ int main(int argc, char *argv[]) {
         Manager *m = NULL;
         FDSet *fds = NULL;
 
+        assert(argc > 0 && !isempty(argv[0]));
```
cgzones, line 2729: Why only an assert and not an unconditional runtime check?
Shouldn't the subsequent code either handle an argc value of 0 just fine or not be called under any circumstance, since the calling method is out of scope.

bluca: that's what the assert does?

evverx: `assert` can be compiled out if NDEBUG is defined so it depends on how `systemd` is compiled. My guess would be that @cgzones hinted at `assert_se` or something like that.

PR #22306, opened 2022-01-30
core: check argc/argv uncoditionally

mrc0mmand:
as `assert()` might be dropped with `-DNDEBUG`.

Follow-up to cf3095a and 1637e75.

---

Prompted by https://github.com/systemd/systemd/pull/22304#discussion_r795181312.

Also, I noticed we already do pretty much the same checks in the `parse_argv()` functions in respective tools, e.g.:

https://github.com/systemd/systemd/blob/cf3095ac2bbdba3db0d8f7fa7afcee8b8c136201/src/journal/journalctl.c#L530-L533

However, not all tools use the `DEFINE_MAIN_FUNCTION*()` macros, so in some cases we might end up doing the same checks twice (once in `main()` if they use the macro, and then in `parse_argv()`) - 1) not sure how much of an issue that actually is, 2) should we unify the checks somehow?, and 3) should they use the `assert_se()` macro as well instead of plain `assert()`?

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -2726,7 +2726,7 @@ int main(int argc, char *argv[]) {
         Manager *m = NULL;
         FDSet *fds = NULL;
 
-        assert(argc > 0 && !isempty(argv[0]));
+        assert_se(argc > 0 && !isempty(argv[0]));
 
         /* SysV compatibility: redirect init → telinit */
         redirect_telinit(argc, argv);
--- a/src/shared/main-func.h
+++ b/src/shared/main-func.h
@@ -15,7 +15,7 @@
 #define _DEFINE_MAIN_FUNCTION(intro, impl, ret)                         \
         int main(int argc, char *argv[]) {                              \
                 int r;                                                  \
-                assert(argc > 0 && !isempty(argv[0]));                  \
+                assert_se(argc > 0 && !isempty(argv[0]));               \
                 save_argc_argv(argc, argv);                             \
                 intro;                                                  \
                 r = impl;                                               \


yuwata:
If possible, let's also use `DEFINE_MAIN_FUNCTION()` in journalctl. But, that's ok in later PR.

> 1) not sure how much of an issue that actually is, 2) should we unify the checks somehow?, 

And, redundant assertion does not cause any issues. Let's keep them as is.

> 3) should they use the `assert_se()` macro as well instead of plain `assert()`?

In DEFINE_MAIN_FUNCTION, I think yes. It checks how users call commands, not our coding issue.

yuwata (approved):
LGTM.

cgzones:
If I understand the implementation correctly `assert_se()` only avoids side affects being skipped in the NDEBUG configuration, e.g. `assert(i++);`. I was more thinking about `assert_return`, similar to the public C API.

mrc0mmand:
> If I understand the implementation correctly `assert_se()` only avoids side affects being skipped in the NDEBUG configuration, e.g. `assert(i++);`. I was more thinking about `assert_return`, similar to the public C API.

In addition to that, `assert_se()` always calls `log_assert_failed()`, which in turn calls `abort()` after printing an error message.

cgzones:
> In addition to that, `assert_se()` always calls `log_assert_failed()`, which in turn calls `abort()` after printing an error message.

My bad, I did not look at `log_assert_failed()`.

yuwata:
@mrc0mmand Is it ok to merge this as is? At least, this should make nothing bad.

mrc0mmand:
> @mrc0mmand Is it ok to merge this as is? At least, this should make nothing bad.

Yup, this should be ready to go. The other raised points can be dealt with later.

poettering:
btw, I'd usually do asserts like this in two lines:

```c
assert_se(argc > 0);
assert_se(!isempty(argv[0]));
```

Why? because that way, if the assertion actually fails, you'll immediately see why in the generated log message. If you instead use && to combine two checks within a single assert_se(), then you have to guess which of the two failed...

but doesn't matter at all...

PR #22307, opened 2022-01-30
hwdb: add new database file for PDA devices

davide125:
Add a new database for handhelds (PDAs, calculators, etc.) that should be
accessible the seat owner.

The database is initially populated with Texas Instruments calculators
and linking cables, which removes the need to installing dedicated udev
rules for them.


yuwata (approved):
LGTM.

PR #22314, opened 2022-01-31
meson: set minimum clang/llvm versions for bpf support

jameshilliard:
The minimum clang/llvm-strip version needed for building libbpf based programs is 10.0.0, this is documented here:
https://github.com/libbpf/libbpf/tree/v0.6.1#bpf-co-re-compile-once--run-everywhere

Using an older version such as 9.0.0 will generate objects that bpftool will not be able to generate skeletons for.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -988,7 +988,7 @@ else
         # Support 'versioned' clang/llvm-strip binaries, as seen on Debian/Ubuntu
         # (like clang-10/llvm-strip-10)
         if meson.is_cross_build() or cc.get_id() != 'clang' or cc.cmd_array()[0].contains('afl-clang') or cc.cmd_array()[0].contains('hfuzz-clang')
-                r = find_program('clang', required : bpf_framework_required)
+                r = find_program('clang', required : bpf_framework_required, version : '>= 10.0.0')
                 clang_found = r.found()
                 if clang_found
                         if meson.version().version_compare('>= 0.55')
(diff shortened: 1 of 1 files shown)


yuwata (approved):

meson.build
```diff
@@ -988,7 +988,7 @@ else
         # Support 'versioned' clang/llvm-strip binaries, as seen on Debian/Ubuntu
         # (like clang-10/llvm-strip-10)
         if meson.is_cross_build() or cc.get_id() != 'clang' or cc.cmd_array()[0].contains('afl-clang') or cc.cmd_array()[0].contains('hfuzz-clang')
-                r = find_program('clang', required : bpf_framework_required)
+                r = find_program('clang', required : bpf_framework_required, version : '>= 10.0.0')
```
yuwata, line 991: Maybe, simply `>= 10` is ok?

jameshilliard: Not sure if that's interpreted the same way for comparisons, this is the output that meson parser for the version which is in the `10.0.0` format:
```
$ clang --version
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
```
$ llvm-strip --version
llvm-strip, compatible with GNU strip
LLVM (http://llvm.org/):
  LLVM version 10.0.0
  
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: znver1
```

jameshilliard: I think this version check shares logic with [`version_compare`](https://mesonbuild.com/Syntax.html#version_compare) where it states in the docs:
> It is best to be unambiguous and specify the full revision level to compare.

yuwata: Thank you!

jameshilliard: I did a test and set the version requires to `==10` and meson does error indicating that a version of `10` is interpreted differently than a version of `10.0.0`(which does not fail in a `==10.0.0` test):
```
meson.build:991:16: ERROR: Invalid version of program, need 'clang' ['== 10'] found '10.0.0'.
```

yuwata: Thank you so much!

PR #22315, opened 2022-01-31
network: add ModemManager support

### LINKED ISSUE #20370: Integration with ModemManager
Setting up WWAN connection with systemd-networkd is the major hurdle at the moment.

Ability to specify a new .network or .netdev link type which would talk with ModemManager to establish and properly configure IP interface managed by systemd-networkd

Right now we have to resort to scripting and manual output parsing which is cumbersome and error-prone.

The typical WWAN connection bring-up looks as follows:
```
mmcli --list-modems
/org/freedesktop/ModemManager1/Modem/5 [QUALCOMM INCORPORATED] QUECTEL Mobile Broadband Module
# this gives the path and modem number e. g. 5:
mmcli --simple-connect="apn=test-internet.myapn" -m 5
successfully connected the modem
mmcli -m 5
...
Bearer                  paths: /org/freedesktop/ModemManager1/Bearer/3
# this (if succeed) gives the number of the established bearer e. g. 3:
mmcli -b 3

  General               path: /org/freedesktop/ModemManager1/Bearer/3
                             type: default  
  Status                 connected: yes
                             suspended: no
                             interface: wwan0
                             ip timeout: 20

  Properties           apn: test-internet.myapn
                            roaming: allowed

  IPv4 configuration       method: static
                                    address: 10.10.42.234
                                    prefix: 30
                                    gateway: 10.10.42.24
                                    dns: 8.8.8.8, 8.8.4.4
                                    mtu: 1500

  Statistics               duration: 60
                              attempts: 1
                              total-duration: 60
```
Now we have to parse the output of the mmcli (or use dbus - whatever) and get the IP address, Gateway and other parameters to generate wwan.network so it can be controlled with systemd-networkd.

Would be way more convenient if systemd-networkd could take care of low-level details and extract that information directly from ModemManager as well as gracefully handle network interruptions etc.

247

yuwata:
Closes #20370.

### DIFF
--- a/src/network/networkd-link.c
+++ b/src/network/networkd-link.c
@@ -1157,11 +1157,14 @@ static int link_configure(Link *link) {
 
 static int link_get_network(Link *link, Network **ret) {
         Network *network;
+        Bearer *b = NULL;
         int r;
 
         assert(link);
         assert(link->manager);
         assert(ret);
+
+        (void) link_get_bearer(link, &b);
 
         ORDERED_HASHMAP_FOREACH(network, link->manager->networks) {
                 bool warn = false;
@@ -1178,7 +1181,8 @@ static int link_get_network(Link *link, Network **ret) {
                                 link->alternative_names,
                                 link->wlan_iftype,
                                 link->ssid,
-                                &link->bssid);
+                                &link->bssid,
+                                b ? b->apn : NULL);
                 if (r < 0)
                         return r;
                 if (r == 0)
@@ -1209,7 +1213,7 @@ static int link_get_network(Link *link, Network **ret) {
         return -ENOENT;
 }
 
-static int link_reconfigure_impl(Link *link, bool force) {
+int link_reconfigure_impl(Link *link, bool force) {
         Network *network = NULL;
         NetDev *netdev = NULL;
         int r;
--- a/src/network/networkd-wwan.c
+++ b/src/network/networkd-wwan.c
@@ -419,6 +419,12 @@ int bearer_update_link(Bearer *b) {
         if (link_get_by_name(b->manager, b->name, &link) < 0)
                 return 0;
 
+        r = link_reconfigure_impl(link, /* force = */ false);
+        if (r < 0)
+                link_enter_failed(link);
+        if (r != 0) /* r > 0 means interface is reconfigured. */
+                return r;
+
         r = link_apply_bearer_impl(link, b);
         if (r < 0)
                 link_enter_failed(link);
@@ -430,6 +436,7 @@ void bearer_drop(Bearer *b) {
         assert(b);
 
         b->connected = false;
+        b->apn = mfree(b->apn);
 
         (void) bearer_update_link(b);
 
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -127,6 +127,15 @@
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>APN=</varname></term>
+        <listitem>
+          <para>A whitespace-separated list of shell-style globs matching the Access Point Name (APN)
+          of the current WWAN connection. If the list is prefixed with a "!", the test is inverted.
+          </para>
+        </listitem>
+      </varlistentry>
+
       <xi:include href="systemd.link.xml" xpointer="host" />
       <xi:include href="systemd.link.xml" xpointer="virtualization" />
       <xi:include href="systemd.link.xml" xpointer="kernel-command-line" />
--- a/src/network/networkd-link.h
+++ b/src/network/networkd-link.h
@@ -236,6 +236,7 @@ int link_stop_engines(Link *link, bool may_keep_dhcp);
 const char* link_state_to_string(LinkState s) _const_;
 LinkState link_state_from_string(const char *s) _pure_;
 
+int link_reconfigure_impl(Link *link, bool force);
 int link_reconfigure(Link *link, bool force);
 int link_reconfigure_after_sleep(Link *link);
 
--- a/src/network/networkd-network-gperf.gperf
+++ b/src/network/networkd-network-gperf.gperf
@@ -58,6 +58,7 @@ Match.SSID,                                  config_parse_match_strv,           
 Match.BSSID,                                 config_parse_ether_addrs,                                 0,                             offsetof(Network, match.bssid)
 Match.Name,                                  config_parse_match_ifnames,                               IFNAME_VALID_ALTERNATIVE,      offsetof(Network, match.ifname)
 Match.Property,                              config_parse_match_property,                              0,                             offsetof(Network, match.property)
+Match.APN,                                   config_parse_match_strv,                                  0,                             offsetof(Network, match.apn)
 Match.Host,                                  config_parse_net_condition,                               CONDITION_HOST,                offsetof(Network, conditions)
 Match.Virtualization,                        config_parse_net_condition,                               CONDITION_VIRTUALIZATION,      offsetof(Network, conditions)
 Match.KernelCommandLine,                     config_parse_net_condition,                               CONDITION_KERNEL_COMMAND_LINE, offsetof(Network, conditions)
--- a/src/shared/net-condition.c
+++ b/src/shared/net-condition.c
@@ -28,6 +28,7 @@ void net_match_clear(NetMatch *match) {
         match->wlan_iftype = strv_free(match->wlan_iftype);
         match->ssid = strv_free(match->ssid);
         match->bssid = set_free(match->bssid);
+        match->apn = strv_free(match->apn);
 }
 
 bool net_match_is_empty(const NetMatch *match) {
@@ -44,7 +45,8 @@ bool net_match_is_empty(const NetMatch *match) {
                 strv_isempty(match->property) &&
                 strv_isempty(match->wlan_iftype) &&
                 strv_isempty(match->ssid) &&
-                set_isempty(match->bssid);
+                set_isempty(match->bssid) &&
+                strv_isempty(match->apn);
 }
 
 static bool net_condition_test_strv(char * const *patterns, const char *string) {
@@ -129,7 +131,8 @@ int net_match_config(
                 char * const *alternative_names,
                 enum nl80211_iftype wlan_iftype,
                 const char *ssid,
-                const struct ether_addr *bssid) {
+                const struct ether_addr *bssid,
+                const char *apn) {
 
         _cleanup_free_ char *iftype_str = NULL;
         const char *path = NULL;
@@ -175,6 +178,9 @@ int net_match_config(
                 return false;
 
         if (match->bssid && (!bssid || !set_contains(match->bssid, bssid)))
+                return false;
+
+        if (!net_condition_test_strv(match->apn, apn))
                 return false;
 
         return true;
--- a/src/shared/net-condition.h
+++ b/src/shared/net-condition.h
@@ -22,6 +22,7 @@ typedef struct NetMatch {
         char **wlan_iftype;
         char **ssid;
         Set *bssid;
+        char **apn;
 } NetMatch;
 
 void net_match_clear(NetMatch *match);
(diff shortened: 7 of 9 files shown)


src/network/networkd-wwan.h
```diff
@@ -0,0 +1,55 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+#pragma once
+
+#include <stdbool.h>
+#include <inttypes.h>
```
bluca, line 4: alphabetic sort?

bluca:
needs a rebase

src/network/networkd-link.c
```diff
@@ -460,6 +461,9 @@ void link_check_ready(Link *link) {
         if (!link->sr_iov_configured)
                 return (void) log_link_debug(link, "%s(): SR-IOV is not configured.", __func__);
 
+        if (!link->bearer_configured)
+                return (void) log_link_debug(link, "%s(): Bearer is not applied.", __func__);
```
keszybz: "is not configured" or "has not been applied" ? The meaning of "is applied" is not clear at all in this context.

src/network/networkd-wwan.h
```diff
@@ -0,0 +1,55 @@
...
+        unsigned ip4_prefixlen;
+        unsigned ip6_prefixlen;
+        union in_addr_union ip4_address;
+        union in_addr_union ip6_address;
+        union in_addr_union ip4_gateway;
+        union in_addr_union ip6_gateway;
+        struct in_addr_data *dns;
+        size_t n_dns;
+        uint32_t ip4_mtu;
+        uint32_t ip6_mtu;
```
keszybz, line 37: ~~Shouldn't this part be a union?~~
I see that both can be used in parallel, so no.

src/network/networkd-wwan-bus.c
```diff
@@ -0,0 +1,362 @@
...
+#include "networkd-wwan.h"
+
+/* From ModemManager-enums.h */
+typedef enum {
+    MM_BEARER_IP_FAMILY_NONE    = 0,
+    MM_BEARER_IP_FAMILY_IPV4    = 1 << 0,
+    MM_BEARER_IP_FAMILY_IPV6    = 1 << 1,
+    MM_BEARER_IP_FAMILY_IPV4V6  = 1 << 2,
+    MM_BEARER_IP_FAMILY_ANY     = 0xFFFFFFFF
+} MMBearerIpFamily;
```
keszybz, line 18: Wouldn't it be better to just reuse `ADDRESS_FAMILY_*` for this? The mapping seems to be 1:1 anyway.

src/network/networkd-wwan-bus.c
```diff
@@ -0,0 +1,362 @@
...
+
+        r = bearer_initialize(b);
+        if (r < 0)
+                return r;
+
+        TAKE_PTR(b);
+        return 0;
+}
+
+static int on_path(const char *path, void *userdata) {
```
keszybz: `on_path` is not a very good name. And I think it's be nicer to do the cast separately:
```C
Set *set = ASSERT_PTR(userdata);
return set_put_strdup(set, path);
```

src/network/networkd-wwan.c
```diff
@@ -0,0 +1,437 @@
...
+                                return r;
+
+                        r = link_request_bearer_route(link, AF_INET, &b->ip4_gateway, &b->ip4_address);
+                        if (r < 0)
+                                return r;
+                }
+
+                if (b->connected && b->ip4_method == MM_BEARER_IP_METHOD_DHCP) {
+                        if (!link_dhcp4_enabled(link))
+                                log_link_notice(link, "The WWAN connection requested DHCPv4 client, but it is disabled.");
```
keszybz, line 309: Hmm, IIUC, we log notice but otherwise ignore our own config. I think this is not very useful: we should either let *our* config take priority here, and not call dhcp4_start(), or we should just ignore our own config. Ideally, we would default link->network->dhcp to ADDRESS_FAMILY_ANY if the support for ModemManager is configured. This would make things "just work", but still allow explicit configuration if wanted.

src/network/networkd-wwan.c
```diff
@@ -0,0 +1,437 @@
...
+                        r = link_request_bearer_route(link, AF_INET6, &b->ip6_gateway, NULL);
+                        if (r < 0)
+                                return r;
+                }
+
+                if (b->connected && b->ip6_method == MM_BEARER_IP_METHOD_DHCP) {
+                        if (!link_ipv6_accept_ra_enabled(link) && !link_dhcp6_enabled(link))
+                                log_link_notice(link,
+                                                "The WWAN connection requested IPv6 dynamic address configuration,"
+                                                "but both IPv6 Router Discovery and DHCPv6 client are disabled.");
```
keszybz, line 336: Same here.

keszybz (commented):
Looks reasonable. I'm a bit surprised that we need to do dbus introspection to figure out the path. @poettering, opinions on that?

src/network/networkd-wwan-bus.c
yuwata: No, ADDRESS_FAMILY_YES is 3, not 4.

src/network/networkd-wwan.c
yuwata: `dhcp4_start()` does nothing and returns 0 when DHCP4 client is disabled. So, the current implementation is equivalent to the first suggestion.

src/network/networkd-wwan.c
yuwata: Also here, `ndisc_start()` and `dhcp6_start()` do nothing if they are disabled.

src/network/networkd-wwan-bus.c
```diff
@@ -0,0 +1,364 @@
...
+
+        assert(message);
+
+        b->slot = sd_bus_slot_unref(b->slot);
+
+        e = sd_bus_message_get_error(message);
+        if (e) {
+                bool removed = false;
+
+                if (sd_bus_error_has_name(e, SD_BUS_ERROR_UNKNOWN_METHOD))
```
poettering, line 176: hmm, this sounds like a weird error to test against here, after all this is about a disappearing object, not method, right?

it appears to me it's a misleading error code returned by mman, no?

if so, maybe also check for SD_BUS_ERROR_UNKNOWN_OBJECT here, as preparation for a future where mman returns a more appropriate error?

src/network/networkd-wwan-bus.c
```diff
@@ -0,0 +1,364 @@
...
+        assert(manager);
+        assert(sd_bus_is_ready(manager->bus) > 0);
+
+        r = sd_bus_call_method_async(
+                        manager->bus,
+                        NULL,
+                        "org.freedesktop.ModemManager1",
+                        "/org/freedesktop/ModemManager/Bearer",
+                        "org.freedesktop.DBus.Introspectable",
+                        "Introspect",
```
poettering, line 306: hmm, doesn't mman implement ObjectManager? judging by https://www.freedesktop.org/software/ModemManager/api/1.0.0/ref-dbus-standard-interfaces-objectmanager.html it does.

I think that's a much better way to enumerate objects, if there's no explicit way to enumerate them.

src/network/networkd-wwan.c
```diff
@@ -0,0 +1,444 @@
...
+#include "networkd-route.h"
+#include "networkd-wwan.h"
+
+/* From ModemManager-enums.h */
+typedef enum {
+    MM_BEARER_IP_METHOD_UNKNOWN = 0,
+    MM_BEARER_IP_METHOD_PPP     = 1,
+    MM_BEARER_IP_METHOD_STATIC  = 2,
+    MM_BEARER_IP_METHOD_DHCP    = 3,
+} MMBearerIpMethod;
```
poettering, line 22: indentation off?

src/network/networkd-wwan.c
```diff
@@ -0,0 +1,444 @@
...
+
+        assert(link);
+        assert(IN_SET(family, AF_INET, AF_INET6));
+        assert(addr);
+
+        if (!in_addr_is_set(family, addr))
+                return 0;
+
+        /* prefixlen is not checked when parsed DBus message. Let's check it here. */
+        if (prefixlen > (family == AF_INET ? 32 : 128)) {
```
poettering, line 188: `FAMILY_ADDRESS_SIZE_SAFE(family)`

src/network/networkd-wwan.c
```diff
@@ -0,0 +1,444 @@
...
+                log_link_debug(link, "Bearer has invalid prefix length %u for %s address, ignoring.",
+                               prefixlen,
+                               family == AF_INET ? "IPv4" : "IPv6");
```
poettering, line 191: af_to_ipv4_ipv6(family)

WhyNotHugo:
As far as I can tell from the changeset, this will make systemd-network read network settings after the modem has connected, but **will not** instruct the modem to connect in the first place, right?

I'm writing a small tool that makes the modem connect automatically on startup or when plugged in, and want to make sure I'm not just duplicating work.

src/network/networkd-wwan-bus.c
aleksander0m: The only way to enumerate the **modem** objects is via the ObjectManager interface. Then, once you have the modem object, you can see the list of bearer objects associated to each modem in the "Bearers" property of the "Modem" interface in the modem object, see https://www.freedesktop.org/software/ModemManager/doc/latest/ModemManager/gdbus-org.freedesktop.ModemManager1.Modem.html#gdbus-property-org-freedesktop-ModemManager1-Modem.Bearers

src/network/networkd-wwan.c
```diff
@@ -0,0 +1,444 @@
...
+
+        SET_FOREACH(route, link->routes) {
+                if (route->source != NETWORK_CONFIG_SOURCE_MODEM_MANAGER)
+                        continue;
+
+                route_mark(route);
+        }
+
+        if (b && FLAGS_SET(b->ip_type, ADDRESS_FAMILY_IPV4)) {
+                if (b->connected && b->ip4_method == MM_BEARER_IP_METHOD_STATIC) {
```
aleksander0m, line 297: What about `MM_BEARER_IP_METHOD_PPP`? looks like that would be totally ignored, right? is it out of scope of what networkd would do?

aleksander0m:
> As far as I can tell from the changeset, this will make systemd-network read network settings after the modem has connected, but **will not** instruct the modem to connect in the first place, right?

That is what it looks like; the code is monitoring bearer objects exposed by MM, regardless of what's the associated modem, and just applying the net interface configs as required by MM (and clearing them when disconnected).

src/network/networkd-wwan-bus.c
aleksander0m: The usual way to work with this would be to get the initial list of modem objects and the added/removed events from the ObjectManager interface; and then monitor property updates in the "Bearers" property of the "Modem" object to get notified about bearer additions/removals, and then monitor the property updates of each of the bearers. It's quite some work without a helper library like libmm-glib, truth be told.

If you truly are only interested in monitoring the bearer objects, listening for property updates on `/org/freedesktop/ModemManager/Bearer` objects doesn't seem that bad idea.

Just beware, because there will be bearer objects reported as connected but without any data interface associated, e.g. the initial EPS bearer
```
aleksander@hestia:~$ mmcli -m a
...
  -----------------------------------
  3GPP EPS |    ue mode of operation: csps-2
           |     initial bearer path: /org/freedesktop/ModemManager1/Bearer/0
...

aleksander@hestia:~$ mmcli -b 0
  -------------------------
  General    |        path: /org/freedesktop/ModemManager1/Bearer/0
             |        type: default-attach
  -------------------------
  Status     |   connected: yes
             |   suspended: no
             | multiplexed: no
             |  ip timeout: 20
  -------------------------
  Properties |         apn: telefonica.es
             |     ip type: ipv4

```

You should ignore those bearers with type `default-attach` and without any interface reported.

shpalani:
Hi author, I'm interested in knowing the commit of this systemd-networkd support. Can you please let me know? Thanks.

sequencer:
Thanks @yuwata, I think after conflicts resolved, I can cherry-pick and test it.

pschichtel:
I considered testing this PR, but I noticed it got flagged as draft. what's the reason for that?

PR #22317, opened 2022-01-31
meson: use bpftool based strip when available

jameshilliard:
This should be useable in bpftool v5.13 or newer based on:
https://github.com/torvalds/linux/commit/d80b2fcbe0a023619e0fc73112f2a02c2662f6ab

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1012,23 +1012,35 @@ else
                 clang_supports_bpf = false
         endif
 
-        if not meson.is_cross_build() and clang_found
-                llvm_strip_bin = run_command(clang, '--print-prog-name', 'llvm-strip',
-                                             check : true).stdout().strip()
-        else
-                llvm_strip_bin = 'llvm-strip'
-        endif
-        llvm_strip = find_program(llvm_strip_bin, required : bpf_framework_required, version : '>= 10.0.0')
-
         # Debian installs this in /usr/sbin/ which is not in $PATH.
         # We check for 'bpftool' first, honouring $PATH, and in /usr/sbin/ for Debian.
         # We use 'bpftool gen' subcommand, it was added by 985ead416df39d6fe8e89580cc1db6aa273e0175 (v5.6).
         bpftool = find_program('bpftool',
                                '/usr/sbin/bpftool',
-                               required : bpf_framework_required,
-                               version : '>= 5.6')
+                               required : false,
+                               version : '>= 5.13.0')
 
-        deps_found = clang_found and clang_supports_bpf and clang_supports_flags and llvm_strip.found() and bpftool.found()
+        if bpftool.found()
+                bpftool_strip = true
+        else
+                bpftool_strip = false
+                bpftool = find_program('bpftool',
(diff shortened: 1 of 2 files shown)


yuwata (approved):

yuwata:
Please rebase.

meson.build
```diff
@@ -1012,23 +1012,35 @@ else
...
+                               version : '>= 5.13.0')
+
+        if bpftool.found()
+                bpftool_strip = true
+        else
+                bpftool_strip = false
+                bpftool = find_program('bpftool',
+                                       '/usr/sbin/bpftool',
+                                       required : bpf_framework_required,
+                                       version : '>= 5.6.0')
```
jameshilliard, line 1030: I changed this to the full 3 part version string since that seems to match the format output from `bpftool`:
```
$ ./bpftool --version
./bpftool v5.16.0-c446fdacb10d
features: libbpf_strict
```

jameshilliard:
rebased

yuwata (approved):
LGTM.

src/core/bpf/meson.build
```diff
@@ -65,13 +65,23 @@ bpf_o_unstripped_cmd += [
...
-        '-g',
-        '@INPUT@',
-        '-o',
-        '@OUTPUT@'
-]
+if bpftool_strip
+        bpf_o_cmd = [
+                bpftool,
+                'g',
+                'o',
```
anakryiko, line 72: 'gen' and 'object' is probably better. It has nothing to do with -g and -o of llvm-strip, btw (pure coincidence).

src/core/bpf/meson.build
```diff
@@ -65,13 +65,23 @@ bpf_o_unstripped_cmd += [
...
+                '@OUTPUT@',
+                '@INPUT@'
```
anakryiko, line 74: is @INPUT@ always a single file? `bpftool gen object` allows statically linking BPF application that consists of multiple .bpf.o objects, that's it's main purpose. Stripping out DWARF info is just a by product. So I'm wondering if ability to do multi-file BPF programs was added by this PR as well or it's purely about stripping out DWARF?

src/core/bpf/meson.build
jameshilliard: I copied the style from the gen skeleton command here, should I change both?:
https://github.com/systemd/systemd/blob/e3759ac43a3c0375e54799986343e2f101d14cc4/src/core/bpf/meson.build#L86-L91

src/core/bpf/meson.build
jameshilliard: > is @input@ always a single file?

Should be at the moment.

> Stripping out DWARF info is just a by product. So I'm wondering if ability to do multi-file BPF programs was added by this PR as well or it's purely about stripping out DWARF?

I didn't make any changes for multi-file BPF program support.

src/core/bpf/meson.build
jameshilliard: I've changed these to full names in #22330.

PR #22321, opened 2022-01-31
test: copy portable profiles into the image if they don't exist there

mrc0mmand:
otherwise the test fails with:

```
systemd-analyze[636]: Failed to find portable profile strict: No such file or directory
```

when run with `-Dportabled=false` build.

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -1163,6 +1163,17 @@ install_systemd() {
         mkdir -p "$initdir/etc/systemd/system/service.d/"
         echo -e "[Service]\nProtectSystem=no\nProtectHome=no\n" >"$initdir/etc/systemd/system/service.d/gcov-override.conf"
     fi
+
+    # If we're built with -Dportabled=false, tests with systemd-analyze
+    # --profile will fail. Since we need just the profile (text) files, let's
+    # copy them into the image if they don't exist there.
+    local portable_dir="${initdir:?}${ROOTLIBDIR:?}/portable"
+    if [[ ! -d "$portable_dir/profile/strict" ]]; then
+        dinfo "Couldn't find portable profiles in the test image"
+        dinfo "Copying them directly from the source tree"
+        mkdir -p "$portable_dir"
+        cp -frv "${SOURCE_DIR:?}/src/portable/profile" "$portable_dir"
+    fi
 }
 
 get_ldpath() {


bluca:
Any chance we can fix the image instead? In which case is it missing?

mrc0mmand:
> Any chance we can fix the image instead? In which case is it missing?

The upstream images are fine, this is just case in RHEL where we build systemd without portable stuff. And it would be a shame to disable the whole test, since the rest of it is useful even without portable profiles.

bluca:
> > Any chance we can fix the image instead? In which case is it missing?
> 
> The upstream images are fine, this is just case in RHEL where we build systemd without portable stuff. And it would be a shame to disable the whole test, since the rest of it is useful even without portable profiles.

It's just the profile text files that are needed, so we don't need to build anything extra, can we add them when building the image from test/test-functions if they are missing?

mrc0mmand:
Pushed a new version which copies the profiles from the source tree into the image if they don't exist there. Tested it with both `-Dportabled=true` and `-Dportabled=false` and it seems to DTRT.

PTAL!

test/TEST-65-ANALYZE/test.sh
```diff
@@ -7,4 +7,18 @@ TEST_DESCRIPTION="test analyze"
 # shellcheck source=test/test-functions
 . "${TEST_BASE_DIR:?}/test-functions"
 
+test_append_files() {(
```
bluca: I think this works only if there's also a new image name defined? From the top of test-functions:

```
# Note that defining a different IMAGE_NAME in a test setup script will only result
# in default.img being copied and renamed. It can then be extended by defining
# a test_append_files() function.
```

Given it's just a handful of text files, I'd just do this from the generic setup that applies to all images. Not worth to create an entire new image just for a few bytes of text.

mrc0mmand: Makes sense.

bluca: The append would work when running the test individually, since the image is built just there, but not when all tests are run in sequence, where the image is reused to save time

mrc0mmand:
Updated (and the test still passes in both cases).

bluca (approved):

mrc0mmand:
The TEST-69 fail is unrelated, see #22323.

PR #22323, opened 2022-01-31
test: use ~/.bashrc instead of /etc/bash.bashrc 

mrc0mmand:
/etc/bash.bashrc appears to be Debian-specific[0] (or specific to bash
versions compiled with -DSYS_BASHRC), which doesn't seem to be the case
for C8S. Let's use ~/.bashrc instead to make the CIs happy again.

[0] https://sources.debian.org/src/bash/5.1-6/debian/README/#L50-L56

### DIFF
--- a/test/TEST-69-SHUTDOWN/test.sh
+++ b/test/TEST-69-SHUTDOWN/test.sh
@@ -26,9 +26,9 @@ test_append_files() {
     # prevent shutdown in test suite, the expect script does that manually.
     rm "${workspace:?}/usr/lib/systemd/tests/testdata/units/end.service"
     inst /usr/bin/screen
-    echo "PS1='screen\$WINDOW # '" >"$workspace/etc/bash.bashrc"
+    echo "PS1='screen\$WINDOW # '" >>"$workspace/root/.bashrc"
     echo 'startup_message off' >"$workspace/etc/screenrc"
-    echo 'bell_msg ""' >>"$1/etc/screenrc"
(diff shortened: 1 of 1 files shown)


test/TEST-69-SHUTDOWN/test.sh
```diff
@@ -9,25 +9,26 @@ TEST_NO_QEMU=1
...
     # prevent shutdown in test suite, the expect script does that manually.
-    rm "$1"/usr/lib/systemd/tests/testdata/units/end.service
+    rm "${workspace:?}/usr/lib/systemd/tests/testdata/units/end.service"
     inst /usr/bin/screen
-    echo "PS1='screen\$WINDOW # '" > "$1"/etc/bash.bashrc
-    echo 'startup_message off' > "$1"/etc/screenrc
-    echo 'bell_msg ""' >> "$1"/etc/screenrc
+    echo "PS1='screen\$WINDOW # '" >"$workspace/etc/bash.bashrc"
+    echo 'startup_message off' >"$workspace/etc/screenrc"
+    echo 'bell_msg ""' >>"$1/etc/screenrc"
```
bluca: $workspace?

mrc0mmand: whoops, thanks

bluca (approved):

mrc0mmand:
Merging a bit earlier to make CI green again, since the issue affected only C8S. I'll go ahead and restart the failed C8S jobs after the merge.

PR #20598, opened 2021-08-31
fstab-generator, core/mount: Avoid special fs target before dependency only if nofail is used

tomty89:
The ordering dependency should not be a problem even when noauto / automount is used.

P.S. Ideally speaking, it appears to me that it's probably best if we could by some means have it for nofail mount units as well. (Like, pull the units with a separate target that is ordered after the existing fs targets?)

PR #22326, opened 2022-01-31
Revert "fstab-generator, core/mount: Avoid special fs target before dependency only if nofail is used"

poettering:
Reverts systemd/systemd#20598

I think this PR is not the way to go, see comments in the PR.

Please don't merge unti discussion around this settled.

### DIFF
--- a/src/core/mount.c
+++ b/src/core/mount.c
@@ -84,6 +84,14 @@ static MountParameters* get_mount_parameters(Mount *m) {
         return get_mount_parameters_fragment(m);
 }
 
+static bool mount_is_automount(const MountParameters *p) {
+        assert(p);
+
+        return fstab_test_option(p->options,
+                                 "comment=systemd.automount\0"
+                                 "x-systemd.automount\0");
+}
+
 static bool mount_is_network(const MountParameters *p) {
         assert(p);
 
@@ -476,7 +484,7 @@ static int mount_add_default_ordering_dependencies(
                 before = SPECIAL_LOCAL_FS_TARGET;
         }
 
-        if (!mount_is_nofail(m)) {
+        if (!mount_is_nofail(m) && !mount_is_automount(p)) {
                 r = unit_add_dependency_by_name(UNIT(m), UNIT_BEFORE, before, true, mask);
                 if (r < 0)
                         return r;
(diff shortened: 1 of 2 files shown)


keszybz:
It isn't clear what to do with this. After re-reading the discussion in https://github.com/systemd/systemd/pull/20598, my understanding is that: a) the patch fixes a real issue with shutdowns, b) the effect of the removal of parallelization only occurs if the mount was triggered during boot, i.e. it only matters in the narrow case where either the mount is explicitly pulled into the initial transaction but nothing blocks on it, or it is implicitly triggered by something which isn't ordered wrt to `local-fs.target` but is running very early. I think that correctness trumps optimization, so we shouldn't merge *this* PR, I'll drop it from the milestone.

It seems we want asymmetrical start/stop behaviour: no ordering on start, but ordering of the mount point stop job after local-fs.target/stop.

tomty89:
> It seems we want asymmetrical start/stop behaviour: no ordering on start, but ordering of the mount point stop job after local-fs.target/stop.

Yes, that's exactly the case. For example `nofail` is still "broken" as it cannot be covered by #20598. While I strongly urge that this should not be merged, it is true that a more general fix is desired for the problem that #20598 partially addresses.

See also https://github.com/systemd/systemd/issues/22336

keszybz:
@poettering comments?

keszybz:
It seems that the whole situation with mount ordering during bootup and shutdown needs a rethink. This issue has seen no movement in months, so let's bump it to the next milestone.

poettering:
I don#t grok what #20598 actually was supposed to fix. Can anyone explain this to me in clear words without just telling me to reread that PR again?

I know what it breaks, I don#t see what it fixes. the commits do not explain that afaics.

poettering:
See my explanation why the PR this one reverts made no sense → https://github.com/systemd/systemd/pull/20598#issuecomment-1256361078

tomty89:
@poettering  Let me ask you an extra question here: are you even aware that we've been relying on the ordering against local-fs.target to delay unmounting? Regardless of whether that's intentional or, historical.

As I've said in one of my other posts, if you think *that* itself is bad/wrong, what you really want to spend time on is getting rid of *that*, not #20598.

Also, you keep saying that #20598 would defeat the point of automount, but you haven't exactly given a concrete example either.

Maybe *I* haven't been conveying the point clearly, but with additionally the well-written clarification / summary from @keszybz, I can't help but feel unsure whether you actually *want* to see the point.

bluca:
This has been done for more than a year, can we close this revert or is there something left to do?

PR #22327, opened 2022-01-31
resolved: misc. small DnsStream refactors and improvements

joanbm:
Various improvements around the DnsStream implementation, mostly related to DNS over TLS.

Summary:

1. **Make event flags logic robust for DoT**: Calling `sd_event_source_get_io_events` and checking for the `EPOLLIN`/`EPOLLOUT` flags like we were doing in order to know whether we want to read/write a packet is wrong, because the TLS library could override the events through the `dnstls_events` field. Conceptually we should be checking the flags before overriding them with `dnstls_events`. (Fixes no actual bug that I know of).

2. **Avoid multiple SSL writes per DoT packet**: During #22132 I noticed (in the socat capture) that for each question we sent out we created two TLS records. This is due to decomposing an iovec into multiple `SSL_write`/`gnutls_record_send` calls. Refactor the logic to create a single TLS record per packet. Possible minor performance improvement.

3. **Read as much as possible per stream `EPOLLIN` event**: This is an almost complete refactor of #22132 using a much simpler idea: Simply try to read as much data as possible when we get an `EPOLLIN` event, until we exhaust the socket. The looping logic is thus much simpler and shared between TCP and TLS. Also possible minor performance improvement.

4. **Allow `test-resolved-stream` to run concurrently**: If possible run `test-resolved-stream` in a network namespace in order to avoid binding ports in the host, this is more robust and allows the test to run concurrently.

Full commit messages follow.

---

# Make event flags logic robust for DoT

Since when handling a DNS over TLS stream, the TLS library can override the requested events through `dnstls_events` for handshake/shutdown purposes, obtaining the event flags through `sd_event_source_get_io_events` and checking for `EPOLLIN` or `EPOLLOUT` does not really tell us whether we want to read/write a packet. Instead, it could just be OpenSSL/GnuTLS doing something else.

To make the logic more robust (and simpler), save the flags that tell us whether we want to read/write a packet, and check them instead of the IO flags.

(& use `uint32_t` for the flags like in `sd_event_source_set_io_events` prototype)

---

# resolved: Avoid multiple SSL writes per DoT packet

In the DoT case, `dns_stream_writev` decomposed an iovec into multiple `dnstls_stream_write` calls, which resulted in multiple SSL writes and multiple TLS records. This can be checked from a network capture, e.g. using socat: `socat -v -x openssl-listen:853,reuseaddr,fork,cert=my.cert,key=my.key,verify=0 openssl:8.8.8.8:853`

Instead, propagate the iovec as-is into the DoT handling code. For GnuTLS, the library provides support for buffering ('corking') a record. OpenSSL has no such facility, so we join the iovec into a single buffer then call `SSL_write`.

socat capture of `resolvectl -4 query --cache=no example.com` before the commit:

```
> 2022/01/30 13:35:52.194200  length=2 from=0 to=1
 00 28                                            .(
--
> 2022/01/30 13:35:52.194253  length=40 from=2 to=41
 1e b2 01 00 00 01 00 00 00 00 00 01 07 65 78 61  .............exa
 6d 70 6c 65 03 63 6f 6d 00 00 01 00 01 00 00 29  mple.com.......)
 ff e4 00 00 00 00 00 00                          ........
--
< 2022/01/30 13:35:52.232798  length=58 from=0 to=57
 00 38 1e b2 81 80 00 01 00 01 00 00 00 01 07 65  .8.............e
 78 61 6d 70 6c 65 03 63 6f 6d 00 00 01 00 01 c0  xample.com......
 0c 00 01 00 01 00 00 53 6f 00 04 5d b8 d8 22 00  .......So..]..".
 00 29 02 00 00 00 00 00 00 00                    .)........
```

socat capture of `resolvectl -4 query --cache=no example.com` after the commit:

```
> 2022/01/30 13:34:47.598099  length=42 from=504 to=545
 00 28 37 86 01 00 00 01 00 00 00 00 00 01 07 65  .(7............e
 78 61 6d 70 6c 65 03 63 6f 6d 00 00 01 00 01 00  xample.com......
 00 29 ff e4 00 00 00 00 00 00                    .)........
--
< 2022/01/30 13:34:47.613203  length=58 from=756 to=813
 00 38 37 86 81 80 00 01 00 01 00 00 00 01 07 65  .87............e
 78 61 6d 70 6c 65 03 63 6f 6d 00 00 01 00 01 c0  xample.com......
 0c 00 01 00 01 00 00 52 5e 00 04 5d b8 d8 22 00  .......R^..]..".
 00 29 02 00 00 00 00 00 00 00                    .)........
```

---

# resolved: Read as much as possible per stream EPOLLIN event

In commit 2aaf6bb, an issue was fixed where systemd-resolved could get stuck for multiple seconds waiting for incoming data, since GnuTLS/OpenSSL can buffer a TLS record, so data could be available, but no `EPOLLIN` event would be generated.

To fix this, a somewhat elaborate logic consisting on asking the TLS library whether it had buffered data, then "faking" an `EPOLLIN` event was implemented.

However, there is a much simpler solution: Always read as much data as available (i.e. until we get an event like `EAGAIN` when trying to read) from the stream when we get an `EPOLLIN` event, instead of at most a single packet per event. This approach does not require asking the TLS library whether it has buffered data, and the logic is exactly the same for both the TCP and TLS case.

`test-resolved-stream` is fixed to avoid a latent double free bug.

---

# resolved: Allow test-resolved-stream to run concurrently

Since `test-resolved-stream` brings up a simple DNS server on 127.0.0.1:12345, only one instance could run at a time, so it would fail when run like `meson test -C build test-resolved-stream --repeat=1000`. Similarly, if by chance something is up on port 12345, the test would fail.

To make the test more reliable, run it in an isolated user + network namespace. If this fails (some distributions disable user namespaces), just run as before.

### DIFF
--- a/src/resolve/test-resolved-stream.c
+++ b/src/resolve/test-resolved-stream.c
@@ -2,10 +2,12 @@
 
 #include <arpa/inet.h>
 #include <fcntl.h>
+#include <net/if.h>
 #include <pthread.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/ioctl.h>
 #include <sys/prctl.h>
 #include <sys/socket.h>
 #include <sys/wait.h>
@@ -13,6 +15,7 @@
 
 #include "fd-util.h"
 #include "log.h"
+#include "macro.h"
 #include "process-util.h"
 #include "resolved-dns-packet.h"
 #include "resolved-dns-question.h"
@@ -327,6 +330,24 @@ static void test_dns_stream(bool tls) {
         log_info("test-resolved-stream: Finished %s test", tls ? "TLS" : "TCP");
 }
 
+static void try_isolate_network(void) {
+        _cleanup_close_ int socket_fd = -1;
+
+        if (unshare(CLONE_NEWUSER | CLONE_NEWNET) < 0) {
+                log_warning("test-resolved-stream: Can't create user and network ns, running on host");
+                return;
+        }
+
+        /* Bring up the loopback interfaceon the newly created network namespace */
+        struct ifreq req = { .ifr_ifindex = 1 };
+        assert_se((socket_fd = socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0)) >= 0);
+        assert_se(ioctl(socket_fd,SIOCGIFNAME,&req) >= 0);
+        assert_se(ioctl(socket_fd, SIOCGIFFLAGS, &req) >= 0);
+        assert_se(FLAGS_SET(req.ifr_flags, IFF_LOOPBACK));
+        req.ifr_flags |= IFF_UP;
+        assert_se(ioctl(socket_fd, SIOCSIFFLAGS, &req) >= 0);
+}
+
 int main(int argc, char **argv) {
         SERVER_ADDRESS = (struct sockaddr_in) {
                 .sin_family = AF_INET,
@@ -335,6 +356,8 @@ int main(int argc, char **argv) {
         };
 
         test_setup_logging(LOG_DEBUG);
+
+        try_isolate_network();
 
         test_dns_stream(false);
 #if ENABLE_DNS_OVER_TLS


src/resolve/resolved-dns-stream.c
```diff
@@ -354,7 +352,7 @@ static int on_stream_io_impl(DnsStream *s, uint32_t revents) {
                 }
         }
 
-        if ((revents & (EPOLLIN|EPOLLHUP|EPOLLRDHUP)) &&
+        while ((revents & (EPOLLIN|EPOLLHUP|EPOLLRDHUP)) &&
             (!s->read_packet ||
              s->n_read < sizeof(s->read_size) + s->read_packet->size)) {
```
yuwata: Indentation.

yuwata (approved):
LGTM.

src/resolve/resolved-dns-stream.c
joanbm: Done.

PR #22330, opened 2022-02-01
meson: use full argument names for bpftool gen commands

jameshilliard:
This should be a purely cosmetic change.


anakryiko (approved):
Thanks!

PR #22332, opened 2022-02-01
network: dhcp-server: allow to specify server address for DNS= or friends

### LINKED ISSUE #15026: networkd: Emit link local as DNS in DHCPServer
I would like to setup a network interface as

```
[Match]
Type=wlan
Name=wlan0

[Network]
Description=Main LAN interface
Address=0.0.0.0/24
Address=::/64
DHCPServer=yes

[DHCPServer]
PoolOffset=10
PoolSize=40
DefaultLeaseTimeSec=600
MaxLeaseTimeSec=7200
EmitDNS=yes
DNS=_link_local
```

But when using `Address=0.0.0.0/24` I have no way to specify link local as a DNS server. Something similar exists in `[IPv6PrefixDelegation]`.

For now, the only workaround (AFAICT) is to specify explicitly the address range to use.

yuwata:
Closes #15026.

### DIFF
--- a/test/test-network/conf/dhcp-server-downstream.network
+++ b/test/test-network/conf/dhcp-server-downstream.network
@@ -10,6 +10,5 @@ DHCPServer=yes
 ServerAddress=192.168.5.1/24
 PoolOffset=10
 PoolSize=50
+UplinkInterface=dummy98
 Router=192.168.5.3
-DNS=_server_address 192.168.5.10
-NTP=_server_address 192.168.5.11
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -4233,6 +4233,7 @@ class NetworkdDHCPServerTests(unittest.TestCase, Utilities):
         'dhcp-client-static-lease.network',
         'dhcp-client-timezone-router.network',
         'dhcp-server.network',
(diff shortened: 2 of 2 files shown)


src/network/networkd-dhcp-server.c
```diff
@@ -639,11 +639,19 @@ int config_parse_dhcp_server_emit(
...
+                        a = IN_ADDR_NULL; /* null address will be converted to the server address. */
+                else {
+                        r = in_addr_from_string(AF_INET, w, &a);
+                        if (r < 0) {
+                                log_syntax(unit, LOG_WARNING, filename, line, r,
+                                           "Failed to parse %s= address '%s', ignoring: %m", lvalue, w);
+                                continue;
+                        }
+
+                        if (in4_addr_is_null(&a.in))
```
bluca: Should this be documented in the manpage too?

bluca (approved):

src/network/networkd-dhcp-server.c
yuwata: I noticed that `DNS=` in `[IPv6SendRA]` ignores a null address. I changed to this so.
And, I do not think it is not necessary to mention about the invalid address. There should be many invalid DNS server addresses, e.g. 255.255.255.255. Mentioning such addresses is overkill.

PR #22338, opened 2022-02-01
tests: rework test macros to not take code as parameters

poettering:
C macros are nasty. We use them, but we try to be conservative with
them. In particular passing literal, complex code blocks as argument is
icky, because of "," handling of C, and also because it's quite a
challange for most code highlighters and similar. Hence, let's avoid
that. Using macros for genreating functions is OK but if so, the
parameters should be simple words, not full code blocks.

hence, rework DEFINE_CUSTOM_TEST_MAIN() to take a function name instead
of code block as argument.

As side-effect this also fixes a bunch of cases where we might end up
returning a negative value from main().

Some uses of DEFINE_CUSTOM_TEST_MAIN() inserted local variables into the
main() functions, these are replaced by static variables, and their
destructors by the static destructor logic.

This doesn't fix any bugs or so, it's just supposed to make the code
easier to work with and improve it easthetically.

Or in other words: let's use macros where it really makes sense, but
let's not go overboard with it.

(And yes, FOREACH_DIRENT() is another one of those macros that take
code, and I dislike that too and regret I ever added that.)

### DIFF
--- a/src/shared/tests.h
+++ b/src/shared/tests.h
@@ -6,6 +6,7 @@
 #include "sd-daemon.h"
 
 #include "macro.h"
+#include "static-destruct.h"
 #include "util.h"
 
 static inline bool manager_errno_skip_test(int r) {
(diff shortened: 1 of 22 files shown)


src/test/test-serialize.c
```diff
@@ -10,7 +10,7 @@
 #include "tests.h"
 #include "tmpfile-util.h"
 
-char long_string[LONG_LINE_MAX+1];
+static char long_string[LONG_LINE_MAX+1];
```
medhefgo, line 13: This could also be done with `= { [0 ... LONG_LINE_MAX] = 'x', '\0'}` to statically initialize. Though, dunno how that affects binary size (if we care for that for test binaries).

medhefgo (approved):
Certainly more readable in some cases.

src/shared/tests.h
```diff
@@ -109,15 +110,27 @@ static inline int run_test_table(void) {
...
         int main(int argc, char *argv[]) {               \
-                int _r = EXIT_SUCCESS;                   \
+                int _r, _q;                              \
                 test_setup_logging(log_level);           \
                 save_argc_argv(argc, argv);              \
-                intro;                                   \
-                _r = run_test_table();                   \
-                outro;                                   \
-                return _r;                               \
+                _r = intro();                            \
```
dnicolodi, line 122: Just a comment passing by: if this is turned into something like `_r = intro ? intro() : EXIT_SUCCESS;` the macro could be called with `NULL` as function parameter, which would be more natural than `test_nop` and knowing that not special setup is requested would not require looking up the definition of `test_nop()`.

medhefgo: Ooh, I like that.

yuwata: Let's do that later.

yuwata: #22356.

PR #22340, opened 2022-02-01
journal: Add features to make handling of .journal~ files easier

michalfabik:
Related: https://bugzilla.redhat.com/show_bug.cgi?id=1433652  
  
- Add flags to choose only `.journal` or `.journal~` when using `sd_journal_open_directory()  
- Add options to `journalctl` to rename and delete corrupted journal files.  

See commits.

### DIFF
--- a/test/units/testsuite-04.sh
+++ b/test/units/testsuite-04.sh
@@ -119,6 +119,26 @@ systemctl start silent-success
 journalctl --sync
 [[ -z "$(journalctl -b -q -u silent-success.service)" ]]
 
+# Test proper handling of .journal~ files
+MACHINE_ID=$(cat /etc/machine-id)
+for _ in {0..4}; do
+    dd if=/dev/urandom bs=1M count=1 | base64 | systemd-cat
+    journalctl --rotate
+    journalctl --sync
+done
+sed -i 's/MESSAGE/XXXXXXX/' /var/log/journal/"$MACHINE_ID"/system\@*.journal
+journalctl --verify-and-rename
+for _ in {0..4}; do
+    dd if=/dev/urandom bs=1M count=1 | base64 | systemd-cat
+    journalctl --rotate
+    journalctl --sync
+done
+journalctl --vacuum-files=1 --vacuum-corrupted=only
+journalctl --vacuum-files=1 --vacuum-corrupted=no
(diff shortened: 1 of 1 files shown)


poettering:
I am sorry, what's the usecase for this?

Why would you want to exclude reading `.journal~` files? They are not something apps should be bothered with, so not sure why this should be exposed in a user facing UI?

michalfabik:
> I am sorry, what's the usecase for this?  
  
Basically, the bottom three comments in https://bugzilla.redhat.com/show_bug.cgi?id=1433652  
  
In `abrt`, we read the journal with `sd_journal_open...` and sometimes we encounter a crash due to a corrupted (`.journal~`) file. Currently, there's no easy way to ignore those. That's what the new flags are for.  
  
The new `journalclt` options are intended mostly for developers' convenience when debugging etc.

poettering:
> Basically, the bottom three comments in https://bugzilla.redhat.com/show_bug.cgi?id=1433652
> 
> In `abrt`, we read the journal with `sd_journal_open...` and sometimes we encounter a crash due to a corrupted (`.journal~`) file. Currently, there's no easy way to ignore those. That's what the new flags are for.

We try very hard to gracefully handle those files. The only reason we rename them is so that we stop writing to them, i.e. to not make things worse. But our journal reading code is supposed to be able to make the best of any journal file it finds, i.e. read the data it can read and ignore the rest. If there are journal files that make the journal read fail completely then this would be a bug in sd-journal and not something to work around in abrt.

Hence: do you have journal files handy that supposedly make journalctl/sd-journal misbheave? If so it would be good to attach them to a bug report (or send the personally to me).

But I am pretty sure we should not work around these issue by simply ignoring the journal files.

michalfabik:
Understood.  
   
> do you have journal files handy that supposedly make journalctl/sd-journal misbheave? If so it would be good to attach them to a bug report (or send the personally to me).  
  
No, not me personally. Let me try to reach out to some of the reporters.

DaanDeMeyer:
Yeah it'd be great if you could attach some of the corrupted journal files. We definitely shouldn't crash on reading corrupted journal files. I'd be happy to look into this if you provide some of the corrupted journal files.

michalfabik:
The handling of corrupted journal files in `abrt` daemons is broken. This is not systemd's fault, closing.

PR #22342, opened 2022-02-01
hwdb: add Logitech G203

vincentbernat:
This setting is confirmed by mouse-dpi-tool. The specs for this mouse says "200
to 6000". Reddit reports there may be an update to get "200 to 8000", but I
don't know how to check the exact list of DPIs supported by this mouse. Nothing
in the manual.

### DIFF
--- a/hwdb.d/70-mouse.hwdb
+++ b/hwdb.d/70-mouse.hwdb
@@ -369,6 +369,10 @@ mouse:usb:v046dpc07e:name:Logitech Gaming Mouse G402:*
 mouse:usb:v046dpc083:name:Logitech G403 Prodigy Gaming Mouse:*
  MOUSE_DPI=400@1000 *800@1000 1600@1000 3200@1000
 
+# Logitech G203
+mouse:usb:v046dpc084:name:Logitech G203 Gaming Mouse:*
+ MOUSE_DPI=400@1000
+
 # Logitech G403 Hero
 mouse:usb:v046dpc08f:name:Logitech G403 HERO Gaming Mouse:*
  MOUSE_DPI=400@1000 *800@1000 1600@1000 3200@1000


whot:
Those logitech devices can be programmed via their software, the DPI list is usually in 50 or 100 steps increment so you get all the intervals between min/max (logitech says 8000 btw, see [here](https://www.logitechg.com/en-us/products/gaming-mice/g203-prodigy-gaming-mouse.910-005081.html#product-tech-specs) unless the G203 is different to the G203 Prodigy)

I don't think adding these with 400 is a good idea - they can (and usually are) changed by users to whatever they feel like anyway, 400 means it's an extremely low DPI default and I'm not sure every mouse even comes with that default.

libratbag has a device entry for this device btw so you should be able to change resolutions with e.g. Piper.

vincentbernat:
Okay, ratbagctl says the active resolution for my mice is 400 but the default one should be 800. I have always used this mouse on Linux without knowledge of such tool, so I don't know why the resolution profile was changed.

```
singing-gundi - Logitech G203 Prodigy Gaming Mouse
             Model: usb:046d:c084:0
 Number of Buttons: 6
    Number of Leds: 1
Number of Profiles: 1
Profile 0: (active)
  Name: n/a
  Report Rate: 1000Hz
  Resolutions:
    0: 400dpi (active)
    1: 800dpi (default)
    2: 1600dpi
    3: 3200dpi
    4: 0dpi
  Button: 0 is mapped to 'button 1'
  Button: 1 is mapped to 'button 2'
  Button: 2 is mapped to 'button 3'
  Button: 3 is mapped to 'button 4'
  Button: 4 is mapped to 'button 5'
  Button: 5 is mapped to 'resolution-cycle-up'
  LED: 0, depth: rgb, mode: cycle, duration: 8000, brightness: 255
```

There are 50 resolutions available (200 to 8000). Should I list them all? There is also 4 configurable rates, the default being 1000. Should I list the 200 possible combinations?

whot:
>   Button: 5 is mapped to 'resolution-cycle-up'

If you press that button, you cycle through the resolutions, so next time you'll get 800, then 1600 etc. I'm assuming you have pressed that button a few times before which is how you ended up at the current 400. Maybe not intentionally, but still.

> Should I list them all?

No, because this list is for **fixed** resolutions so we can work with or around them. The few cases where we have multiple resolutions with a default is for the few mice that have a fixed set of resolutions that cannot be changed. Your device is fully configurable, so simply said - this device does not need an entry in the hwdb because any entry will be wrong on the majority of used devices anyway (sometimes as soon as a button is pressed like in your case).

vincentbernat:
Understood! So, closing this PR.

PR #22343, opened 2022-02-01
efi: add/user assert_se()/ASSERT_SE_PTR() for OOM handling

poettering:
Abortive OOM handling should not be optimized away by NDEBUG.

### DIFF
--- a/src/boot/efi/util.h
+++ b/src/boot/efi/util.h
@@ -27,16 +27,15 @@
 #define xnew_alloc(type, n, alloc)                                           \
         ({                                                                   \
                 UINTN _alloc_size;                                           \
-                if (__builtin_mul_overflow(sizeof(type), (n), &_alloc_size)) \
-                        assert_not_reached();                                \
+                assert_se(!__builtin_mul_overflow(sizeof(type), (n), &_alloc_size)); \
                 (type *) alloc(_alloc_size);                                 \
         })
 
-#define xallocate_pool(size) ASSERT_PTR(AllocatePool(size))
-#define xallocate_zero_pool(size) ASSERT_PTR(AllocateZeroPool(size))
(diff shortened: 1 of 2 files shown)


bluca (approved):

src/boot/efi/util.h
```diff
@@ -27,16 +27,15 @@
 #define xnew_alloc(type, n, alloc)                                           \
         ({                                                                   \
                 UINTN _alloc_size;                                           \
-                if (__builtin_mul_overflow(sizeof(type), (n), &_alloc_size)) \
-                        assert_not_reached();                                \
+                assert_se(!__builtin_mul_overflow(sizeof(type), (n), &_alloc_size)); \
```
medhefgo, line 30: The ocd in me is complaining that the remaining `\` arent't aligned.

medhefgo (approved):
Dunno why that didn't occur to me...

src/boot/efi/util.h
poettering: This is how emacs aligns it, and emacs is the law, don't you know? ;-)

i.e. it aligns \ on the same column over the whole macro, except when lines are above some threshold, in which case it puts it right after the end of the line. This threshold is hit here...

medhefgo: Blurgh. An editor you have learn to use and is efficient on resources? You should go with the flow and use an Electron powered IDE. :stuck_out_tongue:

poettering:
> imho NDEBUG is windows-y stuff that has no reason to exist on Linux. No distro that I know of uses it. But I digress...

Yeah, it's a stupid concept. But it's not Windowsy, it's in ANSI C iirc, and more something for the Gentoofunrollloops crowd (https://www.shlomifish.org/humour/by-others/funroll-loops/Gentoo-is-Rice.html)

Issue #22320, opened 2022-01-31
TEST-56-EXIT-TYPE hags in nspawn when run on a host using legacy cgroups


mrc0mmand:
**systemd version the issue has been seen with**
 > v250.3 (stable)

**Used distribution**
 > RHEL 9

**Linux kernel version used** (`uname -a`)

 > `5.14.0-47.el9.x86_64`

First of all, I'm not sure if `ExitType=cgroup` is supposed to work under legacy cgroups (but the test's QEMU part works fine even with legacy cgroups), so I'm a bit puzzled.

Currently the test hangs and eventually timeouts when running the nspawn part on a host with legacy cgroups (`systemd.unified_cgroup_hierarchy=0 systemd.legacy_systemd_cgroup_controller=1`):

```
[ 1797.954362] H test56-exit-cgroup-parentless.sh[79]: + systemctl kill --signal 9 six
[ 1797.957963] H systemd[1]: Bus private-bus-connection: changing state UNSET → OPENING
[ 1797.958261] H systemd[1]: sd-bus: starting bus private-bus-connection on fds 38/38 (socket:[284205], socket:[284205])...
[ 1797.958366] H systemd[1]: Bus private-bus-connection: changing state OPENING → AUTHENTICATING
[ 1797.958448] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1 iface=org.freedesktop.systemd1.Manager
[ 1797.958790] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/job iface=org.freedesktop.systemd1.Job
[ 1797.958916] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Unit
[ 1797.959388] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Automount
[ 1797.959499] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Device
[ 1797.959590] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Mount
[ 1797.960331] H systemd[1]: Registering bus object implementation for path=/org/freedesktop/systemd1/unit iface=org.freedesktop.systemd1.Path
[...]

PR #22344, opened 2022-02-01
core: warn on ExitType=cgroup with legacy cgroup setup

bluca:
'cgroup empty' notifications are not reliable on v1, so log a warning.

See: https://github.com/systemd/systemd/issues/22320

### DIFF
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -624,6 +624,9 @@ static int service_verify(Service *s) {
         if (s->runtime_max_usec == USEC_INFINITY && s->runtime_rand_extra_usec != 0)
                 log_unit_warning(UNIT(s), "Service has RuntimeRandomizedExtraSec= setting, but no RuntimeMaxSec=. Ignoring.");
 
+        if (s->exit_type == SERVICE_EXIT_CGROUP && cg_unified() < CGROUP_UNIFIED_SYSTEMD)
+                log_unit_warning(UNIT(s), "Service has ExitType=cgroup set, but we are running with legacy cgroups v1, which might not work correctly. Continuing.");
+
         return 0;
 }
 


poettering:
Oh, uh. So far the service_verify() stuff only looked for internal inconsistency of the unit, i.e. it didn't match it with system state. I am not convinced we should change that.

I'd just skip the relevant test on cgroupvs1, but not change PID1.

(maybe add a warning, that ExitType=cgroup will not work correctly, but certainly not fail)

poettering:
> Mmh aside fromthe test, it's really user-unfriendly though, to allow it to start when we know it will be completely borked. Maybe this is not the right place, is there a better one? Refusing it when parsing ExitType? Another place in execute.c?

It's not completely borked. Just unreliable ;-)

It "mostly" works on the host. I think it should fine for typical desktop usage, just not for container stuff or overloaded system stuff...

hence I think a warning when loading the unit should be fine. the setting was added for desktop stuff after all, and for that it should be fine, given that desktop stuff is typically not that conservative to be stuck with cgroupsv1

bluca:
> > Mmh aside fromthe test, it's really user-unfriendly though, to allow it to start when we know it will be completely borked. Maybe this is not the right place, is there a better one? Refusing it when parsing ExitType? Another place in execute.c?
> 
> It's not completely borked. Just unreliable ;-)
> 
> It "mostly" works on the host. I think it should fine for typical desktop usage, just not for container stuff or overloaded system stuff...
> 
> hence I think a warning when loading the unit should be fine. the setting was added for desktop stuff after all, and for that it should be fine, given that desktop stuff is typically not that conservative to be stuck with cgroupsv1

Ok, I've changed it to a non-fatal warning, PTAL.

poettering:
shouldn't the test also be skipped? I mean, that's what the original report was about?

bluca:
> shouldn't the test also be skipped? I mean, that's what the original report was about?

that's the original issue yes, but @mrc0mmand said he was working on that - they can be handled independently

mrc0mmand:
> > shouldn't the test also be skipped? I mean, that's what the original report was about?
> 
> that's the original issue yes, but @mrc0mmand said he was working on that - they can be handled independently

Yup, still plan to do that, unfortunately I got unexpectedly sidetracked, but should be back-to-keyboard soon.

poettering:
lgtm, then.

PR #22345, opened 2022-02-01
systemctl: strings returned by sd_bus_message_read_basic() are immutable

poettering:
It's totally not OK to write to the strings returned by it, the data is
shared by all code that references the message.

While we are at it, simplify the code via
json_variant_set_field_string().

Follow-up for: 5ef599b324efbcb7af317c102b59c662df068500

### DIFF
--- a/src/systemctl/systemctl-set-environment.c
+++ b/src/systemctl/systemctl-set-environment.c
@@ -10,36 +10,36 @@
 
 static int json_transform_message(sd_bus_message *m, JsonVariant **ret) {
         _cleanup_(json_variant_unrefp) JsonVariant *v = NULL;
-        char *text;
+        const char *text;
         int r;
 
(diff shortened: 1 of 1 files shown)


src/systemctl/systemctl-set-environment.c
```diff
@@ -10,36 +10,36 @@
...
 
-                *sep++ = '\0';
+                n = strndup(text, sep - text);
+                if (!n)
+                        return log_oom();
 
-                r = json_build(&w, JSON_BUILD_OBJECT(JSON_BUILD_PAIR(text, JSON_BUILD_STRING(sep))));
-                if (r < 0)
-                        return r;
+                sep ++;
```
medhefgo: Spurious space.

bluca (approved):

PR #22350, opened 2022-02-01
journal: various fixes to journal_file_read_object()

poettering:
This fixes a bunch of issues:

pread() returns ssize_t, and returns errors in 'errno', handle that
correctly.

More importantly: it might incompletely read data in case we hit
EOF. Check for that, and handle it.

Finally, rename the function to journal_file_read_object_header(), since
it really doesn't read full objects, but only their headers.

Follow-up for: 117e21121e857b4b7d81949542e8dd257265970a

### DIFF
--- a/src/libsystemd/sd-journal/journal-file.c
+++ b/src/libsystemd/sd-journal/journal-file.c
@@ -91,8 +91,7 @@
 #  pragma GCC diagnostic ignored "-Waddress-of-packed-member"
 #endif
 
-int journal_file_tail_end(JournalFile *f, uint64_t *ret_offset) {
-        Object tail;
+int journal_file_tail_end_by_pread(JournalFile *f, uint64_t *ret_offset) {
         uint64_t p;
         int r;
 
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
Ouch, good catch, LGTM

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -126,6 +129,43 @@ int journal_file_tail_end(JournalFile *f, uint64_t *ret_offset) {
...
 
+int journal_file_tail_end_by_mmap(JournalFile *f, uint64_t *ret_offset) {
+        uint64_t p;
+        int r;
+
+        assert(f);
+        assert(f->header);
+        assert(ret_offset);
+
+        /* Same as journal_file_tail_end_by_prad() above, but operates with the usual mmap logic */
```
daandemeyer: typo here

poettering:
I added another commit on top. PTAL.

DaanDeMeyer:
I considered the same issue but since I didn't really notice any major regression by using pread(), I left it as is. I did notice major regressions yesterday when I tried to use journal_file_read_object() in some other part of journald so it makes sense to avoid all the syscalls here as well.

PR #22353, opened 2022-02-01
sd-dhcp-lease: store static routes and classless static routes in dif…

yuwata:
…ferent arrays

When classless static routes option is provided, then static routes
option should not be used. Hence, let's not mix and store them in one
storage.

This introduce sd_dhcp_lease_get_static_routes() and
sd_dhcp_lease_get_classless_routes().


bluca (approved):

evverx:
@yuwata I'm not sure `ROUTES` should have been removed. As far as I understand it prevents newer versions of networkd from handling state files written by older versions during upgrades for example.

yuwata:
Note, the lease file is only read by `networkctl`. Hence, it is not necessary to keep the support of `ROUTES`.

PR #22355, opened 2022-02-02
network: tunnel: support external mode

### LINKED ISSUE #22352: RFE: ability to create externally controlled tunnels
Additional .netdev options/properties that would generate an externally controlled tunnel. I noticed there are also restrictions currently on ip6tnl in systemd-networkd where we need to provide `Local=` and `Remote=` but in the desired state these would be `::` (is this the same as "any" in systemd-networkd?). This could also be a separate RFE for this issue to allow `Local=` and `Remote=` to be `::` for this purpose. In the next section I paste a command, but the ideal end state is to generate the following:
```
# ip -d link show tunlany0
4: tunlany0@NONE: <NOARP,UP,LOWER_UP> mtu 1452 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/tunnel6 :: brd :: promiscuity 0
    ip6tnl external addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535
# ip -6 tun
tunlany0: any/ipv6 remote :: local :: encaplimit 0 hoplimit inherit tclass 0x00 flowlabel 0x00000 (flowinfo 0x00000000)
tunlany0: any/ipv6 remote :: local :: encaplimit 0 hoplimit inherit tclass 0x00 flowlabel 0x00000 (flowinfo 0x00000000)
```

I can run `ip link add tunlany0 type ip6tnl external` separately to create the device and then use .network files to configure it. But that's silly when it could be part of .netdev.

249

cc @yuwata for thoughts

yuwata:
Closes #22352.

### DIFF
--- a/test/test-network/conf/25-ip6tnl-external.netdev
+++ b/test/test-network/conf/25-ip6tnl-external.netdev
@@ -1,0 +1,8 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[NetDev]
+Name=ip6tnl-external
+Kind=ip6tnl
+
+[Tunnel]
+Mode=ip6ip6
+External=yes
--- a/test/test-network/conf/netdev-link-local-addressing-yes.network
+++ b/test/test-network/conf/netdev-link-local-addressing-yes.network
@@ -2,6 +2,7 @@
 [Match]
 Name=bareudp99
 Name=batadv99
+Name=ip6tnl-external
 Name=ipvlan99
 Name=ipvtap99
 Name=macvlan99
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -916,6 +916,7 @@ class NetworkdNetDevTests(unittest.TestCase, Utilities):
         'ip6gretun97',
         'ip6gretun98',
         'ip6gretun99',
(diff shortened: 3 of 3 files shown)


yuwata:
cc @anitazha

test/test-network/conf/25-ip6tnl-external.netdev
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
anitazha, line 1: Maybe delete the file in this commit since it gets updated in the next one? At this point in the commit stack it doesn't have `External=true` yet

anitazha (approved):
The commits individually lgtm. I also gave this a whirl and it produced an external tunnel the way I expected.

test/test-network/conf/25-ip6tnl-external.netdev
yuwata: Ouch.

yuwata: The file is moved to the last commit.

anitazha (approved):
Thanks! Still LGTM

PR #22356, opened 2022-02-02
test: allow to set NULL to intro or outro

yuwata:
Addresses https://github.com/systemd/systemd/pull/22338#discussion_r796741033.

### DIFF
--- a/src/shared/tests.h
+++ b/src/shared/tests.h
@@ -110,27 +110,28 @@ static inline int run_test_table(void) {
         return r;
 }
 
-static inline int test_nop(void) {
-        return EXIT_SUCCESS;
-}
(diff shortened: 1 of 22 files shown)


bluca:
This doesn't seem related on focal-arm64, but it also seems new to me, so worth keeping an eye on it:

```
--- test-firewall-util begin ---
Using nftables as firewall backend.
/* test_v4(backend=nftables) */
/* test_v6(backend=nftables) */
Assertion 'fw_add_masquerade(&ctx, true, AF_INET6, &u3, prefixlen) >= 0' failed at src/test/test-firewall-util.c:34, function test_v6(). Aborting.
```

PR #22358, opened 2022-02-02
udevadm: test,test-builtin: support /dev path or device unit

yuwata:


src/udev/udev-builtin-input_id.c
```diff
@@ -112,19 +112,19 @@ static void get_cap_mask(sd_device *pdev, const char* attr,
                 log_device_debug(pdev, "Ignoring %s block %lX which is larger than maximum size", attr, val);
 
         if (test) {
```
bluca: maybe change this to `if (test && DEBUG_LOGGING)`? Saves a bunch of work that is just thrown out otherwise

bluca (approved):

yuwata:
@bluca Thanks. The above comment is addressed. The green label is upgraded.

PR #22361, opened 2022-02-02
test: use rm_rf_physical_and_freep cleanup function

yuwata:


src/test/test-exec-util.c
```diff
@@ -65,20 +66,22 @@ static void test_execute_directory_one(bool gather_stdout) {
...
-        name3 = strjoina(template_lo, "/useless");
-        overridden = strjoina(template_lo, "/overridden");
-        override = strjoina(template_hi, "/overridden");
-        masked = strjoina(template_lo, "/masked");
-        mask = strjoina(template_hi, "/masked");
-        masked2 = strjoina(template_lo, "/masked2");
-        mask2 = strjoina(template_hi, "/masked2");
-        masked2e = strjoina(template_lo, "/masked2e");
-        mask2e = strjoina(template_hi, "/masked2e");
+        assert_se(mkdtemp_malloc(template_lo, &tmp_lo) >= 0);
```
poettering: i'd probably get rid of `template_lo` and just specify `"/tmp/…"` here literally. After all you only have a single use of the string. but doesn't matter...

yuwata: Indeed. Addressed.

poettering:
still lgtm

PR #22366, opened 2022-02-02
journald: some journal file hole punching fixes

poettering:
/cc @DaanDeMeyer

### DIFF
--- a/src/journal/managed-journal-file.c
+++ b/src/journal/managed-journal-file.c
@@ -5,6 +5,7 @@
 
 #include "chattr-util.h"
 #include "copy.h"
+#include "errno-util.h"
 #include "fd-util.h"
 #include "format-util.h"
 #include "journal-authenticate.h"
@@ -91,8 +92,14 @@ static int managed_journal_file_entry_array_punch_hole(JournalFile *f, uint64_t 
                 return 0;
         }
 
-        if (fallocate(f->fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, sz) < 0)
+        if (fallocate(f->fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, sz) < 0) {
+                if (ERRNO_IS_NOT_SUPPORTED(errno)) {
+                        log_debug("Hole punching not supported by backing file system, skipping.");
+                        return -EOPNOTSUPP; /* Make recognizable */
+                }
+
                 return log_debug_errno(errno, "Failed to punch hole in entry array of %s: %m", f->path);
+        }
 
         return 0;
 }
@@ -135,8 +142,12 @@ static int managed_journal_file_punch_holes(JournalFile *f) {
                                 if (le64toh(o.data.n_entries) == 0)
                                         continue;
 
-                                (void) managed_journal_file_entry_array_punch_hole(
-                                        f, le64toh(o.data.entry_array_offset), le64toh(o.data.n_entries) - 1);
+                                r = managed_journal_file_entry_array_punch_hole(
+                                                f, le64toh(o.data.entry_array_offset), le64toh(o.data.n_entries) - 1);
+                                if (r == -EOPNOTSUPP)
+                                        return -EOPNOTSUPP;
+
+                                /* Ignore other errors */
                         }
                 }
         }


src/journal/journald-file.c
```diff
@@ -91,8 +92,14 @@ static int journald_file_entry_array_punch_hole(JournalFile *f, uint64_t p, uint
                 return 0;
         }
 
-        if (fallocate(f->fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, sz) < 0)
+        if (fallocate(f->fd, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, offset, sz) < 0) {
+                if (ERRNO_IS_NOT_SUPPORTED(errno)) {
+                        log_debug("Hole punching not supported by backing file system, skipping.");
+                        return -EOPNOTSUPP; /* Make recognizable */
```
yuwata: `SYNTHETIC_ERRNO()`?

poettering: I thought about that, but it's not as obvious where to place the `/* Make recognizable */` comment then.

yuwata: OK. Let's keep this as is.

src/journal/journald-file.c
```diff
@@ -106,9 +119,12 @@ static int journald_file_punch_holes(JournalFile *f) {
...
 
         for (uint64_t i = p; i < p + sz && n > 0; i += n) {
-                n = pread(f->fd, items, MIN(sizeof(items), p + sz - i), i);
+                size_t m = MIN(sizeof(items), p + sz - i);
+                n = pread(f->fd, items, m, i);
                 if (n < 0)
-                        return n;
+                        return log_debug_errno(errno, "Failed to read hash table items: %m");
+                if ((size_t) n != m)
+                        return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Short read while reading hash table items.");
```
daandemeyer: I'm not sure we should fail on short reads here. I think we already handle those correctly since we check for partially read hash items and exit if we end up not reading anything at all.

DaanDeMeyer (commented):
LGTM, I only have one question

mrc0mmand:
@DaanDeMeyer would it be possible to write a test for this? I tried to, but it looks I'm missing something, because I can't make journald to attempt to punch holes on a vfat partition (or at least it looks like, just going by debug logs). If I understand this correctly (which I probably don't), it should automagically try to punch holes into the journal file during the archiving process (i.e. `journalctl --rotate`. So far it looks like it successfully truncates the file (since the size of each archived file is <8M), but I can't se any debug log messages about failing `fallocate().

DaanDeMeyer:
> @DaanDeMeyer would it be possible to write a test for this? I tried to, but it looks I'm missing something, because I can't make journald to attempt to punch holes on a vfat partition (or at least it looks like, just going by debug logs). If I understand this correctly (which I probably don't), it should automagically try to punch holes into the journal file during the archiving process (i.e. journalctl --rotate. So far it looks like it successfully truncates the file (since the size of each archived file is <8M), but I can't se any debug log messages about failing fallocate().

This is likely because the entry arrays don't have > 512KB of free space. We don't punch holes unless the holes are bigger than 512KB

src/journal/journald-file.c
poettering: hmm, yeah, you actually make use of `n` there, true. will rework.

poettering:
force pushed a new version that dropped the redundant short read check on pread() again. ptal.

src/journal/managed-journal-file.c
```diff
@@ -106,9 +119,10 @@ static int managed_journal_file_punch_holes(JournalFile *f) {
         sz = le64toh(f->header->data_hash_table_size);
 
         for (uint64_t i = p; i < p + sz && n > 0; i += n) {
-                n = pread(f->fd, items, MIN(sizeof(items), p + sz - i), i);
+                size_t m = MIN(sizeof(items), p + sz - i);
```
daandemeyer, line 122: Nit but the variable is a bit superfluous now that we only use it in one place anymore.

src/journal/managed-journal-file.c
```diff
@@ -73,9 +74,15 @@ static int managed_journal_file_entry_array_punch_hole(JournalFile *f, uint64_t
...
+                ssize_t n;
+
                 o.object.size = htole64(offset - p);
-                if (pwrite(f->fd, &o, sizeof(EntryArrayObject), p) < 0)
+
+                n = pwrite(f->fd, &o, sizeof(EntryArrayObject), p);
+                if (n < 0)
                         return log_debug_errno(errno, "Failed to modify entry array object size: %m");
+                if ((size_t) n != sizeof(EntryArrayObject))
+                        return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Short pwrite() while modifying entry array object size.");
```
daandemeyer, line 85: Technically, even if we do a short write, we might still have managed to write the object size which is the only thing we really care about. Maybe the call to pwrite() should be changed to only try to write up until and including the size of the object (so replace `sizeof(EntryArrayObject)` with `offsetof(Object, object.field) + sizeof(le64_t)`)? Kind of a nit though so setting the green label

DaanDeMeyer (approved):
Some nits, nothing really important so setting the green label

PR #22367, opened 2022-02-02
meson: extends putting extra net naming schemes together during confi…

jacekmigacz:
…guration by negative values

In downstream, we define custom combinations and sometimes it is handy to rely
on some already defined set to alias it or to append something on top of it.
It is not possible however to toggle (disable) schema out of such set.

With this patch meson extra-net-naming-schemes can be provided with negative
values, like:

$ meson configure build \
  -Dextra-net-naming-schemes=gargoyle=v239+-slot_function_id+-16bit_index+-replace_strictly
It will yield:
NAMING_GARGOYLE = NAMING_V239 & ~NAMING_SLOT_FUNCTION_ID & ~NAMING_16BIT_INDEX \
  & ~NAMING_REPLACE_STRICTLY

That would effectively drag naming schema from v249, back to v247.

Combining positive and negative values is also valid:

$ meson configure build \
  -Dextra-net-naming-schemes=gargoyle=v241+nspawn_long_hash+-label_noprefix
It will yield:
NAMING_GARGOYLE = NAMING_V241|NAMING_NSPAWN_LONG_HASH & ~NAMING_LABEL_NOPREFIX

It is worth to mention that positive values will be grouped first (and ORed) and then
negative values together (and each one NOTed and ANDed).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -716,11 +716,20 @@ foreach scheme: get_option('extra-net-naming-schemes').split(',')
                 name = scheme.split('=')[0]
                 value = scheme.split('=')[1]
                 NAME = name.underscorify().to_upper()
-                VALUE = []
+                ENABLED_SCHEMAS = []
+                DISABLED_SCHEMAS = []
                 foreach field: value.split('+')
-                        VALUE += 'NAMING_' + field.underscorify().to_upper()
+                        if field.startswith('-')
+                                DISABLED_SCHEMAS += 'NAMING_' + field.strip('-').underscorify().to_upper()
+                        else
+                                ENABLED_SCHEMAS += 'NAMING_' + field.underscorify().to_upper()
+                        endif
                 endforeach
-                extra_net_naming_schemes += 'NAMING_@0@ = @1@,'.format(NAME, '|'.join(VALUE))
+                extra_net_naming_schemes += 'NAMING_@0@ = @1@'.format(NAME, '|'.join(ENABLED_SCHEMAS))
+                foreach disabled: DISABLED_SCHEMAS
+                        extra_net_naming_schemes += ' & ~@0@'.format(disabled)
+                endforeach
+                extra_net_naming_schemes += ','
                 extra_net_naming_map += '{ "@0@", NAMING_@1@ },'.format(name, NAME)
         endif
 endforeach


meson.build
```diff
@@ -716,11 +716,20 @@ foreach scheme: get_option('extra-net-naming-schemes').split(',')
                 name = scheme.split('=')[0]
                 value = scheme.split('=')[1]
                 NAME = name.underscorify().to_upper()
-                VALUE = []
+                ENABLED_SCHEMAS = []
+                DISABLED_SCHEMAS = []
                 foreach field: value.split('+')
-                        VALUE += 'NAMING_' + field.underscorify().to_upper()
+                        if field.startswith('-')
+                                DISABLED_SCHEMAS += 'NAMING_' + field.substring(1).underscorify().to_upper()
```
medhefgo: `substring()` is only available since 0.56.0. Dunno why meson doesn't complain, though. You should be able to do the same with `strip('-')`.

jacekmigacz: Good catch. Thanks you @medhefgo.

meson.build
```diff
@@ -716,11 +716,20 @@ foreach scheme: get_option('extra-net-naming-schemes').split(',')
...
+                DISABLED_SCHEMAS = []
                 foreach field: value.split('+')
-                        VALUE += 'NAMING_' + field.underscorify().to_upper()
+                        if field.startswith('-')
+                                DISABLED_SCHEMAS += 'NAMING_' + field.strip('-').underscorify().to_upper()
+                        else
+                                ENABLED_SCHEMAS += 'NAMING_' + field.underscorify().to_upper()
+                        endif
+                endforeach
+                extra_net_naming_schemes += 'NAMING_@0@ = @1@'.format(NAME, '|'.join(ENABLED_SCHEMAS))
```
medhefgo, line 728: Unless I get my binary ops wrong in my head, you could do this, no?
```
if DISABLED_SCHEMAS.length() > 0
    extra_net_naming_schemes += ' & ~(@0@)'.format('|'.join(DISABLED_SCHEMAS)))
```

jacekmigacz: @medhefgo, you got it right, but I wonder if it is any better? Mine have explicit loop over `disabled`, yours have implicit loop within `join()`. Do I overlook something? I'd love to hear your thoughts.

medhefgo: I just like to point out alternative ways things can be done. It isn't always strictly better. :D

keszybz:
Two questions:

Wouldn't it be nicer to just allow `-` as the separator? `-Dextra-net-naming-schemes=gargoyle=v239-slot_function_id-16bit_index-replace_strictly`. Or if we are worried about `-` looking like a part of the name,  `~` instead: `-Dextra-net-naming-schemes=gargoyle=v239~slot_function_id~16bit_index~replace_strictly`.

With the current implementation, negative entries are moved to the end in the generated expression. So they have higher "priority" and e.g. `v239+-slot_function_id+v239` is equivalent to `v239`. I guess that's actually OK: you would normally use this as taking one base value and then subtracting things from it, and I don't see a scenario where you'd want to add two base values. So I think such an order is OK, but it should be documented. And we should refuse positive entries if a negative entry was seen. By refusing such ambiguous entries we'll have freedom to extend this in a backwards-compatible manner in the future if we need to.

poettering:
Ping?

jacekmigacz:
There was no use of it in RHEL at the end. I'm closing it.

PR #20528, opened 2021-08-24
resolve: remove server 'large' level

ddstreet:
This removes the DNS_SERVER_FEATURE_LEVEL_LARGE, and sets the EDNS0
advertised max packet size as if always in 'large' mode.

Without this, we always send out EDNS0 opts that limit response sizes
to 512 bytes, thus the remote server will never send anything larger
and will always truncate responses larger than 512 bytes, forcing us
to drop from EDNS0 down to TCP, even though one of the primary benefits
of EDNS0 is larger packet sizes.

PR #22368, opened 2022-02-02
Revert "resolve: remove server 'large' level"

poettering:
Reverts systemd/systemd#20528

(Don#t merge this, this is mostly a placeholder, because I don't get #20528, and it doesn't look right to me. This should only be merged once there has been more discussion about what this is supposed to do)


ddstreet:
Hello!

Responding to [this comment](https://github.com/systemd/systemd/pull/20528#issuecomment-1027920509) from the original PR (#20528):


> The original idea was that we'd start out with "large" mode, and if that fails, will fall back to simple edns0 (and thus still get better error codes and stuff, but no large packets) mode. Now you basically removed the latter? Why though?

Because resolved uses a linear list of 'levels', nobody was actually using "large" mode, unless DNSSEC was enabled. For everyone with DNSSEC disabled, the "highest" level was simply normal EDNS0 with 512-byte packets.

> The thing is that whether edns0 works or not is primarily a question of the server implementation. While the packet size is a lot more a question of the gateways in between... Hence it makes sense to discern edns0 mode from large mode, and still to edns0 if the large stuff doesn't work.

The edns0 "large packet" size only controls the size of the *response* packet that the remote, upstream DNS server responds with - not the size of the request packet we send out. And the existing resolved code that detects fragmented packets will (and already did) limit our edns0 max size advertisement. However, the problem is we artificially limited our max size in "normal" edns0 mode to 512 bytes. As mentioned above, if DNSSEC is disabled, we will *never* send out a edns0 request with a max packet size larger than 512 bytes, so we would *never* get a response any larger than 512 bytes. We *must* send out a max packet size with a larger value to even detect fragmentation.

Now, with the "large" level removed and the "normal" edns0 level set to use the interface MTU size, the remote server should actually respond with a packet up to that MTU size, and if we detect fragmentation from that response, we'll automatically lower future edns0 request max response size when talking to that server.

keszybz:
@poettering ^ ?

pemensik:
Note that DNS community has made agreement on different size than MTU: https://dnsflagday.net/2020/#action-dns-software-vendors

EDNS packet from client notifies server how big packets it is willing to accept. Server can choose to send smaller packets than requested, but it should not try sending a bigger packet. Any compliant implementation would be fine with any value systemd can process. Sending 512B size does not make any sense. Justification by @ddstreet  makes sense, I don't understand why this should ever be reverted. It was wrong from the beginning.

keszybz:
Let's close this. There seems to be agreement that what we had before wasn't useful. So even if there are changes needed, a revert is not appropriate.

PR #21609, opened 2021-12-03
core: add StartUnitWithFlags DBUS method

bluca:
When an activation job is skipped because of a Condition*= setting failing,
currently the JobRemoved signal lists 'done' as the result, just as with
a successful job.

This is a problem when doing dbus activation: dbus-broker will receive a
signal that says the job was successful, so then it moves into a state where
it waits for the requested name to appear on the bus, but that never happens
because the job was actually skipped.

Add a new StartUnitWithFlags that lets a caller decide whether the JobRemoved
signal should list 'done' or 'skipped' via a flag.

Fixes #21520

~~Now, this is a backward-incompatible change, technically, as the `JobRemoved` signal return value changes. And there are many projects using this: https://codesearch.debian.net/search?q=JobRemoved&literal=1&perpkg=1 although some of them are already correctly expecting and handling a `skipped` result, even though in practice that can never happen right now for starting jobs (unless one tries to start a device unit).~~

~~Ideas? The obvious would be adding a new StartUnitWithFlags method I suppose, but that's quite heavy handed. Is it worth it?~~

EDIT: added a new method instead, with flags.

PR #22369, opened 2022-02-02
Revert "core: add StartUnitWithFlags DBUS method"

poettering:
Reverts systemd/systemd#21609

This is wrong I am sure. Needs a different fix in dbus-broker.

### DIFF
--- a/man/org.freedesktop.systemd1.xml
+++ b/man/org.freedesktop.systemd1.xml
@@ -74,10 +74,6 @@ node /org/freedesktop/systemd1 {
       StartUnit(in  s name,
                 in  s mode,
                 out o job);
-      StartUnitWithFlags(in  s name,
-                         in  s mode,
-                         in  t flags,
-                         out o job);
       StartUnitReplace(in  s old_unit,
                        in  s new_unit,
                        in  s mode,
@@ -774,8 +770,6 @@ node /org/freedesktop/systemd1 {
 
     <variablelist class="dbus-method" generated="True" extra-ref="StartUnit()"/>
 
-    <variablelist class="dbus-method" generated="True" extra-ref="StartUnitWithFlags()"/>
-
     <variablelist class="dbus-method" generated="True" extra-ref="StartUnitReplace()"/>
 
     <variablelist class="dbus-method" generated="True" extra-ref="StopUnit()"/>
@@ -1194,13 +1188,6 @@ node /org/freedesktop/systemd1 {
 
       <para><function>StartUnitReplace()</function> is similar to <function>StartUnit()</function> but
       replaces a job that is queued for one unit by a job for another unit.</para>
-
-      <para><function>StartUnitWithFlags()</function> is similar to <function>StartUnit()</function> but
-      allows the caller to pass an extra <varname>flags</varname> parameter, which does not support any
-      flags for now, and is reserved for future extensions. The new method also changes the behaviour
-      of the <varname>JobRemoved</varname> signal and make it return <literal>skipped</literal> in case
-      the unit activation job is skipped because a <varname>Condition*=</varname> is not satisfied.
-      With the <varname>StartUnit</varname> method, <literal>done</literal> would be returned instead.</para>
 
       <para><function>StopUnit()</function> is similar to <function>StartUnit()</function> but stops the
       specified unit rather than starting it. Note that the <literal>isolate</literal> mode is invalid for this
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -2799,15 +2799,6 @@ const sd_bus_vtable bus_manager_vtable[] = {
                                  SD_BUS_PARAM(job),
                                  method_start_unit,
                                  SD_BUS_VTABLE_UNPRIVILEGED),
-        SD_BUS_METHOD_WITH_NAMES("StartUnitWithFlags",
-                                 "sst",
-                                 SD_BUS_PARAM(name)
-                                 SD_BUS_PARAM(mode)
-                                 SD_BUS_PARAM(flags),
-                                 "o",
-                                 SD_BUS_PARAM(job),
-                                 method_start_unit,
-                                 SD_BUS_VTABLE_UNPRIVILEGED),
         SD_BUS_METHOD_WITH_NAMES("StartUnitReplace",
                                  "sss",
                                  SD_BUS_PARAM(old_unit)
--- a/src/core/dbus-unit.c
+++ b/src/core/dbus-unit.c
@@ -377,7 +377,6 @@ int bus_unit_method_start_generic(
                 bool reload_if_possible,
                 sd_bus_error *error) {
 
-        BusUnitQueueFlags job_flags = reload_if_possible ? BUS_UNIT_QUEUE_RELOAD_IF_POSSIBLE : 0;
         const char *smode, *verb;
         JobMode mode;
         int r;
@@ -406,23 +405,6 @@ int bus_unit_method_start_generic(
         else
                 verb = job_type_to_string(job_type);
 
-        if (sd_bus_message_is_method_call(message, NULL, "StartUnitWithFlags")) {
-                uint64_t input_flags = 0;
-
-                r = sd_bus_message_read(message, "t", &input_flags);
-                if (r < 0)
-                        return r;
-                /* Let clients know that this version doesn't support any flags at the moment. */
-                if (input_flags != 0)
-                        return sd_bus_reply_method_errorf(message, SD_BUS_ERROR_INVALID_ARGS,
-                                                          "Invalid 'flags' parameter '%" PRIu64 "'",
-                                                          input_flags);
-
-                /* The new method unconditionally uses the new behaviour of returning 'skip' when
-                 * a job is skipped. */
-                job_flags |= BUS_UNIT_QUEUE_RETURN_SKIP_ON_CONDITION_FAIL;
-        }
-
         r = bus_verify_manage_units_async_full(
                         u,
                         verb,
@@ -436,7 +418,8 @@ int bus_unit_method_start_generic(
         if (r == 0)
                 return 1; /* No authorization for now, but the async polkit stuff will call us again when it has it */
 
-        return bus_unit_queue_job(message, u, job_type, mode, job_flags, error);
+        return bus_unit_queue_job(message, u, job_type, mode,
+                                  reload_if_possible ? BUS_UNIT_QUEUE_RELOAD_IF_POSSIBLE : 0, error);
 }
 
 static int bus_unit_method_start(sd_bus_message *message, void *userdata, sd_bus_error *error) {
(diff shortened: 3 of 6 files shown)


bluca:
This interface is already part of a release, so it cannot be removed - apart from that:

> Uh, this looks wrong to me. JobRemoved() is sent to _all_ clients, not just the one who enqueued the job originally. In particular as jobs are shared by multiple clients if multiple enqueue the same one at the same time.
> 
> This should be reverted I am sure.
> 
> Services that declare that they'll take a bus name and then don#t may always exist. Whether a condition makes that happen or because they exit(0) early on before even taking a name doesn't matter, that's an implementation detail...
> 
> dbus-broker has to watch the unit there's no way around that. i.e. once the job finished it needs to check of the unit is running, and if not fail the thing or wait for a unit state change event indicating its stopped.
> 
> I mean, yeah it would have been wise to have a more specific set of values in JobRemoved, but that ship is sailed, and we don't need it here anyway.
> 
> Hence, I think we should revert this and dbus-broker should be improved to watch the service, and check if the service exited or even failed.

I don't see how else it could be made to work? The problem is that there's no way to differentiate. A failed or immediately exited job is one thing, but a skipped one is completely different, but the signal is the same. As a client, you don't have a way to tell the difference. The reaction needs to be different, because they are different situations.

bluca:
> Uh, this looks wrong to me. JobRemoved() is sent to all clients, not just the one who enqueued the job originally. In particular as jobs are shared by multiple clients if multiple enqueue the same one at the same time.

Also I'm not sure if this would be an actual problem in the real world? When I added this, I looked around in codesearch and clients watching for those were either filtering for the job id they added (so it doesn't matter, if they use the new flag they'll do the right thing) or didn't really care about this distinction, and were monitoring for failures explicitly.

poettering:
> Also I'm not sure if this would be an actual problem in the real world? When I added this, I looked around in codesearch and clients watching for those were either filtering for the job id they added (so it doesn't matter, if they use the new flag they'll do the right thing) or didn't really care about this distinction, and were monitoring for failures explicitly.

The thing is you cannot actually wait for the job id you enqueued sanely on D-Bus, it's a design deficiency of D-Bus. That's because the StartUnit() call will return a job object path for you but you can only subscribe to its changes once you get the answer, but that might be quite late. This means, if you want all state changes you need to subscribe generically before you issue the StartUnit(), and thus before you know the job id, and thus you have to wait for all job events really...

bluca:
> > I don't see how else it could be made to work? The problem is that there's no way to differentiate. A failed or immediately exited job is one thing, but a skipped one is completely different, but the signal is the same. As a client, you don't have a way to tell the difference. The reaction needs to be different, because they are different situations.
> 
> dbus-broker should not just issue the start job, but also watch the unit itself. i.e. a unit that reports a successful start but then doesn't actually run is a problem either way. And it doesn't matter if that's because a configured condition check failed and the job was skipped, or because the services' own code just did an exit(0) quite soon: either way it's a problem.
> 
> Example: if a service quux.service shall be activated, dbus-broker should
> 
> 1. Start watching the quux.service for state changes
> 2. Enqueue a start job for the service
> 3. once the start job fails → activation failed
> 4. once the start job success, and the sevice's state changes to dead or failed, → activation failed
> 5. once an actviation timeout dbus-broker maintains is hit → activation failed
> 6. once the bus name is taken → activation suceeded

It does all of that. But the timeout obviously has to be very long, and thus the user experience is abysmal. Also it muddles the root cause and causes confusion. A skipped condition is not a failure, it's "working as intended", and we treat it as such all across our code base and user interface - but due to this discrepancy in the job reporting, we force external users to do the opposite. This is just bad user interface and user experience. We have the correct information, we have it immediately and we have the means to optionally (opt-in) convey it, so why shouldn't this whole bad situation be improved?

> > Also I'm not sure if this would be an actual problem in the real world? When I added this, I looked around in codesearch and clients watching for those were either filtering for the job id they added (so it doesn't matter, if they use the new flag they'll do the right thing) or didn't really care about this distinction, and were monitoring for failures explicitly.
> 
> The thing is you cannot actually wait for the job id you enqueued sanely on D-Bus, it's a design deficiency of D-Bus. That's because the StartUnit() call will return a job object path for you but you can only subscribe to its changes once you get the answer, but that might be quite late. This means, if you want all state changes you need to subscribe generically before you issue the StartUnit(), and thus before you know the job id, and thus you have to wait for all job events for _every_ job, even if you are only interested in one specific job...

Yeah but you discard it later anyway. This (including the dbus-broker change) has been in use on Debian testing and unstable for almost a month, and got no complaints so far.

bluca:
> > It does all of that. But the timeout obviously has to be very long, and thus the user experience is abysmal.
> 
> Hmm? it doesn't have to be long?
> 
> i.e. key is that once the job is done but the unit is not running, that's an immediate reason to consider the activation to have failed. That's really all I am saying.

There's no upper bound on these interactions - eg, the system is low-powered and under load, so the default has to be lax. Even if there was, it's still wrong: the activation succeeded, and if you check the unit status it won't be in the failed state. It was skipped, which is different, and it's treated differently, but it's hidden only in this reporting channel, which results in a bad end-to-end flow.

bluca:
> > > 
> > 
> > 
> > There's no upper bound on these interactions - eg, the system is low-powered and under load, so the default has to be lax. Even if there was, it's still wrong: the activation succeeded, and if you check the unit status it won't be in the failed state. It was skipped, which is different, and it's treated differently, but it's hidden only in this reporting channel, which results in a bad end-to-end flow.
> 
> Hmm? conditions are checked only if a service is not running yet (i.e. is in "dead" or "failed" state), and there was a request to start it via a job. If the condition fails then the actual starting will be skipped, so that unit is still in either "dead" or "failed".
> 
> Thus, if a job succeeded but the unit afterwards is still "dead" or "failed" then this can mean two things:
> 
> 1. the condition failed, and had no effect (i.e. the suggestion described above)
> 2. the condition succeeded, the service was started but immediately issued "exit(0)" for some reason. i.e. job _and_ service claimed success, but things are not in order anyway.
> 
> Regardless which of the two cases it is: from dbus-broker's pov the activation failed...
> 
> It can check this all immediately, no need to wait for the activation to time out. The time out check is only two catch a third case:
> 
> 3. the condition succeeded, the service was started, and is running but never actually takes the name its assumed to take on the bus.

Yes, and being able to tell the difference between these cases is a good thing - right now it's impossible, and you have to basically guess, or parse the logs. I don't understand why optionally giving out the accurate result is a problem? What is the issue exactly?

poettering:
the problem is api compat, that's all. accepting a new enum value on input that we didn't support before is an OK thing to do, but sending out a new enum value in one of our primary interfaces is a bit of a problem. in particular as previously clients could assume everything that isn't "success" is "failure", but with this change they can't do that anymore, there are multiple values for "success".

Yes, if this was a new interface we'd not hide the difference, but it's an established interface

bluca:
> the problem is api compat, that's all. accepting a new enum value on input that we didn't support before is an OK thing to do, but sending out a new enum value in one of our primary interfaces is a bit of a problem. in particular as previously clients could assume everything that isn't "success" is "failure", but with this change they can't do that anymore, there are multiple values for "success".
> 
> Yes, if this was a new interface we'd not hide the difference, but it's an established interface

It's opt-in - default behaviour is the same as before.

poettering:
it's not though. If I write a tool that watches any job happening on the bus I will see it. JobRemoved is broadcast after all...

Tools like that do exist, for example the old systemd-ui package from back in the day does it.

Moreover: jobs are merged if multiple clients enqueue the same job at the same time. if one does this without the flag and one with then one of them will not get the behaviour it wants...

poettering:
to salvage this: clients are supposed to call Subscribe() before watching signals. Its used to suppress sending out signals if noone listens. You could extend as SubscribeWithFlags(), and enable it there. With that it would be he receiver of the signals who could control the flag.

poettering:
hmm, Subscribe() wouldn't work though that easily, as the signals are *broadcast* after all, not unicast, thus if we send out the JobRemoved() signal then in the worst case if there are both clients that called Subscribe() with and without the flag, then we'd have to unicast the signal to each one of them to ensure they get the right value.

So that would suck codewise.

i still don't see why bother. dbus-broker should check the unit status, as suggested, it covers more ground, and if people want to know why the activation failed precisely there are always the logs...

bluca:
> it's not though. If I write a tool that watches any job happening on the bus I will see it. JobRemoved is broadcast after all...
> 
> Tools like that do exist, for example the old systemd-ui package from back in the day does it.

I have not seen any such tool that would care about this change. It has been in Debian for a month.

> Moreover: jobs are merged if multiple clients enqueue the same job at the same time. if one does this without the flag and one with then one of them will not get the behaviour it wants...

There is only one D-Bus service on a system activating units like that, and it either uses the flag or not, it will not mix and match.

poettering:
> There is only one D-Bus service on a system activating units like that, and it either uses the flag or not, it will not mix and match.

Are you saying only the D-Bus daemon can set the flag?

poettering:
> > 
> 
> I have not seen any such tool that would care about this change. It has been in Debian for a month.

Uh, Debian unstable might help weeding out obvious incompat problems, but it's not remotely comparison what RHEL will eventually reveal.

poettering:
> I'm saying that nothing else will be starting the same units that dbus-broker is in the same way, because activation via dbus is strictly managed by the dbus daemon, it's not something that other services on the bus do, so there shouldn't be any conflicts, no? So as things stand now there shouldn't be issues, and we can of course improve the documentation of the new flag so that future unknown use cases are aware of any gotchas.

But the its a generic flag, anyone can set it and if they do things fall apart. Moreover if anyone watches JobRemoved() signals happening they'll see the unexpected value for dbus-broker's activations too.

JobRemoved is *broadcast*, so anyone will get it who wants them, not just the ones who might have requested the job.

Activation doesn't imply ownership. Services can be defined to be activated in five different ways at the same time. i.e. via socket, via bus, or anything else really. networkd for example can be activated via socket and via dbus, for whatever comes first. And users can also start it manually if they like. Thus job merging can and will happen.

So yes, this is a compat break.

bluca:
We currently already return 'skipped' in one case:

https://github.com/systemd/systemd/blob/ff7e7c2b3a2b4367e0a03a51b7d714dfe0837abb/src/core/job.c#L894
https://github.com/systemd/systemd/blob/ff7e7c2b3a2b4367e0a03a51b7d714dfe0837abb/src/core/unit.c#L1842

which if I remember correctly happens because the request was to start an un-startable unit, so technically any generic watch-everything-even-jobs-from-other-callers tool already needs to handle 'skipped' correctly, because it can already happen.

But regardless, (doing my best corporate impression): let's take a step back and forget about the current implementation for a second - dbus-broker is just the first use case, I am going to make heavy use of Condition*= stuff, so clear, immediate and program-friendly feedback on whether a job was skipped instead of started is important to me, as there's a logical difference that requires different handling from an orchestrator.
How could this be implemented?

bluca:
As discussed in the conclave, more limited change that keeps the new interface but makes it behave like the old one for now: https://github.com/systemd/systemd/pull/22682

I'll then work to add a new method for a new signal type separately.

poettering:
Withdrawing in favour of #22682

PR #22372, opened 2022-02-02
udevadm: don't claim a sysattr was write only just because we get EPE…

poettering:
…RM on read

it might be totally inaccessible, hence weaken our language a bit and
just say "not readable", thus making not claim whether it is writable or
not.

Follow-up for 3a90bef55a5a71a03629a762470b0f9014cd8453


yuwata (approved):
LGTM

PR #22373, opened 2022-02-02
journal: Rename JournaldFile to ManagedJournalFile

daandemeyer:
JournalFile and JournaldFile are hard to distinguish from each other.
Let's use ManagedJournalFile instead to make the distinction more clear.


anitazha (approved):
looks like a straight forward rename to me

PR #22079, opened 2022-01-11
shutdown: release the watchdog finally

ldzhong:
During reboot or kexec, watchdog will be setup before systemd is
replaced by systemd-shutdown binary. But while systemd-shutdown is
executed, the watchdog will not really be released. Log shown as
following:

[   39.371202] watchdog: watchdog0: watchdog did not stop!
[   39.383659] systemd-shutdown[1]: Syncing filesystems and block
devices.
[   39.423727] systemd-shutdown[1]: Sending SIGTERM to remaining
processes...
[   39.440857] systemd-journald[526]: Received SIGTERM from PID 1
(systemd-shutdow).
[   39.461047] systemd-shutdown[1]: Sending SIGKILL to remaining
processes...
[   39.466736] systemd-shutdown[1]: Using hardware watchdog 'iTCO_wdt',
version 0, device /dev/watchdog
[   39.467835] systemd-shutdown[1]: Unmounting file systems.
[   39.469721] [2171]: Remounting '/' read-only in with options
'attr2,inode64,logbufs=8,logbsize=32k,noquota'.
[   39.550266] systemd-shutdown[1]: All filesystems unmounted.
[   39.550274] systemd-shutdown[1]: Deactivating swaps.
[   39.550443] systemd-shutdown[1]: All swaps deactivated.
[   39.550449] systemd-shutdown[1]: Detaching loop devices.
[   39.550697] systemd-shutdown[1]: All loop devices detached.
[   39.550703] systemd-shutdown[1]: Detaching DM devices.
[   39.551269] systemd-shutdown[1]: Not all DM devices detached, 1
left.
[   39.551277] systemd-shutdown[1]: Cannot finalize remaining DM
devices, continuing.
[   39.580044] systemd-shutdown[1]: Successfully changed into root
pivot.
[   39.580050] systemd-shutdown[1]: Returning to initrd...
[   39.587921] watchdog: watchdog0: watchdog did not stop!
[   39.774130] dracut Warning: Break before pre-shutdown

after applying fix

[   59.520379] watchdog: watchdog0: watchdog did not stop!
[   59.542286] systemd-shutdown[1]: Syncing filesystems and block
devices.
[   59.583313] systemd-shutdown[1]: Sending SIGTERM to remaining
processes...
[   59.600133] systemd-journald[517]: Received SIGTERM from PID 1
(systemd-shutdow).
[   59.624435] systemd-shutdown[1]: Sending SIGKILL to remaining
processes...
[...]

PR #22375, opened 2022-02-02
fix watchdog handling during reboot

michaelolbrich:
Since #22079 the watchdog is stopped during shutdown. This is wrong. The watchdog should remain enabled until the system resets. Fix that in a way that avoids the kernel message.

Debugging this revealed that `watchdog_set_device()` does not work correctly. This is fixed as well.

Fixes #22079

### DIFF
--- a/src/shutdown/shutdown.c
+++ b/src/shutdown/shutdown.c
@@ -531,8 +531,9 @@ int main(int argc, char *argv[]) {
                           need_md_detach ? " MD devices," : "");
         }
 
-        /* We're done with the watchdog. */
-        watchdog_close(true);
+        /* We're done with the watchdog. Note that the watchdog is explicitly not
+         * stopped here. It remains active to guard against any issues during the
+         * rest of the shutdown sequence. */
         watchdog_free_device();
 
         arguments[0] = NULL;


src/shared/watchdog.c
```diff
@@ -133,6 +133,8 @@ static int open_watchdog(void) {
         fn = !watchdog_device || path_equal(watchdog_device, "/dev/watchdog") ?
                 "/dev/watchdog0" : watchdog_device;
 
+        watchdog_set_device(fn);
```
anitazha: Do you also need to call `watchdog_free_device()` in the error paths?

poettering: if `fn` is `watchdog_device` anway we'll strdup() it and free() the old identical string. Also the call can fail due to OOM. Maybe just do this instead:

```c
if (!streq_ptr(fn, watchdog_device)) {
        /* Keep track of the device we are actually opening */ 
        r = free_and_strdup(&watchdog_device, fn);
        if (r < 0)
                return log_oom_debug();
}
```

src/shutdown/shutdown.c
```diff
@@ -531,10 +531,6 @@ int main(int argc, char *argv[]) {
                           need_md_detach ? " MD devices," : "");
         }
 
-        /* We're done with the watchdog. */
-        watchdog_close(true);
-        watchdog_free_device();
```
poettering, line 536: I think we should keep the watchdog_free_device() in place, so that we release all resources. I mean, the very likely case is that we'll execve() very soon anyway, and that will close the fd anyway.

The kernel watchog API is explicitly document to leave the watchdog "hot" if you just close the device. If you want to disarm it you have to write "V" to it, iirc first, and we do that whenever we want to close the device without leaving it armed. hence: please drop the watchdog_close(), i.e. revert https://github.com/systemd/systemd/pull/22079 please, but do not also drop watchdog_free_device()...

In general it's nice if our code remains "valgrind" clean, even if IRL this doesn#t matter. But it makes our codebase easier to manage and validate and test. 

As far as I can see the only benefit of removing that line is that in case no "exitrd" is used (i.e not on the big distros) that one kernel log line is suppressed, but that sounds like a pretty weak improvement to me. I'd rather see it removed from the kernel, given we knowingly leave the watchdog armed here (well, at least until #22079 was merged, but that should be reverted, as said).

We should also add a comment here, explaining things, so that this doesn't happen again.

poettering:
Yes, I agree that #22079 should be reverted. But I don't think we should leave the watchdog fd open, see my comment above. And I would love it some comments would be to the code so this doesn't happen again...

src/shared/watchdog.c
michaelolbrich: `free_and_strdup()` already does a `streq_ptr()` check internally. If the string matches, then it does nothing.

I changed it to call `free_and_strdup()` anyways and added the oom check.

src/shutdown/shutdown.c
michaelolbrich: Changed it that way and added a comment.

src/shared/watchdog.c
michaelolbrich: We only set a new string here if no device has been set before. So removing it would have to be conditional.
And the string is freed anyways in `watchdog_free_device()` later. So it does not leak.
Adding anything here is just extra complexity with no real benefit.

poettering:
excellent. thanks! lgtm

src/shared/watchdog.c
fbuihuu: then I'm not sure to understand why you didn't stick with `watchdog_set_device(fn);` instead... (with the return value check thought).

Or maybe make it possibly slightly clearer by calling `watchdog_set_device()` only when fn is "dev/watchdog0".

poettering: then I'm not sure to understand why you didn't stick with watchdog_set_device(fn); instead... (with the return value check thought).

you could, but the closing of the fd isn't really necessary then. With @michaelolbrich' current version you have exactly one function call with error handling. if this would use watchdog_set_device() instead you'd have the exact same number of function calls, and it would do stuff we don't really need.

PR #20833, opened 2021-09-24
service: pass exiting service state to triggered On{Failure,Success}= dependencies

pdmorrow:
This PR adds support for passing exit status and other metadata to dependencies triggered via OnFailure= and OnSuccess=. The following environment variables are passed to the ```OnFailure=``` service ```ExecStart=``` and ```ExecStartPre=``` processes:

   * $MONITOR_SERVICE_RESULT
   * $MONITOR_EXIT_CODE
   * $MONITOR_EXIT_STATUS
   * $MONITOR_INVOCATION_ID
   * $MONITOR_UNIT

The first 3 variables are the same as those passed to ExecStop= and ExecStopPost= processes, however the values are related to the service which triggered the OnFailure=/OnSuccess= dependency.

The last 2 variables are new and as their names suggested are used to pass the invocation ID and unit name of the exiting service to the OnFailure= or OnSuccess= dependency.

This is useful for building more complex failure monitoring services.

Fixes #19885

PR #22378, opened 2022-02-02
Follow-ups for #20833

bluca:


src/core/service.c
```diff
@@ -1482,36 +1482,29 @@ static int service_create_monitor_md_env(Job *j, char **ret) {
...
 
-                if (first) {
+                if (first)
                         /* Add the environment variable name first. */
-                        r = strextendf(&var, "MONITOR_METADATA=");
-                        if (r < 0)
-                                return r;
-
-                }
+                        if (!strextend(&var, "MONITOR_METADATA="))
```
medhefgo: `first && !strextend`?

src/core/service.c
```diff
@@ -1482,36 +1482,29 @@ static int service_create_monitor_md_env(Job *j, char **ret) {
...
                                                env_source->main_exec_status.status);
-                        else
-                                r = strextendf(&var, ",EXIT_STATUS=%s",
-                                               signal_to_string(env_source->main_exec_status.status));
-                        if (r < 0)
-                                return r;
+                                if (r < 0)
+                                        return r;
+                        } else
+                                if (!strextend(&var, ",EXIT_STATUS=", signal_to_string(env_source->main_exec_status.status)))
```
medhefgo: Maybe turn this into a proper else if (to reduce indentation)?

src/core/service.c
bluca: generally I find it more legible to avoid mixing logic checks and error checks, but it's not too important, fixed

PR #22380, opened 2022-02-02
sd-boot: rename LOADER_STUB → LOADER_UNIFIED_LINUX

poettering:
No actual code changes, just renaming.

Rationale: the focus here should be on the fact that these are "unified"
images, whether our stub is used or not, or something else doesn't
really matter. Moreover, these are still Linux entries. Hence, emphasize
that these are *unified* images, and *Linux* images, and deemphesize
that our sd-stub is likely used.

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -44,8 +44,8 @@ enum loader_type {
         LOADER_UNDEFINED,
         LOADER_AUTO,
         LOADER_EFI,
-        LOADER_LINUX,
-        LOADER_STUB,
+        LOADER_LINUX,         /* Boot loader spec type #1 entries */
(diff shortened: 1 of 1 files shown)


medhefgo (approved):
Don't bait me like that. I thought you were gonna rename the whole stub itself. What a letdown. :crying_cat_face:

bluca:
```
../src/boot/efi/boot.c: In function ‘config_entry_add_linux’:
../src/boot/efi/boot.c:2199:33: error: ‘LOADER_STUB’ undeclared (first use in this function); did you mean ‘LOADER_AUTO’?
 2199 |                                 LOADER_STUB,
      |                                 ^~~~~~~~~~~
      |                                 LOADER_AUTO
```

yuwata:
Force pushed with one more LOADER_STUB converted. Setting the green label.

PR #22381, opened 2022-02-02
tests: fuzz client_send_message

evverx:
to make sure outgoing packets based on incoming packets are fine.
It's just another follow-up to
https://github.com/systemd/systemd/pull/10200.
Better late than never :-)

@yuwata could you take a look?


yuwata:
Is it not possible to reuse fuzz-dhcp6-client?

evverx:
> Is it not possible to reuse fuzz-dhcp6-client?

It should be possible to reuse it in theory but since it does too much already it would take it a lot of time to even trigger the code paths where outgoing packets are produced. This fuzzer targets that part of the code specifically.

yuwata:
Not sure I can follow, but ok. Introducing more tests is not bad in general.

evverx:
> Introducing more tests is not bad in general.

Agreed :-)

Looking at https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1795921 (which was opened less than a week after https://github.com/systemd/systemd/pull/10200 was opened) I suspect a fuzz target like that was used to discover that particular issue.

PR #22382, opened 2022-02-03
tests: pass FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION to fuzzers

evverx:
to let them use reproducible identifiers, which should make it possible
to really use files copied from OSS-Fuzz to reproduce issues on
GHActions and locally. Prompted by https://github.com/systemd/systemd/pull/22365.


yuwata (approved):

mrc0mmand (approved):
LGTM, thanks!

PR #11223, opened 2018-12-20
fileio: fix read_one_line() when reading bytes > 0x7F

poettering:
Fixes: #11218

PR #22387, opened 2022-02-03
boot: fix pretty nasty typo

poettering:
Note to self: should I ever become a black hat hacker, this is how I'd
hide a vulnerability. ;-)

(No, this one is not a vulnerability)

Follow-up for: 163d1ea5dd69c75d889ba4c2d1b51432ed14127f

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -2446,7 +2446,7 @@ static void config_load_all_entries(
                                       reboot_into_firmware);
 
         if (config->entry_count == 0)
-                return
+                return;
 
         config_write_entries_to_variable(config);
 


poettering:
(sad, that none of the static checkers noticed that alignment didn't match what this does. I mean, gcc does check alignment these days, but not this way apparently)

bluca (approved):

evverx:
Judging by https://pvs-studio.com/en/docs/warnings/v5001/ pvs-studio could have caught this I think. I played with it back in 2018 in https://github.com/systemd/systemd/pull/11223#discussion_r243946461.

mrc0mmand:
> Judging by https://pvs-studio.com/en/docs/warnings/v5001/ pvs-studio could have caught this I think. I played with it back in 2018 in [#11223 (comment)](https://github.com/systemd/systemd/pull/11223#discussion_r243946461).

FTR, I filed https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104367 to check if this couldn't be handled by gcc/clang as well.

medhefgo (approved):
Yikes

PR #22388, opened 2022-02-03
Revert "boot-timestamps: Discard firmware init time when running in a VM"

poettering:
Reverts systemd/systemd#22063

(Don't merge. This is just a placholder, because I think this needs more explanation.)

To me this appears to be a bug in the VM's firmware, which apparently doesn't reset the firmware reported timestamps on reboot. Given the firmware/VM implementation is open source, this sounds like something to fix there.

Taping over this, and rubbishing the reported values for *all* VMs is a bit drastic, since they might not suffer by this bug.

Generally, I don't think the times reported should be assumed to come with more guarantees that we could ever provide: they are as good as the firmware, and if the firmware is buggy the times will be off, but given this is mostly informational I thik we can live with that.

### DIFF
--- a/src/shared/boot-timestamps.c
+++ b/src/shared/boot-timestamps.c
@@ -5,13 +5,11 @@
 #include "efi-loader.h"
 #include "macro.h"
 #include "time-util.h"
-#include "virt.h"
 
 int boot_timestamps(const dual_timestamp *n, dual_timestamp *firmware, dual_timestamp *loader) {
         usec_t x = 0, y = 0, a;
         int r;
         dual_timestamp _n;
-        bool use_firmware = true;
 
         assert(firmware);
         assert(loader);
@@ -26,10 +24,6 @@ int boot_timestamps(const dual_timestamp *n, dual_timestamp *firmware, dual_time
                 r = efi_loader_get_boot_usec(&x, &y);
                 if (r < 0)
                         return r;
-
-                /* If we are running in a VM, the init timestamp would
-                 * be equivalent to the host uptime. */
-                use_firmware = detect_vm() <= 0;
         }
 
         /* Let's convert this to timestamps where the firmware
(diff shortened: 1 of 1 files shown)


poettering:
Anyway, @medhefgo why block this for all VMs? Why do you think this a systematic issue with VMs, and not just a bug that should be fixed in the firmware of one specific (open source) implementation of a vm supervisor?

medhefgo:
The timestamps we produce in sd-boot are based off of TSC provided by the cpu. Whether the hypervisor/vm should be emulating that to 0 at vm start is beyond my knowledge. I wouldn't even know if it actually can.
This just seemed like the simplest solution to me.

You may want to ask some qemu folks.

poettering:
Hmm. Ok, ill accept that. But maybe we should suppress the TSC values in VMs then, so that only that is disabled then.

medhefgo:
> Hmm. Ok, ill accept that. But maybe we should suppress the TSC values in VMs then, so that only that is disabled then.

That's what it does, no? It calls `detect_vm()` for this.

poettering:
i.e. I see that TSC virtualization is something some CPus can do and some cannot, hence it probably does make sense to not use TSC data in VM environments. A fix that makes sd-boot check if we run in a VM and if so not use the TSC should be pretty easy I think. That way systemd can do what it always does, but sd-boot filters out the bad data lready. And if firmware provides us with ACPI data we'll still use that in any case.

poettering:
> That's what it does, no? It calls `detect_vm()` for this.

detect_vm() does a lot more than just check whether we run in a VM, it also checks in which. And I think we should do this check in sd-boot, and not provide the bogus data in the first place, instead of filtering it late. Moreover, there are also the ACPI performance timers on newer firmware, and we should never filter that.

medhefgo:
The check is only in the branch where no ACPI performance data is available.

poettering:
still, we should filter this where we use the TSC, not at a very different place where we then consume it. if the data is crap it shouldn't be generated, and advertised in the first place.

poettering:
I mean the fix is probably easy, just add something like this to rdtsc():

```c
        uint32_t eax, ebx, ecx, edx;
       
        /* http://lwn.net/Articles/301888/ */

        /* First detect whether there is a hypervisor */
        if (__get_cpuid(1, &eax, &ebx, &ecx, &edx) == 0)
                return VIRTUALIZATION_NONE;

        if (ecx & 0x80000000U)
                return 0;
```

medhefgo:
I only explained why I did this, do whatever you want here, it doesn't really matter to me.

poettering:
there's something weird with read_ticks() btw. It returns 1 on archs where it is not supported, but time_usec() appears to think it returns 0 where not supported...

poettering:
Closing this one, replaced by #22446

PR #22391, opened 2022-02-03
Revert "boot: Change boot entry sorting"

poettering:
Reverts systemd/systemd#22066

(This is mostly a revert, so that this is discussed for now. Don#t merge right-away)

See my comments on the original PR. I am not convinced that we should order by human readable strings. We want that developers can pick the ids freely without this spilling needlessly into the human readable UI, and vice versa: allow developers to pick humarn readable titles freely without this having meaning beyond display.

### DIFF
--- a/man/systemd-boot.xml
+++ b/man/systemd-boot.xml
@@ -501,10 +501,10 @@
     considered 'good' from then on.</para>
 
     <para>The boot menu takes the 'tries left' counter into account when sorting the menu entries: entries in 'bad'
-    state are ordered towards the end of the list, and entries in 'good' or 'indeterminate' towards the beginning.
-    The user can freely choose to boot any entry of the menu, including those already marked 'bad'. If the menu entry
-    to boot is automatically determined, this means that 'good' or 'indeterminate' entries are generally preferred as
-    boot entries are tried in sort order, and 'bad' entries will only be considered if there are no 'good' or
+    state are ordered at the beginning of the list, and entries in 'good' or 'indeterminate' at the end. The user can
+    freely choose to boot any entry of the menu, including those already marked 'bad'. If the menu entry to boot is
+    automatically determined, this means that 'good' or 'indeterminate' entries are generally preferred (as the bottom
+    item of the menu is the one booted by default), and 'bad' entries will only be considered if there are no 'good' or
     'indeterminate' entries left.</para>
 
     <para>The <citerefentry><refentrytitle>kernel-install</refentrytitle><manvolnum>8</manvolnum></citerefentry> kernel
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -1511,6 +1511,7 @@ static void config_entry_add_from_file(
 
         entry->device = device;
         entry->id = xstrdup(file);
+        StrLwr(entry->id);
 
         config_add_entry(config, entry);
 
@@ -1622,24 +1623,13 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
         assert(a);
         assert(b);
 
-        /* Order entries that have no tries left towards the end of the list. They have
-         * proven to be bad and should not be selected automatically. */
+        /* Order entries that have no tries left to the beginning of the list */
         if (a->tries_left != 0 && b->tries_left == 0)
-                return -1;
-        if (a->tries_left == 0 && b->tries_left != 0)
                 return 1;
+        if (a->tries_left == 0 && b->tries_left != 0)
+                return -1;
 
-        r = strcasecmp_ptr(a->title ?: a->id, b->title ?: b->id);
-        if (r != 0)
-                return r;
-
-        /* Sort by machine id now so that different installations don't interleave their versions. */
-        r = strcasecmp_ptr(a->machine_id, b->machine_id);
-        if (r != 0)
-                return r;
-
-        /* Reverse version comparison order so that higher versions are preferred. */
-        r = strverscmp_improved(b->version, a->version);
+        r = strverscmp_improved(a->id, b->id);
         if (r != 0)
                 return r;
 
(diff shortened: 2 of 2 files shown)


medhefgo:
Sorting before this was just broken. People expect their list to be sorted by what they see and not by some unseen (possibly cryptic) filename. It just should not have any bearing on the order unless it's a last resort.

Do you really want "arch-lts.conf" to come before "arch.conf"? If you have multiple installations of the same distro you'll get name clashes unless you add something like a machine-id to the filename, but then the sorting order gets quite literally random.

> See my comments on the original PR. I am not convinced that we should order by human readable strings. We want that developers can pick the ids freely without this spilling needlessly into the human readable UI, and vice versa: allow developers to pick humarn readable titles freely without this having meaning beyond display.

If we sort by that id, it quite literally spills into the UI when it should have no bearing on it. Quite frankly, lexicographical sorting of entries is a thing of the 80s. We've had the computing power to sort things by their content/meaning for ages, let's make use of it.

@keszybz What's your opinion?

medhefgo:
> I am not sure I follow. The thing is that the order is not just a visual order, it is first and foremost actually the algorithm that automatically picks what to boot next. i.e. the goal here is not primraily being nice on the eye, but making sense programmatically. Showing the sd-boot menu is supposed to be the "adavanced UI" that you only see if the automatic choice doesn't work. i.e. sd-boot by default doesn't show any UI if at all possible....

And that still happens, based on the logical ordering from the entry *contents* and whatever was chosen as default pattern. Admins and tooling should have been creating entries with *rich* information so that we can produce a proper list anyways.

poettering:
Hmm? Wha? You are saying the display order is now different from the order of precedence what sd-boot automatically would boot into? That's confusing!

medhefgo:
It literally selects/boots in the the order things are displayed now. Whereas before, the entries were selected bottom-to-top.

anitazha:
I have no skin in this game but putting the bad entries at the beginning and good/indeterminate entries at end is pretty counter intuitive.

mbiebl:
Right, I think the the current ordering is much more logical / what one would expect.

poettering:
> It literally selects/boots in the the order things are displayed now. Whereas before, the entries were selected bottom-to-top.

I think we are talking about different issues here? I think sorting by human readable string is a bad idea. I am fine with adjusting the display order of the list as long as we base it on the machine readable strings, i.e. the id and the version. if we order backwards or forwards, and which key precisely we give precedence to is a different question from whether suddenly to make priority choices based on human-facing strings.

i.e. the id names *exist* only to have human-language-independent identifiers for sorting, priorization, and passing along to other layers of the stack. If you suddenly use the human readable strings instead you break all that, because suddenly the ids matter much less anymore, because programs have to order their entries based on the human facing strings...

to give you an example: i think it should be absolutely OK for admins/developers to create snippets they call 10-foo.conf, 20-bar.conf, 30-waldo.conf, 40-quux.conf, and expect them to appear in the UI in exctly this order, and then use title strings in them that are just "Foo", "Bar", "Waldo", "Quux". 

With your change this would not work anymore, suddenly 20-bar would be displayed before 10-foo, and 40-quux before 30-waldo. And that's just borked...

So, maybe I don't understand what you are drying to adress specifically? But basing order on *display* names, i.e. *human-readable* names is the wrong approach.

I mean, if you ran into some snippets where the human readable title and the file name totally badly relate to each other then that appears to be something to fix in that snippet/generator, but not to break the clear distinction between what strings we base our algorithms on and which strings we use for display.

medhefgo:
I'll stand by my point that sorting should be by content and not by filename. Why should an automatic entry like macOS come after a Xubuntu, for example? Imho, it would make more sense to add a `sort-key` option to entries to let people with OCD have their way.

Ultimately, if you cannot be reasoned with on this, at least keep the sane sort ordering of higher priority first instead of reverting this completely.

poettering:
> I'll stand by my point that sorting should be by content and not by filename. Why should an automatic entry like macOS come after a Xubuntu, for example? 

Because we are a Linux boot loader, and explicit stuff always should take precedence over automatically discovered stuff.

> Imho, it would make more sense to add a `sort-key` option to entries to let people with OCD have their way.

Making this configurable is chickening out...

Again: if the order of the title strings picked in snippets is systematically different from the order the filenames of the snippets would suggest, then that feels like a bug in those snippets/the generator for those snippets.

Anyway, it appears the PR made two changes: it switched to using the human readable title string instead of the filename (which as I tried to make clear I think is something we should revert). And it changed the sorting order in some other ways, reversing it so that newest is up-top and oldest at the end? Or what did it do? I have no issue with such a change.

But I am not following entirely, the PR did multiple things at once, and I don#t grok them all.

medhefgo:
See https://github.com/systemd/systemd/pull/22066#issuecomment-1008328101 for an example. Not that the list is lowest to highest priority and in particular, the totally unintuitive (boot-sorted and considered bad) entry `loader4+0.conf`.

poettering:
Moreover, the userspace parser in src/shared/bootspec.[ch] is currently working quite differently from the EFI parser of the same thing regarding sorting order. This needs addressing too.

poettering:
So, I looked into this in closer detail, and I think we should simply revert the whole thing. 

So I still think, sorting by human readable name is simply wrong. I understand the intention of @medhefgo's patch was to sort first alphabetically upwards by entry name and then downwards by entry version. So that if you have three OS "FooOS 0.5", "FooOS 0.6", "AardvarkOS 0.3" you end up with the order "AardvarkOS 0.3", "FooOS 0.6", "FooOS 0.5". (i.e. names go up, and version go down in order) But the simple fact is that we can't do that cleanly right now. i.e. the entry directory names are just strings we compare with a whole with strverscmp(), thus with the current naming regime kernel-install picks we cannot order upwards be OS name and downwards by version because it's all just one string to us.

So without changes to the snippets we cannot really implement the order @medhefgo is looking for (and I agree with @medhefgo on this, it would be better to order by OS name first upwards and then by version down). There are two ways out:

1. change the naming regime for the snippets so that the .conf filenames start with an OS id of some form, i.e. the ID= field of os-release. That way archlinux-9eb1ff8a0b134a608a1a299bc32201a5-5.16.0-0.rc2.18.fc36.x86_64.conf would be ordered before fedora-b870ba0da05642da8aca95b1ba25251b-5.15.0-60.fc36.x86_64.conf as probably makes sense. This change woudl be hard though, as we'd compat would be hard: after all after updating kernel-install we still should be ablet to find the old entry in case "kernel-install remove" is called.

2. add a new key to the files that is expressly intended for sorting. i.e. we call it "sort-key" or so. If set we'd sort the entry by this key first, upwards, and then by the "version" field downwards, and only then by the entry "id". kernel-install would initialize this from ID= in os-release by default. But people could also just put "20", "50", "99" in there if they like. This should be unproblematic for compat.

3. same as 2, but call this expressly "os-id" or so and strictly initialize it from ID= from os-release. Benefit: the field would add a lot more meaning to things beyond mere ordering.

I am kinda leaning towards option 3, but 2 makes a lot of sense to me, too.  Option 2 has the benefit we can synthesize something reasonably useful also for entries that do not encapsulate "just an OS". For example, let's say I build Lennart OS and I want to synthesize additional menu entries "Lennart OS Factory Reset", and "Lennart OS Memory Test", besides the main "Lennart OS" or so. if we define sort-key I could neatly come up with a suitable sort-key to enforce the right ordering...

poettering:
So, I think I prefer option 2 in the end, i.e. adding a "sort-key" field. Does that make sense? Opinions? Will prep a patch for that now, to illustrate what I have in mind.

poettering:
I implemented my ideas now. Let's continue in #22519.

PR #22394, opened 2022-02-03
man: explicitly mention that Requires propagates restarts

bluca:
It's implicit, but the actions are different, so let's be explicit to
avoid any confusion.

### DIFF
--- a/man/systemd.unit.xml
+++ b/man/systemd.unit.xml
@@ -630,8 +630,8 @@
         <para>If this unit gets activated, the units listed will be activated as well. If one of
         the other units fails to activate, and an ordering dependency <varname>After=</varname> on the
         failing unit is set, this unit will not be started. Besides, with or without specifying
-        <varname>After=</varname>, this unit will be stopped if one of the other units is explicitly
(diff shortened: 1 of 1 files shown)


man/systemd.unit.xml
```diff
@@ -630,8 +630,8 @@
         <para>If this unit gets activated, the units listed will be activated as well. If one of
         the other units fails to activate, and an ordering dependency <varname>After=</varname> on the
         failing unit is set, this unit will not be started. Besides, with or without specifying
-        <varname>After=</varname>, this unit will be stopped if one of the other units is explicitly
-        stopped.</para>
+        <varname>After=</varname>, this unit will be stopped (or restarted) if one of the other units is
+        explicitly stopped (or restarted).</para>
```
nh2, line 634: @bluca @anitazha In this changed wording, does `restarted` refer to only `explicitly restarted` (e.g. with `systemctl restart`), or to _any_ restarts (thus also those that are triggered by systemd itself, when a service crashes and gets restarted by systemd)?

PR #22395, opened 2022-02-03
oomd: Dump top offenders after a kill action

benzea:
This hopefully makes it more transparent why a specific cgroup was
killed by systemd-oomd.

CC: @anitazha

### DIFF
--- a/src/oom/oomd-util.c
+++ b/src/oom/oomd-util.c
@@ -216,9 +216,36 @@ int oomd_cgroup_kill(const char *path, bool recurse, bool dry_run) {
         return set_size(pids_killed) != 0;
 }
 
+typedef void (*dump_candidate_func)(const OomdCGroupContext *ctx, FILE *f, const char *prefix);
+
+static int dump_kill_candidates(OomdCGroupContext **sorted, int n, int dump_until, dump_candidate_func dump_func) {
+        /* Try dumping top offendors, ignoring any errors that might happen. */
+        _cleanup_free_ char *dump = NULL;
+        _cleanup_fclose_ FILE *f = NULL;
+        int r;
+        size_t size;
+
+        f = open_memstream_unlocked(&dump, &size);
+        if (!f)
+                return -errno;;
+
+        fprintf(f, "Considered %d cgroups for killing, top candidates were:\n", n);
+        for (int i = 0; i < dump_until; i++)
+                dump_func(sorted[i], f, "\t");
+
+        r = fflush_and_check(f);
+        if (r < 0)
+                return r;
+
+        f = safe_fclose(f);
+
+        return log_dump(LOG_INFO, dump);
+}
+
 int oomd_kill_by_pgscan_rate(Hashmap *h, const char *prefix, bool dry_run, char **ret_selected) {
         _cleanup_free_ OomdCGroupContext **sorted = NULL;
         int n, r, ret = 0;
+        int dump_until;
 
         assert(h);
         assert(ret_selected);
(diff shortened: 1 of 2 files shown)


src/oom/oomd-util.c
```diff
@@ -280,13 +306,35 @@ int oomd_kill_by_swap_usage(Hashmap *h, uint64_t threshold_usage, bool dry_run,
...
+        for (int i = 0; i < dump_until; i++)
+                oomd_dump_swap_cgroup_context(sorted[i], f, "\t");
+
+        r = fflush_and_check(f);
+        if (r < 0)
+                return ret;
+
+        f = safe_fclose(f);
+
+        log_dump(LOG_INFO, dump);
```
poettering: this appears to be the exact same code as the code above, just the function called in the loop is different? Given theset two functions have the same signature, maybe split out this code into a common helper call that accepts the function to call as parameter?

benzea: Sounds good to me!

anitazha (approved):
Thank you! Will be good feature to have for debugging. LGTM

src/oom/oomd-util.c
```diff
@@ -216,9 +216,36 @@ int oomd_cgroup_kill(const char *path, bool recurse, bool dry_run) {
...
+static int dump_kill_candidates(OomdCGroupContext **sorted, int n, int dump_until, dump_candidate_func dump_func) {
+        /* Try dumping top offendors, ignoring any errors that might happen. */
+        _cleanup_free_ char *dump = NULL;
+        _cleanup_fclose_ FILE *f = NULL;
+        int r;
+        size_t size;
+
+        f = open_memstream_unlocked(&dump, &size);
+        if (!f)
+                return -errno;;
```
poettering, line 230: double ";;"

src/oom/oomd-util.c
```diff
@@ -216,9 +216,36 @@ int oomd_cgroup_kill(const char *path, bool recurse, bool dry_run) {
...
 
+typedef void (*dump_candidate_func)(const OomdCGroupContext *ctx, FILE *f, const char *prefix);
+
+static int dump_kill_candidates(OomdCGroupContext **sorted, int n, int dump_until, dump_candidate_func dump_func) {
+        /* Try dumping top offendors, ignoring any errors that might happen. */
+        _cleanup_free_ char *dump = NULL;
+        _cleanup_fclose_ FILE *f = NULL;
+        int r;
+        size_t size;
+
```
poettering, line 227: maybe rename dump_until → dump_max?

src/oom/oomd-util.c
```diff
@@ -216,9 +216,36 @@ int oomd_cgroup_kill(const char *path, bool recurse, bool dry_run) {
...
+        fprintf(f, "Considered %d cgroups for killing, top candidates were:\n", n);
+        for (int i = 0; i < dump_until; i++)
+                dump_func(sorted[i], f, "\t");
+
+        r = fflush_and_check(f);
+        if (r < 0)
+                return r;
+
+        f = safe_fclose(f);
```
poettering, line 240: you can drop this, it's implied due to the cleanup logic above. and the fflush will already fill in the `dump` string, so no need to free this here explicitly

src/oom/oomd-util.c
benzea: I think `dump_until` is more fitting here as it assumes that `dump_until <= n`. We could rename it `dump_count` or so maybe?

PR #22396, opened 2022-02-03
man: systemd-stdio-bridge doesn't connect "two busses"

poettering:
In D-Bus, clients connect to a bus (the usual case), or use direct
questions to each other (the unusual case). A bus is a program one can
connect to and implemented by dbus-daemon or dbus-broker. HOwever,
busses never connect between each other, that doesn't exist. Hence don't
claim so.

This is probably confusion about the fact that sd-bus calls D-Bus
connection objects just "sd_bus" for simplicity, given they are used in
99% of the cases to connect to a bus — only in exceptional cases they
are used for direct connections between peers without involving a bus.

Follow-up for b7bb58ef70b0c876941a1c31ed4e2f5f1dc5ed0e

### DIFF
--- a/man/systemd-stdio-bridge.xml
+++ b/man/systemd-stdio-bridge.xml
@@ -31,12 +31,11 @@
   <refsect1>
     <title>Description</title>
 
-    <para><command>systemd-stdio-bridge</command> implements a proxy for a D-Bus endpoint. It expects to
-    receive an open connection to a bus when started, and will also connect to a (different) bus as a
-    client. It will then act as a server on the first connection, and forward messages between the two
-    busses. This program is suitable for socket activation: the first connection may be a pipe or a socket
-    and must be passed as either standard input, or as an open file descriptor according to the protocol
-    described in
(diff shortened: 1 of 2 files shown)


man/systemd-stdio-bridge.xml
```diff
@@ -31,12 +31,11 @@
...
 
-    <para><command>systemd-stdio-bridge</command> implements a proxy for a D-Bus endpoint. It expects to
-    receive an open connection to a bus when started, and will also connect to a (different) bus as a
-    client. It will then act as a server on the first connection, and forward messages between the two
-    busses. This program is suitable for socket activation: the first connection may be a pipe or a socket
-    and must be passed as either standard input, or as an open file descriptor according to the protocol
-    described in
+    <para><command>systemd-stdio-bridge</command> implements a proxy between STDIN/STDOUT and a D-Bus bus. It
+    expects to receive an open connection via STDIN/STDOUT when started, and will create a new connection to
+    the specified bus. It will then act forward messages between the two connections. This program is
```
bluca: "It will then forward"

bluca (approved):

poettering:
Force pushed the typo fix. Upgrading green label.

PR #22398, opened 2022-02-03
Revert "bootctl: Ignore boot entries (continue #22041)"

poettering:
I am pretty sure we shouldn't totally hide items that sd-boot has not seen yet or that doesn't exist anymore, but we should just present them differently.

The tool should show the data discovered, so that the user can see discrepancies, we should not hide those.

Reverts systemd/systemd#22161


poettering:
See discussion at https://github.com/systemd/systemd/pull/22161#issuecomment-1029104000

poettering:
Replaced by #22487

PR #22399, opened 2022-02-03
hostnamed: trivial optimization

poettering:


bluca (approved):

src/hostname/hostnamed.c
```diff
@@ -185,15 +185,16 @@ static int get_dmi_data(const char *database_key, const char *regular_key, char
...
         if (s) {
                 b = strdup(s);
                 if (!b)
                         return -ENOMEM;
         }
 
-        if (ret)
-                *ret = TAKE_PTR(b);
-
+        *ret = TAKE_PTR(b);
```
daandemeyer, line 197: Should we touch `ret` at all if `b` is `NULL`?

bluca: Yeah it should probably be

```
        if (!ret || !s)
                return !!s;
```

poettering: in systemd we have the rule:

1. don't touch return parameters on failure
2. always initialize return parameters on success
3. failure is defined by a return value < 0
4. success is defined by any return value >= 0

Thus, if b is NULL (because there was no data found) we should initialize the return parameter to NULL, since that#s how we communicate "not found". 

(alternatively the function could be changed to return -ENOENT or so in that case, but I am not convinced htat would be better.)

bluca: makes sense

PR #22403, opened 2022-02-03
hostnamed: drop "interactive" parameter from GetHardwareSerial()

poettering:
Since a long time the D-Bus spec knows a special bit in its message
header for indicating that "interactive" authentication is OK. The
original hostnamed API is before that was added hence most functions
expose that boolean as explicit argument.

For new added functions let's get rid of it, the message flag is good
enough and replaces it with complete functionality.

No new APIs should carry the "interactive" boolean flag explicitly as
argument anymore.

Follow-up for: 9697662915e47a4797b05003cb1970fe2b01e530

### DIFF
--- a/man/org.freedesktop.hostname1.xml
+++ b/man/org.freedesktop.hostname1.xml
@@ -58,8 +58,7 @@ node /org/freedesktop/hostname1 {
                   in  b interactive);
       GetProductUUID(in  b interactive,
                      out ay uuid);
-      GetHardwareSerial(in  b interactive,
-                        out s serial);
+      GetHardwareSerial(out s serial);
       Describe(out s json);
     properties:
       readonly s Hostname = '...';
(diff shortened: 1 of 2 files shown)


bluca:
FYI @nitinexclusively dbus interface change from previous PR

medhefgo:
(Commit title has a typo)

nitinexclusively:
> FYI @nitinexclusively dbus interface change from previous PR

@bluca Thank you for this information !!
I am facing one issue after compiling "systemd" locally . I am sorry , if this is not right place to ask this query and it 
will be helpful, if  there is any comment regarding this. 
I have used below instructions and Wifi doesnot work after this (even i cannot see Wifi panel)
Even i have tried stable system version 248 but same issuei.e Wifi doesnot work . 
-----------------------------------------------------
 git clone https://github.com/systemd/systemd.git
  ./configure
  make && make install
-----------------------------------------------------
Thank you !!

PR #22405, opened 2022-02-03
portable: reopen socket with O_RDONLY

bluca:
Follow-up for e3f7ed944ae750a40685c52349f3cc850db0876e


anitazha (approved):
matches the behavior prior to the change in the linked commit. LGTM

PR #17136, opened 2020-09-23
sd-dhcp6-client: Fix sending prefix delegation request

mog422:

PR #22406, opened 2022-02-03
sd-dhcp6-client: fix sending prefix delegation request during rebind

thom311:
Fixes an assertion failure `pd->type == SD_DHCP6_OPTION_IA_PD` in `dhcp6_option_append_pd()`.

Something similar was done in commit 26a63b81322a ('sd-dhcp6-client: Fix
sending prefix delegation request (#17136)'). The justification is
probably the same.

---

See also: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1068#note_1244436

### DIFF
--- a/src/libsystemd-network/sd-dhcp6-client.c
+++ b/src/libsystemd-network/sd-dhcp6-client.c
@@ -857,7 +857,7 @@ static int client_send_message(sd_dhcp6_client *client, usec_t time_now) {
                                 return r;
                 }
 
-                if (FLAGS_SET(client->request_ia, DHCP6_REQUEST_IA_PD)) {
+                if (FLAGS_SET(client->request_ia, DHCP6_REQUEST_IA_PD) && client->lease->pd.addresses) {
                         r = dhcp6_option_append_pd(&opt, &optlen, &client->lease->pd, NULL);
                         if (r < 0)
                                 return r;


src/libsystemd-network/sd-dhcp6-client.c
```diff
@@ -857,7 +857,7 @@ static int client_send_message(sd_dhcp6_client *client, usec_t time_now) {
                                 return r;
                 }
 
-                if (FLAGS_SET(client->request_ia, DHCP6_REQUEST_IA_PD)) {
+                if (FLAGS_SET(client->request_ia, DHCP6_REQUEST_IA_PD) && client->lease->pd.addresses) {
```
anitazha, line 860: There's a similar looking check under `DHCP6_STATE_SOLICITATION` that doesn't have `&& client->lease->pd.addresses`. Does it also need to be updated? Admittedly I know very little about this code.

evverx:
@thom311 looking at https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/1068#note_1231209 I wonder if NetworkManager replaces `assert_return` with `abort` somewhere? If it does I think it would make sense to does the same when the fuzzers are built because it would allow them to catch issues like that. @yuwata what do you think? For example with `abort` `fuzz-dhcp6-client-send` fails as soon as it starts with
```
==41873== ERROR: libFuzzer: deadly signal
    #0 0x51b1ee in __sanitizer_print_stack_trace (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x51b1ee)
    #1 0x4688a9 in fuzzer::PrintStackTrace() (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x4688a9)
    #2 0x449d06 in fuzzer::Fuzzer::CrashCallback() (.part.0) FuzzerLoop.cpp.o
    #3 0x449dcb in fuzzer::Fuzzer::StaticCrashSignalCallback() (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x449dcb)
    #4 0x7f5e5f8b274f  (/lib64/libc.so.6+0x4274f)
    #5 0x7f5e5f8ff84b in __pthread_kill_implementation (/lib64/libc.so.6+0x8f84b)
    #6 0x7f5e5f8b26a5 in gsignal (/lib64/libc.so.6+0x426a5)
    #7 0x7f5e5f89c7d2 in abort (/lib64/libc.so.6+0x2c7d2)
    #8 0x7f5e602a24c9 in log_assert_failed_return /home/vagrant/systemd/build/../src/basic/log.c:886:9
    #9 0x565976 in dhcp6_lease_set_serverid /home/vagrant/systemd/build/../src/libsystemd-network/sd-dhcp6-lease.c
    #10 0x54e4b9 in client_parse_message /home/vagrant/systemd/build/../src/libsystemd-network/sd-dhcp6-client.c:1169:29
    #11 0x5554b7 in client_receive_advertise /home/vagrant/systemd/build/../src/libsystemd-network/sd-dhcp6-client.c:1412:13
    #12 0x554aa8 in LLVMFuzzerTestOneInput /home/vagrant/systemd/build/../src/libsystemd-network/fuzz-dhcp6-client-send.c:47:13
    #13 0x44a239 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x44a239)
    #14 0x44ada0 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x44ada0)
    #15 0x44d469 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x44d469)
    #16 0x44d920 in fuzzer::Fuzzer::Loop(std::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x44d920)
    #17 0x43288a in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x43288a)
    #18 0x421e46 in main (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x421e46)
    #19 0x7f5e5f89d55f in __libc_start_call_main (/lib64/libc.so.6+0x2d55f)
    #20 0x7f5e5f89d60b in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x2d60b)
    #21 0x421e94 in _start (/home/vagrant/systemd/out/fuzz-dhcp6-client-send+0x421e94)
```

src/libsystemd-network/sd-dhcp6-client.c
thom311: I don't know (I don't full understand the code).

For one, that change was not necessary to fix the assertion failures I had in CI, so I didn't do it.
Also, the code there is slightly different, as it passes `&client->ia_pd` instead of `client->lease.pd`.

But yes. Maybe it should!! Dunno.

thom311:
> I wonder if NetworkManager replaces `assert_return` with `abort` somewhere?

Yes. NM redefines `assert_return()` to behave like `g_return_if_fail()`... which under normal conditions only prints a message. 
But in NM-CI we run with `G_DEBUG=fatal-warnings`, which turns this into an `abort()`.

yuwata:
> If it does I think it would make sense to do the same when the fuzzers are built because it would allow them to catch issues like that. @yuwata what do you think?

@evverx At least partially, addressed in #22412.

yuwata (approved):
LGTM.

PR #22407, opened 2022-02-03
Update year to 2022.

Edu4rdSHL:


medhefgo:
Blurgh. Is a date in that even really necessary? Can't we just drop it?

Edu4rdSHL:
> Blurgh. Is a date in that even really necessary? Can't we just drop it?

I think that it's ok to drop it, I was reading some documentation and noticed the wrong year so I just come here and updated it. Let's wait for what others say.

poettering:
Yeah we should probably replace the line with something else. Don't know what though.

PR #22415, opened 2022-02-04
udev/net: support to set MDI-X mode

### LINKED ISSUE #22386: udev: Support for setting MDI-X state in .link files
Low priority (very rarely needed).
If it is possible I would like to be able to set mdi-x state in similar way as speed/duplex can be set. Classic nework scripts (like in Fedora) allow to specify ethtool options in configuration file (ifcfg-xxx) in convenient way (ETHTOOL_OPTS="mdix on").

Something like MdiX=auto|off|on in [Link] section would be great.

Alternatives considered:
- udev rule calling ethtool
- custom .service started after network-online.target calling ethtool
- calling ethtool in rc.local or other startup script

Feature does not exist in systemd-networkd-249.6-2.fc35.x86_64

yuwata:
Closes #22386.

### DIFF
--- a/man/systemd.link.xml
+++ b/man/systemd.link.xml
@@ -947,6 +947,21 @@
       </varlistentry>
 
       <varlistentry>
+        <term><varname>MDI=</varname></term>
+        <listitem>
+          <para>Specifies the medium dependent interface (MDI) mode for the interface. A MDI describes
+          the interface from a physical layer implementation to the physical medium used to carry the
+          transmission. Takes one of the following words: <literal>straight</literal> (or equivalently:
+          <literal>mdi</literal>), <literal>crossover</literal> (or equivalently:
+          <literal>mdi-x</literal>, <literal>mdix</literal>), and <literal>auto</literal>. When
+          <literal>straight</literal>, the MDI straight through mode will be used. When
+          <literal>crossover</literal>, the MDI crossover (MDI-X) mode will be used. When
+          <literal>auto</literal>, the MDI status is automatically detected. Defaults to unset, and the
+          kernel's default will be used.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><varname>SR-IOVVirtualFunctions=</varname></term>
         <listitem>
           <para>Specifies the number of SR-IOV virtual functions. Takes an integer in the range
--- a/src/shared/ethtool-util.c
+++ b/src/shared/ethtool-util.c
@@ -74,6 +74,15 @@ static const char* const port_table[] = {
 DEFINE_STRING_TABLE_LOOKUP(port, NetDevPort);
 DEFINE_CONFIG_PARSE_ENUM(config_parse_port, port, NetDevPort, "Failed to parse Port setting");
 
+static const char* const mdi_table[] = {
+        [ETH_TP_MDI_INVALID]  = "unknown",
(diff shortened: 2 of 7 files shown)


man/systemd.link.xml
```diff
@@ -946,6 +946,16 @@
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>MDI-X=</varname></term>
+        <listitem>
+          <para>Specifies MDI-X mode for the port. Takes a boolean value or special value
```
poettering: some brief explanation what MDI-X actually is would be great. After reading this man page I am now more confused than before.

man/systemd.link.xml
```diff
@@ -946,6 +946,16 @@
...
+          <literal>auto</literal>. When true, the MDI-X (crossover) mode will be used. When false, the
+          MDI (straight through) mode will be used. When <literal>auto</literal>, the MDI status is
+          automatically detected. Defaults to unset, and the kernel's default will be used.</para>
+        </listitem>
+      </varlistentry>
```
poettering, line 962: so after reading the wikipedia page it appears to me we should normalize this option, i.e.

call it MDI= and then take "mdi" and "mdix" as parameters. i.e. for the same reason the wikipedia page is called "MDI" and not "MDI-X" we should not name our option "MDI-X" either, but just "MDI". 

(might make sense to make this parser a bit flexible and accept the syntaxes "mdi" and "straight" as equivalent, as well as "mdi-x", "midx" and "crossover" as equivalent.)

poettering:
(codewise looks fine too me)

GitArUs:
I have tested copr build, works as expected. Thanks :)

yuwata:
@poettering Thank you for the review. The name of the setting is changed MDI-X= -> MDI=, and takes words like "mdi", "mdi-x", "auto". Also, I copied brief explanation what is MDI from Wikipedia. PTAL.

@GitArUs Thank you for testing the PR so quickly!

GitArUs:
Tested again, works very well.

man/systemd.link.xml
```diff
@@ -946,6 +946,21 @@
...
 
+      <varlistentry>
+        <term><varname>MDI=</varname></term>
+        <listitem>
+          <para>Specifies the medium dependent interface (MDI) mode for the interface. A MDI describes
+          the interface from a physical layer implementation to the physical medium used to carry the
+          transmission. Takes one of the following words: <literal>mdi</literal>,
+          <literal>straight</literal>, <literal>mdi-x</literal>, <literal>mdix</literal>,
+          <literal>crossover</literal>, and <literal>auto</literal>. When <literal>mdi</literal> or
+          <literal>straight</literal>, the MDI straight through mode will be used. When
```
poettering, line 957: maybe write this as:

```
…<literal>straight</literal> (or equivalently: <literal>mdi</literal>), <literal>crossover</literal> (or equivalently: <literal>mdi-x</literal>, <literal>mdix</literal>), or <literal>auto</literal>.…
```

i.e. emphasize that these aren't actually six different options, but just three, as three of them are just aliases for two others.

poettering:
lgtm. tiny suggestion for the man page, looks great otherwise.

yuwata:
@poettering Thank you for the review. The man page is updated according to the above comment. Upgrading the green label.

PR #22416, opened 2022-02-04
Misc

fbuihuu:


yuwata (approved):
LGTM.

poettering:
So far the thinking was that the /run/systemd/netif/ dirs would always exist, so that clients can install inotify watches in it reliably, i.e. if the feature is available at all, then the dir should exist, so that clients can use sd-network APIs safely. Because otherwise, apps will fail with installing those inotify watches and then installing networkd later won't help anyone...

i am not convinced we should split all this up so much. Having the one extra dir there really doesn't matter...

poettering:
And I don't think the resolved/networkd case are comparable really... One is a private dir, the other about a shared file...

fbuihuu:
I see... I was mislead by the fact that 564761fcaeda8c013210f7c6934847a6d0228ec9 moved the networkd sysusers bits in a dedicated file. If those dirs needs to exist all the time maybe we should rename sysusers.d/systemd-network.conf into sysusers.d/systemd.conf to make thing clearer (and also add a comment in the conf file wouldn't hurt).

fbuihuu:
For the time being I dropped the related patch.

poettering:
lgtm

yuwata:
About the /run/systemd/netif directories, still I think it is ok to split them.
Currently, sd-network is not exposed, and only resolved and timesyncd are the users of the directory through sd-network. If these do not work well without the directories, then I think it is better to fix sd-network.

But, I do not like sd-network, and I'd like to replace it near future with varlink. Hence, the above is not a strong opinion.

PR #22420, opened 2022-02-04
tests: also fuzz packets sent in the DHCP6_STATE_SOLICITATION state

evverx:
With aborts enabled the fuzzer can catch issues like
https://github.com/systemd/systemd/commit/26a63b81322a3bd8b9fbd43f75897c391708de2c
Let's extend it a bit to let it cover issues like
https://github.com/systemd/systemd/pull/22406#discussion_r798932098

### DIFF
--- a/src/libsystemd-network/fuzz-dhcp6-client-send.c
+++ b/src/libsystemd-network/fuzz-dhcp6-client-send.c
@@ -44,14 +44,16 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         assert_se(sd_dhcp6_client_set_transaction_id(client, htobe32(0x00ffffff) & ((const DHCP6Message *) data)->transaction_id) == 0);
 
         triple_timestamp_get(&t);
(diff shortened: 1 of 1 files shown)


yuwata (approved):

src/libsystemd-network/fuzz-dhcp6-client-send.c
```diff
@@ -44,14 +44,16 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+        if (r < 0)
                 goto cleanup;
 
         r = sd_event_now(client->event, clock_boottime_or_monotonic(), &time_now);
         if (r < 0)
                 goto cleanup;
 
-        client->state = DHCP6_STATE_REQUEST;
+        if (r == DHCP6_STATE_REQUEST)
+                client->state = DHCP6_STATE_REQUEST;
```
yuwata, line 56: The solicit message is created without previously received messages. Hence, the `client_send_message()` below does almost same thing.

But, this is OK, as this does not disturb any paths previously tested.

Issue #22376, opened 2022-02-02
KERNEL_INSTALL_MACHINE_ID in /etc/machine-info should be revisited


poettering:
I am not too happy with  #21757 and how it played out.

Three points:

1. I would like to understand why /etc/machine-info is ever written to by the script. Consuming might be acceptable, but writing it?
2. I don't think /etc/machine-info is the right place for this, it's not really machine info, it's kernel-install specific configuration. Hence, can we please move this to /etc/kernel/image-prefix-dir or so? (see my comments on the original PR)
3. I don't think the concept should be called "machine-id", because apparently it's *not* the machine ID, and the string can actually be anyhing, i.e. an arbitrary prefix dir.

poettering:
(I'd be fine with still honouring KERNEL_INSTALL_MACHINE_ID if it is set, but I am pretty sure we should move this elsewhere, and I think probably never auto-write this, but maybe I don't understand what this is about...)

DaanDeMeyer:
Writing a separate file in /etc/kernel was my initial idea but @keszybz suggested we use /etc/machine-info. I don't mind switching back to a separate file in /etc/kernel but @keszybz might want to pitch in here as well.

poettering:
@DaanDeMeyer what was the rationale for writing that file automatically though?

(writing to env var files sucks btw, since you basically need an editor. if you we use a file like /etc/kernel/image-prefix-dir or so that just containst the prefix dir string that is much either since if you want to change it you jst rewrite the file as a whole)

Also, /etc/kernel/install.conf exists now too. So one more file even.

This is just a mess, so much configuration options and I doubt we actually want any of this configurable tbh...

DaanDeMeyer:
@poettering The reason for writing a file (either machine-info or a new one in /etc/kernel) automatically is because we need some kind of identifier to identify the directory to use in the boot directory that isn't the machine ID because it's not guaranteed to be available. Like you already said, it doesn't matter what it actually is, so we decided to just generate a UUID and use that. Of course we need to persist the generated UUID somewhere. My initial PR stored it in /etc/kernel/boot-id, but after review I changed it to be stored in /etc/machine-info as requested.

DaanDeMeyer:
See also #22224 which brings up valid points, especially around factory resets of /etc.

[...]

PR #22421, opened 2022-02-04
Kernel install fixes

daandemeyer:
A set of commits to implement some of the suggestions from #22376.


DaanDeMeyer:
Not tested yet so adding the don't merge label. Looking for feedback on the approach

poettering:
So I started working on  #22463 before seeing this one. I think it addresses everything this PR addresses plus a few more things, hence I hope it's OK if I close this?

PR #22423, opened 2022-02-04
repart: fix sector size handling

poettering:
This queries the sector size from libfdisk instead of assuming 512, and
uses that when converting from bytes to the offset/size values libfdisk
expects.

This is an alternative to Tom Yan's #21823, but prefers using libfdisk's
own ideas of the sector size instead of going directly to the backing
device via ioctls. (libfdisk can after all also operate on regular
files, where the sector size concept doesn't necessarily apply the same
way.)

This also makes the "grain" variable, i.e. how we'll align the
partitions. Previously this was hardcoded to 4K, and that still will be
the minimum grain we use, but should the sector size be larger than that
we'll use the next multiple of the sector size instead.

### DIFF
--- a/test/units/testsuite-58.sh
+++ b/test/units/testsuite-58.sh
@@ -189,6 +189,48 @@ grep -qF 'p2 : start=      104448, size=      100319,' /tmp/testsuite-58-issue-2
 rm /tmp/testsuite-58-issue-21817.img /tmp/testsuite-58-issue-21817.dump
 rm -r /tmp/testsuite-58-issue-21817-defs/
 
+testsector()
+{
+    echo "Running sector test with sector size $1..."
+
+    mkdir -p /tmp/testsuite-58-sector
+    cat > /tmp/testsuite-58-sector/a.conf <<EOF
+[Partition]
+Type=root
+SizeMaxBytes=15M
+SizeMinBytes=15M
+EOF
+    cat > /tmp/testsuite-58-sector/b.conf <<EOF
+[Partition]
+Type=linux-generic
+Weight=250
+EOF
+
+    cat > /tmp/testsuite-58-sector/c.conf <<EOF
+[Partition]
+Type=linux-generic
+Weight=750
+EOF
+
+    truncate -s 100m "/tmp/testsuite-58-sector-$1.img"
+    LOOP=$(losetup -b "$1" -P --show -f "/tmp/testsuite-58-sector-$1.img" )
+    systemd-repart --pretty=yes --definitions=/tmp/testsuite-58-sector/ --seed=750b6cd5c4ae4012a15e7be3c29e6a47 --empty=require --dry-run=no "$LOOP"
+    rm -rf /tmp/testsuite-58-sector
+    sfdisk --verify "$LOOP"
+    sfdisk --dump "$LOOP"
+    losetup -d "$LOOP"
+
+    rm "/tmp/testsuite-58-sector-$1.img"
+}
+
+# Valid block sizes on the Linux block layer are >= 512 and <= PAGE_SIZE, and
+# must be powers of 2. Which leaves exactly four different ones to test on
+# typical hardware
+testsector 512
+testsector 1024
+testsector 2048
+testsector 4096
+
 echo OK >/testok
 
 exit 0


poettering:
(dont merge yet, i intend to extend the testcase with a 4K loopback block device test first)

bluca:
```
19:11:02 #0  0x000055d4830aafeb in round_up_size (v=18446744073709551615, p=p@entry=0) at ../build/src/partition/repart.c:224
19:11:02 #1  0x000055d4830b3dff in partition_min_size (context=context@entry=0x55d484193aa0, p=p@entry=0x55d484194c20) at ../build/src/partition/repart.c:444
19:11:02 #2  0x000055d4830b3e3c in partition_min_size_with_padding (context=context@entry=0x55d484193aa0, p=p@entry=0x55d484194c20) at ../build/src/partition/repart.c:489
19:11:02 #3  0x000055d4830b57ed in determine_auto_size (c=c@entry=0x55d484193aa0) at ../build/src/partition/repart.c:4807
19:11:02 #4  0x000055d4830be2e9 in run (argc=<optimized out>, argv=<optimized out>) at ../build/src/partition/repart.c:4972
19:11:02 #5  0x000055d4830be576 in main (argc=<optimized out>, argv=<optimized out>) at ../build/src/partition/repart.c:5029
19:11:02 (gdb) #0  0x000055d4830aafeb in round_up_size (v=18446744073709551615, p=p@entry=0)
19:11:02     at ../build/src/partition/repart.c:224
19:11:02         __unique_prefix_X15 = 18446744073709551615
19:11:02         __unique_prefix_Y16 = 0
19:11:02 #1  0x000055d4830b3dff in partition_min_size (
19:11:02     context=context@entry=0x55d484193aa0, p=p@entry=0x55d484194c20)
19:11:02     at ../build/src/partition/repart.c:444
19:11:02         f = <optimized out>
19:11:02         d = 0
19:11:02         sz = 4096
19:11:02         __PRETTY_FUNCTION__ = "partition_min_size"
19:11:02 #2  0x000055d4830b3e3c in partition_min_size_with_padding (
19:11:02     context=context@entry=0x55d484193aa0, p=p@entry=0x55d484194c20)
19:11:02     at ../build/src/partition/repart.c:489
19:11:02         sz = <optimized out>
19:11:02         __PRETTY_FUNCTION__ = "partition_min_size_with_padding"
19:11:02 #3  0x000055d4830b57ed in determine_auto_size (c=c@entry=0x55d484193aa0)
19:11:02     at ../build/src/partition/repart.c:4807
19:11:02         m = <optimized out>
19:11:02         sum = 1069056
19:11:02         p = 0x55d484194c20
19:11:02         __PRETTY_FUNCTION__ = "determine_auto_size"
19:11:02         __func__ = "determine_auto_size"
19:11:02 #4  0x000055d4830be2e9 in run (argc=<optimized out>, argv=<optimized out>)
19:11:02     at ../build/src/partition/repart.c:4972
19:11:02         loop_device = 0x0
19:11:02         decrypted_image = 0x0
19:11:02         mounted_dir = 0x0
19:11:02         context = 0x55d484193aa0
19:11:02         node = 0x55d484194bf0 "/var/tmp/testsuite-58.img"
19:11:02         backing_fd = 3
19:11:02         from_scratch = true
19:11:02         node_is_our_loop = false
19:11:02         r = 0
19:11:02         __PRETTY_FUNCTION__ = "run"
19:11:02         __func__ = "run"
19:11:02 #5  0x000055d4830be576 in main (argc=<optimized out>, argv=<optimized out>)
19:11:02     at ../build/src/partition/repart.c:5029
19:11:02         r = <optimized out>
19:11:02         __PRETTY_FUNCTION__ = "main"
19:11:02 (gdb) 
```

poettering:
ready for review now. Added a test case for the non-512-sector cases now. ptal

bluca (approved):

test/units/testsuite-58.sh
```diff
@@ -189,6 +189,47 @@ grep -qF 'p2 : start=      104448, size=      100319,' /tmp/testsuite-58-issue-2
...
+[Partition]
+Type=linux-generic
+Weight=750
+EOF
+
+    truncate -s 100m "/tmp/testsuite-58-sector-$1.img"
+    LOOP=$(losetup -b "$1" -P --show -f "/tmp/testsuite-58-sector-$1.img" )
+    systemd-repart --pretty=yes --definitions=/tmp/testsuite-58-sector/ --seed=750b6cd5c4ae4012a15e7be3c29e6a47 --empty=require --dry-run=no "$LOOP"
+    rm -rf /tmp/testsuite-58-sector
+    sfdisk --dump "$LOOP"
```
medhefgo, line 220: `--verify` seems predestined for this, no?

medhefgo:
Not that I know much about this and it probably doesn't matter, but these days I'd expect 1MB aligned partitions…

test/units/testsuite-58.sh
poettering: I wonder what --verify actually does... Man page is so vague about it

medhefgo: `test whether partitions seem correct` D'uh :P

poettering:
Force pushed a new version that also adds an sfdisk --verify call to the test (as suggested), even though I am not sure what it actually entails...

PR #22426, opened 2022-02-04
MemoryZSwapMax directive to configure new memory.zswap.max cgroup file

plv:
NB: This is a Facebook-only PR intended that we into pull in for an upcoming kernel feature. Please disregard.

This PR adds a new directive called `MemoryZSwapMax` for controlling a cgroup's `memory.zswap.max` added by hnaz. It functions similarly to `MemorySwapMax` . Note that since it's not rolled out everywhere internally it checks support by seeing if `/sys/fs/cgroup/init.scope/memory.zswap.max` exists. 

A simple manual test:

```
[root@image ~]# systemd-run --unit=test --slice=test.slice --property="MemoryZSwapMax=4096B" sleep 60
Running as unit: test.service
[root@image ~]# cat /sys/fs/cgroup/test.slice/test.service/memory.zswap.max
4096
[root@image ~]# cat /sys/fs/cgroup/test.slice/test.service/memory.zswap.max
4096
```

cc @cdown @anitazha

### DIFF
--- a/man/systemd.resource-control.xml
+++ b/man/systemd.resource-control.xml
@@ -325,14 +325,15 @@
 
       <varlistentry>
         <term><varname>MemorySwapMax=<replaceable>bytes</replaceable></varname></term>
+        <term><varname>MemoryZSwapMax=<replaceable>bytes</replaceable></varname></term>
 
         <listitem>
-          <para>Specify the absolute limit on swap usage of the executed processes in this unit.</para>
+          <para>Specify the absolute limit on (z)swap usage of the executed processes in this unit.</para>
 
           <para>Takes a swap size in bytes. If the value is suffixed with K, M, G or T, the specified swap size is
           parsed as Kilobytes, Megabytes, Gigabytes, or Terabytes (with the base 1024), respectively. If assigned the
-          special value <literal>infinity</literal>, no swap limit is applied. This controls the
-          <literal>memory.swap.max</literal> control group attribute. For details about this control group attribute,
+          special value <literal>infinity</literal>, no swap limit is applied. These settings control the
+          <literal>memory.(z)swap.max</literal> control group attributes. For details about these control group attributes,
           see <ulink url="https://docs.kernel.org/admin-guide/cgroup-v2.html#memory-interface-files">Memory Interface Files</ulink>.</para>
         </listitem>
       </varlistentry>
--- a/src/core/cgroup.c
+++ b/src/core/cgroup.c
@@ -151,6 +151,7 @@ void cgroup_context_init(CGroupContext *c) {
                 .memory_high = CGROUP_LIMIT_MAX,
                 .memory_max = CGROUP_LIMIT_MAX,
                 .memory_swap_max = CGROUP_LIMIT_MAX,
+                .memory_zswap_max = CGROUP_LIMIT_MAX,
 
                 .memory_limit = CGROUP_LIMIT_MAX,
 
@@ -354,6 +355,9 @@ static int unit_compare_memory_limit(Unit *u, const char *property_name, uint64_
         } else if (streq(property_name, "MemorySwapMax")) {
                 unit_value = c->memory_swap_max;
                 file = "memory.swap.max";
+        } else if (streq(property_name, "MemoryZSwapMax")) {
+                unit_value = c->memory_zswap_max;
+                file = "memory.zswap.max";
         } else
                 return -EINVAL;
 
@@ -396,9 +400,10 @@ static char *format_cgroup_memory_limit_comparison(char *buf, size_t l, Unit *u,
 
         /* memory.swap.max is special in that it relies on CONFIG_MEMCG_SWAP (and the default swapaccount=1).
          * In the absence of reliably being able to detect whether memcg swap support is available or not,
-         * only complain if the error is not ENOENT. */
+         * only complain if the error is not ENOENT. This is similarly the case for memory.zswap.max relying
+         * on CONFIG_ZSWAP. */
         if (r > 0 || IN_SET(r, -ENODATA, -EOWNERDEAD) ||
-            (r == -ENOENT && streq(property_name, "MemorySwapMax")))
+            (r == -ENOENT && STR_IN_SET(property_name, "MemorySwapMax", "MemoryZSwapMax")))
                 buf[0] = 0;
         else if (r < 0) {
                 errno = -r;
@@ -462,6 +467,7 @@ void cgroup_context_dump(Unit *u, FILE* f, const char *prefix) {
                 "%sMemoryHigh: %" PRIu64 "%s\n"
                 "%sMemoryMax: %" PRIu64 "%s\n"
                 "%sMemorySwapMax: %" PRIu64 "%s\n"
+                "%sMemoryZSwapMax: %" PRIu64 "%s\n"
                 "%sMemoryLimit: %" PRIu64 "\n"
                 "%sTasksMax: %" PRIu64 "\n"
                 "%sDevicePolicy: %s\n"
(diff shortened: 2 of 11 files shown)


plv:
Hmm builds are failing because `bus_cgroup_set_zswap_scale` is unused. I know it's generated by a macro I used, but what is it for/how does the regular `bus_cgroup_set*` functions?

src/core/cgroup.c
```diff
@@ -1547,6 +1555,9 @@ static void cgroup_context_apply(
                         cgroup_apply_unified_memory_limit(u, "memory.high", c->memory_high);
                         cgroup_apply_unified_memory_limit(u, "memory.max", max);
                         cgroup_apply_unified_memory_limit(u, "memory.swap.max", swap_max);
+                        if (cg_zswap_supported()) {
+                          cgroup_apply_unified_memory_limit(u, "memory.zswap.max", zswap_max);
```
anitazha: 8sp indents please. no surrounding parens either.

anitazha (changes requested):
Overall this looks good. Small nit.

> Hmm builds are failing because bus_cgroup_set_zswap_scale is unused. I know it's generated by a macro I used, but what is it for/how does it differ from the regular bus_cgroup_set_zswap function?

Take a look at `src/core/dbus-cgroup.c` and use the generated function there. Quite a few of the cgroup2 dbus properties have a  corresponding "scaled" version of the property. So if a percentage is passed in on the unit file it is instead stored as an unsigned value scaled from 0 to uint32_max.

anitazha:
LGTM

anitazha (approved):

poettering:
> NB: This is a Facebook-only PR intended that we into pull in for an upcoming kernel feature. Please disregard.

So, where are we with this?

src/basic/cgroup-util.c
```diff
@@ -177,6 +177,24 @@ bool cg_kill_supported(void) {
...
 
+bool cg_zswap_supported(void) {
+        static thread_local int supported = -1;
+
+        if (supported >= 0)
+                return supported;
+
+        if (cg_all_unified() <= 0)
+                supported = false;
+        else if (access("/sys/fs/cgroup/init.scope/memory.zswap.max", F_OK) < 0) {
```
anitazha: One thing I would change is using `cg_pid_get_path()` here and then using that to get the complete path to check the existence of `memory.zswap.max`.

anitazha (changes requested):
memory.zswap.max is in 5.19 (https://github.com/torvalds/linux/commit/f4840ccfca25db225b3371a8f7b5770febee87c5) so we can merge this (after applying the suggestion). I pinged @plv but if he's not interested I can rebase and carry this through.

src/basic/cgroup-util.c
Werkov: This check relies on memory controller being enabled for `init.scope` (which is technically correct, but it's a dynamic property and cache can have a stale incorrect value). (Also it relies on swap accounting enabled (not disabled) on kernel boot.)

I'd suggest simply ignoring ENOENT upon writes in a particular cgroup as with `memory.swap.max`.

poettering: hmm, why isn't zswap availability reported in /sys/kernel/cgroup/features?

poettering: (and could it be added there?)

Werkov: (The features file lists features that change semantics of the exposed files and match mount options. It could be changed but it'd be redundant to check of the attribute file existence.)

This is IMO simpler to approach the same way like `memory.swap.max` is already handled. (And devise more sophisticated approach if it proves slow.)

anitazha:
I rebased Pasha's changes and it looks consistent with what werkov suggested (treat memory.zswap.max the same as memory.swap.maz). Hopefully someone else can take a look too

anitazha:
post-252 rebase

src/core/cgroup.c
```diff
@@ -396,9 +400,10 @@ static char *format_cgroup_memory_limit_comparison(char *buf, size_t l, Unit *u,
 
         /* memory.swap.max is special in that it relies on CONFIG_MEMCG_SWAP (and the default swapaccount=1).
          * In the absence of reliably being able to detect whether memcg swap support is available or not,
-         * only complain if the error is not ENOENT. */
+         * only complain if the error is not ENOENT. This is similarly the case for memory.zswap.max relying on CONFIG_ZSWAP. */
         if (r > 0 || IN_SET(r, -ENODATA, -EOWNERDEAD) ||
-            (r == -ENOENT && streq(property_name, "MemorySwapMax")))
+            (r == -ENOENT && (streq(property_name, "MemorySwapMax") ||
+                              streq(property_name, "MemoryZSwapMax"))))
```
daandemeyer: This should use `STR_IN_SET()`

src/core/cgroup.c
```diff
@@ -1570,10 +1577,12 @@ static void cgroup_context_apply(
 
                 if (cg_all_unified() > 0) {
                         uint64_t max, swap_max = CGROUP_LIMIT_MAX;
+                        uint64_t zswap_max = CGROUP_LIMIT_MAX;
```
daandemeyer: This can be merged with the previous line

DaanDeMeyer (changes requested):
This is missing a manpage update

DaanDeMeyer:
Fixed the minor nits I noticed and updated the documentation, setting the green label

poettering:
I am sorry, but what *is* this? comes with no explanation whatsoever? is this some facebook internal thing?

man/systemd.resource-control.xml
```diff
@@ -325,14 +325,15 @@
...
-          <para>Specify the absolute limit on swap usage of the executed processes in this unit.</para>
+          <para>Specify the absolute limit on (z)swap usage of the executed processes in this unit.</para>
 
           <para>Takes a swap size in bytes. If the value is suffixed with K, M, G or T, the specified swap size is
           parsed as Kilobytes, Megabytes, Gigabytes, or Terabytes (with the base 1024), respectively. If assigned the
-          special value <literal>infinity</literal>, no swap limit is applied. This controls the
-          <literal>memory.swap.max</literal> control group attribute. For details about this control group attribute,
+          special value <literal>infinity</literal>, no swap limit is applied. These settings control the
+          <literal>memory.(z)swap.max</literal> control group attributes. For details about these control group attributes,
           see <ulink url="https://docs.kernel.org/admin-guide/cgroup-v2.html#memory-interface-files">Memory Interface Files</ulink>.</para>
```
poettering, line 337: so, I assume I am fairly educated with these things, but wtf is this all? please write some docs what this actually is supposed to be. i have no idea what to read from this. what does this *actually* control? what is zswap?

we really should be better at this.

how doe swap limits relate to zswap limits? 

i shouldn't have to consult kerne docs to grok this. And in fact, I did consult them and are more puzzled than before now. This is just not useful. Please provide some docs what this?

poettering:
if this is about compressed swap, I wonder if it wouldn't be better to expose this as MemoryCompressedSwapMax=. The "z" thing is more than cryptic.

Winterhuman:
@poettering Calling it just `MemoryCompressedSwapMax` might be confusing since there's actually "zswap" as well as "zram" which are two separate things:

https://wiki.archlinux.org/title/ZRAM#zram_or_zswap

"The difference compared to ZRAM is that zswap works in conjunction with a swap device (compressing things in RAM before moving that stuff to swap) while zram is a swap device in RAM that does not require a backing swap device. "

anitazha:
The original description was written a while ago; this is a feature available in the kernel now, not Facebook only.

Since someone already commented about zswap vs. zram, I'll elaborate and say memory.zswap.max can limit the amount of memory used for zswap, per cgroup. That way if some cgroups don't want to use zswap at all they can set this property to zero. 

I could expand the man pages with the additional context but would suggest keeping the property name as is instead of renaming it.

Winterhuman:
Maybe https://www.kernel.org/doc/html/v4.18/vm/zswap.html would be better to link to? It seems to describe Zswap pretty well

PR #22433, opened 2022-02-07
networkd: attempt to trigger kernel IPv6LL address generation

### LINKED ISSUE #22424: networkd: fails to restore IPv6LL address generation
`systemd 250 (250.3-2-arch)`

**Used distribution**
Arch Linux 2022.02.01 (installation iso)

`Linux archiso 5.16.4-arch1-1 #1 SMP PREEMPT Sat, 29 Jan 2022 19:08:13 +0000 x86_64 GNU/Linux`

`x86_64`

If a network interface has kernel IPv6LL address generation disabled when networkd is started, networkd re-enables address generation but the interface nevers obtains an IPv6LL address, resulting in the interface being stuck in the `configuring` state.

1. Boot Arch iso
`qemu-system-x86_64 -enable-kvm -cdrom archlinux-2022.02.01-x86_64.iso -m 4G`

2. Disable IPv6LL

```
# sysctl -w net.ipv6.conf.ens3.addr_gen_mode=1
net.ipv6.conf.ens3.addr_gen_mode = 1
```

Note:
```
0: generate address based on EUI64 (default)
1: do no generate a link-local address, use EUI64 for addresses generated from autoconf
```

3. Clear v6 interface addresses
```
# ip link set ens3 down; ip link set ens3 up

# ip address show ens3
2: ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:12:34:56 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    inet 10.0.2.15/24 metric 100 brd 10.0.2.255 scope global dynamic ens3
       valid_lft 86358sec preferred_lft 86358sec

```

4. Restart networkd
```
# systemctl restart systemd-networkd
```

5. Networkd has restored kernel IPv6LL defaults
```
# sysctl net.ipv6.conf.ens3.addr_gen_mode
net.ipv6.conf.ens3.addr_gen_mode = 0
```
... but it's stuck waiting for an address

```
# networkctl status ens3            
● 2: ens3                                                                      >
                     Link File: /usr/lib/systemd/network/99-default.link
                  Network File: /etc/systemd/network/20-ethernet.network
                          Type: ether
                         State: routable (configuring)
                  Online state: online                                         >
             Alternative Names: enp0s3
                          Path: pci-0000:00:03.0
                        Driver: e1000
                        Vendor: Intel Corporation
                         Model: 82540EM Gigabit Ethernet Controller (QEMU Virtu>
              Hardware Address: 52:54:00:12:34:56
                           MTU: 1500 (min: 46, max: 16110)
                         QDisc: fq_codel
  IPv6 Address Generation Mode: eui64
          Queue Length (Tx/Rx): 1/1
              Auto negotiation: yes
                         Speed: 1Gbps
                        Duplex: full
                          Port: tp
                       Address: 10.0.2.15 (DHCP4 via 10.0.2.2)
                       Gateway: 10.0.2.2
                           DNS: 10.0.2.3
```

Note that if you now toggle (!) `addr_gen_mode` manually it _does_ assign an IPv6LL address..

```
# sysctl -w net.ipv6.conf.ens3.addr_gen_mode=1
# sysctl -w net.ipv6.conf.ens3.addr_gen_mode=0
```

Some interesting bits from systemd-networkd logs:

```
Feb 04 18:27:42 archiso systemd-networkd[931]: ens3: Setting IPv6LL address generation mode
...
Feb 04 18:27:42 archiso systemd-networkd[931]: ens3: IPv6LL address generation mode set.
Feb 04 18:27:42 archiso systemd-networkd[931]: ens3: master interface set.
Feb 04 18:27:42 archiso systemd-networkd[931]: ens3: link_check_ready(): link is not activated.
Feb 04 18:27:42 archiso systemd-networkd[931]: ens3: Bringing link up
Feb 04 18:27:42 archiso systemd-networkd[931]: ens3: link_check_ready(): IPv6LL is not configured yet.
```

Full networkd debug log: [networkd.log](https://gist.githubusercontent.com/ius/ce2a016678b351fe9dc7522e2d466fb5/raw/60c0caf93ba023a5bd2aaaddaf40ceb1dd9f54a0/networkd.log)

ius:
Try to ensure kernel IPv6 link local address generation occurs by
setting the per-if addr_gen_mode sysctl instead of solely relying on
the netlink interface (`IFLA_INET6_ADDR_GEN_MODE`).

The netlink setting is sufficient in cases where the interface is not
yet up when networkd configures an interface - bringing the interface
up will trigger in-kernel address generation.

If the interface is already up, yet the interface has no IPv6LL assigned
setting `IFLA_INET6_ADDR_GEN_MODE` has no effect.

Writing the addr_gen_mode sysctl is a best effort attempt at triggering
address generation regardless of interface state because it also works
in cases where the interface is already up.

### Caveat

Because the sysctl needs to be _changed_ to trigger IPv6LL allocation writing the sysctl works in this case (`addr_gen_mode=1` →`addr_gen_mode=0`).

If any other edge case drops all IPv6 adresses but leaves `addr_gen_mode=0` networkd with this PR will still fail because the sysctl isn't _toggled_ (you can simulate this by removing the sysctl write from the test case). I don't know what the best approach is to handle this, or whether we should care at all.

Closes #22424

@yuwata - thoughts?

### DIFF
--- a/src/network/networkd-sysctl.c
+++ b/src/network/networkd-sysctl.c
@@ -256,6 +256,29 @@ static int link_set_ipv6ll_stable_secret(Link *link) {
         return sysctl_write_ip_property(AF_INET6, link->ifname, "stable_secret", str);
 }
 
+static int link_set_ipv6_addr_gen_mode(Link *link) {
+        uint8_t mode;
+
+        assert(link);
+
+        if (!socket_ipv6_is_supported())
+                return 0;
+
+        if (!link->network)
+                return 0;
+
+        if (!link_ipv6ll_enabled(link))
+                return 0;
+        else if (link->network->ipv6ll_address_gen_mode >= 0)
+                mode = link->network->ipv6ll_address_gen_mode;
+        else if (in6_addr_is_set(&link->network->ipv6ll_stable_secret))
+                mode = IN6_ADDR_GEN_MODE_STABLE_PRIVACY;
+        else
+                mode = IN6_ADDR_GEN_MODE_EUI64;
+
+        return sysctl_write_ip_property_uint32(AF_INET6, link->ifname, "addr_gen_mode", mode);
+}
+
 static int link_set_ipv4_accept_local(Link *link) {
         assert(link);
 
@@ -330,6 +353,10 @@ int link_set_sysctl(Link *link) {
         r = link_set_ipv6ll_stable_secret(link);
         if (r < 0)
                 log_link_warning_errno(link, r, "Cannot set stable secret address for IPv6 link local address: %m");
+
+        r = link_set_ipv6_addr_gen_mode(link);
+        if (r < 0)
+                log_link_warning_errno(link, r, "Cannot set IPv6 address generation mode: %m");
 
         r = link_set_ipv4_accept_local(link);
         if (r < 0)
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -2952,6 +2952,32 @@ class NetworkdNetworkTests(unittest.TestCase, Utilities):
         self.assertEqual(read_ipv6_sysctl_attr('test1', 'addr_gen_mode'), '2')
         self.assertEqual(read_ipv6_sysctl_attr('dummy98', 'addr_gen_mode'), '1')
 
+    def test_link_local_addressing_up_without_addr(self):
+        output = check_output('ip link add dummy98 type dummy')
+        print(output)
+
+        print('## do no generate a link-local address')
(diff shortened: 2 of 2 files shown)


yuwata:
Thanks. I will take a look.

yuwata:
@ius Thank you for the analysis in #22424, and submitting this fix.

Based on your analysis, I opened #22452, which has several cleanups and is slightly optimized. PTAL.

yuwata:
Replaced by #22452. Closing.

PR #22434, opened 2022-02-07
watchdog: fix bad memory access

poettering:
In case 'watchdog_device' was initially set to '/dev/watchdog' we'll
patch it to '/dev/watchdog0' and free the original buffer. But that
means we can't use 'fn' anymore, but must use 'watchdog_device' from
then on, otherwise we have a bad memory access.

Follow-up for: f4b1a6b6419b771d26342a9b75b1d77ee9d90133

### DIFF
--- a/src/shared/watchdog.c
+++ b/src/shared/watchdog.c
@@ -137,17 +137,17 @@ static int open_watchdog(void) {
         if (r < 0)
                 return log_oom_debug();
 
-        watchdog_fd = open(fn, O_WRONLY|O_CLOEXEC);
+        watchdog_fd = open(watchdog_device, O_WRONLY|O_CLOEXEC);
         if (watchdog_fd < 0)
-                return log_debug_errno(errno, "Failed to open watchdog device %s, ignoring: %m", fn);
+                return log_debug_errno(errno, "Failed to open watchdog device %s, ignoring: %m", watchdog_device);
 
         if (ioctl(watchdog_fd, WDIOC_GETSUPPORT, &ident) < 0)
-                log_debug_errno(errno, "Hardware watchdog %s does not support WDIOC_GETSUPPORT ioctl, ignoring: %m", fn);
+                log_debug_errno(errno, "Hardware watchdog %s does not support WDIOC_GETSUPPORT ioctl, ignoring: %m", watchdog_device);
         else
                 log_info("Using hardware watchdog '%s', version %x, device %s",
                          ident.identity,
                          ident.firmware_version,
-                         fn);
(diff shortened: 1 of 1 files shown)


bluca (approved):

mrc0mmand:
I originally wanted to fix this last week as well, but in this particular case I think Coverity is "wrong", since we don't touch the original string in case both the source and the target are identical when calling `free_and_strdup()`:

https://github.com/systemd/systemd/blob/01bd5cee6b2c25f8170eaba49a0e1fb961118215/src/basic/string-util.c#L928-L938

or am I missing something else?

poettering:
We upgrade /dev/watchdog → /dev/watchdog0 implicitly. If that code path is hit, we *will* change the string, and thus Coverity is right.

yuwata:
Hmm??? Still I cannot catch the issue...

mrc0mmand:
> We upgrade /dev/watchdog → /dev/watchdog0 implicitly. If that code path is hit, we _will_ change the string, and thus Coverity is right.

In case `watchdog_device = "/dev/watchdog"`, then `fn = "/dev/watchdog0"`, but in `free_and_strdup()` we potentially free the original `watchdog_device` and `strdup()` `fn` into it. But it doesn't touch `fn` in any other way?

mrc0mmand:
I mean, the patch does make sense whether it's an issue or not, I was just puzzled why this wasn't caught by sanitizers (and I tried to reproduce it manually as well without success). So I just want to know where exactly the issue is :-)

yuwata:
Agreed. The patch itself is OK for me. But I cannot understand the issue explained in the commit message or coverity. If the issue is fake, then please update the commit message on merge (by using squash-merge).

poettering:
actually, it's me who was confused. I am wrong, and Coverity too. ;-)

There are only two cases possible: where "fn" points to the static constant string "/dev/watchdog0", and where it points to the status quo ante buffer. In the former case we'll free the status quo ante buffer, but this won't affect fn of course. In the latter case we'll not free the buffer, and leave things as is, and thus fn stays valid too.

Patch should still be merged, given it's really hard to see, and I can totally understand coverity's confusion on this (I mean, obviously, because I got confused too), but the commit msg is rubbish of course. Will shortly force push updated version with same code change but fixed commit msg.

poettering:
Force pushed the promised new commit msg. no code changes. ptal.

yuwata (approved):
LGTM.

mrc0mmand (approved):
LGTM as well, the CentOS CI fail is unrelated (I'll look into it later).

PR #22446, opened 2022-02-08
sd-boot: don't read TSC in VM environments, and other TSC fixes

poettering:
Replaces: #22388

### DIFF
--- a/src/boot/efi/ticks.c
+++ b/src/boot/efi/ticks.c
@@ -2,18 +2,47 @@
 
 #include <efi.h>
 #include <efilib.h>
+#if defined(__i386__) || defined(__x86_64__)
+#include <cpuid.h>
+#endif
 
 #include "ticks.h"
+
+#if defined(__i386__) || defined(__x86_64__)
+static BOOLEAN in_hypervisor(void) {
+        uint32_t eax, ebx, ecx, edx;
+
+        /* The TSC might or might not be virtualized in VMs (and thus might not be accurate or start at zero
+         * at boot), depending on hypervisor and CPU functionality. If it's not virtualized it's not useful
+         * for keeping time, hence don't attempt to use it.
+         *
+         * This is a dumbed down version of src/basic/virt.c's detect_vm() that safely works in the UEFI
+         * environment. */
+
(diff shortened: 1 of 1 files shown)


src/boot/efi/ticks.c
```diff
@@ -2,18 +2,47 @@
...
+#include <cpuid.h>
+#endif
 
 #include "ticks.h"
 
+#if defined(__i386__) || defined(__x86_64__)
+static BOOLEAN in_hypervisor(void) {
+        uint32_t eax, ebx, ecx, edx;
+
+        /* The TSC might or might not be virtualized in VMs (an thus might not be accurate or start at zero
```
yuwata: an thus -> and thus?

yuwata (approved):
LGTM. One possible typo.

poettering:
Force pushed new version with the typo fixed.

src/boot/efi/util.c
```diff
@@ -3,69 +3,9 @@
 #include <efi.h>
 #include <efilib.h>
 
+#include "ticks.h"
```
medhefgo, line 6: It would be better header hygiene to include it where it's actually needed.

src/boot/efi/ticks.c
```diff
@@ -45,6 +45,10 @@ UINT64 ticks_freq(void) {
         BS->Stall(1000);
         ticks_end = ticks_read();
 
+        if (ticks_end < ticks_start) /* Check for an overflow (which is not that unlikely, given on some
+                                      * archs the value is 32bit) */
```
medhefgo, line 81: That's some weird formatting…

medhefgo (approved):
If it works…

It's a shame the EFI timestamp protocol is utter garbage to use. I looked at this some months ago and it overflows within seconds on real hardware. Other similar protocols are just as useless :cry:.

src/boot/efi/util.c
poettering: efivar_set_time_usec() further down in util.c uses time_usec(), that's why we include it here.

src/boot/efi/ticks.c
poettering: It's how emacs line breaks this, and Emacs is God. Of course...

PR #22448, opened 2022-02-08
coredump: raise limits

poettering:


yuwata (approved):
LGTM.

src/coredump/coredump.c
```diff
@@ -48,8 +48,14 @@
 #include "uid-alloc-range.h"
```
dnicolodi, line 48: I think in the commit message "4 orders of magnitude" should be "16 times" as orders of magnitude usually are intended in base 10 😄

poettering: Are you sure that's the case?

I mean https://en.wikipedia.org/wiki/Order_of_magnitude says "usually 10", but then in the introductory paragraphs already points to the fact that one might use "2" as base in computers...

kgibm:
Thank you!

PR #22454, opened 2022-02-08
Add support for `isolated` parameter

swiryaman:
Add the "Isolated" parameter in the *.network file, e.g.,

[Bridge]
Isolated=true|false

When the Isolated parameter is true, traffic coming out of this port
will only be forward to other ports whose Isolated parameter is false.

When Isolated is not specified, the port uses the kernel default
setting (false).

The "Isolated" parameter was introduced in Linux 4.19.
See man bridge(8) for more details.
But even though the kernel and bridge/iproute2 recognize the "Isolated"
parameter, systemd-networkd did not have a way to set it.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -2961,6 +2961,15 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
           </listitem>
         </varlistentry>
         <varlistentry>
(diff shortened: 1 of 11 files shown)


yuwata (approved):
LGTM.

PR #22455, opened 2022-02-09
RFC: inline journal_file_move_to() and mmap_cache_get() top-half

vcaputo:
In commit 258190a0d5 I removed the ret_size parameter from mmap_cache_fd_get() and cleaned up journal_file_move_to_object(), since the context cache should already be doing that for us.

However, these functions are on such a hot path in operations like `journalctl -u $service` that it's actually a visible difference comparing v247 to v248 (and later).

In this series I've explicitly inlined journal_file_move_to() which is a trivial change, since `--buildtype=release` builds (-O1) on my gcc 11.1.0 host aren't automatically doing so.

But more complicatedly, I've done the necessary work to split the context cache lookup portion of mmap_cache_fd_get() into a public inline function, leaving the rest in mmap_cache_fd_get_slow().  It's a pile of commits to tease out the minimum of public structs necessary to perform context cache lookups, rather than just throwing everything into mmap-cache.h.

Here's some perf numbers taken from the last commit's message:

`journalctl -u postfix@-` timings show a ~10%+-2% runtime
improvement when cached, on a mailserver's journals.  That's for
an operation where a majority of the CPU time is spent in the
output_short() side of things:

Before any inlining:

```
 $ time LD_LIBRARY_PATH=build-main/ build-main/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

  real    0m10.716s
  user    0m10.610s
  sys     0m0.090s
```

With just inlined journal_move_to():

```
 $ time LD_LIBRARY_PATH=build-main/ build-main/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

  real    0m10.470s
  user    0m10.365s
  sys     0m0.097s
```

Now, with MMapContextCache hits performed inline too:

```
 $ time LD_LIBRARY_PATH=build-test/ build-test/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

  real    0m9.987s
  user    0m9.876s
  sys     0m0.097s
```

The target to beat, v247, when journal_move_to_object() elided the
second move_to() call:
```
 $ time LD_LIBRARY_PATH=rel-builds/v247/ rel-builds/v247/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

  real    0m10.133s
  user    0m10.015s
  sys     0m0.104s
```

### DIFF
--- a/src/libsystemd/sd-journal/mmap-cache.c
+++ b/src/libsystemd/sd-journal/mmap-cache.c
@@ -130,26 +130,8 @@ static void window_free(Window *w) {
         free(w);
 }
 
-_pure_ static bool mapping_matches(MMapMapping *m, uint64_t offset, size_t size) {
-        assert(m);
-        assert(size > 0);
-
-        return
-                offset >= m->offset &&
-                offset + size <= m->offset + m->size;
-}
-
-_pure_ static bool mapping_matches_fd(MMapMapping *m, MMapFileDescriptor *f, uint64_t offset, size_t size) {
-        assert(m);
-        assert(f);
-
-        return
-                m->fd == f &&
-                mapping_matches(m, offset, size);
-}
-
 _pure_ static bool window_matches(Window *w, uint64_t offset, size_t size) {
-        return mapping_matches(&w->mapping, offset, size);
+        return mmap_mapping_matches(&w->mapping, offset, size);
 }
 
 static Window *window_add(MMapCache *m, MMapFileDescriptor *f, bool keep_always, uint64_t offset, size_t size, void *ptr) {
@@ -393,7 +375,7 @@ outofmem:
         return -ENOMEM;
 }
 
-static int mmap_cache_fd_get_slow(
+int mmap_cache_fd_get_slow(
                 MMapFileDescriptor *f,
                 unsigned context,
                 bool keep_always,
@@ -426,40 +408,6 @@ static int mmap_cache_fd_get_slow(
 
         /* Create a new mmap */
         return add_mmap(f, c, keep_always, offset, size, st, ret);
-}
-
-int mmap_cache_fd_get(
-                MMapFileDescriptor *f,
-                unsigned context,
-                bool keep_always,
-                uint64_t offset,
-                size_t size,
-                struct stat *st,
-                void **ret) {
-
-        MMapContextCache *cc;
-        MMapContext *c;
-
-        assert(f);
-        assert(size > 0);
-        assert(ret);
-        assert(context < MMAP_CACHE_MAX_CONTEXTS);
-
-        cc = *((MMapContextCache **)f);
-        assert(cc);
-
-        /* Check whether the current context is the right one already */
-        c = &cc->contexts[context];
-        if (c->mapping && mapping_matches_fd(c->mapping, f, offset, size)) {
-                c->mapping->keep_always = c->mapping->keep_always || keep_always;
-
-                *ret = (uint8_t*) c->mapping->ptr + (offset - c->mapping->offset);
-                cc->n_context_cache_hit++;
-
-                return 1;
-        }
-
-        return mmap_cache_fd_get_slow(f, context, keep_always, offset, size, st, ret);
 }
 
 void mmap_cache_stats_log_debug(MMapCache *m) {
--- a/src/libsystemd/sd-journal/mmap-cache.h
+++ b/src/libsystemd/sd-journal/mmap-cache.h
@@ -39,7 +39,7 @@ MMapCache* mmap_cache_ref(MMapCache *m);
 MMapCache* mmap_cache_unref(MMapCache *m);
 DEFINE_TRIVIAL_CLEANUP_FUNC(MMapCache*, mmap_cache_unref);
 
-int mmap_cache_fd_get(
+int mmap_cache_fd_get_slow(
         MMapFileDescriptor *f,
         unsigned context,
         bool keep_always,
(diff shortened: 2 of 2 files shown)


vcaputo:
Had some foul weather lately preventing working outside, noticed my commit actually regressed perf a bit in v248 so this happened.

Marked as a RFC because I'm not sure how everyone feels about explicit inlining anymore, considering I've seen commits land previously deliberately going through and uninlining things promising magic from the compiler.  Also not sure how people will feel about the C struct and casting games I played in the interests of keeping most of mmap-cache.c's structs private while still enabling the inline mmap_cache_fd_get() top half.

Draft because I rushed, and didn't test much beyond the timing runs.

vcaputo:
Everybody likes pictures, kcachegrind money shots:

v247, the thing to preferably beat:

v247 xwd


v248, the thing I made worse (not mmap_cache_fd_get() and journal_file_move_to() call counts doubled):

v248 xwd

With this PR:

all-inlined xwd


These are all --buildtype-release builds, gcc 11.1.0 on Arch Linux.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -590,7 +590,8 @@ static unsigned type_to_context(ObjectType type) {
         return type > OBJECT_UNUSED && type < _OBJECT_TYPE_MAX ? type : 0;
 }
 
-static int journal_file_move_to(
+/* XXX keep this inline, this is on a very hot path for things like journalctl matches */
+static inline int journal_file_move_to(
```
poettering, line 594: as a general rule: we don't use "static inline" inside of .c files. The compiler is free to ignore the "inline" anyway, and in fact gcc doesn't really care much: it's able to detect on its own better when it's worth inlining something.

"static inline" has uses in .h files, since it allows defining inline-worthy functions in header files with full body. But outside of .h files it simply shouldn't be used, just using "static" is entirely enough, and gcc will likely inline single-use, static functions anyway.

vcaputo: Except this isn't a single use function, and it's demonstrably influenced by the `inline` qualifier here on a modern gcc in a release build, with a relevant performance impact.

So if we can't actually rely on the inline qualifier, is this an implicit vote for turning it into a macro?  Or do we not care about performance effects of function call overheads in hot paths?

poettering: Are you compiling with -O0? If you do, gcc takes things literally, to make debugging easy. But that's not how production software is built...

vcaputo: I'm just using `meson setup --buildtype=release builddir; meson compile -C builddir`, no explicit manipulation of CFLAGS is going on.  Aren't the release builds optimized? Am I holding it wrong?

vcaputo: It appears when simply using `meson setup --buildtype=release` the optimization level is `-O1`.

And glancing at Arch and Debian source packages, if those qualify as "how production software is built", simply setting the build mode seems to be the norm, unless I'm missing something.

I just want optimization levels representative of what people are likely to be experiencing in common distros out there, the whole point is to characterize and improve if needed the performance for the common case.

vcaputo: Just rebuilt v247, v248, and main+this PR with more explicit meson setup options `-Dmode=release -Ddebug=false -Doptimization=3 -D b_lto=true -D b_pie=true`, and here's the best times of four runs each:

```
$ time LD_LIBRARY_PATH=build-v247-release-O3-lto-pie/ build-v247-release-O3-lto-pie/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

real    0m10.114s
user    0m9.978s
sys     0m0.120s


$ time LD_LIBRARY_PATH=build-v248-release-O3-lto-pie/ build-v248-release-O3-lto-pie/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

real    0m10.708s
user    0m10.573s
sys     0m0.127s

~5.5% slowdown v247->v248

$ time LD_LIBRARY_PATH=build-inlines-PR-release-O3-lto-pie/ build-inlines-PR-release-O3-lto-pie/journalctl --directory rel-builds/test-journals/ -u postfix@- > /dev/null

real    0m9.848s
user    0m9.721s
sys     0m0.114s

8% speedup v248->main+this PR
```

Still significant differences there, smaller than the -O1 from `--buildtype=release`, but multiple whole percents (still nearly 10% vs. v248, but that was made worst than v247).  Callgrind is still seeing the `journal_file_move_to()` calls without this PR as well, -O3 didn't get those inlined by magic.

It's unclear to me which distros are explicitly setting `-Doptimization=3` BTW.  Maybe they have hidden stuff happening in `meson setup` wrappers, but at least in the Arch and Debian source packages what's visible is just a `-Dmode=release`.  It's annoying that we seem to be in a kind of ambiguous place with meson's `--buildtype=release,debug,plain` etc. vs `-Dmode=release,debug` etc.  The former affects optimization flags (-O1), the latter leaves them at the default of 0.

The `arch-meson` wrapper over here spits out the following:
`+ exec meson setup --prefix /usr --libexecdir lib --sbindir bin --buildtype plain --auto-features enabled --wrap-mode nodownload -D b_lto=true -D b_pie=true` and looking at https://github.com/archlinux/svntogit-packages/blob/packages/systemd/trunk/PKGBUILD there's no `-Doptimization=`. *shrug*

vcaputo:
Relevant discussion currently on lkml https://lore.kernel.org/lkml/20220217180735.GM614@gate.crashing.org/T/

In the context of this PR leaves me wondering if we should adopt the kernel's `__always_inline` macro, with the functions inlined by this PR being the first users.

vcaputo:
As this change seems to not be of enough interest, I'm closing.  I guess the assumption is that LTO/WPO will improve these code paths...

I still think making the code more explicit about which part of the MMapCache is the fast-path (the context cache) by making it more publicly directly accessible vs. the slow-path is a worthwhile change to the code, which brings a little perf boost while at it regardless of what compiler/link-time optimizations are in play.

Issue #22370, opened 2022-02-02
#20833 must be reverted (OnFailure=/OnSuccess= metadata propagation)


poettering:
#20833 contains non-trivial to fix memory corruptions, see my comments there. THis needs some more rounds of review. I would have posted a revert PR, but it already got changed since, hence this issue instead.

This really isn't ready for release, and is a blocker issue hence.

bluca:
Follow-ups here: https://github.com/systemd/systemd/pull/22378
For the rest, see comments on #20833

poettering:
#22378 addressed some of the issues that need fixing, but not all. Hence reopening.

poettering:
So I think there's some major disconnect in the discussions here. The current implementation focusses on a model where a single service unit processes OnFailure=/OnSuccess= events for a multitude of services. The problems I see all stem from that idea, i.e. the fact that the linked list will trigger a memory corruption in some cases (or that we need the linked list at all), and that events will be lost, if the handler service is already started when another unit requesting it sends out an event too, or that the environment vairable is so hard to process.

The model I always had in mind for this is that each service that shall be monitored this way will get its own private *instance* of a monitoring service. That removes *all* these issues: things become vastly simpler, we don't need linked lists, no complex env vars and we events will not get lost anymore.

i.e I think if you have a service foobar.service you want to monitor for failures you'd add a OnFailure=mymonitor@%N.service to it. If you have an instanced service quux@waldo.service you want to monitor you'd instead use OnFailure=mymonitor@%p-%i.service or so.

You then define mymonitor@.service as generic template unit that would get instantiated once for each failure service, and would get passed in the failure data as env var, without any ambiguities or complex way to parse things.

So hence, what's the rationale for the approach that was merged? Can we just rework the way I propose above? To me it appears quite clearly as the safer, cleaner approach.

poettering:
or to elaborate on this, the usecase I thought this was for was for people who want to hook failing services to an automated email. i.e. with the design I suggested you'd be able to put together a script like this:

```sh
#!/bin/sh

(
    cat <<EOF
I am sorry to report that unit $MONITOR_UNIT on $HOSTNAME failed.

Service Status
--------------
EOF

    systemctl status $MONITOR_UNIT
    cat <<EOF

Execution Environment
---------------------
EOF

    export
) | mail admin@example.com -s "Unit $MONITOR_UNIT on $HOSTNAME failed!"
```

[...]

PR #22460, opened 2022-02-09
core: split $MONITOR_METADATA and return it only if a single unit triggers OnFailure/OnSuccess

bluca:
Remove the list logic, and simply skip passing metadata if more than one
unit triggered an OnFailure/OnSuccess handler.

Fixes https://github.com/systemd/systemd/issues/22370

### DIFF
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -2459,6 +2459,7 @@ static void service_run_next_control(Service *s) {
                           EXEC_APPLY_SANDBOXING|EXEC_APPLY_CHROOT|EXEC_IS_CONTROL|
                           (IN_SET(s->control_command_id, SERVICE_EXEC_CONDITION, SERVICE_EXEC_START_PRE, SERVICE_EXEC_STOP_POST) ? EXEC_APPLY_TTY_STDIN : 0)|
                           (IN_SET(s->control_command_id, SERVICE_EXEC_STOP, SERVICE_EXEC_STOP_POST) ? EXEC_SETENV_RESULT : 0)|
+                          (IN_SET(s->control_command_id, SERVICE_EXEC_START_PRE, SERVICE_EXEC_START) ? EXEC_SETENV_MONITOR_RESULT : 0)|
                           (IN_SET(s->control_command_id, SERVICE_EXEC_START_POST, SERVICE_EXEC_RELOAD, SERVICE_EXEC_STOP, SERVICE_EXEC_STOP_POST) ? EXEC_CONTROL_CGROUP : 0),
                           &s->control_pid);
         if (r < 0)
@@ -2495,7 +2496,7 @@ static void service_run_next_main(Service *s) {
         r = service_spawn(s,
                           s->main_command,
                           s->timeout_start_usec,
-                          EXEC_PASS_FDS|EXEC_APPLY_SANDBOXING|EXEC_APPLY_CHROOT|EXEC_APPLY_TTY_STDIN|EXEC_SET_WATCHDOG,
+                          EXEC_PASS_FDS|EXEC_APPLY_SANDBOXING|EXEC_APPLY_CHROOT|EXEC_APPLY_TTY_STDIN|EXEC_SET_WATCHDOG|EXEC_SETENV_MONITOR_RESULT,
                           &pid);
         if (r < 0)
                 goto fail;
--- a/test/units/testsuite-68.sh
+++ b/test/units/testsuite-68.sh
@@ -169,8 +169,13 @@ cat >/run/systemd/system/testservice-failure-exit-handler-68.service <<EOF
 Description=TEST-68-PROPAGATE-EXIT-STATUS handle service exiting in failure
 
 [Service]
+# repeat the check to make sure that values are set correctly on repeated invocations
+Type=oneshot
+ExecStartPre=/tmp/check_on_failure.sh
 ExecStartPre=/tmp/check_on_failure.sh
 ExecStart=/tmp/check_on_failure.sh
+ExecStart=/tmp/check_on_failure.sh
+ExecStartPost=test -z '$MONITOR_SERVICE_RESULT'
 EOF
 
 # Template version.
@@ -179,9 +184,13 @@ cat >/run/systemd/system/testservice-failure-exit-handler-68-template@.service <
 Description=TEST-68-PROPAGATE-EXIT-STATUS handle service exiting in failure (template)
 
 [Service]
+Type=oneshot
 ExecStartPre=echo "triggered by %i"
 ExecStartPre=/tmp/check_on_failure.sh
+ExecStartPre=/tmp/check_on_failure.sh
 ExecStart=/tmp/check_on_failure.sh
+ExecStart=/tmp/check_on_failure.sh
+ExecStartPost=test -z '$MONITOR_SERVICE_RESULT'
 EOF
 
 systemctl daemon-reload


pdmorrow:
@bluca @poettering apologies for the extra work this has caused and many thanks for cleaning this up. Let me know if there are any other follow ups I can help with.

NEWS
```diff
@@ -9,6 +9,13 @@ CHANGES WITH 251:
           can still be used by explicitly configuring RouteTable= setting in
           .netdev files.
 
+        * $MONITOR_METADATA will only be passed to OnFailure=/OnSuccess= handlers
+          if exactly one unit was the trigger for their starting job. Therefore,
```
keszybz: Hmm, so I think this needs to be reworded, because in fact one unit is usually the trigger. We check that just one unit lists the handler unit in its OnFailure/OnSuccess.

keszybz (commented):
Looks reasonable. I readed in particular the code that creates the string, and that looks all good. So if the high-level logic is acceptable, I think this is good to merge.

NEWS
bluca: ok, reworded as suggested

poettering:
This looks much better. But the variable seems weird to me, i.e. why stash that all into one? This should be normalized, separate env vars, so that it's easy for scripts to process. i.e. this is weird:

```
MONITOR_METADATA="SERVICE_RESULT=exit-code,EXIT_CODE=exited,EXIT_STATUS=1,INVOCATION_ID=02dd868af2f344b18edaf74b618b2f90,UNIT=failure.service"
```

should be this instead:

```
MONITOR_SERVICE_RESULT=exit-code
MONITOR_EXIT_CODE=exited
MONITOR_EXIT_STATUS=1
MONITOR_INVOCATION_ID=02dd868af2f344b18edaf74b618b2f90
MONITOR_UNIT=failure.service
```

With that behaviour should be simple and systematic, and very easy to write scripts for.

bluca:
It's still easy to parse for scripts. It's an attempt to keep backward compatibility intact with v250. But if we prefer to break compatibility, I can change it.

poettering:
I am pretty sure we should create the normalized environment fields as I suggested. But I'd see that as a separate question from whethre to keep the old MONITOR_METADATA env var too.

I think we should be able to drop it though given that it was not guaranteed to exist anyway (after all, you can start units listed in some other unit's OnFailure=/OnSuccess= lines also directly, in which case the env vars would not exist). i.e. I'd probably just go this way, and drop the old name, document it loudly in NEWS (and maybe highlight in OnFailure=/OnSuccess= that the env vars will only be around if the service is a single other service's OnFailure=/OnSuccess=, and only if that service already exited, and so on, i.e. emphasize the scripts should handle gracefully if the env vars aren't there).

That said, I'd also be OK to keep th old variable in place, but only for compat, i.e. drop it from documentation, and just pass it as one more field, instead of the main conduit for the information.

NEWS
```diff
@@ -9,6 +9,13 @@ CHANGES WITH 251:
           can still be used by explicitly configuring RouteTable= setting in
           .netdev files.
 
+        * $MONITOR_METADATA will only be passed to OnFailure=/OnSuccess= handlers
+          if exactly one unit lists the handler unit as OnFailure=Onsuccess=.
+          Therefore, the content of $$MONITOR_METADATA no longer contains multiple
```
poettering: duplicate $$

NEWS
```diff
@@ -9,6 +9,13 @@ CHANGES WITH 251:
           can still be used by explicitly configuring RouteTable= setting in
           .netdev files.
 
+        * $MONITOR_METADATA will only be passed to OnFailure=/OnSuccess= handlers
+          if exactly one unit lists the handler unit as OnFailure=Onsuccess=.
```
poettering: Onsuccess → OnSuccess

bluca:
Mmh while updating the test, I noticed this change in the dependency handling breaks it, and it no longer deterministically associates the OnFailure dependency. Any idea if I am doing the atoms wrong?

keszybz:
Hmm, wouldn't it be better to keep the compat variable? The code to generate the text is a self-contained chunk and it shouldn't be a problem to keep it. We can remove it a year or two down the road.

bluca:
> Hmm, wouldn't it be better to keep the compat variable? The code to generate the text is a self-contained chunk and it shouldn't be a problem to keep it. We can remove it a year or two down the road.

I don't mind one way or another, i thought @poettering was for removal. I don't think anything is using it, codesearch shows nothing. I can leave it if it's the preference.

More importantly right now the atoms handling is borken and need to figure out why

src/core/unit-dependency-atom.h
```diff
@@ -75,18 +75,21 @@ typedef enum UnitDependencyAtom {
         /* Trigger a dependency on unsuccessful service exit. */
         UNIT_ATOM_ON_FAILURE                          = UINT64_C(1) << 24,
 
+        /* Pass exit status to a triggered OnFailure=/OnSuccess= dependency */
+        UNIT_ATOM_PROPAGATE_EXIT_STATUS               = UINT64_C(1) << 25,
```
poettering: maybe call this UNIT_ATOM_PROPAGATE_EXIT_STATUS_FROM?

i.e. the direction matters: the unit that has the dep is the one that the exit status shall be propagated to, and the unit the dep points to is where it should be propagated from.

src/core/service.c
```diff
@@ -1641,30 +1589,45 @@ static int service_spawn(
...
+        Service *env_source = NULL;
+        const char *monitor_prefix;
         if (flags & EXEC_SETENV_RESULT) {
-                if (asprintf(our_env + n_env++, "SERVICE_RESULT=%s", service_result_to_string(s->result)) < 0)
+                env_source = s;
+                monitor_prefix = "";
+        } else if (flags & EXEC_SETENV_MONITOR_RESULT) {
+                env_source = service_get_triggering_service(s);
+                monitor_prefix = "MONITOR_";
+                assert(env_source);
```
poettering: this assert might very well not be true, no?

src/core/service.c
```diff
@@ -1641,30 +1589,45 @@ static int service_spawn(
...
                                 return -ENOMEM;
                 }
 
-        } else if (flags & EXEC_SETENV_MONITOR_RESULT) {
-                Job *j = UNIT(s)->job;
-                if (j) {
-                        r = service_create_monitor_md_env(j, our_env + n_env++);
-                        if (r < 0)
-                                return r;
+                if (flags & EXEC_SETENV_MONITOR_RESULT) {
```
poettering: this might be misleading if both EXEC_SETENV_MONITOR_RESULT and EXEC_SETENV_RESULT might be set. Maybe instead check `env_source != s` here?

src/core/service.c
```diff
@@ -1584,7 +1532,7 @@ static int service_spawn(
         if (r < 0)
                 return r;
 
-        our_env = new0(char*, 10);
+        our_env = new0(char*, 11);
```
poettering: this should be 12 if i count correctly, including the trailing NULL sentinel

src/core/service.c
```diff
@@ -1443,81 +1443,29 @@ static bool service_exec_needs_notify_socket(Service *s, ExecFlags flags) {
...
+         * one to propagate the exit status. */
+
+        UNIT_FOREACH_DEPENDENCY(other, UNIT(s), UNIT_ATOM_PROPAGATE_EXIT_STATUS) {
+                assert(other->type == UNIT_SERVICE);
+
+                if (candidate) {
+                        log_unit_debug(UNIT(s), "multiple trigger source candidates for exit status propagation, skipping.");
+                        return NULL;
+                } else
+                        candidate = other;
```
poettering: you could drop the `else` and then save one level of indentation on this line

src/core/service.c
bluca: I've added this for debugging, because it doesn't show up when it should, in the test - any idea if I did anything wrong with the atoms glue?

src/core/service.c
```diff
@@ -1443,81 +1443,29 @@ static bool service_exec_needs_notify_socket(Service *s, ExecFlags flags) {
...
-                }
-
-                if (!strextend(&var, ",UNIT=", UNIT(env_source)->id))
-                        return -ENOMEM;
+         * N.B. if there are multiple services which could trigger 's' via OnFailure=
+         * or OnSuccess= then we return NULL. This is since we don't know from which
+         * one to propagate the exit status. */
+
+        UNIT_FOREACH_DEPENDENCY(other, UNIT(s), UNIT_ATOM_PROPAGATE_EXIT_STATUS) {
+                assert(other->type == UNIT_SERVICE);
```
poettering: can we actually rely on this? i don't think so.

better: just skip other units that arent' sevices, i.e.

```c
if (other->type != UNIT_SERVICE)
        continue;
```

man/systemd.exec.xml
```diff
@@ -3623,6 +3622,12 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
 SERVICE_RESULT=result-string,EXIT_CODE=exit-code,EXIT_STATUS=exit-status,INVOCATION_ID=invocation-id,UNIT=triggering.service
           </programlisting>
```
poettering: this needs updating

src/core/unit-dependency-atom.c
```diff
@@ -99,13 +99,13 @@ static const UnitDependencyAtom atom_map[_UNIT_DEPENDENCY_MAX] = {
         [UNIT_REFERENCED_BY]          = UNIT_ATOM_REFERENCED_BY,
         [UNIT_IN_SLICE]               = UNIT_ATOM_IN_SLICE,
         [UNIT_SLICE_OF]               = UNIT_ATOM_SLICE_OF,
+        [UNIT_ON_SUCCESS_OF]          = UNIT_ATOM_PROPAGATE_EXIT_STATUS,
+        [UNIT_ON_FAILURE_OF]          = UNIT_ATOM_PROPAGATE_EXIT_STATUS,
```
poettering: i think these probably should also imply referencing, i.e. shoudl also have te UNIT_REFERENCES atom. So that a failure/success handler will always pin the unit is triggered by, so that it can collect the exit data reliably.

(And the UNIT_ATOM_BACK_REFERENCE_IMPLIED atom should go away)

poettering:
did another review. looks conceptually good to me now, but needs some more polishing i guess.

src/core/service.c
poettering: nothing obvious, but see my comment about the back reference atom

src/core/service.c
bluca: yep, forgot about the NULL

src/core/service.c
bluca: Should we have a check at all, actually? Is it valid to have, say, a socket unit with OnFailure?

keszybz: I think it'd be better to allow it, because why not. One less thing to worry about.

src/core/service.c
bluca: Changed that, but no joy, the OnSuccess still shows nothing:

```
Feb 23 15:23:20 H check_on_success.sh[77]: + env
Feb 23 15:23:20 H check_on_success.sh[78]: PWD=/
Feb 23 15:23:20 H check_on_success.sh[78]: SYSTEMD_EXEC_PID=77
Feb 23 15:23:20 H check_on_success.sh[78]: LANG=C.UTF-8
Feb 23 15:23:20 H check_on_success.sh[78]: INVOCATION_ID=0de52d00d06e487d87146ccb74448d94
Feb 23 15:23:20 H check_on_success.sh[78]: SHLVL=1
Feb 23 15:23:20 H check_on_success.sh[78]: JOURNAL_STREAM=8:11272074
Feb 23 15:23:20 H check_on_success.sh[78]: PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
Feb 23 15:23:20 H check_on_success.sh[78]: _=/usr/bin/env
Feb 23 15:23:20 H check_on_success.sh[77]: + '[' '' '!=' success ']'
Feb 23 15:23:20 H check_on_success.sh[77]: + echo 'MONITOR_SERVICE_RESULT was "", expected "success"'
Feb 23 15:23:20 H check_on_success.sh[77]: MONITOR_SERVICE_RESULT was "", expected "success"
Feb 23 15:23:20 H check_on_success.sh[77]: + exit 1
```

bluca:
> did another review. looks conceptually good to me now, but needs some more polishing i guess.

Updated as requested, but it's still borked, and still no idea what I got wrong. Any suggestion?

man/systemd.exec.xml
```diff
@@ -3598,9 +3598,8 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
           triggered by <varname>OnFailure=</varname> or <varname>OnSuccess=</varname> dependencies.</para>
 
           <para>
-          The contents of this variable consists of a semi-colon separated list of metadata fields associated with the triggering
-          service. For each service which triggered the <varname>OnFailure=</varname> or <varname>OnSuccess=</varname>
-          dependency the following fields will be set:
+          The contents of this variable consists of a list of metadata fields associated with the triggering
+          service. The following fields will be set:
```
keszybz: This chunk is wrong now.

man/systemd.exec.xml
```diff
@@ -3621,6 +3620,12 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
 SERVICE_RESULT=result-string,EXIT_CODE=exit-code,EXIT_STATUS=exit-status,INVOCATION_ID=invocation-id,UNIT=triggering.service
           </programlisting>
 
+          <para>
+          Note that if multiple services trigger the same unit, no metadata will be passed all - consider
```
keszybz: "—" please.

keszybz (commented):
My abandoned review before I push…

bluca:
Thank you so much for the help! Looks great to me

test/units/testsuite-68.sh
```diff
@@ -183,27 +187,43 @@ Description=TEST-68-PROPAGATE-EXIT-STATUS handle service exiting in failure (tem
...
 
+# Let's get rid of the failed units so the tests below don't fail
+systemctl reset-failed
+
 # Test some transient units since these exit very quickly.
+: "-------III--------------------------------------------------"
 systemd-run --unit=testservice-transient-success-68 --property=OnSuccess=testservice-success-exit-handler-68.service sh -c "exit 0"
 wait_on_state_or_fail "testservice-success-exit-handler-68.service" "inactive" "10"
+
+: "-------IIII-------------------------------------------------"
```
dnicolodi, line 216: `IV` ?

keszybz: Apparently Romans used "IIII". "IV" is a medieval form. I thought it'd be nice to go back to the roots.

bluca:
agh it failed on s390x:

```
Mar 02 16:43:00 H check_on_success.sh[79]: + sort
Mar 02 16:43:00 H systemd[1]: testservice-success-exit-handler-68.service: Failed with result 'exit-code'.
Mar 02 16:43:00 H check_on_success.sh[78]: + env
Mar 02 16:43:00 H systemd[1]: testservice-success-exit-handler-68.service: Service will not restart (restart setting)
Mar 02 16:43:00 H check_on_success.sh[79]: INVOCATION_ID=9d8b37b1ac284d9daca90269b7c65355
Mar 02 16:43:00 H check_on_success.sh[79]: JOURNAL_STREAM=9:347257
Mar 02 16:43:00 H check_on_success.sh[79]: LANG=C.UTF-8
Mar 02 16:43:00 H check_on_success.sh[79]: PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
Mar 02 16:43:00 H check_on_success.sh[79]: PWD=/
Mar 02 16:43:00 H check_on_success.sh[79]: SYSTEMD_EXEC_PID=77
Mar 02 16:43:00 H systemd[1]: testservice-success-exit-handler-68.service: Changed running -> failed
Mar 02 16:43:00 H check_on_success.sh[77]: + [  != success ]
Mar 02 16:43:00 H check_on_success.sh[77]: + echo MONITOR_SERVICE_RESULT was '', expected 'success'
Mar 02 16:43:00 H check_on_success.sh[77]: MONITOR_SERVICE_RESULT was '', expected 'success'
Mar 02 16:43:00 H check_on_success.sh[77]: + exit 1
```

keszybz:
That's strange. Let's see if the other arches fail too.

bluca:
`Mar 02 16:43:00 H systemd[1]: testservice-success-exit-handler-68.service: multiple trigger source candidates for exit status propagation (testservice-transient-success-68.service, testservice-success-68.service), skipping.`

bluca:
I guess now we have the opposite problem, and we don't garbage collect in time? I'll add a daemon-reload after reset-failed to force collection

keszybz:
OK, I think this is acceptable and can be merged.

The model where the same cleanup unit is used for multiple service just cannot work too reliably. I thought about enhancing `service_get_triggering_service` to ignore services which haven't failed, but I don't think we can get it to work: we could fix those issues in the testsuite, but it'd introduce wrong behaviour in other cases. In particular, you can have multiple handler services attached to one unit, and they may restart the failed unit, or do reset-failed on it, and we just can't properly keep track of all this if we have a many-to-many mapping. I think it's reasonable to simply push people to use templated cleanup units if they want to watch more than one unit.

bluca:
> OK, I think this is acceptable and can be merged.
> 
> The model where the same cleanup unit is used for multiple service just cannot work too reliably. I thought about enhancing `service_get_triggering_service` to ignore services which haven't failed, but I don't think we can get it to work: we could fix those issues in the testsuite, but it'd introduce wrong behaviour in other cases. In particular, you can have multiple handler services attached to one unit, and they may restart the failed unit, or do reset-failed on it, and we just can't properly keep track of all this if we have a many-to-many mapping. I think it's reasonable to simply push people to use templated cleanup units if they want to watch more than one unit.

Thanks, sounds good to me. @poettering any comment?

bluca:
Latest push is just a rebase, to fix a merge conflict in the NEWS file, no other changes.

src/core/service.c
```diff
@@ -2465,6 +2465,7 @@ static void service_run_next_control(Service *s) {
                           EXEC_APPLY_SANDBOXING|EXEC_APPLY_CHROOT|EXEC_IS_CONTROL|
                           (IN_SET(s->control_command_id, SERVICE_EXEC_CONDITION, SERVICE_EXEC_START_PRE, SERVICE_EXEC_STOP_POST) ? EXEC_APPLY_TTY_STDIN : 0)|
                           (IN_SET(s->control_command_id, SERVICE_EXEC_STOP, SERVICE_EXEC_STOP_POST) ? EXEC_SETENV_RESULT : 0)|
+                          (IN_SET(s->control_command_id, SERVICE_EXEC_START_PRE, SERVICE_EXEC_START) ? EXEC_SETENV_MONITOR_RESULT : 0)|
```
poettering, line 2462: hmm, why is this in this commit? shouldn't this be separate?

src/core/service.c
```diff
@@ -1481,6 +1481,7 @@ static Service *service_get_triggering_service(Service *s) {
 }
 
 static int service_spawn(
+                const char *reason,
```
poettering: hmm, "reason", is that the right word? Maybe call it `caller`? Because that is what is passed here?

src/core/service.c
```diff
@@ -1911,7 +1915,8 @@ static void service_enter_stop_post(Service *s, ServiceResult f) {
         if (s->control_command) {
                 s->control_command_id = SERVICE_EXEC_STOP_POST;
 
-                r = service_spawn(s,
+                r = service_spawn(__func__,
```
poettering: maybe instead rename service_spawn() to service_spawn_internal() and then just do:

```c
#define service_spawn(...) service_spawn_internal(__func__, __VA_ARGS__)
```

that would make the change a lot more minimal, and add less noise to the invocations.

man/systemd.exec.xml
```diff
@@ -4100,33 +4090,31 @@ ExecStart=/bin/mysecondprogram
...
         then <filename index="false">myhandler.service</filename> would be triggered and the
-        <varname>$MONITOR_METADATA</varname> variable would be set as follows:</para>
+        monitor variables would be set as follows:</para>
 
         <programlisting>
-MONITOR_METADATA=SERVICE_RESULT=result-string,EXIT_CODE=exit-code,EXIT_STATUS=exit-status,INVOCATION_ID=invocation-id,UNIT=myfailer.service
+MONITOR_SERVICE_RESULT=result-string
+MONITOR_EXIT_CODE=exit-code
+MONITOR_EXIT_STATUS=exit-status
+MONITOR_INVOCATION_ID=invocation-id
```
poettering: maybe just specify some random id here, for illustrative purposes?

man/systemd.exec.xml
```diff
@@ -4100,33 +4090,31 @@ ExecStart=/bin/mysecondprogram
...
-        would be a list containing exit metadata for both of <filename index="false">myfailer.service</filename>
-        and <filename index="false">mysuccess.service</filename>.</para>
+        monitor variables would be set as follows:</para>
 
         <programlisting>
-MONITOR_METADATA=SERVICE_RESULT=result-string,EXIT_CODE=exit-code,EXIT_STATUS=exit-status,INVOCATION_ID=invocation-id,UNIT=myfailer.service;SERVICE_RESULT=result-string,EXIT_CODE=exit-code,EXIT_STATUS=exit-status,INVOCATION_ID=invocation-id,UNIT=mysuccess.service
+MONITOR_SERVICE_RESULT=result-string
+MONITOR_EXIT_CODE=exit-code
+MONITOR_EXIT_STATUS=exit-status
+MONITOR_INVOCATION_ID=invocation-id
```
poettering: ditto

poettering:
looks great. thanks for looking into this. sorry for taking so long to review. just some nits from my side now.

src/core/service.c
bluca: The commit handles both the issue and enhances the test for check for it:

>   The first ExecStartPre or the first ExecStart commands would get the metadata,
>   but not the subsequent ones. Also check that we do not pass it in
>   ExecStartPost.

It fixes a few other things, I'll split it into two.

test/units/testsuite-68.sh
bluca: my whole school education has been a lie

PR #22461, opened 2022-02-09
ci: remove MULTI_STATUS from superlinter

evverx:
Judging by
```
ERROR! Failed to call GitHub Status API!
```
it doesn't seem to work. Even if it did it would just clutter the status
checks I think so let's just remove MULTI_STATUS along with
GITHUB_TOKEN.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -40,8 +40,6 @@ jobs:
           #   - .*\.(in|SKELETON) - all template/skeleton files
           #   - tools/coverity\.sh - external file (with some modifications)
           FILTER_REGEX_EXCLUDE: .*/(man/.*|src/kernel-install/.*|.*\.(in|SKELETON)|tools/coverity\.sh)$
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          MULTI_STATUS: true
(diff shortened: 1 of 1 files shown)


mrc0mmand:
LGTM. I couldn't make it work either, but I guess we should be fine without the multi-status anyway.

mrc0mmand (approved):

.github/workflows/linter.yml
```diff
@@ -40,8 +40,6 @@ jobs:
           #   - .*\.(in|SKELETON) - all template/skeleton files
           #   - tools/coverity\.sh - external file (with some modifications)
```
evverx, line 41: @mrc0mmand I'm not sure `tools/coverity.sh` should be ignored. It's kind of convoluted though. I'd just replace it with something like https://github.com/util-linux/util-linux/commit/c881bec5fe10eff9cf06e2db4dc479d798f9eae6#diff-b08ad0c56cb48fcac947f04aa72b28c7f500e1accf1c827831e693ecee0474fc and unblock superlinter. I'm not sure why I haven't done it here yet :-)

evverx: I tried to change it and broke something. Probably it is wise to ignore it as long as it works :-)

PR #22462, opened 2022-02-09
network: support Match.Firmware= in .link files

sipraga:
In cbcdcaaa0ec5 ("Add support for conditions on the machines firmware")
a new Firmware= directive was added for .netdev and .network files.
While it was also documented to work on .link files, in actual fact the
support was missing. Add that one extra line to make it work.


bluca:
Needs to be added to `test/fuzz/fuzz-link-parser/directives.link` and `test/fuzz/fuzz-unit-file/directives-all.service`

yuwata (approved):
LGTM.

Issue #22379, opened 2022-02-02
KERNEL_INSTALL_LAYOUT needs love


poettering:
See my comments on #19006.

Questions: where is this actually configured? /etc/machine-info? /etc/kernel/install.conf?

(I am pretty sure it should not be /etc/machine-info, that's supposed to carry machine information, not boot loader details)

this needs documentation, i find this all utterly confusing, and kinda chickening out...

So, I'd like to take a step back from all this. What is the actual problem here? That looking for existance of the $BOOT/loader/entries/ dir is not enought to detect if a spec-compliant boot loader is installed, because the fedora/grub people squatted that directory with their own incompatible stuff?

maybe the right approach to work around that is to make bootctl install add a file `$BOOT/loader/standard` or so that carries an identifier for the used layout.

The thing with the approach of storing that info in /etc is that breaks with one of the fundamental goals of the spec: that $BOOT is a shared space and different OSes can all drop their own stuff in there. If I need to look at /etc to know what to place there we lost already. The data hence belongs in $BOOT instead.

Hence my proposal: deprecate/remove $KERNEL_INSTALL_LAYOUT and instead just use `$BOOT/loader/standard` for figuring this out.

poettering:
So coming back to this, I think there are two things we should distinguish:

1. which layout is actually understood by the installed boot loader
2. which layout is supposed to be generated by an OS' kernel-install script

sd-boot supports two layouts: boot loader spec type `#1` (i.e. .conf snippets) and boot loader spec type `#2` (i.e. .efi unified kernels). Our upstream `90-loaderentry.install`  can generated only type `#1`. Type `#1` is called `bls` in the layout field (`bls` I think is a pretty poor choice of name for this, but that ship has sailed...)

grub on fedora supports at least one layout: let's call it `frankenbootloaderspec`, i.e. something where they took our upstream spec for type `#1` entries and then added variable expansion and whatnot to it (which I think is a tremendously bad idea, and totally misses the point of the boot loader spec, but it is how it is).

Now, I think there are a couple of relevant questions to ask:

1. if I install a new OS and it wants to add its entries to the existing boot loader's menu, how does it figure out what it can drop in?

2. if multiple layouts are supported by the installed boot loader (e.g. type `#1` + `#2`), how is the choice made by kernel-install which to generate? this is about two things: which config files, and which automatic fallback choice.

3. what do no boot loader is found, but kernel-install is already run, what should it do? do nothing? or generate an entry by one specific layout?

[...]

PR #22463, opened 2022-02-09
kernel-install: clean-up machine-id handling 

poettering:
Here's an (incomplete) PR implementing some of my ideas from the #22376 discussion.

This does the following: 

1. instead of auto-persisting the used prefix dir name in /etc/machine-info we'll now add it to the kernel cmdline in generated snippets via the systemd.machine_id= kernel cmdline option that PID 1 uses for initializing /etc/machine-id if it's not initialized yet.

2. It separates $MACHINE_ID from a new variable $ENTRY_TOKEN. The former is strictly a 128bit id, and thus useful in the systemd.machine_id= kernel cmdline option or for the "machine-id" field of Type 1 boot loader spec snippets (where we previously would possible pass garbage data). The latter is the token string we use for actually naming the boot loader spec Type 1 entry .conf files or for naming the kernel image/initrd directory for this OS install.

3. It then links up the two concepts: the systemd.machine_id= field is only generated in case $MACHINE_ID and $ENTRY_TOKEN actually are the same, i.e in which case the machine ID is necessary to identify the boot loader entries we generate.

This is not complete yet. Specifically:

1. This needs documentation, and I think KERNEL_INSTALL_MACHINE_ID should probably not be documented anymore (not because it's inherently evil, but mostly because needlessly confusing). In particular this adds a new /etc/kernel/entry-token file, which needs documentation. 

2. bootctl should get a new switch to explicitly configure which entry token to use during install, i.e. whether to use the machine ID (as originally) or whether to use something else, for example the image-ID. Primary effect of this switch would be that the string is acquired and then written to /etc/kernel/entry-token. Idea would be that if mkosi generates golden images it does "bootctl install --entry-token=image-id" or so, while by default (or via explicit "bootctl install --entry-token=machine-id") you'll get the entries/dirs named after the mchine ID instead.

This is only superficially tested. I figure we have no test suite right now for any of this, so this needs some manual testing love I guess.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -21,13 +21,42 @@ CHANGES WITH 251:
           for breaking backward compatibility instead, as nobody should be
           affected, given the state of the current interface.
 
-        * Service monitor environment variables will only be passed to OnFailure=/OnSuccess=
-          handlers if exactly one unit lists the handler unit as OnFailure=/OnSuccess=.
-          Therefore, $MONITOR_METADATA is no longer used, and instead separate
-          variables are used: $MONITOR_SERVICE_RESULT, $MONITOR_EXIT_CODE,
-          $MONITOR_EXIT_STATUS, $MONITOR_INVOCATION_ID and $MONITOR_UNIT.
-          For cases when a single handler needs to watch multiple units, use a
-          templated handler.
+        * Service monitor environment variables will only be passed to
+          OnFailure=/OnSuccess= handlers if exactly one unit lists the handler
+          unit as OnFailure=/OnSuccess=.  Therefore, $MONITOR_METADATA is no
+          longer used, and instead separate variables are used:
+          $MONITOR_SERVICE_RESULT, $MONITOR_EXIT_CODE, $MONITOR_EXIT_STATUS,
+          $MONITOR_INVOCATION_ID and $MONITOR_UNIT.  For cases when a single
+          handler needs to watch multiple units, use a templated handler.
+
+        * kernel-install's and bootctl's Boot Loader Specification Type #1
+          entry generation logic has been reworked. The user may now pick
+          explicitly by which "token" string to name the installation's boot
+          entries, through the new /etc/kernel/entry-token file or the new
+          --entry-token= switch to bootctl. By default — as before — the
+          entries are named after the local machine ID. However, in "golden
+          image" environments, where the machine ID shall be initialized on
+          first boot (as opposed to at installation time before first boot) the
+          machine ID is not be available at build time to name the entry
+          after. In this case the the --entry-token= switch to bootctl (or the
+          /etc/kernel/entry-token file) may be used to override the "token" to
+          identify the entry by, and use another ID, for example the IMAGE_ID=
+          or ID= fields from /etc/os-release. This will make the OS images
+          independent of any machine ID, and ensure that the images will not
+          carry any identifiable information before first boot, but on the
+          other hand means that multiple parallel installations of the very
+          same image on the same disk cannot be supported. Summary: if you are
+          building golden images that shall acquire identity information
+          exclusively on first boot, make sure to both remove /etc/machine-id
+          *and* to write /etc/kernel/entry-token to the value of the IMAGE_ID
+          or ID field of /etc/os-release or another suitable identifier before
+          deploying the image.
+
+        * The --make-machine-id-directory= switch to bootctl has been replaced
+          by --make-entry-directory=, given that the entry directory is not
+          necessarily named after the machine ID, but after some other suitable
+          ID as selected via --entry-token= described above. The old name of
+          the option is still understood to maximize compatibility.
 
         * Services with Restart=always and a failing ExecCondition= will no longer
           be restarted, to bring ExecCondition= in line with Condition*= settings.


DaanDeMeyer:
Ahh, I also did some stuff in https://github.com/systemd/systemd/pull/22421, did you miss that one? I don't mind closing that one in favor of this one

src/kernel-install/90-loaderentry.install
```diff
@@ -68,17 +69,26 @@ elif [ -r /usr/lib/kernel/cmdline ]; then
...
     BOOT_OPTIONS="$(tr -s "$IFS" '\n' </proc/cmdline | grep -ve '^BOOT_IMAGE=' -e '^initrd=' | tr '\n' ' ')"
 fi
+
 BOOT_OPTIONS="${BOOT_OPTIONS% }"
 
+# If the boot entries are named after the machine ID, then suffix the kernel
+# command line with the machine ID we use, so that the machine ID remains
+# stable, even during factory reset, in the initrd (where the system's machine
+# ID is not directly accessible yet), and if the root file system is volatile.
+if [ "$ENTRY_TOKEN" = "$MACHINE_ID" ] ; then
```
behrmann:
```suggestion
if [ "$ENTRY_TOKEN" = "$MACHINE_ID" ]; then
```
to have the same style as (most of) the rest

src/kernel-install/kernel-install
```diff
@@ -89,17 +89,26 @@ elif [ -r "/usr/lib/kernel/install.conf" ]; then
...
+[ -z "$MACHINE_ID" ] && [ -r /etc/machine-info ] && . /etc/machine-info && MACHINE_ID="$KERNEL_INSTALL_MACHINE_ID"
+[ -z "$MACHINE_ID" ] && [ -r /etc/machine-id ]   && read -r MACHINE_ID </etc/machine-id
+[ -z "$MACHINE_ID" ] && MACHINE_ID="$(systemd-id128 new)"
+
+# Now that we determined the machine ID to use, let's determine the "token" for
+# the boot loader entry to generate. We use that for naming the directory below
+# $BOOT where we want to place the kernel/initrd and related resources, as well
+# for naming the .conf boot loader spec entry. Typically this is just the
+# machine ID, but it can be anything else, too, if we are told so.
+[ -z "$ENTRY_TOKEN" ] && [ -r /etc/kernel/entry-token ] && read -r ENTRY_TOKEN < /etc/kernel/entry-token
```
behrmann:
```suggestion
[ -z "$ENTRY_TOKEN" ] && [ -r /etc/kernel/entry-token ] && read -r ENTRY_TOKEN </etc/kernel/entry-token
```

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
-        Partitions, or that multiple OS instances are not possible.</para></listitem>
+        <para>If set to <option>os-id</option> the entries are named after the OS ID of the running system,
+        i.e. the <varname>ID=</varname> field of
+        <citerefentry><refentrytitle>os-release</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+        (e.g. <literal>fedora</literal>). Similar, if set to <option>os-image-id</option> the entries are
+        named after the OS image ID of the running system, i.e. the <varname>IMAGE_ID=</varname> field of
+        <filename>os-release</filename> (e.g. <literal>vendorx-cashier-system</literal>).</para>
+
+        <para>If set to <option>auto</option> (the default) it is first attempted to read the
+        <filename>/etc/kernel/entry-token</filename> file if it exists, and used the stored entry token
```
behrmann:
```suggestion
        <filename>/etc/kernel/entry-token</filename> file if it exists, and use the stored entry token
```

src/kernel-install/kernel-install
```diff
@@ -72,13 +73,17 @@ else
     [ $# -ge 1 ] && shift
 fi
 
-if [ $# -lt 1 ]; then
-    echo "Error: not enough arguments" >&2
-    exit 1
-fi
+if [ "$COMMAND" = "inspect" ] ; then
```
behrmann:
```suggestion
if [ "$COMMAND" = "inspect" ]; then
```

src/kernel-install/kernel-install
```diff
@@ -89,44 +94,75 @@ elif [ -r "/usr/lib/kernel/install.conf" ]; then
...
+[ -z "$MACHINE_ID" ] && [ -r /etc/machine-id ]   && read -r MACHINE_ID </etc/machine-id
+[ -z "$MACHINE_ID" ] && MACHINE_ID="$(systemd-id128 new)"
+
+# Now that we determined the machine ID to use, let's determine the "token" for
+# the boot loader entry to generate. We use that for naming the directory below
+# $BOOT where we want to place the kernel/initrd and related resources, as well
+# for naming the .conf boot loader spec entry. Typically this is just the
+# machine ID, but it can be anything else, too, if we are told so.
+[ -z "$ENTRY_TOKEN" ] && [ -r /etc/kernel/entry-token ] && read -r ENTRY_TOKEN < /etc/kernel/entry-token
+if [ -z "$ENTRY_TOKEN" ] ; then
```
behrmann:
```suggestion
if [ -z "$ENTRY_TOKEN" ]; then
```

src/kernel-install/kernel-install
```diff
@@ -89,44 +94,75 @@ elif [ -r "/usr/lib/kernel/install.conf" ]; then
...
+    ENTRY_TOKEN_SEARCH="$ENTRY_TOKEN_SEARCH Default"
+else
+    ENTRY_TOKEN_SEARCH="$ENTRY_TOKEN"
+fi
+
+# NB: The $MACHINE_ID is guaranteed to be a valid machine ID, but
+#     $ENTRY_TOKEN can be any string that fits into a VFAT filename, though
+#     typically is just the machine ID.
+
+[ -z "$BOOT_ROOT" ] && for suff in $ENTRY_TOKEN_SEARCH ; do
```
behrmann:
```suggestion
[ -z "$BOOT_ROOT" ] && for suff in $ENTRY_TOKEN_SEARCH; do
```

src/kernel-install/kernel-install
```diff
@@ -89,44 +94,75 @@ elif [ -r "/usr/lib/kernel/install.conf" ]; then
...
     for pref in "/efi" "/boot" "/boot/efi" ; do
         if [ -d "$pref/$suff" ]; then
             BOOT_ROOT="$pref"
+            ENTRY_TOKEN="$suff"
             break 2
         fi
     done
 done
 
+[ -z "$BOOT_ROOT" ] && for pref in "/efi" "/boot" "/boot/efi" ; do
```
behrmann:
```suggestion
[ -z "$BOOT_ROOT" ] && for pref in "/efi" "/boot" "/boot/efi"; do
```

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+        string. Otherwise if the local machine ID is initialized it is used. If it is not initialized the
+        <varname>IMAGE_ID=</varname> or <varname>ID=</varname> fields of <filename>os-release</filename> are
+        used, in this order of precedence, if set.</para>
```
behrmann: The fallback "Default" is not mentioned here.

poettering: "Default" is a fallback used by kernel-install, not bootctl.

man/kernel-install.xml
```diff
@@ -163,36 +175,51 @@
...
+
+    <para><varname>$KERNEL_INSTALL_ENTRY_TOKEN</varname> is set for the plugins to the desired entry "token"
+    to use. It's an identifier that shall be used to identify the local installation, and is often the
+    machine ID, i.e. equivalent to <varname>$KERNEL_INSTALL_MACHINE_ID</varname>, but might also be a
+    different type of identifier, for example a fixed string or the <varname>ID=</varname>,
+    <varname>IMAGE_ID=</varname> values from <filename>/etc/os-release</filename>. The string passed here
+    should be used to name Boot Loader Specification entries, or the directories the kernel image and initial
+    RAM disk images are placed into. Note that while oftentimes
+    <varname>$KERNEL_INSTALL_ENTRY_TOKEN</varname> and <varname>$KERNEL_INSTALL_MACHINE_ID</varname> are set
+    to the same values the latter is guaranteed to be in 128bit hexadecimal character form while the fomrer
```
behrmann:
```suggestion
    to the same values the latter is guaranteed to be in 128bit hexadecimal character form while the former
```

poettering:
(@behrmann's points should be addressed already, thanks for the review!)

man/bootctl.xml
behrmann: Oh, that's true, missed where I was. That being said, There's only three instances of "Default" in [kernel-install.xml](https://github.com/systemd/systemd/blob/3616bf9df28638368036dcec29895aea629a9b3a/man/kernel-install.xml) and they don't describe the fallback, if I didn't miss anything.

poettering:
(btw, I think we should try to add optional boot loader spec type 2 support to all this. Primarily to get a second implementation into place, since abstracting over a single backend is kinda doomed to fail. However, that probably requires changes to the dracut RPM, we can't implement this here on our own. i.e. we have to make sure dracut generates its initrd, and the pick all initrds up and pack them together with objcopy, and maybe sign the result – or alternatively dracut does that all for us)

man/bootctl.xml
poettering: So I am not fully sure "Default" is actually a great idea in retrospective. using $ID/$IMAGE_ID is probably a much better idea, and should not be worse in any way. Hence I am not sure we should document that (even though we honour, for compat reasons).

However, I have updated the docs now, to make clear that kernel-install reads /etc/kernel/entry-token first, and then searchs a few candidate dirs in $BOOT, without specifying which one those are. I think this should be enough.

poettering:
Would be good to get a thorough review on this, as I am not at home in shell as I am in C, and we don't have a test suite for this.

mrc0mmand:
Impressive, we should be able to enable shellcheck for the kernel-install stuff now:

```diff
$ git diff
diff --git a/.github/workflows/linter.yml b/.github/workflows/linter.yml
index 5424f4d6ea..16a9a7ba47 100644
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -35,11 +35,9 @@ jobs:
           # Excludes:
           #   - man/.* - all snippets in man pages (false positives due to
           #              missing shebangs)
-          #   - kernel-install/.* - false-positives for dropins (and I'm afraid
-          #                         to touch kernel-install)
           #   - .*\.(in|SKELETON) - all template/skeleton files
           #   - tools/coverity\.sh - external file (with some modifications)
-          FILTER_REGEX_EXCLUDE: .*/(man/.*|src/kernel-install/.*|.*\.(in|SKELETON)|tools/coverity\.sh)$
+          FILTER_REGEX_EXCLUDE: .*/(man/.*|.*\.(in|SKELETON)|tools/coverity\.sh)$
           VALIDATE_ALL_CODEBASE: false
           VALIDATE_BASH: true
           VALIDATE_GITHUB_ACTIONS: true
diff --git a/src/kernel-install/kernel-install b/src/kernel-install/kernel-install
index d6d4dca709..322c37c20b 100755
--- a/src/kernel-install/kernel-install
+++ b/src/kernel-install/kernel-install
@@ -89,8 +89,10 @@ layout=
 initrd_generator=
 
 if [ -r "/etc/kernel/install.conf" ]; then
+    # shellcheck source=src/kernel-install/install.conf
     . /etc/kernel/install.conf
 elif [ -r "/usr/lib/kernel/install.conf" ]; then
+    # shellcheck source=src/kernel-install/install.conf
     . /usr/lib/kernel/install.conf
 fi
 
@@ -98,6 +100,7 @@ fi
 # generated one. If the user configured an explicit machine ID to use in
 # /etc/machine-info to use for our purpose, we'll use that instead (for
 # compatibility).
+# shellcheck source=/dev/null
 [ -z "$MACHINE_ID" ] && [ -r /etc/machine-info ] && . /etc/machine-info && MACHINE_ID="$KERNEL_INSTALL_MACHINE_ID"
 [ -z "$MACHINE_ID" ] && [ -r /etc/machine-id ]   && read -r MACHINE_ID </etc/machine-id
 [ -z "$MACHINE_ID" ] && MACHINE_ID="$(systemd-id128 new)"
```

Also, I noticed this:
```
$ shellcheck -x src/kernel-install/90-loaderentry.install 

In src/kernel-install/90-loaderentry.install line 21:
shopt -s nullglob
^---------------^ SC3044: In POSIX sh, 'shopt' is undefined.

For more information:
  https://www.shellcheck.net/wiki/SC3044 -- In POSIX sh, 'shopt' is undefined.
```

introduced by 367165a4069ac0c04882a05a8a80f6afb1e42760, which is actually true - we use `/bin/sh` and `shopt` is bashism.

src/kernel-install/kernel-install
```diff
@@ -89,44 +94,75 @@ elif [ -r "/usr/lib/kernel/install.conf" ]; then
...
-
-[ -z "$BOOT_ROOT" ] && for suff in "$MACHINE_ID" "loader/entries"; do
-    for pref in "/efi" "/boot" "/boot/efi" ; do
+# If /etc/machine-id is initialized we'll use it, otherwise we'll use a freshly
+# generated one. If the user configured an explicit machine ID to use in
+# /etc/machine-info to use for our purpose, we'll use that instead (for
+# compatibility).
+[ -z "$MACHINE_ID" ] && [ -r /etc/machine-info ] && . /etc/machine-info && MACHINE_ID="$KERNEL_INSTALL_MACHINE_ID"
+[ -z "$MACHINE_ID" ] && [ -r /etc/machine-id ]   && read -r MACHINE_ID </etc/machine-id
+[ -z "$MACHINE_ID" ] && MACHINE_ID="$(systemd-id128 new)"
```
daandemeyer, line 103: Shouldn't we drop this if we're not going to persist it? We'll end up creating a directory in /efi, /boot with no way to trace it back to a specific installation.

DaanDeMeyer:
Wouldn't this introduce implicit coupling between bootctl and kernel-install where we require bootctl to run first and before kernel-install. We already know that package managers run kernel-install as a post installation script so when bootstrapping a system, if we have to manually run bootctl to configure the correct directory for kernel-install to use, we'll have to either go through hoops to disable kernel-install from running on package installation so we can run it manually later. Or we have to manually remove the stuff it added during package installation so we can run it again later once we've run bootctl.

If we'd also install systemd-boot as a post install script, you'd need kernel-install to depend on systemd-boot or something like that to make sure bootctl runs before kernel-install. But if bootctl would be executed as a post install script, there'd be no way to configure it using command line switches.

Because there might be stuff like package managers automatically running either bootctl or (more likely) kernel-install, It's nice to be able to configure these using a configuration file because those can be picked up even if the tool is executed automatically in a post install script or something similar. In mkosi's case we'd just have to write the configuration file before running the package manager which seems a lot simpler than fiddling with how kernel-install/bootctl are executed by the package manager itself.

It seems safer and less error-prone to me if instead on relying bootctl to have been run first with the correct options, we could also explicitly configure kernel-install to install to a specific directory using the same options that would be implemented in bootctl. Can't we use the install.conf config file that we added for kernel-install to allow specifying the same options as bootctl so we can run kernel-install without needing to run bootctl first?

src/kernel-install/kernel-install
poettering: But we *do* persist it! But not in /etc but in the generated boot entry snippets, i.e. via the systemd.machine_id= kernel cmdline switch! That way, if you do the factory reset all of /etc/ can be flushed out, but you will still get the same machine ID once the machine is back up, and thus you will then also be able to recognize the generated boot entry snippets again.

That's the magic of it: machine is persisted without being persisted ;-)

poettering:
The fact that bootctl *can* create the dir for you doesn't mean you have to use it for that. But yes, if kernel-install drops boot entrys *before* anyone explicitly configures the entry token, then of course, it will use the default machine ID. But that's kinda intended.

So you basically have three choices:

1. Run bootctl first, then install your kernels (as you suggested)
2. Just use the default entry token that is the machine-id, in which case you have nothing to do.
3. do "echo whatever > /etc/kernel/entry-token" first, to tell kernel-install what to drop the kernels into, then install your kernels, and afterwards to "bootctl install" to actully install the boot loader

mkosi should use the third option.

Configuring the entry token is *extremely* simple, hence if you don't want to use the machine-id, use whatever you like and put it in the OS tree before the kernel-install stuff gets run, it doesn#t have to be bootctl that is used for that.

poettering:
(Point i am making: /etc/kernel/entry-token is just a config file. One we can write for you, but you can also do that yourself, with the tool/editor of your choice)

src/kernel-install/kernel-install
daandemeyer: What I'm worried about is that this ends up coupling the machine ID to the installation directory in the ESP quite heavily. Unless mkosi starts writing /etc/kernel/entry-token, images will end up with a systemd.machine_id snippet in the kernel cmdline, which will cause all instances of the image to have the same machine ID suddenly. The way I understand it, the most likely scenario where a machine ID won't be available is when building an image for distribution and that's exactly the scenario where we don't want to persist the machine ID to be persisted but instead want to have it generated on boot. So I figure most scenarios where this logic ends up being executed will be scenarios where we don't actually want it to execute (and the reason it'll be executed will be because someone unfamiliar with the space won't have written /etc/kernel/entry-token). 

The ideal behavior to me would be to have stuff get written under the IMAGE_ID/ID dir if IMAGE_ID/ID is configured, or otherwise "Default" if no machine ID is available when building a golden image.

So to summarize, I'm worried that if we ship this like is, we'll end up breaking compat for image builders since suddenly all instances of an image will end up with a fixed machine ID if you run kernel-install with no machine ID available and nothing in /etc/kernel/entry-token.

poettering: But given the third commit the machine ID is only included in the cmdline that is inserted in the snippet if the snippet is actually placed in the dir named after the machine ID! 

Thus: if you say "yes" (the default) to the machine ID then you'll get snippets with the name ID in *both* path *and* contents. – But if you say "no" then you will get it *neither* in path *nor* in contents.

Or in other words: either you don't want machine IDs in boot stuff then we won't do that *at* *all* or you are fine with machine  ID in boot stuff, and then we use it at both places. I think that's an absolutely reasonable approach, no? What would the value be of omitting it in contents but still having it in the paths of these snippets? That would get you zero benefits!

poettering: > The ideal behavior to me would be to have stuff get written under the IMAGE_ID/ID dir if IMAGE_ID/ID is configured, or otherwise "Default" if no machine ID is available when building a golden image.

Well, yes, gold image builders should do that. But we cannot auto-detect this case, and so far we always use the real machine-id. This change cleans things up, and allows you to clearly pick the right behaviour.

> So to summarize, I'm worried that if we ship this like is, we'll end up breaking compat for image builders since suddenly all instances of an image will end up with a fixed machine ID if you run kernel-install with no machine ID available and nothing in /etc/kernel/entry-token.

But you already end up with exactly that because we use the machine ID in the path to the snippets! So nothing is lost here.

daandemeyer: I might be wrong here, but what I'm thinking of is the following scenario where we do type 1 images in mkosi (without the 250 stuff):

- We configure a machine ID during the image build process because dracut doesn't work without one
- kernel-install runs as a post install script during the image build process and uses the machine ID we generated to install stuff to /efi
- We remove /etc/machine-id at the end of the build process so the generated image doesn't ship with a machine ID configured. When an instance of the image boots, the machine ID gets generated on boot.
- If kernel-install is run again after the first boot, it creates a new directory in /efi named after the final machine ID and installs stuff under that. We end up with 2 directories in /efi containing stuff from kernel-install.

What will happen with this patch included and without changes to mkosi:

- We configure a machine ID during the image build process because dracut doesn't work without one
- kernel-install runs as a post install script during the image build process and uses the machine ID we generated to install stuff to /efi and also includes systemd.machine_id in the kernel cmdline with the machine ID we generated
- We remove /etc/machine-id at the end of the build process so the generated image doesn't ship with a machine ID configured. When an instance of the image boots, the machine ID configured in systemd.machine_id gets used as the machine ID. This will be the same for every instance of the image that's booted.
- If kernel-install is run again after the first boot, it creates a new directory in /efi named after the final machine ID and installs stuff under that. Since the final machine ID is the same as the initial one we generated in mkosi, we end up with a single directory under /efi containing machine ID stuff.

What will happen with this patch included and with changing mkosi to have no machine ID during the build process:

- kernel-install runs as a post install script during the image build process and generates a random machine ID to install stuff to /efi and also includes systemd.machine_id in the kernel cmdline with the machine ID we generated
- We remove /etc/machine-id at the end of the build process so the generated image doesn't ship with a machine ID configured. When an instance of the image boots, the machine ID configured in systemd.machine_id gets used as the machine ID. This will be the same for every instance of the image that's booted.
- If kernel-install is run again after the first boot, it creates a new directory in /efi named after the final machine ID and installs stuff under that. Since the final machine ID is the same as the initial one we generated in mkosi, we end up with a single directory under /efi containing machine ID stuff.

So what I'm worried about is that unless image builders very explicitly configure the directory to use by either running bootctl with the right option first or write to /etc/kernel/entry-token first, they'll end up with a very wrong configuration (same machine ID for every instance of booted images) that isn't immediately obvious and probably won't break things very noticeably in the beginning, but can be a cause for insidious bugs later on. And this stuff isn't immediately obvious to anyone working on an image builder unless they end up reading this discussion or we very clearly document this somewhere where people will end up reading it.

Previously, yeah we'd ship stuff in /efi under the wrong machine ID directory, but it wasn't noticeable because kernel-install would just create a new directory and put new stuff under that. And since we removed the machine-id from the filesystem, a new unique one would be generated for each instance on boot. With this patch, we'll actually end up shipping the same machine ID for every instance of the image.

poettering: Quite frankly, given this was broken in any case (i.e. the orphaned entry dir we so far generate just sucks! and the fact we can't match up what the boot loader reports with what we look for on next book is a bad bug), I am not sure it would really be a regression. We just move broken behaviour from one place to another, which I think is not a total desaster, given the ultimate outcome is an improvement.

Note that for the specific case of mkosi the issue shouldn't matter, since there we generate type 2 entries anyway.

That said, if this is important to you I figure we could tweak this further:

1. a simple hack would be: don't write the machine ID into the generated entry file if we run in a container. that should work, but is slightly skewed I guess. Thunking behind it: we are apparently prepping an image in a container environment we intend to use in a baremetal or VM environment, otherwise we wouldn't bother with kernel-install. But if that's the case we can probably assume it's going to be a golden image situation.

2. another option is to only persist the machine ID in the kernel cmdline if `test -f /etc/kernel/entry-token` also holds, i.e. if the machine ID was *explicitly* configured to be used. This would thus mean the behaviour must be explicitly requested (i.e. the file must be created, and given the file is a new concept people won't get opted into this automatically just like that).

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+      <varlistentry>
+        <term><option>--entry-token=</option></term>
+
+        <listitem><para>Controls how to name and identify boot loader entries for this OS
+        installation. Accepted during <option>install</option>, and takes one of <literal>auto</literal>,
+        <literal>machine-id</literal>, <literal>os-id</literal>, <literal>os-image-id</literal> or an
+        arbitrary string prefixed by <literal>explicit:</literal> as argument.</para>
+
+        <para>If set to <option>machine-id</option> the entries are named after the machine ID of the running
+        systemd (e.g. <literal>b0e793a9baf14b5fa13ecbe84ff637ac</literal>). See
```
daandemeyer: => the running system?

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+
+        <para>If set to <option>auto</option> (the default) it is first attempted to read the
+        <filename>/etc/kernel/entry-token</filename> file if it exists, and use the stored entry token
+        string. Otherwise if the local machine ID is initialized it is used. If it is not initialized the
+        <varname>IMAGE_ID=</varname> or <varname>ID=</varname> fields of <filename>os-release</filename> are
+        used, in this order of precedence, if set.</para>
+
+        <para>Unless set to <literal>machine-id</literal>, or when
+        <option>--make-entry-directory=yes</option> is used the selected token string is written
+        to a file <filename>/etc/kernel/entry-token</filename>, to ensure it will be used for future
```
daandemeyer: We had a report (https://github.com/systemd/systemd/issues/22224) complaining about /etc/kernel/entry-token being written even if kernel-install wasn't installed or used. So I figure we also shouldn't write /etc/kernel/entry-token when --make-entry-directory is set to os-id or os-image-id since kernel-install can find those directories on its own.

man/kernel-install.xml
```diff
@@ -163,36 +175,55 @@
...
-    <para><varname>KERNEL_INSTALL_STAGING_AREA=...</varname> is set for plugins to a path to a directory.
-    Plugins may drop files in that directory, and they will be installed as part of the loader entry, based
-    on the file name and extension.</para>
+    <para>If <varname>$MACHINE_ID</varname> is set and not empty when <command>kernel-install</command> is
+    invoked, it will be used as <replaceable>MACHINE-ID</replaceable>, overriding any automatic detection
+    attempts.  The value must be a valid machine ID (32 hexadecimal characters).</para>
+
+    <para><varname>$KERNEL_INSTALL_MACHINE_ID</varname> is set for the plugins to the desired
+    <replaceable>MACHINE-ID</replaceable> to use. It's always a 128bit ID, and typically the ID from
+    <filename>/etc/machine-id</filename> or the one passed in via <varname>$MACHINE_ID</varname>.</para>
```
daandemeyer: I'd add a note here about MACHINE_ID being some random UUID if no machine ID is provided and /etc/machine-id is empty

src/boot/bootctl.c
```diff
@@ -1174,100 +1285,33 @@ static int install_loader_config(const char *esp_path) {
...
-                                  machine_id, install_layout);
-                        return 0;
-                }
-
-                r = read_full_stream(orig, &contents, &length);
-                if (r < 0)
-                        return log_error_errno(r, "Failed to read /etc/machine-info: %m");
-        }
+        if (!arg_make_entry_directory && arg_entry_token_type != ARG_ENTRY_TOKEN_MACHINE_ID)
+                return 0;
```
daandemeyer: So I figure we should change this to only write /etc/kernel/entry-token if an explicit entry-token was provided. We can find machine-id, os-id and image-os-id in kernel-install automatically without an additional file.

DaanDeMeyer (changes requested):
Looks good but there's a few things that aren't quite clear to me yet.

man/bootctl.xml
poettering: You mean /etc/machine-info I presume? /etc/kernel/entry-token is a new thing this PR introduces after all.

It appears to me #22224 is not really something we support right now. We should, but that would involve adding a --root= switch, so that we operate on some other root fs.

Note that --make-entry-directory= defaults to no, so we don't actually create the dir at all right now, and hence also don't write the token out.

daandemeyer: Yeah that issue was complaining about /etc/machine-info being written, but I'd assume they'd have the same issue with /etc/kernel/entry-token being written when we land this PR.

src/kernel-install/kernel-install
daandemeyer: I think if we add a clear note in the relevant documentation warning image builders that they need to take explicit action here to avoid the machine ID getting coded into the spec files should be sufficient. Users might still end up running into this but they should be able to figure things out if the problem is documented in the relevant documentation.

DaanDeMeyer:
cc @AdamWill this will touch kernel-install semantics again so it'd be great if you could take a look and check that this doesn't cause issues for the Anaconda live install use case.

src/boot/bootctl.c
poettering: Hmm, so the line you highlighted is actually borked. It should read:

```c
if (!arg_make_entry_directory && arg_entry_token_type == ARG_ENTRY_TOKEN_MACHINE_ID)
        return 0;
```

i.e the idea here was that we'll persist the it whenever we are told to create the dir or when it is set to any non-default value.

Why persist in even if the selected id is "os-id" or "image-os-id" instead of purely relying on auto-detection in kernel-install for those cases? Because unlike for the machine ID even if everything is set up OK there's a risk that two distinct images will use the same ID and fight for the directory (think: two images generated from Fedora might use the "fedora" entry token. we should never do that by accident, but only on request). Hence I figured we should not risk that: if a risky way to identify the entries is selected then we should remember that and cut any auto-detection short, and just do what we were told.

does that thinking make any sense?

(I mean, it would be Ok if there was exactly one other string it might be OK to rely on a logic of "machine-id" dir doesn't exist, hence use this one other string, but IRL there are at least three potential strings — os-id, os-image-id, "Default" – and possible more in future hence relying on searching here is icky I think)

man/bootctl.xml
poettering: Anyway, I think this issue doesn't apply here, since we'll not create the dir by default anyway, it's opt-in.

src/kernel-install/kernel-install
poettering: I'll add a NEWS file for that.

src/boot/bootctl.c
daandemeyer: Shouldn't it be as follows then:

```
if (!arg_make_entry_directory || arg_entry_token_type == ARG_ENTRY_TOKEN_MACHINE_ID)
        return 0;
```
> Why persist in even if the selected id is "os-id" or "image-os-id" instead of purely relying on auto-detection in kernel-install for those cases? Because unlike for the machine ID even if everything is set up OK there's a risk that two distinct images will use the same ID and fight for the directory (think: two images generated from Fedora might use the "fedora" entry token. we should never do that by accident, but only on request). Hence I figured we should not risk that: if a risky way to identify the entries is selected then we should remember that and cut any auto-detection short, and just do what we were told.

Should we do searching in kernel-install at all then if we're always going to write /etc/kernel/entry-token? Why not remove the ickyness altogether and require /etc/kernel/entry-token to be written when using anything that's not the machine ID? If we're going to write to entry-token from bootctl anyway, we might as well require it and remove the heuristics from kernel-install.

poettering:
Force pushed new version, with most of @DaanDeMeyer's points addressed. For the other ones see above.

AdamWill:
Thanks for the heads-up. As best as I can figure it out, the Fedora live image case will be similar to Daan's but, if anything, worse. I'm finding it quite hard to reason about, though, since this has gotten so, uh, baroque.

AdamWill:
@jkonecny12 @poncovka @VladimirSlavik heads up about this. It's the follow-up from https://github.com/rhinstaller/anaconda/pull/3770 . We may wind up having to change stuff again.

@martinezjavier @nirik @nullr0ute heads up too, in case this affects container/disk images.

jkonecny12:
Do I understand it correctly that we can get back the `machine-info file` now or something else would be required?

poettering:
> Do I understand it correctly that we can get back the `machine-info file` now or something else would be required?

I don't understand the question, but this stops advertising KERNEL_INSTALL_MACHINE_ID= as a thing, we'll support it in code for compat, but people shouldn't use it. Use /etc/kernel/entry-token instead, it's more generic.

src/boot/bootctl.c
poettering: > Shouldn't it be as follows then:
>
> if (!arg_make_entry_directory || arg_entry_token_type == ARG_ENTRY_TOKEN_MACHINE_ID)
>        return 0;

No. Note that the condition is chekcing when to *not* create the dir. i.e. it's for when to exit early from the function. Or in other words, if you invert it, then you'll see under which conditions the dir is created:

```c
if (arg_make_entry_directory || arg_entry_token_type != ARG_ENTRY_TOKEN_MACHINE_ID) {
…
}
```

i.e. we persist the token when we are told to create dir, or when the token is not the default of machine-id.

poettering: > Should we do searching in kernel-install at all then if we're always going to write /etc/kernel/entry-token? Why not remove the ickyness altogether and require /etc/kernel/entry-token to be written when using anything that's not the machine ID? If we're going to write to entry-token from bootctl anyway, we might as well require it and remove the heuristics from kernel-install.

The main reason is compat: we used to search for "default", and we should probably continue to do so. I see "default" and ID=/IMAGE_ID= in the same boat there...

dunno, we could remove the searching altogether, i#d not be stronlgy opposed to that, but i think it does make some sense...

AdamWill:
@poettering I think the key question for anaconda is whether anything will now *set* `KERNEL_INSTALL_MACHINE_ID` on a fresh install. If not, that would mean we could stop worrying about wiping `/etc/machine-info` when doing a fresh install (which is what we had to do after the first change).

@jkonecny12 @poettering 's reply makes it sound like we might wind up having to wipe `/etc/kernel/entry-token` instead, maybe...

poettering:
No, the code that wrote the machine ID to /etc/machine-info is thoroughly gone, and has not be replaced.

The new file /etc/kernel/entry-token is only created on explicit user request, if they want to name their entries by something different than the machine ID, but I doubt anaconda builds images that way.

Hence, no need to wipe anything: our kernel-install/bootctl code will not put any machine ID or so anywhere in /etc/.

DaanDeMeyer:
> The new file /etc/kernel/entry-token is only created on explicit user request, if they want to name their entries by something different than the machine ID, but I doubt anaconda builds images that way.

I've cloned this locally for testing and currently, if I run `bootctl install --make-entry-directory --entry-token=machine-id`, /etc/kernel/entry-token gets written by bootctl. If I understand the PR correctly, that's not intended and /etc/kernel/entry-token should only be written when we use ID, IMAGE_ID or a custom string?

man/bootctl.xml
daandemeyer: I had to spend a good 15 minute thinking about this before I figured out the intent behind it. I'd reword this sentence as follows:

> If not set to machine-id or when --make-entry-directory=yes is used, the selected token string ...

DaanDeMeyer:
So the default behavior is quite bad at the moment. If I boot an image, clear /efi and run bootctl install followed by kernel-install, kernel-install won't do anything at all since it only installs stuff if $BOOT_ROOT/$ENTRY_TOKEN exists.

This isn't caused by this PR btw, it exists on the main branch as well but I figure we should fix it. Currently the behavior is that we only create $ENTRY_DIR_ABS (which is $BOOT_ROOT/$ENTRY_TOKEN/$KERNEL_VERSION) if $layout == "bls". The default value for $layout is "bls" if "$BOOT_ROOT/$MACHINE_ID" exists or "other" otherwise.

I figure better default behavior would be to unconditionally default $layout to "bls" unless it's set explicitly in install.conf. This makes sure that by default, kernel-install will actually install something. If an image builder like mkosi wants to do something different like a type 2 image, it should write something to install.conf to avoid kernel-install from creating "$BOOT_ROOT/$MACHINE_ID".

So I'd add one more commit to this PR to unconditionally set $layout to "bls" in kernel-install so that running bootctl install followed by kernel-install or running kernel-install on its own actually ends up installing something to $BOOT_ROOT. Currently kernel-install will run, dracut will run and actually generate an image as well but nothing ends up in $BOOT_ROOT and kernel-install will succeed and return zero. So users will think everything succeeded and fail to boot when they reboot their machine because there's no kernel image in /efi.

poettering:
> This isn't caused by this PR btw, it exists on the main branch as well but I figure we should fix it. Currently the behavior is that we only create $ENTRY_DIR_ABS (which is $BOOT_ROOT/$ENTRY_TOKEN/$KERNEL_VERSION) if $layout == "bls". The default value for $layout is "bls" if "$BOOT_ROOT/$MACHINE_ID" exists or "other" otherwise.
> 
> I figure better default behavior would be to unconditionally default $layout to "bls" unless it's set explicitly in install.conf. This makes sure that by default, kernel-install will actually install something. If an image builder like mkosi wants to do something different like a type 2 image, it should write something to install.conf to avoid kernel-install from creating "$BOOT_ROOT/$MACHINE_ID".

Maybe we should do "test -f /EFI/systemd/systemd-bootx64.efi" as a test for checking whether we should use the standard layout. That should be a robust way to determine if "bootctl" was previously run

poettering:
or maybe "test -d /efi/EFI/systemd" is even better since arch independent, and the dir should already suffice.

DaanDeMeyer:
> Maybe we should do "test -f /efi/EFI/systemd/systemd-bootx64.efi" as a test for checking whether we should use the standard layout. That should be a robust way to determine if "bootctl" was previously run

That would still result in a bad user experience when kernel-install is run before bootctl install though. I think we should strive to have systemd-boot and kernel-install be mostly independent from each other.

What about not changing the default for "$layout" to "bls" if unset but still creating $ENTRY_DIR_ABS if "$layout" is unset in install.conf? That might be backwards compatible for Fedora but will still fix the use case where kernel-install is executed with empty /efi and empty install.conf.

poettering:
> That would still result in a bad user experience when kernel-install is run before bootctl install though. 

Well, we that depends on the distro. On Fedora (where Grub and it's frankenbootloaderspec are the default) it's probably a worse user experience if we generate entries that simply cannot be understood.

> I think we should strive to have systemd-boot and kernel-install be mostly independent from each other. Especially since kernel-install usually runs in a package magager post install script and bootctl doesn't. So regularly you'll end up with situations where kernel-install will run before bootctl install.
> 
> What about not changing the default for "$layout" to "bls" if unset but still creating $ENTRY_DIR_ABS if "$layout" is unset in install.conf? That might be backwards compatible for Fedora but will still fix the use case where kernel-install is executed with empty /efi and empty install.conf.

$ENTRY_DIR_ABS only really makes sense if the real boot loader spec is used.

Quite frankly, I wasn't the one adding the layout setting, and I have my doubts about it (see #22379). So I really don#t grok what the intention is with it. i.e. was it to be something to make proper boot loader spec support and opt-in thing, or an opt-out thing? I mean, that#s what it boils down to here.

From my perspective I'd always say it should be an opt-out thing. i.e. we should default to our version of the boot loader spec, and use that by default. The Fedora/grub thing with its variable expension misses the point bigtime. So I actually agree with you: we should default to the layout setting called "bls" if we have no other idea. It would then be up to fedora to figure out how to switch the setting to something else there. One option would be that fedora ships an /etc/kernel/install.conf file that says so, by default, either in the systemd.rpm package (initially at least) or in grub.rpm (eventually). But I really don't understand what the goal here was. @keszybz?

poettering:
I think we shouldn't mix up the discussions about machine-id and about default layout too much. i.e. get this PR merged first, and then look into figuring out how the layout setting can be handled properly.

poettering:
I wrote together my thoughts on what to do about the layout detection and configuration here:

https://github.com/systemd/systemd/issues/22379#issuecomment-1044541442

It provides an avenue to making the fedora case just work, but still default to upstream boot loader type `#1` spec stuff by default.

poettering:
Anyway, is there anything still stopping this PR from being merged?

@keszybz do you have an opinion on this?

DaanDeMeyer (approved):
Tested this out and it all works as expected. The layout thing can be fixed separately.

src/kernel-install/kernel-install
keszybz: Option 1. doesn't really help I think. At least Anaconda is not running anything in a container…
Option 2. requires an explicit step, but that's OK.

src/kernel-install/kernel-install
keszybz: I think this is still valid.

src/kernel-install/kernel-install
```diff
@@ -237,6 +242,14 @@ case "$COMMAND" in
         fi
         ;;
 
+    inspect)
+        echo "KERNEL_INSTALL_MACHINE_ID: $KERNEL_INSTALL_MACHINE_ID"
+        echo "KERNEL_INSTALL_ENTRY_TOKEN: $KERNEL_INSTALL_ENTRY_TOKEN"
+        echo "KERNEL_INSTALL_BOOT_ROOT: $KERNEL_INSTALL_BOOT_ROOT"
+        echo "KERNEL_INSTALL_LAYOUT: $KERNEL_INSTALL_LAYOUT"
+        echo "KERNEL_INSTALL_INITRD_GENERATOR: $KERNEL_INSTALL_INITRD_GENERATOR"
+        echo "ENTRY_DIR_ABS: $KERNEL_INSTALL_BOOT_ROOT/$ENTRY_TOKEN/\$KERNEL_VERSION"
```
keszybz, line 251: Maybe add
```sh
[ "${ENTRY_DIR_ABS%/*}" == "$KERNEL_INSTALL_BOOT_ROOT/$ENTRY_TOKEN" ] || { echo "Wat!?"; exit 1 }
```
as an assert…

man/kernel-install.xml
```diff
@@ -163,36 +175,57 @@
...
-    on the file name and extension.</para>
+    <para>If <varname>$MACHINE_ID</varname> is set and not empty when <command>kernel-install</command> is
+    invoked, it will be used as <replaceable>MACHINE-ID</replaceable>, overriding any automatic detection
+    attempts.  The value must be a valid machine ID (32 hexadecimal characters).</para>
+
+    <para><varname>$KERNEL_INSTALL_MACHINE_ID</varname> is set for the plugins to the desired
+    <replaceable>MACHINE-ID</replaceable> to use. It's always a 128bit ID, and typically the ID from
+    <filename>/etc/machine-id</filename> or the one passed in via <varname>$MACHINE_ID</varname>. (If no
+    machine ID was specified via these methods it might be generated randomly by
+    <command>kernel-installed</command>, in which case it only applies to this invocation.)</para>
```
keszybz: kernel-installed → kernel-install

man/kernel-install.xml
```diff
@@ -163,36 +175,57 @@
...
+
+    <para><varname>$KERNEL_INSTALL_ENTRY_TOKEN</varname> is set for the plugins to the desired entry "token"
+    to use. It's an identifier that shall be used to identify the local installation, and is often the
+    machine ID, i.e. equivalent to <varname>$KERNEL_INSTALL_MACHINE_ID</varname>, but might also be a
```
keszybz: equivalent to → the same as

man/kernel-install.xml
```diff
@@ -163,36 +175,57 @@
...
+    different type of identifier, for example a fixed string or the <varname>ID=</varname>,
+    <varname>IMAGE_ID=</varname> values from <filename>/etc/os-release</filename>. The string passed here
+    should be used to name Boot Loader Specification entries, or the directories the kernel image and initial
```
keszybz: should → will ?

man/kernel-install.xml
```diff
@@ -163,36 +175,57 @@
...
+    RAM disk images are placed into. Note that while oftentimes
+    <varname>$KERNEL_INSTALL_ENTRY_TOKEN</varname> and <varname>$KERNEL_INSTALL_MACHINE_ID</varname> are set
+    to the same values the latter is guaranteed to be in 128bit hexadecimal character form while the former
```
keszybz: Comma missing: "values, the latter"
"is guaranteed to be a valid 32-character ID in lowercase hexadecimals"

man/kernel-install.xml
```diff
@@ -163,36 +175,57 @@
...
+    can be any short string. The entry token to use is read from
+    <filename>/etc/kernel/entry-token</filename>, if it exists. Otherwise a few possible candidates below the
+    <varname>$BOOT</varname> are searched for Boot Loader Specification Type 1 entry directories, and if
+    found the entry token is derived from that. If that is not successful the machine ID is used as
+    fallback.</para>
+
+    <para><varname>$KERNEL_INSTALL_BOOT_ROOT</varname> is set for the plugins to the absolute path of the
+    root directory (mount point, usually) of the hierarchy where boot loader entries, kernel images, and
+    associated resources should be placed. This usually is the path the XBOOTLDR or UEFI ESP are mounted, and
```
keszybz: the path the XBOOTLDR or UEFI ESP → the path where the XBOOTLDR partition or the ESP (EFI System Partition)

man/kernel-install.xml
```diff
@@ -249,35 +291,31 @@
...
-          <filename>/etc/machine-id</filename>
+          <filename>/etc/kernel/entry-token</filename>
         </term>
           <listitem>
-            <para>The content of this file specifies the machine identification
-            <replaceable>MACHINE-ID</replaceable>. If <filename>/etc/machine-id</filename>
-            cannot be read or is temporary (backed by a file on <constant>tmpfs</constant>),
-            <command>kernel-install</command> will use <literal>Default</literal> instead.</para>
+            <para>If exists it is read and used as "entry token" for this system, i.e. is used for naming
+            Boot Loader Specification entries after, see <varname>$KERNEL_INSTALL_ENTRY_TOKEN</varname> above
```
keszybz: " after" → ""

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+        <literal>no</literal>.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><option>--entry-token=</option></term>
+
+        <listitem><para>Controls how to name and identify boot loader entries for this OS
+        installation. Accepted during <option>install</option>, and takes one of <literal>auto</literal>,
+        <literal>machine-id</literal>, <literal>os-id</literal>, <literal>os-image-id</literal> or an
+        arbitrary string prefixed by <literal>explicit:</literal> as argument.</para>
```
keszybz: Do we need "explicit" in the name? I think "literal" is a better word for this, but we don't need to be so verbose. I think just ":" as the magic prefix would be enough. (Or "=" or "+" ?)

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+        <para>If set to <option>os-id</option> the entries are named after the OS ID of the running system,
+        i.e. the <varname>ID=</varname> field of
+        <citerefentry><refentrytitle>os-release</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+        (e.g. <literal>fedora</literal>). Similar, if set to <option>os-image-id</option> the entries are
+        named after the OS image ID of the running system, i.e. the <varname>IMAGE_ID=</varname> field of
+        <filename>os-release</filename> (e.g. <literal>vendorx-cashier-system</literal>).</para>
+
+        <para>If set to <option>auto</option> (the default) it is first attempted to read the
+        <filename>/etc/kernel/entry-token</filename> file if it exists, and use the stored entry token
+        string. Otherwise if the local machine ID is initialized it is used. If it is not initialized the
```
keszybz:
```
<para>If set to <option>auto</option> (the default), the <filename>/etc/kernel/entry-token</filename> file will be read if it exists, and the store value used. 
Otherwise <varname>IMAGE_ID=</varname> from <filename>os-release</filename> will used, if set. 
Otherwise, <varname>ID=</varname> from <filename>os-release</filename> will used, if set.
If none of those mechanisms succeed, …
```

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+        <varname>IMAGE_ID=</varname> or <varname>ID=</varname> fields of <filename>os-release</filename> are
+        used, in this order of precedence, if set.</para>
+
+        <para>Unless set to <literal>machine-id</literal>, or when
+        <option>--make-entry-directory=yes</option> is used the selected token string is written
+        to a file <filename>/etc/kernel/entry-token</filename>, to ensure it will be used for future
+        entries. This file is also read by
+        <citerefentry><refentrytitle>kernel-install</refentrytitle><manvolnum>8</manvolnum></citerefentry>,
+        in order to identify under which name to generate boot loader entries for newly installed
+        kernels.</para>
```
keszybz: … what about kernel removal? The new $ENTRY_TOKEN is also used to find directories to delete.

man/bootctl.xml
```diff
@@ -258,23 +258,60 @@
...
+
+        <para>Using the machine ID for naming the entries is generally preferable, however there are cases
+        where using the other identifiers is a good option. Specifically: if the identification data that the
+        machine ID entails shall not be stored on the (unencrypted) <varname>$BOOT</varname> partition, or if
+        the ID shall be generated on first boot and is not known when the entries are prepared. Note that
+        using the machine ID has the benefit that multiple parallel installations of the same OS can coexist
+        on the same medium, and they can update their boot loader entries independently. By using another
+        identifier (such as and OS ID or OS image ID) parallel installations of the same OS will result in
+        conflicts about the entry names — however parallel installations of different OSes or images
+        (i.e. that use different IDs) may be supported.</para></listitem>
```
keszybz: "By using …" →
"When using another identifier (such as the OS ID or the OS image ID), parallel installations of the same OS would try to use the same entry name. To support parallel installations, the installer must use a different entry token when adding a second installation."

(I think it's better to just say what needs to be done than to beat around the bush.)

keszybz (commented):
Thanks, I think this will work and cover all the cases we need.

keszybz:
> Anyway, is there anything still stopping this PR from being merged?

Sorry for the delay. I think it's good to go with some polish.

man/bootctl.xml
poettering: it's more about keeping the door open here, so that you can do file:/foo/bar/baz.txt later on if you like. i mean, there are so many ways to pick an id here... could also be some key derived from the tpm, so that oyu bind the entry to the tpm. and in that case you might want to write tpm2:/dev/tpmrm0 or so... you never know what might make sense to add here one day...

poettering: will rename to "literal"

man/bootctl.xml
poettering: hmm, this is actually wrong, we prefer the machine ID over the os-lease ids... 

i'll try to correct that and merge both versions

src/kernel-install/kernel-install
poettering: hmm, i am not a shell guru, but what is {} supposed to do here? if i copy/paste your line the shell refuses to process the script.

poettering: do you mean ()? if so it wouldn't work either, since the exit would exit the subshell?

anyway, i now added this, but with if/then/fi.

behrmann: The braces group this into a block (but not in a subshell). I think the problem is the missing `;` after `exit 1`.  Also the `==` inside `[ ]` is not valid POSIX shell, but only allowed inside of bash braces `[[ ]]` (or inside `[ ]` in bash's flavour of POSIX shell), so 
```sh
[ "${ENTRY_DIR_ABS%/*}" = "$KERNEL_INSTALL_BOOT_ROOT/$ENTRY_TOKEN" ] || { echo "Wat!?"; exit 1; }
```
should have worked.

src/kernel-install/kernel-install
poettering: oops, seems i fucked up my rebase. it was fixed in a later commit of this series, will move the fix.

poettering:
Force pushed new version. Addresses everything pointed out, except for https://github.com/systemd/systemd/pull/22463#discussion_r813047200 

I am not sure about that I must say: so far we included the machine-id also in the "machine-id" stanza of generated entries. Thus we placed it at two places already: in the path name, and *in* the files. We just add it at one additional place in the file: in the kernel cmdline, i.e. it's not that big a step.

Also, I kinda *want* that this logic is used for updated system, because it fixes the initrd-has-a-different-machine-id issue we so far had, which is useful in *any* case, i.e. regular Fedora.

Thus I think we should advertise this change in NEWS, and make the change to propagate the machine ID fully in all cases by default. It's of benefit for the majority of users I think, and the golden image users are the more technical I figure, and should be able to read NEWS. Given that the status quo ante was all broken I think the overall effect is good for everyone summar summarum.

Does that make sense?

src/kernel-install/kernel-install
poettering: indeed. will change it to that.

poettering:
changed the assert line, as suggested by @behrmann

DaanDeMeyer:
> Thus I think we should advertise this change in NEWS, and make the change to propagate the machine ID fully in all cases by default. It's of benefit for the majority of users I think, and the golden image users are the more technical I figure, and should be able to read NEWS. Given that the status quo ante was all broken I think the overall effect is good for everyone summar summarum.

I'm not 100% sure about this though. What this does is it enforces that whatever is the machine ID at the time of running kernel-install becomes the machine ID of the booted image. If I understand correctly, any kind of image builder building an image that'll be distributed to > 1 machine will never want to rely on this, since they might end up encoding a machine ID into the final image that will then be shared by all machines that the image is installed on.

But when building an image that'll intended for just 1 machine, there's no need to remove the generated machine ID from the image as you can simply leave it there and it'll be the machine ID of the generated image. In that scenario, there's also no need to put the machine ID in the boot loader entry since it's already there in /etc/machine-id. So what use case is then left for this to address?

poettering:
> I'm not 100% sure about this though. What this does is it enforces that whatever is the machine ID at the time of running kernel-install becomes the machine ID of the booted image. If I understand correctly, any kind of image builder building an image that'll be distributed to > 1 machine will never want to rely on this, since they might end up encoding a machine ID into the final image that will then be shared by all machines that the image is installed on.

But they encoded it into the images previously, too. i.e. the "machine-id" field of the generated snippets contain the machine ID used to generate them, and the path where the snippets are placed are also named after the machine ID. 

hence the machine ID is already there, and it shouldn't be. I totally agree with that, and this PR adds infra for that.

> But when building an image that'll intended for just 1 machine (or when running kernel-install on an already running machine like a user laptop), there's no need to remove the generated machine ID from the image as you can simply leave it there and it'll be the machine ID of the generated image. In that scenario, there's also no need to put the machine ID in the boot loader entry since it's already there in /etc/machine-id. So what use case is then left for this to address?

I am thinking of a classic distro install, installed with a classic distro installer. There the initrd will run with a randomized/transient machine ID. Once the system transitioned into the host fs tree it will thus change to a different machine ID. It's really nice to address that, and ensure tht the machine ID stays the same for the whole runtime of the OS, i.e. including in the initrd. And that's something systemd.machine_id= on the kernel cmdline will achieve.

DaanDeMeyer:
> I am thinking of a classic distro install, installed with a classic distro installer. There the initrd will run with a randomized/transient machine ID. Once the system transitioned into the host fs tree it will thus change to a different machine ID. It's really nice to address that, and ensure tht the machine ID stays the same for the whole runtime of the OS, i.e. including in the initrd. And that's something systemd.machine_id= on the kernel cmdline will achieve.

That's something I hadn't thought about. I'm not sure if this is a problem with today's model of building the initrd from the host system but it'd be very relevant if the initrd becomes just another package to be installed since we can't encode the machine ID in the initrd in that case. In that case this looks fine. I'm going to set the green label since this went through a lot of scrutiny already so it should be good to go.

src/kernel-install/kernel-install
```diff
@@ -205,6 +242,17 @@ case "$COMMAND" in
...
+
+        # Assert that ENTRY_DIR_ABS actually matches what we are printing here
+        [ "${ENTRY_DIR_ABS%/*}" = "$KERNEL_INSTALL_BOOT_ROOT/$ENTRY_TOKEN" ] || { echo "Wat!?"; exit 1; }
```
behrmann: Although I like this because it's quite whimsical, I'm not sure @keszybz meant the error message quite so literally (just guessing), maybe this error could be a bit more descriptive, because looking from a user perspective "`kernel-install inspect` just printed 'Wat!?' on me. Wat!?" might not be the best UX. :)

To be in line with the rest, the error message should also go to stderr.
```suggestion
        [ "${ENTRY_DIR_ABS%/*}" = "$KERNEL_INSTALL_BOOT_ROOT/$ENTRY_TOKEN" ] || { echo "Wat!?" >&2; exit 1; }
```

poettering: It's an assertion, i.e. it's not really user facing, as it should never happen IRL. But yeah you are right, will fix the message to "Assertion failed."

poettering:
> EDIT: Hmm, maybe one more question: Is there a way to get the same benefit for machines booted from a golden image? In those scenarios, how would we get the machine ID into the initrd? We can't directly modify the kernel cmdline in that scenario. Would something like an EFI variable make more sense there? If so, should we try to use that everywhere? I vaguely remember @bluca suggesting something like this a while ago

I'd say this is something to discuss separately from this issue. So far the idea of the machine ID was always to be specific to an installation and you could have multiple of those on the same system. With an EFI var this is hard to do, since they are per system.

I think what could work if sooner or later we use the credentials stuff for this, i.e. that we allow configuring it via the *.creds drop-ins that sd-stub understands these days. Benefit: if we use that we get authentication for free, with TPM and whatnot.

bluca:
@mrc0mmand is there some issue with the testbed?

```
12:47:57 ERROR: Step ‘Archive the artifacts’ failed: no workspace for upstream-vagrant-archlinux #10498
```

mrc0mmand:
> @mrc0mmand is there some issue with the testbed?
> 
> ```
> 12:47:57 ERROR: Step ‘Archive the artifacts’ failed: no workspace for upstream-vagrant-archlinux #10498
> ```

Ah, interesting, yesterday I noticed an openshift pod dying unexpectedly but I couldn't tie it to a a specific job. Looks like you just found it :-)

I restarted the job, should hopefully finish successfully this time.

PR #22465, opened 2022-02-09
Clarify protocol used in systemd-journal-upload

FedericoCeratto:
Point out that systemd-journal-upload is not compatible with syslog.

### DIFF
--- a/man/systemd-journal-upload.service.xml
+++ b/man/systemd-journal-upload.service.xml
@@ -47,7 +47,7 @@
     as they become available.</para>
     
     <para><command>systemd-journal-upload</command> transfers the raw content of journal file and
-    uses HTTP as a transport protocol. It is not compatible with syslog.</para> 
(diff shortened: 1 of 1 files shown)


man/systemd-journal-upload.service.xml
```diff
@@ -45,7 +45,10 @@
     Unless limited by one of the options specified below, all journal entries accessible to the user
     the program is running as will be uploaded, and then the program will wait and send new entries
     as they become available.</para>
-
+    
+    <para><command>systemd-journal-upload</command> transfers the raw content of journal file and
+    uses HTTP as a transport protocol. It is not compatible with syslog.</para>
```
bluca: no need to specify incompatibilities, there's no claim of compatibility anywhere

FedericoCeratto: Updated.

bluca (approved):

PR #22470, opened 2022-02-10
basic: do not warn in mkdir_p() when parent directory exists

keszybz:
This effectively disables warnings about mode+ownership of existing directories
when recursively creating parent directories.

I was recently debugging an issue with the fstab-generator [1], and it says:
'Directory "/tmp" already exists, but has mode 0777 that is too permissive (0644 was requested), refusing.'
which is very specific but totally wrong in this context.
This output was added in 37c1d5e97dbc869edd8fc178427714e2d9428d2b, and I still
think it is worth to do it, because if you actually *do* want the directory, if
there's something wrong, the precise error message will make it much easier to
diagnose. And we can't easily pass the information what failed up the call chain
because there are multiple things we check (ownership, permission mask, type)…
So passing a param whether to warn or not down into the library code seems like
the best solution, despite not being very elegant.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2051285

### DIFF
--- a/src/basic/mkdir.c
+++ b/src/basic/mkdir.c
@@ -51,6 +51,9 @@ int mkdir_safe_internal(
                                                    _mkdirat);
         }
 
+        if (flags & MKDIR_IGNORE_EXISTING)
+                return 0;
+
         if (!S_ISDIR(st.st_mode))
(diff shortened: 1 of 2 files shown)


yuwata (approved):
LGTM.

poettering:
I am pretty sure we should always validate the type though, i.e. move the if check you added three lines further down.

poettering:
maybe rename this to MKDIR_PERMISSIVE instead given it's about both mode and ownership?

PR #3323, opened 2016-05-22
resolved: don't stop handle messages after receiving a zero length UDP packet

evverx:
Fixes:

```
-bash-4.3# ss --udp -l -p
State      Recv-Q Send-Q Local Address:Port                 Peer Address:Port
UNCONN     0      0          *:5355                     *:* users:(("systemd-resolve",pid=601,fd=12))
UNCONN     0      0         :::5355                    :::* users:(("systemd-resolve",pid=601,fd=14))

-bash-4.3# nping --udp -p 5355 --data-length 0 -c 1 localhost

-bash-4.3# journalctl -u systemd-resolved -b --no-hostname
...
May 21 14:59:22 systemd-resolved[601]: Event source llmnr-ipv4-udp (type io) returned error, disabling: Input/output error
...

-bash-4.3# nping --udp -p 5355 --data-length 1000 -c 1 localhost

-bash-4.3# ss --udp -l
State      Recv-Q Send-Q Local Address:Port                 Peer Address:Port
UNCONN     2304   0          *:5355                     *:*
UNCONN     0      0         :::5355                    :::*
```

PR #22479, opened 2022-02-10
resolve: describe mdns event sources

evverx:
when they go down resolved prints
```
Event source mdns-ipv4 (type io) returned error, disabling
```
instead of
```
Event source n/a (type io) returned error, disabling
```


evverx:
It's useful to debug issues like https://github.com/systemd/systemd/pull/3323 :-)

yuwata (approved):

PR #22482, opened 2022-02-11
ci: run all fuzz targets on CIFuzz

evverx:
CIFuzz has been kind of broken for a couple months because
coverage reports downloaded from OSS-Fuzz contain absolute
paths while paths to files changed in PRs are relative and they
don't match. It makes it kind of hard for CIFuzz to figure out
what it should run so it runs either all fuzz targets or just new
fuzz targets. Until that issue is fixed let's just always predictably run
all fuzz targets.


evverx:

Issue #22428, opened 2022-02-05
Keyboard does not work in systemd-boot


nl6720:
**systemd version the issue has been seen with**
 > systemd 250 (250.3-2-arch)
+PAM +AUDIT -SELINUX -APPARMOR -IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK +XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified

**Used distribution**
 > Arch Linux

**Linux kernel version used** (`uname -a`)

 > `Linux archiso 5.16.4-arch1-1 #1 SMP PREEMPT Sat, 29 Jan 2022 19:08:13 +0000 x86_64 GNU/Linux`

**CPU architecture issue was seen on**
 > x86_64

**Expected behaviour you didn't see**
 > Keyboard input working in systemd-boot.

**Unexpected behaviour you saw**
 > Keyboard input did not work in systemd-boot. Pressing any key stops the boot timeout, but that's all that happens. No other interaction is possible. Nothing happens when arrow keys, `e`, or any other key is pressed.
> This was observed on HP ProBook 450 G2 with firmware version 01.54. Another Arch user also encountered the issue on a 2016 macbook air; see https://bbs.archlinux.org/viewtopic.php?id=273765 .

**Steps to reproduce the problem**
 > Boot `archlinux-2022.01.01-x86_64.iso` or `archlinux-2022.02.01-x86_64.iso` and try pressing keys when in systemd-boot.

**Additional program output to the terminal or log subsystem illustrating the issue**

Bisecting yields e628944b11a3465b77bcd704f9fb66b58757df01 as the offending commit.

medhefgo:
Can you give this [branch](https://github.com/medhefgo/systemd/tree/boot-input) a try?

nl6720:
> Can you give this [branch](https://github.com/medhefgo/systemd/tree/boot-input) a try?

Tried it. No change with 87528e11281d41484fd28fe1b2dbbe2f8bbfadf3.

medhefgo:
[...]

PR #22484, opened 2022-02-11
boot: Correctly check the return value of CheckEvent

medhefgo:
Fixes: #22428


medhefgo:
Somebody should check if this check for CheckEvent checks out.

*I'll see myself out now*

poettering:
Ouch!

I now checked that the check for CheckEvent is checking out.

bluca (approved):

PR #22489, opened 2022-02-12
do not call __register_atfork directly

matijaskala:
this way it is cleaner and more portable and systemd links against libpthread anyway

### DIFF
--- a/src/basic/process-util.c
+++ b/src/basic/process-util.c
@@ -4,6 +4,7 @@
 #include <errno.h>
 #include <limits.h>
 #include <linux/oom.h>
+#include <pthread.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -1161,12 +1162,6 @@ void reset_cached_pid(void) {
         cached_pid = CACHED_PID_UNSET;
 }
 
-/* We use glibc __register_atfork() + __dso_handle directly here, as they are not included in the glibc
- * headers. __register_atfork() is mostly equivalent to pthread_atfork(), but doesn't require us to link against
- * libpthread, as it is part of glibc anyway. */
-extern int __register_atfork(void (*prepare) (void), void (*parent) (void), void (*child) (void), void *dso_handle);
-extern void* __dso_handle _weak_;
-
 pid_t getpid_cached(void) {
         static bool installed = false;
         pid_t current_value;
@@ -1194,7 +1189,7 @@ pid_t getpid_cached(void) {
                          * only half-documented (glibc doesn't document it but LSB does — though only superficially)
                          * we'll check for errors only in the most generic fashion possible. */
 
-                        if (__register_atfork(NULL, NULL, reset_cached_pid, __dso_handle) != 0) {
+                        if (pthread_atfork(NULL, NULL, reset_cached_pid) != 0) {
                                 /* OOM? Let's try again later */
                                 cached_pid = CACHED_PID_UNSET;
                                 return new_pid;


bluca:
Why is this necessary? What problem does it solve?

matijaskala:
__register_atfork is an internal glibc function and is not meant to be called directly

poettering:
See the original commit for the rationale for using it instead of pthread_atfork

poettering:
Does this patch solve any actual problem? Or is it a theoretic exercise?

matijaskala:
> See the original commit for the rationale for using it instead of pthread_atfork

The original commit doesn't contain any rationale. The comment does.
Like I said, the rationale has become invalid long ago, since systemd links against libpthread anyway.

matijaskala:
> Does this patch solve any actual problem? Or is it a theoretic exercise?

It solves compilation on musl.

bluca:
It is true that both libsystemd.so and libsystemd-shared.so link directly with libpthread:

```
$ lddtree build/src/shared/libsystemd-shared-250.so
libsystemd-shared-250.so => build/src/shared/libsystemd-shared-250.so (interpreter => none)
    libacl.so.1 => /lib/x86_64-linux-gnu/libacl.so.1
    libblkid.so.1 => /lib/x86_64-linux-gnu/libblkid.so.1
    libcap.so.2 => /lib/x86_64-linux-gnu/libcap.so.2
    libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1
    libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2
    libgcrypt.so.20 => /lib/x86_64-linux-gnu/libgcrypt.so.20
        libgpg-error.so.0 => /lib/x86_64-linux-gnu/libgpg-error.so.0
    libip4tc.so.2 => /lib/x86_64-linux-gnu/libip4tc.so.2
    libkmod.so.2 => /lib/x86_64-linux-gnu/libkmod.so.2
    liblz4.so.1 => /lib/x86_64-linux-gnu/liblz4.so.1
    libmount.so.1 => /lib/x86_64-linux-gnu/libmount.so.1
    libcrypto.so.1.1 => /lib/x86_64-linux-gnu/libcrypto.so.1.1
    libp11-kit.so.0 => /lib/x86_64-linux-gnu/libp11-kit.so.0
        libffi.so.8 => /lib/x86_64-linux-gnu/libffi.so.8
    libpam.so.0 => /lib/x86_64-linux-gnu/libpam.so.0
        libaudit.so.1 => /lib/x86_64-linux-gnu/libaudit.so.1
            libcap-ng.so.0 => /lib/x86_64-linux-gnu/libcap-ng.so.0
    librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1
    libseccomp.so.2 => /lib/x86_64-linux-gnu/libseccomp.so.2
    libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1
        libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0
    libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1
    liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5
    libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
    ld-linux-x86-64.so.2 => /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
$ lddtree build/libsystemd.so.0.33.0
libsystemd.so.0.33.0 => build/libsystemd.so.0.33.0 (interpreter => none)
    librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1
    liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5
    libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1
    liblz4.so.1 => /lib/x86_64-linux-gnu/liblz4.so.1
    libcap.so.2 => /lib/x86_64-linux-gnu/libcap.so.2
    libgcrypt.so.20 => /lib/x86_64-linux-gnu/libgcrypt.so.20
        libgpg-error.so.0 => /lib/x86_64-linux-gnu/libgpg-error.so.0
    libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0
    libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
    ld-linux-x86-64.so.2 => /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
```

@keszybz is this intentional? Is there any config combination where this doesn't happen?

yuwata:
I think it is OK to merge, or we should drop libpthread from the shared libraries (though I do not know if it is possible).

yuwata (approved):

yuwata:
I set the green label, but leave to merge this for other reviewers, @poettering, @keszybz and @bluca.

poettering:
glibc 2.34 (aug 2021) removed libpthread as separate library altogether, it's now part of libc itself. hence the whole issue went away. Thus I guess we should just use the pthread APIs directly for everything, as if they were never anything separate.

(libdl is the same actually, maybe we should rework our meson build logic to "backport" the new glibc setup, i.e. always imply -ldl -lpthread on all our builds without this requiring any explicit configured deps, @keszybz?)

PR #22487, opened 2022-02-11
bootctl: show all discovered entries, but show state + type in details

poettering:
This is supposed to be a replacement for #22161 #22398

This will ensure we comprehensively discover all entries, and then will highlight the ones that the boot loader reported to have seen that we can't see, and vice versa the ones we found that the boot laoder didn't report. This can either mean they have been uninstalled/installed since the last boot, or hint to some misconfiguration.

I am very sure it makes more sense to provide the user with the full info in low-level tool, and highlight relevant bits then to hide things away.

PR #22491, opened 2022-02-12
systemd-boot: add check to EFI / XBOOTLDR partition

kreijack:
This was discussed previously in #22041 #19600

The first 4 patches add the following check:
- ensure that only **one** EFI partition exists 
- ensure that only **one** XBOOTLDR partition exists
- (if used) ensure that the EFI partition has the same random seed as the one returned by systemd-boot (to avoid leak of information, the random seed is returned hashed).

The last patch add a further field in "bootctl list" called "status". It is available or unavaialble depending by the fact that the boot entry has a counter part in the filesystem.

### DIFF
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -508,6 +508,18 @@ static int boot_entry_show(const BootEntry *e, bool show_as_default) {
                         (void) terminal_urlify_path(e->path, NULL, &link);
 
                 printf("       source: %s\n", link ?: e->path);
+
+                /*
+                 * if the entry is a NON "auto-*" one and if the information comes
+                 * from the bootloader only, likely the entry was seen by
+                 * the systemd-boot but then it was removed
+                 */
+                if (!startswith(e->id, "auto-")) {
+                        if (startswith(e->path, "/sys/firmware"))
+                                printf("       status: %sunavailable%s\n", ansi_highlight_red(), ansi_normal());
+                        else
+                                printf("       status: %savailable%s\n", ansi_highlight_green(), ansi_normal());
+                }
         }
         if (e->version)
                 printf("      version: %s\n", e->version);


kreijack:
Removed an unused variables; moreover rebased to the latest systemd HEAD

yuwata:
The first commit looks good to me.

The second commit seems to be a duplicated work contained in #22487, and I prefer the way used in #22487 as it handles the situation gracefully rather than showing warning to the user.

Not sure about the commits about random seed things.

yuwata:
Also, compilers show error:
```
../../src/systemd/src/shared/bootspec.c:1242:19: error: use of undeclared identifier 'blkid_dev_iterate_endp'
        _cleanup_(blkid_dev_iterate_endp) blkid_dev_iterate iter = NULL;
                  ^
../../src/systemd/src/shared/bootspec.c:1242:43: error: unknown type name 'blkid_dev_iterate'
        _cleanup_(blkid_dev_iterate_endp) blkid_dev_iterate iter = NULL;
                                          ^
../../src/systemd/src/shared/bootspec.c:1243:9: error: use of undeclared identifier 'blkid_cache'
        blkid_cache cache = NULL;
        ^
../../src/systemd/src/shared/bootspec.c:1244:9: error: use of undeclared identifier 'blkid_dev'
        blkid_dev dev;
        ^
../../src/systemd/src/shared/bootspec.c:1248:13: error: implicit declaration of function 'blkid_get_cache' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        r = blkid_get_cache(&cache, NULL);
            ^
../../src/systemd/src/shared/bootspec.c:1248:30: error: use of undeclared identifier 'cache'
        r = blkid_get_cache(&cache, NULL);
                             ^
../../src/systemd/src/shared/bootspec.c:1252:9: error: implicit declaration of function 'blkid_probe_all' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        blkid_probe_all(cache);
        ^
../../src/systemd/src/shared/bootspec.c:1252:25: error: use of undeclared identifier 'cache'
        blkid_probe_all(cache);
                        ^
../../src/systemd/src/shared/bootspec.c:1253:16: error: implicit declaration of function 'blkid_dev_iterate_begin' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        iter = blkid_dev_iterate_begin(cache);
               ^
../../src/systemd/src/shared/bootspec.c:1253:40: error: use of undeclared identifier 'cache'
        iter = blkid_dev_iterate_begin(cache);
                                       ^
../../src/systemd/src/shared/bootspec.c:1254:16: error: implicit declaration of function 'blkid_dev_next' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        while (blkid_dev_next(iter, &dev) == 0) {
               ^
../../src/systemd/src/shared/bootspec.c:1254:38: error: use of undeclared identifier 'dev'; did you mean 'div'?
        while (blkid_dev_next(iter, &dev) == 0) {
                                     ^~~
                                     div
/usr/include/stdlib.h:852:14: note: 'div' declared here
extern div_t div (int __numer, int __denom)
             ^
../../src/systemd/src/shared/bootspec.c:1255:27: error: use of undeclared identifier 'blkid_free_probep'
                _cleanup_(blkid_free_probep) blkid_probe b = NULL;
                          ^
../../src/systemd/src/shared/bootspec.c:1255:46: error: unknown type name 'blkid_probe'
                _cleanup_(blkid_free_probep) blkid_probe b = NULL;
                                             ^
../../src/systemd/src/shared/bootspec.c:1259:17: error: use of undeclared identifier 'dev'
                dev = blkid_verify(cache, dev);
                ^
../../src/systemd/src/shared/bootspec.c:1259:23: error: implicit declaration of function 'blkid_verify' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                dev = blkid_verify(cache, dev);
                      ^
../../src/systemd/src/shared/bootspec.c:1259:36: error: use of undeclared identifier 'cache'
                dev = blkid_verify(cache, dev);
                                   ^
../../src/systemd/src/shared/bootspec.c:1259:43: error: use of undeclared identifier 'dev'
                dev = blkid_verify(cache, dev);
                                          ^
../../src/systemd/src/shared/bootspec.c:1260:22: error: use of undeclared identifier 'dev'; did you mean 'div'?
                if (!dev)
                     ^~~
                     div
/usr/include/stdlib.h:852:14: note: 'div' declared here
extern div_t div (int __numer, int __denom)
             ^
../../src/systemd/src/shared/bootspec.c:1260:22: warning: address of function 'div' will always evaluate to 'true' [-Wpointer-bool-conversion]
                if (!dev)
                    ~^~~
../../src/systemd/src/shared/bootspec.c:1260:22: note: prefix with the address-of operator to silence this warning
                if (!dev)
                     ^
                     &
fatal error: too many errors emitted, stopping now [-ferror-limit=]
1 warning and 20 errors generated.
```

kreijack:
> Also, compilers show error:
> 
> ```
> ../../src/systemd/src/shared/bootspec.c:1242:19: error: use of undeclared identifier 'blkid_dev_iterate_endp'
>         _cleanup_(blkid_dev_iterate_endp) blkid_dev_iterate iter = NULL;
>                   ^
> ../../src/systemd/src/shared/bootspec.c:1242:43: error: unknown type name 'blkid_dev_iterate'
>         _cleanup_(blkid_dev_iterate_endp) blkid_dev_iterate iter = NULL;
>                                           ^
> ../../src/systemd/src/shared/bootspec.c:1243:9: error: use of undeclared identifier 'blkid_cache'
>         blkid_cache cache = NULL;

I noticed that, however I have to point out:
- I was unable to reproduce on my setup
- The other "automatic check" ends successfully (I checked Build test / build (clang, 12, gold, gcrypt) (pull_request) , and I was able to find the generation of the bootctl:
```
[...]
2022-02-12T10:46:14.6641894Z [923/2038] clang-12 -Ibootctl.p -I. -I.. -Isrc/basic -I../src/basic -Isrc/fundamental -I../src/fundamental -Isrc/systemd -I../src/systemd -I../src/libsystemd/sd-bus -I../src/libsystemd/sd-device -I../src/libsystemd/sd-event -I../src/libsystemd/sd-hwdb -I../src/libsystemd/sd-id128 -I../src/libsystemd/sd-journal -I../src/libsystemd/sd-netlink -I../src/libsystemd/sd-network -I../src/libsystemd/sd-resolve -Isrc/shared -I../src/shared -I/usr/include/blkid -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=gnu99 -O0 -g -Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wendif-labels -Werror=format=2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=overflow -Werror=override-init -Werror=return-type -Werror=shift-count-overflow -Werror=undef -Wfloat-equal -Winit-self -Wmissing-include-dirs -Wmissing-noreturn -Wnested-externs -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wshadow -Wstrict-aliasing=2 -Wstrict-prototypes -Wunused-function -Wwrite-strings '-Wno-error=#warnings' -Wno-string-plus-int -Wno-typedef-redefinition -Wno-gnu-variable-sized-type-not-at-end -Werror=missing-declarations -Werror=missing-prototypes -fdiagnostics-show-option -fno-common -fno-strict-aliasing -fstack-protector -fstack-protector-strong -fvisibility=hidden --param=ssp-buffer-size=4 -Werror=shadow -include config.h -Werror -MD -MQ bootctl.p/src_boot_bootctl.c.o -MF bootctl.p/src_boot_bootctl.c.o.d -o bootctl.p/src_boot_bootctl.c.o -c ../src/boot/bootctl.c
2022-02-12T10:46:14.7151255Z [924/2038] clang-12  -o bootctl bootctl.p/src_boot_bootctl.c.o -Wl,--as-needed -Wl,--no-undefined -fuse-ld=gold -Wl,--fatal-warnings -Wl,-z,now -Wl,-z,relro -fstack-protector -Wl,--warn-common -Werror '-Wl,-rpath,$ORIGIN/src/shared' -Wl,-rpath-link,/home/runner/work/systemd/systemd/build/src/shared -Wl,--start-group src/shared/libsystemd-shared-250.so /usr/lib/x86_64-linux-gnu/libblkid.so -Wl,--end-group
[...]
```
So I am incline to think that it is a fault of the automatic check:

evverx:
Can you try building systemd with `meson -Dblkid=false && ninja -C ./build`? I suspect the code should be hidden behind `#if HAVE_BLKID`.

kreijack:
> The first commit looks good to me.
> 
> The second commit seems to be a duplicated work contained in #22487, and I prefer the way used in #22487 as it handles the situation gracefully rather than showing warning to the user.
> 
May be that my description was not clear enough. However my 2nd commit check only about the presence of two (or more) EFI SYSTEM gpt partition. #22487 doesn't seems to me that checks this kind of issue.

I am thinking that you are referring to the 5th commit, the one that show the status "available"/"unavailable". I tried to show the truth: the boot entry is not available.

> Not sure about the commits about random seed things.
I think that this check make sense, because is the only thing that can ensure if the boot partition that the system sees is the one that systemd-boot sees.

evverx:
Looks like CIFuzz no longer fails so I'll go ahead and remove the label.

src/shared/bootspec.h
```diff
@@ -84,3 +84,8 @@ static inline const char* boot_entry_title(const BootEntry *entry) {
 
 int find_esp_and_warn(const char *path, bool unprivileged_mode, char **ret_path, uint32_t *ret_part, uint64_t *ret_pstart, uint64_t *ret_psize, sd_id128_t *ret_uuid);
 int find_xbootldr_and_warn(const char *path, bool unprivileged_mode, char **ret_path,sd_id128_t *ret_uuid);
+
+#define GUID_EFI_SYSTEM_PARTITION "c12a7328-f81f-11d2-ba4b-00a0c93ec93b"
+#define GUID_EFI_SYSTEM_PARTITION_DOS "0xef"
+#define GUID_XBOOTLDR_PARTITION "bc13c2ff-59e6-4262-a352-b275fd6f7172"
+#define GUID_XBOOTLDR_PARTITION_DOS "0xea"
```
poettering, line 94: We already have defines for the GPT UUIDs in src/shared/gpt.h. Let's not duplicate this here. I prepped #22507 now that takes inspiration from your patch, but fixes this across the tree in a generic fashion.

The DOS partition type is not a GUID btw, just an 8byte value. We probably should have a helper in bootspec.c that compares things with that:

```c
static int dos_parttype_equal(const char *s, uint8_t v) {
    const char *p;
    uint8_t w; 
    int r;

    p = startswith_nocase(s, "0x");
    if (!p)
            return false;
    r = safe_atoux8(p, &w);
    if (r < 0) 
            return r;
    return w == v;
}

#define DOS_PARTTYPE_ESP 0xefu
#define DOS_PARTTYPE_XBOOTLDR 0xeau
```

and then use:

```c
if (dos_parttype_equal(s, DOS_PARTTYPE_ESP) > 0) …
```

src/boot/bootctl.c
```diff
@@ -68,6 +68,24 @@ static const char *arg_dollar_boot_path(void) {
         return arg_xbootldr_path ?: arg_esp_path;
 }
 
+static void check_efi_path_duplicates(void) {
+        if (check_efi_duplicates(geteuid() != 0) > 1) {
```
poettering, line 75: we usually prefere an "early-exit" coding style over a "deeply nested" coding style.

i.e. reverse this check and return if <= 0.

src/boot/efi/random-seed.c
```diff
@@ -307,5 +315,11 @@ EFI_STATUS process_random_seed(EFI_FILE *root_dir, RandomSeedMode mode) {
         if (EFI_ERROR(err))
                 return log_error_status_stall(err, L"Failed to write random seed to EFI variable: %r", err);
 
+        /* Save a derived new seed values to check the partition id */
+        err = efivar_set_raw(LOADER_GUID, L"LoaderPartitionRandomSeed", for_check_efi_part, HASH_VALUE_SIZE, 0);
```
poettering, line 319: what's the point of this? we aleady have LoaderDevicePartUUID that expressly identifies the GPT partition. Why suddenly turn the random seed into something it is not supposed to be used for so far?

src/boot/bootctl.c
```diff
@@ -508,6 +508,18 @@ static int boot_entry_show(const BootEntry *e, bool show_as_default) {
                         (void) terminal_urlify_path(e->path, NULL, &link);
 
                 printf("       source: %s\n", link ?: e->path);
+
+                /*
+                 * if the entry is a NON "auto-*" one and if the information comes
+                 * from the bootloader only, likely the entry was seen by
+                 * the systemd-boot but then it was removed
+                 */
+                if (!startswith(e->id, "auto-")) {
```
poettering, line 517: PR #22487 already covers this more comprehensively.

poettering (changes requested):

poettering:
I am very sure we should not check unrelated disks. i.e. if somebody plugs in some USB stick with an ESP that is really no reason to warn, because there's nothing wrong with that. There might be a reason to warn about multiple ESPs on the disk backing the root file system, but I am not sure it really deserves so much infrastructure? We could probably return that info in dissect_image() in some additional return value if we wanted, that should be much simpler.

What I am particularly puzzled about though is the random seed thing. Why? We already pass the partition uuid along, which should be all we need to identify partitions/file systems, and it exists *precisely* for this reason.

poettering:
I presume there was a very specific case you @kreijack ran into where things fell appart for you, that made you prep this PR. Can you elaborate on that?

src/shared/bootspec.h
kreijack: I was unaware of "gpt.h"; good to know.

src/boot/bootctl.c
kreijack: ok

src/boot/efi/random-seed.c
kreijack: see my comment below.

kreijack:
Hi Lennart,

the code was inspire by this discussion
https://github.com/systemd/systemd/pull/22161#issuecomment-1032992992

> We at least hope these are the same sources. But it's a real hard problem to find the right devices from Linux userspace that the uefi firmware used before. We typically go via GPT partition GUID, but once that is not unique we have a problem.

what about a system which boots with multiple ESP partition ? this is the case that inspired the checks. 

( a bit of topic: may be that I am so old that when I think to examine an hd in an another host I think to connect it to the sata cable. So it appears from the boot. You are younger to me ( :-) ) so you think to connect it to a USB via an adapter, which is discovered after the boot... )

Both the cases are realistic. The main difference is that in the case of an USB adapter, it will result in false positive and should not requires warning. The former (less likely in these day I have to admit) is still a problem.

Regarding the random-seed vs PARTUUID it could be problematic when the system is cloned to/from another disk.

The random seed is a key unique enough to allow us to guarantee the uniqueness of a partition.

poettering:
> what about a system which boots with multiple ESP partition ? this is the case that inspired the checks.

Sure, but the partition UUID we already have should be fine for that. Why add another ID?

> Both the cases are realistic. The main difference is that in the case of an USB adapter, it will result in false positive and should not requires warning. The former (less likely in these day I have to admit) is still a problem.

But we do pass the UUID over. If there are two ESP partitions accessible from the same system and they carry the *same* UUID, then this a bigger problem.

> Regarding the random-seed vs PARTUUID it could be problematic when the system is cloned to/from another disk.
> 
> The random seed is a key unique enough to allow us to guarantee the uniqueness of a partition.

The random seed is not really created for this purpose and not updated on read-only media. It's also not used if SecureBoot is on (because it is not cryptographically verified, i.e. cannot be trusted to seed the random pool). I am thus not convinced it's useful as an identifier.

It appears to me that instead of bothering with the random seed we should instead just check in bootctl:

1. if there are any ambiguities around the boot loader reported ESP partition UUID, i.e. more than one partition block device matching it
2. if there are more than a single ESP on the partition table backing the rootfs
3. if there are more than a single XBOOTLDR partitions on the partition table backing the rootfs

The first thing should be checked via sd-device (and not via blkid). The second thing should be done via a minor patch to dissect-image.c. Currently, if there are ambiguities we ignore all but the first partition of a specific type. Let's slightly extend DissectedPartition structure with a new boolean `ambiguous` or so, which we'll set to true if this case is hit. Then display this information in bootctl for ESP/XBOOTLDR. See how systemd-gpt-auto-generator does runs dissect_image() for the rootfs device.

kreijack:
> But we do pass the UUID over. If there are two ESP partitions accessible from the same system and they carry the _same_ UUID, then this a bigger problem.

This is still prone to the problem related to the cloned disk. Anyway you are right pointing out that the random seed is not updated when secure-boot is enabled; and because secure-boot is here, we can't follow this path.

[...]
> 
> It appears to me that instead of bothering with the random seed we should instead just check in bootctl:
> 
>     1. if there are any ambiguities around the boot loader reported ESP partition UUID, i.e. more than one partition block device matching it
> 
>     2. if there are more than a single ESP on the partition table backing the rootfs
> 
>     3. if there are more than a single XBOOTLDR partitions on the partition table backing the rootfs
> 
> 
> The first thing should be checked via sd-device (and not via blkid). 

In the code I found the following comment:
```
        /* If we are unprivileged we ask udev for the metadata about the partition. If we are privileged we
         * use blkid instead. Why? Because this code is called from 'bootctl' which is pretty much an
         * emergency recovery tool that should also work when udev isn't up (i.e. from the emergency shell),
         * however blkid can't work if we have no privileges to access block devices directly, which is why
         * we use udev in that case. */
```

This pushed me to implement the double check using both "sd-device" (which is called 'udev') and "blkid". This puzzled me because we can have different behavior depending by the contexts: if it is an emergency shell, if the user is root or not...

If the fear is that udev is not available, we can use blkid only, rebuilding the cache if the user is root.

However we should consider that BLKID is not an hard requirements for sd-boot.

> The second thing should be done via a minor patch to dissect-image.c. Currently, if there are ambiguities we ignore all but the first partition of a specific type. Let's slightly extend DissectedPartition structure with a new boolean `ambiguous` or so, which we'll set to true if this case is hit. Then display this information in bootctl for ESP/XBOOTLDR. See how systemd-gpt-auto-generator does runs dissect_image() for the rootfs device.

Apart the way of doing that, could you elaborate the differences between 1) and 2): in both the cases I check about duplicates of the "partition UUID". My understanding is that 2) applies only when two "partition UUID"s exists in one HD/SSD; but this is dangerous like 1).

What would make sense is to check duplication of ESP only in disks that are present at boot time. 
However the fact that the disk was inserted AFTER the boot, doesn't prevent to be present in the next boot (i.e. a removable disk plugged, and leaved in the pc).

My suggestion is to implement only checks against duplicated ESP/XBOOTLDR without caring the backing disk.

poettering:
> This is still prone to the problem related to the cloned disk

But is this really something we need to focus on? Sure if you have two identical disks in a system and boot from one we might get confused, but you kinda asked for breakage then, so not sure this is worth wasting too much thoughts about? It's not that this is a typical scenario?

poettering:
> ```
> 
> ```
> 
> 
>     
>   
> 
> This pushed me to implement the double check using both "sd-device" (which is called 'udev') and "blkid". This puzzled me because we can have different behavior depending by the contexts: if it is an emergency shell, if the user is root or not...

Hmm, I guess this makes sense then. So typically we very much prefer using udev for everything, as the one source of truth. But yes, bootctl should probably work without it, indeed.

poettering:
> Apart the way of doing that, could you elaborate the differences between 1) and 2): in both the cases I check about duplicates of the "partition UUID". My understanding is that 2) applies only when two "partition UUID"s exists in one HD/SSD; but this is dangerous like 1).

The difference is about the two kinds of uuid partitions have: the partition uuid and the partition *type* uuid. It's a problem if more than one partition *anywhere* carry the same partition uuid. However, it's *not* a problem if more than one partition *anywhere* carry the same partition *type* uuid, in fact it's quite common, since pretty much any PC will have a partition marked with the ESP partition type. However, it's a problem if there are two partitions on the same disk that both have the ESP partition type. Hence it might make sense to have a check explicitly for that, on the disk backing the ESP used for booting.

poettering:
> Hmm, I guess this makes sense then. So typically we very much prefer using udev for everything, as the one source of truth. But yes, bootctl should probably work without it, indeed.

Hmm, so I should probably extend on that: I think using libblkid for enumerating partitions on a given disk is fine, and a good alternative to doing that via udev. However, enumeration disks, is probably something we should not do with libblkid, but always with udev.

poettering:
ping?

poettering:
it would probably be good to split this PR up so that the less controversial parts can be merged already.

kreijack:
> it would probably be good to split this PR up so that the less controversial parts can be merged already.

I will create a new PR about Patch #1
About patch #2, it was addressed by #22507; only the MBR part was not addressed. This may be another separate PR
About patch #3,#4 I don't see consensus so we can ignore it; may be I will reopen a new PR when I find a better idea
About patch #5, it was addressed by #22487

kreijack:
> > it would probably be good to split this PR up so that the less controversial parts can be merged already.
> 
> I will create a new PR about Patch #1 About patch #2, it was addressed by #22507; only the MBR part was not addressed. This may be another separate PR About patch #3,#4 I don't see consensus so we can ignore it; may be I will reopen a new PR when I find a better idea About patch #5, it was addressed by #22487

I reopened a new PR, the #24230

PR #22498, opened 2022-02-13
cgroup: ignore error in attaching process when threaded mode is used

### LINKED ISSUE #22486: `systemctl reload` doesn't work if "threaded" mode subcgroups are created immediately below cgroups of services with delegation enabled

```
$ systemd --version
systemd 249 (249.4-1-cloudflare-2021.9.0)
+PAM +AUDIT +SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT +GNUTLS -OPENSSL +ACL +BLKID +CURL +ELFUTILS -FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP -LIBFDISK +PCRE2 -PWQUALITY -P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified
```

**Used distribution**

Debian Bullseye

```
5.15.15-cloudflare-2022.1.9
```

x86_64, but applies to all

Reload with cgrpupv2 subtrees works as well as it did in cgroupv1.

```
Feb 11 19:52:01 heh systemd[1]: bug.service: Control process exited, code=exited, status=219/CGROUP
```

Create the following systemd unit (named `bug.service`):

```
[Service]
Delegate=true
ExecStartPre=/bin/mkdir /sys/fs/cgroup/system.slice/bug.service/subtree
ExecStartPre=/bin/bash -c "echo threaded > /sys/fs/cgroup/system.slice/bug.service/subtree/cgroup.type"
ExecStart=/bin/sleep 86400
ExecReload=/bin/echo pretending to reload
```

Start the unit, then issue `systemctl reload bug`, then observe:

```
$ sudo systemctl reload bug
Job for bug.service failed.
See "systemctl status bug.service" and "journalctl -xeu bug.service" for details.
```

In the logs you'll see:

```
Feb 11 19:52:01 heh systemd[1]: Reloading bug.service...
Feb 11 19:52:01 heh systemd[1]: bug.service: Control process exited, code=exited, status=219/CGROUP
Feb 11 19:52:01 heh systemd[1]: Reload failed for bug.service.
```

If you `strace` pid 1, you can see:

```
stat("/sys/fs/cgroup/system.slice/bug.service", {st_mode=S_IFDIR|0755, st_size=0, ...}) = 0
mkdir("/sys/fs/cgroup/system.slice/bug.service/.control", 0755) = -1 EEXIST (File exists)
removexattr("/sys/fs/cgroup/system.slice/bug.service/.control", "user.oomd_avoid") = -1 ENODATA (No data available)
removexattr("/sys/fs/cgroup/system.slice/bug.service/.control", "user.oomd_omit") = -1 ENODATA (No data available)
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fe1ca234c10) = 3972575
openat(AT_FDCWD, "/sys/fs/cgroup/system.slice/bug.service/.control/cgroup.procs", O_WRONLY|O_NOCTTY|O_CLOEXEC) = 49
fcntl(49, F_GETFL)                      = 0x8001 (flags O_WRONLY|O_LARGEFILE)
write(49, "3972575\n", 8)               = -1 EOPNOTSUPP (Operation not supported)
close(49)                               = 0
```

Here's a gdb backtrace:

```
#0  __GI___libc_write (fd=49, buf=0x7ffe8973ba0a, nbytes=8) at ../sysdeps/unix/sysv/linux/write.c:26
#1  0x00007fe1ca993665 in _IO_new_file_write (f=0x56161cbe8ec0, data=0x7ffe8973ba0a, n=8) at fileops.c:1181
#2  0x00007fe1ca9929d6 in new_do_write (fp=fp@entry=0x56161cbe8ec0, data=data@entry=0x7ffe8973ba0a "3975788\n", to_do=to_do@entry=8) at libioP.h:948
#3  0x00007fe1ca993d5e in _IO_new_file_xsputn (n=8, data=<optimized out>, f=<optimized out>) at fileops.c:1255
#4  _IO_new_file_xsputn (f=0x56161cbe8ec0, data=<optimized out>, n=8) at fileops.c:1197
#5  0x00007fe1ca987fc7 in __GI__IO_fputs (str=str@entry=0x7ffe8973ba0a "3975788\n", fp=fp@entry=0x56161cbe8ec0) at libioP.h:948
#6  0x00007fe1cad38ecd in write_string_stream_ts (f=0x56161cbe8ec0, line=line@entry=0x7ffe8973ba0a "3975788\n", flags=flags@entry=WRITE_STRING_FILE_DISABLE_BUFFER, ts=ts@entry=0x0)
    at ../src/basic/fileio.c:159
#7  0x00007fe1cad39292 in write_string_file_ts (fn=0x56161ca943f0 "/sys/fs/cgroup/system.slice/bug.service/.control/cgroup.procs", line=0x7ffe8973ba0a "3975788\n", flags=WRITE_STRING_FILE_DISABLE_BUFFER,
    ts=0x0) at ../src/basic/fileio.c:283
#8  0x00007fe1cac93cea in write_string_file (flags=WRITE_STRING_FILE_DISABLE_BUFFER, line=0x7ffe8973ba0a "3975788\n", fn=<optimized out>) at ../src/basic/fileio.h:58
#9  cg_attach (controller=controller@entry=0x56161c67fbc9 "_systemd", path=0x56161cc9d980 "/system.slice/bug.service/.control", pid=pid@entry=3975788) at ../src/shared/cgroup-setup.c:270
#10 0x000056161c5cf875 in exec_spawn (unit=<optimized out>, command=0x56161caae760, context=<optimized out>, params=<optimized out>, runtime=0x0, dcreds=0x56161cc6e5e0, ret=0x7ffe8973bc30)
    at ../src/core/execute.c:4819
#11 0x000056161c604529 in service_spawn (s=s@entry=0x56161cc6db40, c=<optimized out>, timeout=<optimized out>, flags=<optimized out>,
    flags@entry=(EXEC_APPLY_SANDBOXING | EXEC_APPLY_CHROOT | EXEC_IS_CONTROL | EXEC_CONTROL_CGROUP), ret_pid=ret_pid@entry=0x56161cc6e5f4) at ../src/core/service.c:1568
#12 0x000056161c607422 in service_enter_reload (s=0x56161cc6db40) at ../src/core/service.c:2321
#13 service_reload (u=0x56161cc6db40) at ../src/core/service.c:2531
#14 0x000056161c5d20fb in unit_reload (u=0x56161cc6db40) at ../src/core/unit.c:2027
#15 job_perform_on_unit (j=0x7ffe8973beb0) at ../src/core/job.c:780
#16 0x000056161c5e56ca in job_run_and_invalidate (j=<optimized out>) at ../src/core/job.c:844
#17 manager_dispatch_run_queue (source=<optimized out>, userdata=<optimized out>) at ../src/core/manager.c:2200
#18 manager_dispatch_run_queue (source=<optimized out>, userdata=0x56161ca403a0) at ../src/core/manager.c:2189
#19 0x00007fe1cadf8335 in source_dispatch (s=s@entry=0x56161ca40e80) at ../src/libsystemd/sd-event/sd-event.c:3544
#20 0x00007fe1cadf888d in dispatch_exit (e=<optimized out>) at ../src/libsystemd/sd-event/sd-event.c:3654
#21 sd_event_dispatch (e=e@entry=0x56161ca40ba0) at ../src/libsystemd/sd-event/sd-event.c:4071
#22 0x00007fe1cadf9e40 in sd_event_run (e=0x56161ca40ba0, timeout=18446744073709551615) at ../src/libsystemd/sd-event/sd-event.c:4139
#23 0x000056161c5f2b0a in manager_loop (m=0x56161ca403a0) at ../src/core/manager.c:2997
#24 0x000056161c5a7b9a in invoke_main_loop (ret_error_message=<optimized out>, ret_switch_root_init=<optimized out>, ret_switch_root_dir=<optimized out>, ret_fds=<optimized out>,
    ret_shutdown_verb=<optimized out>, ret_retval=<optimized out>, ret_reexecute=<optimized out>, saved_rlimit_memlock=<optimized out>, saved_rlimit_nofile=<optimized out>, m=<optimized out>)
    at ../src/core/main.c:1919
#25 main (argc=4, argv=<optimized out>) at ../src/core/main.c:2926
```

#22095 addressed `systemctl status` bug, but reload still fails.

yuwata:
Closes #22486.

### DIFF
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -4250,6 +4250,12 @@ static int exec_child(
                 }
 
                 r = cg_attach_everywhere(params->cgroup_supported, p, 0, NULL, NULL);
+                if (r == -EUCLEAN) {
+                        *exit_status = EXIT_CGROUP;
+                        return log_unit_error_errno(unit, r, "Failed to attach process to cgroup %s "
+                                                    "because the cgroup or one of its parents or "
+                                                    "siblings is in the threaded mode: %m", p);
+                }
                 if (r < 0) {
                         *exit_status = EXIT_CGROUP;
                         return log_unit_error_errno(unit, r, "Failed to attach to cgroup %s: %m", p);
--- a/src/shared/cgroup-setup.c
+++ b/src/shared/cgroup-setup.c
@@ -346,6 +346,9 @@ int cg_attach(const char *controller, const char *path, pid_t pid) {
         xsprintf(c, PID_FMT "\n", pid);
 
         r = write_string_file(fs, c, WRITE_STRING_FILE_DISABLE_BUFFER);
+        if (r == -EOPNOTSUPP && cg_is_threaded(controller, path) > 0)
+                /* When the threaded mode is used, we cannot read/write the file. Let's return recognizable error. */
+                return -EUCLEAN;
         if (r < 0)
                 return r;
 


src/basic/cgroup-util.c
```diff
@@ -1688,6 +1688,36 @@ int cg_slice_to_path(const char *unit, char **ret) {
...
+        if (r < 0)
+                return r;
+
+        r = read_one_line_file(fs, &contents);
+        if (r == -ENOENT)
+                return false;
+        if (r < 0)
+                return r;
+
+        return streq_ptr(contents, "domain threaded");
```
poettering: hmm, this should probably also look for the string "threaded" here, given how generically the function is called.

src/basic/cgroup-util.c
```diff
@@ -1688,6 +1688,36 @@ int cg_slice_to_path(const char *unit, char **ret) {
...
 
+static int cg_is_threaded_one(const char *controller, const char *path) {
+        _cleanup_free_ char *fs = NULL, *contents = NULL;
+        int r;
+
+        r = cg_get_path(controller, path, "cgroup.type", &fs);
+        if (r < 0)
+                return r;
+
+        r = read_one_line_file(fs, &contents);
```
poettering: rather read_virtual_file() maybe?

src/basic/cgroup-util.c
```diff
@@ -1688,6 +1688,36 @@ int cg_slice_to_path(const char *unit, char **ret) {
...
+                return r;
+
+        return streq_ptr(contents, "domain threaded");
+}
+
+int cg_is_threaded(const char *controller, const char *path) {
+        char prefix[strlen(path) + 1];
+        int r;
+
+        PATH_FOREACH_PREFIX_MORE(prefix, path) {
```
poettering: hmm, the way I understood it regular cgroups can never be below threaded ones, hence going up the tree should not be necessary?

src/core/execute.c
```diff
@@ -4250,7 +4250,9 @@ static int exec_child(
                 }
 
                 r = cg_attach_everywhere(params->cgroup_supported, p, 0, NULL, NULL);
-                if (r < 0) {
+                if (r == -EUCLEAN)
+                        log_unit_warning(unit, "Couldn't attach process to cgroup %s. The threaded mode is detected, ignoring.", p);
```
poettering: I think an explicit error message here is a good idea, but I think we still should fail here, as before with EXIT_CGROUP. The daemon is doing something wrong to place a threaded cgroup immediately below the delegated one, and we should make that discoverable easily, via a friendly log message, but not hide handle this gracefully otherwise.

poettering (changes requested):

poettering:
Codewise looks pretty OK, but I think we should ultimately still fail if we encounter this. it's a bug in the invoked daemon, and we shouldn't tape over that. See comments above. 

And see #22509

yuwata:
@poettering Thank you for the review. Updated. PTAL.

src/basic/cgroup-util.c
```diff
@@ -1688,6 +1688,25 @@ int cg_slice_to_path(const char *unit, char **ret) {
...
+
+        r = read_full_virtual_file(fs, &contents, NULL);
+        if (r == -ENOENT)
+                return false; /* Assume no. */
+        if (r < 0)
+                return r;
+
+        /* If the cgroup is in the threaded mode, it contains "threaded".
+         * If one of the parents is in the threaded more, it may contain "invalid". */
+        return !!STRSTR_IN_SET(contents, "threaded", "invalid");
```
poettering: hmm, strstr() is icky. I mean, one day kernel people might add a new value "non-threaded" or so, and strstr("threaded") will still consider it threaded...

maybe use strv_split() and then strv_contains()? i.e. make sure to use whitespace as field separators, so that we aren't confused by the above

src/core/execute.c
```diff
@@ -4250,6 +4250,10 @@ static int exec_child(
                 }
 
                 r = cg_attach_everywhere(params->cgroup_supported, p, 0, NULL, NULL);
+                if (r == -EUCLEAN) {
+                        *exit_status = EXIT_CGROUP;
+                        return log_unit_error_errno(unit, r, "The threaded mode is detected, and could not attach process to cgroup %s.", p);
```
poettering: maybe: "Unsupported use of threaded control group mode detected. Failed to attach process to cgroup '%s'."

keszybz: Please squash this commit with the previous one… they make no sense separately.

keszybz: > maybe: "Unsupported use of threaded control group mode detected. Failed to attach process to cgroup '%s'."

Or maybe even `"Cgroup %s is threaded. Failed to attach process to cgroup: %m", p` ?

poettering: The thing is that it's a sibling cgroup that is threaded not the one we are interested in. Hence the more vague wording I picked.

keszybz: Hmm, but doesn't the sibling group being threaded make this group threaded too?
`"Failed to attach process to cgroup %s because one of its siblings is threaded: %m", p` ?

yuwata: Maybe "parents" instead of "siblings"? Though, I am not familiar with the correct terms for cgroup...

anitazha: Siblings is the right term to use for cgroups that have the same parent cgroup. It's sprinkled throughout the cgroups documentation.

yuwata: Thanks. But from https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#threads, IIUC cgroup becomes "invald" state when its parent is in threaded mode. So maybe `"Failed to attach process to cgroup %s because the cgroup or one of its siblings or parent is in threaded mode: %m"`?

yuwata:
@poettering, @keszybz, and @anitazha Thank you for your review and comments. Updated.
- strv_split() and strv_contains() are now used for detecting the threaded mode.
- update the log message.

PTAL.

src/core/execute.c
poettering: sounds ok to me.

bluca (approved):

PR #22499, opened 2022-02-13
A couple of Packit/Super-Linter tweaks

mrc0mmand:


yuwata (approved):
LGTM.

evverx:
Thanks!

evverx:
@mrc0mmand I don't think superlinter should complain about GITHUB_TOKEN when MULTI_STATUS isn't set either but turning it off explicitly with MULTI_STATUS=false is better anyway :-)

evverx:
FWIW @mrc0mmand to judge from https://github.com/evverx/systemd/pull/89 `super-linter` has been released with your patch included so I think if those errors have been bugging you it should be possible to pin superlinter to that commit without waiting for Dependabot to update it in March.

mrc0mmand:
> FWIW @mrc0mmand to judge from [evverx#89](https://github.com/evverx/systemd/pull/89) `super-linter` has been released with your patch included so I think if those errors have been bugging you it should be possible to pin superlinter to that commit without waiting for Dependabot to update it in March.

Not really, I was just curious if there's a way to get rid of them, since they might be confusing to users which are not familiar with super-linter. So I guess we can wait for the regular Dependabot update :-)

PR #22500, opened 2022-02-13
Documentation on WantedBy for template units

VannTen:
I just found out a feature that could use a little more documentation, I think.

When enabling a template unit, if this unit list another template unit in its
`WantedBy`, you don't need to specify and instance name; each instance of the
unit listed in `WantedBy` will have a dependency on the corresponding instance
of the first unit. E.g:
`foo@.service`
```
[Service]
ExecStart=echo %i

[Install]
WantedBy=test@.target
```
and `test@.target`
```
[Unit]
Description=Test target
```
`systemctl enable foo@.service && systemctl start test@bar.target` will start
`foo@bar.service` (and `systemctl start test@foo` will start `foo@foo.service`
and so on).

I'm not sure where it should be documented, for now I just corrected the
WantedBy section in systemd.unit.xml regarding template units.

### DIFF
--- a/man/systemd.unit.xml
+++ b/man/systemd.unit.xml
@@ -1894,9 +1894,9 @@
         <para><command>WantedBy=foo.service</command> in a service
         <filename>bar.service</filename> is mostly equivalent to
         <command>Alias=foo.service.wants/bar.service</command> in the
-        same file. In case of template units, <command>systemctl
-        enable</command> must be called with an instance name, and
-        this instance will be added to the
+        same file. In case of template units listing non template units,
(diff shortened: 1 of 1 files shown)


keszybz:
LGTM.

celesteking:
That's great, but what about the reverse case, where you reference a template unit in a non-template unit?
Because right now for `[Install] WantedBy=b@.service` I'm getting a `Unit /usr/lib/systemd/system/a.service is added as a dependency to a non-existent unit b@.service` warning.

If referencing template units in install sections is forbidden, 1) why? 2) update docs to mention that.

PR #22507, opened 2022-02-14
sd-id128: comparison tweaks

poettering:
Inspired by #22491

### DIFF
--- a/src/veritysetup/veritysetup-generator.c
+++ b/src/veritysetup/veritysetup-generator.c
@@ -213,7 +213,7 @@ static int parse_proc_cmdline_item(const char *key, const char *value, void *dat
                         return log_oom();
 
         } else if (proc_cmdline_key_streq(key, "usrhash")) {
-                
+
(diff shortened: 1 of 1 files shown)


yuwata (approved):

src/home/homework-luks.c
```diff
@@ -1859,7 +1859,7 @@ static int make_partition_table(
         if (!t)
                 return log_oom();
 
-        r = fdisk_parttype_set_typestr(t, "773f91ef-66d4-49b5-bd83-d683bf40ad16");
+        r = fdisk_parttype_set_typestr(t, SD_ID128_TO_UUID_STRING(GPT_USER_HOME));
```
keszybz, line 1862: Hmm, but this replaces a const string in static storage with a dynamically allocated buffer we fill byte-by-byte.
We shouldn't do this as a general pattern. I'll submit a patch to use GPT_USER_HOME_STR instead.

PR #22512, opened 2022-02-14
virt: Fix Xen Dom0 detection logic to no longer report as VM

### LINKED ISSUE #22511: `systemd-detect-virt` giving incorrect result for non-nested Xen Dom0
When running `systemd-detect-virt` on a Xen dom0 which isn't nested within another hypervisor, the utility reports to be a Xen VM. It seems that prior to [this patch](https://github.com/systemd/systemd/commit/599be274c13c503806c85073d7beb1a155ac27bd) Xen Dom0 did not report as a Xen VM, which I understand was the correct behaviour.

I think I have identified the problem, and will send a pull request with a fix referencing this issue. More details as follows:

The new behaviour continues after the Xen check in the `detect_vm` function, in order to determine if Xen is running as a nested hypervisor that can be detected via a CPUID check. If dom0 is nested, then this is correct behaviour.

However, if dom0 is not nested and the CPUID check returns `VIRTUALIZATION_NONE`, the change has an unintended consequence because it proceeds with other checks where it previously did not. When the later check `detect_vm_hypervisor` reads `/sys/hypervisor/type`, which exists and reports Xen on both Dom0 and DomU, `VIRTUALIZATION_XEN` is returned. Because the Dom0 check was moved earlier and already disregarded, whether or not we are running Dom0 is not considered at this point, and the result stays as `VIRTUALIZATION_XEN`.

```
# systemctl --version 
systemd 250 (250.3+)
-PAM -AUDIT -SELINUX -APPARMOR +IMA -SMACK +SECCOMP -GCRYPT -GNUTLS -OPENSSL +ACL +BLKID -CURL -ELFUTILS -FIDO2 -IDN2 -IDN -IPTC +KMOD -LIBCRYPTSETUP +LIBFDISK -PCRE2 -PWQUALITY -P11KIT -QRENCODE -BZIP2 -LZ4 -XZ -ZLIB +ZSTD -BPF_FRAMEWORK +XKBCOMMON +UTMP +SYSVINIT default-hierarchy=hybrid
```

**Used distribution**
Custom Linux distribution running linux-yocto 5.10.93 and xen 4.15

`Linux n1sdp 5.10.93-yocto-standard #1 SMP PREEMPT Tue Jan 25 14:35:58 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux`

AArch64

On Xen Dom0:
```
# systemd-detect-virt; echo $?
none
1
```

On Xen Dom0:
 ```
# systemd-detect-virt; echo $?
xen
0
```

`systemd-detect-virt` on Xen Dom0.

On Xen DomU, the output is correct.

Latest behaviour:
```
# SYSTEMD_LOG_LEVEL=debug systemd-detect-virt
Found cgroup2 on /sys/fs/cgroup/unified, unified hierarchy for systemd controller
No virtualization found in DMI vendor table.
Unable to read /sys/firmware/dmi/entries/0-0/raw, using the virtualization information found in DMI vendor table, ignoring: No such file or directory
UML virtualization not found in /proc/cpuinfo.
Virtualization XEN found (/proc/xen exists)
Virtualization XEN, found /sys/hypervisor/properties/features with value 00006805, XENFEAT_dom0 (indicating the 'hardware domain') is set.
No virtualization found in CPUID
Virtualization xen found in /sys/hypervisor/type
Found VM virtualization xen
xen
```

Behaviour from v249-stable (d4406e94a32d423d8a73deb7757fb09890afe2c4):
```
# SYSTEMD_LOG_LEVEL=debug systemd-detect-virt
Found cgroup2 on /sys/fs/cgroup/unified, unified hierarchy for systemd controller
No virtualization found in DMI vendor table.
Unable to read /sys/firmware/dmi/entries/0-0/raw, using the virtualization information found in DMI vendor table, ignoring: No such file or directory
UML virtualization not found in /proc/cpuinfo.
No virtualization found in CPUID
Virtualization XEN found (/proc/xen exists)
Virtualization XEN, found /sys/hypervisor/properties/features with value 00006805, XENFEAT_dom0 (indicating the 'hardware domain') is set.
Found VM virtualization none
none
```

### PR DESCRIPTION
**Pull request by Richard-Neill**
Moving the Xen check before the CPUID check, in order to handle the case where
a Xen domain is nested within a hypervisor which can be detected by via the
CPUID check, had an unintended consequence of causing Dom0 to report as a Xen
VM when it is not nested.

This patch stops further checks once it has been determined that Dom0 is not
nested within another hypervisor, meaning that the non-nested case matches its
previous logic (where it does not report as a VM).

Fixes #22511

### DIFF
--- a/src/basic/virt.c
+++ b/src/basic/virt.c
@@ -331,22 +331,13 @@ static int detect_vm_xen_dom0(void) {
 }
 
 static int detect_vm_xen(void) {
-        int r;
-
-        /* The presence of /proc/xen indicates some form of a Xen domain */
+        /* The presence of /proc/xen indicates some form of a Xen domain
+           The check for Dom0 is handled outside this function */
         if (access("/proc/xen", F_OK) < 0) {
                 log_debug("Virtualization XEN not found, /proc/xen does not exist");
                 return VIRTUALIZATION_NONE;
         }
         log_debug("Virtualization XEN found (/proc/xen exists)");
-
-        /* Ignore the Xen hypervisor if we are in Dom0 */
-        r = detect_vm_xen_dom0();
-        if (r < 0)
-                return r;
-        if (r > 0)
-                return VIRTUALIZATION_NONE;
-
         return VIRTUALIZATION_XEN;
 }
 
@@ -434,7 +425,7 @@ static int detect_vm_zvm(void) {
 int detect_vm(void) {
         static thread_local int cached_found = _VIRTUALIZATION_INVALID;
         bool other = false;
-        int r, dmi;
+        int r, dmi, xen_dom0 = 0;
 
         if (cached_found >= 0)
                 return cached_found;
@@ -461,19 +452,26 @@ int detect_vm(void) {
         r = detect_vm_uml();
         if (r < 0)
                 return r;
-        if (r == VIRTUALIZATION_VM_OTHER)
-                other = true;
-        else if (r != VIRTUALIZATION_NONE)
+        if (r != VIRTUALIZATION_NONE)
                 goto finish;
 
         /* Detect Xen */
         r = detect_vm_xen();
         if (r < 0)
                 return r;
-        if (r == VIRTUALIZATION_VM_OTHER)
-                other = true;
-        else if (r != VIRTUALIZATION_NONE)
-                goto finish;
+        if (r == VIRTUALIZATION_XEN) {
+                 /* If we are Dom0, then we expect to not report as a VM. However, as we might be nested
+                  * inside another hypervisor which can be detected via the CPUID check, wait to report this
+                  * until after the CPUID check. */
+                xen_dom0 = detect_vm_xen_dom0();
+                if (xen_dom0 < 0)
+                        return xen_dom0;
+                if (xen_dom0 == 0)
+                        goto finish;
+
+                r = VIRTUALIZATION_NONE;
+        } else if (r != VIRTUALIZATION_NONE)
+                assert_not_reached();
 
         /* Detect from CPUID */
         r = detect_vm_cpuid();
@@ -482,6 +480,10 @@ int detect_vm(void) {
         if (r == VIRTUALIZATION_VM_OTHER)
                 other = true;
         else if (r != VIRTUALIZATION_NONE)
+                goto finish;
+
+        /* If we are in Dom0 and have not yet finished, finish with the result of detect_vm_cpuid */
+        if (xen_dom0 > 0)
                 goto finish;
 
         /* Now, let's get back to DMI */

### CODE REVIEW

File: src/basic/virt.c [hunk 1]
```diff
@@ -333,20 +333,13 @@ static int detect_vm_xen_dom0(void) {
 static int detect_vm_xen(void) {
         int r;
```
yuwata:
```
../src/basic/virt.c: In function ‘detect_vm_xen’: 01:16
../src/basic/virt.c:334:13: error: unused variable ‘r’ [-Werror=unused-variable] 01:16
  334 |         int r; 01:16
      |             ^ 01:16
```

yuwata (changes requested):

src/basic/virt.c
```diff
@@ -472,8 +465,18 @@ int detect_vm(void) {
...
                 other = true;
-        else if (r != VIRTUALIZATION_NONE)
-                goto finish;
+        else if (r != VIRTUALIZATION_NONE) {
+                 /* If we are Dom0, then we expect to not report as a VM. However, as we might be nested
+                  * inside another hypervisor which can be detected via the CPUID check, wait to report this
+                  * until after the CPUID check. */
+                xen_dom0 = detect_vm_xen_dom0();
+                if (xen_dom0 < 0)
+                        return r;
```
poettering: I figure you mean `return xen_dom0` here?

src/basic/virt.c
```diff
@@ -472,8 +465,18 @@ int detect_vm(void) {
                 return r;
         if (r == VIRTUALIZATION_VM_OTHER)
                 other = true;
-        else if (r != VIRTUALIZATION_NONE)
-                goto finish;
+        else if (r != VIRTUALIZATION_NONE) {
+                 /* If we are Dom0, then we expect to not report as a VM. However, as we might be nested
+                  * inside another hypervisor which can be detected via the CPUID check, wait to report this
+                  * until after the CPUID check. */
+                xen_dom0 = detect_vm_xen_dom0();
```
poettering, line 466: it only makes sense to do this test if `r == VIRTUALIZATION_XEN` already applies, no? i.e. shouldn't the block be:

```c
…
else if (r == VIRTUALIZATION_XEN) {
        /* … */
        xen_dom0 = detect_vm_xen_dom0();
        if (xen_dom0 < 0)
                return xen_dom0;
        if (xen_dom0 > 0) {
                /* If we are in Xen Dom0, then we don't consider this a VM environment, because it is the domain that manages the hardware, i.e. it's the host, not the payload.
                   Don't bother with the other checks either in that case then, since we are only interested in the "innermost" VM environment, and if that's Xen Dom0, then we are not virtualized. */ 
                r = VIRTUALIZATION_NONE; 
                goto finish;
        }

        /* Otherwise: we detected Xen, and are not Dom0, hence return this as Xen */
        goto finish;
} else if (r != VIRTUALIZATION_NONE)
        goto finish;
…
```

poettering (changes requested):
So, once you detect Xen we don't have to check for any other hypervisor, we can just cut the test short: if we detect Dom0 we definitely are a Xen host (and thus should return NONE), and we are not Dom0 then we are Xen payload (and thus should return XEN), and either way shouldn't check CPUID or so anymore, it's not gonna give us any more help.

Richard-Neill:
> So, once you detect Xen we don't have to check for any other hypervisor, we can just cut the test short: if we detect Dom0 we definitely are a Xen host (and thus should return NONE), and we are not Dom0 then we are Xen payload (and thus should return XEN), and either way shouldn't check CPUID or so anymore, it's not gonna give us any more help.

This is what I expected to be the case when I first looked at this, but this patch https://github.com/systemd/systemd/commit/599be274c13c503806c85073d7beb1a155ac27bd implemented logic to handle a case where the Xen hypervisor is itself virtualised, i.e. Xen being nested on a hypervisor that is detected by CPUID. 

The linked patch states that the Xen check was moved above the CPUID check, so that DomU could be reported as Xen and Dom0 would continue to the CPUID check and also report its non-xen hypervisor. Without that patch, I think in the nested case DomU would report results from the CPUID check and Dom0 would report as none, and Xen wouldn't be detected at all.

So in this pull request, my intention is to retain that new logic if Xen is nested, while leaving the old logic (cut the test short) if we are not nested (no result from the CPUID check) and we are running Dom0.

It's not an easy thing to explain, but hopefully this makes sense - do let me know if I can clarify anything (or I have misunderstood). I do not have an environment available to me where Xen is nested within a hypervisor that is detectable through CPUID. Perhaps @bogsen could clarify the above or test on their environment if at all possible!

src/basic/virt.c
Richard-Neill: Correct - thanks!

src/basic/virt.c
```diff
@@ -472,8 +463,18 @@ int detect_vm(void) {
...
+                if (xen_dom0 < 0)
+                        return xen_dom0;
+                else if (xen_dom0 > 0)
+                        r = VIRTUALIZATION_NONE;
+                else
+                        goto finish;
```
poettering, line 470: please write this like this:

```c
if (xen_dom0 < 0)
        return xen_dom0;
if (xen_dom0 == 0)
        goto finish;

r = VIRTUALIZATION_NONE;
…
```

i.e. nest the jumps (i.e. the return + goto) in the conditional blocks, and thus get rid of needless indentation of the main codepath here. i.e. less indentation is a good thing

src/basic/virt.c
```diff
@@ -472,8 +463,18 @@ int detect_vm(void) {
                 return r;
         if (r == VIRTUALIZATION_VM_OTHER)
                 other = true;
-        else if (r != VIRTUALIZATION_NONE)
-                goto finish;
+        else if (r != VIRTUALIZATION_NONE) {
```
poettering: the assumption here is that r is VIRTUALIZATION_XEN. Maybe make this explicit via an assert(), i.e.

```c
…
else if (r != VIRTUALIZATION_NONE) {
        assert(r == VIRTUALIZATION_XEN);
        …
```

or maybe:

```
else if (r == VIRTUALIZATION_XEN) {
        …
else if (r != VIRTUALIZATION_NONE)
        assert_not_reached();
```

or so

poettering (commented):
looks good, just some minor nitpicks

yuwata:
Could you also mention the offending commit hash in the commit message? Then, we can easily backport this to stable branches.

src/basic/virt.c
Richard-Neill: Done

src/basic/virt.c
Richard-Neill: Makes sense to make it explicit. I notice that two checks:

detect_vm_uml
detect_vm_xen

Have no code path which returns VIRTUALIZATION_VM_OTHER. So I have also taken the opportunity to tidy the conditionals to not check this (for example !none and !xen should remain an assertion failure).

Richard-Neill:
> Could you also mention the offending commit hash in the commit message? Then, we can easily backport this to stable branches.

Done.

src/basic/virt.c
```diff
@@ -484,6 +482,10 @@ int detect_vm(void) {
         else if (r != VIRTUALIZATION_NONE)
                 goto finish;
 
+        /* If we are in Dom0 and have not yet finished, finish with the result of detect_vm_cpuid */
```
Richard-Neill, line 485: As detect_vm_cpuid can return VIRTUALIZATION_VM_OTHER and proceed, my previous comment stating that we return VIRTUALIZATION_NONE here was possibly incorrect, so fixed.

yuwata (approved):
LGTM.

Richard-Neill:
Hi @yuwata, do you expect to backport this patch to v250-stable? If so, is there a regular schedule when patches are backported or is it done ad-hoc? Thanks!

yuwata:
Waiting in https://github.com/systemd/systemd-stable/pull/170. We are planning a new point release after the PR is merged.

PR #22517, opened 2022-02-15
test: Stop using TextTestRunner in systemd-networkd-tests.py

daandemeyer:
If we assign our own test runner, passing arguments stops working
as unittest won't instantiate its own test runner after it parses
the arguments from sys.argv.

Consequence is that the tests will write to stderr now instead of
stdout since it doesn't seem possible to configure the stream that
unittest.main() will instantiate its test runner with so it'll
default to sys.stderr.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -5933,5 +5933,4 @@ if __name__ == '__main__':
         env.update({ 'UBSAN_OPTIONS' : ubsan_options })
 
     sys.argv[1:] = unknown_args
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
cc @mrc0mmand do you know if writing to sys.stderr instead of sys.stdout will break things?

mrc0mmand:
> cc @mrc0mmand do you know if writing to sys.stderr instead of sys.stdout will break things?

I don't think so (at least not in CentOS CI).

mrc0mmand (approved):

PR #7392, opened 2017-11-20
man: fix description of --force in halt(8)

keszybz:
https://bugzilla.redhat.com/show_bug.cgi?id=1449751

PR #22518, opened 2022-02-15
Revert "man: fix description of --force in halt(8) (#7392)"

lnussel:
This reverts commit 5d9adb5b60b815b477ba9e6b19ef0fd7e1854a38.

I guess there was some misunderstanding in the original bug report. AFAICT `halt_parse_argv()` always set `arg_force = 2` even if only one `-f` was passed.

### DIFF
--- a/man/halt.xml
+++ b/man/halt.xml
@@ -84,13 +84,8 @@
         <term><option>-f</option></term>
         <term><option>--force</option></term>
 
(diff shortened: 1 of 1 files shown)


keszybz:
Indeed. systemctl implements different behaviour in halt and in 'systemctl halt'. So actually we behave as originally described in the man page and implement behaviour compatible with the original halt command.

yuwata:
LGTM too.

PR #22519, opened 2022-02-15
sd-boot: rework boot entry sorting

poettering:
Replaces: #22391 #22066

Implements what is discussed here:

https://github.com/systemd/systemd/pull/22391#issuecomment-1040092633

### DIFF
--- a/TODO
+++ b/TODO
@@ -1129,6 +1129,10 @@ Features:
   - teach it to prepare an ESP wholesale, i.e. with mkfs.vfat invocation
   - teach it to copy in unified kernel images and maybe type #1 boot loader spec entries from host
   - make it operate on loopback files, dissecting enough to find ESP to operate on
+  - bootspec: properly support boot attempt counters when parsing entry file names
+
+* kernel-install:
+  - optionally, support generating type #2 entries instead of type #1, including signing them
 
 * logind:
   - logind: optionally, ignore idle-hint logic for autosuspend, block suspend as long as a session is around


src/boot/efi/boot.c
```diff
@@ -1624,12 +1633,36 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
...
-                return 1;
+        /* Order entries that have no tries left to the end of the list */
         if (a->tries_left == 0 && b->tries_left != 0)
+                return 1;
+        if (a->tries_left != 0 && b->tries_left == 0)
                 return -1;
 
+        /* First, order by sort key (alphabetically upwards), if it is defined. */
+        if (a->sort_key && b->sort_key) {
+                r = strcmp(a->sort_key, b->sort_key);
```
medhefgo, line 1672: `strcmp_ptr`?

src/boot/efi/boot.c
```diff
@@ -1624,12 +1633,36 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
...
+
+                /* If multiple installations of the same OS are around, group by machine ID */
+                if (a->machine_id && b->machine_id) {
+                        r = strcmp(a->machine_id, b->machine_id);
+                        if (r != 0)
+                                return r;
+                }
+
+                /* If the sort key was defined, then order by version now (downwards, putting the newest first) */
+                if (a->version && b->version) {
```
medhefgo: `strverscmp_improved` is NULL safe, so this can be dropped.

src/boot/efi/boot.c
```diff
@@ -1624,12 +1633,36 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
...
 
+        /* First, order by sort key (alphabetically upwards), if it is defined. */
+        if (a->sort_key && b->sort_key) {
+                r = strcmp(a->sort_key, b->sort_key);
+                if (r != 0)
+                        return r;
+
+                /* If multiple installations of the same OS are around, group by machine ID */
+                if (a->machine_id && b->machine_id) {
+                        r = strcmp(a->machine_id, b->machine_id);
```
medhefgo: here too

src/boot/efi/boot.c
```diff
@@ -1659,7 +1692,7 @@ static UINTN config_entry_find(Config *config, const CHAR16 *needle) {
         if (!needle)
                 return IDX_INVALID;
 
-        for (INTN i = config->entry_count - 1; i >= 0; i--)
+        for (UINTN i = 0; i < config->entry_count; i++)
```
medhefgo: Now this is getting really silly: You sort versions by `sort-key` descending and the stuff without one ascending. But now you select the default entry by the resulting sort order. This is messed up.

This will select the topmost entry, which will be correct for `sort-key` but breaks compatibility for entries without one (it will pick older versions first).

src/shared/bootspec.c
```diff
@@ -645,47 +689,45 @@ static int boot_entries_select_default(const BootConfig *config) {
...
-                for (i = config->n_entries - 1; i >= 0; i--)
-                        if (streq(config->entry_oneshot, config->entries[i].id)) {
-                                log_debug("Found default: id \"%s\" is matched by LoaderEntryOneShot",
-                                          config->entries[i].id);
-                                return i;
-                        }
+        if (config->entry_oneshot) {
+                i = boot_config_find(config, config->entry_oneshot);
+                if (i >= 0) {
+                        log_debug("Found default: id \"%s\" is matched by LoaderEntryOneShot",
```
medhefgo, line 694: Maybe show the pattern used for the efi var matches too?

medhefgo (changes requested):
I still don't like it for the very same reason outlined in the previous discussion. This also doesn't fix the `arch-lts.conf` vs `arch.conf` case either (I will consider reverting my commit without a fix for this a regression).

I do like the idea of `sort-key` though, but now your sorting/default selection is just… interesting as outlined in the comments.

src/boot/efi/boot.c
poettering: we already checked both args for non-NULL here, thus strcmp() is fine?

src/boot/efi/boot.c
poettering: well, not really. if one of the entries has version set and the other does not we don't want to compare by version. if we'd blindly call strverscmp_improved() it would return != in that case however. 

or in other words: if you compare one entry with and one without version we shouldn't use the version field at all, instead of ordering the one without the version before the one with the version

src/boot/efi/boot.c
poettering: as above: we should only compare by machine ID if both entries have one set. If one doesn't this is a reason to try something else, but not a reason to order the one without before the one with

src/boot/efi/boot.c
medhefgo: You'd imho still want entries with sort-key to go before entries without one. Which this fails to do.

src/boot/efi/boot.c
medhefgo: I'd argue the entry with version field should be sorted first as it's more specific.

Also, I don't like how this sorts version first. This could interleave undesirably like this: machine-1-version-2 < machine-2-version2 < machine-1-version-2

src/boot/efi/boot.c
poettering: I am not convinced that's desirable. I'd rather stick to the traditional sorting if we cannot compare by sort key.

src/boot/efi/boot.c
poettering: you are probably right, we should group by machine ID early on.

poettering: hmm, actually i don't follow. we sort by machined *before* version, precisely to have this order:

1. machine1 version 3
2. machine1 version 2
3. machine1 version 1
4. machine2 version 3
5. machine2 version 2
6. machine2 version 1

I'd argue that's exactly what we want.

medhefgo:
> Which specific reason do you mean? Can you elaborate?

Just re-read that PR. I don't feel like repeating myself.

> It doesn't? Why not? Arch can choose any sorting order they want, by carefully picking a sort-key. for example, they could use `sort-key archlinux-30` and `sort-key archlinux-70` or so in the entry they want to order explicitly against each other. sort-key gives them explicit control what should come first

That's not the point. Your PR does not sort entries the way people expect it to, and they shouldn't have to resort to using sort-key for this just because it fails to ignore the file ending when sorting. Even worse, sorting by sort-key vs without creates a list order that makes no sense (as I have explained above) and even confuses the default entry selection algorithm.

I think maybe the right approach is decoupling the entry sorting from default/priority selection...

src/boot/efi/boot.c
poettering: hmm, no? we sort:

1. if sort-key exists:
    1. by sort-key ascending
    2. by machine-id ascending
    3. by version descending
2. by id ascending

And that's exactly what we should do. And if we sorted things that way we should always search from beginning to end if we look for a default pattern or so.

Example:

1. sort-key=fedora version=35.1 machine-id=32dbce7a63f541c5b43d6265b713886f
2. sort-key=fedora version=35.2 machine-id=32dbce7a63f541c5b43d6265b713886f
3. sort-key=fedora version=36.1 machine-id=e3a1f5d8617c4ea7a66b560be6373b04
4. sort-key=fedora version=36.2 machine-id=e3a1f5d8617c4ea7a66b560be6373b04
5. sort-key=arch version=4711 machine-id=159064e98f7f4d38b577c2277c439bfa
6. sort-key=arch version=4712 machine-id=159064e98f7f4d38b577c2277c439bfa

This would result in the following order:

6, 5, 2, 1, 4, 3

And that looks absolutely appropriate and desired to me

src/boot/efi/boot.c
medhefgo: That create an "unstable" sorting order. Realistically, you cannot compare entries with to entries without. You're comparing apples vs oranges in that case, so ignoring the fact and resorting to comparing by color next makes no sense.

src/boot/efi/boot.c
medhefgo: You're right. I think the diff output just confused me

src/boot/efi/boot.c
medhefgo: And then you have entries that lack a sort-key, and versions will be sorted 1 < 2 < 3 (instead of 3 < 2 < 1). This is inconsistent and unexpected and in-fact, breaks creates a regression of version 3 being preferred.

src/boot/efi/boot.c
medhefgo: To put this in more concrete terms, your sorting function is violating transitivity.

Given this list:
```
1: id=a.conf sort-key=50
2: id=b.conf
3: id=c.conf sort-key=40
```

You have these realtions:
```
1 < 2 (id)
2 < 3 (id)
```

Which by transitivity must imply `1 < 3`, but it doesn't because `1 > 3 (sort-key) !!!`

This means that depending on the original order of the list, you can generate an unstable sorting output.

keszybz:
> This allows snippet generators to explicitly order entries: any string can be set as an entry's "sort key". If set, sd-boot will use it to sort entries on display.

> New logic is hence (ignore the boot counting logic)

>  sort-key is set → primary sort key: sort-key (lexicographically increasing order)
>                   → secondary sort key: machine-id (also increasing order)
>                   → tertiary sort key: version (lexicographically decreasing order!)

>   sort-key is not set → primary sort key: entry filename (aka id), lexicographically increasing order)

> With this scheme we can order OSes by their names from A-Z but then put within the same OS still the newest version first. This should clean up the order to match expectations more.

This doesn't sound right. If sort-key is the most important thing set, then it'd mean that distributions would need to somehow use different sort-key prefixes or entries from different distributions would interleave. And with parallel installations of the same distro things would be even worse. It seems that machine-id (in the sense of about-to-be-renamed KERNEL_INSTALL_MACHINE_ID) must be the primary sorting key.

As @medhefgo showed above, entries with and without sort-key cannot be compared. In particular, *adding* or *removing* an entry could cause some other entries to reverse order. The only solution that preserves transitivity is to sort entries without sort-key either as always earlier or later than those with sort-key. Since sort-key is "new", it seems reasonable to sort all such entries later.

Also having a sort-key order that is reverse of the version order sounds just terrible. The natural thing would be to generate entries with "version: someprefix-nnn-somepostfix, sort-key: maybeotherprefix-nnn". Let's just sort in alphanumerical order like with versions.

Sorry, but this seems to make the situation much worse.

poettering:
> This doesn't sound right. If sort-key is the most important thing set, then it'd mean that distributions would need to somehow use different sort-key prefixes or entries from different distributions would interleave.

Hmm? Yes that's the point, Fedora would set the sort key to "fedora", and Arch to "arch". 

> And with parallel installations of the same distro things would be even worse. It seems that machine-id (in the sense of about-to-be-renamed KERNEL_INSTALL_MACHINE_ID) must be the primary sorting key.

if you have parallel installations of the same distro, then they' have the same sort key, but different machine ID, so we'd group them by machine ID.

Not sure where the confusion is? How would *you* sort the stuff? I don#t follow at all anymore.

poettering:
> As @medhefgo showed above, entries with and without sort-key cannot be compared.

I cannot parse this. the patch follows the logic that either we have two "new-style" entries that both have a sort key set, in which case it is used for comparisons. Or one or both lack it, in which case we revert to the status quo ante, and just use the id. What#s wrong with that? I don#t follow what you are requesting instead?

keszybz:
> > This doesn't sound right. If sort-key is the most important thing set, then it'd mean that distributions would need to somehow use different sort-key prefixes or entries from different distributions would interleave.
> 
> Hmm? Yes that's the point, Fedora would set the sort key to "fedora", and Arch to "arch".

Oh, OK. So you want to use the sort-key to distinguish between e.g. "distros", and then the ~machine-id to distinguish between installations, and finally version to order different kernel versions for the same installation. OK, makes sense.

My comment about comparing versions with the key and without also can be ignored.

But the part about non-transitive sorting still stands.

poettering:
> Also having a sort-key order that is reverse of the version order sounds just terrible. The natural thing would be to generate entries with "version: someprefix-nnn-somepostfix, sort-key: maybeotherprefix-nnn". Let's just sort in alphanumerical order like with versions.

Can't parse this either. Are you saying you want Arch appear after or before Fedora in the list? And do you want that version 5 of Arch shall be placed after or before versin 6 of Arch? I assumed people want that Arch is before Fedora, And Arch 6 before ARch 5. But you want something else?

I know for a fact that I am confused now. I have the suspicion that I might not be the only one though?

poettering:
hmm, so you are concerned about cases where a triplet of entries with IDs X, Y, Z of which Y has no sort key but X and Z have B and A?

In that case if one compares Y with X and Z one would assume an order X, Y, Z but if one compares X with Z one would assume an order Z, X? And those are contradictory?

That indeed is a problem. Not sure what the right fix is for that though.

keszybz:
> I cannot parse this. the patch follows the logic that either we have two "new-style" entries that both have a sort key set, in which case it is used for comparisons. Or one or both lack it, in which case we revert to the status quo ante, and just use the id. What#s wrong with that? I don#t follow what you are requesting instead?

So imagine that we are transitioning to this scheme, and we have an entry
```
machine-id A
version 1
title A-1
```
and we add a new entry with
```
machine-id B
sort-key 11
title B-11
```
they will be sorted A, B-11.
And then we add
```
machine-id A
sort-key 12
version 2
title A-12
```
and they will be sorted A-12 < A-1 < B-11.
And then you remove A-1 on upgrade, and they are sorted B-11 < A-12.

I'm requesting instead that we sort anything with sort-key as higher priority then anything w/o. (On the assumption that entries with sort-key are newer.)

Thus B-11, A-12, A-1 first, and then B-11, A-12 when A-1 is removed.

keszybz:
I can't quite figure this out with the current code, but the general principle should be that in the boot menu, entries with later versions are sorted towards the top of the screen. And to keep user sanity, the output from 

With this patch I see the following:
```
$ build/bootctl list
Boot Loader Entries:
         type: Boot Loader Specification Type #1 (.conf)
        title: Fedora Linux 35 (Workstation Edition Prerelease) (default) (not reported/new)
           id: 08a5690a2eed47cf92ac0a5d2e3cf6b0-5.14.9-300.fc35.x86_64.conf
       source: /boot/efi/loader/entries/08a5690a2eed47cf92ac0a5d2e3cf6b0-5.14.9-300.fc35.x86_64.conf
      version: 5.14.9-300.fc35.x86_64
   machine-id: 08a5690a2eed47cf92ac0a5d2e3cf6b0
        linux: /08a5690a2eed47cf92ac0a5d2e3cf6b0/5.14.9-300.fc35.x86_64/linux
       initrd: /08a5690a2eed47cf92ac0a5d2e3cf6b0/5.14.9-300.fc35.x86_64/initrd
      options: ...

         type: Boot Loader Specification Type #1 (.conf)
        title: Fedora Linux 35 (Workstation Edition) (selected) (not reported/new)
           id: 08a5690a2eed47cf92ac0a5d2e3cf6b0-5.15.14-200.fc35.x86_64.conf
       source: /boot/efi/loader/entries/08a5690a2eed47cf92ac0a5d2e3cf6b0-5.15.14-200.fc35.x86_64.conf
      version: 5.15.14-200.fc35.x86_64
   machine-id: 08a5690a2eed47cf92ac0a5d2e3cf6b0
        linux: /08a5690a2eed47cf92ac0a5d2e3cf6b0/5.15.14-200.fc35.x86_64/linux
       initrd: /08a5690a2eed47cf92ac0a5d2e3cf6b0/5.15.14-200.fc35.x86_64/initrd
      options: ...
```
and when I boot with this sd-boot version, it also uses the same order, and selects the older version.

medhefgo:
> That indeed is a problem. Not sure what the right fix is for that though.

The way to make sure transitivity is guaranteed is to always compare a property and only fall back to the next property if they are *equal*. If only one has the property you simply (arbitrarily) define which one goes first (in this example, that's the job of the cmp function):

```
r = cmp(sort-key-a, sort-key-b);
if (r != 0)
    return r;

/* Both either have a key or both don't */

if (sort-key-a) {
    r = cmp(machine-id-a, machine-id-b);
    if (r != 0)
        return r;
    r = cmp(version-a, version-b);
    if (r != 0)
        return r;
}

/* Any further backup cmp. */
```

Also, I still don't like the whole approach as you either still have to maintain the "lower to higher" priority ordering or otherwise you can end up mixing the old stuff "older/broken first, newer last" vs the new stuff with a sort-key "newer first, older/broken last". And, btw, you this still doesn't solve the question of how we select the default entry by pattern: entries without sort-key want the last entry matching, those with one want the very first matching.

Btw, have I mention I hate sorting by filenames when it should be done by content (this applies to userspace too)?

keszybz:
> I hate sorting by filenames when it should be done by content (this applies to userspace too)?

FWIW, I'm with @poetting on this one. We shouldn't sort on content, because this just too fragile. Any time you append " (prerelease)" or " (final)" or change anything in the wording of the title you risk getting the wrong sort order.

keszybz:
> ```
> r = cmp(sort-key-a, sort-key-b);
> if (r != 0)
>     return r;
> 
> /* Both either have a key or both don't */
> 
> if (sort-key-a) {
>     r = cmp(machine-id-a, machine-id-b);
>     if (r != 0)
>         return r;
>     r = cmp(version-a, version-b);
>     if (r != 0)
>         return r;
> }
> 
> /* Any further backup cmp. */
> ```

Hmm, but why the `if`? 
This should just be:

```C
r = cmp(sort-key-a, sort-key-b);
if (r != 0)
     return r;
r = cmp(machine-id-a, machine-id-b);
if (r != 0)
    return r;
r = cmp(version-a, version-b);
if (r != 0)
   return r;
/* Etc. */
```

keszybz:
Hmm, did I say that somewhere?

medhefgo:
> FWIW, I'm with @poetting on this one..

> Hmm, did I say that somewhere?

Without that if you are comparing the *contents*, which you said you don't want.

poettering:
> Thins means that we'd compare two images with no sort key and no machine id by their version id, but that's hardly comparable...

So, previously the "version" field of entries where "local" to the OS, i.e. only was used to distinguish two entries of the *same* OS install. With the suggested comparison this would change, we'd suddenly order by that *before* sorting by OS. 

so, previously an entry with entry fname of "fedora-23.conf" that lacks both sort-key and machine id (but has a version 23 field), would be ordered after "arch-17.conf" that also lacks both sort-key and machine id (but has a version 17 field). But with your proposed series of comparisons, we'd suddenly compare them by version only, i.e. suddenly fedora-23.conf would be placed *before* fedora-17.conf, because it is newer.

So I think we should do what zbigniew originally suggested: put new-style entries that have a sort key before old style entries that do not. Yes, this will mean things are reordered during the transition where everyone adds sort key fields, but things are very systematic, and ordering is transitive, as requested.

medhefgo:
Also, what about users that don't use kernel-install and expect their ordering to be correct? They might even be using boot counting, but *that* will end up preferring/selecting a *bad* entry now if one exists.

poettering:
> Except it doesn't? As stated before A < B < C _and_ C < A:
> 
> ```
> A: id=a.conf sort-key=50
> B: id=b.conf
> C: id=c.conf sort-key=40
> ```

According to @keszybz' original suggestion we'd put those first that have a sort key, and those last that do not. Thus the order would be C, A, B. And regardless which pair of these you'll compare it will always match this order. i.e. A < B (because A has a sort key, and B does not), C < A (because both have a sort key and 40 < 50), C < B (because C has a sort key and B has not).

So it's fully transitive. All good!

poettering:
> Also, what about users that don't use kernel-install and expect their ordering to be correct? They might even be using boot counting, but _that_ will end up preferring/selecting a _bad_ entry now if one exists.

I don't follow at all: your patch switched things over to using the human readable title for sorting, which suddenly reordered everything for everyone, because a completely different string was now used for sorting that wasn't put together with sort order in mind at all? I mean any kind of sort order change isn't ideal. Your change isn't in any way better than this one. If you have a problem with changing hte sort order, then we should revert to the status quo ante. I don't follow anymore at all?

medhefgo:
> According to @keszybz' original suggestion we'd put those first that have a sort key, and those last that do not. Thus the order would be C, A, B. And regardless which pair of these you'll compare it will always match this order. i.e. A < B (because A has a sort key, and B does not), C < A (because both have a sort key and 40 < 50), C < B (because C has a sort key and B has not).
> 
> So it's fully transitive. All good!

I don't see that being done in your comparison function. The list will not be magically in this order or you wouldn't need to sort it in the first place.
(And regardless of whether it makes sense or not, you still need to also make sure the sorting within sort-key entries is transitive too, even if they lack a version or machine-id field.)

> I don't follow at all: your patch switched things over to using the human readable title for sorting, which suddenly reordered everything for everyone, because a completely different string was now used for sorting that wasn't put together with sort order in mind at all? I mean any kind of sort order change isn't ideal. Your change isn't in any way better than this one. If you have a problem with changing hte sort order, then we should revert to the status quo ante. I don't follow anymore at all?

True, I give you that. But your ordering is seems rather inconsistent to me (particularly regarding the default selection). And that  goes doubly so during a transition period.

poettering:
> I don't see that being done in your comparison function. The list will not be magically in this order or you wouldn't need to sort it in the first place.

No, not in the one I original pushed. But @keszybz original proposal was different to what I suggested. he suggested to systematically put entries that have a sort key before those which do not. I have not implemented that yet, I am just arguing, that that's the right approach, in my eyes, now. Will push a patch for that shortly.

poettering:
The new version includes a simple new test that checks if the order matches the expectations.

This does not fix #22580, which is unrelated. Fix for that one should be simple, but let's do that in another PR.

src/boot/efi/boot.c
```diff
@@ -1643,12 +1652,41 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
...
+                r = strcmp_ptr(a->machine_id, b->machine_id);
+                if (r != 0)
+                        return r;
+
+                /* If the sort key was defined, then order by version now (downwards, putting the newest first) */
+                r = -strverscmp_improved(a->version, b->version);
+                if (r != 0)
+                        return r;
+
+        } else if (b->sort_key) /* a is old-style but b new-style */
```
medhefgo: I don't like how this is expressed, putting this branch so far away. Keeping it similar to how the boot counting comparisons are done would be nicer, no?
```c
if (a && !b)
    return -1;
if (!a && b)
    return 1;
if (a && b)
    .....

**Comment by medhefgo** [comment]
I am still sorely missing ignoring the `.efi/.conf` part of the id. And it probably makes to ignore the boot counting part of the file name too for the same reason (and it's already being explicitly dealt with anyways).

**Comment by yuwata** [comment]
(This needs to be rebased.)

File: src/boot/efi/boot.c (hunk 6 above)
Comment (poettering) [reply src/boot/efi/boot.c]: sure

**Comment by poettering** [comment]
> I am still sorely missing ignoring the `.efi/.conf` part of the id. And it probably makes to ignore the boot counting part of the file name too for the same reason (and it's already being explicitly dealt with anyways).

Yeah the boot counting we definitely should drop, but I think that's a separate issue. I'd consider that a separate issue though. I'll add a TODO list item for it.

(That said, for entries with a sort-key removing the boot count from the id is not that important...)

**Comment by poettering** [comment]
Force pushed a new rebased version, with the compariso changes @medhefgo suggested. PTAL.

File: src/boot/efi/boot.c [hunk 7]
```diff
@@ -1643,12 +1652,39 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
...
         assert(b);
 
-        /* Order entries that have no tries left to the beginning of the list */
-        if (a->tries_left != 0 && b->tries_left == 0)
-                return 1;
+        /* Order entries that have no tries left to the end of the list */
         if (a->tries_left == 0 && b->tries_left != 0)
+                return 1;
+        if (a->tries_left != 0 && b->tries_left == 0)
                 return -1;
```
keszybz: Maybe
```
r = CMP(a->tries_left == 0, b->tries_left == 0);
if (r != 0)
      return r;
```

keszybz (commented):
Looks reasonable.

I think that sorting towards lower versions (for newer entries with the sort-key) and then towards higher versions (for older entries) is annoying, but is in some way logical. I *think* we can fix it, but let's consider this is a separate issue later on. We can consider that this PR adds a new type of entries which need to be sorted separately and earlier and with lower versions later. For older entries the behaviour stays at it was, at least for now.

I'll give this a spin on my laptop.

src/boot/efi/boot.c
poettering: didn't want to touch that stuff from the status quo needlessly. But yeah, we should use CMP() for these too, eventually.

keszybz:
focal-arm64 failed in testbed setup,
and focal-ppc64el failed with `TEST-50-DISSECT:                    FAIL     (610 s)`.

PR #22412, opened 2022-02-04
sd-dhcp6-client: several fixes and cleanups

yuwata:

PR #22520, opened 2022-02-15
sd-dhcp6-client: several cosmetic follow-ups

yuwata:
Follow-ups for #22412.

### DIFF
--- a/src/libsystemd-network/dhcp6-option.c
+++ b/src/libsystemd-network/dhcp6-option.c
@@ -236,9 +236,7 @@ int dhcp6_option_append(uint8_t **buf, size_t *buflen, uint16_t code,
         if (r < 0)
                 return r;
 
-        memcpy_safe(*buf, optval, optlen);
-
-        *buf += optlen;
+        *buf = mempcpy_safe(*buf, optval, optlen);
         *buflen -= optlen;
 
         return 0;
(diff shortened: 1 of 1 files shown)


src/libsystemd-network/dhcp-identifier.c
```diff
@@ -168,7 +168,7 @@ static int dhcp_identifier_set_duid_uuid(struct duid *ret_duid, size_t *ret_len)
         unaligned_write_be16(&ret_duid->type, DUID_TYPE_UUID);
         memcpy(ret_duid->raw.data, &machine_id, sizeof(machine_id));
 
-        *ret_len = sizeof(ret_duid->type) + sizeof(machine_id);
+        *ret_len = offsetof(struct duid, uuid) + sizeof(machine_id);
```
poettering: hmme, the memcpy above writes to `ret_duid->raw.data` but here you reference `struct duid`'s `uuid` field instead. It sould point to the same place, but I guess it would be better to fix the memcpy() to use `ret_duid->uuuid.uuid` instead.

yuwata:
@poettering Thank you for the review. Addressed the above point. Upgrading the green label.

keszybz:
```
../src/libsystemd-network/dhcp-identifier.c: In function â€˜dhcp_identifier_set_duid_uuidâ€™:
../src/libsystemd-network/dhcp-identifier.c:169:30: error: incompatible type for argument 1 of â€˜memcpyâ€™
  169 |         memcpy(ret_duid->uuid.uuid, &machine_id, sizeof(machine_id));
      |                ~~~~~~~~~~~~~~^~~~~
      |                              |
      |                              sd_id128_t {aka union sd_id128}
In file included from /usr/include/features.h:490,
                 from /usr/include/sys/types.h:25,
                 from /usr/include/net/ethernet.h:24,
                 from ../src/libsystemd-network/dhcp-identifier.c:4:
/usr/include/bits/string_fortified.h:26:1: note: expected â€˜void * restrictâ€™ but argument is of type â€˜sd_id128_tâ€™ {aka â€˜union sd_id128â€™}
   26 | __NTH (memcpy (void *__restrict __dest, const void *__restrict __src,
      | ^~~~~
```

yuwata:
Ouch. I have no idea why build succeeds on my laptop... Fixed.

PR #22522, opened 2022-02-15
logind: load scheduled shutdown on restart 

lnussel:
Make sure logind doesn't forget about a scheduled shutdown, even if it
gets restarted eg due to a package update.

### DIFF
--- a/src/systemctl/systemctl-logind.c
+++ b/src/systemctl/systemctl-logind.c
@@ -304,24 +304,9 @@ int logind_schedule_shutdown(void) {
         if (r < 0)
                 return r;
 
-        switch (arg_action) {
-        case ACTION_HALT:
-                action = "halt";
-                break;
-        case ACTION_POWEROFF:
-                action = "poweroff";
-                break;
-        case ACTION_KEXEC:
-                action = "kexec";
-                break;
-        case ACTION_EXIT:
-                action = "exit";
-                break;
-        case ACTION_REBOOT:
-        default:
-                action = "reboot";
-                break;
-        }
+        action = action_table[arg_action].verb;
+        if (!action)
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "Scheduling not supported for this action.");
 
         if (arg_dry_run)
                 action = strjoina("dry-", action);


src/login/logind-dbus.c
```diff
@@ -63,7 +64,10 @@
  */
 #define WALL_MESSAGE_MAX 4096
 
+#define SHUTDOWN_SCHEDULE_FILE "/run/systemd/shutdown/scheduled"
```
bluca, line 69: I know this is a general pattern, but what about trying something new-ish and a bit nicer: serialize to memfd, hand it over to the FD store when shutting down, and restoring from there on startup?

src/login/logind-dbus.c
```diff
@@ -2038,6 +2042,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
 
+int manager_load_scheduled_shutdown(Manager *m) {
+        _cleanup_fclose_ FILE *f = NULL;
+        int r;
+
+        f = fopen(SHUTDOWN_SCHEDULE_FILE, "r");
+        if (!f) {
+                if (errno == ENOENT)
+                        return 0;
+                return -errno;
```
anitazha: style suggestion: `return errno == ENOENT ? 0 : -errno;`
this would also allow you to remove the curly braces

src/login/logind-dbus.c
```diff
@@ -2038,6 +2042,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                                log_notice("Failed to parse scheduled shutdown type '%s', ignoring.", val);
+                        else
+                                m->scheduled_shutdown_type = manager_item_for_handle(handle);
+                } else if ((val = startswith(l, "WALL_MESSAGE="))) {
+                        _cleanup_free_ char *unescaped = NULL;
+                        r = cunescape(val, 0, &unescaped);
+                        if (r >= 0)
+                                r = free_and_strdup(&m->wall_message, empty_to_null(unescaped));
+                        if (r < 0)
+                                return log_oom();
```
anitazha: style suggestion:
```
if (r < 0)
        return log_oom();

r = free_and_strdup(&m->wall_message, empty_to_null(unescaped));
```

src/login/logind-dbus.c
```diff
@@ -2038,6 +2042,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                r = manager_setup_shutdown_timers(m);
+                if (r < 0)
+                        return r;
+
+                r = manager_setup_wall_message_timer(m);
+                if (r < 0)
+                        return r;
+        } else {
+                reset_scheduled_shutdown(m);
+        }
```
anitazha: no curly braces around this `else` block

src/login/logind-dbus.c
```diff
@@ -2038,6 +2042,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+                r = read_line(f, LONG_LINE_MAX, &line);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to read serialization line: %m");
+                if (r == 0)
+                        break;
+
+                l = strstrip(line);
+                if (isempty(l)) /* end marker */
+                        break;
```
yuwata: Rather than `break;`, I think it is better to `continue;` the loop. As it may contains blank lines.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2042,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                                log_notice("Failed to parse enable wall messages '%s', ignoring.", val);
+                } else if ((val = startswith(l, "MODE="))) {
+                        HandleAction handle = handle_action_from_string(val);
+                        if (handle == _HANDLE_ACTION_INVALID)
+                                log_notice("Failed to parse scheduled shutdown type '%s', ignoring.", val);
+                        else
+                                m->scheduled_shutdown_type = manager_item_for_handle(handle);
+                } else if ((val = startswith(l, "WALL_MESSAGE="))) {
+                        _cleanup_free_ char *unescaped = NULL;
+                        r = cunescape(val, 0, &unescaped);
```
yuwata: Please add a blank line after the variable declaration.

src/login/logind-dbus.c
yuwata: `free_and_strdup_warn()` can be used.

src/login/logind-dbus.c
```diff
@@ -2046,41 +2112,31 @@ static int update_schedule_file(Manager *m) {
...
+        serialize_item_format(f, "WARN_WALL", "%u", m->enable_wall_messages);
+        serialize_item_format(f, "MODE", "%s", handle_action_to_string(m->scheduled_shutdown_type->handle));
 
         if (!isempty(m->wall_message)) {
-                _cleanup_free_ char *t = NULL;
-
-                t = cescape(m->wall_message);
-                if (!t) {
-                        r = -ENOMEM;
+                r =serialize_item_escaped(f, "WALL_MESSAGE", m->wall_message);
```
yuwata: missing white space after `=`.

src/login/logind-dbus.c
```diff
@@ -2210,35 +2266,63 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
 
-        if (m->nologin_timeout_source) {
-                r = sd_event_source_set_time(m->nologin_timeout_source, nologin_timeout_usec(elapse));
+        r = manager_setup_shutdown_timers(m);
+        if (r < 0)
+                return r;
+
+        r = setup_wall_message_timer(m, message);
+        if (r < 0) {
+                m->scheduled_shutdown_timeout_source = sd_event_source_unref(m->scheduled_shutdown_timeout_source);
```
yuwata: `sd_event_source_disable_unref()`

Is it ok to leave `nologin_timeout_source` as is?

src/login/logind-dbus.c
```diff
@@ -2210,35 +2266,63 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
+                return r;
+
+        return sd_bus_reply_method_return(message, NULL);
+}
+
+static int manager_setup_shutdown_timers(Manager* m) {
+        int r;
+
+        if (m->scheduled_shutdown_timeout_source) {
+                r = sd_event_source_set_time(m->scheduled_shutdown_timeout_source,
```
yuwata: You can use `event_reset_time()`.

src/login/logind-dbus.c
```diff
@@ -2210,35 +2266,63 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
+                r = sd_event_add_time(m->event, &m->scheduled_shutdown_timeout_source,
+                                      CLOCK_REALTIME, m->scheduled_shutdown_timeout, 0,
+                                      manager_scheduled_shutdown_handler, m);
                 if (r < 0)
                         return log_error_errno(r, "sd_event_add_time() failed: %m");
         }
 
-        m->scheduled_shutdown_timeout = elapse;
+        if (m->nologin_timeout_source) {
+                r = sd_event_source_set_time(m->nologin_timeout_source,
```
yuwata: Same here. `event_reset_time()`.

yuwata (changes requested):
LGTM. Several style nits.

src/login/logind-dbus.c
lnussel: In this case logind already writes the file. Was added in 867c37f6bb. It just never uses it for anything. As the format is compatible with what the `(de)serialize_*` functions do I thought might be easiest to just reuse the file.

yuwata: Yeah. Let's keep this as is at least in this PR.

src/login/logind-dbus.c
lnussel: doesn't really matter I guess. logind itself would not produce empty lines creating the file

src/login/logind-dbus.c
lnussel: Would like this then:
```
r = cunescape(val, 0, &unescaped);                                                    
if (r < 0)                                                                            
        return log_oom();                                                             
                                                                                      
r = free_and_strdup_warn(&m->wall_message, empty_to_null(unescaped));                 
if (r < 0)                                                                            
        return r;
```

We miss cunescape to static buffer :)

src/login/logind-dbus.c
bluca: Ok, let's leave it as is then

src/login/logind-dbus.c
lnussel: was that way before but probably not :-) will fix

src/login/logind-dbus.c
lnussel: Nice, that make code much shorter. What accuracy and priority do we need in that case? looks like existing users use 10 * USEC_PER_MSEC, USEC_PER_SEC or 0 as accuracy. Priority seems to be 0 always?

At this point I wonder why there is no inline documentation in headers at all :-)

lnussel: ah, 0 is DEFAULT_ACCURACY_USEC and was used before.

src/systemctl/systemctl-compat-halt.c
```diff
@@ -76,6 +76,8 @@ int halt_parse_argv(int argc, char *argv[]) {
         assert(argc >= 0);
         assert(argv);
 
+        /* called in sysvinit system as last command in
+         * shutdown/reboot so this is always forceful */
```
poettering: btw, we usually break lines at 109 chars, see CODING_STYLE, .editorconfig and so on.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2043,74 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                        if (r < 0)
+                                return r;
+                } else if ((val = startswith(l, "UID="))) {
+                        if (parse_uid(val, &m->scheduled_shutdown_uid) < 0)
+                                log_notice("Failed to parse UID '%s', ignoring.", val);
+                } else if ((val = startswith(l, "TTY="))) {
+                        r = free_and_strdup_warn(&m->scheduled_shutdown_tty, empty_to_null(val));
+                        if (r < 0)
+                                return r;
+                }
```
poettering: tis looks like an environment file, i.e. use parse_env_file() for this?

src/login/logind-dbus.c
```diff
@@ -63,7 +65,10 @@
  */
 #define WALL_MESSAGE_MAX 4096
 
+#define SHUTDOWN_SCHEDULE_FILE "/run/systemd/shutdown/scheduled"
```
poettering, line 69: i'd probably not add the define for this. if the the value is arbitrary or hard to type or changable at build time, sure add a macro for it, but I think it's much easier to read if you just literally spell out the path name at the two places you need it.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2043,74 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
         return LESS_BY(elapse, 5 * USEC_PER_MINUTE);
 }
 
+int manager_load_scheduled_shutdown(Manager *m) {
+        _cleanup_fclose_ FILE *f = NULL;
+        int r;
+
+        f = fopen(SHUTDOWN_SCHEDULE_FILE, "r");
```
poettering: as per coding style: "re"

src/login/logind-dbus.c
```diff
@@ -2038,6 +2043,74 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                        r = cunescape(val, 0, &unescaped);
+                        if (r < 0)
+                                return log_oom();
+
+                        r = free_and_strdup_warn(&m->wall_message, empty_to_null(unescaped));
+                        if (r < 0)
+                                return r;
+                } else if ((val = startswith(l, "UID="))) {
+                        if (parse_uid(val, &m->scheduled_shutdown_uid) < 0)
+                                log_notice("Failed to parse UID '%s', ignoring.", val);
```
poettering: so, we used to warn about parsing errors during serialization/deserialization at LOG_WARNING level. But we downgraded those all to LOG_DEBUG. These messages are relevant for developers after all, not users, and usually the only case this happens is during upgrades from slightly incompatible versions. Hence please always use log_debug()/log_debug_errno() for logging about deserialization issues here.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2043,74 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                if (r == 0)
+                        break;
+
+                l = strstrip(line);
+                if (isempty(l))
+                        continue;
+
+                if ((val = startswith(l, "USEC="))) {
+                        if (deserialize_usec(val, &m->scheduled_shutdown_timeout) < 0)
+                                log_notice("Failed to parse scheduled shutdown time '%s', ignoring.", val);
```
poettering: deserialize_usec() already logs, no need to do that here again

src/login/logind-dbus.c
```diff
@@ -2038,6 +2043,74 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                } else if ((val = startswith(l, "WARN_WALL="))) {
+                        if (safe_atou(val, &m->enable_wall_messages) < 0)
+                                log_notice("Failed to parse enable wall messages '%s', ignoring.", val);
```
poettering: log_debug_errno(r, …)

src/login/logind-dbus.c
```diff
@@ -2038,6 +2043,74 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                } else if ((val = startswith(l, "MODE="))) {
+                        HandleAction handle = handle_action_from_string(val);
+                        if (handle == _HANDLE_ACTION_INVALID)
+                                log_notice("Failed to parse scheduled shutdown type '%s', ignoring.", val);
```
poettering: log_debug()

src/login/logind-dbus.c
```diff
@@ -2046,41 +2119,35 @@ static int update_schedule_file(Manager *m) {
...
 
-        fprintf(f,
-                "USEC="USEC_FMT"\n"
-                "WARN_WALL=%i\n"
-                "MODE=%s\n",
-                m->scheduled_shutdown_timeout,
-                m->enable_wall_messages,
-                handle_action_to_string(m->scheduled_shutdown_type->handle));
+        serialize_usec(f, "USEC", m->scheduled_shutdown_timeout);
+        serialize_item_format(f, "WARN_WALL", "%u", m->enable_wall_messages);
```
poettering: maybe serialize as `%s` and `zero_one(m->enable_wall_messages)`, to make explicit this is a boolean.

src/login/logind-dbus.c
```diff
@@ -2193,54 +2260,45 @@ static int method_schedule_shutdown(sd_bus_message *message, void *userdata, sd_
...
+static int manager_setup_shutdown_timers(Manager* m) {
+        int r;
+
+        r = event_reset_time(m->event, &m->scheduled_shutdown_timeout_source,
+                        CLOCK_REALTIME,
+                        m->scheduled_shutdown_timeout, 0,
+                        manager_scheduled_shutdown_handler, m,
+                        0, "scheduled-shutdown-timeout", true);
+
+        if (r >= 0)
```
poettering: our usual coding style suggests to prefer early-exit over deep-nesting. i.e. always do this:

```c
…
if (r < 0)
        return r;
…
```

over this:

```c
…
if (r >= 0) {
        …
}
…
return …;
```
or in other words: indenting sucks, let's minimize it.

src/login/logind-utmp.c
```diff
@@ -147,20 +148,14 @@ int manager_setup_wall_message_timer(Manager *m) {
...
-                if (r < 0)
-                        return log_error_errno(r, "sd_event_source_set_enabled() failed. %m");
-        } else {
-                r = sd_event_add_time(m->event, &m->wall_message_timeout_source,
-                                      CLOCK_REALTIME, n + elapse, 0, wall_message_timeout_handler, m);
-                if (r < 0)
-                        return log_error_errno(r, "sd_event_add_time() failed. %m");
-        }
+        if (r < 0)
+                return log_error_errno(r, "failed to set up wall message timer: %m");
```
poettering: We usually start our log messages with an uppercase character, as it is customary in English language sentences.

src/login/logind.c
```diff
@@ -1090,6 +1090,10 @@ static int manager_startup(Manager *m) {
         if (r < 0)
                 log_warning_errno(r, "Button enumeration failed: %m");
 
+        r = manager_load_scheduled_shutdown(m);
+        if (r < 0)
+                log_warning_errno(r, "Load scheduled shutdown failed: %m");
```
poettering: for new code we try to be strict in the rules who logs. i.e. for a function figure out whether it shall log about all errors, or about none (usually "library"-like functions should not log on their own and leave that to the caller, while "main-program"-like functions should log on their own, but the line between the two is blurred). Then stick to it, and then either let the caller log about it (in case the function is non-logging) or do not log in the caller (in case it logs on its own).

manager_load_scheduled_shutdown() logs about most errors already, and it's probably more "main-program-like", hence it should log about all. Because of that this log line here in the invocation should just go away.

Following this rule ensures we log about everything there is to log about exactly once — and not twice or not at all.

We didn't always follow this rule, so you'll find place where we are sloppy on this (in particular in logind there are plenty), but for new code please always follow this rule.

See: https://systemd.io/CODING_STYLE/#logging

poettering (changes requested):

src/login/logind-dbus.c
lnussel: ah, ok. didn't know about that one. have to take a look

src/login/logind-dbus.c
lnussel: Ic, that explains the repetition of other file names in the file. If it was only two I probably would not have bothered either, it's 6 though :-) Also in this particular case makes vim `*` search to jump to references easier as the other words in the file name appear in other context too :-)

yuwata:
#22543 is merged. Please rebase.

lnussel:
overhaul using `parse_env_file()` trying to balance no indent vs no goto :)

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+        _cleanup_fclose_ FILE *f = NULL;
+        _cleanup_free_ char *usec = NULL,
+               *warn_wall = NULL,
+               *mode = NULL,
+               *wall_message = NULL,
+               *uid = NULL,
+               *tty = NULL;
+        _cleanup_free_ char *unescaped = NULL;
+        int r;
+
```
bluca, line 2057: `assert(m);`

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                r = -EINVAL;
+                goto out;
+        }
+
+        /* remaining settings are not crucial, so ignore errors */
+
+        if (warn_wall && parse_boolean(warn_wall) >= 0)
+                m->enable_wall_messages = r;
+
+        if(wall_message && cunescape(wall_message, 0, &unescaped) >= 0)
```
bluca: style nit: missing space after `if`

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+        r = deserialize_usec(strnull(usec), &m->scheduled_shutdown_timeout);
+        if (r < 0)
+                goto out; /* without proper time no point handling the rest */
+
+        HandleAction handle = handle_action_from_string(mode);
+        if (handle != _HANDLE_ACTION_INVALID)
+                m->scheduled_shutdown_type = manager_item_for_handle(handle);
+        else {
+                log_debug("Failed to parse scheduled shutdown type '%s', ignoring.", mode);
```
poettering:
```c
r = log_debug_errno(SYNTHETIC_ERRNO(EINVAL), …)
```

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                m->scheduled_shutdown_timeout = 0;
+                r = -EINVAL;
+                goto out;
+        }
+
+        /* remaining settings are not crucial, so ignore errors */
+
+        if (warn_wall && parse_boolean(warn_wall) >= 0)
```
poettering: these days we try hard not to ignore errors entirely, but do a debug log about them whenever something fails, i.e. change this to:

```c
if (warn_wall) {
        r = parse_boolean(warn_wall);
        if (r < 0) {
                log_debug_errno(r, "Failed to parse xyz field: %m");
                goto finish;
        }

File: src/login/logind-dbus.c [hunk 27]
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                m->enable_wall_messages = r;
```
poettering: r doesn't actually contain anything useful here

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+        if(wall_message && cunescape(wall_message, 0, &unescaped) >= 0)
+                free_and_strdup_warn(&m->wall_message, empty_to_null(unescaped));
+
+        if (uid)
+                parse_uid(uid, &m->scheduled_shutdown_uid);
```
poettering: as above...

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,68 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+        free_and_strdup_warn(&m->scheduled_shutdown_tty, empty_to_null(tty));
+
+        r = manager_setup_shutdown_timers(m);
+        if (r >= 0)
+                (void) manager_setup_wall_message_timer(m);
+
+out:
+        if (r < 0)
+                (void) unlink(SHUTDOWN_SCHEDULE_FILE);
+        return r;
```
poettering: we'd usually do this like this:

```c
        …
        return 0;

fail:
        (void) unlink(…);
        return r;
}
```

i.e. do the label only for the failure path, so that we don't have to check `r` again, and have one indentation less.

src/login/logind-dbus.c
lnussel: ok if we bail out at those places I'd reset the scheduled shutdown completely in the fail path. So we don't half apply the file.

src/login/logind-dbus.c
lnussel: :facepalm:

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                        "UID", &uid,
+                        "TTY", &tty);
+        if (r == -ENOENT)
+                return 0;
+        if (r < 0) {
+                log_debug_errno(r, "Failed to parse " SHUTDOWN_SCHEDULE_FILE ": %m");
+                goto fail;
+        }
+
+        r = deserialize_usec(strnull(usec), &m->scheduled_shutdown_timeout);
```
poettering: that's mean, if `usec` is actually NULL you tell it to deserialize the string `"(null)"`... 

I'd rather do this:

```c
if (usec) {
        r = deserialize_usec(…);
        if (r < 0)
                goto fail;
} else
        m->scheduled_shutdown_timeout = 0;
```

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+        if (r < 0)
+                goto fail;
```
poettering: so, given we deserialize stuff we ourselves wrote, i.e. our internal stuff, I think we should handle failures as graceful as we can. i.e. debug log about it, but continue, and nor exit the function early. hence, here, just continue instead of going to fail.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+        HandleAction handle = handle_action_from_string(mode);
+        if (handle == _HANDLE_ACTION_INVALID) {
```
poettering: we usually check for validity of enums with a simple `handle < 0`. i.e. the thinking here is that evertyhing < 0 is invalid/bad/failure, and everything >= 0 is valid/good/success

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Failed to parse scheduled shutdown type '%s'", mode);
```
poettering: log_debug_errno() returns the (negative) error number you pass in (but strips the effect of SYNTHETIC_ERRNO(). Hence you can merge this line and the next into:

```c
r = log_debug_errno(SYNTHETIC_ERRNO(EINVAL), …);
```

poettering: (but as mentioned, I think we should continue with parsing fields here, to handle failures of parsing our own internal serialization as gracefully as possible)

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+        }
+        m->scheduled_shutdown_type = manager_item_for_handle(handle);
+
+        if (wall_message) {
+                r = cunescape(wall_message, 0, &unescaped);
+                if (r < 0) {
+                        log_debug_errno(r, "Failed to parse wall message");
+                        goto fail;
+                }
+                if (free_and_strdup_warn(&m->wall_message, empty_to_null(unescaped)) < 0)
```
poettering: no need to strdup here again, just use free_and_replace()

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
 
+int manager_load_scheduled_shutdown(Manager *m) {
+        _cleanup_fclose_ FILE *f = NULL;
+        _cleanup_free_ char *usec = NULL,
+               *warn_wall = NULL,
+               *mode = NULL,
+               *wall_message = NULL,
+               *uid = NULL,
+               *tty = NULL;
+        _cleanup_free_ char *unescaped = NULL;
```
poettering: move to innermost scope

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Failed to parse scheduled shutdown type '%s'", mode);
+                r = -EINVAL;
+                goto fail;
+        }
+        m->scheduled_shutdown_type = manager_item_for_handle(handle);
+
+        if (wall_message) {
+                r = cunescape(wall_message, 0, &unescaped);
+                if (r < 0) {
+                        log_debug_errno(r, "Failed to parse wall message");
```
poettering: suffix with `: %s", wall_message`

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                        goto fail;
+                }
+                if (free_and_strdup_warn(&m->wall_message, empty_to_null(unescaped)) < 0)
+                        goto fail;
+        }
+
+        if (uid) {
+                r = parse_uid(uid, &m->scheduled_shutdown_uid);
+                if (r < 0) {
+                        log_debug_errno(r, "Failed to parse wall uid");
```
poettering: as above, suffix with `: %s", uid`, i.e. always output the string we can't parse

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+        }
+
+        if (free_and_strdup_warn(&m->scheduled_shutdown_tty, empty_to_null(tty)) < 0)
+                goto fail;
+
+        /* do this last as we'd have to reset on fail and we do not want to repeat any built in default */
+        if (warn_wall) {
+                r = parse_boolean(warn_wall);
+                if (r < 0) {
+                        log_debug_errno(r, "Failed to parse enabling wall messages");
```
poettering: similar

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+fail:
+        m->scheduled_shutdown_timeout = 0;
+        m->scheduled_shutdown_type = NULL;
+        m->scheduled_shutdown_uid = 0;
+        m->wall_message = mfree(m->wall_message);
+        m->scheduled_shutdown_tty = mfree(m->scheduled_shutdown_tty);
+        (void) unlink(SHUTDOWN_SCHEDULE_FILE);
+        return r;
+}
```
poettering, line 2121: as mentioned I think the failure logic is "too fatal" here. i.e. if we encounter one field or one line we can't make sense of then log about it at debug level but continue otherwise. This shuld be a graceful deserialization: take what we can, but don't be too drastic in our reaction if we can't get something.

lnussel:
Updated. So far I left SHUTDOWN_SCHEDULE_FILE in. Is that ok is there a strong desire to spell it out everywhere?

lnussel:
up, collision. will look again

src/login/logind-dbus.c
lnussel: yeah, that's what I did before https://github.com/systemd/systemd/pull/22522#pullrequestreview-887831109

IMO ignoring an invalid shutdown time and action doesn't make sense though as the other values are of no use then.

src/login/logind-dbus.c
lnussel: An alternative would be to make `deserialize_usec` return error on NULL instead of hitting assert.

src/login/logind-dbus.c
poettering: that's a valid point i admit. but it seems some fields are more auxiliary than others, and those should not cause a fatal parsing error.

poettering:
Cherry picked the first commit into git main now. Seems simple enough and unrelated enough to the rest. Hope that's OK.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                m->enable_wall_messages = r;
+        }
+
+        r = manager_setup_shutdown_timers(m);
+        if (r < 0)
+                goto fail;
+
+        r = manager_setup_wall_message_timer(m);
+
+        return r;
```
poettering: i'd merge these two lines, i.e. `return manager_setup_wall_message_timer(m);`

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+fail:
+        m->scheduled_shutdown_timeout = 0;
```
poettering: USEC_INFINITY?

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,94 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+        m->scheduled_shutdown_type = NULL;
+        m->scheduled_shutdown_uid = 0;
```
poettering: UID_INVALID?

src/login/logind.c
```diff
@@ -1090,6 +1090,8 @@ static int manager_startup(Manager *m) {
         if (r < 0)
                 log_warning_errno(r, "Button enumeration failed: %m");
 
+        manager_load_scheduled_shutdown(m);
```
poettering, line 1093: maybe cast to `(void)` explicitly, given you knowingly ignore an error return here.

poettering: (static analyzers tend to check for stuff like that)

src/login/logind-dbus.c
lnussel: those scheduled_* settings are not separately initialized (see https://github.com/systemd/systemd/blob/c972880640ee19e89ce9265d8eae1b3aae190332/src/login/logind.c#L44), so the uninitialized values are zero.

src/login/logind.c
lnussel: I'll make it return void, it's the only user here after all

src/login/logind-dbus.c
lnussel: `reset_scheduled_shutdown()` could probably do that initialization. Then we call it on startup and in the fail case

src/login/logind-dbus.c
lnussel: dito

src/systemctl/systemctl-logind.c
```diff
@@ -304,24 +304,9 @@ int logind_schedule_shutdown(void) {
...
-                break;
-        case ACTION_EXIT:
-                action = "exit";
-                break;
-        case ACTION_REBOOT:
-        default:
-                action = "reboot";
-                break;
-        }
+        action = action_table[arg_action].verb;
```
poettering, line 307: why not: manager_item_for_handle(arg_action)?

lnussel: It's a different table. When refactoring logind I didn't realize that someting called `action_table` already existed in systemctl. We probably should rename the one in logind to avoid confusion.

lnussel:
updated. have to rebase after #22595 though

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,87 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+                        "TTY", &tty);
+
+        /* reset will delete the file */
+        reset_scheduled_shutdown(m);
+
+        if (r < 0) {
+                if (r != -ENOENT)
+                        log_debug_errno(r, "Failed to parse " SHUTDOWN_SCHEDULE_FILE ": %m");
+                return;
+        }
```
keszybz:
```C
if (r == -ENOENT)
     return;
if (r < 0)
     return (void) log_debug_errno(r, "Failed to parse " SHUTDOWN_SCHEDULE_FILE ": %m");
```

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,87 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+        HandleAction handle = handle_action_from_string(mode);
+        if (handle < 0) {
+                log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Failed to parse scheduled shutdown type: %s", mode);
+                return;
+        }
```
keszybz:
```C
if (handle < 0)
    return (void) log_debug_errno(handle, "Failed to parse scheduled shutdown type: %s", mode);
```

And similarly below…

Or you could make this function return `int` and add `(void)` in the caller.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,87 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+        if (!usec) {
+                log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "USEC is required");
+                return;
+        }
+        if(deserialize_usec(usec, &m->scheduled_shutdown_timeout) < 0)
```
keszybz: Space after `if` is missing. And there is no debug message here.

src/login/logind-dbus.c
```diff
@@ -2038,6 +2044,87 @@ static usec_t nologin_timeout_usec(usec_t elapse) {
...
+
+        r = manager_setup_shutdown_timers(m);
+        if (r < 0)
+                reset_scheduled_shutdown(m);
+        else {
+                (void) manager_setup_wall_message_timer(m);
+                (void) update_schedule_file(m);
+        }
+
+        return;
```
keszybz:
```C
if (manager_setup_shutdown_timers(m) < 0)
     return (void) reset_scheduled_shutdown(m);

(void) manager_setup_wall_message_timer(m);
(void) update_schedule_file(m);
```

src/login/logind-utmp.c
```diff
@@ -147,20 +148,16 @@ int manager_setup_wall_message_timer(Manager *m) {
...
-                r = sd_event_source_set_enabled(m->wall_message_timeout_source, SD_EVENT_ONESHOT);
-                if (r < 0)
-                        return log_error_errno(r, "sd_event_source_set_enabled() failed. %m");
-        } else {
-                r = sd_event_add_time(m->event, &m->wall_message_timeout_source,
-                                      CLOCK_REALTIME, n + elapse, 0, wall_message_timeout_handler, m);
-                if (r < 0)
-                        return log_error_errno(r, "sd_event_add_time() failed. %m");
-        }
+        if (r >= 0)
```
keszybz: Please remove the empty newline `if (r >= 0)`.
It doesn't matter greatly, but we prefer to handle errors in the `if`, and continue on success.
Thus this should rather be
```C
if (r < 0) {
    m->wall_message_timeout_source = sd_event_source_unref(m->wall_message_timeout_source);
    return log_error_errno(r, "Failed to set up wall message timer: %m");
}

return 0;
```

src/systemctl/systemctl-logind.c
```diff
@@ -304,24 +304,9 @@ int logind_schedule_shutdown(void) {
...
+        if (!action)
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "Action not supported for scheduling");
```
keszybz: "Scheduling not supported for %s.", arg_action

src/login/logind-action.h
```diff
@@ -21,7 +21,7 @@ typedef enum HandleAction {
 
 typedef struct ActionTableItem ActionTableItem;
 
-#define handle_action_valid(x) (x && (x < _HANDLE_ACTION_MAX))
+#define handle_action_valid(x) (x >= 0 && (x < _HANDLE_ACTION_MAX))
```
keszybz: Please drop the nested parens in the right operand… It doesn't make sense to have them on just one operand.

src/login/logind-button.c
```diff
@@ -84,7 +84,7 @@ static void button_lid_switch_handle_action(Manager *manager, bool is_edge) {
          * differently */
         if (manager_is_docked_or_external_displays(manager))
                 handle_action = manager->handle_lid_switch_docked;
-        else if (!handle_action_valid(manager->handle_lid_switch_ep) && manager_is_on_external_power())
+        else if (handle_action_valid(manager->handle_lid_switch_ep) && manager_is_on_external_power())
```
keszybz, line 87: Ooops.

keszybz (changes requested):
More style tweaks required, sorry.

src/login/logind-utmp.c
keszybz: (But the other version is OK too. Feel free to ignore this comment.)

src/login/logind-dbus.c
lnussel: Damnit. Those f*** spaces. How do I beat vim into telling me beforehand to avoid the embarrassment?  :smile:

lnussel: Anyway `deserialize_usec()` logs itself. Kind of inconsistent but with the current code no logging needed/desired here.

src/systemctl/systemctl-logind.c
lnussel: arg_action is an integer, so doesn't really work like that :-) Could even assert here as it would be some unexpected code path.

src/login/logind-dbus.c
keszybz: Oh, OK.

keszybz: Space?

src/systemctl/systemctl-logind.c
keszybz: OK, but let's write something with better grammar.
What about "Scheduling not supported for this action." ?

keszybz (changes requested):
Almost there.

bluca:
This doesn't look related, but it's also weird:

```
Feb 24 11:10:34 H systemd[30]: testsuite-69.service: Failed to locate executable /usr/lib/systemd/tests/testdata/units/testsuite-69.sh: No such file or directory
Feb 24 11:10:34 H systemd[30]: testsuite-69.service: Failed at step EXEC spawning /usr/lib/systemd/tests/testdata/units/testsuite-69.sh: No such file or directory
```

lnussel:
Weird. I am off tomorrow. Will take a look Monday.

yuwata:
(Hmm, I cannot understand how TEST-69 works...)

lnussel:
Fixed by keeping `m->scheduled_shutdown_timeout = 0`. I couldn't reproduce the `No such file or directory` locally though. No idea how the test could fail in that way when the expected wall message isn't printed.

The test works by having two shells and uses pexpect to check for the wall message in the 2nd shell
https://github.com/systemd/systemd/commit/48f3bc5cc4dd52155a06753da5ef2cd9d48f7b07

lnussel:
current ci failure looks unrelated `fatal: unable to access 'https://salsa.debian.org/systemd-team/systemd.git/': The requested URL returned error: 403`

bluca:
Yes, Salsa is down

lnussel:
does this need a new label to trigger rereview resp ready for merge marker?

src/login/logind-dbus.c
```diff
@@ -2102,14 +2178,18 @@ static void reset_scheduled_shutdown(Manager *m) {
         m->nologin_timeout_source = sd_event_source_unref(m->nologin_timeout_source);
 
         m->scheduled_shutdown_type = NULL;
+        m->scheduled_shutdown_timeout = USEC_INFINITY;
+        m->scheduled_shutdown_uid = UID_INVALID;
+        freep(&m->scheduled_shutdown_tty);
+        freep(&m->wall_message);
```
poettering, line 2184: nope. please do note use freep like this. it's used for `_cleanup_` macro use only. Use mfree() insted:

```c
m->scheduled_shutdown_tty = mfree(m->scheduled_shutdown_tty);
```

lnussel: #22672

PR #22523, opened 2022-02-15
mkosi: Support running systemd-networkd-tests.py in QEMU

daandemeyer:


DaanDeMeyer:
I'm currently stuck on test_vxlan failing when run in QEMU but I'm not sure whether it's because I'm missing a dependency or because adding the "25-" prefix causes some config files to be processed in a different order.

The issue with test_vxlan seems to be that the vxlan-slaac interface is not appearing. cc @yuwata maybe you have an idea what could be causing the issue?

Running in mkosi is easy:

```
$ mkosi qemu
# Wait for boot
$ /root/src/test/test-network/systemd-networkd-tests.py
```

mkosi.default.d/fedora/10-mkosi.fedora
```diff
@@ -75,3 +75,10 @@ Packages=
         netcat
         e2fsprogs
         compsize
+        # Required to run systemd-networkd-tests.py
+        iproute
```
bluca, line 79: can we have these for other distros too?

daandemeyer: Preferably in a separate PR :p. I don't mind adding the packages, but testing everything actually works takes quite a bit of time so I'd rather leave that to someone that actually uses the distro.

DaanDeMeyer:
So the vxlan test fails on Centos CI as well so it seems like it's caused by the file renames. Will have to dig deeper.

DaanDeMeyer:
Found the problem, renaming netdev-link-local-addressing-yes.network to 25-netdev-link-local-addressing-yes.network caused it to get match against veth99 in test_vxlan instead of 25-vxlan-veth99.network which changed the behavior of the test. Fixed by renaming netdev-link-local-addressing-yes.network to 26-link-local-addressing-yes.network instead.

I also had to rename macsec.network to 26-macsec.network instead of 25-macsec.network because 25-macsec.network already exists in test/test-network/conf.

mkosi.default.d/fedora/10-mkosi.fedora
bluca: yeah I wasn't asking to make sure it works, simply to keep the installed packages in sync (to avoid surprises) - for debian and ubuntu, this should be the list:

iproute2
dnsmasq-base
wireguard-tools
isc-dhcp-server

bluca (approved):

yuwata (approved):
Thank you! LGTM.

mkosi.default.d/fedora/10-mkosi.fedora
yuwata: Oops, I forgot this comment and merged this PR. Let's do that later.

PR #22526, opened 2022-02-15
homed: map container UID range for homed home directories

poettering:
Inspired by, based on, and replacing #22239 by @brauner.

It's a very similar patch, but uses append_identity_range() to generate the range (so that in the unlikely event we are not actually using UID_NOBODY on the backing fs but one from the container range things will work reasonably OK). Also, and more importantly this maps the container range we already document in https://systemd.io/UIDS-GIDS/

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,18 +12,39 @@ CHANGES WITH 251:
         * Services with Restart=always and a failing ExecCondition= will no longer
           be restarted, to bring ExecCondition= in line with Condition*= settings.
 
-        * In v250 systemd-homed started making use of UID mapped mounts for the
-          home areas if the kernel and used file system support it. Files are
-          now internally owned by the "nobody" user (i.e. the user typically
-          used for indicating "this ownership is not mapped"), and dynamically
-          mapped to the UID used locally on the system via the UID mapping
-          mount logic of recent kernels.
-          In the current implementation systemd-homed only maps a limited
-          number of UIDs and GIDs making it impossible to run unprivileged
-          containers that want to map a full POSIX compliant UID and GID range
-          with their rootfs located within the systemd-homed managed home area.
-          This will be fixed in subsequent releases. See
-          https://github.com/systemd/systemd/pull/22239 for a proposal.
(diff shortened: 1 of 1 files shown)


brauner (approved):

docs/UIDS-GIDS.md
```diff
@@ -233,6 +233,27 @@ safely use the NSS user database as allocation check, too. Note that if you
...
+files/directories in home directories managed by `systemd-homed` cannot be
+owned by UIDs/GIDs outside of these four ranges (attempts to `chown()` files to
+UIDs outside of these ranges will fail). Thus, if container trees are to be
+placed within a home directory managed by `systemd-homed` they should take
+these ranges into consideration and either place the trees at base UID 0 (and
+then map them to a higher UID range for use in user namespacing via another
+level of UID mapped mounts, at *runtime*) or at a base UID from the container
+UID range. That said, placing container trees (and in fact any
+files/directories not owned by the home directory's user) in home directories
+is generally a questionnable idea (regardless if `systemd-homed` is used or
```
bluca: `questionable`
`regardless of whether`

docs/UIDS-GIDS.md
```diff
@@ -233,6 +233,27 @@ safely use the NSS user database as allocation check, too. Note that if you
...
+not), given this typically breaks quota assumptions, makes it impossible for
+users to properly manage all files in their own home directory due to
+permission problems, introduces security issues around SETUID and severly
```
bluca: `severely`

bluca (approved):

src/home/homework-mount.c
```diff
@@ -209,6 +209,13 @@ static int make_userns(uid_t stored_uid, uid_t exposed_uid) {
         if (r < 0)
                 return log_oom();
 
+        /* Also map the container range. People can user that to place containers owner by high UIDs in their
```
yuwata: `can user` -> `can use`?

poettering:
Addressed all raised points. No other changes. Upgrading label.

PR #22536, opened 2022-02-16
sd-dhcp-server: allow to run without pool

yuwata:


yuwata:
The last commit may be questionable, as `PoolSize=0` was previously handled as all addresses in the subnet, but now it means no address pool. So, strictly speaking this breaks backward compatibility, though I hope no one specifies 0 here.

If we take safety, then maybe keep the previous behavior and deprecate `PoolOffset=` and `PoolSize=`, and introduce a new option, say, `AddressRange=192.168.0.10 192.168.0.250`, and make it also accept boolean value.

WDYT?

PR #22544, opened 2022-02-17
hwdb: 60-keyboard: Support the buttons on CZC P10T tablet

lkundrak:
The machine has tree buttons connected to an EC that acts as a regular
AT-compatible keyboard controller. It can be either in "Windows 7" or
"Android" mode. It boots up with the earlier, but the Android build on
the tablet switches it on bootup (Windows presumably leaves it as-is).

The "Windows 7" mode, the behavior is very inconvenient: the Home button
emits multiple key presses that presumably do something in Windws 7 while
the second button toggles the RF Kill Switch in addition to producing a
scancode (it's labeled "Back" on Android version of the tablet).

The "Android" mode just sends the good ol' scan codes and this patch
handles them. On mainline Linux, the "x86-android-tablets" driver makes
sure we're in the correct mode.

### DIFF
--- a/hwdb.d/60-keyboard.hwdb
+++ b/hwdb.d/60-keyboard.hwdb
@@ -1892,6 +1892,20 @@ evdev:input:b0003v1038p0310*
  KEYBOARD_KEY_70046=f6
 
 ###########################################################
+# CZC
+###########################################################
+
+evdev:atkbd:dmi:bvn*:bvr*:bd*:svnCZC:pnODEON*TPC-10:*
+ KEYBOARD_KEY_01=leftmeta                               # Home button held for 2s
+ KEYBOARD_KEY_db=prog1                                  # Home button
+ KEYBOARD_KEY_dd=rfkill                                 # Second button labeled Wi-Fi
+
+evdev:atkbd:dmi:bvn*:bvr*:bd*:svnViewSonic:pnVPAD10:*
+ KEYBOARD_KEY_01=leftmeta                               # Home button held for 2s
+ KEYBOARD_KEY_db=prog1                                  # Home button
+ KEYBOARD_KEY_dd=back                                   # Second button labeled Back
+
+###########################################################
 # Other
 ###########################################################
 


hwdb.d/60-keyboard.hwdb
```diff
@@ -1891,6 +1891,20 @@ evdev:input:b0003v1038p0310*
...
+
+evdev:atkbd:dmi:bvn*:bvr*:bd*:svnCZC:pnODEON*TPC-10:*
+ KEYBOARD_KEY_01=leftmeta                               # Home button held for 2s
+ KEYBOARD_KEY_db=prog1                                  # Home button
+ KEYBOARD_KEY_dd=rfkill                                 # Second button labeled Wi-Fi
+
+evdev:atkbd:dmi:bvn*:bvr*:bd*:svnViewSonic:pnVPAD10:*
+ KEYBOARD_KEY_01=leftmeta                               # Home button held for 2s
+ KEYBOARD_KEY_db=prog1                                  # Home button
+ KEYBOARD_KEY_dd=back                                   # Second button labeled Back
```
yuwata, line 1906: These two entries are completely same, why do not you merge them?

yuwata:
cc @whot could you review this?

hwdb.d/60-keyboard.hwdb
lkundrak: That's a good suggestion, thanks.
Done.

lkundrak:
Pushed v2.
Changes since v1: Merged the entries together at @yuwata's suggestion

yuwata:
Hm, sorry I have no idea why I did not notice, but the `dd` key for the two models were different (back vs rfkill), but now they are merged. Is it OK?

lkundrak:
> Hm, sorry I have no idea why I did not notice, but the `dd` key for the two models were different (back vs rfkill), but now they are merged. Is it OK?

Ah, you're correct. The two models of the tablet are indeed the same hardware, but the label on the buttons differs -- one is marked as a back button, on another it's marked as radio toggle. Forgot about it, since the commit comes from 2018, but the kernel side got sorted out just recently.

Reverting back to v1.

Issue #22477, opened 2022-02-10
Assertion 'q->bus_request' failed at src/resolve/resolved-bus.c:65, function reply_query_state()


evverx:
**systemd version the issue has been seen with**
```
$ git describe
v250-840-g4ba5594390
```

```sh
# resolvectl service _xmpp-server._tcp gmail.com
Resolve call failed: DNSSEC validation failed: failed-auxiliary

# resolvectl revert eth0

# resolvectl
Global
           Protocols: +LLMNR +mDNS -DNSOverTLS DNSSEC=allow-downgrade/supported
    resolv.conf mode: stub
Fallback DNS Servers: 1.1.1.1#cloudflare-dns.com 8.8.8.8#dns.google 1.0.0.1#cloudflare-dns.com 8.8.4.4#dns.google 2606:4700:4700::1111#cloudflare-dns.com 2001:4860:4860::8888#dns.google
                      2606:4700:4700::1001#cloudflare-dns.com 2001:4860:4860::8844#dns.google

Link 2 (eth0)
Current Scopes: LLMNR/IPv4 LLMNR/IPv6
     Protocols: -DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=allow-downgrade/supported

# resolvectl service _xmpp-server._tcp gmail.com
```

```sh
Added positive unauthenticated non-confidential cache entry for _xmpp-server._tcp.gmail.com IN SRV 900s on eth0/INET/1.1.1.1
Added positive unauthenticated non-confidential cache entry for _xmpp-server._tcp.gmail.com IN SRV 900s on eth0/INET/1.1.1.1
Added positive unauthenticated non-confidential cache entry for _xmpp-server._tcp.gmail.com IN SRV 900s on eth0/INET/1.1.1.1
Added positive unauthenticated non-confidential cache entry for _xmpp-server._tcp.gmail.com IN SRV 900s on eth0/INET/1.1.1.1
Added positive unauthenticated non-confidential cache entry for _xmpp-server._tcp.gmail.com IN SRV 900s on eth0/INET/1.1.1.1
Regular transaction 35985 for <_xmpp-server._tcp.gmail.com IN SRV> on scope dns on */* now complete with <success> from network (unsigned; non-confidential).
Freeing transaction 62013.
Looking up RR for alt1.xmpp-server.l.google.com IN A.
Looking up RR for alt1.xmpp-server.l.google.com IN AAAA.
[...]

PR #22546, opened 2022-02-17
resolved: rework how we reply to D-Bus messages for resolution requests

poettering:
This reworks how we reply to D-Bus messages that come in for resolution
requests. Previously, we'd store them in the .bus_request field of
the main DnsQuery (but not any auxiliary one), and reply to it whenever
we had something to reply. In error paths this could mean we'd
accidentally reply twice.

This cleans this logic up: whenever we reply to a message we'll now go up
the tree of auxiliary queries, to find the primary query, i.e. the one
we actually want to reply to. Once we found it, we take out the bus
message object, resetting it to NULL. This way we can be sure we'll
reply at most once to each message.

Fixes: #22477

### DIFF
--- a/src/resolve/resolved-bus.c
+++ b/src/resolve/resolved-bus.c
@@ -59,57 +59,122 @@ static int dns_query_bus_track(DnsQuery *q, sd_bus_message *m) {
         return 0;
 }
 
-static int reply_query_state(DnsQuery *q) {
-
+static sd_bus_message *dns_query_steal_request(DnsQuery *q) {
         assert(q);
-        assert(q->bus_request);
+
+        /* Find the main query, it's the one that owns the message */
+        while (q->auxiliary_for)
+                q = q->auxiliary_for;
+
+        /* Let's take the request message out of the DnsQuery object, so that we never send requests twice */
+        return TAKE_PTR(q->bus_request);
+}
+
+_sd_printf_(3, 4) static int reply_method_errorf(
+                DnsQuery *query,
+                const char *error_name,
+                const char *format,
+                ...) {
+
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *req = NULL;
+        va_list ap;
+        int r;
+
+        assert(query);
+        assert(format);
+
+        req = dns_query_steal_request(query);
+        if (!req) /* No bus message set anymore? then we already replied already, let's not answer a second time */
+                return 0;
+
+        va_start(ap, format);
+        r = sd_bus_reply_method_errorfv(req, error_name, format, ap);
+        va_end(ap);
+
+        return r;
+}
+
+_sd_printf_(3, 4) static int reply_method_errnof(
+                DnsQuery *query,
+                int err,
+                const char *format,
+                ...) {
+
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *req = NULL;
+        int r;
+
+        assert(query);
+
+        req = dns_query_steal_request(query);
+        if (!req) /* No bus message set anymore? then we already replied already, let's not answer a second time */
+                return 0;
+
+        if (format) {
+                va_list ap;
+
+                va_start(ap, format);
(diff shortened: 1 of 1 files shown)


src/resolve/resolved-bus.c
```diff
@@ -262,11 +312,15 @@ static void bus_method_resolve_hostname_complete(DnsQuery *query) {
                 goto finish;
 
         r = sd_bus_send(q->manager->bus, reply, NULL);
+        if (r < 0)
+                goto finish;
+
+        q->bus_request = sd_bus_message_unref(q->bus_request);
```
yuwata: Hmm... On transaction failure or any other error paths, only `reply_query_state()` or `reply_method_errno()` are called, and there is no fallback on failure in sending it. I think that should be.

But, when a transaction succeeds, then error on sending message triggers fallback. I forgot the detailed implementation of  sd_bus_send(), but if sd_bus_send() failed after sending message, then two messages are still sent.

I prefer to change the order of `unref` and `send` here, for consistency with other error paths.
```
q->bus_request = sd_bus_message_unref(q->bus_request);
r = sd_bus_send(q->manager->bus, reply, NULL);
```

yuwata:
Also, this is a kind of cleanup. I prefer to merge my patch at first, and then apply this. So, the backport can be minimal.

src/resolve/resolved-bus.c
keszybz: Hmm, I don't understand this comment.

sd_bus_send prepares the message and either writes it to the socket or puts it on the queue. If any of the fails, we should invoke the fallback path in `finish`, including sending of the error message. So I don't think we should unset `q->bus_request` until after sd_bus_send has returned successfully.

src/resolve/resolved-bus.c
```diff
@@ -59,54 +59,69 @@ static int dns_query_bus_track(DnsQuery *q, sd_bus_message *m) {
...
 
         switch (q->state) {
 
         case DNS_TRANSACTION_NO_SERVERS:
-                return sd_bus_reply_method_errorf(q->bus_request, BUS_ERROR_NO_NAME_SERVERS, "No appropriate name servers or networks for name found");
+                return sd_bus_reply_method_errorf(req, BUS_ERROR_NO_NAME_SERVERS, "No appropriate name servers or networks for name found");
 
         case DNS_TRANSACTION_TIMEOUT:
-                return sd_bus_reply_method_errorf(q->bus_request, SD_BUS_ERROR_TIMEOUT, "Query timed out");
+                return sd_bus_reply_method_errorf(req, SD_BUS_ERROR_TIMEOUT, "Query timed out");
```
keszybz: Shouldn't this instead be:
```C
switch (q->state) {


case DNS_TRANSACTION_NO_SERVERS:
      return reply_method_errorf(q, SD_BUS_ERROR_TIMEOUT, "All attempts to contact name servers or networks failed");
case DNS_TRANSACTION_INVALID_REPLY:
      return reply_method_errorf(q, BUS_ERROR_INVALID_REPLY, "Received invalid reply");
...
```

It seems inconsistent to open-code the call to dns_query_steal_request() here.

keszybz (commented):
Looks very reasonable… but I admit I didn't grok Yu's comment.

src/resolve/resolved-bus.c
yuwata: I mean, for a failed transaction, sd_bus_send() is called only once. But for a successful transaction, sd_bus_send() is called at most twice. Of course, that happens only when the first send() failed, and thus I hope only one message at most is sent, but I am not sure.

src/resolve/resolved-bus.c
poettering: true

poettering:
Force pushed a new version, with everything addressed. PTAL.

Also incorportated Yu's original patch, hope that works for you @yuwata?

yuwata (approved):
LGTM.

bluca (approved):

PR #22525, opened 2022-02-15
tests: allow running all the services with SYSTEMD_LOG_LEVEL

evverx:
by passing `TEST_SYSTEMD_LOG_LEVEL=...`.

It should make it easier to figure out what exactly services do there.
For example, with SYSTEMD_LOG_LEVEL=debug userdbd (v249) prints
```
varlink-5: New incoming message: {"method":"io.systemd.UserDatabase.GetUserRecord","parameters":{}}
```
before it crashes and systemd-resolved prints
```
varlink-21: New incoming message: {"method":"io.systemd.Resolve.ResolveAddress","parameters":{"address":[127,0,0,1],"flags":0,"ifindex":1000000,"family":0}}
```
and those messages are helpful (especially when scripts causing them
aren't clever enough to keep track of random stuff they send to systemd
:-))

@mrc0mmand could you take a look?

PR #22547, opened 2022-02-17
test: fuzz our dbus interfaces with dfuzzer

mrc0mmand:
Pretty much WIP so far, since there's a couple of issues I have to deal with first to make this useful in production.

Most notably teach dfuzzer to not print shell color codes when not attached to a tty, since it makes the logs quite messy.

@evverx I wonder - how exactly did you manage to hit https://github.com/systemd/systemd/issues/22477 (was it even with "this" dfuzzer?).

Related links:
  * https://github.com/matusmarhefka/dfuzzer/pull/12
  * https://github.com/matusmarhefka/dfuzzer/pull/10
  * https://github.com/matusmarhefka/dfuzzer/pull/9

### DIFF
--- a/test/TEST-21-DFUZZER/test.sh
+++ b/test/TEST-21-DFUZZER/test.sh
@@ -12,6 +12,11 @@ QEMU_TIMEOUT="${QEMU_TIMEOUT:-1800}"
 
 command -v dfuzzer >/dev/null || exit 0
 
+if ! get_bool "$IS_BUILT_WITH_ASAN"; then
+    echo "systemd is built without ASan, skipping..."
+    exit 0
+fi
+
 test_append_files() {
     local workspace="${1:?}"
 


evverx:
> was it even with "this" dfuzzer?

Not exactly. But it's based on this dfuzzer and as far as I can tell the only difference is that it's possible to pass a seed corpus with valid dbus messages to somewhat speed it up. That particular dbus message came straight from the `resolvectl` manpage. I'll need to take a closer look.

evverx:
Having said that even without seed corpora dfuzzer found bugs in systemd in the past so I think it makes sense to bring it here in its current form. I'm not sure it should be run on PRs though. I'd probably run it for a few hours daily with a cron job or something.

test/TEST-99-DFUZZER/test.sh
```diff
@@ -0,0 +1,20 @@
...
+set -e
+
+TEST_DESCRIPTION="tbd"
+TEST_NO_NSPAWN=1
+TEST_MASK_SERVICES=0
+
+# shellcheck source=test/test-functions
+. "${TEST_BASE_DIR:?}/test-functions"
+
+command -v dfuzzer >/dev/null || exit 0
```
evverx: Looking at
```
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA login:
```
it seems it's installed on Arch :-) but I wonder where it came from?

mrc0mmand: https://github.com/systemd/systemd-centos-ci/commit/b6e5dff15d7e33604e21c56de6ef7ba30e972c11 :-)

mrc0mmand:
> Having said that even without seed corpora dfuzzer found bugs in systemd in the past so I think it makes sense to bring it here in its current form. I'm not sure it should be run on PRs though. I'd probably run it for a few hours daily with a cron job or something.

In this case the runtime of the fuzzer is "finite", so it runs for a couple of minutes at most, making it great for an integration test. But I definitely agree having the seeded and/or "long running" version would be great as well. I'm not sure, though, if it's supported by "this" dfuzzer (I don't know about any other implementation).

test/units/testsuite-99.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# Overmount /var/lib/machines with a size-limited tmpfs, as fuzzing
+# the org.freedesktop.machine1 stuff makes quite a mess
+mount -t tmpfs -o size=50M tmpfs /var/lib/machines
+
+# Fuzz both the system and the session buses (where applicable)
+for bus in "${BUS_LIST[@]}"; do
+    echo "Bus: $bus (system)"
+    # Activate the name (if it's activatable) before fuzzing it
+    busctl introspect "$bus" /
```
evverx: Without `--no-pager` it seems in interactive mode it keeps asking to press "Enter"
```
Failed to execute 'pager', using next fallback pager: No such file or directory
WARNING: terminal is not fully functional
Press RETURN to continue
```

I also added `ASAN_OPTIONS=detect_leaks=0` to run it manually under ASan/UBSan.

Another issue I ran into is that it can't seem to activate `homed`
```
+ for bus in "${BUS_LIST[@]}"
+ echo 'Bus: org.freedesktop.home1 (system)'
Bus: org.freedesktop.home1 (system)
+ busctl introspect org.freedesktop.home1 /
Bus n/a: changing state UNSET → OPENING
sd-bus: starting bus by connecting to /run/dbus/system_bus_socket...
Bus n/a: changing state OPENING → AUTHENTICATING
Bus n/a: changing state AUTHENTICATING → HELLO
Sent message type=method_call sender=n/a destination=org.freedesktop.DBus path=/org/freedesktop/DBus interface=org.freedesktop.DBus member=Hello cookie=1 reply_cookie=0 signature=n/a error-name=n/a errora
Got message type=method_return sender=org.freedesktop.DBus destination=:1.37 path=n/a interface=n/a member=n/a cookie=4294967295 reply_cookie=1 signature=s error-name=n/a error-message=n/a
Bus n/a: changing state HELLO → RUNNING
Sent message type=method_call sender=n/a destination=org.freedesktop.home1 path=/ interface=org.freedesktop.DBus.Introspectable member=Introspect cookie=2 reply_cookie=0 signature=n/a error-name=n/a erroa
Got message type=error sender=org.freedesktop.DBus destination=:1.37 path=n/a interface=n/a member=n/a cookie=4294967295 reply_cookie=2 signature=s error-name=org.freedesktop.DBus.Error.NameHasNoOwner er.
Failed to introspect object / of service org.freedesktop.home1: Could not activate remote peer.
Bus n/a: changing state RUNNING → CLOSED
+ at_exit
```
so I had to start it manually as well.

mrc0mmand: Interesting, it worked fine for me locally. Looks like this is going to be fun in CIs :-)

evverx:
WOW
```
[191241.254648] systemd-machined[205]: ==205==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7fa813289730 at pc 0x7fa818e38a5f bp 0x7ffc8c3cd940 sp 0x7ffc8c3cd938
[191241.255480] systemd-machined[205]: WRITE of size 4 at 0x7fa813289730 thread T0
[191241.396637] systemd-machined[205]:     #0 0x7fa818e38a5e in sd_bus_message_read_basic ../src/libsystemd/sd-bus/bus-message.c:3543
[191241.398253] systemd-machined[205]:     #1 0x7fa818e44c36 in sd_bus_message_readv ../src/libsystemd/sd-bus/bus-message.c:4478
[191241.398772] systemd-machined[205]:     #2 0x7fa818e4603e in sd_bus_message_read ../src/libsystemd/sd-bus/bus-message.c:4610
[191241.399780] systemd-machined[205]:     #3 0x432801 in bus_image_method_mark_read_only ../src/machine/image-dbus.c:235
[191241.401957] systemd-machined[205]:     #4 0x7fa818e5aa72 in method_callbacks_run ../src/libsystemd/sd-bus/bus-objects.c:413
[191241.402940] systemd-machined[205]:     #5 0x7fa818e6420e in object_find_and_run ../src/libsystemd/sd-bus/bus-objects.c:1322
[191241.404093] systemd-machined[205]:     #6 0x7fa818e65b58 in bus_process_object ../src/libsystemd/sd-bus/bus-objects.c:1452
[191241.407330] systemd-machined[205]:     #7 0x7fa818ec475d in process_message ../src/libsystemd/sd-bus/sd-bus.c:2963
[191241.408561] systemd-machined[205]:     #8 0x7fa818ec4c77 in process_running ../src/libsystemd/sd-bus/sd-bus.c:3005
[191241.410251] systemd-machined[205]:     #9 0x7fa818ec7ad8 in bus_process_internal ../src/libsystemd/sd-bus/sd-bus.c:3225
[191241.411503] systemd-machined[205]:     #10 0x7fa818ec7c72 in sd_bus_process ../src/libsystemd/sd-bus/sd-bus.c:3252
[191241.413149] systemd-machined[205]:     #11 0x7fa818ecb373 in io_callback ../src/libsystemd/sd-bus/sd-bus.c:3603
[191241.415129] systemd-machined[205]:     #12 0x7fa81908687f in source_dispatch ../src/libsystemd/sd-event/sd-event.c:3598
[191241.415906] systemd-machined[205]:     #13 0x7fa81908f099 in sd_event_dispatch ../src/libsystemd/sd-event/sd-event.c:4183
[191241.416308] systemd-machined[205]:     #14 0x7fa81909020e in sd_event_run ../src/libsystemd/sd-event/sd-event.c:4244
[191241.416832] systemd-machined[205]:     #15 0x7fa8188ecff5 in bus_event_loop_with_idle ../src/shared/bus-util.c:122
[191241.417159] systemd-machined[205]:     #16 0x40b5a4 in manager_run ../src/machine/machined.c:312
[191241.417403] systemd-machined[205]:     #17 0x40bc23 in run ../src/machine/machined.c:357
[191241.417626] systemd-machined[205]:     #18 0x40be3a in main ../src/machine/machined.c:364
[191241.417846] systemd-machined[205]:     #19 0x7fa81739255f in __libc_start_call_main (/lib64/libc.so.6+0x2d55f)
[191241.418814] systemd-machined[205]:     #20 0x7fa81739260b in __libc_start_main_alias_1 (/lib64/libc.so.6+0x2d60b)
[191241.420206] systemd-machined[205]:     #21 0x407104 in _start (/usr/lib/systemd/systemd-machined+0x407104)
[191241.421459] systemd-machined[205]: Address 0x7fa813289730 is located in stack of thread T0 at offset 48 in frame
[191241.422817] systemd-machined[205]:     #0 0x432644 in bus_image_method_mark_read_only ../src/machine/image-dbus.c:226
[191241.424838] systemd-machined[205]:   This frame has 2 object(s):
[191241.425984] systemd-machined[205]:     [48, 49) 'read_only' (line 230) <== Memory access at offset 48 partially overflows this variable
[191241.428698] systemd-machined[205]:     [64, 120) 'details' (line 239)
[191241.432572] systemd-machined[205]: HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
[191241.434348] systemd-machined[205]:       (longjmp and C++ exceptions *are* supported)
```

test/units/testsuite-99.sh
evverx: FWIW Looks like the journal gets lost somewhere along the way
```
-bash-5.1# journalctl --no-pager
Considering root directory '/run/log/journal'.
Root directory /run/log/journal added.
Considering root directory '/var/log/journal'.
Root directory /var/log/journal added.
Considering directory '/var/log/journal/ab68a6bd82294a00a40ad5c229ad7920'.
Directory /var/log/journal/ab68a6bd82294a00a40ad5c229ad7920 added.
Journal effective settings seal=no keyed_hash=yes compress=no compress_threshold_bytes=8B
File /var/log/journal/ab68a6bd82294a00a40ad5c229ad7920/system.journal added.
Journal filter: none
-- Journal begins at Thu 2022-02-17 22:58:30 UTC, ends at Thu 2022-02-17 22:58:30 UTC. --
Feb 17 22:58:30 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA systemd-journald[48]: System Journal (/var/log/journal/ab68a6bd82294a00a40ad5c229ad7920) is 138.9M, max 188.1M, 49.2M free.
Root directory /run/log/journal removed.
Root directory /var/log/journal removed.
Directory /var/log/journal/ab68a6bd82294a00a40ad5c229ad7920 removed.
mmap cache statistics: 389 context cache hit, 3 window list hit, 1 miss
```
so some other dbus methods should probably be blocked somewhere

mrc0mmand: I know there's an issue where fuzzing `org.freedesktop.timedate1` causes the system (and journal) time to be in future and then journald considers it as a corrupted/unclean journal, marks it as such, and rotates it. Not sure if it's the same issue.

evverx: I think https://github.com/systemd/systemd/pull/22525 was partly prompted by issues like that. Among other things I couldn't reliably extract backtraces from the journal so I just forwarded the logs to console and looked for them there.

evverx: @mrc0mmand I'll take a closer look later but as far as I can remember `dfuzzer` sometimes just stops fuzzing for no apparent reason and half the methods kind of stuck in limbo. I'm not sure whether timeouts/connections are configurable there but it would make sense to look at that as well.

evverx:
> making it great for an integration test

Agreed. The backtrace was convincing enough :-) I haven't figured out how to handle backtraces/coredumps yet (considering the journal is (understandably) unreliable. It should be possible to forward logs to the console but `test-functions` needs tweaking if that's the only way to do that.

mrc0mmand:
> > making it great for an integration test
> 
> Agreed. The backtrace was convincing enough :-) I haven't figured out how to handle backtraces/coredumps yet (considering the journal is (understandably) unreliable. It should be possible to forward logs to the console but `test-functions` needs tweaking if that's the only way to do that.

For now I added a pretty simple "safety" net, which checks the journal for coredumps before exiting the script (be it successfully or on an error), not sure how infallible it actually is. As for the time-related issues, I marked `SetLocalRTC` from `org.freedesktop.timedate1` as "destructive" to keep the RTC/system time sane and it seems to have helped.

evverx:
FWIW I'm not sure whether dfuzzer is actively maintained or not. I wonder if it would make sense to fork it (hopefully temporarily) and apply those two PRs on top (and name it `systemd-dfuzzerd` :-)). With them it should be possible to run it under ASan/UBsan.

mrc0mmand:
> Have you tried running it under ASan/UBsan? Can it show that backtrace once the test exits? Ideally, it should show the backtrace along with the message/object causing it?

Yes and so far I haven't managed to trigger any coredump ("only" ASan backtraces, which are already handled by `test-functions`). I'll play around it a bit more, especially with timeout-related stuff.

> 
> FWIW turns out `-L` can be used to make it even more verbose. With it `dfuzzer` fails with
> 
...
> 
> and additionally creates a log file where it's possible to look for "crashes"
> 
> ```
> # grep -i markread LOG/org.freedesktop.machine1
> org.freedesktop.machine1.Image;/org/freedesktop/machine1/image/_2ehost;MarkReadOnly;b;falseCrash
> org.freedesktop.machine1.Image;/org/freedesktop/machine1/image/_2ehost;MarkReadOnly;b;trueCrash
> ```

Hah, that's nifty, thanks for pointing that out!


> FWIW I'm not sure whether dfuzzer is actively maintained or not. I wonder if it would make sense to fork it (hopefully temporarily) and apply those two PRs on top (and name it `systemd-dfuzzerd` :-)). With them it should be possible to run it under ASan/UBsan.

I thought about that as well :-)

evverx:
> I'll play around it a bit more, especially with timeout-related stuff.

I have to admit I don't even know how `dfuzzer` reacts to timeouts :-)

It seems it tends to exit with `3` under ASan/UBsan though:
```
  WARN Get - memory usage 3.4x more than initial memory
```
it would be great if it was possible to somehow turn off that "memory leak detector" there (since LSan can really detect memory leaks) and let the test keep going (or, maybe, it should just ignore `3` itself). I'm not sure.

evverx:
Speaking of memory leaks to catch them in PID1 the test should run `systemctl daemon-reload` somewhere explicitly. Without that they go unnoticed.

test/units/testsuite-99.sh
evverx: > as far as I can remember dfuzzer sometimes just stops fuzzing for no apparent reason

Turns out logs are helpful sometimes :-) and looking at https://github.com/matusmarhefka/dfuzzer/blob/5419e835211366732195ce85f05cdc4761841a60/src/fuzz.c#L956-L960 it seems that happens because "advanced" signatures aren't supported.

mrc0mmand:
> > I'll play around it a bit more, especially with timeout-related stuff.
> 
> I have to admit I don't even know how `dfuzzer` reacts to timeouts :-)
> 
> It seems it tends to exit with `3` under ASan/UBsan though:
> 
> ```
>   WARN Get - memory usage 3.4x more than initial memory
> ```
> 
> it would be great if it was possible to somehow turn off that "memory leak detector" there (since LSan can really detect memory leaks) and let the test keep going (or, maybe, it should just ignore `3` itself). I'm not sure.

Yeah, this is a bit annoying. I tried to tweak the memory threshold (via `-m MEM`), but that only led to systemd-logind/systemd-udevd getting killed by OOM (even when I assigned 8G of RAM to the VM). So I guess having an option to disable the warnings would be useful.

> Speaking of memory leaks to catch them in PID1 the test should run `systemctl daemon-reload` somewhere explicitly. Without that they go unnoticed.

Good point!

evverx:
I "fixed" the memory leak detector with
```diff
diff --git a/src/fuzz.c b/src/fuzz.c
index b6ccd18..692a24a 100644
--- a/src/fuzz.c
+++ b/src/fuzz.c
@@ -727,7 +727,7 @@ int df_fuzz_test_method(const int statfd, long buf_size, const char *name,

        // test passed
        if (leaking_mem_flg)    // warning
-               return 3;
+               return 0;
        df_verbose("\r  \e[32mPASS\e[0m %s\n", df_list.df_method_name);
        return 0;
```
I'm not proud of it :-) but with it on top of all your patches I got it to work under ASan/UBsan.

evverx:
FWIW @mrc0mmand dfuzzer strikes again: https://github.com/systemd/systemd/issues/23097.

evverx:
To catch issues like https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2020-1712 I think polkit should be installed as well. FWIW I suspect that bug was found with `dfuzzer` probably.

test/units/testsuite-99.sh
```diff
@@ -0,0 +1,127 @@
...
+mount -t tmpfs -o size=50M tmpfs /var/lib/machines
+
+# Create a log dir for dfuzzer to store potential failing test cases
+mkdir /dfuzzer-logs
+
+# Fuzz both the system and the session buses (where applicable)
+for bus in "${BUS_LIST[@]}"; do
+    echo "Bus: $bus (system)"
+    # Activate the name (if it's activatable) before fuzzing it
+    busctl --no-pager introspect "$bus" /
```
evverx: Before I forget now that https://github.com/matusmarhefka/dfuzzer/pull/59 is merged it can be removed.

evverx: > Another issue I ran into is that it can't seem to activate homed

As far as I can tell even with that PR included `dfuzzer` can't activate some services. My guess would be that it worked on CentOS CI because it enabled those services somehow somewhere. In my containers I have to run `systemctl enable systemd-homed` explicitly to make it work.

mrc0mmand: Yeah, it's explicitly pulled in in the testsuite service file (https://github.com/systemd/systemd/pull/22547/commits/a98a0a89f5e1f1f627131cff6c528bd1856bf31c?diff=unified&w=0#diff-db05ad3d53fd2ac714b4a714d249f3e237090416860cdca964a1dfbafad429beR4-R7).

evverx: I ran the test with `make clean setup run` so in theory that part should have been included. I'm not sure why it didn't work out.

mrc0mmand: I wonder if this is not a bug on systemd's side, since:

```
dbus-broker-launch[1524]: Activation request for 'org.freedesktop.home1' failed: The systemd unit 'dbus-org.freedesktop.home1.service' could not be found.
```

and indeed, there's no dbus service file for homed:

```
$ ls /usr/lib/systemd/system | grep dbus-
dbus-broker.service
dbus-daemon.service
dbus-org.freedesktop.hostname1.service
dbus-org.freedesktop.import1.service
dbus-org.freedesktop.locale1.service
dbus-org.freedesktop.login1.service
dbus-org.freedesktop.machine1.service
dbus-org.freedesktop.oom1.service
dbus-org.freedesktop.portable1.service
dbus-org.freedesktop.timedate1.service
```

since we don't explicitly install it:

https://github.com/systemd/systemd/blob/main/units/meson.build#L228-L233

evverx: If it isn't intentional I think it's safe to assume `dfuzzer` has found another issue

mrc0mmand: After a quick chat with Lennart, this is actually expected, since in case of the "dynamic" units the dbus service symlink is created when the respective systemd unit is enabled, e.g.:

```
# systemctl enable systemd-homed
Created symlink /etc/systemd/system/dbus-org.freedesktop.home1.service → /usr/lib/systemd/system/systemd-homed.service.
Created symlink /etc/systemd/system/multi-user.target.wants/systemd-homed.service → /usr/lib/systemd/system/systemd-homed.service.
Created symlink /etc/systemd/system/systemd-homed.service.wants/systemd-homed-activate.service → /usr/lib/systemd/system/systemd-homed-activate.service.
Created symlink /etc/systemd/system/sockets.target.wants/systemd-userdbd.socket → /usr/lib/systemd/system/systemd-userdbd.socket.
```

So I guess the test should be tweaked to either do `systemctl preset` or enable the required systemd services.

evverx:
From https://github.com/systemd/systemd/pull/23382#issuecomment-1126935421
> I don't think the "dfuzzer" test should be included in the coverage report to make it easier to spot absolutely untested features

I'd add that it doesn't seem to make much sense to run it without ASan/UBsan so once it's more or less up and running I'd add something like `is_built_with_asan` to skip it. It does make sense to run it under Valgrind but I'm not sure how to avoid skipping it.

mrc0mmand:
@evverx this should be, finally, ready to go. With the last commit this PR is basically a no-op, since I need to wire it up in the CentOS CI sanitizer jobs first, but it should do the right thing given the tests before that.

test/TEST-01-BASIC/test.sh
```diff
@@ -6,6 +6,7 @@ TEST_DESCRIPTION="Basic systemd setup"
 IMAGE_NAME="basic"
 RUN_IN_UNPRIVILEGED_CONTAINER=${RUN_IN_UNPRIVILEGED_CONTAINER:-yes}
 TEST_REQUIRE_INSTALL_TESTS=0
+TEST_MASK_SERVICES=0
```
evverx: I think TEST-01-BASIC is supposed to test all services so `TEST_MASK_SERVICES` should probably be `1` here and `0` everywhere else.

evverx: Looks like the name is confusing. It isn't about testing those services. It's about masking them :-)

evverx: I think I'd rename it to `TEST_SUPPORTING_SERVICES_SHOULD_BE_MASKED` to make it clear what it does but I think it can be done later.

mrc0mmand: Hehe, yeah, I'm bad at naming things, I'll change it to `TEST_SUPPORTING_SERVICES_SHOULD_BE_MASKED` :-)

test/TEST-21-DFUZZER/test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+
+TEST_DESCRIPTION="Fuzz our D-Bus interfaces with dfuzzer"
+TEST_NO_NSPAWN=1
```
evverx, line 6: I think it would be great if it was possible to run it in containers eventually.

In the meantime, I commented this out and ran the test in a container. `dfuzzer` failed to activate `org.freedesktop.network1` (as expected) but the container got stuck
```sh
[138852.713992] testsuite-21.sh[326]: Cannot open suppression file './dfuzzer.conf'
[138852.719082] testsuite-21.sh[326]: Loading suppressions from file '/etc/dfuzzer.conf'
[138852.719903] testsuite-21.sh[326]: [SESSION BUS]
[138852.728448] testsuite-21.sh[326]: Bus not found.
[138852.729512] testsuite-21.sh[326]: [SYSTEM BUS]
[138877.760142] testsuite-21.sh[326]: Error while activating 'org.freedesktop.network1': Timeout was reached.
[138877.763772] testsuite-21.sh[326]: Error while calling method 'GetConnectionUnixProcessID': GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: The connection does not exist
[138877.765029] testsuite-21.sh[326]: Couldn't get the PID of the tested process
[138877.765767] testsuite-21.sh[326]: Exit status: 4
[138877.799794] testsuite-21.sh[325]: Finished with result: exit-code
[138877.800325] testsuite-21.sh[325]: Main processes terminated with: code=exited/status=4
[138877.800901] testsuite-21.sh[325]: Service runtime: 25.089s
[138877.801466] testsuite-21.sh[325]: CPU time consumed: 28ms
[138877.833149] testsuite-21.sh[83]: + at_exit
[138877.833702] testsuite-21.sh[83]: + local found_cd=0
[138877.834067] testsuite-21.sh[83]: + read -r exe
[138877.836818] testsuite-21.sh[334]: ++ coredumpctl -F COREDUMP_EXE
[138877.838069] testsuite-21.sh[335]: ++ sort -u
[138877.912147] testsuite-21.sh[83]: + [[ 0 -eq 0 ]]
[138877.914151] testsuite-21.sh[83]: + journalctl --rotate --vacuum-size=16M
[138878.086090] testsuite-21.sh[83]: + [[ 1 -ne 0 ]]
[138878.086916] testsuite-21.sh[83]: + systemctl start end.service
```

evverx: Looks like it can be "fixed" with `systemctl start --job-mode=isolate end.service` but I have to admit I haven't figured out why the test tries to start `end.service` in the first place.

evverx: Got it. `dfuzzer` kicks `end.service` out with `ClearJobs` when it fuzzes `systemd`. Here it wasn't kicked out because `systemd` wasn't fuzzed so the test got stuck. I think `--job-mode=isolate` should do then.

mrc0mmand: Nice catch, thanks!

test/units/testsuite-21.sh
```diff
@@ -0,0 +1,97 @@
...
+    # TODO: check fo ASan/UBSan errors
+    local found_cd=0
+    while read -r exe; do
+        coredumctl info "$exe"
+        found_cd=1
+    done < <(coredumpctl -F COREDUMP_EXE | sort -u)
+    [[ $found_cd -eq 0 ]] || exit 1
+
+    # Limit the maximum journal size (because fuzzing)
+    journalctl --rotate --vacuum-size=16M
```
evverx: I think if the idea is to avoid running out of space the journal should probably be trimmed on every dfuzzer run. I mean, if the test gets to this point it can safely keep the journal to make it possible to extract coredumps and ASan backtraces as usual and delete it after that.

test/TEST-21-DFUZZER/test.sh
mrc0mmand: In the end I had to switch to `--job-mode=flush`, since `--job-mode=isolate` apparently requires a unit with `AllowIsolate=yes`, which introduced other issues. With `--job-mode=flush` it seems to work in nspawn as well (at least on my machine):

```
[4813392.164242] testsuite-21.sh[79]: + at_exit
[4813392.164595] testsuite-21.sh[79]: + local found_cd=0
[4813392.164835] testsuite-21.sh[79]: + read -r exe
[4813392.165568] testsuite-21.sh[318]: ++ coredumpctl -F COREDUMP_EXE
[4813392.166143] testsuite-21.sh[319]: ++ sort -u
[4813392.195955] testsuite-21.sh[79]: + [[ 0 -eq 0 ]]
[4813392.196240] testsuite-21.sh[79]: + [[ 1 -ne 0 ]]
[4813392.196418] testsuite-21.sh[79]: + systemctl start --job-mode=flush end.service
[4813392.240030] sh[322]: + systemctl poweroff --no-block
[4813392.261461] testsuite-21.sh[321]: Job for end.service canceled.
[  OK  ] Stopped testsuite-21.service.
[  OK  ] Stopped end.service.
```

test/units/testsuite-21.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -eux
+set -o pipefail
+
+# Save the end.service state before we start fuzzing, as it might get changed
+# on the fly by one of the fuzzers
+systemctl is-enabled end.service && SHUTDOWN_AT_EXIT=1 || SHUTDOWN_AT_EXIT=0
```
evverx: It doesn't seem to work as expected in interactive mode where `is-enabled` returns `0` along with `static`
```sh
[145593.489054] testsuite-21.sh[83]: + systemctl is-enabled end.service
[  OK  ] Finished systemd-update-utmp-runlevel.service.
[145593.620193] testsuite-21.sh[84]: static
[145593.628080] testsuite-21.sh[83]: + SHUTDOWN_AT_EXIT=1
```

Having said that I think it would make sense to get this PR merged, turn it on on CentOS CI and polish it with the help of the CI.

mrc0mmand: Ah, good point, I completely forgot we add the `end.service` to the transaction via `systemd.wants=`, hopefully fixed.

PR #22548, opened 2022-02-17
packit: drop unnumbered patches as well

mrc0mmand:
@keszybz has been a *naughty boy* and introduced an unnumbered patch in https://src.fedoraproject.org/rpms/systemd/c/4c2d7265ec2e92e4024bee0a961b33a8f52c2bab?branch=rawhide which breaks the *beautiful* sed we use to drop backported patches.

I vaguely remember a discussion about using unnumbered patches on the fedora-devel ML, so let's drop them as well to make CIs happy again.


yuwata (approved):
LGTM.

PR #22550, opened 2022-02-17
boot: EFI mixed mode support

medhefgo:
This is only tested on QEMU so far and I would prefer to merge this only after it is confirmed that it works on real hardware. Note that OVMF has its own COFF emulation support, so to properly test this a suitable OVMF build without support for it has to be used.

I considered making ia32 builds on 64bit machines automatic, but that would probably make the whole meson logic for gnu-efi even more unsightly. I think asking distros to use a second build dir for this is reasonable. If desired, a "systemd-boot" target alias could be provided so simplify building a bit.

The bootctl changes are only compile-tested.

### DIFF
--- a/man/bootctl.xml
+++ b/man/bootctl.xml
@@ -341,6 +341,11 @@
         token when adding a second installation.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--all-architectures</option></term>
+        <listitem><para>Install binaries for all supported EFI architectures (this implies <option>--no-variables</option>).</para></listitem>
+      </varlistentry>
+
       <xi:include href="standard-options.xml" xpointer="no-pager"/>
       <xi:include href="standard-options.xml" xpointer="json" />
       <xi:include href="standard-options.xml" xpointer="help"/>
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -74,6 +74,7 @@ static enum {
 } arg_entry_token_type = ARG_ENTRY_TOKEN_AUTO;
 static char *arg_entry_token = NULL;
 static JsonFormatFlags arg_json_format_flags = JSON_FORMAT_OFF;
+static bool arg_arch_all = false;
 
 STATIC_DESTRUCTOR_REGISTER(arg_esp_path, freep);
 STATIC_DESTRUCTOR_REGISTER(arg_xbootldr_path, freep);
@@ -411,6 +412,36 @@ finish:
         (void) munmap(buf, st.st_size);
         *v = x;
         return r;
+}
+
+static const char *get_efi_arch(void) {
+        /* Detect EFI firmware architecture of the running system. On mixed mode systems, it could be 32bit
+         * while the kernel is running in 64bit. */
+
+#ifdef __x86_64__
+        _cleanup_free_ char *platform_size = NULL;
+        int r;
+
+        r = read_one_line_file("/sys/firmware/efi/fw_platform_size", &platform_size);
+        if (r == -ENOENT)
+                return EFI_MACHINE_TYPE_NAME;
+        if (r < 0) {
+                log_warning_errno(r, "Error reading EFI firmware word size, assuming '%u': %m", __WORDSIZE);
+                return EFI_MACHINE_TYPE_NAME;
+        }
+
+        if (streq(platform_size, "64"))
+                return EFI_MACHINE_TYPE_NAME;
+        if (streq(platform_size, "32"))
+                return "ia32";
+
+        log_warning(
+                "Unknown EFI firmware word size '%s', using default word size '%u' instead.",
+                platform_size,
+                __WORDSIZE);
+#endif
+
+        return EFI_MACHINE_TYPE_NAME;
 }
 
 static int enumerate_binaries(const char *esp_path, const char *path, const char *prefix) {
@@ -829,7 +860,7 @@ static int copy_one_file(const char *esp_path, const char *name, bool force) {
         return r;
 }
 
-static int install_binaries(const char *esp_path, bool force) {
+static int install_binaries(const char *esp_path, const char *arch, bool force) {
         _cleanup_closedir_ DIR *d = NULL;
         int r = 0;
 
@@ -837,10 +868,13 @@ static int install_binaries(const char *esp_path, bool force) {
         if (!d)
                 return log_error_errno(errno, "Failed to open \""BOOTLIBDIR"\": %m");
 
+        const char *suffix = strjoina(arch, ".efi");
+        const char *suffix_signed = strjoina(arch, ".efi.signed");
+
         FOREACH_DIRENT(de, d, return log_error_errno(errno, "Failed to read \""BOOTLIBDIR"\": %m")) {
                 int k;
 
-                if (!endswith_no_case(de->d_name, ".efi") && !endswith_no_case(de->d_name, ".efi.signed"))
+                if (!endswith_no_case(de->d_name, suffix) && !endswith_no_case(de->d_name, suffix_signed))
                         continue;
 
                 /* skip the .efi file, if there's a .signed version of it */
@@ -1327,6 +1361,8 @@ static int help(int argc, char *argv[], void *userdata) {
                "                       Entry token to use for this installation\n"
                "     --json=pretty|short|off\n"
                "                       Generate JSON output\n"
+               "     --all-architectures\n"
+               "                       Install all supported EFI architectures\n"
                "\nSee the %2$s for details.\n",
                program_invocation_short_name,
                link,
@@ -1349,6 +1385,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_MAKE_ENTRY_DIRECTORY,
                 ARG_ENTRY_TOKEN,
                 ARG_JSON,
+                ARG_ARCH_ALL,
         };
 
         static const struct option options[] = {
@@ -1368,6 +1405,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "make-machine-id-directory", required_argument, NULL, ARG_MAKE_ENTRY_DIRECTORY      }, /* Compatibility alias */
                 { "entry-token",               required_argument, NULL, ARG_ENTRY_TOKEN               },
                 { "json",                      required_argument, NULL, ARG_JSON                      },
+                { "all-architectures",         no_argument,       NULL, ARG_ARCH_ALL                  },
                 {}
         };
 
@@ -1470,7 +1508,10 @@ static int parse_argv(int argc, char *argv[]) {
                         r = parse_json_argument(optarg, &arg_json_format_flags);
                         if (r <= 0)
                                 return r;
+                        break;
 
+                case ARG_ARCH_ALL:
+                        arg_arch_all = true;
                         break;
 
                 case '?':
@@ -1601,13 +1642,14 @@ static int verb_status(int argc, char *argv[], void *userdata) {
 
                 SecureBootMode secure = efi_get_secure_boot_mode();
                 printf("System:\n");
-                printf("     Firmware: %s%s (%s)%s\n", ansi_highlight(), strna(fw_type), strna(fw_info), ansi_normal());
-                printf("  Secure Boot: %sd (%s)\n",
+                printf("      Firmware: %s%s (%s)%s\n", ansi_highlight(), strna(fw_type), strna(fw_info), ansi_normal());
+                printf(" Firmware Arch: %s\n", get_efi_arch());
+                printf("   Secure Boot: %sd (%s)\n",
                        enable_disable(IN_SET(secure, SECURE_BOOT_USER, SECURE_BOOT_DEPLOYED)),
                        secure_boot_mode_to_string(secure));
 
                 s = tpm2_support();
-                printf(" TPM2 Support: %s%s%s\n",
+                printf("  TPM2 Support: %s%s%s\n",
                        FLAGS_SET(s, TPM2_SUPPORT_FIRMWARE|TPM2_SUPPORT_DRIVER) ? ansi_highlight_green() :
                        (s & (TPM2_SUPPORT_FIRMWARE|TPM2_SUPPORT_DRIVER)) != 0 ? ansi_highlight_red() : ansi_highlight_yellow(),
                        FLAGS_SET(s, TPM2_SUPPORT_FIRMWARE|TPM2_SUPPORT_DRIVER) ? "yes" :
@@ -1617,14 +1659,14 @@ static int verb_status(int argc, char *argv[], void *userdata) {
 
                 k = efi_get_reboot_to_firmware();
                 if (k > 0)
-                        printf(" Boot into FW: %sactive%s\n", ansi_highlight_yellow(), ansi_normal());
+                        printf("  Boot into FW: %sactive%s\n", ansi_highlight_yellow(), ansi_normal());
                 else if (k == 0)
-                        printf(" Boot into FW: supported\n");
+                        printf("  Boot into FW: supported\n");
                 else if (k == -EOPNOTSUPP)
-                        printf(" Boot into FW: not supported\n");
+                        printf("  Boot into FW: not supported\n");
                 else {
                         errno = -k;
-                        printf(" Boot into FW: %sfailed%s (%m)\n", ansi_highlight_red(), ansi_normal());
+                        printf("  Boot into FW: %sfailed%s (%m)\n", ansi_highlight_red(), ansi_normal());
                 }
                 printf("\n");
 
@@ -1926,6 +1968,8 @@ static int verb_install(int argc, char *argv[], void *userdata) {
         if (r < 0)
                 return r;
 
+        const char *arch = arg_arch_all ? "" : get_efi_arch();
+
         RUN_WITH_UMASK(0002) {
                 if (install) {
                         /* Don't create any of these directories when we are just updating. When we update
@@ -1940,7 +1984,7 @@ static int verb_install(int argc, char *argv[], void *userdata) {
                                 return r;
                 }
 
-                r = install_binaries(arg_esp_path, install);
+                r = install_binaries(arg_esp_path, arch, install);
                 if (r < 0)
                         return r;
 
@@ -1969,13 +2013,16 @@ static int verb_install(int argc, char *argv[], void *userdata) {
 
         (void) sync_everything();
 
-        if (arg_touch_variables)
-                r = install_variables(arg_esp_path,
-                                      part, pstart, psize, uuid,
-                                      "/EFI/systemd/systemd-boot" EFI_MACHINE_TYPE_NAME ".efi",
-                                      install);
+        if (!arg_touch_variables)
+                return 0;
 
-        return r;
+        if (arg_arch_all) {
+                log_info("Not changing EFI variables with --all-architectures.");
+                return 0;
+        }
+
+        char *path = strjoina("/EFI/systemd/systemd-boot", arch, ".efi");
+        return install_variables(arg_esp_path, part, pstart, psize, uuid, path, install);
 }
 
 static int verb_remove(int argc, char *argv[], void *userdata) {
@@ -2041,7 +2088,13 @@ static int verb_remove(int argc, char *argv[], void *userdata) {
         if (!arg_touch_variables)
                 return r;
 
-        q = remove_variables(uuid, "/EFI/systemd/systemd-boot" EFI_MACHINE_TYPE_NAME ".efi", true);
+        if (arg_arch_all) {
+                log_info("Not changing EFI variables with --all-architectures.");
+                return r;
+        }
+
+        char *path = strjoina("/EFI/systemd/systemd-boot", get_efi_arch(), ".efi");
+        q = remove_variables(uuid, path, true);
         if (q < 0 && r >= 0)
                 r = q;
 


src/boot/efi/meson.build
```diff
@@ -10,17 +10,26 @@ if not get_option('efi') or get_option('gnu-efi') == 'false'
         subdir_done()
 endif
```
poettering: not sure i follow here. why wouldn't the distros just build systemd as a whole for both archs? i.e. what#s the benefit of being able to build 64bit systemd userspace in a combination with 32bit systemd efispace?

medhefgo: Well, a distro like debian could certainly pull the ia32 binaries and add them to the x64 package (though, tbh, this sounds ugly on a debian tooling standpoint to me).

But arch for example has no 32bit port anymore. There you'd want to have a way to build ia32 without having to do a full 32bit cross build just so you can toss out most of it. You'd need to provide a minimum of a lot of 32bit dependencies just so meson could run successfully, even when you just want to issue the build commands for the ia32 efi binaries to speed things up.

(Or I am not getting your question.)

src/boot/efi/util.h
```diff
@@ -24,6 +24,17 @@
...
+/* gnu-efi has no format specifier for (U)INTN. */
+#if __SIZEOF_POINTER__ == 4
+#  define FMT_U L"%u"
+#  define FMT_D L"%d"
+#elif __SIZEOF_POINTER__ == 8
+#  define FMT_U L"%lu"
+#  define FMT_D L"%ld"
+#else
+#  error "Unexpected pointer size"
+#endif
```
poettering: hmm, so C99 PRIu32-style macros do not include the `%` prefix. Our FMT_UID, … macros generally do. That has a reason: the PRIu32-style macros are generic, and still allow you to specify alignment, precision and so on. THe FMT_UID macros on the other hand don't allow that, because we now pretty excactly how UIDs should be formatted.

Now, this here is a generic use case, it is a way to display any natively sized integer. Hence it probably should not contain the "%" prefix.

Hence maybe do:

```c
#define PRIuN L"u"
```

and similar instead, and then use it as:

```c
SPrint(…, "foo %" PRIuN " bar");
```

pretty? no! but I think the most obvious way to do this for people who have accepted C99 syle PRI* macros into their lives.

src/boot/efi/boot.c
```diff
@@ -138,24 +137,20 @@ static void cursor_right(
...
-        if (!line_in)
-                line_in = L"";
-
-        len = StrLen(line_in);
+        len = *line_in ? StrLen(*line_in) : 0;
         size = len + 1024;
         line = xnew(CHAR16, size);
         print = xnew(CHAR16, x_max + 1);
-        StrCpy(line, line_in);
+        StrCpy(line, *line_in ?: L"");
```
poettering: we probably should move strempty() into the fundamental stuff shared between userspace and efispace

src/boot/efi/boot.c
```diff
@@ -531,6 +532,8 @@ static void print_status(Config *config, CHAR16 *loaded_image_path) {
                 if (entry->device)
                     Print(L"        device: %D\n", DevicePathFromHandle(entry->device));
                 ps_string(L"        loader: %s\n", entry->loader);
+                for (CHAR16 **initrd = entry->initrd; initrd && *initrd; initrd++)
```
poettering: we should probably move STRV_FOREACH into the fundamental stuff shared by userspace and efispace

src/boot/efi/boot.c
```diff
@@ -2239,6 +2224,36 @@ static void config_load_xbootldr(
         config_load_entries(config, new_device, root_dir, NULL);
 }
 
+static EFI_STATUS initrd_prepare(const ConfigEntry *entry, CHAR16 **ret_options) {
+        assert(entry);
+        assert(ret_options);
+
+        if (entry->type != LOADER_LINUX || !entry->initrd)
```
poettering: why the entry-type conditionalization? i.e. the right hand conditionalization should conver that too? i.e. if the type is not LOADER_LINUX then initrd is gonna be NULL anyway, no?

src/boot/efi/boot.c
```diff
@@ -2239,6 +2224,36 @@ static void config_load_xbootldr(
...
+                return EFI_SUCCESS;
```
poettering, line 2319: our coding style mandates that we init all return parameters on success and leave them untouched on failure. In this codepath here ret_options is left as is even on success.

src/boot/efi/boot.c
```diff
@@ -2251,6 +2291,8 @@ static EFI_STATUS initrd_prepare(const ConfigEntry *entry, CHAR16 **ret_options)
         }
 
         *ret_options = TAKE_PTR(options);
+        *ret_initrd = TAKE_PTR(initrd);
+        *ret_initrd_size = size;
```
poettering, line 2371: the upper EFI_SUCESS code paths should initialize these to NULL/0, as per coding style

src/boot/efi/pe.c
```diff
@@ -160,6 +166,46 @@ static void locate_sections(
...
+
+                if (compat->type == 0 || compat->size == 0)
+                        break;
+
+                if (compat->type == 1 &&
+                    compat->size >= sizeof(LinuxPeCompat1) &&
+                    compat->machine_type == TARGET_MACHINE_TYPE)
+                        return compat->entry_point;
+
+                addr += compat->size;
```
poettering, line 205: overflow checks?

src/boot/efi/pe.c
```diff
@@ -171,20 +217,30 @@ EFI_STATUS pe_alignment_info(
...
-        assert(ret_size_of_image);
-        assert(ret_section_alignment);
 
         dos = (const struct DosFileHeader *) base;
         if (!verify_dos(dos))
                 return EFI_LOAD_ERROR;
 
         pe = (const struct PeFileHeader*) ((const UINT8 *)base + dos->ExeHeader);
-        if (!verify_pe(pe))
+        if (!verify_pe(pe, TRUE))
```
poettering: maybe write as:

```c
if (!verify_pe(pe, /* allow_alt= */ TRUE))
```

src/boot/efi/pe.c
```diff
@@ -207,7 +263,7 @@ EFI_STATUS pe_memory_locate_sections(
                 return EFI_LOAD_ERROR;
 
         pe = (const struct PeFileHeader*)&base[dos->ExeHeader];
-        if (!verify_pe(pe))
+        if (!verify_pe(pe, FALSE))
```
poettering: ditto and elsewhere too

src/boot/bootctl.c
```diff
@@ -251,6 +259,30 @@ static int get_file_version(int fd, char **v) {
...
+                log_error_errno(r, "Error reading EFI firmware platform size: %m");
+                return EFI_MACHINE_TYPE_NAME;
+        }
+
+        if (streq(platform_size, "64"))
+                return EFI_MACHINE_TYPE_NAME;
+        if (streq(platform_size, "32"))
+                return EFI_MACHINE_TYPE_ALT_NAME;
+
+        log_warning("Unknown EFI platform size: %s", platform_size);
```
poettering: for warnings we usually say what we are doing now, i.e. suffix with ", ignoring" or so. But in this case we should probably say ", using the default word size"

src/boot/bootctl.c
```diff
@@ -251,6 +259,30 @@ static int get_file_version(int fd, char **v) {
...
+static const char *efi_arch(void) {
+#ifdef __x86_64__
+        _cleanup_free_ char *platform_size = NULL;
+        int r;
+
+        r = read_one_line_file("/sys/firmware/efi/fw_platform_size", &platform_size);
+        if (r == -ENOENT)
+                return EFI_MACHINE_TYPE_NAME;
+        if (r < 0) {
+                log_error_errno(r, "Error reading EFI firmware platform size: %m");
```
poettering: given this is not fatal for anything, it's mostly a warning and thus should be downgraded to log_warning_errno()...

So a "platform size" is a weird concept. I figure they mean "word size"? maybe write that in the log message, i.e.

"Error reading EFI firmware platform word size, using the default word size"

man/bootctl.xml
```diff
@@ -277,6 +277,17 @@
...
+      <varlistentry>
+        <term><option>-a</option></term>
+        <term><option>--arch=auto|all|ARCH</option></term>
+        <listitem><para>Install binaries for the specified EFI architecture (e.g. <literal>x64</literal> or
+        <literal>ia32</literal> on x86).</para>
+
+        <para>Defaults to <literal>auto</literal>, which will use the kernel-reported EFI platform size of
+        the running system to decide the right binaries to use. Setting this to <literal>all</literal> will
+        install binaries for all available architectures (this implies <option>--no-variables</option>).</para></listitem>
+      </varlistentry>
```
poettering, line 347: so this feature as it looks right now seems to support exactly one or two archs, which are compiled into bootctl. 

I think we should make this entirely generic, so that we copy basically `/usr/lib/systemd/boot/efi/systemd-boot*.efi`, i.e. any arch that is supported. Or if you specifiy --arch= any subset of it.

i.e. I think distros should be able to implement multiarch to the max if they like and have arm and x86 EFI binaries of sd-boot in place if they like, and "bootctl" should be able to drop it in.

Or in otherwords: instead of compiled in list of archs, just pick up every file that is installed.

src/boot/efi/meson.build
poettering: on debian, with its true multiarch support I presume you#d just be able to install the x86-64, the i386 and the arm package of systemd-boot.deb side by side, no? given the EFI files they install don't conflict, this should just work, no?

poettering: > But arch for example has no 32bit port anymore. There you'd want to have a way to build ia32 without having to do a full 32bit cross build just so you can toss out most of it. You'd need to provide a minimum of a lot of 32bit dependencies just so meson could run successfully, even when you just want to issue the build commands for the ia32 efi binaries to speed things up.

But that sounds like a workaround on a deliberate choice that arch made, that they don't want to support 32bit anymore. Shoehoring a 32bit build into a pretend-64bit build is a hack, no?

anyway, @keszybz what's your take on this?

src/boot/efi/boot.c
medhefgo: You can create an LOADER_EFI config entry with initrds just fine. Does it make any particular sense? Not really. Also, this is just defensive programming.

poettering: good point

src/boot/efi/meson.build
medhefgo: It is perfectly fine for a distro to drop 32bit kernel/userspace. But that still means that you need to provide the necessary tools to boot your 64bit OS. And there are 64bit systems that have been crippled with a 32bit EFI firmware and arch should be able to easily support this. (I feel by your logic it should drop i386-pc grub too.)

Also, I highly doubt dpkg even allows cross-arch package installs (lib32-* packages are still amd64 packages). What seem to suggest is pulling ia32 binaries out of the i386 dpkg and stuff it into the amd64 dpkg. That just sounds like a terrible idea.

@bluca you seem to be involved in debian, what's your take on this?

man/bootctl.xml
medhefgo: > Or in otherwords: instead of compiled in list of archs, just pick up every file that is installed.

This is what the `all` arch is for. And I believe this should be explicitly opt-in. I certainly wouldn't want bootctl to always install ia32 binaries when I don't need them.

I now made it so that you can select from any efi arches we support, but we still default to auto detecting and only installing for that arch, which is imho the right thing to do.

src/boot/efi/meson.build
bluca: debian still has full support for i386 native builds, so for now it's not an issue - although there are recurring talks of dropping it, so at some point it might become an issue. However, it will likely follow ubuntu's example, where select packages are still building i386 for libraries, to support legacy third-party software like steam. I guess we'd ask for a similar exception here. However, at the moment the efi binaries are just in the main systemd package, so they can't be co-installed, which is a problem we need to fix.

There might be a future were even those select i386 builds are deprecated, although it's not anywhere close - at that point, it would be good to be able to build this. Although at that point the hardware with 32 bit efi land might be too obsolete to even bother with. Is there any modern hardware that does that? I thought it was mostly old apple stuff and old atoms netbooks

poettering: > Also, I highly doubt dpkg even allows cross-arch package installs (lib32-* packages are still amd64 packages). What seem to suggest is pulling ia32 binaries out of the i386 dpkg and stuff it into the amd64 dpkg. That just sounds like a terrible idea.

dpkg expressly intends to support *multiarch* (as opposed to *multilib* as the other distros do), i.e. the ability to combine library packages from any arch into one OS tree, including unrelated archs. See: https://wiki.debian.org/Multiarch/HOWTO

poettering: So yeah, what @bluca describes how Ubuntu does it makes sense to me: drop i386 mostly, but do i386 builds still for a select set of packages.

src/boot/efi/pe.c
```diff
@@ -160,6 +166,46 @@ static void locate_sections(
         }
 }
 
+static UINT32 get_compat_entry_address(const struct DosFileHeader *dos, const struct PeFileHeader *pe) {
+        UINTN addr = 0, size = 0;
+        const CHAR8 *sections[] = { (CHAR8 *) ".compat", NULL };
```
poettering: static?

src/boot/efi/pe.c
poettering: no addressed?

poettering:
so in the context of multib we usually talk about the "secondary" arch, which appears to be conceptually close to what you refer to here as "alt" arch. Is the "alt" moniker something used by EFI specs anywhere? Otherwise I#d call this "secondary" here too, to minimize the concepts we introduce here.

I'd prefer if the build system stuff to allow building a 64bit systemd with a 32bit sd-boot would be split out into a separate arch, it seems to be something only some distros are looking for, and not sure it should go in this way.

src/boot/efi/meson.build
xnox: Debian and Ubuntu both have multiarch support and allow installing foo:i386 package on an amd64 system for example. (which results in installing or coinstalling foo_1.23-3_i386.deb and foo_1.23-3_amd64.deb on a single system). However we have been doing some interesting packaging to build package named foo-efi-ia32 as either _amd64 or _all whilst actually containing the EFI 32-bit binaries. So building/installing stuff in general has solutions.

In terms of hardware, I know that obsolete Macbooks Intel Core Duo are 64bit CPU but 32bit efi (note, not Core 2 Duo). And there were Intel MinnowBoards that had 64-bit cpu, had both 32 bit & 64 bit EFI firmware available, however by default out of the factory they shipped with the 32 bit firmware.

I am not aware of any other popular / large systems that have 32bit EFI firmware yet 64bit CPU. My understanding is that no other SKU/vendor repeated that mistake since. And all platforms since ship with matching EFI firmware and CPU, being 64-bit one.

Minnow board users should reflash into 64bit EFI firmware. The core duo macbook I had in University, it is long obsolete and not supported in terms of power efficiency CPU disk space etc. They can continue to use whatever they have on them, and shouldn't need to support or install new OS on them. Their batteries should be long dead too, and impossible to replace.

I am interested to know if there are any other SKUs that are being manufactured and if anybody supports them that ship ia32 efi firmware. I.e. stuff released in the last 10 years, i.e. from 2013 onwards. And has received Spectre Meltdown mitigations.

xnox:
From Ubuntu's point of view, we have never fully provided ia32 EFI enabled installer media or support. And ideally we wouldn't want for that to be built by default during x86_64 build. It is best for this to become an opt-in feature.

src/boot/efi/meson.build
mirh: FWIW 32-bit uefi still shipped up to Bay Trail (if not even some Cherry Trail system?)
Those ~2GHz low power quad cores aren't exactly workhorses, but they are absolutely usable.

xnox: Yeap, after my post above I got pointed at machines sold in 2015 with such setup. Not flagships or anything, but yeah, they were still made and sold less than 8 years ago.

medhefgo:
> so in the context of multib we usually talk about the "secondary" arch, which appears to be conceptually close to what you refer to here as "alt" arch. Is the "alt" moniker something used by EFI specs anywhere? Otherwise I#d call this "secondary" here too, to minimize the concepts we introduce here.

The efi spec says nothing here, as this is out of scope. It really doesn't matter to me what name we use, but considering the kernel puts this in a ".compat" section maybe use that?

> I'd prefer if the build system stuff to allow building a 64bit systemd with a 32bit sd-boot would be split out into a separate arch, it seems to be something only some distros are looking for, and not sure it should go in this way.

What you mean separate arch? Also, what's the big deal about it?

It's a very small change as it's mostly just passing the efi-cflags to a bunch of calls we make. You can hardly call this hacky/shoehorning imho and has real benefits/use cases.
In fact, it eases the ability to do test/debug builds for ia32 (as we both know, the ci has caught the occasional case due to 32 vs 64 UINTN), so that you don't have to setup a full blown qemu container for this or wait for ci to complain.

Also, all of this is clearly opt-in. Unless you *explicitly* ask for it, you will get your regular x64 build.

src/boot/efi/pe.c
medhefgo: The `compat->size > size` some lines above should take care of this, no?

man/bootctl.xml
poettering: I'd argue here that if you have those other flavours installed, then we should install them too. i.e. I'd see that as a drop-in thing: if the other flavours are installed we take them too. Why would you install them in /usr/ if you wouldn't also want them in /efi/?

poettering:
> The efi spec says nothing here, as this is out of scope. It really doesn't matter to me what name we use, but considering the kernel puts this in a ".compat" section maybe use that?

Sounds better than "alt" to me. But not a fan of needlessy cryptic abbreviations. I'd probably still call it secondary then. or at least "compatibility"...

poettering:
> What you mean separate arch? Also, what's the big deal about it?

Building a cross-binary from an otherwise same-arch tree sounds really weird, and a specific ack around one distro which killed one arch but actually still wants to support it at one place...

> It's a very small change as it's mostly just passing the efi-cflags to a bunch of calls we make. You can hardly call this hacky/shoehorning imho and has real benefits/use cases. In fact, it eases the ability to do test/debug builds for ia32 (as we both know, the ci has caught the occasional case due to 32 vs 64 UINTN), so that you don't have to setup a full blown qemu container for this or wait for ci to complain.

Well, but I might instead just cross build the whole of systemd for i386, no? For the testing usecase that'd be a ton more obvious and more enlightening. i.e. why should the UEFI stuff be so special there that we'd support builds that officially are for arch X but some bits are actually for arch Y.

src/boot/efi/util.c
```diff
@@ -57,7 +57,7 @@ EFI_STATUS efivar_set_uint_string(const EFI_GUID *vendor, const CHAR16 *name, UI
         assert(vendor);
         assert(name);
 
-        SPrint(str, ELEMENTSOF(str), L"%u", i);
+        SPrint(str, ELEMENTSOF(str), "%" PRIuN, i);
```
poettering: where did the L go?

medhefgo: Into the ether? :D

medhefgo:
> Building a cross-binary from an otherwise same-arch tree sounds really weird, and a specific ack around one distro which killed one arch but actually still wants to support it at one place...

That's not what arch is doing. It dropped i386 kernel/userspace. But you *still need tools* to get there. For non-EFI boot that's i386 bootloaders. For *crippled* ia32 EFIs on x86_64 CPUs, you need a *ia32* EFI bootloader.

> Well, but I might instead just cross build the whole of systemd for i386, no? For the testing usecase that'd be a ton more obvious and more enlightening. i.e. why should the UEFI stuff be so special there that we'd support builds that officially are for arch X but some bits are actually for arch Y.

But there is no point in doing that. You literally only need ia32 EFI binaries and that's it. No bootctl, no lib32 systemd, nada, nothing else. Meanwhile, to get this working, you'd have to pull in a ton of lib32 stuff (which has to be build and maintained). Just so you drop it all after the build because all you are interestd in are ia32 efi binaries.

poettering:
I'll leave the cross-build thing for @keszybz to okay, as our resident meson guru.

src/boot/efi/boot.c
```diff
@@ -531,6 +534,8 @@ static void print_status(Config *config, CHAR16 *loaded_image_path) {
                 if (entry->device)
                     Print(L"        device: %D\n", DevicePathFromHandle(entry->device));
                 ps_string(L"        loader: %s\n", entry->loader);
+                STRV_FOREACH(initrd, entry->initrd)
+                    Print(L"        initrd: %s\n", *initrd);
```
poettering, line 565: that indentation is killing me

medhefgo:
> I'll leave the cross-build thing for @keszybz to okay, as our resident meson guru.

You keep saying that like a mantra. But repeating it is not gonna improve the meson bus-factor… :joy_cat:

src/boot/efi/boot.c
```diff
@@ -1454,18 +1459,12 @@ static void config_entry_add_from_file(
...
-
-                                s = xpool_print(L"%s initrd=%s", initrd, new);
-                                FreePool(initrd);
-                                initrd = s;
-                        } else
-                                initrd = xpool_print(L"initrd=%s", new);
-
+                        entry->initrd = xreallocate_pool(
+                                entry->initrd,
+                                n_initrd * sizeof(UINT16 *),
```
poettering: the old size is `(n_initrd+1)*sizef(…)`, no?

bluca:
> > Building a cross-binary from an otherwise same-arch tree sounds really weird, and a specific ack around one distro which killed one arch but actually still wants to support it at one place...
> 
> That's not what arch is doing. It dropped i386 kernel/userspace. But you _still need tools_ to get there. For non-EFI boot that's i386 bootloaders. For _crippled_ ia32 EFIs on x86_64 CPUs, you need a _ia32_ EFI bootloader.
> 
> > Well, but I might instead just cross build the whole of systemd for i386, no? For the testing usecase that'd be a ton more obvious and more enlightening. i.e. why should the UEFI stuff be so special there that we'd support builds that officially are for arch X but some bits are actually for arch Y.
> 
> But there is no point in doing that. You literally only need ia32 EFI binaries and that's it. No bootctl, no lib32 systemd, nada, nothing else. Meanwhile, to get this working, you'd have to pull in a ton of lib32 stuff (which has to be build and maintained). Just so you drop it all after the build because all you are interestd in are ia32 efi binaries.

This makes sense to me - while Debian derivatives are not there _yet_, at some point that's the direction everyone has to move toward, simply because the cost of maintaining an entire bootstrappable toolchain just for an EFI bootloader would be so high. Of course 32bit EFI firmwares might completely disappear before we get to that point, which I dearly hope for, but if the latest hardware is from 2015 it's not _that_ long ago, sadly.
And for non-Debian derivatives, without full multiarch, this makes sense already today. The alternative would be to add enough meson glue that you can build _only_ the EFI binaries on a native build, but my spider sense tells me that would be way more invasive.

src/boot/efi/boot.c
```diff
@@ -2344,14 +2344,19 @@ static EFI_STATUS image_start(
         if (EFI_ERROR(err))
                 return log_error_status_stall(err, L"Error registering initrd: %r", err);
 
+        EFI_LOADED_IMAGE *loaded_image;
+        err = BS->OpenProtocol(
+                image,
```
poettering: we actually usually indent continuation after `(` to 16chars, not 8

src/boot/efi/boot.c
```diff
@@ -2361,14 +2366,31 @@ static EFI_STATUS image_start(
 
         efivar_set_time_usec(LOADER_GUID, L"LoaderTimeExecUSec", 0);
         err = BS->StartImage(image, NULL, NULL);
-        if (err != EFI_SUCCESS) {
+        if (err == EFI_SUCCESS)
                 /* Not using EFI_ERROR here because positive values are also errors like with any other
                  * (userspace) program. */
-                graphics_mode(FALSE);
-                return log_error_status_stall(err, L"Failed to execute %s (%s): %r", entry->title_show, entry->loader, err);
+                return err;
```
poettering: doesn't matter, but maybe `return EFI_SUCCESS` here is more obvious to read?

src/boot/efi/boot.c
```diff
@@ -2361,14 +2366,31 @@ static EFI_STATUS image_start(
...
+                if (EFI_ERROR(err)) {
+                        if (err != EFI_UNSUPPORTED)
+                                return log_error_status_stall(err, L"Error finding kernel compat entry address: %r", err);
+                } else {
+                        EFI_IMAGE_ENTRY_POINT kernel_entry =
+                                (EFI_IMAGE_ENTRY_POINT) ((UINT8 *) loaded_image->ImageBase + kernel_entry_address);
+
+                        err = kernel_entry(image, ST);
+                        if (err == EFI_SUCCESS)
+                                return err;
```
poettering: ditto

src/boot/efi/pe.c
```diff
@@ -160,6 +166,46 @@ static void locate_sections(
...
+                if (compat->type == 0 || compat->size == 0 || compat->size > size)
+                        break;
+
+                if (compat->type == 1 &&
+                    compat->size >= sizeof(LinuxPeCompat1) &&
+                    compat->machine_type == TARGET_MACHINE_TYPE)
+                        return compat->entry_point;
+
+                addr += compat->size;
+                size = compat->size;
```
poettering: hmm, i don't follow this here? you skip ovre `compat->size` bytes with the address, and then you look at the next `compat->size` bytes? how precisely is the LinuxPeCompat1 stuff laid out?

src/boot/efi/boot.c
```diff
@@ -2324,12 +2324,47 @@ static EFI_STATUS image_start(
...
+        CHAR16 *options = options_initrd ?: entry->options;
 
         path = FileDevicePath(entry->device, entry->loader);
         if (!path)
                 return log_error_status_stall(EFI_INVALID_PARAMETER, L"Error getting device path.");
 
         err = BS->LoadImage(FALSE, parent_image, path, NULL, 0, &image);
+        if (err == EFI_UNSUPPORTED && entry->type == LOADER_LINUX) {
+                /* If LoadImage fails, we may still start the kernel manually the same way we do for the
+                 * unified stub loader, which will then call the compat entry point if one exists. *
```
poettering: unexpected asterisk at end?

src/boot/efi/boot.c
```diff
@@ -2324,12 +2324,47 @@ static EFI_STATUS image_start(
...
+
+                 * Better be on the safe side. */
+                if (secure_boot_enabled())
+                        return log_error_status_stall(
+                                err,
+                                L"Kernel image '%s' is unsupported by firmware. Refusing to "
+                                L"try compat entry point because of secure boot.",
+                                entry->loader);
```
poettering: what's the rationale? why not allow this?

src/boot/efi/boot.c
```diff
@@ -2324,12 +2324,47 @@ static EFI_STATUS image_start(
...
 
         path = FileDevicePath(entry->device, entry->loader);
         if (!path)
                 return log_error_status_stall(EFI_INVALID_PARAMETER, L"Error getting device path.");
 
         err = BS->LoadImage(FALSE, parent_image, path, NULL, 0, &image);
+        if (err == EFI_UNSUPPORTED && entry->type == LOADER_LINUX) {
+                /* If LoadImage fails, we may still start the kernel manually the same way we do for the
+                 * unified stub loader, which will then call the compat entry point if one exists. *
+
```
poettering: spurious empty line?

src/boot/bootctl.c
```diff
@@ -251,6 +253,33 @@ static int get_file_version(int fd, char **v) {
         return r;
 }
 
+static const char *efi_arch(void) {
+#ifdef __x86_64__
+        _cleanup_free_ char *platform_size = NULL;
+        int r;
```
poettering, line 423: maybe add a comment here why this precisely is done? this is non obvious after all.

src/boot/bootctl.c
```diff
@@ -780,10 +809,17 @@ static int install_binaries(const char *esp_path, bool force) {
         if (!d)
                 return log_error_errno(errno, "Failed to open \""BOOTLIBDIR"\": %m");
 
+        const char *suffix = ".efi", *suffix_signed = ".efi.signed";
+        if (!streq_ptr(arg_arch, "all")) {
+                const char *arch = arg_arch ?: efi_arch();
+                suffix = strjoina(arch, suffix);
+                suffix_signed = strjoina(arch, suffix_signed);
+        }
```
poettering: btw, i think we should generate an error if the selected arch is not supported, i.e. no binary installed for this arch

src/boot/bootctl.c
```diff
@@ -1313,6 +1349,7 @@ static int help(int argc, char *argv[], void *userdata) {
                "                       variables cannot be written\n"
                "     --make-machine-id-directory=yes|no|auto\n"
                "                       Create $BOOT/$MACHINE_ID\n"
+               "  -a --arch=ARCH       EFI arch for installed binaries\n"
```
poettering: abbreviation seems unnecessary. I mean you have it a short option "a" already, no need to abbreviate the long option.

src/boot/bootctl.c
```diff
@@ -1416,6 +1454,17 @@ static int parse_argv(int argc, char *argv[]) {
...
+                case 'a':
+                        if (streq(optarg, "auto"))
+                                arg_arch = mfree(arg_arch);
+                        else if (STR_IN_SET(optarg, "all", "ia32", "x64", "arm", "aa64", "riscv64")) {
+                                r = free_and_strdup(&arg_arch, optarg);
+                                if (r < 0)
+                                        return log_oom();
+                        } else
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "Unknown EFI arch: %s", optarg);
+                        break;
```
poettering, line 1515: i have the suspicion this should actually permit being used multiple times, i.e. be an strv, not a simple string.

And I still think that if unspecified we should copy boot loaders for all archs for which one is installed. i.e. if people don't want some arch they shouldn't install the package for it, or conversely: installing the package for an arch should make the boot lodaer just work for that arch.

src/boot/bootctl.c
```diff
@@ -1924,11 +1974,23 @@ static int verb_install(int argc, char *argv[], void *userdata) {
...
 
-        if (arg_touch_variables)
-                r = install_variables(arg_esp_path,
-                                      part, pstart, psize, uuid,
-                                      "/EFI/systemd/systemd-boot" EFI_MACHINE_TYPE_NAME ".efi",
-                                      install);
+        if (!arg_touch_variables)
+                return r;
+
+        if (streq_ptr(arg_arch, "all"))
```
poettering: allowin an arch that is magic is evil... at the very least it should be  macro, so that it's obvious it's special. But actually I think there should be a separate boolean for the special cases. No need to shove all special cases into a single string.

poettering (changes requested):

src/boot/efi/pe.c
medhefgo: That was supposed to be a `-=` that vanished into the ether.

medhefgo:
(I dropped the `linux_exec` after `LoadImage` as it doesn't seem to work for the one user that I added it for (it was a long shot).)

For now I made bootctl simply default to auto-detection with a `--all-architectures` option (with no specific arch selection option). Partly because I am not particularly invested into bootctl and wanted to keep things simple.

I also disagree that we should default to installing every arch by default as probably 99% of use cases will be "install this on the currently running system" where we can simply detect the arch anyways and would want to change the efi vars too (which pretty much has to be arch dependent).
Also, not every distro splits its packages into 100+ subpackages just for the sake of it (like debian for example). And simply installing everything that is available when only one arch can be used anyways is just wasteful to me. And the detection is quick and easy to do.

src/boot/bootctl.c
medhefgo: This should probably be done separately, as it currently doesn't warn if nothing was copied.

poettering:
needs rebase

src/boot/efi/assert.c
```diff
@@ -6,7 +6,7 @@
 #include "util.h"
 
 void efi_assert(const char *expr, const char *file, unsigned line, const char *function) {
-        log_error_stall(L"systemd-boot assertion '%a' failed at %a:%u, function %a(). Halting.", expr, file, line, function);
+        log_error_stall(L"systemd-boot assertion '%a' failed at %a:%" PRIuN L", function %a(). Halting.", expr, file, line, function);
```
poettering: is this right?

reading the comments in `PrintLib.h` it appears to me %u is size of an int, and hence also size of an unsigned int, and hence the code actually was right here

src/boot/efi/boot.c
```diff
@@ -477,7 +477,7 @@ static void print_status(Config *config, CHAR16 *loaded_image_path) {
             Print(L"           secure boot: %s (%s)\n", yes_no(IN_SET(secure, SECURE_BOOT_USER, SECURE_BOOT_DEPLOYED)), secure_boot_mode_to_string(secure));
           ps_bool(L"                  shim: %s\n",      shim_loaded());
           ps_bool(L"                   TPM: %s\n",      tpm_present());
-            Print(L"          console mode: %d/%d (%lux%lu @%ux%u)\n", ST->ConOut->Mode->Mode, ST->ConOut->Mode->MaxMode - 1LL, x_max, y_max, screen_width, screen_height);
+            Print(L"          console mode: %d/%ld (%" PRIuN L"x%" PRIuN L" @%ux%u)\n", ST->ConOut->Mode->Mode, ST->ConOut->Mode->MaxMode - INT64_C(1), x_max, y_max, screen_width, screen_height);
```
poettering: I'd really prefer if we'd use PRIu32 style stuff for fixed size integers, like `screen_width` and `screen_height`. then it's much cloer to how we do stuff in userspace

src/boot/efi/boot.c
```diff
@@ -2570,11 +2571,10 @@ EFI_STATUS efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *sys_table) {
...
 
                 err = image_start(root_dir, image, &config, entry);
-                if (EFI_ERROR(err)) {
-                        graphics_mode(FALSE);
-                        log_error_stall(L"Failed to execute %s (%s): %r", entry->title_show, entry->loader, err);
+                if (err != EFI_SUCCESS)
+                        /* Not using EFI_ERROR here because positive values are also errors like with any
+                         * other (userspace) program. */
                         goto out;
-                }
```
poettering, line 2630: seems like an unrelated change? where di the graphics mode thing go?

src/boot/efi/boot.c
```diff
@@ -138,24 +137,20 @@ static void cursor_right(
...
         UINTN size, len, first = 0, cursor = 0, clear = 0;
 
-        assert(line_out);
+        assert(line_in);
 
-        if (!line_in)
-                line_in = L"";
-
-        len = StrLen(line_in);
+        len = *line_in ? StrLen(*line_in) : 0;
```
poettering: btw, in userspace we have `strlen_ptr` for this. mabe add here too?

src/boot/efi/boot.c
```diff
@@ -319,8 +314,10 @@ static BOOLEAN line_edit(
                 case KEYPRESS(0, 0, CHAR_CARRIAGE_RETURN):
                 case KEYPRESS(0, CHAR_CARRIAGE_RETURN, 0): /* EZpad Mini 4s firmware sends malformed events */
                 case KEYPRESS(0, CHAR_CARRIAGE_RETURN, CHAR_CARRIAGE_RETURN): /* Teclast X98+ II firmware sends malformed events */
-                        if (StrCmp(line, line_in) != 0)
-                                *line_out = TAKE_PTR(line);
+                        if (!streq(line, *line_in)) {
+                                FreePool(*line_in);
+                                *line_in = TAKE_PTR(line);
```
poettering, line 322: btw, in userspce we have free_and_replace() for this

src/boot/efi/boot.c
```diff
@@ -2251,28 +2251,64 @@ static void config_load_xbootldr(
...
 
         if (entry->type != LOADER_LINUX || !entry->initrd) {
                 ret_options = NULL;
+                ret_initrd = NULL;
+                ret_initrd_size = 0;
                 return EFI_SUCCESS;
         }
 
+        /* Note that order of initrds matters. The kernel will only look for microcode updates in the very
+         * first one it sees. */
```
poettering, line 2323: not just for this, but later initrds override contents of earlier ones, if they contain same inodes

src/boot/efi/boot.c
```diff
@@ -2367,14 +2367,19 @@ static EFI_STATUS image_start(
         if (EFI_ERROR(err))
                 return log_error_status_stall(err, L"Error registering initrd: %r", err);
 
+        EFI_LOADED_IMAGE *loaded_image;
```
poettering, line 2420: shouldn't we issue CloseProtocol in error paths on this?

src/boot/efi/pe.c
```diff
@@ -160,6 +166,45 @@ static void locate_sections(
         }
 }
 
+static UINT32 get_compatibility_entry_address(const struct DosFileHeader *dos, const struct PeFileHeader *pe) {
+        UINTN addr = 0, size = 0;
+        static const CHAR8 *sections[] = { (CHAR8 *) ".compat", NULL };
+
```
poettering, line 172: a comment here what this does would be great, as this is totally not obvious for the drive-by reader

src/boot/efi/pe.c
```diff
@@ -160,6 +166,45 @@ static void locate_sections(
...
+        locate_sections((struct PeSectionHeader *) ((UINT8 *) dos + section_table_offset(dos, pe)),
```
poettering: doesn't matter but strictly speaking you cast the const away, that `dos` carries. to be fully correct the two casts would still have it

src/boot/efi/pe.c
```diff
@@ -160,6 +166,45 @@ static void locate_sections(
...
+
+        if (size == 0)
+                return 0;
+
+        typedef struct {
+                UINT8 type;
+                UINT8 size;
+                UINT16 machine_type;
+                UINT32 entry_point;
+        } _packed_ LinuxPeCompat1;
```
poettering, line 192: is the `_packed_` about actual packing, or about alignment?

src/boot/efi/pe.c
```diff
@@ -160,6 +166,45 @@ static void locate_sections(
...
+                if (compat->type == 0 || compat->size == 0 || compat->size > size)
+                        break;
+
+                if (compat->type == 1 &&
+                    compat->size >= sizeof(LinuxPeCompat1) &&
+                    compat->machine_type == TARGET_MACHINE_TYPE)
+                        return compat->entry_point;
+
+                addr += compat->size;
+                size -= compat->size;
```
poettering, line 206: what about alignment? i figure this is trusted data, so not strictly an issue, but maybe add a comment about that.

(or alternatively validate alignment here, i.e. either as runtime error or as assert check `addr % __alignof__(LinuxPeCompat1) == 0)`, before dereffing the structure

poettering:
did another round of reviews. mostly looks good. not sure about the multlib build thing. will defer to @keszybz on that one, given it#s a build system issue, and he's more immersed in that thatn I am.

poettering:
any chance you can split out the build system thing, into a separate PR, so that we can discuss this separately? the other stuff looks good to be merged quickly now.

src/boot/efi/pe.c
poettering: i mean the function as a whole, not this specific line here

src/boot/efi/assert.c
medhefgo: Actually, I think it makes more sense to cast to `UINT32`.

src/boot/efi/boot.c
medhefgo: `%u` and `%lu` *are* fixed size (32 and 64 resp., I looked up the actual code in `Print.c` to make sure). You can blame gnu-efi for their weird APIs and lack of a native size indicator. And now I am somewhat inclined to just cast any INTN/UINTN to INT64/UINT64 instead of translating them to format macros to keep it simple…

src/boot/efi/boot.c
medhefgo: Yes. But we never did and I am feeling lazy. So I changed it to use `HandleProtcol` instead. :D

src/boot/efi/pe.c
medhefgo: Packing? I mostly used it cause that is how the data is arranged in the section and I'd rather be explicit about it.

src/boot/efi/pe.c
medhefgo: From what I've learned recently with the bcd test code was that the compiler seems to do the right thing when you access a (packed?) struct that's misaligned. At least I didn't get ubsan warnings for the structs, while accessing raw primitive types that way did.

Btw, the PE parsing code as a whole isn't really on the level of paranoid that it should be…

src/boot/efi/boot.c
poettering: jeezus christ, what confusion

src/boot/efi/util.h
```diff
@@ -17,6 +17,17 @@
...
+/* gnu-efi has no format specifier for (U)INTN. */
+#if __SIZEOF_POINTER__ == 4
+#  define PRIuN L"u"
+#  define PRIdN L"d"
+#elif __SIZEOF_POINTER__ == 8
+#  define PRIuN L"lu"
+#  define PRIdN L"ld"
+#else
+#  error "Unexpected pointer size"
+#endif
```
poettering: can you add a comment here, that mentions that %u is UINT32 and %lu is UINT64, and that's fixated that way and hence PRIu64 and PIRu32 and friends are not defined.

otherwise the next one looking at this might add them and then notice it's not how this works...

poettering: also this means we have no way to output an "int" or "unsigned" type right now without casting?

Maybe add:

```c
#define PRIu PRIu32
#define PRIi PRIi32
assert_cc(sizeof(unsigned) == sizeof(uint32_t));
```

or so, to make this complete?

src/boot/efi/util.h
```diff
@@ -17,6 +17,17 @@
 #define UINT64_MAX ((UINT64) -1)
 #endif
 
+/* gnu-efi has no format specifier for (U)INTN. */
+#if __SIZEOF_POINTER__ == 4
+#  define PRIuN L"u"
+#  define PRIdN L"d"
```
poettering: across our userspace codebase we use `PRIi…` almost exclusively, and `PRId…` never, even tough they are identical. Let's stick to that, and also use `PRIiN` here, instead of `PRIdN` (and not even define the other)

src/boot/efi/boot.c
```diff
@@ -2438,10 +2437,29 @@ static EFI_STATUS image_start(
         efivar_set_time_usec(LOADER_GUID, L"LoaderTimeExecUSec", 0);
         err = BS->StartImage(image, NULL, NULL);
         graphics_mode(FALSE);
```
poettering, line 2436: this loks misplaced. if we use the compat proto next, then we should only switch from graphics mode once that fails, too, no?

src/boot/efi/pe.c
poettering: well, the compiler *assumes* that the data is properly aligned of course. And intel cpus are lenient, and will allow unaligned accesses to go through, but are terrible slow for it.

Please, add some kind of check here.

src/boot/bootctl.c
```diff
@@ -2002,11 +2047,14 @@ static int verb_install(int argc, char *argv[], void *userdata) {
...
-                                      "/EFI/systemd/systemd-boot" EFI_MACHINE_TYPE_NAME ".efi",
-                                      install);
+        if (!arg_touch_variables)
+                return r;
+
+        if (arg_arch_all)
+                return log_warning_errno(r, "Not changing EFI variables with --all-architectures.");
+
+        char *path = strjoina("/EFI/systemd/systemd-boot", arch, ".efi");
+        r = install_variables(arg_esp_path, part, pstart, psize, uuid, path, install);
```
poettering: just do `return install_variables(…);`

src/boot/bootctl.c
```diff
@@ -2002,11 +2047,14 @@ static int verb_install(int argc, char *argv[], void *userdata) {
 
         (void) sync_everything();
 
-        if (arg_touch_variables)
-                r = install_variables(arg_esp_path,
-                                      part, pstart, psize, uuid,
-                                      "/EFI/systemd/systemd-boot" EFI_MACHINE_TYPE_NAME ".efi",
-                                      install);
+        if (!arg_touch_variables)
+                return r;
```
poettering: return 0;

src/boot/bootctl.c
```diff
@@ -2002,11 +2047,14 @@ static int verb_install(int argc, char *argv[], void *userdata) {
...
+
+        if (arg_arch_all)
+                return log_warning_errno(r, "Not changing EFI variables with --all-architectures.");
```
poettering: you are propagating `r` here, which has no useful value. You want to return 0 here. 

also, LOG_WARN is too loud of a log level. It's LOG_NOTICE or LOG_INFO at best, since it's nothing actionable at all.

i.e.:

```c
if (arg_arch_all) {
        log_info(…);
        return 0;
}
``

File: src/boot/bootctl.c [hunk 47]
```diff
@@ -2065,7 +2113,11 @@ static int verb_remove(int argc, char *argv[], void *userdata) {
         if (!arg_touch_variables)
                 return r;
 
-        q = remove_variables(uuid, "/EFI/systemd/systemd-boot" EFI_MACHINE_TYPE_NAME ".efi", true);
+        if (arg_arch_all)
+                return log_warning_errno(r, "Not changing EFI variables with --all-architectures.");
```
poettering: I'd really write this as:

```c
if (arg_arch_all) {
        log_warning(…);
        return r;
}
```
why? because the error that `r` might contain actually has nothing to do with this log message, it's a collected error from earlier. hence, yes, pleas propagate, but separately from the log message, as it will otherwise show up in the structure log message metadata, but it doesn't belong there.

poettering (changes requested):

src/boot/efi/pe.c
medhefgo: I was saying that despite x86 cpus supporting unaligned access, ubsan was able to detect those. So I am assuming that it's a non-issue for (packed) structs; the compiler does some magic there.

Nevertheless, having an explicit check doesn't hurt.

src/boot/efi/boot.c
medhefgo: This is deliberate to keep things simple. And it shouldn't really matter anyways.

medhefgo:
I've done away with the format macros in favor of just casting. The `Print()` API is just a mess and since you technically need to deal with `int/unsigned/(U)INT{32,64}` differently, casting seems so much nicer to me. Also, I hate int promotion. Lemme know what you think.

medhefgo:
(if you haven't noticed, I switched it back to PRI macros)

src/boot/efi/boot.c
poettering: MacMode is INT32, hence why subtract INT64_C(1) from it? And why print as `%ld` instead of plain `%d`?

src/boot/efi/util.c
```diff
@@ -57,7 +57,7 @@ EFI_STATUS efivar_set_uint_string(const EFI_GUID *vendor, const CHAR16 *name, UI
         assert(vendor);
         assert(name);
 
-        SPrint(str, ELEMENTSOF(str), L"%u", i);
+        ValueToString(str, FALSE, i);
```
poettering: hmm, this looks wrong. ValueToString() takes a signed integer... but this one is about unsigned

src/boot/efi/util.c
```diff
@@ -236,7 +236,7 @@ void efivar_set_time_usec(const EFI_GUID *vendor, const CHAR16 *name, UINT64 use
         if (usec == 0)
                 return;
 
-        SPrint(str, ELEMENTSOF(str), L"%ld", usec);
+        ValueToString(str, FALSE, usec);
```
poettering: ditto

src/boot/efi/boot.c
```diff
@@ -2301,27 +2301,63 @@ static void config_load_xbootldr(
 static EFI_STATUS initrd_prepare(
                 EFI_FILE *root,
                 const ConfigEntry *entry,
-                CHAR16 **ret_options) {
+                CHAR16 **ret_options,
+                UINT8 **ret_initrd,
```
poettering: void**?

I.e. we generally use `void*` as type for "arbitrary data"...

src/boot/efi/pe.c
```diff
@@ -160,6 +166,52 @@ static void locate_sections(
...
+
+        typedef struct {
+                UINT8 type;
+                UINT8 size;
+                UINT16 machine_type;
+                UINT32 entry_point;
+        } _packed_ LinuxPeCompat1;
+
+        while (size >= sizeof(LinuxPeCompat1)) {
+                LinuxPeCompat1 *compat = (LinuxPeCompat1 *) ((UINT8 *) dos + addr);
```
poettering, line 195: according to C specs you aren't even allowed to to *take* an unaligned pointer of a type, or you are already in undefined behaviour territory. (i.e. the compiler may assume that a pointer of type LinuxPeCompat* is always aligned, and hence the check below is something it can optimize away, since it already knows it is, it after all has the right type).



i.e. you need to check for alignment *before* you cast to the struct, not after.

poettering:
looks good, just some minor stuff left. 

(sorry for the slow review!)

src/boot/efi/util.c
medhefgo: This is exactly what `SPrint` does internally. It does not give a fuck about the sign and will pass UINT32/64 to the INT64 arg of `ValueToString`. Which is why `Print(L"%lu", UINTN_MAX)` gives you `-1`.

I know this is messed up. But I think it's better to call this directly rather than indirectly through the Print machinery.

poettering: please add inline comment about this.

src/boot/efi/boot.c
medhefgo: When I introduced this line you got paranoid about potential underflows should a crazy firmware give us a INT32_MIN as MaxMode. This is just continuing to cover our asses (and I didn't know at the time that `_C` macros existed or I would've used that). :P

src/boot/efi/pe.c
```diff
@@ -160,6 +166,49 @@ static void locate_sections(
...
+                return 0;
+
+        typedef struct {
+                UINT8 type;
+                UINT8 size;
+                UINT16 machine_type;
+                UINT32 entry_point;
+        } _packed_ LinuxPeCompat1;
+
+        while (size >= sizeof(LinuxPeCompat1) && addr % __alignof__(LinuxPeCompat1 *) == 0) {
```
poettering: actually, must be `__alignof__(LinuxPeCompat1)`, i.e. of the struct, not the pointer.

might make sense to wrap is in a helper macro, something like this:

```c
#define POINTER_IS_ALIGNED(p, t) ({ \
        const void *_q = (p); \
        UINTN a = (UINTN) _q; \
        (a % __alignof__(t)) == 0; \
})
```

With that you could just do:

```c
while (size >= sizeof(LinuxPeCompat1) && POINTER_IS_ALIGNED((UINT8*) dos + addr, LinuxPeCompat1) {
…
``

medhefgo: Huh. Interestingly, alignof the _packed_ struct is 1, while 4 without. Does checking really make sense then?? (Aka, it confirms my assumtion about the compiler doing the right thing)

poettering: Yeah, the _packed_ removes the alignment requirements, that's why I asked about that earlier.

But I'd still do the check, since in general we should check alignment before assuming too much from arbitrary pointers.

poettering: alternatively, add an `assert_cc(__alignof__(LinuxPeCompat1) == 1)` before casting into a pointer of that type. That documents clearly that the assumption here is that any pointer is safe for this structure.

medhefgo:
> lgtm, but let's wait for the next cycle, I feel a bit uncomfortable missing something that changes the boot path so much this late in the cycle given we have no testsuite for sd-boot right now... hope that makes sense?

I don't mind, but I think at least the format specifier and devtree commits should go in now as they're fixing bugs (that nobody has complained about yet). Should I make a PR for them or you gonna cherry pick them?

poettering:
hmm, cherry pick isn't straight-forward unfortunately, patches don't apply on git main.

Would be thankful for a split-out PR of those commits. Thanks!

josephlr:
@medhefgo I was going to do some TPM2 related testing w/ this. Could you rebase onto latest `main` now that #23073 is merged? I was able to do the rebase locally, but it didn't apply cleanly.

bluca:
This is a couple of weeks old and there were a few build changes for sd-boot, so I've rebased to make it go through up-to-date CI, just in case

src/boot/efi/meson.build
Viking8: Super late 2 cents, but the Dell Venue 10 Pro 5055 tablet has an 64-bit Atom CPU with 32-bit UEFI. I'm trying to keep it alive by being able to boot a 64-bit Linux seeing as a lot of software has moved to 64-bit only. I managed to get Debian 12 x64 to boot from internal storage, but it ran VERY poorly, and some things didn't work at all (screen brightness settings for instance). When adding a bootia32.efi file to a Linux Mint 21.1 USB, I can boot into the live environment and install, but nothing I do allows me to boot from internal storage. The live usb for LM ran so much faster than the installed debian 12 and there were no hardware issues either.

mirh: Code review is def not the place for technical support, let alone when you bring in problems that are obviously distro-specific.

Viking8: @mirh I didn't ask for any tech support. I was just relaying some info about a device that uses a 64-bit CPU and a 32-bit UEFI. The rest was a little back story in case someone was still working on this so they would know a little about what I did.

PR #22551, opened 2022-02-17
mkosi: Add more packages required to run systemd-networkd-tests.py

daandemeyer:
Let's also add the required packages to run systemd-networkd-tests.py
for Ubuntu and Debian. The Fedora and Arch lists are also updated to
include python which is also required to run the tests.

### DIFF
--- a/mkosi.default.d/debian/10-mkosi.debian
+++ b/mkosi.default.d/debian/10-mkosi.debian
@@ -73,3 +73,9 @@ Packages=
         locales
         nano
         strace
+        # Required to run systemd-networkd-tests.py
+        python3
+        iproute2
+        dnsmasq-base
+        wireguard-tools
+        isc-dhcp-server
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
I had python configured to be installed in a local mkosi config so I missed it in the previous PR

yuwata:
```
Package python is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python-is-python3 2to3 python2-minimal python2 dh-python

E: Package 'python' has no installation candidate
```

mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -73,3 +73,9 @@ Packages=
         locales
         nano
         strace
+        # Required to run systemd-networkd-tests.py
+        python
```
bluca: here and in ubuntu it would be python3

daandemeyer: I figured there was a 50% chance of being right so I decided to let CI figure it out for me :p

yuwata (approved):
LGTM.

bluca (approved):

DaanDeMeyer:
All the mkosi CIs passed so let's merge

PR #22552, opened 2022-02-17
fix nss deadlock between dbus-daemon and PID 1

poettering:
This is more comprehensive fix along the lines of #22038 

In the long run we should get rid of the remaining synchronous calls into dbus from PID 1 too. This starts with one of them, but there are a bunch more, mostly around sd_bus_creds stuff, i.e. identifying peers on the bus. Once we addressed all those the special hack via the env var can go away.

Fixes: #15316
Replaces: #22038

All credit for tracking down the deadlock and prepping the original patch should go to @slyon

### DIFF
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -1681,7 +1681,7 @@ static int service_spawn(
                 return -ENOMEM;
 
         /* System D-Bus needs nss-systemd disabled, so that we don't deadlock */
-        SET_FLAG(exec_params.flags, EXEC_NSS_BYPASS_BUS,
+        SET_FLAG(exec_params.flags, EXEC_NSS_DYNAMIC_BYPASS,
                  MANAGER_IS_SYSTEM(UNIT(s)->manager) && unit_has_name(UNIT(s), SPECIAL_DBUS_SERVICE));
 
         strv_free_and_replace(exec_params.environment, final_env);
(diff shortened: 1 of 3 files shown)


src/core/service.c
```diff
@@ -4322,6 +4324,53 @@ static int service_get_timeout(Unit *u, usec_t *timeout) {
...
+        Service *s;
+        int r;
+
+        assert(reply);
+        assert(u);
+
+        s = SERVICE(u);
+        s->bus_name_pid_lookup_slot = sd_bus_slot_unref(s->bus_name_pid_lookup_slot);
+
+        if (!s->bus_name ||
```
bluca: This check was not immediately obvious to me, and initially I thought you meant `!pid_is_valid()`, would be nice to have a small comment

bluca:
Maybe add to the last commit message: `Co-authored-by: Lukas Märdian <slyon@ubuntu.com>` so that github does full accreditation?

bluca (approved):

poettering:
Made the suggested changes. Upgrading green label.

slyon:
Thank you very much!
I was able to confirm (with another team inside Canonical) that the problem is now gone for good.
This has been a long standing issue and I'm happy that it is now understood and we have a solution in place.

Also, kudos to @nowrep who came up with the initial suggestion for a fix!

Issue #13667, opened 2019-09-26
Service fails to stop when state=restarting and RestartSec=0


filbranden:
**systemd version the issue has been seen with**
> v242

**Used distribution**
 > CentOS 7

**Expected behaviour you didn't see**
 > Service would stop (during container shutdown.)

**Unexpected behaviour you saw**
 > Service was restarting in a loop, would never stop.

**Steps to reproduce the problem**

@ClydeByrdIII just uncovered this problem, container fails to shutdown and looking at the journal logs this is what happens right after shutdown is requested:

```
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Main process exited, code=killed, status=1/HUP
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Succeeded.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Consumed 6ms CPU time.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Service has no hold-off time (RestartSec=0), scheduling restart.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Scheduled restart job, restart counter is at 1.
Sep 26 16:12:14 mycontainer systemd[1]: Stopped Getty on tty1.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Consumed 6ms CPU time.
Sep 26 16:12:14 mycontainer systemd[1]: Started Getty on tty1.
Sep 26 16:12:14 mycontainer login[43]: pam_systemd(login:session): Failed to create session: VT number out of range
Sep 26 16:12:14 mycontainer login[43]: pam_unix(login:session): session opened for user root by LOGIN(uid=0)
Sep 26 16:12:14 mycontainer login[43]: ROOT LOGIN ON console
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Main process exited, code=killed, status=1/HUP
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Succeeded.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Consumed 6ms CPU time.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Service has no hold-off time (RestartSec=0), scheduling restart.
Sep 26 16:12:14 mycontainer systemd[1]: getty@tty1.service: Stop job pending for unit, delaying automatic restart.
[...]

PR #22561, opened 2022-02-18
core: skip automatic restart when a JOB_STOP is pending

fbuihuu:
It's not clear why we rescheduled a service auto restart while a stop job for
the unit was pending. The comment claims that the unit shouldn't be restarted
but the code did reschedule an auto restart meanwhile.

In practice that was rarely an issue because the service waited for the next
auto restart to be rescheduled, letting the queued stop job to be proceed and
service_stop() to be called preventing the next restart to complete.

However when RestartSec=0, the timer expired right away making PID1 to
reschedule the unit again, making the timer expired right away... and so
on. This busy loop prevented PID1 to handle any queued jobs (and hence giving
no chance to the start rate limiting to trigger), which made the busy loop last
forever.

This patch breaks this loop by skipping the reschedule of the unit auto restart
and hence not depending on the value of u->restart_usec anymore.

### DIFF
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -2371,12 +2371,7 @@ static void service_enter_restart(Service *s) {
 
         if (unit_has_job_type(UNIT(s), JOB_STOP)) {
                 /* Don't restart things if we are going down anyway */
-                log_unit_info(UNIT(s), "Stop job pending for unit, delaying automatic restart.");
-
-                r = service_arm_timer(s, usec_add(now(CLOCK_MONOTONIC), s->restart_usec));
-                if (r < 0)
-                        goto fail;
-
+                log_unit_info(UNIT(s), "Stop job pending for unit, skipping automatic restart.");
                 return;
         }
 


fbuihuu:
It might fix issue #13667 hence Cc @filbranden.

fbuihuu:
@filbranden if you can still reproduce the issue you reported, it might be worth to try this patch.

filbranden:
> @filbranden if you can still reproduce the issue you reported, it might be worth to try this patch.

Thanks for tagging me! I lost most of the context of that issue, but I see there's a reproducer in [this comment](https://github.com/systemd/systemd/issues/13667#issuecomment-535728152), you might want to try that to see if you can reproduce it that way and whether the patch fixes that isuse.

(That's what I would do to confirm this fixes that issue, but I guess you might be able to do the same and you might have a test environment more handily available than I do, these days... Cheers!)

fbuihuu:
This [test case](https://github.com/systemd/systemd/issues/13667#issuecomment-535728152) is fixed by this patch. However I wasn't so sure about the initial use case you reported which seemed slightly different.

Anyway I updated the commit message to tag the patch as a fix for issue #13667.

poettering:
looks ok.

Issue #20668, opened 2021-09-08
TPM2 primary key should not have dictionary attack protection enabled


anssih:
**systemd version the issue has been seen with**

master (7728f6aa812f1af5)

**Used distribution**

Buildroot 2021.02.3 - should be reproducible anywhere

**Linux kernel version used** (`uname -a`)

`5.4.123`

**CPU architecture issue was seen on**

x86_64

**Expected behaviour you didn't see**

I expected that the systemd TPM2 primary key would not have dictionary attack protection enabled.

TPM provides a dictionary attack (DA) protection mechanism to prevent exhaustive searches of authorization values (e.g. passwords). The mechanism limits the amount of tries within a time period, causing TPM to enter DA lockout mode when this is exceeded which disallows further tries (TPM spec 2.0, r1p59 Part 1 sec 19.8 Dictionary Attack Protection).

The associated TPM failedTries counter will also be increased on an unclean system shutdown (19.8.6 Non-orderly Shutdown) as a properly timed shutdown could otherwise be used to prevent the counter increment after a real failed authentication attempt.

Depending on TPM configuration, the DA lockout mode may be lifted after a time period or it may require further explicit recovery action (see e.g. tpm2-software/tpm2-tools#1956).

In DA lockout mode, `tpm2_seal()` and `tpm2_unseal()` will fail in `sym_Esys_Create()` and `sym_Esys_Load()` with:

    tpm:warn(2.0): authorizations for objects subject to DA protection are not allowed at this time because the TPM is in DA lockout mode

The authValues of objects automatically get DA protection unless the caller opts out. tpm2-util creates the primary key with userWithAuth attribute but does not opt out of DA protection using noDA.

However, the authValue that is used for the primary key is always zero (primary_sensitive.sensitive.userAuth) as we rely on the PCR policy hash for protection instead, so dictionary attack protection for the authValue does not make sense as the attacker already knows it to be zero.

It seems also to not be generally desirable to have e.g. storage encryption stop working if the user fails an unrelated authentication prompt, or if the system is power-cycled enough times (e.g. several systems I checked go to lockout mode at 3 tries or power-cycles, with 1 try restored after every 1000 seconds - another system was more generous with 31 tries 600 seconds). There may be use cases where that is desirable, but such behavior should not be the default or at least not an **undocumented** default.

The TPM specification itself says (r1p59 Part 1 page 133, sec 19.8):

[...]

PR #22563, opened 2022-02-18
sd-cryptenroll TPM2 PIN protected unlock

### LINKED ISSUE #19229: RFE: Allow passphrase in addition to TPM2 sealed secret in cryptsetup
I'd like to be able to use both a TPM2 sealed secret and a passphrase to unlock encrypted volumes. The TPM enhances security by making the volume passphrase uncloneable and high-entropy, but if e.g. a laptop is stolen, the volume may end up accessible with no further hurdles. An additional passphrase can help remedy this with very little overhead. Passphrases with low entropy (e.g. a numeric PIN) can be feasible here due to the TPM2 dictionary attack lockout mechanism.

I'd like to add an optional --tpm2-auth parameter to systemd-cryptenroll and systemd-cryptsetup which adds a password prompt and includes this password in a TPM2_PolicyAuthValue component to the HMAC secret policy. In this case the policy might be a PolicyAnd(PolicyAuth, PolicyPCR).

With the current choice of tpm2 and passphrase authorization, one can have either the "something you know" or the "something you have" kind of authorization, but not both at the same time. FIDO keys are similar in this regard. There is currently no way I know to achieve this 2-factor model. This change would be a simple and backward-compatible way to allow both.

248

I'm working on a PR for this change, but would be happy to receive comments and guidance.

grigorig:
This adds an optional PIN/password protection for TPM2 unlock. The basic operation is simple: We add an authValue to the keyed hash object when sealing it, and the policy is extended to check the authValue in addition to PCRs. When unsealing, the object is associated with the authValue after it is loaded. LUKS2 metadata is extended with a "tpm2-pin" bool to signal that the modified policy shall be used.

cryptenroll/cryptsetup tools are extended to query PINs for enrollment and unlock, as needed. Non-interactive testing is possible with the new `PIN` and `NEWPIN` environment variables. 

Out of scope for now:
* Parameter encryption to secure the PIN being transmitted (and of course the key transmitted back)
* Support for credential storage
* libcryptsetup token support (it is unfinished and does not support interactive PIN entry, similar to PKCS#11 or FIDO2)
* Support for PINs in sd-repart

This has been tested with the swtpm TPM emulator and a discrete Nuvoton TPM. Integration tests are included, using qemu together with swtpm.

Fixes #19229.

### DIFF
--- a/test/TEST-70-TPM2/test.sh
+++ b/test/TEST-70-TPM2/test.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+
+TEST_DESCRIPTION="cryptenroll/cryptsetup with TPM2 devices"
+IMAGE_NAME="tpm2"
+TEST_NO_NSPAWN=1
+TEST_REQUIRE_INSTALL_TESTS=0
+
+# shellcheck source=test/test-functions
+. "${TEST_BASE_DIR:?}/test-functions"
+
+command -v swtpm >/dev/null 2>&1 || exit 0
+command -v tpm2_pcrextend >/dev/null 2>&1 || exit 0
+
+test_append_files() {
+    (
+        local workspace="${1:?}"
+
+        instmods tpm tpm_tis tpm_ibmvtpm
+        install_dmevent
+        generate_module_dependencies
+        inst_binary tpm2_pcrextend
+    )
+}
+
+machine="$(uname -m)"
+tpmdevice="tpm-tis"
+if [ "$machine" = "ppc64le" ]; then
+    # tpm-spapr support was introduced in qemu 5.0.0. Skip test for old qemu versions.
+    qemu_min_version "5.0.0" || exit 0
+    tpmdevice="tpm-spapr"
+fi
+
+tpmstate=$(mktemp -d)
+swtpm socket --tpm2 --tpmstate dir="$tpmstate" --ctrl type=unixio,path="$tpmstate/sock" &
+trap 'kill %%; rm -rf $tpmstate' SIGINT EXIT
+QEMU_OPTIONS="-chardev socket,id=chrtpm,path=$tpmstate/sock -tpmdev emulator,id=tpm0,chardev=chrtpm -device $tpmdevice,tpmdev=tpm0"
+
+do_test "$@"
--- a/test/units/testsuite-70.service
+++ b/test/units/testsuite-70.service
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=TEST-70-TPM2
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/systemd/tests/testdata/units/%N.sh
--- a/test/units/testsuite-70.sh
+++ b/test/units/testsuite-70.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
+
+export SYSTEMD_LOG_LEVEL=debug
+
+
+# Prepare fresh disk image
+img="/var/tmp/test.img"
+dd if=/dev/zero of=$img bs=1024k count=20 status=none
+echo -n passphrase >/tmp/passphrase
+cryptsetup luksFormat -q --use-urandom $img /tmp/passphrase
+
+# Enroll unlock with default PCR policy
+env PASSWORD=passphrase systemd-cryptenroll --tpm2-device=auto $img
+/usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1
+/usr/lib/systemd/systemd-cryptsetup detach test-volume
+
+# Check with wrong PCR
+tpm2_pcrextend 7:sha256=0000000000000000000000000000000000000000000000000000000000000000
+/usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1 && { echo 'unexpected success'; exit 1; }
+
+# Enroll unlock with PCR+PIN policy
+systemd-cryptenroll --wipe-slot=tpm2 $img
+env PASSWORD=passphrase NEWPIN=123456 systemd-cryptenroll --tpm2-device=auto --tpm2-with-pin=true $img
+env PIN=123456 /usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1
+/usr/lib/systemd/systemd-cryptsetup detach test-volume
+
+# Check failure with wrong PIN
+env PIN=123457 /usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1 && { echo 'unexpected success'; exit 1; }
+
+# Check failure with wrong PCR (and correct PIN)
+tpm2_pcrextend 7:sha256=0000000000000000000000000000000000000000000000000000000000000000
+env PIN=123456 /usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1 && { echo 'unexpected success'; exit 1; }
+
+# Enroll unlock with PCR 0+7
+systemd-cryptenroll --wipe-slot=tpm2 $img
+env PASSWORD=passphrase systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=0+7 $img
+/usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1
+/usr/lib/systemd/systemd-cryptsetup detach test-volume
+
+# Check with wrong PCR 0
+tpm2_pcrextend 0:sha256=0000000000000000000000000000000000000000000000000000000000000000
+/usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1 && exit 1
+
+echo OK >/testok
+
+exit 0
--- a/test/TEST-70-TPM2/Makefile
+++ b/test/TEST-70-TPM2/Makefile
@@ -1,0 +1,6 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+all setup run clean clean-again:
+	@TEST_BASE_DIR=../ ./test.sh --$@
+
+.PHONY: all setup run clean clean-again
--- a/test/test-functions
+++ b/test/test-functions
@@ -1213,7 +1213,7 @@ install_missing_libraries() {
     local lib path
     # A number of dependencies is now optional via dlopen, so the install
     # script will not pick them up, since it looks at linkage.
-    for lib in libcryptsetup libidn libidn2 pwquality libqrencode tss2-esys tss2-rc tss2-mu libfido2 libbpf libelf libdw; do
+    for lib in libcryptsetup libidn libidn2 pwquality libqrencode tss2-esys tss2-rc tss2-mu tss2-tcti-device libfido2 libbpf libelf libdw; do
         ddebug "Searching for $lib via pkg-config"
         if pkg-config --exists "$lib"; then
                 path="$(pkg-config --variable=libdir "$lib")"


src/shared/tpm2-util.c
```diff
@@ -813,6 +835,12 @@ int tpm2_seal(
                 .size = sizeof(hmac_sensitive.sensitive),
                 .sensitive.data.size = 32,
         };
+        if (pin) {
+                size_t pin_len = strlen(pin);
+                assert(pin_len <= 32);
```
poettering: given this is user controlled, should not be an assert(), but just return EINVAL or so here if not met.

poettering: where does the 32 come from? `sizeof(hmac_sensitive.sensitive.userAuth.buffer)` is larger than that, no?

src/shared/tpm2-util.c
```diff
@@ -1010,6 +1039,15 @@ int tpm2_unseal(
                 goto finish;
         }
 
+        if (pin) {
+                TPM2B_AUTH auth = {0};
```
poettering: we use a gcc extension that allows us to drop the `0` thing, and just write `{}` for initializers here, no need to write `{0}`.

src/shared/tpm2-util.c
```diff
@@ -1010,6 +1039,15 @@ int tpm2_unseal(
...
+                size_t pin_len = strlen(pin);
+                assert(pin_len <= 32);
```
poettering: as above, should really return EINVAL, not hit an assert.

src/shared/tpm2-util.c
```diff
@@ -1010,6 +1039,15 @@ int tpm2_unseal(
...
+                memcpy(auth.buffer, pin, pin_len);
+                auth.size = pin_len;
+                sym_Esys_TR_SetAuth(c.esys_context, hmac_key, &auth);
```
poettering: misses error checking?

src/shared/tpm2-util.c
```diff
@@ -1034,7 +1034,10 @@ int tpm2_unseal(
                         &public,
                         &hmac_key);
         if (rc != TSS2_RC_SUCCESS) {
-                r = log_error_errno(SYNTHETIC_ERRNO(ENOTRECOVERABLE),
+                // If we're in lockout mode, we should see a lockout error here,
+                // which we need to translate for the caller.
```
poettering: as per coding style, we use `/* */` style comments for explanatory/documentation stuff, and `//` for FIXME-style comments. This is the former, hence please use `/* */`

src/shared/tpm2-util.c
```diff
@@ -1034,7 +1034,10 @@ int tpm2_unseal(
...
+                int e = (rc == TPM2_RC_LOCKOUT) ? ENOLCK : ENOTRECOVERABLE;
+                r = log_error_errno(SYNTHETIC_ERRNO(e),
```
poettering: no need to use a temporary variable `e`, you only use it once anyway.

src/shared/tpm2-util.c
```diff
@@ -1034,7 +1034,10 @@ int tpm2_unseal(
                         &public,
                         &hmac_key);
         if (rc != TSS2_RC_SUCCESS) {
-                r = log_error_errno(SYNTHETIC_ERRNO(ENOTRECOVERABLE),
+                // If we're in lockout mode, we should see a lockout error here,
+                // which we need to translate for the caller.
+                int e = (rc == TPM2_RC_LOCKOUT) ? ENOLCK : ENOTRECOVERABLE;
```
poettering: the `()` are unnecessary

src/shared/tpm2-util.c
```diff
@@ -1034,7 +1034,10 @@ int tpm2_unseal(
...
+                r = log_error_errno(SYNTHETIC_ERRNO(e),
                                     "Failed to load HMAC key in TPM: %s", sym_Tss2_RC_Decode(rc));
```
poettering: maybe also adjust the human readable message in this case and explicitly say it's about lockout

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -2,6 +2,8 @@
 
 #include "alloc-util.h"
 #include "cryptenroll-tpm2.h"
+#include "ask-password-api.h"
+#include "env-util.h"
```
poettering, line 6: please keep includes sorted alphabeticlly

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,88 @@ static int search_policy_hash(
         return -ENOENT; /* Not found */
 }
 
+static int get_pin(char **ret_pin_str, int *ret_flags) {
+        char *pin_str = NULL;
+        int r = 0;
+        int flags = 0;
```
poettering: merge these two lines?

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,88 @@ static int search_policy_hash(
...
+
+                assert_se(unsetenv_erase("NEWPIN") >= 0);
+        } else {
+                for (int i = 5;; i--) {
+                        size_t pin_len;
+
+                        if (i <= 0) {
+                                return log_error_errno(
+                                        SYNTHETIC_ERRNO(ENOKEY), "Too many attempts, giving up.");
+                        }
```
poettering: no {} around single line if blocks please, see coding style

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,88 @@ static int search_policy_hash(
...
+
+                        pin = strv_free_erase(pin);
+                        r = ask_password_auto(
+                                "Please enter TPM2 PIN:",
+                                "drive-harddisk",
+                                NULL,
+                                "tpm2-pin",
+                                "cryptsetup.tpm2-pin",
```
poettering: cryptsetup → cryptenroll

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,88 @@ static int search_policy_hash(
...
+                                0,
+                                &pin);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to ask for user pin: %m");
+                        assert(strv_length(pin) == 1);
+
+                        /* Enforce a PIN length of at least 4 characters (fewer than that provides no serious
+                         * protection) and 32 characters (maximum of TPM2 authValue). */
+                        pin_len = strlen(*pin);
+                        if (pin_len < 4 || pin_len > 32) {
```
poettering: maybe encapsulate that in a simple helper `tpm2_valid_pin()` or so that does this check, and is used here and everywhere else where we need to superficially check if a password is suitable as TPM2 PIN

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,88 @@ static int search_policy_hash(
...
+                                log_error("Incorrect PIN length (must be 4-32 characters)!");
+                                continue;
+                        }
+
+                        r = ask_password_auto(
+                                "Please enter TPM2 PIN (repeat):",
+                                "drive-harddisk",
+                                NULL,
+                                "tpm2-pin",
+                                "cryptsetup.tpm2-pin",
```
poettering: as above

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,88 @@ static int search_policy_hash(
         return -ENOENT; /* Not found */
 }
 
+static int get_pin(char **ret_pin_str, int *ret_flags) {
+        char *pin_str = NULL;
+        int r = 0;
+        int flags = 0;
+        const char *e;
+        _cleanup_strv_free_erase_ char **pin = NULL;
+        _cleanup_strv_free_erase_ char **pin2 = NULL;
```
poettering: merge these two?

grigorig:
> How difficult is it to set this up? Is it supported natively in Ubuntu?

You are talking about swtpm for testing? That is pretty simple. Start swtpm, and make the emulated TPM appear as a real device with the VTPM proxy kernel module. I use [this script](https://gist.github.com/grigorig/dec727e84524e59401f5d82d91566c4a) to start and initialize swtpm.

bluca:
> > How difficult is it to set this up? Is it supported natively in Ubuntu?
> 
> You are talking about swtpm for testing? That is pretty simple. Start swtpm, and make the emulated TPM appear as a real device with the VTPM proxy kernel module. I use [this script](https://gist.github.com/grigorig/dec727e84524e59401f5d82d91566c4a) to start and initialize swtpm.

Yes - where do you get swtpm from? Is it packaged in Ubuntu?

grigorig:
> Yes - where do you get swtpm from? Is it packaged in Ubuntu?

Sorry, I don't think so. But it should be easy to build, just get it from https://github.com/stefanberger/swtpm?

grigorig:
I've been using version 0.7.0 of swtpm, but looking at the changelog, version 0.6.1 which the PPA provides should be fine too.

src/shared/tpm2-util.c
```diff
@@ -1264,6 +1264,7 @@ int tpm2_make_luks2_json(
                 size_t blob_size,
                 const void *policy_hash,
                 size_t policy_hash_size,
+                int flags,
```
poettering: having `signed` flags is a bit weird, due to sign extension and stuff. We usually do flags fields as enum these days, or at least as "unsigned" or "uint32_t" or so.

src/shared/tpm2-util.c
```diff
@@ -1304,7 +1306,9 @@ int tpm2_make_luks2_json(
                                        JSON_BUILD_PAIR("tpm2-pcrs", JSON_BUILD_VARIANT(a)),
                                        JSON_BUILD_PAIR_CONDITION(!!tpm2_pcr_bank_to_string(pcr_bank), "tpm2-pcr-bank", JSON_BUILD_STRING(tpm2_pcr_bank_to_string(pcr_bank))),
                                        JSON_BUILD_PAIR_CONDITION(!!tpm2_primary_alg_to_string(primary_alg), "tpm2-primary-alg", JSON_BUILD_STRING(tpm2_primary_alg_to_string(primary_alg))),
-                                       JSON_BUILD_PAIR("tpm2-policy-hash", JSON_BUILD_HEX(policy_hash, policy_hash_size))));
+                                       JSON_BUILD_PAIR("tpm2-policy-hash", JSON_BUILD_HEX(policy_hash, policy_hash_size)),
+                                       JSON_BUILD_PAIR_CONDITION(use_pin, "tpm2-flags", JSON_BUILD_ARRAY(JSON_BUILD_STRING("pin"))))
```
poettering: hmm, i'd just serialization that as bool. i.e.

```c
JSON_BUILD_PAIR_CONDITION(use_pin, "tpm2-flags", JSON_BUILD_BOOLEAN(true)),
```

i.e. if we have high-level, friendly bools anyway no need to group thos together with likely unrelated flags.

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -64,7 +119,58 @@ int acquire_tpm2_key(
...
+                                return log_error_errno(SYNTHETIC_ERRNO(EPERM), "Failed to unlock, giving up!");
+
+                        r = get_pin(&pin_str);
+                        if (r == -EPERM)
+                                continue;
+                        if (r < 0)
+                                return r;
+
+                        r = tpm2_unseal(
+                                device,
```
poettering, line 139: our usual coding style suggests to indent line broken parameter lists to 16ch, not 8ch, to make them visually distinct from code block indentation

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -64,7 +119,58 @@ int acquire_tpm2_key(
...
+                        pcr_bank,
+                        primary_alg,
+                        blob,
+                        blob_size,
+                        policy_hash,
+                        policy_hash_size,
+                        NULL,
+                        ret_decrypted_key,
+                        ret_decrypted_key_size);
+        }
```
poettering, line 158: no {} around single-line if blocks

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -64,7 +119,58 @@ int acquire_tpm2_key(
...
+                        if (r == -EPERM)
+                                return r;
+                        if (r == -ENOLCK)
+                                return log_error_errno(
+                                        r,
+                                        "TPM2 is in dictionary attack lockout, no unlock possible right now.");
+                        if (r < 0)
+                                continue;
+
+                        return r;
```
poettering, line 153: so this generates log messages on some errors, but not on others.

Usually our coding style says that functions should either log on all error conditions or on none (in which case the caller should do it). This function has some error paths that log and others which do not. 

See https://systemd.io/CODING_STYLE about this.

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -196,6 +304,31 @@ int find_tpm2_auto_data(
...
+                                        "TPM2 token data flags field is not an array.");
+                        }
+                        JSON_VARIANT_ARRAY_FOREACH(e, w) {
+                                const char *fs = NULL;
+                                if (!json_variant_is_string(e))
+                                        return log_error_errno(
+                                                SYNTHETIC_ERRNO(EINVAL),
+                                                "TPM2 token data flag is not a string.");
+                                fs = json_variant_string(e);
+                                if (!fs)
```
poettering: given you already check that it's a string just a few lines up this cannot fail. hence use assert() or so for this.

man/systemd-cryptenroll.xml
```diff
@@ -299,6 +299,17 @@
         signatures likely will validate against pre-existing certificates.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--tpm2-with-pin=</option><replaceable>BOOL</replaceable></term>
+
+        <listitem><para>When enrolling a TPM2 device, controls whether to require the user to enter a PIN
+        when unlocking the volume in addition to PCR binding, based on TPM2 policy authentication. The pin
+        needs to be 4-32 characters long. Defaults to <literal>no</literal>. Note that incorrect PIN entry
```
poettering: maybe say explicitly that "despite the concept being named PIN this may include any character, not just numbers".

man/systemd-cryptenroll.xml
```diff
@@ -299,6 +299,17 @@
...
+        when unlocking increments the dictionary attack lockout mechanism, and may lock out users for a
```
poettering: maybe say that this is a mechanism of the TPM2 chip, and not of our tools here. otherwise people will look for a config option in our stuff to configure that.

poettering (changes requested):

poettering:
having a test suite for the TPM2 stuff would be phenomenal, of course.

src/shared/tpm2-util.c
grigorig: Hm, my idea is that we could have a generic list of flags that would influence the TPM policy. If you don't like it that's OK. I was thinking of issue #20668 for instance. We could add a new flag keyword to signal that the primary key should be made with DA protection disabled. If I make this a top-level flag instead I should probably rename it, too.

man/systemd-cryptenroll.xml
grigorig: Good point. Maybe we should also point to the tpm-tss2 tools that can be used to change the settings?

src/shared/tpm2-util.c
grigorig: The [TPM 2.0 Structures Specification](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf) says in 10.4.5 that the authValue must be "no larger than the
largest digest produced by a TPM". As far as I know a TPM 2.0 must support at least SHA-256 which has 256 bit / 32 bytes digest size, but SHA-384 etc. are optional. On top the spec recommends the max. length should match the digest used by the object, which is SHA-256. But yeah this is very obtuse, I'll add some comment.

src/shared/tpm2-util.c
poettering: i'd always normalize those things into separate bools. flags are fine, in C code or in binary structs. But in JSON? they have native booleans, I'd always go for those...

man/systemd-cryptenroll.xml
poettering: yeah, that can't hurt.

src/cryptsetup/cryptsetup-tpm2.c
grigorig: OK. Looks like clang-format won't do that?

grigorig: Looks like clang-format can do that with `ContinuationIndentWidth`. Maybe we could set this to 16 chars in `.clang-format`?

poettering: By all means, add a patch to this PR that adds it if we are missing that.

We are currently not using clang-format automatically, since it couldn't really format things for us the way we wanted. But I am sure it's improving, so we might have missed some settings.

ideally, one day it's good enough to be something we can do automatically and enforce, at which point we should hopefully notice mssing tweaks like that more quickly.

src/cryptsetup/cryptsetup-tpm2.c
grigorig: OK, that makes sense (and avoids "double error messages"). The existing code isn't 100% consistent either but I'll try to make it no worse at least. :)

poettering: yeah, there are some parts of the codebase we are really bad it (usually the older parts of the codebase). But we strive to getting better at it.

bluca:
> > * Parameter encryption to secure the PIN being transmitted (and of course the key transmitted back)
> 
> This is highly relevant either way. I tried to find some prior art how this is done elsewhere, but couldn't find anything. As it appears most code that uses the intel TPM2 stack does not employ parameter encryption right now. :-(

Yes, this is a limitation of the Intel stack versus the IBM one (which supports this). But nobody uses the IBM one, and it's largely not maintained in distros. It's really a sad state, especially when you consider how much money and effort has been put into solving this problem by moving the TPM implementation to the firmware (fTPM) or to the CPU (titan, pluton)... all because the Intel stack doesn't implement the on-transit encryption part of the spec

poettering:
Hmm, so right now you pass the PIN literally to the policy. I wonder if it wouldn't be smarter to hash it first, that way the size limitations on the PIN go away. i.e. this detail of the backing technology should probably be hidden away.

we probably should also salt it if we hash it, so that if someone sees the stream of TPM2 commands they at least can't derive the literal password via rainbow tables or so from it.

Hence maybe instead of the JSON boolean field, add a subobject that for now just contains one value, the salt in base64. Place 256 bit of randomness in there, append that to the supplied password, and calculate SHA256 of it, then pass that to the TPM2 stack as PIN.

For the regular password-based unlocking we allow arbitrary keys after all, and I think given how simple this is we should try to make this behave alike.

bluca:
Interesting, I didn't know it was supported

grigorig:
Yeah looks like there is some support for parameter encryption in the tpm2-tss stack, but its a whole new can of worms.

By the way, there is a [free book](https://link.springer.com/book/10.1007%2F978-1-4302-6584-9) on TPM 2.0 available. It's still a bit fuzzy to me how parameter encryption actually works even with the details provided there nonetheless.

grigorig:
> Hmm, so right now you pass the PIN literally to the policy. I wonder if it wouldn't be smarter to hash it first, that way the size limitations on the PIN go away. i.e. this detail of the backing technology should probably be hidden away.

Yes, I agree. I didn't bother to do it to keep it simple. I'll get to it, sounds like a good and simple improvement.

> we probably should also salt it if we hash it, so that if someone sees the stream of TPM2 commands they at least can't derive the literal password via rainbow tables or so from it.
> 
> Hence maybe instead of the JSON boolean field, add a subobject that for now just contains one value, the salt in base64. Place 256 bit of randomness in there, append that to the supplied password, and calculate SHA256 of it, then pass that to the TPM2 stack as PIN.
>

Or just use HMAC? It's already ready to use in systemd utility functions.

grigorig:
By the way, I addressed a bunch of the review issues (thanks for the nice & quick review)! I just added the changes on top for now. Some of the commits I intend to squash into previous commits at the end, but not all. I'm not sure, how important is a "clean history"?

src/shared/tpm2-util.c
grigorig: Replaced it with an enum.

grigorig:
Thinking about hashing and salting once more, I don't think salting is worth the effort. If an attacker can listen to TPM interface traffic, what's the point of protecting the PIN if the LUKS encryption key is transmitted in plain text, after all? If we want to protect against interface snooping, parameter encryption seems to be the only effective way.

poettering:
> Or just use HMAC? It's already ready to use in systemd utility functions.

Sure why not. hmac using the password as key and the salt as message to authenticate sounds great and should be beyond most doubts

poettering:
> Thinking about hashing and salting once more, I don't think salting is worth the effort. If an attacker can listen to TPM interface traffic, what's the point of protecting the PIN if the LUKS encryption key is transmitted in plain text, after all? If we want to protect against interface snooping, parameter encryption seems to be the only effective way.

it's nice in case of password reuse? i.e. if the pw is typed in by humans then they might be tempted to reuse the same password on multiple disks. I mean, we kinda push people towards that with the password caching logic in systemd-cryptsetup... so if it might be good that if you listen to TPM2 traffic you will only be able to unlock one disk instead of all that use the same pw

grigorig:
> it's nice in case of password reuse? i.e. if the pw is typed in by humans then they might be tempted to reuse the same password on multiple disks. I mean, we kinda push people towards that with the password caching logic in systemd-cryptsetup... so if it might be good that if you listen to TPM2 traffic you will only be able to unlock one disk instead of all that use the same pw

Alright, that makes some sense, I didn't have this scenario in mind. If we want to employ parameter encryption (which we do? I intend to work on this as a follow-up to this PR), I still think we should keep it simple for now (i.e. just use a basic, unsalted hash, to allow long PINs to be used). The PIN salt would be superfluous in that case.

poettering:
Dunno, I'd use the HMAC stuff, simply because it is easy to do, i.e. we have most stuff in place: the hmac_sha256() call already exists, and we can store the salt nicely via the base64 helpers we already have for the JSON generator/parser.

But then again, i'd merge it with a simple unsalted sha256 hash or so, too.

grigorig:
I have a couple of tests running now, it was easier than I thought. qemu can directly communicate with swtpm through a socket, so we don't even need the vtpm proxy module. The only problem I can see is that tss2 loads some libraries at runtime, and I haven't found a good way to copy them into place in a portable way.

test/TEST-70-CRYPTENROLL-TPM2/test.sh
```diff
@@ -0,0 +1,42 @@
...
+command -v tpm2_pcrextend >/dev/null 2>&1 || exit 0
+
+test_append_files() {
+    (
+        local workspace="${1:?}"
+
+        install_dmevent
+        generate_module_dependencies
+        inst_binary tpm2_pcrextend
+        # XXX: Needs a better way to install libraries
```
bluca: the way we handle this is via pkg-config - add the required pc files to the list at https://github.com/systemd/systemd/blob/main/test/test-functions#L1216

you can find the available pc files at https://packages.ubuntu.com/focal/amd64/libtss2-dev/filelist

test/units/testsuite-70.service
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=TEST-70-CRYPTENROLL-TPM2
```
bluca: I'd just call this TEST-70-TPM2 as we'll use it for more than cryptenroll

bluca:
> I have a couple of tests running now, it was easier than I thought. qemu can directly communicate with swtpm through a socket, so we don't even need the vtpm proxy module. The only problem I can see is that tss2 loads some libraries at runtime, and I haven't found a good way to copy them into place in a portable way.

Very nice! I'm getting the swtpm package rebuilt in the ppa

test/TEST-70-CRYPTENROLL-TPM2/test.sh
grigorig: Thanks, that seems to have worked fine.

grigorig:
There are two distinct code paths, one with libcryptsetup plugins enabled, one with them disabled. This is configured with the meson toggle `libcryptsetup-plugins`. Can we test both of those configurations in CI?

bluca:
I think one such job was added recently - rebase your PR on main and it will be picked up. Also please clean up the history and squash commits where appropriate.

I have added swtpm and tpm2-tools to the package list for the focal-amd64 job and restarted it

grigorig:
Sorry, just noticed some small issue and outdated commit messages, force pushed to update that.

src/shared/tpm2-util.c
```diff
@@ -717,9 +738,20 @@ static int tpm2_make_pcr_session(
         return r;
 }
 
+static void *hash_pin(const char *pin, uint8_t ret_digest[32]) {
+        struct sha256_ctx hash;
+        size_t pin_len = strlen(pin);
```
bluca: `assert(pin)` - also add some empty lines to make it more readable

src/shared/tpm2-util.c
```diff
@@ -717,9 +738,20 @@ static int tpm2_make_pcr_session(
         return r;
 }
 
+static void *hash_pin(const char *pin, uint8_t ret_digest[32]) {
```
bluca: `ret_digest[static SHA256_DIGEST_LENGTH]`

src/shared/tpm2-util.c
```diff
@@ -1005,11 +1043,37 @@ int tpm2_unseal(
...
+                                        "Failed to load HMAC key in TPM: %s",
+                                        sym_Tss2_RC_Decode(rc));
                 goto finish;
         }
 
+        if (pin) {
+                TPM2B_AUTH auth = {};
+
+                hash_pin(pin, auth.buffer);
+                auth.size = 32;
```
bluca: initialize this parameter inline in the declaration

src/shared/tpm2-util.c
```diff
@@ -1306,3 +1370,10 @@ int tpm2_primary_alg_from_string(const char *alg) {
                 return TPM2_ALG_RSA;
         return -EINVAL;
 }
+
+bool tpm2_valid_pin(const char *pin) {
+        /* Enforce a PIN length of at least 4 characters: Fewer characters provide little protection, even
```
bluca: `assert(pin)`

src/shared/tpm2-util.h
```diff
@@ -3,6 +3,7 @@
 
 #include "json.h"
 #include "macro.h"
+#include <stdbool.h>
```
bluca: external includes in a separate block, above the internal includes

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
         return -ENOENT; /* Not found */
 }
 
+static int get_pin(char **ret_pin_str, systemd_tpm2_flags *ret_flags) {
+        char *pin_str = NULL;
+        int r = 0;
```
bluca: no need to initialize `r`

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
...
+        systemd_tpm2_flags flags = 0;
+        _cleanup_strv_free_erase_ char **pin = NULL, **pin2 = NULL;
+
+        r = getenv_steal_erase("NEWPIN", &pin_str);
+        if (r < 0)
+                return log_error_errno(r, "Failed to acquire PIN from environment: %m");
+        if (r > 0) {
```
bluca: no brackets around single line ifs

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
         return -ENOENT; /* Not found */
 }
 
+static int get_pin(char **ret_pin_str, systemd_tpm2_flags *ret_flags) {
+        char *pin_str = NULL;
+        int r = 0;
+        systemd_tpm2_flags flags = 0;
+        _cleanup_strv_free_erase_ char **pin = NULL, **pin2 = NULL;
```
bluca: this declaration can be moved down in the else block

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
         return -ENOENT; /* Not found */
 }
 
+static int get_pin(char **ret_pin_str, systemd_tpm2_flags *ret_flags) {
+        char *pin_str = NULL;
```
bluca: _cleanup_free_ and use TAKE_PTR when assigning to ret_pin_str

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
...
+                                flags |= TPM2_FLAGS_USE_PIN;
+                                break;
+                        }
+
+                        log_error("PINs didn't match, please try again!");
+                }
+        }
+
+        *ret_flags = flags;
+        *ret_pin_str = pin_str;
```
bluca: check both for NULL before assigning

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
...
+                                        "cryptenroll.tpm2-pin",
+                                        USEC_INFINITY,
+                                        0,
+                                        &pin2);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to ask for user pin: %m");
+                        assert(strv_length(pin) == 1);
+
+                        if (strv_equal(pin, pin2)) {
+                                pin_str = strdup(*pin);
```
bluca, line 112: OOM check

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,77 @@ static int search_policy_hash(
...
+                                        "drive-harddisk",
+                                        NULL,
+                                        "tpm2-pin",
+                                        "cryptenroll.tpm2-pin",
+                                        USEC_INFINITY,
+                                        0,
+                                        &pin);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to ask for user pin: %m");
+                        assert(strv_length(pin) == 1);
```
bluca, line 96: return an error instead of asserting, given it's user input?

src/shared/tpm2-util.h
```diff
@@ -5,6 +5,10 @@
 #include "macro.h"
 #include <stdbool.h>
 
+typedef enum {
```
bluca: `typedef enum tpm2_flags {` - also no need for the systemd_ prefix

src/shared/tpm2-util.h
```diff
@@ -5,6 +5,10 @@
...
+        TPM2_FLAGS_USE_PIN = 1
```
bluca: add `_TPM2_FLAGS_MAX` and `_TPM2_FLAGS_INVALID = -EINVAL`

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,60 @@
...
+#include "env-util.h"
 #include "fileio.h"
 #include "hexdecoct.h"
 #include "json.h"
 #include "parse-util.h"
 #include "random-util.h"
 #include "tpm2-util.h"
 
+static int get_pin(usec_t until, AskPasswordFlags ask_password_flags, bool headless, char **ret_pin_str) {
+        char *pin_str = NULL;
```
bluca: _cleanup_free_ and TAKE_PTR

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,60 @@
...
+        int r = 0;
```
bluca: no need to intialize

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,60 @@
...
+                        "drive-harddisk",
+                        NULL,
+                        "tpm2-pin",
+                        "cryptsetup.tpm2-pin",
+                        until,
+                        ask_password_flags,
+                        &pin);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to ask for user pin: %m");
+                assert(strv_length(pin) == 1);
```
bluca, line 43: same question about returning an error

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,60 @@
...
+
+                if (tpm2_valid_pin(*pin)) {
+                        pin_str = strdup(*pin);
```
bluca: oom check

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,60 @@
...
 #include "json.h"
 #include "parse-util.h"
 #include "random-util.h"
 #include "tpm2-util.h"
 
+static int get_pin(usec_t until, AskPasswordFlags ask_password_flags, bool headless, char **ret_pin_str) {
+        char *pin_str = NULL;
+        int r = 0;
+        _cleanup_strv_free_erase_ char **pin = NULL;
+
```
bluca, line 20: assert(ret_pin_str)

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,60 @@
...
+                if (r < 0)
+                        return log_error_errno(r, "Failed to ask for user pin: %m");
+                assert(strv_length(pin) == 1);
+
+                if (tpm2_valid_pin(*pin)) {
+                        pin_str = strdup(*pin);
+                        goto finish;
+                }
+
+                return log_error_errno(
```
bluca: there's no need for the goto, invert the if above and return this error there

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -64,7 +114,53 @@ int acquire_tpm2_key(
...
+                         * not happen, but this avoids confusing behavior, just in case. */
+                        if (IN_SET(r, -EPERM, -ENOLCK))
+                                return r;
+                        if (r < 0)
+                                continue;
+
+                        return r;
+                }
+        } else
+                return tpm2_unseal(
```
bluca: no need for an else here - to reduce indentation, invert the check and return tpm2_unseal inside the if, and have the for block after

src/cryptsetup/cryptsetup.c
```diff
@@ -1301,9 +1301,16 @@ static int attach_luks_or_plain_or_bitlk_by_tpm2(
...
                                         key_data, key_data_size,
                                         NULL, 0, /* we don't know the policy hash */
+                                        0, /* PIN is currently unhandled in this case */
+                                        until,
+                                        arg_headless,
+                                        arg_ask_password_flags,
                                         &decrypted_key, &decrypted_key_size);
                         if (r >= 0)
                                 break;
+                        if (r == -EACCES || r == -ENOLCK) {
```
bluca: `IN_SET(r, ...)`

src/cryptsetup/cryptsetup.c
```diff
@@ -1369,7 +1378,14 @@ static int attach_luks_or_plain_or_bitlk_by_tpm2(
                                                 NULL, 0, 0, /* no key file */
                                                 blob, blob_size,
                                                 policy_hash, policy_hash_size,
+                                                tpm2_flags,
+                                                until,
+                                                arg_headless,
+                                                arg_ask_password_flags,
                                                 &decrypted_key, &decrypted_key_size);
+                                if (r == -EACCES || r == -ENOLCK) {
```
bluca: `IN_SET(r, ...)`

test/TEST-70-TPM2/test.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+
+TEST_DESCRIPTION="cryptenroll/cryptsetup with TPM2 devices"
+IMAGE_NAME="cryptenrolltpm2"
```
bluca: just `tpm2`

test/TEST-70-TPM2/test.sh
```diff
@@ -0,0 +1,40 @@
...
+    (
+        local workspace="${1:?}"
+
+        install_dmevent
+        generate_module_dependencies
+        inst_binary tpm2_pcrextend
+    )
+}
+
+check_result_qemu() {
```
bluca: isn't this already the default action?

test/units/testsuite-70.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
+set -o pipefail
+
```
bluca, line 4: export `SYSTEMD_LOG_LEVEL=debug` here so that all commands log in verbose mode

bluca:
the last commit can be squashed in the one before

bluca:
@mrc0mmand can we get swtpm and tpm2-tools installed in the CentOS images please?

grigorig:
> the last commit can be squashed in the one before

Just want to make sure it actually works this time. Thanks for the review, will look at the items.

src/cryptenroll/cryptenroll-tpm2.c
grigorig: I think asserts make more sense here, no?

src/cryptenroll/cryptenroll-tpm2.c
grigorig: I adapted this from other usages of `ask_password_auto`. As far as I can tell it should always return a strv of length 1, it's not really influenced by user input.

src/shared/tpm2-util.h
grigorig: Looking through the codebase, I think i should use CamelCase for consistency, too.

src/cryptenroll/cryptenroll-tpm2.c
bluca: ok, leave it then

src/shared/tpm2-util.h
bluca: yes

src/cryptenroll/cryptenroll-tpm2.c
bluca: yeah, that's fine

test/TEST-70-TPM2/test.sh
grigorig: I just removed my own implementation, default seems fine.

grigorig:
Unfortunately the qemu version in Ubuntu focal is too old on PPC64. Support for the necessary `tpm-spapr` devices was [introduced with qemu 5.0.0](https://github.com/qemu/qemu/commit/3676bc69b358d84a6b32d9cd44325048659a32a2), while CI seems to use 4.1.0. Any suggestions how to proceed?

bluca:
Just skip the test if you detect PPC and qemu older than 5.0

mrc0mmand:
> @mrc0mmand can we get swtpm and tpm2-tools installed in the CentOS images please?

Done (https://github.com/systemd/systemd-centos-ci/pull/466).

grigorig:
@mrc0mmand Looks like the CentOS CI Arch test failed. Is there some way to look at the journal logs of the affected test unit?

mrc0mmand:
> @mrc0mmand Looks like the CentOS CI Arch test failed. Is there some way to look at the journal logs of the affected test unit?

Once it finishes it uploads all artifacts (including journals of failed tests), so they can be inspected further.

mrc0mmand:
> > @mrc0mmand Looks like the CentOS CI Arch test failed. Is there some way to look at the journal logs of the affected test unit?
> 
> Once it finishes it uploads all artifacts (including journals of failed tests), so they can be inspected further.

https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/10309/artifact//systemd-centos-ci/index.html

grigorig:
> > > @mrc0mmand Looks like the CentOS CI Arch test failed. Is there some way to look at the journal logs of the affected test unit?
> > 
> > 
> > Once it finishes it uploads all artifacts (including journals of failed tests), so they can be inspected further.
> 
> https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/10309/artifact//systemd-centos-ci/index.html

Thanks. Hm... the VM kernel seems to lack TPM support - no device is available from userspace. Maybe this system has it configured as a kernel module, I'll try to load that. But maybe you have some more insight about the CI system?

mrc0mmand:
> > > > @mrc0mmand Looks like the CentOS CI Arch test failed. Is there some way to look at the journal logs of the affected test unit?
> > > 
> > > 
> > > Once it finishes it uploads all artifacts (including journals of failed tests), so they can be inspected further.
> > 
> > 
> > https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/10309/artifact//systemd-centos-ci/index.html
> 
> Thanks. Hm... the VM kernel seems to lack TPM support - no device is available from userspace. Maybe this system has it configured as a kernel module, I'll try to load that. But maybe you have some more insight about the CI system?

It seems that loading the module should be enough (at least according to the Arch wiki https://wiki.archlinux.org/title/Trusted_Platform_Module#Using_TPM_1.2).

src/shared/tpm2-util.c
```diff
@@ -717,9 +739,26 @@ static int tpm2_make_pcr_session(
...
+        assert(pin);
+        pin_len = strlen(pin);
+
+        sha256_init_ctx(&hash);
+        sha256_process_bytes(pin, pin_len, &hash);
+        sha256_finish_ctx(&hash, ret_digest);
+
+        explicit_bzero_safe(&hash, sizeof(hash));
+
+        return ret_digest;
```
bluca: There's no point in this function having a return value, it cannot fail.

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,83 @@ static int search_policy_hash(
...
+        assert(ret_pin_str);
+        assert(ret_flags);
+
+        r = getenv_steal_erase("NEWPIN", &pin_str);
+        if (r < 0)
+                return log_error_errno(r, "Failed to acquire PIN from environment: %m");
+        if (r > 0)
+                flags |= TPM2_FLAGS_USE_PIN;
+        else {
+                for (int i = 5;; i--) {
```
bluca: We usually use size_t for iterators

src/cryptsetup/cryptsetup-tokens/luks2-tpm2.c
```diff
@@ -1,11 +1,15 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
 #include "alloc-util.h"
+#include "env-util.h"
```
bluca: sort alphabetically

src/cryptsetup/cryptsetup-tokens/luks2-tpm2.h
```diff
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include "tpm2-util.h"
```
bluca, line 5: empty line after the include

test/units/testsuite-70.sh
```diff
@@ -0,0 +1,47 @@
...
+export SYSTEMD_LOG_LEVEL=debug
+
+# Prepare fresh disk image
+dd if=/dev/zero of=/test.img bs=1024k count=100 status=none
```
bluca: put this on /tmp, / might not be writable

test/units/testsuite-70.sh
```diff
@@ -0,0 +1,47 @@
...
+echo -n passphrase >/tmp/passphrase
+cryptsetup luksFormat -q --use-urandom test.img /tmp/passphrase
+
+# Enroll unlock with default PCR policy
+env PASSWORD=passphrase systemd-cryptenroll --tpm2-device=auto /test.img
+/usr/lib/systemd/systemd-cryptsetup attach test-volume /test.img - tpm2-device=auto,headless=1
+cryptsetup luksClose test-volume
```
bluca: systemd-cryptsetup detach ?

test/units/testsuite-70.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Check with wrong PCR
+tpm2_pcrextend 7:sha256=0000000000000000000000000000000000000000000000000000000000000000
+/usr/lib/systemd/systemd-cryptsetup attach test-volume /test.img - tpm2-device=auto,headless=1 && exit 1
```
bluca: When expecting a failure we use the pattern:

`foo && { echo 'unexpected success'; exit 1; }`

test/units/testsuite-70.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
+set -o pipefail
```
bluca: can't have pipefail when expecting commands to fail

bluca (approved):
Just a few more minor details, otherwise looks great to me

src/shared/tpm2-util.c
grigorig: Not even sure why I did that...

test/units/testsuite-70.sh
grigorig: And that results in an OOM error. :) But making the image smaller works.

mrc0mmand: Maybe using `/var/tmp` instead of `/tmp` (which is tmpfs) should work better/more reliably.

grigorig: Well, it works now, but why not, if it is considered good practice.

src/shared/tpm2-util.c
```diff
@@ -12,8 +12,10 @@
 #include "format-table.h"
 #include "fs-util.h"
 #include "hexdecoct.h"
+#include "hmac.h"
```
poettering: this is not actually used, is it? i.e. you opted not to do HMAC() of the passed password after all, right?

src/shared/tpm2-util.c
```diff
@@ -717,9 +739,24 @@ static int tpm2_make_pcr_session(
...
 
+static void hash_pin(const char *pin, uint8_t ret_digest[static SHA256_DIGEST_SIZE]) {
+        struct sha256_ctx hash;
+        size_t pin_len;
+
+        assert(pin);
+        pin_len = strlen(pin);
+
+        sha256_init_ctx(&hash);
+        sha256_process_bytes(pin, pin_len, &hash);
```
poettering: doesn't really matter, but I'd just pass `strlen(pin)` as second parameter directly here, there's no real benefit in using a temporary variable `pin_len`...

src/shared/tpm2-util.c
```diff
@@ -1005,11 +1048,36 @@ int tpm2_unseal(
...
+                else
+                        r = log_error_errno(
+                                        SYNTHETIC_ERRNO(ENOTRECOVERABLE),
+                                        "Failed to load HMAC key in TPM: %s",
+                                        sym_Tss2_RC_Decode(rc));
                 goto finish;
         }
 
+        if (pin) {
+                TPM2B_AUTH auth = { .size = SHA256_DIGEST_SIZE };
```
poettering: we usually would line break this like this:

```c
TPM2B_AUTH auth = {
        .size = SHA256_DIGEST_SIZE,
};
```

but doesn't matter

src/shared/tpm2-util.c
```diff
@@ -1306,3 +1374,12 @@ int tpm2_primary_alg_from_string(const char *alg) {
...
 }
+
+bool tpm2_valid_pin(const char *pin) {
+        assert(pin);
+
+        /* Enforce a PIN length of at least 4 characters: Fewer characters provide little protection, even
+         * with dictionary attack (DA) protection. In the unlock case, validation helps to avoid unnecessary
+         * unlock attempts that increment the DA counter. */
+        return strlen(pin) >= 4;
+}
```
poettering: so i am not sure we should do this kind of validation here, but rather defer to libpwquality on the client side really. i.e. we validate passwords anyway client side, and I am not sure we should validate PINs in any different way (after all it's just a different name for the exact same thing...). 

(I mean, it might be OK to use slightly weaker settings for libpwquality when setting things up with a TPM2 backing, given the DA protection, but the point I am trying to make here is that we shouldn't have two layers of pw quality checks: a simplistic checking here — which ends up being placed server-side – and a complex, detailed checking client-side, it should use the same code/logic, and be placed at the same possitions, and at most differ in selected validation levels)

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -58,11 +60,83 @@ static int search_policy_hash(
...
+                        }
+
+                        log_error("PINs didn't match, please try again!");
+                }
+        }
+
+        *ret_flags = flags;
+        *ret_pin_str = TAKE_PTR(pin_str);
+
+        return r;
```
poettering: I'd just do `return 0` here... i.e. the value of `r` doesn't really make sense to propagate, since we don't end up here as direct consequence of a nested call returning it, but only by consequence of an indirect check via strv_equal()

src/shared/tpm2-util.h
```diff
@@ -6,6 +6,12 @@
 #include "json.h"
 #include "macro.h"
 
+typedef enum TPM2Flags {
+        TPM2_FLAGS_USE_PIN = 1,
+        _TPM2_FLAGS_MAX,
```
poettering: given this is flags field, the _MAX enum doesn't make too much sense, i.e. it would be the OR combination of the lowest and the highest defined bit, but that's not really a maximum value of anything...

src/shared/tpm2-util.c
```diff
@@ -1375,6 +1378,18 @@ int tpm2_primary_alg_from_string(const char *alg) {
         return -EINVAL;
 }
 
+TPM2Flags tpm2_flag_from_string(const char *flag) {
+        if (streq_ptr(flag, "pin"))
+                return TPM2_FLAGS_USE_PIN;
+        return -EINVAL;
```
poettering: hmmm, so what string would you pass here to indicate "flag not set"?

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,63 @@
...
+        if (!r) {
+                if (headless)
+                        return log_error_errno(
+                                        SYNTHETIC_ERRNO(ENOPKG),
+                                        "PIN querying disabled via 'headless' option. "
+                                        "Use the '$PIN' environment variable.");
+
+                pin = strv_free_erase(pin);
+                r = ask_password_auto(
+                        "Please enter TPM2 PIN:",
```
poettering: we usually indent continuation lines by 16ch, not 8ch

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -64,7 +117,54 @@ int acquire_tpm2_key(
...
+                                pcr_bank,
+                                primary_alg,
+                                blob,
+                                blob_size,
+                                policy_hash,
+                                policy_hash_size,
+                                NULL,
+                                ret_decrypted_key,
+                                ret_decrypted_key_size);
+        }
```
poettering: no {} around single-line if blocks, according our coding style

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,63 @@
...
+                        &pin);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to ask for user pin: %m");
+                assert(strv_length(pin) == 1);
+
+                if (!tpm2_valid_pin(*pin)) {
+                        return log_error_errno(
+                                        SYNTHETIC_ERRNO(EPERM),
+                                        "Incorrect PIN length (must be at least 4 characters)!");
+                }
```
poettering: no {} around single-line if blocks, according to coding style

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,63 @@
...
+
```
poettering: it makes a ton of sense to validate PIN syntax when setting the PIN, but when checking it, I think we should just pass the value along as is. i.e. PIN policies might change over time and it should always be allowed to try PINs that might not be OK by today's standards but might have been by yesterday's standards.

(it might make sense to maybe as a friendly hint once the authentication failed with a PIN to check if it is superficially OK looking and then inform the user if it isn't, but that'd just be the cherry on top... I mean, we do no such thing for regular paswords...)

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -1,14 +1,63 @@
...
+                        "tpm2-pin",
+                        "cryptsetup.tpm2-pin",
+                        until,
+                        ask_password_flags,
+                        &pin);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to ask for user pin: %m");
+                assert(strv_length(pin) == 1);
+
+                if (!tpm2_valid_pin(*pin)) {
```
poettering: doesn't matter, but given this in theory is an array of strings, I'd access the first element as `pin[0]` rather than `*pin`, just as a hint to the reader that that's what we do here: not just deref a pointer, but get the first element from an array the pointer points too, even if technically both are the same thing.

src/cryptsetup/cryptsetup-tpm2.c
```diff
@@ -64,7 +117,54 @@ int acquire_tpm2_key(
...
+        }
+
+        for (int i = 5;; i--) {
+                _cleanup_(erase_and_freep) char *pin_str = NULL;
+
+                if (i <= 0)
+                        return -EACCES;
+
+                r = get_pin(until, ask_password_flags, headless, &pin_str);
+                if (r == -EPERM)
```
poettering: add comment what this means (but as mentioned, I think this should actually go away: do password length validation when setting PINs, not when checking PINS)

src/cryptsetup/cryptsetup.c
```diff
@@ -1301,9 +1301,16 @@ static int attach_luks_or_plain_or_bitlk_by_tpm2(
...
+                                        0, /* PIN is currently unhandled in this case */
+                                        until,
+                                        arg_headless,
+                                        arg_ask_password_flags,
                                         &decrypted_key, &decrypted_key_size);
                         if (r >= 0)
                                 break;
+                        if (IN_SET(r, -EACCES, -ENOLCK)) {
+                                return log_error_errno(SYNTHETIC_ERRNO(EAGAIN), "TPM2 PIN unlock failed, falling back to traditional unlocking.");
+                        }
```
poettering: no {} around single-line if blocks please

src/cryptsetup/cryptsetup.c
```diff
@@ -1369,7 +1378,14 @@ static int attach_luks_or_plain_or_bitlk_by_tpm2(
...
                                                 blob, blob_size,
                                                 policy_hash, policy_hash_size,
+                                                tpm2_flags,
+                                                until,
+                                                arg_headless,
+                                                arg_ask_password_flags,
                                                 &decrypted_key, &decrypted_key_size);
+                                if (IN_SET(r, -EACCES, -ENOLCK)) {
+                                        return log_error_errno(SYNTHETIC_ERRNO(EAGAIN), "TPM2 PIN unlock failed, falling back to traditional unlocking.");
+                                }
```
poettering: ditto

man/systemd-cryptenroll.xml
```diff
@@ -299,6 +299,24 @@
         signatures likely will validate against pre-existing certificates.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--tpm2-with-pin=</option><replaceable>BOOL</replaceable></term>
+
+        <listitem><para>When enrolling a TPM2 device, controls whether to require the user to enter a PIN
+        when unlocking the volume in addition to PCR binding, based on TPM2 policy authentication. Defaults
+        to <literal>no</literal>. The PIN needs to be 4-32 characters long. Despite being called PIN,
```
poettering: it can be longer than 32 chars now, given the hashing.

poettering:
excellent work! thanks so much. In particular the testsuite stuff is amazing.

src/shared/tpm2-util.c
grigorig: What's client side and server side? Are you talking about password agents?

Anyway, since this is a bit controversial, I'll just remove the length check completely.

src/shared/tpm2-util.c
grigorig: Hm, why would you need to do that? This is a bit of a left over from the approach with "feature keywords" anyway. The function is not used at the moment and probably not needed. I'll remove it.

src/shared/tpm2-util.c
poettering: Forget client/server side, doesn't matter.

src/shared/tpm2-util.c
grigorig: It's only used for the `SHA256_DIGEST_SIZE` define. I wonder why this isn't in sha256.h. Maybe it should be moved there?

poettering: move it over, please!

src/shared/tpm2-util.c
poettering: yeah, i didn't see it wasn't used anymore. please remove

poettering:
So yes, makes sense to drop the PIN validation for now entirely. In the long run I'd love some basic hookup to the quality_check_password() infra we have in place, which wraps libpwquality. It's not entirely trivial though, since we need to reduce the parameters in some reasonable way, after all a PIN can be much shorter and have less entropy given its backed by TPM. Ideally we could tell libpwquality some target entropy or so, and it would ten figure out the right params on its own.

src/shared/tpm2-util.h
grigorig: I agree, removed.

src/shared/tpm2-util.c
```diff
@@ -813,6 +848,10 @@ int tpm2_seal(
                 .size = sizeof(hmac_sensitive.sensitive),
                 .sensitive.data.size = 32,
         };
+        if (pin) {
+                hash_pin(pin, strlen(pin), hmac_sensitive.sensitive.userAuth.buffer);
+                hmac_sensitive.sensitive.userAuth.size = SHA256_DIGEST_SIZE;
+        }
```
poettering, line 854: i think it would be great to erase this memory when done. i.e. under the `finish` label at the end of this function, add an `explicit_bzero_safe()` call on `hmac_sensitive` (or some part of it, though I personally wouldn't bother)

src/shared/tpm2-util.c
```diff
@@ -1373,3 +1376,10 @@ int tpm2_primary_alg_from_string(const char *alg) {
                 return TPM2_ALG_RSA;
         return -EINVAL;
 }
+
+const char *tpm2_flags_to_string(TPM2Flags flags) {
+        if (flags & TPM2_FLAGS_USE_PIN)
+                return "pin";
+        return NULL;
+}
```
poettering: tpm2_flags_to_string() is not actually used?

src/shared/tpm2-util.h
```diff
@@ -6,6 +6,11 @@
 #include "json.h"
 #include "macro.h"
 
+typedef enum TPM2Flags {
+        TPM2_FLAGS_USE_PIN = 1,
```
poettering: we usually define flags as:

```c
typedef enum xxx {
        yyy = 1 << 0,
        zzz = 1 << 1,
        aaa = 1 << 2,
        …
} xxx;
```

given this is a flags enum, please hence replace the right-hand side of this line with `1 << 0`, even if it's exctly the same. But it will visibly mark this as bitmask, and that's a good thing

poettering (changes requested):

poettering:
almost ready

src/shared/tpm2-util.h
poettering: i documented this now in #22732

poettering:
this will unblock #22630 once merged.

src/shared/tpm2-util.c
grigorig: Right. I thought it's still used in cryptsetup_token_dump(), but that's not true. I'll also remove that function.

src/shared/tpm2-util.c
grigorig: Good idea.

poettering:
excellent work! thanks so much!

PR #22565, opened 2022-02-18
generator: Rename password arg

medhefgo:
This function does not expect a password, but a key file path. The
cryptsetup helper binary even calls it that.

No code changes.

### DIFF
--- a/src/cryptsetup/cryptsetup-generator.c
+++ b/src/cryptsetup/cryptsetup-generator.c
@@ -277,7 +277,7 @@ static int print_dependencies(FILE *f, const char* device_path) {
 static int create_disk(
                 const char *name,
                 const char *device,
-                const char *password,
+                const char *key_file,
                 const char *keydev,
                 const char *headerdev,
                 const char *options,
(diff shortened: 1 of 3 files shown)


yuwata (approved):
LGTM.

medhefgo:
Also, somebody should really take a look at the [other alerts](https://github.com/systemd/systemd/security/code-scanning) (I don't really know/feel comfortable with how to deal with these)…

src/cryptsetup/cryptsetup-generator.c
```diff
@@ -277,15 +277,15 @@ static int print_dependencies(FILE *f, const char* device_path) {
 static int create_disk(
                 const char *name,
                 const char *device,
-                const char *password,
+                const char *key_file,
```
poettering, line 280: this is inherited from the crypttab files from really old distros where the column was called "password", because back then you could specify a literal password there. THis was removed eventually, already before we implemented systemd-cryptsetup but we stuck to the column name.

I changed this in the man page in 6e41f4dd916293f35d7d35cea7eed1807d7ea771. So yes it makes a ton of sense to change this in the code, too.

Maybe mention in the commit msg that this is a follow-up for 6e41f4dd916293f35d7d35cea7eed1807d7ea771 in a way

poettering:
Hmm, you get bonus points if you also fix the two `[PASSWORD]` occurences in cryptsetup.c to `[KEY-FILE]`, they refere to the same column after all.

medhefgo:
> Hmm, you get bonus points if you also fix the two `[PASSWORD]` occurences in cryptsetup.c to `[KEY-FILE]`, they refere to the same column after all.

What can I buy with those? :cat:

PR #22569, opened 2022-02-18
mkosi CI: mask isc-dhcp-server units

bluca:
The packages are installed to provide the dhcpd binary, used by
test/test-network/systemd-networkd-tests.py, but we don't need the units
to run, and in fact in some cases the image fails to boot because of
them:

Spawning container image on /home/runner/work/systemd/systemd/image.raw.
Press ^] three times within 1s to kill container.
● isc-dhcp-server.service  loaded failed failed ISC DHCP IPv4 server
● isc-dhcp-server6.service loaded failed failed ISC DHCP IPv6 server
Container image failed with error code 1.
Error: Process completed with exit code 1.

Mask the units with an --extra-tree.

Example of a failed job: https://github.com/systemd/systemd/runs/5254240178?check_suite_focus=true

### DIFF
--- a/.github/workflows/run_mkosi.sh
+++ b/.github/workflows/run_mkosi.sh
@@ -7,11 +7,20 @@ set -o pipefail
 
 EC=0
 TEMPFILE="$(mktemp)"
-trap "rm -f '$TEMPFILE'" EXIT
+TEMP_EXTRA_TREE="$(mktemp --directory)"
+trap "rm -rf '$TEMPFILE' '$TEMP_EXTRA_TREE'" EXIT
+
+# We need isc-dhcp-server to be installed for the networkd unit tests, but we don't want to
+# run it by default. mktemp creates the directory as 700, so change it, otherwise it will
+# affect the image's root folder permissions.
+chmod 755 "$TEMP_EXTRA_TREE"
+mkdir -p "$TEMP_EXTRA_TREE/etc/systemd/system/"
+ln -s /dev/null "$TEMP_EXTRA_TREE/etc/systemd/system/isc-dhcp-server.service"
+ln -s /dev/null "$TEMP_EXTRA_TREE/etc/systemd/system/isc-dhcp-server6.service"
 
 for ((i = 0; i < 5; i++)); do
     EC=0
-    (sudo python3 -m mkosi "$@") |& tee "$TEMPFILE" || EC=$?
+    (sudo python3 -m mkosi --extra-tree="$TEMP_EXTRA_TREE" "$@") |& tee "$TEMPFILE" || EC=$?
     if [[ $EC -eq 0 ]]; then
         # The command passed — let's return immediately
         break


.github/workflows/run_mkosi.sh
```diff
@@ -9,9 +9,17 @@ EC=0
 TEMPFILE="$(mktemp)"
 trap "rm -f '$TEMPFILE'" EXIT
 
+# We need isc-dhcp-server to be installed for the networkd unit tests, but we don't want to
+# run it by default
+TEMP_EXTRA_TREE="$(mktemp --directory)"
+trap "rm -rf '$TEMP_EXTRA_TREE'" EXIT
```
mrc0mmand: Unfortunately, this won't work (I mean, it doesn't matter in this case, since the machine gets scraped anyway), as you can have only one handler per signal in bash:

```
$ bash
$ trap 'echo 1' EXIT
$ trap 'echo 2' EXIT
$ exit
exit
2
```

So I guess moving them to a function should work

```bash
at_exit() {
    set +e
    [[ -n "${TEMPFILE:-}" ]] && rm -fr "$TEMPFILE"
    [[ -n "${TEMP_EXTRA_TREE:-}" ]] && rm -fr "$TEMP_EXTRA_TREE"
}

trap at_exit EXIT
```

or something along these lines.

bluca: good catch, fixed by simply removing both dirs

bluca:
@DaanDeMeyer any idea what could be the issue?

mrc0mmand:
Oh god. After some painful debugging, the issue is that the `--extra-tree=` option apparently changes permissions of `/`:

```
root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# runuser testuser
runuser: failed to execute /bin/sh: Permission denied

root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# ls -la /
total 57
drwx------   1 root root  160 Feb 19 11:31 .
drwx------   1 root root  160 Feb 19 11:31 ..
lrwxrwxrwx   1 root root    7 Feb 19 11:29 bin -> usr/bin
...
```

After running `chmod 0655 /` everything magically works once again:
```
root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# runuser testuser
$ ^C
$ 

root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# ./test-rm-rf
/* test_rm_rf_chmod */
Successfully forked off '(setresuid)' as PID 3766.
(setresuid) succeeded.
```

yuwata:
Another idea: let's migrate the test which requires dhcpd to use dnsmasq, and drop isc dhcpd from mkosi images.

But actually, the failure is so spurious...

bluca:
> Oh god. After some painful debugging the issue is that the `--extra-tree=` option apparently changes permissions of `/`:
> 
> ```
> root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# runuser testuser
> runuser: failed to execute /bin/sh: Permission denied
> 
> root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# ls -la /
> total 57
> drwx------   1 root root  160 Feb 19 11:31 .
> drwx------   1 root root  160 Feb 19 11:31 ..
> lrwxrwxrwx   1 root root    7 Feb 19 11:29 bin -> usr/bin
> ...
> ```
> 
> After running `chmod 0655 /` everything magically works once again:
> 
> ```
> root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# runuser testuser
> $ ^C
> $ 
> 
> root@mkosi-16853ef8c5c94b30a72bcb8c135e3a11:~/build# ./test-rm-rf
> /* test_rm_rf_chmod */
> Successfully forked off '(setresuid)' as PID 3766.
> (setresuid) succeeded.
> ```

Oh _wat_ -- that doesn't sound right, @DaanDeMeyer is that a known issue?

Anyway, thanks for figuring this out, I 100% failed to notice it - I've switched to --skeleton-tree which doesn't seem to have the same issue

bluca:
> Another idea: let's migrate the test which requires dhcpd to use dnsmasq, and drop isc dhcpd from mkosi images.
> 
> But actually, the failure is so spurious...

If that happens we can remove this, but for now this will avoid these flaky failures

mrc0mmand:
Looks like CI still doesn't like it. Maybe you could revert back to `--extra-tree=` and then in `mkosi.build` do `chmod 0555 /` as a temporary workaround?

DaanDeMeyer:
> Oh wat -- that doesn't sound right, @DaanDeMeyer is that a known issue?

I think it is known. Extra trees are just copied over wholesale with permissions and everything included. We could probably do better here.

bluca:
> Looks like CI still doesn't like it. Maybe you could revert back to `--extra-tree=` and then in `mkosi.build` do `chmod 0555 /` as a temporary workaround?

_hhhh_ ok, I've added a chmod after the mktemp (which creates the directory as 700)

mrc0mmand (approved):

mrc0mmand:
> > Looks like CI still doesn't like it. Maybe you could revert back to `--extra-tree=` and then in `mkosi.build` do `chmod 0555 /` as a temporary workaround?
> 
> _hhhh_ ok, I've added a chmod after the mktemp (which creates the directory as 700)

*This satisfies the CI gods.*

bluca:
> > > Looks like CI still doesn't like it. Maybe you could revert back to `--extra-tree=` and then in `mkosi.build` do `chmod 0555 /` as a temporary workaround?
> > 
> > 
> > _hhhh_ ok, I've added a chmod after the mktemp (which creates the directory as 700)
> 
> _This satisfies the CI gods._

_Praise be!_ this affects only the mkosi jobs, so I'll merge right away, before the machine-gods change their minds

DaanDeMeyer:
@bluca @mrc0mmand I think https://github.com/systemd/mkosi/pull/911 should improve the behavior when using mkosi.extra/. Could you take a look and see if that makes sense and fixes the issue?

Issue #22567, opened 2022-02-18
systemctl timestamp formats inadequate for machine parsing, no epoch


twojstaryzdomu:
**systemd version the issue has been seen with**
systemd 250

**Expected behaviour you didn't see**
None of the timestamp formats available in `systemctl` are machine parseable. For instance, GNU date has issues with all of those.
The `--timestamp` parameter inconveniently accepts only arbitrary human readable values.

**Steps to reproduce the problem**
`# LANG=C systemctl show systemd-journald | grep Timestamp`
no epoch values here

```
# systemctl --help
...
                         'pretty' (default): 'Day YYYY-MM-DD HH:MM:SS TZ
                         'us': 'Day YYYY-MM-DD HH:MM:SS.UUUUUU TZ
                         'utc': 'Day YYYY-MM-DD HH:MM:SS UTC
                         'us+utc': 'Day YYYY-MM-DD HH:MM:SS.UUUUUU UT
```

The epoch provides a convenient time specification for machine parsing. Sadly, systemctl isn't doing much in the way of enabling its output to be parsable by scripts easily.

Seeing as there is a choice of format, one would expect epoch there. For some reason, it's either 'us' or 'utc' or combination of both. Shouldn't there be a greater choice of formats here?

I don't see much difference between these formats anyway. None of them is accepted by `date` for machine parsing:

```
# date -d$(systemctl show --timestamp=pretty systemd-journald -P ActiveEnterTimestamp)
date: extra operand ‘19:20:05’
Try 'date --help' for more information.
# date -d$(systemctl show --timestamp=utc systemd-journald -P ActiveEnterTimestamp)
date: extra operand ‘19:20:05’
Try 'date --help' for more information.
# date -d$(systemctl show --timestamp=us+utc systemd-journald -P ActiveEnterTimestamp)
date: extra operand ‘19:20:05’
Try 'date --help' for more information.
# date -d$(systemctl show --timestamp=us systemd-journald -P ActiveEnterTimestamp)
date: extra operand ‘19:20:05’
Try 'date --help' for more information.
```

Please allow at least epoch to be displayed, or at best a free format string, with sequences similar to FORMAT used by `GNU Coreutils` (`%s` = epoch, etc).

mrc0mmand:
> I don't see much difference between these formats anyway. None of them is accepted by `date` for machine parsing:
> 
> ```
> # date -d$(systemctl show --timestamp=pretty systemd-journald -P ActiveEnterTimestamp)
[...]

PR #22573, opened 2022-02-19
time-util: introduce TIMESTAMP_UNIX

mrc0mmand:
Allow formatting timestamps as number of seconds since the Epoch for easier
machine parsing.

Fixes: #22567

```
$ systemctl show systemd-journald | grep Timestamp
WatchdogTimestampMonotonic=0
ExecMainStartTimestamp=Sat 2021-12-11 15:25:57 CET
ExecMainStartTimestampMonotonic=13030408
ExecMainExitTimestampMonotonic=0
StateChangeTimestamp=Sat 2021-12-11 15:25:57 CET
StateChangeTimestampMonotonic=13049273
InactiveExitTimestamp=Sat 2021-12-11 15:25:57 CET
InactiveExitTimestampMonotonic=13030430
ActiveEnterTimestamp=Sat 2021-12-11 15:25:57 CET
ActiveEnterTimestampMonotonic=13049273
ActiveExitTimestamp=Sat 2021-12-11 15:25:57 CET
ActiveExitTimestampMonotonic=12997236
InactiveEnterTimestamp=Sat 2021-12-11 15:25:57 CET
InactiveEnterTimestampMonotonic=13028890
ConditionTimestamp=Sat 2021-12-11 15:25:57 CET
ConditionTimestampMonotonic=13029539
AssertTimestamp=Sat 2021-12-11 15:25:57 CET
AssertTimestampMonotonic=13029540

$ systemctl show --timestamp=unix systemd-journald | grep Timestamp
WatchdogTimestampMonotonic=0
ExecMainStartTimestamp=@1639232757
ExecMainStartTimestampMonotonic=13030408
ExecMainExitTimestampMonotonic=0
StateChangeTimestamp=@1639232757
StateChangeTimestampMonotonic=13049273
InactiveExitTimestamp=@1639232757
InactiveExitTimestampMonotonic=13030430
ActiveEnterTimestamp=@1639232757
ActiveEnterTimestampMonotonic=13049273
ActiveExitTimestamp=@1639232757
ActiveExitTimestampMonotonic=12997236
InactiveEnterTimestamp=@1639232757
InactiveEnterTimestampMonotonic=13028890
ConditionTimestamp=@1639232757
ConditionTimestampMonotonic=13029539
AssertTimestamp=@1639232757
AssertTimestampMonotonic=13029540
```


src/basic/time-util.h
```diff
@@ -30,6 +30,7 @@ typedef struct triple_timestamp {
 } triple_timestamp;
 
 typedef enum TimestampStyle {
+        TIMESTAMP_EPOCH,
```
bluca: bikeshedding: wouldn't TIMESTAMP_UNIX be more appropriate? That's how it's usually called?

mrc0mmand: I guess that makes sense, updated - naming stuff is hard :(

bluca (approved):

src/basic/time-util.c
```diff
@@ -362,6 +363,13 @@ char *format_timestamp_style(
         if (!localtime_or_gmtime_r(&sec, &tm, utc))
                 return NULL;
 
+        if (style == TIMESTAMP_UNIX) {
+                if (strftime(buf, l, "@%s", &tm) <= 0)
+                        return NULL; /* Doesn't fit */
```
poettering, line 358: hmm, so here you convert microseconds since 1970 to "broken down" (i.e. calendar) time first, via `localtime_or_gmtime()` and then convert it back to seconds since 1970 via `strftime()`'s `%s` specifier. Please just use the value as is, i.e. add right after the timestamp-is-unset check above just:

```c
r = snprintf(buf, l, "@" FMT_USEC, t / USEC_PER_SEC);  /* round down µs → s */
if (r < 0 || (size_t) r >= l)        
        return NULL;
```

That way we don't have do the roundtripping through the calendar time. Which should make things a bit more robust, if between the forth and back conversion the timezone or summer time changes or so.

src/systemctl/systemctl.c
```diff
@@ -298,6 +298,7 @@ static int systemctl_help(void) {
                "     --plain             Print unit dependencies as a list instead of a tree\n"
                "     --timestamp=FORMAT  Change format of printed timestamps.\n"
                "                         'pretty' (default): 'Day YYYY-MM-DD HH:MM:SS TZ\n"
+               "                         'unix': '@seconds-since-the-epoch'\n"
                "                         'us': 'Day YYYY-MM-DD HH:MM:SS.UUUUUU TZ\n"
                "                         'utc': 'Day YYYY-MM-DD HH:MM:SS UTC\n"
                "                         'us+utc': 'Day YYYY-MM-DD HH:MM:SS.UUUUUU UTC\n"
```
poettering: it's kinda uncommon we put such detailed information in --help texts. Usually we just list the available choices and add more detailed info to the man page instead. I think it would be a good time to move it there now, given we are adding antoher entry here...

src/basic/time-util.c
mrc0mmand: Indeed, updated.

src/systemctl/systemctl.c
mrc0mmand: Good point.

src/basic/time-util.c
```diff
@@ -357,6 +359,14 @@ char *format_timestamp_style(
                 return buf;
         }
 
+        if (style == TIMESTAMP_UNIX) {
+                r = snprintf(buf, l, "@" USEC_FMT, t / USEC_PER_SEC);  /* round down µs → s */
+                if (r < 0 || (size_t) r >= l)
+                        return NULL; /* Doesn't fit */
+
+                return buf;
+        }
```
poettering, line 361: i think this should be moved 6 lines up, i.e. before the > 9999 thing. Because the the check it does doesn't apply to TIMESTAMP_UNIX at all: we can output anything we can store in usec_t pretty much.

poettering:
looks good, just one more minor thing

src/basic/time-util.c
mrc0mmand: Updated!

mrc0mmand:
Thanks for the reviews! Applied the last suggestion & upgrading the green label.

poettering:
Failed CI appears unrelated.

Issue #22571, opened 2022-02-19
Multiple delegated prefixes


jmpolom:
**Is your feature request related to a problem? Please describe.**
If an upstream DHCPv6 implementation provides small (/64's) delegated prefixes, only a single network will be able to be configured for stateless PD. Often multiple prefixes may be requested from the upstream provider but it seems networkd does not do this as of v249. See:

```
Feb 18 22:00:09 fedora systemd-networkd[92057]: internet: DHCP6: received PD Prefix 2600:1700:bf30:1d9f::/64
Feb 18 22:00:09 fedora systemd-networkd[92057]: home: DHCPv6-PD address 2600:1700:bf30:1d9f:ae1f:6bff:fe6f:e816/64 (valid for 1h, preferred for 1h)
Feb 18 22:00:09 fedora systemd-networkd[92057]: management: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:00:09 fedora systemd-networkd[92057]: guests: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:00:09 fedora systemd-networkd[92057]: things: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:00:09 fedora systemd-networkd[92057]: diag: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:27:58 fedora systemd-networkd[92057]: management: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:27:58 fedora systemd-networkd[92057]: guests: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:27:58 fedora systemd-networkd[92057]: things: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:27:58 fedora systemd-networkd[92057]: diag: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:56:35 fedora systemd-networkd[92057]: management: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:56:35 fedora systemd-networkd[92057]: guests: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:56:35 fedora systemd-networkd[92057]: things: Couldn't find a suitable prefix. Ran out of address space.
Feb 18 22:56:35 fedora systemd-networkd[92057]: diag: Couldn't find a suitable prefix. Ran out of address space.
```

**Describe the solution you'd like**
[...]

PR #22574, opened 2022-02-20
network: dhcp-pd: fix two issues

### LINKED ISSUE #22559: networkd: assigning IA_PD address to upstream, wrong subnet size
systemd 250.3

**Used distribution**
 > Debian 11, with bullseye-backports

 > `Linux nexushw 5.10.0-11-amd64 #1 SMP Debian 5.10.92-1 (2022-01-18) x86_64 GNU/Linux`

 > x86_64

 > On a systemd-networkd-powered CE router with one upstream and several internal interfaces, connected to a v4/v6 dual-stack link: Using the config example in fc2161e5b98d35c971d0b9b3446fffd1f7f0a2f4, I would expect networkd to assign a /64 address to my upstream/WAN interface, honouring the `[DHCPPrefixDelegation]` parameters like SubnetId, Token, etc. just like it already works fine for internal LAN interfaces. As a consequence, the router itself would be able to use this address for v6 connections to WAN.

 > Using below's interface configuration, and with an ISP-delegated prefix of 2001:db8:205:e900::/56, the upstream/WAN interfaces gets an eui-64 address assigned in 2001:db8:205:e9**00::/56** instead of 2001:db8:205:e9**SubnetId::/64**. A respective route with /56 is added, potentially messing with the also-existing unreachable route for the delegated prefix.

`wan.network` (truncated)
```
[Match]
Name=wan

[Network]
DHCP=yes
IPv6AcceptRA=yes
IPv6SendRA=no
# we want a delegated-prefix address also on WAN
DHCPPrefixDelegation=yes

[DHCPv6]
UseDelegatedPrefix=yes
PrefixDelegationHint=::/56
WithoutRA=solicit

[DHCPPrefixDelegation]
UplinkInterface=:self
SubnetId=0xf
Announce=no
```

Interface status:
```
$ ip -br a s dev wan
wan  UP  x.y.220.189/30 2001:db8:205:e900:225:90ff:fef2:c636/56 metric 256 2001:db8:0:97:24e2:418:61c8:2433/128 fe80::225:90ff:fef2:c636/64 
```
The first v6 is the wrongly assigned /56 WAN IP (instead of /64), and ignoring the `SubnetId=0xf`. The second v6 is the IA_NA assigned IP (not relevant for the issue, would be removed in the future with `UseAddress=no`)

For comparison, the properly assigned /64 subnet on LAN-side, with `SubnetId=0x0a` taken into account:
```
$ ip -br a s dev vlan10
vlan10  UP  192.168.1.1/24 2001:db8:205:e90a::1/64 metric 256 fe80::b8c5:82ff:fe11:fab5/64
```
using this config (`Assign` primarily for visibility):
```
[Match]
Name=vlan10

[Network]
IPv6AcceptRA=no
IPv6SendRA=yes
DHCPPrefixDelegation=yes

[DHCPPrefixDelegation]
SubnetId=0x0a
Announce=yes
Assign=yes
Token=::1
```

Resulting route table (truncated):
```
$ ip -6 route
2a02:8106:205:e90a::/64 dev vlan10 proto kernel metric 256 expires 5382sec pref medium
2a02:8106:205:e900::/56 dev wan proto kernel metric 256 expires 5382sec pref medium
unreachable 2a02:8106:205:e900::/56 dev lo proto dhcp metric 1024 pref medium
```

Is this a misconfiguration on my side, or a misbehaviour of the special case when assigning a PD address on WAN? I would expect a /64 assigned to WAN exactly the same way as LAN interfaces, no?

Thanks and regards

yuwata:
Fixes #22559.
Also drop a restriction about subnet prefixes prompted by #22571.

### DIFF
--- a/src/network/networkd-dhcp-prefix-delegation.c
+++ b/src/network/networkd-dhcp-prefix-delegation.c
@@ -487,19 +487,12 @@ static int dhcp_pd_get_preferred_subnet_prefix(
                                                       "subnet id %" PRIu64 " is out of range. Only have %" PRIu64 " subnets.",
                                                       link->network->dhcp_pd_subnet_id, UINT64_C(1) << (64 - pd_prefix_len));
 
-                if (link_get_by_dhcp_pd_subnet_prefix(link->manager, &prefix, &assigned_link) >= 0 &&
-                    assigned_link != link) {
-                        _cleanup_free_ char *assigned_buf = NULL;
-
-                        (void) in6_addr_to_string(&prefix, &assigned_buf);
-                        return log_link_warning_errno(link, SYNTHETIC_ERRNO(EAGAIN),
-                                                      "The requested prefix %s is already assigned to another link.",
-                                                      strna(assigned_buf));
-                }
-
                 *ret = prefix;
                 return 0;
         }
+
+        if (dhcp_pd_get_assigned_subnet_prefix(link, pd_prefix, pd_prefix_len, ret) >= 0)
+                return 0;
 
         for (uint64_t n = 0; ; n++) {
                 /* If we do not have an allocation preference just iterate
@@ -517,11 +510,16 @@ static int dhcp_pd_get_preferred_subnet_prefix(
 
                 /* Check that the prefix is not assigned to another link. */
                 if (link_get_by_dhcp_pd_subnet_prefix(link->manager, &prefix, &assigned_link) < 0 ||
-                    assigned_link == link) {
-                        *ret = prefix;
-                        return 0;
-                }
+                    assigned_link == link)
+                        break;
         }
+
+        r = link_add_dhcp_pd_subnet_prefix(link, &prefix);
+        if (r < 0)
+                return log_link_warning_errno(link, r, "Failed to save acquired free subnet prefix: %m");
+
+        *ret = prefix;
+        return 0;
 }
 
 static int dhcp_pd_assign_subnet_prefix(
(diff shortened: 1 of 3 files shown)


jmpolom:
The changes in c539b5e452a4b15c1c4a646b0385f6ce43e4950b do not seem to fully address the issue enumerated in #22571

yuwata:
> The changes in [c539b5e](https://github.com/systemd/systemd/commit/c539b5e452a4b15c1c4a646b0385f6ce43e4950b) do not seem to fully address the issue enumerated in #22571

Yeah, sorry, I misunderstood the issue #22571. But still this change should be OK. And I guess there is no way to fix #22571 if the upstream DHCP6 server ignores `PrefixDelegationHint=`.

I updated the commit message of the second commit.

src/network/networkd-address-generation.c
```diff
@@ -252,8 +252,8 @@ static int generate_addresses(
         return 0;
 }
 
-int dhcp_pd_generate_addresses(Link *link, const struct in6_addr *prefix, uint8_t prefixlen, Set **ret) {
-        return generate_addresses(link, link->network->dhcp_pd_tokens, &DHCP_PD_APP_ID, prefix, prefixlen, ret);
+int dhcp_pd_generate_addresses(Link *link, const struct in6_addr *prefix, Set **ret) {
+        return generate_addresses(link, link->network->dhcp_pd_tokens, &DHCP_PD_APP_ID, prefix, 64, ret);
```
bluca, line 256: `64` is used twice, maybe add a `#define` for it?

bluca (approved):

src/network/networkd-address-generation.c
yuwata: Is it? I do not think so... At most, I'd like to leave this to another PR.

keszybz: I don't think we want to do this, because "/64" is used in many places in strings, and we can't use a define there. (We could do something more complicated with templating, but that seems vastly overkill.)  I'd say that "64" is a well-known-constant and we don't need a define for it.

keszybz:
`.github/workflows/build_test.sh: line 113: clang-13: command not found` — that was already fixed today by a different PR.
Let's merge.

PR #22582, opened 2022-02-21
boot: Minor input fixes

medhefgo:


src/boot/efi/boot.c
```diff
@@ -182,6 +182,9 @@ static BOOLEAN line_edit(
                         cursor_color = TEXT_ATTR_SWAP(cursor_color);
 
                         err = console_key_read(&key, 750 * 1000);
+                        if (!IN_SET(err, EFI_SUCCESS, EFI_TIMEOUT, EFI_NOT_READY))
+                                return FALSE;
```
poettering, line 186: hmm, can you elaborate on the commit msg? how could we possibly end up with an uninitialized `key`? I mean, we'd leave this loop only on EFI_SUCCESS, but in that case the `key` should definitely be initialized... what am i missing?

medhefgo: It's mostly theoretical and it doesn't really happen in this invocation. But the one in menu_run and print_status could. I'm just trying to be thorough on all of them.

src/boot/efi/console.c
```diff
@@ -133,20 +132,19 @@ EFI_STATUS console_key_read(UINT64 *key, UINT64 timeout_usec) {
 
                 /* do not distinguish between left and right keys */
                 if (keydata.KeyState.KeyShiftState & EFI_SHIFT_STATE_VALID) {
+                        if (keydata.KeyState.KeyShiftState & (EFI_RIGHT_SHIFT_PRESSED|EFI_LEFT_SHIFT_PRESSED))
```
poettering: hmm, the exprssion on the right is identical to EFI_SHIFT_PRESSED. Hence maybe just write that out like that:

```c
if (keydata.KeyState.KeyShiftState & EFI_SHIFT_PRESSED) 
        shift |= EFI_SHIFT_PRESSED;
```
and then maybe explain that this sets all of the relevant bits if any are set, because this of course will raise eyebrows...

src/boot/efi/boot.c
medhefgo: (Also, this check here is to prevent infinite looping for unexpected/fatal input errors)

src/boot/efi/boot.c
```diff
@@ -436,8 +439,16 @@ static void ps_bool(const CHAR16 *fmt, BOOLEAN value) {
...
 }
 
-static void print_status(Config *config, CHAR16 *loaded_image_path) {
+static BOOLEAN ps_pause(void) {
         UINT64 key;
+        EFI_STATUS err;
+
+        Print(L"\n--- Press any key to continue, ESC or q to quit. ---\n\n");
+        err = console_key_read(&key, UINT64_MAX);
+        return (EFI_ERROR(err) || IN_SET(key, KEYPRESS(0, SCAN_ESC, 0), KEYPRESS(0, 0, 'q'), KEYPRESS(0, 0, 'Q')));
```
poettering: the outer () is not necessary.

src/boot/efi/boot.c
```diff
@@ -436,8 +439,16 @@ static void ps_bool(const CHAR16 *fmt, BOOLEAN value) {
...
+}
```
poettering, line 449: it looks a bit weird to return true on "quit" and false on "continue". Should be the other way round philopsophically?

src/boot/efi/boot.c
```diff
@@ -734,7 +743,9 @@ static BOOLEAN menu_run(
                         beep(idx_highlight + 1);
 
                 err = console_key_read(&key, timeout_remain > 0 ? 1000 * 1000 : UINT64_MAX);
-                if (err == EFI_TIMEOUT) {
+                if (err == EFI_NOT_READY)
+                        continue;
```
poettering, line 749: what does EFI_NOT_READY precisely mean in this context? (maybe add a comment?)

src/boot/efi/boot.c
```diff
@@ -734,7 +743,9 @@ static BOOLEAN menu_run(
...
+                else if (err == EFI_TIMEOUT) {
```
poettering: you can drop the `else` btw.

src/boot/efi/boot.c
```diff
@@ -743,8 +754,12 @@ static BOOLEAN menu_run(
 
                         /* update status */
                         continue;
-                } else
-                        timeout_remain = 0;
+                } else if (EFI_ERROR(err)) {
```
poettering: ditto

src/boot/efi/boot.c
```diff
@@ -734,7 +743,9 @@ static BOOLEAN menu_run(
                         beep(idx_highlight + 1);
 
                 err = console_key_read(&key, timeout_remain > 0 ? 1000 * 1000 : UINT64_MAX);
-                if (err == EFI_TIMEOUT) {
+                if (err == EFI_NOT_READY)
+                        continue;
+                else if (err == EFI_TIMEOUT) {
                         timeout_remain--;
```
poettering, line 752: maybe add an `asssert(timeout_remain > 0);` to clarify that we don't expect to see `EFI_TIMEOUT` if `timeout_remain == 0` holds

bluca (approved):

PR #22583, opened 2022-02-21
hwdb: add CST Laser Trackball

prumian:
Add DPI and frequency for CST Laser Trackball. Tested and proved working on a local system.

### DIFF
--- a/hwdb.d/70-mouse.hwdb
+++ b/hwdb.d/70-mouse.hwdb
@@ -172,7 +172,7 @@ mouse:usb:v04f2p0963:name:Chicony 2.4G Multimedia Wireless Kit:*
 
 # CST Laser Trackball
 mouse:usb:v5332p1400:name:Clearly Superior Technologies. CST Laser Trackball:*
- MOUSE_DPI=400@1000 800@1000 *1600@1000
+ MOUSE_DPI=400@1000 *800@1000 1600@1000
 
 ##########################################
 # Dell


hwdb.d/70-mouse.hwdb
```diff
@@ -166,6 +166,14 @@ mouse:bluetooth:v05acp030d:name:*:*
 mouse:usb:v04f2p0963:name:Chicony 2.4G Multimedia Wireless Kit:*
  MOUSE_DPI=1000@142
 
+##########################################
+# CST
+##########################################
+
+# CST Laser Trackball
+mouse:usb:v5332p1400:name:Clearly Superior Technologies. CST Laser Trackball:*
+ MOUSE_DPI=400@1000 800@1000 *1600@1000
```
poettering: if this is a trackball, it should probably also set `ID_INPUT_TRACKBALL=1`?

prumian: @poettering It's being caught by general rule for trackballs in lines 143-149, so probably not needed explicitly?

whot: yep, not needed explicitly

@prumian: does this thing really ship with 1600 out of the box? the bits I found online say it's button-selectable, is this a hw button or a software-configured button?

prumian: @whot: it's selectable by holding right+middle button, it goes 400-800-1600 with every click. Super hard to tell which resolution it came from factory though. I'll try to find the manual...

prumian: Strangely, the manual says 800DPI is the default. I have updated the PR.

whot (approved):

hwdb.d/70-mouse.hwdb
whot: fwiw, 800 by default makes sense, that and 1000dpi are the two most common ones. Anything else is too fast by default, leading to angry customers screaming at the hotlines (or more likely these days: reddit :). Thanks for the update, this looks good now.

PR #22584, opened 2022-02-21
systemctl: rename "verb" entrypoint functions systematically to verb_xyz()

poettering:
Just some refactoring to make systemctl easier to understand: let's make sure the verb entrypoint functions are systematically named.


src/systemctl/systemctl-daemon-reload.c
```diff
@@ -29,13 +29,8 @@ int verb_daemon_reload(int argc, char *argv[], void *userdata) {
                 method = "Reexecute";
                 break;
 
-        case ACTION_SYSTEMCTL:
```
bluca, line 32: how is this case handled now?

poettering: so previously code flow was weird: we'd always go through verb_daemon_reload() and then would distinguish there again if we actually are called by systemctl and the verb logic, or directly called from the sysv tools without the verb logic. That's kinda ugly: first funneling things to a single abstract interface that then splits up afterwards is just weird.

So with this change we'll do verb_daemon_reload() only when invoked as systemctl, via the verb logic, so that everything is clean then. The inner part of the function are then split out into a new function daemon_reload() however, with a more useful parameter set, and that's the function that the sysv binaries then call. This way, code arrangement should be much simpler: instead of first badly pretending the sysv call was like the systemctl and then dealing with the actual differences inside the handler we no longer pretend anything but focus on sharing the actual code that matters instead of the glue around it.

If that makes any sense?

bluca: makes sense, thanks

bluca: maybe expand the commit message a bit given it needs a rebase anyway?

bluca (approved):
Needs a rebase

poettering:
rebased

PR #22585, opened 2022-02-21
analyze: split out each verb into its own .c/.h pair

poettering:
This is just some refactoring, modelled after what we did for systemctl: split one humongous .c file into multiple smaller ones, relatively systematically with one for each verb.

### DIFF
--- a/src/analyze/analyze-time-data.c
+++ b/src/analyze/analyze-time-data.c
@@ -7,7 +7,6 @@
 #include "bus-map-properties.h"
 #include "bus-unit-util.h"
 #include "special.h"
-#include "strxcpyx.h"
 
 static void subtract_timestamp(usec_t *a, usec_t b) {
         assert(a);
@@ -127,14 +126,11 @@ static int bus_get_uint64_property(sd_bus *bus, const char *path, const char *in
 }
 
 int pretty_boot_time(sd_bus *bus, char **ret) {
-        BootTimes *t;
-        static char buf[4096];
-        size_t size;
-        char *ptr;
-        int r;
-        usec_t activated_time = USEC_INFINITY;
-        _cleanup_free_ char *path = NULL, *unit_id = NULL;
         _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_free_ char *path = NULL, *unit_id = NULL, *text = NULL;
+        usec_t activated_time = USEC_INFINITY;
+        BootTimes *t;
+        int r;
 
         r = acquire_boot_times(bus, &t);
         if (r < 0)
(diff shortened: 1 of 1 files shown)


medhefgo:
I always wondered: Why bother with the `$toolname-` prefix in these splits? It seems unnecessary to me, considering they reside in the `$toolname` folder.

bluca:
Quite a large diff - are there any code changes (besides the last two commits) or is it a plain cut/paste?

poettering:
> I always wondered: Why bother with the `$toolname-` prefix in these splits? It seems unnecessary to me, considering they reside in the `$toolname` folder.

Mostly because most editors only show filenames not dirs when you have multiple buffers open. If you have verbs like "dump" or "time" or "verify" then the chance is really high you'll have other files dump.c/time.c/verify.c all over the place.

Also, why not name it like that? it's not that you actually have to type in the names in full, given completion everywhere?

poettering:
> Quite a large diff - are there any code changes (besides the last two commits) or is it a plain cut/paste?

Yes, I should have mentioned that. The commits whose name start with "split out" are all just that: split outs, i.e. no actual code changes, just copy/paste into new files.

keszybz:
> Mostly because most editors only show filenames not dirs when you have multiple buffers open

FWIW, I configure my emacs to append unique directory suffix, so I get e.g. `file.c/dir1` and `file.c/dir2`. This works better than trying to keep filenames unique, because it also supports the case where you have the same project in two versions, e.g. with two git workdirs. (This is not the default emacs config… By having a suffix that looks like a part of the path it is also easy to switch between the files by name and it doesn't involve any characters that would be annoying to input from the keyboard.)

I guess we can keep the file names unique. Even if isn't really necessary for the editor, it might help us avoid somebody editing the wrong file by mistake.

keszybz:
LGTM.

keszybz:
Thinking about this a bit more: those long names actually *interfere* with bash completion, at least for me. E.g. in `src/resolve/` most files the prefixes `resolved-`, but there's also `resolvectl` and `resolvconf-`, so `re<TAB><letter><TAB><letter><TAB><TAB><letter>` is needed to complete a file name. If we dropped `resolved-`, there would be many more unique prefixes.

But that's a discussion for another time.

evverx:
Looks like it's failing to compile:
```
../src/analyze/analyze-syscall-filter.c:188:5: error: no previous prototype for function 'dump_syscall_filters' [-Werror,-Wmissing-prototypes]
int dump_syscall_filters(int argc, char *argv[], void *userdata) {
    ^
../src/analyze/analyze-syscall-filter.c:188:1: note: declare 'static' if the function is not intended to be used outside of this translation unit
int dump_syscall_filters(int argc, char *argv[], void *userdata) {
^
static
1 error generated.
```

poettering:
ah, seems i fucked up the non-seccomp builds. lemme fix.

poettering:
> ah, seems i fucked up the non-seccomp builds. lemme fix.

See #22594

Issue #22527, opened 2022-02-15
RFE/RFD: lightweight integration of NFT sets for use in firewalls


topimiettinen:
**Is your feature request related to a problem? Please describe.**
Integrating firewall rules with systemd services is difficult or impossible. There are BPF options but they aren't very nice to administrate. Also, kernel's netfilter system for cgroup matching is based on numeric cgroup IDs, which change every time a service is restarted, making them useless in systemd environment. Typically DNS information isn't available for firewalls.

**Describe the solution you'd like**
PID1, networkd and resolved should optionally put discovered network and cgroup ID information to NFT sets, for use with firewall rules.
* PID1 should put cgroupv2 IDs to NFT sets named using cgroup path like `systemd.user.slice/user-1000.slice/user@1000.service/app.slice/`, or perhaps to set specified with `Service.CGroupNFTSet=systemd-app`
* networkd should put interface IP address ranges and possibly other information (at the extreme, same as [NetworkManager-dispatcher](https://developer-old.gnome.org/NetworkManager/unstable/NetworkManager-dispatcher.html)) to set named for example `systemd-eth0-ipv4` or perhaps instead specified with `Network.IPv4NFTSet=systemd-eth0-ipv4`
* resolved should put DNS information from resolved to set named according to requesting service (see also #17126) or explicitly configured

Then the firewall rules can make arbitrarily complex use of the sets. Systemd would not need to gain further firewalling abilities besides the rather simple management of specific NFT sets.

Example NFT rules:
```
table inet filter {
        set systemd-eth0-ipv4 {
                type ipv4_addr;
                flags interval;
        }
        set systemd-app {
                type integer;
        }
        set systemd-timesyncd {
                type ipv4_addr;
        }

        chain service_x_input {
                ip saddr @systemd-eth0-ipv4 counter accept
                counter drop
        }
        chain service_y_input {
                meta cgroup @systemd-app counter accept
                counter drop
        }
        chain systemd_timesyncd_output {
                ip daddr @systemd-timesyncd counter accept
                counter reject with icmpx admin-prohibited
        }
}
```

Related discussion in #7327 and #21899.

**Describe alternatives you've considered**
Daemons to manage cgroup IDs:
https://github.com/helsinki-systems/nft_cgroupv2/
https://github.com/mk-fg/systemd-cgroup-nftables-policy-manager

IPSet module for Unbound DNS puts resolved A and AAAA addresses to IPSet (but it can't identify the calling service):
[...]

PR #22587, opened 2022-02-21
core: NFT sets

topimiettinen:
New directives `ControlGroupNFTSet=`, `DynamicUserNFTSet=`, `IPv4NFTSet=` and
`IPv6NFTSet=` provide methods for integrating configuration of control groups,
dynamic users and dynamic networks into firewall rules with NFT sets.

Examples:
ControlGroupNFTSet:
```
table inet filter {
        set timesyncd {
                type cgroupsv2
        }

        chain ntp_output {
                socket cgroupv2 != @timesyncd counter drop
                accept
        }
}
```

/etc/systemd/system/systemd-timesyncd.service.d/override.conf
```
[Service]
ControlGroupNFTSet=inet:filter:timesyncd
```

```
$ sudo nft list set inet filter timesyncd
table inet filter {
        set timesyncd {
                type cgroupsv2
                elements = { "system.slice/systemd-timesyncd.service" }
        }
}
```

DynamicUser:
```
table inet filter {
        set u {
                typeof meta skuid
        }

        chain service_output {
                meta skuid != @u drop
                accept
        }
}
```
/etc/systemd/system/dunft.service

```
[Service]
ExecStart=/bin/sleep 1000
DynamicUser=yes
DynamicUserNFTSet=inet:filter:u

[Install]
WantedBy=multi-user.target
```

```
$ sudo nft list set inet filter u
table inet filter {
        set u {
                typeof meta skuid
                elements = { 64864 }
        }
}
$ ps -n --format user,group,pid,command -p `pgrep sleep`
    USER    GROUP     PID COMMAND
   64864    64864   55158 /bin/sleep 1000
```
Network:
/etc/systemd/network/lan.network
```
[Network]
...
IPv4NFTSet=netdev:filter:lan-ipv4-address
IPv6NFTSet=netdev:filter:lan-ipv6-address
```

```
table netdev filter {
        set lan-ipv4-address {
                type ipv4_addr
                flags interval
        }
        chain lan_ingress {
                type filter hook ingress device "lan" priority filter; policy drop;
                ip saddr != @lan-ipv4-address drop
                accept
        }
}
```
```
$ sudo nft list set netdev filter lan-ipv4-address
table netdev filter {
        set lan-ipv4-address {
                type ipv4_addr
                flags interval
                elements = { 10.0.0.0/24 }
        }
}
```

Closes: #22527.

### DIFF
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -3164,6 +3164,40 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
   </refsect1>
 
   <refsect1>
+    <title>Firewall Integration</title>
+    <variablelist class='unit-directives'>
+
+      <varlistentry>
+        <term><varname>DynamicUserNFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
+        <listitem><para>This setting provides a method for integrating <varname>DynamicUser=</varname>
+        configuration into firewall rules with NFT sets. This option expects a whitespace separated list of
+        NFT set definitions. Each definition consists of a colon-separated tuple of NFT address family (one
+        of <literal>arp</literal>, <literal>bridge</literal>, <literal>inet</literal>, <literal>ip</literal>,
+        <literal>ip6</literal>, or <literal>netdev</literal>), table name and set name. The names of tables
+        and sets must conform to lexical restrictions of NFT table names. When the unit starts, the user ID
+        will be appended to the NFT sets and it will be removed when the unit is stopped. Failures to manage
+        the sets will be ignored.</para>
+
+          <para>Example:
+          <programlisting>[Service]
+DynamicUserNFTSet=inet:filter:u</programlisting>
+          Corresponding NFT rules:
+          <programlisting>table inet filter {
+        set u {
+                typeof meta skuid
+        }
+        chain service_output {
+                meta skuid != @u drop
+                accept
+        }
+}</programlisting>
+          </para>
+        </listitem>
+      </varlistentry>
+    </variablelist>
+  </refsect1>
+
+  <refsect1>
     <title>System V Compatibility</title>
     <variablelist class='unit-directives'>
 
--- a/src/core/dbus-execute.c
+++ b/src/core/dbus-execute.c
@@ -22,6 +22,7 @@
 #include "execute.h"
 #include "fd-util.h"
 #include "fileio.h"
+#include "firewall-util.h"
 #include "hexdecoct.h"
 #include "io-util.h"
 #include "ioprio-util.h"
@@ -1138,6 +1139,37 @@ static int bus_property_get_exec_dir_symlink(
         return sd_bus_message_close_container(reply);
 }
 
+static int property_get_dynamic_user_nft_set(
+                sd_bus *bus,
+                const char *path,
+                const char *interface,
+                const char *property,
+                sd_bus_message *reply,
+                void *userdata,
+                sd_bus_error *error) {
+
+        ExecContext *c = userdata;
+        int r;
+
+        assert(bus);
+        assert(reply);
+        assert(c);
+
+        r = sd_bus_message_open_container(reply, 'a', "(iss)");
+        if (r < 0)
+                return r;
+
+        for (size_t i = 0; i < c->n_dynamic_user_nft_set_contexts; i++) {
+                NFTSetContext *s = &c->dynamic_user_nft_set_context[i];
+
+                r = sd_bus_message_append(reply, "(iss)", s->nfproto, s->table, s->set);
+                if (r < 0)
+                        return r;
+        }
+
+        return sd_bus_message_close_container(reply);
+}
+
 const sd_bus_vtable bus_exec_vtable[] = {
         SD_BUS_VTABLE_START(0),
         SD_BUS_PROPERTY("Environment", "as", NULL, offsetof(ExecContext, environment), SD_BUS_VTABLE_PROPERTY_CONST),
@@ -1232,6 +1264,7 @@ const sd_bus_vtable bus_exec_vtable[] = {
         SD_BUS_PROPERTY("User", "s", NULL, offsetof(ExecContext, user), SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("Group", "s", NULL, offsetof(ExecContext, group), SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("DynamicUser", "b", bus_property_get_bool, offsetof(ExecContext, dynamic_user), SD_BUS_VTABLE_PROPERTY_CONST),
+        SD_BUS_PROPERTY("DynamicUserNFTSet", "a(iss)", property_get_dynamic_user_nft_set, 0, SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("RemoveIPC", "b", bus_property_get_bool, offsetof(ExecContext, remove_ipc), SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("SetCredential", "a(say)", property_get_set_credential, 0, SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("SetCredentialEncrypted", "a(say)", property_get_set_credential, 0, SD_BUS_VTABLE_PROPERTY_CONST),
@@ -3498,6 +3531,58 @@ int bus_exec_context_set_transient_property(
                         c->temporary_filesystems = NULL;
                         c->n_temporary_filesystems = 0;
 
+                        unit_write_settingf(u, flags, name, "%s=", name);
+                }
+
+                return 1;
+
+        } else if (streq(name, "DynamicUserNFTSet")) {
+                int nfproto;
+                const char *table, *set;
+                bool empty = true;
+
+                r = sd_bus_message_enter_container(message, 'a', "(iss)");
+                if (r < 0)
+                        return r;
+
+                while ((r = sd_bus_message_read(message, "(iss)", &nfproto, &table, &set)) > 0) {
+                        const char *nfproto_name;
+
+                        nfproto_name = nfproto_to_string(nfproto);
+                        if (!nfproto_name)
+                                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Invalid protocol %d.", nfproto);
+
+                        if (nft_identifier_bad(table))
+                                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Invalid NFT table name %s.", table);
+
+                        if (nft_identifier_bad(set))
+                                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Invalid NFT set name %s.", set);
+
+                        if (!UNIT_WRITE_FLAGS_NOOP(flags)) {
+                                r = nft_set_context_add(&c->dynamic_user_nft_set_context, &c->n_dynamic_user_nft_set_contexts, nfproto, table, set);
+                                if (r < 0)
+                                        return r;
+
+                                unit_write_settingf(
+                                                u, flags|UNIT_ESCAPE_SPECIFIERS, name,
+                                                "%s=%s:%s:%s",
+                                                name,
+                                                nfproto_name,
+                                                table,
+                                                set);
+                        }
+
+                        empty = false;
+                }
+                if (r < 0)
+                        return r;
+
+                r = sd_bus_message_exit_container(message);
+                if (r < 0)
+                        return r;
+
+                if (empty) {
+                        c->dynamic_user_nft_set_context = nft_set_context_free_many(c->dynamic_user_nft_set_context, &c->n_dynamic_user_nft_set_contexts);
                         unit_write_settingf(u, flags, name, "%s=", name);
                 }
 
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -4083,6 +4083,43 @@ static int add_shifted_fd(int *fds, size_t fds_size, size_t *n_fds, int fd, int 
         return 1;
 }
 
+static void exec_op_dynamic_user_nft_set(bool add, const ExecContext *c, uid_t uid) {
+        int r;
+
+        assert(c);
+
+        for (size_t i = 0; i < c->n_dynamic_user_nft_set_contexts; i++) {
+                NFTSetContext *s = &c->dynamic_user_nft_set_context[i];
+                if (add)
+                        r = nft_set_element_add_uint32(s, uid);
+                else
+                        r = nft_set_element_del_uint32(s, uid);
+                if (r < 0)
+                        log_warning_errno(r, "%s NFT family %s table %s set %s UID " UID_FMT " failed, ignoring: %m",
+                                          add? "Adding" : "Deleting", nfproto_to_string(s->nfproto), s->table, s->set, uid);
+        }
+}
+
+static void exec_add_dynamic_user_nft_set(const ExecContext *c, uid_t uid) {
+        exec_op_dynamic_user_nft_set(true, c, uid);
+}
+
+void exec_delete_dynamic_user_nft_set(const ExecContext *c, DynamicUser *d) {
+        int r;
+        uid_t uid;
+
+        if (!d)
+                return;
+
+        r = dynamic_user_current(d, &uid);
+        if (r < 0) {
+                log_warning_errno(r, "Can't get current dynamic user, ignoring: %m");
+                return;
+        }
+
+        exec_op_dynamic_user_nft_set(false, c, uid);
+}
+
 static int exec_child(
                 Unit *unit,
                 const ExecCommand *command,
@@ -4283,6 +4320,8 @@ static int exec_child(
 
                 if (dcreds->user)
                         username = dcreds->user->name;
+
+                exec_add_dynamic_user_nft_set(context, uid);
 
         } else {
                 r = get_fixed_user(context, &username, &uid, &gid, &home, &shell);
@@ -5346,6 +5385,8 @@ void exec_context_done(ExecContext *c) {
         c->user = mfree(c->user);
         c->group = mfree(c->group);
 
+        c->dynamic_user_nft_set_context = nft_set_context_free_many(c->dynamic_user_nft_set_context, &c->n_dynamic_user_nft_set_contexts);
+
         c->supplementary_groups = strv_free(c->supplementary_groups);
 
         c->pam_name = mfree(c->pam_name);
@@ -6020,6 +6061,11 @@ void exec_context_dump(const ExecContext *c, FILE* f, const char *prefix) {
                 fprintf(f, "%sGroup: %s\n", prefix, c->group);
 
         fprintf(f, "%sDynamicUser: %s\n", prefix, yes_no(c->dynamic_user));
+        for (size_t i = 0; i < c->n_dynamic_user_nft_set_contexts; i++)
+                fprintf(f, "%sDynamicUserNFTSet: %s:%s:%s\n", prefix,
+                        nfproto_to_string(c->dynamic_user_nft_set_context[i].nfproto),
+                        c->dynamic_user_nft_set_context[i].table,
+                        c->dynamic_user_nft_set_context[i].set);
 
         strv_dump(f, prefix, "SupplementaryGroups", c->supplementary_groups);
 
--- a/src/core/load-fragment.c
+++ b/src/core/load-fragment.c
@@ -6607,3 +6607,20 @@ int config_parse_cgroup_nft_set(
 
         return config_parse_nft_set(unit, filename, line, section, section_line, lvalue, ltype, rvalue, &c->nft_set_context, &c->n_nft_set_contexts, u);
 }
+
+int config_parse_dynamic_user_nft_set(
+                const char *unit,
+                const char *filename,
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
+        ExecContext *c = data;
+        Unit *u = userdata;
+
+        return config_parse_nft_set(unit, filename, line, section, section_line, lvalue, ltype, rvalue, &c->dynamic_user_nft_set_context, &c->n_dynamic_user_nft_set_contexts, u);
+}
--- a/src/core/service.c
+++ b/src/core/service.c
@@ -1877,6 +1877,9 @@ static void service_enter_dead(Service *s, ServiceResult f, bool allow_restart) 
         /* Get rid of the IPC bits of the user */
         unit_unref_uid_gid(UNIT(s), true);
 
+        /* Delete DynamicUserNFTSet= */
+        exec_delete_dynamic_user_nft_set(&s->exec_context, s->dynamic_creds.user);
+
         /* Release the user, and destroy it if we are the only remaining owner */
         dynamic_creds_destroy(&s->dynamic_creds);
 
--- a/src/shared/bus-unit-util.c
+++ b/src/shared/bus-unit-util.c
@@ -16,6 +16,7 @@
 #include "exec-util.h"
 #include "exit-status.h"
 #include "fileio.h"
+#include "firewall-util.h"
 #include "hexdecoct.h"
 #include "hostname-util.h"
 #include "in-addr-util.h"
@@ -432,6 +433,91 @@ static int bus_append_ip_address_access(sd_bus_message *m, int family, const uni
                 return r;
 
         return sd_bus_message_close_container(m);
+}
+
+static int bus_append_nft_set(sd_bus_message *m, const char *field, const char *eq) {
+        int r;
+
+        assert(m);
+
+        if (isempty(eq)) {
+                r = sd_bus_message_append(m, "(sv)", field, "a(iss)", 0);
+                if (r < 0)
+                        return bus_log_create_error(r);
+
+                return 1;
+        }
+
+        r = sd_bus_message_open_container(m, SD_BUS_TYPE_STRUCT, "sv");
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = sd_bus_message_append_basic(m, SD_BUS_TYPE_STRING, field);
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = sd_bus_message_open_container(m, 'v', "a(iss)");
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = sd_bus_message_open_container(m, 'a', "(iss)");
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        for (;;) {
+                _cleanup_free_ char *word = NULL;
+                int family;
+
+                r = extract_first_word(&eq, &word, ":", 0);
+                if (r == -ENOMEM)
+                        return log_oom();
+                if (r < 0)
+                        return log_error_errno(r, "Failed to parse %s: %m", field);
+                if (isempty(word)) {
+                        log_error("Failed to parse %s", field);
+                        return 0;
+                }
+
+                family = nfproto_from_string(word);
+                if (family < 0)
+                        return log_error_errno(family, "Failed to parse %s: %m", field);
+
+                r = extract_first_word(&eq, &word, ":", EXTRACT_CUNESCAPE|EXTRACT_UNESCAPE_SEPARATORS);
+                if (r == -ENOMEM)
+                        return log_oom();
+                if (r < 0)
+                        return log_error_errno(r, "Failed to parse %s: %m", field);
+                if (isempty(word) || isempty(eq)) {
+                        log_error("Failed to parse %s", field);
+                        return 0;
+                }
+
+                _cleanup_free_ char *unescaped = NULL;
+                ssize_t l;
+
+                l = cunescape(eq, 0, &unescaped);
+                if (l < 0)
+                        return log_error_errno(l, "Failed to unescape %s= value: %s", field, eq);
+
+                r = sd_bus_message_append(m, "(iss)", family, word, eq);
+
+                r = sd_bus_message_close_container(m);
+                if (r < 0)
+                        return bus_log_create_error(r);
+        }
+        r = sd_bus_message_close_container(m);
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = sd_bus_message_close_container(m);
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        r = sd_bus_message_close_container(m);
+        if (r < 0)
+                return bus_log_create_error(r);
+
+        return 1;
 }
 
 static int bus_append_cgroup_property(sd_bus_message *m, const char *field, const char *eq) {
@@ -2050,6 +2136,9 @@ static int bus_append_execute_property(sd_bus_message *m, const char *field, con
 
                 return 1;
         }
+
+        if (STR_IN_SET(field, "DynamicUserNFTSet"))
+                return bus_append_nft_set(m, field, eq);
 
         return 0;
 }
--- a/man/org.freedesktop.systemd1.xml
+++ b/man/org.freedesktop.systemd1.xml
@@ -2785,6 +2785,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly b DynamicUser = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
+      readonly a(iss) DynamicUserNFTSet = [...];
+      @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly b RemoveIPC = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly a(say) SetCredential = [...];
@@ -3331,6 +3333,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
     <!--property Group is not documented!-->
 
     <!--property DynamicUser is not documented!-->
+
+    <!--property DynamicUserNFTSet is not documented!-->
 
     <!--property RemoveIPC is not documented!-->
 
@@ -3939,6 +3943,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
     <variablelist class="dbus-property" generated="True" extra-ref="Group"/>
 
     <variablelist class="dbus-property" generated="True" extra-ref="DynamicUser"/>
+
+    <variablelist class="dbus-property" generated="True" extra-ref="DynamicUserNFTSet"/>
 
     <variablelist class="dbus-property" generated="True" extra-ref="RemoveIPC"/>
 
@@ -4679,6 +4685,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2esocket {
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly b DynamicUser = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
+      readonly a(iss) DynamicUserNFTSet = [...];
+      @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly b RemoveIPC = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly a(say) SetCredential = [...];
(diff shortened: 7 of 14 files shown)


src/core/dbus-execute.c
```diff
@@ -1244,6 +1263,7 @@ const sd_bus_vtable bus_exec_vtable[] = {
         SD_BUS_PROPERTY("LogRateLimitBurst", "u", bus_property_get_unsigned, offsetof(ExecContext, log_ratelimit_burst), SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("LogExtraFields", "aay", property_get_log_extra_fields, 0, SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("LogNamespace", "s", NULL, offsetof(ExecContext, log_namespace), SD_BUS_VTABLE_PROPERTY_CONST),
+        SD_BUS_PROPERTY("CgroupNFTSet", "(iss)", property_get_cgroup_nft_set, 0, SD_BUS_VTABLE_PROPERTY_CONST),
```
poettering: in the properties we so fare spelt cgroup as "ControlGroup", please do so here too

Also, given this is a cgroup property this should really be part of CGroupContext, not ExecContext. i.e scope units should have that (as they have CGroupContext, but no ExecContext.

src/core/dbus-execute.c
```diff
@@ -3868,6 +3888,50 @@ int bus_exec_context_set_transient_property(
...
 
+        } else if (streq(name, "CgroupNFTSet")) {
+                int family;
+                const char *table, *set, *family_str;
+
+                r = sd_bus_message_read(message, "(iss)", &family, &table, &set);
+                if (r < 0)
+                        return r;
+
+                if (family > 0) {
```
poettering: hm, i think we should refuse < 0.

src/core/dbus-execute.c
```diff
@@ -3868,6 +3888,50 @@ int bus_exec_context_set_transient_property(
...
+                        family_str = nfproto_to_string(family);
+                        if (!family_str)
+                                return sd_bus_error_set(error, SD_BUS_ERROR_INVALID_ARGS, "Cgroup NFT family not valid");
+                }
+
+                if (isempty(table))
+                        return sd_bus_error_set(error, SD_BUS_ERROR_INVALID_ARGS, "Cgroup NFT table name not valid");
+
+                if (isempty(set))
+                        return sd_bus_error_set(error, SD_BUS_ERROR_INVALID_ARGS, "Cgroup NFT set name not valid");
```
poettering: some validation should take place here. i.e. we should definitely not allow control characters. I am not sure what naming regime nft enforces, but at the very least string_is_safe() should be used here

src/core/dbus-execute.c
```diff
@@ -3868,6 +3888,50 @@ int bus_exec_context_set_transient_property(
...
+
+                                r = free_and_strdup(&c->cgroup_nft_table, table);
+                                if (r < 0)
+                                        return r;
+
+                                r = free_and_strdup(&c->cgroup_nft_set, set);
+                                if (r < 0)
+                                        return r;
+
+                                unit_write_settingf(u, flags, name, "%s=%s:%s:%s", name, family_str, table, set);
```
poettering: this of course will fall apart if people include newlines in the strings. So at the very least strng_is_safe() should be used.

But then names that include `:` will still break the logic here. Hence, mabye use cescape() and force them to be escaped.

src/core/execute.c
```diff
@@ -5147,6 +5148,19 @@ int exec_spawn(Unit *unit,
                 }
         }
 
+        if (context->cgroup_nft_family > 0) {
```
poettering: maybe write this as `context->cgroup_nft_family != NFPROTO_UNSPEC`? i.e. use the symbolic names wherever possible,

src/core/execute.c
```diff
@@ -5147,6 +5148,19 @@ int exec_spawn(Unit *unit,
...
+                r = nft_set_element_add_uint32(context->cgroup_nft_family,
+                                               context->cgroup_nft_table,
+                                               context->cgroup_nft_set,
+                                               unit->cgroup_id);
```
poettering: this should be done at cgroup realizatio time, not when we exec a process into it.

src/core/execute.c
```diff
@@ -5147,6 +5148,19 @@ int exec_spawn(Unit *unit,
...
+                if (r < 0)
+                        log_info("Adding NFT family %s table %s set %s cgroup %" PRIu64 " failed, ignoring: %m",
```
poettering: use `log_warning_errno(r, …` here

i.e. always pass the error you actually got. (otherwise %m will print garbage). And use LOG_WARN log level for log messages that indicate some form of failure that we handle gracefully and ignore.

src/core/load-fragment.c
```diff
@@ -6524,3 +6525,75 @@ int config_parse_tty_size(
...
+                return 0;
+        }
+
+        r = extract_first_word(&rvalue, &word, ":", 0);
+        if (r == -ENOMEM)
+                return log_oom();
+        if (r <= 0 || isempty(rvalue)) {
+                log_syntax(unit, LOG_WARNING, filename, line, r, "Failed to parse cgroup NFT set, ignoring: %s", rvalue);
+                return 0;
+        }
```
poettering: there should be some form of validation here see above, i.e. string_is_safe() at the very least.

And there should be unescaping here, so that ":" in the strings can be encoded. 

(or alternatively, ":" should be refused9

src/core/load-fragment.c
```diff
@@ -6524,3 +6525,75 @@ int config_parse_tty_size(
...
+
+        r = unit_path_printf(u, word, &resolved);
+        if (r < 0 || isempty(word)) {
+                log_syntax(unit, LOG_WARNING, filename, line, r, "Failed to resolve unit specifiers in '%s', ignoring: %m", word);
+                return 0;
+        }
+        table = strdup(resolved);
```
poettering: missing oom check.

also, you leak `table` here on failure. And why dup the string anyway? `resolved' already is yours, you can just use that.

src/core/service.c
```diff
@@ -1920,6 +1921,12 @@ static void service_enter_dead(Service *s, ServiceResult f, bool allow_restart)
         /* Reset TTY ownership if necessary */
         exec_context_revert_tty(&s->exec_context);
 
+        if (s->exec_context.cgroup_nft_family > 0)
+                (void) nft_set_element_del_uint32(s->exec_context.cgroup_nft_family,
+                                                  s->exec_context.cgroup_nft_table,
+                                                  s->exec_context.cgroup_nft_set,
+                                                  UNIT(s)->cgroup_id);
```
poettering: as above. this should be done when we destroy a cgroup, not here.

src/shared/bus-unit-util.c
```diff
@@ -2047,6 +2048,36 @@ static int bus_append_execute_property(sd_bus_message *m, const char *field, con
...
+                        int family;
+
+                        r = extract_first_word(&eq, &word, ":", 0);
+                        if (r == -ENOMEM)
+                                return log_oom();
+                        if (r <= 0 || isempty(word))
+                                return log_error_errno(r, "Failed to parse %s: %m", field);
+                        family = nfproto_from_string(word);
+                        if (family < 0)
+                                return log_error_errno(r, "Failed to parse %s: %m", field);
```
poettering: `r` does not carry an error here

src/shared/bus-unit-util.c
```diff
@@ -2047,6 +2048,36 @@ static int bus_append_execute_property(sd_bus_message *m, const char *field, con
...
+                        r = sd_bus_message_append(m, "(sv)", field, "(iss)", 0, NULL, NULL);
+                else {
+                        _cleanup_free_ char *word = NULL;
+                        int family;
+
+                        r = extract_first_word(&eq, &word, ":", 0);
+                        if (r == -ENOMEM)
+                                return log_oom();
+                        if (r <= 0 || isempty(word))
+                                return log_error_errno(r, "Failed to parse %s: %m", field);
```
poettering: if r == 0 you'll pass `r` as zero here, which will nowadays hit an assert

poettering: (similar below)

src/shared/firewall-util-nft.c
```diff
@@ -1068,3 +1069,74 @@ int fw_nftables_add_local_dnat(
...
+        [NFPROTO_BRIDGE] = "bridge",
+        [NFPROTO_INET] = "inet",
+        [NFPROTO_IPV4] = "ip",
+        [NFPROTO_IPV6] = "ip6",
+        [NFPROTO_NETDEV] = "netdev",
+};
+
+DEFINE_STRING_TABLE_LOOKUP(nfproto, int);
+
+#define NFT_SET_MSGS 3
```
poettering, line 1150: why a macro for this? why not just specify the size literally in the variable declaration, and then use ELEMENTSOF to reference the size when you need it?

src/test/test-nft-set.c
```diff
@@ -0,0 +1,30 @@
...
+        int r;
+        assert(argc == 6);
+
+        test_setup_logging(LOG_DEBUG);
+
+        if (getuid() != 0)
+                return log_tests_skipped("not root");
+
+        int family;
+        family = nfproto_from_string(argv[2]);
```
poettering: add `assert_se(family >= 0)`?

src/test/test-nft-set.c
```diff
@@ -0,0 +1,30 @@
...
+
+        if (streq(argv[1], "add"))
+                r = nft_set_element_add_uint32(family, argv[3], argv[4], strtol(argv[5], NULL, 0));
```
poettering: strtol() is now allowed in our sources. use safe_atou32() instead

src/shared/firewall-util-nft.c
topimiettinen: I have no preference, but similar `#define`s are used elsewhere in this file.

src/core/execute.c
topimiettinen: Problem with `NFPROTO_xxx` is that `#including` `<linux/netfilter.h>` where they are defined causes errors with redefinitions of IP types:
```
In file included from ../src/basic/in-addr-util.h:4,
                 from ../src/basic/ether-addr-util.h:9,
                 from ../src/shared/json.h:12,
                 from ../src/shared/varlink.h:6,
                 from ../src/core/manager.h:18,
                 from ../src/core/emergency-action.h:27,
                 from ../src/core/unit.h:13,
                 from ../src/core/bpf-firewall.h:6,
                 from ../src/core/cgroup.c:12:
/usr/include/netinet/in.h:31:8: error: redefinition of ‘struct in_addr’
   31 | struct in_addr
      |        ^~~~~~~
In file included from /usr/include/linux/netfilter.h:7,
                 from ../src/core/cgroup.c:4:
../src/basic/linux/in.h:89:8: note: originally defined here
   89 | struct in_addr {
      |        ^~~~~~~
```

arianvp:
Hmm it seems that in more recent `nft` you can actually match on Cgroup paths directly? Wouldn't that be the way to go?

e.g.:
https://man.archlinux.org/man/nft.8.en#SOCKET_EXPRESSION
```
table inet x {
  chain y {
    type filter hook input priority filter; policy accept;
    socket cgroupv2 level 2 "system.slice/systemd-network.service" counter
  }
}
```

topimiettinen:
I was wondering why the filter wouldn't work as expected even though the cgroup numbers are correctly inserted to the set. Looking at [here](https://github.com/torvalds/linux/blob/23d04328444a8fa0ca060c5e532220dac8e8bc26/net/netfilter/nft_meta.c#L174), [here](https://github.com/torvalds/linux/blob/23d04328444a8fa0ca060c5e532220dac8e8bc26/include/linux/cgroup-defs.h#L783) and [here](https://github.com/torvalds/linux/blob/23d04328444a8fa0ca060c5e532220dac8e8bc26/include/linux/cgroup-defs.h#L764), it seems that `meta cgroup` operates with `net_cls` cgroup v1 controller.

The cgroupv2 version, `socket cgroupv2 level x "path"` has the same problems as cgroupv1 ID had, the path (/sys/fs/cgroup/path) is resolved immediately when the rules are loaded and this fails since the cgroups don't exist yet. This is exactly what https://github.com/mk-fg/systemd-cgroup-nftables-policy-manager tries to solve. The NFT set approach can't work since strings can't be used in sets.

Maybe cgroupv2 could be made to work with bit more invasive approach: PID1 would install filters which match cgroups (once they are realized). The filter would then set socket marks. Firewall rules would then be able to use the socket marks as they please.

topimiettinen:
> Hmm it seems that in more recent `nft` you can actually match on Cgroup paths directly? Wouldn't that be the way to go?

The path is resolved when the firewall rules are loaded and if the cgroup path doesn't exist (since the service hasn't been started yet), it will fail. This makes them equally useless directly.

Perhaps there's a way (PID1 would manage cgroupv2 rules which use socket marks and store the marks in sets so that rules can reference them).

arianvp:
Yeh sounds to me like upstreaming something along the lines of https://github.com/mk-fg/systemd-cgroup-nftables-policy-manager seems to be the way to go. Also if it turns out that `cgroup` mark only works for `net_cls` (which doesn't exist in cgroups v2)

topimiettinen:
> Yeh sounds to me like upstreaming something along the lines of https://github.com/mk-fg/systemd-cgroup-nftables-policy-manager seems to be the way to go.

It works by loading arbitrary NFT filters when a service starts. I think this level of complexity was rejected earlier, leading to simpler `IPAddressAllow=`/`IPAddressDeny=` features. Maybe this could be implemented with `ExecStartPre=+nft -f /etc/nftables-%p.conf` with no changes, if the cgroup is realized and unchanging at that point. If not, perhaps that could be ensured with a new flag (like `[+-@:!]` now). Also `ExecStartPost=+nft ...` may work, though the starting order may leave a vulnerable window depending on the firewall design.

Perhaps network namespace support could be beefed up so that interfaces would be available inside the namespace without `ExecStartPre=` stuff.

> Also if it turns out that `cgroup` mark only works for `net_cls` (which doesn't exist in cgroups v2)

This is indeed the case.

topimiettinen:
The `ExecStartPre=`/`ExecStartPost=` version seems to be flaky, don't know why but the filter isn't accepted since the cgroup doesn't exist. It works if I start the service and then manually load the rules.

```
/etc/nftables-systemd-timesyncd.conf
#!/usr/sbin/nft -f

flush table inet io.systemd.systemd-timesyncd.service

table inet io.systemd.systemd-timesyncd.service {
        chain systemd-timesyncd_input {
                udp dport 123 counter accept
                counter drop
        }

        chain systemd-timesyncd_output {
                udp dport 123 counter accept
                counter reject with icmpx type admin-prohibited
        }

        chain input {
                type filter hook input priority filter; policy drop;

                socket cgroupv2 level 2 "system.slice/systemd-timesyncd.service" counter jump systemd-timesyncd_input
                # nothing else may use NTP
                udp dport 123 counter drop
                counter accept
        }

        chain output {
                type filter hook output priority filter; policy drop;

                socket cgroupv2 level 2 "system.slice/systemd-timesyncd.service" counter jump systemd-timesyncd_output
                # nothing else may use NTP
                udp dport 123 counter reject with icmpx type admin-prohibited
                counter accept
        }
}
```
```
/etc/systemd/system/systemd-timesyncd.service.d/nftables.conf 
[Service]
ExecStartPre=+/usr/sbin/nft -f /etc/nftables-%p.conf
```
```
$ sudo nft list ruleset|grep cgroup
                socket cgroupv2 level 2 30152 counter packets 0 bytes 0 jump systemd-timesyncd_input
                socket cgroupv2 level 2 30152 counter packets 10 bytes 760 jump systemd-timesyncd_output
$ ls -lid /sys/fs/cgroup/system.slice/systemd-timesyncd.service
30152 drwxr-xr-x. 2 root root 0 Feb 24 21:32 /sys/fs/cgroup/system.slice/systemd-timesyncd.service/
```

src/core/dbus-execute.c
topimiettinen: Found some discussion on [NFT Bugzilla](https://bugzilla.netfilter.org/show_bug.cgi?id=1175).

> Identifiers begin with an alphabetic character (a-z,A-Z), followed zero or more alphanumeric characters (a-z,A-Z,0-9) and the characters slash (/), backslash (\), underscore (_) and dot (.).

> The limit on set identifiers appears to be 31 characters, conflicting with the wiki on this point. It also appears that the limit on chain identifiers is also 31 characters.

topimiettinen:
Updated with `DynamicUserNFTSets=` directive for automatic insertion of UID of the dynamic user into NFT set.

topimiettinen:
In this version: `Network.IPv4NFTSet=nfproto:table:set` and `Network.IPv6NFTSet=nfproto:table:set`.

I noticed a possible problem in testing this that if the set already includes a range (10.0.0.0/8) and `networkd` changes the address (say, because of new address from DHCP) but the ranges overlap (10.0.0.0/24), kernel rejects the new set with EEXISTS. I don't know if this would happen in real life. Does `networkd` always drop the old address (`address_drop()`) before adding the new address (`address_update()`)? If not, the old addresses could be simply dropped from the set before adding new addresses.

Now there's an assumption with directive processing that only one set will ever will be used, but this should be relaxed. The admin could want to use multiple sets in different protocol tables (ARP/bridge/inet/ip/ip6/netdev).

topimiettinen:
@yuwata @poettering  While working on this, I noticed that `sd_nfnl_socket_open()`, called by `fw_nftables_init()`, eventually calls `socket_open()`, which uses `SOCK_NONBLOCK` flag when opening. I don't think the callers (in systemd, at least) expect non-blocking semantics, especially nothing handles `EINPROGRESS` or `EWOULDBLOCK` errno cases. At least when I reused nfnl code here, I saw failures due to`EINPROGRESS` and they disappeared when I explicitly used `fcntl()` to remove `O_NONBLOCK`. Is this correct analysis?

poettering:
@topimiettinen client side queuing on EAGAIN was dropped in  bbe181b489921e5afeaa2b0bc1c16c62aeef8f65. don#t remember the details.

topimiettinen:
Created #22792

src/core/dbus-execute.c
topimiettinen: Added name validation according to the above. It seems at least `-` would have been also OK.

src/core/execute.c
topimiettinen: Not applicable anymore

topimiettinen:
I dropped the cgroupv1 feature (needs more thought for cgroupv2) and I'm not 100% happy with the resolved integration (resolved needs more capabilites etc), so this version only contains integration with `DynamicUser=` (`DynamicUserNFTSet=`) and networkd (`Network.IPv4NFTSet=` and `Network.IPv6NFTSet=`).

topimiettinen:
All green! @poettering ptal.

topimiettinen:
Now with cgroupv2 integration. Sadly it seems that cgroupv2 rules are only usable on the output side of filters.

topimiettinen:
It looks like everything works except for bus operations. For example, `systemctl show systemd-timesyncd` gives error: `Failed to get properties: No such device or address`. But to me it looks like the bus talking code is pretty much identical to what other similar directives use. Any clues?

topimiettinen:
Wohoo, found and fixed the bug with bus protocol.

Meanwhile, a bug in NFT affecting use of `socket cgroupv2` expressions on input filters was also found and a patch was [submitted](https://marc.info/?l=netfilter-devel&m=164950322924987&w=2). With a fixed kernel, cgroups work nicely for input and output.

topimiettinen:
All green. @poettering PTAL.

topimiettinen:
/cc @poettering

Removed also GitHub review request.

topimiettinen:
Split commits for easier review, updated from lessons learned of #23269 and rebased to it.

topimiettinen:
mkosi / ci (arch, rolling) fails with
```
‣ Error: Workspace command bootctl install returned non-zero exit code 1.
```
Unit tests / build (CLANG, gcrypt):
```
384/1394 systemd / test-repart                                                                                TIMEOUT         30.02s   killed by signal 15 SIGTERM
```
and CentOS CI (Arch Linux):
```
TEST-60-MOUNT-RATELIMIT
```
I don't think these are related (nothing to do with networking). @mrc0mmand do they look familiar?

mrc0mmand:
Hah, you've hit the bingo of known issues:

> mkosi / ci (arch, rolling) fails with
> 
> ```
> ‣ Error: Workspace command bootctl install returned non-zero exit code 1.
> ```

That's #23485.

> Unit tests / build (CLANG, gcrypt):
> 
> ```
> 384/1394 systemd / test-repart                                                                                TIMEOUT         30.02s   killed by signal 15 SIGTERM
> ```

Spurious, but known timeout.

> and CentOS CI (Arch Linux):
> 
> ```
> TEST-60-MOUNT-RATELIMIT
> ```

Aaand that's #23424.

> I don't think these are related (nothing to do with networking). @mrc0mmand do they look familiar?

I'll retrigger the jobs and will cross my fingers for better luck :-)

topimiettinen:
@mrc0mmand All green!

@yuwata @poettering please review.

topimiettinen:
I converted my firewalls to use cgroups for examining packets: first by cgroup level 1 (`system.slice` / `user.slice`), then by L4 protocol and port, and finally again by cgroup level 2 (for system services). The order could be equally cgroup levels 1 & 2 first and then by L4 protocol and port for the service.
```
table inet filter {
        set system_slice {
                type cgroupsv2
        }

        set user_slice {
                type cgroupsv2
        }

        set systemd_networkd {
                type cgroupsv2
        }
...
        chain dhcpd_output {
                meta skuid != systemd-network reject with icmpx type admin-prohibited
....
                accept
        }
...
        chain tcp_output_sys {
                tcp dport { 547, 548, 647, 847, 7911 } socket cgroupv2 level 2 @systemd_networkd jump dhcpd_output
...
                reject with icmpx type admin-prohibited
        }
...
        chain system_slice_out {
                meta l4proto vmap { tcp : jump tcp_output_sys, udp : jump udp_output_sys }
                reject with icmpx type admin-prohibited
        }
...
        chain input {
                type filter hook input priority filter; policy drop;
...
                ct state new socket cgroupv2 level 1 @system_slice jump system_slice_in
                ct state new socket cgroupv2 level 1 @user_slice jump user_slice_in
...
        }
...
        chain output {
                type filter hook output priority filter; policy drop;
...
                ct state new socket cgroupv2 level 1 @system_slice jump system_slice_out
                ct state new socket cgroupv2 level 1 @user_slice jump user_slice_out
...
        }
...
}
```
This kind of setup allows very fine firewall control over individual services since all NFT features can be used at that point. This can be contrasted with `IPAddressDeny=`, which also gives a simple firewall for services, or more powerful `BPFProgram=` which isn't very easy to use.

The feature doesn't work well for user services since the cgroups are realized by user manager, which doesn't have privileges to fill in the NFT sets. That doesn't matter much for server systems. I suppose a trusted SUID program like Firejail could manage the sets for desktop apps.

Flushing the NFT rulesets also clears the NFT sets. Then PID1 only repopulates them when the cgroup is restarted and that isn't ever going to happen for `system.slice` / `user.slice`, so the filters may become invalid. A workaround could be that PID1 would also repopulate the sets on for example `daemon-reload` or a DBus call, but it could be also argued that there could be a feature on NFT side to keep the sets, the admin should be careful when flushing rulesets, or that kernel should use cgroup paths instead of cgroup IDs.

DaanDeMeyer:
@topimiettinen Did this get another look by @poettering or @yuwata?

yuwata:
Same here. This needs final review. Temporary setting 'merged-but-needs-fixing' label.

man/systemd.network.xml
```diff
@@ -1141,6 +1141,39 @@ NetLabel=system_u:object_r:localnet_peer_t:s0</programlisting>
           and the reverse operation when the IPv4 address is deconfigured.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>IPv4NFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
+        <term><varname>IPv6NFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
+        <listitem>
+          <para>These settings provide a method for integrating dynamic network configuration into firewall
+          rules with NFT sets. These options expect a whitespace separated list of NFT set definitions. Each
```
keszybz, line 1150: Please link to some explanation of NFT.

src/basic/parse-util.c
```diff
@@ -750,3 +750,38 @@ int parse_loadavg_fixed_point(const char *s, loadavg_t *ret) {
...
+        return true;
+}
+
+static bool nft_next_char_bad(const char c) {
+        if ((c >= 'a' && c <= 'z') ||
+            (c >= 'A' && c <= 'Z') ||
+            (c >= '0' && c <= '9') ||
+            c == '/' || c == '\\' || c == '_' || c == '.')
+                return false;
+        return true;
```
keszybz, line 767: Maybe call this `next_char_good()` (positive conditions are easier to grok, we generally prefer positive booleans and checks), and just write
```C
return (c >= 'a' && c <= 'z') ||
       (c >= 'A' && c <= 'Z') ||
       ...;
```

src/shared/firewall-util-nft.c
```diff
@@ -885,19 +890,78 @@ static int fw_nftables_add_masquerade_internal(
...
+        if (r < 0)
+                return r;
+
+        r = nft_set_element_op_in_addr(nfnl, table, set,
+                                       add, nfproto, af, address, prefixlen);
+
+        (void) in_addr_prefix_to_string(af, address, prefixlen, &addr_str);
+
+        log_debug("%s NFT family %s table %s set %s IP addresss %s", add? "Added" : "Deleted",
+                  nfproto_to_string(nfproto), table, set, strna(addr_str));
```
keszybz, line 934: Hmm, if `r < 0`, this debug stmt is wrong.

src/shared/firewall-util-nft.c
```diff
@@ -1071,3 +1135,222 @@ int fw_nftables_add_local_dnat(
...
         return fw_nftables_add_local_dnat_internal(ctx, add, af, protocol, local_port, remote, remote_port, NULL);
 }
+
+static const char *const nfproto_table[] = {
+        [NFPROTO_ARP] = "arp",
+        [NFPROTO_BRIDGE] = "bridge",
+        [NFPROTO_INET] = "inet",
+        [NFPROTO_IPV4] = "ip",
+        [NFPROTO_IPV6] = "ip6",
+        [NFPROTO_NETDEV] = "netdev",
```
keszybz, line 1145: Please align the strings to the same columns.

src/shared/firewall-util-nft.c
keszybz: I agree that there seems to be little reason to make a define for this, which survives beyond the end of the function, and only use it once in the function.

src/shared/firewall-util-nft.c
```diff
@@ -1071,3 +1135,222 @@ int fw_nftables_add_local_dnat(
...
+        r = sd_nfnl_nft_message_add_setelem_end(transaction[tsize]);
+        if (r < 0)
+                return r;
+        ++tsize;
+        assert(tsize < ELEMENTSOF(transaction));
+        r = sd_nfnl_message_batch_end(nfnl, &transaction[tsize]);
+        if (r < 0)
+                return r;
+
+        ++tsize;
```
keszybz, line 1197: Why is `tsize` incremented twice?

src/shared/firewall-util-nft.c
```diff
@@ -1071,3 +1135,222 @@ int fw_nftables_add_local_dnat(
...
+                }
+
+                nfproto = nfproto_from_string(family_str);
+                if (nfproto < 0) {
+                        log_syntax(unit, LOG_WARNING, filename, line, 0, "Unknown NFT protocol family, ignoring: %s", family_str);
+                        return 0;
+                }
+
+                if (nft_identifier_bad(table))
+                        return log_syntax(unit, LOG_WARNING, filename, line, 0, "Invalid table name %s, ignoring", table);
```
keszybz, line 1336: Here too, `cunescape` probably needs to be used.

src/shared/firewall-util-nft.c
```diff
@@ -1071,3 +1135,222 @@ int fw_nftables_add_local_dnat(
...
+
+                if (nft_identifier_bad(set))
+                        return log_syntax(unit, LOG_WARNING, filename, line, 0, "Invalid set name %s, ignoring", set);
```
keszybz, line 1339: Same.

src/shared/firewall-util-nft.c
topimiettinen: See similar uses in for example [above line 1091](https://github.com/systemd/systemd/blob/9c3e2a2a97d41f3a758c5815421122fc3e3dd3fb/src/shared/firewall-util-nft.c#L1091), `tsize` is used to keep track of used space in the array and `assert()`s guard that the array size isn't exceeded at any point.

topimiettinen:
Created https://github.com/systemd/systemd/pull/23712 for the fixes.

man/systemd.network.xml
```diff
@@ -1141,6 +1141,39 @@ NetLabel=system_u:object_r:localnet_peer_t:s0</programlisting>
           and the reverse operation when the IPv4 address is deconfigured.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>IPv4NFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
+        <term><varname>IPv6NFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
```
poettering, line 1147: BTW, I think it would make sense to implicitly use the network file name as the set name, if not specified, to push people to a clear nomenclature where each .network file gets its own set. i.e. if NFTSet= is set to just "netdev:filter" in a file foobar.network then this would be equivalent to "netdev:filter:foobar"

poettering: BTW, does it really make sense to have two distinct settings for IPv4 and IPv6 here? why is this in the lvalue instead of the rvalue? Can you explain? 

Can't the two kind of addresses be added to the same set? (no nft pro here, as you might guess).

it appears icky to me to allow ipv4 and ipv configuration here to be entirely orthogonal.

man/systemd.resource-control.xml
```diff
@@ -1173,6 +1173,35 @@ DeviceAllow=/dev/loop-control
           </para>
         </listitem>
       </varlistentry>
+      <varlistentry>
+        <term><varname>ControlGroupNFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
```
poettering, line 1177: As in the .network patch: I think the `set` part should be optional, and we should imply the unit name there, so that set names are by default derived from the unit names.

And I am not sure this should carry "Controlgroup" in the name really. that is an implementation detail I think → #24255

src/core/cgroup.c
```diff
@@ -1226,6 +1234,46 @@ static void cgroup_apply_firewall(Unit *u) {
...
+
+        for (size_t i = 0; i < c->n_nft_set_contexts; i++) {
+                NFTSetContext *s = &c->nft_set_context[i];
+                r = nft_set_element_add_uint64(s, u->cgroup_id);
+                if (r < 0)
+                        log_warning_errno(r, "Adding NFT family %s table %s set %s cgroup %" PRIu64 " failed, ignoring: %m",
+                                 nfproto_to_string(s->nfproto),
+                                 s->table,
+                                 s->set,
+                                 u->cgroup_id);
```
poettering, line 1253: you must validate `u->cgroup_id` first here, it's not guaranteed to be initialized, as we require privs for that, we might not have

src/core/cgroup.c
```diff
@@ -1226,6 +1234,46 @@ static void cgroup_apply_firewall(Unit *u) {
...
+
+        for (size_t i = 0; i < c->n_nft_set_contexts; i++) {
+                NFTSetContext *s = &c->nft_set_context[i];
+                r = nft_set_element_del_uint64(s, u->cgroup_id);
+                if (r < 0)
+                        log_warning_errno(r, "Deleting NFT family %s table %s set %s cgroup %" PRIu64 " failed, ignoring: %m",
+                                 nfproto_to_string(s->nfproto),
+                                 s->table,
+                                 s->set,
+                                 u->cgroup_id);
```
poettering, line 1273: ditto

src/basic/parse-util.c
```diff
@@ -750,3 +750,38 @@ int parse_loadavg_fixed_point(const char *s, loadavg_t *ret) {
 
         return store_loadavg_fixed_point(i, f, ret);
 }
+
+static bool nft_first_char_bad(const char c) {
+        if ((c >= 'a' && c <= 'z') ||
+            (c >= 'A' && c <= 'Z'))
+                return false;
+        return true;
+}
```
poettering, line 759: this is weird. In all other cases where we validate strings we keep lists of what is good, not what is bad, and the functions hence are called xyz_valid() or xyz_is_valid() or valid_xyz() or so. This here is the only one that is inverted. Could you please swap this around to match the general style?

src/core/load-fragment.c
```diff
@@ -6520,3 +6522,88 @@ int config_parse_tty_size(
...
+        }
+
+        for (const char *p = rvalue;;) {
+                r = extract_many_words(&p, ":", EXTRACT_CUNESCAPE, &family_str, &table, &set, NULL);
+                if (r == -ENOMEM)
+                        return log_oom();
+                if (r == 0)
+                        break;
+                if (r != 3) {
+                        log_syntax(unit, LOG_WARNING, filename, line, r, "Failed to parse NFT set, ignoring: %s", p);
```
poettering, line 6558: r is not necessarily an error here, but positive (e.g. 1 or 2). but log_syntax will consider it an error code and pass it to ERRNO.

Hence, the 5th argument must be chosen more carefully here

src/core/cgroup.c
```diff
@@ -1226,6 +1234,46 @@ static void cgroup_apply_firewall(Unit *u) {
         (void) bpf_firewall_install(u);
 }
 
+static void cgroup_apply_nft_set(Unit *u) {
+        int r;
+        CGroupContext *c;
+
+        assert(u);
+
```
poettering, line 1242: this should not even be attempted if we aren't in cgroupv2 unified mode

man/systemd.exec.xml
```diff
@@ -3163,6 +3163,40 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
     </variablelist>
   </refsect1>
 
+  <refsect1>
+    <title>Firewall Integration</title>
+    <variablelist class='unit-directives'>
+
+      <varlistentry>
+        <term><varname>DynamicUserNFTSet=</varname><replaceable>family</replaceable>:<replaceable>table</replaceable>:<replaceable>set</replaceable></term>
```
poettering, line 3171: this sounds awfully specific to me. I mean, I can see the usefulness, but why limit this to dynamic users? Why not do that for any UID here? it's really unnecessarily restricting to just say dynamic users.

Given this leaks into the unit file syntax, this really should be fixed before 252

Also, as in the other case I am sure the third part of the expression should be derived from the unit name automatically, if not specified, to push people to a "standardized" name of sets. And the examples provided should always make use of these default namings, instead of making up their own

src/shared/bus-unit-util.c
```diff
@@ -434,6 +435,91 @@ static int bus_append_ip_address_access(sd_bus_message *m, int family, const uni
...
+                }
+
+                _cleanup_free_ char *unescaped = NULL;
+                ssize_t l;
+
+                l = cunescape(eq, 0, &unescaped);
+                if (l < 0)
+                        return log_error_errno(l, "Failed to unescape %s= value: %s", field, eq);
+
+                r = sd_bus_message_append(m, "(iss)", family, word, eq);
```
poettering, line 502: missing error handling

src/shared/bus-unit-util.c
```diff
@@ -434,6 +435,91 @@ static int bus_append_ip_address_access(sd_bus_message *m, int family, const uni
...
+                        return log_error_errno(family, "Failed to parse %s: %m", field);
+
+                r = extract_first_word(&eq, &word, ":", EXTRACT_CUNESCAPE|EXTRACT_UNESCAPE_SEPARATORS);
+                if (r == -ENOMEM)
+                        return log_oom();
+                if (r < 0)
+                        return log_error_errno(r, "Failed to parse %s: %m", field);
+                if (isempty(word) || isempty(eq)) {
+                        log_error("Failed to parse %s", field);
+                        return 0;
```
poettering, line 492: return 0?

should be `return log_error_errno(SYNTHETIC_ERRNO(EINVAL), …`

poettering: (similar above)

src/core/execute.c
```diff
@@ -4284,6 +4321,8 @@ static int exec_child(
                 if (dcreds->user)
                         username = dcreds->user->name;
 
+                exec_add_dynamic_user_nft_set(context, uid);
```
poettering, line 4324: please don't do this from the child, but PID 1, i.e. we send the resolved UID/GID to PID 1 further down, look for  send_user_lookup(). Put the registration in the NFT tables at the receiving end of that.

PR #20538, opened 2021-08-25
hwdb: Add force-release for HP Omen 15 calculator key.

Illuminat10n:
The key doesn't create release event so I came up with this fix to make it work properly.

PR #22588, opened 2022-02-22
hwdb: Force release calculator key on all HP OMEN laptops

xddxdd:
I encountered the same problem as #20538 on my HP OMEN 17 laptop, but my keyboard has a different identifier:

```text
dmi:bvnAMI:bvrF.25:bd12/08/2021:br15.25:efr86.39:svnHP:pnOMENbyHPLaptop17t-ck000:pvr:rvnHP:rn88F7:rvr86.39:cvnHP:ct10:cvrChassisVersion:sku2Y4A3AV:
```

Change the rule to apply to all HP OMEN laptops, as I guess this is the common behavior.


keszybz:
LGTM too.

PR #22590, opened 2022-02-22
systemctl: Show how long a service ran for after it exited in status output

daandemeyer:


src/systemctl/systemctl-show.c
```diff
@@ -433,6 +433,13 @@ static void print_status_info(
                                FORMAT_TIMESTAMP_STYLE(until_timestamp, arg_timestamp_style),
                                FORMAT_TIMESTAMP_RELATIVE(until_timestamp));
                 }
+
+                if (STRPTR_IN_SET(i->active_state, "inactive", "failed")) {
+                        usec_t duration;
+
+                        duration = i->active_exit_timestamp - i->active_enter_timestamp;
+                        printf("   Duration: %s\n", FORMAT_TIMESPAN(duration, USEC_PER_SEC));
```
daandemeyer: Is the `USEC_PER_SEC` correct here? I'm confused on how the accuracy argument for FORMAT_TIMESPAN works

src/systemctl/systemctl-show.c
```diff
@@ -433,6 +433,13 @@ static void print_status_info(
                                FORMAT_TIMESTAMP_STYLE(until_timestamp, arg_timestamp_style),
                                FORMAT_TIMESTAMP_RELATIVE(until_timestamp));
                 }
+
+                if (STRPTR_IN_SET(i->active_state, "inactive", "failed")) {
+                        usec_t duration;
+
+                        duration = i->active_exit_timestamp - i->active_enter_timestamp;
```
poettering, line 445: maybe as safety check also in the conditionalization of this field, that both timestamps are set and the exit timestamp actually larger than the enter timestamp, and if not suppress the output of this entry.

poettering: hmm, so you are using the 'active' timestamps here. i.e. you cover the time the service was fully up. Type=oneshot services will never be "active" though, they only have an "activating" phase, and once that is complete they are immediately dead again. Question is, what to do about those. Two options:

1. maybe use the non-inactive time. In that case we'd include ExecStartPre=/ExecStart= and ExecStop= times and stuff in the "duration". And the Type=oneshot services would get useful output too.

2. accept that this is the way it is and ensure that we suppress output in the type=inactive case. This should not require extra work, if the above safety checks are added, as for Type=oneshot the active-enter timestamp never is set after all..

I am kind leaning towards option 2, but I figure people will be surprised. But I'd accept it.

src/systemctl/systemctl-show.c
poettering: well depends. the question is what level of accuracy would you expect people want to see here? I figure msec accuracy would be good, hence would probably pass MSEC_PER_SEC.

src/systemctl/systemctl-show.c
daandemeyer: Went for the second option and added the necessary checks.

src/systemctl/systemctl-show.c
```diff
@@ -433,6 +433,15 @@ static void print_status_info(
                                FORMAT_TIMESTAMP_STYLE(until_timestamp, arg_timestamp_style),
                                FORMAT_TIMESTAMP_RELATIVE(until_timestamp));
                 }
+
+                if (STRPTR_IN_SET(i->active_state, "inactive", "failed") && i->active_enter_timestamp > 0 &&
```
poettering: maybe use: `timestamp_is_set(i->active_enter_timestamp) && timestamp_is_set(i->active_exit_timestamp) && i->active_exit_timestamp >= i->active_enter_timestamp`

poettering:
lgtm, thanks!

DaanDeMeyer:
Checking the type was a non-trivial change so dropping the green label and requesting another review.

src/systemctl/systemctl-show.c
```diff
@@ -433,6 +434,18 @@ static void print_status_info(
                                FORMAT_TIMESTAMP_STYLE(until_timestamp, arg_timestamp_style),
                                FORMAT_TIMESTAMP_RELATIVE(until_timestamp));
                 }
+
+                if (!streq_ptr(i->type, "target") &&
```
poettering: so far we just did `!endswith(i->id, ".target")` for checks like this. Should be good enough for this too, and simpler

DaanDeMeyer:
Fixed the target check as suggested, upgrading the green label

keszybz:
LGTM too.

bluca (approved):

PR #22591, opened 2022-02-22
ci: build systemd without optional dependencies

evverx:
to catch issues like https://github.com/systemd/systemd/pull/22585#issuecomment-1047640155

`-Dskip-deps=true` can be used to make sure that systemd is buildable without some dependencies
but other than that it should make it easier to build it with MSan without
having to compile all the dependencies with MSan.

### DIFF
--- a/test/fuzz/meson.build
+++ b/test/fuzz/meson.build
@@ -7,10 +7,11 @@ sanitize_address_undefined = custom_target(
                    project_source_root,
                    '@OUTPUT@',
                    'fuzzers',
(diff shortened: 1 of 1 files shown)


evverx:
@mrc0mmand judging by https://github.com/evverx/systemd/actions/runs/1885441036 (where `systemd` failed to compile without https://github.com/systemd/systemd/commit/c04c7c10c0cb42835f51d7f431842eb948119a67) it seems to be working.

mrc0mmand (approved):

PR #22597, opened 2022-02-22
ci: fix clang-13 installation

mrc0mmand:
For some reason Ubuntu Focal repositories now have `llvm-13` virtual
package which can't be installed, but successfully fools our check,
resulting in no clang/llvm being installed...

```
$ apt show llvm-13
Package: llvm-13
State: not a real package (virtual)
N: Can't select candidate version from package llvm-13 as it has no candidate
N: Can't select versions from package 'llvm-13' as it is purely virtual
N: No packages found

$ apt install --dry-run llvm-13
Reading package lists... Done
Building dependency tree
Reading state information... Done
Package llvm-13 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'llvm-13' has no installation candidate
```

/cc @jamacku

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -78,12 +78,12 @@ if [[ "$COMPILER" == clang ]]; then
     # ATTOW llvm-11 got into focal-updates, which conflicts with llvm-11
     # provided by the apt.llvm.org repositories. Let's use the system
     # llvm package if available in such cases to avoid that.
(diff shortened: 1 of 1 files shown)


bluca (approved):

keszybz:
That's just ... strange.

But LGTM too.

keszybz:
The tests that this was supposed to fix now pass, so let's merge.

PR #22598, opened 2022-02-22
Adapt to threaded children in delegated unit

Werkov:
The delegated cgroups on unified hierarchy may have descendants that configure
cgroup.type="threaded", this makes the top unit's cgroup.type="domain
threaded". systemd creates auxiliary child subcgroups to satisfy "no processes
in inner-node" cgroup constraint. However, any new child cgroup is created with
with cgroup.type="domain invalid" by default. It is possible to switch the
auxiliary child cgroup into cgroup.type="threaded" since systemd won't use it
for any (domain) controllers.  And the same time, this doesn't restrict users
who wish to have the delegated root cgroup with cgroup.type="domain
threaded".

### DIFF
--- a/docs/CGROUP_DELEGATION.md
+++ b/docs/CGROUP_DELEGATION.md
@@ -260,11 +260,12 @@ your service's cgroup. This is necessary because by turning on delegation we
 have to assume that the cgroup delegated to your service is now an *inner*
 cgroup, which means that it may not directly contain any processes. Hence, if
 your service has any of these four settings set, you must be prepared that a
-`.control/` subcgroup might appear, managed by the service manager. This also
-means that your service code should have moved itself further down the cgroup
-tree by the time it notifies the service manager about start-up readiness, so
-that the service's main cgroup is definitely an inner node by the time the
-service manager might start `ExecStartPost=`.)
+`.control/` subcgroup might appear, managed by the service manager (that
+includes provisioning of their resource consumption or transitional overcommit).
+This also means that your service code should have moved itself further down
+the cgroup tree by the time it notifies the service manager about start-up
+readiness, so that the service's main cgroup is definitely an inner node by the
+time the service manager might start `ExecStartPost=`.)
 
 ## Three Scenarios
 


test/TEST-70-DELEGATE-THREADED/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+
+TEST_DESCRIPTION="Test threaded cgroup in unit with Delegate=yes"
```
bluca: Can you fold this into the existing TEST-19-DELEGATE? Each new test has additional setup/teardown costs

poettering:
Hmm, so you are saying we should run these auxiliary tools in a threaded cgroup if need be? I'd rathre not do that. Threaded cgroups are weird. People should hide them down the tree, and we should not be starting to manage them ourselves. They are implementaiton details of services, and we should *not* participate in them.

Why bother? where's the problem with inserting an intermediary cgroup to solve this cleanly?

poettering:
So, I am pretty strongly against this PR. Threads are concepts we should never have to deal with, we deal in process and invoke processes, and thus threaded cgroups are something we should never ever touch ourselves, and in particular not create. There's a much better and simpler fix: just expect the services to hide their stuff down the tree. Details how this stuff is done inside of theose daemons should not leak to us and be abstracted away.

poettering:
(I'd be more open to a patch that changes how we set things up for delegated units that have more ExecYZ= than ExecStart= set, i.e. where we ever can get into the position where we need to drop more processes in the cgroup. i.e. we could say that if any ExecXYZ= except for ExecStart= are set we'll put the ExecStart= process in a subcgroup `payload/` or so. That way people can do there whatever they want and we never make changes to that subcrgoup. That would be a reasonably clean approach I think)

Werkov:
Thanks for your feedback on the concept (despite negative). I still have faith in it, so I've addressed the code feedback (force-push) and below are some arguments to the discussion.

systemd and threads
- `cgroup.procs`
  - has not defined content for threaded cgroups, however, the root of the threaded tree (`"domain threaded"` is not purely threaded) simply lists all processes of the subtree, so it can be considered an ordinary leaf cgroup by systemd,
  - allows writes also in the threaded subtree -- migration of whole processes (that's what systemd would need to rely on).
- `cgroup.threads` serves for fine-grained thread handling. I agree systemd shouldn't be dealing with those [1].

The additional wrap-cgroup
- the hierarchy levels come at a cost (depending on controller, workload and absolute depth) so they should not be added without reason or opt-out,
- if it were automatically created by systemd, its name becomes sort of API,
- in absolute terms it doesn't buy anything, the auxiliary `.control` dents into resources (e.g. cpusets) dedicated to the delegation root anyway (and that must be accounted for, that I added into the delegation docs)
-   one purpose I find this useful for is to shift weighted scheduling, if the auxiliary `.control` needed to compete as an equal party to whole delegated subtree (but I'd prefer this to be user's opt-in).

So what do you think?

[1] Neglecting historical kernel bug workaround in `cg_kill()`.

test/TEST-70-DELEGATE-THREADED/test.sh
Werkov: I wasn't aware of the costs. Good idea! (I've moved the test in new version.)

test/units/testsuite-19.sh
```diff
@@ -36,6 +36,35 @@ systemd-run --wait --unit=test4.service true
 # And now check again, "io" should have vanished
 grep -qv io /sys/fs/cgroup/system.slice/cgroup.controllers
 
+
+# And now for something (not) completely different...
+if [ ! -f /sys/fs/cgroup/init.scope/cgroup.type ] ; then
+	echo "Skippint TEST-19-DELEGATE threads test, cgroup v2 doesn't support cgroup.type" >&2
```
keszybz, line 42: typo here

test/units/testsuite-19.sh
```diff
@@ -36,6 +36,35 @@ systemd-run --wait --unit=test4.service true
...
+SERVICE_PATH="$(mktemp /etc/systemd/system/test-delegate-XXX.service)"
+SERVICE_NAME="${SERVICE_PATH##*/}"
+
+cat >"$SERVICE_PATH" <<EOF
+[Service]
+Delegate=true
+ExecStartPre=/bin/mkdir /sys/fs/cgroup/system.slice/$SERVICE_NAME/subtree
+ExecStartPre=/bin/bash -c "echo threaded >/sys/fs/cgroup/system.slice/$SERVICE_NAME/subtree/cgroup.type"
+ExecStart=/bin/sleep 86400
+ExecReload=/bin/echo pretending to reload
```
keszybz, line 55: Please drop the `/bin/` prefixes, they just make the whole thing harder to read.

Also `sh` instead of `bash` to save some cycles.

src/shared/cgroup-setup.c
```diff
@@ -328,6 +328,32 @@ int cg_create_and_attach(const char *controller, const char *path, pid_t pid) {
...
+        if (r < 0)
+                return r;
+
+        if (!streq_skip_trailing_chars(contents, "domain invalid", NEWLINE))
+                return 0;
+
+        r = write_string_file(fs, "threaded", WRITE_STRING_FILE_DISABLE_BUFFER);
+        if (r < 0)
+                return r;
+        return 0;
```
keszybz, line 354: Collapse those into `return write_string_file(…)` ?

keszybz (commented):
I think we should merge this. Threaded mode is in the kernel, and most likely it's here to stay. This patch doesn't make our code more complicated, but it allows delegated subtrees to be set up more easily. It also removes a fairly complicated failure mode, making it more likely that users set things up properly and don't bother us with bug reports and questions. (There's also the technical motivation that @Werkov describes that deep nesting is problematic.) So overall, I'd qualify this as a minor complication on our side that avoid the need for bigger complications in other places.

poettering:
I still very much disagree with this. We should never create threaded cgroups. The origin of the issue here is that in delegated cgroups with ExecReload=/ExecStop= we actually have two managers of that cgroup: ourselves for the control subcgroup and the service itself, since we delegated it to it. I am pretty we should fix that sooner rather than later, since this is likely to always conflict in one way or another, simply because two cooks spoil the food. The whole model of cgroupsv2 is designed around single-writer, where every part of the cgroup tree is managed by a single manager. While we tell everybody that that's how it is, we actually don't follow the rule here. 

I can see this becoming an issue at many other places, for example if people want to use assign fixed guaranteed resources to subcgroups of their service but then everything gets fucked once our control processes are added in their own subcgroup that lacks any resource settings, and thus will affect the ones of the service itself. i.e. it's always a problem if stuff managed by two managers is scheduled agaist each other.

So I think we should fix this properly: let's stop putting subgroups into delegated services from our side. i.e. let's not just fix this for the threaded cgroup mess, but for everything else too: any service that has more ExecXYZ= than ExecStart= set should get its own intermediary subcrgoup called "payload" or so, and that's the one that is delegated. And we leave our own stuff in "control".

There's prior art for this in our tree: if you invoke nspawn in --keep-unit mode (i.e. when it shall take over the unit it is invoked in), then we'll create two subcgroups: `payload/` for the container payload, and that's the one that is delegated. And then `supervisor/` for nspawn itself.

poettering:
And yes, we shouldn't nest needlessly, but I think in this case it's kinda necessary I think, to fix the clean logic of delegation.

poettering:
/cc @htejun maybe you have an opinion on this?

TLDR: currently, if people ask systemd for cgroup delegation on their service, and then put some subcgroups they create below into threaded mode, systemd will fail to run ExecReload=/ExecStop= programs, since so far it simply added them as a new subcgroup of the service, and that will fail now since we'd do so in process mode, not threaded mode, and cgroup siblings can't be of different threaded/process modes...

Question is what to do: this PR suggests to just put the ExecREload=/ExecStop= processes into a threaded subcgroup instead. I think that's a bad idea, and instead we should just move everything one level down: i.e. if you actually define a service with both ExecStart= *and* ExecReload=/ExecStop=, then we'll not delegate the service's main cgroup to you, but only a child of it, maybe called `payload/`. And we'd then run ExecREload=/ExecStop= programs in a sibling cgroup of that, maybe called `control/`. That way we'd have one extra level of cgroups for services that actually do this, but the delegation model is entirely clean again: we'll not create subcrgoups below delegated cgroups from PID 1 anymore, and that's a really good thing I believe.

htejun:
Both options make sense to me.

* While threaded cgroups are a bit nasty, at least from outer systemd's POV, creating a normal sub-cgroup for support command execution and creating a threaded one doesn't make meaningful difference. So, if maintaining the status quo of the outer systemd creating sub-cgroup inside a delegated cgroup is okay, I don't think extending it to sub-cgroups makes it meaningfully worse.
* That said, it is a bit nasty for outer manager to reach into a delegated cgroup and create sub-cgroup as it does make the delegation layer rather fuzzy. So, it is cleaner to be creating sub-cgroups for payload and control workload from the get go. I'll get back to the performance overhead problem later but even ignoring that, I wonder whether this would be surprising for existing users and the behavior might be difficult to understand if the sub-cgroup creation is conditional on the specification of ExecReload/ExecStop.
* As for performance overehead, from kernel side, the goal is making reasonable levels of nesting cheap enough to ignore in most cases. For memory and IO control (w/ iocost at least) and the other misc controllers, this is already the case - whether an application is nested two levels or five levels shouldn't make any meaningful performance difference even in workloads which push hard on the respective resources. Unfortunately, we have a significant nesting overhead problem with the cpu controller which makes extra layers of nesting undesirable. FWIW, there are efforts to remediate this issue as it's PITA for a lot of use cases. It's difficult to make concrete predictions tho.

So, if the current behavior of creating sub-cgroup inside a delegated cgroup is okay, I don't think threaded cgroup is a strong reason to move away from it. It's not the prettiest but not too horrible either.

If folks wanna fix up the delegation boundary problem properly, yeah, it'd be the cleanest to introduce an extra layer (whether conditional or always). I wonder whether it'd make sense to allow specifying a unit for Reload/Stop commands so that they can run in their separate cgroups outside the delegated one so that folks who want both to avoid extra nesting and Exec* have a way out.

poettering:
(btw, in systemd itself we ship two units that have delegation enabled: user@.service (i.e. the per-user systemd instance) and systemd-nspawn@.service (i.e. our own container mgr). Both use only ExecStart= and nothing else, so nesting things like I suggested would mean that most systems would look exactly as before).

Werkov:
Thanks for identifying the additional aspects.

If I wanted to defend my original idea I'd say:
* ExecStop= runs and the end of lifecycle, the service is going to terminate, so messing with resources is without consequences :-p,
* ExecReload= would be disallowed config with Delegate=yes.

Now some diagrams,
```
L1)
foo.service     limits configured
` payload       no limit, parent cap
` .control      no limit, parent cap

L2)
foo.service     no limits
` payload       limits configured
` .control      no limits

W1)
foo.service     weight configured
` payload       implicit half
` .control      implicit half

A1)
foo.service     allocs configured
` payload       no allocs, parent inherit
` .control      no allocs, parent inherit -> clash
```

IMO only W1 is where nested `payload` wrap is adding value. (L1 is no different from wrapless, L2 is weird, A1 is doomed anyway).

At the same time, I don't like Delegate=yes XOR ExecReload= myself, so I will consider the suggestion of conditional nesting when the intrusive ExecXYZ= is specified (intrusive := running after subtree was given up by PID 1 and not at the end of lifecycle). There is a technical cornercase of modifying service config after it's started without wrap while new config would require the wrap and migration is no-no. But this has precedent in impossibility to take `Delegate=yes` away from a running service already.

poettering:
So, I'd like to emphasize that while right now ExecReload= and ExecStop= are the primary cases where you have both payload and control processes running in the cgroup, there's also ExecStartPost=. And ExecStopPost= might even do this too, if people turned off regular process killing.

And then there's the PAM stub process in case of PAMName= use: we need to keep PAM sessions open if this is enabled, which requires a process to stick around for it.

And I think we need to keep the door open for potentially having more control processes around. Specifically, there's a RFE somewhere do add ExecMonitor= as a new setting which allows invoking a process which when it does will cause the service to shutdown. And then there are ideas about adding metrics collection to systemd, which will probably also mean that we'll have to fork off a process per service, that takes the metrics data from cgroups nad writes it to disk or so, continiously. While I can#t say iff and how we'll implement this I think we must keep the door open that sooner or later we'll have more control processes per unit than we have now. Hence, I think we have to prepare for a world where multiple control processes exist, and then arrange our cgroup tree to match that reality instead of trying everything to avoid rearranging it, under the assumption we'll never do control processes for services that have delegation on.

Werkov:
I've implemented a generalization of the proposal here (and an extension of the threaded mode). I'm closing this PR in favor of #22937.

PR #22599, opened 2022-02-22
resolved: synthesize option

jacekmigacz:
Provides a configuration switch for RR synthesis in resolved.

### DIFF
--- a/docs/ENVIRONMENT.md
+++ b/docs/ENVIRONMENT.md
@@ -286,6 +286,11 @@ All tools:
   the installed ones. By default non-UTF-8 locales are suppressed from the
   selection, since we are living in the 21st century.
 
+`systemd-resolved`:
+
+* `$SYSTEMD_RESOLVED_SYNTHESIZE_HOSTNAME` — if set to "0", `systemd-resolved`
+  won't synthesize system hostname on both regular and reverse lookups.
+
 `systemd-sysext`:
 
 * `$SYSTEMD_SYSEXT_HIERARCHIES` — this variable may be used to override which
--- a/src/resolve/resolved-dns-synthesize.c
+++ b/src/resolve/resolved-dns-synthesize.c
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
 #include "alloc-util.h"
+#include "env-util.h"
 #include "hostname-util.h"
 #include "local-addresses.h"
 #include "missing_network.h"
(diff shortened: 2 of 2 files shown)


jacekmigacz:
There is at least this case: https://lists.freedesktop.org/archives/systemd-devel/2021-September/046856.html
Proposed solution is good, but if it's not a `resolvectl` that makes a query, then you are out of luck.

poettering:
So I'd be with a change like this to disable synthesis in resolved server-side, but I think it should be more focussed: only on the local hostname itself (leave the synthesis of localhost and such in place). And it should be controlled via env var rather than get a first class config setting. (i.e. this is a political thing: it should have a hackish feeling to turn on this option)

i.e. if people add Environment=SYSTEMD_RESOLVED_SYNTHESIZE_HOSTNAME=0 to systemd-resolved.service we would turn off the logic. Would need a simple patch to src/resolve/resolved-synthesize.c

poettering:
ping?

src/resolve/resolved-dns-synthesize.c
```diff
@@ -444,6 +445,11 @@ int dns_synthesize_answer(
                 } else if (dns_name_address(name, &af, &address) > 0) {
                         int v, w;
 
+                        if (getenv_bool("SYSTEMD_RESOLVED_SYNTHESIZE_HOSTNAME") == 0) {
+                                nxdomain = true;
```
poettering: hm, nxdomain seems a bit drastic? i.e. if you return nxdomain then this ends the lookup logic. But it appears to me, if the env var is set we simply shouldn't respond at a ll, i.e. just do `continue` here...

also, shouldn't the manager_is_own_hostname() branch above get the same treatment? to disable both directions of local hostname resolution? from address to name and from name to address?

jacekmigacz: For the `nxdomain`, you are right. That is too drastic. I will just `continue` iteration as suggested.

I guess both directions makes sense. I will do it for normal and reverse lookups (and change ENVIRONMENT.md).

docs/ENVIRONMENT.md
```diff
@@ -286,6 +286,11 @@ All tools:
   the installed ones. By default non-UTF-8 locales are suppressed from the
   selection, since we are living in the 21st century.
 
+`systemd-resolved`:
+
+* `$SYSTEMD_RESOLVED_SYNTHESIZE_HOSTNAME=1` — if set, `systemd-resolved` will
+  synthesize system hostname on both regular and reverse lookups.
```
poettering: the text is wrong. It will synthesize this by default. But if you set the variable to zero then it won't.

jacekmigacz: I think I wrongly deducted previous ENVs. Fixed now.

PR #22600, opened 2022-02-22
make more use of timestamp_is_set()

poettering:


poettering:
So the coccinelle script doens't work entirely. For example, in src/basic/time-util.c there are a couple of cases we should catch but don't. Similar in src/test/test-time-util.c

@mrc0mmand maybe you have an idea?

yuwata (approved):
LGTM, though I am not familiar with .cocci file format.

coccinelle/timestamp-is-set.cocci
```diff
@@ -0,0 +1,51 @@
...
+constant USEC_INFINITY = USEC_INFINITY;
+@@
+- x <= 0 || x == USEC_INFINITY
++ !timestamp_is_set(x)
+@@
+expression x;
+constant USEC_INFINITY = USEC_INFINITY;
+@@
+- x <= 0 || x >= USEC_INFINITY
++ !timestamp_is_set(x)
```
keszybz, line 72: Wouldn't a regexp be more effective? ;)

keszybz:
LGTM too.

mrc0mmand:
> So the coccinelle script doens't work entirely. For example, in src/basic/time-util.c there are a couple of cases we should catch but don't. Similar in src/test/test-time-util.c
> 
> @mrc0mmand maybe you have an idea?

For some strange reason the Cocci engine does the commutative transformation in `x == 0 || x == USEC_INFINITY` for both expressions (i.e. it does both `x == 0` and `0 == x`, and `x == USEC_INFINITY` and `USEC_INFINITY`, but it doesn't do the same for the `||` expression (i.e. it doesn't check `x == USEC_INFINITY || x == 0` and other permutations). My Cocci knowledge is unfortunately not enough to see why it happens, though.

I rewrote the rule as
```coccinelle
@@
expression x;
constant USEC_INFINITY = USEC_INFINITY;
/* We want to stick with the literal expression in the implementation of timestamp_is_set(), i.e. in time-util.c */
position p : script:python() { p[0].file != "src/basic/time-util.h" };
binary operator xop = {>, !=};
binary operator uop = {<, !=};
@@
(
/* Handles:
 * x > 0 || x < USEC_INFINITY
 * x > 0 || x != USEC_INFINITY
 * x != 0 || x < USEC_INFINITY
 * x != 0 || x != USEC_INFINITY
 */
- ((x@p xop 0) || (x uop USEC_INFINITY))
+ timestamp_is_set(x)
|
/* Handles:
 * x < USEC_INFINITY || x > 0
 * x < USEC_INFINITY || x != 0
 * x != USEC_INFINITY || x > 0
 * x != USEC_INFINITY || x != 0
 */
- ((x@p uop USEC_INFINITY) || (x xop 0))
+ timestamp_is_set(x)
)

@@
expression x;
constant USEC_INFINITY = USEC_INFINITY;
binary operator xop = {==, <=};
binary operator uop = {==, >=};
@@
(
/* Handles:
 * x == 0 || x == USEC_INFINITY
 * x == 0 || x >= USEC_INFINITY
 * x <= 0 || x == USEC_INFINITY
 * x <= 0 || x >= USEC_INFINITY
 */
- ((x xop 0) || (x uop USEC_INFINITY))
+ !timestamp_is_set(x)
|
/* Handles:
 * x == USEC_INFINITY || x == 0
 * x == USEC_INFINITY || x <= 0
 * x >= USEC_INFINITY || x == 0
 * x >= USEC_INFINITY || x <= 0
 */
- ((x uop USEC_INFINITY) || (x xop 0))
+ !timestamp_is_set(x)
)
```

which works and looks...not that bad, but it incredibly memory hungry (I had to kill it at 20 GB RES).

So, eventually, I had to unroll it into separate statements, which works (given the number of things it found), but it's pretty verbose:

```coccinelle
@@
expression x;
constant USEC_INFINITY = USEC_INFINITY;
/* We want to stick with the literal expression in the implementation of timestamp_is_set(), i.e. in time-util.c */
position p : script:python() { p[0].file != "src/basic/time-util.h" };
@@
(
- x@p > 0 || x < USEC_INFINITY
+ timestamp_is_set(x)
|
- x@p > 0 || x != USEC_INFINITY
+ timestamp_is_set(x)
|
- x@p != 0 || x < USEC_INFINITY
+ timestamp_is_set(x)
|
- x@p != 0 || x != USEC_INFINITY
+ timestamp_is_set(x)
|
- x@p < USEC_INFINITY || x > 0
+ timestamp_is_set(x)
|
- x@p < USEC_INFINITY || x != 0
+ timestamp_is_set(x)
|
- x@p != USEC_INFINITY || x > 0
+ timestamp_is_set(x)
|
- x@p != USEC_INFINITY || x != 0
+ timestamp_is_set(x)
)

@@
expression x;
constant USEC_INFINITY = USEC_INFINITY;
@@
(
- x == 0 || x == USEC_INFINITY
+ !timestamp_is_set(x)
|
- x == 0 || x >= USEC_INFINITY
+ !timestamp_is_set(x)
|
- x <= 0 || x == USEC_INFINITY
+ !timestamp_is_set(x)
|
- x <= 0 || x >= USEC_INFINITY
+ !timestamp_is_set(x)
|
- x == USEC_INFINITY || x == 0
+ !timestamp_is_set(x)
|
- x == USEC_INFINITY || x <= 0
+ !timestamp_is_set(x)
|
- x >= USEC_INFINITY || x == 0
+ !timestamp_is_set(x)
|
- x >= USEC_INFINITY || x <= 0
+ !timestamp_is_set(x)
)

```

mrc0mmand:
oh well, I forgot the `&&` variants, which will make it even more verbose, but you get the idea :-)

yuwata:
We also have `IN_SET(usec, 0, USEC_INFINITY)` at multiple places.

poettering:
Reworked the patch, based on @mrc0mmand's suggestions. Also added the IN_SET() thing.

Please have a look!

yuwata (approved):
LGTM.

poettering:
> For some strange reason the Cocci engine does the commutative transformation in `x == 0 || x == USEC_INFINITY` for both expressions (i.e. it does both `x == 0` and `0 == x`, and `x == USEC_INFINITY` and `USEC_INFINITY == x`, but it doesn't do the same for the `||` expression (i.e. it doesn't check `x == USEC_INFINITY || x == 0` and other permutations).

In C those expressions are not necessarily commutative. They are only commutative if they are side-effect free. Because C says that execution of such expression must be terminated once the result of the whole expression is clear when executing left-to-right. i.e. if you do `foo() || bar();` and `foo()` returns true, then `bar()` is never run at all... Similar, if you do `foo() && bar()` and `foo()` returns false, then `bar()` is never run...

Now, my guess is that Coccinelle is just too dumb to determine whether the expressions here are side-effect free. Because if it doesn't know that it has to assume they have a side effect, and then it cannot swap the order. So my I guess there's some logic that we have to manually write this down in both orders.

mrc0mmand:
> > For some strange reason the Cocci engine does the commutative transformation in `x == 0 || x == USEC_INFINITY` for both expressions (i.e. it does both `x == 0` and `0 == x`, and `x == USEC_INFINITY` and `USEC_INFINITY == x`, but it doesn't do the same for the `||` expression (i.e. it doesn't check `x == USEC_INFINITY || x == 0` and other permutations).
> 
> In C those expressions are not necessarily commutative. They are only commutative if they are side-effect free. Because C says that execution of such expression must be terminated once the result of the whole expression is clear when executing left-to-right. i.e. if you do `foo() || bar();` and `foo()` returns true, then `bar()` is never run at all... Similar, if you do `foo() && bar()` and `foo()` returns false, then `bar()` is never run...
> 
> Now, my guess is that Coccinelle is just too dumb to determine whether the expressions here are side-effect free. Because if it doesn't know that it has to assume they have a side effect, and then it cannot swap the order. So my I guess there's some logic that we have to manually write this down in both orders.

Ah, indeed, that does make sense, thanks!

mrc0mmand (approved):
LGTM as well

PR #22602, opened 2022-02-22
units: drop After=systemd-networkd.service from systemd-resolved.service

poettering:
This ordering existed since resolved was first created, but there should
not be any need to order the two services against each other, as
resolved should be able to pick up networkd DNS metadata either way (as
it works with inotify in /run).

Let's drop this hence, and not cargo-cult this to eternity

Also see: https://github.com/systemd/systemd/pull/22389#issuecomment-1045978403

### DIFF
--- a/units/systemd-resolved.service.in
+++ b/units/systemd-resolved.service.in
@@ -15,7 +15,7 @@ Documentation=https://www.freedesktop.org/wiki/Software/systemd/writing-network-
 Documentation=https://www.freedesktop.org/wiki/Software/systemd/writing-resolver-clients
 
 DefaultDependencies=no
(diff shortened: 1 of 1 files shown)


keszybz:
LGTM, thanks.

yuwata:
@bluca Why do you think it needs to be backported? This does not fix any clear bugs.

bluca:
Well, if it's not needed, then it sounds like a bugfix-ish to me

keszybz:
In principle it fixes a bug, but there's also some level of risk… Rearranging things in early boot is quite fragile. So I'll drop the tag the tag for now. Let's not backport this immediately, but only if we know that it fixes a user-visible issue.

PR #22605, opened 2022-02-22
test-journal-send: fix fd leaks

yuwata:


keszybz:
Looks reasonable.

src/libsystemd/sd-journal/test-journal-send.c
```diff
@@ -90,6 +90,8 @@ static void test_journal_send(void) {
         assert_se(sd_journal_sendv(graph2, 1) == 0);
         assert_se(sd_journal_sendv(message1, 1) == 0);
         assert_se(sd_journal_sendv(message2, 1) == 0);
+
+        closelog();
```
poettering, line 98: maybe add a comment why this is added here, it's otherwise hard to see why we bother, given the fd is not really managed by use, but by libc

yuwata: Sure. A brief comment is added.

evverx:
FWIW I think it would be great if the API provided by `sd-journal` could make it possible to close its fds properly even without valgrind.

keszybz:
@evverx I agree that it'd be nice to export this. But we can do this separately, this PR is good to fix the issue.

Issue #22515, opened 2022-02-15
`networkctl reload` did not apply new prefix length for address


anitazha:
**systemd version the issue has been seen with**
 > v249.4

**Used distribution**
 > CentOS 8 stream

**Linux kernel version used** (`uname -a`)
> `5.6.13` with patches

I have a config that looks like this:
```
$ cat /etc/systemd/network/1-fb_networkd-eth0.network 
[Match]
MACAddress = 48:57:DD:05:C8:DC
Name = eth0

[Network]
Address = <primary address>/64
IPv4LLRoute = true
IPv6AcceptRA = true

[IPv6AcceptRA]
UseMTU = false
UseAutonomousPrefix = false
UseOnLinkPrefix = false
DHCPv6Client = false
```

I apply the config with `networkctl reload && networkctl reconfigure eth0` and `ip addr` shows the expected address with prefix length 64.

But if I change the line in the config from `Address = <primary address>/64` to `Address = <primary address>/128` (i.e. only the prefix length has changed), and do `networkctl reload` and `networkctl reconfigure eth0`, the address and prefix length as shown by `ip addr` is not updated. The only way I successfully got the new prefix length to apply is if I do `networkctl down eth0 && networkctl up eth0`. Is this expected?

yuwata:
Yeah, I know the issue...

Previously, once I changed the hash function for addresses to takes into account prefix length of address, but it causes an issue (though I forgot the detail...). So, now the function hashes only address. I will re-investigate the previous issue and this later.

Note, the issue should be occurs only on IPv6, as the prefix length of IPv4 addresses is already managed by the hash function.

yuwata:
Note, the previous change is 1d30fc5cb64ecba2f03fe42aa0d8c65c3decad82.

yuwata:
Fix is waiting in #22607.

anitazha:
Thanks @yuwata ! Your PR worked like a charm

PR #17851, opened 2020-12-04
network: revert previous changes to address_compare_func()

yuwata:
This partially reverts fe841414ef157f7f01d339c5d5730126e7b5fe0a and 2a236f9fc0ff8fb2152032551436fde74da7217a.

For IPv4, kernel compares the local address, prefix, and prefixlen.
For IPv6, kernel compares only the local address.
Let's follow the kernel's comparison way.

Fixes #17831.

Issue #17831, opened 2020-12-03
networkd: interfaces stuck in "configuring" state


Just4GH:
**systemd version the issue has been seen with**
 > 247.1

**Used distribution**
 > Arch Linux

**Linux kernel version used** (`uname -a`)
 > Linux * 5.9.12-arch1-1 #1 SMP PREEMPT Wed, 02 Dec 2020 16:14:56 +0000 x86_64 GNU/Linux

**CPU architecture issue was seen on**
 > x86

**Expected behaviour you didn't see**
 > All interfaces reach state "configured"

**Unexpected behaviour you saw**
 > `systemd-networkd-wait-online.service loaded failed failed Wait for Network to be Configured`
 > Directly after reboot enp7s0f0 was still configuring, restarting systemd-networkd put eno1 in persistent configuring as well.
```
# networkctl
IDX LINK     TYPE     OPERATIONAL SETUP
  1 lo       loopback carrier     unmanaged
  2 eno1     ether    routable    configuring
  3 eno2     ether    routable    configured
  4 enp7s0f0 ether    routable    configuring
  5 enp7s0f1 ether    off         unmanaged
```
> Further issues:
> Interface lo receives an unreachable route from IPv6 Prefix Delegation:
```
# ip -6 route | grep lo
::1 dev lo proto kernel metric 256 pref medium
unreachable 2001:*:*:fc::/62 dev lo metric 1024 pref medium
```

**Steps to reproduce the problem**
 > Reboot or restart systemd-networkd after upgrade to 247

**More info**
> eno1 is an internal network with prefix delegation
> enp7s0f0 is connected to a router
[eno1.network.txt](https://github.com/systemd/systemd/files/5633456/eno1.network.txt)
[enp7s0f0.network.txt](https://github.com/systemd/systemd/files/5633457/enp7s0f0.network.txt)
[networkd.log](https://github.com/systemd/systemd/files/5633458/networkd.log)

yuwata:
Could you show the result of `ip address` when these interfaces are stuck in `configuring` state?

Just4GH:
Network works normally:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
[...]

PR #22607, opened 2022-02-22
network: compare addresses more strictly

yuwata:
This re-introduce the full hash and compre functions for Address,
which was reverted https://github.com/systemd/systemd/commit/1d30fc5cb64ecba2f03fe42aa0d8c65c3decad82 (https://github.com/systemd/systemd/pull/17851).

The issue https://github.com/systemd/systemd/issues/17831, which is fixed by https://github.com/systemd/systemd/pull/17851, is handled in a different way;
simply ignore to configure the conflicted DHCPv6 address. Previously, we
warn about the conflict, but continue to configure the address anyway, and
the kernel ignores the request. So, it is not necessary to request the
conflicted address in networkd side.

This fixes the following issues:
- when a link has an address, and corresponding .network file has the
  address with different prefix length, then the prefix length specified
  in the .network file will not be applied,
- we cannot specify multiple IPv4 addresses with different prefix
  length, e.g.
  ```
  Address=10.10.10.10/24
  Address=10.10.10.10/16
  ```
  This is spurious setup, but the kernel allows it.

Fixes https://github.com/systemd/systemd/issues/22515.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -2149,6 +2149,18 @@ class NetworkdNetworkTests(unittest.TestCase, Utilities):
         call('ip netns del ns99', stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
 
     def test_address_static(self):
+        # test for #22515. The address will be removed and replaced with /64 prefix.
+        rc = call('ip link add dummy98 type dummy')
+        self.assertEqual(rc, 0)
+        rc = call('ip link set dev dummy98 up')
+        self.assertEqual(rc, 0)
+        rc = call('ip -6 address add 2001:db8:0:f101::15/128 dev dummy98')
+        self.assertEqual(rc, 0)
+        self.wait_address('dummy98', '2001:db8:0:f101::15/128', ipv='-6')
+        rc = call('ip -4 address add 10.3.2.3/16 brd 10.3.255.250 scope global label dummy98:hoge dev dummy98')
+        self.assertEqual(rc, 0)
+        self.wait_address('dummy98', '10.3.2.3/16 brd 10.3.255.250', ipv='-4')
+
         copy_unit_to_networkd_unit_path('25-address-static.network', '12-dummy.netdev')
         start_networkd()
 


bluca:
```
 391/1264 test-network                                                                                                                    FAIL             0.27s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=85 PATH=/home/runner/work/systemd/systemd/build:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin SYSTEMD_LANGUAGE_FALLBACK_MAP=/home/runner/work/systemd/systemd/src/locale/language-fallback-map /home/runner/work/systemd/systemd/build/test-network
stderr:
Assertion 'address_equal(a1, a2)' failed at src/network/test-network.c:211, function test_address_equality(). Aborting.
```

keszybz:
`subprocess.CalledProcessError: Command '['/systemd-meson-build/systemd-networkd-wait-online', '--timeout=20s', '--interface=wg99:routable', '--interface=wg98:routable', '--interface=wg97:carrier']' returned non-zero exit status 1.`
Hmm.

src/network/networkd-address.c
```diff
@@ -1980,8 +2011,10 @@ int network_drop_invalid_addresses(Network *network) {
                         address_free(dup);
                 }
 
-                /* Do not use address_hash_ops_free here. Otherwise, all address settings will be freed. */
-                r = set_ensure_put(&addresses, &address_hash_ops, address);
+                /* Use address_kernel_hash_ops here. The function address_kernel_compare_func() matches
+                 * how kernel compares addresses, and is more lenient than address_compare_func().
+                 * Hence, the logic of dedup here is stricter than the case address_hash_ops is used. */
```
keszybz: … than when address_hash_ops is used.

src/network/networkd-address.c
```diff
@@ -496,92 +550,69 @@ int address_get(Link *link, const Address *in, Address **ret) {
...
+                        continue;
+
+                if (in_addr_is_set(family, &a->in_addr_peer))
+                        continue;
+
+                if (ret)
+                        *ret = a;
+
+                return 0;
         }
```
keszybz, line 600: Maybe add a comment explaining the logic here? It's not obvious why `in_addr_peer` is checked.

src/network/networkd-dhcp6.c
```diff
@@ -181,20 +182,21 @@ static void log_dhcp6_address(Link *link, const Address *address) {
         if (existing->source == NETWORK_CONFIG_SOURCE_NDISC)
                 by_ndisc = true;
 
-        log_link_warning(link, "DHCPv6 address %s/%u (valid %s, preferred %s) conflicts the address %s/%u%s.",
+        log_link_warning(link, "DHCPv6 address %s/%u (valid %s, preferred %s) conflicts the address %s/%u%s. Ignoring the DHCPv6 address.",
```
keszybz: Pre-existing condition, but "a conflicts b" is very awkward. "a conflicts with b" would be better.
Maybe change this to:
```C
log_link_warning(link, 
                        "Ignoring DHCPv6 address %s/%u (valid %s, preferred %s) which conflicts with %s/%u%s.",
                         strna(buffer), address->prefixlen,
                         FORMAT_LIFETIME(address->lifetime_valid_usec),
                         FORMAT_LIFETIME(address->lifetime_preferred_usec),
                         strna(buffer), existing->prefixlen,
                         by_ndisc ? " assigned by NDisc" : "");
if (by_ndisc)
   log_link_info(link, "Hint: use IPv6Token= setting to change the address generated by NDISC or set UseAutonomousPrefix=no.");
```

keszybz (commented):
Looks reasonable, but I'm not familiar with the details here. The CI failure looks like it could be related.

yuwata:
@keszybz Thank you for the review. Addressed the above comments.
Now, cleanups for broadcast handling is added, so the failure in the test for wireguard should be fixed.
PTAL.

keszybz:
focal-ppc64el: TEST-50-DISSECT:                    FAIL     (610 s)
focal-s390x: something in unittests, but I don't see what
rpm-build:https://bugzilla.redhat.com/show_bug.cgi?id=2057735
centos ci: TEST-58-REPART

This looks unrelated.

keszybz:
LGTM.

PR #22609, opened 2022-02-22
udev-builtin-input_id: don't label absolute mice as pointing sticks

whot:
The Getac UX10 tablet exposes a "CUST0000:00 0EEF:C002 Mouse" device
with BTN_LEFT/RIGHT and ABS_X/Y on the i2c bus. This causes the builtin
to incorrectly label it as pointing stick (all i2c mice are
tagged as ID_INPUT_POINTING_STICK, see 3d7ac1c655ec4).

Fix this by adding a separate variable for absolute pointing
devices like the VMmouse USB mouse or this Getac tablet - this way we
skip the pointing stick check.

See https://gitlab.freedesktop.org/libinput/libinput/-/issues/743
for recordings.

### DIFF
--- a/src/udev/udev-builtin-input_id.c
+++ b/src/udev/udev-builtin-input_id.c
@@ -167,6 +167,7 @@ static bool test_pointers(sd_device *dev,
         bool finger_but_no_pen = false;
         bool has_mouse_button = false;
         bool is_mouse = false;
+        bool is_abs_mouse = false;
         bool is_touchpad = false;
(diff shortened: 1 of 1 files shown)


keszybz:
We have this:
```C
if (is_mouse && id->bustype == BUS_I2C)
                is_pointing_stick = true;
```
…
so assigning `is_abs_mouse` instead of `is_mouse` makes this conditions fail. LGTM.

PR #22614, opened 2022-02-24
timesyncd: add runtime servers

zonque:
Currently, NTP servers can be configured through static configuration files or through state files written by `systemd-networkd`. On some systems, however, there might be other sources of information about time servers such as mDNS service announcements.

To support that, this PR adds a new server address type for addresses injected through D-Bus at runtime.

### DIFF
--- a/src/timesync/timesyncd-manager.c
+++ b/src/timesync/timesyncd-manager.c
@@ -856,7 +856,7 @@ int manager_connect(Manager *m) {
                                 return 0;
                         }
 
-                        if (restart && !m->exhausted_servers && m->poll_interval_usec) {
+                        if (restart && !m->exhausted_servers && m->poll_interval_usec > 0) {
                                 log_debug("Waiting after exhausting servers.");
                                 r = sd_event_add_time_relative(m->event, &m->event_retry, CLOCK_BOOTTIME, m->poll_interval_usec, 0, manager_retry_connect, m);
                                 if (r < 0)


sielicki:
I think all this really needs is a change [here](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/man/org.freedesktop.timedate1.xml) to describe the new method.

src/timesync/timesyncd-manager.c
```diff
@@ -831,21 +831,23 @@ int manager_connect(Manager *m) {
...
-                         * fallback list. */
+                         * let's find the next one to iterate. First we try the system list, then the link list,
+                         * then the runtime list. After having processed the link list we jump back to the
+                         * system list, then to the runtime list. However, if all lists are empty, we change to
+                         * the fallback list. */
                         if (!m->current_server_name || m->current_server_name->type == SERVER_LINK) {
                                 f = m->system_servers;
                                 if (!f)
                                         f = m->link_servers;
+                                if (!f)
```
sielicki, line 837: I'm not sure this is right, `f = m->link_servers;` will never be possible because it will simply overwrite with `runtime_servers`.

src/timesync/timesyncd-manager.h
```diff
@@ -42,11 +42,13 @@ struct Manager {
         LIST_HEAD(ServerName, system_servers);
         LIST_HEAD(ServerName, link_servers);
         LIST_HEAD(ServerName, fallback_servers);
+        LIST_HEAD(ServerName, runtime_servers);
```
poettering, line 45: doesn't matter, but swap with previous lines. the "fallback" is conceptually the lst, and should probably be here, too

src/timesync/timesyncd-bus.c
```diff
@@ -23,24 +42,86 @@ static int property_get_servers(
...
+        if (manager_is_connected(m)) {
+                if (m->current_server_address &&
+                    m->current_server_address->name &&
+                    m->current_server_address->name->type == SERVER_RUNTIME &&
+                    !strv_find(names, m->current_server_address->name->string)) {
+                        manager_connect(m);
+                }
+        } else {
+                manager_connect(m);
+        }
```
poettering, line 91: as per coding style, no {} around single-line if blocks

src/timesync/timesyncd-bus.c
```diff
@@ -23,24 +42,86 @@ static int property_get_servers(
...
+
+        if (manager_is_connected(m)) {
+                if (m->current_server_address &&
+                    m->current_server_address->name &&
+                    m->current_server_address->name->type == SERVER_RUNTIME &&
+                    !strv_find(names, m->current_server_address->name->string)) {
+                        manager_connect(m);
+                }
+        } else {
+                manager_connect(m);
```
poettering: if you ignore the return value knowingly (which might be OK) then cast the return value to (void) explicitly, to make this clear (here and above)

src/timesync/timesyncd-bus.c
```diff
@@ -23,24 +42,86 @@ static int property_get_servers(
...
 
-        return sd_bus_message_close_container(reply);
+        if (m->runtime_servers)
+                m->runtime_servers_changed = true;
+
+        if (manager_is_connected(m)) {
+                if (m->current_server_address &&
+                    m->current_server_address->name &&
+                    m->current_server_address->name->type == SERVER_RUNTIME &&
+                    !strv_find(names, m->current_server_address->name->string)) {
```
poettering: strv_contains

src/timesync/timesyncd-bus.c
```diff
@@ -23,24 +42,86 @@ static int property_get_servers(
...
         }
 
-        return sd_bus_message_close_container(reply);
+        if (m->runtime_servers)
+                m->runtime_servers_changed = true;
+
+        if (manager_is_connected(m)) {
+                if (m->current_server_address &&
+                    m->current_server_address->name &&
+                    m->current_server_address->name->type == SERVER_RUNTIME &&
```
poettering: you can drop this conditionalization? if the name matches that should be enough, regardless if it was set statically or dynamically via API

src/timesync/timesyncd-bus.c
```diff
@@ -23,24 +42,86 @@ static int property_get_servers(
...
+
+        _cleanup_strv_free_ char **names = NULL;
+        Manager *m = userdata;
+        char **name;
+        int r;
+
+        assert(m);
+        assert(bus);
+        assert(message);
+
```
poettering, line 59: so, writable D-Bus properties kinda suck, except for stuff tht really is variable like and isolated from everything else. I doubt this one qualifies. hence, please just make the setter a regular method (i.e. SetDynamicServer()).

Also, please hook up with PK. See hwo bus_link_method_set_dns_servers_internal() does that in resolved.

src/timesync/timesyncd-bus.c
```diff
@@ -23,24 +42,86 @@ static int property_get_servers(
...
         if (r < 0)
                 return r;
 
-        LIST_FOREACH(names, p, *s) {
-                r = sd_bus_message_append(reply, "s", p->string);
+        while (m->runtime_servers)
+                server_name_free(m->runtime_servers);
+
+        STRV_FOREACH(name, names) {
+                r = dns_name_is_valid_or_address(*name);
```
poettering, line 65: please validate all params first. and only after that is complete aallocate ServerName objects for them. i.e. we should strive to implement all-or-nothing behaviour here.

src/timesync/timesyncd-bus.c
```diff
@@ -163,6 +244,7 @@ static const sd_bus_vtable manager_vtable[] = {
         SD_BUS_PROPERTY("LinkNTPServers", "as", property_get_servers, offsetof(Manager, link_servers), 0),
         SD_BUS_PROPERTY("SystemNTPServers", "as", property_get_servers, offsetof(Manager, system_servers), SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("FallbackNTPServers", "as", property_get_servers, offsetof(Manager, fallback_servers), SD_BUS_VTABLE_PROPERTY_CONST),
+        SD_BUS_WRITABLE_PROPERTY("RuntimeNTPServers", "as", property_get_runtime_servers, property_set_runtime_servers, 0, 0),
```
poettering: as mentioned, please make this  read-only prop, and instead add a SetRuntimeNTPServers() method for changing them.

src/timesync/timesyncd-server.c
```diff
@@ -54,6 +54,7 @@ int server_name_new(
                 const char *string) {
 
         ServerName *n, *tail;
+        const char *type_name = NULL;
```
poettering: consider using DEFINE_STRING_TABLE_LOOKUP_TO_STRING() to define lookup function with table, and use that.

poettering:
Sorry for the late review, and not sure you still want this, but conceptually looks totally OK to me. Some points review points above.

poettering:
needs policy update, and man page update

poettering:
@sielicki if this will solve your problems, consider picking this up. as i understand @zonque he now intends to pick up the servers he needs directly from resolved instead, hence I take it this PR is not that relevant for him. (correct me if i am wrong @zonque)

src/timesync/timesyncd-bus.c
zonque: Hmm, could you elaborate on the reason for this? Properties give you signals on change for free as well, and my understanding of them is that they are meant for exactly the use-case at hand: expose an internal setting to the outside world.

zonque:
> @sielicki if this will solve your problems, consider picking this up. as i understand @zonque he now intends to pick up the servers he needs directly from resolved instead, hence I take it this PR is not that relevant for him. (correct me if i am wrong @zonque)

Well I think there are other use cases for the functionality of this PR, even though we might solve our case with native mDNS support in timesyncd eventually. I'd like to finish this one up hence.

src/timesync/timesyncd-manager.c
zonque: Not sure I'm following? I'm assuming the runtime servers will only be looked at if there are no link servers and no system servers. What am I missing? :thinking:

sielicki: you are not missing anything, apologies.

src/timesync/timesyncd-bus.c
poettering: read-only properties are OK and we use them all over the place. My issue is with writable properties. for example you cannot do PolicyKit (or any other form of "slow" auth) with them, and we usually talk about system services here...

I am not asking to remove the property, i am just saying: make it read-only and instead provide a setter method, that can do polkit can report proper errors, and so on.

read-only props doesn't mean the props can't change value. It just means you cannot write to them using the generic propery set APIs. And yes, you can still fire PropertiesChanged signals as you wish also for read-only props.

poettering: or to say this differently: i think adding a generic concept of writing to props to the dbus props spec was a mistake. Generalizing reading props, and generalizing props changes is a good thing, but generalizing writing to props was a mistake. It might make sense on the user bus, where access control doesn't matter (but maybe not even then anymore, given flatpak and app sandboxing) but our stuff is almost entirely system level stuff, and thus should not really use writable props.

zonque: Okay, understood. Implemented as requested now.

zonque:
Pushed an updated branch. @sielicki @poettering PTAL again :)

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,60 @@ static int property_get_servers(
...
+
+        assert(m);
+        assert(message);
+
+        r = sd_bus_message_read_strv(message, &msg_names);
+        if (r < 0)
+                return r;
+
+        while (m->runtime_servers)
+                server_name_free(m->runtime_servers);
```
poettering: you flush out the old servers before even just validating the new ones. Ideally we'd prepare the new servers fully before flushing old the old ones to make sure things either succeed or fail entirely but never leave half-completed stuff around if they fail midway. Now, I don't think we should be super strict on this, because it sometimes complicates code more than would be good. But still, we should at the very least try to do validation of user supplied settings before we make any changes.

here it's even quite easy, just exchange this loop with the next...

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,60 @@ static int property_get_servers(
...
+                r = dns_name_is_valid_or_address(*name);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to check validity of NTP server name or address '%s': %m", *name);
+
+                if (r == 0) {
+                        log_error("Invalid NTP server name or address, ignoring: %s", *name);
+                        continue;
+                }
+
+                strv_push(&valid_names, strdup(*name));
```
poettering: strv_extend() will strdup() + strv_push() in one.

poettering: also needs oom check

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,60 @@ static int property_get_servers(
...
+        }
+
+        STRV_FOREACH(name, valid_names) {
+                r = server_name_new(m, NULL, SERVER_RUNTIME, *name);
+                if (r < 0) {
+                        while (m->runtime_servers)
+                                server_name_free(m->runtime_servers);
```
poettering: you have the same loop twice, mayb split out into a helper function `manager_flush_runtime_servers()`?

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,60 @@ static int property_get_servers(
...
+
+                        return log_error_errno(r, "Failed to add runtime server '%s': %m", *name);
+                }
+        }
+
+        if (m->runtime_servers)
+                m->runtime_servers_changed = true;
```
poettering: if an empty list was supplied this should also constitute a change, no? if something was set before…?

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,60 @@ static int property_get_servers(
...
+
+        if (manager_is_connected(m)) {
+                if (m->current_server_address &&
+                    m->current_server_address->name &&
+                    strv_contains(valid_names, m->current_server_address->name->string)) {
+                        (void) manager_connect(m);
+                }
```
poettering: no {} around single-line if blocks, according to our coding style, plesae

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,60 @@ static int property_get_servers(
...
+        Manager *m = userdata;
+        int r;
+
+        assert(m);
+        assert(message);
+
+        r = sd_bus_message_read_strv(message, &msg_names);
+        if (r < 0)
+                return r;
+
```
poettering: pk check missing

src/timesync/timesyncd-manager.c
```diff
@@ -925,6 +932,10 @@ void manager_flush_server_names(Manager  *m, ServerType t) {
         if (t == SERVER_FALLBACK)
                 while (m->fallback_servers)
                         server_name_free(m->fallback_servers);
+
+        if (t == SERVER_RUNTIME)
+                while (m->runtime_servers)
+                        server_name_free(m->runtime_servers);
```
poettering: if you add that helper i suggested, you could even call it here too

src/timesync/timesyncd-server.h
```diff
@@ -12,6 +12,7 @@ typedef enum ServerType {
         SERVER_FALLBACK,
         SERVER_LINK,
         SERVER_RUNTIME,
+        _SERVER_TYPE_MAX,
```
poettering, line 15: while you are at it, maybe also add:

```c
_SERVER_TYPE_INVALID = -EINVAL,
```

we typically have that all across our codebase, let's do thta here too

src/timesync/timesyncd-bus.c
poettering: ignore, a later commit adds that, as i see now

src/timesync/timesyncd-server.h
zonque: done

src/timesync/timesyncd-bus.c
zonque: Yes, that makes sense.

src/timesync/timesyncd-bus.c
zonque: Well, it's a two-liner, and the boilerplate it takes to make this a "public" function is larger than what we save, but okay.

Did that now.

src/timesync/timesyncd-bus.c
zonque: Right, thanks. Done.

src/timesync/timesyncd-bus.c
zonque: Swapped the snippets now.

zonque:
All comments addressed again. Thanks a bunch!

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,58 @@ static int property_get_servers(
...
+        STRV_FOREACH(name, msg_names) {
+                r = dns_name_is_valid_or_address(*name);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to check validity of NTP server name or address '%s': %m", *name);
+
```
poettering: please drop this empty line, we usually try to keep calls and there immediate error handling/return code handling together

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,58 @@ static int property_get_servers(
...
+                if (r == 0) {
+                        log_error("Invalid NTP server name or address, ignoring: %s", *name);
+                        continue;
```
poettering: Why ignore invalid entries? Please return an error to clients, i.e.:

```c
return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Invalid NTP server name or address, refusing: %s", *name);
```

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,58 @@ static int property_get_servers(
...
+                }
+
+                r = strv_extend(&valid_names, *name);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to add runtime server '%s': %m", *name);
```
poettering: the only error strv_extend() returns is ENOMEM, hence maybe just use `return log_oom()` here.

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +45,58 @@ static int property_get_servers(
...
+
+        m->runtime_servers_changed = true;
+
+        if (manager_is_connected(m)) {
+                if (m->current_server_address &&
+                    m->current_server_address->name &&
+                    strv_contains(valid_names, m->current_server_address->name->string))
+                        (void) manager_connect(m);
+        } else
+                (void) manager_connect(m);
```
poettering: seems you can merge the inner and the outer if check? they execute the same operation in the inner and the else branch after all?

poettering (changes requested):
almost there

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,69 @@ static int property_get_servers(
         return sd_bus_message_close_container(reply);
 }
 
+static int method_set_runtime_servers(sd_bus_message *message, void *userdata, sd_bus_error *error) {
```
yuwata, line 52: This should refuse to set servers when the interface is managed by networkd.
See the similar bus methods in resolved.

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,69 @@ static int property_get_servers(
...
+        _cleanup_strv_free_ char **valid_names = NULL;
+        _cleanup_strv_free_ char **msg_names = NULL;
+        Manager *m = userdata;
+        int r;
+
+        assert(m);
+        assert(message);
+
+        r = bus_verify_polkit_async(message, CAP_NET_ADMIN,
```
yuwata: I think, the second and fourth commits should be merged.

zonque: done

src/timesync/timesyncd-bus.c
zonque: done

src/timesync/timesyncd-bus.c
zonque: fair enough!

src/timesync/timesyncd-bus.c
zonque: Hmm. At the time of calling that method the interface could not be managed by networkd yet but only later, so I don't think it makes sense to guard the setting to that? The runtime logic will give link-level servers precedence over runtime servers however - isn't that enough?

zonque:
> Do you have any plan to use the new method in timedatectl?

I didn't anticipate this yet, no. Not sure if that would be the right place either :thinking:

src/timesync/timesyncd-bus.c
poettering: these servers are global, not per interface. if these were per-interface I think we should refuse them if networkd supplies them, but this doesn't apply here.

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,64 @@ static int property_get_servers(
...
+                return r;
+
+        STRV_FOREACH(name, msg_names) {
+                r = dns_name_is_valid_or_address(*name);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to check validity of NTP server name or address '%s': %m", *name);
+                if (r == 0)
+                        return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Invalid NTP server name or address, refusing: %s", *name);
+
+                r = strv_extend(&valid_names, *name);
```
poettering: hmm, valid_naes will contain the exact same entries as msg_names, no? hence just drop valid_names?

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,64 @@ static int property_get_servers(
...
+        r = bus_verify_polkit_async(message, CAP_NET_ADMIN,
+                                    "org.freedesktop.timesync1.set-runtime-servers",
+                                    NULL, true, UID_INVALID,
+                                    &m->polkit_registry, error);
+        if (r < 0)
+                return r;
+
+        if (r == 0)
+                /* Polkit will call us back */
+                return 1;
```
poettering: maybe move the polkit auth until after the basic parameter validation. i.e. polkit is "expensive" (requires UI inteaction), and hence be done when we really are about to apply a change. parsing parameters should be done first I think.

src/timesync/timesyncd-bus.c
zonque: Ah, right. Now it does. I used to ignore invalid servers, hence the intermediate list. Dropped now.

src/timesync/timesyncd-bus.c
zonque: Ok.

src/timesync/timesyncd-bus.c
yuwata: > these servers are global, not per interface.

Ah, I see. Sorry for the noise.

src/timesync/timesyncd-server.c
```diff
@@ -85,6 +85,9 @@ int server_name_new(
         } else if (type == SERVER_FALLBACK) {
                 LIST_FIND_TAIL(names, m->fallback_servers, tail);
                 LIST_INSERT_AFTER(names, m->fallback_servers, tail, n);
+        } else if (type == SERVER_RUNTIME) {
+                LIST_FIND_TAIL(names, m->runtime_servers, tail);
+                LIST_INSERT_AFTER(names, m->runtime_servers, tail, n);
```
yuwata: You can use `LIST_APPEND()`.

src/timesync/timesyncd-manager.c
```diff
@@ -825,21 +826,23 @@ int manager_connect(Manager *m) {
...
                                         f = m->link_servers;
+                                if (!f)
+                                        f = m->runtime_servers;
                         } else {
                                 f = m->link_servers;
                                 if (!f)
                                         f = m->system_servers;
-                                else
+                                if (!f)
+                                        f = m->runtime_servers;
```
yuwata: This means, if at least one link or system server is specified, then runtime server will never used. Is it true?

If this is true, the behavior is different from other daemons, e.g. systemd-resolved or systemd-networkd.
They use runtime settings with the first priority, then, (static) per-link settings, finally, system settings.

zonque: > This means, if at least one link or system server is specified, then runtime server will never used. Is it true?

This is correct and in line with my thinking, yes. Our use case is that we don't have any static servers and we don't get any from the DHCP server, hence the need for runtime servers at all.

> If this is true, the behavior is different from other daemons, e.g. systemd-resolved or systemd-networkd.
They use runtime settings with the first priority, then, (static) per-link settings, finally, system settings.

That said, I'm happy to change that, as per the description above, the actual order doesn't really matter for me. Should we go with the logic that `resolved` implements?

src/timesync/timesyncd-server.c
zonque: This is just copied from the other branches above. I added a commit now to clean that up.

src/timesync/timesyncd-manager.c
yuwata: resolved does not have a way to set runtime system servers. And if per-link runtime servers are specified, then per-link static servers are ignored. And, system servers has higher priority than per-link (runtime and static) servers.

Hence, I prefer runtime servers has higher priority:
runtime servers -> system servers (ignored if runtime servers specified) -> link servers -> (setting restart flag) -> runtime servers -> ...

yuwata: @poettering Do you have any comments or opinions on this?

poettering: i think @yuwata is right.

zonque: Okay, changed now. A 2nd pair of eyes on the logic for selecting servers would be appreciated.

src/timesync/timesyncd-manager.c
```diff
@@ -825,21 +826,23 @@ int manager_connect(Manager *m) {
...
                                         f = m->link_servers;
                         } else {
                                 f = m->link_servers;
+                                if (!f)
+                                        f = m->runtime_servers;
                                 if (!f)
                                         f = m->system_servers;
-                                else
+                                if (f)
                                         restart = false;
```
yuwata: Unset `restart` flag only when link servers exist:
```
f = m->link_server;
if (!f) {
        f = m->runtime_servers;
        if (!f)
                f = m->system_server;
} else
        restart = false;
```

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,58 @@ static int property_get_servers(
...
+                }
+        }
+
+        m->runtime_servers_changed = true;
+
+        if (!manager_is_connected(m) ||
+            (m->current_server_address &&
+             m->current_server_address->name &&
+             strv_contains(msg_names, m->current_server_address->name->string)))
+                (void) manager_connect(m);
```
yuwata: I do not understand why `manager_connect()` is conditionalized... Especially, I have no idea the second condition `(m->current_server_address && ...)`. I think it is OK to call `manager_connect()` unconditionally. At least, please add a comment about the condition.

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,58 @@ static int property_get_servers(
...
+                                    "org.freedesktop.timesync1.set-runtime-servers",
+                                    NULL, true, UID_INVALID,
+                                    &m->polkit_registry, error);
+        if (r < 0)
+                return r;
+        if (r == 0)
+                /* Polkit will call us back */
+                return 1;
+
+        manager_flush_runtime_servers(m);
```
yuwata, line 82: `manager_flush_runtime_servers()` clears previous runtime servers, and if current server is a runtime server, then disconnect from the server.

However, it does not disconnect when a static system server or link server is currently used.
I think, it is better to call `manager_set_server_name(m, NULL)` before `manager_connect()`. Then, the specified new runtime server is used.

src/timesync/timesyncd-manager.c
```diff
@@ -825,21 +826,23 @@ int manager_connect(Manager *m) {
...
                         /* Our current server name list is exhausted,
-                         * let's find the next one to iterate. First
-                         * we try the system list, then the link list.
-                         * After having processed the link list we
-                         * jump back to the system list. However, if
-                         * both lists are empty, we change to the
-                         * fallback list. */
+                         * let's find the next one to iterate. First we try the runtime list, then the system list,
+                         * then the link list. After having processed the link list we jump back to the
+                         * system list, then to the runtime list. However, if all lists are empty, we change to
```
yuwata: "jump back to the runtime list, then to the system list if no runtime server exists."

src/timesync/timesyncd-manager.c
```diff
@@ -825,21 +826,23 @@ int manager_connect(Manager *m) {
                         bool restart = true;
 
                         /* Our current server name list is exhausted,
-                         * let's find the next one to iterate. First
-                         * we try the system list, then the link list.
-                         * After having processed the link list we
-                         * jump back to the system list. However, if
-                         * both lists are empty, we change to the
-                         * fallback list. */
+                         * let's find the next one to iterate. First we try the runtime list, then the system list,
```
yuwata, line 829: "then the system list if no runtime server exists,"

src/timesync/timesyncd-bus.c
```diff
@@ -43,6 +49,58 @@ static int property_get_servers(
...
+        STRV_FOREACH(name, msg_names) {
+                r = server_name_new(m, NULL, SERVER_RUNTIME, *name);
+                if (r < 0) {
+                        manager_flush_runtime_servers(m);
+
+                        return log_error_errno(r, "Failed to add runtime server '%s': %m", *name);
+                }
+        }
+
+        m->runtime_servers_changed = true;
```
yuwata: It can be `m->exhausted_servers = true;`??

yuwata (changes requested):
Sorry for late review.

src/timesync/timesyncd-manager.c
```diff
@@ -852,6 +855,11 @@ int manager_connect(Manager *m) {
                                 return 0;
                         }
 
+                        if (m->runtime_servers_changed) {
+                                m->runtime_servers_changed = false;
+                                restart = false;
+                        }
+
                         if (restart && !m->exhausted_servers && m->poll_interval_usec) {
```
yuwata: This is not your fault, but could you update the last condition to `m->poll_interval_usec > 0`?

src/timesync/timesyncd-bus.c
zonque: Okay, this is a change in behavior but I'm happy with that. My intention was to not disconnect from currently connected servers but only to memorize runtime servers for cases when there is no active connection.

But it doesn't matter in my case, and probably not in anyone else's either. After all, the whole point of introducing the runtime servers is to support systems where all other means of providing that information fails, which means that all other lists are empty anyway.

Changed now.

src/timesync/timesyncd-manager.c
zonque: Added a commit to the PR now, but I'm curious: how can it be negative?

yuwata (approved):
Sorry for late review again. LGTM now. (I fixed super minor things.)

sielicki:
Thank you for your work on this.

PR #22615, opened 2022-02-24
network: dhcp server: support pxe boot systems

### LINKED ISSUE #4403: RFE: permit configuration of TFTP boot dhcp settings in networkd
### Submission type
- [ ] Bug report
- [X] Request for enhancement (RFE)

Hello. As systemd-networkd already implements a Dhcp server, why not complete it with a Tftp server to get a basic PXE server all-in-one? This could be useful to get rid of dnsmasq, for instance. Sorry if I'm saying a stupid thing. Thanks!

yuwata:
Closes #4403.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -2627,6 +2627,23 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
       </varlistentry>
 
       <varlistentry>
+        <term><varname>NextServer=</varname></term>
+
+        <listitem>
+          <para>Takes an IPv4 address. Configures the next server used by e.g. PXE boot systems.
+          Defaults to unset.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>Filename=</varname></term>
+
+        <listitem>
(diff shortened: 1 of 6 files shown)


bluca (approved):

poettering:
Hmm, could we please rename these options in the .network file? "NextServer=" and "Filename=" are just too unspecific. In particular "Filename=" is like the most unspecific name thinkable. ;-)

Maybe name this option PXEServer= and PXEFilename=? 

Or better TFTPServer= and TFTPFilename=?

i.e. just add some prefix to make clear what this is about. I am aware the underlying DHCP options are poorly named, but I think we can do better here.

poettering:
To keep track of this I posted #22661

poettering:
Hmm, so this takes an IPv4 address, but isn't the DHCP option 66 kinda where it's at these days? i.e. googling for PXE it appears that the name you specify there is more relevant.

Hence maybe amend the patch to either take an IP address *or* a domain name. If it parses as IP address fill in the saddr field of the DHCP header, and otherwise fill in DHCP option 66?

PR #22618, opened 2022-02-25
network: dhcp-server: refuse unsafe filename

### LINKED ISSUE #22661: networkd: rename Filename=/NextServer= settings
The options are a bit too generic for my taste. Let's make them a bit more specific, see:

https://github.com/systemd/systemd/pull/22615#issuecomment-1059037246

THis is purely about how we expose the options to users, not about the underlying dhcp options.

yuwata:
Follow-up for #22615.
Fixes #22661.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -2626,19 +2626,38 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
       </varlistentry>
 
       <varlistentry>
-        <term><varname>NextServer=</varname></term>
+        <term><varname>BootServerAddress=</varname></term>
 
         <listitem>
-          <para>Takes an IPv4 address. Configures the next server used by e.g. PXE boot systems.
+          <para>Takes an IPv4 address of the boot server used by e.g. PXE boot systems. When specified,
+          the address is set to the <literal>siaddr</literal> field of the DHCP message header. See
+          <ulink url="https://www.rfc-editor.org/rfc/rfc2131.html">RFC 2131</ulink> for more details.
           Defaults to unset.</para>
         </listitem>
       </varlistentry>
 
       <varlistentry>
-        <term><varname>Filename=</varname></term>
+        <term><varname>BootServerName=</varname></term>
 
         <listitem>
-          <para>Takes a path or url to a file loaded by e.g. a PXE boot loader. Defaults to unset.</para>
+          <para>Takes a name of the boot server used by e.g. PXE boot systems. When specified, the
+          server name is set to the DHCP option 66. See
+          <ulink url="https://www.rfc-editor.org/rfc/rfc2132.html">RFC 2132</ulink> for more details.
+          Defaults to unset.</para>
+          <para>Note that typically one of
+          <varname>BootServerName=</varname>/<varname>BootServerAddress=</varname> is sufficient to be
+          set, but both can be set too, if desired.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>BootFilename=</varname></term>
+
+        <listitem>
+          <para>Takes a path or URL to a file loaded by e.g. a PXE boot loader. The specified path is
+          set to the DHCP option 67. See
+          <ulink url="https://www.rfc-editor.org/rfc/rfc2132.html">RFC 2132</ulink> for more details.
+          Defaults to unset.</para>
         </listitem>
       </varlistentry>
 
--- a/src/libsystemd-network/sd-dhcp-server.c
+++ b/src/libsystemd-network/sd-dhcp-server.c
@@ -12,6 +12,7 @@
 #include "alloc-util.h"
 #include "dhcp-internal.h"
 #include "dhcp-server-internal.h"
+#include "dns-domain.h"
 #include "fd-util.h"
 #include "in-addr-util.h"
 #include "io-util.h"
@@ -163,7 +164,8 @@ static sd_dhcp_server *dhcp_server_free(sd_dhcp_server *server) {
 
         sd_event_unref(server->event);
 
-        free(server->filename);
+        free(server->boot_server_name);
+        free(server->boot_filename);
         free(server->timezone);
 
         for (sd_dhcp_lease_server_type_t i = 0; i < _SD_DHCP_LEASE_SERVER_TYPE_MAX; i++)
@@ -272,24 +274,40 @@ sd_event *sd_dhcp_server_get_event(sd_dhcp_server *server) {
         return server->event;
 }
 
-int sd_dhcp_server_set_next_server(sd_dhcp_server *server, const struct in_addr *next_server) {
+int sd_dhcp_server_set_boot_server_address(sd_dhcp_server *server, const struct in_addr *address) {
         assert_return(server, -EINVAL);
 
-        if (next_server)
-                server->next_server = *next_server;
+        if (address)
+                server->boot_server_address = *address;
         else
-                server->next_server = (struct in_addr) {};
+                server->boot_server_address = (struct in_addr) {};
 
         return 0;
 }
 
-int sd_dhcp_server_set_filename(sd_dhcp_server *server, const char *filename) {
+int sd_dhcp_server_set_boot_server_name(sd_dhcp_server *server, const char *name) {
+        int r;
+
+        assert_return(server, -EINVAL);
+
+        if (name) {
+                r = dns_name_is_valid(name);
+                if (r < 0)
+                        return r;
+                if (r == 0)
+                        return -EINVAL;
+        }
+
+        return free_and_strdup(&server->boot_server_name, name);
+}
+
+int sd_dhcp_server_set_boot_filename(sd_dhcp_server *server, const char *filename) {
         assert_return(server, -EINVAL);
 
         if (filename && (!string_is_safe(filename) || !ascii_is_valid(filename)))
                 return -EINVAL;
 
-        return free_and_strdup(&server->filename, filename);
+        return free_and_strdup(&server->boot_filename, filename);
 }
 
 int sd_dhcp_server_stop(sd_dhcp_server *server) {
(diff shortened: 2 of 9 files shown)


src/basic/string-util.c
```diff
@@ -982,6 +982,9 @@ bool string_is_safe(const char *p) {
         /* Checks if the specified string contains no quotes or control characters */
 
         for (const char *t = p; *t; t++) {
+                if ((unsigned char) *t >= 128) /* no invalid ASCII characters, i.e values between 0 and 127, inclusive. */
+                        return false;
+
```
keszybz: I don't think this is good. `string_is_safe` is used in a dozen different places, and at least e.g. in `json.c` or `kdb-util.c` we shouldn't refuse non-ASCII. (I didn't look in detail.) Part of the problem is the name of this function: "_is_safe" can mean different things to different people and in different contexts, so the name is essentially meaningless.

src/libsystemd-network/sd-dhcp-server.c
```diff
@@ -286,7 +285,7 @@ int sd_dhcp_server_set_next_server(sd_dhcp_server *server, const struct in_addr
 int sd_dhcp_server_set_filename(sd_dhcp_server *server, const char *filename) {
         assert_return(server, -EINVAL);
 
-        if (filename && !ascii_is_valid(filename))
+        if (filename && !string_is_safe(filename))
```
keszybz: I think this should just be `filename && !string_is_safe() or !ascii_is_valid()` instead. Then you don't need to change the meaning of `string_is_safe()`.

src/network/networkd-network-gperf.gperf
```diff
@@ -306,7 +306,7 @@ DHCPServer.SendVendorOption,                 config_parse_dhcp_send_option,
 DHCPServer.SendOption,                       config_parse_dhcp_send_option,                            0,                             offsetof(Network, dhcp_server_send_options)
 DHCPServer.BindToInterface,                  config_parse_bool,                                        0,                             offsetof(Network, dhcp_server_bind_to_interface)
 DHCPServer.NextServer,                       config_parse_in_addr_non_null,                            AF_INET,                       offsetof(Network, dhcp_server_next_server)
-DHCPServer.Filename,                         config_parse_string,                                      0,                             offsetof(Network, dhcp_server_filename)
+DHCPServer.Filename,                         config_parse_safe_string,                                 0,                             offsetof(Network, dhcp_server_filename)
```
keszybz: ... and here add and use `config_parse_safe_ascii_string()` ?

keszybz (changes requested):
Can you add a comment in the patch *why* non-ASCII characters need to be refused?

yuwata:
@keszybz Thank you for the review. `config_parse_safe_ascii_string()` is introduced, and dropped the change to `string_is_safe()`. PTAL.

src/shared/conf-parser.c
```diff
@@ -910,6 +910,38 @@ int config_parse_safe_string(
...
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
+
+        char **s = data;
```
keszybz: `ASSERT_PTR(data)` ?

src/shared/conf-parser.c
```diff
@@ -910,6 +910,38 @@ int config_parse_safe_string(
...
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+        assert(data);
+
+        if (!string_is_safe(rvalue)) {
+                log_syntax(unit, LOG_WARNING, filename, line, 0, "Specified string contains unsafe characters, ignoring: %s", rvalue);
```
keszybz: This print statement will not work well. If the string has control chars, the log will end up garbled. Either journalctl will refuse to show it w/o `--all`, or even the message could be truncated at the first newline. The argument needs to be passed through cescape() before printing, or not printed at all. I think it's worth the trouble here to print it, because the user most likely inserted some ccs by mistake in the config and a good error message will be useful to them.

poettering:
lgtm, but I agree with @keszybz cescaping() the passed string before outputting it does make sense.

yuwata:
Updated.
- Now the log messages contains escaped string.
- addressed #22661.

man/systemd.network.xml
```diff
@@ -2626,19 +2626,27 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
...
         </listitem>
       </varlistentry>
 
       <varlistentry>
-        <term><varname>Filename=</varname></term>
+        <term><varname>BootFilename=</varname></term>
 
         <listitem>
-          <para>Takes a path or url to a file loaded by e.g. a PXE boot loader. Defaults to unset.</para>
+          <para>Takes a path or url to a file loaded by e.g. a PXE boot loader. The specified path is
```
poettering: url → URL

src/network/networkd-dhcp-server.c
```diff
@@ -744,3 +749,45 @@ int config_parse_dhcp_server_address(
...
+
+        if (isempty(rvalue)) {
+                network->dhcp_server_boot_server = (struct in_addr) {};
+                network->dhcp_server_boot_server_name = mfree(network->dhcp_server_boot_server_name);
+                return 0;
+        }
+
+        r = in_addr_from_string(AF_INET, rvalue, &a);
+        if (r >= 0) {
+                network->dhcp_server_boot_server = a.in;
```
poettering: sould probably also do `network->dhcp_server_boot_server_name = mfree(network->dhcp_server_boot_server_name);` here, no?

src/network/networkd-dhcp-server.c
```diff
@@ -744,3 +749,45 @@ int config_parse_dhcp_server_address(
...
+        }
+
+        r = dns_name_is_valid(rvalue);
+        if (r <= 0) {
+                log_syntax(unit, LOG_WARNING, filename, line, r,
+                           "Specified invalid boot server address or name, ignoring assignment: %s", rvalue);
+                return 0;
+        }
+
+        return free_and_strdup_warn(&network->dhcp_server_boot_server_name, rvalue);
```
poettering: conversely, should also do `network->dhcp_server_boot_server = (struct in_addr) {};` here

poettering:
lgtm, just some minor nits

src/network/networkd-dhcp-server.c
yuwata: This is intended. Then, we can configure both siaddr and option 66 through BootServer= setting.
```
BootServer=192.168.0.1
BootServer=boot-server-name
```

yuwata:
Extended a man page to emphasize the above point. PTAL.
(Dropping the green label tentatively, as the point is 'addressed' in a non-requested way.)

poettering:
Hmm, that makes the setting kinda magic, no? i.e. usually we have two kinds of settings: the "overriding" ones, and the "extending" ones. i.e. the former mean whatever was set before is forgotten entirely if used twice (example: Nice= in service unit files). And the latter mean you can use the setting n times and you get n entries, for any n.

but this one would be different: you can use it multiple times, but if you do it would override all previous settings that match a certain syntax, and two syntaxes for that are supported. i.e. if you pass an ipv4 address we'll override the ipv4 field, and if you pass the dns name you override the dns name.

To me this smells as if the logic is not normalized. i.e. this either should be two separate settings, or it should be one setting that can set both fields, or override both.

Hence, maybe:

```
BootServer=yyy
BootServer=47.11.8.16
BootServer=xxx 47.11.8.15
```

i.e. accept one or two words. one word an ipv4 address, the other a domain name. using the setting would always reset both fields. if you actually want to set both fields, you have to pass two words.

and we should refuse any attempts to pass more than two words, or two ip addresses, or two dns names.

yuwata:
> To me this smells as if the logic is not normalized. i.e. this either should be two separate settings, or it should be one setting that can set both fields, or override both.

Then, I prefer to make the setting into two.

In the updated version,
- introduce config_parse_dns_name() and config_parse_hostname(),
- split BootServer= into BootServerAddress= and BootServerName=.

PTAL.

man/systemd.network.xml
```diff
@@ -2626,19 +2626,35 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
...
       </varlistentry>
 
       <varlistentry>
-        <term><varname>Filename=</varname></term>
+        <term><varname>BootServerName=</varname></term>
 
         <listitem>
-          <para>Takes a path or url to a file loaded by e.g. a PXE boot loader. Defaults to unset.</para>
+          <para>Takes a name of the boot server used by e.g. PXE boot systems. When specified, the
+          server name is set to option 66. See
```
poettering: maybe say *DHCP* option 66

man/systemd.network.xml
```diff
@@ -2626,19 +2626,35 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
...
+          Defaults to unset.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>BootFilename=</varname></term>
+
+        <listitem>
+          <para>Takes a path or URL to a file loaded by e.g. a PXE boot loader. The specified path is
+          set to option 67. See
```
poettering: similar

man/systemd.network.xml
```diff
@@ -2626,19 +2626,35 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
...
       <varlistentry>
-        <term><varname>Filename=</varname></term>
+        <term><varname>BootServerName=</varname></term>
 
         <listitem>
-          <para>Takes a path or url to a file loaded by e.g. a PXE boot loader. Defaults to unset.</para>
+          <para>Takes a name of the boot server used by e.g. PXE boot systems. When specified, the
+          server name is set to option 66. See
+          <ulink url="https://www.rfc-editor.org/rfc/rfc2132.html">RFC 2132</ulink> for more details.
+          Defaults to unset.</para>
```
poettering, line 2646: maybe add a comment that typically one of BootServerName=/BootServerAddress= is sufficient to be set, but both can be set too, if desired.

poettering:
lgtm, only minor nits

yuwata:
@poettering Thank you for the review. All above points are addressed. Upgrading the green label.

PR #22619, opened 2022-02-25
network: several netdev cleanups

yuwata:
Split out of #22557.


src/network/netdev/ipvlan.c
```diff
@@ -80,16 +80,11 @@ const NetDevVTable ipvtap_vtable = {
 };
 
 IPVlanMode link_get_ipvlan_mode(Link *link) {
-        NetDev *netdev;
+        IPVlan *ipvlan;
```
bluca, line 84: `assert(link)`

bluca (approved):
The `(cherry picked from commit` lines should probably be dropped from the commit messages? Otherwise it will be confusing after merge

yuwata:
@bluca Thank you for the comments. Addressed the above points. Upgrading the green label.

PR #22624, opened 2022-02-25
file-hierarchy: Document the canonical mountpoint for cgroups

joshtriplett:
file-hierarchy does not mention anything about the expected mountpoint
for cgroups. This may lead some software to believe it will need to
search for it (e.g. by scanning `mountinfo`) rather than just looking in
the canonical location.

Document the canonical mountpoint as `/sys/fs/cgroup`. Also provide enough
information for software looking for `cgroup2` to know that it may want to
look under `/sys/fs/cgroup/unified` as a fallback.

### DIFF
--- a/man/file-hierarchy.xml
+++ b/man/file-hierarchy.xml
@@ -420,6 +420,25 @@
         this directory.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><filename>/sys/fs/cgroup/</filename></term>
+        <listitem><para>A virtual kernel file system exposing process
+        control groups (cgroups). This file system is an API to interface
+        with the kernel and not a place where normal files may be stored. On
+        current systems running in the default "unified" mode,
+        this directory serves as the mount point for the
+        <literal>cgroup2</literal> filesystem, which provides a unified
+        cgroup hierarchy for all resource controllers. On systems with
+        non-default configurations, this directory may instead be a tmpfs
+        filesystem containing mount points for various
+        <literal>cgroup</literal> (v1) resource controllers; in such
+        configurations, if <literal>cgroup2</literal> is mounted it will be
+        mounted on <filename>/sys/fs/cgroup/unified/</filename>, but
+        cgroup2 will not have resource controllers attached. In
+        sandboxed/containerized setups, this directory may either not exist or
+        may include a subset of functionality.
+        </para></listitem>
+      </varlistentry>
     </variablelist>
   </refsect1>
 


keszybz:
Seems reasonable to document this. We certainly want to describe a setup with cgroupv2. I'm not sure if we want to describe hybrid and v1 too though. I'll let others comment.

joshtriplett:
@keszybz Seemed painless to allow for the hybrid setup, insofar as it tells code looking for v2 to consider a fallback to `/sys/fs/cgroup/unified`. Happy to adapt this to whatever degree systemd would prefer to document this, though.

man/file-hierarchy.xml
```diff
@@ -420,6 +420,22 @@
...
+        <listitem><para>A virtual kernel file system exposing process
+        control groups (cgroups). This file system is an API to
+        interface with the kernel and not a place where normal files may
+        be stored. In sandboxed/containerized setups, this directory may
+        either not exist or may include a subset of functionality. On
+        current systems, this directory will refer to the
+        <literal>cgroup2</literal> filesystem, which provides a unified
+        cgroup hierarchy for all resource controllers. On older systems,
+        this directory may refer to the <literal>cgroup</literal> (v1)
+        filesystem, and will contain a subdirectory
```
bluca: s/will/might/ - systems can be booted in legacy-only mode too

joshtriplett: Was not aware that was still a supported configuration. :(

My main goal is to make it clear that if cgroup2 is mounted, it's mounted in one of these two places. I can tweak this to allow for cgroup2 not being mounted at all.

keszybz: We still document and support all three modes: legacy, hybrid, and unified, see e.g. dicussion of `systemd.legacy_systemd_cgroup_controller` in systemd(1), and https://systemd.io/CGROUP_DELEGATION/#three-different-tree-setups-. And those three modes are confusing for people, so I think we should be careful here do describe the three possibilities correctly.


"will refer to …" → "will contain …", or "… will be mounted on", or maybe "uses …".

Please move the sentence about containers after the description of what happen on "real" systems.

> "On older systems,  this directory may refer to the <literal>cgroup</literal> (v1)"

This is incorrect. It will use tmpfs, with subdirs with cgroup and cgroup2.

joshtriplett:
I've updated the PR to more carefully cover hybrid/legacy setups. I've also made it clear that if cgroup2 is mounted on `/sys/fs/cgroup/unified/` it won't have any resource controllers attached, so software looking for cgroup2 resource controllers needn't bother falling back to that directory.

keszybz:
LGTM. I guess we could add further tweaks, like more hyperlinks, but we can do that separately.

PR #22627, opened 2022-02-26
network: cleanups for L2TP tunnel

yuwata:
Split out of #22557.

### DIFF
--- a/src/network/netdev/netdev.c
+++ b/src/network/netdev/netdev.c
@@ -144,7 +144,7 @@ bool netdev_is_managed(NetDev *netdev) {
 static bool netdev_is_stacked_and_independent(NetDev *netdev) {
         assert(netdev);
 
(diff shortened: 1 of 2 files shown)


src/network/netdev/l2tp-tunnel.c
```diff
@@ -245,44 +246,114 @@ static int netdev_l2tp_create_message_session(NetDev *netdev, L2tpSession *sessi
...
+                        return r;
+        }
 
         if (in_addr_is_set(t->family, &t->local)) {
                 /* local address is explicitly specified. */
-                *ret = t->local;
+
+                if (!a) {
+                        if (link)
+                                r = link_get_address(link, t->family, &t->local, 0, &a);
```
bluca, line 314: This could return 0 but `a` could still be NULL - then `address_is_ready` would assert

bluca: ah no it can't, I checked link_address_is_reachable instead of link_get_address

bluca (approved):

Issue #20649, opened 2021-09-06
systemd-oomd: Provide user feedback when killing a cgroup in a user slice


daandemeyer:
**Is your feature request related to a problem? Please describe.**

Currently, systemd-oomd kills user programs without providing any further feedback to the user. This often leaves users confused as to why their program was suddenly terminated.

**Describe the solution you'd like**

Ideally, users would receive a clear notification that systemd-oomd has killed (or is about to kill) one of their programs.

**Describe alternatives you've considered**

/

**The systemd version you checked that didn't have the feature you are asking for**

v249

After some discussion with Lennart, we figured this could work by having systemd-oomd provide a D-Bus interface that allows clients to subscribe for oomd notifications. When oomd kills a program, it would then notify all subscribers via a D-Bus signal. Gnome/KDE could then subscribe to these notifications on the system bus and proxy them to the org.freedesktop.Notifications interface on the user bus.

cc @benzea

benzea:
Stupid question. But couldn't systemd itself rather than systemd-oomd export the information that the unit got into the failed state due to OOM?

DaanDeMeyer:
> Stupid question. But couldn't systemd itself rather than systemd-oomd export the information that the unit got into the failed state due to OOM?

I'm pretty sure it could, but I figure oomd has more information available on why it's killing things that might be interesting to include in the D-Bus signals that systemd itself won't have. Either way, you'd need to proxy notifications the system bus (coming from either systemd or systemd-oomd) to the user bus.

poettering:
I am not sure we should replicate the full oom meta info in pid 1, i.e. i think it's semantically nice this is separated. Yes, PID 1 should receive state changes from oomd, and track that, but detailed reports on what happened should not be proxied/replicated in PID1. I mean, doing such a game of proxying things along oh so many paths is something one should think twice about, and for every single step have a really good reason. But I am not convinced that a client preferring to talk to PID1 rather than oomd is a really good reason. Whether you talk to dbus service foo or dbus service bar really shouldn't matter, should it?

poettering:
BTW, I'd like to extend the discussion to coredumps. Does GNOME currently have a story there? i.e. what happens if an app coredumps? is this handled by GNOME in any way? Or does the app just silently die? What about crashes in system services?

[...]

PR #22629, opened 2022-02-27
core/oomd: Use oom-kill ServiceResult for oomd

nishalkulkarni:
To notify user of kill events from systemd-oomd we now use
`SERVICE_FAILURE_OOM_KILL` as the failure result.

`unit_check_oomd_kill` now calls `notify_cgroup_oom` to
update the service result to `oom-kill`.

We add a new xattr `user.oomd_ooms` to keep track of the OOM kills
initiated by systemd-oomd, this helps us resolve a race between sending
SIGKILL to processes and checking for OOM kill status from the xattr.

Related to: #20649

### DIFF
--- a/man/systemd-system.conf.xml
+++ b/man/systemd-system.conf.xml
@@ -505,7 +505,7 @@
         <term><varname>DefaultOOMPolicy=</varname></term>
 
         <listitem><para>Configure the default policy for reacting to processes being killed by the Linux
-        Out-Of-Memory (OOM) killer. This may be used to pick a global default for the per-unit
+        Out-Of-Memory (OOM) killer or <command>systemd-oomd</command>. This may be used to pick a global default for the per-unit
         <varname>OOMPolicy=</varname> setting. See
         <citerefentry><refentrytitle>systemd.service</refentrytitle><manvolnum>5</manvolnum></citerefentry>
         for details. Note that this default is not used for services that have <varname>Delegate=</varname>
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -1039,7 +1039,7 @@ CapabilityBoundingSet=~CAP_B CAP_C</programlisting>
         normally at 0.</para>
 
         <para>Use the <varname>OOMPolicy=</varname> setting of service units to configure how the service
-        manager shall react to the kernel OOM killer terminating a process of the service.  See
+        manager shall react to the kernel OOM killer or <command>systemd-oomd</command> terminating a process of the service.  See
         <citerefentry><refentrytitle>systemd.service</refentrytitle><manvolnum>5</manvolnum></citerefentry>
         for details.</para></listitem>
       </varlistentry>
--- a/man/systemd.service.xml
+++ b/man/systemd.service.xml
@@ -1140,7 +1140,11 @@
         shall be considered preferred or less preferred candidates for process termination by the Linux OOM
         killer logic. See
         <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry> for
-        details.</para></listitem>
+        details.</para>
+
+        <para>This setting also applies to <command>systemd-oomd</command>, similar to kernel OOM kills
+        this setting determines the state of the service after <command>systemd-oomd</command> kills a cgroup associated
+        with the service.</para></listitem>
       </varlistentry>
 
     </variablelist>


benzea:
Cool, this really looks pretty simple! I think in general we can also just enable it together with `ExitType=croup`, I don't see why that should be an issue.

Note that we *do* have the race condition currently that @Werkov was pointing out already. I think we can resolve this by changing the OOMD code though. i.e. looking at `oomd_cgroup_kill` we just need to move the `increment_oomd_xattr` to happen before the `cg_kill`/`cg_kill_recursive` call.

So, I think we can just add a patch to make that change, and then it should work. And, then maybe also enable the exit code for `ExitType=cgroup`.

CC: @anitazha

benzea:
Hrm, except that we cannot export the number of processes that were killed into `user.oomd_kill` then.

So, we either need to change the definition of that attribute/counter, or we need another attribute to resolve the race condition.

anitazha:
> Hrm, except that we cannot export the number of processes that were killed into `user.oomd_kill` then.
> 
> So, we either need to change the definition of that attribute/counter, or we need another attribute to resolve the race condition.

The original oomd had 2 attributes: oomd_ooms which is incremented before SIGKILL and oomd_kill (the one systemd-oomd also has) which is incremented after. I didn't end up using oomd_ooms in the first iteration of systemd-oomd so I excluded it, but we can add it back to help mitigate a race in this case.

anitazha (commented):
Thanks for working on this the changes look great. If you have cycles to also add the additional xattr increment for oomd_ooms that would also be appreciated :)

@benzea do you think this is enough info to close #20649 ?

src/core/service.c
```diff
@@ -3436,9 +3436,18 @@ static void service_notify_cgroup_empty_event(Unit *u) {
...
 }
 
-static void service_notify_cgroup_oom_event(Unit *u) {
+static void service_notify_cgroup_oom_event(Unit *u, bool managed_oom) {
         Service *s = SERVICE(u);
 
+        if (managed_oom) {
+                log_unit_debug(u, "Process(es) of control group were killed by systemd-oomd.");
+
+                if (s->exit_type == SERVICE_EXIT_MAIN || cgroup_good(s) <= 0)
```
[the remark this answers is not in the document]
nishalkulkarni: Got it from core/service.c#L3579, I misunderstood the comment. 
I'll put up a patch fixing this and adding additional xattr. Thanks!

benzea:
> @benzea do you think this is enough info to close https://github.com/systemd/systemd/issues/20649 ?

In my view, this is enough from the systemd side (We do need `ExitType=cgroup` for KDE at least). If we have this, we can simply write a desktop service that listens for all property changes on units (matching for `arg0` to be `org.freedesktop.systemd1.Unit` and listening for all `org.freedesktop.DBus.Properties.PropertiesChanged` signals otherwise).

src/core/cgroup.c
```diff
@@ -3054,12 +3057,16 @@ int unit_check_oomd_kill(Unit *u) {
         if (!increased)
                 return 0;
 
-        if (n > 0)
+        if (n > 0) {
```
Werkov: I know this isn't a new change but can this condition be ever false provided this is in branch where `increased` is true?

src/core/service.h
```diff
@@ -75,7 +75,8 @@ typedef enum ServiceResult {
         SERVICE_FAILURE_CORE_DUMP,
         SERVICE_FAILURE_WATCHDOG,
         SERVICE_FAILURE_START_LIMIT_HIT,
-        SERVICE_FAILURE_OOM_KILL,
+        SERVICE_FAILURE_OOM_KILL, /* OOM Kill by the Kernel */
+        SERVICE_FAILURE_OOMD_KILL, /* OOM Kill by systemd-oomd */
```
poettering: hmm, oomd is a specific piece of software, not a condition or cause. I'd name this SERVICE_FAILURE_MANAGED_OOM_KILL, i.e. emphasize that this is a *managed* kill, regardless of the fact this is implemented in oomd.

I wonder though: why distinguish this from regular OOM kills? In which way is that relevant to UIs? isn't this just an implementation detail we should abstract away?

or in other words, is this distinction something we should expose in API or just in logs?

src/core/cgroup.c
```diff
@@ -3054,12 +3057,16 @@ int unit_check_oomd_kill(Unit *u) {
...
                 log_unit_struct(u, LOG_NOTICE,
                                 "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
                                 LOG_UNIT_INVOCATION_ID(u),
                                 LOG_UNIT_MESSAGE(u, "systemd-oomd killed %"PRIu64" process(es) in this unit.", n));
 
+                if (UNIT_VTABLE(u)->notify_cgroup_oom)
+                        UNIT_VTABLE(u)->notify_cgroup_oom(u, true);
```
poettering: in new code when we call a function that takes a boolean flag style argument we prefer if a comment is added giving a hint what this parameter is about, i.e.:

```c
UNIT_VTABLE(u)->notify_cgroup_oom(u, /* managed= */ true);
```

src/core/cgroup.c
```diff
@@ -3095,7 +3102,7 @@ int unit_check_oom(Unit *u) {
                         LOG_UNIT_MESSAGE(u, "A process of this unit has been killed by the OOM killer."));
 
         if (UNIT_VTABLE(u)->notify_cgroup_oom)
-                UNIT_VTABLE(u)->notify_cgroup_oom(u);
+                UNIT_VTABLE(u)->notify_cgroup_oom(u, false);
```
poettering: given that these two lines now appear at multiple places, maybe wrap them in a trivial helper. i.e. similar to unit_following() or unit_will_restart() or unit_kill(), which call the vtable method, but become NOPs if the method is NULL.

src/core/cgroup.c
```diff
@@ -2959,6 +2959,9 @@ static int on_cgroup_empty_event(sd_event_source *s, void *userdata) {
                         log_debug_errno(r, "Failed to reenable cgroup empty event source, ignoring: %m");
         }
 
+        /* Update state based on systemd-oomd kills before we notify about cgroup empty event */
+        (void) unit_check_oomd_kill(u);
```
poettering, line 2963: this should probably be a separate commit, no? and we should probably do unit_check_oom() here too, just in case.

poettering: (i.e. as in manager_dispatch_sigchld())

src/core/cgroup.c
poettering: yeah, this looks like a pre-existing bug (or well, ugliness)

poettering:
looks pretty good. Though I am not sure it's a good idea to distinguish the two kinds of oom kills in the API. It appears to me that that's an implementaiton detail we should log about but not expose in the API otherwise, it doesn#t really matter if the kernel did this or us I'd argue.

src/core/service.h
nishalkulkarni: A separate failure state might be useful for troubleshooting, as mentioned by @Werkov in https://github.com/systemd/systemd/issues/20649#issuecomment-965177073

poettering: not sure I am convinced. I mean, making troubleshooting easy is a worthy goal, but to me the answer for that is good logging, not propagating the implementation detail here...

benzea: I am fine with just using `oom-kill` for both. I can't imagine wanting to tell the two types of OOM kills apart. In particular, because a non-managed kill would only happen in odd situation where there is no managed killer or where it is reacting too slowly/incorrectly for some reason.

i.e. you can assume that it was a managed OOM kill if the system has a managed OOM killer running.

nishalkulkarni: Using `oom-kill` for both kernel OOM and systemd-oomd, looks cleaner now. thanks!

src/core/cgroup.c
nishalkulkarni: Added `unit_notify_cgroup_oom()` wrapper

src/core/service.c
```diff
@@ -3436,9 +3436,18 @@ static void service_notify_cgroup_empty_event(Unit *u) {
...
+                        s->result = SERVICE_FAILURE_OOM_KILL;
+
+                return;
```
poettering: hmm, why shortcut the fnction here? I see no reason why we shouldn't continue here like in the kernel oom kill case, i.e. switch to the stop phase, as the switch statement below does. hence, instead of the return here, just continue with the rest of the function, and let it handle setting of SERVICE_FAILURE_OOM_KILL?

nishalkulkarni: I thought the OOMPolicy(continue|stop|kill) was specific to the kernel OOMs and didn't want to mess with it.

So I guess they'll apply to systemd-oomd kills too?

benzea: Hmm, it shouldn't be applicable really. The kernel will only kill one process at a time, so a service can survive an OOM event. In contrast, systemd-oomd will kill the whole cgroup, which is makes it equivalent to a `OOMPolicy=kill` in most cases.

The exception is if we have `Delegate=yes`, I suppose in that case we will currently *not* see the kill as an OOM kill with this patchset. Do we need something smarter that looks into the leaf nodes if when delegation is on?

Werkov: Ooh. Shloudn't there be *someone else* responsible for checking in delegated subtrees, e.g. it'd be the user instance that manages (user) units in the subtree. (Although, I'm not familiar with the general model of systemd-oomd delegation. Is it delegatable at all?)

benzea: I doubt it is delegatable really. Right now it pretty much just looks at all leaf nodes and selects the "best" candidate.

That said, systemd-oomd is part of systemd. It could have some logic to (also) set an attribute on the units cgroup rather than only in the child cgroup where the kill happened. Could we search up for the next cgroup with `.service/.scope` in the name and increment `user.oomd_unit_kill`?

anitazha: `OOMPolicy=` shouldn't be applicable to systemd-oomd kills in general as benzea mentioned.

Thinking a bit about delegated subtrees though, we do partially follow what werkov is suggesting: right now the user manager subtrees are delegated and when a systemd-oomd kill happens, the user manager follows the same code paths as what is being implemented in this PR.

Anyways that's just informational. I do like benzea's idea to increment an xattr on the `service / scope` cgroups. Instead of a new xattr though what if we only increment `user.oomd_ooms` on those `service / scope` cgroups? Otherwise it would go unused on the leaf node (unless the leaf is a `service / scope` cgroup). And then in the delegated case if the `service / scope` cgroup is not empty we treat it similar to `OOMPolicy=continue` (that is, the rest of the unit keeps going).

Side note, FB's container workloads are delegated systemd units and have memory.oom.group=1. That way systemd-oomd kills the whole subtree / workload along with any subcgroups created by the workload.

anitazha (commented):
For testing please update src/oom/test-oomd-util.c accordingly to test that the xattrs are getting incremented.

There is also an extended test TEST-55-OOMD and units (e.g. testsuite-55-testbloat.service) that tests systemd-oomd kills are happening the way we expect. You can use it as an example to set `MemoryHigh=` on a unit and make pressure build up quickly so that you see the messages from pid1.

anitazha:
> Maybe another idea could be to just only consider OOM kills on the leaf-node that contains the main service PID? That would basically delegate other leaf-nodes to the service, while we keep managing the main "service" cgroup from systemd.

I'm not sure I understand why we only focus on the main service PID. Wouldn't we want to go through this flow if a control command triggered the OOM kill as well?

src/core/service.c
anitazha: I just realized I wrote too much without really giving direction. Looking at what I wrote yesterday and the `service_notify_cgroup_oom_event()` state machine again I have a different opinion. Because `OOMPolicy=kill` sets `memory.oom.group=1` and systemd-oomd respects `memory.oom.group`, I don't see why we shouldn't follow the rest of the case/switch as Lennart mentioned. But we need to update the documentation to say this because right now systemd-oomd kills behave like `OOMPolicy=continue` (I don't think that was documented either but that's on me).

Like benzea mentioned below we have the same problem with kernel OOM kills and memory.events as it doesn't look down the whole subtree. We can make this better for systemd-oomd if we wanted to later on but that's out of the scope of this PR.

src/oom/test-oomd-util.c
```diff
@@ -77,6 +77,9 @@ static void test_oomd_cgroup_kill(void) {
                         abort();
                 }
 
+                assert_se(cg_get_xattr_malloc(SYSTEMD_CGROUP_CONTROLLER, cgroup, "user.oomd_ooms", &v) >= 0);
+                assert_se(memcmp(v, i == 0 ? "1" : "2", 2) == 0);
+
```
mrc0mmand, line 83: this also needs

```c
v = mfree(v);
```

to avoid leaking memory (since the pointer is reassigned in the `cg_get_xattr_malloc()` call below).

src/core/cgroup.c
```diff
@@ -3058,11 +3058,12 @@ int unit_check_oomd_kill(Unit *u) {
...
 
-        if (n > 0)
-                log_unit_struct(u, LOG_NOTICE,
-                                "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
-                                LOG_UNIT_INVOCATION_ID(u),
-                                LOG_UNIT_MESSAGE(u, "systemd-oomd killed %"PRIu64" process(es) in this unit.", n));
+        log_unit_struct(u, LOG_NOTICE,
+                        "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
+                        LOG_UNIT_INVOCATION_ID(u),
+                        LOG_UNIT_MESSAGE(u, "systemd-oomd killed some process(es) in this unit."));
```
anitazha: Could you add in the value from `user.oomd_kill` if it is not zero? For example use "some" if it is zero and replace with the number when it is not.

We don't need to store it in `managed_oom_kill_last` since we store `user.oomd_ooms` now, but logging the number of processes can be helpful for debugging

man/systemd.service.xml
```diff
@@ -1140,7 +1140,11 @@
         shall be considered preferred or less preferred candidates for process termination by the Linux OOM
         killer logic. See
         <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry> for
-        details.</para></listitem>
+        details.</para>
+
+        <para>This setting also applies to <constant>systemd-oomd</constant>, similar to kernel OOM kills
```
anitazha: `constant` -> `command`

man/systemd-system.conf.xml
```diff
@@ -503,7 +503,7 @@
         <term><varname>DefaultOOMPolicy=</varname></term>
 
         <listitem><para>Configure the default policy for reacting to processes being killed by the Linux
-        Out-Of-Memory (OOM) killer. This may be used to pick a global default for the per-unit
+        Out-Of-Memory (OOM) killer or systemd-oomd. This may be used to pick a global default for the per-unit
```
anitazha: please use `<command>` tags around systemd-oomd

man/systemd.exec.xml
```diff
@@ -1040,7 +1040,7 @@ CapabilityBoundingSet=~CAP_B CAP_C</programlisting>
         normally at 0.</para>
 
         <para>Use the <varname>OOMPolicy=</varname> setting of service units to configure how the service
-        manager shall react to the kernel OOM killer terminating a process of the service.  See
+        manager shall react to the kernel OOM killer or systemd-oomd terminating a process of the service.  See
```
anitazha: please use `<command>` tags around systemd-oomd

man/systemd.service.xml
```diff
@@ -1140,7 +1140,11 @@
...
+        this setting determines the state of the service after systemd-oomd kills a cgroup associated
```
anitazha: please use `<command>` tags around systemd-oomd

anitazha (commented):
Thanks! A few small things but this looks great

src/core/cgroup.c
nishalkulkarni: added, although didn't we use oomd_ooms to help with the race condition where user.oomd_kill xattr might not have been updated by the time we reach here?

src/core/cgroup.c
```diff
@@ -3058,11 +3058,28 @@ int unit_check_oomd_kill(Unit *u) {
...
                 return 0;
 
-        if (n > 0)
+        n = 0;
+        value = mfree(value);
+        r = cg_get_xattr_malloc(SYSTEMD_CGROUP_CONTROLLER, u->cgroup_path, "user.oomd_kill", &value);
+        if (r >= 0 && !isempty(value)) {
+                r = safe_atou64(value, &n);
+                if (r < 0)
+                        n = 0;
```
anitazha: you can drop the `r < 0` check and do
```
(void) safe_atou64(value, &n);
```

the convention in the systemd code base (including this function) is to only set `ret` values passed in on success.

nishalkulkarni:
Rebased on main, GTM?

src/core/cgroup.c
```diff
@@ -3057,11 +3057,25 @@ int unit_check_oomd_kill(Unit *u) {
...
                 log_unit_struct(u, LOG_NOTICE,
                                 "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
                                 LOG_UNIT_INVOCATION_ID(u),
                                 LOG_UNIT_MESSAGE(u, "systemd-oomd killed %"PRIu64" process(es) in this unit.", n));
+        } else {
+                log_unit_struct(u, LOG_NOTICE,
+                                "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
+                                LOG_UNIT_INVOCATION_ID(u),
+                                LOG_UNIT_MESSAGE(u, "systemd-oomd killed some process(es) in this unit."));
+        }
```
poettering: as per coding style we don't place {} around single-line if blocks

src/core/cgroup.c
```diff
@@ -3057,11 +3057,25 @@ int unit_check_oomd_kill(Unit *u) {
...
+        value = mfree(value);
+        r = cg_get_xattr_malloc(SYSTEMD_CGROUP_CONTROLLER, u->cgroup_path, "user.oomd_kill", &value);
+        if (r >= 0 && !isempty(value))
+                (void) safe_atou64(value, &n);
+
+        if (n > 0) {
                 log_unit_struct(u, LOG_NOTICE,
                                 "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
                                 LOG_UNIT_INVOCATION_ID(u),
                                 LOG_UNIT_MESSAGE(u, "systemd-oomd killed %"PRIu64" process(es) in this unit.", n));
```
poettering: this is a structured log message, hence it would be neat to add the number of killed processes also in a structed field. i.e. add one more line to the expression:

```c

                log_unit_struct(u, LOG_NOTICE,
                                "MESSAGE_ID=" SD_MESSAGE_UNIT_OOMD_KILL_STR,
                                LOG_UNIT_INVOCATION_ID(u),
                                LOG_UNIT_MESSAGE(u, "systemd-oomd killed %"PRIu64" process(es) in this unit.", n),
                                "N_PROCESSES=%" PRIu64, n);
```

src/oom/test-oomd-util.c
```diff
@@ -77,6 +77,10 @@ static void test_oomd_cgroup_kill(void) {
                         abort();
                 }
 
+                assert_se(cg_get_xattr_malloc(SYSTEMD_CGROUP_CONTROLLER, cgroup, "user.oomd_ooms", &v) >= 0);
+                assert_se(memcmp(v, i == 0 ? "1" : "2", 2) == 0);
```
poettering: i'd just use streq() here... no need for memcmp(), i.e.

```c
asset_se(streq(v, i == 0 ? "1" : "2"));
```

poettering (changes requested):
looks excellent, just some minor stuff

poettering:
Perfect! thanks!

src/core/cgroup.c
```diff
@@ -3041,7 +3041,7 @@ int unit_check_oomd_kill(Unit *u) {
         else if (r == 0)
                 return 0;
 
-        r = cg_get_xattr_malloc(SYSTEMD_CGROUP_CONTROLLER, u->cgroup_path, "user.oomd_kill", &value);
+        r = cg_get_xattr_malloc(SYSTEMD_CGROUP_CONTROLLER, u->cgroup_path, "user.oomd_ooms", &value);
```
keszybz, line 3044: Hmm, so is this something that is a public API? Is the change in the name something that should be mentioned in NEWS and treated as a backwards-incompatible change? @anitazha

anitazha: > Hmm, so is this something that is a public API?

The oomd xattrs weren't intended to be public API and isn't documented. So I don't think this is backward incompatible change unless for some reason pid1 and systemd-oomd are on different versions.

man/systemd.service.xml
```diff
@@ -1140,7 +1140,11 @@
         shall be considered preferred or less preferred candidates for process termination by the Linux OOM
         killer logic. See
         <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry> for
-        details.</para></listitem>
+        details.</para>
+
+        <para>This setting also applies to <command>systemd-oomd</command>, similar to kernel OOM kills
+        this setting determines the state of the service after <command>systemd-oomd</command> kills a cgroup associated
+        with the service.</para></listitem>
```
slyon, line 1147: @nishalkulkarni is the "OOMScoreAdjust=" setting recognized by sd-oomd as of this PR (I cannot find anything in code that would relate to this), or has this been the case for previous releases as well, and you just updated the documentation/man page?

slyon: Never mind. This is only about the `OOMPolicy=` setting (which is part of the PR), I was confused.

nishalkulkarni: ah yeah, sorry, its positioning might confuse people... don't think it's worth a new pull request though

PR #22630, opened 2022-02-27
TPM2 parameter encryption

grigorig:
Note: this PR builds upon unmerged PR #22563.

This implements parameter encryption for sealing and unsealing secrets to a TPM. A separate salted and unbound HMAC session is used for that. The session key derivation process seems complicated, but in the end it boils down to encrypting a randomly chosen salt with a public key (in this case, the primary key). Only the TPM knows the secret part, and can decode the salt to derive the key correctly.

As far as security guarantees go, this provides confidentiality, integrity and replay protection. So common attacks based on eavesdropping or manipulating the TPM interface will fail. It does not however protect against man in the middle attacks. We would need to authenticate the TPM for that, possibly with its endorsement certificate (if available).

tpm-tss2 libraries provide a [handy mechanism to dump data](https://github.com/tpm2-software/tpm2-tss/pull/1817) exchanged with TPMs into a `pcap` file, which then can be analyzed with Wireshark. That's what I did for testing. It would still be great if an actual security researcher could check whether the implementation has any obvious loopholes or weaknesses.

### DIFF
--- a/src/shared/tpm2-util.c
+++ b/src/shared/tpm2-util.c
@@ -36,6 +36,7 @@ TSS2_RC (*sym_Esys_PolicyGetDigest)(ESYS_CONTEXT *esysContext, ESYS_TR policySes
 TSS2_RC (*sym_Esys_PolicyPCR)(ESYS_CONTEXT *esysContext, ESYS_TR policySession, ESYS_TR shandle1, ESYS_TR shandle2, ESYS_TR shandle3, const TPM2B_DIGEST *pcrDigest, const TPML_PCR_SELECTION *pcrs) = NULL;
 TSS2_RC (*sym_Esys_StartAuthSession)(ESYS_CONTEXT *esysContext, ESYS_TR tpmKey, ESYS_TR bind, ESYS_TR shandle1, ESYS_TR shandle2, ESYS_TR shandle3, const TPM2B_NONCE *nonceCaller, TPM2_SE sessionType, const TPMT_SYM_DEF *symmetric, TPMI_ALG_HASH authHash, ESYS_TR *sessionHandle) = NULL;
 TSS2_RC (*sym_Esys_Startup)(ESYS_CONTEXT *esysContext, TPM2_SU startupType) = NULL;
+TSS2_RC (*sym_Esys_TRSess_SetAttributes)(ESYS_CONTEXT *esysContext, ESYS_TR session, TPMA_SESSION flags, TPMA_SESSION mask);
 TSS2_RC (*sym_Esys_TR_SetAuth)(ESYS_CONTEXT *esysContext, ESYS_TR handle, TPM2B_AUTH const *authValue) = NULL;
 TSS2_RC (*sym_Esys_Unseal)(ESYS_CONTEXT *esysContext, ESYS_TR itemHandle, ESYS_TR shandle1, ESYS_TR shandle2, ESYS_TR shandle3, TPM2B_SENSITIVE_DATA **outData) = NULL;
 
@@ -66,6 +67,7 @@ int dlopen_tpm2(void) {
                         DLSYM_ARG(Esys_PolicyPCR),
                         DLSYM_ARG(Esys_StartAuthSession),
                         DLSYM_ARG(Esys_Startup),
+                        DLSYM_ARG(Esys_TRSess_SetAttributes),
                         DLSYM_ARG(Esys_TR_SetAuth),
                         DLSYM_ARG(Esys_Unseal));
         if (r < 0)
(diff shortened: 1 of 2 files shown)


grigorig:
I assume the focal-amd64 test fails because of bugs in an old tpm2-tss library version. Is there a way to get detailed journal logs from that test? The autopkgtest site seems to be unresponsive (pages never finish loading).

mrc0mmand:
> I assume the focal-amd64 test fails because of bugs in an old tpm2-tss library version. Is there a way to get detailed journal logs from that test? The autopkgtest site seems to be unresponsive (pages never finish loading).

You should be able to download https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/amd64/s/systemd-upstream/20220227_140531_93b90@/artifacts.tar.gz which should have the test journal in the `artifacts` subfolder.

grigorig:
Now THAT is annoying: [this issue](https://github.com/tpm2-software/tpm2-tss/issues/1590) seems to be culprit. It was fixed in version 2.3.3. And Ubuntu Focal ships... version 2.3.2. :)

Since old Ubuntu versions are not going to ship updated systemd, maybe we could make version 2.3.3 (or 2.4.0) the minimum?

bluca:
yeah we can push it to 2.3.3, I'll upload to the PPA

grigorig:
Great, that's even better, at least for testing purposes. Note that I didn't actually check if version 2.3.3 fully fixes the problem, but if it isn't too much work please update the library.

bluca:
uploaded 2.3.3 and re-triggered

grigorig:
> uploaded 2.3.3 and re-triggered

That seems to have worked, after the network issues (?) were fixed. Thanks for the update.

src/shared/tpm2-util.c
```diff
@@ -596,8 +598,65 @@ static int tpm2_get_best_pcr_bank(
...
+                        TPM2_SE_HMAC,
+                        &symmetric,
+                        TPM2_ALG_SHA256,
+                        &session);
+        if (rc != TSS2_RC_SUCCESS)
+                return log_error_errno(SYNTHETIC_ERRNO(ENOTRECOVERABLE),
+                                       "Failed to open session in TPM: %s", sym_Tss2_RC_Decode(rc));
+
+        TPMA_SESSION sessionAttributes =
+                        (TPMA_SESSION_DECRYPT | TPMA_SESSION_ENCRYPT | TPMA_SESSION_CONTINUESESSION);
```
poettering: () seem unnecessary

src/shared/tpm2-util.c
```diff
@@ -596,8 +598,65 @@ static int tpm2_get_best_pcr_bank(
...
+        rc = sym_Esys_TRSess_SetAttributes(c, session, sessionAttributes, 0xff);
+        if (rc != TSS2_RC_SUCCESS)
+                return log_error_errno(
+                                SYNTHETIC_ERRNO(ENOTRECOVERABLE),
+                                "Failed to configure TPM session: %s",
+                                sym_Tss2_RC_Decode(rc));
+
+        if (ret_session) {
+                *ret_session = session;
+                session = ESYS_TR_NONE;
```
poettering, line 654: doesn't matter, but we might want to add TAKE_ESYS_TR or so one day as a macro:

```c
#define TAKE_ESYS_TR(x) ({ \
        ESYS_TR *_ptr = &(x); \
        ESYS_TR _tr = *_ptr; \
        *_ptr = ESYS_TR_NONE; \
        _tr; \
})
```

That way we can write this here as:

```c
*ret_session = TAKE_ESYS_TR(session);
```

This alls taking inspiration from TAKE_PTR(), TAKE_FD(), TAKE_PID() macros.

src/shared/tpm2-util.c
```diff
@@ -596,8 +598,65 @@ static int tpm2_get_best_pcr_bank(
...
+                .mode = {
+                        .aes = TPM2_ALG_CFB,
+                }
+        };
+        ESYS_TR session = ESYS_TR_NONE;
+        TSS2_RC rc;
+
+        assert(c);
+
+        log_debug("Starting HMAC session.");
```
poettering: i think a comment here what this does and what it is for would be more than welcome.

src/shared/tpm2-util.c
```diff
@@ -596,8 +598,65 @@ static int tpm2_get_best_pcr_bank(
...
+                ESYS_TR *ret_session) {
+
+        static const TPMT_SYM_DEF symmetric = {
+                .algorithm = TPM2_ALG_AES,
+                .keyBits = {
+                        .aes = 128
+                },
+                .mode = {
+                        .aes = TPM2_ALG_CFB,
+                }
```
poettering: doesn't matter at all, but we usually place a trailing `,` into all struct inits, just to minimize later patches that fill in more stuff.

src/shared/tpm2-util.c
```diff
@@ -596,8 +598,65 @@ static int tpm2_get_best_pcr_bank(
         return 0;
 }
 
+static int tpm2_make_hmac_session(
```
poettering: i wonder if this should be called tpm2_setup_parameter_encryption() or so? i.e. focus on the purpose, not on the implementation detail?

src/shared/tpm2-util.c
```diff
@@ -872,7 +935,7 @@ int tpm2_seal(
         rc = sym_Esys_Create(
                         c.esys_context,
                         primary,
-                        ESYS_TR_PASSWORD,
+                        session,
```
poettering: maybe add an inline comment clarifying why this parameter is passed, i.e. suffix by something short like `/* pass HMAC sesssion to enable parameter encryption */` or so

poettering: (and similar at the other calls where this is done)

src/shared/tpm2-util.c
poettering: (i.e. copy tthe stuff from the commit msg here)

src/shared/tpm2-util.c
grigorig: But looks nicer IMHO. I'll remove it though.

src/shared/tpm2-util.c
grigorig: I'll keep it in mind for a possible follow-up cleanup.

src/shared/tpm2-util.c
grigorig: We already fill in the struct completely, though.

poettering: the reader won't know that though.

but doesn't matter at all. (you could earn bonus points though if you also add the `,` to the `keyBits` struct init a few lines above)

bluca:
(removing green label for now to avoid accidental merges, will re add it later)

poettering:
#22563 has been merged now. Please rebase. Should be ready for commit then too.

src/shared/tpm2-util.c
grigorig: I added commas for consistency after all.

bluca (approved):

PR #22632, opened 2022-02-28
man: recommend built-in platform.freedesktop_os_release() in our page

keszybz:
Python gained support for reading os-release, let's advertise it a bit more.
Our open-coded example is still useful, but let's not suggest it as the
default implementation.

I added quotes around the printed string because it looks a bit better
this way.

### DIFF
--- a/man/check-os-release-simple.py
+++ b/man/check-os-release-simple.py
@@ -1,0 +1,12 @@
+#!/usr/bin/python
+# SPDX-License-Identifier: CC0-1.0
+
+import platform
+os_release = platform.freedesktop_os_release()
+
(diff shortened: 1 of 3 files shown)


man/check-os-release-simple.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/python
```
bluca, line 1: env python ?

bluca (approved):

man/check-os-release-simple.py
keszybz: It's visible in the man page, so we don't use `env`. It's the same in the other example.

PR #22634, opened 2022-02-28
mkosi: Add CentOS Stream 9

daandemeyer:
The blocker causing Stream 9 builds to fail was fixed
(https://bugzilla.redhat.com/show_bug.cgi?id=2056276) so we can add
CentOS Stream 9 builds as well now.


DaanDeMeyer:
Hmm, seems the change hasn't been fully deployed yet so the job will likely fail. Will keep this open until the change is fully deployed and then rerun CI.

yuwata (approved):
LGTM.

PR #19637, opened 2021-05-17
core: device: re-trigger uevents missed by udevd

dakr:
This patch addresses a potential performance optimization targeting
well-defined environments such as embedded environments, e.g. automotive
control units, where everything is controlled by systemd and all
existing dependencies are known and properly modelled.

While booting up it is likely that systemd-udevd is started after a
couple of uevents were already triggered by the kernel and therefore
were missed.

Usually, this is resolved by systemd-udev-trigger.service, which simply
re-triggers all devices and subsystems, including also the ones that
were already observed by systemd-udevd.

However, this can take quite some time and is needless in well-defined
environments where everything is executed as systemd unit and all
dependencies are modelled properly.

To speed things up in such environments device_start() is introduced by
this patch and used to re-trigger the corresponding uevent of a device
unit. The device unit is pulled in as a dependency of another unit
depending on the particular device represented by this device unit.

The re-trigger is only done in case the device is not following another
device, was never in plugged state before, is in the dead state, but is
actually plugged (corresponding sysfs entries are existent).

This way we can re-trigger only those uevents that were actually missed
by systemd-udevd. This is much faster than re-triggering all devices and
subsystems and balances the work of processing re-triggered, missed
uevents to the point of time where they're actually needed by another
unit.

This accounts only for device units which are explicitly referenced as
dependency by other units and are tagged in udev rules (TAG+="systemd").
As a consequence device units for all used devices in the system must be
created and a proper dependency tree must be built in order to do
without systemd-udev-trigger.service.

PR #22643, opened 2022-03-02
udev: priorize triggering of certain subsystems (e.g. block) over others

guozhijun111:
block device is critical path of bootup.
the activating time of xxx.device  is define as ueventtime - mount time.
but the block uevent is triggered whith all other type device.I think block uevent should trigger first.
the test result in my computer is(nvme0n1p3.device is the root device):
before modify:dev-nvme0n1p3.device 1140ms
after modify:dev-nvme0n1p3.device 720ms

### DIFF
--- a/units/systemd-udev-trigger.service
+++ b/units/systemd-udev-trigger.service
@@ -19,6 +19,6 @@ ConditionPathIsReadWrite=/sys
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=/bin/udevadm trigger --type=devices -s block --action=add
-ExecStart=/bin/udevadm trigger --type=subsystems --action=add
-ExecStart=/bin/udevadm trigger --type=devices -S block --action=add
+ExecStart=-udevadm trigger --type=devices -s block --action=add
+ExecStart=-udevadm trigger --type=subsystems --action=add
+ExecStart=-udevadm trigger --type=devices --action=add


units/systemd-udev-trigger.service
```diff
@@ -19,5 +19,6 @@ ConditionPathIsReadWrite=/sys
 [Service]
 Type=oneshot
 RemainAfterExit=yes
-ExecStart=-udevadm trigger --type=subsystems --action=add
-ExecStart=-udevadm trigger --type=devices --action=add
+ExecStart=/bin/udevadm trigger --type=devices -s block --action=add
```
floppym: There's no reason to replace `-udevadm` with `/bin/udevadm` here.

floppym:
I wonder if there are any negative effects to scanning block devices before "subsystems".

keszybz:
The comment on the pull request must be the message on the commit itself.

OK, I see how this *could* be faster. Do you have any measurements?

yuwata:
No no no. This is not a way to go. systemd-udevd should process parents device first, then child. Even if this works on your system and improve boot performance, we cannot officially support such hack.

guozhijun111:
>

guozhijun111:
> 
In my systemd,the time to reach multi-user.target is:
before modify:2.71s
after modify:2.61s

guozhijun111:
> No no no. This is not a way to go. systemd-udevd should process parents device first, then child. Even if this works on your system and improve boot performance, we cannot officially support such hack.

if systemd-udev-trigger.service is active,that means the root directory is mounted and block device is ready.

boucman:
It does mean rootfs is up, but it does not always mean block devices are ready... only the block device where the rootfs is.

I mean... USB disk can literally appear at any point during boot.

guozhijun111:
> It does mean rootfs is up, but it does not always mean block devices are ready... only the block device where the rootfs is.
> 
> I mean... USB disk can literally appear at any point during boot.

yes,you are right.so I think I can change it like this:

ExecStart=-udevadm trigger --type=devices -s block --action=add
ExecStart=-udevadm trigger --type=subsystems --action=add
ExecStart=-udevadm trigger --type=devices --action=add

if the system block device is all ready when systemd-udev-trigger active,the first line will trigger block uevent first.
if there is some block not ready,we can trigger block uevent  in the third line.in this case,there is no different with this commit

poettering:
hmm, so this might be OKish if it's clear it has benefits. i mean, i can't deny that block devices are probably the primary type of devices we need to wait for, hence triggering them before the others might make sense. But I'd like some verification that this really has an effect.

Moreover, I'd really prefer if we'd add a new switch to "udevam trigger" called `--priorize-subsystem=` or so, that allows listing subsystems to trigger first. Would take a list, and we'd trigger those subsystems first, in the specified order, and then afterwards everything else. That way people can tweak the order on their specific devices if they need other devices than just block devices to be triggered first...

poettering:
The priorization thing was discussed before, see: https://github.com/systemd/systemd/pull/19637#issuecomment-855966205

As back then I still think it would be the best way to go.

yuwata:
A udev rule may have `ATTRS` or `TAGS` which references a attribute or tag of a parent device, and if a child block device is processed before its parent devices, then the result may be changed by this PR.

If a kind of reordering is applied (this PR or #19637), then we need to do more carefully to keep backward compatibility.

poettering:
Yeah, it's fine to change the order of retriggering of leaves of the tree, but we should always retrigger from trunk to leaves, as first priority.

Thus an "udevadm trigger --priorize=" switch should probably iterate through all listed subsystems, go through all devices there, then first go up the tree trigger all parent devices, before triggering the enumerated devices themselves. It should keep track of all devices it triggered, so that it doesn't trigger them twice. Once it triggered all matching devices and their parents it should then do a sweep through the tree and trigger everything not triggered so far. 

That way we would guarantee that parents are always triggered and thus processed before their children, but the leaves are triggered in a bit smarter way.

poettering:
I think #22662 is the right approach for this. Let's close this one here hence, and continue there.

PR #22644, opened 2022-03-02
Use normal lists in meson instead of splitting on whitespace

keszybz:


bluca (approved):

medhefgo (approved):
Some bikeshedding: Imho, the closing `)` should be on its own line as it reduces the line changed count when a new file is added to the end of a list (meson allows the last entry to have a trailing `,`).

keszybz:
```
libdw.so.1 is not installed: libdw.so.1: cannot open shared object file: No such file or directory
Assertion 'dlopen_dw() >= 0' failed at src/test/test-dlopen-so.c:55, function run(). Aborting.
```
Unrelated, I assume.

keszybz:
> Some bikeshedding: Imho, the closing `)` should be on its own line as it reduces the line changed count when a new file is added to the end of a list (meson allows the last entry to have a trailing `,`).

Yeah, I was thinking about this… But there are two counterarguments: I requires one more line, and the lone brace looks a bit strange. Also, we were already using this style in most places.

I (weakly) prefer the current form, but if y'all prefer to have the closing paren separate, I could live with that too ;)

yuwata:
+1 for @medhefgo's suggestion. But not strong opinion, and it is also ok to do in later PR.

keszybz:
I added a commit with the suggested change.

DaanDeMeyer:
More bikeshedding, but if we do the paren on new line for files(), can we do it for lists as well? Same argument @medhefgo made for files since it allows us to use trailing commas to reduce the diff lines changed when adding new items to the end of a list.

keszybz:
I applied the same logic to file lists. For other cases, I guess it can be done, separately.

poettering:
looks all good to me, but only checked superficially.

bluca:
The packit failures are suspicious - do not seem related, but also do not happen in other PRs?

rules.d/meson.build
```diff
@@ -4,31 +4,30 @@ install_data(
         'README',
         install_dir : udevrulesdir)
 
-rules = files('''
-        60-autosuspend.rules
-        60-block.rules
```
mrc0mmand, line 9: This reformatting breaks the Fedora-only BFQ patch:

https://src.fedoraproject.org/rpms/systemd/blob/rawhide/f/use-bfq-scheduler.patch

which, by extension, breaks Packit, so I guess the patch needs to be regenerated in the Rawhide branch.

keszybz: I'll do it after this is merged. It doesn't make much sense to do it before.

PR #22646, opened 2022-03-02
systemctl: print better message if default target is masked

### LINKED ISSUE #26589: Wrong (unrelated) error message when masking user's `default.target`

253-1-arch

### Used distribution

Arch Linux

### Linux kernel version used

6.1.12-arch1-1

x86_64

### Component

systemctl

When executing `systemctl --user get-default` with `default.target` masked, the output should mention what would happen. If it really is not possible to find the default target, an explanation or relevant error message should be output. As a last resort, the error message should not look like a failed syscall when that is not the case.

The output is `Failed to get default target: Operation not possible due to RF-kill`. This is the case even in a VM with no RF-kill capabilities, no WiFi device, etc.

Moreover the part after the colon looks like a syscall failed with `ERFKILL`, yet this errno is quite unlikely to happen in a minimal virtual machine.

1. `systemctl --user mask default.target` (or whatever the default target is)
2. `systemctl --user get-default`
3. Read the output

```sh
running_in_chroot(): Permission denied
Bus n/a: changing state UNSET → OPENING
sd-bus: starting bus by connecting to /run/user/1001/systemd/private...
Bus n/a: changing state OPENING → AUTHENTICATING
Bus n/a: changing state AUTHENTICATING → RUNNING
Sent message type=method_call sender=n/a destination=org.freedesktop.systemd1 path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=GetDefaultTarget cookie=1 reply_cookie=0 signature=n/a error-name=n/a error-message=n/a
Got message type=error sender=org.freedesktop.systemd1 destination=n/a path=n/a interface=n/a member=n/a  cookie=1 reply_cookie=1 signature=s error-name=System.Error.ERFKILL error-message=Operation not possible due to RF-kill
Failed to get default target: Operation not possible due to RF-kill
Bus n/a: changing state RUNNING → CLOSED
```

dtardon:
If the default target is masked, `systemctl get-default` prints
```
Failed to get default target: Operation not possible due to RF-kill
```
That's a bit too cryptic, so let's make it clear what's actually
happening.

Fixes #26589.

### DIFF
--- a/src/systemctl/systemctl-set-default.c
+++ b/src/systemctl/systemctl-set-default.c
@@ -50,6 +50,8 @@ static int determine_default(char **ret_name) {
 
         if (install_client_side()) {
                 r = unit_file_get_default(arg_scope, arg_root, ret_name);
+                if (r == -ERFKILL)
+                        return log_error_errno(r, "Failed to get default target: Unit file is masked.");
                 if (r < 0)
                         return log_error_errno(r, "Failed to get default target: %m");
                 return 0;
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -2200,6 +2200,8 @@ static int method_get_default_target(sd_bus_message *message, void *userdata, sd
                 return r;
 
         r = unit_file_get_default(m->unit_file_scope, NULL, &default_target);
+        if (r == -ERFKILL)
+                sd_bus_error_setf(error, BUS_ERROR_UNIT_MASKED, "Unit file is masked.");
         if (r < 0)
                 return r;
 


src/systemctl/systemctl-set-default.c
```diff
@@ -51,7 +51,8 @@ static int determine_default(char **ret_name) {
         if (install_client_side()) {
                 r = unit_file_get_default(arg_scope, arg_root, ret_name);
                 if (r < 0)
-                        return log_error_errno(r, "Failed to get default target: %m");
+                        return log_error_errno(r, "Failed to get default target: %s",
+                                               r == -ERFKILL ? "Unit is masked" : strerror(-r));
```
anitazha: nit: use `strerror_safe(r)` which does `abs(r)`

dtardon: Done.

src/systemctl/systemctl-set-default.c
```diff
@@ -66,7 +67,8 @@ static int determine_default(char **ret_name) {
 
                 r = bus_call_method(bus, bus_systemd_mgr, "GetDefaultTarget", &error, &reply, NULL);
                 if (r < 0)
-                        return log_error_errno(r, "Failed to get default target: %s", bus_error_message(&error, r));
+                        return log_error_errno(r, "Failed to get default target: %s",
+                                               r == -ERFKILL ? "Unit is masked" : bus_error_message(&error, r));
```
yuwata: If the dbus error message does not contain good explanation, then please update the dbus error message instead.

src/systemctl/systemctl-set-default.c
```diff
@@ -51,7 +51,8 @@ static int determine_default(char **ret_name) {
         if (install_client_side()) {
                 r = unit_file_get_default(arg_scope, arg_root, ret_name);
                 if (r < 0)
-                        return log_error_errno(r, "Failed to get default target: %m");
+                        return log_error_errno(r, "Failed to get default target: %s",
+                                               r == -ERFKILL ? "Unit is masked" : strerror_safe(r));
```
poettering: plese don't use strerror()/strerror_safe() when you might as well use %m.

The thing is that strerror() is not thread-safe, it uses a static process-wider buffer. While systemctl is not threaded, we still try to avoid thread-unsafe APIs because you never know what kind of bg stuff libraries do, and we want our stuff to be something we can move around in the tree with least amount of headaches. LTDR, plese rework as:

```c
if (r == -ERFKILL)
        return log_error_errno(r, "Failed to get default target, as it is masked.");
if (r < 0)
        return log_error_errno(r, "Failed to get default target: %m");
```

you get the idea.

dtardon: I thought this might be preferred, but tried the short version first... Updated.

poettering:
@yuwata's point was not addressed?

poettering:
other than that looks good now.

src/systemctl/systemctl-set-default.c
poettering: also in case you do want to match for specific errors, please do so via the dbus error name, not the error number we return. We dbus error is a lot more expressive, and we translate it to an error number client-side. Hence if you have a DBusError structure — which you do here – then always use sd_bus_error_has_name() to check for the precise error.

That said, @yuwata is right: why doesn't the english language error string the server returns suffice here? if it's ugly, then it should be fixed *server* side, not client side like you do here

poettering:
ping?

dtardon:
Oops, I've completely forgotten about this one... I'll try to return to it in the next few days.

yuwata (approved):

PR #22647, opened 2022-03-02
core: check size before mmap

stoeckmann:
The data type off_t can be 64 on 32 bit systems if they have large
file support. Since mmap expects a size_t with 32 bits as second
argument truncation could occur. At worst these huge files could
lead to mmaps smaller than the previous check for small files.

This in turn shouldn't have a lot of impact because mmap allocates
at page size boundaries. This also made the PAGE_ALIGN call in
open_mmap unneeded. In fact it was neither in sync with other mmap
calls nor with its own munmap counterpart in error path.

If such large files are encountered, which is very unlikely in these
code paths, treat them with the same error as if they are too small.

### DIFF
--- a/src/basic/locale-util.c
+++ b/src/basic/locale-util.c
@@ -14,6 +14,7 @@
 #include "dirent-util.h"
 #include "env-util.h"
 #include "fd-util.h"
+#include "fileio.h"
 #include "hashmap.h"
 #include "locale-util.h"
 #include "path-util.h"
@@ -111,6 +112,9 @@ static int add_locales_from_archive(Set *locales) {
 
         if (st.st_size < (off_t) sizeof(struct locarhead))
                 return -EBADMSG;
+
+        if (file_offset_beyond_memory_size(st.st_size))
+                return -EFBIG;
 
         p = mmap(NULL, st.st_size, PROT_READ, MAP_SHARED, fd, 0);
         if (p == MAP_FAILED)
--- a/src/libsystemd/sd-hwdb/sd-hwdb.c
+++ b/src/libsystemd/sd-hwdb/sd-hwdb.c
@@ -15,6 +15,7 @@
 
 #include "alloc-util.h"
 #include "fd-util.h"
+#include "fileio.h"
 #include "hashmap.h"
 #include "hwdb-internal.h"
 #include "nulstr-util.h"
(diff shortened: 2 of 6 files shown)


src/basic/locale-util.c
```diff
@@ -109,7 +109,8 @@ static int add_locales_from_archive(Set *locales) {
         if (!S_ISREG(st.st_mode))
                 return -EBADMSG;
 
-        if (st.st_size < (off_t) sizeof(struct locarhead))
+        if (st.st_size < (off_t) sizeof(struct locarhead) ||
+            (uintmax_t) st.st_size > SIZE_MAX)
```
anitazha: nit: this can all be on one line. no need to break lines too eagerly

src/libsystemd/sd-journal/compress.c
```diff
@@ -807,6 +807,9 @@ int decompress_stream_lz4(int in, int out, uint64_t max_bytes) {
         if (fstat(in, &st) < 0)
                 return log_debug_errno(errno, "fstat() failed: %m");
 
+        if ((uintmax_t) st.st_size > SIZE_MAX)
+                return -ENOMEM;
```
anitazha: Maybe EFBIG here? It's not an out of memory situation so ENOMEM doesn't seem right.

src/basic/locale-util.c
```diff
@@ -109,7 +109,8 @@ static int add_locales_from_archive(Set *locales) {
...
                 return -EBADMSG;
```
poettering, line 114: hmm, I think we should return a different error in that case, i.e. EFBIG or so.

src/libsystemd/sd-hwdb/sd-hwdb.c
```diff
@@ -312,6 +312,9 @@ _public_ int sd_hwdb_new(sd_hwdb **ret) {
         if (hwdb->st.st_size < (off_t) offsetof(struct trie_header_f, strings_len) + 8)
                 return log_debug_errno(SYNTHETIC_ERRNO(EIO),
                                        "File %s is too short: %m", hwdb_bin_path);
+        if ((uintmax_t) hwdb->st.st_size > SIZE_MAX)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO),
```
poettering: ditto

poettering:
I wonder if we should add a helper for this, given the cast is non-obvious, i.e. 

```c
static inline bool file_offset_beyond_memory_size(off_t x) {
        assert_cc(OFF_MAX == INT64_MAX);
        if (x < 0) /* off_t is signed, filter that out */
                return false; 
        return (uint64_t) x > (uint64_t) SIZE_MAX;
}
```

And then put that in src/basic/fileio.h somewhere and use it everywhere?

Note that in systemd we require to be compiled with large file support, so off_t is actually always 64bit, though signed. SIZE_MAX can be either 32bit or 64bit, but always unsigned.

poettering:
looks good, but see suggestions above

src/basic/locale-util.c
stoeckmann: done

src/libsystemd/sd-journal/compress.c
stoeckmann: Makes sense, adjusted

src/basic/locale-util.c
stoeckmann: done

src/libsystemd/sd-hwdb/sd-hwdb.c
stoeckmann: done

stoeckmann:
Hi @poettering,

I have added your suggestion to fileio.h. But I left out the assert because OFF_MAX is not defined (and generally not available). If I just missed a systemd-specific definition I will adjust the commit though!

Thanks for review and suggestions (@anitazha as well).

poettering:
excellent, thanks

PR #22649, opened 2022-03-03
Fixups to the unit enablement logic

### LINKED ISSUE #22000: A service file symbolically linked from a directory inside systemd path can be enabled and then gives bad error when started
 >   248.3 (real hardware) and 249.5 (in Virtualbox)

**Used distribution**
> Kubuntu 21.10 and 22.04 (in Virtualbox)

**Linux kernel version used**
>`Linux GreenOne 5.13.0-22-generic #22-Ubuntu SMP Fri Nov 5 13:21:36 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`
`Linux kubuntu 5.13.0-19-generic #19-Ubuntu SMP Thu Oct 7 21:58:00 UTC 2021 x86_64 x86_64 x86_64 
GNU/Linux`

 > Intel i9-9880H anbd virtualbox

We create a unit file in home directory and link it to systemd path. It refuses to enable the unit file saying symbolic links are not supported. I am not sure why that is the case but at least the warning is meaningful. Now try to make a unit file in a subdirectory of systemd path (for example /etc/systemd/system/myown.d). It will enable the service file and when you try to run the file, the error is misleading.

I guess the correct behaviour would be catching that systemd is dealing with a linked file already at enabling phase.

Run this as root:
```
cd ~
cat > test.service <<EOF
[Unit]
Description=Test

[Service]
ExecStart=/bin/sleep infinity

[Install]
WantedBy=multi-user.target
EOF
ln -s test.service /etc/systemd/system/test.service
systemctl daemon-reload
systemctl enable test.service
```

It gives:

    Failed to enable unit: Refusing to operate on alias name or linked unit file: test.service

Now run this:

```
rm /etc/systemd/system/test.service
cd /etc/systemd/system
mkdir myown.d
cp ~/test.service myown.d/
ln -s myown.d/test.service test.service
systemctl daemon-reload
systemctl enable test.service
```
it gives:

    Created symlink /etc/systemd/system/multi-user.target.wants/test.service → /etc/systemd/system/myown.d/test.service.

Now run

    systemctl start test.service

and we get:

    Failed to start test.service: Unit test.service not found.

keszybz:


yuwata:
I think it is better to add more test cases for the fixed issues.

src/basic/unit-file.c
```diff
@@ -261,6 +261,86 @@ static int directory_name_is_valid(const char *name) {
...
+        assert(dir);
+        assert(dirfd >= 0);
+        assert(filename);
+
+        r = readlinkat_malloc(dirfd, filename, &target);
+        if (r < 0)
+                return log_warning_errno(r, "Failed to read symlink %s%s%s: %m",
+                                         dir, dir ? "/" : "", filename);
+
+        const bool is_abs = path_is_absolute(target);
```
poettering: "const bool"? please let's not start with this. Let's use it on pointers, but on local variables it's pointless clutter really.

src/basic/unit-file.c
```diff
@@ -261,6 +261,86 @@ static int directory_name_is_valid(const char *name) {
...
+        r = chase_symlinks(target, root_dir, CHASE_NOFOLLOW | CHASE_NONEXISTENT, &simplified, NULL);
+        if (r < 0)
+                return log_warning_errno(r, "Failed to resolve symlink %s/%s pointing to %s: %m",
+                                         dir, filename, target);
+
+        /* Check if the symlink goes outside of our search path.
+         * If yes, it's a linked unit file or mask, and we don't care about the target name.
+         * Let's just store the link source directly.
+         * If not, let's verify that it's a good symlink. */
+        char *tail = path_startswith_strv(simplified, search_path);
```
poettering: here otoh it would make sense to add a const, since the memory the ptr points to is foreign, and we should not change it

src/basic/unit-file.c
```diff
@@ -261,6 +261,86 @@ static int directory_name_is_valid(const char *name) {
...
+        if (!tail) {
+                log_debug("%s: linked unit file: %s/%s → %s",
+                          __func__, dir, filename, simplified);
+                dst = filename;
+
+        } else {
+                bool self_alias;
+
+                dst = basename(simplified);
```
poettering: if you touch this, maybe also use path_extract_filename()...

src/basic/unit-file.c
```diff
@@ -270,19 +270,32 @@ static int unit_file_resolve_symlink(
...
                                          dir, dir ? "/" : "", filename);
 
+        if (!dir) {
+                assert(is_path(filename));
+                r = path_extract_directory(filename, &_dir);
+                if (r < 0)
+                        return r;
+
+                dir = _dir;
+                filename = basename(filename);
```
poettering: path_extract_filename()?

src/basic/unit-file.c
```diff
@@ -261,6 +261,86 @@ static int directory_name_is_valid(const char *name) {
...
 }
 
+static int unit_file_resolve_symlink(
+                const char *root_dir,
+                char **search_path,
+                const char *dir,
+                int dirfd,
+                const char *filename,
+                char **ret_simplified,
+                const char **ret_dst) {
```
poettering: the semantics of ret_dst are so mean, i.e. returning either a pointer to the pointer or to the simplified allocated buffer.

Afaics the caller doesn't bother about the `simplified` return parameter at all. Only about `dst`. Which it eventually duplicates to add to the set. It sounds much simpler if we'd just drop ret_simplified here, and return ret_dst as properly allocated string, that the caller can then just add to the set as-is, without eeding to duplicate it.

It's simpler and faster...

src/basic/unit-file.c
keszybz: We know that the name is "well formed", and basename works nicely. It also avoid the need to pass a separate allocated pointer around and free it.

poettering: I'd really prefer to add basename() to the list of functions that LGTM and coccinelle will complain/rewrite. The semantics are just too weird I think, and it's always good to have additional checks in place that the assumptions about "well-formedness" actually hold...

keszybz: I changed the patches to use `path_extract_filename()` in new code.

src/basic/unit-file.c
keszybz: It's reworked along the suggested lines now.

bluca:
_holy patchset, Batman_

keszybz:
```
unit-config          FAIL non-zero exit status 1
storage              PASS
networkd-test.py     PASS
build-login          PASS
boot-and-services    PASS
udev                 PASS
unit-tests           FAIL badpkg
blame: https://salsa.debian.org/systemd-team/systemd.git#upstream-ci
badpkg: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
upstream             FAIL badpkg
```
So that looks at least partially unrelated, but `unit-config` is suspicious.

keszybz:
Yeah, so this would need to be adjusted. But let's see if we actually want to do this before changing the tests there.
```
======================================================================
FAIL: test_unit_alias_enable (__main__.EnableTests)
no sysv: enable unit with an alias
----------------------------------------------------------------------
  File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 204, in test_unit_alias_enable
    self.assertEqual(os.readlink(l),
AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
- ../test_enable.service
+ /lib/systemd/system/test_enable.service


======================================================================
FAIL: test_unit_enable (__main__.EnableTests)
no sysv: enable unit
----------------------------------------------------------------------
  File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 105, in test_unit_enable
    self.assertEqual(os.readlink(l),
AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
- ../test_enable.service
+ /lib/systemd/system/test_enable.service


======================================================================
FAIL: test_unit_enable_full_path (__main__.EnableTests)
systemctl enable a unit in a non-default path
----------------------------------------------------------------------
  File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 356, in test_unit_enable_full_path
    self.assertEqual(os.readlink(enable_l), f.name)
AssertionError: '../tmpstnn9ebp.service' != '/tmp/tmpstnn9ebp.service'
- ../tmpstnn9ebp.service
? ^^
+ /tmp/tmpstnn9ebp.service
? ^^^^


======================================================================
FAIL: test_unit_sysv_alias_enable (__main__.EnableTests)
with sysv: enable unit with an alias
----------------------------------------------------------------------
  File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 249, in test_unit_sysv_alias_enable
    self.assertEqual(os.readlink(l),
AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
- ../test_enable.service
+ /lib/systemd/system/test_enable.service


======================================================================
FAIL: test_unit_sysv_enable (__main__.EnableTests)
with sysv: enable unit
----------------------------------------------------------------------
  File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 150, in test_unit_sysv_enable
    self.assertEqual(os.readlink(l),
AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
- ../test_enable.service
+ /lib/systemd/system/test_enable.service
```

poettering:
I cherry-picked a bunch of the initial commits into git main, since they should be riskless and make sense on their own. Please rebase

src/shared/install.c
```diff
@@ -1344,68 +1344,21 @@ static int unit_file_load_or_readlink(
...
+                                      NULL, AT_FDCWD, path,
+                                      true, &info->symlink_target);
+        if (r < 0)
+                return r;
 
+        /* A symlink to /dev/null or an empty file?
+         * When looking under root_dir, we can't expect /dev/ to be mounted,
+         * so let's see if the path is a (possibly dangling) symlink to /dev/null. */
+        if (null_or_empty_path(info->symlink_target) > 0 ||
+            (lp->root_dir && path_equal_ptr(path_startswith(info->symlink_target, lp->root_dir), "dev/null")))
```
poettering: this test expression looks like something that should be its own helper function, i.e. null_or_empty_path_with_root() or so.

src/basic/env-file.c
```diff
@@ -414,30 +414,39 @@ static int load_env_file_push_pairs(
...
                 return r;
 
+        /* Check if the key is present */
+        for (char **t = *m; t && *t; t += 2)
+                if (streq(t[0], key)) {
+                        if (value)
+                                r = free_and_replace(t[1], value);
+                        else
+                                r = free_and_strdup(t+1, "");
+                        goto fini;
```
poettering: what happened to `…sh`?

src/shared/install.c
```diff
@@ -1152,7 +1158,8 @@ static int config_parse_also(
 
                 r = install_name_printf(info, word, info->root, &printed);
                 if (r < 0)
-                        return r;
+                        return log_syntax(unit, LOG_WARNING, filename, line, r,
+                                          "Failed to resolve unit name in Also \"%s\": %m", word);
```
poettering: suffix Also with =, i.e. Also=

src/shared/install.c
```diff
@@ -1199,14 +1206,13 @@ static int config_parse_default_instance(
 
         r = install_name_printf(i, rvalue, i->root, &printed);
         if (r < 0)
-                return r;
+                return log_syntax(unit, LOG_WARNING, filename, line, r,
+                                  "Failed to resolve instance name in DefaultInstance \"%s\": %m", rvalue);
```
poettering: similar

src/shared/specifier.c
```diff
@@ -67,6 +67,12 @@ int specifier_printf(const char *text, size_t max_length, const Specifier table[
                                         size_t k, j;
 
                                         r = i->lookup(i->specifier, i->data, root, userdata, &w);
+                                        if (r == -ENOENT)
+                                                /* Many specifiers refer to machine-id or os-release fields,
+                                                 * which will return -ENOENT if the file is not present.
+                                                 * Let's translate this to -ENODATA to make the error more
+                                                 * understandable in the lower layers of the call stack. */
+                                                return -ENODATA;
```
poettering: hmm, shouldn't the individual specifier callbcks already convert that? Sounds weird to convert the errors so late... And shouldn't this be a bit more exotic errno? i.e. EUNATCH maybe?

src/shared/install.c
```diff
@@ -49,6 +49,7 @@ typedef enum SearchFlags {
 typedef struct {
         OrderedHashmap *will_process;
         OrderedHashmap *have_processed;
+        UnitFileScope scope;
```
poettering: sounds more natural to put this first given it's initialized first?

src/shared/install.c
```diff
@@ -2521,8 +2517,8 @@ int unit_file_add_dependency(
                 size_t *n_changes) {
 
         _cleanup_(lookup_paths_free) LookupPaths lp = {};
-        _cleanup_(install_context_done) InstallContext c = {};
-        UnitFileInstallInfo *i, *target_info;
+        _cleanup_(install_context_done) InstallContext ctx = { .scope = scope };
```
poettering, line 2621: why not add line breaks after the { and before the }?

src/shared/install.c
```diff
@@ -2633,7 +2630,7 @@ int unit_file_disable(
                 size_t *n_changes) {
 
         _cleanup_(lookup_paths_free) LookupPaths lp = {};
-        _cleanup_(install_context_done) InstallContext c = {};
+        _cleanup_(install_context_done) InstallContext ctx = {};
```
poettering: initialization missing?

src/shared/install.c
```diff
@@ -2734,8 +2731,8 @@ int unit_file_get_default(
                 char **name) {
 
         _cleanup_(lookup_paths_free) LookupPaths lp = {};
-        _cleanup_(install_context_done) InstallContext c = {};
-        UnitFileInstallInfo *i;
+        _cleanup_(install_context_done) InstallContext c = { .scope = scope };
```
poettering: line breaks?

src/shared/install.c
```diff
@@ -2769,7 +2766,7 @@ int unit_file_lookup_state(
                 const char *name,
                 UnitFileState *ret) {
 
-        _cleanup_(install_context_done) InstallContext c = {};
+        _cleanup_(install_context_done) InstallContext c = { .scope = scope };
```
poettering: ditto (here and everywhere)

src/shared/specifier.c
```diff
@@ -22,6 +22,7 @@
 #include "specifier.h"
 #include "string-util.h"
 #include "strv.h"
+#include "unit-file.h"
```
poettering: I'd really like to keep the specifier stuff indepdent of unit files. We use the specifier stuff all over the place after all.

hence, instead of this, why not define a flags field to pass in the `data` parameter, i.e.

```c
typedef enum SpecifierFlags {
        SPECIFIER_ROOT_USER = 1 << 0, /* force user/group resolution to return "root", "0" */
        SPECIFIER_FAIL_USER = 1 << 1, /* make user/group specifiers fail */
} SpecifierFlags;
```

and then add a little helper func that can translate the UnitFileScope to SpecifierFlags, so that you can easily set these flags based on the scope.

(Another idea would be to make these flags an extra parameter all specifier callbacks take, separate from `data`)

src/shared/install.c
```diff
@@ -2609,34 +2609,21 @@ int unit_file_add_dependency(
                                      SEARCH_FOLLOW_CONFIG_SYMLINKS, changes, n_changes);
 }
 
-int unit_file_enable(
-                UnitFileScope scope,
-                UnitFileFlags file_flags,
-                const char *root_dir,
+static int _unit_file_enable(
```
poettering: i think the underscore prefix would be fine if we have functions which are wrapper by something else and that something else is really the only way to use the function. But that isn't the case here, it's used by multiple callers.

maybe rename this to unit_file_enable_now()? or so? or prefix with do_?

src/shared/install.c
```diff
@@ -873,7 +890,7 @@ static int find_symlinks(
                         continue;
                 }
 
-                r = find_symlinks_in_directory(d, path, root_dir, i, match_name, ignore_same_name, config_path, same_name_link);
+                r = find_symlinks_in_directory(d, path, root_dir, i, true, match_name, ignore_same_name, config_path, same_name_link);
```
poettering: cryptic and long. maybe line beak? prefix the `true` with `/* ignore_destination= */`?

src/shared/install.c
```diff
@@ -882,7 +899,7 @@ static int find_symlinks(
 
         /* We didn't find any suitable symlinks in .wants or .requires directories, let's look for linked unit files in this directory. */
         rewinddir(config_dir);
-        return find_symlinks_in_directory(config_dir, config_path, root_dir, i, match_name, ignore_same_name, config_path, same_name_link);
+        return find_symlinks_in_directory(config_dir, config_path, root_dir, i, false, match_name, ignore_same_name, config_path, same_name_link);
```
poettering: ditto

src/basic/unit-file.c
```diff
@@ -361,7 +363,7 @@ int unit_file_resolve_symlink(
         }
 
         *ret_destination = TAKE_PTR(dst);
-        return 0;
+        return tail ? false : true;  /* true if linked unit file */
```
poettering: `!tail`?

poettering:
what a PR. did a round of reviews. some points above. I think it would be good to split some parts out of this PR and merge independently.

I already cherry picked a bunch of commits into git main.

needs rebase hence.

src/basic/env-file.c
keszybz: Hmm, what do you mean? I can't parse the comment.

src/shared/install.c
keszybz: I think that as long as only one field is initialized, it's OK to do it in one line. But I don't feel very strongly about this, I can change it if you think it looks better.

src/shared/install.c
keszybz: Ouch!

src/shared/specifier.c
keszybz: So… I don't like this approach. There'd be 1:1 mapping of the new enum and UnitFileScope. UnitFileScope is well established and widely used, and adding a new enum would just introduce a layer of abstraction that doesn't simplify anything. And I think it's OK to pass this as `data`. We don't use `data` for anything else, and for those specifiers, we're unlikely to ever do.

The part that is somewhat ugly are the conversions using `PTR_TO_INT` and `INT_TO_PTR`. But afaics, this part we can't get rid of.

src/basic/unit-file.c
keszybz: Unless I'm utterly confused, this is correct. If we have `tail`, then it's an alias, and this function is supposed to return false.

keszybz:
Updated. I implemented the suggested changes, except as noted above. The only new stuff is commit "basic/stat-util: add null_or_empty_path_with_root()" near the beginning.

keszybz:
```
  ^-- SC2016 (info): Expressions don't expand in single quotes, use double quotes for that.
```
Fuck, that's why I added single quotes!

yuwata:
```
497/980 test-specifier                                                     FAIL             0.22s   killed by signal 6 SIGABRT
>>> SYSTEMD_LANGUAGE_FALLBACK_MAP=/root/src/src/locale/language-fallback-map MALLOC_PERTURB_=189 PATH=/root/build:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /root/build/test-specifier
stderr:
/* test_specifier_escape */
/* test_specifier_escape_strv */
/* test_specifier_printf */
/* test_specifier_real_path */
Assertion 'r >= 0 || r == -EUNATCH' failed at src/test/test-specifier.c:102, function test_specifier_real_path(). Aborting.
```

keszybz:
ci/semaphoreci/pr: debian autopkgtest (lxc) failed with " unit-config          FAIL non-zero exit status 1 23:37 ". This is the same expected failure that was reported above.

bluca:
> Yeah, so this would need to be adjusted. But let's see if we actually want to do this before changing the tests there.
> 
> ```
> ======================================================================
> FAIL: test_unit_alias_enable (__main__.EnableTests)
> no sysv: enable unit with an alias
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 204, in test_unit_alias_enable
>     self.assertEqual(os.readlink(l),
> AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
> - ../test_enable.service
> + /lib/systemd/system/test_enable.service
> 
> 
> ======================================================================
> FAIL: test_unit_enable (__main__.EnableTests)
> no sysv: enable unit
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 105, in test_unit_enable
>     self.assertEqual(os.readlink(l),
> AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
> - ../test_enable.service
> + /lib/systemd/system/test_enable.service
> 
> 
> ======================================================================
> FAIL: test_unit_enable_full_path (__main__.EnableTests)
> systemctl enable a unit in a non-default path
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 356, in test_unit_enable_full_path
>     self.assertEqual(os.readlink(enable_l), f.name)
> AssertionError: '../tmpstnn9ebp.service' != '/tmp/tmpstnn9ebp.service'
> - ../tmpstnn9ebp.service
> ? ^^
> + /tmp/tmpstnn9ebp.service
> ? ^^^^
> 
> 
> ======================================================================
> FAIL: test_unit_sysv_alias_enable (__main__.EnableTests)
> with sysv: enable unit with an alias
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 249, in test_unit_sysv_alias_enable
>     self.assertEqual(os.readlink(l),
> AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
> - ../test_enable.service
> + /lib/systemd/system/test_enable.service
> 
> 
> ======================================================================
> FAIL: test_unit_sysv_enable (__main__.EnableTests)
> with sysv: enable unit
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/tmp/autopkgtest.OvTYXW/build.O10/systemd/debian/tests/unit-config", line 150, in test_unit_sysv_enable
>     self.assertEqual(os.readlink(l),
> AssertionError: '../test_enable.service' != '/lib/systemd/system/test_enable.service'
> - ../test_enable.service
> + /lib/systemd/system/test_enable.service
> ```

Updated:

https://salsa.debian.org/systemd-team/systemd/-/commit/e92d491d5af9304b93d0889bc2f3eb8604257c3f

keszybz:
centos ci (arch linux) – 20:44:09 TEST-46-HOMED

src/shared/tests.c
```diff
@@ -132,6 +134,20 @@ int log_tests_skipped_errno(int r, const char *message) {
...
 }
 
+int write_tmpfile(char *pattern, const char *contents) {
+        _cleanup_close_ int fd = -1;
+
+        fd = mkostemp_safe(pattern);
+        if (fd < 0)
+                return fd;
+
+        ssize_t l = strlen(contents);
```
bluca, line 147: strlen will bomb on null, so there should be at least an assert

test/test-systemctl-enable.sh
```diff
@@ -0,0 +1,523 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
bluca, line 2: is this used from any of our tests?

bluca (changes requested):
> Since the user is allowed
> to move the unit file, or add a new unit file in a different location, and
> we don't actually follow the symlink, I think it's OK to create a dangling
> symlink. The prefix of "../" is useful to give a hint that the link points
> to files that are conceptually "one level up" in the directory hierarchy.

I think this is fine, but it needs to be documented clearly, as it will be quite unexpected to see dandling symlinks. We need to clarify in the manpage that this is an 'internal API' of sorts, and not supposed to point to the actual file. I don't think it's obvious for users.

keszybz:
> I think this is fine, but it needs to be documented clearly, as it will be quite unexpected to see dandling symlinks. We need to clarify in the manpage that this is an 'internal API' of sorts, and not supposed to point to the actual file. I don't think it's obvious for users.

I think this is known and expected, but I'll update the docs.

test/test-systemctl-enable.sh
keszybz: Yikes. I was calling the test by hand and forgot to hook it up.

keszybz:
Updated. I added the asserts, and hooked up the test, and added one more commit that tries to describe the alias/linking rules as they are.

keszybz:
> https://salsa.debian.org/systemd-team/systemd/-/commit/e92d491d5af9304b93d0889bc2f3eb8604257c3f

Thanks!

man/systemd.unit.xml
```diff
@@ -154,8 +154,12 @@
     template instance (e.g. <literal>alias@inst.service</literal>) may be a symlink to different template
     (e.g. <literal>template@inst.service</literal>). In that case, just this specific instance is aliased,
     while other instances of the template (e.g. <literal>alias@foo.service</literal>,
-    <literal>alias@bar.service</literal>) are not aliased. Those rule preserve the requirement that the
-    instance (if any) is always uniquely defined for a given unit and all its aliases.</para>
+    <literal>alias@bar.service</literal>) are not aliased. Those rules preserve the requirement that the
+    instance (if any) is always uniquely defined for a given unit and all its aliases. The target of alias
+    symlink must point to a valid unit file location, i.e. the symlink target name must match the symlink
+    source name as described, and the desination path must be in one of the unit search paths, see UNIT FILE
```
bluca: /desination/destination/

man/systemd.unit.xml
```diff
@@ -175,11 +179,18 @@
...
+    <varname>Wants=</varname> and <varname>Requires=</varname>, see below. The preferred way to create
+    symlinks in the <filename>.wants/</filename> or <filename>.requires/</filename> directories is by
+    specifying the dependency in [Install] section of the target unit, and creating the symlink in the file
+    system with the <command>enable</command> or <command>preset</command> commands of
+    <citerefentry><refentrytitle>systemctl</refentrytitle><manvolnum>1</manvolnum></citerefentry>.  The
+    target can be a normal unit (either plain or a specific instance of a template unit). In case when the
+    source unit is a template, the target can also be a template, in which case the instance will be
+    "propagated" to the target unit to form a valid unit instance. The target of symlinks in
+    <filename>.wants/</filename> or <filename>.requires/</filename> must thus point to a valid unit file
+    location, i.e. the symlink target name must satisfy the described requirements, and the desination path
```
bluca: same typo

bluca:
Looks like the new test is failing here and there

keszybz:
This is new: CentOS CI (CentOS Stream 8):
```
12:08:27 [   60.674243] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.675765] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.676849] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.677478] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.677909] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.679033] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.680076] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.680146] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.779856] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.780586] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
12:08:27 [   60.781654] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
```
@yuwata, any idea?

keszybz:
I hope I fixed the tests. Test for %a is disabled for now.

bluca (approved):

yuwata:
> This is new: CentOS CI (CentOS Stream 8):
> ```
> 12:08:27 [   60.674243] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.675765] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.676849] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.677478] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.677909] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.679033] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.680076] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.680146] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.779856] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.780586] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> 12:08:27 [   60.781654] systemd-udevd[634]: Assertion '!event_pid_changed(s->event)' failed at src/libsystemd/sd-event/sd-event.c:2585, function sd_event_source_set_enabled(). Ignoring.
> ```
> @yuwata, any idea?

Ouch. Will fix it tomorrow.

src/basic/unit-file.c
```diff
@@ -395,77 +471,33 @@ int unit_file_build_name_map(
...
-                                        log_debug("%s: alias: %s/%s → %s", __func__, *dir, de->d_name, dst);
-                                }
 
                         } else {
-                                dst = filename;
+                                dst = TAKE_PTR(_filename_free); /* Grab the copy we made previously, if available. */
+                                if (!dst)
+                                        dst = strdup(filename);
+                                if (!dst)
+                                        return log_oom();
```
poettering: doesnt matter, but why not do:

```c
if (!dst) {
        dst = strdup(filename);
        if (!dst)
                return log_oom();
}
```

i think  that'd be a less suprising way to write this down, no?

src/basic/stat-util.c
```diff
@@ -127,13 +127,16 @@ bool null_or_empty(struct stat *st) {
...
-        /* If we have the path, let's do an easy text comparison first. */
-        if (path_equal(fn, "/dev/null"))
+        /* A symlink to /dev/null or an empty file?
+         * When looking under root_dir, we can't expect /dev/ to be mounted,
+         * so let's see if the path is a (possibly dangling) symlink to /dev/null. */
+
+        if (path_equal_ptr(path_startswith(fn, root ?: "/"), "dev/null"))
                 return true;
 
         if (stat(fn, &st) < 0)
```
poettering: maybe go the whole way properly, and replace this with chase_symlinks_and_stat()? Would make this fully correct and would actualy be minor change only.

src/basic/unit-file.c
```diff
@@ -473,7 +505,7 @@ int unit_file_build_name_map(
 
                                 r = unit_file_resolve_symlink(lp->root_dir, lp->search_path,
                                                               *dir, dirfd(d), de->d_name,
-                                                              &dst);
+                                                              false, &dst);
```
poettering: doesn't matter, but maybe prefix "false" with a comment `/* resolve_destination_target= */` to make this easier to grok?

src/basic/os-util.c
```diff
@@ -170,13 +170,19 @@ int open_extension_release(const char *root, const char *extension, char **ret_p
                         }
                 }
         } else {
-                FOREACH_STRING(p, "/etc/os-release", "/usr/lib/os-release") {
-                        r = chase_symlinks(p, root, CHASE_PREFIX_ROOT,
+                const char *var = secure_getenv("SYSTEMD_OS_RELEASE");
+                if (var)
```
poettering, line 174: i wonder if it wouldn't be better to ignore the env var if root is set?

src/basic/env-file.c
poettering: in 99% of the cases we call the label we jump to for cleaning up "finish". But here you called it "fini", i.e. the "sh" somehow vanished. And I wondered where it went.

src/test/test-specifier.c
```diff
@@ -138,4 +138,18 @@ TEST(specifiers) {
...
 
+TEST(specifiers_missing_data_ok) {
+        _cleanup_free_ char *resolved = NULL;
+
+        assert_se(setenv("SYSTEMD_OS_RELEASE", "/dev/null", 1) == 0);
+        assert_se(specifier_printf("%A-%B-%M-%o-%w-%W", SIZE_MAX, specifier_table, NULL, NULL, &resolved) >= 0);
+        assert_se(streq(resolved, "-----"));
+
+        assert_se(setenv("SYSTEMD_OS_RELEASE", "/nosuchfileordirectory", 1) == 0);
+        assert_se(specifier_printf("%A-%B-%M-%o-%w-%W", SIZE_MAX, specifier_table, NULL, NULL, &resolved) == -ENOENT);
```
poettering: maybe this behaviour isn't ideal. i.e. maybe we should treat a missing os-release file like an empty one in this context,

src/shared/specifier.c
```diff
@@ -270,37 +271,41 @@ int specifier_architecture(char specifier, const void *data, const char *root, c
...
-int specifier_os_id(char specifier, const void *data, const char *root, const void *userdata, char **ret) {
+static int parse_os_release_specifier(const char *root, const char *id, char **ret) {
+        int r;
+
         assert(ret);
-        return parse_os_release(root, "ID", ret);
+
+        /* Translate error for missing os-release file to EUNATCH. */
+        r = parse_os_release(root, id, ret);
+        return r == -ENOENT ? -EUNATCH : r;
```
poettering, line 285: so yeah, I have the suspicion we should just make this graceful, i.e. treat ENOENT like empty

src/shared/install.c
```diff
@@ -865,7 +866,7 @@ static int find_symlinks(
 static int find_symlinks_in_scope(
                 UnitFileScope scope,
```
poettering, line 943: you can drop this param now? it's also available in info->scope?

src/shared/specifier.c
poettering: UnitFileScope is not well established outside of unit file handling. tmpfiles.d, sysusers.d and all the other tools that deal with specifiers cannot make sense of that...

poettering: I mean, I would be more sympathetic to reusing UnitFileScope if it would be generalized, i.e. InvocationScope or AppScope or so, so that we can reasonably use it in all our tools, after all similar concepts kinda do make sense there.

poettering: or tldr: I really don#t like the fact that specifier.h now includes unit-file.h.

poettering: generalizing UnitFileScope as InvocationScope makes a ton of sense btw, we use similar concepts everywhere after all, i.e. when talking about D-Bus busses and stuff

src/basic/unit-file.c
poettering: yes i am just saying, why write such a complex ternary op, if `!tail` also works, and it is much shorter.

src/shared/install.c
```diff
@@ -1794,20 +1794,22 @@ static int install_info_symlink_alias(
                 q = install_name_printf(scope, info, *s, info->root, &dst);
                 if (q < 0) {
                         unit_file_changes_add(changes, n_changes, q, *s, NULL);
-                        return q;
+                        r = r < 0 ? r : q;
```
poettering, line 1886: one day we should add a macro for this, i.e. `r = ERRNO_COLLECT(r, q)` or so

src/basic/unit-file.c
```diff
@@ -111,6 +112,11 @@ int unit_validate_alias_symlink_or_warn(int log_level, const char *filename, con
 
         /* dst checks */
 
+        if (streq(src, dst))
+                return log_debug_errno(SYNTHETIC_ERRNO(ELOOP),
```
poettering, line 116: shouldn't we be careful with ELOOP? might also happen for other reasons, no?

src/shared/install.c
```diff
@@ -2702,17 +2702,66 @@ int unit_file_disable(
...
+        int r;
+
+        STRV_FOREACH(a, names_or_paths) {
+                _cleanup_(install_context_done) InstallContext ctx = { .scope = scope };
+                UnitFileInstallInfo *i = NULL;
+                _cleanup_free_ char *n = NULL;
+
+                r = path_extract_filename(*a, &n);
+                if (r < 0)
+                        return r;
```
poettering, line 2812: btw, it think we should generate an error if path_extrct_filename() returns O_DIRECTORY. (it does that in case the path is suffixed by `/`). It returns 0 if the path is not suffixed like that. And here we only care about files, hence we should refuse such paths to dirs.

src/shared/install.c
```diff
@@ -2702,17 +2702,66 @@ int unit_file_disable(
...
+
+                r = strv_consume(&names, TAKE_PTR(n));
+                if (r < 0)
+                        return r;
+
+                const char *p = NULL;
+                if (i && i->path)
+                        /* Use startswith here, because we know that paths are normalized, and
+                         * path_startswith() would give us a relative path, but we need an absolute path
+                         * relative to i->root. */
```
poettering: an absolute paths that is relative? waahhhh? not following?

this funciton is supposed to normalize, but the paths already *are* normalized?

not following here...

src/shared/install.c
```diff
@@ -1606,7 +1606,8 @@ static int install_info_traverse(
                                 return -ELOOP;
                 }
 
-                r = install_info_follow(ctx, i, lp, flags, false);
+                r = install_info_follow(ctx, i, lp, flags,
+                                        i->type == UNIT_FILE_TYPE_LINKED);  /* If linked, don't look at the target name */
```
poettering: as usual, I'd prefix this with the parameter name in a comment

src/shared/install.c
```diff
@@ -423,21 +423,54 @@ void unit_file_dump_changes(int r, const char *verb, const UnitFileChange *chang
...
          * of the chroot. False negatives are possible. */
 
-        if (!root)
-                root = "/";
+        const char *root = lp->root_dir ?: "/";
+        _cleanup_free_ char *dirname = NULL;
+        int r;
+
+        if (!path_is_absolute(target_a) || !path_is_absolute(target_b)) {
+                dirname = dirname_malloc(src);
```
poettering: path_extract_directory()

src/shared/install.c
```diff
@@ -423,21 +423,54 @@ void unit_file_dump_changes(int r, const char *verb, const UnitFileChange *chang
...
+                if (!dirname)
+                        return log_oom();
+        }
+
+        _cleanup_free_ char *a = strjoin(path_is_absolute(target_a) ? root : dirname, "/", target_a);
+        _cleanup_free_ char *b = strjoin(path_is_absolute(target_b) ? root : dirname, "/", target_b);
```
poettering: path_join()?

(yes, the paths might be normalized, but so what, this is not performance critical. And it is shorter, and easier to read)

poettering (changes requested):

src/basic/os-util.c
keszybz: The opposite: I think it's particularly useful when operating with `--root`, because you might have os-release that is not set up yet, or is only partially filled in. It's also quite useful for testing.

src/basic/env-file.c
keszybz: Oh, I see. No particular reason, I can make it "finish".

src/test/test-specifier.c
keszybz: I thought about this, and that was my initial thought. But we generally expect os-release to be there. And if you're calling 'enable' on files with specifiers, I think it's more likely that something went wrong with the installation and os-release is missing (or in the wrong path or something) than that it's missing on purpose. We only throw the error if specifiers that refer to os-release are used. And it is OK to have an empty os-release file, that is tested in one of the tests. This also lightly pushes people towards installing os-release always, which I think is a reasonable thing to do.

src/basic/unit-file.c
```diff
@@ -259,27 +259,45 @@ static int directory_name_is_valid(const char *name) {
...
+        assert(dir || path_is_absolute(filename));
+        assert(dirfd >= 0 || dirfd == AT_FDCWD);
 
         r = readlinkat_malloc(dirfd, filename, &target);
         if (r < 0)
                 return log_warning_errno(r, "Failed to read symlink %s%s%s: %m",
                                          dir, dir ? "/" : "", filename);
 
+        if (!dir) {
+                assert(is_path(filename));
```
yuwata: `path_is_absolute()`?? Moreover, this is not necessary, as it is checked at the beginning of this function.

src/basic/unit-file.c
```diff
@@ -259,27 +259,45 @@ static int directory_name_is_valid(const char *name) {
...
+                r = path_extract_directory(filename, &_dir);
+                if (r < 0)
+                        return r;
+
+                dir = _dir;
+                filename = basename(filename);
```
yuwata: `path_extract_filename()`

src/basic/unit-file.c
keszybz: It was already used for this, this just moves the creation point to a different spot. The calling code catches ELOOP, so I think this should be OK.

src/shared/install.c
keszybz: I extended the comment.

src/basic/unit-file.c
keszybz: Oh, I see. I changed it like this now. I thought that the explicit form would be clearer, but guess not.

yuwata (changes requested):

src/shared/specifier.c
keszybz: New patch at the end now renames it to LookupScope.

keszybz:
Updated. I think I addressed all the comments. I few things that were questions/suggestions are not done, but I think it's OK to leave them for later. I replied on the comments. Taking liberty to restore the green label.

keszybz:
It's been running for 5+ h. I'll merge this without waiting for focal-*.

buczek:
~~Did this change the semantic of symlink chains like `/etc/systemd/system/default.target` -> `/etc/local/systemd_system_default.target` -> ` /lib/systemd/system/graphical.target` ? It looks like in v242 such a chain would be interpreted as an alias and in v260.2 it would be interpreted as a "linked unit file" with the result that dependencies of graphical.target are ignored. I didn't analyze the code in this PR, but I guess, this change is in here. Is that correct?~~

~~One could argue that the new behavior matches systemd.unit(5), but on the other hand, the man page itself was only updated regarding this in this PR (ecd6c000d3a2e7). It is a surprise and broke our systems.~~

Edit: Sorry, I guess the relevant change was in e8630e6952 ("pid1: use a cache for all unit aliases") and not in this PR.

PR #22650, opened 2022-03-03
sd-stub: do not print warning if filesystem is not supported

alfonsosanchezbeato:
Do not print and stall in case we try to load the file system protocol from an
unsupported file system like ext4, just return EFI_UNSUPPORTED instead.

### DIFF
--- a/src/boot/efi/cpio.c
+++ b/src/boot/efi/cpio.c
@@ -324,6 +324,7 @@ EFI_STATUS pack_cpio(
         _cleanup_freepool_ void *buffer = NULL;
         UINT32 inode = 1; /* inode counter, so that each item gets a new inode */
         EFI_STATUS err;
+        EFI_FILE_IO_INTERFACE *volume;
 
         assert(loaded_image);
         assert(target_dir_prefix);
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
                 return EFI_SUCCESS;
         }
 
-        root = LibOpenRoot(loaded_image->DeviceHandle);
-        if (!root)
-                return log_error_status_stall(EFI_LOAD_ERROR, L"Unable to open root directory.");
+        err = BS->HandleProtocol(loaded_image->DeviceHandle,
+                                 &FileSystemProtocol, (void*)&volume);
+        /* Error will be unsupported if the bootloader doesn't implement the
+         * file system protocol on its file handles.
+         */
+        if (err == EFI_UNSUPPORTED) {
+                *ret_buffer = NULL;
+                *ret_buffer_size = 0;
+                return EFI_SUCCESS;
+        }
+        if (EFI_ERROR(err))
+                return log_error_status_stall(
+                                err, L"Unable to load file system protocol: %r", err);
+
+        err = volume->OpenVolume(volume, &root);
+        if (EFI_ERROR(err))
+                return log_error_status_stall(
+                                err, L"Unable to open root directory: %r", err);
 
         if (!dropin_dir)
                 dropin_dir = rel_dropin_dir = xpool_print(L"%D.extra.d", loaded_image->FilePath);


src/boot/efi/cpio.c
```diff
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
                 return EFI_SUCCESS;
         }
 
-        root = LibOpenRoot(loaded_image->DeviceHandle);
-        if (!root)
-                return log_error_status_stall(EFI_LOAD_ERROR, L"Unable to open root directory.");
+        err = uefi_call_wrapper(BS->HandleProtocol, 3, loaded_image->DeviceHandle,
+                                &FileSystemProtocol, (VOID*)&volume);
+        /* It will be unsupported if not vfat. This can happen if the efi was
```
poettering: why would that be unsupported? Not sure I follow? What does that have to do with vfat vs. ext4?

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
...
+         * loaded by a bootloader with support for non-vfat partitions, like
+         * grub. Do not stall in that case.
+         */
+        if (err == EFI_UNSUPPORTED)
+                return err;
```
poettering: So my understanding is that grub simply doesn't implement a file system driver at all, but just has its own loader, that doesn't implement the `FileSystemProtocol`?

if that's the case, then the commit msg/comments should be changed, to make this clear, the hint regrding ext4 is highly misleading, because it has little to do with the choice of backing fs but everything with the set of APIs Grub decided to implement on the backing handle

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
...
+        if (EFI_ERROR(err)) {
+                return log_error_status_stall(
+                                err, L"Unable to load file system protocol: %r", err);
+        }
+
+        err = uefi_call_wrapper(volume->OpenVolume, 2, volume, &root);
```
poettering: We do not use uefi_call_wrapper() anymore in our codebase.

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
...
+        /* It will be unsupported if not vfat. This can happen if the efi was
+         * loaded by a bootloader with support for non-vfat partitions, like
+         * grub. Do not stall in that case.
+         */
+        if (err == EFI_UNSUPPORTED)
+                return err;
+        if (EFI_ERROR(err)) {
+                return log_error_status_stall(
+                                err, L"Unable to load file system protocol: %r", err);
+        }
```
poettering: no {} around single-line if blocks, as per coding style

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
...
+
+        err = uefi_call_wrapper(volume->OpenVolume, 2, volume, &root);
+        if (EFI_ERROR(err)) {
+                return log_error_status_stall(
+                                err, L"Unable to open root directory: %r", err);
+        }
```
poettering: as above, no {} around single-line if blocks

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,24 @@ EFI_STATUS pack_cpio(
                 return EFI_SUCCESS;
         }
 
-        root = LibOpenRoot(loaded_image->DeviceHandle);
-        if (!root)
-                return log_error_status_stall(EFI_LOAD_ERROR, L"Unable to open root directory.");
+        err = uefi_call_wrapper(BS->HandleProtocol, 3, loaded_image->DeviceHandle,
+                                &FileSystemProtocol, (VOID*)&volume);
```
poettering: we don't use uefi_call_wrapper() anymore in our codebase

src/boot/efi/cpio.c
alfonsosanchezbeato: UEFI mandates a version of FAT for the system partition, so usually implementations will only implement support for that.

src/boot/efi/cpio.c
alfonsosanchezbeato: Right, grub implements a few filesystems itself instead of relying on UEFI to load files. I will reformat the commit message to make this more clear.

src/boot/efi/cpio.c
poettering: Sure, but the `FileSystemProtocol` EFI protocol is fs independent. So if you EFI environment implements ext4, then the protocol should sitll work.

I think Grub doesn't actually implement ext4 as an UEFI fs though, right? it just does its own thing, loads everything into memory on its own, just exposing the loaded files, but never the backing fs.

So, the comments are confusing here, and the commit msg, because it's not about the choice of backing fs, but the fact that grub simply doesn't implement the file system protocol at *all* for the EFI binaries.

poettering: I mean, Grub *could* if it wanted implement the FileSystemProtocol stuff, which is after all exactly what this project does: https://github.com/pbatard/efifs

It just chose not to do that.

medhefgo: Also, if we were loaded from PXE, the device handle would also not likely support file systme protocol.

alfonsosanchezbeato:
Thanks for the comments - to be clear, the scenario here is UEFI loading grub from an ESP, then grub loading the image from a different, non-vfat partition. I've tried to clarified in comment & commit message.

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,23 @@ EFI_STATUS pack_cpio(
                 return EFI_SUCCESS;
         }
 
-        root = LibOpenRoot(loaded_image->DeviceHandle);
-        if (!root)
-                return log_error_status_stall(EFI_LOAD_ERROR, L"Unable to open root directory.");
+        err = BS->HandleProtocol(loaded_image->DeviceHandle,
+                                 &FileSystemProtocol, (VOID*)&volume);
```
medhefgo: Lowercase `void` please.

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,23 @@ EFI_STATUS pack_cpio(
...
 
-        root = LibOpenRoot(loaded_image->DeviceHandle);
-        if (!root)
-                return log_error_status_stall(EFI_LOAD_ERROR, L"Unable to open root directory.");
+        err = BS->HandleProtocol(loaded_image->DeviceHandle,
+                                 &FileSystemProtocol, (void*)&volume);
+        /* EFI only supports vfat, so the error will be "unsupported" if the
+         * image was loaded from a non-vfat partition, which might happen if,
+         * say, UEFI loaded grub from an ESP, but then grub loaded the image
+         * from a different, non-vfat partition. Do not stall in that case.
```
medhefgo: `Do not warn in that case`. The stalling is just a "side-effect" :D

I also think the mention of only vfat being supported should go. I'm pretty sure ASUS will happily boot from NTFS and apple apparently only likes hfs+. The only important thing is whether the image comes from something that doesn't support the protocol that we need to read the files.

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,23 @@ EFI_STATUS pack_cpio(
                 return EFI_SUCCESS;
         }
 
-        root = LibOpenRoot(loaded_image->DeviceHandle);
-        if (!root)
-                return log_error_status_stall(EFI_LOAD_ERROR, L"Unable to open root directory.");
+        err = BS->HandleProtocol(loaded_image->DeviceHandle,
+                                 &FileSystemProtocol, (void*)&volume);
+        /* EFI only supports vfat, so the error will be "unsupported" if the
```
poettering: I think this is still misleading. EFI dictates *at* *least* VFAT are supported, and implementations might support more, and there are projects that make more available via EFI drivers.

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,23 @@ EFI_STATUS pack_cpio(
...
+         * image was loaded from a non-vfat partition, which might happen if,
+         * say, UEFI loaded grub from an ESP, but then grub loaded the image
+         * from a different, non-vfat partition. Do not stall in that case.
+         */
```
poettering, line 344: This should just say that Grub doesn't implement the file system protocol on its file handles, and that's it. Drop all references to supported or non-supported file systems please, it has nothing to do with that.

src/boot/efi/cpio.c
```diff
@@ -336,9 +337,23 @@ EFI_STATUS pack_cpio(
...
+        if (err == EFI_UNSUPPORTED)
+                return err;
```
poettering: actually, please don#t fail here at all. instead return EFI_SUCCESS, i.e. bypass the whole logic, like we do for the case where loaded_image->DeviceHandle is not set at all, as above. And initialize the two return parameters to NULL/0 in that case too, exactly like in that case.

i.e. we have this coding style that functions should either log on all errors or on none. Thus, having a function that suppresses the logging on this one error but leaves it on all others is not according to our coding style, except if it also turns it into EFI_SUCCESS.

alfonsosanchezbeato: Ok, I've repushed changing this now.

poettering:
lgtm

PR #22651, opened 2022-03-03
shared/install: don't fail with dangling symlinks in unit_file_preset…

fbuihuu:
…_all()

4a203a5177b7d9aa499221c315bc0e327a23b5cf changed the behavior of 'systemctl
preset-all' when it encounters dangling symlinks.

Before 4a203a5177b7d9aa499221c315bc0e327a23b5cf, a dangling symlink wasn't
treated as an error and 'preset-all' kept proceeding with the remaining
units. Such unexpected cases however were still reported with:

  Unit test-dangling.service is an alias to a unit that is not present, ignoring.

and the message indicated that the problem was ignored.

After 4a203a5177b7d9aa499221c315bc0e327a23b5cf, 'preset-all' failed right after
encountering dangling symlinks. The process was aborted and could leave the
system with some of the presets unprocessed. The error was reported as follow:

  Failed to preset all units: File test-dangling.service: Link has been severed

This patch restores the old behavior.

### DIFF
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -2206,6 +2206,11 @@ static int install_error(
                                               "Unit file %s does not exist.", changes[i].path);
                         goto found;
 
+                case -ENOLINK:
+                        r = sd_bus_error_setf(error, BUS_ERROR_NO_SUCH_UNIT,
+                                              "Unit file %s is a dangling symlink.", changes[i].path);
+                        goto found;
+
                 default:
                         assert(changes[i].type_or_errno < 0); /* other errors */
                         r = sd_bus_error_set_errnof(error, changes[i].type_or_errno, "File %s: %m", changes[i].path);
--- a/src/shared/install.c
+++ b/src/shared/install.c
@@ -398,12 +398,16 @@ void unit_file_dump_changes(int r, const char *verb, const UnitFileChange *chang
                                         verb, changes[i].path);
                         logged = true;
                         break;
-
                 case -ENOENT:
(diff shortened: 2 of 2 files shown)


fbuihuu:
Cc @keszybz

fbuihuu:
It looks like my analysis is wrong and even with the patch applied, dangling symlinks are still treated as errors. The patch only prevents PID1 from aborting the process prematurely and slightly improve the error message.

Hence with this patch `systemctl preset-all` still reports errors such as:

```
# systemctl preset-all 
Failed to preset all units: Unit file test-dangling.service is a dangling symlink.
# echo $?
1
```

Note for `systemctl --root=/ preset-all`, the old behavior is restored:

```
# systemctl --root=/ preset-all 
Failed to preset unit, unit test-dangling.service is an alias to an unexistent unit.
# echo $?
0
```

fbuihuu:
So maybe we should simply revert 4a203a5177b7d9aa499221c315bc0e327a23b5cf...

poettering:
Looks superficially OK, but let's see what @keszybz has to say

fbuihuu:
I'm not sure it is (at least I still need to update the commit message), see my previous comment.

Basically 4a203a5177b7d9aa499221c315bc0e327a23b5cf makes `preset-all` returns an exit status != 0 when an unexpected case happens (dangling symlinks, masked units, ...).

Before this commit, these corner cases were still logged but not considered as errors (exit status = 0).

So the behavior changed but not sure which one is preferred.

Also there's an inconsistency depending on whether `--root=/` is used or not (ie whether `preset-all` is run on client side or by PID1).

PR #22652, opened 2022-03-03
devnode-acl: use _cleanup_ to free acl_t

dtardon:


src/shared/devnode-acl.c
```diff
@@ -94,68 +92,47 @@ int devnode_acl(const char *path,
...
-        if (acl_calc_mask(&acl) < 0) {
-                r = -errno;
-                goto finish;
-        }
-
-        if (acl_set_file(path, ACL_TYPE_ACCESS, acl) < 0) {
-                r = -errno;
-                goto finish;
-        }
+                return r;
```
poettering: i tink it would be better to write `return 0` here, after all we want to return success here, and the return code of some earlier call doesn't matter really at this point.

poettering:
looks good, just one minor nit

src/shared/devnode-acl.c
```diff
@@ -52,7 +52,7 @@ int devnode_acl(const char *path,
                 bool del, uid_t old_uid,
                 bool add, uid_t new_uid) {
 
-        acl_t acl;
+        _cleanup_(acl_freep) acl_t acl = NULL;
         int r = 0;
```
poettering: you can drop this initialization now

dtardon: Done.

src/shared/devnode-acl.c
dtardon: Done. I've also proposed a second commit that reverses that condition, as it looks slightly better to me that way.

src/shared/devnode-acl.c
```diff
@@ -123,14 +123,13 @@ int devnode_acl(const char *path,
                 }
         }
 
-        if (!changed)
-                return 0;
-
-        if (acl_calc_mask(&acl) < 0)
-                return -errno;
+        if (changed) {
```
yuwata, line 126: I do not see any merits of the second commit. If nothing is changed, then we can safely return the function earlier. This commit increase unnecessary indentation.

yuwata (changes requested):
The first commit looks good to me. Please drop the second commit.

poettering:
I would agree with @yuwata here: exit early is *good*. deep indentation is *bad*. It's not always clear-cut, but I think here the status quo ante was better.

First patch looks perfect now.

bluca:
shall we simply cherry-pick the first commit to main?

poettering:
indeed, done now. thanks!

poettering:
See 203ea2c8f158288fea56c5be980715b2b7e002fe

PR #22656, opened 2022-03-03
portable: add return parameter to GetImageMetadataWithExtensions

keszybz:


bluca:
As I mentioned in the ticket, I'd really rather just break compat here... there's no way the current borken version can be of any use to anybody, and codesearch shows no usage. I know we normally don't, but really, the current version is just useless - and that's my bad of course, I added it, not sure what I was smoking!

keszybz:
There is at least one user: our own code, and we don't require client and server code to be in sync, and we also support remote calls.

But in this case this is all indeed very unlikely to matter. I'd be fine with breaking compat too. I'll update the patch if there's consensus that that's the way to go.

bluca:
> There is at least one user: our own code, and we don't require client and server code to be in sync, and we also support remote calls.
> 
> But in this case this is all indeed very unlikely to matter. I'd be fine with breaking compat too. I'll update the patch if there's consensus that that's the way to go.

@poettering are you ok with breaking compat here? I would strongly prefer to go that way

bluca:
@keszybz what about splitting the other fixes in a separate PR? They look good and could be merged quickly

poettering:
I'd break compat too here. I think we can get away with it.

poettering:
I cherry picked the 4 auxiliary patches into git main now.

poettering:
Patch looks Ok superficially, otherwise, but let's break compat here

bluca:
> I'd break compat too here. I think we can get away with it.

thanks - @keszybz I'm in a bit of a rush to get this backported, so I'll take care of updating this immediately myself

bluca:
Needs to be backported to v249-stable and v250-stable - especially the former, so that I can get the change into Ubuntu 22.04, as it will live for many years.

yuwata:
The last commit is cherry-picked to the main branch.

keszybz:
LGTM.

Issue #22658, opened 2022-03-03
DNS doesn't work anymore in VPN after update


ltuikov:
**systemd version the issue has been seen with**
 > systemd 249 (v249.9-1.fc35)
+PAM +AUDIT +SELINUX -APPARMOR +IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified

**Used distribution**
 > Fedora 35

**Linux kernel version used** (`uname -a`)

 > `Linux localhost.localdomain 5.16.11-200.fc35.x86_64 #1 SMP PREEMPT Wed Feb 23 17:08:49 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**CPU architecture issue was seen on**
 > AMD Ryzen

**Expected behaviour you didn't see**
 > DNS to work in VPN--it used to work BEFORE a package update I did a few weeks ago. (sudo dnf update)

**Unexpected behaviour you saw**
 > DNS cannot resolve a work website after connecting on VPN. It used to work fine before a package update a rew weeks ago.

**Steps to reproduce the problem**
 > Update to the latest packages from Fedora 35, run VPN to my work, try to reach a work website--can't resolve. But if I remote "search ." from /run/systemd/resolve/stub-resolv.conf and /run/systemd/resolve/resolv.conf, it works fine. The problem is "search .".

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
...
```
The problem seems to be the unconditional addition of "search ." in resolv.conf by commit b3ffa2b5f3aa68.
"search ." should be added conditionally, say by a config option in /etc/systemd/resolved.conf, but not unconditionally. Or the correct fix should be implemented as described in the last paragraph in commit b3ffa2b5f3aa68, but definitely not adding "search ." unconditionally.
After connecting to work VPN, this line becomes "search . somthing.com" and this fails all DNS queries to work sites, and this used to work before a recent upgrade.

poettering:
What VPN software do you use?

Does it patch /etc/resolv.conf itself? 

[...]

PR #22659, opened 2022-03-03
Revert "resolved: turn off that a search domain is derived from the h…

ltuikov:
…ost's fqdn"

This reverts commit b3ffa2b5f3aa68dc6ab15893d5eeba8906aa3a9e.

Do not unconditionally add "search ." to resolv.conf, since this breaks DNS
lookup in VPN connections, as it leaves the search line as
"search . something.com".

Fixes: #22658
Cc: Lennart Poettering <lennart@poettering.net>
Signed-off-by: Luben Tuikov <luben.tuikov@amd.com>

### DIFF
--- a/src/resolve/resolv.conf
+++ b/src/resolve/resolv.conf
@@ -16,4 +16,3 @@
 
 nameserver 127.0.0.53
 options edns0 trust-ad
-search .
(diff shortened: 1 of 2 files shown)


ltuikov:
Should probably let distro maintainers know.

poettering:
Hmm, wut? the "search ." line is only added if no search domains are configured.

This patch makes no sense to me, the commit msg in no way matches what this does.

poettering:
Pretty sure this is a bug in your VPN software, see my comment on #22658.

PR #22660, opened 2022-03-04
logind: add uaccess tag for keyboards and mice

abouvier:
Treats keyboards and mice on par with joysticks.
Fixes https://github.com/libretro/RetroArch/issues/5033 for example.


poettering:
No, sorry. Let's not do that. We do input device switching via logind, so that only the fg session gets access to a device, for security reasons. If we'd set "uaccess" on the device that'd become pointless.

Wayland and X11 and all those tools know perfectly well how to get the right session-switched input fds out of logind, I don#t think we need to change anythere here.

poettering:
Generally, applications should get their input via X11 or Waylna,d not from the low level drivres. unless they are display servers of their own, in which case they should just use the logind dbus apis like wayland/x11 do it.

PR #19949, opened 2021-06-16
udevadm: trigger: implement database-{no}match arguments

dakr:
systemd-udev-trigger.service by default triggeres all devices regardless
of whether they were already recognized by systemd-udevd.

There are machines (especially in embedded environments) where
systemd-udev-trigger.service is configured to run at a later stage of
the boot sequence, which can lead to quite a lot of devices being
triggered although they were already recognized by systemd-udevd.

Re-triggering a lot of devices is a relatively expensive operation and
therefore should be avoided if unnecessary.

Therefore this patchset introduces --database-nomatch, which filters out
devices that are already present in the udev database. For consistance
reasons --database-match is implemented as well, which filters out devices
that are *not* already present in the udev database.

PR #22662, opened 2022-03-04
udevadm trigger: add --prioritized-subsystem option

yuwata:
Replaces #19637, #19949, and #22643.

### DIFF
--- a/man/udevadm.xml
+++ b/man/udevadm.xml
@@ -396,6 +396,35 @@
           </listitem>
         </varlistentry>
         <varlistentry>
+          <term><option>--initialized-match</option></term>
+          <term><option>--initialized-nomatch</option></term>
+          <listitem>
+            <para>When <option>--initialized-match</option> is specified, trigger events for devices
+            that are already initialized by <command>systemd-udevd</command>, and skip devices that
+            are not initialized yet.</para>
+            <para>When <option>--initialized-nomatch</option> is specified, trigger events for devices
+            that are not initialized by <command>systemd-udevd</command> yet, and skip devices that
+            are already initialized.</para>
+            <para>Here, initialized devices are those for which at least one udev rule already
+            completed execution – for any action but <literal>remove</literal> — that set a property
+            or other device setting (and thus has an entry in the udev device database). Devices are
+            no longer considered initialized if a <literal>remove</literal> action is seen for them
+            (which removes their entry in the udev device database). Note that devices that have no
+            udev rules are never considered initialized, but might still be announced via the sd-device
+            API (or similar). Typically, it is thus essential that applications which intend to use
+            such a match, make sure a suitable udev rule is installed that sets at least one property
+            on devices that shall be matched.</para>
+            <para>WARNING: <option>--initialized-nomatch</option> can potentially save a significant
+            amount of time compared to re-triggering all devices in the system and e.g. can be used to
+            optimize boot time. However, this is not safe to be used in a boot sequence in general.
+            Especially, when udev rules for a device depend on its parent devices (e.g.
+            <literal>ATTRS</literal> or <literal>IMPORT{parent}</literal> keys, see
+            <citerefentry><refentrytitle>udev</refentrytitle><manvolnum>7</manvolnum></citerefentry>
+            for more details), the final state of the device becomes easily unstable with this option.
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
           <term><option>-w</option></term>
           <term><option>--settle</option></term>
           <listitem>
--- a/src/udev/udevadm-trigger.c
+++ b/src/udev/udevadm-trigger.c
@@ -11,6 +11,7 @@
 #include "device-util.h"
 #include "fd-util.h"
 #include "fileio.h"
+#include "parse-util.h"
 #include "path-util.h"
 #include "process-util.h"
 #include "set.h"
@@ -226,6 +227,8 @@ static int help(void) {
                "  -y --sysname-match=NAME           Trigger devices with this /sys path\n"
                "     --name-match=NAME              Trigger devices with this /dev name\n"
                "  -b --parent-match=NAME            Trigger devices with that parent device\n"
+               "     --initialized-match            Trigger devices that are already initialized\n"
+               "     --initialized-nomatch          Trigger devices that are not initialized yet\n"
                "  -w --settle                       Wait for the triggered events to complete\n"
                "     --wait-daemon[=SECONDS]        Wait for udevd daemon to be initialized\n"
                "                                    before triggering uevents\n"
@@ -243,6 +246,8 @@ int trigger_main(int argc, char *argv[], void *userdata) {
                 ARG_PING,
                 ARG_UUID,
                 ARG_PRIORITIZED_SUBSYSTEM,
+                ARG_INITIALIZED_MATCH,
+                ARG_INITIALIZED_NOMATCH,
         };
 
         static const struct option options[] = {
@@ -260,6 +265,8 @@ int trigger_main(int argc, char *argv[], void *userdata) {
                 { "sysname-match",         required_argument, NULL, 'y'                       },
                 { "name-match",            required_argument, NULL, ARG_NAME                  },
                 { "parent-match",          required_argument, NULL, 'b'                       },
+                { "initialized-match",     no_argument,       NULL, ARG_INITIALIZED_MATCH     },
+                { "initialized-nomatch",   no_argument,       NULL, ARG_INITIALIZED_NOMATCH   },
                 { "settle",                no_argument,       NULL, 'w'                       },
                 { "wait-daemon",           optional_argument, NULL, ARG_PING                  },
                 { "version",               no_argument,       NULL, 'V'                       },
@@ -426,6 +433,12 @@ int trigger_main(int argc, char *argv[], void *userdata) {
                         }
                         break;
                 }
+                case ARG_INITIALIZED_MATCH:
+                case ARG_INITIALIZED_NOMATCH:
+                        r = device_enumerator_add_match_is_initialized(e, c == ARG_INITIALIZED_MATCH ? MATCH_INITIALIZED_YES : MATCH_INITIALIZED_NO);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to set initialized filter: %m");
+                        break;
                 case 'V':
                         return print_version();
                 case 'h':
--- a/shell-completion/bash/udevadm
+++ b/shell-completion/bash/udevadm
@@ -51,7 +51,8 @@ _udevadm() {
         [INFO_STANDALONE]='-r --root -a --attribute-walk -x --export -e --export-db -c --cleanup-db
                            -w --wait-for-initialization --value'
         [INFO_ARG]='-q --query -p --path -n --name -P --export-prefix -d --device-id-of-file --property'
-        [TRIGGER_STANDALONE]='-v --verbose -n --dry-run -q --quiet -w --settle --wait-daemon --uuid'
+        [TRIGGER_STANDALONE]='-v --verbose -n --dry-run -q --quiet -w --settle --wait-daemon --uuid
+                              --initialized-match --initialized-nomatch'
         [TRIGGER_ARG]='-t --type -c --action -s --subsystem-match -S --subsystem-nomatch
                        -a --attr-match -A --attr-nomatch -p --property-match
                        -g --tag-match -y --sysname-match --name-match -b --parent-match
--- a/shell-completion/zsh/_udevadm
+++ b/shell-completion/zsh/_udevadm
@@ -34,6 +34,8 @@ _udevadm_trigger(){
         '--tag-match=property[Trigger events for devices with a matching tag.]' \
         '--sysname-match=[Trigger events for devices with a matching sys device name.]' \
         '--parent-match=[Trigger events for all children of a given device.]' \
+        '--initialized-match[Trigger events for devices that are already initialized.]' \
+        '--initialized-nomatch[Trigger events for devices that are not initialized yet.]' \
         '--uuid[Print synthetic uevent UUID.]' \
         '--prioritized-subsystem=[Trigger events for devices which belong to a matching subsystem earlier.]'
 }


yuwata:
@dakr and @guozhijun111, does this help your issue or improve performance of boot process?

guozhijun111:
> @dakr and @guozhijun111, does this help your issue or improve performance of boot process?

my systemd boot process is kernel+initrd+systemd+lightdm

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -278,16 +299,99 @@ static int device_compare(sd_device * const *_a, sd_device * const *_b) {
...
+
+        typesafe_qsort(devices + m, n - m, device_compare);
+
+        enumerator->n_devices = n;
+        free_and_replace(enumerator->devices, devices);
+        return 0;
+
+failed:
+        for (size_t i = 0; i < n; i++)
+                sd_device_unref(devices[i]);
```
poettering: i wonder if we should add a helper `device_unref_many(sd_device **array, size_t n);` that implements this loop. It appears to me we have that a couple of times.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -278,16 +299,99 @@ static int device_compare(sd_device * const *_a, sd_device * const *_b) {
...
         assert_return(enumerator, -EINVAL);
         assert_return(device, -EINVAL);
 
-        if (!GREEDY_REALLOC(enumerator->devices, enumerator->n_devices + 1))
-                return -ENOMEM;
+        r = sd_device_get_syspath(device, &syspath);
+        if (r < 0)
+                return r;
+
+        r = hashmap_ensure_put(&enumerator->devices_by_syspath, &string_hash_ops, syspath, device);
```
poettering, line 446: string_hash_ops → path_hash_ops?

poettering: hmm, shouldn't this be an ordered hashmap?

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -278,16 +299,99 @@ static int device_compare(sd_device * const *_a, sd_device * const *_b) {
...
+        }
+
+        m = n;
+        while ((device = hashmap_steal_first(enumerator->devices_by_syspath)))
+                devices[n++] = device;
+
+        typesafe_qsort(devices + m, n - m, device_compare);
+
+        enumerator->n_devices = n;
+        free_and_replace(enumerator->devices, devices);
```
poettering, line 424: hmm, don't we have to unref the sd_device objects in the old array first?

src/udev/udevadm-trigger.c
```diff
@@ -229,7 +229,8 @@ static int help(void) {
                "  -w --settle                       Wait for the triggered events to complete\n"
                "     --wait-daemon[=SECONDS]        Wait for udevd daemon to be initialized\n"
                "                                    before triggering uevents\n"
-               "     --uuid                         Print synthetic uevent UUID\n",
+               "     --uuid                         Print synthetic uevent UUID\n"
+               "     --prioritized-subsystem        Trigger devices from a matching subsystem earlier\n",
```
poettering: suffix left column with =SUBSYS or so

man/udevadm.xml
```diff
@@ -297,6 +297,13 @@
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term><option>--prioritized-subsystem=<replaceable>SUBSYSTEM</replaceable></option></term>
+          <listitem>
+            <para>When triggering events for devices, the devices from the specified subsystem and
+            their parents are triggered earlier. This option can be specified multiple times.</para>
```
poettering: earlier → first

poettering:
So, I think we probably should also adjust the systemd-udevd-trigger.service unit file to prioritize the "block" subsystem with this, given it's the primary device we'll wait for. And maybe "tpmrm", too, since we might need TPM to boot through.

poettering:
hmm, so if people use `--prioritized-subsystem=block,tpmrm` then this probably should mean 1st priority is block, second priority is tpmrm, and everything else is coming after that. With this current PR tpmrm and block are both at the same priority though... 

i.e. I think the prioritized_subsystem field that currently is a Set should probably just be a simple array or so (or an ordered set, but not sure that's necessary, presumably the number of prioritized subsystems is relatively small), and then when sorting the devices we should take the order of subsystems in this list into account as additional sort key...

poettering:
> my systemd boot process is kernel+initrd+systemd+lightdm
> in initrd,all uevent triggered once.so most device is already probed and ready to work。
> in systemd all event will trigger again to add some tag and attr.
> all udev event will trigger twice in boot process,this maybe can be improved

The initrd environment is typically dumbed down, i.e. carries fewer/different udev rules, fewer drivers and such. Moreover it might run slightly different versions of things than the host after transition. Hence we really have to retrigger to be safe and correct. And we typically value correctness and robustness over mere performance... So no, retriggering these devices really shouldn't go away.

src/libsystemd/sd-device/device-enumerator.c
yuwata: I initially thought here path_hash_ops should be used, but syspath is always normalized, so it is not necessary, and string_hash_ops is faster.

I cannot find no reason that ordered hashmap should be used here... Why do you think so?

yuwata:
Updated. All comments, except for the hash_ops and ordered hashmap things, are addressed. PTAL.

dakr:
> @dakr and @guozhijun111, does this help your issue or improve performance of boot process?

Yes and no, I tried this already when @poettering originally proposed it and it does improve the boot process in a way that some services are available earlier than before. So, I can confirm it improves things. However, for my specific use-cases this is not enough since,

1. While particular subsystems are preferred and therefore services are unblocked earlier, the general load in the system keeps up, since also devices of no interest within this subsystem are triggered and other subsystems are still triggered afterwards slowing down the startup of the high priority services.
2. It does not decrease the overall work to be done and therefore the overall boot time reduction is limited to the amount we save by optimizing parallelism, which of course is perfectly fine in general, since this is the purpose of this patch.

I found that for my use-cases it is best to start the "services of high priority" in a separate target before sysinit.target and do the triggering with this https://github.com/systemd/systemd/pull/19637 approach, since everything else will still be too slow. However, I agree that this approach is not generic and not for distributions and therefore not for upstream (yet).

For the rest of the boot process I found that this https://github.com/systemd/systemd/pull/19949 is a huge improvement and depending on the amount of devices within the system and the time when systemd-udev-trigger.service is executed, this can easily save a couple of seconds.

In my case, since we run the high priority services in a separate target before, we don't really have the case any longer that at this later stage we'd need to prioritize subsystems. However, I could imagine that combining this (re-triggering only devices not already processed by udevd) plus the --prioritize-subsystem option could have some nice synergies in general.

src/libsystemd/sd-device/device-enumerator.c
poettering: oh, mostly for illustrative purposes, i.e. it makes clear what the contents here is. I mean, does the performance really matter here?

Also, I'd consider it preparation for the future. Given we now have this indexed already, maybe sooner or later we'll expose this, i.e. add sd_device_enumerator_get_device_by_syspath() or so, and once we do that we cannot rely on normalization of the passed path.

but dunno... doesn't really matter, we could of course still change this later if needed.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -62,11 +62,25 @@ _public_ int sd_device_enumerator_new(sd_device_enumerator **ret) {
         return 0;
 }
 
+static void device_unref_many(sd_device **devices, size_t n) {
```
poettering, line 70: maybe add:

```c
assert(devices || n == 0);
```

yuwata:
Added a functionality to skip already processed devices based on #19949, but with a slightly different approach.

yuwata:
cc @dakr. PTAL the last three commits.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -249,26 +248,22 @@ static int device_compare(sd_device * const *_a, sd_device * const *_b) {
...
                         }
                 }
         }
 
         /* md and dm devices are enumerated after all other devices */
-        delay_a = strstr(devpath_a, "/block/md") || strstr(devpath_a, "/block/dm-");
-        delay_b = strstr(devpath_b, "/block/md") || strstr(devpath_b, "/block/dm-");
-        r = CMP(delay_a, delay_b);
+        r = CMP(strstr(devpath_a, "/block/md") || strstr(devpath_a, "/block/dm-"),
+                strstr(devpath_b, "/block/md") || strstr(devpath_b, "/block/dm-"));
```
poettering: if you touch this, maybe we should make this a tiny bit nicer with a helper:

```c
static bool syspath_is_late_block(const char *p) {
        return strstr(p, "/block/md") || strstr(p, "/block/dm-");
}
```

so that you then just call:

```c
r = CMP(syspath_is_late_block(devpath_a), syspath_is_late_block(devpath_b));
```

which is nicer on the eye and less duplication

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -313,6 +405,11 @@ static int enumerator_sort_devices(sd_device_enumerator *enumerator) {
 
         enumerator->sorted = true;
         return 0;
+
+failed:
+        device_unref_many(devices, n);
+        free(devices);
+        return r;
```
poettering, line 432: hmm, so what about an algorithm like this: we first generate a 1:1 array with all devices. And then we go through that list once for each priorized subsystem: if the device we look at is from the priorized subsystem we move it the front, swapping the position with the first device that we haven't processed so far.

```c
size_t p = 0;
STRV_FOREACH(prioritized_subsystem, enumerator->prioritized_subsystems) {
        for (size_t i = 0; i < n; i++) {
                if (!device_matches_subsystem(devices[i], *prioritized_subsystem))
                        continue;
                SWAP_TO(devices[i], devices[p]);
                p++;
        }

        qsort(devices + n_sorted, p - n_sorted);
}
```
(pseudo-code)

with that logic we wouldn#t have to remove/readd anything to the hashmap, we'd just reorder the array in neat partitions, and then sort each partition.

wouldn't thta work?

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -548,20 +548,8 @@ static int enumerator_scan_dir_and_add_devices(sd_device_enumerator *enumerator,
                         continue;
                 }
 
-                /*
```
poettering, line 399: hm, this is quite a change from the status quo ante if i grok this properly. preivously we said that devices for which no props are set just don't exist in our view. With this change now they do. This will mean that on some systems the number of db entries will explode. i.e. some systems with ridiciulous amounts of NUMA memory slots have like a million deivces. so far we'd process those in parallel but not store anything about this. which made this OK. But if we now store a db entry for each things would become quite different.

What's the rationale for this? So far we said: if you want to use the `sd_device_is_initialzed()` to wor, make sure to have a rule on your devices that sets at least one prop. That way people could specifically select the devices they care about, and thus avoid the NUMA memory slot issue, since noone would set a rule for those...

src/udev/udevadm-trigger.c
```diff
@@ -266,6 +269,7 @@ int trigger_main(int argc, char *argv[], void *userdata) {
                 { "help",                  no_argument,       NULL, 'h'                       },
                 { "uuid",                  no_argument,       NULL, ARG_UUID                  },
                 { "prioritized-subsystem", required_argument, NULL, ARG_PRIORITIZED_SUBSYSTEM },
+                { "initialized",           required_argument, NULL, ARG_INITIALIZED           },
```
poettering: so far all flags directly mapping to matches where called xyz-match and xyz-nomatch. That might not be ideal, but it's also not completely bad, so I'd stick to that?

poettering:
looks mostly ok, but I am not a fan of the change that creates a db entry even for devices that have no rules. I think it matters that we don't generate db entries for devices that have zero props, to be able to handle systems with ridiculous numbers of devices, as pointed out.

src/libsystemd/sd-device/device-enumerator.c
yuwata: I think your proposal is inefficient, as `device_matches_subsystem()` must search its all child devices and check their subsystems match `*prioritized_subsystem`. Consider the case that a sd-device object which refers to a PCI bus which has a block device in the tree...

src/udev/udevadm-trigger.c
yuwata: Originally @dakr introduced this as `--database-match` and `--database-nomatch`, but here existence of database is used as if the device is initialized (or processed by udevd), and its connection is not trivial for users, I think. So, I changed the name.
If you'd like to 'match' and 'nomatch', then, how about '--match-initialized' '--match-uninitialized'? Though, still I prefer the current form: '--initialized'.

src/libsystemd/sd-device/device-enumerator.c
yuwata: Yeah, this change make udevd may create huge number of database files. That's expected, and this is for making `sd_device_is_initialized()` work for all devices. But I did not consider the NUMA case. I will drop the commit.

yuwata:
Updated.
- Two more helpers to compare enumerated devices.
- Dropped the commit which generates udev database for all devices.

But keep the naming the new option `--initialized` as is. Please see the above. But I am quite neutral about the naming, Please let me know if you have a better idea.
Also, the logic to sorting devices is kept, see the above comment.

PTAL. 

(@dakr I kept the author of the last commit as your original commit in #19949. If you want to drop or change it, please let me know.)

src/udev/udevadm-trigger.c
poettering: Hmm I think, `--initialized-match=` + `--initialized-nomatch=` sounds like the way to go, is it fits into both naming schemes we so far adopted. i.e. match/unmatch and "initialized"

poettering:
Cherry picked the first two commits into git main, given they are kinda bugfixes and make sense on their own.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -223,49 +223,64 @@ int device_enumerator_add_match_is_initialized(sd_device_enumerator *enumerator)
         return 1;
 }
 
-static int device_compare(sd_device * const *_a, sd_device * const *_b) {
-        sd_device *a = *(sd_device **)_a, *b = *(sd_device **)_b;
-        const char *devpath_a, *devpath_b, *sound_a;
-        bool delay_a, delay_b;
-        int r;
+static int sound_device_compare(const char *a, const char *b) {
```
poettering: it's not clear these params are the devpath of these devices. maybe add comment (or rename params?)

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -223,49 +223,64 @@ int device_enumerator_add_match_is_initialized(sd_device_enumerator *enumerator)
...
+        const char *sound_a, *sound_b;
+        unsigned prefix_len;
```
poettering: actually, size_t is more appropriate

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -995,6 +996,66 @@ _public_ sd_device *sd_device_enumerator_get_subsystem_next(sd_device_enumerator
...
+                if (k < 0)
+                        r = k;
+        } else if (enumerator->match_parent) {
+                k = enumerator_scan_devices_children(enumerator);
+                if (k < 0)
+                        r = k;
+        } else {
+                const char *subsysdir;
+
+                if (access("/sys/subsystem", F_OK) >= 0)
```
poettering: i wonder if we should drop this stuff. `/sys/subsystem` is preparation for a future that never came. And given that the main proponent of this left Linux kernel development (Kay), I doubt this will ever come. So maybe we should start dropping references to /sys/subsystem/ given it's unlikely to materialize anytime soon.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -513,10 +514,45 @@ static bool match_sysname(sd_device_enumerator *enumerator, const char *sysname)
...
+                return true;
+
+        r = sd_device_get_is_initialized(device);
+        if (r < 0) {
+                /* this is necessarily racey, so ignore missing devices */
+                if (r != -ENOENT)
+                        return r;
+
+                return false;
+        }
```
poettering:
```c
if (r == -ENOENT)
        return false;
if (r < 0)
        return r;
```

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -513,10 +514,45 @@ static bool match_sysname(sd_device_enumerator *enumerator, const char *sysname)
...
+
+        if (enumerator->match_initialized == r)
```
poettering: given that both of these are `int`, and not `bool`, but you actually want a boolean comparison here, i'd write this as `if (!enumerator->match_initialized == !r)`

of course, if everything is alright the only values stored in these variables are 0 + 1, but I think it's useful to communicate the boolean concept here to the reader

src/libsystemd/sd-device/device-enumerator.c
poettering: (but this is just a general comment, not really an issue with this PR)

poettering:
looks really good, only some minor points.

yuwata:
Thank you for the review. I hope all comments are addressed. Also, I added a warning in the man page about using `--initialized-nomatch`. PTAL.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -223,49 +223,64 @@ int device_enumerator_add_match_is_initialized(sd_device_enumerator *enumerator)
...
-                                        return r;
-                        }
-                }
-        }
+        if (!sound_a)
+                return 0;
+
+        sound_a += STRLEN("/sound/card");
+        sound_a = strchr(devpath_a, '/');
+
```
poettering: i'd drop this empty line here btw, given the next line is kinda the error check for the preceeding line here, and we usually say that function calls and their immediate error handling should be kept together.

but doesn't matter at all... and I am aware this was this way int he old code too.

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -223,49 +223,64 @@ int device_enumerator_add_match_is_initialized(sd_device_enumerator *enumerator)
...
+        if (r != 0)
+                return r;
 
         /* md and dm devices are enumerated after all other devices */
-        delay_a = strstr(devpath_a, "/block/md") || strstr(devpath_a, "/block/dm-");
-        delay_b = strstr(devpath_b, "/block/md") || strstr(devpath_b, "/block/dm-");
-        r = CMP(delay_a, delay_b);
+        r = CMP(devpath_is_late_block(devpath_a), devpath_is_late_block(devpath_b));
         if (r != 0)
                 return r;
```
poettering, line 325: btw, i am a bit puzzled why we don't use path_compare() instead of strcmp() at the end of device_compare(). Is this historical?

(Yeah, unrelated to this commit, but I just noticed, might be something to improve?)

src/libsystemd/sd-device/device-enumerator.c
```diff
@@ -556,31 +588,13 @@ static int enumerator_scan_dir_and_add_devices(sd_device_enumerator *enumerator,
...
-                 * For now, we can only check these types of devices, we
-                 * might not store a database, and have no way to find out
-                 * for all other types of devices.
-                 */
-                if (!enumerator->match_allow_uninitialized &&
-                    !initialized &&
-                    (sd_device_get_devnum(device, NULL) >= 0 ||
-                     sd_device_get_ifindex(device, NULL) >= 0))
-                        continue;
-
```
poettering, line 414: so previously you could chose between two modes:

1. enumerate all devices
2. enumerate all devices, but exclude the ones that have a device node or a netif and aren't initialized yet

I think the second item stems from a different way of thinking: i.e. it comes from a time where udev was still creating device nodes, and thus there was a guarantee that devices with device nodes would have to have udev rules to work at all. And thus it might some sense to restrict the effect of `match_allow_uninitialized` to such devices, because it was clear that a device is uninitialized if it has no db entry, and to be initialized it had to have a db entry.

But that thinking changed a bit, in particular given that udev has long ceased cretaing device nodes, and devtmpfs creates them now server-side, and udev only otionally then adjusts access modes and such on the. So nowadays there's no guarantee anymore that devices with device nodes will have udev rules, because they really don't have to and will still work fine and be accessible from userspace. So since a longer time the ne thinking was: if you intend to use `match_allow_uninitialized` on some device, then make sure that there's a udev rule for it in place, too. And any use of such a match without also knowing that a rule has been added is simply broken.

Because of that I think we should revisit the logic hre, and simply consider this conditionaliuation here a thing of the past, and support it only for compat. So maybe something like this: support four different match filters:

1. all devices
2. only devices with a db entry
3. only devices without a db entry
4. only devices that have no devnode/ifindex or have a db entry

the latter would then be the compat mode we'd enable if filtering is done via the libudev compat API.

Does this make any sense?

poettering: i.e. this would be a four value enum, not a tristate anymore i guess.

and the foruth option is something we'd only provide for compat, in the new udevadm trigge rfeature we'd not expose that, because it doesn't make sense anymore in times of devtmpfs, where the underlying guarantee makes no sense anymore, as mentioned.

man/udevadm.xml
```diff
@@ -395,6 +395,26 @@
             then each matching result is ORed, that is, all children of each specified device are triggered.</para>
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term><option>--initialized-match</option></term>
+          <listitem>
+            <para>Trigger events for devices that are already initialized by
+            <command>systemd-udevd</command>, and skip devices that are not initialized yet.</para>
```
poettering: maybe explain what initialized means, i.e. something like this:

> Initialized devices are those for which at least one udev rule already completed execution – for any action but "remove" — that set a property or other device setting (and thus has an entry in the udev device database). Devices are no longer considered initialized if a "remove" action is seen for them (which removes their entry in the udev device database). Note that devices that have no udev rules are never considered initialized, but might still be announced via the sd-device API (or similar). Typically, it is thus essential that applications which intend to use such a match, make sure a suitable udev rule is installed that sets at least one property on devices that shall be matched.

Or something like that.

src/libsystemd/sd-device/device-enumerator.c
yuwata: The syspath or devpath are already normalized, hence, strcmp() is enough and faster.

yuwata:
@poettering Thank you for the suggestion. In the revised version, the legacy condition is dropped, and the suggested explanation about initialized devices are added to the man page. PTAL.

src/libsystemd/sd-device/device-enumerator.c
poettering: well, it would be enough for equality checks. But this here is about ordering. and that does make a difference, i.e. what order should these have:

```
a/x
a-b/y
a_/z
```

I think people would probably expect this:

```
a/x
a-b/y
a_a/z
```

but if you use strcmp() instead of path_compare() you'd instead get:

```
a-b/y
a/x
a_a/z
```

That's because `/` is between `-` and `a` in the ascii table. I think that's quite confusing, and we shouldn#t order that way hence.

src/libsystemd/sd-device/device-enumerator-private.h
```diff
@@ -5,6 +5,7 @@
 
 int device_enumerator_scan_devices(sd_device_enumerator *enumeartor);
 int device_enumerator_scan_subsystems(sd_device_enumerator *enumeartor);
```
poettering, line 16: btw, "enumerartor" is a typo, here and above

poettering:
lgtm. (i mean, i raised two points, but they are about stuff that strictly speaking is unrelated to the PR, hence shouldn't block the PR)

src/libsystemd/sd-device/device-enumerator.c
yuwata: Ouch, Indeed...

src/libsystemd/sd-device/device-enumerator-private.h
yuwata: Ouch, Ouch! Will fix soon.

poettering:
→ #22821

AdamWill:
I've been investigating a [bug downstream in Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=2083374) where we have problems with NetworkManager because it starts managing a wired ethernet device while it's still `eth0`, before it gets renamed by udev, then udev renames it and things get broken. Thomas Haller suggested that commit https://github.com/systemd/systemd/commit/d8b50e5df7e01983279e70c6d970fb60d053df97 from this PR may be the cause. Can you take a look? Thanks!

See comment 13 and onwards for the detailed debugging.

yuwata:
Ouch! Thank you for the report.

PR #22663, opened 2022-03-04
Various man page bits

keszybz:


bluca (approved):

bluca:
The second-to-last commit message is missing `#` so it's not auto-resolved, but it seems the PR is not picking up the issues anyway, so doesn't matter too much. Linked them manually.

DaanDeMeyer (approved):

keszybz:
Thank you both for the reviews. There really isn't anything to test here except that XML is syntactically valid, so I'll merge without delay.

PR #22667, opened 2022-03-07
random-util: unify RANDOM_ALLOW_INSECURE and !RANDOM_BLOCK and simplify

zx2c4:
RANDOM_BLOCK has existed for a long time, but RANDOM_ALLOW_INSECURE was
    added more recently, leading to an awkward relationship between the two.
    It turns out that only one, RANDOM_BLOCK, is needed.
    
    RANDOM_BLOCK means return cryptographically secure numbers no matter
    what. If it's not set, it means try to do that, but if it fails, fall
    back to using unseeded randomness.
    
    This part of falling back to unseeded randomness is the intent of
    GRND_INSECURE, which is what RANDOM_ALLOW_INSECURE previously aliased.
    Rather than having an additional flag for that, it makes more sense to
    just use it whenever RANDOM_BLOCK is not set. This saves us the overhead
    of having to open up /dev/urandom.
    
    Additionally, when getrandom returns too little data, but not zero data,
    we currently fall back to using /dev/urandom if RANDOM_BLOCK is not set.
    This doesn't quite make sense, because if getrandom returned seeded data
    once, then it will forever after return the same thing as whatever
    /dev/urandom does. So in that case, we should just loop again.
    
    Since there's never really a time where /dev/urandom is able to return
    some easily but more with difficulty, we can also get rid of
    RANDOM_EXTEND_WITH_PSEUDO. Once the RNG is initialized, bytes
    should just flow normally.
    
    This also makes RANDOM_MAY_FAIL obsolete, because the only case this ran
    was where we'd fall back to /dev/urandom on old kernels and return
    GRND_INSECURE bytes on new kernels. So also get rid of that flag.
    
    Finally, since we're always able to use GRND_INSECURE on newer kernels,
    and we only fall back to /dev/urandom on older kernels, also only fall
    back to using RDRAND on those older kernels. There, the only reason to
    have RDRAND is to avoid a kmsg entry about unseeded randomness.
    
    The result of this commit is that we now cascade like this:
    
      - Use getrandom(0) if RANDOM_BLOCK.
      - Use getrandom(GRND_INSECURE) if !RANDOM_BLOCK.
      - Use /dev/urandom if !RANDOM_BLOCK and no GRND_INSECURE support.
      - Use /dev/urandom if no getrandom() support.
      - Use RDRAND if we would use /dev/urandom for any of the above reasons
        and RANDOM_ALLOW_RDRAND is set.

### DIFF
--- a/src/basic/random-util.c
+++ b/src/basic/random-util.c
@@ -160,27 +160,59 @@ int rdrand(unsigned long *ret) {
 int genuine_random_bytes(void *p, size_t n, RandomFlags flags) {
         static int have_syscall = -1;
         _cleanup_close_ int fd = -1;
-        bool got_some = false;
+
+        if (FLAGS_SET(flags, RANDOM_BLOCK | RANDOM_ALLOW_RDRAND))
+                return -EINVAL;
 
         /* Gathers some high-quality randomness from the kernel (or potentially mid-quality randomness from
          * the CPU if the RANDOM_ALLOW_RDRAND flag is set). This call won't block, unless the RANDOM_BLOCK
-         * flag is set. If RANDOM_MAY_FAIL is set, an error is returned if the random pool is not
-         * initialized. Otherwise it will always return some data from the kernel, regardless of whether the
-         * random pool is fully initialized or not. If RANDOM_EXTEND_WITH_PSEUDO is set, and some but not
-         * enough better quality randomness could be acquired, the rest is filled up with low quality
-         * randomness.
-         *
-         * Of course, when creating cryptographic key material you really shouldn't use RANDOM_ALLOW_DRDRAND
-         * or even RANDOM_EXTEND_WITH_PSEUDO.
-         *
-         * When generating UUIDs it's fine to use RANDOM_ALLOW_RDRAND but not OK to use
-         * RANDOM_EXTEND_WITH_PSEUDO. In fact RANDOM_EXTEND_WITH_PSEUDO is only really fine when invoked via
-         * an "all bets are off" wrapper, such as random_bytes(), see below. */
+         * flag is set. If it doesn't block, it will still always return some data from the kernel, regardless
+         * of whether the random pool is fully initialized or not. When creating cryptographic key material you
+         * should always use RANDOM_BLOCK. */
 
         if (n == 0)
                 return 0;
 
-        if (FLAGS_SET(flags, RANDOM_ALLOW_RDRAND))
+        /* Use the getrandom() syscall unless we know we don't have it. */
+        if (have_syscall != 0 && !HAS_FEATURE_MEMORY_SANITIZER) {
+                for (;;) {
+                        ssize_t l = getrandom(p, n, FLAGS_SET(flags, RANDOM_BLOCK) ? 0 : GRND_INSECURE);
+
+                        if (l > 0) {
+                                have_syscall = true;
+
+                                if ((size_t) l == n)
+                                        return 0; /* Yay, success! */
+
+                                /* We didn't get enough data, so try again */
+                                assert((size_t) l < n);
+                                p = (uint8_t*) p + l;
+                                n -= l;
+                                continue;
+
+                        } else if (l == 0) {
+                                have_syscall = true;
+                                return -EIO;
+
+                        } else if (ERRNO_IS_NOT_SUPPORTED(errno)) {
+                                /* We lack the syscall, continue with reading from /dev/urandom. */
+                                have_syscall = false;
+                                break;
+
+                        } else if (errno == EINVAL) {
+                                /* If we previously passed GRND_INSECURE, and this flag isn't known, then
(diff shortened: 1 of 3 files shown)


poettering:
Hmm, there are three cases in the kernel API what to do in case the pool is not initialized yet:

1. block until it is (with neither flag set)
2. don't block and fail with EAGAIN (with GRND_NONBLOCK set)
3. don't block and give crap randomness (with GRND_INSECURE set)

We expose the first case iff RANDOM_BLOCK is set.

The second case we expose via RANDOM_MAY_FAIL.

The third case we expose via RANDOM_ALLOW_INSECURE.

Weird are the combinations of these flags though, or the absence of all three.

I kinda like the descriptiveness of the flags, i.e. that you look at a call to genuine_random_bytes() it's always clear which of the three strategies is picked. So I wonder if we shouldn't keep the flags around, but strictly require exctly one of them to be passed, to select precisely which behaviour of the three is desired. i.e. add assert() if more than or less than one of these flags are set, and ensure all our invocations pass exactly one of the three.

poettering:
(I wonder btw, if we should drop RANDOM_EXTEND_WITH_PSEUDO. The only case where we call it we also set RANDOM_ALLOW_INSECURE. Hence we'll fall back to /dev/urandom or GRND_INSECURE. But today if reading those sources works once, it will also work for subsequent calls. i.e. there might be benefit in fallback if the syscall or device node is blocked, but having a fallback for the case where it worked once but then not again sounds pretty pointless...)

(And if we drop that, and the RDRAND fallback, then the flags parameter would not be a flags parameter at all anymore, but just an enum that selects which of the three behaviours you want if pool is not initialized)

poettering:
> Yes, without a doubt. I had meant to do that last night but jetlag got the better of me. This option makes no sense for any kernels supported by systemd. Done now: #22671

I think this was originally added to deal with the fact that getrandom() can return short (and getentropy() even enforces an artificial limit of 256bytes)...

So ignoring SIGINT and old GRND_RANDOM behaviour, I take it getrandom() and read from /dev/urandom will never return short, is that correct? I.e. the only reason we still want to loop around getrandom() and read() on /dev/urandom is to deal with EINTR, right? (Given we don't use GRND_RANDOM anyway)

zx2c4:
getrandom() will return fewer bytes than what you asked when there's a pending signal and the requested length is greater than 256, so it's a good idea to loop for this case. But it won't return short because of weird entropy pool things or things like that for which there's no immediate recourse; we don't need to worry about that.

zx2c4:
I simplified this even further. I think you'll be happy with the result. PTAL.

I'm keeping the RDRAND patch still separate for now and we can talk about that separately.

zx2c4:
The rules are now this:

  - Use getrandom(0) if RANDOM_BLOCK.
  - Use getrandom(GRND_INSECURE) if !RANDOM_BLOCK.
  - Use /dev/urandom if !RANDOM_BLOCK and no GRND_INSECURE support.
  - Use /dev/urandom if no getrandom() support.
  - Use RDRAND if we would use /dev/urandom for any of the above reasons  and RANDOM_ALLOW_RDRAND is set.

So the only bad combination is `RANDOM_BLOCK | RANDOM_ALLOW_RDRAND` which no returns `-EINVAL`.

zx2c4:
Looking at the CI, it days CentOS 8S failed, but I can't see how that failure is related to this commit. I think it might be a "flake"? Or something else and maybe I should rebase on master to get the green checkbox?

Either way, I think this PR is good to go and ready to review/merge/whatever.

mrc0mmand:
> Looking at the CI, it days CentOS 8S failed, but I can't see how that failure is related to this commit. I think it might be a "flake"? Or something else and maybe I should rebase on master to get the green checkbox?
> 
> Either way, I think this PR is good to go and ready to review/merge/whatever.

This has already been fixed in the main branch, so I went ahead and retriggered the job, which should make it green once again.

zx2c4:
I don't need to rebase my commit on the main branch first for that to happen? The CI cherry picks or merged rather than checking out the tree?

mrc0mmand:
> I don't need to rebase my commit on the main branch first for that to happen? The CI cherry picks or merges rather than checking out the tree?

Most of the CIs we have check out the `merge` ref (i.e. `refs/pull/22667/merge`) instead of the `head` one, which contains the PR merged with the latest base branch (`main` in this case). This reduces the need to constantly rebase the PR branches and still have relevant CI results.

zx2c4:
@poettering @bluca CI is now all green. Is this mergeable?

poettering:
Hmm, so random_bytes() was the only user of RANDOM_MAY_BLOCK I guess, and it combined it with RANDOM_ALLOW_INSECURE which didn't make much sense. So if it now only uses the latter, you can drop the former, indeed.

poettering:
ok, looks good to me.

PR #22668, opened 2022-03-07
random-util: remove RDRAND usage

zx2c4:
```
    /dev/urandom is seeded with RDRAND. Calling genuine_random_bytes(...,
    ..., 0) will use /dev/urandom as a last resort. Hence, we gain nothing
    here by having our own RDRAND wrapper, because /dev/urandom already is
    based on RDRAND output, even before /dev/urandom has fully initialized.
    
    Furthermore, RDRAND is not actually fast! And on each successive
    generation of new x86 CPUs, from both AMD and Intel, it just gets
    slower.
    
    This commit simplifies things by just using /dev/urandom in cases where
    we before might use RDRAND, since /dev/urandom will always have RDRAND
    mixed in as part of it.
    
    And above where I say "/dev/urandom", what I actually mean is
    GRND_INSECURE, which is the same thing but won't generate warnings in
    dmesg.
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -95,6 +95,20 @@ CHANGES WITH 251:
           handling, and improving compatibility with home directories intended
           to be portable like the ones managed by systemd-homed.
 
+        * All kernels supported by systemd mix RDRAND (or similar) into the
+          entropy pool at early boot. This means that on those systems, even
+          if /dev/urandom is not yet initialized, it still returns bytes that
+          that are at least as high quality as RDRAND. For that reason, we no
+          longer have reason to invoke RDRAND from systemd itself, which has
+          historically been a source of bugs. Furthermore, kernels ≥5.6 provide
+          the getrandom(GRND_INSECURE) interface for returning random bytes
+          before the entropy pool is initialized without warning into kmsg,
+          which is what we attempt to use if available. By removing systemd's
+          direct usage of RDRAND, x86 systems ≥Broadwell that are running an
+          older kernel may experience kmsg warnings that were not seen with
+          250. For newer kernels, non-x86 systems, or older x86 systems,
+          there should be no visible changes.
+
 CHANGES WITH 250:
 
         * Support for encrypted and authenticated credentials has been added.
--- a/docs/RANDOM_SEEDS.md
+++ b/docs/RANDOM_SEEDS.md
@@ -144,33 +144,11 @@ acquired.
 ## Keeping `systemd'`s Demand on the Kernel Entropy Pool Minimal
 
 Since most of systemd's own use of random numbers do not require
-cryptographic-grade RNGs, it tries to avoid reading entropy from the kernel
-entropy pool if possible. If it succeeds this has the benefit that there's no
-need to delay the early boot process until entropy is available, and noisy
-kernel log messages about early reading from `/dev/urandom` are avoided
-too. Specifically:
-
-1. When generating [Type 4
-   UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_\(random\)),
-   systemd tries to use Intel's and AMD's RDRAND CPU opcode directly, if
-   available. While some doubt the quality and trustworthiness of the entropy
-   provided by these opcodes, they should be good enough for generating UUIDs,
-   if not key material (though, as mentioned, today's big distributions opted
-   to trust it for that too, now, see above — but we are not going to make that
-   decision for you, and for anything key material related will only use the
-   kernel's entropy pool). If RDRAND is not available or doesn't work, it will
-   use synchronous `getrandom()` as fallback, and `/dev/urandom` on old kernels
-   where that system call doesn't exist yet. This means on non-Intel/AMD
-   systems UUID generation will block on kernel entropy initialization.
-
-2. For seeding hash tables, and all the other similar purposes systemd first
-   tries RDRAND, and if that's not available will try to use asynchronous
-   `getrandom()` (if the kernel doesn't support this system call,
-   `/dev/urandom` is used). This may fail too in case the pool is not
-   initialized yet, in which case it will fall back to glibc's internal rand()
-   calls, i.e. weak pseudo-random numbers. This should make sure we use good
-   random bytes if we can, but neither delay boot nor trigger noisy kernel log
-   messages during early boot for these use-cases.
+cryptographic-grade RNGs, it tries to avoid blocking reads to the kernel's RNG,
+opting instead for using `getrandom(GRND_INSECURE)`. After the pool is
+initialized, this is identical to `getrandom(0)`, returning cryptographically
+secure random numbers, but before it's initialized it has the nice effect of
+not blocking system boot.
 
 ## `systemd`'s Support for Filling the Kernel Entropy Pool
 
@@ -280,10 +258,8 @@ early-boot entropy in most cases. Specifically:
    hosting provider if they don't. For VMs used in testing environments,
    `systemd.random_seed=` may be used as an alternative to a virtualized RNG.
 
-3. On Intel/AMD systems systemd's own reliance on the kernel entropy pool is
-   minimal (as RDRAND is used on those for UUID generation). This only works if
-   the CPU has RDRAND of course, which most physical CPUs do (but I hear many
-   virtualized CPUs do not. Pity.)
+3. In general, systemd's own reliance on the kernel entropy pool is minimal
+   (due to the use of `GRND_INSECURE`).
 
 4. In all other cases, `systemd-random-seed.service` will help a bit, but — as
    mentioned — is too late to help with early boot.
--- a/src/basic/random-util.c
+++ b/src/basic/random-util.c
@@ -35,137 +35,11 @@
 
 static bool srand_called = false;
 
-int rdrand(unsigned long *ret) {
-
-        /* So, you are a "security researcher", and you wonder why we bother with using raw RDRAND here,
-         * instead of sticking to /dev/urandom or getrandom()?
-         *
-         * Here's why: early boot. On Linux, during early boot the random pool that backs /dev/urandom and
-         * getrandom() is generally not initialized yet. It is very common that initialization of the random
-         * pool takes a longer time (up to many minutes), in particular on embedded devices that have no
-         * explicit hardware random generator, as well as in virtualized environments such as major cloud
-         * installations that do not provide virtio-rng or a similar mechanism.
-         *
-         * In such an environment using getrandom() synchronously means we'd block the entire system boot-up
-         * until the pool is initialized, i.e. *very* long. Using getrandom() asynchronously (GRND_NONBLOCK)
-         * would mean acquiring randomness during early boot would simply fail. Using /dev/urandom would mean
-         * generating many kmsg log messages about our use of it before the random pool is properly
-         * initialized. Neither of these outcomes is desirable.
-         *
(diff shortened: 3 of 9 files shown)


zx2c4:
This is marked as a draft while https://github.com/systemd/systemd/pull/22667 is reviewed first and separately.

poettering:
We so far are using RDRAND for generating UUIDs, i.e. where we want high entropy, but not quite cryptographic quality, and where the kernel random pool is not seeded with RDRAND, because distros traditionally didn't do that, specifically Debian (they set CONFIG_RANDOM_TRUST_CPU=no for the longest time, though that might have changed now). We need a lot of UUIDs (we pass one to each service invocation, as an identifier of a specific instance of a service), hence having to wait for being able to generate UUIDs at boot really sucks.

Is CONFIG_RANDOM_TRUST_CPU=no still a thing? Is this option removed from the kernel now, too?

Usually, we don't hard rely on kernel features for a few years. Are we really there yet with this?

bluca:
>     /dev/urandom is seeded with RDRAND. Calling genuine_random_bytes(...,
    ..., 0) will use /dev/urandom as a last resort. Hence, we gain nothing
    here by having our own RDRAND wrapper, because /dev/urandom already is
    based on RDRAND output, even before /dev/urandom has fully initialized

Is this true on all kernels or just new versions? As per readme we need to support from 3.12 and up

poettering:
> Is this true on all kernels or just new versions? As per readme we need to support from 3.13 and up

Which I'd be fine to raise btw, 3.13 is 2014, i.e. 8 years ago... I think supporting 5y old kernels is already more than enough.

zx2c4:
> where the kernel random pool is not seeded with RDRAND, because distros traditionally didn't do that, specifically Debian (they set CONFIG_RANDOM_TRUST_CPU=no for the longest time, though that might have changed now). 

This is actually a common misconception. CONFIG_RANDOM_TRUST_CPU does not control whether RDSEED is mixed into the pool. RDSEED is always mixed into the pool. Rather, it just controls whether it's credited for it -- whether the RNG unblocks and becomes fully operational on the basis of RDRAND. So even if you set CONFIG_RANDOM_TRUST_CPU=no, it'll still use RDRAND for GRND_INSECURE and such.

(I was hoping we'd get https://github.com/systemd/systemd/pull/22667 in first and then you'd review this after but I guess that didn't happen, so I'll mark this as non-draft now.)

poettering:
> This is actually a common misconception. CONFIG_RANDOM_TRUST_CPU does not control whether RDSEED is mixed into the pool. RDSEED is always mixed into the pool.

Ah, interesting, and that was the case already in 5y old kernels?

So this would only leave one benefit of calling RDSEED then: that we don't see that annoying /dev/urandom kernel log message on old kernels, right? Upcoming kernels don't generate that message at all anymore, right? And all kernels that support GRND_INSECURE don't either (which was added in ~2019 or so), because we go via getrandom() there?

So the question would be whether it's worth keeping the RDSEED support in place to suppress the ugly log message on kernels from before GRND_INSECURE? That's really all it boils down to, right?

poettering:
> Right, exactly. Do you want to maintain complex and historically error-prone code just for the sake or suppressing a log message on ancient kernels that likely won't be used with newer systemds in most cases? For me the choice is an easy one.

I added the RDSEED stuff because people filed all kinds of bugs against systemd here and in distro bug trackers because of that kernel log message under the assumption systemd was doing something wrong there and was crap. GRND_INSECURE was added to kernel 5.6, i.e. April 2020, which is not even 2y ago, so if new systemd would be used with kernels before that people will see the warning. As I know the Debian people they care about really old kernels and about the cosmetics of their kernel logs, so I figure they at least will not like this change.

bluca:
> > Right, exactly. Do you want to maintain complex and historically error-prone code just for the sake or suppressing a log message on ancient kernels that likely won't be used with newer systemds in most cases? For me the choice is an easy one.
> 
> I added the RDSEED stuff because people filed all kinds of bugs against systemd here and in distro bug trackers because of that kernel log message under the assumption systemd was doing something wrong there and was crap. GRND_INSECURE was added to kernel 5.6, i.e. April 2020, which is not even 2y ago, so if new systemd would be used with kernels before that people will see the warning. As I know the Debian people they care about really old kernels and about the cosmetics of their kernel logs, so I figure they at least will not like this change.

It's especially the embedded world, as they tend to get stuck on ancient kernel versions, not their fault but due to bad hardware vendors - see all that noise we got due to the /proc hidepid mount option causing a warn-once on old kernels last year or so...

zx2c4:
Can't they just disable that warning if they need? Debian is already stuck on old systemd + old kernel. When they update systemd, they'll have a new kernel too. What x86 distros have old kernel + new systemd? All I can think of are situations with arm vendor kernels, and there's no rdrand there. It really seems like this is worth a shot, and that practically speaking, impact should be pretty minimal. And if people complain we can just revert.

zx2c4:
> It's especially the embedded world

There is no rdrand code path for those. Non-issue.

zx2c4:
This is now rebased on https://github.com/systemd/systemd/pull/22667, which should show a bit better the actual case where we wind up using RDRAND, and thus more precisely shows what we're getting rid of here.

As mentioned in my last message, I don't think we're getting rid of anything important in practice with this patch.

zx2c4:
Looking at the CI, it days CentOS 8S failed, but I can't see how that failure is related to this commit. I think it might be a "flake"? Or something else and maybe I should rebase on master to get the green checkbox?

Either way, I think this PR is good to go and ready to review/merge/whatever.

bluca:
> /cc @mbiebl

I don't think this is an issue for any Debian usage. There are no Debian x86 releases with ancient kernels but new systemd. At most you can have new systemd via backports with the kernel shipped in stable, but that's v5.10 already, so we are good. You'd need to have a user manually keeping oldstable x86 kernel on stable userspace, plus systemd from backports, but that seems excessively unlikely and it would be an unsupported setup as far as we are concerned.

These issues tend to come up in the embedded world, ie openembedded/yocto and similar. I imagine you were thinking about the procfs hidepid stuff, and that was from yocto/buildroot/etc. But as Jason noted this only affects x86, so although there are x86 builds for those, it would be a weird setup even in that context. It might happen of course, and perhaps 2 years is a bit on the short side of the scale.

Overall it seems to me this is something that we could risk getting in v252 (ie, second half of the year), with little risk of backlash. We can always revert and further postpone it if it happens.

zx2c4:
> These issues tend to come up in the embedded world, ie openembedded/yocto and similar. I imagine you were thinking about the procfs hidepid stuff, and that was from yocto/buildroot/etc. But as Jason noted this only affects x86, so although there are x86 builds for those, it would be a weird setup even in that context. It might happen of course, and perhaps 2 years is a bit on the short side of the scale.

It would be really weird, even in yocto x86 world. But something to keep in mind for the embedded use cases: those extreme edge cases doing weird things are in the best possible position for simply _turning off the kmsgs_ with the kernel command line, precisely _because_ they're embedded systems doing weird things. In other words, old x86 systems that are trying to forcibly upgrade to new systemd can just adjust the thing that RDRAND was there to compensate for. So that seems like another non-issue.

Anyway, just my 2¢; feel free to merge this whenever makes most sense to you. It's obviously not urgent, though it is a maintenance burden of sorts and who knows when you'll get bit by the next irksome AMD RDRAND bug or whatever. I've just rebased this on main, per the label.

poettering:
I am inclined to merge this then, unless anyone is loudly opposed?

poettering:
hmm, docs/RANDOM_SEEDS.md needs updating I guess. It talks a lot about RDSEED still...

src/basic/random-util.c
```diff
@@ -257,8 +103,6 @@ void initialize_srand(void) {
 #if HAVE_SYS_AUXV_H
         const void *auxv;
```
medhefgo, line 104: While this code is touched, I'd love to see this declaration be moved into the other `#if`.

zx2c4: There's a statement in between the two, and afaict, systemd declares variables before statements.

zx2c4:
> hmm, docs/RANDOM_SEEDS.md needs updating I guess. It talks a lot about RDSEED still...

Fixed.

src/basic/random-util.c
medhefgo: That restriction was removed some time ago.

zx2c4: Oh. Maybe send another PR for that? I don't see how it's related to this, and it'd be nice to have this stand-alone revertable in case things don't go as planned.

docs/RANDOM_SEEDS.md
```diff
@@ -297,9 +273,9 @@ This primarily leaves two kind of systems in the cold:
    boot. Alternatively, consider implementing a solution similar to
    systemd-boot's random seed concept in your platform's boot loader.
 
-2. Virtualized environments that lack both virtio-rng and RDRAND, outside of
-   test environments. Tough luck. Talk to your hosting provider, and ask them
-   to fix this.
+2. Virtualized environments that lack both virtio-rng and `GRND_INSECURE`,
```
bluca: Is this an actual thing? GRND_INSECURE being unavailable because of configuration/whatnot by the hypervisor provider?

zx2c4: Not quite. This may have just been a weird edit and maybe I should remove the bullet point all together? I was thinking that `virtio-rng` might initialize the rng pretty fast, so you don't even need GRND_INSECURE, while GRND_INSECURE fills the needs of RDRAND.

Do you want me to just remove the bullet point?

zx2c4: Actually, the original text is fine. I'll restore that chunk. RDRAND _is_ relevant here because it can seed the kernel RNG.

bluca: What about:

```
2. Virtualized environments that lack virtio-rng and a kernel without `GRND_INSECURE`,
   outside of test environments. Tough luck. Talk to your hosting provider, and ask them to
   fix this, or switch your VM to a newer kernel that provides `GRND_INSECURE` (v5.6
   and newer).
```

bluca: sure, that works too

zx2c4: That's what I was thinking at first too, but I just reread the whole section, and it's actually a discussion of bootstrapping the RNG to be secure, so I think the mention of GRND_INSECURE doesn't fit here, whilst the mention of RDRAND does, because:

```c
	if (arch_init && trust_cpu && !crng_ready()) {
		crng_init = 2;
		pr_notice("crng init done (trusting CPU's manufacturer)\n");
	}
```

zx2c4:
CI is all green now. Any thing else holding this up that I should address?

zx2c4:
Added a `NEWS` entry about this to mention the kmsg change.

NEWS
```diff
@@ -54,6 +54,20 @@ CHANGES WITH 251:
           handling, and improving compatibility with home directories intended
           to be portable like the ones managed by systemd-homed.
 
+	* All kernels supported by systemd mix RDRAND (or similar) into the
```
poettering: indentation is off

zx2c4: Tabs vs spaces again... Fixing.

zx2c4:
Fixed `\t` vs ` ` and rebased. Should be good to go now.

poettering:
excellent! thanks!

bluca (approved):

PR #22670, opened 2022-03-07
memory-id: Work-around incorrect "Number of slots"

hadess:
In some BIOSes, the "Number of slots or sockets available for Memory
Devices in this array" is incorrectly set to the number of memory array
that's populated.

Work-around this problem by outputting the number of sockets after
having parsed them so that consumers of this data can carry on expecting
an accurate number in this property.

See https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/1686

I'm still waiting on the dmi dump to test this though.


hadess:
Turns out the HP Z600 dmidecode dump we have shows the problem.

src/udev/dmi_memory_id/dmi_memory_id.c
```diff
@@ -563,6 +563,7 @@ static void dmi_table_decode(const uint8_t *buf, size_t len, uint16_t num) {
 
                 data = next;
         }
+        printf("MEMORY_ARRAY_NUM_DEVICES=%u\n", next_slot_num);
```
poettering: should this be suppressed if next_slot_num is zero? i think so, no?

src/udev/dmi_memory_id/dmi_memory_id.c
```diff
@@ -417,9 +417,10 @@ static void dmi_memory_device_size_detail(
                 dmi_print_memory_size("MEMORY_DEVICE", attr_suffix, slot_num, code, MEMORY_SIZE_UNIT_BYTES);
 }
 
+static unsigned next_slot_num = 0;
```
poettering: hmm, actually, we don't like global variables really. It's fine for stuff that is inherently process-global (i.e. the parsed out fields from the command line or the environment block or so), but for other stuff we generally try to avoid it. AFAICS it's pretty easy to get rid of this one, just make this a local variable in dmi_table_decode() and pass a pointer to dmi_decode().

It's mostly a matter of coding style/hygiene, so that we can copy stuff around in our tree/reuse it without too much work.

I am aware this was a global variable before, too, just one hidden in a function. But if this is touched, let's fix this properly, please

poettering:
lgtm, just some nitpicks

poettering:
excellent. thanks! lgtm

PR #22671, opened 2022-03-07
random-util: remove RANDOM_EXTEND_WITH_PSEUDO

zx2c4:
This depends on #22668. Please review #22668 first.

====

```
    There's never really a time where /dev/urandom is able to return some
    easily but more with difficulty. Once the RNG is initialized, bytes
    should just flow normally. Simplify the logic here by just getting rid
    of this bizarre mode.
    
    This drastically simplifies everything, so that there's now a secure
    mode (RANDOM_BLOCK) and a best-effort mode (0).
```


zx2c4:
I merged this into https://github.com/systemd/systemd/pull/22667

PR #22673, opened 2022-03-07
ci: make the concurrency group identifier unique once again

mrc0mmand:
Otherwise we end up randomly cancelling once of the two centos_epel
jobs.

Follow-up to da637c8fcad1baa530578982d41a6ecb1440f7c5.


DaanDeMeyer:
Yikes it's really easy to forget to update that

bluca (approved):

PR #22676, opened 2022-03-08
udev: 60-persistent-storage-tape.rules: handle duplicate device ID

mwilck:
Some SCSI tape devices use the same device ID (NAA registered device
designator) for the SCSI tape changer device and the first actual tape
device. For example, this one:

https://docs.oracle.com/en/storage/tape-storage/storagetek-sl150-modular-tape-library/slofs/bridged-tape-drives.html

   You must connect the bridged drive to an HBA supporting multiple
   LUNs (also referred to as LUN scanning). The SL150 Library uses a
   single SCSI ID and two logical unit numbers (LUN). LUN 0 controls
   the tape drive and LUN 1 which is configured as a SCSI medium
   changer device controls the robotics. Data is sent to the remaining
   LUN on the bridged drive or to LUNs on the other, unbridged drives
   in the partition, all of which are configured as SCSI
   sequential-access (tape) devices.

This may lead to errors because /dev/tape/by-id symlinks may sometimes
point to the st device representing the tape, and sometimes to the sg
device representing the changer.

Fix this by assigning an increased priority to the tape device, and creating
a separate -changer link for the SCSI tape changer.

Co-developed-by: Martin Wilck <mwilck@suse.com>


yuwata (approved):
LGTM.

PR #22679, opened 2022-03-08
main: log which process send SIGNAL to PID1

lclgo:
This can help users to figure out what makes systemd freeze.
1. Someone kills systemd accidentally, then the sender_pid won't be 1;
2. systemd triggers segfault or assert, then the sender_pid will be 1;

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -228,7 +228,7 @@ _noreturn_ static void freeze_or_exit_or_reboot(void) {
         freeze();
 }
 
-_noreturn_ static void crash(int sig) {
+_noreturn_ static void crash(int sig, siginfo_t *siginfo, void *context) {
         struct sigaction sa;
         pid_t pid;
 
@@ -272,6 +272,14 @@ _noreturn_ static void crash(int sig) {
                 } else {
                         siginfo_t status;
                         int r;
+
+                        if (siginfo) {
+                                _cleanup_free_ char *cmdline = NULL;
+                                pid_t sender_pid = siginfo->si_pid;
+
+                                (void) get_process_cmdline(sender_pid, SIZE_MAX, 0, &cmdline);
+                                log_emergency("Caught <%s> from PID "PID_FMT" (%s)", signal_to_string(sig), sender_pid, strna(cmdline));
+                        }
 
                         /* Order things nicely. */
                         r = wait_for_terminate(pid, &status);
@@ -330,8 +338,8 @@ _noreturn_ static void crash(int sig) {
 
 static void install_crash_handler(void) {
         static const struct sigaction sa = {
-                .sa_handler = crash,
-                .sa_flags = SA_NODEFER, /* So that we can raise the signal again from the signal handler */
+                .sa_sigaction = crash,
+                .sa_flags = SA_NODEFER | SA_SIGINFO, /* So that we can raise the signal again from the signal handler */
         };
         int r;
 


Geass-LL:
With this patch, if someone kills systemd, we have messages like:
systemd[1]:Caught <SEGV> from PID 6092 (-bash)
If systemd triggers freezing itself, we have the following message:
systemd[1]:Caught <SEGV> from PID 1 (/usr/lib/systemd/systemd --switched-root --system --deserialize 16)

src/core/main.c
```diff
@@ -271,8 +271,14 @@ _noreturn_ static void crash(int sig) {
                         _exit(EXIT_EXCEPTION);
                 } else {
                         siginfo_t status;
+                        _cleanup_free_ char *cmdline = NULL;
+                        pid_t sender_pid;
                         int r;
 
+                        sender_pid = siginfo->si_pid;
+                        (void) get_process_cmdline(sender_pid, SIZE_MAX, 0, &cmdline);
+                        log_emergency("Caught <%s> from PID %d (%s)", signal_to_string(sig), sender_pid, strna(cmdline));
```
yuwata: `"%d"` -> `PID_FMT`

src/core/main.c
```diff
@@ -271,8 +271,14 @@ _noreturn_ static void crash(int sig) {
                         _exit(EXIT_EXCEPTION);
                 } else {
                         siginfo_t status;
+                        _cleanup_free_ char *cmdline = NULL;
+                        pid_t sender_pid;
                         int r;
 
+                        sender_pid = siginfo->si_pid;
```
yuwata: I'd like to add `assert(siginfo);`.

yuwata (changes requested):
Looks nice. Two nitpicks.

src/core/main.c
```diff
@@ -228,10 +228,12 @@ _noreturn_ static void freeze_or_exit_or_reboot(void) {
         freeze();
 }
 
-_noreturn_ static void crash(int sig) {
+_noreturn_ static void crash(int sig, siginfo_t *siginfo, void *context) {
         struct sigaction sa;
         pid_t pid;
 
+        assert(siginfo);
```
bluca: is this guaranteed to be set? If not, maybe check instead and skip the new calls if it's NULL?

lclgo: [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html) only says:
```
When delivering a signal with a SA_SIGINFO handler, the kernel
does not always provide meaningful values for all of the fields
of the siginfo_t that are relevant for that signal.
```

bluca: Given it's informational-only, let's play it safe, and wrap the new code in `if (siginfo)` instead of the assert

yuwata: I checked the man page, but it does not guarantee that `siginfo` is always provided. It is safer to do as @bluca commented.

src/core/main.c
```diff
@@ -273,6 +273,14 @@ _noreturn_ static void crash(int sig) {
                         siginfo_t status;
                         int r;
 
+                        if (siginfo) {
+                                _cleanup_free_ char *cmdline = NULL;
+                                pid_t sender_pid;
+                                sender_pid = siginfo->si_pid;
```
bluca: this can be merged in the declaration above - also add an empty new line after the declaration block

src/core/main.c
lclgo: New version force pushed, PLAL. @bluca

src/core/main.c
yuwata: nit. We usually add one blank line after the variable declaration.

And you can merge two lines with `pid_t sender_pid = siginfo->si_pid;`

yuwata: haha. exactly the same comments at the same time :-)

bluca (approved):

yuwata (approved):

src/core/main.c
```diff
@@ -273,6 +273,14 @@ _noreturn_ static void crash(int sig) {
                         siginfo_t status;
                         int r;
 
+                        if (siginfo) {
+                                _cleanup_free_ char *cmdline = NULL;
+                                pid_t sender_pid = siginfo->si_pid;
+
+                                (void) get_process_cmdline(sender_pid, SIZE_MAX, 0, &cmdline);
```
poettering, line 280: this calls malloc internally. We so far avoided doing that from the crash handler, since there's a good chance it is called due to some memory corruption issue, where we shouldn't do more of it, since it likely will fail again, even more catastrophically.

I mean, as a general rule, memory allocations from signal handlers are *not* OK.

hence it might make sense to show the PID of the process that killed us, but I'd avoid anything with reading the cmdline, here, it's not really worth the effrot.

i.e. drop the cmdline output, leave the PID part in.

bluca: I thought about the malloc thing, but given it is used best-effort (failure simply means it shows `n/a`), I thought it would be fine.

src/core/main.c
```diff
@@ -273,6 +273,14 @@ _noreturn_ static void crash(int sig) {
...
+                                log_emergency("Caught <%s> from PID "PID_FMT" (%s)", signal_to_string(sig), sender_pid, strna(cmdline));
+                        }
```
poettering, line 282: note that the sender_pid can also be zero, for example if we are killed from another process outside of our own PID namespace. here it would be displayed as if we killed outselves, since get_process_cmldine()  reads our own cmdline if pid is passed as zero. Thus we should detect this case, and generate a different message instead.

Hence maybe:

```c
if (siginfo->si_pid == 0)
        log_emergency("Caught <%s> from unknown sender process …
else if (siginfo->si_pid == 1)
        log_emergency("Caught <%s> from our own process …
else
        log_emergency("Caught <%s> from PID " PID_FMT …
```

you get the idea.

src/core/main.c
poettering: well, if glibc sees a memory corruption while allocating memory it will hit an assert. We should not risk that in our crash handler, which likely is called when memory corruption already took place.

malloc() is generally not safe to be called from signal handlers, but particularly not in case that signal handler is for SIGSEGV and friends.

poettering:
I filed #22690 now, since I am sure this should be addressed before the upcoming release.

src/core/main.c
bluca: ok, sounds good to me

poettering:
Maybe a lesson to learn here is that the crash handler function should actually come with a big comment explaining the situation and why malloc() or any function that calls it are not OK to call.

bluca:
> Maybe a lesson to learn here is that the crash handler function should actually come with a big comment explaining the situation and why malloc() or any function that calls it are not OK to call.

Yup, that sounds like an excellent idea, it's hard to remember all the rules

PR #22682, opened 2022-03-08
core: do not return 'skipped' when Condition*= fail with StartUnitWithFlags()

bluca:
Backward incompatible change to avoid returning 'skipped' if a condition causes
a job activation to be skipped when using StartUnitWithFlags().
Job results are broadcasted, so it is theoretically possible that existing
software could get confused if they see this result.

Replaces https://github.com/systemd/systemd/pull/22369

Note that instead of a full revert, I'm keeping the D-Bus method in place. We might need it in the future for the next iteration of this feature as discussed, so no need to do a full compat break.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,10 @@ CHANGES WITH 251:
           can still be used by explicitly configuring RouteTable= setting in
           .netdev files.
 
+        * Jobs started via StartUnitWithFlags() will no longer return 'skipped'
+          when a Condition*= check does not succeed, restoring the JobRemoved
+          signal to the behaviour it had before v250.
+
         * The org.freedesktop.portable1 methods GetMetadataWithExtensions and
           GetImageMetadataWithExtensions have been fixed to provide an extra return
           parameter, containing the actual extensions release metadata. The


src/core/dbus-unit.c
```diff
@@ -417,10 +417,6 @@ int bus_unit_method_start_generic(
                         return sd_bus_reply_method_errorf(message, SD_BUS_ERROR_INVALID_ARGS,
                                                           "Invalid 'flags' parameter '%" PRIu64 "'",
                                                           input_flags);
-
-                /* The new method unconditionally uses the new behaviour of returning 'skip' when
-                 * a job is skipped. */
-                job_flags |= BUS_UNIT_QUEUE_RETURN_SKIP_ON_CONDITION_FAIL;
```
yuwata, line 423: The flag `BUS_UNIT_QUEUE_RETURN_SKIP_ON_CONDITION_FAIL` is not used now. Maybe it is better to drop it from the enum?

bluca: As discussed yesterday I'm going to rework this shortly, so I'd prefer to keep churn to a minimum as this will be needed again for that

yuwata: I see, OK.

NEWS
```diff
@@ -9,6 +9,9 @@ CHANGES WITH 251:
           can still be used by explicitly configuring RouteTable= setting in
           .netdev files.
 
+        * Jobs started via StartUnitWithFlags() will no longer return 'skipped'
+          when a Condition*= check does not succeed.
```
yuwata: Not a matter in this PR, but maybe we should create a new section about the compat breakages.

bluca: yes, when we adjust the NEWS before release we should definitely do that

yuwata (approved):
LGTM. But let's wait for @poettering's final review.

NEWS
poettering: please make clear that this reverts to the status quo ante, i.e. unconditionally reverts to semantics JobRemoved had in 249 and before.

poettering:
Could you add a comment next to definition of `BUS_UNIT_QUEUE_RETURN_SKIP_ON_CONDITION_FAIL` saying `FIXME: currently not used, will be changed soon` or so?

NEWS
bluca: done

mrc0mmand:
> @mrc0mmand this is a known one right?
> 
> ```
> 15:05:12 FAILED TASKS:
> 15:05:12 -------------
> 15:05:12 TEST-53-ISSUE-16347
> ```

That's a new one. I'll restart the job just for good measure.

bluca:
I think I've seen it barf on ubuntu. Previous run passed, and the push has only changed NEWS and a comment, so it should be fine.

PR #22685, opened 2022-03-09
core: support MountAPIVFS and RootDirectory in user manager

bluca:
The only piece missing was to somehow make /proc appear in the
new user+mount namespace. It is not possible to mount a new
/proc instance, not even with hidepid=invisible,subset=pid, in
a user namespace unless a PID namespace is created too (and also
at the same time as the other namespaces, it is not possible to
mount a new /proc in a child process that creates a PID namespace
forked from a parent that created a user+mount namespace, it has
to happen at the same time).

Use the host's /proc with a bind-mount as a fallback for this
case. User session services would already run with it, so
nothing is lost.

### DIFF
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -143,9 +143,7 @@
           <title>Mounting logging sockets into root environment</title>
 
           <programlisting>BindReadOnlyPaths=/dev/log /run/systemd/journal/socket /run/systemd/journal/stdout</programlisting>
-        </example>
-
-        <xi:include href="system-only.xml" xpointer="singular"/></listitem>
+        </example></listitem>
       </varlistentry>
 
       <varlistentry>
@@ -276,9 +274,7 @@
 
         <para>In order to allow propagating mounts at runtime in a safe manner, <filename>/run/systemd/propagate</filename>
         on the host will be used to set up new mounts, and <filename>/run/host/incoming/</filename> in the private namespace
-        will be used as an intermediate step to store them before being moved to the final mount point.</para>
-
-        <xi:include href="system-only.xml" xpointer="singular"/></listitem>
+        will be used as an intermediate step to store them before being moved to the final mount point.</para></listitem>
       </varlistentry>
 
       <varlistentry>
@@ -364,9 +360,7 @@
         <varname>InaccessiblePaths=</varname>, or under <filename>/home/</filename> and other protected
         directories if <varname>ProtectHome=yes</varname> is
         specified. <varname>TemporaryFileSystem=</varname> with <literal>:ro</literal> or
-        <varname>ProtectHome=tmpfs</varname> should be used instead.</para>
-
-        <xi:include href="system-only.xml" xpointer="singular"/></listitem>
+        <varname>ProtectHome=tmpfs</varname> should be used instead.</para></listitem>
       </varlistentry>
 
       <varlistentry>
--- a/src/core/namespace.c
+++ b/src/core/namespace.c
@@ -1128,9 +1128,15 @@ static int mount_procfs(const MountEntry *m, const NamespaceInfo *ns_info) {
                 r = path_is_mount_point(entry_path, NULL, 0);
                 if (r < 0)
                         return log_debug_errno(r, "Unable to determine whether /proc is already mounted: %m");
-                if (r == 0)
-                        /* /proc is not mounted. Propagate the original error code. */
-                        return -EPERM;
+                if (r == 0) {
+                        /* We lack permissions to mount a new instance of /proc, and it is not already
+                         * mounted. But we can access the host's, so as a final fallback bind-mount it to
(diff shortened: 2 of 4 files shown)


man/systemd.exec.xml
```diff
@@ -144,8 +144,6 @@
 
           <programlisting>BindReadOnlyPaths=/dev/log /run/systemd/journal/socket /run/systemd/journal/stdout</programlisting>
         </example>
-
-        <xi:include href="system-only.xml" xpointer="singular"/></listitem>
```
yuwata, line 148: Please keep `</listitem>`.

bluca: whoops, deleted too much

bluca:
(first commit is a bug fix and should be backported)

brauner (approved):

src/core/namespace.c
```diff
@@ -1128,9 +1128,23 @@ static int mount_procfs(const MountEntry *m, const NamespaceInfo *ns_info) {
...
-                        /* /proc is not mounted. Propagate the original error code. */
-                        return -EPERM;
+                if (r == 0) {
+                        r = path_is_mount_point("/proc", NULL, 0);
+                        if (r < 0)
+                                return log_debug_errno(r, "Unable to determine whether /proc is already mounted: %m");
+                        if (r == 0)
+                                /* /proc is not mounted, neither at the target nor on the host. Propagate
+                                 * the original error code. */
+                                return -EPERM;
```
poettering: hmm, why this test? /proc/ is definitely going to be mounted on the host, we generally don't need to check that, do we?

Is there are reason for adding this, or just extra safety? I'd really drop this check.

src/core/namespace.c
```diff
@@ -2446,6 +2460,17 @@ int setup_namespace(
 
         /* MS_MOVE does not work on MS_SHARED so the remount MS_SHARED will be done later */
         r = mount_move_root(root);
+        if (r == -EINVAL && root_directory) {
+                /* If we are using root_directory and we don't have privileges (ie: user manager in a user
+                 * namespace) and the root_directory is already a mount point in the parent namespace,
+                 * MS_MOVE will fail as we don't have permission to change it (of course with EINVAL rather
+                 * than EPERM). Attempt to bind-mount it over itself (like we do above if it's not already
```
poettering: "of course"? is that mean sarcastically? because if not i can't see why it would be so "of course"? ;-)

src/core/namespace.c
bluca: Just an extra check, so that if access is denied in some way and the check fails we get a (hopefully) more helpful log_debug above. But I can remove it if that's the preference.

src/core/namespace.c
bluca: very much sarcastically :-)

src/core/namespace.c
poettering: i'd drop it, yeah. just seems unlikely to be hit, at least i don't see how it could ever happen

poettering:
looks good to me. see my comments above.

poettering:
hmm, there might be an issue somewhere you could close with this.

src/core/namespace.c
poettering: maybe add an emoji, to make this clear. 🤯 or so

might not be obvious otherwise.

bluca: ok, I've just removed it to avoid confusion

bluca:
> hmm, there might be an issue somewhere you could close with this.

I did a quick search with some obvious keywords, but couldn't find anything related

test/units/testsuite-43.sh
```diff
@@ -68,6 +68,24 @@ runas testuser systemd-run --wait --user --unit=test-group-fail \
...
+runas testuser systemd-run --wait --user --unit=test-root-dir \
+    -p PrivateUsers=yes -p RootDirectory=/tmp/img \
+    grep MARKER=1 /etc/os-release
+
+mkdir /tmp/img_bind
+mount --bind /tmp/img /tmp/img_bind
+runas testuser systemd-run --wait --user --unit=test-root-dir-bind \
+    -p PrivateUsers=yes -p RootDirectory=/tmp/img_bind \
+    grep MARKER=1 /etc/os-release
+umount /tmp/img_bind
```
yuwata, line 87: Maybe it is better to add an explicit test for MountAPIVFS=?

bluca: It's always implied when using these options - I have more enablement PRs for the user manager sandboxing coming up, I can add this in the next PR

yuwata: OK. Then, that's fine.

src/core/namespace.c
poettering: you missed your chance to add another emoji to the codebase. Sad!

bluca: Surely you meant: "Sad! 😢"

poettering: good point 👉!

PR #22691, opened 2022-03-09
journald: reconsider flushing the data when being restarted

fbuihuu:
b4e26d1d8e582d78e67ed766177f10e8e194404c changed the meaning of
/run/systemd/journal/flushed flag file.

Before this commit /run/systemd/journal/flushed was created during the first
attempt to flush the journal (triggered by systemd-journal.flush.service) and
this regardless of whether the data were actually moved to /var/log/journal or
not. Hence at runtime, the flushed flag prevented subsequent calls to
`journalctl --flush` to be useful although it also indicated journald to switch
to persistent mode (if configured) when the service was restarted (either via
'restart' or via 'stop' + 'start').

After this commit the file flag was created only if the data were actually
moved to the persistent location. Hence attempts to flush the data via
`journalctl --flush` can be used at runtime (even if the first attempt made by
s-j-f.service failed) but restarting journald to switch from volatile to
persistent mode doesn't work anymore (since journald checks the presence of the
flag to determine if it should consider starting in persistent mode).

Hence in the following use case:

 1. boot with /var/log/journal missing and with Storage=auto (hence in volatile mode)
 2. stop
 2. create /var/log/journal
 3. start

journald is restarted in volatile mode whereas it should have switched to
persistent mode.

This patch restores the old meaning of the flushed flag and make the use case
described previously work again.

### DIFF
--- a/src/journal/journald-server.c
+++ b/src/journal/journald-server.c
@@ -1123,10 +1123,9 @@ void server_dispatch_message(
 
 int server_flush_to_var(Server *s, bool require_flag_file) {
         sd_journal *j = NULL;
-        const char *fn;
         unsigned n = 0;
         usec_t start;
-        int r, k;
+        int r;
 
         assert(s);
 
@@ -1219,11 +1218,6 @@ finish:
                                           FORMAT_TIMESPAN(usec_sub_unsigned(now(CLOCK_MONOTONIC), start), 0),
                                           n),
                               NULL);
-
-        fn = strjoina(s->runtime_directory, "/flushed");
-        k = touch(fn);
-        if (k < 0)
-                log_warning_errno(k, "Failed to touch %s, ignoring: %m", fn);
 
         server_refresh_idle_timer(s);
         return r;
@@ -1387,6 +1381,9 @@ int server_process_datagram(
 }
 
 static void server_full_flush(Server *s) {
+        const char *fn;
+        int r;
+
         assert(s);
 
         (void) server_flush_to_var(s, false);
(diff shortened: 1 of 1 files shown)


fbuihuu:
@poettering I think both behaviors are moot. Wouldn't it be simpler if journald considers itself whether it should switch to persistent mode or not instead of relying on a (successful) call to `--flush` ?

I mean journald doesn't need to know whether a flush has already happened or not (since if it's started in volatile mode it won't be the case anyway). The only information it needs is when it can start accessing /var/log/journal and  `systemd-journal-flush.service` has this info.

But the info is  translated to "has a succcessful flush already happened", which is not what journald wants to know when it's (re)started.

WDYT ?

poettering:
I am not convinced this change is desirable. Why not just document that users have to call `journalctl --relinquish-var` + `journalctl --flush` to do the switch forth and back after changing configuration?

fbuihuu:
Because reloading the new settings by restarting journald used to work before b4e26d1d8e582d78e67ed766177f10e8e194404c and I suppose that this use case is widespread enough to not break it.

fbuihuu:
It really seems that `/run/systemd/journal/flushed` was useful only when triggering a flush was only achieved by sending a signal. But now that journald switched to varlink API it doesn't seem necessary anymore (unless it's part of the API and other apps rely on it).

Again the only information journald needs now is "is /var available to enable persistent logging ?".

This info could be added in `systemd-journal-flush.service` by simply adding `ExecStartPre=touch /run/systemd/journal/can_flush`. The flag would be removed when the service is stopped.

Hence journald would simply need to check the presence of the flag each time it's started.

`journalctl --flush` would always contact journald, regardless of whether the flag is set, and would let journald figure it out itself.

We could still keep  /run/systemd/journal/flushed` in case an application triggers a flush by sending SIGUSR1 and relies on the flag to determine when the flush operation is completed. But I doubt since no one reported that the flag meaning changed after b4e26d1d8e582d78e67ed766177f10e8e194404c.

DaanDeMeyer:
@fbuihuu This needs a rebase

fbuihuu:
@DaanDeMeyer hmm this one is pretty old and I actually forgot about it :(

But I think it's been replaced by https://github.com/systemd/systemd/pull/25047

Issue #22690, opened 2022-03-09
get_process_cmdline() (i.e. malloc()) should not be called from crash handler


poettering:
#22679 added a call to get_process_cmdline() to the crash handler of PID 1. That should be removed, see my comments on that issue. 

This is just a tracker bug, because this should not be forgotten before the upcoming release.

PR #22693, opened 2022-03-10
main: drop get_process_cmdline from crash handler

lclgo:
get_process_cmdline calls malloc, which should be avoid in signal handler.

Fixes: #22690

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -274,11 +274,12 @@ _noreturn_ static void crash(int sig, siginfo_t *siginfo, void *context) {
                         int r;
 
                         if (siginfo) {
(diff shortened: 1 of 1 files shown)


yuwata (approved):

Geass-LL:
Sorry for the terrible bug introduced, I have updated the code as @poettering`s suggestion. PTAL.

poettering:
> Sorry for the terrible bug introduced, I have updated the code as @poettering`s suggestion. PTAL.

It's not that "terrible", and it passed through two reviews, so it's not at all your fault, but ours. ;-)

yuwata:
Yeah, that's not your fault, but ours. Please do not hesitate. And thank you for the quick update.

PR #22695, opened 2022-03-10
pid1: add comment clarifying not to use memory allocation from crash handler

poettering:
Also, a refactoring commit that splits out the PID 1 crash handler into its own .c file.

The comment is inspired by #22693 aka #22690 aka #22679.

### DIFF
--- a/src/core/meson.build
+++ b/src/core/meson.build
@@ -197,7 +197,12 @@ libcore = shared_library(
 
 core_includes = [includes, include_directories('.')]
 
-systemd_sources = files('main.c')
+systemd_sources = files(
+        'main.c',
+        'main.h',
+        'crash-handler.c',
+        'crash-handler.h',
+)
 
 in_files = [['system.conf',                     pkgsysconfdir],
             ['user.conf',                       pkgsysconfdir],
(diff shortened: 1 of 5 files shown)


DaanDeMeyer (approved):
One small nit

yuwata:
> One small nit

@DaanDeMeyer I cannot find the 'small nit'...

src/core/meson.build
```diff
@@ -197,7 +197,11 @@ libcore = shared_library(
 
 core_includes = [includes, include_directories('.')]
 
-systemd_sources = files('main.c')
+systemd_sources = files(
+        'main.c',
+        'main.h',
+        'crash-handler.c',
+        'crash-handler.h')
```
daandemeyer: In a recent meson PR by @keszybz we bikeshedded to put the paren on a newline for calls to files() so we can use a  trailing comma which makes future diffs smaller so let's do that here as well.

yuwata (approved):
LGTM.

poettering:
made the change @DaanDeMeyer suggested.

Also updated the comment this PR adds a tiny bit. It sorely needed a couple of bomb emojis, so I added them now. Sorry for missing them earlier. (I also added a pointer to the signal-safety(7) man page, but that part is mostly for decorative purposes.)

poettering:
Now that the bomb emojis are in place I took the liberty to upgrade the green label.

poettering:
💣 💣 💣

PR #22701, opened 2022-03-10
pid1/nspawn: raise RLIMIT_MEMLOCK to 8M matching kernel 5.16's new default

poettering:


yuwata (approved):
LGTM.

Geass-LL:
Maybe also change /etc/systemd/system.conf:
#DefaultLimitMEMLOCK=8M

poettering:
good point. added. also updated the man page, to mention this.

bluca:
Uhm the last commit is the same as https://github.com/systemd/systemd/pull/22707 was that an unintentional push?

bluca:
> Uhm the last commit is the same as #22707 was that an unintentional push?

I've removed the commit given the other PR is merged. There's no code change, so no need to wait for another round of CI.

poettering:
> Uhm the last commit is the same as #22707 was that an unintentional push?

oops, yes, seems i accidently pulled that into the wrong branch. sorry for the confusion

PR #22702, opened 2022-03-10
core: support ExtensionDirectories in user manager

bluca:
Unprivileged overlayfs is supported since Linux 5.11. The only
change needed to get ExtensionDirectories to work is to avoid
hard-coding the staging directory to the system manager runtime
directory, everything else just works (TM).

### DIFF
--- a/src/test/test-namespace.c
+++ b/src/test/test-namespace.c
@@ -208,6 +208,7 @@ TEST(protect_kernel_logs) {
                                     NULL,
                                     NULL,
                                     NULL,
+                                    NULL,
                                     NULL);
                 assert_se(r == 0);
 
--- a/src/test/test-ns.c
+++ b/src/test/test-ns.c
@@ -109,6 +109,7 @@ int main(int argc, char *argv[]) {
                             NULL,
                             NULL,
                             NULL,
+                            NULL,
                             NULL);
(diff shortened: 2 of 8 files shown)


src/test/test-namespace.c
```diff
@@ -208,6 +208,7 @@ TEST(protect_kernel_logs) {
                                     NULL,
                                     NULL,
                                     NULL,
```
yuwata, line 210: An assertion to extension directory is added. So this triggers the assertion, no?

src/test/test-ns.c
```diff
@@ -109,6 +109,7 @@ int main(int argc, char *argv[]) {
                             NULL,
                             NULL,
                             NULL,
+                            NULL,
```
yuwata, line 112: Same here.

src/test/test-namespace.c
bluca: oops, yes indeed - removed the assert, it's redundant anyway, there are other asserts before it is actually used (which won't happen in those tests as the options triggering its use are unset)

bluca: and of course I botched the check, as `append_extensions()` is always called (with empty args) - moved the assert in there below the check for the optional args

Issue #20492, opened 2021-08-20
systemd-cgls --user has no effect, --unit user@1000 fails


keszybz:
I guess that `--user` is treated as a shortcut `--user-units` with an empty list.

```console
$ systemd-cgls --user unit@1000
Failed to query unit control group path: Invalid argument
Failed to list cgroup tree: Invalid argument
```

We should do the same transformations that systemctl does.

PR #22703, opened 2022-03-10
cgls: mangle user-provided unit names

mrc0mmand:
so the CLI interface is now similar to `systemctl`, i.e. if no unit name
suffix is provided, assume `.service`.

Fixes: #20492

Before:
```
$ systemd-cgls --unit user@1000
Failed to query unit control group path: Invalid argument
Failed to list cgroup tree: Invalid argument
```

After:
```
$ build/systemd-cgls --unit user@1000
Unit user@1000.service (/user.slice/user-1000.slice/user@1000.service):
├─session.slice (#4939)
│ ├─pipewire-pulse.service (#5203)
│ │ └─7711 /usr/bin/pipewire-pulse
...
```

---

The "second part" of #20492 makes me wonder why do we even accept `--unit/--user-unit` without an argument in the first place, as the options are effective _only_ with an argument specified (i.e. `arg_show_unit` is used only in the block guarded by `if (!arg_names)`.

### DIFF
--- a/src/cgls/cgls.c
+++ b/src/cgls/cgls.c
@@ -208,7 +208,11 @@ static int run(int argc, char *argv[]) {
 
                         if (arg_show_unit != SHOW_UNIT_NONE) {
                                 /* Command line arguments are unit names */
-                                _cleanup_free_ char *cgroup = NULL;
+                                _cleanup_free_ char *cgroup = NULL, *unit_name = NULL;
+
+                                r = unit_name_mangle(*name, UNIT_NAME_MANGLE_WARN, &unit_name);
+                                if (r < 0)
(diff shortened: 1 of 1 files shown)


src/cgls/cgls.c
```diff
@@ -209,6 +209,11 @@ static int run(int argc, char *argv[]) {
                         if (arg_show_unit != SHOW_UNIT_NONE) {
                                 /* Command line arguments are unit names */
                                 _cleanup_free_ char *cgroup = NULL;
+                                _cleanup_free_ char *unit_name = NULL;
```
bluca: merge above?

bluca (approved):

poettering:
lgtm, too

Issue #20054, opened 2021-06-28
Pretty hostname specifier


petrkutalek:
**Is your feature request related to a problem? Please describe.**
It is not related to any problem, just nice-to-have feature request.

**Describe the solution you'd like**
Systemd offers [`%H` and `%l` specifiers](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers) for hostname and short hostname. It would be useful to also offer a specifier for the `Pretty hostname` (like the entry in `hostnamectl`). The rationale is to offer a nice host name to systemd service units (e.g. `ExecStart=/usr/bin/player --name "Audioplayer @ %█"` could pass "Audioplayer @ Petr's Laptop").

**The systemd version you checked that didn't have the feature you are asking for**
`systemd 248 (248.3-2-arch)`

poettering:
makes sense i guess

PR #22705, opened 2022-03-10
core: introduce %R specifier for pretty hostname

mrc0mmand:
Resolves: #20054

### DIFF
--- a/man/standard-specifiers.xml
+++ b/man/standard-specifiers.xml
@@ -34,6 +34,11 @@
     <entry>Short host name</entry>
     <entry>The hostname of the running system, truncated at the first dot to remove any domain component.</entry>
   </row>
+  <row id='R'>
+    <entry><literal>%R</literal></entry>
+    <entry>Pretty host name</entry>
+    <entry>The pretty hostname of the running system, as read from the <varname>PRETTY_HOSTNAME=</varname> field of <filename>/etc/machine-info</filename>. If not set, resolves to the short hostname. See <citerefentry><refentrytitle>machine-info</refentrytitle><manvolnum>5</manvolnum></citerefentry> for more information.</entry>
+  </row>
   <row id='m'>
     <entry><literal>%m</literal></entry>
     <entry>Machine ID</entry>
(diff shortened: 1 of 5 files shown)


src/shared/specifier.c
```diff
@@ -208,6 +208,21 @@ int specifier_short_host_name(char specifier, const void *data, const char *root
         return 0;
 }
 
+int specifier_pretty_host_name(char specifier, const void *data, const char *root, const void *userdata, char **ret) {
+        char *n = NULL;
+        int r;
+
```
bluca, line 228: `assert(ret)`

mrc0mmand: Given that almost none of the specifier helper functions assert on `ret`, it's been either left out intentionally or should be added everywhere :-)

bluca (approved):

src/shared/specifier.c
```diff
@@ -208,6 +208,23 @@ int specifier_short_host_name(char specifier, const void *data, const char *root
...
+        assert(ret);
+
+        r = get_pretty_hostname(&n);
+        if (r < 0) {
+                n = strdup("(none)");
```
yuwata: Not sure, but if the pretty hostname is not set, then it is better to fallback the short hostname??

Also, you can drop the assignment to `r`.

mrc0mmand: That definitely sounds more sensible, updated, thanks!

man/standard-specifiers.xml
```diff
@@ -34,6 +34,11 @@
     <entry>Short host name</entry>
     <entry>The hostname of the running system, truncated at the first dot to remove any domain component.</entry>
   </row>
+  <row id='R'>
+    <entry><literal>%R</literal></entry>
+    <entry>Pretty host name</entry>
+    <entry>The pretty hostname of the running system. If not set, resolves to the short hostname.</entry>
```
poettering: maybe add a comment that this is PRETTY_HOSTNAME from machine-info(5).

mrc0mmand: Makes sense, updated.

PR #22707, opened 2022-03-10
hostname-util: normalize get_pretty_hostname() call semantics

poettering:
get_pretty_hostname() so far had semantics not in line with our usual
ones: the return parameter was actually freed before the return string
written into it, because that's what parse_env_file() does. Moreover,
when the value was not set it would return NULL but succeed.

Let's normalize this, and only fill in the return value if there's
something set, and never read from it, like we usually do with return
parameter, and in particular those named "ret_xyz".

The existing callers don't really care about the differences, but it's
nicer to normalize behaviour to minimize surprises.

### DIFF
--- a/src/basic/hostname-util.c
+++ b/src/basic/hostname-util.c
@@ -8,6 +8,7 @@
 #include <unistd.h>
 
 #include "alloc-util.h"
+#include "env-file.h"
 #include "hostname-util.h"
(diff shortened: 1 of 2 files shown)


poettering:
Noticed while reviewing #22705

bluca (approved):

PR #22711, opened 2022-03-11
move to C11 (gnu11) and use u8 specifier for special characters

yuwata:


mrc0mmand:
Hmpf, it's weird the warnings surfaced after switching to `-std=gnu11`. I'll look into it...

poettering:
lgtm

mrc0mmand:
Interesting, with this patch:

```diff
diff --git a/src/fundamental/macro-fundamental.h b/src/fundamental/macro-fundamental.h
index 15b93165dc..e15b5c30f1 100644
--- a/src/fundamental/macro-fundamental.h
+++ b/src/fundamental/macro-fundamental.h
@@ -27,12 +27,8 @@
 /* Define C11 noreturn without <stdnoreturn.h> and even on older gcc
  * compiler versions */
 #ifndef _noreturn_
-#if __STDC_VERSION__ >= 201112L
-#define _noreturn_ _Noreturn
-#else
 #define _noreturn_ __attribute__((__noreturn__))
 #endif
-#endif
 
 #define XSTRINGIFY(x) #x
 #define STRINGIFY(x) XSTRINGIFY(x)
```

the CodeQL warnings disappear. Maybe the black-magic ifdef-ery needs to be adjusted?

yuwata:
Interesting. I noticed the block, but I thought it does not intoduce anything bad. Will check it.

medhefgo:
Well, if you do that you can turn the array in `test-bcd.c:get_bcd_title` into a `U""` (UTF-16) string. Also, `test-uft8.c` has a ton of escaping that could probably benefit too?

medhefgo:
Just out of curiosity, why were escape literals even used in the first place? Shouldn't regular strings (with emojis) work already anyways as the compiler creates utf8 strings on linux? (And for the same reason, I fail to see the need to make them explicit u8 strings, not that there is anything wrong with that.)

mrc0mmand:
> Interesting. I noticed the block, but I thought it does not intoduce anything bad. Will check it.

Ok, nevermind, this appears to be a bug in CodeQL. I'll file this to the respective upstream. For now, feel free to ignore the warnings.

CodeQL apparently checks only for the `noreturn` __attribute__ when determining if the function returns or not, but it doesn't take into account `_Noreturn` and `noreturn` macros from C11 (https://github.com/mrc0mmand/codeql-test/pull/1).

yuwata:
@poettering likes this, and there is no objection about this PR. Let's merge this as is, and address other stuff prompted by @medhefgo in later PR.

PR #22712, opened 2022-03-11
fix signature of sd_journal_print_with_location in docs

hellow554:
This is probably wrong since c11edfee986ecef0cacd803e65b5a26373db87af, so as long as the docs exist

The docs ([[1]] [[2]] [[3]]) are saying, that the signature of `sd_journal_print_with_location` is:

```c
int sd_journal_print_with_location( const char *file,
    const char *line,
    const char *func,
    int priority,
    const char *format,
    ...);
```

but [the code](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/systemd/sd-journal.h#L42) shows this signature:

```c
int sd_journal_print_with_location(
    int priority,
    const char *file,
    const char *line,
    const char *func,
    const char *format, ...) _sd_printf_(5, 6);
```

Note the position of priority.

I think the code is more right than the docs? 

[1]: https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/man/sd_journal_print.xml
[2]: https://www.freedesktop.org/software/systemd/man/sd_journal_print.html
[3]: https://man7.org/linux/man-pages/man3/sd_journal_send.3.html

### DIFF
--- a/man/sd_journal_print.xml
+++ b/man/sd_journal_print.xml
@@ -68,10 +68,10 @@
 
       <funcprototype>
         <funcdef>int <function>sd_journal_print_with_location</function></funcdef>
+        <paramdef>int <parameter>priority</parameter></paramdef>
         <paramdef>const char *<parameter>file</parameter></paramdef>
         <paramdef>const char *<parameter>line</parameter></paramdef>
         <paramdef>const char *<parameter>func</parameter></paramdef>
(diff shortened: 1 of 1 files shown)


yuwata (approved):
Thanks. LGTM.

poettering:
lgtm, too

PR #22717, opened 2022-03-12
udev: do not skip events when device is already locked

yuwata:


yuwata:
Now TEST-46-HOMED becomes quite unstable (but it was already unstable):
```
systemd-homework[1081]: Image file '/home/test-user.home' already locked, can't use.
```
But this is a lock for regular file, and should not be (at least directly) related to this PR. I guess there is a race or fd leak in homed or homework.

@poettering Could you take a look https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/10518/artifact/systemd-centos-ci/artifacts_lbwk7s93/vagrant-logs.Pqn/vagrant-arch-testsuite.Orc/TEST-46-HOMED/system.journal ?

yuwata:
Added two commits:
- cleanup commit for homework, though this does not address the instability of the homed test,
- workaround for [the last TEST-58-REPART failure](https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/ppc64el/s/systemd-upstream/20220312_232751_edfc1@/log.gz) on ppc64el.

This is now ready for review. PTAL.

yuwata:
I noticed https://systemd.io/BLOCK_DEVICE_LOCKING needs to be updated. If this is approved, then I will do.

yuwata:
@mrc0mmand Could you run TEST-64 repeatedly with this PR on an affected CI env?

mrc0mmand:
> @mrc0mmand Could you run TEST-64 repeatedly with this PR on an affected CI env?

Sure, unleashed https://github.com/systemd/systemd-centos-ci/pull/474 on this PR.

yuwata:
Hm, not only homed test, but the repart test becomes also unstable...

yuwata:
@mrc0mmand Thanks. Could you re-trigger the CI in https://github.com/systemd/systemd-centos-ci/pull/474? I hope [the last failure](https://jenkins-systemd.apps.ocp.ci.centos.org/job/ci-build/932/artifact//index.html) is solved or at least mitigated.

src/home/homework-luks.c
```diff
@@ -2311,9 +2315,11 @@ int home_create_luks(
                 return log_error_errno(r, "Failed to set up loopback device for %s: %m", setup->temporary_image_path);
         }
 
-        r = loop_device_flock(setup->loop, LOCK_EX); /* make sure udev won't read before we are done */
-        if (r < 0)
-                return log_error_errno(r, "Failed to take lock on loop device: %m");
+        if (path_startswith(ip, "/dev/")) {
+                r = loop_device_flock(setup->loop, LOCK_EX); /* make sure udev won't read before we are done */
```
yuwata, line 2319: Hm, is it necessary? `image_fd` is already locked in the above...

yuwata:
I noticed sd-makefs and sd-growfs do not flock block device...

src/udev/udevd.c
```diff
@@ -406,8 +406,10 @@ static int worker_lock_block_device(sd_device *dev, int *ret_fd) {
 
         fd = open(val, O_RDONLY|O_CLOEXEC|O_NOFOLLOW|O_NONBLOCK);
         if (fd < 0) {
-                log_device_debug_errno(dev, errno, "Failed to open '%s', ignoring: %m", val);
-                return 0;
+                bool ignore = errno == ENOENT;
```
poettering: wouldn't it make sense to treat ENXIO/ENODEV the same? i.e. removal of a device node is async, no? i.e. there might be time window where a device node still exists but the major/minor refer to a device that cannot be opened anymore?

src/udev/udevd.c
```diff
@@ -983,7 +983,7 @@ static int event_queue_start(Manager *manager) {
                                                  event->seqnum,
                                                  strna(device_action_to_string(a)));
 
-                        event_free(event);
+                        event_enter_failed(event);
```
poettering: hmm, under which conditions will event_is_blocked() actually fail? 

wouldn't a better strategy be to warn if the function fails, and then just run the event? i.e. to treat return value < 0 like r == 0 effectively, besides the additional logging?

src/udev/udevd.c
```diff
@@ -754,25 +753,44 @@ static int event_run(Event *event) {
...
+        static bool log_children_max_reached = true;
+        Worker *worker;
+
+        assert(manager);
+
+        HASHMAP_FOREACH(worker, manager->workers) {
+                if (worker->state != WORKER_IDLE)
+                        continue;
+
+                return true;
```
poettering: hmm, i'd invert the check:

```c
HASHMAP_FOREACH(worker, manager->workers) 
        if (worker-state == WORKER_IDLE)
                return true;
```

looks a lot more obvious to me, and shorter

src/udev/udevd.c
```diff
@@ -992,6 +1013,9 @@ static int event_queue_start(Manager *manager) {
                 r = event_run(event);
                 if (r <= 0)
                         return r;
+
+                if (!manager_has_idle_worker(manager))
+                        return 0;
```
poettering: why not move this call as first thing into the loop instead? that way you can drop the earlier invocation.

src/udev/udevd.c
```diff
@@ -483,47 +492,6 @@ static int worker_process_device(Manager *manager, sd_device *dev) {
...
-                 * inotify IN_CLOSE_WRITE event, and then run the rules after all and we then realize that
-                 * this wasn't actually requested (i.e. no OPTIONS+="watch" set) we'll simply turn off the
-                 * watching again (see below). Effectively this means: inotify watching is now enabled either
-                 * a) when the udev rules say so, or b) while the device is locked.
-                 *
-                 * Worst case scenario hence: in the (unlikely) case someone locked the device and we clash
-                 * with that we might do inotify watching for a brief moment for a device where we actually
-                 * weren't supposed to. But that shouldn't be too bad, in particular as BSD locks being taken
-                 * on a block device is kinda an indication that the inotify logic is desired too, to some
-                 * degree — they go hand-in-hand after all. */
```
poettering, line 509: this comment should go somewhere, not be removed altogether

src/udev/udevd.c
```diff
@@ -1010,6 +973,36 @@ static int event_queue_start(Manager *manager) {
                 if (r > 0)
                         continue;
 
+                r = event_lock_block_device(event);
```
poettering: so this will open the block device from the main udev process, and i think that's a problem. opening the device is a device access, and that might be slow/blocking. i.e. think about floppies and such. It's ok to open devices from worker processes, because that's what they are for: parallelize potentially slow stuff. But if the main process opens this and stalls then all of udev is just hanging, and that#s not OK.

i think it's essential that opening the device remains in the worker hence.

yuwata: Hmm, indeed...

poettering:
so i am a bit unsure of this one. "udevadm settle" is weird beast, it only means "wait until queue is empty". People (including our own tests, sloppily) use it as "wait until my device is ready to use". That's not really the same though, as typically means a lot more (i.e. it covers *all* devices after all, the whole queue with *any* device in it and that might be a *lot* on bigger systems; moreover it covers events enqueued *after* the caller started waiting, which is really weird) but can also mean a lot less (as we will eat up events in the block device lock case, if devices are migrated to namespaces or we encounter certain errors). While we generally tell people that stuff run from dev rules should be quick and not run for long many rules/callouts can run for a *loooong* time, we have seen plenty of cases for that in the past, so even if you don't have a stampede of events you might still wait very long for the queue to become empty.

So, I'd like to take a step back here. Do we really want to "fix" udevadm settle for the purposes of waiting for devices to appear?

I am not convinced. I think we should much rather add a proper interface for waiting for specific devices, and rework or tests to use that. i.e. maybe add `udevadm wait /dev/sda5` as a new verb, that waits for exactly the device mentioned, and not less and not more. And it does so with a time-out that can be specified on the cmdline, to make this reasonably robust. This would be more robust and quicker too given it waits exactly what it needs to wait for, but not more.

Implementing this should be easy, we already have a helper for precisely that: `device_wait_for_devlink()`.

yuwata:
@poettering Thank you for the review and comments.

> so this will open the block device from the main udev process, and i think that's a problem. opening the device is a device access, and that might be slow/blocking. i.e. think about floppies and such. It's ok to open devices from worker processes, because that's what they are for: parallelize potentially slow stuff. But if the main process opens this and stalls then all of udev is just hanging, and that#s not OK.
> 
> i think it's essential that opening the device remains in the worker hence.

OK. The logic is changed. Now the event is locked by worker as previously we did. But, if the device is already locked by another process, then now the event is requeued instead of skipping it. This way, we can drop to enable inotify watch unconditionally, and wrapping `mkfs` with `flock` works fine.

> So, I'd like to take a step back here. Do we really want to "fix" udevadm settle for the purposes of waiting for devices to appear?

This does not 'fix' `udevadm settle`. Though once commit makes the command work 'better', but it does not 'fix' issues you mentioned. This PR fixes the underlying issue why @mrc0mmand's commits for the test do not work as expected. See the commit message in https://github.com/systemd/systemd/pull/22717/commits/0a40b90335a8af871cd996b01a2ebb37d419ed08.

> I think we should much rather add a proper interface for waiting for specific devices, and rework or tests to use that. i.e. maybe add udevadm wait /dev/sda5 as a new verb, that waits for exactly the device mentioned, and not less and not more. And it does so with a time-out that can be specified on the cmdline, to make this reasonably robust. This would be more robust and quicker too given it waits exactly what it needs to wait for, but not more.

Sounds nice. But let's do that later. As we already have many helpers in TEST-64 which effectively work in the same way.

PTAL.

yuwata:
@mrc0mmand Could you retrigger https://github.com/systemd/systemd-centos-ci/pull/474 again, as I changed the logic. Sorry for bothering you many times, and thank you for your help.

yuwata:
Wow, all green! PTAL!

src/udev/udevd.c
```diff
@@ -400,8 +400,10 @@ static int worker_lock_block_device(sd_device *dev, int *ret_fd) {
 
         fd = open(val, O_RDONLY|O_CLOEXEC|O_NOFOLLOW|O_NONBLOCK);
         if (fd < 0) {
-                log_device_debug_errno(dev, errno, "Failed to open '%s', ignoring: %m", val);
-                return 0;
+                bool ignore = errno == IN_SET(ENOENT, ENXIO, ENODEV);
```
qhill: `bool ignore = IN_SET(errno, ENOENT, ENXIO, ENODEV);`

yuwata: Ouch! Ouch! Ouch!

yuwata:
(Hmm, not related to this PR, but test-repart still fails. The previous workaround seems to not work as expected...)

src/udev/udevd.c
```diff
@@ -1448,10 +1448,11 @@ static int on_sigchld(sd_event_source *s, const struct signalfd_siginfo *si, voi
                         device_tag_index(worker->event->dev, NULL, false);
 
                         if (manager->monitor) {
-                                /* forward kernel event without amending it */
+                                /* forward kernel event without amending anything */
```
keszybz: Hmm, "amending anything" is a bit strange. We generally use "amend" as "change-in-place". "amend it" is clear — the message is not modified, more than "amending anything". Maybe this should be "Forward kernel event unchanged" ?

src/udev/udevd.c
```diff
@@ -1448,10 +1448,11 @@ static int on_sigchld(sd_event_source *s, const struct signalfd_siginfo *si, voi
...
                                 r = device_monitor_send_device(manager->monitor, NULL, worker->event->dev_kernel);
                                 if (r < 0)
-                                        log_device_error_errno(worker->event->dev_kernel, r, "Failed to send back device to kernel: %m");
+                                        log_device_warning_errno(worker->event->dev_kernel, r,
+                                                                 "Failed to broadcast failed event to libudev listners, ignoring: %m");
```
keszybz: Typo: "list**e**ners"

src/udev/udev-ctrl.c
```diff
@@ -101,7 +101,7 @@ static void udev_ctrl_disconnect(UdevCtrl *uctrl) {
         if (!uctrl)
                 return;
 
-        uctrl->event_source_connect = sd_event_source_unref(uctrl->event_source_connect);
+        uctrl->event_source_connect = sd_event_source_disable_unref(uctrl->event_source_connect);
```
keszybz: For the whole commit: does this result in a some known change in behaviour, or is it just a general precaution?

src/udev/udevd.c
```diff
@@ -1480,7 +1474,13 @@ static int on_post(sd_event_source *s, void *userdata) {
         if (!LIST_IS_EMPTY(manager->events))
                 return 1;
 
-        /* There are no pending events. Let's cleanup idle process. */
+        /* There are no queued events. Let's remove /run/udev/queue and cleanup idle process. */
```
keszybz: "cleanup idle process" → "clean up the idle process"

src/udev/udevd.c
```diff
@@ -1480,7 +1474,13 @@ static int on_post(sd_event_source *s, void *userdata) {
...
+
+        if (unlink("/run/udev/queue") < 0) {
+                if (errno != ENOENT)
+                        log_warning_errno(errno, "Failed to unlink /run/udev/queue, ignoring: %m");
+        } else
+                log_debug("No event is queued, /run/udev/queue is removed.");
```
keszybz: "No events are queued, removing /run/udev/queue" ?

src/udev/udevd.c
```diff
@@ -951,24 +951,21 @@ static int event_queue_start(Manager *manager) {
...
+                if (r > 0)
+                        continue;
                 if (r < 0) {
                         sd_device_action_t a = _SD_DEVICE_ACTION_INVALID;
 
                         (void) sd_device_get_action(event->dev, &a);
                         log_device_warning_errno(event->dev, r,
-                                                 "Failed to check event dependency, "
-                                                 "skipping event (SEQNUM=%"PRIu64", ACTION=%s)",
+                                                 "Failed to check dependency for event (SEQNUM=%"PRIu64", ACTION=%s), "
```
keszybz: "check dependency" → "check dependencies" (I assume that there can be more than one).

"assuming ther is no blocking event" ← typo here.

src/udev/udevd.c
```diff
@@ -951,24 +951,21 @@ static int event_queue_start(Manager *manager) {
...
-
-                        event_free(event);
-                        return r;
                 }
-                if (r > 0)
-                        continue;
 
                 r = event_run(event);
-                if (r <= 0)
+                if (r <= 0) /* 0 means there is no idle workers. Let's escape from the loop. */
```
keszybz: "there are"

src/udev/udevd.c
```diff
@@ -476,44 +483,12 @@ static int worker_process_device(Manager *manager, sd_device *dev) {
         if (!udev_event)
                 return -ENOMEM;
 
+        /* If this is a block device and the device is locked currently via the BSD advisory locks —
+         * someone else is exclusively using it. This means we don't run our udev rules now, to not
+         * interfere. In this case, we skip to process and requeue the event, and try to process the
+         * event again after a tiny delay.
+         * (The user-facing side of this: https://systemd.io/BLOCK_DEVICE_LOCKING) */
```
keszybz:
```
         /* If this is a block device and the device is locked currently via the BSD advisory locks,
         * someone else is using it exclusively. We don't run our udev rules now to not
         * interfere. Instead of processing the event we requeue it, and will try again after a delay.
         * 
         * The user-facing side of this: https://systemd.io/BLOCK_DEVICE_LOCKING
         */
```

src/udev/udevd.c
```diff
@@ -1470,8 +1486,15 @@ static int on_post(sd_event_source *s, void *userdata) {
 
         assert(manager);
 
-        if (!LIST_IS_EMPTY(manager->events))
+        if (!LIST_IS_EMPTY(manager->events)) {
+                /* Try to process pending events if there exist free workers. Why is this necessary?
+                 * When a worker finished a event and became idle, even if there was a pending event,
+                 * the corresponding device might be locked and delayed for a while, and the worker
+                 * might not be possible to start to process the event. Now, the device may be
+                 * unlocked. Let's try again! */
```
keszybz:
```
                /* Try to process pending events if free workers exist. Why is this necessary?
                 * When a worker finished an event and became idle, even if there was a pending event,
                 * the corresponding device might have been locked and the processing of the event delaye
                 * for a while, preventing the worker from processing the event immediately. Now, the device may be
                 * unlocked, so let's try again! */
```

test/units/testsuite-64.sh
```diff
@@ -148,6 +148,42 @@ helper_wait_for_pvscan() {
...
+#
+# Note: the array will be CLEARED before the first assignment
+#
+# Arguments:
+#   $1    - NAME of an array in which the commands/argument will be stored
+#   $2-$n - path to devices
+helper_generate_flock_cmdline() {
+    # Create a name reference to the array passed as the first argument
+    # (requires bash 4.3+)
+    local -n cmd_array="${1:?Missing argument: array name}"
```
keszybz: I think the message can be skipped, `${1:?}` autogenerates a reasonable message.

keszybz (commented):
Looks reasonable. Some spelling and grammar fixups.

src/udev/udev-ctrl.c
yuwata: This should not change any behavior. Will mention that in the commit message.

yuwata:
@keszybz Thank you for your review. and sorry for my bad English... All suggestions are applied. PTAL.

keszybz:
LGTM.

src/udev/udevd.c
```diff
@@ -399,8 +399,10 @@ static int worker_lock_block_device(sd_device *dev, int *ret_fd) {
 
         fd = open(val, O_RDONLY|O_CLOEXEC|O_NOFOLLOW|O_NONBLOCK);
         if (fd < 0) {
-                log_device_debug_errno(dev, errno, "Failed to open '%s', ignoring: %m", val);
-                return 0;
+                bool ignore = IN_SET(errno, ENOENT, ENXIO, ENODEV);
```
poettering: i wonder if we should add ERRNO_IS_DEVICE_ABSENT() as a shortcut helper for this. Sounds like something we could use more than once

src/udev/udevd.c
```diff
@@ -972,6 +963,31 @@ static int event_queue_start(Manager *manager) {
...
+        if (event->worker && event->worker->event == event)
+                event->worker->event = NULL;
+        event->worker = NULL;
+
+        /* add a short delay to suppress busy loop */
+        r = sd_event_now(event->manager->event, clock_boottime_or_monotonic(), &now);
+        if (r < 0)
+                return log_warning_errno(r, "Failed to get current time, requeueing event without delay: %m");
+
+        event->wait_until = usec_add(now, 200 * USEC_PER_MSEC);
```
poettering: hmm, so instead of going just by time, wouldn't it make sense to use "secondary" events as trigger for trying again to process the uevent?

by "secondary" events I mean mostly two things: inotify close events and further uevents for the same device?

i.e. let's say a device is locked, then maybe we should just queue the uevent until something later happens, and in that case process the queued uevent first, and then afterwards everything that happened later on it? i.e. use later happenings as reason to reconsider the queued earlier events for the very same device?

In that scheme I think going by timer also makes sense, but more as a safety net to ensure that uevents don#t remain stuck in the queue indefinitely.

poettering:
So I like the approach of putting the event back in the queue if the device is locked. But I think we should think again about triggers for making the queue "unstuck". i.e. as discussed above, I think:

1. the timeout hsould just be a safety net, to ensure uevents don#t remain queued forever
2. later uevents on a device should be used as reason to check if the older queued uevents on the same device can now be processed
3. later inotify events the same way

net result would be:

1. no uevents would get lost anymore
2. we use proper IO events as trigger to recheck the queue uevents, instead of dumbly going by time
3. we have the guarantee that events won#t remain queued forever via the time logic

Which sounds like a perfect, winning combination to me.

poettering:
And sorry for taking so long for the re-review

src/udev/udevd.c
poettering: → #22856

yuwata:
@poettering Thank you for the review. I hope the commit "udev: assume block device is not locked when a new event is queued" addresses the suggestion. PTAL.

src/udev/udevd.c
```diff
@@ -128,6 +128,7 @@ typedef struct Event {
 
         uint64_t seqnum;
         uint64_t blocker_seqnum;
+        usec_t wait_until;
```
poettering: hmm, maybe call this "retry_again_timeout_usec" or so? i.e. emphasize that this is about retry again?

poettering: so, with this logic we'll still retry idefinitely, no?

I think we should have also a time at which we will give up, and fail the event. So that we actually get the guarantee events are dropped eventually from the queue.

Hence, maybe a logic: first time a worker sends EVENT_RESULT_TRY_AGAIN we'll set two times: 
1. `retry_again_next_usec` to a short time (1s or so)
2. `retry_again_timeout_usec` to a long time (3 min or so)

And then, after each retry bump the "next" but not the "timeout", which we'll fail instead.

src/udev/udevd.c
```diff
@@ -1448,10 +1448,11 @@ static int on_sigchld(sd_event_source *s, const struct signalfd_siginfo *si, voi
                         device_tag_index(worker->event->dev, NULL, false);
 
                         if (manager->monitor) {
-                                /* forward kernel event without amending it */
+                                /* Forward kernel event unchanged */
```
poettering: btw, I think we should revisit this bit, i.e. when we forward the thing after the worker failed, I think we should say so in the event. i.e. add a property "UDEV_WORKER_FAILED=1" or so (and maybe even the exit status/code as separate props).

That's useful for debugging, but also is less confusing for apps: if apps see these events they know why their props are missing.

src/udev/udevd.c
```diff
@@ -404,6 +398,27 @@ static int worker_lock_block_device(sd_device *dev, int *ret_fd) {
         if (r < 0)
                 return log_device_debug_errno(dev, r, "Failed to get devname: %m");
 
+        *ret = val;
+        return 1;
```
poettering, line 419: this function has successfuly exit paths (i.e. where 0 is returned), where *ret is not initialized. that's not in line with our coding style, which after all says all succesful exit paths from a function should initialize all return values. or in other words, you need to sprinkle `*ret = NULL` at before various occurences of `return 0`, above.

poettering (changes requested):

src/udev/udevd.c
```diff
@@ -1055,6 +1103,8 @@ static int on_uevent(sd_device_monitor *monitor, sd_device *dev, void *userdata)
                 return 1;
         }
 
+        (void) event_queue_assume_block_device_unlocked(manager, dev);
```
poettering, line 1137: (doesn't really matter, but we could in theory speed up the inotify case, and immediately unblock the uevents for a device when we see inotify, and not wait until the retrigger uevent percolated through the kernel back to us, no?)

poettering:
will cherry-pick a bunch of the initial commits into git main, given they already passed CI before and are not strictly related to the later stuff.

src/udev/udevd.c
yuwata: Interesting. But let's do that in another later PR.

yuwata:
@poettering Thank you for the clarification. Now, timeout is also set, and inotify event triggers requeued events. PTAL.

PR #22727, opened 2022-03-14
Journald preserve acl when rotating

fbuihuu:


poettering:
hmm, acls are just xattrs, why isn't sufficeint to copy those?

poettering:
for the second patch, adding the COPY_XATTR_ALL flag should suffice, no?

fbuihuu:
> hmm, acls are just xattrs, why isn't sufficeint to copy those?

Well I must admit that I haven't looked at how acls are implemented, but AFAICS they're not part of the xattrs returned by `flistxattr_malloc()`, which are:

```
user.crtime_usec
user.crtime_usec
system.posix_acl_access
```

fbuihuu:
Ah ok then passing `COPY_XATTR_ALL` should be sufficient. Let me give it a test.

fbuihuu:
Passing `COPY_XATTR_ALL` fixed the issue.

I force pushed a new version which does that, PTAL.

DaanDeMeyer (approved):

bluca (approved):

PR #22732, opened 2022-03-14
docs: document how we usually define enums

poettering:


docs/CODING_STYLE.md
```diff
@@ -79,6 +79,46 @@ SPDX-License-Identifier: LGPL-2.1-or-later
...
+  ```
+
+  i.e. use an enum for it, if possible. Indicate bit values via `1 <<`
+  expressions, and align them vertically. Define both an enum and a type for
+  it.
+
+- If you define (non-flags) enums, follow this template:
+
+  ```c
+  typedf enum FoobarMode {
```
daandemeyer: Typo

poettering:
ouch. fixed.

PR #22734, opened 2022-03-14
DECIMAL_STR_WIDTH() + DECIMAL_STR_MAX() tweaks

poettering:
Inspired by e3dd9ea8ea4510221f73071ad30ee657ca77565d

### DIFF
--- a/src/test/test-macro.c
+++ b/src/test/test-macro.c
@@ -410,4 +410,27 @@ TEST(DECIMAL_STR_WIDTH) {
         assert_se(DECIMAL_STR_WIDTH(INT64_MIN) == STRLEN("-9223372036854775808"));
 }
 
+TEST(DECIMAL_STR_MAX) {
+        int8_t s8_longest = INT8_MIN;
+        int16_t s16_longest = INT16_MIN;
+        int32_t s32_longest = INT32_MIN;
+        int64_t s64_longest = INT64_MIN;
+        uint8_t u8_longest = UINT8_MAX;
+        uint16_t u16_longest = UINT16_MAX;
+        uint32_t u32_longest = UINT32_MAX;
+        uint64_t u64_longest = UINT64_MAX;
+
+        /* NB: Always add +1, because DECIMAL_STR_MAX() includes space for trailing NUL byte, but
+         * DECIMAL_STR_WIDTH() does not! */
+        assert_se(DECIMAL_STR_MAX(int8_t) == DECIMAL_STR_WIDTH(s8_longest)+1);
+        assert_se(DECIMAL_STR_MAX(int16_t) == DECIMAL_STR_WIDTH(s16_longest)+1);
+        assert_se(DECIMAL_STR_MAX(int32_t) == DECIMAL_STR_WIDTH(s32_longest)+1);
+        assert_se(DECIMAL_STR_MAX(int64_t) == DECIMAL_STR_WIDTH(s64_longest)+1);
+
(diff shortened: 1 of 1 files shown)


src/basic/macro.h
```diff
@@ -309,22 +309,31 @@ static inline int __coverity_check_and_return__(int condition) {
...
         ({                                                        \
                 typeof(x) _x_ = (x);                              \
-                size_t ans = IS_SIGNED_INTEGER_TYPE(_x_) ? 2 : 1; \
-                while ((_x_ /= 10) != 0)                          \
-                        ans++;                                    \
+                size_t ans;                                       \
+                if (_x_ == 0)                                     \
+                        ans = 1;                                  \
+                else {                                            \
+                        ans = _x_ <= 0 ? 2 : 1;                   \
```
mrc0mmand, line 333: D'oh, neat!

yuwata (approved):
LGTM.

src/basic/macro.h
```diff
@@ -309,15 +309,14 @@ static inline int __coverity_check_and_return__(int condition) {
...
+/* Returns the number of chars needed to format variables of the specified type as a decimal string. Adds in
+ * extra space for a negative '-' prefix for signed types. Includes space for the trailing NUL. */
 #define DECIMAL_STR_MAX(type)                                           \
-        (2U+(sizeof(type) <= 1 ? 3U :                                   \
+        ((size_t) IS_SIGNED_INTEGER_TYPE(type) + 1U +                   \
+            (sizeof(type) <= 1 ? 3U :                                   \
              sizeof(type) <= 2 ? 5U :                                   \
              sizeof(type) <= 4 ? 10U :                                  \
-             sizeof(type) <= 8 ? 20U : sizeof(int[-2*(sizeof(type) > 8)])))
+             sizeof(type) <= 8 ? (IS_SIGNED_INTEGER_TYPE(type) ? 19U : 20U) : sizeof(int[-2*(sizeof(type) > 8)])))
```
medhefgo, line 319: This is confusing me. Why only care about signed for sizeof == 8 while happily over-estimating for the smaller ones? Or am I missing something?

src/test/test-macro.c
```diff
@@ -410,4 +410,27 @@ TEST(decimal_str_width) {
         assert_se(DECIMAL_STR_WIDTH(INT64_MIN) == STRLEN("-9223372036854775808"));
 }
 
+TEST(decimal_str_max) {
```
medhefgo: Doesn't really matter, but so far, uppercase macros also had their test name uppercase. (Goes for the decimal_str_width test too)

src/basic/macro.h
poettering: the commit msg is supposed to explain that...

1. INT8_MIN and UINT8_MAX (as the "longest" decimal numbers fitting in 8 bit) both have three digits (only talking about digits here, ignore the "-" for now, it is handled alreay in the first component of the expression)
2. similar, INT16_MIN and UINT16_MAX have 5 digits
3. INT32_MIN and UINT32_MAX have 10 digits.
4. but INT64_MIN and UINT64_MAX have a different number of digits! INT64_MIn as 19 and UINt64_MAX has 20! hence the special casing.

medhefgo: I see. now it makes sense. But that really doesn't come across in the commit message (to me).

poettering: i extended the commit msg a bit

PR #22721, opened 2022-03-12
journal: convert --list-boots to a table

mrc0mmand:
so it can be output as JSON as well.

```
$ build-san/journalctl --list-boots --file boot-test.journal
#  BOOT ID                          FIRST ENTRY                 LAST ENTRY
-3 39d66eb1925f4d01b8464d502650a714 Sat 2022-03-05 15:20:33 CET Sat 2022-03-05 16:19:21 CET
-2 5dffeb08a27344d5ae9e2fc244bbcbc5 Fri 2022-03-11 17:23:57 CET Fri 2022-03-11 17:38:31 CET
-1 c8ebd52915b642c39eda4bf00f864f79 Fri 2022-03-11 17:38:41 CET Fri 2022-03-11 20:03:46 CET
 0 00bcba97c7094fa88cc5d1cf2a389057 Sat 2022-03-12 20:39:08 CET Sat 2022-03-12 20:40:18 CET

$ build-san/journalctl --list-boots --file boot-test.journal -q
-3 39d66eb1925f4d01b8464d502650a714 Sat 2022-03-05 15:20:33 CET Sat 2022-03-05 16:19:21 CET
-2 5dffeb08a27344d5ae9e2fc244bbcbc5 Fri 2022-03-11 17:23:57 CET Fri 2022-03-11 17:38:31 CET
-1 c8ebd52915b642c39eda4bf00f864f79 Fri 2022-03-11 17:38:41 CET Fri 2022-03-11 20:03:46 CET
 0 00bcba97c7094fa88cc5d1cf2a389057 Sat 2022-03-12 20:39:08 CET Sat 2022-03-12 20:40:18 CET

$ build-san/journalctl --list-boots --file boot-test.journal -o json-pretty
[
        {
                "#" : "-3",
                "boot_id" : "39d66eb1925f4d01b8464d502650a714",
                "first_entry" : 1646490033438495,
                "last_entry" : 1646493561047353
        },
        {
                "#" : "-2",
                "boot_id" : "5dffeb08a27344d5ae9e2fc244bbcbc5",
                "first_entry" : 1647015837289036,
                "last_entry" : 1647016711595489
        },
        {
                "#" : "-1",
                "boot_id" : "c8ebd52915b642c39eda4bf00f864f79",
                "first_entry" : 1647016721056382,
                "last_entry" : 1647025426397414
        },
        {
                "#" : "0",
                "boot_id" : "00bcba97c7094fa88cc5d1cf2a389057",
[...]

Issue #16274, opened 2020-06-25
journalctl: Minor command line option inconsistencies


wlhlm:
**systemd version the issue has been seen with**
> systemd 245 (245.6-7-arch)
+PAM +AUDIT -SELINUX -IMA -APPARMOR +SMACK -SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid

**Used distribution**
 > Arch Linux

**Expected behavior you didn't see**
1. `journalctl --help --no-pager` should print a summary of journalctl's command line options without invoking a pager
2. running `journalctl --list-boots --reverse` should output the list of boots in reverse beginning from the most recent

**Unexpected behavior you saw**
Really, these seem like pretty minor issues, but I'll note them here anyway.

1. `journalctl --help --no-pager` still starts a pager. Note that `journalctl --no-pager --help` *does* work as expected.
2. `journalctl --list-boots --reverse` does not print the boot list in reverse order (order of command line options doesn't matter in this case)

PR #22739, opened 2022-03-14
A couple of follow-ups for #22721

mrc0mmand:
As pointed out by @poettering (https://github.com/systemd/systemd/pull/22721#pullrequestreview-908885714).

### DIFF
--- a/src/journal/journalctl.c
+++ b/src/journal/journalctl.c
@@ -1471,6 +1471,9 @@ static int list_boots(sd_journal *j) {
         if (r < 0)
                 return log_error_errno(r, "Failed to set JSON field name of column 0: %m");
 
+        (void) table_set_sort(table, (size_t) 0);
+        (void) table_set_reverse(table, 0, arg_reverse);
+
         i = 0;
         LIST_FOREACH(boot_list, id, all_ids) {
                 r = table_add_many(table,


mrc0mmand:
Also added a fix for #16274, since thanks to @poettering I found out about a yet another nifty feature of our table-stuff implementation.

src/journal/journalctl.c
```diff
@@ -1460,13 +1460,19 @@ static int list_boots(sd_journal *j) {
...
 
-        table = table_new(OUTPUT_MODE_IS_JSON(arg_output) ? "index" : "idx", "boot id", "first entry", "last entry");
+        table = table_new("idx", "boot id", "first entry", "last entry");
         if (!table)
                 return log_oom();
 
         if (arg_full)
                 table_set_width(table, 0);
 
+        if (OUTPUT_MODE_IS_JSON(arg_output)) {
```
poettering: drop the conditionalization. You can call table_set_json_field_name() in any case. If the json output mode is not selected it of course won't actually do anything, but it doesn't hurt either.

src/journal/journalctl.c
```diff
@@ -1473,6 +1473,14 @@ static int list_boots(sd_journal *j) {
                         return log_error_errno(r, "Failed to set JSON field name of column 0: %m");
         }
 
+        r = table_set_sort(table, (size_t) 0);
+        if (r < 0)
+                return r;
+
+        r = table_set_reverse(table, 0, arg_reverse);
+        if (r < 0)
+                return r;
```
poettering: our coding style says: functions should either log about all their errors, or about none (in which case they should propagate the error code so that the caller can log). the errors we see in `list_boots()` are all logged about. But here you propagate the errors from two functions as-is without logging. hence two options:

1. continue propagating these errors, but start logging about these errors, i.e. do `return log_error_errno(r, …` here instead of just `return r;`
2. just eat up the errors, cast the return value of `table_set_sort()` and `table_set_reverse()` to `(void)`. i.e. don't log and don't propagate. We do this in some other cases like this. i.e. stuff whose only purpose is tweak output a bit is OK to fail silently goes the thinking behind it. 

which of the two options you pick is up to you: we have prior art for both ways to handle this in our codebase. The only option that it isn't OK here is to propagate the error and not log about it.

poettering: similar logic applies to handling of table_set_json_field_name() errors btw. there you chose to log+propagate, which is OK. (but in fact I'd lso be OK if you'd switch to no-log+no-propagate.)

mrc0mmand: I see, thanks. I guess I got confused how it's handled in https://github.com/systemd/systemd/blob/main/src/analyze/analyze-blame.c#L43, will update it.

src/journal/journalctl.c
mrc0mmand: Ah, that's neat, thanks!

src/journal/journalctl.c
```diff
@@ -1457,13 +1460,20 @@ static int list_boots(sd_journal *j) {
...
 
-        table = table_new(OUTPUT_MODE_IS_JSON(arg_output) ? "index" : "idx", "boot id", "first entry", "last entry");
+        table = table_new("idx", "boot id", "first entry", "last entry");
         if (!table)
                 return log_oom();
 
         if (arg_full)
                 table_set_width(table, 0);
 
+        r = table_set_json_field_name(table, 0, "index");
```
yuwata, line 1470: Nice. I did not know the function.

yuwata (approved):
LGTM.

PR #22740, opened 2022-03-14
tree-wide: declare iterator of LIST_FOREACH() in the loop

yuwata:
Prompted by https://lwn.net/Articles/887097/

### DIFF
--- a/src/basic/path-util.c
+++ b/src/basic/path-util.c
@@ -202,9 +202,9 @@ int path_make_relative(const char *from, const char *to, char **ret) {
 }
 
 char* path_startswith_strv(const char *p, char **set) {
-        char **s, *t;
-
         STRV_FOREACH(s, set) {
+                char *t;
+
                 t = path_startswith(p, *s);
                 if (t)
                         return t;
@@ -214,7 +214,6 @@ char* path_startswith_strv(const char *p, char **set) {
 }
 
 int path_strv_make_absolute_cwd(char **l) {
-        char **s;
         int r;
 
         /* Goes through every item in the string list and makes it
@@ -236,7 +235,6 @@ int path_strv_make_absolute_cwd(char **l) {
 }
 
 char **path_strv_resolve(char **l, const char *root) {
-        char **s;
         unsigned k = 0;
         bool enomem = false;
         int r;
@@ -700,12 +698,12 @@ int find_executable_full(const char *name, const char *root, char **exec_search_
                 p = DEFAULT_PATH;
 
         if (exec_search_path) {
-                char **element;
-
                 STRV_FOREACH(element, exec_search_path) {
                         _cleanup_free_ char *full_path = NULL;
+
                         if (!path_is_absolute(*element))
                                 continue;
+
                         full_path = path_join(*element, name);
                         if (!full_path)
                                 return -ENOMEM;
@@ -754,7 +752,6 @@ int find_executable_full(const char *name, const char *root, char **exec_search_
 
 bool paths_check_timestamp(const char* const* paths, usec_t *timestamp, bool update) {
         bool changed = false, originally_unset;
-        const char* const* i;
 
         assert(timestamp);
 
@@ -1339,7 +1336,7 @@ int systemd_installation_has_version(const char *root, unsigned minimal_version)
 
                 _cleanup_strv_free_ char **names = NULL;
                 _cleanup_free_ char *path = NULL;
-                char *c, **name;
+                char *c;
 
                 path = path_join(root, pattern);
                 if (!path)
@@ -1411,8 +1408,6 @@ bool empty_or_root(const char *path) {
 }
 
 bool path_strv_contains(char **l, const char *path) {
-        char **i;
-
         STRV_FOREACH(i, l)
                 if (path_equal(*i, path))
                         return true;
@@ -1421,10 +1416,9 @@ bool path_strv_contains(char **l, const char *path) {
 }
 
 bool prefixed_path_strv_contains(char **l, const char *path) {
-        char **i, *j;
-
         STRV_FOREACH(i, l) {
-                j = *i;
+                const char *j = *i;
+
                 if (*j == '-')
                         j++;
                 if (*j == '+')
--- a/src/basic/strv.h
+++ b/src/basic/strv.h
@@ -122,19 +122,30 @@ static inline int strv_from_nulstr(char ***a, const char *nulstr) {
 
 bool strv_overlap(char * const *a, char * const *b) _pure_;
 
+#define _STRV_FOREACH(s, l, i)                                          \
+        for (typeof(*(l)) *s, *i = (l); (s = i) && *i; i++)
+
 #define STRV_FOREACH(s, l)                      \
-        for ((s) = (l); (s) && *(s); (s)++)
+        _STRV_FOREACH(s, l, UNIQ_T(i, UNIQ))
 
-#define STRV_FOREACH_BACKWARDS(s, l)                                \
-        for (s = ({                                                 \
-                        typeof(l) _l = l;                           \
-                        _l ? _l + strv_length(_l) - 1U : NULL;      \
-                        });                                         \
-             (l) && ((s) >= (l));                                   \
-             (s)--)
+#define _STRV_FOREACH_BACKWARDS(s, l, h, i)                             \
+        for (typeof(*(l)) *s, *h = (l), *i = ({                         \
+                                size_t _len = strv_length(h);           \
+                                _len > 0 ? h + _len - 1 : NULL;         \
+                        });                                             \
+             i && (s = i) >= h;                                         \
+             i--)
 
-#define STRV_FOREACH_PAIR(x, y, l)               \
-        for ((x) = (l), (y) = (x) ? (x+1) : NULL; (x) && *(x) && *(y); (x) += 2, (y) = (x + 1))
+#define STRV_FOREACH_BACKWARDS(s, l)                                    \
+        _STRV_FOREACH_BACKWARDS(s, l, UNIQ_T(h, UNIQ), UNIQ_T(i, UNIQ))
+
+#define _STRV_FOREACH_PAIR(x, y, l, i)                          \
+        for (typeof(*l) *x, *y, *i = (l);                       \
+             i && *(x = i) && *(y = i + 1);                     \
+             i += 2)
+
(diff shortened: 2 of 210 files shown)


src/basic/list.h
```diff
@@ -137,32 +137,32 @@
         (!(item)->name##_prev && !(item)->name##_next)
 
 #define LIST_FOREACH(name,i,head)                                       \
-        for ((i) = (head); (i); (i) = (i)->name##_next)
+        for (typeof(head) i = (head); i; i = i->name##_next)
```
bluca: Can we use the `UNIQ_T` stuff here?

yuwata: No, the iterator must be usable in the loop. Hence, the provided name must be used.

yuwata:
There exists a problem in the current version. The following does not work.
```
typedef struct Foo Foo;
struct Foo {
        LIST_FIELDS(Foo, foo);
};

typedef struct Obj {
        LIST_HEAD(Foo, head);
} Obj;

void func(const Obj *obj) {
        LIST_FOREACH(foo, i, obj->head) {
                ...;
        }
}
```
But `LIST_FOREACH(foo, i, (Foo*) obj->head)` works.

Because, `typeof(obj->head)` is `Foo * const` instead of `const Foo *`...

src/basic/list.h
bluca: I see - but then nesting won't work?

yuwata: Why? There is no variable implicitly declared. We can safely use the macro arbitrary level of nesting:
```
LIST_FOREACH(name1, i, head1)
    LIST_FOREACH(name2, j, head2)
        ...;
```

bluca: ah yes, I misread it, sorry

yuwata:
@poettering, @keszybz and @bluca WDYT about the basic of this PR? If you like this, then I will apply similar changes to other loop macros.

bluca:
> @poettering, @keszybz and @bluca WDYT about the basic of this PR? If you like this, then I will apply similar changes to other loop macros.

I like it! Maybe explicitly document in list.h that the iterator is declared for you by the macro, since that could be somewhat unexpected, but otherwise looks like a good usability improvement to me

yuwata:
All green now. PTAL.

src/libsystemd-network/sd-radv.c
```diff
@@ -659,15 +654,18 @@ int sd_radv_add_prefix(sd_radv *ra, sd_radv_prefix *p) {
         return 0;
 }
 
-sd_radv_prefix *sd_radv_remove_prefix(sd_radv *ra,
-                                               const struct in6_addr *prefix,
-                                               unsigned char prefixlen) {
-        sd_radv_prefix *cur, *next;
+void sd_radv_remove_prefix(
```
bluca: Could you please split this function refactor in a separate commit? Lumped together with the tree-wide iterator change it's very easy to miss

src/resolve/resolved-dns-zone.c
```diff
@@ -78,12 +78,13 @@ void dns_zone_flush(DnsZone *z) {
 }
 
 DnsZoneItem* dns_zone_get(DnsZone *z, DnsResourceRecord *rr) {
-        DnsZoneItem *i;
+        DnsZoneItem *first;
 
         assert(z);
         assert(rr);
 
-        LIST_FOREACH(by_key, i, hashmap_get(z->by_key, rr->key))
```
bluca, line 86: the fact that function calls cannot be inlined anymore also needs to be documented in the list header I think

bluca (approved):
Overall I like the change - just some minor nits. Not being able to inline function calls in the macro invocation anymore might be a the major gotcha, but might be worth it in the end.

src/resolve/resolved-dns-zone.c
yuwata: Or, in this case, you can inline them like the following:
```
LIST_FOREACH(by_key, i, (DnsZoneItem*) hashmap_get(...))
```

bluca: If it's still possible, maybe do that to keep the diff to a minimum? There's other instances as well when a temporary new variable was added to avoid the inline function call

yuwata: Sure.

src/libsystemd-network/sd-radv.c
yuwata: Done.

src/partition/repart.c
```diff
@@ -4797,9 +4772,8 @@ static int resize_backing_fd(
 
 static int determine_auto_size(Context *c) {
         uint64_t sum;
-        Partition *p;
 
-        assert_se(c);
+        assert(c);
```
medhefgo, line 4773: `assert_se -> assert` intentional?

medhefgo (approved):
+1 From me just because of the lines deleted count alone :smile_cat:.

src/partition/repart.c
yuwata: Yes, this is not a test code. Maybe, I should split the commit for the change.

poettering:
Note that we don't have the same underlying problem as the kernel people with this, as we don#t do this containerof() based logic they usually do with embedding. i.e. our stuff uses the containing type as primary type we use, and does not embedd a structure from which one would have to return to the containing type somehow.

That said, it looks worthwhile from a first glance, given how it makes the code shorter.

src/basic/list.h
```diff
@@ -139,14 +139,17 @@
 /* The type of the iterator 'i' is automatically determined by the type of 'head', and declared in the
  * loop. Hence, do not declare the same variable in the outer scope. Sometimes, we set 'head' through
  * hashmap_get(). In that case, you need to explicitly cast the result. */
+#define LIST_FOREACH_WITH_NEXT(name,i,n,head)                           \
```
keszybz, line 142: Shouldn't this be `_LIST_FOREACH_WITH_NEXT()` to match `_LIST_FOREACH_WITH_PREV()` ?

keszybz (commented):
Looks great! Thank you for doing this.

src/basic/path-util.c
```diff
@@ -202,9 +202,9 @@ int path_make_relative(const char *from, const char *to, char **ret) {
 }
 
 char* path_startswith_strv(const char *p, char **set) {
-        char **s, *t;
-
         STRV_FOREACH(s, set) {
+                char *t;
```
medhefgo, line 206: Doesn't really matter, but as I understand it, the iterator is now safe to modify, so this t could be done away with, no?

src/basic/path-util.c
```diff
@@ -1421,10 +1416,9 @@ bool path_strv_contains(char **l, const char *path) {
 }
 
 bool prefixed_path_strv_contains(char **l, const char *path) {
-        char **i, *j;
-
         STRV_FOREACH(i, l) {
-                j = *i;
+                char *j = *i;
```
medhefgo: Same here.

src/basic/strv.h
```diff
@@ -122,19 +122,28 @@ static inline int strv_from_nulstr(char ***a, const char *nulstr) {
...
+
 #define STRV_FOREACH(s, l)                      \
-        for ((s) = (l); (s) && *(s); (s)++)
+        _STRV_FOREACH(s, l, UNIQ_T(i, UNIQ))
+
+#define _STRV_FOREACH_BACKWARDS(s, l, h, i)                             \
+        for (typeof(*(l)) *s, *h = (l),                                 \
+                     *i = h ? h + strv_length(h) - 1U : NULL;           \
+             h && (s = i) >= h;                                         \
+             i--)
```
medhefgo, line 137: Not a total expert, but isn't creating a pointer that points (one element) before a malloc technically UB? (I know the code before this does this already.)

src/basic/path-util.c
poettering: the function doesn't modify the iterator though?

src/basic/path-util.c
```diff
@@ -1421,10 +1416,9 @@ bool path_strv_contains(char **l, const char *path) {
...
+
```
poettering, line 1421: mayb make j const

src/basic/path-util.c
medhefgo: I was implying that we can reuse s instead of declaring t.

poettering:
i cherry-picked the middle PRs into git main, they make sense on their own.

PR looks pretty good to me, see my minor points above. Great work!

src/basic/path-util.c
yuwata: Actually, here `t` can be dropped and the iterator can be reused. But I prefer the current form for readability.

src/basic/list.h
yuwata: The macro `LIST_FOREACH_WITH_NEXT()` is explicitly used in udev-rules.c.

src/basic/path-util.c
yuwata: Also, this is just for readability.

src/basic/path-util.c
yuwata: Done.

src/basic/strv.h
yuwata: This should be safe, but I updated this to make safer. PTAL.

yuwata:
Thank you for all comments.
Rebased, and `STRV_FOREACH_BACKWARD()` is updated to address @medhefgo's concern.
PTAL.

src/basic/strv.h
medhefgo: (This is still decrementing `i` and then checking, which would break for `i` pointing at the 0th string.)

medhefgo:
Also, would be nice if you could move `STRV_FOREACH` into `string-fundamental.h` at the same time, then I won't have to move it myself. :D

src/basic/path-util.c
keszybz: Yeah, it's better to keep it.

src/basic/strv.h
keszybz: I think this is a valid point, and we should make sure to not do UB. But since this is a "preexisting condition", let's leave it for later.

keszybz:
Let's merge this before we get conflicts. Some further follow-ups are certainly possible, but the changes done here stand on their own.

src/basic/strv.h
keszybz: https://github.com/systemd/systemd/pull/22797

PR #22750, opened 2022-03-15
varlink: varlink_error_invalid_parameter(...) always returns EINVAL

vchillara:
varlink_error(...) expects a json object as the third parameter. Passing a string variant causes
parameter sanitization to fail, and it returns -EINVAL. Pass object variant instead.

### DIFF
--- a/src/shared/varlink.c
+++ b/src/shared/varlink.c
@@ -1660,6 +1660,7 @@ int varlink_errorb(Varlink *v, const char *error_id, ...) {
 }
 
 int varlink_error_invalid_parameter(Varlink *v, JsonVariant *parameters) {
+        int r;
 
         assert_return(v, -EINVAL);
         assert_return(parameters, -EINVAL);
@@ -1669,13 +1670,33 @@ int varlink_error_invalid_parameter(Varlink *v, JsonVariant *parameters) {
          * variant in which case we'll pull out the first key. The latter mode is useful in functions that
          * don't expect any arguments. */
 
-        if (json_variant_is_string(parameters))
-                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER, parameters);
+        /* varlink_error(...) expects a json object as the third parameter. Passing a string variant causes
+         * parameter sanitization to fail, and it returns -EINVAL. */
+
+        if (json_variant_is_string(parameters)) {
+                _cleanup_(json_variant_unrefp) JsonVariant *parameters_obj = NULL;
+
+                r = json_build(&parameters_obj,
+                                JSON_BUILD_OBJECT(
+                                        JSON_BUILD_PAIR("parameter", JSON_BUILD_VARIANT(parameters))));
+                if (r < 0)
+                        return r;
+
+                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER, parameters_obj);
+        }
 
         if (json_variant_is_object(parameters) &&
-            json_variant_elements(parameters) > 0)
-                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER,
-                                     json_variant_by_index(parameters, 0));
+            json_variant_elements(parameters) > 0) {
+                _cleanup_(json_variant_unrefp) JsonVariant *parameters_obj = NULL;
+
+                r = json_build(&parameters_obj,
+                                JSON_BUILD_OBJECT(
+                                        JSON_BUILD_PAIR("parameter", JSON_BUILD_VARIANT(json_variant_by_index(parameters, 0)))));
+                if (r < 0)
+                        return r;
+
+                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER, parameters_obj);
+        }
 
         return -EINVAL;
 }


src/shared/varlink.c
```diff
@@ -1669,13 +1671,22 @@ int varlink_error_invalid_parameter(Varlink *v, JsonVariant *parameters) {
...
 
-        if (json_variant_is_string(parameters))
-                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER, parameters);
+        /* varlink_error(...) expects a json object as the third parameter. Passing a string variant causes
+         * parameter sanitization to fail, and it returns -EINVAL. */
+
+        if (json_variant_is_string(parameters)) {
+                r = json_build(&parameters_obj,
+                                JSON_BUILD_OBJECT(
+                                        JSON_BUILD_PAIR("parameter", JSON_BUILD_STRING(json_variant_string(parameters)))));
```
yuwata: Creating a new object in a error handling is not good in general...
I am not familiar with varlink library, but is it possible to introduce, say `varlink_error_string()` or something which takes string json object?

poettering: I think that's OK. THis is so high-level. We have to buffer the generated varlink stuff anyway, hence allocating a json object is fine, too

src/shared/varlink.c
```diff
@@ -1669,13 +1671,22 @@ int varlink_error_invalid_parameter(Varlink *v, JsonVariant *parameters) {
...
+                        return r;
+
+                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER, parameters_obj);
+        }
 
         if (json_variant_is_object(parameters) &&
             json_variant_elements(parameters) > 0)
-                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER,
-                                     json_variant_by_index(parameters, 0));
+                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER, parameters);
```
poettering: this doesn't look right. we should also pass an object here with one field called "parameter" that points to the first element of the parameters list. See the comment at the top of the function about this.

or in other words: the json_build() line you added needs to be added both for the case where a string and where an object is passed.

src/shared/varlink.c
```diff
@@ -1660,6 +1660,8 @@ int varlink_errorb(Varlink *v, const char *error_id, ...) {
 }
 
 int varlink_error_invalid_parameter(Varlink *v, JsonVariant *parameters) {
+        _cleanup_(json_variant_unrefp) JsonVariant *parameters_obj = NULL;
+        int r = 0;
```
yuwata: It is not necessary to initialize `r` here. Please simply `int r;`.

src/shared/varlink.c
yuwata: Still I do not like it, but I am not familiar with this, hence, it is OK, as it is liked by @poettering.

~~But please move `_cleanup_(json_variant_unrefp) JsonVariant *parameters_obj = NULL;` to this `if` block.~~

yuwata (changes requested):

src/shared/varlink.c
vchillara: @yuwata  now that we're using parameters_obj in both if blocks, does it still make sense to move the declaration to within the block(s)?

bluca: yes, they are used independently

bluca: (which means by moving we pay the cost of the _cleanup_ only if it's actually needed and the block is entered, and given each block returns there's no need for the var to live longer)

src/shared/varlink.c
```diff
@@ -1669,13 +1670,33 @@ int varlink_error_invalid_parameter(Varlink *v, JsonVariant *parameters) {
...
-            json_variant_elements(parameters) > 0)
-                return varlink_error(v, VARLINK_ERROR_INVALID_PARAMETER,
-                                     json_variant_by_index(parameters, 0));
+            json_variant_elements(parameters) > 0) {
+                _cleanup_(json_variant_unrefp) JsonVariant *parameters_obj = NULL;
+
+                r = json_build(&parameters_obj,
+                                JSON_BUILD_OBJECT(
+                                        JSON_BUILD_PAIR("parameter", JSON_BUILD_VARIANT(json_variant_by_index(parameters, 0)))));
+                if(r < 0)
```
bluca: style nit: missing space before `(`

bluca (approved):

src/shared/varlink.c
poettering: please replace `JSON_BUILD_STRING(json_variant_string(parameters))` by JSON_BUILD_VARIANT(parameters)`. i.e. we can just reuse the string variant we already have, no need to allocate a new one.

poettering:
looks good, just one minor simplifcation please

src/shared/varlink.c
vchillara: > please replace `JSON_BUILD_STRING(json_variant_string(parameters))` by JSON_BUILD_VARIANT(parameters)`. i.e. we can just reuse the string variant we already have, no need to allocate a new one.

updated

PR #22532, opened 2022-02-16
resolved: Implement continuous mDNS querying as per RFC6762 5.2

vchillara:
Allow for mDNS service/domain/types browsing.
A client can connect to the backend via varlink and receive updates as the requested service becomes available.

The interval between the first two queries MUST be at least one second, the intervals between successive queries MUST increase by at least a factor of two.
When the interval between queries reaches or exceeds 60 minutes, a querier MAY cap the interval to a maximum of 60 minutes, and perform subsequent queries at a steady-state rate of one query per hour.

Cache maintenance performed by issuing queries at 80, 85, 90, 95% of the TTL, and updating/removing the records depending on replies.

TODO:
Improve the DNS transaction logic when multiple clients subscribe to the same service, ensuring that continuous queries are optimized.

PR #22751, opened 2022-03-15
resolved: remove entry from cache when goodbye packet received

vchillara:
ttl of an incoming mdns goodbye packet is updated to 1 as per RFC6762 10.1. However, in dns_cache_put_positive(), we check for min_ttl <= 0 to remove the record from cache.
As a result, the entries for which goodbye packets are received do not get removed from cache. changed the check to min_ttl <= 1 to account for goodbye packets.

### DIFF
--- a/src/resolve/resolved-dns-cache.c
+++ b/src/resolve/resolved-dns-cache.c
@@ -243,6 +243,22 @@ void dns_cache_prune(DnsCache *c) {
         }
 }
 
+bool dns_cache_expiry_in_one_second(DnsCache *c, usec_t t) {
+        DnsCacheItem *i;
+
+        assert(c);
+
+        /* Check if any items expire within the next second */
+        i = prioq_peek(c->by_expiry);
+        if (!i)
+                return false;
+
+        if (i->until <= usec_add(t, USEC_PER_SEC))
+                return true;
+
+        return false;
+}
+
 static int dns_cache_item_prioq_compare_func(const void *a, const void *b) {
         const DnsCacheItem *x = a, *y = b;
 
--- a/src/resolve/resolved-dns-cache.h
+++ b/src/resolve/resolved-dns-cache.h
@@ -58,3 +58,5 @@ bool dns_cache_is_empty(DnsCache *cache);
 unsigned dns_cache_size(DnsCache *cache);
 
 int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p, usec_t ts, unsigned max_rr);
+
+bool dns_cache_expiry_in_one_second(DnsCache *c, usec_t t);
--- a/src/resolve/resolved-mdns.c
+++ b/src/resolve/resolved-mdns.c
@@ -348,6 +348,33 @@ static int mdns_scope_process_query(DnsScope *s, DnsPacket *p) {
         return 0;
 }
 
+static int mdns_goodbye_callback(sd_event_source *s, uint64_t usec, void *userdata) {
+        DnsScope *scope = userdata;
+        int r;
+
+        assert(s);
+        assert(scope);
+
+        scope->mdns_goodbye_event_source = sd_event_source_disable_unref(scope->mdns_goodbye_event_source);
+
+        dns_cache_prune(&scope->cache);
+
+        if (dns_cache_expiry_in_one_second(&scope->cache, usec)) {
+                r = sd_event_add_time_relative(
+                        scope->manager->event,
+                        &scope->mdns_goodbye_event_source,
+                        CLOCK_BOOTTIME,
+                        USEC_PER_SEC,
+                        0,
+                        mdns_goodbye_callback,
+                        scope);
+                if (r < 0)
+                        return log_error_errno(r, "mDNS: Failed to re-schedule goodbye callback: %m");
+        }
+
+        return 0;
+}
+
 static int on_mdns_packet(sd_event_source *s, int fd, uint32_t revents, void *userdata) {
         _cleanup_(dns_packet_unrefp) DnsPacket *p = NULL;
         Manager *m = userdata;
@@ -406,6 +433,22 @@ static int on_mdns_packet(sd_event_source *s, int fd, uint32_t revents, void *us
                                 log_debug("Got a goodbye packet");
                                 /* See the section 10.1 of RFC6762 */
                                 rr->ttl = 1;
+
+                                /* Look at the cache 1 second later and remove stale entries.
+                                 * This is particularly useful to keep service browsers updated on service removal,
+                                 * as there are no other reliable triggers to propogate that info. */
+                                if (!scope->mdns_goodbye_event_source) {
+                                        r = sd_event_add_time_relative(
+                                                        scope->manager->event,
+                                                        &scope->mdns_goodbye_event_source,
+                                                        CLOCK_BOOTTIME,
+                                                        USEC_PER_SEC,
+                                                        0,
+                                                        mdns_goodbye_callback,
+                                                        scope);
+                                        if (r < 0)
+                                                return r;
+                                }
                         }
                 }
 
--- a/src/resolve/resolved-dns-scope.c
+++ b/src/resolve/resolved-dns-scope.c
@@ -41,6 +41,7 @@ int dns_scope_new(Manager *m, DnsScope **ret, Link *l, DnsProtocol protocol, int
                 .protocol = protocol,
                 .family = family,
                 .resend_timeout = MULTICAST_RESEND_TIMEOUT_MIN_USEC,
+                .mdns_goodbye_event_source = NULL,
         };
 
         if (protocol == DNS_PROTOCOL_DNS) {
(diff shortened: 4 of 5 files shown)


src/resolve/resolved-dns-cache.c
```diff
@@ -442,8 +442,9 @@ static int dns_cache_put_positive(
          * RRs are still valid, hence find the lowest here */
         min_ttl = MIN(dns_answer_min_ttl(answer), rr->ttl);
 
-        /* New TTL is 0? Delete this specific entry... */
-        if (min_ttl <= 0) {
+        /* New TTL is 0? Delete this specific entry...
+         * In RFC6762 10.1 for Good Bye packet TTL is 1 */
+        if (min_ttl <= 1) {
```
poettering: hmm, as I read the RFC we should instead when min_ttl is zero go through all our matching cached entries and set the stored TTL to 1s.

vchillara: The ttl of incoming goodbye packet records is already being set to 1 before "dns_cache_put_positive" is called (line 399, resolved-mdns.c).

However in dns_cache_put_positive() for cache deletion ttl is checked for 0 instead of 1.
Due to this, records with goodbye packets aren't getting removed from cache.

The RFC says to change the TTL of goodbye packet records to 1, then take a look at them again after a second and delete them if no updates were received.
The current implementation seems to be taking care of setting the TTL to 1, but it doesn't account for the deletion.

This PR takes care of deletion, but ideally according to the RFC we should be waiting for 1 second before doing so.

yuwata: How about the following? When `min_ttl == 1`, find matching existing positive cache entries and update its TTL and cache lifetime `DnsCacheItem.until`.

vchillara: It makes sense to update the ttl/until of existing matching cache entries to 1 second when we get a TTL==0.
However, there also needs to be some sort of callback after one second to check if the entry was updated in the meantime, and remove the entry if no update was received.
Without this, the entry lingers in the cache indefinitely until the cache is pruned the next time.
We stumbled into this behavior while implementing #22532, with the aim of notifying clients when a service expires.

src/resolve/resolved-mdns.c
```diff
@@ -397,6 +406,17 @@ static int on_mdns_packet(sd_event_source *s, int fd, uint32_t revents, void *us
                                 log_debug("Got a goodbye packet");
                                 /* See the section 10.1 of RFC6762 */
                                 rr->ttl = 1;
+
+                                r = sd_event_add_time_relative(
```
yuwata: It is not necessary to call `dns_cache_prune()` after 1 second. As `dns_cache_prune()` is always called before calling `dns_cache_lookup()` is called. That is, the cache entry should be removed when we search cached entry.

Now I am confused, why do you want to explicitly remove the cache entry from the cache??
What issue do you want to solve??

vchillara: dns_cache_lookup() is only called when a packet is sent, which happens indeterminately.
RFC 6762 section 10.1 does mention that the entry should be deleted form the cache 1 second later.
This helps in keeping the cache up-to-date.
Without this, the entry would linger in cache indefinitely until the next packet is sent.
It also helps to notify the client immediately when a service sends a goodbye packet and expires (#22532).

yuwata: "the entry should be deleted form the cache 1 second later" does not mean we really need to remove the entry after 1 sec. But it is enough to drop the entry when it is requested later than 1sec.

In #22532, please call `dns_cache_prune()` before `dns_cache_lookup()` in `mdns_subscriber_lookup_cache()`.

vchillara: Since goodbye packets are unsolicited, and this logic is actually independent of the RFC6762 section 5.2 stuff implemented in #22532, I figured it would make sense to add here.

poettering:
hmm, what i don't get: why have a timer to remove these entries from the cache at all? it should suffice to clean this up when we query the cache, and encounter an entry which is now run over. we generally only clean the cache when using it, but not independently by timer. so why should we do so here?

patching the ttl to 1 should mean that if we then query the cache it the record will be used still within that 1s but once the 1s is hit it won't anymore, even if it still exists in the cache.

what am i missing?

vchillara:
@poettering thanks for the review. Working on addressing your feedback.
Regarding your question - does https://github.com/systemd/systemd/pull/22751#discussion_r908131303 answer it?

poettering:
> @poettering thanks for the review. Working on addressing your feedback. Regarding your question - does [#22751 (comment)](https://github.com/systemd/systemd/pull/22751#discussion_r908131303) answer it?

Not really. Why would it be beneficial to clean up the cache immediately when the timeout runs out? sounds much better to just do so when we deal with the cache anyway. i.e. usually it makes sense to avoid wakeups for nothing, and instead do all work when we are awake anyway...

vchillara:
> > @poettering thanks for the review. Working on addressing your feedback. Regarding your question - does [#22751 (comment)](https://github.com/systemd/systemd/pull/22751#discussion_r908131303) answer it?
> 
> Not really. Why would it be beneficial to clean up the cache immediately when the timeout runs out? sounds much better to just do so when we deal with the cache anyway. i.e. usually it makes sense to avoid wakeups for nothing, and instead do all work when we are awake anyway...

I see your point about avoiding unnecessary wakeups, but in the context of browsing services, we would like to keep the client up to date. Since we notify them on addition of services, we also want to notify them when a service expires.
Besides, the logic of removing cache entries with a timer would only be active while there is an active subscriber. 
What do you think?

vchillara:
@poettering Added a comment clarifying the need for this.
One thing - in some of the earlier comments on this PR, there was a request to move this functionality to #22532, which we did.
So the question is do we now want it here or in the other PR?

src/resolve/resolved-mdns.c
```diff
@@ -397,6 +406,20 @@ static int on_mdns_packet(sd_event_source *s, int fd, uint32_t revents, void *us
                                 log_debug("Got a goodbye packet");
                                 /* See the section 10.1 of RFC6762 */
                                 rr->ttl = 1;
+
+                                /* Look at the cache 1 second later and remove stale entries.
+                                 * This is particularly useful to keep service browsers updated on service removal,
+                                 * as there are no other reliable triggers to propogate that info. */
+                                r = sd_event_add_time_relative(
```
poettering: so this will create one event source for each incoming mdns packet with ttl set to 0. that is a chance for a DoS, as clients can make us allocate basically unbounded memory, if they send us a ton of these packets within 1s.

Thus the event source object should be stored somewhere (in the scope object?), and if it already exists no new event source should be allocated. morevoer, the handler should check if there are RRs in the cache that should be flushed within less than 1s, and if so should immediately requeue a new event source to cover for that.

That way there's only one event source enqueued at max.

src/resolve/resolved-mdns.c
```diff
@@ -397,6 +406,20 @@ static int on_mdns_packet(sd_event_source *s, int fd, uint32_t revents, void *us
...
+                                                m->event,
+                                                NULL,
+                                                CLOCK_BOOTTIME,
+                                                USEC_PER_SEC,
+                                                0,
+                                                mdns_goodbye_callback,
+                                                scope);
```
poettering: this will result in memory corruption: let's say a scope goes away (because a network interface goes away), but the event source is still queued. it will be called pointing to a scope that doesn't exist anymore. Bad things will then happen.

if you embedd a pointer to the eent source in DnsScope object, then you can also delete the event source whenever the scope goes away.

src/resolve/resolved-dns-cache.h
```diff
@@ -58,3 +58,5 @@ bool dns_cache_is_empty(DnsCache *cache);
 unsigned dns_cache_size(DnsCache *cache);
 
 int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p, usec_t ts, unsigned max_rr);
+
+bool dns_cache_expiry_in_one_second(DnsCache *c, usec_t t);
```
yuwata: Missing line break at the end of the file.

src/resolve/resolved-mdns.c
```diff
@@ -406,6 +433,22 @@ static int on_mdns_packet(sd_event_source *s, int fd, uint32_t revents, void *us
                                 log_debug("Got a goodbye packet");
                                 /* See the section 10.1 of RFC6762 */
                                 rr->ttl = 1;
+
+                                /* Look at the cache 1 second later and remove stale entries.
+                                 * This is particularly useful to keep service browsers updated on service removal,
+                                 * as there are no other reliable triggers to propogate that info. */
+                                if (!scope->mdns_goodbye_event_source) {
+                                        r = sd_event_add_time_relative(
```
yuwata, line 441: Still I cannot follow why we need to actively clear the outdated cache entries here...

yuwata (changes requested):

yuwata:
-> #30096 ?

src/resolve/resolved-mdns.c
vchillara: In the context of continuous browsing (#22532), it would be useful to actively clear the outdated cache entries and notify the browser client(s).
To accomplish that, we would eventually add a call to mdns_notify_browsers_unsolicited_updates()
from mdns_goodbye_callback().

PR #22753, opened 2022-03-15
resolved: make maximum transaction attempts 1 for non-probe MDNS requests

vchillara:
Maximum attempts to send MDNS requests is one except for probe requests. See RFC 6762 Section 8.1
However, TRANSACTION_ATTEMPTS_MAX macro does not account for probe requests.

### DIFF
--- a/src/resolve/resolved-dns-transaction.h
+++ b/src/resolve/resolved-dns-transaction.h
@@ -214,11 +214,31 @@ DnsTransactionSource dns_transaction_source_from_string(const char *s) _pure_;
 /* Maximum attempts to send LLMNR requests, see RFC 4795 Section 2.7 */
 #define LLMNR_TRANSACTION_ATTEMPTS_MAX 3
 
-/* Maximum attempts to send MDNS requests, see RFC 6762 Section 8.1 */
-#define MDNS_TRANSACTION_ATTEMPTS_MAX 3
+/* Maximum attempts to send MDNS requests is one except for probe requests, see RFC 6762 Section 8.1
+ * RFC 6762 differentiates between normal (single-shot/continuous) and probe requests.
+ * It therefore makes sense to attempt each normal query only once with no retries.
+ * Otherwise we'd be sending out three attempts for even a normal query. */
+#define MDNS_TRANSACTION_ATTEMPTS_MAX 1
 
-#define TRANSACTION_ATTEMPTS_MAX(p) (((p) == DNS_PROTOCOL_LLMNR) ? \
-                                         LLMNR_TRANSACTION_ATTEMPTS_MAX : \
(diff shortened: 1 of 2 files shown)


src/resolve/resolved-dns-transaction.h
```diff
@@ -214,11 +214,28 @@ DnsTransactionSource dns_transaction_source_from_string(const char *s) _pure_;
...
-
-#define TRANSACTION_ATTEMPTS_MAX(p) (((p) == DNS_PROTOCOL_LLMNR) ? \
-                                         LLMNR_TRANSACTION_ATTEMPTS_MAX : \
-                                         (((p) == DNS_PROTOCOL_MDNS) ? \
-                                             MDNS_TRANSACTION_ATTEMPTS_MAX : \
-                                             DNS_TRANSACTION_ATTEMPTS_MAX))
+/* Maximum attempts to send MDNS requests is one except for probe requests, see RFC 6762 Section 8.1 */
+#define MDNS_TRANSACTION_ATTEMPTS_MAX 1
+
+#define MDNS_PROBE_TRANSACTION_ATTEMPTS_MAX 3
```
yuwata: RFC 6762 Section 8.1 says about probe requests. Where can I find that other requests should be tried only once?
Could you update comment more, e.g. quoting a relevant sentence in the RFC?

yuwata (changes requested):

src/resolve/resolved-dns-transaction.h
vchillara: The RFC doesn't explicitly state that a non-probe mDNS query should only be attempted once.
However, since RFC 6762 differentiates between normal (single-shot/continuous) and probe requests, it probably makes sense to fire each normal query only once with no retries.
Otherwise we'd be sending out three attempts for a normal query as well.
The exception of course being probing, which should be attempted a max of 3 times as described in section 8.1.
This also messes with the timing of continuous querying in #22532.
Avahi's implementation is also in line with this.

vchillara: This blurb in section 8.1 of RFC6762 also points out the difference in timing restrictions for probe and regular queries.

> Note that probing is the one exception from the normal rule that there should be at least one
   second between repetitions of the same question, and the interval between subsequent repetitions
   should at least double.

yuwata: I see. That part indeed implies the regular queries should be tried once. Could you update the comment in the code with quoting the above sentences?

vchillara: Updated the comment in the code.

evverx:
Just a heads-up, looks this PR will be reverted in https://github.com/systemd/systemd/pull/23875.

Issue #22549, opened 2022-02-17
RFE: Add % specifier that resolves to /run/credentials


flokli:
**Is your feature request related to a problem? Please describe.**

Most processes can be configured to read credentials from from a path described in an environment variable, but there's no way right now to assemble these paths from systemd, without wrapper scripts.

This means, if you want to run an application with SOME_SECRET_LOCATION= env var, which makes use of systemd service credentials, you need an intermediate wrapper script doing the interpolation and then invoking the real service.

**Describe the solution you'd like**

It'd be very convenient if systemd would support substituting `$CREDENTIALS_DIRECTORY` in `Environment=` directives.

The canonical way to do this seems to be Specifiers (from `systemd.unit(5)`), as we don't do variable interpolation there.

With this, something like the following would be possible without any wrapper scripts:

```ini
[Unit]
…
[Service]
ExecStart=/path/to/binary
LoadCredential=google_application_credentials:/run/secrets/google_application_credentials
Environment="GOOGLE_APPLICATION_CREDENTIALS=%c/google_application_credentials"
```

**Describe alternatives you've considered**

Sticking with the wrapper scripts

**The systemd version you checked that didn't have the feature you are asking for**
230

flokli:
I realized we already do interpolate `$CREDENTIALS_DIRECTORY` in `ExecStart=` - so restricted this proposal to environment variables only.

poettering:
We do not resolve env vars in env vars, sorry. We are not a shell. If you want a shell, use a shell.

The only thing we do is resolve them in a very minimalist way in Exec*=, at the moment of invocation. And I must say I kinda regret having added that too.

We really want Unit files to be simple, so that not just our tools can parse them. i.e. not a complex pattern resolution language, but just simple linear assignments. Doing variable expansion inside of variables (and thus requiring ordering, or dealing with recursions) is something we really don't want to do, as it turns units into a form of shell scripts, and that's a bad idea, you can then just use shell scripts in the first place.

flokli:
Nope, I'm not asking to resolve env vars in env vars.

This is about having a specifier for the credentials directory (`%c`) also holding the contents of `$CREDENTIALS_DIRECTORY`.

Similar to how we already have `%S` for `$STATE_DIRECTORY`, `%t` for`$RUNTIME_DIRECTORY`, `%C` for `$CACHE_DIRECTORY` and `%L` for `$LOGS_DIRECTORY`.

PR #22754, opened 2022-03-15
core: add %d specifier for the $CREDENTIALS_DIRECTORY

mrc0mmand:
Resolves: #22549

### DIFF
--- a/man/systemd.unit.xml
+++ b/man/systemd.unit.xml
@@ -1976,6 +1976,11 @@
             <entry>This is either <filename>/var/cache</filename> (for the system manager) or the path <literal>$XDG_CACHE_HOME</literal> resolves to (for user managers).</entry>
           </row>
           <row>
+            <entry><literal>%d</literal></entry>
+            <entry>Credentials directory</entry>
+            <entry>This is the value of the <literal>$CREDENTIALS_DIRECTORY</literal> environment variable if available. See section "Credentials" in <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry> for more information.</entry>
+          </row>
+          <row>
             <entry><literal>%E</literal></entry>
             <entry>Configuration directory root</entry>
             <entry>This is either <filename>/etc/</filename> (for the system manager) or the path <literal>$XDG_CONFIG_HOME</literal> resolves to (for user managers).</entry>
--- a/src/core/unit-printf.c
+++ b/src/core/unit-printf.c
@@ -148,6 +148,20 @@ static int specifier_special_directory(char specifier, const void *data, const c
         return 0;
 }
 
+static int specifier_credentials_dir(char specifier, const void *data, const char *root, const void *userdata, char **ret) {
+        const Unit *u = ASSERT_PTR(userdata);
+        char *d;
+
+        assert(ret);
+
+        d = strjoin(u->manager->prefix[EXEC_DIRECTORY_RUNTIME], "/credentials/", u->id);
+        if (!d)
+                return -ENOMEM;
+
+        *ret = d;
+        return 0;
+}
+
 int unit_name_printf(const Unit *u, const char* format, char **ret) {
         /*
          * This will use the passed string as format string and replace the following specifiers (which should all be
@@ -191,6 +205,7 @@ int unit_full_printf_full(const Unit *u, const char *format, size_t max_length, 
          * %R: the root of this systemd's instance tree (deprecated)
          *
          * %C: the cache directory root (e.g. /var/cache or $XDG_CACHE_HOME)
+         * %d: the credentials directory ($CREDENTIALS_DIRECTORY)
          * %E: the configuration directory root (e.g. /etc or $XDG_CONFIG_HOME)
          * %L: the log directory root (e.g. /var/log or $XDG_CONFIG_HOME/log)
          * %S: the state directory root (e.g. /var/lib or $XDG_CONFIG_HOME)
@@ -227,6 +242,7 @@ int unit_full_printf_full(const Unit *u, const char *format, size_t max_length, 
                 { 'R', specifier_cgroup_root,              NULL },
 
                 { 'C', specifier_special_directory,        UINT_TO_PTR(EXEC_DIRECTORY_CACHE) },
+                { 'd', specifier_credentials_dir,          NULL },
                 { 'E', specifier_special_directory,        UINT_TO_PTR(EXEC_DIRECTORY_CONFIGURATION) },
                 { 'L', specifier_special_directory,        UINT_TO_PTR(EXEC_DIRECTORY_LOGS) },
                 { 'S', specifier_special_directory,        UINT_TO_PTR(EXEC_DIRECTORY_STATE) },
--- a/src/test/test-execute.c
+++ b/src/test/test-execute.c
@@ -1086,6 +1086,7 @@ static void test_exec_specifier(Manager *m) {
         test(m, "exec-specifier.service", 0, CLD_EXITED);
         test(m, "exec-specifier@foo-bar.service", 0, CLD_EXITED);
         test(m, "exec-specifier-interpolation.service", 0, CLD_EXITED);
+        test(m, "exec-specifier-credentials-dir.service", 0, CLD_EXITED);
 }
 
 static void test_exec_standardinput(Manager *m) {
--- a/test/test-execute/exec-specifier-credentials-dir.service
+++ b/test/test-execute/exec-specifier-credentials-dir.service
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=Test for specifiers
+
+[Service]
+Type=oneshot
+Environment=TOP_SECRET=%d/very_top_secret
+# Test if the specifier is resolved correctly both before and after LoadCredential=
+ExecStart=test %d/very_top_secret = "${CREDENTIALS_DIRECTORY}/very_top_secret"
+LoadCredential=very_top_secret
+ExecStart=test %d/very_top_secret = "${CREDENTIALS_DIRECTORY}/very_top_secret"
+ExecStart=sh -c 'test %d/very_top_secret = "$TOP_SECRET"'
--- a/test/test-execute/exec-specifier.service
+++ b/test/test-execute/exec-specifier.service
@@ -20,6 +20,7 @@ ExecStart=test %L = /var/log
 ExecStart=test %E = /etc
 ExecStart=test %T = /tmp
 ExecStart=test %V = /var/tmp
+ExecStart=test %d = %t/credentials/%n
 ExecStart=sh -c 'test %u = $$(id -un)'
 ExecStart=sh -c 'test %U = $$(id -u)'
 ExecStart=sh -c 'test %g = $$(id -gn)'


src/shared/specifier.c
```diff
@@ -301,6 +302,27 @@ int specifier_os_image_version(char specifier, const void *data, const char *roo
...
 }
 
+int specifier_credentials_dir(char specifier, const void *data, const char *root, const void *userdata, char **ret) {
+        const char *d;
+        char *copy;
+        int r;
+
+        assert(ret);
+
+        r = get_credentials_dir(&d);
```
yuwata: IIUC, this reads values from the `CREDENTIALS_DIRECTORY` environment variable, but at the time the unit is loaded, the env var is not set. Hence, I guess this specifier always provides an empty string `""`.

yuwata (changes requested):

poettering:
hmm, this doesn't look right. get_credentials_dir() is something you'd call from service code that reads the env var set by the manager. But the specifier here is something that is resolved in the manager itself, not by the service. Thus it is not the right thing to call.

not that %S, %t, %C and so on all resolve to the *root* where the relevant dirs are placed. The same should apply when we add %c, i.e. the dir we place the per-unit creds dirs in, and not the per-unit creds-dir in full. See build_environment() in src/core/execute.c to see how the credentials dir is currently put together.

or in other words, this should resolve to something like `u->manager->prefix[EXEC_DIRECTORY_RUNTIME]` suffixed with `/credentials`

src/core/unit-printf.c
```diff
@@ -148,6 +148,28 @@ static int specifier_special_directory(char specifier, const void *data, const c
...
+        ec = unit_get_exec_context(u);
+        assert(ec);
+
+        if ((!hashmap_isempty(ec->load_credentials) || !hashmap_isempty(ec->set_credentials)) &&
+            u->manager->prefix[EXEC_DIRECTORY_RUNTIME]) {
+                d = strjoin(u->manager->prefix[EXEC_DIRECTORY_RUNTIME], "/credentials/", u->id);
+                if (!d)
+                        return -ENOMEM;
+        } else
+                d = strdup("");
```
yuwata: No oom check?

mrc0mmand: ouch, thanks!

src/core/unit-printf.c
```diff
@@ -148,6 +148,29 @@ static int specifier_special_directory(char specifier, const void *data, const c
         return 0;
 }
 
+static int specifier_credentials_dir(char specifier, const void *data, const char *root, const void *userdata, char **ret) {
+        const Unit *u = ASSERT_PTR(userdata);
+        const ExecContext *ec = NULL;
+        char *d = NULL;
```
yuwata: `ec` and `d` are not necessary to be initialized.

Also, `n` in the above `specifier_special_directory()` is not necessary to be initialized.

src/core/unit-printf.c
```diff
@@ -148,6 +148,29 @@ static int specifier_special_directory(char specifier, const void *data, const c
...
+
+        assert(ret);
+
+        ec = unit_get_exec_context(u);
+        assert(ec);
```
yuwata: Nowadays, we has `ASSERT_PTR()`, hence
```
ec = ASSERT_PTR(unit_get_exec_context(u));
```
should work fine.

src/core/unit-printf.c
```diff
@@ -148,6 +148,29 @@ static int specifier_special_directory(char specifier, const void *data, const c
...
+
+        if ((!hashmap_isempty(ec->load_credentials) || !hashmap_isempty(ec->set_credentials)) &&
+            u->manager->prefix[EXEC_DIRECTORY_RUNTIME])
+                d = strjoin(u->manager->prefix[EXEC_DIRECTORY_RUNTIME], "/credentials/", u->id);
+        else
+                d = strdup("");
+
+        if (!d)
```
yuwata, line 158: We usually do not insert any blank lines above `if (!d)`.

mrc0mmand:
Oh well, looks like the ExecContext can be `NULL` (at least when running fuzzers):

```
Assertion '_expr_' failed at src/core/unit-printf.c:158, function int specifier_credentials_dir(char, const void *, const char *, const void *, char **)(). Aborting.
==459== ERROR: libFuzzer: deadly signal
    #0 0x42c714 in __sanitizer_print_stack_trace /src/llvm-project/compiler-rt/lib/ubsan/ubsan_diag_standalone.cpp:31:3
    #1 0x47d8c8 in fuzzer::PrintStackTrace() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerUtil.cpp:210:5
    #2 0x462d23 in fuzzer::Fuzzer::CrashCallback() /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:233:3
    #3 0x7f22184213bf  (/lib/x86_64-linux-gnu/libpthread.so.0+0x143bf)
    #4 0x7f221823103a in raise (/lib/x86_64-linux-gnu/libc.so.6+0x4303a)
    #5 0x7f2218210858 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x22858)
    #6 0x7f2218c41d44 in log_assert_failed /work/build/../../src/systemd/src/basic/log.c:866:9
    #7 0x7f2219725b67 in specifier_credentials_dir /work/build/../../src/systemd/src/core/unit-printf.c:158:14
    #8 0x7f2218b83923 in specifier_printf /work/build/../../src/systemd/src/shared/specifier.c:69:45
2022-03-15 19:49:56,492 - root - INFO - Fuzzer: fuzz-unit-file. Detected bug.
2022-03-15 19:49:56,493 - root - INFO - Trying to reproduce crash using: /tmp/tmpkzv7t54m/crash-897d87b3cdd36e10ea01e079b4f006b9917dbfc9.
    #9 0x7f2219724937 in unit_full_printf_full /work/build/../../src/systemd/src/core/unit-printf.c:269:16
    #10 0x7f221960b2dd in unit_path_printf /work/build/../../src/systemd/src/core/unit-printf.h:16:16
    #11 0x7f221960b2dd in config_parse_path_spec /work/build/../../src/systemd/src/core/load-fragment.c:2225:13
    #12 0x7f2218a6cd8d in next_assignment /work/build/../../src/systemd/src/shared/conf-parser.c:149:24
    #13 0x7f2218a6cd8d in parse_line /work/build/../../src/systemd/src/shared/conf-parser.c:256:16
    #14 0x7f2218a6bd2c in config_parse /work/build/../../src/systemd/src/shared/conf-parser.c:399:21
    #15 0x42e205 in LLVMFuzzerTestOneInput /work/build/../../src/systemd/src/core/fuzz-unit-file.c:70:16
    #16 0x4642d2 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:611:15
    #17 0x463aba in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:514:3
    #18 0x465924 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:826:7
    #19 0x465b59 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile> >&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:857:3
    #20 0x4555d0 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:912:6
    #21 0x47e082 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
    #22 0x7f22182120b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x240b2)
    #23 0x407c1d in _start (build-out/fuzz-unit-file+0x407c1d)
```

so I dropped the asssert and moved the check to the condition below.

I wonder if it would make sense to move the context helpers from [execute.c](https://github.com/systemd/systemd/blob/main/src/core/execute.c#L1418-L1472) to `unit.c` and make them public (or at least make them public in `execute.[ch]`), so we could "prettify" the condition a bit (using `ec && exec_context_has_credentials(ec) ...`.

man/systemd.unit.xml
```diff
@@ -1975,6 +1975,11 @@
             <entry>Cache directory root</entry>
             <entry>This is either <filename>/var/cache</filename> (for the system manager) or the path <literal>$XDG_CACHE_HOME</literal> resolves to (for user managers).</entry>
           </row>
+          <row>
+            <entry><literal>%d</literal></entry>
+            <entry>Credentials directory</entry>
+            <entry>This is the value of the <literal>$CREDENTIALS_DIRECTORY</literal> environment variable if available. See section "Credentials" in <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry> for more information.</entry>
```
poettering, line 1981: that's not quite the case. Its the directory *root*, i.e. the unit name has to be suffixed to get the per-unit dir.

mrc0mmand: Is it? We explicitly concatenate it with the unit name in both cases iiuc:

https://github.com/systemd/systemd/blob/c8a4306e63a5387813e6e01164e5303e28dc60dc/src/core/execute.c#L1984

poettering: yes, that's what I am saying: `%d` should map to `p->prefix[EXEC_DIRECTORY_RUNTIME]` suffixed with `/credentials`, but *without* the `u->id` additional suffix.

Thus, i'd call this a "root" directory, because in it you find one subdir for each service.

poettering: or hmm, maybe the way you defined it actually does make more sense.

poettering:
Hmm, so I think I agree with your way to define it after all, i.e. not the root, but the dir itself. Given we can have exactly zero or one credential dirs per user it's safe to do it that way. 

However, the conditoinalization on ExecContext settings data must be dropped. First, I don't think the empty string really is better than a path that doesn't exist. i.e. if we really have a service for which people use the new specifier but where credentials are not used, then the error path for handling a nonexisting dir below /run/credentials/ is likely better htan the error path for handling the empty path as dir. 

More importantly though: specifier expansion takes place at the time of parsing. This means the strings specifiers expand to should *not* be dependent on other settings from the same file. Because if they would then suddenly the order of assignment would start to matter. i.e. consider this:

```ini
[Service]
SetCredential=foo:bar
ExecStart=/bin/echo %d
```

vs. 

```ini
[Service]
ExecStart=/bin/echo %d
SetCredential=foo:bar
``` 

these two units should really have the same effect. But in your current patch they won't. Because in the second case we'll parse ExecStart= first, and then do specifier expansion at a moment where we don't know anything about the credentials, thus resolve things to "". But in the first case we'll properly resolve things to /run/credentials/<unitname>.

hence: simplify htis, just drop the conditionalization on the execcontext. It's simpler, more robust and independent of ine order then.

mrc0mmand:
> Hmm, so I think I agree with your way to define it after all, i.e. not the root, but the dir itself. Given we can have exactly zero or one credential dirs per user it's safe to do it that way.
> 
> However, the conditoinalization on ExecContext settings data must be dropped. First, I don't think the empty string really is better than a path that doesn't exist. i.e. if we really have a service for which people use the new specifier but where credentials are not used, then the error path for handling a nonexisting dir below /run/credentials/ is likely better htan the error path for handling the empty path as dir.
> 
> More importantly though: specifier expansion takes place at the time of parsing. This means the strings specifiers expand to should _not_ be dependent on other settings from the same file. Because if they would then suddenly the order of assignment would start to matter. i.e. consider this:

Ah, thanks, this makes the ExecContext stuff much clearer.

> 
> ```ini
> [Service]
> SetCredential=foo:bar
> ExecStart=/bin/echo %d
> ```
> 
> vs.
> 
> ```ini
> [Service]
> ExecStart=/bin/echo %d
> SetCredential=foo:bar
> ```
> 
> these two units should really have the same effect. But in your current patch they won't. Because in the second case we'll parse ExecStart= first, and then do specifier expansion at a moment where we don't know anything about the credentials, thus resolve things to "". But in the first case we'll properly resolve things to /run/credentials/.
> 
> hence: simplify htis, just drop the conditionalization on the execcontext. It's simpler, more robust and independent of ine order then.

Agreed. Again, thanks for the hints!

Issue #18077, opened 2020-12-24
RFE: support for long unit names


DemiMarie:
**Is your feature request related to a problem? Please describe.**
systemd currently has a hard limit of 255 bytes for the length of a unit name.  This causes problems such as #17684.

**Describe the solution you'd like**
Longer unit names are supported.  This can be done by hashing them before converting them to a filename.

**Describe alternatives you've considered**
Long mount point paths could be handled differently.  For instance, they could be symlinked to somewhere under `/run`.

poettering:
Unsetting milestone again. I think #17684 + #16161 should be enough to have on the milestone.

PR #22759, opened 2022-03-15
Create "hashed" unit names from long paths

msekletar:
Fixes https://github.com/systemd/systemd/issues/18077.

This will allow systemd to manage units (mounts, devices, swaps) that are created from external sources (e.g. /proc/swaps) and we couldn't represent them so far.

### DIFF
--- a/test/units/testsuite-64.sh
+++ b/test/units/testsuite-64.sh
@@ -515,6 +515,7 @@ testcase_long_sysfs_path() {
     echo "UUID=deadbeef-dead-dead-beef-222222222222 $mpoint ext4 defaults 0 0" >>/etc/fstab
     systemctl daemon-reload
     mount "$mpoint"
+    systemctl status "$mpoint"
     test -e "$mpoint/test"
     umount "$mpoint"
 
@@ -525,9 +526,9 @@ testcase_long_sysfs_path() {
     udevadm settle
 
     logfile="$(mktemp)"
-    journalctl -b -q --no-pager -o short-monotonic -p info --grep "Device path.*vda.?' too long to fit into unit name"
+    [[ "$(journalctl -b -q --no-pager -o short-monotonic -p info --grep "Device path.*vda.?' too long to fit into unit name" | wc -l)" -eq 0 ]]
     # Make sure we don't unnecessarily spam the log
-    journalctl -b -q --no-pager -o short-monotonic -p info --grep "/sys/devices/.+/vda[0-9]?" _PID=1 + UNIT=systemd-udevd.service | tee "$logfile"
+    { journalctl -b -q --no-pager -o short-monotonic -p info --grep "/sys/devices/.+/vda[0-9]?" _PID=1 + UNIT=systemd-udevd.service || :;} | tee "$logfile"
     [[ "$(wc -l <"$logfile")" -lt 10 ]]
 
     : >/etc/fstab


mrc0mmand:
```
==11058==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7f865b801a31 at pc 0x7f865f52af68 bp 0x7ffcc6ce5da0 sp 0x7ffcc6ce5550
WRITE of size 9 at 0x7f865b801a31 thread T0
    #0 0x7f865f52af67 in __interceptor_vsprintf (/usr/lib/clang/13.0.1/lib/linux/libclang_rt.asan-x86_64.so+0xcaf67)
    #1 0x7f865f52b1ff in __interceptor_sprintf (/usr/lib/clang/13.0.1/lib/linux/libclang_rt.asan-x86_64.so+0xcb1ff)
    #2 0x7f865e28a7ef in unit_name_hash_long /systemd-meson-build/../build/src/basic/unit-name.c:534:17
    #3 0x7f865e28acd7 in unit_name_from_path /systemd-meson-build/../build/src/basic/unit-name.c:572:21
    #4 0x557492b51967 in test_unit_name_from_path_one /systemd-meson-build/../build/src/test/test-unit-name.c:111:9
    #5 0x557492b42caa in test_unit_name_from_path /systemd-meson-build/../build/src/test/test-unit-name.c:138:9
    #6 0x557492b50dee in run_test_table /systemd-meson-build/../build/src/shared/tests.h:104:33
    #7 0x557492b50dee in main /systemd-meson-build/../build/src/test/test-unit-name.c:867:1
    #8 0x7f865d72d30f in __libc_start_call_main libc-start.c
    #9 0x7f865d72d3c0 in __libc_start_main@GLIBC_2.2.5 (/usr/lib/libc.so.6+0x2d3c0)
    #10 0x557492b424f4 in _start (/systemd-meson-build/test-unit-name+0x254f4)

Address 0x7f865b801a31 is located in stack of thread T0 at offset 49 in frame
    #0 0x7f865e28a59f in unit_name_hash_long /systemd-meson-build/../build/src/basic/unit-name.c:512

  This frame has 3 object(s):
    [32, 49) 'hash' (line 514) <== Memory access at offset 49 overflows this variable
    [96, 352) 'n' (line 514)
    [416, 432) 'k' (line 516)
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow (/usr/lib/clang/13.0.1/lib/linux/libclang_rt.asan-x86_64.so+0xcaf67) in __interceptor_vsprintf
```

mrc0mmand:
GH Webhooks are currently FUBAR, so no CIs getting triggered (https://www.githubstatus.com/). So I guess we'll have to wait until the issue is resolved and then do another force-push.

test/units/testsuite-70.service
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=TEST-70-LONG-UNIT-NAMES
```
bluca: There's a runtime cost in setting up separate test cases, so it would be better to fold this test into an existing one

src/basic/unit-name.c
```diff
@@ -507,6 +509,42 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        suffix = strrchr(name, '.');
+        if (!suffix)
+                return -EINVAL;
+
+        if (unit_type_from_string(suffix+1) < 0)
+                return -EINVAL;
+
+        /* We intentionally hash with k that is all zeros because each time we want the same hash value for the same string. */
+        h = siphash24_string(name, k);
```
anitazha: Maybe define a static `k` declared using `SD_ID128_MAKE()`? It's done in https://github.com/systemd/systemd/blob/main/src/udev/udev-node.c#L256

src/basic/unit-name.c
```diff
@@ -507,6 +509,42 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        for (i = 0; i < 8; i++) {
+                uint64_t m = (uint64_t) 0xff << 8 * i;
+                sprintf(&hash[i*2], "%02hhx", (char) ((h & m) >> 8*i));
+        }
```
poettering: hexmem()?

src/basic/unit-name.c
```diff
@@ -507,6 +509,42 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        stpcpy(stpcpy(stpcpy(mempcpy(n, name, UNIT_NAME_MAX - 1 - strlen(suffix+1) - sizeof(hash) - 1 - 3), "..."), hash), suffix);
```
poettering: strjoin()?

poettering: i think adding a marker for hashed names makes sense, but i'd probably use something shorter, maybe just "_"?

src/basic/unit-name.c
```diff
@@ -507,6 +509,42 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        s = strdup(n);
+        if (!s)
+                return -ENOMEM;
+
+        *ret = TAKE_PTR(s);
+
+        return 0;
```
poettering, line 577: I understand you wanted to minimize memory allocations here, but I wouldn't bother tbh, this is a corner case after all...

src/basic/unit-name.c
```diff
@@ -507,6 +509,42 @@ int unit_name_template(const char *f, char **ret) {
...
 
+int unit_name_hash_long(const char *name, char **ret) {
+        _cleanup_free_ char *s = NULL;
+        char hash[17] = {}, n[UNIT_NAME_MAX] = {}, *suffix;
+        uint64_t h;
+        uint8_t k[16] = {};
+        int i;
+
+        if (strlen(name) < UNIT_NAME_MAX)
+                return 0;
```
poettering: as per coding style, our functions should initialize all return values on success. here you return success, and thus should initialize `*ret`, probably just to `NULL`

src/basic/unit-name.c
```diff
@@ -526,8 +564,18 @@ int unit_name_from_path(const char *path, const char *suffix, char **ret) {
...
+                _cleanup_free_ char *n = NULL;
+
+                log_debug("Unit name too long, falling back to hashed unit name.");
+
+                r = unit_name_hash_long(s, &n);
+                if (r < 0)
+                        return r;
+
+                s = mfree(s);
+                s = TAKE_PTR(n);
```
poettering:
```c
free_and_replace(s, n);
```

src/basic/unit-name.c
```diff
@@ -581,6 +629,9 @@ int unit_name_to_path(const char *name, char **ret) {
         if (r < 0)
                 return r;
 
+        if (strstr(prefix, "..."))
```
poettering: i'd prefer a stricter check here. i.e. add a helper is_unit_name_hashed() or so, that actually checks for the marker and the right number of hex chars

src/test/test-unit-name.c
```diff
@@ -106,14 +106,21 @@ TEST(unit_name_replace_instance) {
...
         puts(strna(t));
         assert_se(streq_ptr(t, expected));
 
         if (t) {
                 _cleanup_free_ char *k = NULL;
-                assert_se(unit_name_to_path(t, &k) == 0);
+
+                /* Unit names that contain "..." in the prefix are created from paths that would otherwise produce unit names that are too long. */
+                r = unit_name_to_path(t, &k);
+                assert_se(r == 0 || r == -ENAMETOOLONG);
```
poettering: move this assert_se() line after the if check below, and you don't have to check for ENAMETOOLONG twice

test/TEST-70-LONG-UNIT-NAMES/Makefile
```diff
@@ -0,0 +1 @@
+../TEST-01-BASIC/Makefile
```
poettering: please use 71, as the sysupdate PR already uses 70, and will be merged shortly

src/basic/unit-name.c
```diff
@@ -507,6 +509,42 @@ int unit_name_template(const char *f, char **ret) {
...
+        s = strdup(n);
+        if (!s)
+                return -ENOMEM;
+
+        *ret = TAKE_PTR(s);
+
+        return 0;
+}
+
 int unit_name_from_path(const char *path, const char *suffix, char **ret) {
```
poettering, line 580: unit_name_from_path_instance() should get the same treatment

src/basic/unit-name.c
```diff
@@ -581,6 +629,9 @@ int unit_name_to_path(const char *name, char **ret) {
         if (r < 0)
                 return r;
 
+        if (strstr(prefix, "..."))
+                return -ENAMETOOLONG;
+
         return unit_name_path_unescape(prefix, ret);
 }
```
poettering, line 667: the various callers of unit_name_to_path() should probaby be updated to handle this error nicely in some way.

poettering (changes requested):
looks good, but some points

test/units/testsuite-70.service
msekletar: Sure, there is a cost but isn't it wort paying for the gained clarity of test results you get by having separate test? I'd fully agree with you if we had a test framework where you could clearly denote that this is separate test case and it's result would be clearly visible and distinct from all other test cases, but we don't have that.

At any rate, I am happy to fold this into some existing test if you think that what I wrote above is a trade-off we want (and to some degree encourage).

bluca: it's worth it when there's a clearly separate component, or a clearly different setup - but there's plenty of tests already running checks on services and units

msekletar: Any idea which test should we merge this into? Maybe make it part of the storage test-suite in `TEST-64-UDEV-STORAGE`?

bluca: TEST-60-MOUNT-RATELIMIT already checks other corner cases of mount units, so I'd add it there - don't worry that it doesn't match the name, I have a TODO to merge lots of these and rename them

src/basic/unit-name.c
msekletar: I've checked this previously and I think that they should be fine but I have double check on that.

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
 
+bool unit_name_is_hashed(const char *name) {
+        char *s;
+        int i;
+
+        if (unit_name_is_valid(name, UNIT_NAME_PLAIN))
+                s = strrchr(name, '.');
+        else if (unit_name_is_valid(name, UNIT_NAME_INSTANCE))
+                s = strrchr(name, '@');
```
poettering: this should look for the *first* `@`. i.e. strrchr() → strchr().

See unit_name_to_prefix() for an example.

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
         return 0;
 }
 
+bool unit_name_is_hashed(const char *name) {
+        char *s;
+        int i;
+
+        if (unit_name_is_valid(name, UNIT_NAME_PLAIN))
```
poettering: maybe use unit_name_classify() instead.

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+                s = strrchr(name, '.');
+        else if (unit_name_is_valid(name, UNIT_NAME_INSTANCE))
+                s = strrchr(name, '@');
+        else
+                return false;
+
+        for (i = UNIT_NAME_HASH_LENGTH; i > 0; i--)
```
poettering: move variable declaration here? i.e.:

```c
for (int i = …
```

poettering: btw, there is really not much reason to count down... should be easier on the eye to count upwards here, i.e.:

```c
for (size_t i = 0; i < UNIT_NAME_HASH_LENGTH; i++) 
```

the compiler might turn this around again, but we shouldn't do the work for the compiler

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+                if (unhexchar(*--s) < 0)
+                        return false;
```
poettering, line 535: see the discussion around undefined behaviour here: #22800, i.e. counting pointers backwards is often enough UB

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+        char *s;
+        int i;
+
+        if (unit_name_is_valid(name, UNIT_NAME_PLAIN))
+                s = strrchr(name, '.');
+        else if (unit_name_is_valid(name, UNIT_NAME_INSTANCE))
+                s = strrchr(name, '@');
+        else
+                return false;
+
```
poettering, line 528: you are missing a check:

```c
if (s - name < UNIT_NAME_HASH_LENGTH + 1)
        return false;
```

src/basic/unit-name.c
```diff
@@ -31,6 +34,9 @@
         VALID_CHARS_WITH_AT                     \
         "[]!-*?"
 
+#define LONG_UNIT_NAME_HASH_KEY SD_ID128_MAKE(ec,f2,37,fb,58,32,4a,32,84,9f,06,9b,0d,21,eb,9a)
+#define UNIT_NAME_HASH_LENGTH 16
```
poettering: is this bytes? or hex chars?

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+        suffix = strrchr(name, '.');
+        if (!suffix)
+                return -EINVAL;
+
+        if (unit_type_from_string(suffix+1) < 0)
+                return -EINVAL;
+
+        h = siphash24_string(name, LONG_UNIT_NAME_HASH_KEY.bytes);
+
+        hash = hexmem(&h, sizeof(h));
```
poettering, line 558: hmm, this would mean we'd has differently on LE and BE machines. maybe instead:

```c
le64_t h;
…
h = le64toh(siphash24_string(…));
…
hash = hexmem(&h, sizeof(h));
```

i.e. explicitly convert to LE before we invoke hexmem(), so that we always turn the same hash value to the same string regardless of endianness

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+                len = UNIT_NAME_MAX - 1 - strlen(at) - strlen(hash) - 1;
+        else
+                len = UNIT_NAME_MAX - 1 - strlen(suffix+1) - strlen(hash) - 2;
+
+        mempcpy(n, name, len);
+
+        s = strjoin(n, "_", hash, at ?: suffix);
+        if (!s)
+                return -ENOMEM;
+        assert_se(unit_name_is_valid(s, UNIT_NAME_PLAIN|UNIT_NAME_INSTANCE));
```
poettering: this will likely be hit if the instance name is too long

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        hash = hexmem(&h, sizeof(h));
+        if (!hash)
+                return -ENOMEM;
+
+        assert_se(strlen(hash) == UNIT_NAME_HASH_LENGTH);
+
+        at = strrchr(name, '@');
+        if (at)
+                len = UNIT_NAME_MAX - 1 - strlen(at) - strlen(hash) - 1;
```
poettering: misses underflow check, i.e. if instance name is just too long

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+        else
+                len = UNIT_NAME_MAX - 1 - strlen(suffix+1) - strlen(hash) - 2;
```
poettering: similar

src/basic/unit-name.c
poettering: that said, i wonder if we should just not do this for now for instantiated units, i.e. only do hashing in case we have non-instantiated unit names. The thing is that instance names can be loooong, and thus we might be tempted to replace those with hashes, but if we do that specifier expansion in unit files with %i/%I cannot work anymore... since neither device nor mount units use templates I'd thus probably avoid this for now, and only do this for simple unit names.

poettering:
I think functions like mount_add_extras() in src/core/mount.c should probably log an error if they get ENAMETOOLNG from unit_name_to_path(), and suggest that people *must* set Where= in the .mount field manually.

src/basic/unit-name.c
```diff
@@ -507,6 +513,68 @@ int unit_name_template(const char *f, char **ret) {
...
+int unit_name_hash_long(const char *name, char **ret) {
+        _cleanup_free_ char *s = NULL, *hash = NULL;
+        char n[UNIT_NAME_MAX] = {}, *suffix, *at;
+        uint64_t h;
+        size_t len;
+
+        if (strlen(name) < UNIT_NAME_MAX) {
+                *ret = NULL;
+                return 0;
+        }
```
dnicolodi: I am not sure I understand this block. All the callers of this function make sure that the unit name is longer than `UNIT_NAME_MAX`, thus the check is redundant. At the same time, none of the callers check whether the value returned in `ret` is NULL, thus storing NULL there while not returning and error seems dangerous.

src/basic/unit-name.c
dnicolodi: Why this? In both call places, the check is anyhow executed right after this function.

msekletar: Sure, but we can add another call site and it is better to be explicit about code invariants.

src/basic/unit-name.c
msekletar: I have no strong preference here, maybe we should just change this to run `-EINVAL` or maybe we should rather copy input parameter value into `*ret=`.
@poettering any opinion about this one?

msekletar: Actually, I've changed the code to return `-EINVAL` for unit names that are shorther than `UNIT_NAME_MAX`.

bluca:
Is this still WIP? If not, please update the title

poettering (changes requested):

src/basic/unit-name.c
msekletar: Ok, I will drop all the code that deals with instances.

src/basic/unit-name.c
msekletar: Hmm, that check is bellow....

poettering: yeah sorry, it was added in the meantime. I had done a review early on, but never hit send becaus i got distracted. a few days late ri then hit send when i came across the issue again and only afterwards noticed some comments were then out of date

msekletar: Anyway, I reworked it as suggested.

src/basic/unit-name.c
msekletar: Dropped handling of templates.

src/basic/unit-name.c
msekletar: Fixed. I was not aware of UB issue.

src/basic/unit-name.c
msekletar: Likewise.

src/basic/unit-name.c
msekletar: What I had there previously was passing on all arches (s390x is BE), but I adopted this version.

src/basic/unit-name.c
msekletar: I've added range check.

src/basic/unit-name.c
msekletar: Not relevant anymore. Handling of templates was dropped.

src/basic/unit-name.c
msekletar: Constant is renamed to make it clear it is in chars.

src/basic/unit-name.c
```diff
@@ -10,7 +10,11 @@
 #include "hexdecoct.h"
 #include "memory-util.h"
 #include "path-util.h"
+#include "random-util.h"
+#include "sd-id128.h"
```
poettering: we usually include `sd-*.h` before the other `#include "…"` lines

src/basic/unit-name.c
```diff
@@ -507,6 +514,68 @@ int unit_name_template(const char *f, char **ret) {
...
+        if (!unit_name_is_valid(name, UNIT_NAME_PLAIN))
+                return false;
+
+        s = strrchr(name, '.');
+
+        if (s - name < UNIT_NAME_HASH_LENGTH_CHARS + 1)
+                return false;
+
+        s -= UNIT_NAME_HASH_LENGTH_CHARS + 1;
+        if (*s++ != '_')
```
poettering: hmm, first decrease then right after increase in a very cryptic way? why not do it like this:

```
s -= UNIT_NAME_HASH_LENGTH_CHARS;
if (s[-1] != '_')
        return false;
```

much less cryptic, at least in my eyes, no?

src/basic/unit-name.c
```diff
@@ -507,6 +514,68 @@ int unit_name_template(const char *f, char **ret) {
         return 0;
 }
 
+bool unit_name_is_hashed(const char *name) {
+        char *s;
+
+        if (!unit_name_is_valid(name, UNIT_NAME_PLAIN))
+                return false;
+
+        s = strrchr(name, '.');
```
poettering: maybe add an assert_se() around this.

src/basic/unit-name.c
```diff
@@ -507,6 +514,68 @@ int unit_name_template(const char *f, char **ret) {
...
+}
+
+int unit_name_hash_long(const char *name, char **ret) {
+        _cleanup_free_ char *s = NULL, *hash = NULL;
+        char n[UNIT_NAME_MAX] = {}, *suffix;
+        le64_t h;
+        size_t len;
+
+        if (strlen(name) < UNIT_NAME_MAX)
+                return -EINVAL;
```
poettering: maybe use a different error code here? just to signal the reader that this is a differnt kind of condition as the EINVAL cases below. The cases below after all are about truly invalid unit names. The one above just about overly long ones. maybe EMSGSIZE?

src/basic/unit-name.c
```diff
@@ -507,6 +514,68 @@ int unit_name_template(const char *f, char **ret) {
...
+        h = le64toh(siphash24_string(name, LONG_UNIT_NAME_HASH_KEY.bytes));
+
+        hash = hexmem(&h, sizeof(h));
+        if (!hash)
+                return -ENOMEM;
+
+        assert_se(strlen(hash) == UNIT_NAME_HASH_LENGTH_CHARS);
+
+        len = UNIT_NAME_MAX - 1 - strlen(suffix+1) - strlen(hash) - 2;
+        if (len <= 0 || len >= UNIT_NAME_MAX)
```
poettering: neither of this can happen, can it? hence assert()?

src/basic/unit-name.c
```diff
@@ -507,6 +514,68 @@ int unit_name_template(const char *f, char **ret) {
...
+                return -ERANGE;
+
+        mempcpy(n, name, len);
+
+        s = strjoin(n, "_", hash, suffix);
+        if (!s)
+                return -ENOMEM;
```
poettering, line 572: what about this:

```c
n = strndup(name, len);
if (!n)
        return -ENOMEM;

if (!strextend(&n, "_", hash, suffix))
        return -ENOMEM;

…
```

src/basic/unit-name.c
```diff
@@ -526,8 +595,17 @@ int unit_name_from_path(const char *path, const char *suffix, char **ret) {
         if (!s)
                 return -ENOMEM;
 
-        if (strlen(s) >= UNIT_NAME_MAX) /* Return a slightly more descriptive error for this specific condition */
-                return -ENAMETOOLONG;
+        if (strlen(s) >= UNIT_NAME_MAX) {
+                _cleanup_free_ char *n = NULL;
+
+                log_debug("Unit name too long, falling back to hashed unit name.");
```
poettering: show the name that is too long here...

poettering (changes requested):

poettering:
> I think functions like mount_add_extras() in src/core/mount.c should probably log an error if they get ENAMETOOLNG from unit_name_to_path(), and suggest that people must set Where= in the .mount field manually.

Not addressed yet?

msekletar:
> Not addressed yet?

I've added the log message.

msekletar:
Updated. @poettering PTAL.

src/basic/unit-name.c
```diff
@@ -507,6 +515,68 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        if (s - name < UNIT_NAME_HASH_LENGTH_CHARS + 1)
+                return false;
+
+        s -= UNIT_NAME_HASH_LENGTH_CHARS;
+        if (s[-1] != '_')
+                return false;
+
+        for (size_t i = 0; i < UNIT_NAME_HASH_LENGTH_CHARS; i++)
+                if (unhexchar(s[i]) < 0)
```
poettering: hmm, here's something I just realized: unhexchar() actually allows both uppercase and lowercase chars. We should only allow lower-case here though. hence maybe:

```c
if (strchr(LOWERCASEHEXDIGITS, s[i]) < 0)
        …
```

And then define LOWERCASEHEXDIGITS near HEXDIGITS at the top of string-util.h.

src/basic/unit-name.c
```diff
@@ -507,6 +515,68 @@ int unit_name_template(const char *f, char **ret) {
...
+                return -EMSGSIZE;
+
+        suffix = strrchr(name, '.');
+        if (!suffix)
+                return -EINVAL;
+
+        if (unit_type_from_string(suffix+1) < 0)
+                return -EINVAL;
+
+        h = le64toh(siphash24_string(name, LONG_UNIT_NAME_HASH_KEY.bytes));
```
poettering: le64toh → htole64

src/basic/unit-name.c
```diff
@@ -507,6 +515,68 @@ int unit_name_template(const char *f, char **ret) {
...
+
+        hash = hexmem(&h, sizeof(h));
+        if (!hash)
+                return -ENOMEM;
+
+        assert_se(strlen(hash) == UNIT_NAME_HASH_LENGTH_CHARS);
+
+        len = UNIT_NAME_MAX - 1 - strlen(suffix+1) - UNIT_NAME_HASH_LENGTH_CHARS - 2;
```
poettering, line 564: hmm, shouldn't the -2 be -1?

poettering: ah, got it

poettering:
two more tweaks.

src/basic/unit-name.c
msekletar: Actually I had that in one of the prior versions ;)

msekletar:
Updated. @poettering PTAL.

poettering:
lgtm, thanks!

poettering:
testcase_long_sysfs_path test failed in CI.

msekletar:
I've updated the test case once again. It seems that previous failure was caused because we can now actually handle long sysfs paths and `journalctl --grep` was failing because it didn't find the errors it was excepting.

poettering:
Ci failure looks unrelated

mwilck:
I haven't seen this before.  I dislike the fact that the hash overwrites the tail of the original path name. More often than not, the most distinctive portion of a device path is at the tail, while the front components of the path are just directories that are common to many devices. This will matter in particular for log messages, making it hard for users to figure out what device a given unit name actually maps to. Or am I overlooking something.

poettering:
This is an edge case when everything is fucked anyway. Usability is long out of the window when this edge case is hit. If you actually care about user friendliness of any of this: don't generate crazy long paths like this in the kernel.

Issue #22635, opened 2022-02-28
Removal of SD_TPM_PCR option breaks grub + sd-stub set-ups


alfonsosanchezbeato:
**systemd version the issue has been seen with**
v250

**Used distribution**
Ubuntu

**Linux kernel version used** (`uname -a`)
`Linux host 5.4.0-100-generic #113-Ubuntu SMP Thu Feb 3 18:43:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**CPU architecture issue was seen on**
x86

**Expected behaviour you didn't see**
 NA

**Unexpected behaviour you saw**
 
Commit https://github.com/systemd/systemd/commit/faacf1807e8fcbee4bf60495e0145e4522d76393 breaks Ubuntu setups. In Ubuntu, we configure grub with PCR8 whilst we configure sd-stub with PCR12. This allows us to seal, compute, and differentiate cmdlines as measured by grub and as measured by sd-stub, on the same system.

Could the commit be reverted please?

Or for example use distinct from grub pcr by default, i.e. 12 - which i guess will not fly, given it is a breaking change for anybody who only ever used sdstub only without grub.

Simultaneous measurements of cmdline to the same pcr by either grub & sd-stub and harder to compute sealing policies for if one wants to support one/another/either.

**Steps to reproduce the problem**
 NA

**Additional program output to the terminal or log subsystem illustrating the issue**
NA

alfonsosanchezbeato:
@xnox ^^

medhefgo:
I don't quite see why grub/sd-boot/sd-stub should be using different PCRs to measure the same thing. That's literally what a PCR measurement is supposed to be for, no?

/cc @poettering

poettering:
I am not sure I follow? Using the kernel cmdline PCR to distinguish boot loaders sounds quite a surprising use for them? Can you elaborate?

Why not just check for contents of `/sys/firmware/efi/efivars/LoaderInfo-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f` if you want to know the boot loader?

poettering:
Dropping from milestone, I don#t think it's obvious what the right approach here is, I simply understand the rationale here.

xnox:
I will provide bios measurements log to explain what is going on.

xnox:
tl;dr it is impossible to compute PCR values when grub chainloads systemd-boot or Boot Loader Specification type 2 binary, because grub measures a lot of things into PCR 8.

```
[...]

PR #22761, opened 2022-03-16
sd-boot: change kernel cmdline PCR from 8 to 12

poettering:
A fix for #22635

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -109,6 +109,19 @@ CHANGES WITH 251:
           250. For newer kernels, non-x86 systems, or older x86 systems,
           there should be no visible changes.
 
+        * sd-boot will now measure the kernel command line into TPM PCR 12
+          rather than PCR 8. This improves usefulness of the measurements on
+          sytems where sd-boot is chainloaded from Grub. Grub measures all
+          commands its executes into PCR 8, which makes it very hard to use
+          reasonably, hence separate ourselves from that and use PCR 12
+          instead, which is already what certain Ubuntu editions use it for. To
+          retain compatibility with systems running older systemd systems a new
+          Meson option 'efi-tpm-pcr-compat' has been added (which defaults to
+          false). If enabled, the measurement is done twice: into the new-style
+          PCR 12 *and* the old-style PCR 8. It's strongly advised to migrate
+          all users to PCR 12 for this purpose in the long run, as we intend to
+          remove this compatibility feature again in two year's time.
+
 CHANGES WITH 250:
 
         * Support for encrypted and authenticated credentials has been added.


meson_options.txt
```diff
@@ -426,6 +426,8 @@ option('efi-libdir', type : 'string',
        description : 'path to the EFI lib directory')
 option('efi-includedir', type : 'string', value : '/usr/include/efi',
        description : 'path to the EFI header directory')
+option('efi-tpm-pcr-compat', type : 'boolean', value : 'false',
```
bluca, line 429: Shouldn't this default to true as we introduce it, and then changed to false later, to reduce the risk of making systems unbootable?

src/boot/efi/cpio.c
```diff
@@ -312,6 +312,7 @@ EFI_STATUS pack_cpio(
                 UINT32 dir_mode,
                 UINT32 access_mode,
                 UINT32 tpm_pcr,
+                UINT32 tpm_pcr2,
```
bluca: Any chance of using a list/array here? Aesthetically speaking it would be a bit nicer than hardcoding one and two

meson_options.txt
medhefgo: And here I thought you hate needles abbreviations and yet you use the very same one you disliked me using…

src/boot/efi/measure.c
```diff
@@ -145,6 +145,10 @@ EFI_STATUS tpm_log_event(UINT32 pcrindex, EFI_PHYSICAL_ADDRESS buffer, UINTN buf
 
         assert(description);
 
+        /* PCR disabled */
+        if (pcrindex == (UINT32) -1)
```
medhefgo: `UINT32_MAX`

src/boot/efi/measure.h
```diff
@@ -3,6 +3,20 @@
...
 
+/* This TPM PCR is where we extend the kernel command line and any passed credentials here. */
+#define TPM_PCR_INDEX_KERNEL_PARAMETERS 12U
+
+/* We used to write the the kernel command line/credentials into PCR 8, in systemd <= 250. Let's provide for
+ * some compatibility. (Remove in 2023!) */
+#if EFI_TPM_PCR_COMPAT
+#define TPM_PCR_INDEX_KERNEL_PARAMETERS_COMPAT 8U
+#else
+#define TPM_PCR_INDEX_KERNEL_PARAMETERS_COMPAT ((UINT32) -1)
```
medhefgo: `UINT32_MAX`

meson_options.txt
poettering: > Shouldn't this default to true as we introduce it, and then changed to false later, to reduce the risk of making systems unbootable? 

I would agree if this was a user-facing setting. But it's not, it's a build-time setting, facing the integrators which build our package. I think we can safely be more agressive there. Or in other words: if integrators/package builders don't read NEWS they are simply bad at their jobs. If users don't read NEWS then they are just normal ;-)

And I'd really not stretch this out more than necessary. I have the suspicion the primary user for this is Ubuntu anyway, and they know about this anyway.

poettering: > And here I thought you hate needles abbreviations and yet you use the very same one you disliked me using…

You have a point, but this one is already firmly established I figure, i.e. `-Dcompat-mutable-uid-boundaries` exists already, `cgroup_compat_warn()`, `exec_runtime_deserialize_compat()`, and so on. It's kinda a full word now in our tree...

poettering: or in other words: do as i say, don't do what i do ;-)

poettering:
addressed he points raised, except for the "compat" thing and the default to compat off thing. ptal.

bluca (approved):

Issue #22390, opened 2022-02-03
sd-boot: reboot-through-nextboot should be made generic, and independent of BitLocker


poettering:
See discussion in:

https://github.com/systemd/systemd/pull/22043#issuecomment-1028964448

I am pretty sure the option knob should be renamed (to drop the bitlocker specificity), hence making this a blocker issue for the upcoming release, since we shouldn't do a release with an option that is going to be renamed.

medhefgo:
I guess it can make sense to generalize it.

poettering:
I prepped #22762 now, which disables the feature and marks it as experimental for now. That way the code can stay in and be used, but we don't have to commit to its current behaviour and naming of the future, but do a v251 release with it.

I'll reassign the milestone for this one now to v252, so that we fix this properly by generalizing this as mentioned. #22762 otoh will be assigned to v251 since it's only a temporary stopgap.

prodrigestivill:
Just to add another use case when generalizing this enhancement, in case it was not thought yet.

As for now sd-boot is not able to boot across different operating systems where the EFI files are spread onto different local hard disks.
For my understanding having this new feature generalized would allow to boot from different local disks (or different ESP partitions) without having to add all the logic GRUB needs (even that in this case it would be required to soft reboot).

This will also solve other requests like #3252.

keszybz:
We made no progress whatsoever on this. It's too late for v253 at this point.

GitMurf:
Curious if this has been re-visited or what the status is (if any)? Thanks!

Issue #12716, opened 2019-06-01
sd-boot: Measure initrd into TPM (attempt #2)


josephlr:
## Describe the solution you'd like

#2587 added support for measuring the kernel command line into the TPM. This is to allow for "Measured Boot" where all components of the boot chain (firmware, bootloader, kernel, etc..) are measured into the TPM for later cryptographic attestation. However, that patch didn't include measuring a key component of the boot chain: the initrd. #6124 attempted to add this support, but @haraldh correctly pointed out that it doesn't make sense to measure the disk and then have the kernel reload the initrd from disk, as this is quite insecure.

It's worth looking at how @mjg59's Grub patches handle this. For context, the Linux kernel has two entry points:
  - [EFISTUB](https://www.kernel.org/doc/Documentation/efi-stub.txt): the standard UEFI entry point
    - This allows the kernel bzImage to masquerade as a normal EFI binary
    - It has the magic `initrd` command line parameter
    - Used by systemd-boot when using normal `boot.c` code
    - Used by Grub with [`chainloader`](https://www.gnu.org/software/grub/manual/grub/html_node/Chain_002dloading.html)
  - [EFI handover](https://www.kernel.org/doc/html/latest/x86/boot.html?highlight=boot#efi-handover-protocol):
    - Similar to the other kernel entry points, you set the `bootparams` struct (née [Zero Page](https://www.kernel.org/doc/html/latest/x86/zero-page.html)) and jump to an address.
    - Requires bootloader to load initrds into memory before jumping
    - Used by systemd-boot when using `shim.c` code
    - Used by (non-mainline) Grub with `linuxefi`.

In Matthew's patches to GRUB, the command line params can be measured into PCR 8 (like systemd-boot) and the initrds can be measured into PCR 9 (unlike systemd-boot) when the kernel is loaded with `linuxefi`.

My proposal is two-fold:

1. If a systemd-boot loader entry uses `linux` and `initrd` parameters, we load the kernel using the EFI handover protocol, similar to what we already do in the stub code.
2. If the `tpm` option is enabled, measure any initrds into the TPM. Which PCR would be determined by a `initrd-pcrindex` option, defaulting to PCR 9.

### Advantages to this approach

- Linux is always loaded the same way with systemd-boot. We would use the EFI handover entry point for the normal bootloader and the stub.
- The PCR measurements are similar between Grub and systemd-boot. This makes it easier for users who only care about UEFI/Secure Boot/Measured Boot to switch to systemd-boot.
[...]

PR #22762, opened 2022-03-16
sd-boot: disable bitlocker reboot feature for now

poettering:
Conceptually the feature is great and should exist, but in its current
form should be worked to be generic (i.e. not specific to
Windows/Bitlocker, but appliable to any boot entry), not be global (but
be a per-entry thing), not require a BootXXXX entry to exist, and not
check for the BitLocker signature (as TPMs are not just used for
BitLocker).

Since we want to get 251 released, mark it in the documentation, in NEWS
and in code as experimental and make clear it will be reworked in a
future release. Also, make it opt-in to make it less likely people come
to rely on it without reading up on it, and understanding that it will
likely change sooner or later.

Follow-up for: #22043
See: #22390

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -52,6 +52,17 @@ CHANGES WITH 251:
           or ID field of /etc/os-release or another suitable identifier before
           deploying the image.
 
+        * sd-boot gained a new *experimental* setting "reboot-for-bitlocker" in
+          loader.conf that implements booting Microsoft Windows from the
+          sd-boot in a way that first reboots the system, to reset the TPM
+          PCRs. This improves compatibility with BitLocker's TPM use, as the
+          PCRs will only record the Windows boot process, and not sd-boot
+          itself, thus retaining the PCR measurements not involving
+          sd-boot. Note that this feature is experimental for now, and is
+          likely going to be generalized, renamed and removed in its current
+          form in a future release, without retaining compatibility with its
+          current implementation.
+
         * The --make-machine-id-directory= switch to bootctl has been replaced
           by --make-entry-directory=, given that the entry directory is not
           necessarily named after the machine ID, but after some other suitable
--- a/man/loader.conf.xml
+++ b/man/loader.conf.xml
@@ -206,8 +206,11 @@
       <varlistentry>
         <term>reboot-for-bitlocker</term>
 
-        <listitem><para>Work around BitLocker requiring a recovery key when the boot loader was
-        updated (enabled by default).</para>
+        <listitem><para>Caveat: This feature is experimental, and is likely to be changed (or removed in its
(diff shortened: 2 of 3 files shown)


bluca (approved):

yuwata (approved):

medhefgo:
Thinking more on this, I dunno if this really makes too much sense to generalize, though.

If using BootNext to launch the kernel, PCR measurements are always gonna change on kernel upgrades, which happen much more frequently than sd-boot ones (and #12716 talks about starting it manually to skip that measurement, which wouldn't work anymore then).

Also, so far BitLocker seems to be the only real use case that has come up. Especially because most people think it's impossible to dualboot with both BitLocker and linux using tpm sealing. In a way the existence of this option and its auto-detection help to make things just work.

I think this should stay the way it is for now including the auto-detection (and not make it experimental). Especially, because the windows boot manager is itself auto detected, so you couldn't turn it on/off with a entry-specific `.conf` option.
Meanwhile, I would argue that it's probably better to wait and see if there will be any real other users of something like this first. My guess is that it's a non issue as sd-cryptenroll defaults to just use PCR7 (and you should really be using secure boot)…

But whatever, do whatever you feel like doing.

poettering:
> If using BootNext to launch the kernel, PCR measurements are always gonna change on kernel upgrades, which happen much more frequently than sd-boot ones (and #12716 talks about starting it manually to skip that measurement, which wouldn't work anymore then).

Well, we can chainload almost anything. For example there are apparently people who chainload grub and sd-boot. I think it would make a ton of sense to optionally chainload let's say RHEL grub from sd-boot this way, and get the RHEL measurements into TPM.

poettering:
> Also, so far BitLocker seems to be the only real use case that has come up. Especially because most people think it's impossible to dualboot with both BitLocker and linux using tpm sealing. In a way the existence of this option and its auto-detection help to make things just work.

Well, it's the major usecase that has come up, but *any* OS that uses the TPM for verification of stuff will need something like this. For example ChromeOS and similar. And even if dual booting chromeos is not feasible at the moment, I think we really should design our basic building blocks generic enough to make that happen.

Again, I actually like this feature a lot, I just think we should make it even more useful.

GitMurf:
> As mentioned I think we should default to this for Windows.

Did this ever get revisited to "default" for Windows? I spent the last couple days trying to get dual boot working with Arch and Windows with BitLocker / TPM and I finally stumbled upon the GPO edits to remove `4` from the PCR list which solved it for me. But curious if this is still expected or if there was any progress made on "defaults" to handle Windows in these situations. In a perfect world, users would not have to modify GPO, especially given often GPO is controlled by an organization (work, school etc.). Thanks!

PR #22764, opened 2022-03-16
conf-parser: Don't treat ENOTDIR as a fatal error

daandemeyer:
It's possible that a directory in the config path doesn't exist which
causes fopen() to fail with ENOTDIR. Let's treat this the same as ENOENT
when parsing config files.

### DIFF
--- a/src/shared/conf-parser.c
+++ b/src/shared/conf-parser.c
@@ -291,13 +291,14 @@ int config_parse(
         if (!f) {
                 f = ours = fopen(filename, "re");
                 if (!f) {
-                        /* Only log on request, except for ENOENT,
+                        /* Only log on request, except for ENOENT and ENOTDIR,
                          * since we return 0 to the caller. */
-                        if ((flags & CONFIG_PARSE_WARN) || errno == ENOENT)
-                                log_full_errno(errno == ENOENT ? LOG_DEBUG : LOG_ERR, errno,
+                        bool is_fatal = !IN_SET(errno, ENOENT, ENOTDIR);
+                        if (FLAGS_SET(flags, CONFIG_PARSE_WARN) || !is_fatal)
+                                log_full_errno(is_fatal ? LOG_ERR : LOG_DEBUG, errno,
                                                "Failed to open configuration file '%s': %m", filename);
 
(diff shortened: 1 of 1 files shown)


src/shared/conf-parser.c
```diff
@@ -291,13 +291,14 @@ int config_parse(
         if (!f) {
                 f = ours = fopen(filename, "re");
                 if (!f) {
-                        /* Only log on request, except for ENOENT,
+                        /* Only log on request, except for ENOENT and ENOTDIR,
                          * since we return 0 to the caller. */
-                        if ((flags & CONFIG_PARSE_WARN) || errno == ENOENT)
-                                log_full_errno(errno == ENOENT ? LOG_DEBUG : LOG_ERR, errno,
+                        bool is_fatal = errno != ENOENT && errno != ENOTDIR;
```
yuwata: Please use `IN_SET()` macro.

src/shared/conf-parser.c
```diff
@@ -291,13 +291,14 @@ int config_parse(
...
+                        if ((flags & CONFIG_PARSE_WARN) || !is_fatal)
```
yuwata: Not a matter, but I prefer `FLAGS_SET()` macro.

yuwata (approved):
LGTM. But two minor comments.

DaanDeMeyer:
Addressed comments, setting the green label

bluca (approved):

poettering:
hmm? that's not really the case though? if I attempt to open a file in a non-existing dir i get ENOENT. Try it out: `cat /doesntexist/foo` will say File not found...

I am confused?

ENOTDIR would be returned if a prefix dir exists but is not a dir... which is something we sould log about and not hide away

poettering:
Also see documentation about open():
```
       ENOTDIR
              A component used as a directory in pathname is not, in fact, a directory, or O_DIRECTORY was specified and pathname was not a directory.
```

Hence, puzzled?

DaanDeMeyer:
Ah damn, my bad, turns out .config/ in home is actually a file on this machine. Never expected that. Let me close this and the other PR

Issue #20989, opened 2021-10-12
nspawn: uidmap setup hosed


mbiebl:
Version: 249.4

Running the autopkgtest suite on Debian bookworm/sid, leads to the following failure in boot-and-services:
```
autopkgtest [13:47:40]: test boot-and-services: [-----------------------
none
Created symlink /etc/systemd/system/default.target → /lib/systemd/system/graphical.target.
bash: line 1: 12745 Killed                  /tmp/autopkgtest.HPdyca/build.4W1/src/debian/tests/boot-and-services 2> >(tee -a /tmp/autopkgtest.HPdyca/boot-and-services-stderr >&2) > >(tee -a /tmp/autopkgtest.HPdyca/boot-and-services-stdout)
autopkgtest [13:47:41]: test process requested reboot with marker boot1
none
test_profile (__main__.AppArmorTest)
AppArmor confined unit ... Replacement succeeded for "violator-test".
ok
test_help (__main__.CLITest)
--help works and succeeds ... ok
test_invalid_option (__main__.CLITest)
Calling with invalid option fails ... ok
test_version (__main__.CLITest)
--version works and succeeds ... ok
test_cpushares (__main__.CgroupsTest)
service with CPUShares ... skipped 'test needs to be reworked on unified cgroup hierarchy'
test_simple (__main__.CgroupsTest)
simple service ... skipped 'test needs to be reworked on unified cgroup hierarchy'
test_bash_crash (__main__.CoredumpTest) ... ok
test_log_for_service (__main__.JournalTest) ... ok
test_no_options (__main__.JournalTest) ... ok
test_boot (__main__.NspawnTest) ... ok
test_service (__main__.NspawnTest) ... ERROR
test_failing (__main__.SeccompTest) ... Job for systemd-nspawn@c1.service failed because the control process exited with error code.
See "systemctl status systemd-nspawn@c1.service" and "journalctl -xeu systemd-nspawn@c1.service" for details.
ok
test_0_init (__main__.ServicesTest)
Verify that init is systemd ... ok
test_cron (__main__.ServicesTest) ... ok
test_dbus (__main__.ServicesTest) ... ok
test_gdm3 (__main__.ServicesTest) ... 324 /usr/sbin/gdm3
328 gdm-session-worker [pam/gdm-launch-environment]
363 /usr/libexec/gdm-wayland-session dbus-run-session -- gnome-session --autostart /usr/share/gdm/greeter/autostart
ok
[...]

PR #22774, opened 2022-03-17
nspawn: uidmap mount fix

poettering:
Fixes: #20989

### DIFF
--- a/src/home/homework-mount.c
+++ b/src/home/homework-mount.c
@@ -216,6 +216,12 @@ static int make_userns(uid_t stored_uid, uid_t exposed_uid) {
         if (r < 0)
                 return log_oom();
 
+        /* Map nspawn's mapped root UID as identity mapping so that people can run nspawn uidmap mounted
+         * containers off $HOME, if they want. */
+        r = strextendf(&text, UID_FMT " " UID_FMT " " UID_FMT "\n", UID_MAPPED_ROOT, UID_MAPPED_ROOT, 1);
+        if (r < 0)
+                return log_oom();
+
         /* Leave everything else unmapped, starting from UID_NOBODY itself. Specifically, this means the
          * whole space outside of 16bit remains unmapped */
 


src/nspawn/nspawn.c
```diff
@@ -3563,7 +3563,7 @@ static int inner_child(
 
 static int setup_notify_child(void) {
         _cleanup_close_ int fd = -1;
-        union sockaddr_union sa = {
+        static const union sockaddr_union sa = {
```
bluca, line 3566: this is not just adding const, but static too, was that intentional?

src/shared/mount-util.c
```diff
@@ -1049,14 +1049,42 @@ int make_mount_point(const char *path) {
         return 1;
 }
 
-static int make_userns(uid_t uid_shift, uid_t uid_range) {
-        char line[DECIMAL_STR_MAX(uid_t)*3+3+1];
+/* If REMOUNT_IDMAP_HOST_ROOT is set we'll include a mapping here that maps the host root user accessing the
+ * idmapped mount to the this user ID on the backing fs. This is the last valid UID in the *signed* 32bit
+ * range. You might wonder why precisely use this specific UID for this purpose? Well, we definitely cannot
+ * use the first 0…65536 UIDs for that, since in most cases that's precisely the file range we intend to map
+ * to some high UID range, ans since UID mappings have to be bijective we thus cannot use them at
```
bluca: /ans/and/

src/shared/mount-util.c
```diff
@@ -1049,14 +1049,42 @@ int make_mount_point(const char *path) {
...
+ * all. Furthermore the UID range beyond INT32_MAX (i.e. the range above the signed 32bit range) is icky,
+ * since many APIs cannot use it (example: setfsuid() returns the old UID as signed integer). Following our
+ * usual logic of assigning a 16bit UID range to each container, so that the upper 16bit of a 32bit UID value
+ * indicate kind of a "container ID" and the lower 16bit map directly to the intended user you can read this
```
bluca: should this be documented in one of our interfaces in docs/ ?

src/shared/mount-util.c
```diff
@@ -1049,14 +1049,42 @@ int make_mount_point(const char *path) {
...
          * process whose only purpose is to give us a new user namespace. It's killed when we got it. */
 
-        xsprintf(line, UID_FMT " " UID_FMT " " UID_FMT "\n", 0, uid_shift, uid_range);
+        if (asprintf(&line, UID_FMT " " UID_FMT " " UID_FMT "\n", 0, uid_shift, uid_range) < 0)
+                return log_oom_debug();
+
+        /* If requested we'll include an entry in the mapping so that the host root user can make changes to
+         * the uidmapped mount like it normally would. Specifically, we'll map the user with UID_HOST_ROOT on
+         * the backing fs to UID 0. This is useful, since nspawn code wants to create various missing inodes
+         * in the OS tree before booting into it, and this becomes very easy and straightforward to do that
```
bluca: /to do that/to do/ (or remove 'this')

src/shared/mount-util.h
```diff
@@ -112,7 +112,11 @@ int mount_image_in_namespace(pid_t target, const char *propagate_path, const cha
 
 int make_mount_point(const char *path);
 
-int remount_idmap(const char *p, uid_t uid_shift, uid_t uid_range);
+typedef enum RemountIdmapFlags {
+        REMOUNT_IDMAP_HOST_ROOT = 1 << 0,
```
bluca, line 123: it seems every call actually passes the flag, is it necessary at all? are there cases where we don't want to do this, or expect we won't?

src/nspawn/nspawn.c
poettering: yes. if we don#t modify we can always make it global, because then it's safe to share with other concurrent users.

src/shared/mount-util.h
poettering: Good point. I added the flag explicitly though since it's not obvious, and it does create some attack surface if misused. Thus, callers should really explicitly request this, even though all our current callers *will* request this.

i.e. adding this extra uid mapping should always be done only after careful consideration that it is safe. And thus be opt-in instead of default.

src/shared/mount-util.c
```diff
@@ -1049,14 +1049,42 @@ int make_mount_point(const char *path) {
         return 1;
 }
 
-static int make_userns(uid_t uid_shift, uid_t uid_range) {
-        char line[DECIMAL_STR_MAX(uid_t)*3+3+1];
```
bluca, line 1053: The size is still fixed, just at twice the length - given asprintf is expensive (or so I'm often told) and this is a library call, maybe just doubling the static buffer and optionally pointing a second xsprintf past the first newline would work? But it's not too important I guess

src/shared/mount-util.c
poettering: For now I'd consider this an internalism of nspawn. files created this way will only be owned by this UID for a very brief time, because we first create the inodes and then chown them. Other tools should never see them

src/shared/mount-util.h
bluca: can you add a comment explaining this where the flag is defined then? It's not super obvious when it shouldn't be used, so an hint would make it easier to avoid mistakes

bluca:
Can TEST-13-NSPAWN-SMOKE be extended to cover this case, so that we don't risk regressing in the future?

src/shared/mount-util.c
poettering: Well, depends on the flag. But this is not called in an inner loop, but exactly once. I think performance doesn't really matter here, but it is a lot more readable and less likely to mistakes if we juse use dynamic memory.

src/shared/mount-util.c
```diff
@@ -1049,14 +1049,43 @@ int make_mount_point(const char *path) {
...
+         * can just do it under its own regular UID. Note that in that case the container's runtime uidmap
+         * (i.e. the one the container payload processes run in) will leave this UID unmapped, i.e. if we
+         * accidentally leave files owned by host root in the already uidmapped tree around they'll show up
+         * as owned by 'nobody', which is safe. (Of course, we shouldn't leave such inodes around, but always
+         * chown() them to the container's own UID range, but it's good to have a safety net, in case we
+         * forget it.) */
+        if (flags & REMOUNT_IDMAP_HOST_ROOT)
+                if (strextendf(&line,
+                               UID_FMT " " UID_FMT " " UID_FMT "\n",
+                               UID_HOST_ROOT, 0, 1) < 0)
```
brauner: UID_HOST_ROOT being INT32_MAX-1?

poettering: yes

src/nspawn/nspawn.c
bluca: Can't the `bind()` modify it in some cases? According to the manpage that can happen when sun_path is not null-terminated - which is not the case here, but it shows that at least in principle it can happen

bluca:
> > Can TEST-13-NSPAWN-SMOKE be extended to cover this case, so that we don't risk regressing in the future?
> 
> The CIs already test for this, no? I mean, that's how this was noticed and the issue is marked as "ci-blocker" due to that.

Our CI never failed though

> > Can TEST-13-NSPAWN-SMOKE be extended to cover this case, so that we don't risk regressing in the future?
> 
> The CIs already test for this, no? I mean, that's how this was noticed and the issue is marked as "ci-blocker" due to that.

There is a test, but it's one of the debian-specific ones, so we don't see it because here we only run it on a very old kernel on Ubuntu 22.04. If it was covered by TEST-13-NSPAWN-SMOKE we would have seen it immediately, as the Arch job runs on new kernels all the time. There's a minimal reproducer shared by @mrc0mmand that should be very easy to add:

https://github.com/systemd/systemd/issues/20989#issuecomment-997024152

```
# mkdir -p /var/lib/machines/my-container/etc
# cp /etc/os-release /var/lib/machines/my-container/etc/
# systemd-nspawn -U -b -D /var/lib/machines/my-container/
Spawning container my-container on /var/lib/machines/my-container.
Press ^] three times within 1s to kill container.
Selected user namespace base 972619776 and range 65536.
Failed to create directory at /var/lib/machines/my-container/usr: Value too large for defined data type
```

src/shared/mount-util.h
poettering: done

src/nspawn/nspawn.c
poettering: the pointer is marked "const", so no, glibc guarantees you it will not touch the structure.

poettering:
> There is a test, but it's one of the debian-specific ones, so we don't see it because here we only run it on a very old kernel on Ubuntu 22.04. If it was covered by TEST-13-NSPAWN-SMOKE we would have seen it immediately, as the Arch job runs on new kernels all the time. There's a minimal reproducer shared by @mrc0mmand that should be very easy to add:

not really, as the invocation of that container will fail anyway, given it has no init binary in it. we'd hve to check where exactly it fails

src/nspawn/nspawn.c
bluca: got it, thanks

bluca (approved):

bluca:
> > There is a test, but it's one of the debian-specific ones, so we don't see it because here we only run it on a very old kernel on Ubuntu 22.04. If it was covered by TEST-13-NSPAWN-SMOKE we would have seen it immediately, as the Arch job runs on new kernels all the time. There's a minimal reproducer shared by @mrc0mmand that should be very easy to add:
> 
> not really, as the invocation of that container will fail anyway, given it has no init binary in it. we'd hve to check where exactly it fails

The test already builds an image that it uses though, that should make it simpler?

PR #22778, opened 2022-03-17
kernel-install/bootctl: layout fixes

poettering:
Addressing the points I made in #22379, except one: I left the names $layout and $initrd_generator for the variables of /etc/kernel/install.conf in place. I think they are kinda poor given we everywhere later reference the same concepts in variables $KERNEL_INSTALL_LAYOUT and $KERNEL_INSTALL_INITRD_GENERATOR, but we can still change that later, or just accept that complains about that are mostly bike-shedding.

### DIFF
--- a/src/kernel-install/kernel-install
+++ b/src/kernel-install/kernel-install
@@ -155,10 +155,29 @@ done
 [ -z "$ENTRY_TOKEN" ] && ENTRY_TOKEN="$MACHINE_ID"
 
 if [ -z "$layout" ]; then
-    # Administrative decision: if not present, some scripts generate into /boot.
-    if [ -d "$BOOT_ROOT/$ENTRY_TOKEN" ]; then
+    # No layout configured by the administrator. Let's try to figure it out
+    # automatically from metadata already contained in $BOOT_ROOT.
+    if [ -e "$BOOT_ROOT/loader/entries.srel" ]; then
+        read -r ENTRIES_SREL <"$BOOT_ROOT/loader/entries.srel"
+        if [ "$ENTRIES_SREL" = "type1" ]; then
+            # The loader/entries.srel file clearly indicates that the installed
+            # boot loader implements the proper standard upstream boot loader
+            # spec for Type #1 entries. Let's default to that, then.
+            layout="bls"
+        else
+            # The loader/entries.srel file indicates some other spec is
+            # implemented and owns the /loader/entries/ directory. Since we
+            # have no idea what that means, let's stay away from it by default.
+            layout="other"
+        fi
+    elif [ -d "$BOOT_ROOT/$ENTRY_TOKEN" ]; then
+        # If the metadata in $BOOT_ROOT doesn't tell us anything, then check if
+        # the entry token directory already exists. If so, let's assume it's
+        # the standard boot loader spec, too.
         layout="bls"
     else
+        # There's no metadata in $BOOT_ROOT, and apparently no entry token
+        # directory installed? Then we really don't know anything.
         layout="other"
     fi
 fi


poettering:
Split out some commits into #22784, and given that's merged now, rebased this one on top.

src/kernel-install/kernel-install
```diff
@@ -155,8 +155,17 @@ done
 [ -z "$ENTRY_TOKEN" ] && ENTRY_TOKEN="$MACHINE_ID"
 
 if [ -z "$layout" ]; then
-    # Administrative decision: if not present, some scripts generate into /boot.
-    if [ -d "$BOOT_ROOT/$ENTRY_TOKEN" ]; then
+    # No layout configured by the administrator. Let's try to figure it out automatically
+    if [ -e "$BOOT_ROOT/loader/entries.srel" ]; then
+        read -r ENTRIES_SREL <"$BOOT_ROOT/loader/entries.srel"
+        if [ "$ENTRIES_SREL" == "standard" ]; then
```
behrmann:
```suggestion
        if [ "$ENTRIES_SREL" = "standard" ]; then
```
since it's unfortunately not POSIX shell

src/boot/bootctl.c
```diff
@@ -164,6 +163,8 @@ static int load_install_machine_id_and_layout(void) {
                 if (r < 0 && !IN_SET(r, -ENOENT, -ENOMEDIUM))
                         return log_error_errno(r, "Failed to get machine-id: %m");
         } else {
+                log_notice("Read kernel install machine ID from /etc/machine-info. Please migrate to /etc/kernel/entry-token instead.");
```
keszybz: Maybe say "Read token from $KERNEL_INSTALL_MACHINE_ID from /etc/machine-info. Please move it to /etc/kernel/entry-token."
("kernel install machine ID" is somewhat misleading here, when we're calling it "entry token" elsewhere).

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -309,6 +309,18 @@ focus for this specification. More specifically, on non-EFI systems
 configuration snippets following this specification cannot be used to spawn
 other operating systems (such as Windows).
 
+Unfortunately, there are implementations of boot loading infrastructure that
+are also using the /loader/entries/ directory, but place files in them that are
+not valid by this specification. This is quite unfortunate. In order to
```
keszybz: Strike "This is quite unfortunate." because the previous sentence already says "Unfortunately", so this is just a repeat.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -309,6 +309,18 @@ focus for this specification. More specifically, on non-EFI systems
...
+minimize confusion a boot loader implementation may place a file
+/loader/entries.srel next to the /loader/entries/ directory containing the
+ASCII string "standard" (suffixed with a UNIX newline). Tools that need to
```
keszybz: I think it'd be better to use the string "bls" instead, like what is used for the install layout. "Standard" has the problem that there may be more than one standard at some point. "bls" clearly says *which* standard is used.

src/kernel-install/kernel-install
```diff
@@ -155,8 +155,17 @@ done
...
+            # The loader/entries.srel file clearly indicates that this uses the proper standard boot loader spec
+            layout="bls"
+        else
+            layout="other"
```
keszybz, line 171: Shoudn't we assume "layout=xxx" when "xxx" is specified in `entries.srel`? If you use some non-standard entry format, it seems reasonable to assume that you're using a corresponding layout of entries.

I'm not sure if it makes sense to create a separate file for this. We already have ~~`layout.conf`~~ [edit:] `loader.conf`: why not add a field there instead? We already put random-seed config in loader.conf.

keszybz (commented):
Looks reasonable in general.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: So I really would avoid the abbreviated "bls" moniker. It's not mentioned even a single time on https://systemd.io/BOOT_LOADER_SPECIFICATION but it *is* mentioned a lot on https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault

Hence I would just give up the short abbreviation "bls" — it's more associated with the fedora/grub thing than with ours stuff. Given that the feature is precisely about distinguishing our stuff from theirs I think it would be a really bad idea to pick the moniker they so far squatted. That just adds more confusion.

I am OK with using basically any moniker for this, but "bls" is the worst choice I think.

I also would like the spec to not carry the systemd name in anything, given it's generic, and if Grub one day maybe implements our upstream spec they shouldn't claim their implementation was "systemd", hence I also wuld not put the "systemd" string in that file.

hence, if not "bls", and not "systemd", what else? "fdo" could work, but we moved aaway from that for new stuff. "official", "upstream", "original" all came to mind, but I figured "standard" is a nicer thing, since it's not the fedora-specific hack but the "standard" version, i.e. the implemenation is more than just a specific implementation by fedora, but more cross-platform...

Anyway, I am not glued to "standard", all I know it shouldn't be "bls", and probably shouldn't be "systemd" either. ideas?

src/kernel-install/kernel-install
poettering: this new file is placed in the ESP/XBOOLDR partition, not /etc/kernel/. I.e. it's about describing which format the boot loader installed in the ESP expects its /loader/entries files. And that's different from selecting in which format kernel-install *creates* these entries (whch is stored in /etc/kernel/install.conf, and for which there already is `$layout` field there).

Hence, this new file is purely informational: it reports "hey, we have a boot loader installed where the /loader/entries/ dir expects standards compliant entries". That makes it quite different from the existing /etc/kernel/install.conf entry which is not informational, but is admin-config: it tells kernel-install "hey you, when you write to /loader/entries/, use this format".

Now, the two fields are not entirely unrelated, after all if there's already a boot loader installed, we should generally install entries in the format it understands. But it's also quite different:

1. One day, once kernel-install learns generating type `#2` boot loader spec entries it might make sense to write `echo layout=unified >> /etc/kernel/install.conf` to select that it should create entries in that format now. But  `/loader/entries.srel` should never carry a string such as "unified", because it is purely about the format of entries in the `/loader/entries/` directory

2. if no boot loader spec is installed yet, but kernel-install invoked we require explicit configuration in which format to generate the entries.

poettering: And to reply more explicitly:

> Shoudn't we assume "layout=xxx" when "xxx" is specified in entries.srel? If you use some non-standard entry format, it seems reasonable to assume that you're using a corresponding layout of entries.

If my boot loader supports both our type 1 and type 2 entry types (which is exacly what sd-boot does) then `$ESP/loader/entries.srel` should contain `standard` (as that#s hte format `$ESP/loader/entries/` drop-ins should have. But I might still want to create files in type 2 format instead, hence such an implicaiton looks wrong. Moreover, I think the grub implementation of our spec might actually support our dropins but not vice versa. Hence the choice of what to generate is not that clear: we might even want to generate our format in case the grub format is supported.

> I'm not sure if it makes sense to create a separate file for this. We already have layout.conf: why not add a field there instead? We already put random-seed config in loader.conf.

loader.conf is an sd-boot config file, and not part of the boot loader spec. Do you mean /etc/kernel/install.conf when you say `layout.conf` (not aware of any file by that name). If so, it's nt in the ESP, so doens't work.

docs/BOOT_LOADER_SPECIFICATION.md
behrmann: To maybe add to the bikeshedding: How about a long name like `io.systemd.bls` (or `io.systemd.boot_loader_specification`). It would make clear which standard (or official, upstream or original version) it is and it would nicely delineate it from something like `org.fedora.bls` and implementors could say "we faithfully implement `io.systemd.bls`" or "our flavour is `org.grub.bls`". Not that I would expect a great many implementations, but with the Fedora stuff there's two already.

src/boot/bootctl.c
poettering: I'll simplify it further to: "Read $KERNEL_INSTALL_MACHINE_ID from /etc/machine-info. Please move it to /etc/kernel/entry-token."

i.e. I dropped "token from". Why? because we actually only parse $KERNEL_INSTALL_MACHINE_ID is as 128bit id, and don't allow anything else. Which is a bit different for the entry token which can be any string. Hence using the same word for the data from $KERNEL_INSTALL_MACHINE_ID and /etc/kernel/entry-token might be slightly misleading, given it's different structure. But if we just drop "token" from the sentence this goes away and the sentence still makes sense.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: @behrmann as mentioned, I'd prefer a string that doesn't include the word "systemd", so that we make it a bit politically easier for grub or so to implement this. i.e. I think it's easier for some people to swallow this if there is no pesky reference to systemd in the stuff placed in the ESP...

poettering:
Force pushed a new version. PTAL.

Addressed the raised points, but still uses "standard" as indicator in $BOOT/loader/entries.srel, simply because I still have no better idea what to use, and "bls" I am very sure isn't it, see above for the discussion about that.

Also still uses `$BOOT/loader/entries.srel` as file to encode this instead of `$BOOT/loader/loader.conf` because the latter is an sd-boot specific config file, and I am pretty sure we should not attempt to standardize that as part of the boot loader spec. See above.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: So I guess we can put the string "type1" in there. I think that would work out nicely, no?

poettering:
Force pushed another version now, that changes the string "standard" → "type1". I think that's an overall improvement, and should address @keszybz' concerns?

ptal!

docs/BOOT_LOADER_SPECIFICATION.md
behrmann: I understand the political argument, although I fear this will be an uphill battle as long as this isn't enshrined in POSIX regardless of how this is named, but I guess you're right since some things like `/etc/os-release` have already filtered through as well. 

I think I like "type1" better than the rest, since it sounds less definitive and more inviting for others to use this, although it leaves open who can make up new ones and how they are to be enumerated, that was what I was trying to get at with the more qualified name. I also think that it would make sense to avoid bls, so as to avoid possible confusions with the extended stuff in Fedora.

keszybz: I'd vote for `type1` or `bls-type1`.

src/kernel-install/kernel-install
keszybz: Yeah, I meant `loader.conf`, sorry for the typo.

OK, I guess it's reasonable to add a new file.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -309,6 +309,18 @@ focus for this specification. More specifically, on non-EFI systems
 configuration snippets following this specification cannot be used to spawn
 other operating systems (such as Windows).
 
+Unfortunately, there are implementations of boot loading infrastructure that
+are also using the /loader/entries/ directory, but place files in them that are
+not valid by this specification. In order to minimize confusion a boot loader
+implementation may place a file /loader/entries.srel next to the
```
keszybz, line 315: I forgot to ask this: why "srel"? Is this short for something?

poettering: well, i didn't want to call it "standard", because that was already he string inside it. So I wanted to name it something that tries to communicate "which standard?". So I came up with "srel", i.e. the "standard release"... choose from fedora standard release, or systemd standard release...

But then i wanted it a bit cryptic so that it can mean anything or nothing.

Anyway, not glued to that name. if you have a better idea, i am game.

keszybz: Oh, OK. I guessed so, but I thought that maybe "srel" is already used somewhere. I guess the extension doesn't matter, as long as it's set to something.

PR #22779, opened 2022-03-17
hwdb: Provide HP 2760p ACCEL_MOUNT_MATRIX

AlexCatze:
This commit does 2 things. First it provides correct mount matrix. Second it marks accel as mounted in display. This acccel is in base, but 2760p is 2 in 1 transformer and accel can be used to rotate display in tablet mode. So, it can be used as workaround until iio-sensor-proxy will not support use of base-mounted accels in tablet mode.

### DIFF
--- a/hwdb.d/60-sensor.hwdb
+++ b/hwdb.d/60-sensor.hwdb
@@ -449,7 +449,7 @@ sensor:modalias:acpi:BMA250E*:dmi:*:svnInsyde*:pni101c:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*Hewlett-Packard*:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*HP*:*
  ACCEL_LOCATION=base
- 
+
 # HP Elitebook 2760p
 # Physicaly accel is in base, but 2760p is 2 in 1 transformer, and accel can be used in tablet mode to rotate screen
 sensor:modalias:platform:lis3lv02d:dmi:*svnHewlett-Packard:pnHPEliteBook2760p:*


AlexCatze:
@yuwata , i think that this is CI bug. Because it fails on 443 line, that i didn`t touched.
https://github.com/systemd/systemd/blob/3dfe57400e3a08490ebe8c4258ad36ff3eec8035/hwdb.d/60-sensor.hwdb#L441-L447
Quote from CI log.

> Cannot parse /builddir/build/BUILD/systemd-250/redhat-linux-build/../hwdb.d/60-sensor.hwdb: Expected stringEnd, found '#'  (at char 18744), (line:443, col:1)

hwdb.d/60-sensor.hwdb
```diff
@@ -449,6 +449,12 @@ sensor:modalias:acpi:BMA250E*:dmi:*:svnInsyde*:pni101c:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*Hewlett-Packard*:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*HP*:*
  ACCEL_LOCATION=base
+ 
+#HP Elitebook 2760p
+#Physicaly accel is in base, but 2760p is 2 in 1 transformer, and accel can be used in tablet mode to rotate screen
```
bluca: missing space after `#` on both lines

keszybz:
The patch has adds a trailing whitespace above the block. Remove that, and the test shall pass.

hwdb.d/60-sensor.hwdb
AlexCatze: I tried, this doesn`t helped.

hwdb.d/60-sensor.hwdb
```diff
@@ -450,6 +450,12 @@ sensor:modalias:platform:lis3lv02d:dmi:*svn*Hewlett-Packard*:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*HP*:*
  ACCEL_LOCATION=base
 
+# HP Elitebook 2760p
+# Physicaly accel is in base, but 2760p is 2 in 1 transformer, and accel can be used in tablet mode to rotate screen
```
poettering, line 454: Physicaly → Physical?

hadess:
> So, it can be used as workaround until iio-sensor-proxy will not support use of base-mounted accels in tablet mode.

This isn't the right place to put a work-around like this. We need to implement the functionality in iio-sensor-proxy (https://gitlab.freedesktop.org/hadess/iio-sensor-proxy/-/issues/345), not implement work-arounds like this one that lie about physical characteristics of a device.

Strong NAK.

hadess:
I'm going to close this so it doesn't get merged by mistake. There isn't any way to "fix" this merge request, you'll need to patiently wait for iio-sensor-proxy to gain the functionality before upstreaming any other changes such as mount matrix quirks, if necessary.

hwdb.d/60-sensor.hwdb
keszybz: FWIW, this is exactly the case where an adverb is needed, i.e. this should have been "physically". In this sentence, an adverb would apply to "is", but an adjective would apply to "accel". I guess we should use a proper word, not "accel" too.

PR #22780, opened 2022-03-17
po: fix "sistemd" typo in it.po

Tachi107:


po/it.po
```diff
@@ -66,7 +66,7 @@ msgstr "Ricarica lo stato di systemd"
 
 #: src/core/org.freedesktop.systemd1.policy.in:65
 msgid "Authentication is required to reload the systemd state."
-msgstr "Autenticazione richiesta per riavviare lo stato di sistemd."
+msgstr "Autenticazione richiesta per riavviare lo stato di systemd."
```
bluca: `ricaricare` is the right verb here instead of `riavviare`

Tachi107: Oh I didn't even notice that, I'll fix that too.

Tachi107: Done :)

PR #22786, opened 2022-03-18
Add HP Elitebook 2760p ACCEL_MOUNT_MATRIX

AlexCatze:
Like   #22779 but without ACCEL_LOCATION, just correct mount matrix.

### DIFF
--- a/hwdb.d/60-sensor.hwdb
+++ b/hwdb.d/60-sensor.hwdb
@@ -449,7 +449,7 @@ sensor:modalias:acpi:BMA250E*:dmi:*:svnInsyde*:pni101c:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*Hewlett-Packard*:*
 sensor:modalias:platform:lis3lv02d:dmi:*svn*HP*:*
  ACCEL_LOCATION=base
- 
+
 # Elitebook 2760p
 sensor:modalias:platform:lis3lv02d:dmi:*svnHewlett-Packard:pnHPEliteBook2760p:*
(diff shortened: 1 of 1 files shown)


hadess:
> Like #22779 but without ACCEL_LOCATION, just correct mount matrix.

How did you test this, and which accelerometer orientation did you use for the mount matrix for the base? We have documentation for the display orientation, but not for the base orientation.

As this patch is half of https://github.com/systemd/systemd/pull/22779 it's unlikely that it is in the right orientation for a base display where the Z axis would shoot upwards from the keyboard, in the same way as it shoots away from the display.

AlexCatze:
> How did you test this
I`d test this in tablet mode, with ACCEL_LOCATION=display. 

> it's unlikely that it is in the right orientation for a base display where the Z axis would shoot upwards from the keyboard, in the same way as it shoots away from the display.

In this device vector that perpendicular to keyboard in laptop mode, equals to vector that perpendicular to screen in tablet mode.

hadess:
> I`d test this in tablet mode, with ACCEL_LOCATION=display.

That means that the Z axis is going in the wrong direction. I'm nak'ing this again. I would really rather you waited until iio-sensor-proxy supported this particular scenario, rather than trying to hack around it.

poettering:
So where are we at this? @keszybz did the suggested changes land in iio sensor package?

keszybz:
https://gitlab.freedesktop.org/hadess/iio-sensor-proxy/-/issues/345 is still open.
@hadess said that it should be fixed there first. Let's close this.

PR #22788, opened 2022-03-18
time-util: add macros around timespec_store() that operates on compun…

poettering:
…d literal allocated timespec struct

This way we can convert usec_t to timespec on-the-fly, without a buffer.

No actual behaviour change just some shortening of code.


bluca (approved):

keszybz:
LGTM too.

PR #22791, opened 2022-03-18
Invert order of entries w/o sort-key in sd-boot menu

keszybz:
Follow up for #22519.

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -1683,10 +1683,10 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
                         return r;
         }
 
-        /* Now order by ID (the version is likely part of the ID, thus note that this might put the oldest
-         * version last, not first, i.e. specifying a sort key explicitly is thus generally preferable, to
-         * take benefit of the explicit sorting above.) */
-        r = strverscmp_improved(a->id, b->id);
+        /* Now order by ID. The version is likely part of the ID, thus note that this will generatelly put
+         * the newer versions earlier. Specifying a sort key explicitly is preferable, because it gives an
+         * explicit sort order. */
+        r = -strverscmp_improved(a->id, b->id);
         if (r != 0)
                 return r;
 
--- a/src/shared/bootspec.c
+++ b/src/shared/bootspec.c
@@ -272,6 +272,7 @@ static int boot_entry_compare(const BootEntry *a, const BootEntry *b) {
         r = CMP(!a->sort_key, !b->sort_key);
         if (r != 0)
                 return r;
+
         if (a->sort_key && b->sort_key) {
                 r = strcmp(a->sort_key, b->sort_key);
                 if (r != 0)
@@ -286,7 +287,7 @@ static int boot_entry_compare(const BootEntry *a, const BootEntry *b) {
                         return r;
         }
 
-        return strverscmp_improved(a->id, b->id);
+        return -strverscmp_improved(a->id, b->id);
 }
 
 static int boot_entries_find(
--- a/src/test/test-bootspec.c
+++ b/src/test/test-bootspec.c
@@ -86,9 +86,9 @@ TEST_RET(bootspec_sort) {
         assert_se(streq(config.entries[2].id, "c.conf"));
 
         /* The following ones have no sort key, hence order by version compared ids, lowest first */
-        assert_se(streq(config.entries[3].id, "a-5.conf"));
+        assert_se(streq(config.entries[3].id, "b.conf"));
         assert_se(streq(config.entries[4].id, "a-10.conf"));
-        assert_se(streq(config.entries[5].id, "b.conf"));
+        assert_se(streq(config.entries[5].id, "a-5.conf"));
 
         return 0;
 }


src/shared/bootspec.c
```diff
@@ -437,12 +435,9 @@ static int find_sections(
 
         _cleanup_free_ struct PeSectionHeader *sections = NULL;
         _cleanup_free_ char *osrelease = NULL, *cmdline = NULL;
-        size_t i, n_sections;
-        struct DosFileHeader dos;
-        struct PeHeader pe;
-        uint64_t start;
         ssize_t n;
 
+        struct DosFileHeader dos;
```
poettering, line 441: hu?

src/boot/efi/boot.c
```diff
@@ -1683,10 +1683,10 @@ static INTN config_entry_compare(const ConfigEntry *a, const ConfigEntry *b) {
...
         }
 
-        /* Now order by ID (the version is likely part of the ID, thus note that this might put the oldest
-         * version last, not first, i.e. specifying a sort key explicitly is thus generally preferable, to
-         * take benefit of the explicit sorting above.) */
-        r = strverscmp_improved(a->id, b->id);
+        /* Now order by ID. The version is likely part of the ID, thus note that this will generatelly put
+         * the newer versions earlier. Specifying a sort key explicitly is preferable, because it gives an
+         * explicit sort order. */
+        r = -strverscmp_improved(a->id, b->id);
```
poettering, line 1689: i don't get this?

before arch would be placed before fedora. But now fedora before arch? 

the `id` field is a concatenation of os name and version typically. This means you can't have both when using strverscmp(): either os names *and* versions are ordered in increasing order, or they are ordered in decreasing order. You can't have os names up but versions down...

The thing is that there is no clear naming regime how the ids are picked, and hence it's hard to reconstruct what is os name and what is os version

poettering:
first two commits look good to me. but the third i don't get...

src/shared/bootspec.c
keszybz: What do you mean? It's a variable declaration that was moved down to the point where it is used so that the reader doesn't have to scan the text up and down to see what is happening.

keszybz: Please look at the whole commit: other variables that are defined and used in the same fashion are moved a bit more. This is the first one, so it's move just a bit down.

src/boot/efi/boot.c
keszybz: `id` is actually the filename. And it is usually built as `<machine-id>-<version>.conf`. In that case, the `<machine-id>` part is an opaque string and it does not matter whether we sort them in increasing or decreasing order. But within a single machine-id group, we sort by `<version>`, and this is the point of the patch: within a group of entries with the same machine-id, entries are sorted with newest at the top. I think this is pretty important, because a) non-dual-boot machines are a big majority, b) people have been confused by reversed sort order and it has been a long-standing issue to sort newest-at-the-top, c) this sort order is consistent with the "new" sort order for entries with sort-key.

It is true that the filename might be constructed differently, or that we might sort groups of entries non-alphabetically. But we already did that, e.g. putting auto-windows or macos last. In fact even with sort-key, the entry titles might be sorted alphabetically or not. And as mentioned above, dual-boot machines are the minority. And dual-boot machines where at least two systems are not windows are even rarer. So overall, this patch should make things better for the common case, and is "meh" for the other cases.

For example:
on my machine I have:
`/boot/efi/loader/entries/08a5690a2eed47cf92ac0a5d2e3cf6b0-5.15.14-200.fc35.x86_64.conf`
`/boot/efi/loader/entries/08a5690a2eed47cf92ac0a5d2e3cf6b0-5.17.0-0.rc5.102.fc36.x86_64.conf`
and this is parsed into:
`id = 08a5690a2eed47cf92ac0a5d2e3cf6b0-5.15.14-200.fc35.x86_64.conf`
and `id = 08a5690a2eed47cf92ac0a5d2e3cf6b0-5.17.0-0.rc5.102.fc36.x86_64.conf`
and results in 5.17.0 being sorted as newer then 5.15.14, which is all I care about as a user.

If I add a second system, and I have
`id = 99abbbba2eed47cf92ac0a5d2e3cf6b0-5.1.14-200.fc29.x86_64.conf`
and `id = 99abbbba2eed47cf92ac0a5d2e3cf6b0-4.19.0-0.fc29.x86_64.conf`,
I care that `5.1.14` is above `4.19.0`. Whether this group is above or below the other one isn't really important. Or in other words, the sort order of `99abbbba2eed47cf92ac0a5d2e3cf6b0` and `08a5690a2eed47cf92ac0a5d2e3cf6b0` doesn't matter, as long as it's stable.

keszybz:
focal-ppc64el:
```
[  146.575683] systemd[1]: dev-hvc0.device: Job dev-hvc0.device/start timed out.
[  146.768754] systemd[1]: Timed out waiting for device /dev/hvc0.
[  146.917058] systemd[1]: Dependency failed for Serial Getty on hvc0.
[  162.916545] systemd[1]: testsuite-58.service: Failed with result 'exit-code'.
[  163.051075] systemd[1]: Failed to start TEST-58-REPART.
[  185.279956] systemd-journald[253]: Failed to send stream file descriptor to service manager: Connection refused
TEST-58-REPART: (failed; see logs)
```
Looks like some issue with the builder.

And rpm-build:* failed with `Submit of the build failed: ("Cannot create a new Copr project`, which is a known issue.

PR #22792, opened 2022-03-19
sd-netlink: use synchronous socket for nftables

topimiettinen:
Callers don't expect EINPROGRESS or EWOULDBLOCK errno cases, so it's better to
use a blocking socket. Keep exported asynchronous socket functions for
compatibility.

### DIFF
--- a/src/libsystemd/sd-netlink/netlink-internal.h
+++ b/src/libsystemd/sd-netlink/netlink-internal.h
@@ -142,11 +142,12 @@ uint32_t message_get_serial(sd_netlink_message *m);
 void message_seal(sd_netlink_message *m);
 
 int netlink_open_family(sd_netlink **ret, int family);
+int netlink_open_family_flags(sd_netlink **ret, int family, int flags);
 bool netlink_pid_changed(sd_netlink *nl);
 int netlink_rqueue_make_room(sd_netlink *nl);
 int netlink_rqueue_partial_make_room(sd_netlink *nl);
 
-int socket_open(int family);
+int socket_open_flags(int family, int flags);
 int socket_bind(sd_netlink *nl);
 int socket_broadcast_group_ref(sd_netlink *nl, unsigned group);
 int socket_broadcast_group_unref(sd_netlink *nl, unsigned group);
--- a/src/libsystemd/sd-netlink/netlink-message-nfnl.c
+++ b/src/libsystemd/sd-netlink/netlink-message-nfnl.c
@@ -329,3 +329,7 @@ int sd_nfnl_nft_message_add_setelem_end(sd_netlink_message *m) {
 int sd_nfnl_socket_open(sd_netlink **ret) {
         return netlink_open_family(ret, NETLINK_NETFILTER);
 }
+
+int sd_nfnl_socket_open_flags(sd_netlink **ret, int flags) {
+        return netlink_open_family_flags(ret, NETLINK_NETFILTER, flags);
+}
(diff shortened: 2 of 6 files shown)


topimiettinen:
TEST-02-UNITTESTS fails in nspawn-root, could be related.

topimiettinen:
It's strange that `test-firewall-util` (tests adding and removing sets of IP addresses) doesn't catch the bug I saw in relation to #22587. Running the test 1000 times doesn't show anything. When run in parallel (32x), there are other errors like `EEXIST` but that's probably because the test is confused by other instances. Grepping the kernel tree for `EINPROGRESS` doesn't show obvious candidates.

poettering:
What's the story here? Is this addressing a real bug you saw?

iirc netlink doesn't really have a socket buffer from userspace to kernel? so, did you actually managed to get an error from the kernel? How can one reproduce this?

topimiettinen:
The background for this is that an early version of #22587 opened a new netlink socket for every request. Sometimes there were failures and stracing showed `EINPROGRESS` as the error in those cases. But later versions reuse an existing netlink socket, so there are no more failures. I suppose our code base doesn't exercise this case well and the bug appears very rarely if at all. But perhaps other users of sd-netlink have such use cases.

Maybe a test in `test-firewall-util.c` could try something many times and open a new socket for each operation and close it each time. I didn't see any failures with the current tests, but they open the socket only once.

I don't know how netlink is implemented in kernel, this might indicate a bug there also or maybe a rate limit could be triggered.

Another solution could be to just document the fact that the socket is asynchronous, so `EINPROGRESS` errors should be expected by callers, maybe also `EWOULDBLOCK` and `EAGAIN` to be safe.

poettering:
> The background for this is that an early version of #22587 opened a new netlink socket for every request. Sometimes there were failures and stracing showed `EINPROGRESS` as the error in those cases. But later versions reuse an existing netlink socket, so there are no more failures. I suppose our code base doesn't exercise this case well and the bug appears very rarely if at all. But perhaps other users of sd-netlink have such use cases.

Hmm, so you are saying that it's not flooding with send() requests that causes EINPROGRESS for you, but flooding with socket() calls? Interesting...

I wonder if this means that we should create the socket without O_NONBLOCK and only turn on O_NONBLOCK afterwards?

poettering:
(btw EWOULDBLOCK and EAGAIN are aliases, it's the same error code)

poettering:
I'd love a reproducer for the original issue.

poettering:
My educated guess is that the "genl" lookup is actually causing the issue here

topimiettinen:
> > The background for this is that an early version of #22587 opened a new netlink socket for every request. Sometimes there were failures and stracing showed `EINPROGRESS` as the error in those cases. But later versions reuse an existing netlink socket, so there are no more failures. I suppose our code base doesn't exercise this case well and the bug appears very rarely if at all. But perhaps other users of sd-netlink have such use cases.
> 
> Hmm, so you are saying that it's not flooding with send() requests that causes EINPROGRESS for you, but flooding with socket() calls? Interesting...
> 
> I wonder if this means that we should create the socket without O_NONBLOCK and only turn on O_NONBLOCK afterwards?

That could work too.

I tried a few simple tests like calling `sd_genl_socket_open()` repeatedly and then also with a small message written, also hand written raw socket calls to drop extra operations but I didn't see any failures with 10000 runs. Maybe  758fdb8762248b052bf143dfcb8de2021ee5537d did something weird.

poettering:
OK, then let's close this for now. If this resurfaces we can look into it again and figure something out.

PR #22794, opened 2022-03-20
Document and test LogNamespace in user units

bluca:
Starting the system unit instance manually is required for now.

### DIFF
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -2912,7 +2912,12 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
         <citerefentry><refentrytitle>journalctl</refentrytitle><manvolnum>1</manvolnum></citerefentry>
         output, unless the <option>--namespace=</option> option is used.</para>
 
-        <xi:include href="system-only.xml" xpointer="singular"/></listitem>
+        <xi:include href="system-or-user-ns.xml" xpointer="singular"/>
+
+        <para>For services running in per-user instances of the service manager, note that it is also
+        required to manually start a system <literal>systemd-journald@<replaceable>NS</replaceable>.service</literal>
+        unit, as it is not possible to have an implicit dependency from a user unit to a system unit.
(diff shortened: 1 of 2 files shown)


bluca:
looks like the test is racy, will rework it to wait for the namespaced journal to be up

man/systemd.exec.xml
```diff
@@ -2912,7 +2912,12 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
         <citerefentry><refentrytitle>journalctl</refentrytitle><manvolnum>1</manvolnum></citerefentry>
         output, unless the <option>--namespace=</option> option is used.</para>
 
-        <xi:include href="system-only.xml" xpointer="singular"/></listitem>
+        <xi:include href="system-or-user-ns.xml" xpointer="singular"/>
+
+        <para>For services running in per-user instances of the service manager, note that it is also
+        required to manually start a system <literal>systemd-journald@<replaceable>NS</replaceable>.service</literal>
+        unit, as it is not possible to have an implicit dependency from a user unit to a system unit.
+        </para></listitem>
```
poettering, line 2920: hmm, we could of course look into making journald work in user mode? i think it mostly should work already?

poettering:
also, the service should not be up, the sockets for it should. i.e. the per-instance journald server is exiting on idle, hence ordering against that pointless

man/systemd.exec.xml
bluca: yes I already added an item for tomorrow's meeting agenda on this

bluca:
> also, the service should not be up, the sockets for it should. i.e. the per-instance journald server is exiting on idle, hence ordering against that pointless

The socket units have `StopWhenUnneeded=yes` so both are needed - but the service requires the socket, so it's not necessary to start it by hand

poettering:
lgtm, but you still marked this as draft?

bluca:
> lgtm, but you still marked this as draft?

iirc we talked about this in a meeting some time ago, and the full solution was to wait for per-user journal instances instead

PR #22796, opened 2022-03-21
Journal fix when var is full

fbuihuu:


src/libsystemd/sd-journal/journal-file.c
```diff
@@ -3447,6 +3447,8 @@ int journal_file_open(
 
         if (f->last_stat.st_size == 0 && f->writable) {
 
+                newly_created = true;
```
poettering: this improves the situation a bit, but there are various earlier causes that might fail, hence is not a comprehensive solution. in particular mmap_cache_add_fd() is something that can fail. I think it would be best if the open() call would directly tell us whether it created a new file or not.

I prepped some code for this now → #22808

Please review, and then rebase this one

src/journal/managed-journal-file.c
```diff
@@ -512,19 +512,25 @@ int managed_journal_file_open_reliably(
...
                 return r;
 
         /* The file is corrupted. Rotate it away and try it again (but only once) */
-        log_warning_errno(r, "File %s corrupted or uncleanly shut down, renaming and replacing.", fname);
+        log_warning_errno(r, "File %s corrupted or uncleanly shut down, trying to move it out of the way: %m", fname);
 
         r = journal_file_dispose(AT_FDCWD, fname);
-        if (r < 0)
+        if (r < 0) {
+                log_warning_errno(r, "Failed to move %s out of the way: %m", fname);
```
poettering, line 517: if you propagate the error and log here this means we'll log twice here. that sucks.

src/journal/managed-journal-file.c
```diff
@@ -495,8 +495,8 @@ int managed_journal_file_open_reliably(
```
poettering, line 489: this commit looks strange to me. i.e. shoehorning this into a function called "open" is wrong. apparently you don't want to open the file in this case, so it shouldn't be in a function called "open".

Hence, split out the stuff you actually are looking for into one or more new functions, and then call those from managed_journal_file_open_reliably(), but don't add "magic" flags to this funciton that does stuff in some magic conditions or does not.

src/journal/managed-journal-file.c
fbuihuu: The 4th patch should remove the redundancy, hence when `vacuum_offline_user_journals()` fails we should see:

```
File system.journal corrupted or uncleanly shut down, trying to move it out of the way: %m
Failed to move system.journal out of the way: %m
Failed to read journal file system.journal for rotation, ignoring: %m
```
which seems OK, no ?

src/libsystemd/sd-journal/journal-file.c
fbuihuu: yes agreed.

src/journal/managed-journal-file.c
fbuihuu: Yes I was pondering that too but I finally decided to go with the "magic" flags because it was already the case: passing  O_RDONLY prevents `managed_journal_file_open_reliably()` from disposing corrupted journals. That is also the case if O_CREAT is not passed.

The patch only changed the meaning of O_CREAT: it is used by the function to decide whether to create a new file if the old one was corrupted and disposed. It seemed that this new meaning made more sense actually.

src/journal/journald-server.c
```diff
@@ -614,19 +606,10 @@ static int vacuum_offline_user_journals(Server *s) {
...
                 if (r < 0) {
-                        log_warning_errno(r, "Failed to read journal file %s for rotation, trying to move it out of the way: %m", full);
-
-                        r = journal_file_dispose(dirfd(d), de->d_name);
-                        if (r < 0)
-                                log_warning_errno(r, "Failed to move %s out of the way, ignoring: %m", full);
-                        else
-                                log_debug("Successfully moved %s out of the way.", full);
-
+                        log_warning_errno(r, "Failed to read journal file %s for rotation, ignoring: %m", full);
```
keszybz, line 607: Preexisting condition, but since you're touching the line: "failed to read" sounds strange. Shouldnt' this just be "failed to open" ?

src/journal/managed-journal-file.c
keszybz: I think it's fine as is. I wanted to write that that the function should be renamed, e.g. from `_open_reliably(`) to `_rename_or_open_reliably()` or something like that, but this doesn't seem needed. After all, `open()` without `O_CREAT` will not create a file either and will return an error. So I think it's fine to make this conditional in a similar way here. But please add a comment at the top of the function that describes the logic.

keszybz (changes requested):

src/journal/managed-journal-file.c
```diff
@@ -508,19 +508,25 @@ int managed_journal_file_open_reliably(
...
         /* The file is corrupted. Rotate it away and try it again (but only once) */
-        log_warning_errno(r, "File %s corrupted or uncleanly shut down, renaming and replacing.", fname);
+        log_warning_errno(r, "File %s corrupted or uncleanly shut down, trying to move it out of the way: %m", fname);
 
         r = journal_file_dispose(AT_FDCWD, fname);
-        if (r < 0)
+        if (r < 0) {
+                log_warning_errno(r, "Failed to move %s out of the way: %m", fname);
+                return r;
+        }
```
poettering, line 519: merge the two lines with `return log_warning_errno(r, …`

src/journal/managed-journal-file.c
```diff
@@ -491,8 +491,8 @@ int managed_journal_file_open_reliably(
 
         int r;
 
-        r = managed_journal_file_open(-1, fname, open_flags, file_flags, mode, compress_threshold_bytes, metrics,
-                               mmap_cache, deferred_closes, template, ret);
+        r = managed_journal_file_open(-1, fname, open_flags, file_flags, mode, compress_threshold_bytes,
+                                      metrics, mmap_cache, deferred_closes, template, ret);
```
poettering, line 493: please drop this change, as it doesn't actually change anything?

src/journal/managed-journal-file.c
poettering: also, even if the whole PR is merged, you'll log about this warning here twice. here and in the caller where you'll warn again

src/journal/journald-server.c
```diff
@@ -589,18 +588,11 @@ static int vacuum_offline_user_journals(Server *s) {
...
-                        continue;
-                }
-
                 /* Make some room in the set of deferred close()s */
                 server_vacuum_deferred_closes(s);
 
                 /* Open the file briefly, so that we can archive it */
-                r = managed_journal_file_open(
+                r = managed_journal_file_open_reliably(
                                 fd,
```
poettering: you removed the `fd` variable above. How does this commit even compile?

poettering (changes requested):

fbuihuu:
After a fresh look it seems that the last patch introduces a behavior change that I didn't notice previously:  in `vacuum_offline_user_journals()` , if `managed_journal_file_open()` fails to open the journal file then the file is *always* disposed whereas with the last patch applied (`managed_journal_file_open()` replaced with `managed_journal_file_open_reliably()`)  the journal file is disposed only for these specific errors: 

```
                    -EBADMSG,           /* Corrupted */
                    -ENODATA,           /* Truncated */
                    -EHOSTDOWN,         /* Other machine */
                    -EPROTONOSUPPORT,   /* Incompatible feature */
                    -EBUSY,             /* Unclean shutdown */
                    -ESHUTDOWN,         /* Already archived */
                    -EIO,               /* IO error, including SIGBUS on mmap */
                    -EIDRM,             /* File has been deleted */
                    -ETXTBSY))          /* File is from the future */

```

otherwise the problematic journal file is left unchanged.

I'm not sure whether this behavior change is acceptable or not...

PR #22797, opened 2022-03-21
basic/strv: avoid potential UB with references to array[-1]

keszybz:
"""
Given an array a[N] of N elements of type T:
- Forming a pointer &a[i] (or a + i) with 0 ≤ i ≤ N is safe.
- Forming a pointer &a[i] with i < 0 or i > N causes undefined behavior.
- Dereferencing a pointer &a[i] with 0 ≤ i < N is safe.
- Dereferencing a pointer &a[i] with i < 0 or i ≥ N causes undefined behavior.
"""

As pointed by by @medhefgo, here we were forming a pointer to a[-1]. a itself
wasn't NULL, so a > 0, and a-1 was also >= 0, and this didn't seem to cause any
problems. But it's better to be formally correct, especially if we move the
code to src/fundamental/ later on and compile it differently.

Compilation shows no size change (with -O0 -g) on build/systemd, so this should
have no effect whatsoever.

### DIFF
--- a/src/basic/strv.h
+++ b/src/basic/strv.h
@@ -133,8 +133,8 @@ bool strv_overlap(char * const *a, char * const *b) _pure_;
                                 size_t _len = strv_length(h);           \
                                 _len > 0 ? h + _len - 1 : NULL;         \
                         });                                             \
-             i && (s = i) >= h;                                         \
-             i--)
+             (s = i);                                                   \
+             i > h ? i-- : (i = NULL))
 
 #define STRV_FOREACH_BACKWARDS(s, l)                                    \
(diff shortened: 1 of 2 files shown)


poettering:
lgtm. somehow i have the suspicion we might have more for loops like this in our codebase which make the same mistake...

src/basic/strv.h
```diff
@@ -133,8 +133,8 @@ bool strv_overlap(char * const *a, char * const *b) _pure_;
                                 size_t _len = strv_length(h);           \
                                 _len > 0 ? h + _len - 1 : NULL;         \
                         });                                             \
-             i && (s = i) >= h;                                         \
-             i--)
+             (s = i);                                                   \
+             i > h ? i-- : (i = NULL))
```
poettering, line 137: btw, i think `i = i > h ? i-1 : NULL` for the last expression is a tiny bit more readable

PR #22800, opened 2022-03-21
Add helper macro PTR_SUB1() to deal with backwards iteration through arrays without UB

poettering:
Inspired by #22797.

### DIFF
--- a/src/shared/dns-domain.c
+++ b/src/shared/dns-domain.c
@@ -168,24 +168,18 @@ int dns_label_unescape_suffix(const char *name, const char **label_terminal, cha
 
         /* Skip current terminal character (and accept domain names ending it ".") */
         if (*terminal == 0)
-                terminal--;
+                terminal = PTR_SUB1(terminal, name);
         if (terminal >= name && *terminal == '.')
-                terminal--;
+                terminal = PTR_SUB1(terminal, name);
 
         /* Point name to the last label, and terminal to the preceding terminal symbol (or make it a NULL pointer) */
-        for (;;) {
-                if (terminal < name) {
-                        /* Reached the first label, so indicate that there are no more */
-                        terminal = NULL;
-                        break;
-                }
-
+        while (terminal) {
                 /* Find the start of the last label */
                 if (*terminal == '.') {
                         const char *y;
                         unsigned slashes = 0;
 
-                        for (y = terminal - 1; y >= name && *y == '\\'; y--)
+                        for (y = PTR_SUB1(terminal, name); y && *y == '\\'; y = PTR_SUB1(y, name))
                                 slashes++;
 
                         if (slashes % 2 == 0) {
@@ -198,7 +192,7 @@ int dns_label_unescape_suffix(const char *name, const char **label_terminal, cha
                         }
                 }
 
-                terminal--;
+                terminal = PTR_SUB1(terminal, name);
         }
 
         r = dns_label_unescape(&name, dest, sz, 0);


src/basic/macro.h
```diff
@@ -463,4 +463,20 @@ typedef struct {
...
+ * for (p = end-1; p >= base; p--) …
+ *
+ * But this is not safe because p before the base is UB in C. With this macro the loop becomes this instead:
+ *
+ * for (p = PTR_SUB1(end, base); p; p = PTR_SUB1(p, base)) …
+ *
+ * And is free from UB! */
+#define PTR_SUB1(p, base)                                \
+        ({                                               \
+                typeof(p) _q = (p);                      \
```
bluca, line 478: is `_q` here safe for nesting? Should it use `UNIT_T`?

poettering: no, it's not safe for nesting. But I am not sure we should complicate stuff with the UNIQ_T stuff unless it's very likely or actually used in a nested way. At least I have serious trouble still following the macros if UNIQ_T is in the mix.

Hence, yes, UNIQ_T where necessary. But where it likely isn't, let's maybe avoid this until it becomes an issue or likely becomes an issue.

I mean, for loops are likely going to be nested, hence using UNIQ_T for the iterator variables of macros like STRV_FOREACH makes a ton of sense. But I doubt people will use `PTR_SUB1(PTR_SUB1(x, foo), bar)`. I mean, we have a pretty good idea here how this is going to be used, I even explained that in the comment, and there is no nesting there.

Thankfully all current compilers will generate very explicit warnings once you actually end up nesting this, so there aren't any mistakes possible here: if people actually nest it, the compiler will give you a clear hint that the macro first needs some love...

keszybz:
I wrote https://github.com/systemd/systemd/pull/22802 in the meantime. I think 22802 should go in first, because it simplifies the code in most cases.

src/basic/macro.h
bluca: makes sense

src/basic/macro.h
```diff
@@ -463,4 +463,20 @@ typedef struct {
...
+                _q && _q > (base) ? &_q[-1] : NULL;      \
```
medhefgo, line 479: `&_q[-1]` works? O_O

Never seen this, imho explicit pointer arithmetic makes this easier to grok.

poettering: > `&_q[-1]` works? O_O

what's worse: `(-1)[q]` works too. Yes, you are reading right... In C `a[b]` is just syntactic sugar for `*(a + b)`, and since additions are commutative, `a[b]` is identical to `b[a]`...

Fuck yeah, C FTW!

src/test/test-macro.c
```diff
@@ -433,4 +433,27 @@ TEST(DECIMAL_STR_MAX) {
         assert_se(DECIMAL_STR_MAX(uint64_t) == DECIMAL_STR_WIDTH(u64_longest)+1);
 }
 
+TEST(PTR_SUB1) {
+        static const uint64_t x[4] = { 2, 3, 4, 5 };
```
medhefgo, line 437: Also, any particular reason why not just `int`?

src/basic/macro.h
poettering: the reason i didn't do pointer arith here, is to ensure PTR_SUB1() is only invokable on pointers/arrays and not on numeric values. Because one thing that C actually *does* is refuse the [] syntax if none of the involved arguments is a pointer/array...

medhefgo: Wow. I just tested this and `1[q]` actually works. I am disgusted.

src/test/test-macro.c
poettering: i wanted something large (i.e. definitely not byte sized) that ensures we count correctly, i.e. take the pointer type size into account correctly. Now `int` are not byte sized either on our relevant archs, but it still felt better to make this explicit by picking a fixed size integer...

poettering:
The CI sanitizer failures puzzle me... These macros are supposed to avoid UB behaviour, but apparently the sanitizers think it triggers even more? anyone gets this?

medhefgo:
> The CI sanitizer failures puzzle me... These macros are supposed to avoid UB behaviour, but apparently the sanitizers think it triggers even more? anyone gets this?

My best guess: The consumer of `skip_slash_or_dot_backward` probably relies on `q < path` in some way (as opposed to `q == NULL` with this patch).

src/core/namespace.c
```diff
@@ -775,12 +775,11 @@ static void drop_nop(MountEntry *m, size_t *n) {
                         bool found = false;
 
                         /* Now let's find the first parent of the entry we are looking at. */
-                        for (p = t-1; p >= m; p--) {
+                        for (p = PTR_SUB1(t, m); p; p = PTR_SUB1(p, m))
```
yuwata: In general, I'd like to split the iterator and `found` entry, and make the iterator defined in `for`. How about the following?
```
MountEntry *found = NULL;

for (MountEntry *p = PTR_SUB1(t, m); p; p = PTR_SUB1(p, m))
        if (path_startswith(...)) {
                found = p;
                break;
        }
```

src/basic/path-util.c
```diff
@@ -930,7 +930,7 @@ int path_find_first_component(const char **p, bool accept_dot_dot, const char **
 static const char *skip_slash_or_dot_backward(const char *path, const char *q) {
         assert(path);
 
-        for (; q >= path; q--) {
+        for (; q; q = PTR_SUB1(q, path)) {
```
yuwata, line 934: Maybe, it is better to add `assert(!q || q >= path);`, as previously `q >= path` is checked for the initially provided value of `q`, but currently not.

yuwata (changes requested):
LGTM. But two points.

yuwata:
> I wrote #22802 in the meantime. I think 22802 should go in first, because it simplifies the code in most cases.

I prefer this PR, and I'd like to merge this first, and merge remaining cleanups in #22802 later.

yuwata:
One commit is cherry-picked from #22802, and added two commits to address my comments in the above. I hope the CI failures are fixed now.

src/shared/dns-domain.c
```diff
@@ -185,7 +185,7 @@ int dns_label_unescape_suffix(const char *name, const char **label_terminal, cha
                         const char *y;
```
keszybz, line 179: `terminal--;` above is UB.

keszybz (changes requested):

yuwata:
@keszybz The above point is also addressed. PTAL.

poettering:
ah, excellent. thank you for picking this up!

medhefgo (approved):
(*grumble* all this because c sucks *grumble*)

PR #22802, opened 2022-03-21
A few more cases which theoretically could constitute UB

keszybz:


src/udev/udev-rules.c
```diff
@@ -1830,9 +1830,9 @@ static int udev_rule_apply_token_to_event(
                         bool found = false;
 
                         /* Drop the last line. */
-                        for (char *p = buf + strlen(buf) - 1; p >= buf; p--)
-                                if (strchr(NEWLINE, *p)) {
-                                        *p = '\0';
+                        for (ssize_t i = strlen(buf) - 1; i >= 0; i--)
```
poettering: the implicit dropping of the sign is a bit sloppy...

poettering:
tbh i prefer using pointer based iteration where we naturally deal in pointers. switching to numeric indexes for that feels like chickening out to me tbh

src/udev/udev-rules.c
medhefgo: `size_t` has no sign? What I am more interested in is: can strlen (malloc really) return more data than can fit into `ssize_t`?

poettering: On UNIX not really, since we have plenty APIs which assume you cannot have larger objects that SSIZE_MAX. read()/write() for starters which accept size_t and return ssize_t. But in generic C at least theoretical nothing stops you from having SSIZE_MAX+1 large objects if you feel like it.

poettering:
i looked through the PR. I am not a fan of using ssize_t based indexes here I must say, I think sticking to pointers appears nicer to me, because more readable (i.e. only one variable instead of two) and in theory more correct since on some theoretic machine objects that fit into size_t but not into ssize_t are thinkable... 

The other stuff, i.e. everything that's not the `ssize_t`-ification in this PR looks good to me though.

keszybz:
I dropped the last patch, and adjusted one other to use `size_t` instead of `ssize_t`. PTAL.

yuwata:
The first commit is cherry picked into #22800.

keszybz:
Let's wait for #22800 to be merged then.

keszybz:
I think there's nothing useful left here.

PR #22803, opened 2022-03-21
meson: Add support for building efi binaries on multilib

medhefgo:
This allows building 32bit versions of efi binaries on x86_64 machines and vice-versa by passing "-Defi-cflags=-m32" to meson, provided the 32bit gnu-efi and gcc-multilib are available.

It is expected that distros that want to provide both ia32 and x64 versions to use a second build dir to build the non-native version by adding -m32 to efi-cflags and then running the sd-boot/sd-stub ninja target directly.

This is much simpler than doing a regular meson-powered cross build (which is still supported), as the 32bit systemd dependencies do not need to be provided to get ia32 EFI binaries on a otherwise x86_64-only distro (such as arch).
This is mostly so that such distros can easily provide the bootloader binaries for mixed EFI machines (32bit EFI on 64bit capable CPU) alongside x64 binaries without having to jump through the cross build hoop, only to discard the vast majority of it.

See #22550 for the initial discussion.

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -9,7 +9,7 @@ success() { echo >&2 -e "\033[32;1m$1\033[0m"; }
 
 ARGS=(
     "--optimization=0"
-    "--optimization=s"
+    "--optimization=s -Dgnu-efi=true -Defi-cflags=-m32 -Defi-libdir=/usr/lib32"
     "--optimization=3 -Db_lto=true -Ddns-over-tls=false"
     "--optimization=3 -Db_lto=false"
     "--optimization=3 -Ddns-over-tls=openssl"
@@ -27,6 +27,7 @@ PACKAGES=(
     kbd
     libblkid-dev
     libbpf-dev
+    libc6-dev-i386
     libcap-dev
     libcurl4-gnutls-dev
     libfdisk-dev
@@ -93,7 +94,7 @@ elif [[ "$COMPILER" == gcc ]]; then
     # Latest gcc stack deb packages provided by
     # https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test
     add-apt-repository -y ppa:ubuntu-toolchain-r/test
-    PACKAGES+=("gcc-$COMPILER_VERSION")
+    PACKAGES+=("gcc-$COMPILER_VERSION" "gcc-$COMPILER_VERSION-multilib")
 else
     fatal "Unknown compiler: $COMPILER"
 fi


.github/workflows/build_test.sh
```diff
@@ -93,7 +93,7 @@ elif [[ "$COMPILER" == gcc ]]; then
     # Latest gcc stack deb packages provided by
     # https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test
     add-apt-repository -y ppa:ubuntu-toolchain-r/test
-    PACKAGES+=("gcc-$COMPILER_VERSION")
+    PACKAGES+=("gcc-$COMPILER_VERSION gcc-$COMPILER_VERSION-multilib")
```
bluca: This gets quoted and confuses apt - I think you need to add it on a new line

.github/workflows/build_test.sh
```diff
@@ -9,7 +9,7 @@ success() { echo >&2 -e "\033[32;1m$1\033[0m"; }
 
 ARGS=(
     "--optimization=0"
-    "--optimization=s"
+    "--optimization=s -Dgnu-efi=true -Defi-cflags=-m32 -Defi-libdir=/usr/lib32"
```
bluca, line 12: on debian-based the canonical location would be `/usr/lib/i386-linux-gnu/` but maybe this is a slightly different use case

medhefgo: The gnu-efi package lists ia32 stuff in `/usr/lib32`, so this should be correct.

bluca:
```
/usr/bin/gcc-10 -c ../src/boot/efi/boot.c -o src/boot/efi/boot.c.o -Wno-format-signedness -Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wendif-labels -Werror=format=2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=overflow -Werror=override-init -Werror=return-type -Werror=shift-count-overflow -Werror=shift-overflow=2 -Werror=undef -Wfloat-equal -Wimplicit-fallthrough=5 -Winit-self -Wlogical-op -Wmissing-include-dirs -Wmissing-noreturn -Wnested-externs -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wshadow -Wstrict-aliasing=2 -Wstrict-prototypes -Wsuggest-attribute=noreturn -Wunused-function -Wwrite-strings -Wno-maybe-uninitialized -Wno-unused-result -fno-stack-protector -fno-strict-aliasing -fpic -fwide-exec-charset=UCS2 -Wall -Wextra -Wsign-compare -nostdlib -std=gnu99 -ffreestanding -fshort-wchar -fvisibility=hidden -isystem /usr/include/efi -isystem /usr/include/efi/ia32 -I /home/runner/work/systemd/systemd/src/fundamental -DSD_BOOT -DGNU_EFI_USE_MS_ABI -include src/boot/efi/efi_config.h -include version.h -m32 -mno-sse -mno-mmx -Werror -ggdb -DEFI_DEBUG -Os -Werror
../src/boot/efi/boot.c: In function ‘boot_windows_bitlocker’:
../src/boot/efi/boot.c:2048:23: error: argument to variable-length array may be too large [-Werror=vla-larger-than=]
 2048 |                 CHAR8 buf[block_io->Media->BlockSize];
      |                       ^~~
```

poettering:
Would love @keszybz' take on this.

bluca (approved):
lgtm

src/boot/efi/meson.build
```diff
@@ -182,7 +182,7 @@ efi_cflags = cc.get_supported_arguments(
         '-DGNU_EFI_USE_MS_ABI',
         '-include', efi_config_h,
         '-include', version_h,
-]
+] + get_option('efi-cflags')
```
keszybz: I was wondering about the ordering here, but putting the user-specified options last is appropriate. This will allow the user to possibly override our default flags if necessary.

src/boot/efi/meson.build
```diff
@@ -54,14 +63,17 @@ if efi_ld == 'auto'
         endif
 endif
 
+efi_multilib = '/usr/lib' / run_command(
+        cc.cmd_array(),
+        '-print-multi-os-directory',
+        get_option('efi-cflags'),
+        check: false
+).stdout().strip()
```
keszybz: Hmm, doing `run_command` as an operand of `/` seems a bit too much. Maybe split that into two steps?

medhefgo:
I moved the cflags further down so it really does come after our own provided flags. And I edited the PR rationale.

PR #22804, opened 2022-03-21
A couple of TEST-69 tweaks

mrc0mmand:
I noticed a couple of spurious TEST-69 fails in the recent runs, so let's make it a bit easier to see what's going on when things go south.

### DIFF
--- a/test/TEST-69-SHUTDOWN/test.sh
+++ b/test/TEST-69-SHUTDOWN/test.sh
@@ -24,7 +24,12 @@ EOF
 test_append_files() {
     local workspace="${1:?}"
     # prevent shutdown in test suite, the expect script does that manually.
-    rm "${workspace:?}/usr/lib/systemd/tests/testdata/units/end.service"
+    mkdir -p "${workspace:?}/etc/systemd/system/end.service.d"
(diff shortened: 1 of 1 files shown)


test/TEST-69-SHUTDOWN/test.sh
```diff
@@ -13,18 +13,23 @@ _ORIG_NSPAWN="${SYSTEMD_NSPAWN:?}"
...
+EOF
     chmod 755 "${STATEDIR:?}"/run-nspawn
 }
 
 test_append_files() {
     local workspace="${1:?}"
     # prevent shutdown in test suite, the expect script does that manually.
-    rm "${workspace:?}/usr/lib/systemd/tests/testdata/units/end.service"
+    mkdir -p "${workspace:?}/etc/systemd/system/end.service.d"
+    cat >"${workspace}/etc/systemd/system/end.service.d/99-override.conf" <<EOF
```
bluca: `${workspace:?}` ?

mrc0mmand: I skipped it intentionally, since we already check the variable in the previous statement (I guess I should've dropped the braces as well).

PR #22808, opened 2022-03-21
journal: add openat() wrapper that reports file creation, and use it in journal

poettering:


poettering:
Inspired by #22796

bluca (approved):

src/basic/fs-util.c
```diff
@@ -1083,3 +1083,42 @@ int open_mkdir_at(int dirfd, const char *path, int flags, mode_t mode) {
 
         return TAKE_FD(fd);
 }
+
+int openat_report_new(int dirfd, const char *pathname, int flags, mode_t mode, bool *ret_newly_created) {
+        unsigned attempts = 7;
+
+        /* Just like openat(), but adds one thing: optionally returns whether we created the file anew or if
+         * it already existed before. This is only relevant of O_CREAT is set without O_EXCL, and thus will
```
fbuihuu, line 1091: typo 'of' -> 'if'

src/basic/fs-util.c
```diff
@@ -1083,3 +1083,42 @@ int open_mkdir_at(int dirfd, const char *path, int flags, mode_t mode) {
...
+         * shortcut to openat() otherwise */
+
+        if (!FLAGS_SET(flags, O_CREAT) || FLAGS_SET(flags, O_EXCL) || !ret_newly_created)
```
fbuihuu, line 1094: Since the main (only ?) reason to use this function is to figure out whether the file has been created or not, I would make the presence of `ret_newly_created` mandatory.

src/basic/fs-util.c
```diff
@@ -1083,3 +1083,42 @@ int open_mkdir_at(int dirfd, const char *path, int flags, mode_t mode) {
...
+                        *ret_newly_created = true;
+                        return fd;
+                }
+                if (errno != EEXIST)
+                        return -errno;
+
+                /* Hmm, so now we got EEXIST? So it apparently exists now? If so, let's try to open again
+                 * without the two flags. But let's not spin forever, hnce put a limit on things */
+
+                if (--attempts == 0) /* Give up eventually, somebody is playing with us */
```
fbuihuu, line 1121: Does this attempt to handle a race really necessary ? that makes this function differs with `openat()` in this regard and it seems that returning an error when the race happens is just fine anyway.

src/test/test-fs-util.c
```diff
@@ -970,6 +970,56 @@ TEST(open_mkdir_at) {
...
+        fd = openat_report_new(AT_FDCWD, j, O_RDWR|O_CREAT, 0666, &b);
+        assert_se(fd >= 0);
+        fd = safe_close(fd);
+        assert_se(b);
+
+        fd = openat_report_new(AT_FDCWD, j, O_RDWR|O_CREAT, 0666, &b);
+        assert_se(fd >= 0);
+        fd = safe_close(fd);
+        assert_se(!b);
+
```
fbuihuu, line 1010: Aren't these 2 blocks redundant with the first 3 ones ?

src/basic/fs-util.c
fbuihuu: Also whenever `ret_newly_created` is passed, we should initialize it. I think the caller can expect that regardless of the value of the flags.

poettering: > Since the main (only ?) reason to use this function is to figure out whether the file has been created or not, I would make the presence of ret_newly_created mandatory.

dunno, people might want to call this from some wrapper call without having to synthesize an argument if it isn't clear if the caller of that wrapper needs this parameter, hence i think we should make return params optionall all the way through the innermost stack frames these days if it even vaguely makes sense...

> Also whenever ret_newly_created is passed, we should initialize it. I think the caller can expect that regardless of the value of the flags.

indeed, this is a bug. will fix.

src/basic/fs-util.c
poettering: I think hitting the race is pretty easy. just invoke two instances of journal-remote tools  or so at once that intend to write the same file...

src/test/test-fs-util.c
poettering: well, testing is redundant as a whole if you so will... just checking that this also works after unlinking the file...

PR #22810, opened 2022-03-21
meson: Detect python instead of hard-coding python3

heirecka:


meson.build
```diff
@@ -638,7 +638,9 @@ endif
 
 ############################################################
 
-if run_command('python3', '-c', 'import jinja2', check : false).returncode() != 0
+pymod = import('python')
+python = pymod.find_installation('python3')
```
kepstin: If you aren't using any of the other functionality of the python module, it's better to use the generic method `find_program('python3')` instead.

(This uses the exact same method for finding the python executable as the python module does, but does less extra work since it doesn't also try to figure out how to build and install python libraries.) (edit: I was mistaken, it doesn't use the *same* method, but it does have some special fallback code for using the python meson was run with if it can't find a 'python3' binary to use)

heirecka: Updated the PR with this suggestion.

poettering:
lacks rationale

meson.build
```diff
@@ -638,7 +638,8 @@ endif
 
 ############################################################
 
-if run_command('python3', '-c', 'import jinja2', check : false).returncode() != 0
+python = find_program('python3')
+if run_command(python, '-c', 'import jinja2', check : false).returncode() != 0
```
eli-schwartz, line 642: This is semantically identical, since the first thing run_command does with the initial argument is to internally run the find_program() implementation on it.

man/meson.build
```diff
@@ -235,6 +235,6 @@ update_man_rules = custom_target(
         output : 'update-man-rules',
         command : [sh, '-c',
                    'cd @0@ && '.format(project_build_root) +
-                   'python3 @0@/tools/update-man-rules.py $(find @0@ -wholename "*/man/*.xml") >t && '.format(project_source_root) +
+                   python.path() + ' @0@/tools/update-man-rules.py $(find @0@ -wholename "*/man/*.xml") >t && '.format(project_source_root) +
                    'mv t @0@/rules/meson.build'.format(meson.current_source_dir())],
```
eli-schwartz: This is incorrect. It embeds the resolved python command into a complicated shell script, but does so incorrectly because that interpolation needs to be shell-quoted.

Admittedly, project_build_root and project_source_root are not shell-quoted and are both more likely to be broken than the path to python. After all, the path to python is generally in /usr/bin while the path to the source or build tree may trivially have whitespace in it.

Traditionally, one would pass all 3 strings as argv to the sh command (remembering to fill in a dummy arg for argv0), then use `"$1"` etc.

This highlights the fact that really, the entire command here looks quite strange. Why can't tools/update-man-rules.py be rewritten to find and iterate over xml files itself, and write the `meson.build` file directly once complete?

eli-schwartz (changes requested):

man/meson.build
keszybz: Two advantages: this version will log the full list of files in the build logs, and it is also fairly easy to call update-man-rules with some subset of files for debugging/testing. We *could* rewrite it to make update-man-rules.py look for files itself, but as @poettering said, there is still no rationale what this patch is trying to fix.

heirecka:
> lacks rationale

It allows to specify the desired python executable (and version) via meson's native file if there are multiple versions available.

meson.build
heirecka: The documentation doesn't mention this and to be honest I haven't looked at the source, but if that's true it doesn't respect the configured python3 from a native file.

eli-schwartz: Okay, double checking it seems I am actually wrong.

The behavior I describe should be the way custom_target works, but not run_command. Instead, run_command turns the string internally into an `ExternalProgram(string)` and just checks if that is usable (found).

I think this probably counts as a Meson bug, we "should* respect machine file overrides.

man/meson.build
eli-schwartz: Well, there is the broken terribad shell quoting.

If you just wanted to log the files, you could have it accept the filesnames as argv when given an interactive `--files` argument or something, and otherwise do globbing inside python while using `print()` to perform logging.

I'm a bit confused about this logging thing actually, since it isn't being logged in the command line contained in build.ninja, so what is supposed to log the filenames? sh itself in xtrace mode? The script using `print()`?

One bonus if you simplify it to a single python script which does work internally, is you can just find_program(scriptname) and run that as is. Meson will find the correct interpreter for you, and if needed it will use its own sys.executable.

keszybz: OK. I'll change the program to glob internally then.

bluca:
sounds reasonable in principle - but there was some mistake in the rebase I think? There's a load of external commits, please rebase correctly on latest main and force push again

heirecka:
> sounds reasonable in principle - but there was some mistake in the rebase I think? There's a load of external commits, please rebase correctly on latest main and force push again

Rebased on current main and force-pushed.

man/meson.build
keszybz: https://github.com/systemd/systemd/pull/22829

keszybz:
We have a bazillion helpers which do have `#!/usr/bin/env python3`. How would this work for them?

eli-schwartz:
> We have a bazillion helpers which do have `#!/usr/bin/env python3`. How would this work for them?

If the script is executable, Meson will simply try to run it as is (it's an opaque program), and the shebang must somehow work.

If the script is not executable, Meson will detect the shebang and invoke `shebang-command path/to/script`. It then special cases the shebang-command for both `/usr/bin/env python3` and `/any/path/ending/in/python3` -- in both cases, it runs using Meson's own `sys.executable` (which is definitely a working version of some version of python3 >= 3.7).

yuwata:
Please rebase again.

heirecka:
Rebased.

keszybz:
Looks reasonable.

keszybz:
> > We have a bazillion helpers which do have `#!/usr/bin/env python3`. How would this work for them?
> 
> If the script is executable, Meson will simply try to run it as is (it's an opaque program), and the shebang must somehow work.
> 
> If the script is not executable, Meson will detect the shebang and invoke `shebang-command path/to/script`. It then special cases the shebang-command for both `/usr/bin/env python3` and `/any/path/ending/in/python3` -- in both cases, it runs using Meson's own `sys.executable` (which is definitely a working version of some version of python3 >= 3.7).

Our python scripts are all executable. Should we make them `-x` to trigger this magic behaviour for python3? I think we'd prefer to use the version of python that meson uses.

PR #22814, opened 2022-03-22
random-util: use correct minimum pool size constant

zx2c4:
```
The actual minimum size of the pool across supported kernel versions is
32 bytes. So adjust this minimum.

I've audited every single usage of random_pool_size(), and cannot see
anywhere that this would have any impact at all on anything. We could
actually just not change the constant and everything would be fine, or
we could change it here and that's fine too. From both a functionality
and crypto perspective, it doesn't really seem to make a substantive
difference any which way, so long as the value is ≥32. However, it's
better to be correct and have the function do what it says, so clamp it
to the right minimum.
```


zx2c4:
@mrc0mmand Seems like CentOS flaked on something?

PR #22820, opened 2022-03-22
docs: add some docs about building OS images

poettering:
It's not trivial to know what to reset how. Let's document this a bit.

### DIFF
--- a/docs/BUILDING_IMAGES.md
+++ b/docs/BUILDING_IMAGES.md
@@ -1,0 +1,227 @@
+---
+title: Safely Building Images
+category: Concepts
+layout: default
+SPDX-License-Identifier: LGPL-2.1-or-later
+---
+
+# Safely Building Images
+
+In many scenarios OS installations are shipped as pre-built images, that
+require no further installation process beyond simple `dd`-ing the image to
+disk and booting it up. When building such "golden" OS images for
+`systemd`-based OSes a few points should be taken into account.
+
+Most of the points described here are implemented by the
+[`mkosi`](https://github.com/systemd/mkosi) OS image builder developed and
+maintained by the systemd project. If you are using or working on another image
+builder it's recommended to keep the following concepts and recommendations in
+mind.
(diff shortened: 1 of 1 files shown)


docs/BUILDING_IMAGES.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Resources to Reset
+
+Typically the same OS image shall be deployable in multiple instances, and each
+instance should automatically acquire its own identifying credentials on first
+boot. For that it's essential to:
+
+1. Remove the
+   [`/etc/machine-id`](https://www.freedesktop.org/software/systemd/man/machine-id.html)
+   file or write the string `uninitialized\n` into it. This file is supposed to
```
bluca, line 29: Might be worth noting that when /etc is read-only, you instead want the file to be empty, so that it's mounted over. It's a non-trivial detail, so it might be helpful to note it here.

bluca:
It would be good to link to this from other docs, like the partition one?

keszybz:
Let's treat this similarly to NEWS, i.e. let's not make comments about spelling and enhancements, because that'll just lead to duplicated work. If you see something, push something ;)

docs/BUILDING_IMAGES.md
poettering: So this whole text doesn't cover fully immutable/stateless images right now. I think this should be discussed in another section at the bottom of the doc eventually, but more comprehensively. For now, the text focusses on images that are mutable (specifically: at least /etc/ + /var/ being mutable).

I mean, we do link to the machine-id man page which details the behaviour regarding immutable images already, so there *is* some documentation about this already, except that maybe it's not as easy to get an overview of as one wished.

Hence, yes, we should add more docs about how to do images with immutable /etc. But let's do that comprehensively later, if that makes any sense.

poettering:
> It would be good to link to this from other docs, like the partition one?

done now.

PR #22821, opened 2022-03-22
Udev tweaklets

poettering:


yuwata (approved):
LGTM. Thanks.

bluca (approved):

bluca:
> Wow, two pull requests with two bit-for-bit identical commits. That doesn't happen often.

I guess they fixed something in the Matrix!

poettering:
but mine was a split second faster ;-) hehe

PR #22822, opened 2022-03-22
sd-device-enumerator: two minor fixlets

yuwata:
Addresses comments in #22662.


yuwata:
Haha, #22821 is faster. Closing.

PR #22823, opened 2022-03-22
fs-util: openat_report_new() follow-up

poettering:
@fbuihuu noticed that I fucked up the openat_report_new() return parameters: there are success paths where the return params are not initialized. Fix that.


bluca:
> There's still an issue: in the `for(;;)` loop, you assume that `ret_newly_created` is always passed whereas you decided to make it optional.

That's fine though, as there's a check at the beginning - it cannot get into that loop at all if `ret_newly_created` is not passed

bluca:
> Indeed but then there's no need to check whether `ret_newly_created` is null in the previous `if {}` block.

Yes, that can be removed

poettering:
aarg, apparently programming is hard ;-)

poettering:
dropped the redundant if check, ptal

bluca (approved):

Issue #22763, opened 2022-03-16
Setting DuplicateAddressDetection=none doesn't disable DAD for link-local IPs


slyon:
**systemd version the issue has been seen with**
 > 249.5

**Used distribution**
 > Ubuntu Jammy 22.04

**Linux kernel version used** (`uname -a`)

 > 5.15

**CPU architecture issue was seen on**
 > x86

**Expected behaviour you didn't see**
 > Setting `DuplicateAddressDetection=none` in a .network file does not disable DAD for link-local IPs. I think this commit https://github.com/systemd/systemd/commit/1cf4ed142d6c1e2b9dc6a0bc74b6a83ae30b0f8e may be related (landed in systemd v249). I've also tested the out-of-the-box systemd versions for Focal and Impish but there I can effectively disable DAD. If the IPs are set to non-link local addresses then the problem is not observed anymore.
> 
> I think at least this new behavior should be documented in https://systemd.network/systemd.network.html#DuplicateAddressDetection=

**Unexpected behaviour you saw**
 > I've noticed that setting DuplicateAddressDetection=none for an interface with a link-local address (e.g., 169.254.*) via a .network file added to /etc/systemd/network/ doesn't really disable Duplicate Address Detection.

**Steps to reproduce the problem**
> 1- Set up Ubuntu 22.04 VM
> 2- Increase systemlog level:
> 
>   mkdir -p /etc/systemd/system/systemd-networkd.service.d/
>   cat > /etc/systemd/system/systemd-networkd.service.d/10-debug.conf <<EOF
>   [Service]
>   Environment=SYSTEMD_LOG_LEVEL=debug
>   EOF
>   systemctl daemon-reload && systemctl restart systemd-networkd
> 
> 3- Configure a link-local address to a network interface on the VM:
> vi /etc/systemd/network/10-netplan-enp7s0.network
> 
> [Match]
> Name=enp7s0
> 
> [Address]
> Address=169.254.240.10/24
> DuplicateAddressDetection=ipv4
> 
> 4- Restart services and apply
> 
> systemctl daemon-reload && systemctl restart systemd-networkd && systemctl restart networkd-dispatcher.service
> netplan apply
> 
> 5- Check if DAD is running (it should):
> 
> journalctl -b -u systemd-networkd --no-pager -o short-precise | grep -i ACD
> 
> 6- Set DuplicateAddressDetection=none in
> /etc/systemd/network/10-netplan-enp7s0.network
> 
[...]

PR #22824, opened 2022-03-22
network: do not enable IPv4 ACD for IPv4 link-local address if ACD is disabled explicitly

yuwata:
The commit https://github.com/yuwata/systemd/commit/1cf4ed142d6c1e2b9dc6a0bc74b6a83ae30b0f8e makes the IPv4 ACD
enabled unconditionally for IPv4 link-local addresses even if users
explicitly disable ACD.

This makes the IPv4 ACD is enabled by default, but honor user setting.

Fixes https://github.com/systemd/systemd/issues/22763.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -1045,7 +1045,8 @@ Table=1234</programlisting></para>
           Detection. See <ulink url="https://tools.ietf.org/html/rfc5227">RFC 5227</ulink>.
           When <literal>ipv6</literal>, performs IPv6 Duplicate Address Detection. See
           <ulink url="https://tools.ietf.org/html/rfc4862">RFC 4862</ulink>. Defaults to
-          <literal>ipv6</literal>.</para>
+          <literal>ipv4</literal> for IPv4 link-local addresses, <literal>ipv6</literal> for IPv6
+          addresses, and <literal>none</literal> otherwise.</para>
         </listitem>
       </varlistentry>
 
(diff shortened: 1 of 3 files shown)


poettering:
hmm, so far we tried to avoid boolean options that are prefixed with "Enable", as that is pretty much redundant wording. it's boolean after all...

yuwata:
> hmm, so far we tried to avoid boolean options that are prefixed with "Enable", as that is pretty much redundant wording. it's boolean after all...

OK. The renaming part was dropped. PTAL.

yuwata:
The issue is not critical, but the offending commit changed previous behavior. Hence, please backport this to v249 and v250.

keszybz:
I think we can simplify this further. I took the liberty to force-push into your branch, because I thought that just a simple cleanup can be done and I started working on a patch for that. In the end I realized that we can simplify the logic. Please drop my patch if you think it's wrong.

yuwata:
@keszybz No no, you confused the current unfortunate situation: `DuplicateAddressDetection=yes` was equivalent to `DuplicateAddressDetection=none`. That's soooo unfortunate.... For backward compatibility, we cannot safely use boolean string for the setting. Ugh!

I'd like to adopt something like you posted to clean up the current situation. But please drop the second commit, as it does not work as expected.

To make it clear, the very unfortunate code is the following:
```
n->duplicate_address_detection = r ? ADDRESS_FAMILY_NO : ADDRESS_FAMILY_YES;
```

yuwata:
(The second commit was dropped. PTAL)

keszybz:
This does not apply cleanly to v250-stable. Please provide a pull request.

keszybz:
@yuwata ^

yuwata:
Oops. Will do.

yuwata:
v250 backport: https://github.com/systemd/systemd-stable/pull/209
v249 backport: https://github.com/systemd/systemd-stable/pull/210

PR #22825, opened 2022-03-22
Assorted cleanups

keszybz:


bluca (approved):

src/basic/errno-to-name.awk
```diff
@@ -3,7 +3,7 @@
 BEGIN{
```
dnicolodi, line 3: `s/errno-to-awk/errno-to-name/` in the commit message :)

keszybz: Oops. But I don't think it's worth pushing and retriggering the CI for this.

keszybz:
focal-ppc64el: TEST-50-DISSECT:                    FAIL     (611 s), hopefully unrelated.

PR #22831, opened 2022-03-22
xdg-autostart-service: Re-escape desktop Exec line

davidedmundson:
A valid desktop entry may be in the form
Exec="/bin/path with spaces/" someArgs

The generator correctly expands this treating the quoted string as a
single executable name. However this is then lost when writing back
into a systemd service. The executable is written without quotes or
the space escaped into the ExecStart line.

This patch uses an escape function that will re-insert quotes if needed.

### DIFF
--- a/src/xdg-autostart-generator/xdg-autostart-service.c
+++ b/src/xdg-autostart-generator/xdg-autostart-service.c
@@ -412,7 +412,7 @@ int xdg_autostart_format_exec_start(
                         if (r < 0)
                                 return log_info_errno(r, "Exec binary '%s' does not exist: %m", c);
 
-                        escaped = cescape(executable);
+                        escaped = shell_maybe_quote(executable, 0);
                         if (!escaped)
                                 return log_oom();
 


davidedmundson:
@benzea

benzea:
Outch. Looks good to me.

Further down there is this:
```
                escaped = cescape(p);
                if (!escaped)
                        return log_oom();
                quoted = strjoin("\"", escaped, "\"");
                if (!quoted)
                        return log_oom();
```

Can we also simply use `shell_maybe_quote` there?

davidedmundson:
>Can we also simply use shell_maybe_quote there?

At that point, we could removal all escaping within the loop and replace the str_join with `quote_command_line(char**)`.

It works fine for most cases, but I'm not 100% sure it's right. It breaks the unit tests, some is obviously safe losing the quotes, but I also see:

```
In: '/bin/sleep ";\""', out: '/bin/sleep ";\\\""', expected: '/bin/sleep ";\""'
```

benzea:
Hmm, yeah, that looks wrong, however, the following seems to work here:

```
diff --git a/src/xdg-autostart-generator/test-xdg-autostart.c b/src/xdg-autostart-generator/test-xdg-autostart.c
index c7a816bc25..5c46d116e4 100644
--- a/src/xdg-autostart-generator/test-xdg-autostart.c
+++ b/src/xdg-autostart-generator/test-xdg-autostart.c
@@ -25,13 +25,13 @@ static void test_xdg_format_exec_start_one(const char *exec, const char *expecte
 }
 
 static void test_xdg_format_exec_start(void) {
-        test_xdg_format_exec_start_one("/bin/sleep 100", "/bin/sleep \"100\"");
+        test_xdg_format_exec_start_one("/bin/sleep 100", "/bin/sleep 100");
 
         /* All standardised % identifiers are stripped. */
         test_xdg_format_exec_start_one("/bin/sleep %f \"%F\" %u %U %d %D\t%n %N %i %c %k %v %m", "/bin/sleep");
 
         /* Unknown % identifier currently remain, but are escaped. */
-        test_xdg_format_exec_start_one("/bin/sleep %X \"%Y\"", "/bin/sleep \"%%X\" \"%%Y\"");
+        test_xdg_format_exec_start_one("/bin/sleep %X \"%Y\"", "/bin/sleep %%X %%Y");
 
         test_xdg_format_exec_start_one("/bin/sleep \";\\\"\"", "/bin/sleep \";\\\"\"");
 }
diff --git a/src/xdg-autostart-generator/xdg-autostart-service.c b/src/xdg-autostart-generator/xdg-autostart-service.c
index d450341b9f..fb0efdbf4a 100644
--- a/src/xdg-autostart-generator/xdg-autostart-service.c
+++ b/src/xdg-autostart-generator/xdg-autostart-service.c
@@ -396,7 +396,7 @@ int xdg_autostart_format_exec_start(
 
         first_arg = true;
         for (i = n = 0; exec_split[i]; i++) {
-                _cleanup_free_ char *c = NULL, *raw = NULL, *p = NULL, *escaped = NULL, *quoted = NULL;
+                _cleanup_free_ char *c = NULL, *raw = NULL, *percent = NULL;
                 ssize_t l;
 
                 l = cunescape(exec_split[i], 0, &c);
@@ -412,11 +412,7 @@ int xdg_autostart_format_exec_start(
                         if (r < 0)
                                 return log_info_errno(r, "Exec binary '%s' does not exist: %m", c);
 
-                        escaped = cescape(executable);
-                        if (!escaped)
-                                return log_oom();
-
-                        free_and_replace(exec_split[n++], escaped);
+                        free_and_replace(exec_split[n++], executable);
                         continue;
                 }
 
@@ -445,23 +441,16 @@ int xdg_autostart_format_exec_start(
                 raw = strreplace(c, "%%", "%");
                 if (!raw)
                         return log_oom();
-                p = strreplace(raw, "%", "%%");
-                if (!p)
-                escaped = cescape(p);
-                if (!escaped)
-
-                quoted = strjoin("\"", escaped, "\"");
-                if (!quoted)
+                percent = strreplace(raw, "%", "%%");
+                if (!percent)
                         return log_oom();
 
-                free_and_replace(exec_split[n++], quoted);
+                free_and_replace(exec_split[n++], percent);
         }
         for (; exec_split[n]; n++)
                 exec_split[n] = mfree(exec_split[n]);
 
-        res = strv_join(exec_split, " ");
+        res = quote_command_line(exec_split, SHELL_ESCAPE_EMPTY);
         if (!res)
                 return log_oom();
 
```

keszybz:
@benzea can you submit this as PR? It looks reasonable.

PR #22833, opened 2022-03-23
calendarspec: fix possibly skips next elapse

gibeom-gwon:
If the time unit changes after adding the repetition value, the
timer may skip the next elapse. This patch reset sub time units
to minimum value when upper unit is changed.

Example cases:

```
$ systemd-analyze --base-time '2022-02-28 01:00:00' calendar '*-*-01/5 04:00:00'
Normalized form: *-*-01/5 04:00:00
    Next elapse: Sun 2022-03-06 04:00:00 KST << should be Sun 2022-03-01 04:00:00 KST
       (in UTC): Sat 2022-03-05 19:00:00 UTC
       From now: 2 weeks 3 days ago
```

```
$ systemd-analyze --base-time '2022-04-30 01:00:00' calendar '*-*-01/7 04:00:00'
Normalized form: *-*-01/7 04:00:00
    Next elapse: Sun 2022-05-08 04:00:00 KST << should be Sun 2022-05-01 04:00:00 KST
       (in UTC): Sat 2022-05-07 19:00:00 UTC
       From now: 1 month 15 days left
```


yuwata:
Could you add several testcases for the issue?

cc @keszybz

src/shared/calendarspec.c
```diff
@@ -1196,12 +1197,19 @@ static int tm_within_bounds(struct tm *tm, bool utc) {
...
 
         /* Did any normalization take place? If so, it was out of bounds before */
-        int cmp = CMP(t.tm_year, tm->tm_year) ?:
-                  CMP(t.tm_mon, tm->tm_mon) ?:
-                  CMP(t.tm_mday, tm->tm_mday) ?:
-                  CMP(t.tm_hour, tm->tm_hour) ?:
-                  CMP(t.tm_min, tm->tm_min) ?:
-                  CMP(t.tm_sec, tm->tm_sec);
+        if ((cmp = CMP(t.tm_year, tm->tm_year)) != 0)
+                t.tm_mon = 0;
```
poettering, line 1207: shouldn't this be 1?

src/shared/calendarspec.c
```diff
@@ -1196,12 +1197,19 @@ static int tm_within_bounds(struct tm *tm, bool utc) {
...
+                t.tm_mday = 1;
+        else if ((cmp = CMP(t.tm_mday, tm->tm_mday)) != 0)
+                t.tm_hour = 0;
+        else if ((cmp = CMP(t.tm_hour, tm->tm_hour)) != 0)
+                t.tm_min = 0;
+        else if ((cmp = CMP(t.tm_min, tm->tm_min)) != 0)
+                t.tm_sec = 0;
+        else if ((cmp = CMP(t.tm_sec, tm->tm_sec)) != 0) {
+                // no op
+        }
```
poettering: drop the {} if there is no op to execute, i.e. end in:

```c
…
else if ((cmp = CMP(t.tm_min, tm->tm_min)) != 0)
        t.tm_sec = 0;
else
        cmp = CMP(t.tm_sec, tm->tm_sec);
```

poettering: (also as per coding style https://systemd.io/CODING_STYLE/#formatting we use `//` style comments for FIXME stuff, and `/**/` comments for explanatory stuff. the no op comment is the latter... but I'd drop it entirely)

src/shared/calendarspec.c
```diff
@@ -1196,12 +1197,19 @@ static int tm_within_bounds(struct tm *tm, bool utc) {
                 return negative_errno();
 
         /* Did any normalization take place? If so, it was out of bounds before */
```
poettering: somewhere here it would be good to add a comment, explaining why we reset those fields if we found a difference

src/shared/calendarspec.c
gibeom-gwon: mktime(3) says valid tm_mon is 0-11, so I think `tm_mon` should be 0. https://github.com/systemd/systemd/blob/0be1504704caff1c48618d100c5d4ae0b4198485/src/shared/calendarspec.c#L1272-L1274
It looks like systemd internally uses +1 biased month, but restored after use.

src/shared/calendarspec.c
gibeom-gwon: Fixed.

src/shared/calendarspec.c
gibeom-gwon: Fixed.

gibeom-gwon:
> Could you add several testcases for the issue?
> 
> cc @keszybz

Added 2 test cases.

src/shared/calendarspec.c
```diff
@@ -1195,13 +1196,25 @@ static int tm_within_bounds(struct tm *tm, bool utc) {
...
+        else if ((cmp = CMP(t.tm_mon, tm->tm_mon)) != 0)
+                t.tm_mday = 1;
+        else if ((cmp = CMP(t.tm_mday, tm->tm_mday)) != 0)
+                t.tm_hour = 0;
+        else if ((cmp = CMP(t.tm_hour, tm->tm_hour)) != 0)
+                t.tm_min = 0;
+        else if ((cmp = CMP(t.tm_min, tm->tm_min)) != 0)
+                t.tm_sec = 0;
+        else
+                cmp = CMP(t.tm_sec, tm->tm_sec);
```
yuwata, line 1217: Hmm, I am not familiar with the code, but I am confused why we need to do like that here.
About year or month, we have similar code to handle that in `find_next()` for day or other units. However, the handling for other units looks slightly different. That is, the sequence of the operation `+1`, `find_matching_component()`, and `-1` is done only for year and month, but not for other units...

My question and suggestion is, should we do the similar way for other units in `find_next()`, and keep `tm_within_bounds()` as is???

yuwata: But, let's wait for @keszybz's review. IIRC, when @keszybz edited that part previously, I was also confused with the code...

PR #22837, opened 2022-03-23
Revert "lgtm: disable cpp/missing-return"

mrc0mmand:
This reverts commit 6f4bffb586dfb0ce8db4e02ccb7f076a45bca419.

Should be, hopefully, fixed by https://github.com/github/codeql/issues/8409.

---

The fix might not yet be available in LGTM/CodeQL production, so let's keep this one open as a reminder.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -4,8 +4,6 @@
 
 # Explicitly enable certain checks which are hidden by default
 queries:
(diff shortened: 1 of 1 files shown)


mrc0mmand:
Looks like LGTM is not complaining anymore, so this should be ready to go.

Issue #11221, opened 2018-12-20
bootctl/systemd-boot api


jameshilliard:
**Is your feature request related to a problem? Please describe.**
There doesn't currently seem to be a good way for applications such as [swupdate](https://sbabic.github.io/swupdate/) to manage systemd-boot.

**Describe the solution you'd like**
There should be a c header api/library that allows external apps to control installation of systemd-boot in the same way bootctl does. 

For example one could create a swupdate systemd-boot environment [handler](https://sbabic.github.io/swupdate/handlers.html#supplied-handlers) if bootctl's functionality were exposed externally. 

The API should provide the ability to programmatically add, remove and select systemd-boot entries. The API should ideally expose a way to generate and parse the systemd-boot conf files. 

**Describe alternatives you've considered**
Alternatives would be re-implementing bootctl's functionality within swupdate or wrapping bootctl in a script.

I see there are some functions exposed in [bootspec](https://github.com/systemd/systemd/blob/master/src/shared/bootspec.h) and [efivars](https://github.com/systemd/systemd/blob/master/src/shared/efivars.h) although they don't seem to be capable of handling systemd-boot installation or manipulation of the conf files.

mhill-holoplot:
I would also be looking for such an interface to integrate sd-boot with rauc.

poettering:
We do have a Varlink IPC API now for bootctl, but it is very limited, i.e. is mostly about introspection, but not about changing things. But it's a start, and it's a great template for exposing the full functionality.

Original commit was 79ec39958d70e3eeb141f7ca1f57ea52533727b6 but received numerous updates since.

PR #22840, opened 2022-03-23
util-lib: efi variable access refactorings

poettering:
Nothing major, just some refactorings, splitting up of .c/.h files and such.

### DIFF
--- a/src/shared/efi-loader.c
+++ b/src/shared/efi-loader.c
@@ -1,537 +1,22 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
-#include <stdlib.h>
-#include <sys/stat.h>
-#include <unistd.h>
-
 #include "alloc-util.h"
-#include "dirent-util.h"
 #include "efi-loader.h"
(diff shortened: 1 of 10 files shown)


src/shared/efi-loader.c
```diff
@@ -1,537 +1,22 @@
...
-
-        typesafe_qsort(list, count, cmp_uint16);
-
-        *options = TAKE_PTR(list);
-
-        return count;
-}
-
-static int read_usec(const char *variable, usec_t *u) {
+static int read_usec(const char *variable, usec_t *ret) {
```
yuwata, line 13: I think it is better to such renaming of variables or arguments should be in the previous commit.

That is, the commit splitting a file into two or more should only do that for reviewers.

yuwata (approved):
LGTM. But one comments.

jameshilliard:
@poettering Think this gets us closer to having a bootctl API(#11221)?

yuwata:
Hm?? mkosi tests are now broken with this?

PR #22841, opened 2022-03-23
man: allow "+" in os-release VERSION_ID and IMAGE_VERSION

bettio:
Semantic versioning makes use of plus sign for denoting build metadata,
such as 1.0.0+20130313144700 or 1.0.0+21AF26D3.
Previous to this change "+" wasn't allowed, making impossible using
semver in os-release fields.

This change is still not enough for enabling full semver support,
however it covers the complete semver semantic, and also uppercase
letters are not really common in real world scenarios.

**Following this change os-release parsers must start allowing "+" in
VERSION_ID and IMAGE_VERSION in order to be compliant.**

**To the reviewers**: *I didn't add `A-Z` range since it would be an addition
that doesn't enable a significant number of use-cases, but it would be quite
an extensive addition to VERSION_ID and IMAGE_VERSION supported
chars range.
Let me know if it might be of interest having full semver support.*

### DIFF
--- a/man/os-release.xml
+++ b/man/os-release.xml
@@ -242,7 +242,7 @@
           <term><varname>VERSION_ID=</varname></term>
 
           <listitem><para>A lower-case string (mostly numeric, no spaces or other characters outside of 0–9,
-          a–z, ".", "_" and "-") identifying the operating system version, excluding any OS name information
+          a–z, ".", "_", "-" and "+") identifying the operating system version, excluding any OS name information
           or release code name, and suitable for processing by scripts or usage in generated filenames. This
           field is optional.</para>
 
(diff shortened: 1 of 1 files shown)


bluca:
> Let me know if it might be of interest having full semver support.

I think we should yes, and that should also include the `~` character I think?

medhefgo:
This nicely clashes with systemd-boot's boot counting scheme (and version info is usually used in the entry filename). :D

/cc @poettering

keszybz:
I think we don't want A-Z, because we have to deal with FAT filesystems, and case might get mangled, and names that differ only in case might be treated as identical or not identical depending on who's accessing the systemd and the phase of the moon.

Adding `+` sound reasonable. Semver is **the** versioning scheme, and we should be compatible with as much as possible.

We could change the separator for boot counting, e.g. to `:`.

Issue #22580, opened 2022-02-21
sd-boot: Incorrect default boot entry info from bootctl


JiriPodhorny:
**systemd version the issue has been seen with**
 > 250.3-4

**Used distribution**
 > ArchLinux

**Linux kernel version used** (`uname -a`)

 > `5.16.10-arch1-1 #1 SMP PREEMPT Wed, 16 Feb 2022 19:35:18 +0000 x86_64`

**CPU architecture issue was seen on**
 > x86_64

**Expected behaviour you didn't see**
 > Default boot entry should be marked with `(default)` in output from commands `bootctl`, `bootctl status` and `bootctl list`.

**Unexpected behaviour you saw**
 > If some of `auto-*` options set as default, then alphabetically last non-auto option defined by `*.conf` file is set as default in output from above mentioned commands (in my case `memtest94`. 

**Steps to reproduce the problem**
 > On my work notebook I need boot to MS Windows by default, so my `loader.conf` looks like:
```
timeout 1
console-mode keep
default auto-windows
```
During boot is option `Windows 10` correctly selected and notebook boots to Windows (that's OK).
When I select and boot to `ArchLinux` and want to check boot options, above mentioned comands returns incorrect results.

**Additional program output to the terminal or log subsystem illustrating the issue**

```
$ bootctl list
Boot Loader Entries:
        title: Arch Linux (manual entry)
           id: arch1.conf
       source: /boot/loader/entries/arch1.conf
      version: manual entry
        linux: /vmlinuz-linux
       initrd: /amd-ucode.img
               /initramfs-linux.img
      options: root=LABEL=ARCHFS notsc rw ipv6.disable=1 8250.nr_uarts=2 fsck.mode=force fsck.repair=yes

        title: Arch Linux (rolling) 
           id: archlinux.efi
       source: /boot/EFI/Linux/archlinux.efi
      version: rolling
        linux: EFI/Linux/archlinux.efi
      options: root=LABEL=ARCHFS notsc rw ipv6.disable=1 8250.nr_uarts=1

        title: PassMark MemTest86 v9.4 (default)
           id: memtest94.conf
       source: /boot/loader/entries/memtest94.conf

        title: Windows Boot Manager
           id: auto-windows
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f

        title: EFI Shell
           id: auto-efi-shell
[...]

PR #18387, opened 2021-01-27
JSON output support for `bootctl list`

Batcastle:
This was a pain to figure out. I could not find ANY documentation on how to use `json.h` or `format-table.h`, so I ended up doing this manually. I figure in most cases this should work just fine.

This PR was made to resolve issue https://github.com/systemd/systemd/issues/18094.

Issue #18094, opened 2020-12-28
RFE: JSON output option for bootctl


Batcastle:
**Is your feature request related to a problem? Please describe.**
Currently, it takes seemingly unnecessary parsing to get the output of `bootctl list` in a machine-readable format.

**Describe the solution you'd like**
Having a JSON option, or proper YAML formatting (The current output is already very close to YAML formatting, but not close enough to be parse correctly.), would make it very easy for any third-party developer to use a standard lib to parse the output, in order to provide information to the user about their system, or inform a decision on what to do in a program.

**Describe alternatives you've considered**
Manually fixing the current output in each application is a viable option. If we go for YAML, we just need to change the number of leading spaces on each line so that each line's description lines up, instead of the value for each line lining up. Like so:

Instead of this:
```
Boot Loader Entries:
        title: Reboot Into Firmware Interface
           id: auto-reboot-to-firmware-setup
       source: /sys/firmware/efi/efivars/LoaderEntries-****
```
We would have this:
```yaml
Boot Loader Entries:
       title: "Reboot Into Firmware Interface"
       id: "auto-reboot-to-firmware-setup"
       source: "/sys/firmware/efi/efivars/LoaderEntries-****"
```
As you can see, the only real difference is line indentation. This, however, requires unnecessary preprocessing that can can be avoided by formatting the output as YAML in the first place.
<br/><br/>
YAML can be achieved without even adding a new flag, and it will maintain the same level of readability for users. JSON would likely require a new flag due to it's reduced readability for average users, despite being (arguably) easier to parse and generate. An example of the above `bootctl list` output, in JSON output:

```json
{
    "Boot Loader Entries": [
        {
            "title": "Reboot Into Firmware Interface",
            "id": "auto-reboot-to-firmware-setup",
            "source": "/sys/firmware/efi/efivars/LoaderEntries-****",
            "default": false
        }
    ]
}
```

Is something like this possible? Thanks!!!

poettering:
yaml is such a broken language. json is fine though. we can add that.

Batcastle:
I can get behind that. I preferred JSON anyways. I just looked mostly at YAML cause I figured it would be the least work to add with the current output format.

Batcastle:
Hey. Just wondering where this feature is at? 

I understand it's probably low on the priority list. Just curious if it is already being worked on, just isn't committed yet, or if doing it myself (if I can figure it out XD) and making a PR would be a quicker way of getting it included in the next release.

Thanks!

[...]

PR #22843, opened 2022-03-23
bootctl: bootspec improvements and clean-ups

poettering:
Fixes: #22580 #18094
Replaces: #18387

Various othe refactorings and improvements.

### DIFF
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -709,7 +709,7 @@ static int status_entries(
                        SD_ID128_FORMAT_VAL(dollar_boot_partition_uuid));
         printf("\n\n");
 
-        r = boot_entries_load_config(esp_path, xbootldr_path, &config);
+        r = boot_config_load(&config, esp_path, xbootldr_path);
         if (r < 0)
                 return r;
 
@@ -1816,7 +1816,7 @@ static int verb_list(int argc, char *argv[], void *userdata) {
         /* If XBOOTLDR and ESP actually refer to the same block device, suppress XBOOTLDR, since it would find the same entries twice */
         bool same = arg_esp_path && arg_xbootldr_path && devid_set_and_equal(esp_devid, xbootldr_devid);
 
-        r = boot_entries_load_config(arg_esp_path, same ? NULL : arg_xbootldr_path, &config);
+        r = boot_config_load(&config, arg_esp_path, same ? NULL : arg_xbootldr_path);
         if (r < 0)
                 return r;
 
@@ -1826,7 +1826,7 @@ static int verb_list(int argc, char *argv[], void *userdata) {
         else if (r < 0)
                 log_warning_errno(r, "Failed to determine entries reported by boot loader, ignoring: %m");
         else
-                (void) boot_entries_augment_from_loader(&config, efi_entries, /* only_auto= */ false);
+                (void) boot_config_augment_from_loader(&config, efi_entries, /* only_auto= */ false);
 
         r = boot_config_select_special_entries(&config);
         if (r < 0)
--- a/src/shared/bootspec.c
+++ b/src/shared/bootspec.c
@@ -834,10 +834,10 @@ int boot_config_select_special_entries(BootConfig *config) {
         return 0;
 }
 
-int boot_entries_load_config(
+int boot_config_load(
+                BootConfig *config,
                 const char *esp_path,
-                const char *xbootldr_path,
-                BootConfig *config) {
+                const char *xbootldr_path) {
 
         const char *p;
         int r;
@@ -882,10 +882,10 @@ int boot_entries_load_config(
         return 0;
 }
 
-int boot_entries_load_config_auto(
+int boot_config_load_auto(
+                BootConfig *config,
                 const char *override_esp_path,
-                const char *override_xbootldr_path,
-                BootConfig *config) {
+                const char *override_xbootldr_path) {
 
         _cleanup_free_ char *esp_where = NULL, *xbootldr_where = NULL;
         dev_t esp_devid = 0, xbootldr_devid = 0;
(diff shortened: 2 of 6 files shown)


src/boot/bootctl.c
```diff
@@ -1811,7 +1823,44 @@ static int verb_list(int argc, char *argv[], void *userdata) {
...
+                                                       JSON_BUILD_PAIR_CONDITION(e->machine_id, "machineId", JSON_BUILD_STRING(e->machine_id)),
+                                                       JSON_BUILD_PAIR_CONDITION(e->architecture, "architecture", JSON_BUILD_STRING(e->architecture)),
+                                                       JSON_BUILD_PAIR_CONDITION(opts, "options", JSON_BUILD_STRING(opts)),
+                                                       JSON_BUILD_PAIR_CONDITION(e->kernel, "linux", JSON_BUILD_STRING(e->kernel)),
+                                                       JSON_BUILD_PAIR_CONDITION(e->efi, "efi", JSON_BUILD_STRING(e->efi)),
+                                                       JSON_BUILD_PAIR_CONDITION(!strv_isempty(e->initrd), "initrd", JSON_BUILD_STRV(e->initrd)),
+                                                       JSON_BUILD_PAIR_CONDITION(e->device_tree, "deviceTree", JSON_BUILD_STRING(e->device_tree)),
+                                                       JSON_BUILD_PAIR_CONDITION(!strv_isempty(e->device_tree_overlay), "deviceTreeOverlay", JSON_BUILD_STRV(e->device_tree_overlay))));
```
medhefgo: sd-boot only does one device tree (the last one wins). So this being a strv is wrong…

src/shared/bootspec.c
```diff
@@ -275,20 +275,18 @@ static int boot_entry_compare(const BootEntry *a, const BootEntry *b) {
 }
 
 static int boot_entries_find(
+                BootConfig *config,
                 const char *root,
-                const char *dir,
-                BootEntry **entries,
-                size_t *n_entries) {
+                const char *dir) {
```
medhefgo, line 334: assert? (the other ones too)

src/boot/bootctl.c
medhefgo: Also, afaik it doesn't do overlays. So calling it one is also wrong…

keszybz: The field is called `devicetree-overlay`, so this should be "devicetreeOveraly".
Also "devicetree" above. Let's maintain a consistent transliteration from the keys in the spec. Our internal variable names are less important.

I think making this a strv is OK: we may add support for multiple in the future, and then the format won't be broken.

man/bootctl.xml
```diff
@@ -92,7 +92,10 @@
         url="https://systemd.io/BOOT_LOADER_SPECIFICATION">Boot Loader Specification</ulink>, as well as any
         other entries discovered or automatically generated by a boot loader implementing the <ulink
         url="https://systemd.io/BOOT_LOADER_INTERFACE">Boot Loader
-        Interface</ulink>.</para></listitem>
+        Interface</ulink>.</para>
+
+        <para>When combined with <option>--json=</option> outputs the boot loader entries in a series of
+        simple JSON objects.</para></listitem>
```
keszybz: Maybe `JSON output may be requested with <option>--json=</option>.`.

("object" is not something that users care about.)

src/shared/bootspec.c
keszybz: Hmm, what do you mean? I see `assert(config);` added below.

src/shared/bootspec.c
```diff
@@ -274,6 +276,58 @@ static int boot_entry_compare(const BootEntry *a, const BootEntry *b) {
...
+        r = CMP(a->st_dev, b->st_dev);
+        if (r != 0)
+                return r;
+
+        return CMP(a->st_ino, b->st_ino);
+}
+
+DEFINE_HASH_OPS_WITH_KEY_DESTRUCTOR(inode_hash_ops, struct stat, inode_hash_func, inode_compare_func, free);
+
+static int config_check_inode(BootConfig *config, int fd, const char *fname) {
```
keszybz: Maybe call this `config_inode_relevant_and_unseen()`, so it's self-explanatory without the comment in the calller.

src/shared/bootspec.c
```diff
@@ -975,6 +976,15 @@ int boot_entries_augment_from_loader(
...
                 };
+
+                added = true;
+        }
+
+        if (added) {
+                /* If we added a new entry that the loader found, let's reasses which entries are
+                 * default/selected */
+                config->default_entry = boot_entries_select_default(config);
+                config->selected_entry = boot_entries_select_selected(config);
```
keszybz: That seems ugly, because it implies we calculate the default and selected entries twice. Shouldn't we instead postpone the calculation to do it just one at the right time?

Or maybe make `boot_entries_get_{default,selected}` caching getters that calculate the right entry on first access?

keszybz (changes requested):

src/boot/bootctl.c
poettering: well, the spec and the parse know the overlays, so we should show them here. If sd-boot currently supports them is another thing. In theory at least sd-boot is *one* implementation of the boot loader spec, but the spec should be generic.

src/shared/bootspec.c
poettering: yeah, all params are assert()ed on

poettering:
Force pushed a new version, with the raised points addressed. I also added three more minor refactroing commits on top. PTAL

keszybz:
mkosi / ci (centos_epel, 8-stream) (pull_request) failed with some package dependency issue.
focal-ppc64el: TEST-50-DISSECT:                    FAIL     (611 s)

keszybz:
LGTM.

PR #22844, opened 2022-03-23
udev/net_id: avoid slot based names only for single function devices

msekletar:
If we have two or more devices that share the same slot but they are
also multifunction then it is OK to use the slot information even if it
is the same for all of them. Name conflict will be avoided because we
will append function number and form names like, ens1f1, ens1f2...

### DIFF
--- a/man/systemd.net-naming-scheme.xml
+++ b/man/systemd.net-naming-scheme.xml
@@ -403,6 +403,18 @@
           </listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><constant>v251</constant></term>
+
+          <listitem><para>Since version <constant>v247</constant> we no longer set
+          <varname>ID_NET_NAME_SLOT</varname> if we detect that a PCI device associated with a slot is a PCI
+          bridge as that would create naming conflict when there are more child devices on that bridge. Now,
+          this is relaxed and we will use slot information to generate the name based on it but only if
+          the PCI device has multiple functions. This is safe because distinct function number is a part of
+          the device name for multifunction devices.</para>
+          </listitem>
+        </varlistentry>
+
       </variablelist>
 
     <para>Note that <constant>latest</constant> may be used to denote the latest scheme known (to this
--- a/src/shared/netif-naming-scheme.h
+++ b/src/shared/netif-naming-scheme.h
@@ -22,20 +22,21 @@
  * OS versions, but not fully stabilize them. */
 typedef enum NamingSchemeFlags {
         /* First, the individual features */
-        NAMING_SR_IOV_V            = 1 << 0, /* Use "v" suffix for SR-IOV, see 609948c7043a */
-        NAMING_NPAR_ARI            = 1 << 1, /* Use NPAR "ARI", see 6bc04997b6ea */
-        NAMING_INFINIBAND          = 1 << 2, /* Use "ib" prefix for infiniband, see 938d30aa98df */
-        NAMING_ZERO_ACPI_INDEX     = 1 << 3, /* Use zero acpi_index field, see d81186ef4f6a */
-        NAMING_ALLOW_RERENAMES     = 1 << 4, /* Allow re-renaming of devices, see #9006 */
-        NAMING_STABLE_VIRTUAL_MACS = 1 << 5, /* Use device name to generate MAC, see 6d3646406560 */
-        NAMING_NETDEVSIM           = 1 << 6, /* Generate names for netdevsim devices, see eaa9d507d855 */
-        NAMING_LABEL_NOPREFIX      = 1 << 7, /* Don't prepend ID_NET_LABEL_ONBOARD with interface type prefix */
-        NAMING_NSPAWN_LONG_HASH    = 1 << 8, /* Shorten nspawn interfaces by including 24bit hash, instead of simple truncation  */
-        NAMING_BRIDGE_NO_SLOT      = 1 << 9, /* Don't use PCI hotplug slot information if the corresponding device is a PCI bridge */
-        NAMING_SLOT_FUNCTION_ID    = 1 << 10, /* Use function_id if present to identify PCI hotplug slots */
-        NAMING_16BIT_INDEX         = 1 << 11, /* Allow full 16-bit for the onboard index */
-        NAMING_REPLACE_STRICTLY    = 1 << 12, /* Use udev_replace_ifname() for NAME= rule */
-        NAMING_XEN_VIF             = 1 << 13, /* GEnerate names for Xen netfront devices */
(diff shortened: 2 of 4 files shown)


yuwata:
I think it is better to add a new NAMING_ flag for this change.

msekletar:
> I think it is better to add a new NAMING_ flag for this change.

Ok, I will add that.

poettering:
patch looks ok, but please add a new naming scheme flag for that. Also please update the log message a bit according to the new conditionalization.

msekletar:
Updated. @yuwata @poettering PTAL.

man/systemd.net-naming-scheme.xml
```diff
@@ -403,6 +403,18 @@
           </listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><constant>v251</constant></term>
+
+          <listitem><para>Since version <constant>v247</constant> we no longer set
+          <varname>ID_NET_NAME_SLOT</varname> if we detect that a PCI device associated with a slot is a PCI
```
keszybz, line 410: This doesn't match the constant that appears in the code below.

src/shared/netif-naming-scheme.h
```diff
@@ -22,20 +22,21 @@
...
+        NAMING_STABLE_VIRTUAL_MACS      = 1 << 5, /* Use device name to generate MAC, see 6d3646406560 */
+        NAMING_NETDEVSIM                = 1 << 6, /* Generate names for netdevsim devices, see eaa9d507d855 */
+        NAMING_LABEL_NOPREFIX           = 1 << 7, /* Don't prepend ID_NET_LABEL_ONBOARD with interface type prefix */
+        NAMING_NSPAWN_LONG_HASH         = 1 << 8, /* Shorten nspawn interfaces by including 24bit hash, instead of simple truncation  */
+        NAMING_BRIDGE_NO_SLOT           = 1 << 9, /* Don't use PCI hotplug slot information if the corresponding device is a PCI bridge */
+        NAMING_SLOT_FUNCTION_ID         = 1 << 10, /* Use function_id if present to identify PCI hotplug slots */
+        NAMING_16BIT_INDEX              = 1 << 11, /* Allow full 16-bit for the onboard index */
+        NAMING_REPLACE_STRICTLY         = 1 << 12, /* Use udev_replace_ifname() for NAME= rule */
+        NAMING_XEN_VIF                  = 1 << 13, /* Generate names for Xen netfront devices */
+        NAMING_SINGLE_FUNCTION_NO_SLOT  = 1 << 14, /* Don't use PCI hotplug slot information if associated PCI device has only single function */
```
keszybz: This is a "negative" name. I think it'd be better to make it into a positive name. This will make the code easier to maintain in the long term, see below.

src/udev/udev-builtin-net_id.c
```diff
@@ -451,8 +451,15 @@ static int dev_pci_slot(sd_device *dev, const LinkInfo *info, NetNames *names) {
...
-                                        return 0;
+                                        if (naming_scheme_has(NAMING_SINGLE_FUNCTION_NO_SLOT) && !is_pci_multifunction(names->pcidev)) {
+                                                log_device_debug(dev, "Not using slot information because the PCI device associated with the hotplug slot is a bridge and the PCI device has single function.");
+                                                return 0;
+                                        }
+
+                                        if (!naming_scheme_has(NAMING_SINGLE_FUNCTION_NO_SLOT)) {
+                                                log_device_debug(dev, "Not using slot information because the PCI device is a bridge.");
+                                                return 0;
+                                        }
```
keszybz: ... maybe something like this:
```C
if (naming_scheme_has(NAMING_BRIDGE_NO_SLOT) && is_pci_bridge(hotplug_slot_dev)) {
      if (naming_scheme_has(NAMING_BRIDGE_MULTIFUNCTION_SLOT) && !is_pci_multifunction(names->pcidev))
            log_device_debug(dev, "Not using slot information because the PCI device associated with the hotplug slot is a bridge and the PCI device has single function.");
             return 0;
     }
     ...
```

This has the advantage that the option is named after the case that if affects, not after the case that it skips.

keszybz:
I *think* it'll be easier to grok if the name is reversed. But if you disagree, ignore my comment.

msekletar:
@keszybz, OK, so let's rename the option. But I am adding dont-merge label because I got some reports from Red Hat QE folks about issues with this change while they were doing testing internally.

msekletar:
This is now updated and should be ready to merge but please wait until we figure out issues regarding testing of this change. I will drop "dont-merge" label once this is fully ready.

msekletar:
I've force pushed my branch with small update to manual page (removed erroneous dot at the end of added paragraph which was causing man rendering problem).

src/shared/netif-naming-scheme.h
```diff
@@ -22,20 +22,21 @@
...
+        NAMING_STABLE_VIRTUAL_MACS       = 1 << 5, /* Use device name to generate MAC, see 6d3646406560 */
+        NAMING_NETDEVSIM                 = 1 << 6, /* Generate names for netdevsim devices, see eaa9d507d855 */
+        NAMING_LABEL_NOPREFIX            = 1 << 7, /* Don't prepend ID_NET_LABEL_ONBOARD with interface type prefix */
+        NAMING_NSPAWN_LONG_HASH          = 1 << 8, /* Shorten nspawn interfaces by including 24bit hash, instead of simple truncation  */
+        NAMING_BRIDGE_NO_SLOT            = 1 << 9, /* Don't use PCI hotplug slot information if the corresponding device is a PCI bridge */
+        NAMING_SLOT_FUNCTION_ID          = 1 << 10, /* Use function_id if present to identify PCI hotplug slots */
+        NAMING_16BIT_INDEX               = 1 << 11, /* Allow full 16-bit for the onboard index */
+        NAMING_REPLACE_STRICTLY          = 1 << 12, /* Use udev_replace_ifname() for NAME= rule */
+        NAMING_XEN_VIF                   = 1 << 13, /* Generate names for Xen netfront devices */
+        NAMING_BRIDGE_MUTLIFUNCTION_SLOT = 1 << 14, /* Use PCI hotplug slot information associated with bridge, but only if PCI device is multifunction */
```
poettering: mutli → multi

src/udev/udev-builtin-net_id.c
poettering: not addressed?

src/shared/netif-naming-scheme.h
msekletar: Good catch!

msekletar:
@poettering I've fixed the typo and marked the comment from @keszybz as addressed.

yuwata:
@msekletar Is the tag 'dont-merge' still needed?

msekletar:
RH networking QE folks got back to me and are reporting that change is fine.

PR #22846, opened 2022-03-23
portable: allow reattaching when one image has a version and the other does not

bluca:
A reattach might go from img.raw to img_0.1.raw or viceversa, but this is
not allowed right now as we try to match the full name.

Also take into account that running strcspn(a, '/') on an image name, without
leading path, will return the length of the full string, but the versions
might be different so they won't match, eg:

img_0.1.raw -> 12
img_0.1.1.raw -> 14

So adjust the check to take that into account, and skip it if we are not
dealing with directories

### DIFF
--- a/src/portable/portable.c
+++ b/src/portable/portable.c
@@ -1429,17 +1429,28 @@ static bool marker_matches_images(const char *marker, const char *name_or_path, 
                         size_t l;
 
                         /* We shall match against a path. Let's ignore any prefix here though, as often there are many ways to
-                        * reach the same file. However, in this mode, let's validate any file suffix. */
+                         * reach the same file. However, in this mode, let's validate any file suffix.
+                         * But also ensure that we don't fail if both components don't have a '/' at all
+                         * (strcspn returns the full length of the string in that case, which might not
+                         * match as the versions might differ). */
 
                         l = strcspn(a, "/");
                         b = last_path_component(*image_name_or_path);
 
-                        if (strcspn(b, "/") != l)
+                        if ((a[l] != '/') != !strchr(b, '/')) /* One is a directory, the other is not */
+                                return false;
+
+                        if (a[l] != 0 && strcspn(b, "/") != l)
                                 return false;
 
                         underscore = strchr(b, '_');
                         if (underscore)
                                 l = underscore - b;
+                        else { /* Either component could be versioned */
+                                underscore = strchr(a, '_');
+                                if (underscore)
+                                        l = underscore - a;
+                        }
 
                         if (!strneq(a, b, l))
                                 return false;
--- a/test/units/testsuite-29.sh
+++ b/test/units/testsuite-29.sh
@@ -88,6 +88,14 @@ systemctl is-active app1.service
 status="$(portablectl is-attached --extension app1 minimal_0)"
 [[ "${status}" == "running-runtime" ]]
 
+# Ensure that adding or removing a version to the image doesn't break reattaching
+cp /usr/share/app1.raw /tmp/app1_2.raw
+portablectl "${ARGS[@]}" reattach --now --runtime --extension /tmp/app1_2.raw /usr/share/minimal_1.raw app1
+
+systemctl is-active app1.service
+status="$(portablectl is-attached --extension app1_2 minimal_1)"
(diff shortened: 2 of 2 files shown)


src/portable/portable.c
```diff
@@ -1425,21 +1425,32 @@ static bool marker_matches_images(const char *marker, const char *name_or_path,
...
                         size_t l;
 
                         /* We shall match against a path. Let's ignore any prefix here though, as often there are many ways to
-                        * reach the same file. However, in this mode, let's validate any file suffix. */
+                         * reach the same file. However, in this mode, let's validate any file suffix. But also ensure that we
+                         * don't fail if both components don't have a '/' at all (strcspn returns the full length of the string
+                         * in that case, which might not match as the versions might differ). */
 
+                        e = strchr(a, '/');
                         l = strcspn(a, "/");
```
poettering, line 1437: these two lines to some level do the same thing but one returns a pointer, the other a length. one of them should go or be replaced by the other.

poettering: i.e.:

```c
l = e ? e - a : strlen(a);
```

src/portable/portable.c
```diff
@@ -1425,21 +1425,32 @@ static bool marker_matches_images(const char *marker, const char *name_or_path,
...
                         b = last_path_component(*image_name_or_path);
 
-                        if (strcspn(b, "/") != l)
+                        if (!!e != !!strchr(b, '/')) /* One is a directory, the other is not */
```
poettering: why not use `(a[l] != '/') == !strchr(b, '/'))` as test? if so, you don't need e at all..

poettering:
tbh, i'd prefer if the comparison would be split out into a helper func with a test case. it's too complicated to follow fully with all side-effects

src/portable/portable.c
bluca: done - but with `!=` (if one has the `/` and the other hasn't, then its' not a match and we return false), and `e` is necessary as the next check needs to be skipped if it's not defined

bluca:
> tbh, i'd prefer if the comparison would be split out into a helper func with a test case. it's too complicated to follow fully with all side-effects

that would be good, but let's do it later please - it's an existing function, and this is a user-reported issue and I'd like to have it fixed in the next release

src/portable/portable.c
```diff
@@ -1425,21 +1425,32 @@ static bool marker_matches_images(const char *marker, const char *name_or_path,
...
                         size_t l;
 
                         /* We shall match against a path. Let's ignore any prefix here though, as often there are many ways to
-                        * reach the same file. However, in this mode, let's validate any file suffix. */
+                         * reach the same file. However, in this mode, let's validate any file suffix. But also ensure that we
+                         * don't fail if both components don't have a '/' at all (strcspn returns the full length of the string
+                         * in that case, which might not match as the versions might differ). */
 
-                        l = strcspn(a, "/");
+                        e = strchr(a, '/');
```
poettering: strchrnul() then?

src/portable/portable.c
```diff
@@ -1425,21 +1425,32 @@ static bool marker_matches_images(const char *marker, const char *name_or_path,
...
+                        l = e ? (size_t)(e - a) : strlen(a);
                         b = last_path_component(*image_name_or_path);
 
-                        if (strcspn(b, "/") != l)
+                        if ((a[l] != '/') != !strchr(b, '/')) /* One is a directory, the other is not */
+                                return false;
+
+                        if (e && strcspn(b, "/") != l)
```
poettering: and if you use strchrnul(), then you can replace the first part of the expression with `*e != 0`

bluca:
@poettering updated as suggested, PTAL

src/portable/portable.c
poettering: actually, not entirely sure why bother at all, you can just use strcspn() as before and check check with 'a[l] != 0` if you found the slash or not.

bluca: ok, updated again

poettering:
lgtm

bluca:
focal-arm64 has been stuck in the queue for more than 15 hours, looks like infra issues

PR #22847, opened 2022-03-24
journal-remote: request_handler return 0 when using -Dgnutls=false

JetXujing:
when using -Dgnutls=false to build systemd and if
journal_remote_server_global->check_trust is true, request_handler
will return code which is random.

### DIFF
--- a/src/journal-remote/journal-remote-main.c
+++ b/src/journal-remote/journal-remote-main.c
@@ -266,7 +266,8 @@ static mhd_result request_handler(
                 void **connection_cls) {
 
         const char *header;
-        int r, code, fd;
+        int r, fd;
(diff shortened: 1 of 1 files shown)


yuwata:
Thank you for investigating the issue, however, initializing the variable with 0 looks not a valid way to fix the issue.
I opened PR #22848 to fix the issue in another way. Could you test the PR?

JetXujing:
> Thank you for investigating the issue, however, initializing the variable with 0 looks not a valid way to fix the issue. I opened PR #22848 to fix the issue in another way. Could you test the PR?

In fact, I'm not very familiar with the code, I just saw check_permissions setting *code default to 0, so I did this.

PR #22851, opened 2022-03-24
inotify-util: declare iterator in FOREACH_INOTIFY_EVENT()

yuwata:
This also makes the macro check if the event is actually in the buffer.

### DIFF
--- a/src/basic/inotify-util.h
+++ b/src/basic/inotify-util.h
@@ -6,12 +6,31 @@
 #include <stddef.h>
 #include <sys/inotify.h>
 
+#include "log.h"
+
 #define INOTIFY_EVENT_MAX (offsetof(struct inotify_event, name) + NAME_MAX + 1)
 
-#define FOREACH_INOTIFY_EVENT(e, buffer, sz) \
-        for ((e) = &buffer.ev;                                \
-             (uint8_t*) (e) < (uint8_t*) (buffer.raw) + (sz); \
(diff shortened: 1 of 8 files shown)


src/basic/inotify-util.h
```diff
@@ -8,10 +8,16 @@
...
-             (uint8_t*) (e) < (uint8_t*) (buffer.raw) + (sz); \
-             (e) = (struct inotify_event*) ((uint8_t*) (e) + sizeof(struct inotify_event) + (e)->len))
+#define _FOREACH_INOTIFY_EVENT(e, buffer, sz, start, end)               \
+        for (struct inotify_event                                       \
+                     *start = &((buffer).ev),                           \
+                     *end = (struct inotify_event*) ((uint8_t*) start + (sz)), \
+                     *e = start;                                        \
+             (uint8_t*) e + sizeof(struct inotify_event) <= (uint8_t*) end && \
+             (uint8_t*) e + sizeof(struct inotify_event) + e->len <= (uint8_t*) end; \
+             e = (struct inotify_event*) ((uint8_t*) e + sizeof(struct inotify_event) + e->len))
```
keszybz, line 24: I'm not sure if the added checks are useful. It's part of the inotify api that events in the buffer are delivered atomically, i.e. if `e < start + size`, we can assume that `*e` is a valid event. Maybe this should be an assert instead?

yuwata: `assert()` may be used, but for safety, in the revised version, I added a log message about the invalid event,
as usually we use `assert()` for _our_ programming issues.

bluca (approved):

PR #22854, opened 2022-03-24
xdg-autostart-service: Fix binary escaping and simplify code a bit

benzea:
Instead of escaping each component separately, we can instead use
quote_command_line. Doing so simplifies the code and fixes an issue
where spaces inside the executable name were not escaped.

Co-Authored-By: David Edmundson <kde@davidedmundson.co.uk>

Superseeds #22831

CC: @davidedmundson


keszybz:
LGTM.

PR #22855, opened 2022-03-24
TEST-68: replace daemon-reload with separate handler units

keszybz:


bluca (approved):

mrc0mmand (approved):
I checked the various CIs manually and all seems to be in order, so LGTM as well.

PR #22858, opened 2022-03-24
portable: clarify error when detach fails to match to existing unit

bluca:
Clarify that it's looking at attached units, not at units inside the referenced
image. Also take into account extensions.

### DIFF
--- a/src/portable/portable.c
+++ b/src/portable/portable.c
@@ -1520,6 +1520,7 @@ int portable_detach(
 
         _cleanup_(lookup_paths_free) LookupPaths paths = {};
         _cleanup_set_free_ Set *unit_files = NULL, *markers = NULL;
+        _cleanup_free_ char *extensions = NULL;
         _cleanup_closedir_ DIR *d = NULL;
         const char *where, *item;
         int ret = 0;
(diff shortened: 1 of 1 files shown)


src/portable/portable.c
```diff
@@ -1674,8 +1675,20 @@ int portable_detach(
...
+        log_debug("No unit files associated with '%s%s%s' found attached to the system. Image not attached?",
+                  name_or_path,
+                  isempty(extensions) ? "" : "' or any of its extensions '",
+                  isempty(extensions) ? "" : extensions);
+        return sd_bus_error_setf(error,
+                                 BUS_ERROR_NO_SUCH_UNIT,
+                                 "No unit files associated with '%s%s%s' found attached to the system. Image not attached?",
+                                 name_or_path,
+                                 isempty(extensions) ? "" : "' or any of its extensions '",
+                                 isempty(extensions) ? "" : extensions);
```
poettering: please reuse the message. i.e.:

```c
r = sd_bus_error_setf(error, …);
log_debug_errno(r, "%s", error->message);
return r;
```

i.e. no need to have the very same format string in place twice here.

bluca: pre-existing condition :-) but sure, fixed

keszybz:
LGTM.

src/portable/portable.c
```diff
@@ -1674,8 +1675,18 @@ int portable_detach(
...
+                return -ENOMEM;
+
+        r = sd_bus_error_setf(error,
+                              BUS_ERROR_NO_SUCH_UNIT,
+                              "No unit files associated with '%s%s%s' found attached to the system. Image not attached?",
+                              name_or_path,
+                              isempty(extensions) ? "" : "' or any of its extensions '",
+                              isempty(extensions) ? "" : extensions);
+        log_debug_errno(r, "%s", error->message);
+        return r;
```
poettering: actually you can merge these two lines even:

```c
return log_debug_errno(r, "%s", error->message);
```

bluca: done

poettering:
lgtm

Issue #22642, opened 2022-03-01
journald: processing of SIGTERM can get starved by persistent incoming journal messages


dustymabe:
# Description

We are seeing [an issue in Fedora CoreOS](https://github.com/coreos/fedora-coreos-tracker/issues/1108) where the `systemd-journald` process gets leaked from the initramfs into the real root and is constantly trying to create a new journal file/directory structure in old root from the initramfs.

Our best efforts to characterize this behavior have lead us to the conclusion that the `systemd-journald` gets in a situation where the [sent SIGTERM](https://github.com/systemd/systemd/blob/c8aeb9d672fac7ac2d1e350431b7b4e734b90a5d/src/core/main.c#L1903-L1905) gets starved because of its [lowered priority](https://github.com/systemd/systemd/blob/c8aeb9d672fac7ac2d1e350431b7b4e734b90a5d/src/journal/journald-server.c#L1507-L1508).

How could this happen? What we've noticed by inspection is that `systemd-journald` is continously trying to create `/run/` so it can then create the journal in `/run/log/...`. The attempt to try to create `/run/` fails because of an SELinux denial (we think).

We're not sure exactly but we think the SELinux denial is happening because of the switch to the real root. The `systemd-journald` process still has a `cwd` in the initramfs `/` but now that SELinux is loaded `systemd-journald` can't create things in `/` (it's trying to create `/run/`):

```
avc:  denied  { associate } for  pid=254 comm="systemd-journal" name="run" scontext=system_u:object_r:var_run_t:s0 tcontext=system_u:object_r:root_t:s0 tclass=filesystem permissive=0^M
```

The SELinux denial triggers another audit message to be written to the journal, which then gets added to the queue of things that are higher priority than processing the SIGTERM. We're fairly confident in this theory because on a system where `systemd-journald` is in this state we can simply `setenforce 0` and the leaked `systemd-journald` will then process the SIGTERM immediately.

SELinux context of the leaked `systemd-journald` process:

```
system_u:system_r:kernel_t:s0   root         318       1 92 18:43 ?        00:00:58 @journald
```

CWD of the leaked `systemd-journald` process:

```
[root@cosa-devsh ~]# ls -l /proc/318/cwd
lrwxrwxrwx. 1 root root 0 Mar  1 18:44 /proc/318/cwd -> /
[root@cosa-devsh ~]# ls -l /proc/318/cwd/
total 0
```

Stace (`strace -p $PID`) for the process shows us this loop over and over:

```
[...]

PR #22861, opened 2022-03-24
journald: don't let SIGTERM starve indefinitely

poettering:
Fixes: #22642

### DIFF
--- a/src/libsystemd/sd-journal/journal-file.c
+++ b/src/libsystemd/sd-journal/journal-file.c
@@ -1931,10 +1931,17 @@ static int post_change_thunk(sd_event_source *timer, uint64_t usec, void *userda
 }
 
 static void schedule_post_change(JournalFile *f) {
+        sd_event *e;
         int r;
 
         assert(f);
         assert(f->post_change_timer);
+
+        assert_se(e = sd_event_source_get_event(f->post_change_timer));
+
+        /* If we are aleady going down, post the change immediately. */
+        if (IN_SET(sd_event_get_state(e), SD_EVENT_EXITING, SD_EVENT_FINISHED))
+                goto fail;
 
         r = sd_event_source_get_enabled(f->post_change_timer, NULL);
         if (r < 0) {


cgwalters:
I think in order to avoid the AVC denial though, we need to block the switchroot on this, right?

dustymabe:
> I think in order to avoid the AVC denial though, we need to block the switchroot on this, right?

I think you might be right, but just making sure the process eventually does go down the system will be fine. An AVC denial might get logged but the system will be working fine.

poettering:
> I think in order to avoid the AVC denial though, we need to block the switchroot on this, right?

We so far explicitly exclude journald from it, so that it can collect logs from the entire transition, i.e. is killed as late as possible.

src/journal/journald-server.c
```diff
@@ -1447,12 +1447,82 @@ static int dispatch_sigusr2(sd_event_source *es, const struct signalfd_siginfo *
...
+        r = sd_event_add_defer(s->event, &news, NULL, NULL); /* NULL handler means → exit when triggered */
+        if (r < 0) {
+                log_error_errno(r, "Failed to allocate exit idle event handler: %m");
+                goto fail;
+        }
+
+        (void) sd_event_source_set_description(news, "exit-idle");
+
+        /* Run everything relevant before this. */
+        r = sd_event_source_set_priority(news, SD_EVENT_PRIORITY_NORMAL+20);
```
DaanDeMeyer, line 1485: Nit: whitespace between the '+' missing here

src/journal/journald-server.c
```diff
@@ -1447,12 +1447,82 @@ static int dispatch_sigusr2(sd_event_source *es, const struct signalfd_siginfo *
...
+         * written to disk. These goals are incompatible, hence we try to find a middle ground: we'll process
+         * SIGTERM with high priority, but from the handler (this one right here) we'll install two new event
+         * sources: one low priority idle one that will issue the exit once everything else is processed (and
+         * which is hopefully the regular, clean codepath); and one high priority timer that acts as safety
+         * net: if our idle handler isn't run within 10s, we'll exit anyway.
+         *
+         * TLDR: we'll exit either when everything is processed, or after 10s max, depending on what happens
+         * first.
+         *
+         * Note that exiting before the idle event hit doesn't typically mean that we lose any data, as
```
daandemeyer: typo => "hit" => "is hit"

src/journal/journald-server.c
```diff
@@ -1447,12 +1447,82 @@ static int dispatch_sigusr2(sd_event_source *es, const struct signalfd_siginfo *
...
+                goto fail;
+        }
+
+        news = sd_event_source_unref(news);
+
+        log_debug("Exit event sources are now pending.");
+        return 0;
+
+fail:
         sd_event_exit(s->event, 0);
```
daandemeyer, line 1526: Shouldn't we exit the event loop with a nonzero value in case of abnormal failures?

src/journal/journald-server.c
```diff
@@ -1447,12 +1447,82 @@ static int dispatch_sigusr2(sd_event_source *es, const struct signalfd_siginfo *
...
+
+        r = sd_event_add_time_relative(s->event, &news, CLOCK_MONOTONIC, 10 * USEC_PER_SEC, 0, NULL, NULL);
+        if (r < 0) {
+                log_error_errno(r, "Failed to allocate exit timeout event handler: %m");
+                goto fail;
+        }
+
+        (void) sd_event_source_set_description(news, "exit-timeout");
+
+        r = sd_event_source_set_priority(news, SD_EVENT_PRIORITY_IMPORTANT-20); /* This is a safety net, with highest priority */
```
DaanDeMeyer, line 1508: Ditto for whitespace like above

src/journal/journald-server.c
```diff
@@ -1504,8 +1574,8 @@ static int setup_signals(Server *s) {
         if (r < 0)
                 return r;
 
-        /* Let's process SIGTERM late, so that we flush all queued messages to disk before we exit */
-        r = sd_event_source_set_priority(s->sigterm_event_source, SD_EVENT_PRIORITY_NORMAL+20);
+        /* Let's process SIGTERM early, so that we definitely react to it */
+        r = sd_event_source_set_priority(s->sigterm_event_source, SD_EVENT_PRIORITY_IMPORTANT-10);
```
DaanDeMeyer, line 1578: Ditto whitespace

src/journal/journald-server.c
```diff
@@ -1515,7 +1585,7 @@ static int setup_signals(Server *s) {
         if (r < 0)
                 return r;
 
-        r = sd_event_source_set_priority(s->sigint_event_source, SD_EVENT_PRIORITY_NORMAL+20);
+        r = sd_event_source_set_priority(s->sigint_event_source, SD_EVENT_PRIORITY_IMPORTANT-10);
```
DaanDeMeyer, line 1588: Ditto whitespace

DaanDeMeyer (approved):
LGTM, few nits and one question

src/journal/journald-server.c
poettering: Dunno. It doesn't really matter here… It's one of the most graceful places we could exit. I think a log message plus graceful handling is totally appropriate. I wouldn't make something fatal that really doesn't have to be fatal...

cgwalters:
> An AVC denial might get logged but the system will be working fine.

But we want to be able to check as part of our tests that no AVC denials are logged - xref https://github.com/coreos/coreos-assembler/commit/fe010b25a5ec4f4e16999e926f90b4995e8525eb

And it's not just that, but the policy load also blocks journald from flushing these messages IIUC.

Hmm...why is journald-in-initramfs buffering in its heap and then flushing to `/run`?  Why doesn't it just immediately write messages into a file in `/run`?

poettering:
> Hmm...why is journald-in-initramfs buffering in its heap and then flushing to `/run`? Why doesn't it just immediately write messages into a file in `/run`?

It's not.

yuwata (approved):

poettering:
So yeah, i guess we should somehow wait for journald to die before the transition. While my sympathy for the selinux issue is limited (not a fan of the current selinux situation in general) it *is* kinda ugly that journald tries to write to /run after /run and everything else has already been cleared out.

dustymabe:
Preliminary testing of this looks promising: https://github.com/coreos/fedora-coreos-tracker/issues/1108#issuecomment-1080661906

We'll see what happens when it goes through our full CI run.

cgwalters:
> While my sympathy for the selinux issue is limited (not a fan of the current selinux situation in general) it is kinda ugly that journald tries to write to /run after /run and everything else has already been cleared out.

For sure, SELinux adds a whole lot of complexity to everything.  But, it is good at finding bugs like this.  Really that's the core tradeoff; needing to add random permissions to the centralized policy and/or rework things in a way that's better for policy, but it also has mitigated or prevented a number of CVEs over the years.

PR #22867, opened 2022-03-25
add small "udevadm lock" tool for locking a block device

poettering:
This adds a tiny tool "systemd-lockdev" that will correctly lock a block device as per https://systemd.io/BLOCK_DEVICE_LOCKING/

### DIFF
--- a/docs/BLOCK_DEVICE_LOCKING.md
+++ b/docs/BLOCK_DEVICE_LOCKING.md
@@ -75,6 +75,25 @@ And please keep in mind: BSD file locks (`flock()`) and POSIX file locks
 orthogonal. The scheme discussed above uses the former and not the latter,
 because these types of locks more closely match the required semantics.
 
+If multiple devices are to be locked at the same time (for example in order to
+format a RAID file system), the devices should be locked in the order of the
+the device nodes' major numbers (primary ordering key, ascending) and minor
+numbers (secondary ordering key, ditto), in order to avoid ABBA locking issues
+between subsystems.
+
+Note that the locks should only be taken while the device is repartitioned,
+file systems formatted or `dd`'ed in, and similar cases that
+apply/remove/change superblocks/partition information. It should not be held
+during normal operation, i.e. while file systems on it are mounted for
+application use.
+
+The [`udevadm
+lock`](https://www.freedesktop.org/software/systemd/man/udevadm.html) command
+is provided to lock block devices following this scheme from the command line,
+for the use in scripts and similar. (Note though that it's typically preferable
+to use native support for block device locking in tools where that's
+available.)
+
 Summarizing: it is recommended to take `LOCK_EX` BSD file locks when
 manipulating block devices in all tools that change file system block devices
 (`mkfs`, `fsck`, …) or partition tables (`fdisk`, `parted`, …), right after
--- a/man/udevadm.xml
+++ b/man/udevadm.xml
@@ -51,6 +51,9 @@
     <cmdsynopsis>
       <command>udevadm wait <optional>options</optional> <replaceable>device|syspath</replaceable></command>
     </cmdsynopsis>
+    <cmdsynopsis>
+      <command>udevadm lock <optional>options</optional> <replaceable>command</replaceable></command>
+    </cmdsynopsis>
   </refsynopsisdiv>
 
   <refsect1><title>Description</title>
@@ -747,6 +750,87 @@
         <xi:include href="standard-options.xml" xpointer="help" />
       </variablelist>
     </refsect2>
+
+    <refsect2>
+      <title>udevadm lock
+      <arg choice="opt"><replaceable>options</replaceable></arg>
+      <arg choice="opt"><replaceable>command</replaceable></arg>
+      …
+      </title>
+
+      <para><command>udevadm lock</command> takes an (advisory) exclusive lock(s) on a block device (or
+      multiple therof), as per <ulink url="https://systemd.io/BLOCK_DEVICE_LOCKING">Locking Block Device
+      Access</ulink> and invokes a program with the lock(s) taken. When the invoked program exits the lock(s)
+      are automatically released.</para>
+
+      <para>This tool is in particular useful to ensure that
+      <citerefentry><refentrytitle>systemd-udevd.service</refentrytitle><manvolnum>8</manvolnum></citerefentry>
+      does not probe a block device while changes are made to it, for example partitions created or file
+      systems formatted. Note that many tools that interface with block devices natively support taking
+      relevant locks, see for example
+      <citerefentry><refentrytitle>sfdisk</refentrytitle><manvolnum>8</manvolnum></citerefentry>'s
+      <option>--lock</option> switch.</para>
+
+      <para>The command expects at least one block device specified via <option>--device=</option> or
+      <option>--backing=</option>, and a command line to execute as arguments.</para>
+
+      <variablelist>
+        <varlistentry>
+          <term><option>--device=<replaceable>DEVICE</replaceable></option></term>
+          <term><option>-d <replaceable>DEVICE</replaceable></option></term>
+
+          <listitem><para>Takes a path to a device node of the device to lock. This switch may be used
+          multiple times (and in combination with <option>--backing=</option>) in order to lock multiple
+          devices. If a partition block device node is specified the containing "whole" block device is
+          automatically determined and used for the lock, as per the specification. If multiple devices are
+          specified, they are deduplicated, sorted by the major/minor of their device nodes and then locked
+          in order.</para>
+
+          <para>This switch must be used at least once, to specify at least one device to
+          lock. (Alternatively, use <option>--backing=</option>, see below.)</para></listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><option>--backing=<replaceable>PATH</replaceable></option></term>
+          <term><option>-b <replaceable>PATH</replaceable></option></term>
+
+          <listitem><para>If a path to a device node is specified, identical to
+          <option>--device=</option>. However, this switch alternatively accepts a path to a regular file or
+          directory, in which case the block device of the file system the file/directory resides on is
+          automatically determined and used as if it was specified with
+          <option>--device=</option>.</para></listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><option>--timeout=<replaceable>SECS</replaceable></option></term>
+          <term><option>-t <replaceable>SECS</replaceable></option></term>
+
+          <listitem><para>Specifies how long to wait at most until all locks can be taken. Takes a value in
+          seconds, or in the usual supported time units, see
+          <citerefentry><refentrytitle>systemd.time</refentrytitle><manvolnum>7</manvolnum></citerefentry>. If
+          specified as zero the lock is attempted and if not successful the invocation will immediately
+          fail. If passed as <literal>infinity</literal> (the default) the invocation will wait indefinitely
+          until the lock can be acquired. If the lock cannot be taken in the specified time the specified
+          command will not be executed and the invocation will fail.</para></listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><option>--print</option></term>
+          <term><option>-p</option></term>
+
+          <listitem><para>Instead of locking the specified device(s) and executing a command, just print the
+          device path(s) that would be locked, and execute no command. This command is useful to determine
+          the "whole" block device in case a partition block device is specified. The devices will be sorted
+          by their device node major number as primary ordering key and the minor number as secondary
+          ordering key (i.e. they are shown in the order they'd be locked). Note that the number of lines
+          printed here can be less than the the number of <option>--device=</option> and
+          <option>--backing=</option> switches specified in case these resolve to the same "whole"
+          devices.</para></listitem>
+        </varlistentry>
+
+        <xi:include href="standard-options.xml" xpointer="help" />
+      </variablelist>
+    </refsect2>
   </refsect1>
 
   <refsect1>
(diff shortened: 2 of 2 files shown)


poettering:
I guess we should start using this in some of the test cases. @mrc0mmand?

keszybz:
Does it work for the case where btrfs raid is used?

src/lockdev/lockdev.c
```diff
@@ -0,0 +1,300 @@
...
+
+                        r = safe_fork("(timed-flock)", FORK_DEATHSIG|FORK_LOG, &flock_pid);
+                        if (r < 0)
+                                return r;
+                        if (r == 0) {
+                                /* Child */
+
+                                if (flock(fd, LOCK_EX) < 0) {
+                                        log_error_errno(errno, "Failed to lock device '%s': %m", whole_node);
+                                        _exit(EXIT_FAILURE);
```
yuwata:
```
../src/lockdev/lockdev.c: In function ‘run’: 05:03
../src/lockdev/lockdev.c:235:41: error: implicit declaration of function ‘_exit’ [-Werror=implicit-function-declaration] 05:03
  235 |                                         _exit(EXIT_FAILURE); 05:03
      |                                         ^~~~~ 05:03
../src/lockdev/lockdev.c:235:41: error: incompatible implicit declaration of built-in function ‘_exit’ [-Werror] 05:03
../src/lockdev/lockdev.c:238:33: error: incompatible implicit declaration of built-in function ‘_exit’ [-Werror] 05:03
  238 |                                 _exit(EXIT_SUCCESS); 05:03
      |                                 ^~~~~ 05:03
../src/lockdev/lockdev.c:291:17: error: implicit declaration of function ‘execvp’ [-Werror=implicit-function-declaration] 05:03
  291 |                 execvp(arg_cmdline[0], arg_cmdline); 05:03
      |                 ^~~~~~ 05:03
../src/lockdev/lockdev.c:294:17: error: incompatible implicit declaration of built-in function ‘_exit’ [-Werror] 05:03
  294 |                 _exit(EXIT_FAILURE); 05:03
      |                 ^~~~~ 05:03
```

yuwata (changes requested):
- Why not to introduce a new verb in `udevadm`, instead of introducing a new executable?
- How about to add `--device` and `--file` options, and drop `--find` option? And also make these options specified multiple times, to support multi device filesystems.
- `--find` seems to not support files in btrfs raid.

poettering:
> * How about to add `--device` and `--file` options, and drop `--find` option? And also make these options specified multiple times, to support multi device filesystems.

So I wanted to avoid the "--file" switch, because I wanted to emphasize that this is not about locking files, but devices. Via a switch --file it would kinda communicate that we'd actually lock a regular file here, but the tool will never do that, it will *find* the backing block device for that file, and lock that. This is important in particular because in other, but related contexts (notable systemd-homed) we'll actually lock regular files too (i.e. we'll a times lock the loopback device node we create for each home dir *and* the regular file that backs it). (Moreover, this also is capable of figuring out the block device of a directory, not just regular files, hence `--file` might be slightly misleading anyway, in particular as I'd assume real-life invocations would probably be more with dirs than regular files.)

Hence, I think we should stick to the `--find` naming here (or some other word that communicates clearly that we don't take what's specified but try to track what it is backed by). 

does that make sense?

poettering:
> * `--find` seems to not support files in btrfs raid.

Yes, it currently does not. The tool (like our entire other codebase) is not capable of tracking devices that back btrfs raid. Adding this would be something we definitely can add later, and I think there would be quite some value in that, given that locking multiple devices is alway hard, due to locking ABBA problems, and we could neatly handle that, i.e. sort dev_ts or so. But I'd leave that for a later PR really. It's not something we have to include right away.

poettering:
> Does it work for the case where btrfs raid is used?

Well, that could mean two things:

1. We could eventually add support for `--find` to automatically pick up backing devices of btrfs raid. Right now we only support btrfs file systems backed by a single device. Adding this isn't hard, and I think makes a lot of sense, see my earlier comments.

2. We could add support for specifying more than one devices on the cmdline, that shall be locked together. THis would make a ton of sense too, also because of the ABBA think. That said, you can chain systemd-lockdev to make this work already.

Hence, yes you can make it work if you want to set up a btrfs raid. But no, it won't make this particularly easy, and yes, it should make it easier eventually, but this doesn't have to be done in the initial PR I think.

poettering:
> So this could basically replace this commit [41d1aaea64](https://github.com/systemd/systemd/commit/41d1aaea64)?

That's the idea, yes.

poettering:
force pushed a new version, that hopefully fixes the CI issue.

(also added a note to docs/BLOCK_DEVICE_LOCKING.md pointing to the new tool)

ptal

yuwata:
> > Why not to introduce a new verb in udevadm, instead of introducing a new executable?
> 
> So my thinking was that the flock()-based block device locking is kinda independent of udev. i.e. conceptually you can ignore the existance of udev and still use it to ensure multiple partitioner tools don't step on each other's toes or so. i.e. it doesn't depend on the udev db or anything like: it's a concept that makes sense even if all you have a plain device node (like a simplistic container manager might give you). thus, i figured it should be a separate tool.
> 
> as a counter example, the "udevadm wait" command I suggested otoh is purely and exclusively about the udev db: it's a tool to for synchronizing on udev db changes, hence I think it fits perfectly under the udevadm umbrella and belongs there.
> 
> It's a blurry line though, I am not tied to makig this a separate tool It would be OK to move it into udevadm too I think. i.e. I am leaning weakely to leaving it a separate tool rather than integrating it into udevadm.
> 
> In a way "systemd-inhibit" was my template for this, which behaves very similar, and is also about synchronizing stuff.
> 
> anyway, other opinions on this? @keszybz?

> So my thinking was that the flock()-based block device locking is kinda independent of udev.

That's true, but still I prefer this to be a part of udevadm. 
- Actually, this works without udevd or udev database. But, this provides a very low-level device management, hence it is good to fit to udevadm.
- I guess 'systemd-' prefix makes users guess the tool is something related to PID1. But this is not...
- Users can safely operate setting up devices by using this tool and 'udevadm wait' in scripts, as you can see in testsuite-64.sh. So, I think it is better both commands are explained in the same man page udevadm(8). And 'udevadm lock' and 'udevadm wait' sounds good pair for me.
- I'd like to this command being always installed in every systems. I guess 99.9% systems (including systemd-less systems) have udevadm. If this is introduced as independent, then this may not be installed or packaged in a wrong place.
(BTW, are there any udev-less systems in the world?? eudev is just a clone of udevd before it became a systemd component).

yuwata:
> > * How about to add `--device` and `--file` options, and drop `--find` option? And also make these options specified multiple times, to support multi device filesystems.
> 
> So I wanted to avoid the "--file" switch, because I wanted to emphasize that this is not about locking files, but devices. Via a switch --file it would kinda communicate that we'd actually lock a regular file here, but the tool will never do that, it will _find_ the backing block device for that file, and lock that. This is important in particular because in other, but related contexts (notable systemd-homed) we'll actually lock regular files too (i.e. we'll at times lock the loopback device node we create for each home dir _and_ the regular file that backs it). (Moreover, this also is capable of figuring out the block device of a directory, not just regular files, hence `--file` might be slightly misleading anyway, in particular as I'd assume real-life invocations would probably be more with dirs than regular files.)
> 
> Hence, I think we should stick to the `--find` naming here (or some other word that communicates clearly that we don't take what's specified but try to track what it is backed by).
> 
> does that make sense?

Actually, the naming '--file' is not good. But, the important part of the suggestion is this command should take multiple devices or files. Also, we should always support regular files unconditionally.
Ideas:
- systemd-lockdev --target /dev/sda5 --target /path/to/file [COMMAND]
- systemd-lockdev /dev/sda5 /path/to/file -- [COMMAND]

poettering:
So I merged this into "udevadm" now, as a new "lock" verb.

I also added support for locking multiple devices at once, with a systematic ordering to avoid ABBA issues.

I did not add automatic btrfs raid device enumeration though. We can easily add that though if there's desire for that.

ptal

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+#include "udevadm.h"
+
+static usec_t arg_timeout_usec = USEC_INFINITY;
+static char **arg_devices = NULL;
+static char **arg_backing = NULL;
+static char **arg_cmdline = NULL;
+static bool arg_print = false;
+
+STATIC_DESTRUCTOR_REGISTER(arg_devices, strv_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_cmdline, strv_freep);
```
yuwata, line 32: missing destructor for `arg_backing`.

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+
+static int help(void) {
+        _cleanup_free_ char *link = NULL;
+        int r;
+
+        r = terminal_urlify_man("udevadm", "1", &link);
```
yuwata: 1 -> 8?

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+
+        printf("%s [OPTIONS...] COMMAND\n"
+               "%s [OPTIONS...] --print\n"
+               "\n%sLock a block device and run a comand.%s\n\n"
+               "  -h --help            Print this message\n"
+               "  -V --version         Print version of the program\n"
+               "  -d --device=DEVICE   Block device to lock\n"
+               "  -b --backing=FILE    File whose backing block device to lock\n"
+               "  -t --timeout=SECS    Block at most the specified time waiting for lock\n"
+               "  -p --print           Only show which block device the lock would be taken on\n"
```
yuwata, line 50: Maybe `--dry-run`?  `udevadm trigger` uses `--dry-run`.

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+        _cleanup_(fdset_freep) FDSet *fds = NULL;
+        _cleanup_free_ dev_t *devnos = NULL;
+        size_t n_devnos = 0;
+        usec_t deadline;
+        pid_t pid;
+        int r;
+
+        log_show_color(true);
+        log_parse_environment();
+        log_open();
```
yuwata: These are not necessary, as `udevadm.c` calls them.

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+        if (r == 0) {
+                /* Child */
+
+                execvp(arg_cmdline[0], arg_cmdline);
+                log_open();
+                log_error_errno(errno, "Failed to execute %s: %m", argv[optind]);
+                _exit(EXIT_FAILURE);
+        }
+
+        return wait_for_terminate_and_check(argv[optind], pid, 0);
```
yuwata: Not a matter, but `arg_cmdline[0]` is better?

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+
+        r = safe_fork("(lock)", FORK_RESET_SIGNALS|FORK_DEATHSIG|FORK_CLOSE_ALL_FDS|FORK_RLIMIT_NOFILE_SAFE|FORK_LOG, &pid);
+        if (r < 0)
+                return r;
+        if (r == 0) {
+                /* Child */
+
+                execvp(arg_cmdline[0], arg_cmdline);
+                log_open();
+                log_error_errno(errno, "Failed to execute %s: %m", argv[optind]);
```
yuwata: `argv[optind]` -> `arg_cmdline[0]`

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+        assert(devnos);
+        assert(n_devnos);
+        assert(*devnos || *n_devnos == 0);
+        assert(device);
+
+        if (stat(device, &st) < 0)
+                return log_error_errno(errno, "Failed to stat '%s': %m", device);
+
+        if (S_ISBLK(st.st_mode))
+                devt = st.st_rdev;
```
yuwata, line 160: This means, `--backing=/dev/sda1` works well as `--device=/dev/sda1`, right?

P.S.
Ah, the man page explicitly say so. Then, why not to the --device accept both device node and regular file?

src/udev/udevadm-lock.c
```diff
@@ -0,0 +1,402 @@
...
+                /* If major(st.st_dev) is zero, this might mean we are backed by btrfs, which needs special
+                 * handing, to get the backing device node. */
+
+                regfd = open(device, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
+                if (regfd < 0)
+                        return log_error_errno(errno, "Failed to open '%s': %m", device);
+
+                /* Extra safety: let's check we are still looking at the same file */
+                if (fstat(regfd, &st2) < 0)
+                        return log_error_errno(errno, "Failed top stat '%s': %m", device);
```
yuwata: 'top' -> 'to'

yuwata (changes requested):

src/udev/udevadm-lock.c
poettering: Hmm, so I think --dry-run would be right if we'd just log to stderr here, i.e. log what we are doing, but not actually do it, for the purpose of debugging/testing. But here we write what we found out to stdout, where it can be processed further, and it is not just for debugging/testing, but also for actually using that returned data for other stuff. Hence I think --dry-run would not be the right name for this, as here the focus should be on generating further usable data, and not just about logging about stuff.

If that makes any sense?

src/udev/udevadm-lock.c
poettering: I kept this separate for a reason, because I got burnt but an automatism like this myself when I developed systemd-repart. The thing is that if you use disk tools on regular files it could mean two things:

1. consider the regular file a disk image and operate on that as a regular file
2. find out the backing disk for that file, and operate on that

Tools like sfdisk/fdisk/cfdisk follow logic 1, while here we are tempted to follow logic 2. But if we pick the wrong logic the results can be fatal. And this happened to me as mentioned: by accident I repartitioned the disk my test image was on rather then the test image myself. And that sucked hard! :-(

I think it's much safer to assume people are looking for logic 1 instead of logic 2 when doing disk management and getting a path to a regular file passed. Hence I think logic 2 should be opt-in, i.e. explicitly require a concious decision to be used by users. My attempt with `--backing=` is exactly that: by naming hte option that way the user hopefully understands that this is about locking the *backing* disk, and we do not lock the regular file itself.

Does that make any sense?

src/udev/udevadm-lock.c
yuwata: Yeah, thanks.

src/udev/udevadm-lock.c
yuwata: Indeed. I agree with you. Thanks.

poettering:
Force pushed new version, with all points addressed. PTAL.

yuwata (approved):
LGTM.

PR #17201, opened 2020-09-29
resolved: turn off that a search domain is derived from the host's fqdn

poettering:
If the hostname of a system is set to an fqdn, glibc traditionally
derives a search domain from it if none is explicitly configured.

This is a bit weird, and we currently don't do that in our own search
path logic.

Following #17193 let's turn this behaviour off for now.

Yes, this has a slight chance of pissing people off who think this
behaviour is good. If this is indeed an issue, we can revisit the issue
but in that case if we readd the concept we should do it properly:
derive the search domain from the fqdn in our codebase too and report it
in resolvectl, and in our generated stub files. But I have the suspicion
most people who set the hostname to an fqdn aren#t even aware of this
behaviour nor want it, so let's wait until people complain.

Fixes: #17193

PR #22868, opened 2022-03-25
RFC: resolved: add option to toggle DNS search on hostdomain

LorbusChris:
If the domain list is empty, since https://github.com/systemd/systemd/pull/17201/commits/c1a072766edafd3bd6d2f626685cbd1fd7a217fe
resolved has been adding a `search .` line to `resolv.conf` in order to
ensure the FQDN of the host does not imply a DNS search domain.

In some cases it may be desirable however to enable this behavior.
In such a case the `search .` line should not be written to
`resolv.conf`.

This commit adds a `SearchHostDomain=` option to resolved's
configuration that can be used toggle this behavior. Defaults to `no`.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1874419

Beware: This is an untested Friday afternoon PR. Only RFC on this approach at this time.

### DIFF
--- a/man/resolved.conf.xml
+++ b/man/resolved.conf.xml
@@ -329,6 +329,14 @@ DNSStubListenerExtra=udp:[2001:db8:0:f102::13]:9953</programlisting>
         url="https://www.iab.org/documents/correspondence-reports-documents/2013-2/iab-statement-dotless-domains-considered-harmful/">IAB
         Statement</ulink>, and may create a privacy and security risk.</para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>SearchHostDomain=</varname></term>
+        <listitem><para>Takes a boolean argument. If <literal>no</literal> (the default), <command>systemd-resolved</command> will add
+        the line <emphasis>search .</emphasis> to <filename>/etc/systemd/resolved.conf</filename>. If <literal>yes</literal>, a search
+        domain that is derived from the hostname is added instead.</para></listitem>
+      </varlistentry>
+
     </variablelist>
   </refsect1>
 
--- a/src/resolve/resolved-gperf.gperf
+++ b/src/resolve/resolved-gperf.gperf
@@ -32,3 +32,4 @@ Resolve.ReadEtcHosts,              config_parse_bool,                    0,     
 Resolve.ResolveUnicastSingleLabel, config_parse_bool,                    0,                   offsetof(Manager, resolve_unicast_single_label)
 Resolve.DNSStubListenerExtra,      config_parse_dns_stub_listener_extra, 0,                   offsetof(Manager, dns_extra_stub_listeners)
 Resolve.CacheFromLocalhost,        config_parse_bool,                    0,                   offsetof(Manager, cache_from_localhost)
+Resolve.SearchHostDomain,          config_parse_bool,                    0,                   offsetof(Manager, search_host_domain)
--- a/src/resolve/resolved-manager.c
+++ b/src/resolve/resolved-manager.c
@@ -584,6 +584,7 @@ int manager_new(Manager **ret) {
                 .need_builtin_fallbacks = true,
                 .etc_hosts_last = USEC_INFINITY,
                 .read_etc_hosts = true,
+                .search_host_domain = false,
         };
 
         r = dns_trust_anchor_load(&m->trust_anchor);
--- a/src/resolve/resolved-manager.h
+++ b/src/resolve/resolved-manager.h
@@ -77,6 +77,7 @@ struct Manager {
         bool need_builtin_fallbacks;
         bool read_resolv_conf;
         bool resolve_unicast_single_label;
+        bool search_host_domain;
 
         struct stat resolv_conf_stat;
 
--- a/src/resolve/resolved-resolv-conf.c
+++ b/src/resolve/resolved-resolv-conf.c
@@ -253,7 +253,7 @@ static void write_resolv_conf_search(
         fputs("\n", f);
 }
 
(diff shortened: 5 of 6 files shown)


poettering:
What's the usecase? If you manually reconfigure resolved like this, why not just add the search domain instead? Not grokking this?

LorbusChris:
The specific need is to get rid of this ugly hack in OKD (OpenShift on Fedora CoreOS): https://github.com/openshift/okd-machine-os/blob/5686ca6a95b95437e70c79d9c72cbcdadd736f3e/overlay.d/99okd/usr/lib/systemd/system/fix-resolv-conf-search.service

There was a whole saga surrounding this, i.e. making DNS work in OKD with resolved and NM in tandem, but I'm not too familiar with the details.

AFAICT _something_ adds the correct search domain to `/run/systemd/resolve/stub-resolv.conf`, but only if `search .` is not already present there.
I also wonder whether it might suffice to let NM manage `/etc/resolv.conf` alone in this case (e.g. by symlinking `/etc/resolv.conf` to `/run/NetworkManager/resolv.conf` instead - it has been said that won't work but I haven't tested that myself).

CC'ing some folks who were involved and/or might be able to explain this issue better than me: @fortinj66 @vrutkovs @dustymabe

man/resolved.conf.xml
```diff
@@ -329,6 +329,14 @@ DNSStubListenerExtra=udp:[2001:db8:0:f102::13]:9953</programlisting>
         url="https://www.iab.org/documents/correspondence-reports-documents/2013-2/iab-statement-dotless-domains-considered-harmful/">IAB
         Statement</ulink>, and may create a privacy and security risk.</para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>SearchHostDomain=</varname></term>
+        <listitem><para>Takes a boolean argument. If <literal>no</literal> (the default), <command>systemd-resolved</command> will add
+        the line <emphasis>search .</emphasis> to <filename>/etc/systemd/resolved.conf</filename>. If <literal>yes</literal>, a search
```
LorbusChris, line 336:
```suggestion
        the line <emphasis>search .</emphasis> to <filename>/run/systemd/resolve/stub-resolv.conf</filename>. If <literal>yes</literal>, a search
```

LorbusChris:
This PR description has a bit more reasoning in it: https://github.com/openshift/okd-machine-os/pull/158

Maybe this means my above assertion is wrong and the correct search domain isn't added to resolv.conf after all in either case - it just breaks for us when `search .` is present, because we need the search domain to be implicitly derived from the hostname.

poettering:
Why doesn't openshift install the search domain it wants explicitly? relying on the obscure logic of glibc to derive one search domain off the hostname if it is set to an fqdn and no search domain otherwise configured is just super duper weird. It falls apart as soon as people set manual search domains anyway, as the logic is disabled in glibc then anyway.

Seriously, fix openshift on this. if you want a search domain configure a search domain like everyone else.

LorbusChris:
Ok, I think that's fair. What would be the right way to set the search domain with resolved programmatically? Do I just append a line with `Domains=<mydomain>` to `/etc/systemd/resolved.conf`?

poettering:
that really depends on where the DNS config comes from in general. DHCP? static global config? per-network networkd .network files? programmatically via NM?

LorbusChris:
It could be from DHCP or static, via NM.

poettering:
> UPI stands for User Provisioned Infrastructure (i.e. bring your own machines). I guess we can focus on this case for now. If there's a preferred way to set the search domain in such a case, please let me know.

Sorry, I can't parse this. Note I have no clue about OpenShift or "fcos"?

Anyway, there's nothing left to fix here? instead maybe post an issue on "fcos" (for whatever that is) or OpenShift?

LorbusChris:
My bad, I'm confused by this myself. FCOS stands for Fedora CoreOS (the rpm-ostree based distro).
I've had a few more conversations and this is what seems to be going on:

The machine is provisioned with static ip and hostname, which are both passed in through the `ip=` kernel arg. On RHEL CoreOS - which *does not* use systemd-resolved - the DNS search domain is then automatically derived from the hostname, which enables the machines in the cluster to discover each other. On FCOS however - which uses systemd-resolved in tandem with NetworkManager - the existence of the `search .` line in `resolv.conf` prevents this automatic derivation, and the machines are then unable to discover each other. Removing the line fixes this.

The question now is how to either A) specifically configure the DNS search domain at provisioning time or B) allow for the automatic derivation to happen on FCOS without resorting to hacking around it with https://github.com/openshift/okd-machine-os/blob/8c7935a25405c9239e310973082bd61dfeded2f3/overlay.d/99okd/usr/lib/systemd/system/fix-resolv-conf-search.service

fortinj66:
Note that this is the code which results in the current behavior.  Prior to this change "it just worked"  

https://github.com/systemd/systemd/pull/17201

poettering:
@LorbusChris it sounds strange to me to rely on that behaviour.

We could probably add a feature in resolved to pick up DNS config from the kernel cmdline. Given that networkd already picks up IP config from there, it would only be natural to pick it up from there too.

or in other words, you then could add `systemd.dns-search-domains=foobar.com` or so on the kernel cmdline.

Would that work?

fortinj66:
As long as that would prevent the `search .` from being added this would work.

The issue really is the added `search .` It causes issues with OpenShift pods and internal Openshift DNS search such that pods cannot resolve addresses in the cluster space.

LorbusChris:
@poettering a `systemd.dns-search-domains=` kernel arg sounds great

LorbusChris:
> can you file an RFE issue requesting that please?

will do.

LorbusChris:
Filed https://github.com/systemd/systemd/issues/24103

PR #22871, opened 2022-03-25
udev: append error code on failure

yuwata:
Addresses https://github.com/systemd/systemd/pull/22717#discussion_r834420878.

### DIFF
--- a/test/units/testsuite-17.03.sh
+++ b/test/units/testsuite-17.03.sh
@@ -3,6 +3,7 @@
 set -ex
 
 test_rule="/run/udev/rules.d/49-test.rules"
+KILL_PID=
 
 setup() {
     mkdir -p "${test_rule%/*}"
@@ -22,18 +23,37 @@ EOF
 teardown() {
     set +e
 
+    if [[ -n "$KILL_PID" ]]; then
+        kill "$KILL_PID"
+    fi
+
+    rm -rf "$TMPDIR"
+
     mv -f /etc/udev/udev.conf.bckp /etc/udev/udev.conf
     rm -f "$test_rule"
     systemctl restart systemd-udevd.service
 }
 
 run_test() {
+    local since
+
     since="$(date '+%F %T')"
 
-    SYSTEMD_LOG_LEVEL=debug udevadm trigger --verbose --settle --action add /dev/null
+    TMPDIR=$(mktemp -d -p /tmp udev-tests.XXXXXX)
+    udevadm monitor --udev --property --subsystem-match=mem >"$TMPDIR"/monitor.txt &
+    KILL_PID="$!"
 
-    for _ in {1..20}; do
+    SYSTEMD_LOG_LEVEL=debug udevadm trigger --verbose --action add /dev/null
+
+    for _ in {1..40}; do
         if coredumpctl --since "$since" --no-legend --no-pager | grep /bin/udevadm ; then
+            kill "$KILL_PID"
+            KILL_PID=
+
+            cat "$TMPDIR"/monitor.txt
+            grep -q 'UDEV_WORKER_FAILED=1' "$TMPDIR"/monitor.txt
+            grep -q 'UDEV_WORKER_SIGNAL=6' "$TMPDIR"/monitor.txt
+            grep -q 'UDEV_WORKER_SIGNAL_NAME=ABRT' "$TMPDIR"/monitor.txt
             return 0
         fi
         sleep .5


bluca (approved):

src/udev/udevd.c
```diff
@@ -363,13 +362,18 @@ static void device_broadcast(sd_device_monitor *monitor, sd_device *dev) {
         if (!monitor)
                 return;
 
+        if (result != EVENT_RESULT_SUCCESS) {
+                (void) device_add_property(dev, "UDEV_WORKER_FAILED", "1");
+                (void) device_add_propertyf(dev, "UDEV_WORKER_ERROR_CODE", "%i", abs(result));
```
poettering: maybe use %s and errno_to_name()?

(or set both the numeric error here?)

I'd call this field UDEV_WORKER_ERRNO btw, to communicate clearly what kind of error code this is, i.e. a UNIX `errno` style error code.

src/udev/udevd.c
```diff
@@ -1541,7 +1543,7 @@ static int on_sigchld(sd_event_source *s, const struct signalfd_siginfo *si, voi
                         device_tag_index(worker->event->dev, NULL, false);
 
                         /* Forward kernel event to libudev listeners */
-                        device_broadcast(manager->monitor, worker->event->dev);
+                        device_broadcast(manager->monitor, worker->event->dev, WEXITSTATUS(status));
```
poettering: hmm, so what is the value supposed to be? an errno style error code (as the code above suggests), or an exit status of a process (i.e. something in the 0…255 range?)

and it can also be EVENT_RESULT_TRY_AGAIN, right?

So what does the value 1 mean in this parameter? it could mean three things, no? EXIT_FAILURE (i.e. a process exited with exit status 1), it could mean EPERM (i.e. we saw an errno failure somewhere, and 1 means EPERM), or it could mean EVENT_RESULT_TRY_AGAIN, which is also defined to 1, no?

I mean, I think it's OK to pass this all through the same channel here, but we should take care of not making things ambiguous. 

i.e. maybe say: values < 0 are errno-style failures, values 0…255 are process returns (because that's all UNIX supports here), and the value 256 or so would mean EVENT_RESULT_TRY_AGAIN or so. i.e. define the enum like this:

```c
typedef enum EventResult {
        EVENT_RESULT_NERRNO_MIN = -ERRNO_MAX,
        EVENT_RESULT_NERRNO_MAX = -1,
        EVENT_RESULT_EXIT_STATUS_BASE = 0,
        EVENT_RESULT_EXIT_STATUS_MAX = 255,
        EVENT_RESULT_TRY_AGAIN = 256,
        _EVENT_RESULT_MAX,
        _EVENT_RESULT_INVALID = -EINVAL,
} EventResult;
```

or something like that?

poettering: and of course the field in the udev event should be differently named depending if it is an errno value or an exit status.

yuwata:
@poettering Thank you for the review. I hope the above comments are addressed. PTAL.

src/udev/udevd.c
```diff
@@ -363,13 +368,28 @@ static void device_broadcast(sd_device_monitor *monitor, sd_device *dev) {
...
+                (void) device_add_property(dev, "UDEV_WORKER_FAILED", "1");
+                if (result < 0) {
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_ERRNO", "%i", -result);
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_ERRNO_NAME", "%s", strna(errno_to_name(result)));
+                } else if (result <= EVENT_RESULT_EXIT_STATUS_MAX)
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_EXIT_STATUS", "%i", result);
+                else {
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_SIGNAL", "%i", result - EVENT_RESULT_SIGNAL_BASE);
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_SIGNAL_NAME", "%s", strna(signal_to_string(result - EVENT_RESULT_SIGNAL_BASE)));
+                }
```
poettering, line 397: doesn't matter at all, but this would be super pretty with gcc range switch statements, i.e.:

```c
switch(result)

case EVENT_RESULT_NERRNO_MIN..EVENT_RESULT_NERRNO_MAX:
        …
case EVENT_RESULT_EXIT_STATUS_BASE..EVENT_RESULT_EXIT_STATUS_MAX:
        …
case EVENT_RESULT_SIGNAL_BASE..EVENT_RESULT_SIGNAL_MAX:
        …
default:
        assert_not_reached();
}
```
or so.

src/udev/udevd.c
```diff
@@ -1541,7 +1559,9 @@ static int on_sigchld(sd_event_source *s, const struct signalfd_siginfo *si, voi
                         device_tag_index(worker->event->dev, NULL, false);
 
                         /* Forward kernel event to libudev listeners */
-                        device_broadcast(manager->monitor, worker->event->dev);
+                        device_broadcast(manager->monitor, worker->event->dev,
+                                         WIFEXITED(status) ? WEXITSTATUS(status) :
```
poettering: doesn't matter but may write this as `EVENT_RESULT_EXIT_STATUS_BASE + WEXITSTATUS(status)`? just to make things even more explicit

poettering:
I like this a lot!

yuwata:
@poettering Thank you! The above suggestions are applied. Upgrading the green label.

src/udev/udevd.c
```diff
@@ -365,13 +370,40 @@ static void device_broadcast(sd_device_monitor *monitor, sd_device *dev) {
         if (!monitor)
                 return;
 
+        if (result != 0) {
+                (void) device_add_property(dev, "UDEV_WORKER_FAILED", "1");
+
+                switch (result) {
+                case EVENT_RESULT_NERRNO_MIN ... EVENT_RESULT_NERRNO_MAX:
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_ERRNO", "%i", -result);
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_ERRNO_NAME", "%s", strna(errno_to_name(result)));
```
keszybz, line 379: Hmm, I think it'd be better to just not set the property if errno_to_name() fails.

src/udev/udevd.c
```diff
@@ -365,13 +370,40 @@ static void device_broadcast(sd_device_monitor *monitor, sd_device *dev) {
...
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_EXIT_STATUS", "%i", result - EVENT_RESULT_EXIT_STATUS_BASE);
+                        break;
+
+                case EVENT_RESULT_TRY_AGAIN:
+                        assert_not_reached();
+                        break;
+
+                case EVENT_RESULT_SIGNAL_BASE ... EVENT_RESULT_SIGNAL_MAX:
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_SIGNAL", "%i", result - EVENT_RESULT_SIGNAL_BASE);
+                        (void) device_add_propertyf(dev, "UDEV_WORKER_SIGNAL_NAME", "%s", strna(signal_to_string(result - EVENT_RESULT_SIGNAL_BASE)));
```
keszybz, line 392: Same here.

keszybz (approved):
LGTM too.

The one suggested change can go in separately later (or not).

keszybz:
focal-amd64 passed. The other focals are being really slow. There is nothing arch-specific here, let's merge.

src/udev/udevd.c
poettering: btw, simple device_add_property() (as opposed to device_add_propertyf()) works too for this string property.

yuwata: Indeed.

PR #22872, opened 2022-03-25
udevadm: introduce 'wait' command

yuwata:
Prompted by https://github.com/systemd/systemd/pull/22717#issuecomment-1067348496.

### DIFF
--- a/src/udev/udev-node.c
+++ b/src/udev/udev-node.c
@@ -598,7 +598,7 @@ int udev_node_apply_permissions(
                 gid_t gid,
                 OrderedHashmap *seclabel_list) {
 
-        const char *devnode, *subsystem, *id = NULL;
+        const char *devnode, *subsystem;
         bool apply_mode, apply_uid, apply_gid;
         _cleanup_close_ int node_fd = -1;
         struct stat stats;
@@ -616,32 +616,24 @@ int udev_node_apply_permissions(
         r = sd_device_get_devnum(dev, &devnum);
         if (r < 0)
                 return log_device_debug_errno(dev, r, "Failed to get devnum: %m");
-        (void) device_get_device_id(dev, &id);
 
         if (streq(subsystem, "block"))
                 mode |= S_IFBLK;
         else
                 mode |= S_IFCHR;
 
-        node_fd = open(devnode, O_PATH|O_NOFOLLOW|O_CLOEXEC);
+        node_fd = sd_device_open(dev, O_PATH|O_CLOEXEC);
         if (node_fd < 0) {
-                if (errno == ENOENT) {
-                        log_device_debug_errno(dev, errno, "Device node %s is missing, skipping handling.", devnode);
+                if (ERRNO_IS_DEVICE_ABSENT(node_fd)) {
+                        log_device_debug_errno(dev, node_fd, "Device node %s is missing, skipping handling.", devnode);
                         return 0; /* This is necessarily racey, so ignore missing the device */
                 }
 
-                return log_device_debug_errno(dev, errno, "Cannot open node %s: %m", devnode);
+                return log_device_debug_errno(dev, node_fd, "Cannot open node %s: %m", devnode);
         }
 
         if (fstat(node_fd, &stats) < 0)
                 return log_device_debug_errno(dev, errno, "cannot stat() node %s: %m", devnode);
-
-        if ((mode != MODE_INVALID && (stats.st_mode & S_IFMT) != (mode & S_IFMT)) || stats.st_rdev != devnum) {
-                log_device_debug(dev, "Found node '%s' with non-matching devnum %s, skipping handling.",
-                                 devnode, strna(id));
-                return 0; /* We might process a device that already got replaced by the time we have a look
-                           * at it, handle this gracefully and step away. */
-        }
 
         apply_mode = mode != MODE_INVALID && (stats.st_mode & 0777) != (mode & 0777);
         apply_uid = uid_is_valid(uid) && stats.st_uid != uid;
--- a/src/udev/udevd.c
+++ b/src/udev/udevd.c
@@ -422,12 +422,11 @@ static int worker_send_result(Manager *manager, int result) {
         return loop_write(manager->worker_watch[WRITE_END], &result, sizeof(result), false);
 }
 
-static int device_get_whole_disk(sd_device *dev, const char **ret) {
+static int device_get_whole_disk(sd_device *dev, sd_device **ret_device, const char **ret_devname) {
         const char *val;
         int r;
 
         assert(dev);
-        assert(ret);
 
         if (device_for_action(dev, SD_DEVICE_REMOVE))
                 goto irrelevant;
@@ -463,16 +462,23 @@ static int device_get_whole_disk(sd_device *dev, const char **ret) {
         if (r < 0)
                 return log_device_debug_errno(dev, r, "Failed to get devname: %m");
 
-        *ret = val;
+        if (ret_device)
+                *ret_device = dev;
+        if (ret_devname)
+                *ret_devname = val;
         return 1;
 
 irrelevant:
-        *ret = NULL;
+        if (ret_device)
+                *ret_device = NULL;
+        if (ret_devname)
+                *ret_devname = NULL;
         return 0;
 }
 
 static int worker_lock_whole_disk(sd_device *dev, int *ret_fd) {
         _cleanup_close_ int fd = -1;
+        sd_device *dev_whole_disk;
         const char *val;
         int r;
 
@@ -484,19 +490,19 @@ static int worker_lock_whole_disk(sd_device *dev, int *ret_fd) {
          * event handling; in the case udev acquired the lock, the external process can block until udev has
          * finished its event handling. */
 
-        r = device_get_whole_disk(dev, &val);
+        r = device_get_whole_disk(dev, &dev_whole_disk, &val);
         if (r < 0)
                 return r;
         if (r == 0)
                 goto nolock;
 
-        fd = open(val, O_RDONLY|O_CLOEXEC|O_NOFOLLOW|O_NONBLOCK);
+        fd = sd_device_open(dev_whole_disk, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
         if (fd < 0) {
-                bool ignore = ERRNO_IS_DEVICE_ABSENT(errno);
+                bool ignore = ERRNO_IS_DEVICE_ABSENT(fd);
 
-                log_device_debug_errno(dev, errno, "Failed to open '%s'%s: %m", val, ignore ? ", ignoring" : "");
+                log_device_debug_errno(dev, fd, "Failed to open '%s'%s: %m", val, ignore ? ", ignoring" : "");
                 if (!ignore)
-                        return -errno;
+                        return fd;
 
                 goto nolock;
         }
@@ -545,15 +551,9 @@ static int worker_mark_block_device_read_only(sd_device *dev) {
         if (STARTSWITH_SET(val, "dm-", "md", "drbd", "loop", "nbd", "zram"))
                 return 0;
 
-        r = sd_device_get_devname(dev, &val);
-        if (r == -ENOENT)
-                return 0;
-        if (r < 0)
-                return log_device_debug_errno(dev, r, "Failed to get devname: %m");
-
-        fd = open(val, O_RDONLY|O_CLOEXEC|O_NOFOLLOW|O_NONBLOCK);
+        fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
         if (fd < 0)
-                return log_device_debug_errno(dev, errno, "Failed to open '%s', ignoring: %m", val);
+                return log_device_debug_errno(dev, fd, "Failed to open '%s', ignoring: %m", val);
 
         if (ioctl(fd, BLKROSET, &state) < 0)
                 return log_device_warning_errno(dev, errno, "Failed to mark block device '%s' read-only: %m", val);
@@ -1093,7 +1093,7 @@ static int event_queue_assume_block_device_unlocked(Manager *manager, sd_device 
          * device is not locked anymore. The assumption may not be true, but that should not cause any
          * issues, as in that case events will be requeued soon. */
 
-        r = device_get_whole_disk(dev, &devname);
+        r = device_get_whole_disk(dev, NULL, &devname);
         if (r <= 0)
                 return r;
 
@@ -1106,7 +1106,7 @@ static int event_queue_assume_block_device_unlocked(Manager *manager, sd_device 
                 if (event->retry_again_next_usec == 0)
                         continue;
 
-                if (device_get_whole_disk(event->dev, &event_devname) <= 0)
+                if (device_get_whole_disk(event->dev, NULL, &event_devname) <= 0)
                         continue;
 
                 if (!streq(devname, event_devname))
@@ -1403,18 +1403,13 @@ static int synthesize_change(sd_device *dev) {
             !startswith(sysname, "dm-")) {
                 _cleanup_(sd_device_enumerator_unrefp) sd_device_enumerator *e = NULL;
                 bool part_table_read = false, has_partitions = false;
-                const char *devname;
                 sd_device *d;
                 int fd;
-
-                r = sd_device_get_devname(dev, &devname);
-                if (r < 0)
-                        return r;
 
                 /* Try to re-read the partition table. This only succeeds if none of the devices is
                  * busy. The kernel returns 0 if no partition table is found, and we will not get an
                  * event for the disk. */
-                fd = open(devname, O_RDONLY|O_CLOEXEC|O_NOFOLLOW|O_NONBLOCK);
+                fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
                 if (fd >= 0) {
                         r = flock(fd, LOCK_EX|LOCK_NB);
                         if (r >= 0)
--- a/src/udev/udev-builtin-blkid.c
+++ b/src/udev/udev-builtin-blkid.c
@@ -302,11 +302,12 @@ static int builtin_blkid(sd_device *dev, sd_netlink **rtnl, int argc, char *argv
         if (r < 0)
                 return log_device_debug_errno(dev, r, "Failed to get device name: %m");
 
-        fd = open(devnode, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
+        fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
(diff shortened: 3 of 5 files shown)


yuwata:
TODO:
- support syspath
- check initialized, and add --initialized=BOOL
- add --removed

bluca (approved):

yuwata:
> TODO:
> 
>     * support syspath
> 
>     * check initialized, and add --initialized=BOOL
> 
>     * add --removed

done.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -419,6 +419,31 @@ _public_ int sd_device_new_from_stat_rdev(sd_device **ret, const struct stat *st
...
+        if (!path_startswith(devname, "/dev"))
+                return -EINVAL;
+
+        if (stat(devname, &st) < 0)
+                return errno == ENOENT ? -ENODEV : -errno;
+
+        return sd_device_new_from_stat_rdev(ret, &st);
+}
+
+_public_ int sd_device_new_from_syspath_or_devname(sd_device **ret, const char *path) {
```
poettering: maybe call this `sd_device_new_from_path()`? that leaves the door open that we can add more kinds of paths? dunno?

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,421 @@
...
+        STRV_FOREACH(p, c->devices) {
+                path_simplify(*p);
+
+                if (!path_is_safe(*p))
+                        return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                               "Device path cannot contain \"..\".");
+
+                if (!is_device_path(*p))
+                        return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                               "Specified path \"%s\" does not start with \"/dev\" or \"/sys\".", *p);
```
poettering: `/dev` → `/dev/` ?

`/sys` → `/sys/` ?

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,421 @@
...
+
+        r = sd_event_source_set_description(s, "inotify-event-source");
+        if (r < 0)
+                return r;
+
+        r = sd_event_source_set_floating(s, true);
+        if (r < 0)
+                return r;
+
+        c->inotify_fd = TAKE_FD(fd);
```
poettering: note that there is sd_event_add_inotify()?

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,421 @@
...
+                _cleanup_free_ char *p = NULL;
+
+                if (only_syspath)
+                        return 0;
+
+                r = path_extract_directory(path, &p);
+                if (r < 0)
+                        return r;
+
+                if (inotify_add_watch(c->inotify_fd, p, IN_CREATE | IN_DELETE | IN_MOVED_TO) < 0)
```
poettering: uh oh, please don't use inotify for this. Use our sd_device_monitor_xyz stuff to watch until a device shows up.

inotify is not useful here, as it dosn't tell us that the device node is ready, it just tells us that it exists. i.e. access mode, acls definitely won't be set up when we get the inotify event, and maybe some udev rules callouts should be called first too.

poettering (changes requested):
so, the concept is great, but i am very sure we shouldn#t use inotify for this, but use proper sd-event apis, i.e. the device monitor.

i.e. more along the lines of device_wait_for_devlink().

i.e. it is crucial for this tool to be useful that the complete udev ruleset ran for the device, i.e. we should only report ready once the udev db exists.

src/udev/udevadm-wait.c
yuwata: Indeed. I have no idea why I did not use sd-device-monitor. Thanks.

yuwata:
@poettering Thank you for the review. All comments are addressed. PTAL.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -415,6 +415,36 @@ _public_ int sd_device_new_from_stat_rdev(sd_device **ret, const struct stat *st
...
+        assert_return(path, -EINVAL);
+
+        if (path_startswith(path, "/dev"))
+                return sd_device_new_from_devname(ret, path);
+
+        if (path_startswith(path, "/sys"))
+                return sd_device_new_from_syspath(ret, path);
+
+        /* Assume that the path is without "/sys" prefix, like the path obtained by sd_device_get_devpath(). */
+        path = strjoina("/sys", path);
```
poettering: hmm, so I am not conviced by this fallback, since it makes it a lot harder to extend this function with different kinds of paths.

I mean, I think the fallback makes sense in a user-facing tool, because users want to be quicky and sloppy, but it does create some conceptual ambiguity as /dev/block/7:0 could in theory mean two things with such a fallback: actually /dev/block/7:0, the device node, or /sys/dev/block/7:0 which also exists! (really, it does, it's loop0). The ambiguity is only suppressed because this code prefers understanding of /dev over /sys, but I think that's still ugly.

So I'd just remove the ambiguity as part of the sd-device. i.e. if the path starts neither with /dev/ nor with /sys/ I'd just return an error, and thus avoid any ambiguity, and keeping things perfectly extensible if we should ever need to add more stuff here...

(and this should have been path_join(), since otherwise passing "xyz" as argument would have created a path "/sysxyz" here, which definitely is not desirable...)

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,298 @@
...
+STATIC_DESTRUCTOR_REGISTER(arg_devices, strv_freep);
+
+static int check_device(const char *path) {
+        _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
+        int r;
+
+        assert(path);
+
+        if (access(path, F_OK) < 0)
+                return arg_removed ? true : -errno;
```
poettering: i think this check should be dropped. this check after all only checks if the device node exists, but as mentioned that's not enough as the ownership/mode might not yet be in order. we really want to check that the db entry exists, i.e. sd_device_new_from_path() must succeed.

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,298 @@
...
+
+        if (arg_removed)
+                return false;
+
+        if (!arg_initialized)
+                return true;
+
+        r = sd_device_new_from_path(&dev, path);
```
poettering, line 51: this returns ENODEV if the device doesn't exist. We should return false in that case.

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,298 @@
...
+                                                  arg_removed ? "is removed" : arg_initialized ? "is initialized" : "exists",
+                                                  arg_removed ? "not removed" : arg_initialized ? "not initialized" : "not exist");
+                        return false;
+                }
+        }
+
+        return true;
+}
+
+static int device_monitor_handler(sd_device_monitor *monitor, sd_device *device, void *userdata) {
```
poettering: hmm, i think we should probably use the data in sd_device a bit before checking in /sys. i.e. add a helper

```c
static int device_matches_path(sd_device *d);
```

or so that goes through all relevant paths in the supplied device object, i.e. the syspath, the devpaths and the symlinks. And only if at least one of them matches against `arg_devices` we should go for check().

I am a bit afraid that we'll get a ton of events here, and if we never bother with the event data we'll always end up accessing /dev/ and/or /sys/ to verify the state. I think it would be much better if we would instead try to already do superficial verification via the data we got sent anyway, and only if it that looks good, check fully.

(in fact, I'd go even further: extend the `check()` routine, so that it gets an additional, optional `sd_device *focus` parameter or so, that we will initialize from the monitor callback device object. `check()` should skip ny further verification of all device paths that match this object as we *know* the device is already present if we got an event for it.)

poettering (changes requested):

yuwata:
@poettering Thank you for the review. Hopefully, all comments are addressed. PTAL.

yuwata:
The CI failures are by the tests for homed or dissect, should not be related.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -2222,3 +2222,62 @@ _public_ int sd_device_trigger_with_uuid(
...
+        dev_t devnum;
+        uint64_t diskseq = 0;
+        struct stat st;
+        int r;
+
+        assert_return(device, -EINVAL);
+
+        r = sd_device_get_devname(device, &devname);
+        if (r == -ENOENT)
+                return -EOPNOTSUPP;
```
poettering: hmm, EOPNOTSUPP to me is really about something that could be implemented but currently is not. Here however we have a device that simply is not something we could open at all, hence maybe ENOEXEC maybe?

src/libsystemd/sd-device/sd-device.c
```diff
@@ -2222,3 +2222,62 @@ _public_ int sd_device_trigger_with_uuid(
...
+
+        fd = open(devname, O_CLOEXEC|O_NOFOLLOW|O_PATH);
+        if (fd < 0)
+                return -errno;
+
+        if (fstat(fd, &st) < 0)
+                return -errno;
+
+        if (st.st_rdev != devnum)
+                return -ENXIO;
```
poettering, line 2280: we should also check if sd_event structure and st.st_mode agree on whether it's a block or char device.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -2222,3 +2222,62 @@ _public_ int sd_device_trigger_with_uuid(
...
+
+        if (diskseq != 0) {
+                _cleanup_close_ int fd3 = -1;
+                uint64_t seq;
+
+                fd3 = open(FORMAT_PROC_FD_PATH(fd), O_CLOEXEC|O_NOCTTY|O_NONBLOCK|O_RDONLY);
```
poettering: i think we should not open the device more than once. There are weird devices out there, i.e. cdroms where the tray is locked and things. Hence, only open the device node once with O_PATH and once without, and use the user chosen flags for the latter, right-away, and then invoke the diskseq stuff on that.

poettering: and this also means: if we are told to open the device with O_PATH we simply can't verify the diskseq and hence shouldn't

poettering: (which is of course somethng to document. in a comment or so for now I guess)

src/udev/udev-node.c
```diff
@@ -636,7 +636,7 @@ int udev_node_apply_permissions(
         if (fstat(node_fd, &stats) < 0)
                 return log_device_debug_errno(dev, errno, "cannot stat() node %s: %m", devnode);
 
-        if ((mode != MODE_INVALID && (stats.st_mode & S_IFMT) != (mode & S_IFMT)) || stats.st_rdev != devnum) {
+        if (mode != MODE_INVALID && (stats.st_mode & S_IFMT) != (mode & S_IFMT)) {
                 log_device_debug(dev, "Found node '%s' with non-matching devnum %s, skipping handling.",
```
poettering: as mentioned earlier, sd_device_open() should also verify whether we are talking about a block or char device here, and hence this whole if check can then go away.

src/udev/udevd.c
```diff
@@ -451,6 +451,8 @@ static int device_get_block_device(sd_device *dev, const char **ret) {
                 return log_device_debug_errno(dev, r, "Failed to get devtype: %m");
         if (r >= 0 && streq(val, "partition")) {
                 r = sd_device_get_parent(dev, &dev);
+                if (r == -ENODEV) /* The device may be already removed. */
```
poettering: looking at the code of sd_device_get_parent() now it is impossible it returns ENODEV, no?

some ENODEV/ENOENT confusion here?

src/libsystemd/sd-device/sd-device.c
```diff
@@ -415,6 +415,31 @@ _public_ int sd_device_new_from_stat_rdev(sd_device **ret, const struct stat *st
         return sd_device_new_from_devnum(ret, type, st->st_rdev);
 }
 
+_public_ int sd_device_new_from_devname(sd_device **ret, const char *devname) {
+        struct stat st;
+
+        assert_return(ret, -EINVAL);
+        assert_return(devname, -EINVAL);
+
+        if (!path_startswith(devname, "/dev"))
```
poettering: hmm, we probably shoudl check if the part after /dev is non-empty, btw. i.e. change to:

```c
if (isempty(path_startswith(devname, "/dev"))
        return -EINVAL;
```

src/libsystemd/sd-device/sd-device.c
```diff
@@ -415,6 +415,31 @@ _public_ int sd_device_new_from_stat_rdev(sd_device **ret, const struct stat *st
...
+                return ERRNO_IS_DEVICE_ABSENT(errno) ? -ENODEV : -errno;
+
+        return sd_device_new_from_stat_rdev(ret, &st);
+}
+
+_public_ int sd_device_new_from_path(sd_device **ret, const char *path) {
+        assert_return(ret, -EINVAL);
+        assert_return(path, -EINVAL);
+
+        if (path_startswith(path, "/dev"))
```
poettering, line 452: similar

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,363 @@
...
+}
+
+static bool check(void) {
+        int r;
+
+        if (arg_settle) {
+                r = udev_queue_is_empty();
+                if (r <= 0) {
+                        if (r < 0)
+                                log_warning_errno(r, "Failed to check if udev queue is empty, assuming not empty: %m");
```
poettering: if we can't check the queue I'd actually assume its empty

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,363 @@
...
+                        return false;
+                }
+        }
+
+        STRV_FOREACH(p, arg_devices) {
+                r = check_device(*p);
+                if (r <= 0) {
+                        if (r < 0)
+                                log_warning_errno(r, "Failed to check if device \"%s\" %s, assuming %s.: %m",
```
poettering: `.:` → `:`

src/udev/udevadm-wait.c
```diff
@@ -0,0 +1,363 @@
...
+#include "parse-util.h"
+#include "path-util.h"
+#include "static-destruct.h"
+#include "strv.h"
+#include "udev-util.h"
+#include "udevadm.h"
+
+static usec_t arg_timeout_usec = USEC_INFINITY;
+static bool arg_initialized = true;
+static bool arg_removed = false;
```
poettering: these two should be a single enum, no? i.e.

```c
enum {
        ARG_WAIT_UNTIL_ADDED,
        ARG_WAIT_UNTIL_INITIALIZED,
        ARG_WAIT_UNTIL_REMOVED,
} arg_wait_until = ARG_INITIALIZED;
```

no? and then maybe also expose it like this in the cmdline, i.e. `udevadm wait --until=added /dev/sda1` or so?

test/units/testsuite-64.sh
```diff
@@ -658,24 +625,16 @@ testcase_iscsi_lvm() {
...
-    for link in "${expected_symlinks[@]}"; do
-        test ! -e "$link"
-    done
+    udevadm wait --settle --timeout=30 --removed "/dev/$vgroup" "/dev/disk/by-label/mylvpart1" "${expected_symlinks[@]}"
     helper_check_device_symlinks "/dev/disk"
     # Reconnect the iSCSI devices and check if everything get detected correctly
     iscsiadm --mode discoverydb --type sendtargets --portal "$target_ip" --discover
     iscsiadm --mode node --targetname "$target_name" --portal "$target_ip:$target_port" --login
-    udevadm settle
+    udevadm wait --settle --timeout=30 "${expected_symlinks[@]}"
```
poettering, line 634: do we really need `--settle` on these?

poettering (changes requested):
looks really good

src/libsystemd/sd-device/sd-device.c
poettering: i think this effectively should check the subsystem. If returned as "block", then S_ISBLK(st.st_mode) must hold, and in all other cases S_ISCHR()

poettering:
Another candidate for sd_device_open() is src/gpt-auto-generator/gpt-auto-generator.c's open_parent_block_device()...

test/units/testsuite-64.sh
yuwata: Yes. Not only 'add' but 'change' event may be already queued for the devices.

src/libsystemd/sd-device/sd-device.c
yuwata: Here, 'isempty()' is not necessary when `isempty()` is checked in 'sd_device_new_from_devname()'.

yuwata:
@poettering Thank you for the review. Most comments are addressed. PTAL.

> Another candidate for sd_device_open() is src/gpt-auto-generator/gpt-auto-generator.c's open_parent_block_device()...

Let's update it in later PR.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -415,6 +415,31 @@ _public_ int sd_device_new_from_stat_rdev(sd_device **ret, const struct stat *st
         return sd_device_new_from_devnum(ret, type, st->st_rdev);
 }
 
+_public_ int sd_device_new_from_devname(sd_device **ret, const char *devname) {
```
poettering, line 418: strictly speaking this will also accept symlinks, i.e. "devlink"s in udev nomenclature. But I think this fact doesn't have to show up in the function name, so i think we are ok here.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -415,6 +415,31 @@ _public_ int sd_device_new_from_stat_rdev(sd_device **ret, const struct stat *st
...
+        struct stat st;
+
+        assert_return(ret, -EINVAL);
+        assert_return(devname, -EINVAL);
+
+        if (isempty(path_startswith(devname, "/dev")))
+                return -EINVAL;
```
poettering, line 429: So, I think we should tweak this one tiny bit more. i.e. we should use device_path_parse_major_minor() here, i.e. try to parse the major/minor directly from the path if the path is a /dev/block/maj:min path. 

Why bother? because it means sd_device_new_from_devname() actually can work for these paths even if udev isn't running, and hence no /dev/block/ + /dev/char/ symlink created yet.

poettering (approved):
lgtm

src/libsystemd/sd-device/sd-device.c
poettering: but maybe add a comment to the function that mentions this, i.e.

```c
/* this function actually accepts both devlinks and devnames, i.e. both symlinks and actualy device nodes below /dev/ */
```

yuwata:
@poettering Thank you. The above two comments are addressed. Upgrading the green label.

PR #22877, opened 2022-03-27
core: add [Un]SubscribeWithFlags method and JobRemovedEx signal

### LINKED ISSUE #21520: RFE: allow discerning a job that completes successfully because job is conditioned out from one that completes actually successfully
 > 249

**Used distribution**
 > openSUSE Tumbleweed

 > `Linux localhost.localdomain 5.15.3-1-default #1 SMP Fri Nov 19 08:35:39 UTC 2021 (404f576) x86_64 x86_64 x86_64 GNU/Linux`

 > x86_64

The `org.bluez` service on the system DBus is configured to use systemd activation to start `dbus-org.bluez.service`.

That unit is provided by `bluetooth.service`:

```
[Unit]
Description=Bluetooth service
Documentation=man:bluetoothd(8)
ConditionPathIsDirectory=/sys/class/bluetooth

[Service]
Type=dbus
BusName=org.bluez
ExecStart=/usr/libexec/bluetooth/bluetoothd
NotifyAccess=main
#WatchdogSec=10
#Restart=on-failure
CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_BIND_SERVICE
LimitNPROC=1
ProtectHome=true
ProtectSystem=full

[Install]
WantedBy=bluetooth.target
Alias=dbus-org.bluez.service
```

I'm running this in a VM, so this unit will not be started because `/sys/class/bluetooth` does not exist.

When anything does a DBus call to `org.bluez` and thus triggers activation, it never gets a reply because the activation timed out.
This is because `dbus-daemon` waits for either the activated service to appear on the bus or an `ActivationFailure` signal from systemd. However, on condition failure, neither happens. systemd should probably send an `ActivationFailure` back.

```
linux@localhost:~> dbus-send --system --type=method_call --dest=org.bluez --print-reply /asdf as.df.asdf
Error org.freedesktop.DBus.Error.TimedOut: Failed to activate service 'org.bluez': timed out (service_start_timeout=25000ms)
```

`dbus-monitor --system` shows:
```
signal time=1637772534.319864 sender=org.freedesktop.DBus -> destination=org.freedesktop.systemd1 serial=443 path=/org/freedesktop/DBus; interface=org.freedesktop.systemd1.Activator; member=ActivationRequest
   string "dbus-org.bluez.service"
...
signal time=1637772534.322047 sender=:1.0 -> destination=(null destination) serial=11394 path=/org/freedesktop/systemd1; interface=org.freedesktop.systemd1.Manager; member=JobNew
   uint32 4433
   object path "/org/freedesktop/systemd1/job/4433"
   string "bluetooth.service"
...
signal time=1637772534.322982 sender=:1.0 -> destination=(null destination) serial=11397 path=/org/freedesktop/systemd1; interface=org.freedesktop.systemd1.Manager; member=JobRemoved
   uint32 4433
   object path "/org/freedesktop/systemd1/job/4433"
   string "bluetooth.service"
   string "done"
...
signal time=1637772534.323592 sender=:1.0 -> destination=(null destination) serial=11399 path=/org/freedesktop/systemd1/unit/bluetooth_2eservice; interface=org.freedesktop.DBus.Properties; member=PropertiesChanged
   string "org.freedesktop.systemd1.Unit"
   array [
      dict entry(
         string "ActiveState"
         variant             string "inactive"
      )
      dict entry(
         string "FreezerState"
         variant             string "running"
      )
      dict entry(
         string "SubState"
         variant             string "dead"
      )
...
error time=1637772559.343660 sender=org.freedesktop.DBus -> destination=:1.308 error_name=org.freedesktop.DBus.Error.TimedOut reply_serial=2
   string "Failed to activate service 'org.bluez': timed out (service_start_timeout=25000ms)"
```

```
linux@localhost:~> systemctl status dbus-org.bluez.service
○ bluetooth.service - Bluetooth service
     Loaded: loaded (/usr/lib/systemd/system/bluetooth.service; enabled; vendor preset: disabled)
     Active: inactive (dead)
       Docs: man:bluetoothd(8)
```

The journal shows:
```
Nov 24 11:48:54 localhost.localdomain systemd[1]: Got message type=signal sender=org.freedesktop.DBus destination=org.freedesktop.systemd1 path=/org/freedesktop/DBus interface=org.freedesktop.systemd1.Activator member=ActivationRequest cookie=443 reply_cookie=0 signature=s error-name=n/a error-message=n/a
Nov 24 11:48:54 localhost.localdomain systemd[1]: Unit dbus-org.bluez.service has alias bluetooth.service.
Nov 24 11:48:54 localhost.localdomain systemd[1]: Sent message type=method_call sender=n/a destination=org.freedesktop.DBus path=/org/freedesktop/DBus interface=org.freedesktop.DBus member=AddMatch cookie=11391 reply_cookie=0 signature=s error-name=n/a error-message=n/a
Nov 24 11:48:54 localhost.localdomain systemd[1]: Sent message type=method_call sender=n/a destination=org.freedesktop.DBus path=/org/freedesktop/DBus interface=org.freedesktop.DBus member=GetNameOwner cookie=11392 reply_cookie=0 signature=s error-name=n/a error-message=n/a
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: Watching D-Bus name 'org.bluez'.
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: Trying to enqueue job bluetooth.service/start/replace
Nov 24 11:48:54 localhost.localdomain systemd[1]: systemd-journald-audit.socket: Cannot add dependency job, ignoring: Unit systemd-journald-audit.socket not found.
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: Installed new job bluetooth.service/start as 4433
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: Enqueued job bluetooth.service/start as 4433
Nov 24 11:48:54 localhost.localdomain systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=UnitNew cookie=11393 reply_cookie=0 signature=so error-name=n/a error-message=n/a
Nov 24 11:48:54 localhost.localdomain systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=JobNew cookie=11394 reply_cookie=0 signature=uos error-name=n/a error-message=n/a
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: ConditionPathIsDirectory=/sys/class/bluetooth failed.
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: Starting requested but condition failed. Not starting unit.
Nov 24 11:48:54 localhost.localdomain systemd[1]: bluetooth.service: Job 4433 bluetooth.service/start finished, result=done
Nov 24 11:48:54 localhost.localdomain systemd[1]: Condition check resulted in Bluetooth service being skipped.
Nov 24 11:48:54 localhost.localdomain systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1/unit/bluetooth_2eservice interface=org.freedesktop.DBus.Properties member=PropertiesChanged cookie=11395 reply_cookie=0 signature=sa{sv}as error-name=n/a error-message=n/a
```

bluca:
The new signal is always sent, and it returns 'skipped' when Condition*= fail.

Always set the job result to 'skipped' when a condition check fails, but
fix it back to 'done' when sending the old JobRemoved signal. Let the new
JobRemoved2 signal return the true result.

The old signal is only sent if the old Subscribe method has been
called by at least one client, otherwise only the new signal is used.

An environment variable is provided to force always sending also the
legacy signal, even when nobody called Subscribe().

Fixes #21520

### DIFF
--- a/src/login/logind-dbus.c
+++ b/src/login/logind-dbus.c
@@ -3761,7 +3761,7 @@ static int session_jobs_reply(Session *s, uint32_t jid, const char *unit, const 
         if (!s->started)
                 return 0;
 
-        if (result && !streq(result, "done")) {
+        if (result && !STR_IN_SET(result, "done", "skipped")) {
                 _cleanup_(sd_bus_error_free) sd_bus_error e = SD_BUS_ERROR_NULL;
 
                 sd_bus_error_setf(&e, BUS_ERROR_JOB_FAILED,
@@ -3787,6 +3787,12 @@ int match_job_removed(sd_bus_message *message, void *userdata, sd_bus_error *err
         if (r < 0) {
                 bus_log_parse_error(r);
                 return 0;
+        }
+
+        if (sd_bus_message_is_signal(message, NULL, "JobRemovedEx")) {
+                r = sd_bus_message_skip(message, "ay");
+                if (r < 0)
+                        return bus_log_parse_error(r);
         }
 
         if (m->action_job && streq(m->action_job, path)) {
--- a/src/shared/bus-wait-for-jobs.c
+++ b/src/shared/bus-wait-for-jobs.c
@@ -5,6 +5,7 @@
 #include "set.h"
 #include "bus-util.h"
 #include "bus-internal.h"
+#include "bus-locator.h"
 #include "unit-def.h"
 #include "escape.h"
 #include "strv.h"
@@ -48,6 +49,12 @@ static int match_job_removed(sd_bus_message *m, void *userdata, sd_bus_error *er
                 return 0;
         }
 
+        if (sd_bus_message_is_signal(m, NULL, "JobRemovedEx")) {
+                r = sd_bus_message_skip(m, "ay");
+                if (r < 0)
+                        return bus_log_parse_error(r);
+        }
+
         found = set_remove(d->jobs, (char*) path);
         if (!found)
                 return 0;
@@ -78,12 +85,37 @@ BusWaitForJobs* bus_wait_for_jobs_free(BusWaitForJobs *d) {
         return mfree(d);
 }
 
+int bus_subscribe_with_flags_cached(sd_bus *bus) {
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *reply = NULL;
+        static int cached = -1;
+        int r;
+
+        assert(bus);
+
+        if (cached >= 0)
+                return cached;
+
+        r = bus_call_method(bus,
+                        bus_systemd_mgr,
+                        "SubscribeWithFlags",
+                        &error, &reply, "t", 0);
+        if (r < 0 && !sd_bus_error_has_name(&error, SD_BUS_ERROR_UNKNOWN_METHOD))
+                return log_error_errno(r, "Failed to call SubscribeWithFlags: %s", bus_error_message(&error, r));
+
+        return cached = r >= 0;
+}
+
 int bus_wait_for_jobs_new(sd_bus *bus, BusWaitForJobs **ret) {
         _cleanup_(bus_wait_for_jobs_freep) BusWaitForJobs *d = NULL;
+        bool new_signal_available = false;
         int r;
 
         assert(bus);
         assert(ret);
+
+        r = bus_subscribe_with_flags_cached(bus);
+        new_signal_available = r > 0;
 
         d = new(BusWaitForJobs, 1);
         if (!d)
@@ -102,7 +134,7 @@ int bus_wait_for_jobs_new(sd_bus *bus, BusWaitForJobs **ret) {
                         bus->bus_client ? "org.freedesktop.systemd1" : NULL,
                         "/org/freedesktop/systemd1",
                         "org.freedesktop.systemd1.Manager",
-                        "JobRemoved",
+                        new_signal_available ? "JobRemovedEx" : "JobRemoved",
                         match_job_removed, NULL, d);
         if (r < 0)
                 return r;
--- a/src/systemctl/systemctl-start-unit.c
+++ b/src/systemctl/systemctl-start-unit.c
@@ -348,7 +348,18 @@ int verb_start(int argc, char *argv[], void *userdata) {
         }
 
         if (arg_wait) {
-                r = bus_call_method_async(bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
+                bool new_signal_available = false;
+
+                r = bus_subscribe_with_flags_cached(bus);
+                if (r < 0)
+                        log_warning_errno(r, "Failed to determine whether JobRemovedEx signal is available, falling back to legacy JobRemoved: %m");
+                else
+                        new_signal_available = r > 0;
+
+                if (new_signal_available)
+                        r = bus_call_method_async(bus, NULL, bus_systemd_mgr, "SubscribeWithFlags", NULL, NULL, "t", 0);
+                else
+                        r = bus_call_method_async(bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
                 if (r < 0)
                         return log_error_errno(r, "Failed to enable subscription: %m");
 
--- a/src/login/logind.c
+++ b/src/login/logind.c
@@ -13,6 +13,7 @@
 #include "bus-locator.h"
 #include "bus-log-control-api.h"
 #include "bus-polkit.h"
+#include "bus-wait-for-jobs.h"
 #include "cgroup-util.h"
 #include "daemon-util.h"
 #include "def.h"
@@ -628,6 +629,7 @@ static int manager_reserve_vt(Manager *m) {
 }
 
 static int manager_connect_bus(Manager *m) {
+        bool new_signal_available = false;
         int r;
 
         assert(m);
@@ -637,6 +639,12 @@ static int manager_connect_bus(Manager *m) {
         if (r < 0)
                 return log_error_errno(r, "Failed to connect to system bus: %m");
 
+        r = bus_subscribe_with_flags_cached(m->bus);
+        if (r < 0)
+                log_warning_errno(r, "Failed to determine whether JobRemovedEx signal is available, falling back to legacy JobRemoved: %m");
+        else
+                new_signal_available = r > 0;
+
         r = bus_add_implementation(m->bus, &manager_object, m);
         if (r < 0)
                 return r;
@@ -645,9 +653,9 @@ static int manager_connect_bus(Manager *m) {
         if (r < 0)
                 return r;
 
-        r = bus_match_signal_async(m->bus, NULL, bus_systemd_mgr, "JobRemoved", match_job_removed, NULL, m);
+        r = bus_match_signal_async(m->bus, NULL, bus_systemd_mgr, new_signal_available ? "JobRemovedEx" : "JobRemoved", match_job_removed, NULL, m);
         if (r < 0)
-                return log_error_errno(r, "Failed to request match for JobRemoved: %m");
+                return log_error_errno(r, "Failed to request match for JobRemoved%s: %m", new_signal_available ? "Ex": "");
 
         r = bus_match_signal_async(m->bus, NULL, bus_systemd_mgr, "UnitRemoved", match_unit_removed, NULL, m);
         if (r < 0)
@@ -668,7 +676,10 @@ static int manager_connect_bus(Manager *m) {
         if (r < 0)
                 return log_error_errno(r, "Failed to request match for Reloading: %m");
 
-        r = bus_call_method_async(m->bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
+        if (new_signal_available)
+                r = bus_call_method_async(m->bus, NULL, bus_systemd_mgr, "SubscribeWithFlags", NULL, NULL, "t", 0);
+        else
+                r = bus_call_method_async(m->bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
         if (r < 0)
                 return log_error_errno(r, "Failed to enable subscription: %m");
 
--- a/src/machine/machined-dbus.c
+++ b/src/machine/machined-dbus.c
@@ -1230,6 +1230,12 @@ int match_job_removed(sd_bus_message *message, void *userdata, sd_bus_error *err
                 return 0;
         }
 
+        if (sd_bus_message_is_signal(message, NULL, "JobRemovedEx")) {
+                r = sd_bus_message_skip(message, "ay");
+                if (r < 0)
+                        return bus_log_parse_error(r);
+        }
+
         machine = hashmap_get(m->machine_units, unit);
         if (!machine)
                 return 0;
@@ -1238,7 +1244,7 @@ int match_job_removed(sd_bus_message *message, void *userdata, sd_bus_error *err
                 machine->scope_job = mfree(machine->scope_job);
 
                 if (machine->started) {
-                        if (streq(result, "done"))
+                        if (STR_IN_SET(result, "done", "skipped"))
                                 machine_send_create_reply(machine, NULL);
                         else {
                                 _cleanup_(sd_bus_error_free) sd_bus_error e = SD_BUS_ERROR_NULL;
--- a/src/machine/machined.c
+++ b/src/machine/machined.c
@@ -11,6 +11,7 @@
 #include "bus-locator.h"
 #include "bus-log-control-api.h"
 #include "bus-polkit.h"
+#include "bus-wait-for-jobs.h"
 #include "cgroup-util.h"
 #include "daemon-util.h"
 #include "dirent-util.h"
@@ -184,6 +185,7 @@ static int manager_enumerate_machines(Manager *m) {
 }
 
 static int manager_connect_bus(Manager *m) {
+        bool new_signal_available = false;
         int r;
 
         assert(m);
@@ -193,13 +195,19 @@ static int manager_connect_bus(Manager *m) {
         if (r < 0)
                 return log_error_errno(r, "Failed to connect to system bus: %m");
 
+        r = bus_subscribe_with_flags_cached(m->bus);
+        if (r < 0)
+                log_warning_errno(r, "Failed to determine whether JobRemovedEx signal is available, falling back to legacy JobRemoved: %m");
+        else
+                new_signal_available = r > 0;
+
         r = bus_add_implementation(m->bus, &manager_object, m);
         if (r < 0)
                 return r;
 
-        r = bus_match_signal_async(m->bus, NULL, bus_systemd_mgr, "JobRemoved", match_job_removed, NULL, m);
+        r = bus_match_signal_async(m->bus, NULL, bus_systemd_mgr, new_signal_available ? "JobRemovedEx" : "JobRemoved", match_job_removed, NULL, m);
         if (r < 0)
-                return log_error_errno(r, "Failed to add match for JobRemoved: %m");
+                return log_error_errno(r, "Failed to add match for JobRemoved%s: %m", new_signal_available ? "Ex" : "");
 
         r = bus_match_signal_async(m->bus, NULL, bus_systemd_mgr, "UnitRemoved", match_unit_removed, NULL, m);
         if (r < 0)
@@ -220,7 +228,10 @@ static int manager_connect_bus(Manager *m) {
         if (r < 0)
                 return log_error_errno(r, "Failed to request match for Reloading: %m");
 
-        r = bus_call_method_async(m->bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
+        if (new_signal_available)
+                r = bus_call_method_async(m->bus, NULL, bus_systemd_mgr, "SubscribeWithFlags", NULL, NULL, "t", 0);
+        else
+                r = bus_call_method_async(m->bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
         if (r < 0)
                 return log_error_errno(r, "Failed to enable subscription: %m");
 
--- a/src/shared/bus-wait-for-jobs.h
+++ b/src/shared/bus-wait-for-jobs.h
@@ -12,5 +12,7 @@ BusWaitForJobs* bus_wait_for_jobs_free(BusWaitForJobs *d);
 int bus_wait_for_jobs_add(BusWaitForJobs *d, const char *path);
 int bus_wait_for_jobs(BusWaitForJobs *d, bool quiet, const char* const* extra_args);
 int bus_wait_for_jobs_one(BusWaitForJobs *d, const char *path, bool quiet, const char* const* extra_args);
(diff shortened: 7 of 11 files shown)


bluca:
Ideas for better names than `JobRemoved2`? `JobRemovedNew`? Not that much better... `JobRemovedFull`? `JobRemovedShouldHaveAlwaysBeenLikeThat`?

src/shared/bus-wait-for-jobs.c
```diff
@@ -78,13 +79,72 @@ BusWaitForJobs* bus_wait_for_jobs_free(BusWaitForJobs *d) {
...
+
+        assert(name);
+
+        if (streq(name, "JobRemoved2"))
+                *ret_available = true;
+
+        return 0;
+}
+
+int bus_jobremoved2_signal_available(sd_bus *bus, bool *ret_available) {
```
yuwata: Why not cache the result in this function?

src/shared/bus-wait-for-jobs.c
```diff
@@ -78,13 +79,72 @@ BusWaitForJobs* bus_wait_for_jobs_free(BusWaitForJobs *d) {
...
+                        &reply,
+                        NULL);
+        if (r < 0)
+                return log_debug_errno(r, "Could not introspect /org/freedesktop/systemd1: %s", bus_error_message(&error, r));
+
+        r = sd_bus_message_read(reply, "s", &xml);
+        if (r < 0)
+                return bus_log_parse_error(r);
+
+        r = parse_xml_introspect("/org/freedesktop/systemd1", xml, &ops, ret_available);
```
yuwata: How about the following?
```
static int available = -1;
bool b = false;
...
r = parse_xml_introspect(..., &b);
if (r < 0)
        return r;

return available = b;
```

src/shared/bus-wait-for-jobs.c
bluca: I wanted the decision to the caller, it might not always be appropriate

yuwata: But, at least currently, all caller cache the result. At least, how about to introduce `bus_jobremoved2_signal_available_cached()`?

bluca: ok, I've changed to cache into the call

man/org.freedesktop.systemd1.xml
```diff
@@ -268,6 +270,10 @@ node /org/freedesktop/systemd1 {
                  o job,
                  s unit,
                  s result);
+      JobRemoved2(u id,
```
poettering: for various bus methods we are using the "Ex" suffix for newer extended flavours of existing calls. please follow the same naming scheme here.

src/systemctl/systemctl-start-unit.c
```diff
@@ -348,7 +348,18 @@ int verb_start(int argc, char *argv[], void *userdata) {
         }
 
         if (arg_wait) {
-                r = bus_call_method_async(bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
+                bool new_signal_available = false;
+
+                r = bus_jobremoved2_signal_available(bus);
```
poettering: uh, no. please don't use the crappy XML introspection for anything but debugging please.

Just issue SubscribeWithFlags() directly. If you get NoSuchMethod error back you know new-style JobRemovedEx signal is not available.

src/core/manager.h
```diff
@@ -277,6 +277,9 @@ struct Manager {
         and it is much simpler that way. */
         sd_bus_track *subscribed;
         char **deserialized_subscribed;
+        /* We'll send out legacy JobRemoved signals only if at least one subscriber asks for it by calling
+         * the legacy Subscribe() method. */
+        size_t n_legacy_subscribers;
```
poettering, line 282: please just maintain two sd_bus_track objects: one for new-style and one for old-style subscribers. And send the JobRemoved if the old-style one is populated and JobRemoevedEx if the new-style one is populated.

man/org.freedesktop.systemd1.xml
poettering: so if we fiddle with this, I think it would be good to take this one step further and add one more thing at the same time: the invocation ID associated with the unit at the moment the job completed/failed, u.e. the unit's invocation_id field. I think that would be a major useful addition, as it means you can issue a start job, get a job id back, then once the job is complete you get the invocation id and that you can then use to watch the service's iteration you triggered.

the addition would be trivial to add. just copy the id into the signal.

src/core/dbus-job.c
```diff
@@ -286,14 +287,45 @@ static int send_removed_signal(sd_bus *bus, void *userdata) {
...
+                        "/org/freedesktop/systemd1",
+                        "org.freedesktop.systemd1.Manager",
+                        "JobRemoved");
+        if (r < 0)
+                return r;
+
+        /* The old JobRemoved signal never returns 'skipped' as a result, unless the job was about a device
+         * unit, for backward-compatibility reasons. */
+        JobResult result;
+        if (j->result == JOB_SKIPPED && j->unit->type != UNIT_DEVICE)
```
poettering, line 324: hmm, where was that conditionalization on devices? i missed that?

poettering (changes requested):

src/systemctl/systemctl-start-unit.c
poettering: this means the Subscribe() call must come first, and then match installation to JobRemovedEx second, but that should not matter.

src/core/dbus-job.c
bluca: It always returned 'skipped' when the  unit doesn't have a `start()` method, which means device unit:

https://github.com/systemd/systemd/blob/ff7e7c2b3a2b4367e0a03a51b7d714dfe0837abb/src/core/unit.c#L1916
https://github.com/systemd/systemd/blob/ff7e7c2b3a2b4367e0a03a51b7d714dfe0837abb/src/core/unit.c#L1842
https://github.com/systemd/systemd/blob/ff7e7c2b3a2b4367e0a03a51b7d714dfe0837abb/src/core/job.c#L894

src/core/manager.h
bluca: It's quite _a lot_ of churn to duplicate everything, are we sure it's worth it? It's just checked in one place after all

src/systemctl/systemctl-start-unit.c
bluca: But there are some places, like in timedated, where it just watches for the signal, it doesn't subscribe. So the only thing I can do in that situation is instrospection. What's the issue with introspection? It's cached, so even if it's a bit slow, it only happens on first invocation.

poettering: it's actually a bug if we don't subscribe. We really should. 

maybe add a helper for all this. i.e. bus_subscribe_job_removed() or so, which will do the right thing (i.e. subscribe + install match), and then change all code to just call that.

src/core/manager.h
poettering: there usually should not be duplicate entries. i.e. old clients in one, and new clients in the other. so you got the same number of entries either way. All you pay is that you allocate the track object twice. But that's cheap... in particular as we lazy-allocate the sd_bus_track object iirc, and thus if you have a fully new-style system the old object will never be allocated.

bluca: I'm not worried about the runtime cost - by churn I meant code churn. Everything needs to be duplicated, and handled in tandem everywhere, save for one place - where the signal is sent. It's quite a lot of duplication.

src/core/dbus-job.c
poettering: uh, oh, so "skipped" actually means something entirely different than "condition failed". It means "you asked me for an operation i can not execute for you on this unit type, but all is good, so i skipped it".

we need a new JobResult for indicating skipped conditions. "conditioned" or so?

poettering: Ah, fuck, ExecCondition= has already squatted on this result, too. what a mess... see job.h, the JobResult definition.

we need to clean that up.


i.e. so what about this. We introduce:

JOB_DONE_CONDITION + JOB_DONE_EXEC_CONDITION or so as new "special flavours" of JOB_DONE. That way clients can look for "success conditions" via `streq(result, "done") || startswith(result, "done-")`, in a generic fashion.

Then in the compat JobRemoved signal mangle JOB_DONE_EXEC_CONDITION into JOB_SKIPPED and
JOB_DONE_CONDITION into JOB_DONE. And in the new JobRemovedEx we'd instead pass them down in their proper names.

With that in place things  should be reasonably clean again and one result will mean exactly one thing again.

src/core/manager.h
poettering: "everything"? it's not that much, is it?

src/core/dbus-job.c
bluca: Aside from that "on this unit type", which is an implementation detail, to me it seems a description that fits pretty well for this case though - the operation could not be executed, but the reason it could not be executed is within the parameters set by the system configuration, so it's all good, no?

It seems to fit with the original description of the job result even when it was introduced: https://github.com/systemd/systemd/commit/d68201e9aa5e9ebd6085b1bb8892c42e9d20be75

src/systemctl/systemctl-start-unit.c
bluca: Two more issues with that:

1) in all places where it happens, the subscribe is done as the last thing, after all signals match are installed, I assume to avoid races - this would have to be swapped
2) in all places where it happens, the subscribe is called async, so there's no immediate feedback on whether the method exists or not, these would have to all become sync, blocking calls

bluca: implemented this - somewhat - as usage varies too wildly

man/org.freedesktop.systemd1.xml
bluca: makes sense, added and renamed

src/core/dbus-job.c
bluca: > JOB_DONE_CONDITION + JOB_DONE_EXEC_CONDITION or so as new "special flavours" of JOB_DONE. That way clients can look for "success conditions" via `streq(result, "done") || startswith(result, "done-")`, in a generic fashion.

Why are different return values for Condition/ExecCondition needed? The latter is just a free-form of the former, no? In which situation should a caller do something different if a condition caused a skip vs if an execcondition caused a skip?

src/core/manager.h
bluca: Everything to do with this queue, including serializing/deserializing, etc

docs/ENVIRONMENT.md
```diff
@@ -260,6 +260,9 @@ All tools:
   it is either set to `system` or `user` depending on whether the NSS/PAM
   module is called by systemd in `--system` or `--user` mode.
 
+* `$SYSTEMD_FORCE_LEGACY_JOB_REMOVED_SIGNAL` — always send the legacy `JobRemoved`
+  signal, even if all clients called `SubscribeWithFlags` instead of `Subscribe`.
```
keszybz, line 273: What happens in no one called `SubscribeWithFlags`? Shouldn't this say "always send …, even if nobody subscribed to it"? (I don't know, I'm just starting to read the patch.)

man/org.freedesktop.systemd1.xml
```diff
@@ -1303,13 +1316,18 @@ node /org/freedesktop/systemd1 {
...
       interested in signals need to call this method. Signals are only sent out if at least one client
-      invoked this method. <function>Unsubscribe()</function> reverts the signal subscription that
-      <function>Subscribe()</function> implements. It is not necessary to invoke
-      <function>Unsubscribe()</function> as clients are tracked. Signals are no longer sent out as soon as
-      all clients which previously asked for <function>Subscribe()</function> either closed their connection
-      to the bus or invoked <function>Unsubscribe()</function>.</para>
+      invoked this method. <function>UnsubscribeWithFlags()</function> reverts the signal subscription that
+      <function>SubscribeWithFlags()</function> implements. It is not necessary to invoke
+      <function>UnsubscribeWithFlags()</function> as clients are tracked. Signals are no longer sent out as
+      soon as all clients which previously asked for <function>SubscribeWithFlags()</function> either closed
```
keszybz: Preexisting condition, but not "asked for SubscribeWithFlags()", but just "called SubscribeWithFlags()".

man/org.freedesktop.systemd1.xml
```diff
@@ -1303,13 +1316,18 @@ node /org/freedesktop/systemd1 {
...
+      their connection to the bus or invoked <function>UnsubscribeWithFlags()</function>. Calling the legacy
+      <function>Subscribe()</function> causes the legacy <function>JobRemoved()</function> signal to be
```
keszybz: singal to be sent → signals to be sent

JobRemoved() → JobRemoved

man/org.freedesktop.systemd1.xml
```diff
@@ -1303,13 +1316,18 @@ node /org/freedesktop/systemd1 {
...
+      sent, in addition to the new <function>JobRemovedEx</function> which is always sent if there are any
+      kind of subscribers. The legacy <function>Unsubscribe()</function> works for
+      <function>Subscribe()</function> as <function>UnsubscribeWithFlags()</function> works for
+      <function>SubscribeWithFlags()</function>.</para>
```
keszybz: Hmm, is the flags argument documented anywhere?

src/core/dbus-job.c
keszybz: I agree. It sounds nicer to have the same result for both.

keszybz (changes requested):

docs/ENVIRONMENT.md
bluca: If no client calls any of the subscribes, no signal is sent - this is existing behavior, documented in the API page

src/shared/bus-wait-for-jobs.c
```diff
@@ -78,13 +85,40 @@ BusWaitForJobs* bus_wait_for_jobs_free(BusWaitForJobs *d) {
...
+
+        assert(bus);
+
+        if (cached >= 0)
+                return cached;
+
+        r = sd_bus_call_method(bus,
+                        "org.freedesktop.systemd1",
+                        "/org/freedesktop/systemd1",
+                        "org.freedesktop.systemd1.Manager",
```
yuwata: `bus-locator.h` is included, but not used. Here, you can use `bus_systemd_mgr`.

src/systemctl/systemctl-start-unit.c
```diff
@@ -348,7 +348,18 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
+                r = bus_subscribe_with_flags_cached(bus);
+                if (r < 0)
+                        log_warning_errno(r, "Failed to determine whether JobRemovedEx signal is available, falling back to legacy JobRemoved: %m");
+                else
+                        new_signal_available = r > 0;
+
+                if (new_signal_available)
+                        r = bus_call_method_async(bus, NULL, bus_systemd_mgr, "SubscribeWithFlags", NULL, NULL, "t", 0);
+                else
+                        r = bus_call_method_async(bus, NULL, bus_systemd_mgr, "Subscribe", NULL, NULL, NULL);
```
yuwata, line 362: If PID1 has `SubscribeWithFlags` method, then the method is called twice.

Maybe, `bus_subscribe_with_flags_cached()` always call `SubscribeWithFlags` or `Subscribe`.
And, also make them called with async.

yuwata (changes requested):

src/systemctl/systemctl-start-unit.c
bluca: It cannot be async, as we need to wait for the result before we know which one to call. It won't hurt to queue the second async call, it will just be a no-op.

yuwata: I mean, this is still incomplete, but I think we should do something like that: https://github.com/yuwata/systemd/commit/5fcb45500f00e2d35a22a4a557bbb0473322fb67.

Currently, all `Subscribe` method calls and `JobRemoved` signal match installations are done in asynchronously. I think we should use `bus_wait_for_jobs_new()` everywhere, and it handles Ex and non-Ex gracefully.

yuwata:
BTW, I'd like this to be split into at least 3 commits:
1. drop the unused flag,
2. introduce new method and signal,
3. use the new method and signal in services or tools.

src/core/dbus-job.c
poettering: > Why are different return values for Condition/ExecCondition needed? The latter is just a free-form of the former, no? In which situation should a caller do something different if a condition caused a skip vs if an execcondition caused a skip?

I think they are quite different, because the exec one actually executed code, had a blocking runtime and so on. 

But the main reason is because you want to mangle them differently later on for the legacy JobRemoved. You can't do that if you don't distinguish them. i.e. one should be mangled into JOB_SKIPPED and the other into JOB_DONE, see above. Hence, if we have to internally distinguish them I think it makes a lot of sense to also dsitinguish them in the modern external interface too.

src/core/dbus-manager.c
```diff
@@ -1364,6 +1380,22 @@ static int method_unsubscribe(sd_bus_message *message, void *userdata, sd_bus_er
                         return r;
                 if (r == 0)
                         return sd_bus_error_set(error, BUS_ERROR_NOT_SUBSCRIBED, "Client is not subscribed.");
+
+                if (sd_bus_message_is_method_call(message, NULL, "Unsubscribe"))
+                        m->n_legacy_subscribers--;
```
poettering, line 1385: maybe do saturated subtraction, just in case...

poettering:
So, this needs some more tweaks I guess, see my comments.

Also, the counting of legacy clients is borked, because clients that subscribe with the legcy iface and then drop off the bus are not properly counted.

I'd really just use another sd_bus_track object for the legacy clients. It's cheap...

src/login/logind-dbus.c
```diff
@@ -3789,6 +3789,12 @@ int match_job_removed(sd_bus_message *message, void *userdata, sd_bus_error *err
                 return 0;
         }
 
+        if (sd_bus_message_is_signal(message, NULL, "JobRemovedEx")) {
```
poettering, line 3792: we expect logind and machined to be update along with pid 1, hence just rework to use JobRemovedEx unconditionally

src/shared/bus-wait-for-jobs.c
```diff
@@ -78,13 +85,38 @@ BusWaitForJobs* bus_wait_for_jobs_free(BusWaitForJobs *d) {
...
+
+        r = bus_call_method(bus,
+                        bus_systemd_mgr,
+                        "SubscribeWithFlags",
+                        &error, &reply, "t", 0);
+        if (r < 0 && !sd_bus_error_has_name(&error, SD_BUS_ERROR_UNKNOWN_METHOD))
+                return log_error_errno(r, "Failed to call SubscribeWithFlags: %s", bus_error_message(&error, r));
+
+        return cached = r >= 0;
+}
```
poettering, line 107: Hmm, I think we should probably update bus_wait_for_jobs() to issue `Subscribe()` too, and remove the explicit calls from logind/machined. That#d be more correct.

konsolebox:
Is this PR abandoned?

PR #22878, opened 2022-03-28
network: rename netdev kind virtual-wlan -> wlan

yuwata:
The Kind= setting in [Match] section of .network files takes "wlan".
This makes the same setting in .netdev files matches the one in .network
files.


yuwata:
Kind=virtual-wlan was introduced after v250. Hence, we can freely rename the value. Please review and merge this before v251.

bluca (approved):

poettering:
also needs an update in the NEWS file?

yuwata:
> also needs an update in the NEWS file?

Sure. I will do soon.

PR #22869, opened 2022-03-25
inotify-util: fix buffer overrun warnings in FOREACH_INOTIFY_EVENT()

yuwata:
Follow-up for 00adc340bb15bc9d634db6caa48f1c964b99f79a.

Hopefully fixes CID#1487019-CID#1487024.

PR #22880, opened 2022-03-28
inotify-util: Fix FOREACH_INOTIFY_EVENT() operator precedence bug

daandemeyer:
'&&' has higher priority than '?:' so the warning was always being
logged on the last iteration of the for loop. Fix this by adding
explicit parentheses to indicate the precedence.


poettering:
ouch

yuwata:
This is already addressed by #22869...

poettering:
yeah i was wondering about https://github.com/systemd/systemd/pull/22869... it does a lot more though?

bluca:
there were several issues raised by coverity, that PR fixes those too

PR #22885, opened 2022-03-28
time-util: assume CLOCK_BOOTTIME always exists

poettering:
Let's raise our supported baseline a bit: CLOCK_BOOTTIME started to work
with timerfd in kernel 3.15 (i.e. back in 2014), let's require support
for it now.

This will raise our baseline only modestly from 3.13 → 3.15.

### DIFF
--- a/README
+++ b/README
@@ -30,7 +30,7 @@ LICENSE:
         LGPL-2.1-or-later for all code, exceptions noted in LICENSES/README.md
 
 REQUIREMENTS:
-        Linux kernel >= 3.13
+        Linux kernel >= 3.15
         Linux kernel >= 4.2 for unified cgroup hierarchy support
         Linux kernel >= 4.10 for cgroup-bpf egress and ingress hooks
         Linux kernel >= 4.15 for cgroup-bpf device hook
--- a/src/basic/time-util.c
+++ b/src/basic/time-util.c
@@ -77,7 +77,7 @@ triple_timestamp* triple_timestamp_get(triple_timestamp *ts) {
 
         ts->realtime = now(CLOCK_REALTIME);
         ts->monotonic = now(CLOCK_MONOTONIC);
-        ts->boottime = clock_boottime_supported() ? now(CLOCK_BOOTTIME) : USEC_INFINITY;
+        ts->boottime = now(CLOCK_BOOTTIME);
 
         return ts;
 }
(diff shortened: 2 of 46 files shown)


yuwata (approved):
LGTM.

bluca (approved):

mnme:
I randomly stumbled upon this pull request and scrolled through the code, I don't have deep knowledge of Linux syscalls or systemd, but some things stood out so I wanted to see if I'm right about it or not. Sorry for bothering if I completely misunderstood something here.

---

The following comment was deleted in `sd-event.c` :

```c
/* Generate a clean error in case CLOCK_BOOTTIME is not available. Note that don't use clock_supported() here,
 * for a reason: there are systems where CLOCK_BOOTTIME is supported, but CLOCK_BOOTTIME_ALARM is not, but for
 * the purpose of getting the time this doesn't matter. */
```

This contradicts what my `man timerfd_create` says about it:

```
CLOCK_BOOTTIME (Since Linux 3.15)
CLOCK_BOOTTIME_ALARM (since Linux 3.11)
```

This leaves me with two possible scenarios:

1. The comment was correct and there are systems where `CLOCK_BOOTTIME` is supported, but `CLOCK_BOOTTIME_ALARM` is not.
2. The comment was wrong and the man page is right, i.e. both `CLOCK_BOOTTIME` and `CLOCK_BOOTTIME_ALARM` are always supported on Linux >= 3.15

Now, I believe there is an impact for either of the scenarios:

### 1. If the comment was correct

In this case, the complete removal of the check inside the `TIMER_MONOTONIC_CLOCK` macro in `core/timer.h` is probably incorrect, as it should still check for `CLOCK_BOOTTIME_ALARM` support.

### 2. If the man page is right

In this case, the `clock_supported` function in `time-util.c` should probably retain the `CLOCK_BOOTTIME_ALARM` case and return true for it, like for the other three clocks. `clock_supported(CLOCK_BOOTTIME_ALARM)` is called in `sleep-config.c`.

---

Unrelated to the above, there is one occurrence of `t->wake_system ? CLOCK_BOOTTIME_ALARM : CLOCK_MONOTONIC` in `core/timer.c` where you could probably use the `TIMER_MONOTONIC_CLOCK` macro. Looking at the code it was probably left as is on purpose to make the code "symmetric" with the `CLOCK_REALTIME` branch. Just good to know that this exists if the macro ever needs to be changed.

poettering:
@mnme the confusion stems from the fact that CLOCK_BOOTTIME was initially available for clock_gettime() but not timerfd_create(). And then CLOCK_BOOTTIME_ALARM was added, in timerfd_create(). And only after that CLOCK_BOOTTIME support was also added to timerfd_create(). Confusing? Yes…

mnme:
@poettering Thanks for the clarification, confusing indeed. Learned something new today :)

PR #22886, opened 2022-03-28
veritysetup: fix typo

gportay:


src/veritysetup/veritysetup.c
```diff
@@ -151,7 +151,7 @@ static int run(int argc, char *argv[]) {
                 size_t l;
 
                 if (argc < 6)
-                        return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "attach requires at least two arguments.");
+                        return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "attach requires at least five arguments.");
```
yuwata: four??

poettering: depends if you count "attach" as an argument. I would not... So yeah, should be four

gportay: I have hesitated. Will change.

PR #22887, opened 2022-03-28
core: don't kill usermode helpers when switching to root fs

fbuihuu:
Since commit e45154c770567507cc51eb78b28a1fae1fcdf396, we rely on the presence
of PF_KTHREAD for figuring out whether a given process is a kernel thread or a
regular process.

This helped solving the issue reported at
~~https://lists.freedesktop.org/archives/systemd-devel/2018-September/041414~~
https://lists.freedesktop.org/archives/systemd-devel/2018-September/041414.html
where a usermode helper (UMH) was not killed by the systemd-shutdown killing
spree because it was considered as a kthread (due to its empty 'cmdline' proc
file), which prevented references on the root fs hold by the UMH from being
released on shutdown.

Although it makes sense to kill usermode helpers during shutdown to free theirs
system resources, it might become a problem when switching to the root fs.

Indeed a couple of drivers use UMHs too for being loaded asynchronously,
'mlx4_core' module being one of these. When this driver is loaded it requests
'mlx4_en' to be loaded asynchronously, which is done by mean of UMH. However
killing the UMH process can lead to the driver half way initialized resulting
in, for example, 6 out of 8 CX3 NICs not being activated although the driver is
reported as fully loaded by the kernel afterwards.

And this can happen during the killing spree of PID1 happening right before
switching to the new root fs and when this driver is embedded in initrd but not
involved in the setup of the root fs [1].

Although this can be considered as a driver design issue, it might make sense
to skip UMH processes during PID1's killing spree as this appears to be unsafe
(as described previously) and allow this driver to be loaded as early as
possible.

This patch hence prevents PID1 from killing UMHs during the switch of root fs
while still killed at shutdown. The detection of UMHs is done based by on the
assumption that their 'root' ancestor is kthreadd which has always the
PF_KTHREAD bit set.

[1] This is the case when the initrd is generated with host-only mode disabled
for example or when the driver is explicitly included in an host-only initrd
(--add-drivers of dracut) in order to pre-load the module since this operation
is slow.

### DIFF
--- a/src/basic/process-util.c
+++ b/src/basic/process-util.c
@@ -398,18 +398,12 @@ int rename_process(const char name[]) {
         return !truncated;
 }
 
-int is_kernel_thread(pid_t pid) {
+static int get_process_flags(pid_t pid, unsigned long long *ret_flags) {
         _cleanup_free_ char *line = NULL;
-        unsigned long long flags;
         size_t l, i;
         const char *p;
         char *q;
         int r;
-
-        if (IN_SET(pid, 0, 1) || pid == getpid_cached()) /* pid 1, and we ourselves certainly aren't a kernel thread */
-                return 0;
-        if (!pid_is_valid(pid))
-                return -EINVAL;
 
         p = procfs_file_alloca(pid, "stat");
         r = read_one_line_file(p, &line);
@@ -449,11 +443,59 @@ int is_kernel_thread(pid_t pid) {
                 return -EINVAL;
         q[l] = 0;
 
-        r = safe_atollu(q, &flags);
+        return safe_atollu(q, ret_flags);
+}
+
+int is_kernel_thread(pid_t pid) {
+        unsigned long long flags;
+        int r;
+
+        if (IN_SET(pid, 0, 1) || pid == getpid_cached()) /* pid 1, and we ourselves certainly aren't a kernel thread */
+                return 0;
+        if (!pid_is_valid(pid))
+                return -EINVAL;
+
+        r = get_process_flags(pid, &flags);
         if (r < 0)
                 return r;
 
-        return !!(flags & PF_KTHREAD);
+        return FLAGS_SET(flags, PF_KTHREAD);
+}
+
+int is_usermode_helper(pid_t pid) {
+
+        /* Usermode helpers are special processes spawned by the kernel (kthreadd more
+         * precisely). A few drivers rely on them for being loaded asynchronously for
+         * example. Hence it's usually not a good idea to kill them except during shutdown. These
+         * processes don't have PF_KTHREAD set but its parent (kthreadd) does. */
+
+        /* pid 1, and we ourselves certainly aren't an usermode helper */
+        if (IN_SET(pid, 0, 1) || pid == getpid_cached())
+                return 0;
+        if (!pid_is_valid(pid))
+                return -EINVAL;
+
+        for (;;) {
+                unsigned long long flags;
+                pid_t ppid;
+                int r;
+
+                r = get_process_ppid(pid, &ppid);
+                if (r < 0)
+                        return r;
+                if (ppid == 1)          /* our child can't be a kernel thread */
+                        return 0;
+                if (!pid_is_valid(ppid))
+                        return -EINVAL;
+
+                r = get_process_flags(ppid, &flags);
+                if (r < 0)
+                        return r;
+                if (FLAGS_SET(flags, PF_KTHREAD)) /* if one of our ancestor is a kthread then we are */
+                        return true;
+
+                pid = ppid;
+        }
 }
 
 int get_process_capeff(pid_t pid, char **ret) {
--- a/src/basic/process-util.h
+++ b/src/basic/process-util.h
@@ -71,6 +71,7 @@ int kill_and_sigcont(pid_t pid, int sig);
 
 int rename_process(const char name[]);
 int is_kernel_thread(pid_t pid);
+int is_usermode_helper(pid_t pid);
 
 int getenv_for_pid(pid_t pid, const char *field, char **_value);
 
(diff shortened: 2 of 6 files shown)


poettering:
Uh, I am not convinced. user mode helpers are backed by the rootfs. Which we want to release here.

UMH are really a broken concept, kernel code really should stop doing that. And if it needs some stuff to work, it probably shouldn't ignore the failure this returns?

I am sorry, but this looks *very* wrong to me.

poettering:
> This helped solving the issue reported at
> https://lists.freedesktop.org/archives/systemd-devel/2018-September/041414

This link does not work?

fbuihuu:
> > This helped solving the issue reported at
> > https://lists.freedesktop.org/archives/systemd-devel/2018-September/041414
> 
> This link does not work?

Please use https://lists.freedesktop.org/archives/systemd-devel/2018-September/041414.html instead.

Werkov:
Hello, @poettering. This is tagged as 'needs-discussion', I've provided some food for thought. I'd like to ask you to consider these and your opinion on them. (It'll be helpful for us to decide how to tackle the situation with the peculiar drivers described previously.)

poettering:
this is supposed to be a tiny bit symmetric to the shutdown logic we run before transitioning back into the initrd: yes, the clean isolate logic should take care of most, and we then just step in to clean up stuff that was accidentally left-over. Of course in a much weaker way (i.e. just SIGTERM, no SIGKILL), but that's mostly because we don't want to wait for this, and can collect the children easily after the transition and attribute them to something reasonable, i.e. the units they originate from (if that exists), and because they don't keep file systems busy we really really have to disassemble.

Thing is, initrds leave all kind of weird stuff around, i.e. gpg/ssh agents forked off fuse mounts, or children of kernel mode helpers (kernel keyring callouts, yuck...), stuff run with KillMode=none and so on. There's just *soo* much that can stick around and that *does* stick around (and we get the reports for that), and which we really should get rid of by default. It's an illusion to believe that initrds had that high quality, that this doesn't happen, but it *does* happen, all the time.

I mean, during shutdown, you could argue that we don't need the second phase of shutdown either, because the first ("clean") phase already took care of it. But reality tells us different, that part is absolutely necessary, things get cleaned up by it *all* *the* *time*...

Or to say this differently: for the stuff that excluded itself from the cgroup tree logic we want an allowlist, and clean up everything else.

Werkov:
Thanks for the quick response.
Ad symmetry and leftovers -- I acknowledge proper full cleanup at shutdown (or when switching away from persistent FS) since it's functionally necessary. I consider the cleanup when switching away from initrd non-functional optimization (to release consumed resources), hence the compromise of symmetry.

> There's just _soo_ much that can stick around and that _does_ stick around (and we get the reports for that), and which we really should get rid of by default.

I think these leftovers must be distinguished into two categories:
a) forgotten stuff not needed anymore (IMO the agents in your examples),
b) processes finishing/continuing their jobs throughout switch root.

The a) ideally should be resolved by better initrd design (proper unit configuration) -- maybe systemd should warn about these so that such a feedback gets through to initrd authors?

The b) should be avoided when killing (actually, your example of keyring callouts sounds similar to the mlx drivers), alas, not all of them are marked with `@` in argv[0].

> Or to say this differently: for the stuff that excluded itself from the cgroup tree logic we want an allowlist, and clean up everything else.

E.g. applying such an allowlist as the @fbuihuu's change (this PR)?

poettering:
> E.g. applying such an allowlist as the @fbuihuu's change (this PR)?

No, as in sets the argv[0][0] = '@'...

I am really not convinced about @fbuihuu's patches because as I read it, left-over agent stuff forked into the backrgound from the keyring upcall for example would be excluded from killing, but it's precisely the stuff we *should* kill.

Werkov:
Sorry, what did you mean with: "stuff that excluded itself from the cgroup tree logic"? My (mis)understanding was those were processes in the root cgroup (i.e. excluded from regular cgroup logic) and those are effectively ignored with that patch.

Werkov:
@poettering Could you please elaborate more about the keyring helpers that you deemed problematic? (IMO it's similar to the modprobe here, i.e. they shouldn't be killed without warranty.)

Also note that it's inconsistent to kill(2) processes that can't be wait(2)ed for (UMHs are children of kthreadd).

Finally, could you also elaborate more about the gpg/ssh agents problems? Those should be in a non-root cgroup, i.e. subject to regular (unit induced) termination. The processes the remain in the root cgroup should rather be excluded (for the multiple reasons explained above), here, effectively achieved by filtering via kthread (forking) ancestry.

poettering:
> @poettering Could you please elaborate more about the keyring helpers that you deemed problematic? (IMO it's similar to the modprobe here, i.e. they shouldn't be killed without warranty.)

The keyring usermode helpers are basically an application server, they can fork *anything* of, see /etc/request-key.conf... By default it even forks off some shell scripts if you look there, the kind of stuff that likes to do "&" to fork something into the bg and leave around. It's a terrible concept if you ask me, and all outside of the cgroup hierarchy.

poettering:
> Finally, could you also elaborate more about the gpg/ssh agents problems? Those should be in a non-root cgroup, i.e. subject to regular (unit induced) termination.

people fork them off their shells. And since request-key stuff forks off shell scripts we should clean that up.

keszybz:
To summarize:
- there's a real problem with `mlx4_core` / `mlx4_en` drivers
- the proposed patch skips killing of all children of kernel threads
- those children can be arbitrary things, incl. shell scripts with left-over background jobs
- we want to kill those processes and @poettering only wants to exclude things explicitly marked with `argv[0][0] = '@'`.

Based on this, I think this patch is the wrong approach and it's not something we'd want to merge. Instead, those things that kernel spawns should mark themselves appropriately. I assume that this should be simple patch, if one knows where to do the change. (I looked at the sources for `mlx_core`/`mlx_en`, but I couldn't figure out where it spawns the helper.)

PR #17336, opened 2020-10-13
units: update serial-getty@.service to support 57600 baud rate

mbiebl:
Fixes: #17306

Issue #17306, opened 2020-10-10
systemd: serial-getty@.service should support 57600 baud


mbiebl:
Version: 246.6
Forwarding this verbatim from the downstream bug report at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=969144

Dear Maintainer,

My BIOS provides pre-boot access to the system over serial, and sets
the port up for 57600.  So I have configured the client (conserver)
to use that speed.  This has happened to work by accident with the
systemd serial-getty@.service because of the  --keep-baud
parameter to agetty.

But today I ended up changing the speed of the port while temporarily
testing another client.  And was surprised to learn that it is
impossible to switch back to 57600 with the default serial-getty@.service 

I see no reason why this standard baud rate should be excluded from
the list of rates.  Please fix

```
- --- a/lib/systemd/system/serial-getty@.service   2020-04-27 19:02:57.000000000 +0200
+++ b/etc/systemd/system/serial-getty@.service   2020-08-28 09:45:03.000000000 +0200
@@ -31,7 +31,7 @@
 # The '-o' option value tells agetty to replace 'login' arguments with an
 # option to preserve environment (-p), followed by '--' for safety, and then
 # the entered username.
- -ExecStart=-/sbin/agetty -o '-p -- \\u' --keep-baud 115200,38400,9600 %I $TERM
+ExecStart=-/sbin/agetty -o '-p -- \\u' --keep-baud 115200,57600,38400,9600 %I $TERM
 Type=idle
 Restart=always
 UtmpIdentifier=%I
```

poettering:
send a PR

mvduin:
This seems like a very bad idea. The auto-baud mechanism of getty relies on a sequence of consecutive zero-bits transmitted at a low baudrate being recognized as a break (i.e. 10 consecutive 0-bits, assuming 8 data bits and no parity or 7 data bits with parity) by the receiver configured at a high baudrate. This only works if the baudrate ratio is sufficiently large.

Switching from 38400 to 9600 requires only 3 consecutive zero-bits (transmitted at 9600 baud), while switching from 115200 to 38400 requires 4 consecutive zero-bits. Assuming we're using 8 data bits and no parity (as is typical), four consecutive zero-bits (one of which may be the start bit) can be triggered by any control character (e.g. pressing enter, escape, or ctrl-C), space, and also the `A` of modem AT-commands.

[...]

PR #22888, opened 2022-03-28
Revert "units: update serial-getty@.service to support 57600 baud rate"

mvduin:
This reverts commit 7fe7547ba3b9 (PR #17336 for issue #17306).

The automatic baudrate switching of agetty relies on having a
sufficiently large ratio between the receive baudrate and the transmit
baudrate, such that some sequence of zero-bits in the transmitted data
is interpreted as a break condition by the receiver.

For 115200 -> 38400 (ratio 3) and 38400 -> 9600 (ratio 4) this will be
triggered by a user pressing enter, space, or ctrl-C, as well as by a
modem AT command or CONNECTED message.

Inserting 57600 into the baudrate list breaks this:

For 115200 -> 57600 (ratio 2), pressing space will still cause a switch
but none of the other aforementioned triggers work anymore.

For 57600 -> 38400 (ratio 1.5), even pressing space may no longer work,
depending on the uart.

### DIFF
--- a/units/serial-getty@.service.in
+++ b/units/serial-getty@.service.in
@@ -33,7 +33,7 @@ Before=rescue.service
 # The '-o' option value tells agetty to replace 'login' arguments with an
 # option to preserve environment (-p), followed by '--' for safety, and then
 # the entered username.
-ExecStart=-/sbin/agetty -o '-p -- \\u' --keep-baud 115200,57600,38400,9600 - $TERM
+ExecStart=-/sbin/agetty -o '-p -- \\u' --keep-baud 115200,38400,9600 - $TERM
 Type=idle
 Restart=always
 UtmpIdentifier=%I


poettering:
@mbiebl what's your take on this?

mbiebl:
The patch is by @bmork , I merely forwarded it.

Would probably be a good idea to have input from either @bmork or from @karelzak as maintainer of agetty.

I'm not really familiar with this particular aspect of agetty.
@mvduin how would you handle the original issue then, i.e. #17306 ?

mvduin:
@mbiebl It appears `--keep-baud` just doesn't work properly. Its description in the manpage claims that the "original baud rate is also saved to the end of the wanted baud rates list. This can be used to return to the original baud rate after unexpected BREAKs." but based on testing this appears to be plain false. Once you switch away from your original baudrate, it will just cycle through the listed baudrates and never back to the original one. So that's just a bug in agetty.

Note however that even if this is fixed, if agetty changes the baudrate and then the user successfully logs in, the original baudrate will never be restored after logout hence the next agetty spawned on that tty will preserve whatever baudrate was left behind by the previous session and the original custom baudrate will be lost. This behaviour seems bad to me, and I feel like the original baudrate _ought_ to be restored when a session ends, but I'm not sure how that might be best achieved.

Regardless, I don't think it's appropriate to break the autobaud detection that getty normally has for the convenience of one specific user's unusual serial baudrate. I'd suggest to them for now to just use a customized service file for their setup.

mbiebl:
I guess this is all the more reason to get @karelzak involved here to find a robust solution

bmork:
> The patch is by @bmork , I merely forwarded it.

And I had completely forgot about this. Didn't really have any strong feelings about it.  I sent the patch because I thought it was The Right Thing To Do, in the hope that it would help others with the same issue.  But if that assumption was wrong, then I'm going to manage fine without it.

However.  I ended up looking at this again now, and I see a few points I believe are interesting for this discussion:

1. agetty saves the original speed to the speed list: https://github.com/util-linux/util-linux/commit/0d855a83095adc953cc9a0df8998a2f916242695

I wasn't aware of that fix since I presumable was using an older version fraom Debian back then.  The original speed ends up last in the list, so proper breaks are required to switch to it.  But this should work fine and solve my original problem.

2. I find the statement "for the convenience of one specific user's unusual serial baudrate" to be be plain ignorant.  I don't think there is any need to prove that 57600 is a quite common serial console speed.  See for example the issue triggering the fix in agetty: https://github.com/util-linux/util-linux/issues/1025

So this is "for the convenience of **two** specific users",  Which is twice as many as you, @mvduin , if you still believe that's a real argument....

3. Then there is the undocumented claim about the "automatic baudrate switching of agetty".  Yes, I understand that agetty a)will work that way provided that 
b) you sort the speed list in descending order, and 
c) list only speeds sufficiently far apart on a logarithmic scale, and
d) receive varying enough data to get some of the characters which are accidentally interpreted as "break" 

But AFAICS from reading the agetty source and manual, this is an undocumented side effect.  The manual (https://github.com/util-linux/util-linux/blob/master/term-utils/agetty.8.adoc#arguments) says:
```
baud_rate,…​

    A comma-separated list of one or more baud rates. Each time agetty receives a BREAK character it advances through the list, which is treated as if it were circular.

    Baud rates should be specified in descending order, so that the null character (Ctrl-@) can also be used for baud-rate switching.

    This argument is optional and unnecessary for virtual terminals.

    The default for serial terminals is keep the current baud rate (see --keep-baud) and if unsuccessful then default to '9600'.
```

Ctrl-@ is the only character which is documented to trigger the automatic switch (in addition to break). This works as long as the ratio is at least 10/8, which always will be true.  There is absolutely nothing supporting a demand that other characters should trigger a speed change.  There isn't even anything documenting that they might.  

So after looking more into the background, source and docs here, I can only conclude that this request is based on a misunderstanding. agetty works as documented with the current speed list.  There are multiple reports of real issues caused by 57600 missing.  There is been no ducumentation or pointers provided by @mvduin to back up the claims made. And personally I really don't see any problems sending either ctrl-@ or break to switch speeds.  Talk about convenience.... 

You didn't answer how you would handle the original issue, @mvduin .  I guess that's because that's only my problem, while you can't be bothered to send a break?  Right....


.

mvduin:
> I wasn't aware of that fix since I presumable was using an older version fraom Debian back then.

Ahh! Yep, turns out the system I was using for testing had an older version of agetty than the one where I was checking the man page. That was a dumb mistake on my part, my apologies for assuming it was a bug.

> But AFAICS from reading the agetty source and manual, this is an undocumented side effect.

I guess wrongly assumed that this is intended behaviour. In practice the preconditions used to be met, and the result was that pressing enter (which is the single most intuitive thing to get a response from a serial console) Just Works™, which is an obvious good thing. This also means there's a risk people may be relying on this without even being aware of it, and having no idea who or what's to blame for it suddenly no longer working or how to fix it (sending a break is not something anyone would guess if they don't already know about it). But, I guess that's just hypothetical and if there's multiple people who want 57600 in the list then that may weigh more strongly.

I originally just wanted to make sure people didn't overlook that adding baudrates to the list isn't a benign trivial change but has some actual downsides as well that should be taken into consideration. I think I worded some of my comments more strongly than I should have, apologies for that.

I do still think that a superior solution would be one that works for _any_ custom baudrate configured by the bootloader and not just 57600. As you pointed out, part of the fix for that is already in agetty. For the remaining part, perhaps an option for systemd to save the terminal settings when starting the service and restoring them when it stops/restarts?

mvduin:
> You didn't answer how you would handle the original issue

Also, I did answer that: the "right" solution would be to ensure that whatever baudrate is configured by the bootloader is retained as option, but the simple and immediate solution is for users who need a custom baudrate to just locally customize the service file to add that custom baudrate.

karelzak:
> agetty relies on having a
> sufficiently large ratio between the receive baudrate and the transmit
> baudrate

agetty does not care about the ratio between baud rates. agetty reacts to the read() syscall result and it switches to the next seed when read() returns zero or when it reads zero (or after parity bit remove "& 0177" it gets zero). Not sure if the kernel plays something more advanced behind the scene on read().

> the original baudrate will never be restored after logout hence the next agetty spawned on that tty will preserve whatever baudrate was left behind by the previous session and the original custom baudrate will be lost. This behaviour seems bad to me, and I feel like the original baudrate ought to be restored when a session ends, but I'm not sure how that might be best achieved.

The original agetty behavior is to always set serial line speed (so ignore what kernel setup for the serial line and do always the reset). This is crazy in many situations because it resets something that already works (for example to print kernel boot messages). The option --keep-baud assumes that agetty is executed on the terminal which is already usable, so it does not touch the speed setting. 

> For the remaining part, perhaps an option for systemd to save the terminal settings when starting the service and restoring them when it stops/restarts?

You're right that the sessions re-cycling on the serial line is a little bit fragile as nothing save/restore the speed due to --keep-baud, but the idea is that nothing (including agetty) modifies the speed. Maybe saving the speed after boot and restoring always before agetty execution is a good idea to make things more robust  (in theory it's possible the user modifies the speed).

mvduin:
> agetty does not care about the ratio between baud rates

The ratio between baudrates is what results in various bytes transmitted at a lower baudrate being recognized as a break or a NUL byte when received at a higher baudrate.

> You're right that the sessions re-cycling on the serial line is a little bit fragile as nothing save/restore the speed due to --keep-baud, but the idea is that nothing (including agetty) modifies the speed.

Well the thing is that the current default config uses --keep-baud combined with a list of baudrates, which means agetty actually _will_ change the baudrate when break or NUL is received, and of course it has no way to restore it after the session since agetty exec()s login.

karelzak:
ad 57600 - I have absolutely no clue how common is this speed, but I guess it's not so common if we had no this speed on agetty command line for years ;-)

mvduin:
Personally I switched to 460800 when I ended up spending some time working via a serial console to fidget with networking on an embedded system. 115200 can still be noticably sluggish for terminal-heavy applications like vim, and 460800 is (surprise!) 4x as fast.

karelzak:
> > agetty does not care about the ratio between baud rates
> 
> The ratio between baudrates is what results in various bytes transmitted at a lower baudrate being recognized as a break or a NUL byte when received at a higher baudrate.

Yes, I only want to say that agetty itself reacts to the zeros, the way how consecutive zero-bits are generated is another story :-)

> > You're right that the sessions re-cycling on the serial line is a little bit fragile as nothing save/restore the speed due to --keep-baud, but the idea is that nothing (including agetty) modifies the speed.
> 
> Well the thing is that the current default config uses --keep-baud combined with a list of baudrates, which means agetty actually _will_ change the baudrate when break or NUL is received, and of course it has no way to restore it after the session since agetty exec()s login.

Yes, but NUL should not be something common I guess. The list of baud rates is there as a fallback solution and agetty continues only when the user is able to type login name otherwise it continues with all specified speeds including the original speed ... (Frankly, I'm not sure how common and real is that baud rate changes "on the fly" after system boot or during an active log-in session.)

mvduin:
Well, it is very much "real", otherwise the original issue (#17306) didn't exist.

karelzak:
Anyway, I don't see anything we can change/fix in agetty, save/restore sounds like a feature request for systems.

mbiebl:
@poettering what do you suggest we do now? It appears whatever we do some use case is broken.

mbiebl:
I.e. keep the status quo?

mbiebl:
I honestly don't have a strong feeling either way.

poettering:
> Note however that even if this is fixed, if agetty changes the baudrate and then the user successfully logs in, the original baudrate will never be restored after logout hence the next agetty spawned on that tty will preserve whatever baudrate was left behind by the previous session and the original custom baudrate will be lost. This behaviour seems bad to me, and I feel like the original baudrate _ought_ to be restored when a session ends, but I'm not sure how that might be best achieved.

We could in theory save and restore that in systemd, we after already reset the tty before and after starting tty services. But I am not sure this is really desirable here. I mean, the way I see it we just stick to the last working thing. So initially that's what the kernel set up for us, but if that didn't work, agetty might have switched to something else, but that's fine, if it works. And I think it makes sense to stick to that then.

I mean, my assumption would be that the hw attached to serial ports isn't reconfigured or replaced all the time. And if it is it's OK to require a manual restart of the service (or an stty invocation). So I am not convinced we should change anything here in this respect. 

I mean, serial ports are not plug and play and a bit messy. They are tech from the 60's, so noone should be surprised. We can make them work to some point automatically but I see no issue at all if there are usecases not covered by our defaults if we cover a large part of them.

mvduin:
> I mean, my assumption would be that the hw attached to serial ports isn't reconfigured or replaced all the time. And thus it's _good_ to stick to what's worked.

That assumption is however not borne out by any use-case presented or referenced in this thread or the original issue though. The whole reason this all started was because agetty changed the baudrate and didn't leave any way to switch back to the original one. Restoring the original tty config would have fixed this.

poettering:
hence, I am tempted to just close this (and the associated issue), with the recommendation to just override the unit file's ExecStart= line with an explicit baud rate configuration if the default logic is not sufficient. There's nothing inherently wrong with requiring configuration for special cases.

poettering:
> That assumption is however not borne out by any use-case presented or referenced in this thread or the original issue though. The whole reason this all started was because agetty changed the baudrate and didn't leave any way to switch back to the original one. Restoring the original tty config would have fixed this.

hmm? not sure follow? Why specifically did it switch the baudrate if the pre-configured work was the right one? it shouldn't siwtch if that works...

mvduin:
> Why specifically did it switch the baudrate if the pre-configured work was the right one? it shouldn't siwtch if that works...

Original issue said:

> But today I ended up changing the speed of the port while temporarily
> testing another client. And was surprised to learn that it is
> impossible to switch back

Because once you trigger a baudrate switch and then attempt to log in, the original baudrate set by the bootloader is lost until you reboot.

keszybz:
I'd lean towards merging this patch.
@mvduin's use case that *used to work* is broken. If we merge the revert, the other use case will be broken, but  @bmork said that that'd be OK. Basically, to get 57600, it needs to be configured explicitly. If we need to make a choice, not regressing is better than supporting another equivalently important use case.

mvduin:
Yeah, I can personally work around it of course, but it is still a regression that I'm worried would probably end up causing at least a few people out there to scratch their head why a thing that used to work no longer does. I _personally_ don't think this is outweighed by the benefits of adding this one specific custom baudrate, but it's also not my decision to make. I mostly just wanted to make sure this regression wasn't overlooked.

poettering:
> I'd lean towards merging this patch.
> @mvduin's use case that _used to work_ is broken. 

But that was back in 245 times, i.e. more than 2y ago...

A regression compared to 2y ago might not really be a regression...

mvduin:
> But that was back in 245 times, i.e. more than 2y ago...

Huh, for some reason I had in my head that it was a recent change, but now that I check again it seems you're right. Yeah that does sound like I'm too late and it may be better to stick with the status quo.

poettering:
OK, given @mvduin last comment, let's close this.

PR #22889, opened 2022-03-28
Allow setting owner/mode for mount targets, and for them to inherit the underlying owner/mode

nabijaczleweli:
Set{User,Group,Mode}= comes up a lot in the issues for https://github.com/systemd/zram-generator, where a common use-case is to create a global /tmp or /tmp-equivalent backed by ext4-on-zram, which is hindered by being unable to, without [horrible hacks](https://github.com/systemd/zram-generator/blob/1bf6b08d8bd15624280482b1c125902f90c706aa/README.md#mount-points), set the mountpoint to a=rwx,o+t.

ReplicateUnderlying=, additionally, mimics the tmpfs behaviour of setting the ugid and mode to that of the underlying directory, which is what everyone Actually Wants 98% of the time.

`reload`ing the mount unit re-applies the settings. None of them failing are actually fatal to the mounts proceeding.

It's not immediately obvious to me if, for example, we want to not use the user's primary group from `SetUser=`? And if not, then allow numerical ugids?
My instinct was to also make these lchown() instead of chown(), but?

The core of this (mount_chown and most of its driver) was stolen from socket.c.

Ref: https://github.com/systemd/zram-generator/issues/150, https://github.com/systemd/zram-generator/issues/146, &c.
Cc: @keszybz for zram-generator

### DIFF
--- a/man/org.freedesktop.systemd1.xml
+++ b/man/org.freedesktop.systemd1.xml
@@ -6127,6 +6127,7 @@ node /org/freedesktop/systemd1/unit/home_2emount {
       readonly b ForceUnmount = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
       readonly b ReadWriteOnly = ...;
+      readonly b ReplicateUnderlying = ...;
       readonly u SetUser = ...;
       readonly u SetGroup = ...;
       readonly u SetMode = ...;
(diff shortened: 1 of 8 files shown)


poettering:
usually, we use tmpfiles to adjust accessmode/ownership of /tmp during boot. Why wouldn't that work here?

poettering:
next question, if the tmpfiles.d/ thing doesn#t suffice (though I don#t grok why it would not), then why should this be a systemd feature and not a /bin/mount feature, available from any such cmdline. i.e. it already supports X-mount.mkdir, and this could be added therein the same fashion.

poettering:
I see this has now been proposed here:
https://lore.kernel.org/util-linux/20220406113814.twdo4mwkevwuayht@ws.net.home/T/#ma1da7aed9ad69dd3501fdfc2d2f4185c798c1466

Let's close this hence here.

PR #22890, opened 2022-03-29
udev: do not call sd_event_source_disable_unref() in workers for event sources created by the main process

yuwata:
Fixes a bug introduced by 9612da361a825d70a9fd392f3ee5a53bf8896887.

C.f. https://github.com/systemd/systemd/pull/22649#issuecomment-1080980217


keszybz:
Let's merge. It looks reasonable, and it's effectively a noop, because before we'd get the warning about PID being different, so this just silences the warning.

PR #22894, opened 2022-03-29
Update hwdb again

keszybz:


bluca (approved):

keszybz:
centos ci: 14:10:42 TEST-50-DISSECT, looks unrelated.

src/cryptsetup/cryptsetup.c
```diff
@@ -1414,7 +1414,7 @@ static int attach_luks_or_plain_or_bitlk_by_tpm2(
 
                         if (is_efi_boot() && !efi_has_tpm2())
                                 return log_notice_errno(SYNTHETIC_ERRNO(EAGAIN),
-                                                        "No TPM2 hardware discovered and EFI bios indicates no support for it either, assuming TPM2-less system, falling back to traditional unlocking.");
```
poettering, line 1417: actually "bios" is kinda wrong.. if you touch this, you could also change this to "firmware", which would be more accurate

poettering:
lgtm

keszybz:
Oh, copr reports that the packages were successfully built: https://copr.fedorainfracloud.org/coprs/packit/systemd-systemd-22894/build/3910857/. This seems only to be a communication problem.

I'll force push with the one-word change and immediately merge. It don't think we need to build this again.

PR #22895, opened 2022-03-29
socket-util: don't attempt SO_RCVBUFFORCE/SO_SNDBUFFORCE if we do not…

fbuihuu:
… have enough privileges

Otherwise such attempts triggers DENIED audit records for applications with
apparmor profile in enforcing mode.

### DIFF
--- a/src/basic/socket-util.c
+++ b/src/basic/socket-util.c
@@ -16,6 +16,7 @@
 #include <linux/if.h>
 
 #include "alloc-util.h"
+#include "capability-util.h"
 #include "errno-util.h"
 #include "escape.h"
 #include "fd-util.h"
@@ -735,6 +736,12 @@ int fd_set_sndbuf(int fd, size_t n, bool increase) {
         if (r >= 0 && l == sizeof(value) && increase ? (size_t) value >= n*2 : (size_t) value == n*2)
                 return 1;
 
+        /* Setting the buffer size beyond the kernel limit needs net_admin capability. If we know
+         * this cap is missing don't even try SO_SNDBUFFORCE to avoid audit records reporting this
+         * unauthorized attempt.  */
+        if (!have_effective_cap(CAP_NET_ADMIN))
+                return -EPERM;
+
         /* If we have the privileges we will ignore the kernel limit. */
         r = setsockopt_int(fd, SOL_SOCKET, SO_SNDBUFFORCE, n);
         if (r < 0)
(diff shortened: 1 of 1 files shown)


poettering:
uh, let's not second guess the kernel. cap checks change, caps are added. This is something to fix in aa policy.

We will not duplicate all kinds of priv checks for individual kernel calls in userspace. This can only fail, we cannot possible do this correctly.

Sorry, but no.

fbuihuu:
I agree but the thing is that the `net_admin` cap needs to be granted through the aa profile simply because it relies on `sd_notify()`, which ends up calling `fd_inc_sndbuf(fd, 8MB)`.

It's not obvious and I'm not sure it's correct either.

noelpower:
There are a couple of things here
Regarding "This is something to fix in aa policy", well it's like this there are 2 options to fix this with an aa policy, none are ideal
1. putting a deny in the smbd (and other) profiles of any systemd process that calls sd_notify (e.g. apache)
2. allow the net_admin privilege

1. has the disadvantage of totally squashing any logs to do with net_admin privileges, if the program involved at some stage really needs this capability it will fail (silently) and will be hard to track down
2. giving a process unneeded elevated privileges. In this case the privilege is net_admin which allows reconfiguration of your network, I'd say there are 2 changes of that being configured, none & none

I also wonder why the size that is being configured here is so large (that it triggers this situation) since it seems that no one cares if this attempt to force the buffer size fails (sd_pid_notify_with_fds(..) calls (void) fd_inc_sndbuf(fd, SNDBUF_SIZE))

This was fixed in the past for this very same problem so it kindof feels like this is a regression

If such a change here really cannot happen for some reason is there any other way to avoid this ? would it be possible for the code to be changed in such a way that the caller of sd_notify could influence whether there is an attempt to force the buffersize (via env variable or api change)

poettering:
hmm? selinux has the ability to make AVC denials silent. Do the same here. It's OK to deny the sockopt for lack of privs. We execue the sockopt full well knowing it won#t work if we have no privs. 

We only issue the sockopt since it unclogs booting a bit, since we less likely have to sync on sd_notify() if we use it a lot. If the socket buffer is short there's a realistic chance we have to wait for PID 1 to catch up processing our messages, and we'd rather avoid that.

This is a best effort thing: if it works great, if it doesn't, things will work correctly too, just maybe a tiny bit slower in some apps that issue a lot of sd_notify(). that's all.

PR #22899, opened 2022-03-29
network: automatically determine timeout of waiting for carrier regain

### LINKED ISSUE #19832: systemd-networkd: Conditional bonding based on SSID doesn't works (interface not enslaved)
 > systemd 248 (248.3-2-arch)

**Used distribution**
 > Arch Linux

 > `Linux arch 5.12.9-arch1-1 #1 SMP PREEMPT Thu, 03 Jun 2021 11:36:13 +0000 x86_64 GNU/Linux`

 > x86_64

**Expected behavior you didn't see**
 > I've configured wired/wireless bonding using this very simple configuration: https://gist.github.com/Zoddo/4d709bf0099cbe072ce2319b697b055d  
> This works well (when unplugging the ethernet cable, the connection seamlessly switches over Wifi), and both physical interfaces are shown as enslaved:
> ```
> $ networkctl
> IDX LINK   TYPE     OPERATIONAL SETUP     
>   1 lo     loopback carrier     unmanaged
>   2 bond0  bond     routable    configured
>   3 enp8s0 ether    enslaved    configured
>   4 wlan0  wlan     enslaved    configured
> ```
>
> However, I want the bonding to be active only when connected to a specific SSID.
> Therefore, I changed `25-wireless.network` to a simple DHCP configuration, and moved the bonding configuration in `21-wireless-zoddo.network` as shown in this gist: https://gist.github.com/Zoddo/169817399995bac59dc35809dcda24e1 (other files are unchanged)
>
> I expect the `wlan0` interface to be part of `bond0` whilst connected to SSID `Zoddo`, and to use the "standalone" DHCP configuration when connected to another network.

 > When using this "conditional bonding" configuration, `wlan0` isn't properly enslaved when connected to the `Zoddo` SSID.  
 > `networkctl` and `ip addr` output can be found here: https://gist.github.com/Zoddo/9af07cf1b3d886fbf6990bb231805f36  
 > I've also included journalctl logs of what happens when connecting for the first time to the SSID `Zoddo`.
 >
 > Looking through these logs shows that when connecting to that SSID, `wlan0` is quickly reset (line 14) before being added to the bond. I don't know if this reset happens because of the switch to another config file or because of the interface being added to the bond.
 >
 > So, the interface is successfully added in the bond as a down link (line 16), and is then removed again (line 23) because of networkd switching again to the default configuration (probably because we are not connected anymore to the SSID).  
 > We are quickly reconnected to the SSID, and networkd switch again to the SSID-specific configuration (line 40), but the interface is not added again to the bond.
 >
 >
 > I think the issue lies in the interface reset occuring. I don't know if this reset is triggered by systemd-networkd, or by the kernel.  
 > Anyway, I think this can be fixed on the networkd-side by either not resetting the interface (if done by networkd, and if possible), or by not switching back to the default configuration whilst the carrier remain down (probably the only fix if this reset is done by the kernel).

 > Try to bond an interface conditionally based on the current SSID as shown above (ie., using the config provided in gists)

I will happily provide more logs or other debug information, or test some fixes if needed.

yuwata:
Closes #19832.

### DIFF
--- a/src/network/networkd-dhcp4.c
+++ b/src/network/networkd-dhcp4.c
@@ -735,7 +735,6 @@ static int dhcp4_request_routes(Link *link) {
 }
 
 static int dhcp_reset_mtu(Link *link) {
-        uint16_t mtu;
         int r;
 
         assert(link);
@@ -743,18 +742,9 @@ static int dhcp_reset_mtu(Link *link) {
         if (!link->network->dhcp_use_mtu)
                 return 0;
 
-        r = sd_dhcp_lease_get_mtu(link->dhcp_lease, &mtu);
-        if (r == -ENODATA)
-                return 0;
-        if (r < 0)
-                return log_link_error_errno(link, r, "DHCP error: failed to get MTU from lease: %m");
-
-        if (link->original_mtu == mtu)
-                return 0;
-
         r = link_request_to_set_mtu(link, link->original_mtu);
         if (r < 0)
-                return log_link_error_errno(link, r, "DHCP error: could not reset MTU: %m");
+                return log_link_error_errno(link, r, "DHCP error: Could not queue request to reset MTU: %m");
 
         return 0;
 }


man/systemd.network.xml
```diff
@@ -917,17 +917,36 @@ Table=1234</programlisting></para>
...
+          <literal>infinite</literal> is equivalent to <literal>yes</literal>.</para>
+
+          <para>Setting a finite timespan may be useful when e.g. in the following cases:
+          <itemizedlist>
+            <listitem>
+              <para>A wireless interface connecting to a network which has multiple access points with
+              the same SSID.</para>
+            </listitem>
+            <listitem>
+              <para>Enslaving a wireless interface to a bond interface, as which may disconnect from
```
bluca: `interface, which may disconnect`

man/systemd.network.xml
```diff
@@ -917,17 +917,36 @@ Table=1234</programlisting></para>
...
+              the connected access point and causes its carrier lost.</para>
```
bluca: `causes its carrier to be lost`

man/systemd.network.xml
```diff
@@ -917,17 +917,36 @@ Table=1234</programlisting></para>
...
+            </listitem>
+            <listitem>
+              <para>The driver of the interface resets when the MTU is changed.</para>
+            </listitem>
+          </itemizedlist>
           </para>
+
+          <para>When <varname>Bond=</varname> is specified to a wireleass interface, defaults to 3
```
bluca: /wireleass/wireless/

src/network/networkd-dhcp4.c
```diff
@@ -735,26 +735,16 @@ static int dhcp4_request_routes(Link *link) {
...
-        if (r < 0)
-                return log_link_error_errno(link, r, "DHCP error: failed to get MTU from lease: %m");
-
-        if (link->original_mtu == mtu)
-                return 0;
-
         r = link_request_to_set_mtu(link, link->original_mtu);
         if (r < 0)
-                return log_link_error_errno(link, r, "DHCP error: could not reset MTU: %m");
+                return log_link_error_errno(link, r, "DHCP error: Could not request to reset MTU: %m");
```
bluca: maybe `Could not queue request to reset MTU`?

bluca (approved):

yuwata:
@bluca Thank you for the review. All comments are addressed. Upgrading the green label.

PR #22904, opened 2022-03-29
man: update root-hash-signature option with value

gportay:
This documents two possible values expected by the option
root-hash-signature for veritytab and veritysetup-generator.

### DIFF
--- a/man/systemd-veritysetup-generator.xml
+++ b/man/systemd-veritysetup-generator.xml
@@ -87,7 +87,7 @@
         <listitem><para>Takes a comma-separated list of dm-verity options. Expects the following options
         <option>ignore-corruption</option>, <option>restart-on-corruption</option>, <option>ignore-zero-blocks</option>,
         <option>check-at-most-once</option>, <option>panic-on-corruption</option> and
-        <option>root-hash-signature</option>. See
+        <option>root-hash-signature=<replaceable>PATH</replaceable>|base64:<replaceable>HEX</replaceable></option>. See
         <citerefentry project='die-net'><refentrytitle>veritysetup</refentrytitle><manvolnum>8</manvolnum></citerefentry> for more
         details.</para></listitem>
       </varlistentry>
--- a/man/veritytab.xml
+++ b/man/veritytab.xml
@@ -95,7 +95,7 @@ This is based on crypttab(5).
       </varlistentry>
 
       <varlistentry>
-        <term><option>root-hash-signature=</option></term>
+        <term><option>root-hash-signature=<replaceable>PATH</replaceable>|base64:<replaceable>HEX</replaceable></option></term>
 
         <listitem><para>A base64 string encoding the root hash signature prefixed by <literal>base64:</literal> or a
         path to roothash signature file used to verify the root hash (in kernel). This feature requires Linux kernel


bluca (approved):

man/systemd-veritysetup-generator.xml
```diff
@@ -87,7 +87,7 @@
         <listitem><para>Takes a comma-separated list of dm-verity options. Expects the following options
         <option>ignore-corruption</option>, <option>restart-on-corruption</option>, <option>ignore-zero-blocks</option>,
         <option>check-at-most-once</option>, <option>panic-on-corruption</option> and
-        <option>root-hash-signature</option>. See
+        <option>root-hash-signature=<replaceable>bash64:HEX|PATH</replaceable></option>. See
```
keszybz: I have to admit I was confused by this. The syntax is informal, but it's not clear if ":' or "|" have the higher precedence. So please write this as `root-hash-signature=<replaceable>path</replaceable>|base64:<replaceable>hex</replaceable>`. Obviously there's a typo also.

keszybz: I think it'd be nice to say something more about the options too. Contrary to the suggestion, veritysetup(8) doesn't say anything about those options. Some hints what `<replaceable>path</replaceable>` should point to, etc., would be useful.

gportay: > I have to admit I was confused by this. The syntax is informal, but it's not clear if ":' or "|" have the higher precedence. So please write this as `root-hash-signature=<replaceable>path</replaceable>|base64:<replaceable>hex</replaceable>`. Obviously there's a typo also.

Oh, good catch for the typo. I put `PATH` and `HEX` in upper case; tell me if you prefer to stick to lower case.

yuwata (approved):

man/systemd-veritysetup-generator.xml
gportay: > I think it'd be nice to say something more about the options too. Contrary to the suggestion, veritysetup(8) doesn't say anything about those options. Some hints what `<replaceable>path</replaceable>` should point to, etc., would be useful.

Sorry, it is unclear to me. By options, you means all the options? I mean `ignore-corruption`, `restart-on-corruption`, `ignore-zero-blocks`, `check-at-most-once`, `panic-on-corruption` and `root-hash-signature`; or only `root-hash-signature`.

`veritysetup(8)` says:

```
       --root-hash-signature=FILE
              Path to roothash signature file used to verify the root hash (in kernel).  This feature requires Linux kernel  version 5.4 or more recent.
```

and the `dm-verity` kernel documentation says

```
root_hash_sig_key_desc <key_description>

    This is the description of the USER_KEY that the kernel will lookup to get the pkcs7 signature of the roothash. The pkcs7 signature is used to validate the root hash during the creation of the device mapper block device. Verification of roothash depends on the config DM_VERITY_VERIFY_ROOTHASH_SIG being set in the kernel. The signatures are checked against the builtin trusted keyring by default, or the secondary trusted keyring if DM_VERITY_VERIFY_ROOTHASH_SIG_SECONDARY_KEYRING is set. The secondary trusted keyring includes by default the builtin trusted keyring, and it can also gain new certificates at run time if they are signed by a certificate already in the secondary trusted keyring.
```

I guess, the kernel documentation is more complete. Maybe I can link the [verity](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/verity.html) documentation.

gportay: I wonder if a note about how to create the root-hash signature file would not be more appropriate.

PR #22908, opened 2022-03-30
veritysetup: fix parsing of root-hash-signature= option

keszybz:
The function was named confusingly and we managed to confused ourselves. The
parameter was assigned incorrectly and then reassigned correctly in the caller.
Let's simplify the whole thing by just saving the optarg param.

I considered moving the unhexmemming and/or reading of the file to the parse
function, but decided against it. I think it's nicer to parse all options
before opening external files.

/cc @gportay

### DIFF
--- a/src/veritysetup/veritysetup.c
+++ b/src/veritysetup/veritysetup.c
@@ -16,9 +16,7 @@
 #include "terminal-util.h"
 
 static uint32_t arg_activate_flags = CRYPT_ACTIVATE_READONLY;
-static char *arg_root_hash_signature = NULL;
-
-STATIC_DESTRUCTOR_REGISTER(arg_root_hash_signature, freep);
+static const char *arg_root_hash_signature = NULL;
 
 static int help(void) {
         _cleanup_free_ char *link = NULL;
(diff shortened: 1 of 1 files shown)


bluca:
```
./src/veritysetup/veritysetup.c: In function ‘parse_options’:
../src/veritysetup/veritysetup.c:101:29: error: implicit declaration of function ‘parse_roothashsig_option’; did you mean ‘save_roothashsig_option’? [-Werror=implicit-function-declaration]
  101 |                         r = parse_roothashsig_option(val, true);
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~
      |                             save_roothashsig_option
..
```

bluca (approved):

src/veritysetup/veritysetup.c
```diff
@@ -16,9 +16,7 @@
 #include "terminal-util.h"
 
 static uint32_t arg_activate_flags = CRYPT_ACTIVATE_READONLY;
-static char *arg_root_hash_signature = NULL;
-
-STATIC_DESTRUCTOR_REGISTER(arg_root_hash_signature, freep);
+static const char *arg_root_hash_signature = NULL;
```
poettering, line 19: this is not OK. you must copy the string if it comes from the option string, since that one is freed when parse_options() is left.

keszybz: Is it? I think it's just called with argv[6], i.e. a static string.

poettering: look again, i.e. the invocation of the function with "val" is with an allocated string

poettering: → #22942

PR #22910, opened 2022-03-30
udev: do not append unknown errno or signal name

yuwata:
Follow-up for 6467bda59d571696b645e8bbdf31926676890956.

Addresses https://github.com/systemd/systemd/pull/22871#discussion_r837705779.


yuwata:
The CentOS CI failure should be unrelated. See #22911.

PR #22915, opened 2022-03-31
Fix "link-local" language inconsistencies

smopucilowski:
"Link-local" and "link local" are used throughout man pages and program
output, with the former used far more than the latter. This commit makes
it consistent throughout the project.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -366,7 +366,7 @@
       <varlistentry>
         <term><varname>IPv6LinkLocalAddressGenerationMode=</varname></term>
         <listitem>
-          <para>Specifies how IPv6 link local address is generated. Takes one of
+          <para>Specifies how IPv6 link-local address is generated. Takes one of
           <literal>eui64</literal>, <literal>none</literal>, <literal>stable-privacy</literal> and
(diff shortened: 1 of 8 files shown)


src/basic/linux/if_ether.h
```diff
@@ -81,7 +81,7 @@
 #define ETH_P_ATMMPOA	0x884c		/* MultiProtocol Over ATM	*/
 #define ETH_P_PPP_DISC	0x8863		/* PPPoE discovery messages     */
 #define ETH_P_PPP_SES	0x8864		/* PPPoE session messages	*/
-#define ETH_P_LINK_CTL	0x886c		/* HPNA, wlan link local tunnel */
+#define ETH_P_LINK_CTL	0x886c		/* HPNA, wlan link-local tunnel */
```
yuwata: Please do not edit this file. This is imported from the kernel source.

yuwata (changes requested):
Thank you. Mostly good. But one comment.

src/basic/linux/if_ether.h
smopucilowski: Many thanks for reviewing. Corrected.

bluca (approved):

Issue #22906, opened 2022-03-29
RFE: provide a helper for opening `/dev/XYZ` in a race-free way


DemiMarie:
**Is your feature request related to a problem? Please describe.**
Currently, opening almost any file under `/dev` is racy against device hotplug: the kernel uses major and minor numbers to identify the device, but those are subject to reuse.  I would like to be able to open such a device without a risk of reuse.

**Describe the solution you'd like**
I suggest using a layer of indirection: create a `/dev/disk/by-diskseq` directory, and have every block-device symlink point to one 

**Describe alternatives you've considered**
None

**The systemd version you checked that didn't have the feature you are asking for**
git HEAD

poettering:
well, but how would that work? from userspace we cannot atomically have the device node removed whenver a major/minor is gone. I mean, the device nodes are on the lower level *always* about major/minor ids, no diskseqs, and we can't fake that...

So, if you find a device in the udev db, and you see major/minor + diskseq in the db entry, then the only fully safe way to open the device would be to open() the device node with O_PATH, pinning the device node inode, verify major/minor are right, then open the node via `/proc/self/fd/<fd>` properly (i.e. without O_PATH, so that we get the backing device), and then issue the diskseq ioctl, and compare it with the diskseq if the db entry. Only if everything matches we the fd refers to the device the udev db entry was about.

This is admittedly quite complex, but i don#t see how we could make this work otherwise.

We could add an sd-device API for that maybe, i.e.


```c
int sd_device_open(sd_device *d, int mode);
```

Which would return you an fd to the block device node you could then use.

I kinda like this idea.

@yuwata what do you say?

poettering:
I guess, for starters, the various udev components that actually open devices for probing info should start using such a new api...

DemiMarie:
> well, but how would that work? from userspace we cannot atomically have the device node removed whenver a major/minor is gone. I mean, the device nodes are on the lower level _always_ about major/minor ids, no diskseqs, and we can't fake that...

[...]

PR #22916, opened 2022-03-31
udev: add /dev/disk/by-diskseq symlink for block devices

poettering:
This adds another symlink for block devices:

    /dev/disk/by-diskseq/<number>

where the number is the diskseq number as exposed by the kernel. It's
useful for apps because they can use it to open a device by diskseq, in
a way that is safe against device node reuse. I.e. if a device node path
like this is passed to an app it could open the device node via the
symlink and also parse the diskseq from the path. Once the device is
opened it could compare the parsed diskseq with the one returned by
BLKGETDISKSEQ on the open node, and if it matches they know they are
talking to the right device.

Fixes: #22906

### DIFF
--- a/rules.d/60-persistent-storage.rules
+++ b/rules.d/60-persistent-storage.rules
@@ -121,4 +121,9 @@ ENV{DEVTYPE}=="partition", ENV{ID_WWN_WITH_EXTENSION}=="?*", SYMLINK+="disk/by-i
 ENV{ID_PART_ENTRY_UUID}=="?*", SYMLINK+="disk/by-partuuid/$env{ID_PART_ENTRY_UUID}"
 ENV{ID_PART_ENTRY_SCHEME}=="gpt", ENV{ID_PART_ENTRY_NAME}=="?*", SYMLINK+="disk/by-partlabel/$env{ID_PART_ENTRY_NAME}"
 
+# by-diskseq link (if an app is told to open a path like this, they may parse
+# the diskseq number from the path, then issue BLKGETDISKSEQ to verify they really got
+# the right device, to access specific disks in a race-free fashion)
+ENV{DISKSEQ}=="?*", ENV{DEVTYPE}!="partition", SYMLINK+="disk/by-diskseq/$env{DISKSEQ}"
+
 LABEL="persistent_storage_end"


DemiMarie:
Looks good.  Still not as good as proper kernel support but such is life.

bluca (approved):

yuwata:
On my laptop:
```
$ grep -r DISKSEQ /sys/class/block/*/uevent
/sys/class/block/loop0/uevent:DISKSEQ=5
/sys/class/block/loop1/uevent:DISKSEQ=6
/sys/class/block/nvme0n1/uevent:DISKSEQ=1
/sys/class/block/nvme0n1p1/uevent:DISKSEQ=1
/sys/class/block/nvme0n1p2/uevent:DISKSEQ=1
/sys/class/block/sda/uevent:DISKSEQ=4
/sys/class/block/zram0/uevent:DISKSEQ=3
```
So, it seems DISKSEQ is shared with the whole block device and its partitions.

rules.d/60-persistent-storage.rules
```diff
@@ -121,4 +121,9 @@ ENV{DEVTYPE}=="partition", ENV{ID_WWN_WITH_EXTENSION}=="?*", SYMLINK+="disk/by-i
 ENV{ID_PART_ENTRY_UUID}=="?*", SYMLINK+="disk/by-partuuid/$env{ID_PART_ENTRY_UUID}"
 ENV{ID_PART_ENTRY_SCHEME}=="gpt", ENV{ID_PART_ENTRY_NAME}=="?*", SYMLINK+="disk/by-partlabel/$env{ID_PART_ENTRY_NAME}"
 
+# by-diskseq link (if an app is told to open a path like this, they may parse
+# the diskseq number from the path, then issue BLKGETDISKSEQ to verify they really got
+# the right device, to access specific disks in a race-free fashion)
+ENV{DISKSEQ}=="?*",SYMLINK+="disk/by-diskseq/$env{DISKSEQ}"
```
yuwata: So, at least, `ENV{DEVTYPE}=="disk"` or `ENV{DEVTYPE}!="partition"` needs to be added.

poettering: indeed

poettering:
Force pushed new version, and verified it works. ptal

yuwata (approved):
LGTM.

PR #22921, opened 2022-03-31
userns uid range tweaks: taint systemd if assigned userns uid range too short, and show userns uid range in userdbctl output

poettering:
Let's make it easier to work with containers that have userns enabled:

1. if the userns uid range is less than 2^16 in size, let's taint systemd. We rely on the range to be around, and document that, hence if stuff is missing, complain.
2. show assigned userns of current contianer in "userdbctl" to give a better idea how UID ranges are looking.

### DIFF
--- a/src/userdb/userdbctl.c
+++ b/src/userdb/userdbctl.c
@@ -17,6 +17,7 @@
 #include "socket-util.h"
 #include "strv.h"
 #include "terminal-util.h"
+#include "uid-range.h"
 #include "user-record-show.h"
 #include "user-util.h"
 #include "userdb.h"
@@ -167,13 +168,21 @@ static const struct {
         },
 };
 
-static int table_add_uid_boundaries(Table *table) {
+static int table_add_uid_boundaries(
+                Table *table,
+                const UidRange *p,
+                size_t n) {
         int r;
 
         assert(table);
+        assert(p || n == 0);
 
         for (size_t i = 0; i < ELEMENTSOF(uid_range_table); i++) {
                 _cleanup_free_ char *name = NULL, *comment = NULL;
+
+                if (n > 0 &&
+                    !uid_range_covers(p, n, uid_range_table[i].first, uid_range_table[i].last - uid_range_table[i].first + 1))
+                        continue;
 
                 name = strjoin(special_glyph(SPECIAL_GLYPH_ARROW_DOWN),
                                " begin ", uid_range_table[i].name, " users ",
@@ -199,7 +208,7 @@ static int table_add_uid_boundaries(Table *table) {
                                 TABLE_SET_COLOR, ansi_grey(),
                                 TABLE_EMPTY,
                                 TABLE_EMPTY,
-                                TABLE_INT, -1); /* sort before an other entry with the same UID */
+                                TABLE_INT, -1); /* sort before any other entry with the same UID */
                 if (r < 0)
                         return table_log_add_error(r);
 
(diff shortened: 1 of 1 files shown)


src/test/test-uid-range.c
```diff
@@ -7,6 +7,8 @@
 #include "uid-range.h"
 #include "user-util.h"
 #include "util.h"
+#include "virt.h"
+#include "errno-util.h"
```
bluca: sorting?

src/test/test-uid-range.c
```diff
@@ -72,4 +74,24 @@ TEST(uid_range) {
         assert_se(p[0].nr == 1983);
 }
 
+TEST(load_userns) {
+        _cleanup_free_ UidRange *p = NULL;
+        size_t n = 0;
+        int r;
+
+        r = uid_range_load_userns(&p, &n, NULL);
+        if (ERRNO_IS_NOT_SUPPORTED(r))
```
bluca, line 104: maybe create a mock file so that the test works everywhere?

src/core/manager.c
```diff
@@ -4349,15 +4350,28 @@ int manager_dispatch_user_lookup_fd(sd_event_source *source, int fd, uint32_t re
...
+static int short_uid_range(const char *path) {
+        _cleanup_free_ UidRange *p = NULL;
+        size_t n = 0;
+        int r;
+
+        r = uid_range_load_userns(&p, &n, NULL);
+        if (ERRNO_IS_NOT_SUPPORTED(r))
+                return false;
+        if (r < 0)
+                return log_debug_errno(r, "Failed to load %s: %m", path);
```
bluca, line 4368: `strna(path)`? or assert

src/userdb/userdbctl.c
```diff
@@ -237,6 +246,104 @@ static int table_add_uid_boundaries(Table *table) {
...
+                        TABLE_SET_COLOR, ansi_grey(),
+                        TABLE_EMPTY,
+                        TABLE_UID, start,
+                        TABLE_SET_COLOR, ansi_grey(),
+                        TABLE_EMPTY,
+                        TABLE_STRING, "First unavailable user",
+                        TABLE_SET_COLOR, ansi_grey(),
+                        TABLE_EMPTY,
+                        TABLE_EMPTY,
+                        TABLE_INT, -1); /* sort before an other entry with the same UID */
```
bluca, line 275: /an/any/

bluca: or /an other/another/ ?

bluca (approved):
needs rebase

src/core/manager.c
```diff
@@ -4350,16 +4351,31 @@ int manager_dispatch_user_lookup_fd(sd_event_source *source, int fd, uint32_t re
...
 }
 
+static int short_uid_range(const char *path) {
+        _cleanup_free_ UidRange *p = NULL;
+        size_t n = 0;
+        int r;
+
+        assert(path);
+
+        r = uid_range_load_userns(&p, &n, NULL);
```
yuwata: `NULL` -> `path`??

Otherwise, `path` is used only in the log message. And gid map will never checked.

yuwata (changes requested):

src/core/manager.c
poettering: ouch

poettering:
fix the one issue @yuwata found. no other changes. Taking liberty to set green label again

yuwata (approved):
LGTM.

PR #22933, opened 2022-04-01
process-util: refactor APIs for reading /proc/self/xyz symlinks

poettering:
The three functions for reading cwd, exe and root symlinks of processes
already share a common core: get_process_link_contents(). Let's refactor
that a bit, and move formatting of the /proc/self/ path into this helper
function instead of doing that in the caller, thus sharing more code.

While we are at it, make the return parameters optional, in case the
information if the links are readable is interesting, but the contents
is not. (This also means safe_getcwd() and readlinkat_malloc() are
updated to make the return parameter optional, as these are called by
the relevant three functions)

### DIFF
--- a/src/basic/fs-util.c
+++ b/src/basic/fs-util.c
@@ -119,7 +119,6 @@ int readlinkat_malloc(int fd, const char *p, char **ret) {
         size_t l = PATH_MAX;
 
         assert(p);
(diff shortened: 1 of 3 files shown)


yuwata (approved):
LGTM.

PR #22934, opened 2022-04-01
tests: make test-resolved-stream suceed even when run as root with restrictive access mode on build tree dir

poettering:


yuwata (approved):
LGTM.

PR #22937, opened 2022-04-01
RFC Partition cgroup subtree of services with enabled Delegate=

Werkov:
This series adds a new configuration option to service units that allows users to specify how the delegated cgroup subtree should be organized. It builds upon distinction of payload and control processes. Control processes are basically anything besides `ExecStart=`. This is motivated by the conflict where we promise to give exclusive rights over the delegated subtree to the service but it can be violated when systemd needs to spawn a control command for the given service. This implements a proposal from #22598 of having two separated subtrees for delegate services.

In order to maintain backwards compatibility this separation is only an opt-in that must be explicitly configured. Additionally, this allows more flexibility, such as not having two separate subtrees, instead the service unit file provides information to systemd about internal structure of delegated subtree, which can be leveraged to place control commands into custom cgroups (demonstration is with the in-house `user@.service`).

There are couple of reasons why I mark this RFC:

 * I still find the payload placement an implementation detail not useful to users (in contrast to have configurable placement of control commands),
 * it opens a possibility for users to shoot themselves in the leg when control commands wrap cgroup path is an inner-node, thus preventing execution of control commands (it's mostly deliberate since fallback would be the original non-ideal behavior),
 * it keeps the provisions for threaded mode from #22598,
 * the cgroup limit/weight/alloc attributes are set for the top service cgroup, while user access is delegated to the payload subtree only,
    * that relies on proper cgroup v2 inheritance semantics,
 * there are few FIXMEs, XXXs in the code (before feedback on the approach),
 * user docs is missing (before feedback on the approach).


src/core/cgroup.c
```diff
@@ -243,6 +243,9 @@ void cgroup_context_remove_socket_bind(CGroupSocketBindItem **head) {
 void cgroup_context_done(CGroupContext *c) {
         assert(c);
 
+        c->delegate_path_control = mfree(c->delegate_path_control);
+        c->delegate_path_payload = nfree(c->delegate_path_payload);
```
yuwata: mfree()?

Werkov: Good catch. Rebase mistake, forcefully pushed correction now.

poettering:
I am sorry, but no, let's not make this configurable. "Compatibility" you say, but with what? Programs that hardcode their cgroup path are borked anyway, we don't have to bother. Programs should run the in the cgroup they are invoked in, or something below in case delegation is on, that's a very simple rule that was always in place.

poettering:
I am still very much against fiddling with threaded cgroups, as mentioned before. Let people do that down the hierarchy.

Again, I am still very sure that if there's more than ExecStart= in a unit with delegation turned on we should put payload/ and control/ cgroups in the unit's main cgroup. That is simple, systematic and should solve the problem comprehensively.

Werkov:
FWIW, there's a planned discussion around this topic on the upcoming [LPC](https://lpc.events/event/16/contributions/1184/).

Werkov:
I'm closing this proposal in favor of #24961.

PR #22941, opened 2022-04-01
cgroup: Drop redundant condition when reading cgroup accounts

Werkov:
The various *_accounting variables are just configuration options from
users and all controllers actually carry out accounting if they are
enabled for whatever reason (e.g. configured limit).

Drop the redundant check in getter functions (it is redundant since
cgroup_realized_mask is checked in respective code paths).

As a result, this provides more convenience, it is sufficient to just
configure e.g. a MemoryMax=$foo and read `systemctl -p MemoryCurrent`
without dealing with `MemoryAccounting=yes`.

Note: Similar does not apply to IPAccounting=, since there's no related
config that'd enable IP accounting implicitly.

### DIFF
--- a/src/core/cgroup.c
+++ b/src/core/cgroup.c
@@ -3586,9 +3586,6 @@ int unit_get_memory_available(Unit *u, uint64_t *ret) {
          * claim before hitting the configured cgroup limits (if any). Consider both MemoryHigh
          * and MemoryMax, and also any slice the unit might be nested below. */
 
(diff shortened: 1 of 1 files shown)


poettering:
I added this explicitly actually. People hsould enable accounting explicitly if they want accounting. Otherwise people might end up relying on side-effects, and that sucks.

See: cf3b4be101acb396fe3b9504685a970be7f86764

Werkov:
Ack. Thanks for the reference, I see how this could give inaccurate data for runtime.

PR #22942, opened 2022-04-01
veritysetup: fix memory corruption

poettering:
We must copy the option string, since in one case we are called with a
pointer into dynamic memory that will be freed by the caller.

As discussed here: https://github.com/systemd/systemd/pull/22908/files#r839394490

Follow-up for: #22908

### DIFF
--- a/src/veritysetup/veritysetup.c
+++ b/src/veritysetup/veritysetup.c
@@ -17,7 +17,9 @@
 #include "terminal-util.h"
 
 static uint32_t arg_activate_flags = CRYPT_ACTIVATE_READONLY;
-static const char *arg_root_hash_signature = NULL;
+static char *arg_root_hash_signature = NULL;
+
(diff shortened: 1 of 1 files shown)


src/veritysetup/veritysetup.c
```diff
@@ -45,7 +47,9 @@ static int save_roothashsig_option(const char *option, bool strict) {
                         return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP),
                                                "Activation of verity device with signature requested, but cryptsetup does not support crypt_activate_by_signed_key().");
 
-                arg_root_hash_signature = option;
+                if (free_and_strdup(&arg_root_hash_signature, option) < 0)
```
yuwata: `free_and_strdup_warn()` can be used.

poettering: indeed!

bluca (approved):

yuwata (approved):
LGTM.

Issue #22709, opened 2022-03-10
systemd-detect-virt: Incorrectly identifies KVM VMs when using hv_passthrough


holmanb:
**systemd version the issue has been seen with**
249.9

**Used distribution**
Ubuntu Jammy (22.04)

**Linux kernel version used** (`uname -a`)
`Linux ubuntu 5.15.0-18-generic #18-Ubuntu SMP Fri Jan 21 14:57:54 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**CPU architecture issue was seen on**
x86_64

**Expected behaviour you didn't see**
`systemd-detect-virt` should identify KVM-accelerated VMs as "kvm" or something indicating that KVM is in use

**Unexpected behaviour you saw**
`systemd-detect-virt` identifies KVM-accelerated VMs as "qemu" when `hv_passthrough` is used

**Steps to reproduce the problem**
```
#!/bin/sh

wget https://cloud-images.ubuntu.com/jammy/current/jammy-server-cloudimg-amd64.img

# cloud-config sets user password
cat >user-data <<EOF
#cloud-config
password: password
chpasswd:
  expire: False
EOF

# requires package cloud-utils, adds user/password to cloud image
cloud-localds seed.img user-data

# The second invocation uses hv_passthrough, which causes systemd-detect-virt to report "qemu" as the virtualization method
# Log in and execute `systemd-detect-virt`
#
# Moniker: ubuntu
# Password: password
#

# Works as expected:
qemu-system-x86_64 -net nic -net user -machine accel=kvm,type=q35 -cpu host -m 512 -nographic -hda jammy-server-cloudimg-amd64.img -hdb seed.img

# Reports "qemu" as virtualization type
qemu-system-x86_64 -net nic -net user -machine accel=kvm,type=q35 -cpu host,hv_passthrough -m 512 -nographic -hda jammy-server-cloudimg-amd64.img -hdb seed.img
```

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
# without hv_passthrough:

ubuntu@ubuntu:~$  SYSTEMD_LOG_LEVEL=debug systemd-detect-virt
Failed to read $container of PID 1, ignoring: Permission denied
Found cgroup2 on /sys/fs/cgroup/, full unified hierarchy
Found container virtualization none.
Virtualization QEMU found in DMI (/sys/class/dmi/id/sys_vendor)
UML virtualization not found in /proc/cpuinfo.
[...]

PR #22945, opened 2022-04-01
Identify kvm + hv_passthrough as "kvm"

holmanb:
This PR proposes a "fix" for https://github.com/systemd/systemd/issues/22709, assuming that desired behavior is to report "kvm". If a unique output ("kvm hv_passthrough" maybe?) is preferred, I will update accordingly.

I don't see tests for detect-virt in the codebase so I haven't tried to include any.

The following output based on the latest `main` head: bd4dea76de6355778079ba4945846324cef2a043
```
ubuntu@ubuntu:~/systemd$ git log HEAD
commit c0fc2d9c41dee04108f076fbc857ddc213931d42 (HEAD -> holmanb/hv, origin/holmanb/hv)
Author: Brett Holman <bholman.devel@gmail.com>
Date:   Thu Mar 31 14:04:57 2022 -0600

    Identify kvm + hv_passthrough as "kvm"

commit bd4dea76de6355778079ba4945846324cef2a043
Author: Lennart Poettering <lennart@poettering.net>
Date:   Fri Apr 1 17:01:29 2022 +0200

    veritysetup: fix memory corruption
    
    We must copy the option string, since in one case we are called with a
    pointer into dynamic memory that will be freed by the caller.
    
    As discussed here: https://github.com/systemd/systemd/pull/22908/files#r839394490
    
    Follow-up for: #22908

commit 95cd21928fae6e1806176cd56853cc1e490607ff
Merge: b17a681be5 798931160e
ubuntu@ubuntu:~/systemd$ meson compile -C build
ninja: Entering directory `/home/ubuntu/systemd/build'
[1/491] Generating version.h with a custom command
ubuntu@ubuntu:~/systemd$ ./build/systemd-detect-virt
kvm
ubuntu@ubuntu:~/systemd$ SYSTEMD_LOG_LEVEL=debug ./build/systemd-detect-virt
Failed to read $container of PID 1, ignoring: Permission denied
Virtualization QEMU found in DMI (/sys/class/dmi/id/sys_vendor)
UML virtualization not found in /proc/cpuinfo.
Virtualization XEN not found, /proc/xen does not exist
Virtualization found, CPUID=Linux KVM Hv
Found VM virtualization kvm
kvm
```


yuwata (changes requested):
LGTM, but please drop the `virtualization_table` in test-table.c.

holmanb:
> LGTM, 

Thanks for the review.

> but please drop the `virtualization_table` in test-table.c.

Done

yuwata (approved):

poettering:
Hmm, should this really be a separate entry in the enum? I am not sure. Why would you think so? Isn't this just a minor flavour of regular kvm?

I have no idea what "hv_passthrough" actually is, can you elaborate?

If a new entry is really the right choice here (which from my perspective doesn't look like it), then this is missing doc updates, specifically an entry in the table in man/systemd-detect-virt.xml)

poettering:
> LGTM, but please drop the `virtualization_table` in test-table.c.

Why that?

poettering:
Oh, I understand now, this PR works by mapping two enums to the same string "kvm". I really don't like that I must say to have more enum values that mapped strings. Also, once any of our code starts using virtualization_from_string() things will fall apart, as the mapping then isn't unique.

Please just rework the `vm_table[]` array to be like `dmi_vendor_table[]`, i.e. a series of structs that pair the type and string. It's probably a better idea anyway, since we don#t need the indexing by type anyway, and the array as it stands now is mostly NULL pointers already.

poettering:
To speed things I implemented what I proposed now in #22952. @holmanb any chance you can verify this still works for you?

holmanb:
Presumably I'm in a different time zone, as I'm just seeing all of this now. Thanks for working on this.

> I have no idea what "hv_passthrough" actually is, can you elaborate?

In all honesty I'm not sure why one would use it for a Linux guest. My understanding is that `hv_passthrough` is an entitlement for accelerating Windows guests. I came across it because [LXC uses it on kernels >5.10](https://github.com/lxc/lxd/blob/173c1dcc6612a80d8302b3ebef3b9434af79aa9c/lxd/instance/drivers/driver_qemu.go#L1251), which caused an unexpected behavior change in cloud-init (which has support for lxc). 

> Oh, I understand now, this PR works by mapping two enums to the same string "kvm".

Correct.

> I really don't like that I must say to have more enum values that mapped strings. Also, once any of our code starts using virtualization_from_string() things will fall apart, as the mapping then isn't unique.

Oops, we don't want that obviously.

> To speed things I implemented what I proposed now in #22952. @holmanb any chance you can verify this still works for you?

I tested your reworked version and it works as expected (tested on main branch, which is currently on 666efe94c998dad2b94be9b7cf471071):
```
ubuntu@ubuntu:~/systemd$ ./build/systemd-detect-virt
kvm
ubuntu@ubuntu:~/systemd$ SYSTEMD_LOG_LEVEL=debug ./build/systemd-detect-virt
Failed to read $container of PID 1, ignoring: Permission denied
Virtualization QEMU found in DMI (/sys/class/dmi/id/sys_vendor)
UML virtualization not found in /proc/cpuinfo.
Virtualization XEN not found, /proc/xen does not exist
Virtualization found, CPUID=Linux KVM Hv
Found VM virtualization kvm
kvm
```
Thanks again!

keszybz:
Note to self: this is already in v251, but needs a backport to v250.

PR #22951, opened 2022-04-04
Fix entry selection in bootctl status

keszybz:


src/boot/bootctl.c
```diff
@@ -682,12 +711,12 @@ static int boot_entry_show(
 }
 
 static int status_entries(
+                const BootConfig *config,
```
poettering, line 714: add an assert for this?

src/boot/bootctl.c
```diff
@@ -709,21 +738,13 @@ static int status_entries(
                        SD_ID128_FORMAT_VAL(dollar_boot_partition_uuid));
         printf("\n\n");
 
-        r = boot_config_load(&config, esp_path, xbootldr_path);
-        if (r < 0)
-                return r;
-
-        r = boot_config_select_special_entries(&config);
-        if (r < 0)
-                return r;
```
poettering, line 718: hmm, why not call boot_config_load_and_select() from here?

(I mean, i am not opposed to moving this, I am just wondering, given that `status_entries()` only has a single caller)

keszybz: I wanted to keep the loading and printing separate. Looking at it again, either option should work about as well…

keszybz:
centois (arch linux + sanitizers): ninja-test_sanitizers → 385/1297 test-sd-device   SIGABRT.
Doesn't look related, but it's suspious anyway.

I'll update with the added assert.

PR #22952, opened 2022-04-04
virt: tweak kvm with hyperv enlightenments

poettering:
Implements what I suggested here:

https://github.com/systemd/systemd/pull/22945#issuecomment-1087255741

PR #22955, opened 2022-04-04
detect-virt: make virtualization table entries unique again

keszybz:
fbe043001581df609eae40e0d2f2319d52b5237a wanted to add a new match pattern for
VIRTUALIZATION_KVM, but because the code was using _to_string(), it just added
a duplicate enum entry. I think that this is confusing, because we'd have two
entries that are displayed the same, and we'd need to remember to match for
both in relevant places.

Instead, let's just invert the match table to allow multiple matches that map
to the same enum value, and make the enum values unique again.

This inverts most of the changes in fbe043001581df609eae40e0d2f2319d52b5237a,
but not the important part, i.e. the addition of a new match pattern.


bluca:
conflicts with https://github.com/systemd/systemd/pull/22952 ?

yuwata:
Yeah, #22952 is faster.

PR #22956, opened 2022-04-04
network: ignore all errors in loading .network files

### LINKED ISSUE #22954: Since v250 systemd-networkd aborts if it don't have permissions to read one of the .network conf files
More accurately this behavior change was introduced by 9202b567bcdd0c1f6a1fc2a5f36602e619960813.

@yuwata can you please tell if this is intended ?

I'm not sure whether networkd should fail when such condition happen, at least this wasn't the case before.

Thanks.

yuwata:
Fixes a bug introduced by 9202b567bcdd0c1f6a1fc2a5f36602e619960813.

Fixes #22954.

### DIFF
--- a/src/udev/net/link-config.c
+++ b/src/udev/net/link-config.c
@@ -218,10 +218,8 @@ int link_load_one(LinkConfigContext *ctx, const char *filename) {
         assert(filename);
 
         r = null_or_empty_path(filename);
-        if (r == -ENOENT)
-                return 0;
         if (r < 0)
-                return r;
+                return log_warning_errno(r, "Failed to check if \"%s\" is empty: %m", filename);
         if (r > 0) {
                 log_debug("Skipping empty file: %s", filename);
                 return 0;
@@ -229,11 +227,11 @@ int link_load_one(LinkConfigContext *ctx, const char *filename) {
 
         name = strdup(filename);
         if (!name)
-                return -ENOMEM;
+                return log_oom();
 
         config = new(LinkConfig, 1);
         if (!config)
-                return -ENOMEM;
+                return log_oom();
 
         *config = (LinkConfig) {
                 .filename = TAKE_PTR(name),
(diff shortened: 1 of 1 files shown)


src/network/netdev/netdev.c
```diff
@@ -873,8 +873,8 @@ int netdev_load(Manager *manager, bool reload) {
 
         STRV_FOREACH(f, files) {
                 r = netdev_load_one(manager, *f);
-                if (r < 0)
-                        log_error_errno(r, "Failed to load %s, ignoring: %m", *f);
+                if (r < 0 && !ERRNO_IS_PRIVILEGE(r))
+                        return log_error_errno(r, "Failed to load %s: %m", *f);
```
bluca: maybe keep the log message for the other cases, with 'ignoring', at debug level?

bluca (approved):

src/network/netdev/netdev.c
yuwata: I thought so, but `config_parse()` already shows a log message in that case. Hence, it is not necessary, I think.

bluca: ok, makes sense then, no need to duplicate

src/network/networkd-network.c
```diff
@@ -592,7 +592,7 @@ int network_load(Manager *manager, OrderedHashmap **networks) {
 
         STRV_FOREACH(f, files) {
                 r = network_load_one(manager, networks, *f);
-                if (r < 0)
+                if (r < 0 && !ERRNO_IS_PRIVILEGE(r))
                         return log_error_errno(r, "Failed to load %s: %m", *f);
```
poettering: hmm, so something's afoot regarding logging here. According to our usual coding style functions should either log about all errors or about none (in which case the caller should do that). 

Previously we'd log here about all errors, thus implying that network_load_one() would not log about any (which is not correct, is it?).

With the new code we'll now log about some errors, but not about others, which seems to contradict the rule quite a bit.

i.e. it appears to me there are two issues here: 

1. we should clean up logging so that every error is logged once and once only
2. we should eat up ERRNO_IS_PRIVILEGE() errors after LOG_WARNING about them, so that we continue running if we encounter an issue like that.

poettering: actually, I wonder if it wouldn't be more appropriate to continue running regardless of any error with loading configuration files. i.e. if some unexpected error happens while loading some .netdev it should not take down networkd as a whole and other netdevs...

i.e. the change should cover more than ERRNO_IS_PRIVILEGE, no?

yuwata:
@poettering Thank you for the comments. Now all errors in loading .network files are ignored. And error messages are deduplicated. PTAL.

yuwata:
Please backport the first commit into v250-stable.

keszybz:
Looks reasonable.

PR #22959, opened 2022-04-04
tree-wide: use 'Architecture' type consistently across the tree

poettering:
Some parts of our tree used 'Architecture' for storing architectures,
others used ints. Let's unify on the former.

Inspired by #22952's rework of the 'Virtualization' enum.

### DIFF
--- a/src/basic/architecture.h
+++ b/src/basic/architecture.h
@@ -48,7 +48,7 @@ typedef enum {
         _ARCHITECTURE_INVALID = -EINVAL,
 } Architecture;
 
-int uname_architecture(void);
+Architecture uname_architecture(void);
 
 /*
  * LIB_ARCH_TUPLE should resolve to the local library path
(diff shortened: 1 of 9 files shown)


src/basic/architecture.h
```diff
@@ -3,6 +3,7 @@
 
 #include <endian.h>
 
+#include "errno-list.h"
```
bluca: what's the reason for adding this include?

src/basic/architecture.h
```diff
@@ -48,7 +49,7 @@ typedef enum {
         _ARCHITECTURE_INVALID = -EINVAL,
```
yuwata, line 48: Maybe, it is better to include -ERRNO_MAX?

poettering: the functions returning `Architecture` cannot fail actually, right now. (I mean only architecture_to_string() can, but only via the `_ARCHITECTURE_INVALID`/`-EINVAL` special case). Hence, following our usual rule of adding a `-ERRNO_MAX` entry only where we actually need it I didn't add it here.

bluca (approved):

PR #22962, opened 2022-04-04
man: properly conditionalize kernel-install man page

poettering:
Change f887eab1da85d0053321d43228042d90720eb77e conditionalized the
building of ther kenel-install man page in the generated meson output,
instead of the source in the XML markup. Thus, whenever the rules file
is updated the conditionalization is lost. Correct that.

Follow-up for: f887eab1da85d0053321d43228042d90720eb77e


bluca (approved):

keszybz:
Oops.

PR #22963, opened 2022-04-04
sd-device diskseq fix + udevadm info improvements

poettering:
We weren't actually reading the diskseq from the uevent file, hence fix that. While debugging I figured we should show more info in "udevadm info" and also document hte involveed APIs a bit.


src/udev/udevadm-info.c
```diff
@@ -171,27 +172,77 @@ static int print_device_chain(sd_device *device) {
...
         assert(device);
 
-        (void) sd_device_get_devpath(device, &str);
-        printf("P: %s\n", str);
+        /* We don't show syspath here, because it's identical to devpath (modulo the "/sys" prefix).
+         *
+         * We don't show action/seqnum here because that only makes sense for records synthesized from
+         * uevents, not for those synthesized from database entries.
+         *
+         * We don't show sysattrs here, because they can be expensive and potential issue expensive driver
```
mrc0mmand:
```suggestion
         * We don't show sysattrs here, because they can be expensive and potentially issue expensive driver
```

src/udev/udevadm-info.c
```diff
@@ -171,27 +172,77 @@ static int print_device_chain(sd_device *device) {
...
+         * IO
+         *
+         * Coloring: let's be conservative with coloring. Let's use it to group related fields. Right now:
+         *
+         *     • white for fields that give the device a name
+         *     • green for fields that categorize the device into subsystem/devtype and similar
+         *     • cyan for fields about associated device nodes/sysmlinks/network interfaces and such
```
mrc0mmand:
```suggestion
         *     • cyan for fields about associated device nodes/symlinks/network interfaces and such
```

src/udev/udevadm-info.c
```diff
@@ -171,27 +171,65 @@ static int print_device_chain(sd_device *device) {
...
 
         assert(device);
 
-        (void) sd_device_get_devpath(device, &str);
+        /* We don't show syspath here, because it's identical to devpath (modulo the "/sys" prefix).
+         *
+         * We don't show action/seqnum here because that only makes sense for records synthesized from
+         * uevents, not for those synthesized from database entries.
+         *
+         * We don't show sysattrs here, because they can be expensive and potential issue expensive driver
```
bluca: /potential/potentially/

src/udev/udevadm-info.c
```diff
@@ -584,14 +635,21 @@ int info_main(int argc, char *argv[], void *userdata) {
...
                 _cleanup_(sd_device_unrefp) sd_device *device = NULL;
 
                 r = find_device(*p, NULL, &device);
-                if (r == -EINVAL)
-                        return log_error_errno(r, "Bad argument \"%s\", expected an absolute path in /dev/ or /sys or a unit name: %m", *p);
-                if (r < 0)
-                        return log_error_errno(r, "Unknown device \"%s\": %m",  *p);
+                if (r < 0) {
+                        if (r == -EINVAL)
+                                log_error_errno(r, "Bad argument \"%s\", expected an absolute path in /dev/ or /sys or a unit name: %m", *p);
```
mrc0mmand:
```suggestion
                                log_error_errno(r, "Bad argument \"%s\", expected an absolute path in /dev/ or /sys/ or a unit name: %m", *p);
```

bluca (approved):

man/sd_device_get_syspath.xml
```diff
@@ -0,0 +1,200 @@
...
+    returns the device number with major/minor <literal>4:7</literal>.</para>
+
+    <para><function>sd_device_get_ifindex()</function> returns the network interface index of the specified
+    device record, if the device encapsulates a network interface device, i.e. belongs to the
+    <literal>net</literal> subsystem. Example: the <literal>lo</literal> interface typically has interface
+    index 1.</para>
+
+    <para><function>sd_device_get_driver()</function> returns the kernel driver name attached to the
+    device. Note that the driver field is set on the devices consumed by the driver, not on the device
+    created by it. Example: the a PCI device <filename>/sys/bus/pci/devices/0000:00:1f.6</filename> might be
```
mrc0mmand:
```suggestion
    created by it. Example: a PCI device <filename>/sys/bus/pci/devices/0000:00:1f.6</filename> might be
```

bluca:
> sd_device_get_diskseq() actually called device_read_uevent_file() but
that function didn't actually parse DISKSEQ= so far. Fix that.

wait, was it all borken then?

man/udevadm.xml
```diff
@@ -239,6 +263,14 @@
               <entry><literal>S:</literal></entry>
               <entry>Device node symlink</entry>
             </row>
+            <row>
+              <entry><literal>Q:</literal></entry>
+              <entry>Block device sequence sequence number (DISKSEQ)</entry>
```
mrc0mmand:
```suggestion
              <entry>Block device sequence number (DISKSEQ)</entry>
```

bluca:
> > wait, was it all borken then?
> 
> no, just some parts. i.e. when getting an sd_device object via netlink all was OK. But when getting it independently from that (i.e. loading it off udev db) it didn#t get loaded...

ah, phew!

poettering:
Force pushed a new version, with @mrc0mmand's points addressed. Upgrading green label

src/udev/udevadm-info.c
```diff
@@ -171,27 +171,65 @@ static int print_device_chain(sd_device *device) {
...
         if (sd_device_get_devname(device, &str) >= 0) {
                 assert_se(val = path_startswith(str, "/dev/"));
                 printf("N: %s\n", val);
-        }
+        } else
+                str = NULL;
 
-        if (device_get_devlink_priority(device, &i) >= 0)
-                printf("L: %i\n", i);
+        if (str) {
```
yuwata: It is not necessary to assign `str = NULL;`.

yuwata (approved):
LGTM. But one minor request.

src/udev/udevadm-info.c
poettering: it is, because we otherwise will see the `str` initialization from further up, i.e. in sd_device_get_devtype() and earlier calls hence we must explicitly reset it to NULL if we want to use it as a checker for "success".

yuwata: I mean this can be like the following.
```
        if (sd_device_get_devname(device, &str) >= 0) {
                assert_se(val = path_startswith(str, "/dev/"));
                printf("N: %s\n", val);

                /* Don't bother with devlink priority if we don't have any node in /dev/ anyway */
                if (device_get_devlink_priority(device, &i) >= 0)
                        printf("L: %i\n", i);

                ...
        }
```

poettering: oh indeed! will fix

poettering: done

PR #22965, opened 2022-04-04
oomd: calculate 'used' memory with MemAvailable instead of MemFree

enr0n:
The calculation for used memory in oomd_system_context_acquire is given
by MemTotal - MemFree from /proc/meminfo. This is too strict of a
calculation because it does not consider memory that is still available
for starting new applictions without swapping (MemAvailable). As a
result, systemd-oomd can start to kill processes before it is necessary.
This is more apparent on systems with low swap space.

Instead, compute 'used' memory as MemTotal - MemAvailable in
oomd_system_context_acquire and procfs_memory_get (which is used by
oomd_cgroup_context_acquire). And, rename oomd_mem_free_below to
oomd_mem_available_below for clarity.

### DIFF
--- a/src/basic/procfs-util.c
+++ b/src/basic/procfs-util.c
@@ -219,7 +219,7 @@ int convert_meminfo_value_to_uint64_bytes(const char *word, uint64_t *ret) {
 }
 
 int procfs_memory_get(uint64_t *ret_total, uint64_t *ret_used) {
-        uint64_t mem_total = UINT64_MAX, mem_free = UINT64_MAX;
+        uint64_t mem_total = UINT64_MAX, mem_available = UINT64_MAX;
         _cleanup_fclose_ FILE *f = NULL;
         int r;
 
@@ -242,9 +242,9 @@ int procfs_memory_get(uint64_t *ret_total, uint64_t *ret_used) {
                 if (p)
                         v = &mem_total;
                 else {
-                        p = first_word(line, "MemFree:");
+                        p = first_word(line, "MemAvailable:");
                         if (p)
-                                v = &mem_free;
+                                v = &mem_available;
                         else
                                 continue;
                 }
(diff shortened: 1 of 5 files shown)


julian-klode:
The semaphoreci/pr seems to be a networking/keyserver issue, afaict.

julian-klode:
There are two points I want to point out so they don't get lost:

1. earlyoom also uses available memory instead of free memory to calculate used memory
2. The cgroup's memory usage still includes the page cache as we read `memory.current`, this seems somewhat suboptimal. I'm not sure how much that is an issue, as I think we order cgroups to kill by swap usage (or PSI). To fix this, one would have to read `memory.stat` which is a bit more involved; and does not align with the limits we read from `.max` and `.low`, as the approach to include page caches makes sense from a memory containment perspective, but not a global "should this cgroup be killed" perspective.

yuwata:
cc @anitazha

anitazha:
> 2. The cgroup's memory usage still includes the page cache as we read `memory.current`, this seems somewhat suboptimal. I'm not sure how much that is an issue, as I think we order cgroups to kill by swap usage (or PSI). To fix this, one would have to read `memory.stat` which is a bit more involved; and does not align with the limits we read from `.max` and `.low`, as the approach to include page caches makes sense from a memory containment perspective, but not a global "should this cgroup be killed" perspective.
> 

This change should only affect the "swap kill" behavior a bit and w.r.t. there shouldn't be an issue as only `memory.swap.current` is used there. For memory pressure kills we primarily rely on `pgscan` from `memory.stat` as part of picking cgroups during memory pressure kills. We do read and store `memory.current` but generally it doesn't come into play if reclaim rate is non-zero.

anitazha (approved):

MarkJeronimus:
I had a crash just now and I think it's related to this issue. The crash happened when I started a large application. It was taking longer than expected and the disk was crunching a lot. I peered at my system monitor graphs and one thing I noticed was that real memory was about 50% filled, the other 50% filled with page cache. This is nothing out of the ordinary: it literally always looks like this on every pc and every distro I use. Even now, 30 minutes after restart as I'm writing this.

After restart, I found these lines in the log:

> Sep 17 09:48:11 confucius systemd-oomd[1097]: Killed /user.slice/user-1000.slice/session-3.scope due to memory used (16340660224) / total (16675287040) and swap used (15186862080) / total (16835928064) being more than 90.00%

> Sep 17 09:48:12 confucius systemd[1]: session-3.scope: systemd-oomd killed 381 process(es) in this unit.

Confusingly, my log contained a few of these lines beforehand:

> Sep 17 09:28:52 confucius pcp-pmie[2298]: Severe demand for real memory 366pgsout/s@confucius
Sep 17 09:36:52 confucius pcp-pmie[2298]: High aggregate context switch rate 101826ctxsw/s@confucius

PR #22968, opened 2022-04-05
Recommend kernel 4.15 or newer

keszybz:


bluca (approved):

README
```diff
@@ -30,14 +30,18 @@ LICENSE:
...
+                     ≥ 4.6 for cgroup namespaces
+                     ≥ 4.10 for cgroup-bpf egress and ingress hooks
+                     ≥ 4.15 for cgroup-bpf device hook and cpu controller in cgroup v2
+                     ≥ 4.17 for cgroup-bpf socket address hooks
+                     ≥ 5.3 for bounded-loops in BPF program
+                     ≥ 5.4 for signed Verity images support
+                     ≥ 5.7 for BPF links and the BPF LSM hook
+
+        Versions below 4.15 have significant gaps in functionality and are not
+        recommended for use with this version of systemd.
```
poettering: if this is changed, also needs updating of `KERNEL_BASELINE_VERSION` in src/basic/def.h

src/basic/fs-util.c
```diff
@@ -75,46 +75,6 @@ int rmdir_parents(const char *path, const char *stop) {
...
 }
 
-int rename_noreplace(int olddirfd, const char *oldpath, int newdirfd, const char *newpath) {
-        int r;
-
-        /* Try the ideal approach first */
-        if (renameat2(olddirfd, oldpath, newdirfd, newpath, RENAME_NOREPLACE) >= 0)
-                return 0;
-
-        /* renameat2() exists since Linux 3.15, btrfs and FAT added support for it later. If it is not implemented,
```
poettering: hmm, iirc renameat2() support was added much later than 3.15 in some file systems. We shouldn't drop the compat kludge here unless we checked that all relevant file systems support it these days.

poettering:
vfat got renameat2() RENAME_NOREPLACE support only in kernel 4.9, according to the man page.

I wonder if nfs/smb have it at all?

poettering:
so afaics nfs4 supports neither RENAME_EXCHANGE nor RENAME_NOREPLACE. I figure we have to keep the compat kludge :-(

keszybz:
Updated.

> Also needs update in NEWS

I'm leaving NEWS for an asynchronous update before the release. I may be wrong, but I think this PR is not the last word on the subject ;)

bluca (approved):

yuwata (approved):

keszybz:
ci (centos_epel, 8-stream)   - nothing provides libmount = 2.32.1-35.el8 needed by libmount-devel-2.32.1-35.el8.x86_64

logo semaphore Blazing-fast build and deploy! + ip route list 0/0 56:34
+ [ -z  ] 56:34
+ sleep 1 56:37
+ ip route list 0/0

Both look unrelated.

PR #22969, opened 2022-04-05
udevadm: add new "udevadm info --tree" command

poettering:
This shows the device hierarchy as a tree, i.e. as what it actually is.

### DIFF
--- a/TODO
+++ b/TODO
@@ -164,9 +164,6 @@ Features:
   sd_device object, so that data passed into sd_device_new_from_devnum() can
   also be queried.
 
-* udevadm: a new "tree" verb that shows tree of devices as syspath hierarchy,
-  along with their properties. uninitialized devices should be greyed out.
-
 * bootctl: show whether UEFI audit mode is available
 
 * sd-event: optionally, if per-event source rate limit is hit, downgrade


src/udev/udevadm-info.c
```diff
@@ -486,6 +497,201 @@ static int help(void) {
...
+                r = sd_device_enumerator_add_match_parent(e, below);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to install parent enumerator match: %m");
+        }
+
+        r = sd_device_enumerator_allow_uninitialized(e);
+        if (r < 0)
+                return log_error_errno(r, "Failed to enable enumeration of uninitialized devices: %m");
+
+        FOREACH_DEVICE(e, child) {
```
yuwata: This may not show e.g. driver device??

yuwata (commented):
The logic of the last commit looks complicated for me. And IIUC the comment in the code, I think it is better to fix the enumerator to actually enumerate all devices??
The enumerator now sorts the devices which mostly fits for showing in tree format.

poettering:
As discussed elsewhere this new rework includes any sysfs kobjects that aren't linked from /sys/class or /sys/bus also in the enumerations if they are parent devices of devices found that way.

Various other fixes.

src/udev/udevadm-info.c
```diff
@@ -486,6 +497,156 @@ static int help(void) {
...
+                if (r < 0)
+                        return r;
+
+                i = j;
+        }
+
+        return 0;
+}
+
+static int print_tree(sd_device* below, const char *prefix, unsigned level) {
```
yuwata: The second and third arguments are never used.

poettering:
fixed the issue @yuwata pointed out.

yuwata:
Hm, all full-task CIs are red...

poettering:
ah, the pitfalls of log_trace()...

poettering:
Ci should now be fixed. ptal

yuwata (approved):
LGTM.

PR #22730, opened 2022-03-14
Add AV production devices to hwdb

GeorgesStavracas:
Builds on top of https://github.com/systemd/systemd/pull/20561

This adds support for multimedia devices, including everything in https://github.com/systemd/systemd/pull/20561, but also Elgato Stream Deck devices.

I did not apply any of the corrections or comments from https://github.com/systemd/systemd/pull/20561, but will be happy to take it over and push this to the finish line if @Be-ing agrees.

CC @hadess

PR #22971, opened 2022-04-05
Add Logitech receivers and devices to hwdb and add uaccess

hadess:
This adds support for Logitech USB receivers and Bluetooth devices to be
accessed by users at the console, so that they can use an application
like Solaar out-of-the-box.

Note, this is currently untested, I'm not sure about the property name for this one.

### DIFF
--- a/hwdb.d/70-input-devices.hwdb
+++ b/hwdb.d/70-input-devices.hwdb
@@ -1,0 +1,27 @@
+# This file is part of systemd.
+#
+# Database for input devices that need hidraw access for all the features to work.
+#
+# This currently only covers Logitech receivers (Bolt, Unifying, etc.) and Bluetooth
+# Logitech devices.
+#
+# To add local entries, copy this file to
+#   /etc/udev/hwdb.d/
+# and add your rules there. To load the new rules execute (as root):
+#   systemd-hwdb update
+#   udevadm trigger
+
+################
+# Logitech
+################
+# USB-connected Logitech receivers and devices
+usb:v046d*
+ ID_INPUT_DEVICE_NEEDS_RAW=1
+
+# Lenovo Nano Receiver
+usb:v17efp6042*
+ ID_INPUT_RECEIVER_NEEDS_RAW=1
+
+# Bluetooth-connected Logitech devices
+bluetooth:v046d*
+ ID_INPUT_RECEIVER_NEEDS_RAW=1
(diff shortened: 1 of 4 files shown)


poettering:
hmm, but that means we turn off session switching for logitech devices, which sounds like quite a major security issue...

poettering:
it appears to me we should revisit this whole uaccess thing on hidraw thing and maybe go another way: add a dbus api to logind to open these for clients. Then put a polkit auth in the middle. That way polkit could make some smart decision here, i.e. verify by password that app xyz shall get raw access to a device...

hadess:
Yeah, there's no revoke support for hidraw, or USB devices, both of which are important things to have. I guess that those could also be problematic for game controllers (#22860) or other input devices (#22730) but it's certainly much less effort extracting info from somebody's keyboard events than anything else.

But do note that this is already a problem for anyone using the Solaar Logitech receiver management app that allows things like pairing and unpairing devices from the receiver: https://github.com/pwr-Solaar/Solaar/blob/master/rules.d/42-logitech-unify-permissions.rules

keszybz:
I don't think we can do this… I just checked a usb logitech keyboard+mouse combo with a unified receiver, and it has `046d:c534`, so it'd be covered by this rule.

hadess:
Right, somebody will need to implement revoke support for hidraw to be able to do this. I'll send a mail to the hid list soon.

hadess:
> Right, somebody will need to implement revoke support for hidraw to be able to do this. I'll send a mail to the hid list soon.

@whot got started on that:
https://github.com/whot/linux/tree/wip/hid-revoke
https://github.com/whot/systemd/tree/wip/logind-hidraw

keszybz:
I'll mark this as `postponed` for lack of a better label, since we're blocked on kernel changes.

bluca:
What's the status here? @hadess could you move this branch to your fork? We don't use members' branches in this repo so it looks out of place

YHNdnzj:
Given the lack of response, and that the existing approach is not feasible, I'm tempted to just close this. The linked branches haven't really been updated for quite some time either.

If there's interest in the future, and the kernel patch is ready, we can always revisit this. But carrying the WIP branch here forever isn't something we want.

whot:
I won't have time to work on this in the near future, but the basic summary is: we had a detour of using BPF for hidraw revoke, eventually realised that's way over scale and the current plan is to go with the above approach of the ioctl. Once someone finds the time to submit this for the kernel...

YHNdnzj:
With https://github.com/systemd/systemd/pull/33970#issuecomment-2305216088 we should finally be able to close this.

Issue #22936, opened 2022-04-01
Sleep key can't be ignored


jolmg:
**systemd version the issue has been seen with**

> 250.4

**Used distribution**
 
> Archlinux

**Linux kernel version used** (`uname -a`)

 > Linux aspero 5.16.16-arch1-1 #1 SMP PREEMPT Mon, 21 Mar 2022 22:59:40 +0000 x86_64 GNU/Linux

**CPU architecture issue was seen on**
 > x86-64

**Expected behaviour you didn't see**
 > To not suspend when `HandleSuspendKey=ignore` and `systemd-logind[16899]: Suspend key pressed short.`

**Unexpected behaviour you saw**
 > It suspended when short pressing the suspend key, despite having `HandleSuspendKey=ignore` in /etc/systemd/logind.conf

**Steps to reproduce the problem**
 > 1. Set HandleSuspendKey=ignore in /etc/systemd/logind.conf
 > 2. reboot
 > 3. short-press suspend key

**Additional program output to the terminal or log subsystem illustrating the issue**

Here is the journal output of systemd with log level "debug" from the moment logind reports the sleep button press, to when the suspension starts:

https://gist.github.com/jolmg/73dbd0a0d706434bff9fd682695c26e3

Additionally,

```
$ systemd-analyze cat-config systemd/logind.conf
# /etc/systemd/logind.conf
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it under the
#  terms of the GNU Lesser General Public License as published by the Free
#  Software Foundation; either version 2.1 of the License, or (at your option)
#  any later version.
#
# Entries in this file show the compile time defaults. Local configuration
# should be created by either modifying this file, or by creating "drop-ins" in
# the logind.conf.d/ subdirectory. The latter is generally recommended.
# Defaults can be restored by simply deleting this file and all drop-ins.
#
# Use 'systemd-analyze cat-config systemd/logind.conf' to display the full config.
#
# See logind.conf(5) for details.

[Login]
#NAutoVTs=6
#ReserveVT=6
#KillUserProcesses=no
#KillOnlyUsers=
#KillExcludeUsers=root
#InhibitDelayMaxSec=5
#UserStopDelaySec=10
HandlePowerKey=ignore
HandleSuspendKey=ignore
HandleSuspendKeyLongPress=ignore
#HandleHibernateKey=hibernate
HandleLidSwitch=ignore
HandleLidSwitchExternalPower=ignore
[...]

PR #22975, opened 2022-04-05
logind: "long press" follow-up

poettering:
Seems we forgot some bits in a520bb665417af7aa98dcb983f3583659bbab807. Let's add this.

See: #22936


yuwata (approved):
LGTM.

PR #22977, opened 2022-04-05
network: drop support for old kernels which cannot set prefix route with non-main routa table

yuwata:
Not sure when the issue was fixed.
- kernel-3.10 on CentOS 7 has the issue,
- kernel-4.18 on CentOS 8 works fine.

Note, the workaround dropped by the commit is not incomplete:
with an old kernel which has the issue, all non-prefix routes are
configured on the specified route table, but the prefix route is
configured on the main table. That should not work for most cases,
hence, the workaround is mostly meaningless.

### DIFF
--- a/src/network/networkd-dhcp4.c
+++ b/src/network/networkd-dhcp4.c
@@ -27,8 +27,6 @@
 #include "strv.h"
 #include "sysctl-util.h"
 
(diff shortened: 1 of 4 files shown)


keszybz:
I think this is reasonable, but I'd prefer to do it after the next release, so there's more time for this to "bake".

bluca (approved):

PR #22981, opened 2022-04-05
udev: cleanups for device node handling

yuwata:


yuwata:
BTW, what is the usecase of static_node-tags ??

PR #22983, opened 2022-04-05
login: shorten code a bit

yuwata:
The symlinks under /run/udev/static_node-tags are pointing to the
corresponding device nodes. Let's read the symlinks.

### DIFF
--- a/src/shared/devnode-acl.c
+++ b/src/shared/devnode-acl.c
@@ -9,9 +9,9 @@
 #include "device-util.h"
 #include "devnode-acl.h"
 #include "dirent-util.h"
-#include "escape.h"
 #include "fd-util.h"
 #include "format-util.h"
+#include "fs-util.h"
 #include "set.h"
 #include "string-util.h"
 #include "util.h"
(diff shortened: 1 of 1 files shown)


src/shared/devnode-acl.c
```diff
@@ -195,21 +195,12 @@ int devnode_acl_all(const char *seat,
...
-                        ssize_t l;
-
-                        l = cunescape(de->d_name, UNESCAPE_RELAX, &unescaped_devname);
-                        if (l < 0)
-                                return l;
-
-                        n = path_join("/dev", unescaped_devname);
-                        if (!n)
-                                return -ENOMEM;
+                        r = readlinkat_malloc(dirfd(dir), de->d_name, &n);
```
poettering, line 198: maybe skip ENOENT here, in case this code runs in paralell with udev's logic and an entry that was still in the directory moments ago is not anymore by the time we try to read it.

In fact, I'd probably do it like this: 

```c
r = readlink_malloc(dirfd(dir), de->d_name, &n);
if (r == -ENOENT)
        continue;
if (r < 0) {
        log_debug_errno("Unable to read symlink '/run/udev/static_node-tags/uaccess/%s', ignoring: %m", de->d_name);
        continue;
}
```

i.e. totally ignore ENOENT, and log but ignore all other errors.

yuwata:
@poettering Thank you for the review. The suggestion is applied. Upgrading the green label.

PR #22984, opened 2022-04-05
core: taint if /usr is unmerged

bluca:
9afd5e7b975e8051c011ff9c07c95e80bd954469 introduced a build-time
taint, introduce a runtime one as well, in preparation for
removing support for unmerged-usr in a future release

### DIFF
--- a/README
+++ b/README
@@ -353,6 +353,13 @@ WARNINGS and TAINT FLAGS:
         For more information on this issue consult
         https://www.freedesktop.org/wiki/Software/systemd/separate-usr-is-broken
 
+        systemd will warn if the filesystem is not usr-merged (i.e.: /bin, /sbin
+        and /lib* are not symlinks to their counterparts under /usr). Taint flag
+        'unmerged-usr' will be set when this condition is detected.
+
+        For more information on this issue consult
+        https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge
+
         systemd requires that the /run mount point exists. systemd also
         requires that /var/run is a symlink to /run. Taint flag 'var-run-bad'
         will be set when this condition is detected.
--- a/src/core/manager.c
+++ b/src/core/manager.c
@@ -4377,11 +4377,15 @@ char* manager_taint_string(const Manager *m) {
 
         assert(m);
 
-        const char* stage[11] = {};
+        const char* stage[12] = {};
         size_t n = 0;
 
         if (m->taint_usr)
                 stage[n++] = "split-usr";
+
+        _cleanup_free_ char *usrbin = NULL;
+        if (readlink_malloc("/bin", &usrbin) < 0 || !PATH_IN_SET(usrbin, "usr/bin", "/usr/bin"))
+                stage[n++] = "unmerged-usr";
 
         if (access("/proc/cgroups", F_OK) < 0)
                 stage[n++] = "cgroups-missing";


keszybz:
Now that #22968 got merged, please update this to describe the new taint flag in the README.

src/core/manager.c
```diff
@@ -949,10 +949,14 @@ int manager_new(LookupScope scope, ManagerTestRunFlags test_run_flags, Manager *
                         return r;
         }
 
-        m->taint_usr =
+        m->taint_split_usr =
                 !in_initrd() &&
                 dir_is_empty("/usr") > 0;
 
+        m->taint_unmerged_usr =
```
poettering: no need to cache this. we can check this cheaply any time...

The reason why the split usr thing is cached was that it reports whether /usr is initially empty when pid 1 initializes, and thus it must be done early since if we do it later we can't really do the test anymore since something most likely will be mounted over /usr/.

Hence, just do the check from manager_taint_string()

src/core/manager.c
```diff
@@ -949,10 +949,14 @@ int manager_new(LookupScope scope, ManagerTestRunFlags test_run_flags, Manager *
...
+                !in_initrd() &&
+                is_symlink("/bin") == 0;
```
poettering: i'd probably generate a LOG_DEBUG message in case this returns < 0.

poettering:
as mentioned elsewhere, I'd love one more taint flag if sbin isn't a symlink to bin... ;-)

poettering:
i wonder btw if we shouldn't make the check stricter, i.e. check that the symlink contents is exactly "usr/bin".

bluca:
> as mentioned elsewhere, I'd love one more taint flag if sbin isn't a symlink to bin... ;-)

One thing at a time :-)

src/core/manager.c
bluca: refactored to do the same as the /var/run check with readlink_malloc

README
```diff
@@ -353,6 +353,13 @@ WARNINGS and TAINT FLAGS:
         For more information on this issue consult
         https://www.freedesktop.org/wiki/Software/systemd/separate-usr-is-broken
 
+        systemd will warn if the filesystem is not usr-merged (ie: /bin, /sbin
+        and /lib* are symlinks to their counterparts under /usr). Taint flag
+        'unmerged-usr' will be set when this condition is detected.
```
keszybz: ie → i.e.

The sense of the phrase in parenthesis is reversed, a "not" is missing there.

src/core/manager.c
```diff
@@ -4383,6 +4383,7 @@ char *manager_taint_string(Manager *m) {
         assert(m);
 
         buf = new(char, sizeof("split-usr:"
+                               "unmerged-usr:"
```
keszybz: Let's merge #22993 first and rebase this on top. (Since it needs updating anyway.)

yuwata:
#22993 is merged. Please rebase.

README
yuwata: This is not addressed?

bluca: ops, missed the second one

bluca: fixed

yuwata (approved):
LGTM.

poettering:
lgtm, too

bluca:
```
19:01:51 FAILED TASKS:
19:01:51 -------------
19:01:51 TEST-64-UDEV-STORAGE
```

unrelated

PR #22985, opened 2022-04-05
Semaphore CI fixes

bluca:


bluca:
Other than the setup issue, whatever changed in semaphoreci also broke the autopkgtest-reboot functionality. I've simply disabled the two tests that use that for now, so that we can get the rest going again, filed https://github.com/systemd/systemd/issues/22991 to track it

.semaphore/semaphore-runner.sh
```diff
@@ -36,8 +36,11 @@ create_container() {
     # enable source repositories so that apt-get build-dep works
     sudo lxc-attach -n "$CONTAINER" -- sh -ex <<EOF
 sed 's/^deb/deb-src/' /etc/apt/sources.list >> /etc/apt/sources.list.d/sources.list
-# wait until online
-while [ -z "\$(ip route list 0/0)" ]; do sleep 1; done
+# We might attach the console too soon
+while ! systemctl --quiet --wait is-system-running; do sleep 1; done
```
mrc0mmand, line 40: I wonder - why is this run in a loop when `systemctl --wait` should _wait_ for the boot process to complete before returning?

mrc0mmand:
LGTM, thanks for working on this!

.semaphore/semaphore-runner.sh
bluca: That works only if systemctl can connect to the socket, otherwise it doesn't wait and it exits with failure immediately:

```
+ sed s/^deb/deb-src/ /etc/apt/sources.list
+ systemctl --quiet --wait is-system-running
Failed to connect to bus: No such file or directory
+ sleep 1
+ systemctl --quiet --wait is-system-running
+ + debconf-set-selections
```

mrc0mmand: I see, thanks.

PR #22989, opened 2022-04-06
analyze: Fix verify exit status regression

daandemeyer:
Previously, systemd-analyze verify would return 0 even if warnings
were raised during analysis of the specified units or their
dependencies. With 3cc3dc7736727973df852c4bdf7d312c2e752e5e, verify
was changed to return 1 when warnings were raised.

This commit introduces a new mode "never" which becomes the new
default and makes 'verify' behave like before, all specified units
and their dependencies are analyzed but zero is returned even if
warnings happen during analysis of any of the units.


man/systemd-analyze.xml
```diff
@@ -818,11 +818,13 @@ $ systemd-analyze verify /tmp/source:alias.service
 
         <listitem><para>Control verification of units and their dependencies and whether
         <command>systemd-analyze verify</command> exits with a non-zero process exit status or not. With
+        <command>compat</command>, return zero as the exit status regardless of whether warnings arise during
```
bluca: Bikeshedding: call it 'never' instead? Ie, describe what it does, not why it's there

src/analyze/analyze-verify-util.c
```diff
@@ -339,9 +339,10 @@ int verify_units(char **filenames, LookupScope scope, bool check_man, bool run_g
 }
 
 static const char* const recursive_errors_table[_RECURSIVE_ERRORS_MAX] = {
-        [RECURSIVE_ERRORS_NO]  = "no",
-        [RECURSIVE_ERRORS_YES] = "yes",
-        [RECURSIVE_ERRORS_ONE] = "one",
+        [RECURSIVE_ERRORS_NO]     = "no",
+        [RECURSIVE_ERRORS_YES]    = "yes",
+        [RECURSIVE_ERRORS_ONE]    = "one",
+        [RECURSIVE_ERRORS_COMPAT] = "compat",
```
bluca: missed one

daandemeyer: Yikes, that's what I get for not compiling

bluca (approved):

yuwata:
Is this really good for backporting to the stable branch(es)? I mean, this introduced a new command line option. Usually, this is not good to backport. Are there any other way to fix the regression??

bluca:
I guess it could be backported without the table change, so that the 'never' value cannot be specified, but it's still the default?

DaanDeMeyer:
I guess we could reuse _RECURSIVE_ERRORS_INVALID as the default value instead of introducing a new value. That would fix the issue without introducing a new table value.

DaanDeMeyer:
New version just changes the default value without introducing a new mode

yuwata (approved):
LGTM. And I think this is good to be backported to stable repositories.

Issue #17469, opened 2020-10-27
TEST-50-DISSECT: Failed to set up mount namespacing: No such file or directory


mrc0mmand:
**systemd version the issue has been seen with**
 > latest master

**Used distribution**
 > CentOS 7 (haven't seen it on Arch so far)

**Linux kernel version used** (`uname -a`)
 > 3.10.0-1160.2.2.el7.x86_64

**CPU architecture issue was seen on**
 > x86_64

The `TEST-50-DISSECT` occasionally fails with `Failed to set up mount namespacing: No such file or directory` on CentOS 7 (seen several times in the last few days):

```
[   11.405446] systemd[1]: Started run-u17.service.
...
[   11.406667] systemd[1]: Failed to read pids.max attribute of cgroup root, ignoring: No data available
[   11.406804] systemd[1]: Found unit run-u17.service at /run/systemd/transient/run-u17.service (regular file)
[   11.406943] systemd[889]: Waiting for device 'loop0' to initialize for 44.999997s.
[   11.408158] systemd[1]: Preset files don't specify rule for run-u17.service. Enabling.
[   11.408320] systemd[889]: Successfully waited for device 'loop0' to initialize for 271us.
[   11.408485] systemd[1]: Sent message type=method_return sender=n/a destination=:1.17 path=n/a interface=n/a member=n/a cookie=5031 reply_cookie=4 signature=a{sv} error-name=n/a error-message=n/a
[   11.408767] systemd[889]: sd-device-enumerator: Failed to open parent directory /sys/devices/virtual/block/loop0/loop0p1/trace: No such file or directory
...
[   11.408926] systemd[1]: Failed to read pids.max attribute of cgroup root, ignoring: No data available
[   11.409094] systemd[889]: Failed to dissect image: No such file or directory
...
[   11.421316] systemd[889]: run-u17.service: Failed to set up mount namespacing: No such file or directory
[   11.421450] systemd[889]: run-u17.service: Failed at step NAMESPACE spawning cat: No such file or directory
[   11.422549] systemd[1]: Received SIGCHLD from PID 889 ((cat)).
[   11.422753] systemd[1]: Child 889 ((cat)) died (code=exited, status=226/NAMESPACE)
[   11.422931] systemd[1]: run-u17.service: Child 889 belongs to run-u17.service.
[   11.423145] systemd[1]: run-u17.service: Main process exited, code=exited, status=226/NAMESPACE
[...]

PR #22992, opened 2022-04-06
loop-util/image dissect fixes

poettering:
Another attempt at tackling #17469 even on older kernels, with very different means.

### DIFF
--- a/src/partition/repart.c
+++ b/src/partition/repart.c
@@ -12,6 +12,7 @@
 #include <sys/ioctl.h>
 #include <sys/stat.h>
 
+#include "sd-device.h"
 #include "sd-id128.h"
 
 #include "alloc-util.h"
@@ -3643,14 +3644,13 @@ static int resolve_copy_blocks_auto_candidate(
                 sd_id128_t *ret_uuid) {
 
         _cleanup_(blkid_free_probep) blkid_probe b = NULL;
-        _cleanup_free_ char *p = NULL;
+        _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
         _cleanup_close_ int fd = -1;
-        const char *pttype, *t;
+        const char *pttype, *t, *p;
         sd_id128_t pt_parsed, u;
         blkid_partition pp;
         dev_t whole_devno;
         blkid_partlist pl;
-        struct stat st;
         int r;
 
         /* Checks if the specified partition has the specified GPT type UUID, and is located on the specified
@@ -3673,21 +3673,19 @@ static int resolve_copy_blocks_auto_candidate(
                                 major(partition_devno), minor(partition_devno),
                                 major(restrict_devno), minor(restrict_devno));
 
-        r = device_path_make_major_minor(S_IFBLK, whole_devno, &p);
+        r = sd_device_new_from_devnum(&dev, 'b', whole_devno);
         if (r < 0)
-                return log_error_errno(r, "Failed to convert block device to device node path: %m");
+                return log_error_errno(r, "Failed to create sd-device for block device %u:%u: %m",
+                                       major(whole_devno), minor(whole_devno));
 
-        fd = open(p, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
+        r = sd_device_get_devname(dev, &p);
+        if (r < 0)
+                return log_error_errno(r, "Failed to get name of block device %u:%u: %m",
+                                       major(whole_devno), minor(whole_devno));
+
+        fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
         if (fd < 0)
-                return log_error_errno(r, "Failed to open '%s': %m", p);
-
-        if (fstat(fd, &st) < 0)
-                return log_error_errno(r, "Failed to stat '%s': %m", p);
-
-        if (!S_ISBLK(st.st_mode) || st.st_rdev != whole_devno)
-                return log_error_errno(
-                                SYNTHETIC_ERRNO(EPERM),
-                                "Opened and determined block device don't match, refusing.");
+                return log_error_errno(fd, "Failed to open block device %s: %m", p);
 
         b = blkid_new_probe();
         if (!b)
@@ -3805,7 +3803,7 @@ static int resolve_copy_blocks_auto(
                 sd_id128_t type_uuid,
                 const char *root,
                 dev_t restrict_devno,
-                char **ret_path,
+                dev_t *ret_devno,
                 sd_id128_t *ret_uuid) {
 
         const char *try1 = NULL, *try2 = NULL;
@@ -3814,8 +3812,6 @@ static int resolve_copy_blocks_auto(
         sd_id128_t found_uuid = SD_ID128_NULL;
         dev_t devno, found = 0;
         int r;
-
-        assert(ret_path);
 
         /* Enforce some security restrictions: CopyBlocks=auto should not be an avenue to get outside of the
          * --root=/--image= confinement. Specifically, refuse CopyBlocks= in combination with --root= at all,
@@ -3926,9 +3922,8 @@ static int resolve_copy_blocks_auto(
                 return log_error_errno(SYNTHETIC_ERRNO(ENXIO),
                                        "Unable to automatically discover suitable partition to copy blocks from.");
 
-        r = device_path_make_major_minor(S_IFBLK, found, ret_path);
-        if (r < 0)
-                return log_error_errno(r, "Failed to convert dev_t to device node path: %m");
+        if (ret_devno)
+                *ret_devno = found;
 
         if (ret_uuid)
                 *ret_uuid = found_uuid;
@@ -3972,32 +3967,43 @@ static int context_open_copy_block_paths(
                                                        "Copying from block device node is not permitted in --image=/--root= mode, refusing.");
 
                 } else if (p->copy_blocks_auto) {
+                        _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
+                        const char *devname;
+                        dev_t devno;
 
-                        r = resolve_copy_blocks_auto(p->type_uuid, root, restrict_devno, &opened, &uuid);
+                        r = resolve_copy_blocks_auto(p->type_uuid, root, restrict_devno, &devno, &uuid);
                         if (r < 0)
                                 return r;
 
-                        source_fd = open(opened, O_RDONLY|O_CLOEXEC|O_NOCTTY);
+                        r = sd_device_new_from_devnum(&dev, 'b', devno);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to create sd-device object for device %u:%u: %m", major(devno), minor(devno));
+
+                        r = sd_device_get_devname(dev, &devname);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to get device name of %u:%u: %m", major(devno), minor(devno));
+
+                        opened = strdup(devname);
+                        if (!opened)
+                                return log_oom();
+
+                        source_fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
                         if (source_fd < 0)
-                                return log_error_errno(errno, "Failed to open automatically determined source block copy device '%s': %m", opened);
+                                return log_error_errno(source_fd, "Failed to open automatically determined source block copy device '%s': %m", opened);
 
                         if (fstat(source_fd, &st) < 0)
                                 return log_error_errno(errno, "Failed to stat block copy file '%s': %m", opened);
-
-                        /* If we found it automatically, it must be a block device, let's enforce that */
-                        if (!S_ISBLK(st.st_mode))
-                                return log_error_errno(SYNTHETIC_ERRNO(EBADF),
-                                                       "Automatically detected source block copy device '%s' is not a block device, refusing: %m", opened);
-                }  else
+                } else
                         continue;
 
                 if (S_ISDIR(st.st_mode)) {
-                        _cleanup_free_ char *bdev = NULL;
+                        _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
+                        const char *bdev;
 
                         /* If the file is a directory, automatically find the backing block device */
 
                         if (major(st.st_dev) != 0)
-                                r = device_path_make_major_minor(S_IFBLK, st.st_dev, &bdev);
+                                r = sd_device_new_from_devnum(&dev, 'b', st.st_dev);
                         else {
                                 dev_t devt;
 
@@ -4009,22 +4015,23 @@ static int context_open_copy_block_paths(
                                 if (r < 0)
                                         return log_error_errno(r, "Unable to determine backing block device of '%s': %m", opened);
 
-                                r = device_path_make_major_minor(S_IFBLK, devt, &bdev);
+                                r = sd_device_new_from_devnum(&dev, 'b', devt);
                         }
                         if (r < 0)
-                                return log_error_errno(r, "Failed to determine block device path for block device backing '%s': %m", opened);
+                                return log_error_errno(r, "Failed to create sd-device object for block device backing '%s': %m", opened);
+
+                        r = sd_device_get_devpath(dev, &bdev);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to get device name for block device backing '%s': %m", opened);
 
                         safe_close(source_fd);
 
-                        source_fd = open(bdev, O_RDONLY|O_CLOEXEC|O_NOCTTY);
+                        source_fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NONBLOCK);
                         if (source_fd < 0)
-                                return log_error_errno(errno, "Failed to open block device '%s': %m", bdev);
+                                return log_error_errno(source_fd, "Failed to open block device '%s': %m", bdev);
 
                         if (fstat(source_fd, &st) < 0)
                                 return log_error_errno(errno, "Failed to stat block device '%s': %m", bdev);
-
-                        if (!S_ISBLK(st.st_mode))
-                                return log_error_errno(SYNTHETIC_ERRNO(ENOTBLK), "Block device '%s' is not actually a block device, refusing.", bdev);
                 }
 
                 if (S_ISREG(st.st_mode))


poettering:
Oh man, I am so curious if this one will pass CI fine...

src/test/test-loop-block.c
```diff
@@ -228,19 +228,27 @@ int main(int argc, char *argv[]) {
...
+                for (unsigned i = 0; i < N_THREADS; i++)
+                        assert_se(pthread_create(threads + i, NULL, thread_func, FD_TO_PTR(fd)) == 0);
+        }
 
         log_notice("All threads started now.");
 
-        for (unsigned i = 0; i < N_THREADS; i++) {
-                log_notice("Joining thread #%u.", i);
+        if (N_THREADS == 1)
+                assert_se(thread_func(FD_TO_PTR(fd)) == NULL);
```
bluca: maybe return 0 immediately here? otherwise the "Threads are all terminated now" message below is a bit confusing

src/test/test-loop-block.c
```diff
@@ -10,27 +10,30 @@
...
 #include "string-util.h"
 #include "strv.h"
 #include "tests.h"
 #include "tmpfile-util.h"
 #include "user-util.h"
 #include "virt.h"
 
-#define N_THREADS 5
-#define N_ITERATIONS 3
+static unsigned arg_n_threads = 1;
```
bluca: Any reason to change the default from 5 to 1?

src/shared/dissect-image.c
```diff
@@ -529,91 +147,74 @@ static void check_partition_flags(
...
-        int r;
-
-        start = now(CLOCK_MONOTONIC);
-        left = usec_sub_unsigned(deadline, start);
+static void dissected_partition_done(DissectedPartition *p) {
+        assert(p);
 
-        if (DEBUG_LOGGING) {
-                const char *sn = NULL;
+        free(p->fstype);
```
bluca, line 156: `mfree`?

src/shared/dissect-image.c
```diff
@@ -529,91 +147,74 @@ static void check_partition_flags(
...
+                int fd,
+                const char *name,
+                int nr,
+                uint64_t start,
+                uint64_t size) {
 
-        for (;;) {
-                usec_t local_deadline, n;
-                bool last_try;
+        assert(fd >= 0);
```
bluca, line 177: `assert(name)` (strlen will barf on nul)

src/shared/dissect-image.c
```diff
@@ -529,91 +147,74 @@ static void check_partition_flags(
...
-}
-#endif
+        assert(whole_devname);
+        assert(nr > 0);
 
-#define DEVICE_TIMEOUT_USEC (45 * USEC_PER_SEC)
+        /* Given a whole block device node name (e.g. /dev/sda or /dev/loop7) generate a partition device
+         * name (e.g. /dev/sda7 or /dev/loop7p5). The rule the kernel uses is simple: if whole block device
+         * node name ends in a digit, then suffix a 'p', followed by the partition number. Otherwise, just
+         * suffix the partition number without any 'p'. */
```
bluca, line 214: How 'stable' is this kernel rule? What happens if it changes in the future, say for some fancy new device type?

src/shared/dissect-image.c
```diff
@@ -915,6 +487,31 @@ int dissect_image(
...
 
+                r = make_partition_devname(devname, nr, &node);
+                if (r < 0)
+                        return r;
+
+                /* So here's the thing: after the main ("whole") block device popped up it might take a while
+                 * before the kernel fully probed the partition table. Waiting for that to finish is icky in
+                 * userspace. So here's what we do instead. We issue the BLKPG_ADD_PARTITION ioctl to add the
+                 * partition ourselves, racing against the kernel. Good thing is: if this call fails with
+                 * EBUSY then the kernel was quicker than us , and that's totally OK, the outcome is good for
```
bluca: /us ,/us,/

src/shared/dissect-image.c
```diff
@@ -915,6 +487,31 @@ int dissect_image(
...
+                 * us: the device node will exist. If OTOH our call was successful we won the race. Which is
+                 * also good as the outcome is the same: the partition block device exists, and we can use
+                 * it.
+                 *
+                 * Kernel returns EBUSY if there's already a partition by that number or an overlapping
+                 * partition already existent. */
```
bluca, line 509: Do we know how 'reliable' this mechanism is? Would it be worth it to keep the old code as fallback in case we get a failure that is not EBUSY, and what check can we do to ensure even if this works, that everything looks the way it's supposed to?

bluca:
```
../src/shared/dissect-image.c:168:12: error: unused function 'ioctl_partition_add' [-Werror,-Wunused-function]
static int ioctl_partition_add(
           ^
../src/shared/dissect-image.c:197:12: error: unused function 'make_partition_devname' [-Werror,-Wunused-function]
static int make_partition_devname(
```

bluca:
Looks nice (assuming the kernel side is reliable), I can restart the focal tests a few times, just to make sure, when they succeed

src/shared/dissect-image.c
```diff
@@ -529,91 +147,74 @@ static void check_partition_flags(
...
+        if (isempty(whole_devname)) /* Make sure there *is* a last char */
+                return -EINVAL;
 
-        free(p->fstype);
-        free(p->node);
-        free(p->label);
-        free(p->decrypted_fstype);
-        free(p->decrypted_node);
-        free(p->mount_options);
+        need_p = strchr(DIGITS, whole_devname[strlen(whole_devname)-1]); /* Last char a digit? */
```
dnicolodi, line 219: `need_p = isdigit(whole_devname[strlen(whole_devname) - 1])`?

poettering: isdigit() sucks. It's locale dependent (i.e. there's even isdigit_l). We really want only ASCII 0…9 here, hence we avoid the function.

(I mean let's hope that all locales define that the same way, but already the possibility that they could depart from ASCII 0…9 and add some other number chars to it, sounds like something we really should avoid).

`DIGITS` is defined by us, we know exactly what it is...

bluca:
on focal-ppc64el:

```
TEST-50-DISSECT:                    FAIL     (611 s)
```

src/test/test-loop-block.c
poettering: the way i see it, the in this special case the worker thread is the main process, and that's a thread too...

src/test/test-loop-block.c
poettering: nope, that was a forgotten left-over change. will revert to old params

src/shared/dissect-image.c
poettering: note we explicitly reset the whole struct further down anyway.

(also note that this is unmodified, pre-existing code actually, github's diff logic is just terrible at recognizing that, i guess the change is too massive to recognize that these lines existed before)

src/shared/dissect-image.c
poettering: Very. It's encoded in libfdisk and lots of other places. If they change that *a lot* would break.

I think it's very clearly de-facto API these days

src/shared/dissect-image.c
poettering: well, the CI is supposed to find out how stable this is. I think this new code is sooo much simpler though, and a lot more compatible with older kernels, that I'd rather not attempt to maintain fallback code that we expressly want to replace because problematic in various environments.

src/shared/dissect-image.c
bluca: ah yes, missed that

src/shared/dissect-image.c
dnicolodi: The man page for the libc isxxx() functions says that these functions are locale dependent but it also says "isdigit() checks for a digit (0 through 9)". The "0 through 9" part made me believe that this is less locale dependent than what it probably is...

src/shared/loop-util.c
```diff
@@ -104,26 +111,40 @@ static int device_has_block_children(sd_device *d) {
...
 
-                r = sd_device_get_sysname(q, &sn);
-                if (r < 0)
+                r = sd_device_get_devtype(q, &dt);
+                if (r < 0) {
+                        log_debug_errno(r, "Failed to get devtype of child '%s', ignoring: %m", sn);
                         continue;
+                }
 
-                if (streq(sn, main_sn))
```
yuwata: So, this spurious check is dropped. So, sn and main_sn are not used any more, and can be dropped.

Note, sn is used for logging, but you can use log_device_debug_errno().

yuwata (approved):
LGTM. One minor comment.

bluca:
```
>>> PATH=/home/runner/work/systemd/systemd/build:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin SYSTEMD_LANGUAGE_FALLBACK_MAP=/home/runner/work/systemd/systemd/src/locale/language-fallback-map MALLOC_PERTURB_=55 /home/runner/work/systemd/systemd/build/test-loop-block
stdout:
Checking that no-one is using this disk right now ... OK

Disk /var/tmp/.#sfdisk97c4f3c7664e6cff: 256 MiB, 268435456 bytes, 524288 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Script header accepted.
>>> Created a new GPT disklabel (GUID: 7CA4239B-AF79-894B-9CF7-4260D6CF99D4).
/var/tmp/.#sfdisk97c4f3c7664e6cff1: Created a new partition 1 of type 'EFI System' and of size 32 MiB.
/var/tmp/.#sfdisk97c4f3c7664e6cff2: Created a new partition 2 of type 'Linux extended boot' and of size 32 MiB.
/var/tmp/.#sfdisk97c4f3c7664e6cff3: Created a new partition 3 of type 'Linux swap' and of size 32 MiB.
/var/tmp/.#sfdisk97c4f3c7664e6cff4: Created a new partition 4 of type 'Linux root (x86-64)' and of size 32 MiB.
/var/tmp/.#sfdisk97c4f3c7664e6cff5: Created a new partition 5 of type 'Linux home' and of size 32 MiB.
/var/tmp/.#sfdisk97c4f3c7664e6cff6: Done.

New situation:
Disklabel type: gpt
Disk identifier: 7CA4239B-AF79-894B-9CF7-4260D6CF99D4

Device                              Start    End Sectors Size Type
/var/tmp/.#sfdisk97c4f3c7664e6cff1   2048  67583   65536  32M EFI System
/var/tmp/.#sfdisk97c4f3c7664e6cff2  67584 133119   65536  32M Linux extended boot
/var/tmp/.#sfdisk97c4f3c7664e6cff3 133120 198655   65536  32M Linux swap
/var/tmp/.#sfdisk97c4f3c7664e6cff4 198656 264191   65536  32M Linux root (x86-64)
/var/tmp/.#sfdisk97c4f3c7664e6cff5 264192 329727   65536  32M Linux home

The partition table has been altered.
Syncing disks.
stderr:
Wed 2022-04-06 18:25:34 UTC (28671) Found cgroup2 on /sys/fs/cgroup/unified, unified hierarchy for systemd controller
Wed 2022-04-06 18:25:34 UTC (28671) Found container virtualization none.
Wed 2022-04-06 18:25:34 UTC (28671) Skipping non-partition child 'loop3' (devtype=disk).
Wed 2022-04-06 18:25:34 UTC (28671) Successfully acquired /dev/loop3, devno=7:3, nr=3, diskseq=0
Wed 2022-04-06 18:25:34 UTC (28671) Assertion 'dissect_image(loop->fd, NULL, NULL, loop->diskseq, loop->uevent_seqnum_not_before, loop->timestamp_not_before, 0, &dissected) >= 0' failed at src/test/test-loop-block.c:214, function int run(int, char **)(). Aborting.
```

poettering:
> Wed 2022-04-06 18:25:34 UTC (28671) Successfully acquired /dev/loop3, devno=7:3, nr=3, diskseq=0
> Wed 2022-04-06 18:25:34 UTC (28671) Assertion 'dissect_image(loop->fd, NULL, NULL, loop->diskseq, loop->uevent_seqnum_not_before, loop->timestamp_not_before, 0, &dissected) >= 0' failed at src/test/test-loop-block.c:214, function int run(int, char **)(). Aborting.

ah, eh, that CI is compiled without libblkid. I guess this has to fail.

bluca:
```
../src/test/test-loop-block.c:149:19: error: unused variable 'threads' [-Werror,-Wunused-variable]
        pthread_t threads[arg_n_threads];
```

bluca:
```
ESC[31mFAIL: test-loop-blockESC[0m
Wed 2022-04-06 20:23:30 UTC (19440) Found cgroup2 on /sys/fs/cgroup/unified, unified hierarchy for systemd controller
Wed 2022-04-06 20:23:30 UTC (19440) Found container virtualization none.
Checking that no-one is using this disk right now ... OK

Disk /var/tmp/.#sfdiskfe68c7488d20e43e: 256 MiB, 268435456 bytes, 524288 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Script header accepted.
>>> Created a new GPT disklabel (GUID: 94307F57-3439-4244-9A63-E987C1A44275).
/var/tmp/.#sfdiskfe68c7488d20e43e1: Created a new partition 1 of type 'EFI System' and of size 32 MiB.
/var/tmp/.#sfdiskfe68c7488d20e43e2: Created a new partition 2 of type 'Linux extended boot' and of size 32 MiB.
/var/tmp/.#sfdiskfe68c7488d20e43e3: Created a new partition 3 of type 'Linux swap' and of size 32 MiB.
/var/tmp/.#sfdiskfe68c7488d20e43e4: Created a new partition 4 of type 'Linux root (x86-64)' and of size 32 MiB.
/var/tmp/.#sfdiskfe68c7488d20e43e5: Created a new partition 5 of type 'Linux home' and of size 32 MiB.
/var/tmp/.#sfdiskfe68c7488d20e43e6: Done.

New situation:
Disklabel type: gpt
Disk identifier: 94307F57-3439-4244-9A63-E987C1A44275

Device                              Start    End Sectors Size Type
/var/tmp/.#sfdiskfe68c7488d20e43e1   2048  67583   65536  32M EFI System
/var/tmp/.#sfdiskfe68c7488d20e43e2  67584 133119   65536  32M Linux extended boot
/var/tmp/.#sfdiskfe68c7488d20e43e3 133120 198655   65536  32M Linux swap
/var/tmp/.#sfdiskfe68c7488d20e43e4 198656 264191   65536  32M Linux root (x86-64)
/var/tmp/.#sfdiskfe68c7488d20e43e5 264192 329727   65536  32M Linux home

The partition table has been altered.
Syncing disks.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/power/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/holders/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu0/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/queue/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/queue/iosched/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/integrity/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/trace/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) sd-device: the uevent file "/sys/devices/virtual/block/loop0/slaves/uevent" does not exist.
Wed 2022-04-06 20:23:30 UTC (19440) loop0: Skipping non-partition child (devtype=disk).
Wed 2022-04-06 20:23:30 UTC (19440) Successfully acquired /dev/loop0, devno=7:0, nr=0, diskseq=0
Wed 2022-04-06 20:23:30 UTC (19440) Image name .#sfdiskfe68c7488d20e43e is not valid, ignoring
Wed 2022-04-06 20:23:30 UTC (19440) Kernel was quicker than us in adding partition 1.
Wed 2022-04-06 20:23:30 UTC (19440) Kernel was quicker than us in adding partition 2.
Wed 2022-04-06 20:23:30 UTC (19440) Kernel was quicker than us in adding partition 3.
Wed 2022-04-06 20:23:30 UTC (19440) Kernel was quicker than us in adding partition 4.
Wed 2022-04-06 20:23:30 UTC (19440) Kernel was quicker than us in adding partition 5.
Wed 2022-04-06 20:23:30 UTC (19440) No type detected on partition /dev/loop0p4
Wed 2022-04-06 20:23:30 UTC (19440) No type detected on partition /dev/loop0p5
Wed 2022-04-06 20:23:30 UTC (19440) No type detected on partition /dev/loop0p2
Wed 2022-04-06 20:23:30 UTC (19440) No type detected on partition /dev/loop0p3
Wed 2022-04-06 20:23:30 UTC (19440) Successfully forked off '(mkfs)' as PID 19444.
mkfs.vfat: unable to open /dev/loop0p1: No such file or directory
mkfs.fat 4.1 (2017-01-24)
Wed 2022-04-06 20:23:30 UTC (19440) (mkfs) failed with exit status 1.
Wed 2022-04-06 20:23:30 UTC (19440) Assertion 'make_filesystem(dissected->partitions[PARTITION_ESP].node, "vfat", "EFI", id, true) >= 0' failed at src/test/test-loop-block.c:229, function run(). Aborting.
```

also TEST-58-REPART seems sad everywhere, not sure if related

poettering:
For now I split out the initial commits into #23008. They should be safe to merge either way.

The mkfs/ENOENT issue sucks. My educated guess is that it is caused by the fact that the partition is first added to the in-kernel partition table array in the `add_partition()` kernel function via RCU, and only after that created in sysfs/devtmpfs/netlink. This means if another process races against that then it will see EBUSY *before* the devnode of the earlier code is there. And that causes the problem.

bluca:
`TEST-58-REPART` failed in 3 tests, sounds related?

yuwata:
Yay! All CentOS CIs are green!

src/partition/repart.c
```diff
@@ -3972,32 +3970,33 @@ static int context_open_copy_block_paths(
...
                         r = resolve_copy_blocks_auto(p->type_uuid, root, restrict_devno, &opened, &uuid);
                         if (r < 0)
                                 return r;
 
-                        source_fd = open(opened, O_RDONLY|O_CLOEXEC|O_NOCTTY);
+                        r = sd_device_new_from_path(&dev, opened);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to create sd-device object for device '%s': %m", opened);
+
+                        source_fd = sd_device_open(dev, O_RDONLY|O_CLOEXEC|O_NOCTTY);
```
yuwata: By the way, is the O_NOCTTY flag correct? Why? Maybe O_NONBLOCK?

bluca:
I've restarted the two focal tests, will do so a couple of times to put it under pressure - should probably do it on Monday as well when the infra is under heavier load

poettering:
Oh nice, @yuwata indeed, with my rework the locking would mean udev couldn#t create the major/minor symlinks anymore. Thanks for the fix, looks perfect.

src/partition/repart.c
poettering: So, O_NOCTTY is probably something we always should specify if there's the tiniest of chances that what we are opening there ends up being a tty. I mean, if Linux had a nicer API it would be inverted, and be called O_CTTY because the behaviour it enables should be opt-out, not opt-in. (In fact some BSDs do just that.)

O_NONBLOCK only carries value on some specific devices, and fifos. It's a much unclearer situation.

I guess the rule should be:

1. Specify O_NOCTTY in all open() calls, except when O_DIRECTORY (because if something is a dir, it definitely isn't a tty), O_PATH (because an inode fd definitely is not a tty), O_CREAT|O_EXCL (a new file cannot be a tty) or O_TMPFILE (same) is already specified.

2. Specify O_NONBLOCK whenever we expect to open a regular file only. Because not specifying O_NONBLOCK means we might otherwise hang in the open() call.

(I guess one of those days we should add a wrapper open_regular() or so that safely opens a regular file and safely refuses weird stuff like fifos or special devices, and will not block on them in any cases)

yuwata: Thanks.

mrc0mmand:
This is awesome change, thanks! I can finally drop quite a lot of kludges https://github.com/systemd/systemd-centos-ci/pull/484.

freedge:
I am wondering if this is what is causing https://bugzilla.redhat.com/show_bug.cgi?id=2083900

systemd-gpt-auto-generator was passively waiting for partitions, now it actively calls BLKPG ioctl which needs some extra capabilities, triggering SELinux AVC at the moment.

PR #22993, opened 2022-04-06
Refactor manager_taint_string()

keszybz:


src/test/test-manager.c
```diff
@@ -0,0 +1,20 @@
+#include "manager.h"
```
yuwata, line 3: No SPDX comment?

yuwata (approved):
LGTM. One minor comment.

src/test/test-manager.c
```diff
@@ -0,0 +1,20 @@
...
+#include "tests.h"
+
+TEST(manager_taint_string) {
+        Manager m = {};
+
+        _cleanup_free_ char *a = manager_taint_string(&m);
+        assert_se(a);
+        log_debug("taint string w/o split-usr: '%s'", a);
+        /* split-usr is the only one that is cached in Manager, so we know it's not present */
```
bluca: Shouldn't this check what the value is? It's not enforced besides the taint, no, so it could be already true here?

src/test/test-manager.c
keszybz: oops.

src/test/test-manager.c
keszybz: I extended the comment to explain this more, and added one more test.

bluca (approved):

yuwata (approved):

PR #22998, opened 2022-04-06
journal: Add compact mode

daandemeyer:
This PR contains the first set of commits from #21183 that are (hopefully) less controversial and can be merged on their own already. See #21183 for more context and information.

### DIFF
--- a/man/journalctl.xml
+++ b/man/journalctl.xml
@@ -854,6 +854,17 @@
         cryptographic theory it is based on.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--convert=</option></term>
+
+        <listitem><para>Converts the specified journal files to the latest supported journal format. Takes
+        the path to store the converted journal files. The path should include the filename to be used for
+        the converted files, with the <literal>.journal</literal> extension (e.g.
+        <filename>/a/b/c/converted.journal</filename> will store the journal files in the
+        <filename>/a/b/c</filename> directory using <filename>converted.journal</filename> as the filename).
+        </para></listitem>
+      </varlistentry>
+
       <xi:include href="standard-options.xml" xpointer="help" />
       <xi:include href="standard-options.xml" xpointer="version" />
     </variablelist>
--- a/meson.build
+++ b/meson.build
@@ -2293,11 +2293,13 @@ public_programs += executable(
         install : true)
 
 if get_option('link-journalctl-shared')
-        journalctl_link_with = [libshared]
+        journalctl_link_with = [libshared,
+                                libjournal_core]
 else
         journalctl_link_with = [libsystemd_static,
                                 libshared_static,
-                                libbasic_gcrypt]
+                                libbasic_gcrypt,
+                                libjournal_core]
 endif
 
 public_programs += executable(
--- a/src/journal/journalctl.c
+++ b/src/journal/journalctl.c
@@ -44,6 +44,7 @@
 #include "locale-util.h"
 #include "log.h"
 #include "logs-show.h"
+#include "managed-journal-file.h"
 #include "memory-util.h"
 #include "mkdir.h"
 #include "mount-util.h"
@@ -128,6 +129,7 @@ static uint64_t arg_vacuum_size = 0;
 static uint64_t arg_vacuum_n_files = 0;
 static usec_t arg_vacuum_time = 0;
 static char **arg_output_fields = NULL;
+static const char *arg_convert = NULL;
 static const char *arg_pattern = NULL;
 static pcre2_code *arg_compiled_pattern = NULL;
 static PatternCompileCase arg_case = PATTERN_COMPILE_CASE_AUTO;
@@ -162,6 +164,7 @@ static enum {
         ACTION_ROTATE_AND_VACUUM,
         ACTION_LIST_FIELDS,
         ACTION_LIST_FIELD_NAMES,
+        ACTION_CONVERT,
 } arg_action = ACTION_SHOW;
 
 typedef struct BootId {
@@ -387,6 +390,7 @@ static int help(void) {
                "     --dump-catalog          Show entries in the message catalog\n"
                "     --update-catalog        Update the message catalog database\n"
                "     --setup-keys            Generate a new FSS key pair\n"
+               "     --convert=PATH          Convert the journal to the latest journal format\n"
                "\nSee the %2$s for details.\n",
                program_invocation_short_name,
                link,
@@ -441,6 +445,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_NO_HOSTNAME,
                 ARG_OUTPUT_FIELDS,
                 ARG_NAMESPACE,
+                ARG_CONVERT,
         };
 
         static const struct option options[] = {
@@ -508,6 +513,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "no-hostname",          no_argument,       NULL, ARG_NO_HOSTNAME          },
                 { "output-fields",        required_argument, NULL, ARG_OUTPUT_FIELDS        },
                 { "namespace",            required_argument, NULL, ARG_NAMESPACE            },
+                { "convert",              required_argument, NULL, ARG_CONVERT              },
                 {}
         };
 
@@ -1033,6 +1039,11 @@ static int parse_argv(int argc, char *argv[]) {
                         }
                         break;
                 }
+
+                case ARG_CONVERT:
+                        arg_action = ACTION_CONVERT;
+                        arg_convert = optarg;
+                        break;
 
                 case '?':
                         return -EINVAL;
@@ -2093,6 +2104,52 @@ static int wait_for_change(sd_journal *j, int poll_fd) {
         return 0;
 }
 
+static int journal_convert(sd_journal *j) {
+        _cleanup_(managed_journal_file_closep) ManagedJournalFile *to = NULL;
+        _cleanup_(mmap_cache_unrefp) MMapCache *mmap = NULL;
+        int r;
+
+        assert(arg_convert);
+
+        mmap = mmap_cache_new();
+        if (!mmap)
+                return -ENOMEM;
+
+        r = managed_journal_file_open(-1, arg_convert, O_RDWR | O_CREAT, JOURNAL_COMPRESS, 0640, UINT64_MAX,
+                                      &(JournalMetrics) { -1, -1, -1, -1, -1, -1 }, mmap, NULL, NULL, &to);
+        if (r < 0)
+                return log_error_errno(r, "Failed to open journal: %m");
+
+        SD_JOURNAL_FOREACH(j) {
+                Object *o;
+                JournalFile *from;
+
+                from = j->current_file;
+                assert(from && from->current_offset > 0);
+
+                r = journal_file_move_to_object(from, OBJECT_ENTRY, from->current_offset, &o);
+                if (r < 0)
+                        return log_error_errno(r, "Can't read entry: %m");
+
+                r = journal_file_copy_entry(from, to->file, o, from->current_offset);
+                if (r >= 0)
+                        continue;
+
+                if (!journal_shall_try_append_again(to->file, r))
+                        return log_error_errno(r, "Can't write entry: %m");
+
+                r = managed_journal_file_rotate(&to, mmap, JOURNAL_COMPRESS, UINT64_MAX, NULL);
+                if (r < 0)
+                        return r;
+
+                r = journal_file_copy_entry(from, to->file, o, from->current_offset);
+                if (r < 0)
+                        return log_error_errno(r, "Can't write entry: %m");
+        }
+
+        return 0;
+}
+
 int main(int argc, char *argv[]) {
         _cleanup_(loop_device_unrefp) LoopDevice *loop_device = NULL;
         _cleanup_(umount_and_rmdir_and_freep) char *unlink_dir = NULL;
@@ -2203,6 +2260,7 @@ int main(int argc, char *argv[]) {
         case ACTION_ROTATE_AND_VACUUM:
         case ACTION_LIST_FIELDS:
         case ACTION_LIST_FIELD_NAMES:
+        case ACTION_CONVERT:
                 /* These ones require access to the journal files, continue below. */
                 break;
 
@@ -2356,6 +2414,10 @@ int main(int argc, char *argv[]) {
         case ACTION_SHOW:
         case ACTION_LIST_FIELDS:
                 break;
+
+        case ACTION_CONVERT:
+                r = journal_convert(j);
+                goto finish;
 
         default:
                 assert_not_reached();
--- a/src/journal/journald-server.c
+++ b/src/journal/journald-server.c
@@ -30,6 +30,7 @@
 #include "io-util.h"
 #include "journal-authenticate.h"
 #include "journal-internal.h"
+#include "journal-util.h"
 #include "journal-vacuum.h"
 #include "journald-audit.h"
 #include "journald-context.h"
@@ -769,55 +770,6 @@ static void server_cache_hostname(Server *s) {
         free_and_replace(s->hostname_field, x);
 }
 
-static bool shall_try_append_again(JournalFile *f, int r) {
-        switch (r) {
-
-        case -E2BIG:           /* Hit configured limit          */
-        case -EFBIG:           /* Hit fs limit                  */
-        case -EDQUOT:          /* Quota limit hit               */
-        case -ENOSPC:          /* Disk full                     */
-                log_debug("%s: Allocation limit reached, rotating.", f->path);
-                return true;
-
-        case -EIO:             /* I/O error of some kind (mmap) */
-                log_warning("%s: IO error, rotating.", f->path);
-                return true;
-
-        case -EHOSTDOWN:       /* Other machine                 */
-                log_info("%s: Journal file from other machine, rotating.", f->path);
-                return true;
-
-        case -EBUSY:           /* Unclean shutdown              */
-                log_info("%s: Unclean shutdown, rotating.", f->path);
-                return true;
-
-        case -EPROTONOSUPPORT: /* Unsupported feature           */
-                log_info("%s: Unsupported feature, rotating.", f->path);
-                return true;
-
-        case -EBADMSG:         /* Corrupted                     */
-        case -ENODATA:         /* Truncated                     */
-        case -ESHUTDOWN:       /* Already archived              */
-                log_warning("%s: Journal file corrupted, rotating.", f->path);
-                return true;
-
-        case -EIDRM:           /* Journal file has been deleted */
-                log_warning("%s: Journal file has been deleted, rotating.", f->path);
-                return true;
-
-        case -ETXTBSY:         /* Journal file is from the future */
-                log_warning("%s: Journal file is from the future, rotating.", f->path);
-                return true;
-
-        case -EAFNOSUPPORT:
-                log_warning("%s: underlying file system does not support memory mapping or another required file system feature.", f->path);
-                return false;
-
-        default:
-                return false;
-        }
-}
-
 static void write_to_journal(Server *s, uid_t uid, struct iovec *iovec, size_t n, int priority) {
         bool vacuumed = false, rotate = false;
         struct dual_timestamp ts;
@@ -872,7 +824,7 @@ static void write_to_journal(Server *s, uid_t uid, struct iovec *iovec, size_t n
                 return;
         }
 
-        if (vacuumed || !shall_try_append_again(f->file, r)) {
+        if (vacuumed || !journal_shall_try_append_again(f->file, r)) {
                 log_ratelimit_full_errno(LOG_ERR, r, "Failed to write entry (%zu items, %zu bytes), ignoring: %m", n, IOVEC_TOTAL_SIZE(iovec, n));
                 return;
         }
@@ -1202,7 +1154,7 @@ int server_flush_to_var(Server *s, bool require_flag_file) {
                 if (r >= 0)
                         continue;
 
-                if (!shall_try_append_again(s->system_journal->file, r)) {
+                if (!journal_shall_try_append_again(s->system_journal->file, r)) {
                         log_error_errno(r, "Can't write entry: %m");
                         goto finish;
                 }
--- a/src/shared/journal-util.c
+++ b/src/shared/journal-util.c
@@ -136,3 +136,52 @@ int journal_access_check_and_warn(sd_journal *j, bool quiet, bool want_other_use
 
         return r;
 }
+
+bool journal_shall_try_append_again(JournalFile *f, int r) {
+        switch (r) {
+
+        case -E2BIG:           /* Hit configured limit          */
+        case -EFBIG:           /* Hit fs limit                  */
+        case -EDQUOT:          /* Quota limit hit               */
+        case -ENOSPC:          /* Disk full                     */
+                log_debug("%s: Allocation limit reached, rotating.", f->path);
+                return true;
+
+        case -EIO:             /* I/O error of some kind (mmap) */
+                log_warning("%s: IO error, rotating.", f->path);
+                return true;
+
+        case -EHOSTDOWN:       /* Other machine                 */
+                log_info("%s: Journal file from other machine, rotating.", f->path);
+                return true;
+
+        case -EBUSY:           /* Unclean shutdown              */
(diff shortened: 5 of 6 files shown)


DaanDeMeyer:
@mrc0mmand I'm not downgrading the clang used by specifying the version explicitly am I? I vaguely remember a PR from you that switched the minimum clang to 12 somewhere. Was that in the CI job I'm modifying here?

mrc0mmand:
> @mrc0mmand I'm not downgrading the clang used by specifying the version explicitly am I? I vaguely remember a PR from you that switched the minimum clang to 12 somewhere. Was that in the CI job I'm modifying here?

Nope, this is fine, the clang-12 shenanigans were in the *Build test* CI job.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1683,7 +1683,15 @@ uint64_t journal_file_entry_array_n_items(Object *o) {
...
 
-        return (sz - offsetof(Object, entry_array.items)) / sizeof(uint64_t);
+        return (sz - offsetof(Object, entry_array.items)) / journal_file_entry_array_item_size(f);
+}
+
+uint64_t journal_file_entry_array_item(JournalFile *f, Object *o, size_t i) {
+        assert(o);
+        assert(o->object.type == OBJECT_ENTRY_ARRAY);
+
+        return JOURNAL_HEADER_COMPACT(f->header) ? (uint64_t) le32toh(o->entry_array.items.compact[i]) :
```
keszybz: Is the cast actually needed? I'd think that an up-cast is implicitly OK.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -2574,14 +2593,14 @@ static int generic_array_bisect(
                 return 0;
 
         /* Let's cache this item for the next invocation */
-        chain_cache_put(f->chain_cache, ci, first, a, le64toh(array->entry_array.items[0]), t, subtract_one ? (i > 0 ? i-1 : UINT64_MAX) : i);
+        chain_cache_put(f->chain_cache, ci, first, a, journal_file_entry_array_item(f, array, 0), t, subtract_one ? (i > 0 ? i-1 : UINT64_MAX) : i);
```
keszybz, line 2708: Line break this?

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1927,10 +1924,16 @@ static int journal_file_append_entry_internal(
...
+
         if (JOURNAL_HEADER_COMPACT(f->header)) {
+                if (f->header->head_entry_monotonic == 0)
+                        f->header->head_entry_monotonic = htole64(ts->monotonic);
+
                 o->entry.compact.seqnum = htole64(journal_file_new_entry_seqnum(f, seqnum));
-                o->entry.compact.realtime = htole64(ts->realtime);
-                o->entry.compact.monotonic = htole64(ts->monotonic);
+                o->entry.compact.realtime = htole32(ts->realtime - le64toh(f->header->head_entry_realtime));
+                o->entry.compact.monotonic = htole32(ts->monotonic - le64toh(f->header->head_entry_monotonic));
```
keszybz: Shouldn't this raise an error so that the file is rotated? Or at least saturate to the maximum value to avoid wraparound? Let's say that we're appending to a file which was created on a system with no RTC, so the first few entries are at time close to zero, and then we start getting entries with normal timestamps…

src/libsystemd/sd-journal/journal-file.h
```diff
@@ -201,13 +201,13 @@ static inline size_t journal_file_entry_seqnum_offset(JournalFile *f) {
 
 static inline uint64_t journal_file_entry_realtime(JournalFile *f, Object *o) {
         return JOURNAL_HEADER_COMPACT(f->header)
-                        ? le64toh(o->entry.compact.realtime)
+                        ? le32toh(o->entry.compact.realtime) + le64toh(f->header->head_entry_realtime)
```
keszybz: saturate_add ?

src/libsystemd/sd-journal/journal-file.h
```diff
@@ -201,13 +201,13 @@ static inline size_t journal_file_entry_seqnum_offset(JournalFile *f) {
...
                         : le64toh(o->entry.regular.realtime);
 }
 
 static inline uint64_t journal_file_entry_monotonic(JournalFile *f, Object *o) {
         return JOURNAL_HEADER_COMPACT(f->header)
-                        ? le64toh(o->entry.compact.monotonic)
+                        ? le32toh(o->entry.compact.monotonic) + le64toh(f->header->head_entry_monotonic)
```
keszybz: saturate_add ?

src/libsystemd/sd-journal/journal-file.h
```diff
@@ -189,7 +189,7 @@ static inline bool VALID_EPOCH(uint64_t u) {
 
 static inline uint64_t journal_file_entry_seqnum(JournalFile *f, Object *o) {
         return JOURNAL_HEADER_COMPACT(f->header)
-                        ? le64toh(o->entry.compact.seqnum)
+                        ? le32toh(o->entry.compact.seqnum) + le64toh(f->header->head_entry_seqnum)
```
keszybz: Same here?

docs/JOURNAL_FILE_FORMAT.md
```diff
@@ -518,7 +517,10 @@ the format indicated by the **compact** struct instead of the **regular** struct
 The **realtime** field is stored as a 32-bit offset to the **head_entry_realtime**
 field in the header. Similarly, The **monotonic** field is stored as a 32-bit
 offset to the **head_entry_monotonic** field in the header. The **seqnum** field is
-stored as a 32-bit offset to the **head_entry_seqnum** field in the header.
+stored as a 32-bit offset to the **head_entry_seqnum** field in the header. The
+boot ID is not stored per entry anymore and is always identical to the
+**boot_id** field in the header. As a consequence, journal files written in
+compact mode are limited to a single boot ID.
```
keszybz: Hmm. I have to say that this seems like a potentially bad tradeoff. Some machines reboot regularly, and this will force rotation on every reboot. What about a slightly different approach: in the header, create an array of up to 16 boot_ids, and a corresponding array of seqnums. When looking up the boot_id for a seqnum, find the entry in the second array with the highest entry smaller than the seqnum at hand, and return the boot_id in the same position. This would give essentially the same size savings without forcing rotation…

keszybz (changes requested):
This looks very nice. I'm impressed how this is all so nicely split into patches, this makes the review much easier than it'd otherwise be. Apart from some stylistic comments, my only real gripe is with the last patch.

src/libsystemd/sd-journal/journal-file.c
daandemeyer: Added some logic to check that the timestamps aren't too early or too far in the future. We return an error if that's the case

DaanDeMeyer:
I just realized using a 32-bit offset for a microsecond timestamp only covers 71 minutes. Unless we drop the precision of the timestamp offsets, we'd be rotating every hour even on systems that hardly produce any logs. Dropping to millisecond precision gives us 50 days for a single journal file, but then you're getting a significant drop in timestamp precision when using compact mode.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -318,13 +318,22 @@ static int journal_file_init_header(JournalFile *f, JournalFileFlags file_flags,
         seal = FLAGS_SET(file_flags, JOURNAL_SEAL) && journal_file_fss_load(f) >= 0;
 #endif
 
+        r = getenv_bool("SYSTEMD_JOURNAL_COMPACT");
```
poettering: add doc to ENVIRONMENT.md

src/libsystemd/sd-journal/journal-file.h
poettering: hmm, overflow check in case rogue file as super high seqnum set in header? not sure what a good strategy is in this case though, maybe return an invalid seqnum (i.e. UINT64_MAX) and then catch that everywhere and make it an EBADMSG issue

src/libsystemd/sd-journal/journal-file.h
```diff
@@ -123,8 +123,9 @@ typedef struct JournalFile {
 } JournalFile;
 
 typedef enum JournalFileFlags {
-        JOURNAL_COMPRESS = 1 << 0,
-        JOURNAL_SEAL     = 1 << 1,
+        JOURNAL_COMPRESS   = 1 << 0,
+        JOURNAL_SEAL       = 1 << 1,
+        JOURNAL_MULTI_BOOT = 1 << 2,
```
poettering: maybe add a brief comment here, as what this is for, not's not obvious I think to most readers

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -318,13 +318,17 @@ static int journal_file_init_header(JournalFile *f, JournalFileFlags file_flags,
...
 
-        r = getenv_bool("SYSTEMD_JOURNAL_COMPACT");
-        if (r < 0) {
-                if (r != -ENXIO)
-                        log_debug_errno(r, "Failed to parse $SYSTEMD_JOURNAL_COMPACT environment variable, ignoring: %m");
-                compact = true;
-        } else
-                compact = r;
+        if (FLAGS_SET(file_flags, JOURNAL_MULTI_BOOT))
+                compact = false;
```
poettering: maybe add explicit comment here: `/* the compact mode does not support multiple different boot ids per journal file, hence disable compat mode if that's requested */

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -370,10 +374,6 @@ static int journal_file_refresh_header(JournalFile *f) {
         else if (r < 0)
                 return r;
 
-        r = sd_id128_get_boot(&f->header->boot_id);
-        if (r < 0)
-                return r;
```
poettering: what's the benefit of initializing this lazily? It's unlikely to get you much.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1920,16 +1921,29 @@ static int journal_file_append_entry_internal(
...
+                        return -EXDEV;
+
                 o->entry.compact.seqnum = htole32(journal_file_new_entry_seqnum(f, seqnum) - le64toh(f->header->head_entry_seqnum));
                 o->entry.compact.realtime = htole64(ts->realtime);
                 o->entry.compact.monotonic = htole64(ts->monotonic);
-                o->entry.compact.boot_id = f->header->boot_id;
                 o->entry.compact.xor_hash = htole64(xor_hash);
                 memcpy_safe(o->entry.compact.items, items, n_items * sizeof(EntryItem));
         } else {
+                if (!sd_id128_equal(f->header->boot_id, b))
```
poettering: this check is redundant... if it's already equal you might as well set it to b again, it won't change a thing

poettering (changes requested):
looks good, some comments

poettering:
Hmm, so here's an idea: what about adding a flag to the `flags` field in the `ObjectHeader` indicating that something is a compact object? first of all I think that would be nice because objects would then be self-descriptive: look at an object without other source of truth and you know what you have. In particular good for recovery scenarios.

But more imporantly: we could then allow mixing compact and non-compat objects inside a journal file which has compat mode enabled. Specifically, we'd always write compact objects except if what we are about to write doesn't "fit" in a compact object, for example because the boot id field doesn#t match, or because the timestamp overran 32bit and son.

This could then allow us to relax restrictions a bit: as long as the majority of records fit into the compact fields we are good. But if some objects don't, it's still OK, all we lost is a tiny bit of compactness.

poettering:
The EntryObject size is not aligned to 64bit anymore. It probably should be. What about changing the fields to:
```
le32_t seqnum_lo;
le32_t realtime_lo;
le32_t monotonic_lo;
le8_t seqnum_hi;
le8_t realtime_hi;
le8_t monotonic_hi;
le8_t _reserved;
le64_t xor_hash;
EntryItem items[];
```

With that, items[] starts on a 64bit boundary again, and you have 40 bits for the three timestamps/sequence numbers, which with µs granularity should give you more than 10 days of coverage.

and you save 4 bytes compared to your current definition (28 vs. 24 bytes) and even still got 8 free bytes for other uses left.

DaanDeMeyer:
Reworked PR to just include the 3 most impactful changes that didn't require large changes to the journal format.

1. Store entry array offsets as 32-bit
2. Store entry item offsets as 32-bit and drop the unused hash field from entry items
3. Add offsets to the tail entry array to the header and individual data objects to speed up appending entries

This halves the space required by my laptop's journal.

man/journalctl.xml
```diff
@@ -854,6 +854,17 @@
...
 
+      <varlistentry>
+        <term><option>--convert=</option></term>
+
+        <listitem><para>Converts the specified journal files to the latest supported journal format. Takes
+        the path to store the converted journal files. The path should include the filename to be used for
+        the converted files, with the <literal>.journal</literal> extension (e.g.
+        <filename>/a/b/c/converted.journal</filename> will store the journal files in the
+        <filename>/a/b/c</filename> directory using <filename>converted.journal</filename> as the filename).
+        </para></listitem>
```
poettering, line 865: I find this very confusing? "The specified journal files"? what does that mean, which ones? If I invoke "journalctl --convert=foo.journal" then I am not specifiying any source files.

What about calling this "journalctl --rewrite=…" or so? because you are not really just converting, you are entirely rewriting stuff, i.e. journal files might be differently split up afterwards.

And maybe say the path specified is the output journal file path, but if rotation is taking place files will be placed to the specified file in the same directory.

docs/ENVIRONMENT.md
```diff
@@ -468,3 +468,9 @@ SYSTEMD_HOME_DEBUG_SUFFIX=foo \
   when kernel-install is invoked. This can be useful if kernel-install is invoked
   unconditionally as a child process by another tool, such as package managers
   running kernel-install in a postinstall script.
+
+`systemd-journald`:
+
+* `$SYSTEMD_JOURNAL_COMPACT` - Takes a boolean. If enabled, journal files are written
+  in a more compact format that reduces the amount of disk space required by the
+  journal. Enabled by default.
```
poettering: maybe say explicitly that individual journal files are limited to 4G in size in this case, because this is technically what it boils down to.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -318,13 +318,22 @@ static int journal_file_init_header(JournalFile *f, JournalFileFlags file_flags,
         seal = FLAGS_SET(file_flags, JOURNAL_SEAL) && journal_file_fss_load(f) >= 0;
 #endif
 
+        r = getenv_bool("SYSTEMD_JOURNAL_COMPACT");
+        if (r < 0) {
+                if (r != -ENXIO)
+                        log_debug_errno(r, "Failed to parse $SYSTEMD_JOURNAL_COMPACT environment variable, ignoring: %m");
+                compact = true;
+        } else
+                compact = r;
```
poettering: maybe split this out into a helper function `compact_mode_requested()` or so?

src/libsystemd/sd-journal/journal-def.h
```diff
@@ -117,7 +117,10 @@ struct HashTableObject {
 struct EntryArrayObject {
         ObjectHeader object;
         le64_t next_entry_array_offset;
-        le64_t items[];
+        union {
+                le64_t regular[0];
+                le32_t compact[0];
```
poettering, line 133: i think i asked this in a previous review, but i forgot: why specify the size here? why not leave it empty? i.e. `[]` rather than `[0]`?

poettering:
looks good. just some nitpicking over the --convert= switch name.

src/libsystemd/sd-journal/journal-def.h
daandemeyer: Variable length arrays in unions are not standard C, but do work in GCC this way. Just try to define a union containing an array without size and see if it compiles.

man/journalctl.xml
daandemeyer: I removed this commit from the PR, I'll prepare a separate PR for it with the requested fixed

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -295,24 +295,39 @@ JournalFile* journal_file_close(JournalFile *f) {
...
+        r = getenv_bool("SYSTEMD_JOURNAL_KEYED_HASH");
+        if (r >= 0)
+                return r;
+        if (r != -ENXIO)
+                log_debug_errno(r, "Failed to parse $SYSTEMD_JOURNAL_COMPACT environment variable, ignoring: %m");
+
+        return true;
+}
+
+
```
poettering: spurious double empty line

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -324,7 +339,8 @@ static int journal_file_init_header(JournalFile *f, JournalFileFlags file_flags,
         h.incompatible_flags |= htole32(
                         FLAGS_SET(file_flags, JOURNAL_COMPRESS) *
                         COMPRESSION_TO_HEADER_INCOMPATIBLE_FLAG(DEFAULT_COMPRESSION) |
-                        keyed_hash * HEADER_INCOMPATIBLE_KEYED_HASH);
+                        keyed_hash_requested() * HEADER_INCOMPATIBLE_KEYED_HASH |
+                        compact_mode_requested() * HEADER_INCOMPATIBLE_COMPACT);
```
poettering, line 342: ah so much nicer!

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -295,24 +295,38 @@ JournalFile* journal_file_close(JournalFile *f) {
...
 }
 
+static bool keyed_hash_requested(void) {
+        int r;
+
+        r = getenv_bool("SYSTEMD_JOURNAL_KEYED_HASH");
+        if (r >= 0)
+                return r;
+        if (r != -ENXIO)
+                log_debug_errno(r, "Failed to parse $SYSTEMD_JOURNAL_COMPACT environment variable, ignoring: %m");
```
yuwata: Copy-and-paste error: `"$SYSTEMD_JOURNAL_KEYED_HASH"`

src/journal/test-journal-flush.c
```diff
@@ -70,6 +71,16 @@ int main(int argc, char *argv[]) {
...
+static int intro(void) {
+        assert_se(setenv("SYSTEMD_JOURNAL_COMPACT", "0", 1) >= 0);
+        run_test(saved_argc, saved_argv);
 
-        return 0;
+        assert_se(setenv("SYSTEMD_JOURNAL_COMPACT", "1", 1) >= 0);
+        run_test(saved_argc, saved_argv);
+
+        return EXIT_SUCCESS;
 }
```
yuwata: This is just a test code, hence no problem, but running tests in `intro()` is not a good idea. We usually use `intro()` for checking testing condition or so...

src/journal/test-journal-interleaving.c
```diff
@@ -302,6 +302,12 @@ static int intro(void) {
 
         arg_keep = saved_argc > 1;
 
+        assert_se(setenv("SYSTEMD_JOURNAL_COMPACT", "0", 1) >= 0);
+        test_sequence_numbers();
+
+        assert_se(setenv("SYSTEMD_JOURNAL_COMPACT", "1", 1) >= 0);
+        test_sequence_numbers();
+
```
yuwata: Also here.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1840,8 +1840,8 @@ static int journal_file_link_entry_item(JournalFile *f, Object *o, uint64_t offs
                                               offset);
 }
 
-static int journal_file_link_entry(JournalFile *f, Object *o, uint64_t offset) {
-        uint64_t n;
+static int journal_file_link_entry(
+                JournalFile *f, Object *o, uint64_t offset, const EntryItem items[], size_t n_items) {
```
yuwata: We usually break lines for arguments one by one.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1405,42 +1404,13 @@ int journal_file_find_data_object_with_hash(
...
-                        }
-#else
-                        return -EPROTONOSUPPORT;
-#endif
-                } else if (le64toh(o->object.size) == osize &&
-                           memcmp(o->data.payload, data, size) == 0) {
+                r = journal_file_data_payload(f, o, p, NULL, 0, 0, &d, &rsize);
+                if (r < 0)
+                        return r;
+                assert(r > 0); /* journal_file_data_payload() always returns > 0 if no field is provided. */
```
yuwata, line 1417: The comment looks not correct for me. `maybe_decompress_payload()` may return 0.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1658,6 +1628,103 @@ static int journal_file_append_data(
...
+                if (field) {
+                        r = decompress_startswith(compression, payload, size, &f->compress_buffer, field,
+                                                  field_length, '=');
+                        if (r < 0)
+                                return log_debug_errno(r,
+                                                       "Cannot decompress %s object of length %" PRIu64 ": %m",
+                                                       compression_to_string(compression),
+                                                       size);
+                        if (r == 0)
+                                return 0;
```
yuwata: We usually initialize `ret_` arguments on success.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -1658,6 +1628,103 @@ static int journal_file_append_data(
...
+                if (ret_data)
+                        *ret_data = f->compress_buffer;
+                if (ret_size)
+                        *ret_size = rsize;
+#else
+                return -EPROTONOSUPPORT;
+#endif
+        } else {
+                if (field && (size < field_length + 1 || memcmp(payload, field, field_length) != 0 || payload[field_length] != '='))
+                        return 0;
```
yuwata, line 1696: Also here.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -3802,51 +3869,20 @@ int journal_file_copy_entry(JournalFile *from, JournalFile *to, Object *o, uint6
...
-                                return r;
-
-                        data = from->compress_buffer;
-                        l = rsize;
-#else
-                        return -EPROTONOSUPPORT;
-#endif
-                } else
-                        data = o->data.payload;
+                assert(r > 0);
```
yuwata, line 3924: This can be also triggered...

src/libsystemd/sd-journal/sd-journal.c
```diff
@@ -2437,23 +2336,21 @@ _public_ int sd_journal_enumerate_data(sd_journal *j, const void **data, size_t
...
                 p = journal_file_entry_item_object_offset(f, o, j->current_field);
-                r = journal_file_move_to_object(f, OBJECT_DATA, p, &o);
+                r = journal_file_data_payload(f, NULL, p, NULL, 0, j->data_threshold, &d, &l);
                 if (IN_SET(r, -EADDRNOTAVAIL, -EBADMSG)) {
                         log_debug_errno(r, "Entry item %"PRIu64" data object is bad, skipping over it: %m", j->current_field);
                         continue;
                 }
                 if (r < 0)
                         return r;
+                assert(r > 0);
```
yuwata, line 2350: Also here.

yuwata (changes requested):

src/libsystemd/sd-journal/journal-file.c
daandemeyer: Only if 'field' is not 'NULL', so 'r' must be larger than zero or negative here.

DaanDeMeyer:
Adressed the issues, the asserts should not be triggered if "field" is "NULL", PTAL

bluca (approved):

poettering:
Hmm, so this was now committed under the "--convert" name. i still think that's a bad name. since it suggests we'd convert individual files, but we don't. we convert a whole stream that is a combination of files and chop it up potentially very differently. Hence I#d call this "--rewrite" really.. see above

poettering:
Hmm, looking back: wasn't the --convert commit supposed to be removed from the PR before merge? what happened?

yuwata:
Yeah, I haven't reviewed the convert commit...

DaanDeMeyer:
Ah I messed up, let me revert

PR #22999, opened 2022-04-06
core: fix dm-verity auto-discovery in MountImageUnit()

bluca:
The implementation of MountImageUnit()/systemctl mount-image was
changed to use a /proc/self/fd path as the source, but that causes
the dm-verity files autodiscovery to fail, as it looks for files
in the same directory as the image.

Use the original file path when setting up dm-verity.

### DIFF
--- a/src/core/namespace.c
+++ b/src/core/namespace.c
@@ -1216,7 +1216,7 @@ static int mount_image(const MountEntry *m, const char *root_directory) {
         }
 
         r = verity_dissect_and_mount(
-                                mount_entry_source(m), mount_entry_path(m), m->image_options,
+                                /* src_fd= */ -1, mount_entry_source(m), mount_entry_path(m), m->image_options,
                                 host_os_release_id, host_os_release_version_id, host_os_release_sysext_level, NULL);
         if (r == -ENOENT && m->ignore)
                 return 0;
--- a/src/shared/mount-util.c
+++ b/src/shared/mount-util.c
@@ -789,6 +789,7 @@ static int mount_in_namespace(
         bool mount_slave_created = false, mount_slave_mounted = false,
                 mount_tmp_created = false, mount_tmp_mounted = false,
                 mount_outside_created = false, mount_outside_mounted = false;
+        _cleanup_free_ char *chased_src_path = NULL;
         struct stat st, self_mntns_st;
         pid_t child;
         int r;
@@ -826,9 +827,10 @@ static int mount_in_namespace(
         if (r < 0)
                 return log_debug_errno(r == -ENOENT ? SYNTHETIC_ERRNO(EOPNOTSUPP) : r, "Target does not allow propagation of mount points");
 
-        r = chase_symlinks(src, NULL, CHASE_TRAIL_SLASH, NULL, &chased_src_fd);
+        r = chase_symlinks(src, NULL, 0, &chased_src_path, &chased_src_fd);
         if (r < 0)
                 return log_debug_errno(r, "Failed to resolve source path of %s: %m", src);
+        log_debug("Chased source path of %s to %s", src, chased_src_path);
 
         if (fstat(chased_src_fd, &st) < 0)
                 return log_debug_errno(errno, "Failed to stat() resolved source path %s: %m", src);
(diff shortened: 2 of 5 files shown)


yuwata (approved):
LGTM.

src/shared/mount-util.c
```diff
@@ -872,12 +872,22 @@ static int mount_in_namespace(
...
         mount_tmp_created = true;
 
-        if (is_image)
-                r = verity_dissect_and_mount(FORMAT_PROC_FD_PATH(chased_src_fd), mount_tmp, options, NULL, NULL, NULL, NULL);
-        else
+        if (is_image) {
+                _cleanup_free_ char *chased_src_path = NULL;
+
+                /* For the dm-verity autodiscovery to work, we need to pass the original path. */
+                r = fd_get_path(chased_src_fd, &chased_src_path);
```
poettering: hmm, why retrieve the path from the fd? chase_symlinks() rturns it anyway if you want? is this about pinning?

btw, the chase_symlinks() call above uses CHASE_TRAIL_SLASH flag which I figure makes no sense right now, if the path is not requested anyway...?

src/shared/mount-util.c
```diff
@@ -872,12 +872,22 @@ static int mount_in_namespace(
...
+
+                r = verity_dissect_and_mount(chased_src_path, mount_tmp, options, NULL, NULL, NULL, NULL);
+                if (r < 0)
+                        goto finish;
+        } else {
                 r = mount_follow_verbose(LOG_DEBUG, FORMAT_PROC_FD_PATH(chased_src_fd), mount_tmp, NULL, MS_BIND, NULL);
-        if (r < 0)
-                goto finish;
+                if (r < 0)
+                        goto finish;
```
poettering: you can move these two lines after the if/else block and share it between both branches

poettering:
so staring at this for a while: this doesn't look right. the current implementation pins the source inode, throughout the whole operation. Which is nice. With this change we stop pinning, and go by path again. which is quite a step back.

I think it's OK to search sidecar files via the path, but I think we should retain the pinning of the image file.

Hence, I'd propose that verity_dissect_and_mount() is changed, so that it gets two input parameters: `const char *src_path` *and* `int src_fd`, where the latter can be -1. 

Then, use src_fd whenever we can, when it is >= 0. But if it isn't, go by src_path instead.

(fd_get_path() is something we can use when we really have to, but if we have the path anyway through other means then please always use that, because fd_get_path() is kinda hosed when it comes to the `(deleted)` suffix).

bluca:
Updated as suggested

src/core/namespace.c
```diff
@@ -1216,7 +1216,7 @@ static int mount_image(const MountEntry *m, const char *root_directory) {
         }
 
         r = verity_dissect_and_mount(
-                                mount_entry_source(m), mount_entry_path(m), m->image_options,
+                                -1 /* src_fd */, mount_entry_source(m), mount_entry_path(m), m->image_options,
```
poettering: we usually write these commands like this:

```c
…, /* src_fd= */ -1, …
```

i.e. write them before the argument, and sufficed by `=`

src/shared/mount-util.c
```diff
@@ -826,7 +827,7 @@ static int mount_in_namespace(
         if (r < 0)
                 return log_debug_errno(r == -ENOENT ? SYNTHETIC_ERRNO(EOPNOTSUPP) : r, "Target does not allow propagation of mount points");
 
-        r = chase_symlinks(src, NULL, CHASE_TRAIL_SLASH, NULL, &chased_src_fd);
+        r = chase_symlinks(src, NULL, 0, is_image ? &chased_src_path : NULL, &chased_src_fd);
```
poettering: I'd actually be sloppy here, and drop the conditionalization. it's OK to acquire the path here, even if we don#t need it in some cases.

Or to say this differently: i much prefer code with the least number of conditionalizations, if this is not performance sensitive. And i'd rather see the benefit in having the path here, and add a log_debug() line saying to which file we traced things...

bluca: Changed, and added log_debug

keszybz:
I don't grok what is happening here, but the code changes look superficially ok ;)

PR #23000, opened 2022-04-06
macro: call __gcov_dump() before _exit() w/ coverage enabled

mrc0mmand:
_exit() skips at-exit hooks, causing lost coverage from processes
utilizing it.

Hopefully resolves systemd/systemd-centos-ci#482

---

/cc @yuwata - this feels pretty nasty, but I'm not sure if there's a better way. If so, please let me know. Marking as a draft to make sure it works everywhere as expected (and in hopes someone comes up with a more elegant solution), but from my local tests I finally get coverage from code run by udev workers.

### DIFF
--- a/test/TEST-02-UNITTESTS/test.sh
+++ b/test/TEST-02-UNITTESTS/test.sh
@@ -4,6 +4,10 @@ set -e
 
 TEST_DESCRIPTION="Run unit tests under containers"
 RUN_IN_UNPRIVILEGED_CONTAINER=yes
+# Some tests make collecting coverage impossible (like test-mount-util, which
+# remounts the whole / as read-only), so let's ignore the gcov errors in such
+# case
+IGNORE_MISSING_COVERAGE=yes
 
 # embed some newlines in the kernel command line to stress our test suite
 KERNEL_APPEND="
--- a/test/test-functions
+++ b/test/test-functions
@@ -1406,8 +1406,8 @@ check_coverage_reports() {
     # usually due to the sandbox being too restrictive (e.g. ProtectSystem=yes,
     # ProtectHome=yes) or the $BUILD_DIR being inaccessible to non-root users - see
     # `setfacl` stuff in install_compiled_systemd().
-
-    if "${JOURNALCTL:?}" -q --no-pager -D "${root:?}/var/log/journal" --grep "profiling:.+?gcda:[Cc]annot open"; then
+    if ! get_bool "${IGNORE_MISSING_COVERAGE:=}" && \
+       "${JOURNALCTL:?}" -q --no-pager -D "${root:?}/var/log/journal" --grep "profiling:.+?gcda:[Cc]annot open"; then
         derror "Detected possibly missing coverage, check the journal"
         return 1
     fi


yuwata (approved):
Interesting. If it works, then I think that's fine.

poettering:
i doubt we can do that. we only call _exit() in child processes, where we explicitly don't want the cleanup handlers to run, because the are assumed to run in the child...

is gcov even ready to be called in forked off processes?

can we call that gcov specific handler explicitly?

yuwata (approved):
Looks better now.

medhefgo:
This does not belong in fundamental headers.

mrc0mmand:
> This does not belong in fundamental headers.

Indeed, will move it.

@yuwata the test build looks promising (we finally got past the 60% mark) - https://coveralls.io/builds/48064359/source?filename=src%2Fudev%2Fudevd.c#L485. I noticed another issue which popped up, I'll deal with it tomorrow (together with this PR).

mrc0mmand:
> > This does not belong in fundamental headers.
> 
> Indeed, will move it.
> 
> @yuwata the test build looks promising (we finally got past the 60% mark) - https://coveralls.io/builds/48064359/source?filename=src%2Fudev%2Fudevd.c#L485. I noticed another issue which popped up, I'll deal with it tomorrow (together with this PR).

The systemd-networkd part looks much better as well https://coveralls.io/builds/48064359/source?filename=src%2Fudev%2Fnet%2Flink-config.c

src/basic/macro.h
```diff
@@ -29,6 +29,19 @@
 #define _alignptr_ __attribute__((__aligned__(sizeof(void*))))
 #define _warn_unused_result_ __attribute__((__warn_unused_result__))
 
+#if defined(BUILT_WITH_COVERAGE) && BUILT_WITH_COVERAGE
+/* We need to explicitly call __gcov_dump() in places where we use _exit(), since
+ * _exit() skips at-exit hooks resulting in lost coverage */
+#  include <unistd.h>
+extern void __gcov_dump(void);
+
+_unused_ _noreturn_ static void _coverage__exit(int status) {
```
poettering: please also make this inline, otherwise it shouldn't be in a header file.

And once you make it `inline` you can drop the `_unused_`, because you don't need it anymore

src/basic/macro.h
```diff
@@ -29,6 +29,19 @@
...
+    __gcov_dump();
+    _exit(status);
```
poettering: please indent by 8 spaces

poettering:
looks great, just some nitpicks.

src/basic/macro.h
mrc0mmand: D'oh!

src/basic/macro.h
mrc0mmand: Indeed, thanks!

poettering:
lgtm. If this works, let's merge (though I presume there might be a reason you marked this as "Draft"?)

mrc0mmand:
> lgtm. If this works, let's merge (though I presume there might be a reason you marked this as "Draft"?)

Yup, there's another issue which popped up after applying this "fix" I need to take care of first.

bluca (approved):

yuwata (approved):

src/basic/macro.h
```diff
@@ -29,6 +29,19 @@
 #define _alignptr_ __attribute__((__aligned__(sizeof(void*))))
 #define _warn_unused_result_ __attribute__((__warn_unused_result__))
 
+#if defined(BUILT_WITH_COVERAGE) && BUILT_WITH_COVERAGE
```
keszybz, line 32: Please drop the `defined(BUILT_WITH_COVERAGE) && `.

The big advantage of using `.set10()` instead of `.set()` is that the variable is always defined and we can catch typos. In ancient times autoconf would only do `#define WITH_FOO`, so the usual check was `#ifdef WITH_FOO`. The problem with that is that it's fairly easy to remove the definition without getting any errors about remaining uses, or make a typo in the name. When we dropped autoconf support finally converted to use `.set10()` and `-Werror=undef` there were a few places with dead code because of messed up `ifdef`s.

src/basic/macro.h
```diff
@@ -29,6 +29,19 @@
...
+ * _exit() skips at-exit hooks resulting in lost coverage */
+#  include <unistd.h>
+extern void __gcov_dump(void);
+
+_noreturn_ static inline void _coverage__exit(int status) {
+        __gcov_dump();
+        _exit(status);
+}
+#  define _exit(x) _coverage__exit(x)
+#endif
```
keszybz, line 43: Hmm, this feels dangerous. It'll only work if `macro.h` is included. If it's not included, the wrapper will not be called [1]. The problem is there will be no indication about the issue, except possibly missing coverage. So I think we need to bite the bullet and invert this: define a function with a different name that just does `_exit` if `!BUILT_WITH_COVERAGE` and replace all calls to `_exit()` with a call to it. (`do__exit()` ?)

[1]:
```console
$ git grep -n -l _exit '*.c' | xargs git grep -L macro.h
man/inotify-watch-tmp.c
man/sd-bus-container-read.c
src/analyze/analyze-exit-status.c
src/analyze/analyze.c
src/busctl/busctl.c
src/core/dbus-cgroup.c
src/core/dbus-execute.c
src/core/dbus-manager.c
src/core/dbus-path.c
src/core/dbus-scope.c
src/core/dbus-service.c
src/core/dbus-socket.c
src/core/dbus-timer.c
src/core/dbus-unit.c
src/core/dbus-util.c
src/core/load-fragment.c
src/core/main.c
src/core/mount.c
src/core/service.c
src/core/socket.c
src/core/swap.c
src/core/timer.c
src/core/unit-serialize.c
src/cryptsetup/cryptsetup.c
src/delta/delta.c
src/fsck/fsck.c
src/fuzz/fuzz-varlink.c
src/home/homectl.c
src/home/homed-home.c
src/home/homed-manager.c
src/home/homework-cifs.c
src/home/homework-fscrypt.c
src/home/homework-luks.c
src/import/export-raw.c
src/import/export-tar.c
src/import/export.c
src/import/import-common.c
src/import/import-raw.c
src/import/import-tar.c
src/import/import.c
src/import/importd.c
src/import/pull-common.c
src/import/pull.c
src/journal-remote/journal-remote-main.c
src/journal-remote/journal-upload.c
src/journal/journald-server.c
src/libsystemd-network/test-dhcp-client.c
src/libsystemd-network/test-ndisc-ra.c
src/libsystemd-network/test-ndisc-rs.c
src/libsystemd/sd-bus/bus-container.c
src/libsystemd/sd-bus/bus-control.c
src/libsystemd/sd-bus/bus-message.c
src/libsystemd/sd-bus/bus-objects.c
src/libsystemd/sd-bus/test-bus-benchmark.c
src/libsystemd/sd-bus/test-bus-marshal.c
src/libsystemd/sd-bus/test-bus-watch-bind.c
src/libsystemd/sd-netlink/netlink-genl.c
src/libsystemd/sd-netlink/netlink-message.c
```

keszybz (commented):
Oh, I was slow with review.

src/basic/macro.h
medhefgo: An alternative would be to put this in its own header and add an '-include' to the compile args.

keszybz: I'd say that our command-lines are long-enough already.

mrc0mmand: > Hmm, this feels dangerous. It'll only work if `macro.h` is included. If it's not included, the wrapper will not be called [1]. The problem is there will be no indication about the issue, except possibly missing coverage. So I think we need to bite the bullet and invert this: define a function with a different name that just does `_exit` if `!BUILT_WITH_COVERAGE` and replace all calls to `_exit()` with a call to it. (`do__exit()` ?)

Yeah, I thought about this, but didn't want to introduce such a big change "just" because of collecting coverage. But I guess there's no better option.

src/basic/macro.h
mrc0mmand: Ah, I forgot it there from the previous version, where the code was located in `macro-fundamental.h` (and `BUILT_WITH_COVERAGE` was not defined when building EFI stuff).

src/basic/macro.h
keszybz: "Big" is relative — this one should be doable with a simple `sed -i` invocation, so the amount of changed lines doesn't really matter.

mrc0mmand: I wonder - by introducing a wrapper for `_exit()` uncoditionally, we'd have to keep that in mind when introducing new code using `_exit()` (or have a Coccinelle transformation for that).

Given all this, I'm kind of inclined to the solution proposed by Jan, which would make it more "future proof" without introducing additional code changes, which are, in the end, useful for only a very specific use case (collecting coverage), thus making all this a bit more "transparent".

keszybz: Dunno, I think an approach with an explicit wrapper would be clearer. But we can try the other way too. Submit a PR and we can see how it looks.

PR #23002, opened 2022-04-06
udev: use child event source

yuwata:


yuwata:
Hmm??
```
systemd-udevd[77062]: sd-device: Failed to get target of '/sys/devices/virtual/net': Too many open files
```

src/udev/udevd.c
```diff
@@ -1544,59 +1554,69 @@ static int on_sighup(sd_event_source *s, const struct signalfd_siginfo *si, void
...
+                result = EVENT_RESULT_SIGNAL_BASE + si->si_status;
+                break;
+
+        case CLD_TRAPPED:
+                log_device_info(dev, "Worker ["PID_FMT"] trapped.", si->si_pid);
+                return 0;
 
-                if ((!WIFEXITED(status) || WEXITSTATUS(status) != 0) && worker->event) {
-                        log_device_error(worker->event->dev, "Worker ["PID_FMT"] failed", pid);
+        case CLD_STOPPED:
```
poettering: these two cannot happen given that you passed WEXITED to sd_event_add_child() and not `WEXITED|WCONTINUED|WSTOPPED`.

hence, maybe just add `assert_nort_reached()` for these two

src/udev/udevd.c
```diff
@@ -1544,59 +1554,69 @@ static int on_sighup(sd_event_source *s, const struct signalfd_siginfo *si, void
...
+                log_device_info(dev, "Worker ["PID_FMT"] continued.", si->si_pid);
+                return 0;
 
-                        /* Forward kernel event to libudev listeners */
-                        device_broadcast(manager->monitor, worker->event->dev,
-                                         WIFEXITED(status) ? EVENT_RESULT_EXIT_STATUS_BASE + WEXITSTATUS(status):
-                                         WIFSIGNALED(status) ? EVENT_RESULT_SIGNAL_BASE + WTERMSIG(status) : 0);
-                }
+        default:
+                log_device_warning(dev, "Worker ["PID_FMT"] failed due to unknown reason (%i).",
```
poettering: hmm, here too: this should not be possible? i.e. assert_not_reached()?

src/libsystemd/sd-event/sd-event.c
```diff
@@ -1465,6 +1460,12 @@ _public_ int sd_event_add_child(
                 e->need_process_child = true;
         }
 
+        r = hashmap_put(e->child_sources, PID_TO_PTR(pid), s);
+        if (r < 0)
+                return r;
```
poettering, line 1471: hmm, if you swap this you need some clean-up patch changes here, i.e. undo the effect of source_child_pidfd_register on failure

poettering: maybe the other option is to move the hashmap_put() further up, i.e. close to the old initialization of `s->child.pid`

yuwata: I think such cleanup is already done in `source_disconnect()`, e.g. unregistering pidfd. Or, do I misunderstand your point?

src/udev/udevd.c
yuwata: How about CLD_TRAPPED?

poettering: shouldn#t happen either... if weonly watch WEXITED we should not get that..

yuwata: Thanks. 

BTW, where can I get the official information about that? The man page wait(2) does not explicitly explain the relation CLD_XXX code vs WXXX flag, especially for trapped case.

yuwata:
The first commit should be backported to stable repository, I think. PTAL.

yuwata:
I think all commits for sd-event, except for the one about comments, should be backported. PTAL.

src/udev/udevd.c
poettering: yeah, it's not explicitly documented. But you can derive it from the description of WIFSTOPPED() in a way...

src/libsystemd/sd-event/sd-event.c
```diff
@@ -706,6 +706,9 @@ static void event_unmask_signal_data(sd_event *e, struct signal_data *d, int sig
                 return;
         }
 
+        if (event_pid_changed(e))
```
poettering, line 709: hm, i am not convinced we shoud check for that here. I mean, it's not supported to use sd-event across forks(), and we generate an error on most entry points already.

I figure we should just change source_free() and event_free() to start with a line like this:

```c
assert_return(!event_pid_changed(e), NULL);
```

this means we'll somewhat gracefully handle attempts to free event or event source objects allocated in children: we just ignore the request and debug log about it.

src/udev/udevd.c
```diff
@@ -265,6 +270,10 @@ static int worker_new(Worker **ret, Manager *manager, sd_device_monitor *worker_
         assert(worker_monitor);
         assert(pid > 1);
 
+        r = sd_event_add_child(manager->event, &s, pid, WEXITED, on_sigchld, manager);
```
poettering: hmm, shouldn't this get the newly allocated worker rather than the manager as `userdata` argument?

src/udev/udevd.c
```diff
@@ -1544,59 +1554,55 @@ static int on_sighup(sd_event_source *s, const struct signalfd_siginfo *si, void
         return 1;
 }
 
-static int on_sigchld(sd_event_source *s, const struct signalfd_siginfo *si, void *userdata) {
-        Manager *manager = userdata;
+static int on_sigchld(sd_event_source *s, const siginfo_t *si, void *userdata) {
+        Manager *manager = ASSERT_PTR(userdata);
```
poettering: as mentioned above: if `userdata` would be the worker object, rather than the manager object, you wouldn't have to look up the worker via the hashmap by the pid.

and the worker has a `manager` field anyway, where you can acquire the manager object trivial if you need it

poettering (changes requested):
So, judging by the three commits that deal with sd-event freeing/disconnection in child processes: there seems to be something wrong there. we really should not try to free event sources or event loops in children which we allocated in the main process. we should just "leak" that memory.

yes, that#s messy, but "fork()" generally is messy. Ideally we'd insert an execve() btw, to makae sure the workers get a clean context.

yuwata:
> So, judging by the three commits that deal with sd-event freeing/disconnection in child processes: there seems to be something wrong there. we really should not try to free event sources or event loops in children which we allocated in the main process. we should just "leak" that memory.
> 
> yes, that#s messy, but "fork()" generally is messy. Ideally we'd insert an execve() btw, to makae sure the workers get a clean context.

One exception is udevd. It forks many workers, and all workers unref() unnecessary sd_event and sd_event_source.

yuwata:
@poettering Thank you for the review. Only comments for udevd are addressed.

src/libsystemd/sd-event/sd-event.c
keszybz: source_free → source_disconnect → source_child_pidfd_unregister.
So this should be fine.

keszybz:
I think this makes a lot of sense. The public entry points require cannot be used after fork, and that is reasonable. But you still want to do cleanup in a way that frees memory and other resources specific to the process cleanly. This is very important for debugging: otherwise you get some issues reported by valgrind and it's hard to figure out if they are legitimate. The check with `event_pid_changed(e)` is a reasonable way to do this.

poettering:
> I think this makes a lot of sense. The public entry points require cannot be used after fork, and that is reasonable. But you still want to do cleanup in a way that frees memory and other resources specific to the process cleanly. This is very important for debugging: otherwise you get some issues reported by valgrind and it's hard to figure out if they are legitimate. The check with `event_pid_changed(e)` is a reasonable way to do this.

I disagree. You cannot reasonably clean up memory in forked children that was allocated in the parent. Given threads and stuff. We shouldn't even try I am sure.

yuwata:
Still I do not understand the point why we should not free them. But, sd_event_unref() or sd_event_source_unref() are public, and kind of free functions. Hence, I think we should not trigger any assertion on that, and do their task gracefully.
Even if udev workers will changed to not free the sd_event or sd_event_source allocated by the main process, but other applications may already do the similar thing.

Or, are there any issues triggered when a child free a memory which is allocated by the parent?

keszybz:
It is probably true that this cannot be solved in the general case. But we're not making any big promises here: the code just does cleanup that works for us. This is important to make debugging easier. Or in other words, those changes should make things strictly better, even if they don't solve all corner cases.

PR #23003, opened 2022-04-06
sd-ipv4acd: actually drop the arp packet from one of the host interface

### LINKED ISSUE #23001: ACD marks packet as a conflict when hw address belongs to host
Sorry to forego the usual bug report form, but I have only seen this problem in code.
I believe [this line in sd-ipv4acd.c](https://github.com/systemd/systemd/blob/main/src/libsystemd-network/sd-ipv4acd.c#L332) to be erroneous. 
The if statement above seems to check whether the hardware address belongs to the host and if it does, it returns true indicating a conflict, but it should return false as this is not a conflict. The comment above indicates that it is checking if the address belongs to the host and if it does, it should be ignored.

Digging deeper, this check is carried out by `check_mac_callback`, which is a function dynamically set by `sd_ipv4acd_set_check_mac_callback`, which appears to be called [in networkd-ipv4acd.c](https://github.com/systemd/systemd/blob/187df3dd91105385c27b4eb0eb5ab05981f6732c/src/network/networkd-ipv4acd.c#L182), setting it to `ipv4acd_check_mac`.
`ipv4acd_check_mac` is defined [also in networkd-ipv4acd.c](https://github.com/systemd/systemd/blob/187df3dd91105385c27b4eb0eb5ab05981f6732c/src/network/networkd-ipv4acd.c#L115) whose primary functionality seems to rely on a call to `link_get_by_hw_addr` which is defined in [networkd-link.c](https://github.com/systemd/systemd/blob/b2e7ded16b16181767822d471d214cd8eba3c2d5/src/network/networkd-link.c#L278).
Presumably the logic here is that there is a hashmap (`m->links_by_hw_addr`) that contains the host's hardware addresses and if an address can be found in this hashmap, it belongs to the host, otherwise it does not. So if `link_get_by_hw_addr` returns 0, then the address belongs to the host. `ipv4acd_check_mac` checks that the return value is a nonnegative number and returns the boolean from that check. So when `ipv4acd_arp_conflict` calls `check_mac_callback`, it will check that it is larger than 0 (Which is actually redundant), because we want to check that it is an address owned by the host.

A packet where the sender hardware address belongs to the host, should not be seen as a conflict (Per RFC 5227 section 2.1.1., most of the relevant text is in the comment), and thus if my reasoning is correct, the function should return false. 
This line of reasoning does not consider the behaviour of BPF, I'm not familiar with what packets are filtered.

I have not seen this be a problem in any actual usage scenario, but I ended up seeing this part of the code by coincidence and it worried me a bit.

yuwata:
Fixes a bug in 7f77917c0effe92d5fed52503bceddabcb4667ba.

Fixes #23001.

### DIFF
--- a/src/libsystemd-network/sd-ipv4acd.c
+++ b/src/libsystemd-network/sd-ipv4acd.c
@@ -329,7 +329,7 @@ static bool ipv4acd_arp_conflict(sd_ipv4acd *acd, const struct ether_arp *arp, b
         if (acd->check_mac_callback &&
             acd->check_mac_callback(acd, (const struct ether_addr*) arp->arp_sha, acd->check_mac_userdata) > 0)
                 /* sender hardware is one of the host's interfaces, ignoring. */
-                return true;
+                return false;
 
         return true; /* conflict! */
 }


yuwata:
Please backport this to v249 and v250-stable.

bluca (approved):

bluca:
Looks unrelated but also seems new, FYI @mrc0mmand 

```
TEST-01-BASIC: ● systemd-journal-flush.service loaded failed failed Flush Journal to Persistent Storage
```

PR #23009, opened 2022-04-07
Fix detection of libsystemd-shared with suffix

keszybz:


src/nspawn/nspawn-util.c
```diff
@@ -0,0 +1,78 @@
...
+
+                _cleanup_strv_free_ char **names = NULL;
+                _cleanup_free_ char *path = NULL;
+                char *c;
+
+                path = path_join(root, pattern);
+                if (!path)
+                        return -ENOMEM;
+
+                r = glob_extend(&names, path, 0);
```
poettering, line 37: i wished we could do that in a chroot() aware way, i.e. use chase_symlinks() somehow for this

src/nspawn/test-nspawn-util.c
```diff
@@ -0,0 +1,19 @@
...
+
+#include "nspawn-util.h"
+#include "string-util.h"
+#include "tests.h"
+
+TEST(systemd_installation_has_version) {
+        const unsigned versions[] = {0, 231, PROJECT_VERSION, 999};
+        int r;
+
+        for (unsigned i = 0; i < ELEMENTSOF(versions); i++) {
```
poettering: ELEMENTSOF() returns size_t, hence iterator i should be size_t too

src/nspawn/test-nspawn-util.c
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "nspawn-util.h"
+#include "string-util.h"
+#include "tests.h"
+
+TEST(systemd_installation_has_version) {
+        const unsigned versions[] = {0, 231, PROJECT_VERSION, 999};
```
poettering: also static please

src/nspawn/nspawn-util.c
```diff
@@ -58,11 +59,17 @@ int systemd_installation_has_version(const char *root, unsigned minimal_version)
                         if (!t2)
                                 continue;
 
-                        t2[0] = '\0'; /* truncate the suffix */
+                        /* truncate the version suffix at the first dash, underscore, or dot. */
+                        t2 = t + strcspn(t, "-_.");
 
-                        r = safe_atou(t, &version);
+                        string = strndup(t, t2 - t);
```
poettering: hmm, why do you calculate `t2` as pointer first, just to then get back to a size right after, i.e. just change this line to:

```c
string = strndup(t, strcspn(t, "-_."));
```
and don#t bother with t2

bluca (approved):

src/nspawn/nspawn-util.c
```diff
@@ -49,6 +49,7 @@ int systemd_installation_has_version(const char *root, unsigned minimal_version)
                         /* This is most likely to run only once, hence let's not optimize anything. */
                         char *t, *t2;
                         unsigned version;
+                        _cleanup_free_ char *string = NULL;
```
poettering: quite frankly I think we should drop the numeric version comparison altogether. instead let's just use strverscmp_improved() for this stuff. We have it for stuff like that, and it should handle this in a much nicer way.

src/basic/path-util.c
```diff
@@ -373,52 +372,6 @@ char *path_simplify(char *path) {
         return path;
 }
 
-int path_simplify_and_warn(
-                char *path,
-                unsigned flag,
-                const char *unit,
-                const char *filename,
-                unsigned line,
-                const char *lvalue) {
```
poettering, line 385: not sure I follow, but why is src/basic/ not linked into libsystemd-shared, too?

src/nspawn/test-nspawn-util.c
keszybz: We know that it goes up to 3, so even a `char` would be OK ;) But I changed it to `size_t` now.

src/nspawn/nspawn-util.c
keszybz: strverscmp_improved() works great.

src/nspawn/test-nspawn-util.c
keszybz: ... and then to FOREACH_STRING() in a later version.

src/basic/path-util.c
keszybz: `libsystemd-basic.a` obviously _is_ linked into `libsystemd-shared.so`. In fact, we use `link_whole` in that particular case, so all of `libsystemd-basic.a` will be copied into `libsystemd-shared.so`. So for anything that links to `libsystemd-shared.so`, there should be no difference, because we end up with the object in the library anyway. But other things which use `libsystemd-basic.a` but don't link to `-shared`, could end up with a copy too. I *think* it should be removed when in some optimization pass, but, as shown by the grep that is included in the commit message, this doesn't (always?) happen.

In general, if we can move things from `basic/` to `shared/`, this is always strictly better. It reduces the dependency scope of `basic/`, so it makes partial compilations faster. In some cases it also reduces the size and dependencies of things that don't link to `libsystemd-shared.so`. (Not in this case though.)

PR #23016, opened 2022-04-08
udev-builtin-input_id: use heuristics to detect joysticks

whot:
Several keyboard devices are erroneously tagged with ID_INPUT_JOYSTICK
because of random buttons they set. For example, the LiteOn Lenovo
Calliope USB Keyboard sets BTN_TRIGGER, BTN_TOP2, BTN_PINKIE and
BTN_BASE, see libinput issue 745 for details.

ID_INPUT_JOYSTICK triggers the uaccess rules, making those keyboards
easily accessible. That's not a problem in the LiteOn example since that
event node doesn't contain the normal keys and eavesdropping on volume
keys is probably not very interesting.

Improve the joystick detection by adding heuristics similar to what
libinput 1.20 uses: check for some specific set of keys that are common
on keyboards but very unlikely on joysticks. If enough of those are
present (or the device has less than 2 axes or joysticks), don't tag it
as joystick.

libinput also checks for > 10 keyboard keys, but this is not done here
to be more conservative.

cc @JoseExposito

### DIFF
--- a/src/udev/udev-builtin-input_id.c
+++ b/src/udev/udev-builtin-input_id.c
@@ -156,7 +156,8 @@ static bool test_pointers(sd_device *dev,
         bool has_abs_coordinates = false;
         bool has_rel_coordinates = false;
         bool has_mt_coordinates = false;
-        bool has_joystick_axes_or_buttons = false;
+        size_t num_joystick_axes = 0;
+        size_t num_joystick_buttons = 0;
         bool has_pad_buttons = false;
         bool is_direct = false;
         bool has_touch = false;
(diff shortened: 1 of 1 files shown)


whot:
This was prompted by #22860, cc @smcv

src/udev/udev-builtin-input_id.c
```diff
@@ -235,9 +240,9 @@ static bool test_pointers(sd_device *dev,
                         is_abs_mouse = true;
                 else if (has_touch || is_direct)
                         is_touchscreen = true;
-                else if (has_joystick_axes_or_buttons)
+                else if (num_joystick_buttons || num_joystick_axes)
```
medhefgo: We don't do implicit bool conversion for integer types, please use `> 0`.

src/udev/udev-builtin-input_id.c
```diff
@@ -262,6 +267,30 @@ static bool test_pointers(sd_device *dev,
...
+                        KEY_LEFTCTRL, KEY_CAPSLOCK, KEY_NUMLOCK, KEY_INSERT,
+                        KEY_MUTE, KEY_CALC, KEY_FILE, KEY_MAIL, KEY_PLAYPAUSE,
+                        KEY_BRIGHTNESSDOWN,
+                };
+                size_t num_well_known_keys = 0;
+
+                for (size_t i = 0; has_keys && i < ELEMENTSOF(well_known_keyboard_keys); i++) {
+                        if (test_bit(well_known_keyboard_keys[i], bitmask_key))
+                                num_well_known_keys++;
+                }
```
yuwata: `has_keys` is not changed in the loop, hence it is not necessary to evaluated every time. Also, `{}` can be dropped:
```
        if (has_keys)
                for (size_t i = 0; i < ELEMENTSOF(well_known_keyboard_keys); i++)
                        if (test_bit(well_known_keyboard_keys[i], bitmask_key))
                                num_well_known_keys++;
```

yuwata (approved):
LGTM. One minor comment.

src/udev/udev-builtin-input_id.c
whot: done, thanks

src/udev/udev-builtin-input_id.c
whot: fixed, thanks

src/udev/udev-builtin-input_id.c
yuwata: Nah,,, not changed...

whot: sigh, talk about tunnel vision, I changed the one 2 lines south but not this one. Done now, this time for both :)

src/udev/udev-builtin-input_id.c
```diff
@@ -262,6 +267,30 @@ static bool test_pointers(sd_device *dev,
...
+                };
+                size_t num_well_known_keys = 0;
+
+                if (has_keys)
+                        for (size_t i = 0; i < ELEMENTSOF(well_known_keyboard_keys); i++)
+                                if (test_bit(well_known_keyboard_keys[i], bitmask_key))
+                                        num_well_known_keys++;
+
+                if (num_well_known_keys >= 4 || num_joystick_buttons + num_joystick_axes < 2)
+                        is_joystick = false;
```
poettering, line 294: maybe add a debug message here, i.e.

```c
log_device_debug(dev, "Input device has joystick buttons and axes, but also has multiple keyboard keys, assuming it is a keyboard and not a joystick.");
```

or so.

keszybz:
LGTM too.

src/udev/udev-builtin-input_id.c
whot: Done!

keszybz:
It's been building for 10+ h. Te failure in packit build is because of missing bpftool on i386, not related. I don't think this code would be tested by anything in the ubuntuautopkg tests, so I'll just merge this without further delay.

PR #23020, opened 2022-04-08
meson: explicitly include coverage tweaks when built w/ --coverage

mrc0mmand:
To make sure we don't miss any _exit() calls let's move the
coverage-related tweaks into a separate header file and include it
explicitly on the compiler command line using -include when a coverage
build is requested.

Follow-up to c6552ad381003a23cde7c3228e7071f30465df35.

---

As discussed in https://github.com/systemd/systemd/pull/23000, cc @keszybz.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -47,9 +47,6 @@ fuzzer_build = want_ossfuzz or want_libfuzzer
 # More items are added later after they have been detected.
 summary({'build mode' : get_option('mode')})
 
-# GCOV doesn't define any macro when compiled with, so let's define it ourselves
-conf.set10('BUILT_WITH_COVERAGE', get_option('b_coverage'))
-
 #####################################################################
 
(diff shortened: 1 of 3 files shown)


bluca (approved):
Looks reasonable to me, the extra compiler line argument is only added in coverage builds, so it shouldn't add to the noise in normal cases.

keszybz:
Thanks. Looks reasonable. I wonder if it'll change the coverage stats. After #23000 was merged, there was +1.4%.

mrc0mmand:
> Thanks. Looks reasonable. I wonder if it'll change the coverage stats. After #23000 was merged, there was +1.4%.

I'll give it a spin right after it's merged, since I'm curious as well.

yuwata (approved):

mrc0mmand:
Oh well, literally no change. But at least the coverage didn't decrease.

PR #23021, opened 2022-04-08
Tmpfiles fix precedence with plus sign

fbuihuu:


bluca:
```
TEST-22-TMPFILES:                   FAIL     (  5 s)
```

src/tmpfiles/tmpfiles.c
```diff
@@ -2896,6 +2896,26 @@ static int parse_age_by_from_arg(const char *age_by_str, Item *item) {
...
+        for (size_t n = 0; n < existing->n_items; n++) {
+                Item *e = existing->items + n;
+
+                if (item_compatible(e, i))
+                        continue;
+
+                /* Only multiple 'w+' lines for the same path are allowed. */
+                if (e->type != WRITE_FILE || !e->append_or_force ||
+                    i->type != WRITE_FILE || !i->append_or_force)
+                    return true;
```
yuwata: The condition must be inverted.
```
if (e->type == WRITE_FILE && e->append_or_force &&
    i->type == WRITE_FILE && i->append_or_force)
        continue;

return true;
```

BTW, why not to add the condition into `item_compatible()`?

fbuihuu: Because It looked more readable but it's subjective so I've no strong argument.

yuwata: Hm, wait, my comment was not correct. Sorry. It seems your code is fine.

But please fix the indentation.

bluca (approved):

yuwata (approved):

PR #23024, opened 2022-04-08
specifier: use %q for pretty hostname

poettering:
%R is already used in service manager specifier expansion (cgroup root),
hence use a different char, that was so far not used.

Follow-up for: 6ceb0a4094908dd213a78b9f6d0c59a684831ab0


bluca (approved):
Given there's loads of these, and they are not all defined together, we could really use some automated testing that checks for duplication (no idea how it would look like tho)

PR #23033, opened 2022-04-09
systemctl: Allow forcing a handover on switch-root

dasJ:
Ummm yes, this is probably a very weird use case but here I am…
Maybe we should name the variable something different but I don't know what we would call it so I'm open to suggestions.

Anyway, this is why I want this in systemd:
I'm currently porting the NixOS initrd from a veery long bash script to systemd.
The issue I'm having is that `systemctl switch-root` expects the `init=` parameter to be the same as `SYSTEMD_BINARY_PATH` to have a proper handover with state serialization.
In NixOS, we set this macro to `/run/current-system/systemd/lib/systemd` which is correct in the real system but which does not exist in the initrd at `/sysroot/run`.
So what I'm doing right now is creating `/run/current-system/…` in the initrd and let systemd move `/run` to `/sysroot/run` on switch.
This works well, but systemctl will notice that the inodes are not the same (in fact the inode lookup fails for `/sysroot/run/current-system/…` because the mount is not moved yet) so the handover is not done properly.
This Pull Request solves the issue by allowing us to bypass this check.

### DIFF
--- a/src/systemctl/systemctl-switch-root.c
+++ b/src/systemctl/systemctl-switch-root.c
@@ -2,6 +2,7 @@
 
 #include "bus-error.h"
 #include "bus-locator.h"
+#include "env-util.h"
 #include "parse-util.h"
 #include "path-util.h"
 #include "proc-cmdline.h"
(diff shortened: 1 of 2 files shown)


src/systemctl/systemctl-switch-root.c
```diff
@@ -37,7 +38,9 @@ int verb_switch_root(int argc, char *argv[], void *userdata) {
         }
 
         init = empty_to_null(init);
-        if (init) {
+        if (getenv_bool("SYSTEMCTL_FORCE_HANDOVER") > 0)
+            init = NULL;
```
poettering, line 42: indentation off.

But I really don't get this. You already can specify an empty string as third argument to achieve exactly this behaviour here. i.e. the `empty_to_null()` makes sure that an empty string is automatically turned into `NULL`.

i.e. if you invoke this:

```sh
systemctl switch-root /sysroot ""
```

you should already get exactly what you want? what am i missing?

dasJ: Thanks, that works like I need it :) systemd already doing what I need without me patching it again :P

PR #23035, opened 2022-04-09
compression: add build-time option to select default

bluca:
Compression and decompression are controlled by the same build flag,
so if one wants to use, say, LZ4 to compress, ZSTD has to be disabled,
which means one loses the ability to read zstd-compressed journals.

Add a default-compression meson option, that allows to select any of
the available compression algorithms as the default.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1452,6 +1452,30 @@ conf.set10('HAVE_ZSTD', have_zstd)
 
 conf.set10('HAVE_COMPRESSION', have_xz or have_lz4 or have_zstd)
 
+compression = get_option('default-compression')
+if compression == 'auto'
+        if have_zstd
+                compression = 'zstd'
+        elif have_lz4
+                compression = 'lz4'
+        elif have_xz
+                compression = 'xz'
+        else
+                compression = 'none'
+        endif
+elif compression == 'zstd' and not have_zstd
+        error('default-compression=zstd requires zstd')
+elif compression == 'lz4' and not have_lz4
+        error('default-compression=lz4 requires lz4')
+elif compression == 'xz' and not have_xz
+        error('default-compression=xz requires xz')
+endif
+conf.set('OBJECT_COMPRESSED_NONE', 0)
+conf.set('OBJECT_COMPRESSED_XZ', 1)
+conf.set('OBJECT_COMPRESSED_LZ4', 2)
+conf.set('OBJECT_COMPRESSED_ZSTD', 4)
+conf.set('DEFAULT_COMPRESSION', 'OBJECT_COMPRESSED_@0@'.format(compression.to_upper()))
+
 want_xkbcommon = get_option('xkbcommon')
 if want_xkbcommon != 'false' and not skip_deps
         libxkbcommon = dependency('xkbcommon',
--- a/src/libsystemd/sd-journal/compress.h
+++ b/src/libsystemd/sd-journal/compress.h
@@ -18,15 +18,15 @@ int compress_blob_zstd(const void *src, uint64_t src_size,
 static inline int compress_blob(const void *src, uint64_t src_size,
                                 void *dst, size_t dst_alloc_size, size_t *dst_size) {
         int r;
-#if HAVE_ZSTD
+#if DEFAULT_COMPRESSION == OBJECT_COMPRESSED_ZSTD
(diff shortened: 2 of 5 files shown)


meson.build
```diff
@@ -1452,6 +1452,35 @@ conf.set10('HAVE_ZSTD', have_zstd)
...
+        default_compression_lz4 = false
+        default_compression_xz = true
+else
+        default_compression_zstd = false
+        default_compression_lz4 = false
+        default_compression_xz = false
+endif
+conf.set10('DEFAULT_COMPRESSION_ZSTD', default_compression_zstd)
+conf.set10('DEFAULT_COMPRESSION_LZ4', default_compression_lz4)
+conf.set10('DEFAULT_COMPRESSION_XZ', default_compression_xz)
```
keszybz: This whole block is very unwieldy. Could we introduce a variable for this?
```python
compression = get_option('default-compression');
...  # handle "auto" here
conf.set('DEFAULT_COMPRESSION',
             compression == 'none' ? 0 : 'OBJECT_COMPRESSED_@0@'.format(compression.toupper()))
```
?

bluca: ok, done, PTAL

src/libsystemd/sd-journal/compress.h
```diff
@@ -5,6 +5,11 @@
 
 #include "journal-def.h"
 
+#define OBJECT_COMPRESSED_NONE 0
+#define OBJECT_COMPRESSED_ZSTD 1
+#define OBJECT_COMPRESSED_LZ4 2
+#define OBJECT_COMPRESSED_XZ 3
```
yuwata: Is it not necessary to be defined in config.h? Otherwise, config.h is not self-contained.

At least, please align them.

bluca: Builds were fine - but I've moved them now

meson.build
```diff
@@ -1452,6 +1452,30 @@ conf.set10('HAVE_ZSTD', have_zstd)
...
+        error('default-compression=zstd requires zstd')
+elif compression == 'lz4' and not have_lz4
+        error('default-compression=lz4 requires lz4')
+elif compression == 'xz' and not have_xz
+        error('default-compression=xz requires xz')
+endif
+conf.set('OBJECT_COMPRESSED_NONE', 0)
+conf.set('OBJECT_COMPRESSED_XZ', 1)
+conf.set('OBJECT_COMPRESSED_LZ4', 2)
+conf.set('OBJECT_COMPRESSED_ZSTD', 4)
```
yuwata, line 1476: Why 4, instead of 3?

yuwata: Ah, these are flag.

yuwata (approved):

src/libsystemd/sd-journal/journal-def.h
```diff
@@ -42,11 +42,9 @@ typedef enum ObjectType {
         _OBJECT_TYPE_MAX
 } ObjectType;
 
-/* Object flags */
+/* Object flags
+ * The per-compression enums are defined in meson.build so that config.h is self-contained */
 enum {
-        OBJECT_COMPRESSED_XZ   = 1 << 0,
-        OBJECT_COMPRESSED_LZ4  = 1 << 1,
-        OBJECT_COMPRESSED_ZSTD = 1 << 2,
```
poettering, line 49: oh, please let's not remove this here. The idea was that journal-def.h was a more or less complete description of the objects that make up a journal file. If you remove these enums here, then parts of the file format will be define in our build scripts. And that is kinda weird.

Everything in this file is *ABI* and the canonical reference definition of the data structures in journa files, hence never remove anything here really.

Hence, please find some other way to use these flags from meson.build. for example, name the flags differently there, and use an assert_cc() somewhere to ensure they match the values here.

poettering:
conceptually ok, but please, let's make sure journal-def.h remains a comprehensive description of the file format.

src/libsystemd/sd-journal/journal-def.h
bluca: There were no build issues with having config.h not self-contained, so I can simply move the definitions back to here - I have no strong opinion either way

poettering: please do

bluca: https://github.com/systemd/systemd/pull/23121

PR #23038, opened 2022-04-10
udev: log once

yuwata:


src/udev/net/link-config.c
```diff
@@ -686,7 +673,27 @@ static int link_apply_rtnl_settings(Link *link, sd_netlink **rtnl) {
...
+        static int cached = -1;
+        bool b;
+        int r;
+
+        if (cached >= 0)
+                return cached;
+
+        r = proc_cmdline_get_bool("net.ifnames", &b);
+        if (r < 0)
+                log_warning_errno(r, "Failed to parse net.ifnames kernel command line option, ignoring: %m");
```
poettering: maybe suffix `net.ifnames` → `net.ifnames=`, after all it expects an argument, and that's how we usually try to denote that fact in log messages.

yuwata:
@poettering Thank you for the review. The above comment is addressed. Upgrading the green label.

keszybz:
The failures seems unrelated, something about timeouts and internal server errors when retrieving packages.

PR #9551, opened 2018-07-09
udev: fix race condition in symlink generation code

mwilck:
This request replaces the previous pull requests https://github.com/systemd/systemd/pull/8667 and https://github.com/systemd/systemd/pull/8819. These requests belong together, as https://github.com/systemd/systemd/pull/8819 introduced currently failing tests that were fixed by https://github.com/systemd/systemd/pull/8667. Especially in https://github.com/systemd/systemd/pull/8667,  several updates from my side may confuse readers. Therefore I decided to create a new pull request.

With lots of contenders for a certain symlink, the current udev code suffers from two problems:

 1. the symlink creation is racy if udev events from several contenders are processed simultaneously. This is a real world problem that has been reported to SUSE by customers. The patch set extends the udev test suite by new test cases that reproduce this failure reliably, in particular on multicore systems (my 4-core laptop suffices).
 2. the algorithm for selecting the best symlink is highly inefficient for many contenders, because all potential symlink targets are represented by files flatly stored under `/run/udev/links/$LINK`. To obtain  the highest-prio link, udev must read all entries and retrieve their priority from the udev db, which is again inherently racy. This inefficiency is strongly aggravated if locking is used to solve the first problem.

Another issue which I faced when looking for a solution for this problem was the bad scaling of `fcntl()` based file locking in current kernels. Neil Brown has developed a fix for this, but it will take a lot of time until that's broadly available. Therefore I am using a different approach for locking based on a set of SysV semaphores, which doesn't have kernel-side scaling issues.

In the discussion for https://github.com/systemd/systemd/pull/8667, @poettering had asked for a solution without locking ("simple rescan"). After having conducted lots of experiment and research in alternative solutions I can assure that no such solution exists. Like in other, similar cases, locking the shared data structure is the only safe way to avoid the race.

Solving both 1.) and 2.) above requires both locking and changing the internal data structures udev uses for managing symlink targets. With the method presented here, the race is avoided, and yet the code scales better than the original in my tests.

The patch set also fixes some minor issues I encountered while working on the udev test suite.

PR #23043, opened 2022-04-11
udev-node: use flock() for symlink stack directory

yuwata:


src/udev/udev-node.c
```diff
@@ -246,129 +238,96 @@ static int update_timestamp(sd_device *dev, const char *path, struct stat *prev)
...
-                r = update_timestamp(dev, dirname, &st);
-                if (r < 0)
-                        return log_device_debug_errno(dev, r, "Failed to update timestamp of %s: %m", dirname);
+        r = mkdir_p(dirname, 0755);
+        if (r < 0)
+                return r;
 
-                return 0;
-        }
+        fd = open(dirname, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_NONBLOCK | O_RDONLY);
```
poettering: I'd drop `O_NONBLOCK` here.

`O_NONBLOCK` makes sense if you intend to open a regular file, but the path for some reason points to a device node of the kind where opening blocks (old serial port devices), which you don't want.

But here, you specify O_DIRECTORY anyway, hence the operation will fail anyway if we are not dealing with a directory. Hence O_NONBLOCK is unnecessary amd hence I'd not place it here.

src/udev/udev-node.c
```diff
@@ -246,129 +238,96 @@ static int update_timestamp(sd_device *dev, const char *path, struct stat *prev)
...
+        if (fd < 0)
+                return -errno;
 
-        return log_device_debug_errno(dev, SYNTHETIC_ERRNO(ELOOP), "Failed to create symbolic link %s: %m", filename);
+        return TAKE_FD(fd);
 }
```
poettering, line 309: this function reminds me a bit of `open_mkdir_at()`, which is not recursive though... but maybe we should extend that?

src/udev/udev-node.c
```diff
@@ -201,7 +203,8 @@ static int link_find_prioritized(sd_device *dev, bool add, const char *stackdir,
                         r = free_and_strdup(&target, devnode);
                         if (r < 0)
                                 return r;
-                }
+                } else
+                        continue;
```
poettering: maybe debug log here? failures here are unexpected, no? hence maybe log about the error?

src/udev/udev-node.c
```diff
@@ -329,6 +255,9 @@ static int stack_directory_open(const char *dirname) {
         if (fd < 0)
                 return -errno;
 
+        if (flock(fd, LOCK_EX) < 0)
```
poettering, line 305: the dir is world-accessible, no? that kinda rules out flock() for this I fear. Because you can read a directory, then you can also lock it. That seriously limits the usefulness of flock(), but it is how it is...

Otherwise any arbitrary process on the system can make udev hang/DoS...

poettering: but we can make the dir have access mode 0700, right? if so this would be OK to do I guess. or would we lose anything if we made that change? i.e. should have unpriv userspace have a good reason to read these dirs?

src/udev/udev-node.c
```diff
@@ -70,18 +72,18 @@ static int node_symlink(sd_device *dev, const char *devnode, const char *slink)
...
+        r = tempfn_xxxxxx(slink, NULL, &slink_tmp);
         if (r < 0)
-                return log_device_debug_errno(dev, r, "Failed to get device id: %m");
-
-        slink_tmp = strjoina(slink, ".tmp-", id);
-        (void) unlink(slink_tmp);
+                log_device_debug_errno(dev, r,
+                                       "Failed to generate temporary file name for symlink '%s', "
+                                       "creating symlink non-atomically, ignoring: %m",
+                                       slink);
```
poettering: why make this graceful? why not just fail this? is there any valid case where tempfn_xxxxx() can fail here?

src/udev/udev-node.c
```diff
@@ -246,129 +238,96 @@ static int update_timestamp(sd_device *dev, const char *path, struct stat *prev)
...
+                const char *devname;
+                int priority;
 
-                if (unlink(filename) < 0)
-                        unlink_error = -errno;
+                r = sd_device_get_devname(dev, &devname);
+                if (r < 0)
+                        return r;
 
-                if (rmdir(dirname) >= 0 || errno == ENOENT)
```
poettering, line 346: so, hmm, if this is not cleaned up anymore, then these might just accumulate without bounds, no?

i.e. i.e. consider `/dev/disk/by-uuid/` symlinks. each time a new device pops up, we'll create one stack dir for that, and if that never gets cleaned up, then eventually /run is filled up for good? what am i missing?

src/udev/udev-node.c
poettering: other option, create a lock file inside the dir, and jst make the lock file 0700. Slightly complicates removing the dir when it is not used anymore though

src/udev/udev-node.c
yuwata: Added a cleanup logic by the last commit.

src/udev/udev-node.c
yuwata: Maybe. But, I'd like to leave that for a later PR.

src/udev/udev-node.c
yuwata: Now the directory is created with 0700, as creating a file in the directory makes the cleanup logic complicated.

src/udev/udev-node.c
yuwata: If the filename is too long, it fails.

yuwata:
@poettering Thank you for the review. Most of comments are addressed. PTAL.

src/udev/udev-node.c
```diff
@@ -34,7 +34,6 @@
```
poettering, line 26: this commit msg is a bit misleading now, give you add a clean-up logic later?

src/udev/udev-node.c
```diff
@@ -310,25 +240,28 @@ static int stack_directory_update(sd_device *dev, int fd, bool add) {
...
+        r = mkdir_parents(dirname, 0755);
         if (r < 0)
                 return r;
 
-        fd = open(dirname, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
+        fd = open_mkdir_at(AT_FDCWD, dirname, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY, 0700);
         if (fd < 0)
+                return fd;
+
+        /* To prevent an arbitrary processe locks the directory. */
```
poettering: processe → process

but maybe:

```c
/* Prevent that unprivileged processes may lock the directory. */
```

src/udev/udev-node.c
```diff
@@ -310,25 +240,28 @@ static int stack_directory_update(sd_device *dev, int fd, bool add) {
...
+        if (fchmod(fd, 0700) < 0)
```
poettering: given you already specified 0700 during mkdir creation above this line is actually unnecessary...

poettering: (but the comment is good, and should stay)

src/udev/udev-node.c
```diff
@@ -26,6 +26,31 @@
...
+
+        FOREACH_DIRENT_ALL(de, dir, break) {
+                if (de->d_name[0] == '.')
+                        continue;
+
+                if (de->d_type != DT_DIR)
+                        continue;
+
+                if (unlinkat(dirfd(dir), de->d_name, AT_REMOVEDIR) < 0 && errno != ENOTEMPTY)
+                        log_debug_errno(errno, "Failed to remove directory '/run/udev/links/%s', ignoring: %m", de->d_name);
```
poettering: this is racy, no?

let's say some worker mkdir()s the dir and immediately open()s it as per your earlier commit using `open_mkdir_at()`. But given this is not atomic there's the chance that this cleanup logic here will rmdir() the dir in the middle…

I figure open_mkdir_at() should be called in a loop to handle ENOENT. Or maybe extend it to optionally call things in a loop a couple of times, in case the dir keeps vanishing if we try to open it.

poettering (changes requested):

src/udev/udev-node.c
yuwata: The directory may be created by an old udevd. In that case, we need to explicitly update the access mode, as open_mkdir_at() does not change the access mode for the existing directory.

poettering: true, but in that case you are actually vulnerable, because bad code might have locked the dir beforehand...

not sure it's worth really thinking about that too much, but to me it sounds pointless trying to address a case that is flawed anyway...

src/udev/udev-node.c
yuwata: This should not happen. When the function is called. No event is queued and no worker exists. Hence, there should be no one who locks and access the directory.
But for safety, we can use another `flock()` here.

yuwata:
@poettering Thank you for the review. The commit message and the comment in the code is updated. Also, the cleanup logic is updated. PTAL.

src/udev/udev-node.c
poettering: but the extra flock won't help, because you cannot create a directory flocked...

i.e. there's a window between creating the dir and your ability to open + flock it...

poettering: if this is safe, because this cannot happen because all workers are known dead at this point, then maybe add a comment about that. because it is not obvious.

src/udev/udev-node.c
```diff
@@ -26,6 +26,44 @@
...
+                if (de->d_type != DT_DIR)
+                        continue;
+
+                fd = openat(dirfd(dir), de->d_name, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
+                if (fd < 0) {
+                        log_debug_errno(errno, "Failed to open directory '/run/udev/links/%s', ignoring: %m", de->d_name);
+                        continue;
+                }
+
+                if (flock(fd, LOCK_EX | LOCK_NB) < 0) {
```
poettering: so yeah, as mentioned, this will not help, as the workers can't atomically create the dr and lock it. hence please drop this bit

but if this is called only when the workers are all gone, then this should indeed be safe, but please add a comment about that, as mentioned.

src/udev/udev-node.c
yuwata: Hmm, indeed... But, if the directory was not locked, then we should avoid any further lock by unprivileged processes, right?

poettering: well, i mean, it doesn't properly fix the problem, but it certainly makes it a bit better....

maybe add a comment `/* Try to lock down the access mode to cover for upgrades, even though this won't kick out clients who already locked the dir in case it was world-readable as it was historically */`

src/udev/udev-node.c
yuwata: OK. The another flock() was dropped.

src/basic/path-util.c
```diff
@@ -197,6 +197,24 @@ int path_make_relative(const char *from, const char *to, char **ret) {
...
+        /* Similar to path_make_relative(), but provides the relative path from the parent directory of
+         * 'from_child'. This may be useful when creating relative symlink. */
+
+        r = path_extract_directory(from_child, &from);
+        if (r < 0)
+                return r;
+
+        return path_make_relative(from, to, ret);
+}
+
```
keszybz, line 213: Maybe add a test in `test-path-util.c`?

src/udev/udev-node.c
keszybz: Shouldn't this be higher than `debug`?

src/udev/udev-node.c
```diff
@@ -310,25 +240,31 @@ static int stack_directory_update(sd_device *dev, int fd, bool add) {
...
+        r = mkdir_parents(dirname, 0755);
         if (r < 0)
                 return r;
 
-        fd = open(dirname, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
+        fd = open_mkdir_at(AT_FDCWD, dirname, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY, 0700);
         if (fd < 0)
+                return fd;
+
+        /* Prevent that unprivileged processes may lock the directory.
```
keszybz: "Prevent unprivileged processes from locking the directory."
"direcotry" → "directory"
"may be already" → "may have been already"
"does not change" → "will not change"

src/udev/udev-node.c
keszybz: Hmm, shouldn't we just use a stack directory with a different name to avoid the locking issues?

src/udev/udev-node.c
keszybz: I think this should be handled in `tempfn_xxxxxxx` instead. If we assume `tempfn_xxxxxxx` can fail, we'll have to have the same fallback in every caller, or at least in those where the filename is controlled externally or unpredictable.
Let's make `tempfn_xxxxxxx` smarter: only use up to `NAME_MAX - strlen(extra) - strlen(".#" "XXXXXX") ` of the original file name. This way the function can only fail on oom, and callers become simpler.

src/udev/udev-node.c
```diff
@@ -26,6 +26,34 @@
 
 #define UDEV_NODE_HASH_KEY SD_ID128_MAKE(b9,6a,f1,ce,40,31,44,1a,9e,19,ec,8b,ae,f3,e3,2f)
 
+int udev_node_cleanup(void) {
+        _cleanup_closedir_ DIR *dir = NULL;
+
+        /* This MUST be called when no worker exists. Otherwise, it causes an race between mkdir()
```
keszybz: I think the comment should be inverted here: "This must not be called when any workers exist. It would cause a race …"

(Right now the sentence says that it's important to call this… but this is not what the comment was trying to convey.)

keszybz (changes requested):

poettering:
> Hmm, shouldn't we just use a stack directory with a different name to avoid the locking issues.

I kinda like the idea, but how to handle upgrades then?

poettering:
> I kinda like the idea, but how to handle upgrades then?

Maybe we should just tell people to retrigger everything on update. Which they really should do anyway. If so, then the old dir just doesn't matter (though we wouldn't clean it up either, but I'd be OK with that)

poettering:
> think this should be handled in tempfn_xxxxxxx instead. If we assume tempfn_xxxxxxx can fail, we'll have to have the same fallback in every caller, or at least in those where the filename is controlled externally or unpredictable.
> Let's make tempfn_xxxxxxx smarter: only use up to NAME_MAX - strlen(extra) - strlen(".#" "XXXXXX") of the original file name. This way the function can only fail on oom, and callers become simpler.

So I thought about this approach, and I think we should implement this, but this doesn't solve the problem fully, because if you want to create a temporary path for something that has NAME_MAX length already, and the last compononent (i.e. the filename) is just one character, then you cannot possibly stick six XXXXXX characters in there anymore. So while it pushes the problem out, it doesn't solve it...

keszybz:
Can you give an example? I don't understand the comment.

poettering:
> Can you give an example? I don't understand the comment.

I presume the one about XXXXXX.

Consider you intend to create a file with path `/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/x`

(these are 4095 characters, i.e. the longest allowed path, given PATH_MAX is 4096)

Now you pass this path to `tempfn_xxxxxxx()`, to generate a temporary path from it. To do that it must include at least six `X` characters in the filename. That means the last component would grow from one char to six, and thus the whole path would become 4100 chars long. But that's larger than PATH_MAX, hence not allowed....

So whtever you do, tempfn_xxxxxx() might not be ablet to produce a valid temporary filename for you.... :-(

keszybz:
I was talking about NAME_MAX. PATH_MAX is a different issue, and we can't solve that. If we're asked to create a temporary file name, and only given on char of space to work with, this is not going to work anyway. But here we can expect that the total length will be OK. At the same time, it's quite possible that the name of the last component may be close to the NAME_MAX limit.

src/udev/udev-node.c
yuwata: > Shouldn't this be higher than `debug`?

Not sure, but the function is something like a library function now, and the similar errors are ignored. I'd like to keep it the debug level.

src/udev/udev-node.c
yuwata: > Hmm, shouldn't we just use a stack directory with a different name to avoid the locking issues?

Now, ".lock" file is created under the directory. Then, I think it is not necessary to rename the directory.

yuwata:
@poettering and @keszybz Thank you for many comments.
- tempfn_xxxxxx() or friends now truncate the original filename.
- adds several tests for tempfn_xxxxxx() and friends.
- adds several tests for path_make_relative_parent().
- stack_directory_find_prioritized_devnode() checks the file type before calling readlink().
- instead of the stack directory, now ".lock" file is created in the directory, and lock it. Then, we can avoid an unprivileged user locks the directory, as the lock file is created after the directory is chmod()ed to 0700.

PTAL.

yuwata:
@msekletar If you'd like, could you take a look? I hope this is better than the previous logic of 'try-and-wait' loop, and possibly improves performance, though I have not run benchmark yet.

fbuihuu:
Glad to see that the userspace lock free algo thingy is going to be abandoned eventually. I'm not sure to understand why the flock approach is now acceptable by upstream though.

Anyway, @yuwata is there anything wrong with `make_lock_file_for()` for creating the lock ?

I'm asking because SUSE's implementation of  (f)locking `/run/udev/links` is based on this function and I would like to understand whether it might cause any issue.

src/udev/udev-node.c
fbuihuu: > the dir is world-accessible, no? that kinda rules out flock() for this I fear. Because you can read a directory, then you can also lock it. That seriously limits the usefulness of flock(), but it is how it is...
> 
> Otherwise any arbitrary process on the system can make udev hang/DoS...

Sorry but I don't get this. The lock file is created with 0600 perms.

How can an unpriv process could also lock it ?

yuwata:
The lock file is not exposed to unprivileged users, hence we can safely use the _UNPOSIX_ lock used by `make_lock_file_for()`. But, I do not like the implementation that it unlink() the lock file on unlock.
Also, the `make_lock_file_for()` is implemented as it works on NFS. But, that's not necessary, as /run/udev/links is on tmpfs.

yuwata:
> Glad to see that the userspace lock free algo thingy is going to be abandoned eventually. I'm not sure to understand why the flock approach is now acceptable by upstream though.

I am not sure the current implementation that SUSE uses, but this takes a lock per symlink (that is, /run/udev/links/xxxxx), instead of the whole links directory (/run/udev/links) which is used by #9551.
Hence, workers blocked only when they try to create/update/remove the same symlink.

fbuihuu:
> I am not sure the current implementation that SUSE uses, but this takes a lock per symlink (that is, /run/udev/links/xxxxx), instead of the whole links directory (/run/udev/links) which is used by #9551. Hence, workers blocked only when they try to create/update/remove the same symlink.

Well I don't think that was the concern for using a proper lock API at that time.

fbuihuu:
> The lock file is not exposed to unprivileged users, hence we can safely use the _UNPOSIX_ lock used by `make_lock_file_for()`. 

Could you explain why it would be the problem if the directory containing the lock was readable ?

Lennart already said "he dir is world-accessible, no? that kinda rules out flock() for this I fear. Because you can read a directory, then you can also lock it." but I don't see how an unprivileged process can lock udev's lock since the lock file is created with `0600` permissions.

> But, I do not like the implementation that it unlink() the lock file on unlock. 

Why would this be an issue ?

> Also, the `make_lock_file_for()` is implemented as it works on NFS. But, that's not necessary, as /run/udev/links is on tmpfs.

The function uses UNPOSIX locks by default and falls back to BSD locks if the former are not supported by the kernel. The fact that UNPOSIX locks support NFS is just a consequence of the use of UNPOSIX lock but callers of this function shouldn't really care.

msekletar (commented):
LGTM

msekletar:
@yuwata thank you all for working on this. Getting rid of the current "lock-free" algorithm is great!

yuwata:
@fbuihuu Sorry for late response. I completely overlooked your comment.

> > The lock file is not exposed to unprivileged users, hence we can safely use the _UNPOSIX_ lock used by `make_lock_file_for()`.
> 
> Could you explain why it would be the problem if the directory containing the lock was readable ?
> 
> Lennart already said "he dir is world-accessible, no? that kinda rules out flock() for this I fear. Because you can read a directory, then you can also lock it." but I don't see how an unprivileged process can lock udev's lock since the lock file is created with `0600` permissions.

If the lock file is readable by an unprivileged user, then the user can lock the file, and udev workers are blocked to access the dir.
If the directory is not workd-accessible, then there should be no problem.

> > But, I do not like the implementation that it unlink() the lock file on unlock.
> 
> Why would this be an issue ?

If the lock file is removed by a worker on unlock, then another worker needs to recreate it before lock the file. Hence, simple open() + flock() does not work...

> > Also, the `make_lock_file_for()` is implemented as it works on NFS. But, that's not necessary, as /run/udev/links is on tmpfs.
> 
> The function uses UNPOSIX locks by default and falls back to BSD locks if the former are not supported by the kernel. The fact that UNPOSIX locks support NFS is just a consequence of the use of UNPOSIX lock but callers of this function shouldn't really care.

Right. But the lock file is only used internally. We can use arbitrary lock method. There is no stronger opinion about that, but I prefer flock() as it looks simpler for me.

poettering:
Please split out the tempfn commit into a separate PR that we can review separately. It doesn't really have too much to do with the rest

src/udev/udev-node.c
```diff
@@ -157,15 +156,18 @@ static int link_find_prioritized(sd_device *dev, bool add, const char *stackdir,
                 if (streq(de->d_name, id))
                         continue;
 
-                path = path_join(stackdir, de->d_name);
-                if (!path)
-                        return -ENOMEM;
-
-                if (readlink_malloc(path, &buf) >= 0) {
+                if (de->d_type == DT_LNK) {
```
poettering, line 172: hmm, readdir() might actually return DT_UNKNOWN in d_type. At the very least, use dirent_ensure_type() to make sure we can rely on it.

src/udev/udev-node.c
```diff
@@ -27,6 +27,45 @@
...
+                return log_debug_errno(errno, "Failed to open directory '/run/udev/links', ignoring: %m");
+        }
+
+        FOREACH_DIRENT_ALL(de, dir, break) {
+                _cleanup_free_ char *lockfile = NULL;
+
+                if (de->d_name[0] == '.')
+                        continue;
+
+                if (de->d_type != DT_DIR)
```
poettering, line 49: as above, maybe do dirent_ensure_type first

src/udev/udevd.c
```diff
@@ -1151,6 +1153,9 @@ static int on_uevent(sd_device_monitor *monitor, sd_device *dev, void *userdata)
 
         (void) event_queue_assume_block_device_unlocked(manager, dev);
 
+        /* To make the stack directory /run/udev/links cleaned up later. */
+        manager->udev_node_needs_cleanup = true;
```
poettering: why not just set this when actually creating/changing the dir?

src/udev/udev-node.c
```diff
@@ -27,6 +27,45 @@
...
+                        continue;
+
+                lockfile = path_join(de->d_name, ".lock");
+                if (!lockfile)
+                        return log_oom_debug();
+
+                if (unlinkat(dirfd(dir), lockfile, 0) < 0 && errno != ENOENT) {
+                        log_debug_errno(errno, "Failed to remove '/run/udev/links/%s', ignoring: %m", lockfile);
+                        continue;
```
poettering, line 61: maybe add a comment here why it is OK to remove the lock file here

poettering:
looks mostly ok, but some points above

src/udev/udev-node.c
```diff
@@ -204,24 +206,111 @@ static int link_find_prioritized(sd_device *dev, bool add, const char *stackdir,
...
+        _cleanup_close_ int fd = -1;
+        int r;
+
+        assert(dirname);
+
+        r = mkdir_parents(dirname, 0755);
+        if (r < 0)
+                return r;
+
+        fd = open_mkdir_at(AT_FDCWD, dirname, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY, 0700);
```
poettering: this can be relaxed now that you use the lock file, no? i.e. 0755 to make things easier to debug

src/udev/udev-node.c
poettering: ok, but then also remove the this fchmod line, as we won't need it anymore

src/udev/udev-node.c
yuwata: dirent_ensure_type() is called in FOREACH_DIRENT_ALL().

src/udev/udevd.c
yuwata: Creating/changing the directory is done by workers. But cleanup must be done by the manager on idle. Hence, the flag must be set by the main process.

yuwata:
@poettering Thank you for the review. All comments are addressed. PTAL.

@fbuihuu Sorry for late response again. 

> > If the lock file is readable by an unprivileged user, then the user can lock the file, and udev workers are blocked to access the dir. If the directory is not workd-accessible, then there should be no problem.
> 
> Again the lock file is created with permissions 600, hence let me ask again how an unprivileged process could open the lock file.

Hm, yeah, the file is created with 0600. There is no problem.

> > If the lock file is removed by a worker on unlock, then another worker needs to recreate it before lock the file. Hence, simple open() + flock() does not work...
> 
> `make_lock_file_for()` seems to be prepared to handle this case, no ?

Not sure, maybe. If so, we can rewrite with it. But let's do that later.

PR #23049, opened 2022-04-11
core: update to use new-style sd-bus macros

amarjargal:
Replace old SD_BUS_METHOD_WITH_NAMES and SD_BUS_SIGNAL_WITH_NAMES macros 
to the new SD_BUS_METHOD_WITH_ARGS and SD_BUS_SIGNAL_WITH_ARGS macros.


medhefgo (approved):
I would prefer to not have the extra text in the commit message, it's restating the commit title.

(A question to other maintainers: I wonder if these should be sorted by name while they're being touched anyways?)

yuwata:
> (A question to other maintainers: I wonder if these should be sorted by name while they're being touched anyways?)

Yeah!

@amarjargal If you'd like, please do that. Or, leaving it as is now, and do it in a later PR is also OK. Thank you!

amarjargal:
> I would prefer to not have the extra text in the commit message, it's restating the commit title.
> 
> (A question to other maintainers: I wonder if these should be sorted by name while they're being touched anyways?)

OK. I'll keep in mind the next time. In general, do you prefer commits without any messages at all or is it okay if the messages are meaningful or necessary?

medhefgo:
Commit messages are good. We love them! But they should be meaningful and provide extra value that doesn't otherwise fit into the code itself. It should explain *why* something was done, list alternative ideas that might have been discarded (and why) and other gotchas.
Just restating the same things that was already said in the title provides little value. As would an explanation that is just "obvious" (this of course lies in the eye of the beholder).

amarjargal:
> > (A question to other maintainers: I wonder if these should be sorted by name while they're being touched anyways?)
> 
> Yeah!
> 
> @amarjargal If you'd like, please do that. Or, leaving it as is now, and do it in a later PR is also OK. Thank you!

Yes, I can do that in another PR. But let me check if I got it right. Do you mean to sort them alphabetically? I think the current way is grouping them logically which might be the better way, but I don't know.

Issue #22970, opened 2022-04-05
Spurious fails in `test-sd-device/sd_device_new_from_syspath()`


mrc0mmand:
**systemd version the issue has been seen with**
 > latest main

**Used distribution**
 > Arch Linux/CentOS Stream 8

In a couple of recent jobs I noticed a fail in `test-sd-device`:

```
382/984 test-sd-device                                                     FAIL              1.43s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=43 SYSTEMD_LANGUAGE_FALLBACK_MAP=/build/src/locale/language-fallback-map PATH=/systemd-meson-build:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl /systemd-meson-build/test-sd-device
――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――
Listing only the last 100 lines from a long log.
syspath:/sys/devices/software subsystem:event_source id:+event_source:software initialized:yes
syspath:/sys/devices/system/clockevents/broadcast subsystem:clockevents id:+clockevents:broadcast initialized:yes
syspath:/sys/devices/system/clockevents/clockevent0 subsystem:clockevents id:+clockevents:clockevent0 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent1 subsystem:clockevents id:+clockevents:clockevent1 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent2 subsystem:clockevents id:+clockevents:clockevent2 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent3 subsystem:clockevents id:+clockevents:clockevent3 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent4 subsystem:clockevents id:+clockevents:clockevent4 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent5 subsystem:clockevents id:+clockevents:clockevent5 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent6 subsystem:clockevents id:+clockevents:clockevent6 initialized:yes
syspath:/sys/devices/system/clockevents/clockevent7 subsystem:clockevents id:+clockevents:clockevent7 initialized:yes
syspath:/sys/devices/system/clocksource/clocksource0 subsystem:clocksource id:+clocksource:clocksource0 initialized:yes
[...]

PR #23051, opened 2022-04-11
udev: three minor tweaks

poettering:
Split out of #22969, since mostly independent from that.


yuwata (approved):

poettering:
hmm, the centos CI failure actually does look suspicious...

mrc0mmand:
> hmm, the centos CI failure actually does look suspicious...

Yup, that's already filed as https://github.com/systemd/systemd/issues/22970 (and it's not caused by this PR).

PR #23057, opened 2022-04-12
Various doc tweaks

keszybz:


man/sd-device.xml
```diff
@@ -0,0 +1,61 @@
...
+    for more information about the functions available.</para>
+  </refsect1>
+
+  <xi:include href="libsystemd-pkgconfig.xml" />
+
+  <refsect1>
+    <title>See Also</title>
+    <para>
+      <citerefentry><refentrytitle>systemd</refentrytitle><manvolnum>1</manvolnum></citerefentry>,
+      <citerefentry><refentrytitle>sd-event</refentrytitle><manvolnum>3</manvolnum></citerefentry>,
```
yuwata, line 57: Why sd-event here??

man/sd-device.xml
```diff
@@ -42,7 +42,8 @@
     <filename>libudev.h</filename>.</para>
 
     <para>See
-    <literallayout><citerefentry><refentrytitle>sd_device_get_syspath</refentrytitle><manvolnum>3</manvolnum></citerefentry>
+    <literallayout><citerefentry><refentrytitle>sd_device_get_syspath</refentrytitle><manvolnum>3</manvolnum></citerefentry>,
+<citerefentry><refentrytitle>sd_device_ref</refentrytitle><manvolnum>3</manvolnum></citerefentry>
```
yuwata, line 46: Indentation?

man/sd_device_ref.xml
```diff
@@ -0,0 +1,83 @@
...
+        <paramdef>sd_device *<parameter>device</parameter></paramdef>
+      </funcprototype>
+
+      <funcprototype>
+        <funcdef>void <function>sd_device_unrefp</function></funcdef>
+        <paramdef>sd_device **<parameter>device</parameter></paramdef>
+      </funcprototype>
+    </funcsynopsis>
+
+    <para><function>sd_device_ref()</function> increases the reference counter of
```
yuwata: Maybe, "internal reference counter"?

README
```diff
@@ -241,7 +241,7 @@ REQUIREMENTS:
           DESTDIR=... meson install -C build/
 
         A tarball can be created with:
-          git archive --format=tar --prefix=systemd-222/ v222 | xz > systemd-222.tar.xz
+          v=222 && git archive --prefix=systemd-$v/ v$v | zstd >systemd-$v.tar.zstd
```
yuwata: 222 -> 250?? This is not necessary to follow the recent release, but 222 is too old...

man/sd-device.xml
keszybz: It's another library from the same set. People generally might want to also use it.

man/sd-device.xml
keszybz: It needs to be like this. See `sd-bus.xml`.

keszybz:
I added "internal" in various places (new commit at the end) and changed the version as requested.

man/sd-device.xml
yuwata: Hmm.. interesting...

yuwata (approved):

README
```diff
@@ -225,7 +225,7 @@ REQUIREMENTS:
         polkit (optional)
 
         To build in directory build/:
-          meson setup build/ && meson compile -C build/
+          meson setup build/ && ninja -C build/
```
poettering, line 228: interesting, i though meson prefered the syntax via meson, because they wanted to permit some level of abstraction of used backend?

poettering:
lgtm

README
eli-schwartz: That's a matter of some opinion.

Many ninja targets are simply not allowed to be run via `meson compile` because they don't compile anything. Not all versions of Meson actually support meson compile (some versions simply try to configure a new build directory at ./compile/ instead) and in the best case scenario, it will first check whether you are using ninja or Microsoft Visual Studio, and effectively run `ninja "$@"` or `msbuild "$@"`.

PR #23061, opened 2022-04-12
some chase_symlinks() tweaks

poettering:
Nothing particularly relevant.


yuwata (approved):
LGTM.

src/basic/chase-symlinks.c
```diff
@@ -161,17 +165,17 @@ int chase_symlinks(
                 path_simplify(root);
 
                 if (flags & CHASE_PREFIX_ROOT) {
```
medhefgo, line 167: Might as well use `FLAGS_SET` here too.

PR #23065, opened 2022-04-12
pid1: pass additional env vars about exec context to generators

poettering:


src/core/manager.c
```diff
@@ -3665,8 +3665,70 @@ static int manager_run_environment_generators(Manager *m) {
...
+        if (r < 0)
+                return r;
+
+        *ret = TAKE_PTR(nl);
+        return 0;
+}
+
 static int manager_run_generators(Manager *m) {
         _cleanup_strv_free_ char **paths = NULL;
+        _cleanup_free_ char **ge = NULL;
```
yuwata: `_cleanup_strv_free_`

yuwata (approved):
LGTM. But one memleak.

src/core/manager.c
```diff
@@ -3665,8 +3665,70 @@ static int manager_run_environment_generators(Manager *m) {
...
+        nl = strv_copy(m->transient_environment);
+        if (!nl)
+                return -ENOMEM;
+
+        r = strv_env_assign(&nl, "SYSTEMD_SCOPE", MANAGER_IS_SYSTEM(m) ? "system" : "user");
+        if (r < 0)
+                return r;
+
+        if (MANAGER_IS_SYSTEM(m)) {
+                /* Note that $SYSTEMD_INITRD may be used to override the initrd detection in much of our
```
keszybz: The variable is actually called `SYSTEMD_IN_INITRD`…

src/core/manager.c
```diff
@@ -3665,8 +3665,70 @@ static int manager_run_environment_generators(Manager *m) {
...
+                 * codebase. This is hence more than purely informational. It will shorcut detection of the initrd
+                 * state if generators invoke our own tools. But that's OK, as it would come to the same results
+                 * (hopefully) */
```
keszybz: Dot.

src/core/manager.c
```diff
@@ -3665,8 +3665,70 @@ static int manager_run_environment_generators(Manager *m) {
...
+                        return r;
+
+                r = strv_env_assign(&nl, "SYSTEMD_VIRTUALIZATION_IS_VM", one_zero(VIRTUALIZATION_IS_VM(v)));
+                if (r < 0)
+                        return r;
+
+                r = strv_env_assign(&nl, "SYSTEMD_VIRTUALIZATION_IS_CONTAINER", one_zero(VIRTUALIZATION_IS_CONTAINER(v)));
+                if (r < 0)
+                        return r;
+        }
```
keszybz: Hmm, let's maybe define this a bit differently.
SYSTEMD_VIRTUALIZATION="type[:detail]", where "type" would be one of "", "chroot", "container", "vm", and the second part would only be set if known, for "vm" and "container" types. This way this becomes generic and the same variable can be used for other purposes later without too much verbosity.

man/systemd.generator.xml
```diff
@@ -128,6 +128,71 @@
...
+  <refsect1>
+    <title>Environment</title>
+
+    <para>The service manager sets a number of environment variables when invoking generator
+    executables. They carry information about the execution context of the generator, in order to simplify
+    conditionalizing generators to specific environments. The following environment variables are set:</para>
+
+    <variablelist class='environment-variables'>
+      <varlistentry>
+        <term><varname>$SYSTEM_SCOPE</varname></term>
```
keszybz: 'D' missing.

man/systemd.generator.xml
```diff
@@ -128,6 +128,71 @@
...
+        <literal>1</literal>. If it is run from the regular host (i.e. after the transition from initrd to
+        host) it is set to <literal>0</literal>. This environment variable is only set for system generators
+        and remains unset in user generators.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>$SYSTEMD_FIRST_BOOT</varname></term>
+
+        <listitem><para>If this boot-up cycle is considered a "first boot", this is set to
+        <literal>1</literal>; if it is a subsequent, regular boot is is set to <literal>0</literal>. For
```
keszybz: "is is"

man/systemd.generator.xml
```diff
@@ -128,6 +128,71 @@
...
+        details see the documentation of <varname>ConditionFirstBoot=</varname> in
+        <citerefentry><refentrytitle>systemd.unit</refentrytitle><manvolnum>5</manvolnum></citerefentry>. This
+        environment variable is only set for system generators and remains unset in user
+        generators.</para></listitem>
```
keszybz: Strike " and remains unset in user generators", here and in other places.

keszybz (changes requested):
Looks reasonable.

src/core/manager.c
keszybz: In particular, if we want to allow overriding autodetection with a variable in the future, it's much easier to have just one variable, so that there can be no conflicting information.

poettering: hmm, my assumption was that it is much easier from a shell script to parse $SYSTEMD_VIRTUALIZATION_IS_CONTAINER, and that's what 99% of the generators probably care about.

i for one wouldn't know how to test the var quickly if for "is this a container?" if i get "container:systemd-nspawn"?

src/core/manager.c
poettering: ouch

src/core/manager.c
keszybz: In C: `startswith_ptr(getenv("SYSTEMD_VIRTUALIZATION"), "container")`, in bash: `[[ $x = *container* ]]`, in Python `os.getenv('SYSTEMD_VIRTUALIZATION', '').startswith('container')`, etc. I don't think it's any harder to check.

poettering: and in bourne shell?

poettering: I must say, I disagree with this request. If I was to write a generator script in shell I'd probably just shell out `systemd-detect-virt -c` instead of trying to parse the `container:` thing out of the env var... But anyway, I reworked this as requested now, given I am not the audience for this feature anyway...

dtardon: > and in bourne shell?

```sh
case "$SYSTEMD_VIRTUALIZATION" in
container*) ... ;;
esac
```
or
```sh
if [ "${SYSTEMD_VIRTUALIZATION#container}" != "$SYSTEMD_VIRTUALIZATION"]; then
...
fi
```

yuwata (approved):
LGTM.

src/core/manager.c
keszybz: If it turns out to be a problem, you have the right to say blame it on me ;)

PR #23068, opened 2022-04-13
update hostnamed to get BIOS version

cerebro1:
Here is my PR for outreachy contribution : 
- defined function to get BIOS version from /sys/class/dmi/id/bios_version.
- defined method to call the above function after validation and autherization.
- Created a sd_bus_method_with_arg to get BIOS version using method.
Please review and let me know if i have understood the requirements correctly.

Do i need to update any man page corresponding to this ?


src/hostname/hostnamed.c
```diff
@@ -822,6 +822,30 @@ static int property_get_uname_field(
         return sd_bus_message_append(reply, "s", (char*) &u + PTR_TO_SIZE(userdata));
 }
 
+static int get_bios_version(char **ret) {
```
poettering: hmm, any chance you can generalize the existing get_hardware_serial() call a bit for this?

i.e. make it accept the sysattr name to read as additional param (or two, as get_hardware_serial() wants that). and then just provide two tiny wrappers around this generalization. one that passes the strings "product_serial"/"board_serial", and the other one: "bios_version".

get_dmi_data() does something similar, but for device properties instead of attributes. But it can be used as a template here.

(in case you wonder: udev device properties are uppercase fields assigned by udev in userspace to a device, and sysfs device attributes are lowercase fields exposed by the kernel directly in sysfs. The former might be based on the latter though. Confusin? yes! :-))

src/hostname/hostnamed.c
```diff
@@ -1126,9 +1150,47 @@ static int method_get_hardware_serial(sd_bus_message *m, void *userdata, sd_bus_
...
+        r = sd_bus_message_new_method_return(m, &reply);
+        if (r < 0)
+                return r;
+
+        r = sd_bus_message_append(reply, "s", biosversion);
+        if (r < 0)
+                return r;
+
+        return sd_bus_send(NULL, reply, NULL);
+}
```
poettering: and here, too, would prefer if we could share some code here with the hardware serial getter function. i.e. maybe use a function pointer parameter to choose whether to call the existing get_hardware_serial() or get_bios_version().

See how property_get_hardware_property() does this with property_get_hardware_vendor() and property_get_hardware_model(), for inspiration

src/hostname/hostnamed.c
```diff
@@ -1320,7 +1383,11 @@ static const sd_bus_vtable hostname_vtable[] = {
                                 SD_BUS_RESULT("s", json),
                                 method_describe,
                                 SD_BUS_VTABLE_UNPRIVILEGED),
-
+        SD_BUS_METHOD_WITH_ARGS("GetBiosVersion",
```
poettering: please replace the string "bios" with "firmware" everywhere please.

BIOS is the PC firmware from the 1990s. Today we use UEFI instead of BIOS. Many people use BIOS and firmware interchangable, but it is a bit incorrect to call a modern UEFI fimware "bios", hence let's avoid that here.

poettering (changes requested):
please also update "hostnamectl" to show the new field in its status output

src/hostname/hostnamed.c
nitinexclusively: I think we can remove polkit authentication as bios version can be read without root permission . We can get bios version without root permission using "/sys/class/dmi/id/bios_version" 
@poettering it will be helpful, if you can relay your comment on this .. Thank you

src/hostname/hostnamed.c
nitinexclusively: Just a suggestion - how about changing it to "uefi_firmware" , just to avoid any confusion with other firmware like ecfw (Embedded controller firmware) etc . Thank you

cerebro1:
Thanks @poettering for review. I have gone through your comments and been working on that. I should be able to bring a better version by tomorrow. 

@nitinexclusively you want the output to be displayed when we run hostnamectl command so that will change the order in which output is displayed right now. Any specific preference on which line version has to be printed or I could print it at the last.

nitinexclusively:
> 
> @nitinexclusively you want the output to be displayed when we run hostnamectl command so that will change the order in which output is displayed right now. Any specific preference on which line version has to be printed or I could print it at the last.

@cerebro1 I dont have any specific preference. I am beginner of  systemd , so , not sure, if maintainers of 'systemd' has any specific preference or not . Thank you !

cerebro1:
@poettering There is a comment in hostnamed.c
/* The product UUID and hardware serial is only available to privileged clients */
Will firmware_version be part of it ? I mean should i add it here ?

src/hostname/hostnamed.c
cerebro1: Yes. Updated the get_hardware_serial as get_hardware_firmware_version that passes strings "product_serial"/"board_serial", and the other one: "bios_version". and returns accordingly

cerebro1:
> please also update "hostnamectl" to show the new field in its status output

added

cerebro1:
> > Do i need to update any man page corresponding to this ?
> 
> At the very least you need to run "ninja -C build update-dbus-docs". Ideally you'd then fill in what that command filled in as blanks in the docs.
> 
> But admittedly we are kinda sloppy with documenting dbus APIs. Asking you to fill this in would require higher standards from you then we usually require from ourselves. Hence: would be excellent if you documented the new props in the the man pages, but it's not a strict requirement.

Sure will do, once my code works!

cerebro1:
@poettering Am i missing something in the codebase ?
When i run the command hostnamectl on my system and forcefully try to print whatever value is present in firmware_version, output is : 
`Firmware Version: n/a`
But the file /sys/class/dmi/id/bios_version is present on my machine and it has bios version. Am i executing it wrong?

cerebro1:
> please also update "hostnamectl" to show the new field in its status output

updated. Please review. Thanks!

bluca:
> @poettering There is a comment in hostnamed.c /* The product UUID and hardware serial is only available to privileged clients */ Will firmware_version be part of it ? I mean should i add it here ?

I don't think so, the version is the same across all devices, it cannot be used to identify a specific user

bluca:
```
2022-04-16T12:07:13.1195449Z ../src/hostname/hostnamed.c: In function ‘method_get_firmware_version’:
2022-04-16T12:07:13.1196258Z ../src/hostname/hostnamed.c:1141:18: error: unused variable ‘c’ [-Werror=unused-variable]
2022-04-16T12:07:13.1196790Z  1141 |         Context *c = userdata;
```

src/hostname/hostnamed.c
```diff
@@ -5,6 +5,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <string.h>
```
bluca: sort includes alphabetically please

src/hostname/hostnamed.c
```diff
@@ -218,8 +219,9 @@ static int get_hardware_serial(char **ret) {
         if (r < 0)
                 return log_debug_errno(r, "Failed to open /sys/class/dmi/id device, ignoring: %m");
 
-        (void) sd_device_get_sysattr_value(device, "product_serial", &s);
-        if (isempty(s))
+        if (dmi_data)
```
bluca: just `assert(dmi_data)` at the top of the function, after the variables declarations, no point in allowing this to be called without a data type

src/hostname/hostnamed.c
bluca: actually, you can remove this

src/hostname/hostnamed.c
```diff
@@ -218,8 +219,9 @@ static int get_hardware_serial(char **ret) {
...
+                (void) sd_device_get_sysattr_value(device, dmi_data, &s);
+        if (!s && strcmp(dmi_data,"product_serial"))
```
bluca: given this is now a generic helper, do the fallback in `get_hardware_serial()` below instead

src/hostname/hostnamed.c
```diff
@@ -1126,9 +1135,33 @@ static int method_get_hardware_serial(sd_bus_message *m, void *userdata, sd_bus_
         return sd_bus_send(NULL, reply, NULL);
 }
 
+static int method_get_firmware_version(sd_bus_message *m, void *userdata, sd_bus_error *error) {
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *reply = NULL;
+        _cleanup_free_ char *firmware_version = NULL;
+        Context *c = userdata;
```
bluca: this can be removed

src/hostname/hostnamed.c
bluca: this also returns the version of old-style bios, not just uefi, so uefi_firmware wouldn't be accurate

bluca:
> @poettering Am i missing something in the codebase ? When i run the command hostnamectl on my system and forcefully try to print whatever value is present in firmware_version, output is : `Firmware Version: n/a` But the file /sys/class/dmi/id/bios_version is present on my machine and it has bios version. Am i executing it wrong?

That is because GetFirmwareVersion is a D-Bus method, but it was integrated in hostnamectl with the D-Bus property getter, so it cannot work.
The UUID/serial are privileged information as they can identify a single machine, so they were implemented as methods so that PolicyKit can be used to assert privileges of the caller.

But as discussed the firmware version comes from the vendor, so there's no need for PolicyKit, hence there's no need to implement this as a method - just change it to be a property, and then it should work.

cerebro1:
> > @poettering Am i missing something in the codebase ? When i run the command hostnamectl on my system and forcefully try to print whatever value is present in firmware_version, output is : `Firmware Version: n/a` But the file /sys/class/dmi/id/bios_version is present on my machine and it has bios version. Am i executing it wrong?
> 
> That is because GetFirmwareVersion is a D-Bus method, but it was integrated in hostnamectl with the D-Bus property getter, so it cannot work. The UUID/serial are privileged information as they can identify a single machine, so they were implemented as methods so that PolicyKit can be used to assert privileges of the caller.
> 
> But as discussed the firmware version comes from the vendor, so there's no need for PolicyKit, hence there's no need to implement this as a method - just change it to be a property, and then it should work.

@bluca Thanks for review. If this is to be implemented as property, then we do not need a new function (get_hardware_firmware_data). get_dmi_data is already available for property. I will make use of it and update accordingly. Thanks!

cerebro1:
I was reverting my old commits and did a forced push, i think that closed my PR. Will update new PR link here just to not get this conversation on PR lost because this really guided me to understand how to work on this issue

bluca:
No worries, it happens, rules on GitHub are weird and sometimes it lets you reopen it, sometimes it doesn't

cerebro1:
New PR here : https://github.com/systemd/systemd/pull/23112

PR #23069, opened 2022-04-13
networkctl: obey --full with lldp command

### LINKED ISSUE #22806: systemd-networkd doesn't show full lldp information with networkctl lldp
I can reproduce the issue on the current debian stable and arch linux.

```
$ systemd --version
systemd 247 (247.3-6)
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +ZSTD +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=unified
```
```
$ systemctl --version                                                                                                                                                                                                                                                                                                         
systemd 250 (250.4-2-arch)
```

<!-- **NOTE:** Do not submit bug reports about anything but the two most recently released *major* systemd versions upstream! -->
<!--           If there have been multiple stable releases for that major version, please consider updating to a recent one before reporting an issue. -->
<!--           When using a distro package, please make sure that the version reported is meaningful for upstream. -->
<!-- See https://github.com/systemd/systemd-stable/tags for the list of most recent releases. -->
<!-- For older version please use distribution trackers (see https://systemd.io/CONTRIBUTING#filing-issues). -->

**Used distribution**
```
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
```
$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux-logo
```

```
$ uname -a
Linux tonno 5.10.0-7-amd64 #1 SMP Debian 5.10.40-1 (2021-05-28) x86_64 GNU/Linux
```
```
$ uname -a
Linux nyu 5.16.11-arch1-1 #1 SMP PREEMPT Thu, 24 Feb 2022 02:18:20 +0000 x86_64 GNU/Linux
```

```
$ lscpu | head -1
Architecture:                    x86_64
```

I would expect a full output of the lldp info like `GigabitEthernet1/0/10`.

A shortened version of the info that always cuts off after a fixed number of chars e.g. `GigabitEthernet1...`.

Example output:

```
$ networkctl lldp
LINK             CHASSIS ID        SYSTEM NAME      CAPS        PORT ID           PORT DESCRIPTION
eno3             ff:00:00:bb:aa:aa core.home.netwk  ..b.r...... GigabitEthernet1… GigabitEthernet…
eno4             ff:00:00:bb:aa:aa core.home.netwk  ..b.r...... GigabitEthernet2… GigabitEthernet…

Capability Flags:
b - Bridge; r - Router

2 neighbors listed.
```

Connect device A via ethernet to device B (other pc, switch, router, etc) that supports LLDP and set lldp information on device B that exceed 16 characters for port id and/or description. Then issue `networkctl lldp` command on device A to see shortened output.

<!-- Please paste relevant program terminal or journal output here, ideally
     when generated in debug mode (try setting the SYSTEMD_LOG_LEVEL=debug
     environment variable). For very long copy/pasted data consider using a
     service like https://gist.github.com/. Where copy/paste is not possible
     (for example early boot or late shutdown), a photo of the screen might do
     too, but text is always much preferred. -->
```
$ env SYSTEMD_LOG_LEVEL=debug networkctl lldp
Successfully forked off '(pager)' as PID 1506181.
Skipping PR_SET_MM, as we don't have privileges.
Pager executable is "pager", options "FRSXMK", quit_on_interrupt: yes
LINK             CHASSIS ID        SYSTEM NAME      CAPS        PORT ID           PORT DESCRIPTION
eno3             ff:00:00:bb:aa:aa core.home.netwk  ..b.r...... GigabitEthernet1… GigabitEthernet…
eno4             ff:00:00:bb:aa:aa core.home.netwk  ..b.r...... GigabitEthernet2… GigabitEthernet…

Capability Flags:
b - Bridge; r - Router

2 neighbors listed.
```

ackxolotl:
Do not ellipsize output when `-l` or `--full` is handed to `networkctl lldp`. Fixes #22806.

### DIFF
--- a/src/network/networkctl.c
+++ b/src/network/networkctl.c
@@ -2552,28 +2552,30 @@ static int link_lldp_status(int argc, char *argv[], void *userdata) {
                         (void) sd_lldp_neighbor_get_system_name(n, &system_name);
                         (void) sd_lldp_neighbor_get_port_description(n, &port_description);
 
-                        if (chassis_id) {
-                                cid = ellipsize(chassis_id, 17, 100);
-                                if (cid)
-                                        chassis_id = cid;
-                        }
+                        if (!arg_full) {
+                                if (chassis_id) {
+                                        cid = ellipsize(chassis_id, 17, 100);
+                                        if (cid)
+                                                chassis_id = cid;
+                                }
 
-                        if (port_id) {
-                                pid = ellipsize(port_id, 17, 100);
-                                if (pid)
-                                        port_id = pid;
-                        }
+                                if (port_id) {
+                                        pid = ellipsize(port_id, 17, 100);
+                                        if (pid)
+                                                port_id = pid;
+                                }
 
-                        if (system_name) {
-                                sname = ellipsize(system_name, 16, 100);
-                                if (sname)
-                                        system_name = sname;
-                        }
+                                if (system_name) {
+                                        sname = ellipsize(system_name, 16, 100);
+                                        if (sname)
+                                                system_name = sname;
+                                }
 
-                        if (port_description) {
-                                pdesc = ellipsize(port_description, 16, 100);
-                                if (pdesc)
-                                        port_description = pdesc;
+                                if (port_description) {
+                                        pdesc = ellipsize(port_description, 16, 100);
+                                        if (pdesc)
+                                                port_description = pdesc;
+                                }
                         }
 
                         if (sd_lldp_neighbor_get_enabled_capabilities(n, &cc) >= 0) {


src/network/networkctl.c
```diff
@@ -2552,28 +2552,30 @@ static int link_lldp_status(int argc, char *argv[], void *userdata) {
...
                         (void) sd_lldp_neighbor_get_port_description(n, &port_description);
 
-                        if (chassis_id) {
-                                cid = ellipsize(chassis_id, 17, 100);
-                                if (cid)
-                                        chassis_id = cid;
-                        }
+                        if (!arg_full) {
+                                if (chassis_id) {
+                                        cid = ellipsize(chassis_id, 17, 100);
```
yuwata, line 2557: In the above, we have
```
        if (arg_full)
                table_set_width(table, 0);
```
and table_add_many() ellipsizes strings gracefully.

Hence, please unconditionally put the given strings, e.g. chassis_id, to the table.

yuwata (changes requested):

src/network/networkctl.c
ackxolotl: Oh, so you mean all this ellipsizing code from line 2555 to 2578 is superfluous since `table_add_many()` ellipsizes strings anyway? I just tried it by commenting those lines out but nothing is ellipsized in my table.

yuwata: See,
https://github.com/systemd/systemd/blob/c87c30780624df257ed96909a2286b2b933f8c44/src/shared/format-table.c#L136-L137
And, by default, the width is SIZE_MAX. Hence, each strings are truncated if you run the command on a terminal with short width.

ackxolotl: Ok, but as I said: Nothing is truncated on my shell when I set it to 60 characters and remove those calls to `ellipsize()`, i.e., `ellipsize(chassis_id, 17, 100)` and so on. So either your claim is wrong or something else is going on.

yuwata: Please provide the diff you tested.

ackxolotl: This:

```
diff --git a/src/network/networkctl.c b/src/network/networkctl.c
index d656027ec4..d2a72dec27 100644
--- a/src/network/networkctl.c
+++ b/src/network/networkctl.c
@@ -2551,7 +2551,7 @@ static int link_lldp_status(int argc, char *argv[], void *userdata) {
                         (void) sd_lldp_neighbor_get_port_id_as_string(n, &port_id);
                         (void) sd_lldp_neighbor_get_system_name(n, &system_name);
                         (void) sd_lldp_neighbor_get_port_description(n, &port_description);
-
+/*
                         if (chassis_id) {
                                 cid = ellipsize(chassis_id, 17, 100);
                                 if (cid)
@@ -2575,7 +2575,7 @@ static int link_lldp_status(int argc, char *argv[], void *userdata) {
                                 if (pdesc)
                                         port_description = pdesc;
                         }
-
+*/
                         if (sd_lldp_neighbor_get_enabled_capabilities(n, &cc) >= 0) {
                                 capabilities = lldp_capabilities_to_string(cc);
                                 all |= cc;

```

None of the columns is truncated when executing `ninja -C build/ && ./build/networkctl lldp`, the table overflows my terminal. Terminal widths:

```
[nix-shell:~/systemd]$ tput cols
60
```

yuwata: Maybe, `table_set_minimum_width()` in the above is too large for your terminal width?

ackxolotl: Even with a width of `145` the table overflows as no string in any column is truncated. I have one very long name in the column `SYSTEM NAME`. Please try it out yourself, the column formatting definitely works not as expected.

yuwata:
Replaced with #23072.

PR #23076, opened 2022-04-14
Add ConditionCPUFeature to load-fragment-gperf.gperf

### LINKED ISSUE #23075: ConditionCPUFeature not recognized with systemd 248
Using systemd 248 (248.3-1ubuntu8.2) on  Ubuntu Impish (21.10) with kernel 5.13.0-1021 on amd64. systemd chokes on the following sample service unit that uses `ConditionCPUFeature`:

```
Unknown key name 'ConditionCPUFeature' in section 'Unit', ignoring.
```

Here is the sample service unit:
```
# Common configuration options for units of all types...
[Unit]
 
# Human readable description...
Description=My Server
 
# Don't start us until we have these things...
After=network.target                                                            \
    postgresql.service                                                          \
    auditd.service                                                              \
    remote-fs.target                                                            \
    time-sync.target                                                            \
    sound.target
 
# If the daemon is started but the database is not, start the database first...
Requires=postgresql.service
 
# When database is stopped or restarted, so should we...
PartOf=postgresql.service
 
# If this file exists, disable the service...
ConditionPathExists=!/etc/myserver/disable_myserver
 
# Server configuration must exist before launching. This is an array type so
#  overriding it in a drop-in requires resetting the value first...
AssertPathExists=/etc/myserver/server.conf
 
# Verify CPU features are available via CPUID instruction. This condition only
#  does something on x86 processors. On other processors it is assumed that the
#  CPU does not support the given feature. This requires systemd >= 248...
ConditionCPUFeature=avx
ConditionCPUFeature=avx2
ConditionCPUFeature=bmi1
ConditionCPUFeature=bmi2
 
 
# Configuration specific to system service...
[Service]
 
# Expect notification from daemon's ::sd_notify(3) usage...
Type=notify
 
# Location daemon writes out its process ID. This isn't really necessary because
#  systemd(1) can already identify the PID because it does the forking itself...
PIDFile=/run/myserver/myserver.pid
 
# Expect service updates (e.g. ::sd_notify(3)) from the main process only...
NotifyAccess=main
 
# Don't automatically restart the service if it exits, crashes, or a timeout is
#  reached...
Restart=no
 
# Expect that the daemon should not take any longer than three minutes at most
#  to receive READY notification after attempting to activate it...
TimeoutStartSec=3m
 
# If the daemon is still running after this much time from having asked it
#  politely (SIGTERM) to quit, terminate via SIGKILL...
TimeoutStopSec=3m
 
# Path to binary. This is an array type so overriding it in a drop-in requires
#  resetting the value first...
ExecStart=/usr/bin/myserver --pidfile=/run/myserver/myserver.pid --verbose
 
# Run as myserver user...
User=_myserver
 
# Create /run/myserver directory and bind it to the lifetime of our daemon's
#  runtime. This must be done through systemd because myserver is unprivileged
#  and therefore cannot create this runtime directory on its own...
RuntimeDirectory=myserver
 
# Maximum number of file descriptors that can be opened at any given time...
LimitNOFILE=65536
 
# Create our own private instance of /tmp and /var/tmp. Their life-cycles will
#  be bound to the service runtime...
PrivateTmp=true
 
# Common configuration options for units of all types. This is used by
#  systemctl(1) during enable and disable commands only...
[Install]
 
# Add our daemon as a dependency for setting up a non-graphical multi-user
#  system...
WantedBy=multi-user.target
```

This feature [should work](https://github.com/systemd/systemd/blob/4bcf5c11c8a951f9feef2964a721cc245b753873/src/shared/condition.c#L1267).

Edu4rdSHL:
- Fixes #23075


kiplingw (approved):

kiplingw (approved):

Edu4rdSHL:
The failing test is caused by a DNS problem in the container unrelated to the actual change.

```
E: Unable to locate package llvm-14
+ wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key
+ gpg --yes --dearmor --output /usr/share/keyrings/apt-llvm-org.gpg
--2022-04-14 02:13:37--  https://apt.llvm.org/llvm-snapshot.gpg.key
Resolving apt.llvm.org (apt.llvm.org)... failed: Temporary failure in name resolution.
wget: unable to resolve host address ‘apt.llvm.org’
gpg: no valid OpenPGP data found.
```

yuwata (approved):
LGTM. Thank you!

kiplingw:
It might be helpful if someone updated [NEWS](https://github.com/systemd/systemd/blob/main/NEWS) because this is a user visible change that, although previously announced in 248, it actually wasn't available for the above reasons until very recently.

Issue #19256, opened 2021-04-08
Toggle unit state - `start` if stopped and `stop` if running


tuxiqae:
**Is your feature request related to a problem? Please describe.**
The request is not related to a problem.

**Describe the solution you'd like**
 I think that it could be really efficient to have a single command which will `start` a unit if it is stopped and `stop` a unit if it is running.
Such a solution should be used for daily use of the terminal and not for script-writing.

**Describe alternatives you've considered**
Alternatives may include writing custom scripts which check the state of the unit and acts accordingly.

**The systemd version you checked that didn't have the feature you are asking for**
248.
I have also checked the man and the repository and couldn't find any code or issues regarding such feature.

amarjargal:
Hello everyone. I'm an outreachy applicant hoping to make some contributions to systemd. I'd like to work on this issue if it's still available.

tuxiqae:
Still is as far as I'm aware

amarjargal:
I tried to tackle this issue in PR #23077. Please let me know if this solution is OK.

poettering:
Hmm, anyone knows what the precise usecase for this is? i.e. why would you want to an issue where you can't guess the possible outcome? Would be good to shed some light on this.

behrmann:
I would like to have something like this mostly during development, where I start a service, see that something isn't working the way I want it to, stop the service, tweak the config and/or recompile, start the service again. It's not like this is something I couldn't live without, I have `systemctl start foo.service` and `systemctl stop foo.service` in my shell's history and I won't do much typing anyway, but it would still be a quality of life thing.

I do think, though, it would make sense if this command were verbose by default, e.g.
```
$ systemctl toggle foo.service
starting foo.service
$ systemctl toggle foo.service
stoppping foo.service
```

james-lawrence:
@behrmann that sounds like 'restart' does what you want.

tuxiqae:
> 

But it doesn't, `restart` `stop`s and then **immediately** `start`s the service.
Toggle does one or the other at a time, not both.

tuxiqae:
A better example might be the following:

```
$ systemctl toggle foo.service
starting foo.service

# ... Do something ...

$ systemctl toggle foo.service
stoppping foo.service
```

james-lawrence:
not really. doesn't explain why going from unknown state -> unknown state is useful or desirable.

- if you want to ensure a service is running but don't want to do anything if its already running -> use systemctl start.
[...]

PR #23077, opened 2022-04-14
systemctl: add a command that toggles unit state - start if inactive and stop if active

amarjargal:
Fixes: https://github.com/systemd/systemd/issues/19256

### DIFF
--- a/man/systemctl.xml
+++ b/man/systemctl.xml
@@ -451,6 +451,14 @@ Jan 12 10:46:45 example.com bluetoothd[8900]: gatt-time-server: Input/output err
           </listitem>
         </varlistentry>
         <varlistentry>
+          <term><command>toggle <replaceable>PATTERN</replaceable>…</command></term>
+
+          <listitem>
+            <para>Toggle the states of one or more units specified on the command line: start the units if inactive
+            (dead) and stop if active (running).</para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
           <term><command>isolate <replaceable>UNIT</replaceable></command></term>
 
           <listitem>
--- a/src/systemctl/systemctl-start-unit.c
+++ b/src/systemctl/systemctl-start-unit.c
@@ -399,3 +399,42 @@ int verb_start(int argc, char *argv[], void *userdata) {
 
         return ret;
 }
+
+int verb_toggle(int argc, char *argv[], void *userdata) {
+        _cleanup_strv_free_ char **names = NULL;
+        sd_bus *bus;
+        int r, ret = EXIT_SUCCESS;
+
+        if (!argv[1])
+                return 0;
+
+        r = mangle_names("to toggle", strv_skip(argv, 1), &names);
+        if (r < 0)
+                return r;
+
+        r = acquire_bus(BUS_MANAGER, &bus);
+        if (r < 0)
+                return r;
+
+        STRV_FOREACH(name, names) {
+                UnitActiveState active_state;
+                bool is_active = false;
+                char *new_args[3];
+
+                r = get_state_one_unit(bus, *name, &active_state);
+                if (r < 0)
+                        return r;
+
+                is_active = IN_SET(active_state, UNIT_ACTIVE, UNIT_RELOADING);
+
+                new_args[0] = (char*) (is_active ? "stop" : "start");
+                new_args[1] = basename(*name);
+                new_args[2] = NULL;
+
+                r = verb_start(2, new_args, userdata);
+                if (r < 0)
+                        ret = EXIT_FAILURE;
+        }
+
+        return ret;
+}
--- a/src/systemctl/systemctl.c
+++ b/src/systemctl/systemctl.c
@@ -161,6 +161,7 @@ static int systemctl_help(void) {
                "                                      otherwise start or restart\n"
                "  try-reload-or-restart UNIT...       If active, reload one or more units,\n"
                "                                      if supported, otherwise restart\n"
+               "  toggle UNIT...                      Start inactive or stop active units\n"
                "  isolate UNIT                        Start one unit and stop all others\n"
                "  kill UNIT...                        Send signal to processes of a unit\n"
                "  clean UNIT...                       Clean runtime, cache, state, logs or\n"
@@ -1040,6 +1041,7 @@ static int systemctl_main(int argc, char *argv[]) {
                 { "condreload",            2,        VERB_ANY, VERB_ONLINE_ONLY, verb_start                   }, /* For compatibility with ALTLinux */
                 { "condrestart",           2,        VERB_ANY, VERB_ONLINE_ONLY, verb_start                   }, /* For compatibility with RH */
                 { "isolate",               2,        2,        VERB_ONLINE_ONLY, verb_start                   },
+                { "toggle",                2,        VERB_ANY, VERB_ONLINE_ONLY, verb_toggle                  },
                 { "kill",                  2,        VERB_ANY, VERB_ONLINE_ONLY, verb_kill                    },
                 { "clean",                 2,        VERB_ANY, VERB_ONLINE_ONLY, verb_clean_or_freeze         },
                 { "freeze",                2,        VERB_ANY, VERB_ONLINE_ONLY, verb_clean_or_freeze         },
--- a/test/units/testsuite-48.sh
+++ b/test/units/testsuite-48.sh
@@ -81,6 +81,13 @@ systemctl restart testservice-48.target
 
 systemctl is-active testservice-48.service
 
+# Check toggle
+state="$(systemctl is-active testservice-48.service)"
+systemctl toggle testservice-48.service
+test "${state}" != "$(systemctl is-active testservice-48.service)"
+systemctl toggle testservice-48.service
+test "${state}" == "$(systemctl is-active testservice-48.service)"
+
 echo OK >/testok
 
 exit 0
--- a/src/systemctl/systemctl-start-unit.h
+++ b/src/systemctl/systemctl-start-unit.h
@@ -4,6 +4,7 @@
 #include "systemctl.h"
 
 int verb_start(int argc, char *argv[], void *userdata);
+int verb_toggle(int argc, char *argv[], void *userdata);
 
 struct action_metadata {
         const char *target;


src/systemctl/systemctl-start-unit.c
```diff
@@ -15,6 +15,7 @@
 #include "systemctl-util.h"
 #include "systemctl.h"
 #include "terminal-util.h"
+#include "systemctl-is-active.h"
```
bluca: sort includes alphabetically please

src/systemctl/systemctl.c
```diff
@@ -161,6 +161,8 @@ static int systemctl_help(void) {
                "                                      otherwise start or restart\n"
                "  try-reload-or-restart UNIT...       If active, reload one or more units,\n"
                "                                      if supported, otherwise restart\n"
+               "  toggle UNIT...                      Start a unit if it's active (running) and\n"
```
bluca: Opposite? Start if inactive, stop if active

bluca (changes requested):
Nice! Please update the documentation as well `man/systemctl.xml`. 

Also it would be nice to have test coverage - no need to add an entire new test case, so maybe just call toggle a couple of times and check the state in `test/units/testsuite-48.sh`?

src/systemctl/systemctl.c
amarjargal: Oops, yes it's the opposite. I'll fix it.

src/systemctl/systemctl-start-unit.c
amarjargal: OK

amarjargal:
> Nice! Please update the documentation as well `man/systemctl.xml`.
> 
> Also it would be nice to have test coverage - no need to add an entire new test case, so maybe just call toggle a couple of times and check the state in `test/units/testsuite-48.sh`?

I've been looking through  `test/units/testsuite-48.sh` to add the tests. But I couldn't run this script because 'testservice-48.service' was not found. Was 'testservice-48.service' in the script meant to be 'testsuite-48.service'? Or am I missing something?

bluca:
> > Nice! Please update the documentation as well `man/systemctl.xml`.
> > Also it would be nice to have test coverage - no need to add an entire new test case, so maybe just call toggle a couple of times and check the state in `test/units/testsuite-48.sh`?
> 
> I've been looking through `test/units/testsuite-48.sh` to add the tests. But I couldn't run this script because 'testservice-48.service' was not found. Was 'testservice-48.service' in the script meant to be 'testsuite-48.service'? Or am I missing something?

You cannot run it manually, it's part of the integration suite, and it will run in a VM/container. If you have all the build dependencies installed, this should run it:

```
sudo make -C test/TEST-48-START-STOP-NO-RELOAD/
```

See `test/README.testsuite` - but also feel free to just push and let the CI run it while you get it running locally, it should be a simple change

src/systemctl/systemctl.c
```diff
@@ -161,6 +161,8 @@ static int systemctl_help(void) {
                "                                      otherwise start or restart\n"
                "  try-reload-or-restart UNIT...       If active, reload one or more units,\n"
                "                                      if supported, otherwise restart\n"
+               "  toggle UNIT...                      Start a unit if it's inactive (dead) and\n"
```
medhefgo: Maybe just `Start inactive or stop active units`?

test/units/testsuite-48.sh
```diff
@@ -81,6 +81,17 @@ systemctl restart testservice-48.target
 
 systemctl is-active testservice-48.service
 
+# Check toggle
+for i in 1 2
```
medhefgo: we prefer `for ...; do`

medhefgo: Though, You can just do this to also get around the lint error (as in, do away with the loop):

```
state="$(systemctl is-active testservice-48.service)"
systemctl toggle testservice-48.service
test "${state}" != "$(systemctl is-active testservice-48.service)"
systemctl toggle testservice-48.service
test "${state}" == "$(systemctl is-active testservice-48.service)"
```

amarjargal: Yes, this is much better, thank you!

medhefgo (approved):

medhefgo:
Please add `Fixes: #19256` to the commit message for these kinds of changes. It'll auto-close the linked issues on merge.

bluca (approved):

amarjargal:
> Please add `Fixes: #19256` to the commit message for these kinds of changes. It'll auto-close the linked issues on merge.

Got it. I was wondering how to link the original issue. Will do the next time.

src/systemctl/systemctl-show.c
```diff
@@ -298,12 +298,24 @@ static void format_active_state(const char *active_state, const char **active_on
                 *active_on = *active_off = "";
 }
 
+static void format_enable_state(const char *enable_state, const char **enable_on, const char **enable_off) {
+        if (streq_ptr(enable_state, "disabled")) {
+                *enable_on = ansi_highlight_red();
```
bluca: The ultimate bikeshedding! Red is usually associated with an error, but disabled is not an error - maybe yellow?

amarjargal: Sorry, I was meant to create another pull request so removed this one. Still trying to figure out git.

bluca: no problem at all

amarjargal: I guess, if I want to create new pull request while the previous one is still in progress, I have to checkout new branch?

bluca: yes, just use a different branch (and base it on main, rather than on the other topic branch

amarjargal:
I tried to look into the failing test log but I'm at lost. Can you please help me how to fix the failing test?

bluca:
> I tried to look into the failing test log but I'm at lost. Can you please help me how to fix the failing test?

Don't worry about that one, it's a flacky clock test that is unrelated to your changes

man/systemctl.xml
```diff
@@ -450,6 +450,14 @@ Jan 12 10:46:45 example.com bluetoothd[8900]: gatt-time-server: Input/output err
                  don't document that. -->
           </listitem>
         </varlistentry>
+        <varlistentry>
+          <term><command>toggle <replaceable>PATTERN</replaceable>…</command></term>
+
+          <listitem>
+            <para>Toggle the states of one or more units specified on the command line: start the units if inactive
+            (dead) and stop if active (running)</para>
```
yuwata: Missing period at the end.

test/units/testsuite-48.sh
```diff
@@ -81,6 +81,14 @@ systemctl restart testservice-48.target
...
 systemctl is-active testservice-48.service
 
+# Check toggle
+state="$(systemctl is-active testservice-48.service)"
+systemctl toggle testservice-48.service
+test "${state}" != "$(systemctl is-active testservice-48.service)"
+systemctl toggle testservice-48.service
+test "${state}" == "$(systemctl is-active testservice-48.service)"
+
+
```
yuwata: duplicated blank line

src/systemctl/systemctl-start-unit.c
```diff
@@ -298,6 +299,17 @@ int verb_start(int argc, char *argv[], void *userdata) {
                                 job_type = "start";
                                 mode = "isolate";
                                 suffix = ".target";
+                        } else if (streq(argv[0], "toggle")) {
+                                /* A "systemctl toggle <unit1> <unit2> …" command */
+                                r = verb_is_active(argc, argv, userdata);
```
yuwata: If unit1 is active and unit2 is inactive, then this tries to stop both unit1 and unit2... But, this should stop unit1 and start unit2 in that case.

Also, `verb_is_active()` -> `check_unit_generic()` prints unit state when `arg_quiet` is false. The output should not be printed here.

yuwata (changes requested):

src/systemctl/systemctl-start-unit.c
amarjargal: Yes. I'll fix them and update.

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +400,49 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
+                return r;
+
+        r = acquire_bus(BUS_MANAGER, &bus);
+        if (r < 0)
+                return r;
+
+        STRV_FOREACH(name, names) {
+                is_active = false;
+
+                r = get_state_one_unit(bus, names[j], &active_state);
```
bluca: just use `name` and drop `j`

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +400,49 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
+                if (r < 0)
+                        return r;
+
+                for (i = 0; i < nb_states; ++i)
```
bluca: declare `i` inline

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +400,49 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
+                        if (good_states[i] == active_state)
+                                is_active = true;
```
bluca: no need to loop, you can just use `is_active = IN_SET(active_state, UNIT_ACTIVE, UNIT_RELOADING)`

amarjargal: thank you for pointing them out, the code is so much better! pushing new commit now.

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +400,41 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
         return ret;
 }
+
+int verb_toggle(int argc, char *argv[], void *userdata) {
+        _cleanup_strv_free_ char **names = NULL;
+        UnitActiveState active_state;
+        sd_bus *bus;
+        int r;
+        bool is_active = false;
+        char *new_args[3];
```
bluca: You can move the declaration of new_args, is_active and active_state inside the STRV_FOREACH

medhefgo: You beat me to it.

Also, please add the "Fixes: #19256" to the commit message while you're at it.

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +400,41 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
+                if (r < 0)
+                        return r;
+
+                is_active = IN_SET(active_state, UNIT_ACTIVE, UNIT_RELOADING);
+
+                new_args[0] = (char*) (is_active ? "stop" : "start");
+                new_args[1] = basename(*name);
+                new_args[2] = NULL;
+
+                r = verb_start(2, new_args, userdata);
```
bluca, line 434: return value will be overwritten if there's more than one unit, it should be checked immediately

src/systemctl/systemctl-start-unit.c
amarjargal: Done!

bluca (approved):

src/systemctl/systemctl-start-unit.c
medhefgo: Shouldn't this rather try to toggle the remaining units and return the first error then? How do the other verbs behave?

amarjargal: I only checked start verb, and it does go through all the units even if there's one erroneous unit among them and return that error. I'll make a fix.

amarjargal: Thinking further, maybe it's better to leave it as before without checking? That way it still prints the error message but also continues on toggling other units.

bluca: The usual pattern we use is something like:

```
int r = 0, k;

k = foo()
if (k < 0 && r == 0)
     r = k;
```

This way if the first one errors out and the last one doesn't, we still return an error, which is important as that's what's expected

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +399,42 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
```
daandemeyer, line 434: We shouldn't limit ourselves to the interface of `verb_start` for this. It's interface is specifically designed to take arguments from the command line which is not what we're doing here.

What I'd do instead is to split out a common function out of `verb_start` that does the stuff that's needed for both `verb_start` and `verb_toggle` and to then call that function from both `verb_start` and `verb_toggle`.

The interface of this function would probably look something like:

```
int do_action(enum action action, const char **units)
```

src/systemctl/systemctl-start-unit.c
```diff
@@ -399,3 +399,42 @@ int verb_start(int argc, char *argv[], void *userdata) {
...
+                return r;
+
+        STRV_FOREACH(name, names) {
+                UnitActiveState active_state;
+                bool is_active = false;
+                char *new_args[3];
+
+                r = get_state_one_unit(bus, *name, &active_state);
+                if (r < 0)
+                        return r;
```
daandemeyer, line 426: Doing all units one by one isn't terribly efficient, especially if the --wait option is provided. Since systemctl will wait for each unit to be fully toggled before sending the request for the next unit to be toggled.

Ideally, we'd first get the active state of each provided unit, start/stop them all at once and only then wait for all the jobs to complete.

DaanDeMeyer:
I also think a better implementation of this feature could be to add a ToggleUnit D-Bus method which is called directly from systemctl just like StartUnit and StopUnit. Aside from the boilerplate to add the new method, the main logic change would be in `job_type_collapse()` I think where we'd check the active unit state and change the job type to `JOB_START` or `JOB_STOP` depending on the unit active state.

poettering:
Hmm, anyone knows what the precise usecase for this is? i.e. why would you want to an issue where you can't guess the possible outcome? Would be good to shed some light on this.

Regarding implementaiton: i am pretty sure this should happen server-side, i.e. extend the existing D-Bus calls EnqueueJob()/EnqueueUnitJob() to expose this behaviour, so that we can make it a single bus call to request this.

poettering:
specifically: extend bus_unit_method_enqueue_job() in src/core/dbus-unit.c to look for a new "toggle" job type, and then mangle that into JOB_START or JOB_STOP as needed. There are already two job types handled specially, this would be a third.

client side would become relavtively simple I think, then, i.e. unit_actions[] table in src/systemctl/systemctl-start-unit.c would get a new entry, and otherwise client hookup could probably reuse most of verb_start() unmodified.

amarjargal:
Hmm, I tried to solve this by touching the existing code as little as possible. But I see that it's not a good way and it needs to cover more areas than I anticipated. The outreachy contribution period is almost over. But can I still work on this? I want to study the code more and try to implement it per your suggestions.

bluca:
> Hmm, I tried to solve this by touching the existing code as little as possible. But I see that it's not a good way and it needs to cover more areas than I anticipated. The outreachy contribution period is almost over. But can I still work on this? I want to study the code more and try to implement it per your suggestions.

Of course, if you have the time, please do so

src/systemctl/systemctl-start-unit.c
tuxiqae: What if the units provided are a mix of stopped and active?
Would you run 2 jobs, one for stop and one for start?

PR #23080, opened 2022-04-14
Redundant code should be removed because manager_dispatch_delayed() does not return a value less than 0 after this PR https://github.com/systemd/systemd/commit/8207b8321bbbcbd19a345deb77d455d98e6ffb84

jcg190701:
Redundant code should be removed because manager_dispatch_delayed() does not return a value less than 0 after this PR https://github.com/systemd/systemd/commit/8207b8321bbbcbd19a345deb77d455d98e6ffb84

### DIFF
--- a/src/login/logind.c
+++ b/src/login/logind.c
@@ -1142,8 +1142,6 @@ static int manager_run(Manager *m) {
                 manager_gc(m, true);
 
                 r = manager_dispatch_delayed(m, false);
-                if (r < 0)
-                        return r;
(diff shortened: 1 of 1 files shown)


poettering:
Nah, let's keep these functions reasonably abstract. We generally return errors as negative errno values. And if some function currently cannot fail, that's OK, but this fact should not be leaked into the caller code. 

The compiler should be able to recognize this on its own, and optimize it away.

PR #23083, opened 2022-04-14
man: update TPM2 PCR documentation

poettering:
The assignments where partly simply incorrect documented, partly changed
with 4d32507f5186a89e98093659fbbe386787a97b9f and partly where missing.
Moreover kernel 5.17 now measures all initrds to PCR 9 on its own
(https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f046fff8bc4c4d8f8a478022e76e40b818f692df)

Let's correct all this and bring it up-to-date.

And while we are add it extend the docs about this in systemd-stub, with
a new table that indicates which OS resource is protected by which PCR.

### DIFF
--- a/man/systemd-cryptenroll.xml
+++ b/man/systemd-cryptenroll.xml
@@ -251,7 +251,7 @@
 
               <row>
                 <entry>4</entry>
-                <entry>Boot loader; changes on boot loader updates. The shim project will measure the PE binary it chain loads into this PCR.</entry>
+                <entry>Boot loader and additional drivers; changes on boot loader updates. The shim project will measure the PE binary it chain loads into this PCR. If the Linux kernel is invoked as UEFI PE binary, it is measured here, too. <citerefentry><refentrytitle>sd-stub</refentrytitle><manvolnum>7</manvolnum></citerefentry> measures system extension images read from the ESP here too (see <citerefentry><refentrytitle>systemd-sysext</refentrytitle><manvolnum>8</manvolnum></citerefentry>).</entry>
               </row>
 
               <row>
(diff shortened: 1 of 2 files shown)


man/systemd-cryptenroll.xml
```diff
@@ -273,15 +273,21 @@
...
               </row>
 
               <row>
                 <entry>10</entry>
                 <entry>The IMA project measures its runtime state into this PCR.</entry>
               </row>
 
+              <row>
+                <entry>12</entry>
+                <entry><citerefentry><refentrytitle>sd-stub</refentrytitle><manvolnum>7</manvolnum></citerefentry> measures any manually specified kernel command line and loaded credentials into this PCR.</entry>
```
medhefgo: Actualy, sd-boot measures cmdline options too. (So they end up getting logged twice, lul.)

bluca (approved):

poettering:
updated as suggested. ptal.

poettering:
a, well, will upgrade the green label, it's just docs after all, and the weak green label was already set. of course, reviews are still welcome ;-)

poettering:
> Btw, did you that github lets you can directly request reviews on the sidebar to the right? No need to create these silly "/cc" comments. stuck_out_tongue

Yeah, I am aware. I mostly do the cc thing because assigning someone as a reviewer seems quite a strong operation to me. i.e. to me it feels as if assigning you as a reviewer would communicate to you that I feel like being your manager or so, and I could assign you tasks like a manager would. Which of course is complete rubbish. After all, mostly here I am just politely asking for your input, your take, your opinion here, if you can arrange for providing it, and have time, and feel like it. a quick "cc" line transports that better I think. But of course, maybe these nuances exist in my head only...

I'll keep in mind to properly assign you as a reviewer in future now. ;-)

bluca:
> > Btw, did you that github lets you can directly request reviews on the sidebar to the right? No need to create these silly "/cc" comments. stuck_out_tongue
> 
> Yeah, I am aware. I mostly do the cc thing because assigning someone as a reviewer seems quite a strong operation to me. i.e. to me it feels as if assigning you as a reviewer would communicate to you that I feel like being your manager or so, and I could assign you tasks like a manager would. Which of course is complete rubbish. After all, mostly here I am just politely asking for your input, your take, your opinion here, if you can arrange for providing it, and have time, and feel like it. a quick "cc" line transports that better I think. But of course, maybe these nuances exist in my head only...
> 
> I'll keep in mind to properly assign you as a reviewer in future now. ;-)

One can always remove themselves, I think - the nice thing about the interface is that all pending review requests appear under https://github.com/pulls/assigned so it's much easier to keep track of TODOs

PR #23084, opened 2022-04-14
creds: add semi-automatic fallback support for initrd credentials on systems lacking TPM2

poettering:
credentials are supposed to be useful to pass parameters securely to signed, unified kernels. For that it's great to encrypt/sign them against a key in the TPM2 chip.

unfortunately though there are systems lacking a TPM2 chip. Let's implement some reasonable fallback logic for cases like this. Specifically: `systemd-creds encrypt --with-key=auto-initrd` can now be used to either encrypt/sign with a key in the TPM2 chip (if that's available) or with an automatic fallback to a fixed zero length key (i.e. a providing no confidentiality nor authenticity). 

Effect of this: a credential encrypted like this intended for the initrd will be protected as good as the local hw allows it, but not better.

### DIFF
--- a/TODO
+++ b/TODO
@@ -114,12 +114,6 @@ Features:
   - sd-stub: automatically pick up microcode from ESP (/loader/microcode/*)
     and synthesize initrd from it, and measure it. Signing is not necessary, as
     microcode does that on its own. Pass as first initrd to kernel.
-  - systemd-creds should have a fallback logic that uses neither TPM nor the
-    system key in /var for encryption and instead some fixed key. This should
-    be opt in (since it provides no security properties) but be used by
-    kernel-install when encrypting the creds it generates on systems that lack
-    a TPM, so that we can have very similar codepaths on TPM and TPM-less
-    systems. i.e. --with-key=tpm-graceful or so.
   - sd-stub should measure the kernel/initrd/… into a separate PCR, so that we
     have one PCR we can bind the encrypted creds to that is not effected by
     anything else but what we drop in via kernel-install, i.e. by earlier EFI


src/shared/creds-util.c
```diff
@@ -475,7 +475,7 @@ int encrypt_credential_and_warn(
         uint8_t md[SHA256_DIGEST_LENGTH];
         const EVP_CIPHER *cc;
 #if HAVE_TPM2
-        bool try_tpm2 = false;
+        bool try_tpm2;
```
medhefgo: Makes more sense to move this further down, would reduce the number of #ifdefs.

src/shared/creds-util.h
```diff
@@ -43,5 +43,10 @@ int get_credential_host_secret(CredentialSecretFlags flags, void **ret, size_t *
 #define CRED_AES256_GCM_BY_TPM2_HMAC          SD_ID128_MAKE(0c,7c,c0,7b,11,76,45,91,9c,4b,0b,ea,08,bc,20,fe)
 #define CRED_AES256_GCM_BY_HOST_AND_TPM2_HMAC SD_ID128_MAKE(93,a8,94,09,48,74,44,90,90,ca,f2,fc,93,ca,b5,53)
 
+/* Special ID to pick automatic mode (i.e. tpm2+host if TPM2 exists, only host otherwise). This ID will never
+ * be stored on disk, but is useful only internally while figuring out what precisely to write to disk. To
+ * mark that this isn't a "real" type, we'll prefix it with an underscore. */
+#define _CRED_AUTO                            SD_ID128_MAKE(a2,19,cb,07,85,b2,4c,04,b1,6d,18,ca,b9,d2,ee,01)
```
medhefgo, line 55: Maybe `CRED_AUTO_INTERNAL`?

src/shared/creds-util.c
```diff
@@ -572,12 +589,17 @@ int encrypt_credential_and_warn(
...
+                else if (sd_id128_equal(with_key, _CRED_AUTO_INITRD))
+                        id = CRED_AES256_GCM_BY_TPM2_ABSENT;
                 else
                         return log_error_errno(SYNTHETIC_ERRNO(ENOTRECOVERABLE),
                                                "TPM2 not available and host key located on temporary file system, no encryption key available.");
         } else
                 id = with_key;
 
+        if (sd_id128_equal(id, CRED_AES256_GCM_BY_TPM2_ABSENT))
+                log_warning("Warning: using a null key for encryption and signing. Confidentiality or authenticity will not be be provided.");
```
medhefgo: It's already a log_warning, does a "Warning:" prefix really make sense? (the message already gets a different output color, no?

src/shared/creds-util.h
```diff
@@ -38,15 +38,22 @@ typedef enum CredentialSecretFlags {
...
-/* Special ID to pick automatic mode (i.e. tpm2+host if TPM2 exists, only host otherwise). This ID will never
- * be stored on disk, but is useful only internally while figuring out what precisely to write to disk. To
- * mark that this isn't a "real" type, we'll prefix it with an underscore. */
+/* Two special IDs to pick a general automatic mode (i.e. tpm2+host if TPM2 exists, only host otherwise) or
+ * an initrd-specific automatic mode (i.e. tpm2 if firmware can do it, otherwise fixed zero-length key, and
+ * never involve host keys). These IDs will never be stored on disk, but are useful only internally while
+ * figuring out what precisely to write to disk. To mark that these aren't a "real" type, we'll prefix them
+ * with an underscore. */
 #define _CRED_AUTO                            SD_ID128_MAKE(a2,19,cb,07,85,b2,4c,04,b1,6d,18,ca,b9,d2,ee,01)
+#define _CRED_AUTO_INITRD                     SD_ID128_MAKE(02,dc,8e,de,3a,02,43,ab,a9,ec,54,9c,05,e6,a0,71)
```
medhefgo, line 56: And `CRED_AUTO_INITRD_INTERNAL`? (I really hate _names)

man/systemd-creds.xml
```diff
@@ -263,23 +263,31 @@
...
+        <literal>auto-initrd</literal>. See above for details on the three key types. If set to
+        <literal>auto</literal> (which is the default) the TPM2 key is used if a TPM2 device is found and not
+        running in a container. The host key is used if <filename>/var/lib/systemd/</filename> is on
+        persistent media. This means on typical systems the encryption is by default bound to both the TPM2
+        chip and the OS installation, and both need to be available to decrypt the credential again. If
+        <literal>auto</literal> is selected but neither TPM2 is available (or running in container) nor
+        <filename>/var/lib/systemd/</filename> is on persistent media, encryption will fail. If set to
+        <literal>tpm2-absent</literal> a fixed zero length key is used (thus, in this mode no confidentiality
+        nor authenticity are provided!). This logic is useful to cover for systems that lack a TPM2 chip but
+        where credentials shall be generated. Note that decryption of such credentials is refused on systems
```
medhefgo, line 278: I was gonna ask the reason behind this but then saw the code comment later on. I think it makes sense to add an explanation to the docs.

src/shared/creds-util.c
yuwata: "be be"

src/shared/creds-util.h
poettering: usually we use the `_` prefix to mark enums and similar that are kinda "special", and not real things, hence I just followed the same logic here, given this isn't a "real" cred key type, but just an internal placeholder... 

i guess whether to name stuff xxxx_INTERNAL or _xxxx is ultimately bikeshedding, no ;-)

src/shared/creds-util.h
poettering: well, as mentioned, it's kinda established already in our codebase...

poettering: i.e. for regular enums we have _MAX and _INVALID special values usually, and they are underscored, to indicate that, and set them apart in tables.

poettering:
force pushed a new version. fixed everything @medhefgo and @yuwata found (except for the underscore vs. INTERNAL thing)

PR #12592, opened 2019-05-16
[WIP] journal: support match negation

mrc0mmand:
To get myself familiar further with the systemd codebase (and to fix an issue which has been nagging me for a while), I implemented (or at least tried to) a fix for https://github.com/systemd/systemd/issues/2720.

First of all, this is just a (working, so far) mockup, as I'd like to get a feedback if this is even *the right* approach.

To avoid collisions with bash operators and such (i.e. using `!` or `-` for negation) I chose `~` as a prefix for a field user wants to NOT match.

For example:
`journalctl -b ~MESSAGE="this is sparta"`
shows all messages for this boot which are not "this is sparta".

```
$ build/journalctl --no-pager --file system.journal --since "2019-01-14 14:21:13" | tail -n 4
Jan 14 14:21:13 pyrelight systemd[1]: Stopped Getty on tty2.
Jan 14 14:21:13 pyrelight systemd[1]: Stopped Getty on tty5.
Jan 14 14:21:13 pyrelight systemd[1]: Stopped Getty on tty3.
Jan 14 14:21:13 pyrelight kernel: audit: type=1131 audit(1547472073.653:1033): pid=8784 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:spc_t:s0 msg='unit=getty@tty3 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'

$ build/journalctl --no-pager --file system.journal --since "2019-01-14 14:21:13" MESSAGE="Stopped Getty on tty3." | tail -n 4
-- Logs begin at Mon 2019-01-14 14:15:24 CET, end at Mon 2019-01-14 14:21:13 CET. --
Jan 14 14:21:13 pyrelight systemd[1]: Stopped Getty on tty3.

$ build/journalctl --no-pager --file system.journal --since "2019-01-14 14:21:13" ~MESSAGE="Stopped Getty on tty3." | tail -n 4
Jan 14 14:21:13 pyrelight kernel: audit: type=1131 audit(1547472073.653:1032): pid=8784 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:spc_t:s0 msg='unit=getty@tty5 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Jan 14 14:21:13 pyrelight systemd[1]: Stopped Getty on tty2.
Jan 14 14:21:13 pyrelight systemd[1]: Stopped Getty on tty5.
[...]

PR #23086, opened 2022-04-15
journalctl: add --exclude-identifier option

zhmylove:
### Exclude syslog identifiers
There is `-t` (`--identifier=STRING`) journalctl option which allows to specify `SYSLOG_IDENTIFIER` in a handy manner. This option could be used several times, each of them appends new layer of OR condition. A lot of `journalctl(1)` users are using it to show log entries related to only particular syslog identifiers. There are a lot of situations when it's required not to _select_ a particular syslog identifiers, but to _exclude_ some of them leaving all the others visible. There is a real problem in implementing this using Match* functionality built into journals as it match any part of the object by hash, while this could be effectively achieved during the journal entries output. This PR introduces `-T` (`--exclude=STRING`) option which implements this functionality for output_short (output_verbose, output_json, and others are intentionally left intact so far). 
Example:
```
# ./build/journalctl |head -5
Apr 15 02:05:40 rc1 systemd-logind[364]: Removed session c11.
Apr 15 02:06:01 rc1 CRON[23073]: pam_unix(cron:session): session opened for user root by (uid=0)
Apr 15 02:06:01 rc1 CRON[23074]: (root) CMD (/root/redis/balancer.pl)
Apr 15 02:06:01 rc1 CRON[23073]: pam_unix(cron:session): session closed for user root
Apr 15 02:06:39 rc1 su[23076]: (to root) korg on pts/0
# ./build/journalctl -T CRON |head -5
Apr 15 02:05:40 rc1 systemd-logind[364]: Removed session c11.
Apr 15 02:06:39 rc1 su[23076]: (to root) korg on pts/0
Apr 15 02:06:39 rc1 su[23076]: pam_unix(su-l:session): session opened for user root by korg(uid=1000)
Apr 15 02:06:39 rc1 systemd-logind[364]: New session c12 of user root.
Apr 15 02:06:39 rc1 systemd[1]: Started Session c12 of user root.
# ./build/journalctl -T CRON -T su |head -5
Apr 15 02:05:40 rc1 systemd-logind[364]: Removed session c11.
Apr 15 02:06:39 rc1 systemd-logind[364]: New session c12 of user root.
Apr 15 02:06:39 rc1 systemd[1]: Started Session c12 of user root.
Apr 15 02:06:47 rc1 systemd[1]: session-c12.scope: Succeeded.
Apr 15 02:06:47 rc1 systemd-logind[364]: Session c12 logged out. Waiting for processes to exit.
```

### DIFF
--- a/man/journalctl.xml
+++ b/man/journalctl.xml
@@ -363,6 +363,18 @@
       </varlistentry>
 
       <varlistentry>
+        <term><option>-T</option></term>
+        <term><option>--exclude-identifier=<replaceable>SYSLOG_IDENTIFIER</replaceable></option></term>
+
+        <listitem><para>Exclude messages for the specified syslog identifier
+        <replaceable>SYSLOG_IDENTIFIER</replaceable>.</para>
+
+        <para>This parameter can be specified multiple times.</para>
+
+        <xi:include href="version-info.xml" xpointer="v256"/></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><option>-p</option></term>
         <term><option>--priority=</option></term>
 
--- a/shell-completion/bash/journalctl
+++ b/shell-completion/bash/journalctl
@@ -46,9 +46,9 @@ _journalctl() {
                       --show-cursor --dmesg -k --pager-end -e -r --reverse
                       --utc -x --catalog --no-full --force --dump-catalog
                       --flush --rotate --sync --no-hostname -N --fields'
-        [ARG]='-b --boot -D --directory --file -F --field -t --identifier --facility
-                      -M --machine -o --output -u --unit --user-unit -p --priority
-                      --root --case-sensitive'
+        [ARG]='-b --boot -D --directory --file -F --field -t --identifier
+                      -T --exclude-identifier --facility -M --machine -o --output
+                      -u --unit --user-unit -p --priority --root --case-sensitive'
         [ARGUNKNOWN]='-c --cursor --interval -n --lines -S --since -U --until
                       --after-cursor --cursor-file --verify-key -g --grep
                       --vacuum-size --vacuum-time --vacuum-files --output-fields'
@@ -107,7 +107,7 @@ _journalctl() {
             --user-unit)
                 comps=$(journalctl -F '_SYSTEMD_USER_UNIT' 2>/dev/null)
                 ;;
-            --identifier|-t)
+            --identifier|-t|--exclude-identifier|-T)
                 comps=$(journalctl -F 'SYSLOG_IDENTIFIER' 2>/dev/null)
                 ;;
             --case-sensitive)
--- a/src/journal/journalctl.c
+++ b/src/journal/journalctl.c
@@ -122,6 +122,7 @@ static bool arg_force = false;
 static usec_t arg_since = 0, arg_until = 0;
 static bool arg_since_set = false, arg_until_set = false;
 static char **arg_syslog_identifier = NULL;
+static char **arg_exclude_identifier = NULL;
 static char **arg_system_units = NULL;
 static char **arg_user_units = NULL;
 static const char *arg_field = NULL;
@@ -146,6 +147,7 @@ STATIC_DESTRUCTOR_REGISTER(arg_file, strv_freep);
 STATIC_DESTRUCTOR_REGISTER(arg_facilities, set_freep);
 STATIC_DESTRUCTOR_REGISTER(arg_verify_key, freep);
 STATIC_DESTRUCTOR_REGISTER(arg_syslog_identifier, strv_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_exclude_identifier, strv_freep);
 STATIC_DESTRUCTOR_REGISTER(arg_system_units, strv_freep);
 STATIC_DESTRUCTOR_REGISTER(arg_user_units, strv_freep);
 STATIC_DESTRUCTOR_REGISTER(arg_root, freep);
@@ -380,6 +382,8 @@ static int help(void) {
                "  -u --unit=UNIT             Show logs from the specified unit\n"
                "     --user-unit=UNIT        Show logs from the specified user unit\n"
                "  -t --identifier=STRING     Show entries with the specified syslog identifier\n"
+               "  -T --exclude-identifier=STRING\n"
+               "                             Hide entries with the specified syslog identifier\n"
                "  -p --priority=RANGE        Show entries with the specified priority\n"
                "     --facility=FACILITY...  Show entries with the specified facilities\n"
                "  -g --grep=PATTERN          Show entries with MESSAGE matching PATTERN\n"
@@ -519,6 +523,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "image-policy",         required_argument, NULL, ARG_IMAGE_POLICY         },
                 { "header",               no_argument,       NULL, ARG_HEADER               },
                 { "identifier",           required_argument, NULL, 't'                      },
+                { "exclude-identifier",   required_argument, NULL, 'T'                      },
                 { "priority",             required_argument, NULL, 'p'                      },
                 { "facility",             required_argument, NULL, ARG_FACILITY             },
                 { "grep",                 required_argument, NULL, 'g'                      },
@@ -564,7 +569,7 @@ static int parse_argv(int argc, char *argv[]) {
         assert(argc >= 0);
         assert(argv);
 
-        while ((c = getopt_long(argc, argv, "hefo:aln::qmb::kD:p:g:c:S:U:t:u:NF:xrM:i:", options, NULL)) >= 0)
+        while ((c = getopt_long(argc, argv, "hefo:aln::qmb::kD:p:g:c:S:U:t:T:u:NF:xrM:i:", options, NULL)) >= 0)
 
                 switch (c) {
 
@@ -954,6 +959,12 @@ static int parse_argv(int argc, char *argv[]) {
 
                 case 't':
                         r = strv_extend(&arg_syslog_identifier, optarg);
+                        if (r < 0)
+                                return log_oom();
+                        break;
+
+                case 'T':
+                        r = strv_extend(&arg_exclude_identifier, optarg);
                         if (r < 0)
                                 return log_oom();
                         break;
@@ -1570,6 +1581,19 @@ static int add_syslog_identifier(sd_journal *j) {
                 return r;
 
         return 0;
+}
+
+static int add_exclude_identifier(sd_journal *j) {
+        _cleanup_set_free_ Set *excludes = NULL;
+        int r;
+
+        assert(j);
+
+        r = set_put_strdupv(&excludes, arg_exclude_identifier);
+        if (r < 0)
+                return r;
+
+        return set_free_and_replace(j->exclude_syslog_identifiers, excludes);
 }
 
 #if HAVE_GCRYPT
@@ -2407,6 +2431,10 @@ static int run(int argc, char *argv[]) {
         r = add_syslog_identifier(j);
         if (r < 0)
                 return log_error_errno(r, "Failed to add filter for syslog identifiers: %m");
+
+        r = add_exclude_identifier(j);
+        if (r < 0)
+                return log_error_errno(r, "Failed to add exclude filter for syslog identifiers: %m");
 
         r = add_priorities(j);
         if (r < 0)
--- a/src/libsystemd/sd-journal/journal-internal.h
+++ b/src/libsystemd/sd-journal/journal-internal.h
@@ -87,6 +87,7 @@ struct sd_journal {
         uint64_t current_field;
 
         Match *level0, *level1, *level2;
+        Set *exclude_syslog_identifiers;
 
         uint64_t origin_id;
 
--- a/src/libsystemd/sd-journal/sd-journal.c
+++ b/src/libsystemd/sd-journal/sd-journal.c
@@ -2327,6 +2327,8 @@ _public_ void sd_journal_close(sd_journal *j) {
 
         hashmap_free_free(j->errors);
 
+        set_free(j->exclude_syslog_identifiers);
+
         free(j->path);
         free(j->prefix);
         free(j->namespace);
--- a/src/shared/logs-show.c
+++ b/src/shared/logs-show.c
@@ -523,6 +523,9 @@ static int output_short(
                 return 0;
         }
 
+        if (identifier && set_contains(j->exclude_syslog_identifiers, identifier))
+                return 0;
+
         if (!(flags & OUTPUT_SHOW_ALL))
                 strip_tab_ansi(&message, &message_len, highlight_shifted);
 
--- a/test/testsuite-04.units/verbose-success.service
+++ b/test/testsuite-04.units/verbose-success.service
@@ -1,0 +1,6 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=Verbose successful service
+
+[Service]
+ExecStart=/bin/echo success
--- a/test/units/testsuite-04.journal.sh
+++ b/test/units/testsuite-04.journal.sh
@@ -107,6 +107,15 @@ systemctl start silent-success
 journalctl --sync
 [[ -z "$(journalctl -b -q -u silent-success.service)" ]]
 
+# Test syslog identifiers exclusion
+systemctl start verbose-success
+journalctl --sync
+[[ -n "$(journalctl -b -q -u verbose-success.service -t systemd)" ]]
+[[ -n "$(journalctl -b -q -u verbose-success.service -t echo)" ]]
+[[ -n "$(journalctl -b -q -u verbose-success.service -T systemd)" ]]
+[[ -n "$(journalctl -b -q -u verbose-success.service -T echo)" ]]
+[[ -z "$(journalctl -b -q -u verbose-success.service -T echo -T '(echo)' -T systemd -T '(systemd)' -T systemd-executor)" ]]
+
 # Exercise the matching machinery
 SYSTEMD_LOG_LEVEL=debug journalctl -b -n 1 /dev/null /dev/zero /dev/null /dev/null /dev/null
 journalctl -b -n 1 /bin/true /bin/false


src/basic/string-util.c
```diff
@@ -583,6 +583,24 @@ static void advance_offsets(
...
+        char *p = *ibuf;
+        size_t len = 0;
+
+        while (*p & ~'\n') {
+                len++;
+                p++;
+        }
+
+        *p = '\0';
+        *_isz = len;
```
pyih: you can exclude "len" variable at all and calculate "*_isz" so:
*_isz = p - ibuf;

zhmylove: Thank you. Fixed in 32325564557be491e7b20800d19d66df53bb970e

src/shared/logs-show.c
```diff
@@ -491,9 +491,17 @@ static int output_short(
...
         }
 
+        if (identifier && set_contains(j->exclude, identifier)) {
+                return 0;
+        }
+
         if (!(flags & OUTPUT_SHOW_ALL))
                 strip_tab_ansi(&message, &message_len, highlight_shifted);
 
+        if (flags & OUTPUT_STRIP) {
```
bluca: no braces around single-line blocks

src/journal/journalctl.c
```diff
@@ -1804,6 +1820,22 @@ static int add_syslog_identifier(sd_journal *j) {
         return 0;
 }
 
+static int add_exclude_identifier(sd_journal *j) {
+        int r;
+
+        assert(j);
+        j->exclude = set_new(&string_hash_ops);
+        assert(j->exclude);
```
bluca: `return -ENOMEM`

src/journal/journalctl.c
```diff
@@ -117,6 +118,7 @@ static bool arg_force = false;
 static usec_t arg_since, arg_until;
 static bool arg_since_set = false, arg_until_set = false;
 static char **arg_syslog_identifier = NULL;
+static char **arg_exclude_identifier = NULL;
```
bluca, line 125: no destructor was added for this

src/basic/string-util.c
```diff
@@ -583,6 +583,22 @@ static void advance_offsets(
                 shift[1] += size;
 }
 
+char *strip_newline(char **ibuf, size_t *_isz) {
```
bluca: pretty sure we already have helpers to strip strings of whitespace

.gitignore
```diff
@@ -9,6 +9,7 @@
 .deps/
 .mypy_cache/
 __pycache__/
+cscope*
```
bluca: what is this?

src/journal/journalctl.c
```diff
@@ -1804,6 +1820,22 @@ static int add_syslog_identifier(sd_journal *j) {
         return 0;
 }
 
+static int add_exclude_identifier(sd_journal *j) {
+        int r;
+
+        assert(j);
+        j->exclude = set_new(&string_hash_ops);
```
bluca: either assert that it is null, or free it first, for cleanliness

bluca (changes requested):
manpage update is missing. Also would be good to have a test/unit test for this, there should be already tests for the journal. Finally, rebase and squash fixup commits, so that history is clean. Would be good to have the two new options added as separate commits, since they are not related.

src/journal/journalctl.c
zhmylove: Isn't it the destructor:
https://github.com/systemd/systemd/blob/f5c756a254819a94173f04cae0028868fe1339f0/src/journal/journalctl.c#L2863

src/basic/string-util.c
zhmylove: We do have `truncate_nl()`, but it has 2 major disadvantages: 

1. it uses non-performant `strcspn()` (in most libc implementations it'll build a table of "reject" characters and this will take much more instructions to process)
2. it is unable to return truncated string length, so it requires an additional strlen() call to set `message_len`

We also have `string_truncate_lines()` which I think is not suitable as it makes the copy of the string.

From my perspective it's too much of work if we gonna to call this for each message we're going to show. What do you think?

In case `strcspn()` will be considered quite performant for this and `--strip` should react not only on `'\n'`, but also on `'\r'`, then I suggest:
```c
char *truncate_nl_len(char *s, size_t *len) {
        assert(s);
        assert(len);

        size_t idx = strcspn(s, NEWLINE);
        s[idx] = 0;
        *len = idx;
        return s;
}
```

.gitignore
zhmylove: Just like there is an exclusion for `vim` standard `*.swp` files above, this is an exclusion for `vim` standard `cscope` files.

bluca: it's not related to this PR, so at the very least add a separate commit for it

src/basic/string-util.c
bluca: I have no opinion w.r.t. performances, I'll leave that to @poettering and @DaanDeMeyer

zhmylove:
> Would be good to have the two new options added as separate commits, since they are not related.
@bluca should I send them as different PRs or just 2 separate commits under one PR?

bluca:
> @bluca many thanks for your review!
> 
> > Would be good to have the two new options added as separate commits, since they are not related.
> 
> Should I send them as different PRs or just 2 separate commits under one PR?
> 
> I'll work on man/tests later on.

same PR is fine by me

zhmylove:
@bluca I've re-ordered changes into 3 separate commits and added manpage update.
Still looking for a way to add unit tests. So far I can't find any of them for journalctl.

src/journal/journalctl.c
bluca: yes, missed that

bluca:
> @bluca I've re-ordered changes into 3 separate commits and added manpage update. Still looking for a way to add unit tests. So far I can't find any of them for journalctl.

There are tests for the journal in `test/units/testsuite-04.sh`

zhmylove:
@bluca thank you for pointing me out on them. 

I've added unit test for `--exclude` and I belive we're done with this part. 

I'd like to discuss with you the test for `--strip`. Usually we use `systemd-cat` to save log messages into the log. Unfortunately it's not able of adding multiline messages as it splits stdin by newlines. 
We can't rely on `systemd-journal-remote` as `HAVE_MICROHTTPD` could be `false`, like on my system.
Neither can we rely on `logger(1)` which could be missing on several systems. For example, on Debian-based it's part of `bsdutils` package, not the `util-linux` and I believe it's not installed by default.
What will be the best way to create a multiline message in journal from the testsuite?
Thanks!

bluca:
> @bluca thank you for pointing me out on them.
> 
> I've added unit test for `--exclude` and I belive we're done with this part.
> 
> I'd like to discuss with you the test for `--strip`. Usually we use `systemd-cat` to add log messages into the log. Unfortunately it's not able of adding multiline messages as it splits stdin by newlines:
> 
> ```
> # systemd-cat -t "$ID" sh -c 'echo line1 ;echo line2'
> # journalctl -q -b -t "$ID"
> Apr 16 11:41:57 rc1 q[29341]: line1
> Apr 16 11:41:57 rc1 q[29341]: line2
> ```
> 
> We can't rely on `systemd-journal-remote` as `HAVE_MICROHTTPD` could be `false`, like on my system. Neither can we rely on `logger(1)` which could be missing on several systems. For example, on Debian-based it's part of `bsdutils` package, not the `util-linux`, and I believe it's not installed by default. What will be the best way to create a multiline message in journal from the testsuite?
> 
> As `python >= 3.5` is build-requirement for systemd we can try to utilize it like:
> 
> ```
> # python -c 'import syslog; syslog.openlog("qqq"); syslog.syslog("line1\nline2")'
> # journalctl -q -b -t qqq
> Apr 16 12:04:34 rc1 qqq[29513]: line1
>                                 line2
> ```
> 
> But I'm not sure it'll be available inside the testsuite. Should we add `python` to `BASICTOOLS` of `test/test-functions`?
> 
> Thanks!

Use `logger` and simply check if it's available with `command -v`, and skip the test if not - I'll make sure it's installed in the CI

zhmylove:
@bluca added test for `--strip`. Thank you!

bluca:
> @bluca added test for `--strip`. Thank you!

You need to add a `test_append_files` in `test/TEST-04-JOURNAL/test.sh` that installs logger if it's available - see `test/TEST-43-PRIVATEUSER-UNPRIV/test.sh` as an example

zhmylove:
@bluca this time included. Thanks!

bluca:
> @bluca this time included. Thanks!

The test is now failing on s390x, you can fetch the logs from https://autopkgtest.ubuntu.com/results/autopkgtest-focal-upstream-systemd-ci-systemd-ci/focal/s390x/s/systemd-upstream/20220416_143415_22e40@/artifacts.tar.gz

bluca:
here too: https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux-sanitizers/11170/console

src/basic/string-util.h
```diff
@@ -105,6 +105,8 @@ char *strstrip(char *s);
 char *delete_chars(char *s, const char *bad);
 char *delete_trailing_chars(char *s, const char *bad);
 char *truncate_nl(char *s);
+char *truncate_nl_len(char **ibuf, size_t *_isz);
+
```
yuwata: Unnecessary blank line.

src/basic/string-util.c
```diff
@@ -176,6 +176,19 @@ char *truncate_nl(char *s) {
         return s;
 }
 
+char *truncate_nl_len(char **ibuf, size_t *_isz) {
```
yuwata: Please extend `truncate_nl()`. How about the following?
```
char *truncate_nl_full(char *s, size_t *ret_len) {
        size_t n;

        assert(s);

        n = strcspn(s, NEWLINE);
        s[n] = '\0';
        if (ret_len)
                *ret_len = n;
        return s;
}
```
Then, in string-util.h
```
char *truncate_nl_full(char *s, size_t *ret_len);
static inline char *truncate_nl(char *s) {
        return truncate_nl_full(s, NULL);
}
```

src/journal/journalctl.c
```diff
@@ -1804,6 +1820,24 @@ static int add_syslog_identifier(sd_journal *j) {
         return 0;
 }
 
+static int add_exclude_identifier(sd_journal *j) {
+        int r;
+
+        assert(j);
+        set_free(j->exclude);
```
yuwata: Please insert a blank line after assertion(s).

Also, clear `j->exclude` explicitly, that is, `j->exclude = set_free(j->exclude);`

src/journal/journalctl.c
```diff
@@ -1804,6 +1820,24 @@ static int add_syslog_identifier(sd_journal *j) {
...
+        j->exclude = set_new(&string_hash_ops);
+        if (!j->exclude)
+                return -ENOMEM;
```
yuwata: please drop them, and use `set_ensure_put()` in the below.

src/journal/journalctl.c
```diff
@@ -1804,6 +1820,24 @@ static int add_syslog_identifier(sd_journal *j) {
...
+
+        STRV_FOREACH(i, arg_exclude_identifier) {
+                r = set_put(j->exclude, (const void *)*i);
```
yuwata: set_put() -> set_ensure_put().

Also, the cast should not be necessary.

yuwata (changes requested):

src/basic/string-util.c
zhmylove: @yuwata thank you for the review! 
From my perspective name this function with prefix "_full" is not a completely good idea as in the future someone would probably want to override it as well. I suggest using "truncate_nl_len" (like in my comment above) but I like your implementation. I'll commit new version soon.

yuwata: The naming is not a matter, hence `truncate_nl_len()` is ok.

bluca: it's common convention to call functions that get shortened by other helpers `_full`

src/basic/string-util.c
zhmylove: Below we agreed to use `strcspn()`, so implemented in a better way.

zhmylove:
@bluca @yuwata thank you very much for the comments!
Do you know, if there is any way to get more logs from `focal-s390x` autopkgtest? It's failing most of runs, although once it succeeded, I see no correlation with changes, and there is virtually nothing interesting inside `log.gz` available via the `Details` link.
Otherwise I have to use different approach: send commits one-by-one and see which part is failing. But it'll take much more time.

man/journalctl.xml
```diff
@@ -408,6 +408,14 @@
         printed.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>-s</option></term>
+        <term><option>--strip</option></term>
+
+        <listitem><para>Strip each message so that only first line is
```
behrmann: I think there's an article missing here
```suggestion
        <listitem><para>Strip each message so that only the first line is
```

zhmylove: Thank you!

zhmylove:
@bluca @yuwata many thanks for your time and help! Now all your recommendations are fulfilled and CI checks are completely passed.

src/journal/journalctl.c
```diff
@@ -1804,6 +1820,22 @@ static int add_syslog_identifier(sd_journal *j) {
...
 
+static int add_exclude_identifier(sd_journal *j) {
+        int r;
+
+        assert(j);
+
+        j->exclude = set_free(j->exclude);
+
+        STRV_FOREACH(i, arg_exclude_identifier) {
+                r = set_ensure_put(&j->exclude, &string_hash_ops, *i);
```
yuwata: I noticed that `j->exclude` references strings owned by `arg_exclude_identifier`.

As you can see, `arg_exclude_identifier` is freed in main() *before* `sd_journal *j` is freed.

Please use `set_put_strdup()`.

test/TEST-04-JOURNAL/test.sh
```diff
@@ -7,4 +7,10 @@ TEST_DESCRIPTION="Journal-related tests"
 # shellcheck source=test/test-functions
 . "${TEST_BASE_DIR:?}/test-functions"
 
+test_append_files() {
+    (
```
yuwata: Why `()` is necessary here?

src/shared/logs-show.c
```diff
@@ -491,9 +491,15 @@ static int output_short(
                 return 0;
         }
 
+        if (identifier && set_contains(j->exclude, identifier))
```
yuwata: Is it not necessary to applied to other output functions, e.g. `output_json()`?

src/shared/logs-show.c
```diff
@@ -491,9 +491,15 @@ static int output_short(
...
+                return 0;
+
         if (!(flags & OUTPUT_SHOW_ALL))
                 strip_tab_ansi(&message, &message_len, highlight_shifted);
 
+        if (flags & OUTPUT_STRIP)
```
yuwata: Also this one. Is it not necessary to apply this to other methods?

yuwata (changes requested):
BTW, I think it is better to introduce `sd_journal_add_invert_match()` or so, instead of filtering in the output functions.

src/shared/logs-show.c
zhmylove: The full list of output functions are: 
```
output_cat     -- this feature not needed if "cat"
output_export  -- this feature not needed if "export"
output_json    -- typically json output is used for parsing/scripting, so I believe it's not needed for this as well
output_short   -- this feature covers all the variants of output_short
output_verbose -- this feature not needed if "verbose"
```

src/shared/logs-show.c
zhmylove: Same as above

test/TEST-04-JOURNAL/test.sh
zhmylove: I'm not so familiar with systemd testsuite but 14 implementations of `test_append_files` use subshell, 4 do not, so I decided to make like the majority.

zhmylove:
> BTW, I think it is better to introduce `sd_journal_add_invert_match()` or so, instead of filtering in the output functions.

There is an abandoned PR from 2019 with similar functionality, but from my perspective it's a little bit complex. To improve speed, `journalctl` utilizes several levels of `Match*` and gets the next journal entry when any of the fields has the same `jenkins_hash64()`. We can not simply negate such an approach: we still be iterating through all the hashes, moreover: even on positive match, we still have to iterate through all the fields to ensure this entry is not _excluded_ -- this is a performance degradation.

src/journal/journalctl.c
zhmylove: It's meaningless:
1. `set_put_strdup()` (here more applicable `set_put_strdup_full()` as we want specify hash_ops for strings is:
```
130 int _set_put_strdup_full(Set **s, const struct hash_ops *hash_ops, const char *p  HASHMAP_DEBUG_PARAMS);
131 #define set_put_strdup_full(s, hash_ops, p) _set_put_strdup_full(s, hash_ops, p  HASHMAP_DEBUG_SRC_ARGS)
132 #define set_put_strdup(s, p) set_put_strdup_full(s, &string_hash_ops_free, p)
```
2. `_set_put_strdup_full()` is:
```
1844 int _set_put_strdup_full(Set **s, const struct hash_ops *hash_ops, const char *p  HASHMAP_DEBUG_PARAMS) {
...
1851         r = _set_ensure_allocated(s, hash_ops  HASHMAP_DEBUG_PASS_ARGS);
...
1858         c = strdup(p);
...
1862         return set_consume(*s, c);
```
3. `set_consume()` is:
```
1797 int set_consume(Set *s, void *value) {
1798         int r;
1799
1800         assert(s);
1801         assert(value);
1802
1803         r = set_put(s, value);
1804         if (r <= 0)
1805                 free(value);
1806
1807         return r;
1808 }
```
So it calls `free()` of a value after calling `set_put()`

In this PR `free()` is called by `strv_free()`, while `set_put()` is called directly via `set_ensure_put()`

zhmylove: Thanks! Done

test/TEST-04-JOURNAL/test.sh
yuwata: OK. Thanks.

.gitignore
daandemeyer: Tip: you can edit .git/info/exclude which acts as a local gitignore file.

man/journalctl.xml
```diff
@@ -537,6 +537,18 @@
...
+      <varlistentry>
+        <term><option>-T</option></term>
+        <term><option>--exclude=<replaceable>SYSLOG_IDENTIFIER</replaceable></option></term>
+
+        <listitem><para>Exclude messages for the specified syslog
+        identifier
+        <replaceable>SYSLOG_IDENTIFIER</replaceable>.</para>
+
+        <para>This parameter can be specified multiple
+        times.</para></listitem>
```
daandemeyer: Our line length is set to 109 so I think there's too many newlines here

DaanDeMeyer (commented):
I think we should give more thought to adding support for inverse matches which would solve this in a more generic way and for all fields instead of only for one specific field.

If you look at `MatchType` in journal-internal.h, we could add a new type `MATCH_DISCRETE_INVERT` or so and then modify `next_for_match()` and `find_location_for_match()` to accommodate for the new type.

Making those changes won't be entirely trivial but it doesn't seem impossible either. The key difference would likely be that instead of bisecting the data object's entry array to find the next entry object, we'd search the global entry array for the next entry object that we can't find in the data object's entry array. 

The only issue is that this isn't exactly an optimized access pattern since we don't store any information that helps us find the first object that doesn't store a particular field. The existing Data objects make it easy enough to check if a particular entry contains a field or not, but what we can't do easily is find the next object that doesn't have a particular field.

Best I can think of is iterating the Data object's entry array and the global entry array simultaneously until you find an object offset that's in the global array but not in the data object entry array..

CLI interface could be something like `journalctl !_SYSTEMD_UNIT=systemd-networkd.service`. A shorthand option such as the proposed --exclude could then be implemented on top of this.

mbiebl:
@bluca 
I notice that you added a `bsdutils` test dependency to the upstream test suite.

Is there a specific reason for that given that the `bsdutils` package is Essential and thus guaranteed to be installed on Debian/Ubuntu?

bluca:
> @bluca I notice that you added a `bsdutils` test dependency to the upstream test suite.
> 
> Is there a specific reason for that given that the `bsdutils` package is Essential and thus guaranteed to be installed on Debian/Ubuntu?

I don't remember to be honest, I think it was requested?

man/journalctl.xml
```diff
@@ -408,6 +408,14 @@
         printed.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>-s</option></term>
+        <term><option>--strip</option></term>
+
+        <listitem><para>Strip each message so that only the first line is
```
poettering: hmm, "strip" sounds like a weird name for this. Maybe call thus "--truncate-newline"?

poettering:
So, I sympathize with both patches, but I agree with @DaanDeMeyer on the negation thing: we should support inverted matches generically in the journal match logic. It's certainly more involved but if we have it there we can optimize it much nicer and allow it for all kinds of matches. 

I am not a fan of just naming this --exclude=, since it's not clear it's about the syslog tag field. Moreover, I am pretty sure different people want to exclude different stuff, for example, some people might want to invert the effect of --unit= and similar. Thus if we introduce negated matches, we probably should give it a naming we can systematically extend to --identifier=, --unit=, --user-unit= and so on. i.e. we could just say that --exclude-identifier=, --exclude-unit=, --exclude-user-unit= would be the systematic way we'd name these options.

The --strip thing sounds more like something we could merge quickly. But I don't like the name "strip". truncating evertyhing after the first newline is a lot more than just "stripping". Hence, maybe rename the switch --truncate-newline? 

Given the two patches are kinda unrelated, they should probably be split into two PRs, of which the --truncate-newline one we can probably merge much more quickly than the --exclude-xyz= one

poettering:
So I see now that the --strip thing already has been split out, in #23473... so ignore my requests asking for such a split-out

poettering:
please rebase the rest on current main

zhmylove:
Sure, I'll do it soon

zhmylove:
@poettering I've rebased the rest on current main. Thank you!

zhmylove:
@poettering I've reviewed logs for 2x failing checks and I doubt they're relative to this PR.
* jammy-ppc64le fails in test of TPM2 as it looks like there is no TPM module for that platform.
* jammy-s390x fails on creating nova instance.

zhmylove:
@poettering I've rebased on current main and now other tests are failing. Looks not relevant again

zhmylove:
@mrc0mmand hello, it looks like CentOS CI checks are failing due to some not code-related issue (see logs below).
According to empty `.failed_tasks` and non-empty `.passed_tasks` files, all the tests are actually passing well, right?

```
23:17:03 2023-12-11 20:17:03,747 [agent-control/fetch_artifacts] INFO: Fetching artifacts from node n27-42-29.pool.ci.centos.org: (remote: /var/crash, local: ./artifacts_q9x7o23r/kdumps)
23:17:03 2023-12-11 20:17:03,747 [agent-control/execute_local_command] INFO: Executing a LOCAL command: /usr/bin/scp -r -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@n27-42-29.pool.ci.centos.org:/var/crash ./artifacts_q9x7o23r/kdumps
23:17:03 2023-12-11 20:17:03,815 [agent-control/main] ERROR: Execution failed
23:17:03 Traceback (most recent call last):
23:17:03   File "/tmp/workspace/upstream-centos8-8302/systemd-centos-ci/./agent-control.py", line 508, in main
23:17:03     ac.execute_remote_command(command, artifacts_dir="~/bootstrap-logs*")
23:17:03   File "/tmp/workspace/upstream-centos8-8302/systemd-centos-ci/./agent-control.py", line 173, in execute_remote_command
23:17:03     raise RuntimeError("Remote command exited with an unexpected return code "
23:17:03 RuntimeError: Remote command exited with an unexpected return code (got: 1, expected: 0), bailing out
```

mrc0mmand:
> @mrc0mmand hello, it looks like CentOS CI checks are failing due to some not code-related issue (see logs below). According to empty `.failed_tasks` and non-empty `.passed_tasks` files, all the tests are actually passing well, right?
> 
> ```
> 23:17:03 2023-12-11 20:17:03,747 [agent-control/fetch_artifacts] INFO: Fetching artifacts from node n27-42-29.pool.ci.centos.org: (remote: /var/crash, local: ./artifacts_q9x7o23r/kdumps)
> 23:17:03 2023-12-11 20:17:03,747 [agent-control/execute_local_command] INFO: Executing a LOCAL command: /usr/bin/scp -r -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@n27-42-29.pool.ci.centos.org:/var/crash ./artifacts_q9x7o23r/kdumps
> 23:17:03 2023-12-11 20:17:03,815 [agent-control/main] ERROR: Execution failed
> 23:17:03 Traceback (most recent call last):
> 23:17:03   File "/tmp/workspace/upstream-centos8-8302/systemd-centos-ci/./agent-control.py", line 508, in main
> 23:17:03     ac.execute_remote_command(command, artifacts_dir="~/bootstrap-logs*")
> 23:17:03   File "/tmp/workspace/upstream-centos8-8302/systemd-centos-ci/./agent-control.py", line 173, in execute_remote_command
> 23:17:03     raise RuntimeError("Remote command exited with an unexpected return code "
> 23:17:03 RuntimeError: Remote command exited with an unexpected return code (got: 1, expected: 0), bailing out
> ```
> 
> It'd be nice if you point out on some mistakes I did and help me fixing them. Thank you!

The C8S fail is unrelated (borked kernel, already working on a workaround). As for the Arch Linux job, there's an actual issue:

```
[  276.808394] testsuite-04.sh[12216]: ++ journalctl -b -q -u verbose-success.service -T echo
[  276.814310] testsuite-04.sh[11695]: + [[ -n Dec 11 20:12:02 H systemd[1]: verbose-success.service: Trying to enqueue job verbose-success.service/start/replace
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Installed new job verbose-success.service/start as 2018
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Enqueued job verbose-success.service/start as 2018
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Will spawn child (service_enter_start): /bin/echo
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Passing 0 fds to service
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: About to execute: /bin/echo success
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Forked /bin/echo as 12209
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Changed dead -> running
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Job 2018 verbose-success.service/start finished, result=done
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: Started verbose-success.service.
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd-executor[12209]: SELinux enabled state cached to: disabled
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H (echo)[12209]: Found cgroup2 on /sys/fs/cgroup/, full unified hierarchy
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H (echo)[12209]: Found cgroup2 on /sys/fs/cgroup/, full unified hierarchy
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Child 12209 belongs to verbose-success.service.
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Main process exited, code=exited, status=0/SUCCESS (success)
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Deactivated successfully.
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Service will not restart (restart setting)
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Changed running -> dead
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Consumed 8ms CPU time, 1.5M memory peak, 0B memory swap peak.
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Releasing resources...
[  276.814310] testsuite-04.sh[11695]: Dec 11 20:12:02 H systemd[1]: verbose-success.service: Collecting. ]]
[  276.814960] testsuite-04.sh[12217]: ++ journalctl -b -q -u verbose-success.service -T echo -T systemd
[  276.822148] testsuite-04.sh[11695]: + [[ -z Dec 11 20:12:02 H systemd-executor[12209]: SELinux enabled state cached to: disabled
[  276.822148] testsuite-04.sh[11695]: Dec 11 20:12:02 H (echo)[12209]: Found cgroup2 on /sys/fs/cgroup/, full unified hierarchy
[  276.822148] testsuite-04.sh[11695]: Dec 11 20:12:02 H (echo)[12209]: Found cgroup2 on /sys/fs/cgroup/, full unified hierarchy ]]
[  276.822675] testsuite-04.sh[382]: + echo 'Subtest /usr/lib/systemd/tests/testdata/units/testsuite-04.journal.sh failed'
[  276.822675] testsuite-04.sh[382]: Subtest /usr/lib/systemd/tests/testdata/units/testsuite-04.journal.sh failed
```

mrc0mmand:
> @mrc0mmand thank you SO much for your instant reply!
> 
> Should I upload a fix for the Arch Linux job as soon as I ready or should I wait for your workaround on C8S in order to avoid double-running the pipeline? In latter case may I ask you to notify me when the workaround is there, please?

Feel free to push the fix once you have it ready; don't worry about the pipeline, the CIs automagically cancel previous running jobs each time you (force-) push to the PR.

zhmylove:
Hello @poettering, as you've asked I rebased it on current main.

There are 2 jammy checks (ppc64 & s390x) that seems stuck already for 3 days.
All the other checks have passed successfully, thanks to mrc0mmand.

Everything looks good and PR is waiting for it's merge.

zhmylove:
@bluca @DaanDeMeyer @poettering @yuwata hello.
Is there any chance you can take a look on this? I've already rebased this PR 3x times, each time it took some efforts. It would be more effective if I rebased it only once: when you ready to give it attention.

People do require this functionality for many years. There were at least 2 PRs delivering similar stuff, already buried in time.

Thank you!

zhmylove:
@bluca @DaanDeMeyer @poettering @yuwata Merry Christmas and happy New Year. Any updates on this?

man/journalctl.xml
```diff
@@ -362,6 +362,18 @@
         <xi:include href="version-info.xml" xpointer="v217"/></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>-T</option></term>
+        <term><option>--exclude=<replaceable>SYSLOG_IDENTIFIER</replaceable></option></term>
```
bluca: The long option should be --exclude-identifier or so as --exclude is too generic given this only applies to this single field

src/journal/journalctl.c
```diff
@@ -2408,6 +2434,11 @@ static int run(int argc, char *argv[]) {
         if (r < 0)
                 return log_error_errno(r, "Failed to add filter for syslog identifiers: %m");
 
+        r = add_exclude_identifier(j);
+        if (r < 0) {
```
bluca: no brackets around single line if blocks

bluca (changes requested):
Please also add bash completion support for the flag to shell-completion/bash/journalctl

YHNdnzj:
> I am not a fan of just naming this --exclude=, since it's not clear it's about the syslog tag field. Moreover, I am pretty sure different people want to exclude different stuff, for example, some people might want to invert the effect of --unit= and similar. Thus if we introduce negated matches, we probably should give it a naming we can systematically extend to --identifier=, --unit=, --user-unit= and so on. i.e. we could just say that --exclude-identifier=, --exclude-unit=, --exclude-user-unit= would be the systematic way we'd name these options.

Hmm, this is not yet resolved? Also, how does this relate to https://github.com/systemd/systemd/pull/12592 ?

bluca (approved):

man/journalctl.xml
zhmylove: Thank you! Done

src/journal/journalctl.c
zhmylove: Thank you! Done

shell-completion/bash/journalctl
```diff
@@ -46,9 +46,9 @@ _journalctl() {
                       --show-cursor --dmesg -k --pager-end -e -r --reverse
                       --utc -x --catalog --no-full --force --dump-catalog
                       --flush --rotate --sync --no-hostname -N --fields'
-        [ARG]='-b --boot -D --directory --file -F --field -t --identifier --facility
-                      -M --machine -o --output -u --unit --user-unit -p --priority
-                      --root --case-sensitive'
+        [ARG]='-b --boot -D --directory --file -F --field -t --identifier
+                      --exclude-identifier --facility -M --machine -o --output
```
yuwata: Missing `-T`

shell-completion/bash/journalctl
```diff
@@ -110,6 +110,9 @@ _journalctl() {
             --identifier|-t)
                 comps=$(journalctl -F 'SYSLOG_IDENTIFIER' 2>/dev/null)
                 ;;
+            --exclude-identifier|-T)
+                comps=$(journalctl -F 'SYSLOG_IDENTIFIER' 2>/dev/null)
```
yuwata, line 111: Not a matter, but can be merged with the above case `--identifier|-t`.

src/journal/journalctl.c
```diff
@@ -380,6 +382,7 @@ static int help(void) {
                "  -u --unit=UNIT             Show logs from the specified unit\n"
                "     --user-unit=UNIT        Show logs from the specified user unit\n"
                "  -t --identifier=STRING     Show entries with the specified syslog identifier\n"
+               "  -T --exclude-identifier=STRING Hide entries with the specified identifier\n"
```
yuwata: Please break the line if the string does not fit the current alignment.

src/journal/journalctl.c
```diff
@@ -1572,6 +1582,22 @@ static int add_syslog_identifier(sd_journal *j) {
...
 
+static int add_exclude_identifier(sd_journal *j) {
+        int r;
+
+        assert(j);
+
+        j->exclude = set_free_free(j->exclude);
+
+        STRV_FOREACH(i, arg_exclude_identifier) {
+                r = set_put_strdup_full(&j->exclude, &string_hash_ops, *i);
```
yuwata: Why `string_has_ops` is used instead of `string_has_ops_free`??

Please use `set_put_strdupv()`, and use `set_free()` in the above.

src/libsystemd/sd-journal/sd-journal.c
```diff
@@ -2327,6 +2327,8 @@ _public_ void sd_journal_close(sd_journal *j) {
 
         hashmap_free_free(j->errors);
 
+        set_free_free(j->exclude);
```
yuwata: Please use `set_free()`.

src/shared/logs-show.c
```diff
@@ -523,6 +523,11 @@ static int output_short(
                 return 0;
         }
 
+        if (identifier && set_contains(j->exclude, identifier)) {
+                log_debug("Skipping manually excluded message.");
```
yuwata: Please drop the debugging log, otherwise, the functionality of the feature is just replacing matching journal entry.

yuwata (changes requested):

src/journal/journalctl.c
zhmylove: Thanks! What is the convenient way to break the line? I didn't find any long options as a reference not fitting the line, so I'm not sure about break position and indentation.

mrc0mmand: You can check `systemctl`'s help output and look for the `--check-inhibitors=` option.

src/shared/logs-show.c
zhmylove: Done

src/libsystemd/sd-journal/sd-journal.c
zhmylove: Done

src/journal/journalctl.c
zhmylove: Done

src/journal/journalctl.c
zhmylove: Thanks a lot, done!

shell-completion/bash/journalctl
zhmylove: Anyways, merged. Thanks

shell-completion/bash/journalctl
zhmylove: Done

zhmylove:
@YHNdnzj 

> Hmm, this is not yet resolved? 

Already resolved.

> Also, how does this relate to https://github.com/systemd/systemd/pull/12592 ?

The idea looks similar, people do really want this functionality for a long time.
But that PR is probably abandoned by the author for last three years.
There were several of them, I think.

@bluca Thanks a lot for your approval!

@yuwata Thank you for the comments, everything is fixed now.

src/journal/journalctl.c
```diff
@@ -1572,6 +1583,22 @@ static int add_syslog_identifier(sd_journal *j) {
...
+
+        assert(j);
+
+        j->exclude = set_free(j->exclude);
+
+        STRV_FOREACH(i, arg_exclude_identifier) {
+                r = set_put_strdupv(&j->exclude, i);
+                if (r < 0)
+                        return r;
+        }
```
yuwata: `set_put_strdupv()` takes strv, so it is not necessary to run over `arg_exclude_identifier`. Simply pass it.
Also, please create a new set, then replace it.
```c
_cleanup_set_free_ Set *excludes = NULL;
int r;

assert(j);

r = set_put_strdupv(&excludes, arg_exclude_identifier);
if (r < 0)
        return r;

return set_free_and_replace(j->excludes, excludes);
```

src/libsystemd/sd-journal/journal-internal.h
```diff
@@ -87,6 +87,7 @@ struct sd_journal {
         uint64_t current_field;
 
         Match *level0, *level1, *level2;
+        Set *exclude;
```
yuwata: I'd like to make the name more descriptive, e.g. `exclude_syslog_identifiers` or so.

yuwata (changes requested):
Currently, the new option only works with output type which uses `output_short()`.
I think such feature should be implemented in sd-journal side.
But, this is a good start point. Let's take the current way, and extend the feature later.

src/libsystemd/sd-journal/journal-internal.h
zhmylove: Refactoring done

src/journal/journalctl.c
zhmylove: Thanks, changed the code that way

yuwata (approved):
LGTM. Thank you, and sorry for late review process.

PR #23087, opened 2022-04-15
udev: resolve race in saving inotify watch handle

yuwata:


src/udev/udevd.c
```diff
@@ -1487,23 +1487,22 @@ static int synthesize_change(sd_device *dev) {
...
 
         l = read(fd, &buffer, sizeof(buffer));
         if (l < 0) {
                 if (ERRNO_IS_TRANSIENT(errno))
-                        return 1;
+                        return 0;
 
-                return log_error_errno(errno, "Failed to read inotify fd: %m");
+                log_warning_errno(errno, "Failed to read inotify fd: %m");
+                return 0;
```
poettering: hmm, why eat up this error? if we can't read from inotify fd we shouldn't ignore this, otherwise we'll busy-loop?

under which conditions would this return an error which is OK to ignore?

src/udev/udevd.c
```diff
@@ -1509,6 +1509,14 @@ static int on_inotify(sd_event_source *s, int fd, uint32_t revents, void *userda
                 _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
                 const char *devnode;
 
+                /* Do not handle IN_IGNORED here. The same watch handle may be already reused for
+                 * another device node. The old symlink will be removed by a worker which processes the
```
poettering: hmm, watch handles can be reused? i wasn't aware… under which conditions?

src/udev/udev-watch.c
```diff
@@ -10,16 +10,97 @@
...
+                _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
+                _cleanup_free_ char *buf = NULL;
+
+                if (dot_or_dot_dot(de->d_name))
+                        continue;
+
+                if (readlinkat_malloc(dirfd(dir), de->d_name, &buf) < 0)
+                        continue;
+
+                if (!streq(buf, wd_str))
```
poettering: i think it would be safer to parse the specified buffer via `safe_atoi()` and then do a numeric comparison wiht `ẁd`, instead of first formatting `wd` into `wd_str`.

src/udev/udev-watch.c
```diff
@@ -29,115 +110,116 @@ int udev_watch_restore(int inotify_fd) {
 
         if (rename("/run/udev/watch", "/run/udev/watch.old") < 0) {
                 if (errno != ENOENT)
-                        return log_warning_errno(errno, "Failed to move watches directory /run/udev/watch. "
+                        return log_warning_errno(errno,
+                                                 "Failed to move watches directory /run/udev/watch. "
```
poettering: we usually suffix paths referring to directories with a trailing `/` in log messages and documentation. Maybe do so here too, given you touch the line

src/udev/udev-watch.c
```diff
@@ -29,115 +110,116 @@ int udev_watch_restore(int inotify_fd) {
...
                 return 0;
         }
 
         dir = opendir("/run/udev/watch.old");
-        if (!dir)
-                return log_warning_errno(errno, "Failed to open old watches directory /run/udev/watch.old. "
-                                         "Old watches will not be restored: %m");
+        if (!dir) {
+                r = log_warning_errno(errno,
+                                      "Failed to open old watches directory /run/udev/watch.old. "
```
poettering: ditto

src/udev/udev-watch.c
```diff
@@ -10,16 +10,97 @@
...
+        if (wd < 0)
+                return -EBADF;
+
+        xsprintf(wd_str, "%d", wd);
+
+        dir = opendir("/run/udev/watch");
+        if (!dir)
+                return -errno;
+
+        FOREACH_DIRENT_ALL(de, dir, break) {
```
poettering: hmm, so this moves behaviour from O(1) to O(n)? Is that really a good idea?

poettering:
hmm, so i understand this is all supposed to address issues around inotify watch descriptor usage? is that really something that is easy to trigger?


wouldn't a nicer way be to track IN_IGNORED events for each wd?

i.e whenever we remove a wd from an inotify, put the wd in a "quarantine" Set object. Then, when we call read() on the next time on the inotify fd we ignore all events which have a wd currently registered in that `Set` object. Once we see an IN_IGNORED event we'll remove the wd from the set. And then as additional rule we'll flush out the Set object in full whenever we see IN_Q_OVERFLOW or whenever we attempt to read from the inotify fd and get EAGAIN back (i.e. nothing is queued anymore).

That would properly solve the issue, no? it would mean that wd *may* be reused, but if an event is already queued for the old wd we'll never mistake it for an event for the new wd.

or am I totally missing what the issue is?

src/udev/udevd.c
yuwata: Hm, indeed. `ERRNO_IS_TRANSIENT()` is already filtered out. Will drop the change.

src/udev/udevd.c
yuwata: Not sure, but possibly. I will rephrase the comment.

src/udev/udev-watch.c
```diff
@@ -73,71 +73,130 @@ int udev_watch_restore(int inotify_fd) {
...
+        /* 3. check if the symlink wd -> ID is owned by the device. */
+        if (!streq(buf, id)) {
+                r = log_device_debug_errno(dev, SYNTHETIC_ERRNO(ENOENT),
+                                           "Symlink '/run/udev/watch/%s' is owned by another device '%s'.", wd_str, buf);
+                goto finalize;
+        }
+
+        /* 4. remove symlink wd -> ID only when the symlink is owned by the device. */
+        if (unlinkat(dirfd, wd_str, 0) < 0 && errno != -ENOENT)
+                log_device_debug_errno(dev, errno, "Failed to remove '/run/udev/watch/%s', ignoring: %m", wd_str);
```
poettering, line 159: is it a problem that this is not atomic? i.e. the readlink/streq/unlinkat?

src/udev/udev-watch.c
```diff
@@ -73,71 +73,130 @@ int udev_watch_restore(int inotify_fd) {
...
+        r = 0;
+
+finalize:
+        if (unlinkat(dirfd, id, 0) < 0 && errno != -ENOENT)
+                log_device_debug_errno(dev, errno, "Failed to remove '/run/udev/watch/%s': %m", id);
+
+        if (r >= 0 && ret_wd)
+                *ret_wd = wd;
+        return r;
+}
```
poettering, line 172: the error path is weird. wouldn't this make more sense:

```c
        if (ret_wd)
                *ret_wd = wd;
        r = 0;

finish:
        if (unlinkat(…))
                 …
        return r;
}
```

i.e. move the `*ret_wd` initialization before the jump label, since you need that only in the success path...

src/udev/udev-watch.c
```diff
@@ -73,71 +73,130 @@ int udev_watch_restore(int inotify_fd) {
...
+        if (r < 0)
+                return log_device_debug_errno(dev, r, "Failed to get device ID: %m");
+
+        r = mkdir_p("/run/udev/watch", 0755);
+        if (r < 0)
+                return log_device_debug_errno(dev, r, "Failed to create '/run/udev/watch/': %m");
+
+        dirfd = open("/run/udev/watch", O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
+        if (dirfd < 0)
+                return log_device_debug_errno(dev, errno, "Failed to open '/run/udev/watch/': %m");
```
poettering: maybe use `open_mkdir_at()` to combine these two things?

src/udev/udev-watch.c
```diff
@@ -28,49 +28,54 @@ int udev_watch_restore(int inotify_fd) {
...
+                        continue;
                 }
 
-                log_device_debug(dev, "Restoring old watch");
                 (void) udev_watch_begin(inotify_fd, dev);
-unlink:
-                (void) unlinkat(dirfd(dir), ent->d_name, 0);
         }
 
         (void) closedir(dir);
```
poettering: might be the right time to switch to `_cleanup_closedir_`

src/udev/udev-watch.c
yuwata: The symlink is removed only when it is assigned to the currently processing device. Hence, if the above `readlink()` and `streq()` pass, then no other workers remove the symlink at the same time.

yuwata:
@poettering Thank you for your review and sorry for late reply. Now, all suggestions are addressed. PTAL.

src/udev/udev-watch.c
```diff
@@ -73,71 +73,128 @@ int udev_watch_restore(int inotify_fd) {
...
         r = sd_device_get_devname(dev, &devnode);
         if (r < 0)
-                return log_device_debug_errno(dev, r, "Failed to get device name: %m");
+                return log_device_debug_errno(dev, r, "Failed to get device node: %m");
 
+        r = device_get_device_id(dev, &id);
+        if (r < 0)
+                return log_device_debug_errno(dev, r, "Failed to get device ID: %m");
+
+        r = dirfd = open_mkdir_at(AT_FDCWD, "/run/udev/watch", O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_RDONLY, 0755);
```
poettering: O_DIRECTORY and O_NOFOLLOW are actually implied.

(ad thinking about it, we should really imply O_RDONLY, too)

src/udev/udev-watch.c
poettering: hmm, can you add this as a comment?

src/udev/udev-watch.c
```diff
@@ -20,57 +20,61 @@
...
 int udev_watch_restore(int inotify_fd) {
-        DIR *dir;
+        _cleanup_closedir_ DIR *dir = NULL;
         int r;
 
         /* Move any old watches directory out of the way, and then restore the watches. */
 
         assert(inotify_fd >= 0);
 
         if (rename("/run/udev/watch", "/run/udev/watch.old") < 0) {
```
poettering, line 55: hmm, for robustness reasons shouldn't this first try to `rm_rf("/run/udev/watch.old")` in case the dir still remains from a previous invocation of udev where we failed to remove the thing?

src/udev/udev-watch.c
yuwata: Sure. Done.

yuwata:
@poettering Thank you for the review. Addressed all comments in the above. Upgrading the green label.

PR #23088, opened 2022-04-15
udev: cleanups for event blocker

yuwata:


src/udev/udevd.c
```diff
@@ -854,9 +854,21 @@ static int event_run(Event *event) {
...
+
+        if (!a || !b)
+                return false;
+
+        for (; *a != '\0' && *b != '\0'; a++, b++)
+                if (*a != *b)
+                        return false;
+
+        return *a == '/' || *b == '/' || *a == *b;
+}
```
poettering: could use a test case I think. given this is not obvious

poettering: hmm, *a == *b only holds if both are NUL, right?

poettering:
PR looks good to me, but the wider implications are not entirely clear to me, hence: stuff for v251? or v252? what do you think?

src/udev/udevd.c
yuwata: > hmm, *a == *b only holds if both are NUL, right?

Right.

yuwata:
> PR looks good to me, but the wider implications are not entirely clear to me, hence: stuff for v251? or v252? what do you think?

I'd like this to be in v251 with other PRs for udev. Especially, the third commit should be important to make inotify watch stabler.

src/udev/udevd.c
yuwata: Added testcases.

keszybz:
LGTM.

PR #23089, opened 2022-04-15
sd-device: use path_extract_filename()

yuwata:


src/libsystemd/sd-device/sd-device.c
```diff
@@ -1171,24 +1171,11 @@ _public_ int sd_device_get_devname(sd_device *device, const char **devname) {
 static int device_set_sysname_and_sysnum(sd_device *device) {
         _cleanup_free_ char *sysname = NULL;
         const char *sysnum = NULL;
-        const char *pos;
         size_t len = 0;
 
-        if (!device->devpath)
```
bluca, line 1177: did you intend to remove this check? if so, it should have an assert

bluca, line 1177: did you intend to remove this check? if so, it should have an assert

yuwata: The condition should be already handled by `path_extract_filename()`.

bluca: but `device` is dereferenced before that runs?

yuwata: Ah, indeed. Will fix it. Thanks.

bluca (approved):

src/basic/fs-util.c
```diff
@@ -156,24 +156,16 @@ int readlink_malloc(const char *p, char **ret) {
...
-                return -ENOENT;
-
-        value = strdup(value);
-        if (!value)
-                return -ENOMEM;
-
-        *ret = value;
-
-        return 0;
+        return path_extract_filename(link, ret);
```
poettering, line 168: so, path_extract_filename() will return O_DIRECTORY if the path ends in a `/`. Which we shouldn't allow here I guess. Maybe hence add:

```c
_cleanup_free char *value = NULL;

r = path_extract_filename(link, &value);
if (r < 0)
        return r;
if (r == O_DIRECTORY)
        return _EINVAL;
*ret = TAKE_PTR(value);
return 0;
```

src/libsystemd/sd-device/sd-device.c
```diff
@@ -1171,24 +1171,13 @@ _public_ int sd_device_get_devname(sd_device *device, const char **devname) {
...
-
-        /* devpath is not a root directory */
-        if (*pos == '\0' || pos <= device->devpath)
-                return -EINVAL;
+        assert(device);
 
-        sysname = strdup(pos);
-        if (!sysname)
-                return -ENOMEM;
+        r = path_extract_filename(device->devpath, &sysname);
```
poettering, line 1178: maybe here to, refuse if O_DIRECTORY is returned?

src/libsystemd/sd-device/sd-device.c
```diff
@@ -1180,21 +1180,14 @@ static int device_set_sysname_and_sysnum(sd_device *device) {
...
-                        sysname[len] = '/';
-
-                len++;
-        }
-
-        /* trailing number */
-        while (len > 0 && isdigit(sysname[--len]))
-                sysnum = &sysname[len];
+        for (p = strchrnul(sysname, '!'); *p != '\0'; p = strchrnul(p, '!'))
+                *p = '/';
```
poettering, line 1184: hmm, why not use regular strchr()? sounds simpler:

```c
for (p = strchr(sysname, '!'); *p; p = strchr(p, '!'))
    *p = '/';
```

poettering: btw, we have a very similar function efi_tilt_backslashes() already. maybe we should unify this into a common helper?

i.e. `char *string_replace_char(char *buf, char old_char, char new_char);`

and then make efi_tilt_backslashes() just a trivial inline wrapper around this?

src/libsystemd/sd-device/sd-device.c
```diff
@@ -1180,21 +1180,14 @@ static int device_set_sysname_and_sysnum(sd_device *device) {
...
 
-        if (len == 0)
-                sysnum = NULL;
+        /* trailing number (refuse number only sysname)*/
+        for (; p > sysname && isdigit(p[-1]); p--)
```
poettering, line 1187: btw, isdigit() is kinda nasty, since in theory locale dependent (i.e. there's even `isdigt_l()` to run it with a different locale!). To be locale-independent we should probably replace most cases where we use `isdigit(x)` right now by `strchr(DIGITS, x)` instead. Since most of the time we only care about ASCII, like in this case.

src/basic/fs-util.c
yuwata: Ouch!

PR #23090, opened 2022-04-15
firewall-util: cleanups

yuwata:


medhefgo (approved):

bluca (approved):

src/shared/firewall-util-nft.c
```diff
@@ -29,21 +29,24 @@
 
 #define UDP_DPORT_OFFSET 2
 
-static int nfnl_netlink_sendv(sd_netlink *nfnl,
-                              sd_netlink_message *messages[],
-                              size_t msgcount) {
+static int nfnl_netlink_sendv(
+                sd_netlink *nfnl,
+                sd_netlink_message **messages,
```
poettering, line 34: i think it makes sense to write this as `*messages[]` rather than `**messages`, even though it's actually the same. For the simple reason that it indicates to the reader that we expect more than a pointer to a pointer here, but actually a pointer to an array here.

In fact, given this requires at least one element to be specified we could even write:

```c
sd_netlink_message *messages[static 1],
```

here.

PR #23094, opened 2022-04-15
systemctl: colorize "enabled" in 'systemctl status ...'

### LINKED ISSUE #16932: Colorize "enabled" in `systemctl status ...`
Reducing reading time of output of `systemctl status ...`.

`systemctl statatus ...` already colotizes "active (running)" into bold green.  it would be nice if "enabled" is also colorized.  If not in bold green, then maybe in just bold white.

Likewise, colorize the `disabled`s into some gray color.

This way, when we want to check if service is enabled, our eyes can only scan for colorized texts, such as bold green or bold white.  This is more convenient than scanning for word `enabled` specially that there is 2 `enabled` (one for service and one for vendor preset).

amarjargal:
Fixes #16932. 

"enabled" state is highlighted in green and "disabled" state is 
highlighted in yellow because I felt that white and grey colors were not 
so distinguishable. Other states are not highlighted. Any other coloring 
suggestions are welcome!

### DIFF
--- a/src/systemctl/systemctl-show.c
+++ b/src/systemctl/systemctl-show.c
@@ -298,12 +298,27 @@ static void format_active_state(const char *active_state, const char **active_on
                 *active_on = *active_off = "";
 }
 
+static void format_enable_state(const char *enable_state, const char **enable_on, const char **enable_off) {
+        assert(enable_on);
+        assert(enable_off);
+
+        if (streq_ptr(enable_state, "disabled")) {
+                *enable_on = ansi_highlight_yellow();
+                *enable_off = ansi_normal();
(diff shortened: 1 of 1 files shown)


src/systemctl/systemctl-show.c
```diff
@@ -298,12 +298,24 @@ static void format_active_state(const char *active_state, const char **active_on
                 *active_on = *active_off = "";
 }
 
+static void format_enable_state(const char *enable_state, const char **enable_on, const char **enable_off) {
+        if (streq_ptr(enable_state, "disabled")) {
```
bluca, line 305: `assert(enable_on)` (also for enable_off)

bluca (approved):

PR #23099, opened 2022-04-16
sd-bus: fix counter

### LINKED ISSUE #23097: There seems to be a heap-use-after-free in systemd
It was discovered by `dfuzzer` with a patch allowing it to poke void methods: https://github.com/matusmarhefka/dfuzzer/issues/18. I haven't figured out how to reproduce it without `dfuzzer` yet but here's the backtrace just in case:
```sh
[ 4373.606755] systemd[248]: ==248==ERROR: AddressSanitizer: heap-use-after-free on address 0x61f000008d58 at pc 0x7fb821dbaa93 bp 0x7ffc8f350b90 sp 0x7ffc8f350b88
[ 4373.607549] systemd[248]: READ of size 8 at 0x61f000008d58 thread T0 (systemd)
[ 4373.790923] systemd[248]:     #0 0x7fb821dbaa92 in bus_unit_track_handler ../src/core/dbus-unit.c:2522
[ 4373.796728] systemd[248]:     #1 0x7fb82072aa12 in bus_track_dispatch ../src/libsystemd/sd-bus/bus-track.c:375
[ 4373.800353] systemd[248]:     #2 0x7fb820752d20 in dispatch_track ../src/libsystemd/sd-bus/sd-bus.c:2975
[ 4373.801292] systemd[248]:     #3 0x7fb820752f7d in process_running ../src/libsystemd/sd-bus/sd-bus.c:2994
[ 4373.802140] systemd[248]:     #4 0x7fb820755e87 in bus_process_internal ../src/libsystemd/sd-bus/sd-bus.c:3224
[ 4373.802864] systemd[248]:     #5 0x7fb820756021 in sd_bus_process ../src/libsystemd/sd-bus/sd-bus.c:3251
[ 4373.803478] systemd[248]:     #6 0x7fb820759722 in io_callback ../src/libsystemd/sd-bus/sd-bus.c:3602
[ 4373.804835] systemd[248]:     #7 0x7fb82091a734 in source_dispatch ../src/libsystemd/sd-event/sd-event.c:3591
[ 4373.806036] systemd[248]:     #8 0x7fb820922f4e in sd_event_dispatch ../src/libsystemd/sd-event/sd-event.c:4175
[ 4373.806848] systemd[248]:     #9 0x7fb8209240b2 in sd_event_run ../src/libsystemd/sd-event/sd-event.c:4236
[ 4373.811527] systemd[248]:     #10 0x7fb821f1b153 in manager_loop ../src/core/manager.c:3042
[ 4373.814282] systemd[248]:     #11 0x41c1d3 in invoke_main_loop ../src/core/main.c:1893
[ 4373.815231] systemd[248]:     #12 0x4247c9 in main ../src/core/main.c:2952
[ 4373.816323] systemd[248]:     #13 0x7fb81ec0655f in __libc_start_call_main (/lib64/libc.so.6+0x2d55f)
[ 4373.817461] systemd[248]:     #14 0x7fb81ec0660b in __libc_start_main_impl (/lib64/libc.so.6+0x2d60b)
[ 4373.818385] systemd[248]:     #15 0x4072d4 in _start (/usr/lib/systemd/systemd+0x4072d4)
[ 4373.819141] systemd[248]: 0x61f000008d58 is located 216 bytes inside of 2960-byte region [0x61f000008c80,0x61f000009810)
[ 4373.819840] systemd[248]: freed by thread T0 (systemd) here:
[ 4373.820439] systemd[248]:     #0 0x7fb822a40627 in free (/usr/lib64/libasan.so.6.0.0+0xae627)
[ 4373.824006] systemd[248]:     #1 0x7fb8220d7457 in unit_free ../src/core/unit.c:810
[ 4373.824933] systemd[248]:     #2 0x7fb821f070df in manager_clear_jobs_and_units ../src/core/manager.c:1430
[ 4373.825942] systemd[248]:     #3 0x7fb821f1fc12 in manager_reload ../src/core/manager.c:3340
[ 4373.826820] systemd[248]:     #4 0x41c56d in invoke_main_loop ../src/core/main.c:1926
[ 4373.827461] systemd[248]:     #5 0x4247c9 in main ../src/core/main.c:2952
[ 4373.838268] systemd[248]:     #6 0x7fb81ec0655f in __libc_start_call_main (/lib64/libc.so.6+0x2d55f)
[ 4373.839159] systemd[248]: previously allocated by thread T0 (systemd) here:
[ 4373.839913] systemd[248]:     #0 0x7fb822a40ad7 in calloc (/usr/lib64/libasan.so.6.0.0+0xaead7)
[ 4373.840525] systemd[248]:     #1 0x7fb8220bff49 in unit_new ../src/core/unit.c:95
[ 4373.841326] systemd[248]:     #2 0x7fb8220c1518 in unit_new_for_name ../src/core/unit.c:138
[ 4373.843384] systemd[248]:     #3 0x7fb821f47228 in mount_setup_new_unit ../src/core/mount.c:1561
[ 4373.844561] systemd[248]:     #4 0x7fb821f483e1 in mount_setup_unit ../src/core/mount.c:1732
[ 4373.845704] systemd[248]:     #5 0x7fb821f48cad in mount_load_proc_self_mountinfo ../src/core/mount.c:1777
[ 4373.846515] systemd[248]:     #6 0x7fb821f4ab3c in mount_enumerate ../src/core/mount.c:1927
[ 4373.847380] systemd[248]:     #7 0x7fb821f0a5c8 in manager_enumerate ../src/core/manager.c:1580
[ 4373.848309] systemd[248]:     #8 0x7fb821f0cba9 in manager_startup ../src/core/manager.c:1786
[ 4373.849373] systemd[248]:     #9 0x4243fc in main ../src/core/main.c:2924
[ 4373.850426] systemd[248]:     #10 0x7fb81ec0655f in __libc_start_call_main (/lib64/libc.so.6+0x2d55f)
[ 4373.851264] systemd[248]: SUMMARY: AddressSanitizer: heap-use-after-free ../src/core/dbus-unit.c:2522 in bus_unit_track_handler
[ 4373.851923] systemd[248]: Shadow bytes around the buggy address:
[ 4373.852529] systemd[248]:   0x0c3e7fff9150: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[ 4373.853091] systemd[248]:   0x0c3e7fff9160: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[ 4373.853715] systemd[248]:   0x0c3e7fff9170: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[ 4373.854319] systemd[248]:   0x0c3e7fff9180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
[ 4373.855106] systemd[248]:   0x0c3e7fff9190: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
[ 4373.855937] systemd[248]: =>0x0c3e7fff91a0: fd fd fd fd fd fd fd fd fd fd fd[fd]fd fd fd fd
[ 4373.856565] systemd[248]:   0x0c3e7fff91b0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
[ 4373.856845] systemd[248]:   0x0c3e7fff91c0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
[ 4373.857083] systemd[248]:   0x0c3e7fff91d0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
[ 4373.857429] systemd[248]:   0x0c3e7fff91e0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
[ 4373.857684] systemd[248]:   0x0c3e7fff91f0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
[ 4373.857915] systemd[248]: Shadow byte legend (one shadow byte represents 8 application bytes):
[ 4373.858144] systemd[248]:   Addressable:           00
[ 4373.858530] systemd[248]:   Partially addressable: 01 02 03 04 05 06 07
[ 4373.858768] systemd[248]:   Heap left redzone:       fa
[ 4373.858997] systemd[248]:   Freed heap region:       fd
[ 4373.859744] systemd[248]:   Stack left redzone:      f1
[ 4373.860839] systemd[248]:   Stack mid redzone:       f2
[ 4373.861553] systemd[248]:   Stack right redzone:     f3
[ 4373.862096] systemd[248]:   Stack after return:      f5
[ 4373.862766] systemd[248]:   Stack use after scope:   f8
[ 4373.863735] systemd[248]:   Global redzone:          f9
[ 4373.864856] systemd[248]:   Global init order:       f6
[ 4373.866015] systemd[248]:   Poisoned by user:        f7
[ 4373.866830] systemd[248]:   Container overflow:      fc
[ 4373.867627] systemd[248]:   Array cookie:            ac
[ 4373.868245] systemd[248]:   Intra object redzone:    bb
[ 4373.868820] systemd[248]:   ASan internal:           fe
[ 4373.869573] systemd[248]:   Left alloca redzone:     ca
[ 4373.870241] systemd[248]:   Right alloca redzone:    cb
[ 4373.870929] systemd[248]:   Shadow gap:              cc
[ 4373.871283] systemd[248]: ==248==ABORTING
```

yuwata:
Fixes #23097.

### DIFF
--- a/src/libsystemd/sd-bus/test-bus-track.c
+++ b/src/libsystemd/sd-bus/test-bus-track.c
@@ -10,6 +10,7 @@
 
 static bool track_cb_called_x = false;
 static bool track_cb_called_y = false;
+static bool track_destroy_called_z = false;
 
 static int track_cb_x(sd_bus_track *t, void *userdata) {
 
@@ -39,9 +40,17 @@ static int track_cb_y(sd_bus_track *t, void *userdata) {
         return 0;
 }
 
+static int track_cb_z(sd_bus_track *t, void *userdata) {
+        assert_not_reached();
+}
+
+static void track_destroy_z(void *userdata) {
+        track_destroy_called_z = true;
+}
+
 int main(int argc, char *argv[]) {
         _cleanup_(sd_event_unrefp) sd_event *event = NULL;
-        _cleanup_(sd_bus_track_unrefp) sd_bus_track *x = NULL, *y = NULL;
+        _cleanup_(sd_bus_track_unrefp) sd_bus_track *x = NULL, *y = NULL, *z = NULL;
         _cleanup_(sd_bus_unrefp) sd_bus *a = NULL, *b = NULL;
         bool use_system_bus = false;
         const char *unique;
@@ -82,6 +91,53 @@ int main(int argc, char *argv[]) {
         assert_se(sd_bus_get_unique_name(a, &unique) >= 0);
 
         assert_se(sd_bus_track_add_name(y, unique) >= 0);
+
+        /* Basic tests. */
+        assert_se(sd_bus_track_new(a, &z, track_cb_z, NULL) >= 0);
+
+        /* non-recursive case */
+        assert_se(sd_bus_track_set_recursive(z, false) >= 0);
+        assert_se(sd_bus_track_get_recursive(z) == 0);
+        assert_se(!sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 0);
+        assert_se(sd_bus_track_remove_name(z, unique) == 0);
+        assert_se(sd_bus_track_add_name(z, unique) >= 0);
+        assert_se(sd_bus_track_add_name(z, unique) >= 0);
+        assert_se(sd_bus_track_add_name(z, unique) >= 0);
+        assert_se(sd_bus_track_set_recursive(z, true) == -EBUSY);
+        assert_se(sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 1);
+        assert_se(sd_bus_track_remove_name(z, unique) == 1);
+        assert_se(!sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 0);
+        assert_se(sd_bus_track_remove_name(z, unique) == 0);
+
+        /* recursive case */
+        assert_se(sd_bus_track_set_recursive(z, true) >= 0);
+        assert_se(sd_bus_track_get_recursive(z) == 1);
+        assert_se(!sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 0);
+        assert_se(sd_bus_track_remove_name(z, unique) == 0);
+        assert_se(sd_bus_track_add_name(z, unique) >= 0);
+        assert_se(sd_bus_track_add_name(z, unique) >= 0);
+        assert_se(sd_bus_track_add_name(z, unique) >= 0);
+        assert_se(sd_bus_track_set_recursive(z, false) == -EBUSY);
+        assert_se(sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 3);
+        assert_se(sd_bus_track_remove_name(z, unique) == 1);
+        assert_se(sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 2);
+        assert_se(sd_bus_track_remove_name(z, unique) == 1);
+        assert_se(sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 1);
+        assert_se(sd_bus_track_remove_name(z, unique) == 1);
+        assert_se(!sd_bus_track_contains(z, unique));
+        assert_se(sd_bus_track_count_name(z, unique) == 0);
+        assert_se(sd_bus_track_remove_name(z, unique) == 0);
+
+        assert_se(sd_bus_track_set_destroy_callback(z, track_destroy_z) >= 0);
+        z = sd_bus_track_unref(z);
+        assert_se(track_destroy_called_z);
 
         /* Now make b's name disappear */
         sd_bus_close(b);


yuwata:
cc @evverx and @mrc0mmand

src/libsystemd/sd-bus/bus-track.c
```diff
@@ -190,14 +190,8 @@ _public_ int sd_bus_track_add_name(sd_bus_track *track, const char *name) {
 
         i = hashmap_get(track->names, name);
         if (i) {
-                if (track->recursive) {
-                        unsigned k = i->n_ref + 1;
-
-                        if (k < i->n_ref) /* Check for overflow */
-                                return -EOVERFLOW;
```
evverx, line 197: I'm not sure `track_item_ref` can replace it. As far as I can see there instead of returning `-EOVERFLOW` the counter either overflows if `systemd` is built with `NDEBUG` or the function crashes when `assert` is triggered.

evverx:
FWIW `dfuzzer` can no longer trigger https://github.com/systemd/systemd/issues/23097 with this patch applied

src/libsystemd/sd-bus/bus-track.c
yuwata: Actually, this changes the previous behavior. But, I do not think we need to keep the previous behavior.

E.g.
```
for (unsigned i = 0; i < UINT_MAX; i++)
    sd_bus_message_ref(m);
assert_se(m->n_ref == 0);
sd_bus_message_unref(m);
```
Then, the last `sd_bus_message_unref()` triggers assertion.

Of course, in the macro which defines _ref(), we can use `assert_return(p->n_ref < UINT_MAX, NULL);`, and in bus-track.c
```
if (track->recursive && !track_item_ref(i))
        return -EOVERFLOW;
```

However, in most places, we assume `xxx_ref()` returns the same as the input. If we use `assert_return()`, then the assumption is broken...

src/basic/macro.h
```diff
@@ -396,7 +396,11 @@ static inline int __coverity_check_and_return__(int condition) {
                 if (!p)                                         \
                         return NULL;                            \
                                                                 \
+                /* For type check. */                           \
+                _unused_ unsigned *q = &p->n_ref;               \
                 assert(p->n_ref > 0);                           \
+                assert(p->n_ref < UINT_MAX);                    \
```
poettering: idea: drop the `_unused`, and change the two asserts to use `*q` instead?

src/libsystemd/sd-bus/bus-track.c
```diff
@@ -191,12 +191,12 @@ _public_ int sd_bus_track_add_name(sd_bus_track *track, const char *name) {
         i = hashmap_get(track->names, name);
         if (i) {
                 if (track->recursive) {
-                        unsigned k = track->n_ref + 1;
+                        unsigned k = i->n_ref + 1;
```
poettering: ouch!

src/libsystemd/sd-bus/bus-track.c
```diff
@@ -191,12 +191,12 @@ _public_ int sd_bus_track_add_name(sd_bus_track *track, const char *name) {
...
 
-                        if (k < track->n_ref) /* Check for overflow */
+                        if (k < i->n_ref) /* Check for overflow */
```
poettering: we should not do overflow checks this way really. We used to, but it's not in compliance with C standard. i.e. overflows are undefined behaviour in C, and thus the compiler may assume that the the earlier +1 doesn't overflow, and thus optimize this check away.

thus, this should be reworked to:

```c
if (i->n_ref >= UINT_MAX)
        return -EOVEFLOW;

i->n_ref++;
```

i.e. never execute the +1 unless we verified it's safe.

src/libsystemd/sd-bus/bus-track.c
```diff
@@ -264,14 +262,12 @@ _public_ int sd_bus_track_remove_name(sd_bus_track *track, const char *name) {
         i = hashmap_get(track->names, name);
         if (!i)
                 return -EUNATCH;
-        if (i->n_ref <= 0)
-                return -EUNATCH;
 
-        i->n_ref--;
-
-        if (i->n_ref <= 0)
+        if (i->n_ref <= 1)
```
poettering, line 264: maybe add `assert(i->n_ref >= 1)` first, just in case

src/libsystemd/sd-bus/bus-track.c
yuwata: Sure, but this code block will be replaced with `track_item_ref()` in the later commit. I'd like to keep the commit as is.

yuwata:
@poettering Thank you for the review. Updated. PTAL.

src/libsystemd/sd-bus/bus-track.c
evverx: Given that `sd_bus_track_add_name` is public I wonder if it makes sense to mention in NEWS that it's going to either crash or overflow `i->n_ref` with NDEBUG defined if its callers manage to trigger it enough times. FWIW I don't think PID1 should call functions that can crash it like that.

poettering: People shouldn't set NDEBUG. That's generally a bad idea. That said, the overflow assert here should be upgraded to assert_se(), as it should never be optimized away. I filed #23138 for that.

And yes, I guess given that this is directly user controllable, we should not use the generic ref counting code here (where I think the assertion is OK, as there shoud be limits on resources by other means anyway), nd instead have manual one that returns an error instead of an assertion back.

i.e. assertions are OK for local programming errors, and in general local ref counting overflows are that: objects simply should have been limited in other ways beforehand. But this here is not a local ref counting logic, but a directly remotely accessible in the usual ways, hence we should return an error here.

poettering: → #23139

PR #23101, opened 2022-04-17
sd-device: several cleanups

yuwata:


src/libsystemd/sd-device/device-private.c
```diff
@@ -938,8 +938,7 @@ static int device_tag(sd_device *device, const char *tag, bool add) {
                 if (r < 0)
                         return r;
         } else {
-                r = unlink(path);
-                if (r < 0 && errno != ENOENT)
+                if (unlink(path) < 0 && errno != ENOENT)
```
bluca: can be merged with the `else` above

bluca (approved):

yuwata:
@bluca Thank you for the review. The above points are addressed with a slightly different (and more aggressive) way. Upgrading the green label.

Issue #22920, opened 2022-03-31
CentOS CI's C8S job randomly times out during boot with multipath


mrc0mmand:
**systemd version the issue has been seen with**
 > latest main

**Used distribution**
 > CentOS Stream 8

In the recent couple of days some of the C8S jobs abort during reboot caused by a time out. Closer inspection reveals that dracut stalls the boot for a while before discovering the multipath rootfs. However, the dracut's version haven't changed for a while on C8S so I wonder if it's not caused by some of the recent patches.

```
[    8.794182] multipathd[672]: --------start up--------
[    8.794182] multipathd[672]: read /etc/multipath.conf
[    8.797702] multipathd[672]: failed to increase buffer size
[    8.798386] multipathd[672]: path checkers start up
[    8.827617] kernel: device-mapper: multipath service-time: version 0.3.0 loaded
[    8.850143] dracut-initqueue[695]: Scanning devices sda2 sdb2  for LVM logical volumes vg_n35/root vg_n35/swap
[    8.855366] multipathd[672]: mpatha: load table [0 467664896 multipath 0 0 1 1 service-time 0 2 1 8:0 1 8:16 1]
[    8.913834] dracut-initqueue[733]: Volume group "vg_n35" not found
[    8.913834] dracut-initqueue[733]: Cannot process volume group vg_n35
[    9.056417] systemd[1]: Found device /dev/mapper/vg_n35-swap.
[    9.057946] systemd[1]: Starting Resume from hibernation using device /dev/mapper/vg_n35-swap...
[    9.064928] systemd-hibernate-resume[758]: Could not resume from '/dev/mapper/vg_n35-swap' (253:3).
[    9.064889] kernel: PM: Image not found (code -22)
[    9.065890] systemd[1]: systemd-hibernate-resume@dev-mapper-vg_n35\x2dswap.service: Deactivated successfully.
[    9.066164] systemd[1]: Finished Resume from hibernation using device /dev/mapper/vg_n35-swap.
[    9.066295] audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=systemd-hibernate-resume@dev-mapper-vg_n35\x2dswap comm="systemd" exe="/usr/lib/systemd/systemd" hostnam>
[...]

PR #23102, opened 2022-04-17
cryptenroll,homectl: Introduce --fido2-credential-algorithm option

MkfsSion:
* Some authenticators(like Yubikey) support credential algorithm other than ES256
* Introduce a new option so users can make use of it

### DIFF
--- a/man/homectl.xml
+++ b/man/homectl.xml
@@ -357,6 +357,19 @@
       </varlistentry>
 
       <varlistentry>
+        <term><option>--fido2-credential-algorithm=</option><replaceable>STRING</replaceable></term>
+        <listitem><para>Specify COSE algorithm used in credential generation. The default value is
+        <literal>es256</literal>. Supported values are <literal>es256</literal>, <literal>rs256</literal>
+        and <literal>eddsa</literal>.</para>
+
+        <para><literal>es256</literal> denotes ECDSA over NIST P-256 with SHA-256. <literal>rs256</literal>
+        denotes 2048-bit RSA with PKCS#1.5 padding and SHA-256. <literal>eddsa</literal> denotes
+        EDDSA over Curve25519 with SHA-512.</para>
+
+        <para>Note that your authenticator may not support some algorithms.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><option>--fido2-device=</option><replaceable>PATH</replaceable></term>
 
         <listitem><para>Takes a path to a Linux <literal>hidraw</literal> device
--- a/man/systemd-cryptenroll.xml
+++ b/man/systemd-cryptenroll.xml
@@ -123,6 +123,19 @@
       </varlistentry>
 
       <varlistentry>
+        <term><option>--fido2-credential-algorithm=</option><replaceable>STRING</replaceable></term>
+        <listitem><para>Specify COSE algorithm used in credential generation. The default value is
+        <literal>es256</literal>. Supported values are <literal>es256</literal>, <literal>rs256</literal>
+        and <literal>eddsa</literal>.</para>
+
+        <para><literal>es256</literal> denotes ECDSA over NIST P-256 with SHA-256. <literal>rs256</literal>
+        denotes 2048-bit RSA with PKCS#1.5 padding and SHA-256. <literal>eddsa</literal> denotes
+        EDDSA over Curve25519 with SHA-512.</para>
+
+        <para>Note that your authenticator may not support some algorithms.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><option>--fido2-device=</option><replaceable>PATH</replaceable></term>
 
         <listitem><para>Enroll a FIDO2 security token that implements the <literal>hmac-secret</literal>
--- a/src/cryptenroll/cryptenroll.c
+++ b/src/cryptenroll/cryptenroll.c
@@ -39,6 +39,11 @@ static size_t arg_n_wipe_slots = 0;
 static WipeScope arg_wipe_slots_scope = WIPE_EXPLICIT;
 static unsigned arg_wipe_slots_mask = 0; /* Bitmask of (1U << EnrollType), for wiping all slots of specific types */
 static Fido2EnrollFlags arg_fido2_lock_with = FIDO2ENROLL_PIN | FIDO2ENROLL_UP;
+#if HAVE_LIBFIDO2
+static int arg_fido2_cred_alg = COSE_ES256;
+#else
+static int arg_fido2_cred_alg = 0;
+#endif
 
 assert_cc(sizeof(arg_wipe_slots_mask) * 8 >= _ENROLL_TYPE_MAX);
 
@@ -89,6 +94,8 @@ static int help(void) {
                "     --recovery-key    Enroll a recovery key\n"
                "     --pkcs11-token-uri=URI\n"
                "                       Specify PKCS#11 security token URI\n"
+               "     --fido2-credential-algorithm=STRING\n"
+               "                       Specify COSE algorithm for FIDO2 credential\n"
                "     --fido2-device=PATH\n"
                "                       Enroll a FIDO2-HMAC security token\n"
                "     --fido2-with-client-pin=BOOL\n"
@@ -129,6 +136,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_FIDO2_WITH_PIN,
                 ARG_FIDO2_WITH_UP,
                 ARG_FIDO2_WITH_UV,
+                ARG_FIDO2_CRED_ALG,
         };
 
         static const struct option options[] = {
@@ -137,6 +145,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "password",                     no_argument,       NULL, ARG_PASSWORD         },
                 { "recovery-key",                 no_argument,       NULL, ARG_RECOVERY_KEY     },
                 { "pkcs11-token-uri",             required_argument, NULL, ARG_PKCS11_TOKEN_URI },
+                { "fido2-credential-algorithm",   required_argument, NULL, ARG_FIDO2_CRED_ALG   },
                 { "fido2-device",                 required_argument, NULL, ARG_FIDO2_DEVICE     },
                 { "fido2-with-client-pin",        required_argument, NULL, ARG_FIDO2_WITH_PIN   },
                 { "fido2-with-user-presence",     required_argument, NULL, ARG_FIDO2_WITH_UP    },
@@ -239,6 +248,12 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_pkcs11_token_uri = TAKE_PTR(uri);
                         break;
                 }
+
+                case ARG_FIDO2_CRED_ALG:
+                        r = parse_fido2_algorithm(optarg, &arg_fido2_cred_alg);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to parse COSE algorithm: %s", optarg);
+                        break;
 
                 case ARG_FIDO2_DEVICE: {
                         _cleanup_free_ char *device = NULL;
@@ -566,7 +581,7 @@ static int run(int argc, char *argv[]) {
                 break;
 
         case ENROLL_FIDO2:
-                slot = enroll_fido2(cd, vk, vks, arg_fido2_device, arg_fido2_lock_with);
+                slot = enroll_fido2(cd, vk, vks, arg_fido2_device, arg_fido2_lock_with, arg_fido2_cred_alg);
                 break;
 
         case ENROLL_TPM2:
--- a/src/home/homectl.c
+++ b/src/home/homectl.c
@@ -61,6 +61,11 @@ static uint64_t arg_disk_size_relative = UINT64_MAX;
 static char **arg_pkcs11_token_uri = NULL;
 static char **arg_fido2_device = NULL;
 static Fido2EnrollFlags arg_fido2_lock_with = FIDO2ENROLL_PIN | FIDO2ENROLL_UP;
+#if HAVE_LIBFIDO2
+static int arg_fido2_cred_alg = COSE_ES256;
+#else
+static int arg_fido2_cred_alg = 0;
+#endif
 static bool arg_recovery_key = false;
 static JsonFormatFlags arg_json_format_flags = JSON_FORMAT_OFF;
 static bool arg_and_resize = false;
@@ -1114,7 +1119,7 @@ static int acquire_new_home_record(UserRecord **ret) {
         }
 
         STRV_FOREACH(i, arg_fido2_device) {
-                r = identity_add_fido2_parameters(&v, *i, arg_fido2_lock_with);
+                r = identity_add_fido2_parameters(&v, *i, arg_fido2_lock_with, arg_fido2_cred_alg);
                 if (r < 0)
                         return r;
         }
@@ -1473,7 +1478,7 @@ static int acquire_updated_home_record(
         }
 
         STRV_FOREACH(i, arg_fido2_device) {
-                r = identity_add_fido2_parameters(&json, *i, arg_fido2_lock_with);
+                r = identity_add_fido2_parameters(&json, *i, arg_fido2_lock_with, arg_fido2_cred_alg);
                 if (r < 0)
                         return r;
         }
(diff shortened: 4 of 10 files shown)


src/shared/libfido2-util.h
```diff
@@ -14,6 +14,14 @@ typedef enum Fido2EnrollFlags {
...
 } Fido2EnrollFlags;
 
+typedef enum COSEAlgorithm {
+        COSE_ALG_RS256 = -257,
+        COSE_ALG_EDDSA = -8,
+        COSE_ALG_ES256 = -7,
+        _COSE_ALG_TYPE_MAX,
+        _COSE_ALG_TYPE_INVALID = -EINVAL,
+} COSEAlgorithm;
+
```
yuwata: Why these are (mostly) redefined? Why not simply use `COSE_ES256` or friends?

If you really want to define an enum, then please use like that
```
typedef enum COSEAlgorithm {
        COSE_ALG_RS256 = COSE_RS256,
        COSE_ALG_EDDSA = COSE_EDDSA,
        COSE_ALG_ES256 = COSE_ES256,
} COSEAlgorithm;
```
We usually adds _MAX and _INVALID, but I think this is not suitable here. As, it is kind of definition, nothing enumerated.

src/home/homectl.c
```diff
@@ -3485,6 +3488,17 @@ static int parse_argv(int argc, char *argv[]) {
...
 
+                case ARG_FIDO2_CRED_ALG:
+                        if (streq(optarg, "ES256"))
+                                arg_fido2_cred_alg = COSE_ALG_ES256;
+                        else if(streq(optarg, "RS256"))
+                                arg_fido2_cred_alg = COSE_ALG_RS256;
+                        else if(streq(optarg, "EDDSA"))
+                                arg_fido2_cred_alg = COSE_ALG_EDDSA;
+                        else
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "Unsupported COSE algorithm: %s", optarg);
```
yuwata: How about to introduce a simple helper `parse_fido2_algorithm()`?

I am not familiar with each algorithm, but why `COSE_ECDH_ES256` is excluded here?

src/shared/libfido2-util.c
```diff
@@ -628,10 +629,10 @@ int fido2_generate_hmac_hash(
                 return log_error_errno(SYNTHETIC_ERRNO(EIO),
                                        "Failed to set FIDO2 credential relying party ID/name: %s", sym_fido_strerr(r));
 
-        r = sym_fido_cred_set_type(c, COSE_ES256);
+        r = sym_fido_cred_set_type(c, cred_alg);
         if (r != FIDO_OK)
                 return log_error_errno(SYNTHETIC_ERRNO(EIO),
-                                       "Failed to set FIDO2 credential type to ES256: %s", sym_fido_strerr(r));
+                                       "Failed to set FIDO2 credential type to %d: %s", cred_alg, sym_fido_strerr(r));
```
yuwata: Let's also introduce a simple helper `fido2_algorithm_to_string()`, and use here.

yuwata (changes requested):

src/shared/libfido2-util.h
MkfsSion: The first attempt failed build due to COSE_* macro missing. Possiblely due to the lack of libfido2.

src/home/homectl.c
MkfsSion: > How about to introduce a simple helper `parse_fido2_algorithm()`?
> 
> I am not familiar with each algorithm, but why `COSE_ECDH_ES256` is excluded here?

Ok I'll consider introduce the helper.
As libfido2 manpage `man fido_cred_set_type` states, it only accepts the three values.
Source code:https://github.com/Yubico/libfido2/blob/main/src/cred.c#L968-L969
Manpage state:

>   The fido_cred_set_type() function sets the type of cred to cose_alg, where cose_alg is COSE_ES256,
>   COSE_RS256, or COSE_EDDSA.  The type of a credential may only be set once.  Note that not all authentica‐
>   tors support COSE_RS256 or COSE_EDDSA.

src/shared/libfido2-util.h
yuwata: I guess the following should work without defining the enum.
```
/* libfido2-util.c */
#if HAVE_LIBFIDO2
int parse_fido2_algorithm(const char *s, int *ret) {
        int a;

        assert(s);

        if (streq(s, "ES256"))
                a = COSE_ES256;
        else if(streq(s, "RS256"))
                a = COSE_RS256;
        else if(streq(s, "EDDSA"))
                a = COSE_EDDSA;
        else
                return -EINVAL;

        if (ret)
                *ret = a;
        return 0;
}
#endif
```
```
/* libfido2-util.h*/
#if HAVE_LIBFIDO2
int parse_fido2_algorithm(const char *s, int *ret);
#else
static inline int parse_fido2_algorithm(const char *s, int *ret) {
        return -EOPNOTSUPP;
}
#endif
```

src/home/homectl.c
yuwata: I see. Thanks.

src/shared/libfido2-util.h
MkfsSion: But we still need magic value `COSE_ES256` for the deault value for `arg_fido2_cred_alg`, it's that ok?

yuwata: Please also use `#if,else,endif` there.

src/shared/libfido2-util.c
```diff
@@ -1124,3 +1138,24 @@ int fido2_have_device(const char *device) {
...
+        else if(streq(s, "EDDSA"))
+                a = COSE_EDDSA;
+        else
+                return -EINVAL;
+
+        if (ret)
+                *ret = a;
+        return 0;
+}
+#endif
```
bluca: missing newline

src/shared/libfido2-util.h
```diff
@@ -109,15 +109,21 @@ int fido2_generate_hmac_hash(
...
+static inline int parse_fido2_algorithm(const char *s, int *ret) {
+        return -EOPNOTSUPP;
+}
 #endif
 
 int fido2_list_devices(void);
 int fido2_find_device_auto(char **ret);
 
-int fido2_have_device(const char *device);
+int fido2_have_device(const char *device);
```
bluca: missing newline

src/shared/libfido2-util.c
MkfsSion: done

src/shared/libfido2-util.h
MkfsSion: done

src/cryptenroll/cryptenroll.c
```diff
@@ -240,6 +249,16 @@ static int parse_argv(int argc, char *argv[]) {
                         break;
                 }
 
+                case ARG_FIDO2_CRED_ALG: {
```
yuwata: Unnecessary `{}`, as this code block has no variable declaration.

src/cryptenroll/cryptenroll.c
```diff
@@ -240,6 +249,16 @@ static int parse_argv(int argc, char *argv[]) {
...
+                        if (isempty(optarg))
+                                break;
```
yuwata: This is not necessary, as the option requires argument.

src/cryptenroll/cryptenroll.c
```diff
@@ -240,6 +249,16 @@ static int parse_argv(int argc, char *argv[]) {
...
+
+                        r = parse_fido2_algorithm(optarg, &arg_fido2_cred_alg);
+                        if (r < 0)
+                                return log_error_errno(SYNTHETIC_ERRNO(r), "Failed to parse COSE algorithm: %s", optarg);
```
yuwata: Please drop SYNTHETIC_ERRNO(), which is used when an errno is explicitly specified, e.g. `SYNTHETIC_ERRNO(EINVAL)`.
So,
```
return log_error_errno(r, "Failed to ...", ...);
```
is enough.

src/home/homectl.c
```diff
@@ -3485,6 +3492,15 @@ static int parse_argv(int argc, char *argv[]) {
                         strv_uniq(arg_pkcs11_token_uri);
                         break;
 
+                case ARG_FIDO2_CRED_ALG:
+                        if (isempty(optarg))
+                                break;
```
yuwata: Same here. Please drop this.

src/home/homectl.c
```diff
@@ -3485,6 +3492,15 @@ static int parse_argv(int argc, char *argv[]) {
...
+
+                        r = parse_fido2_algorithm(optarg, &arg_fido2_cred_alg);
+                        if (r < 0)
+                                return log_error_errno(SYNTHETIC_ERRNO(r), "Failed to parse COSE algorithm: %s", optarg);
```
yuwata: Also here. Please drop `SYNTHETIC_ERRNO()`.

yuwata (changes requested):
Several minor comments.

src/cryptenroll/cryptenroll.c
MkfsSion: Thanks for your review. Done.

src/home/homectl.c
MkfsSion: Done.

yuwata (approved):
Superficially LGTM. But, I am not familiar with the detail. Let's wait for other reviewers.

src/shared/libfido2-util.c
```diff
@@ -450,6 +450,19 @@ static int fido2_use_hmac_hash_specific_token(
...
 }
 
+static const char *fido2_algorithm_to_string(int alg) {
+        switch(alg) {
+                case COSE_ES256:
+                        return "ES256";
+                case COSE_RS256:
+                        return "RS256";
+                case COSE_EDDSA:
+                        return "EDDSA";
```
poettering: libfido2 appears to also know COSE_ECDH_ES256?

src/shared/libfido2-util.c
```diff
@@ -450,6 +450,19 @@ static int fido2_use_hmac_hash_specific_token(
...
+                default:
+                        return NULL;
+        }
+}
```
poettering, line 465: so almost all our string enumerations use lowercase spelling. I think we should continue that rule here, and thus this should be "es256", "rs256" and "eddsa"...

src/shared/libfido2-util.c
```diff
@@ -1124,3 +1138,24 @@ int fido2_have_device(const char *device) {
...
+
+#if HAVE_LIBFIDO2
+int parse_fido2_algorithm(const char *s, int *ret) {
+        int a;
+
+        assert(s);
+
+        if (streq(s, "ES256"))
+                a = COSE_ES256;
+        else if(streq(s, "RS256"))
```
poettering: space between `if` and `(`, please. (here and below)

man/homectl.xml
```diff
@@ -356,6 +356,19 @@
...
+        <term><option>--fido2-credential-algorithm=</option><replaceable>STRING</replaceable></term>
+        <listitem><para>Specify COSE algorithm used in credential generation. The default value is
+        <literal>ES256</literal>. Supported values are <literal>ES256</literal>, <literal>RS256</literal>
+        and <literal>EDDSA</literal>.</para>
+
+        <para><literal>ES256</literal> denotes ECDSA over NIST P-256 with SHA-256. <literal>RS256</literal>
+        denotes 2048-bit RSA with PKCS#1.5 padding and SHA-256. <literal>EDDSA</literal> denotes
+        EDDSA over Curve25519 with SHA-512.</para>
+
+        <para>Note that your authenticator may not support some algorithms.</para></listitem>
```
poettering, line 369: is there a way to enumerate supported algorithms of a device?

poettering (changes requested):
looks pretty good

src/shared/libfido2-util.c
MkfsSion: No, `fido_cred_set_type` only accept three values as the man page states

> The fido_cred_set_type() function sets the type of cred to cose_alg, where cose_alg is COSE_ES256,
> COSE_RS256, or COSE_EDDSA. The type of a credential may only be set once. Note that not all authentica‐
> tors support COSE_RS256 or COSE_EDDSA.

man/homectl.xml
MkfsSion: Yes, we can use `fido_cbor_info_algorithm_cose` to enumerate supported algorithm on device. I'll going to implement it.

src/shared/libfido2-util.c
MkfsSion: Should I change the argument for the option too?

src/shared/libfido2-util.c
poettering: might be good adding a comment to this function saying that this function only maps the stuff usable for creds. And thus COSE_ECDH_ES256 is not listed.

man/homectl.xml
poettering: that'd certainly be welcome though not necessary, i guess.

question is were to add this. maybe just query it in verify_features() and show it as debug info like the other bits of info we show there right now?

src/shared/libfido2-util.c
poettering: the C enum/defines should be all uppercase, but the string representaiton we parse or output should be lowercase. That's how we do it for almost all enums or enum-like things.

but not sure i grok the question fully?

MkfsSion: So I think it's not necessary to make the argument(like ES256) of the newly added option(fido2-credential-algorithm) to lowercase.

man/homectl.xml
MkfsSion: Yes, I think it's not necessary since if the authenticator doesn't support the specified algorithm, `fido_dev_make_cred` will fail with `FIDO_ERR_UNSUPPORTED_ALGORITHM`.

MkfsSion: After some search I found the function will only work for authenticator that support CTAP 2.1. There is no way to enumerate supported algorithms on authenticator before CTAP 2.1. `authenticatorGetInfo` method won't return `algorithms` field before that version.

Reference:
CTAP 2.0: https://fidoalliance.org/specs/fido-v2.0-rd-20170927/fido-client-to-authenticator-protocol-v2.0-rd-20170927.html#authenticatorgetinfo-0x04
CTAP 2.1: https://fidoalliance.org/specs/fido-v2.1-rd-20210309/fido-client-to-authenticator-protocol-v2.1-rd-20210309.html#authenticatorGetInfo

MkfsSion:
@mrc0mmand Seems the CI failure doesn't relate to the PR

mrc0mmand:
> @mrc0mmand Seems the CI failure doesn't relate to the PR

Indeed, that's a know issue (#22920), I'll re-trigger it.

man/homectl.xml
```diff
@@ -356,6 +356,19 @@
...
 
+      <varlistentry>
+        <term><option>--fido2-credential-algorithm=</option><replaceable>STRING</replaceable></term>
+        <listitem><para>Specify COSE algorithm used in credential generation. The default value is
+        <literal>ES256</literal>. Supported values are <literal>ES256</literal>, <literal>RS256</literal>
+        and <literal>EDDSA</literal>.</para>
+
+        <para><literal>ES256</literal> denotes ECDSA over NIST P-256 with SHA-256. <literal>RS256</literal>
+        denotes 2048-bit RSA with PKCS#1.5 padding and SHA-256. <literal>EDDSA</literal> denotes
+        EDDSA over Curve25519 with SHA-512.</para>
```
poettering, line 367: needs updating to use lowercase spelling of accepted values

man/systemd-cryptenroll.xml
```diff
@@ -122,6 +122,19 @@
         <filename>/etc/crypttab</filename> line.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--fido2-credential-algorithm=</option><replaceable>STRING</replaceable></term>
+        <listitem><para>Specify COSE algorithm used in credential generation. The default value is
+        <literal>ES256</literal>. Supported values are <literal>ES256</literal>, <literal>RS256</literal>
+        and <literal>EDDSA</literal>.</para>
```
poettering: ditto

poettering:
codewise looks excellent, but doc needs updating now, see comments.

man/homectl.xml
poettering: ok, it would be the cherry on top only anyway

src/shared/libfido2-util.c
poettering: hmm? my point was to make all our config files/cmdline switches to accept the lowercase spelling only, as that's kinda what we do across thre tree for all our enum and enum-like settings

MkfsSion: Ok, I'll update the code.

PR #23107, opened 2022-04-17
man: recommend that .network or friends should have a numeric prefix

### LINKED ISSUE #23041: UseDelegatedPrefix doesn't enable DHCPPrefixDelegation
 > 250.4

**Used distribution**
 > Arch Linux

 > `5.17.1-arch1-1`

 > x86_64

**Issue Description**
The `systemd.network(5)` docs state the following about the `UseDelegatedPrefix` option of the `[DHCPv6]` section:
>UseDelegatedPrefix=
           When true (the default), the client will request the DHCPv6 server to delegate prefixes. If the server provides prefixes to be delegated, then subnets of the prefixes are assigned to the interfaces which enables DHCPPrefixDelegation=. See also the
           DHCPPrefixDelegation= setting in the [Network] section, settings in the [DHCPPrefixDelegation] section, and RFC 8415[19].

This issue is specifically about the sentence:
>If the server provides prefixes to be delegated, then subnets of the prefixes are assigned to the interfaces which enables DHCPPrefixDelegation=.

This sentence implies that, if a prefix is delegated, then `DHCPPrefixDelegation=yes` is set automatically. This is not correct, because `DHCPPrefixDelegation=yes` would also mean, that an address of that prefix is assigned to the respective interface, which doesn't happen. An address is assigned when explicitly setting `DHCPPrefixDelegation=yes`.

Thus I suspect that the docs are wrong and the reception of a prefix doesn't enable `DHCPPrefixDelegation`. Maybe the sentence means that a delegated prefix enables the use of the `DHCPPrefixDelegation` setting. In this case the wording should be adjusted IMO.

### LINKED ISSUE #23105: documentation: `systemd.network` files must include numeric prefix?

```
$ systemctl --version
systemd 250 (250.4-2-arch)
```

**Used distribution**

ArchLinux

aarch64

Save the following to `/usr/lib/systemd/network/some-name.network`:

```dosini
[Match]
Name=wwan0

[Network]
DHCP=yes
```

DHCP should be activated for `wwan0`.

DHCP was not activated for `wwan0`

**Additional notes**

It seems that using  `/usr/lib/systemd/network/90-some-name.network` instead works.

I think a numeric prefix is required but not documented..?

yuwata:
Closes #23105.
Closes #23041.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -2149,7 +2149,7 @@ Table=1234</programlisting></para>
         <listitem>
           <para>When true (the default), the client will request the DHCPv6 server to delegate
           prefixes. If the server provides prefixes to be delegated, then subnets of the prefixes are
-          assigned to the interfaces which enables <varname>DHCPPrefixDelegation=</varname>.
+          assigned to the interfaces that have <varname>DHCPPrefixDelegation=yes</varname>.
           See also the <varname>DHCPPrefixDelegation=</varname> setting in the [Network] section,
           settings in the [DHCPPrefixDelegation] section, and
           <ulink url="https://www.rfc-editor.org/rfc/rfc8415.html#section-6.3">RFC 8415</ulink>.


WhyNotHugo (approved):
:+1:

man/systemd.link.xml
```diff
@@ -32,15 +32,19 @@
...
-    <filename>.link</filename>; other extensions are ignored. All link files are collectively sorted
-    and processed in lexical order, regardless of the directories in which they live. However, files
-    with identical filenames replace each other. Files in <filename>/etc/</filename> have the highest
-    priority, files in <filename>/run/</filename> take precedence over files with the same name in
-    <filename>/usr/lib/</filename>. This can be used to override a system-supplied link file with a
+    <filename>/etc/systemd/network</filename>. All configuration files are collectively sorted and
+    processed in alphanumeric order, regardless of the directories in which they live. However, files
+    with identical filenames replace each other. It is recommended that each filename is prefixed with
+    a number (e.g. <filename>10-eth0.link</filename>). Otherwise, the default
+    <filename>.link</filename> files or generated by
```
bluca: "or those generated by"

bluca (approved):

yuwata:
@bluca Thank you for the review. The above point is addressed. Upgrading the green label.

PR #23117, opened 2022-04-19
virt: detect OpenStack instance

### LINKED ISSUE #23116: [OpenStack Nova]: systemd-detect-virt export none, not kvm
When the ARM system is created by OpenStack Nova, it's technically based on KVM, but `systemd-detect-virt` output is `none`, not `kvm`
> [root@kwephisprm76911 ~]$ systemd-detect-virt
> none

`systemd-detect-virt` output `kvm` when  the system is created by OpenStack Nova

the product name in `/sys/class/dmi/id/product_name` is `OpenStack Nova` when the system is created by OpenStack Nova. 
> [root@kwephisprm76911 ~]$ cat /sys/class/dmi/id/product_name
> OpenStack Nova

so that when it is detected that the file `/sys/class/dmi/id/product_name` contains the words `OpenStack Nova`, systemd-detect-virt output `kvm`

main

YoruStar:
Fixes #23116.

### DIFF
--- a/src/basic/virt.c
+++ b/src/basic/virt.c
@@ -156,6 +156,7 @@ static Virtualization detect_vm_dmi_vendor(void) {
                 Virtualization id;
         } dmi_vendor_table[] = {
                 { "KVM",                 VIRTUALIZATION_KVM       },
+                { "OpenStack",           VIRTUALIZATION_KVM       }, /* Detect OpenStack instance as KVM in non x86 architecture */
                 { "Amazon EC2",          VIRTUALIZATION_AMAZON    },
                 { "QEMU",                VIRTUALIZATION_QEMU      },
                 { "VMware",              VIRTUALIZATION_VMWARE    }, /* https://kb.vmware.com/s/article/1009458 */


poettering:
hmm, how come this wasn't caught by detect_vm_cpuid()? i.e. what does it report in that CPUID VM leaf?

src/basic/virt.c
```diff
@@ -156,6 +156,7 @@ static Virtualization detect_vm_dmi_vendor(void) {
                 Virtualization id;
         } dmi_vendor_table[] = {
                 { "KVM",                 VIRTUALIZATION_KVM       },
+                { "OpenStack Nova",      VIRTUALIZATION_KVM       },
```
poettering: https://cloudinit.readthedocs.io/en/latest/topics/datasources/openstack.html

Hmm, that suggests "OpenStack Compute" is also something used?

Does openstack document the used strings anywhere?

YoruStar:
> hmm, how come this wasn't caught by detect_vm_cpuid()? i.e. what does it report in that CPUID VM leaf?

CPUID is an x86 specific interfac, but the system I use is ARM

poettering:
ah, can you add that as a comment maybe?

src/basic/virt.c
YoruStar: The document says that in x86 architecture , `openstack Nova` and `openstack compute` can be used to determine whether it is openstack. 
But I'm not sure if arm architecture has a product name `OpenStack Compute`

YoruStar: It's better that use `OpenStack` instead of `OpenStack Nova` or `OpenStack Compute`

src/basic/virt.c
```diff
@@ -156,6 +156,7 @@ static Virtualization detect_vm_dmi_vendor(void) {
                 Virtualization id;
         } dmi_vendor_table[] = {
                 { "KVM",                 VIRTUALIZATION_KVM       },
+                { "OpenStack",           VIRTUALIZATION_KVM       }, /* Detect OpenStack instance as KVM in non x86 architectur */
```
bluca: typo: architectur -> architecture

YoruStar: fixed

poettering:
lgtm

poettering:
(btw, doesn't arm have some cpuid-like feature, where we can determine KVM support in a generic fashion from the CPU instead of DMI?)

PR #23118, opened 2022-04-19
nspawn: fix --ephemeral with --machine

bluca:
Follow-up for https://github.com/systemd/systemd/commit/2362fdde1bd4bf54772383ef29431f683729ba76

When --machine is specified with --ephemeral, no random suffix is added, so
the recently added assert would fail.

Add a top-level variable with the expected file name for nspawn files, and
compute it when the rest of the names are computed.


yuwata (approved):
LGTM.

poettering:
lgtm. thanks for fixing this so quickly.

bluca:
(pushed to fix typo in code comment `wont'` -> `won't`, nothing else)

PR #23121, opened 2022-04-19
compression: add separate pre-processor definitions

bluca:
Follow-up for https://github.com/systemd/systemd/commit/cd3c6322dbc6370448bafc216ee4e19e32a79d9e

journal-def.h should be self-contained too, as it represents the journal object ABI.
Duplicate the enums, as they also need to be in config.h for it to be self-contained,
and enums are not available to the preprocessor. Use an assert to ensure they don't
diverge.

### DIFF
--- a/src/libsystemd/sd-journal/journal-def.h
+++ b/src/libsystemd/sd-journal/journal-def.h
@@ -43,11 +43,20 @@ typedef enum ObjectType {
 } ObjectType;
 
 /* Object flags
- * The per-compression enums are defined in meson.build so that config.h is self-contained */
+ * The per-compression enums are also redefined in meson.build so that config.h is self-contained */
 enum {
+        OBJECT_COMPRESSED_XZ   = 1 << 0,
+        OBJECT_COMPRESSED_LZ4  = 1 << 1,
+        OBJECT_COMPRESSED_ZSTD = 1 << 2,
         OBJECT_COMPRESSION_MASK = (OBJECT_COMPRESSED_XZ | OBJECT_COMPRESSED_LZ4 | OBJECT_COMPRESSED_ZSTD),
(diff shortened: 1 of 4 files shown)


src/libsystemd/sd-journal/journal-def.h
```diff
@@ -43,12 +43,19 @@ typedef enum ObjectType {
...
+        OBJECT_COMPRESSED_XZ   = 1 << 0,
+        OBJECT_COMPRESSED_LZ4  = 1 << 1,
+        OBJECT_COMPRESSED_ZSTD = 1 << 2,
         OBJECT_COMPRESSION_MASK = (OBJECT_COMPRESSED_XZ | OBJECT_COMPRESSED_LZ4 | OBJECT_COMPRESSED_ZSTD),
         _OBJECT_COMPRESSED_MAX = OBJECT_COMPRESSION_MASK,
 };
 
+assert_cc(OBJECT_COMPRESSED_XZ == COMPRESSION_XZ);
+assert_cc(OBJECT_COMPRESSED_LZ4 == COMPRESSION_LZ4);
+assert_cc(OBJECT_COMPRESSED_ZSTD == COMPRESSION_ZSTD);
```
poettering, line 58: maybe move to compress.h?

poettering: or actually, maybe not.

bluca: done

bluca: lol - moving back then

yuwata, line 58: If you want to make the header self-contained, then maybe, we should add `#if defined(DEFAULT_COMPRESSION)` or similar here?

bluca: sure, done

PR #23122, opened 2022-04-19
tpm2: beef up tpm2 support checks

poettering:
This primarily adds `systemd-creds has-tpm2` as new command for reporting whether the system is equipped with a TPM2 device we can use.

Usecase: we eventually should update `kernel-install` to automatically augment installed kernel EFI images with credentials for things like the root pw (so that you can log into the initrd the same way as into the host), but we should only do so if we can encrypt/authenticate the password with a TPM2, and properly lock it to one specific kernel version via PCRs. Hence, with this we have a simple mechanism to do this, i.e. along these lines:

```sh
if systemd-creds has-tpm2 -q ; then
        getent shadow root | cut -d: -f 2 | systemd-creds --with-key=tpm2 encrypt - $TARGETDIR/passwd.hashedpassword.root
fi
```

### DIFF
--- a/man/systemd-creds.xml
+++ b/man/systemd-creds.xml
@@ -163,6 +163,20 @@
         and thus decryption is entirely automatic.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><command>has-tpm2</command></term>
+
+        <listitem><para>Reports whether the system is equipped with a TPM2 device usable for protecting
+        credentials. If the a TPM2 device has been discovered, is supported, and is being used by firmware,
+        by the OS kernel drivers and by userspace (i.e. systemd) this prints <literal>yes</literal> and exits
+        with exit status zero. If no such device is discovered/supported/used, prints
+        <literal>no</literal>. Otherwise prints <literal>partial</literal>. In either of these two cases
+        exits with non-zero exit status. It also shows three lines indicating separately whether drivers,
+        firmware and the system discovered/support/use TPM2.</para>
+
+        <para>Combine with <option>--quiet</option> to suppress the output.</para></listitem>
+      </varlistentry>
+
       <xi:include href="standard-options.xml" xpointer="help" />
       <xi:include href="standard-options.xml" xpointer="version" />
     </variablelist>
@@ -305,6 +319,14 @@
         <citerefentry><refentrytitle>systemd-cryptenroll</refentrytitle><manvolnum>1</manvolnum></citerefentry>.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--quiet</option></term>
+        <term><option>-q</option></term>
+
+        <listitem><para>When used with <command>has-tpm2</command> suppresses the output, and only returns an
+        exit status indicating support for TPM2.</para></listitem>
+      </varlistentry>
+
       <xi:include href="standard-options.xml" xpointer="no-pager" />
       <xi:include href="standard-options.xml" xpointer="no-legend" />
       <xi:include href="standard-options.xml" xpointer="json" />
@@ -315,6 +337,12 @@
     <title>Exit status</title>
 
     <para>On success, 0 is returned.</para>
+
+    <para>In case of the <command>has-tpm2</command> command returns 0 if a TPM2 device is discovered,
+    supported and used by firmware, driver, and userspace (i.e. systemd). Otherwise returns the OR
+    combination of the value 1 (in case firmware support is missing), 2 (in case driver support is missing)
+    and 4 (in case userspace support is missing). If no TPM2 support is available at all, value 7 is hence
+    returned.</para>
   </refsect1>
 
   <refsect1>


src/shared/tpm2-util.c
```diff
@@ -1453,3 +1455,24 @@ int tpm2_primary_alg_from_string(const char *alg) {
                 return TPM2_ALG_RSA;
         return -EINVAL;
 }
+
+Tpm2Support tpm2_support(void) {
+        Tpm2Support support = 0;
```
medhefgo: `= TPM2_SUPPORT_NONE`?

src/boot/bootctl.c
```diff
@@ -1723,7 +1724,15 @@ static int verb_status(int argc, char *argv[], void *userdata) {
                 printf("  Secure Boot: %sd (%s)\n",
                        enable_disable(IN_SET(secure, SECURE_BOOT_USER, SECURE_BOOT_DEPLOYED)),
                        secure_boot_mode_to_string(secure));
-                printf(" TPM2 Support: %s\n", yes_no(efi_has_tpm2()));
+
+                s = tpm2_support();
+                printf(" TPM2 Support: %s%s%s\n",
+                       FLAGS_SET(s, TPM2_SUPPORT_FIRMWARE|TPM2_SUPPORT_DRIVER) ? ansi_highlight_green() :
+                       (s & (TPM2_SUPPORT_FIRMWARE|TPM2_SUPPORT_DRIVER)) != 0 ? ansi_highlight_red() : ansi_highlight_yellow(),
```
medhefgo, line 1731: I think you have yellow and red reversed here or this should be `==`…

src/creds/creds.c
```diff
@@ -525,6 +526,31 @@ static int verb_setup(int argc, char **argv, void *userdata) {
...
+                printf("%sfirmware\n"
+                       "%sdriver\n"
+                       "%ssystem\n",
+                       plus_minus(s & TPM2_SUPPORT_FIRMWARE),
+                       plus_minus(s & TPM2_SUPPORT_DRIVER),
+                       plus_minus(s & TPM2_SUPPORT_SYSTEM));
+        }
+
+        /* Return inverted bit flags */
+        return ~s & 7;
```
medhefgo: I assume this is so that TPM2_SUPPORT_NONE does not signal success by returning 0? The comment should be more clear. Imho, this should use TPM2_SUPPORT_FULL too.

man/systemd-creds.xml
```diff
@@ -163,6 +163,20 @@
...
+
+        <listitem><para>Reports whether the system is equipped with a TPM2 device usable for protecting
+        credentials. If the a TPM2 device has been discovered, is supported, and is being used by firmware,
+        by the OS kernel drivers and by userspace (i.e. systemd) this prints <literal>yes</literal> and exits
+        with exit status zero. If no such device is discovered/supported/used, prints
+        <literal>no</literal>. Otherwise prints <literal>partial</literal>. In either of these two cases
+        exits with non-zero exit status. It also shows three lines indicating separately whether drivers,
+        firmware and the system discovered/support/use TPM2.</para>
+
+        <para>Combine with <option>-q</option> to suppress the output.</para></listitem>
```
medhefgo: Do we prefer long or short names when suggesting other cmdline options?

src/boot/bootctl.c
poettering: nope, this was done on purpose. i.e. having partial tpm support is a weird thing. having no tpm support is quite ok. hence i highlighted the partial (i.e. kinda broken) support as red, but the tpm being absent just as yellow, because it is noteworthy but kinda ok if your platform doesn't have it.

src/creds/creds.c
bluca: yeah the magic number is not great here

man/systemd-creds.xml
poettering: actually, long ones.

poettering:
Force pushed a new version. PTAL!

src/shared/tpm2-util.c
```diff
@@ -1453,3 +1455,24 @@ int tpm2_primary_alg_from_string(const char *alg) {
                 return TPM2_ALG_RSA;
         return -EINVAL;
 }
+
+Tpm2Support tpm2_support(void) {
+        Tpm2Support support = TPM2_SUPPORT_NONE;
```
bluca, line 1460: Is it work caching this locally? Result won't change during execution, right?

bluca (approved):

src/shared/tpm2-util.c
poettering: so we already cache the firmware check in efi_has_tpm2(), so we cache *some* thing.

The driver check we don't cache, because initially at boot there's a good chance the drivers aren't loaded yet, but later they will. Hence it's only safe to check each time, and thus not cache.

Hence: i think we already cache as much as makes sense here.

medhefgo (approved):
Is it just me or does github not highlight whitespace only changes anymore? (So that re-arrangements of tables only have the new whitespace highlighted)

medhefgo:
That one is about hiding whitespace-only changes. What I am talking about is keeping the whitespace changes visible, but have the new or removed spaces highlighted like with any other printable character.

PR #17780, opened 2020-11-30
udev-builtin-blkid: add support for --hint offsets (v2)

karelzak:
The next libblkid v2.37 is going to support session offsets for
multi-session CD/DVDs. This feature is implemented by "hint offsets"
these offsets are optional and prober specific (e.g., iso,   udf,
...). For this purpose, the library provides a new function
blkid_probe_set_hint(), and blkid(8) provides a new command-line
option --hint <name>=<offset>. For CD/DVD, the offset name is
"session_offset".

The difference between classic --offset and the new --hint is that --offset
is very restrictive and defines the probing area and the rest of the
the device is invisible to the library. The new --hint works like a
suggestion, it provides a hint where the user assumes the filesystem, but
the rest of the device is still readable for the library (for example, to
get some additional superblock information etc.).

The option --hint implementation in udev-builtin-blkid.c is backwardly
compatible. If compiled against old libblkid, then the option is used in
the same way as --offset.

Addresses: https://github.com/karelzak/util-linux/issues/1161
Addresses: https://github.com/systemd/systemd/pull/17424

(Note the udev change is not tested yet. We have verified the change in libbkid, but it would be nice to re-test it with udev too.)

PR #23127, opened 2022-04-19
fix 60-persistent-storage.rules for /dev/sr*

jcfrosty:
It appears the IMPORT lines affecting /dev/sr* drive rules can cause udev to hang when putting certain discs in the optical drive. After udev hangs for a while and times out, it kills /dev/sr* . watching/ripping DVDs and Blurays would break when entering a disc that didn't agree with the use of those IMPORT lines requiring reboot. This could also hang dockers and create unkillable zombie processes that were using /dev/sr* devices.

This has also been recognized in the makemkv community HERE: https://forum.makemkv.com/forum/viewtopic.php?t=25357

I've experience this in Unraid, Fedora, and Manjaro and have found all were affected and fixed using this suggested change.

### DIFF
--- a/rules.d/60-persistent-storage.rules
+++ b/rules.d/60-persistent-storage.rules
@@ -101,10 +101,10 @@ KERNEL=="vd*[0-9]", ENV{ID_PATH}=="pci-*", SYMLINK+="disk/by-path/virtio-$env{ID
 
 # probe filesystem metadata of optical drives which have a media inserted
 KERNEL=="sr*", ENV{DISK_EJECT_REQUEST}!="?*", ENV{ID_CDROM_MEDIA_TRACK_COUNT_DATA}=="?*", ENV{ID_CDROM_MEDIA_SESSION_LAST_OFFSET}=="?*", \
-  IMPORT{builtin}="blkid --hint=session_offset=$env{ID_CDROM_MEDIA_SESSION_LAST_OFFSET}"
+  GOTO="persistent_storage_end"
 # single-session CDs do not have ID_CDROM_MEDIA_SESSION_LAST_OFFSET
 KERNEL=="sr*", ENV{DISK_EJECT_REQUEST}!="?*", ENV{ID_CDROM_MEDIA_TRACK_COUNT_DATA}=="?*", ENV{ID_CDROM_MEDIA_SESSION_LAST_OFFSET}=="", \
-  IMPORT{builtin}="blkid --noraid"
+  GOTO="persistent_storage_end"
 
 # probe filesystem metadata of disks
 KERNEL!="sr*|mmcblk[0-9]boot[0-9]", IMPORT{builtin}="blkid"


jcfrosty:
Just so we can all see what the issue looks like before the rule change:

```
[  324.178091] usb 4-1: reset SuperSpeed USB device number 2 using xhci_hcd
[  324.190405] sr 0:0:0:0: [sr0] tag#0 UNKNOWN(0x2003) Result: hostbyte=0x03 driverbyte=DRIVER_OK cmd_age=30s
[  324.190410] sr 0:0:0:0: [sr0] tag#0 CDB: opcode=0x28 28 00 00 00 04 00 00 00 02 00
[  324.190412] blk_update_request: I/O error, dev sr0, sector 4096 op 0x0:(READ) flags 0x80700 phys_seg 1 prio class 0
[  354.383938] usb 4-1: reset SuperSpeed USB device number 2 using xhci_hcd
[  473.812944] udevd[1688]: worker [27853] /devices/pci0000:00/0000:00:08.1/0000:03:00.3/usb4/4-1/4-1:1.0/host0/target0:0:0/0:0:0:0/block/sr0 timeout; kill it
[  473.812954] udevd[1688]: seq 11256 '/devices/pci0000:00/0000:00:08.1/0000:03:00.3/usb4/4-1/4-1:1.0/host0/target0:0:0/0:0:0:0/block/sr0' killed
```

poettering:
I am sorry, what?

The disk apparently has an IO error, and readig then hangs. and udev handles it quite OK by killing the udev worker after a while? so all seems to be in order?

So where's the problem precisely? I mean, the hw/disk is faulty apparently?

If the kernel hangs here without timeout on its own, then this is ugly, and probably should be fixed in kernel but udev appears to really nicely handle this.

marcosfrm:
It would undo https://github.com/systemd/systemd/pull/17780

jcfrosty:
@poettering 

Its NOT hanging in the kernel. Same disc, remove IMPORT lines, I can insert DVD disc and watch the video just fine without udev killing /dev/sr0. I still have complete control of /dev/sr0 and can eject the drive etc, no zombie processes or hung tasks that were hooked up to /dev/sr0. This is 100% an issue with how the udev rules are handling /dev/sr0 after entering a disc.

poettering:
> Its NOT hanging in the kernel. 

Well, the blkid builtin just does a couple of ioctls+read()s, and that's it. If that hangs, then this suggests a kernel issue, and your dmesg suggest the same.

> Same disc, remove IMPORT lines,

Yes, this might stop triggering the issue, but the fact is that there appears to be some IO issues with your disk...

>  I can insert DVD disc and watch the video just fine without udev killing /dev/sr0. 

I am not sure what "killing" /dev/sr0 is even supposed to mean...

> I still have complete control of /dev/sr0 and can eject the drive etc, no zombie processes or hung tasks that were hooked up to /dev/sr0. This is 100% an issue with how the udev rules are handling /dev/sr0 after entering a optical disc.

hung tasks are kernel issues, not userspace issues. udev's blkid builtin might trigger the issue, but it's not the culprit.

jcfrosty:
@poettering That make sense, I'm following you now. as for "killing" terminology I'm referring to this line in which /dev/sr0 become unresponsive until after reboot:

```
udevd[1688]: worker [27853] /devices/pci0000:00/0000:00:08.1/0000:03:00.3/usb4/4-1/4-1:1.0/host0/target0:0:0/0:0:0:0/block/sr0 timeout; kill it
udevd[1688]: seq 11256 '/devices/pci0000:00/0000:00:08.1/0000:03:00.3/usb4/4-1/4-1:1.0/host0/target0:0:0/0:0:0:0/block/sr0' killed
```

karelzak:
What medium is it? I guess encrypted movie DVD in an external USB drive.

 The problem sounds like this:

https://bugzilla.kernel.org/show_bug.cgi?id=207327
https://bugzilla.redhat.com/show_bug.cgi?id=1825533

and if you really have time:
https://bugzilla.redhat.com/show_bug.cgi?id=1822948

https://bugzilla.redhat.com/show_bug.cgi?id=1822948#c142


If I good understand the problem, then blkid reads areas that are not readable by read(), and in some cases, the kernel reacts by hand. So, blkid should be more careful, and the kernel should be more robust ...

@pali, any idea?

pali:
You need to look which sectors is blkid trying to read and compare it with layout of that problematic medium. Maybe medium is broken and reports bogus data about layout? Hard to guess.

poettering:
I am pretty sure there's nothing to fix in systemd here though. The kernel shouldn't allow hangs like that. Now I sense the kernel dvd stack might not get the love anymore it deserves, but even then this should probably dealt with in blkid and not systemd i guess. 

Anyway, I see nothing actionable here in systemd/udevd. Let's close hence. @karelzak posted links to other bug trackers where the issue is tracked, so please direct further comments there.

PR #23133, opened 2022-04-20
Add a short page with a video showing sd-boot

keszybz:
I have no idea how to effectively preview this. The video is 4.5 MB, so we should probably not merge this until we're fairly sure that this is wanted.

### DIFF
--- a/docs/SYSTEMD-BOOT.md
+++ b/docs/SYSTEMD-BOOT.md
@@ -1,0 +1,25 @@
+---
+title: systemd-boot
+category: Booting
+layout: default
+SPDX-License-Identifier: LGPL-2.1-or-later
+---
+
+# systemd-boot
+
+`systemd-boot` (short: `sd-boot`) is a simple UEFI boot manager. It provides a
+textual menu to select the entry to boot and an editor for the kernel command
+line. `systemd-boot` supports systems with UEFI firmware only.
+
+See the man pages for details:
+[systemd-boot(7)](https://www.freedesktop.org/software/systemd/man/systemd-boot.html),
+[bootctl(1)](https://www.freedesktop.org/software/systemd/man/bootctl.html).
+
+<video src='/assets/sd-boot.webm' width=180/>
+
+## Links
+
+[`systemd-boot(7)`](https://www.freedesktop.org/software/systemd/man/systemd-boot.html)<br>
+[`bootctl(1)`](https://www.freedesktop.org/software/systemd/man/bootctl.html)<br>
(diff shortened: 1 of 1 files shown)


keszybz:
https://github.com/keszybz/sd-boot-video has the source components.

keszybz:
I agree that a simpler kernel version would be nice. But, pfff, regenerating this from scratch would be a pain, all timestamps would be off.

keszybz:
I want to extend the page to say how sd-boot and bootctl and the other APIs tie together. I'll rework this.

behrmann:
In other screencasts people sometimes overlay what buttons are pressed. This probably can't be done in sd-boot, so it would need to be added in post production (which might be more work than this is worth), but I find this always a nice alternative to voiceover.

keszybz:
It's already there.

behrmann:
Ah, sorry, I had looked at the video you in the second post, assuming they were the same. 

A few nits, though:
- "This entry is autogenerated" the "this" means the "Reboot Into Firmware Interface" entry, right? This only appears for a very short time, so it might not be immediately clear, whether this entry is meant (when looking at the subtitle) or any other
- "Macintosh and Microsoft systems are also autodetected", sounds a bit oldtimey. How about "MacOS and Windows are also autodetected"?
- "There are two "Fedora Linux 36" entries, kernel version the is shown to disambiguate", since that is mostly a full sentence already it might be better as "There are two "Fedora Linux 36" entries, *the* kernel version the is shown to disambiguate" or it can be shortened to not be a full sentence anymore "wo "Fedora Linux 36" entries, kernel version the is shown to disambiguate". The previous sentence about the Fedora 34 could maybe also be taken off that screen, to have fewer subtitles on the screen.
- "When menu is disabled, any key held when sd-boot activated will open the menu", maybe longer as "When *the* menu is disabled, any key held when sd-boot activated will open the menu" and maybe with a gerund instead of a passive construction "When (the) menu is disabled, holding any key when sd-boot activates will open the menu". This could also be shortened by leaving out "when sd-boot activates", since this should be pretty clear from context. 
- In the description of `e` "a bare-bones" instead of "an bare-bones", also "Emacs" instead of "emacs" since it is a proper name. I'd leave out the `M-b`, `M-f` and the rest of the keybindings, either people know or they don't, but this is probably not the place to teach them (also I am a heathen and prefer `C-<left>`/`C-<right>` ;) )

All in all, great work! Something really nice to show people.

flokli:
There's also a typo on the `a - boot Micintosh` (at 71s)

Issue #23120, opened 2022-04-19
Improving the touchpad precision vaio FE14


davijosw:
**systemd version the issue has been seen with**
 > 249.11-1.fc35

**Used distribution**
 > Fedora 35

**Linux kernel version used** (`uname -a`)

 > `5.16.20-200.fc35`

**CPU architecture issue was seen on**
 > Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz - 64bits

I'm doubt if the hwdb file has to be a pull request here or in libinut gitlab. I don't know if the two projects are managed together, like two brothers in the same family hahah. Anyway, if it is a make a pull request asap and modify the file to be more general for other users. Thanks. 

There's the file.
[61-evdev-local.hwdb.txt](https://github.com/systemd/systemd/files/8513715/61-evdev-local.hwdb.txt)

I'm based on `libinput measure touchpad-size` tool.

poettering:
/cc @whot 

please submit as proper PR

PR #23136, opened 2022-04-20
Update 60-evdev.hwdb

davijosw:
By the libinput measure touchpad-size tool, I identify that the touchpad's precision needs to be improved. If some adjust is in need let me know,

### DIFF
--- a/hwdb.d/60-evdev.hwdb
+++ b/hwdb.d/60-evdev.hwdb
@@ -743,4 +743,3 @@ evdev:input:b0003v172Fp0047*
 evdev:input:b0003v172Fp0505*
  EVDEV_ABS_00=::160
  EVDEV_ABS_01=::160
- 


davijosw:
https://github.com/systemd/systemd/issues/23120

hwdb.d/60-evdev.hwdb
```diff
@@ -732,3 +732,13 @@ evdev:input:b0003v172Fp0047*
...
+ 
+#########################################
+# Positivo-Vaio
+#########################################
+# Vaio FE14
+evdev:name:SYNA3602:00 0911:5288 Touchpad:dmi:bvn*:bvr*:bd*:svnPositivoBahia-VAIO:pnVJFE41F11*:pvr*
+ EVDEV_ABS_00=::28
+ EVDEV_ABS_01=::27
+ EVDEV_ABS_35=::28
+ EVDEV_ABS_36=::27
```
whot: This file is ordered alphabetically by vendor, can you move this block up please to where it belongs, thanks

hwdb.d/60-evdev.hwdb
```diff
@@ -732,3 +732,13 @@ evdev:input:b0003v172Fp0047*
 evdev:input:b0003v172Fp0505*
  EVDEV_ABS_00=::160
  EVDEV_ABS_01=::160
+ 
+#########################################
+# Positivo-Vaio
+#########################################
+# Vaio FE14
+evdev:name:SYNA3602:00 0911:5288 Touchpad:dmi:bvn*:bvr*:bd*:svnPositivoBahia-VAIO:pnVJFE41F11*:pvr*
```
whot:
```suggestion
evdev:name:SYNA3602:00 0911:5288 Touchpad:dmi:*svnPositivoBahia-VAIO:pnVJFE41F11*
```
No need to specify all the others with globs.

whot (changes requested):

hwdb.d/60-evdev.hwdb
davijosw: Hey, I made a pull request on the hwdb keyboard file and this particular line was pretty much the same (with globs). Do I have to make a new pull request to correct this aspect? Or the keyboard and the evdev are differents in the system reading aspect?

whot: hwdb files are all parsed by the same parser, if that's what you're referring to. Whether you specify all sub-components is just convention, looks like `60-keyboard.hwdb` primarily uses the long forma, `60-evdev.hwbd` primarily uses the shortened format. I think it's better to use the short version here for consistency (plus: it's less visual detritus).

whot:
merged (with fixed commit message), thanks!

hwdb.d/60-evdev.hwdb
davijosw: thanks!

PR #23137, opened 2022-04-20
terminal-util: get_color_mode checks COLORTERM

cerebro1:
TODO : https://github.com/systemd/systemd/blob/main/TODO#L134
**get_color_mode() should probably check the $COLORTERM environment variable which most terminal environments appear to set.**
This mentions https://github.com/termstandard/colors#checking-for-colorterm
that `The S-Lang library has a check that $COLORTERM contains either "truecolor" or "24bit" (case sensitive).`

Parsed COLORTERM.

### DIFF
--- a/src/basic/terminal-util.c
+++ b/src/basic/terminal-util.c
@@ -1278,6 +1278,11 @@ ColorMode get_color_mode(void) {
                         /* We only check for the presence of the variable; value is ignored. */
                         cached_color_mode = COLOR_OFF;
 
+                else if (STRPTR_IN_SET(getenv("COLORTERM"),
+                                       "truecolor",
+                                       "24bit"))
+                        cached_color_mode = COLOR_24BIT;
+
                 else if (getpid_cached() == 1)
                         /* PID1 outputs to the console without holding it open all the time.
                          *
--- a/src/basic/terminal-util.h
+++ b/src/basic/terminal-util.h
@@ -100,16 +100,19 @@ typedef enum AcquireTerminalFlags {
 /* Limits the use of ANSI colors to a subset. */
 typedef enum ColorMode {
         /* No colors, monochrome output. */
-        COLOR_OFF = 0,
+        COLOR_OFF,
 
         /* All colors, no restrictions. */
-        COLOR_ON  = 1,
+        COLOR_ON,
 
         /* Only the base 16 colors. */
-        COLOR_16  = 16,
+        COLOR_16,
 
         /* Only 256 colors. */
-        COLOR_256 = 256,
+        COLOR_256,
+
+        /* For truecolor or 24bit color support.*/
+        COLOR_24BIT,
 
         _COLOR_INVALID = -EINVAL,
 } ColorMode;


cerebro1:
@bluca @anitazha @DaanDeMeyer Is the change as expected?

DaanDeMeyer:
I'm not familiar with the terminal logic so I can't say for sure, let's wait for @poettering to take a look.

src/basic/terminal-util.c
```diff
@@ -1278,6 +1278,9 @@ ColorMode get_color_mode(void) {
                         /* We only check for the presence of the variable; value is ignored. */
                         cached_color_mode = COLOR_OFF;
 
+                else if (getenv("COLORTERM"))
```
yuwata: Hm? If I believe your comment, it has "truecolor" or "24bit". Why not to parse it and set it to COLOR_16777216 or COLOR_24bit?

src/basic/terminal-util.h
```diff
@@ -111,6 +111,10 @@ typedef enum ColorMode {
         /* Only 256 colors. */
         COLOR_256 = 256,
 
+        /* COLORTERM default value truecolor for terminator, 16,777,216 colors. 24-bit color Value. */
+        /* Other Possible value gnome-terminal */
+        COLOR_TERM = 24,
```
yuwata: How about to rename `COLOR_TERM` -> `COLOR_16777216` or `COLOR_24BIT`?
And, the assigned value should be 16777216, instead of 24.

src/basic/terminal-util.h
```diff
@@ -111,6 +111,10 @@ typedef enum ColorMode {
         /* Only 256 colors. */
         COLOR_256 = 256,
 
+        /* COLORTERM default value truecolor for terminator, 16,777,216 colors. 24-bit color Value. */
+        /* Other Possible value gnome-terminal */
```
yuwata: I am not native, but the comment sounds strange for me.

At least, please merge the two lines, that is, drop the slashes at the end of the first line, and the beginning of the second line.

yuwata (changes requested):

src/basic/terminal-util.c
cerebro1: Agreed

src/basic/terminal-util.h
cerebro1: Sure

src/basic/terminal-util.h
cerebro1: Updated comment in sync with previous comment for other values.

cerebro1:
Changes done.

poettering:
So, when I added this to the TODO list, this was meant as something to figure out the details of first, i.e. whether we actually have any benefit if we'd start bothering about that env var.

I think it might be unnecessary, since it indicaes 24bit color support in terminals, which however we do not use. Moreover, we actually use 256 color support by default, unless there is indication that it's not going to work. i.e. we are looking for indications that things are "bad", not that things are "good", i.e. we do the 256 color support in an opt-out fashion, not an opt-in fashion.

Hence it might not actually be a useful thing for us to check, because it gets us nothing... Dunno. 

@keszybz last touched this code, do you have a take on this?

src/basic/terminal-util.c
keszybz: Hmm, you said "agreed", but the variable is still not being parsed ??

src/basic/terminal-util.h
```diff
@@ -111,6 +111,9 @@ typedef enum ColorMode {
         /* Only 256 colors. */
         COLOR_256 = 256,
 
+        /* Only 16,777,216 colors. */
+        COLOR_24BIT = 16777216,
```
keszybz: I know that 16777216 was requested here, but this is not useful. Instead, drop the value assignments for COLOR_OFF, COLOR_ON, COLOR_16, COLOR_256 and COLOR_24BIT so that the compiler can assign 0, 1, 2, 3, 4 automatically.

Also "*only* 16 mil colors" is not a useful comment. It was copied from the comment above, which was copied from the comment about 16 colors, where it made sense. Please adjust this also.

keszybz (changes requested):
I think it's useful to add support for COLORTERM. If it is set to "truecolor" or "24bit", this gives us a strong indication that the terminal has full color support.

src/basic/terminal-util.c
cerebro1: Needed an input here actually but did not ask. When parsed, i need validate it against "truecolor" and "24bit", right?

src/basic/terminal-util.h
cerebro1: ok sure.

src/basic/terminal-util.c
cerebro1: Adding now. Missed to ask value to validate against. Going with "truecolor" and "24bit"

src/basic/terminal-util.h
cerebro1: > I know that 16777216 was requested here, but this is not useful. Instead, drop the value assignments for COLOR_OFF, COLOR_ON, COLOR_16, COLOR_256 and COLOR_24BIT so that the compiler can assign 0, 1, 2, 3, 4 automatically.

Dropped for all mentioned.

cerebro1:
Just realised that variable declaration has to be done at the beginning of function and then later can be used in function. 
`const char *colorterm;`
Missed CODING_STYLE. I will make that change in next commit with any other changes that are supposed to go after review. @bluca @keszybz @yuwata please review. 
Thanks

src/basic/terminal-util.c
```diff
@@ -1272,12 +1272,16 @@ ColorMode get_color_mode(void) {
...
         if (cached_color_mode < 0) {
                 m = parse_systemd_colors();
+                const char *colorterm = getenv("COLORTERM");
                 if (m >= 0)
                         cached_color_mode = m;
                 else if (getenv("NO_COLOR"))
                         /* We only check for the presence of the variable; value is ignored. */
                         cached_color_mode = COLOR_OFF;
 
+                else if (colorterm != NULL && (streq(colorterm, "truecolor") || streq(colorterm, "24bit")))
```
bluca: you can use `streq_ptr` and drop the manual null check

keszybz: … or even use `STRPTR_IN_SET` !

keszybz: We don't want to call `getenv` unless we're going to use the return value. It needs to search the array of variables, so it is "expensive".

So this should be like this:
```C
... else if (STRPTR_IN_SET(getenv("COLORTERM"),
               "truecolor",
               "24bit")) 

        ...
```

keszybz: We don't want to call `getenv` unless we're going to use the return value. It needs to search the array of variables, so it is "expensive".

So this should be like this:
```C
     ...
     else if (STRPTR_IN_SET(getenv("COLORTERM"),
               "truecolor",
               "24bit")) 

        ...
```

cerebro1: Thanks for sharing the util. Updated the commit. I will though understand how STRPTR_IN_SET or strreq_ptr better suited here.

src/basic/terminal-util.c
```diff
@@ -1278,6 +1278,11 @@ ColorMode get_color_mode(void) {
                         /* We only check for the presence of the variable; value is ignored. */
                         cached_color_mode = COLOR_OFF;
 
+                else if (STRPTR_IN_SET(getenv("COLORTERM"),
+                        "truecolor",
+                        "24bit"))
```
keszybz: The code is OK, but the alignment is wrong (misleading). `"truecolor"` needs to be aligned to the same column that `getenv`, because those are arguments to the same macro call.

cerebro1: ok. Done.

bluca (approved):

PR #23140, opened 2022-04-21
logind: add support for hidraw devices

whot:
Add support for opening /dev/hidraw devices via logind's TakeDevice().
Same semantics as our support for evdev devices, but it requires the
HIDIOCREVOKE ioctl in the kernel.

Todo:
- [ ] bit unsure about the udev rule change for the seat tag and whether that pulls in more than we need

cc @hadess

- Kernel repo for testing: https://github.com/whot/linux/tree/wip/hid-revoke
- Patch on LKML: https://patchwork.kernel.org/project/linux-input/patch/YmEAPZKDisM2HAsG@quokka/
- Tool to verify kernel works: https://github.com/whot/hidiocrevoke-test

### DIFF
--- a/src/login/logind-session-device.c
+++ b/src/login/logind-session-device.c
@@ -15,6 +15,7 @@
 #include "logind-session-device.h"
 #include "missing_drm.h"
 #include "missing_input.h"
+#include "missing_hidraw.h"
 #include "parse-util.h"
 #include "util.h"
 
@@ -103,6 +104,20 @@ static void sd_eviocrevoke(int fd) {
         }
 }
 
+static void sd_hidiocrevoke(int fd) {
+        static bool warned = false;
+
+        assert(fd >= 0);
+
+        if (ioctl(fd, HIDIOCREVOKE, NULL) < 0) {
+
+                if (errno == EINVAL && !warned) {
+                        log_warning_errno(errno, "Kernel does not support hidraw-revocation: %m");
+                        warned = true;
+                }
+        }
+}
+
 static int sd_drmsetmaster(int fd) {
         assert(fd >= 0);
         return RET_NERRNO(ioctl(fd, DRM_IOCTL_SET_MASTER, 0));
@@ -145,6 +160,11 @@ static int session_device_open(SessionDevice *sd, bool active) {
         case DEVICE_TYPE_EVDEV:
                 if (!active)
                         sd_eviocrevoke(fd);
+                break;
+
+        case DEVICE_TYPE_HIDRAW:
+                if (!active)
+                        sd_hidiocrevoke(fd);
                 break;
 
         case DEVICE_TYPE_UNKNOWN:
@@ -190,6 +210,17 @@ static int session_device_start(SessionDevice *sd) {
                 CLOSE_AND_REPLACE(sd->fd, r);
                 break;
 
+        case DEVICE_TYPE_HIDRAW:
+                /* Hidraw devices are revoked while inactive. Reopen it and we are fine. */
+                r = session_device_open(sd, true);
+                if (r < 0)
+                        return r;
+
+                /* For hidraw devices, the file descriptor might be left uninitialized. This might happen while resuming
+                 * into a session and logind has been restarted right before. */
+                CLOSE_AND_REPLACE(sd->fd, r);
+                break;
+
         case DEVICE_TYPE_UNKNOWN:
         default:
                 /* fallback for devices without synchronizations */
@@ -229,6 +260,14 @@ static void session_device_stop(SessionDevice *sd) {
                 sd_eviocrevoke(sd->fd);
                 break;
 
+        case DEVICE_TYPE_HIDRAW:
+                /* Revoke access on hidraw file-descriptors during deactivation.
+                 * This will basically prevent any operations on the fd and
+                 * cannot be undone. Good side is: it needs no CAP_SYS_ADMIN
+                 * protection this way. */
+                sd_hidiocrevoke(sd->fd);
+                break;
+
         case DEVICE_TYPE_UNKNOWN:
         default:
                 /* fallback for devices without synchronization */
@@ -252,6 +291,9 @@ static DeviceType detect_device_type(sd_device *dev) {
         } else if (streq(subsystem, "input")) {
                 if (startswith(sysname, "event"))
                         type = DEVICE_TYPE_EVDEV;
+        } else if (streq(subsystem, "hidraw")) {
+                if (startswith(sysname, "hidraw"))
+                        type = DEVICE_TYPE_HIDRAW;
         }
 
         return type;
@@ -285,7 +327,7 @@ static int session_device_verify(SessionDevice *sd) {
                 if (sd_device_get_syspath(dev, &sp) < 0)
                         return -ENODEV;
 
-        } else if (sd->type != DEVICE_TYPE_DRM)
+        } else if (sd->type != DEVICE_TYPE_DRM && sd->type != DEVICE_TYPE_HIDRAW)
                 /* Prevent opening unsupported devices. Especially devices of
                  * subsystem "input" must be opened via the evdev node as
                  * we require EVIOCREVOKE. */
--- a/src/basic/meson.build
+++ b/src/basic/meson.build
@@ -138,6 +138,7 @@ basic_sources = files(
         'missing_drm.h',
         'missing_fcntl.h',
         'missing_fs.h',
+        'missing_hidraw.h',
         'missing_input.h',
         'missing_ioprio.h',
         'missing_keyctl.h',
--- a/src/basic/missing_hidraw.h
+++ b/src/basic/missing_hidraw.h
@@ -1,0 +1,9 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+#pragma once
+
+#include <linux/hidraw.h>
+
+/* Proposed, pending */
+#ifndef HIDIOCREVOKE
+#define HIDIOCREVOKE _IOW('H', 0x0D, int)
+#endif
--- a/src/login/71-seat.rules.in
+++ b/src/login/71-seat.rules.in
@@ -12,6 +12,7 @@ ACTION=="remove", GOTO="seat_end"
 TAG=="uaccess", SUBSYSTEM!="sound", TAG+="seat"
 SUBSYSTEM=="sound", KERNEL=="card*", TAG+="seat"
 SUBSYSTEM=="input", KERNEL=="input*", TAG+="seat"
+SUBSYSTEM=="hidraw", KERNEL=="hidraw*", TAG+="seat"
 SUBSYSTEM=="graphics", KERNEL=="fb[0-9]*", TAG+="seat"
 
 # Assign keyboard and LCD backlights to the seat
--- a/src/login/logind-session-device.h
+++ b/src/login/logind-session-device.h
@@ -11,6 +11,7 @@ enum DeviceType {
         DEVICE_TYPE_UNKNOWN,
         DEVICE_TYPE_DRM,
         DEVICE_TYPE_EVDEV,
+        DEVICE_TYPE_HIDRAW,
 };
 
 struct SessionDevice {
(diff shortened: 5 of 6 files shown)


keszybz:
Looks reasonable.

About the process: I think we should wait for the reviews of the kernel patch. If it looks likely to be merged, we can merge the systemd side. We don't need for a released kernel.

hadess:
I had already looked at the event revoke code to start implementing USB revoke which I never finished, and there was always something somewhat unclear to me.

Am I correct in thinking that the revoking code only revokes access to a single fd, meaning that all the fd we want revoked need to be obtained through logind's API? That's potentially going to be a lot of code to port to that behaviour, but necessary...

If that's the case, does the current patchset have the unintended consequence of making all hidraw devices accessible through logind? I'm guessing this is where we would need to add some "authorisation" code, as @poettering has mentioned in the past, and we would also need to remove all the direct "uaccess" from hidraw devices that are intended to be shared between seats.

> Tool to verify kernel works: https://github.com/whot/hidiocrevoke-test

Maybe making the test use 2 separate processes would show the feature better?

whot:
> Am I correct in thinking that the revoking code only revokes access to a single fd, meaning that all the fd we want revoked need to be obtained through logind's API? That's potentially going to be a lot of code to port to that behaviour, but necessary...

Correct. The end-game will (sadly) be: logind->mutter->impl.portal->portal->application, with happy fd passing through. logind would keep a copy of the fd and so will the portal so either can revoke, either on VT switch or whatever else the portal does.

>  the unintended consequence of making all hidraw devices accessible through logind?

yes, but logind's `TakeDevice` only opens devices for the controlling process of the session. You can't just call it over DBus and get an fd (otherwise testing would've been a lot easier :)

hadess:
> Correct. The end-game will (sadly) be: logind->mutter->impl.portal->portal->application, with happy fd passing through.

At least with USB you could patch libusb, or libgusb and cover a large number of apps...

> logind would keep a copy of the fd and so will the portal so either can revoke, either on VT switch or whatever else the portal does.

Could this have an impact on some devices that might only start spewing data when the device is open?

> yes, but logind's `TakeDevice` only opens devices for the controlling process of the session. You can't just call it over DBus and get an fd (otherwise testing would've been a lot easier :)

OK, then would we want to figure out a way to tag devices that mutter should ask authorisation for, ones it should offer without any auth, and (maybe) others it shouldn't ever allow opening?

whot:
> At least with USB you could patch libusb, or libgusb and cover a large number of apps...

Doesn't cover Bluetooth HID devices though? I'd say revoke for USB is largely independent. For hidraw what we will likely need is a `LD_PRELOAD`-capable wrapper that routes `open("/dev/hidraw")` through whichever portal we want to use.

> Could this have an impact on some devices that might only start spewing data when the device is open?

I don't think there's significant difference to the current situation. logind only opens the device on `TakeDevice` and you're supposed to `ReleaseDevice` when you're done. The delay between the actual `open()` and the application receiving the fd means more events will be queued up than with a direct `open()` call by the application but I doubt this matters much.

> OK, then would we want to figure out a way to tag devices that mutter should ask authorisation for, ones it should offer without any auth, and (maybe) others it shouldn't ever allow opening?

Yes, we need to figure that out, though I think this is largely a portal policy decision, like (thought bubble) `portal.GameMode` allowing any `ID_INPUT_JOYSTICK` hidraw device for a registered game.

hadess:
> Doesn't cover Bluetooth HID devices though? I'd say revoke for USB is largely independent.

It was supposed to be a comment on the amount of work needed for integration, you took it as a comment on this PR ;)

> For hidraw what we will likely need is a `LD_PRELOAD`-capable wrapper that routes `open("/dev/hidraw")` through whichever portal we want to use.

Doing blocking D-Bus'y things in single-threaded apps is definitely just a stop-gap.

> I don't think there's significant difference to the current situation. logind only opens the device on `TakeDevice` and you're supposed to `ReleaseDevice` when you're done. The delay between the actual `open()` and the application receiving the fd means more events will be queued up than with a direct `open()` call by the application but I doubt this matters much.

OK.

> Yes, we need to figure that out, though I think this is largely a portal policy decision, like (thought bubble) `portal.GameMode` allowing any `ID_INPUT_JOYSTICK` hidraw device for a registered game.

I was thinking about something at the logind/udev/hwdb level, because I don't think we want gnome-shell to be 100% responsible for gatekeeping certain hidraw devices.

bentiss:
> > For hidraw what we will likely need is a `LD_PRELOAD`-capable wrapper that routes `open("/dev/hidraw")` through whichever portal we want to use.
> 
> Doing blocking D-Bus'y things in single-threaded apps is definitely just a stop-gap.
> 

Chiming in there.

There is an alternative/complement for the new ioctl with eBPF. See [this repo](https://gitlab.freedesktop.org/bentiss/logind-hidraw/) for an example.
Basically, we need an helper in the kernel that is marked as `__weak` and `ALLOW_ERROR_INJECTION` (as I requested on @whot's kernel patch). Then we can have logind that loads a BPF program that monitors every `/dev/hidraw` opened files. If a file is opened in the session, we know it has access to it (with `uaccess`). We can also raise an event to logind with the pid that opened the file, and logind can decide or not to add the file descriptor to the list of potentially revoked ones.

Once we do a VT switching, logind then mark all opened files as revoked in its BPF program, and this has exactly the same effect than calling the ioctl.

The nice advantage here is that we don't need any code change in clients, it is all transparent for them. The only difference with the existing state is that their devices will be "unplugged" when there is a fast user switching, which is something they should handle properly already.

IMO we can have both (in term of kernel): the new ioctl and the BPF capability: the ioctl can be used for the flatpak portals, because we have to ask the portal for the fd, and thus we are in charge of that fd, while for steam/SDL games, we can use the emulation trough BPF.

hadess:
> If a file is opened in the session, we know it has access to it (with `uaccess`).

The problem here is that we also want to make sure that not all hidraw devices are available to end-users, and we also need a way to open those, so in quite a few cases (not the joystick/game controller case) we'll need a privileged D-Bus client to do that job for us.

bentiss:
> > If a file is opened in the session, we know it has access to it (with `uaccess`).
> 
> The problem here is that we also want to make sure that not all hidraw devices are available to end-users, and we also need a way to open those, so in quite a few cases (not the joystick/game controller case) we'll need a privileged D-Bus client to do that job for us.

BPF has access to all of the informations you need: who opened the file (its pid) and which file has been opened (the hidraw number), so nothing prevents to have a generic udev rule that tags with uaccess the gamepads, but also prematurely revoke access to hidraw to pids who are not allowed to do so: when a hidraw node is opened, logind gets the notification, decides whether or not the access is valid (by showing up a popup to the user), and then do nothing or revoke the file access if it is not supposed to happen.

The application will be able to open the file, but once it tries to read events (as soon as logind revokes the access), the fd will be unusable.

hadess:
That sounds racy, and that it might cause the problems mentioned in https://gitlab.freedesktop.org/bentiss/logind-hidraw/-/issues/1:
```
Plus, the ioctl chain does not actually open the device until everyone agrees, so the portal can present some UI. With BPF, we would need to pause the already-open fd until the UI is complete, then unmute the fd. That may work for input reports but it's unlikely to work for output/feature reports.
```

We'll also need a way to figure out, at runtime, whether hidraw BPF (and hidraw revoke) are available, as you really don't want to be tagging uaccess for all the hidraw devices if there's no hidraw BPF support to revoke access as soon as opened.

bentiss:
> That sounds racy, and that it might cause the problems mentioned in https://gitlab.freedesktop.org/bentiss/logind-hidraw/-/issues/1:
> 
> ```
> Plus, the ioctl chain does not actually open the device until everyone agrees, so the portal can present some UI. With BPF, we would need to pause the already-open fd until the UI is complete, then unmute the fd. That may work for input reports but it's unlikely to work for output/feature reports.
> ```

This is racy, yes,  but do we really care that the device is readable for a few (milli)seconds? Muting the device is rather problematic, and if you think about it, it's not worse than what we have now. With that approach, we can revoke files without touching a line in the client code.

(I'll reply in more detail to this issue in the gitlab repo).

> 
> We'll also need a way to figure out, at runtime, whether hidraw BPF (and hidraw revoke) are available, as you really don't want to be tagging uaccess for all the hidraw devices if there's no hidraw BPF support to revoke access as soon as opened.

That's easy enough: if you managed to load the BPF program, then you have both :)
That's the advantage of BPF with this patch compared to ioctl: with the ioctl you need to check the return value of the command after the fact, while with bpf, if `hidraw_is_revoked()` is not exported by the kernel, you know it won't be available and the BPF program will fail at loading.

hadess:
> This is racy, yes, but do we really care that the device is readable for a few (milli)seconds?

Yes, absolutely. Not all "HID" attached devices are harmless input devices.

bentiss:
> > This is racy, yes, but do we really care that the device is readable for a few (milli)seconds?
> 
> Yes, absolutely. Not all "HID" attached devices are harmless input devices.

I am not suggesting to open up to the world to all hidraw nodes. #22860 is IMO what we need: some tagging to "safe" devices. All of the other hidraw access either need root (and your machine is already compromised), either will need to go through this dbus-ioctl approach, so we can add some more rules and prevent access until we are judging it's safe to do so.

hadess:
> I am not suggesting to open up to the world to all hidraw nodes. #22860 is IMO what we need: some tagging to "safe" devices. All of the other hidraw access either need root (and your machine is already compromised), either will need to go through this dbus-ioctl approach, so we can add some more rules and prevent access until we are judging it's safe to do so.

OK, that works for me.

src/login/logind-session-device.c
```diff
@@ -285,7 +327,7 @@ static int session_device_verify(SessionDevice *sd) {
                 if (sd_device_get_syspath(dev, &sp) < 0)
                         return -ENODEV;
 
-        } else if (sd->type != DEVICE_TYPE_DRM)
+        } else if (sd->type != DEVICE_TYPE_DRM && sd->type != DEVICE_TYPE_HIDRAW)
```
keszybz, line 330: `!IN_SET(d->type, DEVICE_TYPE_DRM, DEVICE_TYPE_HIDRAW))`

keszybz (commented):
> bit unsure about the udev rule change for the seat tag and whether that pulls in more than we need

What the PR does seems OK.

So… what is the status here? Did the kernel patches get queued up?

poettering:
could looks excellent to me, btw. if this can fly on the kernel side, I am loving this.

i wonder wht this means for the fido2 stuff we already have, i.e. right now web browsers open the hidraw stuff directly, ideally they wouldn't do that anymore...

poettering:
> Doesn't cover Bluetooth HID devices though? I'd say revoke for USB is largely independent. For hidraw what we will likely need is a `LD_PRELOAD`-capable wrapper that routes `open("/dev/hidraw")` through whichever portal we want to use.

sandboxing environments could probably use the new seccomp fd passing stuff to hook into the open(), and then install an fd acquired via logind into the processes.

i.e. using SECCOMP_IOCTL_NOTIF_ADDFD, see https://man7.org/linux/man-pages/man2/seccomp_unotify.2.html

bentiss:
>> revoke ioctl at the VFS level, not per subsystem.
> wasn't that attempted before and failed?

Yes, that was one of the remark in the thread by Greg KH.
OTOH, we *could* have a small generic stub ioctl REVOKE that subsystem could then implement as they want.

> Also: having al viro in the discussion sounds like a great way to derail it...

Yep, good point. But as mentioned, the BPF approach is shinier, as logind will not be the man in the middle.

hadess:
> wasn't that attempted before and failed? Also: having al viro in the discussion sounds like a great way to derail it...

Yes, and the semantics of the revoke()/frevoke() calls in other Unixes don't actually match what we want to do here. So it's BPF or bust.

poettering:
any update on this? what#s the state of the bpf thing?

whot:
on the backburner atm, I don't have time for it right now until (most likely) the end of the year.

poettering:
any update on this? what happened to the kernel work here?

whot:
Thanks for the reminder, I forgot about this one. I'm closing this, can be zombied at some later point if need be.

@bentiss' hid-bpf work has since been merged into the upstream and there's a chance that we'll be using that one, once someone actually finds the time to work on it...

PR #23147, opened 2022-04-21
bus-unit-util: make sure we can set LoadCredentials= property with a …

poettering:
…single string

LoadCredentials= in unit files supports a syntax passing a single string
only (in which case the credentials are propagated down from the host).
but systemd-run's --property= setting doesn't allow that yet. Fix that.


yuwata (approved):
LGTM.

PR #23155, opened 2022-04-21
tpm2-util: if we run in a container, ignore /sys/class/tpmrm/* contents

poettering:


src/shared/tpm2-util.c
```diff
@@ -1460,12 +1461,18 @@ Tpm2Support tpm2_support(void) {
         Tpm2Support support = TPM2_SUPPORT_NONE;
         int r;
 
-        r = dir_is_empty("/sys/class/tpmrm");
-        if (r < 0) {
-                if (r != -ENOENT)
-                        log_debug_errno(r, "Unable to test whether /sys/class/tpmrm/ exists and is populated, assuming it is not: %m");
-        } else if (r == 0) /* populated! */
-                support |= TPM2_SUPPORT_DRIVER;
+        if (detect_container() <= 0) {
```
yuwata, line 1464: Maybe, it is better to use `udev_available()`?

poettering: Well, it's not tied to udev here. This doesn't use udev APIs (and not sure it should...)

PR #23157, opened 2022-04-21
execute: refactor credential passing code, and fix two bugs

poettering:
This is basically a follow-up for 3989bdc1ad7cca4d75c06cdf601fea2cb37ba337, and fixes two issues that introduced:

1. SetCredential=/LoadCredential= no longer worked as fallback for each other (i.e. the idea is that if you have SetCredential=/LoadCredential= set for the same credential, the former acts as fallback value if the latter's source file does not exist). This documented behaviour is restored.

2. LoadCredential= with a relative source path name was supposed to "inherit" the credential from further up, i.e. nspawn#s --set-credential= switch for example. This documented behaviour is also restored.

Various other minor refactorings.

### DIFF
--- a/test/TEST-54-CREDS/test.sh
+++ b/test/TEST-54-CREDS/test.sh
@@ -3,6 +3,7 @@
 set -e
 
 TEST_DESCRIPTION="test credentials"
+NSPAWN_ARGUMENTS="--set-credential=mynspawncredential:strangevalue"
 
 # shellcheck source=test/test-functions
 . "${TEST_BASE_DIR:?}/test-functions"
--- a/test/units/testsuite-54.sh
+++ b/test/units/testsuite-54.sh
@@ -22,6 +22,20 @@ echo piff > /tmp/ts54-fallback
 rm /tmp/ts54-fallback
 [ "$(systemd-run -p LoadCredential=paff:/tmp/ts54-fallback -p SetCredential=paff:poff --pipe --wait systemd-creds cat paff)" = "poff" ]
 
+if systemd-detect-virt -q -c ; then
+    # If this test is run in nspawn a credential should have been passed to us. See test/TEST-54-CREDS/test.sh
+    [ "$(systemd-creds --system cat mynspawncredential)" = "strangevalue" ]
+
+    # Test that propagation from system credential to service credential works
+    [ "$(systemd-run -p LoadCredential=mynspawncredential --pipe --wait systemd-creds cat mynspawncredential)" = "strangevalue" ]
+
+    # Check it also works, if we rename it while propagating it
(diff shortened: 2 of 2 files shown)


src/core/execute.c
```diff
@@ -2620,14 +2616,19 @@ static int load_credential(
...
+        if (path_is_absolute(path) || read_dfd >= 0) {
                 /* If this is an absolute path, read the data directly from it, and support AF_UNIX sockets */
-                source = lc->path;
+                source = path;
                 flags |= READ_FULL_FILE_CONNECT_SOCKET;
 
                 /* Pass some minimal info about the unit and the credential name we are looking to acquire
                  * via the source socket address in case we read off an AF_UNIX socket. */
-                if (asprintf(&bindname, "@%" PRIx64"/unit/%s/%s", random_u64(), unit, lc->id) < 0)
+                if (asprintf(&bindname, "@%" PRIx64"/unit/%s/%s", random_u64(), unit, id) < 0)
```
bluca, line 2635: `strempty(id)` or `assert(id)`? I think printf would barf when given null

bluca:
Can a test for this also be added, to avoid further regressions?

> LoadCredential= with a relative source path name was supposed to "inherit" the credential from further up, i.e. nspawn#s --set-credential= switch for example. This documented behaviour is also restored.

src/core/execute.c
```diff
@@ -2833,13 +2833,17 @@ static int acquire_credentials(
...
-         * and thus act as a "default" if the same credential is specified multiple times */
+        /* Second, we add in literally specified credentials. If the credentials already exist, we'll not add
+         * them, so that they can act as a "default" if the same credential is specified multiple times. */
         HASHMAP_FOREACH(sc, context->set_credentials) {
                 _cleanup_(erase_and_freep) void *plaintext = NULL;
                 const char *data;
                 size_t size, add;
 
+                /* Note that we check ahead of time here instead of relying on O_EXCL|O_CREAT later to return
+                 * EEXIST if the credential already exists. That's because the TPM2-based decrytion is kinda
```
dnicolodi: typo: `s/decrytion/decryption/`

poettering:
> Can a test for this also be added, to avoid further regressions?

done

bluca:
> > Can a test for this also be added, to avoid further regressions?
> 
> done

Did the push fail? I don't see the new version

poettering:
> Did the push fail? I don't see the new version

No, my "ninja -C build test && git push -f" just took a bit longer than I hoped.

poettering:
(also fixed the other issues raised. ptal)

bluca (approved):

poettering:
CI failure appears to be LXC network woes

PR #23160, opened 2022-04-21
meson: simplify setting of default compression

keszybz:
Follow-up for da13d2ca0731b413841663052f2cc6832a855334. Instead of having
separate definitions of the bitmask flags, just define DEFAULT_COMPRESSION_FOO=0|1
directly.

(It *should* be possible to do this more simply, but the problem is that
anything that is used in #if cannot refer to C constants or enums. This is the
simplest I could come up with that preserves the property that we don't use #ifdef.)

The return value from compress_blob() is changed to propagate the error instead
of always return -EOPNOTSUPP. The callers don't care either way.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1470,9 +1470,9 @@ elif compression == 'lz4' and not have_lz4
 elif compression == 'xz' and not have_xz
         error('default-compression=xz requires xz')
 endif
-conf.set10('DEFAULT_COMPRESSION_ZSTD', compression == 'zstd')
-conf.set10('DEFAULT_COMPRESSION_LZ4', compression == 'lz4')
-conf.set10('DEFAULT_COMPRESSION_XZ', compression == 'xz')
+conf.set('DEFAULT_COMPRESSION',
+         compression == 'none' ? 0 :
+                 'OBJECT_COMPRESSED_@0@'.format(compression.to_upper()))
 
 want_xkbcommon = get_option('xkbcommon')
 if want_xkbcommon != 'false' and not skip_deps
--- a/src/libsystemd/sd-journal/compress.h
+++ b/src/libsystemd/sd-journal/compress.h
@@ -53,37 +53,42 @@ int decompress_stream_zstd(int fdf, int fdt, uint64_t max_size);
 
 static inline int compress_blob(const void *src, uint64_t src_size,
                                 void *dst, size_t dst_alloc_size, size_t *dst_size) {
-#if DEFAULT_COMPRESSION_ZSTD
-        return compress_blob_zstd(src, src_size, dst, dst_alloc_size, dst_size);
-#elif DEFAULT_COMPRESSION_LZ4
-        return compress_blob_lz4(src, src_size, dst, dst_alloc_size, dst_size);
-#elif DEFAULT_COMPRESSION_XZ
-        return compress_blob_xz(src, src_size, dst, dst_alloc_size, dst_size);
-#else
-        return -EOPNOTSUPP;
-#endif
+        switch (DEFAULT_COMPRESSION) {
+        case OBJECT_COMPRESSED_ZSTD:
+                return compress_blob_zstd(src, src_size, dst, dst_alloc_size, dst_size);
+        case OBJECT_COMPRESSED_LZ4:
+                return compress_blob_lz4(src, src_size, dst, dst_alloc_size, dst_size);
+        case OBJECT_COMPRESSED_XZ:
+                return compress_blob_xz(src, src_size, dst, dst_alloc_size, dst_size);
+        default:
+                return -EOPNOTSUPP;
+        }
 }
 
 static inline int compress_stream(int fdf, int fdt, uint64_t max_bytes, uint64_t *ret_uncompressed_size) {
-#if DEFAULT_COMPRESSION_ZSTD
-        return compress_stream_zstd(fdf, fdt, max_bytes, ret_uncompressed_size);
-#elif DEFAULT_COMPRESSION_LZ4
-        return compress_stream_lz4(fdf, fdt, max_bytes, ret_uncompressed_size);
-#elif DEFAULT_COMPRESSION_XZ
-        return compress_stream_xz(fdf, fdt, max_bytes, ret_uncompressed_size);
-#else
-        return -EOPNOTSUPP;
-#endif
+        switch (DEFAULT_COMPRESSION) {
+        case OBJECT_COMPRESSED_ZSTD:
+                return compress_stream_zstd(fdf, fdt, max_bytes, ret_uncompressed_size);
+        case OBJECT_COMPRESSED_LZ4:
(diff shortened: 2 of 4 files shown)


yuwata (approved):
Yeah, that's much simpler. LGTM.

yuwata:
test-journal and test-repart failed. I pushed a follow-up commit on this.
@keszybz If you like the fix, then please squash the change on merge. If you do not like it, feel free to drop it.

src/libsystemd/sd-journal/compress.h
```diff
@@ -72,19 +51,38 @@ int decompress_stream_xz(int fdf, int fdt, uint64_t max_size);
...
+        return compress_blob_zstd(src, src_size, dst, dst_alloc_size, dst_size);
+#elif DEFAULT_COMPRESSION_LZ4
+        return compress_blob_lz4(src, src_size, dst, dst_alloc_size, dst_size);
+#elif DEFAULT_COMPRESSION_XZ
+        return compress_blob_xz(src, src_size, dst, dst_alloc_size, dst_size);
+#else
+        return -EOPNOTSUPP;
+#endif
+}
+
```
dnicolodi, line 67: Another possibility would be to rewrite this as a C construct instead than as a preprocessor macro. The compiler should be able to optimize it away:
```c
static inline int compress_blob(const void *src, uint64_t src_size,
                                void *dst, size_t dst_alloc_size, size_t *dst_size) {
        switch (DEFAULT_COMPRESSION) {
        case COMPRESSION_ZSTD:
                return compress_blob_zstd(src, src_size, dst, dst_alloc_size, dst_size);
        case ...
        default:
                return -EOPNOTSUPP;
        }
}
```

keszybz: This has the problem that DEFAULT_COMPRESSION must be defined to one of the constants. So the constants must be defined, and we need to duplicate their definitions in `meson.build`. This patch removes this duplication, which means that this approach is not possible.

keszybz:
> test-journal and test-repart failed. I pushed a follow-up commit on this. @keszybz If you like the fix, then please squash the change on merge. If you do not like it, feel free to drop it.

Oh, I missed that call site somehow.

I'll squash you patch, but rework this a bit. I think it's nicer to return the right flag from `compress_blob()` as it was before, rather then let the caller assume what the compression was.

src/libsystemd/sd-journal/compress.h
dnicolodi: Maybe I should have been more explicit: turning this into a C construct you don't need the defined in Meson, you can just set DEFAULT_COMPRESSION to a C enum label
```meson
conf.set('DEFAULT_COMPRESSION', 'COMPRESSION_ZSTD')
```
as the check is done by the compiler and not by the preprocessor.

keszybz: Nope, this doesn't work.

keszybz:
PTAL again.

src/libsystemd/sd-journal/compress.h
dnicolodi: I just tested to make sure, and it works. Minimal example:
```meson
project('test','c')

conf = configuration_data()
conf.set('DEFAULT_COMPRESSION', 'COMPRESSION_ZSTD')

configure_file(
  output: 'config.h',
  configuration: conf)

executable('main', 'main.c')
```
```c
#include <stdio.h>

#include "config.h"

enum setting {
  COMPRESSION_ZSTD,
  COMPRESSION_XZ,
};

int main(int argc, char **argv)
{
  switch (DEFAULT_COMPRESSION) {
  case COMPRESSION_ZSTD:
    printf("zstd\n");
    return 0;
  case COMPRESSION_XZ:
    printf("xz\n");
    return 0;
  default:
    return 1;
  }
}
```
Whether you like it or not, or whether it is preferable to go this way or another way is a different matter, but it works just fine.

bluca: The problem is that you need to define COMPRESSION_XYZ too in meson, so that config.h is self contained, and I guess this PR wants to avoid that

bluca (approved):

bluca:
CI is still unhappy:

```
 559/1337 test-compress-benchmark                                                                                 TIMEOUT         90.01s   killed by signal 15 SIGTERM
```

src/libsystemd/sd-journal/compress.h
keszybz: systemd uses `-Werror=undef`. This example ~~will~~ would fail with an error about the line `#define DEFAULT_COMPRESSION COMPRESSION_ZSTD`, something like "COMPRESSION_ZSTD is undefined".

dnicolodi: Are you sure? I just extended the example above to set `-Werror=undef` and with a second source files that includes `config.h` but not the definition of the enum and I don't get any error. However, this is with clang, maybe gcc is more strict.

keszybz: Oh, OK. If the macro is never used in `#if`, but only in C code, we don't get the error. This means that all places that use `#if` have to be refactored into C. But indeed, I think this looks a bit nicer I'll push that as a separate commit.

meson.build
```diff
@@ -1470,9 +1470,7 @@ elif compression == 'lz4' and not have_lz4
 elif compression == 'xz' and not have_xz
         error('default-compression=xz requires xz')
 endif
-conf.set10('DEFAULT_COMPRESSION_ZSTD', compression == 'zstd')
-conf.set10('DEFAULT_COMPRESSION_LZ4', compression == 'lz4')
-conf.set10('DEFAULT_COMPRESSION_XZ', compression == 'xz')
+conf.set('DEFAULT_COMPRESSION', 'OBJECT_COMPRESSED_@0@'.format(compression.to_upper()))
```
bluca: compat.h is no longer self-defined now (it was raised as a point in the previous PR)

dnicolodi: I think you mean `config.h` and `config.h` already contains defines for `ANSI_OK_COLOR` and `STATUS_UNIT_FORMAT_DEFAULT` to enum identifiers or other preprocessor macros not defined in `config.h`.

bluca:
```
../src/libsystemd/sd-journal/journal-file.c: In function ‘journal_file_open’:
../src/libsystemd/sd-journal/journal-file.c:3380:12: error: ‘r’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 3380 |         if (r < 0) {
```

meson.build
keszybz: It's "self-defined" in the sense that you can include it and it is valid. Some of the defines are not meaningful without other headers, but that's OK.

bluca: won't the preprocessor barf because OBJECT_COMPRESSED_XYZ does not exist?

keszybz: It should not. See the discussion above.

yuwata (approved):
LGTM.

PR #23162, opened 2022-04-22
udev: check stats of .link files and their drop-in files

### LINKED ISSUE #23128: systemd-udev does not use newer values from networkd drop-in link configuration
 > 249, 250, master

**Used distribution**
 > CentOS 8

 > `Linux 5.6.13`

 > x86_64

 > Expected that after changing a configuration value for a `.link` drop-in and triggering with `udevadm trigger --action add <sys/path/to/interface>`, the newer configuration value should take effect

 > The drop-in `.link` configuration value wasn't effective unless there was a change in the main `.link` configuration

Assume a system has the following configuration
*/etc/systemd/network/foo.link*
``` 
[Link]
RxBufferSize = 2000

[Match]
OriginalName = eth0
```
*/etc/systemd/network/foo.link.d/1.conf*
``` 
[Link]
RxBufferSize = 3000

[Match]
OriginalName = eth0
```
and `ethtool -g eth0` correctly shows the Rx buffer to be `3000`(overridden).

Now change the value in drop-in for `RxBufferSize` to, say `5000` and trigger with `udevadm trigger --class add /sys/class/net/eth0`. You'd expect udevd to load the new configuration value and `ethtool -g eth0` to show `5000`, however it does not. Now if you change some random value in the main, `foo.link`, link configuration and `udevadm trigger` then the expect Rx buffer value shows up.

This problem is obvious when browsing the source code as it [reloads configuration](https://github.com/systemd/systemd/blob/da13d2ca0731b413841663052f2cc6832a855334/src/udev/net/link-config.c#L335) by using the timestamp of the base directory alone, `/etc/systemd/network`, and not the drop-ins.

This look like an undesired behavior and I can work on a patch if there is consensus that this should be fixed.

yuwata:
Fixes #23128.

### DIFF
--- a/src/shared/conf-parser.c
+++ b/src/shared/conf-parser.c
@@ -34,6 +34,7 @@
 #include "set.h"
 #include "signal-util.h"
 #include "socket-util.h"
+#include "stat-util.h"
 #include "string-util.h"
 #include "strv.h"
 #include "syslog-util.h"
@@ -556,6 +557,27 @@ int config_parse_many_nulstr(
                                        ret_stats_by_path);
 }
 
+static int config_get_dropin_files(
+                const char* const* conf_file_dirs,
+                const char *dropin_dirname,
+                char ***ret) {
+
+        _cleanup_strv_free_ char **dropin_dirs = NULL;
+        const char *suffix;
+        int r;
+
+        assert(conf_file_dirs);
+        assert(dropin_dirname);
+        assert(ret);
+
+        suffix = strjoina("/", dropin_dirname);
+        r = strv_extend_strv_concat(&dropin_dirs, (char**) conf_file_dirs, suffix);
+        if (r < 0)
+                return r;
+
+        return conf_files_list_strv(ret, ".conf", NULL, 0, (const char* const*) dropin_dirs);
+}
+
 /* Parse each config file in the directories specified as strv. */
 int config_parse_many(
                 const char* const* conf_files,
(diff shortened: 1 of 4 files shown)


bluca:
CI looks sad:

```
07:56:33 FAILED TASKS:
07:56:33 -------------
07:56:33 systemd-networkd_sanitizers
```

gogsbread (approved):

src/shared/conf-parser.c
```diff
@@ -568,21 +590,127 @@ int config_parse_many(
...
+        /* First read the main config file. */
+        r = RET_NERRNO(stat(conf_file, &st));
+        if (r >= 0) {
+                r = hashmap_put_stats_by_path(&stats_by_path, conf_file, &st);
+                if (r < 0)
+                        return r;
+        } else if (r != -ENOENT)
+                return r;
+
+
```
daandemeyer: Spurious newline

src/shared/conf-parser.c
```diff
@@ -568,21 +590,127 @@ int config_parse_many(
...
+        /* Then read all the drop-ins. */
+        STRV_FOREACH(fn, files) {
+                if (stat(*fn, &st) < 0) {
+                        if (errno == ENOENT)
+                                continue;
```
daandemeyer, line 650: There's probably a good reason but I don't entirely understand why we need to ignore ENOENT here. Can you eplxain that (and maybe add a comment)?

DaanDeMeyer (approved):
LGTM, some minor comments

src/shared/conf-parser.c
```diff
@@ -568,21 +590,127 @@ int config_parse_many(
...
+
+                        return -errno;
+                }
+
+                r = hashmap_put_stats_by_path(&stats_by_path, *fn, &st);
+                if (r < 0)
+                        return r;
+        }
+
```
poettering, line 659: this should also detect mtime of the drop-in dir itself, so that we can detect added files.

src/shared/conf-parser.c
poettering: TOCTTOU. i.e. you enumerate the file with readddir()/getdents(), but when you try to stat it, it's already gone. All across our codebase we should thus handle cases gracefully where after seeing a dir in readdir() the stat() on it fails with ENOENT

src/shared/conf-parser.c
```diff
@@ -568,21 +590,127 @@ int config_parse_many(
...
+                        if (errno == ENOENT)
+                                continue;
+
+                        return -errno;
+                }
+
+                r = hashmap_put_stats_by_path(&stats_by_path, *fn, &st);
+                if (r < 0)
+                        return r;
+        }
```
poettering, line 658: btw, i wonder if we should maybe extend the readdir_all() stuff we have now, to do a statx() of an entire directory listing as well depending on a flag, or so. could be handy for things like that, because we could go by fd and stuff...

(but dunno, stuff for another time)

src/shared/conf-parser.c
gogsbread: `stats_by_path_equal` does a hashmap size comparison which should detect newly added and deleted files including the drop-ins?

src/shared/conf-parser.c
yuwata: Right, the file may be removed after readdir().

src/shared/conf-parser.c
yuwata: Here, I think @gogsbread is right. storing stats for drop-in dir is not necessary, and moreover, it triggers unnecessary reload: e.g. if we have 
- /usr/lib/systemd/network/10-eth0.link.d/a.conf
- /usr/lib/systemd/network/10-eth0.link.d/b.conf
- /etc/systemd/network/10-eth0.link.d/a.conf

Then, it is not necessary to reload configs when /usr/lib/systemd/network/10-eth0.link.d/a.conf is updated, but we need to reload configs when /usr/lib/systemd/network/10-eth0.link.d/b.conf is updated.

src/shared/conf-parser.c
yuwata: Sorry, I cannot follow the suggestion. But it sounds not relevant to this PR. Anyway, I'd like to leave extending `readdir()` to another PR.

yuwata:
@DaanDeMeyer, @poettering and @gogsbread Thank you for reviewing this PR. Only addressed the doubled blank line. For other comments or suggestions, please see the above replies.

src/shared/conf-parser.c
daandemeyer: I can't think of a scenario where the current approach wouldn't detect added files either.

DaanDeMeyer (approved):

src/shared/conf-parser.c
daandemeyer: One question: Do we store full paths in the stats_by_path hashmap or just the relative path to /usr or /etc or some other config directory? Asking because if we store the full path, then updating /usr/lib/systemd/network/a.conf if /etc/systemd/network/a.conf exists will lead to a reload no, even if it isn't needed in that case because the file in /etc overrides the file in /usr?

yuwata: If /usr/lib/systemd/network/10-eth0.link.d/a.conf exists, then /etc/systemd/network/10-eth0.link.d/a.conf is not listed by `config_get_dropin_files()`. Then, storing full path does not trigger unnecessary reload.

Moreover, when /usr/lib/systemd/network/10-eth0.link.d/a.conf is removed, and /etc/systemd/network/10-eth0.link.d/a.conf has take an effect, then we should reload configs. Of course, in that case, `struct stat` is also different, but it is better to detect the change also from the path. Hence, I think it is better to store full path.

Issue #15266, opened 2020-03-30
libmount utab mount options get lost when we switch root, and thus x-initrd.mount vanishes during initrd transitions


fbuihuu:
x-initrd.mount (for instance) is used for setting mounts up in initrd. This "application" option is recorded in "utab" so applications (such as systemd) can figure out later which application options have been used for a given mount.

However	in initrd, mount points described in fstab are usually mounted under /sysroot, therefore mount paths in utab are usually prefixed with "/sysroot". A typical entry in utab looks like:

`SRC=/dev/vdb1 TARGET=/sysroot/mnt ROOT=/ OPTS=x-initrd.mount,_netdev`

The problem appears after swithing to the main system: PID1 tries to retrieve the mount options (via mnt_fs_get_options()) for /mnt (not /sysroot/mnt) but nothing is found since the relevant entry has `TARGET=/sysroot/mnt`.

The previous example is based on x-initrd.mount but the issue can also happen with any other x-* options.

poettering:
Hmm, I wonder what the right approach here... somehow libmount needs to either convert utab or better automatically handle this better.

Maybe the correct approach would be if libmount only ever uses the numeric mount ID (rather than the mount point path) for identifying mounts? that way things should remain robust regardless how many times we switch root.

@karelzak any chance you could rework libmount like that? i.e. replace the utab key with the numeric mount ID?

karelzak:
I'll try it. @poettering suggested to use  /proc/self/fdinfo/<mountpoint-fd> to get the ID -- it seems elegant way how to bypass the current mount(2) stupidity (it does not provide any feedback after success). In the future with the new mount API we can get the ID from fsmount() FD.

karelzak:
@poettering has suggested mount ID to join utab and mountinfo. The question is how usable mount ID is as it seems kernel is cycling the IDs ;-(

Is there any possibility to detect that the current root is in /sysroot directory? If yes, we can improve the way how libmount search in the utab.

fbuihuu:
@poettering in the meantime maybe we could workaround this issue with something like https://github.com/fbuihuu/systemd/commit/303c959147e07b2b3865228dda73e62033983450 ?

fbuihuu:
> @poettering in the meantime maybe we could workaround this issue with something like [fbuihuu@303c959](https://github.com/fbuihuu/systemd/commit/303c959147e07b2b3865228dda73e62033983450) ?

[...]

PR #23168, opened 2022-04-22
Inform libmount when switching root

fbuihuu:


poettering:
I am not sure I understand why doing this in systemd is really a good idea. Why doesn't mount maintain its utab by mount ids rather than paths? If so the problem would pretty much just go away.

@karelzak can you comment?

poettering:
Do I understand that this approach here requires both patches in libmount *and* here?

An approach to fix libmount to use mount ids as keys sounds a million time better, because it will fix this without work in systemd, and in fact will fix any other code doing switch root too.

fbuihuu:
Using mount IDs seems a better approach indeed but it seems a rather substantial change for libmount, which I guess won't happen any time soon (issue #15266 is 2 years old).

OTOH this change + https://github.com/util-linux/util-linux/pull/1660 are simpler so could be quickly accepted and they could quickly fix issue #15266. And once libmount will gain support for mount IDs and the systemd requirement on libmount bumped we could simply drop this.

poettering:
> Well I'm afraid that adding mount IDs looks requiring (much) more changes (that will probably won't be easy to backport too). Maybe I'm wrong so let's wait for @karelzak's opinion.

systemd is not a dumping ground for "quick hacks". We very much prefer proper fixes. Thank you. We are not in this because it's easy, but because we want to get things mostly right.

poettering:
> Well, classic kernel API does not return any ID after mount(2) call, and the library has no ID in the time it writes to utab.

Well, you can get it easily with statx() these days, even in a way that doesn't trigger the fs. of course it would not be atomic (i.e. in theory something else could jump in between your mount() and your statx()), but given that we are talking about a status quo ante where things where using dirs I think the benefit still outweighs things by far.

You could add it as "fake" mount option in utab, i.e. as auxiliary info at the end somewhere, if statx() gives it to you. And then only use it to match up entries if the data is in utab and the client can query the mntid via statx() safely too.

karelzak:
@poettering do you mean statfs->f_fsid (or the same for statvfs()) ? 

The ideal solution would be to get mount ID from descriptor created by fsmount()  or open_tree(), then we can do it in an atomic way. (I'd like to rewrite the library to use the new API if possible.)

karelzak:
Cool, I have tested it and it works. You can get mount ID before the mount node is attached to the tree, so it's atomic.

For example code below mount device without classic mount(8) call.
```
static int api_create_mount(void)
{        
        int fd, mfd;
        struct statx st;
         
        fd = fsopen("ext4", FSOPEN_CLOEXEC);
        if (fd < 0)
                err(EXIT_FAILURE, "fsopen failed");
         
        if (fsconfig(fd, FSCONFIG_SET_STRING, "source", "/dev/sdc1", 0))
                err(EXIT_FAILURE, "fsconfig set path failed");
        if (fsconfig(fd, FSCONFIG_SET_STRING, "errors", "continue", 0))
                err(EXIT_FAILURE, "fsconfig set string failed");
        if (fsconfig(fd, FSCONFIG_SET_FLAG, "ro", NULL, 0))
                err(EXIT_FAILURE, "fsconfig set flag failed");
         
        if (fsconfig(fd, FSCONFIG_CMD_CREATE, NULL, NULL, 0))
                err(EXIT_FAILURE, "fsconfig create failed");
      
        mfd = fsmount(fd, FSMOUNT_CLOEXEC, 0);
        if (mfd < 0)
                err(EXIT_FAILURE, "fsmount failed");
         
        if (statx(mfd, "", AT_EMPTY_PATH, STATX_MNT_ID, &st) < 0)
                err(EXIT_FAILURE, "statfs failed");
         
        fprintf(stderr, "KZAK>>> ID=%" PRIu64 "\n", (uint64_t) st.stx_mnt_id);
         
        close(fd);
        return mfd;
}        
         
static int api_attache_mount(int fd, const char *target)
{        
        if (move_mount(fd, "", AT_FDCWD, target, MOVE_MOUNT_F_EMPTY_PATH))
                err(EXIT_FAILURE, "attache FS failed");
        return 0;
}        
         
int main(int argc, char **argv)
{        
        int fd;
         
        fd = api_create_mount();
        api_attache_mount(fd, "/mnt/test");
         
        close(fd);
        return 0;
}        
 ```

poettering:
@karelzak So, what's the perspective here? Should we close this here, under the assumption libmount will eventuall handle this nicely for us?

karelzak:
> @karelzak So, what's the perspective here? Should we close this here, under the assumption libmount will eventuall handle this nicely for us?

I have already merged the mount points paths update on the move, the ID-based utab management is not implemented yet, but it's my plan for the next release. So yes, it's something we can handle in libmount.

poettering:
OK, then let's close this here, and continue discussions over on https://github.com/util-linux/util-linux/issues/1765

PR #23170, opened 2022-04-22
import system credentials from sd-stub + qemu fw_cfg + kernel cmdline explicitly in PID 1

poettering:
Previously, you could path in system creds from a container manager to systemd (e.g. nspawn's --set-credential= and --load-credential= switch). This adds support to doing the same from qemu (via "fw_cfg") and the kernel cmdline. Also, creds passed in via sd-stub are similarly picked up.

Also, search for creds to import in /etc/credstore/, /run/credstore and /usr/lib/credstore.

Also add some longer docs about creds

### DIFF
--- a/TODO
+++ b/TODO
@@ -165,11 +165,6 @@ Features:
   don't query this unnecessarily in entirely uninitialized
   containers. (i.e. containers with empty /etc).
 
-* systemd creds hookup with qemu fw_cfg. (Quite possibly might not need any
-  code at all, given the fw_cfg stuff are just files, but we should then
-  document how to use it). Goal: provide symmetric ways to pass creds to nspawn
-  containers and qemu VMs. (maybe also pick up env vars from fw_cfg?)
-
 * beef up sd_notify() to support AV_VSOCK in $NOTIFY_SOCKET, so that VM
   managers can get ready notifications from VMs, just like container managers
   from their payload. Also pick up address from qemu/fw_cfg if set there.
@@ -535,14 +530,7 @@ Features:
 
 * expose MS_NOSYMFOLLOW in various places
 
-* make LoadCredential= automatically find credentials in /etc/creds,
-  /run/creds, … and so on, if path component is unqualified
-
-* teach LoadCredential=/LoadCredentialEncrypted= to load credentials from
-  kernel cmdline, maybe: LoadCredentialEncrypted=foobar:proc-cmdline:foobar
-
 * credentials system:
-  - acquire from kernel command line
   - acquire from EFI variable?
   - acquire via via ask-password?
   - acquire creds via keyring?


poettering:
I'll add an integeration test for this soon.

poettering:
> I'll add an integeration test for this soon.

done. ptal!

mrc0mmand:
TEST-54 seems to be unhappy under qemu:

```
[   10.169906] H testsuite-54.sh[241]: + '[' poff = poff ']'
[   10.170001] H testsuite-54.sh[241]: + systemd-detect-virt -q -c
[   10.183454] H testsuite-54.sh[264]: ++ systemd-creds --system cat kernelcmdlinecred
[   10.192094] H testsuite-54.sh[241]: + '[' uff = uff ']'
[   10.192094] H testsuite-54.sh[241]: + systemd-detect-virt -q -v
[   10.202390] H testsuite-54.sh[241]: + expected_credential=myqemucredential
[   10.202390] H testsuite-54.sh[241]: + expected_value=othervalue
[   10.202650] H testsuite-54.sh[241]: + '[' myqemucredential '!=' '' ']'
[   10.204200] H testsuite-54.sh[266]: ++ systemd-creds --system cat myqemucredential
[   10.211567] H systemd-creds[266]: Credential 'myqemucredential' not set.
[   10.212330] H testsuite-54.sh[241]: + '[' '' = othervalue ']'
[   10.213316] H systemd[1]: Received SIGCHLD from PID 241 (bash).
[   10.213457] H systemd[1]: Child 241 (bash) died (code=exited, status=1/FAILURE)
[   10.213592] H systemd[1]: testsuite-54.service: Child 241 belongs to testsuite-54.service.
[   10.213692] H systemd[1]: testsuite-54.service: Main process exited, code=exited, status=1/FAILURE
[   10.213957] H systemd[1]: testsuite-54.service: Failed with result 'exit-code'.
[   10.214076] H systemd[1]: testsuite-54.service: Service will not restart (restart setting)
[   10.214177] H systemd[1]: testsuite-54.service: Changed start -> failed
[   10.214348] H systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.DBus.Properties member=PropertiesChanged cookie=112 reply_cookie=0 sig>
[   10.228505] H systemd[1]: testsuite-54.service: Failed to delete cgroup entry from LSM BPF map: No such file or directory
[   10.228719] H systemd[1]: testsuite-54.service: Job 166 testsuite-54.service/start finished, result=failed
[   10.228841] H systemd[1]: Failed to start TESTSUITE-54-CREDS.

```

Maybe this is related?:

```
[    6.770284] H systemd[1]: Detected virtualization kvm.
[    6.770291] H systemd[1]: Detected architecture x86-64.
[    6.770342] H systemd[1]: Detected initialized system, this is not the first boot.
[    6.770351] H systemd[1]: Kernel version 5.17.3-arch1-1, our baseline is 4.15
[    6.780110] H systemd[1]: Hostname set to <H>.
[    6.784049] H systemd[1]: Initializing machine ID from random generator.
[    6.794219] H systemd[1]: Failed to add address 127.0.0.1 to loopback interface: File exists
[    6.794260] H systemd[1]: Failed to add address ::1 to loopback interface: File exists
[    6.794288] H systemd[1]: Successfully brought loopback interface up
[    6.794392] H systemd[1]: Setting '/proc/sys/fs/file-max' to '9223372036854775807
[    6.794399] H systemd[1]: '
[    6.794569] H systemd[1]: No change in value '9223372036854775807
[    6.794574] H systemd[1]: ', suppressing write
[    6.794606] H systemd[1]: Setting '/proc/sys/fs/nr_open' to '2147483640
[    6.794610] H systemd[1]: '
[    6.794649] H systemd[1]: Couldn't write fs.nr_open as 2147483640, halving it.
[    6.794671] H systemd[1]: Skipping bump, value is already larger.
[    6.794797] H systemd[1]: Not importing credentials, $CREDENTIALS_DIRECTORY or $ENCRYPTED_CREDENTIALS_DIRECTORY already set.
[    6.794822] H systemd[1]: Failed to link $CREDENTIALS_DIRECTORY to /run/credentials/@system: File exists
[    6.795267] H systemd[1]: Found cgroup2 on /sys/fs/cgroup/, full unified hierarchy
```

Full journal: https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/11411/artifact//systemd-centos-ci/artifacts_93jgmz47/vagrant-logs.IHC/vagrant-arch-testsuite.FkZ/TEST-54-CREDS/system.journal

test/TEST-54-CREDS/test.sh
```diff
@@ -4,8 +4,15 @@ set -e
 
 TEST_DESCRIPTION="test credentials"
 NSPAWN_ARGUMENTS="--set-credential=mynspawncredential:strangevalue"
+QEMU_OPTIONS="-fw_cfg  name=opt/io.systemd.credentials/myqemucredential,string=othervalue"
```
mrc0mmand: Please tweak this to something like this:

```suggestion
QEMU_OPTIONS="${QEMU_OPTIONS:-} -fw_cfg  name=opt/io.systemd.credentials/myqemucredential,string=othervalue"
```

(and the same for `NSPAWN_ARGUMENTS=` and `KERNEL_APPEND=`), since in CIs we sometimes tweak these options from the outside as well.

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+acquiring and passing credential data to systems and services. The precise
+nature of the credential data is up to applications, but the concept is
+intended to provide systems and services with potentially security sensitive
+cryptographic keys, certificates, passwords, identity information and similar
+types of information. It may also be used as generic infrastructure for
+parameterizing systems and services.
+
+Traditionally, data of this nature has often been provided to services via
+environment variables (which is problematic because by default it is inherited
+down the process tree, has size limitations, and issues with binary data) or
```
behrmann:
```suggestion
environment variables (which is problematic because by default they are inherited
down the process tree, have size limitations, and issues with binary data) or
```
silly nit: The data (it) is inherited down the process tree and there are limitations on its size because the environment variables (they) are inherited down and have those limitations, so I think it would be nicer to pin this on the environment variables directly.

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+simple, unencrypted files on disk. `systemd`'s system and service credentials
+are supposed to provide a better alternative for this purpose. Specifically,
+the following features are provided:
+
+1. Service credentials are acquired at moment of service activation, and
```
behrmann:
```suggestion
1. Service credentials are acquired at the moment of service activation, and
```

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+sha256sum $FOOBARPATH
+
+```
+
+A service defined like this will get the contents of the file
+`/etc/myfoobarcredential.txt` passed as credential `foobar`, which is hence
+accessible under `$CREDENTIALS_DIRECTORY/foobar`. Since we additionally pass
+the path to it as environment variable `$FOOBARPATH` the credential is also
+accessible as the path in that environment variable. When invoked, the service
+will hence show the same SHA256 hash value of the `/etc/myfoobarcredential.txt`
```
behrmann: Something's missing or too much
```suggestion
will hence show the same SHA256 hash value of `/etc/myfoobarcredential.txt`
```
or 
```suggestion
will hence show the same SHA256 hash value of the file at `/etc/myfoobarcredential.txt`
```

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+`${CREDENTIAL_PATH}` in the `ExecStart=` command line to reference the
+credentials directory. For daemons that allow passing credentials via a path
+supplied as environment variabe, use the `%d` specifier in the `Environment=`
+setting to build valid paths to specific credentials.
+
+## Tools
+
+The
+[`systemd-creds`](https://www.freedesktop.org/software/systemd/man/systemd-creds.html)
+tool is provided to work with system and service credentials. It may be use to
```
behrmann:
```suggestion
tool is provided to work with system and service credentials. It may be used to
```

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+access and enumerate system and service credentials, or to encrypt/decrypt credentials
+(for details about the latter, see below).
+
+When invoked from service context, `systemd-creds` passed without further
+parameters will list passed credentials. The `systemd-creds cat xyz` command
+may be used to write the contents of credential `xyz` to standard output. If
+these calls are combined with `--system` switch credentials passed to the
```
behrmann:
```suggestion
these calls are combined with the `--system` switch, credentials passed to the
```

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+Credentials are supposed to be useful for carrying sensitive information, such
+as cryptographic key material. For this kind of data (symmetric) encryption and
+authentication is provided to make storage of the data at rest safer. The data
+may be encrypted and authenticated with AES256-GCM. The encryption key can
+either be one derived from the local TPM2 device, or one stored in
+`/var/lib/systemd/credential.secret`, or a combination of both. If a TPM2
+device is available and `/var/` resides on persistent storage the default
+behaviour is to use the combination of both for encryption, thus ensuring that
+credentials protected this way can only be decrypted and validated on the
+locale hardware and OS installation. Encrypted credentials stored on disk thus
```
behrmann:
```suggestion
local hardware and OS installation. Encrypted credentials stored on disk thus
```

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+
+3. Access to credentials is restricted to the service's user. Unlike
+   environment variables the credential data is not propagated down the process
+   tree. Instead each time a credential is accessed an access check is enforced
+   by the kernel. If the service is using file system namespacing the loaded
+   credential data is invisble to any other services.
+
+4. Service credentials may be acquired from files on disk, may be specified as
+   literal strings in unit files, may be acquired from another service
+   dynamically via an `AF_UNIX` socket, or inherited from the system
```
behrmann: The many "may be"s made me stumble a bit when reading this. I think this could be shorter like this
```suggestion
4. Service credentials may be acquired from files on disk, specified as
   literal strings in unit files, acquired from another service
   dynamically via an `AF_UNIX` socket, or inherited from the system
```
or, what I think made me pause when reading this, add another "may be"
```suggestion
4. Service credentials may be acquired from files on disk, may be specified as
   literal strings in unit files, may be acquired from another service
   dynamically via an `AF_UNIX` socket, or may be inherited from the system
```

behrmann (commented):
The CREDENTIALS docs are super helpful! Here's a couple of typos and style nits.

poettering:
force pushed a new version with all raised points addressed. Let's see if this one passes CI

src/core/import-creds.c
```diff
@@ -0,0 +1,534 @@
...
+        return 0;
+}
+
+static int import_credentials_boot(void) {
+        _cleanup_(import_credentials_context_free) ImportCredentialContext context = {
+                .target_dir_fd = -1,
+        };
+        int r;
+
+        /* systemd-stub will wrap sidecard *.cred files from the UEFI kernel image directory into initrd
```
daandemeyer: `sidecard` => `sidecar`

bluca:
Linter is sad, but all fairly simple stuff:

```
In /github/workspace/test/units/testsuite-54.sh line 44:
    [ "$(systemd-creds --system cat $expected_credential)" = "$expected_value" ]
                                    ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
    [ "$(systemd-creds --system cat "$expected_credential")" = "$expected_value" ]


In /github/workspace/test/units/testsuite-54.sh line 47:
    [ "$(systemd-run -p LoadCredential=$expected_credential --pipe --wait systemd-creds cat $expected_credential)" = "$expected_value" ]
                                       ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                                            ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
    [ "$(systemd-run -p LoadCredential="$expected_credential" --pipe --wait systemd-creds cat "$expected_credential")" = "$expected_value" ]


In /github/workspace/test/units/testsuite-54.sh line 50:
    [ "$(systemd-run -p LoadCredential=miau:$expected_credential --pipe --wait systemd-creds cat miau)" = "$expected_value" ]
                                            ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
    [ "$(systemd-run -p LoadCredential=miau:"$expected_credential" --pipe --wait systemd-creds cat miau)" = "$expected_value" ]


In /github/workspace/test/units/testsuite-54.sh line 53:
    [ "$(systemd-run -p LoadCredential=$expected_credential -p SetCredential=$expected_credential:zzz --pipe --wait systemd-creds cat $expected_credential)" = "$expected_value" ]
                                       ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                             ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                                                                                      ^------------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
    [ "$(systemd-run -p LoadCredential="$expected_credential" -p SetCredential="$expected_credential":zzz --pipe --wait systemd-creds cat "$expected_credential")" = "$expected_value" ]
```

man/systemd.exec.xml
```diff
@@ -3039,20 +3039,30 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
...
+        which may be used to propagate credentials from an invoking environment (e.g. a container manager
+        that invoked the service manager) into a service. If no matching system credential is found, the
+        directories `/etc/credstore/`, `/run/credstore/` and `/usr/lib/credstore/` are searched for files under
+        the credential's name — which hence are recommended locations for credential data on disk. If
+        <varname>LoadCredentialEncrypted=</varname> is used <filename>/run/credstore.encrypted/</filename>,
+        <filename>/etc/credstore.encrypted/</filename>, and
+        <filename>/usr/lib/credstore.encrypted/</filename> are searched as well.</para>
+
+        <para>If the file system path is omitted it is chosen identical to the credential name, i.e. this is
+        a terse way do declare credentials to inherit from the service manager into a service. This option
```
daandemeyer: do => to

man/systemd.exec.xml
```diff
@@ -3077,10 +3087,23 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
...
+        e.g. <literal>ExecStart=cat ${CREDENTIALS_DIRECTORY}/mycred</literal>. In order to reference the path
+        a credential may be read from within a <varname>Environment=</varname> line use
+        <literal>%d/mycred</literal>, e.g. <literal>Environment=MYCREDPATH=%d/mycred</literal>.</para>
 
         <para>Currently, an accumulated credential size limit of 1 MB per unit is enforced.</para>
 
+        <para>The service manager itself may receive system credentials that can be propagated to services
+        from a hosting container manager or VM hypervisor. See the <ulink
+        url="https://systemd.io/CONTAINER_INTERFACE">Container Interface</ulink> documentation for details
+        about the former. For the latter use the <command>qemu</command> <literal>fw_cfg</literal> node
```
daandemeyer: For the latter, ...

man/systemd.exec.xml
```diff
@@ -3077,10 +3087,23 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
...
+        <literal>opt/io.systemd.credentials/</literal>. Example qemu switch: <literal>-fw_cfg
```
daandemeyer, line 3101: Why prefix with opt/? Is that just common convention?

man/systemd.exec.xml
```diff
@@ -3077,10 +3087,23 @@ StandardInputData=SWNrIHNpdHplIGRhIHVuJyBlc3NlIEtsb3BzLAp1ZmYgZWVtYWwga2xvcHAncy
...
+        name=opt/io.systemd.credentials/mycred,string=supersecret</literal>. They may also be specified on
+        the kernel command line using the <literal>systemd.set-credential=</literal> switch (see
```
daandemeyer: set-credential => set_credential?

src/core/import-creds.c
```diff
@@ -0,0 +1,549 @@
...
+                        _cleanup_close_ int cfd = -1, nfd = -1;
+                        _cleanup_free_ char *n = NULL;
+                        const char *e;
+                        struct stat st;
+
+                        e = endswith(d->d_name, ".cred");
+                        if (!e)
+                                continue;
+
+                        n = strndup(d->d_name, e - d->d_name);
```
daandemeyer, line 192: This removes .cred from the filename right? I'd add a comment as this isn't immediately obvious. Or maybe a helper function with a clear name would help to explain the intent here.

src/core/import-creds.c
```diff
@@ -0,0 +1,549 @@
...
+        ImportCredentialContext *c = ASSERT_PTR(data);
+        _cleanup_free_ char *n = NULL;
+        _cleanup_close_ int nfd = -1;
+        const char *colon;
+        size_t l;
+        int r;
+
+        assert(key);
+
+        if (!proc_cmdline_key_streq(key, "systemd.set-credential"))
```
daandemeyer: Mismatch with documentation (which lists systemd.set_credential). I guess this should be systemd.set_credential as well.

src/core/import-creds.c
```diff
@@ -0,0 +1,549 @@
...
+        return 0;
+}
+
+int import_credentials(void) {
+        const char *received_creds_dir = NULL, *received_encrypted_creds_dir = NULL;
+        bool envvar_set = false;
+        int r, q;
+
+        r = get_credentials_dir(&received_creds_dir);
+        if (r < 0 && r != -ENXIO)
```
daandemeyer: Why ignore ENXIO here? That seems like it deserves a comment.

src/core/import-creds.c
```diff
@@ -0,0 +1,549 @@
...
+                log_debug_errno(r, "Failed to determine credentials directory, ignoring: %m");
```
daandemeyer: Can't we log at warning level here and elsewhere in this function? That shouldn't be too noisy I think

src/core/execute.c
```diff
@@ -2595,6 +2595,42 @@ static int write_credential(
...
+
+                if (strv_extend_strv(&l, CONF_PATHS_STRV("credstore.encrypted"), /* filter_duplicates= */ true) < 0)
+                        return NULL;
+        }
+
+        if (params->received_credentials_directory)
+                if (strv_extend(&l, params->received_credentials_directory) < 0)
+                        return NULL;
+
+        if (strv_extend_strv(&l, CONF_PATHS_STRV("credstore"), /* filter_duplicates= */ true) < 0)
```
daandemeyer, line 2622: This stuff is quite systemd specific, shouldn't we make credstore a subdirectory of etc/systemd, /usr/lib/systemd, ... then?

src/core/execute.c
```diff
@@ -2624,10 +2661,24 @@ static int load_credential(
         assert(write_dfd >= 0);
         assert(left);
 
-        if (path_is_absolute(path) || read_dfd >= 0) {
-                /* If this is an absolute path (or a directory fd is specifier relative which to read), read
-                 * the data directly from it, and support AF_UNIX sockets */
+        if (read_dfd >= 0) {
+                /* If a directory fd is specified, then read the file directly from that dir. In this case we
+                 * won't do AF_UNIX stuff (we simply don't want to recursively do IPC). It's OK if a file
```
daandemeyer: How does a directory fd imply recursive IPC?

src/core/execute.c
```diff
@@ -2636,25 +2687,50 @@ static int load_credential(
...
+        if (search_path) {
+                STRV_FOREACH(d, search_path) {
+                        _cleanup_free_ char *j = NULL;
+
+                        j = path_join(*d, path);
+                        if (!j)
+                                return -ENOMEM;
+
+                        r = read_full_file_full(
+                                        AT_FDCWD, j,
```
daandemeyer: The paths are absolute if I'm correct, so AT_FDCWD is just a default value here?

src/core/execute.c
```diff
@@ -2636,25 +2687,50 @@ static int load_credential(
...
+                                        UINT64_MAX,
+                                        maxsz,
+                                        flags,
+                                        bindname,
```
daandemeyer: Since we don't support the AF_UNIX stuff in this mode, maybe pass NULL as bindname to more clearly indicate that? I got a little confused here.

src/creds/creds.c
```diff
@@ -60,63 +60,72 @@ static const char* transcode_mode_table[_TRANSCODE_MAX] = {
...
         int r;
 
-        if (arg_system) {
-                _cleanup_free_ char *cd = NULL;
-
-                r = getenv_for_pid(1, "CREDENTIALS_DIRECTORY", &cd);
+        assert(ret_dir);
+
+        if (arg_system)
+                /* PID 1 creates a fixed path symlinks for system credentials */
```
daandemeyer: This comment reads kinda weird, should probably be reworded

src/creds/creds.c
```diff
@@ -60,63 +60,72 @@ static const char* transcode_mode_table[_TRANSCODE_MAX] = {
...
-                        return -EINVAL;
+                        return log_error_errno(r, "Failed to get credentials directory: %m");
+        }
 
-                j = path_join("/proc/1/root", cd);
-                if (!j)
-                        return -ENOMEM;
+        d = opendir(p);
+        if (!d) {
+                if (arg_system && errno == ENOENT)  /* No such dir? Then no creds where passed. */
```
daandemeyer: Why conditionalize this on arg_system? Is the directory guaranteed to be there for service specific directories?

DaanDeMeyer (commented):
Didn't get to everything yet, will finish later. Looks really good though

man/systemd.exec.xml
poettering: yes, as per qemu docs. they reverse domain name notation too.

https://github.com/qemu/qemu/blob/master/docs/specs/fw_cfg.txt#L258

man/systemd.exec.xml
poettering: it's actually the same. but you are right we usually write kernel cmdline options with an underscore

src/core/import-creds.c
poettering: doesn't actually matter. kernel's own cmdline handling considers underscore and dash as equal and we followed that rule and do so to. That's what proc_cmdline_key_streq() does. 

That said we should use the same syntax here, to minimize surprises.

src/core/execute.c
poettering: I thought about this, but it wasn't clear to me, and I figured there might be value in keeping this generic, given that the concept is not really specific to systemd. Similar systems do exist in container solution (though usually not as elaborate and careful). So I figure we can just leave the door open for others to go to this dir too without necessarily acknowledging systemd.

src/creds/creds.c
poettering: the per-service stuff relies on an env var, the per-system stuff uses hardcode path. if the env var points to non-existing path something is weird, and this should be considered an error. if the fixed path in the system case does not exist that's however totally ok, it just means nothing was passed.

will extend comment.

src/core/execute.c
poettering: yeah, most xyz_at() apis that operate on dir fds work that way. will add comment.

src/core/execute.c
poettering: will extend comment.

poettering:
force pushed a new version with all of @DaanDeMeyer's points addressed.

ptal!

src/core/import-creds.c
```diff
@@ -0,0 +1,551 @@
...
+ * generators invoked by it) can acquire credentials from outside, to mimic how we support it for containers,
+ * but on VM/physical environments.
+ *
+ * This does three things:
+ *
+ * 1. It imports credentials picked up by sd-boot (and placed in the /.extra/credentials/ dir in the initrd)
+ *    and puts them in /run/credentials/@encrypted/. Note that during the initrd→host transition the initrd root
+ *    file system is cleaned out, thus it is essential we pick up these files before they are deleted. Note
+ *    that these credentials originate from an untrusted source, i.e. the ESP and are not
+ *    pre-authenticated. They still have to be authenticated before use.
```
daandemeyer, line 33: I'm not sure I follow this bit. If I understand correctly, this directory is intended for use with `LoadCredentialEncrypted`. Does that mean all credentials that end up in this directory must be encrypted with `sd-creds encrypt`? And does that mean systemd still handles all that for the user before passing the credentials to a service?

src/creds/creds.c
```diff
@@ -330,19 +376,58 @@ static int verb_cat(int argc, char **argv, void *userdata) {
...
 
-                r = read_full_file_full(
-                                dirfd(d), *cn,
-                                UINT64_MAX, SIZE_MAX,
-                                READ_FULL_FILE_SECURE|READ_FULL_FILE_WARN_WORLD_READABLE,
-                                NULL,
-                                (char**) &data, &size);
-                if (r < 0) {
+                /* Look both in regular and in encrypted credentials */
+                for (encrypted = 0; encrypted < 2; encrypted ++) {
```
daandemeyer, line 380: Spurious whitespace here

src/creds/creds.c
```diff
@@ -60,63 +60,76 @@ static const char* transcode_mode_table[_TRANSCODE_MAX] = {
 
 DEFINE_PRIVATE_STRING_TABLE_LOOKUP_FROM_STRING(transcode_mode, TranscodeMode);
 
-static int open_credential_directory(DIR **ret) {
-        _cleanup_free_ char *j = NULL;
+static int open_credential_directory(
+                DIR **ret_dir,
+                const char **ret_prefix,
+                bool encrypted) {
```
daandemeyer, line 66: Shouldn't this be the first argument since we always put return arguments at the end of the signature?

src/creds/creds.c
```diff
@@ -330,19 +376,58 @@ static int verb_cat(int argc, char **argv, void *userdata) {
...
+                if (encrypted >= 2) { /* Found nowhere */
+                        log_error_errno(SYNTHETIC_ERRNO(ENOENT), "Credential '%s' not set.", *cn);
+                        if (ret >= 0)
+                                ret = -ENOENT;
+
                         continue;
                 }
 
+                if (encrypted) {
+                        _cleanup_(erase_and_freep) void *plaintext = NULL;
```
daandemeyer, line 414: Do we really need the _cleanup_ here since if decrypt_credential_and_warn() fails plaintext will never be set in the first place? I support keeping it though since it'll be less error prone if we add an extra failure path here in the future.

DaanDeMeyer (approved):
LGTM, some more minor stuff and one more question

src/core/kmod-setup.c
```diff
@@ -80,6 +81,10 @@ static bool has_virtio_rng(void) {
 
         return r > 0;
 }
+
+static bool in_qemu(void) {
+        return IN_SET(detect_vm(), VIRTUALIZATION_KVM, VIRTUALIZATION_QEMU);
```
bluca, line 86: mmh are we sure about `VIRTUALIZATION_KVM`? There are other hypervisors that run on KVM, eg: https://github.com/cloud-hypervisor/cloud-hypervisor would those be matched too here?

poettering: docs clearly say "kvm" means "qemu+kvm".

Quoting systemd-detect-vrit man page:

> "Linux KVM kernel virtual machine, in combination with QEMU. Not used for other virtualizers using the KVM interfaces, such as Oracle VirtualBox or Amazon EC2 Nitro, see below."

src/creds/creds.c
poettering: hmm, where?

src/core/import-creds.c
poettering: well it kinda means that, but also not.

It's more about ensuring that automatic propagation doesn't happen for stuff we can't trust because it comes from an untrusted source. i.e. we should not auto-propagate stuff into services if it potentially contains rogue stuff. 

What do I mean by "auto-propagate"? mostly if LoadCredential= is used with just a cred name (ie. without specifying an explicit source), so that the credential is searched automatically in /etc/credstore/ or among the system creds. That said, it's actual OK if we propagate such an untrusted cred if the unit file makes this explicit. How can it make it explicit? By specifying a full, absolute source path. Or in other words:

A simple line like this:
```
LoadCredential=foo
```
should auto-search the credential to propagate from stuff in /etc/credstore/ and among creds passed in from a trusted hypervisor/container manager or so.

vs.

But for loading untrusted creds to pass to a service, we should require the developer to explicitly indicate that it's OK to source from an unencrypted (i.e. untrusted) source, via a line like this:

```
LoadCredential=foo:/run/credentials/@encrypted/foo
```

This line will basically upgrade the untrusted system credential "foo" into a trusted  service credential "foo" *without* validation.

A much better way of course is this:

```
LoadCredentialEncrypted=foo
```
 
This will auto-search and then validate/decrypt the credential and pass it on. This line hence takes an untrusted system credential "foo" and turns it into a trusted service credential *with* validation.

or in other words: we place trusted/untrusted credentials in two separate dirs so that the creds passed in that way are not picked up by code that expects them to be trusted even though they are not trusted at all, because they are loaded from an untrusted source (i.e. the ESP).we want to make sure that we have two distinct pools: one which contains "trusted" credentials, i.e. which can just be used, as the come from a trusted source. and "untrusted" credentials, that still need to be validated before use. It's OK if credentials in the "trusted" pool are consumed with little restrictions and hence LoadCredential= will just search the stuff from the trusted pool. However stuff from the "untrusted" pool should not consumed that easily, it either requires validation against something, or the user should explicitly say that it is OK to propagate this cred without authentication. 

Why did I call the dir "encrypted" rather than "untrusted"? Initially while I worked on it I actually named it that way. But I changed this to "encrypted" because we have so far not used the term "trust" in code for this purpose. So far we only distinguish encrpted and not encrypted creds, and the authentication is implicit to the encryption mechanism. (i.e. we actually used authenticated encryption, but this is only ever mentioned in the docs, not in the naming of the functionality). So in order to simplify things and in the hope that people will more likely use this properly I want to stick to this: there are only two types of creds, and the one way to get from one to the other is through LoadCredentialEncrypt=, and only then things are nice, and automatic.

does that make any sense?

src/core/kmod-setup.c
poettering: (I mean, of course, in retrospect we should nevre have added the explicit "kvm" enum value ere and just make it "qemu", but back then it wasn't clear that kvm would be used by a number of different hypervisors one day)

src/creds/creds.c
daandemeyer: Between `encrypted` and `++`?

src/creds/creds.c
poettering: we do not. but it also doesn't hurt... 

I mean, sooner or later the code might get more complex, and we add additional exit paths. I think it's quite valuable if we can be reasonably sure that we won't regress here even then, without having to take care of updating the exit paths.

I mean, I think optimzing the exit paths of functions that are not inner loops is not a priority at all, but in particular so if the data involved is actually secret/plaintext data. I think for that kind of data we should basicallly always add an `erase` destructor declaration even if there's really no way it could ever be triggered, simply to indicate we know it' sensitive, and whatever happens we'll always consider it sensitive...

src/core/import-creds.c
daandemeyer: That makes sense, thank you for the detailed explanation.

One thing I don't completely understand is why credentials picked up by sd-boot would be untrusted. What exactly distinguishes those credentials from the trusted ones that makes them untrusted?

poettering:
Ok, let's merge this. I'll fix the two issues @DaanDeMeyer pointed out with a follow-up commit.

I kinda love that the CI is green for this, I need to cease the opportunity to get something as complex as this with n all green CI merged ;-)

src/core/import-creds.c
poettering: > One thing I don't completely understand is why credentials picked up by sd-boot would be untrusted. What exactly distinguishes those credentials from the trusted ones that makes them untrusted?

In any environment that uses FDE or verity or something like that I think we can assume the areas of the file hierarchy that are only writable to root can be reasonably considered "trusted", as then there's protection that neither online nor offline (at least to some degree) it can be changed.

This is quite different from data stored in the ESP: it's trivially accessible offline, there is no encryption or integrity checking, it's a FAT file system you can trivially modify.

Of course, if you do not employ FDE then there's no difference in the guarantees you get for /etc/ and for the ESP but I think we should focus on secure scenarios where it is used, and thus distinguish stuff that never comes with any sort of trust guarantee from the stuff that will on systems using FDE comes with them (even if people can turn that off in which case it is not any better than the other stuff)

daandemeyer: Got it, I didn't think of offline modification, that makes total sense

src/core/kmod-setup.c
bluca: Well, the KVM CPUID stuff predates systemd-detect-virt, I think. Running under cloud-hypervisor we indeed detect kvm:

```
$ ./cloud-hypervisor-static --kernel ./hypervisor-fw --disk path=/tmp/focal-server-cloudimg-amd64.raw --cpus boot=2 --memory size=2048M
cloud-hypervisor: 5.568027101s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x87
cloud-hypervisor: 5.606556807s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x2f9
cloud-hypervisor: 5.606595571s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x2f9
cloud-hypervisor: 5.606628522s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x2f9
cloud-hypervisor: 5.606670021s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x2f9
cloud-hypervisor: 5.60670979s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x2f9
cloud-hypervisor: 5.606730758s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x2f9
cloud-hypervisor: 5.60681496s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x3e9
cloud-hypervisor: 5.606844019s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x3e9
cloud-hypervisor: 5.606869118s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x3e9
cloud-hypervisor: 5.606894391s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x3e9
cloud-hypervisor: 5.6069225s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x3e9
cloud-hypervisor: 5.606949647s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x3e9
cloud-hypervisor: 5.607026592s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x2e9
cloud-hypervisor: 5.607053484s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x2e9
cloud-hypervisor: 5.60707703s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x2e9
cloud-hypervisor: 5.607101702s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x2e9
cloud-hypervisor: 5.607126966s: <vcpu0> WARN:vmm/src/vm.rs:470 -- Guest PIO read to unregistered address 0x2e9
cloud-hypervisor: 5.607152985s: <vcpu0> WARN:vmm/src/vm.rs:495 -- Guest PIO write to unregistered address 0x2e9

Ubuntu 20.04.4 LTS ubuntu hvc0

ubuntu login: root
Password: 
Welcome to Ubuntu 20.04.4 LTS (GNU/Linux 5.4.0-109-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Mon May  2 15:02:13 UTC 2022

  System load: 0.23              Memory usage: 9%   Processes:       122
  Usage of /:  67.5% of 1.96GB   Swap usage:   0%   Users logged in: 0

1 update can be applied immediately.
To see these additional updates run: apt list --upgradable



The programs included with the Ubuntu system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.

root@ubuntu:~# SYSTEMD_LOG_LEVEL=debug systemd-detect-virt 
No virtualization found in DMI
Virtualization found, CPUID=KVMKVMKVM
Found VM virtualization kvm
kvm
```

So this would break things, no?

poettering: well, i think if there is a VM we currently don't detect correctly, we should fix that.

Note that various other implementations that use kvm actually pass a different CPUID than "KVMKVMKVM". I think it's kinda sloppy if "could-hypervisor" uses KVMKVMKVM for that...

But anyway, I think we should stick to our definition of "kvm", and add detection entries for other VM hypervisors as necessary. We followed this rule so far, I see no reason to depart from it now.

poettering: note that the misdetection isn't too bad here: we'll try to load qemu_fw_cfg, and that might not do anything, but this is handled gracefully on all levels. i.e. we do a bit of unnecessary work here on any hypervisor misrecognized as being kvm+qemu, but it doesn't break anything, it just wastes a a handful of CPU cycles...

docs/CREDENTIALS.md
```diff
@@ -0,0 +1,388 @@
...
+
+2. Service credentials are accessible to service code as regular files, the
+   path to access them is derived from the environment variable
+   `$CREDENTIALS_DIRECTORY`.
+
+3. Access to credentials is restricted to the service's user. Unlike
+   environment variables the credential data is not propagated down the process
+   tree. Instead each time a credential is accessed an access check is enforced
+   by the kernel. If the service is using file system namespacing the loaded
+   credential data is invisble to any other services.
```
bluca, line 37: /invisble/invisible/

poettering:
Given how trivial I took liberty to push fixes for the remaining issues @bluca and @DaanDeMeyer found directly to git main.

https://github.com/systemd/systemd/commit/0ba05e754cbd028fb8487564cfd90aa18318f516
https://github.com/systemd/systemd/commit/07a442ef8b7822af2bbddc5b8938525e39de9981

Issue #23180, opened 2022-04-24
Test failure with 250.4 on HPPA


thesamesam:
**systemd version the issue has been seen with**
 > 250.4

**Used distribution**
 > Gentoo

**Linux kernel version used** (`uname -a`)

 > `5.10.95`

**CPU architecture issue was seen on**
 > HPPA

**Expected behaviour you didn't see**
 > All tests pass, including `test-seccomp`.

**Unexpected behaviour you saw**
 > `test-seccomp` failed.

**Steps to reproduce the problem**
 > Run the test suite on HPPA.

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
518/648 test-seccomp                              FAIL             1.69s   killed by signal 6 SIGABRT
00:28:05 MALLOC_PERTURB_=7 SYSTEMD_LANGUAGE_FALLBACK_MAP=/var/tmp/portage/sys-apps/systemd-250.4-r1/work/systemd-stable-250.4/src/locale/language-fallback-map PATH=/var/tmp/portage/s
ys-apps/systemd-250.4-r1/work/systemd-stable-250.4-.hppa:/var/tmp/portage/sys-apps/systemd-250.4-r1/temp/python3.9/bin:/usr/lib/portage/python3.9/ebuild-helpers/xattr:/usr/lib/portag
e/python3.9/ebuild-helpers:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin /var/tmp/portage/sys-apps/systemd-250.4-r1/work/systemd-stable-250.4-.hppa/test-secco
mp
----------------------------------- output -----------------------------------
stderr:
/* test_restrict_suid_sgid */
Not privileged, skipping test_restrict_suid_sgid
/* test_lock_personality */
Not privileged, skipping test_lock_personality
/* test_native_syscalls_filtered */
Not privileged, skipping test_native_syscalls_filtered
/* test_load_syscall_filter_set_raw */
Not privileged, skipping test_load_syscall_filter_set_raw
/* test_restrict_archs */
Not privileged, skipping test_restrict_archs
/* test_memory_deny_write_execute_shmat */
Not privileged, skipping test_memory_deny_write_execute_shmat
/* test_memory_deny_write_execute_mmap */
Not privileged, skipping test_memory_deny_write_execute_mmap
/* test_restrict_realtime */
Not privileged, skipping test_restrict_realtime
/* test_restrict_address_families */
[...]

PR #23181, opened 2022-04-24
Add seccomp support for PARISC (HPPA)

thesamesam:
Fixes: #23180


src/shared/seccomp-util.c
```diff
@@ -132,6 +137,14 @@ const char* seccomp_arch_to_string(uint32_t c) {
                 return "mips64-le";
         case SCMP_ARCH_MIPSEL64N32:
                 return "mips64-le-n32";
+#ifdef SCMP_ARCH_PARISC
```
thesamesam, line 140: I've opted to keep the #ifdef guards but they're trivial to remove. Support for HPPA landed in libseccomp in 2016 but was released in libseccomp 2.4.0 (2019).

yuwata: We have similar ifdef for `SCMP_ARCH_RISCV64`. So, that's OK. But please add similar `#ifdef` at all places where `SCMP_ARCH_PARISC` or friends are used.

src/shared/seccomp-util.c
```diff
@@ -80,6 +80,11 @@ uint32_t seccomp_local_archs[] = {
                 SCMP_ARCH_MIPSEL64,
                 SCMP_ARCH_MIPS64N32,
                 SCMP_ARCH_MIPSEL64N32, /* native */
+#elif defined(__hppa64__)
+                SCMP_ARCH_PARISC,
+                SCMP_ARCH_PARISC64,    /* native */
+#elif defined(__hppa__)
+                SCMP_ARCH_PARISC,
```
yuwata: Please add `&& defined (SCMP_ARCH_PARISC)` or similar here.

yuwata (changes requested):
Looks mostly good.

yuwata:
BTW, please read http://systemd.io/PORTING_TO_NEW_ARCHITECTURES/, and if you'd like, please also update them.
1 is already addressed, 4 and 5 are addressed by this PR. I am not sure about the other items.

yuwata:
Also, please drop the Signed-off-by from the commit message. We do not use it.

thesamesam:
Thanks for the feedback. I'll get on it soon!

yuwata:
Ping?

thesamesam:
Let me finish this off. I'll take a look tonight. I was sort of waiting to see if I could get away without the exec stack stuff but while there's been kernel changes, not seen anything else there yet, so..

yuwata (approved):

thesamesam:
Two failures are unrelated (timeouts). focal-arm64 seems stuck.

PR #23187, opened 2022-04-25
mkosi: Remove locale workaround

daandemeyer:
Let's expect mkosi to set a valid UTF-8 locale and fail early if it
doesn't instead of messing with this ourselves.

We also pull in the latest mkosi commit which guarantees that the
locale will be set to C.UTF-8.

### DIFF
--- a/.github/workflows/mkosi.yml
+++ b/.github/workflows/mkosi.yml
@@ -49,7 +49,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
-    - uses: systemd/mkosi@93098e2406e12ea7f06f962d4808952b8a06d345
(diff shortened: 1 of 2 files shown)


bluca (approved):

mrc0mmand:
CI seems to be unhappy:

```
 (11/12) Warn about old perl modules
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "C.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
(12/12) Updating the info directory file...
Generating locales...
  C.UTF-8... done
Generation complete.
*** locale does not support UTF-8. ***
```

DaanDeMeyer:
Yeah we need an nspawn fix to make this work, closing for now

PR #23195, opened 2022-04-26
core: Release resources of services in SERVICE_EXITED state

### LINKED ISSUE #23193: systemd keeps standard file descriptors open of services with RemainAfterExit=yes
 > main branch

**Used distribution**
 > /

 > /

 > /

 > Standard file descriptors are released after the service configured with RemainAfterExit=yes enters the SERVICE_EXITED state

 > Standard file descriptors stay open until the unit is garbage collected

 > Write a program that creates a pipe, starts a transient unit with RemainAfterExit=yes and configures one end of the pipe as the standardoutput file descriptor of the transient service using StandardOutputFileDescriptor. Then wait for EOF in that program on the pipe in the same program. EOF will only be generated once the transient service is stopped explicitly and not when the transient service command exits.

Relevant code is [here](https://github.com/systemd/systemd/blob/c84fc00b371cb30b1b0d4bfde015662e1a8542a7/src/core/unit.c#L411)

I can't immediately think of a use case where we would want to keep the service resources (standard file descriptors and fd store) intact after the service exits so to me it makes sense to extend the check I linked to include service units in the SERVICE_EXITED state.

```text
…
```

daandemeyer:
Services with RemainAfterExit=yes will finish but won't get gc-ed
just yet. Let's make sure we release the resources of these
services early instead of keeping them alive until the service is
explicitly stopped.

Fixes #23193

### DIFF
--- a/src/core/unit.c
+++ b/src/core/unit.c
@@ -408,8 +408,9 @@ bool unit_may_gc(Unit *u) {
 
         state = unit_active_state(u);
 
-        /* If the unit is inactive and failed and no job is queued for it, then release its runtime resources */
-        if (UNIT_IS_INACTIVE_OR_FAILED(state) &&
+        /* If the unit is inactive, failed or is a service with RemainAfterExit=yes and no job is queued for it, then release its runtime resources */
+        if ((UNIT_IS_INACTIVE_OR_FAILED(state) ||
+             (u->type == UNIT_SERVICE && SERVICE(u)->state == SERVICE_EXITED)) &&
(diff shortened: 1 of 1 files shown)


poettering:
As mentioned elsewhere, this contradicts the express goal with the fdstore: that it is kept around for restarts

DaanDeMeyer:
Closing pending further discussion

poettering:
I mean, we could make this behaviour configurable. I mean I do sympathize a bit wiht the usecase... But not sure it would be worth it...

PR #23204, opened 2022-04-27
meson: add install_tag and build alias for systemd-boot, libsystemd and libudev

bluca:
Allows to 'meson install --tags sd-boot --no-rebuild' to install only the EFI
binaries, skipping the rest, for a very quick build:

$ ninja src/boot/efi/linuxx64.efi.stub
[21/21] Generating src/boot/efi/linuxx64.efi.stub with a custom command
$ ninja src/boot/efi/systemd-bootx64.efi
[10/10] Generating src/boot/efi/systemd-bootx64.efi with a custom command
$ DESTDIR=/tmp/foo meson install --tags sd-boot --no-rebuild
Installing src/boot/efi/systemd-bootx64.efi to /tmp/foo/usr/lib/systemd/boot/efi
Installing src/boot/efi/linuxx64.elf.stub to /tmp/foo/usr/lib/systemd/boot/efi
Installing src/boot/efi/linuxx64.efi.stub to /tmp/foo/usr/lib/systemd/boot/efi

Requires Meson 0.60 to be used, prints a warning for unknown keyword
in earlier versions, but there's no failure

### DIFF
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -428,7 +428,7 @@ foreach tuple : [['systemd-boot@0@.@1@', systemd_boot_objects, false, 'systemd-b
                 install_tag: tuple[3],
                 install_dir : bootlibdir)
 
-        custom_target(
+        efi = custom_target(
                 tuple[0].format(efi_arch[0], 'efi'),
                 input : elf,
                 output : tuple[0].format(efi_arch[0], 'efi'),
@@ -449,4 +449,6 @@ foreach tuple : [['systemd-boot@0@.@1@', systemd_boot_objects, false, 'systemd-b
                 install : true,
                 install_tag: tuple[3],
                 install_dir : bootlibdir)
+
+        alias_target(tuple[3], efi)
 endforeach


bluca:
This trips the CI because we pass `--fatal-meson-warnings`, suggestions?

src/boot/efi/meson.build
```diff
@@ -446,5 +447,6 @@ foreach tuple : [['systemd-boot@0@.@1@', systemd_boot_objects, false],
                            efi_format,
                            '@INPUT@', '@OUTPUT@'],
                 install : true,
+                install_tag: 'sd-boot',
```
poettering, line 450: probably should have another tag though. sd-stub?

src/boot/efi/meson.build
```diff
@@ -425,6 +425,7 @@ foreach tuple : [['systemd-boot@0@.@1@', systemd_boot_objects, false],
                            '-lgnuefi',
                            '-lgcc'],
                 install : tuple[2],
+                install_tag: 'sd-boot',
```
poettering: the tag to be used should probably be specified in the array we iterate through further up, so that sd-stub and sd-boot can have two distinct tags

poettering:
lovely meson feature, i wasn't aware of this. i wonder if we could drop some build system conditionalizations now that this exists

poettering:
@keszybz should probably comment on this

keszybz:
> This trips the CI because we pass `--fatal-meson-warnings`, suggestions?

I'd tweak CI so that it passes `--fatal-meson-warnings` only if meson is at least some specified version (e.g. 0.60). The param was added to catch warnings with new meson versions. We are fine with a warning on old meson versions.

In general, I think this is a very nice feature. We should probably figure out what tags we'd want to use. We'd probably want on for `libsystemd`, maybe also one for the standalone variants of tools, and one for tests, man pages, html pages.

BTW.: can we have a tag that is not installed by default? This would allow us to drop the config option `-Dinstall-tests`.

bluca:
> > This trips the CI because we pass `--fatal-meson-warnings`, suggestions?
> 
> I'd tweak CI so that it passes `--fatal-meson-warnings` only if meson is at least some specified version (e.g. 0.60). The param was added to catch warnings with new meson versions. We are fine with a warning on old meson versions.
> 
> In general, I think this is a very nice feature. We should probably figure out what tags we'd want to use. We'd probably want on for `libsystemd`, maybe also one for the standalone variants of tools, and one for tests, man pages, html pages.

I've added a tag for libsystemd and libudev, as they are public APIs so it makes some sense to have separately (likely shipped by different packages, so this might help). I'd rather not add more right now to avoid too much bikeshedding - I think the sd-boot stuff is useful given the discussion on the mailing list, and also the recent change to make the 32bit cross-build possible, so I think we should get these in by themselves.

Besides, the documentation says there are default tags, and for example if we switch to `install_man` we'll get an automated `install --tags man` created for us, but we should look into that separately I think.

https://mesonbuild.com/Installing.html#installation-tags

> BTW.: can we have a tag that is not installed by default? This would allow us to drop the config option `-Dinstall-tests`.

Don't know, but I guess we'll need to keep support for old stuff until we bump the required version to 0.60 for now.

evverx:
> I'd tweak CI so that it passes --fatal-meson-warnings only if meson is at least some specified version (e.g. 0.60)

The CI works as expected. `meson-0.62` complained that "Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target" so this patch produces warnings with all meson versions basically and to fix it the meson version in `meson.build` should be bumped from `>= 0.53.2` to `>= 0.60`.

.github/workflows/build_test.sh
```diff
@@ -123,7 +123,9 @@ for args in "${ARGS[@]}"; do
     #   src/boot/efi/meson.build:52:16: ERROR: Fatal warnings enabled, aborting
     # when LINKER is set to lld so let's just not turn meson warnings into errors with lld
     # to make sure that the build systemd can pick up the correct efi-ld linker automatically.
-    if [[ "$LINKER" != lld ]]; then
+    # We use some features (like install_tag) that were introduced in 0.60, but that don't
+    # break running with older versions, so don't enable this with older versions
+    if [[ "$LINKER" != lld ]] && dpkg --compare-versions "$(meson --version)"  ge 0.60; then
```
evverx: `meson` is installed with `pip` there using https://github.com/systemd/systemd/blob/main/.github/workflows/requirements.txt so `apt` and `dpkg` can't be used to compare versions.

bluca: it's just used as a version string comparator, as it knows all the "gotchas" already, doesn't matter where the strings come from

bluca:
> > I'd tweak CI so that it passes --fatal-meson-warnings only if meson is at least some specified version (e.g. 0.60)
> 
> The CI works as expected. `meson-0.62` complained that "Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target" so this patch produces warnings with all meson versions basically and to fix it the meson version in `meson.build` should be bumped from `>= 0.53.2` to `>= 0.60`.

Yeah checking the version is not enough - I've removed the parameter for now, with a FIXME to add it back once we bump the minimum version in the future

evverx:
> with a FIXME to add it back once we bump the minimum version in the future

The point of this check among other things is to keep `meson.build` in sync with the features the `systemd` build system relies on so it tags are here to stay the build system should fail when they don't work.

bluca:
> > with a FIXME to add it back once we bump the minimum version in the future
> 
> The point of this check among other things is to keep `meson.build` in sync with the features the `systemd` build system relies on so if tags are here to stay the build system should fail when they don't work.

No need to break backward compatibility, it's an optional cmdline helper, it's fine if it doesn't work

evverx:
> it's an optional cmdline helper, it's fine if it doesn't work

It breaks scripts where `--fatal-meson-warnings` is used though so it isn't exactly silently ignored everywhere.

keszybz:
I don't think we can require 0.60 yet. F35 has 0.59.4, debian stable (bullseye) has 0.56.2. But meh, we can live with a warning on older systems.

keszybz:
E: Failed to fetch http://ppa.launchpad.net/upstream-systemd-ci/systemd-ci/ubuntu/pool/main/t/tpm2-tss/libtss2-esys0_2.3.3-1~bpo20.04.1_amd64.deb  Cannot initiate the connection to ppa.launchpad.net:80 (2620:2d:4000:1001::111). - connect (101: Network is unreachable) Could not connect to ppa.launchpad.net:80 (185.125.189.231), connection timed out [IP: 185.125.189.231 80]

bluca:
It's just a warning, we can live with it for a couple of years

evverx:
> It's just a warning, we can live with it for a couple of years

@bluca it's not a warning when `--fatal-meson-warnings` is used. I'm not sure why because of that niche feature scripts using recent meson versions supported the tags have to be edited anyway to ignore all warnings for a couple of years.

Tachi107:
> It's just a warning, we can live with it for a couple of years

You're currently using `install_tags` unconditionally, right? If so, I believe that builds will fail if systemd is built with Meson versions older than 0.60.0, as they would try to use a kwarg that doesn't yet exist.

evverx:
> there are already warnings printed for other unrelated reasons

There are no other warnings without this PR with more or less recent versions of meson.

> This is with 0.56.2 on Bullseye:

It will fail later though with
```
usage: meson [-h] {setup,configure,dist,install,introspect,init,test,wrap,subprojects,help,rewrite,compile} ...
meson: error: unrecognized arguments: --tags sd-boot
```

bluca:
> > there are already warnings printed for other unrelated reasons
> 
> There are no other warnings without this PR with more or less recent versions of meson.

I see some in bullseye, and it's fine

> > This is with 0.56.2 on Bullseye:
> 
> It will fail later though with
> 
> ```
> usage: meson [-h] {setup,configure,dist,install,introspect,init,test,wrap,subprojects,help,rewrite,compile} ...
> meson: error: unrecognized arguments: --tags sd-boot
> ```

Of course, that's a new command line argument - just don't use it. It's not mandatory.

medhefgo:
Imho, it would make sense to add a sd-boot target so that you can do `ninja -C build systemd-boot` to build the binares. The paths  aren't exactly nice to use and even contain the EFI arch that is being build.

bluca:
> Imho, it would make sense to add a sd-boot target so that you can do `ninja -C build systemd-boot` to build the binares. The paths aren't exactly nice to use and even contain the EFI arch that is being build.

Sounds like a good idea - added

medhefgo:
I feel like spelling the tag/target out fully (`systemd-boot/systemd-stub`) feels more consistent…

bluca:
> I feel like spelling the tag/target out fully (`systemd-boot/systemd-stub`) feels more consistent…

ok, renamed

keszybz:
OK, let's do this.

The warnings are annoying, but in practice they don't see *too* bad: the configuration step prints a lot of output anyway. And the warnings are not printed later during build or install phases. During configuration, on some systems we have warnings anyway (e.g. about nobody not matching), and one gets attenuated quickly.

evverx:
> in practice they don't see too bad

My understanding is that this PR was supposed to address only one specific feature request and to keep `meson-0.56` afloat at the same time so only two tags are used. If all the targets were tagged properly (as they should) those warnings would be hard to miss because basically every place where `install_tag` would be used would produce those warnings.

> the configuration step prints a lot of output anyway

And exactly because of that `fatal-meson-warnings` is used to avoid looking at all that stuff when there are no issues there.

eli-schwartz:
```meson
if meson.version().version_compare('>=0.62.0')
    custom_target(..., install_tag: 'foo')
else
    custom_target(...)
endif
```

is the canonical way to avoid warnings.

Please note that sheer coincidence:



> It won't fail, there will be a warning printed and then it continues, which is fine as there are already warnings printed for other unrelated reasons. If `--fatal-meson-warnings` is used, it can just be removed, it's not a big deal.
> 
> This is with 0.56.2 on Bullseye:
> 
> ```
> ../src/boot/efi/meson.build:415: WARNING: Passed invalid keyword argument "install_tag".
> WARNING: This will become a hard error in the future.
> WARNING: Unknown keyword arguments in target systemd-bootx64.elf: install_tag
> ../src/boot/efi/meson.build:431: WARNING: Passed invalid keyword argument "install_tag".
> WARNING: This will become a hard error in the future.
> WARNING: Unknown keyword arguments in target systemd-bootx64.efi: install_tag
> ../src/boot/efi/meson.build:415: WARNING: Passed invalid keyword argument "install_tag".
> WARNING: This will become a hard error in the future.
> WARNING: Unknown keyword arguments in target linuxx64.elf.stub: install_tag
> ../src/boot/efi/meson.build:431: WARNING: Passed invalid keyword argument "install_tag".
> WARNING: This will become a hard error in the future.
> WARNING: Unknown keyword arguments in target linuxx64.efi.stub: install_tag
> ```

Meson 0.60.0 made this (passing an invalid kwarg to the custom_target function) a hard error.

It just so happened that that is *also* the version that added a new kwarg "install_tag". So all versions that don't support "install_tag", also warn-but-don't-error on unknown kwargs.

You cannot apply this logic generically. Adding new kwargs while still supporting old versions of Meson is not "supposed" to work, and next time it probably won't.

evverx:
> if meson.version().version_compare('>=0.62.0')
...
is the canonical way to avoid warnings.

@eli-schwartz thanks! Personally I agree it should be wrapped into `if` but it would make it look like a kludge (which it is) and as far as I understand looking at https://lists.freedesktop.org/archives/systemd-devel/2022-April/047797.html the idea was to integrate it "seamlessly". Another upside would be that to avoid duplicating more targets than necessary the "libudev" and "libsystemd" tags would be gone until meson could be bumped properly (I don't think anyone has ever asked for them). Then again I don't think it should have been merged in the first place.

evverx:
In the meantime I opened https://github.com/systemd/systemd/pull/23229. I don't like it but if that feature is important enough that duplication should be warranted I assume

bluca:
> ```meson
> if meson.version().version_compare('>=0.62.0')
>     custom_target(..., install_tag: 'foo')
> else
>     custom_target(...)
> endif
> ```
> 
> is the canonical way to avoid warnings.

@eli-schwartz thanks for the suggestion, but please consider adding a more dev-friendly way to avoid these. Maybe give --fatal-warnings a value, or add something like -Wno-foo as the compilers do. For command-line driven things like these, it should really just be ignored by default, as the missing command line switch is more than enough to make it clear it's not supported. Forcing every developer to litter the meson file with a million conditionals makes it look like an autoconf script... if we wanted autoconf, we'd use autoconf ;-)

evverx:
> Maybe give --fatal-warnings a value, or add something like -Wno-foo as the compilers do

-Wno-foo is usually used to turn off false positives and in this particular case it's not a false positive so I'm not sure why `meson` should add `-Wno-foo` to make it easier to hide kludges like that.

> For command-line driven things like these, it should really just be ignored by default, as the missing command line switch is more than enough to make it clear it's not supported

Assuming that the missing command line switch is `--tags` it's passed to `meson install` and it's usually a separate step. How would meson know that whatever it configures is installed without `--tags`?

eli-schwartz:
I'm not exactly sure how that is supposed to work. My understanding is that compilers tag every single error with the "type" of warning and then let you turn specific types of warnings on and off; meson doesn't have specific types of warnings, just warning warnings, and none of them are optional.

And really, this would just fit into the subcategory of "FeatureNew warnings" so you would have to turn off ALL FeatureNew warnings to turn this one off.

Which... makes sense, because this is a new feature you are using and this one single cherry-picked example of a FeatureNew that is:
- unusually, not an error in older versions of meson
- unusually, not a feature that changes the output product of a standard `ninja && ninja install`

Does not scale sensibly to general meson stuff.

I already pointed this out above, of course. Today I'm just repeating myself.

Anyway, what I'm getting out of this is that you want a new command-line option to be added to Meson which exists, as far as I can tell, solely to make `install_tag` and that alone, not trigger a fatal warning. At first glance, it seems to me that it cannot be used for ***any*** other warning, because ***any*** other warning is either something that causes actual, real problems, or a kwarg that, when ignored, causes older versions of Meson to generate invalid binaries or libraries and thus, indirectly, disaster.

I warmly encourage you to change my mind by writing up a proposal to implement this feature, which also explains who would use it and why it's useful (for more than one cherry-picked case). I'm open to being convinced -- because as I'm not actually interested in this feature and don't feel like implementing it myself, I also haven't devoted a lot of effort to evaluating the pros and cons of it, or wrapping my brain around a good way to structure it.

> Forcing every developer to litter the meson file with a million conditionals makes it look like an autoconf script... if we wanted autoconf, we'd use autoconf ;-)

Not really sure I understand this comparison. There are a bunch of reasons why autoconf isn't very elegant to read or write, chief among them being that it uses 2 different languages intermixed in one file, and neither of them have a strong type system.

There are a lot of cases where you'd need to do if/else in an autoconf script, and many of them still need if/else in meson too. Option handling, enabling parts of a build, checking for different dependencies or defines based on OS or kernel or libc version...

There are many things that meson makes more ergonomic. Generating a different build output depending on your version of meson itself doesn't seem like it should be one of those things. If you are doing different things on different versions of Meson, 99 times out of 100 you need fallback logic to reimplement whatever you wanted to do, and in all 100 cases  you should be able to just use the old way of doing things without warnings, and not suffer from any loss of functionality you always had in the old versions.

No one is forced to litter conditionals anywhere. :) Of course, you can if you want to (and it important enough to you that you have a new feature immediately).

medhefgo:
Quite frankly, meson's behavior wrt backwards/forwards compatibility is just broken. When you specify `= version` in `project` you're stuck at that exact version and a newer meson outright refuses to operate. But meson afaik also does not support installing multiple version side-by-side for this kind of thing.

If you use `>= version`, a typical programmer (like myself) expects the API of that exact version. That *should* mean that you do not get any warnings about future behavior changes, nor should such behavior changes even affect code targeting older versions.
In fact, it outright refuse to use new features at all unless they're behind a version check guard to really drive home the point.

Effectively, this means meson should really be supporting both, the old behavior (for example `check` defaulting to false for `run_command`) when targeting a version `< flag_day` and defaulting to true for version `>= flag_day`.
And *then* you also wanna add some upgrade feature/flag that helps the user identify any behavior changes and maybe even automatically migrate trivial cases (say a flag like `--update-project-version='>= new_version'`.

But as it stands right now, meson does the same crazy thing as python does: 'We promise we will regularly break stuff and expect you to keep up to date if you want to use a newer meson version. And if you want to use both an older and newer version expect to use ugly kludges or overly verbose function arguments. Oh, and btw, good luck trying to build a older project that hasn't seen much love by their maintainers.'

eli-schwartz:
> But meson afaik also does not support installing multiple version side-by-side for this kind of thing.

I see no reason at all why that wouldn't work just fine. Meson has extensive documentation on how to install it, which isn't limited to "put up with the limitations of pip".

> If you use `>= version`, a typical programmer (like myself) expects the API of that exact version. That should mean that you do not get any warnings about future behavior changes, nor should such behavior changes even affect code targeting older versions.

I'm completely confused. What are you suggesting? Future behavior changes such as "accept new kwargs as valid and do something with them" should only occur when bumping the minimum version? Wouldn't that mean... erroring out instead of being halfhearted about it and just printing warnings?

> In fact, it outright refuse to use new features at all unless they're behind a version check guard to really drive home the point.

It refuses to what now? Isn't the entire point of this discussion about the fact that meson agrees to use new features?

Despite you first saying you want it to error out, then complaining that that's a bad thing?

> But as it stands right now, meson does the same crazy thing as python does: 'We promise we will regularly break stuff and expect you to keep up to date if you want to use a newer meson version [...]

Can you *please* clarify what you're referring to here? Meson certainly doesn't *promise* it will "regularly" break stuff that was formerly considered valid and documented functionality, and while stuff might very well be broken anyway, Meson's promise is that it won't and that that is therefore considered a bug.

For my sanity, I beg of you -- do not tell me that you meant Meson "regularly breaks old projects" by printing deprecation warnings.

medhefgo:
> I see no reason at all why that wouldn't work just fine. Meson has extensive documentation on how to install it, which isn't limited to "put up with the limitations of pip".

Is that seriously your idea of backwards compatibility?
"Use different --prefix=" and install ten different versions of meson and pick the right one when building $project"

> Can you _please_ clarify what you're referring to here? Meson certainly doesn't _promise_ it will "regularly" break stuff that was formerly considered valid and documented functionality, and while stuff might very well be broken anyway, Meson's promise is that it won't and that that is therefore considered a bug.

Eh? `check` parameter defaulting to `true` in some future version? Pretty much anytime you've changed some param to default to something else at some point is pretty much the definition of this, no?
Recently, I couldn't configure systemd-249 while hunting a regression with meson `>= 0.60.0` because some type has lost it's `+` operator.

Let's rephrase what meson *should* do for `= version`:
 - Allow building with a *newer* version of meson than requested by `version`.
 - *Refuse* any use of features that were not part of `version`.
 - *Not warn* about any (incompatible) future changes.
 - Any behavior changes that are part of `> version` should *not* take effect (e.g. check defaulting to false when requesting an older version).
 - Provide a command line option to detect code that needs attention when devs decide they want to port to `new_version`.

Additionally, `>= version` ought to do basically the same, except allow use of new features behind feature flags or explicit version checks. Inside of version check `if` blocks, the code shall have access to new features and also use new default behaviors.

Can't be that hard to understand.

The general issue is that right now `= version` is *useless*, so everyone uses `>= version`. But that one is fraught with compatibility warnings nobody sees or reads and things get even funnier when multiple versions of meson are used to build something and some of them disagree on what is the proper default. The former can be dealt with by using fatal-warnings, so that we get some hope of stability from CI, but it's not exactly pretty.

eli-schwartz:
> Is that seriously your idea of backwards compatibility?
"Use different --prefix=" and install ten different versions of meson and pick the right one when building $project"

Nice strawman you have there. You're the one who said you *wanted* multiple versions installed at the same time. I merely corrected you when you said it's *not possible*.

> Eh? `check` parameter defaulting to true in some future version? Pretty much anytime you've changed some param to default to something else at some point is pretty much the definition of this, no?

Thank you for clarifying. Do you have an example of where this has ever actually occurred, or are you just blindly assuming that because Meson warns about `check`, that it does so "regularly" to use your scare tactics?

Because the warning by `check` is specifically agnostic about the timeline as a result of the fact that we don't want to break backwards compat and have no plans for when to carry through on that. This is no different from any functionality marked as deprecated, with the exception that `check` implicit false is not "deprecated" but rather "dangerously unexpected by most people" so Meson doesn't restrict itself to warning you based on your minimum Meson version.

To date, the only realistic suggestion for when to drop  deprecated features and change defaults such as `check`, is to do so whenever Meson decides to finally release a 1.0 semver-incompatible major version, and this too has no real timeline.

> Recently, I couldn't configure systemd-249 while hunting a regression with meson `>= 0.60.0` because some type has lost it's `+` operator.

Since you're talking about Meson "promising" to "regularly" break stuff, are you saying that meson documented "haha lol we're intentionally removing this operator"? Or are you complaining that meson has bugs?

In the latter case, has that been reported as a bug?

> Let's rephrase what meson should do for `= version`:

This is an opinion, and not one I agree with. Just like `dependency('foo', version: '=1.0')` doesn't allow you to use any other version "but refuse to use features of libfoo.so.1.0 that come from libfoo.so.2.0", so too the same version algorithm is used to do version checks for meson.

Since all these things you list except that last one are features of `>= version`, I see no reason to break everything and introduce it to `=`.

The only reason `=` works in meson_version is because the version comparison algorithm is shared between that, find_program, dependency, and str.version_compare; I don't recommend its use.

> Additionally, `>= version` ought to do basically the same, except allow use of new features behind feature flags or explicit version checks. Inside of version check `if` blocks, the code shall have access to new features and also use new default behaviors.

This is already essentially the case, code inside a version checked if block isn't interpreted on older versions of Meson because the if block is false and therefore is not entered.

We're not interested in turning meson into cmake and having "policies" by whatever name constituting a language fork of behaviors.

> The general issue is that right now = version is useless, so everyone uses `>= version`.

No, the general issue is that it only exists as a side effect of the universal version_compare functionality. You can do `<= version` too if you really want.

I would be amenable to adding validation to this that makes its use an error ("only `>` and `>=` are supported") but I am afraid to do so because *it might break existing projects in the wild* oh no we can't have that, compatibility at all costs, how disappointing.

Maybe I can write a PR to make it a warning and threaten to remove support for it "at some point in the future" but never actually do so because compatibility...

> But that one is fraught with compatibility warnings nobody sees or reads and things get even funnier when multiple versions of meson are used to build something and some of them disagree on what is the proper default.

Of course people see or read those compatibility warnings. That's why we added them, and rather a lot of people have thanked us for it because they read them and follow the advice from it.

Meson even helpfully provides a summary at the end for all compatibility warnings, so you don't even have to scroll up to see if you're using too-new features.

It's really easy to see and read, because it's literally there at the bottom where you can't miss it.

eli-schwartz:
> > Recently, I couldn't configure systemd-249 while hunting a regression with meson `>= 0.60.0` because some type has lost it's `+` operator.
> 
> Since you're talking about Meson "promising" to "regularly" break stuff, are you saying that meson documented "haha lol we're intentionally removing this operator"? Or are you complaining that meson has bugs?
> 
> In the latter case, has that been reported as a bug?

:/ https://github.com/mesonbuild/meson/issues/10336

I didn't realize you were being facetious. While a number of other types document that they have a plus operator, bool does not. Furthermore, bool *cannot*, because the documentation explicitly states that that behavior doesn't work and isn't allowed since it violates one of the most basic and fundamental rules of meson syntax.

Mea culpa. I should have clarified that there's a third case, which is neither "meson dropped support for something" nor "a bug", but rather "a bug was fixed that caused invalid meson.build files to error out rather than silently doing the wrong thing".

In this third case, we will typically evaluate the fallout, particularly with regard to how widespread use of the invalid logic is, and either leave things as they are or add a compatibility shims along with a mandatory deprecation warning (e.g. mentioning the word "invalid") to make sure people stop using the invalid construct ASAP.

Meson compatibility guarantees don't apply to cases where the documentation explicitly says something does not work, even if we offer a grace period.

medhefgo:
> I didn't realize you were being facetious.

Reporting bugs (when asked to) is facetious now?! WTF?!! 

Mind you, I do not much care for the reasons about why that was changed and whether it was right or wrong. I just reported it because I got hit by it when regression testing and you were asking if it was reported, that's all. Feel free to close it.

evverx:
https://github.com/mesonbuild/meson/issues/10336 is kind of my fault in the end. Sorry about that. If I've learned anything it's that meson features that work by accident (especially if they produce warnings) can (and probably will) go sideways in mysterious ways (a few years later probably).

evverx:
@eli-schwartz as far as I understand https://github.com/mesonbuild/meson/issues/10336 essentially boils down to how to take the latest version of meson and use it to build a certain version of systemd (during `git bisect` for example). It isn't always possible for various reasons (in that particular case the behavior I relied on was fixed and `meson>=60` can no longer build it). As far as I can tell https://github.com/systemd/systemd/pull/23204#issuecomment-1114214727 describes a way to make it work without downgrading `meson`. I don't know if `meson` could/should support something like that though.

Regarding breaking changes as far as I can remember there were issues in weird environments like OSS-Fuzz where everything (including linker flags) is passed via CFLAGS. They were all reported and fixed and covered by the tests. To make it easier to catch regression `meson` was pinned and now it's updated via PRs like https://github.com/systemd/systemd/pull/23240. Updates like that rely on the fact that `--meson-fatal-warnings` can be passed there and unfortunately with this PR merged that workflow is kind of broken in general and if new targets are added all those warning will most likely be ignored eventually and turned off altogether (as was done in this PR to make CI "happy").

There are things that don't work: https://github.com/systemd/systemd/pull/23158 but I have to admit I haven't figured out how it should work. I came up with some kludges and wait for that part of OSS-Fuzz to be stabilized so I haven't reported anything yet.

Anyway, I hope this PR is either reverted or `install_tag` is wrapped into if.

PR #23205, opened 2022-04-27
tmpfiles: Split networkd entries into a separate file

daandemeyer:
Many distributions ship systemd-networkd as a separate file so we
need to be able to ship the tmpfiles networkd entries as part of
that separate networkd package. Let's split the networkd entries
into a separate file to make that possible.

### DIFF
--- a/tmpfiles.d/meson.build
+++ b/tmpfiles.d/meson.build
@@ -5,15 +5,15 @@ enable_tmpfiles = conf.get('ENABLE_TMPFILES') == 1
 files = [['README',               ''],
          ['home.conf',            ''],
          ['journal-nocow.conf',   ''],
+         ['portables.conf',       'ENABLE_PORTABLED'],
+         ['systemd-network.conf', 'ENABLE_NETWORKD'],
          ['systemd-nologin.conf', 'HAVE_PAM'],
          ['systemd-nspawn.conf',  'ENABLE_MACHINED'],
+         ['systemd-pstore.conf',  'ENABLE_PSTORE'],
          ['systemd-resolve.conf', 'ENABLE_RESOLVE'],
          ['systemd-tmp.conf',     ''],
-         ['portables.conf',       'ENABLE_PORTABLED'],
-         ['systemd-pstore.conf',  'ENABLE_PSTORE'],
          ['tmp.conf',             ''],
          ['x11.conf',             ''],
-         ['systemd-network.conf', 'ENABLE_NETWORKD'],
         ]
 
 foreach pair : files


bluca:
Didn't we have some issues in the past with splitting tmpfiles, due to users overriding them and suddenly the override not working anymore as the settings moved, or something like that? How was that solved?

tmpfiles.d/meson.build
```diff
@@ -2,17 +2,18 @@
...
+         ['journal-nocow.conf',    ''],
+         ['systemd-nologin.conf',  'HAVE_PAM'],
+         ['systemd-nspawn.conf',   'ENABLE_MACHINED'],
+         ['systemd-resolve.conf',  'ENABLE_RESOLVE'],
+         ['systemd-tmp.conf',      ''],
+         ['portables.conf',        'ENABLE_PORTABLED'],
+         ['systemd-pstore.conf',   'ENABLE_PSTORE'],
+         ['tmp.conf',              ''],
+         ['x11.conf',              ''],
+         ['systemd-networkd.conf', 'ENABLE_NETWORKD'],
```
yuwata: For resolved, we have systemd-resolve.conf. So, maybe systemd-network.conf is better??

Note, we have systemd-network.conf for sysusers.

yuwata (changes requested):

tmpfiles.d/meson.build
```diff
@@ -13,6 +13,7 @@ files = [['README',               ''],
          ['systemd-pstore.conf',  'ENABLE_PSTORE'],
          ['tmp.conf',             ''],
          ['x11.conf',             ''],
+         ['systemd-network.conf', 'ENABLE_NETWORKD'],
```
yuwata: That's not your fault, but could you sort the entries?

yuwata (approved):
LGTM. Only one minor thing.

I guess, 'backward incompatibility' should not a big problem here. As, it is only used by our daemons that uses sd-network, and it is not exported yet.

DaanDeMeyer:
Added a second commit that sorts the entries

poettering:
hmm, so i though this was proposed before, but there were some reservations. Anyone remembers? @keszybz you?

poettering:
can't find it in the issues/prs though

poettering:
ah, right, now i remember: these dirs are used for inotify watching by various of our tools (resolved for example), thus we can't really add them in when networkd is later installed because the stuff waiting for networkd notifications would then not start watching the stuff magically, because it's too late then.

the dirs and the system user for this should just be installed always, even if the networkd binary is dropped in later only.

At least as long as any component that is a client to networkd is installed.

DaanDeMeyer:
Should we create the dirs unconditionally then? The issue is that in centos, there's two different source rpms for the main systemd rpm and the systemd-networkd rpm which lives in EPEL. The main systemd rpm is compiled without networkd support and so these tmpfiles stanzas are not included in the main systemd tmpfiles.d/ snippet.

The systemd-networkd is (naturally) built with systemd-networkd enabled, but the main systemd snippet (that now has the networkd stuff inside of it) does not become part of the networkd package. The end result is that on a system with systemd and systemd-networkd installed, those networkd directories don't get created because they're not part of any tmpfiles.d/ snippet.

If we unconditionally create the systemd-networkd user and tmpfiles.d/ directory, that problem would be fixed.

yuwata:
I think this should be merged, and networkd rpm should restart resolved when it is installed.

poettering:
well, but why? having those dirs doesn't matter at all, no?

it's not just resolved, it's also timesyncd and wait online stuff, i.e. everything using sd_network_monitor...

i am not sure why bother with this. /run/systemd/ is systemd's private playground. if you enable networkd at all during build it should be totally ok to create some extra dirs there even though they aren't filled untill networkd is actually installed. after all the dirs are *used* already before that, i.e. that timesyncd/resolved/wait-online subscribe to it...

yuwata:
Hm, indeed. After re-considering, I think the directories are unconditionally created.

yuwata:
Wait, but the entries requires user systemd-network, but the user is added by sysusers.d/systemd-network.conf. And the file may be shipped with separated package for networkd...

DaanDeMeyer:
Yes exactly, that's the problem I'm trying to fix.

If I understand @poettering correctly, he suggests we additionally create these directories if resolved or timesyncd are enabled. That would then require the systemd-network.conf sysusers file to ship with the main systemd rpm instead of the networks rpm.

keszybz:
> these dirs are used for inotify watching by various of our tools (resolved for example), thus we can't really add them in when networkd is later installed because the stuff waiting for networkd notifications would then not start watching the stuff magically, because it's too late then.

Can we fix that? IIRC, in other places we watch the parent directory also. If we could allow the directory to be created later, this would all become simpler.

> I think this should be merged, and networkd rpm should restart resolved when it is installed.

Ah, that is another solution. resolved can be restarted at any time, so we can just a %trigger scriptlet.

Also, I see that the scriptlets do not restart networkd on package upgrades. Would it be safe to do so?

anitazha:
> Also, I see that the scriptlets do not restart networkd on package upgrades. Would it be safe to do so?

I was wondering about this. If you're using something like userspace RA wouldn't this cause any drops when restarting systemd-networkd?

yuwata:
> > Also, I see that the scriptlets do not restart networkd on package upgrades. Would it be safe to do so?
> 
> I was wondering about this. If you're using something like userspace RA would this cause any drops when restarting systemd-networkd?

Right. Restarting networkd causes that RA, DHCPv6, and IPv4LL addresses are dropped. Only DHCPv4 address is kept by default.

DaanDeMeyer:
So what's the conclusion here? Do we go the other way and create the dies and user if resolved or timesyncd are enabled?

yuwata:
I think the tmpfiles can be split by making sd-network watch parent directories if /run/systemd/network/links or friends do not exist.
IIRC, sd-bus does the similar thing.

DaanDeMeyer:
> I think the tmpfiles can be split by making sd-network watch parent directories if /run/systemd/network/links or friends do not exist.

I just checked and we already do this it seems? See `monitor_add_inotify_watch()` and `sd_network_monitor_flush()` which updates the monitor as the necessary directories are added.

yuwata:
Indeed. Then, the initial concern https://github.com/systemd/systemd/pull/23205#issuecomment-1110936055 is already solved.

DaanDeMeyer:
https://github.com/systemd/systemd/pull/23254 is slightly related

tmpfiles.d/meson.build
```diff
@@ -2,18 +2,18 @@
 
 enable_tmpfiles = conf.get('ENABLE_TMPFILES') == 1
 
-files = [['README',               ''],
-         ['home.conf',            ''],
+files = [['home.conf',            ''],
          ['journal-nocow.conf',   ''],
+         ['portables.conf',       'ENABLE_PORTABLED'],
+         ['README',               ''],
```
keszybz: Let's keep README at the top, unsorted.

DaanDeMeyer:
Added a NEWS entry to describe the change.

yuwata (approved):

PR #23209, opened 2022-04-27
RFC: cryptsetup: Add experimental EC keys support for PKCS#11 token

MkfsSion:
Currently, cryptsetup only support RSA algorirthm for PKCS#11 token via encryption and decrpytion of random LUKS key. As EC keys doesn't support encryption directly, the PR uses ECDH protocol to derive a shared secret and then use HKDF_SHA512 to derive the LUKS key(72 byte). I'm not an expert of cryptography, so maybe there is better way to support EC keys. Discussion is demanded and suggestions are highly welcomed.
Detailed process:
Enrollment:
First we generate a pair of EC key on enrollment and save public key, then we use ECDH protocol to derive a shared secret.
Then we use HKDF_SHA512 algorithm to derive the final LUKS key(72 byte)
Decryption:
We get the saved public key in `pkcs11-key` field and use ECDH protocol to derive shared secret on token. Then we use HKDF_SHA512 algorithm  to derive the final LUKS key.
TODO:
- [ ] Don't touch returned argument on failure
- [x] OpenSSL 3.0 compatibility
- [x] Update documentation
- [x] Coding style

### DIFF
--- a/man/crypttab.xml
+++ b/man/crypttab.xml
@@ -104,10 +104,14 @@
       see above and below.</para></listitem>
 
       <listitem><para>The key may be acquired via a PKCS#11 compatible hardware security token or
-      smartcard. In this case an encrypted key is stored on disk/removable media, acquired via
-      <constant>AF_UNIX</constant>, or stored in the LUKS2 JSON token metadata header. The encrypted key is
-      then decrypted by the PKCS#11 token with an RSA key stored on it, and then used to unlock the encrypted
-      volume. Use the <option>pkcs11-uri=</option> option described below to use this mechanism.</para></listitem>
+      smartcard. In this case a savedata used in unlock process is stored on disk/removable media, acquired via
+      <constant>AF_UNIX</constant>, or stored in the LUKS2 JSON token metadata header.</para>
+      <para>For RSA private key, the savedata is an encrypted key. The encrypted key is then decrypted by
+      the PKCS#11 token with an RSA key stored on it, and then used to unlock the encrypted volume.</para>
+      <para> For EC private key, the savedata is the generated public key in enrollment process. The public key
+      is then used to derive shared secret with the private key stored in the PKCS#11 token. Then the derived
+      shared secret is used to unlock the volume.</para><para>Use the <option>pkcs11-uri=</option> option
+      described below to use this mechanism.</para></listitem>
 
       <listitem><para>Similar, the key may be acquired via a FIDO2 compatible hardware security token (which
       must implement the "hmac-secret" extension). In this case a (during enrollment) randomly generated key
@@ -548,7 +552,7 @@
         <term><option>pkcs11-uri=</option></term>
 
         <listitem><para>Takes either the special value <literal>auto</literal> or an <ulink
-        url="https://tools.ietf.org/html/rfc7512">RFC7512 PKCS#11 URI</ulink> pointing to a private RSA key
+        url="https://tools.ietf.org/html/rfc7512">RFC7512 PKCS#11 URI</ulink> pointing to a private key
         which is used to decrypt the encrypted key specified in the third column of the line. This is useful
         for unlocking encrypted volumes through PKCS#11 compatible security tokens or smartcards. See below
         for an example how to set up this mechanism for unlocking a LUKS2 volume with a YubiKey security
@@ -562,11 +566,11 @@
         <literal>auto</literal>. In this mode the third column of the line should remain empty (that is,
         specified as <literal>-</literal>).</para>
 
-        <para>The specified URI can refer directly to a private RSA key stored on a token or alternatively
-        just to a slot or token, in which case a search for a suitable private RSA key will be performed.  In
-        this case if multiple suitable objects are found the token is refused. The encrypted key configured
-        in the third column of the line is passed as is (i.e. in binary form, unprocessed) to RSA
-        decryption. The resulting decrypted key is then Base64 encoded before it is used to unlock the LUKS
+        <para>The specified URI can refer directly to a private key stored on a token or alternatively
+        just to a slot or token, in which case a search for a suitable private key will be performed.  In
+        this case if multiple suitable objects are found the token is refused. The savedata configured
+        in the third column of the line is passed as is (i.e. in binary form, unprocessed) to unlock
+        process. The resulting decrypted key is then Base64 encoded before it is used to unlock the LUKS
         volume.</para>
 
         <para>Use <command>systemd-cryptenroll --pkcs11-token-uri=list</command> to list all suitable PKCS#11
@@ -792,7 +796,7 @@ external   /dev/sda3       keyfile:LABEL=keydev keyfile-timeout=10s,cipher=xchac
       <title>Yubikey-based PKCS#11 Volume Unlocking Example</title>
 
       <para>The PKCS#11 logic allows hooking up any compatible security token that is capable of storing RSA
-      decryption keys for unlocking an encrypted volume. Here's an example how to set up a Yubikey security
+      or EC decryption keys for unlocking an encrypted volume. Here's an example how to set up a Yubikey security
       token for this purpose on a LUKS2 volume, using <citerefentry
       project='debian'><refentrytitle>ykmap</refentrytitle><manvolnum>1</manvolnum></citerefentry> from the
       yubikey-manager project to initialize the token and
--- a/man/systemd-cryptenroll.xml
+++ b/man/systemd-cryptenroll.xml
@@ -34,7 +34,7 @@
     supports tokens and credentials of the following kind to be enrolled:</para>
 
     <orderedlist>
-      <listitem><para>PKCS#11 security tokens and smartcards that may carry an RSA key pair (e.g. various
+      <listitem><para>PKCS#11 security tokens and smartcards that may carry an RSA or EC key pair (e.g. various
       YubiKeys)</para></listitem>
 
       <listitem><para>FIDO2 security tokens that implement the <literal>hmac-secret</literal> extension (most
@@ -108,8 +108,13 @@
         be specified, in order to automatically determine the URI of a currently plugged in security token
         (of which there must be exactly one). The special value <literal>list</literal> may be used to
         enumerate all suitable PKCS#11 tokens currently plugged in. The security token must contain an RSA
-        key pair which is used to encrypt the randomly generated key that is used to unlock the LUKS2
-        volume. The encrypted key is then stored in the LUKS2 JSON token header area.</para>
+        or EC key pair which is used to generate LUKS2 volume key.</para><para>For RSA key, the private key
+        is ued to encrypt the randomly generated key that is used to unlock the LUKS2 volume. The encrypted key
+        is the savedata.</para><para>For EC key, we first generate a pair of EC key in same EC group. Then we
+        use the newly generated private key to derive shared secret with the public key in the PKCS#11 token.
+        Finally the shared secret is used to unlock the LUKS2 volume. The newly generated public key is the
+        savedata and the private key is simplely erased.</para><para>The savedata is then stored in the
+        LUKS2 JSON token header area.</para>
 
         <para>In order to unlock a LUKS2 volume with an enrolled PKCS#11 security token, specify the
         <option>pkcs11-uri=</option> option in the respective <filename>/etc/crypttab</filename> line:</para>
--- a/src/cryptenroll/cryptenroll-pkcs11.c
+++ b/src/cryptenroll/cryptenroll-pkcs11.c
@@ -6,7 +6,6 @@
 #include "memory-util.h"
 #include "openssl-util.h"
 #include "pkcs11-util.h"
-#include "random-util.h"
 
 int enroll_pkcs11(
                 struct crypt_device *cd,
@@ -18,11 +17,10 @@ int enroll_pkcs11(
         _cleanup_(erase_and_freep) char *base64_encoded = NULL;
         _cleanup_(json_variant_unrefp) JsonVariant *v = NULL;
         _cleanup_free_ char *keyslot_as_string = NULL;
-        size_t decrypted_key_size, encrypted_key_size;
-        _cleanup_free_ void *encrypted_key = NULL;
+        size_t decrypted_key_size, savedata_size;
+        _cleanup_free_ void *savedata = NULL;
         _cleanup_(X509_freep) X509 *cert = NULL;
         const char *node;
-        EVP_PKEY *pkey;
         int keyslot, r;
 
         assert_se(cd);
@@ -36,27 +34,9 @@ int enroll_pkcs11(
         if (r < 0)
                 return r;
 
-        pkey = X509_get0_pubkey(cert);
-        if (!pkey)
-                return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to extract public key from X.509 certificate.");
-
-        r = rsa_pkey_to_suitable_key_size(pkey, &decrypted_key_size);
+        r = X509_certificate_generate_volume_key(cert, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
         if (r < 0)
-                return log_error_errno(r, "Failed to determine RSA public key size.");
-
-        log_debug("Generating %zu bytes random key.", decrypted_key_size);
-
-        decrypted_key = malloc(decrypted_key_size);
-        if (!decrypted_key)
-                return log_oom();
-
-        r = crypto_random_bytes(decrypted_key, decrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to generate random key: %m");
-
-        r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to encrypt key: %m");
+                return log_error_errno(r, "Failed to generate secret key for X.509 certifcate: %m");
 
         /* Let's base64 encode the key to use, for compat with homed (and it's easier to type it in by
          * keyboard, if that might ever end up being necessary.) */
@@ -86,7 +66,7 @@ int enroll_pkcs11(
                                        JSON_BUILD_PAIR("type", JSON_BUILD_CONST_STRING("systemd-pkcs11")),
                                        JSON_BUILD_PAIR("keyslots", JSON_BUILD_ARRAY(JSON_BUILD_STRING(keyslot_as_string))),
                                        JSON_BUILD_PAIR("pkcs11-uri", JSON_BUILD_STRING(uri)),
-                                       JSON_BUILD_PAIR("pkcs11-key", JSON_BUILD_BASE64(encrypted_key, encrypted_key_size))));
+                                       JSON_BUILD_PAIR("pkcs11-key", JSON_BUILD_BASE64(savedata, savedata_size))));
         if (r < 0)
                 return log_error_errno(r, "Failed to prepare PKCS#11 JSON token object: %m");
 
--- a/src/home/homectl-pkcs11.c
+++ b/src/home/homectl-pkcs11.c
@@ -8,7 +8,6 @@
 #include "memory-util.h"
 #include "openssl-util.h"
 #include "pkcs11-util.h"
-#include "random-util.h"
 #include "strv.h"
 
 static int add_pkcs11_encrypted_key(
@@ -157,11 +156,10 @@ static int acquire_pkcs11_certificate(
 }
 
 int identity_add_pkcs11_key_data(JsonVariant **v, const char *uri) {
-        _cleanup_(erase_and_freep) void *decrypted_key = NULL, *encrypted_key = NULL;
+        _cleanup_(erase_and_freep) void *decrypted_key = NULL, *savedata = NULL;
         _cleanup_(erase_and_freep) char *pin = NULL;
-        size_t decrypted_key_size, encrypted_key_size;
+        size_t decrypted_key_size, savedata_size;
         _cleanup_(X509_freep) X509 *cert = NULL;
-        EVP_PKEY *pkey;
         int r;
 
         assert(v);
@@ -170,27 +168,9 @@ int identity_add_pkcs11_key_data(JsonVariant **v, const char *uri) {
         if (r < 0)
                 return r;
 
-        pkey = X509_get0_pubkey(cert);
-        if (!pkey)
-                return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to extract public key from X.509 certificate.");
-
-        r = rsa_pkey_to_suitable_key_size(pkey, &decrypted_key_size);
+        r = X509_certificate_generate_volume_key(cert, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
         if (r < 0)
-                return log_error_errno(r, "Failed to extract RSA key size from X509 certificate.");
-
-        log_debug("Generating %zu bytes random key.", decrypted_key_size);
-
-        decrypted_key = malloc(decrypted_key_size);
-        if (!decrypted_key)
-                return log_oom();
-
-        r = crypto_random_bytes(decrypted_key, decrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to generate random key: %m");
-
-        r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to encrypt key: %m");
+                return log_error_errno(r, "Failed to generate secret key for X.509 certifcate: %m");
 
         /* Add the token URI to the public part of the record. */
         r = add_pkcs11_token_uri(v, uri);
@@ -201,7 +181,7 @@ int identity_add_pkcs11_key_data(JsonVariant **v, const char *uri) {
         r = add_pkcs11_encrypted_key(
                         v,
                         uri,
-                        encrypted_key, encrypted_key_size,
+                        savedata, savedata_size,
                         decrypted_key, decrypted_key_size);
         if (r < 0)
                 return r;
--- a/src/shared/openssl-util.c
+++ b/src/shared/openssl-util.c
@@ -3,6 +3,7 @@
 #include "openssl-util.h"
 #include "alloc-util.h"
 #include "hexdecoct.h"
+#include "random-util.h"
 
 #if HAVE_OPENSSL
 int openssl_hash(const EVP_MD *alg,
@@ -42,7 +43,7 @@ int openssl_hash(const EVP_MD *alg,
         return 0;
 }
 
-int rsa_encrypt_bytes(
+static int rsa_encrypt_bytes(
                 EVP_PKEY *pkey,
                 const void *decrypted_key,
                 size_t decrypted_key_size,
@@ -79,15 +80,15 @@ int rsa_encrypt_bytes(
         return 0;
 }
 
-int rsa_pkey_to_suitable_key_size(
+static int rsa_pkey_to_suitable_key_size(
                 EVP_PKEY *pkey,
                 size_t *ret_suitable_key_size) {
 
         size_t suitable_key_size;
         int bits;
 
-        assert_se(pkey);
-        assert_se(ret_suitable_key_size);
+        assert(pkey);
+        assert(ret_suitable_key_size);
 
         /* Analyzes the specified public key and that it is RSA. If so, will return a suitable size for a
          * disk encryption key to encrypt with RSA for use in PKCS#11 security token schemes. */
@@ -106,6 +107,91 @@ int rsa_pkey_to_suitable_key_size(
                 return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Uh, RSA key size too short?");
 
         *ret_suitable_key_size = suitable_key_size;
+        return 0;
+}
+
+static int pkey_generate_ec_key(int nid, EVP_PKEY **ret_ppkey) {
+
+        assert(ret_ppkey);
+        assert_se(nid != NID_undef);
+
+        int r = 0;
+        _cleanup_(EVP_PKEY_CTX_freep) EVP_PKEY_CTX *pctx = NULL;
+        _cleanup_(EVP_PKEY_CTX_freep) EVP_PKEY_CTX *kctx = NULL;
+        _cleanup_(EVP_PKEY_freep) EVP_PKEY *params = NULL;
+
+        pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
+        if (pctx == NULL)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to allocate pkey context");
+
+        r = EVP_PKEY_paramgen_init(pctx);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to initialze pkey parameters context");
+
+        r = EVP_PKEY_CTX_set_ec_paramgen_curve_nid(pctx, nid);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to set pkey context ec curve nid");
+
+        r = EVP_PKEY_paramgen(pctx, &params);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to generate pkey parameters");
+
+        kctx = EVP_PKEY_CTX_new(params, NULL);
+        if (!kctx)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to allocate pkey context");
+
+        r = EVP_PKEY_keygen_init(kctx);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to initialize pkey keygen context");
+
+        *ret_ppkey = NULL;
+
+        r = EVP_PKEY_keygen(kctx, ret_ppkey);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to generate pkey");
+
+        return 0;
+}
+
+static int pkey_ecdh_derive_shared_secret(
+                EVP_PKEY *pkey,
+                EVP_PKEY *peer_key,
+                uint8_t *ret_shared_secret,
+                size_t *ret_shared_secret_len) {
+
+        _cleanup_(EVP_PKEY_CTX_freep) EVP_PKEY_CTX *ctx = NULL;
+        int r = 0;
+        size_t secret_len = 0;
+
+        assert(pkey);
+        assert(peer_key);
+        assert(ret_shared_secret_len);
+
+        ctx = EVP_PKEY_CTX_new(pkey, NULL);
+        if (!ctx)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to allocate pkey context");
+
+        r = EVP_PKEY_derive_init(ctx);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to initialize pkey derive context");
+
+        r = EVP_PKEY_derive_set_peer(ctx, peer_key);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to set peer key for derivation");
+
+        r = EVP_PKEY_derive(ctx, NULL, &secret_len);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to get derived key size");
+
+        if (!ret_shared_secret) {
+                *ret_shared_secret_len = secret_len;
+                return 0;
+        }
+
+        r = EVP_PKEY_derive(ctx, ret_shared_secret, ret_shared_secret_len);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to derive shared secret");
+
         return 0;
 }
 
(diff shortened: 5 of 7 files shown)


MkfsSion:
If anyone wants to test the patch, make sure your EC private key has `CKA_DERIVE` usage set. You can check it using `pkcs11-tool --list-objects --login`.

meson.build
```diff
@@ -2004,7 +2004,26 @@ if conf.get('HAVE_LIBCRYPTSETUP_PLUGINS') == 1
...
+                                     libshared],
+                        dependencies : [libcryptsetup,
+                                        libp11kit,
+                                        libopenssl,
+                                        versiondep],
+                        link_depends : cryptsetup_token_sym,
+                        install_rpath : rootlibexecdir,
+                        install : true,
+                        install_dir : libcryptsetup_plugins_dir)
+        endif
```
poettering: this seems unnecessary, just add the openssl dep unconditionally. cryptenroll has it anyway, there's no point in conditionalizing it here (also, please don't duplicate this many lines, you can build up deps list with various expressions quite easily, just look around in our meson files, there are plenty examples)

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+
+                r = genuine_random_bytes(decrypted_key, decrypted_key_size, RANDOM_BLOCK);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to generate random key: %m");
+
+                r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to encrypt key: %m");
+        }
+        else if (type == EVP_PKEY_EC) {
```
poettering: coding style, i.e. pleas place `}` and `else` on same line

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                _cleanup_(EVP_PKEY_freep) EVP_PKEY *pkey_new = NULL;
+                _cleanup_(erase_and_freep) uint8_t *shared_secret = NULL;
+                size_t shared_secret_len = 0;
+                EC_KEY *ec_key = EVP_PKEY_get0_EC_KEY(pkey);
+                int nid;
+
+                if (!ec_key)
+                        return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "pkey doesn't have EC_KEY associated");
+
+                nid = EC_GROUP_get_curve_name(EC_KEY_get0_group(ec_key));
```
poettering: can this fail? if yes, please add error handling. If not, wrap in `assert_se()` to indicate that this is known to always return something

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                r = pkey_generate_ec_key(nid, &pkey_new);
+                if (r != 0)
```
poettering: in our codebase it is customary to return errors as negative errno values, and everything >= is a form of success. Thus error checking of our functios usually looks like `if (r < 0)…` everywhere. Given that pkey_generate_ec_key() is defined by this patch please stick to this scheme

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                        return log_error_errno(r, "Failed to generate ec key");
```
poettering: when we log about an error we receive from somewhere, we usually include it in the output, i.e. suffix the format string here with `…: %m"`

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+
+                r = pkey_ecdh_derive_shared_secret(pkey_new, pkey, NULL, &shared_secret_len);
+                if (r != 0 || shared_secret_len == 0)
+                        return log_error_errno(r, "Failed to determine derived shared secret size");
```
poettering: if r == 0 and shared_secret_len == 0 then you will pass a 0 as first parameter to log_error_errno(). This will result in an assertion actually, log_error_errno() expects an error to be passed, and 0 is not an error...

Can shared_secret_len == 0 actually happen? Reading the function it dosn't look like it. Any reason you explicitly check for this?

also, please suffix with `…: %m"` again.

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+
+                shared_secret = malloc(shared_secret_len);
+                if (!shared_secret)
+                        return log_oom();
+
+                r = pkey_ecdh_derive_shared_secret(pkey_new, pkey, shared_secret, &shared_secret_len);
+                if (r != 0)
+                        return log_error_errno(r, "Failed to derive shared secret");
```
poettering: here too, and everywhere else, please stick to normalized error handling, as mentioned, i.e. return < 0 for errors, and output them in the log messages with %m

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+
+                decrypted_key = malloc(72);
```
poettering: where does the 72 come from? given this is used multiple times and is quite arbitrary, please make this a macro.

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                decrypted_key_size = 72;
+                encrypted_key_size = i2d_PUBKEY(pkey_new, NULL);
+                if (encrypted_key_size == 0)
+                        return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to determine encoded public key size");
+
+                encrypted_key = malloc(encrypted_key_size);
+                if (!encrypted_key)
+                        return log_oom();
+
+                // i2d_PUBKEY function has a side effect that makes *pp point to end of the allocated buffer
```
poettering: we reserve // comments for FIXME style comments, and use /* */ for explanatory comments. this qualifies as the latter, hence please use /*…*/

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                if (r < 0)
+                        return log_error_errno(r, "Failed to generate random key: %m");
+
+                r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to encrypt key: %m");
+        }
+        else if (type == EVP_PKEY_EC) {
+                _cleanup_(EVP_PKEY_freep) EVP_PKEY *pkey_new = NULL;
+                _cleanup_(erase_and_freep) uint8_t *shared_secret = NULL;
```
poettering: why uint8_t* btw? why not void*? I mean, you don#t actually access the data yourself here, this it's mostly just an opaque blob we pass around. Hence should be void*

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                encrypted_key_size = i2d_PUBKEY(pkey_new, NULL);
+                if (encrypted_key_size == 0)
+                        return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to determine encoded public key size");
+
+                encrypted_key = malloc(encrypted_key_size);
+                if (!encrypted_key)
+                        return log_oom();
+
+                // i2d_PUBKEY function has a side effect that makes *pp point to end of the allocated buffer
+                uint8_t *buffer = encrypted_key;
```
poettering: i2d_PUBKEY() actually wants an `unsigned char*` pointer, no? why not define it as that then? makes all casts unnecessary

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                encrypted_key_size = i2d_PUBKEY(pkey_new, (uint8_t **)&buffer);
+                if (encrypted_key_size == 0)
+                        return  log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to get encoded public key");
+        }
+        else
```
poettering: as above, please place } and else on the same line, matching our usual coding style

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Unsuported public key type: %s", OBJ_nid2sn(type));
```
poettering: it might make sense to split each of the two if branches into their own helper function? this was previously already quite long, but now is even longer....

src/shared/pkcs11-util.c
```diff
@@ -508,22 +508,65 @@ int pkcs11_token_read_x509_certificate(
...
 #endif
 
+static int pkcs11_token_find_private_key_from_template(
+        CK_FUNCTION_LIST *m,
+        CK_SESSION_HANDLE session,
+        CK_ATTRIBUTE_PTR template,
+        CK_ULONG templateCount,
+        CK_OBJECT_HANDLE_PTR objects,
+        CK_ULONG objectCount,
+        CK_ULONG *ret_nobjects) {
```
poettering: as per our usual coding style, please indent fuction parameter lists to double of the usual indentation, i.e. 16 spaces

src/cryptsetup/cryptsetup-tokens/luks2-pkcs11.c
```diff
@@ -79,9 +80,45 @@ static int luks2_pkcs11_callback(
...
+
+                shared_secret = malloc(shared_secret_len);
+                if (!shared_secret)
+                        return log_oom();
+
+                r = pkcs11_token_ecdh_derive_shared_secret(m, session, object, pkey, shared_secret, &shared_secret_len);
+                if (r < 0)
+                        return r;
+
+                data->decrypted_key = malloc(72);
```
poettering: 72 again...

src/cryptsetup/cryptsetup-tokens/luks2-pkcs11.c
```diff
@@ -79,9 +80,45 @@ static int luks2_pkcs11_callback(
...
+                if (!data->decrypted_key)
+                        return log_oom();
```
poettering: logging is not right here.

in systemd we want that functions either log about all errors, or about none (in which case the errors are propagated so that the caller can do that). usually we want more library-like utility functions to be non-logging, and main-program-like to be logging, but the line is blurry. What matters that you figure out for each function whether it better should be logging or non-logging and then stick to it. (in this context logging at LOG_DEBUG leve is not considered "real" logging and is always ok). I this function here, you handle some errors without logging, but others you do log (i.e. log_oom() here is of the latter kind, but further up you propagate -EIO/-EINVAL without logging.)

Now that is the general rule, but of course there are always exceptions and sometimes things become messy. These cryptsetup plugins for example should probably log via crypt_log_error(), i.e. libcryptsetup's logging infrastructure...

Anyway, it#s a bit messy, but please don't make this worse than it already is, and thus please stick to crypt_log_error() based logging in the plugin, and follow the other rule everywhere else.

src/cryptsetup/cryptsetup-tokens/luks2-pkcs11.c
```diff
@@ -79,9 +80,45 @@ static int luks2_pkcs11_callback(
...
+
+                data->decrypted_key_size = 72;
+                r = hkdf_sha512(shared_secret, shared_secret_len, data->decrypted_key_size, data->decrypted_key);
+                if (r < 0)
+                        return r;
```
poettering: this if body could be a helper function I think

src/shared/openssl-util.c
```diff
@@ -109,6 +109,126 @@ int rsa_pkey_to_suitable_key_size(
         return 0;
 }
 
+int hkdf_sha512(const uint8_t *in, const size_t in_len,  size_t out_len, uint8_t *ret_out) {
```
poettering: there's a double whitespace before size_t

src/shared/openssl-util.c
```diff
@@ -109,6 +109,126 @@ int rsa_pkey_to_suitable_key_size(
...
+
+        kctx = EVP_PKEY_CTX_new(params, NULL);
+        if (!kctx)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to allocate pkey context");
+
+        r = EVP_PKEY_keygen_init(kctx);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to initialize pkey keygen context");
+
+        *ret_ppkey = NULL;
```
poettering, line 147: we generally follow the rule that our functions should initialize all return parameters on success, and non on failure. Here you are initializing ret_ppkey before the last call that can still fail, thus there's a codepath where we'd touch the return param on failure.

thus please, store the result in a local variable first, and only once you know that nothing can fail anymore, copy to the real return param.

src/shared/openssl-util.c
```diff
@@ -109,6 +109,126 @@ int rsa_pkey_to_suitable_key_size(
...
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to set peer key for derivation");
+
+        r = EVP_PKEY_derive(ctx, NULL, &secret_len);
+        if (r != 1)
+                return log_debug_errno(SYNTHETIC_ERRNO(EIO), "Failed to get derived key size");
+
+        if (!ret_shared_secret) {
+                *ret_shared_secret_len = secret_len;
+                return 0;
```
poettering, line 188: as above

src/shared/pkcs11-util.c
```diff
@@ -508,22 +508,65 @@ int pkcs11_token_read_x509_certificate(
...
         _cleanup_free_ CK_ATTRIBUTE *attributes_buffer = NULL;
         CK_ULONG n_attributes, a, n_objects;
         CK_ATTRIBUTE *attributes = NULL;
         CK_OBJECT_HANDLE objects[2];
-        CK_RV rv, rv2;
+        CK_ATTRIBUTE key_type_template = {
+                .type = CKA_KEY_TYPE,
+        };
+        CK_RV rv;
+        const CK_BBOOL yes = true;
```
poettering: const without static on a non-pointer is weird. I figure this should be static as well

src/shared/pkcs11-util.c
```diff
@@ -556,90 +599,94 @@ int pkcs11_token_find_private_key(
...
                 return log_error_errno(SYNTHETIC_ERRNO(ENOENT),
-                                       "Failed to find selected private key suitable for decryption on token.");
+                                       "Failed to find selected private key suitable for decryption or derivation on token.");
         if (n_objects > 1)
                 return log_error_errno(SYNTHETIC_ERRNO(ENOTUNIQ),
-                                       "Configured private key URI matches multiple keys, refusing.");
+                                       "Configured private key URI matches multiple derivation keys, refusing.");
+
+        key_type_template.pValue = ret_key_type;
+        key_type_template.ulValueLen = sizeof(*ret_key_type);
```
poettering: maybe initialize the whole struct here:

```c
key_type_template = (CK_ATTRIBUTE) {
        .type = CKA_TYPE,
        .pValue = …,
        .ulValueLen = …,
};
```

src/shared/pkcs11-util.c
```diff
@@ -698,6 +745,127 @@ int pkcs11_token_decrypt_data(
...
+
+        assert(m);
+        assert(peer_key);
+        assert(ret_shared_secret_len);
+
+        _cleanup_(freep) uint8_t *buffer = NULL;
+        size_t shared_secret_len = 0;
+        size_t public_data_len = 0;
+        const CK_BBOOL yes = CK_TRUE;
+        const CK_BBOOL no = CK_FALSE;
```
poettering: as above, make these static. and combine on one line

src/shared/pkcs11-util.c
```diff
@@ -698,6 +745,127 @@ int pkcs11_token_decrypt_data(
...
+                uint8_t *ret_shared_secret,
+                size_t *ret_shared_secret_len) {
+
+        assert(m);
+        assert(peer_key);
+        assert(ret_shared_secret_len);
+
+        _cleanup_(freep) uint8_t *buffer = NULL;
+        size_t shared_secret_len = 0;
+        size_t public_data_len = 0;
```
poettering: combine these lines

src/shared/pkcs11-util.c
```diff
@@ -698,6 +745,127 @@ int pkcs11_token_decrypt_data(
...
+        const CK_OBJECT_CLASS shared_secret_class = CKO_SECRET_KEY;
+        const CK_KEY_TYPE shared_secret_type = CKK_GENERIC_SECRET;
+        CK_RV rv;
+        CK_ECDH1_DERIVE_PARAMS params = {
+                .kdf = CKD_NULL,
+                .pSharedData = NULL,
+                .ulSharedDataLen = 0,
+        };
+        CK_ATTRIBUTE shared_secret_template[] = {
+                {CKA_TOKEN, (void *)&no, sizeof(no)},
```
poettering: we usually place a space after { and before }

src/shared/pkcs11-util.c
```diff
@@ -698,6 +745,127 @@ int pkcs11_token_decrypt_data(
...
+                {CKA_ENCRYPT, (void *)&yes, sizeof(yes)},
+                {CKA_DECRYPT, (void *)&yes, sizeof(yes)}
+        };
+        CK_ATTRIBUTE shared_secret_attr = {
+                .type = CKA_VALUE,
+        };
+        CK_MECHANISM mechanism = {
+                .mechanism = CKM_ECDH1_DERIVE,
+                .pParameter = &params,
+                mechanism.ulParameterLen = sizeof(CK_ECDH1_DERIVE_PARAMS)
```
poettering: this doesnt look right: this is a line with side-effect on the struct itself. weird.

i figure you just want to say `.ulParameterLen` here...

src/shared/pkcs11-util.c
```diff
@@ -1232,9 +1401,45 @@ int pkcs11_crypt_device_callback(
...
+                data->decrypted_key = malloc(72);
+                if (!data->decrypted_key)
+                        return log_oom();
+
+                data->decrypted_key_size = 72;
+                r = hkdf_sha512(shared_secret, shared_secret_len, data->decrypted_key_size, data->decrypted_key);
+                if (r < 0)
+                        return r;
+        }
+        else {
```
poettering: coding style

src/shared/pkcs11-util.c
```diff
@@ -1232,9 +1401,45 @@ int pkcs11_crypt_device_callback(
...
+
+                shared_secret = malloc(shared_secret_len);
+                if (!shared_secret)
+                        return log_oom();
+
+                r = pkcs11_token_ecdh_derive_shared_secret(m, session, object, pkey, shared_secret, &shared_secret_len);
+                if (r < 0)
+                        return r;
+
+                data->decrypted_key = malloc(72);
```
poettering: still wondering where the 72 comes from

src/shared/pkcs11-util.c
```diff
@@ -1232,9 +1401,45 @@ int pkcs11_crypt_device_callback(
...
+                if (!data->decrypted_key)
+                        return log_oom();
+
+                data->decrypted_key_size = 72;
+                r = hkdf_sha512(shared_secret, shared_secret_len, data->decrypted_key_size, data->decrypted_key);
+                if (r < 0)
+                        return r;
+        }
```
poettering: this whole if block we already have seen somewhere else ust like this, no? share the code?

poettering (changes requested):
Mostly coding style issues.

Superficially this looks OK to me, but me too I am not a cryptographer, and I never played around with ECDH.

Is there any prior art of anyone using EC keys for unlocking stuff like this?

poettering:
> HKDF_SHA512 to derive the LUKS key(72 byte)

I am puzzled by this. 512 bits is 64 bytes, not 72? where do the extra 8 come from?

poettering:
Hmm can you explain exactly how your construction works? I mean, ECDH involves to peers with a key pairs. one side is stored on the token, but what about the other? And why would the resulting symmetric key be constant? don#t get it?

poettering:
hmm, why the hkdf? luks key management uses a kdf anyway (pbkdf or argon), so no need to do this manually here first? what am i missing?

MkfsSion:
> hmm, why the hkdf? luks key management uses a kdf anyway (pbkdf or argon), so no need to do this manually here first? what am i missing?

HKDF is used here since OpenSSL wiki page states we should never use shared secret directly. Since I'm not familier with cryptography, in my superficial opinion, use HKDF makes the key more random.

And maybe we should add needs-discussion label for the PR.

Reference:
1: https://wiki.openssl.org/index.php/Elliptic_Curve_Diffie_Hellman
2: https://crypto.stackexchange.com/questions/51070/is-the-openssl-implementation-of-ecdh-missing-something

poettering:
> HKDF is used here since OpenSSL wiki page states we should never use shared secret directly. Since I'm not familier with cryptography, in my superficial opinion, use HKDF makes the key more random.

well, but we won't. luks apply a kdf anyway, so i see no need here to do this here already once

MkfsSion:
> > HKDF is used here since OpenSSL wiki page states we should never use shared secret directly. Since I'm not familier with cryptography, in my superficial opinion, use HKDF makes the key more random.
> 
> well, but we won't. luks apply a kdf anyway, so i see no need here to do this here already once

HKDF is fast, and we can get a longer key (noticed that LUKS key length in RSA implmementation is much longer)  and key length is independent of curves. Anyway, it won't hurt.

poettering:
> > 
> 
> HKDF is fast, and we can get a longer key (noticed that LUKS key length in RSA implmementation is much longer) and key length is independent of curves. Anyway, it won't hurt.

dunno, i still would drop this. it's unnecssary complexity, luks applies a kdf anyway as mentioned (and it's not something you could even turn off). And even in the long run, should we ever hook this up with other usecases such as homed (i.e. for unlocking user accounts rather than luks volumes), there too a first-level kdf is not much use since the unix password hashing stuff used there already does it anyway too.

MkfsSion:
> 

Ok, I'll drop this.
@simo5 Thanks for your clarification.
I have an another question: Should we refactor our pkcs11 code to use libp11? Maybe we shouldn't reinvent the wheel.

simo5:
I have one more question, any reason why you did not implement ECIES: https://cryptobook.nakov.com/asymmetric-key-ciphers/ecies-public-key-encryption ?
You should have all the necessary primitives available and would use an actual standard which has been looked at by experts.

MkfsSion:
> I have one more question, any reason why you did not implement ECIES: https://cryptobook.nakov.com/asymmetric-key-ciphers/ecies-public-key-encryption ? You should have all the necessary primitives available and would use an actual standard which has been looked at by experts.

ECIES is more complicated, and we needn't do encryption and decryption in the case. I don't think ECIES can have much benifit in the case.

simo5:
> > I have one more question, any reason why you did not implement ECIES: https://cryptobook.nakov.com/asymmetric-key-ciphers/ecies-public-key-encryption ? You should have all the necessary primitives available and would use an actual standard which has been looked at by experts.
> 
> ECIES is more complicated, and we don't actually need do encryption and decryption in the case. I don't think ECIES can have much benifit in the case.

Thanks, I wasn't implying it has to be used, was just wondering if you had considered and decided against for some reason. Seem that is the case.

poettering:
BTW: if a pkcs#11 token contains both EC and RSA keys, we probably should prefer EC because faster (and more contemporary crypto), no?

Also needs some doc chnages, in man/systemd-cryptenroll.xml and man/crypttab.xml

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -40,24 +40,78 @@ int enroll_pkcs11(
...
+                encrypted_key = malloc(encrypted_key_size);
+                if (!encrypted_key)
+                        return log_oom();
+
+                // i2d_PUBKEY function has a side effect that makes *pp point to end of the allocated buffer
+                uint8_t *buffer = encrypted_key;
+                encrypted_key_size = i2d_PUBKEY(pkey_new, (uint8_t **)&buffer);
+                if (encrypted_key_size == 0)
+                        return  log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to get encoded public key");
+        }
```
poettering: btw, somewhere here, at the top of the enrollment code it would be good to have a coment explaining what the construction is, i.e. mention ECDH, mention which keys are stored on the PCKS#11 device, which key is stored in the unencrypted LUKS2 metadata header, and which key is simply dumped and forgotten. Also mention that the resulting symmetric key is not expanded here but later by the LUKS2 key management.

simo5:
> BTW: if a pkcs#11 token contains both EC and RSA keys, we probably should prefer EC because faster (and more contemporary crypto), no?

It's not so clear that we should. So far the understanding about Quantum Computers and Shor-derived algorithms is that the size of the key is the determining factor, and RSA has much longer keys than ECC. Of course it is also not clear if that will be significant or Quantum computers can quickly evolve to cover all the necessary coherent qbits between ECC and RSA within a short time frame.
I would say that the user should decide, as they may have their own uses and policies for keys.

> Also needs some doc chnages, in man/systemd-cryptenroll.xml and man/crypttab.xml

MkfsSion:
@poettering Should we make static helper functions not to touch returned arguments on failure? They will only be called by functions (like `X509_certificate_generate_volume_key()`) in same file scope. Make them not touch returned arguments would add duplicated code.

src/shared/openssl-util.c
MkfsSion: This is intended since we have to call the function with NULL arguments to determine buffer length

man/crypttab.xml
```diff
@@ -104,10 +104,14 @@
       see above and below.</para></listitem>
 
       <listitem><para>The key may be acquired via a PKCS#11 compatible hardware security token or
-      smartcard. In this case an encrypted key is stored on disk/removable media, acquired via
-      <constant>AF_UNIX</constant>, or stored in the LUKS2 JSON token metadata header. The encrypted key is
-      then decrypted by the PKCS#11 token with an RSA key stored on it, and then used to unlock the encrypted
-      volume. Use the <option>pkcs11-uri=</option> option described below to use this mechanism.</para></listitem>
+      smartcard. In this case a savedata used in unlock process is stored on disk/removable media, acquired via
```
poettering, line 107: "savedata"?

man/crypttab.xml
```diff
@@ -104,10 +104,14 @@
...
+      <constant>AF_UNIX</constant>, or stored in the LUKS2 JSON token metadata header.</para>
+      <para>For RSA private key, the savedata is an encrypted key. The encrypted key is then decrypted by
+      the PKCS#11 token with an RSA key stored on it, and then used to unlock the encrypted volume.</para>
+      <para> For EC private key, the savedata is the generated public key in enrollment process. The public key
+      is then used to derive shared secret with the private key stored in the PKCS#11 token. Then the derived
+      shared secret is used to unlock the volume.</para><para>Use the <option>pkcs11-uri=</option> option
```
poettering, line 113: you are inserting a docbook paragraph break here, but you hide this within the same line, please also insert an ascii paragraph break here

man/systemd-cryptenroll.xml
```diff
@@ -108,8 +108,13 @@
         be specified, in order to automatically determine the URI of a currently plugged in security token
         (of which there must be exactly one). The special value <literal>list</literal> may be used to
         enumerate all suitable PKCS#11 tokens currently plugged in. The security token must contain an RSA
-        key pair which is used to encrypt the randomly generated key that is used to unlock the LUKS2
-        volume. The encrypted key is then stored in the LUKS2 JSON token header area.</para>
+        or EC key pair which is used to generate LUKS2 volume key.</para><para>For RSA key, the private key
+        is ued to encrypt the randomly generated key that is used to unlock the LUKS2 volume. The encrypted key
```
poettering, line 112: "ued"

man/systemd-cryptenroll.xml
```diff
@@ -108,8 +108,13 @@
...
+        is the savedata.</para><para>For EC key, we first generate a pair of EC key in same EC group. Then we
+        use the newly generated private key to derive shared secret with the public key in the PKCS#11 token.
+        Finally the shared secret is used to unlock the LUKS2 volume. The newly generated public key is the
+        savedata and the private key is simplely erased.</para><para>The savedata is then stored in the
+        LUKS2 JSON token header area.</para>
```
poettering, line 118: "savedata" is not a word. why not call it the "saved key"?

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -18,11 +17,10 @@ int enroll_pkcs11(
         _cleanup_(erase_and_freep) char *base64_encoded = NULL;
         _cleanup_(json_variant_unrefp) JsonVariant *v = NULL;
         _cleanup_free_ char *keyslot_as_string = NULL;
-        size_t decrypted_key_size, encrypted_key_size;
-        _cleanup_free_ void *encrypted_key = NULL;
+        size_t decrypted_key_size, savedata_size;
+        _cleanup_free_ void *savedata = NULL;
```
poettering, line 21: "savedata" is weird. Call it `saved_key`?

src/home/homectl-pkcs11.c
```diff
@@ -170,27 +168,9 @@ int identity_add_pkcs11_key_data(JsonVariant **v, const char *uri) {
...
-
-        r = crypto_random_bytes(decrypted_key, decrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to generate random key: %m");
-
-        r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
+        r = X509_certificate_generate_volume_key(cert, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
         if (r < 0)
-                return log_error_errno(r, "Failed to encrypt key: %m");
+                return log_error_errno(r, "Failed to generate secret key for X.509 certifcate: %m");
```
poettering, line 173: "certifcate" is misspelled

src/home/homectl-pkcs11.c
```diff
@@ -170,27 +168,9 @@ int identity_add_pkcs11_key_data(JsonVariant **v, const char *uri) {
...
-        decrypted_key = malloc(decrypted_key_size);
-        if (!decrypted_key)
-                return log_oom();
-
-        r = crypto_random_bytes(decrypted_key, decrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to generate random key: %m");
-
-        r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
+        r = X509_certificate_generate_volume_key(cert, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
```
poettering, line 171: can you introduce this new helper call in a separate, initial commit? it seems like a good idea to do regardless of the followin EC stuff.

src/cryptenroll/cryptenroll-pkcs11.c
```diff
@@ -36,27 +34,9 @@ int enroll_pkcs11(
...
-
-        r = crypto_random_bytes(decrypted_key, decrypted_key_size);
-        if (r < 0)
-                return log_error_errno(r, "Failed to generate random key: %m");
-
-        r = rsa_encrypt_bytes(pkey, decrypted_key, decrypted_key_size, &encrypted_key, &encrypted_key_size);
+        r = X509_certificate_generate_volume_key(cert, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
         if (r < 0)
-                return log_error_errno(r, "Failed to encrypt key: %m");
+                return log_error_errno(r, "Failed to generate secret key for X.509 certifcate: %m");
```
poettering, line 39: certifcate is a typo.

so here you do duplicate logging. your new helper x509_certificate_generate_volume_key() already logs about all errors, hence you shouldn't do so here again.

In systemd we have this rule that all functions should be one of two kinds: either they log about all errors on their own, or they log about no errors, and propagate it instead, so that the caller can log. For each function determine which typo it should be, and then stick to this rule. Typically, "main program"-like functions are of the logging type and "library"-like functions about the non-logging type. (logging at LOG_DEBUG level is always OK, it does not count as logging in this context).

By following this rule we ensure that we neither log twice about errors nor not at all.

Here, X509_certificate_generate_volume_key() is of the logging kind, hence you should drop the exta logging here, and just do `return r`.

src/home/homectl-pkcs11.c
poettering: also, logging redudant, see above

src/shared/openssl-util.c
```diff
@@ -3,6 +3,7 @@
 #include "openssl-util.h"
 #include "alloc-util.h"
 #include "hexdecoct.h"
+#include "random-util.h"
```
poettering, line 6: hmm, not the fault of this PR, but we should order include lists alphabetically, can you fix this, if you touch this list?

src/shared/openssl-util.c
```diff
@@ -109,6 +110,91 @@ int rsa_pkey_to_suitable_key_size(
         return 0;
 }
 
+static int pkey_generate_ec_key(int nid, EVP_PKEY **ret_ppkey) {
+
+        assert(ret_ppkey);
+        assert_se(nid != NID_undef);
+
+        int r = 0;
```
poettering, line 118: needs no initialzation

src/shared/openssl-util.c
```diff
@@ -109,6 +110,91 @@ int rsa_pkey_to_suitable_key_size(
...
+        _cleanup_(EVP_PKEY_CTX_freep) EVP_PKEY_CTX *pctx = NULL;
+        _cleanup_(EVP_PKEY_CTX_freep) EVP_PKEY_CTX *kctx = NULL;
+        _cleanup_(EVP_PKEY_freep) EVP_PKEY *params = NULL;
+
+        pctx = EVP_PKEY_CTX_new_id(EVP_PKEY_EC, NULL);
+        if (pctx == NULL)
```
poettering, line 124: usually, when comparing a pointer to NULL we just write `if (ptcx)`, rely on C's pointer-to-bool downgrade concept.

src/shared/openssl-util.c
poettering: not adressed yet

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+
+/* We use derived shared secret as volume key for EC keys.
+ * The following is the procedure.
+ * First we generate a pair of EC key of the same curve group.
+ * Then we use newly generated private key to derive shared secret
+ * with the public key in the token.
+ * Finally, we simplely forget the private key, store the public key
+ * to pkcs11-key field of LUKS token and use the dervied shared secret
+ * as the volume key (we don't apply any KDF here since LUKS will apply
+ * PBKDF on the key).
```
poettering, line 235: can you line break this to 109ch width. That is what we usually use, and what all the vim/emacs/editorconfig/llvm-format metadata files in our tree also indicate. Please stick to that.

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+        int nid = NID_undef;
+        int r;
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000
+        _cleanup_free_ char *curve_name = NULL;
+        size_t len = 0;
+        if (EVP_PKEY_get_group_name(pkey, NULL, 0, &len) != 1 || len == 0)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to determine PKEY group name length");
+        len += 1;
+        curve_name = malloc(len);
```
poettering, line 254: we prefer typesafe allocations, i.e `malloc(len)` → `new(char, len)`

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+        nid = EC_GROUP_get_curve_name(EC_KEY_get0_group(ec_key));
+#endif
+
+        r = pkey_generate_ec_key(nid, &pkey_new);
+        if (r < 0)
+                return log_error_errno(r, "Failed to generate ec key: %m");
+
+        r = pkey_ecdh_derive_shared_secret(pkey_new, pkey, NULL, ret_decrypted_key_size);
+        if (r < 0 || *ret_decrypted_key_size == 0)
+                return log_error_errno(r, "Failed to determine derived shared secret size: %m");
```
poettering, line 278: if `pkey_ecdh_derive_shared_secret()` succeeds, but `*ret_decrypted_key_size` is zero, then you'll hit this error path, with r == 0, and propagate that as error, but zero is no error. This should be two error paths, one `r < 0` and one `*ret_decrypted_key_size == 0`

poettering: moreover, we have this rule in systemd that functions should not touch their return parameters on failure, and should initialize them fully on success. It's not OK to modify return values on failure. Here, `*ret_decrypted_key_size` might be initialized, but then later the function might still fail, conflicting with the rule.

Hence, please make pkey_ecdh_derive_shared_secret() write its result into a local temporary variable first, and then when nothing can fail anymore, copy it over to the actual return parameter.

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+
+        *ret_decrypted_key = malloc(*ret_decrypted_key_size);
+        if (!*ret_decrypted_key)
```
poettering, line 281: similar here

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+                return log_oom();
+
+        r = pkey_ecdh_derive_shared_secret(pkey_new, pkey, *ret_decrypted_key, ret_decrypted_key_size);
+        if (r < 0)
+                return log_error_errno(r, "Failed to derive shared secret: %m");
+
+        *ret_savedata_size = i2d_PUBKEY(pkey_new, NULL);
+        if (*ret_savedata_size == 0)
```
poettering, line 289: and here

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+                return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to determine encoded public key size");
+
+        *ret_savedata = malloc(*ret_savedata_size);
```
poettering, line 292: and here

poettering: (and everywhere else, not going to mention this anymore)

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+                size_t *ret_decrypted_key_size,
+                void **ret_savedata,
+                size_t *ret_savedata_size) {
+
+        assert(pkey);
+        assert(ret_decrypted_key);
+        assert(ret_savedata);
+        assert(ret_savedata_size);
+
+        int r = rsa_pkey_to_suitable_key_size(pkey, ret_decrypted_key_size);
```
poettering, line 318: so `int r` is our special, favourite variable to store return codes of functions in. Because of that it gets this special privilege, to always be defined uptop, and not "lazily" where it is needed. (doesn't matter though)

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+        assert_se(ret_decrypted_key);
+        assert_se(ret_decrypted_key_size);
+        assert_se(ret_savedata);
+        assert_se(ret_savedata_size);
+
+        _cleanup_free_ void *decrypted_key = NULL;
+        size_t decrypted_key_size = 0;
+        _cleanup_free_ void *savedata = NULL;
+        size_t savedata_size = 0;
+        int r = 0;
```
poettering, line 357: no need to initialize

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+        if (r < 0)
+                return log_error_errno(r, "Failed to generate random key: %m");
+
+        r = rsa_encrypt_bytes(pkey, *ret_decrypted_key, *ret_decrypted_key_size, ret_savedata, ret_savedata_size);
+        if (r < 0)
+                return log_error_errno(r, "Failed to encrypt key: %m");
+
+        return 0;
+
+}
```
poettering, line 338: spurious empty line

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+                void **ret_savedata,
+                size_t *ret_savedata_size) {
+
+        assert_se(cert);
+        assert_se(ret_decrypted_key);
+        assert_se(ret_decrypted_key_size);
+        assert_se(ret_savedata);
+        assert_se(ret_savedata_size);
+
+        _cleanup_free_ void *decrypted_key = NULL;
```
poettering, line 353: secrets should use `_cleanup_(erase_and_freep)` instead of `_cleanup_free_`, so that the contents is actively erased before freeing.

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+        size_t decrypted_key_size = 0;
+        _cleanup_free_ void *savedata = NULL;
+        size_t savedata_size = 0;
+        int r = 0;
+
+        EVP_PKEY *pkey = X509_get0_pubkey(cert);
+        if (!pkey)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to extract public key from X.509 certificate.");
+
+
```
poettering, line 363: spurious double empty line

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+        int type = EVP_PKEY_base_id(pkey);
+        if (type == EVP_PKEY_RSA)
+                r = pkey_generate_volume_key_rsa(pkey, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
+        else if (type == EVP_PKEY_EC)
+                r = pkey_generate_volume_key_ec(pkey, &decrypted_key, &decrypted_key_size, &savedata, &savedata_size);
+        else
+                return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Unsuported public key type: %s", OBJ_nid2sn(type));
```
poettering, line 370: Unsuported is a typo

src/shared/openssl-util.c
```diff
@@ -137,4 +223,159 @@ int string_hashsum(
...
+
```
poettering, line 371: we have this rule that function calls and their immediate error handling should be kept together. that rule would mean to drop this empty line here.

src/shared/pkcs11-util.c
```diff
@@ -556,96 +593,276 @@ int pkcs11_token_find_private_key(
...
+
+        /* Curves like secp521r1 will return 521 for EVP_PKEY_bits call,
+         * We need round it up to a size that can hold the generated shared secret */
+        shared_secret_len = (EVP_PKEY_bits(peer_key) + 7) / 8;
+
+        if (!ret_shared_secret)
+                *ret_shared_secret_len = shared_secret_len;
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000
+        if(EVP_PKEY_set_utf8_string_param(peer_key,
```
poettering, line 748: missing space between `if` and `(`

src/shared/pkcs11-util.c
```diff
@@ -556,96 +593,276 @@ int pkcs11_token_find_private_key(
...
+
+        public_data_len = EC_POINT_point2oct(EC_KEY_get0_group(ec_key),
+                                             EC_KEY_get0_public_key(ec_key),
+                                             POINT_CONVERSION_UNCOMPRESSED,
+                                             NULL, 0, NULL);
+        if (public_data_len == 0)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO), "Failed to determine public data size");
+
+        buffer = malloc(public_data_len);
+        if(!buffer)
```
poettering, line 775: missing space after `if`

src/shared/pkcs11-util.c
```diff
@@ -691,10 +908,76 @@ int pkcs11_token_decrypt_data(
...
+                r = pkcs11_token_decrypt_data_ec(m, session, object, encrypted_data, encrypted_data_size, &decrypted_data, &decrypted_data_size);
+#endif
+        else {
+                const char *key_type_str = pkcs11_key_type_to_string(key_type);
+                if (key_type_str != NULL)
+                        return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Unsupported private key type: %s", key_type_str);
+                else
+                        return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Unsupported private key type: %ld", key_type);
+        }
+
```
poettering, line 973: remove empty line, as above

src/shared/pkcs11-util.c
```diff
@@ -691,10 +908,76 @@ int pkcs11_token_decrypt_data(
...
+#if HAVE_OPENSSL
+        else if (key_type == CKK_EC)
+                r = pkcs11_token_decrypt_data_ec(m, session, object, encrypted_data, encrypted_data_size, &decrypted_data, &decrypted_data_size);
+#endif
+        else {
+                const char *key_type_str = pkcs11_key_type_to_string(key_type);
+                if (key_type_str != NULL)
+                        return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Unsupported private key type: %s", key_type_str);
+                else
+                        return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Unsupported private key type: %ld", key_type);
```
poettering, line 971: we don't really know what type `key_type` has, CK_KEY_TYPE could be anything, hence maybe cast it to (unsigned long) here?

poettering:
looks quite good, but lots of minor coding style issues.

poettering:
ping?

poettering:
looks so close to being ready! ping?

MkfsSion:
> looks so close to being ready! ping?

Sorry, I am busying in my class these days and my working machine is broken. I have to put it aside for a while.

H5117:
@MkfsSion If you don't mind, I will try to finish your patch.

MkfsSion:
> @MkfsSion If you don't mind, I will try to finish your patch.

Sorry, I can not continue to work on the PR for some reason. I would appreciate it if you could do it.

Issue #23210, opened 2022-04-28
Deploy initial time via timesyncd


darkdragon-001:
**systemd version the issue has been seen with**
> systemd 244 (244.5+)
> -PAM -AUDIT -SELINUX +IMA -APPARMOR -SMACK +SYSVINIT +UTMP -LIBCRYPTSETUP -GCRYPT -GNUTLS +ACL +XZ -LZ4 -SECCOMP +BLKID -ELFUTILS +KMOD -IDN2 -IDN -PCRE2 default-hierarchy=hybrid

**Used distribution**
> Yocto Dunfell

**Linux kernel version used** (`uname -a`)

 > Linux HOSTNAME 5.4.142+g05557cd55b18 #1 SMP PREEMPT Sat Sep 11 22:36:03 UTC 2021 armv7l armv7l armv7l GNU/Linux

**CPU architecture issue was seen on**
 > armv7l

**Expected behaviour you didn't see**
 > Setting modification timestamp of `/var/lib/systemd/timesync/clock` to the deployment time should set the time on first boot to this time when `systemd-timesyncd` is enabled (see [docs](https://www.freedesktop.org/software/systemd/man/systemd-timesyncd.service.html#/var/lib/systemd/timesync/clock))

**Unexpected behaviour you saw**
 > After first boot this time is reset to some older value (maybe _systemd build date_?)

**Steps to reproduce the problem**
1. Create file `/var/lib/systemd/timesync/clock` with desired `mtime` on freshly built root file system
2. Enable `systemd-timesyncd.service`
3. Configure non-existing NTP server to make sure time is not overwritten by NTP
4. Copy root file system to device
5. Boot device
6. See that time is not set to the desired one

**Additional program output to the terminal or log subsystem illustrating the issue**

Tagging you as you created the feature in #20142:
@dimitry-ishenko @poettering

darkdragon-001:
I was writing the wrong file. Works now.

Would still be interesting to know how the _systemd build date_ is determined...

poettering:
> Would still be interesting to know how the _systemd build date_ is determined...

SOURCE_DATE_EPOCH env var at build (as per reproducible build specs)

if not set and is built from a git tree, uses timestamp of last tag

if not set and not built from a git tree, uses timestamp of `NEWS` file.

[...]

PR #23214, opened 2022-04-28
Improve docs for timesyncd

darkdragon-001:
See https://github.com/systemd/systemd/issues/23210#issuecomment-1111942579

### DIFF
--- a/man/systemd-timesyncd.service.xml
+++ b/man/systemd-timesyncd.service.xml
@@ -78,8 +78,9 @@
           <para>The modification time ("mtime") of this file is updated on each successful NTP synchronization
           or after each <varname>SaveIntervalSec=</varname> time interval, as specified in
           <citerefentry><refentrytitle>timesyncd.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
-          At the minimum, it will be set to the systemd build date. It is used to ensure that the system clock
-          remains roughly monotonic across reboots, in case no local RTC is available.</para>
+          At the minimum, it will be set to the systemd build date (determined via modification time ("mtime")
+          of <filename>/usr/lib/clock-epoch</filename> or <filename>NEWS</filename>). It is used to ensure that
+          the system clock remains roughly monotonic across reboots, in case no local RTC is available.</para>
         </listitem>
       </varlistentry>
 


man/systemd-timesyncd.service.xml
```diff
@@ -78,7 +78,8 @@
           <para>The modification time ("mtime") of this file is updated on each successful NTP synchronization
           or after each <varname>SaveIntervalSec=</varname> time interval, as specified in
           <citerefentry><refentrytitle>timesyncd.conf</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
-          At the minimum, it will be set to the systemd build date. It is used to ensure that the system clock
+          At the minimum, it will be set to the systemd build date (specified as modification time ("mtime") of
+          <filename>/usr/lib/clock-epoch</filename>). It is used to ensure that the system clock
```
poettering: that's not really how this works. the build date is used when /usr/lib/clock-epoch doesn't exist. otherwise its mtime is used.

keszybz: @darkdragon-001 ?

darkdragon-001: I think I still don't understand how it works...

@poettering So the `mtime` of `/var/lib/systemd/timesync/clock` is used if it exists, otherwise the `mtime` of `/usr/lib/clock-epoch` is used if it exists and if both don't then the build time is used? If so where is the build time stored?

poettering: pid1 early on uses /usr/lib/clock-epoch's mtime if the file exists. if it does not, it uses the built-in time epoch (which is typically derived from the NEWS file timestamp ad build-time, but there are other options)

systemd-timesyncd then much later on might bump again via /var/lib/systemd/timesync/clock's mtime.

the stuff pid 1 does is a complete baseline concept, it is based on static timestamps that are never updated during runtime.

the stuff timesyncd is much more useful, since it ensures rough monotonicity of the clock. However it can do this only much later during boot, at the point where /var/ becomes accessible. Since this logic actually maintains a dynamic timestamp it's a much more powerful concept.

So, pid 1 clock bump is early and static, and timesyncd clock bump is late and dynamic, if you so will.

darkdragon-001: Thanks for the explanation! I gave it another try. Can you have a look if it is now accurate enough?
If not, could you please provide some draft?

poettering:
Sorry for dropping the ball on this. I have have prepped a patch similar to yours now in #28005, ptal.

i mentioned that this is based on your work. i hope it's ok i put this together in an independent patch though since i made this a bigger thing now. ptal.

closing this one

darkdragon-001:
> Sorry for dropping the ball on this. I have have prepped a patch similar to yours now in #28005, ptal.
> 
> i mentioned that this is based on your work. i hope it's ok i put this together in an independent patch though since i made this a bigger thing now. ptal.
> 
> closing this one

Awesome, thanks!

PR #17431, opened 2020-10-23
udev: make algorithm that selects highest priority devlink less susceptible to race conditions

msekletar:

Issue #11997, opened 2019-03-14
During switchroot, devices transition plugged -> dead -> plugged


owtaylor:
**systemd version the issue has been seen with**
v240 and v241

**Used distribution**
 Fedora

**Unexpected behaviour you saw**
 With systemd 240 and 241, at switch root time, all devices change plugged -> dead -> plugged. (I tested fedora 29 + v239, v240, and v241, and it was a v239 -> v240 regression.)

In some cases, this might be mostly harmless, but see https://pagure.io/fedora-silverblue/issue/8 for a problem that it causes.

**Steps to reproduce the problem**
* Install Fedora 30 nightly in a VM
* Boot with `systemd.log-level=debug printk.devkmsg=on log_buf_len=1M` on the kernel command line
 * Search for `plugged -> dead` in the logs

**Notes**

With some poking and debug message insertion, I determined that the initial call to device_enumerate() after re-exec is not finding any devices - which is what I think is causing the problem - I couldn't easily find the cause in the v239 ... v240 diff.

yuwata:
I think this is caused by 1fb70e66488d6c124d1d1e8d7cc2a98d2da5eb23.

yuwata:
I hope #12013 fixes this.

poettering:
I have the suspicion it would be a better fix to drop the initrd udev db cleanup service from the initrd. #otaylor, does that fix stuff for you?

poettering:
(i mean, a don't really see why we would flush the udev db like we currently do. We trigger everything anyway thus making sure the new rules are in effect and its just extra worl to flush the db explicitly and it means we drop props and tags for a time during boot which creates this issue here.)

poettering:
The only other option i see would be to freeze device unit states in pid1 while udev is not running and until all devices have been triggered and processed once. But that's always going to be awfully racy. And what happens with devices that actually really disappear during the transition? I mean, the user might actually pull a usb stick from the device right when we transition and that should definitely make the device unit in pid1 to go down

poettering:
hmm, another idea might be to set the "db_persist" option on all devices in the initrd we consider relevant for systemd. it's frickin ugly though, but would make sure that once a device is relevant to systemd it stays so across the transition

poettering:
So, yeah, I think the best approach for now would be to set "db_persist" on the select few devices where we really care and let's see if that makes things work.

owtaylor:
[...]

Issue #9751, opened 2018-07-30
race condition between "systemctl daemon-reload" and "udevadm trigger"


ssapalski:
**systemd version the issue has been seen with**
> v239

**Used distribution**
 > embedded system (but also reproduced with Debian and systemd v239)

**Expected behavior you didn't see**
 > serialized/deserialized device jobs don't hang forever after "daemon-reload"

**Unexpected behavior you saw**
 > serialized/deserialized device jobs hang forever after "daemon-reload"

**Steps to reproduce the problem**

To be executed in two different shells, a couple of iterations are enough. With proper timing this issue can be also reproduced with just one "udevadm trigger" & "systemctl daemon-reload" call in parallel if the latter is delayed a bit.
```
$ while true; do systemctl daemon-reload; done
$ while true; do udevadm trigger; done
```

----

If running device jobs get intercepted by "systemctl daemon-reload" (and in turn get serialized/deserialized), these will hang forever after "daemon-reload" returned. These device jobs hang in state "nop waiting" forever as can be seen afterwards with "systemctl list-jobs".

**I've bisected this behavior down to the following commit:**
```
f332611abee97247e40311d99492d53697152a1a is the first bad commit
commit f332611abee97247e40311d99492d53697152a1a
Author: Boucman <jeremy.rosen@enst-bretagne.fr>
Date: Fri Nov 10 17:00:32 2017 +0100

device : reload when udev generates a "changed" event (#6850)
```

This commit isn't the root cause of the problem but the one which caused to reproduce the issue easily with the above mentioned procedure. In general it seems that either all device jobs or "nop waiting" jobs which get serialized/deserialized hang forever after "daemon-reload" returns.

This issue can easily be reproduced, are further debug logs needed?

ssapalski:
@poettering 
I've updated the description since I've analyzed the issue a bit further.
Do you have an idea what happens here? Are further logs needed?

r0l1:
@poettering  Any updates on this issue?

PR #23218, opened 2022-04-28
pid1: several fixlets for device handling

### LINKED ISSUE #12953: Devices transition plugged -> dead -> plugged is still not completely fixed

### LINKED ISSUE #23208: Boot to emergency mode if devices are switched to SYSTEMD_READY=0 (multipath scenario)

yuwata:
Fixes #12953.
Fixes #23208.
Replaces #23215.

### DIFF
--- a/src/core/device.c
+++ b/src/core/device.c
@@ -163,14 +163,57 @@ static int device_coldplug(Unit *u) {
         assert(d->state == DEVICE_DEAD);
 
         /* First, let's put the deserialized state and found mask into effect, if we have it. */
-
-        if (d->deserialized_state < 0 ||
-            (d->deserialized_state == d->state &&
-             d->deserialized_found == d->found))
+        if (d->deserialized_state < 0)
                 return 0;
 
-        d->found = d->deserialized_found;
-        device_set_state(d, d->deserialized_state);
+        Manager *m = u->manager;
+        DeviceFound found = d->deserialized_found;
+        DeviceState state = d->deserialized_state;
+
+        /* On initial boot, switch-root, reload, reexecute, the following happen:
+         * 1. MANAGER_IS_RUNNING() == false
+         * 2. enumerate devices: manager_enumerate() -> device_enumerate()
+         *    Device.enumerated_found is set.
+         * 3. deserialize devices: manager_deserialize() -> device_deserialize()
+         *    Device.deserialize_state and Device.deserialized_found are set.
+         * 4. coldplug devices: manager_coldplug() -> device_coldplug()
+         *    deserialized properties are copied to the main properties.
+         * 5. MANAGER_IS_RUNNING() == true: manager_ready()
+         * 6. catchup devices: manager_catchup() -> device_catchup()
+         *    Device.enumerated_found is applied to Device.found, and state is updated based on that.
+         *
+         * Notes:
+         * - On initial boot, no udev database exists. Hence, no devices are enumerated in the step 2.
+         *   Also, there is no deserialized device. Device units are (a) generated based on dependencies of
+         *   other units, or (b) generated when uevents are received.
+         *
+         * - On switch-root, the udev databse may be cleared, except for devices with sticky bit, i.e.
+         *   OPTIONS="db_persist". Hence, almost no devices are enumerated in the step 2. However, in general,
+         *   we have several serialized devices. So, DEVICE_FOUND_UDEV bit in the deserialized_found must be
+         *   ignored, as udev rules in initramfs and the main system are often different. If the deserialized
+         *   state is DEVICE_PLUGGED, we need to downgrade it to DEVICE_TENTATIVE (or DEVICE_DEAD if nobody
+         *   sees the device). Unlike the other starting mode, Manager.honor_device_enumeration == false
+         *   (maybe, it is better to rename the flag) when device_coldplug() and device_catchup() are called.
+         *   Hence, let's conditionalize the operations by using the flag. After switch-root, systemd-udevd
+         *   will (re-)process all devices, and the Device.found and Device.state will be adjusted.
+         *
+         * - On reload or reexecute, we can trust enumerated_found, deserialized_found, and deserialized_state.
+         *   Of course, deserialized parameters may be outdated, but the unit state can be adjusted later by
+         *   device_catchup() or uevents. */
+
+        if (!m->honor_device_enumeration && !MANAGER_IS_USER(m)) {
+                found &= ~DEVICE_FOUND_UDEV; /* ignore DEVICE_FOUND_UDEV bit */
+                if (state == DEVICE_PLUGGED)
+                        state = DEVICE_TENTATIVE; /* downgrade state */
+                if (found == DEVICE_NOT_FOUND)
+                        state = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
+        }
+
+        if (d->found == found && d->state == state)
+                return 0;
+
+        d->found = found;
+        device_set_state(d, state);
         return 0;
 }
 
@@ -644,13 +687,9 @@ static void device_found_changed(Device *d, DeviceFound previous, DeviceFound no
 }
 
 static void device_update_found_one(Device *d, DeviceFound found, DeviceFound mask) {
-        Manager *m;
-
         assert(d);
 
-        m = UNIT(d)->manager;
-
-        if (MANAGER_IS_RUNNING(m) && (m->honor_device_enumeration || MANAGER_IS_USER(m))) {
+        if (MANAGER_IS_RUNNING(UNIT(d)->manager)) {
                 DeviceFound n, previous;
 
                 /* When we are already running, then apply the new mask right-away, and trigger state changes


yuwata:
cc @mwilck, @fbuihuu and @hreinecke.

yuwata:
I hope most commits, except for the last two, are trivial.

yuwata:
From https://github.com/yuwata/systemd/commit/c7fd2d5c00b62e6ca53aab052c5e8f6ec6b7c3f8#commitcomment-72476156
> Thanks for looking into this! Note that my comment in [systemd#23208](https://github.com/systemd/systemd/issues/23208) was a bit misleading perhaps. In the case I quoted, udev had indeed taken over the symlink. But that's not necessarily the case if you look at the udev code. A worker may or may not take over the symlink. In particular, systemd has no concept of udev's link priority.
> 
> IMO the only safe way for systemd to derive the current state of symlinks is to read the symlink from the filesystem, at least after each ADD/CHANGE/REMOVE uevent.

@mwilck Thank you for the comment. But, sorry, I cannot follow the comment. Could you elaborate more?

mwilck:
Can you explain how this relates to my #23215 ?

mwilck:
> The last commit is mostly equivalent to your PR, no?

Right ... I didn't realize this looking over the commit messages. Why didn't you just use my patch?

mwilck:
> Right ... I didn't realize this looking over the commit messages. Why didn't you just use my patch?

I see you moved the flag modification from `device_deserialize_item()` to `device_coldplug()`. I suppose that's ok, but I'd prefer fixing the flags as soon as possible, thereby avoiding any risk to use the "wrong" flags in case the ordering of function calls in this part of the code changes in the future. Could you explain why you think doing it in `device_coldplug()` is better?

fbuihuu:
> > The last commit is mostly equivalent to your PR, no?
> 
> Right ... I didn't realize this looking over the commit messages. Why didn't you just use my patch?

@yuwata that doesn't sound really fair actually. As you can guess, Martin spent a lot o time investigating, analyzing and fixing the issues (introduced roughly 4 years ago by https://github.com/systemd/systemd/commit/66f3fdbb073d8263b34bd65982a7e59d09086b23 that no one could fix since then). So at minimum it would be nice to give him credit of the detailed analyzis although I'm not sure to understand why you didn't provide feedback on his PR in the first place, that's what we usually do.

yuwata:
> The code in [link_update()](https://github.com/systemd/systemd/blob/2d14ec9b8a2a7232b1a0b9229c4de4423790b5b3/src/udev/udev-node.c#L449) tries to set the symlink to the device with the highest link priority. If priorities are equal, it prefers itself (the device for which it's executing).

Right.

> We know that this may result in races between different devices claiming the same symlink.

The race should be solved by #23043. If you'd like, please take a look.

> If one device has higher link priority than all others, it will win. Regardless of which device "wins", all possible candidates will have the symlink listed in their `DEVLINKS` property.

Right.

> Thus if systemd processes an uevent with a certain devlink set, it can't be sure that the device it's currently processing actually owns this symlink.

It CAN, by using `sd_device_new_from_devname()`.

> In the case of multipath, `SYSTEMD_READY="0"` will be set for all devices except the winning device mapper device, which also has the highest priority. The same holds for MD RAID. IIUC this **should** cause the symlink to be set correctly in systemd. There's nothing that would enforce this behavior (setting `SYSTEMD_READY="0"` for all but the highest prio candidate) in udev rules, though.

Right. I forgot that in the previous version. Now, `device_is_ready()`, which checks `SYSTEMD_READY=`, is called after `sd_device_new_from_devname()`.

> In order to be certain to have the "right" device mapped to any symlink pseudo-device, systemd would need to read the symlinks that udev creates. The udev db itself doesn't contain this information, unless there's exactly one device with higher link prio than all others.

Right. So, `sd_device_new_from_devname()` reads the symlink.

> Alternatively, one might argue that if several devices claim the same symlink with the same prio, it doesn't really matter which one wins, and except for very rare situations it wouldn't really hurt that systemd and udev are in inconsistent state (if it did hurt, we'd see a lot more errors in this area than we actually do). If we accept this argument, it'd be sufficient if systemd read the link priority and mapped symlinks to the device with the highest prio, without having to actually read the symlinks.

I am not sure. May be true, but may not.

mwilck:
> The race should be solved by https://github.com/systemd/systemd/pull/23043. If you'd like, please take a look.

Interesting that you did it this way now... the basic idea looks similar to my ancient PRs #8667 and #9551, which have later been superseded by the "lockless" solution #17431. Back then, I had scaling issues with a very large number of symlink contenders (1000 and more) for a symlink. This was mostly due to bad scaling in the kernel, solved by fd7732e033e3 ("fs/locks: create a tree of dependent requests.") in kernel v5.0.

> Right. So, sd_device_new_from_devname() reads the symlink.

Alright. Good to know. Your patch set here is pretty big, I still have to dig deeper into it. Next week, probably.

mwilck:
> > Thus if systemd processes an uevent with a certain devlink set, it can't be sure that the device it's currently processing actually owns this symlink.
> 
> It CAN, by using `sd_device_new_from_devname()`.

I see. But now it can happen that when systemd processes an event for device A and reads the devlinks, the links point to device B. Different devlinks can even point to different devices C, D, … Not saying this is a problem, we just need to be aware of the possibility.

yuwata:
Hm...
From https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux-sanitizers/11428/
```
Apr 30 06:52:17 H systemd-journald[286]: Failed to send WATCHDOG=1 notification message: Connection refused
Apr 30 06:53:57 H systemd-journald[286]: Failed to send WATCHDOG=1 notification message: Transport endpoint is not connected
Apr 30 06:55:17 H systemd-journald[286]: Failed to send WATCHDOG=1 notification message: Transport endpoint is not connected
Apr 30 06:56:57 H systemd-journald[286]: Failed to send WATCHDOG=1 notification message: Transport endpoint is not connected
Apr 30 06:58:17 H systemd-journald[286]: Failed to send WATCHDOG=1 notification message: Transport endpoint is not connected
```

src/core/device.c
```diff
@@ -158,19 +158,65 @@ static void device_set_state(Device *d, DeviceState state) {
...
+         * - On reload or reexecute, we can trust enumerated_found, deserialized_found, and deserialized_state.
+         *   Of course, deserialized parameters may be outdated, but the unit state can be adjusted later by
+         *   device_catchup() or uevents. */
+
+        if (!m->honor_device_enumeration && !MANAGER_IS_USER(m)) {
+                f &= ~DEVICE_FOUND_UDEV; /* ignore DEVICE_FOUND_UDEV bit */
+                if (s == DEVICE_PLUGGED)
+                        s = DEVICE_TENTATIVE; /* downgrade state */
+                if (f == DEVICE_NOT_FOUND)
+                        s = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
```
mwilck: I assume that `deserialized_found == DEVICE_NOT_FOUND` implies `deserialized_state == DEVICE_DEAD`. In that case the clause would be redundant. But if this implication does **not** hold for whatever reason, we'd better understand how that comes to pass before we reset the state to "dead", which is dangerous.

mwilck:
@yuwata, AFAICS commit https://github.com/systemd/systemd/pull/23215/commits/57910e581ecd85de524c21a8eac213dd9c54732c (my proposed fix for #12953) is not included in your set. Could you explain please?

src/core/device.c
```diff
@@ -639,13 +685,9 @@ static void device_found_changed(Device *d, DeviceFound previous, DeviceFound no
...
 
 static void device_update_found_one(Device *d, DeviceFound found, DeviceFound mask) {
-        Manager *m;
-
         assert(d);
 
-        m = UNIT(d)->manager;
-
-        if (MANAGER_IS_RUNNING(m) && (m->honor_device_enumeration || MANAGER_IS_USER(m))) {
+        if (MANAGER_IS_RUNNING(UNIT(d)->manager)) {
```
mwilck, line 692: This drops the condition `m->honor_device_enumeration`. If the udev db is not initialized, `d->enumerated_found` will be `DEVICE_NOT_FOUND`. Thus `manager_startup() → manager_ready() → unit_catchup() → device_catchup() → device_update_found_one()` will see `MANAGER_OK`, and will reset `d->found` to `d->enumerated_found == DEVICE_FOUND_NONE`, with possibly fatal consequences. What am I missing here?

mwilck (changes requested):

src/core/device.c
mwilck: IOW, I think the test for `m->honor_device_enumeration` must be kept here.

src/core/device.c
```diff
@@ -705,9 +698,111 @@ static bool device_is_ready(sd_device *dev) {
...
+                if (PATH_STARTSWITH_SET(devlink, "/dev/block/", "/dev/char/"))
+                        continue;
+
+                if (device_setup_devlink_unit_one(m, devlink, &assigned) <= 0)
+                        continue;
+
+                if (sd_device_get_syspath(assigned, &s) < 0)
+                        continue;
+
+                if (path_equal(s, syspath))
```
mwilck: Should this be `!path_equal()` ?

mwilck:
> @yuwata, AFAICS commit [57910e5](https://github.com/systemd/systemd/commit/57910e581ecd85de524c21a8eac213dd9c54732c) (my proposed fix for #12953) is not included in your set. Could you explain please?

See https://github.com/systemd/systemd/pull/23234#issuecomment-1114704517

src/core/device.c
```diff
@@ -705,9 +698,111 @@ static bool device_is_ready(sd_device *dev) {
...
+                if (!path_startswith(d->path, "/dev/"))
+                        continue;
+
+                if (device_setup_devlink_unit_one(m, d->path, &assigned) <= 0)
+                        continue;
+
+                if (sd_device_get_syspath(assigned, &s) < 0)
+                        continue;
+
+                if (path_equal(s, syspath))
```
mwilck: Again, `!path_equal()`?

src/core/device.c
yuwata: The condition holds if `deserialized_found == DEVICE_FOUND_UDEV`.
In that case, `f == DEVICE_NOT_FOUND` and the device was neither mounted nor used as a swap.
Hence, I think it is safe to go to `DEVICE_DEAD`.

In other words, entering the state `DEVICE_TENTATIVE` with `DEVICE_NOT_FOUND` sounds quite strange.

src/core/device.c
yuwata: No, I intend to exclude this to make `ready_devlinks` only contain links not assigned to the device currently processing.

The .device units which corresponds to the syspath that is currently processing will be handled by `device_update_found_by_sysfs()`. However, .device units that get involved in this function whose syspath is different from the syspath currently processed must be handled by `device_update_found_by_name()`. So, we need to create the list of the involved devices here.

src/core/device.c
yuwata: This is also intended.

src/core/device.c
mwilck: It would be strange to find this combination (`found == DEVICE_NOT_FOUND` and `state != DEVICE_DEAD`) in the serialized state, yes. Maybe we should emit a warning if we hit this case.

src/core/device.c
mwilck: ok, thanks for the clarification.

yuwata:
Reply for https://github.com/systemd/systemd/pull/23234#issuecomment-1114704517
> @yuwata, you're confusing me. Why do we have a separate PR for this now, which is also included in #23218?

Sorry, previously one of the commit in this PR had a memleak, and CI were unhappy, to investigate the reason, I split the PR. Now, the memleak is fixed. I will close #23234.

> > @mwilck I do not think the first commit of #23215 is necessary. Moreover, it makes devices removed during reload or reexec processes not handled correctly.
> 
> The commit is necessary for systemd reload operations that take place before the udev DB is fully initialized, e.g. during boot, as explained int the commit message ([57910e5](https://github.com/systemd/systemd/commit/57910e581ecd85de524c21a8eac213dd9c54732c)). In this case, the `DEVICE_FOUND_UDEV` bit would be clear in `d->enumerated_found`, and we'd reset `d->found` in `device_catchup()`, causing the `plugged→dead→plugged` transition (#11997, #12953). If systemd is reloading, the only reason to reset `DEVICE_FOUND_UDEV` would be the arrival of a REMOVE uevent. If this actually happens during a reload, the uevent will be handled after the reload, and the device state will be correct again, even if we haven't reset it in `device_catchup()`.
> 
> You're right that this does not hold for `systemctl daemon-reexec`, because `reexec` closes systemd's device monitor socket, creating a time window in which uevents might get lost. That's no problem for the switching-root case, where udev coldplug will fix device states. `systemctl daemon-reexec` without switching root might indeed miss uevents. The "risk" that my patch introduces is only that _a device is removed while systemd is re-executing_, and systemd misses the REMOVE uevent. This is quite unlikely, and easily fixed by running `systemctl restart systemd-udev-trigger.service`. IMO this risk is rather unimportant compared to the frequent issues with systemd _reload_ operations during boot leading to "plugged→dead→plugged" transitions.
> 
> The only way to "fix" the reexec problem would be to distinguish between "reload" and "reexec" in `device_catchup()`, and apply my patch only in the "reload" case. But even if we did that, `daemon-reexec` triggered during the boot procedure would be at risk to falsely consider devices "dead" if it calls `device_enumerate` before udev coldplugging has finished. We know from past discussions about `udev settle` that determining when udev is **done** processing devices is basically impossible. IMHO the best way to avoid this would be to generally block or postpone "reexecute" operations (except for switching root) until `default.target` (or at least `local-fs.target`) has been reached. But that goes far beyond the scope of my PR.

With the second commit in #23215, or the second commit in this PR, all device units will not have `DEVICE_FOUND_UDEV` bit in `d->found`, and the state is `DEVICE_TENTATIVE` (or DEVICE_DEAD with my commit) after `device_coldplug()`. Hence, the devices with the tentative state will never enter the dead state in `device_catchup()` (unless the device is not mounted or used for swapping anymore). 
If PID1 is reloaded or reexecuted soon after switching root, the `deserialized_found` does not have `DEVICE_FOUND_UDEV` bit and `deserialized_state == DEVICE_TENTATIVE`, and these are simply copied to the main parameters. Hence the tentative devices will never enter the dead state in `device_catchup()` called in the reload or reexecute process (again, unless the device is not mounted or used for swapping anymore).

yuwata:
@mrc0mmand I noticed that if PID1 has a memleak, then several tests in CentOS CI (Arch Linux + sanitizers) fail. But, unfortunately, the logs do not contain anything about why the tests failed. E.g. https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux-sanitizers/11428/
Is this expected? Is it possible to improve the report like e.g. when networkd has memleak?

mwilck:
> With the second commit in #23215, or the second commit in this PR, all device units will not have `DEVICE_FOUND_UDEV` bit in `d->found`, and the state is `DEVICE_TENTATIVE` (or DEVICE_DEAD with my commit) after `device_coldplug()`. 

Hm, you're right. I didn't realize this because I'd tackled the "plugged→dead→plugged" before I looked at the multipath issue.

mrc0mmand:
> @mrc0mmand I noticed that if PID1 has a memleak, then several tests in CentOS CI (Arch Linux + sanitizers) fail. But, unfortunately, the logs do not contain anything about why the tests failed. E.g. https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux-sanitizers/11428/ Is this expected? Is it possible to improve the report like e.g. when networkd has memleak?

I've stumbled across this as well when debugging stuff in RHEL. The issue here, imo, is that the memleak happens when a logging facility is not available (e.g. before/after journald is started/stopped), so the log goes to /dev/null. I guess it would make sense to use `log_path=` for ASan/UBSan logs for PID as well (as we already do for journald), so they don't get accidentally lost. I'll put this on my TODO list to look at it later.

src/core/device.c
mwilck: `m->honor_device_enumeration` matters only if switching root. Following a similar argument as https://github.com/systemd/systemd/pull/23218#issuecomment-1114901938, the device would have been reset to `DEVICE_TENTATIVE` beforehand. If the device had been mounted before switching root, `device_enumerate()` should have seen it and `enumerated_found` should be `DEVICE_FOUND_MOUNTED`.
So I *suppose* this change does no harm. It's the kind of thing that scares me a bit though, because of past experience with the subtleties of the device handling code.

As `m->honor_device_enumeration` now doesn't affect any more whether device enumeration is honored, this change is definitely an argument for changing the field name from `m->honor_device_enumeration` to `m->honor_serialized_device_state` or perhaps simply `m->switching_root`.

mwilck:
@yuwata, I would appreciate if we could isolate the switching root / device activation fixes from the code that modifies the symlink handling. The former is required to fix actual bugs we are seeing. The latter, while highly worthwhile, is not known to cause any real-world issues.

You are probably aware that the issues we're working on here are outside the realm of CI. Only testing on a variety of diversely configured real systems by real people reveals possible issues. Most of these users don't test with upstream systemd, and won't be easily persuaded to do so.

My patches from #23215 have been verified by users on SUSE releases of systemd, based on systemd 249 and 246, respectively. I'll provide backports of  b5f9f2b7 to these users for testing, but I will not include the rest of your changes because they are too intrusive for us. I hope you can understand that.

fbuihuu:
@yuwata a slightly different approach can be found at https://github.com/systemd/systemd/compare/main...fbuihuu:device-fix-plug-dead-plug-again, see the first three commits. It basically always forces to forget all devices that is no more present in udev DB (done by `device_coldplug()`) and let `device_catchup()` decide for the final state (in case the device is referenced by a mount/swap unit). This is more or less what the old code did. It also introduces `MANAGER_IS_SWITCHING_ROOT()` which makes the code slightly easier to parse.

The fourth commit is more interesting because it drops `device_coldplug()` entirely. Since this function was basically used to restore the serialized device properties, which is supposed to be done be `device_deserialize_items()`. Hence it let `device_catchup()` decides for the final state of the device and generate the state transition event when needed.

yuwata:
Sorry for late reply. I will update this and reply the comments later today.

yuwata:
Dropped the commits about symlink update. Will reply the comments one-by-one.

src/core/device.c
yuwata: > `m->honor_device_enumeration` matters only if switching root. Following a similar argument as [#23218 (comment)](https://github.com/systemd/systemd/pull/23218#issuecomment-1114901938), the device would have been reset to `DEVICE_TENTATIVE` beforehand. If the device had been mounted before switching root, `device_enumerate()` should have seen it and `enumerated_found` should be `DEVICE_FOUND_MOUNTED`. So I _suppose_ this change does no harm. It's the kind of thing that scares me a bit though, because of past experience with the subtleties of the device handling code.

`device_enumerate()` sets `DEVICE_FOUND_UDEV` bit if the device is already processed by systemd-udevd, instead of `DEVICE_FOUND_MOUNT`.
As I commented in the code, on switching root, `device_enumerate()` lists almost no devices, as udev database is cleared. Hence, `device_catchup()` does not set `DEVICE_FOUND_UDEV` flag for devices, and `device->state` is kept as `DEVICE_TENTATIVE`.

> As `m->honor_device_enumeration` now doesn't affect any more whether device enumeration is honored, this change is definitely an argument for changing the field name from `m->honor_device_enumeration` to `m->honor_serialized_device_state` or perhaps simply `m->switching_root`.

Yeah, I agree. But let's do that later.

yuwata:
> I have run the first two patches from this PR ([5cfb16a](https://github.com/systemd/systemd/commit/5cfb16af539ded421328e44ce0f51785778cda75) and [b5f9f2b](https://github.com/systemd/systemd/commit/b5f9f2b79731aaa5fc332bada88bdc52870f6a6a)) through extensive reboot tests, both on systems that would used to exhibit boot failures without these patches and on others (to watch out for regressions). Two customers, who could reproduce multipath booting issues on their systems easily before, have verified the same fix in similar extensive reboot tests. All tests were done with SUSE systemd packages with the mentioned two patches cherry-picked.
> 
> **I consider these two commits validated by real-world testing.**

Yay!! Thank you!

src/core/device.c
mwilck: My concern here was that `device_coldplug()` now calls `device_set_state()` even if `honor_device_enumeration` is false. But as I said previously, I believe it's ok, and my testing yielded favourable results. This conversation can be considered resolved.

>  As I commented in the code, on switching root, `device_enumerate()` lists almost no devices, as udev database is cleared. 

FTR: true,  but `manager_enumerate()` also calls `mount_enumerate()`, which may set `DEVICE_FOUND_MOUNT`.

yuwata:
I guess all concerns are solved now? At least I think all questions are answered.
PTAL!

src/core/device.c
yuwata: OK. Let's add the warning later.

src/core/device.c
```diff
@@ -158,19 +158,65 @@ static void device_set_state(Device *d, DeviceState state) {
...
         assert(d);
         assert(d->state == DEVICE_DEAD);
 
         /* First, let's put the deserialized state and found mask into effect, if we have it. */
+        if (d->deserialized_state < 0)
+                return 0;
+
+        m = u->manager;
+        f = d->deserialized_found;
+        s = d->deserialized_state;
```
keszybz: Why not write this as 
```C
Manager *m = u->manager;
DeviceFound found = d->deserialized_found;
DeviceState state = d->deserialized_state;
```
Much easier to read…

src/core/device.c
```diff
@@ -158,19 +158,65 @@ static void device_set_state(Device *d, DeviceState state) {
...
+
+        /* On initial boot, switch-root, reload, reexecute, the following happen:
+         * 1. MANAGER_IS_RUNNING() == false
+         * 2. enumerate devices: manager_enumerate() -> device_enumerate()
+         *    Device::enumerated_found is set.
```
keszybz: `::` is C++. Maybe just `Device.enumerated_found` ?

src/core/device.c
```diff
@@ -158,19 +158,65 @@ static void device_set_state(Device *d, DeviceState state) {
...
+         * 3. deserialize devices: manager_deserialize() -> device_deserialize()
+         *    Device::deserialize_state and Device::deserialized_found are set.
+         * 4. coldplug devices: manager_coldplug() -> device_coldplug()
+         *    deserialized properties are copied to the main properties.
+         * 5. MANAGER_IS_RUNNING() == true: manager_ready()
+         * 6. catchup devices: manager_catchup() -> device_catchup()
+         *    Device::enumerated_found is applied to Device::found, and state is updated based on that.
+         *
+         * Notes:
+         * - On initial boot, no udev database exist. Hence, no devices are enumerated in the step 2.
```
keszybz: exist → exists

src/core/device.c
```diff
@@ -158,19 +158,65 @@ static void device_set_state(Device *d, DeviceState state) {
...
+         *   Also, there is no deserialized device. Device units are (a) generated based on dependencies of
+         *   other units, or (b) generated when uevents are received.
+         *
+         * - On switch-root, no udev databse exist, except for devices with sticky bit, i.e. OPTIONS="db_persist".
```
keszybz: databse exist → database exists

Note that this description is only true for "traditional initramfs". We (I) stopped doing the db wipe on switch root now. Maybe change "no udev database exists" to "the udev database may be removed" ?

keszybz (commented):
This looks reasonable, but the logic is so complicated that I don't feel confident saying it's correct. Let's merge this and release -rc3 to give people a few days to test this.

keszybz:
mkosi / ci (fedora, 35): test-acl: sh: line 1: getfacl: command not found

mrc0mmand:
> mkosi / ci (fedora, 35): test-acl: sh: line 1: getfacl: command not found

That's being resolved in https://github.com/systemd/systemd/pull/23365.

yuwata:
> mkosi / ci (fedora, 35): test-acl: sh: line 1: getfacl: command not found

#23365.

fbuihuu:
> The macro `MANAGER_IS_SWITCHING_ROOT()` is true only when the _previous_ process in the shutting down stage. Hence, it is always false in `device_coldplug()` or `device_catchup()`.

@yuwata I don't think so, otherwise this code would have never been able to pass my basic testing. See the call to `manager_set_switching_root(m, arg_switched_root);` in `main()`.

> 
> > The fourth commit is more interesting because it drops `device_coldplug()` entirely. Since this function was basically used to restore the serialized device properties, which is supposed to be done be `device_deserialize_items()`. Hence it let `device_catchup()` decides for the final state of the device and generate the state transition event when needed.
> 
> If the (deserialized) found and the `enumerated_found` are equivalent, then it does not call `device_set_state()`...

Well that's the whole point of the patch: "forget" all devices after switching root that are no more referenced by udev nor by systemd (mount/swap) units.

I think you should reread it more carefully. This version seems to be simpler, which is rather a good argument since the logic is already complicated.

Thanks.

yuwata:
@fbuihuu Could you open a PR? Then we can easily review and discuss about the code there.

fbuihuu:
> @fbuihuu Could you open a PR? Then we can easily review and discuss about the code there.

OK I will.

mwilck:
> It's not clear to me whether it will cause issue in practice (maybe @mwilck knows ?) but this `dead->plugged` transition for devices with `db_persist` doen't look correct and wasn't present before. And udev man page seems to indicate that this transition should not occur.

Why would this even be mentioned in an the udev man page? It's genuine systemd functionality.

I have observed these transitions (I see them all the time actually because I work with multipath-tools, but I didn't bother about it), but definitely no negative impact. 

I have a test case here where root FS (btrfs) is on top of LVM on top of MD RAID on top of multipath. Thus the multipath devices (which have `db_persist` set) are in use once root is mounted, but from the PoV of systemd, they have neither `DEVICE_FOUND_MOUNT` nor `DEVICE_FOUND_SWAP` set. Thus they start out "dead" after switching root, just like in @fbuihuu's example. I can see these dead→plugged transitions with loglevel debug, but everything works just fine during boot.

mwilck:
@yuwata. recall [this discussion](https://github.com/systemd/systemd/pull/23218#discussion_r862654712) we had previously? If we removed the lines 
```
+                if (found == DEVICE_NOT_FOUND)
+                        state = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
```
from 75d7b59, we'd still see transition messages but only "tentative→plugged", not the confusing "dead→plugged". I don't think this can be completely avoided. We apply the result from enumeration after the result from deserialization on purpose.

mwilck:
Ah, and my example above shows that it can actually happen that `deserialized_state == DEVICE_PLUGGED` and `(deserialized_found & ~DEVICE_FOUND_UDEV) == DEVICE_NOT_FOUND`. (https://github.com/systemd/systemd/pull/23218#discussion_r862841031).

yuwata:
https://github.com/systemd/systemd/issues/13775#issuecomment-1211271191 requests to backport 75d7b5989f99125e52d5c0e5656fa1cd0fae2405. Setting the backport label.

keszybz:
@yuwata I know this and #23517 have been requested for backports, but I don't feel confident enough to do it myself. Could you open a PR against v250-stable?

yuwata:
Sure, I will.

keszybz:
@yuwata ping

nikita-dodia:
Hi All,
We have an issue where all devices change to dead state during systemd daemon-reload. The issue is seen on systemd-stable v-253.6.
The systemd daemon-reload is performed during startup. We found the pull request [#23215](https://github.com/systemd/systemd/pull/23215) on github that seems to address this issue.
We tracked that #23215 got superseeded by this pull request,but this pull request does not include
the change [57910e5](https://github.com/systemd/systemd/pull/23215/commits/57910e581ecd85de524c21a8eac213dd9c54732c)  that we need for our issue
Is there a specific reason why this change from #23215 got dropped when "replacing" it with the new pull request?

yuwata:
@nikita-dodia Could you open a new issue about your issue, if it can be reproducible with a recent release of systemd? Otherwise we can easily forget your comment on an already closed PR.

nikita-dodia:
We can only test it on latest release of v253-stable which is v253.29. Would you accept a new issue based on reproduction on this 253.29 version?

mwilck:
It's almost 3y ago and a really complicated matter.

I can just say that back then, after lengthy discussions with @yuwata and others, we concluded that #23218 was correct. And since then I haven't observed the kind of device failures during boot that were the topic of this PR. I have no clue about the state of the backports.

yuwata:
@nikita-dodia Sorry, v253 is too old here. Please reproduce with a recent release.

nikita-dodia:
> It's almost 3y ago and a really complicated matter.
> 
> I can just say that back then, after lengthy discussions with @yuwata and others, we concluded that #23218 was correct. And since then I haven't observed the kind of device failures during boot that were the topic of this PR. I have no clue about the state of the backports.

I have two questions regarding the fix:

1. Did your startup scenario change so that you are not concerned about doing systemctl daemon-reload during startup any more?
2. In the original fix #23215 you mentioned in the first point of the conversation that the issue is seen during systemctl reload. Was it systemctl reload for a service or daemon-reload for the complete manager?

nikita-dodia:
> @nikita-dodia BTW, maybe related to #9751 ?

Our current understanding is that the conditions are different (i.e we do not trigger udevadm trigger in parallel to systemctl daemon-reload) so its not related to #9751.

The issue that we see is occasional.

PR #23211, opened 2022-04-28
ci: bring fatal-meson-warnings back

evverx:
It's a follow-up to https://github.com/systemd/systemd/pull/23204

PR #23221, opened 2022-04-28
ci: actually turn on fatal-meson-warnings in the "build" workflow

evverx:
It's a follow-up to https://github.com/systemd/systemd/pull/23211

I'm not sure how I missed that.


yuwata (approved):
Ouch! LGTM.

Issue #23223, opened 2022-04-29
[Bug] Buffer overflow in sd-dhcp-client.c


yiyuaner:
In the file `libsystemd-network/sd-dhcp-client.c`, the function `client_receive_message_raw` has the following [code](https://github.com/systemd/systemd/blob/cfa7d57ba1bbc8317ad7757380a62abe8b77ef84/src/libsystemd-network/sd-dhcp-client.c#L1976):

```
static int client_receive_message_raw(
                sd_event_source *s,
                int fd,
                uint32_t revents,
                void *userdata)  {
    ...
    buflen = next_datagram_size_fd(fd);
    if (buflen < 0) {
        if (ERRNO_IS_TRANSIENT(buflen) || ERRNO_IS_DISCONNECT(buflen))
            return 0;

        log_dhcp_client_errno(client, buflen, "Failed to determine datagram size to read, ignoring: %m");
        return 0;
    }

    packet = malloc0(buflen);
    if (!packet)
        return -ENOMEM;    
    ...
    r = dhcp_packet_verify_headers(packet, len, checksum, client->port);
    ...
}
```

The problem is when `buflen = next_datagram_size_fd(fd) = 0`,  `packet` will be allocated to a one-byte buffer by `malloc0`.
Then it is passed to `dhcp_packet_verify_headers` as a normal `DHCPPacket`, which will cause buffer overflow.

evverx:
@yiyuaner since it doesn't seem to be the first bug report like that I wonder if it was generated by a rudimentary static analyzer of some kind that appears to completely ignore what happens to `packet` before it's passed to `dhcp_packet_verify_headers`?

yiyuaner:
> @yiyuaner since it doesn't seem to be the first bug report like that I wonder if it was generated by a rudimentary static analyzer of some kind that appears to completely ignore what happens to `packet` before it's passed to `dhcp_packet_verify_headers`?

Yes, this is a report generated by a static analyzer. 

Before `packet` is passed to `dhcp_packet_verify_headers`,  the only thing important is this call:
```
len = recvmsg_safe(fd, &msg, 0);
```

If this call to `recvmsg_safe` **must** fail when `buflen=0`, then the code is fine. But I don't think this is always the case (or any specification of recv can guarantee it).

evverx:
[...]

Issue #22873, opened 2022-03-26
[Bug] Passing zero length value to recv in sd-dhcp-client.c is an undefined behavior


yiyuaner:
The function `client_receive_message_udp` in the file `sd-dhcp-client.c` has the following [code](https://github.com/systemd/systemd/blob/017228934833d8618774353dcccbc5ec9f78ce98/src/libsystemd-network/sd-dhcp-client.c#L1906-#L1919):

```
        buflen = next_datagram_size_fd(fd);
        if (buflen < 0) {
                if (ERRNO_IS_TRANSIENT(buflen) || ERRNO_IS_DISCONNECT(buflen))
                        return 0;

                log_dhcp_client_errno(client, buflen, "Failed to determine datagram size to read, ignoring: %m");
                return 0;
        }

        message = malloc0(buflen);
        if (!message)
                return -ENOMEM;

        len = recv(fd, message, buflen, 0);
```
The call to `next_datagram_size_fd(fd)` could return a zero value. When `buflen` is zero, `message` is allocated to an one-byte buffer and `recv` is passed a zero length value.  This is undefined behavior and should be avoided.

yuwata:
> When `buflen` is zero, `message` is allocated to an one-byte buffer and `recv` is passed a zero length value. This is undefined behavior and should be avoided.

From recv(2), I cannot find anything about undefined behavior. Moreover, it says
> The value 0 may also be returned if the requested number of bytes to receive from a stream socket was 0.

Could you give us references about the undefined behavior?

yiyuaner:
@yuwata There seems to be some controversies on this. Check this stack overflow [link](https://stackoverflow.com/questions/5994724/recv-with-length-zero-is-valid#:~:text=Returning%20zero%20normally%20means%20that,recv()%20with%20length%20%3D%200.).

Personally, I think it will be clearer if the code handles the `len==0` case separately.

If you are certain that it won't cause any issue for the type of socket used here, please close this issue. Thanks.

evverx:
`recv` there is intentional: https://github.com/systemd/systemd/pull/3317.

yuwata:
@evverx Thank you for the pointer.

PR #23226, opened 2022-04-29
Add assert about DHCP packet length in libsystemd-network

keszybz:


yuwata (approved):
LGTM.

evverx:
FWIW In https://github.com/systemd/systemd/issues/23223 the code was analyzed by a static analyzer of some kind and I think with `assert` it would have reported either a crash caused by that `assert` or if it weeds out `assert`s because they can be compiled out it would have still reported the buffer overflow.

evverx:
Looking at https://github.com/systemd/systemd/issues/22873 I think that that static analyzer is based on stack overflow questions and it would report stuff like that until it's fixed regardless of whether `assert`s are used or not. Having said that I think it makes sense to add it. It's just that it can't help avoid getting bug reports like that.

keszybz:
Please note that I didn't say that this commit fixes anything. I called it a "clarification". ;)

evverx:
@keszybz yeah, I saw "let's add an assert to clarify this for the reader". What I was trying to say is that static analyzers generating those "bug reports" will keep ignoring it anyway. It would be great if experimental stuff like that was verified and cross-verified by other more stable tools before dumping it here and letting someone else figure out whether those are false positives or not. At least that's how research in that space is supposed to be done.

Issue #23224, opened 2022-04-29
systemd-analyze security segfaults on some units


vedranmiletic:
**systemd version the issue has been seen with**

```
systemd 250 (250.5-1-arch)
+PAM +AUDIT -SELINUX -APPARMOR -IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK +XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified
```

**Used distribution**

Arch Linux

**Linux kernel version used** (`uname -a`)

`Linux ares.miletic.net 5.17.5-arch1-1 #1 SMP PREEMPT Wed, 27 Apr 2022 20:56:11 +0000 x86_64 GNU/Linux`

**CPU architecture issue was seen on**

x86-64

**Expected behaviour you didn't see**

No segfault

**Unexpected behaviour you saw**

Segfault

**Steps to reproduce the problem**

```
$ systemd-analyze security systemd-udevd.service
$ systemd-analyze security systemd-networkd.service
$ systemd-analyze security systemd-journald.service
$ systemd-analyze security systemd-logind.service
```

**Additional program output to the terminal or log subsystem illustrating the issue**

Output from each of these commands:

```
Segmentacijska greška (**krah**: stanje (dump) memorije je zapisano!)
```

dmesg:

```
[89654.403252] systemd-analyze[274922]: segfault at 1 ip 00007fb2c66604bd sp 00007fff50764478 error 4 in libc.so.6[7fb2c6509000+176000]
[89654.403283] Code: f8 77 c3 66 2e 0f 1f 84 00 00 00 00 00 f3 0f 1e fa 89 f8 48 89 fa c5 f9 ef c0 25 ff 0f 00 00 3d e0 0f 00 00 0f 87 33 01 00 00 <c5> fd 74 0f c5 fd d7 c1 85 c0 74 57 f3 0f bc c0 c5 f8 77 c3 66 66
```

mrc0mmand:
Also reproducible with the latest main:

```gdb
 bt
[...]

PR #23228, opened 2022-04-29
analyze: fix crash with online security check

bluca:
1449b0f8a96b27 fixed seccomp arch check for the offline case,
but broke it for the normal case, as when coming from D-Bus the
list of seccomp architectures is already converted to string.

Fixes https://github.com/systemd/systemd/issues/23224

### DIFF
--- a/src/analyze/analyze-security.c
+++ b/src/analyze/analyze-security.c
@@ -539,19 +539,16 @@ static int assess_system_call_architectures(
                 uint64_t *ret_badness,
                 char **ret_description) {
 
-        uint32_t native = 0;
         char *d;
         uint64_t b;
 
         assert(ret_badness);
         assert(ret_description);
 
-        assert_se(seccomp_arch_from_string("native", &native) >= 0);
-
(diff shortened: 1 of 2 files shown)


yuwata:
@bluca The offending commit was backported to v250-stable, hence, this also needs to be backported.

bluca:
> @bluca The offending commit was backported to v250-stable, hence, this also needs to be backported.

It was? Oh, forgot about that, thanks

yuwata (approved):
LGTM.

keszybz:
We have this fix in v251, and since the issue was introduced in v251-rc1-42-g1449b0f8a9, it isn't needed elsewhere.

yuwata:
> We have this fix in v251, and since the issue was introduced in v251-rc1-42-g1449b0f8a9, it isn't needed elsewhere.

No no, please read the avobe discussion.

yuwata:
Backport is waiting at https://github.com/systemd/systemd-stable/pull/208

Issue #23212, opened 2022-04-28
systemd fails to compile with --fatal-meson-warnings


evverx:
Since https://github.com/systemd/systemd/pull/23204 was merged `systemd` has been failing to compile with `--fatal-meson-warnings`

```sh
meson --version
0.62.0

meson --fatal-meson-warnings build
...
src/boot/efi/meson.build:429: WARNING: Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target.

src/boot/efi/meson.build:415:8: ERROR: Fatal warnings enabled, aborting
```

evverx:
https://github.com/systemd/systemd/pull/23204#issuecomment-1113696550
```python
if meson.version().version_compare('>=0.62.0')
    custom_target(..., install_tag: 'foo')
else
    custom_target(...)
endif
```
is the canonical way to avoid warnings.

eli-schwartz:
What was the ultimate resolution of this?

evverx:
systemd still fails to compile with `--fatal-meson-warnings` and it's totally fine by systemd standards. It will keep failing to build until certain distributions can switch to meson >= 0.60.

evverx:
As far as I can see there was an attempt to bump meson a bit in https://github.com/systemd/systemd/pull/26728 but looks like it was opened without asking permission first.

keszybz:
We're waiting for ~~the anchor to be lifted~~ Debian Bookworm to be released.

evverx:
I understand that systemd follows that release cycle. But given that it seems that that half-baked rust integration is on its way with its warnings I don't think that bumping meson can help much with `--fatal-meson-warnings` generally.

eli-schwartz:
I see. :/

(By the way, github now supports closing an issue as "not planned" to distinguish it from "completed".)

evverx:
Apologies. I closed a bunch of issues that day so I might have missed the right button.

evverx:
For the record https://github.com/systemd/systemd/pull/35946 broke it again.

PR #23229, opened 2022-04-30
meson: add install_tag conditionally

evverx:
Based on https://github.com/systemd/systemd/pull/23204#issuecomment-1113696550

Closes https://github.com/systemd/systemd/issues/23212

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -123,11 +123,6 @@ for args in "${ARGS[@]}"; do
     #   src/boot/efi/meson.build:52:16: ERROR: Fatal warnings enabled, aborting
     # when LINKER is set to lld so let's just not turn meson warnings into errors with lld
     # to make sure that the build systemd can pick up the correct efi-ld linker automatically.
-
-    # The install_tag feature introduced in 0.60 causes meson to fail with fatal-meson-warnings
-    # "Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target"
-    # It can be safely removed from the CI since it isn't actually used anywhere to test anything.
-    find . -type f -name meson.build -exec sed -i '/install_tag/d' '{}' '+'
     if [[ "$LINKER" != lld ]]; then
         additional_meson_args="--fatal-meson-warnings"
     fi
--- a/.github/workflows/unit_tests.sh
+++ b/.github/workflows/unit_tests.sh
@@ -60,10 +60,6 @@ for phase in "${PHASES[@]}"; do
                 # The docs build is slow and is not affected by compiler/flags, so do it just once
                 MESON_ARGS+=(-Dman=true)
             fi
-            # The install_tag feature introduced in 0.60 causes meson to fail with fatal-meson-warnings
-            # "Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target"
-            # It can be safely removed from the CI since it isn't actually used anywhere to test anything.
-            find . -type f -name meson.build -exec sed -i '/install_tag/d' '{}' '+'
             MESON_ARGS+=(--fatal-meson-warnings)
             run_meson -Dnobody-group=nogroup --werror -Dtests=unsafe -Dslow-tests=true -Dfuzz-tests=true "${MESON_ARGS[@]}" build
             ninja -C build -v
@@ -84,10 +80,6 @@ for phase in "${PHASES[@]}"; do
                     MESON_ARGS+=(-Dskip-deps=true)
                 fi
             fi
-            # The install_tag feature introduced in 0.60 causes meson to fail with fatal-meson-warnings
-            # "Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target"
-            # It can be safely removed from the CI since it isn't actually used anywhere to test anything.
-            find . -type f -name meson.build -exec sed -i '/install_tag/d' '{}' '+'
             MESON_ARGS+=(--fatal-meson-warnings)
             run_meson -Dnobody-group=nogroup --werror -Dtests=unsafe -Db_sanitize=address,undefined "${MESON_ARGS[@]}" build
             ninja -C build -v


meson.build
```diff
@@ -1912,7 +1912,6 @@ libsystemd = shared_library(
                         librt],
         link_depends : libsystemd_sym,
         install : true,
-        install_tag: 'libsystemd',
```
bluca, line 1915: this was requested here https://github.com/systemd/systemd/pull/23204#issuecomment-1110977393

src/boot/efi/meson.build
```diff
@@ -412,43 +412,67 @@ endforeach
...
+        elf_command = [cc.cmd_array(),
+                   '-o', '@OUTPUT@',
+                   efi_cflags,
+                   efi_ldflags,
+                   '@INPUT@',
+                   '-lefi',
+                   '-lgnuefi',
+                   '-lgcc']
+
+        if meson.version().version_compare('>=0.60.0')
```
bluca, line 430: More install_tag will be added soon - I don't think all this duplication and verbosity is needed, just to save one line in the local CI script. Better to just leave the workaround in the CI script. None of the major distributions use `--fatal-meson-warnings` anyway, it's just in our CI, so it's not very important:

https://git.launchpad.net/ubuntu/+source/systemd/tree/debian/rules
https://tracker.debian.org/media/packages/s/systemd/rules-251rc1-3
https://src.fedoraproject.org/rpms/systemd/blob/rawhide/f/systemd.spec
https://build.opensuse.org/package/view_file/Base:System/systemd/systemd.spec?expand=1

bluca (changes requested):

meson.build
evverx: I might be wrong but I think it was requested in the sense that it would probably be nice to have those tags eventually. It was before it was pointed out how kludgy it is in terms of making it work with all meson versions `systemd` claims to support.

src/boot/efi/meson.build
evverx: > More install_tag will be added soon

I'm not sure why they would be added. If it was discussed somewhere could you point me in the right direction?

> I don't think all this duplication and verbosity is needed

I don't think `install_tags` is needed in its current form either.

> just to save one line in the local CI script

The CI script was just the first place where `systemd` failed to build because `meson` rightfully complained. It's not the only place where it's used.

> None of the major distributions use --fatal-meson-warnings anyway

Is my understand correct that `systemd` doesn't care about backward compatibility in general if major distributions are intact? If so could you add it to `README` to make it a bit more obvious?

medhefgo:
The clean way to do this would be to use `kwargs`:
```meson
args = { ... }
if meson.version().version_compare('>=0.62.0')
    args += { 'install_tag': 'systemd-boot' }
endif
custom_target(..., kwargs: args)
```

But quite frankly, I am not convinced we need this and would much rather we depend on a newer version of meson anyways for the new fs module methods. It's not like older distros are gonna upgrade to latest systemd without also updating meson. Meanwhile, for devs and other self-compilers, a new meson is just a `pip install meson` away.

evverx:
> The clean way to do this would be to use kwargs

With `kwargs` meson still complains about `install_tag`
```python
        if meson.version().version_compare('>=0.60.0')
                elf_kwargs += { 'install_tag' : tuple[3] }
        endif
        elf = custom_target(
                tuple[0].format(efi_arch[0], 'elf'),
                kwargs : elf_kwargs
        )
```
```
src/boot/efi/meson.build:437: WARNING: Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target.

src/boot/efi/meson.build:435:8: ERROR: Fatal warnings enabled, aborting
```

eli-schwartz:
Using `kwargs: ` is, in general, a valid way to only pass new kwargs on new versions of meson.

The problem is that on new versions of meson it will also still warn you... simply because Meson is clever enough to suppress FeatureNew warnings inside a version guard, but not clever enough to detect when a dictionary of strings is built in a version guard and used outside of one.

It would be fantastic if someone could figure out a clever hack to propagate this information, but in the meantime if your goal is to be warning-free then writing out both branches (or waiting before using the feature) seems like the best bet.

meson.build
yuwata: > this was requested here [#23204 (comment)](https://github.com/systemd/systemd/pull/23204#issuecomment-1110977393)

It did not clarify explicit usages. It just said 'probably ...'.

From https://github.com/systemd/systemd/pull/23204#issuecomment-1110977393,
> In general, I think this is a very nice feature. We should probably figure out what tags we'd want to use. We'd probably want on for `libsystemd`, maybe also one for the standalone variants of tools, and one for tests, man pages, html pages.

yuwata (approved):
I like this PR. If there is a way to avoid emitting warnings, then we should do that.
Let's revert the hack when we bump the requirement for meson.

yuwata:
> But quite frankly, I am not convinced we need this and would much rather we depend on a newer version of meson anyways for the new fs module methods. It's not like older distros are gonna upgrade to latest systemd without also updating meson. Meanwhile, for devs and other self-compilers, a new meson is just a `pip install meson` away.

https://github.com/systemd/systemd/pull/23204#issuecomment-1111100476
> I don't think we can require 0.60 yet. F35 has 0.59.4, debian stable (bullseye) has 0.56.2. But meh, we can live with a warning on older systems.

How about to bump the requirement after systemd v251 is released? At least, f36 will be released soon, and it seems to have meson-0.62.

keszybz:
> At least, f36 will be released soon, and it seems to have meson-0.62.

This doesn't help much. We want to support all extant fedora releases, so we'd need to wait until F34 and F35 are retired, i.e. approx. 13 months from now.

yuwata:
> > At least, f36 will be released soon, and it seems to have meson-0.62.
> 
> This doesn't help much. We want to support all extant fedora releases, so we'd need to wait until F34 and F35 are retired, i.e. approx. 13 months from now.

I see, thanks. But that fact does not prohibit this PR being merged, right?

keszybz:
Sorry, but I think we shouldn't merge this.

`--fatal-meson-warnings` is a tool for *developers*. Just like `-Werror`, it is not something that is useful for downstream packagers. In particular, systemd config will already emit warnings about nobody, and split-usr, non-default prefixes and some other stuff. So this warning is just one more warning that you might need to ignore when building on an older system, or even one that is non-standard in some ways.

The important part is that the CI works. And we have that today. The diff in the patch is terrible: we are replacing three ugly but straightforward `find` calls with a lot of churn. A few months down the road we'll want to drop the workarounds, and it's much nicer to do that with the current `find`s. As @bluca said, we'll want to add more tags. Doing the same complicated dance for every case just isn't worth the gains.

src/boot/efi/meson.build
keszybz: > Is my understand correct that `systemd` doesn't care about backward compatibility in general if major distributions are intact? If so could you add it to `README` to make it a bit more obvious?

That is not true. We definitely care about backwards compat. But sometimes there's no good answer, and we need to make a choice based on the trade-offs involved. In *that* case, something major like Debian or Fedora weighs more significantly, because they simply have the most users, and it's where the developer are, etc. This is particularly true in a case like here, where the issue is not fatal, but more of an annoyance.

evverx:
> --fatal-meson-warnings is a tool for developers

I don't think it is. It helps to catch various issues when packages are built.

> Just like -Werror, it is not something that is useful for downstream packagers

I think it's debatable. `-Werror` with stable compilers helps not to ship stuff riddled with buffer overflows, uninitialized variables and so on. I agree it's easier to turn it off and hope for the best though.

> In particular, systemd config will already emit warnings about nobody, and split-usr, non-default prefixes and some other stuff

If it was always the case the CI couldn't use it. Then again this feature isn't useful for build pipelines where warnings are routinely ignored and tests are run basically just in case.

> So this warning is just one more warning that you might need to ignore when building on an older system

As far as I understand new tags are supposed to be added and every place with `install_tag` will produce warnings like that so I wouldn't say it's just one warning. There are going to be a lot of warnings.

> The important part is that the CI works. And we have that today.

The knee-jerk reaction was to turn it off altogether though. And this kind of pushing stuff till it works is what concerns me.

> Sorry, but I think we shouldn't merge this.

To be honest I didn't expect it to be merged but as far as I can see I haven't been outvoted here yet. One more NACK should do it though

src/boot/efi/meson.build
evverx: > But sometimes there's no good answer, and we need to make a choice based on the trade-offs involved

Can't argue with that. I think what I don't understand is why this feature is so important. Was it discussed anywhere apart from https://lists.freedesktop.org/archives/systemd-devel/2022-April/047797.html? By the way even there I haven't been able to find any confirmation that this is actually needed.

keszybz:
> > --fatal-meson-warnings is a tool for developers
> 
> I don't think it is. It helps to catch various issues when packages are built.
> 
> > Just like -Werror, it is not something that is useful for downstream packagers
> 
> I think it's debatable. `-Werror` with stable compilers helps not to ship stuff riddled with buffer overflows, uninitialized variables and so on. I agree it's easier to turn it off and hope for the best though.

In my experience, `-Werror` is problematic because even minor changes to the compiler version can result in the compiler suddenly emitting new warnings. So yeah, if you have some fixed version of the compiler for a specific version of the project, it works. But if it's all fixed, what would the point of rebuilds be? If you're rebuilding, it means that something must be changing, and especially with optimization on, `-Werror` can strike at any time.
 
> > The important part is that the CI works. And we have that today.
> 
> The knee-jerk reaction was to turn it off altogether though. And this kind of pushing stuff till it works is what concerns me.

I think "knee-jerk" is overstating it. It was turned off with the assumption that we'll figure out a fix later on, which we did. And "pushing stuff till it works" is in general how this all works: we push the code until the unit tests pass, then until the CI passes, then until the fuzzers don't complain, and move on to the next pressing issue. I don't think we even pretend that it's all bug-free.

> Was it discussed anywhere
apart from https://lists.freedesktop.org/archives/systemd-devel/2022-April/047797.html?

It's a long-standing problem. Autotoolz had install targets, and for the conversion to meson `-Dinstall-tests` was invented. It works, but not very well (since it's unintuitive and requires a pointless recompilation when changed). Similarly, people have been asking to be able to install only some parts for ages and ages. Our answer always was that you should just install everything and pick out the parts that you want. This solution sort-of-works, especially in something like rpm builds where you want to build everything, but then split it between packages at the very end. But I think it's fair to say that we said that because we didn't have anything better. Install tags can be clearly superior for some cases. We're just starting to experiment with this, so it's  we don't know where it'll all go, but I expect that we'll find a bunch of uses.

evverx:
> In my experience, -Werror is problematic

I agree it isn't perfect and I agree it shouldn't be turned on by default. What I was trying to say is that it's not for developers only and in some cases it's useful for people packaging stuff as well

> for the conversion to meson -Dinstall-tests was invented. It works, but not very well (since it's unintuitive and requires a pointless recompilation when changed)

Given that `meson` doesn't support targets with tags that can't be installed by default I don't think it's a good example. Even if `meson` did support that removing `-Dinstall-tests` would break the debian package along with Ubuntu CI and 
 unlike the upstream CI that isn't something that can be broken apparently

> Similarly, people have been asking to be able to install only some parts for ages and ages

I don't doubt there are use cases where it's needed but given that some distros are stuck with `meson < 0.60` for various reasons I'm not sure when this feature is going to be used in practice actually.

evverx:
> I think "knee-jerk" is overstating it

I'm sorry but I don't think it was an overstatement. I still remember that patch where buffer overflows were "fixed" with a fork and I was told there that the feature was optional and I didn't have to use it (as usual when I review patches like that). I'm afraid this pattern sprawls and it would be great if it could be stopped.

>  It was turned off with the assumption that we'll figure out a fix later on

There wasn't any assumptions like that in that PR. It was just turned off.

Anyway, what I'm getting at is that I think this PR should be closed and that PR should be reverted (apart from the part where the aliases suggested by @medhefgo were introduced. That part makes sense). I'm pretty sure once `meson>=0.60.0` is more widely adopted it would be more convincing.

keszybz:
meson-0.62.1-1 is in updates for Fedora [34](https://bodhi.fedoraproject.org/updates/FEDORA-2022-87318054d1) and [35](https://bodhi.fedoraproject.org/updates/FEDORA-2022-9c317a63c7).

evverx:
> meson-0.62.1-1 is in test updates for Fedora [34](https://bodhi.fedoraproject.org/updates/FEDORA-2022-87318054d1) and [35](https://bodhi.fedoraproject.org/updates/FEDORA-2022-9c317a63c7).

Good to know. Thanks. I wonder why meson can't be bumped properly in `meson.build` then?

keszybz:
> why meson can't be bumped properly in `meson.build` then?

Because that'd make it *impossible* to build on older systems without patching the config file. A warning during build is annoying, but an outright error is worse.

evverx:
> that'd make it impossible to build on older systems

Why would the latest systemd be built on systems where meson is ancient?

evverx:
Anyway, if meson can't be bumped I think that PR should be reverted. I'll keep this PR open until it receives one more NACK though. All the arguments I've heard so far about why this feature should be kept in its current form to cover hypothetical use cases weren't convincing (to me at least).

evverx:
Looks like Gentoo is the first distribution that is going to actually use this feature so it's now less hypothetical. Their use case doesn't cover scenarios where the nss modules and the development headers are packaged separately though (making it somewhat less granular than it was initially advertised). As usual It wasn't discussed anywhere and was already merged so it is what it is apparently.

evverx:
Anyway, since I saw real-life use case I'll go ahead and close this PR. I just hope `systemd` stops accepting patches like that dragging the upstream CI down along with them. https://github.com/bus1/dbus-broker/pull/273 would be a great example of how it should be done ideally.

floppym:
> Looks like Gentoo is the first distribution that is going to actually use this feature so it's now less hypothetical.

Speaking as the Gentoo maintainer: I was actually planning to wait until the minimum meson version was bumped for some other reason before submitting patches to sprinkle `install_tag` on various targets.

It's really just an optimization for us in that it allows us to build/install a subset of systemd while still utilizing `meson install`. If you want to revert all the `install_tag` stuff until some later date, I have no strong objection.

evverx:
@floppym sorry. I didn't mean to say that https://github.com/systemd/systemd/pull/23378 should be reverted. My objection to `install_tag` was based on the assumption that all the use cases I saw at the time were either hypothetical or unsupported by `meson` and https://github.com/systemd/systemd/pull/23378 proved me wrong. If it's actually used I have no problem with keeping it and (temporarily) getting the warnings around with that "sed" kludge. I think https://github.com/systemd/systemd/pull/23391 (where `meson` was bumped) is the way to go and there it should be possible to figure out why `meson-0.54` and `meson-0.56` should be kept afloat upstream. My guess would be that it's partly because of Ubuntu CI and Debian CI but http://ppa.launchpad.net/upstream-systemd-ci/systemd-ci/ubuntu was supposed to keep them in sync with the systemd repository. It hasn't been actively maintained for a while but I guess it's time to start maintaining that stuff again (if the idea is to still test systemd upstream instead of making sure the package is buildable with `meson-0.54).

PR #23233, opened 2022-04-30
Fix orientation detection for Asus Transformer T100TAF, copied T100TA

techtino:
I have a 'Asus Transformer T100TAF', and orientation was incorrect in landscape mode. After a bit of digging I found that the T100TA model of the same device has this fixed. I've copied the line for the T100TA that is already included in the hwdb config file. Tested it locally and the rule functions as expected. If this can be combined into the T100TA line easily, feel free to do so.


yuwata (approved):
LGTM.

PR #23234, opened 2022-04-30
core/device: ignore DEVICE_FOUND_UDEV on switching root

### LINKED ISSUE #12953: Devices transition plugged -> dead -> plugged is still not completely fixed

yuwata:
See the comments in the code.

This is based on the work by Martin Wilck.

Fixes #12953 and #23208.
Replaces #23215.
Split out of #23218.

### DIFF
--- a/src/core/device.c
+++ b/src/core/device.c
@@ -158,19 +158,65 @@ static void device_set_state(Device *d, DeviceState state) {
 
 static int device_coldplug(Unit *u) {
         Device *d = DEVICE(u);
+        DeviceFound f;
+        DeviceState s;
+        Manager *m;
 
         assert(d);
         assert(d->state == DEVICE_DEAD);
 
         /* First, let's put the deserialized state and found mask into effect, if we have it. */
-
-        if (d->deserialized_state < 0 ||
-            (d->deserialized_state == d->state &&
-             d->deserialized_found == d->found))
+        if (d->deserialized_state < 0)
                 return 0;
 
-        d->found = d->deserialized_found;
-        device_set_state(d, d->deserialized_state);
+        m = u->manager;
+        f = d->deserialized_found;
+        s = d->deserialized_state;
+
+        /* On initial boot, switch-root, reload, reexecute, the following happen:
+         * 1. MANAGER_IS_RUNNING() == false
+         * 2. enumerate devices: manager_enumerate() -> device_enumerate()
+         *    Device::enumerated_found is set.
+         * 3. deserialize devices: manager_deserialize() -> device_deserialize()
+         *    Device::deserialize_state and Device::deserialized_found are set.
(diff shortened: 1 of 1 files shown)


bluca:
> This is based on the work by Martin Wilck.

Github understands the line `Co-authored-by: name <email>` if you want to give trackable attribution

yuwata:
This is mostly equivalent in its effect to the second commit of #23215, but, the deserialized parameters are adjusted in `device_coldplug()`. Please also see https://github.com/systemd/systemd/pull/23218#issuecomment-1113430645.

@mwilck 
I do not think the first commit of #23215 is necessary. Moreover, it makes devices removed during reload or reexec processes not handled correctly.

mwilck:
@yuwata, you're confusing me. Why do we have a separate PR for this now, which is also included in #23218?

> @mwilck I do not think the first commit of #23215 is necessary. Moreover, it makes devices removed during reload or reexec processes not handled correctly.

The commit is necessary for systemd reload operations that take place before the udev DB is fully initialized, e.g. during boot, as explained int the commit message (57910e58). In this case, the `DEVICE_FOUND_UDEV` bit would be clear in `d->enumerated_found`, and we'd reset `d->found` in `device_catchup()`, causing the `plugged→dead→plugged` transition (#11997, #12953).  If systemd is reloading, the only reason to reset `DEVICE_FOUND_UDEV` would  be the arrival of a REMOVE uevent. If this actually happens during a reload, the uevent will be handled after the reload, and the device state will be correct again, even if we haven't reset it in `device_catchup()`.

You're right that this does not hold for `systemctl daemon-reexec`, because `reexec` closes systemd's device monitor socket, creating a time window in which uevents might get lost. That's no problem for the switching-root case, where udev coldplug will fix device states. `systemctl daemon-reexec` without switching root might indeed miss uevents. The "risk" that my patch introduces is only that *a device is removed while systemd is re-executing*, and systemd misses the REMOVE uevent. This is quite unlikely, and easily fixed by running `systemctl restart systemd-udev-trigger.service`. IMO this risk is rather unimportant compared to the frequent issues with systemd *reload* operations during boot leading to "plugged→dead→plugged" transitions.

The only way to "fix" the reexec problem would be to distinguish between "reload" and "reexec" in `device_catchup()`, and apply my patch only in the "reload" case. But even if we did that, `daemon-reexec` triggered during the boot procedure would be at risk to falsely consider devices "dead" if it calls `device_enumerate` before udev coldplugging has finished. We know from past discussions about `udev settle` that determining when udev is **done** processing devices is basically impossible. IMHO the best way to avoid this would be to generally block or postpone "reexecute" operations (except for switching root) until `default.target` (or at least `local-fs.target`) has been reached. But that goes far beyond the scope of my PR.

yuwata:
@mwilck Thank you for the comment. I replied in #23218.

Issue #23220, opened 2022-04-28
ConditionDirectoryNotEmpty change with hidden files results in systemd-binfmt service starting and failing


puppetjoy:
**systemd version the issue has been seen with**
 > systemd 250 (250)
 > +PAM -AUDIT -SELINUX -APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID -CURL -ELFUTILS -FIDO2 +IDN2 -IDN -IPTC +KMOD -LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY -P11KIT -QRENCODE -BZIP2 +LZ4 -XZ -ZLIB +ZSTD -BPF_FRAMEWORK -XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified

**Used distribution**
 > Gentoo

**Linux kernel version used** (`uname -a`)

 > `Linux osprey 5.16.14-gentoo #1 SMP Fri Apr 15 18:52:25 EDT 2022 armv7l ARMv7 Processor rev 2 (v7l) Generic AM33XX (Flattened Device Tree) GNU/Linux`

**CPU architecture issue was seen on**
 > ARM

**Expected behaviour you didn't see**
 > systemd-binfmt service is skipped based on the condition that the directories /lib/binfmt.d /usr/lib/binfmt.d /usr/local/lib/binfmt.d /etc/binfmt.d /run/binfmt.d are empty except for hidden files. In systemd 249, with just hidden files installed by the package manager, the service is skipped.
```
> ls -al /lib/binfmt.d /usr/lib/binfmt.d /usr/local/lib/binfmt.d /etc/binfmt.d /run/binfmt.d
ls: cannot access '/lib/binfmt.d': No such file or directory
ls: cannot access '/usr/local/lib/binfmt.d': No such file or directory
ls: cannot access '/run/binfmt.d': No such file or directory
/etc/binfmt.d:
total 7
drwxr-xr-x  2 root root   3 Apr 28 14:05 .
drwxr-xr-x 66 root root 149 Apr 28 16:11 ..
-rw-r--r--  1 root root   0 Apr 22 13:48 .keep_sys-apps_systemd-0

/usr/lib/binfmt.d:
total 98
drwxr-xr-x  2 root root   3 Apr 28 13:51 .
drwxr-xr-x 57 root root 834 Apr 28 15:25 ..
[...]

PR #23236, opened 2022-04-30
stat-util: ignore hidden_or_backup_file when checking if dir is empty

bluca:
Commit https://github.com/systemd/systemd/commit/a068aceafbf
changed dir_is_emtpy_at to use FOREACH_DIRENT_IN_BUFFER instead of
FOREACH_DIRENT, but used dot_or_dotdot which just checks if the name
is literally '.' or '..' which is not enough, previous behaviour was
to ignore all hidden files, so restore that and add a test case.

Fixes https://github.com/systemd/systemd/issues/23220

### DIFF
--- a/src/basic/stat-util.c
+++ b/src/basic/stat-util.c
@@ -106,7 +106,7 @@ int dir_is_empty_at(int dir_fd, const char *path) {
         msan_unpoison(&buffer, n);
 
         FOREACH_DIRENT_IN_BUFFER(de, &buffer.de, n)
-                if (!dot_or_dot_dot(de->d_name))
+                if (!hidden_or_backup_file(de->d_name))
                         return 0;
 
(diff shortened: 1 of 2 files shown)


yuwata (approved):
LGTM.

poettering:
This isn't right. if we ignore hidden files, then it's not sufficient to read exactly 3 directory entries, we might have to read more.

poettering:
→ #23261

bluca:
> This isn't right. if we ignore hidden files, then it's not sufficient to read exactly 3 directory entries, we might have to read more.

whops, overlooked that - thanks for the follow-up

keszybz:
I reverted https://github.com/systemd/systemd/commit/a068aceafbf instead of trying to fix it up in v250-stable.

PR #22067, opened 2022-01-09
fuzz-bus-message: add a testcase triggering an invalid read

evverx:
on 32-bit architectures so that it won't get lost.
```
==169373==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf4900780 at pc 0xf75dd9cf bp 0xfff63438 sp 0xfff6342c
READ of size 1 at 0xf4900780 thread T0
    #0 0xf75dd9ce in bus_message_parse_fields ../src/libsystemd/sd-bus/bus-message.c:5272
    #1 0xf75b68c3 in bus_message_from_malloc ../src/libsystemd/sd-bus/bus-message.c:568
    #2 0x80497f3 in LLVMFuzzerTestOneInput ../src/libsystemd/sd-bus/fuzz-bus-message.c:30
    #3 0x804a0ba in main ../src/fuzz/fuzz-main.c:50
    #4 0xf6f1f468 in __libc_start_call_main (/lib/libc.so.6+0x25468)
    #5 0xf6f1f549 in __libc_start_main_impl (/lib/libc.so.6+0x25549)
    #6 0x804928b in _start (/home/vagrant/systemd/build/fuzz-bus-message+0x804928b)

0xf4900780 is located 0 bytes to the right of 16-byte region [0xf4900770,0xf4900780)
allocated by thread T0 here:
    #0 0xf7abb39b in malloc (/lib/libasan.so.6+0xb239b)
    #1 0xf744cf1f in memdup ../src/basic/alloc-util.c:16
    #2 0x8049723 in LLVMFuzzerTestOneInput ../src/libsystemd/sd-bus/fuzz-bus-message.c:28
    #3 0x804a0ba in main ../src/fuzz/fuzz-main.c:50
    #4 0xf6f1f468 in __libc_start_call_main (/lib/libc.so.6+0x25468)
```

PR #23246, opened 2022-05-02
meson: Improve public header tests

medhefgo:


medhefgo:
I also considered adding `-pedantic`, but we are using some extensions that I am not certain if they are used intentional or not:
```
/home/medhefgo/systemd/src/systemd/sd-bus.h:90:43: error: ISO C restricts enumerator values to range of ‘int’ [-Werror=pedantic]
   90 |         SD_BUS_CREDS_UNIQUE_NAME        = 1ULL << 31,
/home/medhefgo/systemd/src/systemd/sd-bus-vtable.h:48:54: error: ISO C restricts enumerator values to range of ‘int’ [-Werror=pedantic]
   48 |         _SD_BUS_VTABLE_CAPABILITY_MASK             = 0xFFFFULL << 40
/home/medhefgo/systemd/src/systemd/sd-id128.h:156:28: error: anonymous variadic macros were introduced in C99 [-Werror=variadic-macros]
  156 | #define sd_id128_in_set(a, ...) \
/home/medhefgo/systemd/src/systemd/sd-bus.h:59:43: error: use of C99 long long integer constant [-Werror=long-long]
   59 |         SD_BUS_CREDS_PID                = 1ULL << 0,
/home/medhefgo/systemd/src/systemd/sd-bus.h:105:53: error: comma at end of enumerator list [-Werror=pedantic]
  105 |         SD_BUS_MESSAGE_DUMP_SUBTREE_ONLY = 1ULL << 1,
/home/medhefgo/systemd/src/systemd/sd-bus-vtable.h:70:17: error: type of bit-field ‘type’ is a GCC extension [-Werror=pedantic]
   70 |         uint8_t type:8;
/home/medhefgo/systemd/src/systemd/sd-bus.h:93:43: error: use of C++11 long long integer constant [-Werror=long-long]
   93 |         SD_BUS_CREDS_AUGMENT            = 1ULL << 63, /* special flag, if on sd-bus will augment creds struct, in a potentially race-full way. */
/home/medhefgo/systemd/src/libudev/libudev.h:91:15: error: ISO C90 does not support ‘long long’ [-Werror=long-long]
   91 | unsigned long long int udev_device_get_seqnum(struct udev_device *udev_device);
```

This list is just an summary. Note that without `-pedantic`, we cannot automatically catch things like flexible array members for `-ansi`.
I know that coding style says we aren't that strict with the public headers, that's why I am mostly putting this out there in case these were an oversight.

yuwata (approved):

meson.build
```diff
@@ -4012,6 +4013,7 @@ foreach tuple : sanitizers
                         if fuzz_tests
                                 test('@0@_@1@_@2@'.format(b, c, sanitizer),
                                      env,
+                                     suite : 'sanitizers',
```
evverx: These are fuzz targets. They are just built with ASan/UBSan when `-Dfuzz-test=true` is passed to `meson`. I think they should go to the "fuzzers" suite as well.

evverx: On the other hand if the idea is to run the fuzz targets with `valgrind` using `--wrapper` they shouldn't be included in the "fuzzers" suite. I think maybe it would make sense to rename it to "fuzzers-sanitizers" or something like that. I'm not sure.

src/systemd/meson.build
```diff
@@ -73,13 +77,17 @@ endif
...
-                if want_tests != 'false'
-                        test(name,
-                             check_compilation_sh,
-                             args : cc.cmd_array() + ['-c', '-x'] + opt +
-                                    ['-Werror', '-include',
-                                     meson.current_source_dir() / header])
-                endif
+                test(''.join(['cc-', header.split('/')[-1], '_', opt[0], std_name]),
+                     env,
+                     suite : 'headers',
```
evverx, line 82: That's nice. Thanks! I'd add that apart from running just the "headers" suite it makes it possible to exclude those tests. It should make it easier to run the whole systemd testsuite under Valgrind because some parts including the "headers" suite have to be skipped with `--no-suite check-compilation headers`. With `bash` it failed under Valgrind. Now that `bash` is gone it just doesn't make sense to run the compiler under Valgrind.

meson.build
medhefgo: I renamed it to `fuzzers+sanitizers`. Hopefully, that makes more sense.

evverx: Thanks! Ideally this particular part of the systemd suite should be removed and Ubuntu CI limping along with no sanitizers whatsoever should be fixed to make it compatible with ASan/UBsan and run separate ASan/UBsan builds. That can't be addressed with a bunch of half-baked one-off patches though so probably it's never going to happen.

poettering:
> /home/medhefgo/systemd/src/systemd/sd-bus.h:90:43: error: ISO C restricts enumerator values to range of ‘int’ [-Werror=pedantic]

Interesting, I wasn't aware this was a gnuism, i assume this was ISO C...

poettering:
looks superficially OK to me, but I'll leave this to @keszybz, as my meson-fu is a lot more limited than his

meson.build
evverx: Looks like `-Dfuzz-tests=true` is no longer used on Ubuntu CI so I guess it should be safe to remove that suite eventually. It's redundant when separate ASan/UBSan builds are run and have to be edited every time it's run in new environments by analogy with https://github.com/systemd/systemd/pull/22067/commits/2ced9f86ee237bc163ff11a22f9a71f39d10e479 anyway.

src/systemd/meson.build
```diff
@@ -77,11 +77,14 @@ endif
 foreach header : _systemd_headers + _not_installed_headers + [libudev_h_path]
         foreach opt : opts
                 std_name = opt.length() == 2 ? '_'.join(opt[1].split(':')) : ''
-                name = ''.join(['cc-', header.split('/')[-1], '_', opt[0], std_name])
-                test(name,
-                     check_compilation_sh,
-                     args : cc.cmd_array() + ['-c', '-x'] + opt +
-                             ['-Werror', '-include',
-                             meson.current_source_dir() / header])
+                test(''.join(['cc-', header.split('/')[-1], '_', opt[0], std_name]),
```
keszybz: That join isn't doing much here…
`test('cc-' + header.split('/')[-1] + '_' + opt[0] + std_name, …` ?

meson.build
keszybz: This PR has the downside that meson prints the suite name in the test output:
```
915/1052 systemd:fuzzers / fuzz-varlink_array
308/314 systemd:fuzzers+sanitizers / fuzz-varlink_timeout-d8a88bf4adea54537d21e3afb396e1a55c5b58bf_address,undefined 
```

`systemd:fuzzers+sanitizers` is very wide again… Can we maybe make it narrower again? The name doesn't really need to be super precise.

keszybz (commented):
This looks quite useful, and I think we should do this.

Can you also add a suite for `check-version-*` and `check-help-*`? I don't have an idea for a name though… `version+help`? `distcheck` (because it was inspired by autotoolz' `dist-check`)? `output`? `console`?

meson.build
evverx: > Can we maybe make it narrower again?

@keszybz as long as the names are different I think it should be fine to rename that suite again. It's just that the fuzz targets built with ASan/UBSan can't be run with `--wrapper valgrind` along with the "fuzzer" suite.

keszybz: I don't understand your comment.

I was just asking for a string that is shorter than "fuzzers+sanitizers". If we can do that, that'd be great. If not, that's OK too.

keszybz:
> I also considered adding `-pedantic`, but we are using some extensions that I am not certain if they are used intentional or not:

Some of those we should fix. E.g. "comma at end of enumerator list". And the variadic macro arg thingy.

meson.build
medhefgo: > Looks like `-Dfuzz-tests=true` is no longer used on Ubuntu CI so I guess it should be safe to remove that suite eventually.

If you think these can be removed, why not make a PR for it? Then I can just skip giving it a name altogether :D

Otherwise: suggestions welcome. Or I could just keep it nameless regardless.

keszybz:
Looks reasonable. We can bikeshed about the target names later on ;)

PR #23249, opened 2022-05-02
man: correct a major missed opportunity

poettering:
I don't know why this didn't occur to me earlier, but of course, it
*has* to be this data.

(This replaces some German prose about Berlin, that i guess only very
few people will get. With the new blob I think we have a much broader
chance of delivering smiles.)

### DIFF
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -2820,11 +2820,13 @@ SystemCallErrorNumber=EPERM</programlisting>
 
         <programlisting>…
 StandardInput=data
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):
I wonder what this could be ...

bluca (approved):

medhefgo:
Nobody'll appreciate this…

medhefgo:
Wouldn't this be the perfect place to put your let's-take-over-the-linux-kernel-manifest?

PR #23252, opened 2022-05-02
nspawn: set LANG to SYSTEMD_DEFAULT_LOCALE

jameshilliard:
Not all systems have `C.UTF-8` support so lets use the `SYSTEMD_DEFAULT_LOCALE` which should be `C.UTF-8` on systems which have `C.UTF-8` support.

### DIFF
--- a/src/nspawn/nspawn.c
+++ b/src/nspawn/nspawn.c
@@ -3491,7 +3491,7 @@ static int inner_child(
                 /* If we're running a command in the container, let's default to the C.UTF-8 locale as it's
                  * part of glibc these days and was backported to most distros a long time before it got
                  * added to upstream glibc. */
-                envp[n_env] = strdup("LANG=C.UTF-8");
+                envp[n_env] = strdup("LANG=" SYSTEMD_DEFAULT_LOCALE);
                 if (!envp[n_env])
                         return log_oom();
                 n_env++;


thesamesam:
Thanks. I was just about to suggest this myself. It was added upstream in glibc-2.35 finally after a few false starts. While many distros did backport it, not all did. I remember seeing a few and being surprised when I checked in the past but I can't remember which hadn't.

In any case, glibc-2.35 was only released in Feb, and we can't be sure whether people did or didn't backport it, so it seems best to wait a while.

poettering:
I am not sure I agree with this. The thing is that the default locale so far was understood to apply to the host system if nothing else is set. But here we are talking about an nspawn container payload, i.e. *another* system than the host. It's quite a stretch to assume that the same locale that we think is right for the host also exists in the container.

C.UTF-8 otoh is kinda special: in recent glibc (and before that already on various distros) it has a special status and is available right in glibc itself, instead of an add-on locale package. 

Thus there's a good chance that it will exist and work in most modern container payloads.

Hence, I am not convinced we should merge this.

DaanDeMeyer:
Can we get some examples of systems that ship a recent version of systemd without shipping a recent version of glibc with C.UTF-8 included? I know Arch is one but we want to gently push Arch towards including C.UTF-8 in their glibc package by default.

I also agree with what @poettering said, `SYSTEMD_DEFAULT_LOCALE` doesn't seem like a good fit here.

poettering:
same for inheriting the locale from the host btw, i.e. propagating LANG from the calling process. For that too we have the issue that what exists on the host might not exist for the payload.

hence, I think we only have three options for the payload:

1. set nothing
2. set LANG=C
3. set LANG=C.UTF-8

because only these three cases are kinda universally supported these days. Of course, the first two have the same effect and will give you a 7bit ASCII charset which is garbage in 2022. Hence I think the third option is really the way to go.

thesamesam:
Okay, fair enough. Before it was finally merged in glibc, I was worried that it might deviate somewhat from the version everyone was using in distro, but bar some minor tweaks, that didn't happen.

I think it's worth including in the release notes pointedly for 251 to give a prominent nudge to any remaining distros though.

keszybz:
I suspect that if the payload has no support for `C.UTF-8`, it'll fall back to `C` anyway. So this shouldn't matter too much. Let's keep it as is, and if people start complaining  because their old payloads emit a warning about unknown locale, we can always make this (separately) configurable.

jameshilliard:
> Can we get some examples of systems that ship a recent version of systemd without shipping a recent version of glibc with C.UTF-8 included?

I don't think [we](https://buildroot.org/) really support it yet properly either for glibc based builds since we don't carry large glibc backports generally like the C.UTF-8 one. For some services(due to libarchive needing a UTF-8 locale it seems) I've had to set `en_US.UTF-8` in the env for them to work right at least.

PR #23254, opened 2022-05-03
sd-network: Keep inotify watch if watch descriptor didn't change

daandemeyer:
In sd_network_monitor_flush(), we shouldn't remove the inotify
watch for the current directory if the directory the network
monitor is waiting for wasn't created yet.

inotify_add_watch() returns the same unique watch descriptor if a
path is already being watched. Let's return the watch descriptor
from monitor_add_inotify_watch() so we can check if it's the same
as the watch descriptor of the inotify event. If they are equal,
we're still watching the same path and we don't need to remove the
inotify watch just yet.

### DIFF
--- a/src/libsystemd/sd-network/sd-network.c
+++ b/src/libsystemd/sd-network/sd-network.c
@@ -413,25 +413,25 @@ static sd_network_monitor* FD_TO_MONITOR(int fd) {
 }
 
 static int monitor_add_inotify_watch(int fd) {
-        int k;
+        int wd;
 
-        k = inotify_add_watch(fd, "/run/systemd/netif/links/", IN_MOVED_TO|IN_DELETE);
-        if (k >= 0)
-                return 0;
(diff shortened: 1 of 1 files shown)


yuwata (approved):

src/libsystemd/sd-network/sd-network.c
```diff
@@ -486,13 +486,16 @@ int sd_network_monitor_flush(sd_network_monitor *m) {
...
+                        if (wd < 0)
+                                return wd;
 
-                        k = inotify_rm_watch(fd, e->wd);
-                        if (k < 0)
-                                return -errno;
+                        if (wd != e->wd) {
+                                if (inotify_rm_watch(fd, e->wd) < 0)
+                                        return -errno;
+                        }
```
poettering, line 498: you should store `e->wd = wd;` somewhere here, no?

PR #23257, opened 2022-05-03
tests: make it possible to install valgrind

evverx:


mrc0mmand:
Looks good, thanks! The commits seem mis-attributed though:

```
From: Your Name <you@example.com>
Date: Tue, 3 May 2022 20:16:04 +0000
Subject: [PATCH 1/2] tests: make it possible to install valgrind
```

Is that on purpose?

evverx:
@mrc0mmand let me fix it. Long story short, I ran `dfuzzer` in a container and it somehow brought down the host in the sense that it was completely destroyed (which is interesting and shouldn't have happened) so I had to set up `git` again.

evverx:
The host was a VM so it wasn't that bad but still

mrc0mmand:
> The host was a VM so it wasn't that bad but still

That's indeed concerning (and interesting at the same time).

mrc0mmand (approved):

PR #23260, opened 2022-05-04
libsystemd-network: two fixlets

### LINKED ISSUE #23258: Possible bug on checking the size parameter in the function sd_ndisc_router_from_raw
In the file `ndisc-router.c`, the function `sd_ndisc_router_from_raw` has the following [code](https://github.com/systemd/systemd/blob/fa2ba7aea87ad1812f730f0c2ad86fbfe2ca6552/src/libsystemd-network/ndisc-router.c#L34):

```
int sd_ndisc_router_from_raw(sd_ndisc_router **ret, const void *raw, size_t raw_size) {
    _cleanup_(sd_ndisc_router_unrefp) sd_ndisc_router *rt = NULL;
    int r;

    assert_return(ret, -EINVAL);
    assert_return(raw || raw_size <= 0, -EINVAL);

    rt = ndisc_router_new(raw_size);
}
```

Why the assertion checks whether `raw_size` is less than or equal to zero?  Since  `raw_size` is an unsigned type, this check is actually equivalent to ` raw_size == 0`.

Moreover, if some caller calls the function as: `sd_ndisc_router_from_raw(ret, raw, -1)` , then the checks can be by passed. Inside `ndisc_router_new`,  an integer overflow will happen when computing the allocation size of the buffer, which will lead to subsequent buffer overflow:

```
sd_ndisc_router *ndisc_router_new(size_t raw_size) {
    sd_ndisc_router *rt;

    rt = malloc0(ALIGN(sizeof(sd_ndisc_router)) + raw_size);
    ...
}
```

The function `sd_ndisc_router_from_raw` seems to be public and the assertion in the code suggests the above scenario.  I think either there is a bug in the code or the assertion should be changed to be less misleading.

yuwata:
Closes #23258.

### DIFF
--- a/src/libsystemd-network/lldp-neighbor.c
+++ b/src/libsystemd-network/lldp-neighbor.c
@@ -652,7 +652,8 @@ int sd_lldp_neighbor_from_raw(sd_lldp_neighbor **ret, const void *raw, size_t ra
         if (!n)
                 return -ENOMEM;
 
-        memcpy(LLDP_NEIGHBOR_RAW(n), raw, raw_size);
+        memcpy_safe(LLDP_NEIGHBOR_RAW(n), raw, raw_size);
+
         r = lldp_neighbor_parse(n);
         if (r < 0)
                 return r;


evverx:
@yuwata I wonder why `sd_ndisc_router_from_raw` is kept in the repository? It looks like it has never been used anywhere. (except that it helped to find a potential overflow in `ndisc_recv`)

yuwata:
> @yuwata I wonder why `sd_ndisc_router_from_raw` is kept in the repository? It looks like it has never been used anywhere. (except that it helped to find a potential overflow in `ndisc_recv`)

@evverx Thanks. The function is dropped. PTAL.

DaanDeMeyer (approved):

evverx:
Looks like `assert_return(raw || raw_size <= 0, -EINVAL);` is used in `sd_lldp_neighbor_from_raw` as well and that's what triggered the static analyzer initially. `raw_size <= 0` is indeed misleading there.

evverx:
FWIW @yuwata I think it would make sense to mention in the commit messages whether they fix real or theoretical issues. For example, while `memcpy_safe` is better than `memcpy` it has never been called with `src = NULL` (and that's what UBSan usually complains about). Other than that `networkctl` where `sd_lldp_neighbor_from_raw` is used rejects data larger than 4Kib and so on.

keszybz:
LGTM too.

yuwata:
> FWIW @yuwata I think it would make sense to mention in the commit messages whether they fix real or theoretical issues. For example, while `memcpy_safe` is better than `memcpy` it has never been called with `src = NULL` (and that's what UBSan usually complains about). Other than that `networkctl` where `sd_lldp_neighbor_from_raw` is used rejects data larger than 4Kib and so on.

Unfortunately, it is already merged. But, I think, the commit message is clear; it states that the buffer size may be zero.
Of course, as you mentioned, the all and only user of the function passes non-zero buffer size, though.

PR #23269, opened 2022-05-04
networkd: NetLabel integration

topimiettinen:
New directives `IPv4NetLabelUnlabel=` and `IPv6NetLabelUnlabel=` provide a
method for integrating dynamic network configuration into Linux NetLabel
subsystem rules, used by Linux security modules (LSMs) for network access
control. The options expect a whitespace separated list of NetLabel
labels. The labels must conform to lexical restrictions of LSM labels. When
an interface is configured with IP addresses, the addresses and subnetwork
masks will be appended to the NetLabel Fallback Peer Labeling rules. They will
be removed when the interface is deconfigured. Failures to manage the labels
will be ignored.

Example:
```
IPv4NetLabelUnlabel=system_u:object_r:localnet_peer_t:s0
```

With the above rules for interface `eth0`, when the interface is configured with
an IPv4 address of 10.0.0.0/8, `systemd-networkd` performs the equivalent of
`netlabelctl` operation

```
$ sudo netlabelctl unlbl add interface eth0 address:10.0.0.0/8 label:system_u:object_r:localnet_peer_t:s0
```

Result:
```
$ sudo netlabelctl -p unlbl list
...
 interface: eth0
   address: 10.0.0.0/8
    label: "system_u:object_r:localnet_peer_t:s0"
...
```

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -1109,6 +1109,38 @@ Table=1234</programlisting></para>
           Defaults to <literal>no</literal>.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname><replaceable>label</replaceable></term>
+        <listitem>
+
+          <para>This setting provides a method for integrating dynamic network configuration into Linux
+          NetLabel subsystem rules, used by Linux security modules (LSMs) for network access control. The
+          option expects a whitespace separated list of NetLabel labels. The labels must conform to lexical
+          restrictions of LSM labels. When an interface is configured with IP addresses, the addresses and
+          subnetwork masks will be appended to the NetLabel Fallback Peer Labeling rules. They will be
+          removed when the interface is deconfigured. Failures to manage the labels will be ignored.</para>
+
+          <para>Warning: Once labeling is enabled for network traffic, a lot of LSM access control points in
+          Linux networking stack go from dormant to active. It is easy for someone not familiar with the LSM
+          per-packet access controls to get into a situation where for example remote connectivity is
+          broken. Also note that additional configuration with <citerefentry
+          project='man-pages'><refentrytitle>netlabelctl</refentrytitle><manvolnum>8</manvolnum></citerefentry>
+          is needed.</para>
+
+          <para>Example:
+          <programlisting>[Address]
+NetLabel=system_u:object_r:localnet_peer_t:s0</programlisting>
+
+          With the example rules applying for interface <literal>eth0</literal>, when the interface is
+          configured with an IPv4 address of 10.0.0.0/8, <command>systemd-networkd</command> performs the
+          equivalent of <command>netlabelctl</command> operation
+
+          <programlisting>netlabelctl unlbl add interface eth0 address:10.0.0.0/8 label:system_u:object_r:localnet_peer_t:s0</programlisting>
+
+          and the reverse operation when the IPv4 address is deconfigured.</para>
+        </listitem>
+      </varlistentry>
     </variablelist>
   </refsect1>
 
@@ -2050,6 +2082,13 @@ Table=1234</programlisting></para>
           <ulink url="https://tools.ietf.org/html/rfc5227">RFC 5227</ulink>. Defaults to false.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
+        </listitem>
+      </varlistentry>
     </variablelist>
   </refsect1>
 
@@ -2163,6 +2202,7 @@ Table=1234</programlisting></para>
         <term><varname>UseNTP=</varname></term>
         <term><varname>UseHostname=</varname></term>
         <term><varname>UseDomains=</varname></term>
+        <term><varname>NetLabel=</varname></term>
         <listitem>
           <para>As in the [DHCPv4] section.</para>
         </listitem>
@@ -2262,6 +2302,13 @@ Table=1234</programlisting></para>
           <para>The metric of the route to the delegated prefix subnet. Takes an unsigned integer in
           the range 0…4294967295. When set to 0, the kernel's default value is used. Defaults to 256.
           </para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
         </listitem>
       </varlistentry>
     </variablelist>
@@ -2519,6 +2566,13 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
           <varname>WithoutRA=</varname> in the [DHCPv6] section is enabled, or
           <varname>UplinkInterface=:self</varname> in the [DHCPPrefixDelegation] section is
           specified. Defaults to true.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
         </listitem>
       </varlistentry>
     </variablelist>
--- a/src/basic/in-addr-util.c
+++ b/src/basic/in-addr-util.c
@@ -595,6 +595,45 @@ struct in_addr* in4_addr_prefixlen_to_netmask(struct in_addr *addr, unsigned cha
         return addr;
 }
 
+struct in6_addr* in6_addr_prefixlen_to_netmask(struct in6_addr *addr, unsigned char prefixlen) {
+        assert(addr);
+        assert(prefixlen <= 128);
+
+        for (unsigned int i = 0; i < 16; i++) {
+                uint8_t mask;
+
+                if (prefixlen >= 8) {
+                        mask = 0xFF;
+                        prefixlen -= 8;
+                } else if (prefixlen > 0) {
+                        mask = 0xFF << (8 - prefixlen);
+                        prefixlen = 0;
+                } else {
+                        assert(prefixlen == 0);
+                        mask = 0;
+                }
+
+                addr->s6_addr[i] = mask;
+        }
+
+        return addr;
+}
+
+int in_addr_prefixlen_to_netmask(int family, union in_addr_union *addr, unsigned char prefixlen) {
+        assert(addr);
+
+        switch (family) {
+        case AF_INET:
+                in4_addr_prefixlen_to_netmask(&addr->in, prefixlen);
+                return 0;
+        case AF_INET6:
+                in6_addr_prefixlen_to_netmask(&addr->in6, prefixlen);
+                return 0;
+        default:
+                return -EAFNOSUPPORT;
+        }
+}
+
 int in4_addr_default_prefixlen(const struct in_addr *addr, unsigned char *prefixlen) {
         uint8_t msb_octet = *(uint8_t*) addr;
 
--- a/src/libsystemd/sd-netlink/netlink-types-genl.c
+++ b/src/libsystemd/sd-netlink/netlink-types-genl.c
@@ -221,15 +221,26 @@ static const NLType genl_wireguard_types[] = {
         [WGDEVICE_A_PEERS]       = { .type = NETLINK_TYPE_NESTED, .type_system = &genl_wireguard_peer_type_system },
 };
 
+/***************** genl NetLabel type systems *****************/
+static const NLType genl_netlabel_types[] = {
+        [NLBL_UNLABEL_A_IPV4ADDR] = { .type = NETLINK_TYPE_IN_ADDR, .size = sizeof(struct in_addr) },
+        [NLBL_UNLABEL_A_IPV4MASK] = { .type = NETLINK_TYPE_IN_ADDR, .size = sizeof(struct in_addr) },
+        [NLBL_UNLABEL_A_IPV6ADDR] = { .type = NETLINK_TYPE_IN_ADDR, .size = sizeof(struct in6_addr) },
+        [NLBL_UNLABEL_A_IPV6MASK] = { .type = NETLINK_TYPE_IN_ADDR, .size = sizeof(struct in6_addr) },
+        [NLBL_UNLABEL_A_IFACE]    = { .type = NETLINK_TYPE_STRING, .size = IFNAMSIZ-1 },
+        [NLBL_UNLABEL_A_SECCTX]   = { .type = NETLINK_TYPE_STRING },
+};
+
 /***************** genl families *****************/
 static const NLTypeSystemUnionElement genl_type_systems[] = {
-        { .name = CTRL_GENL_NAME,    .type_system = TYPE_SYSTEM_FROM_TYPE(genl_ctrl),      },
-        { .name = BATADV_NL_NAME,    .type_system = TYPE_SYSTEM_FROM_TYPE(genl_batadv),    },
-        { .name = FOU_GENL_NAME,     .type_system = TYPE_SYSTEM_FROM_TYPE(genl_fou),       },
-        { .name = L2TP_GENL_NAME,    .type_system = TYPE_SYSTEM_FROM_TYPE(genl_l2tp),      },
-        { .name = MACSEC_GENL_NAME,  .type_system = TYPE_SYSTEM_FROM_TYPE(genl_macsec),    },
-        { .name = NL80211_GENL_NAME, .type_system = TYPE_SYSTEM_FROM_TYPE(genl_nl80211),   },
-        { .name = WG_GENL_NAME,      .type_system = TYPE_SYSTEM_FROM_TYPE(genl_wireguard), },
+        { .name = CTRL_GENL_NAME,               .type_system = TYPE_SYSTEM_FROM_TYPE(genl_ctrl),      },
+        { .name = BATADV_NL_NAME,               .type_system = TYPE_SYSTEM_FROM_TYPE(genl_batadv),    },
+        { .name = FOU_GENL_NAME,                .type_system = TYPE_SYSTEM_FROM_TYPE(genl_fou),       },
+        { .name = L2TP_GENL_NAME,               .type_system = TYPE_SYSTEM_FROM_TYPE(genl_l2tp),      },
+        { .name = MACSEC_GENL_NAME,             .type_system = TYPE_SYSTEM_FROM_TYPE(genl_macsec),    },
+        { .name = NL80211_GENL_NAME,            .type_system = TYPE_SYSTEM_FROM_TYPE(genl_nl80211),   },
+        { .name = WG_GENL_NAME,                 .type_system = TYPE_SYSTEM_FROM_TYPE(genl_wireguard), },
+        { .name = NETLBL_NLTYPE_UNLABELED_NAME, .type_system = TYPE_SYSTEM_FROM_TYPE(genl_netlabel),  },
 };
 
 /* This is the root type system union, so match_attribute is not necessary. */
--- a/src/network/networkd-address.c
+++ b/src/network/networkd-address.c
@@ -12,6 +12,7 @@
 #include "networkd-dhcp-server.h"
 #include "networkd-ipv4acd.h"
 #include "networkd-manager.h"
+#include "networkd-netlabel.h"
 #include "networkd-network.h"
 #include "networkd-queue.h"
 #include "networkd-route-util.h"
@@ -137,6 +138,7 @@ Address *address_free(Address *address) {
 
         config_section_free(address->section);
         free(address->label);
+        set_free(address->netlabels);
         return mfree(address);
 }
 
@@ -492,6 +494,8 @@ static int address_update(Address *address) {
         if (r < 0)
                 return log_link_warning_errno(link, r, "Could not enable IP masquerading: %m");
 
+        address_add_netlabel(address);
+
         if (address_is_ready(address) && address->callback) {
                 r = address->callback(address);
                 if (r < 0)
@@ -517,6 +521,8 @@ static int address_drop(Address *address) {
         r = address_set_masquerade(address, false);
         if (r < 0)
                 log_link_warning_errno(link, r, "Failed to disable IP masquerading, ignoring: %m");
+
+        address_del_netlabel(address);
 
         if (address->state == 0)
                 address_free(address);
@@ -1937,6 +1943,41 @@ int config_parse_duplicate_address_detection(
 
         TAKE_PTR(n);
         return 0;
+}
+
+int config_parse_address_netlabel(
+                const char *unit,
+                const char *filename,
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
+
+        Network *network = userdata;
+        _cleanup_(address_free_or_set_invalidp) Address *n = NULL;
+        int r;
+
+        assert(filename);
+        assert(section);
+        assert(lvalue);
+        assert(rvalue);
+        assert(data);
+        assert(network);
+
+        r = address_new_static(network, filename, section_line, &n);
+        if (r == -ENOMEM)
+                return log_oom();
+        if (r < 0) {
+                log_syntax(unit, LOG_WARNING, filename, line, r,
+                           "Failed to allocate new address, ignoring assignment: %m");
+                return 0;
+        }
+
+        return config_parse_netlabel(unit, filename, line, section, section_line, lvalue, ltype, rvalue, &n->netlabels, network);
 }
 
 static int address_section_verify(Address *address) {
--- a/src/network/networkd-netlabel.c
+++ b/src/network/networkd-netlabel.c
@@ -1,0 +1,194 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "netlink-util.h"
+#include "networkd-address.h"
+#include "networkd-link.h"
+#include "networkd-manager.h"
+#include "networkd-netlabel.h"
+#include "networkd-network.h"
+
+static int netlabel_handler(sd_netlink *rtnl, sd_netlink_message *m, Link *link) {
+        int r;
+
+        assert_se(rtnl);
+        assert_se(m);
+        assert_se(link);
+
+        r = sd_netlink_message_get_errno(m);
+        if (r < 0) {
+                log_link_message_warning_errno(link, m, r, "NetLabel operation failed, ignoring");
+                return 1;
+        }
+
+        log_link_debug(link, "NetLabel operation successful");
+
+        return 1;
+}
+
+static int netlabel_command(uint16_t command, const char *label, const Address *address) {
+        _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
+        int r;
+
+        assert(address);
+        assert(address->link);
+        assert(address->link->manager);
+        assert(address->link->manager->genl);
+        assert(address->link->network);
+        assert(IN_SET(address->family, AF_INET, AF_INET6));
+
+        r = sd_genl_message_new(address->link->manager->genl, NETLBL_NLTYPE_UNLABELED_NAME, command, &m);
+        if (r < 0)
+                return r;
+
+        r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_IFACE, address->link->ifname);
+        if (r < 0)
+                return r;
+
+        if (command == NLBL_UNLABEL_C_STATICADD) {
+                assert(label);
+                r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_SECCTX, label);
+                if (r < 0)
+                        return r;
+        }
+
+        union in_addr_union netmask;
+
+        r = in_addr_prefixlen_to_netmask(address->family, &netmask, address->prefixlen);
+        if (r < 0)
+                return r;
+
+        if (address->family == AF_INET) {
+                r = sd_netlink_message_append_in_addr(m, NLBL_UNLABEL_A_IPV4ADDR, &address->in_addr.in);
+                if (r < 0)
+                        return r;
+
+                r = sd_netlink_message_append_in_addr(m, NLBL_UNLABEL_A_IPV4MASK, &netmask.in);
+        } else if (address->family == AF_INET6) {
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6ADDR, &address->in_addr.in6);
+                if (r < 0)
+                        return r;
+
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6MASK, &netmask.in6);
+        }
+        if (r < 0)
+                return r;
+
+        r = netlink_call_async(address->link->manager->genl, NULL, m, netlabel_handler, link_netlink_destroy_callback,
+                               address->link);
+        if (r < 0)
+                return r;
+
+        link_ref(address->link);
+        return 0;
+}
+
+static void address_add_netlabel_set(const Address *address, Set *labels) {
+        _cleanup_free_ char *addr_str = NULL;
+        int r;
+        const char *label;
+
+        (void) in_addr_prefix_to_string(address->family, &address->in_addr, address->prefixlen, &addr_str);
+
+        SET_FOREACH(label, labels) {
+                r = netlabel_command(NLBL_UNLABEL_C_STATICADD, label, address);
+                if (r < 0)
+                        log_link_warning_errno(address->link, r, "Adding NetLabel %s for IP address %s failed, ignoring",
+                                               label, strna(addr_str));
+                else
+                        log_link_debug(address->link, "Adding NetLabel %s for IP address %s", label, strna(addr_str));
+        }
+}
+
+void address_add_netlabel(const Address *address) {
+        assert(address);
+        assert(address->link);
+
+        if (!address->link->network || !IN_SET(address->family, AF_INET, AF_INET6))
+                return;
+
+        switch (address->source) {
+        case NETWORK_CONFIG_SOURCE_DHCP4:
+                return address_add_netlabel_set(address, address->link->network->dhcp_netlabels);
+        case NETWORK_CONFIG_SOURCE_DHCP6:
+                return address_add_netlabel_set(address, address->link->network->dhcp6_netlabels);
+        case NETWORK_CONFIG_SOURCE_DHCP_PD:
+                return address_add_netlabel_set(address, address->link->network->dhcp_pd_netlabels);
+        case NETWORK_CONFIG_SOURCE_NDISC:
+                return address_add_netlabel_set(address, address->link->network->ndisc_netlabels);
+        case NETWORK_CONFIG_SOURCE_STATIC:
+                return address_add_netlabel_set(address, address->netlabels);
+        default:
+                return;
+        }
+}
+
+void address_del_netlabel(const Address *address) {
+        int r;
+        _cleanup_free_ char *addr_str = NULL;
+
+        assert(address);
+        assert(address->link);
+
+        if (!address->link->network || !IN_SET(address->family, AF_INET, AF_INET6))
+                return;
+
+        (void) in_addr_prefix_to_string(address->family, &address->in_addr, address->prefixlen, &addr_str);
+
+        r = netlabel_command(NLBL_UNLABEL_C_STATICREMOVE, NULL, address);
+        if (r < 0)
+                log_link_warning_errno(address->link, r, "Deleting NetLabels for IP address %s failed, ignoring",
+                                       strna(addr_str));
+        else
+                log_link_debug(address->link, "Deleting NetLabels for IP address %s", strna(addr_str));
+}
+
+int config_parse_netlabel(
+                const char *unit,
+                const char *filename,
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
+        int r;
+        Set **set = data;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+        assert(set);
+
+        if (isempty(rvalue)) {
+                *set = set_free(*set);
+                return 0;
+        }
+
+        for (const char *p = rvalue;;) {
+                _cleanup_free_ char *w = NULL;
+
+                r = extract_first_word(&p, &w, NULL, 0);
+                if (r == -ENOMEM)
+                        return log_oom();
+                if (r < 0) {
+                        log_syntax(unit, LOG_WARNING, filename, line, r,
+                                   "Failed to extract NetLabel label, ignoring: %s", rvalue);
+                        return 0;
+                }
+                if (r == 0)
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        log_syntax(unit, LOG_WARNING, filename, line, 0,
+                                   "Bad NetLabel label, ignoring: %s", w);
+                        continue;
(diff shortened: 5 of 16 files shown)


topimiettinen:
Information for [Fallback Peer Labeling](https://github.com/SELinuxProject/selinux-notebook/blob/main/src/network_support.md#packet-controls-netlabel---fallback-peer-labeling) for SELinux could be helpful to the reviewers.

topimiettinen:
focal-amd failure most likely unrelated.

@yuwata PTAL.

src/libsystemd/sd-netlink/netlink-types-genl.c
```diff
@@ -221,15 +221,26 @@ static const NLType genl_wireguard_types[] = {
         [WGDEVICE_A_PEERS]       = { .type = NETLINK_TYPE_NESTED, .type_system = &genl_wireguard_peer_type_system },
 };
 
+/***************** genl NetLabel type systems *****************/
+static const NLType genl_netlabel_types[] = {
+        [NLBL_UNLABEL_A_IPV4ADDR] = { .type = NETLINK_TYPE_IN_ADDR },
+        [NLBL_UNLABEL_A_IPV4MASK] = { .type = NETLINK_TYPE_IN_ADDR },
```
yuwata: Not sure, but maybe, `.size = sizeof(struct in_addr)`??

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+        _cleanup_free_ char *addr_str = NULL;
+
+        assert(address);
+        assert(address->link);
+
+        if (!address->link->network)
+                return;
+
+        int r2;
+        r2 = in_addr_to_string(address->family, &address->in_addr, &addr_str);
```
yuwata: This is only used for logging. Please voidify the function and use `strna()` in the logger:

Also, `in_addr_prefix_to_string()` can be used here.

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+
+        r = netlabel_op(false, NULL, address);
+        if (r < 0)
+                log_warning_errno(r, "Deleting NetLabel for IP address %s/%d failed, ignoring: %m",
```
yuwata: Please use `log_link_warning_errno(address->link, r, "...")`

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+                                  r2 == 0? addr_str: "(error while translating to string)",
+                                  address->prefixlen);
+        log_debug("Deleted NetLabels for interface %s IP address %s/%d",
```
yuwata: Also here. Please use `log_link_debug()`, then `for interface %s` can be dropped.

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+        assert(address);
+        assert(address->link);
+
+        if (!address->link->network)
+                return;
+
+        if (address->family == AF_INET)
+                labels = address->link->network->ipv4_netlabel_unlbl_labels;
+        else if (address->family == AF_INET6)
+                labels = address->link->network->ipv4_netlabel_unlbl_labels;
```
yuwata: Typo: ipv4 -> ipv6

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+        else
+                return;
+
+        int r2;
+
+        r2 = in_addr_to_string(address->family, &address->in_addr, &addr_str);
```
yuwata: Same here. Please voidify and use `in_addr_prefix_to_string()`.

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+
+        const char *label;
+        SET_FOREACH(label, labels) {
+                r = netlabel_op(true, label, address);
+                if (r < 0) {
+                        log_warning_errno(r, "Adding NetLabel %s for IP address %s/%d failed, ignoring: %m",
```
yuwata: `log_link_warning_errno()`

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+                                          label,
+                                          r2 == 0? addr_str: "(error while translating to string)",
+                                          address->prefixlen);
+                }
+        }
+
+        log_debug("Added NetLabel %s for interface %s IP address %s/%d",
```
yuwata: `log_link_debug()`

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+        _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
+        int r;
+
+        assert(address);
+        assert(address->link);
+        assert(address->link->network);
+
+        r = sd_genl_socket_open(&genl);
+        if (r < 0)
+                return r;
```
yuwata: Please do not open new socket, but use `address->link->manager->genl`.

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+        if (add) {
+                r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_SECCTX, label);
+                if (r < 0)
+                        return r;
+        }
+
+        union in_addr_union mask;
+
+        memset(&mask, 0xff, sizeof(mask));
+        r = in_addr_mask(address->family, &mask, address->prefixlen);
```
yuwata: Why do you initialize the mask with `0xff` here?? What value should we pass to the kernel?

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
 }
 
+static int netlabel_op(bool add, const char *label, const Address *address) {
+        _cleanup_(sd_netlink_unrefp) sd_netlink *genl = NULL;
+        _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
+        int r;
+
+        assert(address);
+        assert(address->link);
+        assert(address->link->network);
```
yuwata: Maybe, it is better to add `assert(IN_SET(address->family, AF_INET, AF_INET6));`

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6ADDR, &address->in_addr.in6);
+                if (r < 0)
+                        return r;
+
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6MASK, &mask.in6);
+        }
+        if (r < 0)
+                return r;
+
+        r = sd_netlink_send(genl, m, NULL);
```
yuwata: Please send the message asynchronously.

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+        return r < 0 ? r : 0;
```
yuwata: We usually do not distinguish non-negative values. Hence. `returns sd_netlink_send()` is OK.

src/network/networkd-network.c
```diff
@@ -1161,6 +1164,95 @@ int config_parse_ntp(
...
+                Set **set) {
+        int r;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+        assert(set);
+
+        if (isempty(rvalue)) {
+                set_free(*set);
```
yuwata: `*set = set_free(*set);` Otherwise, this triggers use-after-free.

src/network/networkd-network.c
```diff
@@ -1161,6 +1164,95 @@ int config_parse_ntp(
...
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        log_syntax(unit, LOG_WARNING, filename, line, r,
+                                   "Bad NetLabel label, ignoring: %s", rvalue);
+                        return 0;
+                }
+
+                r = set_ensure_put(set, &string_hash_ops, TAKE_PTR(w));
```
yuwata: `set_ensure_consume()`, otherwise `w` is leaked on failure.

yuwata (changes requested):
Instead of introducing `IPv4NetLabelUnlabel=`and `IPv6NetLabelUnlabel=` in `[Network]` section, please add `NetLabel=` or so in `[Address]`, `[DHCPv4]`, `[DHCPv6]`, `[IPv6AcceptRA]`, `[DHCPPrefixDelegation]` sections.
Then, we can assign different labels to each address.

Also, please move the netlabel related functions to networkd-netlabel.[ch] or so, as networkd-address.c and networkd-network.c are already too big.

src/libsystemd/sd-netlink/netlink-types-genl.c
topimiettinen: OK. Grepping for other uses of `NETLINK_TYPE_IN_ADDR` shows that the `.size` isn't used very consistently, even for IPv6:
src/libsystemd/sd-netlink/netlink-types-genl.c:119:        [FOU_ATTR_LOCAL_V4]          = { .type = NETLINK_TYPE_IN_ADDR },
src/libsystemd/sd-netlink/netlink-types-genl.c:121:        [FOU_ATTR_LOCAL_V6]          = { .type = NETLINK_TYPE_IN_ADDR },
src/libsystemd/sd-netlink/netlink-types-genl.c:149:        [L2TP_ATTR_IP_DADDR]          = { .type = NETLINK_TYPE_IN_ADDR },
src/libsystemd/sd-netlink/netlink-types-genl.c:153:        [L2TP_ATTR_IP6_DADDR]         = { .type = NETLINK_TYPE_IN_ADDR },

Maybe they should be fixed too?

src/network/networkd-address.c
topimiettinen: I think the mask should have bits set according to prefixlen, like 255.255.0.0 for prefix 16 in IPv4. Perhaps it would be better instead to add a new function `in_addr_prefixlen_to_netmask()`, using existing `in4_addr_prefixlen_to_netmask()` and (not yet existing) `in6_addr_prefixlen_to_netmask()`.

src/libsystemd/sd-netlink/netlink-types-genl.c
yuwata: Yeah, please if you'd like.

src/network/networkd-address.c
topimiettinen: I implemented asynchronous sending, but looking at the logs I'm not sure it's a good idea. I think at least for address configuration the operation should be synchronous, so that the label is used as soon as possible or at least before the address is configured. For deconfiguration doing label removal asynchronously probably would not be critical by itself, but then the removal operation could be mixed with a new address configuration event and the order of operations could get wrong. This is also made worse by a property of NetLabel operation for removal: it isn't symmetric with addition, the label can't be specified but all labels for the address are removed.

src/libsystemd/sd-netlink/netlink-types-genl.c
topimiettinen: Made PR #23302.

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+
+        if (add) {
+                r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_SECCTX, label);
+                if (r < 0)
+                        return r;
+        }
+
+        union in_addr_union mask;
+
+        memset(&mask, 0xff, sizeof(mask));
```
poettering: sounds like we should have IN_ADDR_ALLF or so, to match IN_ADDR_NULL we already have in in-addr-util.h

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+                        return r;
+
+                r = sd_netlink_message_append_in_addr(m, NLBL_UNLABEL_A_IPV4MASK, &mask.in);
+        } else if (address->family == AF_INET6) {
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6ADDR, &address->in_addr.in6);
+                if (r < 0)
+                        return r;
+
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6MASK, &mask.in6);
+        }
```
poettering: have an else branch that returns -EAFNOSUPPORT? or hits an assert?

src/network/networkd-address.c
poettering: order is one think, doing stuff synchronous is another. just make sure your calls here are enqueued before the netlink calls to add the addresses?

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+
+        r2 = in_addr_to_string(address->family, &address->in_addr, &addr_str);
+
+        const char *label;
+        SET_FOREACH(label, labels) {
+                r = netlabel_op(true, label, address);
+                if (r < 0) {
+                        log_warning_errno(r, "Adding NetLabel %s for IP address %s/%d failed, ignoring: %m",
+                                          label,
+                                          r2 == 0? addr_str: "(error while translating to string)",
```
poettering: instead of checking r2, just check if addr_str is non-NULL?

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
...
+                                          address->prefixlen);
+                }
+        }
+
+        log_debug("Added NetLabel %s for interface %s IP address %s/%d",
+                  label, address->link->ifname,
+                  r2 == 0? addr_str: "(error while translating to string)", address->prefixlen);
```
poettering: ditto

src/network/networkd-address.c
```diff
@@ -448,6 +448,127 @@ static int address_set_masquerade(Address *address, bool add) {
         return 0;
 }
 
+static int netlabel_op(bool add, const char *label, const Address *address) {
```
poettering: instead of passing a bool as first arg, maybe just pass `uint16_t command` as first argument?

that makes the invocations more self-explanatory (since you imemdiately see if this is about add or about remove, instead of a cryptic "true" or "false), and shortens the code.

topimiettinen: Nice!

src/network/networkd-address.c
topimiettinen: Ok, kept the asynchronous version then.

src/network/networkd-address.c
topimiettinen: Introduced instead `in_addr_prefixlen_to_netmask()`.

src/basic/in-addr-util.c
```diff
@@ -595,6 +595,57 @@ struct in_addr* in4_addr_prefixlen_to_netmask(struct in_addr *addr, unsigned cha
...
+}
+
+int in_addr_prefixlen_to_netmask(int family, union in_addr_union *addr, unsigned char prefixlen) {
+        assert(addr);
+
+        switch (family) {
+        case AF_INET: {
+                struct in_addr *ret_addr;
+
+                ret_addr = in4_addr_prefixlen_to_netmask(&addr->in, prefixlen);
```
yuwata: It does not fail. Hence, simply,
```
switch (family) {
case AF_INET:
        in4_addr_prefixlen_to_netmask(...);
        return 0;
case AF_INET6:
        in6_addr_prefixlen_to_netmask(...);
        return 0;
default:
        return -EAFNOSUPPORT;
}
```

src/basic/in-addr-util.c
```diff
@@ -595,6 +595,57 @@ struct in_addr* in4_addr_prefixlen_to_netmask(struct in_addr *addr, unsigned cha
...
+                        mask = 0;
+                }
+
+                addr->s6_addr[i] = mask;
+        }
+
+        return addr;
+}
+
+int in_addr_prefixlen_to_netmask(int family, union in_addr_union *addr, unsigned char prefixlen) {
```
yuwata, line 622: Please add tests for this.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+
+        assert_se(rtnl);
+        assert_se(m);
+        assert_se(userdata);
+
+        Link *link = userdata;
+
+        r = sd_netlink_message_get_errno(m);
+        if (r < 0) {
+                log_link_warning_errno(link, r, "NetLabel operation failed: %m");
```
yuwata: Please use `log_link_message_warning_errno()`.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+        r = sd_genl_message_new(address->link->manager->genl, NETLBL_NLTYPE_UNLABELED_NAME, command, &m);
+        if (r < 0)
+                return r;
+
+        r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_IFACE, address->link->ifname);
+        if (r < 0)
+                return r;
+
+        if (command == NLBL_UNLABEL_C_STATICADD) {
+                r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_SECCTX, label);
```
yuwata, line 49: Please add `assert(label);` in this block.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6ADDR, &address->in_addr.in6);
+                if (r < 0)
+                        return r;
+
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6MASK, &netmask.in6);
+        }
+        if (r < 0)
+                return r;
+
+        return sd_netlink_call_async(address->link->manager->genl, NULL, m, netlabel_handler, NULL, address->link, 0, NULL);
```
yuwata: Please use `netlink_call_async()` (that is, without 'sd_' prefix), and if this succeeds, please increment the reference counter of `address->link`.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+        int r;
+        const char *label;
+
+        (void) in_addr_prefix_to_string(address->family, &address->in_addr, address->prefixlen, &addr_str);
+
+        SET_FOREACH(label, labels) {
+                r = netlabel_command(NLBL_UNLABEL_C_STATICADD, label, address);
+                if (r < 0)
+                        log_link_warning_errno(address->link, r, "Adding NetLabel %s for IP address %s failed, ignoring: %m",
+                                               label, strna(addr_str));
```
yuwata, line 96: If this is failed, `log_link_debug()` below is also called, please add `else` or `continue`.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+                log_link_debug(address->link, "Added NetLabel %s for IP address %s",
```
yuwata: Not a big matter, but the label is not added yet. "Added" -> "Adding".

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+        assert(address);
+        assert(address->link);
+
+        if (!address->link->network || !IN_SET(address->family, AF_INET, AF_INET6))
+                return;
+
+        if (address->family == AF_INET && IN_SET(address->source, NETWORK_CONFIG_SOURCE_DHCP4, NETWORK_CONFIG_SOURCE_DHCP_PD) && address->link->network->dhcp4_netlabels)
+                address_add_netlabel_set(address, address->link->network->dhcp4_netlabels);
+        else if (address->family == AF_INET6 && IN_SET(address->source, NETWORK_CONFIG_SOURCE_DHCP6, NETWORK_CONFIG_SOURCE_DHCP_PD) && address->link->network->dhcp6_netlabels)
+                address_add_netlabel_set(address, address->link->network->dhcp6_netlabels);
```
yuwata: I think it is better to split label for DHCP_PD from DHCP4 or DHCP6.
```
switch (address->source) {
case NETWORK_CONFIG_SOURCE_STATIC:
        return address_add_netlabel_set(address, address->netlabels);
case NETWORK_CONFIG_SOURCE_DHCP4:
        return address_add_netlabel_set(address, address->link->network->dhcp4_netlabels);
...
}
```

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+                return;
+
+        (void) in_addr_prefix_to_string(address->family, &address->in_addr, address->prefixlen, &addr_str);
+
+        r = netlabel_command(NLBL_UNLABEL_C_STATICREMOVE, NULL, address);
+        if (r < 0)
+                log_link_warning_errno(address->link, r, "Deleting NetLabels for IP address %s failed, ignoring: %m",
+                                       strna(addr_str));
+        log_link_debug(address->link, "Deleted NetLabels for IP address %s",
+                       strna(addr_str));
```
yuwata: Same here, the debugging log is added even `netlabel_command()` fails.

Also, `Deleted` -> `Deleting` or `Removing`

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+                        return 0;
+                }
+                if (r == 0)
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        log_syntax(unit, LOG_WARNING, filename, line, r,
+                                   "Bad NetLabel label, ignoring: %s", rvalue);
+                        return 0;
```
yuwata: How about to ignore `w` but continue to parse?
```
if (!string_is_safe(w)) {
        log_syntax(..., w);
        continue;
}
```

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+                const char *unit,
+                const char *filename,
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                Set **set) {
```
yuwata: Please use `data` to accept `set`. Then, this become a normal conf parser.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,213 @@
...
+                        return 0;
+                }
+
+                r = set_ensure_consume(set, &string_hash_ops, TAKE_PTR(w));
+                if (r < 0)
+                        return log_oom();
+        }
+}
+
+int config_parse_dhcp_or_ra_netlabel(
```
yuwata: Then, this can be dropped.

src/network/networkd-network-gperf.gperf
```diff
@@ -243,6 +245,7 @@ DHCPv4.SendVendorOption,                     config_parse_dhcp_send_option,
 DHCPv4.RouteMTUBytes,                        config_parse_mtu,                                         AF_INET,                       offsetof(Network, dhcp_route_mtu)
 DHCPv4.FallbackLeaseLifetimeSec,             config_parse_dhcp_fallback_lease_lifetime,                0,                             0
 DHCPv4.Use6RD,                               config_parse_bool,                                        0,                             offsetof(Network, dhcp_use_6rd)
+DHCPv4.NetLabel,                             config_parse_dhcp_or_ra_netlabel,                         AF_INET,                       0
```
yuwata: `AF_INET` -> `0`
`0` -> `offsetof(Network, dhcp_netlabel)`

src/network/networkd-network-gperf.gperf
```diff
@@ -260,6 +263,7 @@ DHCPv6.SendOption,                           config_parse_dhcp_send_option,
 DHCPv6.IAID,                                 config_parse_iaid,                                        AF_INET6,                      0
 DHCPv6.DUIDType,                             config_parse_duid_type,                                   0,                             offsetof(Network, dhcp6_duid)
 DHCPv6.DUIDRawData,                          config_parse_duid_rawdata,                                0,                             offsetof(Network, dhcp6_duid)
+DHCPv6.NetLabel,                             config_parse_dhcp_or_ra_netlabel,                         AF_INET6,                      0
```
yuwata: `offsetof(Network, dhcp6_netlabel)`

src/network/networkd-network-gperf.gperf
```diff
@@ -277,6 +281,7 @@ IPv6AcceptRA.PrefixDenyList,                 config_parse_in_addr_prefixes,
 IPv6AcceptRA.RouteAllowList,                 config_parse_in_addr_prefixes,                            AF_INET6,                      offsetof(Network, ndisc_allow_listed_route_prefix)
 IPv6AcceptRA.RouteDenyList,                  config_parse_in_addr_prefixes,                            AF_INET6,                      offsetof(Network, ndisc_deny_listed_route_prefix)
 IPv6AcceptRA.Token,                          config_parse_address_generation_type,                     0,                             offsetof(Network, ndisc_tokens)
+IPv6AcceptRA.NetLabel,                       config_parse_dhcp_or_ra_netlabel,                         AF_INET6,                      0
```
yuwata: `offsetof(Network, ndisc_netlabel)`

src/network/networkd-network-gperf.gperf
```diff
@@ -343,6 +348,7 @@ DHCPPrefixDelegation.Assign,                 config_parse_bool,
 DHCPPrefixDelegation.ManageTemporaryAddress, config_parse_bool,                                        0,                             offsetof(Network, dhcp_pd_manage_temporary_address)
 DHCPPrefixDelegation.Token,                  config_parse_address_generation_type,                     0,                             offsetof(Network, dhcp_pd_tokens)
 DHCPPrefixDelegation.RouteMetric,            config_parse_uint32,                                      0,                             offsetof(Network, dhcp_pd_route_metric)
+DHCPPrefixDelegation.NetLabel,               config_parse_dhcp_or_ra_netlabel,                         AF_INET,                       0
```
yuwata: `offsetof(Network, dhcp_pd_netlabel)`

src/network/networkd-network.c
```diff
@@ -705,6 +705,9 @@ static Network *network_free(Network *network) {
         set_free(network->ndisc_deny_listed_route_prefix);
         set_free(network->ndisc_allow_listed_route_prefix);
 
+        set_free(network->dhcp4_netlabels);
+        set_free(network->dhcp6_netlabels);
```
yuwata: Please put them together with dhcp4 or dhcp6 related parameters.

src/network/networkd-network.h
```diff
@@ -353,6 +353,9 @@ struct Network {
 
         /* NTP */
         char **ntp;
+
+        /* NetLabel */
+        Set *dhcp4_netlabels, *dhcp6_netlabels;
```
yuwata: Also here. Please move them.

yuwata (changes requested):

topimiettinen:
Updated according to comments, please review.

mrc0mmand:
> mkosi f35: `sh: line 1: getfacl: command not found`, not related

This should be resolved by #23365.

> CentOS CI (Arch Linux + sanitizers) shows a lot of memory leaks in networkd, which could be related. @mrc0mmand WDYT?

They're definitely related, since they don't appear in other PRs.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,190 @@
...
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        log_syntax(unit, LOG_WARNING, filename, line, r,
+                                   "Bad NetLabel label, ignoring: %s", w);
+                        continue;
+                }
+
+                r = set_ensure_consume(set, &string_hash_ops, TAKE_PTR(w));
```
mrc0mmand: I wonder if this shouldn't be `&string_hash_ops_free` (and probably the `*set = set_free(*set);` line above shouldn't be `*set = set_free_free(*set);`). Based on a quick skim though the existing code (and a criminal lack of coffee).

topimiettinen: Excellent, perhaps coffee isn't required for proper operation?

src/network/networkd-network.c
```diff
@@ -687,6 +687,8 @@ static Network *network_free(Network *network) {
         free(network->dhcp6_mudurl);
         strv_free(network->dhcp6_user_class);
         strv_free(network->dhcp6_vendor_class);
+        set_free(network->dhcp_netlabels);
```
mrc0mmand, line 691: Ah, following the previous change, maybe this should be `set_free_free(...)` as well in this and the following three cases?

topimiettinen: Thanks, nice catch!

topimiettinen:
Hmm, still fails because of memory leaks.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,190 @@
...
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6ADDR, &address->in_addr.in6);
+                if (r < 0)
+                        return r;
+
+                r = sd_netlink_message_append_in6_addr(m, NLBL_UNLABEL_A_IPV6MASK, &netmask.in6);
+        }
+        if (r < 0)
+                return r;
+
+        r = netlink_call_async(address->link->manager->genl, NULL, m, netlabel_handler, NULL, address->link);
```
yuwata: Please specify the destroy callback, otherwise, the reference counter of `address->link` will never be decremented.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,190 @@
...
+        assert_se(m);
+        assert_se(link);
+
+        r = sd_netlink_message_get_errno(m);
+        if (r < 0) {
+                log_link_message_warning_errno(link, m, r, "NetLabel operation failed, ignoring");
+                return 1;
+        }
+
+        log_link_message_debug_errno(link, m, r, "NetLabel operation successful");
```
yuwata: Here, please use `log_link_debug()`. `log_link_message_debug_errno()` append error message in netlink message.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,190 @@
...
+
+        switch (address->source) {
+        case NETWORK_CONFIG_SOURCE_DHCP4:
+                return address_add_netlabel_set(address, address->link->network->dhcp_netlabels);
+        case NETWORK_CONFIG_SOURCE_DHCP6:
+                return address_add_netlabel_set(address, address->link->network->dhcp6_netlabels);
+        case NETWORK_CONFIG_SOURCE_DHCP_PD:
+                return address_add_netlabel_set(address, address->link->network->dhcp_pd_netlabels);
+        case NETWORK_CONFIG_SOURCE_STATIC:
+                return address_add_netlabel_set(address, address->netlabels);
```
yuwata, line 119: SOURCE_NDISC is missing.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,190 @@
...
+        int r;
+        Set **set = data;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+        assert(set);
+
+        if (isempty(rvalue)) {
+                *set = set_free_free(*set);
```
yuwata: Please use `set_free()`, as hash_op has free func, i.e., you specified string_hash_ops_free.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,190 @@
...
+                        log_syntax(unit, LOG_WARNING, filename, line, r,
+                                   "Failed to extract NetLabel label, ignoring: %s", rvalue);
+                        return 0;
+                }
+                if (r == 0)
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        log_syntax(unit, LOG_WARNING, filename, line, r,
```
yuwata: `r` -> `0`

src/network/networkd-network.c
```diff
@@ -687,6 +687,8 @@ static Network *network_free(Network *network) {
         free(network->dhcp6_mudurl);
         strv_free(network->dhcp6_user_class);
         strv_free(network->dhcp6_vendor_class);
+        set_free_free(network->dhcp_netlabels);
+        set_free_free(network->dhcp6_netlabels);
```
yuwata: set_free

src/network/networkd-network.c
```diff
@@ -753,6 +755,8 @@ static Network *network_free(Network *network) {
         ordered_hashmap_free(network->dhcp6_client_send_vendor_options);
         set_free(network->dhcp_pd_tokens);
         set_free(network->ndisc_tokens);
+        set_free_free(network->dhcp_pd_netlabels);
+        set_free_free(network->ndisc_netlabels);
```
yuwata: set_free

src/network/networkd-address.c
```diff
@@ -137,6 +138,7 @@ Address *address_free(Address *address) {
 
         config_section_free(address->section);
         free(address->label);
+        set_free_free(address->netlabels);
```
yuwata: set_free

yuwata (changes requested):

topimiettinen:
Thanks for the review, updated.

topimiettinen:
All green!

evverx:
FWIW to avoid scenarios like this it should be possible to somewhat enforce approvals with https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule
> Optionally, to require approvals before a pull request can be merged, select Require approvals, click the Required number of approvals before merging drop-down menu, then select the number of approving reviews you would like to require on the branch.

It's probably bureaucracy but on the other hand it appears a lot of PRs have been approved like that anyway so it seems it shouldn't be too annoying.

yuwata:
> But it was reviewed by @yuwata.

I reviewed several times, but I have neither said 'LGTM' nor set the green label...

Sorry for late final review, but this needs to be re-reviewed. Temporary setting 'merged-but-needs-fixing'.

man/systemd.network.xml
```diff
@@ -1109,6 +1109,38 @@ Table=1234</programlisting></para>
...
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname><replaceable>label</replaceable></term>
+        <listitem>
+
+          <para>This setting provides a method for integrating dynamic network configuration into Linux
+          NetLabel subsystem rules, used by Linux security modules (LSMs) for network access control. The
+          option expects a whitespace separated list of NetLabel labels. The labels must conform to lexical
+          restrictions of LSM labels. When an interface is configured with IP addresses, the addresses and
+          subnetwork masks will be appended to the NetLabel Fallback Peer Labeling rules. They will be
```
keszybz, line 1121: It'd be good to add some hyperrefences here, e.g. to kernel docs for LSM and NetLabel.

man/systemd.network.xml
```diff
@@ -2050,6 +2082,13 @@ Table=1234</programlisting></para>
           <ulink url="https://tools.ietf.org/html/rfc5227">RFC 5227</ulink>. Defaults to false.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
```
keszybz, line 2089: "This setting accepts the same arguments as <varname>NetLabel=</varname> in the [Address] section."

man/systemd.network.xml
```diff
@@ -2264,6 +2304,13 @@ Table=1234</programlisting></para>
           </para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
```
keszybz, line 2311: Here too.

man/systemd.network.xml
```diff
@@ -2521,6 +2568,13 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
           specified. Defaults to true.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
```
keszybz, line 2575: Here too.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,194 @@
...
+static int netlabel_handler(sd_netlink *rtnl, sd_netlink_message *m, Link *link) {
+        int r;
+
+        assert_se(rtnl);
+        assert_se(m);
+        assert_se(link);
+
+        r = sd_netlink_message_get_errno(m);
+        if (r < 0) {
+                log_link_message_warning_errno(link, m, r, "NetLabel operation failed, ignoring");
```
keszybz, line 19: %m missing.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,194 @@
...
+                return 1;
+        }
+
+        log_link_debug(link, "NetLabel operation successful");
```
keszybz, line 23: Hmm, this message is not very informative. If I saw it without context, I would have no idea what was being done and why. Can you rephrase it?

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,194 @@
...
+                                   "Failed to extract NetLabel label, ignoring: %s", rvalue);
+                        return 0;
+                }
+                if (r == 0)
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        log_syntax(unit, LOG_WARNING, filename, line, 0,
+                                   "Bad NetLabel label, ignoring: %s", w);
```
keszybz, line 186: If `w` is not safe, it shouldn't be printed. Please `cescape` it before pritning.

src/network/networkd-netlabel.c
keszybz: %m is missing.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,194 @@
...
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
+        int r;
+        Set **set = data;
```
keszybz, line 157:
```C
Set **set = ASSERT_PTR(data);
int r;
```

src/network/networkd-netlabel.c
yuwata: `log_link_message_warning_errno()` macro automatically adds `%m` at the end.
Please see `log_link_message_full_errno()`.

man/systemd.network.xml
topimiettinen: I used the same short text 'As in ...' like is used elsewhere in this manual page (`MUDURL=`, `UseDNS=`, `MACAddress=` etc). Wouldn't it be more consistent to use similar style also here so that it can be sort of skipped by the reader more easily?

src/network/networkd-netlabel.c
topimiettinen: I agree it would be nice to have more verbose message explaining what exactly was successful (or what failed). But there isn't any context related to the operation available in the callback, just `Link` (which may have several `Address`) and the raw message.

In the earlier version I used synchronous sending and then the errors messages were able to be more precise.

topimiettinen:
Created https://github.com/systemd/systemd/pull/23712 for the fixes.

PR #23272, opened 2022-05-05
Logind man and rules

keszybz:


bluca:
packit is sad:

```
+ /usr/bin/patch -p1 -s --fuzz=0 --no-backup-if-mismatch -f
1 out of 1 hunk FAILED -- saving rejects to file rules.d/meson.build.rej
error: Bad exit status from /var/tmp/rpm-tmp.94jfml (%prep)
    Bad exit status from /var/tmp/rpm-tmp.94jfml (%prep)
```

mrc0mmand:
Looks like it conflicts with https://src.fedoraproject.org/rpms/systemd/blob/rawhide/f/use-bfq-scheduler.patch, so at some point this needs to be merged with failing Packit and then the patch needs to be regenerated in the Fedora dist-git.

yuwata:
LGTM. Maybe we should drop all patches for fedora when building for packit.

PR #23275, opened 2022-05-05
journal: Make journal_default_metrics() static

daandemeyer:
Only caller resides in the same file as were it's defined so no
need to have it be public.


bluca (approved):

keszybz:
Since this is just moving code around, if it compiles in one place, we should be fine.

PR #23286, opened 2022-05-06
basic: Propagate SIGBUS signal info when re-raising signals

daandemeyer:
raise() and kill() won't propagate the siginfo information of the
signal that's re-raised. rt_sigqueueinfo() allows us to provide
the original siginfo struct which makes sure it is propagated to
the next signal handler (or to the coredump).

### DIFF
--- a/src/basic/sigbus.c
+++ b/src/basic/sigbus.c
@@ -7,6 +7,8 @@
 
 #include "macro.h"
 #include "memory-util.h"
+#include "missing_syscall.h"
+#include "process-util.h"
 #include "sigbus.h"
 
 #define SIGBUS_QUEUE_MAX 64
@@ -88,7 +90,7 @@ static void sigbus_handler(int sn, siginfo_t *si, void *data) {
 
         if (si->si_code != BUS_ADRERR || !si->si_addr) {
                 assert_se(sigaction(SIGBUS, &old_sigaction, NULL) == 0);
-                raise(SIGBUS);
+                rt_sigqueueinfo(getpid_cached(), SIGBUS, si);
                 return;
         }
 


src/basic/sigbus.c
```diff
@@ -88,7 +90,7 @@ static void sigbus_handler(int sn, siginfo_t *si, void *data) {
 
         if (si->si_code != BUS_ADRERR || !si->si_addr) {
                 assert_se(sigaction(SIGBUS, &old_sigaction, NULL) == 0);
-                raise(SIGBUS);
+                rt_sigqueueinfo(getpid_cached(), SIGBUS, si);
```
anitazha, line 93: You should probably add some assertions for `si`. Things like
- `si->si_signo == SIGBUS`
- `si->si_code != SI_TKILL`

The man page seems to suggest that there are restrictions on the `siginfo_t` set in `rt_sigqueueinfo()` that could be different from what the kernel is allowed to set elsewhere.

daandemeyer: Those restrictions in the manpage are conditionalized on the following though:

> If the signal is being sent to any process other than the caller itself, the following restrictions apply ...

So that made me think we don't need to check for those restrictions here since we're sending the signal to ourselves.

anitazha: Ah gotcha that makes sense.

So I guess the other thing I'm not sure about is, `rt_sigqueueinfo()` could possibly hit the queue limit for real-time signals. Compared to `raise()` which sends a standard signal that doesn't queue. Does this matter?

daandemeyer: From reading the man pages, the way I understand it, Only real-time signals can be queued. Standard signals cannot be queued. Real-time signals are signals numbered from 32-63. The actual function used to send the signal doesn't affect whether it's a standard signal or a real-time signal. So since we're sending a standard signal using rt_sigqueueinfo(), it won't be queued at all just like before when we sent it with raise().

anitazha (approved):

PR #23292, opened 2022-05-06
network: clarify relationship between RA flags and DHCPv6 modes

alexhenrie:
In the documentation, using the term "managed" for both the RA flag and
the DHCPv6 mode is confusing because the mode is referred to as
"solicit" both in the official DHCPv6 documentation (see RFC 8415) and
in the WithoutRA option.

Furthermore, calling the other RA flag "other information" or "other
address configuration" is confusing because its official name is simply
"other configuration" (see RFC 4861 and RFC 5175) and it isn't used to
assign IP addresses.

Rewrite the documentation for DHCPv6Client and WithoutRA to make it
clear that getting the "managed" RA flag triggers the same kind of DHCP
request as WithoutRA=solicit, whereas getting the "other configuration"
RA flag triggers the same kind of DHCP request as
WithoutRA=information-request.

### DIFF
--- a/man/systemd.network.xml
+++ b/man/systemd.network.xml
@@ -328,9 +328,10 @@
           <literal>no</literal>.</para>
 
           <para>Note that DHCPv6 will by default be triggered by Router Advertisement, if that is
-          enabled, regardless of this parameter. By enabling DHCPv6 support explicitly, the DHCPv6
-          client will be started regardless of the presence of routers on the link, or what flags the
-          routers pass. See <literal>IPv6AcceptRA=</literal>.</para>
+          enabled, regardless of this parameter. By explicitly enabling DHCPv6 support here, the
+          DHCPv6 client will be started in the mode specified by the <variable>WithoutRA=</variable>
+          setting in the [DHCPv6] section, regardless of the presence of routers on the link, or
+          what flags the routers pass. See <literal>IPv6AcceptRA=</literal>.</para>
 
           <para>Furthermore, note that by default the domain name specified through DHCP is not used
           for name resolution. See option <option>UseDomains=</option> below.</para>


yuwata (approved):
LGTM. Thank you.

man/systemd.network.xml
```diff
@@ -2172,9 +2172,9 @@ Table=1234</programlisting></para>
       <varlistentry>
         <term><varname>WithoutRA=</varname></term>
         <listitem>
-          <para>Allows DHCPv6 client to start without router advertisements's managed or other
-          address configuration flag. Takes one of <literal>no</literal>, <literal>solicit</literal>
-          or <literal>information-request</literal>. If this is not specified,
+          <para>Allows DHCPv6 client to start without router advertisements's "managed" or "other
```
yuwata: Please use `<literal>` tag instead of `"`.

man/systemd.network.xml
```diff
@@ -2172,9 +2172,9 @@ Table=1234</programlisting></para>
...
+          configuration" flag. Takes one of <literal>no</literal>, <literal>solicit</literal> or
```
yuwata: This is not your fault, but could you add a comma after `<literal>solicit</literal>`?

man/systemd.network.xml
alexhenrie: Done.

man/systemd.network.xml
alexhenrie: Done.

alexhenrie:
While we're at it, I just added another small commit to clarify the relationship between `DHCP=` and `WithoutRA=`. The previous documentation implied that all you have to do to get DHCPv6 working without RA is to set `DHCP=true` or `DHCP=ipv6`, but in reality `WithoutRA` must also be set to the desired DHCP mode.

alexhenrie:
Sorry for the noise, I just noticed that I missed a quotation mark. They are all `<literal>` tags now.

And thanks for the speedy review!

alexhenrie:
Thank you!

PR #23293, opened 2022-05-06
shared/json: disentangle flag mixup

keszybz:
_DEBUG and _RELAX had the same value… I'm not sure what the effect
of this is.

### DIFF
--- a/src/shared/json.h
+++ b/src/shared/json.h
@@ -331,9 +331,9 @@ typedef enum JsonDispatchFlags {
         JSON_SAFE       = 1 << 3, /* Don't accept "unsafe" strings in json_dispatch_string() + json_dispatch_string() */
         JSON_RELAX      = 1 << 4, /* Use relaxed user name checking in json_dispatch_user_group_name */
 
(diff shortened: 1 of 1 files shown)


src/shared/json.h
```diff
@@ -332,8 +332,8 @@ typedef enum JsonDispatchFlags {
         JSON_RELAX      = 1 << 4, /* Use relaxed user name checking in json_dispatch_user_group_name */
 
         /* The following two may be passed into log_json() in addition to the three above */
-        JSON_DEBUG      = 1 << 4, /* Indicates that this log message is a debug message */
-        JSON_WARNING    = 1 << 5, /* Indicates that this log message is a warning message */
+        JSON_DEBUG      = 1 << 5, /* Indicates that this log message is a debug message */  //
```
bluca: what's that `//` at the end?

keszybz: It was my mark to mark that this needs fixing, which of course I forgot to remove ;)

bluca (approved):

PR #23297, opened 2022-05-07
meson: Compile with -ftrivial-auto-var-init

medhefgo:


medhefgo:
I can't say if this would have a significant runtime effect, but it should certainly improve security a bit. Also, a quick check with compiler explorer suggests that gcc is intelligent enough to elide the write if it can prove all code paths are initializing the var.

If desired, we could also only enable it for hardened (and debug) builds…

meson.build
```diff
@@ -401,6 +401,12 @@ if cc.get_id() == 'clang'
         ]
 endif
 
+if get_option('mode') == 'developer'
+        possible_common_cc_flags += '-ftrivial-auto-var-init=pattern'
```
evverx: Looks like it messes up `MSan` and `valgrind`. Without it they can detect an unitialized variable:
```sh
$ clang -fsanitize=memory  main.c
$ ./a.out
==50642==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x49d724 in main (/home/vagrant/systemd/a.out+0x49d724)
    #1 0x7f234c472f1f in __libc_start_call_main (/lib64/libc.so.6+0x40f1f)
    #2 0x7f234c472fcf in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x40fcf)
    #3 0x41c314 in _start (/home/vagrant/systemd/a.out+0x41c314)

SUMMARY: MemorySanitizer: use-of-uninitialized-value (/home/vagrant/systemd/a.out+0x49d724) in main
Exiting

$ clang  main.c
$ DEBUGINFOD_URLS= valgrind ./a.out
==50651== Memcheck, a memory error detector
==50651== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==50651== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==50651== Command: ./a.out
==50651==
==50651== Conditional jump or move depends on uninitialised value(s)
==50651==    at 0x40111F: main (in /home/vagrant/systemd/a.out)
```
With `-ftrivial-auto-var-init=pattern` they miss it completely
```sh
$ clang -fsanitize=memory -ftrivial-auto-var-init=pattern main.c
$ ./a.out

$ clang -ftrivial-auto-var-init=pattern main.c
$ DEBUGINFOD_URLS= valgrind ./a.out
...
==50675== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

medhefgo: I guess it's working as intended? :D

I changed the logic a little. I hope that's enough, short of adding another meson build option.

evverx: > I guess it's working as intended? :D

If it had crashed it would have worked. My understanding is that `-ftrivial-auto-var-init=pattern` can't be used to detect unitialized variables reliably so I don't think it should be used when `systemd` is built with MSan or run under Valgrind.

evverx:
FWIW I don't think flags like that should be passed by default because it would most likely mess up "hardened" builds where stuff like that is added via `CFLAGS/LDFLAGS` when all packages are built.

meson.build
```diff
@@ -401,6 +401,12 @@ if cc.get_id() == 'clang'
         ]
 endif
 
+if get_option('mode') == 'release'
+        possible_common_cc_flags += '-ftrivial-auto-var-init=zero'
+elif not get_option('valgrind') and get_option('b_sanitize') == 'none'
```
evverx: I have to admit I'm not sure how to detect whether the sanitizers are used or not. `b_sanitize` is kind of broken (https://github.com/mesonbuild/meson/pull/9825) so it isn't always used to pass `-fsanitize=`. On OSS-Fuzz everything is passed via `CFLAGS/LDFLAGS`.

`valgrind` should be easy because `-Dvalgrind=true` is supposed to be passed to make `systemd` compatible with it.

yuwata:
CIs are unhappy.

meson.build
medhefgo: It's "working as intended" because the flag makes the compiler initialize those vars, removing the undefined behavior altogether.

evverx: The idea behind `-ftrivial-auto-var-init=pattern` as far as I understand is to initialize variables using values that most likely trigger crashes down the road. compilers aren't smart enough to inject values that trigger crashes all the time.

evverx: Anyway, I get what you're saying. Sorry.

medhefgo: Well, it depends on the type of var. If it was a pointer, it should have crashed.

medhefgo:
Maybe this should just be for release builds as defensive measure and leave the dev builds as is`

meson.build
evverx: Agreed. I think it there had been a pointer there it had crashed.

FWIW With clang it appears `-ftrivial-auto-var-init=pattern` isn't always usable: https://reviews.llvm.org/D68115. That probably can explain why the tests where unions were initialized explicitly with `{}` triggered the assertions.

medhefgo:
I could make a rather complicated check for valgrind/sanatizers/clang and only then enable pattern. But it sounds a little brittle at that point… so I've opted for just release build zeroing. Unless someone really wants it. :D

keszybz:
Do we have any idea what the impact is on performance?

keszybz:
```
In file included from ../src/basic/siphash24.h:11,
                 from ../src/basic/hash-funcs.h:6,
                 from ../src/basic/hashmap.h:8,
                 from ../src/shared/fdset.h:6,
                 from ../src/shared/bpf-program.h:9,
                 from ../src/core/unit.h:11,
                 from ../src/core/all-units.h:4,
                 from ../src/core/manager.c:23:
../src/basic/time-util.h: In function â€˜manager_dispatch_jobs_in_progressâ€™:
../src/basic/time-util.h:140:38: error: â€˜xâ€™ may be used uninitialized [-Werror=maybe-uninitialized]
  140 | #define FORMAT_TIMESPAN(t, accuracy) format_timespan((char[FORMAT_TIMESPAN_MAX]){}, FORMAT_TIMESPAN_MAX, t, accuracy)
      |                                      ^~~~~~~~~~~~~~~
In function â€˜manager_print_jobs_in_progressâ€™,
    inlined from â€˜manager_dispatch_jobs_in_progressâ€™ at ../src/core/manager.c:3007:9:
../src/core/manager.c:219:18: note: â€˜xâ€™ was declared here
  219 |         uint64_t x;
      |                  ^
```
Hmmm.

medhefgo:
> Do we have any idea what the impact is on performance?

I genuinely don't know. The only thing I can offer is that a quick look at compiler explorer suggests to me that the cost would only be paid for variables where gcc cannot prove all code paths are initializing the var anyways.

keszybz:
I compiled out test-suite at -O2 with and without this patch, and it's certainly not slower with the patch. I actually got a slight speedup, but I didn't do enough tests to get statistical significance.

PR #23298, opened 2022-05-07
core/device: Use DEVICE_NOT_FOUND instead of 0.

daandemeyer:
Use the enum name instead of the value zero. This is a noop change
as DEVICE_NOT_FOUND == 0.


yuwata (approved):

DaanDeMeyer:
Changed one more occurrence of 0 to DEVICE_NOT_FOUND.

yuwata (approved):
Still LGTM.

DaanDeMeyer:
Found and fixed two more occurrences

bluca (approved):

PR #23299, opened 2022-05-07
cgroup: support config cgroup cpu.idle by systemd

wineway:
this [patch](https://lore.kernel.org/lkml/162971078674.25758.15464079371945307825.tip-bot2@tip-bot2/#r) introduced a new option named `cpu.idle` in cgroup/cpu, and already added to https://github.com/opencontainers/runtime-spec/pull/1136, seems we should also make systemd could manage it


poettering:
I don't understand what the patch does? the patch seems to add a boolean option, but your patch here adds a numeric option?

so, what does this do? can you explain?

please add an brief explanation of the option to man/systemd.resource-control.xml, where we document all stuff backed by cgroup stuff...

htejun:
I think it was from google. They wanted hierarchical idle scheduling - ie. at each level, if some cgroups have idle set, they are guaranteed to run only when no non-idle sibling is runnable. It's a niche feature that google and scheduler folks wanted. Maybe leave it be until some people really want it?

kolyshkin:
> the patch seems to add a boolean option, but your patch here adds a numeric option?

Quoting the kernel commit message (https://lore.kernel.org/lkml/20210730020019.1487127-2-joshdon@google.com/):

> The per-task_group idle field is an integer that currently only holds
> either a 0 or a 1. This is explicitly typed as an integer to allow for
> further extensions to this API. For example, a negative value may
> indicate a highly latency-sensitive cgroup that should be preferred for
> preemption/placement/etc.

poettering:
@wineway so is that something you actually intend to use? or something you just wanted to add support for for completeness?

bluca:
> @wineway so is that something you actually intend to use? or something you just wanted to add support for for completeness?

Actually I think I could make good use of this setting, sounds like it would nicely complement the Condition PSI settings for long lived services. We could even configure this in background.slice maybe?

poettering:
> > the patch seems to add a boolean option, but your patch here adds a numeric option?
> 
> Quoting the kernel commit message (https://lore.kernel.org/lkml/20210730020019.1487127-2-joshdon@google.com/):
> 
> > The per-task_group idle field is an integer that currently only holds
> > either a 0 or a 1. This is explicitly typed as an integer to allow for
> > further extensions to this API. For example, a negative value may
> > indicate a highly latency-sensitive cgroup that should be preferred for
> > preemption/placement/etc.

We don't necessarily want to expose things 1:1 exposing the raw kernel semantics. We genreally try to be friendlier than the kernel. i.e. to me this appears to be something that is boolean for now, and maybe supports more values later, but still a small number. hence maybe expose this conceptually as an enum for now, that happens to also parse a boolean value... We do that for a couple of settings, for example ProtectSystem=, which is primarily a boolean, but also takes some additional, special values.

I am pretty sure exposing a raw integer is not the right approach...

That all said, I'd like to understand what the usecase here is: should we really care? are there people who actually want this? I'd rather avoid adding support for this, unless there's real value for people in this.

bluca:
> That all said, I'd like to understand what the usecase here is: should we really care? are there people who actually want this? I'd rather avoid adding support for this, unless there's real value for people in this.

I think I'd use this - for short-lived tasks we now have the ConditionCPUPressure which works nicely, but for long-running services obviously that doesn't help. This could fill that gap.
Another example, we could set this in `background.slice`, it should fit the definition pretty well.

poettering:
but why wouldn't `CPUShares=<somethinglow>` not be the better, safer choice for that? totally starving out processes sounds like risky business, no?

wineway:
> @wineway so is that something you actually intend to use? or something you just wanted to add support for for completeness?

for this kernel feature, yes, and for the systemd's implementation, for me mainly for the completeness.

we intend to use this kernel feature in container environments to run a regular service and a less latency-sensitive service in one host machine

before this patch merged, many other applications(like `kubernetes qos class`) use a minimum `cpu.weight` intend to do the same thing

with this feature, we can use a more explicit configuration and get a better effect.

since mostly container runtimes rely on `systemd` driver for `cgroup`, though using `cgroupfs` directly is an alternative solution, for completeness and for the one who uses container with `systemd-cgroup` driver seems to be worthy to support it.

and as @bluca mentioned, this feature is useful besides what is described above.

bluca:
> but why wouldn't `CPUShares=<somethinglow>` not be the better, safer choice for that? totally starving out processes sounds like risky business, no?

Those are hard limits. We use them too, of course - but just like with the conditions, some tasks are of the type 'I'd like to run and use a lot of resources, but it's fine if I don't because of high-prio items'. With a fixed setting you have a hard ceiling regardless of the state of the system.

bluca:
> > Those are hard limits. We use them too, of course - but just like with the conditions, some tasks are of the type 'I'd like to run and use a lot of resources, but it's fine if I don't because of high-prio items'. With a fixed setting you have a hard ceiling regardless of the state of the system.
> 
> i am not sure i grok the "hard limits" and "hard ceiling" references? `CPUWeight=1` means "if the system is under CPU pressure give siblings of mine each 100x more CPU than me". (because `CPUWeight=100` is the default)
> 
> As I understand the `cpu.idle` thing it means "if the system is under CPU pressure give me nothing".
> 
> What concerns me about the `cpu.idle` thing is that it makes processes totally hang if there's just one runnable task that has idle not set. Since traditionally batch jobs would use CPUWeight=1 or nice=19 or something like this it means that if you run anything like that in parallel then your service doesn't scheduled _at_all* anymore. and i doubt people would expect that...

Uhm CPUWeight takes into account current pressure? It wasn't obvious from our manpage, but maybe I'm conflating that with CPUQuota in my head

wineway:
> What concerns me about the cpu.idle thing is that it makes processes totally hang if there's just one runnable task that has idle not set. Since traditionally batch jobs would use CPUWeight=1 or nice=19 or something like this it means that if you run anything like that in parallel then your service doesn't scheduled atall* anymore. and i doubt people would expect that...

seems the sematic we want is: 

- Make non-idle processes use full of CPU as if no idle processes are running
- Minimize the impact of idle processes on non-idle processes.

in our use case, we have two services `service a` and `serivce b` , `service a` has a dynamic CPU usage and a higher priority, and `service b` is a batch job that always tries to use all CPU time but we only expect it make full use of remained    CPU time and finish in an arbitrary future time.

we benched `cpu.idle` and minimum `cpu.weight`, the processes assigned `cpu.idle` have much less impact on regular processes.

🤔I think it might not a niche requirement, especially in cloud and container environments。

wineway:
seems I should make `CPUIlde` an enum value? 
```
CPUIldle=yes
CPUIdle=no
CPUIlde=#something append for future?
```
and since we didn't expose it 1:1 exposing the raw kernel semantics. maybe we can rename it like `CPUWeightIdle`?
```
```

Werkov:
My 0.02€ -- as systemd adds own abstract layer on top of kernel knobs, let's allow setting `CPUWeight=0` (internally translates to `cpu.idle` attribute). It's a conservative change but it makes user aware that setting weight to 0 is something special, if kernel API is extended in the future, new abstraction (such as `CPUIdle=`) can be added to systemd too.

wineway:
> as systemd adds own abstract layer on top of kernel knobs, let's allow setting `CPUWeight=0` (internally translates to `cpu.idle` attribute). It's a conservative change but it makes user aware that setting weight to 0 is something special, if kernel API is extended in the future, new abstraction (such as `CPUIdle=`) can be added to systemd too.

@Werkov In my opinion, `CPUWeight` name same as `cpu.weight` in cgroup, we had better make them have semantic equally, otherwise it might make confuse, but I'm not sure about this, is there any further discussion?

keszybz:
I think we should support this knob in some fashion. If the kernel added this in cgroups-v2, this means pretty much means we should support it, since we generally try to provide a complete interface for cgroups-v2.

>  let's allow setting `CPUWeight=0`

I don't think we should conflate the two. `cpu.weight` exists also for processes with `cpu.idle=1`, so we need two switches to configure those independently.

But please make this take a boolean argument in the config. We'll 'yes|no' for now, and if the kernel interface is extended, we can extend the parser. OTOH, internally and in the dbus interface, I think it's fine if we keep this an integer. If kernel allows other values, we'll want to map them 1:1 to our own values.

wineway:
@keszybz @htejun 
made `CPUIdle` take a boolean argument and rebased to main, can I ask you review it?

htejun:
As for the interface, I have no objection.

man/systemd.resource-control.xml
```diff
@@ -198,6 +198,18 @@
...
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>CPUIdle=</varname></term>
+
+        <listitem>
+          <para>Assign the idle scheduling strategy to the processes executed, both the unified and legacy control group
+          is effiecient. Takes a boolean argument, default to no. If true, processes have idle schedule policy, which means mininum sched weight
+          and could be preemt by other non-idle processes in the same level of task group, Note that when setting this param
+          to yes, configure with CPUWeight is illegal.
```
keszybz: Maybe like this:
> Takes a boolean argument, defaults to false.
> When true, mark the cgroup for "idle scheduling", which means that it will get CPU resources only when there are no processes *not* marked in this way to execute in this cgroup or its siblings.
> This setting corresponds to the `cpu.idle` cgroup attribute (link).

Is CPUWeight really illegal to set?

keszybz: Maybe like this:
> Takes a boolean argument, defaults to false.
> When true, mark the cgroup for "idle scheduling", which means that it will get CPU resources only when there are no processes *not* marked in this way to execute in this cgroup or its siblings.
> This setting corresponds to the `cpu.idle` cgroup attribute.

Is CPUWeight really illegal to set?

src/basic/cgroup-util.h
```diff
@@ -115,12 +115,24 @@ extern const uint64_t cgroup_io_limit_defaults[_CGROUP_IO_LIMIT_TYPE_MAX];
 const char* cgroup_io_limit_type_to_string(CGroupIOLimitType t) _const_;
 CGroupIOLimitType cgroup_io_limit_type_from_string(const char *s) _pure_;
 
+/* Special values for the cpu.idle attribute */
+#define CGROUP_CPU_IDLE_INVALID INT64_MAX
+#define CGROUP_CPU_IDLE_DEFAULT INT64_C(0)
+#define CGROUP_CPU_IDLE_MIN INT32_MIN
+#define CGROUP_CPU_IDLE_MAX INT32_MAX
```
keszybz: This seems iffy. Why allow signed values? Shouldn't this be `uint64_t` ?

src/basic/cgroup-util.h
```diff
@@ -115,12 +115,24 @@ extern const uint64_t cgroup_io_limit_defaults[_CGROUP_IO_LIMIT_TYPE_MAX];
...
 /* Special values for the cpu.shares attribute */
 #define CGROUP_CPU_SHARES_INVALID UINT64_MAX
 #define CGROUP_CPU_SHARES_MIN UINT64_C(2)
 #define CGROUP_CPU_SHARES_MAX UINT64_C(262144)
 #define CGROUP_CPU_SHARES_DEFAULT UINT64_C(1024)
 
+static inline bool CGROUP_CPU_IDLE_IS_OK(int64_t x) {
+        return
+            x == CGROUP_CPU_IDLE_INVALID ||
+            (x >= CGROUP_CPU_IDLE_MIN && x <= CGROUP_CPU_IDLE_MAX);
```
keszybz: Hmm, shouldn't it reject anything except 0 or 1 ?

src/core/cgroup.c
```diff
@@ -858,6 +862,21 @@ static int lookup_block_device(const char *p, dev_t *ret) {
...
+static bool cgroup_context_has_cpu_idle(CGroupContext *c) {
+        return c->cpu_idle != CGROUP_CPU_IDLE_INVALID;
+}
+
+static int64_t cgroup_context_cpu_idle(CGroupContext *c, ManagerState state) {
+        return c->cpu_idle;
+}
+
+static void cgroup_apply_unified_cpu_idle(Unit *u, int64_t idle) {
+        char buf[DECIMAL_STR_MAX(int64_t) + 2];
```
keszybz: `+ 1`, DECIMAL_STR_MAX already takes the terminator into account.

src/core/cgroup.c
```diff
@@ -1383,6 +1402,10 @@ static void cgroup_context_apply(
          * we couldn't even write to them if we wanted to). */
         if ((apply_mask & CGROUP_MASK_CPU) && !is_local_root) {
 
+                if (cgroup_context_has_cpu_idle(c)) {
+                        cgroup_apply_unified_cpu_idle(u, cgroup_context_cpu_idle(c, state));
+                }
```
keszybz: No {} around single-line blocks.

src/core/dbus-cgroup.c
```diff
@@ -903,6 +904,36 @@ BUS_DEFINE_SET_CGROUP_LIMIT(memory_protection, CGROUP_MASK_MEMORY, physical_memo
...
+
+        assert(p);
+
+        r = sd_bus_message_read(message, "x", &v);
+        if (r < 0)
+                return r;
+
+        if (CGROUP_CPU_IDLE_IS_OK(v)) {
+                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Value specified in %s is out of range", name);
+        }
```
keszybz: No {}.

src/core/dbus-cgroup.c
```diff
@@ -903,6 +904,36 @@ BUS_DEFINE_SET_CGROUP_LIMIT(memory_protection, CGROUP_MASK_MEMORY, physical_memo
...
+
+        if (!UNIT_WRITE_FLAGS_NOOP(flags)) {
+                *p = v;
+                unit_invalidate_cgroup(u, CGROUP_MASK_CPU);
+                unit_write_settingf(u, flags, name,
+                "%s=%" PRId64, name, v);
```
keszybz: Indentation.

man/systemd.resource-control.xml
wineway: yes, look at [sched_group_set_shares](https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee#diff-75a417d0e6ebb27175a48c6810294ab6e21847776b1e46c62de9cccfb428883eR11494), set CPUWeight to idle group will return EINVAL

wineway: > Is CPUWeight really illegal to set?

@keszybz  yes, look at [sched_group_set_shares](https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee#diff-75a417d0e6ebb27175a48c6810294ab6e21847776b1e46c62de9cccfb428883eR11494), set CPUWeight to idle group will return EINVAL

src/basic/cgroup-util.h
wineway: > This seems iffy. Why allow signed values? Shouldn't this be `uint64_t` ?

this is because `cpu.idle` is [s64](https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee#diff-cc1a82129952a910fdc4292448c2a097a2ba538bebefcf3c06381e45639ae73eR10371) in the kernel, maybe there will be more values like `-1` in the future, I think `int64_t` can make more compatibility

src/basic/cgroup-util.h
wineway: > Hmm, shouldn't it reject anything except 0 or 1 ?

yes, we can  reject anything except 0 or 1 now

Werkov:
> In my opinion, `CPUWeight` name same as `cpu.weight` in cgroup, we had better make them have semantic equally, otherwise it might make confuse, but I'm not sure about this, is there any further discussion?

I'm not aware of such a discussion. Just look at the example of `CPUQuota=` vs `cpu.max` (stems from legacy name `cpu.runtime_quota_us`). In my opinion there's value added in the abstraction. Hardcoded 1:1 translation may be also interpreted that systemd stands in the way of a much simpler approach (e.g. special key-value configs that'd simply write through to cgroup attributes).


> If the kernel added this in cgroups-v2, this means pretty much means we should support it, since we generally try to provide a complete interface for cgroups-v2.

I agree with the completeness goal. Note, however, that `cpu.idle` is [not officially documented](https://docs.kernel.org/admin-guide/cgroup-v2.html#interface-files). I consider the concept immature.

> I don't think we should conflate the two. cpu.weight exists also for processes with cpu.idle=1, so we need two switches to configure those independently.

`cpu.weight` exists but can it be [really configured](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/kernel/sched/fair.c?h=v5.18.10#n11684)? (I don't know what's the intention, this is just to illustrate that the concept looks half-baked to me.)


@wineway Do you have an use case where you'd need configure `cpu.idle` via systemd? (Specifically, is not setting `Delegate=cpu` and assigning `cpu.idle` manually in a *subtree* sufficient for you?) Also, could you please clarify what is your expectation of interaction between `cpu.idle=1` and any `cpu.weight`?

wineway:
> > In my opinion, `CPUWeight` name same as `cpu.weight` in cgroup, we had better make them have semantic equally, otherwise it might make confuse, but I'm not sure about this, is there any further discussion?
> 
> Just look at the example of `CPUQuota=` vs `cpu.max` (stems from legacy name `cpu.runtime_quota_us`)

for `CPUQuota`, we can recognize it might have a different value range since they have a different name with cgroup entry, but for `CPUWeight`, I think the same name might cause confusion.

> `cpu.weight` exists but can it be [really configured](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/kernel/sched/fair.c?h=v5.18.10#n11684)? (I don't know what's the intention, this is just to illustrate that the concept looks half-baked to me.)

this is because `cpu.idle` already has the [minimum weight](https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee#diff-75a417d0e6ebb27175a48c6810294ab6e21847776b1e46c62de9cccfb428883eR11565), we also can't set priority to a process which has an idle policy. just think as `cpu.idle` is idle policy of `cgroup`.

> 
> @wineway Do you have an use case where you'd need configure `cpu.idle` via systemd? (Specifically, is not setting `Delegate=cpu` and assigning `cpu.idle` manually in a _subtree_ sufficient for you?) Also, could you please clarify what is your expectation of interaction between `cpu.idle=1` and any `cpu.weight`?

we use this entry in our container environment, and use `cgroupfs` as cgroup driver, so we don't really need to configure `cpu.idle` via systemd now, we are mainly for completeness to add this support ;)

poettering:
so my understanding is that the CPUWeight= and this new idle feature are exclusive to each other? either you mark a cgroup as idle, or you give it a weight?

If so, I think it might actually make sense to make this a special case for CPUWeight=. i.e. `CPUWeight=idle` means turn on idle mode for this. Internally (and maybe even over dbus) we'd map that to zero, but this would not be visible in the unit file syntax.

(I'd also be OK if this was a boolean value of its own, but given this isn't really orthogonal to `CPUWeight=` I think folding this conceptually into becoming a special `CPUWeight=` case would make sense)

poettering:
> @Werkov In my opinion, `CPUWeight` name same as `cpu.weight` in cgroup, we had better make them have semantic equally, otherwise it might make confuse, but I'm not sure about this, is there any further discussion?

So, we are higher level than the kernel, so what we expose is supposed to be higher level too. Of course we are not going to expose stuff we couldn't implement on top of the kernel, and we'll not depart from the semantics/values/ranges the kernel defines, but we definitely want to expose a friendlier way to configure things.

poettering:
> I don't think we should conflate the two. `cpu.weight` exists also for processes with `cpu.idle=1`, so we need two switches to configure those independently.

is this true? i assumed the weight as no effect once you turn on idle?

poettering:
> I think we should support this knob in some fashion. If the kernel added this in cgroups-v2, this means pretty much means we should support it, since we generally try to provide a complete interface for cgroups-v2.

i am not sure i agree with that. Let's support what makes sense, I'd not expose stuff that noone needs and is of questionnable use outside of very specific usecases. For example, stuff that let's say android needs, but noone else definitely shouldn't be exposed by us.

man/org.freedesktop.systemd1.xml
```diff
@@ -2503,6 +2503,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
       @org.freedesktop.DBus.Property.EmitsChangedSignal("false")
       readonly t StartupCPUShares = ...;
       @org.freedesktop.DBus.Property.EmitsChangedSignal("false")
+      readonly x CPUIdle = ...;
```
poettering: this certainly should not be of type "x". It should either be a boolean, or a string. I understand it's actually a boolean in the kernel right now too, but exported as an integer for "future additions". But we should not expose a low-level concept like that. If it's an enum (i.e. "regular" + "idle") then expose it as string (because dbus doesn't have enums). But exposing this as integer is definitely wrong, because it doesn't really take arbitrary numeric values.

man/systemd.resource-control.xml
```diff
@@ -198,6 +198,19 @@
...
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>CPUIdle=</varname></term>
+
+        <listitem>
+          <para>Takes a boolean argument, defaults to false.
+          When true, mark the cgroup for "idle scheduling", which means that it will get CPU resources only
+          when there are no processes not marked in this way to execute in this cgroup or its siblings.
+          Note that setting CPUWeight is illegal if CPUIdle is true.
```
poettering: Enclose CPUWeight with `<varname>` and suffix settings with `=`, to clarify, they are seetings, i.e.

`<varname>CPUWeight=</varname>`

Also, the fact that CPUIdle=/CPUWeight= are exclusive to each other and cannot both be used is am artifact of the fact that a cgroup can either be considered of idle or regular scheduling, which is what you should say. 

Moreover afaics it's not actually illegal to set both in the current patch, it's just that the weight has no effect if idle mode is used.

man/systemd.resource-control.xml
```diff
@@ -198,6 +198,19 @@
...
+          This setting corresponds to the cpu.idle cgroup attribute.
```
poettering: use `<literal>` for `cpu.idle`

man/systemd.resource-control.xml
```diff
@@ -198,6 +198,19 @@
...
+          </para>
+        </listitem>
+      </varlistentry>
```
poettering: also needs an update to `CPUWeight=` docs, saying that `CPUWeight=` has no effect if `CPUIdle=` is set (that's how this works, no?)

Also would be good to briefly say what happens if i have a slice with idle mode and a service below it which does not have idle mode. Is the service then of idle mode anyway?

Also would be good to briefly mention how the process SCHED_IDLE stuff relates to this per cgroup stuff: what wins?

src/basic/cgroup-util.h
```diff
@@ -115,12 +115,24 @@ extern const uint64_t cgroup_io_limit_defaults[_CGROUP_IO_LIMIT_TYPE_MAX];
 const char* cgroup_io_limit_type_to_string(CGroupIOLimitType t) _const_;
 CGroupIOLimitType cgroup_io_limit_type_from_string(const char *s) _pure_;
 
+/* Special values for the cpu.idle attribute */
+#define CGROUP_CPU_IDLE_INVALID INT64_MIN
+#define CGROUP_CPU_IDLE_DEFAULT INT64_C(0)
+#define CGROUP_CPU_IDLE_MIN 0
+#define CGROUP_CPU_IDLE_MAX 1
```
poettering: it's not really a number. the kernel interface is just weird, but let's not expose it like that. make it a boolean, a tri-state (i.e. a boolean encoded in an `int`, which has the additional value -1 for "unset") or a proper enum. But definitely not an int64_t.

src/core/dbus-cgroup.c
```diff
@@ -903,6 +904,34 @@ BUS_DEFINE_SET_CGROUP_LIMIT(memory_protection, CGROUP_MASK_MEMORY, physical_memo
...
+        int r;
+
+        assert(p);
+
+        r = sd_bus_message_read(message, "x", &v);
+        if (r < 0)
+                return r;
+
+        if (CGROUP_CPU_IDLE_IS_OK(v))
+                return sd_bus_error_setf(error, SD_BUS_ERROR_INVALID_ARGS, "Value specified in %s is out of range", name);
```
poettering: please always specify which value, i.e. the invalid value

src/nspawn/nspawn-oci.c
```diff
@@ -1288,6 +1289,21 @@ static int oci_cgroup_cpu_shares(const char *name, JsonVariant *v, JsonDispatchF
...
 }
 
+static int oci_cgroup_cpu_idle(const char *name, JsonVariant *v, JsonDispatchFlags flags, void *userdata) {
+        int64_t *u = userdata;
+        int64_t k;
+
+        assert(u);
+
+        k = json_variant_integer(v);
+        if (k <= INT32_MIN || k >= INT32_MAX)
```
poettering: hmm, what? in OCI it can have other values than 0 and 1?

src/shared/cgroup-setup.c
```diff
@@ -153,6 +153,26 @@ bool cg_is_hybrid_wanted(void) {
...
+
+        if (isempty(s)) {
+                *ret = CGROUP_CPU_IDLE_INVALID;
+                return 0;
+        }
+
+        /* cpu.idle only support 0 or 1 now, and if this value accept more value in the future, we will add
+         * parse logic here. */
+        r = parse_boolean(s);
+
```
poettering: please keep function calls and their immediate error handling together. i.e. remove this empty line. 

place empty lines that separate the error handling of a call from the next call. but do not place empty lines between call and its error handling.

poettering (changes requested):

wineway:
> 
> If so, I think it might actually make sense to make this a special case for CPUWeight=. i.e. `CPUWeight=idle` means turn on idle mode for this. Internally (and maybe even over dbus) we'd map that to zero, but this would not be visible in the unit file syntax.

🤔I think `CPUWeight=idle` is more reasonable, we can distinguish it from other numeric values and realize its different, so I think we can make `CPUWeight` accept input as follow:
1. string: 'idle'
2. uint: greater than 0

and store `idle` as `const uint32_t CGROUP_CPU_SHARES_IDLE=0` internally.

poettering:
> thinkingI think `CPUWeight=idle` is more reasonable, we can distinguish it from other numeric values and realize its different, so maybe we can make `CPUWeight` accept input as follow:
> 
>     1. string: 'idle'
> 
>     2. uint: greater than 0
> 
> 
> and store `idle` as `const uint32_t CGROUP_CPU_SHARES_IDLE=0` internally.

Yupp, that's what I am proposing. And the D-Bus cpu weight property can also use 0 to encode idle mode.

man/systemd.resource-control.xml
```diff
@@ -180,14 +180,23 @@
...
-          <literal>cpu.weight</literal> control group attribute. The allowed range is 1 to 10000. Defaults to
-          100. For details about this control group attribute, see <ulink
-          url="https://docs.kernel.org/admin-guide/cgroup-v2.html">Control Groups v2</ulink>
-          and <ulink url="https://docs.kernel.org/scheduler/sched-design-CFS.html">CFS
-          Scheduler</ulink>.  The available CPU time is split up among all units within one slice relative to
-          their CPU time weight. A higher weight means more CPU time, a lower weight means less.</para>
+          <para>These options may take integer value or string value "idle"</para>
+            <itemizedlist>
+              <listitem>
+                <para>If it take integer value, Assign the specified CPU time weight to the processes executed,
```
poettering: lowercase `Assign` → `assign`, please

man/systemd.resource-control.xml
```diff
@@ -180,14 +180,23 @@
...
         <listitem>
-          <para>Assign the specified CPU time weight to the processes executed, if the unified control group
-          hierarchy is used on the system. These options take an integer value and control the
-          <literal>cpu.weight</literal> control group attribute. The allowed range is 1 to 10000. Defaults to
-          100. For details about this control group attribute, see <ulink
-          url="https://docs.kernel.org/admin-guide/cgroup-v2.html">Control Groups v2</ulink>
-          and <ulink url="https://docs.kernel.org/scheduler/sched-design-CFS.html">CFS
-          Scheduler</ulink>.  The available CPU time is split up among all units within one slice relative to
-          their CPU time weight. A higher weight means more CPU time, a lower weight means less.</para>
+          <para>These options may take integer value or string value "idle"</para>
```
poettering: `These options accept an integer value or a the special string "idle":`

man/systemd.resource-control.xml
poettering: also `If set to an integer value, … `

man/systemd.resource-control.xml
```diff
@@ -1188,20 +1197,27 @@ DeviceAllow=/dev/loop-control
...
-          <varname>CPUShares=</varname> applies to normal runtime of the system, and if the former is not set also to
-          the startup and shutdown phases. Using <varname>StartupCPUShares=</varname> allows prioritizing specific services at
-          boot-up and shutdown differently than during normal runtime.</para>
-
-          <para>Implies <literal>CPUAccounting=yes</literal>.</para>
-
+          <para>These options may take integer value or string value "idle"</para>
+            <itemizedlist>
+              <listitem>
+                <para>If it take integer value, Assign the specified CPU time share weight to the processes executed. These options
```
poettering, line 1203: as above

man/systemd.resource-control.xml
```diff
@@ -1188,20 +1197,27 @@ DeviceAllow=/dev/loop-control
...
-          weight.</para>
-
-          <para>While <varname>StartupCPUShares=</varname> applies to the startup and shutdown phases of the system,
-          <varname>CPUShares=</varname> applies to normal runtime of the system, and if the former is not set also to
-          the startup and shutdown phases. Using <varname>StartupCPUShares=</varname> allows prioritizing specific services at
-          boot-up and shutdown differently than during normal runtime.</para>
-
-          <para>Implies <literal>CPUAccounting=yes</literal>.</para>
-
+          <para>These options may take integer value or string value "idle"</para>
```
poettering: No. CPUShares= is obsolete. Please do not add new features to it. Just ignore CPUShares.

src/basic/cgroup-util.h
```diff
@@ -84,8 +84,10 @@ static inline CGroupMask CGROUP_MASK_EXTEND_JOINED(CGroupMask mask) {
 CGroupMask get_cpu_accounting_mask(void);
 bool cpu_accounting_is_cheap(void);
 
+#define CGROUP_IDLE_VALUE UINT64_C(0)
 /* Special values for all weight knobs on unified hierarchy */
 #define CGROUP_WEIGHT_INVALID UINT64_MAX
+#define CGROUP_WEIGHT_IDLE CGROUP_IDLE_VALUE
```
poettering: why indirect via two macros? one suffices, i.e. just do:

```c
#define CGROUP_WEIGHT_IDLE UINT64_C(0)
```

src/basic/cgroup-util.h
```diff
@@ -117,6 +119,7 @@ CGroupIOLimitType cgroup_io_limit_type_from_string(const char *s) _pure_;
 
 /* Special values for the cpu.shares attribute */
 #define CGROUP_CPU_SHARES_INVALID UINT64_MAX
+#define CGROUP_CPU_SHARES_IDLE CGROUP_IDLE_VALUE
```
poettering: nope, do not add this to legacy CPUShares, please

src/core/cgroup.c
```diff
@@ -952,6 +957,21 @@ static void cgroup_apply_unified_cpu_weight(Unit *u, uint64_t weight) {
         (void) set_attribute_and_warn(u, "cpu", "cpu.weight", buf);
 }
 
+static int set_cpu_idle_and_warn(Unit *u, uint32_t value) {
+        int r;
+
+        r = cg_set_attribute("cpu", u->cgroup_path, "cpu.idle", value ? "1" : "0");
```
poettering: use one_zero()

src/core/cgroup.c
```diff
@@ -952,6 +957,21 @@ static void cgroup_apply_unified_cpu_weight(Unit *u, uint64_t weight) {
         (void) set_attribute_and_warn(u, "cpu", "cpu.weight", buf);
 }
 
+static int set_cpu_idle_and_warn(Unit *u, uint32_t value) {
```
poettering: I think this should take an `uint64_t weight` value

src/core/cgroup.c
```diff
@@ -987,11 +1007,17 @@ static void cgroup_apply_legacy_cpu_quota(Unit *u, usec_t quota, usec_t period)
...
+        if (shares == CGROUP_CPU_SHARES_IDLE)
+                return CGROUP_WEIGHT_IDLE;
+
         return CLAMP(shares * CGROUP_WEIGHT_DEFAULT / CGROUP_CPU_SHARES_DEFAULT,
                      CGROUP_WEIGHT_MIN, CGROUP_WEIGHT_MAX);
 }
 
 static uint64_t cgroup_cpu_weight_to_shares(uint64_t weight) {
+        if (weight == CGROUP_WEIGHT_IDLE)
+                return CGROUP_CPU_SHARES_IDLE;
```
poettering: does the idle stuff even exist on cgroupv1? If not, such a conversion doesn't make any sense.

src/core/cgroup.c
```diff
@@ -952,6 +957,21 @@ static void cgroup_apply_unified_cpu_weight(Unit *u, uint64_t weight) {
...
+        if (r < 0 && (r != -ENOENT || value))
+                log_unit_full_errno(u, LOG_LEVEL_CGROUP_WRITE(r), r, "Failed to set '%s' attribute on '%s' to '%" PRIu32 "': %m",
+                                    "cpu.idle", empty_to_root(u->cgroup_path), value);
+
+        return r;
+}
+
+static void cgroup_apply_unified_cpu_idle(Unit *u, bool idle) {
+        (void) set_cpu_idle_and_warn(u, idle);
+}
```
poettering, line 968: what's the point of this helper?

src/core/cgroup.c
```diff
@@ -1381,7 +1407,9 @@ static void cgroup_context_apply(
          * we couldn't even write to them if we wanted to). */
         if ((apply_mask & CGROUP_MASK_CPU) && !is_local_root) {
 
-                if (cg_all_unified() > 0) {
+                if (cgroup_context_is_cpu_idle(c))
+                        cgroup_apply_unified_cpu_idle(u, true);
```
poettering: how does the idle field actually work? if i want to switch a running service from idle mode to non-idle mode, wouldn't i have to first write "0" into the idle cgroupfs attr file, and then fix the weight value? Or does setting the weight value suffice to turn off idle mode again?

it appears to me, you kinda have to write the idle field always to make this robust, no?

src/core/cgroup.c
```diff
@@ -1416,6 +1445,7 @@ static void cgroup_context_apply(
                         else
                                 shares = CGROUP_CPU_SHARES_DEFAULT;
 
+                        cgroup_apply_unified_cpu_idle(u, false);
```
poettering: nope, as mentioned, don't add features to cgroupsv1 stuff. Just ignore cgroupsv1. And also as mentioned: are you sure even if we wanted to support the idle stuff on cgroupsv1: does it even support that? i doubt it, but too lazy to check

src/core/dbus-cgroup.c
```diff
@@ -894,15 +894,71 @@ static int bus_cgroup_set_boolean(
...
+        if (!CGROUP_WEIGHT_IS_OK(v) && v != CGROUP_WEIGHT_IDLE)
+                return sd_bus_error_setf(
+                                error, SD_BUS_ERROR_INVALID_ARGS, "Value specified in %s is out of range", name);
+        if (!UNIT_WRITE_FLAGS_NOOP(flags)) {
+                *p = v;
+                unit_invalidate_cgroup(u, CGROUP_MASK_CPU);
+                if (v == CGROUP_WEIGHT_INVALID)
+                        unit_write_settingf(u, flags, name, "%s=", name);
+                else if (v == CGROUP_WEIGHT_IDLE)
+                        unit_write_settingf(u, flags, name, "%s=%s", name, "idle");
```
poettering: given you output a literal string here as value, you might just use `"%s=idle"` as format string

src/core/dbus-cgroup.c
```diff
@@ -894,15 +894,71 @@ static int bus_cgroup_set_boolean(
...
+                else
+                        unit_write_settingf(u, flags, name, "%s=%" PRIu64, name, v);
+        }
+        return 1;
+}
+
+static int bus_cgroup_set_cpu_shares(
```
poettering: again, nope, no new features to cgrouspv1 CPUShares, thanks.

src/core/cgroup.c
wineway: > does the idle stuff even exist on cgroupv1? If not, such a conversion doesn't make any sense.

yes, idle exists in cgroupv1

poettering: ok, but still: we do not add new features to our cgroupsv1 support. So even if it supports it, our code should not.

src/core/cgroup.c
wineway: > how does the idle field actually work? 

If we switch service from idle mode to non-idle mode, we have to set idle attr to `0` and then give a weight, and if we need to switch it from non-idle to idle, we only need set idle attr to `1` and mustn't give it a weight.

> it appears to me, you kinda have to write the idle field always to make this robust, no?

*I am not sure I got it* 
yes, we should always write this idle field. we can't modify the weight of an idle service, so write this idle field to `0` before modifying its weight. do you means we should do: 
```
if (cgroup_context_is_cpu_idle(c))
  cgroup_apply_unified_cpu_idle(u, true);
else {
  cgroup_apply_unified_cpu_idle(u, false);
  // modify some weight
}

Comment (poettering) [reply src/core/cgroup.c]: yes, something like that

File: src/core/cgroup.c (hunk 27 above)
Comment (wineway) [reply src/core/cgroup.c]: > nope, as mentioned, don't add features to cgroupsv1 stuff. Just ignore cgroupsv1. And also as mentioned: are you sure even if we wanted to support the idle stuff on cgroupsv1: does it even support that? i doubt it, but too lazy to check

update:
ok, we can warn if user setting `CPUWeight=idle` when using cgroupv1

----------------

yes, idle is supported in cgroupv1, we can make `CPUShares` not accept `idle`, but `CPUWeight` will be transformed to `cpu.shares` if the host is using cgroupv1.

Comment (poettering) [reply src/core/cgroup.c]: when parsing CPUShares=idle, don't handle it specially, just let parsing fail.

Comment (poettering) [reply src/core/cgroup.c]: and if CPuWeight= gets converted into CPUShares when on cgroupsv1, just map "idle" to lowest share value we know. THe mapping is imperfect anyway...

Comment (wineway) [reply src/core/cgroup.c]: > and if CPuWeight= gets converted into CPUShares when on cgroupsv1, just map "idle" to lowest share value we know. THe mapping is imperfect anyway...

I do this mapping and add instructions in manual

File: src/core/cgroup.c (hunk 25 above)
Comment (wineway) [reply src/core/cgroup.c:968]: > what's the point of this helper?

removed this helper

File: src/core/cgroup.c [hunk 30]
```diff
@@ -952,6 +952,15 @@ static void cgroup_apply_unified_cpu_weight(Unit *u, uint64_t weight) {
         (void) set_attribute_and_warn(u, "cpu", "cpu.weight", buf);
 }
 
+static void cgroup_apply_unified_cpu_idle(Unit *u, uint64_t weight) {
+        int r;
+
+        r = cg_set_attribute("cpu", u->cgroup_path, "cpu.idle", one_zero(weight));
```
poettering: this looks wrong. You want to enable idle mode if weight is 0, i.e. use:

```c
…, one_zero(weight == CGROUP_WEIGHT_IDLE));
```

src/core/cgroup.c
```diff
@@ -952,6 +952,15 @@ static void cgroup_apply_unified_cpu_weight(Unit *u, uint64_t weight) {
...
+        if (r < 0 && (r != -ENOENT || weight))
```
poettering: similar, check needs to be: `weight == CGROUP_WEIGHT_IDLE

src/core/cgroup.c
```diff
@@ -952,6 +952,15 @@ static void cgroup_apply_unified_cpu_weight(Unit *u, uint64_t weight) {
...
+                log_unit_full_errno(u, LOG_LEVEL_CGROUP_WRITE(r), r, "Failed to set '%s' attribute on '%s' to '%" PRIu64 "': %m",
+                                    "cpu.idle", empty_to_root(u->cgroup_path), weight);
```
poettering: log message is confusing, it should probably show the result of `one_zero(weight == CGROUP_WEIGHT_IDLE)` and not the literal value of `weight`

src/core/cgroup.c
```diff
@@ -1397,7 +1410,9 @@ static void cgroup_context_apply(
                         } else
                                 weight = CGROUP_WEIGHT_DEFAULT;
 
-                        cgroup_apply_unified_cpu_weight(u, weight);
+                        cgroup_apply_unified_cpu_idle(u, !weight);
+                        if (weight != CGROUP_WEIGHT_IDLE)
+                                cgroup_apply_unified_cpu_weight(u, weight);
```
poettering: please move the `weight != CGROUP_WEIGHT_IDLE` check into the `cgroup_apply_unified_cpu_weight()` function and exit early there.

src/shared/cgroup-setup.c
```diff
@@ -173,6 +173,31 @@ int cg_weight_parse(const char *s, uint64_t *ret) {
...
+        r = safe_atou64(s, &u);
+        if (r < 0)
+                return r;
+
+        if (u < CGROUP_WEIGHT_MIN || u > CGROUP_WEIGHT_MAX)
+                return -ERANGE;
+
+        *ret = u;
+        return 0;
+}
```
poettering, line 180: make this a wrapper around the regular cg_weight_parse(), i.e.:

```c
int cg_cpu_weight_parse(const char *s, uint64_t *ret) {
        if (streq_ptr(s, "idle"))
                return *ret = CGROUP_WEIGHT_IDLE;
        return cg_weight_parse(s, ret);
 }

File: man/systemd.resource-control.xml [hunk 35]
```diff
@@ -180,14 +180,25 @@
...
+                and <ulink url="https://docs.kernel.org/scheduler/sched-design-CFS.html">CFS
+                Scheduler</ulink>.  The available CPU time is split up among all units within one slice relative to
+                their CPU time weight. A higher weight means more CPU time, a lower weight means less.</para>
+              </listitem>
+              <listitem><para>If set to the special string "idle", mark the cgroup for "idle scheduling", which means
+              that it will get CPU resources only when there are no processes not marked in this way to execute in this
+              cgroup or its siblings. This setting corresponds to the <literal>cpu.idle</literal> cgroup attribute.</para>
+
+              <para>Note that this value is only efficient for cgroup-v2, for cgroup-v1 it is equivalent to the minimum weight</para>
+              </listitem>
```
poettering, line 201: Please indent the two list items similar in style.

poettering (changes requested):

src/core/cgroup.c
```diff
@@ -1397,7 +1410,9 @@ static void cgroup_context_apply(
                         } else
                                 weight = CGROUP_WEIGHT_DEFAULT;
 
-                        cgroup_apply_unified_cpu_weight(u, weight);
+                        cgroup_apply_unified_cpu_idle(u, !weight);
```
poettering: drop the `!`. Just pass the weight directly, and make it a boolean balue inside of cgroup_apply_unified_cpu_idle()

man/systemd.resource-control.xml
wineway: > Please indent the two list items similar in style

sorry, my fault...how about now

man/systemd.resource-control.xml
```diff
@@ -180,14 +180,26 @@
...
+                100. For details about this control group attribute, see <ulink
+                url="https://docs.kernel.org/admin-guide/cgroup-v2.html">Control Groups v2</ulink>
+                and <ulink url="https://docs.kernel.org/scheduler/sched-design-CFS.html">CFS
+                Scheduler</ulink>.  The available CPU time is split up among all units within one slice relative to
+                their CPU time weight. A higher weight means more CPU time, a lower weight means less.</para>
+              </listitem>
+              <listitem>
+                <para>If set to the special string "idle", mark the cgroup for "idle scheduling", which means
+              that it will get CPU resources only when there are no processes not marked in this way to execute in this
+              cgroup or its siblings. This setting corresponds to the <literal>cpu.idle</literal> cgroup attribute.</para>
```
poettering: hmm, this indentation is particularly weird now

poettering:
lgtm, but the indentation of the man page docbook XML is still a bit weird

man/systemd.resource-control.xml
wineway: 👀 fixed here……

man/systemd.resource-control.xml
```diff
@@ -180,14 +180,26 @@
...
+                and <ulink url="https://docs.kernel.org/scheduler/sched-design-CFS.html">CFS
+                Scheduler</ulink>.  The available CPU time is split up among all units within one slice relative to
+                their CPU time weight. A higher weight means more CPU time, a lower weight means less.</para>
+              </listitem>
+              <listitem>
+                <para>If set to the special string "idle", mark the cgroup for "idle scheduling", which means
+                that it will get CPU resources only when there are no processes not marked in this way to execute in this
+                cgroup or its siblings. This setting corresponds to the <literal>cpu.idle</literal> cgroup attribute.</para>
+
+                <para>Note that this value is only efficient for cgroup-v2, for cgroup-v1 it is equivalent to the minimum weight</para>
```
poettering: is only efficient for → only has an effect on

Also, please end the sentence in a full stop (`.`)

PR #23300, opened 2022-05-07
core/device: Add more logging

daandemeyer:


src/core/device.c
```diff
@@ -598,8 +598,11 @@ static void device_process_new(Manager *m, sd_device *dev, const char *sysfs) {
         }
 
         /* Add additional units for all explicitly configured aliases */
-        if (sd_device_get_property_value(dev, "SYSTEMD_ALIAS", &alias) < 0)
+        r = sd_device_get_property_value(dev, "SYSTEMD_ALIAS", &alias);
+        if (r < 0) {
```
yuwata, line 602: Please ignore -ENOENT here.

yuwata (changes requested):

yuwata:
Also, CIs are unhappy...

DaanDeMeyer:
CI failures seem unrelated cc @mrc0mmand

mrc0mmand:
> CI failures seem unrelated cc @mrc0mmand

Yeah, please ignore them. There's something wrong with the network and it, of course, surfaced during Friday's evening, so fix has to wait until Monday...

src/core/device.c
```diff
@@ -908,6 +908,8 @@ static int device_dispatch_io(sd_device_monitor *monitor, sd_device *dev, void *
                 return 0;
         }
 
+        log_device_debug(dev, "Processing \"%s\" udev action", device_action_to_string(action));
```
yuwata: Please use `log_device_uevent()`.

src/core/device.c
```diff
@@ -992,12 +972,20 @@ void device_found_node(Manager *m, const char *node, DeviceFound found, DeviceFo
         if ((found & mask) != 0) {
                 _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
 
+                /* Return early if we're not dealing with a device. */
+                if (!path_startswith(node, "/dev"))
+                        return;
+
```
yuwata: This is gracefully handled by `sd_device_new_from_devname()`, no? Why do you explicitly exclude this?

src/core/device.c
```diff
@@ -709,16 +709,34 @@ static void device_update_found_by_name(Manager *m, const char *path, DeviceFoun
...
 
 static bool device_is_ready(sd_device *dev) {
+        int r;
+
         assert(dev);
 
-        if (device_is_renaming(dev) > 0)
+        r = device_is_renaming(dev);
+        if (r < 0)
+                log_device_error_errno(dev, r, "Failed to check if device is renaming: %m");
```
yuwata: Maybe, better to mention that it is assumed that the device is not renaming.

src/core/device.c
```diff
@@ -709,16 +709,34 @@ static void device_update_found_by_name(Manager *m, const char *path, DeviceFoun
...
+        if (r > 0) {
+                log_device_debug(dev, "Device busy: renaming");
                 return false;
+        }
 
         /* Is it really tagged as 'systemd' right now? */
-        if (sd_device_has_current_tag(dev, "systemd") <= 0)
+        r = sd_device_has_current_tag(dev, "systemd");
+        if (r < 0)
+                log_device_error_errno(dev, r, "Failed to check if device has \"systemd\" tag: %m");
```
yuwata: Also here. I think it is better to mention that we assume that the device does not have 'systemd' tag.

src/core/device.c
```diff
@@ -709,16 +709,34 @@ static void device_update_found_by_name(Manager *m, const char *path, DeviceFoun
...
+        if (r == 0)
+                log_device_debug(dev, "Device busy: not tagged with \"systemd\"");
+        if (r <= 0)
                 return false;
 
-        return device_get_property_bool(dev, "SYSTEMD_READY") != 0;
+        r = device_get_property_bool(dev, "SYSTEMD_READY");
+        if (r < 0 && r != -ENOENT)
+                log_device_error_errno(dev, r, "Failed to get device SYSTEMD_READY property: %m");
```
yuwata: Also here. I think it is better to mention that we assume that the device is ready.

src/core/device.c
```diff
@@ -709,16 +709,34 @@ static void device_update_found_by_name(Manager *m, const char *path, DeviceFoun
...
+        int r;
+
         assert(dev);
 
-        if (device_is_renaming(dev) > 0)
+        r = device_is_renaming(dev);
+        if (r < 0)
+                log_device_error_errno(dev, r, "Failed to check if device is renaming: %m");
+        if (r > 0) {
+                log_device_debug(dev, "Device busy: renaming");
```
yuwata: I guess this is quite noisy. Please drop it or introduce `log_device_trace()` and use it.

src/core/device.c
```diff
@@ -709,16 +709,34 @@ static void device_update_found_by_name(Manager *m, const char *path, DeviceFoun
...
                 return false;
+        }
 
         /* Is it really tagged as 'systemd' right now? */
-        if (sd_device_has_current_tag(dev, "systemd") <= 0)
+        r = sd_device_has_current_tag(dev, "systemd");
+        if (r < 0)
+                log_device_error_errno(dev, r, "Failed to check if device has \"systemd\" tag: %m");
+        if (r == 0)
+                log_device_debug(dev, "Device busy: not tagged with \"systemd\"");
```
yuwata: Also here. Please drop it or use log_device_trace().

src/core/device.c
```diff
@@ -709,16 +709,34 @@ static void device_update_found_by_name(Manager *m, const char *path, DeviceFoun
...
+        if (r <= 0)
                 return false;
 
-        return device_get_property_bool(dev, "SYSTEMD_READY") != 0;
+        r = device_get_property_bool(dev, "SYSTEMD_READY");
+        if (r < 0 && r != -ENOENT)
+                log_device_error_errno(dev, r, "Failed to get device SYSTEMD_READY property: %m");
+        if (r == 0)
+                log_device_debug(dev, "Device busy: SYSTEMD_READY property is false");
```
yuwata: Also here.

src/core/device.c
```diff
@@ -716,10 +716,10 @@ static bool device_is_ready(sd_device *dev) {
         r = device_is_renaming(dev);
         if (r < 0)
                 log_device_error_errno(dev, r, "Failed to check if device is renaming: %m");
-        if (r > 0) {
+        if (r > 0)
                 log_device_debug(dev, "Device busy: renaming");
+        if (r != 0)
```
yuwata: This changes the previous behavior... Why do you change it??

src/core/device.c
```diff
@@ -736,7 +736,7 @@ static bool device_is_ready(sd_device *dev) {
         if (r == 0)
                 log_device_debug(dev, "Device busy: SYSTEMD_READY property is false");
 
-        return r != 0;
+        return r > 0 || r == -ENOENT;
```
yuwata: Also here. Why do you change this?

yuwata (changes requested):
The last two commits sound questionable for me. ID_RENAMING or SYSTEMD_READY properties are optional properties. I do not think we should handle the failures as critical.

src/core/device.c
yuwata: Also, please downgrade this to debug.

src/core/device.c
daandemeyer: sd_device_new_from_devname() will return EINVAL which results in a lot of log noise since this function is called with quite a few nodes that are not device nodes. Skipping early here means we don't get log noise from these nodes.

src/core/device.c
daandemeyer: Why should this be a debug log? Unless we expect this to happen often, it seems to make more sense to log at error level so it's more noticeable when it happens

src/core/device.c
daandemeyer: At least in a mkosi QEMU VM with tun/tap device it never fires, but maybe that's not representative of all scenarios. Same for the other debug logs in this function.

src/core/device.c
daandemeyer: I figured it'd be good to not consider a device ready if an unexpected error happens, but if it might cause issues, better to not include it, I dropped the commit.

src/core/device.c
daandemeyer: Ditto, I dropped the commit.

src/core/device.c
daandemeyer: Downgraded the error logging in this function to warning instead

src/core/device.c
daandemeyer: Changed to not log when EINVAL is returned by sd-device.

poettering:
lgtm

src/core/device.c
yuwata: I thought this is a kind of library function. But OK, Let's see what happens.

src/core/device.c
yuwata: I guess this is (or will be) repeatedly called for the same device. But let's see how often it is logged.

PR #23309, opened 2022-05-08
basic: Add log context

daandemeyer:
This commit adds support for attaching extra metadata to log
messages written to the journal via log.h. We keep track of a
thread local log context in log.c onto which we can push extra
metadata fields that should be logged. Once a field is no longer
relevant, it can be popped again from the log context.

On top of this, we then add macros to allow pushing extra fields
onto the log context.

LOG_CONTEXT_PUSH() will push the provided field onto the log context
and pop the last field from the log context when the current block
ends. LOG_CONTEXT_PUSH_STRV() will do the same but for all fields in
the given strv.

Using the macros is as simple as putting them anywhere inside a block
to add a field to all following log messages logged from inside that
block.

```
void myfunction(...) {
	...

	LOG_CONTEXT_PUSH("MYMETADATA=abc");

	// Every journal message logged will now have the MYMETADATA=abc 
        // field included.
}
```

One special case to note is async code, where we use callbacks that
are invoked to continue processing when some event occurs. For async
code, there's usually an associated "userdata" struct containing
all the information associated with the async request. In this
"userdata" struct, we can add an extra "log_fields" field which is
a strv containing extra fields to be attached to log messages. At
the start of each callback, we can then call
"LOG_CONTEXT_PUSH_STRV(userdata->log_fields)" which makes sure the
fields are attached for the duration of the callback.

### DIFF
--- a/src/libsystemd/sd-device/device-monitor.c
+++ b/src/libsystemd/sd-device/device-monitor.c
@@ -242,10 +242,14 @@ _public_ int sd_device_monitor_stop(sd_device_monitor *m) {
 
 static int device_monitor_event_handler(sd_event_source *s, int fd, uint32_t revents, void *userdata) {
         _cleanup_(sd_device_unrefp) sd_device *device = NULL;
+        _unused_ _cleanup_(log_context_freep) LogContext *c = NULL;
         sd_device_monitor *m = ASSERT_PTR(userdata);
 
         if (device_monitor_receive_device(m, &device) <= 0)
                 return 0;
+
+        if (log_context_enabled())
+                c = log_context_new_consume(device_make_log_fields(device));
 
         if (m->callback)
                 return m->callback(m, device, m->userdata);
--- a/src/libsystemd/sd-device/device-util.c
+++ b/src/libsystemd/sd-device/device-util.c
@@ -5,6 +5,7 @@
 #include "devnum-util.h"
 #include "fd-util.h"
 #include "string-util.h"
+#include "strv.h"
 
 int devname_from_devnum(mode_t mode, dev_t devnum, char **ret) {
         _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
@@ -62,4 +63,79 @@ int device_open_from_devnum(mode_t mode, dev_t devnum, int flags, char **ret) {
         }
 
         return TAKE_FD(fd);
+}
+
+static int add_string_field(
+                sd_device *device,
+                const char *field,
+                int (*func)(sd_device *dev, const char **s),
+                char ***strv) {
+
+        const char *s;
+        int r;
+
+        assert(device);
+        assert(field);
+        assert(func);
+        assert(strv);
+
+        r = func(device, &s);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"%s\" property, ignoring: %m", field);
+        if (r >= 0)
+                (void) strv_extend_assignment(strv, field, s);
+
+        return 0;
+}
+
+char** device_make_log_fields(sd_device *device) {
+        _cleanup_strv_free_ char **strv = NULL;
+        dev_t devnum;
+        int ifindex;
+        sd_device_action_t action;
+        uint64_t seqnum, diskseq;
+        int r;
+
+        assert(device);
+
+        (void) add_string_field(device, "SYSPATH", sd_device_get_syspath, &strv);
+        (void) add_string_field(device, "SUBSYSTEM", sd_device_get_subsystem, &strv);
+        (void) add_string_field(device, "DEVTYPE", sd_device_get_devtype, &strv);
+        (void) add_string_field(device, "DRIVER", sd_device_get_driver, &strv);
+        (void) add_string_field(device, "DEVPATH", sd_device_get_devpath, &strv);
+        (void) add_string_field(device, "DEVNAME", sd_device_get_devname, &strv);
+        (void) add_string_field(device, "SYSNAME", sd_device_get_sysname, &strv);
+        (void) add_string_field(device, "SYSNUM", sd_device_get_sysnum, &strv);
+
+        r = sd_device_get_devnum(device, &devnum);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"DEVNUM\" property, ignoring: %m");
+        if (r >= 0)
+                (void) strv_extendf(&strv, "DEVNUM="DEVNUM_FORMAT_STR, DEVNUM_FORMAT_VAL(devnum));
+
+        r = sd_device_get_ifindex(device, &ifindex);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"IFINDEX\" property, ignoring: %m");
+        if (r >= 0)
+                (void) strv_extendf(&strv, "IFINDEX=%i", ifindex);
+
+        r = sd_device_get_action(device, &action);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"ACTION\" property, ignoring: %m");
+        if (r >= 0)
+                (void) strv_extendf(&strv, "ACTION=%s", device_action_to_string(action));
+
+        r = sd_device_get_seqnum(device, &seqnum);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"SEQNUM\" property, ignoring: %m");
+        if (r >= 0)
+                (void) strv_extendf(&strv, "SEQNUM=%"PRIu64, seqnum);
+
+        r = sd_device_get_diskseq(device, &diskseq);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"DISKSEQ\" property, ignoring: %m");
+        if (r >= 0)
+                (void) strv_extendf(&strv, "DISKSEQ=%"PRIu64, diskseq);
+
+        return TAKE_PTR(strv);
 }
--- a/src/libsystemd/sd-device/device-util.h
+++ b/src/libsystemd/sd-device/device-util.h
@@ -99,3 +99,5 @@ static inline int devname_from_stat_rdev(const struct stat *st, char **ret) {
         return devname_from_devnum(st->st_mode, st->st_rdev, ret);
 }
 int device_open_from_devnum(mode_t mode, dev_t devnum, int flags, char **ret);
+
+char** device_make_log_fields(sd_device *device);


src/basic/log.c
```diff
@@ -617,15 +623,15 @@ static int write_to_journal(
...
-                IOVEC_MAKE_STRING(header),
-                IOVEC_MAKE_STRING("MESSAGE="),
-                IOVEC_MAKE_STRING(buffer),
-                IOVEC_MAKE_STRING("\n"),
-        };
+        log_context[0] = IOVEC_MAKE_STRING(header);
+        log_context[1] = IOVEC_MAKE_STRING("MESSAGE=");
+        log_context[2] = IOVEC_MAKE_STRING(buffer);
+        log_context[3] = IOVEC_MAKE_STRING("\n");
+        /* Make sure to check if LOG_CONTEXT_RESERVED is large enough when adding more fields here. */
```
bluca: can this be done with a cc_assert?

DaanDeMeyer:
Removed the CONSUME versions of the macros as they didn't allow for proper OOM handling.

DaanDeMeyer:
Reworked implementation to make sure we add the extra metadata everywhere. All stack allocated iovecs are now replaced by use of the log context iovec array to avoid having to copy iovecs around.

poettering:
Hmpf. I really dislike using TLS for things like this. TLS is costly, in particular if you stick an array in it, and releasing it is cumbersome.

I mean, i see why this is useful, but can't we do this without tls? for example using block based stack allocations. i.e. have something like this:


```c
struct log_context {
        const char *field;
        struct log_context *prev;
};

#define LOG_CONTEXT_BEGIN \
        struct log_context *_log_context = NULL;

#define LOG_CONTEXT_ADD(s) \
        _log_context = &(struct log_context) { .field = s, .prev = _log_context }
```

Then you could do: 

```c
int foo(void) {
        LOG_CONTEXT_BEGIN;

        LOG_CONTEXT_ADD("foo=bar");
        LOG_CONTEXT_ADD("wuff=kkk");

        log_info("ydfkjhdlkfhdskjf"); /* this macro would then have to pick up _log_context as parameter implicitly */
}
```

next step would then be to change LOG_CONTEXT_ADD() to also use `_cleanup_` stuff to automatically remove an item from the list when the current block is left, so that this would work:

```c
int foo(void) {
        LOG_CONTEXT_BEGIN;

        LOG_CONTEXT_ADD("foo=bar");
        LOG_CONTEXT_ADD("wuff=kkk");

        for (;;) {
                _cleanup_free_ char *p = NULL;
                asprintf(&p, "random=%" PRIx64, random_u64());
                LOG_CONTEXT_ADD(p);
                
                log_info("something happened");
        }
```

poettering:
Hmpf. I really dislike using TLS for things like this. TLS is costly, in particular if you stick an array in it, and releasing it is cumbersome.

I mean, i see why this is useful, but can't we do this without tls? for example using block based stack allocations. i.e. have something like this:


```c
struct log_context {
        const char *field;
        struct log_context *prev;
};

#define LOG_CONTEXT_BEGIN \
        struct log_context *_log_context = NULL;

#define LOG_CONTEXT_ADD(s) \
        _log_context = &(struct log_context) { .field = s, .prev = _log_context }
```

Then you could do: 

```c
int foo(void) {
        LOG_CONTEXT_BEGIN;

        LOG_CONTEXT_ADD("foo=bar");
        LOG_CONTEXT_ADD("wuff=kkk");

        log_info("ydfkjhdlkfhdskjf"); /* this macro would then have to pick up _log_context as parameter implicitly */
}
```

next step would then be to change LOG_CONTEXT_ADD() to also use `_cleanup_` stuff to automatically remove an item from the list when the current block is left, so that this would work:

```c
int foo(void) {
        LOG_CONTEXT_BEGIN;

        LOG_CONTEXT_ADD("foo=bar");
        LOG_CONTEXT_ADD("wuff=kkk");

        for (;;) {
                _cleanup_free_ char *p = NULL;
                asprintf(&p, "random=%" PRIx64, random_u64());
                LOG_CONTEXT_ADD(p);
                
                log_info("something happened");
        }
```

poettering:
i'd be more OK with using TLS if we only store the anchor of such a linked list in it... i.e. singular fields that require no cleanup should be ok.

poettering:
i'd be more OK with using TLS if we only store the anchor of such a linked list in it... i.e. singular fields that require no cleanup should be ok.

DaanDeMeyer:
Hmm, that's a neat approach, let me try that out, I think it should work.

DaanDeMeyer:
Re-implemented using the approach suggested by @poettering. PTAL!

src/basic/log.h
```diff
@@ -367,3 +367,37 @@ int log_syntax_invalid_utf8_internal(
...
+
+struct log_context {
+        char *const *fields;
+        struct log_context *prev;
+};
+
+extern thread_local struct log_context *_log_context;
+
+static inline void _log_context_pop(struct log_context **c) {
+        if (!*c)
```
bluca: `if (!c || !*c)`

src/test/test-log.c
```diff
@@ -70,6 +70,25 @@ static void test_log_syntax(void) {
...
 }
 
+static void test_log_context(void) {
+        char *strv[] = { (char*) "MYDATA=abc", NULL };
+
+        {
+                LOG_CONTEXT_PUSH("MYDATA=abc");
+                LOG_CONTEXT_PUSH("MYDATA=def");
+                LOG_CONTEXT_PUSH_STRV(strv);
+                LOG_CONTEXT_PUSH_STRV(strv);
```
bluca, line 80: Should it work if calling a function after this line, and doing more LOG_CONTEXT_PUSH from there? If so, would probably be good to cover that case too in the test

src/basic/log.h
daandemeyer: We don't do this check either in `DEFINE_TRIVIAL_CLEANUP_FUNC()` so I left it out here as well. I think it's guaranteed c itself won't be `NULL`?

src/test/test-log.c
daandemeyer: Extended the test a bit more in general. Went for a nested block since that's pretty much the same as a function in this scenario.

src/basic/log.h
```diff
@@ -367,3 +367,37 @@ int log_syntax_invalid_utf8_internal(
 #define DEBUG_LOGGING _unlikely_(log_get_max_level() >= LOG_DEBUG)
 
 void log_setup(void);
+
+struct log_context {
+        char *const *fields;
+        struct log_context *prev;
```
bluca: we have linked list stuff already in src/basic/list.h, can that be used instead?

src/basic/log.h
bluca: force of habit... doesn't cost much to check, and it's a bit friendlier to onlookers who might not know what is guaranteed and what is not - in fact if it's guaranteed, then an assert() to assert it would be a nice way to document it clearly

src/basic/log.h
```diff
@@ -367,3 +368,37 @@ int log_syntax_invalid_utf8_internal(
 #define DEBUG_LOGGING _unlikely_(log_get_max_level() >= LOG_DEBUG)
 
 void log_setup(void);
+
```
bluca: Maybe add a brief comment here, explaining what's this useful for, and how to use it (maybe explicitly mention that the caller owns any memory buffer that gets pushed, and popping doesn't free them)?

daandemeyer: Added a blurb

src/basic/log.c
```diff
@@ -607,15 +632,19 @@ static int write_to_journal(
 
         log_do_header(header, sizeof(header), level, error, file, line, func, object_field, object, extra_field, extra);
 
-        struct iovec iovec[4] = {
+        struct iovec iovec[JOURNAL_MAX_ENTRIES] = {
```
keszybz: Does JOURNAL_MAX_ENTRIES need to be so big? `struct iovec` is 16 bytes, so we're allocating 16k here on the stack.

src/basic/log.c
```diff
@@ -588,6 +592,27 @@ static int log_do_header(
         return 0;
 }
 
+static void log_do_context(struct iovec *iovec, size_t iovec_len, size_t *n) {
+        assert(iovec);
+        assert(n);
+
+        if (iovec_len <= 1)
+                return;
```
keszybz: Is this check needed? We don't expect such short iovec, and the check below should handle it if we get one.

src/basic/log.c
```diff
@@ -588,6 +592,27 @@ static int log_do_header(
...
+                STRV_FOREACH(s, _log_context->fields) {
+                        if (*n >= iovec_len - 1)
+                                break;
+
+                        iovec[(*n)++] = IOVEC_MAKE_STRING(*s);
+                        iovec[(*n)++] = IOVEC_MAKE_STRING("\n");
+                }
+
+                if (*n >= iovec_len - 1)
+                        break;
```
keszybz: This repeated check is annoying. Why not just drop the second check and `return` from the first? We don't expect any items in this list with empty fields, so we don't need to optimize for that.

src/basic/log.c
```diff
@@ -588,6 +592,27 @@ static int log_do_header(
...
 
+static void log_do_context(struct iovec *iovec, size_t iovec_len, size_t *n) {
+        assert(iovec);
+        assert(n);
+
+        if (iovec_len <= 1)
+                return;
+
+        LIST_FOREACH(ll, c, _log_context) {
+                STRV_FOREACH(s, _log_context->fields) {
```
keszybz: I'm afraid this needs to be `c->fields` ;(

Maybe add a test that this logging actually works in TEST-04-JOURNAL ?

src/basic/log.h
```diff
@@ -367,3 +368,68 @@ int log_syntax_invalid_utf8_internal(
...
+        LIST_PREPEND(ll, _log_context, c)
+
+#define LOG_CONTEXT_PUSH(s) \
+        _LOG_CONTEXT_PUSH(s, UNIQ_T(c, UNIQ))
+
+#define _LOG_CONTEXT_PUSH_STRV(strv, c)                                             \
+        _cleanup_(_log_context_pop) struct log_context *c = &(struct log_context) { \
+                .fields = (strv),                                                   \
+        };                                                                          \
+        LIST_PREPEND(ll, _log_context, c)
```
keszybz: Hmm, do we need the separate version for strv? Do you expect any uses where the strv would be passed from outside?
I'd imagine that the context is always constructed locally, so we could have just a single macro LOG_CONTEXT_PUSH() which accepts one or more strings and calls STRV_MAKE() on them? (Also, STRV_MAKE should be used above…)

keszybz (changes requested):

src/basic/log.c
daandemeyer: I copied the max amount of entries that a single journal message can contain. We can certainly lower this though. I'll put in a lower value

src/basic/log.h
daandemeyer: I explained this a bit in the comment, but I figure if we want to use this for e.g. a resolved DNS request, and we want to retain context across asynchronous calls, we'd gather the context fields in a strv, store that strv in some `userdata` struct and push it onto the context at the start of each callback associated with that DNS request. That's the primary use case I had in mind for this macro.

src/basic/log.c
daandemeyer: Hmm, let me extend systemd-cat with support for adding extra fields and use this new functionality there immediately. Then I can use that in TEST-04-JOURNAL to test that this works.

src/basic/log.h
keszybz: Ah, OK. I remember reading that, but it was a while ago… But in that case, shouldn't LOG_CONTEXT_PUSH just be a wrapper for `LOG_CONTEXT_PUSH_STRV(STRV_MAKE(__VARARGS__))` ?

daandemeyer: That makes sense, I'll rework this to do that

src/basic/log.c
daandemeyer: Lowered to 128. This gives us space for 64 fields

src/test/test-log.c
```diff
@@ -70,6 +70,54 @@ static void test_log_syntax(void) {
         assert_se(log_syntax("unit", LOG_ERR, "filename", 10, SYNTHETIC_ERRNO(ENOTTY), "ENOTTY: %s: %m", "hogehoge") == -ENOTTY);
 }
 
+static void test_log_context(void) {
+        char *strv[] = { (char*) "MYDATA=abc", NULL };
```
poettering: STRV_MAKE?

maybe add multiple entries here, to actually use the fact an strv is accepted here

src/basic/log.c
```diff
@@ -32,12 +32,14 @@
...
 #include "string-table.h"
 #include "string-util.h"
+#include "strv.h"
 #include "syslog-util.h"
 #include "terminal-util.h"
 #include "time-util.h"
 #include "utf8.h"
 
 #define SNDBUF_SIZE (8*1024*1024)
+#define MAX_IOVEC 128
```
poettering: usually we put the MAX last, i.e. IOVEC_MAX?

poettering:
looks ok to me, just some minor points...

src/basic/log.c
poettering: btw, why do this? why not instead a counter also as thread_local variable that encodes how many entries are needed? it's increased whenever a context is added, and decreased whenever one is removed.

daandemeyer: To avoid unbounded stack allocations, I figure it makes sense to impose an upper limit on the amount of stuff we put into the iovec. If we already have a max size, tracking the size of the log context linked list individually doesn't seem like it provides much benefit

src/basic/list.h
```diff
@@ -1,8 +1,6 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 #pragma once
 
-#include "macro.h"
```
yuwata, line 4: Why this is dropped?

src/basic/log.c
```diff
@@ -588,6 +592,21 @@ static int log_do_header(
         return 0;
 }
 
+static void log_do_context(struct iovec *iovec, size_t iovec_len, size_t *n) {
+        assert(iovec);
+        assert(n);
+
+        LIST_FOREACH(ll, c, _log_context) {
```
yuwata: You can drop `{`.

src/basic/log.c
```diff
@@ -588,6 +592,21 @@ static int log_do_header(
...
+                STRV_FOREACH(s, c->fields) {
+                        if (*n >= iovec_len - 1)
```
yuwata: This should be `-2`??

src/basic/log.c
```diff
@@ -1045,16 +1071,24 @@ int log_struct_iovec_internal(
                 char header[LINE_MAX];
                 log_do_header(header, sizeof(header), level, error, file, line, func, NULL, NULL, NULL, NULL);
 
-                struct iovec iovec[1 + n_input_iovec*2];
-                iovec[0] = IOVEC_MAKE_STRING(header);
+                /* Each input iovec translates to 2 iovecs in our iovec array because we need to add a newline. */
+                if (n_input_iovec > IOVEC_MAX / 2)
```
yuwata: `if (n_input_iovec * 2 + 1 > IOVEC_MAX)`??

src/basic/log.h
```diff
@@ -8,6 +8,7 @@
 #include <syslog.h>
 
 #include "macro.h"
+#include "list.h"
```
yuwata, line 10: Please sort the headers.

src/basic/log.h
```diff
@@ -367,3 +368,62 @@ int log_syntax_invalid_utf8_internal(
...
+        assert(c);
+
+        if (!*c)
+                return;
+
+        LIST_REMOVE(ll, _log_context, *c);
+}
+
+#define LOG_CONTEXT_PUSH(...) \
+        LOG_CONTEXT_PUSH_STRV(STRV_MAKE(__VA_ARGS__))
```
yuwata, line 476: CONST?

yuwata (changes requested):

src/basic/list.h
daandemeyer: We get compilation failures if we leave it in:

```
➜  systemd git:(log-context) ninja -C build
ninja: Entering directory `build'
[165/2017] Compiling C object src/libsystemd/libsystemd_static.a.p/sd-netlink_netlink-types.c.o
FAILED: src/libsystemd/libsystemd_static.a.p/sd-netlink_netlink-types.c.o 
cc -Isrc/libsystemd/libsystemd_static.a.p -Isrc/libsystemd -I../src/libsystemd -Isrc/basic -I../src/basic -Isrc/fundamental -I../src/fundamental -Isrc/systemd -I../src/systemd -I. -I.. -I../src/libsystemd/sd-bus -I../src/libsystemd/sd-device -I../src/libsystemd/sd-event -I../src/libsystemd/sd-hwdb -I../src/libsystemd/sd-id128 -I../src/libsystemd/sd-journal -I../src/libsystemd/sd-netlink -I../src/libsystemd/sd-network -I../src/libsystemd/sd-resolve -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=gnu11 -O0 -g -Wno-format-signedness -Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wendif-labels -Werror=format=2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=overflow -Werror=override-init -Werror=return-type -Werror=shift-count-overflow -Werror=shift-overflow=2 -Werror=undef -Wfloat-equal -Wimplicit-fallthrough=5 -Winit-self -Wlogical-op -Wmissing-include-dirs -Wmissing-noreturn -Wnested-externs -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wshadow -Wstrict-aliasing=2 -Wstrict-prototypes -Wsuggest-attribute=noreturn -Wunused-function -Wwrite-strings -Wno-unused-result -Werror=missing-declarations -Werror=missing-prototypes -fdiagnostics-show-option -fno-common -fno-strict-aliasing -fstack-protector -fstack-protector-strong -fvisibility=hidden --param=ssp-buffer-size=4 -fno-omit-frame-pointer -Werror=shadow -include config.h -fPIC -pthread -fvisibility=default -MD -MQ src/libsystemd/libsystemd_static.a.p/sd-netlink_netlink-types.c.o -MF src/libsystemd/libsystemd_static.a.p/sd-netlink_netlink-types.c.o.d -o src/libsystemd/libsystemd_static.a.p/sd-netlink_netlink-types.c.o -c ../src/libsystemd/sd-netlink/netlink-types.c
In file included from ../src/basic/macro.h:452,
                 from ../src/basic/list.h:4,
                 from ../src/libsystemd/sd-netlink/netlink-internal.h:8,
                 from ../src/libsystemd/sd-netlink/netlink-types.c:6:
../src/basic/log.h:405:9: error: expected specifier-qualifier-list before ‘LIST_FIELDS’
  405 |         LIST_FIELDS(struct log_context, ll);
      |         ^~~~~~~~~~~
../src/basic/log.h:408:51: error: unknown type name ‘_log_context’
  408 | extern thread_local LIST_HEAD(struct log_context, _log_context);
      |                                                   ^~~~~~~~~~~~
../src/basic/log.h: In function ‘_log_context_pop’:
../src/basic/log.h:416:9: error: implicit declaration of function ‘LIST_REMOVE’ [-Werror=implicit-function-declaration]
  416 |         LIST_REMOVE(ll, _log_context, *c);
      |         ^~~~~~~~~~~
../src/basic/log.h:416:9: warning: nested extern declaration of ‘LIST_REMOVE’ [-Wnested-externs]
../src/basic/log.h:416:21: error: ‘ll’ undeclared (first use in this function)
  416 |         LIST_REMOVE(ll, _log_context, *c);
      |                     ^~
../src/basic/log.h:416:21: note: each undeclared identifier is reported only once for each function it appears in
../src/basic/log.h:416:25: error: ‘_log_context’ undeclared (first use in this function); did you mean ‘log_context’?
  416 |         LIST_REMOVE(ll, _log_context, *c);
      |                         ^~~~~~~~~~~~
      |                         log_context
cc1: some warnings being treated as errors
[174/2017] Compiling C object src/libsystemd/libsystemd_static.a.p/sd-resolve_sd-resolve.c.o
```

src/basic/log.c
daandemeyer: I went for `n_input_iovec * 2 >= IOVEC_MAX)` which should be equivalent

src/basic/list.h
yuwata: How about to include macro.h at the end of the file, like macro.h includes logs.h? Otherwise, we cannot safely use the macros defined in list.h without separately including macro.h.

yuwata (approved):
LGTM.

poettering:
So this PR currently doesn't port anything over to the new interface. Can you maybe add that? We should generally have at least one user of the stuff we have here.

DaanDeMeyer:
Reworked the interface a bit, and added a few things that I figured out would be nice while trying to convert some code to use this. Still need to push the code that makes some part of the codebase use this feature.

- Added LOG_CONTEXT_PUSHF() to add a field via formatting
- We now expose the underlying log context struct (renamed to LogContext) to make the entire feature more flexible overall. For example, we can now write a function `log_context_push_device()` that creates a new log context containing all the fields relevant to a `sd_device` instance.

src/basic/stdio-util.h
```diff
@@ -9,13 +9,18 @@
...
 #include "memory-util.h"
 
-#define snprintf_ok(buf, len, fmt, ...)                                \
-        ({                                                             \
-                char *_buf = (buf);                                    \
-                size_t _len = (len);                                   \
-                int _snpf = snprintf(_buf, _len, (fmt), __VA_ARGS__);  \
-                _snpf >= 0 && (size_t) _snpf < _len ? _buf : NULL;     \
-        })
+static inline char *snprintf_ok(char *buf, size_t len, const char *format, ...) {
```
yuwata, line 12: I think it is better to add `_printf_` attribute.

yuwata (changes requested):

src/basic/log.c
```diff
@@ -1478,3 +1511,35 @@ void log_setup(void) {
...
+        LIST_REMOVE(ll, _log_context, c);
+        return NULL;
+}
+
+LogContext* log_context_new(void) {
+        LogContext *c = new(LogContext, 1);
+        if (!c)
+                return NULL;
+
+        *c = (LogContext) {};
```
yuwata: Please use `new0()`.

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +114,83 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+
+        assert(device);
+        assert(field);
+        assert(func);
+
+        r = func(device, &s);
+        if (r < 0 && r != -ENOENT)
+                log_error_errno(r, "Failed to get device \"%s\" property: %m", field);
+
+        if (r == 0)
```
yuwata: Please use `r >= 0`.

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +114,83 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
 
         return false;
 }
+
+static int add_string_field(
+                sd_device *device,
+                const char *field,
+                int (*func)(sd_device *dev, const char **s),
+                char ***ret) {
```
yuwata: `ret` here is also an input. Please rename it.

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +114,83 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+
+        const char *s;
+        int r;
+
+        assert(device);
+        assert(field);
+        assert(func);
```
yuwata, line 79: Missing assertion for `ret`.

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +114,83 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+
+        r = func(device, &s);
+        if (r < 0 && r != -ENOENT)
+                log_error_errno(r, "Failed to get device \"%s\" property: %m", field);
```
yuwata: Please downgrade error and mention that the failure is ignored.

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +114,83 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+        (void) add_string_field(device, "DEVNAME", sd_device_get_devname, &strv);
+        (void) add_string_field(device, "SYSNAME", sd_device_get_sysname, &strv);
+        (void) add_string_field(device, "SYSNUM", sd_device_get_sysnum, &strv);
+
+        r = sd_device_get_devnum(device, &devnum);
+        if (r < 0 && r != -ENOENT)
+                log_error_errno(r, "Failed to get device \"DEVNUM\" property: %m");
+
+        if (r == 0)
+                (void) strv_extendf(&strv, "DEVNUM=%u:%u", major(devnum), minor(devnum));
```
yuwata: Nowadays we have macro for formatting devnum.

yuwata (changes requested):

src/core/device.c
```diff
@@ -1083,6 +1083,8 @@ static int device_dispatch_io(sd_device_monitor *monitor, sd_device *dev, void *
 
         assert(dev);
 
+        LOG_CONTEXT_PUSH_DEVICE(dev);
```
daandemeyer: I wonder if we should move this into sd-device and call it just before we invoke the handler. That's an easy way to apply it everywhere. Since this macro doesn't do any logging itself, it should be OK to add to sd-device, we would have to degrade the warning logs to debug though.

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +115,84 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+        int r;
+
+        assert(device);
+        assert(field);
+        assert(func);
+        assert(strv);
+
+        r = func(device, &s);
+        if (r < 0 && r != -ENOENT)
+                log_debug_errno(r, "Failed to get device \"%s\" property, ignoring: %m", field);
```
yuwata: `log_device_debug_errno()`?

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +115,84 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+        (void) add_string_field(device, "DEVTYPE", sd_device_get_devtype, &strv);
+        (void) add_string_field(device, "DRIVER", sd_device_get_driver, &strv);
+        (void) add_string_field(device, "DEVPATH", sd_device_get_devpath, &strv);
+        (void) add_string_field(device, "DEVNAME", sd_device_get_devname, &strv);
+        (void) add_string_field(device, "SYSNAME", sd_device_get_sysname, &strv);
+        (void) add_string_field(device, "SYSNUM", sd_device_get_sysnum, &strv);
+
+        r = sd_device_get_devnum(device, &devnum);
+        if (r < 0 && r != -ENOENT)
+                log_debug_errno(r, "Failed to get device \"DEVNUM\" property, ignoring: %m");
```
yuwata: Also here, `log_device_debug_errno()` is better?

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +115,84 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+                (void) strv_extendf(&strv, "SEQNUM=%"PRIu64, seqnum);
+
+        r = sd_device_get_diskseq(device, &diskseq);
+        if (r < 0 && r != -ENOENT)
+                log_debug_errno(r, "Failed to get device \"DISKSEQ\" property, ignoring: %m");
+
+        if (r >= 0)
+                (void) strv_extendf(&strv, "DISKSEQ=%"PRIu64, diskseq);
+
+        return TAKE_PTR(strv);
```
yuwata, line 140: How about to cash the strv in sd-device object? Then, we can enable/disable log context frequently with almost no performance regression.

src/libsystemd/sd-device/device-monitor.c
```diff
@@ -245,6 +245,8 @@ static int device_monitor_event_handler(sd_event_source *s, int fd, uint32_t rev
         if (device_monitor_receive_device(m, &device) <= 0)
                 return 0;
 
+        LOG_CONTEXT_CONSUME_STRV(device_log_context_fields(device));
```
yuwata: Setting the log context here may be too early, as an uevent may trigger very long operation.

yuwata (changes requested):

src/libsystemd/sd-device/device-monitor.c
daandemeyer: I don't understand, this is right before we invoke the callback no? If this is too early, where would you suggest we set the log context for the device?

yuwata: In other words, the usage of log context I thought was that each sd-device object manages its log context for the device, and push and pop within single call of `log_device_xxx()`. Then, log messages unrelated to the device are not polluted by the log context.

But now, you propose all logs during a uevent attaches the log context about the received device. That might help in some case. But when we widely use the log context for variety of objects, then each log messages has toooo many fields...

daandemeyer: The device log context is only valid for the duration of the callback though. And I figure that during the callback for a specific device, it makes sense to attach the log context fields related to the device to every log message logged during the callback.

This makes it very easy to show all log messages related to the device by simply doing `journalctl DEVNAME=<devname>`.

The idea is that during a device monitor callback for a device X, even when we're calling into code that isn't aware of the device being processed, we can still attach the necessary context to those log messages to be able to figure out later that they were logged when device X was being processed.

To be able to only show the fields you're interested in, I've been working on https://github.com/systemd/systemd/pull/23635 which will allow to format the journal output to only show the fields the user is interested in.

As for disk space constraints, the journal deduplicates identical fields so logging the same fields many times for many log messages is cheap.

src/libsystemd/sd-device/device-util.c
daandemeyer: See the other discussion, I don't think we'll be enabling/disabling the log context very frequently, so I'd like to leave this optimization until we find a use case where we're enabling/disabling the same log context very frequently.

src/libsystemd/sd-bus/sd-bus.c
```diff
@@ -2933,6 +2933,7 @@ static int process_message(sd_bus *bus, sd_bus_message *m) {
         bus->current_message = m;
         bus->iteration_counter++;
 
+        LOG_CONTEXT_CONSUME_STRV(bus_message_log_context_fields(m));
```
bluca: should this happen only if log level is debug?

bluca: the message below is logged only at debug level, right? so that would avoid doing the processing for nothing

daandemeyer: This doesn't only apply to the debug log message below. During `process_message()`, all the callbacks registered in the vtable for the service get called as well.

The idea of adding the fields to the context here is to add these fields to any log message that's logged from those callbacks, so that later, we can easily query for all the log messages that happened during processing of a specific D-Bus method by adding a `DBUS_METHOD=Reload` match to journalctl for example.

bluca: ah got it, thanks

bluca (approved):

src/basic/list.h
```diff
@@ -28,22 +28,23 @@
...
+        ({                                                              \
                 typeof(*(head)) **_hhead = &(head), *_tail;             \
-                LIST_FIND_TAIL(name, *_hhead, _tail);                   \
+                _tail = LIST_FIND_TAIL(name, *_hhead);                  \
                 LIST_INSERT_AFTER(name, *_hhead, _tail, item);          \
-        } while (false)
+        })
 
 /* Remove an item from the list */
 #define LIST_REMOVE(name,head,item)                                     \
```
poettering, line 48: hmm, let's do the conversion comprehensively, i.e. also for LIST_REMOVE() and everything else in this header. Always return the item that is added/removed

src/basic/log.c
```diff
@@ -588,6 +592,20 @@ static int log_do_header(
         return 0;
 }
 
+static void log_do_context(struct iovec *iovec, size_t iovec_len, size_t *n) {
+        assert(iovec);
+        assert(n);
+
+        LIST_FOREACH(ll, c, _log_context)
+                STRV_FOREACH(s, c->fields) {
+                        if (*n >= iovec_len - 2)
```
poettering: might underflow if iovec_len is smaller than 2?

add assert? or other way to handle corner case?

src/basic/log.c
```diff
@@ -607,15 +625,19 @@ static int write_to_journal(
 
         log_do_header(header, sizeof(header), level, error, file, line, func, object_field, object, extra_field, extra);
 
-        struct iovec iovec[4] = {
+        struct iovec iovec[IOVEC_MAX] = {
```
poettering: that's quite an increase. i.e. 32 byte before. 1K now. in a world where 4K stacks are often used... For a log function that should try to be as little intrusive as we can be.

maybe count how much we actually need, and size the thing properly?

src/basic/log.h
```diff
@@ -367,3 +369,68 @@ int log_syntax_invalid_utf8_internal(
...
+ * One special case to note is async code, where we use callbacks that are invoked to continue processing
+ * when some event occurs. For async code, there's usually an associated "userdata" struct containing all the
+ * information associated with the async operation. In this "userdata" struct, we can store a log context
+ * allocated with log_context_new() and freed with log_context_free(). We can then add and remove fields to
+ * the `fields` member of the log context object and all those fields will be logged along with each log
+ * message.
+ */
+
+typedef struct LogContext {
+        char **fields;
```
poettering: Can you add a comment here: `/* depending on which destructor is used (log_context_free() or log_context_detach()) the memory referenced by this is freed or not */` or so?

If there are two ways memory is owned there should always be a specific note about it where the field for it is declared.

src/basic/stdio-util.h
```diff
@@ -8,6 +8,7 @@
 
 #include "macro.h"
 
+_printf_(3, 4)
 static inline char *snprintf_ok(char *buf, size_t len, const char *format, ...) {
```
poettering, line 12: should probably be a separate commit

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +115,84 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+        (void) add_string_field(device, "DRIVER", sd_device_get_driver, &strv);
+        (void) add_string_field(device, "DEVPATH", sd_device_get_devpath, &strv);
+        (void) add_string_field(device, "DEVNAME", sd_device_get_devname, &strv);
+        (void) add_string_field(device, "SYSNAME", sd_device_get_sysname, &strv);
+        (void) add_string_field(device, "SYSNUM", sd_device_get_sysnum, &strv);
+
+        r = sd_device_get_devnum(device, &devnum);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"DEVNUM\" property, ignoring: %m");
+
```
poettering: drop this empty line given you are still processing the same fuction return code here

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +115,84 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+        if (r >= 0)
+                (void) strv_extendf(&strv, "DEVNUM="DEVNUM_FORMAT_STR, DEVNUM_FORMAT_VAL(devnum));
+
+        r = sd_device_get_ifindex(device, &ifindex);
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"IFINDEX\" property, ignoring: %m");
+
```
poettering: ditto and everywhere else

src/libsystemd/sd-device/device-util.c
```diff
@@ -113,3 +115,84 @@ bool device_match_parent(sd_device *device, Set *match_parent, Set *nomatch_pare
...
+        if (r < 0 && r != -ENOENT)
+                log_device_debug_errno(device, r, "Failed to get device \"%s\" property, ignoring: %m", field);
+
+        if (r >= 0)
+                (void) strv_extendf(strv, "%s=%s", field, s);
+
+        return 0;
+}
+
+char** device_log_context_fields(sd_device *device) {
```
poettering: maybe call this `device_make_log_fields()`?

src/basic/log.h
```diff
@@ -367,3 +369,68 @@ int log_syntax_invalid_utf8_internal(
...
+        _LOG_CONTEXT_PUSH_STRV(strv, UNIQ_T(c, UNIQ))
+
+/* LOG_CONTEXT_CONSUME_STRV() is identical to LOG_CONTEXT_PUSH_STRV() except it takes ownership of the given
+ * strv argument. */
+
+#define _LOG_CONTEXT_CONSUME_STRV(strv, c) \
+        _unused_ _cleanup_(log_context_freep) LogContext *c = log_context_new(strv);
+
+#define LOG_CONTEXT_CONSUME_STRV(strv) \
+        _LOG_CONTEXT_CONSUME_STRV(strv, UNIQ_T(c, UNIQ))
```
poettering, line 504: we really should have `LOG_CONTET_CONSUME_STR()` too, no? how would we do that?

src/test/test-log.c
```diff
@@ -70,6 +71,85 @@ static void test_log_syntax(void) {
...
+                assert(i == 1);
+
+                /* Test that everything still works with a manually configured log context. */
+                test_log_struct();
+                test_long_lines();
+                test_log_syntax();
+        }
+
+        {
+                char **strv = strv_new("ABC", "DEF");
```
poettering: assert_se()

src/test/test-log.c
```diff
@@ -70,6 +71,85 @@ static void test_log_syntax(void) {
...
+                LIST_FOREACH(ll, c, _log_context)
+                        i++;
+                assert(i == 4);
+        }
+
+        /* Check that the log context is cleaned up correctly. */
+        assert(_log_context == NULL);
+
+        {
+                _cleanup_(log_context_freep) LogContext *ctx = log_context_new(NULL);
```
poettering: assert_se()

src/libsystemd/sd-bus/bus-message.c
```diff
@@ -4649,3 +4649,42 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
...
 }
+
+char** bus_message_log_context_fields(sd_bus_message *m) {
+        _cleanup_strv_free_ char **strv = NULL;
+        const char *s;
+
+        assert(m);
+
+        if ((s = bus_message_type_to_string(m->header->type)))
+                (void) strv_extendf(&strv, "DBUS_MESSAGE_TYPE=%s", s);
```
poettering: maybe let's add to strv.h a helper:

```c
int strv_extend_assignment(char ***l, const char *lhs, const char *rhs) {
        assert(l);
        assert(lhs);

        if (!rhs) /* value is optional, in which case we suppress the field */
                  return 0;
    
        j = strjoin(lhs, "=", "rhs");
        if (!j)
                return NULL;

         return strv_consume(l, j); 
}
``

File: src/libsystemd/sd-bus/bus-message.c [hunk 44]
```diff
@@ -4649,3 +4649,42 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
...
+                (void) strv_extendf(&strv, "DBUS_PATH=%s", s);
+
+        if ((s = sd_bus_message_get_interface(m)))
+                (void) strv_extendf(&strv, "DBUS_INTERFACE=%s", s);
+
+        if ((s = sd_bus_message_get_member(m)))
+                (void) strv_extendf(&strv, "DBUS_MEMBER=%s", s);
+
+        (void) strv_extendf(&strv, "DBUS_MESSAGE_COOKIE=" PRIu64, BUS_MESSAGE_COOKIE(m));
+        (void) strv_extendf(&strv, "DBUS_MESSAGE_REPLY_COOKIE=" PRIu64, m->reply_cookie);
```
poettering: reply_cookie might not be set, check for that.

src/basic/log.c
daandemeyer: Added counting, but kept the upper limit just in case

src/basic/log.h
daandemeyer: Added `LOG_CONTEXT_CONSUME_STR()` using `strv_new()`.

Maybe we should have `strv_ops` that we can pass to `strv_free()` to indicate how the strv should be freed.

src/basic/stdio-util.h
```diff
@@ -9,13 +9,19 @@
...
-                size_t _len = (len);                                   \
-                int _snpf = snprintf(_buf, _len, (fmt), __VA_ARGS__);  \
-                _snpf >= 0 && (size_t) _snpf < _len ? _buf : NULL;     \
-        })
+_printf_(3, 4)
+static inline char *snprintf_ok(char *buf, size_t len, const char *format, ...) {
+        va_list ap;
+        int r;
+
+        va_start(ap, format);
```
poettering, line 16: hmm, i wonder how the generated code here actually compares to the status quo ante? previously this would just be an invocation of snprintf(). Now it packs stuff on the stack for the va_list stuff I presume, and then calls vsnprintf(). i.e. invocation via registers got changed to invocation via stack. might be something to look into compiler explorer for

What's the rationale for this change?

src/basic/strv.c
```diff
@@ -904,6 +904,22 @@ int strv_extend_n(char ***l, const char *value, size_t n) {
         return -ENOMEM;
 }
 
+int strv_extend_assignment(char ***l, const char *lhs, const char *rhs) {
```
poettering, line 792: this smells a lot like strv_env_assign()...

I wonder if it wouldn't be better to make this new helper more generic, i.e. call this strv_extend_join() that combined strjoin() + strv_extend(). You'd thus call:

```c
r = strv_extend_join(&l, "foo", "=", "bar");
```

I think that would be almost as good to read, less redundant given that strv_env_assign() already exists, and I am sure we can find various other places to use this.

src/basic/list.h
```diff
@@ -190,3 +188,5 @@
                         LIST_REMOVE(name, *_a, _p);                     \
                 _p;                                                     \
         })
+
+#include "macro.h"
```
poettering, line 194: hmm, this feels like a hack, but i got no better idea atm.

But please add a comment why this is needed. i.e something like this:

```c
/* Now include "macro.h", because we want our definition of assert() which the macros above use. We include it down here instead of up top, since macro.h pulls in log.h which in turn needs our own definitions. */
```

or something like that

src/basic/log.c
```diff
@@ -588,6 +592,21 @@ static int log_do_header(
...
 }
 
+static void log_do_context(struct iovec *iovec, size_t iovec_len, size_t *n) {
+        assert(iovec);
+        assert(iovec_len >= 2);
+        assert(n);
+
+        LIST_FOREACH(ll, c, _log_context)
+                STRV_FOREACH(s, c->fields) {
+                        if (*n >= iovec_len - 2)
```
poettering: maybe write as `if (*n + 2 >= iovec_len)` to avoid the underflow and the assert check?

src/basic/log.c
```diff
@@ -1478,3 +1516,47 @@ void log_setup(void) {
...
+
+size_t log_context_num_fields(LogContext *context) {
+        size_t num_fields = 0;
+
+        LIST_FOREACH(ll, c, context)
+                STRV_FOREACH(s, c->fields)
+                        num_fields++;
+
+        return num_fields;
+}
```
poettering, line 1598: hmm, if we have large numbers of fields this likely will get quite expensive... a static counter next to the linked list might be more future proof there? This function is called on every log message after all

src/basic/log.c
```diff
@@ -1478,3 +1516,47 @@ void log_setup(void) {
...
+LogContext* log_context_detach(LogContext *c) {
+        if (!c)
+                return NULL;
+
+        LIST_REMOVE(ll, _log_context, c);
+        return NULL;
+}
+
+LogContext* log_context_new(char **fields) {
+        LogContext *c = new(LogContext, 1);
```
poettering, line 1568: maybe add an `strv_isempty()` shortcut to bypass things?

src/basic/log.h
```diff
@@ -406,3 +408,79 @@ typedef struct LogRateLimit {
...
+
+#define LOG_CONTEXT_PUSH_STRV(strv) \
+        _LOG_CONTEXT_PUSH_STRV(strv, UNIQ_T(c, UNIQ))
+
+/* LOG_CONTEXT_CONSUME_STR()/LOG_CONTEXT_CONSUME_STRV() are identical to
+ * LOG_CONTEXT_PUSH_STR()/LOG_CONTEXT_PUSH_STRV() except they take ownership of the given str/strv argument.
+ */
+
+#define _LOG_CONTEXT_CONSUME_STR(s, c) \
+        _unused_ _cleanup_(log_context_freep) LogContext *c = log_context_new(strv_new(s))
```
poettering: this will leak memory if the strv_new() succeeds but the log_context_new() does not

src/libsystemd/sd-device/device-monitor.c
```diff
@@ -246,6 +246,8 @@ static int device_monitor_event_handler(sd_event_source *s, int fd, uint32_t rev
         if (device_monitor_receive_device(m, &device) <= 0)
                 return 0;
 
+        LOG_CONTEXT_CONSUME_STRV(device_make_log_fields(device));
```
poettering: hmm, i am not sure i like this use here...

I mean, think about the callback function called below might start doing pretty much anything, for example a screen redraw in a GUI app. But if some graphics function then logs, should that really carry device info? I am not sure...

Using this funcitonality is good, but when doing a generic callback dispatch then that's going to far I think, because you have no clue what the callback will do and if it will issue more than just stuff directly related to the device even we have here...

src/libsystemd/sd-device/device-util.h
```diff
@@ -83,3 +83,6 @@
 
 int devpath_from_devnum(mode_t mode, dev_t devnum, char **ret);
 int device_open_from_devnum(mode_t mode, dev_t devnum, int flags, char **ret);
+
+char** device_make_log_fields(sd_device *device);
+
```
poettering: trailing empty line

src/libsystemd/sd-bus/bus-message.c
```diff
@@ -4649,3 +4649,26 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
...
+
+        (void) strv_extend_assignment(&strv, "DBUS_MESSAGE_TYPE", bus_message_type_to_string(m->header->type));
+        (void) strv_extend_assignment(&strv, "DBUS_SENDER", sd_bus_message_get_sender(m));
+        (void) strv_extend_assignment(&strv, "DBUS_DESTINATION", sd_bus_message_get_destination(m));
+        (void) strv_extend_assignment(&strv, "DBUS_PATH", sd_bus_message_get_path(m));
+        (void) strv_extend_assignment(&strv, "DBUS_INTERFACE", sd_bus_message_get_interface(m));
+        (void) strv_extend_assignment(&strv, "DBUS_MEMBER", sd_bus_message_get_member(m));
+
+        (void) strv_extendf(&strv, "DBUS_MESSAGE_COOKIE=" PRIu64, BUS_MESSAGE_COOKIE(m));
+        if (m->reply_cookie)
```
poettering: coding style: this is not boolean/boolean-like or a pointer, but a numeric integer, hence check with `reply->cookie != 0`

poettering (changes requested):

src/basic/list.h
yuwata: (BTW, we already do that in macro.h, i.e. log.h is included at the end of macro.h)

src/basic/stdio-util.h
daandemeyer: I want to pass a pre-allocated buffer via a compound initializer. If we have the macro + statement expressions, the compound initializer block will be scoped to the macro block. If we make it a static inline function, it will be scoped to the block of the function calling snprintf_ok().

src/libsystemd/sd-device/device-monitor.c
daandemeyer: Maybe this is excessive information to attach by default. But it's still immensely useful information to know that the graphics function logged a message during the processing of a device monitor callback for device X.

You can go pretty wild with the context concept. If you want to be able to filter all log messages logged during the execution of a particular function, just push some unique field on the context and you can filter by that in journalctl.

yuwata: Still I do not like this. And I agree with @poettering's concern in the above. Relevant discussion is here https://github.com/systemd/systemd/pull/23309#discussion_r947240763 (sorry for late response, though).

yuwata (changes requested):

yuwata:
Or, maybe, instead of adding full device information, how about to add basic callback information e.g. `SYSTEMD_CALLBACK="sd-device-monitor(description)=sysname"`?

DaanDeMeyer:
I dropped the device stuff, we can discuss it in a separate PR, now only the D-Bus context fields are left

src/basic/stdio-util.h
keszybz: Can you add this rationale to the commit description?

>  hmm, i wonder how the generated code here actually compares to the status quo ante?

That's a good question, but since we only have 30 invocations of snprintf_ok in the codebase, we don't really care if it's particularly efficient.

src/basic/stdio-util.h
```diff
@@ -7,7 +7,6 @@
 #include <sys/types.h>
 
 #include "macro.h"
-#include "memory-util.h"
```
keszybz, line 10: Please add rationale to the commit message.

src/libsystemd/sd-bus/bus-message.c
```diff
@@ -4667,3 +4667,26 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
...
+        _cleanup_strv_free_ char **strv = NULL;
+
+        assert(m);
+
+        (void) strv_extend_join(&strv, "DBUS_MESSAGE_TYPE", "=", bus_message_type_to_string(m->header->type));
+        (void) strv_extend_join(&strv, "DBUS_SENDER", "=", sd_bus_message_get_sender(m));
+        (void) strv_extend_join(&strv, "DBUS_DESTINATION", "=", sd_bus_message_get_destination(m));
+        (void) strv_extend_join(&strv, "DBUS_PATH", "=", sd_bus_message_get_path(m));
+        (void) strv_extend_join(&strv, "DBUS_INTERFACE", "=", sd_bus_message_get_interface(m));
+        (void) strv_extend_join(&strv, "DBUS_MEMBER", "=", sd_bus_message_get_member(m));
```
keszybz: Hmm, would `strv_extend_join` ever be used with something else than "=" as the joiner? It'd be more effiecient to change those calls to
```C
 (void) strv_extend_join(&strv, "DBUS_MESSAGE_TYPE=", bus_message_type_to_string(m->header->type));
...
```

keszybz (commented):
I wasn't following the previous discussion, so apologies if I'm asking something that is already answered in some buried comment… Partial review, I didn't look at commit 5/6 that does the actual implementation yet.

src/libsystemd/sd-bus/bus-message.c
```diff
@@ -4667,3 +4667,26 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
...
+        (void) strv_extend_join(&strv, "DBUS_MESSAGE_TYPE=", bus_message_type_to_string(m->header->type));
+        (void) strv_extend_join(&strv, "DBUS_SENDER=", sd_bus_message_get_sender(m));
+        (void) strv_extend_join(&strv, "DBUS_DESTINATION=", sd_bus_message_get_destination(m));
+        (void) strv_extend_join(&strv, "DBUS_PATH=", sd_bus_message_get_path(m));
+        (void) strv_extend_join(&strv, "DBUS_INTERFACE=", sd_bus_message_get_interface(m));
+        (void) strv_extend_join(&strv, "DBUS_MEMBER=", sd_bus_message_get_member(m));
+
+        (void) strv_extendf(&strv, "DBUS_MESSAGE_COOKIE=" PRIu64, BUS_MESSAGE_COOKIE(m));
+        if (m->reply_cookie != 0)
+                (void) strv_extendf(&strv, "DBUS_MESSAGE_REPLY_COOKIE=" PRIu64, m->reply_cookie);
```
keszybz: '%' is missing: `"DBUS_MESSAGE_COOKIE=%"PRIu64`, `"DBUS_MESSAGE_REPLY_COOKIE=%"PRIu64`.

keszybz (commented):
Looks good otherwise.

yuwata:
Sorry, but I do not like the last commit, as the same reason https://github.com/systemd/systemd/pull/23309#discussion_r965907330.

keszybz:
> Sorry, but I do not like the last commit, as the same reason [#23309 (comment)](https://github.com/systemd/systemd/pull/23309#discussion_r965907330).

OK, so you are worried that other applications using libsystemd can do arbitrary things in nested calls from libsystemd and that this metadata would be unwanted there. I guess this makes sense. @DaanDeMeyer what about the adding a flag for opting in? I.e. have some (unexported at least for now) function bus_enable_log_context() and systemd programs would call that and get the context logs after allocating the bus object.

DaanDeMeyer:
> OK, so you are worried that other applications using libsystemd can do arbitrary things in nested calls from libsystemd and that this metadata would be unwanted there. I guess this makes sense. @DaanDeMeyer what about the adding a flag for opting in? I.e. have some (unexported at least for now) function bus_enable_log_context() and systemd programs would call that and get the context logs after allocating the bus object.

That's OK for me. @yuwata would this address your concerns?

yuwata:
> > OK, so you are worried that other applications using libsystemd can do arbitrary things in nested calls from libsystemd and that this metadata would be unwanted there. I guess this makes sense. @DaanDeMeyer what about the adding a flag for opting in? I.e. have some (unexported at least for now) function bus_enable_log_context() and systemd programs would call that and get the context logs after allocating the bus object.
> 
> That's OK for me. @yuwata would this address your concerns?

Yeah. @poettering WDYT?

DaanDeMeyer:
I added back the device monitor log context stuff as well with the same fix so that enabling the log context for the device monitor requires an extra (private) function call

src/basic/strv.c
```diff
@@ -789,6 +789,22 @@ int strv_extend_n(char ***l, const char *value, size_t n) {
         return -ENOMEM;
 }
 
+int strv_extend_join(char ***l, const char *lhs, const char *rhs) {
```
poettering: hmm, one of these days this should be rework to take variadic params, and dispatch to some flavour of strjoin_real()

src/basic/strv.c
```diff
@@ -789,6 +789,22 @@ int strv_extend_n(char ***l, const char *value, size_t n) {
...
+        char *j;
+
+        assert(l);
+        assert(lhs);
+
+        if (!rhs) /* value is optional, in which case we suppress the field */
+                return 0;
```
poettering, line 799: hmm, these semantics sound quite specific to me. should this really be in strv?

at the very least it should have a different name, i.e. strv_extend_with_assignment() or so?

shouldn't the `=` be inserted automatically here?

src/basic/log.c
```diff
@@ -68,6 +70,17 @@ static bool prohibit_ipc = false;
...
+typedef struct LogContext {
+        /* Depending on which destructor is used (log_context_free() or log_context_detach()) the memory
+         * referenced by this is freed or not */
+        char **fields;
+        bool owned;
+        LIST_FIELDS(struct LogContext, ll);
+} LogContext;
+
+thread_local LIST_HEAD(LogContext, _log_context) = NULL;
+thread_local size_t _log_context_num_fields = 0;
```
poettering: no `static`?

src/basic/log.c
```diff
@@ -1477,3 +1523,72 @@ void log_setup(void) {
         if (log_on_console() && show_color < 0)
                 log_show_color(true);
 }
+
+LogContext* log_context_attach(LogContext *c) {
+        assert(c);
+
+        STRV_FOREACH(s, c->fields)
+                _log_context_num_fields++;
```
poettering:
```c
_log_context_num_fields += strv_length(c->fields)
```

src/basic/log.c
```diff
@@ -1477,3 +1523,72 @@ void log_setup(void) {
...
+
+        return LIST_PREPEND(ll, _log_context, c);
+}
+
+LogContext* log_context_detach(LogContext *c) {
+        if (!c)
+                return NULL;
+
+        STRV_FOREACH(s, c->fields)
+                _log_context_num_fields--;
```
poettering: similar.

also, please add assert() checking for underflow ahead of time

src/basic/log.c
```diff
@@ -1477,3 +1523,72 @@ void log_setup(void) {
...
+
+        LIST_REMOVE(ll, _log_context, c);
+        return NULL;
+}
+
+LogContext* log_context_new(char **fields, bool owned) {
+        if (strv_isempty(fields))
+                return NULL;
```
poettering: hmm, this makes it hard to distinguish allocation failures from invocations with an empty list.

maybe add:

```c
#define LOG_CONTEXT_EMPTY ((LogContext*) (uintptr_t) 1)
```

as a special value you can return here to indicate an empty log context.

And then handle that specially in log_context_attach(), log_context_free() and so on.

poettering: or, alternatively, just define a static empty one as:

```c
static const LogContext empty_log_context = {};
```

and then return a pointer to that?

src/basic/log.c
```diff
@@ -1477,3 +1523,72 @@ void log_setup(void) {
...
+
+        log_context_detach(c);
+
+        if (c->owned)
+                strv_free(c->fields);
+
+        return mfree(c);
+}
+
+LogContext* log_context_consume(char **fields) {
```
poettering: maybe name log_context_new_consume()?

src/libsystemd/sd-bus/bus-message.c
```diff
@@ -4667,3 +4667,26 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
         m->sensitive = true;
         return 0;
 }
+
+char** bus_message_make_log_fields(sd_bus_message *m) {
+        _cleanup_strv_free_ char **strv = NULL;
+
+        assert(m);
+
+        (void) strv_extend_join(&strv, "DBUS_MESSAGE_TYPE=", bus_message_type_to_string(m->header->type));
```
poettering: I think strv_extend_join() shiuld really be called strv_append_assignment(), and then insert the `=` on its own

src/libsystemd/sd-bus/bus-message.c
```diff
@@ -4667,3 +4667,26 @@ _public_ int sd_bus_message_sensitive(sd_bus_message *m) {
...
+        (void) strv_extendf(&strv, "DBUS_MESSAGE_COOKIE=%" PRIu64, BUS_MESSAGE_COOKIE(m));
+        if (m->reply_cookie != 0)
+                (void) strv_extendf(&strv, "DBUS_MESSAGE_REPLY_COOKIE=%" PRIu64, m->reply_cookie);
+
+        (void) strv_extend_join(&strv, "DBUS_SIGNATURE=", m->root_container.signature);
+        (void) strv_extend_join(&strv, "DBUS_ERROR_NAME=", m->error.name);
+        (void) strv_extend_join(&strv, "DBUS_ERROR_MESSAGE=", m->error.message);
+
+        return TAKE_PTR(strv);
+}
```
poettering, line 4692: instead of enabling this manually in all our code, shouldn't this be something we just enable when `DEBUG_LOGGING` holds?

That way, once you enable debug logging you get the extra fields for *all* messages just like that, but without it we won't waste time on all those memory allocations?

poettering: or here's another idea, instead of adding an boolean settng to the sd_bus object, and another one to sd_device and so on – maybe instead add one to log.c called "log_request_context_requested" or so that sd-bus, sd-device and so on all check on their own. i.e. a single global option I think is much nicer than having to enable this in every subsystem individually.

poettering: or maybe make log_request_context_requested a tristate:

1. if 0: no extended log contexts are added anywhere
2. if > 0: extended log contexts are added everywhere
3. if < 0: extended log contexts are done if debug logging is on

< 0 would be the default of course.

And then maybe add an env var for this too, so that callers can request this from the command lines explicitly without necessarily having to enable debug logging.

poettering:
looks good, but I think the logic should be enabled differently, see my comments above.

src/basic/log.c
daandemeyer: I just removed the empty `fields` check, if there are no fields, we still allocate a log context object. We can't use a static const instance here because we always have to return an allocated object that the user can free.

bluca (approved):

PR #23310, opened 2022-05-08
Suppress fuzzer timeouts and errors

keszybz:
I was looking at the list of open issues on oss-fuzz. No real errors were reported, but rather timeouts
caused by large input size and bad algorithmic complexity in our code. Options and checked return
values as adjusted to not see those errors.

Code is formatted / cleaned-up in various trivial ways.

### DIFF
--- a/src/shared/calendarspec.c
+++ b/src/shared/calendarspec.c
@@ -76,26 +76,21 @@ static int component_compare(CalendarComponent * const *a, CalendarComponent * c
 }
 
 static void normalize_chain(CalendarComponent **c) {
-        CalendarComponent **b, *i, **j, *next;
-        size_t n = 0, k;
-
         assert(c);
 
-        for (i = *c; i; i = i->next) {
+        size_t n = 0;
+        for (CalendarComponent *i = *c; i; i = i->next) {
                 n++;
 
-                /*
-                 * While we're counting the chain, also normalize `stop`
-                 * so the length of the range is a multiple of `repeat`
-                 */
+                /* While we're counting the chain, also normalize 'stop'
+                 * so the length of the range is a multiple of 'repeat'. */
                 if (i->stop > i->start && i->repeat > 0)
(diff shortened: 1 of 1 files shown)


bluca:
```
../src/xdg-autostart-generator/test-xdg-autostart.c:50:9: error: suspicious concatenation of string literals in an array initialization; did you mean to separate the elements with a comma? [-Werror,-Wstring-concatenation]
        "Hidden=\t true\n",
        ^
../src/xdg-autostart-generator/test-xdg-autostart.c:49:9: note: place parentheses around the string literal to silence warning
        "[Desktop Entry]\n"
        ^
```

keszybz:
> ```
> ../src/xdg-autostart-generator/test-xdg-autostart.c:50:9: error: suspicious concatenation of string literals in an array initialization; did you mean to separate the elements with a comma? [-Werror,-Wstring-concatenation]
>         "Hidden=\t true\n",
>         ^
> ../src/xdg-autostart-generator/test-xdg-autostart.c:49:9: note: place parentheses around the string literal to silence warning
>         "[Desktop Entry]\n"
>         ^
> ```

Oh, I'll drop the last commit.

bluca (approved):

src/systemctl/fuzz-systemctl-parse-argv.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+max_len = 65535
```
evverx: It doesn't work with `AFL` and `honggfuzz`. Something like `if (size > 65535) return 0;` should be added to the fuzz target to cover those fuzzing engines as well.

keszybz: Oh, OK. I changed it to just refuse long command lines. ("Long" in the sense of number of entries. This is more useful than limiting size, because in this particular case we care about O(n²) in strv_append().)

yuwata (approved):

PR #23319, opened 2022-05-09
test: JSON_BUILD_REAL nowadays expects 'double', not 'long double'

poettering:
Follow-up for 337712e777bff389f53e26d5b378d2ceba7d98a8, aka "the great
un-long-double-ification of 2021".

### DIFF
--- a/src/test/test-json.c
+++ b/src/test/test-json.c
@@ -315,7 +315,7 @@ TEST(build) {
         a = json_variant_unref(a);
         b = json_variant_unref(b);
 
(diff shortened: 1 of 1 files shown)


poettering:
(makes this patch go away https://git.openembedded.org/openembedded-core/tree/meta/recipes-core/systemd/systemd/0021-test-json.c-define-M_PIl.patch when porting to musl)

yuwata (approved):

bluca (approved):

kraj:
> /cc @kraj

ah thanks for fixing it ! one less to carry  on my end :)

PR #23323, opened 2022-05-09
mkosi: add shadow package to SUSE Tumbleweed

bluca:
[1958/1958] Generating export-dbus-interfaces with a custom command
/root/mkosi.build: line 70: groupadd: command not found

### DIFF
--- a/mkosi.default.d/opensuse/10-mkosi.opensuse
+++ b/mkosi.default.d/opensuse/10-mkosi.opensuse
@@ -39,6 +39,7 @@ BuildPackages=
         python3-Jinja2
         python3-lxml
         qrencode-devel
+        shadow
         system-user-nobody
         systemd-sysvinit
         zlib-devel


DaanDeMeyer:
I figure @Werkov should pitch in here, it seems a bit off for the package to suddenly disappear.

bluca:
> I figure @Werkov should pitch in here, it seems a bit off for the package to suddenly disappear.

Maybe things are getting re-arranged - but we can fix the job now, and revise later if needed

Werkov:
The openSUSE Tumbleweed mkosi image is based on very minimal base image and most requirements must be explicit. The reorg of basic packages (providing groups, basic utilities) happens from time to time. I wasn't able to track down the particular change (`shadow` might have been pulled in transitively too or some other packages ensure existence of group IDs and `groupadd` was not resolved&invoked.)

All in all, since the missing groupadd was the [only breakage](https://github.com/systemd/systemd/runs/6352823808?check_suite_focus=true), explicitly adding its provider is a good fix. Thanks!

PR #23329, opened 2022-05-10
Set auto-suspend delay on LG850-GL

superm1:
It has been shown that the default autosuspend delay for this device and the signal quality checking
by modem manager will race with suspend and cause system suspend
failures.
    
This occurred in ChromiumOS on a chromebook, but there is no reason
it won't happen in regular notebooks with the same WWAN.  To avoid
the race delay autosuspend to a frequency longer than the polling rate
used by modem manager for signal information.
    
Link: https://chromium-review.googlesource.com/c/chromiumos/overlays/board-overlays/+/3635003


yuwata (approved):
LGTM.

keszybz:
Nice!

PR #23334, opened 2022-05-10
fuzz-bootspec: limit input size

keszybz:
https://oss-fuzz.com/testcase-detail/5680508182331392 has the
first timeout with 811kb of input. As in the other cases, the code
is known to be slow with lots of repeated entries and we're fine with
that.


yuwata (approved):
LGTM.

PR #23335, opened 2022-05-10
More coverage in fuzz-json

keszybz:


src/shared/json.c
```diff
@@ -4499,14 +4489,8 @@ int json_dispatch_strv(const char *name, JsonVariant *variant, JsonDispatchFlags
...
-        JsonVariant **p = userdata;
-
-        assert(variant);
-        assert(p);
-
-        json_variant_unref(*p);
-        *p = json_variant_ref(variant);
-
+        JsonVariant **p = ASSERT_PTR(userdata);
+        JSON_VARIANT_REPLACE(*p, ASSERT_PTR(variant));
```
yuwata: missing `json_variant_ref()`.

yuwata (changes requested):
LGTM. But one issue.

src/shared/json.c
keszybz: Eh, indeed. I tried a few different forms here and apparently managed to introduce the mistake on the way.

keszybz:
> Though I am not sure it is related to the above point, CIs are unhappy.

Yeah, I'm not sure either. I pushed without any other changes for now.

PR #23336, opened 2022-05-10
More coverage in fuzz-calendarspec

keszybz:


src/shared/calendarspec.c
```diff
@@ -321,10 +328,14 @@ static void format_chain(FILE *f, int space, const CalendarComponent *c, bool us
 
         if (c->next) {
                 fputc(',', f);
-                format_chain(f, space, c->next, usec);
+                _format_chain(f, space, c->next, false, usec);
         }
 }
 
+static void format_chain(FILE *f, int space, const CalendarComponent *c, bool usec) {
+        _format_chain(f, space, c, true, usec);
```
yuwata: Nowadays we usually add comment for boolean argument: `/* start = */ true`

yuwata (approved):
LGTM. One minor suggestion.

src/shared/calendarspec.c
keszybz: Updated.

PR #23339, opened 2022-05-10
tree-wide: add support for connecting to AF_UNIX sockets in the file system beyond the 108ch limit

poettering:
This adds some infrastructure to make sure we can connect to AF_UNIX sockets in the filesystem beyond the 108 char limit that `sockaddr_un` enforces. It uses O_PATH to work around the limit (i.e. connects to `/proc/self/fd/<fd>` instead of the literal path).

Most of our internal sockets do not use such paths, but there's value to allowing paths configured externally to be connectible regardless of their length.

### DIFF
--- a/src/shared/nscd-flush.c
+++ b/src/shared/nscd-flush.c
@@ -1,4 +1,6 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include <fcntl.h>
 #include <poll.h>
 
 #include "fd-util.h"
@@ -18,17 +20,12 @@ struct nscdInvalidateRequest {
         char dbname[];
 };
 
-static const union sockaddr_union nscd_sa = {
-        .un.sun_family = AF_UNIX,
-        .un.sun_path = "/run/nscd/socket",
-};
-
 static int nscd_flush_cache_one(const char *database, usec_t end) {
         size_t req_size, has_written = 0, has_read = 0, l;
         struct nscdInvalidateRequest *req;
         _cleanup_close_ int fd = -1;
         int32_t resp;
-        int events;
+        int events, r;
 
         assert(database);
 
@@ -60,11 +57,12 @@ static int nscd_flush_cache_one(const char *database, usec_t end) {
          * nice way to connect() to a server synchronously with a time limit that would also cover dealing with the
          * backlog limit. After all SO_RCVTIMEO and SR_SNDTIMEO don't apply to connect(), and alarm() is frickin' ugly
          * and not really reasonably usable from threads-aware code.) */
-        if (connect(fd, &nscd_sa.sa, SOCKADDR_UN_LEN(nscd_sa.un)) < 0) {
-                if (errno == EAGAIN)
-                        return log_debug_errno(errno, "nscd is overloaded (backlog limit reached) and refuses to take further connections: %m");
-                if (errno != EINPROGRESS)
-                        return log_debug_errno(errno, "Failed to connect to nscd socket: %m");
+        r = connect_unix_path(fd, AT_FDCWD, "/run/nscd/socket");
+        if (r < 0) {
+                if (r == -EAGAIN)
+                        return log_debug_errno(r, "nscd is overloaded (backlog limit reached) and refuses to take further connections: %m");
+                if (r != -EINPROGRESS)
+                        return log_debug_errno(r, "Failed to connect to nscd socket: %m");
 
                 /* Continue in case of EINPROGRESS, but don't bother with send() or recv() until being notified that
                  * establishing the connection is complete. */
--- a/src/userdb/userdbctl.c
+++ b/src/userdb/userdbctl.c
@@ -942,25 +942,19 @@ static int display_services(int argc, char *argv[], void *userdata) {
 
         FOREACH_DIRENT(de, d, return -errno) {
                 _cleanup_free_ char *j = NULL, *no = NULL;
-                union sockaddr_union sockaddr;
-                socklen_t sockaddr_len;
                 _cleanup_close_ int fd = -1;
 
                 j = path_join("/run/systemd/userdb/", de->d_name);
                 if (!j)
                         return log_oom();
 
-                r = sockaddr_un_set_path(&sockaddr.un, j);
-                if (r < 0)
-                        return log_error_errno(r, "Path %s does not fit in AF_UNIX socket address: %m", j);
-                sockaddr_len = r;
-
                 fd = socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, 0);
                 if (fd < 0)
                         return log_error_errno(errno, "Failed to allocate AF_UNIX/SOCK_STREAM socket: %m");
 
-                if (connect(fd, &sockaddr.sa, sockaddr_len) < 0) {
-                        no = strjoin("No (", errno_to_name(errno), ")");
+                r = connect_unix_path(fd, dirfd(d), de->d_name);
+                if (r < 0) {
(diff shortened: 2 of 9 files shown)


src/basic/socket-util.c
```diff
@@ -1426,3 +1426,51 @@ int socket_get_mtu(int fd, int af, size_t *ret) {
...
+
+                /* if dir_fd is specified, then we need to go the indirect O_PATH route, because connectat()
+                 * does not exist. If the path is too long, we also need to take the indirect route, since we
+                 * cant fit this into a sockaddr_run directly. */
+
+                inode_fd = openat(dir_fd, path, O_PATH|O_CLOEXEC);
+                if (inode_fd < 0)
+                        return -errno;
+
+                proc = FORMAT_PROC_FD_PATH(inode_fd);
```
yuwata, line 1467: Is this safe?? I thought it can be only used as a function argument.

src/shared/nscd-flush.c
```diff
@@ -1,4 +1,5 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
+#include <fcntl.h>
```
yuwata, line 3: Not a matter, but we usually add a blank line after SPDX header.

bluca:
```
../src/basic/fileio.c:766:37: error: unused variable 'dfd' [-Werror,-Wunused-variable]
                _cleanup_close_ int dfd = -1, sk = -1;
                                    ^
```

yuwata (changes requested):

src/basic/socket-util.c
```diff
@@ -1426,3 +1426,51 @@ int socket_get_mtu(int fd, int af, size_t *ret) {
...
+        if (dir_fd == AT_FDCWD && path_len < sizeof(sa.un.sun_path)) {
+                memcpy(sa.un.sun_path, path, path_len + 1);
+                salen = offsetof(struct sockaddr_un, sun_path) + path_len + 1;
+        } else {
+                const char *proc;
+                size_t proc_len;
+
+                /* if dir_fd is specified, then we need to go the indirect O_PATH route, because connectat()
+                 * does not exist. If the path is too long, we also need to take the indirect route, since we
+                 * cant fit this into a sockaddr_run directly. */
```
floppym: Typo: `sockaddr_run` should be `sockaddr_un`.

bluca: also /cant/can't/

poettering:
> So the path will be invalid after the function returns, right? IE, `/proc/self/fd/<fd>` will be closed and no longer point to the socket once the connection completes. Would that cause issues? Especially given it could be reused at some point? Also other programs will see `/proc/self/fd<fd>` eg, in `netstat -x`, but obviously `self` is actually another process?

When connecting to a socket the `/proc/self/fd/<fd>` should note have any negative effects I would be aware of. Things like /proc/net/unix don't show the target path used for connection anyway... AF_UNIX sockets in the file system are weird beasts, they only act as entrypoints, and the connection socket does not carry info about the entrypoint anymore. i.e. if you do getpeername() of an AF_UNIX socket you connected like that you'll not get info about the socket node in the fs that you connected to, but about the AF_UNIX inode that is behind it...

bluca:
> > So the path will be invalid after the function returns, right? IE, `/proc/self/fd/<fd>` will be closed and no longer point to the socket once the connection completes. Would that cause issues? Especially given it could be reused at some point? Also other programs will see `/proc/self/fd<fd>` eg, in `netstat -x`, but obviously `self` is actually another process?
> 
> When connecting to a socket the `/proc/self/fd/<fd>` should note have any negative effects I would be aware of. Things like /proc/net/unix don't show the target path used for connection anyway... AF_UNIX sockets in the file system are weird beasts, they only act as entrypoints, and the connection socket does not carry info about the entrypoint anymore. i.e. if you do getpeername() of an AF_UNIX socket you connected like that you'll not get info about the socket node in the fs that you connected to, but about the AF_UNIX inode that is behind it...

Got it, thanks

src/basic/socket-util.c
poettering: this is safe, c99 compound literals are valid from the point on they are declared. What's key though is that the memory is only valid within the current {} block. Which we comply here by, we copy it into the sockaddr_un further down, via memcpy(), and then don't need the memory anmyore.

poettering: this is safe, c99 compound literals are valid from the point on they are declared. What's key though is that the memory is only valid within the current {} block. Which we comply here by, we copy it into the sockaddr_un further down, via memcpy(), and then don't need the memory anmyore.

poettering:
> When connecting to a socket the /proc/self/fd/<fd> should note have any negative effects I would be aware of. Things like /proc/net/unix don't show the target path used for connection anyway... AF_UNIX sockets in the file system are weird beasts, they only act as entrypoints, and the connection socket does not carry info about the entrypoint anymore. i.e. if you do getpeername() of an AF_UNIX socket you connected like that you'll not get info about the socket node in the fs that you connected to, but about the AF_UNIX inode that is behind it...

Also, what's the alternative here? that is to simply fail. I think making this just work is the better approach even if it's a rather arcane indirection we have to implement here...

poettering:
New version pushed, addressing all issues raised. ptal. given this fixes a bad error propagation and comes with tests (i.e. should be relatively safe) I think this should probably be merged before 251. but i could also just cherry-pick that one relevant patch...

src/userdb/userdbctl.c
```diff
@@ -942,25 +942,19 @@ static int display_services(int argc, char *argv[], void *userdata) {
...
                         return log_oom();
 
-                r = sockaddr_un_set_path(&sockaddr.un, j);
-                if (r < 0)
-                        return log_error_errno(r, "Path %s does not fit in AF_UNIX socket address: %m", j);
-                sockaddr_len = r;
-
                 fd = socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC|SOCK_NONBLOCK, 0);
                 if (fd < 0)
                         return log_error_errno(r, "Failed to allocate AF_UNIX/SOCK_STREAM socket: %m");
```
yuwata: Another bug. r -> errno.

yuwata (changes requested):

yuwata:
Could you add a test for connect_unix_path()?

bluca:
> New version pushed, addressing all issues raised. ptal. given this fixes a bad error propagation and comes with tests (i.e. should be relatively safe) I think this should probably be merged before 251. but i could also just cherry-pick that one relevant patch...

It's very late in the cycle, I think it would safer to just cherry-pick the error propagation fix and merge the rest later? It would probably be fine, but I don't think there's an urgent need either, right?

poettering:
> Could you add a test for connect_unix_path()?

There's a test for this already in place, via the read_file_full() test, which tests behaviour with overlay long sockets paths. I wouldn't know which corner cases wouldn't already be covered by that...

yuwata:
> > New version pushed, addressing all issues raised. ptal. given this fixes a bad error propagation and comes with tests (i.e. should be relatively safe) I think this should probably be merged before 251. but i could also just cherry-pick that one relevant patch...
> 
> It's very late in the cycle, I think it would safer to just cherry-pick the error propagation fix and merge the rest later? It would probably be fine, but I don't think there's an urgent need either, right?

#23369.

> > Could you add a test for connect_unix_path()?
> 
> There's a test for this already in place, via the read_file_full() test, which tests behaviour with overlay long sockets paths. I wouldn't know which corner cases wouldn't already be covered by that...

Ah, thanks.

yuwata:
Rebased as #23369 is merged.

Issue #23263, opened 2022-05-04
test-execute.c:874:test_exec_dynamicuser: exec-dynamicuser-statedir.service: can_unshare=yes: exit status 1, expected 0


mbiebl:
Seen with v250 and latest git HEAD.

When running the unit-tests via LXC/autopkgtest I get the following failure in test-execute:
```
exec-dynamicuser-statedir.service: Child 6387 belongs to exec-dynamicuser-statedir.service.
exec-dynamicuser-statedir.service: Main process exited, code=exited, status=0/SUCCESS (success)
exec-dynamicuser-statedir.service: Running next main command for state start.
exec-dynamicuser-statedir.service: Will spawn child (service_run_next_main): sh
exec-dynamicuser-statedir.service: Passing 0 fds to service
exec-dynamicuser-statedir.service: About to execute sh -x -c "test \$\$(find / \\( -path /var/tmp -o -path /tmp -o -path /proc -o -path /dev/mqueue -o -path /dev/shm -o -path /sys/fs/bpf -o -path /dev/.lxc \\) -prune -o -type d -writable -print 2>/dev/null | sort -u | tr -d \"\\\\n\") = /var/lib/private/quux/pief/var/lib/private/waldo"
exec-dynamicuser-statedir.service: Forked sh as 6388
exec-dynamicuser-statedir.service: User lookup succeeded: uid=65325 gid=65325
Received SIGCHLD from PID 6388 (sh).
Child 6388 (sh) died (code=exited, status=1/FAILURE)
exec-dynamicuser-statedir.service: Child 6388 belongs to exec-dynamicuser-statedir.service.
exec-dynamicuser-statedir.service: Main process exited, code=exited, status=1/FAILURE
exec-dynamicuser-statedir.service: Failed with result 'exit-code'.
exec-dynamicuser-statedir.service: Service will not restart (restart setting)
exec-dynamicuser-statedir.service: Changed start -> failed
exec-dynamicuser-statedir.service: Unit entered failed state.
Spawning thread to nuke /tmp/systemd-private-00dcaf0615b640509dbdecf874e9e22e-exec-dynamicuser-statedir.service-6LeDti
Spawning thread to nuke /var/tmp/systemd-private-00dcaf0615b640509dbdecf874e9e22e-exec-dynamicuser-statedir.service-kNixoB
UID 65325 is no longer referenced, cleaning up its IPC.
GID 65325 is no longer referenced, cleaning up its IPC.
[...]

PR #23343, opened 2022-05-11
test: ignore LXC filesystem when checking for writable locations

bluca:
test-execute checks that only /var/lib/private/waldo is writable, but there are
some filesystems that are always writable and excluded. Add /sys/devices/system/cpu
which is created by lxcfs.

Fixes https://github.com/systemd/systemd/issues/23263

### DIFF
--- a/test/test-execute/exec-dynamicuser-statedir.service
+++ b/test/test-execute/exec-dynamicuser-statedir.service
@@ -14,7 +14,7 @@ ExecStart=test -f /var/lib/private/quux/pief/yayyay
 ExecStart=sh -x -c 'test "$$STATE_DIRECTORY" = "%S/waldo:%S/quux/pief"'
 
 # Make sure that /var/lib/private/waldo is really the only writable directory besides the obvious candidates
-ExecStart=sh -x -c 'test $$(find / \\( -path /var/tmp -o -path /tmp -o -path /proc -o -path /dev/mqueue -o -path /dev/shm -o -path /sys/fs/bpf -o -path /dev/.lxc \\) -prune -o -type d -writable -print 2>/dev/null | sort -u | tr -d "\\\\n") = /var/lib/private/quux/pief/var/lib/private/waldo'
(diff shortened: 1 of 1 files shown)


poettering:
hmm, what?

/dev/.lxc and /sys/devices/system/cpu are writable to arbitrary users inside an LXC container?

world-writable dirs? that looks like a bug in LXC or in the build system that uses LXC here. Either way, I doubt we should work around it.

poettering:
i mean, the test uses DynamicUser=1 to acquire a new UID, and then checks whether access(W_OK) will fail everywhere except known world-wrtiable dirs (i.e. /tmp/ + /var/tmp/ + /dev/shm/, and the one dir it manages itself). If there are more dirs like that in an LXC environment, then that's almost certainly a bug somewhere.

bluca:
> hmm, what?
> 
> /dev/.lxc and /sys/devices/system/cpu are writable to arbitrary users inside an LXC container?
> 
> world-writable dirs? that looks like a bug in LXC or in the build system that uses LXC here. Either way, I doubt we should work around it.

Perhaps, I'm really not familiar with it - but I do note there's already an exclusion for `/dev/.lxc` which I assume is somewhat related. Regardless, I don't think we should have our tests fail because of this, there's nothing we can/should do about it and it's an annoyance for maintainers. I don't think it's a problem in our code or configuration?

bluca:
@brauner do you know anything about this lxcfs stuff?

brauner:
> @brauner do you know anything about this lxcfs stuff?

It's created by a tiny fuse filesystem (LXCFS) that provides a containerized cpu view.

poettering:
@hallyn lxcfs is a Serge Hallyn thing, right? Any chance you could comment? access(W_OK) seems to be incorrectly implemented on lxcfs for some files. unpriv users supposedly have write access to that dir, which looks wrong.

bluca:
> can we make "find" exclude all fuse file systems? i.e. match on fs magic? if so we could just exclude all fuse stuff in one go, under the assumption that access(W_OK) is quite possibly implemented incorrectly in them. I find that more acceptable than working around specifically for bugs...

mmh maybe? There is an `-fstype` option

bluca:
> `-fstype` doesn't work too great because the fstype is something like `fuse.lxc`, and we need a literal match. (E.g. for sshfs it's `fuse.sshfs`.) So maybe we should merge this, and figure out something better in the future?

I see, we'd trade a bunch of paths for a bunch of filesystems types - I'd leave it as-is too then

keszybz:
OK, let's go with this version then.

PR #23344, opened 2022-05-11
boot: Small config entry add changes

medhefgo:
- boot: Remove trivial config entry add helpers
- boot: Rename functions adding entries to match userspace

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -1421,7 +1421,7 @@ static void config_entry_bump_counters(ConfigEntry *entry, EFI_FILE *root_dir) {
         }
 }
 
-static void config_entry_add_from_file(
+static void config_entry_add_type1(
                 Config *config,
                 EFI_HANDLE *device,
                 EFI_FILE *root_dir,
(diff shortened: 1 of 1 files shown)


src/boot/efi/boot.c
```diff
@@ -2266,18 +2212,22 @@ static void config_entry_add_linux(
...
-                                /* title= */ good_name,
-                                /* loader= */ path,
-                                /* sort_key= */ good_sort_key,
-                                good_version);
+                ConfigEntry *entry = xnew(ConfigEntry, 1);
+                *entry = (ConfigEntry) {
+                        .id = xstrdup(f->FileName),
+                        .type = LOADER_UNIFIED_LINUX,
+                        .title = xstrdup(good_name),
+                        .version = xstrdup(good_version),
```
keszybz: Hmm, are we certain that `good_name` or `good_version` are set? Looking at ` bootspec_pick_name_version_sort_key()`, this doesn't seem guaranteed.

medhefgo: They should be: https://github.com/systemd/systemd/blob/5f1077af1eda0bfee7a7ec68b4912ed8e2d86484/src/fundamental/bootspec-fundamental.c#L49

medhefgo: I guess it's now only guaranteed for good_name. Updated the commit.

keszybz:
LGTM.

PR #23348, opened 2022-05-11
shutdown: Log processes that block umount

medhefgo:


medhefgo:
I tried testing this on my current system where something is keep /oldroot/sys busy, but it turns out it's not blocked by any running process. :/

So this is essentially untested and I am at a loss as to how to otherwise test it during a real shutdown. Suggestions are welcome.
The general logic does work though, as I developed the function in a unit test.

src/shutdown/umount.c
```diff
@@ -524,6 +527,57 @@ static bool nonunmountable_path(const char *path) {
                 || path_startswith(path, "/run/initramfs");
 }
 
+static void log_mount_blockers(int umount_log_level, const char *mnt) {
+        /* Logging proccesses that block umount could get very verbose. So only log when we are at the later
```
bluca: /proccesses/processes/

src/shutdown/umount.c
```diff
@@ -524,6 +527,57 @@ static bool nonunmountable_path(const char *path) {
...
+        FOREACH_DIRENT_ALL(de, dir, break) {
+                pid_t pid;
+
+                if (!IN_SET(de->d_type, DT_DIR, DT_UNKNOWN))
+                        continue;
+
+                if (parse_pid(de->d_name, &pid) < 0)
+                        continue;
+
+                const char *fd_path = procfs_file_alloca(pid, "fd");
```
bluca: I thought alloca in a loop was a no-no?

mrc0mmand: LGTM has the same opinion :-)

bluca: Good bot!

src/shutdown/umount.c
```diff
@@ -524,6 +528,63 @@ static bool nonunmountable_path(const char *path) {
...
+                                continue;
+
+                        if (PATH_STARTSWITH_SET(open_file, "/dev", "/sys", "/proc"))
+                                continue;
+
+                        if (get_process_comm(pid, &comm) < 0)
+                                continue;
+
+                        strextend_with_separator(&blockers, ", ", comm);
+                        strextend(&blockers, "(", de->d_name, ")");
```
bluca: check return values?

src/shutdown/umount.c
```diff
@@ -584,8 +646,12 @@ static int umount_with_timeout(MountPoint *m, int umount_log_level) {
                  * filesystem less busy so the unmount might succeed (rather
                  * than return EBUSY). */
                 r = umount2(m->path, MNT_FORCE);
```
bluca: `r = RET_NERRNO(umount2...)`?

medhefgo:
In my case the binary blob nvidia driver keeps /oldroot/sys alive for some reason (I found this out by unloading the modules and then trying to unmount again). The second commit tries to work around such cases by utilizing lazy unmounting. An alternative would be to try unloading kernel modules, but that sounds too intrusive to me.

ghtesting2020:
Nvidia has filed an internal bug report to look into the unmount issue. NVBug 3673743

https://github.com/NVIDIA/open-gpu-kernel-modules/issues/289#issuecomment-1148063425

src/shutdown/umount.c
```diff
@@ -651,6 +651,17 @@ static int umount_with_timeout(MountPoint *m, int umount_log_level) {
                 if (r == -EBUSY)
                         log_umount_blockers(umount_log_level, m->path);
 
+                /* If API filesystems under /oldroot cannot be unmounted we can still lazily unmount
+                 * them to unblock /oldrood. They serve no function to us anymore and should be
```
bluca: /oldrood -> /oldroot

src/shutdown/umount.c
```diff
@@ -651,6 +651,17 @@ static int umount_with_timeout(MountPoint *m, int umount_log_level) {
...
+                 * memory-only and hence safe to unmount like this. */
+                if (r == -EBUSY && in_initrd() &&
```
bluca: given it's a critical path, shouldn't this be in the previous r == -EBUSY block, and have log_umount_blockers as an else branch? No point in going through open files if the lazy one succeeds for these, no?

medhefgo: I think there is still value in reporting any processes blocking these even if we then lazily unmount them. (Though, in this particular case there are no such processes as it's the nvidia kernel module itself blocking it somehow.)

bluca: The issue is that there will be a cost to do those iterations, and also if all works out fine, but there are scary logs anyway, a lot of users would freak out unnecessarily. What about logging here if the lazy umount fails?

bluca:
> I was wondering: is there a way to lazy unmount /olrdoot and then wait (with timeout) until the kernel has cleaned up whatever was left in there? Would sure beat the complexity we have right now…

I don't think so, that would require a reliable and race-free mountinfo, and there's no such thing given fsinfo() never went anywhere

src/shutdown/umount.c
medhefgo: That's what `umount_log_level` is for. It starts out at LOG_INFO and the blockers aren't logged until the log level is lower than that. It's only raised to LOG_ERR for a final umount round if we have no initrd to switch to (either because there is none or because we are in it). So it shouldn't add any costs.

It probably makes sense to change that param to a bool as it doesn't really reflect the regular log level to make it more clear.

Unless I got something wrong, we never even look for unmount blockers unless we are desperate, at which point we really should look and report for them.

bluca (approved):

yuwata (approved):
LGTM too.

DAC324:
Errors like 
```
 [   250.982832 ]  sd-umount[1177]: Failed to unmount /oldroot: Device or resource busy
 [   250.983401 ]  sd-umount[1178]: Failed to unmount /oldroot/dev: Device or resource busy
 [   250.984415 ]  shutdown[1]: Failed to finalize file systems, ignoring. 
```
are occurring randomly because of Nvidia drivers. Forcefully unloading them on shutdown prevents the error from occurring.

Solution:

Create a script with the following contents
```
#!/bin/sh
# Remove all Nvidia modules on shutdown
# to avoid errors like
# [35509.312925] sd-umoun[56020]: Failed to unmount /oldroot: Device or resource busy
# [35509.317505] sd-umoun[56021]: Failed to unmount /oldroot/sys: Device or resource busy
# [35509.322349] shutdown[1]: Failed to finalize file systems, ignoring.
#
for MODULE in nvidia_drm nvidia_modeset nvidia_uvm nvidia
   do
# we will have to check if the modules are still loaded because they aren't always
       if lsmod | grep "$MODULE" &> /dev/null ; then
          rmmod $MODULE
       fi
   done  
```
Name this script, for example, `nvidia.shutdown`, make it executable as usual and place it in `/usr/lib/systemd/system-shutdown`.

src/shutdown/shutdown.c
```diff
@@ -501,13 +497,12 @@ int main(int argc, char *argv[]) {
...
                 }
 
-                if (!changed && umount_log_level == LOG_INFO && !can_initrd) {
-                        /* There are things we cannot get rid of. Loop one more time
-                         * with LOG_ERR to inform the user. Note that we don't need
-                         * to do this if there is an initrd to switch to, because that
-                         * one is likely to get rid of the remaining mounts. If not,
-                         * it will log about them. */
-                        umount_log_level = LOG_ERR;
+                if (!changed && last_try && !can_initrd) {
```
poettering, line 500: this doesn't look right. must be `!last_try`…

src/shutdown/umount.c
```diff
@@ -524,6 +528,58 @@ static bool nonunmountable_path(const char *path) {
...
+                if (!IN_SET(de->d_type, DT_DIR, DT_UNKNOWN))
+                        continue;
+
+                pid_t pid;
+                if (parse_pid(de->d_name, &pid) < 0)
+                        continue;
+
+                _cleanup_closedir_ DIR *pid_dir = xopendirat(dirfd(dir), de->d_name, 0);
+                if (!pid_dir)
+                        continue;
```
poettering, line 548: hmm, what's the point of maintaining pid_dir here? why not just do:

```c
_cleanup_free_ char *fd_path = NULL;

fd_path = path_join(de->d_name, fd);
if (!fd_path)
        return log_oom();

fd_dir = xopendirat(dirfddir), fd_path, 0);
```

That's two syscalls less!

(I man, I usually don't care about optimizations like that, but iterating through /proc/ is so slow anyway, let's not waste syscalls in inner loops)

src/shutdown/umount.c
```diff
@@ -648,6 +648,17 @@ static int umount_with_timeout(MountPoint *m, bool last_try) {
 
                         if (r == -EBUSY && last_try)
                                 log_umount_blockers(m->path);
+
+                        /* If API filesystems under /oldroot cannot be unmounted we can still lazily unmount
+                         * them to unblock /oldroot. They serve no function to us anymore and should be
+                         * memory-only and hence safe to unmount like this. */
```
poettering, line 654: hmm, oldroot? what is this about? What issue you adressing here?

src/shutdown/umount.c
```diff
@@ -648,6 +648,17 @@ static int umount_with_timeout(MountPoint *m, bool last_try) {
...
+                        if (in_initrd() &&
+                            PATH_STARTSWITH_SET(m->path, "/oldroot/dev", "/oldroot/proc", "/oldroot/sys")) {
+                                log_info("Lazily unmounting '%s' instead.", m->path);
```
poettering, line 657: I don't get this? when would /oldroot ever become visible in the initrd?

poettering: (btw, we probably should kill oldroot stuff altogether in switch root, as per the upated comments in pivot_root(), which indicate we don't actually need oldroot at all)

poettering (commented):
This looks like it needs some love still.

I don't grok the oldroot thing. This works around some breakage, but I wonder if there's a bug report about that. And anyway it appears to me the oldroot stuff is something we can get rid of entirely, which would be a much nicer solution.

See pivot_root man page, i.e.

```c
           chdir(new_root);
           pivot_root(".", ".");
           umount2(".", MNT_DETACH);
```

poettering:
BTW, to say this clearly: work-arounds for broken closed source software have no place in systemd. after reading the whole PR I understand that oldroot thing is a workaround for nvidia, but really, that does *not* belong in our tree.

It appears to me nvidia uses some hacks around the fact that device model is off limits for closed source crap and opens the sysfs files via in-kernel fds? man, that's so ugly...

Anyway, I think the oldroot hack really should go away, let's not do stuff like that. By implementing the pivot_root sequence from the man page we can get rid of oldroot altogether though, which is a general improvement, and I'd be happy to merge, and as side effect also makes the nvidia mess go away.

src/shutdown/umount.c
medhefgo: > I don't get this? when would /oldroot ever become visible in the initrd?

Once you've pivoted back to the initrd during final shutdown? D'uh!

And just to refresh your memory for the hundredth time on this: arch uses systemd-shutdown in its initrd to reduce duplication of effort (well, technically, it creates /run/initramfs as ramfs during shutdown with a service unit).

medhefgo:
I am all for lazy unmounting /oldroot. But how do you wait for the kernel to be finished with that? It afaik is a async operation by design.

Also, this is not nvidia specific. There could certainly be other cases where api filesystems are blocked, and we should try too not let that get in the way of cleaning things up.

poettering:
See: #24425

Can you check if that works for you, too?

poettering:
> Anyway, I think the oldroot hack really should go away, let's not do stuff like that. By implementing the pivot_root sequence from the man page we can get rid of oldroot altogether though, which is a general improvement, and I'd be happy to merge, and as side effect also makes the nvidia mess go away.

So to clarify this comment:

I understand now that the /oldroot/ issue this is about is not about the switch root from initrd into host fs at boot, but from host fs back into the initrd at shutdown. My suggestion to use the MNT_DETACH based switch root as suggested in the pivot_root man page is really about the initrd→host transition at boot, and not about the host→initrd transition at shutdown. For initrd→host transition asynchronous MNT_DETACH behaviour is good (as initrd is backed by tmpfs), but for the host→initrd transition asynchronous MNT_DETACH behaviour is bad (as we *want* to synchronously wait for the mount to complete).

keyiflerolsun:
> Errors like
> 
> ```
>  [   250.982832 ]  sd-umount[1177]: Failed to unmount /oldroot: Device or resource busy
>  [   250.983401 ]  sd-umount[1178]: Failed to unmount /oldroot/dev: Device or resource busy
>  [   250.984415 ]  shutdown[1]: Failed to finalize file systems, ignoring. 
> ```
> 
> are occurring randomly because of Nvidia drivers. Forcefully unloading them on shutdown prevents the error from occurring.
> 
> Solution:
> 
> Create a script with the following contents
> 
> ```
> #!/bin/sh
> # Remove all Nvidia modules on shutdown
> # to avoid errors like
> # [35509.312925] sd-umoun[56020]: Failed to unmount /oldroot: Device or resource busy
> # [35509.317505] sd-umoun[56021]: Failed to unmount /oldroot/sys: Device or resource busy
> # [35509.322349] shutdown[1]: Failed to finalize file systems, ignoring.
> #
> for MODULE in nvidia_drm nvidia_modeset nvidia_uvm nvidia
>    do
> # we will have to check if the modules are still loaded because they aren't always
>        if lsmod | grep "$MODULE" &> /dev/null ; then
>           rmmod $MODULE
>        fi
>    done  
> ```
> 
> Name this script, for example, `nvidia.shutdown`, make it executable as usual and place it in `/usr/lib/systemd/system-shutdown`.

this is fixed my `Failed to unmount /oldroot: Device or resource busy` error message

PR #23350, opened 2022-05-11
man: improve VtableExample

bluca:
The methods published by the example have a reply in the signature, but
the code was not sending any, so the client gets stuck waiting for a
response that doesn't arrive. Echo back the input string.

Update the object path to follow what would be the canonical format.

Request a service name on the bus, so that the code can be dropped in a
service and it can be dbus-activatable. It also makes it easier to see
on busctl list.

For users getting started with D-Bus these are non-trivial details, so beefing up the example will help them a bit.

### DIFF
--- a/man/vtable-example.c
+++ b/man/vtable-example.c
@@ -9,6 +9,14 @@
 
 #define _cleanup_(f) __attribute__((cleanup(f)))
 
+#define check(x) ({                             \
+  int r = (x);                                  \
+  errno = r < 0 ? -r : 0;                       \
+  printf(#x ": %m\n");                          \
+  if (r < 0)                                    \
+    return EXIT_FAILURE;                        \
+  })
+
 typedef struct object {
   char *name;
   uint32_t number;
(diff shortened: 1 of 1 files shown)


man/vtable-example.c
```diff
@@ -9,13 +9,33 @@
 
 #define _cleanup_(f) __attribute__((cleanup(f)))
 
+#define check(x) ({                             \
+  int r = x;                                    \
```
keszybz: `(x)` ?

man/vtable-example.c
```diff
@@ -9,13 +9,33 @@
...
+  if (sd_bus_message_is_method_call(m,
+                                    "org.freedesktop.systemd.VtableExample",
+                                    "Method4"))
+    return 1;
+
+  const char *string;
+
+  check(sd_bus_message_read(m, "s", &string));
+
+  check(sd_bus_reply_method_return(m, "s", string));
```
keszybz: Maybe remove the two empty lines here… It's better to keep docs shorter.

man/vtable-example.c
bluca: this was just moved from below - but will fix it

man/vtable-example.c
yuwata: Not addressed? Though not a big matter.

bluca: lol I removed 2 empty lines elsewhere

PR #23352, opened 2022-05-11
sd-device-enumerator: fix inverted return value of match_initialized()

yuwata:
Fixes a bug introduced by d8b50e5df7e01983279e70c6d970fb60d053df97.

Fixes https://github.com/systemd/systemd/pull/22662#issuecomment-1124163773.
Fixes RHBZ#2083374 (https://bugzilla.redhat.com/show_bug.cgi?id=2083374).


yuwata:
cc @AdamWill @thom311

AdamWill:
Thanks, I'll check and see if this resolves my issue.

bluca (approved):

AdamWill:
Testing looks good, ran the problematic openQA test 20 times with a base image that includes a systemd build with this fix, passed 20 times. Without this fix it usually fails at least 1 in 5 times.

PR #23353, opened 2022-05-11
systemd-escape: add --object option

bluca:
Allows to encode/decode unit names in the D-Bus object path format.

$ systemd-escape -o /org/freedesktop/systemd1/unit/dbus-broker.service
/org/freedesktop/systemd1/unit/dbus_2dbroker_2eservice

### DIFF
--- a/man/systemd-escape.xml
+++ b/man/systemd-escape.xml
@@ -95,6 +95,15 @@
       </varlistentry>
 
       <varlistentry>
+        <term><option>--object</option></term>
+        <term><option>-o</option></term>
+
+        <listitem><para>When escaping or unescaping a string, assume it refers to a D-Bus object path. This
+        eliminates (or adds) escape sequences that are necessary to encode/decode in a D-Bus-friendly manner.
+        </para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><option>--unescape</option></term>
         <term><option>-u</option></term>
 
@@ -149,6 +158,10 @@ Hallöchen, Meister</programlisting>
     <para>To generate the mount unit for a path:</para>
     <programlisting>$ systemd-escape -p --suffix=mount "/tmp//waldi/foobar/"
 tmp-waldi-foobar.mount</programlisting>
+
+    <para>To generate the D-Bus object path for a unit:</para>
+    <programlisting>$ systemd-escape -o /org/freedesktop/systemd1/unit/dbus-broker.service
+/org/freedesktop/systemd1/unit/dbus_2dbroker_2eservice</programlisting>
 
     <para>To generate instance names of three strings:</para>
     <programlisting>$ systemd-escape --template=systemd-nspawn@.service 'My Container 1' 'containerb' 'container/III'
--- a/src/escape/escape.c
+++ b/src/escape/escape.c
@@ -21,6 +21,7 @@ static enum {
 static const char *arg_suffix = NULL;
 static const char *arg_template = NULL;
 static bool arg_path = false;
+static bool arg_object = false;
 static bool arg_instance = false;
 
 static int help(void) {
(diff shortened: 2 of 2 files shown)


yuwata:
typo in commit message: systemd-encode -> systemd-escape

man/systemd-escape.xml
```diff
@@ -94,6 +94,15 @@
         </para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--object</option></term>
+        <term><option>-o</option></term>
+
+        <listitem><para>When escaping or unescaping a string, assume it refers to a D-Bus object path. This
+        eliminates (or adds) escape sequences that are necessary to encode/decode in a D-Bus-frienly manner.
```
yuwata: frienly -> friendly?

yuwata (approved):
LGTM.

yuwata:
CIs are red, as the help message is not updated.

poettering:
hmm, that's awfully specific, no? what's the usecase?

poettering:
Does this really deserve a short option? Sounds too low-level for that?

I don't understand the example you are giving: it's already an object path?

You mention this is for units, so shouldn't it prefix the /org/freedesktop/systemd1/units/ prefix?

I don't like that this uses path_extract_directory() for this, given these are not paths in the fs, but behave quite differently (i.e. "." and ".." and "//" are not a thing in dbus paths, and trailing slash is not ok, and leading slash missing is not ok either). I am pretty sure any such feature should generate an error if the result would be an invalid dbus object path...

bluca:
> hmm, that's awfully specific, no? what's the usecase?

AFAIK there's no way to un-escape dbus objects in the terminal. So if I do something like `busctl call ... GetUnitByPID ...` I get back an escaped unit name. I want to have a way to un-escape it in the shell.

poettering:
sd-bus actually exposes generic bus path escaping functions. It appears to me this feature should not be in systemd-escape, but in busctl hence.

In fact, it should be a concept of "mangling", i.e. if you pass an object path that is invalid, we should automatically make it valid if we can, by escaping each label. 

See sd_bus_path_*() apis

poettering:
and then we should log about this if we do this, i.e. a LOG_NOTICE level message "Hey, you passed an invalid bus path xyz, we mangled it into foo for you automatically, hope you like it."

bluca:
You mean adding `busctl [un]escape` verbs?

poettering:
no, i mean that all those verbs like "busctl call" or "busctl get-property" and so on, should check their object path parameter: if it's a valid bus path, then just use it. If it's invalid, then split it at "/", escape each component, glue it together again, prefix a / if missing, remove one at the end if there is one, then check if that's a valid bus path now. If so, use it, log about the fact the bus path was just mangled, and proceed.

poettering:
or in other words: don't require the user to manually combine a systemd-escape invocation with a busctl invocation. Instead just do the escaping/mangling automatically if it appears to be needed.

bluca:
> or in other words: don't require the user to manually combine a systemd-escape invocation with a busctl invocation. Instead just do the escaping/mangling automatically if it appears to be needed.

But how do I know a returned string needs to be un-escaped? I can't just un-escape any string, there's no way to know if one was a dbus path or if it just matches some characters... Did you mean to have a `--mangle` option to make it optional?

poettering:
why would you want to unescape an object path returned by some bus service? object paths at that point should really be considered opaque handles i guess...

bluca:
> why would you want to unescape an object path returned by some bus service? object paths at that point should really be considered opaque handles i guess...

Because the GetUnitByPid method returns an object path, and i need to get the unit name from it

poettering:
> Because the GetUnitByPid method returns an object path, and i need to get the unit name from it

you could query the "Id" property from that path...

Maybe we should introduce GetUnitByPid2() btw, that returns object path + unit name + invocation id. In particular the latter actually matters I think, since after all a PID is member of a specific invocation of a unit, and not the unit generally...

bluca:
> > Because the GetUnitByPid method returns an object path, and i need to get the unit name from it
> 
> you could query the "Id" property from that path...
> 
> Maybe we should introduce GetUnitByPid2() btw, that returns object path + unit name + invocation id. In particular the latter actually matters I think, since after all a PID is member of a specific invocation of a unit, and not the unit generally...

Yeah that works, I can use that instead, thanks

PR #23358, opened 2022-05-12
fuzzers: add input size limits, always configure limits in two ways

keszybz:
Without the size limits, oss-fuzz creates huge samples that time out. Usually
this is because some of our code has bad algorithmic complexity. For data like
configuration samples we don't need to care about this: non-rogue configs are
rarely more than a few items, and a bit of a slowdown with a few hundred items
is acceptable. This wouldn't be OK for processing of untrusted data though.

We need to set the limit in two ways: through .options and in the code. The
first because it nicely allows libFuzzer to avoid wasting time, and the second
because fuzzers like hongfuzz and afl don't support .options.

While at it, let's fix an off-by-one (65535 is the largest offset for a
power-of-two size, but we're checking the size here).

Co-authored-by: Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>

### DIFF
--- a/src/core/fuzz-unit-file.c
+++ b/src/core/fuzz-unit-file.c
@@ -21,7 +21,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         const char *name;
         long offset;
 
-        if (size > 65536)
+        if (outside_size_range(size, 0, 65536))
                 return 0;
 
         f = data_to_file(data, size);
--- a/meson.build
+++ b/meson.build
@@ -50,6 +50,11 @@ endif
 skip_deps = want_ossfuzz or get_option('skip-deps')
 fuzzer_build = want_ossfuzz or want_libfuzzer
 
+# If we're building *not* for actual fuzzing, allow input samples of any size
+# (for testing and for reproduction of issues discovered with previously-higher
+# limits).
+conf.set10('FUZZ_USE_SIZE_LIMIT', fuzzer_build)
+
 # Create a title-less summary section early, so it ends up first in the output.
 # More items are added later after they have been detected.
 summary({'build mode' : get_option('mode')})
--- a/src/boot/efi/fuzz-bcd.c
+++ b/src/boot/efi/fuzz-bcd.c
@@ -11,7 +11,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         _cleanup_free_ void *p = NULL;
 
         /* This limit was borrowed from src/boot/efi/boot.c */
-        if (size > 100*1024)
+        if (outside_size_range(size, 0, 100*1024))
                 return 0;
 
         if (!getenv("SYSTEMD_LOG_LEVEL"))
--- a/src/fuzz/fuzz-bootspec.c
+++ b/src/fuzz/fuzz-bootspec.c
@@ -84,7 +84,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         _cleanup_(boot_config_free) BootConfig config = BOOT_CONFIG_NULL;
         int r;
 
-        if (size > 65536)
+        if (outside_size_range(size, 0, 65536))
                 return 0;
 
         /* Disable most logging if not running standalone */
--- a/src/fuzz/fuzz-env-file.c
+++ b/src/fuzz/fuzz-env-file.c
@@ -12,7 +12,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         _cleanup_fclose_ FILE *f = NULL;
         _cleanup_strv_free_ char **rl = NULL, **rlp =  NULL;
 
-        if (size > 65536)
+        if (outside_size_range(size, 0, 65536))
                 return 0;
 
         f = data_to_file(data, size);
--- a/src/fuzz/fuzz.h
+++ b/src/fuzz/fuzz.h
@@ -4,6 +4,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include "env-util.h"
 #include "fileio.h"
 
 /* The entry point into the fuzzer */
@@ -14,4 +15,15 @@ static inline FILE* data_to_file(const uint8_t *data, size_t size) {
                 return fopen("/dev/null", "re");
         else
                 return fmemopen_unlocked((char*) data, size, "re");
+}
+
+/* Check if we are within the specified size range.
+ * The upper limit is ignored if FUZZ_USE_SIZE_LIMIT is unset.
+ */
+static inline bool outside_size_range(size_t size, size_t lower, size_t upper) {
+        if (size < lower)
+                return true;
+        if (size > upper)
+                return FUZZ_USE_SIZE_LIMIT;
+        return false;
 }
--- a/src/journal-remote/fuzz-journal-remote.c
+++ b/src/journal-remote/fuzz-journal-remote.c
@@ -24,7 +24,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         _cleanup_(journal_remote_server_destroy) RemoteServer s = {};
         int r;
 
-        if (size <= 2 || size > 65536)
+        if (outside_size_range(size, 3, 65536))
                 return 0;
 
         if (!getenv("SYSTEMD_LOG_LEVEL"))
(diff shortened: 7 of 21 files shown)


keszybz:
Rescued from https://github.com/systemd/systemd/pull/21939.

yuwata:
Ah, nice. Thank you.

evverx:
As I mentioned I'm not a fan of arbitrary limits. They can hide real issues: https://googleprojectzero.blogspot.com/2021/12/this-shouldnt-have-happened.html. I think it's better to have a few bogus issues opened than to miss real bugs

evverx:
By the way I wonder why the fuzz targets seem to have been getting much more attention lately? It's not that I'm opposed to it. It's just unexpected.

src/libsystemd-network/fuzz-dhcp-server.c
```diff
@@ -54,7 +54,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         struct in_addr address = { .s_addr = htobe32(UINT32_C(10) << 24 | UINT32_C(1))};
         _cleanup_free_ uint8_t *duped = NULL;
 
-        if (size < sizeof(DHCPMessage))
+        if (size < sizeof(DHCPMessage) || size > 600)
```
evverx: I wonder where 600 comes from?

src/core/fuzz-unit-file.c
```diff
@@ -21,7 +21,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         const char *name;
         long offset;
 
+        if (size > 65536)
```
evverx: https://github.com/systemd/systemd/pull/21939#issuecomment-1003162677

src/libsystemd-network/fuzz-dhcp-server.c
keszybz: From the .options file. It was added thus in the initial commit in 2bd37c5be2f09950f26351506a00929c8719ea15.

keszybz:
> By the way I wonder why the fuzz targets seem to have been getting much more attention lately? It's not that I'm opposed to it. It's just unexpected.

I added some new fuzzers and coverage and I'm also trying to clean up extant issues.

src/libsystemd-network/fuzz-dhcp-server.c
evverx: I don't know why it was added there either. But since the file has never affected AFL and honggfuzz it effectively has been fuzzed without any arbitrary limits. As far as I can remember the dns packets were limited somewhere for no apparent reason as well.

keszybz:
> As I mentioned I'm not a fan of arbitrary limits. They can hide real issues: https://googleprojectzero.blogspot.com/2021/12/this-shouldnt-have-happened.html. I think it's better to have a few bogus issues opened than to miss real bugs

Please see the commit message. I'm not setting a blanket limit everywhere, but instead a limit for the cases where we know that it's easy to generate quadratic behaviour with some parser and we don't really care about that. Without the limit oss-fuzz is going to keep finding new cases that don't do us any good. We know that `strv_append()` is slow, and there just isn't anything interesting to discover here.

evverx:
> I added some new fuzzers and coverage and I'm also trying to clean up extant issues.

I saw those PRs and I think it's great that the fuzzing infrastructure is documented much better now. Thanks! I've been trying to add a link to the coverage report since January unsuccessfully. What I was trying to ask if what prompted it.

> Without the limit oss-fuzz is going to keep finding new cases that don't do us any good

OSS-Fuzz isn't going to report all of them though. All timeouts are compressed into one issue per a fuzz target. Most of them have been opened for years by the way and I don't think it was very annoying.

keszybz:
> I've been trying to add a link to the coverage report since January unsuccessfully.

Is there something that is blocked on review or something like that? (Sorry, it's not clear if "trying" implies some issue that I can help with.)

> What I was trying to ask if what prompted it.

Nothing in particular. I started working on fuzz-bootspec, and then saw some low-hanging fruit, and so it went.

> > Without the limit oss-fuzz is going to keep finding new cases that don't do us any good
> 
> OSS-Fuzz isn't going to report all of them though. All timeouts are compressed into one issue per a fuzz target. Most of them have been opened for years by the way and I don't think it was very annoying.

That's still not useful though. Oss-fuzz has some limited amount of cpu cycles it will devote to our fuzzers. Whatever fraction of that time it spends on busy-looping on large inputs for code that does poorly with large inputs is essentially time wasted. The docs for oss-fuzz say that a single run should be < 1 ms. If it's spending 30s or 90s on one input, that is a *lot* of other inputs it didn't try. And we actually care about the inputs where it finds something wrong with smaller size.

Also, it's easier to manage issues if the list is empty (except for new stuff).

evverx:
> Is there something that is blocked on review or something like that?

No there isn't. Sorry. What I was trying to say it that unlike your PRs that PR stalled and the documentation has never been updated. With your PRs merged the documentation is much better now.

> Oss-fuzz has some limited amount of cpu cycles it will devote to our fuzzers. Whatever fraction of that time it spends on busy-looping on large inputs for code that does poorly with large inputs is essentially time wasted

Clusterfuzz is smart enough to deduce meaningful limits itself. It doesn't just keep fuzzing stuff like that for eternity so I don't think it's an issue.

Anyway, it would be great if at least the fuzz targets handling network packets could keep fuzz stuff without those arbitrary limits. Unlike local file parsers it wouldn't be great to miss new buffer overflows just because timeouts can be a bit annoying.

keszybz:
> Anyway, it would be great if at least the fuzz targets handling network packets could keep fuzz stuff without those arbitrary limits. Unlike local file parsers it wouldn't be great to miss new buffer overflows just because timeouts can be a bit annoying.

That makes sense. But then I think the limits should be raised (or dropped altogether) consistently, i.e. both in the code and .options files. Let's merge this PR (which doesn't add any of those limits), and then handle network stuff in later PRs.

bluca (approved):

evverx:
> which doesn't add any of those limits

I does though. `fuzz-dhcp-server`, `fuzz-dhcp6-client` and so on have been updated as far as I can see.

src/libsystemd-network/fuzz-dhcp-server.c
```diff
@@ -54,7 +54,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         struct in_addr address = { .s_addr = htobe32(UINT32_C(10) << 24 | UINT32_C(1))};
         _cleanup_free_ uint8_t *duped = NULL;
 
-        if (size < sizeof(DHCPMessage))
+        if (outside_size_range(size, sizeof(DHCPMessage), 600))
```
evverx: I don't think this limit should be added here. I think the bogus options file should be removed instead

evverx (changes requested):

src/libsystemd-network/fuzz-dhcp-server.c
keszybz: OK, I'll remove the options file instead.

yuwata:
CIFuzz is red...

evverx:
Looks like the condition in `fuzz-compress` should be reverted. Having said that that fuzz target has never actually fuzzed anything (because it depends on external libraries like `zstd`, `zlib` and so on that should be linked statically with the fuzz target and `systemd` intentionally doesn't support use cases like that) so maybe it should be just dropped.

Anyway, I still can't say I'm happy about introducing those limits but as long as the "network" fuzz targets are hopefully intact I can live with it.

It would be great if CIFuzz could report "coverage" (https://github.com/google/oss-fuzz/issues/7190) to figure out how those limits affect coverage but I supposed it can be tracked manually as usual

src/core/fuzz-unit-file.c
```diff
@@ -21,7 +21,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
         const char *name;
         long offset;
 
+        if (outside_size_range(size, 0, 65536))
```
evverx, line 24: Looking at the parser it appears at list one corner case is no longer tested by the fuzzer because LONG_LINE_MAX is 1024 * 1024. I suspect most corner cases like that are no longer covered with those arbitrary limits in place which makes the fuzz targets less useful than they could probably be I think. I'm not sure those timeouts are worth it to be honest.

evverx: I mean it should be possible to scale down those constants when the fuzz targets are built if the idea is to keep discovering corner cases without handling too much data but that's not what happened here and in PRs where the other arbitrary limits were introduced as far as I can see. I'm not sure how I missed them. Then again, those timeouts aren't exactly an issue that needs fixing I think

Issue #22826, opened 2022-03-22
RFE: Provide templates for creating new systemd unit files


danimo:
**Is your feature request related to a problem? Please describe.**

The best way for a user to create a new unit/service file is `systemctl edit --full --force my.service`. The rest is left to the user, who needs to look up even basic boiler plating for writing a standard unit file. As a result, this often does not happen ([annecdotic evidence](https://twitter.com/fluepke/status/1506199342369193985)).

**Describe the solution you'd like**

1. `systemctl edit --full --force my.service` should look up a template from a template folder, depending on the unit type, and provide it to the editor environment. The elegance of a template folder is that it can be customized by downstream with ease.
2.  Consider introducing `systemctl create` or `systemctl create-unit` as a shorthand for `systemctl edit --full --force`

**Describe alternatives you've considered**

Writing a wrapper script. But it's never there where and when it is needed, so it fails to solve the issue.

**The systemd version you checked that didn't have the feature you are asking for**

I checked `master`, and Lennart asked me to file this RFE, so I guess it's safe to assume it's not there?

keszybz:
I think `systemctl edit --new` would be better than a completely new top-level verb.

I think this would be useful. We'd just need to add code to look in a template directory for new units, which shouldn't be too hard, and then start populating that directory.

poettering:
Yeah, "--full --force" is really weird wording for creating a new file. --new makes a lot of sense to me.

DaanDeMeyer:
The relevant source code to modify will be in src/systemctl/systemctl.h, src/systemctl/systemctl.c, src/systemctl/systemctl-edit.h and src/systemctl/sytemctl-edit.c.

As discussed above, you should add a `--new` option to systemctl. When this is specified, we'll look in a directory for template files (e.g. /usr/share/systemd/unit/template) for a template file, and pre-populate the contents of the new unit file with the contents from the template file. To select the correct template, --new could take a string argument that allows users to specify which template they want to use. If no string argument is given for --new, we could list the possible templates the user can choose from.

To add support for --new, you'll need to modify `systemctl_parse_argv()` in systemctl.c. You can add a new global variable (this isn't frowned upon for argument variables) `arg_new` that is populated when the --edit option is specified. The code to look up the template and use it to prepopulate the unit file should go somewhere in systemctl-edit.c.

This should be enough information to get started. Take a look at the source code and try some stuff out. If you come up with any more questions, you can ask them here.

cerebro1:
[...]

PR #23359, opened 2022-05-12
systemctl : add --new to create new unit using template

cerebro1:
https://github.com/systemd/systemd/issues/22826

### DIFF
--- a/src/systemctl/systemctl-edit.c
+++ b/src/systemctl/systemctl-edit.c
@@ -374,7 +374,7 @@ static int run_editor(char **paths) {
         return 0;
 }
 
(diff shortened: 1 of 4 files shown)


cerebro1:
@bluca When --new will be introduced, we plan on removing --full --force option from systemctl-edit or are going to keep it as backward compatibility ?
--full and --force have significance at other verb so they can not be completely removed. That i why i want to understand if i can only remove it from systemctl-edit since all the features will be available with --new
- create unit with default template
- create unit by passing existing template

bluca:
We can't break backward compat, so the existing functionality must remain too

bluca:
we have moved on since this, so let's close for now, and eventually reopen later if it becomes needed again

PR #23364, opened 2022-05-12
Add a "test" that prints the SBAT table

keszybz:
The SBAT section was included in a special section in the EFI code, but
the contents weren't directly visible in any way. Let's add a "test" that
prints them for visual inspection.

If there's some external linter for this format, we could hook it up in the
future.

### DIFF
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -3,6 +3,8 @@
 conf.set10('ENABLE_EFI', get_option('efi'))
 conf.set10('HAVE_GNU_EFI', false)
 
+efi_config_h_dir = meson.current_build_dir()
+
 if not get_option('efi') or get_option('gnu-efi') == 'false'
         if get_option('gnu-efi') == 'true'
                 error('gnu-efi support requested, but general efi support is disabled')
--- a/src/test/meson.build
+++ b/src/test/meson.build
@@ -478,6 +478,10 @@ tests += [
 
         [files('test-date.c')],
 
+        [files('test-sbat.c'),
+         [], [], [], 'HAVE_GNU_EFI', '',
+         ['-I@0@'.format(efi_config_h_dir)]],
+
         [files('test-sleep.c')],
 
         [files('test-tpm2.c')],
--- a/src/test/test-sbat.c
+++ b/src/test/test-sbat.c
@@ -1,0 +1,23 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+/* We include efi_config.h after undefining PROJECT_VERSION which is also defined in config.h. */
+#undef PROJECT_VERSION
+#include "efi_config.h"
+
+#include "build.h"
+#include "sbat.h"
+#include "tests.h"
+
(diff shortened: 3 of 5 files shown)


src/test/test-sbat.c
```diff
@@ -0,0 +1,22 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#undef PROJECT_VERSION
+#include "efi_config.h"
```
bluca, line 5: sorting?

bluca (approved):

yuwata:
```
src/test/meson.build:43:0: ERROR: Unknown variable "efi_config_h_dir".
```

src/boot/efi/meson.build
```diff
@@ -171,6 +171,7 @@ endif
 efi_config_h = configure_file(
         output : 'efi_config.h',
         configuration : efi_conf)
+efi_config_h_dir = meson.current_build_dir()
```
yuwata: I think it is better to move the test at src/boot/efi.

src/test/meson.build
```diff
@@ -478,6 +478,10 @@ tests += [
 
         [files('test-date.c')],
 
+        [files('test-sbat.c'),
+         [], [], [], 'ENABLE_EFI', '',
+         ['-I@0@'.format(efi_config_h_dir)]],
```
medhefgo, line 483: You can just do `'-include', efi_config_h`, no need for this extra var, no?

keszybz: No, this doesn't work because both `config.h` and `efi_config.h` define `PROJECT_VERSION`. And `-include config.h -UPROJECT_VERSION -include efi_config.h` doesn't work as expected. Thus I changed to have `#undef` and `#include` in the file itself.

src/boot/efi/meson.build
keszybz: All tests for src/fundamental are in src/test, and also src/boot/efi is for efi code. So I think this place is better.

src/test/test-sbat.c
keszybz: It needs to be first.

src/boot/efi/meson.build
yuwata: ~~Or, src/boot? Then, conditions for adding the test may become simpler. Currently, it is necessary to check the build supports EFI. Otherwise the above CI failure is triggered.~~

medhefgo:
Alternatively, you could assemble the sbat text in meson and pass it to config and efi_config…

src/test/meson.build
```diff
@@ -478,6 +478,10 @@ tests += [
 
         [files('test-date.c')],
 
+        [files('test-sbat.c'),
+         [], [], [], 'ENABLE_EFI', '',
```
medhefgo: This needs to depend on `HAVE_GNU_EFI`.

keszybz: Updated with that changed only.

yuwata (approved):

PR #23365, opened 2022-05-12
CI: use Fedora 36

yuwata:


mrc0mmand (approved):

mkosi.default.d/fedora/10-mkosi.fedora
```diff
@@ -65,6 +65,7 @@ BuildPackages=
         /usr/bin/xsltproc
 
 Packages=
+        acl
```
mrc0mmand, line 67: Hmm, CI still complains:

```
/usr/lib/systemd/systemd: error while loading shared libraries: libip4tc.so.2: cannot open shared object file: No such file or directory
Error: Process completed with exit code 127.
```

I guess we should either drop `pkgconfig(libiptc)` from the `BuildPackages=` or add `iptables-legacy` to the `Packages=`.

yuwata: Thanks. libiptc is now dropped.

yuwata:
mkosi CI for fedora 36 is green now. Merging.

PR #23366, opened 2022-05-12
network: do not update interface group by default

yuwata:
Follow-up for 10af8bb24b39a815079f6bf31b449c6e5aaa2adf.


DaanDeMeyer (approved):
LGTM but not sure if we should merge too much stuff after the rc

yuwata:
> LGTM but not sure if we should merge too much stuff after the rc

This fixes a bug introduced by 10af8bb24b39a815079f6bf31b449c6e5aaa2adf. Before the commit, interface group was set only when `Group=` is explicitly specified, otherwise the interface group was kept. However, after the commit, we need to specify `Group=` with an empty string to keep the current interface group.

I will also mention that in the commit message.

DaanDeMeyer:
That makes sense, then this can go in, setting green label

PR #23367, opened 2022-05-13
core/mount: fix dependency issues

### LINKED ISSUE #19983: systemd merges deps of mount unit files and /proc/self/mountinfo, leading to surprising results: systemd automatically unmounts manually mounted filesystems [Take #2]
PR https://github.com/systemd/systemd/pull/19322 is supposed to fix issue https://github.com/systemd/systemd/issues/9869 but AFAICS it's not the case.

IIUC StopPropagatedFrom= is supposed to address the aforementioned issue but I fail to see how exactly: as long as the dependency on the wrong device is still included in the mount dependencies, the problem isn't solved.

The fact that the dependency on the device is now edge-triggered doesn't help, it just moves the problem when the wrong device (the one PID1 initially tracked when the generated mount unit was loaded) is being stopped now.

yuwata:
Hopefully fixes #19983 and #23552.

### DIFF
--- a/test/units/testsuite-60.sh
+++ b/test/units/testsuite-60.sh
@@ -3,6 +3,155 @@
 set -eux
 set -o pipefail
 
+# shellcheck source=test/units/assert.sh
+. "$(dirname "$0")"/assert.sh
+
+teardown_test_dependencies() (
+    set +eux
+
+    if mountpoint /tmp/deptest; then
+        umount /tmp/deptest
+    fi
+
+    if [[ -n "${LOOP}" ]]; then
+        losetup -d "${LOOP}" || :
+    fi
+    if [[ -n "${LOOP_0}" ]]; then
+        losetup -d "${LOOP_0}" || :
+    fi
+    if [[ -n "${LOOP_1}" ]]; then
+        losetup -d "${LOOP_1}" || :
+    fi
+
+    rm -f /tmp/testsuite-60-dependencies-0.img
+    rm -f /tmp/testsuite-60-dependencies-1.img
+
+    rm -f /run/systemd/system/tmp-deptest.mount
+    systemctl daemon-reload
+
+    return 0
+)
+
+setup_loop() {
+    truncate -s 30m "/tmp/testsuite-60-dependencies-${1?}.img"
+    sfdisk --wipe=always "/tmp/testsuite-60-dependencies-${1?}.img" <<EOF
+label:gpt
+
+name="loop${1?}-part1"
+EOF
+    LOOP=$(losetup -P --show -f "/tmp/testsuite-60-dependencies-${1?}.img")
+    udevadm wait --settle --timeout=10 "${LOOP}p1"
+    udevadm lock --device="${LOOP}p1" mkfs.ext4 -L "partname${1?}-1" "${LOOP}p1"
+}
+
+check_dependencies() {
+    local escaped_0 escaped_1 after
+
+    escaped_0=$(systemd-escape -p "${LOOP_0}p1")
+    escaped_1=$(systemd-escape -p "${LOOP_1}p1")
+
+    if [[ -f /run/systemd/system/tmp-deptest.mount ]]; then
+        after=$(systemctl show --property=After --value tmp-deptest.mount)
+        assert_not_in "local-fs-pre.target" "$after"
+        assert_in "remote-fs-pre.target" "$after"
+        assert_in "network.target" "$after"
+    fi
+
+    # mount LOOP_0
+    mount -t ext4 "${LOOP_0}p1" /tmp/deptest
+    sleep 1
+    after=$(systemctl show --property=After --value tmp-deptest.mount)
+    assert_in "local-fs-pre.target" "$after"
+    assert_not_in "remote-fs-pre.target" "$after"
+    assert_not_in "network.target" "$after"
+    assert_in "${escaped_0}.device" "$after"
+    assert_in "blockdev@${escaped_0}.target" "$after"
+    assert_not_in "${escaped_1}.device" "$after"
+    assert_not_in "blockdev@${escaped_1}.target" "$after"
+    umount /tmp/deptest
+
+    if [[ -f /run/systemd/system/tmp-deptest.mount ]]; then
+        after=$(systemctl show --property=After --value tmp-deptest.mount)
+        assert_not_in "local-fs-pre.target" "$after"
+        assert_in "remote-fs-pre.target" "$after"
+        assert_in "network.target" "$after"
+    fi
+
+    # mount LOOP_1 (using fake _netdev option)
+    mount -t ext4 -o _netdev "${LOOP_1}p1" /tmp/deptest
+    sleep 1
+    after=$(systemctl show --property=After --value tmp-deptest.mount)
+    assert_not_in "local-fs-pre.target" "$after"
+    assert_in "remote-fs-pre.target" "$after"
+    assert_in "network.target" "$after"
+    assert_not_in "${escaped_0}.device" "$after"
+    assert_not_in "blockdev@${escaped_0}.target" "$after"
+    assert_in "${escaped_1}.device" "$after"
+    assert_in "blockdev@${escaped_1}.target" "$after"
+    umount /tmp/deptest
+
+    if [[ -f /run/systemd/system/tmp-deptest.mount ]]; then
+        after=$(systemctl show --property=After --value tmp-deptest.mount)
+        assert_not_in "local-fs-pre.target" "$after"
+        assert_in "remote-fs-pre.target" "$after"
+        assert_in "network.target" "$after"
+    fi
+
+    # mount tmpfs
+    mount -t tmpfs tmpfs /tmp/deptest
+    sleep 1
+    after=$(systemctl show --property=After --value tmp-deptest.mount)
+    assert_in "local-fs-pre.target" "$after"
+    assert_not_in "remote-fs-pre.target" "$after"
+    assert_not_in "network.target" "$after"
+    assert_not_in "${escaped_0}.device" "$after"
+    assert_not_in "blockdev@${escaped_0}.target" "$after"
+    assert_not_in "${escaped_1}.device" "$after"
+    assert_not_in "blockdev@${escaped_1}.target" "$after"
+    umount /tmp/deptest
+
+    if [[ -f /run/systemd/system/tmp-deptest.mount ]]; then
+        after=$(systemctl show --property=After --value tmp-deptest.mount)
+        assert_not_in "local-fs-pre.target" "$after"
+        assert_in "remote-fs-pre.target" "$after"
+        assert_in "network.target" "$after"
+    fi
+}
+
+test_dependencies() {
+    if systemd-detect-virt --quiet --container; then
+        echo "Skipping test_dependencies in container"
+        return
+    fi
+
+    trap teardown_test_dependencies RETURN
+
+    setup_loop 0
+    LOOP_0="${LOOP}"
+    LOOP=
+    setup_loop 1
+    LOOP_1="${LOOP}"
+    LOOP=
+
+    mkdir -p /tmp/deptest
+
+    # without .mount file
+    check_dependencies
+
+    # create .mount file
+    mkdir -p /run/systemd/system
+    cat >/run/systemd/system/tmp-deptest.mount <<EOF
+[Mount]
+Where=/tmp/deptest
+What=192.168.0.1:/tmp/mnt
+Type=nfs
+EOF
+    systemctl daemon-reload
+
+    # with .mount file
+    check_dependencies
+}
+
 test_issue_20329() {
     local tmpdir unit
     tmpdir="$(mktemp -d)"
@@ -107,6 +256,9 @@ fi
 # Verify that the mount units are always cleaned up at the end.
 # Give some time for units to settle so we don't race between exiting the rate limit state and cleaning up the units.
 timeout 2m bash -c 'while systemctl list-units -t mount tmp-meow* | grep -q tmp-meow; do systemctl daemon-reload; sleep 10; done'
+
+# test for issue #19983 and #23552.
+test_dependencies
 
 # test that handling of mount start jobs is delayed when /proc/self/mouninfo monitor is rate limited
 test_issue_20329


yuwata:
TEST-60-MOUNT-RATELIMIT failed on CentOS CI. @mrc0mmand Is this new? Could you restart the CI?

mrc0mmand:
> TEST-60-MOUNT-RATELIMIT failed on CentOS CI. @mrc0mmand Is this new? Could you restart the CI?

Haven't seen this one yet, I'll restart the job and look into the fail.

yuwata:
> > TEST-60-MOUNT-RATELIMIT failed on CentOS CI. @mrc0mmand Is this new? Could you restart the CI?
> 
> Haven't seen this one yet, I'll restart the job and look into the fail.

Thanks.

DaanDeMeyer (approved):

DaanDeMeyer:
Should this go into v251? I'm tempted to include it but we already released rc3 so it's a bit risky unless we want to do another rc

yuwata:
> Should this go into v251? I'm tempted to include it but we already released rc3 so it's a bit risky unless we want to do another rc

I added three more commits and reordered commits.
I think the first three commits are trivial and safe, but at the same time they do not have any merits to be included in v251.
The last three commits should be important, and hopefully fix #19983. But yeah, it is risky to be merged before v251.

yuwata:
> > TEST-60-MOUNT-RATELIMIT failed on CentOS CI. @mrc0mmand Is this new? Could you restart the CI?
> 
> Haven't seen this one yet, I'll restart the job and look into the fail.

I found that the test failed on another PR.
https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-vagrant-archlinux/11795/artifact/systemd-centos-ci/index.html
Hence, the previous failure is not related to this PR.

DaanDeMeyer:
@yuwata Can you add more detail to the commit messages of the last 3 commits? It's a bit hard to follow otherwise.

yuwata:
> @yuwata Can you add more detail to the commit messages of the last 3 commits? It's a bit hard to follow otherwise.

Sure. Will do. Sorry, I am often lazy about writing explanation in commit message...

fbuihuu:
> > @yuwata Can you add more detail to the commit messages of the last 3 commits? It's a bit hard to follow otherwise.
> 
> Sure. Will do. Sorry, I am often lazy about writing explanation in commit message...

I agree, commit messages are not enough detailed. I know it's much work for you but otherwise the extra work is put on reviewer's shoulder ;)

fbuihuu:
> Hopefully fixes #19983.

It seems to me that the actual fix is " core/mount: set Mount.from_proc_self_mountinfo flag before adding default dependencies", no ?

With this patch the device deps are correctly cleaned up when the mount unit enters in dead state and hence it doesn't keep refs on the "old" device.

So I'm not sure to see why you tagged the last commit instead as fix for #19983.

src/core/mount.c
```diff
@@ -1646,7 +1646,7 @@ static int mount_setup_existing_unit(
                 /* If things changed, then make sure that all deps are regenerated. Let's
                  * first remove all automatic deps, and then add in the new ones. */
 
-                unit_remove_dependencies(u, UNIT_DEPENDENCY_MOUNTINFO_IMPLICIT);
+                unit_remove_dependencies(u, UNIT_DEPENDENCY_MASK_MOUNTINFO);
```
fbuihuu: So basically you now remove the deps involved with `DefaultDependencies=`. Given that these deps depend on the presence of a unit file, I'm not sure to follow...

Could you explain ? (and add the rational in the commit message ?)

fbuihuu: Also I'm not sure that hiding the list of bits(only 2 in this case) behind a new "mask" macro helps in term of readability.

src/core/unit.h
```diff
@@ -86,15 +86,19 @@ typedef enum UnitDependencyMask {
          * DefaultDependencies= and thus also involving configuration from UNIT_DEPENDENCY_FILE sources */
         UNIT_DEPENDENCY_MOUNTINFO_DEFAULT  = 1 << 6,
 
-        UNIT_DEPENDENCY_MASK_MOUNTINFO     = UNIT_DEPENDENCY_MOUNTINFO_IMPLICIT | UNIT_DEPENDENCY_MOUNTINFO_DEFAULT,
+        /* A dependency created because of data read from /proc/self/mountinfo, but fallback to unit configuration
+         * sources */
+        UNIT_DEPENDENCY_MOUNTINFO_OR_FILE  = 1 << 7,
```
fbuihuu, line 91: Can you explain the need for this new flag please, I'm not following. What problem is this commit supposed to solve exactly ?

yuwata:
@DaanDeMeyer and @fbuihuu Sorry for late. The commit messages are updated. PTAL.

> > Hopefully fixes #19983.
> 
> It seems to me that the actual fix is " core/mount: set Mount.from_proc_self_mountinfo flag before adding default dependencies", no ?
> 
> With this patch the device deps are correctly cleaned up when the mount unit enters in dead state and hence it doesn't keep refs on the "old" device.
> 
> So I'm not sure to see why you tagged the last commit instead as fix for #19983.

Does it? I did not expect that. Interesting.

> Also not sure to see why you added the `fstab-generator` label to this PR. `pid1` seems to be enough to me.

The first commit is a trivial cleanup for fstab-generator.

src/core/mount.c
keszybz: I agree with @fbuihuu that the new mask is not helpful. I think this will be clearer with an explicit OR of the two values.

> Could you explain ? (and add the rational in the commit message ?)

I'd phrase the commit as the following:
"""
Dependencies tagged with UNIT_DEPENDENCY_MOUNTINFO_DEFAULT are created
based on /proc/self/mountinfo, but are conditionalized by DefaultDependencies= in the corresponding .mount unit file. Hence, when information from mountinfo is changed, e.g. device mounted
on the path, we need to update dependencies of both types (UNIT_DEPENDENCY_MOUNTINFO_DEFAULT
and  UNIT_DEPENDENCY_MOUNTINFO_IMPLICIT).
"""

src/core/unit.h
keszybz: Yeah, this seems strange. This flag seems to be essentially saying "we don't know where this dependency came from exactly". It breaks the orthogonality of the different flags. After reading the commit message, I'd expect that we simply propagate the information about the appropriate mask from the caller down into the functions adding the dependencies. If they are added based on the unit file, use one mask; if they are added based on mountinfo, use the other mask.

keszybz (commented):
This needs a rebase now…

d-hatayama:
Let me write the same thing I wrote at https://github.com/systemd/systemd/issues/23552#issuecomment-1143410198 in case missed reading, but  in more detail with logs.

With these commits, #23552 is apparently fixed but there is still issue. Please look at the following trace log I collected using systemtap when I ran mount -o _netdev /dev/vdb1 /opt, where Before=local-fs.target is still temporarily configured for opt.mount, meaning that wrong dependency is being configured temporarily. This is not good behavior.

```
Thu Jun  2 16:18:34 2022 mount (3431) kprobe.function("__x64_sys_mount")? "/dev/vdb1", "/opt", "xfs", 0x0, 0x0
...snip...
Thu Jun  2 16:18:34 2022 systemd (1) process("/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so").function("unit_add_dependency_by_name@../src/core/unit.c:3196") "opt.mount" "local-fs.target" dep:14 mask:64 options:"rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota" 0x7fdcdcabedc0 : unit_add_dependency_by_name+0x0/0x130 [/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so]
...snip...
Thu Jun  2 16:18:34 2022 systemd (1) process("/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so").function("unit_add_dependency_by_name@../src/core/unit.c:3196") "opt.mount" "local-fs.target" dep:14 mask:64 options:"rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota" 0x7fdcdcabedc0 : unit_add_dependency_by_name+0x0/0x130 [/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so]
...snip...
Thu Jun  2 16:18:34 2022 mount (3431) kprobe.function("__x64_sys_write") "SRC=/dev/vdb1 TARGET=/opt ROOT=/ OPTS=_netdev\n"
Thu Jun  2 16:18:44 2022 mount (3431) kprobe.function("__x64_sys_rename")? "/run/mount/utab.NHrdAb", "/run/mount/utab" t" "remote-fs.target" dep:14 mask:64 options:"rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota,_netdev" 0x7fdcdcabedc0 : unit_add_dependency_by_name+0x0/0x130 [/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so]
...snip...
Thu Jun  2 16:18:44 2022 systemd (1) process("/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so").function("unit_add_dependency_by_name@../src/core/unit.c:3196") "opt.mount" "remote-fs-pre.target" dep:15 mask:64 options:"rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota,_netdev" 0x7fdcdcabedc0 : unit_add_dependency_by_name+0x0/0x130 [/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so]
Thu Jun  2 16:18:44 2022 systemd (1) process("/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so").function("unit_add_dependency_by_name@../src/core/unit.c:3196") "opt.mount" "network.target" dep:15 mask:64 options:"rw,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota,_netdev" 0x7fdcdcabedc0 : unit_add_dependency_by_name+0x0/0x130 [/usr/lib/systemd/libsystemd-core-251.1-2.fc36.test.so]
```

It seems to me that implicit dependencies for some kind of mount options like _netdev are undecidable until /run/mount/utab file is updated.

src/core/unit.h
yuwata: Strictly speaking, the dependencies tagged with this new label _always_ depend on /proc/self/mountinfo in some sense. Of course, if there exists a relevant entry in /proc/self/mountinfo, then the dependencies depend on it. Even if there is _no_ corresponding entry in /proc/self/mountinfo, the dependencies _are_ based on the fact that there is no entry in /proc/self/mountinfo. Hence, when /proc/self/mountinfo is updated, we need to regenerate the dependencies, and they must not be labeled with `UNIT_DEPENDENCY_FILE`.

Though, the name `UNIT_DEPENDENCY_MOUNTINFO_OR_FILE` may be misleading. But I have no idea about better name. Do you have any suggestions?

yuwata:
> > Does it? I did not expect that. Interesting.
> 
> Yes it seems so.
> 
> Could you check it and update the commit messages to reflect this fact after checking that it's true (there's a test case provided in issue #19983) ?
> 
> Also in this case, could you also provide test cases to show which problems the 2 last commits are supposed to fix ?
> 
> Thanks.

The commit `core/mount: set Mount.from_proc_self_mountinfo flag before adding default dependencies` partially fixes #19983. But the test added at the last should require the whole commits in this PR.

src/core/mount.c
yuwata: The mask is now dropped, and the three labels are explicitly specified in `unit_remove_dependencies()`.

yuwata:
@fbuihuu and @keszybz Sorry for late response. I rebased the PR, dropped the mask, and added a test case. See the above comments for more details. PTAL.

d-hatayama:
@yuwata I tried to test this new update  of the PR for issue#23552 but the issue occurs. It looks that the old dependencies such as Before=local-fs.target and After=local-fs-pre.target are not removed.
```
# LANG=C systemctl show -p Before,After opt.mount
Before=umount.target remote-fs.target local-fs.target
After=systemd-journald.socket local-fs-pre.target network-online.target blockdev@dev-vda2.target system.slice -.mount network.target remote-fs-pre.target dev-vda2.device
```
In the previous version, it worked well. What is the difference from the previous version?

fbuihuu:
@yuwata sorry I didn't have time to take a look before my vacation. I'll do it when I'll return in 3 weeks if this can wait until then.

One thing I discovered recently is that commit c52c2dc64f2443dd2e1f0cc82f577b68520ceb8f seems to have fixed issue #19983, at least I couldn't reproduce the provided test case.

d-hatayama:
> @yuwata I tried to test this new update of the PR for issue#23552 but the issue occurs. It looks that the old dependencies such as Before=local-fs.target and After=local-fs-pre.target are not removed.
> 
> ```
> # LANG=C systemctl show -p Before,After opt.mount
> Before=umount.target remote-fs.target local-fs.target
> After=systemd-journald.socket local-fs-pre.target network-online.target blockdev@dev-vda2.target system.slice -.mount network.target remote-fs-pre.target dev-vda2.device
> ```
> 
> In the previous version, it worked well. What is the difference from the previous version?

I got why.

Before=local-fs.target is configured with UNIT_DEPENDENCY_FILE:

```                                                                                                                                               
(gdb) bt                                                                                                                                          
#0  unit_add_dependency_by_name (u=0x1561460, d=UNIT_BEFORE, name=0x7f64c5e099b8 "local-fs.target", add_reference=true, mask=UNIT_DEPENDENCY_FILE) at ../src/core/unit.c:3203                                                                                                                      
#1  0x00007f64c5d3858c in mount_add_default_ordering_dependencies (m=0x1561460, p=0x15618a8, mask=UNIT_DEPENDENCY_FILE) at ../src/core/mount.c:50\
0                                                                                                                                                 
#2  0x00007f64c5d3874a in mount_add_default_dependencies (m=0x1561460) at ../src/core/mount.c:538                                                 
#3  0x00007f64c5d393d5 in mount_add_non_exec_dependencies (m=0x1561460) at ../src/core/mount.c:630                                                
#4  0x00007f64c5d39564 in mount_add_extras (m=0x1561460) at ../src/core/mount.c:676                                                               
#5  0x00007f64c5d397b6 in mount_load (u=0x1561460) at ../src/core/mount.c:718                                                                     
#6  0x00007f64c5db792e in unit_load (u=0x1561460) at ../src/core/unit.c:1637                                                                      
#7  0x00007f64c5d2c251 in manager_dispatch_load_queue (m=0x12903c0) at ../src/core/manager.c:2080                                                 
#8  0x00007f64c5d3fbf1 in mount_process_proc_self_mountinfo (m=0x12903c0) at ../src/core/mount.c:1979                                             
#9  0x00007f64c5d3ff5b in mount_dispatch_io (source=0x1297010, fd=15, revents=1, userdata=0x12903c0) at ../src/core/mount.c:2073                  
#10 0x00007f64c594cb56 in source_dispatch (s=0x1297010) at ../src/libsystemd/sd-event/sd-event.c:3602                                             
#11 0x00007f64c594ec54 in sd_event_dispatch (e=0x11a4860) at ../src/libsystemd/sd-event/sd-event.c:4186                                           
#12 0x00007f64c594f137 in sd_event_run (e=0x11a4860, timeout=18446744073709551615) at ../src/libsystemd/sd-event/sd-event.c:4247                  
#13 0x00007f64c5d301c9 in manager_loop (m=0x12903c0) at ../src/core/manager.c:3073                                                                
#14 0x00000000004104e4 in invoke_main_loop (m=0x12903c0, saved_rlimit_nofile=0x7ffe2f2f3470, saved_rlimit_memlock=0x7ffe2f2f3480, ret_retval=0x7ffe2f2f3398, ret_shutdown_verb=0x7ffe2f2f33e8, ret_fds=0x7ffe2f2f3400, ret_switch_root_dir=0x7ffe2f2f33d8, ret_switch_root_init=0x7ffe2f2f33e0, ret_error_message=0x7ffe2f2f33f0) at ../src/core/main.c:1912                                                                                        
#15 0x00000000004142b0 in main (argc=5, argv=0x7ffe2f2f36d8) at ../src/core/main.c:2979                                                           
```

which is not removed by unit_remove_dependencies() at the entry of mount_add_non_exec_dependencies().

```c
static int mount_add_non_exec_dependencies(Mount *m) {                                                                                            
        int r;                                                                                                                                    
                                                                                                                                                  
        assert(m);                                                                                                                                
                                                                                                                                                  
        /* Any dependencies based on /proc/self/mountinfo may be now stale. */                                                                    
        unit_remove_dependencies(UNIT(m),                                                                                                         
                                 UNIT_DEPENDENCY_MOUNTINFO_IMPLICIT |                                                                             
                                 UNIT_DEPENDENCY_MOUNTINFO_DEFAULT |                                                                              
                                 UNIT_DEPENDENCY_MOUNTINFO_OR_FILE);
```

I think it correct to use UNIT_DEPENDENCY_MOUNTINFO_OR_FILE also in mount_add_default_dependencies().

```
@@ -533,9 +532,7 @@ static int mount_add_default_dependencies(Mount *m) {                                                                         
         if (!p)                                                                                                                                  
                 return 0;                                                                                                                        
                                                                                                                                                  
-        mask = m->from_fragment ? UNIT_DEPENDENCY_FILE : UNIT_DEPENDENCY_MOUNTINFO_DEFAULT;                                                      
-                                                                                                                                                 
-        r = mount_add_default_ordering_dependencies(m, p, mask);                                                                                 
+        r = mount_add_default_ordering_dependencies(m, p, UNIT_DEPENDENCY_MOUNTINFO_OR_FILE);                                                    
         if (r < 0)                                                                                                                               
                 return r;
```

yuwata:
@d-hatayama Wow, thank you! The suggested change is applied. Also, the last commit tests `_netdev` mount option. PTAL!

src/core/mount.c
```diff
@@ -2079,25 +2076,17 @@ static int mount_process_proc_self_mountinfo(Manager *m) {
...
+                        /* Exclude devices currently used */
+                        free(set_remove(gone, mount->parameters_proc_self_mountinfo.what));
 
                 /* Reset the flags for later calls */
                 mount->proc_flags = 0;
         }
 
-        SET_FOREACH(what, gone) {
-                if (set_contains(around, what))
-                        continue;
```
poettering: this doesn't look right.

If a specific device is mounted to multiple places, and one of them is removed, then we should not mistake this for it being gone entirely...

That's why the old code maintained two lists: of the devices definitely around, and the ones that appear gone. But if one of the gone devices actually appears in the around list, we'd know it's not actually gone...

src/core/mount.c
```diff
@@ -646,9 +646,6 @@ static int mount_add_extras(Mount *m) {
 
         if (!m->where) {
                 r = unit_name_to_path(u->id, &m->where);
-                if (r == -ENAMETOOLONG)
-                        log_unit_error_errno(u, r, "Failed to derive mount point path from unit name, because unit name is hashed. "
-                                                   "Set \"Where=\" in the unit file explicitly.");
```
poettering: hmm? unit_name_to_path is defined like this:

```c
int unit_name_to_path(const char *name, char **ret) {
        _cleanup_free_ char *prefix = NULL;
        int r;

        assert(name);

        r = unit_name_to_prefix(name, &prefix);
        if (r < 0)
                return r;

        if (unit_name_is_hashed(name))
                return -ENAMETOOLONG;

        return unit_name_path_unescape(prefix, ret);
}
```

so it definitely can return ENAMETOOLONG?

poettering (changes requested):

src/core/mount.c
poettering: I guess the confusion here is between unit_name_to_path() and unit_name_from_path(). Only the latter stopped generating ENAMETOOLONG...

yuwata:
@poettering Thank you for the review. The above points are fixed. PTAL.

poettering:
lgtm

keszybz:
Arrgh, this has lots of conflicts with v251-stable. Not sure what to do here.

keszybz:
This seems risky. So unless somebody really asks for this, or even better provides a tested pull request, I'll punt.

cdhowie:
@YHNdnzj I just ran into #1741 this week on a Debian Bookworm system. Which version of systemd should be fixed? (I cannot comment on #1741 because it is locked.)

YHNdnzj:
> @YHNdnzj I just ran into #1741 this week on a Debian Bookworm system. Which version of systemd should be fixed? (I cannot comment on #1741 because it is locked.)

Hmm, it should be fixed in v252, i.e. what Debian 12 is using. If not, please provide debug logs. I've unlocked the original issue.

PR #23368, opened 2022-05-13
boot: Build with -flto=auto instead

medhefgo:


medhefgo:
Do we care about gcc-9? According to docs, this mode was introduced with 10.

bluca:
> Do we care about gcc-9? According to docs, this mode was introduced with 10.

This would cut off Debian oldoldstable and Ubuntu 18.04, which is not a problem from my point of view. Might want to check with the CentOS/RHEL folks though. Also should be documented in README.

medhefgo:
This should only be an issue for people building with `-Db_lto=true` if they pass lto by `CFLAGS`, things stay the same. So probably not an issue?

bluca (approved):

PR #23372, opened 2022-05-14
hwdb: Add Google Hangouts Meet speakermic

pceballosgoogle:
This device implements the phone mute HID usage as a toggle switch,
where 1 indicates muted, and 0 indicates unmuted. However, for a key
event 1 indicates that the key has been pressed and 0 indicates it has
been released. This mismatch causes issues, so prevent key events from
being generated for this HID usage.


yuwata (changes requested):
Please move the entry to 60-keyboard.hwdb with a short comment.

pceballosgoogle:
Thanks. I've addressed the comments. Please review.

bluca (approved):

PR #23383, opened 2022-05-15
meson: suppress warning

yuwata:


meson.build
```diff
@@ -1005,7 +1013,7 @@ libbpf = dependency('libbpf', required : bpf_framework_required, version : '>= 0
 conf.set10('HAVE_LIBBPF', libbpf.found())
 
 if want_bpf_framework == 'false' or not libbpf.found()
-        conf.set10('BPF_FRAMEWORK', 0)
+        conf.set10('BPF_FRAMEWORK', false)
```
evverx, line 1007: Looks like the CI didn't catch it because the bpf framework ignores `skip-deps`. I think maybe `skip-deps` should be added to the condition by analogy with all the other dependencies.

Regarding `version_compare('>= 0.55')` as usual I think `meson` should be bumped :-) I'm not sure why upstream projects should cater to slowly moving distributions trying to build the latest systemd with outdated toolchains.

yuwata: > Looks like the CI didn't catch it because the bpf framework ignores `skip-deps`. I think maybe `skip-deps` should be added to the condition by analogy with all the other dependencies.

Added one more commit for that.

> Regarding `version_compare('>= 0.55')` as usual I think `meson` should be bumped :-) I'm not sure why upstream projects should cater to slowly moving distributions trying to build the latest systemd with outdated toolchains.

Yeah, I agree, but let's reconsider that after v251.

evverx: > Added one more commit for that.

Thanks! I think it should help to set `BPF_FRAMEWORK` to `false`. `HAVE_LIBBPF` can still be true as far as I understand and some code relying on libbpf can still be compiled:
```sh
$ git grep HAVE_LIBB
meson.build:conf.set10('HAVE_LIBBPF', libbpf.found())
src/analyze/analyze-filesystems.c:#if ! HAVE_LIBBPF
src/core/dbus-execute.c:#if HAVE_LIBBPF
src/core/execute.c:#if HAVE_LIBBPF
src/core/execute.c:#if HAVE_LIBBPF
src/core/execute.c:#if HAVE_LIBBPF
src/core/execute.c:#if HAVE_LIBBPF
src/core/main.c:#if HAVE_LIBBPF
src/core/manager.c:#if HAVE_LIBBPF
src/shared/bpf-dlopen.c:#if HAVE_LIBBPF
src/shared/bpf-dlopen.h:#if HAVE_LIBBPF
src/shared/meson.build:if conf.get('HAVE_LIBBPF') == 1
src/test/test-dlopen-so.c:#if HAVE_LIBBPF
```

I think technically it has nothing to do with the warning and the CI so it can be fixed later. Thanks!

meson.build
```diff
@@ -641,7 +641,11 @@ foreach prog : progs
                                    '/usr/sbin/' + prog[0],
                                    '/sbin/' + prog[0],
                                    required: false)
-                path = exe.found() ? exe.path() : prog[1]
+                if meson.version().version_compare('>= 0.55')
+                        path = exe.found() ? exe.full_path() : prog[1]
+                else
+                        path = exe.found() ? exe.path() : prog[1]
+                endif
```
eli-schwartz: You don't really need to do this at all. Meson will not issue a warning (yellow "WARNING" text, potentially triggers --fatal-meson-warnings) for the use of `.path()` until and unless you bump the minimum version of Meson to 0.55.0

The difference between the two is nonexistent. We renamed it for consistency between external programs and built programs, and deprecated the older name in order to encourage consistency.

You may do if/else, or not, as you please. Some people don't care, and just change things when they bump the minimum. Other people like to add the if/else because they find it makes it easier to bump the minimum -- simply bump it, and no need to upgrade deprecated code at the same time, and it's easier to drop the else statement than it is to spend even 30 seconds figuring out what the correct way to upgrade is.

However...

> Regarding version_compare('>= 0.55') as usual I think meson should be bumped :-) I'm not sure why upstream projects should cater to slowly moving distributions trying to build the latest systemd with outdated toolchains.

There's no reason to bump the minimum version of Meson for such a trifling change.

Just stick with `.path()` until you are ready to bump the minimum version of Meson, then change things. It causes zero harm whatsoever. This is infinitely better than locking out users of 0.54 (such as Ubuntu 20.04) for... literally no improvement.

meson.build
```diff
@@ -669,8 +673,12 @@ gperf_test_format = '''
...
 @1@
 '''
-gperf_snippet_format = 'echo foo,bar | @0@ -L ANSI-C'
-gperf_snippet = run_command(sh, '-c', gperf_snippet_format.format(gperf.path()),
+if meson.version().version_compare('>= 0.55')
+        gperf_path = gperf.full_path()
+else
+        gperf_path = gperf.path()
+endif
+gperf_snippet = run_command(sh, '-c', 'echo foo,bar | @0@ -L ANSI-C'.format(gperf_path),
```
eli-schwartz: This actually points out a more general issue, which is that both before and after this PR, this code snippet is wrong. See #23388

eli-schwartz (changes requested):

meson.build
evverx: > There's no reason to bump the minimum version of Meson for such a trifling change.

Agreed. Thanks! I was thinking more about `install_tag` when I was writing that.

eli-schwartz: install_tag is different, yeah.

Specifically, because I'm the current state of affairs it prevents a warning-free build (that is, it's *impossible* no matter what options or configurations you use, to suppress the warning).

yuwata: @eli-schwartz Thank you for the suggestion. The change is dropped.

keszybz:
LGTM.

bluca (approved):

PR #23388, opened 2022-05-15
meson: use better shellscript argument passing

eli-schwartz:
Passing potentially arbitrary data into a shellscript is potentially very broken if you do not correctly quote it for use. This quoting must be done as part of the interpretation of the data itself, e.g. python's shlex.quote; simply formatting it into a string with double quotes is NOT sufficient.

An alternative is to communicate the data reliably via argv to the shell process, and allow the shell to internally handle it via `"$1"`, which is quote-safe and will expand the data from argv as a single tokenized word.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -669,8 +669,7 @@ gperf_test_format = '''
 const char * in_word_set(const char *, @0@);
 @1@
 '''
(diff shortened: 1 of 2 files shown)


yuwata (approved):
Thanks. LGTM.

PR #23396, opened 2022-05-16
logind: remember our idle state and use it to detect idle level transitions

### LINKED ISSUE #16391: systemd: logind parameters "IdleAction=Lock" is not working
> Version: 241-7

**Used distribution**
 >Debian 10.4 Desktop

> In logind.conf `IdleAction=Lock` should send `Lock` event when system is Idle

 >Not received any `Lock` event After the above configuration and restart of system, 

* Disable other Auto Lock features in desktop environment.
* Set 'Blank screen' timeout to 5min
* Configured the logind for recieving the `Lock` event action after reaching system Idle.
logind.conf
```
[Login]
IdleAction=lock
IdleActionSec=1min
```
* Restart the system and waited using a sample application which is registered for dbus event "Lock"
* waited for more than 10min after system lid is off, not received any "Lock" event, and when i saw the status logs the idle action was initiating but no dbus event generated
```
========================================================================================
● systemd-logind.service - Login Service
   Loaded: loaded (/lib/systemd/system/systemd-logind.service; static; vendor
preset: enabled)
   Active: active (running) since Tue 2020-06-16 19:48:42 IST; 2 days ago
     Docs: man:systemd-logind.service(8)
           man:logind.conf(5)
           https://www.freedesktop.org/wiki/Software/systemd/logind
           https://www.freedesktop.org/wiki/Software/systemd/multiseat
 Main PID: 634 (systemd-logind)
   Status: "Processing requests..."
    Tasks: 1 (limit: 4915)
   Memory: 2.6M
   CGroup: /system.slice/systemd-logind.service
           └─634 /lib/systemd/systemd-logind

Jun 19 18:09:08 debian systemd-logind[634]: System idle. Taking action.
Jun 19 18:10:38 debian systemd-logind[634]: System idle. Taking action.
Jun 19 18:12:08 debian systemd-logind[634]: System idle. Taking action.
Jun 19 18:13:38 debian systemd-logind[634]: System idle. Taking action.
Jun 19 18:15:08 debian systemd-logind[634]: System idle. Taking action.
========================================================================================
```
With curiosity i have checked the source code and after applying the below change i could see the `Lock` event is generating and  my sample application could recieve it.

```diff
diff --git a/src/login/logind.c b/src/login/logind.c
index 95ec0a57c6..ea505ceed8 100644
--- a/src/login/logind.c
+++ b/src/login/logind.c
@@ -1019,7 +1019,7 @@ static int manager_dispatch_idle_action(sd_event_source
*s, uint64_t t, void *us
                     (m->idle_action_not_before_usec <= 0 || n >=
m->idle_action_not_before_usec + m->idle_action_usec)) {
                         log_info("System idle. Taking action.");

-                        manager_handle_action(m, 0, m->idle_action, false,
false);
+                        manager_handle_action(m, 0, m->idle_action, false,
true);
                         m->idle_action_not_before_usec = n;
                 }
```
Is there any specific reason blocking the "Lock" action in the [source code](https://github.com/systemd/systemd/blob/d7b34e384108ff431401d942ec0daaf156ec27c1/src/login/logind-action.c#L94)?

can we apply the above change to the master branch?

Thanks,
venkata.

msekletar:
Fixes #16391

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -1736,6 +1736,13 @@ install_basic_tools() {
     image_install -o ldconfig.real
     # for TEST-35-LOGIN
     image_install -o evemu-device evemu-event
+    if command -v expect >/dev/null && command -v tclsh >/dev/null ; then
+        # shellcheck disable=SC2016
+        version="$(tclsh <<< 'puts $tcl_version')"
+
+        image_install expect
+        inst_recursive /usr/lib64/tcl"$version" /usr/share/tcl"$version"
+    fi
 }
 
 install_debug_tools() {
--- a/test/units/testsuite-35.sh
+++ b/test/units/testsuite-35.sh
@@ -360,6 +360,89 @@ EOF
     fi
 }
 
+setup_idle_action_lock() {
+    useradd testuser ||:
+
+    mkdir -p /run/systemd/logind.conf.d/
+    cat >/run/systemd/logind.conf.d/idle-action-lock.conf <<EOF
+[Login]
+IdleAction=lock
+IdleActionSec=1s
+EOF
+    systemctl restart systemd-logind.service
+    systemctl service-log-level systemd-logind.service debug
+}
+
+teardown_idle_action_lock() {(
+    set +ex
+    rm -f /run/systemd/logind.conf.d/idle-action-lock.conf
+    pkill -9 -u "$(id -u testuser)"
+    userdel -r testuser
+    systemctl restart systemd-logind.service
+)}
+
+test_lock_idle_action() {
+    if ! command -v expect >/dev/null ; then
+        echo >&2 "expect not installed, skiping test ${FUNCNAME[0]}"
+        return 0
+    fi
+
+    setup_idle_action_lock
+    trap teardown_idle_action_lock RETURN
+
+    if loginctl --no-legend | awk '{ print $3; }' | sort -u | grep -q testuser ; then
+        echo >&2 "Session of the \'testuser\' is already present."
+        return 1
+    fi
+
+    # IdleActionSec is set 1s but the accuracy of associated timer is 30s so we
+    # need to sleep in worst case for 31s to make sure timer elapsed. We sleep
+    # here for 35s to accomodate for any possible scheudling delays.
+    cat > /tmp/test.exp <<EOF
+spawn systemd-run -G -t -p PAMName=login -p User=testuser bash
+send "sleep 35\r"
+send "echo foobar\r"
+send "sleep 35\r"
+send "exit\r"
+interact
+wait
+EOF
+
+    ts="$(date '+%H:%M:%S')"
+    busctl --match "type='signal',sender='org.freedesktop.login1',interface='org.freedesktop.login1.Session',member='Lock'" monitor  > dbus.log &
+
+    expect /tmp/test.exp &
+
+    # Sleep a bit to give expect time to spawn systemd-run before we check for
+    # the presence of resulting session.
+    sleep 1
+    if [ "$(loginctl --no-legend | awk '{ print $3; }' | sort -u | grep -c testuser)" != 1 ] ; then
+        echo >&2 "\'testuser\' is expected to have exactly one session running."
+        return 1
+    fi
+
+    wait %2
+    kill %1
+
+    # We slept for 35s , in that interval all sessions should have become idle
+    # and "Lock" signal should have been sent out. Then we wrote to tty to make
+    # session active again and next we slept for another 35s so sessions have
+    # become idle again. 'Lock' signal is sent out for each session, we have at
+    # least one session, so minimum of 2 "Lock" signals must have been sent.
+    if [ "$(grep -c Member=Lock dbus.log)" -lt 2 ]; then
+        echo >&2 "Too few 'Lock' D-Bus signal sent, expected at least 2."
+        return 1
+    fi
+
+    journalctl -b -u systemd-logind.service --since="$ts" > logind.log
+    if [ "$(grep -c 'System idle. Will be locked now.' logind.log)" -lt 2 ]; then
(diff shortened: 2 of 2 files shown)


src/login/logind.c
```diff
@@ -963,19 +963,35 @@ static int manager_dispatch_idle_action(sd_event_source *s, uint64_t t, void *us
...
+
+                        if (m->idle_action == HANDLE_LOCK && !is_edge)
+                                /* We are idle and we were before so we are actually not taking any action. */
+                                log_debug("System idle.");
+                        else
+                                log_info("System idle. Doing %s operation.", handle_action_to_string(m->idle_action));
+
+                        manager_handle_action(m, 0, m->idle_action, false, is_edge);
                         m->idle_action_not_before_usec = n;
+
```
keszybz: Please drop this empty line.

src/login/logind.c
```diff
@@ -963,19 +963,35 @@ static int manager_dispatch_idle_action(sd_event_source *s, uint64_t t, void *us
...
 
                 if (n >= since.monotonic + m->idle_action_usec &&
                     (m->idle_action_not_before_usec <= 0 || n >= m->idle_action_not_before_usec + m->idle_action_usec)) {
-                        log_info("System idle. Doing %s operation.", handle_action_to_string(m->idle_action));
+                        bool is_edge = false;
 
-                        manager_handle_action(m, 0, m->idle_action, false, false);
+                        /* We weren't idle previously or some activity happened while we were sleeping, and now we are
+                         * idle. Let's remember that for the next time and make this an edge transition. */
+                        if (!m->was_idle || since.monotonic > m->idle_action_not_before_usec) {
```
keszybz: `>` → `>=`. We use `>=` everywhere.

keszybz (commented):
This looks reasonable, but it's hard to have confidence. I was wondering if there are some other place where `was_idle` should be reset.

It'd be nice to get confirmation that this fixes the OP's issue.

msekletar:
Updated. @keszybz PTAL.

Btw, regarding testing, so far I tested this manually. In my other PR that I have now opened (#23569) I am adding new integration test for logind so I will add test case for this once we merge both PR.

yuwata:
Please add a test.

keszybz:
Yeah. With some form of a test this would be much easier.

test/units/testsuite-35.sh
```diff
@@ -360,6 +360,89 @@ EOF
...
+    if loginctl --no-legend | awk '{ print $3; }' | sort -u | grep -q testuser ; then
+        echo >&2 "Session of the \'testuser\' is already present."
+        return 1
+    fi
+
+    # IdleActionSec is set 1s but the accuracy of associated timer is 30s so we
+    # need to sleep in worst case for 31s to make sure timer elapsed. We sleep
+    # here for 35s to accomodate for any possible scheudling delays.
+    cat > /tmp/test.exp <<EOF
+spawn systemd-run -G -t -p PAMName=login -p User=testuser bash
```
yuwata, line 402: Not sure, but why not to simply run a script with `systemd-run`??

test/units/testsuite-35.sh
```diff
@@ -360,6 +360,89 @@ EOF
...
+[Login]
+IdleAction=lock
+IdleActionSec=1s
+EOF
+    systemctl restart systemd-logind.service
+    systemctl service-log-level systemd-logind.service debug
+}
+
+teardown_idle_action_lock() {
+    set +e
```
yuwata: Please use subshell:
```
teardown_idle_action_lock() {(
    set +ex
    ...
)}
```

test/TEST-35-LOGIN/test.sh
```diff
@@ -7,4 +7,14 @@ TEST_DESCRIPTION="LOGIN"
 # shellcheck source=test/test-functions
 . "${TEST_BASE_DIR:?}/test-functions"
 
+test_append_files() {
+    if command -v expect >/dev/null && command -v tclsh >/dev/null ; then
```
yuwata: Please move this in `install_basic_tools()` in test-functions. The testing image is shared with other tests, hence these will not be included when multiple tests are executed, like our CIs do.

yuwata (changes requested):

test/units/testsuite-35.sh
msekletar: I am using expect because I want tty allocated and AFAICT you can't do systemd-run --pty from context where you don't have any tty.

test/test-functions
```diff
@@ -1736,6 +1736,13 @@ install_basic_tools() {
     image_install -o ldconfig.real
     # for TEST-35-LOGIN
     image_install -o evemu-device evemu-event
+    if command -v expect >/dev/null && command -v tclsh >/dev/null ; then
+        # shellcheck disable=SC2016
+        version="$(tclsh <<< 'puts $tcl_version')"
+
+        image_install expect
+        inst_recursive /usr/lib64/tcl"$version" /usr/share/tcl"$version"
```
bluca, line 1744: I missed this, right now the dependencies happen not to be installed in the ubuntu/debian tests, but if they were (intentionally or pulled in accidentally by something else), this would break, as the path is specific to 64bit rpm distros

bluca: on debian/ubuntu, `/usr/bin/tclsh` is a symlink to `/usr/bin/tclsh<version>`

bluca: there's also both `/usr/lib/tcl<version>` and multi-arch `/usr/lib/x86_64-linux-gnu/tcl<version>`, but no `/usr/share/tcl<version>`

yuwata: Ouch. Will address them.

yuwata: #23930.

PR #19166, opened 2021-03-31
coredump: compress on the fly

bluca:
coredump: check cgroups memory limit if storing on tmpfs
    
When /var/lib/systemd/coredump/ is backed by a tmpfs, all disk usage
will be accounted under the systemd-coredump process cgroup memory
limit.
If MemoryMax is set, this might cause systemd-coredump to be terminated
by the kernel oom handler when writing large uncompressed core files,
even if the compressed core would fit within the limits.

Detect if a tmpfs is used, and if so check MemoryMax from the process
and slice cgroups, and do not write uncompressed core files that are
greater than half the available memory. If the limit is breached,
stop writing and compress the written chunk immediately, then delete
the uncompressed chunk to free more memory, and resume compressing
directly from STDIN.

Example debug log when this situation happens:

```
systemd-coredump[737455]: Setting max_size to limit writes to 51344896 bytes.
systemd-coredump[737455]: ZSTD compression finished (51344896 -> 3260 bytes, 0.0%)
systemd-coredump[737455]: ZSTD compression finished (1022786048 -> 47245 bytes, 0.0%)
systemd-coredump[737455]: Process 737445 (a.out) of user 1000 dumped core.
```

Fixes https://github.com/systemd/systemd/issues/13968

PR #23398, opened 2022-05-16
coredump: directly compress into output file

MarkSchmitt:
Directly compress coredumps from input stream to compressed
output file instead of first (sometimes partially) writing
to uncompressed file and then compressing it. This safes
I/O and lowers ram usage in case tmpfs is used.

Notes:
I did zoom through the issues, but couldn't find one matching what I'm trying so solve.

We had some trouble storing large coredumps on tmpfs. This change directly compresses the coredump into the compressed file, instead of first writing it uncompressed. This saves I/O and also should make it possible to store very large core dumps, larger than the available memory (in tmpfs) when assuming the compression ratio is high.

On non-tmpfs, it saves a lot of potentially slow I/O.

I ran tests locally, and also tested a backported version of this change with v246 and v250 on Dunfell and Kirkstone Yocto images.

I'll be happy to work on findings or suggestions. This is my first systemd PR, please be kind! I hope I've adhered to all rules, I certainly tried!

Best regards,
Mark

Mandatory legal disclaimer from my company:

The program was tested solely for our own use cases, which might differ from yours.

<sub>Mark Schmitt [mark.schmitt@mercedes-benz.com](mailto:mark.schmitt@mercedes-benz.com), Mercedes-Benz Tech Innovation GmbH, [imprint](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -395,14 +395,9 @@ static int save_external_coredump(
         /* If storage is on tmpfs, the kernel oomd might kill us if there's MemoryMax set on
          * the service or the slice it belongs to. This is common on low-resources systems,
          * to avoid crashing processes to take away too many system resources.
-         * Check the cgroup settings, and set max_size to a bit less than half of the
-         * available memory left to the process.
-         * Then, attempt to write the core file uncompressed first - if the write gets
-         * interrupted, we know we won't be able to write it all, so instead compress what
-         * was written so far, delete the uncompressed truncated core, and then continue
-         * compressing from STDIN. Given the compressed core cannot be larger than the
-         * uncompressed one, and 1KB for metadata is accounted for in the calculation, we
-         * should be able to at least store the full compressed core file. */
+         * Check the cgroup settings and set max_size to at most the cgroup limit minus
+         * 1KB for compression meta data.
+         */
 
         storage_on_tmpfs = fd_is_temporary_fs(fd) > 0;
         if (storage_on_tmpfs && arg_compress) {
@@ -434,7 +429,7 @@ static int save_external_coredump(
                 }
 
                 max_size = MIN(cgroup_limit, max_size);
-                max_size = LESS_BY(max_size, 1024U) / 2; /* Account for 1KB metadata overhead for compressing */
+                max_size = LESS_BY(max_size, 1024U); /* Account for 1KB metadata overhead for compressing */
                 max_size = MAX(PROCESS_SIZE_MIN, max_size); /* Impose a lower minimum */
 
                 /* tmpfs might get full quickly, so check the available space too.
@@ -445,12 +440,6 @@ static int save_external_coredump(
 
                 log_debug("Limiting core file size to %" PRIu64 " bytes due to cgroup memory limits.", max_size);
         }
-
-        r = copy_bytes(input_fd, fd, max_size, 0);
-        if (r < 0)
-                return log_error_errno(r, "Cannot store coredump of %s (%s): %m",
-                                context->meta[META_ARGV_PID], context->meta[META_COMM]);
-        truncated = r == 1;
 
 #if HAVE_COMPRESSION
         if (arg_compress) {
@@ -470,24 +459,9 @@ static int save_external_coredump(
                 if (fd_compressed < 0)
                         return log_error_errno(fd_compressed, "Failed to create temporary file for coredump %s: %m", fn_compressed);
 
-                r = compress_stream(fd, fd_compressed, max_size, &uncompressed_size);
+                r = compress_stream(input_fd, fd_compressed, max_size, &uncompressed_size);
                 if (r < 0)
                         return log_error_errno(r, "Failed to compress %s: %m", coredump_tmpfile_name(tmp_compressed));
-
-                if (truncated && storage_on_tmpfs) {
-                        uint64_t partial_uncompressed_size = 0;
-
-                        /* Uncompressed write was truncated and we are writing to tmpfs: delete
-                         * the uncompressed core, and compress the remaining part from STDIN. */
-
-                        tmp = unlink_and_free(tmp);
-                        fd = safe_close(fd);
-
-                        r = compress_stream(input_fd, fd_compressed, max_size, &partial_uncompressed_size);
-                        if (r < 0)
-                                return log_error_errno(r, "Failed to compress %s: %m", coredump_tmpfile_name(tmp_compressed));
-                        uncompressed_size += partial_uncompressed_size;
-                }
 
                 r = fix_permissions(fd_compressed, tmp_compressed, fn_compressed, context, uid);
                 if (r < 0)
@@ -509,6 +483,12 @@ static int save_external_coredump(
                 return 0;
         }
 #endif
+
+        r = copy_bytes(input_fd, fd, max_size, 0);
+        if (r < 0)
+                return log_error_errno(r, "Cannot store coredump of %s (%s): %m",
+                                context->meta[META_ARGV_PID], context->meta[META_COMM]);
+        truncated = r == 1;
 
         if (truncated)
                 log_struct(LOG_INFO,


bluca:
This was already proposed and rejected, please see the discussion at: https://github.com/systemd/systemd/pull/19166#discussion_r604742481

MarkSchmitt:
> This was already proposed and rejected, please see the discussion at: [#19166 (comment)](https://github.com/systemd/systemd/pull/19166#discussion_r604742481)

Hi @bluca , I tried understanding the discussion ... I might have been a bit premature in removing the division by 2 in   https://github.com/systemd/systemd/pull/23398/files#diff-25e03646f09722b2e97b77bb7effb48a21a1f0d9171ef69e5770f0e29cf272e9R432 because we still need to allocate some memory during compression. In https://github.com/systemd/systemd/pull/23398/files#diff-25e03646f09722b2e97b77bb7effb48a21a1f0d9171ef69e5770f0e29cf272e9L486 it might happen, that if the compression is bad, you still go OOM _or_ if you do hit  `max_bytes` the function returns an error (in lz4 case, haven't checked the other two compression algos yet), which would not finish writing the coredump properly https://github.com/systemd/systemd/pull/23398/files#diff-25e03646f09722b2e97b77bb7effb48a21a1f0d9171ef69e5770f0e29cf272e9L488. That's something that imho can happen with the current implementation, and would also happen with my change (not getting worse though with my change!).

I think with these two modifications, compresse coredumps could be written directly and with the option to truncate them, if we run out of disk space or memory (tmpfs):
1.) (tmpfs only) we need an idea how much memory needs to be available after the coredump was written. this would also need to include the memory, that the compression algorithm allocates for buffers and stuff that happens directly after `save_external_coredump` exited.
2.) (also necessary for normal disks, not only tmpfs) we need compress_stream to exit gracefully and truncate the file when hitting `max_bytes`.

I'd be interested in some feedback. I think I could make the above proposed modifications. Please let me know if I have overlooked something.

bluca:
The problem is not the 'how', but what was rejected is going directly to compressed output in all cases without taking into account whether it's necessary or not. Are you saying the current implementation doesn't take into account the available space on the target filesystem? IIRC it should, but there might be a bug somewhere ofc

MarkSchmitt:
So first let's talk about potential issues with the current code:;

> The problem is not the 'how', but what was rejected is going directly to compressed output in all cases without taking into account whether it's necessary or not. Are you saying the current implementation doesn't take into account the available space on the target filesystem? IIRC it should, but there might be a bug somewhere ofc

Yes I think there's still some issues with the current implementation:
`max_size` is only half of the available cgroup limit, so in case compression is active and we're on tmpfs, we will write up to `max_size` of the coredump uncompressed, detect that it was truncated, then compress it, which should fit into the tmpfs / cgroup limit (as the compression ratio should be at least 1), then delete the uncompressed partial coredump, and then compress directly from input stream into the compressed_fd writing up to `max_size`. 

And here is where you get into trouble:
1.) if we would hit the max_size, the compress_stream function returns with an error. we don't get a truncated coredump, we have nothing. that could be improved, albeit it's not that easy with compression ... we would need to stop the algorithm gracefully or simply stop writing, but we would need to make sure, that it's still uncompressible afterwards .. in any case, that needs to be done in the compress_stream function or the way we handle its errors.
2.) we can still go OOM. it's very much a constructed scenario, only when the coredump cannot be compressed good enough, or limits are very low and the buffers allocated during compression eat up too much memory.

Now .. back to the business of directly compressing the input_fd into the output file without doing uncompressed writes...

@bluca youw rote:
>The problem is not the 'how', but what was rejected is going directly to compressed output in all cases without taking into account whether it's necessary or not

Quoting from: https://github.com/systemd/systemd/pull/19166#discussion_r604742481
>hmm, so as I understand you now compress first, and then uncompress again, i.e. you trade CPU usage against disk space use.

@bluca I assume you're fererring to the comment and the thread stated by @poettering.

But .. maybe I'm blind, but I don't see that happening anywhere. The the OP was referring to does not exist any more ( https://github.com/systemd/systemd/pull/19166/files/13c87159c0215cc2035a1db89de7c121cc5ce144..8b20f3289e235992c91cd6a1fde98824bee8f332#diff-25e03646f09722b2e97b77bb7effb48a21a1f0d9171ef69e5770f0e29cf272e9 ), but I assume there was some code somewhere compressing first and then uncompressing. But imho it's not happening now, neither in the main branch nor with the change I'm proposing. But if it does, please show me. :confused:

bluca:
> So first let's talk about potential issues with the current code:;
> 
> > The problem is not the 'how', but what was rejected is going directly to compressed output in all cases without taking into account whether it's necessary or not. Are you saying the current implementation doesn't take into account the available space on the target filesystem? IIRC it should, but there might be a bug somewhere ofc
> 
> Yes I think there's still some issues with the current implementation: `max_size` is only half of the available cgroup limit, so in case compression is active and we're on tmpfs, we will write up to `max_size` of the coredump uncompressed, detect that it was truncated, then compress it, which should fit into the tmpfs / cgroup limit (as the compression ratio should be at least 1), then delete the uncompressed partial coredump, and then compress directly from input stream into the compressed_fd writing up to `max_size`.
> 
> And here is where you get into trouble: 1.) if we would hit the max_size, the compress_stream function returns with an error. we don't get a truncated coredump, we have nothing. that could be improved, albeit it's not that easy with compression ... we would need to stop the algorithm gracefully or simply stop writing, but we would need to make sure, that it's still uncompressible afterwards .. in any case, that needs to be done in the compress_stream function or the way we handle its errors. 2.) we can still go OOM. it's very much a constructed scenario, only when the coredump cannot be compressed good enough, or limits are very low and the buffers allocated during compression eat up too much memory.

For the latter, uncompressible core files are out of scope, it's really uncommon (never seen it) and in that case directly compressing doesn't help either anyway, you still go OOM, so it's not a very interesting case.
For the former, can you provide a reproducer that shows that issue?

> Now .. back to the business of directly compressing the input_fd into the output file without doing uncompressed writes...
> 
> @bluca youw rote:
> 
> > The problem is not the 'how', but what was rejected is going directly to compressed output in all cases without taking into account whether it's necessary or not
> 
> Quoting from: [#19166 (comment)](https://github.com/systemd/systemd/pull/19166#discussion_r604742481)
> 
> > hmm, so as I understand you now compress first, and then uncompress again, i.e. you trade CPU usage against disk space use.
> 
> @bluca I assume you're referring to the comment and the thread stated by @poettering.
> 
> But .. maybe I'm blind, but I don't see that happening anywhere. The the OP was referring to does not exist any more ( https://github.com/systemd/systemd/pull/19166/files/13c87159c0215cc2035a1db89de7c121cc5ce144..8b20f3289e235992c91cd6a1fde98824bee8f332#diff-25e03646f09722b2e97b77bb7effb48a21a1f0d9171ef69e5770f0e29cf272e9 ), but I assume there was some code somewhere compressing first and then uncompressing. But imho it's not happening now, neither in the main branch nor with the change I'm proposing. But if it does, please show me. confused

That's how it must work, the uncompressed core must be made available too for analysis and to log various things to the journal. So if it's changed to compress directly, then de-compression must be re-added, which goes back to the objection. This is how that PR was first implemented. The current compromise is that this happens all the time, apart from the corner cases where the cgroup/fs limit kicks in. But for the vast majority of cases where there's no issue, the uncompressed core must be there too for analysis.

MarkSchmitt:
>That's how it must work, the uncompressed core must be made available too for analysis and to log various things to the journal. So if it's changed to compress directly, then de-compression must be re-added, which goes back to the objection. 

Oh, I see, I haven't understood that part before. Thanks for clarifying. I'll dig in a bit deeper, but I think it's save to close the PR.

>> .) if we would hit the max_size, the compress_stream function returns with an error. we don't get a truncated coredump, we have nothing. that could be improved ...

>For the latter, uncompressible core files are out of scope, it's really uncommon (never seen it) and in that case directly compressing doesn't help either anyway, you still go OOM, so it's not a very interesting case.
For the former, can you provide a reproducer that shows that issue?

@bluca I'll try, but I'll open a new issue - or if I have a solution PR - for that.

Thanks for your patience! Closing the PR as it does not look like this is a viable change.

bobrik:
See also: #29263.

PR #23391, opened 2022-05-15
meson: bump required version

yuwata:
- Fedora
  - Fedora 35: ~~0.62.2 (https://bodhi.fedoraproject.org/updates/FEDORA-2022-6db1b2d367)~~ EOL
  - Fedora 36: 0.62.2 (https://bodhi.fedoraproject.org/updates/FEDORA-2022-6f0e3f2c82)
  - Fedora 37+: 1.0.0

- CentOS
  - 8 with PowerTools repo: 0.58.2 (https://centos.pkgs.org/8-stream/centos-powertools-x86_64/meson-0.58.2-2.el8.noarch.rpm.html)
  - 9 with CRB repo: 0.58.2 (https://centos.pkgs.org/9-stream/centos-crb-x86_64/meson-0.58.2-1.el9.noarch.rpm.html)

- Ubuntu: https://packages.ubuntu.com/search?keywords=meson
  - focal (20.04LTS): 0.53.2-2ubuntu2
  - impish (21.10): 0.57.0+really0.56.2-0.1ubuntu1
  - jammy (22.04LTS): 0.61.2-1

- Debian: https://packages.debian.org/search?keywords=meson
  - bullseye (stable): 0.56.2-1
  - bullseye-backports: 1.0.0-1.1~bpo11+1
  - bookworm (testing): 1.0.0-1.1
  - sid (unstable): 1.0.1-4

PR #23399, opened 2022-05-16
Clean up various meson constructs

keszybz:


yuwata:
hm? what's the difference from #23383?

keszybz:
It's only the first half of it.

yuwata:
#23383 is the exactly same as this...

yuwata:
Sorry, I did not mention that, but #23391 has all commits in #23383.

bluca:
I think the confusion might be because there are 3 PRs?

this one
https://github.com/systemd/systemd/pull/23391
https://github.com/systemd/systemd/pull/23383

keszybz:
Indeed.

PR #23403, opened 2022-05-16
Say that we support recent version of major distros and other doc updates

keszybz:


README
```diff
@@ -267,6 +267,24 @@ REQUIREMENTS:
         - strace             (used by test/test-functions)
         - capsh              (optional, used by test-execute)
 
+POLICY FOR SUPPORT OF DISTRIBUTIONS AND ARCHITECTURES:
+        systemd main branch and latest major or stable releases are generally
+        expected to compile on current versions of popular distributions (at
+        least all non-EOL versions of Fedora, Debian unstable/testing/stable,
+        latest Ubuntu LTS and non-LTS relases, openSUSE Tumbleweed/Leap,
```
bluca: /relases/releases/

README
```diff
@@ -267,6 +267,24 @@ REQUIREMENTS:
...
+        non-EOL versions of various distributions. Features which would break
+        compilation on slightly-older distributions will only be introduced if
+        there are significant reasons for this (i.e. supporting them interferes
+        with development or requires too many resources to support). In some
+        cases backports of specific libraries or tools might be required.
+
+        The policy is similar wrt. architecture support. systemd is regularly
+        tested on popular architectures (currently amd64, i386, arm64, ppc64el,
+        and s390x), but should compile and work also on other architectures, for
+        which support has been added. Systemd will emit warnings when
```
bluca: /Systemd/systemd/

bluca (approved):

README
```diff
@@ -271,34 +267,49 @@ REQUIREMENTS:
...
         - capsh              (optional, used by test-execute)
 
+POLICY FOR SUPPORT OF DISTRIBUTIONS AND ARCHITECTURES:
+        systemd main branch and latest major or stable releases are generally
+        expected to compile on current versions of popular distributions (at
+        least all non-EOL versions of Fedora, Debian unstable/testing/stable,
+        latest Ubuntu LTS and non-LTS relases, openSUSE Tumbleweed/Leap,
+        up-to-date Arch, etc.)  We will generally attempt to support also other
+        non-EOL versions of various distributions. Features which would break
+        compilation on slightly-older distributions will only be introduced if
```
evverx: I'm not sure changes breaking builds on newer platforms are covered here. I think this sentence should imply that features aren't available on all platforms presumably supported by `systemd` and its build system shouldn't be introduced.

evverx: Thinking about it more I'm not sure what "breaking compilation" means. If it means that `systemd` shouldn't fail to compile spectacularly it's one thing. It it means that scripts that have been working for years shouldn't stop working it would be another thing.

README
```diff
@@ -271,34 +267,49 @@ REQUIREMENTS:
         - strace             (used by test/test-functions)
         - capsh              (optional, used by test-execute)
 
+POLICY FOR SUPPORT OF DISTRIBUTIONS AND ARCHITECTURES:
+        systemd main branch and latest major or stable releases are generally
+        expected to compile on current versions of popular distributions (at
+        least all non-EOL versions of Fedora, Debian unstable/testing/stable,
```
yuwata, line 274: How about the latest CentOS stream with EPEL? This is purely a question. Not a request to add it.

yuwata (approved):

README
daandemeyer: We backport latest systemd to centos stream (currently 8 and 9) so it'd definitely be nice to explicitly mention those here as well

keszybz: Added now.

README
keszybz: I don't want to make this a law contract. "break compilation" is generally clear. It's not about failing spectacularly — instead, if you can expect that after doing `ninja -C build` you'll have working binaries. I don't think we can say that every feature and every way to consume the source code will always work. There will always be various judgement calls. In particular, I don't want to say that systemd will compile without warnings, or with `-Werror` or `--fatal-meson-warnings`.

evverx: >  if you can expect that after doing ninja -C build you'll have working binaries

In theory it would mean that `install-tests` (or any other option) used by the debian package could be safely removed but my understanding is that patches like that wouldn't be accepted and that idea by the way was immediately rejected because the sytemd build system should be compatible until `meson` can be bumped. On the other hand https://github.com/systemd/systemd/issues/23212 isn't considered important even though technically it's the same thing.

> There will always be various judgement calls

Agreed. I just thought that the policy would be more specific.

yuwata: > I don't want to say that systemd will compile without warnings, or with `-Werror` or `--fatal-meson-warnings`.

I agree with this.

evverx: @yuwata I agree with that as well. By "more specific" I meant that it seems to me that the actual policy is that the debian package should be kept afloat upstream for whatever reason and I don't know why it isn't mentioned anywhere explicitly.

yuwata: I do not know what kind of issues exist in Debian packaging. But this policy does not say that we support build script for packaging on each distribution. And we should not say that, I think.
For example, Fedora also has several custom scripts to building package, and we sometimes update them to support newer release.

evverx: > I do not know what kind of issues exist in Debian packaging

@yuwata I was referring to https://github.com/systemd/systemd/pull/23204#issuecomment-1111043730 where it was mentioned how suddenly important it was to wait for meson to be bumped. That particular feature doesn't work but still.

Anyway, luckily stuff I care about can't be even run or built on Ubuntu  (and its so-called CI) (in the context of the upstream systemd repository) because it has never been brought to the 21 century so I don't have to interact with it very often. Patches like https://github.com/systemd/systemd/pull/22547#issuecomment-1126937590 cut it off from that kind of testing entirely (I don't think it matters though because those kinds of issues aren't a priority there).

README
```diff
@@ -267,6 +267,26 @@ REQUIREMENTS:
         - strace             (used by test/test-functions)
         - capsh              (optional, used by test-execute)
 
+POLICY FOR SUPPORT OF DISTRIBUTIONS AND ARCHITECTURES:
+
+        systemd main branch and latest major or stable releases are generally
+        expected to compile on current versions of popular distributions (at
+        least all non-EOL versions of Fedora, Debian unstable/testing/stable,
+        latest Ubuntu LTS and non-LTS releases, openSUSE Tumbleweed/Leap,
+        CentoOS Stream 8 and 9, up-to-date Arch, etc.)  We will generally
```
marcosfrm, line 276: CentOS

README
evverx: @mbiebl I noticed you downvoted my comments. They were too harsh and I apologize for that.

On a constructive note it doesn't make sense to run some tests without ASan/UBSan and autopkgtest simply doesn't support that. As far as I can remember it was discussed somewhere but nobody was willing to make it work. If anything has changed since then I'd appreciate it if you could point me to autopkgtest issues/PRs where it was implemented so that the CI could be hooked up to it. I'd be happy to do that.

PR #23404, opened 2022-05-16
rpm: remove check if systemd is running in a few cases

keszybz:
rpms can be installed in two different modes: into a chroot, where the system
is not running, and onto a live system. In the first mode, where should create
all changes that are "permanent", and in the second mode, all changes which are
"permanent" but also those which only affect the running system. Thus, changes
like new modprobe rules, tmpfiles rules, binfmt rules, udev rules, etc., are
guarded by 'test -d "/run/systemd/system"' which is the official way to check
if systemd is running, so that they are *not* executed when installed into a
chroot. But the same logic does not apply to sysusers, hwdb, and the journal
catalog: all those files can and should result in changes being performed
immediately to the system. This makes the creation of immutable images possible
(because there are no permanent changes to executed after a reboot), and allows
other packages to depend on the the effect of those changes.

Thus, the guard to check if we're not in a chroot is dropped from triggers for
sysusers, hwdb, and the journal catalog. This means that those triggers will
execute, and no subsequent work is needed. systemd-sysusers.service,
systemd-journal-catalog-update.service, and systemd-hwdb-update.service.in all
have ConditionNeedsUpdate= so they they generally won't be invoked after a
reboot. (systemd.rpm does not touch /usr to trigger the condition, because the
%transfiletriggers make that unnecessary.)

https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=2085481

### DIFF
--- a/src/rpm/triggers.systemd.in
+++ b/src/rpm/triggers.systemd.in
@@ -72,37 +72,31 @@ end
 -- This script will process files installed in {{SYSUSERS_DIR}} to create
 -- specified users automatically. The priority is set such that it
 -- will run before the tmpfiles file trigger.
-if posix.access("/run/systemd/system") then
-    pid = posix.fork()
-    if pid == 0 then
-        assert(posix.execp("systemd-sysusers"))
-    elseif pid > 0 then
-        posix.wait(pid)
-    end
+pid = posix.fork()
+if pid == 0 then
+    assert(posix.execp("systemd-sysusers"))
+elseif pid > 0 then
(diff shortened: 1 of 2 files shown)


bluca (approved):

bluca:
Is this safe enough for v251?

keszybz:
Yes. It's a clear bug fix.

I tested the build in rawhide and it now works as expected.

keszybz:
v249-stable now.

PR #23407, opened 2022-05-17
Clean up bpf cleanup warning

keszybz:


src/core/bpf-lsm.c
```diff
@@ -125,13 +125,15 @@ static int mac_bpf_use(void) {
         }
 }
 
-bool lsm_bpf_supported(void) {
+bool lsm_bpf_supported(bool initialize) {
```
bluca, line 128: Maybe `cache_only` or something like that instead, to reflect what it does rather than what calls it? Just a nit, not too important

src/core/bpf-lsm.c
```diff
@@ -267,7 +269,8 @@ int lsm_bpf_cleanup(const Unit *u) {
         assert(u);
         assert(u->manager);
 
-        if (!lsm_bpf_supported())
+        /* If we never successfully detected support, there is nothing to clean up. */
+        if (!lsm_bpf_supported(false))
```
yuwata: Please add a comment `/* initialize = */`.

yuwata (approved):
LGTM.

src/core/bpf-lsm.c
keszybz: So for me, "initialize" **is** about what it does: it answers the question "should the function do initialization?", and the answer to that is provided by the caller. The fact that there's some caching going on underneath is an implementation detail of the function.

bluca (approved):

PR #23411, opened 2022-05-17
journal: Add new _INITRD field

daandemeyer:
The _INITRD field is a boolean field (0 or 1) that specifies whether
a message was processed by systemd-journald in the initrd or not.

### DIFF
--- a/man/systemd.journal-fields.xml
+++ b/man/systemd.journal-fields.xml
@@ -398,6 +398,14 @@
         for details about journal namespaces.</para>
         </listitem>
       </varlistentry>
+      <varlistentry>
+        <term><varname>_SYSTEM_CONTEXT=</varname></term>
+
+        <listitem><para>A string field that specifies the context in which the message was logged. If
+        <literal>initrd</literal>, the log message was processed while systemd-journald
+        was running inside the initrd. If <literal>main</literal>, the log message was generated after
+        journald switched root to the root filesystem.</para></listitem>
+      </varlistentry>
     </variablelist>
   </refsect1>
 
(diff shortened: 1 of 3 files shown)


poettering:
I think it would be better to maybe have _SYSTEM_LOCALITY=initrd instead? that would pave ground for more "localities" later on, such as _SYSTEM_LOCALITY=exitrd or so?

not sure if "locality" is a great word for this. maybe "context"?

DaanDeMeyer:
> I'd just not set it for "main" context...

But then how do we skip initrd messages? We don't have inverted matches yet and I don't think we can select all entries that don't have a specific field yet either.

keszybz:
I agree with @poettering that having a field with multiple values seems much more forward-looking. In particular, we'll want something like "exitrd" and "crashrd" (for when we reboot into a special kernel after an oops, whatitsacalled).

>  We don't have inverted matches yet

That'd be super neat to add.

dtardon:
> "crashrd" (for when we reboot into a special kernel after an oops, whatitsacalled).

kdump

poettering:
> But then how do we skip initrd messages? We don't have inverted matches yet and I don't think we can select all entries that don't have a specific field yet either.

ok, then maybe it's ok to also add a field for the "maint" context.

Hence _SYSTEM_CONTEXT=initrd|main|exitrd|kdump sounds ok to me.

or _SYSTEM_SCOPE=? _SYSTEM_LOCALITY= _SYSTEM_DISPOSITION?

bluca (approved):

poettering:
lgtm

PR #11814, opened 2019-02-24
semaphoreci: Run subset of autopkgtests in LXC

martinpitt:
Run build/test in LXC for now, as full nested QEMU is too brittle right
now: https://github.com/semaphoreci/semaphore/issues/37
But this at least runs some tests. It ensures that systemd generally
works in containers, as well as provides some backup results if the main
Ubuntu CI is down.

---

Still lots of things to fix (see `FIXME`s in the commit), but this should first demonstrate that this works in principle.

 - [x] This builds on top of PR #11834 , land that first.

PR #23413, opened 2022-05-17
ci: switch to bookworm on Semaphore CI

evverx:
This part of the CI is supposed to test Debian Testing:
https://github.com/systemd/systemd/pull/11814#issuecomment-468157657
Looks like it hasn't been updated for a while.

### DIFF
--- a/.semaphore/semaphore-runner.sh
+++ b/.semaphore/semaphore-runner.sh
@@ -54,6 +54,10 @@ apt-get install -y fdisk tree libfdisk-dev libp11-kit-dev libssl-dev libpwqualit
 apt-get purge --auto-remove -y unattended-upgrades
 systemctl unmask systemd-networkd
 systemctl enable systemd-networkd
+
+# https://github.com/systemd/systemd/issues/23415
+apt-get install -y accountsservice
+systemctl disable accounts-daemon
 EOF
     sudo lxc-stop -n "$CONTAINER"
 }


.semaphore/semaphore-runner.sh
```diff
@@ -6,7 +6,7 @@ set -o pipefail
 
 # default to Debian testing
 DISTRO="${DISTRO:-debian}"
-RELEASE="${RELEASE:-bullseye}"
+RELEASE="${RELEASE:-bookwarm}"
```
daandemeyer: bookworm?

evverx:
I'm not interested in testing bullseye.

> that mkosi already covers it

It doesn't run it in lxc containers and that is the point of that part of the CI among other thins.

.semaphore/semaphore-runner.sh
```diff
@@ -54,6 +54,10 @@ apt-get install -y fdisk tree libfdisk-dev libp11-kit-dev libssl-dev libpwqualit
 apt-get purge --auto-remove -y unattended-upgrades
 systemctl unmask systemd-networkd
 systemctl enable systemd-networkd
+
+# https://github.com/systemd/systemd/issues/23415
+apt-get install -y accountsservice
+systemctl disable accounts-daemon
 EOF
```
mbiebl, line 61: This doesn't make sense.

evverx: Could you be more specific? If there is a better way to address https://github.com/systemd/systemd/issues/23415 I'd appreciate it if you could point me in the right direction

evverx: Anyway I have no idea what exactly doesn't make sense. If by analogy with 8821518b00453559 the test was supposed to be turned off entirely I don't think that's the right approach in general. Minor packaging issues shouldn't prevent systemd from being tested. If it was supposed to mean that it shouldn't be here in the first place I agree. The package should be fixed so as not bug upstream systemd contributors with this kind of irrelevant stuff. It's not the Debian CI after all.

Looking at https://salsa.debian.org/systemd-team/systemd/-/commit/5738b62f5544d040550a018dcd02701bac4feec8 my best guess would be that stuff is turned off downstream and it should go there. I have no idea.

evverx: Looking at https://salsa.debian.org/systemd-team/systemd/-/commit/5738b62f5544d040550a018dcd02701bac4feec8 a bit closer it appears it's expected that the tests are run on bullseye so I can't seem to just switch to bookworm here. I'm not going to even try to touch that so I'll go ahead and close the PR. I'm pretty sure it isn't going to be merged anyway :-)

yuwata:
Why did you close this? Are there any progress on this?

.semaphore/semaphore-runner.sh
yuwata: I have no idea about the workaround, but the basic concept of this PR looks good to me, and I think we should definitely switch to bookworm.

bluca: No, i want bullseye covered until it becomes oldstable next year

evverx: > No, i want bullseye covered until it becomes oldstable next year

I wonder why that would be useful upstream?

evverx:
> Why did you close this?

That part of the CI for some reason appears to have started testing that systemd is buildable on bullseye :-) I don't know what the point of that is but the "upstream" branch of the debian package now moves backwards instead of rolling forward along with systemd and it can't be changed here without touching the debian package.

PR #23414, opened 2022-05-17
systemd-analyze compare-versions

keszybz:


src/test/test-string-util.c
```diff
@@ -868,43 +879,187 @@ TEST(strverscmp_improved) {
...
         /* invalid characters */
         assert_se(strverscmp_improved("123_aa2-67.89", "123aa+2-67.89") == 0);
 }
 
+#define RPMVERCMP(a, b, c) \
+        test_strverscmp_improved_one(STRINGIFY(a), STRINGIFY(b), (c))
+
+TEST(strverscmp_improved_rpm) {
+        /* Tests copied from rmp's rpmio test suite.
+         * The original form is retained for easy comparisons and updates. */
```
bluca: can you add also the link and the note about the license from the commit message? easier to see it as a comment

keszybz: Done.

yuwata (approved):

man/systemd-analyze.xml
```diff
@@ -118,6 +118,13 @@
       <arg choice="plain">cat-config</arg>
       <arg choice="plain" rep="repeat"><replaceable>NAME</replaceable>|<replaceable>PATH</replaceable></arg>
     </cmdsynopsis>
+    <cmdsynopsis>
+      <command>systemd-analyze</command>
+      <arg choice="opt" rep="repeat">OPTIONS</arg>
+      <arg choice="plain">vercmp</arg>
```
bluca: Nit: other verbs use the long format for the name rather than abbreviating, so what about `version-compare` for internal consistency?

man/systemd-analyze.xml
```diff
@@ -547,6 +554,31 @@ NAutoVTs=8
...
+    <refsect2>
+      <title><command>systemd-analyze vercmp <replaceable>VERSION1</replaceable> <replaceable>VERSION2</replaceable></command></title>
+
+      <para>This command will compare the two specified version strings and print
+      either <literal><replaceable>VERSION1</replaceable> &lt; <replaceable>VERSION2</replaceable></literal>,
+      or <literal><replaceable>VERSION1</replaceable> == <replaceable>VERSION2</replaceable></literal>,
+      or <literal><replaceable>VERSION1</replaceable> &gt; <replaceable>VERSION2</replaceable></literal>
+      as appropriate.</para>
+
+      <para>The exit status is <constant>0</constant> if the versions are equal, <constant>11</constant> if
```
bluca, line 573: So not sure we need it, but `dpkg --compare-versions` has a nice interface because it allows the user to specify what the expected result should be, so it's really nice and flexible, eg:

```
$ dpkg --compare-versions 3.2.1.0 ge 0.1.2.3; echo $?
0
$ dpkg --compare-versions 3.2.1.0 lt 0.1.2.3; echo $?
1
```

Do we maybe want the same here?

bluca:
I know it's a very thin wrapper around the function which already has a unit test, but could you please add some minimal exercise of the binary in `test/units/testsuite-65.sh`? It should be really quick to do, and would cover the return code, which is part of the interface

man/systemd-analyze.xml
keszybz: So… I made it `vercmp` to match `rpmdev-vercmp`, but I'm fine with the longer version too if people prefer that.

bluca: If we want this to be part of our public api, and part of the image story, then to me it seems better to go for internal consistency - but I do not have a strong opinion, just an observation

keszybz:
Updated. I renamed the verb to `compare-versions`, and the interfaces of both `rpmdev-vercmp` and `dpkg --compare-versions` are now supported. This should make it easy for people to switch over to our implementation. The test is added as requested.

bluca (approved):

bluca:
> Updated. I renamed the verb to `compare-versions`, and the interfaces of both `rpmdev-vercmp` and `dpkg --compare-versions` are now supported. This should make it easy for people to switch over to our implementation. The test is added as requested.

Nice, thanks!

keszybz:
mkosi / ci (arch, rolling) failed because the internet is broken.

man/systemd-analyze.xml
```diff
@@ -547,6 +555,52 @@ NAutoVTs=8
...
+      <replaceable>VERSION2</replaceable></literal>, or <literal><replaceable>VERSION1</replaceable> &gt;
+      <replaceable>VERSION2</replaceable></literal> as appropriate.</para>
+
+      <para>The exit status is <constant>0</constant> if the versions are equal, <constant>11</constant> if
+      the version of the right is smaller, and <constant>12</constant> if the version of the left is
+      smaller. (This matches the convention used by <command>rpmdev-vercmp</command>.)</para>
+
+      <para>In the second mode — when <replaceable>OP</replaceable> is specified — it will compare the two
+      version strings using the operation <replaceable>OP</replaceable> and return <constant>0</constant>
+      (success) if they condition is satisfied, and <constant>0</constant> (failure)
```
dtardon: `<constant>1</constant>` here?

man/systemd-analyze.xml
```diff
@@ -1197,8 +1251,12 @@ $ systemd-analyze verify /tmp/source:alias.service
...
     <title>Exit status</title>
 
-    <para>On success, 0 is returned, a non-zero failure code
-    otherwise.</para>
+    <para>For most commands, 0 is returned on success, and a non-zero failure code otherwise.</para>
+
+    <para>With the verb <command>compare-versions</command>, in the two-argument form,
+    <constant>12</constant>, <constant>0</constant>, <constant>11</constant> is returned if the second
+    version string is respectively larger, equal, or smaller to the first. In the three-argument form,
+    <constant>0</constant> or <constant>0</constant> if the condition is respectively true or false.</para>
```
dtardon: Here too, `<constant>1</constant>` for failure?

poettering:
I think we should also rework the function to not use locale dependent calls anymore, so that it has the same results everywhere.

i.e. isdigit() and isalpha() are all locale dependent. we should implement simpler replacements that just care about 7bit ASCII and nothing else.

src/analyze/analyze-blame.c
```diff
@@ -61,5 +61,9 @@ int verb_blame(int argc, char *argv[], void *userdata) {
 
         pager_open(arg_pager_flags);
 
-        return table_print(table, NULL);
+        r = table_print(table, NULL);
+        if (r < 0)
+                return r;
+
+        return 0;
```
poettering: maybe `return EXIT_SUCCESS` now?

src/analyze/analyze-capability.c
```diff
@@ -48,5 +48,9 @@ int verb_capabilities(int argc, char *argv[], void *userdata) {
 
         pager_open(arg_pager_flags);
 
-        return table_print(table, NULL);
+        r = table_print(table, NULL);
+        if (r < 0)
+                return r;
+
+        return 0;
```
poettering: ditto (and everywhere else).

i.e. if this is propagated as is now if > 0 we should probably use only EXIT_xyz names for returning here

src/analyze/analyze-compare-versions.c
```diff
@@ -0,0 +1,45 @@
...
+        int r;
+
+        assert(argc == 3 || argc == 4);
+        assert(argv);
+
+        if (argc == 3) {
+                r = strverscmp_improved(ASSERT_PTR(argv[1]), ASSERT_PTR(argv[2]));
+                printf("%s %s %s\n",
+                       argv[1],
+                       r < 0 ? "<" : r > 0 ? ">" : "==",
```
poettering, line 20: it's a bit weird, we output `<`, `>`and `==`, but we aprse `lt`, `gt`, `eq`.

i think we should at the very least accept both when parsing...

poettering: Also, there's some code to share here with ConditionKernelVersion= i think, i.e. the OrderOperator enum

keszybz: `<` and `>` are inconvenient to type because of the need to quote… I can add them, but maybe let's not document them in the man page.

keszybz:
Updated. There's some more commits now, so a re-review is needed.

bluca:
New one?

```
+ /usr/bin/patch -p1 -s --fuzz=0 --no-backup-if-mismatch -f
+ /usr/lib/rpm/rpmuncompress /builddir/build/SOURCES/f58b96d3e8d1cb0dd3666bc74fa673918b586612.patch
+ /usr/lib/rpm/rpmuncompress /builddir/build/SOURCES/0001-Revert-meson-create-new-libsystemd-core.so-private-s.patch
+ /usr/bin/patch -p1 -s --fuzz=0 --no-backup-if-mismatch -f
1 out of 2 hunks FAILED -- saving rejects to file meson.build.rej
```

bluca (approved):

poettering:
What about sharing the parser with ConditionKernelVersion?

keszybz:
> What about sharing the parser with ConditionKernelVersion?

The important part is parsing of the version, which is shared, because both paths use `strverscmp_improved()`.
For the rest, we could try to share the part that extracts the operator, but I think that'd be more trouble than it's worth. The check is completely trivial here, because we get the operator as a separate item in argv[], at a known position, and the interpretation of the operator is also very simple. So I think this is fine as is.

src/fundamental/string-util-fundamental.c
```diff
@@ -77,19 +77,18 @@ sd_char* endswith_no_case(const sd_char *s, const sd_char *postfix) {
         return (sd_char*) s + sl - pl;
 }
 
-#ifdef SD_BOOT
-static sd_bool isdigit(sd_char a) {
+static sd_bool is_digit(sd_char a) {
```
poettering, line 80: maybe we should use a more explicit name. The underscore is not a good indicator that this is strictly doing ASCII stuff. hence, following the logic of ascii_toupper(), ascii_strlower(), ascii_strcasecmp(), we should call this helper ascii_isdigit() I think.

poettering: and probably export it, in fact.

PR #23417, opened 2022-05-17
docs: add dark mode

bnf:
Thanks for the awesome documents on https://systemd.io – a nice dark mode would be the icing on the cake.

This PR adds a bunch of preflight fixes to streamline HTML markup and structure with the goal to add dark mode support in a clean way using CSS variables.

We respect the user agent property  `prefers-color-scheme` by
rendering the documentation in dark mode, if requested.

The CSS is reorganised to store the color palette into CSS variables
and reference these from another set of CSS variables that
are dynamically switched based on the prefers-color-scheme
media query.

Light mode variables stay they same as before,
but to ease color mixing, the current RGB values
have been transformed to HSL (using chrome devtools).
The current body background is now --sd-gray-extralight,
the current body color is now --sd-gray-extradark.
Other gray-variations, needed for dark-mode constrasts
are derived from these colors using HSL lightness shifting.

The systemd brand black color is used as dark mode background and a matching extralight gray font color has been selected.

The link font-weight is reduced to 400 in dark mode,
as the green color on dark ground becomes to overwhelming
with a bold font.

The systemd logo color is dynamically swapped by using
the special fill value `currentColor` for dark/light-mode
depending parts – as per specification on brand.systemd.io

A preview of this branch is available on https://systemd.bnf.dev (served by https://github.com/bnf/systemd/tree/docs which contains one additional commit to configure for the preview domain) and can be tested via Google Chrome, Gnome Web:

```bash
google-chrome-stable --enable-features=WebUIDarkMode --force-dark-mode https://systemd.bnf.dev/
```

```bash
epiphany https://systemd.bnf.dev/
# enable dark mode
gsettings set org.gnome.desktop.interface color-scheme prefer-dark
# (re)set to light mode
gsettings set org.gnome.desktop.interface color-scheme prefer-light
```

Related to https://github.com/systemd/brand.systemd.io/issues/4 – I would be willing to add the adaptions to [brand.systemd.io](https://github.com/systemd/brand.systemd.io/) once we agree on the changes.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -33,63 +33,63 @@ Other parts include a logging daemon, utilities to control basic system configur
 
 ---
 
-<pre style="color:white; background-color:black; font-size:smaller; padding:6pt 8pt">
-Welcome to <span style="color:blue">Fedora 20 (Heisenbug)</span>!
+<pre class="intro-code-block">
+Welcome to <span class="color-blue">Fedora 20 (Heisenbug)</span>!
 
-[  <span style="color:green">OK</span>  ] Reached target Remote File Systems.
-[  <span style="color:green">OK</span>  ] Listening on Delayed Shutdown Socket.
-[  <span style="color:green">OK</span>  ] Listening on /dev/initctl Compatibility Named Pipe.
-[  <span style="color:green">OK</span>  ] Reached target Paths.
-[  <span style="color:green">OK</span>  ] Reached target Encrypted Volumes.
-[  <span style="color:green">OK</span>  ] Listening on Journal Socket.
+[  <span class="color-green">OK</span>  ] Reached target Remote File Systems.
+[  <span class="color-green">OK</span>  ] Listening on Delayed Shutdown Socket.
+[  <span class="color-green">OK</span>  ] Listening on /dev/initctl Compatibility Named Pipe.
+[  <span class="color-green">OK</span>  ] Reached target Paths.
+[  <span class="color-green">OK</span>  ] Reached target Encrypted Volumes.
+[  <span class="color-green">OK</span>  ] Listening on Journal Socket.
          Mounting Huge Pages File System...
          Mounting POSIX Message Queue File System...
          Mounting Debug File System...
          Starting Journal Service...
-[  <span style="color:green">OK</span>  ] Started Journal Service.
+[  <span class="color-green">OK</span>  ] Started Journal Service.
          Mounting Configuration File System...
          Mounting FUSE Control File System...
-[  <span style="color:green">OK</span>  ] Created slice Root Slice.
-[  <span style="color:green">OK</span>  ] Created slice User and Session Slice.
-[  <span style="color:green">OK</span>  ] Created slice System Slice.
-[  <span style="color:green">OK</span>  ] Reached target Slices.
-[  <span style="color:green">OK</span>  ] Reached target Swap.
+[  <span class="color-green">OK</span>  ] Created slice Root Slice.
+[  <span class="color-green">OK</span>  ] Created slice User and Session Slice.
+[  <span class="color-green">OK</span>  ] Created slice System Slice.
+[  <span class="color-green">OK</span>  ] Reached target Slices.
+[  <span class="color-green">OK</span>  ] Reached target Swap.
          Mounting Temporary Directory...
-[  <span style="color:green">OK</span>  ] Reached target Local File Systems (Pre).
+[  <span class="color-green">OK</span>  ] Reached target Local File Systems (Pre).
          Starting Load Random Seed...
          Starting Load/Save Random Seed...
-[  <span style="color:green">OK</span>  ] Mounted Huge Pages File System.
-[  <span style="color:green">OK</span>  ] Mounted POSIX Message Queue File System.
-[  <span style="color:green">OK</span>  ] Mounted Debug File System.
-[  <span style="color:green">OK</span>  ] Mounted Configuration File System.
-[  <span style="color:green">OK</span>  ] Mounted FUSE Control File System.
-[  <span style="color:green">OK</span>  ] Mounted Temporary Directory.
-[  <span style="color:green">OK</span>  ] Started Load Random Seed.
-[  <span style="color:green">OK</span>  ] Started Load/Save Random Seed.
-[  <span style="color:green">OK</span>  ] Reached target Local File Systems.
+[  <span class="color-green">OK</span>  ] Mounted Huge Pages File System.
+[  <span class="color-green">OK</span>  ] Mounted POSIX Message Queue File System.
+[  <span class="color-green">OK</span>  ] Mounted Debug File System.
+[  <span class="color-green">OK</span>  ] Mounted Configuration File System.
+[  <span class="color-green">OK</span>  ] Mounted FUSE Control File System.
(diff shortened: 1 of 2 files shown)


docs/style.css
```diff
@@ -38,6 +38,7 @@ h1, h2, h3, h4, h5, h6 {
 h1 {
   text-align: center;
   font-size: 30px;
+  font-weight: 400;
```
medhefgo, line 88: Something's off here, one line down replaces this with `font-weight: 100;`?

docs/style.css
```diff
@@ -12,6 +12,52 @@
...
+
+  --sd-highlight-bg-light: rgba(255, 255, 255, 1);
+  --sd-highlight-bg-dark: rgba(0, 0, 0, .6);
+  --sd-highlight-inline-bg-light: rgba(0, 0, 0, 0.07);
+  --sd-highlight-inline-bg-dark: rgba(255, 255, 255, 0.1);
+
+  --sd-font-weight-normal: 400;
+  --sd-font-weight-bold: 600;
+
+  /* Light mode variables */
```
medhefgo, line 37: My html/css-fu is from the stone age… Why are the light-mode values the default here while further down they're behind a media selector?

docs/index.md
```diff
@@ -33,63 +33,63 @@ Other parts include a logging daemon, utilities to control basic system configur
 
 ---
 
-<pre style="color:white; background-color:black; font-size:smaller; padding:6pt 8pt">
-Welcome to <span style="color:blue">Fedora 20 (Heisenbug)</span>!
+<pre class="intro-code-block">
```
medhefgo, line 36: I disagree on how this block looks in light mode. The brand green/blue makes for terrible contrast with the new lighter background color. Dark mode looks okay, though.

medhefgo (commented):
I must say, looks really neat.

To somebody who has the stakes in this while I noticed this now: Why are the manpages not on systemd.io also?

docs/style.css
bnf: oops, rebase mistake. Fixed.

docs/index.md
bnf: Thanks, I fully agree, that looked totally wrong in light mode.
It broke during late-minute refactoring and I didn't retest that case. Fixed by using brand-black as background color.

docs/style.css
bnf: The dark-mode css further down (code highlighting) is not able to overwrite all selectors, as `rougify` produced more css rules for the light-mode style, than for the dark mode, that means both mode would collide, If I wouldn't make them exclusive.

Now, in this case `light` mode acts as default where *all* variables are overwirtten by the dark-mode media query.

We could also encapsulate the light-mode values into a media query, I'm not aware of any browser that would fail in that case (e.g. not applying any color at all) – but to be safe I thought It would be good to provide these as defaults, just in case a browser would ignore both `prefers-color-scheme: dark` and `prefers-color-scheme: dark` media query.

For the case further down, It was ok for me in case an odd browser shows no custom colors at all, therefore the distinct definitions over there.

bluca (approved):
I know nothing about CSS, but the dark mode docs look absolutely smashing

bnf:
Thanks @medhefgo for your comments. Fixed and force pushed: https://github.com/systemd/systemd/compare/9180d64..2eca6d6

medhefgo (approved):

keszybz:
Thanks! This looks very nice. Can we close https://github.com/systemd/brand.systemd.io/issues/4? You mentioned some adjustments there…

bnf:
Thanks. https://github.com/systemd/brand.systemd.io/issues/4 can be kept open for now as I'll file a PR that applies the changes to https://brand.systemd.io/ as well.

Tachi107:
I personally find the background a bit too purple-y, but looks great nonetheless. Thanks!

bnf:
> I personally find the background a bit too purple-y, but looks great nonetheless. Thanks!

I wasn't sure entirely sure either.
Still, here are the arguments that let me use this purple-y black as background color:

- The color used is the systemd-brand black color as defined on http://brand.systemd.io/.
- Not using true-black is in line with all the others (Material Design, Adwaita, Github) that uses dark gray background-variants but not just `#000000`.
- The other existing dark color (which is used for the light mode text-color) is actually purple-bluish (plurple) and would actually offer less contrast, so that wasn't an option either.
- I didn't want to mix an additional color.
- The color offers a nice individual visual touch (fully subjective).

Maybe @bertob could comment whether it's OK to use the brand black `#201A26` as full-surface background color, or whether it was intended to only be used in direct context with the [logo](http://brand.systemd.io/assets/svg/systemd-dark.svg).

Tachi107:
Hey @bnf, sorry for my late reply.

> here are the arguments that let me use this purple-y black as background color:
> 
>    * The color used is the systemd-brand black color as defined on http://brand.systemd.io/.
> 
>    * Not using true-black is in line with all the others (Material Design, Adwaita, Github) that uses dark gray background-variants but not just `#000000`.

Yes, and `#000000` also makes text harder to read (or is it just me?)

>    * I didn't want to mix an additional color.

I agree, reusing existing colors is generally better for consistency.

>    * The color offers a nice individual visual touch (fully subjective).

Yep, that purple definitely has some character.

That being said, I came here again after spending some more time with the dark mode. Especially on my monitor, the background color causes quite some eye strain for me. This isn't a big issue on my phone though, that has a high quality IPS display, so maybe this particular background color becomes problematic only on low quality TN panels?

I experimented a bit with different background colors, ~all taken from Adwaita. I chose this color palette because Adwaita and GNOME are a big part of the Linux ecosystem (more than Material Design, for example). The list of colors is available here: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/named-colors.html.

- `#242424`: this is the dark `window_bg_color`. It is the main color used for window background of apps using dark mode. I find it a bit to bright for our use case.
image

- `#1e1e1e`: slightly darker, used for widgets. This doesn't look bad IMO.
image

- `#101010`: mentioned by @A6GibKm in [this MR](https://gitlab.gnome.org/GNOME/geary/-/merge_requests/714), it is used by some GNOME/Adwaita apps when the main content is white text on a dark background. Maybe a little too dim.
image

- `#121212`: not a classic Adwaita color, but it it used as the background color of the Adwaita page I linked earlier. Doesn't look bad, but maybe it is still a bit too dim?
image

Maybe I'm the only one who's having issues with this purple-y background, but I wanted to leave my feedback nonetheless :)

bnf:
Thanks @Tachi107 for your analysis and nitpicking.
The more I look at it I also get a bit annoyed by the salient purple-y background.

So I would be fine with desaturating the background – I'd like to see a shade between your first two variants and it would be nice to derive it from the systemd brand black `hsl(270, 19%, 13%)`. I would therefore propose to reduce saturation from 19% to 7%: `hsl(270, 7%, 13%)` (that is basically `#211f23` in RGB, which is between your variants 1 and 2) 

Preview:
https://systemd.bnf.dev/
systemd-io-bg-7-satuation

What do you think?

Tachi107:
Using a color derived from systemd brand black is a great idea! It 
looks nice too, as the purple tint is really subtle. I'm not a UI 
designer, but as a normal user I can say that I like this :)

keszybz:
Looks nice.

PR #23418, opened 2022-05-17
resolved: default to OpenSSL

bluca:
If both gnutls and openssl are available, prefer openssl.
We are gradually moving toward supporting openssl only as the
crypto library, and the resolved gnutls backend will be dropped
at some point, so start nudging users toward the openssl one.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1595,20 +1595,20 @@ if dns_over_tls != 'false'
                 error('Sorry, -Ddns-over-tls=gnutls is not supported when openssl is used as the cryptolib')
         endif
 
-        if dns_over_tls == 'openssl' or conf.get('PREFER_OPENSSL') == 1
-                have_gnutls = false
-        else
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -1595,20 +1595,20 @@ if dns_over_tls != 'false'
...
         endif
-        if dns_over_tls == 'gnutls' or have_gnutls
-                have_openssl = false
+        if dns_over_tls == 'openssl' or have_openssl
+                have_gnutls = false
         else
-                have_openssl = conf.get('HAVE_OPENSSL') == 1
-                if dns_over_tls != 'auto' and not have_openssl
-                        str = dns_over_tls == 'openssl' ? ' with openssl' : ''
+                have_gnutls = (conf.get('HAVE_GNUTLS') == 1 and libgnutls.version().version_compare('>= 3.6.0'))
```
keszybz: () not necessary here. (I know that this was preexsting, but still…)

keszybz (commented):
Looks OK, but please write in the commit message *why* we're doing this change now.

bluca:
Updated

PR #23420, opened 2022-05-18
meson: Disable elf semantic interposition in release mode

crrodriguez:
Using ELF interposition on any systemd component is unlikely
to be wise or recommended but what is absolutely not supported
is interposing symbols with different semantics or side effects,
yet we are still paying for it.

Use -fno-semantic-interposition by default which usually generates
better code, sometimes quite dramatically.

Note that clang does this by default and therefore does not
have this switch, only -fsemantic-interposition to enable the
misfeature explicitly.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -453,13 +453,13 @@ possible_cc_flags = possible_common_cc_flags + [
         '-fstack-protector-strong',
         '-fvisibility=hidden',
         '--param=ssp-buffer-size=4',
+        '-fno-semantic-interposition',
 ]
 
 if get_option('buildtype') != 'debug'
         possible_cc_flags += [
                 '-ffunction-sections',
                 '-fdata-sections',
-                '-fno-semantic-interposition',
         ]
 
         possible_link_flags += '-Wl,--gc-sections'


meson.build
```diff
@@ -429,6 +429,7 @@ if get_option('buildtype') != 'debug'
         possible_cc_flags += [
                 '-ffunction-sections',
                 '-fdata-sections',
+                '-fno-semantic-interposition',
```
keszybz: I think we should do this unconditionally, i.e. not exclude debug mode from this. With the exclusion, the code we test in debug mode could behave significantly differently from the release code, which we don't want.

poettering: i agree, this should just be done always

crrodriguez: ok, let me check.

crrodriguez: did as requested

evverx:
As far as I can tell CIFuzz failed because https://github.com/llvm/llvm-project/commit/bc56097817bec3446edd9f06e2d78eab95033027 hasn't been included in the OSS-Fuzz base builder image yet.

Looking at https://reviews.llvm.org/D102453 and https://reviews.llvm.org/D117183 I'm not sure how safe it is to use `-fno-semantic-interposition` with `clang < 14` by default.

evverx:
I took a look at `cpython` (which as far as I can tell is the only large project where `-fno-semantic-interposition` is used upstream). To judge from https://github.com/python/cpython/blob/5f3c9fda1825737fa7b671b995f84a8ab9a4adb8/configure.ac#L1713-L1743 it's passed to `gcc` only when `--enable-optimizations` is set. I think If it was safe to roll it out by default everywhere it would have been done there probably. Anyway, I don't think this flag should be passed when `clang` is used to build `systemd`.

Issue #23416, opened 2022-05-17
`networkctl reload`: Failed to reload network settings: No route to host


g0tar:
**systemd version the issue has been seen with**

Current git HEAD.

**Used distribution**

PLD Linux

**Linux kernel version used** (`uname -a`)

5.9.6

**CPU architecture issue was seen on**

x86_64

**Expected behaviour you didn't see**

`networkctl` reloading configuration

**Unexpected behaviour you saw**

`Failed to reload network settings: No route to host`

**Steps to reproduce the problem**

0. craft brand new system,
1. compile current code (https://github.com/systemd/systemd/issues/23315#issuecomment-1123135986),
2. take freshly compiled `systemd-networkd`, `networkctl` and `systemd`, call `systemd daemon-reexec`,
3. `systemctl start systemd-networkd`
4. 
```
# networkctl
IDX LINK   TYPE     OPERATIONAL SETUP    
  1 lo     loopback carrier     unmanaged
  2 eth0   ether    carrier     unmanaged
  3 eth1   ether    off         unmanaged
  4 eth2   ether    off         unmanaged
  5 eth3   ether    off         unmanaged
  6 eth4   ether    off         unmanaged
  7 eth5   ether    off         unmanaged
  8 eth0.6 vlan     routable    unmanaged
```
5. `mkdir /etc/systemd/network`
6. `/etc/systemd/network/05-x5.network`
```
[Match]
PermanentMACAddress=90:1b:0e:0a:1a:e3
Type=ether
```
7. `/etc/systemd/network/20-x5-v111.netdev`
```
[NetDev]
Name=v111
Kind=vlan

[VLAN]
Id=111
```
8. `/etc/systemd/network/05-x5.network.d/v111.conf`
```
[Network]
VLAN=v111
```
9.
```
# networkctl reload
Failed to reload network settings: No route to host
```
10. 
```
# systemctl restart systemd-networkd.service
# networkctl
IDX LINK   TYPE     OPERATIONAL SETUP     
  1 lo     loopback carrier     unmanaged
  2 eth0   ether    carrier     unmanaged
  3 eth1   ether    off         unmanaged
  4 eth2   ether    off         unmanaged
  5 eth3   ether    off         unmanaged
  6 eth4   ether    off         unmanaged
  7 eth5   ether    no-carrier  configured
  8 eth0.6 vlan     routable    unmanaged
  9 v111   vlan     off         unmanaged
# journalctl -u systemd-networkd.service
[...]

PR #23421, opened 2022-05-18
networkctl: show error message provided through dbus

yuwata:
Prompted by #23416.


DaanDeMeyer (approved):

yuwata:
'after-next-release'?? I think this is safe to be in v251, though this is not so important.

DaanDeMeyer:
> 'after-next-release'?? I think this is safe to be in v251, though this is not critical.

I figured on the day of release we should only merge critical stuff. If others disagree I don't mind merging though.

yuwata:
OK. No problem for me.

PR #23427, opened 2022-05-18
dissect-image: Explicitly remove partitions when done with image

daandemeyer:
When closing a loop device, the kernel will asynchronously remove
the probed partitions. This can lead to race conditions where we
try to reuse a partition device that still needs to be removed by
the kernel. To avoid such issues, let's explicitly try to remove
any partitions using BLKPG_DEL_PARTITION when we're done with an
image.

### DIFF
--- a/src/gpt-auto-generator/gpt-auto-generator.c
+++ b/src/gpt-auto-generator/gpt-auto-generator.c
@@ -768,6 +768,8 @@ static int enumerate_partitions(dev_t devnum) {
                         r = k;
         }
 
+        dissected_image_relinquish(m);
+
         return r;
 }
 
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -148,10 +148,44 @@ static void check_partition_flags(
                 log_debug("Unexpected partition flag %llu set on %s!", bit, node);
         }
 }
+
+static int ioctl_partition_remove(int fd, const char *name, int nr) {
+        assert(fd >= 0);
+        assert(name);
+        assert(nr > 0);
+
+        struct blkpg_partition bp = {
+                .pno = nr,
+        };
+
+        struct blkpg_ioctl_arg ba = {
+                .op = BLKPG_DEL_PARTITION,
+                .data = &bp,
+                .datalen = sizeof(bp),
+        };
+
+        if (strlen(name) >= sizeof(bp.devname))
+                return -EINVAL;
+
+        strcpy(bp.devname, name);
+
+        return RET_NERRNO(ioctl(fd, BLKPG, &ba));
+}
 #endif
 
-static void dissected_partition_done(DissectedPartition *p) {
+static void dissected_partition_done(int fd, DissectedPartition *p) {
+        assert(fd >= 0);
         assert(p);
+
+#if HAVE_BLKID
+        if (p->node && p->partno > 0 && !p->relinquished) {
+                int r;
+
+                r = ioctl_partition_remove(fd, p->node, p->partno);
+                if (r < 0)
+                        log_debug_errno(r, "BLKPG_DEL_PARTITION failed, ignoring: %m");
+        }
+#endif
 
         free(p->fstype);
         free(p->node);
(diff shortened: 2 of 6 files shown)


bluca:
Nice! I'll re-trigger the focal jobs a few times after they finish

bluca (approved):

src/shared/dissect-image.c
```diff
@@ -332,9 +366,14 @@ int dissect_image(
                 return -ENOMEM;
 
         *m = (DissectedImage) {
+                .fd = -1,
                 .has_init_system = -1,
         };
 
+        m->fd = fcntl(fd, F_DUPFD_CLOEXEC, 3);
+        if (m->fd < 0)
+                return m->fd;
```
bluca: -errno?

daandemeyer: Yikes

src/shared/dissect-image.c
```diff
@@ -148,11 +148,45 @@ static void check_partition_flags(
...
+        assert(fd >= 0);
         assert(p);
 
+#if HAVE_BLKID
+        if (p->node && p->partno > 0 && !p->relinquished) {
+                int r;
+
+                r = ioctl_partition_remove(fd, p->node, p->partno);
+                if (r < 0)
+                        log_debug_errno(r, "BLKPG_DEL_PARTITION failed: %m");
```
yuwata: Please mention that the error will be ignored.

src/shared/dissect-image.c
```diff
@@ -790,10 +829,14 @@ int dissect_image(
                                          * scheme in OS images. */
 
                                         if (!PARTITION_DESIGNATOR_VERSIONED(designator) ||
-                                            strverscmp_improved(m->partitions[designator].label, label) >= 0)
+                                            strverscmp_improved(m->partitions[designator].label, label) >= 0) {
+                                                r = ioctl_partition_remove(fd, node, nr);
+                                                if (r < 0)
+                                                        log_debug_errno(r, "BLKPG_DEL_PARTITION failed: %m");
```
yuwata: Also here.

src/shared/dissect-image.c
```diff
@@ -863,8 +906,12 @@ int dissect_image(
                                 const char *sid, *options = NULL;
 
                                 /* First one wins */
-                                if (m->partitions[PARTITION_XBOOTLDR].found)
+                                if (m->partitions[PARTITION_XBOOTLDR].found) {
+                                        r = ioctl_partition_remove(fd, node, nr);
+                                        if (r < 0)
+                                                log_debug_errno(r, "BLKPG_DEL_PARTITION failed: %m");
```
yuwata: Also here.

src/gpt-auto-generator/gpt-auto-generator.c
```diff
@@ -768,6 +768,8 @@ static int enumerate_partitions(dev_t devnum) {
                         r = k;
         }
 
+        dissected_image_relinquish(m);
```
daandemeyer, line 771: This fixes gpt-auto-generator in mkosi qemu VMs but I'm not sure if it's actually the right fix or why it's necessary at all. Shouldn't the partitions get destroyed regardless at the end of this block when the parent block device fd gets closed?

bluca:
I've restarted the focal-amd64 and focal-ppc64el jobs a few times, they never failed again. Just did it one more time - if it still goes through, I think we can merged this.

src/gpt-auto-generator/gpt-auto-generator.c
yuwata: Please mention why this is necessary or what we need to investigate later.

daandemeyer: Looking at it again, it makes sense. In gpt-auto-generator, we're operating on a block device and not a loop device. When closing the fd to the block device, I think the partitions just don't get removed automatically like what happens when closing a loop device fd. If that's the case, it makes sense that we have to call relinquish here to make sure we don't remove the partitions manually either.

Without this in place, booting using gpt-auto-generator will fail as systemd will time out waiting for the root partition device to appear (which will never appear since we remove it in the dissected_image_unref function unless we call relinquish first.

PR #23432, opened 2022-05-18
docs: use relative links

bnf:
Allows for links to work both on systemd.io (or forks) and
when viewed on https://github.com/systemd/systemd/tree/main/docs

Note that the markdown links are converted by jekyll-relative-links[1]
to html. This plugin is enabled by default on github pages[2][3].

Due to a bug in jekyll-relative-links – see
https://github.com/benbalter/jekyll-relative-links/issues/61 –
we need to avoid line-wrapped links when using relative markdown links.

Also adapt the header levels in `JOURNAL_EXPORT_FORMATS.md` to ensure that the section-links (from `JOURNAL_FILE_FORMAT.md` make sense.

Preview available on https://systemd.bnf.dev/

[1] https://github.com/benbalter/jekyll-relative-links
[2] https://github.blog/2016-12-05-relative-links-for-github-pages/
[3] https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/about-github-pages-and-jekyll#plugins

### DIFF
--- a/docs/JOURNAL_EXPORT_FORMATS.md
+++ b/docs/JOURNAL_EXPORT_FORMATS.md
@@ -1,11 +1,13 @@
 ---
-title: Journal Export Format
+title: Journal Export Formats
 category: Interfaces
 layout: default
 SPDX-License-Identifier: LGPL-2.1-or-later
 ---
 
-# Journal Export Format
+# Journal Export Formats
+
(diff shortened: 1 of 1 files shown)


keszybz:
Nice! Those absolute links were annoying.

Issue #23429, opened 2022-05-18
PR#23218 introduced a regression for devices with `db_persist` option


fbuihuu:
Devices with `OPTIONS+="db_persist"` are not supposed to create `dead->plugged` transition right after switching root, at least that wasn't the case before and the udev man page describes the option to actually prevent it from happening.

However pr #23218 changed this behavior.  I don't know currently whether it will cause any real regression in practice but I think it deserves a discussion before releasing v251.

mwilck:
I wouldn't call this a regression. No actual negative impact is known except the debug messages.

See also https://github.com/systemd/systemd/pull/23218#issuecomment-1130496007

fbuihuu:
Hi Martin, it's pretty hard to say for sure that it won't have any negative impact but the facts are that this transition never happened before and the doc clearly describes the use of "db_persist" as way to make sure that the device unit state remains always plugged.

yuwata:
https://github.com/systemd/systemd/pull/23218#issuecomment-1130496007
@mwilck 
> > It's not clear to me whether it will cause issue in practice (maybe @mwilck knows ?) but this `dead->plugged` transition for devices with `db_persist` doen't look correct and wasn't present before. And udev man page seems to indicate that this transition should not occur.
> 
> Why would this even be mentioned in an the udev man page? It's genuine systemd functionality.
> 
> I have observed these transitions (I see them all the time actually because I work with multipath-tools, but I didn't bother about it), but definitely no negative impact.
> 
> I have a test case here where root FS (btrfs) is on top of LVM on top of MD RAID on top of multipath. Thus the multipath devices (which have `db_persist` set) are in use once root is mounted, but from the PoV of systemd, they have neither `DEVICE_FOUND_MOUNT` nor `DEVICE_FOUND_SWAP` set. Thus they start out "dead" after switching root, just like in @fbuihuu's example. I can see these dead→plugged transitions with loglevel debug, but everything works just fine during boot.

https://github.com/systemd/systemd/pull/23218#issuecomment-1130503768
> @yuwata. recall [this discussion](https://github.com/systemd/systemd/pull/23218#discussion_r862654712) we had previously? If we removed the lines
> 
> ```
> +                if (found == DEVICE_NOT_FOUND)
> +                        state = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
> ```
> 
[...]

PR #23436, opened 2022-05-19
Device fix plug dead plug again

fbuihuu:


fbuihuu:
@yuwata here's the PR you asked me to create for discussing a possible and simpler way to fix issues addressed by # 23218. The second commit contains the fix whereas the last one is a tentative to drop `device_coldplug()` as it seems to not be needed after all.

src/core/device.c
```diff
@@ -158,59 +158,38 @@ static void device_set_state(Device *d, DeviceState state) {
...
+                 * trust the enumeration step anymore until udev DB is repopulated again. Until
+                 * this mechanism is used during initrd transitions, we have no other choice than
+                 * forgetting all devices and relying on a full retriggering of events (based on
+                 * the host set of udev rules this time). Hence we avoid generating wrong device
+                 * state transitions by not assuming wrongly that devices missing in udev DB were
+                 * unplugged. The downside is that PID1 won't generate plugged->dead transitions
+                 * for devices that were unplugged during the initrd transition. */
+
+                found = DEVICE_NOT_FOUND;
+                state = DEVICE_DEAD;
```
yuwata: This re-introduces the previous plugged -> dead -> plugged issue for devices without persistent database.

src/core/device.c
```diff
@@ -187,21 +176,10 @@ static int device_coldplug(Unit *u) {
...
-        return 0;
-}
-
-static void device_catchup(Unit *u) {
-        Device *d = DEVICE(u);
-
-        assert(d);
-
-        /* Second, let's update the state with the enumerated state */
         device_update_found_one(d, d->enumerated_found, DEVICE_FOUND_MASK);
```
yuwata, line 183: This does update `Device.found` and `Device.state`, and may not trigger required transactions.
E.g. on reload, if `d->state == DEVICE_PLUGGED` and `d->found == d->enumerated_found` and they have `DEVICE_FOUND_UDEV` flag, then `device_update_found_one()` does not call device_set_state()...

yuwata (changes requested):
The first and third commits LGTM. I misunderstood them previously, sorry.

However, the second and fourth commits look not correct. Please see the comments below.

keszybz:
```
2022-05-19T07:47:33.3416115Z ../src/core/device.c: In function ‘device_init’:
2022-05-19T07:47:33.3417051Z ../src/core/device.c:95:17: error: unused variable ‘d’ [-Werror=unused-variable]
2022-05-19T07:47:33.3417572Z    95 |         Device *d = DEVICE(u);
2022-05-19T07:47:33.3417904Z       |                 ^
2022-05-19T07:47:33.3418267Z cc1: all warnings being treated as errors
```

src/core/device.c
keszybz: Yeah, this sounds really wrong.

> When this happens (due to PID1switching root), udev is inactive and its DB has been cleared

So this sounds like the issue is caused by this clearing of the database. I know this used to be done, but let's not try to hardcode this (IMHO completely broken) behaviour in even more places.

fbuihuu: I'm sorry but I don't get this. Can you please elaborate ?

src/core/device.c
fbuihuu: Again I'm not seeing which transition that would be needed in your example ? `plugged->plugged` ?

fbuihuu:
> ```
> 2022-05-19T07:47:33.3416115Z ../src/core/device.c: In function ‘device_init’:
> 2022-05-19T07:47:33.3417051Z ../src/core/device.c:95:17: error: unused variable ‘d’ [-Werror=unused-variable]
> 2022-05-19T07:47:33.3417572Z    95 |         Device *d = DEVICE(u);
> 2022-05-19T07:47:33.3417904Z       |                 ^
> 2022-05-19T07:47:33.3418267Z cc1: all warnings being treated as errors
> ```

That's strange because `d` is referenced by the 2 assertions just below the statement of `d`.

src/core/device.c
yuwata: With this change, if a device does NOT have persistent database, then the device is going to dead state here, and later possibly goes to plugged state on uevent. Hence, if the device is already mounted on somewhere, it will be once unmounted. Thus, the original issue is re-introduced.

If a device HAS persistent databse, then the up-to-date flags (and state) will be applied in device_catchup(). Hence, it is not necessary to downgrade the flag and state here. So, the condition added by the second commit is correct.

So, a correct way to fix #23429 is simply add the condition.

src/core/device.c
yuwata: We need to call device_set_state() at least once in device_coldplug() or device_catchup(), even there is no transition like `plugged -> plugged`.

src/core/device.c
yuwata: Consider the case that a device does NOT have persistent database, and it is mounted on somewhere.
With this PR, the device will enter the dead state, as `enumerated_found` does not have DEVICE_FOUND_UDEV flag when `device_coldplug()` is called. Hence, the corresponding mount unit will be stopped. After systemd-udevd processes the device, then the device unit again goes to the plugged state, and the mount unit is restarted.

This is mostly equivalent to the original issue, right?

fbuihuu: No.

If the device is mounted, then `d->enumerated_found=DEVICE_FOUND_MOUNT` and the device enters in "tentative state".

Then later udevd is started and the events are replayed. From this point, the device enters in "plugged" state through `tentative->plugged` transition, which doesn't stop the mount unit.

I think you should give it a try.

src/core/device.c
fbuihuu: Why ?

yuwata:
Superseded by #23517. Closing.

PR #23437, opened 2022-05-19
core/device: fix state transition on switching root with persistent database

### LINKED ISSUE #23429: PR#23218 introduced a regression for devices with `db_persist` option

yuwata:
Fixes #23429.

### DIFF
--- a/src/core/device.c
+++ b/src/core/device.c
@@ -208,7 +208,7 @@ static int device_coldplug(Unit *u) {
                 found &= ~DEVICE_FOUND_UDEV; /* ignore DEVICE_FOUND_UDEV bit */
                 if (state == DEVICE_PLUGGED)
                         state = DEVICE_TENTATIVE; /* downgrade state */
-                if (found == DEVICE_NOT_FOUND)
+                if (found == DEVICE_NOT_FOUND && d->enumerated_found == DEVICE_NOT_FOUND)
                         state = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
         }
 


src/core/device.c
```diff
@@ -190,18 +190,19 @@ static int device_coldplug(Unit *u) {
...
+         *   are called. Hence, let's conditionalize the operations by using the flag. After switch-root,
+         *   systemd-udevd will (re-)process all devices, and the Device.found and Device.state will be
+         *   adjusted.
          *
          * - On reload or reexecute, we can trust enumerated_found, deserialized_found, and deserialized_state.
          *   Of course, deserialized parameters may be outdated, but the unit state can be adjusted later by
          *   device_catchup() or uevents. */
 
-        if (MANAGER_IS_SWITCHING_ROOT(m)) {
+        if (MANAGER_IS_SWITCHING_ROOT(m) && !FLAGS_SET(d->enumerated_found, DEVICE_FOUND_UDEV)) {
```
fbuihuu, line 205: It looks similar to the condition I use in #23436, which fixes #23429 too, so I guess this should do the trick.

bluca: if you are able to reproduce the issue, can you confirm whether this fixes it for you?

fbuihuu: yep I'll give it a test.

yuwata: Thank you.

fbuihuu: I'm not seeing these "dead->plugged" transitions anymore right after switching root for devices with the "db_persist" option set.

mwilck: ~~I think this is wrong. We must reset the state to `DEVICE_TENTATIVE` to prevent premature device activation in the switching root case. The argument is the same as in #23208. Even with `db_persist`, the udev rules may have changed while switching root, and a device which was "plugged" in the initramfs may not be any more in the mounted root (SYSTEMD_READY="0" may be set on it). IMO `DEVICE_READY` should never be set after switching root until an actual "add" uevent for the device has been processed.~~

After looking at the dm-crypt example, I have to say this statement of mine was wrong. dm-crypt devices apparently need to be in "plugged" state early on. We can't wait for udev coldplug to establish this state, because `cryptsetup@,service` may need to be activated before udev after switching root, and that wouldn't happen if the device was only in tentative state.

mwilck (changes requested):

src/core/device.c
fbuihuu: Well if db_persist is used it's clearly an indication that the device must remain plugged, see the doc. If SYSTEMD_READY=0 is used in the host then it seems actually expected that the device goes through the `plugged->dead` transition when the new set of rules from the host is processed by udev.

mwilck: That's one possible intepretation of the text in **udev(7)**. But you're just inventing it :-)

`db_persist` dates back to 2011, before the systemd/udev merge. The man page paragraph was added much later. I don't think the relationship of `db_persist` and systemd's `state` attribute has ever been clearly defined. 

`db_persist` was created in the past for LVM, because LVM uses "change" events for device activation rather than "add" events, and needs special treatment for coldplug "add" events for devices that are already active (see `10-dm.rules`). LVM doesn't care about systemd's `state`, so as far as LVM is concerned, it doesn't matter whether we translate `db_persist` to `DEVICE_READY` or `DEVICE_TENTATIVE`.

I still think `DEVICE_TENTATIVE` is safer. It doesn't do harm, because the actual state will be resolved sooner or later either way, and it avoids premature activation.

yuwata: > I think this is wrong. We must reset the state to `DEVICE_TENTATIVE` to prevent premature device activation in the switching root case. The argument is the same as in #23208. Even with `db_persist`, the udev rules may have changed while switching root, and a device which was "plugged" in the initramfs may not be any more in the mounted root (SYSTEMD_READY="0" may be set on it). IMO `DEVICE_READY` should never be set after switching root until an actual "add" uevent for the device has been processed.

@mwilck The host PID1 has used and honored the persistent database generated in initrd at least more than two years. Why do you want to change the behavior? Of course, we have not documented the behavior, but IIRC, there is no issue related to that.
As reported in #23429, after PR #23218, the device with persistent database has unnecessary state transition. It may be harmless in most cases, but such state transition did not occur previously. If we can avoid that, why not to do that?

> Well if db_persist is used it's clearly an indication that the device must remain plugged, see the doc. If SYSTEMD_READY=0 is used in the host then it seems actually expected that the device goes through the `plugged->dead` transition when the new set of rules from the host is processed by udev.

@fbuihuu I agree with you. Even if the behavior is not documented, we have done so for a while. There is no reason to change it.

src/core/manager-serialize.c
```diff
@@ -397,15 +394,6 @@ int manager_deserialize(Manager *m, FILE *f, FDSet *fds) {
                         else
                                 m->service_watchdogs = b;
 
-                } else if ((val = startswith(l, "honor-device-enumeration="))) {
```
mwilck, line 400: Wouldn't this cause a warning if an older systemd version passed `honor-device-enumeration` in?

mwilck:
Whatever we eventually do here, can we please separate the `MANAGER_IS_SWITCHING_ROOT` change from the bug fix?

src/core/device.c
mwilck: > @mwilck The host PID1 has used and honored the persistent database generated in initrd at least more than two years. Why do you want to change the behavior?

@yuwata, you're right that as a matter of fact, systemd kept `DEVICE_READY` state over switching root. It did this for **all** devices though, which was wrong, and was fixed in #23218. #23218 did indeed "change the behavior". I hope we agree it did so for the better. 

The open question is how to deal with `db_persist`. We need to realize that we have two conflicting semantic rules:

 1. udev rules are allowed to be different between initrd and root fs, and thus need to be re-parsed and re-executed after switching root before switching device states and forwarding them to dependent units.
 2. `db_persist` devices keep state over switching root.

@fbuihuu and you think that rule 2) defines an exception to rule 1). I disagree. 1) is a general principle. **Nothing** prevents udev rules from changing between initrd and root FS. Thus if we run `device_set_state(DEVICE_READY)` on these devices, we'd open up new ways for the race described in #23208 to occur, which may cause boot failure.

`db_persist` is only used by dm and md. After switching root, the  udev rules for these subsystems need to distinguish between "activation" events and other events, which requires certain udev properties to survive the cleanup (see e.g. [this dracut commit](https://github.com/dracutdevs/dracut/commit/b874d4b5be4c3b81cf394726c354ef7131b5699d)). This distinction is made in udev rules and relies 100% on properties in the udev db. It has nothing to do with systemd's device activation. Thus for the current consumers of `db_persist`, it doesn't matter if we set the state to "tentative" rather than "plugged" here. 

It does matter for the further processing of dependent devices, though. The processing of dm and md devices during boot relies heavily on (udev) coldplug. It's not in the interest of these subsystems to have dependent units activated before the (udev) coldplug events have been processed.  Strictly speaking, after switching root, we should never call `device_set_state(DEVICE_READY)`, except from `device_dispatch_io()`.

**However**, I realize that this hunk is not the point. With this hunk, we'll set `DEVICE_READY` in `device_coldplug()` for `db_persist` devices, whereas without it, we set it in `device_catchup()`. If my argument above is correct, both calls would be premature. @fbuihuu, with this observation I don't understand [how this patch fixes your issue](https://github.com/systemd/systemd/pull/23437#discussion_r877134440). The extra "dead→plugged" transition should still observed, just a little bit earlier.

yuwata:
@mwilck Thank you for the investigation. Still I think this PR is OK, but let's postpone to v252.

src/core/manager-serialize.c
yuwata: Ouch. Thanks.

yuwata:
@mwilck 
https://github.com/systemd/systemd/pull/23218#issuecomment-1130503768
> @yuwata. recall [this discussion](https://github.com/systemd/systemd/pull/23218#discussion_r862654712) we had previously? If we removed the lines
> 
> ```
> +                if (found == DEVICE_NOT_FOUND)
> +                        state = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
> ```
> 
> from [75d7b59](https://github.com/systemd/systemd/commit/75d7b5989f99125e52d5c0e5656fa1cd0fae2405), we'd still see transition messages but only "tentative→plugged", not the confusing "dead→plugged". I don't think this can be completely avoided. We apply the result from enumeration after the result from deserialization on purpose.

https://github.com/systemd/systemd/pull/23437#discussion_r877306966
> I still think DEVICE_TENTATIVE is safer. It doesn't do harm, because the actual state will be resolved sooner or later either way, and it avoids premature activation.

I hope the last commit in this PR addresses that. WDYT?

mwilck:
> I hope the last commit in this PR addresses that. WDYT?

I guess so, but I haven't taken a closer look. I've tested these patches one-by-one.
One thing that I don't like so much about this PR is that it mixes `enumerated_found` into the `device_coldplug()` logic. Previouly, `device_coldplug()` would handle `deserialized_found` and either `device_enumerate` or `device_catchup()` would handle `enumerated_found`. Now, `enumerated_found` is part of the logic of both `coldplug` and `catchup`.

I've been wondering if it'd be possible (in the switching-root case at least) to postpone the call to `device_set_state()` , and perhaps the entire state calculation logic, to `device_catchup()`. Why do we need to call `device_set_state()` twice?

fbuihuu:
> I've been wondering if it'd be possible (in the switching-root case at least) to postpone the call to `device_set_state()` , and perhaps the entire state calculation logic, to `device_catchup()`. Why do we need to call `device_set_state()` twice?

This is basically what the last commit of https://github.com/systemd/systemd/pull/23436 is all about.

mwilck:
> Then issue #23429 is not a regression.

Well unfortunately it seems to be a regression for systems using dm-crypt for the root FS. Another example for the general theorem that whatever we do in this area, we overlook something :sob: 

I also have to revoke my [statement above](https://github.com/systemd/systemd/pull/23437#discussion_r877205840) that "tentative" is always better than "plugged". I still think that we have a conflict between two principles here, as stated [above](https://github.com/systemd/systemd/pull/23437#discussion_r877871523), but at least the way things are set up today, we have to carry over "plugged" state from the initrd for `db_persist` devices. I can think about other approaches for dm-crypt, but let's not overcomplicate matters at this stage.

src/core/manager-serialize.c
yuwata: Fixed.

yuwata:
Superseded by #23517. I will pick the cleanup commits by @fbuihuu in another PR. Closing.

PR #23441, opened 2022-05-19
Add sys/stat.h for S_IFDIR

kraj:
Fixes
../git/src/shared/mkdir-label.c:13:61: error: use of undeclared identifier 'S_IFDIR'
        r = mac_selinux_create_file_prepare_at(dirfd, path, S_IFDIR);

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/shared/mkdir-label.c
+++ b/src/shared/mkdir-label.c
@@ -1,5 +1,7 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
+#include <sys/stat.h>
+
 #include "mkdir-label.h"
 #include "selinux-util.h"
 #include "smack-util.h"


bluca (approved):

src/shared/mkdir-label.h
```diff
@@ -2,6 +2,7 @@
 #pragma once
 
 #include <fcntl.h>
+#include <sys/stat.h>
```
yuwata: Please include this in mkdir-label.c instead of .h file.

kraj: Done. thats how I had it but changed it since system headers were included in .h file.

src/shared/mkdir-label.c
```diff
@@ -4,6 +4,7 @@
 #include "selinux-util.h"
 #include "smack-util.h"
 #include "user-util.h"
+#include <sys/stat.h>
```
bluca: order in a new block above the internal headers plz

src/shared/mkdir-label.h
kraj: yeah done

src/shared/mkdir-label.c
kraj: done

yuwata (approved):

src/shared/mkdir-label.h
floppym: Just to explain: mode_t, uid_t, and gid_t are defined in fcntl.h and sys/types.h. Those types are referenced directly in mkdir-label.h.

The S_IFDIR macro is only referenced in mkdir-label.c; we don't need it in mkdir-label.h.

PR #23447, opened 2022-05-20
22826/templates for new units

amit-tewari:
Issue: https://github.com/systemd/systemd/issues/22826

Instead of adding a new switch, modify `edit` to check if a unit exists or not (If `--new` was added, it would have to check a unit exists anyway)

Unit template identification added in TODO comments. Need to be extended to actually use from a template when a new unit is being created.


amit-tewari:
manual tests done
```
ninja -C build
build/systemctl edit abc --user
build/systemctl edit abc --user
build/systemctl edit abc --user --full
build/systemctl edit abc --user --runtime

sudo build/systemctl edit abc
sudo build/systemctl edit abc
sudo build/systemctl edit abc --full

sudo build/systemctl edit abc --runtime
```

Issue #23448, opened 2022-05-20
portabled crashes in extension_release_validate when directories are passed to portablectl


evverx:
**systemd version the issue has been seen with**
https://github.com/systemd/systemd/commit/1ab191b3a95cd511e630e24555ed50205d1ed220

```sh
touch t2/usr/lib/extension-release.d/extension-release.t2
portablectl attach /root/t1 --extension=/root/t2
(Matching unit files with prefix 't1'.)
AttachImageWithExtensions failed: Remote peer disconnected
```
```sh
Assertion '!isempty(host_os_release_id)' failed at src/shared/extension-release.c:21, function extension_release_validate(). Aborting.

Program received signal SIGABRT, Aborted.
0x00007ffff4b9c59c in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install audit-libs-3.0.8-1.fc35.x86_64 glibc-2.34-32.fc35.x86_64 iptables-legacy-libs-1.8.7-13.fc35.x86_64 kmod-libs-29-4.fc35.x86_64 libacl-2.3.1-2.fc35.x86_64 libasan-11.3.1-2.fc35.x86_64 libattr-2.5.1-3.fc35.x86_64 libblkid-2.37.4-1.fc35.x86_64 libcap-2.48-3.fc35.x86_64 libcap-ng-0.8.2-8.fc35.x86_64 libeconf-0.4.0-2.fc35.x86_64 libffi-3.1-29.fc35.x86_64 libgcc-11.3.1-2.fc35.x86_64 libgcrypt-1.9.4-1.fc35.x86_64 libgpg-error-1.43-1.fc35.x86_64 libibverbs-39.0-1.fc35.x86_64 libmount-2.37.4-1.fc35.x86_64 libnl3-3.5.0-8.fc35.x86_64 libselinux-3.3-1.fc35.x86_64 libstdc++-11.3.1-2.fc35.x86_64 libubsan-11.3.1-2.fc35.x86_64 libxcrypt-4.4.28-1.fc35.x86_64 libzstd-1.5.2-1.fc35.x86_64 openssl-libs-1.1.1n-1.fc35.x86_64 pcre2-10.39-1.fc35.x86_64 xz-libs-5.2.5-9.fc35.x86_64 zlib-1.2.11-30.fc35.x86_64
(gdb) bt
#0  0x00007ffff4b9c59c in __pthread_kill_implementation () from /lib64/libc.so.6
[...]

Issue #23445, opened 2022-05-19
portabled can't be introspected via DBus


evverx:
**systemd version the issue has been seen with**
https://github.com/systemd/systemd/commit/1ab191b3a95cd511e630e24555ed50205d1ed220

The `/org/freedesktop/portable1/image` hierarchy can't be introspected and it prevents programs like `busctl`, `gdbus` and `dfuzzer` from discovering images automatically (which is the point of the introspection in general)
```sh
# busctl tree org.freedesktop.portable1 --no-pager
Failed to introspect object /org/freedesktop/portable1/image of service org.freedesktop.portable1: Expected image name, not path in place of ''.
└─/org
  └─/org/freedesktop
    ├─/org/freedesktop/LogControl1
    └─/org/freedesktop/portable1

# busctl introspect org.freedesktop.portable1 /org/freedesktop/portable1/image
Failed to introspect object /org/freedesktop/portable1/image of service org.freedesktop.portable1: Expected image name, not path in place of ''.
```

PR #23452, opened 2022-05-20
portable: reject root directories without an ID field in os-release

bluca:
Fixes https://github.com/systemd/systemd/issues/23448

### DIFF
--- a/src/portable/portable.c
+++ b/src/portable/portable.c
@@ -582,6 +582,8 @@ static int extract_image_and_extensions(
                                    "PORTABLE_PREFIXES", &prefixes);
                 if (r < 0)
                         return r;
+                if (isempty(id))
+                        return sd_bus_error_set_errnof(error, SYNTHETIC_ERRNO(ESTALE), "Image %s os-release metadata lacks the ID field", name_or_path);
 
                 if (prefixes) {
                         valid_prefixes = strv_split(prefixes, WHITESPACE);
--- a/test/units/testsuite-29.sh
+++ b/test/units/testsuite-29.sh
@@ -152,6 +152,16 @@ umount /tmp/rootdir
 umount /tmp/app0
 umount /tmp/app1
 
+# Lack of ID field in os-release should be rejected, but it caused a crash in the past instead
+mkdir -p /tmp/emptyroot/usr/lib
+mkdir -p /tmp/emptyext/usr/lib/extension-release.d
+touch /tmp/emptyroot/usr/lib/os-release
+touch /tmp/emptyext/usr/lib/extension-release.d/extension-release.emptyext
+
+# Remote peer disconnected -> portabled crashed
+res="$(! portablectl attach --extension /tmp/emptyext /tmp/emptyroot 2> >(grep "Remote peer disconnected"))"
+test -z "${res}"
+
 echo OK >/testok
 
 exit 0


evverx:
> Fixes https://github.com/systemd/systemd/issues/23445

Looks like this PR is supposed to fix https://github.com/systemd/systemd/issues/23448. The introspection is still broken so https://github.com/systemd/systemd/issues/23445 isn't fixed.

src/portable/portable.c
```diff
@@ -582,6 +582,8 @@ static int extract_image_and_extensions(
                                    "PORTABLE_PREFIXES", &prefixes);
                 if (r < 0)
                         return r;
+                if (isempty(id))
+                        return sd_bus_error_set_errnof(error, SYNTHETIC_ERRNO(ESTALE), "Image %s os-release metadata lacks the ID field", name_or_path);
```
evverx, line 586: According to https://www.freedesktop.org/software/systemd/man/os-release.html#ID=
> If not set, a default of "ID=linux" may be used

I'm not sure why portabled would start rejecting images without IDs. It has always allowed attaching images like that.

test/units/testsuite-29.sh
```diff
@@ -152,6 +152,18 @@ umount /tmp/rootdir
...
 umount /tmp/app1
 
+# Lack of ID field in os-release should be rejected, but it caused a crash in the past instead
+mkdir -p /tmp/emptyroot/usr/lib
+mkdir -p /tmp/emptyext/usr/lib/extension-release.d
+touch /tmp/emptyroot/usr/lib/os-release
+touch /tmp/emptyext/usr/lib/extension-release.d/extension-release.emptyext
+
+set +e
+res=$(portablectl attach --extension /tmp/emptyext /tmp/emptyroot)
```
medhefgo: With `res=$(! portablectl ...)` you don't need the `set +e`.

src/portable/portable.c
keszybz: Yeah, I don't think we should start becoming more strict here.

test/units/testsuite-29.sh
bluca: Thanks, done

src/portable/portable.c
bluca: > Yeah, I don't think we should start becoming more strict here.

The spec is quite clear that an os-release is required https://systemd.io/PORTABLE_SERVICES/ and the os-release spec itself clearly says `This field is optional.` when a field is not mandatory. It doesn't say so for `ID`. I'm not sure an os-release without at least ID can be claimed to be well-formed? Is there any distro actually doing that?

keszybz: OK, for portable images that is true. So I think we can merge this, even if doesn't fix all the issues.

evverx: It's a breaking change nonetheless so I think it should be mentioned in `NEWS`

bluca: > OK, for portable images that is true. So I think we can merge this, even if doesn't fix all the issues.

Let's put it in v252 - it's really not urgent, it can wait till later (hopefully tonight?)

keszybz:
Nah, it fixes a crash.

bluca:
> Nah, it fixes a crash.

I'm not very convinced it's that important... it's an artificial corner case that doesn't exist in the real world, for an exit-on-idle service, IMHO it can wait

evverx:
By the way I agree it's a corner case but looking at the "extensions" feature a bit closer it appears it wasn't integrated into portabled smoothly. Apart from crashes and memory leaks the "consistency" checks are basically pointless because it isn't that hard to end up having service files belonging to images that have been detached.

keszybz:
It's a corner case, but still, we shouldn't crash. The fix is fairly obvious, let's merge it.

PR #23453, opened 2022-05-20
Add strv methods with externally-supplied size argument and speed up parsing of /etc/hosts

keszybz:


src/basic/strv.h
```diff
@@ -35,18 +35,32 @@ size_t strv_length(char * const *l) _pure_;
 int strv_extend_strv(char ***a, char * const *b, bool filter_duplicates);
 int strv_extend_strv_concat(char ***a, char * const *b, const char *suffix);
 int strv_prepend(char ***l, const char *value);
-int strv_extend(char ***l, const char *value);
+
+int strv_extend_with_size(char ***l, size_t *n, const char *value);
```
bluca, line 43: Can you add a brief comment here, replicating the line about speed? Otherwise it won't be super obvious to someone else some time in the future

src/basic/strv.c
```diff
@@ -403,27 +403,34 @@ char* strv_join_full(char * const *l, const char *separator, const char *prefix,
...
+        /* n is a pointer to a variable to store the size of l.
+         * If not given, n will be calculated using strv_length().
+         * If given, the size after the push will be returned. */
 
         if (!value)
                 return 0;
 
-        n = strv_length(*l);
+        if (size == SIZE_MAX)
+                size = strv_length(*l);
```
medhefgo, line 417: Moving the declaration of size down here would simply things a bit...

keszybz:
Updated with the two suggestions.

bluca (approved):

keszybz:
This is not urgent, I think we shouldn't merge it before the release.

yuwata (approved):

PR #23449, opened 2022-05-20
shared/extension-release: fix assertion

yuwata:
Fixes #23445.
Fixes #23448.

PR #23454, opened 2022-05-20
Fix bus introspection of portable1

keszybz:
This is on top of https://github.com/systemd/systemd/pull/23452. Once that's merged, it should be fine to merge this one too, the shared commits don't matter. I took one commit from https://github.com/systemd/systemd/pull/23449.

### DIFF
--- a/src/portable/portabled-image-bus.c
+++ b/src/portable/portabled-image-bus.c
@@ -1045,19 +1045,23 @@ int bus_image_acquire(
                 /* If it's a short name, let's search for it */
                 r = image_find(IMAGE_PORTABLE, name_or_path, NULL, &loaded);
                 if (r == -ENOENT)
-                        return sd_bus_error_setf(error, BUS_ERROR_NO_SUCH_PORTABLE_IMAGE, "No image '%s' found.", name_or_path);
+                        return sd_bus_error_setf(error, BUS_ERROR_NO_SUCH_PORTABLE_IMAGE,
+                                                 "No image '%s' found.", name_or_path);
 
                 /* other errors are handled below… */
         } else {
                 /* Don't accept path if this is always forbidden */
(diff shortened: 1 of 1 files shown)


evverx:
It would be great if it was possible to add `busctl tree org.freedesktop.portable1 --no-pager` to the portabled testsuite. Its DBus interface is generally broken but at least the test would help to keep it more or less sane. On the bright side even `dfuzzer` has never seen services where the introspection would be that broken so in a way `portabled` made `dfuzzer` better: https://github.com/dbus-fuzzer/dfuzzer/pull/100

bluca (approved):

yuwata (approved):

yuwata:
> It would be great if it was possible to add `busctl tree org.freedesktop.portable1 --no-pager` to the portabled testsuite. Its DBus interface is generally broken but at least the test would help to keep it more or less sane. On the bright side even `dfuzzer` has never seen services where the introspection would be that broken so in a way `portabled` made `dfuzzer` better: [dbus-fuzzer/dfuzzer#100](https://github.com/dbus-fuzzer/dfuzzer/pull/100)

Let's do that later.

src/shared/dissect-image.c
```diff
@@ -3157,7 +3157,7 @@ int verity_dissect_and_mount(
          * First, check the distro ID. If that matches, then check the new SYSEXT_LEVEL value if
          * available, or else fallback to VERSION_ID. If neither is present (eg: rolling release),
          * then a simple match on the ID will be performed. */
-        if (required_host_os_release_id) {
+        if (!isempty(required_host_os_release_id)) {
```
poettering, line 3160: hmm, this looks wrong. We should have filtered this much earlier, not here. I think we should revert this patch and figure out where we missed suppressing the ID earlier.

PR #23455, opened 2022-05-20
sd-bus: do not return empty string as success in sd_bus_path_decode()

bluca:
3970	        e = object_path_startswith(path, prefix);
(gdb) p path
$1 = 0x55c5a166f768 "/org/freedesktop/portable1/image"
(gdb) p prefix
$2 = 0x55c59ffc2928 "/org/freedesktop/portable1/image"
(gdb) p e
$1 = 0x5581a1675788 ""

This happens when calling 'busctl tree org.freedesktop.portable1'

Fixes https://github.com/systemd/systemd/issues/23445

I am not 100% sure it's safe to change as such, given it's a public api... but it does seem strange to return success and an empty string? It could be worked around in the callers instead

### DIFF
--- a/src/libsystemd/sd-bus/sd-bus.c
+++ b/src/libsystemd/sd-bus/sd-bus.c
@@ -3968,7 +3968,7 @@ _public_ int sd_bus_path_decode(const char *path, const char *prefix, char **ext
         assert_return(external_id, -EINVAL);
 
         e = object_path_startswith(path, prefix);
-        if (!e) {
+        if (isempty(e)) {
                 *external_id = NULL;
                 return 0;
         }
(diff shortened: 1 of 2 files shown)


keszybz:
No, I don't think this is right. An implementation *may* respond to the path. E.g. imagine a case where the subtree corresponds to a subtree on disk, and you want to return information about the file system. In that case, returning something meaning for the root path would be useful. In our cases we don't do this, but I think this decision should be taken individually by the bus_*_find() functions. See the commit message in  #23454.

bluca:
> No, I don't think this is right. An implementation _may_ respond to the path. E.g. imagine a case where the subtree corresponds to a subtree on disk, and you want to return information about the file system. In that case, returning something meaning for the root path would be useful. In our cases we don't do this, but I think this decision should be taken individually by the bus_*_find() functions. See the commit message in #23454.

Yeah, the use case makes sense. I found the behaviour a bit surprising, I'll add a test and a comment.

bluca:
> > No, I don't think this is right. An implementation _may_ respond to the path. E.g. imagine a case where the subtree corresponds to a subtree on disk, and you want to return information about the file system. In that case, returning something meaning for the root path would be useful. In our cases we don't do this, but I think this decision should be taken individually by the bus_*_find() functions. See the commit message in #23454.
> 
> Yeah, the use case makes sense. I found the behaviour a bit surprising, I'll add a test and a comment.

https://github.com/systemd/systemd/pull/23456

PR #23460, opened 2022-05-21
docs: rename COREDUMP_PACKAGE_METADATA → ELF_PACKAGE_METADATA

keszybz:
The format described by this document is used not only for coredumps,
but also for other purposes, so we've outgrown the old title.

A redirect was added based on https://codepo8.github.io/github-redirection-demo/.
I tried to use a relative link, but it's hard to test if it works
before merging.

@bnf maybe you know if the redirect is done correctly?

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <!-- SPDX-License-Identifier: LGPL-2.1-or-later -->
 
-<html>
+<html lang="en">
 
   {% include head.html %}
 
--- a/docs/_layouts/forward.html
+++ b/docs/_layouts/forward.html
@@ -1,14 +1,26 @@
+<!DOCTYPE html>
+<!-- SPDX-License-Identifier: LGPL-2.1-or-later -->
+
 <html lang="en">
-  <!-- SPDX-License-Identifier: LGPL-2.1-or-later -->
   <head>
+    <meta charset="utf-8">
     <meta http-equiv="refresh" content="0;url={{ page.target }}"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta name="theme-color" content="#201A26">
+    <meta name="robots" content="noindex,follow">
+    <link rel="stylesheet" href="{{ "/style.css" | prepend: site.baseurl }}">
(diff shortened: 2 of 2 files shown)


bluca (approved):

bnf:
@keszybz The `forward.html` layout file is missing. That files needs to includes the `<meta http-equiv="refresh" content=""` tag to interact as workaround for missing github-pages server-side redirects.

You can add it like so: https://github.com/bnf/systemd/commit/9abb73fe52f70fd0f919902b07a4565194a15530
(part of my https://github.com/bnf/systemd/commits/docs branch)

Also the `target` needs to be absolute, otherwise `/COREDUMP_PACKAGE_METADATA/` would redirect to `/COREDUMP_PACKAGE_METADATA/ELF_PACKAGE_METADATA`.
See: https://github.com/bnf/systemd/commit/397652a70ab62ed109662813a6624032db4f2815

With those fixes the pseudo-redirect will work: https://systemd.bnf.dev/COREDUMP_PACKAGE_METADATA/

docs/COREDUMP_PACKAGE_METADATA.md
```diff
@@ -1,149 +1,4 @@
 ---
-title: Package Metadata for Core Files
-category: Interfaces
-layout: default
-SPDX-License-Identifier: LGPL-2.1-or-later
+layout: forward
```
bnf, line 2: forward.html missing in docs/_layouts/forward.html

docs/COREDUMP_PACKAGE_METADATA.md
```diff
@@ -1,149 +1,4 @@
...
+target: ELF_PACKAGE_METADATA
```
bnf: target needs to be absolute, `/ELF_PACKAGE_METADATA`, to avoid a calculated destination like:  `/COREDUMP_PACKAGE_METADATA/ELF_PACKAGE_METADATA`

bnf (changes requested):

keszybz:
Thanks. I merged a stub forward.html layout into the commit. We don't need the whole thing with formatting, we're not going to use it.

bnf:
Thanks, we need one fix though, maybe you can squash https://github.com/keszybz/systemd/pull/6 into it, then it's done.

keszybz:
CI doesn't care, let's assume that this works.

PR #23466, opened 2022-05-21
docs: improve table readability

bnf:
Let the table span more than the default content width,
if the table contains alot of data (controlled by width: auto)
(720px is very good for continuous text, but too narrow for tables).

The container class is therefore adapted to put the
width restriction on the elements itself, allowing for
exceptions for individual elements like `<table>` and
`<h1>` (which used an offset margin before and is now
streamlined to use a max-width as well).

Also add a striped background to ease reading rows
and allow for horizontal mobile scrolling without
overflowing the entire document, only the table itself.

## Current View
https://systemd.io/PORTABILITY_AND_STABILITY#general-portability-of-systemds-apis

current-tables-light

## Updated View

https://systemd.bnf.dev/PORTABILITY_AND_STABILITY#general-portability-of-systemds-apis

wide-tables-light

wide-tables-dark

### Support for narrow tables without rendering them too wide

https://systemd.bnf.dev/ELF_PACKAGE_METADATA/#well-known-keys

narrow-tables-light

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -43,6 +43,8 @@
   --sd-highlight-bg: var(--sd-highlight-bg-light);
   --sd-highlight-inline-bg: var(--sd-highlight-inline-bg-light);
   --sd-link-font-weight: var(--sd-font-weight-bold);
+  --sd-table-row-bg: var(--sd-highlight-inline-bg-light);
+  --sd-table-row-hover-bg: var(--sd-gray);
 }
 
 @media (prefers-color-scheme: dark) {
@@ -56,6 +58,8 @@
     --sd-highlight-bg: var(--sd-highlight-bg-dark);
     --sd-highlight-inline-bg: var(--sd-highlight-inline-bg-dark);
     --sd-link-font-weight: var(--sd-font-weight-normal);
+    --sd-table-row-bg: var(--sd-highlight-inline-bg-dark);
+    --sd-table-row-hover-bg: var(--sd-highlight-bg-dark);
   }
 }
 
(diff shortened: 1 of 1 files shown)


medhefgo:
The first example looks/feels a bit weird when you are a boomer and have a default zoom level (125% in my case). The lack of a margin around the table feels really oppressing.

bnf:
I agree, that looks weird on small desktops/zoom-mode.
Force pushed an adaption.

medhefgo (approved):
Looks really nice. Something I just thought of: maybe add a mouse hover for the current row? Would make it even easier to read things…

bnf:
> Something I just thought of: maybe add a mouse hover for the current row? Would make it even easier to read things…

I personally don't need it, but fine by me to add it.
Forced pushed the change to add that highlighting.

bluca (approved):

yuwata (approved):

medhefgo:
@bnf Since you seem to be a web guy, any chance I can interest you in beefing up the journal web ui in `src/journal-remote`?

medhefgo:
Nice! One thing: it probably makes sense to split this into separately served files (css/js/html maybe new things). Should be easy to do: https://github.com/systemd/systemd/blob/35207b14cf2bdd4b80742abbe3852b23aa6a3559/src/journal-remote/journal-gatewayd.c#L845 https://github.com/systemd/systemd/blob/35207b14cf2bdd4b80742abbe3852b23aa6a3559/src/journal-remote/meson.build#L63

PR #23473, opened 2022-05-22
journalctl: add --truncate-newline option

zhmylove:
Part of #23086

### Truncate by newlines
Sometimes log messages contain several lines of text. The most inconvenient are java backtraces among other useful logs. Unfortunately `journalctl(1)` lacks any option to output log messages only until the end of the first line to suppress any garbage in the message. This PR introduces `--truncate-newline` option which does exactly this: truncate messages by the first newline character, if any. 
Example:
```
# ./build/journalctl -t test
Apr 15 03:03:53 rc1 test[31481]: Error1
                                 Details...
                                 Details...
                                 Details...
Apr 15 03:03:59 rc1 test[31482]: Error2
                                 Details...
                                 Details...
                                 Details...
Apr 15 03:04:03 rc1 test[31486]: Error3
                                 Details...
                                 Details...
                                 Details...
# ./build/journalctl -t test --truncate
Apr 15 03:03:53 rc1 test[31481]: Error1
Apr 15 03:03:59 rc1 test[31482]: Error2
Apr 15 03:04:03 rc1 test[31486]: Error3
```

### DIFF
--- a/man/journalctl.xml
+++ b/man/journalctl.xml
@@ -430,6 +430,13 @@
           </varlistentry>
 
           <varlistentry>
+            <term><option>--truncate-newline</option></term>
+
+            <listitem><para>Truncate each log message at the first newline character on output, so that only the
+            first line of each message is displayed.</para></listitem>
+          </varlistentry>
+
+          <varlistentry>
             <term><option>short-precise</option></term>
             <listitem><para>is very similar, but shows classic syslog timestamps with full microsecond
             precision.</para></listitem>
--- a/src/journal/journalctl.c
+++ b/src/journal/journalctl.c
@@ -96,6 +96,7 @@ static bool arg_all = false;
 static PagerFlags arg_pager_flags = 0;
 static int arg_lines = ARG_LINES_DEFAULT;
 static bool arg_no_tail = false;
+static bool arg_truncate_newline = false;
 static bool arg_quiet = false;
 static bool arg_merge = false;
 static bool arg_boot = false;
@@ -367,6 +368,7 @@ static int help(void) {
                "  -a --all                   Show all fields, including long and unprintable\n"
                "  -f --follow                Follow the journal\n"
                "     --no-tail               Show all lines, even in follow mode\n"
+               "     --truncate-newline      Truncate entries by first newline character\n"
                "  -q --quiet                 Do not show info messages and privilege warning\n"
                "\n%3$sPager Control Options:%4$s\n"
                "     --no-pager              Do not pipe output into a pager\n"
@@ -445,6 +447,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_RELINQUISH_VAR,
                 ARG_SMART_RELINQUISH_VAR,
                 ARG_ROTATE,
+                ARG_TRUNCATE_NEWLINE,
                 ARG_VACUUM_SIZE,
                 ARG_VACUUM_FILES,
                 ARG_VACUUM_TIME,
@@ -465,6 +468,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "full",                 no_argument,       NULL, 'l'                      },
                 { "no-full",              no_argument,       NULL, ARG_NO_FULL              },
                 { "lines",                optional_argument, NULL, 'n'                      },
+                { "truncate-newline",     no_argument,       NULL, ARG_TRUNCATE_NEWLINE     },
                 { "no-tail",              no_argument,       NULL, ARG_NO_TAIL              },
                 { "new-id128",            no_argument,       NULL, ARG_NEW_ID128            }, /* deprecated */
                 { "quiet",                no_argument,       NULL, 'q'                      },
@@ -620,6 +624,10 @@ static int parse_argv(int argc, char *argv[]) {
 
                 case ARG_NO_TAIL:
                         arg_no_tail = true;
+                        break;
+
+                case ARG_TRUNCATE_NEWLINE:
+                        arg_truncate_newline = true;
                         break;
 
                 case ARG_NEW_ID128:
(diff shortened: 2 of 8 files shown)


DaanDeMeyer (changes requested):
LGTM, but should we really add a short option for this? --strip seems short enough that we don't benefit too much from adding a shorthand, and it leaves -s open for another option that we might add in the future.

Hmm, maybe --strip isn't the right name for this, what about --no-newlines or --no-line-break?

dnicolodi:
> Hmm, maybe --strip isn't the right name for this, what about --no-newlines or --no-line-break?

The `--strip` option name is confusing enough that this is not what the options does. The options truncates the message at the first newline character. Maybe a better approach would be to c-escape newlines and call the option `--escape-newlines` or to rename the option `--first-line-only` or `--first-line` or something that is not `--strip`.

zhmylove:
@DaanDeMeyer I'm hesitating between `--strip` and `--truncate`, but the first is more shorter. 
From my perspective, `--no-newlines` and the other look more like `s/\n//` which is not the case.

vcaputo:
I'm not sure if this really fits, but GNU `ls` has precedent of using `-1` to limit output to one file per line.  If I squint hard enough I can see how that could be analogous to one line per message in `journalctl`, which doesn't already use `-1`.

zhmylove:
@DaanDeMeyer taking all the above into account, I've pushed a new version

dnicolodi:
I think the option name could still be improved. Using `--truncate` leaves still unclear where the truncation is going to take place. The most general solution would be to have the `--truncate` option to take an argument: `--truncate=newline` to truncate to the first newline, or `--truncate=N` to truncate at `N` characters.

zhmylove:
@dnicolodi we do already have `truncate_nl` and I can't imagine a situation where other separator is needed.

dnicolodi:
What I am trying to say is that the option name as spelled now is not a nice user interface as the user is left wondering where the truncation will happen. The argument to `--truncate` in my proposal is either `newline` or a number of characters. I am not proposing to use any other separator. Another option is to rename the option to something that makes it clear that the truncation happens at the first newline.

man/journalctl.xml
```diff
@@ -408,6 +408,12 @@
         printed.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--truncate</option></term>
```
poettering: maybe call this `--truncate-newline` or even `--truncate-at-newline`, to make clear where it truncates.

poettering (changes requested):
lgtm, but i think the name should make clear what it truncates, see comment.

poettering:
hmm, i wonder if it wouldn't be better to suffix the truncated lines with something to tell the user about this. possibly a pretty unicode char of some form via the special_glyph(). maybe one of these ellipsis chars?

https://en.wikipedia.org/wiki/Ellipsis#Computer_representations

or maybe the bottom right corner arrow? this one? ⇲

That way, we can communicate in a friendly way that a truncation took place

poettering:
And to me this doesn't sound like something we should add a short switch for, it's not a thing one would do all the time? or would one? dunno, if there's strong desire we can also give a short switch...

man/journalctl.xml
zhmylove: Thanks, I'll send a new commit soon considering all you comments

zhmylove:
> hmm, i wonder if it wouldn't be better to suffix the truncated lines with something to tell the user about this. possibly a pretty unicode char of some form via the special_glyph(). maybe one of these ellipsis chars?
> 
> https://en.wikipedia.org/wiki/Ellipsis#Computer_representations
> 
> or maybe the bottom right corner arrow? this one? ⇲
> 
> That way, we can communicate in a friendly way that a truncation took place

I'm not sure what is the proper way to do it (and if this really required). Pass some flag to `print_multiline()`?

src/journal/journalctl.c
```diff
@@ -435,6 +437,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_RELINQUISH_VAR,
                 ARG_SMART_RELINQUISH_VAR,
                 ARG_ROTATE,
+                ARG_TRUNCATE,
```
poettering: call this the same way as the exported option, i.e. `ARG_TRUNCATE_NEWLINE`

src/journal/journalctl.c
```diff
@@ -89,6 +89,7 @@ static bool arg_all = false;
 static PagerFlags arg_pager_flags = 0;
 static int arg_lines = ARG_LINES_DEFAULT;
 static bool arg_no_tail = false;
+static bool arg_truncate = false;
```
poettering: rename to arg_truncate_newline, i.e. please stick to the same name all throught the codebase

src/shared/output-mode.h
```diff
@@ -42,11 +42,12 @@ typedef enum OutputFlags {
         OUTPUT_BEGIN_NEWLINE  = 1 << 5,
         OUTPUT_UTC            = 1 << 6,
         OUTPUT_NO_HOSTNAME    = 1 << 7,
+        OUTPUT_TRUNCATE       = 1 << 8,
```
poettering: here too, call this OUTPUT_TRUNCATE_NEWLINE

man/journalctl.xml
```diff
@@ -407,6 +407,12 @@
             precision.</para></listitem>
           </varlistentry>
 
+          <varlistentry>
+            <term><option>--truncate-newline</option></term>
+
+            <listitem><para>Truncate each message so that only the first line is displayed.</para></listitem>
```
poettering: maybe say "Truncate each log message at the first newline character on output, so that only the first line of each message is displayed."

poettering (changes requested):
sorry for dropping the ball on this.

PR looks actually quite good. just some nitpicking about names.

i think we can merge this quickly if you clean up the text/naming as requested

poettering:
(the continuation glyph thing can be done another time)

zhmylove:
@poettering thank you for the feedback!
I've updated the PR, please review

poettering:
thanks!

PR #23474, opened 2022-05-22
systemctl: make show/status honour --state and --type

### LINKED ISSUE #23207: systemctl status --failed, please?
To show more details about failed units and to avoid typing lengthy service names, something like 
```
systemctl status --failed
```
would be very nice. Currently (250.4) the "--failed" appears to be ignored (without usage hint).

Just a suggestion, of course.

keszybz:
This makes the interface more flexible, by allowing the same filtering
for show and status as is done for list-units.

Fixes #23207.

### DIFF
--- a/src/systemctl/systemctl-show.c
+++ b/src/systemctl/systemctl-show.c
@@ -2070,6 +2070,7 @@ static int get_unit_dbus_path_by_pid(
 
 static int show_all(
                 sd_bus *bus,
+                SystemctlShowMode show_mode,
                 bool *new_line,
                 bool *ellipsized) {
 
@@ -2095,7 +2096,7 @@ static int show_all(
                 if (!p)
                         return log_oom();
 
-                r = show_one(bus, p, u->id, SYSTEMCTL_SHOW_STATUS, new_line, ellipsized);
+                r = show_one(bus, p, u->id, show_mode, new_line, ellipsized);
                 if (r < 0)
                         return r;
                 if (r > 0 && ret == 0)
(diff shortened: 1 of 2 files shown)


bluca:
Do the shell completions need a tweak for this?

keszybz:
> Do the shell completions need a tweak for this?

I don't think so, because it's just extending coverage of existing options to new verbs. I don't think any of our completions are smart enough to know which verbs care about which options.

src/systemctl/systemctl-show.c
```diff
@@ -2196,17 +2197,28 @@ int verb_show(int argc, char *argv[], void *userdata) {
...
+                        if (show_mode == SYSTEMCTL_SHOW_PROPERTIES)
+                                r = show_one(bus, "/org/freedesktop/systemd1", NULL, show_mode, &new_line, &ellipsized);
+                        else
+                                r = show_system_status(bus);
+                        if (r < 0)
+                                return r;
 
-                show_system_status(bus);
-                new_line = true;
+                        new_line = true;
```
daandemeyer, line 2217: Is it intended that we also always show a newline when we're following the `SYSTEMCTL_SHOW_PROPERTIES` logic path now? `show_one` seems to set `new_line` itself so it's weird that we just override it here now.

bluca (approved):

src/systemctl/systemctl-show.c
keszybz: Yes. The newline separates the system status from the per-unit statuses. (And the per-unit statuses are also separated by an empty line.)

keszybz:
?

yuwata:
Maybe better to add a test?

keszybz:
In principle I'd agree, but we have exactly 0 tests for `--failed` or `list-units`, except for incidental use in other tests. We should add a full test for systemctl that at least verifies the main aspects of operation… But I don't want to be on the hook for that for this relatively simple pull request.

yuwata:
OK. Let's add tests later in another PR.

Issue #23374, opened 2022-05-14
Bashism in `/usr/lib/kernel/install.d/90-loaderentry.install`


thesamesam:
**systemd version the issue has been seen with**
 > 251-rc2

**Used distribution**
 > Gentoo

**Linux kernel version used** (`uname -a`)

 > `5.15.39-gentoo-dist-hardened`

**CPU architecture issue was seen on**
 > amd64

**Expected behaviour you didn't see**
 > No errors when calling `installkernel` (indirectly here, through `eclean-kernel` in Gentoo).

**Unexpected behaviour you saw**
 > Error message during run: `/usr/lib/kernel/install.d/90-loaderentry.install: 21: shopt: not found`

**Steps to reproduce the problem**
 > Set `/bin/sh` to be provided by a POSIX-compliant shell like `dash` (definitely not `bash`)
> Try to run `eclean-kernel` in Gentoo (I suspect `installkernel` with appropriate args would work too, but the issue is obvious anyway)
> Get a Bashism error:
```
/usr/lib/kernel/install.d/90-loaderentry.install: 21: shopt: not found
```

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
# eclean-kernel -a
Preserving currently running kernel (5.15.39-gentoo-dist-hardened)
Legend:
[-] file being removed
[x] file does not exist (anymore)
[+] file being kept (used by other kernels)

* Removing kernel 5.15.38-gentoo-dist-hardened (not referenced by bootloader (symlinks))
/usr/lib/kernel/install.d/90-loaderentry.install: 21: shopt: not found
 [-] /boot/8c595ddc91b6b00a140266195f3f00ea/5.15.38-gentoo-dist-hardened/initrd
 [-] /lib/modules/5.15.38-gentoo-dist-hardened
 [-] /boot/8c595ddc91b6b00a140266195f3f00ea/5.15.38-gentoo-dist-hardened/linux
 [-] /boot/8c595ddc91b6b00a140266195f3f00ea/5.15.38-gentoo-dist-hardened
Removed 1 kernels
```

thesamesam:
Introduced by 367165a4069ac0c04882a05a8a80f6afb1e42760 (cc @keszybz).

The gist is that it's okay to use Bashisms like `shopt`, but if we're going to, the shebang needs to be for Bash. In 662f45e3ea9f6e933234b81bec532d584bda6ead, @nabijaczleweli ported it to use `/bin/sh` and avoid Bashisms.

[...]

PR #23475, opened 2022-05-22
Actually run shellcheck on CI

nabijaczleweli:
#23374

### DIFF
--- a/src/kernel-install/kernel-install.in
+++ b/src/kernel-install/kernel-install.in
@@ -112,6 +112,7 @@ fi
 
 if [ -n "$install_conf" ]; then
     [ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] && echo "Reading $install_conf…"
+    # shellcheck source=/dev/null
     . "$install_conf"
     # FIXME: This may override configuration in environment variables, e.g. $BOOT_ROOT.
 fi
@@ -130,6 +131,7 @@ fi
 # generated one. If the user configured an explicit machine ID to use in
 # /etc/machine-info to use for our purpose, we'll use that instead (for
 # compatibility).
+# shellcheck source=/dev/null
 if [ -z "$MACHINE_ID" ] && [ -r /etc/machine-info ] && . /etc/machine-info && MACHINE_ID="$KERNEL_INSTALL_MACHINE_ID"; then
     [ -n "$MACHINE_ID" ] && [ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] && \
         echo "machine-id $MACHINE_ID acquired from /etc/machine-info"
@@ -157,6 +159,7 @@ if [ -z "$ENTRY_TOKEN" ]; then
     # the IMAGE_ID= and ID= fields from /etc/os-release and finally the fixed
     # string "Default"
     ENTRY_TOKEN_SEARCH="$MACHINE_ID"
+    # shellcheck source=/dev/null
     [ -r /etc/os-release ] && . /etc/os-release
     [ -n "$IMAGE_ID" ] && ENTRY_TOKEN_SEARCH="$ENTRY_TOKEN_SEARCH $IMAGE_ID"
     [ -n "$ID" ] && ENTRY_TOKEN_SEARCH="$ENTRY_TOKEN_SEARCH $ID"
@@ -291,7 +294,7 @@ PLUGINS="$(
 IFS="
 "
 
-[ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] && echo -e "Plugin files:\n$PLUGINS"
+[ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] && printf '%s\n' "Plugin files:" "$PLUGINS"
 
 case "$COMMAND" in
     add)


mrc0mmand (changes requested):
I don't think this is correct. Answering the comment from https://github.com/systemd/systemd/issues/23374#issuecomment-1133902874 there are two factors that might be culprits here:

1) We do only differential shellcheck in each PR (`VALIDATE_ALL_CODEBASE=true`), i.e. we check only files changed by that PR

2) The "miss" in https://github.com/systemd/systemd/issues/23374#issuecomment-1133902874 was most ilkely caused by the  `FILTER_REGEX_EXCLUDE: .*/(man/.*|.*\.(in|SKELETON)|tools/coverity\.sh)$` regex, which ignores the template files (`*.in`) by default, since they are not usually compatible with shellcheck before they're processed by a templating engine

nabijaczleweli:
In workable state now, IMO

.github/workflows/linter.yml
```diff
@@ -37,14 +37,16 @@ jobs:
...
-          #        ubuntu-22.04 as a valid runner specification. This should
-          #        be fixed once https://github.com/github/super-linter/pull/2897
-          #        is merged, which includes
-          #        https://github.com/rhysd/actionlint/commit/09a60b0f5761c2d066def294e541c25a89756eea
+          #    the current version of super-linter doesn't recognize
+          #    ubuntu-22.04 as a valid runner specification. This should
+          #    be fixed once https://github.com/github/super-linter/pull/2897
+          #    is merged, which includes
+          #    https://github.com/rhysd/actionlint/commit/
+          #            09a60b0f5761c2d066def294e541c25a89756eea
```
keszybz: Please drop this chunk.

.github/workflows/linter.yml
```diff
@@ -37,14 +37,16 @@ jobs:
           #   - man/.* - all snippets in man pages (false positives due to
           #              missing shebangs)
           #   - .*\.(in|SKELETON) - all template/skeleton files
+          #                         except kernel-install
           #   - tools/coverity\.sh - external file (with some modifications)
-          FILTER_REGEX_EXCLUDE: .*/(man/.*|.*\.(in|SKELETON)|tools/coverity\.sh)$
+          FILTER_REGEX_EXCLUDE: .*/(man/.*|([^k]|k(k|ek)*([^ek]|e[^kr]))*(k(k|ek)*e?)?\.(in|SKELETON)|tools/coverity\.sh)$
```
keszybz, line 42: This is scary ;) Can we maybe use a negative match here:
`…|(?<!kernel-install)\.(in|SKELETON)|…` ?
Don't know if this is supported by this engine.

.github/workflows/linter.yml
```diff
@@ -37,14 +37,16 @@ jobs:
...
           VALIDATE_ALL_CODEBASE: false
-          VALIDATE_BASH: true
```
mrc0mmand, line 43: Why this change? The validators default to `false` when at least one of them is set to `true`, so the original `VALIDATE_BASH: true` should work just fine.

.github/workflows/linter.yml
```diff
@@ -37,14 +37,16 @@ jobs:
...
+          VALIDATE_SHELL_SHFMT: false
           # FIXME: temporarily disable validation of GH Action files, since
-          #        the current version of super-linter doesn't recognize
-          #        ubuntu-22.04 as a valid runner specification. This should
-          #        be fixed once https://github.com/github/super-linter/pull/2897
-          #        is merged, which includes
-          #        https://github.com/rhysd/actionlint/commit/09a60b0f5761c2d066def294e541c25a89756eea
```
mrc0mmand: Unnecessary ident change?

.github/workflows/linter.yml
```diff
@@ -37,14 +37,16 @@ jobs:
           #   - man/.* - all snippets in man pages (false positives due to
           #              missing shebangs)
           #   - .*\.(in|SKELETON) - all template/skeleton files
+          #                         except kernel-install
           #   - tools/coverity\.sh - external file (with some modifications)
-          FILTER_REGEX_EXCLUDE: .*/(man/.*|.*\.(in|SKELETON)|tools/coverity\.sh)$
```
mrc0mmand, line 41: Oh well, this is unfortunate, but it is what it is, thanks for figuring this out.

mrc0mmand (commented):
A couple of nits, otherwise LGTM, thanks!

mrc0mmand:
(To test this out you'll have to explicitly "touch" the kernel-install script, since by reverting the fixup commit this PR effectively doesn't touch kernel-install at all.)

.github/workflows/linter.yml
nabijaczleweli: Right, the defaults table misled me, and the BASH=true -> SHFMT=false conversion made the YAML validator run, which exploded on >80-character lines; fixed

nabijaczleweli:
Blegh, yeah; validated, fixed, should be g2g

mrc0mmand (approved):
Awesome, LGTM, thanks!

.github/workflows/linter.yml
mrc0mmand: Unfortunately, no, this uses Bash's POSIX regex engine (`regex(3)`), so no fancy lookaheads/lookbehinds here.

nabijaczleweli: That's what I tried (and helpfully the documentation doesn't say; well it says it's RE, which it isn't), but it uses bash `[[ =~ ]]`, which is ERE, to match these, and EREs don't have lookaheads

PR #23480, opened 2022-05-23
journal: ensure the logical status of LOCATION_TAIL and LOCATION_HEAD

### LINKED ISSUE #9934: sd_journal_next returns one entry per journal file after sd_journal_seek_tail
This issue appears for me in systemd 239 on Gentoo

I expected to get no entries from sd_journal_next immediately after calling sd_journal_seek_tail or, if any, just entries that were just created (after my call to sd_journal_seek_tail but before my call to sd_journal_next).

Instead I got 19 entries of ages up to 438 days. The problem is reproducible, I get the same 19 entries every time. None of them seem to be clustered around a specific point in time (each many hours in age apart).

There is a total number of 44 system journal files in /var/log/journal/.../ but when calling the file tool on them I get only 19 that do not return "no read permission". This read permission seems to be granted (apparently only on files since early June 2017) via ACLs for the adm and wheel groups.

When running the code as root (via sudo) I get 44 entries. This suggests that it returns one entry from each file.

I wrote a minimal C program to reproduce the problem, at least with the current state of my journal files. You can find it in this Gist:

https://gist.github.com/taladar/e44e7cebc8cdb97c6651da654067239a

ChenQi1989:
sd_journal_next call after sd_journal_seek_tail should always return
0 instead of go anywhere else to seek for a journal entry. The same
logic applies to sd_journal_previous and sd_journal_seek_head.

The LOCATION_TAIL and LOCATION_HEAD only have logical meaning, the
end/beginning of the journal. The real entry location is determined
by some follow-up journal location related call. For example, the real
location of LOCATION_TAIL, by desgin I guess, is determined by the
time an immediate follow-up call, sd_journal_previous, is made.

So any sd_journal_next call immediately after sd_journal_seek_tail
should return 0. sd_journal_previous and sd_journal_seek_head, the same.

Fixes #9934

Signed-off-by: Chen Qi <Qi.Chen@windriver.com>

### DIFF
--- a/src/libsystemd/sd-journal/sd-journal.c
+++ b/src/libsystemd/sd-journal/sd-journal.c
@@ -826,6 +826,10 @@ static int real_journal_next(sd_journal *j, direction_t direction) {
         assert_return(j, -EINVAL);
         assert_return(!journal_pid_changed(j), -ECHILD);
 
+        if ((j->current_location.type == LOCATION_TAIL && direction == DIRECTION_DOWN) ||
+            (j->current_location.type == LOCATION_HEAD && direction == DIRECTION_UP))
+                return 0;
+
         r = iterated_cache_get(j->files_cache, NULL, &files, &n_files);
         if (r < 0)
                 return r;


yuwata:
CIs are in red.

ChenQi1989:
Any comment on this one?

DaanDeMeyer:
I'm not too familiar with this code but shouldn't the fix be made in `next_beyond_location()`? In fact, in `next_beyond_location()`, we already have logic to check for EOF and to return early if that's the case. Can you check why that logic isn't triggered? I'm pretty sure if possible we should fix that condition instead of adding a new one to `real_journal_next()`.

DaanDeMeyer (changes requested):
I'm not too familiar with this code but shouldn't the fix be made in next_beyond_location()? In fact, in next_beyond_location(), we already have logic to check for EOF and to return early if that's the case. Can you check why that logic isn't triggered? I'm pretty sure if possible we should fix that condition instead of adding a new one to real_journal_next().

ChenQi1989:
@DaanDeMeyer 
Are you referring to the following code block in next_beyond_location?
        /* If we hit EOF before, we don't need to look into this file again
         * unless direction changed or new entries appeared. */
        if (f->last_direction == direction && f->location_type == LOCATION_TAIL &&
            n_entries == f->last_n_entries)
                return 0;

We can see that it's checking the file's location instead of the journal's location.
Looking at the codes, it's iterating all journal files, if the above condition meets, it proceeds to parse the remaining journal files.
The codes are as below:
        for (i = 0; i < n_files; i++) {
                JournalFile *f = (JournalFile *)files[i];
                bool found;

                r = next_beyond_location(j, f, direction);
                if (r < 0) {
                        log_debug_errno(r, "Can't iterate through %s, ignoring: %m", f->path);
                        remove_file_real(j, f);
                        continue;
                } else if (r == 0) {
                        f->location_type = LOCATION_TAIL;
                        continue;
                }

sd_journal_seek_tail/head is setting journal's LOCATION_TAIL/HEAD, and we need a check against it.

ChenQi1989:
@DaanDeMeyer
Are you referring to the following code block in next_beyond_location?

`
/* If we hit EOF before, we don't need to look into this file again * unless direction changed or new entries appeared. */ 
if (f->last_direction == direction && f->location_type == LOCATION_TAIL && 
    n_entries == f->last_n_entries) 
        return 0;
`

We can see that it's checking the file's location instead of the journal's location.
Looking at the codes, it's iterating all journal files, if the above condition meets, it proceeds to parse the remaining journal files.
The codes are as below:

`
for (i = 0; i < n_files; i++) {
            JournalFile *f = (JournalFile *)files[i];
            bool found;

            r = next_beyond_location(j, f, direction);
            if (r < 0) {
                    log_debug_errno(r, "Can't iterate through %s, ignoring: %m", f->path);
                    remove_file_real(j, f);
                    continue;
            } else if (r == 0) {
                    f->location_type = LOCATION_TAIL;
                    continue;
            }
.....
}
`

However, sd_journal_seek_tail/head is setting journal's LOCATION_TAIL/HEAD, and we need a check against it.

ChenQi1989:
@DaanDeMeyer 
I tried to make the comment a little easier to read, but github's 'add code' button does not work well enough.
Hope you still get what I mean.

ChenQi1989:
@DaanDeMeyer ping

ChenQi1989:
ping

DaanDeMeyer:
@ChenQi1989 Can you test https://github.com/systemd/systemd/pull/23909? I think that might be a better solution.

PR #23482, opened 2022-05-23
ci: bump gcc in the "build test" workflow

evverx:
gcc-12 has been released


mrc0mmand:
Interesting, both gcc jobs have finished successfully, but GH still shows them as running. I don't see anything wrong with the change itself, so I guess (and hope) there's something going on with the GH runners. Let's give it a moment if it reassesses itself.

PR #23489, opened 2022-05-24
Attempted fix for regression with LUKS devices introduced by PR#23218

### LINKED ISSUE #23429: PR#23218 introduced a regression for devices with `db_persist` option

mwilck:
On switching root, a device may have a persistent databse. In that case,
Device.enumerated_found may have DEVICE_FOUND_UDEV flag, and it is not
necessary to downgrade the Device.deserialized_found and
Device.deserialized_state. Otherwise, the state of the device unit may
be changed plugged -> dead -> plugged, if the device has not been mounted.

This is #23437, just reduced to a minimal bug fix.
Fixes #23429 (hopefully)

### DIFF
--- a/test/TEST-24-CRYPTSETUP/test.sh
+++ b/test/TEST-24-CRYPTSETUP/test.sh
@@ -66,6 +66,16 @@ EOF
     # Forward journal messages to the console, so we have something
     # to investigate even if we fail to mount the encrypted /var
     echo ForwardToConsole=yes >> "$initdir/etc/systemd/journald.conf"
+
+    if [[ -z "$INITRD" ]]; then
+        if command -v dracut >/dev/null; then
+            dracut --add crypt "$TESTDIR/initrd.img"
+            INITRD="$TESTDIR/initrd.img"
+        elif command -v mkinitcpio >/dev/null; then
+            mkinitcpio --addhooks sd-encrypt --generate "$TESTDIR/initrd.img"
+            INITRD="$TESTDIR/initrd.img"
+        fi
+    fi
 }
 
 cleanup_root_var() {
--- a/test/test-functions
+++ b/test/test-functions
@@ -337,6 +337,9 @@ qemu_min_version() {
 # Return 0 if qemu did run (then you must check the result state/logs for actual
 # success), or 1 if qemu is not available.
 run_qemu() {
+    if [[ -z "$INITRD" && -n "$TESTDIR" && -f "$TESTDIR/initrd.img" ]]; then
+        INITRD="$TESTDIR/initrd.img"
+    fi
     if [ -f /etc/machine-id ]; then
         read -r MACHINE_ID </etc/machine-id
         [ -z "$INITRD" ] && [ -e "$EFI_MOUNT/$MACHINE_ID/$KERNEL_VER/initrd" ] \


fbuihuu:
I don't see the point of this PR. We already have https://github.com/systemd/systemd/pull/23437 and PR #23436, this one is adding only more confusion.

Again there is no rush so let's discuss the regression in on of the aforementioned PRs.

bluca:
well, if there's a regression slipped in v251, we need to fix it with some urgency

fbuihuu:
I think we should try to reproduce the prob instead based on the description given in https://github.com/systemd/systemd/issues/23429#issuecomment-1135570920

mwilck:
> As I wrote we already have 2 PRs. If we need a quick fix, working on #23437 should be enough I think. And "Fixes #23429 (hopefully)" doesn't make me confident at all.

I proposed a minimal fix. That's my understanding how a regression should be addressed. Both other PRs contain additional unrelated changes. I wrote "hopefully" because the fix hadn't been verified, and I didn't want to boast of something I wasn't 100% certain of.

> See comments in #23429 where this wasn't supposed to be a regression and wasn't supposed to introduce any issue.

I'm unsure what you mean. I don't deny that some of my statements were wrong. But I don't think I'm the only one. At least I have tried to analyze the issues in depth. I haven't realized the LUKS issue. But nobody else did, either.

fbuihuu:
Sorry I wasn't explained myself well enough. We all made some mistakes because it's hard to foresee the consequences of any changes in this part of the code even if we put a lot of thoughts. But I think my suggestion to reproduce the regression first and then verify that the fix actually is fixing it is reasonable instead of running for a fix that we're not again 100% sure.

https://github.com/systemd/systemd/pull/23437 is already an attempt to fix this problem and has the same change that this PR is introducing. Indeed it has 2 small additional commits but those should be considered right now anyway as they're likely to be committed soon or later. We would avoid testing this stuff twice.

I'm re-opening this PR and let @yuwata decides if he prefers dealing with this PR rather than his.

mwilck:
My thinking was that if this PR actually fixes the issue(s) reported, it might be easier to merge this minimal PR than #23437. But I'm not familiar with how this is handled in systemd. Most importantly, we need tests by the affected users. @tanriol, as a Gentoo user, may be able to test this rather easily. @keszybz seems to have ways to distribute fixes to Fedora users quickly, too. That should cover most regressions reported so far.

yuwata:
Sorry for late response. Will take a look.

tanriol:
Sorry, applying this patch on top of systemd 251 did not fix the issue on my system.

mwilck:
@tanriol, I've pushed another patch which has fixed the issue for me (I was able to reproduce your problem). I'd be very grateful if you could apply 0543ebd on top and test again.

bluca:
If you can reproduce it, can you add a test case too please to TEST-24-CRYPTSETUP? Otherwise we'll never know if another regression happens

mwilck:
> If you can reproduce it, can you add a test case too please to TEST-24-CRYPTSETUP? 

What I meant is I was able to set up a box with LUKS encrypted root, saw the same issue, and was able to fix it. I don't know how to transform this into a unit test. I can have a look later, but I know almost nothing about the systemd unit tests, except for the udev test which I've been working on in the past.

dustymabe:
If we can get the `rpm-build:fedora-rawhide-x86_64` action to succeed I can take the resulting RPM and run it through our Fedora CoreOS test suite.

mrc0mmand:
> If we can get the `rpm-build:fedora-rawhide-x86_64` action to succeed I can take the resulting RPM and run it through our Fedora CoreOS test suite.

I've already pinged @keszybz, since we might need to revert https://src.fedoraproject.org/rpms/systemd/c/9a48377e0adb6af26f8e7a89dd0c186cc951efa8?branch=rawhide once again, as the builders are still having some issues.

mrc0mmand:
> > If you can reproduce it, can you add a test case too please to TEST-24-CRYPTSETUP?
> 
> What I meant is I was able to set up a box with LUKS encrypted root, saw the same issue, and was able to fix it. I don't know how to transform this into a unit test. I can have a look later, but I know almost nothing about the systemd unit tests, except for the udev test which I've been working on in the past.

I've opened #23517 which should add a coverage for this. Feel free to cherry-pick the commit into this branch, as it should make it easier to test the changes with the test included.

yuwata:
@mwilck Thank you for the analysis of the issue. LGTM. I cherry-picked @mrc0mmand's commit about test update.

tanriol:
Will test this tomorrow, sorry for delay.

tanriol:
systemd-251 with 2659b041a5bb505aa0f6661205567aa9a7640e47 and 0543ebd9c0262d94ad32a5f450b186b3f625ebd9 on top works for me.

yuwata:
> systemd-251 with [2659b04](https://github.com/systemd/systemd/commit/2659b041a5bb505aa0f6661205567aa9a7640e47) and [0543ebd](https://github.com/systemd/systemd/commit/0543ebd9c0262d94ad32a5f450b186b3f625ebd9) on top works for me. Do I need to retest on newer commits in this PR?

Thanks. I think retest is not necessary, as they are just rebased.

yuwata:
The two main commits are in PR #23517. Closing.

PR #23492, opened 2022-05-24
boot: Fix bad CompareMem call

medhefgo:


keszybz:
This was introduced in v249-1389-g393955fd21, so no need to backport to v249-stable.

PR #23493, opened 2022-05-24
Simplify random number selection

zx2c4:
Most modern systems supported by systemd have getrandom(). On ones that
don't, there's /dev/urandom. But these systems are pretty rare.
Nonetheless we keep the fallback.

Systems that don't have getrandom() but also don't have /dev/urandom are
basically non-existent and would likely choke to run systemd anyway. So
let's stop playing games with that extreme and unlikely edge case. This
allows us to simplify the convoluted RNG situation into two cases with
optimized algorithms for each:

1) Randomness for crypto: this one needs to wait until the RNG is
   initialized. So it uses getrandom(0). If that's not available, it
   polls on /dev/random, and then reads from /dev/urandom.

2) Randomness for other things: this one uses getrandom(GRND_INSECURE).
   If it's not available it uses getrandom(GRND_NONBLOCK). And if that
   would block, then it falls back to /dev/urandom.

Notice that nowhere in here is the ugly libc rand(). There's simply no
need to play games with that stuff. Instead we can insist on getrandom()
or /dev/urandom working and being available, as this is essentially
needed anyway on modern Linux and extremely unlikely to be broken
anywhere that systemd can run.

### DIFF
--- a/src/basic/random-util.c
+++ b/src/basic/random-util.c
@@ -30,183 +30,86 @@
 #include "missing_syscall.h"
 #include "parse-util.h"
 #include "random-util.h"
-#include "siphash24.h"
 #include "time-util.h"
 
-static bool srand_called = false;
-
-int genuine_random_bytes(void *p, size_t n, RandomFlags flags) {
-        static int have_syscall = -1;
+void random_bytes(void *p, size_t n) {
+        static bool have_getrandom = true, have_grndinsecure = true;
         _cleanup_close_ int fd = -1;
 
-        /* Gathers some high-quality randomness from the kernel. This call won't block, unless the RANDOM_BLOCK
-         * flag is set. If it doesn't block, it will still always return some data from the kernel, regardless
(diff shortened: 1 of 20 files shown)


bluca:
> Notice that nowhere in here is the ugly libc rand(). There's simply no
> need to play games with that stuff. Instead we can insist on getrandom()
> or /dev/urandom working and being available, as this is essentially
> needed anyway on modern Linux and extremely unlikely to be broken
> anywhere that systemd can run.

I'm not sure about this - the device might be available in the sense that it might be there on the system, but it might be blocked by a variety of things - lsm, mac, dac, and so on. I don't see any reason to do hard breakages in those cases, if the defaults when everything is set up are working correctly? They are just fallbacks, no?

zx2c4:
Has there ever actually been something in the real world that breaks it? I'm wary against adding all this code complexity and craziness over edge cases that don't really exist.

If you insist, I can write up a better fallback than what was there before. But meh.

zx2c4:
I'll close this and open a new one for a different approach.

PR #23496, opened 2022-05-24
Simplify random number selection

zx2c4:
We currently have a convoluted and complex selection of which random numbers to use. We can simplify this down to two functions that cover all of our use cases:

1) Randomness for crypto: this one needs to wait until the RNG is initialized. So it uses getrandom(0). If that's not available, it polls on /dev/random, and then reads from /dev/urandom. This function returns whether or not it was successful, as before.

2) Randomness for other things: this one uses getrandom(GRND_INSECURE). If it's not available it uses getrandom(GRND_NONBLOCK). And if that would block, then it falls back to /dev/urandom. And if /dev/urandom isn't available, it uses the fallback code. It never fails and doesn't return a value.

These two cases match all the uses of randomness inside of systemd.

I would prefer to make both of these return void, and get rid of the fallback code, and simply assert in the incredibly unlikely case that /dev/urandom doesn't exist, which is what #23493 does. But @bluca disagrees, so this commit attempts to instead keep case (1) returning a return value, which all the callers already check, and fix the fallback code in (2) to be less bad than before.

For the less bad fallback code for (2), we now use auxval and some timestamps, together with various counters representing the invocation, hash it all together and provide the output. Provided that AT_RANDOM is secure, this construction is probably okay too.

CC @poettering

### DIFF
--- a/src/basic/random-util.c
+++ b/src/basic/random-util.c
@@ -30,183 +30,142 @@
 #include "missing_syscall.h"
 #include "parse-util.h"
 #include "random-util.h"
-#include "siphash24.h"
+#include "sha256.h"
 #include "time-util.h"
 
-static bool srand_called = false;
-
-int genuine_random_bytes(void *p, size_t n, RandomFlags flags) {
-        static int have_syscall = -1;
-        _cleanup_close_ int fd = -1;
-
-        /* Gathers some high-quality randomness from the kernel. This call won't block, unless the RANDOM_BLOCK
-         * flag is set. If it doesn't block, it will still always return some data from the kernel, regardless
-         * of whether the random pool is fully initialized or not. When creating cryptographic key material you
-         * should always use RANDOM_BLOCK. */
-
-        if (n == 0)
-                return 0;
-
-        /* Use the getrandom() syscall unless we know we don't have it. */
-        if (have_syscall != 0 && !HAS_FEATURE_MEMORY_SANITIZER) {
-                for (;;) {
-                        ssize_t l = getrandom(p, n, FLAGS_SET(flags, RANDOM_BLOCK) ? 0 : GRND_INSECURE);
-
-                        if (l > 0) {
-                                have_syscall = true;
-
-                                if ((size_t) l == n)
-                                        return 0; /* Yay, success! */
-
-                                /* We didn't get enough data, so try again */
-                                assert((size_t) l < n);
-                                p = (uint8_t*) p + l;
-                                n -= l;
-                                continue;
-
-                        } else if (l == 0) {
-                                have_syscall = true;
-                                return -EIO;
-
-                        } else if (ERRNO_IS_NOT_SUPPORTED(errno)) {
-                                /* We lack the syscall, continue with reading from /dev/urandom. */
-                                have_syscall = false;
-                                break;
-
-                        } else if (errno == EINVAL) {
-                                /* If we previously passed GRND_INSECURE, and this flag isn't known, then
-                                 * we're likely running an old kernel which has getrandom() but not
-                                 * GRND_INSECURE. In this case, fall back to /dev/urandom. */
-                                if (!FLAGS_SET(flags, RANDOM_BLOCK))
-                                        break;
-
-                                return -errno;
-                        } else
-                                return -errno;
-                }
-        }
-
-        fd = open("/dev/urandom", O_RDONLY|O_CLOEXEC|O_NOCTTY);
-        if (fd < 0)
-                return errno == ENOENT ? -ENOSYS : -errno;
-
-        return loop_read_exact(fd, p, n, true);
-}
-
-static void clear_srand_initialization(void) {
-        srand_called = false;
-}
-
-void initialize_srand(void) {
-        static bool pthread_atfork_registered = false;
-        unsigned x;
-
-        if (srand_called)
-                return;
+/* This is a "best effort" kind of thing, but has no real security value.
+ * So, this should only be used by random_bytes(), which is not meant for
+ * crypto. This could be made better, but we're *not* trying to roll a
+ * userspace prng here, or even have forward secrecy, but rather just do
+ * the shortest thing that is at least better than libc rand(). */
+static void fallback_random_bytes(void *p, size_t n) {
+        static thread_local uint64_t fallback_counter = 0;
+        struct {
+                char label[32];
+                uint64_t call_id, block_id;
+                usec_t stamp_mono, stamp_real;
+                pid_t pid, tid;
+                uint8_t auxval[16];
+        } state = {
+                /* Arbitrary domain separation to prevent other usage of AT_RANDOM from clashing. */
+                .label = "systemd fallback random bytes v1",
+                .call_id = fallback_counter++,
+                .stamp_mono = now(CLOCK_MONOTONIC),
+                .stamp_real = now(CLOCK_REALTIME),
+                .pid = getpid(),
+                .tid = gettid()
+        };
 
 #if HAVE_SYS_AUXV_H
-        /* The kernel provides us with 16 bytes of entropy in auxv, so let's try to make use of that to seed
-         * the pseudo-random generator. It's better than nothing... But let's first hash it to make it harder
-         * to recover the original value by watching any pseudo-random bits we generate. After all the
-         * AT_RANDOM data might be used by other stuff too (in particular: ASLR), and we probably shouldn't
-         * leak the seed for that. */
-
-        const void *auxv = ULONG_TO_PTR(getauxval(AT_RANDOM));
-        if (auxv) {
-                static const uint8_t auxval_hash_key[16] = {
-                        0x92, 0x6e, 0xfe, 0x1b, 0xcf, 0x00, 0x52, 0x9c, 0xcc, 0x42, 0xcf, 0xdc, 0x94, 0x1f, 0x81, 0x0f
-                };
-
-                x = (unsigned) siphash24(auxv, 16, auxval_hash_key);
-        } else
-#endif
-                x = 0;
-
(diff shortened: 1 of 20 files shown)


zx2c4:
Whoops, sure, sorry.

zx2c4:
Right, got it. Should be all set now.

(And hopefully this looks better to you than the previous one.)

zx2c4:
Huh, I'm seeing some CI errors, but I wonder if these are caused by the base commit rather than this new code?

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
-         * flag is set. If it doesn't block, it will still always return some data from the kernel, regardless
-         * of whether the random pool is fully initialized or not. When creating cryptographic key material you
-         * should always use RANDOM_BLOCK. */
+#if HAVE_SYS_AUXV_H
+        memcpy(state.auxval, ULONG_TO_PTR(getauxval(AT_RANDOM)), sizeof(state.auxval));
+#endif
 
-        if (n == 0)
-                return 0;
+        while (n) {
```
medhefgo: `n > 0`

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
-        } else
-#endif
-                x = 0;
-
-        x ^= (unsigned) now(CLOCK_REALTIME);
-        x ^= (unsigned) gettid();
-
-        srand(x);
-        srand_called = true;
+        for (;;) {
```
medhefgo, line 88: `while (have_getrandom)`

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
+                        p = (uint8_t *) p + l;
+                        n -= l;
+                        continue; /* Interrupted by a signal; keep going. */
+                } else if (ERRNO_IS_NOT_SUPPORTED(errno)) {
+                        have_getrandom = false;
+                        break; /* No syscall; fallback to /dev/urandom. */
+                } else if (errno == EINVAL && have_grndinsecure) {
+                        have_grndinsecure = false;
+                        continue; /* No GRND_INSECURE; fallback to GRND_NONBLOCK. */
+                } else if (errno == EAGAIN && !have_grndinsecure) {
```
medhefgo: No `{}` for this block.

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
+static void fallback_random_bytes(void *p, size_t n) {
+        static thread_local uint64_t fallback_counter = 0;
+        struct {
+                char label[32];
+                uint64_t call_id, block_id;
+                usec_t stamp_mono, stamp_real;
+                pid_t pid, tid;
+                uint8_t auxval[16];
+        } state = {
+                .label = "systemd fallback random bytes v1",
```
medhefgo, line 51: What is this for?

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
-#if RAND_STEP >= 3
-                if ((size_t) (q - (uint8_t*) p + 2) < n)
-                        q[2] = rr >> 16;
-#endif
-#if RAND_STEP >= 2
-                if ((size_t) (q - (uint8_t*) p + 1) < n)
-                        q[1] = rr >> 8;
-#endif
-                q[0] = rr;
+        for (;;) {
```
medhefgo, line 130: `while (have_getrandom)`

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
+                if (l > 0) {
+                        if ((size_t) l == n)
+                                return 0; /* Done reading, success. */
+                        p = (uint8_t *) p + l;
+                        n -= l;
+                        continue; /* Interrupted by a signal; keep going. */
+                } else if (ERRNO_IS_NOT_SUPPORTED(errno)) {
+                        have_getrandom = false;
+                        break; /* No syscall; fallback to /dev/urandom. */
+                } else if (l == 0) {
```
medhefgo: No `{}`

medhefgo (commented):
Just some coding style points. Can't really say much about the random changes (though, it does look nicer to me).

poettering:
getrandom() was added in kernel 3.17. Our current baseline is 3.15. We wanted to raise the baseline to 4.3 anyway soon (to be able to rely on ambient caps). Thus we might even go further here and just hard rely on getrandom().

poettering:
> getrandom() was added in kernel 3.17. Our current baseline is 3.15. We wanted to raise the baseline to 4.3 anyway soon (to be able to rely on ambient caps). Thus we might even go further here and just hard rely on getrandom().

What we'd lose is compat with unnecessarily restrictive seccomp filters (i.e. container managers which block getrandom(), but allow /dev/urandom)

src/basic/random-util.c
```diff
@@ -30,183 +30,140 @@
...
-         *
-         * This function is hence great for things like seeding hash tables, generating random numeric UNIX
-         * user IDs (that are checked for collisions before use) and such.
-         *
-         * This function is hence not useful for generating UUIDs or cryptographic key material.
-         */
-
-        if (genuine_random_bytes(p, n, 0) >= 0)
-                return;
+                poll_fd = open("/dev/random", O_RDONLY);
```
poettering: O_CLOEXEC

src/basic/random-util.c
zx2c4: Preventing other uses of AT_RANDOM from colliding. "Crypto hygiene".

src/basic/random-util.c
zx2c4: Okay, will do.

src/basic/random-util.c
zx2c4: That seems like a "just because we can doesn't mean we should" situation. Clearer without that. Not gonna do this.

src/basic/random-util.c
zx2c4: Will do.

src/basic/random-util.c
zx2c4: Ditto here.

src/basic/random-util.c
zx2c4: Will do.

src/basic/random-util.c
zx2c4: Thanks, will do.

src/basic/random-util.h
```diff
@@ -5,15 +5,8 @@
...
-typedef enum RandomFlags {
-        RANDOM_BLOCK              = 1 << 0, /* Rather block than return crap randomness (only if the kernel supports that) */
-} RandomFlags;
-
-int genuine_random_bytes(void *p, size_t n, RandomFlags flags); /* returns "genuine" randomness, optionally filled up with pseudo random, if not enough is available */
-void pseudo_random_bytes(void *p, size_t n);                    /* returns only pseudo-randommess (but possibly seeded from something better) */
-void random_bytes(void *p, size_t n);                           /* returns genuine randomness if cheaply available, and pseudo randomness if not. */
-
-void initialize_srand(void);
+void random_bytes(void *p, size_t n); /* returns random bytes suitable for most uses. */
```
bluca: Shouldn't the comment here be more explicit about the fact that bad quality data might be returned?

bluca:
> > getrandom() was added in kernel 3.17. Our current baseline is 3.15. We wanted to raise the baseline to 4.3 anyway soon (to be able to rely on ambient caps). Thus we might even go further here and just hard rely on getrandom().
> 
> What we'd lose is compat with unnecessarily restrictive seccomp filters (i.e. container managers which block getrandom(), but allow /dev/urandom)

seccomp, apparmor, selinux, bpf lsm, device not mounted, file descriptor exaustion... there's tons of stuff that could go really wrong, and cause grief. The fallback, especially the new one, seems simple enough to me. This area doesn't seem to be particularly high maintenance, cost-wise, at least judging from the git log.

What about adding a new taint if we end up using the fallback, instead of removing it?

src/basic/random-util.h
zx2c4: The one before wasn't, so I didn't here, but I agree with you, so I'll change that and force push.

zx2c4:
That really just sounds like adding bloat for extreme edge cases where those edge cases are the things that we should be encouraging people to fix. Having fallback code of meh security at all is really unfortunate. The closer we can get to removing that the better.

My experience with taint bits is it doesn't really motivate much, unless it's super annoying and in your face. Like: every systemd CLI utility prints in bright blinking red "TAINTED!" before and after the command output and rings the terminal bell and and... Less than this and people just ignore it.

But, anyway, that's not what this PR is about, so I think we can merge this, and then maybe I'll send a few different PRs for the next steps and we can see what they look like?

src/basic/random-util.c
bluca: please add a comment, as it's not immediately obvious

bluca (approved):

src/basic/random-util.c
zx2c4: Done.

dnicolodi: Just trying to understand: wouldn't a collision require also all other parts of the `state` struct to be identical?

zx2c4: Yes. The idea is to separate the domain definitively so, rather than probably so. For this reason, most crypto protocols (TLS, Signal, WireGuard, etc) use labels at the beginning of hash chains. I mentioned "hygiene" in my last message, because this is something I do out of habit. We could _probably_ reason that it's not necessary and that this is overly paranoid. But it's cheap enough to do, for code that will never be used anyway, and it's one less thing to think about.

zx2c4:
Looks like the CI is all green (the arch issue was fixed I guess). Is this mergeable now?

bluca:
we should probably wait for @poettering but he's out of office at the moment

keszybz:
LGTM.

Issue #23424, opened 2022-05-18
TEST-60-MOUNT-RATELIMIT is unstable


mrc0mmand:
**systemd version the issue has been seen with**
 > latest main

**Used distribution**
 > Arch Linux

Since May 13th I've noticed several failed runs with `TEST-60-MOUNT-RATELIMIT` being the culprit:

```
[  133.292006] testsuite-60.sh[446]: + journalctl -u init.scope
[  133.292968] testsuite-60.sh[447]: + grep -q '(mount-monitor-dispatch) left rate limit'
[  133.530574] testsuite-60.sh[239]: + sleep 5
[  138.533647] testsuite-60.sh[449]: ++ date -u +%s
[  138.536337] testsuite-60.sh[239]: + [[ 1652841000 -le 1652840999 ]]
[  138.536785] testsuite-60.sh[239]: + '[' 0 = 0 ']'
[  138.536916] testsuite-60.sh[239]: + exit 24
[  138.538091] systemd[1]: systemd-journald.service: Received EPOLLHUP on stored fd 68 (stored), closing.
[  138.538664] systemd[1]: Received SIGCHLD from PID 239 (bash).
[  138.538779] systemd[1]: Child 239 (bash) died (code=exited, status=24/n/a)
[  138.539026] systemd[1]: testsuite-60.service: Child 239 belongs to testsuite-60.service.
[  138.539133] systemd[1]: testsuite-60.service: Main process exited, code=exited, status=24/n/a
[  138.539293] systemd[1]: testsuite-60.service: Failed with result 'exit-code'.
[  138.539395] systemd[1]: testsuite-60.service: Service will not restart (restart setting)
[  138.541385] systemd[1]: testsuite-60.service: Changed start -> failed
[  138.541492] systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.DBus.Properties member=PropertiesChanged cookie=643 reply_cookie=0 signa>
[  138.569260] systemd[1]: testsuite-60.service: Failed to delete cgroup entry from LSM BPF map: No such file or directory
[  138.569425] systemd[1]: testsuite-60.service: Job 167 testsuite-60.service/start finished, result=failed
[  138.569521] systemd[1]: Failed to start TEST-60-MOUNT-RATELIMIT.
```

According to the logs it looks like the mount unit never leaves the rate limited state (or at least the respective log message is missing):

```
[...]

Issue #4641, opened 2016-11-10
systemd-boot ordering incorrect for debian kernel versioning


sjoerd-ccu:
### Submission type

  - [x] Bug report
  - [ ] Request for enhancement (RFE)

### systemd version the issue has been seen with

232

### Used distribution

Debian unstable

### In case of bug report: Expected behaviour you didn't see

Kernels are ordered based on their version and the latest is selected by defualt

### In case of bug report: Unexpected behaviour you saw
My system has the following versions (form the loader snippets):

version    4.8.0-1-amd64
version    4.8.0-2-amd64
version    4.8.0-rc5-amd64
version    4.8.0-rc8-amd64
version    4.8.0-trunk-amd64

systemd-boot orders them in this order as well, meaning the rc kernels will be picked up before the release kernels. The same goes for -trunk, which in the debian kernel versioning are experimental.

I don't know what e.g. fedora does with their kernel versioning, but it seems we might need to either tune the version that ends up in the boot snippets or change the systemd-boot ordering so it just works.

poettering:
but isn't debian's naming a bit weird here?

sjoerd-ccu:
Not for me, but i'm used to it :p 

In any case that versioning policy of debian is unlikely to change, so we either have to tweak the systemd-boot ordering or tweak the calls to kernel-install from the debian side to adjust the naming to make sd-boot happy.

poettering:
iirc, the version sorting algorithm used by sd-boot originates in debian actually. @msekletar, @kaysievers might now more. The code is here:

https://github.com/systemd/systemd/blob/master/src/boot/efi/boot.c#L871

medhefgo:
Version sorting (and selection) have changed recently. I now get this order (highest to lowest priority):
```
version 4.8.0-2-amd64
version 4.8.0-1-amd64
version 4.8.0-trunk-amd64
version 4.8.0-rc8-amd64
version 4.8.0-rc5-amd64
```

keszybz:
The order listed by @medhefgo seems correct and expected.

[...]

PR #23504, opened 2022-05-24
Refactor the BLS and add a description of version sorts

keszybz:
There'll be other steps to do, e.g. the discussion of boot counting, but I think it's better to start small. Wording suggestions (and semantic fixes) are welcome, I'll try to incorporate them into the text.

### DIFF
--- a/docs/BOOT_LOADER_SPECIFICATION.md
+++ b/docs/BOOT_LOADER_SPECIFICATION.md
@@ -35,52 +35,58 @@ The target audience for this specification is:
 
 ## The boot partition
 
-Everything described below is located on a file system that will be called
-`$BOOT`. The boot loader or user-space programs reading the boot loader
-configuration should pick `$BOOT` according to the following rules:
+Everything described below is located on one or two partitions. The boot loader
+or user-space programs reading the boot loader configuration should locate them
+in the following manner:
 
 * On disks with an MBR partition table:
 
-  * A partition with the type ID of 0xEA shall be used as `$BOOT`.
+  * The boot partition — partition with the type ID of 0xEA — shall be used
+    for boot loader configuration and entries.
 
 * On disks with GPT (GUID Partition Table)
 
-  * If an Extended Boot Loader Partition (XBOOTLDR partition for short) — a
-    partition with GPT type GUID of `bc13c2ff-59e6-4262-a352-b275fd6f7172` —
-    is found, it should be used as `$BOOT`.
+  * The EFI System Partition (ESP for short) — a partition with GPT type GUID
+    of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — should be used for boot loader
+    configuration and boot entries.
 
-  * Otherwise, the EFI System Partition (ESP for short) — a partition with
-    GPT type GUID of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — should
-    be used as `$BOOT`.
+  * Optionally, an Extended Boot Loader Partition (XBOOTLDR partition for
+    short) — a partition with GPT type GUID of
+    `bc13c2ff-59e6-4262-a352-b275fd6f7172` — may be used as an additional
+    location for boot loader entries. This partition must be located on the
+    same disk as the ESP.
+
+In the text below, `$BOOT` will be used to refer to (the root of) the first of
+the two partitions (the boot partition on MBR disks and the ESP on GPT disks),
+and `$XBOOTLDR` will be used to refer to (the root of) the optional second
+partition.
 
 An installer for the operating system should use this logic when selecting or
 creating partitions:
 
   * If `$BOOT` is not found, a new suitably sized partition (let's say 500MB)
-    should be created and used as `$BOOT`, matching the characteristics
-    described above. On disks with GPT, just an ESP partition without a
-    XBOOTLDR partition should be created.
+    should be created, matching the characteristics described above. On disks
+    with GPT, only the ESP partition without the XBOOTLDR partition should be
+    created.
 
   * If the OS is installed on a disk with GPT and the ESP partition is found
     but is too small, a new suitably sized (let's say 500MB) XBOOTLDR partition
-    shall be created and used as `$BOOT`.
+    shall be created.
 
-This placeholder file system shall be determined during _installation time_,
-and an fstab entry may be created. It should be mounted to either `/boot/` or
-`/efi/`. Additional locations like `/boot/efi/` (with `/boot/` being a separate
-file system) might be supported by implementations. This is not recommended
-because the mounting of `$BOOT` is then dependent on and requires the mounting
-of the intermediate file system.
+Those file systems shall be determined during _installation time_, and an fstab
+entry may be created. If only one partition is used, it should be mounted on
+`/boot/`. If both XBOOTLDR partition and the ESP are used, they should be
+mounted on `/boot` and `/efi`, or on `/boot` and `/boot/efi`.
 
-**Note:** _`$BOOT` is **shared** among all OS installations of a system.
-Instead of maintaining one `$BOOT` per installed OS (as `/boot/` was
-traditionally handled), all installed OSes use the same place for boot-time
+**Note:** _Those file systems are **shared** among all OS installations on the
+system. Instead of maintaining one boot partition per installed OS (as `/boot/`
+was traditionally handled), all installed OSes use the same place for boot-time
 configuration._
 
-For systems where the firmware is able to read file systems directly, `$BOOT`
-must be a file system readable by the firmware. For other systems and generic
-installation and live media, `$BOOT` must be a VFAT (16 or 32) file system.
-Applications accessing `$BOOT` should hence not assume that fancier file system
+For systems where the firmware is able to read file systems directly, the ESP
+must — and the XBOOTLDR partition should — be a file system readable by the
+firmware. For most systems this means VFAT (16 or 32 bit). Applications
+accessing both partitions should hence not assume that fancier file system
 features such as symlinks, hardlinks, access control or case sensitivity are
 supported.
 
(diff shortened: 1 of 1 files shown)


src/fundamental/string-util-fundamental.c
```diff
@@ -124,8 +124,10 @@ sd_int strverscmp_improved(const sd_char *a, const sd_char *b) {
          *  (newer) 124-1
          */
 
-        if (isempty(a) || isempty(b))
-                return CMP(strcmp_ptr(a, b), 0);
+        if (!a)
+                a = (const sd_char *) "";
```
medhefgo: `strempty`?

bluca (approved):

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -7,120 +7,63 @@ SPDX-License-Identifier: LGPL-2.1-or-later
...
-    `$BOOT`.
-  * Otherwise, if the OS is installed on a disk with an MBR partition table, a
-    new partition with type id of 0xEA shall be created, of a suitable size
-    (let's say 500MB), and it should be used as `$BOOT`.
 * On disks with GPT (GUID Partition Table)
-  * If the OS is installed on a disk with GPT, and an Extended Boot Loader
-    Partition (or XBOOTLDR partition for short), i.e. a partition with GPT type
-    GUID of `bc13c2ff-59e6-4262-a352-b275fd6f7172`, already exists, it should
+
+  * If an Extended Boot Loader Partition (XBOOTLDR partition for short) — a
```
medhefgo: I don't like the wording of this. It sounds (to me) like XBOOT partitions replaces the ESP, when it really extends it. It should also be stated that only the ESP can contain the actual bootloader and its configuration (loader.conf etc.). The XBOOT can only old boot entries (type 1 or 2).

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -187,102 +132,129 @@ the root of the file system. Specifically, if `$BOOT` is the ESP, then
...
+  unique. If a boot loader discovers two entries with the same title it should
+  show more than just the raw title in the UI, for example by appending the
+  `version` field. This field is optional.
+
+  Example: `title Fedora 18 (Spherical Cow)`
+
+* `version` is a human-readable version for this menu item. This is usually the
+  kernel version and is intended for use by OSes to install multiple kernel
+  versions with the same `title` field. This field shall be in a syntax that is
+  useful for Debian-style version sorts, so that the boot loader UI can
```
medhefgo: Funny you mention debian-style version sort. Afaik our version compare isn't fully debian compatible (#4641).

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -190,14 +191,14 @@ The following keys are recognized:
...
-  usually is not). If non-unique the the `machine-id` (lexicographically
-  increasing) and `version` (lexicographically decreasing, i.e. newest version
-  first) fields described above are used as secondary/ternary sorting keys. If
-  this field is not set entries are typically sorted by the `.conf` file name
-  of the entry.
+  [os-release](https://www.freedesktop.org/software/systemd/man/os-release.html),
+  possibly with an additional suffix. This field is optional. If set, it is
+  used as primary sorting key for the entries on display (lexicographically
+  increasing). It does not have to be unique (and usually is not). If
+  non-unique the the `machine-id` (lexicographically increasing) and `version`
```
medhefgo: Double the.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -393,6 +388,85 @@ creating a partition and file system for it) and creates the `/loader/entries/`
 directory in it. It then installs an appropriate boot loader that can read
 these snippets. Finally, it installs one or more kernel packages.
 
+## Sorting
+
+The boot loader menu should generally show entries in some order meaningful to
+the user. The `title` key is free-form and not suitable to be used as the
+primary sorting key. Instead, in order of priority, the `sort-key`
+(lexicographically increasing), `machine-id` (lexicographically increasing),
```
medhefgo: User interfaces these days say just `A-Z`/`Z-A` instead of complicated words such as (lexicographically) ascending. Imho, this would much easier to grok.

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: That was in the original text, which I only move around here. This text is later removed. (I think it'd be maybe worth it to add a subsection that enumerates differences to rpm and deb version comparisons, since we're not following either exactly.)

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: This is also just moving the text around. It is fixed in a later commit.

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: Right, but the thing is that this is not just an alphanumerical sort, but that the string has "structure". So I can't say that comparsion 9-0 is used for digits, because it's more like "skip 0, but not if it's the only thing, and then evaluate as number, order +∞ - 0".

I added "A-Z" or "Z-A" in the description, but didn't remove the existing text.

keszybz: Hmm, but now I realize that the text is wrong in one important aspect: it implies that strverscmp can be used for all keys, including e.g. machine-id. But a normal alphabetical sort should be used there.

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: Thanks for pointing this out. I reworked the text quite a bit, adding one more commit at the end.

keszybz:
Updated. As mentioned above:
- the confusion about what $BOOT is and what goes there is hopefully fixed
- the description of sorting is updated. It now clearly says that alphanumerical sorts should be used for e.g. the machine ID (which matters, because 0002a… should be before 01a…, and with strverscmp_improved, it'd be after), and a new subsection briefly refers to strcmp.
- a minor issue in the `compare-versions` output if fixed.

TODO
```diff
@@ -1420,6 +1420,7 @@ Features:
   - make it operate on loopback files, dissecting enough to find ESP to operate on
   - bootspec: properly support boot attempt counters when parsing entry file names
   - support --quiet in is-installed, update
+  - can we load Type#2 kernels from XBOOTLDR?
```
medhefgo: You got a bit confused there. It can load type1 and 2 from both ESP and XBOOT. But it does not look at any other configuration (loader.conf, random-seed).

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -35,52 +35,56 @@ The target audience for this specification is:
 
 ## The boot partition
 
-Everything described below is located on a file system that will be called
-`$BOOT`. The boot loader or user-space programs reading the boot loader
-configuration should pick `$BOOT` according to the following rules:
+Everything described below is located on one or two partitions. The boot loader
+or user-space programs reading the boot loader configuration should locate the
```
medhefgo: This sentence is borked

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -35,52 +35,56 @@ The target audience for this specification is:
...
-    GPT type GUID of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — should
-    be used as `$BOOT`.
+  * Optionally, an Extended Boot Loader Partition (XBOOTLDR partition for
+    short) — a partition with GPT type GUID of
+    `bc13c2ff-59e6-4262-a352-b275fd6f7172` — may be used as an additional
+    location for boot loader entries.
+
+In the text below, `$BOOT` will be used to refer to (the root of) the first of
+the two partitions (the boot partition on MBR disks and the ESP on GPT disks),
+and `$BOOT2` will be used to refer to (the root of) the optional second partition.
```
medhefgo: `$XBOOT` or `$XBOOTLDR` would be better

TODO
keszybz: Ah, fuck. Indeed. I was looking at the code and got confused. I'll readjust the text again.

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: Hmm, then I think I'd just use XBOOTLDR, without the dollar.

medhefgo: Maybe just use ESP, XBOOTLDR and $BOOT if it's either?

keszybz: There are following cases:
1. ESP or MBR /boot (for config)
2. ESP or MBR /boot and XBOOTLDR (for entries)

And I think in 2. it's better to explicitly say that both places are checked, then to imply that by using something like $BOOT to refer to either. So overall, $BOOT doesn't seem to be useful in the text at all. I think that especially with s/$BOOT2/XBOOTLDR/ the text will be much more comprehensible than before.

keszybz:
centos ci (arch linux):  TEST-60-MOUNT-RATELIMIT failed, looks unrelated.

mrc0mmand:
> centos ci (arch linux): TEST-60-MOUNT-RATELIMIT failed, looks unrelated.

Yup, that's #23424.

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: Updated like that. When I started changing the text, I realized that your original idea works better with markup. `XBOOTLDR/some/path` looks strange. So in the end it's `$BOOT` and `$XBOOTLDR`.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -35,52 +35,57 @@ The target audience for this specification is:
...
-    partition with GPT type GUID of `bc13c2ff-59e6-4262-a352-b275fd6f7172` —
-    is found, it should be used as `$BOOT`.
+  * The EFI System Partition (ESP for short) — a partition with GPT type GUID
+    of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — should be used for boot loader
+    configuration and boot entries.
 
-  * Otherwise, the EFI System Partition (ESP for short) — a partition with
-    GPT type GUID of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — should
-    be used as `$BOOT`.
+  * Optionally, an Extended Boot Loader Partition (XBOOTLDR partition for
```
medhefgo, line 53: Just remembered: The XBOOTLDR partition must reside on the same disk where the ESP is located.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -355,23 +357,23 @@ EFI boot loaders.
...
-`$BOOT/EFI/Linux/*.efi`. It may also add additional entries, for example a
-"Reboot into firmware" option. Optionally it may sort the menu based on the
-`sort-key`, `machine-id` and `version` fields, and possibly others. It uses the
-file name to identify specific items, for example in case it supports storing
-away default entry information somewhere. A boot loader should generally not
-modify these files.
+A _boot loader_ locates `$BOOT` and `$XBOOTLDR`, then simply reads all the files
+`$BOOT/loader/entries/*.conf` and `$XBOOTLDR/loader/entries/*.conf`, and populates
+its boot menu. On EFI, it then extends this with any unified kernel images
+found in `$BOOT/EFI/Linux/*.efi`. It may also add additional entries, for
```
medhefgo: XBOOTLDR too

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: Good points both. I updated the last commit.

medhefgo (approved):

poettering:
Hmmpf, I am seeing this only now, but afaics the spec is not completely add odds with bootctl where $BOOT is referenced all the time, and where it is either the XBOOTLDR partition where it exists, or the ESP otherwise.

This is now utterly confusing, it's all completely out of sync. We shouldn't redefine $BOOT like that in a spec, after all this time, and in particular not when it conflicts with our own code...

PR #23507, opened 2022-05-25
core/device: honor deserialized state and check device existence on switching-root

### LINKED ISSUE #23429: PR#23218 introduced a regression for devices with `db_persist` option

yuwata:
This reverts 75d7b5989f99125e52d5c0e5656fa1cd0fae2405.
Fixes #23429.
Replaces #23436, #23437, #23489.

### DIFF
--- a/src/core/device.c
+++ b/src/core/device.c
@@ -163,14 +163,6 @@ static int device_coldplug(Unit *u) {
         assert(d);
         assert(d->state == DEVICE_DEAD);
 
-        /* First, let's put the deserialized state and found mask into effect, if we have it. */
-        if (d->deserialized_state < 0)
-                return 0;
-
-        Manager *m = u->manager;
-        DeviceFound found = d->deserialized_found;
-        DeviceState state = d->deserialized_state;
-
         /* On initial boot, switch-root, reload, reexecute, the following happen:
          * 1. MANAGER_IS_RUNNING() == false
          * 2. enumerate devices: manager_enumerate() -> device_enumerate()
@@ -188,34 +180,57 @@ static int device_coldplug(Unit *u) {
          *   Also, there is no deserialized device. Device units are (a) generated based on dependencies of
          *   other units, or (b) generated when uevents are received.
          *
-         * - On switch-root, the udev database may be cleared, except for devices with sticky bit, i.e.
-         *   OPTIONS="db_persist". Hence, almost no devices are enumerated in the step 2. However, in general,
-         *   we have several serialized devices. So, DEVICE_FOUND_UDEV bit in the deserialized_found must be
-         *   ignored, as udev rules in initramfs and the main system are often different. If the deserialized
-         *   state is DEVICE_PLUGGED, we need to downgrade it to DEVICE_TENTATIVE (or DEVICE_DEAD if nobody
-         *   sees the device). Unlike the other starting mode, Manager.honor_device_enumeration == false
-         *   (maybe, it is better to rename the flag) when device_coldplug() and device_catchup() are called.
-         *   Hence, let's conditionalize the operations by using the flag. After switch-root, systemd-udevd
-         *   will (re-)process all devices, and the Device.found and Device.state will be adjusted.
+         * - On switch-root (or reload/reexecute triggered soon after switch-root and systemd-udevd has not
+         *   finished to re-process devices yet, see issue #12953), the udev database may be cleared, except
+         *   for devices with sticky bit, i.e. OPTIONS="db_persist". Hence, almost no devices are enumerated
+         *   in the step 2. However, in general, we have several serialized devices. To make the devices not
+         *   enter dead state needlessly, we need to mask DEVICE_FOUND_UDEV bit in Device.enumerated_found.
+         *   See device_found_mask_on_catchup() for more details. After switch-root, systemd-udevd will
+         *   (re-)process all devices, and the Device.found and Device.state will be adjusted.
          *
          * - On reload or reexecute, we can trust enumerated_found, deserialized_found, and deserialized_state.
          *   Of course, deserialized parameters may be outdated, but the unit state can be adjusted later by
          *   device_catchup() or uevents. */
 
-        if (!m->honor_device_enumeration && !MANAGER_IS_USER(m)) {
-                found &= ~DEVICE_FOUND_UDEV; /* ignore DEVICE_FOUND_UDEV bit */
-                if (state == DEVICE_PLUGGED)
-                        state = DEVICE_TENTATIVE; /* downgrade state */
-                if (found == DEVICE_NOT_FOUND)
-                        state = DEVICE_DEAD; /* If nobody sees the device, downgrade more */
-        }
-
-        if (d->found == found && d->state == state)
+        /* First, let's put the deserialized state and found mask into effect, if we have it. */
+        if (d->deserialized_state < 0)
                 return 0;
 
-        d->found = found;
-        device_set_state(d, state);
+        if (d->found == d->deserialized_found && d->state == d->deserialized_state)
+                return 0;
+
+        d->found = d->deserialized_found;
+        device_set_state(d, d->deserialized_state);
         return 0;
+}
+
+static bool device_found_mask_on_catchup(Device *d) {
+        assert(d);
+
+        /* On switching root or PID1 is reloaded before the device is processed by the systemd-udevd, the
+         * database for the device may not exist, or the device itself may be removed from the sysfs. The
+         * function device_enumerate() cannot distinguish the two cases. Let's explicitly check if the
+         * device is still around. */
+
+        if (FLAGS_SET(d->enumerated_mask, DEVICE_FOUND_UDEV))
+                /* The device is already enumerated. Let's use the all bits in enumerated_found. */
+                return DEVICE_FOUND_MASK;
+
+        if (!d->path)
+                /* We have no way to check if the device exists. Assume the device is unplugged. */
+                return DEVICE_FOUND_MASK;
+
+        _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
+        if (sd_device_new_from_path(&dev, d->path) < 0)
+                /* The device does not exist any more. Note, device units generated from SYSTEMD_ALIAS=
+                 * without udev database are always assumed as removed, as we have no way to get original
+                 * syspath from the alias. */
+                return DEVICE_FOUND_MASK;
+
+        /* The device still exists in sysfs, but is not ready yet. In that case, we will receive a uevent
+         * for the device when the device is re-processed by systemd-udevd or unplugged. Let's ignore the
+         * DEVICE_FOUND_UDEV bit in enumerated_found now. */
+        return DEVICE_FOUND_MASK & ~DEVICE_FOUND_UDEV;
 }
 
 static void device_catchup(Unit *u) {
@@ -224,7 +239,7 @@ static void device_catchup(Unit *u) {
         assert(d);
 
         /* Second, let's update the state with the enumerated state */
-        device_update_found_one(d, d->enumerated_found, DEVICE_FOUND_MASK);
+        device_update_found_one(d, d->enumerated_found, device_found_mask_on_catchup(d));
 }
 
 static const struct {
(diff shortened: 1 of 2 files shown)


yuwata:
cc @mwilck, @fbuihuu, and @tanriol.

src/core/device.c
```diff
@@ -189,41 +182,64 @@ static int device_coldplug(Unit *u) {
...
+                return 0;
 
-        if (d->found == found && d->state == state)
+        if (d->found == d->deserialized_found && d->state == d->deserialized_state)
                 return 0;
 
-        d->found = found;
-        device_set_state(d, state);
+        d->found = d->deserialized_found;
+        device_set_state(d, d->deserialized_state);
```
mwilck, line 203: You'll call `device_set_state(DEVICE_PLUGGED)` here if `deserialized_state == DEVICE_PLUGGED`. This will cause #23208 to occur again.  Note that you also set `DEVICE_FOUND_UDEV` in `d->found`.

src/core/device.c
```diff
@@ -703,10 +737,12 @@ static void device_update_found_one(Device *d, DeviceFound found, DeviceFound ma
                 d->found = n;
 
                 device_found_changed(d, previous, n);
-        } else
+        } else {
                 /* We aren't running yet, let's apply the new mask to the shadow variable instead, which we'll apply as
                  * soon as we catch-up with the state. */
                 d->enumerated_found = (d->enumerated_found & ~mask) | (found & mask);
+                d->enumerated_mask |= mask;
```
mwilck, line 744: This code path is only used if `device_update_found_one()` is called while `MANAGER_IS_RUNNING()` is false. 

The `DEVICE_FOUND_UDEV` bit in `enumerated_mask` will be set in the call chain `device_enumerate()→device_update_found_by_sysfs()→device_update_found_one()` if and only if the device was encountered by `device_enumerate()`, in which case the bit will be set in `d->enumerated_found`, too.

Tracking `enumerated_mask` separately would only make sense if the `DEVICE_FOUND_UDEV` bit could be cleared in `enumerated_found` later in a different code path. I don't see how this could happen.

src/core/device.c
```diff
@@ -295,6 +311,9 @@ static int device_serialize(Unit *u, FILE *f, FDSet *fds) {
         assert(f);
         assert(fds);
 
+        if (d->path)
```
mwilck, line 314: I'd be grateful if we could separate out he attempts to solve #23208 and #23429 from the new device path tracking logic.

src/core/device.c
```diff
@@ -189,41 +182,64 @@ static int device_coldplug(Unit *u) {
...
+        /* On switching root, the database for the device may be cleared, or the device itself may be
+         * unplugged. The function device_enumerate() cannot distinguish the two cases. Let's explicitly
+         * check if the device is still around. */
+
+        Manager *m = UNIT(d)->manager;
+        if (m->honor_device_enumeration || MANAGER_IS_USER(m))
+                /* Not switching root. Let's use the all bits. */
+                return DEVICE_FOUND_MASK;
+
+        if (FLAGS_SET(d->enumerated_mask, DEVICE_FOUND_UDEV))
```
mwilck, line 215: AFAICS you might as well test for `FLAGS_SET(d->enumerated_found, DEVICE_FOUND_UDEV)` here. `enumerated_mask` is not needed. See my comment in `device_update_found_one()`.

src/core/device.c
```diff
@@ -189,41 +182,64 @@ static int device_coldplug(Unit *u) {
...
+        if (!d->path)
+                /* We have no way to check if the device exists. Assume the device is unplugged. */
+                return DEVICE_FOUND_MASK;
+
+        _cleanup_(sd_device_unrefp) sd_device *dev = NULL;
+        if (sd_device_new_from_path(&dev, d->path) < 0)
+                /* The device does not exist any more. */
+                return DEVICE_FOUND_MASK;
+
+        /* The device still exists, and we will receive uevent for the device when the device is re-processed
```
mwilck: IIUC we'd get here if `FLAGS_SET(d->enumerated_found, DEVICE_FOUND_UDEV)` is false (i.e. the device has been "forgotten" while switching root) but still exists in sysfs. [^1]

In this case you'll call `device_update_found_one()` with `DEVICE_FOUND_UDEV` cleared in the mask. In `device_found_changed()`, you'll end up setting the device state to `DEVICE_DEAD` because the previous `found` value has `DEVICE_FOUND_UDEV` set (see me remark for `device_coldplug()`), and the new `found` value has not.

Unless I'm completely on the wrong track, this is not what we want.

I like the basic idea to check device existence in sysfs before udev's coldplug processing occurs. But IMO we need to apply a different logic (roughly speaking, we should check sysfs if `serialized_found` indicates the devices is present but `enumerated_found` does not — I haven't thought it through yet).

[^1]: the device must have existed before switching root, otherwise we'd never know about it. While it can have been removed on the kernel side, udev can hardly have processed a REMOVE event yet at this point in time, when we're switching root.

mwilck (changes requested):
@yuwata, thanks for your efforts. See my remarks. I may be misunderstanding your intentions, but AFAICS the logic isn't correct yet. 

I strongly recommend that we analyze the problems mentioned in #23429 first before we come up with additional patches.

src/core/device.c
yuwata: > You'll call `device_set_state(DEVICE_PLUGGED)` here if `deserialized_state == DEVICE_PLUGGED`. Note that you also set `DEVICE_FOUND_UDEV` in `d->found`.

Right.

> This will cause #23208 to occur again. 

Ah, maybe. I will re-consider the issue.

src/core/device.c
yuwata: > This code path is only used if `device_update_found_one()` is called while `MANAGER_IS_RUNNING()` is false.

Yes.

> The `DEVICE_FOUND_UDEV` bit in `enumerated_mask` will be set in the call chain `device_enumerate()→device_update_found_by_sysfs()→device_update_found_one()` if and only if the device was encountered by `device_enumerate()`, in which case the bit will be set in `d->enumerated_found`, too.
> 
> Tracking `enumerated_mask` separately would only make sense if the `DEVICE_FOUND_UDEV` bit could be cleared in `enumerated_found` later in a different code path. I don't see how this could happen.

Currently, that's right. But in general, that DEVICE_FOUND_UDEV flag set in `xxx_found` means the device is ready, and it is more than the fact that the device exists. Here, I'd like to make it clear if the device has udev database or not.

src/core/device.c
yuwata: Unfortunately, I have no idea to split it. To check if the device exists, we need syspath or path to the device node, and they are not explicitly stored previously. Of course, in most cases, it is encoded in the unit name, but not always, as the syspath may be too long.

src/core/device.c
yuwata: See mt reply in the above. We can optimize it. But I cannot find any merit to do that, except for suppressing a tiny amount of memory usage.

src/core/device.c
mwilck: > Here, I'd like to make it clear if the device has udev database or not.

But that's exactly the meaning of  `FLAG_SET(d->enumerated_found, DEVICE_FOUND_UDEV)`, or what am I missing here?

src/core/device.c
mwilck: I find it confusing to introduce a redundant new variable. The code is difficult enough to understand. I guess it's a matter of taste.

src/core/device.c
yuwata: No. If `FLAG_SET(d->enumerated_found, DEVICE_FOUND_UDEV)` is true, then `device_is_ready()` for the device is true.
If the flag is not set, there are multiple possible situations.
- the device is removed from sysfs,
- the udev database for the device has been removed,
- the device is not ready for PID1, e.g. SYSTEMD_READY=0.

src/core/device.c
yuwata: > IIUC we'd get here if `FLAGS_SET(d->enumerated_found, DEVICE_FOUND_UDEV)` is false (i.e. the device has been "forgotten" while switching root) but still exists in sysfs.

Right.

> In this case you'll call `device_update_found_one()` with `DEVICE_FOUND_UDEV` cleared in the mask. In `device_found_changed()`, you'll end up setting the device state to `DEVICE_DEAD` because the previous `found` value has `DEVICE_FOUND_UDEV` set (see me remark for `device_coldplug()`), and the new `found` value has not.
>
> Unless I'm completely on the wrong track, this is not what we want.

No, new `found` still has the DEVICE_FOUND_UDEV flag, as the mask does not have the bit.

src/core/device.c
mwilck: > No. If `FLAG_SET(d->enumerated_found, DEVICE_FOUND_UDEV)` is true, then `device_is_ready()` for the device is true. If the flag is not set, there are multiple possible situations.
> 
>     * the device is removed from sysfs,

This is unrelated to `d->enumerated_found`, because udevd isn't running at this point in time. The device may have disappeared from sysfs, but udev wouldn't have noticed, so the device would still be in the udev db (provided it had the `db_persist` flag before switching root — otherwise the device hadn't been enuerated anyway).

>     * the udev database for the device has been removed,
>     * the device is not ready for PID1, e.g. SYSTEMD_READY=0.

Right, these two can't be distinguished, but AFAICS they wouldn't be distinguished with `enumerated_mask`, either??

src/core/device.c
yuwata: More seriously, I noticed that this re-introduces #12953. I will update this soon.

src/core/device.c
yuwata: > > No. If `FLAG_SET(d->enumerated_found, DEVICE_FOUND_UDEV)` is true, then `device_is_ready()` for the device is true. If the flag is not set, there are multiple possible situations.
> > ```
> > * the device is removed from sysfs,
> > ```
> 
> This is unrelated to `d->enumerated_found`, because udevd isn't running at this point in time.

This _is_ related.

> The device may have disappeared from sysfs, but udev wouldn't have noticed, so the device would still be in the udev db (provided it had the `db_persist` flag before switching root — otherwise the device hadn't been enuerated anyway).

`sd_device_enumerator` checks existence of both syspath and udev database. Hence, if a device is not enumerated, then it is removed from sysfs, or udev database is removed.

> > ```
> > * the udev database for the device has been removed,
> > * the device is not ready for PID1, e.g. SYSTEMD_READY=0.
> > ```
> 
> Right, these two can't be distinguished, but AFAICS they wouldn't be distinguished with `enumerated_mask`, either??

Right, ~~and this is why we need to check the existence of syspath in `device_found_mask_on_catchup()`.~~

src/core/device.c
mwilck: Right, sorry. I misread the code.

This means in the case described, you wouldn't clear the `DEVICE_FOUND_UDEV` bit. Thus in the case at hand, device state would only change if `DEVICE_FOUND_MOUNT` or `DEVICE_FOUND_SWAP` changed. Which basically can't happen because we wouldn't have (de)activated any mounts or swaps at the point in time when `enumerated_found` is initialized. IOW, the device state remains the same that had been set in `device_coldplug()`.

yuwata: > This means in the case described, you wouldn't clear the `DEVICE_FOUND_UDEV` bit. Thus in the case at hand, device state would only change if `DEVICE_FOUND_MOUNT` or `DEVICE_FOUND_SWAP` changed. Which basically can't happen because we wouldn't have (de)activated any mounts or swaps at the point in time when `enumerated_found` is initialized.

Right.

>  IOW, the device state remains the same that had been set in `device_coldplug()`.

No. `device_dispatch_io()` will drop the `DEVICE_FOUND_UDEV` flag on remove uevent or `SYSTEMD_READY=0` is set on add/change uevent.

src/core/device.c
yuwata: In the updated version, #12953 should be fixed again. But, I have no idea about fixing #23208 without breaking LUKS setup.

src/core/device.c
mwilck: Can that happen before `device_catchup()` is called? (Sorry for bothering you here. I think it's important we clarify everything).

yuwata: Unfortunately no. As mentioned in the above, I have no idea about fixing #23208 now...

yuwata: > (Sorry for bothering you here. I think it's important we clarify everything).

No problem. I am enjoying this whole discussion.

src/core/device.c
mwilck: I am still debugging. But you're generating code faster than me catching up :grinning:

src/core/device.c
yuwata: OK. I will drop `enumerated_mask` in the next push.

yuwata:
Still I like this approach, but I cannot find a way to fix #23208. Closing.

PR #23508, opened 2022-05-25
core/device: fix devlink handling 

yuwata:
~~After #23507.~~

### DIFF
--- a/src/core/device.c
+++ b/src/core/device.c
@@ -826,12 +826,12 @@ static void device_process_new(Manager *m, sd_device *dev, const char *sysfs) {
         /* Add the main unit named after the sysfs path. If this one fails, don't bother with the rest, as
          * this one shall be the main device unit the others just follow. (Compare with how
          * device_following() is implemented, see below, which looks for the sysfs device.) */
-        if (device_setup_unit(m, dev, sysfs, true) < 0)
+        if (device_setup_unit(m, dev, sysfs, /* main = */ true) < 0)
                 return;
 
         /* Add an additional unit for the device node */
         if (sd_device_get_devname(dev, &dn) >= 0)
-                (void) device_setup_unit(m, dev, dn, false);
+                (void) device_setup_unit(m, dev, dn, /* main = */ false);
 
         /* Add additional units for all explicitly configured aliases */
         r = sd_device_get_property_value(dev, "SYSTEMD_ALIAS", &alias);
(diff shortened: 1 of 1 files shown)


src/core/device.c
```diff
@@ -1001,27 +1102,49 @@ static int device_dispatch_io(sd_device_monitor *monitor, sd_device *dev, void *
...
+                }
 
+                /* Add additional units for all symlinks */
+                (void) device_setup_devlink_units(m, dev, &ready_devlinks);
+        }
+
+        if (!IN_SET(action, SD_DEVICE_ADD, SD_DEVICE_REMOVE, SD_DEVICE_MOVE))
+                device_propagate_reload_by_sysfs(m, sysfs);
+
+        STRV_FOREACH(devlink, ready_devlinks)
```
msekletar: This block is run for all event types. Is that what we want? It seems to me we don't want to propagate reloads on REMOVE events, right?

yuwata: To be frank, I have no idea about that. What does `manager_propagate_reload()` do?? It is difficult to understand transaction.c and job.c...

src/core/manager.h
```diff
@@ -406,6 +406,8 @@ struct Manager {
         char *switch_root;
         char *switch_root_init;
 
+        bool switching_root;
```
poettering, line 410: why have that as a separate boolean? Why isn't `m->objective == MANAGER_SWITCH_ROOT` enough?

yuwata: `MANAGER_SWITCH_ROOT` is set before the system is going to switch root. The `switching_root` flag is set the PID1 started after switching root. Maybe, the name of the boolean flag should be changed. Do you have any better idea?

poettering: might suffice to add a big comment here, explaining that this is set both before *and* after the switch root, in the old and the new PID 1 process.

src/core/device.c
```diff
@@ -719,9 +712,111 @@ static bool device_is_ready(sd_device *dev) {
...
+        assert(devlink);
+
+        if (sd_device_new_from_devname(&dev, devlink) < 0 || !device_is_ready(dev)) {
+                /* the devlink is already removed or not ready */
+                device_update_found_by_name(m, devlink, 0, DEVICE_FOUND_UDEV);
+                *ret = NULL;
+                return 0; /* not ready */
+        }
+
+        r = device_setup_unit(m, dev, devlink, false);
```
poettering: maybe do the `/* param_name= */` dance on the fourth agument

yuwata:
@poettering Thank you for the review. All comments are addressed. Upgrading the green label.

PR #23509, opened 2022-05-25
sysusers/basic: add soft-static entry for "mail" group

lucab:
This adds a sysusers.d entry for the "mail" group, using a configurable
soft-static GID.
It is typically involved in system-local mail spool setup, thus the group
name may need to be referenced by the systemd-tmpfiles fragment responsible
for creating `/var/spool/mail`.
This group does not seem to have a common static ID across different
distributions, thus they would need to configure the value at build-time
(e.g. Debian uses GID `8`, Fedora uses GID `12`).


keszybz:
Hmm, why should this be part of systemd, instead of the some package that provides mail capabilities? We generally include rules for users that we need ourselves (in particular because we have udev or tmpfiles rules which refer to those users), or because they are part of the basic infra needed to boot a system. Local mail is something of a fringe setup these days…

keszybz:
If the justification is "there are multiple mail implementations, exim4, postfix, sendmail, fetchmail, …, and there is no shared infrastructure between them, so let's add this here", this is good enough for me. Systemd shouldn't be a dumping ground, but we can carry some additional configuration when there is no good place anywhere else.

keszybz:
This shouldn't be a problem. Just configure the right uid with `-Dmail-gid` and it should be fine.

BTW, what other groups does this script generate?

mbiebl:
> BTW, what other groups does this script generate?

https://salsa.debian.org/debian/base-passwd/-/blob/master/group.master
https://salsa.debian.org/debian/base-passwd/-/blob/master/passwd.master

The resulting basic.conf looks like this:
```
$ cat /usr/lib/sysusers.d/basic.conf 
# generated from /usr/share/base-passwd/{passwd,group}.master
g adm        4     -
g tty        5     -
g disk       6     -
g man        12    -
g kmem       15    -
g dialout    20    -
g fax        21    -
g voice      22    -
g cdrom      24    -
g floppy     25    -
g tape       26    -
g sudo       27    -
g audio      29    -
g dip        30    -
g operator   37    -
g src        40    -
g shadow     42    -
g utmp       43    -
g video      44    -
g sasl       45    -
g plugdev    46    -
g staff      50    -
g games      60    -
g users      100   -
g nogroup    65534 -

u root       0       - /root                /bin/bash
u daemon     1       - /usr/sbin            /usr/sbin/nologin
u bin        2       - /bin                 /usr/sbin/nologin
u sys        3       - /dev                 /usr/sbin/nologin
u sync       4:65534 - /bin                 /bin/sync
u games      5:60    - /usr/games           /usr/sbin/nologin
u man        6:12    - /var/cache/man       /usr/sbin/nologin
u lp         7       - /var/spool/lpd       /usr/sbin/nologin
u mail       8       - /var/mail            /usr/sbin/nologin
u news       9       - /var/spool/news      /usr/sbin/nologin
u uucp       10      - /var/spool/uucp      /usr/sbin/nologin
u proxy      13      - /bin                 /usr/sbin/nologin
u www-data   33      - /var/www             /usr/sbin/nologin
u backup     34      - /var/backups         /usr/sbin/nologin
u list       38      - /var/list            /usr/sbin/nologin
u irc        39      - /run/ircd            /usr/sbin/nologin
u gnats      41      - /var/lib/gnats       /usr/sbin/nologin
u nobody     65534:65534 - /nonexistent         /usr/sbin/nologin
```

keszybz:
Wouldn't it make more sense for setup to provide a sysusers file with the appropriate contents? It already has the list, so we could convert it to some other format at build time.

travier:
I think that the `setup` RPM sysusers have to define all users and groups as it may be installed in containers where we won't always install systemd.

keszybz:
> I'm fine with bringing back the mail group to only live in the setup sysusers fragment, if that's the proposal.

Yep. I think it's fine to have setup.rpm keep the `passwd` and `group` files as they are, but just provide sysusers files in addition. They wouldn't be normally used for anything, because sysusers ignores users/groups that are already present in `/etc/{passwd,group}`.

Sysusers will complain if there are two definitions for the same user/group, and they are not exactly equal. Thus, if we have the same fields defined twice, we'd need to take care to have exactly the same values, down to the "GECOS" description. Fortunately it's not defined for groups, but only for users. I guess we'd want to use to make those consistent. Systemd uses "Nobody" but setup.rpm has "Kernel Overflow User". The uids/gids that are used should be already consistent, because the configuration for systemd is based on the same table that setup uses.

And I think we want to keep the definitions we already have in systemd, to stay self-contained. In particular you can just install systemd and some libraries into an empty chroot and this should be functional.

poettering:
I am not convinced we should carry stuff that only caters for something I think is somewhat legacy (i.e. a local MTA). I am pretty sure outside of actual mail servers this is a model we don't really should focus on from upstream. Sounds like something distros should carry, i.e. in the fedora case in setup.rpm.

lucab:
Thanks all for the feedback. I think I can gather that there is consensus on:
 1. not pushing more legacy/distro-specific entries in here (i.e. no `mail` and such)
 2. having distributions provide their own sysusers fragment for system users, on the side of systemd `basic.conf` and augmenting it
 3. being ok with having overlapping entries between those fragments, as long as we can manage to keep all the entries consistent
 
 Let me know if there is anything I got wrong up to here, otherwise I'll just close this.

lucab:
>  Thus, if we have the same fields defined twice, we'd need to take care to have exactly the same values, down to the "GECOS" description.

On this topic, I just realized that for the `root` user both Debian and Fedora do use "root" as the GECOS label.
Instead systemd has
https://github.com/systemd/systemd/blob/a52765a5504108fb1d89f3d79b639b0a20efbd5e/sysusers.d/basic.conf.in#L9

@keszybz in which direction do you prefer to fix this misalignment?

keszybz:
The general idea is to use the "GECOS" field to provide additional information for users who don't know the purpose of the user. So for example, `apache — Apache Web Server` is much better than `apache — Apache`, because if you don't know what apache stands for, you're going to be completely lost. And if you google for it, you're going to find info about the tribe, or Apache Open Office, or APA stocks.

That said, I don't think "Super User" is terribly useful. I would prefer something like "Administrator". It sounds like MS Windows, and that might not be a bad thing, because it'll make sense for a lot of newbie users.

> it has a different shell too.

Hmm. We use `/bin/sh` because that's safe and will always exist. But it's not very nice to use. I wonder if we should make `nss-systemd` and `sysusers.d` use `/bin/bash` instead if present. That'd be an improvement for usability…

PR #23512, opened 2022-05-25
boot: Add string functions

medhefgo:
The ultimate goal of this is to eventually drop gnu-efi in favor of native clang/gcc-mingw64 efi support. There are a lot of annoyances and sometimes bugs about gnu-efi functions and I am getting tired of working around those. By building with native support we would also have a much cleaner build altogether.
To do that, we need to stop depending on gnu-efi library functions first (from libefi.a). This is the first step to do just that.

Some design choices in this are:
 - All of these functions come with unit tests.
 - They are NULL pointer safe.
 - Use of 8/16 suffix. It helps reduce type confusion when reading code and lets us unit test them without heavy ifdefs.
 - The suffix-less ifdefs are only supposed to be used in fundamental code.
 - Certain standard headers such as `stddef.h` are okay to use in `-ffreestanding` code. I looked at what these headers do, but gcc docs says so too. This means that we can safely use standard provided types such as size_t as there is nothing particular special about them.
 - We (will) use `char` to represent `CHAR8`. The UEFI spec itself does not say anything about whether the type is signed or not, so the standard "maybe" signed char is perfectly fine. In fact, edk2 typdefs it to char (and so does the heavily patched fedora gnu-efi).
 - We (will) use `char16_t` to represent `CHAR16` for pretty much the same reasons. We also use `u""` in code to drive home that point. This also makes it less likely that the size difference vs `wchat_t` doesn't bite us in userspace unit tests.

Currently, there are a few cases where casting is necessary due to `CHAR8` being unsigned. Eventually, once most/all of gnu-efi functions have been replaced I plan to convert the rest of the code to standard types. At that point those casts will be dropped too.

Once `libefi.a` is dropped, I plan to also drop usage of gnu-efi headers in favor of our own (possibly even with no screaming in type and function names :). After that would be changing the way we build sd-boot.

One small benefit I got out of this already was detection of this due to a compiler warning: 2d5d72c62b3d7bfd4d0d3e59cf252852237aa10c
I already have a working printf implementation that would give us proper compile-time format string checking, which would be another great benefit once that is finished.

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -418,11 +418,11 @@ static CHAR16 *update_timeout_efivar(UINT32 *t, BOOLEAN inc) {
 
         switch (*t) {
         case TIMEOUT_UNSET:
-                return xstrdup(L"Menu timeout defined by configuration file.");
+                return xstrdup16(u"Menu timeout defined by configuration file.");
         case TIMEOUT_MENU_FORCE:
-                return xstrdup(L"Timeout disabled, menu will always be shown.");
+                return xstrdup16(u"Timeout disabled, menu will always be shown.");
         case TIMEOUT_MENU_HIDDEN:
-                return xstrdup(L"Menu disabled. Hold down key at bootup to show menu.");
+                return xstrdup16(u"Menu disabled. Hold down key at bootup to show menu.");
         default:
                 return xpool_print(L"Menu timeout set to %u s.", *t);
         }
@@ -861,7 +861,7 @@ static BOOLEAN menu_run(
                 case KEYPRESS(0, 0, 'H'):
                 case KEYPRESS(0, 0, '?'):
                         /* This must stay below 80 characters! Q/v/Ctrl+l/f deliberately not advertised. */
-                        status = xstrdup(L"(d)efault (t/T)timeout (e)dit (r/R)resolution (p)rint (h)elp");
+                        status = xstrdup16(u"(d)efault (t/T)timeout (e)dit (r/R)resolution (p)rint (h)elp");
                         break;
 
                 case KEYPRESS(0, 0, 'Q'):
@@ -873,13 +873,13 @@ static BOOLEAN menu_run(
                 case KEYPRESS(0, 0, 'D'):
                         if (config->idx_default_efivar != idx_highlight) {
                                 FreePool(config->entry_default_efivar);
-                                config->entry_default_efivar = xstrdup(config->entries[idx_highlight]->id);
+                                config->entry_default_efivar = xstrdup16(config->entries[idx_highlight]->id);
                                 config->idx_default_efivar = idx_highlight;
-                                status = xstrdup(L"Default boot entry selected.");
+                                status = xstrdup16(u"Default boot entry selected.");
                         } else {
                                 config->entry_default_efivar = mfree(config->entry_default_efivar);
                                 config->idx_default_efivar = IDX_INVALID;
-                                status = xstrdup(L"Default boot entry cleared.");
+                                status = xstrdup16(u"Default boot entry cleared.");
                         }
                         config->use_saved_entry_efivar = FALSE;
                         refresh = TRUE;
@@ -1360,13 +1360,13 @@ good:
         entry->tries_left = left;
         entry->tries_done = done;
 
-        entry->path = xstrdup(path);
-        entry->current_name = xstrdup(file);
+        entry->path = xstrdup16(path);
+        entry->current_name = xstrdup16(file);
 
         next_left = left <= 0 ? 0 : left - 1;
         next_done = done >= (UINTN) -2 ? (UINTN) -2 : done + 1;
 
-        prefix = xstrdup(file);
+        prefix = xstrdup16(file);
         prefix[i] = 0;
 
         entry->next_name = xpool_print(L"%s+%" PRIuN L"-%" PRIuN L"%s", prefix, next_left, next_done, strempty(suffix));
@@ -1547,7 +1547,7 @@ static void config_entry_add_type1(
                 return;
 
         entry->device = device;
-        entry->id = xstrdup(file);
+        entry->id = xstrdup16(file);
         strtolower16(entry->id);
 
         config_add_entry(config, entry);
@@ -1792,7 +1792,7 @@ static void config_title_generate(Config *config) {
         for (UINTN i = 0; i < config->entry_count; i++) {
                 assert(!config->entries[i]->title_show);
                 unique[i] = TRUE;
-                config->entries[i]->title_show = xstrdup(config->entries[i]->title ?: config->entries[i]->id);
+                config->entries[i]->title_show = xstrdup16(config->entries[i]->title ?: config->entries[i]->id);
         }
 
         if (entries_unique(config->entries, unique, config->entry_count))
@@ -1911,11 +1911,11 @@ static ConfigEntry *config_entry_add_loader_auto(
 
         ConfigEntry *entry = xnew(ConfigEntry, 1);
         *entry = (ConfigEntry) {
-                .id = xstrdup(id),
+                .id = xstrdup16(id),
                 .type = LOADER_AUTO,
-                .title = xstrdup(title),
+                .title = xstrdup16(title),
                 .device = device,
-                .loader = xstrdup(loader),
+                .loader = xstrdup16(loader),
                 .key = key,
                 .tries_done = UINTN_MAX,
                 .tries_left = UINTN_MAX,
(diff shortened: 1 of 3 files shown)


src/boot/efi/efi-string.c
```diff
@@ -0,0 +1,44 @@
...
+        STRNLEN_U;
+}
+
+size_t strlen8(const char *s) {
+        return strnlen8(s, SIZE_MAX);
+}
+
+size_t strlen16(const char16_t *s) {
+        return strnlen16(s, SIZE_MAX);
+}
```
yuwata: Maybe, these can be inline?

src/boot/efi/efi-string.c
```diff
@@ -42,3 +43,67 @@ size_t strlen8(const char *s) {
...
+
+#define STRNCASECMP_U(cmp_type, case)                \
+        ({                                           \
+                if (!s1 || !s2)                      \
+                        return CMP(s1, s2);          \
+                                                     \
+                while (n > 0) {                      \
+                        cmp_type c1 = *s1, c2 = *s2; \
+                        if (case) {                  \
```
yuwata: I prefer `case` -> `tolower` or something, as "case" is reserved. Of course, this is a macro. But still...

src/boot/efi/efi-string.c
```diff
@@ -42,3 +43,67 @@ size_t strlen8(const char *s) {
...
+        return TOLOWER(c);
+}
+
+#define STRNCASECMP_U(cmp_type, case)                \
+        ({                                           \
+                if (!s1 || !s2)                      \
+                        return CMP(s1, s2);          \
+                                                     \
+                while (n > 0) {                      \
+                        cmp_type c1 = *s1, c2 = *s2; \
```
yuwata: Maybe, `typeof` can be used?

src/boot/efi/efi-string.c
```diff
@@ -42,3 +43,67 @@ size_t strlen8(const char *s) {
...
+        STRNCASECMP_U(char16_t, false);
+}
+
+int strcmp8(const char *s1, const char *s2) {
+        return strncmp8(s1, s2, SIZE_MAX);
+}
+
+int strcmp16(const char16_t *s1, const char16_t *s2) {
+        return strncmp16(s1, s2, SIZE_MAX);
+}
```
yuwata, line 97: Inline?

src/boot/efi/efi-string.c
```diff
@@ -42,3 +43,67 @@ size_t strlen8(const char *s) {
...
+        STRNCASECMP_U(char16_t, true);
+}
+
+int strcasecmp8(const char *s1, const char *s2) {
+        return strncasecmp8(s1, s2, SIZE_MAX);
+}
+
+int strcasecmp16(const char16_t *s1, const char16_t *s2) {
+        return strncasecmp16(s1, s2, SIZE_MAX);
+}
```
yuwata: Inline?

src/boot/efi/boot.c
```diff
@@ -554,7 +554,7 @@ static void print_status(Config *config, CHAR16 *loaded_image_path) {
                     Print(L"  config entry: %" PRIuN L"/%" PRIuN L"\n", i + 1, config->entry_count);
                 ps_string(L"            id: %s\n", entry->id);
                 ps_string(L"         title: %s\n", entry->title);
-                ps_string(L"    title show: %s\n", streq_ptr(entry->title, entry->title_show) ? NULL : entry->title_show);
+                ps_string(L"    title show: %s\n", streq16(entry->title, entry->title_show) ? NULL : entry->title_show);
```
yuwata, line 557: Could you mention in the commit message that `streq16()` accepts NULL hence streq16() can be used as the replacement for `streq_ptr()`.

src/boot/efi/efi-string.c
```diff
@@ -151,3 +151,25 @@ char *strcpy8(char *restrict dest, const char *restrict src) {
...
 }
+
+#define STRCHR_U(ret_type)                           \
+        ({                                           \
+                if (!s)                              \
+                        return NULL;                 \
+                                                     \
+                while (*s) {                         \
+                        if (*s == c)                 \
+                                return (ret_type) s; \
```
yuwata: `ret_type` -> `typeof(c)*` ?

src/boot/efi/efi-string.h
```diff
@@ -5,6 +5,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <uchar.h>
+#include "macro-fundamental.h"
```
yuwata, line 8: Please insert a blank line.

src/boot/efi/efi-string.c
```diff
@@ -173,6 +182,37 @@ char16_t *strchr16(const char16_t *s, char16_t c) {
...
+        ({                                                \
+                if (!s)                                   \
+                        return NULL;                      \
+                                                          \
+                size_t len = _len;                        \
+                size_t size = len * sizeof(type);         \
+                                                          \
+                type *dup = xmalloc(size + sizeof(type)); \
+                efi_memcpy(dup, s, size);                 \
+                dup[len] = '\0';                          \
```
yuwata, line 160: Maybe `u'\0'` for `char16_t`??

src/boot/efi/efi-string.c
```diff
@@ -0,0 +1,44 @@
...
+ * userspace does not allow for this (strlen8(NULL) returns 0 like strlen_ptr(NULL) for example). To make it
+ * easier to tell in code which kind of string they work on, we use 8/16 suffixes. This also makes is easier
+ * to unit test them.
+ *
+ * To avoid repetition and because char/char16_t string functions are fundamentally the same, we implement
+ * them with macros. Note that these macros are to be only used within their respective function they
+ * implement and therefore do not care about parameter hygiene and will also use the function parameters
+ * without declaring them in the macro params. */
+
+#define STRNLEN_U                       \
```
yuwata: Of course, this is used only in defining functions below, but I prefer to declare macros in a function like form:
```
#define STRLEN(s) \
({ \
    typeof _s = (s); \
    if (!_s) \
        return 0; \
    ...
})
```

yuwata (changes requested):

src/boot/efi/efi-string.c
medhefgo: Doesn't matter because it gets promoted to `char16_t` (boot.c does basically the same for long char literals). Also, this needs to work for both `char` and `char16_t`.

src/boot/efi/efi-string.c
medhefgo: I don't think it adds much to the table. The `return` statements would still make this not a function-like macro. And reordering the control flow because statement expressions can only return something on the last line would make this way too ugly imho.

src/boot/efi/efi-string.c
medhefgo: Unfortunately, the `const` gets in the way. But I just realized that I can just promote to `int` anyways. :D

src/boot/efi/efi-string.c
medhefgo: And now I've side-stepped the matter by using `memset`. That way I can use `void*` and drop the `type` arg.

src/boot/efi/efi-string.c
```diff
@@ -0,0 +1,46 @@
...
+                        return 0;                       \
+                                                        \
+                size_t len = 0, _n = n;                 \
+                while ((until_nul || len < _n) && *s) { \
+                        s++;                            \
+                        len++;                          \
+                }                                       \
+                                                        \
+                return len;                             \
+        })
```
yuwata: Still I am not a fun of macros in this form.
How about to introduce a macro that declares a function like `DEFINE_CONFIG_PARSE_PTR()` in conf-parser.h?
```c
#define DEFINE_STRNLEN(function, type) \
    size_t function(const type *s, size_t n) { \
        size_t len = 0; \
        while ((n == SIZE_MAX || len < n) && s && *s) { \
            s++; \
            len++; \
        } \
        return len; \
    }

DEFINE_STRNLEN(strnlen8, char);
DEFINE_STRNLEN(strnlen16, char16_t);
static inline size_t strlen8(const char *s) {
        return strnlen8(s, SIZE_MAX);
}
static inline size_t strlen16(const char16_t *s) {
        return strnlen16(s, SIZE_MAX);
}
```

medhefgo: Ah yes, that looks way better. Dunno why my brain didn't go that route.

src/boot/efi/efi-string.c
```diff
@@ -35,3 +36,58 @@ size_t strlen8(const char *s) {
...
+
+#define TOLOWER(c) ((c >= 'A' && c <= 'Z') ? c + ('a' - 'A') : c)
+
+char tolower8(char c) {
+        return TOLOWER(c);
+}
+
+char16_t tolower16(char16_t c) {
+        return TOLOWER(c);
+}
```
yuwata: Are they used?? If not, please drop them.

src/boot/efi/efi-string.c
```diff
@@ -35,3 +36,58 @@ size_t strlen8(const char *s) {
 size_t strlen16(const char16_t *s) {
         return strnlen16(s, SIZE_MAX);
 }
+
+#define TOLOWER(c) ((c >= 'A' && c <= 'Z') ? c + ('a' - 'A') : c)
```
yuwata: I'd like to avoid double evaluation.

yuwata (commented):
Looks mostly good now for me.

src/boot/efi/efi-string.c
medhefgo: I mostly added them to unit test it. Though, I guess it's covered by strtolower…

yuwata (approved):
LGTM.

> I though of a type-safe way to do these. Compiler explorer tells me it compiles to pretty much the same code (-O1+) as the macro version. But I am not really sure which would be easier to stomach :/

Not strong opinion, but I prefer the current form.

PR #23513, opened 2022-05-25
bootctl: print version comparison when updating

keszybz:
(The first patch is also in the other PR. I moved it here to avoid a conflict later.)

### DIFF
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -718,10 +718,13 @@ static int version_check(int fd_from, const char *from, int fd_to, const char *t
                                         to);
 
         r = compare_version(a, b);
+        log_debug("Comparing versions: \"%s\" %s \"%s", a, comparison_operator(r), b);
         if (r < 0)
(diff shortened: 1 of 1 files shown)


bluca (approved):

src/fundamental/string-util-fundamental.h
```diff
@@ -76,6 +76,10 @@ static inline const sd_char *yes_no(sd_bool b) {
         return b ? STR_C("yes") : STR_C("no");
 }
 
+static inline const sd_char* comparison_operator(int result) {
```
medhefgo, line 79: Really belongs to string-util.h. Or do you intend to use it in efi code?

keszybz: The version comparison stuff is in this file… I expect that sooner or later it'll be useful also in the efi code. And since this is just an inline stub, it doesn't matter.

yuwata (approved):

PR #23516, opened 2022-05-25
namespace: create new root directories with the correct selinux labels

bluca:
Use the _label variants everywhere, so that the policy can be
taken into account for /run/systemd/unit-root/ when services
are instantiated.

Requested by the SELinux policy maintainers.

### DIFF
--- a/src/core/namespace.c
+++ b/src/core/namespace.c
@@ -892,7 +892,7 @@ add_symlink:
                      DEVNUM_FORMAT_VAL(st.st_rdev)) < 0)
                 return log_oom();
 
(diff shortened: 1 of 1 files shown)


poettering:
what#s with this btw? looks conceptually Ok to me.

bluca:
> what#s with this btw? looks conceptually Ok to me.

selinux devs are not sure it's the right thing in general and are discussing (and on vacation), so it's to be held back for now

poettering:
ping?

bluca:
looks like this is not happening, so closing for now

PR #23518, opened 2022-05-25
Implement --root option for systemd-hwdb query

enr0n:
I came across this using `systemd-hwdb` to test additional hwdb rules. Based on the help output of `systemd-hwdb`, I was expecting the `--root` flag to also work with the `query` verb:

```
$ systemd-hwdb -h
systemd-hwdb [OPTIONS...] COMMAND ...

Update or query the hardware database.

Commands:
  update          Update the hwdb database
  query MODALIAS  Query database and print result

Options:
  -h --help       Show this help
     --version    Show package version
  -s --strict     When updating, return non-zero exit value on any parsing error
     --usr        Generate in /lib/udev instead of /etc/udev
  -r --root=PATH  Alternative root path in the filesystem

See the systemd-hwdb(8) man page for details.
```

These patches implement that functionality by adding a `sd_hwdb_new_from_path` function which permits loading a hwdb from a specified path.

### DIFF
--- a/src/shared/hwdb-util.c
+++ b/src/shared/hwdb-util.c
@@ -650,14 +650,27 @@ int hwdb_update(const char *root, const char *hwdb_bin_dir, bool strict, bool co
         return r;
 }
 
-int hwdb_query(const char *modalias) {
+int hwdb_query(const char *modalias, const char *root) {
         _cleanup_(sd_hwdb_unrefp) sd_hwdb *hwdb = NULL;
-        const char *key, *value;
+        const char *key, *value, *p;
         int r;
 
         assert(modalias);
 
-        r = sd_hwdb_new(&hwdb);
+        if (!isempty(root))
+                NULSTR_FOREACH(p, hwdb_bin_paths) {
+                        _cleanup_free_ char *hwdb_bin = NULL;
+
+                        hwdb_bin = path_join(root, p);
+                        if (!hwdb_bin)
+                                return -ENOMEM;
+
+                        r = sd_hwdb_new_from_path(hwdb_bin, &hwdb);
+                        if (r >= 0)
+                                break;
+                }
+        else
+                r = sd_hwdb_new(&hwdb);
         if (r < 0)
                 return r;
 
--- a/src/hwdb/hwdb.c
+++ b/src/hwdb/hwdb.c
@@ -18,7 +18,7 @@ static const char *arg_root = NULL;
 static bool arg_strict = false;
 
 static int verb_query(int argc, char *argv[], void *userdata) {
-        return hwdb_query(argv[1]);
+        return hwdb_query(argv[1], arg_root);
 }
 
 static int verb_update(int argc, char *argv[], void *userdata) {
--- a/src/shared/hwdb-util.h
+++ b/src/shared/hwdb-util.h
@@ -7,4 +7,4 @@
 
 bool hwdb_validate(sd_hwdb *hwdb);
 int hwdb_update(const char *root, const char *hwdb_bin_dir, bool strict, bool compat);
-int hwdb_query(const char *modalias);
+int hwdb_query(const char *modalias, const char *root);
--- a/src/udev/udevadm-hwdb.c
+++ b/src/udev/udevadm-hwdb.c
@@ -95,7 +95,7 @@ int hwdb_main(int argc, char *argv[], void *userdata) {
         }
 
         if (arg_test)
-                return hwdb_query(arg_test);
+                return hwdb_query(arg_test, NULL);
 
         return 0;
 }


src/libsystemd/sd-hwdb/sd-hwdb.c
```diff
@@ -339,6 +346,16 @@ _public_ int sd_hwdb_new(sd_hwdb **ret) {
         return 0;
 }
 
+_public_ int sd_hwdb_new_from_path(sd_hwdb **ret, const char *path) {
```
bluca: nit: return parameters are usually last

src/libsystemd/sd-hwdb/sd-hwdb.c
```diff
@@ -294,19 +294,26 @@ _public_ int sd_hwdb_new(sd_hwdb **ret) {
...
-                if (errno != ENOENT)
-                        return log_debug_errno(errno, "Failed to open %s: %m", hwdb_bin_path);
-        }
+        /* find hwdb.bin in hwdb_bin_paths, or from an explicit path if provided */
+        if (!isempty(path)) {
+                log_debug("Trying to open \"%s\"...", path);
+                hwdb->f = fopen(path, "re");
+                if (!hwdb->f)
+                        return log_debug_errno(errno, "Failed to open %s: %m", path);
+        } else {
```
yuwata, line 303: Unnecessary `{` after else.

src/shared/hwdb-util.c
```diff
@@ -650,14 +650,27 @@ int hwdb_update(const char *root, const char *hwdb_bin_dir, bool strict, bool co
...
 
-        r = sd_hwdb_new(&hwdb);
+        if (!isempty(root))
+                NULSTR_FOREACH(p, hwdb_bin_paths) {
+                        hwdb_bin = path_join(root, p);
+                        if (!hwdb_bin)
+                                return -ENOMEM;
+
+                        r = sd_hwdb_new_from_path(&hwdb, hwdb_bin);
+                        if (r == 0)
```
yuwata: Please use `if (r >= 0)`.

src/libsystemd/sd-hwdb/sd-hwdb.c
enr0n: I actually moved the `if (!hwdb->f)` check below into this else block, so there is more than one statement.

yuwata: Oops. indeed.

enr0n:
I have pushed changes to address review comments. I also added one more commit for a build issue I noticed when writing the unit test.

src/libsystemd/libsystemd.sym
```diff
@@ -643,6 +643,7 @@ global:
         sd_hwdb_unref;
 
         sd_hwdb_new;
+        sd_hwdb_new_from_path;
```
yuwata: Please add v252 entry and add the function there. I.e., please append the following at the end of the file:
```
LIBSYSTEMD_252 {
global:
        sd_hwdb_new_from_path;
} LIBSYSTEMD_251;
```

man/sd_hwdb_new.xml
```diff
@@ -66,7 +76,8 @@
     <title>Return Value</title>
 
     <para>On success, <function>sd_hwdb_new()</function> returns a non-negative integer. On
-    failure, it returns a negative errno-style error code.</para>
+    failure, it returns a negative errno-style error code. The same is true for
+    <function>sd_hwdb_new_from_path()</function>.</para>
```
yuwata: How about the following?
```
On success, <function>sd_hwdb_new()</function> and <function>sd_hwdb_new_from_path()</function> return a non-negative integer.
```

src/shared/hwdb-util.c
```diff
@@ -650,14 +650,27 @@ int hwdb_update(const char *root, const char *hwdb_bin_dir, bool strict, bool co
...
+                        if (!hwdb_bin)
+                                return -ENOMEM;
+
+                        r = sd_hwdb_new_from_path(hwdb_bin, &hwdb);
+                        if (r >= 0)
+                                break;
+                }
+        else
+                r = sd_hwdb_new(&hwdb);
+
```
yuwata: Please drop the empty line. We usually do not insert empty line between function call and error check.

yuwata (changes requested):

man/sd_hwdb_new.xml
enr0n: That sounds better. Thanks.

src/libsystemd/libsystemd.sym
enr0n: Whoops, thank you.

enr0n:
I have pushed changes to address review comments. Thanks.

yuwata (approved):
LGTM. Thank you!

src/shared/hwdb-util.c
```diff
@@ -650,14 +650,26 @@ int hwdb_update(const char *root, const char *hwdb_bin_dir, bool strict, bool co
...
+        _cleanup_free_ char *hwdb_bin = NULL;
+        const char *key, *value, *p;
         int r;
 
         assert(modalias);
 
-        r = sd_hwdb_new(&hwdb);
+        if (!isempty(root))
+                NULSTR_FOREACH(p, hwdb_bin_paths) {
+                        hwdb_bin = path_join(root, p);
```
bluca, line 664: this leaks memory on each iteration, the delcaration of hwdb_bin needs to be moved to this block

src/test/test-sd-hwdb.c
```diff
@@ -1,10 +1,12 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
 #include "sd-hwdb.h"
+#include "hwdb-internal.h"
```
bluca: sort this in the group below

src/shared/hwdb-util.c
enr0n: Good catch, thanks. Fixed in the latest push.

bluca (approved):

PR #23529, opened 2022-05-26
Don't linebreak after each initrd in kernel-install verbose mode

nabijaczleweli:
https://github.com/systemd/systemd/commit/76b1274a5cb54acaa4a0f0c2e570d751f9067c06#r74618476; @keszybz


keszybz:
Thanks, this seems to do the trick.

medhefgo:
I think this is causing a regression in stable: #23681

PR #23519, opened 2022-05-25
test: add coverage for #23481

mrc0mmand:
Provide some coverage for systemd/systemd#23481.

Without 794da5a:
```
[   34.730815] testsuite-29.sh[600]: + portablectl detach --now --runtime --enable /tmp/rootdir minimal-app0
         Stopping minimal-app0-foo.service...
[  OK  ] Stopped minimal-app0-foo.service.
         Stopping minimal-app0.service...
[  OK  ] Stopped minimal-app0.service.
[   34.878050] testsuite-29.sh[1383]: =================================================================
[   34.878421] testsuite-29.sh[1383]: ==1383==ERROR: LeakSanitizer: detected memory leaks
[   34.878784] testsuite-29.sh[1383]: Direct leak of 48 byte(s) in 2 object(s) allocated from:
[   34.879174] testsuite-29.sh[1383]:     #0 0x7fdf9c8b0f8c in reallocarray (/lib64/libasan.so.6+0xaef8c)
[   34.879554] testsuite-29.sh[1383]:     #1 0x7fdf9b4270f5 in unit_file_changes_add ../src/shared/install.c:282
[   34.879926] testsuite-29.sh[1383]:     #2 0x7fdf9b2ad9e5 in bus_deserialize_and_dump_unit_file_changes ../src/shared/bus-unit-util.c:2688
[   34.880267] testsuite-29.sh[1383]:     #3 0x40bc27 in maybe_enable_disable ../src/portable/portablectl.c:582
[   34.880673] testsuite-29.sh[1383]:     #4 0x40ef56 in maybe_stop_disable ../src/portable/portablectl.c:810
[   34.881064] testsuite-29.sh[1383]:     #5 0x410bc2 in detach_image ../src/portable/portablectl.c:924
[   34.881493] testsuite-29.sh[1383]:     #6 0x7fdf9b5df424 in dispatch_verb ../src/shared/verbs.c:103
[   34.881953] testsuite-29.sh[1383]:     #7 0x41604a in run ../src/portable/portablectl.c:1427
[   34.882459] testsuite-29.sh[1383]:     #8 0x416106 in main ../src/portable/portablectl.c:1430
[   34.882947] testsuite-29.sh[1383]:     #9 0x7fdf99d5de8f in __libc_start_call_main (/lib64/libc.so.6+0x44e8f)
[   34.883368] testsuite-29.sh[1383]: Indirect leak of 104 byte(s) in 2 object(s) allocated from:
[...]

PR #23533, opened 2022-05-27
portable: remove drop-in configs

yuwata:
Replaces #23519.

### DIFF
--- a/test/units/testsuite-29.sh
+++ b/test/units/testsuite-29.sh
@@ -152,6 +152,14 @@ portablectl inspect --cat --extension app0 --extension app1 rootdir app0 app1 | 
 
 portablectl detach --now --runtime --extension /tmp/app0 --extension /tmp/app1 /tmp/rootdir app0 app1
 
+# Attempt to disable the app unit during detaching. Requires --copy=symlink to reproduce.
+# Provides coverage for https://github.com/systemd/systemd/issues/23481
+portablectl "${ARGS[@]}" attach --copy=symlink --now --runtime /tmp/rootdir minimal-app0
+portablectl detach --now --runtime --enable /tmp/rootdir minimal-app0
+# attach and detach again to check if all drop-in configs are removed even if the main unit files are removed
+portablectl "${ARGS[@]}" attach --copy=symlink --now --runtime /tmp/rootdir minimal-app0
+portablectl detach --now --runtime --enable /tmp/rootdir minimal-app0
+
 umount /tmp/rootdir
 umount /tmp/app0
 umount /tmp/app1


yuwata:
Or, maybe `DisableUnitFilesWithFlags` should also remove drop-in configs? @keszybz WDYT?

bluca:
I'm a little confused, why doesn't the unit file exist?

yuwata:
> I'm a little confused, why doesn't the unit file exist?

`portablectl --enable --runtime detach` disables units with `DisableUnitFilesWithFlags` bus method, and the method removes the symlinked unit but does not its drop-ins.

bluca:
> > I'm a little confused, why doesn't the unit file exist?
> 
> `portablectl --enable --runtime detach` disables units with `DisableUnitFilesWithFlags` bus method, and the method removes the symlinked unit but does not its drop-ins.

Something's not quite right - `DisableUnitFilesWithFlags` is from PID1 and should disable the units, but it's also removing the symlink of the unit itself. I don't think this should happen.

bluca:
It can be reproduced with systemctl, without involving portabled in any way:

```
$ cat <<EOF>/tmp/foo.service
[Service]
ExecStart=echo foo
[Install]
WantedBy=multi-user.target
EOF
$ sudo ln -s /tmp/foo.service /run/systemd/system/
$ sudo systemctl daemon-reload
$ sudo systemctl enable --runtime foo
Created symlink /run/systemd/system/multi-user.target.wants/foo.service → /tmp/foo.service.
$ ls -l /run/systemd/system/
total 8
lrwxrwxrwx 1 root root  16 May 27 13:30 foo.service -> /tmp/foo.service
drwxr-xr-x 2 root root  60 May 27 13:32 multi-user.target.wants
$ sudo systemctl disable --runtime foo
Removed /run/systemd/system/multi-user.target.wants/foo.service.
Removed /run/systemd/system/foo.service.
$ ls /run/systemd/system/
$
```

bluca:
> It can be reproduced with systemctl, without involving portabled in any way:
> 
> ```
> $ cat <<EOF>/tmp/foo.service
> [Service]
> ExecStart=echo foo
> [Install]
> WantedBy=multi-user.target
> EOF
> $ sudo ln -s /tmp/foo.service /run/systemd/system/
> $ sudo systemctl daemon-reload
> $ sudo systemctl enable --runtime foo
> Created symlink /run/systemd/system/multi-user.target.wants/foo.service → /tmp/foo.service.
> $ ls -l /run/systemd/system/
> total 8
> lrwxrwxrwx 1 root root  16 May 27 13:30 foo.service -> /tmp/foo.service
> drwxr-xr-x 2 root root  60 May 27 13:32 multi-user.target.wants
> $ sudo systemctl disable --runtime foo
> Removed /run/systemd/system/multi-user.target.wants/foo.service.
> Removed /run/systemd/system/foo.service.
> $ ls /run/systemd/system/
> $
> ```

@keszybz says this is working as intended - if so, can you add a comment that explains why this is done?

bluca:
> > It can be reproduced with systemctl, without involving portabled in any way:
> > ```
> > $ cat <<EOF>/tmp/foo.service
> > [Service]
> > ExecStart=echo foo
> > [Install]
> > WantedBy=multi-user.target
> > EOF
> > $ sudo ln -s /tmp/foo.service /run/systemd/system/
> > $ sudo systemctl daemon-reload
> > $ sudo systemctl enable --runtime foo
> > Created symlink /run/systemd/system/multi-user.target.wants/foo.service → /tmp/foo.service.
> > $ ls -l /run/systemd/system/
> > total 8
> > lrwxrwxrwx 1 root root  16 May 27 13:30 foo.service -> /tmp/foo.service
> > drwxr-xr-x 2 root root  60 May 27 13:32 multi-user.target.wants
> > $ sudo systemctl disable --runtime foo
> > Removed /run/systemd/system/multi-user.target.wants/foo.service.
> > Removed /run/systemd/system/foo.service.
> > $ ls /run/systemd/system/
> > $
> > ```
> 
> @keszybz says this is working as intended - if so, can you add a comment that explains why this is done?

I mean a comment next to the code change in portable.c, as it's not very obvious

src/portable/portable.c
```diff
@@ -1559,40 +1559,54 @@ int portable_detach(
...
-                if (set_contains(unit_files, de->d_name))
+                if (set_contains(unit_files, unit_name))
                         continue;
 
-                if (!IN_SET(de->d_type, DT_LNK, DT_REG))
+                if (dot ? !IN_SET(de->d_type, DT_LNK, DT_DIR) : !IN_SET(de->d_type, DT_LNK, DT_REG))
                         continue;
 
-                r = test_chroot_dropin(d, where, de->d_name, name_or_path, extension_image_paths, &marker);
+                if (dot && laccess(unit_name, F_OK) >= 0)
```
bluca: This is not necessary, no? This block only deals with the unit name, so whether it's sourced from the drop-in dir or the unit it doesn't change anything, it will be stored in the set anyway

src/portable/portable.c
```diff
@@ -1559,40 +1559,54 @@ int portable_detach(
...
                         return r;
                 if (r == 0)
                         continue;
 
-                r = unit_file_lookup_state(LOOKUP_SCOPE_SYSTEM, &paths, de->d_name, &state);
-                if (r < 0)
-                        return log_debug_errno(r, "Failed to determine unit file state of '%s': %m", de->d_name);
-                if (!IN_SET(state, UNIT_FILE_STATIC, UNIT_FILE_DISABLED, UNIT_FILE_LINKED, UNIT_FILE_RUNTIME, UNIT_FILE_LINKED_RUNTIME))
-                        return sd_bus_error_setf(error, BUS_ERROR_UNIT_EXISTS, "Unit file '%s' is in state '%s', can't detach.", de->d_name, unit_file_state_to_string(state));
+                if (!dot) {
```
bluca: this should happen regardless?

keszybz: Or alternatively, if the check cannot be done if we don't have the main file, maybe the check is not necessary at all?

yuwata: > Or alternatively, if the check cannot be done if we don't have the main file, maybe the check is not necessary at all?

Indeed. Dropped.

yuwata:
- The call of `unit_file_lookup_state()` is dropped.
- comment is added.

PTAL.

src/portable/portable.c
```diff
@@ -1559,40 +1559,46 @@ int portable_detach(
         }
 
         FOREACH_DIRENT(de, d, return log_debug_errno(errno, "Failed to enumerate '%s' directory: %m", where)) {
-                _cleanup_free_ char *marker = NULL;
-                UnitFileState state;
+                _cleanup_free_ char *marker = NULL, *unit_name = NULL;
+                const char *dot;
+
+                /* "portablectl --enable detach" calls DisableUnitFilesWithFlags DBus method, and the main
```
bluca: This needs to specify "if it is symlinked via --copy=symlink", as that's the condition that triggers it

evverx:
@mrc0mmand @yuwata thanks! As far as I can see this PR also fixes another issue where images with empty unit files can be attached but can't be detached :-) Honestly the API is bogus (because more or less complex stuff like keeping track of images and extensions was offloaded onto clients (and documented :-)) so I'm not sure any new tests can fix those design decisions and make it any better.

keszybz:
LGTM.

src/portable/portable.c
bluca: @yuwata this was not addressed?

yuwata: Ouch, I mistakenly mention --runtime instead of --copy. Will post a follow-up.

yuwata: #23571.

Issue #23520, opened 2022-05-25
systemd-login disturbs all ptys on suspend


cgzones:
**systemd version the issue has been seen with**
 > 251

**Used distribution**
 > Debian sid

**Linux kernel version used** (`uname -a`)
 > `Linux debianHome 5.17.0-2-amd64 #1 SMP PREEMPT Debian 5.17.6-1 (2022-05-14) x86_64 GNU/Linux`

**CPU architecture issue was seen on**
 > amd64

**Expected behaviour you didn't see**
 > ptys (e.g. Konsole or tmux) unchanged after suspend

**Unexpected behaviour you saw**
 >
 ```

Broadcast message from christian@debianHome on tty1 (Wed 2022-05-25 23:49:31 CEST):

The system is going down for suspend NOW!
```
> in every terminal

**Steps to reproduce the problem**
 > suspend the system, e.g. on a laptop close and reopen the lid

Please either disable broadcast messages for suspend or add a configuration switch to disable it.

Also reported at https://www.spinics.net/lists/fedora-devel/msg300498.html

Related commit: https://github.com/systemd/systemd/commit/cdf370626f

AdrianusWest:
I join in. It's a madhouse with these constant notifications!

hl037:
+1 on archlinux

VorpalBlade:
If this was only on shutdown I could understand it. However it is nonsensical and annoying to send these messages every time I suspend or hibernate my laptop by closing the lid. If I close the lid I'm *fully aware* that it is going to be suspend. I in fact *expect* it to suspend. But it messes up my terminals when I get back.

hl037:
I think it's intended for servers or nfs to advise users of a shared system that it is going down, but they forgot to provide a mean of configuration...

arthurflor23:
+1 on arch (so annoying)

ffernand:
Also on Arch & some have suggested adding `mesg n` to bash profile.  But it's a no-go for me. Even with `mesg n`, whether on shutdown or sleep, all open terminals are spammed with broadcast messages.

ItIsUday:
I believe it's because of this commit https://github.com/systemd/systemd/commit/cdf370626f08ed509a5dde9d5618eed29d625032

AdrianusWest:
> I believe it's because of this commit [cdf3706](https://github.com/systemd/systemd/commit/cdf370626f08ed509a5dde9d5618eed29d625032)


I don't think. When disabled, the messages remain.

[...]

PR #23538, opened 2022-05-27
login: do not issue wall messages on ptys for suspend and hibernate

cgzones:
Fixes: #23520

### DIFF
--- a/src/login/logind-utmp.c
+++ b/src/login/logind-utmp.c
@@ -47,6 +47,15 @@ bool logind_wall_tty_filter(const char *tty, void *userdata) {
         const char *p;
 
         assert(m);
+        assert(m->scheduled_shutdown_action);
+
+        if (IN_SET(m->scheduled_shutdown_action->handle,
+                HANDLE_SUSPEND,
+                HANDLE_HIBERNATE,
+                HANDLE_HYBRID_SLEEP,
+                HANDLE_SUSPEND_THEN_HIBERNATE) &&
+                path_startswith(tty, "/dev/pts/"))
+                return false;
 
         if (!m->scheduled_shutdown_tty)
                 return true;


src/login/logind-utmp.c
```diff
@@ -47,6 +47,15 @@ bool logind_wall_tty_filter(const char *tty, void *userdata) {
...
 
         assert(m);
+        assert(m->scheduled_shutdown_action);
+
+        if (IN_SET(m->scheduled_shutdown_action->handle,
+                HANDLE_SUSPEND,
+                HANDLE_HIBERNATE,
+                HANDLE_HYBRID_SLEEP,
+                HANDLE_SUSPEND_THEN_HIBERNATE) &&
+                path_startswith(tty, "/dev/pts/"))
```
keszybz, line 57: Indentation is borked, which makes this hard to read: please align `HANDLE_*` to the same column as `m->`.

keszybz (commented):
Hmmm. Hmmm. I had to reread all the code that sets and uses `scheduled_shutdown_tty` to figure out what is going on here. IIUC, the idea is that we record the tty where the command to shut down / hibernate / etc. originated. When doing the wall message, we loop over all utmp entries, and send the message to terminals where users are logged in, except for the originating tty.

This patch "fixes" the issue by dropping the notifications on pseudo-ttys, but I don't think this makes much sense. In particular, it'll drop notifications for ssh sessions, which would seem to be quite useful and which we should keep.

It seems that the patch which enabled wall messages by default exposed a latent bug. People weren't generally requesting notifications, so nobody noticed that they were being delivered improperly. But if somebody requested them, the same issue would appear.

The question instead is: why do people have their pseudoterminal sessions registered in utmp? On my machine, `utmpdump </var/run/utmp` yields `tty2` for my graphical session, and `pts/12` for an ssh login. So I see the wall message exactly once, on the ssh pty.

pieniacy (commented):
IMO any hardcoded mechanism is not a solution to this problem, this should be configurable at least with respect to suspend/hibernate and shutdown/reboot.

See https://github.com/systemd/systemd/issues/23520#issuecomment-1140206521

keszybz:
After looking at this even more, I don't think we should take this approach. As described above, this patch hides the issue, but it makes the logic even more accidental. I think it's fair to say that the logic of using utmp lists for wall messages has severe problems. I think we should revert https://github.com/systemd/systemd/commit/cdf370626f08ed509a5dde9d5618eed29d625032 instead. I'll submit a PR with a proper justification there.

PR #23540, opened 2022-05-27
meson: Build header tests with -pedantic

medhefgo:
By using __extension__, we can silence pedantic errors we cannot or
do not want to fix.

This in particular silences enum values being outside of int range,
variadic macros, long long being C99 and a few public header functions
having C99 bool args.


medhefgo:
I just learned about this and though this would be a good way to follow up on (#23246) and future proof things. I'll leave it to others to decide if this is really worth the effort and slight uglyfication of code.

Edu4rdSHL (approved):
LGTM. The failing test is unrelated to the actual change.

src/systemd/sd-bus.h
```diff
@@ -94,15 +94,15 @@ enum {
...
+__extension__ enum {
         SD_BUS_NAME_REPLACE_EXISTING  = 1ULL << 0,
         SD_BUS_NAME_ALLOW_REPLACEMENT = 1ULL << 1,
         SD_BUS_NAME_QUEUE             = 1ULL << 2
 };
 
-enum {
+__extension__ enum {
         SD_BUS_MESSAGE_DUMP_WITH_HEADER  = 1ULL << 0,
-        SD_BUS_MESSAGE_DUMP_SUBTREE_ONLY = 1ULL << 1,
```
yuwata, line 105: We usually append a comma also in the last entry, to make diffs smaller when we extend enums. Why do you want to drop them?

medhefgo: These are not allowed with `-ansi`, which the public headers are supposed to target.

yuwata (approved):

PR #23542, opened 2022-05-28
meson: Document why -Wimplicit-fallthrough is not used with clang

medhefgo:
Btw, what's our minimum supported gcc version? Readme only has something to say about clang.


yuwata (approved):

thom311:
follow-up at https://github.com/systemd/systemd/pull/23926

PR #23543, opened 2022-05-28
Move homectl and userdbctl to bindir

mbiebl:
Those binaries aren't needed during early boot.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2571,8 +2571,7 @@ if conf.get('ENABLE_USERDB') == 1
                 link_with : [libshared],
                 dependencies : [threads],
                 install_rpath : rootlibexecdir,
-                install : true,
-                install_dir : rootbindir)
+                install : true)
 endif
 
 if conf.get('ENABLE_HOMED') == 1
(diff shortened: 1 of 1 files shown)


mbiebl:
Seems the existing docs/tests already expect those binaries to be in /usr/bin: 
```
test/units/testsuite-46.sh:if ! test -x /usr/bin/homectl ; then
man/userdbctl.xml:AuthorizedKeysCommand /usr/bin/userdbctl ssh-authorized-keys %u
man/userdbctl.xml:AuthorizedKeysCommand /usr/bin/userdbctl ssh-authorized-keys %u --chain /usr/bin/othertool %u
```

Edu4rdSHL:
What are the advantages of doing this? What's the point?

keszybz:
We can merge this for now. Hopefully we can start drop the split soon.

keszybz:
> What are the advantages of doing this? What's the point?

Some ancient systems need this, nobody else understands it ;)

Edu4rdSHL:
I would have preferred that docs/tests have been updated to reflect the change before the merge.

mbiebl:
> I would have preferred that docs/tests have been updated to reflect the change before the merge.

The existing tests and documentation expect the binaries to be in `/usr/bin`, but we were installing the binaries to `/bin`.
On a non-merged system it doesn't matter that much but on an unmerged system this caused the above to be silently skipped.

Instead of updating the documentation and tests to use `/bin/` I figured it was preferable to install the binaries to `/usr/bin`.
Sooner or later we would be doing that anyway.

PR #23545, opened 2022-05-28
clang-format: Adjust style of pointers

medhefgo:
We gernerally have our pointers aligned to the right. SpaceAroundPointerQualifiers
makes sure "const char * const" is formatted nicely.


medhefgo:
While I would like to add `PPIndentWidth: 2`, it seems that complete borks multi-line `#defines` :crying_cat_face:

Edu4rdSHL (approved):

yuwata (approved):

PR #23548, opened 2022-05-28
rpm: adjust lua trigger for udevadm and use use rpm.execute()

tpgxyz:
Hi,
this is trivial to adjust test for `/run/udev/control` then run udevadm triggers.


keszybz:
LGTM.

tpgxyz:
Thanks

PR #23555, opened 2022-05-30
Revert "logind: enable wall messages by default"

### LINKED ISSUE #23520: systemd-login disturbs all ptys on suspend

keszybz:
This reverts commit cdf370626f08ed509a5dde9d5618eed29d625032.

The motivation for that change was reasonable: users were asked to authorize
two times when doing a commandline operation. This was certainly annoying, and
it seemed we can fix this nicely by flipping the default. But this clearly
causes problems, and seems to be a fix in the wrong place.

As described in #23520, users are getting unexpected wall messages. Xterm,
urxvt, (anything using libutempter?), and tmux (in some configurations),
register local pty sessions in utmp. The information in utmp is not enough to
distinguish them from e.g. ssh logins. So we can't guess if the wall messages
should go to those sessions after the fact. We should instead honour the
configuration that was passed by the caller.

Above I wrote that the fix is in the wrong place: on the commandline, it often
make sense to try to be smart in some generic command, and we can change some
details if the general intent of the command remains unchanged. But the dbus
API is an API like any other, and it should remain stable including the meaning
of various parameters and their defaults, and we shouldn't try to remove the
need for an additional call if it makes the call ambiguous. Thus, instead of
changing the default, we should find a different way to avoid the double polkit
auth, e.g. by adding a new call that sets the wall message too in one step.

Fixes #23520.

### DIFF
--- a/src/login/logind.c
+++ b/src/login/logind.c
@@ -54,7 +54,6 @@ static int manager_new(Manager **ret) {
         *m = (Manager) {
                 .console_active_fd = -1,
                 .reserve_vt_fd = -1,
-                .enable_wall_messages = true,
                 .idle_action_not_before_usec = now(CLOCK_MONOTONIC),
         };
 


yuwata (approved):
LGTM.

mrc0mmand (approved):

mbiebl:
As mentioned elsewhere, with that commit reverted I still got wall messages on suspend on the console.

yuwata:
Tentatively drops the green label. See Dropped pin
https://maps.app.goo.gl/BwcuowS2gbjuqzFJA

mbiebl:
> As mentioned elsewhere, with that commit reverted I still got wall messages on suspend on the console.

@keszybz maybe I was doing something wrong. Have you verified that the wall messages are gone for you?

mrc0mmand:
I think reverting the commit will also break `TEST-69-SHUTDOWN`, which might need some tweaking (and possibly would benefit for adding a test case for suspend as well).

yuwata:
Also see the suggestion https://github.com/systemd/systemd/issues/23520#issuecomment-1140991757.

mrc0mmand:
And when at it, let's bring @lnussel into the loop as well.

lnussel:
hmm, hmm, hmm :thinking:  I still think it makes sense to send the notification on direct dbus calls. Independent of any polkit stuff, that's just a side effect.
Whether or not the feature turns out annoying  on a desktop depends on a combination of distro, DE and terminal I guess. XFCE on openSUSE seems to be entirely silent. xterm has `+ut` resp `-ut` options to control utmp registration (on openSUSE default on). KDE even registers a pty in  utmp to display popups with the message.
Assuming the field is reliable maybe logind could look at the `ut_host` field to be smarter on where to send the notification. Eg not send to local logins if the triggering terminal was not remote, at least for suspend like stuff.

keszybz:
I was trying to wrap my head when exactly the login message is emitted, and after looking at this for a few hours, I'm ever more confused. But looking at this again, I think the revert doesn't really make much sense. The state of EnableWallMessages is not reset after a call, so if you e.g. hibernate and then resume,  or schedule a shutdown and cancel it, the original "default" doesn't matter. 

> Assuming the field is reliable maybe logind could look at the `ut_host` field to be smarter on where to send the notification. Eg not send to local logins if the triggering terminal was not remote, at least for suspend like stuff.

I guess this could work. Can you submit a PR?

lnussel:
> I was trying to wrap my head when exactly the login message is emitted, and after looking at this for a few hours, I'm ever more confused. But looking at this again, I think the revert doesn't really make much sense. The state of EnableWallMessages is not reset after a call, so if you e.g. hibernate and then resume, or schedule a shutdown and cancel it, the original "default" doesn't matter.

Yeah, it's a rabbit hole. Better don't look too close unless you want to keep yourself busy for weeks, btdt :-) There's more fishy things in there. Scheduling and inhibit handling for example.

> > Assuming the field is reliable maybe logind could look at the `ut_host` field to be smarter on where to send the notification. Eg not send to local logins if the triggering terminal was not remote, at least for suspend like stuff.
> 
> I guess this could work. Can you submit a PR?

Sure. Probably won't be quick this week though.

keszybz:
> Sure. Probably won't be quick this week though.

Since people seem to be quite annoyed by this, I took a stab at this myself. I'm closing this one in favour of #23574.

Issue #19204, opened 2021-04-03
RFE: cryptenroll custom TPM2 PCR values


grawity:
**Is your feature request related to a problem? Please describe.**

I would like to use `systemd-cryptenroll --tpm2-pcrs=` against the PCR values I specify, instead of the ones currently read from the TPM.

**Describe the solution you'd like**

An option to provide the PCR values as a file, or a documented format to generate LUKS tokens that would be understood by systemd-cryptsetup-generator.

**Describe alternatives you've considered**

 - Using my existing non-systemd unlock script.
 - Trying to adapt my existing script to manually create a LUKS token that would be understood by systemd-cryptsetup.

**The systemd version you checked that didn't have the feature you are asking for**

v248

poettering:
Yeah, this make sense (and is already listed in the TODO file, actually)

poettering:
let's drop this from the milestone. I think this makes a ton of sense to have, but I don't think this should really marked for the milestone, I don't think any of the maintainers need this, and hence are unlikely to work on this. Would be delighted to review a patch for this howver, for example if #24597 is resuscitated.

flixman:
@poettering I have just forked systemd and I am trying to implement this functionality in https://github.com/flixman/systemd/pull/1. The idea is to reuse as much code as possible, so the specification of the registers would go like 
```
 --tpm2-pcrs=7+11:sha256=<hash>+14
```
I have made a number of changes triggered by the update of the signatures in some inner functions, but the idea is that the changes are transparent for methods that do not use them. I have added a couple of tests for the new specification to test-tpm2 and seems to go ok, but I need still to extend the tpm2_unseal and tpm2_make_luks2_json. Do you think you can give it a preliminary look, to assess if it goes in the direction you expect?

poettering:
@flixman please submit this as a PR on this repo, and mark it as draft. let's do reviews here, so that they are generally visible to anyone doing reviews here.

generally: please follow coding style. (https://systemd.io/CODING_STYLE/#formatting) We are usually not to keen on reviewing patches that don't even get these superficialities right.

And no fixed-size pre-allocation of data structures please. Also, needs to be somewhat hash func independent, i.e. cover for sha1 and sha256 reasonably. I think it would make sense to store the hash values in a `Hashmap` object, and make the values a structure that contains pcr nr, hash func, and he hash value and hash size. Then simply pass that around.

flixman:
[...]

PR #23556, opened 2022-05-30
cryptenroll: Allow specifying PCR values in a file

fergus-dall:
cryptenroll gains a new option --tpm2-pcr-file which contains PCR values to use instead of the current PCR values in the TPM. This allows a volume to be re-bound to an expected future system state, rather then just the current state.

Fixes: #19204

### DIFF
--- a/man/systemd-cryptenroll.xml
+++ b/man/systemd-cryptenroll.xml
@@ -321,6 +321,34 @@
       </varlistentry>
 
       <varlistentry>
+        <term><option>--tpm2-pcr-file=</option><replaceable>PATH</replaceable></term>
+
+        <listitem><para>When enrolling a TPM2 device, bind against the PCR values in a file insead of the
+        current PCR values in the TPM. The file should contain a JSON object like this:</para>
+
+        <programlisting>
+{
+    "tpm2-pcr-bank" : "sha256",
+    "tpm2-pcrs" : {
+        "0" : "0000000000000000000000000000000000000000000000000000000000000000",
+        "7" : "0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF"
+    }
+}
+        </programlisting>
+
+        <para>Field <literal>tpm2-pcr-bank</literal> must be <literal>sha1</literal> or
+        <literal>sha256</literal>.</para>
+        <para>Field <literal>tpm2-pcrs</literal> must be an object who's keys are the indices of PCR
+        registers, and who's values are (case insensitive) hex-strings of the desired PCR values.</para>
+        <para>Note that which PCRs to bind against is controlled by <option>--tpm2-pcrs</option>, not this
+        file. PCRs with an entry in the file that are not listed in <option>--tpm2-pcrs</option> are
+        ignored. Passing a file that lacks an entry for a PCR listed in <option>--tpm2-pcrs</option> is an
+        error. You cannot bind partially against the current PCR values and partially against fixed PCR
+        values by passing a file that lacks entries for some PCRs.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><option>--tpm2-with-pin=</option><replaceable>BOOL</replaceable></term>
 
         <listitem><para>When enrolling a TPM2 device, controls whether to require the user to enter a PIN
--- a/src/cryptenroll/cryptenroll-tpm2.c
+++ b/src/cryptenroll/cryptenroll-tpm2.c
@@ -7,6 +7,8 @@
 #include "hexdecoct.h"
 #include "json.h"
 #include "memory-util.h"
+#include "sha256.h"
+#include "stdio-util.h"
 #include "tpm2-util.h"
 
 static int search_policy_hash(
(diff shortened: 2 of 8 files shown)


Vogtinator:
Nice, this is pretty much exactly our use case as well.

What about using the same format for PCR values as generated/expected by tpm2-tools instead of a custom JSON format?

Vogtinator:
Currently it always uses the list of PCRs given by `--tpm2-pcrs=` and only complains if one of the listed PCRs is not included in the PCR file. IMO it's more intuitive to use all of the specified PCR values if `--tpm2-pcrs=` is not explicitly given. Or the parameter `--tpm2-pcr-file` should be renamed to something like `--tpm2-pcr-values`.

poettering:
So, we definitely want something like this, but my thinking was to integrate it in the existing PCR selection syntax. i.e. right now you can write

```
--tpm2-pcrs=3+4+9
```

I was thinking to extend this to allow this:

```
--tpm2-pcrs=3:0ef149998289474e4bb31813edda6ad7f3c991b2d8dec6e8fe4db7a1f039f2d1+4:
7620558a01660aa12000dd62765e4ba03dfca8a39305592a0f069eeb8f8dd130+9:581bc875fc72a0f2e5aa508488e0ae44b18bfdebad2799c529c3d6e4661029c4
```

i.e. allow optionally suffixing each specified PCR index with a colon and a fixed hashsum, which is then is used in place of what is currently measured in the TPM.

And then, extend this futhre like this:

```
--tpm2-pcrs=3+4:/boot/vmlinux+9
```

which would mean: "use current TPM values for PCR `#3` and PCR `#9`, and hash the `/boot/vmlinux` file for PCR `#4`".

i.e. allow specified file system paths of files to automatically hash and use as PCR values. After all most of the time that's what you want: have the PCR values tested against some specific files you install. This would have to be implemented carefully, so that one can specify multiple files, and the right thing happens, i.e.

```
--tpm2-pcrs=3+4:/boot/vmlinux+4:/boot/initrd+9
```

poettering:
Also, I think it's essential that we allow specification of *both* SHA1 *and* SHA256 values for the same PCR within the same string, because some systems do not support SHA256, and we still want to allow people to provide a single string that can then appropriately apply to both the SHA1 bank (if only htat is available) ot the SHA256 bank (otherwise). i.e. the enrollment should be able to receive both types of hash values, and then just ignore all values but the ones that apply to the local TPM hash support.

This would then probably mean that the syntax for `--tpm2-pcrs=` that takes literal values, should take a hash type prefix in some way, i.e.:

```
--tpm2-pcrs=4:sha1:efdc817b8da21651cf391b39a63e15fabe6247ad,4:sha256:581bc875fc72a0f2e5aa508488e0ae44b18bfdebad2799c529c3d6e4661029c4
```

You get the idea.

poettering:
I think we should probably define a PcrPolicy object in some way, that carries the set of hashes for all PCRs, for both banks. And then have a parser from the strings as I suggest into this structure, and a formatter to build the string from the structure again. The string parser should then maybe already talk to the TPM and read the current PCR values for the cases where neither the explicit hash value or a source file is specified. then pass that PCrPolicy around.

Vogtinator:
I just checked, tpm2-tools describes their format as YAML and it supports specifying multiple banks.
There's also sha384 and sha512 meanwhile (and sm3_256, which we could probably ignore) as well as some sha3 variants.

What I'd also like in the future is to support more policies like policyor to allow multiple sets of PCR values. That would complicate the interface even more...

aplanas:
> And then, extend this futhre like this:
> 
> ```
> --tpm2-pcrs=3+4:/boot/vmlinux+9
> ```
> 
> which would mean: "use current TPM values for PCR `#3` and PCR `#9`, and hash the `/boot/vmlinux` file for PCR `#4`".

That is a good idea! IIUC it will require:

* Read the event log
* Parse the binary format
* Track the extensions for this PCR number 
* Re-create extensions until /boot/vmlinux has been measured
* Hash the asset and extend and calculate the new extension
* Continue with the extensions in the event log

Also I guess a bank should be specified (sha1, sha256), or detected based on how the event logs gets extended.

aplanas:
> FWICT that would just take the value of 3 and 9 as-is and just use the sha1/sha256/... hash of the given file as value for PCR 4.
> The event log doesn't need to be read for that.

In a general case it should, for example `4:/boot/initrd` (in grub would be PCR#8) cannot be done directly. There is currently an extension of the kernel and the current initrd. In a general case all the extensions should be re-calculated, replaced the entry specified in the CLI, and continue the extension chain, all based on the information from the event log.

poettering:
> That is a good idea! IIUC it will require:

Oh, the cryptographic operation the PCR logic implements is dead-easy, I think we should just calculate this "dry", i.e. without consulting the TPM or the logs or anything like that. i.e. start from an assumed PCR value of zero (or possibly a specified one), then hash the files, and extend the values with that, via `PCR[n] = H(PCR[n] || DATA)`, as per TPM spec.

This will be immediately useful to bind policy to initrds, since new kernels measure all initrds to PCR 9, which is zero before at boot. A separate step would then be to also make this useful for binding policy to the kernel itself, via a minor addition to sd-stub: automatically measure the kernel itself to some PCR that is so far not used (11?), before it is invoked. 

Note that PCR measurements from grub are utter garbage, and I think we should just ignore them, as they measure not the boot script, but the execution path through the boot script. It's not really systematically reproducible, afaics. Hence I'd just ignore it…

Things become very simple if we have two PCRs where one strictly and only contains the kernel image and nothing else, and the other only the initrds, and nothing else.

poettering:
> I mean, how would you even specify which element precisely to replace? what do you use as "handle" to reference the right element in the chain?

I mean, let me say this clearly: I think grub's model of just measuring any rubbish is a bad idea, as it means enrollment must be happen *locally*, and is extremely brittle. If Grub is uses the rule is: unless you booted once you won't know the PCR values in place. But that's just awully limiting. Instead of trying to make that work we should focus on clean PCR measurements, where we know exactly how they came to be. Because in that case we can easily calculate them offline, systematically, and robustly. i.e. we know what the PCR values are going to be on any system, because they are always going to be the same wherever we boot.

Fix the *generation* of the PCRs, don't try to make the terrible soup of PCRs work when *consuming* them. It'll make you a lot happier.

And note that it's fine to measure things twice, i.e. have the kernel measure both into PCR 4 and into PCR 11. It's redundant of course, but it doesn't really hurt if it's helpful for compat.

aplanas:
I think that I agree on every point, but I have some questions (maybe answered in the TODO).

>  And then include these signatures of the PCR values in the kernel PE image

IIUC we need to store the signed policy hash, and the policy structs so we can re-create it, and later validated by the TPM, so we cannot store the PCR values in the PE image directly.

Also, I am wondering that even if I am wrong and the PCR values can be stored in the stub PE header, in the `systemd` model the `initrd` will also be attached in there (`EFI stub `+ `kernel` + `initrd` + signature), so the PCRs are for this `kernel` and this `initrd`.  Can this be extended for a model where the `initrd` is outside the `EFI stub` (one kernel / multiple initrd)?  Could be that the idea is to store multiple policies in the PE header?

aplanas:
@fergus-dall I will give a shot next week. Lets see.

PR #23558, opened 2022-05-30
Actually delay running of mount start jobs when /p/s/mountinfo is rate limited

msekletar:


src/core/unit.c
```diff
@@ -1868,6 +1868,10 @@ int unit_start(Unit *u) {
 
         assert(u);
 
+        /* Let's hold off running start jobs for mount units when /proc/self/mountinfo monitor is rate limited. */
+        if (u->type == UNIT_MOUNT && sd_event_source_is_ratelimited(u->manager->mount_event_source))
+                return -EAGAIN;
```
yuwata, line 1873: Maybe, it is better to introduce a function in unit vtable, e.g. `.can_start_early()`??

yuwata (approved):
LGTM. One minor suggestion.

src/core/unit.c
msekletar: I don't know if it is worth adding virtual function if we now that only mount unit type would provide an implementation.

yuwata: OK. Then let's merge this as is, and introduce the vtable entry later when necessary.

PR #23559, opened 2022-05-30
sd-device: Directly compare character instead of using strchr

benzea:
The strchr when removing trailing \n characters is always just reading
the single character (as there is a \0 character behind anyway). So,
change it to a simple comparison to make it a bit more obvious that the
string is always of length 1.

### DIFF
--- a/src/libsystemd/sd-device/sd-device.c
+++ b/src/libsystemd/sd-device/sd-device.c
@@ -2150,7 +2150,7 @@ _public_ int sd_device_get_sysattr_value(sd_device *device, const char *sysattr,
                         return r;
 
                 /* drop trailing newlines */
-                while (size > 0 && strchr(NEWLINE, value[--size]))
+                while (size > 0 && value[--size] == NEWLINE)
                         value[size] = '\0';
         }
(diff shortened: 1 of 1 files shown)


src/libsystemd/sd-device/sd-device.c
```diff
@@ -2150,7 +2150,7 @@ _public_ int sd_device_get_sysattr_value(sd_device *device, const char *sysattr,
                         return r;
 
                 /* drop trailing newlines */
-                while (size > 0 && strchr(NEWLINE, value[--size]))
+                while (size > 0 && value[--size] == NEWLINE)
```
yuwata, line 2153: NEWLINE is a string...

yuwata (changes requested):
I cannot follow the change. What issue do you want to fix?

src/libsystemd/sd-device/sd-device.c
benzea: Hah, yeah, I was stupid. tbh. I thought I had not opened the PR yet when I noticed …

PR #23561, opened 2022-05-30
cgroup-util: Properly handle conditions where cgroup.threads is empty  after SIGKILL but processes still remain

msizanoen1:
After sending a SIGKILL to a process, the process might disappear from `cgroup.threads` but still show up in `cgroup.procs` and still remains in the cgroup and cause migrating new processes to `Delegate=yes` cgroups to fail with `-EBUSY`. This is especially likely for heavyweight processes that consume more kernel CPU time to clean up.

Fix this by only returning 0 when both `cgroup.threads` and `cgroup.procs` are empty.

How to reproduce:
 - Make sure you are on kernel 5.17.11 or later with full kernel preemption enabled
 - Compile the following C code:
```c
#include <string.h>
#include <unistd.h>
#include <sys/mman.h>

int main(void) {
    char *mem = mmap(NULL, 1024 * 1024 * 1024, PROT_READ|PROT_WRITE, MAP_ANONYMOUS|MAP_PRIVATE, -1, 0);
    memset(mem, 0xFF, 1024 * 1024 * 1024);
    pause();
}
```
   This serves as an example of a heavyweight process that requires some kernel CPU time to clean up.
 - Run `systemd-run --user -p KillMode=none <binary path> && systemctl --no-block restart user@$(id -u).service`

You will get the following error messages in the system journal:
```
May 30 22:06:07 redacted systemd[1]: user@1000.service: Found left-over process 221030 (heavyprocess) in control group while starting unit. Ignoring.
May 30 22:06:07 redacted systemd[1]: This usually indicates unclean termination of a previous run, or service implementation deficiencies.
May 30 22:06:07 redacted systemd[1]: Starting user@1000.service - User Manager for UID 1000...
May 30 22:06:07 redacted systemd[221629]: user@1000.service: Failed to attach to cgroup /user.slice/user-1000.slice/user@1000.service: Device or resource busy
May 30 22:06:07 redacted systemd[221629]: user@1000.service: Failed at step CGROUP spawning /usr/lib/systemd/systemd: Device or resource busy
May 30 22:06:07 redacted systemd[1]: user@1000.service: Main process exited, code=exited, status=219/CGROUP
May 30 22:06:07 redacted systemd[1]: user@1000.service: Killing process 221030 (heavyprocess) with signal SIGKILL.
May 30 22:06:07 redacted systemd[1]: user@1000.service: Killing process 221030 (heavyprocess) with signal SIGKILL.
May 30 22:06:07 redacted systemd[1]: user@1000.service: Failed with result 'exit-code'.
May 30 22:06:07 redacted systemd[1]: user@1000.service: Unit process 221030 (heavyprocess) remains running after unit stopped.
May 30 22:06:07 redacted systemd[1]: Failed to start user@1000.service - User Manager for UID 1000.
```

### DIFF
--- a/src/basic/cgroup-util.c
+++ b/src/basic/cgroup-util.c
@@ -357,20 +357,29 @@ int cg_kill(
                 Set *s,
                 cg_kill_log_func_t log_kill,
                 void *userdata) {
-        int r;
+
+        int r, ret;
 
         r = cg_kill_items(controller, path, sig, flags, s, log_kill, userdata, "cgroup.procs");
         if (r < 0 || sig != SIGKILL)
                 return r;
 
+        ret = r;
+
         /* Only in case of killing with SIGKILL and when using cgroupsv2, kill remaining threads manually as
            a workaround for kernel bug. It was fixed in 5.2-rc5 (c03cd7738a83), backported to 4.19.66
            (4340d175b898) and 4.14.138 (feb6b123b7dd). */
         r = cg_unified_controller(controller);
-        if (r <= 0)
+        if (r < 0)
+                return r;
+        if (r == 0)
+                return ret;
+
+        r = cg_kill_items(controller, path, sig, flags, s, log_kill, userdata, "cgroup.threads");
+        if (r < 0)
                 return r;
 
-        return cg_kill_items(controller, path, sig, flags, s, log_kill, userdata, "cgroup.threads");
+        return r > 0 || ret > 0;
 }
 
 int cg_kill_kernel_sigkill(const char *controller, const char *path) {


src/basic/cgroup-util.c
```diff
@@ -357,20 +357,26 @@ int cg_kill(
...
                 cg_kill_log_func_t log_kill,
                 void *userdata) {
-        int r;
+        int r, r2;
 
         r = cg_kill_items(controller, path, sig, flags, s, log_kill, userdata, "cgroup.procs");
         if (r < 0 || sig != SIGKILL)
                 return r;
 
+        r2 = r;
```
yuwata: Not a matter, but I'd like to rename `r2` -> `ret`, as in `cg_kill_item()` or `cg_kill_recursive()` we use `ret`.

src/basic/cgroup-util.c
```diff
@@ -357,20 +357,26 @@ int cg_kill(
...
+
         /* Only in case of killing with SIGKILL and when using cgroupsv2, kill remaining threads manually as
            a workaround for kernel bug. It was fixed in 5.2-rc5 (c03cd7738a83), backported to 4.19.66
            (4340d175b898) and 4.14.138 (feb6b123b7dd). */
         r = cg_unified_controller(controller);
         if (r <= 0)
                 return r;
```
yuwata: This should be
```
if (r < 0)
        return r;
if (r == 0)
        return ret;
```

src/basic/cgroup-util.c
```diff
@@ -357,20 +357,26 @@ int cg_kill(
                 Set *s,
                 cg_kill_log_func_t log_kill,
                 void *userdata) {
-        int r;
+        int r, r2;
```
yuwata: This is not your fault, but could you insert a blank line after the function declaration?

yuwata (changes requested):
Several minor points.

src/basic/cgroup-util.c
```diff
@@ -357,20 +357,26 @@ int cg_kill(
...
         r = cg_unified_controller(controller);
-        if (r <= 0)
+        if (r < 0)
                 return r;
+        if (r == 0)
+                return ret;
 
-        return cg_kill_items(controller, path, sig, flags, s, log_kill, userdata, "cgroup.threads");
+        r = cg_kill_items(controller, path, sig, flags, s, log_kill, userdata, "cgroup.threads");
+        return r > 0 || ret > 0;
```
yuwata, line 382: Oh, you (mistakenly?) dropped the error check.
```
if (r < 0)
    return r;

return r > 0 || ret > 0;
```

yuwata (approved):
LGTM.

PR #23565, opened 2022-05-30
man: nspawn related wording improvements

bnf:
Main hickup while reading the man page was "makes use of the --boot which is not the default" (missing "option").
Also fixes and streamlines some other findings in the nspawn man page.


medhefgo (approved):

keszybz:
LGTM.

PR #23569, opened 2022-05-31
Don't spawn systemd --user instance for background (i.e. cron) sessions

msekletar:


src/login/logind-dbus.c
```diff
@@ -1305,7 +1305,7 @@ static int method_set_user_linger(sd_bus_message *message, void *userdata, sd_bu
                         return r;
 
                 if (manager_add_user_by_uid(m, uid, &u) >= 0)
-                        user_start(u);
+                        user_start(u, true);
```
yuwata: Recently, we usually add comment `/* need_user_instance = */` for boolean argument.

test/TEST-73-LOGIND/test.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+
+TEST_DESCRIPTION="Test various logind features"
```
yuwata: Please specify image name, e.g. `IMAGE_NAME="logind"`. Otherwise, the default image may be used if it exists, which may not contain required commands.

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,81 @@
...
+    pkill -u "$(id -u test)"
+    pkill crond
+    crontab -r -u test
+    userdel -r test
+}
+
+test_no_user_instance_for_cron() {
+    # We actually want to run error code path in case any of the needed binaries are not present.
+    # shellcheck disable=SC2015
+    command -v passwd && command -v useradd && command -v userdel && command -v crond && command -v crontab || {
```
yuwata: If `IMAGE_NAME=` is specified, then this line can be dropped.

src/login/logind-user.c
```diff
@@ -523,7 +525,11 @@ int user_stop(User *u, bool force) {
                         r = k;
         }
 
-        user_stop_service(u, force);
+        r = manager_unit_is_active(u->manager, u->service, &error);
+        if (r < 0)
+                log_debug_errno(r, "Failed to determine whether unit '%s' is active, ignoring: %s", u->service, bus_error_message(&error, r));
+        if (r != 0)
+                user_stop_service(u, force);
```
keszybz: Hmm, is this chunk needed at all? `manager_stop_unit()` calls `StopUnit`, which should be a noop if the unit is not running.

keszybz (changes requested):
I think this is reasonable. We *might* want to add a way to override this later on, but not starting `user@.service` for background sessions seems like a much better default.

Please adjust the text in pam_systemd(8). It currently says that `user@.service` will be started.

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,81 @@
...
+
+    return 0
+}
+
+: > /failed
+
+test_no_user_instance_for_cron
+
+rm /failed
+: > /testok
```
keszybz: Please drop the spaces after redirect op.

test/units/testsuite-73.sh
msekletar: I think it makes sense to keep it here. Running test outside of test suite (i.e. `make setup run clean`) should be supported I think (after all we do have systemd-tests sub-package in Fedora).

mbiebl:
Wouldn't this break code which makes use of a dbus user bus?

src/login/logind-user.c
msekletar: Actually, `StopUnit()` on non-existing unit returns `org.freedesktop.systemd1.NoSuchUnit` and I think we want to avoid that.

msekletar:
> Wouldn't this break code which makes use of a dbus user bus?

This is very good point, I haven't considered this trade-off before. User bus is launched on demand by user instance of systemd. For cron session there won't be any `systemd --user` so no `dbus-broker --scope=user` instance will be launched either.

I personally think this is OK trade-off for cron sessions. I'd say that if you want to run periodic actions in user context with full session you should enable lingering for the user and use timer units instead.

msekletar:
Updated. @yuwata @keszybz PTAL.

src/login/logind-session.c
```diff
@@ -701,7 +701,7 @@ int session_start(Session *s, sd_bus_message *properties, sd_bus_error *error) {
         if (s->started)
                 return 0;
 
-        r = user_start(s->user);
+        r = user_start(s->user, s->class != SESSION_BACKGROUND);
```
keszybz: /* need_user_instance= */

keszybz (commented):
```
I: Masking supporting services
'/var/tmp/systemd-test.pklcif/root/etc/systemd/system/systemd-hwdb-update.service' -> '/dev/null'
'/var/tmp/systemd-test.pklcif/root/etc/systemd/system/systemd-journal-catalog-update.service' -> '/dev/null'
'/var/tmp/systemd-test.pklcif/root/etc/systemd/system/systemd-networkd.service' -> '/dev/null'
'/var/tmp/systemd-test.pklcif/root/etc/systemd/system/systemd-networkd.socket' -> '/dev/null'
'/var/tmp/systemd-test.pklcif/root/etc/systemd/system/systemd-resolved.service' -> '/dev/null'
W: Failed to install 'crond'
I: Skipping program crond as it cannot be found and is flagged to be optional
TEST-73-LOGIND RUN: Test various logind features
```

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,81 @@
...
+    # shellcheck disable=SC2015
+    command -v passwd && command -v useradd && command -v userdel && command -v crond && command -v crontab || {
+        echo >&2 "Missing support for running tasks via crond under non-root user."
+        return 1
+    }
+
+    trap teardown_cron EXIT
+    setup_cron
+
+    [[ $(loginctl --no-legend list-sessions | grep -c test) -ge 1 ]] || {
```
mrc0mmand: Is there a reason why not to use "standard" `if` statements over the `[[ ... ]] || { ... }` expressions?

```suggestion
    if [[ $(loginctl --no-legend list-sessions | grep -c test) -lt 1 ]]; then
    ...
```

yuwata:
> > Wouldn't this break code which makes use of a dbus user bus?
> 
> This is very good point, I haven't considered this trade-off before. User bus is launched on demand by user instance of systemd. For cron session there won't be any `systemd --user` so no `dbus-broker --scope=user` instance will be launched either.
> 
> I personally think this is OK trade-off for cron sessions. I'd say that if you want to run periodic actions in user context with full session you should enable lingering for the user and use timer units instead.

Could you also mention that point in NEWS?

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+set -eux
+set -o pipefail
+
+setup_cron() {
+    # Setup test user and cron
+    useradd test
+    echo "test" | passwd --stdin test
```
yuwata: Is it necessary to set password for the user?

Note, IIRC, passwd in Ubuntu CI does not support `--stdin`.

test/TEST-73-LOGIND/test.sh
```diff
@@ -0,0 +1,15 @@
...
+IMAGE_NAME="logind"
+
+# shellcheck source=test/test-functions
+. "${TEST_BASE_DIR:?}/test-functions"
+
+test_append_files() {
+    image_install -o useradd userdel passwd crond crontab pkill
```
yuwata: Now `userdadd`, `userdel`, and `pkill` are included in the default image.
I do not think it is necessary to build new image just for `crond` and `crontab` (and `passwd` if it is really necessary).
How about to optionally install them in `install_basic_tools()`?

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,78 @@
...
+* * * * * /bin/sleep 65
+EOF
+    # Let's a bit more than one interval to make sure that cron session is started already
+    sleep 70
+}
+
+teardown_cron() {
+    set +e
+
+    pkill -11  -u "$(id -u test)"
```
yuwata: doubled space??

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,78 @@
...
+        echo >&2 "User instance state is unexpected, expected \"dead\", got \"$state\""
+        return 1
+    fi
+
+    return 0
+}
+
+: >/failed
+
+if command -v passwd && command -v useradd && command -v userdel && command -v crond && command -v crontab ; then
```
yuwata: check for useradd and userdel can be dropped.

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,78 @@
...
+    crond -s -n &
+    # Install crontab for the test user that runs sleep every minute. But let's sleep for
+    # 65 seconds to make sure there is overlap between two consecutive runs, i.e. we have
+    # always a cron session running.
+    crontab -u test - <<EOF
+RANDOM_DELAY=0
+* * * * * /bin/sleep 65
+EOF
+    # Let's a bit more than one interval to make sure that cron session is started already
+    sleep 70
```
yuwata: To reduce runtime, how about to run a loop to check at least one session is started?

yuwata (changes requested):
Could you merge the test with TEST-35?

test/units/testsuite-73.sh
msekletar: Setting password shouldn't be necessary.

test/units/testsuite-73.sh
msekletar: Ok

test/units/testsuite-73.sh
msekletar: Done!

test/units/testsuite-73.sh
msekletar: Fixed

test/TEST-73-LOGIND/test.sh
msekletar: Thanks for the suggestion!

yuwata (approved):
LGTM.

mrc0mmand:
I think this change just broke dfuzzer:

```
[  465.562946] testsuite-21.sh[360]: Bus: org.freedesktop.systemd1 (session)
[  465.565156] testsuite-21.sh[360]: + systemd-run --machine testuser@.host --user --pipe --wait -- dfuzzer -b 10000 -n org.freedesktop.systemd1
[  466.142828] testsuite-21.sh[1355]: Failed to start transient service unit: Transport endpoint is not connected
```

because no user daemon was spawned. How can this be remedied?

src/login/logind-user.c
keszybz: I still think it'd be better to call `user_stop_service` unconditionally, and check if `org.freedesktop.systemd1.NoSuchUnit` was returned and treat that as success.

src/login/logind-user.c
```diff
@@ -442,7 +442,7 @@ static int user_update_slice(User *u) {
         return 0;
 }
 
-int user_start(User *u) {
+int user_start(User *u, bool need_user_instance) {
```
keszybz: `need` → `want`? I think this expresses the meaning of this variable better.

keszybz:
> because no user daemon was spawned. How can this be remedied?

I guess it'll be a matter of spawning the user daemon explicitly.

src/core/execute.c
```diff
@@ -1213,6 +1213,10 @@ static int setup_pam(
 
                 if (getttyname_malloc(STDIN_FILENO, &q) >= 0)
                         tty = strjoina("/dev/", q);
+                else
+                        /* If everything else failed then let's just use value "systemd". This will cause that session
+                         * isn't going to be marked as "background" and user manager will be started. */
+                        tty = "systemd";
```
yuwata, line 1219: This sounds hackish for me...
At least, `pam_sm_open_session()` in pam_systemd.c needs to handle this gracefully.

src/core/execute.c
```diff
@@ -1213,6 +1213,10 @@ static int setup_pam(
...
         }
 
         if (tty) {
```
yuwata: This condition is not necessary, if the above "workaround" is approved.

test/units/testsuite-35.sh
```diff
@@ -360,6 +360,68 @@ EOF
...
+EOF
+    # Let's wait (at most one interval plus 10s to accomodate for slow machines) for at least one session of test user
+    timeout 70s bash -c "while true; do loginctl --no-legend list-sessions | awk '{ print \$3 }' | grep -q test && break || sleep 1 ; done"
+}
+
+teardown_cron() {
+    set +e
+    pkill -9 -u "$(id -u test)"
+    pkill crond
+    crontab -r -u test
```
yuwata, line 465: Maybe, `userdel test`??

test/units/testsuite-35.sh
```diff
@@ -360,6 +360,68 @@ EOF
...
+    crontab -u test - <<EOF
+RANDOM_DELAY=0
+* * * * * /bin/sleep 65
+EOF
+    # Let's wait (at most one interval plus 10s to accomodate for slow machines) for at least one session of test user
+    timeout 70s bash -c "while true; do loginctl --no-legend list-sessions | awk '{ print \$3 }' | grep -q test && break || sleep 1 ; done"
+}
+
+teardown_cron() {
+    set +e
```
yuwata: Please use subshell:
```
teardown_cron() {(
    set +ex

    ...
)}
```

yuwata (changes requested):

src/core/execute.c
msekletar: We have conflict of requirements. I don't want background sessions (of cron) to spawn user instance while at the same time are relying on (currently) background sessions created from sd_bus to spawn user manager so that connection to user bus can be established.

https://github.com/systemd/systemd/blob/3bdb60055e16776c8cc4d44f26abb5f2cbe197bf/src/libsystemd/sd-bus/sd-bus.c#L1567

I am essentially saying here that no PAM sessions created from systemd through `PAMName=` directive will be considered background (this was effectively the previous behaviour even though they were marked as background).

yuwata: yeah, I understand the motivation of this change. How about the following?
```diff
diff --git a/src/login/pam_systemd.c b/src/login/pam_systemd.c
index 2047dc1817..1519b25182 100644
--- a/src/login/pam_systemd.c
+++ b/src/login/pam_systemd.c
@@ -792,6 +792,10 @@ _public_ PAM_EXTERN int pam_sm_open_session(
                              * much later (this is because it doesn't know yet if it needs one at all, as whether to
                              * register a pty or not is negotiated much later in the protocol). */
 
+        } else if (streq(tty, "systemd")) {
+                if (isempty(class))
+                        class = "user";
+                tty = NULL;
         } else
                 /* Chop off leading /dev prefix that some clients specify, but others do not. */
                 tty = skip_dev_prefix(tty);
```

msekletar: This seems sensible. Thank you for the suggestion.

test/units/testsuite-35.sh
msekletar: Actually, I think curly brackets are redundant if we have subshell, i.e. ().

msekletar:
Updated. @yuwata @poettering PTAL.

yuwata (approved):
LGTM.

yuwata:
Please rebase.

yuwata:
This was previously all green. Merging.

src/core/execute.c
```diff
@@ -1213,13 +1213,15 @@ static int setup_pam(
 
                 if (getttyname_malloc(STDIN_FILENO, &q) >= 0)
                         tty = strjoina("/dev/", q);
+                else
+                        /* If everything else failed then let's just use value "systemd". This will cause that session
+                         * isn't going to be marked as "background" and user manager will be started. */
+                        tty = "systemd";
         }
```
poettering, line 1220: I don't follow here? PAMName= without a tty pretty much *is* background stuff? so why mark it specially? What do you have in mind that sets PAMName= without tty here that you don't want to affect?

poettering: (I am a bit conservative about adding more special somethings to the PAM tty field that encodes stuff that clearly is not about actual ttys)

src/login/logind-session.c
```diff
@@ -664,12 +664,12 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
                                 /* These two have StopWhenUnneeded= set, hence add a dep towards them */
                                 STRV_MAKE(s->user->runtime_dir_service,
-                                          s->user->service),
+                                          s->class != SESSION_BACKGROUND ? s->user->service : NULL),
                                 /* And order us after some more */
                                 STRV_MAKE("systemd-logind.service",
                                           "systemd-user-sessions.service",
                                           s->user->runtime_dir_service,
-                                          s->user->service),
+                                          s->class != SESSION_BACKGROUND ? s->user->service : NULL),
```
poettering, line 672: this one looks wrong.

in systemd ordering deps and requirement deps are unconditional. It might make sense to drop the requiement deps for some tasks, but i think the ordering should always be kept.

i.e. if something else pulls in user@.service anyway (for example: lingering is enabled for the user), then we should honour the order. And if it is not pulled in by anything the configured order has no effect. Hence we should just leave this in.

src/login/logind.c
```diff
@@ -1118,9 +1117,7 @@ static int manager_startup(Manager *m) {
         HASHMAP_FOREACH(seat, m->seats)
                 (void) seat_start(seat);
 
-        HASHMAP_FOREACH(user, m->users)
-                (void) user_start(user);
-
+        /* Users are started by respective sessions */
```
poettering, line 1120: this change is wrong. think about users with "lingering" turned on.

src/login/pam_systemd.c
```diff
@@ -790,7 +790,10 @@ _public_ PAM_EXTERN int pam_sm_open_session(
                              * does the PAM session registration early for new connections, and registers a pty only
                              * much later (this is because it doesn't know yet if it needs one at all, as whether to
                              * register a pty or not is negotiated much later in the protocol). */
-
+        } else if (streq(tty, "systemd")) {
+                if (isempty(class))
+                        class = "user";
+                tty = NULL;
```
poettering, line 796: so this doesn't make sense to me. see above. please elaborate

src/login/logind-user.c
```diff
@@ -442,7 +442,7 @@ static int user_update_slice(User *u) {
         return 0;
 }
 
-int user_start(User *u) {
+int user_start(User *u, bool want_user_instance) {
```
poettering, line 445: this is broken, because racy. Think of a "background" session that is long running (cron job backup script). Starts some time after boot. A bit later you log in interactively. Nothing will start the user@.service now. You'll permanently be stuck in broken session without user@.service.

src/login/logind-user.c
```diff
@@ -465,7 +465,8 @@ int user_start(User *u) {
         (void) user_update_slice(u);
 
         /* Start user@UID.service */
-        user_start_service(u);
+        if (want_user_instance)
```
poettering, line 468: this should not be a bool you pass, but a function user_needs_manager() or so which is called, which iterates through all sessions of the user, and returns true if there's one that isn't "background". It should also return true if lingering is enabled.

poettering (commented):
Hmpf. This is not quite right yet.

src/login/logind-session.c
msekletar: Hmm, it makes sense to keep ordering if you don't strictly require other service but do use it in case it is pulled-in by some else. I don't want background instances to interact with user@.service in any way so I don't see why ordering would be beneficial.

src/login/logind-user.c
msekletar: Hmm, interactive session will start user@.service...

poettering:
So, I am also pretty sure that we should leave the default for cron actually at starting user@.service. At this point it's far from clear to me that this is what people really want and would be less confusing to users. At this point, security agents and whatnot are all run from user@.service (and rightfully so I think), hence making this unavailable (as well as breaking XDG_RUNTIME_DIR and the like) sounds quite a lot to just like that willy-nilly. I mean, I value "correctness" and "uniformity" a lot more than "performance". And given the main (only?) reason for this PR is "performance" I think we should not change defaults here.

src/login/logind-session.c
poettering: well, you are *not* connecting these sessions from user@.service though. You are making it a race: maybe they will have user@.service, maybe they won't, it depends on whether the user is logged in otherwise. I think that is actually OK behaviour, I just want that to be systematic: i.e. if something pulls it in anyway, then wait for it. if nothing does, then don't.

src/login/logind.c
msekletar: Ahh, you are right. I forgot lingering means automatic startup of user manager as well...Cleaner solution would be that lingering as just that, i.e. user staying lingering after last session is gone and automatic startup of user manager is achieved via normal `systemctl enable user@$(id -u).service`. But I guess we can't really change that now.

msekletar: I think this will require a bit more work. I will file PR with revert of this commit to unbreak lingering.

poettering:
one more question about the usecase: if somebody wants user@.service to be off, and is in the backrgound, then why bother with logind at all? it appears then you don't want any of the concepts logind offers to you.

hence, maybe the right approach is to add a new env var pam_systemd.so honours calls XDG_SESSION_TRACK=no or so, which disables logind to bother at all, and shortcut itself.

I mean, with the scheme implemented here, things are a bit weird: you'll get half a hookup to logind, but only half, and things might work (because some other interactive session started user@.service for you) or might not (because nothing did). i wonder if the usecases that this is intended for wouldn't be better served with turning off logind for those services entirely?

Now, they can already do that, by editing the pam stacks and removing pam_systemd.so from the stacks. But of course, this doesn't work well for "superserver" services, like crond, where you have one daemon starting man PAM sessions with entirely different purposes and behaviours. Hence, allowing this to be configurable for each PAM session (which the XDG_SESSION_TRACK env var would allow) sounds like a better approach to me.

src/login/logind-user.c
poettering: user_start() shortcuts itself if already set up, afaics, so no, if a "backrgound" session comes first, and an "interactive" session is added later, then the user@.service instance wont be pulled in afaics.

msekletar:
Ok, I will file the revert of this one and go with new lightweight class.

msekletar:
IIRC we discussed this and I've mentioned we have background class that isn't really used for anything at this point except cron and that cron people actually hate the current behaviour where we start user@.service for cron sessions. I *think* we agreed to turn this around and not start user@.service for background class. The problem is that we rely on that (on starting user@.service for background) ourselves in implementation of --machine user@.host (that is why we now set PAM_TTY to systemd and make resulting session class=user without tty).

Anyway, we didn't have complete picture when we made a decision and even after learning about the problem with --machine I pushed it too far.

poettering:
Well, that might be, but who says I cannot contradict myself? ;-) From my current perspective I think the new class looks like a much better approach. (and XDG_SESSION_TRACK=no)

poettering:
> In fact given that classic cronjobs allow setting env vars, you might even be able to set the class from classic cronjob syntax, no? (this would require cron to pass the env vars into the pam session instead of setting it afterwards. if it doesn't already do that it really should).

I checked the sources of cronie. afaics it does *not* pass the job's env vars into PAM currently. But I think it should. might be worth filing an RFE to cronie about this.

poettering:
btw, to give a bit more background: i have though about some logind innner workings recently, and what i currently dislike is that the PAM session that user@.service requests is currently so weirdly handled by pam_systemd: it bypasses logind altogether, and then does weird, manual things to nonetheless get XDG_RUNTIME_DIR set properly. This special handling is done to avoid the cyclic dependency issue here, because that PAM session is different from all other PAM sessions in that it should not be delayed by user@.service being started up (because that would mean a deadlock: user@.service would only start once user@.service completed starting, ELOGICLOOP, anyone?). So this made me think how to handle this better: make user@.service a bit less special. And I came back to something close to the issue this PR tries to address: we should just let user@.service register a PAM session all the way to logind, but then specify an explicit class for it, maybe called "service-manager" or so. Which would be really similar to the "lightweight" class proposed above: but would differ in the After= handling. sessions of class "service-manager" after all *provide* the service manager that "lightweight" sessions should be able to consume if it exists, but not wait for.

Anyway, this is beyond the scope of this PR, I just wanted to give a bit of context here, why I am coming out of the woodwork with lots of new ideas... ;-)

msekletar:
From what I see in practice, people either want a full session, hence, logind object+session scope+user manager+device chown+XDG variables. Or they want all of that minus user-manager. In other words, the only complaint that I've heard so far was about resource consumption (CPU, memory and log bloat) that comes as a price for having user-manager even when users don't need it. By introducing this "leightweight" class we are addressing that concern.

Frankly, I don't really see  the need for introducing `XDG_SESSION_TRACK=`. What would be the actual use-case for it?

PR #23572, opened 2022-05-31
Remove message repeat

alshopov:
Original catalog file is in English
Remove the repeating German message

Signed-off-by: Alexander Shopov <ash@kambanaria.org>


keszybz:
Please move it to `catalog/systemd.de.catalog.in` instead.

yuwata (approved):

PR #23573, opened 2022-05-31
Update Bulgarian translation

alshopov:
po: catalog: Bring bg.po and catalog file to 100%

Signed-off-by: Alexander Shopov <ash@kambanaria.org>


keszybz:
I didn't check the translations, but looks reasonable in general.

alshopov:
> I didn't check the translations, but looks reasonable in general.

Just to clarify - I am longtime FLOSS translator and contributor of Bulgarian translation. I am also coordinator of GNOME translation team. Previous translation was also contributed by me: https://github.com/systemd/systemd/commit/c66c2c7c4f21acb7fe31f28d1eb8b9909df6d21e
This is a legitimate Bulgarian translation. Any error would be inadvertant and if pointed out - I will gladly fix it.

keszybz:
I didn't mean to imply that the patch was in doubt in anyway. We generally try to have at least one through review of all patches. For long-time and first-time any other contributor the policy is the same. I usually try to review translations for languages where I can, but here I only did a brief check, hence the comment.

alshopov:
No offence taken. I just wanted to confirm that my submission was in good faith.

PR #23574, opened 2022-05-31
Do not print logind wall message to local terminals

keszybz:


src/test/test-utmp.c
```diff
@@ -0,0 +1,61 @@
...
+
+TEST(dump_run_utmp) {
+        _unused_ _cleanup_(utxent_cleanup) bool utmpx = false;
+
+        utmpx = utxent_start();
+
+        for (struct utmpx *u; (u = getutxent()); ) {
+                char _type_buf[DECIMAL_STR_MAX(short)];
+                const char *type =
+                        u->ut_type == EMPTY         ? "EMPTY" :
```
medhefgo, line 28: switch?

medhefgo (approved):

src/login/logind-utmp.c
```diff
@@ -43,19 +43,17 @@ _const_ static usec_t when_wall(usec_t n, usec_t elapse) {
...
-                return true;
-
-        p = path_startswith(tty, "/dev/");
+        const char *p = path_startswith(tty, "/dev/");
         if (!p)
                 return true;
 
-        return !streq(p, m->scheduled_shutdown_tty);
+        /* Do not write to local pseudo-terminals */
+        if (startswith(tty, "pts/") && is_local)
```
yuwata: Hm? I cannot follow the code. Maybe `tty` -> `p`??

keszybz: Arrgh, indeed.

keszybz: So… after fixing this issue, messages are correctly suppressed, and this exposes a problem with KDE: as mentioned by @lnussel it registers a listener for wall messages, and that is now broken, `shutdown 5` does not show a message in the KDE. And looking at the utmp list, there's nothing in utmp that'd allow us to make a reasonable distinction between kde5 and xterm:
```
[fedora@rawhide-uefi systemd]$ build/test-utmp
/* test_dump_run_utmp */
     BOOT_TIME          0 line=~       id=~~   name=reboot   session=0 host=5.18.0-0.rc6.20220510git9be9ed2612b5aed.49.fc37.x86_64 addr=0.0.0.0
       RUN_LVL         51 line=~       id=~~   name=runlevel session=0 host=5.18.0-0.rc6.20220510git9be9ed2612b5aed.49.fc37.x86_64 addr=0.0.0.0
 LOGIN_PROCESS        606 line=tty1    id=tty1 name=LOGIN    session=606 host= addr=0.0.0.0
 LOGIN_PROCESS        609 line=ttyS0   id=tyS0 name=LOGIN    session=609 host= addr=0.0.0.0
  USER_PROCESS        641 line=pts/0   id=ts/0 name=fedora   session=0 host=192.168.122.1 addr=192.168.122.1
  USER_PROCESS        724 line=tty2    id=     name=fedora   session=0 host= addr=0.0.0.0
  USER_PROCESS        788 line=pts/1   id=ts/1 name=fedora   session=0 host=:0 addr=0.0.0.0
  USER_PROCESS       1024 line=pts/2   id=ts/2 name=fedora   session=0 host=:0 addr=0.0.0.0
[fedora@rawhide-uefi systemd]$ ps 788 1024
    PID TTY      STAT   TIME COMMAND
    788 ?        Ssl    0:01 /usr/bin/kded5
   1024 ?        S      0:00 /usr/bin/xterm
```

src/test/test-utmp.c
keszybz: So… switch cannot be used in an initalizer, so I'd trade one type of repetition for another
```diff
-                       u->ut_type == EMPTY         ? "EMPTY" :
+                       case EMPTY:
+                               type = "EMPTY";
+                               break;
```
Doesn't seem to be worth it.

mbiebl:
Maybe a much simpler solution is to simply drop the wall messages on suspend.
Afaics this is what most users are annoyed with.

keszybz:
Yep. I added @cgzones' patch on top now. The utmp code is just a mess, and even though the patch introduces heuristics that are not entirely clean, I don't think we can do anything better with the mess that utmp is.

mbiebl:
> Maybe a much simpler solution is to simply drop the wall messages on suspend. Afaics this is what most users are annoyed with.

I.e. #23538 without the `&& path_startswith(tty, "/dev/pts/"` bits.

src/login/logind-utmp.c
```diff
@@ -44,13 +44,23 @@ _const_ static usec_t when_wall(usec_t n, usec_t elapse) {
 
 bool logind_wall_tty_filter(const char *tty, bool is_local, void *userdata) {
         Manager *m = ASSERT_PTR(userdata);
+        assert(m->scheduled_shutdown_action);
```
yuwata, line 48: Please insert a blank line.

keszybz:
I dropped `path_startswith(tty, "/dev/pts/")` now. I think we don't need it in the end, see the comment in the patch.

src/login/logind-utmp.c
```diff
@@ -44,13 +44,22 @@ _const_ static usec_t when_wall(usec_t n, usec_t elapse) {
...
 
         const char *p = path_startswith(tty, "/dev/");
         if (!p)
                 return true;
 
-        /* Do not write to local pseudo-terminals */
-        if (startswith(p, "pts/") && is_local)
+        /* Do not information about events which do not destroy local sessions to local terminals. We can
```
yuwata: information -> inform or notify ??

src/login/logind-utmp.c
```diff
@@ -44,13 +44,22 @@ _const_ static usec_t when_wall(usec_t n, usec_t elapse) {
...
+         * assume that if the system enters sleep or hibernation, this will be visible in an obvious way for
+         * any local user. And once the systems exits sleep or hibernation, the notication would be just
```
yuwata: local user -> local users ??
exits sleep -> exits from sleep ??

src/login/logind-utmp.c
```diff
@@ -42,20 +42,27 @@ _const_ static usec_t when_wall(usec_t n, usec_t elapse) {
...
+        Manager *m = ASSERT_PTR(userdata);
+        assert(m->scheduled_shutdown_action);
 
-        p = path_startswith(tty, "/dev/");
+        const char *p = path_startswith(tty, "/dev/");
         if (!p)
                 return true;
 
-        return !streq(p, m->scheduled_shutdown_tty);
+        /* Do not information about events which do not destroy local sessions to local terminals. We can
```
mrc0mmand:
```suggestion
        /* Do not inform about events which do not destroy local sessions to local terminals. We can
```

or maybe a missing verb before `information`?

src/login/logind-utmp.c
keszybz: I think those are both fine. "Any" works with singular, "all" with plural. And "exits sleep" is also OK.

src/login/logind-utmp.c
keszybz: Updated.

src/login/logind-utmp.c
yuwata: OK. Thanks.

mbiebl:
> Yep. I added @cgzones' patch on top now. The utmp code is just a mess, and even though the patch introduces heuristics that are not entirely clean, I don't think we can do anything better with the mess that utmp is.

This still breaks the KDE/kded watch, right?

keszybz:
> This still breaks the KDE/kded watcher, right?

Not really. It shows messages about shutdown (or its cancellation). It won't show messages about suspend, but I think that's acceptable. It essentially restores behaviour from before cdf370626f08ed509a5dde9d5618eed29d625032 in this scenario.

src/login/logind-dbus.c
```diff
@@ -2357,10 +2355,12 @@ static int method_cancel_scheduled_shutdown(sd_bus_message *message, void *userd
                 }
 
                 username = uid_to_name(uid);
-                utmp_wall("The system shutdown has been cancelled",
+                utmp_wall("System shutdown has been cancelled",
```
yuwata: The line below needs to be updated:
https://github.com/systemd/systemd/blob/cd25e170014783503d5035ef4b067e7b00c01221/test/test-shutdown.py#L67

yuwata (changes requested):

src/login/logind-dbus.c
keszybz: I moved that chunk to the other PR.

src/login/logind-utmp.c
lnussel: Longer term for desktops it would probably make sense to come up with a dbus based notification mechanism the desktops can listen to.  There's already PrepareForShutdown/PrepareForSleep signals but those ones are only sent right before the action takes place.  Unlike the wall message that announce scheduled shutdowns too.

src/login/logind-utmp.c
```diff
@@ -45,12 +45,22 @@ _const_ static usec_t when_wall(usec_t n, usec_t elapse) {
...
+        /* Do not send information about events which do not destroy local sessions to local terminals. We
+         * can assume that if the system enters sleep or hibernation, this will be visible in an obvious way
+         * for any local user. And once the systems exits sleep or hibernation, the notication would be just
+         * noise, in particular for auto-suspend. */
+        if (is_local &&
+            IN_SET(m->scheduled_shutdown_action->handle,
+                   HANDLE_SUSPEND,
+                   HANDLE_HIBERNATE,
+                   HANDLE_HYBRID_SLEEP,
+                   HANDLE_SUSPEND_THEN_HIBERNATE))
```
lnussel, line 63: another way to express this set would be to put a boolean into `handle_action_data_table`

yuwata (approved):

keszybz:
focal-amd64 fails in TEST-* (4 different tests). I see it's failing in other places too.

poettering:
Hmm, I think a better approach might be to just ignore utmp altogether, and instead go by our own session list we maintain anyway in memory in logind. If we do that the problem just goes away, as terminal multiplexers like graphical sessions will be just *one* session that way, and not *many* as in utmp.

I figure it would be wise anyway to forego utmp as *source* of truth wherever we can, and instead use better information we have anyway, and consider utmp just a *display* of our own concept of state. Already because we allow systemd to be built without utmp support, and utmp is kinda a legacyish concept.

keszybz:
I considered that, but it doesn't seem to work at all. See the discussion about kde5 above: it is not registered as a logind session, nor should it. And there are many other examples where existing expectations would be broken. Xterm has an explicit switch to enable/disable utmp registration, so if we start ignoring that, we'll break those setups.

Overall, I think we could build a better system that uses dbus signals that anybody can subscribe to, but it'd serve as something to add in parallel. We'd need to keep the utmp-based system anyway.

lnussel:
exactly. I assume the kde hack exists exactly because there is just no other channel for the admin or system to notify/warn all users of an event. If systemd/logind starts to offer a better mechanism, the desktops would likely start using it. Could probably even replace `/usr/bin/wall` with one that uses logind instead of utmp.

keszybz:
So… maybe instead of trying to invent a new mechanism, we could just reuse the journal for this?
We currently send a line to ttys, but not the journal. If we start doing that with an appropriate message-id, people could easily listen for that.

PR #23576, opened 2022-05-31
network: support erspan version 0 and 2

### LINKED ISSUE #23570: networkd: Allow to set version for erspan tunnel
There is no possibility to the the desired erspan version (IFLA_GRE_ERSPAN_VER) when creating an erspan tunnel, right?
I would assume something like ERSPANVersion in the [Tunnel] section for netdevs.

yuwata:
Closes #23570.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -699,8 +699,10 @@ class Utilities():
             check_output(*args, env=env)
         except subprocess.CalledProcessError:
             for link in links_with_operstate:
-                output = check_output(*networkctl_cmd, '-n', '0', 'status', link.split(':')[0], env=env)
-                print(output)
+                name = link.split(':')[0]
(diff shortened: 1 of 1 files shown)


keszybz:
LGTM.

I think we could try to clean up the error messages a bit: I think it'd be easier both for us, when writing the code, and for the user when getting error messages, if we just used the original option name instead of trying to come up with a descriptive replacement. I.e. say "Invalid ERSPANHardwareID=", not "Invalid erspan hardware identifier". But it's a big job, because I think we should do try to do this comprehensively, i.e. change whole swaths of such strings. If we did this in a smart way, we could also reduce the static data size quite a bit. E.g. if we used `"Invalid %s=%s: %m", lvalue, rvalue`, we'd have a single string reused quite a bit. It wouldn't make sense to use the same format everywhere, but maybe we could reuse some formats at least a bit.

Issue #19208, opened 2021-04-05
RFE: cryptsetup: support automatic LUKS header FIDO2 param discovery even if multiple FIDO2 tokens are enrolled


daenney:
**systemd version the issue has been seen with**
 > v248

**Used distribution**
 > Arch Linux

**Linux kernel version used** (`uname -a`)
 > `Linux unicorn 5.11.11-arch1-1 #1 SMP PREEMPT Tue, 30 Mar 2021 14:10:17 +0000 x86_64 GNU/Linux`

**CPU architecture issue was seen on**
 > amd64

**Expected behaviour you didn't see**
 > I'm able to unlock my root volume with one of my fido2 keys

**Unexpected behaviour you saw**
 > I was unable to unlock, with either `fido2-device=auto` or `fido2-device=/dev/hidraw1` passed as `rd.luks.options=`

**Steps to reproduce the problem**
 > Enrolled 2 keys using `systemd-cryptenroll --fido2-device=auto /dev/nvme0n1p4` each time (with only one key plugged in each time)

**Additional program output to the terminal or log subsystem illustrating the issue**
When booting with `fido2-device=auto`:

```
Apr 05 01:35:12 archlinux kernel: Command line: rd.luks.options=fido2-device=auto rd.luks.uuid=c399df9f-f0db-4b5d-8743-27ebfe445161 rd.luks.name=c399df9f-f0db-4b5d-8743-27ebfe445161=unicorn root=/dev/mapper/unicorn debug
...
Apr 05 01:35:12 archlinux systemd-cryptsetup[209]: Multiple FIDO2 tokens enrolled, cannot automatically determine token.
Apr 05 01:35:12 archlinux systemd-cryptsetup[209]: Automatic FIDO2 metadata discovery was not possible because missing or not unique, falling back to traditional unlocking.
```

When booting with `fido2-device=/dev/hidraw1`
```
Apr 05 02:40:07 archlinux kernel: Command line: rd.luks.options=fido2-device=/dev/hidraw1 rd.luks.uuid=c399df9f-f0db-4b5d-8743-27ebfe445161 rd.luks.name=c399df9f-f0db-4b5d-8743-27ebfe445161=unicorn root=/dev/mapper/unicorn debug
...
Apr 05 02:40:07 archlinux systemd-cryptsetup[203]: Multiple FIDO2 tokens enrolled, cannot automatically determine token.
[...]

PR #23577, opened 2022-05-31
libfido2-util: Do pre-flight to determine whether the credential is in the token

MkfsSion:
* Avoid mutiple touches if the keyslot is not the first one

Possibly fix: #19208

Only device without builtin UV is tested since I don't own a builtin UV device (like Yubikey BIO series).
Test is highly welcomed especially for builtin UV device user.

### DIFF
--- a/src/shared/libfido2-util.c
+++ b/src/shared/libfido2-util.c
@@ -194,6 +194,162 @@ static int verify_features(
         return 0;
 }
 
+static int fido2_assert_set_basic_properties(
+                fido_assert_t *a,
+                const char *rp_id,
+                const void *cid,
+                size_t cid_size) {
+        int r;
+
+        assert(a);
+        assert(rp_id);
+        assert(cid);
+        assert(cid_size > 0);
+
+        r = sym_fido_assert_set_rp(a, rp_id);
+        if (r != FIDO_OK)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                       "Failed to set FIDO2 assertion ID: %s", sym_fido_strerr(r));
+
+        r = sym_fido_assert_set_clientdata_hash(a, (const unsigned char[32]) {}, 32);
+        if (r != FIDO_OK)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                       "Failed to set FIDO2 assertion client data hash: %s", sym_fido_strerr(r));
+
+        r = sym_fido_assert_allow_cred(a, cid, cid_size);
+        if (r != FIDO_OK)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                       "Failed to add FIDO2 assertion credential ID: %s", sym_fido_strerr(r));
+
+        return 0;
+}
+
+static int fido2_common_assert_error_handle(int r) {
+
+        switch (r) {
+
+        case FIDO_OK:
+                break;
+        case FIDO_ERR_NO_CREDENTIALS:
+                return log_error_errno(SYNTHETIC_ERRNO(EBADSLT),
+                                       "Wrong security token; needed credentials not present on token.");
+        case FIDO_ERR_PIN_REQUIRED:
+                return log_error_errno(SYNTHETIC_ERRNO(ENOANO),
+                                       "Security token requires PIN.");
+        case FIDO_ERR_PIN_AUTH_BLOCKED:
+                return log_error_errno(SYNTHETIC_ERRNO(EOWNERDEAD),
+                                       "PIN of security token is blocked, please remove/reinsert token.");
+#ifdef FIDO_ERR_UV_BLOCKED
+        case FIDO_ERR_UV_BLOCKED:
+                return log_error_errno(SYNTHETIC_ERRNO(EOWNERDEAD),
+                                       "Verification of security token is blocked, please remove/reinsert token.");
+#endif
+        case FIDO_ERR_PIN_INVALID:
+                return log_error_errno(SYNTHETIC_ERRNO(ENOLCK),
+                                       "PIN of security token incorrect.");
+        case FIDO_ERR_UP_REQUIRED:
+                return log_error_errno(SYNTHETIC_ERRNO(EMEDIUMTYPE),
+                                       "User presence required.");
+        case FIDO_ERR_ACTION_TIMEOUT:
+                return log_error_errno(SYNTHETIC_ERRNO(ENOSTR),
+                                       "Token action timeout. (User didn't interact with token quickly enough.)");
+        default:
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                        "Failed to ask token for assertion: %s", sym_fido_strerr(r));
+        }
+
+        return 0;
+}
+
+static int fido2_is_cred_in_specific_token(
+                const char *path,
+                const char *rp_id,
+                const void *cid,
+                size_t cid_size,
+                char **pins,
+                Fido2EnrollFlags flags) {
+
+        assert(path);
+        assert(rp_id);
+        assert(cid);
+        assert(cid_size);
+
+        _cleanup_(fido_dev_free_wrapper) fido_dev_t *d = NULL;
+        _cleanup_(fido_assert_free_wrapper) fido_assert_t *a = NULL;
+        bool has_up = false, has_uv = false;
+        int r;
+
+        d = sym_fido_dev_new();
+        if (!d)
+                return log_oom();
+
+        r = sym_fido_dev_open(d, path);
+        if (r != FIDO_OK)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                       "Failed to open FIDO2 device %s: %s", path, sym_fido_strerr(r));
+
+        r = verify_features(d, path, LOG_ERR, NULL, NULL, &has_up, &has_uv);
+        if (r < 0)
+                return r;
+
+        a = sym_fido_assert_new();
+        if (!a)
+                return log_oom();
+
+        r = fido2_assert_set_basic_properties(a, rp_id, cid, cid_size);
+        if (r < 0)
+                return log_error_errno(r,
+                                       "Failed to set FIDO2 assertion basic properties: %m");
+
+        /* According to CTAP 2.1 specification, to do pre-flight we need to set up option to false
+         * with optionally pinUvAuthParam in assertion[1]. But for authenticator that doesn't support
+         * user presence, once up option is present, the authenticator may return CTAP2_ERR_UNSUPPORTED_OPTION[2].
+         * So we simplely omit the option in that case.
+         * Reference:
+         * 1: https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#pre-flight
+         * 2: https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#authenticatorGetAssertion (in step 5)
+         */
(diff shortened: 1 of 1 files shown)


mrc0mmand:
> @mrc0mmand Seems CI failure doesn't relate to the PR, could you please re-trigger the CI?

I re-triggered Semaphore CI, but Ubuntu CIs are out of my reach - @bluca can you help with that?

src/shared/libfido2-util.c
```diff
@@ -195,6 +195,122 @@ static int verify_features(
...
+
+        r = sym_fido_assert_set_clientdata_hash(a, (const unsigned char[32]) {}, 32);
+        if (r != FIDO_OK)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                       "Failed to set FIDO2 assertion client data hash: %s", sym_fido_strerr(r));
+
+        r = sym_fido_assert_allow_cred(a, cid, cid_size);
+        if (r != FIDO_OK)
+                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                       "Failed to add FIDO2 assertion credential ID: %s", sym_fido_strerr(r));
```
poettering: these same three calls are also done by fido2_use_hmac_hash_specific_token() and fido2_generate_hmac_hash(), now you add a third. Can you split this out to a new helper, that is calle dby all three?

```c
static fido2_assert_set_basic_properties(
                fido_assert *a,
                const char *rp_id,
                const void *cid,
                size_t cid_size) {
        int r;

        assert(a);
        assert(cid);
        assert(cid_size > 0);

        r = sym_fido_assert_set_rp(a, rp_id);
        if (r != FIDO_OK)
                return log_error_errno(SYNTHETIC_ERRNO(EIO),
                                       "Failed to set FIDO2 assertion ID: %s", sym_fido_strerr(r));

        r = sym_fido_assert_set_clientdata_hash(a, (const unsigned char[32]) {}, 32);
        if (r != FIDO_OK)
                return log_error_errno(SYNTHETIC_ERRNO(EIO),
                                       "Failed to set FIDO2 assertion client data hash: %s", sym_fido_strerr(r));

        r = sym_fido_assert_allow_cred(a, cid, cid_size);
        if (r != FIDO_OK)
                return log_error_errno(SYNTHETIC_ERRNO(EIO),
                                       "Failed to add FIDO2 assertion credential ID: %s", sym_fido_strerr(r));

        return 0;
}
```

or something like that. You get the idea.

mrc0mmand:
> > @mrc0mmand Seems CI failure doesn't relate to the PR, could you please re-trigger the CI?
> 
> I re-triggered Semaphore CI, but Ubuntu CIs are out of my reach - @bluca can you help with that?

Also, it looks like the PR needs to be rebased, as Semaphore CI complains about issue already fixed by b5472417284.

src/shared/libfido2-util.c
```diff
@@ -195,6 +195,122 @@ static int verify_features(
...
+                        case FIDO_ERR_PIN_INVALID:
+                                return log_error_errno(SYNTHETIC_ERRNO(ENOLCK),
+                                                       "PIN of security token incorrect.");
+                        case FIDO_ERR_PIN_BLOCKED:
+                                return log_error_errno(SYNTHETIC_ERRNO(EOWNERDEAD),
+                                                       "PIN of security token is blocked, please remove/reinsert token.");
+                        default:
+                                return log_error_errno(SYNTHETIC_ERRNO(EIO),
+                                                       "Failed to get assertion result: %s", sym_fido_strerr(r));
+                }
```
poettering: similar, this switch statement is a subset of the one in fido2_use_hmac_hash_specific_token(). Can we share that in a helper?

i.e.

```c
static int fido2_common_assert_error_handle(int r) {
         …
}
```

You get the idea.

poettering: And then call it here:

```c
switch (r) {
        case FIDO_OK:
                return true;
        case FIDO_ERR_NO_CREDENTIALS;
                return false;
        default:
                return fido2_common_assert_error_handle(r);
}
```

src/shared/libfido2-util.c
```diff
@@ -195,6 +195,122 @@ static int verify_features(
...
+        }
+
+        if (r == FIDO_ERR_NO_CREDENTIALS)
+                *ret_in_token = false;
+        else
+                *ret_in_token = true;
+
+        return 0;
```
poettering: commonly across our codebase we have functions that return "extended booleans", i.e. they return == 0 as "false", > 0 as "true" and < 0 as "failure" (with the precise code indicating an error code.

please follow this here too, i.e. fold the `ret_in_token` return parameter into the actual return value, by return it as 0 (if token doesn't have the creds) or 1 (if it has) or -EIO (or some other errno) if we couldn#t determine that.

src/shared/libfido2-util.c
```diff
@@ -520,6 +636,32 @@ int fido2_use_hmac_hash(
                         goto finish;
                 }
 
+                bool in_token = false;
+
+                r = fido2_is_cred_in_specific_token(path, rp_id, cid, cid_size, pins, required, &in_token);
+                /* We handle PIN related errors here since we have used PIN in the check.
+                 * If PIN error happens, we can avoid pin retries decreased the second time. */
+                if (r < 0) {
```
poettering: we like shallow indentation, hence just drop this `if (r < 0)` early check here, just check against the various values in the if checks below.

src/shared/libfido2-util.c
```diff
@@ -520,6 +636,32 @@ int fido2_use_hmac_hash(
...
+                        if (IN_SET(r, -ENOANO, -ENOLCK, -EOWNERDEAD)) {
+                                /* If it's not the last device, try next one */
+                                if (i < found - 1)
+                                        continue;
+                                else {
```
poettering: you can drop this `else {` without ill effects.

src/shared/libfido2-util.c
```diff
@@ -520,6 +636,32 @@ int fido2_use_hmac_hash(
...
+                                        if (r == -EOWNERDEAD)
+                                                r = -EAGAIN;
```
poettering: what's the rationale for this?

src/shared/libfido2-util.c
```diff
@@ -520,6 +636,32 @@ int fido2_use_hmac_hash(
...
+                                        goto finish;
+                                }
+                        } else if (r == -NODEV)
+                                continue;
+                        else
+                                r = log_error_errno(r, "Failed to determine whether the credential is in the token, try anyway: %m");
```
poettering: you can drop the `r = ` prefix here without any change in behaviour

poettering:
looks conceptually ok.

src/shared/libfido2-util.c
MkfsSion: EOWNERDEAD is returned when PIN is blocked, map it to the EAGAIN so users can reinsert the token and try again.

MkfsSion: This makes it behave same with https://github.com/systemd/systemd/blob/main/src/shared/libfido2-util.c#L529

poettering: there should be a comment

src/shared/libfido2-util.c
```diff
@@ -403,36 +549,9 @@ static int fido2_use_hmac_hash_specific_token(
...
-                return log_error_errno(SYNTHETIC_ERRNO(EMEDIUMTYPE),
-                                       "User presence required.");
-        case FIDO_ERR_ACTION_TIMEOUT:
-                return log_error_errno(SYNTHETIC_ERRNO(ENOSTR),
-                                       "Token action timeout. (User didn't interact with token quickly enough.)");
-        default:
-                return log_error_errno(SYNTHETIC_ERRNO(EIO),
-                                       "Failed to ask token for assertion: %s", sym_fido_strerr(r));
-        }
+        int err = fido2_common_assert_error_handle(r);
```
poettering, line 552: why introduce `int err` here? why not just use `r = fido2_common_assert_error_handle(r)` here? you don't need libfido's error code anymore here, do you?

src/shared/libfido2-util.c
```diff
@@ -194,6 +194,162 @@ static int verify_features(
...
+                return r;
+
+        a = sym_fido_assert_new();
+        if (!a)
+                return log_oom();
+
+        r = fido2_assert_set_basic_properties(a, rp_id, cid, cid_size);
+        if (r < 0)
+                return log_error_errno(r,
+                                       "Failed to set FIDO2 assertion basic properties: %m");
```
poettering, line 302: fido2_assert_set_basic_properties() already logs about all errors, hence you shouldn't log about the errors here again.

in systemd, we have this rule that all functions should either log about all errors, or about none (in which case the error should be propagated, so that the caller can log). That way, we ensure that each error is only logged about once, and exactly once, i.e. not more than once or even not at all. (LOG_DEBUG level logging is always OK though, it's not considered "logging" in this context). Hence for each function figure out what it should be. Typically, if it's more "library-like" it should not log at all, and if it's more "main-program"-like, it should log about all errors. Now, fido2_assert_set_basic_properties() already is a logging function, hence remove the extra logging here, and just do `return r` here.

src/shared/libfido2-util.c
```diff
@@ -519,6 +638,30 @@ int fido2_use_hmac_hash(
...
                 }
 
+                r = fido2_is_cred_in_specific_token(path, rp_id, cid, cid_size, pins, required);
+                /* We handle PIN related errors here since we have used PIN in the check.
+                 * If PIN error happens, we can avoid pin retries decreased the second time. */
+                if (IN_SET(r, -ENOANO, -ENOLCK, -EOWNERDEAD)) {
+                        /* If it's not the last device, try next one */
+                        if (i < found - 1)
+                                continue;
+                        else {
```
poettering, line 648: you can remove this else indentation, after all the if branch already die `continue`. We are big fans of shallow indentation.

src/shared/libfido2-util.c
```diff
@@ -519,6 +638,30 @@ int fido2_use_hmac_hash(
                         goto finish;
                 }
 
+                r = fido2_is_cred_in_specific_token(path, rp_id, cid, cid_size, pins, required);
+                /* We handle PIN related errors here since we have used PIN in the check.
+                 * If PIN error happens, we can avoid pin retries decreased the second time. */
+                if (IN_SET(r, -ENOANO, -ENOLCK, -EOWNERDEAD)) {
```
poettering, line 644: maybe write this as:

```c
if (IN_SET(r, -ENOANO,        /* → pin required */
              -ENOLCK,        /* → pin incorrect */
              -EOWNERDEAD)) { /* → verification (uv) blocked */
```

or so. i.e. we use unix error codes here, and they are really cryptic, hence it's always good to explain what an error we handle specifically means where we handle it.

src/shared/libfido2-util.c
```diff
@@ -519,6 +638,30 @@ int fido2_use_hmac_hash(
...
+                        /* If it's not the last device, try next one */
+                        if (i < found - 1)
+                                continue;
+                        else {
+                                /* -EOWNERDEAD is returned when PIN is blocked. Map it to EAGAIN so users can reinsert
```
poettering, line 649: PIN is block? uv is blocked! no?

poettering:
it would be excellent, if the commit could be split out, e.g. fido2_assert_set_basic_properties() done in a first commit, and so on.

PeterCxy:
Here is an updated version of what is in this PR with the latest reviews addressed: https://github.com/PeterCxy/systemd/commits/fido2-preflight

@MkfsSion Do you still intend to keep this PR going? If not I can take over and re-open a new PR from my branch with all the new changes.

bluca:
close for https://github.com/systemd/systemd/pull/25268

PR #23579, opened 2022-05-31
sha256: use memcpy() to accept unaligned result buffer

### LINKED ISSUE #23578: systemd-cryptenroll fails under UBSan
 > a52765a5504108

```sh
env PASSWORD=passphrase NEWPIN=123456 systemd-cryptenroll --tpm2-device=auto --tpm2-with-pin=true $img
...
../src/fundamental/sha256.c:125:42: runtime error: store to misaligned address 0x7f466b5e5132 for type 'uint32_t', which requires 4 byte alignment
0x7f466b5e5132: note: pointer points here
 00 00  20 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00
              ^
    #0 0x7f4671851735 in sha256_finish_ctx ../src/fundamental/sha256.c:125
    #1 0x7f467161fdb0 in hash_pin ../src/shared/tpm2-util.c:816
    #2 0x7f4671623da7 in tpm2_unseal ../src/shared/tpm2-util.c:1147
    #3 0x4170e6 in enroll_tpm2 ../src/cryptenroll/cryptenroll-tpm2.c:177
    #4 0x411e4d in run ../src/cryptenroll/cryptenroll.c:588
    #5 0x4121c1 in main ../src/cryptenroll/cryptenroll.c:613
    #6 0x7f466fab143f in __libc_start_call_main (/lib64/libc.so.6+0x4043f)
    #7 0x7f466fab14ef in __libc_start_main_alias_2 (/lib64/libc.so.6+0x404ef)
    #8 0x403964 in _start (/usr/bin/systemd-cryptenroll+0x403964)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/fundamental/sha256.c:125:42 in
```
Looking at https://github.com/systemd/systemd/blob/3e597d0d3547f9f140fb01b8755698a91b984127/src/fundamental/sha256.c#L96-L100
it seems `sha256_finish_ctx` expects `RESBUF` to be aligned properly and `tpm2_unseal` passes a member of a TPM2B_AUTH structure. As far as I can see it looks like
```
{
    UINT16 size;
    BYTE buffer[sizeof(TPMU_HA)];
};
```

yuwata:
Fixes #23578.

### DIFF
--- a/src/test/test-hmac.c
+++ b/src/test/test-hmac.c
@@ -10,59 +10,59 @@ static void hmac_sha256_by_string(const char *key, const char *value, uint8_t re
 }
 
 TEST(hmac) {
-    uint8_t result[SHA256_DIGEST_SIZE];
-    char *hex_result = NULL;
+        uint8_t result[SHA256_DIGEST_SIZE];
+        char *hex_result = NULL;
 
-    /* Results compared with output of 'echo -n "<input>" | openssl dgst -sha256 -hmac "<key>"' */
+        /* Results compared with output of 'echo -n "<input>" | openssl dgst -sha256 -hmac "<key>"' */
 
-    hmac_sha256_by_string("waldo",
-                          "",
-                          result);
-    hex_result = hexmem(result, sizeof(result));
-    assert_se(streq_ptr(hex_result, "cadd5e42114351181f3abff477641d88efb57d2b5641a1e5c6d623363a6d3bad"));
-    hex_result = mfree(hex_result);
+        hmac_sha256_by_string("waldo",
+                              "",
+                              result);
+        hex_result = hexmem(result, sizeof(result));
+        assert_se(streq_ptr(hex_result, "cadd5e42114351181f3abff477641d88efb57d2b5641a1e5c6d623363a6d3bad"));
+        hex_result = mfree(hex_result);
 
-    hmac_sha256_by_string("waldo",
-                          "baldohaldo",
-                          result);
-    hex_result = hexmem(result, sizeof(result));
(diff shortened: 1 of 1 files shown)


src/fundamental/sha256.c
```diff
@@ -122,7 +119,7 @@ void *sha256_finish_ctx(struct sha256_ctx *ctx, void *resbuf) {
 
         /* Put result from CTX in first 32 bytes following RESBUF.  */
         for (size_t i = 0; i < 8; ++i)
-                ((uint32_t *) resbuf)[i] = SWAP(ctx->H[i]);
+                memcpy((uint8_t*) resbuf + i * sizeof(uint32_t), (uint32_t[]) { SWAP(ctx->H[i]) }, sizeof(uint32_t));
```
keszybz: We have UNALIGNED_P() defined below. Shouldn't this macro be also used here to avoid unnecessary copying?

I think we should (not necessarily in this PR) have a test for this hashing code. In particular, with unaligned buffers.

yuwata:
Updated. Now `UNALIGNED_P` macro is reused, and test for sha256 is added. PTAL.

src/test/test-sha256.c
```diff
@@ -0,0 +1,50 @@
...
+                        _cleanup_free_ char *hex_result = NULL;
+
+                        sha256_init_ctx(&ctx);
+                        sha256_process_string(str + i, &ctx);
+                        sha256_finish_ctx(&ctx, result + j);
+
+                        hex_result = hexmem(result + j, SHA256_DIGEST_SIZE);
+                        assert_se(streq_ptr(hex_result, expect));
+                }
+        }
```
keszybz, line 36: Nice, thanks for adding the test.

PR #23582, opened 2022-05-31
resolved: define source address for proxy-only stub replies

### LINKED ISSUE #23495: Stub on 127.0.0.54 responds from wrong address
 > systemd-251-1.fc37.x86_64

**Used distribution**
 > Fedora Rawhide (37)

 > `Linux *.redhat.com 5.18.0-0.rc7.54.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Mon May 16 14:40:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

 > x86_64

 > Just working ``dig @127.0.0.54``

 > 
```
# dig @127.0.0.54
;; reply from unexpected source: 127.0.0.53#53, expected 127.0.0.54#53

;; reply from unexpected source: 127.0.0.53#53, expected 127.0.0.54#53

;; reply from unexpected source: 127.0.0.53#53, expected 127.0.0.54#53


; <<>> DiG 9.16.28-RH <<>> @127.0.0.54
; (1 server found)
;; global options: +cmd
;; connection timed out; no servers could be reached
```

 > ``dig @127.0.0.54``

(nothing more required)

This is a misfeature again. LLMNR behaviour has to be fixed different way, so it does work by default both on localhost and from dependent containers or VMs as well. It should accept on the normal socket offered by default DNS queries. Another socket for special cases is wrong workaround. Please fix the primary socket instead.

bnf:
`DnsPacket.ifindex=1` (loopback) is normalized to `0` whenever a message is
received on the loopback iface, so for both listeners, `127.0.0.53` and
`127.0.0.54`, the `ifindex` will be set to `0` by `manager_recv()` for queries
that have a local origin.

Replies to such local messages need to set a proper `ifindex` in any
case, as the supplied source-address would otherwise be ignored in
`manager_ipv4_send()` (`CMSG` generation is skipped due to `ifindex > 0` check).

Note that this change only forces `ifindex` to loopback if it was actually
normalized to `0` before (due to a loopback detection) in order to keep the
nat-to-127.0.0.54-from-another-interface usecase that was described in
a8d09063447568d87288a8e868fe386c1da7ce09 intact.
Also note that nat is not supported for the main stub 127.0.0.53 which is
why forcing LOOPBACK_IFINDEX was/is fine for that case.

Fixes #23495

Can be tested locally via `dig +notcp @127.0.0.54` which then no longer complaints about `;; reply from unexpected source: 127.0.0.53#53, expected 127.0.0.54#53`

Also nat'ing 127.0.0.54 from a container still works on uses the correct source interface for replies:

```
echo 1 > /proc/sys/net/ipv4/conf/ve-fedora/route_localnet 
iptables -t nat -I PREROUTING -p udp -i ve-fedora --dport 53 -j DNAT --to 127.0.0.54:53
machinectl shell fedora /usr/bin/dig +notcp +short @_gateway
```

### DIFF
--- a/src/resolve/resolved-dns-stub.c
+++ b/src/resolve/resolved-dns-stub.c
@@ -530,7 +530,7 @@ static int dns_stub_send(
         else {
                 int fd, ifindex;
 
-                fd = find_socket_fd(m, l, p->family, &p->sender, SOCK_DGRAM);
+                fd = find_socket_fd(m, l, p->family, &p->destination, SOCK_DGRAM);
                 if (fd < 0)
                         return fd;
 


yuwata (approved):
LGTM.

src/resolve/resolved-dns-stub.c
```diff
@@ -528,18 +528,26 @@ static int dns_stub_send(
...
+                         * and ifindex was normalized to 0 by manager_recv(). */
+                        ifindex = p->ifindex ?: LOOPBACK_IFINDEX;
+                else
+                        /* Force loopback iface if this is the main listener stub. */
+                        ifindex = l ? p->ifindex : LOOPBACK_IFINDEX;
+
                 /* Note that it is essential here that we explicitly choose the source IP address for this
                  * packet. This is because otherwise the kernel will choose it automatically based on the
-                 * routing table and will thus pick 127.0.0.1 rather than 127.0.0.53. */
+                 * routing table and will thus pick 127.0.0.1 rather than 127.0.0.53/54. */
```
dnicolodi, line 547: Maybe I am missing something, but isn't 127.0.0.54 just one possible example out of 16777216 possible addresses that can be assigned to the loopback interface and to which resolved can be listening to?

bnf: You mean via something like `DNSStubListenerExtra=127.0.0.123`?
Yes, true, those cases would apply as well, but as the default stub and proxy stub addresses aren't configurable – and as `DNSStubListenerExtra` is intended for actually configured addresses – I think it is fine to only mention these two "static" loopback addresses that would really be affected here.

Anyway, for sure, the same applies to loopback addresses defined via `DNSStubListenerExtra` as well, and I wanted verify that and now actually noticed one additional bug thanks to your comment:
While the wrong ifindex(=0) is also used for loopback `DNSStubListenerExtra` replies (fixable only by dropping the ifindex normalization in `manager_recv()`), the kernel still choose the correct sender address because the correct file descriptor was (re)used. So only proxy stub replies where wrong because they used a wrong `fd` (the one from the default stub) *and* wrong ifindex.
I'll push a separate commit to fix that.

PR #23583, opened 2022-05-31
boot: follow-ups for efi-string functions

yuwata:
Several points that I overlooked in the previous review.

### DIFF
--- a/src/boot/efi/efi-string.c
+++ b/src/boot/efi/efi-string.c
@@ -60,14 +60,13 @@ DEFINE_STRTOLOWER(char16_t, strtolower16);
                         return CMP(s1, s2);                  \
                                                              \
                 while (n > 0) {                              \
-                        int c1 = *s1;                        \
-                        int c2 = *s2;                        \
+                        type c1 = *s1, c2 = *s2;             \
                         if (tolower) {                       \
                                 c1 = TOLOWER(c1);            \
                                 c2 = TOLOWER(c2);            \
                         }                                    \
                         if (!c1 || c1 != c2)                 \
-                                return c1 - c2;              \
+                                return CMP(c1, c2);          \
                                                              \
                         s1++;                                \
                         s2++;                                \
@@ -141,13 +140,16 @@ DEFINE_STRNDUP(char, xstrndup8, strnlen8);
 DEFINE_STRNDUP(char16_t, xstrndup16, strnlen16);
 
 int efi_memcmp(const void *p1, const void *p2, size_t n) {
+        const uint8_t *up1 = p1, *up2 = p2;
+        int r;
+
         if (!p1 || !p2)
                 return CMP(p1, p2);
 
-        const uint8_t *up1 = p1, *up2 = p2;
         while (n > 0) {
-                if (*up1 != *up2)
-                        return *up1 - *up2;
+                r = CMP(*up1, *up2);
+                if (r != 0)
+                        return r;
 
                 up1++;
                 up2++;
--- a/src/boot/efi/test-efi-string.c
+++ b/src/boot/efi/test-efi-string.c
@@ -335,6 +335,7 @@ TEST(efi_memcmp) {
         assert_se(efi_memcmp("a", "A", 1) > 0);
         assert_se(efi_memcmp("abc", "ab", 2) == 0);
         assert_se(efi_memcmp("ab", "abc", 3) < 0);
+        assert_se(efi_memcmp("abc", "ab", 3) > 0);
         assert_se(efi_memcmp("ab\000bd", "ab\000bd", 6) == 0);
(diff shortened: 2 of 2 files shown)


src/boot/efi/efi-string.c
```diff
@@ -60,14 +60,13 @@ DEFINE_STRTOLOWER(char16_t, strtolower16);
...
-                        int c1 = *s1;                        \
-                        int c2 = *s2;                        \
+                        type c1 = *s1, c2 = *s2;             \
                         if (tolower) {                       \
                                 c1 = TOLOWER(c1);            \
                                 c2 = TOLOWER(c2);            \
                         }                                    \
                         if (!c1 || c1 != c2)                 \
-                                return c1 - c2;              \
+                                return CMP(c1, c2);          \
```
medhefgo, line 69: `c1 - c2` is faster, no?

src/boot/efi/efi-string.c
```diff
@@ -160,9 +160,12 @@ int efi_memcmp(const void *p1, const void *p2, size_t n) {
 }
 
 void *efi_memcpy(void * restrict dest, const void * restrict src, size_t n) {
-        if (!dest || !src || n == 0)
+        if (n == 0)
```
medhefgo: I dislike this. The whole point for the strings library is to also be NULL pointer safe and doing something sensible rather than getting our asses kicked at runtime in obscure corner cases. Assert would certainly work, but not with `NDEBUG` builds.

src/boot/efi/efi-string.c
```diff
@@ -141,13 +140,16 @@ DEFINE_STRNDUP(char, xstrndup8, strnlen8);
...
+        int r;
+
         if (!p1 || !p2)
                 return CMP(p1, p2);
 
-        const uint8_t *up1 = p1, *up2 = p2;
         while (n > 0) {
-                if (*up1 != *up2)
-                        return *up1 - *up2;
+                r = CMP(*up1, *up2);
```
medhefgo, line 150: Same here? Also, I'd prefer to declare the `int r` here at least.

yuwata: Here, `*up1` and `*up2` are unsigned. So, I'd like to use `CMP()` for safety.
Also, we usually declare `r` at the beginning, as it is a kind of special variable.

src/boot/efi/efi-string.c
yuwata: If we use `int` for `c1` and `c2`, then `c1 - c2` is ok, at least, not bad. But I still prefer `CMP()` as this is a compare function. In many compare functions we declare, `CMP()` is heavily used and we do not use subtraction.

src/boot/efi/efi-string.c
yuwata: Hmm... Then, I think we should slightly rename it to emphasize that this accepts NULL even if `n != 0`. At least, a brief comment should be added.

medhefgo: Well, the comment at the top of the file still applies ;D

If you really want you could rename to `efi_memcpy_safe` etc.

src/boot/efi/efi-string.c
```diff
@@ -37,14 +37,6 @@
 DEFINE_STRNLEN(char, strnlen8);
 DEFINE_STRNLEN(char16_t, strnlen16);
 
-size_t strlen8(const char *s) {
```
medhefgo, line 40: I just realized I can provide inline forms for these too (similar for strcmp) making this commit unnecessary…:
```c
#define DEFINE_STRNLEN(type, suffix)                                                     \
        static inline size_t _strnlen##suffix(const type *s, bool until_end, size_t n) { \
                if (!s)                                                                  \
                        return 0;                                                        \
                                                                                         \
                size_t len = 0;                                                          \
                while ((until_end || len < n) && *s) {                                   \
                        s++;                                                             \
                        len++;                                                           \
                }                                                                        \
                                                                                         \
                return len;                                                              \
        }                                                                                \
        size_t strnlen##suffix(const type *s, size_t n) {                                \
                return _strnlen##suffix(s, false, n);                                    \
        }                                                                                \
        size_t strlen##suffix(const type *s) {                                           \
                return _strnlen##suffix(s, true, 0);                                     \
        }
```

medhefgo: Then again, I am not sure if shaving off a compare like that gains anything…

src/boot/efi/efi-string.c
yuwata: OK. Will drop the commit.

src/boot/efi/efi-string.c
yuwata: Yeah, I do not think it provides anything better... But if you'd like, please do it.

medhefgo (approved):

PR #23585, opened 2022-06-01
meson: install libsystemd-shared into pkglibdir

mbiebl:
Introduce rootpkglibdir for installing libsystemd-{shared,core}.so.
The benefit over using rootlibexecdir is that this path can be
Multi-Arch aware. This is something we'd like to make use of in
Debian/Ubuntu to make libsystemd-shared co-installable.

Using (pkg)libdir instead of libexecdir for installing a library also
is conceptually more correct.

See also https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=990547

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -146,6 +146,7 @@ rootlibdir = get_option('rootlibdir')
 if rootlibdir == ''
         rootlibdir = rootprefixdir / libdir.split('/')[-1]
 endif
+rootpkglibdir = rootlibdir / 'systemd'
 
 install_sysconfdir = get_option('install-sysconfdir') != 'false'
 install_sysconfdir_samples = get_option('install-sysconfdir') == 'true'
@@ -1984,7 +1985,7 @@ if conf.get('HAVE_LIBCRYPTSETUP_PLUGINS') == 1
                                         tpm2,
                                         versiondep],
                         link_depends : cryptsetup_token_sym,
-                        install_rpath : rootlibexecdir,
+                        install_rpath : rootpkglibdir,
                         install : true,
                         install_dir : libcryptsetup_plugins_dir)
         endif
@@ -2002,7 +2003,7 @@ if conf.get('HAVE_LIBCRYPTSETUP_PLUGINS') == 1
                                         libfido2,
                                         versiondep],
                         link_depends : cryptsetup_token_sym,
-                        install_rpath : rootlibexecdir,
+                        install_rpath : rootpkglibdir,
                         install : true,
                         install_dir : libcryptsetup_plugins_dir)
         endif
@@ -2020,7 +2021,7 @@ if conf.get('HAVE_LIBCRYPTSETUP_PLUGINS') == 1
                                         libp11kit,
                                         versiondep],
                         link_depends : cryptsetup_token_sym,
-                        install_rpath : rootlibexecdir,
+                        install_rpath : rootpkglibdir,
                         install : true,
                         install_dir : libcryptsetup_plugins_dir)
         endif
@@ -2155,7 +2156,7 @@ exe = executable(
                      libshared],
         dependencies : [versiondep,
                         libseccomp],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : true,
         install_dir : rootlibexecdir)
 dbus_programs += exe
@@ -2173,7 +2174,7 @@ exe = executable(
                      libshared],
         dependencies : [versiondep,
                         libseccomp],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : conf.get('ENABLE_ANALYZE'))
 public_programs += exe
 
@@ -2194,7 +2195,7 @@ executable(
                         liblz4,
                         libselinux,
                         libzstd],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : true,
         install_dir : rootlibexecdir)
 
@@ -2205,7 +2206,7 @@ public_programs += executable(
         link_with : [libjournal_core,
                      libshared],
         dependencies : [threads],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : true)
 
 public_programs += executable(
@@ -2219,7 +2220,7 @@ public_programs += executable(
                         liblz4,
                         libzstd,
                         libdl],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : true,
         install_dir : rootbindir)
 
@@ -2228,7 +2229,7 @@ executable(
         'src/getty-generator/getty-generator.c',
         include_directories : includes,
         link_with : [libshared],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : true,
         install_dir : systemgeneratordir)
 
@@ -2237,7 +2238,7 @@ executable(
         'src/debug-generator/debug-generator.c',
         include_directories : includes,
         link_with : [libshared],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
         install : true,
         install_dir : systemgeneratordir)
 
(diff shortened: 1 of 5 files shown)


mbiebl:
@keszybz would you be ok with this change? It should basically not affect Fedora.

If this patch is not accepted upstream, we'd have to carry it downstream which would make it unmaintainable due to the constant churn in meson.build.

meson.build
```diff
@@ -2155,7 +2156,7 @@ exe = executable(
                      libshared],
         dependencies : [versiondep,
                         libseccomp],
-        install_rpath : rootlibexecdir,
+        install_rpath : rootpkglibdir,
```
keszybz, line 2159: Hmm, the commit message talks about libraries, but this is about systemd exe itself (and other executables below)?

keszybz (commented):
Please adjust the commit title and message to actually describe what is being done… Maybe also give some example (useful for people who are not familiar with multiarch).

I think this will break the heuristics in systemd_installation_has_version(). That'd need to be adjusted too.

meson.build
mbiebl: Well, if you change the location of the library you have to adjust the rpath of the binaries. But that is an implementation detail. Am I missing something?

mbiebl:
> Please adjust the commit title and message to actually describe what is being done… 

Hm, I thought the title was pretty explicit. But I'm happy to take input on how this could be improved.

Maybe also give some example (useful for people who are not familiar with multiarch).
> 

Ok.

> I think this will break the heuristics in systemd_installation_has_version(). That'd need to be adjusted too.

Thankfully this is not the case. Multi-Arch doesn't mean you can mix different versions of systemd / libsystemd-shared.

meson.build
keszybz: Oh, it's `install_rpath` not `install_path`. My bad.

mbiebl:
> > > I think this will break the heuristics in systemd_installation_has_version(). That'd need to be adjusted too.
> > 
> > 
> > Thankfully this is not the case. Multi-Arch doesn't mean you can mix different versions of systemd / libsystemd-shared.
> 
> But will the new path still match the globs that are used there?

You are of course right. I've updated the heuristics in  `systemd_installation_has_version()` to also check in `/usr/lib/*/systemd/`, as this would be the path we'd like to use for libsystemd-shared.

mbiebl:
> > Maybe also give some example (useful for people who are not familiar with multiarch).
> 
> Ok.

Updated the commit message a little. Hopefully it is a bit clearer now.

mbiebl:
```
$ ./build-deb/test-nspawn-util /srv/chroot/sid-ma/
/* test_systemd_installation_has_version */
Found libsystemd shared at "/srv/chroot/sid-ma//usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
/srv/chroot/sid-ma/ has systemd >= 0: yes
Found libsystemd shared at "/srv/chroot/sid-ma//usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
/srv/chroot/sid-ma/ has systemd >= 231: yes
Found libsystemd shared at "/srv/chroot/sid-ma//usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
/srv/chroot/sid-ma/ has systemd >= 251: yes
Found libsystemd shared at "/srv/chroot/sid-ma//usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (too old).
/srv/chroot/sid-ma/ has systemd >= 999: no
```

src/nspawn/nspawn-util.c
```diff
@@ -23,6 +23,7 @@ int systemd_installation_has_version(const char *root, const char *minimal_versi
                         */
                        "/lib/systemd/libsystemd-shared-*.so",
                        "/lib64/systemd/libsystemd-shared-*.so",
+                       "/usr/lib/*/systemd/libsystemd-shared-*.so",
```
keszybz: Hmm, can we make this glob a bit tighter? There can be hundreds of directories that match `/usr/lib/*/` on a big installation, and resolving this glob can be a tad expensive. (E.g. on my machine, with caches cleared, it can be ~0.5 s). Maybe `/usr/lib/*-linux-*/systemd/libsystemd-shared-*.so` ?

mbiebl: Hm, this is on my old, messy laptop, with thousands of packets installed:
```
 $ find  /usr/lib/ -maxdepth 1 -type d | wc -l
150
$ time ls /usr/lib/*/systemd/libsystemd-shared-*.so
/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so

real	0m0,007s
user	0m0,000s
sys	0m0,007s
```

I actually thought about this but figured it's not much of an issue even if a tad inefficient, as this function is called so rarely. Especially, since you would call it inside/for an nspawn container, which is typically much more minimal and doesn't have 100s of directories.
We also exit early, so if you have a Fedora container, you shouldn't get the "expensive" glob, right?
Sorting this glob to the end of the list might be a good idea, though, so it is tried last.
I was also considering including `ROOTPKGLIBDIR "/libsystemd-shared-*.so"` in that list. So as long as the architecture matches (which should be the predominant case), we'd try an exact match first.

mbiebl: On the same system with the 150 directories
```
$ time ./build-deb/test-nspawn-util 
/* test_systemd_installation_has_version */
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 0: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 231: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 251: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (too old).
Current installation has systemd >= 999: no

real	0m0,011s
user	0m0,001s
sys	0m0,011s
```

mbiebl: Using `/usr/lib/*-linux-*/systemd/libsystemd-shared-*.so`:

```
$ time ./build-deb/test-nspawn-util 
/* test_systemd_installation_has_version */
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 0: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 231: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 251: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (too old).
Current installation has systemd >= 999: no

real	0m0,009s
user	0m0,005s
sys	0m0,004s
```
A very slight improvement

mbiebl: With `ROOTPKGLIBDIR` added to the search list:
```
$ time ./build-deb/test-nspawn-util 
/* test_systemd_installation_has_version */
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 0: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 231: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (OK).
Current installation has systemd >= 251: yes
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (too old).
Found libsystemd shared at "/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-251.so", version 251 (too old).
Current installation has systemd >= 999: no

real	0m0,009s
user	0m0,005s
sys	0m0,005s
```
No improvement

mbiebl: Given that using an explicit `ROOTPKGLIBDIR` didn't bring any improvements I went with your suggestion and used `*-linux-*`. I've updated the comment in the source code as well.

PR #23589, opened 2022-06-01
boot: Drop use of several gnu-efi functions

medhefgo:


man/loader.conf.xml
```diff
@@ -110,7 +110,14 @@
               </row>
```
yuwata, line 110: Commit message: collcation -> collation?

man/loader.conf.xml
```diff
@@ -110,7 +110,14 @@
...
             </tbody>
           </tgroup>
-        </table></listitem>
+        </table>
+
+        <para>Note that the patterns supported are limited compared to
+        <citerefentry><refentrytitle>glob</refentrytitle><manvolnum>7</manvolnum></citerefentry>: Only
+        <literal>?</literal>, <literal>*</literal> and <literal>[]</literal> (including ranges) are
```
yuwata: We usually add comma before "and": `<literal>*</literal>, and <literal>[]</literal>`

src/boot/efi/efi-string.c
```diff
@@ -173,6 +173,86 @@ char16_t *xstrdup16(const char16_t *s) {
...
+
+                case '[':
+                        if (h == '\0')
+                                /* Premature end of haystack. */
+                                return false;
+
+                        for (char16_t low = 0;;) {
+                                pattern++;
+                                p = TOLOWER(*pattern);
+                                if (!p || p == ']')
```
yuwata: `if (IN_SET(p, ']', '\0'))`?

From glob(7)
> Character classes
       An  expression  "[...]"  where  the first character after the leading '[' is not an '!' matches a single character, namely any of the characters enclosed by the brackets.  The string enclosed by the brackets cannot be empty;
       therefore ']' can be allowed between the brackets, provided that it is the first character.  (Thus, "[][!]" matches the three characters '[', ']', and '!'.)

However, this does not allow "[][!]".

src/boot/efi/efi-string.c
```diff
@@ -173,6 +173,86 @@ char16_t *xstrdup16(const char16_t *s) {
...
+                                        /* Invalid pattern or no match in set. */
+                                        return false;
+
+                                if (p == '-') {
+                                        /* Range pattern. */
+                                        pattern++;
+                                        p = TOLOWER(*pattern);
+                                        if (!low || !p || p == ']')
```
yuwata: `low` is integer. Please do `if (low == '\0' || IN_SET(p, ']', '\0'))`.

From glob(7)
> Ranges
       There is one special convention: two characters separated by '-' denote a range.  (Thus, "[A-Fa-f0-9]" is equivalent to "[ABCDEFabcdef0123456789]".)  One may include '-' in its literal meaning by making it the first or  last
       character between the brackets.  (Thus, "[]-]" matches just the two characters ']' and '-', and "[--0]" matches the three characters '-', '.', '0', since '/' cannot be matched.)

However, this refuses "[]-]".

src/boot/efi/efi-string.c
```diff
@@ -173,6 +173,86 @@ char16_t *xstrdup16(const char16_t *s) {
...
+                                                /* Invalid pattern. */
+                                                return false;
+
+                                        if (h >= low && h <= p)
+                                                /* Range match. */
+                                                break;
+
+                                        low = 0;
```
yuwata: `low = '\0'`?

src/boot/efi/test-efi-string.c
```diff
@@ -322,6 +322,43 @@ TEST(xstrdup16) {
         free(s);
 }
 
+TEST(metaimatch) {
+        assert_se(metaimatch(u"", u""));
+        assert_se(metaimatch(u"a", u"a"));
+        assert_se(metaimatch(u"A", u"a"));
+        assert_se(metaimatch(u"a", u"A"));
+        assert_se(metaimatch( u"abc", u"abc"));
```
yuwata: unnecessary space.

src/boot/efi/efi-string.c
```diff
@@ -173,6 +173,86 @@ char16_t *xstrdup16(const char16_t *s) {
...
+                case '?':
+                        if (h == '\0')
+                                /* Premature end of haystack. */
+                                return false;
+                        break;
+
+                case '*':
+                        do {
+                                /* Try matching haystack with remaining pattern. */
+                                if (metaimatch(haystack, pattern + 1))
```
yuwata: This recursion may be dangerous, e.g. 
`metaimatch("hoge", "******************************************************")`.

src/boot/efi/shim.c
```diff
@@ -102,7 +102,6 @@ static EFIAPI EFI_STATUS security2_policy_authentication (const EFI_SECURITY2_PR
 static EFIAPI EFI_STATUS security_policy_authentication (const EFI_SECURITY_PROTOCOL *this, UINT32 authentication_status,
```
yuwata, line 102: commit message: "n need" -> "no need" ?

yuwata (changes requested):

src/boot/efi/efi-string.c
medhefgo: `metaimatch` is intentionally written to exhibit the same (quirky) behavior as EDK2 `MetaiMatch` (since most firmware is written against that and we eventually want to call the firmware implementation). This probably wasn't clear enough, so I added comments to clarify. The previous version wasn't 100% there, which is now fixed. I even made sure by verifying all the test cases against the EDK2 function. I also adjusted the man page to hopefully make it clear it's not 100% fnmatch.

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,93 @@ DEFINE_STRCHR(char16_t, strchr16);
 DEFINE_STRNDUP(char, xstrndup8, strnlen8);
 DEFINE_STRNDUP(char16_t, xstrndup16, strnlen16);
 
+/* Patterns are fnmatch-like, but have to be compatible to EDK2 MetaiMatch behavior.
+ * See unit test cases for some corner-case examples. */
+static bool metaimatch_fallback(const char16_t *haystack, const char16_t *pattern, int max_depth) {
```
daandemeyer: We tend to suffix these kind of functions with "internal" so I'd name this `metaimatch_internal` instead

medhefgo: Once collation protocol is added/used, fallback makes more sense…

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,93 @@ DEFINE_STRCHR(char16_t, strchr16);
...
+
+                                        if (h >= low && h <= p)
+                                                /* Range match. */
+                                                break;
+                                }
+
+                                low = p;
+                                if (p == h)
+                                        /* Single char match. */
+                                        break;
```
yuwata: Why did you drop 'else' here?? Do you want to support `[-1-3]`?? Please add such a case in the test.

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,93 @@ DEFINE_STRCHR(char16_t, strchr16);
...
+                                pattern++;
+                                p = TOLOWER(*pattern);
+                                if (IN_SET(p, '\0', ']'))
+                                        /* Invalid pattern or no match in set. */
+                                        return false;
+
+                                if (p == '-') {
+                                        /* Range pattern. */
+                                        p = TOLOWER(*(pattern + 1));
+                                        if (IN_SET(p, '\0', ']'))
```
yuwata: Why did you drop the condition `low == '\0'` here?
It looks `"[-3]"` matches `"2"` for me. Please add the case to the test.

yuwata (changes requested):

src/boot/efi/efi-string.c
medhefgo: Yes, `MetaiMatch` is weird. It should already be covered by `assert_se(metaimatch(u"8", u"[1-7-9]")); /* ! */`

src/boot/efi/efi-string.c
medhefgo: This should already be covered by:
```c
        assert_se(metaimatch(u"+", u"[-@]")); /* ! */
        assert_se(!metaimatch(u"-", u"[-,]")); /* ! */
```

medhefgo:
An alternative would be to not bother with MetaiMatch and instead make the syntax fnmatch compatible and use the collation protocol's StriColl to compare the chars (not sure about the cost of calling that function for single chars, though).

medhefgo:
Now that I think about it: making this fnmatch compatible is the way to go. All it really takes is lower-casing the pattern.

medhefgo:
It should be fnmatch compatible now (which is enforced by the unit test). I also discovered an interesting edge case: `[a-c-f]` is equivalent to `[-abcf]`. I honestly expected that pattern to be invalid.

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,109 @@ DEFINE_STRCHR(char16_t, strchr16);
...
+
+                                /* Range pattern if '-' is not first or last in set. */
+                                if (*p == '-' && can_range && !first && *(p + 1) != ']') {
+                                        p++;
+                                        if (*p == '\\')
+                                                p++;
+                                        if (*p == '\0')
+                                                return false;
+
+                                        match |= (*(p - 2) <= *h && *h <= *p);
```
yuwata: `(p - 2)` seems not correct. How about the case when the pattern include `'\\'`, i.e. `"[a-\\x]"`?

src/boot/efi/boot.c
```diff
@@ -1170,7 +1170,7 @@ static void config_defaults_load_from_file(Config *config, CHAR8 *content) {
                 }
 
                 if (streq8((char *) key, "default")) {
-                        if (value[0] == '@' && !streq8((char *) value, "@saved")) {
+                        if (value[0] == '@' && !strcaseeq8((char *) value, "@saved")) {
```
yuwata, line 1172: Is this change relevant to the commit which introduces fnmatch? If not, then please move to another commit.

src/boot/efi/boot.c
```diff
@@ -1606,6 +1606,11 @@ static void config_load_defaults(Config *config, EFI_FILE *root_dir) {
 
         (void) efivar_get(LOADER_GUID, L"LoaderEntryDefault", &config->entry_default_efivar);
 
+        strtolower16(config->entry_default_config);
+        strtolower16(config->entry_default_efivar);
+        strtolower16(config->entry_oneshot);
+        strtolower16(config->entry_saved);
```
yuwata, line 1611: Also here. Is this change relevant to the commit which introduces fnmatch? If not, then please move to another commit.

yuwata (changes requested):

src/boot/efi/efi-string.c
medhefgo: Nice catch. This is now covered by this test: `[$-\\+]`

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,110 @@ DEFINE_STRCHR(char16_t, strchr16);
...
 
+/* Patterns are fnmatch-compatible (with reduced feature support). */
+static bool efi_fnmatch_internal(const char16_t *p, const char16_t *h, int max_depth) {
+        assert(p);
+        assert(h);
+
+        if (max_depth == 0)
+                return false;
+
+        for (;; p++, h++) {
```
yuwata: `{` can be dropped.

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,110 @@ DEFINE_STRCHR(char16_t, strchr16);
...
+                        for (;; first = false) {
+                                p++;
+                                if (*p == '\0')
+                                        return false;
+
+                                if (*p == '\\') {
+                                        p++;
+                                        if (*p == '\0')
+                                                return false;
+                                        match |= *p == *h;
```
yuwata, line 200: Not familiar with C standard, but should this be `match = match || (*p == *h);`??

src/boot/efi/efi-string.c
```diff
@@ -139,6 +139,110 @@ DEFINE_STRCHR(char16_t, strchr16);
...
+                                /* Range pattern if '-' is not first or last in set. */
+                                if (*p == '-' && can_range && !first && *(p + 1) != ']') {
+                                        char16_t low = *(p - 1);
+                                        p++;
+                                        if (*p == '\\')
+                                                p++;
+                                        if (*p == '\0')
+                                                return false;
+
+                                        match |= (low <= *h && *h <= *p);
```
yuwata: Also here. `match = match || (low <= *h && *h <= *p)`??

yuwata (approved):
LGTM. Several minor comments.

src/boot/efi/efi-string.c
medhefgo: For bools it's equivalent. But I can change it if desired.

yuwata (approved):
LGTM. But I'd like to add fuzzer or any better way to test `efi_fnmatch()` in a later PR. @mrc0mmand and @evverx Do you have any idea about that?

src/boot/efi/efi-string.c
yuwata: I cannot find reference that bitwise operator `|` can be used for bool (or cannot used). So, I have no idea. But we do not usually use `|` for boolean but use `||`.

evverx:
> But I'd like to add fuzzer or any better way to test efi_fnmatch() in a later PR. @mrc0mmand and @evverx Do you have any idea about that?

Looking at https://seclists.org/oss-sec/2022/q2/178 I think it certainly makes sense to fuzz bootloaders :-) But external dependencies like `gnu-efi` prevent some code from being fuzzed on OSS-Fuzz. If by analogy with `fuzz-bcd` it's possible to test that code without `gnu-efi` I think it should be doable.

medhefgo:
efi-string.c was designed to be tested in userspace so it should be easy enough to add fuzzers. Though, I am not sure how to go about feeding the data blob we get from the fuzzer to those mem/string functions in a meaningful way.

evverx:
If there are no high-level functions receiving only strings of files I think it should be possible to parse that blob and extract strings, numbers and so on and then pass them to the string/mem functions. To impose that structure something  like `LLVMFuzzerCustomMutator` could be used to generate "valid" inputs.

evverx:
I started playing with `efi_fnmatch` and generated a couple of string crashing it
```
efi_fnmatch(u"袈袈*?*?㼀", u"袈袈뎳\001灠Ѐ塾\x5fd\a\xda70\xdadaߚ瀀\xdada\xdada\x7b3瀀\xdada\xdada墳\x5fd\a\xda70\xdada돚\xdada墳\x5fd\a\xda70\xdada돚");
```
```
==2334016==ERROR: AddressSanitizer: global-buffer-overflow on address 0x0000005673c2 at pc 0x00000054cb90 bp 0x7ffcb53be300 sp 0x7ffcb53be2f8
READ of size 2 at 0x0000005673c2 thread T0
    #0 0x54cb8f in efi_fnmatch_internal /home/vagrant/systemd/build/../src/boot/efi/efi-string.c:180:34
    #1 0x54c7d9 in efi_fnmatch_internal /home/vagrant/systemd/build/../src/boot/efi/efi-string.c:175:37
    #2 0x54b6b5 in efi_fnmatch /home/vagrant/systemd/build/../src/boot/efi/efi-string.c:243:16
```
It could be I messed something up trying to extract two `char16_t` strings though but it was just a prototype :-)

yuwata:
Hmm, I added the testcase into test-efi-string.c and ran with valgrind, but valgrind does not detect the issue.

evverx:
Could you try building with ASan:
```diff
diff --git a/src/boot/efi/test-efi-string.c b/src/boot/efi/test-efi-string.c
index 5aaa1f713f..da1a471d36 100644
--- a/src/boot/efi/test-efi-string.c
+++ b/src/boot/efi/test-efi-string.c
@@ -368,6 +368,7 @@ TEST(efi_fnmatch) {
         TEST_FNMATCH_ONE("[b]", "z-a", false);
         TEST_FNMATCH_ONE("[a\\-z]", "b", false);
         TEST_FNMATCH_ONE("?a*b[.-0]c", "/a/b/c", true);
+        efi_fnmatch(u"袈袈*?*?㼀", u"袈袈뎳\001灠Ѐ塾\x5fd\a\xda70\xdadaߚ瀀\xdada\xdada\x7b3瀀\xdada\xdada墳\x5fd\a\xda70\xdada돚\xdada墳\x5fd\a\xda70\xdada돚");
 }

 TEST(efi_memcmp) {
```
```
meson -Db_sanitize=address build
ninja -C ./build/ -v test-efi-string
./build/test-efi-string
==2338663==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000000412762 at pc 0x00000040907d bp 0x7ffc58da48e0 sp 0x7ffc58da48d8
READ of size 2 at 0x000000412762 thread T0
    #0 0x40907c in efi_fnmatch_internal ../src/boot/efi/efi-string.c:180
    #1 0x409032 in efi_fnmatch_internal ../src/boot/efi/efi-string.c:175
    #2 0x4096a7 in efi_fnmatch ../src/boot/efi/efi-string.c:243
    #3 0x40712d in test_efi_fnmatch ../src/boot/efi/test-efi-string.c:371
    #4 0x401978 in run_test_table ../src/shared/tests.h:106
    #5 0x407de3 in main ../src/boot/efi/test-efi-string.c:426
    #6 0x7f0eaea9143f in __libc_start_call_main (/lib64/libc.so.6+0x4043f)
    #7 0x7f0eaea914ef in __libc_start_main_alias_2 (/lib64/libc.so.6+0x404ef)
    #8 0x401224 in _start (/home/vagrant/systemd/build/test-efi-string+0x401224)
```

yuwata:
Yeah, sanitizer detects the issue. Thanks.

PR #23591, opened 2022-06-01
core: default {State,Cache,Log,Configuration}DirectoryMode to 0750

klemensn:
These directories may contain sensitive data which must be protected,
but world-readable by default is not safe.

The recent CVE-2022-32198[0] shows how easy it is to deploy services
with unit files lacking proper `*DirectoryMode` directives, thus
exposing passwords to the world.

Use `0750` aka. `u=rwx,g=rx,o=` as a more sensitive default;
services that need to expose data can still do so explicitly by
overriding the respective `*Mode` value.

0: https://github.com/NixOS/nixpkgs/pull/175729
   "jellyfin: fix permissions on state directory"

### DIFF
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -1421,7 +1421,7 @@ StateDirectory=aaa/bbb ccc</programlisting>
         <listitem><para>Specifies the access mode of the directories specified in <varname>RuntimeDirectory=</varname>,
         <varname>StateDirectory=</varname>, <varname>CacheDirectory=</varname>, <varname>LogsDirectory=</varname>, or
         <varname>ConfigurationDirectory=</varname>, respectively, as an octal number.  Defaults to
-        <constant>0755</constant>. See "Permissions" in <citerefentry
+        <constant>0750</constant>. See "Permissions" in <citerefentry
         project='man-pages'><refentrytitle>path_resolution</refentrytitle><manvolnum>7</manvolnum></citerefentry> for a
         discussion of the meaning of permission bits.</para></listitem>
       </varlistentry>
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -5299,7 +5299,7 @@ void exec_context_init(ExecContext *c) {
         c->timer_slack_nsec = NSEC_INFINITY;
         c->personality = PERSONALITY_INVALID;
         for (ExecDirectoryType t = 0; t < _EXEC_DIRECTORY_TYPE_MAX; t++)
-                c->directories[t].mode = 0755;
+                c->directories[t].mode = 0750;
         c->timeout_clean_usec = USEC_INFINITY;
         c->capability_bounding_set = CAP_ALL;
         assert_cc(NAMESPACE_FLAGS_INITIAL != NAMESPACE_FLAGS_ALL);


poettering:
I am not convinced. This is a compat break. Various subsystems expect to be able to read each others files, and they have every right to.

I am pretty sure password files and suchlike should be indiviudally protected. Relying on the surrounding directory is a poor idea, already because of hardlinks and stuff. We shouldn't make it our problem, if apps are so extremely sloppy, that they cannot even be bothered to write their own files with secrets with a strict access modes...

I think it definitely deserves explicit mention in the man pages though. 

But I am not convinced this PR like this is desirable and it is a major compat break.

medhefgo:
Here is an idea: how about a version field in the `[Unit]` section that lets services target a certain unit API level? Then whenever there is something that should really be changed but cannot because of compat is put behind a new version.

lheckemann:
>they have every right to.

I think that when this is necessary, it's good to have it explicit. Either via explicitly world-readable directories, or via group membership. systemd already provides great facilities for implementing this.

>I am pretty sure password files and suchlike should be indiviudally protected.

Yes. But sloppy software exists, and defence in depth helps.

I agree that this is a massive break in compatibility, but think it _would have been good to have it this way from the beginning_, and don't think that -- in the long term -- compatibility is a good reason not to move towards better defaults.

Another idea: `systemd-analyze security` should warn about this. I'm not sure if this is possible, but I think it would be nice if it assigned a relatively high score to 0755 mode _only if it comes from the default_.

klemensn:
> > they have every right to.
> 
> I think that when this is necessary, it's good to have it explicit. Either via explicitly world-readable directories, or via group membership. systemd already provides great facilities for implementing this.
> 
> > I am pretty sure password files and suchlike should be indiviudally protected.
> 
> Yes. But sloppy software exists, and defence in depth helps.
> 
> I agree that this is a massive break in compatibility, but think it _would have been good to have it this way from the beginning_, and don't think that -- in the long term -- compatibility is a good reason not to move towards better defaults.

I lack insight into systemd development and deployment outside of my own systems, but I tend to agree with Linus here that issues like the recent CVE could have probably been prevented if safe defaults had been in place from the beginning.
 
> Another idea: `systemd-analyze security` should warn about this. I'm not sure if this is possible, but I think it would be nice if it assigned a relatively high score to 0755 mode _only if it comes from the default_.

This should be possible but I'm pessimistic about the usefulness, given how discouraging the output of `systemd-analyze security`
already is on a regular NixOS desktop system:
```
UNIT                                  EXPOSURE PREDICATE HAPPY
ModemManager.service                       6.3 MEDIUM    😐
NetworkManager-dispatcher.service          9.6 UNSAFE    😨
NetworkManager.service                     7.8 EXPOSED   🙁
accounts-daemon.service                    5.5 MEDIUM    😐
bluetooth.service                          6.0 MEDIUM    😐
dbus.service                               9.6 UNSAFE    😨
display-manager.service                    9.6 UNSAFE    😨
earlyoom.service                           9.6 UNSAFE    😨
emergency.service                          9.5 UNSAFE    😨
getty@tty1.service                         9.6 UNSAFE    😨
logrotate.service                          9.6 UNSAFE    😨
nix-daemon.service                         9.6 UNSAFE    😨
nix-gc.service                             9.6 UNSAFE    😨
nix-optimise.service                       9.6 UNSAFE    😨
nscd.service                               8.2 EXPOSED   🙁
pcscd.service                              9.6 UNSAFE    😨
polkit.service                             9.6 UNSAFE    😨
reload-systemd-vconsole-setup.service      9.6 UNSAFE    😨
rescue.service                             9.5 UNSAFE    😨
rtkit-daemon.service                       7.2 MEDIUM    😐
sshd.service                               9.6 UNSAFE    😨
systemd-ask-password-console.service       9.4 UNSAFE    😨
systemd-ask-password-wall.service          9.4 UNSAFE    😨
systemd-journald.service                   4.3 OK        🙂
systemd-logind.service                     2.8 OK        🙂
systemd-timesyncd.service                  2.1 OK        🙂
systemd-udevd.service                      6.7 MEDIUM    😐
udisks2.service                            9.6 UNSAFE    😨
upower.service                             2.4 OK        🙂
user@1000.service                          9.4 UNSAFE    😨
wpa_supplicant.service                     9.6 UNSAFE    😨
zfs-zed.service                            9.6 UNSAFE    😨
zpool-trim.service                         9.6 UNSAFE    😨
```

I have my doubts that folks seriously
1. users go through this output
2. decide to take action
3. see a warning default `0755` modes in some unit specific output
4. go and fix the unit

(I'd be glad to be wrong on this one...)

poettering:
> Another idea: `systemd-analyze security` should warn about this. I'm not sure if this is possible, but I think it would be nice if it assigned a relatively high score to 0755 mode _only if it comes from the default_.

That might be OK, consider filing an issue about that.

I still think this is mostly a documentation issue though: we should clearly state what the issue here is.

In general, I am not convinced this is that much of an issue really If software deals with sensitive data it has to be careful about it, there's no way around it. And if it doesn't get something as basic as unix file access modes right, then I am pretty sure there's so much wrong with the software, that this is the least of the problems.

There's also the fact that it's typically a good idea if admins do much of their stuff as regular user instead of root. if you hide non-sensitive stuff away by making it unreadable to root, you thus force the admin to raise their privileges all the time to root — unnecessarily! But that's not a good idea. Hence, so far the thinking in most Linux services actually is: make things world-readable, but allow only the service user itself to write, i.e. go for 0755 or so by default, and then individually lock down the more sensitive stuff to 0700 or so. And I think that's a good default behaviour.

Hence, I think yes, we should update docs and warn users. And yes, it might make sense to make "systemd-analyze security" pick up on this stuff. But changing defaults? Sorry, but I am not convinced.

PR #23596, opened 2022-06-02
Silence messages from libbpf

keszybz:
For https://bugzilla.redhat.com/show_bug.cgi?id=2084955#c14. ~~It probably doesn't fix the issue completely, because there should be still one more message `Failed to load BPF object`.~~

### DIFF
--- a/src/core/bpf-firewall.c
+++ b/src/core/bpf-firewall.c
@@ -543,7 +543,7 @@ int bpf_firewall_compile(Unit *u) {
                 return supported;
         if (supported == BPF_FIREWALL_UNSUPPORTED)
                 return log_unit_debug_errno(u, SYNTHETIC_ERRNO(EOPNOTSUPP),
-                                            "BPF firewalling not supported on this manager, proceeding without.");
+                                            "bpf-firewall: BPF firewalling not supported, proceeding without.");
         if (supported != BPF_FIREWALL_SUPPORTED_WITH_MULTI && u->type == UNIT_SLICE)
                 /* If BPF_F_ALLOW_MULTI is not supported we don't support any BPF magic on inner nodes (i.e. on slice
                  * units), since that would mean leaf nodes couldn't do any BPF anymore at all. Under the assumption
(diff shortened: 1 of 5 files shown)


keszybz:
Seems to work fine.

src/core/bpf-util.c
```diff
@@ -0,0 +1,37 @@
...
+
+        r = cg_unified_controller(SYSTEMD_CGROUP_CONTROLLER);
+        if (r < 0) {
+                log_warning_errno(r, "Can't determine whether the unified hierarchy is used: %m");
+                return (supported = false);
+        }
+
+        if (r == 0) {
+                log_info_errno(SYNTHETIC_ERRNO(EOPNOTSUPP),
+                               "Not running with unified cgroup hierarchy, BPF cgroup filters are not supported");
```
daandemeyer: Copy paste error with the log messages?

src/core/bpf-util.c
```diff
@@ -0,0 +1,37 @@
...
+                return (supported = false);
+        }
+
+        r = dlopen_bpf();
+        if (r < 0) {
+                log_info_errno(r, "Failed to open libbpf, BPF cgroup filters disabled: %m");
```
daandemeyer: Ditto

src/core/bpf-util.c
keszybz: Hmm, can you be more specific?

daandemeyer: > BPF cgroup filters are not supported

This looked specific for one use case of BPF but I might understand BPF wrong

keszybz: "BPF cgroup filters" was supposed to be a generic term that covers the types of filters we attach… I think it's OK,  but suggestions are welcome.

daandemeyer: I got tripped up by the function being named bpf_supported(). I figure we might use bpf for something different than cgroup filters in the future and then the message would be wrong if we also use bpf_supported() there to check for bpf support. 

So I'd say either the function should be renamed to something more specific to indicate it's checking more than just whether bpf is supported or the error message should be updated to say we don't support any BPF features unless we're using the unified hierarchy, so something like "Not running with unified cgroup hierarchy, disabling BPF features"

keszybz: I renamed the function to `cgroup_bpf_supported()` and adjusted the texts a bit.

yuwata (approved):

PR #23601, opened 2022-06-02
man: add missing argument to systemd-creds synopsis

aafeijoo-suse:
Add missing `COMMAND` and `ARGS` to systemd-creds synopsis man page, so it should be:

```
SYNOPSIS
       systemd-creds [OPTIONS...] COMMAND [ARGS...]
```

### DIFF
--- a/man/systemd-creds.xml
+++ b/man/systemd-creds.xml
@@ -25,6 +25,8 @@
     <cmdsynopsis>
       <command>systemd-creds</command>
       <arg choice="opt" rep="repeat">OPTIONS</arg>
+      <arg choice="plain">COMMAND</arg>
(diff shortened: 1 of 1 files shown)


man/systemd-creds.xml
```diff
@@ -25,6 +25,7 @@
     <cmdsynopsis>
       <command>systemd-creds</command>
       <arg choice="opt" rep="repeat">OPTIONS</arg>
+      <arg choice="plain">COMMAND</arg>
```
keszybz, line 28: I think it would be nice to add `[ARGS]` or `[ARGUMENTS]` after `COMMAND`.

aafeijoo-suse: Done.

PR #23611, opened 2022-06-03
resolve: known-answer suppression fix for ageing entries

vchillara:
Optimization for known answer suppression for ageing entries.
RFC 6762 section 7.1 suggests that a Multicast DNS querier SHOULD NOT include records in the Known-Answer list whose remaining TTL is less than half of their original TTL.
This is useful in general to keep the information in cache current, but particularly so when browsing for mDNS services (PR #22532 ).
Skipping cached records whose age is >50% TTL from the known answers enables us to receive updated info from the network if the service is still alive.

### DIFF
--- a/src/resolve/resolved-dns-cache.c
+++ b/src/resolve/resolved-dns-cache.c
@@ -1252,10 +1252,13 @@ int dns_cache_check_conflicts(DnsCache *cache, DnsResourceRecord *rr, int owner_
 int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
         unsigned ancount = 0;
         DnsCacheItem *i;
+        usec_t t;
         int r;
 
         assert(cache);
         assert(p);
+
+        t = now(CLOCK_BOOTTIME);
 
         HASHMAP_FOREACH(i, cache->by_key)
                 LIST_FOREACH(by_key, j, i) {
@@ -1263,6 +1266,11 @@ int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
                                 continue;
 
                         if (!j->shared_owner)
+                                continue;
+
+                        /* RFC6762 7.1: Don't append records with less than half the TTL remaining
+                         * as known answers. */
+                        if (usec_sub_unsigned(j->until, t) < j->rr->ttl * USEC_PER_SEC / 2)
                                 continue;
 
                         r = dns_packet_append_rr(p, j->rr, 0, NULL, NULL);


src/resolve/resolved-dns-cache.c
```diff
@@ -1265,6 +1266,14 @@ int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
                         if (!j->shared_owner)
                                 continue;
 
+                        if (t <= 0)
+                                t = now(CLOCK_BOOTTIME);
```
medhefgo: This should be done outside the loop.

src/resolve/resolved-dns-cache.c
```diff
@@ -1265,6 +1266,14 @@ int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
...
+
+                        /* RFC6762 7.1: Don't append records with less than
+                         * half the TTL remaining as known answers. */
+                        if ((t >= j->until) || (j->until - t) < (0.5 * (j->rr->ttl * USEC_PER_SEC)))
```
medhefgo: I doubt thing this needs float math:
```suggestion
                        if ((t >= j->until) || (j->until - t) < (j->rr->ttl * (USEC_PER_SEC / 2)))
```

src/resolve/resolved-dns-cache.c
vchillara: Moved outside

src/resolve/resolved-dns-cache.c
vchillara: Changed as suggested

vchillara:
@medhefgo addressed your comments. Any further feedback?

src/resolve/resolved-dns-cache.c
```diff
@@ -1265,6 +1268,11 @@ int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
                         if (!j->shared_owner)
                                 continue;
 
+                        /* RFC6762 7.1: Don't append records with less than
+                         * half the TTL remaining as known answers. */
+                        if ((t >= j->until) || (j->until - t) < (j->rr->ttl * (USEC_PER_SEC / 2)))
```
yuwata: Please drop redundant `()`, and use `usec_sub_unsigned()`:
`if (usec_sub_unsigned(t, j->until) < j->rr->ttl * USEC_PER_SEC / 2)`

src/resolve/resolved-dns-cache.c
```diff
@@ -1265,6 +1268,11 @@ int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
                         if (!j->shared_owner)
                                 continue;
 
+                        /* RFC6762 7.1: Don't append records with less than
+                         * half the TTL remaining as known answers. */
```
yuwata: Not a big matter, but it is not necessary to wrap comment so frequently. We allow about 109ch per line.

src/resolve/resolved-dns-cache.c
```diff
@@ -1252,11 +1252,14 @@ int dns_cache_check_conflicts(DnsCache *cache, DnsResourceRecord *rr, int owner_
 int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
```
yuwata, line 1252: But please wrap the long line in the commit message.

yuwata (changes requested):

src/resolve/resolved-dns-cache.c
```diff
@@ -1265,6 +1268,12 @@ int dns_cache_export_shared_to_packet(DnsCache *cache, DnsPacket *p) {
                         if (!j->shared_owner)
                                 continue;
 
+                        /* RFC6762 7.1: Don't append records with less than half the TTL remaining
+                         * as known answers.
+                         */
```
yuwata: Not a fun of unnecessary line break at the end of comment. But, not a big matter.

yuwata (approved):
LGTM.

PR #23612, opened 2022-06-03
meson: Add nspawn-locale meson option

daandemeyer:
https://github.com/systemd/systemd/pull/23192 caused breakage in
Arch Linux's build tooling. Let's give users an opt-out aside from
reverting the patch. It's hardly any maintenance work on our side
and gives users an easy way to revert the locale change if needed.

Of course, by default we still pick C.UTF-8 if the option is not
specified.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -302,6 +302,8 @@ option('gshadow', type : 'boolean',
        description : 'support for shadow group')
 option('default-locale', type : 'string', value : '',
        description : 'default locale used when /etc/locale.conf does not exist')
+option('nspawn-locale', type : 'string', value : 'C.UTF-8',
+       description : 'default locale used by systemd-nspawn when executing commands in a container')
 option('localegen-path', type : 'string', value : '',
        description : 'absolute path to the locale-gen binary in case the system is using locale-gen')
 option('service-watchdog', type : 'string', value : '3min',
--- a/meson.build
+++ b/meson.build
@@ -933,6 +933,9 @@ if default_locale == ''
 endif
 conf.set_quoted('SYSTEMD_DEFAULT_LOCALE', default_locale)
 
+nspawn_locale = get_option('nspawn-locale')
+conf.set_quoted('SYSTEMD_NSPAWN_LOCALE', nspawn_locale)
+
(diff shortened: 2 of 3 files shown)


medhefgo:
A link would be nice. I am wondering how this breaks arch tooling when arch supports C.UTF-8 (afaik).

nl6720:
It breaks because `C.UTF-8` is not a [builtin](https://sourceware.org/glibc/wiki/Proposals/C.UTF-8#Builtin), so it is not available by default. See https://bugs.archlinux.org/task/74864 for details.

meson_options.txt
```diff
@@ -302,6 +302,8 @@ option('gshadow', type : 'boolean',
        description : 'support for shadow group')
 option('default-locale', type : 'string', value : '',
        description : 'default locale used when /etc/locale.conf does not exist')
+option('nspawn-locale', type : 'string', value : 'C.UTF-8',
+       description : 'default locale used by systemd-nspawn when executing commands in a container')
```
yuwata, line 306: I think it is not necessary to introduce a new knob, but please reuse `default-locale`.

yuwata (changes requested):

meson_options.txt
daandemeyer: This was discussed in an earlier PR where I tried to do this and it was rejected. There's no guarantee that the default locale from the host is also the right locale to use for nspawn. I'm on mobile now but I'll link the PR when I get back to a PC.

daandemeyer: See discussion here: https://github.com/systemd/systemd/pull/23252

yuwata: OK. Thank you for the pointer.

yuwata: From https://github.com/systemd/systemd/pull/23252#issuecomment-1118678927
> I suspect that if the payload has no support for `C.UTF-8`, it'll fall back to `C` anyway. So this shouldn't matter too much. Let's keep it as is, and if people start complaining because their old payloads emit a warning about unknown locale, we can always make this (separately) configurable.

That's make sense for me.

yuwata (approved):

PR #23613, opened 2022-06-03
meson: Switch default-locale default to C.UTF-8

daandemeyer:
We're already using C.UTF-8 as the default locale for nspawn. Let's
make the same change for the default-locale option instead of deciding
what to use based on the locale used by the host system. Users can
still override the locale using the default-locale option if needed.


medhefgo (approved):
Can't wait for people to come crying about this. :smile_cat:

keszybz:
Yeah, let's just do this.

eli-schwartz:
The reason anyone cared about objecting to the use of C.UTF-8 here was back when there was no such actual locale. The option was created to cater to those people, and it's actually a really good point. The exact implementation of defaults selection is probably a bikeshed issue...

Things have changed a bit since then. glibc 2.35 has an upstream C.UTF-8 locale, which works on more systems than just Debian and Fedora (i.e. if it doesn't work everywhere, then someone has been naughty by disabling it).

yuwata (approved):

PR #23615, opened 2022-06-03
core: refuse daemon-reload request during boot time

steelman:
daemon-reload intereferes with boot process. If invoked from a service
being started, it may interrupt starting of other services. We have
observed user@X.service being closed because a login process after
opening session via pam_systemd was removed from session-X.scope. Most
probably daemon-reload should not be invoked this way, however, we think
such invocations should be explicitely blocked or at least logged.

Signed-off-by: Hyotaek Shim <hyotaek.shim@samsung.com>
Signed-off-by: Łukasz Stelmach <l.stelmach@samsung.com>

### DIFF
--- a/src/core/dbus-manager.c
+++ b/src/core/dbus-manager.c
@@ -1463,6 +1463,11 @@ static int method_reload(sd_bus_message *message, void *userdata, sd_bus_error *
         if (r < 0)
                 return r;
 
+        if (!MANAGER_IS_FINISHED(m)) {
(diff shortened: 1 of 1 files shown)


steelman:
Apparently `initrd-parse-etc.service` does exactly this and we broke it )-:

keszybz:
Yeah, I don't think this is the right approach. If a job got lost due to a daemon-reload, that's a significant bug that needs to be fixed.

PR #23616, opened 2022-06-03
Add macros that allocate a fixed buffer for in_addr_to_string(), in_addr_prefix+to_string()

keszybz:


src/basic/in-addr-util.h
```diff
@@ -68,14 +69,45 @@ int in_addr_prefix_range(
...
+                char *buf,
+                size_t buf_len);
+
+static inline const char* _in_addr_prefix_to_string(
+                int family,
+                const union in_addr_union *u,
+                unsigned prefixlen,
+                char *buf,
+                size_t buf_len) {
+        int r = in_addr_prefix_to_string(family, u, prefixlen, buf, buf_len);
```
mrc0mmand: Maybe drop `r` altogether and do just

```c
       if (in_addr_prefix_to_string(family, u, prefixlen, buf, buf_len) < 0)
              return NULL;
```
?

keszybz: Ah, good point. I'll adjust it like that on the next push.

src/network/networkd-routing-policy-rule.c
```diff
@@ -427,14 +427,14 @@ static void log_routing_policy_rule_debug(const RoutingPolicyRule *rule, const c
                 return;
 
         (void) network_config_state_to_string_alloc(rule->state, &state);
-        (void) in_addr_prefix_to_string(rule->family, &rule->from, rule->from_prefixlen, &from);
-        (void) in_addr_prefix_to_string(rule->family, &rule->to, rule->to_prefixlen, &to);
```
yuwata, line 431:
```
../src/network/networkd-routing-policy-rule.c:417:59: error: unused variable 'to' [-Werror,-Wunused-variable]
        _cleanup_free_ char *state = NULL, *from = NULL, *to = NULL, *table = NULL;
                                                          ^
../src/network/networkd-routing-policy-rule.c:417:45: error: unused variable 'from' [-Werror,-Wunused-variable]
        _cleanup_free_ char *state = NULL, *from = NULL, *to = NULL, *table = NULL;
                                            ^
2 errors generated.
```

src/basic/in-addr-util.h
```diff
@@ -68,10 +69,20 @@ int in_addr_prefix_range(
...
 static inline int in6_addr_to_string(const struct in6_addr *u, char **ret) {
         return in_addr_to_string(AF_INET6, (const union in_addr_union*) u, ret);
 }
+
+/* Note: the lifetime of the compound literal is the immediately surrounding block,
+ * see C11 §6.5.2.5, and
+ * https://stackoverflow.com/questions/34880638/compound-literal-lifetime-and-if-blocks */
+#define IN_ADDR_MAX CONST_MAX(INET_ADDRSTRLEN, INET6_ADDRSTRLEN)
+#define IN_ADDR_TO_STRING(family, addr) inet_ntop(family, addr, (char[IN_ADDR_MAX]){}, IN_ADDR_MAX)
+#define IN4_ADDR_TO_STRING(addr) IN_ADDR_TO_STRING(AF_INET, addr)
```
yuwata: Not a matter, but we can use `INET_ADDRSTRLEN` for IPv4 case.

src/libsystemd-network/sd-ipv4ll.c
```diff
@@ -257,8 +256,8 @@ static int ipv4ll_pick_address(sd_ipv4ll *ll) {
         } while (addr == ll->address ||
                  IN_SET(be32toh(addr) & 0x0000FF00U, 0x0000U, 0xFF00U));
 
-        (void) in_addr_to_string(AF_INET, &(union in_addr_union) { .in.s_addr = addr }, &address);
-        log_ipv4ll(ll, "Picked new IP address %s.", strna(address));
+        log_ipv4ll(ll, "Picked new IP address %s.",
+                   IN4_ADDR_TO_STRING(&(union in_addr_union) { .in.s_addr = addr }));
```
yuwata: It is not necessary to create a new compound literal here, simply `&addr` can be used.

src/basic/in-addr-util.h
```diff
@@ -68,10 +69,20 @@ int in_addr_prefix_range(
...
 int in_addr_to_string(int family, const union in_addr_union *u, char **ret);
 static inline int in6_addr_to_string(const struct in6_addr *u, char **ret) {
         return in_addr_to_string(AF_INET6, (const union in_addr_union*) u, ret);
 }
+
+/* Note: the lifetime of the compound literal is the immediately surrounding block,
+ * see C11 §6.5.2.5, and
+ * https://stackoverflow.com/questions/34880638/compound-literal-lifetime-and-if-blocks */
+#define IN_ADDR_MAX CONST_MAX(INET_ADDRSTRLEN, INET6_ADDRSTRLEN)
+#define IN_ADDR_TO_STRING(family, addr) inet_ntop(family, addr, (char[IN_ADDR_MAX]){}, IN_ADDR_MAX)
```
yuwata: BTW, I'd like to introduce a tiny inline wrapper of `inet_ntop()` to introduce type safety.
```
static inline typesafe_inet_ntop(int family, const union in_addr_union *a, char *buf, size_t len) {
        return inet_ntop(family, a, buf, len);
}
static inline typesafe_inet_ntop4(const struct in_addr *a, char *buf, size_t len) {
        return inet_ntop(family, a, buf, len);
}
static inline typesafe_inet_ntop6(const struct in6_addr *a, char *buf, size_t len) {
        return inet_ntop(family, a, buf, len);
}
```

src/libsystemd-network/sd-ndisc.c
```diff
@@ -229,8 +228,8 @@ static int ndisc_recv(sd_event_source *s, int fd, uint32_t revents, void *userda
 
                 switch (r) {
                 case -EADDRNOTAVAIL:
-                        (void) in_addr_to_string(AF_INET6, (const union in_addr_union*) &rt->address, &addr);
-                        log_ndisc(nd, "Received RA from non-link-local address %s. Ignoring", addr);
+                        log_ndisc(nd, "Received RA from non-link-local address %s. Ignoring.",
+                                  IN6_ADDR_TO_STRING((const union in_addr_union*) &rt->address));
```
yuwata: The cast can be dropped.

keszybz:
Updated. Now both sets of macros use typesafe helpers. I added a few more changes.

mbiebl:
@keszybz I moved the last commit into a separate PR #23633.
Hope that's ok.

keszybz:
I changed the `_in_addr_prefix_to_string()` wrapper to `assert` success. In CI, gcc-11.2 on Debian was complaining that we pass a NULL pointer to `fprintf()`. I don't see how that could happen, but the code was assuming that we don't actually return NULL, so let's make that explicit.

yuwata (approved):
LGTM.

medhefgo:
Dunno how this skipped past ci, but now I get this error with clang on arch:
```
../src/shared/firewall-util-nft.c: In function ‘nft_set_element_op_in_addr_open’:
../src/shared/firewall-util-nft.c:931:65: error: passing argument 4 of ‘in_addr_prefix_to_string’ from incompatible pointer type [-Werror=incompatible-pointer-types]
  931 |         (void) in_addr_prefix_to_string(af, address, prefixlen, &addr_str);
      |                                                                 ^~~~~~~~~
      |                                                                 |
      |                                                                 char **
In file included from ../src/shared/firewall-util.h:7,
                 from ../src/shared/firewall-util-nft.c:18:
../src/basic/in-addr-util.h:100:23: note: expected ‘char *’ but argument is of type ‘char **’
  100 |                 char *buf,
      |                 ~~~~~~^~~
../src/shared/firewall-util-nft.c:931:16: error: too few arguments to function ‘in_addr_prefix_to_string’
  931 |         (void) in_addr_prefix_to_string(af, address, prefixlen, &addr_str);
      |                ^~~~~~~~~~~~~~~~~~~~~~~~
../src/basic/in-addr-util.h:96:5: note: declared here
   96 | int in_addr_prefix_to_string(
```

keszybz:
> Dunno how this skipped past ci, but now I get this error with ~clang~ gcc on arch:

Two conflicting pull requests were merged. I pushed a resolution to `main` right now, because it's very simple and I don't think it was worth waiting of the CI. (Or in other words, if I got this wrong, the CI on any pull request will tell us as much.)

PR #23620, opened 2022-06-03
hwdb: Add accel orientation quirk for the Aya Neo Next

Maccraft123:
Has been tested only on their Aya Neo Next Pro model,
should work on Aya Neo Next and Aya Neo Next Advance,
however I've been unable to find someone to test it, and
these devices differ only in minor hardware changes so
it should be fine to treat them like one.


yuwata (approved):

Maccraft123:
After this force-push it should also match on different firmware version,
since it has been found out that some units have `pnAYANEONEXT` instead of `pnNEXT`.
It should also match on Next Advance and Next now.

Issue #23619, opened 2022-06-03
systemd built with clang and --optimization>=2 is unusable


evverx:
**systemd version the issue has been seen with**
 > d48cfc98ebe3ceb

```sh
$ clang --version
clang version 14.0.0 (Fedora 14.0.0-1.fc36)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

$ CC=clang CXX=clang++ meson --optimization=2 build
$ ninja -C ./build -v
$ ninja -C ./build test
```

Since https://github.com/systemd/systemd/commit/0bd292567a543d124cd303f7dd61169a209cae64 was merged `systemd` has been using `__builtin_dynamic_object_size` to prevent issues like https://github.com/systemd/systemd/issues/22801. It appears to work with `gcc` (where `__builtin_dynamic_object_size` returns `-1` most of the time). `clang` on the other hand can keep track of pointers passed to functions and returned from them (in most cases) so the code allocating memory like `GREEDY_REALLOC` uses actual sizes while some other code receiving pointers to that memory through a chain of function calls can get either the actual size or `-1`. This leads to segfaults like
```sh
#0  journal_importer_cleanup (imp=0x6572756365730a36) at ../src/shared/journal-importer.c:28
#1  0x0000000000404112 in source_free (source=0x6572756365730a36) at ../src/journal-remote/journal-remote-parse.c:14
#2  0x0000000000403725 in remove_source (s=s@entry=0x7fffffffe410, fd=fd@entry=8) at ../src/journal-remote/journal-remote.c:210
#3  0x0000000000403e3a in journal_remote_server_destroy (s=0x7fffffffe410) at ../src/journal-remote/journal-remote.c:359
#4  0x0000000000402b47 in LLVMFuzzerTestOneInput (data=<optimized out>, size=<optimized out>) at ../src/journal-remote/fuzz-journal-remote.c:91
#5  0x0000000000402d6f in main (argc=2, argv=0x7fffffffe678) at ../src/fuzz/fuzz-main.c:50
```
and test failures like
```sh
# ./build/test-alloc-util
/* test_alloca */
/* test_GREEDY_REALLOC */
/* test_memdup_multiply_and_greedy_realloc */
usable size: 24, dynamic size: -1, object size: -1
usable size: 88, dynamic size: 80, object size: -1
usable size: 88, dynamic size: -1, object size: -1
[...]

Issue #23150, opened 2022-04-21
udev assertion 'i < num' in device_update_properties_bufs() since 0bd29256


tambry:
**systemd version the issue has been seen with**
Bisected to 0bd292567a543d124cd303f7dd61169a209cae64

**Used distribution**
Debian Unstable

**Linux kernel version used** (`uname -a`)
`Linux v11.cleveronmobility.dev 5.16.0-6-amd64 #1 SMP PREEMPT Debian 5.16.18-1 (2022-03-29) x86_64 GNU/Linux`

**CPU architecture issue was seen on**
amd64

**Expected behaviour you didn't see**
Sucessful startup

**Unexpected behaviour you saw**
An assertion. As a result drivers necessary for the internal disk aren't loaded during initramfs and the system fails to boot.

**Steps to reproduce the problem**
Boot latest systemd master on PC Engines APU2

**Additional program output to the terminal or log subsystem illustrating the issue**
[Full boot log with `udev.log-priority=debug`](https://github.com/systemd/systemd/files/8529862/udev.boot.log.txt)

**Additional details**
Compiled using LLVM aed0e8b8056a with `-march=ivybridge`.

I built systemd with Debian tools and mostly downstream Debian patches. I'm hesitant to blame those at first as the offending commit seems innocuous.  
Will try building with the latest LLVM head soon. Happy to provide any further information or help with debugging.

yuwata:
Maybe, compiler bug? Is it reproducible with a stable version of LLVM?

tambry:
Reproduced with newer LLVM 7a98d8351b27 (yesterday's HEAD, no notable related changes committed since).

Not affected by disabling/enabling fortify and stackprotector. `-march=ivybridge` also isn't the culprit. Will try with an older LLVM release within a few days.

thesamesam:
We've had people hit this in [Gentoo](https://bugs.gentoo.org/841770) with LLVM 14.0.3 and systemd-251-rc2.

thesamesam:
CC @marxin @siddhesh in case you're interested, as it's related to #22801. But know you're both GCC people.

evverx:
I'm not sure why GitHub didn't close it automatically. It was fixed by reverting that patch in https://github.com/systemd/systemd/pull/23621. Closing.

Issue #22801, opened 2022-03-21
*** buffer overflow detected *** with GCC 12 and -D_FORTIFY_SOURCE=3


marxin:
Using the new `-D_FORTIFY_SOURCE` level, I see the following buffer overflow:

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff7aad1e3 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007ffff7a5d306 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff7a46813 in __GI_abort () at abort.c:79
#4  0x00007ffff7aa01b7 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff7be63cf "*** %s ***: terminated\n") at ../sysdeps/posix/libc_fatal.c:155
#5  0x00007ffff7b4530a in __GI___fortify_fail (msg=msg@entry=0x7ffff7be6375 "buffer overflow detected") at fortify_fail.c:26
#6  0x00007ffff7b438b6 in __GI___chk_fail () at chk_fail.c:28
#7  0x00007ffff7b43d3a in __read_chk (fd=<optimized out>, buf=<optimized out>, nbytes=<optimized out>, buflen=<optimized out>) at read_chk.c:30
#8  0x00007ffff78554a4 in read_virtual_file.constprop () at lib/libudev.so.1
#9  0x00007ffff784c9bf in device_read_uevent_file () at lib/libudev.so.1
#10 0x00007ffff784d026 in sd_device_get_devnum () at lib/libudev.so.1
#11 0x00007ffff785b7a8 in device_read_db_internal.constprop () at lib/libudev.so.1
```

It happens here:
https://github.com/systemd/systemd/blob/e7949be7900af8e045cb530a22d023037eac273c/src/basic/fileio.c#L469-L482
where `MALLOC_SIZEOF_SAFE` is defined as:
https://github.com/systemd/systemd/blob/02036ca7e98f282b262274bea246b453e9e2b8d1/src/basic/alloc-util.h#L173-L183

Even though, the comment counts with `-D_FORTIFY_SOURCE` it behaves wrongly as `SIZE_MAX` is returned for an unknown object. Thus we end up with `size == malloc_usable_size` and the assert happens.

[...]

Issue #19203, opened 2021-04-03
cryptenroll --recovery-key crashes


grawity:
**systemd version the issue has been seen with**
 > v248 (Arch Linux build)

**Used distribution**
 > Arch

**Linux kernel version used** (`uname -a`)
 > 5.11.8-arch1-1

**CPU architecture issue was seen on**
 > amd64

**Expected behaviour you didn't see**
 > It doesn't crash.

**Unexpected behaviour you saw**
 > It crashes.

**Steps to reproduce the problem**
 > `systemd-cryptenroll --recovery-key`

**Additional program output to the terminal or log subsystem illustrating the issue**

```text
# SYSTEMD_LOG_LEVEL=debug systemd-cryptenroll --recovery-key /dev/nvme0n1p2
Trying to load LUKS2 crypt type from device /dev/nvme0n1p2.
Crypto backend (OpenSSL 1.1.1k  25 Mar 2021) initialized in cryptsetup library version 2.3.5.
Detected kernel Linux 5.11.8-arch1-1 x86_64.
Loading LUKS2 header (repair disabled).
Acquiring read lock for device /dev/nvme0n1p2.
Opening lock resource file /run/cryptsetup/L_259:2
Verifying lock handle for /dev/nvme0n1p2.
Device /dev/nvme0n1p2 READ lock taken.
Trying to read primary LUKS2 header at offset 0x0.
Opening locked device /dev/nvme0n1p2
Veryfing locked device handle (bdev)
LUKS2 header version 2 of size 16384 bytes, checksum sha256.
Checksum:a8b702198d6c275f8a05700c82d96b262d6a04ad21783ad61cf13ac613705737 (on-disk)
Checksum:a8b702198d6c275f8a05700c82d96b262d6a04ad21783ad61cf13ac613705737 (in-memory)
Trying to read secondary LUKS2 header at offset 0x4000.
Reusing open ro fd on device /dev/nvme0n1p2
LUKS2 header version 2 of size 16384 bytes, checksum sha256.
Checksum:4cf340067f1f510071ddf0db08e5c03466d95f230fa412576e52a1612b2a58d2 (on-disk)
Checksum:4cf340067f1f510071ddf0db08e5c03466d95f230fa412576e52a1612b2a58d2 (in-memory)
Device size 549755813888, offset 2097152.
Device /dev/nvme0n1p2 READ lock released.
[...]

PR #23621, opened 2022-06-03
ci: build systemd with clang with -Dmode=release --optimization=2

evverx:
This is what's most likely used to build systemd with clang in
practice so let's test it as well.

In preparation for reverting https://github.com/systemd/systemd/commit/0bd292567a543d124cd303f7dd61169a209cae64
(which replaced a bogus buffer overflow found with _FORTIFY_SOURCE=3
with actual segfaults).

0bd292567a543d124c is reverted as well.
It isn't guaranteed anywhere that __builtin_dynamic_object_size can
always deduce the size of every object passed to it so systemd
can end up using either malloc_usable_size or
__builtin_dynamic_object_size when pointers are passed around,
which in turn can lead to actual segfaults like the one mentioned in
https://github.com/systemd/systemd/issues/23619.

Apparently __builtin_object_size can return different results for
pointers referring to the same memory as well but somehow it hasn't
caused any issues yet. Looks like this whole
malloc_usable_size/FORTIFY_SOURCE stuff should be revisited.

Closes https://github.com/systemd/systemd/issues/23619 and https://github.com/systemd/systemd/issues/23150.
Reopens https://github.com/systemd/systemd/issues/22801

### DIFF
--- a/src/basic/alloc-util.h
+++ b/src/basic/alloc-util.h
@@ -174,23 +174,13 @@ void* greedy_realloc0(void **p, size_t need, size_t size);
  * is compatible with _FORTIFY_SOURCES. If _FORTIFY_SOURCES is used many memory operations will take the
  * object size as returned by __builtin_object_size() into account. Hence, let's return the smaller size of
  * malloc_usable_size() and __builtin_object_size() here, so that we definitely operate in safe territory by
- * both the compiler's and libc's standards. Note that _FORTIFY_SOURCES=3 handles also dynamically allocated
- * objects and thus it's safer using __builtin_dynamic_object_size if _FORTIFY_SOURCES=3 is used (#22801).
- * Moreover, when NULL is passed malloc_usable_size() is documented to return zero, and
+ * both the compiler's and libc's standards. Note that __builtin_object_size() evaluates to SIZE_MAX if the
+ * size cannot be determined, hence the MIN() expression should be safe with dynamically sized memory,
+ * too. Moreover, when NULL is passed malloc_usable_size() is documented to return zero, and
  * __builtin_object_size() returns SIZE_MAX too, hence we also return a sensible value of 0 in this corner
  * case. */
-
-#if defined __has_builtin
-#  if __has_builtin(__builtin_dynamic_object_size)
-#    define MALLOC_SIZEOF_SAFE(x) \
-        MIN(malloc_usable_size(x), __builtin_dynamic_object_size(x, 0))
-#  endif
-#endif
-
-#ifndef MALLOC_SIZEOF_SAFE
 #define MALLOC_SIZEOF_SAFE(x) \
         MIN(malloc_usable_size(x), __builtin_object_size(x, 0))
-#endif
 
 /* Inspired by ELEMENTSOF() but operates on malloc()'ed memory areas: typesafely returns the number of items
  * that fit into the specified memory block */


thesamesam:
Good catch with CI. I was wondering why it didn't hit it.

evverx:
The CI mostly uses clang to test systemd with ASan/UBsan and somewhat MSan (where `malloc_usable_size` is intercepted correctly with all optimization levels). It just wasn't prepared for plain builds with -O2 where `MALLOC_SIZEOF_SAFE/MALLOC_ELEMENTSOF` can return either `__builtin_dynamic_object_size` or `malloc_usable_size` (which is usually larger than `__builtin_dynamic_object_size`) for pointers referring to the same memory in different places.

evverx:
Judging by
```
autopkgtest: WARNING: Test dependencies are unsatisfiable - calling apt install on test deps directly for further data about failing dependencies in test logs
blame: https://salsa.debian.org/systemd-team/systemd.git#upstream-ci
badpkg: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
autopkgtest [09:56:55]: ERROR: erroneous package: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
```
Ubuntu CI and Sempahore CI broke down again.

mbiebl:
> Judging by
> 
> ```
> autopkgtest: WARNING: Test dependencies are unsatisfiable - calling apt install on test deps directly for further data about failing dependencies in test logs
> blame: https://salsa.debian.org/systemd-team/systemd.git#upstream-ci
> badpkg: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
> autopkgtest [09:56:55]: ERROR: erroneous package: Test dependencies are unsatisfiable. A common reason is that your testbed is out of date with respect to the archive, and you need to use a current testbed or run apt-get update or use -U.
> ```
> 
> Ubuntu CI and Sempahore CI broke down again.

I rebased the upstream-ci branch on current debian/master which includes
https://salsa.debian.org/systemd-team/systemd/-/commit/a8a310c5e6f6e466f638a246ac8840615e5019f0

This package is provided via bullseye-backports, but it seems this repository is not enabled in Semaphore CI.
For Ubuntu Focal, the package should be available via https://launchpad.net/~upstream-systemd-ci/+archive/ubuntu/systemd-ci

Not quite sure why that didn't work.

@bluca any idea?

Anyway, I reverted that commit for now to unbreak CI.

Please retrigger.

mbiebl:
ah, seems package-notes was only built for jammy and not for focal

mbiebl:
So, Semaphore CI failed with a different error (which I guess is a good sign).

As for Ubuntu CI, might be that they are caching the git repo as it seems the still have the old state.

evverx:
> So, Semaphore CI failed with a different error (which I guess is a good sign).

I opened https://github.com/systemd/systemd/issues/23630

evverx:
FWIW I'm not sure if https://salsa.debian.org/systemd-team/systemd/-/commit/453537cfbead2f0780b51c5b24d598496575770f was released or not but that warning boils down to `journal-remote` crashing due to `assert_not_reached()` in `check_permissions` without `gnutls`.

mbiebl:
Argh, sorry. Wasn't really my intention to disrupt CI that much by updating upstream-ci.
Should I revert https://salsa.debian.org/systemd-team/systemd/-/commit/453537cfbead2f0780b51c5b24d598496575770f again?
Or should 507fc32e1b1a8efa936218571790c7a2e8f11ef0 be merged into main?

Then again, upstream-ci, afaics, is also used for systemd-stable, so this commit would have to be pulled there as well.

mbiebl:
> FWIW I'm not sure if https://salsa.debian.org/systemd-team/systemd/-/commit/453537cfbead2f0780b51c5b24d598496575770f was released or not

This change was released in 251.1-1 to unstable.
See https://tracker.debian.org/news/1327809/accepted-systemd-2511-1-source-into-unstable/

evverx:
> Should I revert https://salsa.debian.org/systemd-team/systemd/-/commit/453537cfbead2f0780b51c5b24d598496575770f again?
Or should https://github.com/systemd/systemd/commit/507fc32e1b1a8efa936218571790c7a2e8f11ef0 be merged into main?

I think if it's expected that `journald-remote` should crash there https://github.com/systemd/systemd/commit/507fc32e1b1a8efa936218571790c7a2e8f11ef0 should probably be merged. I haven't looked at that part of the code closely though so I can't say for sure.

evverx:
Looks like `__builtin_object_size` is equally dangerous when pointers are passed around. I'm not sure why it has never manifested itself in actual buffer overruns (but now I know where to look for them closer :-)) but I think it would make sense to revert 6df28e1f847d68ad37ff as well. I don't think the part of systemd allocating memory should rely on compiler features that can return different results in different places depending on how pointers are passed around.
```sh
# cat test.c
#include <malloc.h>

void print_sizes(void *p) {
	printf("usable_size: %ld, dynamic_size: %ld, object_size: %ld\n", malloc_usable_size(p), __builtin_dynamic_object_size(p, 0), __builtin_object_size(p, 0));
}

int main(int argc, char *argv[]) {
	char *p = malloc(10);
	printf("usable_size: %ld, dynamic_size: %ld, object_size: %ld\n", malloc_usable_size(p), __builtin_dynamic_object_size(p, 0), __builtin_object_size(p, 0));
	print_sizes(p);
	free(p);
	return 0;
}
```
```sh
# gcc --version
gcc (GCC) 12.1.1 20220507 (Red Hat 12.1.1-1)
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

# gcc -O2 test.c
# ./a.out
usable_size: 24, dynamic_size: 10, object_size: 10
usable_size: 24, dynamic_size: -1, object_size: -1
```

keszybz:
/cc @marxin 

I guess we should merge this to fix those crashes. `clang` and `-O2` is super standard and if the code doesn't work with that, we don't have much choice. Maybe we can figure out a way to support `-D_FORTIFY_SOURCE=3` later on.

> but I think it would make sense to revert [6df28e1](https://github.com/systemd/systemd/commit/6df28e1f847d68ad37ffe3f4ff47745b55233861) as well. I don't think the part of systemd allocating memory should rely on compiler features that can return different results in different places depending on how pointers are passed around.

Hmm, maybe. But let's do this one step at a time.

evverx:
> But let's do this one step at a time

Agreed. Since it hasn't caused any issues yet I think it should be fine to keep it for now. I'm a bit concerned though that those codepaths are completely hidden from ASan, UBSan, MSan and Valgrind and whatever is tested by the CI isn't what is actually shipped by distributions.

> Maybe we can figure out a way to support -D_FORTIFY_SOURCE=3 later on.

As far as I understand at this point `-D_FORTIFY_SOURCE` isn't compatible with `malloc_usable_size` and it can be fixed either by getting rid of `malloc_usable_size` or by intercepting `malloc_usable_size` correctly with `_FORTIFY_SOURCE` somehow by analogy with ASan, UBSan, MSan and Valgrind.

marxin:
> Looks like `__builtin_object_size` is equally dangerous when pointers are passed around. I'm not sure why it has never manifested itself in actual buffer overruns (but now I know where to look for them closer :-))

Can you please explain why is dangerous? Note that `-1` is of `size_t` so it's a huge number and it will also satisfy checks like:

```c
if (size < buffer_size)
   abort();
```

Can you please explain how exactly does it crash with `clang` and would it be possible to isolate a reproducer?

marxin:
> As far as I understand at this point `-D_FORTIFY_SOURCE` isn't compatible with `malloc_usable_size` and it can be fixed either by getting rid of `malloc_usable_size` or by intercepting `malloc_usable_size` correctly with `_FORTIFY_SOURCE` somehow by analogy with ASan, UBSan, MSan and Valgrind.

Yes, it does not work correctly with `malloc_usable_size` and I would consider not using it by such a system tool like `systemd`. It seems to me like a super-micro optimization that does not worth it.

evverx:
> I would consider not using it by such a system tool like systemd

@marxin agreed. I think ideally it would be great if _FORTIFY_SOURCE was compatible with `malloc_usable_size` somehow (to avoid false positives in general) but even if it worked by analogy with say Valgrind right now if would basically mean that d4b604baeadbb24 would never work as intended with systemd packages built with FORTIFY_SOURCE (and as far as I know basically all distributions include _FORTIFY_SOURCE=2 in their CFLAGS one way or another).

marxin:
> @marxin agreed. I think ideally it would be great if _FORTIFY_SOURCE was compatible with `malloc_usable_size` somehow (to avoid false positives in general) but even if it worked by analogy with say Valgrind right now if would basically mean that [d4b604b](https://github.com/systemd/systemd/commit/d4b604baeadbb2498e4f2c3e260260eed210f5d6) would never work as intended with systemd packages built with FORTIFY_SOURCE (and as far as I know basically all distributions include _FORTIFY_SOURCE=2 in their CFLAGS one way or another).

Good! And yes, most of the distros likely use `_FORTIFY_SOURCE=2`. So are planning a commit that will remove the usage of `malloc_usable_size` or should I do that?

evverx:
@marxin I'd wait for @poettering to chime in. Looking at https://github.com/systemd/systemd/issues/22801#issuecomment-1073962482 and https://github.com/systemd/systemd/issues/19203#issuecomment-814194892 I'm not sure the idea of removing `malloc_usable_size` would be approved right away unanimously.

PR #23624, opened 2022-06-04
execute: fix resource leak

sshedi:
CID#1431998


sshedi:
@yuwata This as well, can you please take a look?

keszybz:
LGTM.

PR #23628, opened 2022-06-04
fix: sd_id128_get_invocation now works for user services if keyring is unavailable

i-do-cpp:
`sd_id128_get_invocation` only returned success if `get_invocation_from_keyring` succeeded. This is a problem when no keyring is available.
This patch makes `sd_id128_get_invocation` first try `get_invocation_from_environment` (and return on success) and then try `get_invocation_from_keyring`.  
This has the effect that `sd_id128_get_invocation` now correctly returns the `invocation id` for user services if no keyring is available.

### DIFF
--- a/src/libsystemd/sd-id128/sd-id128.c
+++ b/src/libsystemd/sd-id128/sd-id128.c
@@ -256,7 +256,10 @@ _public_ int sd_id128_get_invocation(sd_id128_t *ret) {
                 /* We first check the environment. The environment variable is primarily relevant for user
                  * services, and sufficiently safe as long as no privilege boundary is involved. */
                 r = get_invocation_from_environment(&saved_invocation_id);
-                if (r < 0 && r != -ENXIO)
+                if (r >= 0) {
+                        *ret = saved_invocation_id;
+                        return 0;
+                } else if (r != -ENXIO)
                         return r;
 
                 /* The kernel keyring is relevant for system services (as for user services we don't store


src/libsystemd/sd-id128/sd-id128.c
```diff
@@ -253,17 +253,21 @@ _public_ int sd_id128_get_invocation(sd_id128_t *ret) {
         assert_return(ret, -EINVAL);
 
         if (sd_id128_is_null(saved_invocation_id)) {
-                /* We first check the environment. The environment variable is primarily relevant for user
-                 * services, and sufficiently safe as long as no privilege boundary is involved. */
-                r = get_invocation_from_environment(&saved_invocation_id);
-                if (r < 0 && r != -ENXIO)
-                        return r;
```
yuwata: Please do not change the order of the two `get_invocation_from_xxx()`.
I guess the issue can be fixed by fixing the return value handling:
```
r = get_invocation_from_environment();
if (r != -ENXIO)
        return r;
```

yuwata (changes requested):

src/libsystemd/sd-id128/sd-id128.c
i-do-cpp: Thank you for reviewing this PR!
The suggested change would still fail without a keyring as the following `get_invocation_from_keyring` call overrides the `get_invocation_from_environment` call.

We now return the result of `get_invocation_from_environment` if the call succeeded, please have another look.

src/libsystemd/sd-id128/sd-id128.c
```diff
@@ -258,6 +258,11 @@ _public_ int sd_id128_get_invocation(sd_id128_t *ret) {
                 r = get_invocation_from_environment(&saved_invocation_id);
                 if (r < 0 && r != -ENXIO)
                         return r;
+                else if (!r)
+                {
+                        *ret = saved_invocation_id;
+                        return 0;
+                }
```
yuwata: We do not use implicit cast from integer to boolean. How about the following?
```
if (r >= 0) {
        *ret = saved_invocation_id;
        return 0;
} else if (r != -ENXIO)
        return r;
```

yuwata (changes requested):

src/libsystemd/sd-id128/sd-id128.c
i-do-cpp: sounds good, the diff is now exactly what you suggested

yuwata (approved):
LGTM. Thank you.

PR #23635, opened 2022-06-05
journal: Add "format" output mode

daandemeyer:
Draft since I still need to add specifiers for timestamps and documentation, etc. I'm slightly unsure on the best way to do specifiers for timestamps though, since there's many different ways to output the time. Should we reuse a bunch of the remaining specifier variables to each represent a different time output or do we need something more clever.

See commit messages for more details on the feature.

### DIFF
--- a/src/journal/journalctl.c
+++ b/src/journal/journalctl.c
@@ -128,6 +128,7 @@ static uint64_t arg_vacuum_size = 0;
 static uint64_t arg_vacuum_n_files = 0;
 static usec_t arg_vacuum_time = 0;
 static char **arg_output_fields = NULL;
+static char *arg_format = NULL;
 static const char *arg_pattern = NULL;
 static pcre2_code *arg_compiled_pattern = NULL;
 static PatternCompileCase arg_case = PATTERN_COMPILE_CASE_AUTO;
@@ -441,6 +442,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_NO_HOSTNAME,
                 ARG_OUTPUT_FIELDS,
                 ARG_NAMESPACE,
+                ARG_FORMAT,
         };
 
         static const struct option options[] = {
@@ -508,6 +510,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "no-hostname",          no_argument,       NULL, ARG_NO_HOSTNAME          },
                 { "output-fields",        required_argument, NULL, ARG_OUTPUT_FIELDS        },
                 { "namespace",            required_argument, NULL, ARG_NAMESPACE            },
+                { "output-format",        required_argument, NULL, ARG_FORMAT               },
                 {}
         };
 
@@ -1034,6 +1037,15 @@ static int parse_argv(int argc, char *argv[]) {
                         break;
                 }
 
+                case ARG_FORMAT:
+                        arg_output = OUTPUT_FORMAT;
+
+                        r = free_and_strdup(&arg_format, optarg);
+                        if (r < 0)
+                                return log_oom();
+
+                        break;
+
                 case '?':
                         return -EINVAL;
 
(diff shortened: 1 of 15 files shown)


src/libsystemd/sd-journal/sd-journal.c
```diff
@@ -2236,26 +2236,24 @@ _public_ int sd_journal_get_monotonic_usec(sd_journal *j, uint64_t *ret, sd_id12
...
 
-static bool field_is_valid(const char *field) {
-        const char *p;
-
+static bool field_is_valid(const char *field, size_t field_length) {
         assert(field);
 
-        if (isempty(field))
+        if (field_length == 0)
                 return false;
```
poettering, line 2237: maybe add `if (field_length == SIZE_MAX) field_lenght = strlen(field)` here? And then just call the functions with SIZE_MAX as second parameter?

poettering:
If this is done it should probably be aligned closely to how we do things in the catalog entry formatting code. It's the very same problem after all, and we should share a single templating logic for that.

src/journal/journalctl.c
```diff
@@ -1103,6 +1115,11 @@ static int parse_argv(int argc, char *argv[]) {
                 arg_system_units = strv_free(arg_system_units);
         }
 
+        if (arg_format && arg_output != OUTPUT_FORMAT)
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                       "--format cannot be used with output mode \"%s\"",
```
poettering: → `--format=` 

(i.e. suffix `=` please)

src/journal/journalctl.c
```diff
@@ -525,6 +527,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "no-hostname",          no_argument,       NULL, ARG_NO_HOSTNAME          },
                 { "output-fields",        required_argument, NULL, ARG_OUTPUT_FIELDS        },
                 { "namespace",            required_argument, NULL, ARG_NAMESPACE            },
+                { "format",               required_argument, NULL, ARG_FORMAT               },
```
poettering: name this "output-format" here too? should have same name in all tools

src/journal/journalctl.c
poettering: i think it would make sense to default to OUTPUT_FORMAT if --output-format= is used, but --output= is not.

src/shared/logs-show.c
```diff
@@ -1228,14 +1258,234 @@ static int output_cat(
...
+        { 'T', format_specifier_field, "_TRANSPORT" },
+        { 'c', format_specifier_field, "_SYSTEMD_CGROUP" },
+        { 's', format_specifier_field, "_SYSTEMD_SLICE" },
+
+        { 'm', format_specifier_field, "MESSAGE"},
+        { 'i', format_specifier_field, "SYSLOG_IDENTIFIER" },
+        { 'f', format_specifier_field, "CODE_FILE" },
+        { 'F', format_specifier_field, "CODE_FUNC" },
+        { 'l', format_specifier_field, "CODE_LINE" },
+        { 'X', format_specifier_field, "ERRNO" },
```
poettering: I'd really stick to the field names for this, and not bother with short specifiers. that adds another level of indirection for very little reason.

src/journal/journalctl.c
poettering: we should also fail here if --output=format is used, but --output-format= wasn't specified

poettering:
looks like something that would make a lot of people happy. but as mentioned should relaly reuse the templating stuff of the catalog entry display.

src/shared/logs-show.c
daandemeyer: We want to present an attractive API for power users though no? I really feel that we should have a few specifiers to reduce the amount of typing required to get useful information when using this option.

poettering: but i think it would be easier for everyone to just use the long field names we also use for filtering. Sounds weird using one type of specifier for filtering and another for display

daandemeyer: But let's say you want to include the time in the format, do you really want users to type out SOURCE_MONOTONIC_TIMESTAMP or stuff like that?

In general, the timestamps need special processing and can't be specified by a single field, especially if you want to mimick the date formatting that's done by default. Ideally that'd be doable with a single specifier

poettering: So yeah, that might make sense, use % specifiers for formatting key fields in non-literal ways. But let's keep a lid on it, i.e. only where it really makes sense. i.e. displaying raw UNIX timestamp values probably makes little sense, hence having a % specifier to format it in human readable way makes sense to me. But let's not add specifiers for stuff that would be 1:1 the original field raw data...

DaanDeMeyer:
> looks like something that would make a lot of people happy. but as mentioned should relaly reuse the templating stuff of the catalog entry display.

I'm not sure I follow here. What exactly is the templating stuff of the catalog entry display?

DaanDeMeyer:
CI is supposed to fail, I still need to implement the timestamp formatting functions

DaanDeMeyer:
Closing until I have time to work on this again

PR #23639, opened 2022-06-06
scope: allow unprivileged delegation on scopes

### LINKED ISSUE #21683: Delegation to unprivileged users doesn't work for scopes
 > systemd-249

**Used distribution**
 > Fedora

 > `Linux localhost 5.16.0-0.rc3.24.fc36.x86_64 #1 SMP PREEMPT Mon Nov 29 13:43:51 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

 > x86_64

> I would expect that bash process started by systemd-run is placed in a cgroup that can be managed by test user.

> $subject, i.e. cgroup is owned by root user.

```bash
$ useradd test
$ systemd-run --uid=test -p Delegate=yes --scope --unit foo.scope --slice bar.slice /bin/bash
$ ls -ld /sys/fs/cgroup/bar.slice/foo.scope/
drwxr-xr-x. 2 root root 0 Dec  8 08:07 /sys/fs/cgroup/bar.slice/foo.scope/
```

msekletar:
Previously it was possible to set delegate property for scope, but you
were not able to allow unprivileged process to manage the scope's cgroup
hierarchy. This is useful when launching manager process that  will run
unprivileged but is supposed to manage its own (scope) sub-hierarchy.

Fixes #21683

### DIFF
--- a/src/basic/unit-def.c
+++ b/src/basic/unit-def.c
@@ -169,6 +169,7 @@ DEFINE_STRING_TABLE_LOOKUP(path_state, PathState);
 
 static const char* const scope_state_table[_SCOPE_STATE_MAX] = {
         [SCOPE_DEAD]         = "dead",
+        [SCOPE_START_CHOWN]  = "start-chown",
         [SCOPE_RUNNING]      = "running",
         [SCOPE_ABANDONED]    = "abandoned",
         [SCOPE_STOP_SIGTERM] = "stop-sigterm",
--- a/src/basic/unit-def.h
+++ b/src/basic/unit-def.h
@@ -114,6 +114,7 @@ typedef enum PathState {
 
 typedef enum ScopeState {
         SCOPE_DEAD,
+        SCOPE_START_CHOWN,
         SCOPE_RUNNING,
         SCOPE_ABANDONED,
         SCOPE_STOP_SIGTERM,
--- a/src/core/dbus-scope.c
+++ b/src/core/dbus-scope.c
@@ -186,6 +186,12 @@ int bus_scope_set_property(
                 r = bus_kill_context_set_transient_property(u, &s->kill_context, name, message, flags, error);
                 if (r != 0)
                         return r;
+
+                if (streq(name, "User"))
+                        return bus_set_transient_user_relaxed(u, name, &s->user, message, flags, error);
+
+                if (streq(name, "Group"))
+                        return bus_set_transient_user_relaxed(u, name, &s->group, message, flags, error);
         }
 
         return 0;
--- a/src/core/scope.c
+++ b/src/core/scope.c
@@ -6,6 +6,7 @@
 #include "alloc-util.h"
 #include "dbus-scope.h"
 #include "dbus-unit.h"
+#include "exit-status.h"
 #include "load-dropin.h"
 #include "log.h"
 #include "process-util.h"
@@ -18,9 +19,11 @@
 #include "strv.h"
 #include "unit-name.h"
 #include "unit.h"
+#include "user-util.h"
 
 static const UnitActiveState state_translation_table[_SCOPE_STATE_MAX] = {
         [SCOPE_DEAD] = UNIT_INACTIVE,
+        [SCOPE_START_CHOWN] = UNIT_ACTIVATING,
         [SCOPE_RUNNING] = UNIT_ACTIVE,
         [SCOPE_ABANDONED] = UNIT_ACTIVE,
         [SCOPE_STOP_SIGTERM] = UNIT_DEACTIVATING,
@@ -39,6 +42,7 @@ static void scope_init(Unit *u) {
         s->runtime_max_usec = USEC_INFINITY;
         s->timeout_stop_usec = u->manager->default_timeout_stop_usec;
         u->ignore_on_isolate = true;
+        s->user = s->group = NULL;
 }
 
 static void scope_done(Unit *u) {
@@ -50,6 +54,9 @@ static void scope_done(Unit *u) {
         s->controller_track = sd_bus_track_unref(s->controller_track);
 
         s->timer_event_source = sd_event_source_disable_unref(s->timer_event_source);
+
+        s->user = mfree(s->user);
+        s->group = mfree(s->group);
 }
 
 static usec_t scope_running_timeout(Scope *s) {
@@ -107,7 +114,7 @@ static void scope_set_state(Scope *s, ScopeState state) {
         old_state = s->state;
         s->state = state;
 
-        if (!IN_SET(state, SCOPE_STOP_SIGTERM, SCOPE_STOP_SIGKILL))
+        if (!IN_SET(state, SCOPE_STOP_SIGTERM, SCOPE_STOP_SIGKILL, SCOPE_START_CHOWN))
                 s->timer_event_source = sd_event_source_disable_unref(s->timer_event_source);
 
         if (IN_SET(state, SCOPE_DEAD, SCOPE_FAILED)) {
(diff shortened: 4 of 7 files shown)


src/core/scope.c
```diff
@@ -353,35 +359,87 @@ static void scope_enter_signal(Scope *s, ScopeState state, ScopeResult f) {
...
+                        const char *group = s->group;
+
+                        r = get_group_creds(&group, &gid, 0);
+                        if (r < 0) {
+                                log_unit_error_errno(UNIT(s), r, "Failed to resolve group \"%s\": %m", group);
+                                _exit(EXIT_GROUP);
+                        }
+                }
+
+                if (uid_is_valid(uid) && unit_cgroup_delegate(u)) {
```
Werkov: This function is only called with `s->user` and `get_user_creds` should err if that was invalid input, so this `uid_is_valid` check here seems superfluous.
Also, can you factor the `unit_cgroup_delegate()` check out so that the fork is saved when unnecessary?

src/basic/unit-def.h
```diff
@@ -114,6 +114,7 @@ typedef enum PathState {
 
 typedef enum ScopeState {
         SCOPE_DEAD,
+        SCOPE_CHOWN_CGROUP,
```
Werkov: This should be added to `state_translation_table` as `UNIT_ACTIVATING`.

src/core/scope.c
```diff
@@ -353,35 +359,87 @@ static void scope_enter_signal(Scope *s, ScopeState state, ScopeResult f) {
...
-        if (unit_has_name(u, SPECIAL_INIT_SCOPE))
-                return -EPERM;
+        (void) unit_realize_cgroup(u);
+        (void) unit_reset_accounting(u);
 
-        if (s->state == SCOPE_FAILED)
-                return -EPERM;
+        r = scope_arm_timer(s, usec_add(now(CLOCK_MONOTONIC), 30*USEC_PER_SEC));
+        if (r < 0)
+                return r;
```
Werkov, line 373: This hard-coded timeout looks fishy.
Couldn't this be charged against `JobRunningTimeoutSec=`? (I know it's not the same, especially diverts from service start, however, it'd reuse the existing mechanisms.)

src/core/scope.c
```diff
@@ -416,6 +474,32 @@ static int scope_start(Unit *u) {
...
+        /* We can't fulfill this right now, please try again later */
+        if (IN_SET(s->state, SCOPE_STOP_SIGTERM, SCOPE_STOP_SIGKILL))
+                return -EAGAIN;
+
+        assert(s->state == SCOPE_DEAD);
+
+        if (!u->transient && !MANAGER_IS_RELOADING(u->manager))
+                return -ENOENT;
+
+        if (s->user)
```
Werkov: This is simple(r). Is the omission of `s->group` intentional?

src/core/scope.c
```diff
@@ -353,35 +359,87 @@ static void scope_enter_signal(Scope *s, ScopeState state, ScopeResult f) {
...
         r = unit_acquire_invocation_id(u);
         if (r < 0)
                 return r;
 
-        (void) unit_realize_cgroup(u);
-        (void) unit_reset_accounting(u);
+        if (!u->cgroup_realized) {
+                (void) unit_realize_cgroup(u);
+                (void) unit_reset_accounting(u);
+        }
```
Werkov: Why do you move `unit_realize_cgroup()` into `scope_enter_running`? (It'd be simpler to keep it in `scope_start()`, no?)
Is the process attach vs chown order important?

Werkov (changes requested):
It looks useful. I'm slightly scared by the added complexity with the helper process, however, it's necessary and you seem to follow the precedent of (e.g.) socket units.
No remarks to the concept, just some inline observations, comments.

src/core/scope.c
msekletar: Actually no, this is on purpose. This works the same way for services, i.e. having 'Delegate=yes Group=foo' w/o specifing User= has no effect.

```shell
$ useradd test

# w/o User= the delegation is not enabled
$ systemd-run --unit test.service --wait -t -p Delegate=yes -p Group=test ls -ld /sys/fs/cgroup/system.slice/test.service 2>/dev/null
drwxr-xr-x 2 root root 0 Jun  9 13:12 /sys/fs/cgroup/system.slice/test.service

# With User= the delegation is set up
$ systemd-run --unit test.service --wait -t -p Delegate=yes -p User=test -p Group=test ls -ld /sys/fs/cgroup/system.slice/test.service 2>/dev/null
drwxr-xr-x 2 test test 0 Jun  9 13:12 /sys/fs/cgroup/system.slice/test.service
```

Maybe this behavior is unexpected, but for start I wanted scopes to behave the same way as services do.

Werkov: Thanks for your example. Please add a comment about consistency with services (however unexpected it is).

(FTR, "Delegate=yes Group=foo' w/o specifing User=" still has effect on what cgroup controllers are enabled in the subtree. There's no effect on ownership indeed.)

src/core/scope.c
msekletar: I wanted to preserve the order of externally visible events the same for the common case (i.e. no unpriv. delegation). But it probably doesn't really matter.

msekletar:
Updated. @Werkov PTAL.

msekletar:
@poettering PTAL

Semaphore CI failure is unrelated.

src/basic/unit-def.c
```diff
@@ -169,6 +169,7 @@ DEFINE_STRING_TABLE_LOOKUP(path_state, PathState);
 
 static const char* const scope_state_table[_SCOPE_STATE_MAX] = {
         [SCOPE_DEAD]         = "dead",
+        [SCOPE_CHOWN_CGROUP] = "chown-cgroup",
```
poettering: event though this is strictly about cgroups, I'd probably call this more generically "chown". I also would say "start" in the name, to make clear this is about starting the scope. Hence maybe call this SCOPE_START_CHOWN. 

This also has the nice effect that the state is similarly called to SOCKET_START_CHOWN of socket units. There's its about chowning the socket inode, instead of the cgroup, but if we'd add support for more stuff to chown, then it would also subsumed in SOCKET_START_CHOWN there, so having a generic name is good.

but yeah, this is pure bikeshedding

src/core/dbus-scope.c
```diff
@@ -186,6 +186,12 @@ int bus_scope_set_property(
                 r = bus_kill_context_set_transient_property(u, &s->kill_context, name, message, flags, error);
                 if (r != 0)
                         return r;
+
+                if (streq(name, "User"))
+                        return bus_set_transient_user_relaxed(u, name, &s->user , message, flags, error);
```
poettering: spurious space?

src/core/scope.c
```diff
@@ -353,36 +360,78 @@ static void scope_enter_signal(Scope *s, ScopeState state, ScopeResult f) {
...
+static int scope_enter_chown_cgroup(Scope *s) {
+        Unit *u = UNIT(s);
+        pid_t pid;
         int r;
 
         assert(s);
 
-        if (unit_has_name(u, SPECIAL_INIT_SCOPE))
-                return -EPERM;
+        r = scope_arm_timer(s, usec_add(now(CLOCK_MONOTONIC), u->job_running_timeout));
```
poettering: hmm, that's a surprising timeout to use. I think we probably should introduce a new field in scope units `timeout_start_usec` to match service units, and then use that for this purpose. that would be cleaner...

src/core/scope.c
```diff
@@ -353,36 +360,78 @@ static void scope_enter_signal(Scope *s, ScopeState state, ScopeResult f) {
...
 }
 
-static int scope_start(Unit *u) {
-        Scope *s = SCOPE(u);
+static int scope_enter_chown_cgroup(Scope *s) {
+        Unit *u = UNIT(s);
+        pid_t pid;
         int r;
 
         assert(s);
```
poettering, line 368: scope_enter_chown_cgroup() expects to be called with s->user set, right?

So we should have an assert for that here.

Alternatively, move the check before the scope_enter_chown_cgroup() invocation into this function, and then make it return 0 if there's no chowning to be done, and > 0 if we are now in the chowning state.

That way, there would be only place to check the setting (i.e. inside this function) instead of two (in the caller via `if` and here via `assert()`).

up to you...

src/core/scope.c
```diff
@@ -585,6 +675,11 @@ static int scope_dispatch_timer(sd_event_source *source, usec_t usec, void *user
                 scope_enter_dead(s, SCOPE_FAILURE_TIMEOUT);
                 break;
 
+        case SCOPE_CHOWN_CGROUP:
+                log_unit_warning(UNIT(s), "User lookup timed out. Entering failed mode.");
```
poettering: mode? state!

poettering (changes requested):
looks excellent, just some minor things,

src/core/scope.c
msekletar: So maybe we should use default manager's start timeout (`default_timeout_start_usec`)? Frankly, I don't think that another user configurable timeout is needed. I can't think of the case when the sys admin would want to configure such low-level detail, especially for operation that shouldn't take too much time anyway (chown() on cgroupfs should be very quick).

In any case, if you insist I will add the explicit option.

src/core/scope.c
msekletar: I will add an `assert()` and keep the `if` in the caller because I don't want to obscure the fact that we `chown()` the cgroup (i.e. we actually enter the state) only if delegation is turned on and `user` is set.

msekletar:
@poettering updated!

I just remembered, there is one more important thing that may need to be addressed before this is finally good to go. That is (de)serialization. I am not sure if PIDs watched by the unit are rewatched automatically after deserialisation. Not dropping "needs-rework" just yet.

src/core/scope.c
poettering: looks good

src/core/scope.c
```diff
@@ -39,6 +42,7 @@ static void scope_init(Unit *u) {
         s->runtime_max_usec = USEC_INFINITY;
         s->timeout_stop_usec = u->manager->default_timeout_stop_usec;
         u->ignore_on_isolate = true;
+        s->user = s->group = NULL;
```
poettering, line 45: you can drop this actually, all fields are initialized to NULL anyway initially

poettering:
(this shouldnt need serialization afaics, hence the merge)

PR #23645, opened 2022-06-06
journalctl: Use STATIC_DESTRUCTOR_REGISTER()

daandemeyer:


src/journal/journalctl.c
```diff
@@ -138,6 +139,19 @@ static pcre2_code *arg_compiled_pattern = NULL;
...
 #endif
 
+STATIC_DESTRUCTOR_REGISTER(arg_file, strv_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_facilities, set_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_verify_key, freep);
+STATIC_DESTRUCTOR_REGISTER(arg_syslog_identifier, strv_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_system_units, strv_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_user_units, strv_freep);
+STATIC_DESTRUCTOR_REGISTER(arg_root, freep);
+STATIC_DESTRUCTOR_REGISTER(arg_image, freep);
```
medhefgo, line 149: So this was leaked before, right?
Also, might as well sort these…

medhefgo (approved):

src/journal/journalctl.c
daandemeyer: I went for the same order as they appear in the list above, I figure that's more consistent than going alphabetically. And yeah, I guess it was leaked before since it didn't get freed at the end unlike the other ones.

Issue #23648, opened 2022-06-06
rootlibdir broken on multiarch


mbiebl:
While working on #23585 I noticed that our current code to compute rootlibdir is broken.
If `rootlibdir` is not explicitly set, it uses
```
rootlibdir = get_option('rootlibdir')
if rootlibdir == ''
        rootlibdir = rootprefixdir / libdir.split('/')[-1]
endif
```
On a multiarch enabled system like Debian, `libdir` defaults to `/usr/lib/$(DEB_HOST_MULTIARCH)`, e.g. /usr/lib/x86_64-linux-gnu/

This will lead to rootlibdir being set to `/x86_64-linux-gnu/`
```
$ meson build
$ grep ROOTLIBDIR build/config.h 
#define ROOTLIBDIR "/x86_64-linux-gnu"
```

In Debian we dodge this issue by explicitly setting `-Drootlibdir=` in debian/rules.

floppym:
libdir defaults to "lib" generally. Overriding it as "lib64" or "/usr/lib64" is typical on amd64. Given that you could technically use any path for libdir, I'm not sure there is a to reliably derive rootlibdir from it in all cases.

mbiebl:
> libdir defaults to "lib" generally.

Not on multiarch enabled systems, ie. Debian/Ubuntu. There meson defaults to `/usr/lib/$(DEB_HOST_MULTIARCH)`

mbiebl:
> I'm not sure there is any way to reliably derive rootlibdir from it in all cases.

Instead of using `libdir.split('/')[-1]` I was thinking of removing `prefixdir` from `libdir` as a more robust solution.

mbiebl:
> > libdir defaults to "lib" generally.
> 
> Not on multiarch enabled systems, ie. Debian/Ubuntu. There meson defaults to `/usr/lib/$(DEB_HOST_MULTIARCH)`

https://github.com/mesonbuild/meson/blob/master/mesonbuild/mesonlib/universal.py#L951

floppym:
I think this might do the trick:
```
rootlibdir = get_option('rootlibdir')
if rootlibdir == ''
        rootlibdir = rootprefixdir / get_option('libdir')
elif not rootlibdir.startswith('/')
        rootlibdir = rootprefixdir / rootlibdir
endif
```
This works even if you call `meson --prefix /usr --libdir /usr/lib/x86_64-linux-gnu` because meson automatically strips prefix from libdir if it is a subdirectory of prefix.

PR #23650, opened 2022-06-07
meson: adjust rootlibdir default for multiarch

floppym:
On Debian, libdir is commonly something like 'lib/x86_64-linux-gnu'.

The result of get_option('libdir') is normalized to a prefix-relative
path by meson, so we can just append it to rootprefixdir.

Fixes https://github.com/systemd/systemd/issues/23648.


mbiebl:
LGTM, thanks!

PR #23653, opened 2022-06-07
cryptsetup: improve password prompt text

aafeijoo-suse:
Instead of always asking for passphrase, if the device has LUKS2 header check:
- If only regular passphrases are registered, ask for passphrase.
- If only recovery keys are registered, ask for recovery key.
- If both regular passphrases and recovery keys are registered, ask for
passphrase or recovery key.

Extracted from the TODO list: https://github.com/systemd/systemd/blob/369151c9c73b12fb7a88fc2b558499c2d4832982/TODO#L632-L633

### DIFF
--- a/TODO
+++ b/TODO
@@ -826,9 +826,6 @@ Features:
 
 * introduce a new group to own TPM devices
 
-* cryptsetup: if only recovery keys are registered and no regular passphrases,
-  ask user for "recovery key", not "passphrase"
-
 * cyptsetup: add option for automatically removing empty password slot on boot
 
 * cryptsetup: optionally, when run during boot-up and password is never


src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,96 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
 
+static PasswordRegisteredType check_registered_passwords(struct crypt_device *cd) {
+
+        int slots = 0, slot_max;
+        PasswordRegisteredType r = _PASSWORD_REGISTERED_TYPE_INVALID;
+
+        assert(cd);
+
+        if (!streq_ptr(crypt_get_type(cd), CRYPT_LUKS2)) {
+                log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Non-LUKS2 devices only support passphrase.");
```
keszybz: Please drop the errno here, just use `log_debug()`. And it'd be better to say something like
```C
log_debug("%s: not a LUKS2 device, only passphrases are supported", pretty);
```
(`pretty` can be passed in by the caller.)

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,96 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+        /* Search all used slots */
+        assert_se((slot_max = crypt_keyslot_max(CRYPT_LUKS2)) > 0);
+        for (int slot = 0; slot < slot_max; slot++) {
+                crypt_keyslot_info status;
+
+                status = crypt_keyslot_status(cd, slot);
+                if (!IN_SET(status, CRYPT_SLOT_ACTIVE, CRYPT_SLOT_ACTIVE_LAST))
+                        continue;
+
+                slots += slot + 1;
```
keszybz: Hmm, I'm not sure if this works. If slots can be referenced multiple times, subtraction below could arrive at 0 even if some slot is not referenced. Can you explain (in a comment) how this is supposed to work?
(It might be easier to just use a set or a `bool array[slot_max]`).

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,96 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+                _cleanup_(json_variant_unrefp) JsonVariant *v = NULL;
+                const char *type;
+                JsonVariant *w, *z;
+                int tk;
+
+                tk = cryptsetup_get_token_as_json(cd, token, NULL, &v);
+                if (IN_SET(tk, -ENOENT, -EINVAL))
+                        continue;
+                if (tk < 0) {
+                        log_warning_errno(tk, "Failed to read JSON token data off disk, ignoring: %m");
```
keszybz: "off disk" is assuming too much. This might not even be a disk. Just say "Failed to read JSON token data".

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,96 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+                }
+
+                w = json_variant_by_key(v, "type");
+                if (!w || !json_variant_is_string(w)) {
+                        log_warning("Token JSON data lacks type field, ignoring.");
+                        continue;
+                }
+
+                type = json_variant_string(w);
+                if (streq(type, "systemd-recovery") || streq(type, "systemd-pkcs11") || streq(type, "systemd-fido2") || streq(type, "systemd-tpm2")) {
```
keszybz: `STR_IN_SET(type, "systemd-recovery", "systemd-pkcs11", …)`

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,96 @@ static char *friendly_disk_name(const char *src, const char *vol) {
         return name_buffer;
 }
 
+static PasswordRegisteredType check_registered_passwords(struct crypt_device *cd) {
+
+        int slots = 0, slot_max;
+        PasswordRegisteredType r = _PASSWORD_REGISTERED_TYPE_INVALID;
```
keszybz: We reserve `r` for `int r`. Please call this something more meaningful, e.g. `type`.

src/cryptsetup/cryptsetup.c
```diff
@@ -46,6 +46,14 @@
 #define CRYPT_SECTOR_SIZE 512
 #define CRYPT_MAX_SECTOR_SIZE 4096
 
+typedef enum PasswordRegisteredType {
+        PASSWORD_PASSPHRASE,
+        PASSWORD_RECOVERY_KEY,
+        PASSWORD_PASSPHRASE_AND_RECOVERY_KEY,
+        _PASSWORD_REGISTERED_TYPE_MAX,
+        _PASSWORD_REGISTERED_TYPE_INVALID = -EINVAL,
+} PasswordRegisteredType;
```
keszybz: This is all passed within one file, so I don't think it's useful to deal bother with table. Why not just move the enum into the function (to make the code clearer), but return a string? Something like this:
```C
static const char* registered_passwords_phrase(const char *pretty, struct crypt_device *cd) {
   if (!streq_ptr(crypt_get_type(cd), CRYPT_LUKS2)) {
       ...
       return "passphrase";
  }
   ...
  if (…)
      return "passphrase";
  else
     return "passphrase or recovery key";
}
```

keszybz (changes requested):

src/cryptsetup/cryptsetup.c
aafeijoo-suse: Ok, thanks for this hint.

src/cryptsetup/cryptsetup.c
aafeijoo-suse: Ok.

src/cryptsetup/cryptsetup.c
aafeijoo-suse: Message copied from https://github.com/systemd/systemd/blob/c3191c6d4f3c66b78d66f1fa31f63d92db6f6fd7/src/cryptenroll/cryptenroll-list.c#L46-L52, so maybe that one should be changed as well.

src/cryptsetup/cryptsetup.c
aafeijoo-suse: I think you're right, and currently for LUKS2 it's at most 32 key-slots, so better keep the logic simple.

src/cryptsetup/cryptsetup.c
aafeijoo-suse: I'm not sure how to handle with this approach the case where all the slots are referenced by systemd tokens, but no recovery key is present.

src/cryptsetup/cryptsetup.c
aafeijoo-suse: Would it be enough to use `crypt_get_device_name(cd)`?

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,98 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+        assert(cd);
+
+        if (!streq_ptr(crypt_get_type(cd), CRYPT_LUKS2)) {
+                log_debug("%s: not a LUKS2 device, only passphrases are supported", crypt_get_device_name(cd));
+                return PASSWORD_PASSPHRASE;
+        }
+
+        /* Search all used slots */
+        assert_se((slot_max = crypt_keyslot_max(CRYPT_LUKS2)) > 0);
+        bool slots[slot_max];
```
poettering: hmm, i'd prefer if this was just heap memory. the maximum might be huge one day. I mean, I think it's relatively small right now, but who knows what happens if there's a revision of luks2. i.e. if this suddenly allows a million or two, then things will explode.

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,98 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+        for (int slot = 0; slot < slot_max; slot++) {
+                crypt_keyslot_info status;
+
+                status = crypt_keyslot_status(cd, slot);
+                slots[slot] = IN_SET(status, CRYPT_SLOT_ACTIVE, CRYPT_SLOT_ACTIVE_LAST);
```
poettering: merge this three lines into one?

```c
slots[slot] = IN_SET(crypt_keyslot_status(cd, slot), CRYPT_SLOT_ACTIVE, CRYPT_SLOT_ACTIVE_LAST);
```

src/cryptsetup/cryptsetup.c
```diff
@@ -46,6 +46,14 @@
 #define CRYPT_SECTOR_SIZE 512
 #define CRYPT_MAX_SECTOR_SIZE 4096
 
+typedef enum PasswordRegisteredType {
```
poettering: sounds like a misnomer, this is not about passwords really. It's about the two flavours of passphrase. hence maybe do:

```
typedef enum PassphraseType {
        PASSPHRASE_NONE,
        PASSPHRASE_REGULAR = 1 << 0,
        PASSPHRASE_RECOVERY_KEY = 1 << 1,
        PASSPHRASE_BOTH = PASSPHRASE_REGULAR|PASSPHRASE_RECOVERY_KEY,
        _PASSPHRASE_TYPE_MAX,
        _PASSPHRASE_TYPE_INVALID = -1,
} PassphraseType;
```

(as long as we only have two real types here we have the luxury we can make this a bit field without losing density, so let's do this)

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +584,98 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+                                        continue;
+                                }
+
+                                at = safe_atou(json_variant_string(z), &u);
+                                if (at < 0) {
+                                        log_warning_errno(at, "Token JSON data's keyslot field is not an integer formatted as string, ignoring.");
+                                        continue;
+                                }
+
+                                slots[u] = false;
```
poettering, line 663: please verify that `u` is `< slot_max` first...

otherwise this is a memory corruption triggerable by an attacker

poettering (changes requested):
Looks pretty good, but some nitpicks

poettering:
BTW, please include a commit in this PR that removes the relevant line from the TODO file in the git repo main dir:

```
* cryptsetup: if only recovery keys are registered and no regular passphrases,
  ask user for "recovery key", not "passphrase"
``

File: src/cryptsetup/cryptsetup.c (hunk 10 above)
Comment (aafeijoo-suse) [reply src/cryptsetup/cryptsetup.c:663]: Right, thanks.

File: src/cryptsetup/cryptsetup.c (hunk 9 above)
Comment (aafeijoo-suse) [reply src/cryptsetup/cryptsetup.c]: Ok.

File: src/cryptsetup/cryptsetup.c (hunk 8 above)
Comment (aafeijoo-suse) [reply src/cryptsetup/cryptsetup.c]: Ok.

File: src/cryptsetup/cryptsetup.c (hunk 7 above)
Comment (aafeijoo-suse) [reply src/cryptsetup/cryptsetup.c]: Ok, let's stay safe. But if the maximum number of slots grows to a million, I don't know if it's worth the time lost in iterating all of them just for the purpose of this PR.

File: src/cryptsetup/cryptsetup.c [hunk 11]
```diff
@@ -565,11 +585,103 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+
+        /* Search all used slots */
+        assert_se((slot_max = crypt_keyslot_max(CRYPT_LUKS2)) > 0);
+        slots = new(bool, slot_max);
+        if (!slots)
+                return log_oom();
+
+        for (int slot = 0; slot < slot_max; slot++) {
+                slots[slot] = IN_SET(crypt_keyslot_status(cd, slot), CRYPT_SLOT_ACTIVE, CRYPT_SLOT_ACTIVE_LAST);
+        }
```
poettering: please remove {} around single line for blocks.

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +585,103 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+                        log_warning("Token JSON data lacks type field, ignoring.");
+                        continue;
+                }
+
+                type = json_variant_string(w);
+                if (STR_IN_SET(type, "systemd-recovery", "systemd-pkcs11", "systemd-fido2", "systemd-tpm2")) {
+
+                        /* At least exists one recovery key */
+                        if (passphrase_type == PASSPHRASE_NONE && streq(type, "systemd-recovery"))
+                                passphrase_type = PASSPHRASE_RECOVERY_KEY;
```
poettering: given its a bitfield now you can do:

```c
if (streq(type, "systemd-recovery"))
        passphrase_type |= PASSPHRASE_RECOVERY_KEY;
```

src/cryptsetup/cryptsetup.c
```diff
@@ -565,11 +585,103 @@ static char *friendly_disk_name(const char *src, const char *vol) {
...
+
+                                slots[u] = false;
+                        }
+                }
+        }
+
+        /* Check if any of the slots is not referenced by systemd tokens */
+        for (int slot = 0; slot < slot_max; slot++)
+                if (slots[slot])
+                        return (passphrase_type == PASSPHRASE_RECOVERY_KEY) ? PASSPHRASE_BOTH : PASSPHRASE_REGULAR;
```
poettering: I'd do:

```c
if (slots[slot]) {
        passphrase_type |= PASSPHRASE_REGULAR;
        break;
}
``

poettering:
looks good, just some nitpicking

src/cryptsetup/cryptsetup.c
aafeijoo-suse: ok

src/cryptsetup/cryptsetup.c
aafeijoo-suse: ok, it makes sense

src/cryptsetup/cryptsetup.c
aafeijoo-suse: ok, sorry

aafeijoo-suse:
sorry, somehow my email server was blocking some notifications from github last week

poettering:
thanks! lgtm!

PR #23658, opened 2022-06-07
Fixes for NULL-pointer arguments based on gcc warnings

keszybz:
Continuation of #23654.

PR #23654, opened 2022-06-07
Fixes for NULL-pointer arguments based on gcc warnings

keszybz:

Issue #23646, opened 2022-06-06
../src/fundamental/string-util-fundamental.h:25:21: error: argument 2 null where non-null expected [-Werror=nonnull]


evverx:
**systemd version the issue has been seen with**
3cf63830acdef9d8afd

It can be triggered by `-finline-limit=1000` (which is passed by `honggfuzz` by default) and `-O2`:
```sh
FAILED: src/basic/libbasic.a.p/unit-file.c.o
cc -Isrc/basic/libbasic.a.p -Isrc/basic -I../src/basic -Isrc/fundamental -I../src/fundamental -Isrc/systemd -I../src/systemd -I. -I.. -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=gnu11 -O2 -g -Wno-format-signedness -Wno-missing-field-initializers -Wno-unused-parameter -Wdate-time -Wendif-labels -Werror=format=2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=overflow -Werror=override-init -Werror=return-type -Werror=shift-count-overflow -Werror=shift-overflow=2 -Werror=undef -Wfloat-equal -Wimplicit-fallthrough=5 -Winit-self -Wlogical-op -Wmissing-include-dirs -Wmissing-noreturn -Wnested-externs -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wshadow -Wstrict-aliasing=2 -Wstrict-prototypes -Wsuggest-attribute=noreturn -Wunused-function -Wwrite-strings -Wno-unused-result -Werror=missing-declarations -Werror=missing-prototypes -fdiagnostics-show-option -fno-common -fno-strict-aliasing -fstack-protector -fstack-protector-strong -fvisibility=hidden --param=ssp-buffer-size=4 -Werror=shadow -include config.h -finline-limit=10000 -fPIC -pthread -fvisibility=default -MD -MQ src/basic/libbasic.a.p/unit-file.c.o -MF src/basic/libbasic.a.p/unit-file.c.o.d -o src/basic/libbasic.a.p/unit-file.c.o -c ../src/basic/unit-file.c
In file included from ../src/basic/string-util.h:10,
                 from ../src/basic/path-util.h:9,
                 from ../src/basic/dirent-util.h:9,
                 from ../src/basic/unit-file.c:6:
In function ‘add_names’,
[...]

PR #23659, opened 2022-06-07
ci: build systemd with -O2 and -finline-limit=1000

evverx:
It should help to catch issues like https://github.com/systemd/systemd/issues/23646.

It should also make it easier to build `systemd` with `honggfuzz` where `-finline-limit` is passed by default.

Depends on https://github.com/systemd/systemd/pull/23654 and https://github.com/systemd/systemd/pull/23658


evverx:
Closing because GHActions are already clogged even without this.

Issue #23666, opened 2022-06-08
test-seccomp/test_lock_personality fails under ASan on ppc64le


mrc0mmand:
**systemd version the issue has been seen with**
 > latest main

**Used distribution**
 > CentOS 8 Stream, Fedora 36

**Linux kernel version used** (`uname -a`)

 > `4.18.0-383.el8.ppc64le`, `5.17.12-300.fc36.ppc64le`

**CPU architecture issue was seen on**
 > ppc64le

Without ASan:

```
$ sudo build-basic/test-seccomp
...
/* test_lock_personality */
current personality=0
Failed to enable personality lock for architecture ppc, skipping: Operation canceled
Failed to enable personality lock for architecture ppc64, skipping: Operation canceled
lockpersonalityseccomp succeeded.
```

With ASan:
```
$ sudo -E build/test-seccomp
...
/* test_lock_personality */
current personality=0
Failed to enable personality lock for architecture ppc, skipping: Operation canceled
Failed to enable personality lock for architecture ppc64, skipping: Operation canceled
Assertion '(unsigned long) safe_personality(current) == current' failed at src/test/test-seccomp.c:946, function test_lock_personality(). Aborting.
lockpersonalityseccomp terminated by signal ABRT.
Assertion 'wait_for_terminate_and_check("lockpersonalityseccomp", pid, WAIT_LOG) == EXIT_SUCCESS' failed at src/test/test-seccomp.c:971, function test_lock_personality(). Aborting.
Aborted
```

There's also this:
```
$ ASAN_OPTIONS= sudo -E build/test-seccomp
...
Failed to add rule for system call utimensat_time64() / -10240, ignoring: Numerical argument out of domain
System call utrap_install is not known, ignoring.
Failed to add rule for system call vm86old() / -10072, ignoring: Numerical argument out of domain
@known succeeded.
AddressSanitizer:DEADLYSIGNAL
=================================================================
==313598==ERROR: AddressSanitizer: SEGV on unknown address 0x000010031be0 (pc 0x7ffff783677c bp 0x7fffffffe8e0 sp 0x7fffffffe8e0 T0)
==313598==The signal is caused by a UNKNOWN memory access.
    #0 0x7ffff783677c in __asan_stack_free_1 (/lib64/libasan.so.8+0x4677c)
    #1 0x10015b04 in test_filter_sets ../src/test/test-seccomp.c:163
    #2 0x10012824 in run_test_table ../src/shared/tests.h:106
[...]

PR #23667, opened 2022-06-08
test: account for ADDR_NO_RANDOMIZE on ppc64le w/ sanitizers

mrc0mmand:
On pp64le sanitizers disable ASLR, but the additional
`ADDR_NO_RANDOMIZE` flag is not honored by `opinionated_personality()`.
Let's tweak the flag in such cases manually.

See: https://github.com/llvm/llvm-project/commit/78f7a6eaa601bfdd6ae70ffd3da2254c21ff77f9

Resolves: #23666

### DIFF
--- a/src/test/test-seccomp.c
+++ b/src/test/test-seccomp.c
@@ -934,8 +934,15 @@ TEST(lock_personality) {
         }
 
         assert_se(opinionated_personality(&current) >= 0);
+        /* On ppc64le sanitizers disable ASLR (i.e. by setting ADDR_NO_RANDOMIZE),
+         * which opinionated_personality() doesn't return. Let's tweak the current
+         * personality ourselves in such cases.
+         * See: https://github.com/llvm/llvm-project/commit/78f7a6eaa601bfdd6ae70ffd3da2254c21ff77f9
+         */
+        if (FLAGS_SET(safe_personality(PERSONALITY_INVALID), ADDR_NO_RANDOMIZE))
+                current |= ADDR_NO_RANDOMIZE;
 
-        log_info("current personality=%lu", current);
+        log_info("current personality=0x%lX", current);
 
         pid = fork();
         assert_se(pid >= 0);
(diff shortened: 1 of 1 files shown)


mrc0mmand:
Another possible solution without ifdef-ery might be:

```diff
diff --git a/src/test/test-seccomp.c b/src/test/test-seccomp.c
index 45fe8f7c59..eb2e9ab15f 100644
--- a/src/test/test-seccomp.c
+++ b/src/test/test-seccomp.c
@@ -934,6 +934,8 @@ TEST(lock_personality) {
         }
 
         assert_se(opinionated_personality(&current) >= 0);
+        if (safe_personality(PERSONALITY_INVALID) & ADDR_NO_RANDOMIZE)
+                current |= ADDR_NO_RANDOMIZE;
 
         log_info("current personality=%lu", current);
 
@@ -948,10 +950,11 @@ TEST(lock_personality) {
                 /* Note, we also test that safe_personality() works correctly, by checkig whether errno is properly
                  * set, in addition to the return value */
                 errno = 0;
-                assert_se(safe_personality(PER_LINUX | ADDR_NO_RANDOMIZE) == -EPERM);
+                assert_se(safe_personality(PER_LINUX | MMAP_PAGE_ZERO) == -EPERM);
                 assert_se(errno == EPERM);
 
-                assert_se(safe_personality(PER_LINUX | MMAP_PAGE_ZERO) == -EPERM);
+                if (!FLAGS_SET(current, ADDR_NO_RANDOMIZE))
+                        assert_se(safe_personality(PER_LINUX | ADDR_NO_RANDOMIZE) == -EPERM);
                 assert_se(safe_personality(PER_LINUX | ADDR_COMPAT_LAYOUT) == -EPERM);
                 assert_se(safe_personality(PER_LINUX | READ_IMPLIES_EXEC) == -EPERM);
                 assert_se(safe_personality(PER_LINUX_32BIT) == -EPERM);
```

which is a bit nicer (and shorter). Not sure if we want to be explicit with the ppc64le arch or make this more generic & shorter.

evverx:
@mrc0mmand I think it would make sense to just skip that test there and call it a day :-) instead of trying to make it work. Given that the test boils down to calling `personality` provided by `glibc` I'm not sure how useful it is to run it under ASan in general.

Just out of curiosity, are there any other tests failing under ASan there? Last time I checked the fuzz targets went sideways almost immediately with weird segfaults.

mrc0mmand:
In the end I went for the *simple fix* over disabling the test completely, since in this particular case the fix is fairly trivial. If it causes any other issues in the future, we can disable it altogether, as done in a couple of other seccomp-related tests.

yuwata (approved):
LGTM.

evverx:
> In the end I went for the simple fix over disabling the test completely, since in this particular case the fix is fairly trivial. If it causes any other issues in the future, we can disable it altogether, as done in a couple of other seccomp-related tests.

Fair enough :-)

> So far it looks like that's the only one failing. Once this is fixed, I should be able to run unit & integration tests on ppc64le under sanitizers at least daily in CentOS CI (https://github.com/systemd/systemd-centos-ci/pull/308).

Good to know. Thanks!

PR #23550, opened 2022-05-29
oss-fuzz: support i386

evverx:
Closes https://github.com/systemd/systemd/issues/23532

It's being tested in https://github.com/google/oss-fuzz/pull/7777

PR #23670, opened 2022-06-08
cifuzz: build and run fuzz targets on i386 as well

evverx:
It's a follow-up to https://github.com/systemd/systemd/pull/23550.

https://github.com/google/oss-fuzz/pull/7779 has just been merged so it's unclear whether it's included in the CIFuzz image yet.


mrc0mmand (approved):
Nice, thank you!

PR #23672, opened 2022-06-08
meson: Build with frame pointers in developer mode

daandemeyer:
Profiling tools tend to work better when binaries and libraries
are compiled with frame pointers as without them there's no easy
and fast way to get the current stacktrace.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -450,6 +450,10 @@ if get_option('buildtype') != 'debug'
         possible_link_flags += '-Wl,--gc-sections'
 endif
 
+if get_option('mode') == 'developer'
+        possible_cc_flags += '-fno-omit-frame-pointer'
+endif
+
 add_project_arguments(cc.get_supported_arguments(basic_disabled_warnings), language : 'c')
 add_project_arguments(cc.get_supported_arguments(possible_cc_flags), language : 'c')
 add_project_link_arguments(cc.get_supported_link_arguments(possible_link_flags), language : 'c')


meson.build
```diff
@@ -450,6 +450,10 @@ if get_option('buildtype') != 'debug'
         possible_link_flags += '-Wl,--gc-sections'
 endif
 
+if get_option('mode') == 'developer'
+        possible_cc_flags += ['-fno-omit-frame-pointer']
+endif
```
yuwata, line 455: Not sure this is conditionalized. If it is useful to get stacktrace, then should this be enabled unconditionally?
E.g. fedora provides debuginfo packages but built with `-mode=release`.
cc @keszybz ?

daandemeyer: Whether to do this unconditionally should be a distro decision. I'm actually drafting a change proposal for Fedora to build with -fno-omit-frame-pointer by default, but the decision should still be left to the distro I believe.

meson.build
```diff
@@ -450,6 +450,10 @@ if get_option('buildtype') != 'debug'
         possible_link_flags += '-Wl,--gc-sections'
 endif
 
+if get_option('mode') == 'developer'
+        possible_cc_flags += ['-fno-omit-frame-pointer']
```
medhefgo: You can actually drop the `[]`

medhefgo (approved):

yuwata (approved):
OK. Then LGTM. But leave this to @keszybz.

meson.build
keszybz: I have to admit that I don't know how this works exactly. We get the `-debuginfo` packages and they seem to work, we get a full backtrace with all details. But if something is worth changing there, I agree that a Change proposal is the way to go.

keszybz:
I think this is reasonable. Gcc turns `-fomit-frame-pointer` on for `-O1`, so it makes sense to disable that to get easier debugging in developer mode.

bluca (approved):

PR #23675, opened 2022-06-08
Use udev_available() where appropriate

enr0n:
I came across a couple places where the use of `path_is_read_only_fs("/sys")` is intended to check if udev is available.


yuwata (approved):
LGTM.

PR #23677, opened 2022-06-09
bpf: use __always_inline macro in restrict-ifaces.bpf.c

jameshilliard:
This appears to be more reliable at inlining with some compilers.

Note we use this in the other BPF programs:
https://github.com/systemd/systemd/blob/377157e6bd743af8de1e89b6109282de4fc5a066/src/core/bpf/socket_bind/socket-bind.bpf.c#L85

### DIFF
--- a/src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
+++ b/src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
@@ -20,7 +20,7 @@ struct {
 #define DROP 0
 #define PASS 1
 
-static inline int restrict_network_interfaces_impl(const struct __sk_buff *sk) {
+static __always_inline int restrict_network_interfaces_impl(const struct __sk_buff *sk) {
         __u32 zero = 0, ifindex;
         __u8 *lookup_result;
 


src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
```diff
@@ -20,7 +20,7 @@ struct {
 #define DROP 0
 #define PASS 1
 
-static inline int restrict_network_interfaces_impl(const struct __sk_buff *sk) {
+static __always_inline int restrict_network_interfaces_impl(const struct __sk_buff *sk) {
```
yuwata, line 23: We usually not set `inline` or `__always_inline` for function in .c files. Why this needs to be inlined? Could you elaborate more?

yuwata:
BTW, we do not usually use Signed-off-by tag. Please drop next time when you post a PR.

yuwata:
Hmm, most CIs are unhappy...

yuwata:
Could you rebase and force-push this PR?

src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
jameshilliard: This is a BPF program which can hit errors like this if in-lining fails:
```c
error: indirect call in function, which are not supported by eBPF
```

jameshilliard:
> Please drop next time when you post a PR.

Removed.

> Could you rebase and force-push this PR?

Rebased.

src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
yuwata: Ah, thanks.

yuwata (approved):
LGTM.

bluca (approved):

PR #23680, opened 2022-06-09
Move boot counting into BLS proper

keszybz:


medhefgo (approved):

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -391,25 +391,77 @@ creating a partition and file system for it) and creates the `/loader/entries/`
...
+## Boot counting
+
+The main idea is that when boot entries are initially installed, they are
+marked as "indeterminate" and assigned a number of boot attempts. Each time the
+boot loader tries to boot an entry, it decreases this count by one. If the
+operating system considers the boot as successful, it removes the counter
+altogether and the entry becomes "good". Otherwise, once the assigned number of
+boots is exhausted, the entry is marked as "bad".
+
+Which boots are "successful" is determined by the operating system. Systemd
```
bluca: /Systemd/systemd/

keszybz: Hmm, lower-case proper names are so awkward. Can we agree to say "Systemd" at the beggining of a sentence?

bluca (approved):

docs/BOOT_LOADER_SPECIFICATION.md
bluca: I don't mind one way or another, but I think right now it's mandated lower case everywhere? Ie, this should be consistent with the generic rule, and if we want to change that everywhere I'm fine with that, but probably separately

keszybz: OK. Updated with that one-bit change.

PR #23683, opened 2022-06-09
Default to default-status-unit-format=name in developer mode

keszybz:


meson.build
```diff
@@ -4171,6 +4171,7 @@ summary({
         'default net.naming-scheme value' : default_net_naming_scheme,
         'default KillUserProcesses value' : kill_user_processes,
         'default locale' :                  default_locale,
+        'default status unit format':       status_unit_format_default,
```
yuwata: Better to add a space before the colon?

yuwata (approved):

keszybz:
Thanks, updated.

keszybz:
CI passed previously, so let's assume that the whitespace change didn't break things ;)

PR #23686, opened 2022-06-10
network: set maximum length to be read by read_full_file_full()

yuwata:


DaanDeMeyer (approved):

DaanDeMeyer:
test_wireguard failed on centOS CI, that seems like it's related

src/network/netdev/macsec.c
```diff
@@ -971,8 +971,8 @@ static int macsec_read_key_file(NetDev *netdev, SecurityAssociation *sa) {
                 return 0;
 
         r = read_full_file_full(
-                        AT_FDCWD, sa->key_file, UINT64_MAX, SIZE_MAX,
-                        READ_FULL_FILE_SECURE | READ_FULL_FILE_UNHEX | READ_FULL_FILE_WARN_WORLD_READABLE | READ_FULL_FILE_CONNECT_SOCKET,
+                        AT_FDCWD, sa->key_file, UINT64_MAX, 16,
```
poettering, line 974: given the 16 is used at at least two places now, maybe add a macro for it? MACSEC_KEY_LEN?

poettering:
so my educated guess is that the CI stem from the fact that the size limit passed to read_full_file() refers to the *encoded* size, but the returned size of the *decoded* size. after all these calls do a base64 decode run on the data loaded from the file.

Now, we could scale the value by 8/6th, to deal with this, but that doesn't really work, since base64 typically contains whitespace/newlines and stuff which we ignore while parsing. But still, I guess we can assume that the encoded size is bounded in all common cases by the expected output size multiplied by some reasonably fixed factor.

hence maybe define a factor like this: 

```c
#define BASE64_ESTIMATED_AMPLIFICATION_BOUNDARY 3
```

or so. Which would mean when loading a base64 encoded file we'd assume that a file is at worst blown up by factor 3 by base64 encoding it, i.e. this is sho much extra whitespace, that noone should ever really need it...

yuwata:
Superseded by #29400.

PR #23688, opened 2022-06-10
boot/efi-string: check the end of haystack before testing remaining pattern

yuwata:
Fixes buffer-overflow reported at https://github.com/systemd/systemd/pull/23589#issuecomment-1151820341.


evverx:
The fuzz target no longer crashes there as far as I can tell. Thanks! I'd wait for @medhefgo though

medhefgo (approved):
Wups.

yuwata:
The CentOS CI failure is unrelated, and should be addressed by #23690.

PR #23584, opened 2022-05-31
test: import logind test from debian/ubuntu test suite

yuwata:

PR #23689, opened 2022-06-10
test: import timedated test from debian/ubuntu test suite

yuwata:


mrc0mmand (approved):
LGTM, thanks! I've flagged both this and #23584 with the `needs-sanitizer-coverage` label, so I don't forged to add these two tests to the CentOS CI sanitizer jobs.

PR #23691, opened 2022-06-10
efi: Add parse_number8/16

medhefgo:


src/boot/efi/efi-string.c
```diff
@@ -239,6 +239,38 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
         return efi_fnmatch_internal(pattern, haystack, 32);
 }
 
+#define DEFINE_PARSE_NUMBER(type, name)                                    \
+        bool name(const type *s, uint64_t *ret_u, const type **ret_tail) { \
```
poettering, line 243: hmm, why return "bool" instead of EFI_RESULT?

I mean, in userspace we have safe_atou32() and friends, and they do return errors codes. I am pretty sure we should try to do this similar here, and systematically return EFI_RESULT as errors, and not bools...

medhefgo: The only really meaningful efi statuses for this is `EFI_SUCCESS/EFI_INVALID_PARAMETER`. bool is simpler and also doesn't require pulling in efi.h/redifining this stuff for userspace testing.

poettering: Well, it's a matter of having a uniform coding style, and about having the ability to bubble up errors. Here you don't because you have to convert them to EFI errors first.

PR #23695, opened 2022-06-10
 boot: use efi_memcpy inplace of memcpy

sshedi:
Fixes: CID#1489876

### DIFF
--- a/src/boot/efi/util.h
+++ b/src/boot/efi/util.h
@@ -67,7 +67,7 @@ static inline void *xmalloc_multiply(size_t size, size_t n) {
 _malloc_ _alloc_(3) _returns_nonnull_ _warn_unused_result_
 static inline void *xrealloc(void *p, size_t old_size, size_t new_size) {
         void *r = xmalloc(new_size);
-        memcpy(r, p, MIN(old_size, new_size));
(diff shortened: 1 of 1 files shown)


yuwata:
It seems the issue is false positive. As memcpy(), which is an alias of efi_memcpy(), accepts NULL.

Maybe, explicitly (and unconditionally) calling efi_memcpy() is better.

cc @medhefgo

sshedi:
I think in this case, it's using __builtin_memcpy but agree we can use `efi_memcpy` directly instead of doing null check.

medhefgo:
I think it might be better to hoist the NULL checks out of efi_ mem functions and put them into the #define. Otherwise we end up playing whack-a-mole with these.

medhefgo:
Or maybe do as @yuwata suggested before and keep the definitions of mem* to the standard and use assert…

medhefgo:
I think in the end I prefer calling `efi_memcpy` unconditionally here. xstrndup8/16 already does the same.

sshedi:
Agree, will update the PR. Thanks for the inputs.

yuwata (approved):
LGTM too.

PR #23698, opened 2022-06-10
docs: desaturate dark-mode background color

bnf:
The documentations dark-mode background color as
added in #23417 was perceived to be too purple-y [1]
and is therefore replaced by a desaturated black
that is derived from the systemd brand-black using
12% less HSL saturation.

Preview:
https://systemd.bnf.dev/
systemd-io-bg-7-satuation

[1] https://github.com/systemd/systemd/pull/23417#issuecomment-1146323820


keszybz:
LGTM. CI doesn't care about this, so I'll merge immediately.

PR #23699, opened 2022-06-10
docs/CONTRIBUTING: ask people to send a ping on stalled reviews

keszybz:
A few related changes: be explicit that 'main' is the branch we're referring
to. There was a case recently of somebody rebasing on 'master' by mistake.
It's better to be explicit, since 'main' is still not the default for git
(AFAIK).

Also, github seems to send mails about force-pushes, so drop that as a
justification. Commenting about changes that were done is useful even if
github were to send a notification.

And finally, $subject. We do that ourselves, but outside contributors might not
know that that's expected.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -31,13 +31,14 @@ See [reporting of security vulnerabilities](SECURITY.md).
 
 ## Posting Pull Requests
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):

bluca (approved):

keszybz:
OK, let's just merge this.

PR #23701, opened 2022-06-10
boot: Use memcpy/memset provided by firmware

medhefgo:


medhefgo:
I tested the CopyMem from OVMF and of my ASUS board and both have the same performance as a naive memcpy that copies uint64_t at a time (instead of uint8_t that we currently do). So it seems worth it.

yuwata (approved):

PR #23706, opened 2022-06-10
boot: Add printf functions

medhefgo:
This is the last big (and somewhat complicated) piece to add.

### DIFF
--- a/src/boot/efi/boot.c
+++ b/src/boot/efi/boot.c
@@ -443,22 +443,9 @@ static bool unicode_supported(void) {
         return cache;
 }
 
-static void ps_string(const char16_t *fmt, const void *value) {
-        assert(fmt);
-        if (value)
-                Print(fmt, value);
-}
-
-static void ps_bool(const char16_t *fmt, bool value) {
-        assert(fmt);
-        Print(fmt, yes_no(value));
-}
-
 static bool ps_continue(void) {
-        if (unicode_supported())
-                Print(L"\n─── Press any key to continue, ESC or q to quit. ───\n\n");
-        else
-                Print(L"\n--- Press any key to continue, ESC or q to quit. ---\n\n");
+        const char16_t *sep = unicode_supported() ? u"───" : u"---";
+        printf("\n%ls Press any key to continue, ESC or q to quit. %ls\n\n", sep, sep);
 
         uint64_t key;
         return console_key_read(&key, UINT64_MAX) == EFI_SUCCESS &&
@@ -480,112 +467,143 @@ static void print_status(Config *config, char16_t *loaded_image_path) {
         secure = secure_boot_mode();
         (void) efivar_get(LOADER_GUID, L"LoaderDevicePartUUID", &device_part_uuid);
 
-        /* We employ some unusual indentation here for readability. */
-
-        ps_string(L"  systemd-boot version: %a\n",      GIT_VERSION);
-        ps_string(L"          loaded image: %s\n",      loaded_image_path);
-        ps_string(L" loader partition UUID: %s\n",      device_part_uuid);
-        ps_string(L"          architecture: %a\n",      EFI_MACHINE_TYPE_NAME);
-            Print(L"    UEFI specification: %u.%02u\n", ST->Hdr.Revision >> 16, ST->Hdr.Revision & 0xffff);
-        ps_string(L"       firmware vendor: %s\n",      ST->FirmwareVendor);
-            Print(L"      firmware version: %u.%02u\n", ST->FirmwareRevision >> 16, ST->FirmwareRevision & 0xffff);
-            Print(L"        OS indications: %lu\n",     get_os_indications_supported());
-            Print(L"           secure boot: %s (%s)\n", yes_no(IN_SET(secure, SECURE_BOOT_USER, SECURE_BOOT_DEPLOYED)), secure_boot_mode_to_string(secure));
-          ps_bool(L"                  shim: %s\n",      shim_loaded());
-          ps_bool(L"                   TPM: %s\n",      tpm_present());
-            Print(L"          console mode: %d/%ld (%" PRIuN L"x%" PRIuN L" @%ux%u)\n", ST->ConOut->Mode->Mode, ST->ConOut->Mode->MaxMode - INT64_C(1), x_max, y_max, screen_width, screen_height);
+        printf("  systemd-boot version: " GIT_VERSION "\n");
+        if (loaded_image_path)
+                printf("          loaded image: %ls\n", loaded_image_path);
+        if (device_part_uuid)
+                printf(" loader partition UUID: %ls\n", device_part_uuid);
+        printf("          architecture: " EFI_MACHINE_TYPE_NAME "\n");
+        printf("    UEFI specification: %u.%02u\n", ST->Hdr.Revision >> 16, ST->Hdr.Revision & 0xffff);
+        printf("       firmware vendor: %ls\n", ST->FirmwareVendor);
+        printf("      firmware version: %u.%02u\n", ST->FirmwareRevision >> 16, ST->FirmwareRevision & 0xffff);
+        printf("        OS indications: %#" PRIx64 "\n", get_os_indications_supported());
+        printf("           secure boot: %ls (%ls)\n",
+                        yes_no(IN_SET(secure, SECURE_BOOT_USER, SECURE_BOOT_DEPLOYED)),
+                        secure_boot_mode_to_string(secure));
+        printf("                  shim: %ls\n", yes_no(shim_loaded()));
+        printf("                   TPM: %ls\n", yes_no(tpm_present()));
+        printf("          console mode: %i/%" PRIi64 " (%zux%zu @%ux%u)\n",
+                        ST->ConOut->Mode->Mode, ST->ConOut->Mode->MaxMode - INT64_C(1),
+                        x_max, y_max, screen_width, screen_height);
 
         if (!ps_continue())
                 return;
 
         switch (config->timeout_sec_config) {
         case TIMEOUT_UNSET:
-            break;
+                break;
         case TIMEOUT_MENU_FORCE:
-            Print(L"      timeout (config): menu-force\n"); break;
+                printf("      timeout (config): menu-force\n");
+                break;
         case TIMEOUT_MENU_HIDDEN:
-            Print(L"      timeout (config): menu-hidden\n"); break;
+                printf("      timeout (config): menu-hidden\n");
+                break;
         default:
-            Print(L"      timeout (config): %u s\n", config->timeout_sec_config);
+                printf("      timeout (config): %u s\n", config->timeout_sec_config);
         }
 
         switch (config->timeout_sec_efivar) {
         case TIMEOUT_UNSET:
-            break;
+                break;
         case TIMEOUT_MENU_FORCE:
-            Print(L"     timeout (EFI var): menu-force\n"); break;
+                printf("     timeout (EFI var): menu-force\n");
+                break;
         case TIMEOUT_MENU_HIDDEN:
-            Print(L"     timeout (EFI var): menu-hidden\n"); break;
+                printf("     timeout (EFI var): menu-hidden\n");
+                break;
         default:
-            Print(L"     timeout (EFI var): %u s\n", config->timeout_sec_efivar);
+                printf("     timeout (EFI var): %u s\n", config->timeout_sec_efivar);
         }
 
-        ps_string(L"      default (config): %s\n", config->entry_default_config);
-        ps_string(L"     default (EFI var): %s\n", config->entry_default_efivar);
-        ps_string(L"    default (one-shot): %s\n", config->entry_oneshot);
-        ps_string(L"           saved entry: %s\n", config->entry_saved);
-          ps_bool(L"                editor: %s\n", config->editor);
-          ps_bool(L"          auto-entries: %s\n", config->auto_entries);
-          ps_bool(L"         auto-firmware: %s\n", config->auto_firmware);
-          ps_bool(L"                  beep: %s\n", config->beep);
-          ps_bool(L"  reboot-for-bitlocker: %s\n", config->reboot_for_bitlocker);
+        if (config->entry_default_config)
+                printf("      default (config): %ls\n", config->entry_default_config);
+        if (config->entry_default_efivar)
+                printf("     default (EFI var): %ls\n", config->entry_default_efivar);
+        if (config->entry_oneshot)
+                printf("    default (one-shot): %ls\n", config->entry_oneshot);
+        if (config->entry_saved)
+                printf("           saved entry: %ls\n", config->entry_saved);
+        printf("                editor: %ls\n", yes_no(config->editor));
+        printf("          auto-entries: %ls\n", yes_no(config->auto_entries));
+        printf("         auto-firmware: %ls\n", yes_no(config->auto_firmware));
+        printf("                  beep: %ls\n", yes_no(config->beep));
+        printf("  reboot-for-bitlocker: %ls\n", yes_no(config->reboot_for_bitlocker));
 
         switch (config->secure_boot_enroll) {
         case ENROLL_OFF:
-                Print(L"    secure-boot-enroll: off\n"); break;
+                printf("    secure-boot-enroll: off\n");
+                break;
         case ENROLL_MANUAL:
-                Print(L"    secure-boot-enroll: manual\n"); break;
+                printf("    secure-boot-enroll: manual\n");
+                break;
         case ENROLL_FORCE:
-                Print(L"    secure-boot-enroll: force\n"); break;
+                printf("    secure-boot-enroll: force\n");
+                break;
         default:
                 assert_not_reached();
         }
 
         switch (config->console_mode) {
         case CONSOLE_MODE_AUTO:
-            Print(L" console-mode (config): %s\n", L"auto"); break;
+                printf(" console-mode (config): auto\n");
+                break;
         case CONSOLE_MODE_KEEP:
-            Print(L" console-mode (config): %s\n", L"keep"); break;
+                printf(" console-mode (config): keep\n");
+                break;
         case CONSOLE_MODE_FIRMWARE_MAX:
-            Print(L" console-mode (config): %s\n", L"max"); break;
+                printf(" console-mode (config): max\n");
+                break;
         default:
-            Print(L" console-mode (config): %ld\n", config->console_mode); break;
+                printf(" console-mode (config): %" PRIi64 "\n", config->console_mode);
+                break;
         }
 
         /* EFI var console mode is always a concrete value or unset. */
         if (config->console_mode_efivar != CONSOLE_MODE_KEEP)
-            Print(L"console-mode (EFI var): %ld\n", config->console_mode_efivar);
+                printf("console-mode (EFI var): %" PRIi64 "\n", config->console_mode_efivar);
 
         if (!ps_continue())
                 return;
 
         for (UINTN i = 0; i < config->entry_count; i++) {
                 ConfigEntry *entry = config->entries[i];
+                EFI_DEVICE_PATH *dp = NULL;
+                _cleanup_free_ char16_t *dp_str = NULL;
 
-                _cleanup_free_ char16_t *dp = NULL;
-                if (entry->device)
-                        (void) device_path_to_str(DevicePathFromHandle(entry->device), &dp);
+                if (entry->device &&
+                    BS->HandleProtocol(entry->device, &(EFI_GUID) EFI_DEVICE_PATH_PROTOCOL_GUID, (void **) &dp) ==
+                                    EFI_SUCCESS)
+                        (void) device_path_to_str(dp, &dp_str);
 
-                    Print(L"  config entry: %" PRIuN L"/%" PRIuN L"\n", i + 1, config->entry_count);
-                ps_string(L"            id: %s\n", entry->id);
-                ps_string(L"         title: %s\n", entry->title);
-                ps_string(L"    title show: %s\n", streq16(entry->title, entry->title_show) ? NULL : entry->title_show);
-                ps_string(L"      sort key: %s\n", entry->sort_key);
-                ps_string(L"       version: %s\n", entry->version);
-                ps_string(L"    machine-id: %s\n", entry->machine_id);
-                ps_string(L"        device: %s\n", dp);
-                ps_string(L"        loader: %s\n", entry->loader);
+                printf("  config entry: %zu/%zu\n", i + 1, config->entry_count);
+                printf("            id: %ls\n", entry->id);
+                if (entry->title)
+                        printf("         title: %ls\n", entry->title);
+                if (entry->title_show && !streq16(entry->title, entry->title_show))
+                        printf("    title show: %ls\n", entry->title_show);
+                if (entry->sort_key)
+                        printf("      sort key: %ls\n", entry->sort_key);
+                if (entry->version)
+                        printf("       version: %ls\n", entry->version);
+                if (entry->machine_id)
+                        printf("    machine-id: %ls\n", entry->machine_id);
+                if (dp_str)
+                        printf("        device: %ls\n", dp_str);
+                if (entry->loader)
+                        printf("        loader: %ls\n", entry->loader);
                 STRV_FOREACH(initrd, entry->initrd)
-                    Print(L"        initrd: %s\n", *initrd);
-                ps_string(L"    devicetree: %s\n", entry->devicetree);
-                ps_string(L"       options: %s\n", entry->options);
-                  ps_bool(L" internal call: %s\n", !!entry->call);
+                        printf("        initrd: %ls\n", *initrd);
+                if (entry->devicetree)
+                        printf("    devicetree: %ls\n", entry->devicetree);
+                if (entry->options)
+                        printf("       options: %ls\n", entry->options);
+                printf(" internal call: %ls\n", yes_no(!!entry->call));
 
-                  ps_bool(L"counting boots: %s\n", entry->tries_left >= 0);
+                printf("counting boots: %ls\n", yes_no(entry->tries_left >= 0));
                 if (entry->tries_left >= 0) {
-                    Print(L"         tries: %u left, %u done\n", entry->tries_left, entry->tries_done);
-                    Print(L"  current path: %s\\%s\n",  entry->path, entry->current_name);
-                    Print(L"     next path: %s\\%s\n",  entry->path, entry->next_name);
+                        printf("         tries: %i left, %i done\n", entry->tries_left, entry->tries_done);
+                        printf("  current path: %ls\\%ls\n", entry->path, entry->current_name);
+                        printf("     next path: %ls\\%ls\n", entry->path, entry->next_name);
                 }
 
                 if (!ps_continue())
@@ -641,7 +659,7 @@ static bool menu_run(
         ST->ConOut->EnableCursor(ST->ConOut, false);
 
         /* draw a single character to make ClearScreen work on some firmware */
-        Print(L" ");
+        ST->ConOut->OutputString(ST->ConOut, (char16_t *) u" ");
 
         err = console_set_mode(config->console_mode_efivar != CONSOLE_MODE_KEEP ?
                                config->console_mode_efivar : config->console_mode);
@@ -2729,7 +2747,7 @@ out:
 EFI_STATUS efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *sys_table) {
         InitializeLib(image, sys_table);
 
-        debug_hook(L"systemd-boot");
+        debug_hook("systemd-boot");
         /* Uncomment the next line if you need to wait for debugger. */
         // debug_break();
 
--- a/src/boot/efi/secure-boot.c
+++ b/src/boot/efi/secure-boot.c
@@ -44,16 +44,16 @@ EFI_STATUS secure_boot_enroll_at(EFI_FILE *root_dir, const char16_t *path) {
 
         clear_screen(COLOR_NORMAL);
 
-        Print(u"Enrolling secure boot keys from directory: %s\n", path);
+        printf("Enrolling secure boot keys from directory: %ls\n", path);
 
         /* Enrolling secure boot keys is safe to do in virtualized environments as there is nothing
          * we can brick there. */
         if (!in_hypervisor()) {
-                Print(u"Warning: Enrolling custom Secure Boot keys might soft-brick your machine!\n");
+                printf("Warning: Enrolling custom Secure Boot keys might soft-brick your machine!\n");
 
                 unsigned timeout_sec = 15;
                 for (;;) {
-                        Print(u"\rEnrolling in %2u s, press any key to abort.", timeout_sec);
+                        printf("\rEnrolling in %2u s, press any key to abort.", timeout_sec);
 
                         uint64_t key;
                         err = console_key_read(&key, 1000 * 1000);
--- a/src/boot/efi/stub.c
+++ b/src/boot/efi/stub.c
@@ -420,7 +420,7 @@ static EFI_STATUS real_main(EFI_HANDLE image) {
 EFI_STATUS efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *sys_table) {
         InitializeLib(image, sys_table);
 
-        debug_hook(L"systemd-stub");
+        debug_hook("systemd-stub");
         /* Uncomment the next line if you need to wait for debugger. */
         // debug_break();
 
--- a/src/boot/efi/util.h
+++ b/src/boot/efi/util.h
@@ -11,20 +11,6 @@
 #define UINTN_MAX (~(UINTN)0)
 #define INTN_MAX ((INTN)(UINTN_MAX>>1))
 
-/* gnu-efi format specifiers for integers are fixed to either 64bit with 'l' and 32bit without a size prefix.
- * We rely on %u/%d/%x to format regular ints, so ensure the size is what we expect. At the same time, we also
- * need specifiers for (U)INTN which are native (pointer) sized. */
-assert_cc(sizeof(int) == sizeof(uint32_t));
-#if __SIZEOF_POINTER__ == 4
-#  define PRIuN L"u"
-#  define PRIiN L"d"
-#elif __SIZEOF_POINTER__ == 8
-#  define PRIuN L"lu"
-#  define PRIiN L"ld"
-#else
-#  error "Unexpected pointer size"
-#endif
-
 static inline void free(void *p) {
         if (!p)
                 return;
@@ -180,7 +166,7 @@ void debug_break(void);
 extern uint8_t _text, _data;
 /* Report the relocated position of text and data sections so that a debugger
  * can attach to us. See debug-sd-boot.sh for how this can be done. */
-#  define debug_hook(identity) Print(identity L"@0x%lx,0x%lx\n", POINTER_TO_PHYSICAL_ADDRESS(&_text), POINTER_TO_PHYSICAL_ADDRESS(&_data))
+#  define debug_hook(identity) printf(identity "@%p,%p\n", &_text, &_data)
 #else
 #  define debug_hook(identity)
 #endif


yuwata:
```
../src/boot/efi/util.c: In function ‘log_wait’:
/home/runner/work/systemd/systemd/src/fundamental/macro-fundamental.h:170:31: error: comparison of integer expressions of different signedness: ‘unsigned int’ and ‘int’ [-Werror=sign-compare]
  170 |                 UNIQ_T(A, aq) < UNIQ_T(B, bq) ? UNIQ_T(A, aq) : UNIQ_T(B, bq); \
      |                               ^
/home/runner/work/systemd/systemd/src/fundamental/macro-fundamental.h:165:19: note: in expansion of macro ‘__MIN’
  165 | #define MIN(a, b) __MIN(UNIQ, (a), UNIQ, (b))
      |                   ^~~~~
../src/boot/efi/util.c:419:19: note: in expansion of macro ‘MIN’
  419 |         BS->Stall(MIN(2 * log_count, 8) * 1000 * 1000);
      |                   ^~~
```

src/boot/efi/util.c
```diff
@@ -411,6 +411,15 @@ EFI_STATUS file_read(EFI_FILE *dir, const CHAR16 *name, UINTN off, UINTN size, C
         return err;
 }
 
+static unsigned log_count = 0;
+void log_wait(void) {
+        if (log_count == 0)
+                return;
+
+        BS->Stall(MIN(2 * log_count, 8) * 1000 * 1000);
```
yuwata: `8` -> `8u`??

yuwata:
semaphoreci
```
../src/boot/efi/efi-string.c: In function ‘format_status’: 01:14
../src/boot/efi/efi-string.c:367:14: error: ‘EFI_WARN_UNKNOWN_GLYPH’ undeclared (first use in this function); did you mean ‘EFI_WARN_UNKOWN_GLYPH’? 01:14
  367 |         case EFI_WARN_UNKNOWN_GLYPH:    s = u"Unknown glyph"; break; 01:14
      |              ^~~~~~~~~~~~~~~~~~~~~~ 01:14
      |              EFI_WARN_UNKOWN_GLYPH 01:14
../src/boot/efi/efi-string.c:367:14: note: each undeclared identifier is reported only once for each function it appears in
```

medhefgo:
Just FYI, github sends out notifications for CI failures on its own.

yuwata:
> Just FYI, github sends out notifications for CI failures on its own.

OK. Sorry for the noise.

medhefgo:
Some of the mkosi cis timeout out because sd-boot was failing. What concerns me is that this isn't detected as a ci failure.

@mrc0mmand Maybe you have some thoughts on that?

mrc0mmand:
> Some of the mkosi cis timeout out because sd-boot was failing. What concerns me is that this isn't detected as a ci failure.
> 
> @mrc0mmand Maybe you have some thoughts on that?

Eventually the job gets cancelled and the run with at least one cancelled job is marked as failed. However, the default timeout (360 minutes) is unnecessarily generous (imo), so maybe making it a bit shorter might help, e.g.:

```
diff --git a/.github/workflows/mkosi.yml b/.github/workflows/mkosi.yml
index 7061631c01..b0e120b3d3 100644
--- a/.github/workflows/mkosi.yml
+++ b/.github/workflows/mkosi.yml
@@ -25,6 +25,7 @@ env:
 jobs:
   ci:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     concurrency:
       group: ${{ github.workflow }}-${{ matrix.distro }}-${{ matrix.release }}-${{ github.ref }}
       cancel-in-progress: true
```

See: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
 DEFINE_PARSE_NUMBER(char, parse_number8);
 DEFINE_PARSE_NUMBER(char16_t, parse_number16);
 
+static void format_number(
+                char16_t buf[static SCRATCH_BUF_SIZE],
+                size_t min_digits,
+                bool hex,
+                bool hex_prefix,
+                bool lower,
+                bool is_signed,
```
yuwata: We usually introduce and use a enum if a function takes multiple boolean flags.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                uint64_t u) {
+
+        size_t n = 0;
+        int base = hex ? 16 : 10;
+        char16_t tmp[SCRATCH_BUF_SIZE];
+        const char16_t * const digits = lower ? u"0123456789abcdef" : u"0123456789ABCDEF";
```
yuwata: Why the second `const`?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+        if (is_signed && (int64_t) u < 0) {
+                buf[0] = '-';
+                buf++;
+
+                int rem = -((int64_t) u % base);
+                u = (int64_t) u / -base;
+                tmp[n++] = digits[rem];
```
yuwata: Why not simply `u = -(int64_t) u;`??

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        }
+
+        while (u > 0 || n == 0) {
+                int rem = u % base;
+                u /= base;
+                tmp[n++] = digits[rem];
+        }
+
+        assert(n + min_digits + 5 < SCRATCH_BUF_SIZE);
```
yuwata: Why `+5`??

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+        if (hex_prefix) {
+                buf[0] = '0';
+                buf[1] = lower ? 'x' : 'X';
+                buf += 2;
```
yuwata:
```
*buf++ = '0';
*buf++ = lower ? 'x' : 'X';
```

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        }
+
+        while (min_digits > n) {
+                buf[0] = '0';
+                buf++;
```
yuwata: `*buf++ = '0';`

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                min_digits--;
+        }
+
+        while (n > 0) {
+                *buf = tmp[--n];
```
yuwata: `*buf++ = tmp[--n];`

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+                default:
+                        assert_not_reached();
+                }
+}
+
+/* efi_vsnprintf is largely compatible to userspace vsnprintf. Any features omitted should trigger asserts.
+ * Notable differences:
+ *  - Passing NULL to %s is permitted and will print "(nil)".
+ *  - If buf is too small, -1 is returned instead of needed buf size (printf/xasprintf will do a allocation
```
yuwata: "a allocation" -> "an allocation"

Maybe, "will do _for_ an allocation loop"??

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+                case '*':
+                case '1' ... '9': {
+                        uint64_t u;
+
+                        if (*format == '*')
+                                u = va_arg(*ap, int);
+                        else {
+                                if (!parse_number8(format, &u, &format) || u > INT_MAX)
+                                        assert_not_reached();
```
yuwata: Even if the format is checked by the compiler with `_printf_` attribute, I'd like to return the function with `-EINVAL` and entirely refuse to formatting the remaining string.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                                *ret_precision = u;
+                        else if (!have_field_width && *ret_field_width < 0)
+                                *ret_field_width = u;
+                        else
+                                assert_not_reached();
+                        break;
+                }
+
+                case '%':
+                        *ret_str16 = u"%";
```
yuwata: What happens when field width or precision is specified? E.g. `%9.9%`??

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                        return format;
+
+                case 'm':
+                        format_number(tmp,
+                                      6,
+                                      /* hex= */ true,
+                                      /* has_prefix= */ true,
+                                      /* lower= */ true,
+                                      /* is_signed= */ false,
+                                      status);
```
yuwata: Printing status as a number??

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                        else
+                                *ret_str8 = va_arg(*ap, const char *);
+
+                        return format;
+
+                case 'm':
+                        format_number(tmp,
+                                      6,
+                                      /* hex= */ true,
+                                      /* has_prefix= */ true,
```
yuwata: has_prefix -> hex_prefix?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                }
+
+                case 'p': {
+                        const void *ptr = va_arg(*ap, const void *);
+
+                        if (ptr) {
+                                format_number(tmp,
+                                              /* min_digits= */ 8,
+                                              /* hex= */ true,
+                                              /* has_prefix= */ true,
```
yuwata: Also here.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                }
+
+                default:
+                        assert_not_reached();
+                }
+}
+
+/* efi_vsnprintf is largely compatible to userspace vsnprintf. Any features omitted should trigger asserts.
+ * Notable differences:
+ *  - Passing NULL to %s is permitted and will print "(nil)".
```
yuwata: I prefer "(null)" instead of "(nil)", as we use `strnull()` at many places.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                va_list *ap,
+                char16_t tmp[SCRATCH_BUF_SIZE],
+                bool *ret_pad_zero,
+                int *ret_precision,
+                int *ret_field_width,
+                const char **ret_str8,
+                const char16_t **ret_str16,
+                const wchar_t **ret_wstr) {
+
+        assert(ap);
```
yuwata: Please add comments what kind of specifier is supported.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,419 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
 DEFINE_PARSE_NUMBER(char, parse_number8);
 DEFINE_PARSE_NUMBER(char16_t, parse_number16);
 
+static void format_number(
```
yuwata: Maybe, it is better to introduce tests for this function.

yuwata (changes requested):
Reviewed only the first commit. Please rebase.

src/boot/efi/efi-string.c
medhefgo: `INT64_MIN * -1` is an int overflow. I should probably have add a comment to that fact.

src/boot/efi/efi-string.c
medhefgo: This function is tested already with the printf tests. I originally wrote test cases for this, but they ended up being identical to the printf ones.

yuwata: Right. But, even so, still I'd like to have tests for the function, though that's not strong requirement.

src/boot/efi/efi-string.c
yuwata: Interesting.

src/boot/efi/efi-string.c
medhefgo: gcc will complain about any of these: `%9.9% %9.% %0% %.0% %#%`

src/boot/efi/efi-string.c
medhefgo: I would argue that any invalid format string goes to the same category as assert on internal function parameters. If we are fine with "don't call this function with a NULL pointer", it's only logical to apply the same for "don't call this with invalid/unsupported format string". All these are internal functions and aren't supposed to be exposed to outside users, so imho asserting should be safe and keeps things simpler.

Assert also makes it much more likely that an invalid/unsupported format does not get unnoticed instead of being silently swallowed somewhere.

medhefgo:
I am hitting a really weird bug in some of the mkosi distros somewhere that causes the kernel to fail at some point during early boot.

The interesting part about this is that it seems to be caused by code size: it will break even when the printf functions aren't called from code whatsoever. The easiest way to trigger this was by changing the format_status lookup table. At some point I reduced code size to a single extra char in one of the strings. Interestingly, I cannot seem to reproduce this by only adding a very large string table onto the main branch (so without any of the printf functions defined).
Also note that despite the kernel breaking, the printf functions work perfectly fine.

In a dev branch where I have dropped libefi.a, this also doesn't happen. So my best guess is either some OVMF bug or the way the final binary is assembled (using gnu-efi).

Originally I simply decided to ignore this by dropping format_status until libefi.a is removed, but I found a simple workaround: building with at least -O1 seems to fix this (I even added a *giant* list of strings to the table just to make sure).

I have now looked at the functions for way longer than I like and cannot see anything that would cause this. I am also confused as to why other (mkosi) distros work fine. It doesn't seem to be because of distro-patched gnu-efi as at least debian seems to be using vanilla upstream.

I think using `-O1` until libefi.a is dropped is the way to go. I also reduced the size of this PR like I originally did to make it more palatable as it shouldn't be necessary to "rush" dropping libefi.a anymore.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,500 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
 DEFINE_PARSE_NUMBER(char, parse_number8);
 DEFINE_PARSE_NUMBER(char16_t, parse_number16);
 
+typedef enum NumberFormatFlags {
+        NUMBER_FORMAT_DEFAULT    = 1 << 0,
+        NUMBER_FORMAT_HEX        = 1 << 1,
+        NUMBER_FORMAT_HEX_PREFIX = 1 << 2,
+        NUMBER_FORMAT_LOWER      = 1 << 3,
+        NUMBER_FORMAT_SIGNED     = 1 << 4,
```
poettering: Can you add brief comments to the righ there what these flags do?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,500 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                char16_t buf[static SCRATCH_BUF_SIZE],
+                size_t min_digits,
+                NumberFormatFlags flags,
+                uint64_t u) {
+
+        size_t n = 0;
+        int base = FLAGS_SET(flags, NUMBER_FORMAT_HEX) ? 16 : 10;
+        char16_t tmp[SCRATCH_BUF_SIZE];
+        const char16_t *digits = FLAGS_SET(flags, NUMBER_FORMAT_LOWER) ? u"0123456789abcdef" :
+                                                                         u"0123456789ABCDEF";
```
poettering: we have LOWERCASE_HEXDIGITS in src/basic/string-util.h already. Maybe move that to fundamental and add UPPERCASE_HEXDIGITS?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,500 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+        if (FLAGS_SET(flags, NUMBER_FORMAT_SIGNED) && (int64_t) u < 0) {
+                *buf++ = '-';
```
poettering: at least I personally do not know precedence of operations between `*` and `++` I must say, and whenever I see this I must think twice.

Maybe do: `*(buf++)` just to make this more obvious, even though it's not necessary.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,500 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+                /* We cannot just do "u = -(int64_t)u" here because -INT64_MIN overflows. */
+
+                int rem = -((int64_t) u % base);
+                u = (int64_t) u / -base;
+                tmp[n++] = digits[rem];
+        }
+
+        while (u > 0 || n == 0) {
+                int rem = u % base;
```
poettering: while this should fit in an `int`, given that `base` is 16 or 10, I'd still make this a uint64_t, because a division/modulo shouldn't really change the type of `u`

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,500 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+static const char16_t * const warn_table[] = {
+        u"Success",
+        u"Unknown glyph",
+        u"Delete failure",
+        u"Write failure",
+        u"Buffer too small",
+        u"Stale data",
+        u"File system",
+        u"Reset required",
```
poettering: why not use a `[…] =` table, like we do everywhere else?

poettering: (I mean, I think I understand why, but still, please just define the EFI errors also if you compile this in userspace via `#ifndef`, it's not that many)

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,500 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                strcpy16(tmp, warn_table[status]);
+                return;
+        }
+
+        if (status >= EFI_ERROR_MIN && status <= EFI_ERROR_MAX) {
+                strcpy16(tmp, err_table[status - EFI_ERROR_MIN]);
+                return;
+        }
+
+        format_number(tmp, 6, NUMBER_FORMAT_POINTER, status);
```
poettering: where does the 6 come from? add comment?

src/boot/efi/assert.c
```diff
@@ -6,7 +6,11 @@
 #include "util.h"
 
 void efi_assert(const char *expr, const char *file, unsigned line, const char *function) {
-        log_error_stall(L"systemd-boot assertion '%a' failed at %a:%u, function %a(). Halting.", expr, file, line, function);
+        log_error("systemd-boot assertion '%s' failed at %s:%u, function %s(). Halting.",
```
poettering: actually "systemd-boot" is a misleading here, might also be in "systemd-stub"...

poettering (changes requested):
looks pretty good to me. but i am of course not entirely sure i grapsed all details about the format string logic, but it does look ok, so let's hope the tests found the corner cases

src/boot/efi/efi-string.c
medhefgo: Imho, it's really "Not that many". I've changed it so that one of each table is provided for unit tests, though.

src/boot/efi/assert.c
medhefgo: This is an assert message, and changing this is too much work for little gain. Nobody is expected to see this. The important thing is the mention of systemd so that the right people get blamed. :D

medhefgo:
I've simplified things quite a bit by moving the allocation loop into the actual printf functions. I also dropped snprintf as there's hardly any users for it.

I also decided to keep the differences to std printf to a minimum, so there will be no direct support for device paths, guids and optional printf output as I originally planned/did.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
 DEFINE_PARSE_NUMBER(char, parse_number8);
 DEFINE_PARSE_NUMBER(char16_t, parse_number16);
 
+enum {
+        SCRATCH_BUF_SIZE = 64,
+};
```
poettering: btw, I am not sure I agree with your love for enums even for stuff that is not actually enum-like or flags.

The thing is that in particuar for limits you typically want unsigned values actually, but enums are never unsigned. i.e. I think:

```c
#define SCRATCH_BUF_SIZE 64U
```

is much more appropriate actually.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        const char *digits = FLAGS_SET(flags, NUMBER_LOWER) ? "0123456789abcdef" : "0123456789ABCDEF";
+
+        if (FLAGS_SET(flags, NUMBER_SIGNED) && (int64_t) u < 0) {
+                *(buf++) = '-';
+
+                /* We cannot just do "u = -(int64_t)u" here because -INT64_MIN overflows. */
+
+                uint64_t rem = -((int64_t) u % base);
+                u = (int64_t) u / -base;
+                tmp[n++] = digits[rem];
```
poettering: (in userspace we usually use hexchar() for this)

src/boot/efi/efi-string.c
medhefgo: Don't bother looking at this code for now. I already majorly reworked this code to make it much nicer to look at. I am just doing some finishing touches right now.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                size_t min_digits,
+                NumberFormatFlags flags,
+                uint64_t u) {
+
+        size_t n = 0;
+        int base = FLAGS_SET(flags, NUMBER_HEX) ? 16 : 10;
+        char16_t tmp[SCRATCH_BUF_SIZE];
+        const char *digits = FLAGS_SET(flags, NUMBER_LOWER) ? "0123456789abcdef" : "0123456789ABCDEF";
+
+        if (FLAGS_SET(flags, NUMBER_SIGNED) && (int64_t) u < 0) {
```
poettering: are you sure this cast is safe? isn't this one of these areas where the c compiler can do whatever it wants if u > INT64_MAX because undef behaviour or so?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
 DEFINE_PARSE_NUMBER(char, parse_number8);
 DEFINE_PARSE_NUMBER(char16_t, parse_number16);
 
+enum {
+        SCRATCH_BUF_SIZE = 64,
```
poettering: can we determine how much precisely we need at max here? i.e. largest 64bit number in octal with most decoration we can have?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        assert(ret_wstr);
+
+        bool alternative_form = false, have_field_width = false, long_arg = false;
+        size_t field_length = sizeof(int);
+        int precision = -1, field_width = -1;
+
+        for (;; format++)
+                switch (*format) {
+                case '0':
+                        *ret_pad_zero = true;
```
poettering: we have this coding style rule that on success we initialize all return params, and on failure none. afaics this might be set or might be left alone on success so this is conflicting with this rule.

You might be able to get away by simply dropping the "ret_" prefix, i.e. by stopping to claim that this is a return paremeter, but instead an input/output param.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                case '*':
+                case '1' ... '9': {
+                        int i;
+
+                        if (*format == '*')
+                                i = va_arg(*ap, int);
+                        else {
+                                uint64_t u;
+                                if (!parse_number8(format, &u, &format) || u > INT_MAX)
+                                        assert_not_reached();
```
poettering: instead of:

```c
if (!x)
        asser_not_reached()
```

write:

```c
assert_se(x);
```

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                                field_width = i;
+                        break;
+                }
+
+                case '%':
+                        *ret_str16 = u"%";
+                        return format;
+
+                case 'c':
+                        /* char/char16_t/wchat_t get promoted to int/wint_t. */
```
poettering: wchat is a typo

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                        if (status <= ELEMENTSOF(warn_table) - 1) {
+                                *ret_str8 = warn_table[status];
+                                return format;
+                        }
+
+                        if (status >= EFI_ERROR_MASK &&
+                            status <= ((ELEMENTSOF(err_table) - 1) | EFI_ERROR_MASK)) {
+                                *ret_str8 = err_table[NOERR(status)];
+                                return format;
+                        }
```
poettering: i'd really add a helper call for this that takes the status as input and returns a const char * if it finds it in either of the two tables

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
 
+enum {
+        SCRATCH_BUF_SIZE = 64,
+};
+
+typedef enum FormatFlags {
+        NUMBER_DEFAULT    = 1 << 0, /* Print as decimal. */
+        NUMBER_HEX        = 1 << 1, /* Print as hex. */
+        NUMBER_HEX_PREFIX = 1 << 2, /* Print as hex with 0x prefix. */
+        NUMBER_LOWER      = 1 << 3, /* Print hex numbers and prefix in lowercase letters. */
```
poettering: NUMBER_HEX_LOWER?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                        SET_FLAG(flags, NUMBER_LOWER, IN_SET(*format, 'x'));
+
+                        switch (field_length) {
+                        case sizeof(int32_t):
+                                v = FLAGS_SET(flags, NUMBER_SIGNED) ? (uint64_t) va_arg(*ap, int32_t) :
+                                                                      va_arg(*ap, uint32_t);
+                                break;
+                        case sizeof(int64_t):
+                                v = FLAGS_SET(flags, NUMBER_SIGNED) ? va_arg(*ap, int64_t) :
+                                                                      va_arg(*ap, uint64_t);
```
poettering: hmm, compilers don't warn that you return different types in both branches of the ternary op? weird...

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                }
+
+               /* Grow buf. */
+                if (n + len + padding >= n_buf) {
+                        n_buf = 2 * (n + len + padding);
+
+                        char16_t *new_buf = xnew(char16_t, n_buf);
+                        memcpy(new_buf, buf, n * sizeof(*buf));
+
+                        free(dyn_buf);
```
poettering: xrealloc()?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,453 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+                for (size_t i = 0; i < len; i++)
+                        buf[n++] = str16 ? str16[i] : (str8 ? str8[i] : wstr[i]);
+        }
+
+        buf[n++] = '\0';
+
+        if (ret_buf) {
+                if (buf == stack_buf) {
+                        *ret_buf = xnew(char16_t, n);
+                        memcpy(*ret_buf, buf, n * sizeof(*buf));
```
poettering: in userspace we have memdup() and newdup(), maybe add that here too?

src/boot/efi/efi-string.c
medhefgo: This can never happen at runtime, as this is the passed inside of format string branch and will throw a compiler error:
`error: ‘%2147483648i’ directive output of 2147483648 bytes exceeds ‘INT_MAX’`

src/boot/efi/efi-string.c
medhefgo: I think int conversion happens first.

src/boot/efi/efi-string.c
medhefgo: That would free `stack_buf` for the first time the buffer is too small.

src/boot/efi/efi-string.c
medhefgo: Yeah, it's implementation defined. And for gcc it's two's complement: https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html#Integers-implementation
So this should be fine and testcases ought to be covering this anyways.

Also, it's interesting to know gcc only supports two's complement integers.

medhefgo:
I've restructured the code. It should be easier to grok now that there's less indentation and code that deals with adding to the buf moved into helper functions.

During all this I've found a ton of printf corner case behaviors that are all dealt with now (they should all be part of the unit tests). I even added support for more flags.

I am fairly confident about this implementation now.

keszybz:
Needs a rebase now (prolly because of the #24632 that was just merged).
Do we need to merge this before v252? It seems that there's some risk, even if it's a matter of issues somewhere in the toolchain. If possible, I'd prefer to merge this early in the next cycle.

medhefgo:
This can certainly wait for next cycle.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,495 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+#ifdef SD_BOOT
+        [EFI_WARN_UNKOWN_GLYPH]     = "Unknown glyph",
+        [EFI_WARN_DELETE_FAILURE]   = "Delete failure",
+        [EFI_WARN_WRITE_FAILURE]    = "Write failure",
+        [EFI_WARN_BUFFER_TOO_SMALL] = "Buffer too small",
+#  ifdef EFI_WARN_RESET_REQUIRED
+        [EFI_WARN_STALE_DATA]       = "Stale data",
+        [EFI_WARN_FILE_SYSTEM]      = "File system",
+        [EFI_WARN_RESET_REQUIRED]   = "Reset required",
+#  endif
```
keszybz: Does it make sense to idef this? Even if not used, it's just a few dozens of bytes…

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,495 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        [NOERR(EFI_END_OF_MEDIA)]         = "End of media",
+        [29]                              = "Reserved (29)",
+        [30]                              = "Reserved (30)",
+        [NOERR(EFI_END_OF_FILE)]          = "End of file",
+        [NOERR(EFI_INVALID_LANGUAGE)]     = "Invalid language",
+        [NOERR(EFI_COMPROMISED_DATA)]     = "Compromised data",
+#  ifdef EFI_HTTP_ERROR
+        [NOERR(EFI_IP_ADDRESS_CONFLICT)]  = "IP address conflict",
+        [NOERR(EFI_HTTP_ERROR)]           = "HTTP error",
+#  endif
```
keszybz: … here too.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,495 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+
+        wchar_t c;
+        const char *str;
+        const wchar_t *wstr;
+
+        /* For numbers. */
+        bool is_signed:1;
+        bool lowercase:1;
+        int8_t base;
+        char sign_pad; /* For + and (space) flags. */
```
keszybz: Hmm, I'm pretty sure that that the bitfields just make the code larger and slower. Bitfields make sense when there's many copies of some structure, but I don't think there'll ever be more than a few of this one… So it's probably better to drop the bitfield annotations and spend a few bytes on extra memory and reduce code size.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,495 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+} SpecifierContext;
+
+typedef struct {
+        char16_t stack_buf[128]; /* We use stack_buf first to avoid allocations in most cases. */
```
keszybz, line 469: 128 → 1024? It's still very small, but will cover more cases.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,495 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+static void grow_buf(FormatContext *ctx, size_t need) {
+        assert(ctx);
+
+        if (ctx->n + need < ctx->n_buf)
+                return;
+
+        ctx->n_buf = 2 * (ctx->n + need);
+
+        char16_t *new_buf = xnew(char16_t, ctx->n_buf);
+        memcpy(new_buf, ctx->buf, ctx->n * sizeof(*ctx->buf));
```
keszybz, line 494: Not that it matters too much, but why not use `realloc`?

src/boot/efi/secure-boot.c
```diff
@@ -109,7 +109,7 @@ EFI_STATUS secure_boot_enroll_at(EFI_FILE *root_dir, const char16_t *path) {
 
                 err = efivar_set_raw(&sb_vars[i].vendor, sb_vars[i].name, sb_vars[i].buffer, sb_vars[i].size, sb_vars_opts);
                 if (err != EFI_SUCCESS) {
-                        log_error_stall(L"Failed to write %s secure boot variable: %r", sb_vars[i].name, err);
+                        log_error("Failed to write %ls secure boot variable: %m", sb_vars[i].name);
```
keszybz: Hmm, `%m` should only be used with `log_*_errno`. `err` seems to have been lost.

src/boot/efi/secure-boot.c
```diff
@@ -95,7 +95,7 @@ EFI_STATUS secure_boot_enroll_at(EFI_FILE *root_dir, const char16_t *path) {
         for (size_t i = 0; i < ELEMENTSOF(sb_vars); i++) {
                 err = file_read(dir, sb_vars[i].filename, 0, 0, &sb_vars[i].buffer, &sb_vars[i].size);
                 if (err != EFI_SUCCESS) {
-                        log_error_stall(L"Failed reading file %s\\%s: %r", path, sb_vars[i].filename, err);
+                        log_error("Failed reading file %ls\\%ls: %m", path, sb_vars[i].filename);
```
keszybz: Here too.

src/boot/efi/boot.c
```diff
@@ -481,110 +468,140 @@ static void print_status(Config *config, char16_t *loaded_image_path) {
...
-                ps_string(L"      sort key: %s\n", entry->sort_key);
-                ps_string(L"       version: %s\n", entry->version);
-                ps_string(L"    machine-id: %s\n", entry->machine_id);
-                if (entry->device)
-                    Print(L"        device: %D\n", DevicePathFromHandle(entry->device));
-                ps_string(L"        loader: %s\n", entry->loader);
+                printf("  config entry: %zu/%zu\n", i + 1, config->entry_count);
+                printf("            id: %ls\n", entry->id);
+                if (entry->title)
+                        printf("         title: %ls\n", entry->title);
```
keszybz, line 581: This changes the logic to only print certain lines conditionally. Please add a comment why in the commit message.

src/boot/efi/boot.c
```diff
@@ -2031,10 +2048,9 @@ static EFI_STATUS boot_windows_bitlocker(void) {
 
         for (UINTN i = 0; i < boot_order_size / sizeof(uint16_t); i++) {
                 _cleanup_free_ char *buf = NULL;
-                char16_t name[sizeof(L"Boot0000")];
                 UINTN buf_size;
 
-                SPrint(name, sizeof(name), L"Boot%04x", (uint32_t) boot_order[i]);
+                _cleanup_free_ char16_t *name = xasprintf("Boot%04x", boot_order[i]);
```
keszybz, line 2031: Looks like this was supposed to be part of preceding commit.

keszybz (commented):
I didn't go into the details of the printf implementation, I was only mostly reviewing the usage of the new functions. This all looks pretty clean, apart from a few minor issues.

EDIT: I was reviewing the version before the last push. So some comments might be out-of-date.

src/boot/efi/efi-string.c
medhefgo: gnu-efi doesn't provide these right now and I am too lazy to add them to missing_efi.h

src/boot/efi/efi-string.c
medhefgo: Turns out that the struct is the same size regardless. Silly padding rules. :D

src/boot/efi/efi-string.c
medhefgo: This goes on the stack (and because it's char16_t it's double the bytes). Since this is efispace I'd rather be on the conservative side here. In practice it's only around one such allocation during a regular boot.

src/boot/efi/efi-string.c
medhefgo: This could be stack_buf, which we shall remain forever imprisoned on stack. I've added a comment to make this clear.

src/boot/efi/secure-boot.c
medhefgo: And here I thought I've stared enough times at these conversion to think I've fixed them all. :D

src/boot/efi/efi-string.c
keszybz: Yeah, `stack_buf` would need to be handled explicitly.
```C
if (ctx->dyn_buf)
    realloc(ctx->dyn_buf, ctx->n * sizeof(*ctx->buf));
else {
    new_buf = xnew(char16_t, ctx->n_buf);
    memcpy(new_buf, ctx->buf, ctx->n * sizeof(*ctx->buf));
    free(ctx->dyn_buf);
}
```
After writing this out, I see that it's annoyingly longer. Let's keep the current version.

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,496 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        free(ctx->dyn_buf);
+        ctx->buf = ctx->dyn_buf = new_buf;
+}
+
+static void push_padding(FormatContext *ctx, char pad, size_t len) {
+        assert(ctx);
+        while (len > 0) {
+                len--;
+                ctx->buf[ctx->n++] = pad;
+        }
```
keszybz, line 505: Maybe add an assert that `ctx->n + len < ctx->n_buf` ?

src/boot/efi/efi-string.c
```diff
@@ -272,6 +274,496 @@ bool efi_fnmatch(const char16_t *pattern, const char16_t *haystack) {
...
+        int8_t base;
+        char sign_pad; /* For + and (space) flags. */
+} SpecifierContext;
+
+typedef struct {
+        char16_t stack_buf[128]; /* We use stack_buf first to avoid allocations in most cases. */
+        char16_t *dyn_buf;       /* Allocated buf or NULL if stack_buf is used. */
+        char16_t *buf;           /* Points to the current active buf. */
+        size_t n_buf;
+        size_t n;
```
keszybz: Please also add comments what those are, incl. allowed ranges / relationship between `n` and `n_buf`.

keszybz (commented):
LGTM.

We'll certainly want to create a fuzzer for the printf stuff. I can do it after this is merged, but if somebody else wants to, please go ahead.

yuwata:
needs to be rebased.

BTW, As I commented few weeks (months?) ago, this should be definitely tested with a fuzzer, though I do not block this PR without it, and it's ok to add that later. But, please please add a fuzzer for this feature.

src/boot/efi/efi-string.c
medhefgo: This should really be covered by the `assert(ctx.n < ctx.n_buf);` in `push_num/push_str` and the final one in `printf_internal`. (There is imho only so much paranoia before it gets silly.)

medhefgo:
How do you even fuzz printf? Random format strings itself would create tons of UB pitfalls thanks to the insanity that is vararg functions…

bluca:
needs a rebase

medhefgo:
I've added a fuzzer. Not sure how good that one is, though.

keszybz:
CI looks unhappy:

```
Run sudo timeout -k 30 10m mkosi qemu
Couldn't find swtpm binary, not invoking qemu with TPM2 device.
‣ Running command:
‣ qemu-system-x86_64 -machine type=q35,accel=tcg,smm=on -smp 1 -m 2G -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0,id=rng-device0 -cpu max -nographic -nodefaults -serial mon:stdio -nic tap,script=no,downscript=no,ifname=vt-image,model=virtio-net-pci -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE.secboot.fd -global ICH9-LPC.disable_s3=1 -global driver=cfi.pflash01,property=secure,value=on -drive file=/var/tmp/.mkosi-0cazehvu,if=pflash,format=raw -drive 'if=none,id=hd,file=/home/runner/work/systemd/systemd/mkosi.output/centos_epel~8-stream/image.raw,format=raw' -device virtio-scsi-pci,id=scsi -device scsi-hd,drive=hd,bootindex=1

3hBdsDxe: loading Boot0001 "UEFI QEMU QEMU HARDDISK " from PciRoot(0x0)/Pci(0x3,0x0)/Scsi(0x0,0x0)
BdsDxe: starting Boot0001 "UEFI QEMU QEMU HARDDISK " from PciRoot(0x0)/Pci(0x3,0x0)/Scsi(0x0,0x0)
systemd-boot@0x7d356000,0x7d371000
!!!! X64 Exception Type - 0D(#GP - General Protection)  CPU Apic ID - 00000000 !!!!
ExceptionData - 0000000000000000
RIP  - 00000000709D43FC, CS  - 0000000000000038, RFLAGS - 0000000000000206
RAX  - 000A226DD2013274, RCX - 0000000000001380, RDX - 00000000709DC168
RBX  - 000000007DAB3B98, RSP - 000000007EFB4428, RBP - 000000007EFB47E0
RSI  - 00000000709DB0B8, RDI - 00000000709BC000
R8   - 0000000000000018, R9  - 0000000000000000, R10 - 000000007DAD1098
R11  - 000000007DC1B65C, R12 - 0000000000000000, R13 - 0000000000000000
R14  - 000000007EFB4568, R15 - 000000007DB1C298
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010033, CR2 - 0000000000000000, CR3 - 000000007EC01000
CR4  - 0000000000000668, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 0000000000000400
GDTR - 000000007E9E3000 0000000000000047, LDTR - 0000000000000000
IDTR - 000000007E4E3018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 000000007EFB4080
!!!! Find image based on IP(0x709D43FC) (No PDB)  (ImageBase=00000000709BC000, EntryPoint=00000000709C1000) !!!!
qemu-system-x86_64: terminating on signal 15 from pid 48414 (timeout)
Error: Process completed with exit code 124.
```

keszybz:
→ https://github.com/systemd/systemd/pull/25766

medhefgo:
It's just another case of overlapping PE sections (which in this case are not detected by the stub as code execution doesn't get far enough). The proper solution is porting mkosi to use ukify. Alternatively, one could use [this branch](https://github.com/medhefgo/systemd/tree/efi-clang-next), which does the little extra work to drop libefi.a, which should reduce the size pressure.

keszybz:
From the other PR:
```
 systemd-boot@0x7D0FB000,0x7D117000
systemd-stub@0x72349000,0x7235E000
Overlapping PE sections detected. Boot may fail due to image memory corruption!
systemd-boot assertion 'description' failed at ../src/src/boot/efi/measure.c:149, function tpm_log_event(). Halting.
```

Took me a while to see the issue:
```
Idx Name          Size      VMA               LMA               File off  Algn
  0 .text         00013aa0  0000000000005000  0000000000005000  00000370  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  1 .reloc        0000000a  0000000000019000  0000000000019000  00013f70  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  2 .data         000051a8  000000000001a000  000000000001a000  00014170  2**4
                  CONTENTS, ALLOC, LOAD, DATA
  3 .dynamic      00000100  0000000000020000  0000000000020000  00019370  2**2
                  CONTENTS, ALLOC, LOAD, DATA
  4 .osrel        0000029c  0000000000020000  0000000000020000  00019570  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  5 .rela         000014e8  0000000000021000  0000000000021000  00019970  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  6 .dynsym       00000018  0000000000023000  0000000000023000  0001af70  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  7 .sbat         000000d5  0000000000025980  0000000000025980  0001b170  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  8 .sdmagic      00000027  0000000000025a60  0000000000025a60  0001b370  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  9 .cmdline      00000032  0000000000030000  0000000000030000  0001b570  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
 10 .linux        00c285e8  0000000002000000  0000000002000000  0001b770  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
 11 .initrd       038a76ee  0000000003000000  0000000003000000  00c43d70  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
```
Both `.dynamic` and `.osrel` are allocated at the same VMA. This cannot work ;(

DaanDeMeyer:
@medhefgo Can you rebase this one? Should get rid of the mkosi CI failures

DaanDeMeyer:
@medhefgo needs another rebase it seems

medhefgo:
> @medhefgo needs another rebase it seems

I was just doing that when I saw https://github.com/systemd/systemd/pull/26087. This branch is just too damn easy to conflict with. :(

src/boot/efi/stub.c
```diff
@@ -421,3 +416,15 @@ EFI_STATUS efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *sys_table) {
...
+        InitializeLib(image, sys_table);
+
+        debug_hook(L"systemd-stub");
+        /* Uncomment the next line if you need to wait for debugger. */
+        // debug_break();
+
+        EFI_STATUS err = real_main(image);
+        log_wait();
+        return err;
+}
```
poettering, line 430: this reminds me a lot of what we have in userspace with DEFINE_MAIN_FUNCTION(). I think we might want to do something similar here sooner or later?

(but i guess that can be done another time)

src/fundamental/macro-fundamental.h
```diff
@@ -74,11 +74,11 @@
...
                 #define assert_not_reached() __builtin_unreachable()
         #else
-                #define assert(expr) ({ _likely_(expr) ? VOID_0 : efi_assert(#expr, __FILE__, __LINE__, __PRETTY_FUNCTION__); })
-                #define assert_not_reached() efi_assert("Code should not be reached", __FILE__, __LINE__, __PRETTY_FUNCTION__)
+                #define assert(expr) ({ _likely_(expr) ? VOID_0 : efi_assert(#expr, __FILE__, __LINE__, __func__); })
+                #define assert_not_reached() efi_assert("Code should not be reached", __FILE__, __LINE__, __func__)
         #endif
         #define static_assert _Static_assert
-        #define assert_se(expr) ({ _likely_(expr) ? VOID_0 : efi_assert(#expr, __FILE__, __LINE__, __PRETTY_FUNCTION__); })
+        #define assert_se(expr) ({ _likely_(expr) ? VOID_0 : efi_assert(#expr, __FILE__, __LINE__, __func__); })
```
poettering, line 81: if you change this, please also change this for the other asserts in src/basic/macro.h

i.e. across the whole codebase please.

if i "git grep" through things right now I see most `__func__`, the asserts re the only holdouts.

poettering:
lgtm

src/boot/efi/stub.c
medhefgo: I'll think on it

medhefgo:
I converted the other `__PRETTY_FUNCTION__` too.

PR #23709, opened 2022-06-11
test: import hostname and locale tests from debian/ubuntu test suite

yuwata:


yuwata:
TODO: support debian patch: https://salsa.debian.org/systemd-team/systemd/-/blob/debian/experimental/debian/patches/debian/Use-Debian-specific-config-files.patch

yuwata:
Now this is ready. Please take a look!

test/test-functions
```diff
@@ -2609,6 +2671,20 @@ inst_any() {
     return 1
 }
 
+inst_recursive() {
+    local i p
+
+    for p in in "$@"; do
```
mrc0mmand:
```suggestion
    for p in "$@"; do
```

test/test-functions
```diff
@@ -2609,6 +2671,20 @@ inst_any() {
...
+        for i in $(find "$p" 2>/dev/null || :); do
```
mrc0mmand: This will fall apart with paths containing spaces, maybe use something like:

```bash
while read -r item; do
...
done < <(find "$p" 2>/dev/null)
# No need for || : here
```

test/units/assert.sh
```diff
@@ -3,42 +3,56 @@
...
+assert_true() {(
     local rc
 
-    set +e
+    set +ex
+
     "$@"
     rc=$?
-    set -e
     if [[ "$rc" != "0" ]]; then
```
mrc0mmand:
```suggestion
    if [[ $rc -ne 0 ]]; then
```

test/units/assert.sh
```diff
@@ -3,42 +3,56 @@
...
+
+    if [[ "$2" =~ $1 ]]; then
+        echo "FAIL: '$1' found in:" >&2
+        echo "$2" >&2
+        exit 1
+    fi
+)}
+
+assert_rc() {(
+    local rc exp=$1
```
mrc0mmand:
```suggestion
    local rc exp="${1:?}"
```

test/units/assert.sh
```diff
@@ -3,42 +3,56 @@
...
     fi
-}
+)}
+
 
+assert_eq() {(
+    set +ex
 
-assert_eq() {
     if [[ "$1" != "$2" ]]; then
```
mrc0mmand:
```suggestion
    if [[ "${1?}" != "${2?}" ]]; then
```
(No colons, since we want fail only on unset values, not empty ones)

test/units/assert.sh
```diff
@@ -3,42 +3,56 @@
...
     fi
-}
+)}
+
+assert_in() {(
+    set +ex
 
-assert_in() {
-    if ! echo "$2" | grep -q "$1"; then
+    if ! [[ "$2" =~ $1 ]]; then
```
mrc0mmand:
```suggestion
    if ! [[ "${2?}" =~ ${1?} ]]; then
```

test/units/assert.sh
```diff
@@ -3,42 +3,56 @@
...
         echo "$2" >&2
         exit 1
     fi
-}
+)}
+
+assert_not_in() {(
+    set +ex
+
+    if [[ "$2" =~ $1 ]]; then
```
mrc0mmand:
```suggestion
    if [[ "${2?}" =~ ${1?} ]]; then
```

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,404 @@
...
+        mv /tmp/default-keyboard.bak /etc/default/keyboard
+    else
+        rm -f /etc/default/keyboard
+        rmdir --ignore-fail-on-non-empty /etc/default
+    fi
+}
+
+wait_vconsole_setup() {
+    local i ss
+    for ((i=0;i<20;i++)); do
```
mrc0mmand:
```suggestion
    for ((i = 0; i < 20; i++)); do
```

test/units/testsuite-73.sh
```diff
@@ -0,0 +1,404 @@
...
+test_vc_keymap() {
+    local i output
+
+    if [[ -z "$(localectl list-keymaps)" ]]; then
+        echo "No vconsole keymap installed, skipping test."
+        return
+    fi
+
+    backup_keymap
+    trap restore_keymap EXIT
```
mrc0mmand: This could be, theoretically, simplified by using

```suggestion
    trap restore_keymap RETURN
```

and dropping all the

```bash
    restore_keymap
    trap - EXIT
```

statements below. Similarly in other functions.

mrc0mmand (changes requested):
Looks excellent, just a couple of nits, thanks for working on this!

test/units/assert.sh
yuwata: Should this with colon?

mrc0mmand: Ah, mea culpa, also without colon, I blame the muscle memory :-)

yuwata:
@mrc0mmand Thank you for the review. I hope all points are addressed. PTAL.

mrc0mmand (approved):
LGTM, thank you!

PR #23710, opened 2022-06-11
systemd multiseat more than one user on a graphic card

Gert-Vanhaerents:
Hi Lennart and/or others,
First of all, systemd is a very big improvement for the Linux community, much faster and better than the earlier systems.
I'm the IT person of a school, earlier we used multiseatcomputers for the school, i have maded with a XGL implementation and it works fine but not so fantastic. The school wants that i build new computers but the XGL project is for now too outdated so i can't use it anymore. 

I have tested with the systemd loginctl tool, it works fantastic, very fast, very stable and do almost all what we want. But i can use only seat on one graphic card. So i can make only multiseats with 3 users, and that's not enough. Is there a possibility to make the multiseats with more than one users on one graphic card? 

The only thing i see is that the graphic card not allow to make more than one seat on one card. But on windows multiseats systems is works perfectly with 2 or even 3 seats on one card and it's also not crashing. I'm sure what Windows can Linux can also!
How can i make a multiseatcomputer with more then one user on one card? 
How can i change or rebuild the loginctl the use more than one seat on one card? Is there any config file that I can change somewhere so that this restriction is no longer used?

Hopefully someone can help, the children and school will allways thank you!
Thanks in advance,
With Kind Regards,
Gert

### DIFF
--- a/src/network/networkctl.c
+++ b/src/network/networkctl.c
@@ -36,6 +36,7 @@
 #include "locale-util.h"
 #include "logs-show.h"
 #include "macro.h"
+#include "macvlan-util.h"
 #include "main-func.h"
 #include "netlink-util.h"
 #include "network-internal.h"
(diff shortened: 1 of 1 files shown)


yuwata:
Please open an issue page, rather than opening an irrelevant PR. Closing.

PR #23714, opened 2022-06-11
A couple of assorted tweaks for the systemd-networkd test suite

mrc0mmand:
Only the first commit is somewhat important, the rest are just clean ups based on pylint's suggestions.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -295,9 +295,6 @@ def expectedFailureIfFQPIEIsNotAvailable():
     return f
 
 def setUpModule():
-    # pylint: disable=global-statement
-    global running_units
-
     os.makedirs(network_unit_file_path, exist_ok=True)
     os.makedirs(networkd_conf_dropin_path, exist_ok=True)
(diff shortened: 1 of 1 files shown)


yuwata (approved):

mrc0mmand:
@yuwata @evverx just a quick heads-up - I finally finished https://github.com/systemd/systemd-centos-ci/pull/308, which means the CentOS CI runs the integration test suite (selected tests + the networkd test suite) on C8S ppc64le with ASan+UBSan. For now it's run only in a daily cron job (because the alt-arch machine pool is very limited), but it should, hopefully, be of some use. We have also aarch64 available as well, but that needs further work, since the machines are... strange.

yuwata:
@mrc0mmand Nice. Where can I see the result?

mrc0mmand:
> @mrc0mmand Nice. Where can I see the result?

Currently the latest ones can be found in the CI job results - https://jenkins-systemd.apps.ocp.ci.centos.org/job/ci-build/1053/artifact//index.html.

Once the cron job succeeds (looks like there was a network outage during today's night), it should be part of the cron job results here - https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-cron-build/.

PR #23722, opened 2022-06-12
po: add a false positive to POTFILES.skip

piotrdrag:


yuwata:
I do not understand the change. What issue do you want to fix? Could you elaborate more?

piotrdrag:
Scripts used to detect files that should be in POTFILES.in, like intltool-update -m used on https://l10n.gnome.org/module/systemd/, falsely detect this file as containing translations. This patch avoids this behavior by putting the file in POTFILES.skip.

yuwata (approved):
Thank you for the explanation. And sorry for late reply.

PR #23726, opened 2022-06-12
inotify-util: use address of raw instead of ev

sshedi:
Fixes:
CID#1487033
CID#1487023
CID#1487022
CID#1487021
CID#1487020
CID#1487019

### DIFF
--- a/src/basic/inotify-util.h
+++ b/src/basic/inotify-util.h
@@ -12,7 +12,7 @@
 
 #define _FOREACH_INOTIFY_EVENT(e, buffer, sz, log_level, start, end)    \
         for (struct inotify_event                                       \
-                     *start = &((buffer).ev),                           \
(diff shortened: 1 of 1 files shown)


yuwata:
These issues are false positive. And the fix is not programmatically correct. Closing.

yuwata:
Also read the discussion here: https://github.com/systemd/systemd/pull/22869#discussion_r836581253

sshedi:
@yuwata Agree that using `&(buffer.ev)` is programmatically correct and prbably a bug in coverity. But it might have thought that, we have initialized `ev` struct but trying to fetch `raw` buffer, so it might be throwing warning.

I have a question, why can't we use `&(buffer)` instead of `&(buffer.ev)` or `&(buffer.raw)`?

PR #23727, opened 2022-06-12
coredump: initialize copy array

sshedi:
Fixes: CID#1469723

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -1074,7 +1074,8 @@ static int send_iovec(const struct iovec_wrapper *iovw, int input_fd) {
                         .msg_iov = iovw->iovec + i,
                         .msg_iovlen = 1,
                 };
-                struct iovec copy[2];
+
+                struct iovec copy[2] = {};
 
                 for (;;) {
                         if (sendmsg(fd, &mh, MSG_NOSIGNAL) >= 0)


src/coredump/coredump.c
```diff
@@ -1074,7 +1074,8 @@ static int send_iovec(const struct iovec_wrapper *iovw, int input_fd) {
                         .msg_iov = iovw->iovec + i,
                         .msg_iovlen = 1,
                 };
-                struct iovec copy[2];
+
+                struct iovec copy[2] = {0};
```
yuwata: Note, we use `{}` instead of `{0}`.

yuwata (changes requested):
This also sounds false positive for me...

src/coredump/coredump.c
sshedi: I saw that but I think Coverity is not happy with it so decided to use `{0}`, I will fix it. Thanks.

src/coredump/coredump.c
```diff
@@ -1074,7 +1074,8 @@ static int send_iovec(const struct iovec_wrapper *iovw, int input_fd) {
                         .msg_iov = iovw->iovec + i,
                         .msg_iovlen = 1,
                 };
-                struct iovec copy[2];
+
+                struct iovec copy[2] = { };
```
yuwata: We do not use `{ }`. Please use `{}`.

yuwata (changes requested):
Still I do not understand the issue. Please explain more.

sshedi:
> Still I do not understand the issue. Please explain more.
Again, this is a false positive and an attempt to keep coverity quiet.

PR #23728, opened 2022-06-12
logs-show: initialze buf to zero

sshedi:
Fixes: CID#1469722

### DIFF
--- a/src/shared/logs-show.c
+++ b/src/shared/logs-show.c
@@ -339,7 +339,7 @@ static int output_timestamp_monotonic(FILE *f, sd_journal *j, const char *monoto
 }
 
 static int output_timestamp_realtime(FILE *f, sd_journal *j, OutputMode mode, OutputFlags flags, const char *realtime) {
-        char buf[CONST_MAX(FORMAT_TIMESTAMP_MAX, 64U)];
+        char buf[CONST_MAX(FORMAT_TIMESTAMP_MAX, 64U)] = {};
         uint64_t x;
         int r;
 


vcaputo:
@sshedi Maybe I'm missing a memo, but your commit message is totally meaningless to me and presumably everyone else that doesn't have visibility into whatever database that "CID" is indexing.

Please actually include the substance of what CID#1469722 actually refers to in the commit message, this is totally opaque.

yuwata:
The issue sounds false positive for me. All paths initialize the buffer. Could you elaborate more?

sshedi:
@vcaputo @yuwata - Sorry about that. This is not about initialization.

CID says: (at line number 425 of this file)
```
CID 1469722 (#1 of 1): String not null terminated (STRING_NULL)20. string_null: Passing unterminated string buf to strlen, which expects a null-terminated string.
``` 

I will include this from next time, thanks for the input.

src/shared/logs-show.c
```diff
@@ -339,7 +339,7 @@ static int output_timestamp_monotonic(FILE *f, sd_journal *j, const char *monoto
 }
 
 static int output_timestamp_realtime(FILE *f, sd_journal *j, OutputMode mode, OutputFlags flags, const char *realtime) {
-        char buf[CONST_MAX(FORMAT_TIMESTAMP_MAX, 64U)];
+        char buf[CONST_MAX(FORMAT_TIMESTAMP_MAX, 64U)] = {0};
```
yuwata: Please use `{}` instead of `{0}`.

yuwata (changes requested):
Still I think this is false positive. In which path the buffer lacks nul-termination?

sshedi:
> Still I think this is false positive. In which path the buffer lacks nul-termination?
> 
> Also, copy-n-pasting the error message from Coverity is not enough. Please explain more.

I agree with you that this might be a false positive, this an attempt to make coverity keep quiet.
I'm not entirely sure that coverity will not report this issue after this change but it doesn't hurt to initialize this buffer with an appropriate value.

BTW, how to mark an issue as false positive in Coverity?

vcaputo:
I'm pretty sure we don't make such pointless commits just to make some other broken linter happy.  It generates unnecessary noise in the history and churn in the code, it's making work for maintainers and other folks who try pay attention to what's happening upstream and if things are CVEs and need backports etc.

This isn't my call but it's frankly annoying to have such merge requests where we're effectively being spammed with scrutinizing the validity of often frivolous changes that aren't even impactful because they're mostly false positives.

It's great that you're looking to contribute upstream, but maybe spend a bit more time understanding the validity what coverity is flagging and learning how to tell it to shut up.

yuwata:
Yeah, @vcaputo's comment makes sense. Closing.

> BTW, how to mark an issue as false positive in Coverity?

Not sure. cc @keszybz?

sshedi:
Thank you @mrc0mmand and @yuwata

PR #23729, opened 2022-06-12
sd-dhcp-server: use calloc to init message buffer

sshedi:
Fixes: CID#1469721


yuwata:
The issue is also false positive, as we have the following:
```
if ((size_t) len < sizeof(DHCPMessage))
        return 0;
```

PR #23730, opened 2022-06-12
udev: initialize tbuffers to zero

sshedi:
Fixes: CID#1469719


yuwata:
The issue is false positive. Closing.

PR #23731, opened 2022-06-12
bootctl: add --root and --image

bluca:


bluca:
> This pull request **introduces 1 alert** when merging [ffd71c2](https://github.com/systemd/systemd/commit/ffd71c2f11d660434c4d2af177ead17f1779a4c5) into [ef55efe](https://github.com/systemd/systemd/commit/ef55efe82331e42208078117169c8751f2aa3f74) - [view on LGTM.com](https://lgtm.com/projects/g/systemd/systemd/rev/pr-6ce22c89308fe6d3956c852d3b27400fd9f60921)
> 
> **new alerts:**
> 
> * 1 for Call to alloca in a loop

It's a statically fixed and bound loop so I wouldn't bother fixing it...

src/boot/bootctl.c
```diff
@@ -607,6 +613,9 @@ static int boot_config_load_and_select(
         if (r < 0)
                 return r;
 
+        if (arg_root)
```
medhefgo: This skips over `boot_config_select_special_entries`, which I think is still relevant with `--root`…

src/shared/find-esp.c
```diff
@@ -438,8 +443,8 @@ int find_esp_and_warn(
                 goto found;
         }
 
-        FOREACH_STRING(_path, "/efi", "/boot", "/boot/efi") {
-                path = _path;
+        FOREACH_STRING(p, "/efi", "/boot", "/boot/efi") {
+                path = prefix_roota(root, p);
```
medhefgo: I'd say this no alloca rule in loops exists for a reason. So this should either be reworked or add a suppression comment with explanation…

bluca: The reason is that input-dependent loop length could cause stack exhaustion, but this is static

medhefgo: Hence why I said this should have a https://lgtm.com/help/lgtm/alert-suppression then.

bluca: yep, adding it

src/boot/bootctl.c
bluca: refactored

medhefgo (approved):

src/shared/find-esp.c
evverx: > this is static

Last time I checked LGTM didn't complain about "static" loops. It complains here because `root` is passed via `--arg-root` and isn't limited anywhere. I can imagine scenarios where `ARG_MAX` * 3 can overflow the stack so if this is supposed to stay I don't think the warning should be suppressed.

medhefgo: Well, if there is a chance for this to blow the stack, it should certainly be changed.

bluca: It's a false positive because of the fixed loop: "1 for Call to alloca in a **loop**"

evverx: I wonder where exactly arg_root is bounded?

src/boot/bootctl.c
```diff
@@ -1523,6 +1568,18 @@ static int parse_argv(int argc, char *argv[]) {
                         assert_not_reached();
                 }
 
+        if ((arg_root || arg_image) && !STRPTR_IN_SET(argv[optind], "status", "list", "install", "update",
+                        "remove", "is-installed", "random-seed"))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                       "Options --root= and --image= are not supported with verb %s.",
```
evverx, line 1619: I wonder what happens when there is no verb. Does it crash or does it print unhelpful messages with `(n/a)/(null)`? More generally, why are verbs analyzed here? I thought that's what `dispatch_verb` is for

src/boot/bootctl.c
```diff
@@ -2388,6 +2459,25 @@ static int run(int argc, char *argv[]) {
...
+        if (arg_image) {
+                assert(!arg_root);
+
+                r = mount_image_privately_interactively(
+                                arg_image,
+                                DISSECT_IMAGE_GENERIC_ROOT |
+                                DISSECT_IMAGE_RELAX_VAR_CHECK,
+                                &unlink_dir,
+                                &loop_device,
+                                &decrypted_image);
```
evverx, line 2517: I wonder what happens with those loop devices when bootcl is interrupted by a signal or something like that?

src/shared/find-esp.c
```diff
@@ -673,6 +679,8 @@ int find_xbootldr_and_warn(
         /* Similar to find_esp_and_warn(), but finds the XBOOTLDR partition. Returns the same errors. */
 
         if (path) {
+                path = prefix_roota(root, path);
```
poettering: we should really stop using prefix_roota() in new code.

Please use `chase_symlinks(path, root, CHASE_PREFIX_ROOT, &p, NULL)` and use the resulting `p`.

That change is trivial to make and has the big benefit that symlinks with absolute targets will be taken relative to the specified root instead of the host.

Here and everywhere.

src/boot/bootctl.c
poettering: they ara marked autoclear, so should be cleaned up automatically.

src/boot/bootctl.c
```diff
@@ -1022,6 +1035,12 @@ static int install_variables(const char *esp_path,
         uint16_t slot;
         int r;
 
+        if (arg_root) {
+                log_warning("Acting on external %s, skipping EFI variable setup.",
```
poettering: "external"? not sure what is "external" about that?

I think this should be downgraded a bit, it's not a warning. It's notice or even info

src/boot/bootctl.c
```diff
@@ -865,16 +876,18 @@ static int copy_one_file(const char *esp_path, const char *name, bool force) {
 
 static int install_binaries(const char *esp_path, const char *arch, bool force) {
         _cleanup_closedir_ DIR *d = NULL;
+        const char *path;
         int r = 0;
 
-        d = opendir(BOOTLIBDIR);
+        path = prefix_roota(arg_install_from_host ? NULL : arg_root, BOOTLIBDIR);
+        d = opendir(path);
```
poettering: chase_symlinks_and_opendir()

src/boot/bootctl.c
poettering: depends on libc used. glibc will output (null), other libcs might crash.

so yes, thia call should cover for a NULL verb better.

poettering (changes requested):
looks excellent.

src/boot/bootctl.c
evverx: At least `_cleanup_(unlink_tempfilep)` doesn't work when programs where it's used are interrupted with Ctrl-C so I'm not sure why `_cleanup_(loop_device_unrefp)` should be any different.

evverx: I ran `test-loop-block` (where `_cleanup_(loop_device_unrefp)` is used), interrupted it with CTRL-C and got a bunch of `/dev/loop*` lying around:
```sh
$ ls -l /dev/loop*
brw-rw----. 1 root disk  7,   0 Jul  6 16:32 /dev/loop0
brw-rw----. 1 root disk  7,   1 Jul  6 16:32 /dev/loop1
brw-rw----. 1 root disk  7,   2 Jul  6 16:32 /dev/loop2
brw-rw----. 1 root disk  7,   3 Jul  6 16:32 /dev/loop3
brw-------. 1 root root  7,   4 Jul  6 16:32 /dev/loop4
crw-rw----. 1 root disk 10, 237 Jul  6 12:25 /dev/loop-control
```
so it seems to me that loop devices are cleaned up when they are set up and managed by PID1. The command line tools can't expect systemd to clean them up.

poettering: > I'm not sure why _cleanup_(loop_device_unrefp) should be any different.

we set the autoclear flag of the loopback device. The kernel will then free the device once it is not referenced anymore, i.e. not mounted or otherwise opened.

poettering: > I ran test-loop-block (where _cleanup_(loop_device_unrefp) is used), interrupted it with CTRL-C and got a bunch of /dev/loop* lying around:

That's misleading. autoclean means the stuff is detached, but the devices stay around. but that's fine, as the next user will use them then. For the next process needing a loopback device there's no difference between "loopback device is gone", and "loopback device exists but not attached" the api calls are the same to allocate/attach something.

Allocation always happens from the bottom, so the worst that happens is that there might be one detached loopback device around for a bit until it is recycled.

evverx: > Allocation always happens from the bottom, so the worst that happens is that there might be one detached loopback device around for a bit until it is recycled.

Got it. Thanks! It's just that I run various tests where stuff gets killed/restarted millions of times and when it leads to a lot of leftovers lying around it usually falls apart somewhere. With loop devices it doesn't seem to be possible to overflow/break anything so it seems it should be fine to rely on the kernel here.

bluca:
> > Note that due to the usual races with the kernel and partitions/device blocks, most of the times it's necessary to use SYSTEMD_RELAX_ESP_CHECKS=1 or it will fail, as /dev/block/x:y is often not ready in time...
> 
> We should relax these checks implicitly if --image= is specified. I mean, if we decided the ESP is good enough to be mounted to /efi/, then we don#t need to check that a second time.
> 
> I.e. maybe change verify_esp() to take a flags param instead of the two bools, and then add a third flag that disables the ESP chekcs, and set then implicitly if arg_root is set in bootctl.
> 
> i.e. pass the flags param all the way through find_esp_and_warn() and so on.

@poettering updated, ptal

src/boot/bootctl.c
poettering: i mean, the API of loopback devices is terrible in a million ways, and the fact these devices stay around in detached state is pretty awful, but it's not a total loss. I wish they'd clean this up more comprehensively.

src/boot/bootctl.c
```diff
@@ -842,7 +853,15 @@ static int copy_one_file(const char *esp_path, const char *name, bool force) {
...
-        p = strjoina(BOOTLIBDIR "/", name);
+        r = chase_symlinks(BOOTLIBDIR "/", arg_root, CHASE_PREFIX_ROOT | CHASE_TRAIL_SLASH, &path, NULL);
+        if (r < 0)
+                return log_error_errno(r,
+                                       "Failed to resolve path %s%s%s: %m",
+                                       BOOTLIBDIR,
+                                       arg_root ? " under directory " : "",
+                                       arg_root ?: "");
+
+        p = strjoina(path, name);
```
poettering: path_join() sounds like the better choice here. then you can drop the "/" suffixing above and the CHASE_TRAI_SLASH, and the (unlikely) special case that BOOTLIBDIR might be the root dir will be covered nicely...

src/boot/bootctl.c
```diff
@@ -842,7 +853,15 @@ static int copy_one_file(const char *esp_path, const char *name, bool force) {
...
         q = strjoina(esp_path, "/EFI/systemd/", dest_name);
```
poettering: should use chase_symlinks() too I guess, to ensure we stay under the image too

src/boot/bootctl.c
bluca: pre-existing conditions but ok, updated, ptal

src/boot/bootctl.c
```diff
@@ -833,29 +843,53 @@ static int create_subdirs(const char *root, const char * const *subdirs) {
...
                 int k;
                 char *v;
 
                 /* Create the EFI default boot loader name (specified for removable devices) */
-                v = strjoina(esp_path, "/EFI/BOOT/BOOT", e);
+                v = strjoina("/EFI/BOOT/BOOT", e);
                 ascii_strupper(strrchr(v, '/') + 1);
 
-                k = copy_file_with_version_check(p, v, force);
+                r = chase_symlinks(v, esp_path, CHASE_PREFIX_ROOT | CHASE_NONEXISTENT, &default_dest_path, NULL);
```
poettering, line 898: you cannot use `r` here. 

The logic here is that we have two copies here, and we execute the second even if the first fails, and then return the first error we encountered. That first error is stored in `r` here.

That said, I find this rather confusing. You tripped up on this, and I think everyone would.

hence I think it would be wiser to maybe use `ret` or so as name for the "saved" return code, and then just use "r" for all the local uses as we usually do.

man/bootctl.xml
```diff
@@ -258,6 +258,16 @@
         switch of the same name.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--install-source</option></term>
```
poettering: we generally try to suffix switches that expect a parameter with `=`, hence please write `--install-source=` here.

man/bootctl.xml
```diff
@@ -258,6 +258,16 @@
...
+        <listitem><para>When installing binaries with <option>--root</option> or
```
poettering: similar

man/bootctl.xml
poettering: also, teh commit msg should probably be updated to reflect the new name

src/boot/bootctl.c
bluca: Yeah completely missed that. Fixed and added a comment to make it more prominent.

PR #23733, opened 2022-06-13
homework: initialize we_detached

sshedi:
Fixes: CID#1469711 CID#1469712

CID 1469711 (#1 of 1): Uninitialized scalar variable (UNINIT)
14. uninit_use: Using uninitialized value we_detached.

CID 1469712 (#1 of 1): Unused value (UNUSED_VALUE)
returned_value: Assigning value from sd_id128_from_string(word + 2, &boot_id) to r here,
but that stored value is overwritten before it can be used.

### DIFF
--- a/src/libsystemd/sd-journal/sd-journal.c
+++ b/src/libsystemd/sd-journal/sd-journal.c
@@ -987,6 +987,8 @@ _public_ int sd_journal_seek_cursor(sd_journal *j, const char *cursor) {
                 case 'b':
                         boot_id_set = true;
                         r = sd_id128_from_string(word + 2, &boot_id);
+                        if (r < 0)
+                                return r;
                         break;
 
                 case 'm':


src/home/homework-luks.c
```diff
@@ -1621,7 +1621,6 @@ int home_deactivate_luks(UserRecord *h, HomeSetup *setup) {
                         return log_error_errno(r, "Failed to initialize cryptsetup context for %s: %m", setup->dm_name);
                 if (r == 0) {
                         log_debug("LUKS device %s has already been detached.", setup->dm_name);
-                        we_detached = false;
                 }
```
yuwata: Please drop unnecessary `{}`.

src/home/homework-luks.c
```diff
@@ -1633,10 +1632,9 @@ int home_deactivate_luks(UserRecord *h, HomeSetup *setup) {
                 r = sym_crypt_deactivate_by_name(setup->crypt_device, setup->dm_name, 0);
                 if (ERRNO_IS_DEVICE_ABSENT(r) || r == -EINVAL) {
                         log_debug_errno(r, "LUKS device %s is already detached.", setup->dm_node);
-                        we_detached = false;
-                } else if (r < 0)
+                } else if (r < 0) {
                         return log_info_errno(r, "LUKS device %s couldn't be deactivated: %m", setup->dm_node);
-                else {
+                } else {
```
yuwata: Also here. Please drop unnecessary `{}`.

src/home/homework-luks.c
```diff
@@ -1603,7 +1603,7 @@ int home_activate_luks(
 }
 
 int home_deactivate_luks(UserRecord *h, HomeSetup *setup) {
-        bool we_detached;
+        bool we_detached = false;
```
yuwata, line 1606: The issue is CID#1469711 should be false positive, as when `acquire_open_luks_device()` returns `1`, `setup->crypt_device` is set.

But the change looks good to me, as it is simpler and clear for readers.

src/libsystemd/sd-journal/sd-journal.c
```diff
@@ -987,6 +987,8 @@ _public_ int sd_journal_seek_cursor(sd_journal *j, const char *cursor) {
                 case 'b':
                         boot_id_set = true;
                         r = sd_id128_from_string(word + 2, &boot_id);
+                        if (r < 0)
+                                return r;
```
yuwata, line 991: Please split the change into another commit.

yuwata (changes requested):
Mostly good.
- Please split changes.
- Please follow our coding style.

src/home/homework-luks.c
sshedi: I though the first `if` is using `{}` even though it has one call, so though of keeping it consistent.

src/home/homework-luks.c
yuwata: Not addressed. Please drop `{}`. See our coding style.

src/home/homework-luks.c
```diff
@@ -1633,7 +1632,6 @@ int home_deactivate_luks(UserRecord *h, HomeSetup *setup) {
                 r = sym_crypt_deactivate_by_name(setup->crypt_device, setup->dm_name, 0);
                 if (ERRNO_IS_DEVICE_ABSENT(r) || r == -EINVAL) {
                         log_debug_errno(r, "LUKS device %s is already detached.", setup->dm_node);
-                        we_detached = false;
                 } else if (r < 0)
```
yuwata: Here too.

yuwata (changes requested):

yuwata (approved):
LGTM.

PR #23735, opened 2022-06-13
hwdb: analyzers: remove generic "STM Device in DFU Mode"

jluebbe:
The USB ID v0483pDF11 is used by the ROM code in many STMicroelectronics
devices (for firmware download) and not just signal analyzers.


yuwata (approved):
Not familiar with the device, but the same USB ID is in 20-usb-vendor-model.hwdb with the following:
```
usb:v0483pDF11*
 ID_MODEL_FROM_DATABASE=STM Device in DFU Mode
```
Hence, the change looks good to me, as devices in DFU mode usually do not work as analyzer, I guess.

PR #23643, opened 2022-06-06
test: fix (not only) TEST-70 under sanitizers

mrc0mmand:
Addresses:
  * https://github.com/systemd/systemd/issues/23578#issuecomment-1144089821
  * https://github.com/systemd/systemd-centos-ci/pull/496#issuecomment-1144640305

/cc @evverx

PR #23741, opened 2022-06-14
test: wrap certain uninstrumented binaries to make them work w/ ASan

mrc0mmand:
Uninstrumented binaries which touch instrumented stuff will fail with

```
==633==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```

One possible solution would be to set `LD_PRELOAD=<path-to-asan-rt>` globally,
but that breaks a _lot_ of stuff, and also reports many ASan/UBSan errors (which
are either false positives or errors we can't do anything about, since they
are not in "our" stuff).

Another, less-invasive solution (yet still ugly), is to wrap the problematic
binaries in a short shell-wrapper, configure the environment in it as necessary,
and then exec() the original binary with the new environment. This seems
to work (I was able to run most of TEST-64-UDEV-STORAGE tests, TEST-25-IMPORT,
and several others, which were originally failing under sanitizers).

---

@evverx I tried to extend and... let's say "polish" the original patch from https://github.com/systemd/systemd/pull/23643 based on the discussion from https://github.com/systemd/systemd-centos-ci/pull/496, and so far it seems to be doing the right thing (at lest from what I've observed). However, it's still pretty ugly :-)

For now I'm curious how it's going to fare in CIs, and also for comments how to make this nicer.

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -228,6 +228,8 @@ DEBUGTOOLS=(
 )
 
 is_built_with_asan() {
+    local _bin="${1:?}"
+
     if ! type -P objdump >/dev/null; then
         ddebug "Failed to find objdump. Assuming systemd hasn't been built with ASAN."
         return 1
@@ -235,7 +237,7 @@ is_built_with_asan() {
 
     # Borrowed from https://github.com/google/oss-fuzz/blob/cd9acd02f9d3f6e80011cc1e9549be526ce5f270/infra/base-images/base-runner/bad_build_check#L182
     local _asan_calls
-    _asan_calls="$(objdump -dC "$SYSTEMD_JOURNALD" | grep -E "(callq?|brasl?|bl)\s.+__asan" -c)"
+    _asan_calls="$(objdump -dC "$_bin" | grep -E "(callq?|brasl?|bl)\s.+__asan" -c)"
     if ((_asan_calls < 1000)); then
         return 1
     else
@@ -251,7 +253,7 @@ is_built_with_coverage() {
     meson configure "${BUILD_DIR:?}" | grep 'b_coverage' | awk '{ print $2 }' | grep -q 'true'
 }
 
-IS_BUILT_WITH_ASAN=$(is_built_with_asan && echo yes || echo no)
+IS_BUILT_WITH_ASAN=$(is_built_with_asan "$SYSTEMD_JOURNALD" && echo yes || echo no)
 IS_BUILT_WITH_COVERAGE=$(is_built_with_coverage && echo yes || echo no)
 
 if get_bool "$IS_BUILT_WITH_ASAN"; then
@@ -2429,7 +2431,9 @@ inst_binary() {
         wrap_binary=1
     fi
 
-    if get_bool "$wrap_binary"; then
+    # If the target binary is built with ASan support, we don't need to wrap
+    # it, as it should handle everything by itself
+    if get_bool "$wrap_binary" && ! is_built_with_asan "$bin"; then
         dinfo "Creating ASan-compatible wrapper for binary '$target'"
         # Install the target binary with a ".orig" suffix
         inst_simple "$bin" "${target}.orig"


mrc0mmand:
> I think if it works it should be good to go.
> 
> I wonder if it would be possible to create those wrappers automatically when the binaries are installed though? It seems in theory `ldd` could help to figure out where `libasan` should be preloaded but on the other hand with `dlopen` and stuff like that it probably wouldn't always work so it would be necessary to create some wrappers manually anyway.

I like that idea! Let me try it out.

evverx:
@mrc0mmand I think another plus would be that if all binaries like that were wrapped automatically it would be possible not to set `LD_PRELOAD` globally in https://github.com/systemd/systemd/blob/b105d413045037cd177120c40428188fba464943/test/test-functions#L839. It should make the ASan wrapper easier to reason about and more predictable. I suspect most memory leaks suppressed via `systemd-lsan.supp` would be gone too. Then again I'm not sure how hard it is to implement that and I can imagine scenarios where it could go sideways anyway :-)

mrc0mmand:
I pushed a first version where we try to detect & wrap such binaries automagically, and it seems to be working (at least for the cases I tried so far). I had to resort to wrapping some binaries unconditionally (like `tar` for `TEST-25`), but thankfully it's not _that_ ugly even with that. I'll play around with this more the first thing tomorrow :-)

mrc0mmand:
> @mrc0mmand I think another plus would be that if all binaries like that were wrapped automatically it would be possible not to set `LD_PRELOAD` globally in
> 
> https://github.com/systemd/systemd/blob/b105d413045037cd177120c40428188fba464943/test/test-functions#L839
> . It should make the ASan wrapper easier to reason about and more predictable. I suspect most memory leaks suppressed via `systemd-lsan.supp` would be gone too. Then again I'm not sure how hard it is to implement that and I can imagine scenarios where it could go sideways anyway :-)

That indeed seems to be the case, since with the last commit everything\* seems to be still working correctly. Let's see what breaks in the CI :-)

\* everything I tried so far manually

test/test-functions
```diff
@@ -2463,7 +2418,35 @@ inst_binary() {
...
-    inst_simple "$bin" "$target"
+
+    # Same as above, but we need to wrap certain libraries unconditionally
+    #
+    # tar - called by machinectl in TEST-25
+    if get_bool "$IS_BUILT_WITH_ASAN" && [[ "$bin" =~ /(tar)$ ]]; then
+        wrap_binary=1
+    fi
+
+    if get_bool "$wrap_binary"; then
```
mrc0mmand: Also, maybe as an additional safeguard we might check if the `$target` binary is built with ASan (i.e. using the stuff from `is_built_with_asan()`), and if so, don't create a wrapper for it. But I can't think of any scenario where it would be useful (at least in current situation), since all systemd+ASan binaries are not installed by this function.

Maybe it would make sense if this was run on a system where not only systemd is instrumented?

evverx: > But I can't think of any scenario where it would be useful (at least in current situation)

https://github.com/bus1/dbus-broker/issues/291 (where `dbus-broker` was instrumented along with `systemd`) would be an example of a scenario where it could be useful :-)

mrc0mmand: Good point, updated :-)

test/test-functions
```diff
@@ -2448,6 +2399,12 @@ inst_binary() {
         # by ldd attempting to use the unprefixed RPATH.
         [[ "$line" =~ libsystemd.*\ not\ found ]] && continue
 
+        # We're built with ASan and the target binary loads one of the systemd's
+        # DSOs, so we need to tweak the environment before executing the binary
+        if get_bool "$IS_BUILT_WITH_ASAN" && [[ "$line" =~ /(libudev|libsystemd).so ]]; then
```
evverx: I think it should look for all shared libraries `systemd` ships
```
$ find -name '*.so'
./build/libnss_systemd.so
./build/libcryptsetup-token-systemd-fido2.so
./build/libcryptsetup-token-systemd-pkcs11.so
./build/libnss_mymachines.so
./build/pam_systemd_home.so
./build/libudev.so
./build/libcryptsetup-token-systemd-tpm2.so
./build/libnss_resolve.so
./build/libsystemd.so
./build/src/shared/libsystemd-shared-251.so
./build/src/core/libsystemd-core-251.so
./build/libnss_myhostname.so
./build/pam_systemd.so
```
Without them it wouldn't be possible to log in in "interactive" mode for example:
```sh
H login: root
==69==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```

evverx: Though I suspect there `dlopen` is used to load the pam modules so the script can't figure it out with `ldd`

mrc0mmand: Ah, damn, good catch. I tried to make the "instrumented DSO detection mechanism" a bit more generic by checking if any of the DSOs the binary tries to load is compiled with ASan, but the performance of the image build is abysmal with it, pity. I guess I'll just make a list of DSOs systemd ships/builds and check just that list for each binary.

mrc0mmand: Oh well, I tried several different *possible solutions*, and eventually just changed the regex to `'/(libudev|libsystemd.*|.+[\-_]systemd([\-_].+)?|libnss_(mymachines|myhostname|resolve)).so'` to keep my sanity, since apart from just hunting down all the systemd DSOs (which might be in several places, as we don't have to install a compiled systemd), there's also issue with various formats of the lines from `ldd` (sometimes it says something like `libsystemd... not found`, since the test libsystemd version is different from the host's one, etc.)

As for the interactive session, wrapping `login` unconditionally helped, but the terminal input with that is... very weird, I need to look into more.

evverx: @mrc0mmand I think it would make sense to roll back to the first version of the PR for now and bring back `LD_PRELOAD` and the suppression file and create a bunch of wrappers manually. Given that the tests didn't fail anywhere it appears the pam and nss modules aren't covered by the testsuite so it's hard to tell what else can be broken. I think it can be revisited once those libraries are covered at least partially.

mrc0mmand: I wonder if it would suffice to just drop the last two commits, and keep the automatic wrapper creation, but with `LD_PRELOAD=` and suppression file in place (just so we don't have to maintain the list manually, apart from the exceptions).

mrc0mmand: Interesting. Turns out the "weird tty behavior" happens on the latest main as well, at least on my machine, so it's possible my setup is somehow borked, *sigh*. With wrapping `login` it seems to work as expected. I'll repush the previous version with the `login` binary and regex tweaked, just to see if CIs are going to complain.

evverx (approved):
LGTM. Thanks!

@mrc0mmand Ubuntu CI got stuck at some point. Could you force-push it once again just to make sure it was just a glitch?

yuwata:
> Ubuntu CI got stuck at some point.

Now all green!

mrc0mmand:
@evverx I wanted to extend the list of tests we run under sanitizers (again) and stumbled across an interesting issue - calls to `textdomain(3)` under ASan cause a deadlock. Unfortunately, this is used by `procps` utils like `ps`, `pidof`, `pkill`, etc. (utilized, for example, in `TEST-35-LOGIN`, which render the test broken under ASan.

This can be easily reproduced (tested on latest Fedora Rawhide):

```
# LD_PRELOAD=/usr/lib64/libasan.so.8.0.0 ps --help
<hangs>

# LD_PRELOAD=/usr/lib64/libasan.so.8.0.0 ltrace -- ps --help
__cxa_atexit(0x56537682dd70, 0, 0x565376835f80, 0x565376835f78)                          = 0
strrchr("ps", '/')                                                                       = nil
setlocale(LC_ALL, "")                                                                    = "C.UTF-8"
bindtextdomain("procps-ng", "/usr/share/locale")                                         = "/usr/share/locale"
textdomain("procps-ng"
```

With clang's RTs it behaves the same:
```
# LD_PRELOAD=/usr/lib64/clang/14.0.0/lib/libclang_rt.asan-x86_64.so ltrace -- ps --help
__cxa_atexit(0x5652c8e3ad70, 0, 0x5652c8e42f80, 0x5652c8e42f78)                          = 0
strrchr("ps", '/')                                                                       = nil
setlocale(LC_ALL, "")                                                                    = "C.UTF-8"
bindtextdomain("procps-ng", "/usr/share/locale")                                         = "/usr/share/locale"
textdomain("procps-ng"
```

I found https://bugzilla.redhat.com/show_bug.cgi?id=1716928 with the same issue, unfortunately it was closed without resolution.

Have you, by any chance, stumble across this in your ASan endeavors? I wonder if it would make sense to file this to clang/gcc upstream issue tracker.

mrc0mmand:
Looks like this is already reported in clang's upstream issue tracker, with no response so far: https://github.com/llvm/llvm-project/issues/49223.

mrc0mmand:
And I also summed it up in https://bugzilla.redhat.com/show_bug.cgi?id=2098125.

evverx:
> I wonder if it would make sense to file this to clang/gcc upstream issue tracker.

It certainly would but I don't think it's going to be fixed anytime soon.

Those binaries should be linked statically to avoid preloading libasan. Another (fragile) option would be to copy the shared libraries those binaries are linked against to a separate directory and point the loader to it with `LD_LIBRARY_PATH` at runtime or patch `rpath` somewhere. Then again, I'm not sure it makes sense to turn the script into something that generates images fully compatible with ASan :-) It's good enough already.

evverx:
I think another option would be to recompile `procps` without `libsystemd` (as far as I understand that's what pulls libasan there) and install those binaries before running the script. It seems to be the easiest way to get the logind test to work. Having said that I don't know whether the test relies on features provided by `pkill`/`ps` via `libsystemd` or not so it can go sideways in mysterious ways :-)

PR #23749, opened 2022-06-15
logind-session-dbus: allow to set display name via dbus

dtardon:
Currently, the only way to set display name of a graphical session is to
pass it to CreateSession(). But modern display managers like gdm start
the display server as part of the user session, which means that the
display name isn't known yet when the session is being created. Hence,
let's make it possible to set it later.

### DIFF
--- a/man/org.freedesktop.login1.xml
+++ b/man/org.freedesktop.login1.xml
@@ -1075,6 +1075,7 @@ node /org/freedesktop/login1/session/1 {
       TakeControl(in  b force);
       ReleaseControl();
       SetType(in  s type);
+      SetDisplay(in  s display);
       TakeDevice(in  u major,
                  in  u minor,
                  out h fd,
@@ -1172,6 +1173,8 @@ node /org/freedesktop/login1/session/1 {
 
     <variablelist class="dbus-method" generated="True" extra-ref="SetType()"/>
 
+    <variablelist class="dbus-method" generated="True" extra-ref="SetDisplay()"/>
+
     <variablelist class="dbus-method" generated="True" extra-ref="TakeDevice()"/>
 
     <variablelist class="dbus-method" generated="True" extra-ref="ReleaseDevice()"/>
@@ -1267,6 +1270,11 @@ node /org/freedesktop/login1/session/1 {
       control is released, either by calling <function>ReleaseControl()</function> or closing the D-Bus
       connection. This should help prevent a session from entering an inconsistent state, for example if the
       controller crashes. The only argument <varname>type</varname> is the new session type.</para>
+
+      <para><function>SetDisplay()</function> allows the display name of the graphical session to be changed. This is
+      useful if the display server is started as part of the session. It can only be called by session's current
+      controller. If <function>TakeControl()</function> has not been called, this method will fail. The only argument
+      <varname>display</varname> is the new display name.</para>
 
       <para><function>TakeDevice()</function> allows a session controller to get a file descriptor for a
       specific device. Pass in the major and minor numbers of the character device and
--- a/src/login/logind-session-dbus.c
+++ b/src/login/logind-session-dbus.c
@@ -413,6 +413,30 @@ static int method_set_type(sd_bus_message *message, void *userdata, sd_bus_error
         return sd_bus_reply_method_return(message, NULL);
 }
 
+static int method_set_display(sd_bus_message *message, void *userdata, sd_bus_error *error) {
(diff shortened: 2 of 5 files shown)


src/login/logind-session-dbus.c
```diff
@@ -901,6 +922,11 @@ static const sd_bus_vtable session_vtable[] = {
                                 SD_BUS_NO_RESULT,
                                 method_set_type,
                                 SD_BUS_VTABLE_UNPRIVILEGED),
+        SD_BUS_METHOD_WITH_ARGS("SetDisplay",
+                                SD_BUS_ARGS("s", type),
```
yuwata: Maybe it is better the method name more specific, e.g. `SetDisplayName`.
And the argument name should be `type` -> `name`.

man/org.freedesktop.login1.xml
```diff
@@ -1268,6 +1271,11 @@ node /org/freedesktop/login1/session/1 {
       connection. This should help prevent a session from entering an inconsistent state, for example if the
       controller crashes. The only argument <varname>type</varname> is the new session type.</para>
 
+      <para><function>SetDisplay()</function> allows to change the display name of a graphical session. This is useful
+      if the display server is started as part of the session. It can only be called by session's current controller.
+      If <function>TakeControl()</function> has not been called, this method will fail. The only argument
+      <varname>display</varname> is the new display name.</para>
```
yuwata, line 1277: `display` -> `name` (or at least `type` if you'd prefer the current argument name.)

src/login/logind-session-dbus.c
dtardon: > Maybe it is better the method name more specific, e.g. `SetDisplayName`.

That would only cause confusion, as the corresponding (read-only) property for this is named `Display`.

> And the argument name should be `type` -> `name`.

Sorry, copypasta...

yuwata: > > Maybe it is better the method name more specific, e.g. `SetDisplayName`.
> 
> That would only cause confusion, as the corresponding (read-only) property for this is named `Display`.

That makes sense for me.

man/org.freedesktop.login1.xml
```diff
@@ -1268,6 +1271,11 @@ node /org/freedesktop/login1/session/1 {
       connection. This should help prevent a session from entering an inconsistent state, for example if the
       controller crashes. The only argument <varname>type</varname> is the new session type.</para>
 
+      <para><function>SetDisplay()</function> allows to change the display name of a graphical session. This is useful
```
halfline: this seems useful indeed.

src/login/logind-session-dbus.c
```diff
@@ -413,6 +413,27 @@ static int method_set_type(sd_bus_message *message, void *userdata, sd_bus_error
...
+        assert(message);
+
+        r = sd_bus_message_read(message, "s", &display);
+        if (r < 0)
+                return r;
+
+        if (!session_is_controller(s, sd_bus_message_get_sender(message)))
+                return sd_bus_error_set(error, BUS_ERROR_NOT_IN_CONTROL, "You must be in control of this session to set display");
+
+        r = free_and_strdup(&s->display, display);
```
halfline: Does the right thing happen if the session type is not a graphical session type?

Shouldn't this call be `session_set_display` or something and have it save the results?

dtardon: It seems `logind` doesn't use the display for anything internally, so it's likely not a big deal (e.g., `CreateSession()` assigns the display unconditionally as well). But restricting it makes sense and it also follows what the man page says, therefore I've added a check for this.

keszybz: This needs to have the same logic as `method_set_type()` above, i.e. save the value and send a signal in addition to changing the internal value. We currently don't restart logind because of bugs, but we certainly want to, and this setting would be lost then.

dtardon: Updated. PTAL.

man/org.freedesktop.login1.xml
keszybz: > allows to change the display name of a graphical session

allows the display name of the graphical session to be changed

dtardon: Updated.

man/org.freedesktop.login1.xml
Hamidb14: I dont understand 😖

Hamidb14 (approved):

PR #23750, opened 2022-06-15
resolve: provide service resolve over varlink

vchillara:
ported the d-bus implementation of service resolve to varlink

### DIFF
--- a/src/resolve/resolved-varlink.c
+++ b/src/resolve/resolved-varlink.c
@@ -17,6 +17,15 @@ typedef struct LookupParameters {
         char *name;
 } LookupParameters;
 
+typedef struct LookupParametersResolveService {
+        const char *name;
+        const char *type;
+        const char *domain;
+        int family;
+        int ifindex;
+        uint64_t in_flags;
+} LookupParametersResolveService;
+
 static void lookup_parameters_destroy(LookupParameters *p) {
         assert(p);
         free(p->name);
@@ -160,14 +169,62 @@ static bool validate_and_mangle_flags(
         return true;
 }
 
+static int find_addr_records(
+                JsonVariant **array,
+                DnsQuestion *question,
+                DnsQuery *q,
+                DnsResourceRecord **canonical,
+                const char *search_domain) {
+        DnsResourceRecord *rr;
+        int ifindex, r;
+
+        DNS_ANSWER_FOREACH_IFINDEX(rr, ifindex, q->answer) {
+                _cleanup_(json_variant_unrefp) JsonVariant *entry = NULL;
+                int family;
+                const void *p;
+
+                r = dns_question_matches_rr(question, rr, search_domain);
+                if (r < 0)
+                        return r;
+                if (r == 0)
+                        continue;
+
+                if (rr->key->type == DNS_TYPE_A) {
+                        family = AF_INET;
+                        p = &rr->a.in_addr;
+                } else if (rr->key->type == DNS_TYPE_AAAA) {
+                        family = AF_INET6;
+                        p = &rr->aaaa.in6_addr;
+                } else {
+                        return -EAFNOSUPPORT;
+                }
+
+                r = json_build(&entry,
+                               JSON_BUILD_OBJECT(
+                                               JSON_BUILD_PAIR_CONDITION(ifindex > 0, "ifindex", JSON_BUILD_INTEGER(ifindex)),
+                                               JSON_BUILD_PAIR("family", JSON_BUILD_INTEGER(family)),
+                                               JSON_BUILD_PAIR("address", JSON_BUILD_BYTE_ARRAY(p, FAMILY_ADDRESS_SIZE(family)))));
+                if (r < 0)
+                        return r;
+
+                r = json_variant_append_array(array, entry);
+                if (r < 0)
+                        return r;
+
+                if (canonical && !*canonical)
+                        *canonical = dns_resource_record_ref(rr);
+        }
+
+        return 0;
+}
+
 static void vl_method_resolve_hostname_complete(DnsQuery *query) {
         _cleanup_(dns_resource_record_unrefp) DnsResourceRecord *canonical = NULL;
         _cleanup_(json_variant_unrefp) JsonVariant *array = NULL;
         _cleanup_(dns_query_freep) DnsQuery *q = query;
         _cleanup_free_ char *normalized = NULL;
-        DnsResourceRecord *rr;
         DnsQuestion *question;
-        int ifindex, r;
+        int r;
 
         assert(q);
 
@@ -191,43 +248,9 @@ static void vl_method_resolve_hostname_complete(DnsQuery *query) {
 
         question = dns_query_question_for_protocol(q, q->answer_protocol);
 
-        DNS_ANSWER_FOREACH_IFINDEX(rr, ifindex, q->answer) {
-                _cleanup_(json_variant_unrefp) JsonVariant *entry = NULL;
-                int family;
-                const void *p;
-
-                r = dns_question_matches_rr(question, rr, DNS_SEARCH_DOMAIN_NAME(q->answer_search_domain));
-                if (r < 0)
-                        goto finish;
-                if (r == 0)
-                        continue;
-
-                if (rr->key->type == DNS_TYPE_A) {
-                        family = AF_INET;
-                        p = &rr->a.in_addr;
-                } else if (rr->key->type == DNS_TYPE_AAAA) {
-                        family = AF_INET6;
-                        p = &rr->aaaa.in6_addr;
-                } else {
-                        r = -EAFNOSUPPORT;
-                        goto finish;
-                }
-
-                r = json_build(&entry,
-                               JSON_BUILD_OBJECT(
-                                               JSON_BUILD_PAIR_CONDITION(ifindex > 0, "ifindex", JSON_BUILD_INTEGER(ifindex)),
-                                               JSON_BUILD_PAIR("family", JSON_BUILD_INTEGER(family)),
-                                               JSON_BUILD_PAIR("address", JSON_BUILD_BYTE_ARRAY(p, FAMILY_ADDRESS_SIZE(family)))));
-                if (r < 0)
-                        goto finish;
-
-                if (!canonical)
-                        canonical = dns_resource_record_ref(rr);
-
-                r = json_variant_append_array(&array, entry);
-                if (r < 0)
-                        goto finish;
-        }
+        r = find_addr_records(&array, question, q, &canonical, DNS_SEARCH_DOMAIN_NAME(q->answer_search_domain));
+        if (r < 0)
+                goto finish;
 
         if (json_variant_is_blank_object(array)) {
                 r = varlink_error(q->varlink_request, "io.systemd.Resolve.NoSuchResourceRecord", NULL);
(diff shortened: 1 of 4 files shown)


yuwata:
Which systemd component uses the varlink method?

vchillara:
@yuwata 
This is intended to complete the workflow of discovering and resolving services through varlink.
Clients can browse for mDNS services via #22532, and resolve services with this PR.

yuwata:
> @yuwata This is intended to complete the workflow of discovering and resolving services through varlink. Clients can browse for mDNS services via #22532, and resolve services with this PR.

Sorry, I cannot follow. This provides a verlink method, but no one use it... Unlike DBus methods, we do not expose varlink methods, and only they are used internally, at least currently.

vchillara:
Our basic goal is to enable mDNS service discovery and resolution like in avahi.
Our initial (rather naive) approach was to dump the cache over d-bus in #20053.
But Poettering suggested doing this over varlink and sticking to the RFCs, which is what we did in #22532.
The idea is for an independent client, maybe implemented with libvarlink, to be able to connect to APIs exposed in resolved-varlink and get notified continuously.
The client could then make a follow-up call to io.systemd.Resolve.ResolveService (implemented here) to resolve any service of interest.

src/resolve/resolved-varlink.c
```diff
@@ -516,6 +532,547 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+                                                        JSON_BUILD_PAIR("family", JSON_BUILD_INTEGER(family)),
+                                                        JSON_BUILD_PAIR("address", JSON_BUILD_BYTE_ARRAY(p, FAMILY_ADDRESS_SIZE(family)))));
+                                if (r < 0)
+                                        return r;
+
+                                r = json_variant_append_array(_addr, entry);
+                                if (r < 0)
+                                        return r;
+                        }
+                }
```
poettering, line 675: this loop exists in almost the exact same way in `vl_method_resolve_hostname_complete()` already. please unify it in a common helper.

vchillara: Updated in the latest commit

vchillara:
> So I think this makes sense to add. But we really need some test case for this, that checks that this even remotely does what is intended.
> 
> For the address/hostname resolution we pretty quickly notice if things go wrong, since it's used all the time by nss-resolve. But this here has no user in our own codebase, hence relaly needs some proper testcase.

Testing is being discussed in https://github.com/systemd/systemd/pull/22532 (at least for mDNS)

src/resolve/resolved-varlink.c
```diff
@@ -534,6 +564,514 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+                log_error_errno(r, "Failed to send address reply: %m");
+                r = varlink_error_errno(q->varlink_request, r);
+        }
+}
+
+static int vl_method_resolve_service(Varlink* link, JsonVariant* parameters, VarlinkMethodFlags flags, void* userdata) {
+        static const JsonDispatch dispatch_table[] = {
+                { "name",    JSON_VARIANT_STRING,   json_dispatch_string, offsetof(LookupParametersResolveService, name),     JSON_MANDATORY },
+                { "type",    JSON_VARIANT_STRING,   json_dispatch_string, offsetof(LookupParametersResolveService, type),     JSON_MANDATORY },
+                { "domain",  JSON_VARIANT_STRING,   json_dispatch_string, offsetof(LookupParametersResolveService, domain),   JSON_MANDATORY },
```
poettering: please use json_dispatch_const_string(). That way the strings don't have to be copied. Given you don't need them after this function finished that sounds like a good idea.

src/resolve/resolved-varlink.c
```diff
@@ -157,45 +173,23 @@ static bool validate_and_mangle_flags(
...
-        _cleanup_(dns_resource_record_unrefp) DnsResourceRecord *canonical = NULL;
-        _cleanup_(json_variant_unrefp) JsonVariant *array = NULL;
-        _cleanup_(dns_query_freep) DnsQuery *q = query;
-        _cleanup_free_ char *normalized = NULL;
+static int find_addr_records(
+                JsonVariant **array,
+                DnsQuestion *question,
+                DnsQuery *q,
+                DnsResourceRecord **canonical,
+                bool search_domain_flag) {
```
poettering: hmmm, why not make this parameter a `const char* search_domain`? i.e. already call DNS_SEARCH_DOMAIN_NAME() in the caller where appropriate and pass NULL here if not appropriate. I think that would be more readable.

src/resolve/resolved-varlink.c
```diff
@@ -216,16 +209,53 @@ static void vl_method_resolve_hostname_complete(DnsQuery *query) {
...
-                if (!canonical)
-                        canonical = dns_resource_record_ref(rr);
+                if (canonical && !*canonical)
+                        *canonical = dns_resource_record_ref(rr);
 
-                r = json_variant_append_array(&array, entry);
+                r = json_variant_append_array(array, entry);
                 if (r < 0)
-                        goto finish;
+                        return r;
```
poettering, line 212: i'd really update the array first, since this can fail, and if it does then you already updated `*canonical`... if you switch this, then the function will either fail or succeed, but never return half-initialized data...

src/resolve/resolved-varlink.c
```diff
@@ -534,6 +564,514 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+        assert(rr->key);
+
+        if (rr->key->type != DNS_TYPE_TXT)
+                return 0;
+
+        LIST_FOREACH(items, i, rr->txt.items) {
+                if (i->length <= 0)
+                        continue;
+
+                r = json_build(&entry,  JSON_BUILD_BYTE_ARRAY(i->data, i->length));
```
poettering: please encode this as base63, not as byte array.

also, use json_build() only if you build complex json objects. for simple stuff just call the indiviudal constructors directly, i.e.:


```c
r = json_variant_new_base64(&entry, i->data, i->length);
```

src/resolve/resolved-varlink.c
```diff
@@ -534,6 +564,514 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+                if (r < 0)
+                        return r;
+
+                r = json_variant_append_array(_txt, entry);
+                if (r < 0)
+                        return r;
+        }
```
poettering, line 579: this leaks memory, as you do not free the entry object here. i.e. move the declaration of `entry` into the loop

src/resolve/resolved-varlink.c
```diff
@@ -534,6 +564,514 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+
+        /* Note that above we appended the hostname as encoded in the
+         * SRV, and here the canonical hostname this maps to. */
+        r = json_build(_norm,
+                        JSON_BUILD_OBJECT(
+                                        JSON_BUILD_PAIR("normalized", JSON_BUILD_STRING(normalized))));
+        if (r < 0)
+                return r;
+
+        return 1;
```
poettering, line 690: so we generally have this rule that parameters that are purely output should be filled in only on success, and not touched on failure. here, you will touch `_srv` early on, even if later stuff can still fail. Hence, please always build your stuff in temporary vairables, and only if you know nothing can fail anymore, move it to the return parameters. (usually TAKE_PTR() is a good way to do so).

also, please name return variables `ret_xyz` as indicated above

src/resolve/resolved-varlink.c
```diff
@@ -534,6 +564,514 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+        /* Find the main query */
+        while (aux->auxiliary_for)
+                aux = aux->auxiliary_for;
+
+        return aux->varlink_request;
+}
+
+static void resolve_service_all_complete(DnsQuery *query) {
+        _cleanup_(dns_query_freep) DnsQuery *q = query;
+        _cleanup_(json_variant_unrefp) JsonVariant *_srv = NULL, *_addr = NULL, *_txt = NULL, *_norm = NULL, *_canonical = NULL;
```
poettering: why the underscore here?

src/resolve/resolved-varlink.c
```diff
@@ -534,6 +564,514 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+        if (r < 0)
+                goto finish;
+
+        r = json_build(&_canonical,
+                               JSON_BUILD_OBJECT(
+                                               JSON_BUILD_PAIR("name", JSON_BUILD_STRING(name)),
+                                               JSON_BUILD_PAIR("type", JSON_BUILD_STRING(type)),
+                                               JSON_BUILD_PAIR("domain", JSON_BUILD_STRING(domain))));
+        if (r < 0)
+                goto finish;
```
poettering: why build this object separately? you can just build this as part of varlink_replyb(), simply by inserting JSON_BUILD_OBJECT() where you currently use JSON_BUILD_VARIANT() for the canonical field

src/resolve/resolved-varlink.c
```diff
@@ -15,11 +15,27 @@ typedef struct LookupParameters {
         char *name;
 } LookupParameters;
 
+typedef struct LookupParametersResolveService {
+        char *name;
+        char *type;
+        char *domain;
```
poettering: as mentioned further down, please use `const char*` here, and fill in the fields via json_dispatch_const_string() as then you don't need to copy them, and also not free them

poettering (changes requested):
this really needs some form of testing, i.e some simple addition to TEST-75-RESOLVED.

poettering:
and sorry here too for the slow slow reviews. mea culpa!

vchillara:
@poettering addressed your comments.
Looking into testing for this

vchillara:
@poettering added test for resolve over varlink in TEST-75-RESOLVED

vchillara:
Rebased and resolved conflicts.
Please review @poettering

src/resolve/resolved-varlink.c
```diff
@@ -530,6 +553,516 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
 }
 
+static int append_txt(JsonVariant **ret_txt, DnsResourceRecord *rr) {
+        int r;
+
+        assert(rr);
+        assert(rr->key);
+
+        if (rr->key->type != DNS_TYPE_TXT)
+                return 0;
```
poettering, line 564: you return from the function here successfully (i.e. with an error code >= 0).

according to our coding style functions should initialize all return paramaters on success (and leave them untouched on failure). THus, you need to add `*ret_txt = NULL` here, if you return 0.

src/resolve/resolved-varlink.c
```diff
@@ -530,6 +553,516 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+                _cleanup_(json_variant_unrefp) JsonVariant *entry = NULL;
+
+                if (i->length <= 0)
+                        continue;
+
+                r = json_variant_new_base64(&entry, i->data, i->length);
+                if (r < 0)
+                        return r;
+
+                r = json_variant_append_array(ret_txt, entry);
```
poettering: hmm, so now i am confused. our coding style says that you should name return parameters ret_xyz. A return parameter is a purely output param. But here you extend what you get. Hence either the code is wrong or the name (i.e. drop the "ret_" prefix from the parmaete name)

my comment regarding return parameter initialization above is rooted in the misunderstanding that I thought this was a purely output param because of the name, which it appaerntly isn't after all

src/resolve/resolved-varlink.c
```diff
@@ -530,6 +553,516 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+        /* Find the main query */
+        while (aux->auxiliary_for)
+                aux = aux->auxiliary_for;
+
+        return aux->varlink_request;
+}
+
+static void resolve_service_all_complete(DnsQuery *query) {
+        _cleanup_(dns_query_freep) DnsQuery *q = query;
+        _cleanup_(json_variant_unrefp) JsonVariant *ret_srv = NULL, *ret_addr = NULL, *ret_txt = NULL, *ret_norm = NULL;
```
poettering: please don't name local variables `ret_xyz`. As per our coding style that's for output-only (i.e. "return" params), hence never for local variables. just drop the `ret_` prefix

test/units/testsuite-75.sh
```diff
@@ -389,6 +390,23 @@ grep -qF "myservice.signed.test:1234" "$RUN_OUT"
...
+{
+  "method": "io.systemd.Resolve.ResolveService",
+  "more": false,
+  "parameters":{"name":"","type":"_mysvc._tcp","domain":"signed.test"}
+}\0')| nc -U /run/systemd/resolve/io.systemd.Resolve > $VL_RESOLVE_LOGS
+
+grep -qF '"srv":{"priority":10,"weight":5,"port":1234,"hostname":"myservice.signed.test"}' "$VL_RESOLVE_LOGS"
+grep -qF '{"ifindex":4,"family":10,"address":[253,0,222,173,190,239,202,254,0,0,0,0,0,0,0,23]}' "$VL_RESOLVE_LOGS"
+grep -qF '{"ifindex":4,"family":2,"address":[10,0,0,20]}' "$VL_RESOLVE_LOGS"
+grep -qF '"normalized":{"normalized":"myservice.signed.test"}' "$VL_RESOLVE_LOGS"
```
poettering: hmm a field which contains an object whith a field also called like the object#s field? that doesn't look right

test/units/testsuite-75.sh
```diff
@@ -389,6 +390,23 @@ grep -qF "myservice.signed.test:1234" "$RUN_OUT"
...
+grep -qF '"canonical":{"name":null,"type":"_mysvc._tcp","domain":"signed.test"}' "$VL_RESOLVE_LOGS"
+TXT_OUT=$(grep -a -o -P '(?<=\"txt\"\:\[\").*(?=\"\])' "$VL_RESOLVE_LOGS" | base64 --decode)
+assert_in "This is TXT for myservice" "$TXT_OUT"
```
poettering, line 403: please add a test with the varlinkctl tool

poettering:
CI failed badly.

Please add varlink introspection support, i.e. add the method call, all errors and the structures you added to src/shared/varlink-io.systemd.Resolve.c

poettering:
And sorry for taking so terribly long to review this.

test/units/testsuite-75.sh
vchillara: removed the wrapper field

src/resolve/resolved-varlink.c
vchillara: dropped the ret_ prefix.

src/resolve/resolved-varlink.c
vchillara: fixed

test/units/testsuite-75.sh
vchillara: added test with varlinkctl

vchillara:
@poettering fixed the CI failures due to TEST-75 additions. 
Added introspection support.

A couple of mkosi checks seem to be failing, but it looks unrelated to the code changes.

src/shared/varlink-io.systemd.Resolve.c
```diff
@@ -32,6 +32,33 @@ static VARLINK_DEFINE_METHOD(
...
+                ResolveService,
+                VARLINK_DEFINE_INPUT(name, VARLINK_STRING, 0),
+                VARLINK_DEFINE_INPUT(type, VARLINK_STRING, 0),
+                VARLINK_DEFINE_INPUT(domain, VARLINK_STRING, 0),
+                VARLINK_DEFINE_INPUT(ifindex, VARLINK_INT, VARLINK_NULLABLE),
+                VARLINK_DEFINE_INPUT(family, VARLINK_INT, VARLINK_NULLABLE),
+                VARLINK_DEFINE_INPUT(flags, VARLINK_INT, VARLINK_NULLABLE),
+                VARLINK_DEFINE_OUTPUT_BY_TYPE(srv, ResolvedService, 0),
+                VARLINK_DEFINE_OUTPUT_BY_TYPE(addr, ResolvedAddress, VARLINK_ARRAY),
+                VARLINK_DEFINE_OUTPUT(txt, VARLINK_STRING, VARLINK_ARRAY),
```
evverx, line 58: Looks like this is triggered when services comes with no key/value pairs:
```
Varlink-IDL: Mandatory field 'txt' is null or missing on object, refusing.
varlink-26: Return parameters for method reply ResolveService() didn't pass validation on field 'txt', ignoring: No anode
varlink-26: Sending message: {"parameters":{"srv":{"priority":0,"weight":0,"port":123,"hostname":"host-5.local"},"addr":[{"ifindex":3,"family":2,"address":[192,168,56,3]},{"ifindex":3,"family":10,"address":[254,128,0,0,0,0,0,0,0,224,25,56,10,209,68,157]}],"txt":null,"normalized":"host-5.local","canonical":{"name":"X","type":"_ipp._tcp","domain":"local"}}}
```
it should probably emit empty arrays instead of `null` when it receives TXT RRs with a single empty string
```
            Name: X._ipp._tcp.local                                                                                                                           
            Type: TXT (Text strings) (16)                                                                                                                     
            .000 0000 0000 0001 = Class: IN (0x0001)                                                                                                          
            1... .... .... .... = Cache flush: True                                                                                                           
            Time to live: 4500 (1 hour, 15 minutes)                                                                                                           
            Data length: 1                                                                                                                                    
            TXT Length: 0                                                                                                                                     
            TXT:                                  
```
or zero-length RDATA.

src/resolve/resolved-varlink.c
```diff
@@ -530,6 +553,509 @@ static int vl_method_resolve_address(Varlink *link, JsonVariant *parameters, Var
...
+                return varlink_error_invalid_parameter(link, JSON_VARIANT_STRING_CONST("type"));
+
+        r = dns_name_is_valid(p.domain);
+        if (r < 0)
+                return r;
+        if (r == 0)
+                return varlink_error_invalid_parameter(link, JSON_VARIANT_STRING_CONST("domain"));
+
+        if (!validate_and_mangle_flags(p.name, &p.in_flags, 0))
+                return varlink_error_invalid_parameter(link, JSON_VARIANT_STRING_CONST("flags"));
```
evverx, line 1031: I'm not sure if it's intentional but it seems it isn't possible to pass `SD_RESOLVED_NO_TXT` or `SD_RESOLVED_NO_ADDRESS` (corresponding to `--service-txt=no` and `--service-txt=no`). Currently resolved rejects those flags with
```
varlink-26: Received message: {"method":"io.systemd.Resolve.ResolveService","parameters":{"name":"Y","type":"_ipp._tcp","domain":"local","flags":192}}
varlink-26: Changing state idle-server → processing-method
varlink-26: Sending message: {"error":"org.varlink.service.InvalidParameter","parameters":{"parameter":"flags"}}
varlink-26: Changing state processing-method → processed-method
varlink-26: Changing state processed-method → idle-server
```

PR #23751, opened 2022-06-15
resolve: mark mDNS RRs in the Additional Records section as cacheable

vchillara:
RFC 6762 does not restrict caching of records in the Additional section.
Caching can improve resolve efficiency.
Avahi also allows caching of records from the AR section.


yuwata (approved):
LGTM.

Issue #23742, opened 2022-06-14
Services with RootImage hitting the restart limit lead to failed modprobe@dm_verity.service


AlbanBedel:
**systemd version the issue has been seen with**
 > 247

This version is a bit older but it is on an embedded system where I can't easily switch to a newer version. From my analysis of the related code and unit files I strongly believe this bug is still present in the latest version.

**Unexpected behaviour you saw**
 > When a portable service hit the restart limit I also see modprobe@dm_verity.service as failed because it also hit the restart limit.

**Steps to reproduce the problem**
 > Make sure dm_verity module is not available, create a service with a RootImage that fails right away, configure it for automatic restart, then launch it.

Looking at the code, units with a RootImage automatically get several modprobe@ added as dependency, which lead to this problem if the respective module doesn't exists. The modprobe@ unit in itself doesn't fail when the module is missing as the command is allowed to fail, but it then hit the restart limit because the parent service keep restarting it.

MR #15027 used to mitigate this problem when the modules can be loaded but it doesn't help when the module can't get loaded. But it was reverted anyways. On the other hand using `RemainAfterExit=yes` in modprobe@.service as was suggested in #14819 do solve this problem. Removing the restart limit in the modprobe service would be another solution.

yuwata:
> Removing the restart limit in the modprobe service would be another solution.

I think this is reasonable. @poettering @bluca WDYT?

yuwata:
> Removing the restart limit in the modprobe service would be another solution.

@AlbanBedel Could you provide a PR?

PR #23753, opened 2022-06-15
units: remove the restart limit on the modprobe@.service

AlbanBedel:
They are various cases where the same module might be repeatedly
loaded in a short time frame, for example if a service depending on a
module keep restarting, or if many instances of such service get
started at the same time. If this happend the modprobe@.service
instance will be marked as failed because it hit the restart limit.

Overall it doesn't seems to make much sense to have a restart limit on
the modprobe service so just disable it.

Fixes: #23742


yuwata (approved):
LGTM. But let's wait for other reviewers.

keszybz:
Looks reasonable.

PR #23754, opened 2022-06-15
nspawn: fix UID map string

### LINKED ISSUE #23664: nspawn: more than one --bind-user coredumps
systemd v250

ubuntu 20.04

5.4.0-113-generic

amd64

When one --bind-user specified everything goes as expected, with more than one switch it coredumps.
But it is expected, that mapping of more than one user is available here.

To reproduce try to use BindUser or --bind-user twice or more

systemd-nspawn[124716]: Assertion 'previous_uid <= payload_uid' failed at src/nspawn/nspawn.c:4120, function make_uid_map_string(). Aborting.

yuwata:
We send/recv the set of payload uid, host uid, payload gid, host gid.
Hence, the index must be incremented with 4, instead of 2.

Fixes #23664.

### DIFF
--- a/src/nspawn/nspawn.c
+++ b/src/nspawn/nspawn.c
@@ -4134,8 +4134,8 @@ static int make_uid_map_string(
          * quadruplet, consisting of host and container UID + GID. */
 
         for (size_t i = 0; i < n_bind_user_uid; i++) {
-                uid_t payload_uid = bind_user_uid[i*2+offset],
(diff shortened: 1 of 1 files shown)


yuwata:
The issue was introduced by 2f8930449079403b26c9164b8eeac78d5af2c8df (v249).

DaanDeMeyer (approved):

PR #23755, opened 2022-06-15
oomd: try the next cgroup candidate when no processes are killed

enr0n:
The oomd_cgroup_kill function returns 0 if no processes were killed, and
returns 1 if processes were killed. The oomd_kill_by_pgscan_rate and
oomd_kill_by_swap_usage do not currently handle the case when 0 is
returned, i.e. no processes were killed.

This can lead to a case where oomd is stuck in a loop trying to kill the
same cgroup over and over. When e.g. the swap usage threshold is
reached, oomd will try to kill the top candidate to relieve swap usage.
But, if no processes are actually killed, no swap is freed up, and oomd
will not try the next candidate. Thus, when oomd again sees that the
swap usage threshold is reached, this pattern continues.

Fix this by making sure oomd_kill_by_pgscan_rate and oomd_kill_by_swap_usage
try the next candidate cgroup when oomd_cgroup_kill returns 0.

### DIFF
--- a/src/oom/oomd-util.c
+++ b/src/oom/oomd-util.c
@@ -266,8 +266,8 @@ int oomd_kill_by_pgscan_rate(Hashmap *h, const char *prefix, bool dry_run, char 
                 r = oomd_cgroup_kill(sorted[i]->path, true, dry_run);
                 if (r == -ENOMEM)
                         return r; /* Treat oom as a hard error */
-                if (r < 0) {
-                        if (ret == 0)
+                if (r <= 0) {
+                        if (r < 0 && ret == 0)
                                 ret = r;
                         continue; /* Try to find something else to kill */
                 }
@@ -310,8 +310,8 @@ int oomd_kill_by_swap_usage(Hashmap *h, uint64_t threshold_usage, bool dry_run, 
                 r = oomd_cgroup_kill(sorted[i]->path, true, dry_run);
                 if (r == -ENOMEM)
                         return r; /* Treat oom as a hard error */
-                if (r < 0) {
-                        if (ret == 0)
+                if (r <= 0) {
+                        if (r < 0 && ret == 0)
                                 ret = r;
                         continue; /* Try to find something else to kill */
                 }


yuwata (approved):
LGTM. But let's wait for @anitazha's review.

anitazha (commented):
The original version was written as intended. If systemd-oomd selected a candidate but didn't kill anything, we want to return to the parent function in oomd-manager.c. In the parent function it will go through the timer event loop again and recalculate the candidates. This should ensure we don't inadvertently kill the next cgroup in the list without first checking if the system/cgroups are still over their thresholds.

anitazha:
> Here is a log snippet that demonstrates the issue I observed. Note that despite systemd-oomd reporting it killed org.gnome.Shell@wayland.service, the desktop session apparently survived.

Memory and swap basically remained unchanged despite all the processes in `org.gnome.Shell@wayland.service` being killed? That doesn't seem right at all. Because of `THRESHOLD_SWAP_USED_PERCENT`, systemd-oomd selects candidates using more than 5% of swap ("big" consumers). Based on your logs it doesn't look like 5% of swap is being restored. Does this repro consistently? If you could send me instructions I would also like to repro this myself

enr0n:
> > Here is a log snippet that demonstrates the issue I observed. Note that despite systemd-oomd reporting it killed [org.gnome.Shell@wayland.service](mailto:org.gnome.Shell@wayland.service), the desktop session apparently survived.
> 
> Memory and swap basically remained unchanged despite all the processes in `org.gnome.Shell@wayland.service` being killed? That doesn't seem right at all. Because of `THRESHOLD_SWAP_USED_PERCENT`, systemd-oomd selects candidates using more than 5% of swap ("big" consumers). Based on your logs it doesn't look like 5% of swap is being restored. Does this repro consistently? If you could send me instructions I would also like to repro this myself

Yes, it does repro pretty consistently, but it is not very straight forward. Basically, I run a bunch of apps like chromium, firefox, libreoffice, etc. and create heavy workloads (lots of tabs, many open documents, etc.) and hope that when `SwapUsedLimit` is reached, `org.gnome.Shell@wayland.service` has the highest swap usage. I have tried to automate this with `stress-ng`, but weirdly in that case `org.gnome.Shell@wayland.service` is killed properly.

First, start a kinetic VM with lxc:

```bash
 lxc launch images:ubuntu/22.04/desktop ubuntu --vm -c limits.cpu=4 -c limits.memory=4GiB --console=vga
 ```
 Once that boots, copy and run this script in the VM:
 
 ```bash
 #!/bin/bash
set -euo pipefail

if [[ $EUID != 0 ]]; then
    echo "This script must be run as root."
    exit 1
fi

if ! grep "/swapfile" /etc/fstab; then
    # Create a swap file
    fallocate -l 1G /swapfile
    chmod 600 /swapfile
    mkswap /swapfile
    swapon /swapfile
    echo "/swapfile swap swap defaults 0 0" >> /etc/fstab
fi

# Install some desktop applications
snap install firefox chromium snap-store
apt install libreoffice -y

# Install the latest systemd and re-enable swap kill
cat <<EOF >/etc/apt/sources.list.d/ubuntu-$(lsb_release -cs)-proposed.list
# Enable Ubuntu proposed archive
deb http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs)-proposed restricted main multiverse universe
EOF

apt update
apt install systemd -y

sed -i 's/auto/kill/g' /lib/systemd/system/-.slice.d/10-oomd-root-slice-defaults.conf

mkdir -p /etc/systemd/system/systemd-oomd.service.d
cat <<EOF >/etc/systemd/system/systemd-oomd.service.d/override.conf
[Service]
Environment=SYSTEMD_LOG_LEVEL=debug
EOF

systemctl daemon-reload
systemctl restart systemd-oomd
 ```

From there, I create a heavy workload as described above. I wish I had better instructions, but hopefully this is enough to help you observe the issue.

enr0n:
> ```shell
>  lxc launch images:ubuntu/kinetic/desktop ubuntu --vm -c limits.cpu=4 -c limits.memory=4GiB --console=vga
> ```
I edited this from the original, because it should have been `kinetic`, not `22.04`.

keszybz:
@anitazha please make a decision here.

anitazha:
I moved this to my personal TODO. Closing since this PR won't address the problem AFAICT

PR #23756, opened 2022-06-15
analyze: fix seccomp filter handling in security command

### LINKED ISSUE #23663: [REGRESSION] Wrong `systemd-analyze security` output with regard to `SystemCallFilter=`

```
systemd 251 (251.2-1-arch)
+PAM +AUDIT -SELINUX -APPARMOR -IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT -QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK +XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified
```

**Used distribution**

Arch Linux

```
[~]> uname -a
Linux spock 5.18.0-pf2 #1 SMP PREEMPT_DYNAMIC Mon, 06 Jun 2022 14:51:13 +0000 x86_64 GNU/Linux
```

`x86_64`

Having multiple `SystemCallFilter=` lines in the `.service` file I expect to see that the first one defines default behaviour (_allowlisting vs denylisting_) as explained in the documentation:

> If you specify both types of this option (i.e. allow-listing and deny-listing), the first encountered will take precedence and will dictate the default action (termination or approval of a system call). Then the next occurrences of this option will add or delete the listed system calls from the set of the filtered system calls, depending of its type and the default action.

This is not the case however.

I've got the following lines in my `.service` file:

```
[~]> grep -n SystemCallFilter= /usr/lib/systemd/system/tinc@.service
33:SystemCallFilter=@system-service
34:SystemCallFilter=~@resources @privileged
```

And this is what `systemd-analyze security` gives me:

```
[~]> systemd-analyze security --offline=true /usr/lib/systemd/system/tinc@.service | grep SystemCallFilter=
✗ SystemCallFilter=~@swap                                     System call allow list defined for service, and @swap is included (e.g. swapoff is allowed)                  0.2
✗ SystemCallFilter=~@resources                                System call allow list defined for service, and @resources is included (e.g. ioprio_set is allowed)          0.2
✗ SystemCallFilter=~@reboot                                   System call allow list defined for service, and @reboot is included (e.g. kexec_file_load is allowed)        0.2
✗ SystemCallFilter=~@raw-io                                   System call allow list defined for service, and @raw-io is included (e.g. ioperm is allowed)                 0.2
✗ SystemCallFilter=~@privileged                               System call allow list defined for service, and @privileged is included (e.g. chown is allowed)              0.2
✗ SystemCallFilter=~@obsolete                                 System call allow list defined for service, and @obsolete is included (e.g. _sysctl is allowed)              0.1
✗ SystemCallFilter=~@mount                                    System call allow list defined for service, and @mount is included (e.g. chroot is allowed)                  0.2
✗ SystemCallFilter=~@module                                   System call allow list defined for service, and @module is included (e.g. delete_module is allowed)          0.2
✗ SystemCallFilter=~@debug                                    System call allow list defined for service, and @debug is included (e.g. lookup_dcookie is allowed)          0.2
✗ SystemCallFilter=~@cpu-emulation                            System call allow list defined for service, and @cpu-emulation is included (e.g. modify_ldt is allowed)      0.1
✗ SystemCallFilter=~@clock                                    System call allow list defined for service, and @clock is included (e.g. adjtimex is allowed)                0.2
```

This is not correct, because I've excluded at least `@privileged` and `@resources` explicitly.

(_the same applies to online check, so `--offline=true` option doesn't make any difference_)

My full `.service` file is shown below:

```
[Unit]
Description=Tinc net %i
Documentation=info:tinc
Documentation=man:tinc(8) man:tinc.conf(5)
Documentation=http://tinc-vpn.org/docs/
PartOf=tinc.service
ReloadPropagatedFrom=tinc.service

[Service]
DynamicUser=true
User=tinc-%i
Group=tinc-%i
AmbientCapabilities=CAP_NET_ADMIN CAP_NET_BIND_SERVICE
CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_BIND_SERVICE
RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
RestrictNamespaces=true
PrivateDevices=true
BindPaths=/dev/net/tun
DeviceAllow=/dev/net/tun rw
ProtectClock=true
ProtectControlGroups=true
ProtectHome=true
ProtectKernelLogs=true
ProtectKernelModules=true
ProtectKernelTunables=yes
ProtectProc=invisible
ProcSubset=pid
SystemCallArchitectures=native
RestrictRealtime=true
LockPersonality=true
MemoryDenyWriteExecute=true
ProtectHostname=true
SystemCallFilter=@system-service
SystemCallFilter=~@resources @privileged
TasksMax=2
UMask=0066
WorkingDirectory=/var/lib/tinc/%i
ConfigurationDirectory=tinc/%i
RuntimeDirectory=tinc/%i
StateDirectory=tinc/%i
PIDFile=${RUNTIME_DIRECTORY}/pid
LoadCredential=%i:/etc/credstore/tinc/%i
ExecStart=/usr/bin/tincd -D --pidfile=${RUNTIME_DIRECTORY}/pid -n %i -o Ed25519PrivateKeyFile=${CREDENTIALS_DIRECTORY}/%i_ed25519_key.priv -o PrivateKeyFile=${CREDENTIALS_DIRECTORY}/%i_rsa_key.priv -o Interface=tinc-%i
ExecReload=/usr/bin/tinc --pidfile=${RUNTIME_DIRECTORY}/pid -n %i reload
Type=notify
WatchdogSec=10
Restart=on-failure

[Install]
WantedBy=tinc.service
```

It was not the case in the past. I cannot say when the issue was introduced, but IIRC v250 behaved correctly.

Also, if I swap lines in the `.service` file so that the denylist comes first, then `systemd-analyze security` doesn't report any issue:

```
[~]> grep -n SystemCallFilter= /usr/lib/systemd/system/tinc@.service
33:SystemCallFilter=~@resources @privileged
34:SystemCallFilter=@system-service

[~]> systemd-analyze security --offline=true /usr/lib/systemd/system/tinc@.service | grep SystemCallFilter=
✓ SystemCallFilter=~@swap                                     System call deny list defined for service, and @swap is included
✓ SystemCallFilter=~@resources                                System call deny list defined for service, and @resources is included
✓ SystemCallFilter=~@reboot                                   System call deny list defined for service, and @reboot is included
✓ SystemCallFilter=~@raw-io                                   System call deny list defined for service, and @raw-io is included
✓ SystemCallFilter=~@privileged                               System call deny list defined for service, and @privileged is included
✓ SystemCallFilter=~@obsolete                                 System call deny list defined for service, and @obsolete is included
✓ SystemCallFilter=~@mount                                    System call deny list defined for service, and @mount is included
✓ SystemCallFilter=~@module                                   System call deny list defined for service, and @module is included
✓ SystemCallFilter=~@debug                                    System call deny list defined for service, and @debug is included
✓ SystemCallFilter=~@cpu-emulation                            System call deny list defined for service, and @cpu-emulation is included
✓ SystemCallFilter=~@clock                                    System call deny list defined for service, and @clock is included
```

But this goes against the documentation.

yuwata:
Fixes #23663.

### DIFF
--- a/test/units/testsuite-65.sh
+++ b/test/units/testsuite-65.sh
@@ -3,6 +3,9 @@
 # shellcheck disable=SC2016
 set -eux
 
+# shellcheck source=test/units/assert.sh
+. "$(dirname "$0")"/assert.sh
+
 systemd-analyze log-level debug
 export SYSTEMD_LOG_LEVEL=debug
 
@@ -605,6 +608,63 @@ if systemd-analyze --version | grep -q -F "+ELFUTILS"; then
 fi
 
 systemd-analyze --threshold=90 security systemd-journald.service
+
+# issue 23663
+check() {(
+    set +x
+    output=$(systemd-analyze security --offline="${2?}" "${3?}" | grep -F 'SystemCallFilter=')
+    assert_in "System call ${1?} list" "$output"
+    assert_in "[+✓] SystemCallFilter=~@swap" "$output"
+    assert_in "[+✓] SystemCallFilter=~@resources" "$output"
+    assert_in "[+✓] SystemCallFilter=~@reboot" "$output"
+    assert_in "[+✓] SystemCallFilter=~@raw-io" "$output"
+    assert_in "[-✗] SystemCallFilter=~@privileged" "$output"
+    assert_in "[+✓] SystemCallFilter=~@obsolete" "$output"
+    assert_in "[+✓] SystemCallFilter=~@mount" "$output"
+    assert_in "[+✓] SystemCallFilter=~@module" "$output"
+    assert_in "[+✓] SystemCallFilter=~@debug" "$output"
+    assert_in "[+✓] SystemCallFilter=~@cpu-emulation" "$output"
+    assert_in "[-✗] SystemCallFilter=~@clock" "$output"
+)}
+
+export -n SYSTEMD_LOG_LEVEL
+
+mkdir -p /run/systemd/system
+cat >/run/systemd/system/allow-list.service <<EOF
+[Service]
+ExecStart=false
+SystemCallFilter=@system-service
+SystemCallFilter=~@resources:ENOANO @privileged
+SystemCallFilter=@clock
+EOF
+
+cat >/run/systemd/system/deny-list.service <<EOF
+[Service]
+ExecStart=false
+SystemCallFilter=~@known
+SystemCallFilter=@system-service
+SystemCallFilter=~@resources:ENOANO @privileged
+SystemCallFilter=@clock
+EOF
+
+systemctl daemon-reload
+
+check allow yes /run/systemd/system/allow-list.service
+check allow no allow-list.service
+check deny yes /run/systemd/system/deny-list.service
+check deny no deny-list.service
+
+output=$(systemd-run -p "SystemCallFilter=@system-service" -p "SystemCallFilter=~@resources:ENOANO @privileged" -p "SystemCallFilter=@clock" sleep 60 2>&1)
+name=$(echo "$output" | awk '{ print $4 }')
+
+check allow yes /run/systemd/transient/"$name"
+check allow no "$name"
+
+output=$(systemd-run -p "SystemCallFilter=~@known" -p "SystemCallFilter=@system-service" -p "SystemCallFilter=~@resources:ENOANO @privileged" -p "SystemCallFilter=@clock" sleep 60 2>&1)
+name=$(echo "$output" | awk '{ print $4 }')
+
+check deny yes /run/systemd/transient/"$name"
+check deny no "$name"
 
 systemd-analyze log-level info
 


yuwata:
The issue is caused by 1624114d74f55ad9791b7624b08d89d2339a68b3 (v250) and its follow-up dd51e725df9aec2847482131ef601e0215b371a0 (v251).

src/core/dbus-execute.c
```diff
@@ -375,6 +375,10 @@ static int property_get_syscall_filter(
                 char *s;
                 int num = PTR_TO_INT(val);
 
+                if (c->syscall_allow_list && num >= 0)
```
bluca, line 378: This looks like a backward incompatible change? And also I'm not sure it's right? It means nothing will be returned when it's allow-listed

src/core/dbus-execute.c
```diff
@@ -2470,7 +2470,7 @@ int bus_exec_context_set_transient_property(
                                                                          -1,
                                                                          c->syscall_filter,
                                                                          SECCOMP_PARSE_PERMISSIVE |
-                                                                         SECCOMP_PARSE_ALLOW_LIST | invert_flag,
```
bluca, line 2469: are you sure it's always 0?

src/basic/hashmap.c
```diff
@@ -1853,10 +1853,13 @@ int _set_put_strdup_full(Set **s, const struct hash_ops *hash_ops, const char *p
         if (r < 0)
                 return r;
 
-        if (set_contains(*s, (char*) p))
-                return 0;
+        if (n == SIZE_MAX) {
+                if (set_contains(*s, (char*) p))
```
bluca, line 1857: why is this check done only for the n == SIZE_MAX case?

src/core/dbus-execute.c
yuwata: Strictly speaking, yes, it breaks compatibility with recent releases. However this fixes a compatibility breakage introduced in the past by 68acc1afbe5cec50da1ffdc411dadda504e4caf5. Before the commit, an allow-list does not contain denied syscalls.

Note, if the list is an allow-list, then it contains all allowed syscalls (with `num == -1`) and denied syscalls specified with errno or error action (`num >= 0`). This change makes that the bus property returns only allowed syscalls.
The bus property was designed before we accept denied syscalls with errno or error action, hence, we have no way to provide denied syscalls with errno without breaking (very old) compatibility.
If people want to get syscalls with errno in allowed list, then we need to provide another bus property, say, SystemCallFilterEx= or so.

I will update commit message above points if you approve the change.

src/core/dbus-execute.c
yuwata: Yes. This block is called only when the syscall filter was empty and requested filter is an allow-list. Hence `invert_flag == 0`.

src/basic/hashmap.c
yuwata: When `n != SIZE_MAX`, to check the existence, we need to call `strndup()` first. But calling `set_contains()` then `set_consume()` is quite redundant. Hence, the check is done only when `n == SIZE_MAX`.

Or in other words, if and only if `n == SIZE_MAX`, we can check the existence without duplicating the string.

bluca (approved):

yuwata:
Comments are updated. Setting the green label.

PR #23762, opened 2022-06-16
udev: init EVENT_RESULT_EXIT_STATUS_BASE to 1 to remove dead code.

sshedi:
Fixes: CID#1487104

CID 1487104 (#1 of 1): Logically dead code (DEADCODE)
dead_error_condition: The switch governing value result cannot be EVENT_RESULT_EXIT_STATUS_BASE.

### DIFF
--- a/src/udev/udevd.c
+++ b/src/udev/udevd.c
@@ -395,7 +395,7 @@ static void device_broadcast(sd_device_monitor *monitor, sd_device *dev, EventRe
                                 (void) device_add_property(dev, "UDEV_WORKER_ERRNO_NAME", str);
                         break;
                 }
-                case EVENT_RESULT_EXIT_STATUS_BASE ... EVENT_RESULT_EXIT_STATUS_MAX:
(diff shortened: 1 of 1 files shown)


yuwata:
Unfortunately, CIs are unhappy with the change...
I think the issue should be also marked as false positive, as the switch case is a range, not a single value.

JoeyeS0:
Why not just let it run for a few

yuwata:
I guess CI can be happy if we introduce a new enum e.g. `EVENT_RESULT_EXIT_STATUS_MIN_FAIL = 1`, but it will not be used any other place.

Anyway, as we already discussed in other PRs, let's not introduce workarounds to make Coverity into silence. Closing.

yuwata:
@sshedi But, thank you for your recent works about checking Coverity reports!

JoeyeS0:
Okay sorry no expert in any of this just curiosity really. Don't mind me!

sshedi:
@yuwata no problem. I understand. 

@yuwata  & @JoeyeS0 Thanks for your time and patience.

PR #18401, opened 2021-01-29
oomd: implement avoid/omit support for cgroups

anitazha:
There may be situations where a cgroup should be protected from killing
or deprioritized as a candidate. In FB oomd xattrs are used to bias oomd
away from supervisor cgroups and towards worker cgroups in container
tasks. On desktops this can be used to protect important units with
unpredictable resource consumption.

The patch allows systemd-oomd to understand 2 xattrs:
"user.oomd_avoid" and "user.oomd_omit". If systemd-oomd sees these
xattrs set to 1 on a candidate cgroup (i.e. while attempting to kill something)
AND the cgroup is owned by root:root, it will either deprioritize the cgroup as
a candidate (avoid) or remove it completely as a candidate (omit).

Usage is restricted to root:root cgroups to prevent situations where an
unprivileged user can set their own cgroups lower in the kill priority than
another user's (and prevent them from omitting their units from
systemd-oomd killing).

The service manager is also extended with ManagedOOMPreference= to make
it easier to set these xattrs on units.

cc @dschatzberg @jaredp

PR #19007, opened 2021-03-15
RFC: Permit user to use ManagedOOMPreference=avoid

benzea:
The idea is to first select the user from which to kill a unit, and
allow OOM killing preference within the user.

With this, we could set ManagedOOMPreference=avoid (or even omit) on
important units in session.slice in order to protect them from being
killed.

NOTE: This code has not been tested

CC: @anitazha

PR #23764, opened 2022-06-16
oomd: allow ManagedOOMPreference on all cgroups 

enr0n:
Firstly, I consider this a sort of RFC because I know that it cannot be merged in the current state. But, I think making this change is worthwhile, specifically for desktop environments using systemd-oomd.

The motivation behind these patches is to allow desktop environments to have some control over which applications are targeted (or avoided) by systemd-oomd when it needs to act on candidate cgroups. In the current state, the `ManagedOOMPreference` property is ignored for any cgroups which are not owned by root. This is unfortunate for desktop environments where applications are launched as descendants of `user@.service`.

In practice, when systemd-oomd needs to act on desktop systems, user applications such as browsers are killed first. For desktop users, it is very likely that their browser (or some other desktop application) is their most critical program, and it may make sense to de-prioritize the browser's cgroup as a kill candidate. For this reason, I think it is important that systemd-oomd honors `ManagedOOMPreference` on cgroups which are not owned by root.

Based on the conversation in https://github.com/systemd/systemd/pull/18401/, I understand why this restriction on `ManagedOOMPreference` exists -- to "prevent situations where an unprivileged user can set their own cgroups lower in the kill priority than another user's". The "RFC" part of this PR is then, how can we retain the ability of root to have discretion over `ManagedOOMPreference` without completely restricting the property from cgroups owned by non-root users?

### DIFF
--- a/src/oom/oomd-util.c
+++ b/src/oom/oomd-util.c
@@ -323,7 +323,7 @@ int oomd_kill_by_pgscan_rate(Hashmap *h, const char *prefix, bool dry_run, char 
                         continue; /* Try to find something else to kill */
                 }
 
-                dump_until = MAX(dump_until, i);
+                dump_until = MAX(dump_until, i + 1);
                 char *selected = strdup(sorted[i]->path);
                 if (!selected)
                         return -ENOMEM;
@@ -367,7 +367,7 @@ int oomd_kill_by_swap_usage(Hashmap *h, uint64_t threshold_usage, bool dry_run, 
                         continue; /* Try to find something else to kill */
                 }
 
-                dump_until = MAX(dump_until, i);
+                dump_until = MAX(dump_until, i + 1);
                 char *selected = strdup(sorted[i]->path);
                 if (!selected)
                         return -ENOMEM;


anitazha:
We also had https://github.com/systemd/systemd/pull/19007 that is a bit outdated now since it was written before we had systemd-oomd settings for user services. I need to recall the idea I had before but it was something like a permission/UID check between the cgroup with xattrs and the user manager that "owns" the systemd-oomd settings. That way we can ignore the xattrs if we see it on e.g. the system.slice level, but respect it if it's at the user@.slice level. (I'll be traveling for the next week or so so my responses will be spotty.)

enr0n:
> We also had #19007 that is a bit outdated now since it was written before we had systemd-oomd settings for user services. I need to recall the idea I had before but it was something like a permission/UID check between the cgroup with xattrs and the user manager that "owns" the systemd-oomd settings. 

Similarly to the way this is done for `ManagedOOMSwap` and `ManagedOOMMemoryPressure` in `process_managed_oom_message`?

> That way we can ignore the xattrs if we see it on e.g. the system.slice level, but respect it if it's at the [user@.slice](mailto:user@.slice) level. (I'll be traveling for the next week or so so my responses will be spotty.)

I don't think I follow, can you please elaborate? When exactly would we ignore the xattrs?

enr0n:
> Similar to the ManagedOOM*= settings themselves, the xattrs should also be grouped according to the requester. Example: if we're processing candidates for a cgroup owned by root (e.g. pid1, service manager), we might see `user.oomd_omit` xattr set on `/user.slice/user-1000.slice/user@1000.service/app.slice/myuserservice.service`. In this case we would ignore that xattr because it was set by the user manager for UID 1000. This would avoid unprivileged users affecting the kill priority of an ancestor cgroup.

Okay, I think I understand. Does that mean, for example, that if system was configured with `ManagedOOM{Swap,MemoryPressure}=kill` on `user@.service`, but not on any ancestor cgroups, then unprivileged users would be able to configure `ManagedOOMPreference`? And, if a system was configured with `ManagedOOM{Swap,MemoryPressure}=kill` on e.g. `-.slice`, then unprivileged users would **never** be able to configure `ManagedOOMPreference`?

enr0n:
I pushed new commits to implement what we have discussed, based on my understanding of the approach.

enr0n:
I see the failing tests, and I will push changes to address that.

enr0n:
The remaining test failure appears to be unrelated to these changes.

man/systemd.resource-control.xml
```diff
@@ -1150,9 +1150,16 @@ DeviceAllow=/dev/loop-control
...
-          <command>systemd-oomd</command> will ignore these extended attributes if the unit's cgroup is not
-          owned by the root user.</para>
+          in order to use <option>avoid</option> or <option>omit</option>.</para>
+
+          <para><command>systemd-oomd</command> will only respect these extended attributes if
+          the unit's cgroup owner, and the owner of the monitored ancestor cgroup are the
+          same. For example, if <filename>-.slice</filename> is monitored by <command>systemd-oomd</command>,
+          then these extended attributes would be ignored if set on descendants of
+          <filename>/user.slice/user-1000.slice/user@1000.service/</filename>. However, if no ancestors of
+          <filename>/user.slice/user-1000.slice/user@1000.service/</filename> are monitored by
```
anitazha: > However, if no ancestors of <filename>/user.slice/user-1000.slice/user@1000.service/</filename> are monitored by...

This should be allowed as long as the avoid/omit cgroup is a candidate for a monitored cgroup that has the same cgroup owner. So for example, if we have something like this:

- -.slice # monitoring for pressure threshold 60%
  - user.slice
    - user-1000.slice
      - user@1000.service # monitoring for pressure threshold 30%
        - app.slice
          - my.service # oomd_omit

xattrs on `my.service` can be respected if it is a candidate for `user@1000.service`. But if we're calculating the candidates for `-.slice`, then oomd xattrs on `my.service` should be ignored. Basically regardless of whether `user@1000.service` has an ancestor being monitored or not. Does that make sense?

This last part should be reworded to reflect that (maybe remove the stuff about "no ancestors"). I think this might also simplify some of the code such that the check between UIDs would only need to happen during candidate selection/ordering.

src/oom/oomd-util.c
```diff
@@ -355,10 +357,10 @@ int oomd_cgroup_context_acquire(const char *path, OomdCGroupContext **ret) {
         if (r < 0)
                 return log_debug_errno(r, "Error parsing memory pressure from %s: %m", p);
 
-        r = cg_get_owner(SYSTEMD_CGROUP_CONTROLLER, path, &uid);
-        if (r < 0)
-                log_debug_errno(r, "Failed to get owner/group from %s: %m", path);
-        else if (uid == 0) {
+        if ((r = cg_get_owner(SYSTEMD_CGROUP_CONTROLLER, path, &cg_uid)) < 0 ||
+            (r = cg_get_owner(SYSTEMD_CGROUP_CONTROLLER, monitored_ancestor_path, &ancestor_cg_uid)) < 0)
+                log_debug_errno(r, "Failed to get owner/group from %s and %s: %m", path, monitored_ancestor_path);
```
anitazha: and -> and/or (?)
Alternatively split these conditions up and log the messages separately

anitazha (changes requested):

man/systemd.resource-control.xml
enr0n: > > However, if no ancestors of /user.slice/user-1000.slice/user@1000.service/ are monitored by...
> 
> This should be allowed as long as the avoid/omit cgroup is a candidate for a monitored cgroup that has the same cgroup owner. So for example, if we have something like this:
> 
> * -.slice # monitoring for pressure threshold 60%
>   
>   * user.slice
>     
>     * user-1000.slice
>       
>       * [user@1000.service](mailto:user@1000.service) # monitoring for pressure threshold 30%
>         
>         * app.slice
>           
>           * my.service # oomd_omit
> 
> xattrs on `my.service` can be respected if it is a candidate for `user@1000.service`. But if we're calculating the candidates for `-.slice`, then oomd xattrs on `my.service` should be ignored. Basically regardless of whether `user@1000.service` has an ancestor being monitored or not. Does that make sense?

Yes, this is a good clarifying example. It is more clear looking at the memory pressure scenario because `oomd_pressure_above` does initial filtering of the cgroups which means (using the example above) `user@1000.service` could be a target, but not `-.slice`. But, with swap kill every monitored cgroup is an initial candidate and will be sorted in `oomd_kill_by_swap_usage`, which led me to the monitored ancestor logic.

> This last part should be reworded to reflect that (maybe remove the stuff about "no ancestors"). I think this might also simplify some of the code such that the check between UIDs would only need to happen during candidate selection/ordering.

Yes, I think this should simplify the code quite a bit. Thanks for taking a look.

enr0n:
I believe the remaining test failures are unrelated (I know the focal tests are infrastructure related, not sure about CentOS CI). PTAL.

src/oom/oomd-manager.c
```diff
@@ -404,21 +406,35 @@ static int monitor_swap_contexts_handler(sd_event_source *s, uint64_t usec, void
...
 
                 threshold = m->system_context.swap_total * THRESHOLD_SWAP_USED_PERCENT / 100;
-                r = oomd_kill_by_swap_usage(candidates, threshold, m->dry_run, &selected);
-                if (r == -ENOMEM)
-                        return log_oom();
-                if (r < 0)
-                        log_notice_errno(r, "Failed to kill any cgroup(s) based on swap: %m");
+
+                n = oomd_sort_cgroup_contexts(m->monitored_swap_cgroup_contexts,
+                                              compare_cgroup_path_and_swap_usage,
```
anitazha: I'm pretty sure we don't want to do this. The idea with the swap kill policy is to operate globally to kill the largest candidate. Adding priority based on their position in the cgroup hierarchy goes against this intent.

anitazha (commented):
Sorry for the delay. I did one round over the PR and it looks great. I would want to do another round tomorrow to be sure.
However I'm not sure we want the swap kill candidate change (commented inline).

src/oom/oomd-manager.c
enr0n: Yeah, I see what you mean. How exactly should we determine when a swap kill candidate's OOM preference should be honored then? Since swap kill candidates are sorted all at once, regardless of which monitored cgroup context they are descendant of, we cannot use the same logic as the memory pressure scenario (this commit was an attempt to make that possible).

If we just dropped "oomd: calculate swap kill candidates one monitored cgroup at a time" from this PR, then the existing behavior (only cgroups owned by root can set `ManagedOOMPreference`) would remain for the swap kill scenario.

anitazha: > If we just dropped "oomd: calculate swap kill candidates one monitored cgroup at a time" from this PR, then the existing behavior (only cgroups owned by root can set ManagedOOMPreference) would remain for the swap kill scenario.

I think this makes the most sense. Given that swap kill thresholds are globally controlled, root should be able to use `ManagedOOMPreference=`. Non-root cgroups that don't want to get killed could have someone privileged set up a more fine grained policy where certain cgroups are omitted altogether from the candidate list. For example:
```
/ [ManagedOOMSwap=auto]
  /system.slice [ManagedOOMSwap=kill]
  /user.slice [ManagedOOMSwap=auto]
    /user.slice/user-1000.slice [ManagedOOMSwap=kill]
    /user.slice/user-3000.slice [ManagedOOMSwap=auto]
```
Which would skip `/user.slice/user-3000.slice` as a swap kill candidate, but still include everything else

enr0n: Okay, I have removed this commit from the PR. I believe the test failures are unrelated.

src/oom/test-oomd-util.c
```diff
@@ -451,6 +406,90 @@ static void test_oomd_sort_cgroups(void) {
...
+        test_xattrs = !ERRNO_IS_PRIVILEGE(r) && !ERRNO_IS_NOT_SUPPORTED(r);
+
+        if (test_xattrs) {
+                assert_se(oomd_get_cgroup_oom_preference(ctx, NULL) == 0);
+                assert_se(cg_set_xattr(SYSTEMD_CGROUP_CONTROLLER, cgroup, "user.oomd_omit", "1", 1, 0) >= 0);
+                assert_se(cg_set_xattr(SYSTEMD_CGROUP_CONTROLLER, cgroup, "user.oomd_avoid", "1", 1, 0) >= 0);
+        }
+
+        /* omit takes precedence over avoid when both are set to true */
+        if (test_xattrs) {
```
anitazha: nit: can combine these `if` statements

anitazha (approved):
lgtm. minor nit

src/oom/test-oomd-util.c
enr0n: Thanks, fixed!

src/oom/oomd-util.c
```diff
@@ -142,10 +143,49 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
...
+        r = cg_get_owner(SYSTEMD_CGROUP_CONTROLLER, prefix, &prefix_uid);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to get owner/group from %s: %m", ctx->path);
+
+        if (uid == prefix_uid) {
+                /* Ignore most errors when reading the xattr since it is usually unset and cgroup xattrs are only used
+                 * as an optional feature of systemd-oomd (and the system might not even support them). */
+                r = cg_get_xattr_bool(SYSTEMD_CGROUP_CONTROLLER, ctx->path, "user.oomd_avoid");
+                if (r == -ENOMEM)
+                        return r;
```
poettering: i think it would be good to debug log about unexpected errors, i.e. everything except for ENODATA

src/oom/oomd-util.c
```diff
@@ -142,10 +143,49 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
...
+                ctx->preference = r == 1 ? MANAGED_OOM_PREFERENCE_AVOID : ctx->preference;
+
+                r = cg_get_xattr_bool(SYSTEMD_CGROUP_CONTROLLER, ctx->path, "user.oomd_omit");
+                if (r == -ENOMEM)
```
poettering, line 177: ditto

src/oom/oomd-util.c
```diff
@@ -142,10 +143,49 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
...
+        if (r < 0)
+                return log_debug_errno(r, "Failed to get owner/group from %s: %m", ctx->path);
+
+        if (uid == prefix_uid) {
+                /* Ignore most errors when reading the xattr since it is usually unset and cgroup xattrs are only used
+                 * as an optional feature of systemd-oomd (and the system might not even support them). */
+                r = cg_get_xattr_bool(SYSTEMD_CGROUP_CONTROLLER, ctx->path, "user.oomd_avoid");
+                if (r == -ENOMEM)
+                        return r;
+                ctx->preference = r == 1 ? MANAGED_OOM_PREFERENCE_AVOID : ctx->preference;
```
poettering: doesn't matter, but I think it would be nicer to do this check as `r > 0` rather than `r == 1`.

When doing "ternary" booleans, i.e. booleans that are true or false, or indicate some error, we say > 0 means true, == 0 means false and < 0 means error. Hence to be systematic here it would be good to do write the check above as > 0, even if of course 1 is the only relevant value here.

src/oom/oomd-util.c
```diff
@@ -142,10 +143,49 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
         return (ctx->swap_total - ctx->swap_used) < swap_threshold;
 }
 
+int oomd_get_cgroup_oom_preference(OomdCGroupContext *ctx, const char *prefix) {
```
poettering: maybe call this "fetch" rather than "get", after all it's not a traditional "getter" which returns the value as ret val or ret param, but instead just writes it into the context object.

poettering:
(my comments dont matter much, I am aware the points I raised are already existent in the code before this commit — and mostly bikeshedding – but it would still be good to get this fixed)

enr0n:
> (my comments dont matter much, I am aware the points I raised are already existent in the code before this commit — and mostly bikeshedding – but it would still be good to get this fixed)

Thanks for your review. I believe I addressed all of your comments in the last push.

src/oom/oomd-util.c
```diff
@@ -142,10 +143,53 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
...
+                return log_debug_errno(r, "Failed to get owner/group from %s: %m", ctx->path);
+
+        if (uid == prefix_uid) {
+                /* Ignore most errors when reading the xattr since it is usually unset and cgroup xattrs are only used
+                 * as an optional feature of systemd-oomd (and the system might not even support them). */
+                r = cg_get_xattr_bool(SYSTEMD_CGROUP_CONTROLLER, ctx->path, "user.oomd_avoid");
+                if (r == -ENOMEM)
+                        return r;
+                if (r < 0 && r != -ENODATA)
+                        log_debug_errno(r, "Failed to get xattr user.oomd_avoid: %m");
```
yuwata: Please add `", ignoring"` to the log message for ignored error.

src/oom/oomd-util.c
```diff
@@ -142,10 +143,53 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
...
+        r = cg_get_owner(SYSTEMD_CGROUP_CONTROLLER, prefix, &prefix_uid);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to get owner/group from %s: %m", ctx->path);
+
+        if (uid == prefix_uid) {
+                /* Ignore most errors when reading the xattr since it is usually unset and cgroup xattrs are only used
+                 * as an optional feature of systemd-oomd (and the system might not even support them). */
+                r = cg_get_xattr_bool(SYSTEMD_CGROUP_CONTROLLER, ctx->path, "user.oomd_avoid");
+                if (r == -ENOMEM)
+                        return r;
```
yuwata: Please use `log_oom_debug()`.

src/oom/oomd-util.c
```diff
@@ -142,10 +143,53 @@ bool oomd_swap_free_below(const OomdSystemContext *ctx, int threshold_permyriad)
...
+                if (r < 0 && r != -ENODATA)
+                        log_debug_errno(r, "Failed to get xattr user.oomd_avoid: %m");
+                ctx->preference = r > 0 ? MANAGED_OOM_PREFERENCE_AVOID : ctx->preference;
+
+                r = cg_get_xattr_bool(SYSTEMD_CGROUP_CONTROLLER, ctx->path, "user.oomd_omit");
+                if (r == -ENOMEM)
+                        return r;
+                if (r < 0 && r != -ENODATA)
+                        log_debug_errno(r, "Failed to get xattr user.oomd_omit: %m");
```
yuwata: Same here. Please use `log_oom_debug()` and append `", ignoring"`.

yuwata (approved):
LGTM. But several minor comments.

src/oom/oomd-util.c
enr0n: Done, thanks!

enr0n:
I think the remaining failure (focal amd64 autopkgtest) is unrelated.

emansom:
I usually have a ridicilious number of tabs both in Chromium and Firefox open, and would prefer `systemd-oomd` seeing that memory hog as a candidate to kill.

Current behavior yields a livelock eventually.

How would one go about configuring that?

PR #23765, opened 2022-06-16
meson: Reorder cflags for sd-boot

medhefgo:
Currently, -Wall and -Wextra override previously passed flags like
-Wno-unused-parameter. This reorders them to be passed before any
optional flags. -Wsign-compare is part of -Wextra and therefore dropped.
-nostdlib is a link-stage flag and dropped as it is already part of
efi_ldflags.


DaanDeMeyer (approved):

yuwata (approved):
LGTM.

PR #23766, opened 2022-06-16
test: send stdout/stderr of testsuite units to journal & console

mrc0mmand:
to make debugging in CIs slightly easier.

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -2941,6 +2941,18 @@ test_setup() {
             mask_supporting_services
         fi
 
+        # Send stdout/stderr of testsuite-*.service units to both journal and
+        # console to make debugging in CIs easier
+        # Note: we can't use a dropin for `testsuite-.service`, since that also
+        #       overrides 'sub-units' of some tests that already use a specific
+        #       value for Standard(Output|Error)=
+        #       (e.g. test/units/testsuite-66-deviceisolation.service)
+        if ! get_bool "$INTERACTIVE_DEBUG"; then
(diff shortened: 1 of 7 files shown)


DaanDeMeyer:
This seems super useful

test/test-functions
```diff
@@ -724,6 +724,16 @@ setup_basic_environment() {
...
+
+    # Send stdout/stderr of testsuite-*.service units to both journal and
+    # console to make debugging in CIs easier
+    # Note: we can't use a dropin for `testsuite-.service`, since that also
+    #       overrides 'sub-units' of some tests that already use a specific
+    #       value for Standard(Output|Error)=
+    #       (e.g. test/units/testsuite-66-deviceisolation.service)
+    local dropin_dir="${initdir:?}/etc/systemd/system/testsuite-${TESTID:?}.service.d"
+    mkdir -p "$dropin_dir"
+    printf '[Service]\nStandardOutput=journal+console\nStandardError=journal+console' >"$dropin_dir/99-stdout.conf"
```
evverx: I'm not sure it should be turned on by default in interactive mode. I think it would make sense to wrap it in `[[ ! get_bool "$INTERACTIVE_DEBUG" ]]` or something like that.

mrc0mmand: Good point! Fixed.

mrc0mmand:
Pushed another change moving the dropin creation from `setup_basic_environment()` to `test_setup()`, since `setup_basic_environment()` is called only when creating a new image, which might not be always true if the image is already created and cached.

yuwata (approved):
LGTM.

PR #23771, opened 2022-06-17
A couple of test-output-related tweaks

mrc0mmand:
Let's reduce the amount of noise on the console output from tests by keeping systemd debug logs in the journal and output only stdout/stderr from the testsuite-*.sh scripts. If a test fails, we have the respective journal at hand for further investigation, so in the end it should make the debugging bit easier by making the logs produced by CIs slightly more readable without the need to go through megabytes of debug logs first.


yuwata (approved):
LGTM. But Lint Code Base reveals several issues.

mrc0mmand:
Added a couple more commits, since the changes broke `TEST-52-HONORFIRSTSHUTDOWN`.

yuwata (approved):

PR #23773, opened 2022-06-17
Deduplicate linking of code under src/basic and src/libsystemd/

keszybz:


src/libsystemd/sd-netlink/netlink-message.c
```diff
@@ -524,19 +524,19 @@ _public_ int sd_netlink_message_append_ether_addr(sd_netlink_message *m, unsigne
         return 0;
 }
 
-int netlink_message_append_hw_addr(sd_netlink_message *m, unsigned short type, const struct hw_addr_data *data) {
+_public_ int sd_netlink_message_append_hw_addr(sd_netlink_message *m, unsigned short type, const uint8_t *data, size_t length) {
```
yuwata: I do not think this has worth to be exported. Users can be use `sd_netlink_message_{append,read}_data()` instead.

src/libsystemd/sd-netlink/netlink-util.h
```diff
@@ -135,7 +135,21 @@ static inline int netlink_message_append_hw_addr(sd_netlink_message *m, unsigned
...
+                sd_netlink_message *m,
+                unsigned short type,
+                int family,
+                const union in_addr_union *data) {
+        switch (family) {
+        case AF_INET:
+                return sd_netlink_message_append_in_addr(m, type, &data->in);
+        case AF_INET6:
+                return sd_netlink_message_append_in6_addr(m, type, &data->in6);
+        default:
```
yuwata, line 148: How about to simply call `netlink_message_append_addr(m, type, family, data->bytes)`?

src/libsystemd/sd-netlink/netlink-util.h
```diff
@@ -149,7 +163,16 @@ static inline int netlink_message_read_hw_addr(sd_netlink_message *m, unsigned s
         return r;
 }
 
-int netlink_message_read_in_addr_union(sd_netlink_message *m, unsigned short type, int family, union in_addr_union *data);
+static inline int netlink_message_read_in_addr_union(sd_netlink_message *m, unsigned short type, int family, union in_addr_union *data) {
+        switch (family) {
+        case AF_INET:
+                return sd_netlink_message_read_in_addr(m, type, &data->in);
+        case AF_INET6:
+                return sd_netlink_message_read_in6_addr(m, type, &data->in6);
```
yuwata, line 184: Also here. `netlink_message_read_addr(m, type, family, data->bytes)` is enough.

src/libsystemd/sd-netlink/netlink-util.h
```diff
@@ -150,7 +150,16 @@ static inline int netlink_message_append_in_addr_union(
         }
 }
 
-int netlink_message_append_sockaddr_union(sd_netlink_message *m, unsigned short type, const union sockaddr_union *data);
+static inline int netlink_message_append_sockaddr_union(sd_netlink_message *m, unsigned short type, const union sockaddr_union *data) {
+        switch (data->sa.sa_family) {
+        case AF_INET:
+                return sd_netlink_message_append_sockaddr_in(m, type, &data->in);
+        case AF_INET6:
+                return sd_netlink_message_append_sockaddr_in6(m, type, &data->in6);
```
yuwata, line 158: Also here.

src/libsystemd/libsystemd.sym
```diff
@@ -782,4 +782,173 @@ global:
 LIBSYSTEMD_252 {
 global:
         sd_hwdb_new_from_path;
+
+        sd_netlink_new_from_fd;
+        sd_netlink_open;
+        sd_netlink_open_fd;
+        sd_netlink_inc_rcvbuf;
```
yuwata: I'd like to rename `inc` -> `increase`.

src/libsystemd/libsystemd.sym
```diff
@@ -782,4 +782,173 @@ global:
...
+        sd_netlink_ref;
+        sd_netlink_unref;
+        sd_netlink_send;
+        sd_netlink_sendv;
```
yuwata: I do not think sendv() is ready to be exported. Now I am working on this, but I am not sure I can finish the work before v252.

src/libsystemd/libsystemd.sym
```diff
@@ -782,4 +782,173 @@ global:
...
+        sd_nfnl_message_batch_end;
+        sd_nfnl_nft_message_del_table;
+        sd_nfnl_nft_message_new_table;
+        sd_nfnl_nft_message_new_basechain;
+        sd_nfnl_nft_message_new_rule;
+        sd_nfnl_nft_message_new_set;
+        sd_nfnl_nft_message_new_setelems_begin;
+        sd_nfnl_nft_message_del_setelems_begin;
+        sd_nfnl_nft_message_add_setelem;
+        sd_nfnl_nft_message_add_setelem_end;
```
yuwata: I think these are not stable yet. I am refactoring them now.

yuwata (changes requested):

src/libsystemd/sd-netlink/netlink-util.h
keszybz: The goal is to let sd-message.c only export `_public_` functions. The public interface only has `_append_in_addr()` and `_append_in6_addr()`.

src/libsystemd/sd-netlink/netlink-util.h
keszybz: Same as above.

src/libsystemd/sd-netlink/netlink-message.c
keszybz: Hmm, but then we lose the check for `message_attribute_has_type(m, NULL, type, NETLINK_TYPE_IN_ADDR)`. Isn't this check useful?

src/libsystemd/libsystemd.sym
keszybz: Hmm, if https://github.com/systemd/systemd/pull/23774 goes in, I think those won't be used. Otherwise this becomes quite messy to unexport them.

keszybz:
Updated:
- sd_nfnl_*() and sd_netlink_sendv() are unexported again
- I reworked the descriptions of commits to explain why the helper functions in `netlink-util.h` are the way they are

src/libsystemd/sd-netlink/netlink-message.c
keszybz: OK, I reworked this as suggested. The new commit title is 'sd-netlink: drop netlink_message_{append,read}_hw_addr'.

src/libsystemd/sd-netlink/netlink-util.h
```diff
@@ -131,11 +131,25 @@ int rtnl_log_create_error(int r);
...
+
 int netlink_message_append_in_addr_union(sd_netlink_message *m, unsigned short type, int family, const union in_addr_union *data);
 int netlink_message_append_sockaddr_union(sd_netlink_message *m, unsigned short type, const union sockaddr_union *data);
 
-int netlink_message_read_hw_addr(sd_netlink_message *m, unsigned short type, struct hw_addr_data *data);
+static inline int netlink_message_read_hw_addr(sd_netlink_message *m, unsigned short type, struct hw_addr_data *data) {
+        int r;
+
+        assert(data);
+        r = sd_netlink_message_read(m, type, sizeof(data->bytes), data->bytes);
```
yuwata: Missing empty line between assertion and actual code?

src/libsystemd/sd-netlink/netlink-message.c
```diff
@@ -452,31 +452,31 @@ _public_ int sd_netlink_message_append_data(sd_netlink_message *m, unsigned shor
         return 0;
 }
 
-int netlink_message_append_in_addr_union(sd_netlink_message *m, unsigned short type, int family, const union in_addr_union *data) {
+static int _netlink_message_append_in_addr_union(sd_netlink_message *m, unsigned short type, int family, const union in_addr_union *addr) {
```
yuwata, line 363: We usually use `_internal` or `_impl` instead of prefixing `_`.

src/libsystemd/sd-bus/bus-dump.c
```diff
@@ -46,14 +46,12 @@ static char *indent(unsigned level, uint64_t flags) {
         return p;
 }
 
-_public_ int sd_bus_message_dump(sd_bus_message *m, FILE *f, uint64_t flags) {
+int _bus_message_dump(sd_bus_message *m, FILE *f, uint64_t flags) {
```
yuwata, line 49: Also here, `_internal()` or `_impl()`, rather than prefixing with `_`.

TODO
```diff
@@ -1046,6 +1046,10 @@ Features:
```
yuwata, line 1046: typo in commit message: libystemd -> libsystemd

Also, the commit message contains comment about conflict on rebase?

meson.build
```diff
@@ -1902,29 +1902,27 @@ basic_includes = include_directories(
...
-        'src/libsystemd/sd-hwdb',
-        'src/libsystemd/sd-id128',
-        'src/libsystemd/sd-journal',
-        'src/libsystemd/sd-netlink',
-        'src/libsystemd/sd-network',
-        'src/libsystemd/sd-resolve')]
+libsystemd_includes = [
+        basic_includes,
+        include_directories(
+                'src/libsystemd/internal',
```
yuwata, line 1908: Hm, I cannot understand why moving type 3 functions to libsystemd/internal can save file size...
And, I am not a fun of the idea. libsystemd/internal now has variety of codes...
Moreover, e.g. sd_device_monitor and sd_device_enumerator are exported but moved to internal. That's not intuitive and confusing for me...

yuwata (changes requested):
CIs are unhappy with the change.

bluca:
@mbiebl with this change, we'll have to make libsystemd0 depend on libsystemd-shared (which as a very large set of dependencies), add a new single-library package libsystemd-core, or put this new internal library inside libsystemd0 (which breaks the best practice of one shared object per library package) - I'd tend toward the last one

mbiebl:
I would prefer to keep the dependencies of libsystemd minimal. We have stuff like apt linking against it

mbiebl:
Ie libsystemd should not link against libsystemd-core

mbiebl:
This is a NAK from my part

bluca:
> Ie libsystemd should not link against libsystemd-core/basic

We can ship this new basic lib in libsystemd0 - not great, but not terrible either, pretty much the same amount of code as before, just split into two objects

mbiebl:
The standalone variants/builds of systemd-tmpfiles and systemd-sysusers are broken by this change.

They started to link against libsystemd-basic without setting an RPATH
```
$ objdump -x systemd-sysusers  | grep NEEDED
  NEEDED               libsystemd-basic-251.so
  NEEDED               libcrypt.so.1
  NEEDED               libselinux.so.1
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2
$ objdump -x systemd-sysusers  | grep PATH
```
This could be fixed by installing an RPATH but I think the standalone variants should have minimal dependencies and as such not link against libsystemd-basic (or libsystemd)

src/libsystemd/sd-netlink/netlink-message.c
keszybz: We have two conventions in place… I think that `_` is much better: it conveys pretty much the same information ("internal" or "private" function") but is much less verbose. This additional verbosity brings no benefit that I can see, so I prefer the short&sweet.

src/libsystemd/sd-netlink/netlink-util.h
keszybz: Fixed.

mbiebl:
```
$ objdump -x usr/lib/x86_64-linux-gnu/libsystemd.so.0 | grep NEEDED
  NEEDED               libsystemd-basic-251.so
  NEEDED               librt.so.1
  NEEDED               libgcrypt.so.20
  NEEDED               liblzma.so.5
  NEEDED               libzstd.so.1
  NEEDED               liblz4.so.1
  NEEDED               libpthread.so.0
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2
$ objdump -x usr/lib/x86_64-linux-gnu/libsystemd.so.0 | grep PATH
```
libsystemd links against libsystemd-basic-251.so but doesn't set a RUNPATH/RPATH.

Wouldn't this break all (non-systemd) applications linking against libsystemd ?

meson.build
keszybz: > Hm, I cannot understand why moving type 3 functions to libsystemd/internal can save file size... 

No, the moving of functions isn't supposed to save file size. Size is saved because we stop installing certain compiled objects twice. Namely, the public (`_public_`) functions that form libsystemd. The moving of (other) functions to src/libsystemd/internal is done so we can link them into *both* libsystemd and non-libsystemd-parts without pulling in libsystemd into the non-systemd parts. (Or in other words, src/libsystemd/internal now is used like src/libsystemd was used as a whole previously.)

I tried to express this in the commit message. I think the text *does* explain this, but I'm open to suggestions.

> Moreover, e.g. sd_device_monitor and sd_device_enumerator are exported but moved to internal.

That was an error in the patch. `device-monitor.c` and `device-enumerator.c` are now back under `src/libsystemd/sd-device/` and some of the code that was there is in `src/libsystemd/internal/device-internal.c`.

> And, I am not a fun of the idea. libsystemd/internal now has variety of codes...  That's not intuitive and confusing for me...

This is generally how the rest of shared code is organized. `src/fundamental`, `src/basic`, and `src/shared` all consist of code that has a variety of purposes, grouped by what's it's for, not what it does. I'm not trying to say that the split between files, or the file naming, or ordering of things within files is ideal. I mostly tried to minimize changes, e.g. keeping the same order when moving text between files, but I'm sure this could be improved. But this kind of cleanup (especially reordering within files or renaming of files) can be easily done later.

keszybz:
> @mbiebl with this change, we'll have to make libsystemd0 depend on libsystemd-shared (which as a very large set of dependencies), add a new single-library package libsystemd-core, or put this new internal library inside libsystemd0 (which breaks the best practice of one shared object per library package) - I'd tend toward the last one

> We can ship this new basic lib in libsystemd0 - not great, but not terrible either, pretty much the same amount of code as before, just split into two objects

Yeah, putting the new file together with `libsystemd.so` seems reasonable.

keszybz:
> libsystemd links against libsystemd-basic-251.so but doesn't set a RUNPATH/RPATH.

Indeed, it should. This is fixed now.

> The standalone variants/builds of systemd-tmpfiles and systemd-sysusers are broken by this change.

They are now linked statically (again).

OK, I think I fixed all the issues that were pointed out. There was a bit of churn, so the measurements that are in the commit messages might be a bit off. I'll redo them once the series looks like it's approaching the final form.

Note: I renamed `libbasic_shared → libbasic`, `libbasic → libbasic_static`. This fits better with `libshared` and `libshared_static`.

mrc0mmand:
networkd seems to be unhappy:

```
[  963.648063] systemd-networkd[70]: Failed to increase receive buffer size for general netlink socket, ignoring: Operation not permitted
[  963.653707] systemd[1]: systemd-networkd.socket: Incoming traffic
[  963.654206] systemd[1]: systemd-networkd.socket: Changed listening -> running
[  963.654369] systemd-networkd[70]: Assertion 'data' failed at src/libsystemd/internal/netlink-util.h:167, function int netlink_message_read_hw_addr(sd_netlink_message *, unsigned short, struct hw_addr_data *)(). Aborting.
[  963.654520] systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1/unit/systemd_2dnetworkd_2esocket interface=org.freedesktop.DBus.Properties member=PropertiesChanged cookie=121 reply_cookie=0 signature=sa{sv}as error-name=n/a error-me>
[  963.654659] systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1/unit/systemd_2dnetworkd_2esocket interface=org.freedesktop.DBus.Properties member=PropertiesChanged cookie=122 reply_cookie=0 signature=sa{sv}as error-name=n/a error-me>
[  963.733985] systemd[1]: systemd-journald.service: Received EPOLLHUP on stored fd 3 (stored), closing.
[  963.734499] systemd[1]: Received SIGCHLD from PID 70 (4).
[  963.734686] systemd[1]: Child 70 (4) died (code=dumped, status=6/ABRT)
[  963.734894] systemd[1]: systemd-networkd.service: Child 70 belongs to systemd-networkd.service.
[  963.735048] systemd[1]: systemd-networkd.service: Main process exited, code=dumped, status=6/ABRT
[  963.735639] systemd[1]: systemd-networkd.service: Failed with result 'core-dump'.
```

src/libsystemd/sd-netlink/netlink-message.c
yuwata: OK. That's not strong opinion.

src/analyze/analyze-critical-chain.c
```diff
@@ -3,7 +3,7 @@
 #include "analyze-critical-chain.h"
 #include "analyze-time-data.h"
 #include "analyze.h"
-#include "bus-error.h"
+#include "bus-error-util.h"
```
yuwata, line 6: I'd like that this renaming is done in a different commit, as this commit is one of the most important commits in this PR. Making this commit smaller is better for reviewing this PR.

src/libsystemd/meson.build
```diff
@@ -82,80 +124,50 @@ sd_login_sources = files('sd-login/sd-login.c')
...
         'sd-netlink/netlink-message.c',
         'sd-netlink/netlink-slot.c',
         'sd-netlink/netlink-slot.h',
         'sd-netlink/netlink-socket.c',
-        'sd-netlink/netlink-types-genl.c',
-        'sd-netlink/netlink-types-internal.h',
-        'sd-netlink/netlink-types-nfnl.c',
-        'sd-netlink/netlink-types-rtnl.c',
-        'sd-netlink/netlink-types.c',
-        'sd-netlink/netlink-types.h',
```
yuwata, line 148: Let me confirm that files in src/libsystemd/internal are build twice for libsystemd and libsystemd-shared, right?
And, files in src/libsystemd except for those in internal are only build for libsystemd?

If my understanding is correct, then these netlink-types*.[ch] should be in libsystemd/sd-netlink, as IIRC, these are not directly used by files in shared or other executables.

yuwata (changes requested):

mbiebl:
I did a test build of the Debian package with this PR applied to get an idea about the space savings and what kind of changes it would cause:
```
$ debdiff --show-moved main/systemd_251.2-6_amd64.changes full-split/systemd_251.2-6_amd64.changes | colordiff
[The following lists of changes regard files as different if they have
different names, permissions or owners.]

New files in second set of .debs, found in package libsystemd-dev
-----------------------------------------------------------------
-rw-r--r--  root/root   /usr/include/systemd/sd-netlink.h
lrwxrwxrwx  root/root   /usr/share/man/man3/sd_bus_error_setfv.3.gz -> sd_bus_error.3.gz
lrwxrwxrwx  root/root   /usr/share/man/man3/sd_bus_message_read_strv_extend.3.gz -> sd_bus_message_read_strv.3.gz
lrwxrwxrwx  root/root   /usr/share/man/man3/sd_hwdb_new_from_path.3.gz -> sd_hwdb_new.3.gz
lrwxrwxrwx  root/root   /usr/share/man/man3/sd_id128_string_equal.3.gz -> sd-id128.3.gz

New files in second set of .debs, found in package libsystemd0
--------------------------------------------------------------
-rw-r--r--  root/root   /usr/lib/x86_64-linux-gnu/systemd/libsystemd-basic-251.so


Control files of package libnss-myhostname: lines which differ (wdiff format)
-----------------------------------------------------------------------------
Installed-Size: [-288-] {+292+}

No differences were encountered between the control files of package libnss-mymachines

Control files of package libnss-resolve: lines which differ (wdiff format)
--------------------------------------------------------------------------
Installed-Size: [-272-] {+276+}

Control files of package libnss-systemd: lines which differ (wdiff format)
--------------------------------------------------------------------------
Installed-Size: [-453-] {+457+}

No differences were encountered between the control files of package libpam-systemd

Control files of package libsystemd-dev: lines which differ (wdiff format)
--------------------------------------------------------------------------
Installed-Size: [-1148-] {+1170+}

Control files of package libsystemd-shared: lines which differ (wdiff format)
-----------------------------------------------------------------------------
Depends: libacl1 (>= 2.2.23), libapparmor1 (>= 2.13), libaudit1 (>= 1:2.2.1), libblkid1 (>= 2.24), libc6 (>= 2.33), [-libcap2 (>= 1:2.10),-] libcrypt1 (>= 1:4.4.0), libgcrypt20 (>= 1.10.0), libip4tc2 (>= 1.8.3), libkmod2 (>= 5~), liblz4-1 (>= 0.0~r130), liblzma5 (>= 5.1.1alpha+20120614), libmount1 (>= 2.30), libpam0g (>= 0.99.7.1), libseccomp2 (>= 2.4.1), libselinux1 (>= 3.1~), libssl3 (>= 3.0.0), {+libsystemd0 (= 251.2-6),+} libzstd1 (>= 1.5.2)
Installed-Size: [-5154-] {+3990+}

Control files of package libsystemd0: lines which differ (wdiff format)
-----------------------------------------------------------------------
Installed-Size: [-981-] {+1661+}

No differences were encountered between the control files of package libudev-dev

Control files of package libudev1: lines which differ (wdiff format)
--------------------------------------------------------------------
Installed-Size: [-304-] {+308+}

Control files of package libudev1-udeb: lines which differ (wdiff format)
-------------------------------------------------------------------------
Installed-Size: [-173-] {+177+}

Control files of package systemd: lines which differ (wdiff format)
-------------------------------------------------------------------
Installed-Size: [-9756-] {+9772+}

Control files of package systemd-boot: lines which differ (wdiff format)
------------------------------------------------------------------------
Depends: libc6 (>= 2.33), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} systemd-boot-efi (= 251.2-6)

No differences were encountered between the control files of package systemd-boot-efi

Control files of package systemd-container: lines which differ (wdiff format)
-----------------------------------------------------------------------------
Depends: libacl1 (>= 2.2.23), libbz2-1.0, libc6 (>= 2.33), libcurl4 (>= 7.16.3), liblzma5 (>= 5.1.1alpha+20120614), libseccomp2 (>= 2.3.1), libselinux1 (>= 3.1~), libssl3 (>= 3.0.0), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} zlib1g (>= 1:1.1.4), systemd, default-dbus-system-bus | dbus-system-bus

Control files of package systemd-coredump: lines which differ (wdiff format)
----------------------------------------------------------------------------
Depends: libc6 (>= 2.33), liblz4-1 (>= 0.0~r130), liblzma5 (>= 5.1.1alpha+20120614), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} libzstd1 (>= 1.5.2), adduser, systemd

Control files of package systemd-homed: lines which differ (wdiff format)
-------------------------------------------------------------------------
Depends: libblkid1 (>= 2.24), libc6 (>= 2.33), libcap2 (>= 1:2.10), libfdisk1 (>= 2.33), libpam0g (>= 1.1.1), libssl3 (>= 3.0.0), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} systemd-userdbd (= 251.2-6), systemd (= 251.2-6), libpam-runtime
Installed-Size: [-1129-] {+1133+}

Control files of package systemd-journal-remote: lines which differ (wdiff format)
----------------------------------------------------------------------------------
Depends: libc6 (>= 2.33), libcurl4 (>= 7.16.2), libmicrohttpd12 (>= 0.9.50), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} systemd, adduser

Control files of package systemd-oomd: lines which differ (wdiff format)
------------------------------------------------------------------------
Depends: libc6 (>= 2.4), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} adduser

No differences were encountered between the control files of package systemd-standalone-sysusers

No differences were encountered between the control files of package systemd-standalone-tmpfiles

No differences were encountered between the control files of package systemd-sysv

Control files of package systemd-tests: lines which differ (wdiff format)
-------------------------------------------------------------------------
Installed-Size: [-20294-] {+19555+}

Control files of package systemd-timesyncd: lines which differ (wdiff format)
-----------------------------------------------------------------------------
Depends: libc6 (>= 2.33), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} adduser, systemd

Control files of package systemd-userdbd: lines which differ (wdiff format)
---------------------------------------------------------------------------
Depends: libc6 (>= 2.10), libsystemd-shared (= 251.2-6), {+libsystemd0 (= 251.2-6),+} systemd (= 251.2-6)

No differences were encountered between the control files of package udev

No differences were encountered between the control files of package udev-udeb
```

libsytemd-shared: Installed-Size: [-5154-] {+3990+}
libsystemd0: Installed-Size: [-981-] {+1661+}

So the space savings is about 480K

mbiebl:
As this didn't result in additional dependencies for the libsystemd0 package, I retract my NAK.
Shipping libsystemd-basic within the libsystemd0 package is a bit meh, but I can live with that.

src/analyze/analyze-critical-chain.c
keszybz: Actually, it's not a rename. I'm splitting `bus-error.h` into two files: `bus-error.h` and `bus-error-util.h`. The include is changed because the files outside of `src/libsystemd/` need to include the new header. It wouldn't make sense to make this split without the other changes.

src/libsystemd/meson.build
keszybz: > Let me confirm that files in src/libsystemd/internal are build twice for libsystemd and libsystemd-shared, right? And, files in src/libsystemd except for those in internal are only build for libsystemd?

Yes.

> If my understanding is correct, then these netlink-types*.[ch] should be in libsystemd/sd-netlink, as IIRC, these are not directly used by files in shared or other executables.

"directly" is not relevant. What matters whether they are used *only* through `_public_` functions, in which case it is enough to link to `libsystemd`, or whether they are used by directly or through another non-public function. In this case, `netlink_seal_message` calls `message_seal`, and `nfnl_message_batch` which calls `message_new` which calls `type_system_root_get_type_system_and_header_size`, and then the dependency tree cascades. So maybe it'd be possible to split this better, but I think it'll require a rethinking of how `netlink-types*.[ch]` are supposed to be used. We can always do that in a later PR.

keszybz:
I'll close this PR. I did a build with Fedora rpms similar to what @mbiebl did above with debs, and the savings are rather small. So I don't think it's worth the hassle. I'll push the other patches as a new pull request, but I'll drop the creation of `libsystemd-basic.so` and linking changes. Maybe we can return to this at some later point if we have a better understanding why the deduplication of code does not lead to the expected space savings.

PR #12633, opened 2019-05-22
networkctl: add support to display Transmit/Recieve queue length

ssahani:
```
(networkctl)⚡ % build/networkctl status veth-test                                                                                                                            ~/tt/networkctl
● 13: veth-test
            Link File: /usr/lib/systemd/network/99-default.link
         Network File: /usr/lib/systemd/network/veth0.network
                 Type: ether
                State: routable (configured)
               Driver: veth
           HW Address: 8a:a6:1c:3f:a6:1a
                  MTU: 1500
          Minimum MTU: 68
          Maximum MTU: 65535
Transmit Queue Length: 1
 Receive Queue Length: 1
              Address: 192.168.5.31
                       fe80::88a6:1cff:fe3f:a61a
              Gateway: 192.168.5.1
                  DNS: 192.168.5.1
                  NTP: 192.168.5.1
            Time Zone: Asia/Kolkata
         Connected To: Zeus on port peer-test
```

PR #23775, opened 2022-06-17
networkctl: fix output of "status": replace "Queue Length" by "Number of Queues"

marckleinebudde:
Commit 0307afc681e1 ("networkctl: add support to display Transmit/Recieve queue length (#12633)") added the display of the number of RX and TX Queues to the output of `networkctl status $DEV`. However the row description says "Queue Length".

This patch fixes the output by replacing "Queue Length" by "Number of Queues".

Fixes: 0307afc681e1 ("networkctl: add support to display Transmit/Recieve queue length #12633)")


marckleinebudde:
FYI: @ssahani

yuwata (approved):
Thanks. LGTM.

PR #23790, opened 2022-06-21
Add sys/file.h for LOCK_

landgraf:
Fixes
| ../git/src/shared/dissect-image.c: In function 'mount_image_privately_interactively':
| ../git/src/shared/dissect-image.c:2986:34: error: 'LOCK_SH' undeclared (first use in this function)
|  2986 |         r = loop_device_flock(d, LOCK_SH);
|       |                                  ^~~~~~~

### DIFF
--- a/src/core/namespace.c
+++ b/src/core/namespace.c
@@ -4,6 +4,7 @@
 #include <linux/loop.h>
 #include <sched.h>
 #include <stdio.h>
+#include <sys/file.h>
 #include <sys/mount.h>
 #include <unistd.h>
 #include <linux/fs.h>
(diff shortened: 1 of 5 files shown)


src/core/namespace.c
```diff
@@ -7,6 +7,7 @@
 #include <sys/mount.h>
 #include <unistd.h>
 #include <linux/fs.h>
+#include <sys/file.h>
```
yuwata: Please sort headers.

yuwata (changes requested):
Is this for supporting musl or other non-glibc? If so, please mention that in the commit message.

src/core/namespace.c
landgraf: Should be fixed now

yuwata (approved):
LGTM.

bluca (approved):

PR #23791, opened 2022-06-21
Several dfuzzer-related follow-ups

mrc0mmand:


test/units/testsuite-21.sh
```diff
@@ -8,22 +8,16 @@ set -o pipefail
...
+    set +e
+    # We have to call the end.service/poweroff explicitly even if it's specified on
     # the kernel cmdline via systemd.wants=end.service, since dfuzzer calls
     # org.freedesktop.systemd1.Manager.ClearJobs() which drops the service
     # from the queue
-    [[ $SHUTDOWN_AT_EXIT -ne 0 ]] && systemctl start --job-mode=flush end.service
+    if [[ $SHUTDOWN_AT_EXIT -ne 0 ]]; then
+        systemctl poweroff
+        sleep 15
+        systemctl poweroff -ff
```
evverx: I think it can be replaced with something like
```
if ! systemctl poweroff; then
    # PID1 is down let's try to save the journal
    journalctl --sync || true # journal can be down as well so let's ignore exit codes here
    systemctl -ff poweroff # sync() and reboot(RB_POWER_OFF)
fi
```
It should probably cover the worst scenario where `dbus` is down, `systemd` is frozen and `journal` isn't responsive (hopefully it shouldn't happen very often these days)

mrc0mmand:
As mentioned in dbus-fuzzer/dfuzzer#106, with this PR and dbus-fuzzer/dfuzzer#105 the test now seems to be stable and vastly faster (now it's down to ~10 minutes from the original 30 - 45 minutes under sanitizers). I'll do several other tests, mainly running both the clang and gcc jobs in a loop to check if it's indeed stable, and then https://github.com/systemd/systemd-centos-ci/pull/506 should be finally ready to go.

test/units/testsuite-21.sh
```diff
@@ -8,22 +8,16 @@ set -o pipefail
...
-    # TODO: check fo ASan/UBSan errors
-    local found_cd=0
-    while read -r exe; do
-        coredumctl info "$exe"
-        found_cd=1
-    done < <(coredumpctl -F COREDUMP_EXE | sort -u)
-    [[ $found_cd -eq 0 ]] || exit 1
-
-    # We have to call the end.service explicitly even if it's specified on
+    set +e
```
evverx, line 11: I think either this or `|| :` below could be removed. I'd keep `|| :` and drop `set +e`

evverx: It can be done later though. I'll go ahead and merge the PR to make it possible to finally turn the test on.

evverx (approved):
LGTM. Thanks!

mrc0mmand:
I just merged https://github.com/systemd/systemd-centos-ci/pull/506 so the TEST-21-DFUZZER should be now a part of CI :-)

mrc0mmand:
Also, I should probably mention that it won't have any effect on the test coverage (which is, I guess, a good thing overall), since the coverage run runs without sanitizers and TEST-21-DFUZZER explicitly requires sanitizers.

evverx:
> I guess, a good thing overall

Agreed. If it was included in the coverage report it would be hard to tell what's actually tested. Ideally stuff related to fuzzing should go to a separate coverage report by analogy with https://storage.googleapis.com/oss-fuzz-coverage/systemd/reports/20220621/linux/index.html but I'm not sure at this point it's worth it. Maybe if there was `vfuzzer` separate fuzzing reports could be considered :-)

PR #23792, opened 2022-06-21
core: introduce env variable that makes user manager to not enumerate  devices

msekletar:
Not tracking state of device and mount units can save significant amount
of memory on systems where there is a lot of devices and a lot of users
logged in at the same time.

Assumption is that in vast majority of cases a user unit files don't
actually reference specific devices or mounts. Rather user units assume
that system is fully booted.

### DIFF
--- a/docs/ENVIRONMENT.md
+++ b/docs/ENVIRONMENT.md
@@ -274,6 +274,15 @@ All tools:
   it is either set to `system` or `user` depending on whether the NSS/PAM
   module is called by systemd in `--system` or `--user` mode.
 
+* `$SYSTEMD_SUPPORT_DEVICE`, `$SYSTEMD_SUPPORT_MOUNT`, `$SYSTEMD_SUPPORT_SWAP` -
+  can be set to `0` to mark respective unit type as unsupported. Generally,
+  having less units saves system resources so these options might be useful
+  for cases where we don't need to track given unit type, e.g. `--user` manager
+  often doesn't need to deal with device or swap units because they are
+  handled by the `--system` manager (PID 1). Note that setting certain unit
+  type as unsupported may not prevent loading some units of that type if they
+  are referenced by other units of another supported type.
+
 `systemd-remount-fs`:
 
 * `$SYSTEMD_REMOUNT_ROOT_RW=1` — if set and no entry for the root directory
--- a/src/core/unit.c
+++ b/src/core/unit.c
@@ -22,6 +22,7 @@
 #include "dbus-unit.h"
 #include "dbus.h"
 #include "dropin.h"
+#include "env-util.h"
 #include "escape.h"
 #include "execute.h"
 #include "fd-util.h"
@@ -4781,9 +4782,26 @@ int unit_setup_dynamic_creds(Unit *u) {
 }
 
 bool unit_type_supported(UnitType t) {
+        static int8_t cache[_UNIT_TYPE_MAX] = {}; /* -1: disabled, 1: enabled: 0: don't know */
+        int r;
+
         if (_unlikely_(t < 0))
                 return false;
         if (_unlikely_(t >= _UNIT_TYPE_MAX))
+                return false;
+
+        if (cache[t] == 0) {
+                char *e;
+
+                e = strjoina("SYSTEMD_SUPPORT_", unit_type_to_string(t));
+
+                r = getenv_bool(ascii_strupper(e));
+                if (r < 0 && r != -ENXIO)
+                        log_debug_errno(r, "Failed to parse $%s, ignoring: %m", e);
+
+                cache[t] = r == 0 ? -1 : 1;
+        }
+        if (cache[t] < 0)
                 return false;
 
         if (!unit_vtable[t]->supported)


poettering:
This is a hack I am sure. It#s OK as env var I guess, but should not be advertised as feature. People should be able to rely on these unit types to work. And if peiople want to break things anyway, it might be OK to do so, but it should be hidden behind an env var i am sure.

docs/ENVIRONMENT.md
```diff
@@ -269,6 +269,13 @@ All tools:
   it is either set to `system` or `user` depending on whether the NSS/PAM
   module is called by systemd in `--system` or `--user` mode.
 
+* `$SYSTEMD_SKIP_DEV_ENUMERATE` - manager will skip enumerating device, mount
```
poettering: i don't like pointless abbreviations. SYSTEMD_IGNORE_DEVICES sounds like an ok name too, no need to abbreviate that as "dev"

src/core/mount.c
```diff
@@ -1891,6 +1892,9 @@ static void mount_enumerate(Manager *m) {
 
         assert(m);
 
+        if (MANAGER_IS_USER(m) && getenv_bool("SYSTEMD_SKIP_DEV_ENUMERATE") == 1)
+                return;
```
poettering: i'd drop the conditionalization on the --user mode. if its hidden between an env var, i see no reason not to let people shoot themselves in the foot.

and after all it might make sense to give people control over this in container environments in system mode or so.

poettering:
So, the unit types already have a .supported() vtable function, that for example disables swap and devices in containers and such.

I am pretty sure we should hook this new env var to that.

Specifically, have env vars SYSTEMD_SUPPORT_MOUNT=0, SYSTEMD_SUPPORT_DEVICE=0, SYSTEMD_SUPORT_SWAP=0 and so on, for all unit types really. (given this is a priv feature, i think it's ok to even disable service units with that, if people feel like that).

or in other words, add a generic check into unit_type_supported() for this, in unit.c

msekletar:
@poettering PTAL.

docs/ENVIRONMENT.md
```diff
@@ -272,6 +272,15 @@ All tools:
   it is either set to `system` or `user` depending on whether the NSS/PAM
   module is called by systemd in `--system` or `--user` mode.
 
+* `$SYSTEMD_SUPPORT_DEVICE, SYSTEMD_SUPPORT_MOUNT, SYSTEMD_SUPPORT_SWAP` -
```
poettering: Write as:

```
`$SYSTEMD_SUPPORT_DEVICE`, `$SYSTEMD_SUPPORT_MOUNT`, `$SYSTEMD_SUPPORT_SWAP` – …
```

i.e. an individual set of backticks, and prefixing `$` for each var

docs/ENVIRONMENT.md
```diff
@@ -272,6 +272,15 @@ All tools:
...
+  can be set to `0` to mark respective unit type as unsupported. Generally,
+  having less units saves system resources so these options might be useful
+  for cases where we don't need to track given unit type, e.g. `--user` manager
+  usually doesn't need to deal with device or swap units because they are
```
poettering: usually → often

docs/ENVIRONMENT.md
```diff
@@ -272,6 +272,15 @@ All tools:
...
+  handled by the `--system` manager (PID 1). Note that setting certain unit
+  type as unsupported may not prevent creating some units of that type if they
```
poettering: creating → loading

poettering:
I'd really implement this generically in unit_type_supported().

i.e. add something like this there:

```c
bool unit_type_supported(UnitType t) {
        static int8_t cache[_UNIT_TYPE_MAX] = {}; /* -1: disabled, 1: enabled: 0: don't know */

        if (_unlikely_(t < 0))
                return false;
        if (_unlikely_(t >= _UNIT_TYPE_MAX))
                return false;

        if (cache[t] == 0) { 
                j = strjoina("SYSTEMD_SUPPORT_", unit_type_to_string(t));

                r = getenv_bool(ascii_strupper(j));
                if (r < 0 && r != -ENXIO)
                        log_debug_errno(r, "Failed to parse $%s, ignoring: %m", j); 
                
                cache[t] = r == 0 ? -1 : 1;
        }
        if (cache[t] < 0)
                return false;

        if (!unit_vtable[t]->supported)
                return true;

        return unit_vtable[t]->supported();
}

msekletar:
Hmm, but if we implement changes in `unit_type_supported()` as suggested it would be possible to turn off support for vital unit types like targets and services. Do we **really** want this to be possible? I think we should rather have only subset of unit types for which we check the supported env var.

poettering:
I consider this a debugging feature, a hacker feature. I think for features like that it's really OK if you can shoot yourself in the foot.

I'd see this as a positive thing: things should fail cleanly if you for example disable support for services. If however things fail in a dirty way, hit asserts/segfaults and so on we have a problem. Now we have a sane way to test that

msekletar:
Fair enough, let's support disabling all of them but document only some of them. Rest as you say is for hackers to play with not for actual users.

msekletar:
@poettering updated as per your suggestion.

poettering:
lgtm. thanks!

yuwata (approved):

PR #23793, opened 2022-06-21
core: Fix memory leaks

steelman:
arg_early_core_pattern and arg_watchdog_device hold pointers to memory
allocated with strdup() (inside path_make_absolute_cwd). The memory needs
to be freed in reset_arguments() during reload rather than forgotten.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -2377,8 +2377,8 @@ static void reset_arguments(void) {
         arg_reboot_watchdog = 10 * USEC_PER_MINUTE;
         arg_kexec_watchdog = 0;
         arg_pretimeout_watchdog = 0;
-        arg_early_core_pattern = NULL;
-        arg_watchdog_device = NULL;
+        arg_early_core_pattern = mfree(arg_early_core_pattern);
+        arg_watchdog_device = mfree(arg_watchdog_device);
         arg_watchdog_pretimeout_governor = mfree(arg_watchdog_pretimeout_governor);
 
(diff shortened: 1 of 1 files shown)


src/core/main.c
```diff
@@ -141,9 +141,9 @@ static usec_t arg_runtime_watchdog;
 static usec_t arg_reboot_watchdog;
 static usec_t arg_kexec_watchdog;
 static usec_t arg_pretimeout_watchdog;
-static char *arg_early_core_pattern;
+static char *arg_early_core_pattern = NULL;
 static char *arg_watchdog_pretimeout_governor;
-static char *arg_watchdog_device;
+static char *arg_watchdog_device = NULL;
```
keszybz: Please drop this chunk: see the comment above in line 117.

keszybz (changes requested):
Looks OK, apart from the one thing.

src/core/main.c
steelman: Fixed. I've seen a commit that removed NULL assignments there but because it moved them to reset_arguments() I wasn't sure it is going to be fine (I am never 100% sure how automatic initialization works for each variable)

keszybz: static variables are initialized to 0. So initializing them to e.g. NULL is a cosmetic thing to make it clearer to the reader what the initialization value is. But here we are initializing them explicitly in a different place, so initialization was removed here.

bluca (approved):

PR #23801, opened 2022-06-21
resolved: allow user to configure DNS stream timeout

### LINKED ISSUE #22196: Persistent DNS connections
I use strict DNS-over-TLS ("on" mode), but I noticed systemd-resolved closes the connection after 10 seconds of inactivity.
It's too bad because it will then have to reconnect to satisfy the next cache-missed request.  This triples the time needed on my machine: 60ms to establish the TLS tunnel, and 30ms for any following requests in the 10s.

I'd like a config switch to make the outgoing DNS connections persistent, or configure the delay (currently hardcoded as `DNS_STREAM_ESTABLISHED_TIMEOUT_USEC`).

This features does not seems implemented in latest upstream nor git main.

bytesuji:
Implements configurable DNS timeout.

Closes #22196.

### DIFF
--- a/man/resolved.conf.xml
+++ b/man/resolved.conf.xml
@@ -306,6 +306,13 @@ DNSStubListenerExtra=udp:[2001:db8:0:f102::13]:9953</programlisting>
       </varlistentry>
 
       <varlistentry>
+        <term><varname>DNSStreamTimeout=</varname></term>
+        <listitem><para>Takes an integer argument. Specifies the delay in seconds after which <command>systemd-resolved
+        </command> will close an outgoing DNS stream connection. The default setting is 10 seconds.
+        </para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><varname>ReadEtcHosts=</varname></term>
         <listitem><para>Takes a boolean argument. If <literal>yes</literal> (the default),
         <command>systemd-resolved</command> will read <filename>/etc/hosts</filename>, and try to resolve
--- a/src/resolve/resolved-dns-stream.c
+++ b/src/resolve/resolved-dns-stream.c
@@ -454,7 +454,7 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
 
         /* If we did something, let's restart the timeout event source */
         if (progressed && s->timeout_event_source) {
-                r = sd_event_source_set_time_relative(s->timeout_event_source, DNS_STREAM_ESTABLISHED_TIMEOUT_USEC);
+                r = sd_event_source_set_time_relative(s->timeout_event_source, s->manager->dns_stream_timeout_usec);
                 if (r < 0)
                         log_warning_errno(errno, "Couldn't restart TCP connection timeout, ignoring: %m");
         }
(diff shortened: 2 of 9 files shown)


man/resolved.conf.xml
```diff
@@ -305,6 +305,13 @@ DNSStubListenerExtra=udp:[2001:db8:0:f102::13]:9953</programlisting>
         </para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>DNSStreamTimeout=</varname></term>
+        <listitem><para>Takes an integer argument. Specifies the delay in seconds after which <literal>systemd-resolved
```
yuwata: Please use command tag instead of literal.

src/resolve/test-resolved-stream.c
```diff
@@ -210,8 +210,8 @@ static int on_stream_complete_do_nothing(DnsStream *s, int error) {
 }
 
 static void test_dns_stream(bool tls) {
-        Manager manager = {};
-         _cleanup_(dns_stream_unrefp) DnsStream *stream = NULL;
+        Manager manager = {.dns_stream_timeout_usec = 10 * USEC_PER_SEC};
```
yuwata: Please add a space and trailing comma: `{ .dns_stream_timeout_usec = 10 * USEC_PER_SEC, }`

man/resolved.conf.xml
```diff
@@ -305,6 +305,13 @@ DNSStubListenerExtra=udp:[2001:db8:0:f102::13]:9953</programlisting>
         </para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>DNSStreamTimeout=</varname></term>
+        <listitem><para>Takes an integer argument. Specifies the delay in seconds after which <command>systemd-resolved
```
yuwata, line 310: "Takes a timespan. Specifies the delay after ...", as we support multiple units, e.g. 3 min.

Also, please mention that the stream will not be closed if "infinity" is specified.

src/resolve/resolved-dns-stream.c
```diff
@@ -454,7 +454,7 @@ static int on_stream_io(sd_event_source *es, int fd, uint32_t revents, void *use
 
         /* If we did something, let's restart the timeout event source */
         if (progressed && s->timeout_event_source) {
-                r = sd_event_source_set_time_relative(s->timeout_event_source, DNS_STREAM_ESTABLISHED_TIMEOUT_USEC);
+                r = sd_event_source_set_time_relative(s->timeout_event_source, s->manager->dns_stream_timeout_usec);
```
yuwata, line 457: The timeout may be infinity. Please conditionally add the timer event source in `dns_stream_new()` when the timeout is not `USEC_INFINITY`.

src/resolve/resolved-gperf.gperf
```diff
@@ -32,3 +32,4 @@ Resolve.ReadEtcHosts,              config_parse_bool,                    0,
 Resolve.ResolveUnicastSingleLabel, config_parse_bool,                    0,                   offsetof(Manager, resolve_unicast_single_label)
 Resolve.DNSStubListenerExtra,      config_parse_dns_stub_listener_extra, 0,                   offsetof(Manager, dns_extra_stub_listeners)
 Resolve.CacheFromLocalhost,        config_parse_bool,                    0,                   offsetof(Manager, cache_from_localhost)
+Resolve.DNSStreamTimeout,          config_parse_sec,                     0,                   offsetof(Manager, dns_stream_timeout_usec)
```
yuwata, line 35: I guess we should refuse too small value (say, < 1 sec ?) here.
Please introduce config_parse_stream_timeout() and refuse such values.

man/resolved.conf.xml
```diff
@@ -305,6 +305,13 @@ DNSStubListenerExtra=udp:[2001:db8:0:f102::13]:9953</programlisting>
...
+        </command> will close an outgoing DNS stream connection. The default setting is 10 seconds.
```
yuwata, line 311: Maybe, it is better to clarify that closing stream happens when no IO on the stream during the period.

Issue #23796, opened 2022-06-21
systemd: mount units fail with "Mount process finished, but there is no mount."


jcg190701:
**systemd version the issue has been seen with**
 > [root@fedora ~]# systemctl --version
systemd 250 (v250.3-8.fc36)
+PAM +AUDIT +SELINUX -APPARMOR +IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS 
+FIDO2 +IDN2 -IDN -IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK +XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified

**Used distribution**
 > Fedora release 36 (Thirty Six)

**Linux kernel version used** (`uname -a`)

 > [root@fedora ~]# uname -a
Linux fedora 5.17.5-300.fc36.x86_64 #1 SMP PREEMPT Thu Apr 28 15:51:30 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**CPU architecture issue was seen on**
 > x86_64

**Expected behaviour you didn't see**
 > mounts work

**Unexpected behaviour you saw**
 > 
 Mounting units fail with "Mount process finished, but there is no mount.". 
It's hard to reproduce with 'systemctl daemon-reload' after https://github.com/systemd/systemd/issues/10872 , but it's easy to reproduce with 'systemctl daemon-reexec'.
```
× root-test.mount - mount unit for test
     Loaded: loaded (/proc/self/mountinfo; static)
     Active: failed (Result: protocol) since Tue 2022-06-21 08:38:28 EDT; 4s ago
      Where: /root/test
       What: /dev/loop0
        CPU: 3ms

Jun 21 08:38:28 fedora systemd[1]: root-test.mount: Child 3065 belongs to root-test.mount.
Jun 21 08:38:28 fedora systemd[1]: root-test.mount: Mount process exited, code=exited, status=0/SUCCESS (success)
Jun 21 08:38:28 fedora systemd[1]: root-test.mount: Mount process finished, but there is no mount.
Jun 21 08:38:28 fedora systemd[1]: root-test.mount: Failed with result 'protocol'.
[...]

PR #23803, opened 2022-06-22
mount:check from_proc_self_mountinfo if deserialized_state is MOUNT_M…

jcg190701:
…OUNTING.

log messages:
Mount process finished, but there is no mount.
Failed with result 'protocol'.

if daemon-reexec is done in parallel with starting mount unit, it is
possible that things happen in following order and result in above failure.

1.the mount unit state changes to "mounting"
2.daemon-reexec saves the unit state
3.kernel completes the mount and /proc/self/mountinfo is updated
4.daemon-reexec restores the saved unit state, that is "mounting"
5.systemd notices the mount command has finished but the unit state
is still "mounting".

Fixes:#23796

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -27,9 +27,9 @@ jobs:
         env:
           - { COMPILER: "gcc",   COMPILER_VERSION: "11", LINKER: "bfd",  CRYPTOLIB: "gcrypt"  }
           - { COMPILER: "gcc",   COMPILER_VERSION: "12", LINKER: "gold", CRYPTOLIB: "openssl" }
-          - { COMPILER: "clang", COMPILER_VERSION: "12", LINKER: "gold", CRYPTOLIB: "gcrypt"  }
-          - { COMPILER: "clang", COMPILER_VERSION: "13", LINKER: "lld",  CRYPTOLIB: "openssl" }
-          - { COMPILER: "clang", COMPILER_VERSION: "14", LINKER: "bfd",  CRYPTOLIB: "auto"    }
+          - { COMPILER: "clang", COMPILER_VERSION: "13", LINKER: "gold", CRYPTOLIB: "gcrypt"  }
+          - { COMPILER: "clang", COMPILER_VERSION: "14", LINKER: "lld",  CRYPTOLIB: "openssl" }
+          - { COMPILER: "clang", COMPILER_VERSION: "15", LINKER: "bfd",  CRYPTOLIB: "auto"    }
     env: ${{ matrix.env }}
     steps:
       - name: Repository checkout
--- a/src/basic/hashmap.c
+++ b/src/basic/hashmap.c
@@ -1190,7 +1190,7 @@ static int resize_buckets(HashmapBase *h, unsigned entries_add) {
                 } while (rehash_next);
         }
 
-        assert(n_rehashed == n_entries(h));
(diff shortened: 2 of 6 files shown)


src/core/mount.c
```diff
@@ -759,7 +759,7 @@ static int mount_coldplug(Unit *u) {
         assert(m);
         assert(m->state == MOUNT_DEAD);
 
-        if (m->deserialized_state != m->state)
+        if (m->deserialized_state != m->state && m->deserialized_state != MOUNT_MOUNTING)
                 new_state = m->deserialized_state;
         else if (m->from_proc_self_mountinfo)
                 new_state = MOUNT_MOUNTED;
```
yuwata: Not sure, but if the unit already has info from /proc/self/mountinfo, then the state should be MOUNTED, except for when the deserialized state is MOUNT_UNMOUNTING??

jcg190701: I tested it and this patch made the problem go away. And if you have a better fix, please do as soon as possible.

lclgo: Maybe this:
```c
if (m->deserialized_state != m->state)
        new_state = m->deserialized_state;
if (m->from_proc_self_mountinfo && new_state == MOUNT_MOUNTING)
        new_state = MOUNT_MOUNTED;
```
This means we were mounting something before reloading, now it is mounted, the new_state should keep consistent with proc_self_mountinfo.
@yuwata @poettering PTAL.

yuwata: @Geass-LL That should be better. But how about the other states? Especially when `new_state` is `MOUNT_DEAD`, `MOUNT_MOUNTING_DONE`, or `MOUNT_REMOUNTING`.
Also, if `m->from_proc_self_mountinfo` is `NULL`, then the state should be also adjusted, I guess.

lclgo: @yuwata I thought we can copy the logic in `mount_process_proc_self_mountinfo` yesterday. 
```c
MOUNT_DEAD/MOUNT_FAILED -> MOUNT_MOUNTED
MOUNT_MOUNTING_DONE -> MOUNT_MOUNTING_DONE
MOUNT_MOUNTING/MOUNT_REMOUNTING -> MOUNT_MOUNTING_DONE
```
Because at that time, I didn't konw systemd would process `/proc/self/mountinfo`  just after receiving sigchld of mount.
<https://github.com/systemd/systemd/blob/f0390fa03456ecb4d5f1482bd20a28ea98ae4d58/src/core/mount.c#L1376-L1390>
Now I'm adding some log to `mount_process_proc_self_mountinfo` to see why systemd didn't change the mount->state from MOUNT_MOUNTING to MOUNT_MOUNTING_DONE.

lclgo: @yuwata Figured out why. Before processing `/proc/self/mountinfo`, systemd will check if there are any mount events by calling `drain_libmount`. But we have closed the mount monitor when reexecuting, the event happened before we reestablish the monitor will be losed. So there is no change, and `mnt_monitor_next_change` returns 1. Then the following steps in `mount_process_proc_self_mountinfo` are skipped and mount->state will not change to MOUNT_MOUNTING_DONE even if `has` been mounted.
So I think we should add a flag to force check `/proc/self/mountinfo` after reexecuting.

jcg190701: Thank you all for your discussion and I look forward to a better fix.

PR #23811, opened 2022-06-23
github: Issue forms templates follow-up

jamacku:
- Use `SHA1` for actions versioning
- Fix typo: `github-issue-praser` -> `github-issue-parser`
- Define exact permissions

Follow-up to: #23693

PR #23813, opened 2022-06-23
github: Issue forms - fix GA `SHA1` ref

jamacku:
Fix `SHA1` reference for github action `stefanbuck/github-issue-parser` to point to correct commit.

Follow-up to: #23811

### DIFF
--- a/.github/workflows/issue_labeler.yml
+++ b/.github/workflows/issue_labeler.yml
@@ -26,7 +26,7 @@ jobs:
           node-version: '16'
 
       - name: Parse issue form
-        uses: stefanbuck/github-issue-parser@fb73b92b5c4d283c3e9a198f2bb7dbbf6cf87079
+        uses: stefanbuck/github-issue-parser@fc06b2a0adc5ccb7702ab6b641fd8a742a5e9cc0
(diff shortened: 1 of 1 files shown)


evverx:
I'm not sure https://github.com/stefanbuck/github-issue-parser/commit/fc06b2a0adc5ccb7702ab6b641fd8a742a5e9cc0 is compatible with Dependabot. My guess would be that it either won't be able to update the action or it will bump it to the next release without the "dist" folder.

evverx:
I'll go ahead and merge it to fix the action. Thanks!

jamacku:
There is still an option to use tags, and dependabot should be able to post updates.

evverx:
Unlike SHAs tags can be moved and it prevents the CI from being reproducible. Other than that https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

jamacku:
> Unlike SHAs tags can be moved and it prevents the CI from being reproducible. Other than that https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Good to know, thank you.

PR #23817, opened 2022-06-24
sd-device: send udev database version

### LINKED ISSUE #23799: Wrong current_tags from device created with udev_monitor_receive_device
 > 249.11-0ubuntu4

**Used distribution**
 > Ubuntu Kinetic Kudu (development branch)

 > `Linux z370m-d3h 5.15.0-27-generic #28-Ubuntu SMP Thu Apr 14 04:55:28 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

 > x86_64

 > ...

 > ...

 > …

```text
…
```

`udev_device_get_current_tags_list_entry` returns always same tags as `udev_device_get_tags_list_entry` if udev device is created with `udev_monitor_receive_device`. Works correctly if device is created with `udev_device_new_from_syspath`.

Problem seems to be that `udev_device` created from monitor has `sd_device` with `database_version == 0`.

Pull request https://github.com/systemd/systemd/pull/17622 added `device_database_supports_current_tags` that checks if `database_version` is `>=1`. If check fails, current_tags just returns regular tags.

Same problem if device is created with `udev_device_new_from_environment`.

yuwata:
Fixes #23799.

### DIFF
--- a/src/libsystemd/sd-device/device-private.c
+++ b/src/libsystemd/sd-device/device-private.c
@@ -356,6 +356,10 @@ static int device_amend(sd_device *device, const char *key, const char *value) {
                         if (r < 0)
                                 return log_device_debug_errno(device, r, "sd-device: Failed to add tag '%s': %m", word);
                 }
+        } else if (streq(key, "UDEV_DATABASE_VERSION")) {
+                r = safe_atou(value, &device->database_version);
+                if (r < 0)
+                        return log_device_debug_errno(device, r, "sd-device: Failed to parse udev database version '%s': %m", value);
         } else {
                 r = device_add_property_internal(device, key, value);
                 if (r < 0)
@@ -525,6 +529,15 @@ static int device_update_properties_bufs(sd_device *device) {
 
         if (!device->properties_buf_outdated)
                 return 0;
+
+        /* append udev database version */
+        buf_nulstr = newdup(char, "UDEV_DATABASE_VERSION=" STRINGIFY(LATEST_UDEV_DATABASE_VERSION) "\0",
+                            STRLEN("UDEV_DATABASE_VERSION=" STRINGIFY(LATEST_UDEV_DATABASE_VERSION)) + 2);
+        if (!buf_nulstr)
+                return -ENOMEM;
+
+        nulstr_len += STRLEN("UDEV_DATABASE_VERSION=" STRINGIFY(LATEST_UDEV_DATABASE_VERSION)) + 1;
+        num++;
 
         FOREACH_DEVICE_PROPERTY(device, prop, val) {
                 size_t len = 0;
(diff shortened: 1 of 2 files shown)


yuwata:
The current tag feature was introduced in v247.

src/libsystemd/sd-device/device-private.c
```diff
@@ -536,59 +535,57 @@ static int device_update_properties_bufs(sd_device *device) {
...
         assert(device);
-        assert(strv);
 
         r = device_update_properties_bufs(device);
         if (r < 0)
                 return r;
 
-        *strv = device->properties_strv;
+        if (ret)
+                *ret = device->properties_strv;
```
keszybz, line 602: Hmm, this change is not described in the commit message. If it's done, I'd prefer it to be split out to a separate commit. But I don't think it makes much sense: if `ret==NULL`, `device_get_properties_strv` doesn't make much sense.

src/libsystemd/sd-device/device-private.c
```diff
@@ -536,59 +535,57 @@ static int device_update_properties_bufs(sd_device *device) {
...
                 return r;
 
-        *nulstr = device->properties_nulstr;
-        *len = device->properties_nulstr_len;
-
+        if (ret_nulstr)
+                *ret_nulstr = device->properties_nulstr;
+        if (ret_len)
+                *ret_len = device->properties_nulstr_len;
         return 0;
```
keszybz: Same here: this change seems marginally useful, but the change in semantics should be a separate commit.

keszybz (commented):
Looks reasonable, apart from some minor squabbles about the first commit.

poettering:
lgtm

yuwata:
@keszybz and @poettering Thank you for your review. The first commit was split into two. PTAL.

src/libsystemd/sd-device/device-private.c
```diff
@@ -517,16 +521,23 @@ int device_new_from_nulstr(sd_device **ret, uint8_t *nulstr, size_t len) {
...
 
         if (!device->properties_buf_outdated)
                 return 0;
 
+        /* append udev database version */
+        buf_nulstr = newdup(char, "UDEV_DATABASE_VERSION=1\0", STRLEN("UDEV_DATABASE_VERSION=1") + 2);
+        if (!buf_nulstr)
+                return -ENOMEM;
+
+        nulstr_len += STRLEN("UDEV_DATABASE_VERSION=1") + 1;
```
muktupavels: Is there reason why version is hard-coded here? Would not it be better to use `LATEST_UDEV_DATABASE_VERSION` from `device-internal.h`?

yuwata: Good catch. I will update. Thank you.

keszybz:
LGTM.

keszybz:
I'll give this a few days to bake in `main` before backporting.

PR #23819, opened 2022-06-24
ci: set top-level permissions as well

evverx:
It should turn on the "restricted" mode by default regardless of
whether the global setting is on or not. New jobs in this action
should have to overwrite it explicitly to gain write access in any
way.

It should also make the action consistent with the other actions
writing various stuff like 'labeler' and 'codeql'.

### DIFF
--- a/.github/workflows/issue_labeler.yml
+++ b/.github/workflows/issue_labeler.yml
@@ -5,6 +5,9 @@ on:
   issues:
     types: [ opened ]
 
+permissions:
+  contents: read
+
 jobs:
   label-component:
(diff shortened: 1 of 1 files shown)


jamacku (approved):

evverx:
> it sounds safer for me

Ideally GitHub should do all that by default but because of I presume backward compatibility the default has never been changed. Actually even with read-only tokens it's possible to write some stuff to repositories but since it leads to overcharging paid accounts for resources they haven't consumed it doesn't seem to be a priority there.

PR #23821, opened 2022-06-24
Allow ASCII fallback for Unicode characters in logs

dtardon:


src/libsystemd/sd-bus/sd-bus.c
```diff
@@ -521,7 +521,7 @@ void bus_set_state(sd_bus *bus, enum bus_state state) {
         if (state == bus->state)
                 return;
 
-        log_debug("Bus %s: changing state %s → %s", strna(bus->description), table[bus->state], table[state]);
+        log_debug("Bus %s: changing state %s -> %s", strna(bus->description), table[bus->state], table[state]);
```
yuwata: Why not to use special_glyph() here?

src/libsystemd/sd-event/sd-event.c
```diff
@@ -405,7 +405,7 @@ _public_ int sd_event_new(sd_event** ret) {
         e->epoll_fd = fd_move_above_stdio(e->epoll_fd);
 
         if (secure_getenv("SD_EVENT_PROFILE_DELAYS")) {
-                log_debug("Event loop profiling enabled. Logarithmic histogram of event loop iterations in the range 2^0 … 2^63 us will be logged every 5s.");
+                log_debug("Event loop profiling enabled. Logarithmic histogram of event loop iterations in the range 2^0 ... 2^63 us will be logged every 5s.");
```
yuwata: Also here.

yuwata (changes requested):

src/libsystemd/sd-bus/sd-bus.c
dtardon: Hmm, why not indeed... `libsystemd` can use stuff from `basic`... Updated.

src/libsystemd/sd-event/sd-event.c
dtardon: Fixed too.

PR #23827, opened 2022-06-25
sd-event: process buffered inotify data

### LINKED ISSUE #23826: `sd_event_prepare()` fails to consider buffered inotify events

250

### Used distribution

Fedora 36

### Linux kernel version used

5.18.5-200.fc36.x86_64

x86_64

### Component

other

When there are pending inotify events (in the `inotify_data_buffered` of the event loop) then `sd_event_prepare()` ought to return non-zero and transition the loop to `SD_EVENT_PENDING`.

`sd_event_prepare()` returns `0` and the event loop is armed, instead.

The issue appears to be related to the special treatment of `inotify_data_buffered` in `process_epoll()` in `sd-event.c`.  In particular, it sets the timeout to 0 in case there is buffered data:

```
          /* If we still have inotify data buffered, then query the other fds, but don't wait on it */
          if (e->inotify_data_buffered)
                  timeout = 0;
```

There's no way to know this when trying to iterate the loop from outside, though, and there's no equivalent check in `sd_event_prepare()`.  That allows things to function normally when systemd is driving the loop, but not when trying to drive it from outside.  This issue therefore also affects the example "glib glue" code in `man/glib-event-glue.c`.

_No response_

yuwata:
Fixes #23826.

### DIFF
--- a/src/libsystemd/sd-event/test-event.c
+++ b/src/libsystemd/sd-event/test-event.c
@@ -757,4 +757,56 @@ TEST(inotify_self_destroy) {
         assert_se(sd_event_loop(e) >= 0);
 }
 
+struct inotify_process_buffered_data_context {
+        const char *path[2];
+        unsigned i;
+};
+
+static int inotify_process_buffered_data_handler(sd_event_source *s, const struct inotify_event *ev, void *userdata) {
+        struct inotify_process_buffered_data_context *c = ASSERT_PTR(userdata);
+        const char *description;
+
+        assert_se(sd_event_source_get_description(s, &description) >= 0);
+
+        assert_se(c->i < 2);
+        assert_se(streq(c->path[c->i], description));
+        c->i++;
+
+        return 1;
+}
+
+TEST(inotify_process_buffered_data) {
+        _cleanup_(rm_rf_physical_and_freep) char *p = NULL, *q = NULL;
+        _cleanup_(sd_event_source_unrefp) sd_event_source *a = NULL, *b = NULL;
+        _cleanup_(sd_event_unrefp) sd_event *e = NULL;
+        _cleanup_free_ char *z = NULL;
+
+        /* For issue #23826 */
+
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
I've tested this change and confirm that it fixes the issue, as reported.

Also worth noting that the `goto pending;` path immediately calls `sd_event_wait(e, 0);`, thus preserving the intended behaviour of "If we still have inotify data buffered, then query the other fds, but don't wait on it".

Thanks for such a fast fix!

yuwata:
Note the issue exists since 97ef5391697c34ee1c763fa9bddcd20a29ff3159 (v239), which introduces inotify event source.

@allisonkarlitskaya Thank you for the confirmation.

yuwata:
The second commit is split into two to make the fix easily backported to old releases.

DaanDeMeyer:
LGTM, but do we need to be careful about this being changing behavior? Or is the current behavior completely broken in that it can't work at all so it should be safe to merge?

yuwata:
> LGTM, but do we need to be careful about this changing behavior? Or is the current behavior completely broken in that it can't work at all so it should be safe to merge?

The current behavior is completely broken. If `sd_event_prepare()` returns 0 even if there exists buffered data, we need to call `sd_event_wait()` to wait for another event comes to a fd to make later call of `sd_event_prepare()` return 1 for the buffered data.

src/libsystemd/sd-event/sd-event.c
```diff
@@ -1799,7 +1799,7 @@ static void event_free_inode_data(
         if (!d)
                 return;
 
-        assert(!d->event_sources);
+        assert(LIST_IS_EMPTY(d->event_sources));
```
keszybz, line 1802: I don't think `LIST_IS_EMPTY` is useful. We could drop it altogether, but that's a discussion for a separate PR. If we have it, it's fine to use it here.

keszybz (commented):
LGTM.

PR #23828, opened 2022-06-25
sd-netlink: several cleanups for netfilter

yuwata:


src/shared/firewall-util-nft.c
```diff
@@ -637,51 +806,46 @@ static uint32_t concat_types2(enum nft_key_types a, enum nft_key_types b) {
         return type;
 }
 
-/* enough space to hold netlink messages for table skeleton */
-#define NFT_INIT_MSGS 16
 static int fw_nftables_init_family(sd_netlink *nfnl, int family) {
-        sd_netlink_message *batch[NFT_INIT_MSGS] = {};
-        size_t msgcnt = 0, i, ip_type_size;
+        sd_netlink_message *batch[12] = {};
```
daandemeyer: Changes this from 16 to 12 seems like a behavior change? That's unexpected in a cleanup commit

src/shared/firewall-util-nft.c
```diff
@@ -848,56 +1011,57 @@ static int nft_message_add_setelem_ip6range(
...
-                FirewallContext *ctx,
+                sd_netlink *nfnl,
                 bool add,
                 int af,
                 const union in_addr_union *source,
                 unsigned int source_prefixlen) {
 
-        sd_netlink_message *transaction[NFT_MASQ_MSGS] = {};
-        size_t tsize;
+        sd_netlink_message *transaction[4] = {};
```
daandemeyer: Ditto for the changed size

src/shared/firewall-util-nft.c
```diff
@@ -952,13 +1118,16 @@ static int fw_nftables_add_local_dnat_internal(
                 uint16_t remote_port,
                 const union in_addr_union *previous_remote) {
 
-        sd_netlink_message *transaction[NFT_DNAT_MSGS] = {};
+        sd_netlink_message *transaction[5] = {};
```
daandemeyer: Ditto

src/libsystemd/sd-netlink/netlink-message-nfnl.c
```diff
@@ -40,41 +42,176 @@ static int nft_message_new(sd_netlink *nfnl, sd_netlink_message **ret, int nfpro
...
+        r = nfnl_message_new_batch(nfnl, &batch_end, subsys, NFNL_MSG_BATCH_END);
+        if (r < 0)
+                return r;
+
+        netlink_seal_message(nfnl, batch_begin);
+        netlink_seal_message(nfnl, batch_end);
+
+        iovs[0] = IOVEC_MAKE(batch_begin->hdr, batch_begin->hdr->nlmsg_len);
+
+        for (size_t i = 0; i < n_messages; i++) {
```
daandemeyer, line 139: Maybe pull `i` out of the for loop and use it for the batch begin/end as well and as the total iovec size?

src/libsystemd/sd-netlink/netlink-message-nfnl.c
```diff
@@ -40,41 +42,176 @@ static int nft_message_new(sd_netlink *nfnl, sd_netlink_message **ret, int nfpro
...
         return 0;
 }
 
-int sd_nfnl_message_batch_begin(sd_netlink *nfnl, sd_netlink_message **ret) {
-        return nfnl_message_batch(nfnl, ret, NFNL_MSG_BATCH_BEGIN);
+int sd_nfnl_send_batch(
+                sd_netlink *nfnl,
+                sd_netlink_message **messages,
+                size_t n_messages,
+                uint32_t **ret_serial) {
```
daandemeyer: `ret_serials`?

DaanDeMeyer (approved):
LGTM, few small questions and nits

src/shared/firewall-util-nft.c
keszybz: There's 11 `++msgcnt` below, so 12 is OK. The last four positions were unused.

But… I find the way that the counting is done a bit confusing. Because the increase of the count is done *before* the next operation, not immediately *after* the operation that added an item. Wouldn't it be better to move the increment operations right after the addition operations?

src/shared/firewall-util-nft.c
keszybz: Same here, 3 `++tsize`, so the use of the array seems OK.

But `++tsize` right before `nfnl_netlink_sendv` is strange.

keszybz (commented):
LGTM too, except one stylistic issue.

yuwata:
@DaanDeMeyer and @keszybz Thank you for the review. All comments are addressed. Upgrading the green label.

PR #23829, opened 2022-06-26
meson.build: Fix GNUism on date(1)

lanodan:


yuwata (commented):
Which implementation of `date` command do you want to use?
At leaset, `date` provided by coreutils (and busybox) seem to support the `-I` option.

(Note, we do not support busybox, which does not support several required options in e.g. mount command.)

In this case, the output is used only on meson's summary, hence it is OK to approve this, but I am not willing to merge.

Conan-Kudo:
`-I` is also supported in FreeBSD `date(1)`, not that it matters _that_ much since we don't support FreeBSD.

yuwata:
No response from the author. Closing.

PR #23833, opened 2022-06-26
emacs: ignore .dir-locals-2.el (personal customization) versioning

jacekmigacz:
Ignores .dir-locals-2.el personal customization versioning as defined in https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ __pycache__/
 # Ignore any mkosi config files with "local" in the name
 /mkosi.default.d/**/*local*.conf
 /tags
+.dir-locals-2.el


yuwata (changes requested):
Not familiar with the file, but why only -2.el? How about to also add `.dir-locals.el`??

jacekmigacz:
Systemd source tree is already emacs-aware and likely you don't want to ignore .dir-locals.el, since and holds global configuration (indention and style).

yuwata:
Ah, indeed. We already have .dir-locals.el at several directories.

yuwata (approved):

PR #23834, opened 2022-06-26
A couple of tweaks to allow running TEST-21-DFUZZER in a container

mrc0mmand:
Inspired by https://github.com/systemd/systemd/pull/22547#discussion_r901076650.

/cc @evverx

### DIFF
--- a/test/units/testsuite-21.sh
+++ b/test/units/testsuite-21.sh
@@ -15,7 +15,7 @@ at_exit() {
     # from the queue
     if [[ $SHUTDOWN_AT_EXIT -ne 0 ]] && ! systemctl poweroff; then
         # PID1 is down let's try to save the journal
-        journalctl --sync || : # journal can be down as well so let's ignore exit codes here
+        journalctl --sync      # journal can be down as well so let's ignore exit codes here
         systemctl -ff poweroff # sync() and reboot(RB_POWER_OFF)
     fi
 }


test/units/testsuite-21.sh
```diff
@@ -33,17 +33,25 @@ BUS_LIST=(
...
 
 # systemd-oomd requires PSI
 if tail -n +1 /proc/pressure/{cpu,io,memory}; then
-    BUS_LIST+=(org.freedesktop.oom1)
+    BUS_LIST+=(
+        org.freedesktop.oom1
+    )
+fi
+
+# systemd-networkd and systemd-timesyncd don't work in containers
```
evverx: `networkd` can be run in containers if it has `CAP_NET_ADMIN` there. I think it should be skipped in containers if it can't be activated with `busctl introspect org.freedesktop.network1 /`. Something like
```
if ! systemd-detect-virt --container; then
    BUS_LIST+=(
        org.freedesktop.network1
        org.freedesktop.timesync1
    )
elif busctl introspect org.freedesktop.network1 / >&/dev/null then
    BUS_LIST+=(org.freedesktop.network1)
fi
```
should do I think.

mrc0mmand: Ah, interesting, I thought that running systemd-networkd in a container was impossible. Let me tweak that, thanks!

test/TEST-21-DFUZZER/test.sh
```diff
@@ -12,6 +11,13 @@ TEST_FORCE_NEWIMAGE=1
 # shellcheck source=test/test-functions
 . "${TEST_BASE_DIR:?}/test-functions"
 
+# Run the test either only under nspawn (if $TEST_PREFER_NSPAWN is set to true)
+# or only uder qemu otherwise, to avoid running the test twice on machines where
+# we can do both.
+if ! get_bool "${TEST_PREFER_NSPAWN:=}"; then
+    TEST_NO_NSPAWN=1
+fi
```
evverx, line 19: As far as I understand `TEST_NO_QEMU` should be set to `1` if `TEST_PREFER_NSPAWN` is `1`

evverx: Looks like `do_test` takes care of that

evverx: @mrc0mmand as far as I can tell if `TEST_NO_QEMU=yes` is set explicitly the test is skipped completely without `TEST_PREFER_NSPAWN` and that's weird in the sense that I'm not sure why I should pass
```
TEST_PREFER_NSPAWN=yes
```
if I already pass `TEST_NO_QEMU=yes`. I think I'll go ahead and merge the PR once the CI is green but all those variables like `TEST_PREFER_NSPAWN` that haven't been integrated into the testsuite properly should probably be revisited.

PR #23845, opened 2022-06-27
resolve: mdns: do not append cached entries if using cached entries is disabled

yuwata:
Prompted when reviewing #22751.

cc @vchillara

### DIFF
--- a/src/resolve/resolved-dns-transaction.c
+++ b/src/resolve/resolved-dns-transaction.c
@@ -1856,7 +1856,7 @@ static int dns_transaction_make_packet_mdns(DnsTransaction *t) {
         DNS_PACKET_HEADER(p)->qdcount = htobe16(qdcount);
 
         /* Append known answer section if we're asking for any shared record */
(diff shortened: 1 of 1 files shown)


yuwata:
(Please do not merge this. I will rebase after #23875 is merged.)

vchillara:
@yuwata Do you still plan on merging this?

yuwata:
> @yuwata Do you still plan on merging this?

To be honest, I am not sure about this. I will take a look again later. What is your opinion about this? Does this still make some sense?

vchillara:
I think this change does still make sense.
When using the SD_RESOLVED_NO_CACHE flag, we expect to get up-to-date data from the network.
However, appending known-answers to the query eliminates the possibility of receiving fresh info.

vchillara:
@yuwata Come to think of it, this probably causes more harm than good. It'd cause redundant traffic on the network.
As for the corner case in #22532 mentioned above, we now handle that with some additional logic in that PR itself.

PR #23848, opened 2022-06-27
core: fix SYSTEMD_WANTS and StopWhenUnneeded=

### LINKED ISSUE #23410: `StopWhenUnneeded=true` ignored if the unit was spawned as dependency of a device unit
 > systemd 249 (249.7)

**Used distribution**
 > NixOS 21.11

 > `Linux ft-ssy-wznb 5.17.5 #1-NixOS SMP PREEMPT Wed Apr 27 12:41:17 UTC 2022 x86_64 GNU/Linux`

 > x86_64

 > A unit with `StopWhenUnneeded=true` stops when `systemctl list-dependencies --reverse that-unit.service` yields only `that-unit.service`

 > `that-unit.service` keeps running forever, even after the dependent which triggered `that-unit`s start becomes `inactive (dead)`

1. Add the following unit config and udev rule to the system
`that-unit.service`:
```ini
[Unit]
StopWhenUnneeded=true

[Service]
ExecStartPre=/bin/date
ExecStop=/bin/date
RemainAfterExit=true
Type=oneshot
```

`99-that-unit.rules`:
```udev
SUBSYSTEM=="usb", ACTION!="remove", ATTRS{idVendor}=="dead", ATTRS{idProduct}=="beef", TAG+="systemd", ENV{SYSTEMD_WANTS}+="that-unit.service"
```
2. Plug a device matching the udev rule in question. A device unit for it will become active, which triggers the start of`that-unit.service`. `systemctl list-dependencies --reverse that-unit.service` correctly lists the device unit(s) as dependents of `that-unit.service`.
3. Unplug the device. Its respective device units change from `active (plugged)` to `inactive (dead)`. `systemctl list-dependencies --reverse that-unit.service` now names no dependents of `that-unit.service`. Still, `systemctl status that-unit.service` lists `that-unit.service` as `active (exited)`. I would expect `that-unit.service` to be `inactive (dead)`, since `man systemd.unit` claims that `StopWhenUnneeded=true` `a unit will be automatically cleaned up if no other active unit requires it`. However, `that-unit.service` keeps `active (exited)`.

_Assumption 1_: The manpage speaks of `this unit will be stopped when it is no longer used`. I interpret this as when `systemctl list-dependencies --reverse` lists no other units, than the unit is no longer used
_Assumption 2_: The manpage speaks of `will be stopped`. I assume that this will be the equivalent in outcome to `systemctl stop`.

Possibly duplicate of https://github.com/systemd/systemd/issues/15676, however a new issue seems to be the right course of action here.

yuwata:
Fixes #23410.

### DIFF
--- a/test/units/testsuite-17.07.sh
+++ b/test/units/testsuite-17.07.sh
@@ -1,0 +1,205 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
+set -o pipefail
+
+# shellcheck source=test/units/assert.sh
+. "$(dirname "$0")"/assert.sh
+
+wait_service_active() {(
+    set +ex
+    for (( i = 0; i < 20; i++ )); do
+        if (( i != 0 )); then sleep 0.5; fi
+        if systemctl --quiet is-active "${1?}"; then
+            return 0
+        fi
+    done
+    return 1
+)}
+
+wait_service_inactive() {(
+    set +ex
+    for (( i = 0; i < 20; i++ )); do
+        if (( i != 0 )); then sleep 0.5; fi
+        systemctl --quiet is-active "${1?}"
+        if [[ "$?" == "3" ]]; then
+            return 0
+        fi
+    done
+    return 1
+)}
+
+mkdir -p /run/systemd/system
+cat >/run/systemd/system/both.service <<EOF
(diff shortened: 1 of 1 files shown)


keszybz:
@yuwata I didn't include this in stable backports. Please set the label if it should be.

PR #23849, opened 2022-06-28
Use https for freedesktop.org

mbiebl:
`grep -l -r http:// | xargs sed -E -i s'#http://(.*).freedesktop.org#https://\1.freedesktop.org#'`

### DIFF
--- a/shell-completion/bash/bootctl
+++ b/shell-completion/bash/bootctl
@@ -14,7 +14,7 @@
 # General Public License for more details.
 #
 # You should have received a copy of the GNU Lesser General Public License
-# along with systemd; If not, see <http://www.gnu.org/licenses/>.
+# along with systemd; If not, see <https://www.gnu.org/licenses/>.
 
 __contains_word () {
     local w word=$1; shift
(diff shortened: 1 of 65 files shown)


man/os-release.xml
```diff
@@ -354,7 +354,7 @@
           <term><varname>LOGO=</varname></term>
 
           <listitem><para>A string, specifying the name of an icon as defined by <ulink
-          url="http://standards.freedesktop.org/icon-theme-spec/latest">freedesktop.org Icon Theme
+          url="https://standards.freedesktop.org/icon-theme-spec/latest".freedesktop.org Icon Theme
```
yuwata: This breaks the tag...

mbiebl: Thanks, fixed.

yuwata (approved):

medhefgo:
Why only for freedesktop.org? This should really be done for all urls that have https…

medhefgo:
I thought we did away with license boiler plate, but I guess that was forgotten for shell scripts?

mbiebl:
@medhefgo I was too lazy to go through all http refs and just picked the low-hanging fruit.

If you want to see the other urls changed, please send a follow-up PR!

PR #23855, opened 2022-06-28
basic/list: drop LIST_IS_EMPTY

keszybz:
This was a trivial wrapper that didn't provide any added value. With more
complicated structures like strvs, hashmaps, sets, and arrays, it is possible
to have an empty container. But in case of a list, the list is empty only when
the head is missing.

Also, we generally want the positive condition, so we replace many
if (!LIST_IS_EMPTY(x)) with just if (x).

### DIFF
--- a/src/test/test-list.c
+++ b/src/test/test-list.c
@@ -6,7 +6,7 @@
 int main(int argc, const char *argv[]) {
         size_t i;
         typedef struct list_item {
-                LIST_FIELDS(struct list_item, item);
+                LIST_FIELDS(struct list_item, item_list);
         } list_item;
         LIST_HEAD(list_item, head);
         LIST_HEAD(list_item, head2);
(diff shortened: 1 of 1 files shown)


yuwata:
Personally, I'd like to keep LIST_IS_EMPTY macro even if it is trivial, as it suggests the variable is a list, not a single object. But that is not strong opinion. And I do not refuse this, as at least two persons approve this.

keszybz:
For me it's pointless obfuscation. Not type safe, verbose, and actually hiding what is happening underneath.

yuwata:
> For me it's pointless obfuscation. Not type safe, verbose, and actually hiding what is happening underneath.

Right. But removing the macro hides that the variable is a list.
E.g. we usually write cleanup code like the followings:
```
LIST_FOREACH(name, i, c->objects)
        free_object(i);
```
or
```
while ((i = c->objects))
        free_object(i);
```
But, if the variable pointing the head (`c->objects` in the above) is not plural, e.g. `c->object`, then the context this is a list may be easily lost if we drop LIST_IS_EMPTY. And we may wrongly write like the following:
```
free_object(c->object);
```

If we drop LIST_IS_EMPTY, then I'd like to also rename all list heads into plural.

(BTW, if we drop the macro, why not to also drop LIST_HEAD and LIST_HEAD_INIT?? Of course, I'd like to also keep them.)

src/libsystemd/sd-event/sd-event.c
```diff
@@ -126,7 +126,7 @@ struct sd_event {
         LIST_HEAD(struct inode_data, inode_data_to_close);
 
         /* A list of inotify objects that already have events buffered which aren't processed yet */
-        LIST_HEAD(struct inotify_data, inotify_data_buffered);
+        LIST_HEAD(struct inotify_data, buffered_inotify_datas);
```
dnicolodi: I don't know if it matters, but "data" has no plural form. It is debatable whether data itself is singular or plural, but almost certainly, "datas" does not exist as an English word. The commend above this line suggests that using "objects" instead may be appropriate here.

yuwata: Interesting. But I recently heard that 'datas' is becoming approved as a plural of 'data', even though 'data' is already plural.

keszybz: "datas" would be nice, but dictionaries don't list it, so it seems that it's not widely accepted yet.
I renamed it to `buffered_inotify_data_objects` now.

src/libsystemd/sd-event/sd-event.c
```diff
@@ -123,7 +123,7 @@ struct sd_event {
         Hashmap *inotify_data; /* indexed by priority */
 
         /* A list of inode structures that still have an fd open, that we need to close before the next loop iteration */
-        LIST_HEAD(struct inode_data, inode_data_to_close);
+        LIST_HEAD(struct inode_data, inode_datas_to_close);
```
dnicolodi: Similar issue with "datas". As in the previous case, this should probably be read "inode_data"s as in "a list of `struct inode_data`" but I don't know a good way to concisely convey this as a C identifier. The comment above the declaration uses "inode structures" thus this may be a good candidate.

src/core/cgroup.c
```diff
@@ -1689,7 +1689,7 @@ static bool unit_get_needs_bpf_foreign_program(Unit *u) {
         if (!c)
                 return false;
 
-        return !LIST_IS_EMPTY(c->bpf_foreign_programs);
+        return c->bpf_foreign_programs;
```
poettering: i find this hard to read btw. i.e. if i read this changed line i'd assume this would return an `BPFProgram*` object, but it actually returns a boolean

src/libsystemd/sd-event/sd-event.c
```diff
@@ -126,7 +126,7 @@ struct sd_event {
         LIST_HEAD(struct inode_data, inode_data_to_close);
 
         /* A list of inotify objects that already have events buffered which aren't processed yet */
-        LIST_HEAD(struct inotify_data, inotify_data_buffered);
+        LIST_HEAD(struct inotify_data, buffered_inotify_data_objects);
```
poettering: "objects"? i am not sure i would call this "objects". they are dead structs, not more...

this renaming smells a bit like bike shedding to me though, but I don't mind, if this is important to you.

src/libsystemd/sd-event/sd-event.c
poettering: "datas" sounds really weird...

maybe just call it inode_data_list if you really don't like the simple "inode_data_to_close". I would leave the current name in place though.

src/systemctl/systemctl-show.c
```diff
@@ -64,7 +64,7 @@ typedef struct ExecStatusInfo {
 
         ExecCommandFlags flags;
 
-        LIST_FIELDS(struct ExecStatusInfo, exec);
+        LIST_FIELDS(struct ExecStatusInfo, exec_status_infos);
```
poettering: "infos" is a weird thing too. not a proper word. When I read it like this I'd assume a German tried to write English here, and didn't really succeed...

(becuase "infos" is a German word, but not an English one)

Hence, ehre too i'd leave the status quo ante in place, but if you insist renaming it, maybe suffix with "_list"

poettering:
> For me it's pointless obfuscation. Not type safe, verbose, and actually hiding what is happening underneath.

you could make it typesafe though (or "more" typesafe at least)

src/core/cgroup.c
keszybz: Yeah, I was thinking about this too. I changed it to use `!!`.

src/libsystemd/sd-event/sd-event.c
keszybz: I changed the name again, to `buffered_inotify_data_list`.

I don't particularly *like* the verbose name, but @yuwata requested a change, and I think it makes sense to have the general rule that container-like fields have a name that indicates that. This makes it easier for the reader to avoid a mistake where they forget that the thing is just one-of-many. If you don't like the new name, I can drop the commit, but I think it's better to rename.

keszybz:
Updated with a bunch of (hopefully better) new names.

PR #23858, opened 2022-06-28
test: copy libgcc_s.so.1 to TPM2 test image on Debian-like systems

enr0n:
On Ubuntu, cryptsetup does not link against libgcc_s.so.1 which leads to
the following test failure in TEST-70-TPM2:

 systemd[1]: Starting testsuite-70.service...
 systemd[329]: testsuite-70.service: Executing: /usr/lib/systemd/tests/testdata/units/testsuite-70.sh
 testsuite-70.sh[329]: + export SYSTEMD_LOG_LEVEL=debug
 testsuite-70.sh[329]: + SYSTEMD_LOG_LEVEL=debug
 testsuite-70.sh[329]: + img=/var/tmp/test.img
 testsuite-70.sh[329]: + dd if=/dev/zero of=/var/tmp/test.img bs=1024k count=20 status=none
 testsuite-70.sh[329]: + echo -n passphrase
 testsuite-70.sh[329]: + cryptsetup luksFormat -q --use-urandom /var/tmp/test.img /tmp/passphrase
 testsuite-70.sh[333]: libgcc_s.so.1 must be installed for pthread_exit to work
 testsuite-70.sh[329]: /usr/lib/systemd/tests/testdata/units/testsuite-70.sh: line 12:   333 Aborted                 (core dumped) cryptsetup luksFormat -q --use-urandom $img /tmp/passphrase

To fix this, manually install libgcc_s.so.1 to the test image if running
on Debian-like systems.

### DIFF
--- a/test/TEST-70-TPM2/test.sh
+++ b/test/TEST-70-TPM2/test.sh
@@ -21,6 +21,11 @@ test_append_files() {
         install_dmevent
         generate_module_dependencies
         inst_binary tpm2_pcrextend
+
+        # On Ubuntu, cryptsetup does not link against libgcc_s.so.1
+        if get_bool "$LOOKS_LIKE_DEBIAN"; then
+            inst_library "/lib/$(dpkg-architecture -qDEB_HOST_MULTIARCH)/libgcc_s.so.1"
+        fi
     )
 }
 


mbiebl:
where have you seen this failure?

enr0n:
> a bit weird, shouldn't it link against it?

Yeah, I don't understand the full story yet.

I think libpthread loads libgcc_s via `dlopen`. But, on systems with glibc >= 2.34 glibc ships with libpthread bundled with libc, which means cryptsetup does not link against libpthread.so.0. I am not familiar enough with this test suite to know if that has to do with it, but I know Debian and Ubuntu initramfs-tools have special handling for this [1][2].

[1] https://bugs.launchpad.net/ubuntu/+source/initramfs-tools/+bug/1958594
[2] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=950254

enr0n:
> where have you seen this failure?

Working on getting 251.2 into Ubuntu Kinetic, so it may be specific to Ubuntu if the glibc version is in fact relevant.

mbiebl:
> > where have you seen this failure?
> 
> Working on getting 251.2 into Ubuntu Kinetic, so it may be specific to Ubuntu if the glibc version is in fact relevant.

I was asking because I can't reproduce the failure on Debian sid.

yuwata:
> > > where have you seen this failure?
> > 
> > 
> > Working on getting 251.2 into Ubuntu Kinetic, so it may be specific to Ubuntu if the glibc version is in fact relevant.
> 
> I was asking because I can't reproduce the failure on Debian sid.

@enr0n If this is an ubuntu specific issue, then could you update the comment and commit message?

enr0n:
I have spent some time looking into this, and I have not yet been able to figure out the root cause, i.e. if this needs to be fixed in cryptsetup, libargon2, or the toolchain (or something else).

Despite this test currently working on sid, AFIACT `cryptsetup` does not link against `libgcc_s.so.1` on sid either. It's possible that the difference between sid and kinetic is that some other binary which *does* link against `libgcc_s.so.1` is being copied into the sid test image.

Since cryptsetup requires `libgcc_s.so.1`, and cryptsetup is critical to this test case, I think the workaround should be applied for now.

bluca (approved):

mbiebl:
Without knowing the root cause it becomes harder to drop the workaround again, though.

enr0n:
> @enr0n could you post the post the commands you used to setup the test environment which revealed this issue?

This demonstrates the issue with systemd 251.2-7 from sid, but running in a kinetic autopkgtest environment (I don't have a sid image handy at the moment):
```bash
apt install autopkgtest ubuntu-dev-tools -y # If needed 
autopkgtest-buildvm-ubuntu-cloud -r kinetic
pull-lp-source systemd unstable -D debian
autopkgtest systemd_251.2-7.dsc --shell-fail -U --test-name upstream -- qemu --ram-size=8192 --cpus=4 autopkgtest-kinetic-amd64.img
```

mrc0mmand:
I came across this issue on CentOS Stream 8 as well, but in TEST-01-BASIC:
```
$ journalctl --file system.journal -u rngd -o short-monotonic
[    0.985727] localhost rngd[297]: Initializing available sources
[    0.985727] localhost rngd[297]: [hwrng ]: Initialized
[    0.985727] localhost rngd[297]: [rdrand]: Enabling RDSEED rng support
[    0.985727] localhost rngd[297]: [rdrand]: Initialized
[    1.028676] localhost rngd[297]: [jitter]: Initializing AES buffer
[    4.450879] localhost rngd[297]: [jitter]: Enabling JITTER rng support
[    4.450879] localhost rngd[297]: [jitter]: Initialized
[    4.716227] localhost systemd[1]: Stopping rngd.service...
[   15.866611] H rngd[297]: [hwrng ]: Shutting down
[   15.866611] H rngd[297]: [rdrand]: Shutting down
[   15.866611] H rngd[297]: [jitter]: Shutting down
[   15.866611] H rngd[297]: libgcc_s.so.1 must be installed for pthread_cancel to work
[   15.922480] H systemd[1]: rngd.service: Installed new job rngd.service/stop as 231
[   16.764075] H systemd[1]: rngd.service: Failed to load configuration: No such file or directory
[   16.771694] H systemd[1]: rngd.service: Changed dead -> failed
```

And yeah, `libgcc_s.so` is `dlopen()`ed by libpthread (https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/unwind-link.c;hb=9c443ac4559a47ed99859bd80d14dc4b6dd220a1#l50), hence it's not pulled automagically to the test images. I guess we should install it explicitly if it's available to avoid such issues.

Issue #23287, opened 2022-05-06
Unclear systemd-journald.socket destructive transaction error during shutdown


daandemeyer:
**systemd version the issue has been seen with**
 > 250.3

**Used distribution**
 > CentOS Stream 8

**Linux kernel version used** (`uname -a`)

 > `5.6`

**CPU architecture issue was seen on**
 > x86_64

**Steps to reproduce the problem**
 > Unsure

**Additional program output to the terminal or log subsystem illustrating the issue**

I'm looking into common journald error logs on our machines and one of the most common ones is the following:

```
systemd[1]: systemd-journald.socket: Failed to queue service startup job (Maybe the service file is missing or not a non-template unit?): Transaction for systemd-journald.service/start is destructive (shutdown.target has 'start' job queued, but 'stop' is inclu## in transaction).
```

It's a bit unclear to me what this exactly means. I suspect that something is accessing the journald socket during shutdown which tries to activate journald but this isn't allowed at that point in time? Maybe we can improve the logging message and maybe there's something we need to fix in the unit dependencies so that this can't happen (by stopping the journald socket if we know that activating journald is going to fail?)

DaanDeMeyer:
Specifically, I don't quite get how systemd-journald.socket would cause a transaction to include shutdown.target since neither it nor systemd-journald.service seem to include shutdown.target.

DaanDeMeyer:
> But we just stopped that, and since it conflicts with shutdown.target actually starting it again this late, would mean reversing shutdown.target/start to shutdown.target/stop, which we expressly don#t want to do.

This is what's confusing me. I'm probably just blind but I don't see how systemd-journald conflicts with shutdown.target from its service file.

> after all during shutdown we terminate the journald service before its socket

Why do we do this? If the socket can't do its job after journald is shut down, why destroy journald before it's socket?

keszybz:
From discussion during videoconf: set `IgnoreOnIsolate=yes` on `systemd-journald.service`. It might help, though it's unclear if this is a fix for this issue.

DaanDeMeyer:
Let's close this, we should try to expose better logging in the debug message but let's track that in a separate issue

PR #23863, opened 2022-06-29
units: add IgnoreOnIsolate=yes to systemd-journald too

keszybz:
We already had it on the socket units, so it's possible that
systemd-journald.service would be stopped and then restarted when trafic hits
the sockets when something logs. Let's not try to stop it. It is supposed to
run until the end and be eventually killed in the final killing spree.

This might (or not) help with #23287.

### DIFF
--- a/units/systemd-journald-dev-log.socket
+++ b/units/systemd-journald-dev-log.socket
@@ -13,9 +13,9 @@ Documentation=man:systemd-journald.service(8) man:journald.conf(5)
 DefaultDependencies=no
 Before=sockets.target
 
-# Mount and swap units need this. If this socket unit is removed by an
-# isolate request the mount and swap units would be removed too,
-# hence let's exclude this from isolate requests.
+# Mount and swap units need this. If this socket unit is removed by an isolate
+# request the mount and swap units would be removed too, hence let's exclude
+# systemd-journald and its sockets from isolate requests.
 IgnoreOnIsolate=yes
 
 [Socket]
--- a/units/systemd-journald.service.in
+++ b/units/systemd-journald.service.in
@@ -15,6 +15,11 @@ Requires=systemd-journald.socket
 After=systemd-journald.socket systemd-journald-dev-log.socket systemd-journald-audit.socket syslog.socket
 Before=sysinit.target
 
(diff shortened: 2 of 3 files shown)


DaanDeMeyer:
I already opened https://github.com/systemd/systemd/pull/23857 but I'll close that one in favor of this one

keszybz:
> I already opened #23857 but I'll close that one in favor of this one

Oops. I checked the bug, but didn't think to look for other PRs.

mbiebl:
Could this interfere with emergency/rescue mode?
See also https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=999695

keszybz:
I don't think it'll "interfere". Basically, we have two conflicting goals: to have systemd-journald running so that we have working logging, and the ability to remount the fs read-only. Right now, you cannot do the latter anyway, because the socket units are not stopped, so systemd-journald can be restarted. Nowadays `journalctl --relinquish-var` should be used to make allow remount of root ro, and I think it's reasonable to require users who want to manipulate the root device to do this explicitly. This PR is independent of this.

mbiebl:
> Nowadays `journalctl --relinquish-var` should be used to make allow remount of root ro, and I think it's reasonable to require users who want to manipulate the root device to do this explicitly.

Could we do that automatically when we enter emergency/rescue mode?

keszybz:
> > Nowadays `journalctl --relinquish-var` should be used to make allow remount of root ro, and I think it's reasonable to require users who want to manipulate the root device to do this explicitly.
> 
> Could we do that automatically when we enter emergency/rescue mode?

I think we *could*, but I don't think we *should*. It'd only be needed rarely, and to fix things in rescue mode, you need to know what you're doing anyway.

mbiebl:
> I think we _could_, but I don't think we _should_. It'd only be needed rarely, and to fix things in rescue mode, you need to know what you're doing anyway. I don't think we should try to add even more fancy logic to rescue mode, it's hard to get it working already.

Problem is, that the `--relinquish-var` feature is not particularly well known or easily discoverable. 
A manual fsck is something users are more familiar with.

yuwata (approved):

PR #23865, opened 2022-06-29
sd-id128: avoid an unnecessary function call in inline helper

keszybz:
When optimizing, the compiler will most likely replace the call to memcmp(),
but at -O0, the code that is emitted builds the call preamble and does the
call. Let's use the same pattern as with sd_id128_is_null() and
sd_id128_is_allf() and avoid the call.

Split out of #23773, since apparently it's making the compiler unhappy.

### DIFF
--- a/src/home/homework-luks.c
+++ b/src/home/homework-luks.c
@@ -179,7 +179,7 @@ static int probe_file_system_by_path(const char *path, char **ret_fstype, sd_id1
 
         fd = open(path, O_RDONLY|O_CLOEXEC|O_NOCTTY|O_NONBLOCK);
         if (fd < 0)
-                return -errno;
+                return negative_errno();
 
         return probe_file_system_by_fd(fd, ret_fstype, ret_uuid);
 }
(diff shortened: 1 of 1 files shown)


yuwata:
```
In file included from ../src/systemd/sd-bus.h:27,
                 from ../src/journal/journalctl.c:21:
In function â€˜sd_id128_equalâ€™,
    inlined from â€˜mainâ€™ at ../src/journal/journalctl.c:2697:46:
../src/systemd/sd-id128.h:119:43: error: â€˜previous_boot_id.qwords[0]â€™ may be used uninitialized [-Werror=maybe-uninitialized]
  119 |         return a.qwords[0] == b.qwords[0] && a.qwords[1] == b.qwords[1];
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/journal/journalctl.c: In function â€˜mainâ€™:
../src/journal/journalctl.c:2167:20: note: â€˜previous_boot_id.qwords[0]â€™ was declared here
 2167 |         sd_id128_t previous_boot_id;
      |                    ^~~~~~~~~~~~~~~~
```

src/home/homework-luks.c
```diff
@@ -1229,7 +1229,7 @@ int home_setup_luks(
                 PasswordCache *cache,
                 UserRecord **ret_luks_home) {
 
-        sd_id128_t found_partition_uuid = SD_ID128_NULL, found_luks_uuid = SD_ID128_NULL, found_fs_uuid = SD_ID128_NULL;
+        sd_id128_t found_partition_uuid, found_luks_uuid, found_fs_uuid;
```
yuwata: `found_luks_uuid` may not be initialized, as `luks_open()` is called only when `!setup->crypt_device`.

yuwata (changes requested):
The CI failure seems related.

src/home/homework-luks.c
keszybz: Arggh, indeed. Should be better now.

Issue #22560, opened 2022-02-18
Anonymous auth incompatible with dbus implementation


anthonybrandon:
I've noticed that since commit 1ed4723d38cd0d1423c8fe650f90fa86007ddf55 the dbus anonymous auth method seems incompatible with with other dbus implementations (specifically https://gitlab.freedesktop.org/dbus/dbus/)
Is this intentional?

The problem comes from two places: the unexpected DATA in sasl_auth_anonymous
And the server never sending the expected DATA back.

How to reproduce:
Try to connect to a remote https://gitlab.freedesktop.org/dbus/dbus/ server using anonymous auth and call some dbus method.

leiflm:
https://github.com/altdesktop/python-dbus-next/pull/121 might be helpful to others.

anthonybrandon:
@leiflm so am I interpreting that comment correctly that the sdbus implementation is correct and it's the other library that should be changed?

dvdhrm:
Honestly, what dbus-daemon does looks wrong. It seems to require an inline argument to `AUTH <mech>` while the spec does not mandate this. `dbus/dbus-auth.c:handle_auth()` is what I am looking at. It calls `process_data()` unconditionally on any AUTH line, even though data is not mandatory in AUTH-lines. But not sure what exactly fails. Can you elaborate what the issue exactly is?

leiflm:
> Honestly, what dbus-daemon does looks wrong. 

I'm looking at `systemd-stdio-bridge`.

> It seems to require an inline argument to `AUTH <mech>` while the spec does not mandate this. `dbus/dbus-auth.c:handle_auth()` is what I am looking at.

The [relevant statement](https://github.com/systemd/systemd/blob/3a2a0806f5520dc1a979f6f3faebfd5c6f10301b/src/libsystemd/sd-bus/bus-socket.c#L419) within the server context boils down to the question of whether the server should answer `OK` or `DATA`, if the client sends `AUTH ANONYMOUS`?

I'd argue the server should reply `OK` right away, as the client actually confirms that it is done sending credentials by sending `BEGIN` as reply to an `OK`. But I am not familiar with SASL. @dvdhrm do you know whether SASL mandates this?

So - IMO - what _should_ happen is:
```
...
C: AUTH ANONYMOUS
S: OK
C: DATA mytraceinfo
S: OK
C: BEGIN
```

[...]

PR #23869, opened 2022-06-30
bus: use inline trace argument for ANONYMOUS auth

dvdhrm:
Rather than using a separate DATA round to transmit the trace-string of
the ANONYMOUS authentication scheme, transmit it inline as argument.
This requires a refactor of the client-side SASL parser, as we now have
a different set of replies depending on the mode used.

This fixes an issue where libdbus-1 does not query for trace-strings if
not transmit inline as AUTH-ANONYMOUS argument. It is unclear from the
wording of the spec whether this is a violation by libdbus-1. However,
we can work around it by simply changing our mode of transmittal.

Ref: #22560

### DIFF
--- a/src/libsystemd/sd-bus/bus-socket.c
+++ b/src/libsystemd/sd-bus/bus-socket.c
@@ -156,75 +156,86 @@ static int bus_socket_write_auth(sd_bus *b) {
 }
 
 static int bus_socket_auth_verify_client(sd_bus *b) {
-        char *d, *e, *f, *start;
+        char *l, *lines[4] = {};
         sd_id128_t peer;
+        size_t i, n;
         int r;
 
         assert(b);
 
         /*
-         * We expect three response lines:
-         *   "DATA\r\n"
+         * We expect up to three response lines:
+         *   "DATA\r\n"                 (optional)
          *   "OK <server-id>\r\n"
          *   "AGREE_UNIX_FD\r\n"        (optional)
          */
 
-        d = memmem_safe(b->rbuffer, b->rbuffer_size, "\r\n", 2);
-        if (!d)
-                return 0;
-
-        e = memmem_safe(d + 2, b->rbuffer_size - (d - (char*) b->rbuffer) - 2, "\r\n", 2);
-        if (!e)
-                return 0;
-
-        if (b->accept_fd) {
-                f = memmem_safe(e + 2, b->rbuffer_size - (e - (char*) b->rbuffer) - 2, "\r\n", 2);
-                if (!f)
-                        return 0;
-
-                start = f + 2;
-        } else {
-                f = NULL;
-                start = e + 2;
+        n = 0;
+        lines[n] = b->rbuffer;
+        for (i = 0; i < 3; ++i) {
+                l = memmem_safe(lines[n], b->rbuffer_size - (lines[n] - (char*) b->rbuffer), "\r\n", 2);
+                if (l)
+                        lines[++n] = l + 2;
+                else
+                        break;
         }
 
-        /* Nice! We got all the lines we need. First check the DATA line. */
+        /*
+         * If we sent a non-empty initial response, then we just expect an OK
+         * reply. We currently do this if, and only if, we picked ANONYMOUS.
+         * If we did not send an initial response, then we expect a DATA
+         * challenge, reply with our own DATA, and expect an OK reply. We do
+         * this for EXTERNAL.
+         * If FD negotiation was requested, we additionally expect
+         * an AGREE_UNIX_FD response in all cases.
+         */
+        if (n < (b->anonymous_auth ? 1U : 2U) + !!b->accept_fd)
+                return 0; /* wait for more data */
 
-        if (d - (char*) b->rbuffer == 4) {
-                if (memcmp(b->rbuffer, "DATA", 4))
+        i = 0;
+
+        /* In case of EXTERNAL, verify the first response was DATA. */
+        if (!b->anonymous_auth) {
+                l = lines[i++];
+                if (lines[i] - l == 4 + 2) {
+                        if (memcmp(l, "DATA", 4))
+                                return -EPERM;
+                } else if (lines[i] - l == 3 + 32 + 2) {
+                        /*
+                         * Old versions of the server-side implementation of
+                         * `sd-bus` replied with "OK <id>" to "AUTH" requests
+                         * from a client, even if the "AUTH" line did not
+                         * contain inlined arguments. Therefore, we also accept
+                         * "OK <id>" here, even though it is technically the
+                         * wrong reply. We ignore the "<id>" parameter, though,
+                         * since it has no real value.
+                         */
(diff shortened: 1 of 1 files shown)


src/libsystemd/sd-bus/bus-socket.c
```diff
@@ -156,93 +156,109 @@ static int bus_socket_write_auth(sd_bus *b) {
...
-                start = f + 2;
-        } else {
-                f = NULL;
-                start = e + 2;
+        /*
+         * In case of ANONYMOUS, we expect OK. In case of EXTERNAL, we expect
+         * DATA+OK. If FD negotiation was requested, we additionally expect
+         * an AGREE_UNIX_FD response.
+         */
+        if (n < (b->anonymous_auth ? 1U : 2U) + !!b->accept_fd)
```
smcv: If I'm remembering RFC 4422 correctly, `b->anonymous_auth` here is acting as a proxy for: did we send a non-empty initial response in the `AUTH` command? But because systemd currently only implements `EXTERNAL` (*without* an initial response, because systemd wants to send an empty response to the empty challenge, so that it'll work across container boundaries) and `ANONYMOUS` (*with* an initial response, after this MR), that's equivalent to asking whether we're using `ANONYMOUS`.

It would probably make sense to mention that in the comment.
("Did we send"? "Do we plan to send"? Pipelining isn't good at preserving causality.)

dvdhrm: > It would probably make sense to mention that in the comment.

Mh. I tried to explain that in the comment you quoted. The text matches one-to-one (in-order) the code. It says if ANONYMOUS was used (`b->anonymous_auth == true`), we expect 1 answer. In case of EXTERNAL (`b->anonymous_auth == false`) we expect 2. In both cases, if FD-negotiation was used, an additional answer is expected.

Are you suggesting that `(b->anonymous_auth == false) => EXTERNAL` is not clear? I that case, I can repalce `in case of EXTERNAL` with `in all other cases` maybe?

src/libsystemd/sd-bus/bus-socket.c
```diff
@@ -156,93 +156,109 @@ static int bus_socket_write_auth(sd_bus *b) {
...
+        lines[n] = b->rbuffer;
+        l = memmem_safe(b->rbuffer, b->rbuffer_size, "\r\n", 2);
+        if (l) {
+                lines[++n] = l + 2;
+                l = memmem_safe(l + 2, b->rbuffer_size - (l - (char*) b->rbuffer) - 2, "\r\n", 2);
+                if (l) {
+                        lines[++n] = l + 2;
+                        l = memmem_safe(l + 2, b->rbuffer_size - (l - (char*) b->rbuffer) - 2, "\r\n", 2);
+                        if (l)
+                                lines[++n] = l + 2;
```
poettering: uh, this smells as if there should be a helper call, given it basically repeats the same complex expression thrice

src/libsystemd/sd-bus/bus-socket.c
smcv: > Are you suggesting that `(b->anonymous_auth == false) => EXTERNAL` is not clear? I that case, I can repalce `in case of EXTERNAL` with `in all other cases` maybe?

No, I'm suggesting that it's not clear that `b->anonymous_auth` is equivalent to "we are using the initial-response optimization".

When you say "inline" in the commit message, what you're really saying is that you're sending a non-empty initial response. See RFC 4422 §3a: `AUTH` is D-Bus' implementation of the "message to initiate the authentication exchange", the first argument is the "field for carrying the name of the mechanism", and the second argument is the "optional field for carrying an initial response".

So perhaps something like this?

```
          * If we sent a non-empty initial response, then we just expect an OK reply.
          * We currently do this if and only if ANONYMOUS.
          * If we did not send an initial response, then we expect the DATA challenge,
          * reply with our own DATA, and expect an OK reply. We do this for EXTERNAL.
          * If FD negotiation was requested, we additionally expect
          * an AGREE_UNIX_FD response. 
```

dvdhrm: Fair enough, updated!

src/libsystemd/sd-bus/bus-socket.c
```diff
@@ -156,93 +156,113 @@ static int bus_socket_write_auth(sd_bus *b) {
...
+                }
+        }
 
-                start = f + 2;
-        } else {
-                f = NULL;
-                start = e + 2;
+        /*
+         * If we sent a non-empty initial response, then we just expect an OK
```
poettering, line 184: doesn't matter, but we usually line break at 109ch (.editorconfig, .vimrc, .dir-locals.el, .clang-format all carry that info btw)

src/libsystemd/sd-bus/bus-socket.c
```diff
@@ -156,93 +156,113 @@ static int bus_socket_write_auth(sd_bus *b) {
...
+                         */
+                        if (memcmp(l, "OK ", 3))
+                                return -EPERM;
+                } else
+                        return -EPERM;
         }
 
-        /* Nice! We got all the lines we need. First check the DATA line. */
+        /* Now check the OK line. */
+        {
```
poettering: why the { block here?

src/libsystemd/sd-bus/bus-socket.c
```diff
@@ -156,93 +156,113 @@ static int bus_socket_write_auth(sd_bus *b) {
...
-
-        if (f)
+        if (b->accept_fd) {
+                l = lines[i++];
                 b->can_fds =
-                        (f - e == STRLEN("\r\nAGREE_UNIX_FD")) &&
-                        memcmp(e + 2, "AGREE_UNIX_FD",
+                        (lines[i] - l == STRLEN("\r\nAGREE_UNIX_FD")) &&
+                        memcmp(l, "AGREE_UNIX_FD",
                                STRLEN("AGREE_UNIX_FD")) == 0;
```
poettering: startswith()?

poettering (changes requested):
looks good to me, just some minor nitpicks.

I am not going to check if the claims about libdbus-1 behaviour and so on are true, I trust you on that...

src/libsystemd/sd-bus/bus-socket.c
dvdhrm: You wrote that code, I am just moving things around ;) But sure, I can try extracting this bit.

src/libsystemd/sd-bus/bus-socket.c
dvdhrm: For visual guidance? I can drop it.

src/libsystemd/sd-bus/bus-socket.c
dvdhrm: Sure.

poettering:
thanks, lgtm

bluca (approved):
do we need to backport this?

keszybz:
Queued up for v251.5.

PR #23870, opened 2022-06-30
logind: log wall messages to the journal

keszybz:
Currently kde installs a fake utmp session to listen for this. This provides an
alternative mechanism as discussed in #23574.

Example with 'shutdown 6 -r' and shutdown -c':
    PRIORITY=6
    SYSLOG_FACILITY=4
    SYSLOG_IDENTIFIER=systemd-logind
    ...
    CODE_FILE=src/login/logind-utmp.c
    CODE_LINE=90
    CODE_FUNC=warn_wall
    MESSAGE_ID=9e7066279dc8403da79ce4b1a69064b2
    OPERATOR=root
    MESSAGE=The system will reboot at Thu 2022-06-30 12:16:43 CEST!
    ACTION=reboot

    PRIORITY=5
    SYSLOG_FACILITY=4
    SYSLOG_IDENTIFIER=systemd-logind
    ...
    OPERATOR=root
    CODE_FILE=src/login/logind-dbus.c
    CODE_LINE=2407
    CODE_FUNC=method_cancel_scheduled_shutdown
    MESSAGE=System shutdown has been cancelled
    MESSAGE_ID=249f6fb9e6e2428c96f3f0875681ffa3
    ACTION=reboot

### DIFF
--- a/src/systemd/sd-messages.h
+++ b/src/systemd/sd-messages.h
@@ -203,6 +203,14 @@ _SD_BEGIN_DECLARATIONS;
 #define SD_MESSAGE_TIME_SYNC              SD_ID128_MAKE(7c,8a,41,f3,7b,76,49,41,a0,e1,78,0b,1b,e2,f0,37)
 #define SD_MESSAGE_TIME_SYNC_STR          SD_ID128_MAKE_STR(7c,8a,41,f3,7b,76,49,41,a0,e1,78,0b,1b,e2,f0,37)
 
+#define SD_MESSAGE_LOGIND_SHUTDOWN        SD_ID128_MAKE(9e,70,66,27,9d,c8,40,3d,a7,9c,e4,b1,a6,90,64,b2)
+#define SD_MESSAGE_LOGIND_SHUTDOWN_STR    SD_ID128_MAKE_STR(9e,70,66,27,9d,c8,40,3d,a7,9c,e4,b1,a6,90,64,b2)
+
+#define SD_MESSAGE_LOGIND_SHUTDOWN_CANCELED \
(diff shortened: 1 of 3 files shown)


mbiebl:
Hm, this requires root/admin permissions to read the system journal and kded runs as unprivileged user.
A D-Bus message seems like a more intuitive choice for this kind of use case.

keszybz:
I think that dbus would be an option too. But regardless, I think we should log to the journal.

src/systemd/sd-messages.h
```diff
@@ -203,6 +203,14 @@ _SD_BEGIN_DECLARATIONS;
 #define SD_MESSAGE_TIME_SYNC              SD_ID128_MAKE(7c,8a,41,f3,7b,76,49,41,a0,e1,78,0b,1b,e2,f0,37)
 #define SD_MESSAGE_TIME_SYNC_STR          SD_ID128_MAKE_STR(7c,8a,41,f3,7b,76,49,41,a0,e1,78,0b,1b,e2,f0,37)
 
+#define SD_MESSAGE_LOGIND_SHUTDOWN        SD_ID128_MAKE(9e,70,66,27,9d,c8,40,3d,a7,9c,e4,b1,a6,90,64,b2)
```
bluca, line 206: shouldn't these two be added to `catalog/systemd.catalog.in`?

keszybz: I think the messages are self-explanatory. I don't think there's any reason to provide an additional explanation.

bluca (approved):

bluca:
> I think that dbus would be an option too. But regardless, I think we should log to the journal.

This can be done separately if desired

PR #23882, opened 2022-07-01
Add tests for org.freedesktop.login1.Session SetType and SetDisplay

dtardon:


test/units/testsuite-35.sh
```diff
@@ -244,11 +244,12 @@ test_session_tear_down() {
     rm -f /run/udev/rules.d/70-logindtest-scsi_debug-user.rules
     udevadm control --reload
 
+    loginctl terminate-session $SESSION
     systemctl stop getty@tty2.service
     rm -rf /run/systemd/system/getty@tty2.service.d
```
yuwata: Maybe, it is better to call `loginctl terminate-session` after `systemctl stop getty@tty2`??

Also, `$SESSION` may be empty here. I'd like to this be conditional.

test/units/testsuite-35.sh
```diff
@@ -253,6 +253,8 @@ test_session_tear_down() {
     userdel logind-test-user
 
     rmmod scsi_debug
+
+    set -e
```
yuwata: We usually make trap function subshell, and unset "x":
```
test_session_tear_down() (
    set +ex

    ...
)
```

test/TEST-35-LOGIN/test-session-properties.c
```diff
@@ -0,0 +1,65 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
```
yuwata: I'd like to locate this .c file in src/login.

yuwata (changes requested):
Looks nice! Several minor comments.

keszybz:
This looks reasonable. I think with @yuwata's points addressed, this should be merged.

test/TEST-35-LOGIN/test-session-properties.c
dtardon: Makes sense, moved.

src/login/test-session-properties.c
```diff
@@ -0,0 +1,110 @@
...
+TEST(set_display) {
+        _cleanup_(sd_bus_flush_close_unrefp) sd_bus* bus = NULL;
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_free_ char *display = NULL;
+
+        assert_se(sd_bus_open_system(&bus) >= 0);
+
+        /* Display is unset by default */
+        assert_se(bus_get_property_string(bus, &session, "Display", NULL, &display) >= 0);
+        assert_se(strempty(display));
```
yuwata: `isempty()`

src/login/test-session-properties.c
```diff
@@ -0,0 +1,110 @@
...
+        assert_se(bus_call_method(bus, &session, "SetDisplay", NULL, NULL, "s", ":0") >= 0);
+        display = mfree(display);
+        assert_se(bus_get_property_string(bus, &session, "Display", NULL, &display) >= 0);
+        assert_se(streq(display, ":0"));
+
+        /* Empty display can be set too */
+        assert_se(bus_call_method(bus, &session, "SetDisplay", NULL, NULL, "s", "") >= 0);
+        display = mfree(display);
+        assert_se(bus_get_property_string(bus, &session, "Display", NULL, &display) >= 0);
+        assert_se(strempty(display));
```
yuwata: `isempty()`

test/TEST-35-LOGIN/test.sh
```diff
@@ -9,6 +9,7 @@ TEST_DESCRIPTION="Tests for systemd-logind"
 
 test_append_files() {
     image_install -o evemu-device evemu-event crond crontab
+    image_install "${BUILD_DIR}"/test-session-properties
```
yuwata: @mrc0mmand Is this necessary? I thought our test binaries are always installed into the basic image.

test/units/testsuite-35.sh
```diff
@@ -462,6 +462,28 @@ EOF
...
+    if [[ ! -c /dev/tty2 ]]; then
+        echo "/dev/tty2 does not exist, skipping test ${FUNCNAME[0]}."
+        return
+    fi
+
+    trap cleanup_session RETURN
+    create_session
+
+    s=$(busctl call org.freedesktop.login1 /org/freedesktop/login1 org.freedesktop.login1.Manager ListSessions |\
+        awk '{ if ($5 == "\"logind-test-user\"") print $7 }' | tr -d \")
```
yuwata: Does this work if there exist multiple sessions??
I think it is safer to build session object path from the output of `loginctl`:
```
s = $(loginctl --no-legend | grep logind-test-user | awk '{ print $1 }')
/usr/lib/systemd/tests/manual/test-session-properties "/org/freedesktop/login1/session/_3${s?}"
```

yuwata (changes requested):

test/TEST-35-LOGIN/test.sh
mrc0mmand: I think it shouldn't be necessary, but the path in the test itself needs to be adjusted (to `/usr/lib/systemd/tests/test-session-properties` IIRC).

mrc0mmand: Ah, nevermind, the test already uses the path, so this can be simply dropped altogether.

test/units/testsuite-35.sh
dtardon: If it's reliable to construct the object path like this, why not... But it's even possible that there be more than one session, unless there's a bug in the test code?

test/TEST-35-LOGIN/test.sh
dtardon: Good, I'll drop it.

test/units/testsuite-35.sh
dtardon: Anyway, I changed this.

src/login/test-session-properties.c
dtardon: Done.

src/login/test-session-properties.c
dtardon: Done.

test/units/testsuite-35.sh
```diff
@@ -224,6 +228,8 @@ cleanup_session() (
     pkill -u "$(id -u logind-test-user)"
     sleep 1
     pkill -KILL -u "$(id -u logind-test-user)"
+
+    return 0
```
yuwata, line 232: Oh, nice catch. Thank you!

yuwata (approved):
LGTM.

PR #23851, opened 2022-06-28
mount: load /proc/self/mountinfo unconditionally when SIGCHLD received

lclgo:
Fixes #23796.
Replaces #23803.

PR #23883, opened 2022-07-01
core: do not filter out systemd.unit= from kernel command line

yuwata:
Fixes a bug introduced by 846f1da465beda990c1c01346311393f485df467.

The commit 846f1da465beda990c1c01346311393f485df467 made systemd.unit=
filtered out from the command line. That causes debug-generator does not
work as expected on daemon-reexecute, and we cannot call `systemctl
daemon-reexecute` in our test suite running on nspawn.

Fixes issue reported in https://github.com/systemd/systemd/pull/23851#issuecomment-1170992052.

### DIFF
--- a/test/units/testsuite-03.sh
+++ b/test/units/testsuite-03.sh
@@ -3,6 +3,10 @@
 set -eux
 set -o pipefail
 
+# Simple test for that daemon-reexec works in container.
+# See: https://github.com/systemd/systemd/pull/23883
+systemctl daemon-reexec
+
 # Test merging of a --job-mode=ignore-dependencies job into a previously
 # installed job.
 


yuwata:
cc @mrc0mmand and @keszybz

mrc0mmand:
Nice, thanks!

Could you throw a `systemctl daemon-reexec` into `TEST-03-JOBS` (since it runs only in nspawn and runs for a while, which has a high chance of hitting this issue)?

```diff
diff --git a/test/units/testsuite-03.sh b/test/units/testsuite-03.sh
index 070e978cda..67a88ec2a3 100755
--- a/test/units/testsuite-03.sh
+++ b/test/units/testsuite-03.sh
@@ -3,6 +3,9 @@
 set -eux
 set -o pipefail
 
+# See: https://github.com/systemd/systemd/pull/23883
+systemctl daemon-reexec
+
 # Test merging of a --job-mode=ignore-dependencies job into a previously
 # installed job.
 
```

should probably do the trick.

mrc0mmand (approved):
LGTM, thanks!

keszybz:
Is this safe enough to backport to v251 and earlier?

yuwata:
> Is this safe enough to backport to v251 and earlier?

Yeah, I think so.

PR #23886, opened 2022-07-02
tree-wide: use html links for kernel docs

keszybz:
Instead of using "*.txt" as reference name, use the actual destination title.


DaanDeMeyer (approved):

keszybz:
Oh, I pushed another cleanup here because I didn't see you review. Maybe take a look again, though I don't think  it should be fairly non-controversial. ;)

bluca (approved):

yuwata (approved):

PR #23888, opened 2022-07-02
network: NetLabel integration

topimiettinen:
New directive `NetLabel=` provides a method for integrating dynamic network
configuration into Linux NetLabel subsystem rules, used by Linux security
modules (LSMs) for network access control. The option expects a whitespace
separated list of NetLabel labels. The labels must conform to lexical
restrictions of LSM labels. When an interface is configured with IP addresses,
the addresses and subnetwork masks will be appended to the NetLabel Fallback
Peer Labeling rules. They will be removed when the interface is
deconfigured. Failures to manage the labels will be ignored.

Example:
```
[DHCP]
NetLabel=system_u:object_r:localnet_peer_t:s0
```

With the above rules for interface `eth0`, when the interface is configured with
an IPv4 address of 10.0.0.0/8, `systemd-networkd` performs the equivalent of
`netlabelctl` operation

```
$ sudo netlabelctl unlbl add interface eth0 address:10.0.0.0/8 label:system_u:object_r:localnet_peer_t:s0
```

Result:
```
$ sudo netlabelctl -p unlbl list
...
 interface: eth0
   address: 10.0.0.0/8
    label: "system_u:object_r:localnet_peer_t:s0"
...
```

### DIFF
--- a/test/test-network/conf/25-address-static.network
+++ b/test/test-network/conf/25-address-static.network
@@ -361,3 +361,8 @@ Address=10.3.3.251/16
 Address=10.3.3.252/16
 Address=10.3.3.253/16
 Address=10.3.3.254/16
+
+[Address]
+Address=10.4.3.2/24
+# just a random label which should exist
+NetLabel=system_u:object_r:root_t:s0
--- a/test/test-network/conf/25-dhcp-client-ipv4-only.network
+++ b/test/test-network/conf/25-dhcp-client-ipv4-only.network
@@ -26,6 +26,8 @@ SendDecline=yes
 # DenyList= will be ignored
 AllowList=192.168.5.0/24 192.168.6.0/24
 DenyList=192.168.5.0/24
+# just a random label which should exist
+NetLabel=system_u:object_r:root_t:s0
 
 [Route]
 Destination=192.168.5.0/24
--- a/test/test-network/conf/25-dhcp-pd-downstream-dummy98.network
+++ b/test/test-network/conf/25-dhcp-pd-downstream-dummy98.network
@@ -14,3 +14,5 @@ SubnetId=0
 Announce=no
 Token=eui64
 Token=::1a:2b:3c:4d
+# just a random label which should exist
+NetLabel=system_u:object_r:root_t:s0
--- a/test/test-network/conf/25-ipv6-prefix-veth.network
+++ b/test/test-network/conf/25-ipv6-prefix-veth.network
@@ -7,3 +7,5 @@ IPv6AcceptRA=true
 
 [IPv6AcceptRA]
 UseDomains=yes
+# just a random label which should exist
+NetLabel=system_u:object_r:root_t:s0
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -889,6 +889,18 @@ class Utilities():
 
         self.assertNotRegex(output, address_regex)
 
+    def check_netlabel(self, interface, address, label='system_u:object_r:root_t:s0'):
+        if not shutil.which('selinuxenabled'):
+            print(f'## Checking NetLabel skipped: selinuxenabled command not found.')
+        elif call_quiet('selinuxenabled') != 0:
+            print(f'## Checking NetLabel skipped: SELinux disabled.')
+        elif not shutil.which('netlabelctl'): # not packaged by all distros
+            print(f'## Checking NetLabel skipped: netlabelctl command not found.')
+        else:
+            output = check_output('netlabelctl unlbl list')
+            print(output)
+            self.assertRegex(output, f'interface:{interface},address:{address},label:"{label}"')
+
 class NetworkctlTests(unittest.TestCase, Utilities):
 
     def setUp(self):
@@ -2208,6 +2220,8 @@ class NetworkdNetworkTests(unittest.TestCase, Utilities):
         self.assertIn('inet6 2001:db8:0:f103::20 peer 2001:db8:0:f103::10/128 scope global', output)
         self.assertIn('inet6 2001:db8:1:f101::1/64 scope global deprecated', output)
         self.assertRegex(output, r'inet6 fd[0-9a-f:]*1/64 scope global')
+
+        self.check_netlabel('dummy98', '10\.4\.3\.0/24')
 
         # Tests for #20891.
         # 1. set preferred lifetime forever to drop the deprecated flag for testing #20891.
@@ -4267,6 +4281,9 @@ class NetworkdRATests(unittest.TestCase, Utilities):
         print(output)
         self.assertRegex(output, '2002:da8:1:0')
 
+        self.check_netlabel('veth99', '2002:da8:1::/64')
+        self.check_netlabel('veth99', '2002:da8:2::/64')
+
     def test_ipv6_token_static(self):
         copy_network_unit('25-veth.netdev', '25-ipv6-prefix.network', '25-ipv6-prefix-veth-token-static.network')
         start_networkd()
@@ -4559,6 +4576,8 @@ class NetworkdDHCPClientTests(unittest.TestCase, Utilities):
         self.assertIn('DHCPDISCOVER(veth-peer) 192.168.5.11', output)
         self.assertIn('client provides name: test-hostname', output)
         self.assertIn('26:mtu', output)
+
+        self.check_netlabel('veth99', '192\.168\.5\.0/24')
 
     def test_dhcp_client_ipv4_use_routes_gateway(self):
         first = True
@@ -5116,6 +5135,8 @@ class NetworkdDHCPPDTests(unittest.TestCase, Utilities):
         output = check_output('ip -6 route show dev dummy99')
         print(output)
         self.assertRegex(output, '3ffe:501:ffff:[2-9a-f]02::/64 proto dhcp metric [0-9]* expires')
+
+        self.check_netlabel('dummy98', '3ffe:501:ffff:[2-9a-f]00::/64')
 
     def verify_dhcp4_6rd(self, tunnel_name):
         print('### ip -4 address show dev veth-peer scope global')


topimiettinen:
Clangs fail due to timeout in test-repart and one focal fails, probably not related.

@yuwata @keszybz please re-review.

src/basic/in-addr-util.c
```diff
@@ -585,6 +585,45 @@ struct in_addr* in4_addr_prefixlen_to_netmask(struct in_addr *addr, unsigned cha
         return addr;
 }
 
+struct in6_addr* in6_addr_prefixlen_to_netmask(struct in6_addr *addr, unsigned char prefixlen) {
+        assert(addr);
+        assert(prefixlen <= 128);
+
+        for (unsigned int i = 0; i < 16; i++) {
```
poettering: as per coding style, we write simply "unsigned", never "unsigned int"

src/basic/in-addr-util.c
```diff
@@ -585,6 +585,45 @@ struct in_addr* in4_addr_prefixlen_to_netmask(struct in_addr *addr, unsigned cha
         return addr;
 }
 
+struct in6_addr* in6_addr_prefixlen_to_netmask(struct in6_addr *addr, unsigned char prefixlen) {
+        assert(addr);
+        assert(prefixlen <= 128);
```
poettering, line 606: maybe add a comment somewhere that explains briefly what this new function does in english language

src/basic/in-addr-util.c
```diff
@@ -585,6 +585,45 @@ struct in_addr* in4_addr_prefixlen_to_netmask(struct in_addr *addr, unsigned cha
...
+                        prefixlen -= 8;
+                } else if (prefixlen > 0) {
+                        mask = 0xFF << (8 - prefixlen);
+                        prefixlen = 0;
+                } else {
+                        assert(prefixlen == 0);
+                        mask = 0;
+                }
+
+                addr->s6_addr[i] = mask;
```
poettering, line 622: this looks wrong, no?

should be `|=`, no?

src/test/test-in-addr-util.c
```diff
@@ -364,4 +364,35 @@ TEST(in_addr_to_string) {
...
+        static const char *const ipv4_netmasks[] = {
+                "0.0.0.0", "128.0.0.0", "192.0.0.0", "224.0.0.0", "240.0.0.0",
+                "248.0.0.0", "252.0.0.0", "254.0.0.0", "255.0.0.0",
+                "255.128.0.0", "255.192.0.0", "255.224.0.0", "255.240.0.0",
+                "255.248.0.0", "255.252.0.0", "255.254.0.0", "255.255.0.0",
+                "255.255.128.0", "255.255.192.0", "255.255.224.0", "255.255.240.0",
+                "255.255.248.0", "255.255.252.0", "255.255.254.0", "255.255.255.0",
+                "255.255.255.128", "255.255.255.192", "255.255.255.224", "255.255.255.240",
+                "255.255.255.248", "255.255.255.252", "255.255.255.254", "255.255.255.255",
+        };
```
poettering, line 378: maybe add some examples here that actually contain one bits right of zero bits...

src/test/test-in-addr-util.c
```diff
@@ -364,4 +364,35 @@ TEST(in_addr_to_string) {
...
+
+        for (unsigned char prefixlen = 0; prefixlen <= 32; prefixlen++) {
+                _cleanup_free_ char *r = NULL;
```
poettering: we generally call `r` the local integer var that takes error return codes. maybe find another name for this. `ret` for example?

src/test/test-in-addr-util.c
```diff
@@ -364,4 +364,35 @@ TEST(in_addr_to_string) {
...
+
+                assert_se(in_addr_prefixlen_to_netmask(AF_INET, &addr, prefixlen) >= 0);
+                assert_se(in_addr_to_string(AF_INET, &addr, &r) >= 0);
+                printf("test_in_addr_prefixlen_to_netmask: %s == %s\n", ipv4_netmasks[prefixlen], r);
+                assert_se(streq(ipv4_netmasks[prefixlen], r));
+        }
+
+        for (unsigned char prefixlen = 0; prefixlen <= 128; prefixlen++) {
+                _cleanup_free_ char *r = NULL;
```
poettering: ditto

man/systemd.network.xml
```diff
@@ -2050,6 +2085,13 @@ Table=1234</programlisting></para>
           <ulink url="https://tools.ietf.org/html/rfc5227">RFC 5227</ulink>. Defaults to false.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>As in [Address] section.</para>
```
poettering: this deserves a longer comment. people might misunderstand that this would enable respecting some field in the DHCP lease that configures a label. But this is not what it does, instead it just effects traffic to/from the IP address/network configured via DHCP on this iface get labelled with that is specified here. This should be configured explicitly

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,200 @@
...
+                if (r == 0)
+                        return 0;
+
+                /* Label semantics depend on LSM but let's do basic checks */
+                if (!string_is_safe(w)) {
+                        _cleanup_free_ char *esc = NULL;
+
+                        esc = cescape(w);
+                        log_syntax(unit, LOG_WARNING, filename, line, 0,
+                                   "Bad NetLabel label, ignoring: %s", esc);
```
poettering: strna(esc), in case of OOM

poettering:
looks pretty ok to me, codewise. not a netlabel person though. maybe someone who is deeply involved with netlabel upstream could briefly state that this conceptually makes sense.

i guess @yuwata should have a look at this too

src/basic/in-addr-util.c
topimiettinen: It's like in4_addr_prefixlen_to_netmask() (which hasn't deserved any such comments so far), but with IPv6 addresses. But I can add descriptions.

src/basic/in-addr-util.c
topimiettinen: No, like `in4_addr_prefixlen_to_netmask()`, the `addr` is completely overwritten with the netmask.

src/test/test-in-addr-util.c
topimiettinen: Those would not be [netmasks](https://www.ipaddressguide.com/netmask). The function only generates masks like above, where the leftmost bits (if any) are one and the remaining rightmost bits (if any) are zero.

topimiettinen:
> looks pretty ok to me, codewise. not a netlabel person though. maybe someone who is deeply involved with netlabel upstream could briefly state that this conceptually makes sense.

There was a short discussion with @pcmoore (author of [netlabel_tools](https://github.com/netlabel/netlabel_tools) and corresponding kernel features) at #22527, but perhaps a new check is in order.

pcmoore:
Sorry @topimiettinen, this must have fallen through my inbox, or I just missed it :/

I'm not very knowledgeable on systemd, but from what I can see the genetlink communication bits looks sane from a NetLabel perspective.  However, I did have one question: does this do the right thing whenever the interface's IP address changes due to a new DHCP lease and/or local configuration change?

poettering:
btw, if the goal here is to mark packets based on the interfaces they are on, then going by IP address/mask means we'll be abiguous if the same IP address range is available on multiple interfaces. Hence, isn't there a proper way to do this in kernel, i.e. match packets by interface instead of IP range?

(my educated guess is that this is hard, because interfaces for outgoing packets are only available late, after the routing decision took place, but still makes me wonder if that wouldn't make more sense? or is the goal to actually influence routing decisions based on netlabel?)

poettering:
Ok, so let's maybe take a step back: @topimiettinen what's your usecase for all this? How do you intend to use this? Can you elaborate?

topimiettinen:
> btw, if the goal here is to mark packets based on the interfaces they are on, then going by IP address/mask means we'll be abiguous if the same IP address range is available on multiple interfaces. Hence, isn't there a proper way to do this in kernel, i.e. match packets by interface instead of IP range?

The labeling command indeed specifies the interface. It's possible to use default rules which apply to all interfaces but we're not using them here.

> (my educated guess is that this is hard, because interfaces for outgoing packets are only available late, after the routing decision took place, but still makes me wonder if that wouldn't make more sense? or is the goal to actually influence routing decisions based on netlabel?)

The labels can be used for access control where the source IP address determines the label. With SELinux, the labels can be used with `peer` class, for example:
```
allow mozilla_t localnet_peer_t:peer recv;
```
Domains without such allow rules can't receive anything from addresses labeled `localnet_peer_t`.

topimiettinen:
> It looks like this PR is only exposing the static/fallback labeling functionality of NetLabel which assigns a security label to _incoming_ traffic that matches a particular address/network.

Yes. I don't think CIPSO/CALIPSO would need this kind of dynamic configuration.

src/basic/in-addr-util.c
topimiettinen: fixed

src/basic/in-addr-util.c
topimiettinen: Added comments

src/test/test-in-addr-util.c
topimiettinen: r -> result

src/test/test-in-addr-util.c
topimiettinen: ditto

man/systemd.network.xml
topimiettinen: Improved the comment

src/network/networkd-netlabel.c
topimiettinen: fixed

topimiettinen:
Rebased and updated according to review comments.

Other changes:
- there can be only one NetLabel for (interface, address) tuple :facepalm:
- when adding, kernel adds the address to hash table _applying also the netmask_ ([net/netlabel/netlabel_unlabeled.c](https://github.com/torvalds/linux/blob/68e77ffbfd06ae3ef8f2abf1c3b971383c866983/net/netlabel/netlabel_unlabeled.c#L243)), but on removal, an exact match is required _without netmask applied_ ([net/netlabel/netlabel_addrlist.c](https://github.com/torvalds/linux/blob/68e77ffbfd06ae3ef8f2abf1c3b971383c866983/net/netlabel/netlabel_addrlist.c#L244)), so apply the mask on both operations.
- don't remove NetLabels unconditionally but only if configuration exists

I didn't test the DHCP address change yet.

src/test/test-in-addr-util.c
```diff
@@ -364,4 +364,35 @@ TEST(in_addr_to_string) {
...
+                assert_se(streq(ipv4_netmasks[prefixlen], result));
+        }
+
+        for (unsigned char prefixlen = 0; prefixlen <= 128; prefixlen++) {
+                _cleanup_free_ char *result = NULL;
+
+                assert_se(in_addr_prefixlen_to_netmask(AF_INET6, &addr, prefixlen) >= 0);
+                assert_se(in_addr_to_string(AF_INET6, &addr, &result) >= 0);
+                printf("test_in_addr_prefixlen_to_netmask: %s\n", result);
+        }
```
poettering, line 416: maybe for some corner cases (i.e. prefixlen=0, =1, =2, =7, =8, =9, =63, =127, =128) some explicit string checks would make sense here? i understand you don#t want to carry the full table with 128 entries here, but i thing some selected corner cases should still be checked.

man/systemd.network.xml
```diff
@@ -1121,6 +1121,41 @@ Table=1234</programlisting></para>
...
+
+          <para>This setting provides a method for integrating dynamic network configuration into Linux
+          <ulink url="https://docs.kernel.org/netlabel/index.html">NetLabel</ulink> subsystem rules, used by
+          <ulink url="https://en.wikipedia.org/wiki/Linux_Security_Modules">Linux security modules
+          (LSMs)</ulink> for network access control. The option expects a single NetLabel label. The label
+          must conform to lexical restrictions of LSM labels. When an interface is configured with IP
+          addresses, the addresses and subnetwork masks will be appended to the <ulink
+          url="https://github.com/SELinuxProject/selinux-notebook/blob/main/src/network_support.md">NetLabel
+          Fallback Peer Labeling</ulink> rules. They will be removed when the interface is
+          deconfigured. Failures to manage the labels will be ignored.</para>
```
poettering, line 1147: This should really explain what this all means. i.e. why one would do this.

i.e. write this for somebody who is *not* at home in the NetLabel world, for example me.

man/systemd.network.xml
```diff
@@ -1121,6 +1121,41 @@ Table=1234</programlisting></para>
...
+
+          <para>Warning: Once labeling is enabled for network traffic, a lot of LSM access control points in
+          Linux networking stack go from dormant to active. It is easy for someone not familiar with the LSM
```
poettering: i'd not make this personal. i.e. remove "it is easy for someone not familiar with the blablablabla" and write instead something like "Care should be taken, not to blablablabla, because blablabla"

man/systemd.network.xml
poettering: also, this doesn't mention whether this is about ingress or egress

man/systemd.network.xml
```diff
@@ -2062,6 +2097,13 @@ Table=1234</programlisting></para>
           <ulink url="https://tools.ietf.org/html/rfc5227">RFC 5227</ulink>. Defaults to false.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>See <varname>NetLabel=</varname> in [Address] section.</para>
```
poettering: referencing the [Address] section is good, but please explicitly say that this option here applies labels to the DHCP acquired IP addresses the same way as NetLabel in [Address] applies them to statically configured ones.

man/systemd.network.xml
```diff
@@ -2276,6 +2319,13 @@ Table=1234</programlisting></para>
           </para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>See <varname>NetLabel=</varname> in [Address] section.</para>
```
poettering: similar

man/systemd.network.xml
```diff
@@ -2533,6 +2583,13 @@ Token=prefixstable:2002:da8:1::</programlisting></para>
           specified. Defaults to true.</para>
         </listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>NetLabel=</varname></term>
+        <listitem>
+          <para>See <varname>NetLabel=</varname> in [Address] section.</para>
```
poettering: ditto

src/network/networkd-address.h
```diff
@@ -61,6 +61,9 @@ struct Address {
 
         /* Called when address become ready */
         address_ready_callback_t callback;
+
+        /* NetLabel */
```
poettering: that is one useless comment. If the comment consists of just the name of the variable with different capitalization, maybe drop the comment.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,193 @@
...
+        case NETWORK_CONFIG_SOURCE_DHCP_PD:
+                return address->link->network->dhcp_pd_netlabel;
+        case NETWORK_CONFIG_SOURCE_NDISC:
+                return address->link->network->ndisc_netlabel;
+        case NETWORK_CONFIG_SOURCE_STATIC:
+                return address->netlabel;
+        default:
+                return NULL;
+        }
+        return NULL;
```
poettering: redundant line

poettering:
did another review round. codewise looks ok, but I'll leave the final review to @yuwata.

The man page could use some love, i.e. for me as someone who is not at home in selinux/netlabel that much, I personally cannot understand from reading this what this is good for. Write docs for someone who needs docs, i.e. might be 50% there, but not 100%...

poettering:
integration tests would be good, that verify that the label rules are properly installed.

src/test/test-in-addr-util.c
topimiettinen: Added some checks.

man/systemd.network.xml
topimiettinen: This is a new option to complement rather advanced setup with NetLabels, so this will never be a beginner level option, but I tried to put more descriptive text here. About ingress, I'm not sure how Smack uses NetLabels and if it has the same limitations but at least for SELinux the checks are only for receiving side.

man/systemd.network.xml
topimiettinen: Fixed

man/systemd.network.xml
topimiettinen: Now more explicit

man/systemd.network.xml
topimiettinen: ditto

src/network/networkd-address.h
topimiettinen: Dropped the comment and merged the variable definition with another.

src/network/networkd-netlabel.c
topimiettinen: I hope the compilers are smarter than they used to be, so dropped the line.

topimiettinen:
> integration tests would be good, that verify that the label rules are properly installed.

This would be pretty hard to test completely. The setup should have SELinux (or maybe Smack) enabled and `netlabelctl` tools installed. Then SELinux policy should not make the test system lose connectivity (assuming that it's needed to get information out of the test system) when the network controls are enabled and I think distro SELinux policies may choke on that point. The policy should be also enhanced to allow `networkd` to set up NetLabel rules. Perhaps the system could be booted with `enforcing=0`. A test label should be defined by the policy (the rule can't be installed if the label isn't defined first), but maybe it could be just a random existing type label like `root_t`. Then `networkd` should be made to toggle an interface up and down while `netlabelctl` (or a new test tool) would be used to read the rules.

topimiettinen:
I made some tests with two DHCP servers. Switching between them is fine, the labels are removed before new ones are installed.

pcmoore:
> The setup should have SELinux (or maybe Smack) enabled ...

It's been a while since I discussed this with the Smack folks, but if I remember correctly, in the case of Smack the NetLabel configuration should be done via Smack and not directly using the NetLabel config tools.  Casey would be the best one to ask about this (not sure if he is on GH).

pcmoore:
> I made some tests with two DHCP servers. Switching between them is fine, the labels are removed before new ones are installed.

Thanks for all your work on this @topimiettinen, including testing all these different scenarios.

topimiettinen:
@yuwata please review.

man/systemd.network.xml
```diff
@@ -1121,6 +1121,62 @@ Table=1234</programlisting></para>
...
+      <varlistentry>
+        <term><varname>NetLabel=</varname><replaceable>label</replaceable></term>
+        <listitem>
+
+          <para>This setting provides a method for integrating static and dynamic network configuration into
+          Linux <ulink url="https://docs.kernel.org/netlabel/index.html">NetLabel</ulink> subsystem rules,
+          used by <ulink url="https://en.wikipedia.org/wiki/Linux_Security_Modules">Linux Security Modules
+          (LSMs)</ulink> for network access control. The label, with suitable LSM rules, can be used to
+          control connectivity of (for example) a service with peers in the local network. At least with
+          SELinux, only the ingress can be controlled but not egress. The benefit of using this setting is
```
poettering, line 1134: does any LSM other than SELinux actually care about NetLabel?

Sounds to me, if the answer to that question is "no", we shouldn't mention LSM here, and just say SELinux.

poettering:
looks pretty OK to me. Can't judge how useful this IRL, but codewise all fine be me. I'll leave merging to @yuwata

man/systemd.network.xml
topimiettinen: Smack [seems to care](https://wiki.tizen.org/Security:SmackNetworking) but the interface may be different. If so, it would need some work to add Smack support (by someone knowledgeable). I suppose the concept of NetLabel configuration by networkd still makes sense. Perhaps the documentation should warn that only SELinux is implemented now.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+        default:
+                return NULL;
+        }
+}
+
+void address_add_netlabel(const Address *address) {
+        int r;
+        assert(address);
+
+        const char *label;
```
yuwata: Please insert a empty line between the variable declaration(s) and assertion(s).
Nowadays, we allow to declare variables when needed, but in this case, I prefer to declare `label` at the beginning of the function.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+
+        if (command == NLBL_UNLABEL_C_STATICADD) {
+                assert(label);
+                r = sd_netlink_message_append_string(m, NLBL_UNLABEL_A_SECCTX, label);
+                if (r < 0)
+                        return r;
+        }
+
+        union in_addr_union netmask, masked_addr;
+
```
yuwata: Please drop the empty line.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+                const char *filename,
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
+        char **label = ASSERT_PTR(data);
```
yuwata: Please insert an empty line before variable declaration and function arguments.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+                           "Bad NetLabel label, ignoring: %s", strna(esc));
+                return 0;
+        }
+
+        r = free_and_strdup(label, rvalue);
+        if (r < 0)
+                return log_oom();
+
+        return 0;
+}
```
yuwata: You can use `free_and_strdup_warn()`:
```
return free_and_strdup_warn(label, rvalue);
```

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+        case NETWORK_CONFIG_SOURCE_DHCP4:
+                return address->link->network->dhcp_netlabel;
+        case NETWORK_CONFIG_SOURCE_DHCP6:
+                return address->link->network->dhcp6_netlabel;
+        case NETWORK_CONFIG_SOURCE_DHCP_PD:
+                return address->link->network->dhcp_pd_netlabel;
+        case NETWORK_CONFIG_SOURCE_NDISC:
+                return address->link->network->ndisc_netlabel;
+        case NETWORK_CONFIG_SOURCE_STATIC:
+                return address->netlabel;
```
yuwata: Here, the `Address` object is owned by `Link`. But, `Address->netlabel` is owned by another `Address` object owned by `Network` object. Hence, if we configure a static address, and soon later the .network file is modified and the interface is reconfigured, then accessing `address->netlabel` causes use-after-free.

Please update `address_dup()` to make the netlabel copied.

Also, please copy netlabels for dynamic addresses before `link_request_address()`. E.g. `dhcp4_request_address()` in networkd-dhcp4.c.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+        else
+                log_link_debug(address->link, "Adding NetLabel %s for IP address %s", label,
+                               IN_ADDR_PREFIX_TO_STRING(address->family, &address->in_addr, address->prefixlen));
+}
+
+void address_del_netlabel(const Address *address) {
+        int r;
+        assert(address);
+
+        const char *label;
```
yuwata: Also here.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+        _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
+        int r;
+
+        assert(address);
+        assert(address->link);
+        assert(address->link->ifname);
+        assert(address->link->manager);
+        assert(address->link->manager->genl);
+        assert(address->link->network);
+        assert(IN_SET(address->family, AF_INET, AF_INET6));
```
yuwata, line 39: Maybe, better to add an assertion for `command`?

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+static int netlabel_command(uint16_t command, const char *label, const Address *address) {
+        _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
+        int r;
+
+        assert(address);
+        assert(address->link);
+        assert(address->link->ifname);
+        assert(address->link->manager);
+        assert(address->link->manager->genl);
+        assert(address->link->network);
```
yuwata: `address->link->network` seems not used. Is the assertion necessary?

yuwata (changes requested):
Sorry for late review.

Please add test cases for the feature. At least one for static address and another for dynamic address.

yuwata:
Please also rebase.

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,197 @@
...
+}
+
+void address_del_netlabel(const Address *address) {
+        int r;
+        assert(address);
+
+        const char *label;
+        label = get_label(address);
+        if (!label)
+                return;
```
yuwata: Is this necessary? On removal, the netlink message does not contain label. I think we should unconditionally remove netlabel for the address.

src/network/networkd-netlabel.c
topimiettinen: Sorry, I'm not sure if I follow. The various netlabel strings as returned by this function are only used immediately in `netlabel_command()`, where the string is copied to the netlink command, and then second time when logging on exit of `address_add_netlabel()` or `address_del_netlabel()`. Is this use-after-free scenario then possible? I'd think that it would require that for example the async callback would access the label string, which is not the case.

src/network/networkd-netlabel.c
topimiettinen: That could conflict with labels configured with for example `netlabelctl` when there's no networkd configuration for NetLabels. I'd only remove the label if and only if we installed it.

topimiettinen:
> Sorry for late review.

Many thanks for the review!

> Please add test cases for the feature. At least one for static address and another for dynamic address.

Unfortunately testing even just only adding or removing labels is bit hard, since the NetLabel for SELinux can't be added unless the policy defines the label and it allows networkd to add it (unless the whole system or networkd domain is in permissive mode). Also SELinux needs to be enabled. So I suppose the test wouldn't be able to run on usual CI systems.

man/systemd.network.xml
topimiettinen: Added a note that SELinux is only implemented

src/network/networkd-netlabel.c
topimiettinen: fixed

src/network/networkd-netlabel.c
topimiettinen: dropped

src/network/networkd-netlabel.c
topimiettinen: inserted

src/network/networkd-netlabel.c
topimiettinen: Thanks, much better.

src/network/networkd-netlabel.c
topimiettinen: No, removed

src/network/networkd-netlabel.c
topimiettinen: added assert

src/network/networkd-netlabel.c
topimiettinen: fixed

topimiettinen:
I've added a test for static and DHCP addresses, let's see how it goes.

topimiettinen:
@mrc0mmand `autopkgtest` fails because package `netlabel-tools` (Fedora: `netlabel_tools`) isn't installed, but I've added it to every package list I found. Any idea what am I missing?

mrc0mmand:
> @mrc0mmand `autopkgtest` fails because package `netlabel-tools` (Fedora: `netlabel_tools`) isn't installed, but I've added it to every package list I found. Any idea what am I missing?

`autopkgtest` stuff is handled outside of this repo - https://salsa.debian.org/systemd-team/systemd/-/blob/upstream-ci/debian/tests/control#L119 - this is something @bluca can tweak as necessary

bluca:
Updated and restarted all the jobs https://salsa.debian.org/systemd-team/systemd/-/commit/de985ca1a621ee051a31ea40f4d795df0fa79c88

src/network/networkd-netlabel.c
```diff
@@ -0,0 +1,194 @@
...
+                const char *unit,
+                const char *filename,
+                unsigned line,
+                const char *section,
+                unsigned section_line,
+                const char *lvalue,
+                int ltype,
+                const char *rvalue,
+                void *data,
+                void *userdata) {
```
yuwata: This can be dropped and replaced with `config_parse_string()` with `CONFIG_PARSE_STRING_SAFE`.

mkosi.default.d/centos_epel/10-mkosi.centos_epel
```diff
@@ -88,3 +88,4 @@ Packages=
         wireguard-tools
         dhcp-server
         kernel-modules-extra
+        netlabel_tools
```
yuwata: I do not think it is necessary to add `netlabel_tools` for mkosi files.

src/network/networkd-netlabel.c
yuwata: I mean, `address->label` here may be already freed, as it is owned by _another_ `Address` object managed by `Network`.

yuwata (changes requested):

src/network/networkd-netlabel.c
topimiettinen: Updated to copy the labels.

src/network/networkd-netlabel.c
topimiettinen: Nice, replaced.

mkosi.default.d/centos_epel/10-mkosi.centos_epel
topimiettinen: Removed

src/network/networkd-address.c
```diff
@@ -399,6 +401,10 @@ int address_dup(const Address *src, Address **ret) {
                         return r;
         }
 
+        r = free_and_strdup(&dest->netlabel, src->netlabel);
```
yuwata, line 405: Please add `dest->netlabel = NULL;` in the above. Otherwise, `dest->netlabel` and `src->netlabel` points to the same memory, and causes use-after-free.

src/network/networkd-address.c
```diff
@@ -1940,6 +1950,41 @@ int config_parse_duplicate_address_detection(
...
+        r = address_new_static(network, filename, section_line, &n);
+        if (r == -ENOMEM)
+                return log_oom();
+        if (r < 0) {
+                log_syntax(unit, LOG_WARNING, filename, line, r,
+                           "Failed to allocate new address, ignoring assignment: %m");
+                return 0;
+        }
+
+        return config_parse_string(unit, filename, line, section, section_line, lvalue, ltype, rvalue, &n->netlabel, network);
```
yuwata: `ltype` -> `CONFIG_PARSE_STRING_SAFE`

test/networkd-test.py
```diff
@@ -457,6 +458,18 @@ def do_test(self, coldplug=True, ipv6=False, extra_opts='',
         else:
             self.fail('nameserver 192.168.5.1 not found in ' + RESOLV_CONF)
 
+        if netlabel and selinux.is_selinux_enabled():
```
yuwata: Could you split out about adding tests into another commit? And instead of networkd-test.py, please update test-network/systemd-networkd-tests.py. networkd-test.py is not maintained.

mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -78,7 +78,9 @@ Packages=
         gcc
         # Required to run systemd-networkd-tests.py
         python3
+        python3-selinux
         iproute2
         dnsmasq-base
         wireguard-tools
         isc-dhcp-server
+        netlabel-tools
```
yuwata: Ah, you added these packages for systemd-networkd-tests.py. Sorry, please forget my previous comment, and also add them for other distro?

yuwata (changes requested):

src/network/networkd-address.c
topimiettinen: Fixed

src/network/networkd-address.c
topimiettinen: Thanks, updated

test/networkd-test.py
topimiettinen: Split commits. I moved the tests over to systemd-networkd-tests.py and also added tests for `IPv6AcceptRA.NetLabel=` and `DHCPPrefixDelegation.NetLabel=`.

mkosi.default.d/debian/10-mkosi.debian
topimiettinen: Added. Arch doesn't seem to have netlabel_tools packaged.

topimiettinen:
OpenSUSE fails with
'FileNotFoundError: [Errno 2] No such file or directory: '/var/tmp/mkosi-ec67po6p/root/usr/etc/pam.d/login'
and Arch with
FAILED TASKS:
-------------
TEST-02-UNITTESTS,
probably not related.

topimiettinen:
Rebased

yuwata:
Force-pushed:
- fixed genl family name,
- sort genl attributes,
- dropped netlabelctl from mkosi again, as it is for networkd-test.py, not systemd-networkd-tests.py,
- many fixes and cleanups for systemd-networkd-tests.py,
- rearrange commits.

yuwata:
@mrc0mmand Could you install netlabel_tools in CentOS CI environments?

mrc0mmand:
> @mrc0mmand Could you install netlabel_tools in CentOS CI environments, and restart CentOS CIs?

Done. Unfortunately, I can't restart passed jobs, so you'll need to do a force-push.

yuwata:
> > @mrc0mmand Could you install netlabel_tools in CentOS CI environments, and restart CentOS CIs?
> 
> Done. Unfortunately, I can't restart passed jobs, so you'll need to do a force-push.

@mrc0mmand Thanks. Now force-pushed.

Issue #23703, opened 2022-06-10
Main branch source code is not formatted despite `.clang-format` present 


igo95862:
**systemd version the issue has been seen with**
 > main branch

**Unexpected behaviour you saw**
 > Running `ninja clang-format` on main branch would format a large amount of files 

**Steps to reproduce the problem**
 > meson setup build
> cd build
> ninja clang-format

In total there are 1811 not formatted source code files.

`git ls-files --modified | wc --lines`

I am running in to issues writing C code because if I try to format a large part of code I am not working on will get formatted and mess up trying to create git commits.

I would recommend setting up a Continuous Integration to verify that source code is properly formatted.

yuwata:
Please open a PR to fix trivial formatting issue.
But we sometimes intentionally break format. E.g. https://github.com/systemd/systemd/blob/288bafa952b1d515e10bdf8c8f0adde3acbbe4be/src/boot/efi/boot.c#L484-L495
So, I guess it is hard to introduce a CI about formatting.

poettering:
It was decided to not make clang formatting unconditional, because iirc it couldn't properly express the formatting we so far use, in particular when it comes to aligning tables (i.e. literal arrays of structs), and corrupting those is a bad idea, as is adding reformatting escapes around them all...

Hence, the clang formatting rules we provide are supposed to help, but they do not imply a regime, and as long as the formatter doesn't get smarter regarding such tables I don't see that changing...

igo95862:
> particular when it comes to aligning tables (i.e. literal arrays of structs),

There is a new option `AlignArrayOfStructures` in clang format 13

```
struct test demo[] =
{
    {56, 23,    "hello"},
    {-1, 93463, "world"},
    {7,  5,     "!!"   }
};

```
```
struct test demo[] =
{
    {56,    23, "hello"},
    {-1, 93463, "world"},
    { 7,     5,    "!!"}
};
```

PR #23702, opened 2022-06-10
Fix ObjectManager interfaces for `GetManagedObjects`, `InerfacesAdded` and `InterfacesRemoved`

igo95862:
Objects without ObjectManager should not have `org.freedesktop.DBus.ObjectManager` interface.
Objects with ObjectManager should do.

I started with a test that verifies that `GetManagedObjects` has correct interfaces. It fails right because current behavior is bugged but I am working on fix next.

Related issues:

#8444
#8448

Issue #15415, opened 2020-04-12
Please add a .clang-format file


nolange:
I am contributing to various projects, and the differing Coding styles
are a big timesink for making code "fitting". I guess some of your time is wasted as well by pointing out violations.

So I would ask you to "invest" in a best-fit .clang-format, even if its not a 100% fit its a good starting point. 
In case you don't know, `git clang-format [--diff]` will apply/show corrections only on changed lines (possible to integrate this into CI)

poettering:
See: #11811

keszybz:
This is never going to happen until clang-format gets much much smarter. There were attempts previously, but we have lots of formatting rules to make the code readable in different contexts. Automatic formatting is generally not flexible enough to support this.

I'll close this issue to avoid false hope.

nolange:
Thing is, that I will still use clang-format or equivalents for systemd, even if this is not deemed sufficient its better than having nothing.
So just throw it under `tools/` with some notes that this is no substitute to reading an adhering to the style guide.

keszybz:
Indeed, we should probably do that.

nolange:
Thanks for making this happen.

It seems though, that the backslashes are messing up the `ForEachMacros`.
Thats what I get after a `clang-format --style=file --dump-config`:
(which I would recommend before committing this)

```
ForEachMacros:
  - BITMAP_FOREACH
  - CMSG_FOREACH
  - _DNS_ANSWER_FOREACH
  - DNS_ANSWER_FOREACH
  - '\
  _DNS_ANSWER_FOREACH_FLAGS'
  - DNS_ANSWER_FOREACH_FLAGS
  - _DNS_ANSWER_FOREACH_FULL
  - '\
  DNS_ANSWER_FOREACH_FULL'
```

PR #11254, opened 2018-12-23
code style format: clang-format configuration suggestion

zebastian:
The following .clang-format matches to ~93%, meaning that 93% of the lines
of the source code are left untouched (tested with latest commit and
clang-format-6.0.

In order to test the auto formatting, you will need to install clang-format
(eg.: apt-get install clang-format-6.0).
For more information, see the file header of the .clang-format file in this
patch.

I haven't seen any discussions on using an automated code style system on
the systemd code base yet. So i thought you may find this useful. There are
a couple of "issues" this file addresses, for example consistent pointer
alignment, some fixes for spacing and consistent line breaking rules.
On the other hand clang-format also destroys some custom parameter block alignments.

Let me know what you think of this. I can adjust some of the rules in this
file, but these rules come already pretty close to the coding style
already present in the code base.

More information on clang-format:

ClangFormat reformats source code of different programming languages
(mostly C and C++) and can apply a consistent coding style.
clang-format is non destructive and does not change the semantics of the code.

ClangFormat is a CLI-tool which uses the clang-tokenizer to tokenize the code
and recombine the code again by applying a certain specified code style.
The code style can be defined in a .clang-format file and can be tweaked to
apply to almost any "common" style. In comparison to astyle, ClangFormat styles
more aggressively and has more configuration options.
More information here: https://clang.llvm.org/docs/ClangFormat.html

ClangFormat is present in most GNU/Linux based Distros (Debian, Fedora, Ubuntu,
...). Current common version in recent Distro images are 5.0.0 - 6.0.0.
The versions of ClangFormat are not backward or forward compatible, but the
style format is forward compatible and the style only changes slightly in
newer versions. Also some package managers keep different versions of
ClangFormat to be installed at any time.

PR #23890, opened 2022-07-03
clang-format all test source files

igo95862:
Makes it easier to develop new tests as running clang format
on newly developed test won't cause other parts of file get
reformatted.

If some parts are not supposed to be formatted it is possible to add `/* clang-format off */` `/* clang-format on */` comments to disable formatting for certain areas.

See #23703


evverx:
I think the following comment
https://github.com/systemd/systemd/blob/5932c870004fbcb3c25ca3fceb92b640f3406c88/.clang-format#L1-L5
was included in `.clang-format` for a reason.

Other than that as discussed back in 2018 in https://github.com/systemd/systemd/pull/11254#issuecomment-450481407 I don't think it makes much sense to keep it unless it can be safely applied automatically. I'm not sure why `.clang-format` was added in the first place. Judging by the state of the codebase it doesn't seem to be used at all.

igo95862:
I feel like test source are more of the write once and forget so auto formatting would be more appropriate.

The reason why I created this PR is #23702 . I usually write some code out of style and when press auto-format. But because the files would be reformatted far beyond where I added code it would force me to use `git add --patch` and spend a lot more time.

evverx:
> I feel like test sources are more of the write once and forget so auto formatting would be more appropriate.

The tests get updated to cover new features, to make sure that old bugs can't make it to the repository again and so on. If they were formatted with `.clang-format` they wouldn't be consistent with the rest of the codebase and I don't think it's ideal given that the idea behind auto-formatting in general is to make code consistent and to make it easier for everyone (including CI) to follow style guides automatically.

> The reason why I created this PR is https://github.com/systemd/systemd/pull/23702 . I usually write some code out of style and when press auto-format. But because the files would be reformatted far beyond where I added code it would force me to use git add --patch and spend a lot more time.

I agree it would be great if it was possible to auto-format the code but unfortunately `.clang-format` in its current form can't be used for that.

Just to clarify I'm not opposed to the idea in general and I think it makes sense to use tools like that when it's possible (for example I don't think I have ever seen rust or go projects that haven't been auto-formatted or where the standard guidelines haven't been imposed by their CI). It's just that `systemd` can't be covered by `.clang-format` and maybe it would make sense to remove it so as not to confuse contributors. As far as I understand it isn't actually used.

igo95862:
> in its current form can't be used for that.

Is there a list of issues with clang format?

For example, I found that clang format wants to sort imports but some systemd headers depend on custom order. I made a pull request to disable include sorting: #23891

evverx:
> Is there a list of issues with clang format?

I'm not sure. I think some of them might have probably been mentioned in the PR where `.clang-format` was introduced (along with that comment).

I think the easiest way would be to try to reformat `systemd` and compare the result with "CODING_STYLE".

evverx:
@keszybz looking at https://github.com/systemd/systemd/commit/f3a604e4cf84f00a595aceca6c3e78fd03de652d I still have no idea why it was merged. If it should be kept for some reason I wonder if it would make sense to move it to `tools` (as was suggested in https://github.com/systemd/systemd/issues/15415#issuecomment-613418940) so that editors wouldn't pick it up automatically?

igo95862:
> > I feel like test sources are more of the write once and forget so auto formatting would be more appropriate.
> 
> No, not at all.

The reason why I thought that is because there was a big lack of comments and poor variable names in the test files. (at least in the files that I worked with)

igo95862:
@keszybz I think I will close it for now.

Once the #23702 gets resolved. I think I will revisit formatting the file that I worked on. Maybe look in to improving the clang format file to better fit the project formatting code. (the #23891 already has the clang format improvements)

PR #23891, opened 2022-07-03
Disable include sorting for clang format

igo95862:
`libc-compat.h` is required to be in specific place otherwise
redefinition error will be raised.

Clang format wants to sort includes alphabetically.

Alternatively `/* clang-format off */ /* clang-format on */` comments can be included everywhere sorting includes would break the build.

See #23703.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -118,6 +118,7 @@ PenaltyBreakString: 0
 PenaltyExcessCharacter: 10
 PenaltyReturnTypeOnItsOwnLine: 100
 PointerAlignment: Right
+SortIncludes: false
 SpaceAfterCStyleCast: true
 SpaceAroundPointerQualifiers: Both
 SpaceBeforeParens: ControlStatementsExceptForEachMacros


yuwata:
> `libc-compat.h` is required to be in specific place

It is not directly included in our code, except for imported kernel headers.

DaanDeMeyer:
This is probably because you ran clang-format indiscriminately over some included kernel headers? Regardless, we should fix headers that depend on specific include orders instead of disabling include sorting altogether, hence closing

igo95862:
> This is probably because you ran clang-format indiscriminately over some included kernel headers? Regardless, we should fix headers that depend on specific include orders instead of disabling include sorting altogether, hence closing

No. I ran over all test files: #23890 

In one of them changing include order broke compilation.

DaanDeMeyer:
See https://github.com/systemd/systemd/pull/23914 for a fix

igo95862:
> See #23914 for a fix

Great!

Should I look in to if there are any more of those issues with sorted imports?

DaanDeMeyer:
Sure, if there's a way to have clang-format only check if includes are sorted, we can make that a check in CI

igo95862:
> Sure, if there's a way to have clang-format only check if includes are sorted, we can make that a check in CI

Why not just format with a clang format and test if it compiles.

I found two more issues this way:

```
../src/shared/libcrypt-util.h:9:45: error: ‘NULL’ undeclared (first use in this function)
```
```
In file included from ../src/test/test-firewall-util.c:5:
../src/shared/firewall-util-private.h:28:22: error: unknown type name ‘FirewallContext’
```

PR #23892, opened 2022-07-03
meson: Assign tests a suite based on their directory

daandemeyer:
This can be used to run only a subset of tests, e.g.
"meson test -C build --suite journal" to run only the journal
unit tests.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3794,6 +3794,10 @@ foreach tuple : tests
                 type = ''
         endif
 
+        suite = fs.name(fs.parent('@0@'.format(sources[0])))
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -3794,6 +3794,18 @@ foreach tuple : tests
                 type = ''
         endif
 
+        path = '@0@'.format(sources[0])
```
medhefgo: `fs.parent`?

daandemeyer: Fixed but we still have to use `'@0@'.format(sources[0])` because the fs module only works on file objects from meson 0.59.0 onwards.

medhefgo (approved):

PR #23893, opened 2022-07-04
core/mount: adjust deserialized state based on if the corresponding mountinfo entry exists or not

### LINKED ISSUE #23796: systemd: mount units fail with "Mount process finished, but there is no mount."

yuwata:
Fixes #23796.
Replaces #23803 and #23851.

### DIFF
--- a/test/units/testsuite-60.sh
+++ b/test/units/testsuite-60.sh
@@ -53,6 +53,45 @@ EOF
     }
 }
 
+test_issue_23796() {
+    local mount_path mount_mytmpfs
+
+    mount_path="$(command -v mount 2>/dev/null)"
+    mount_mytmpfs="${mount_path/\/bin/\/sbin}.mytmpfs"
+    cat >"$mount_mytmpfs" <<EOF
+#!/bin/bash
+sleep ".\$RANDOM"
+exec -- $mount_path -t tmpfs tmpfs "\$2"
+EOF
+    chmod +x "$mount_mytmpfs"
+
+    mkdir -p /run/systemd/system
+    cat >/run/systemd/system/tmp-hoge.mount <<EOF
+[Mount]
+What=mytmpfs
+Where=/tmp/hoge
+Type=mytmpfs
+EOF
+
+    # shellcheck disable=SC2064
+    trap "rm -f /run/systemd/system/tmp-hoge.mount '$mount_mytmpfs'" RETURN
+
+    for ((i = 0; i < 10; i++)); do
+        systemctl --no-block start tmp-hoge.mount
+        sleep ".$RANDOM"
+        systemctl daemon-reexec
+
+        sleep 1
+
+        if [[ "$(systemctl is-failed tmp-hoge.mount)" == "failed" ]] || \
+           journalctl -u tmp-hoge.mount -q --grep "but there is no mount"; then
+                exit 1
+        fi
+
+        systemctl stop tmp-hoge.mount
+    done
+}
+
 : >/failed
 
 systemd-analyze log-level debug
@@ -110,6 +149,9 @@ timeout 2m bash -c 'while systemctl list-units -t mount tmp-meow* | grep -q tmp-
 
 # test that handling of mount start jobs is delayed when /proc/self/mouninfo monitor is rate limited
 test_issue_20329
+
+# test for reexecuting with background mount job
+test_issue_23796
 
 systemd-analyze log-level info
 


yuwata:
cc @Geass-LL and @jcg190701.

src/core/manager.c
```diff
@@ -1757,6 +1757,10 @@ static void manager_ready(Manager *m) {
                         UID_INVALID, GID_INVALID, 0444);
 
         m->honor_device_enumeration = true;
+
+        /* Re-read /proc/self/mountinfo when the first SIGCHLD event for mount unit is received. As, the
+         * mount states set by mount_enumerate() may be overwritten by the outdated deserialized states. */
```
lclgo: I think it's better to also mention that `mount_enumerate may failed to catch the change event`.

yuwata: What is "the change event"??

lclgo: I mean the change of `/proc/self/mountinfo`.
Emm... tought twice, I was wrong. The mount states set by `mount_enumerate()` should always be up-to-date, even if the libmount monitor is closed when systemd reexcutes.
The comment dosn't need any modify.

jcg190701 (approved):

test/units/testsuite-60.sh
```diff
@@ -53,6 +53,51 @@ EOF
...
+EOF
+    chmod +x "$MOUNT_MYTMPFS"
+
+    mkdir -p /run/systemd/system
+    cat >/run/systemd/system/tmp-hoge.mount <<EOF
+[Mount]
+What=mytmpfs
+Where=/tmp/hoge
+Type=mytmpfs
+EOF
```
mrc0mmand: This is going to break on systems where `mount` is already in `/sbin`, like Arch:

```
# command -v mount
/usr/sbin/mount
# MOUNT=$(command -v mount 2>/dev/null)
# echo "$(echo "$MOUNT" | sed -e 's|bin|sbin|')".mytmpfs
/usr/ssbin/mount.mytmpfs
```

What about something like this:

```suggestion
    local mount_path mount_mytmpfs

    mount_path="$(command -v mount 2>/dev/null)"
    mount_mytmpfs="${mount_path/\/bin/\/sbin}.mytmpfs"

    cat >"$mount_mytmpfs" <<EOF
#!/bin/bash

sleep ".\$RANDOM"
exec -- $mount_path -t tmpfs tmpfs "\$2"
EOF
    chmod +x "$mount_mytmpfs"

    mkdir -p /run/systemd/system
    cat >/run/systemd/system/tmp-hoge.mount <<EOF
[Mount]
What=mytmpfs
Where=/tmp/hoge
Type=mytmpfs
EOF

    # shellcheck disable=SC2064
    trap "rm -f /run/systemd/system/tmp-hoge.mount '$mount_mytmpfs" RETURN
```

src/core/mount.c
```diff
@@ -1801,6 +1801,7 @@ static int mount_load_proc_self_mountinfo(Manager *m, bool set_flags) {
                 (void) mount_setup_unit(m, device, path, options, fstype, set_flags);
         }
 
+        m->mountinfo_uptodate = true;
```
daandemeyer: Maybe name this `mountinfo_dirty` instead? (and invert all the conditions/assignments)

test/units/testsuite-60.sh
yuwata: Thank you!

src/core/mount.c
yuwata: Not strong opinion, but I prefer `mountinfo_uptodate`, as if `mountinfo_dirty` or something inverted is used, then we need to set the flag on initialization:
```
*m = (Manager) {
        ...,
        .mountinfo_dirty = true,
};
```

test/units/testsuite-60.sh
```diff
@@ -53,6 +53,53 @@ EOF
     }
 }
 
+cleanup_23796() {(
```
mrc0mmand: You can drop this function now.

yuwata: Ouch. Indeed.

test/units/testsuite-60.sh
```diff
@@ -53,6 +53,44 @@ EOF
...
+    mkdir -p /run/systemd/system
+    cat >/run/systemd/system/tmp-hoge.mount <<EOF
+[Mount]
+What=mytmpfs
+Where=/tmp/hoge
+Type=mytmpfs
+EOF
+
+    # shellcheck disable=SC2064
+    trap "rm -f /run/systemd/system/tmp-hoge.mount '$mount_mytmpfs" RETURN
```
mrc0mmand: Whoops, I missed `'` in my previous comment, sorry about that:

```suggestion
    trap "rm -f /run/systemd/system/tmp-hoge.mount '$mount_mytmpfs'" RETURN
```

Also, I had to add a sleep between `systemctl start` and `systemctl daemon-reexec` below, to reproduce the issue reliably after reverting the first commit from this PR, i.e. the final patch to make it all work is

```diff
diff --git a/test/units/testsuite-60.sh b/test/units/testsuite-60.sh
index c282e9ac68..ebb9ae9936 100755
--- a/test/units/testsuite-60.sh
+++ b/test/units/testsuite-60.sh
@@ -74,10 +74,11 @@ Type=mytmpfs
 EOF
 
     # shellcheck disable=SC2064
-    trap "rm -f /run/systemd/system/tmp-hoge.mount '$mount_mytmpfs" RETURN
+    trap "rm -f /run/systemd/system/tmp-hoge.mount '$mount_mytmpfs'" RETURN
 
     for ((i = 0; i < 10; i++)); do
         systemctl --no-block start tmp-hoge.mount
+        sleep ".$RANDOM"
         systemctl daemon-reexec
 
         sleep 1
```

src/core/manager.c
```diff
@@ -1757,6 +1757,10 @@ static void manager_ready(Manager *m) {
...
+        m->mountinfo_uptodate = false;
```
poettering: hmm, I am confused. This is not hooked to SIGCHLD at all? Is the comment outdated?

yuwata:
(The failures in mkosi CIs should not be related.)

Geass-LL:
ping. Now CI is green, could you please review and merge this PR?

src/core/mount.c
```diff
@@ -761,8 +761,31 @@ static int mount_coldplug(Unit *u) {
...
+                        break;
+                case MOUNT_MOUNTING:
+                        new_state = MOUNT_MOUNTING_DONE;
+                        break;
+                default:
+                        break;
+                }
+        else
+                switch (new_state) {
+                case MOUNT_MOUNTING_DONE:
```
msekletar: Maybe we should add an assertion here that we have to have control_pid or at least warn when we don't have it.

src/core/mount.c
```diff
@@ -761,8 +761,31 @@ static int mount_coldplug(Unit *u) {
...
+                        new_state = MOUNT_MOUNTING;
+                        break;
+                case MOUNT_UNMOUNTING:
+                        new_state = MOUNT_DEAD;
```
msekletar: Because of this we will not really coldplug units that were previously unmounting. Do you know how we make sure that control_pid is rewatched and mount is not marked as dead as prematurely? I mean umounting can take some time specifically for NFS and other network FS types.

yuwata:
@jcg190701 Which variant?

jcg190701:
> @jcg190701 Which variant?

variant_0

yuwata:
@jcg190701 Sorry for late response.
During running your test, do you see `(mount-monitor-dispatch) entered rate limit` or `(mount-monitor-dispatch) left rate limit` in journal?

yangmingtai:
@yuwata Some checks were not successful:  focal-s390x — autopkgtest finished (failure)
Please solve it when you are free. Thanks

keszybz:
LGTM.

I restarted s390x/focal now, let's merge if the tests pass.

bluca (approved):

keszybz:
I'm adding this to v251.5.

PR #23640, opened 2022-06-06
Estimate battery level

cerebro1:
Task 3 : https://github.com/systemd/systemd/issues/21165 : Estimate Battery Consumption
@bluca @DaanDeMeyer @anitazha 

Solution Details after fix -
HibernateDelaySec=15min
New VM launch and suspended.
Result :
Auto wakeup after 15 min
10 min bat discharged
Discharge rate 39
Set sleep timer for 1h 48 min 27s
Wakeup again and this time no battery dischargee
System suspended for same duration again
Logs : https://rentry.co/rvsuc

—------------------------------------
Suspend and manual wakeup in 3min
Discharge rate file used to estimate suspend interval
Set timer for 2h 3min 50s
Manual wakeup in 3min & 11% bat discharged, estimate discharge rate
Dis rate is 183
Logs : https://rentry.co/tbqme

—------------------------------------
Exit VM and login again : check file persistency across reboots
Suspend again and wakeup in time almost close to autowakeup i.e : 2min
(Since discharge rate and current bat charge is already known so estimated suspend interval)
Result : System used stored discharge rate to calculate suspend interval 2 min 47s and successfully wake up as manual wakeup in 2 min 19s
Logs : https://rentry.co/ap9z9

PR #23895, opened 2022-07-04
suspend-then-hibernate: Add support for ACPI _BTP so that the bios/hw will trigger a wakeup when battery is low

cerebro1:


src/shared/sleep-config.c
```diff
@@ -102,6 +102,120 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+        if (r < 0)
+               return log_debug_errno(r, "Failed to read discharge rate from %s: %m", filepath);
+
+        r = safe_atoi(discharge_rate, &stored_discharge_rate);
+        if (r < 0)
+               return log_debug_errno(r, "Failed to parse discharge rate read from %s location: %m", filepath);
+
+        if (stored_discharge_rate > 0 && stored_discharge_rate < 200)
+               return stored_discharge_rate;
+        return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Invalid battery discharge percentage rate per hour: %m");
```
poettering: btw, as a very vague rule we prefer if the error code paths are the indented codepaths, and the non-indented codepath is the main success code path.

or in other words, would be cool to invert the if check here, so that the log_debug_errno() call is in the indented block, and the `return stored_discharge_rate` is on the top-level, if you follow what I mean

src/shared/sleep-config.c
```diff
@@ -102,6 +102,120 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+
+        r = read_one_line_file("/sys/class/power_supply/BAT0/energy_now", &battery_energy_now_string);
+        if (r < 0)
+               return log_debug_errno(r, "Failed to read /sys/class/power_supply/BAT0/energy_now: %m");
+
+        r = safe_atoi(battery_energy_now_string, &battery_energy_now);
+        if (r < 0)
+               return log_debug_errno(r, "Failed to parse battery present energy: %m");
+
+        r = read_one_line_file("/sys/class/power_supply/BAT0/alarm", &battery_alarm_string);
```
poettering: i am pretty sure this is optional, hence needs graceful handling on ENOENT here

src/sleep/sleep.c
```diff
@@ -263,41 +263,99 @@ static int execute(
...
+                        return r;
+
+                r = fd_wait_for_event(tfd, POLLIN, 0);
+                if (r < 0)
+                        return log_error_errno(r, "Error polling timerfd: %m");
+
+                if (!FLAGS_SET(r, POLLIN))  { /* We woke up before the alarm time, calculate battery discharged and estimate rate. */
+                        after_timestamp = now(CLOCK_BOOTTIME);
+
+                        time_elapsed = (after_timestamp - before_timestamp) / USEC_PER_SEC;
```
poettering: btw, we generally want to use usecs for all internal time calculations, i.e. please change time_elapsed to `usec_t`, and display such values then with "FORMAT_TIMESPAN()" which will make a nice human readable string from it, on the fly.

src/shared/sleep-config.c
poettering: don't we have to set up the alarm first? i mean, i assume by default laptops won't wake up from suspend when the alarm is configured, we must enable that first, no?

cerebro1:
Sorry. Just saw the review here. I was checking only this #23640 . Thanks for review @poettering . These suggested changes will be part of #23640 so will make changes and push. This PR will be build on the success of #23640 . I should have waited for that to merge before raising this.

src/shared/sleep-config.c
cerebro1: yes. will keep that in mind and revert wherever needed

src/sleep/sleep.c
cerebro1: > btw, we generally want to use usecs for all internal time calculations, i.e. please change time_elapsed to `usec_t`, and display such values then with "FORMAT_TIMESPAN()" which will make a nice human readable string from it, on the fly.

ahh. sorry about that. that is a left-over from my debugging. We do not need that as it is already displayed in logs how much suspend-then-hibernate took. I added that just to validate my calculations are correct. Removing it.
Example logs that already exist -
Jul 06 13:50:07 localhost systemd[1]: systemd-suspend-then-hibernate.service: Consumed 17min 33.653s CPU time.

src/shared/sleep-config.c
cerebro1: fixed.

src/shared/sleep-config.c
```diff
@@ -526,6 +536,88 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+                r = safe_atoi(&offset_char, &offset_int);
+                if (r < 0)
+                        return log_debug_errno(r, "Failed to parse wakeup type offset value: %m");
+
+                *ret = offset_int;
+                return SMBIOS_WAKEUP_BIT_SET;
+        }
+        log_debug("DMI BIOS System Information does not indicate wakeup type bit is set.");
+        return SMBIOS_WAKEUP_BIT_UNSET;
```
bluca: This can be simplified: we are not interest in the exact bit set, only if it was timer or not. So instead of the enum, simply make this function return 1 if it was timer, 0 if it wasn't, and negative error at usual on errors.

src/shared/sleep-config.c
```diff
@@ -526,6 +536,88 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        r = safe_atoi(batterytrippoint, &battery_alarm);
+        if (r < 0)
+                return r;
+
+        *ret = battery_alarm;
+
+        return 0;
+}
+
+bool battery_trip_point_alarm_exists(void) {
```
bluca: Errors can be returned, so should be int here

src/shared/sleep-config.c
```diff
@@ -526,6 +536,88 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                r = read_power_supply_battery_alarm_file(alarm_filepath, &battery_alarm);
+                if (r < 0) {
+                        log_device_debug_errno(dev, r, "Failed to read battery alarm from %s, ignoring: %m", alarm_filepath);
+                        continue;
+                }
+                if (battery_alarm <= 0)
+                        return false;
+        }
+
+        return true;
```
bluca, line 555: This will return true if there are no batteries, or if they are all skipped by any of the continues

src/shared/sleep-config.c
```diff
@@ -41,10 +41,14 @@
 #include "string-util.h"
 #include "strv.h"
 #include "time-util.h"
+#include "unaligned.h"
+
```
yuwata: doubled empty line.

src/shared/sleep-config.c
cerebro1: In systems like Ubuntu 22.04 on Lenovo Thinqpad, which have btp support so alarm file exists but the architecture does not expose smbios dmi entries. so this will need to return 0 if wakeup bit is not set. If any system exposes dmi entries and wakeup bit is set no matter what value, we need to check its set at the time of suspend.
if bit  corresponds to APM, timer then hibernate. So the bit has to be returned to check on suspend as well as on wakeup.

bluca: @poettering do you know what's the story with these dmi entries not being available?

src/shared/sleep-config.c
cerebro1: fixed

src/shared/sleep-config.c
cerebro1: fixed

src/shared/sleep-config.c
cerebro1: fixed

cerebro1:
@yuwata @DaanDeMeyer @anitazha Please review.

src/shared/sleep-config.c
```diff
@@ -526,6 +529,90 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                        log_device_debug_errno(dev, r, "Failed to read battery name, ignoring: %m");
+                        continue;
+                }
+
+                alarm_filepath = path_join(POWER_SUPPLY_DIRPATH, battery_name, "alarm");
+                r = read_power_supply_battery_alarm_file(alarm_filepath, &battery_alarm);
+                if (r < 0) {
+                        log_device_debug_errno(dev, r, "Failed to read battery alarm from %s, ignoring: %m", alarm_filepath);
+                        continue;
+                }
```
yuwata: It seems not necessary to read battery name and build path to sysfs attribute. I guess the following should be simpler:
```
const char *s;
int a;

r = sd_device_get_sysattr_value(dev, "alarm", &s);
if (r < 0) {
        log_device_debug_errno(...);
        continue;
}

r = safe_atoi(s, &a);
if (r < 0)
        ...
```

src/shared/sleep-config.c
```diff
@@ -526,6 +529,90 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0) {
+                log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+                return r;
+        }
+        if (readsize < 25 || s[1] < 25) {
+                log_debug("Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
+                return 0;
```
yuwata: I think we should return negative errno in this case, maybe -EINVAL.
```
return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "...");
```

src/shared/sleep-config.c
```diff
@@ -526,6 +529,90 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        size_t readsize;
+        int offset, r;
+
+        assert(ret);
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0) {
+                log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+                return r;
```
yuwata: Please write like the following:
```
if (r < 0)
        return log_debug_errno(r, "...");
```

src/shared/sleep-config.c
```diff
@@ -526,6 +529,90 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        offset = s[24];
+        /* 0 is Reserved and 8 is AC Power Restored. As per table 12 in
+         * https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf */
+        if (offset > 0 && offset < 9) {
+                log_debug("DMI BIOS System Information indicates wakeup type bit is set in range 1-8");
+                *ret = offset;
+                return 1;
+        }
+
+        return 0;
```
yuwata: Please always set `ret` when the function return non-negative:
```
*ret = 0;
return 0;
```

src/shared/sleep-config.c
```diff
@@ -526,6 +529,90 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        if (r < 0) {
+                log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+                return r;
+        }
+        if (readsize < 25 || s[1] < 25) {
+                log_debug("Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
+                return 0;
+        }
+
+        offset = s[24];
```
yuwata: Not a matter, but the naming `offset` for a value sounds strange for me...

src/sleep/sleep.c
```diff
@@ -264,77 +264,110 @@ static int execute(
 
 static int execute_s2h(const SleepConfig *sleep_config) {
         _cleanup_hashmap_free_ Hashmap *last_capacity = NULL, *current_capacity = NULL;
-        int r;
+        int wakeup_type_bit, r;
```
yuwata: The failure of `check_wakeup_type()` is ignored. You need to initialize `wakup_type_bit` with 0 at the beginning, otherwise, the variable may be used without initialization.

src/sleep/sleep.c
```diff
@@ -264,77 +264,110 @@ static int execute(
...
-                        log_debug_errno(r, "Failed to estimate suspend interval using previous discharge rate, ignoring: %m");
-                else
-                        suspend_interval = total_suspend_interval;
-
-                log_debug("Set timerfd wake alarm for %s", FORMAT_TIMESPAN(suspend_interval, USEC_PER_SEC));
-                /* Wake alarm for system with or without battery to hibernate or estimate discharge rate whichever is applicable */
-                timespec_store(&ts.it_value, suspend_interval);
+        r = check_wakeup_type(&wakeup_type_bit);
+        if (r < 0)
+                log_debug_errno(r, "Failed to fetch wakeup-type bit: %m");
```
yuwata: Please add ", ignoring" in the log message.

src/sleep/sleep.c
```diff
@@ -264,77 +264,110 @@ static int execute(
...
 
-                if (timerfd_settime(tfd, 0, &ts, NULL) < 0)
-                        return log_error_errno(errno, "Error setting battery estimate timer: %m");
+        r = battery_trip_point_alarm_exists();
+        if (r < 0)
+                log_debug_errno(r, "Failed to check acpi_btp support enabled or not: %m");
```
yuwata: Also here.

src/sleep/sleep.c
```diff
@@ -264,77 +264,110 @@ static int execute(
...
 
+        if (r > 0 && wakeup_type_bit > 0) {
+                log_debug("Attempting to suspend...");
                 r = execute(sleep_config, SLEEP_SUSPEND, NULL);
                 if (r < 0)
                         return r;
 
-                r = fd_wait_for_event(tfd, POLLIN, 0);
+                r = check_wakeup_type(&wakeup_type_bit);
```
yuwata: Hm? Why this is called again?

src/sleep/sleep.c
```diff
@@ -264,77 +264,110 @@ static int execute(
...
-                if (r < 0) {
-                        /* In case of no battery or error while getting charge level, no need to measure
-                         * discharge rate. Instead system should wakeup if it is manual wakeup or
-                         * hibernate if this is a timer wakeup.   */
-                        log_debug_errno(r, "Battery capacity percentage unavailable, cannot estimate discharge rate: %m");
-                        if (!woken_by_timer)
-                                return 0;
-                        break;
-                }
+                if (wakeup_type_bit != 3)
```
yuwata: Why not check the value earlier??

src/sleep/sleep.c
```diff
@@ -264,77 +264,110 @@ static int execute(
...
+                        /* For APM Timer wakeup, system should hibernate else wakeup */
+                        return 0;
+        }
+        else {
+                while (battery_is_low() == 0) {
```
yuwata: Is this `else` really necessary?? What happens `wakeup_type_bit == 3` in the above?

If the logic is correct, then please move `else` after `}`, and drop unnecessary `{`:
```
} else
        while (battery_is_low() == 0)
```

yuwata (changes requested):

src/shared/sleep-config.c
cerebro1: Yeah makes sense. Will fix.

src/shared/sleep-config.c
cerebro1: Yes. fixed

src/shared/sleep-config.c
cerebro1: ohh sorry. fixed.

src/shared/sleep-config.c
cerebro1: sure

src/shared/sleep-config.c
cerebro1: https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf
Actually using the term from Table 10 in this doc[Line 957].

src/sleep/sleep.c
cerebro1: Yes. initialized now. The reason to ignore failure is because there are certain system that by architecture do not expose the `/sys/firmware/dmi/entries` filepath. In that case the normal method of estimating the timer and suspending the machine should be used. thus ignored failure

src/sleep/sleep.c
cerebro1: After the indefinite suspend, if the system wakes up, there could 2 possibilities
- manual wakeup
- APM Timer

If it is APM timer wakeup for which this wakeup bit will be set to 3 as per Table 12 [Line 973]
https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf , it means the battery is Low. In that case the system should hibernate.

src/sleep/sleep.c
cerebro1: We need to check this at the time of wakeup after the indefinite suspend

src/sleep/sleep.c
cerebro1: same as above.
ok. will fix else

src/shared/sleep-config.c
yuwata: No, no, 24 is offset (or in other words, index). And `s[24]` is value.

cerebro1: yes agree with you. will change it to `offset_value` or `wakeup_type_bit`

src/shared/sleep-config.c
```diff
@@ -526,6 +528,71 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        if (r < 0)
+                return log_debug_errno(r, "Failed to initialize battery enumerator: %m");
+
+        FOREACH_DEVICE(e, dev) {
+                const char *s;
+                int battery_alarm;
+
+                r = sd_device_get_sysattr_value(dev, "alarm", &s);
+                if (r < 0) {
+                        log_device_debug_errno(dev, r, "Failed to read battery name, ignoring: %m");
```
bluca: these shouldn't be ignored, same below - unless it can be guaranteed that alarms are supported on all detected batteries, the fallback should be used instead

src/shared/sleep-config.c
```diff
@@ -526,6 +528,71 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                        continue;
+                }
+
+                r = safe_atoi(s, &battery_alarm);
+                if (r < 0) {
+                        log_device_debug_errno(dev, r, "Failed to parse battery alarm, ignoring: %m");
+                        continue;
+                }
+
+                battery_count++;
```
bluca: it's not necessary to count, simply return false immediately if any enumerated battery doesn't have the alarm

src/shared/sleep-config.c
```diff
@@ -526,6 +528,71 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+        if (readsize < 25 || s[1] < 25)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
+
+        wakeup_type_bit = s[24];
+        /* 0 is Reserved and 8 is AC Power Restored. As per table 12 in
+         * https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf */
+        if (wakeup_type_bit > 0 && wakeup_type_bit < 9) {
+                log_debug("DMI BIOS System Information indicates wakeup type bit is set in range 1-8");
+                *ret = wakeup_type_bit;
```
bluca: There's no need to expose the byte to the caller. Simply return 1 if it's `3`, 0 if it's not, and negative in case of error or out of the expected range

src/shared/sleep-config.c
```diff
@@ -526,6 +528,71 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0)
+                return log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+
+        if (readsize < 25 || s[1] < 25)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
+
+        wakeup_type_bit = s[24];
```
bluca: this is a byte (rather than a bit), and also it needs to be of type `uint8_t` rather than int. int are 4/8 bytes large, this is a single byte.

src/shared/sleep-config.c
```diff
@@ -41,10 +41,12 @@
 #include "string-util.h"
 #include "strv.h"
 #include "time-util.h"
+#include "unaligned.h"
```
bluca: what is this needed for?

cerebro1: sorry. will remove.

src/shared/sleep-config.c
```diff
@@ -526,6 +527,59 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0)
+                return log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+
+        if (readsize < 25 || s[1] < 25)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
+
+        wakeup_type_byte = s[24];
```
bluca: needs an explicit cast here `(uint8_t)`

src/shared/sleep-config.c
```diff
@@ -526,6 +527,59 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        /* 0 is Reserved and 8 is AC Power Restored. As per table 12 in
+         * https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf */
+        if (wakeup_type_byte == 3) {
```
bluca, line 584: you should check if it's > 8 and return `SYNTHETIC_ERRNO(EINVAL)` before this

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
-                timespec_store(&ts.it_value, suspend_interval);
+        r = check_wakeup_type();
+        if (r < 0)
+                log_debug_errno(r, "Failed to fetch wakeup-type bit, ignoring: %m");
 
-                if (timerfd_settime(tfd, 0, &ts, NULL) < 0)
-                        return log_error_errno(errno, "Error setting battery estimate timer: %m");
+        s = battery_trip_point_alarm_exists();
+        if (s < 0)
+                log_debug_errno(r, "Failed to check acpi_btp support enabled or not, ignoring: %m");
```
bluca: "to check whether acpi_btp support is enabled or not,"

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
                 if (r < 0)
                         return r;
 
-                r = fd_wait_for_event(tfd, POLLIN, 0);
+                r = check_wakeup_type();
                 if (r < 0)
-                        return log_error_errno(r, "Error polling timerfd: %m");
-                /* Store fd_wait status */
-                woken_by_timer = FLAGS_SET(r, POLLIN);
+                        return log_debug_errno(r, "Failed to fetch wakeup-type bit: %m");
```
bluca: "Failed to check hardware wakeup type:"

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
-                        log_debug_errno(r, "Failed to estimate suspend interval using previous discharge rate, ignoring: %m");
-                else
-                        suspend_interval = total_suspend_interval;
-
-                log_debug("Set timerfd wake alarm for %s", FORMAT_TIMESPAN(suspend_interval, USEC_PER_SEC));
-                /* Wake alarm for system with or without battery to hibernate or estimate discharge rate whichever is applicable */
-                timespec_store(&ts.it_value, suspend_interval);
+        r = check_wakeup_type();
+        if (r < 0)
+                log_debug_errno(r, "Failed to fetch wakeup-type bit, ignoring: %m");
```
bluca: "Failed to check hardware wakeup type,"

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
 
 static int execute_s2h(const SleepConfig *sleep_config) {
         _cleanup_hashmap_free_ Hashmap *last_capacity = NULL, *current_capacity = NULL;
-        int r;
+        int r, s;
```
bluca: use `k` instead of `s` for consistency with other files

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
-                if (r < 0) {
-                        /* In case of no battery or error while getting charge level, no need to measure
-                         * discharge rate. Instead system should wakeup if it is manual wakeup or
-                         * hibernate if this is a timer wakeup.   */
-                        log_debug_errno(r, "Battery capacity percentage unavailable, cannot estimate discharge rate: %m");
-                        if (!woken_by_timer)
-                                return 0;
-                        break;
-                }
+                if (r == 1)
```
bluca: this should be `r == 0`

bluca (approved):
Tested on my Lenovo running Debian stable, seems to work as intended

src/shared/sleep-config.c
```diff
@@ -526,6 +527,62 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                        return log_device_debug_errno(dev, r, "Failed to parse battery alarm: %m");
+                if (battery_alarm <= 0)
+                        return false;
+        }
+        return true;
+}
+
+/* Return true if wakeup type is APM timer */
+int check_wakeup_type(void) {
+        _cleanup_free_ char *s = NULL;
```
poettering, line 560: make this `uint8_t` please, it's binary data

src/shared/sleep-config.c
```diff
@@ -526,6 +527,62 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        uint8_t wakeup_type_byte = 0;
+        size_t readsize;
+        int r;
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0)
+                return log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+
+        if (readsize < 25 || s[1] < 25)
```
poettering: if you don't make `s` uint8_t, then on most archs it will end up being signed, and thus the s[1] < 25 check will make things fail once the size reaches 127...

src/shared/sleep-config.c
```diff
@@ -526,6 +527,62 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
+
+        wakeup_type_byte = (uint8_t) s[24];
+        /* 0 is Reserved and 8 is AC Power Restored. As per table 12 in
+         * https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf */
+        if (wakeup_type_byte > 9)
```
poettering: they might add more wake-up types in later releases. Which should be OK. Just ignore all but the APM timers. regardless which range

poettering: also i guess you mean `>= 9` here anyway?

poettering (changes requested):
hmm, so i am not sure if we should rely on the wakeup source alone. It sounds to me we should always do *both*: wakeup by time *and* by battery alarm. 

Also, the alarm should be enabled explicitly (by writing to the sysfs attribute) if it isn't available yet.

I think we can be more aggressive with the time logic (i.e. pick longer wake-up intervals) if we see that all batteries support the alarm stuff, but even then I'd still always set it, for safety reasons if the alarm stuff might not be wired up correctly.

bluca:
> hmm, so i am not sure if we should rely on the wakeup source alone. It sounds to me we should always do _both_: wakeup by time _and_ by battery alarm.
> 
> Also, the alarm should be enabled explicitly (by writing to the sysfs attribute) if it isn't available yet.
> 
> I think we can be more aggressive with the time logic (i.e. pick longer wake-up intervals) if we see that all batteries support the alarm stuff, but even then I'd still always set it, for safety reasons if the alarm stuff might not be wired up correctly.

Having used this I'm not sure it would be worth it? I mean we always say that systemd shouldn't be the dumping ground for broken hardware workarounds - if an hardware timer is exposed as available by the kernel but doesn't work properly, there's already a block list in the kernel driver that can be expanded, and then we'll do the right thing automatically.
The advantage of the hardware timer is that you don't need to rely on manual estimates that are never going to be 100% accurate, and thus burn a lot of power needlessly. Doing this would negate that advantage.

src/shared/sleep-config.c
cerebro1: to read from virtual file, i need a char variable.

src/shared/sleep-config.c
cerebro1: better to store this is some unsigned char variable before comparison to avoid that.

src/shared/sleep-config.c
cerebro1: i am still not sure removing this condition would be safe because we are also checking for whether the bit is set or not. There should be a valid range that we could use to be on safe side. May be 0-127. Considering in future it is going to be a long list of timers. in case it is set then only we are going for suspend. the APM Timer byte comes in role on wakeup.

src/shared/sleep-config.c
```diff
@@ -526,6 +527,63 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                if (battery_alarm <= 0)
+                        return false;
+        }
+        return true;
+}
+
+/* Return true if wakeup type is APM timer */
+int check_wakeup_type(void) {
+        _cleanup_free_ char *s = NULL;
+        uint8_t wakeup_type_byte = 0, tablesize = 0;
```
bluca: no need to initialize these

src/shared/sleep-config.c
```diff
@@ -526,6 +527,63 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        int r;
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0)
+                return log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+
+        tablesize = (uint8_t) s[1];
+        if (readsize < 25 || tablesize < 25)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
```
bluca: the log message is no longer accurate if table_size is < 25, so have two separate checks/logs

src/shared/sleep-config.c
```diff
@@ -526,6 +527,67 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+        wakeup_type_byte = (uint8_t) s[24];
+        /* 0 is Reserved and 8 is AC Power Restored. As per table 12 in
+         * https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf */
+        if (wakeup_type_byte >= 128)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Expected value in range 0-127");
+
+        if (wakeup_type_byte == 3) {
+                log_debug("DMI BIOS System Information indicates wakeup type is APM Timer");
+                return 1;
```
yuwata: `return true;`

src/shared/sleep-config.c
```diff
@@ -526,6 +527,67 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+        }
+
+        return 0;
```
yuwata: `return false;`

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
+                        } else
+                                log_debug("System woke up too early to estimate discharge rate");
+
+                        if (!woken_by_timer)
+                                /* Return as manual wakeup done. This also will return in case battery was charged during suspension */
+                                return 0;
+
+                        r = check_wakeup_type();
+                        if (r < 0)
+                                log_debug_errno(r, "Failed to check hardware wakeup type: %m");
```
yuwata: Please add `", ignoring"`

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
+                        if (r == 1) {
```
yuwata: Not a big matter, but please use `r > 0`.

yuwata (approved):
LGTM. Only several minor comments.

bluca:
> Also, the alarm should be enabled explicitly (by writing to the sysfs attribute) if it isn't available yet.

Missed this: the kernel enables the alarm by default if it's supported. I don't think we should override it - a user might want to disable it (eg: hardware is broken but kernel doesn't blocklist it yet), and they can do so. If we always override it, they can't.

src/shared/sleep-config.c
daandemeyer: The log message seems wrong? We expect to read at least 25 bytes if we're going to access the 24th index no?

src/shared/sleep-config.c
```diff
@@ -526,6 +527,67 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+                if (r < 0)
+                        return log_device_debug_errno(dev, r, "Failed to read battery alarm: %m");
+
+                r = safe_atoi(s, &battery_alarm);
+                if (r < 0)
+                        return log_device_debug_errno(dev, r, "Failed to parse battery alarm: %m");
+                if (battery_alarm <= 0)
+                        return false;
+        }
+        return true;
```
daandemeyer, line 555: Nit: missing newline before the return

src/shared/sleep-config.c
```diff
@@ -526,6 +527,67 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+int check_wakeup_type(void) {
+        _cleanup_free_ char *s = NULL;
+        uint8_t wakeup_type_byte, tablesize;
+        size_t readsize;
+        int r;
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0)
+                return log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
```
daandemeyer: The "ignoring" here doesn't seem correct, since we're most definitely not ignoring the error as we're returning early because of it.

src/sleep/sleep.c
```diff
@@ -264,78 +264,109 @@ static int execute(
...
-                timespec_store(&ts.it_value, suspend_interval);
+        r = check_wakeup_type();
+        if (r < 0)
+                log_debug_errno(r, "Failed to check hardware wakeup type, ignoring: %m");
 
-                if (timerfd_settime(tfd, 0, &ts, NULL) < 0)
-                        return log_error_errno(errno, "Error setting battery estimate timer: %m");
+        k = battery_trip_point_alarm_exists();
+        if (k < 0)
+                log_debug_errno(r, "Failed to check whether acpi_btp support is enabled or not, ignoring: %m");
```
daandemeyer: Should use `k` here instead of `r`

DaanDeMeyer (approved):
Found a few nits, please address before merging

src/shared/sleep-config.c
cerebro1: fixed

src/shared/sleep-config.c
cerebro1: fixed

src/shared/sleep-config.c
cerebro1: yes. fixed

src/sleep/sleep.c
cerebro1: fixed

cerebro1:
Please review this. I have moved the while loop as a seperate helper now.

src/sleep/sleep.c
```diff
@@ -262,12 +262,10 @@ static int execute(
         return r;
 }
 
-static int execute_s2h(const SleepConfig *sleep_config) {
+static int custom_timer_suspend(const SleepConfig *sleep_config) {
         _cleanup_hashmap_free_ Hashmap *last_capacity = NULL, *current_capacity = NULL;
         int r;
```
bluca, line 268: `assert(sleep_config);`

src/sleep/sleep.c
```diff
@@ -335,7 +333,55 @@ static int execute_s2h(const SleepConfig *sleep_config) {
...
+                if (r < 0)
+                        return log_debug_errno(r, "Failed to check hardware wakeup type: %m");
+
+                if (r == 0)
+                        /* For APM Timer wakeup, system should hibernate else wakeup */
+                        return 0;
+        } else {
+                r = custom_timer_suspend(sleep_config);
+                if(r < 0)
+                        log_debug_errno(r, "Failed to set custom timer suspend");
```
bluca: it was returning on error from the loop, so this should return too

src/sleep/sleep.c
```diff
@@ -335,7 +335,55 @@ static int execute_s2h(const SleepConfig *sleep_config) {
...
+                if (r < 0)
+                        return log_debug_errno(r, "Failed to check hardware wakeup type: %m");
+
+                if (r == 0)
+                        /* For APM Timer wakeup, system should hibernate else wakeup */
+                        return 0;
+        } else {
+                r = custom_timer_suspend(sleep_config);
+                if(r < 0)
+                        return log_debug_errno(r, "Failed to set custom timer suspend");
```
bluca: "Suspend cycle with manual battery discharge rate estimation failed: %m"

PR #23909, opened 2022-07-05
journal: Make sd_journal_previous/next() return 0 at HEAD/TAIL

daandemeyer:
Currently, both these functions don't return 0 if we're at HEAD/TAIL
and move in the corresponding direction. Let's fix that.

Replaces #23480

### DIFF
--- a/src/journal/test-journal-interleaving.c
+++ b/src/journal/test-journal-interleaving.c
@@ -158,6 +158,7 @@ static void test_skip_one(void (*setup)(void)) {
          */
         assert_ret(sd_journal_open_directory(&j, t, 0));
         assert_ret(sd_journal_seek_head(j));
+        assert_ret(sd_journal_previous(j) == 0);
         assert_ret(sd_journal_next(j));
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
This logic needs some more general cleaning up but I'll do that later. For now, let's make sure we fix the issue.

pobrn:
It seems this has caused two sd journal related tests in the pipewire test suite to fail. Could someone please look over them and advise how to fix the code? The pipewire tests made the assumption that `sd_journal_seek_tail()` goes to the *current* tail, and then we can read the new entries using `sd_journal_next()` but looks like this has been changed.

Test code: https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/ea4e5cf364d5925b992c5cb16abd2ff8872f8762/test/test-logger.c#L488
Related pipewire issue: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/2823

yuwata:
@pobrn Could you open a new issue page? Otherwise, we easily forget your report. Thank you.

Issue #6450, opened 2017-07-25
wrong error message for deleted service units ("Access denied")


sourcejedi:
### Submission type

 - Bug report

### systemd version the issue has been seen with

systemd-233-6.fc26.x86_64

### Used distribution

Fedora 26

### In case of bug report: Expected behaviour you didn't see

Error messages should be technically correct (even if they're too broad to actually identify a problem).

### In case of bug report: Unexpected behaviour you saw

"Access denied" message when the fault is a service unit that does not exist.

### In case of bug report: Steps to reproduce the problem

```
# systemctl edit --full systemd-suspend.service  # and exit the editor without making changes
# rm /etc/systemd/system/systemd-suspend.service # delete the override file
# systemctl start systemd-suspend.service
Failed to start systemd-suspend.service: Access denied
See system logs and 'systemctl status systemd-suspend.service' for details.
# systemctl status systemd-suspend.service
Failed to get properties: Access denied
```

This made me think systemd was denying access to all requests.  But actually it was failing with the wrong error only for requests which read `systemd-suspend.service`.

sourcejedi:
Technically it's not a log message.  The journal is never consulted; `systemctl status` aborts early.  I can see an Access Denied error that would be generated in `mac_selinux_generic_access_check()`.  Although it sets a bus reply message "Failed to get file context on %s.", which I'm not seeing in output.

Because it's deliberately suppressed, sigh.

```
const char *bus_error_message(const sd_bus_error *e, int error) {

        if (e) {
                /* Sometimes, the D-Bus server is a little bit too verbose with
                 * its error messages, so let's override them here */
                if (sd_bus_error_has_name(e, SD_BUS_ERROR_ACCESS_DENIED))
                        return "Access denied";
```

I don't know how to resolve that.

poettering:
Yeah, it's messy, the errors dbus spits out are just undigestale for users, they contain whole match strings and XML bits and we really don#t want to show all tha tin user facing UI...

What I don't get, how come mac_selinux_generic_access_check() returns an access error if the unit file doesn't exist? That appears to be the thing to fix here...

keszybz:
I think this was fixed by 173471b771a52baa7f705019600956e875b0cb9a. When I follow your recipe to reproduce, `start` and `status` work without issues.

sourcejedi:
I get the same result, even after `fixfiles restore && setenforce 1`.  I'll make sure to re-report if I notice it again :).

[...]

PR #10023, opened 2018-09-06
re-add SELinux checks for unit handling

cgzones:
- add mac_check callback to unit_file_* functions to check unit actions
      Use the filecontext of /lib/systemd/system/{UNIT_NAME} as reference label for checks if available,
      else the systemd runtime context
      (this helps when operating on edited files in /etc/systemd/system or links to /dev/null)

    - add mac checks for { bus_job_method_get_waiting_jobs / method_unref_unit /
      method_list_units_by_names / method_lookup_dynamic_user_by_name /
      method_lookup_dynamic_user_by_uid / method_get_dynamic_users }

    - rename { kexec / switch_root / environment } checks to "admin"

This is a first attempt to solve #1050.
There is still some debug code present.

Some questions:
  - What exactly does `method_unref_unit` do? What access verb would fit?
  - How are the invocations in ` src/systemctl/systemctl.c` triggered?
    They are currently ignored via `mac_nocheck`.

PR #23910, opened 2022-07-05
selinux: return ENOENT when a unit file is not found rather than EPERM

bluca:
When checking the label of a unit, and the file is not found, access
denied is returned to the caller, which is misleading.

Fixes https://github.com/systemd/systemd/issues/6450

### DIFF
--- a/src/core/selinux-access.c
+++ b/src/core/selinux-access.c
@@ -239,7 +239,10 @@ int mac_selinux_access_check_internal(
                         if (!enforce)
                                 return 0;
 
-                        return sd_bus_error_setf(error, SD_BUS_ERROR_ACCESS_DENIED, "Failed to get file context on %s.", path);
+                        return sd_bus_error_setf(error,
+                                                 IN_SET(r, -ENOENT, -ENODATA) ? SD_BUS_ERROR_FILE_NOT_FOUND : SD_BUS_ERROR_ACCESS_DENIED,
+                                                 "Failed to get file context on %s.",
+                                                 path);
                 }
 
                 tclass = "service";


pebenito:
This looks good to me. It continues to return an error, but it is more specific.

src/core/selinux-access.c
```diff
@@ -239,7 +239,10 @@ int mac_selinux_access_check_internal(
                         if (!enforce)
                                 return 0;
 
-                        return sd_bus_error_setf(error, SD_BUS_ERROR_ACCESS_DENIED, "Failed to get file context on %s.", path);
+                        return sd_bus_error_setf(error,
+                                                 r == -ENOENT ? SD_BUS_ERROR_FILE_NOT_FOUND : SD_BUS_ERROR_ACCESS_DENIED,
+                                                 "Failed to get file context on %s.",
+                                                 path);
```
poettering, line 245: this is ugly: there can be so many reasons this fails, ENOENT is just one among many, and specially handling that one, over all the other ones is a mess.

let's maybe fix this properly: cache the label in the `Unit` object whenever we load the file, and then load it from the open fd, instead of doing a separate path-based lookup here.

Doing second level security checks by paths is racy as fuck, and revolting to be frank.

If you do it that way, the issue goes away entirely, and we don't have to special case error codes anymore.

poettering: this was proposed in some other issue already. and would make things a ton nicer: we could then report the label as bus property and so on, and the label would follow our usual load cycle.

cgzones: That method was objected at https://github.com/systemd/systemd/pull/10023#issuecomment-650025249

bluca: Other than that comment, having to do a daemon-reload on policy update is a problem for me, as that's a major stall on the system that I really need to avoid.

bluca: > this is ugly: there can be so many reasons this fails, ENOENT is just one among many, and specially handling that one, over all the other ones is a mess.

But which other SD_BUS_ERROR value would match an errno? I struggle to see any that could be returned by that call?

keszybz:
LGTM. We start returning a specific error for one specific errno, and otherwise things are as before. We can certainly merge this as a quick fix, and discuss caching in a separate PR.

src/core/selinux-access.c
poettering: > That method was objected at https://github.com/systemd/systemd/pull/10023#issuecomment-650025249

Well, I disagree with the reasoning. If selinux wants to be taken seriously as a security tech, then the project should learn that applying/reading labels always via file system paths is a racy mess, and undestand we don#t really do things anymore, but rather want to pin inodes while we operate on them, and then get a reasonably congruent view of the world.

poettering: > But which other SD_BUS_ERROR value would match an errno? I struggle to see any that could be returned by that call?

EOPTNOTSUPP (if selinux is not supported on the fs, or so, or in the execution environment/container)

ELOOP (if the unit file became part of a symlink loop)

ENOTDIR... ESTALE...

and so on and so on.

bluca: can that libselinux api actually return any of those?

poettering: it's just a wrapper around getxattr(), no? and I doubt they eat up those errors... (but then again, my hopes might be too high)

bluca: It does get bubbled up - according to the manpage, getxattr can return E2BIG, ERANGE, ENODATA, ENOTSUP. We shouldn't worry about the first one here, ERANGE is handled by the library with a retry, and ENOTSUP seems fine to map it to ACCESS_DENIED. I'll add ENODATA to the NOT_FOUND mapping.

bluca: > > That method was objected at [#10023 (comment)](https://github.com/systemd/systemd/pull/10023#issuecomment-650025249)
> 
> Well, I disagree with the reasoning. If selinux wants to be taken seriously as a security tech, then the project should learn that applying/reading labels always via file system paths is a racy mess, and undestand we don#t really do things anymore, but rather want to pin inodes while we operate on them, and then get a reasonably congruent view of the world.

I am not going to solve that here. If all of that happens at some point in the future, all the better - but right now there's a very real problem (can't tell between a permission denial and a file that's gone), and I need a fix for it, and it can't really wait for a hypothetical major redesign that might arrive down the line

cgzones: I am not sure about `ENODATA`.
According to getxattr(2) there are two causes for it:
  - The named attribute does not exist,
  -  or the process has no access to this attribute.

On a SELinux enabled system all files *should* have a security context (named attribute of `"security.selinux"`) either by creation or initial file system setup and it can't be removed. So a unit file not having a SELinux context is very unlikely and serious.

Also for the second reason `ENODATA` should not be mapped to `SD_BUS_ERROR_FILE_NOT_FOUND`.

bluca: > or the process has no access to this attribute.

I'm not sure this could realistically happen to pid1 given the existing policy? Which means it can only be the first one - which might be worrying, but it's entirely correct to map as NOT_FOUND. It doesn't mean it gets ignored, it just means there's a meaningful error reported back, so that one can do the appropriate action.

poettering: > * 
> 
> 
> On a SELinux enabled system all files _should_ have a security context (named attribute of `"security.selinux"`) either by creation or initial file system setup and it can't be removed. So a unit file not having a SELinux context is very unlikely and serious.

What happens if i mount a foreign (and thus unlabelled) ext4 fs on an selinux-enabled host? do its inodes magically acquire labels?

bluca: > > On a SELinux enabled system all files _should_ have a security context (named attribute of `"security.selinux"`) either by creation or initial file system setup and it can't be removed. So a unit file not having a SELinux context is very unlikely and serious.
> 
> What happens if i mount a foreign (and thus unlabelled) ext4 fs on an selinux-enabled host? do its inodes magically acquire labels?

AFAIK no, it would all stay unlabeled, it would require a restorecon on the volume to apply labels according to the current policy

cgzones: > What happens if i mount a foreign (and thus unlabelled) ext4 fs on an selinux-enabled host? do its inodes magically acquire labels?

Then the file will indeed have no label. (The kernel falls back for access checks to the file initial security identifier - commonly `unlabeled_t`.)
But is `SD_BUS_ERROR_FILE_NOT_FOUND` a more meaningful error than `SD_BUS_ERROR_ACCESS_DENIED` in this case, as the actual file exists?

(I am not strongly against this change, just asking.)

bluca: > But is SD_BUS_ERROR_FILE_NOT_FOUND a more meaningful error than SD_BUS_ERROR_ACCESS_DENIED in this case, as the actual file exists?

Well, we are limited in what we can map by the dbus errors we have, and that seems better to me than the status quo.

poettering:
So taking back a step I wonder if we really should make this change at all. The thing is that there are two distinct concepts here: 

1. we should communicate that the call was refused due to an access check issue
2. we should communicate details about the access check issue

So far, the first fact was communicated via the D-Bus error name, the second fact via the english language error message. Which is not perfect (i.e. varlink allow structured errors with arbitrary payload, so that both can be communicated in a computer-readable way), but I think is pretty OK: it's clear that you have an access control issue, but there's also helpful info about why precisely.

Now, if you change the ENOENT case to the FileNotFound error, then things get really confusing, because it's not clear anymore that this is merely about access control: it looks as if some file needed to the method call itself is missing. but that's not really the case, because we don't even get that far here, the body of the method call is never even reached, we already refused because the selinux stuff is not in order, and that's an access control issue first and foremost.

Hence, I am not convinced this is desirable at all?

If this is about debugging selinux issues easier: the dbus error message string specifically exists to make debugging easier, and it is already quite explicit here. Why isn't that sufficient already?

or in other words: shouldn't this be addressed by showing the return message text more prominently? 

(and optionally: by making the string we generate more expressive)

poettering:
Or to say this differently: maybe we should revisit bus_error_message()'s magic overriding of the DBusError's `message` field. 

i.e. this bit:
```c
                /* Sometimes, the D-Bus server is a little bit too verbose with
                 * its error messages, so let's override them here */
                if (sd_bus_error_has_name(e, SD_BUS_ERROR_ACCESS_DENIED))
                        return "Access denied";
```

bluca:
> Now, if you change the ENOENT case to the FileNotFound error, then things get really confusing, because it's not clear anymore that this is merely about access control: it looks as if some file needed to the method call itself is missing. but that's not really the case, because we don't even get that far here, the body of the method call is never even reached, we already refused because the selinux stuff is not in order, and that's an access control issue first and foremost.

But it's not an access control issue. It's a "this unit doesn't exist" issue, and the first thing that happens to check it is selinux. If selinux is disabled, you'd get "unit not found" or somesuch. Because selinux is enabled, instead you get "access denied" - it's the status quo that seems confusing to me? Depending on build time configuration, in the same situation, you get a completely different error.

This is not about debugging, this is about an orchestrator program reacting properly to this error. "Access denied" looks like a security issue and gets flagged up the chain, as it should. "File not found" can be expected, and in this case it is, so it needs to be ignored. The caller cannot know right now.

poettering:
think about it: you start a unit. then delete its unit file. Then try to stop it. We now can't read the label off the unit file anymore, hence any attempt to stop the unit generated from it will fail (which is a really terrible situation actually, and one of the reasons we should stop going to disk via path all the time), but with this PR they'll see FileNotFound, and thus will assume the whole *unit* doesn't exist, and assume it's not running then — except it *does* exist as runtime concept, and it could be stopped in theory (in in reality too, indirectly via a dependency for example, i.e. `systemd-run -p Conflicts=xyz.service /bin/true`).

Hence I think we should not generate FileNotFound here, but an AccessDenied error, but then provide a useful error message that is also displayed properly by the receiving side. Because the issue here is an access issue, much less than a non-existance issue.

poettering:
or to say this differently: for the unit engine the fact that the unit file is gone is not that big of a deal. It's none whatsoever while we havent reloaded, and once we reloaded it's not that terrible either, because for running services we then use a skeleton unit that is just good enough to keep track of things and stop things. 

For the selinux access part otoh the unit file missing just means everything is broken, it cannot work at all anymore: all attempts to do access checks must be denied because there is no way to run the checks properly at all whatsoever.

hence I think we should return AccessDenied, because that makes clear that it's about the access layer (with more info in the message string), and not FileNotFound, since that suggests the unit engine couldn't work, but in fact it's totally OK with this state, and would be happy to process your stop request if the access layer hadn't refused it already.

bluca:
None of that helps with the actual problems that are reported - that is, how can a program react? It needs to be able to tell that in this case it's a no-op and should just ignore it. In a real case of a selinux denial, it needs to escalate. How can it tell the difference?

poettering:
Well, i guess this boils down to: we have an error condition we could return as error code x, or error code y, and both are ambiguous. Which kinda suggest introducing error code z for them.

However, by fixing this properly we wouldn't even have this error condition at all. So I think this is really the way to go: let's read the label at the moment we parse the unit file, and use that. i think the patch for that shouldn#t be too long. 10 lines? i mean, sure this patch has even fewer lines, but caching the label at moment of unit parsing is not that hard...

poettering:
I guess we can close this now that #23986 is happening.

PR #23914, opened 2022-07-05
test-netlink: Make sure netinet/in.h is not include sorted

daandemeyer:


bluca (approved):

src/libsystemd/sd-netlink/test-netlink.c
```diff
@@ -1,13 +1,16 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
-#include <net/if.h>
-#include <netinet/ether.h>
+/* We make sure netinet/in.h is not include sorted automatically by clang-format because it will happily
+ * redefine in_addr even if it has already been defined by the linux headers. */
 #include <netinet/in.h>
```
keszybz, line 5: How does this work? If there's an empty line, clang-format doesn't reorder across it?

daandemeyer: Exactly

yuwata:
We already do like that in many places, especially, src/network. If you touch it, please do more consistently.

DaanDeMeyer:
I've opened https://sourceware.org/bugzilla/show_bug.cgi?id=29354 to hopefully get the bug fixed in glibc instead.

PR #23918, opened 2022-07-06
Revert "dissect: ID from os-release should be non-empty, not just non-NULL"

yuwata:
This reverts commit a2cf73f0b602a93a32107cfc066a5e307263c577.

This is not necessary after 78ab2b5064a0f87579ce5430f9cb83bba0db069a.

Addresses https://github.com/systemd/systemd/pull/23454#discussion_r913611798.

### DIFF
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -3220,6 +3220,8 @@ int verity_dissect_and_mount(
         if (required_host_os_release_id) {
                 _cleanup_strv_free_ char **extension_release = NULL;
 
+                assert(!isempty(required_host_os_release_id));
+
                 r = load_extension_release_pairs(dest, dissected_image->image_name, &extension_release);
                 if (r < 0)
                         return log_debug_errno(r, "Failed to parse image %s extension-release metadata: %m", dissected_image->image_name);


bluca:
This check doesn't hurt and it's minimal, why should it be removed? If it's always true, then shouldn't it become an assertion?

poettering:
It hides errors. 

It would make sense for the whole function to fail with EINVAL if a non-NULL parameter is specified that is an empty string. Or an assert() would be OKish too.

But hiding the error entirely like this is done right now is wrong.

yuwata:
Added one more commit that add an assertion. PTAL.

bluca (approved):

poettering:
lgtm. (that said, did anyone check if we correctly filter such invalid ids earlier on? we should not allow bad os-release data to trigger this issue..

yuwata:
At leaset, `verity_dissect_and_mount()` is called at two places, and only core/namespace.c calls it with non-NULL release ID.

bluca:
> lgtm. (that said, did anyone check if we correctly filter such invalid ids earlier on? we should not allow bad os-release data to trigger this issue..

yeah, see https://github.com/systemd/systemd/commit/78ab2b5064a0f87579ce5430f9cb83bba0db069a

PR #23921, opened 2022-07-06
smack: Add DefaultSmackProcessLabel to user.conf and system.conf

steelman:
DefaultSmackProcessLabel tells systemd what label to assign to its child
process in case SmackProcessLabel is not set in the service file. By
default, when DefaultSmackProcessLabel is not set child processes inherit
label from systemd.

### DIFF
--- a/man/systemd-system.conf.xml
+++ b/man/systemd-system.conf.xml
@@ -525,6 +525,18 @@
         details. Note that this setting has no effect on the OOM score adjustment value of the service
         manager process itself, it retains the original value set during its invocation.</para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>DefaultSmackProcessLabel=</varname></term>
+
+        <listitem><para>Takes a <option>SMACK64</option> security label as the argument. The process executed
+        by a unit will be started under this label if <varname>SmackProcessLabel=</varname> is not set in the
+        unit. See <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+        for the details.</para>
+
+        <para>If the value is <literal>/</literal>, only labels specified with <varname>SmackProcessLabel=</varname>
+        are assigned and the compile-time default is ignored.</para></listitem>
+      </varlistentry>
     </variablelist>
   </refsect1>
 
--- a/src/core/execute.c
+++ b/src/core/execute.c
@@ -3240,6 +3240,7 @@ static int setup_credentials(
 
 #if ENABLE_SMACK
 static int setup_smack(
+                const Manager *manager,
                 const ExecContext *context,
                 int executable_fd) {
         int r;
@@ -3251,20 +3252,17 @@ static int setup_smack(
                 r = mac_smack_apply_pid(0, context->smack_process_label);
                 if (r < 0)
                         return r;
-        }
-#ifdef SMACK_DEFAULT_PROCESS_LABEL
-        else {
+        } else if (manager->default_smack_process_label) {
                 _cleanup_free_ char *exec_label = NULL;
 
                 r = mac_smack_read_fd(executable_fd, SMACK_ATTR_EXEC, &exec_label);
                 if (r < 0 && !IN_SET(r, -ENODATA, -EOPNOTSUPP))
                         return r;
 
-                r = mac_smack_apply_pid(0, exec_label ? : SMACK_DEFAULT_PROCESS_LABEL);
+                r = mac_smack_apply_pid(0, exec_label ? : manager->default_smack_process_label);
                 if (r < 0)
                         return r;
         }
-#endif
 
         return 0;
 }
@@ -4853,7 +4851,7 @@ static int exec_child(
                 /* LSM Smack needs the capability CAP_MAC_ADMIN to change the current execution security context of the
                  * process. This is the latest place before dropping capabilities. Other MAC context are set later. */
                 if (use_smack) {
-                        r = setup_smack(context, executable_fd);
+                        r = setup_smack(unit->manager, context, executable_fd);
                         if (r < 0 && !context->smack_process_label_ignore) {
                                 *exit_status = EXIT_SMACK_PROCESS_LABEL;
                                 return log_unit_error_errno(unit, r, "Failed to set SMACK process label: %m");
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -170,6 +170,7 @@ static void *arg_random_seed;
 static size_t arg_random_seed_size;
 static int arg_default_oom_score_adjust;
 static bool arg_default_oom_score_adjust_set;
+static char *arg_default_smack_process_label;
 
 /* A copy of the original environment block */
 static char **saved_env = NULL;
@@ -658,6 +659,11 @@ static int parse_config_file(void) {
                 { "Manager", "CtrlAltDelBurstAction",        config_parse_emergency_action,      0,                        &arg_cad_burst_action             },
                 { "Manager", "DefaultOOMPolicy",             config_parse_oom_policy,            0,                        &arg_default_oom_policy           },
                 { "Manager", "DefaultOOMScoreAdjust",        config_parse_oom_score_adjust,      0,                        NULL                              },
+#if ENABLE_SMACK
+                { "Manager", "DefaultSmackProcessLabel",     config_parse_string,                0,                        &arg_default_smack_process_label  },
+#else
+                { "Manager", "DefaultSmackProcessLabel",     config_parse_warn_compat,           DISABLED_CONFIGURATION,   NULL                              },
+#endif
                 {}
         };
 
@@ -730,6 +736,8 @@ static void set_manager_defaults(Manager *m) {
         m->default_oom_policy = arg_default_oom_policy;
         m->default_oom_score_adjust_set = arg_default_oom_score_adjust_set;
         m->default_oom_score_adjust = arg_default_oom_score_adjust;
+
+        (void) manager_set_default_smack_process_label(m, arg_default_smack_process_label);
 
         (void) manager_set_default_rlimits(m, arg_default_rlimit);
 
@@ -2404,6 +2412,7 @@ static void reset_arguments(void) {
         arg_clock_usec = 0;
 
         arg_default_oom_score_adjust_set = false;
+        arg_default_smack_process_label = mfree(arg_default_smack_process_label);
 }
 
 static void determine_default_oom_score_adjust(void) {
--- a/src/core/manager.c
+++ b/src/core/manager.c
@@ -1548,6 +1548,8 @@ Manager* manager_free(Manager *m) {
         free(m->switch_root);
         free(m->switch_root_init);
 
+        free(m->default_smack_process_label);
+
         rlimit_free_all(m->rlimit);
 
         assert(hashmap_isempty(m->units_requiring_mounts_for));
@@ -3877,6 +3879,19 @@ int manager_get_effective_environment(Manager *m, char ***ret) {
 
         *ret = l;
         return 0;
+}
+
+int manager_set_default_smack_process_label(Manager *m, const char *label) {
+        assert(m);
+
+#ifdef SMACK_DEFAULT_PROCESS_LABEL
+        if (!label)
+                return free_and_strdup(&m->default_smack_process_label, SMACK_DEFAULT_PROCESS_LABEL);
+#endif
+        if (streq_ptr(label, "/"))
+                return free_and_strdup(&m->default_smack_process_label, NULL);
+
+        return free_and_strdup(&m->default_smack_process_label, label);
 }
 
 int manager_set_default_rlimits(Manager *m, struct rlimit **default_rlimit) {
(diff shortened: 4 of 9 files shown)


src/core/main.c
```diff
@@ -658,6 +659,11 @@ static int parse_config_file(void) {
                 { "Manager", "CtrlAltDelBurstAction",        config_parse_emergency_action,      0,                        &arg_cad_burst_action             },
                 { "Manager", "DefaultOOMPolicy",             config_parse_oom_policy,            0,                        &arg_default_oom_policy           },
                 { "Manager", "DefaultOOMScoreAdjust",        config_parse_oom_score_adjust,      0,                        NULL                              },
+#if ENABLE_SMACK
+                { "Manager", "DefaultSmackProcessLabel",     config_parse_string,                0, &arg_default_smack_process_label       },
```
poettering: please align rightmost column with lines above

src/core/main.c
```diff
@@ -2404,6 +2411,13 @@ static void reset_arguments(void) {
         arg_clock_usec = 0;
 
         arg_default_oom_score_adjust_set = false;
+#ifdef SMACK_DEFAULT_PROCESS_LABEL
+        if (free_and_strdup(arg_default_smack_process_label,
+                            SMACK_DEFAULT_PROCESS_LABEL) < 0)
+                log_oom();
```
poettering: reset_arguments() is not supposed to fail...

I think it would be better if arg_default_smack_process_label would always default to NULL, but then, when you copy it into Manager you fill in SMACK_DEFAULT_PROCESS_LABEL if still NULL.

that way you remove one pointless string copy, and don't have the reset logic here fail, and things would generally be simpler

man/systemd-system.conf.xml
```diff
@@ -525,6 +525,15 @@
...
         manager process itself, it retains the original value set during its invocation.</para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>DefaultSmackProcessLabel=</varname></term>
+
+        <listitem><para>Takes a <option>SMACK64</option> security label as the argument. The process executed
+        by a unit will be started under this label if <varname>SmackProcessLabel=</varname> is not set in the
+        unit. See <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+        for the details.
```
poettering: should there be a way to disable the setting of the label explicitly in case SMACK_DEFAULT_PROCESS_LABEL was set during compile time?

If so, it would probably make sense to define a special parameter here you can set to disable the setting of the label altogether, i.e some string that is nice to look at but not itself a valid SMACK label.

src/core/execute.c
```diff
@@ -3249,20 +3250,17 @@ static int setup_smack(
                 r = mac_smack_apply_pid(0, context->smack_process_label);
                 if (r < 0)
                         return r;
-        }
-#ifdef SMACK_DEFAULT_PROCESS_LABEL
-        else {
+        } else if (!isempty(manager->default_smack_process_label)) {
```
poettering: why use isempty() here? why not a simple NULL check?

man/systemd-system.conf.xml
steelman: Should it prevent assigning labels defined with SmackProcessLabel in unit files too? (IMHO no, just asking to be sure.)

src/core/execute.c
steelman: For no particular reason. I didn't check how config_string_parse handle empty strings. I will change it to NULL check.

man/systemd-system.conf.xml
poettering: nah

src/core/main.c
```diff
@@ -731,6 +737,16 @@ static void set_manager_defaults(Manager *m) {
...
 
+#ifdef SMACK_DEFAULT_PROCESS_LABEL
+        if (!arg_default_smack_process_label)
+                m->default_smack_process_label = SMACK_DEFAULT_PROCESS_LABEL;
+        else
+#endif
+        if (streq(arg_default_smack_process_label, "/"))
+                m->default_smack_process_label = NULL;
+        else
+                m->default_smack_process_label = arg_default_smack_process_label;
```
poettering: nah, please please keep this string copy in place. the Manager object should own the memory of the label string. it should really be self contained when it comes to memory.

Hence, here please copy the thing once from `arg_default_smack_process_label`/`SMACK_DEFAULT_PROCES_LABEL` into `m->default_smack_process_label` with `strdup()`.

man/systemd-system.conf.xml
```diff
@@ -525,6 +525,18 @@
...
+      <varlistentry>
+        <term><varname>DefaultSmackProcessLabel=</varname></term>
+
+        <listitem><para>Takes a <option>SMACK64</option> security label as the argument. The process executed
+        by a unit will be started under this label if <varname>SmackProcessLabel=</varname> is not set in the
+        unit. See <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+        for the details.</para>
+
+        <para>If the value is <literal>/</literal> only labels labels specified with <varname>SmackProcessLabel=</para>
+        are assigned and the compile-time default is ignored.</listitem>
```
poettering: is this common in SMACK land, to use `/` as marker for "unset"? or did you invent this here?

src/core/main.c
```diff
@@ -731,6 +737,16 @@ static void set_manager_defaults(Manager *m) {
...
+
```
poettering, line 741: hmm, I think it would be better to move this into a function `manager_set_default_smack_process_label()`, and hide the default fallback stuff there.

man/systemd-system.conf.xml
steelman: This is my invention. According to SMACK [documentation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/LSM/Smack.rst?id=e8a4e1c1bb697b1d9fc48f0e56dc0f50bc024bee#n438) 

> […] Single character labels using special characters, that being anything other than a letter or digit, are reserved for use by the Smack development team.

but 

> […] Smack labels cannot contain unprintable characters, the "/" (slash), the "\" (backslash), the "'" (quote) and '"' (double-quote) characters. Smack labels cannot begin with a '-'. This is reserved for special options. […]

Which means we can be sure `/` won't be used anywhere in SMACK.

> Every task on a Smack system is assigned a label. The Smack label
of a process will usually be assigned by the system initialization
mechanism.

Therefore, there is no notion of _unset_ in SMACK. This is going to by purely systemd logic, to use kernel default instead of the compiled in systemd default.

src/core/manager.c
```diff
@@ -3879,6 +3881,31 @@ int manager_get_effective_environment(Manager *m, char ***ret) {
...
+                p = strdup(SMACK_DEFAULT_PROCESS_LABEL);
+        else
+#endif
+        if (!label || streq(label, "/")) {
+                m->default_smack_process_label = mfree(m->default_smack_process_label);
+                return 0;
+        } else
+                p = strdup(label);
+
+
```
poettering, line 3893: spurious double newline

src/core/manager.c
```diff
@@ -3879,6 +3881,31 @@ int manager_get_effective_environment(Manager *m, char ***ret) {
...
+        _cleanup_free_ char *p = NULL;
+
+        assert(m);
+
+        if (streq_ptr(m->default_smack_process_label, label))
+                return 0;
+
+#ifdef SMACK_DEFAULT_PROCESS_LABEL
+        if (!label)
+                p = strdup(SMACK_DEFAULT_PROCESS_LABEL);
```
poettering:
```c
return free_and_strdup(&m->default_smack_process_lanel, SMACK_DEFAULT_PROCESS_LABEL);
```

src/core/manager.c
```diff
@@ -3879,6 +3881,31 @@ int manager_get_effective_environment(Manager *m, char ***ret) {
...
+        else
+#endif
+        if (!label || streq(label, "/")) {
+                m->default_smack_process_label = mfree(m->default_smack_process_label);
+                return 0;
+        } else
+                p = strdup(label);
```
poettering:
```c
return free_and_strdup(&m->default_smack_process_label, label);
```

man/systemd-system.conf.xml
poettering: but "-" sounds like a safer option then, no?

steelman: The documentation says.

> Smack labels cannot begin with a '-'. This is reserved for special options.

I'd rather not use something that is marked as reserved.

src/core/manager.c
```diff
@@ -3879,6 +3881,20 @@ int manager_get_effective_environment(Manager *m, char ***ret) {
...
 
+int manager_set_default_smack_process_label(Manager *m, const char *label) {
+        assert(m);
+
+#ifdef SMACK_DEFAULT_PROCESS_LABEL
+        if (!label)
+                return free_and_strdup(&m->default_smack_process_label, SMACK_DEFAULT_PROCESS_LABEL);
+        else
+#endif
+        if (streq(label, "/"))
```
poettering: this will segfault if called with a NULL parameter, and SMACK_DEFAULT_PROCESS_LABEL is not defined.

man/systemd-system.conf.xml
```diff
@@ -525,6 +525,18 @@
...
+
+      <varlistentry>
+        <term><varname>DefaultSmackProcessLabel=</varname></term>
+
+        <listitem><para>Takes a <option>SMACK64</option> security label as the argument. The process executed
+        by a unit will be started under this label if <varname>SmackProcessLabel=</varname> is not set in the
+        unit. See <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>5</manvolnum></citerefentry>
+        for the details.</para>
+
+        <para>If the value is <literal>/</literal> only labels labels specified with <varname>SmackProcessLabel=</varname>
```
yuwata: "labels labels"

Also, I'd like to add a comma after `<literal>/</literal>`.

BTW, as already discussed at https://github.com/systemd/systemd/pull/23921#discussion_r916666613, still I'd lie to use `-` for disabling the compile-time defaults. As you already commented, normal labels cannot start with '-'. That means we can safely use '-'.

src/core/manager.c
```diff
@@ -3879,6 +3881,20 @@ int manager_get_effective_environment(Manager *m, char ***ret) {
         return 0;
 }
 
+int manager_set_default_smack_process_label(Manager *m, const char *label) {
+        assert(m);
+
+#ifdef SMACK_DEFAULT_PROCESS_LABEL
+        if (!label)
+                return free_and_strdup(&m->default_smack_process_label, SMACK_DEFAULT_PROCESS_LABEL);
+        else
```
yuwata: unnecessary `else`.

src/core/execute.c
```diff
@@ -3251,20 +3252,17 @@ static int setup_smack(
...
                         return r;
-        }
-#ifdef SMACK_DEFAULT_PROCESS_LABEL
-        else {
+        } else if (manager->default_smack_process_label) {
                 _cleanup_free_ char *exec_label = NULL;
 
                 r = mac_smack_read_fd(executable_fd, SMACK_ATTR_EXEC, &exec_label);
                 if (r < 0 && !IN_SET(r, -ENODATA, -EOPNOTSUPP))
                         return r;
```
yuwata, line 3260: This is called only when `DefaultSmackProcessLabel=` or compile-time default is specified. Is it expected?
Is it not necessary to read and apply the attribute even if the default is not specified??

IOW, why this block is conditionalized with `if (manager->default_smack_process_label)`?

yuwata (changes requested):

man/systemd-system.conf.xml
steelman: '-' is actually used in a special way (see `-CIPSO` and `-DELETE`) and I'd like to stay away from it, even if only to avoid confusion (LSM/MAC stuff is complicated enough) between the ways the kernel and systemd use '-'.

src/core/manager.c
steelman: Indeed.

man/systemd-system.conf.xml
poettering: hmm, can we use `~`?

src/core/execute.c
steelman: >This is called only when DefaultSmackProcessLabel= or compile-time default is specified.

Yes. This has been this way since e174dce271 ("smack: add default smack process label config").

> IOW, why this block is conditionalized with if (manager->default_smack_process_label)?

It is further explained in e419a0e310 ("core: set default process label only exec label is none"). I simply follow the logic that was before: the else block was executed if `SMACK_DEFAULT_PROCESS_LABEL` had been defined.

man/systemd-system.conf.xml
steelman: I am afraid we can't

> Single character labels using special characters, that being anything other than a letter or digit, are reserved for use by the Smack development team.

bluca (approved):

PR #23924, opened 2022-07-06
os-release: define SUPPORT_END=

### LINKED ISSUE #21764: [RFE] `SUPPORT_END` field for `/etc/os-release`
When managing multiple systems of different distributions and different versions, it is frustrating to have no neutral way to determine end of support for an OS.

It would be helpful if `/etc/os-release` added a standardized (optional) field like `SUPPORT_END`.  Ideally supporting something like RFC3339 style dates.  From there building a vendor neutral nag tool becomes very simple.

https://www.freedesktop.org/software/systemd/man/os-release.html

Various OS specific tools/tricks each requiring manual tricks.

v249

**Example script using this new attribute**
Some sort of `timer` could run this script on a relevant interval.

```shell
#!/bin/sh

source /etc/os-release

if [ "x$SUPPORT_END" == "x" ]; then
        exit 0
fi

DAYS_LEFT_90=7776000
DAYS_LEFT_60=5184000
DAYS_LEFT_30=2592000
DAYS_LEFT_7=604800

EPOCH_EOL=`date -d $SUPPORT_END +%s`
EPOCH_NOW=`date +%s`

TIME_LEFT=`expr $EPOCH_EOL - $EPOCH_NOW`

PRIORITY="info"

if [ $TIME_LEFT -lt $DAYS_LEFT_7 ]; then
        PRIORITY="crit"
else if [ $TIME_LEFT -lt $DAYS_LEFT_30 ]; then
        PRIORITY="err"
else if [ $TIME_LEFT -lt $DAYS_LEFT_60 ]; then
        PRIORITY="warn"
else if [ $TIME_LEFT -lt $DAYS_LEFT_90 ]; then
        PRIORITY="notice"
fi
fi
fi
fi

echo "$PRETTY_NAME end of support is $SUPPORT_END" | systemd-cat -p $PRIORITY
```

keszybz:
Fixes #21764.

I think is very simple, but flexible. The date may be set early, for distros
that have a fixed schedule, but it doesn't have to. So for example Debian could
push out an update that sets a few months before the release goes EOL. And
various tools, in particular graphical desktops, can start nagging people to
upgrade a few weeks before the date.

As discussed in the bug, we don't need granularity higher than a day. And this
means that we can use a simple human- and machine-readable format.
I was considering other names, e.g. something with "EOL", but I think that
"SUPPORT_END" is better because it doesn't imply that the machine will somehow
stop working. This is supposed to be an advisory, nothing more.

### DIFF
--- a/man/os-release.xml
+++ b/man/os-release.xml
@@ -351,6 +351,16 @@
         </varlistentry>
 
         <varlistentry>
+          <term><varname>SUPPORT_END=</varname></term>
+
+          <listitem><para>The date at which support for this version of the OS ends. (What exactly "lack of
+          support" means varies between vendors, but generally users should assume that updates, including
+          security fixes, will not be provided.) The value is a date in the format
+          <literal>YYYY-MM-DD</literal>, and specifies the last day on which support <emphasis>is</emphasis>
+          provided.</para></listitem>
+        </varlistentry>
+
+        <varlistentry>
           <term><varname>LOGO=</varname></term>
 
           <listitem><para>A string, specifying the name of an icon as defined by <ulink


man/os-release.xml
```diff
@@ -350,6 +350,16 @@
           <literal>BUG_REPORT_URL="https://bugzilla.redhat.com/"</literal>.</para></listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><varname>SUPPORT_END=</varname></term>
+
+          <listitem><para>The date at which support for this version of the OS ends. (What exactly "lack of
+          support" means varies between vendors, but generally users should assume that updates, including
+          security fixes, will not be provided.) The value is a date in the format
+          <literal>YYYY-MM-DD</literal>, and specifies the last day on which support <emphasis>is</emphasis>
```
poettering, line 359: hmm, i guess there's a bikeshedding discussion to be had about the picked date format.

Could also be unix seconds? Or ISO?

I guess the format you picked is a subset of what systemd.time(5) describes, which I guess is a nice property. So I guess this is fine to pick this one.

bluca (approved):

poettering:
Should hostnamed expose a property for this? and then, as a unix timestamp?

and hostnamectl show this?

poettering:
lgtm

poettering:
maybe also add SUPPORT_END_URL= or so, right away, with a helpful link what to do when support ends. i.e. make the stuff actionable.

man/os-release.xml
medhefgo: This is ISO 8601 date format and imho the only sane way to define this.

yuwata (approved):

poettering:
I added a bunch of TODO list items regarding the other ideas:

https://github.com/systemd/systemd/commit/81a965187d01542a965daad7523993f4386fbc2d

PR #23929, opened 2022-07-07
core, mount: Fix early misconfiguration of unit dependencies before /run/mount/utab is updated

d-hatayama:


src/core/mount.c
```diff
@@ -314,7 +322,7 @@ static int mount_add_mount_dependencies(Mount *m) {
         pm = get_mount_parameters_fragment(m);
         if (pm && pm->what &&
             path_is_absolute(pm->what) &&
-            (mount_is_bind(pm) || mount_is_loop(pm) || !mount_is_network(pm))) {
+            (mount_is_bind(pm) || mount_is_loop(pm) || !mount_is_network(m))) {
```
yuwata, line 325: `mount_is_network()` now returns integer. Please do not use implicit cast to boolean.

src/core/mount.c
```diff
@@ -84,8 +84,16 @@ static MountParameters* get_mount_parameters(Mount *m) {
...
 }
 
-static bool mount_is_network(const MountParameters *p) {
-        assert(p);
+static int mount_is_network(Mount *m) {
+        MountParameters *p;
+
+        p = get_mount_parameters(m);
+        if (!p)
+                return -1;
```
yuwata, line 92: Please use negative errno. Maybe `-EBUSY` or `-EAGAIN`?

src/core/mount.c
```diff
@@ -84,8 +84,16 @@ static MountParameters* get_mount_parameters(Mount *m) {
...
+
+        if (m->from_proc_self_mountinfo &&
+            UNIT(m)->manager->last_libmount_event_type != MNT_MONITOR_TYPE_USERSPACE)
+                return -1;
```
yuwata, line 99: Also here.

yuwata (changes requested):

yuwata:
Could you add a test case in TEST-60 or somewhere?

src/core/mount.c
```diff
@@ -84,15 +84,23 @@ static MountParameters* get_mount_parameters(Mount *m) {
...
-                return true;
+        p = get_mount_parameters(m);
+        if (!p)
+                return -1;
 
         if (p->fstype && fstype_is_network(p->fstype))
                 return true;
 
+        if (m->from_proc_self_mountinfo &&
+            UNIT(m)->manager->last_libmount_event_type != MNT_MONITOR_TYPE_USERSPACE)
```
yuwata, line 98: Hmm... `mount_is_network()` is called in the several paths:
- mount_load() -> mount_add_extras() -> mount_add_non_exec_dependencies()
- mount_setup_unit() -> mount_setup_{new,existing}_unit() -> mount_add_non_exec_dependencies()

What happens the following?
1. the manager receives a userspace event which is triggered by a path e.g. /pathA,
2. the manager receives a kernel event triggered by another path e.g. /pathB,
3. then, the mount unit for /pathA gets `MOUNT_PROC_JUST_CHANGED` flag...
4. and subsequent call of `mount_is_network()` for /pathA returns negative...

Does this patch really work as expected??

Anyway, please add a testcase.

d-hatayama: > > Hmm... `mount_is_network()` is called in the several paths:
> > 
> > * mount_load() -> mount_add_extras() -> mount_add_non_exec_dependencies()
> > * mount_setup_unit() -> mount_setup_{new,existing}_unit() -> mount_add_non_exec_dependencies()
> > 
> > What happens the following?
> > 
> > 1. the manager receives a userspace event which is triggered by a path e.g. /pathA,
> > 2. the manager receives a kernel event triggered by another path e.g. /pathB,
> > 3. then, the mount unit for /pathA gets `MOUNT_PROC_JUST_CHANGED` flag...
> > 4. and subsequent call of `mount_is_network()` for /pathA returns negative...
> > 
> > Does this patch really work as expected??

No, this patch doesn't work as you point out.

Honestly, I don't fully understand your example, but it's easy to show another example where this patch doesn't work as expected. For example, first note that mount command updates /run/mount/utab file only when mount options contain those for user-mode applications such as _netdev. Then, the following scenario could occur easily:

1. Assume that systemd process (PID 1) is sleeping in this time for now.
2. Some user mounts some filesystem on /opt with _netdev (thus updateing /run/mount/utab occurs).
3. Some user further mounts another filesystem on /mnt with no mount options for usre-mode applications (thus no updating /run/mount/utab occurs).
4. The currently sleeping systemd process (PID 1) gets awaken here.

In this case, the last libmount event type we get is MNT_MONITOR_TYPE_KERNEL from the step 3. and the MNT_MONITOR_TYPE_USERSPACE from the step 2. is lost. The patch doesn't work like this.

Next, I tried to find another method to know whether the corresponding mount entry is updated with /run/mount/utab in libmount, but I got there is no such interface in libmount now. There is such internal flag in libmount, but it's not exported for users.

The internal flag is MNT_FS_MERGED, which is set within mnt_table_merge_user_fs() at https://github.com/util-linux/util-linux/blob/master/libmount/src/tab_parse.c#L1154. MNT_FS_MERGED is controlled by flags member of libmnt_fs structure at https://github.com/util-linux/util-linux/blob/master/libmount/src/mountP.h#L218 . But libmnt_fs structure is not exported to users. It looks an incomplete type, for example by gdb:

```
(gdb) ptype fs
type = struct libmnt_fs {
    <incomplete type>
} *
```

So, libmount provides several accsessor functions to get access to hidden members of libmnt_fs structure, but the one for flags member is not exported to users, which is defined as static at https://github.com/util-linux/util-linux/blob/master/libmount/src/fs.c#L555-L558. There are still functions for checking the flag state of libmnt_fs structure at https://github.com/util-linux/util-linux/blob/master/libmount/src/fs.c#L595-L637, but unfortunately, only the one for MNT_FS_MERGED is not being provided...

Hence, it is the current situation that we have no method to check if information of /run/mount/utab is reflected for a mount entry given by libmount.

d-hatayama: I've just noticed that even with MNT_FS_MERGED, this patch never fix the issue. Consider again that mount command updates /run/mount/utab file only when mount options contain those for user-mode applications such as _netdev. For such mount entries whose mount options not containing these for user-mode applications, MNT_FS_MERGED is never set. No matter how long we wait, we cannot determine whether a given mount entry is local one or network one for such mount entries. I wish libmount updates utab at all mount.

d-hatayama:
I'm closing this PR since as I mentioned at https://github.com/systemd/systemd/pull/23929#discussion_r925454825 and https://github.com/systemd/systemd/pull/23929#discussion_r925474311, libmount currently provides no interface to check whether a given mount entry is updated with /run/mount/utab and thus fixing the issue in this PR's direction is impossible as long as such interface is newly added into libmount.

I'll switch to PR#23367 as an interim fix.

PR #23932, opened 2022-07-07
boot: Miscellaneous cleanups

medhefgo:


src/boot/efi/linux.c
```diff
@@ -16,7 +16,7 @@
 #include "pe.h"
 #include "util.h"
 
-static EFI_LOADED_IMAGE * loaded_image_free(EFI_LOADED_IMAGE *img) {
+EFI_LOADED_IMAGE_PROTOCOL *loaded_image_free(EFI_LOADED_IMAGE_PROTOCOL *img) {
```
yuwata: Why 'static' dropped??

yuwata (commented):
Looks mostly good. One question.

src/boot/efi/linux.c
medhefgo: It's not needed (everywhere else we use the typedef version anyway).

medhefgo: nvm, by brain saw this as a `struct`.

yuwata (approved):
LGTM.

PR #23933, opened 2022-07-07
time-util: fix buffer-over-run

### LINKED ISSUE #23928: Buffer overrun in format_timespan

251

### Used distribution

_No response_

### Linux kernel version used

_No response_

_No response_

### Component

_No response_

_No response_

In the file `time-util.c`, the function `format_timespan` has the following [code](https://github.com/systemd/systemd/blob/b5fe371c0ea1814738b685634c5d405cfb750a6a/src/basic/time-util.c#L500):

```
char *format_timespan(char *buf, size_t l, usec_t t, usec_t accuracy) {
    char *p = buf;
    for (size_t i = 0; i < ELEMENTSOF(table); i++) {
        size_t n;
        ....
        if (!done) {
            k = snprintf(p, l,
                                "%s"USEC_FMT"%s",
                                p > buf ? " " : "",
                                a,
                                table[i].suffix);

            t = b;
        }
        n = MIN((size_t) k, l);
        l -= n;
        p += n;
    }
    *p = 0;
    return buf;
}
```
The problem of the above code is that `n = MIN((size_t) k, l);` can assign the buffer size `l` to `n`. Then `p += n;` will cause `p` to point to one byte after the buffer `buf`, leading to a buffer overwrite in `*p=0` (an off-by-one error).

To reproduce the buffer overrun, just run the following test code

```
int main() {
    char buf[5];
    char *p;
    usec_t t = 100005;
    usec_t accuracy = 1000;
    p = format_timespan(buf, sizeof(buf), t, accuracy);
    printf("%s\n",p);
    return 0;
}
```
`format_timespan` will write to `buf[5]`, which is an error.

_No response_

yuwata:
Fixes #23928.

### DIFF
--- a/src/test/test-time-util.c
+++ b/src/test/test-time-util.c
@@ -238,6 +238,11 @@ TEST(format_timespan) {
         test_format_timespan_accuracy(1);
         test_format_timespan_accuracy(USEC_PER_MSEC);
         test_format_timespan_accuracy(USEC_PER_SEC);
+
+        /* See issue #23928. */
+        _cleanup_free_ char *buf;
+        assert_se(buf = new(char, 5));
+        assert_se(buf == format_timespan(buf, 5, 100005, 1000));
 }
 
 TEST(verify_timezone) {
--- a/src/basic/time-util.c
+++ b/src/basic/time-util.c
@@ -591,7 +591,7 @@ char *format_timespan(char *buf, size_t l, usec_t t, usec_t accuracy) {
                         t = b;
                 }
 
-                n = MIN((size_t) k, l);
+                n = MIN((size_t) k, l-1);
 
                 l -= n;
                 p += n;


src/test/test-time-util.c
```diff
@@ -238,6 +238,11 @@ TEST(format_timespan) {
         test_format_timespan_accuracy(1);
         test_format_timespan_accuracy(USEC_PER_MSEC);
         test_format_timespan_accuracy(USEC_PER_SEC);
+
+        /* See issue #23928. */
+        _cleanup_free_ char *buf;
+        assert_se(buf = new(char, 5));
+        format_timespan(buf, 5, 100005, 1000);
```
mrc0mmand: I suspect this might need to be prefixed with `(void) ` to make compilers happy, if the results should indeed be ignored.

yuwata: Interesting. I have no idea why my compiler does not complain about that...

src/test/test-time-util.c
```diff
@@ -238,6 +238,11 @@ TEST(format_timespan) {
         test_format_timespan_accuracy(1);
         test_format_timespan_accuracy(USEC_PER_MSEC);
         test_format_timespan_accuracy(USEC_PER_SEC);
+
+        /* See issue #23928. */
+        _cleanup_free_ char *buf;
```
medhefgo, line 243: Why not just `char buf[5]`?

yuwata: I do not know why, but valgrind does not detect the buffer-over-run if `buf[5]` is used.
ASAN may detect the issue even with `buf[5]`. But this is just a test code, hence I hope that's no problem.

dtardon: > I do not know why, but valgrind does not detect the buffer-over-run if `buf[5]` is used.

Because valgrind's `memcheck` doesn't check stack allocations.

yuwata: Ah, thanks.

PR #23937, opened 2022-07-07
Several ASan-related test suite tweaks

mrc0mmand:
With this patchset it's possible to run almost all integration tests under ASan/UBSan.

Two exceptions (as far as I know) are `TEST-35-LOGIN` and `TEST-47-ISSUE-14566`, which suffer from https://github.com/llvm/llvm-project/issues/49223 (https://bugzilla.redhat.com/show_bug.cgi?id=2098125). However, a fix for this is already under review https://reviews.llvm.org/D128992.


yuwata:
@mrc0mmand I guess this works with a paired change in systemd-centos-ci repository, right? Should we wait for the change before merging this PR?

PR #23949, opened 2022-07-08
mac: rework labelling code to be simpler, and less racy

poettering:
This merges the various labelling calls into a single label_fix_full(),
which can operate on paths, on inode fds, and in a dirfd/fname style
(i.e. like openat()). It also systematically separates the path to look
up in the db from the path we actually use to reference the inode to
relabel.

This then ports tmpfiles over to labelling by fd. This should make the
code a bit less racy, as we'll try hard to always operate on the very
same inode, pinning it via an fd.

User-visibly the behaviour should not change.

### DIFF
--- a/src/shared/selinux-util.c
+++ b/src/shared/selinux-util.c
@@ -233,46 +233,19 @@ static int mac_selinux_reload(int seqno) {
 }
 #endif
 
-int mac_selinux_fix_container(const char *path, const char *inside_path, LabelFixFlags flags) {
-
-        assert(path);
-        assert(inside_path);
-
 #if HAVE_SELINUX
-        _cleanup_close_ int fd = -1;
+static int selinux_fix_fd(
+                int fd,
+                const char *label_path,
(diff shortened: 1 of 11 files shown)


yuwata (approved):

src/shared/selinux-util.c
```diff
@@ -282,42 +255,85 @@ int mac_selinux_fix_container_fd(int fd, const char *path, const char *inside_pa
...
+
+                        return -errno;
+                }
+
+                inode_fd = opened_fd;
+        } else
+                inode_fd = atfd;
+
+        if (!label_path) {
+                if (path_is_absolute(inode_path))
```
cgzones, line 322: `label_fix_full()` does not guarantee either `label_path` or `inode_path` is non-NULL and `path_is_absolute()` does not accept a NULL pointer.

bluca: The latter is fixed already by https://github.com/systemd/systemd/pull/23946

Issue #23723, opened 2022-06-12
Possible performance improvements in the integration test suite


mrc0mmand:
As a response to @yuwata's https://github.com/systemd/systemd-centos-ci/issues/475 I introduce several time-saving features in the CentOS CI, which cut down the runtime of the CentOS CI jobs by ~25 - 35%. Apart from that, I noticed that there are several areas for further improvements, which would be beneficial for CIs in general, not only CentOS CI.

### 1) Check the `sleep()`s in the systemd-networkd test suite

There's quite a lot of `sleep()`s in the systemd-networkd test suite and some of them might be reduced or possibly dropped/replaced by other means:

```python
test/test-network/systemd-networkd-tests.py:440:    time.sleep(1)
test/test-network/systemd-networkd-tests.py:472:    time.sleep(1)
test/test-network/systemd-networkd-tests.py:539:                    time.sleep(.2)
test/test-network/systemd-networkd-tests.py:604:        time.sleep(sleep_sec)
test/test-network/systemd-networkd-tests.py:626:                time.sleep(1)
test/test-network/systemd-networkd-tests.py:657:                time.sleep(1)
test/test-network/systemd-networkd-tests.py:714:                time.sleep(1)
test/test-network/systemd-networkd-tests.py:724:                time.sleep(1)
test/test-network/systemd-networkd-tests.py:1323:                time.sleep(1)
test/test-network/systemd-networkd-tests.py:1327:                time.sleep(1)
test/test-network/systemd-networkd-tests.py:2363:        time.sleep(1)
test/test-network/systemd-networkd-tests.py:2552:            time.sleep(1)
test/test-network/systemd-networkd-tests.py:2582:        time.sleep(1)
test/test-network/systemd-networkd-tests.py:2774:        time.sleep(1)
test/test-network/systemd-networkd-tests.py:2810:        time.sleep(1)
test/test-network/systemd-networkd-tests.py:2846:        time.sleep(2)
test/test-network/systemd-networkd-tests.py:3089:        time.sleep(1)
test/test-network/systemd-networkd-tests.py:3099:        time.sleep(1)
test/test-network/systemd-networkd-tests.py:3172:        time.sleep(2)
test/test-network/systemd-networkd-tests.py:3186:        time.sleep(2)
test/test-network/systemd-networkd-tests.py:3194:        time.sleep(2)
test/test-network/systemd-networkd-tests.py:3202:        time.sleep(2)
[...]

PR #23950, opened 2022-07-08
test-network: merge DHCP client tests to improve performance

yuwata:
C.f. #23723.


yuwata:
Compared with #23723,
CentOS CI (Arch Linux): 39:18.83 -> 34:34.36

yuwata:
All CentOS CIs are green now. PTAL!
Note that the test is currently only run in CentOS CIs.

PR #23953, opened 2022-07-08
test: start test user session before idle action setting is changed

### LINKED ISSUE #23952: TEST-35-LOGIN appears to be flaky when systemd is built with coverage
I haven't seen this fail, yet, on a non-coverage build.

As described in https://github.com/systemd/systemd/pull/23930#issuecomment-1178781984, the expected D-Bus message in `test_lock_idle_action()` might never appear, causing the test to fail. This is reliably reproducible with

```
# meson build -Dinstall-tests=true -Db_coverage=true
# ninja -C build
# make -C test/TEST-35-LOGIN clean setup run QEMU_SMP=1
```

```
Jul 08 04:45:10 H systemd-logind[595]: Got message type=signal sender=:1.1 destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=Reloading cookie=3258 reply_cookie=0 signature=b error-name=n/a error-message=n/a
Jul 08 04:45:10 H systemd-logind[595]: System manager has been reloaded, rechecking sessions...
Jul 08 04:45:10 H systemd-logind[595]: System idle. Will be locked now.
Jul 08 04:45:10 H systemd-logind[595]: Locking sessions...
...
Jul 08 04:45:12 H testsuite-35.sh[260]: + assert_eq tty2 tty2
Jul 08 04:45:12 H testsuite-35.sh[676]: + set +ex
Jul 08 04:45:12 H testsuite-35.sh[260]: + timeout 35 bash -c 'while [[ "$(journalctl -b -u systemd-logind.service --since=02:45:10 | grep -c '\''Sent message type=signal .* member=Lock'\'')" -lt 1 ]]; do sleep 1; done'
Jul 08 04:45:37 H systemd-logind[595]: System idle.
Jul 08 04:45:37 H systemd[1]: systemd-journald.service: Got notification message from PID 225 (WATCHDOG=1)
Jul 08 04:45:47 H systemd[1]: Received SIGCHLD from PID 814 (sleep).
Jul 08 04:45:47 H systemd[1]: Child 814 (sleep) died (code=killed, status=15/TERM)
Jul 08 04:45:47 H testsuite-35.sh[260]: + cleanup_test_user
```

Full log: https://jenkins-systemd.apps.ocp.ci.centos.org/job/upstream-cron-build/665/artifact//systemd-centos-ci/artifacts_all/artifacts_mq_ifqlt/vagrant-logs.rfi/vagrant-arch-coverage-testsuite.cKl/TEST-35-LOGIN/system.journal

As I mentioned in #23930, this doesn't seem like a timing issue, since the behavior remains same with large timeout values as well (10 minutes).

/cc @yuwata

yuwata:
Otherwise, idle action may be triggered before starting the test user
session.

This also introduce create_session() and cleanup_session() helper
functions.

Fixes #23952.

### DIFF
--- a/test/units/testsuite-35.sh
+++ b/test/units/testsuite-35.sh
@@ -214,7 +214,7 @@ test_shutdown() {
     assert_eq "$(systemctl show systemd-logind.service -p ExecMainPID --value)" "$pid"
 }
 
-teardown_session() (
+cleanup_session() (
     set +ex
 
     systemctl stop getty@tty2.service
@@ -224,6 +224,12 @@ teardown_session() (
     pkill -u "$(id -u logind-test-user)"
     sleep 1
     pkill -KILL -u "$(id -u logind-test-user)"
+)
+
+teardown_session() (
+    set +ex
+
+    cleanup_session
 
     rm -f /run/udev/rules.d/70-logindtest-scsi_debug-user.rules
     udevadm control --reload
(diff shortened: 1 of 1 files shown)


mrc0mmand (approved):
Tried it locally and I can't reproduce the original issue anymore. Thanks a lot for the quick fix!

yuwata:
@mrc0mmand 
> Tried it locally and I can't reproduce the original issue anymore. Thanks a lot for the quick fix!

Thank you for the quick confirmation! BTW, the green label is set by mistake? Or does this need any other checks?

mrc0mmand:
> @mrc0mmand
> 
> > Tried it locally and I can't reproduce the original issue anymore. Thanks a lot for the quick fix!
> 
> Thank you for the quick confirmation! BTW, the green label is set by mistake? Or does this need any other checks?

Whoops, that was definitely set by a mistake, fixed.

Issue #23955, opened 2022-07-08
Suspicious caching of broken-trust "dual stack" domains in resolved 


mrc0mmand:
### systemd version the issue has been seen with

251

### Used distribution

Fedora 36

### Component

systemd-resolved

### Expected behaviour you didn't see

```
# resolvectl flush-caches
# dig A untrusted.test

; <<>> DiG 9.16.30-RH <<>> A untrusted.test
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4800
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;untrusted.test.			IN	A

;; ANSWER SECTION:
untrusted.test.		7197	IN	A	10.0.0.121

;; Query time: 1 msec
;; SERVER: 127.0.0.53#53(127.0.0.53)
;; WHEN: Fri Jul 08 15:15:16 CEST 2022
;; MSG SIZE  rcvd: 59

# dig AAAA untrusted.test

; <<>> DiG 9.16.30-RH <<>> AAAA untrusted.test
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 29466
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;untrusted.test.			IN	AAAA

;; ANSWER SECTION:
untrusted.test.		7195	IN	AAAA	fd00:dead:beef:cafe::121

;; Query time: 1 msec
;; SERVER: 127.0.0.53#53(127.0.0.53)
;; WHEN: Fri Jul 08 15:15:18 CEST 2022
;; MSG SIZE  rcvd: 71
```

### Unexpected behaviour you saw

```
# resolvectl flush-caches
# dig A untrusted.test

; <<>> DiG 9.16.30-RH <<>> A untrusted.test
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 4152
[...]

Issue #23951, opened 2022-07-08
nss-resolve returns an IPv6 address even when IPv6 is disabled


mrc0mmand:
### systemd version the issue has been seen with

251

### Used distribution

Fedora 36

### Expected behaviour you didn't see

nss-resolve should return an IPv4 address when `net.ipv6.conf.all.disable_ipv6=1`.

### Unexpected behaviour you saw

nss-resolve return an IPv6 address with `net.ipv6.conf.all.disable_ipv6=1`

### Steps to reproduce the problem

```
# cat /etc/hosts
10.0.0.1               ns1.unsigned.test
fd00:dead:beef:cafe::1 ns1.unsigned.test
# sysctl -w net.ipv6.conf.all.disable_ipv6=0
net.ipv6.conf.all.disable_ipv6 = 0
# getent -s resolve hosts ns1.unsigned.test
fd00:dead:beef:cafe::1 ns1.unsigned.test

# sysctl -w net.ipv6.conf.all.disable_ipv6=1
net.ipv6.conf.all.disable_ipv6 = 1
# getent -s resolve hosts ns1.unsigned.test
fd00:dead:beef:cafe::1 ns1.unsigned.test
```

pemensik:
I think resolved should still deliver IPv6 address if requested explicitly by ``-t AAAA``. It might just not start AAAA queries when requested by *getaddrinfo* call with *AF_UNSPEC*.

flokli:
This works as intended - see https://github.com/systemd/systemd/pull/14125 for the rationale.

It's up to clients (implementing RFC6724) to determine how to reach a given destination, whether to consider the v6 records returned or not. The resolver trying to be smart about it is almost always wrong.

flokli:
You have a point. I'd still err on the side of caution.

resolved can speak UDP on port 53, and DNS lookups popping up there might come from systems with IPv6 enabled in the kernel (think about local (firecracker) VMs, which bring their own kernel).

Sure, it's highly unlikely that such a VM then has v6 connectivity over a L2, while IPv6 is explicitly disabled in the host kernel, but still…

--

It feels like adding this complexity in resolved doesn't gain us a lot, except saving some DNS requests that are usually cached anyways.

I'd rather have some `SkipAAAAOnAFUnspec=` config knob in `resolved.conf` that sysadmins can set to `yes` if they really care. I'd argue most browsers etc. doing happy eyeballs do explicit requests anyways.

[...]

PR #23956, opened 2022-07-08
test: cover (not only) IPv6 in the resolved test suite

mrc0mmand:
Marking as draft for now, since it fails due to #23951 and #23955.

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -2641,8 +2641,9 @@ inst_binary() {
     # chown, getent, login, su, useradd, userdel - dlopen()s (not only) systemd's PAM modules
     # ls, stat - pulls in nss_systemd with certain options (like ls -l) when
     #            nsswitch.conf uses [SUCCESS=merge] (like on Arch Linux)
+    # delv, dig - pulls in nss_resolve if `resolve` is in nsswitch.conf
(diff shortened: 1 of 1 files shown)


mrc0mmand:
I commented the failing tests caused by https://github.com/systemd/systemd/issues/23951 and https://github.com/systemd/systemd/issues/23955 to not block the PR further, since the rest should be ready to go.

test/units/testsuite-75.sh
```diff
@@ -16,6 +22,15 @@ run() {
     "$@" |& tee "$RUN_OUT"
 }
 
+disable_ipv6() {
+    sysctl -w net.ipv6.conf.all.disable_ipv6=1
+}
+
+enable_ipv6() {
+    sysctl -w net.ipv6.conf.all.disable_ipv6=0
+    networkctl reconfigure dns0
```
yuwata, line 31: 'networkctl reconfigure' waits for the interface reconfigured, but does not wait for the updated setting propagated to resolved. You may need to add a loop that IPv6 DNS server passed to resolved, though not sure it is really necessary or not.

yuwata (approved):
LGTM. I commented about one concern about enabling/disabling IPv6. But let's address that later when necessary.

PR #23962, opened 2022-07-09
Taint flag "support-ended"

keszybz:


src/basic/os-util.c
```diff
@@ -317,3 +317,35 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+                return log_full_errno(quiet ? LOG_DEBUG : LOG_WARNING, SYNTHETIC_ERRNO(EINVAL),
+                                      "Failed to parse SUPPORT_END= in os-release file, ignoring: %m");
+
+        time_t eol = mktime(&tm);
+        if (eol == (time_t) -1)
+                return log_full_errno(quiet ? LOG_DEBUG : LOG_WARNING, SYNTHETIC_ERRNO(EINVAL),
+                                      "Failed to convert SUPPORT_END= in os-release file, ignoring: %m");
+
+        usec_t ts = now(CLOCK_REALTIME);
+        return ts / USEC_PER_SEC > (usec_t) eol;
```
yuwata: please use DIV_ROUND_UP().

src/test/test-os-util.c
```diff
@@ -72,4 +72,18 @@ TEST(load_os_release_pairs) {
...
+
+        assert_se(os_release_support_ended("1999-01-01", false) == true);
+        assert_se(os_release_support_ended("2090-01-01", false) == false);
+        assert_se(os_release_support_ended("-1-1-1", true) == -EINVAL);
+
+        r = os_release_support_ended(NULL, false);
+        if (r < 0)
+                log_info_errno(r, "Failed to check host: %m");
+        else
+                log_info_errno(r, "Host is unsupported: %s", yes_no(r));
```
yuwata: I prefer `"supported"` with `yes_no(!r)`, though this is a test code, hence no problem.

mrc0mmand:
i*86 is unhappy, since it still thinks the world is going to end in 2038

```
 491/1066 systemd:test / test-os-util                                          FAIL              0.08s   killed by signal 6 SIGABRT
>>> MALLOC_PERTURB_=224 PATH=/builddir/build/BUILD/systemd-251/redhat-linux-build:/builddir/.local/bin:/builddir/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin SYSTEMD_LANGUAGE_FALLBACK_MAP=/builddir/build/BUILD/systemd-251/src/locale/language-fallback-map /builddir/build/BUILD/systemd-251/redhat-linux-build/test-os-util
stderr:
/* test_os_release_support_ended */
Failed to convert SUPPORT_END= in os-release file, ignoring: Invalid argument
Assertion 'os_release_support_ended("2090-01-01", false) == false' failed at src/test/test-os-util.c:79, function test_os_release_support_ended(). Aborting.
```

yuwata (changes requested):

man/org.freedesktop.systemd1.xml
```diff
@@ -1544,6 +1544,15 @@ node /org/freedesktop/systemd1 {
           UTC.</para></listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><literal>support-ended</literal></term>
+
+          <listitem><para>The system is running past the end of support declared by the vendor. See the
+          discussion of <varname>SUPPORT_END=</varname> in
```
medhefgo: Documentation isn't really a discussion…

src/basic/os-util.c
```diff
@@ -317,3 +317,35 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+
+int os_release_support_ended(const char *support_end, bool quiet) {
+        _cleanup_free_ char *t = NULL;
+        int r;
+
+        if (!support_end) {
+                r = parse_os_release(NULL,
+                                     "SUPPORT_END", &t);
+                if (r < 0)
+                        return log_full_errno(r == -ENOENT || quiet ? LOG_DEBUG : LOG_WARNING, r,
```
medhefgo: So you're silently treating `true` as a `LOG_` constant? This looks ugly; is this used anywhere else?

src/basic/os-util.c
```diff
@@ -317,3 +317,35 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+                                              "Failed to read os-release file, ignoring: %m");
+                if (!t)
+                        return false;  /* no end date defined */
+
+                support_end = t;
```
medhefgo: Did you forget `TAKE_PTR`?

src/basic/os-util.c
medhefgo: nvm, I read this as `r == -ENOENT || (quiet ? LOG_DEBUG : LOG_WARNING)`. Could use some `()` around the condition if you ask me.

src/test/test-os-util.c
```diff
@@ -72,4 +72,18 @@ TEST(load_os_release_pairs) {
         assert_se(unsetenv("SYSTEMD_OS_RELEASE") == 0);
 }
 
+TEST(os_release_support_ended) {
+        int r;
+
+        assert_se(os_release_support_ended("1999-01-01", false) == true);
+        assert_se(os_release_support_ended("2090-01-01", false) == false);
```
medhefgo: Surely, systemd will be obsolete by then… :joy_cat:

src/core/main.c
```diff
@@ -1312,29 +1312,35 @@ static int enforce_syscall_archs(Set *archs) {
...
-                                     "\nWelcome to \x1B[%sm%s\x1B[0m!\n",
-                                     isempty(ansi_color) ? "1" : ansi_color,
-                                     isempty(pretty_name) ? "Linux" : pretty_name);
-        else
-                return status_printf(NULL, 0,
-                                     "\nWelcome to %s!\n",
-                                     isempty(pretty_name) ? "Linux" : pretty_name);
+        if (support_end && os_release_support_ended(support_end, false) > 0)
+                status_printf(ANSI_HIGHLIGHT_RED "  !!  " ANSI_NORMAL, 0,
+                              "OS is past its end-of-support date (%s)", support_end);
```
medhefgo: I'd prefer using pretty name and version in this message. Simply stating that "Debian" (or simply OS) is unsupported seems weird to me.

src/basic/os-util.c
keszybz: No, it's supposed to be cleaned up. Either we get a param from the outside, or we read it in.

keszybz: I renamed it and added a comment.

src/test/test-os-util.c
keszybz: Damn, I had to change this to 2037 because of i686. We'll need all to look for a new hobby soon ;)

src/core/main.c
keszybz: "OS" was meant as "this os", not "the os is general". But I changed the text to display the name and version now.

src/test/test-os-util.c
medhefgo: Wait, we aren't using 64bit versions of time APIs? O_O

src/basic/os-util.c
medhefgo: > I renamed it and added a comment.

That wasn't really necessary. For some reason I looked at support_end and thought it was a `char **`.

medhefgo (approved):

src/test/test-os-util.c
bluca: IIRC for some i386 distros that transition is not yet done, because it's ABI-breaking. I should probably book January 2038 as vacation in advance...

bluca (approved):

man/os-release.xml
```diff
@@ -355,9 +355,12 @@
 
           <listitem><para>The date at which support for this version of the OS ends. (What exactly "lack of
           support" means varies between vendors, but generally users should assume that updates, including
-          security fixes, will not be provided.) The value is a date in the format
-          <literal>YYYY-MM-DD</literal>, and specifies the last day on which support <emphasis>is</emphasis>
-          provided.</para></listitem>
+          security fixes, will not be provided.) The value is a date in the ISO 8601 format
```
poettering, line 358: i think this is slightly misleading. ISO 8601 allows so many other ways to specify dates and times. Here, we want to allow a small subset of it. (A subset that also happens to be compatible with our `systemd.time` syntax, btw.)

Hence I would clarify this, and write something like this:

```
The value is a date in the format <literal>YYYY-MM-DD</literal> (this date format is a valid special case of both ISO 8601 and systemd.time(5)).
```

I think that makes clearer that we do not allow all ISO 8601 time formats here.

man/os-release.xml
```diff
@@ -355,9 +355,12 @@
...
+          <literal>YYYY-MM-DD</literal>, and specifies the first day on which support <emphasis>is
+          not</emphasis> provided.</para>
```
poettering, line 360: is not → is no longer?

README
```diff
@@ -359,7 +359,7 @@ WARNINGS and TAINT FLAGS:
         will break if /usr is on a separate late-mounted partition, many of its
         dependencies very likely will break sooner or later in one form or
         another. For example, udev rules tend to refer to binaries in /usr,
-        binaries that link to libraries in /usr or binaries that refer to data
+        binaries that link to libraries in /usr, or binaries that refer to data
```
poettering, line 362: if you touch this, please also suffix /usr with /. It's customary in our documentation to suffix paths referring to directories with a /.

man/org.freedesktop.systemd1.xml
```diff
@@ -1494,21 +1494,87 @@ node /org/freedesktop/systemd1 {
...
+          <command>systemd</command> is invoked. See
+          <ulink url="https://www.freedesktop.org/wiki/Software/systemd/separate-usr-is-broken">
+          Booting Without /usr is Broken</ulink> for details why this is bad.</para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><literal>unmerged-usr</literal></term>
+
+          <listitem><para><filename>/bin</filename>, <filename>/sbin</filename> and
```
poettering, line 1519: please suffix these dirs with /

man/org.freedesktop.systemd1.xml
```diff
@@ -1494,21 +1494,87 @@ node /org/freedesktop/systemd1 {
...
+          <ulink url="https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge">
+            The Case for the /usr Merge
+          </ulink>.</para></listitem>
+
+        </varlistentry>
+
+        <varlistentry>
+          <term><literal>cgroups-missing</literal></term>
+
+          <listitem><para>Support for cgroups is unavailable.</para></listitem>
```
poettering, line 1531: i think an important keyword is that it is mssing in the *kernel*

man/org.freedesktop.systemd1.xml
```diff
@@ -1494,21 +1494,87 @@ node /org/freedesktop/systemd1 {
...
+        </varlistentry>
+
+        <varlistentry>
+          <term><literal>cgroupsv1</literal></term>
+
+          <listitem><para>The system is using the old cgroup hierarchy.</para></listitem>
```
poettering, line 1537: this is a bit misleading, our hierarchy actually looks the same on cgroupsv1 or cgroupsv2, it's just mounted somewhere else...

maybe say "legacy v1 control group subsystem"

man/org.freedesktop.systemd1.xml
```diff
@@ -1494,21 +1494,87 @@ node /org/freedesktop/systemd1 {
...
+          <term><literal>overflowgid-not-65534</literal></term>
+
+          <listitem><para>The kernel overflow UID or GID have a value other than 65534.</para></listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><literal>short-uid-range</literal></term>
+          <term><literal>short-gid-range</literal></term>
+
+          <listitem><para>The UID or GID range assigned to the running systemd instance covers less than
```
poettering, line 1581: the range is not assigned to *systemd*, but to the local *system* as a whole

src/basic/os-util.c
```diff
@@ -317,3 +317,38 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+        if (!support_end) {
+                /* If the caller has the variably handy, they can pass it in. If not, we'll read it
+                 * ourselves. */
+
+                r = parse_os_release(NULL,
+                                     "SUPPORT_END", &_support_end_alloc);
+                if (r < 0)
+                        return log_full_errno((r == -ENOENT || quiet) ? LOG_DEBUG : LOG_WARNING, r,
+                                              "Failed to read os-release file, ignoring: %m");
+                if (!_support_end_alloc)
```
poettering, line 334: probably should use isempty(). people might explicitly set it to the empty string to override the end date locally or so.

src/basic/os-util.c
```diff
@@ -317,3 +317,38 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+        struct tm tm = {};
+
+        const char *k = strptime(support_end, "%Y-%m-%d", &tm);
+        if (!k || *k)
+                return log_full_errno(quiet ? LOG_DEBUG : LOG_WARNING, SYNTHETIC_ERRNO(EINVAL),
+                                      "Failed to parse SUPPORT_END= in os-release file, ignoring: %m");
+
+        time_t eol = mktime(&tm);
+        if (eol == (time_t) -1)
+                return log_full_errno(quiet ? LOG_DEBUG : LOG_WARNING, SYNTHETIC_ERRNO(EINVAL),
```
poettering, line 349: mktime() sets eerrno, hence return that, do not make up any error

src/basic/os-util.c
```diff
@@ -317,3 +317,38 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+
+                support_end = _support_end_alloc;
+        }
+
+        struct tm tm = {};
+
+        const char *k = strptime(support_end, "%Y-%m-%d", &tm);
+        if (!k || *k)
+                return log_full_errno(quiet ? LOG_DEBUG : LOG_WARNING, SYNTHETIC_ERRNO(EINVAL),
+                                      "Failed to parse SUPPORT_END= in os-release file, ignoring: %m");
```
poettering, line 345: strptime() doesn't appear to set errno, hence %m should not be used.

src/basic/os-util.c
```diff
@@ -317,3 +317,38 @@ int load_extension_release_pairs(const char *root, const char *extension, char *
...
+
+        time_t eol = mktime(&tm);
+        if (eol == (time_t) -1)
+                return log_full_errno(quiet ? LOG_DEBUG : LOG_WARNING, SYNTHETIC_ERRNO(EINVAL),
+                                      "Failed to convert SUPPORT_END= in os-release file, ignoring: %m");
+
+        usec_t ts = now(CLOCK_REALTIME);
+        return DIV_ROUND_UP(ts, USEC_PER_SEC) > (usec_t) eol;
```
poettering, line 353: i think technically more correct would be `ts > (usec_t) eol * USEC_PER_SEC`, no?

src/core/manager.c
```diff
@@ -4494,6 +4495,9 @@ char* manager_taint_string(const Manager *m) {
         if (clock_is_localtime(NULL) > 0)
                 stage[n++] = "local-hwclock";
 
+        if (os_release_support_ended(NULL, true) > 0)
```
poettering, line 4498: maybe add `/* quiet= */` before the second arg?

poettering:
looks ok, but found some nitpicks, and the only thing that matters is that one error handling thing.

PR #23963, opened 2022-07-10
Replace __sync intrinsics with __atomic

matoro:
As documented in the [GCC manual](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html), the `__sync` intrinsics are considered legacy.  The `__atomic` intrinsics should be used instead.  For now, I have specified `__ATOMIC_SEQ_CST` as the memory model to use for all operations.  This does result in a bit of ugliness in some areas, specifically when translating `__sync_bool_compare_and_swap` to `__atomic_compare_exchange_n`, because of two reasons:  the latter requires the second parameter to be a pointer while the former does not, and the latter makes a write to the second parameter on failure, while the former does not.

In GCC at least, if the platform is missing a hardware implementation for a `__sync` intrinsic, it is up to the user to provide a software implementation.  But, for platforms lacking a hardware implementation for an `__atomic` intrinsic, GCC provides a software implementation in libatomic.  This would allow compiling on such platforms.

There are no platforms which have `__sync` intrinsics for a specific byte-size which lack a corresponding `__atomic` intrinsic for the same size.  However, at least according to https://github.com/buildroot/buildroot/commit/6856e417da4f3aa77e2a814db2a89429af072f7d, there ARE platforms for which GCC has a hardware implementation of the `__sync` intrinsics for a particular size but not one for the corresponding `__atomic` ones, which therefore require linking with libatomic in order to support the `__atomic` intrinsics of the same size.

This requires GCC 4.7 as the minimum version.

I used [this downstream Void patch for apr](https://github.com/void-linux/void-packages/blob/df7ebc36a9ffb208a0d60da58a98a2c4936b1735/srcpkgs/apr/patches/atomic64.patch) and [this ffmpeg patch](https://ffmpeg.org/pipermail/ffmpeg-devel/2014-October/164556.html) as references.

### DIFF
--- a/README
+++ b/README
@@ -207,7 +207,8 @@ REQUIREMENTS:
         python >= 3.5
         meson >= 0.53.2
         ninja
-        gcc, awk, sed, grep, and similar tools
+        gcc >= 4.7
+        awk, sed, grep, and similar tools
         clang >= 10.0, llvm >= 10.0 (optional, required to build BPF programs
                 from source code in C)
         gnu-efi >= 3.0.5 (optional, required for systemd-boot)


src/basic/sigbus.c
```diff
@@ -29,23 +29,29 @@ static void sigbus_push(void *addr) {
 
         /* Find a free place, increase the number of entries and leave, if we can */
         for (size_t u = 0; u < SIGBUS_QUEUE_MAX; u++)
-                if (__sync_bool_compare_and_swap(&sigbus_queue[u], NULL, addr)) {
-                        __sync_fetch_and_add(&n_sigbus_queue, 1);
+                /* OK to initialize this here since we haven't started the atomic ops yet */
+                void *tmp = NULL;
```
medhefgo, line 33: The `{}` fix from next patch needs to be in this one, really.

matoro: Oops!  Not sure why my compiler warned for the other one but not for this one.

src/basic/sigbus.c
```diff
@@ -81,8 +81,8 @@ int sigbus_pop(void **ret) {
 
                         /* OK if we clobber addr here, since we either immediately return
                          * or it will be immediately reinitialized on next loop */
-                        if (__atomic_compare_exchange_n(&sigbus_queue[u], &addr, NULL),
-                            false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST) {
+                        if (__atomic_compare_exchange_n(&sigbus_queue[u], &addr, NULL,
+                            false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)) {
```
poettering: that fix should be squashed into the previous commit?

src/journal/managed-journal-file.c
```diff
@@ -167,19 +167,31 @@ static void managed_journal_file_set_offline_internal(ManagedJournalFile *f) {
...
-                        if (!__sync_bool_compare_and_swap(&f->file->offline_state, OFFLINE_CANCEL, OFFLINE_DONE))
-                                continue;
-                        return;
+                        {
+                                OfflineState tmp_state = OFFLINE_CANCEL;
+                                if (!__atomic_compare_exchange_n(&f->file->offline_state, &tmp_state, OFFLINE_DONE,
+                                    false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST))
+                                        continue;
+                                return;
+                        }
```
poettering, line 174: we usually would indent this like this:

```c
case OFFLINE_CANCEL: {
        OfflineState tmp_state …
}
case OFFLINE_AGAIN_FROM_SYNCING:…
```

please follow the same logic here

src/journal/managed-journal-file.c
```diff
@@ -167,19 +167,31 @@ static void managed_journal_file_set_offline_internal(ManagedJournalFile *f) {
         for (;;) {
                 switch (f->file->offline_state) {
                 case OFFLINE_CANCEL:
-                        if (!__sync_bool_compare_and_swap(&f->file->offline_state, OFFLINE_CANCEL, OFFLINE_DONE))
-                                continue;
-                        return;
+                        {
+                                OfflineState tmp_state = OFFLINE_CANCEL;
+                                if (!__atomic_compare_exchange_n(&f->file->offline_state, &tmp_state, OFFLINE_DONE,
+                                    false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST))
```
poettering: continuation lines of function calls we usually indent with 16ch

src/journal/managed-journal-file.c
```diff
@@ -166,19 +166,28 @@ static void managed_journal_file_set_offline_internal(ManagedJournalFile *f) {
 
         for (;;) {
                 switch (f->file->offline_state) {
-                case OFFLINE_CANCEL:
-                        if (!__sync_bool_compare_and_swap(&f->file->offline_state, OFFLINE_CANCEL, OFFLINE_DONE))
-                                continue;
+                case OFFLINE_CANCEL: {
+                                OfflineState tmp_state = OFFLINE_CANCEL;
```
poettering: indentation should be 8 instead of 16 here.

src/basic/sigbus.c
```diff
@@ -72,8 +79,13 @@ int sigbus_pop(void **ret) {
                         if (!addr)
                                 continue;
 
-                        if (__sync_bool_compare_and_swap(&sigbus_queue[u], addr, NULL)) {
-                                __sync_fetch_and_sub(&n_sigbus_queue, 1);
+                        /* OK if we clobber addr here, since we either immediately return
+                         * or it will be immediately reinitialized on next loop */
+                        if (__atomic_compare_exchange_n(&sigbus_queue[u], &addr, NULL,
+                            false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)) {
```
yuwata: spurious indentation.

src/basic/sigbus.c
```diff
@@ -28,24 +28,31 @@ static void sigbus_push(void *addr) {
...
                 c = n_sigbus_queue;
 
                 if (c > SIGBUS_QUEUE_MAX) /* already overflow */
                         return;
 
-                if (__sync_bool_compare_and_swap(&n_sigbus_queue, c, c + SIGBUS_QUEUE_MAX))
+                /* OK if we clobber c here, since we either immediately return
+                 * or it will be immediately reinitialized on next loop */
+                if (__atomic_compare_exchange_n(&n_sigbus_queue, &c, c + SIGBUS_QUEUE_MAX,
+                    false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST))
```
yuwata: indentation.

src/basic/sigbus.c
```diff
@@ -28,24 +28,31 @@ static void sigbus_push(void *addr) {
...
 
         /* Find a free place, increase the number of entries and leave, if we can */
-        for (size_t u = 0; u < SIGBUS_QUEUE_MAX; u++)
-                if (__sync_bool_compare_and_swap(&sigbus_queue[u], NULL, addr)) {
-                        __sync_fetch_and_add(&n_sigbus_queue, 1);
+        for (size_t u = 0; u < SIGBUS_QUEUE_MAX; u++) {
+                /* OK to initialize this here since we haven't started the atomic ops yet */
+                void *tmp = NULL;
+                if (__atomic_compare_exchange_n(&sigbus_queue[u], &tmp, addr,
+                    false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)) {
```
yuwata: also here.

src/journal/managed-journal-file.c
```diff
@@ -166,19 +166,28 @@ static void managed_journal_file_set_offline_internal(ManagedJournalFile *f) {
 
         for (;;) {
                 switch (f->file->offline_state) {
-                case OFFLINE_CANCEL:
-                        if (!__sync_bool_compare_and_swap(&f->file->offline_state, OFFLINE_CANCEL, OFFLINE_DONE))
+                case OFFLINE_CANCEL: {
+                        OfflineState tmp_state = OFFLINE_CANCEL;
+                        if (!__atomic_compare_exchange_n(&f->file->offline_state, &tmp_state, OFFLINE_DONE,
+                                false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST))
```
yuwata: also here.

yuwata (changes requested):
Please fix the indentation for `__atomic_compare_exchange_n()`.

vcaputo:
> This requires GCC 4.7 as the minimum version.

The README should reflect this, gcc is mentioned currently but without any minimum version.  Unless I'm mistaken, with this patch, the minimum version becomes established at 4.7.

src/journal/managed-journal-file.c
matoro: This one is correct I think?

yuwata: We usually use the following style:
```c
func(a, b, c, d);
func(a, b,
     c, d); /* aligned with the first argument */
some_long_name_func(
                long_argument_a,  /* 16ch indent */
                long_argument_b,
                long_argument_c,
                long_argument_d);

```

poettering:
and as mentioned before, function call continuation lines should double indentation (i.e. 16ch)

matoro:
> and as mentioned before, function call continuation lines should double indentation (i.e. 16ch)

Is it 16ch or aligned with the first argument as @yuwata commented?  https://github.com/systemd/systemd/pull/23963#discussion_r918133735

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -205,22 +205,34 @@ static int journal_file_set_online(JournalFile *f) {
                         break;
 
                 case OFFLINE_SYNCING:
-                        if (!__sync_bool_compare_and_swap(&f->offline_state, OFFLINE_SYNCING, OFFLINE_CANCEL))
-                                continue;
+                        {
```
medhefgo: `{}` style should be like the others.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -205,22 +205,34 @@ static int journal_file_set_online(JournalFile *f) {
...
+                                OfflineState tmp_state = OFFLINE_SYNCING;
+                                if (!__atomic_compare_exchange_n(&f->offline_state, &tmp_state, OFFLINE_CANCEL,
+                                    false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST))
```
medhefgo, line 210: Indent like the others.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -205,22 +205,34 @@ static int journal_file_set_online(JournalFile *f) {
...
+                                        continue;
+                        }
                         /* Canceled restart from syncing, no need to wait. */
                         wait = false;
                         break;
 
                 case OFFLINE_AGAIN_FROM_OFFLINING:
-                        if (!__sync_bool_compare_and_swap(&f->offline_state, OFFLINE_AGAIN_FROM_OFFLINING, OFFLINE_CANCEL))
-                                continue;
+                        {
```
medhefgo: `{}` style.

src/libsystemd/sd-journal/journal-file.c
```diff
@@ -205,22 +205,34 @@ static int journal_file_set_online(JournalFile *f) {
...
+                                OfflineState tmp_state = OFFLINE_AGAIN_FROM_OFFLINING;
+                                if (!__atomic_compare_exchange_n(&f->offline_state, &tmp_state, OFFLINE_CANCEL,
+                                    false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST))
```
medhefgo, line 230: `indent`

src/libsystemd/sd-journal/journal-send.c
```diff
@@ -58,7 +58,9 @@ static int journal_fd(void) {
 
         fd_inc_sndbuf(fd, SNDBUF_SIZE);
 
-        if (!__sync_bool_compare_and_swap(&fd_plus_one, 0, fd+1)) {
+        int zero = 0;
+        if (!__atomic_compare_exchange_n(&fd_plus_one, &zero, fd+1,
+            false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)) {
```
medhefgo: indent.

src/libsystemd/sd-journal/journal-send.c
```diff
@@ -58,7 +58,9 @@ static int journal_fd(void) {
 
         fd_inc_sndbuf(fd, SNDBUF_SIZE);
 
-        if (!__sync_bool_compare_and_swap(&fd_plus_one, 0, fd+1)) {
+        int zero = 0;
+        if (!__atomic_compare_exchange_n(&fd_plus_one, &zero, fd+1,
```
medhefgo: Wonder if this is better…
```suggestion
        if (!__atomic_compare_exchange_n(&fd_plus_one, &(int){0}, fd+1,
```

src/basic/process-util.c
```diff
@@ -1158,7 +1158,13 @@ pid_t getpid_cached(void) {
          * https://sourceware.org/git/gitweb.cgi?p=glibc.git;h=c579f48edba88380635ab98cb612030e3ed8691e
          */
 
-        current_value = __sync_val_compare_and_swap(&cached_pid, CACHED_PID_UNSET, CACHED_PID_BUSY);
+        __atomic_compare_exchange_n(
+                &cached_pid,
```
medhefgo: indent with 16ch.

medhefgo (commented):
FYI, a good code editor should let you just format selections using clang-format. Which is usually the right format.

src/libsystemd/sd-journal/journal-send.c
matoro: I had no idea you could do this.  Learned something new today

src/libsystemd/sd-journal/journal-send.c
```diff
@@ -58,7 +58,8 @@ static int journal_fd(void) {
 
         fd_inc_sndbuf(fd, SNDBUF_SIZE);
 
-        if (!__sync_bool_compare_and_swap(&fd_plus_one, 0, fd+1)) {
+        if (!__atomic_compare_exchange_n(&fd_plus_one, &(int){0}, fd+1,
```
poettering, line 61: please make this `&(const int) {0}`, so that the compiler can share the value with other stuff.

matoro: This won't work actually, because the second parameter must now be writable, hence why I couldn't just use a literal `0`.

If I add that I get:
```
../src/libsystemd/sd-journal/journal-send.c:62:17: warning: passing argument 2 of ‘__atomic_compare_exchange_4’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
```

PR #23972, opened 2022-07-11
sd-device: do not return -ENOENT by sd_device_get_is_initialized()

yuwata:


src/libsystemd/sd-device/sd-device.c
```diff
@@ -1589,6 +1589,9 @@ _public_ int sd_device_get_is_initialized(sd_device *device) {
         assert_return(device, -EINVAL);
 
         r = device_read_db(device);
+        if (r == -ENOENT)
+                /* The device may be already removed or renamed. */
+                return false;
```
poettering, line 1594: hmm, device_read_db_internal_filename() already eats up ENOENT internally? This should not be necessary?

yuwata: `device_read_db_internal_filename()` does.
But `device_read_db_internal()` -> `device_get_device_id()` -> `sd_device_get_subsystem()` may return ENOENT.

PR #23973, opened 2022-07-11
Fix automatic screen rotation for Asus Transformer T100TAM

MikkCZ:
This fixes screen rotation on Asus Transformer T100TAM in landscape mode, where the screen without this patch is rotated upside down. This is the same fix as for T100TAF in #23233.

**QUESTION:** Wouldn't `sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnT100TA*:*` work for the whole line of T100TA Asus Transformers? Would it be a viable change instead?

### DIFF
--- a/hwdb.d/60-sensor.hwdb
+++ b/hwdb.d/60-sensor.hwdb
@@ -132,6 +132,7 @@ sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:pnT300CHI:*
 sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnM80TA:*
 sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnT100TA:*
 sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnT100TAF:*
+sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnT100TAM:*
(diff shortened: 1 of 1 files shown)


yuwata (approved):

yuwata:
> **QUESTION:** Wouldn't `sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnT100TA*:*` work for the whole line of T100TA Asus Transformers? Would it be a viable change instead?

`pnT100TA` matches with `pnT100TA*`. Hence, if you now the devices with all suffixes after pnT100TA support the same parameter, then it is ok to merge the three lines.

poettering:
> **QUESTION:** Wouldn't `sensor:modalias:acpi:INVN6500*:dmi:*svnASUSTeK*:*pnT100TA*:*` work for the whole line of T100TA Asus Transformers? Would it be a viable change instead?

it's always easy to one day merge multiple lines that allow this into one, and using a `*` for it. Going the other way is much harder, because you won't know anymore what the known devices where.

Hence, I'd probably use the explicit non-wildcarded lines until it's extremely clear that it applies to the whole line of devices. Not sure if that's this point is reached yet... Can't really say when it is reached ;-) Maybe only once the line ceases production entirely...

MikkCZ:
Right, that's a good point. Let's keep this PR to only fix the M variant.

PR #23979, opened 2022-07-11
nspawn: Support relative paths for --bind and --overlay

daandemeyer:


src/nspawn/nspawn-mount.c
```diff
@@ -83,13 +83,22 @@ static int custom_mount_compare(const CustomMount *a, const CustomMount *b) {
...
+                if (!allow_empty)
+                        return -EINVAL;
 
-        if (*p == '+')
-                p++;
+                *ret = NULL;
+                return 0;
+        }
+
+        if (*p == '+' && !path_is_absolute(p++))
```
poettering: i seriously dislike if checks with side-effects... it's confusing, so confusing in fact that this is wrong, no? you mean ++p, not p++...

src/nspawn/nspawn-mount.c
```diff
@@ -83,13 +83,22 @@ static int custom_mount_compare(const CustomMount *a, const CustomMount *b) {
         return CMP(a->type, b->type);
 }
 
-static bool source_path_is_valid(const char *p) {
+static int source_path_parse(const char *p, bool allow_empty, char **ret) {
```
poettering: allow_empty sounds if you'd just allow empty strings. But you do more than that.. you convert them into NULL...

i am not sure this should be a bool param. instead, just drop it here entirely, but add a new helper func: source_path_parse_nullable() that wraps source_path_path(), but converts empty string to NULL too.

src/nspawn/nspawn-mount.c
daandemeyer: Yeah,  I figured this was pushing it a bit, I split it into multiple lines

src/nspawn/nspawn-mount.c
```diff
@@ -83,15 +83,33 @@ static int custom_mount_compare(const CustomMount *a, const CustomMount *b) {
...
+
+        if (strempty(p)) {
+                *ret = NULL;
+                return 0;
+        }
+
+        return source_path_parse(p, ret);
+}
+
+
```
poettering: spurious double new line

src/nspawn/nspawn-mount.c
```diff
@@ -235,10 +253,11 @@ int bind_mount_parse(CustomMount **l, size_t *n, const char *s, bool read_only)
...
 
-        if (isempty(source))
-                source = mfree(source);
-        else if (!source_path_is_valid(source))
-                return -EINVAL;
+        r = source_path_parse_nullable(source, &p);
+        if (r < 0)
+                return r;
+
+        free_and_replace(source, p);
```
poettering: instead of moving p → source, why not just change the one use of `source` below to just use `p`?

src/nspawn/nspawn-mount.c
```diff
@@ -295,43 +314,60 @@ int overlay_mount_parse(CustomMount **l, size_t *n, const char *s, bool read_onl
...
+
+                r = source_path_parse(lower[1], &p);
+                if (r < 0)
+                        return r;
+
+                free_and_replace(lower[1], p);
 
                 upper = TAKE_PTR(lower[1]);
 
                 destination = strdup(upper[0] == '+' ? upper+1 : upper); /* take the destination without "+" prefix */
```
poettering, line 349: source_path_parse() already dropped the '+' prefix, afaics

daandemeyer: Modified source_path_parse() to not drop the '+' prefix

src/nspawn/nspawn-mount.c
```diff
@@ -295,43 +316,60 @@ int overlay_mount_parse(CustomMount **l, size_t *n, const char *s, bool read_onl
...
         if (k == 2) {
+                _cleanup_free_ char *p = NULL;
+
                 /* If two parameters are specified, the first one is the lower, the second one the upper directory. And
                  * we'll also define the destination mount point the same as the upper. */
 
-                if (!source_path_is_valid(lower[0]) ||
-                    !source_path_is_valid(lower[1]))
-                        return -EINVAL;
+                r = source_path_parse(lower[0], &p);
```
poettering, line 335: so, previously, if you passed the empty string here, then this would result in EINVAL.

With this change you'll convert the empty string to the current working directory. I am not sure that's desirable (that's because calling path_make_absolute_cwd() with an empty string will return the cwd)

daandemeyer: Yikes this is not my proudest pull request

PR #23981, opened 2022-07-11
udev: add flag to allow disabling blkid probing

dlitz:
This can be useful for users of slow block devices.

For example, the persistent-storage rules are needed for USB floppy
drives be recognized by udisks2, but the extra blkid calls cause
thrashing for 25+ seconds after every disk change.

With this change, a user wishing to avoid the extra blkid invocation(s)
could create /etc/udev/rules.d/55-floppy-noprobe.rules as follows:

    # Don't probe PC floppy drives
    SUBSYSTEM=="block", KERNEL=="fd*", \
        ENV{UDEV_DISABLE_PERSISTENT_STORAGE_BLKID_FLAG}="1"

    # Don't probe USB floppy drives
    SUBSYSTEM=="block", SUBSYSTEMS=="usb", \
        ATTRS{bInterfaceClass}=="08", ATTRS{bInterfaceSubClass}=="04", \
        ENV{UDEV_DISABLE_PERSISTENT_STORAGE_BLKID_FLAG}="1"

I didn't exclude floppies by default in this change, because floppy
devices are also emulated by some BIOSes/hypervisors in some cases, and
I don't know how many systems would fail to boot if /dev/disk/by-uuid/*
became unavailable for 'floppy disks' on those systems.

### DIFF
--- a/rules.d/60-persistent-storage.rules
+++ b/rules.d/60-persistent-storage.rules
@@ -99,6 +99,9 @@ ENV{DEVTYPE}=="partition", ENV{ID_PATH_ATA_COMPAT}=="?*", SYMLINK+="disk/by-path
 KERNEL=="vd*[!0-9]", ENV{ID_PATH}=="pci-*", SYMLINK+="disk/by-path/virtio-$env{ID_PATH}"
 KERNEL=="vd*[0-9]", ENV{ID_PATH}=="pci-*", SYMLINK+="disk/by-path/virtio-$env{ID_PATH}-part%n"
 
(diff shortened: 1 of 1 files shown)


poettering:
physical floppy disks in 2022?

poettering:
I guess I am not opposed to this... @yuwata?

dlitz:
Heh, yes, I'm doing some long-overdue data archival. :)

yuwata:
Sorry for late response. LGTM.

PR #23988, opened 2022-07-12
core: drop ambient capabilities in user manager

steelman:
Neither ambient capabilities should be passed implicitly to user
services. Dropping them does not affect the permitted and effective sets
which are important for the manager itself to operate.

### DIFF
--- a/src/core/main.c
+++ b/src/core/main.c
@@ -2825,6 +2825,11 @@ int main(int argc, char *argv[]) {
                 /* clear the kernel timestamp, because we are not PID 1 */
                 kernel_timestamp = DUAL_TIMESTAMP_NULL;
 
+                /* Clear ambient capabilities, so services do not inherit them implicitly. Dropping them does
+                 * not affect the permitted and effective sets which are important for the manager itself to
+                 * operate. */
+                capability_ambient_set_apply(0, /* also_inherit= */ false);
+
                 if (mac_selinux_init() < 0) {
                         error_message = "Failed to initialize SELinux support";
                         goto finish;


steelman:
@yuwata does `pid1` label mean only the system manager or does it cover `systemd --user` too? Because this patch is meant  for the user manager.

yuwata:
> @yuwata does `pid1` label mean only the system manager or does it cover `systemd --user` too? Because this patch is meant for the user manager.

Nice catch.

src/core/main.c
```diff
@@ -2816,6 +2816,10 @@ int main(int argc, char *argv[]) {
                 /* clear the kernel timestamp, because we are not PID 1 */
                 kernel_timestamp = DUAL_TIMESTAMP_NULL;
 
+                /* Clear ambient capabilities, so services do not inherit
+                them implicitly. */
+                capability_ambient_set_apply(0, false);
```
poettering: nowadays, when invoking functions with boolean arguments, we prefer if a comment is inserted like this:

```c
capability_ambient_set_apply(0, /* also_inherit= */ false);
```

to make it easier to grok what that param is about.

src/core/main.c
```diff
@@ -2816,6 +2816,10 @@ int main(int argc, char *argv[]) {
                 /* clear the kernel timestamp, because we are not PID 1 */
                 kernel_timestamp = DUAL_TIMESTAMP_NULL;
 
+                /* Clear ambient capabilities, so services do not inherit
+                them implicitly. */
```
poettering: can you please add the comment you added to the commit msg here, i.e. that dropping the ambient caps doesn't affect the caps effective for the service manager itself, but only what it passes further down. this is important info for the reader of the code.

(btw, we line break at 109chars these days)

poettering:
can you drop the "RFC" from the commit msg?

poettering:
lgtm, otherwise

src/core/main.c
hlovdal: There are three ways to write code:

# 1) Write unreadable code

```C
capability_ambient_set_apply(0, false);
```

# 2) Write unreadable code, and then use comments to make it readable

```C
capability_ambient_set_apply(0, /* also_inherit= */ false);
```
_"comments often are used as a deodorant"_ -- [Martin Fowler, Kent Beck](https://martinfowler.com/books/refactoring.html).

# 3) Write readable code

```C
enum capability_inherit_mode {
    INHERIT,
    NO_INHERIT,
};
...
capability_ambient_set_apply(0, NO_INHERIT);
```

If you have a function named for instance `SetVisible`, then having a boolean argument is perfectly fine. A function named `capability_ambient_set_apply` not so much.

Applying the list *Characteristics of a Good API* from Joshua Bloch's excellent talk [How To Design A Good API and Why it Matters](https://www.youtube.com/watch?v=aAb7hSCtvGw) show that using `bool`  fails on the majority of criteria while `enum` fails on none:

|criteria|using `bool` argument in `capability_ambient_set_apply`|using `enum` argument in `capability_ambient_set_apply`|
|--------|-----|-----|
|Easy to learn|❌ FAIL|✅ PASS|
|Easy to use, even without documentation|❌ FAIL|✅ PASS|
|Hard to misuse|❌ FAIL|✅ PASS|
|Easy to read and maintain code that uses it|❌ FAIL|✅ PASS|
|Sufficiently powerful to satisfy requirements|✅ PASS|✅ PASS|
|Easy to extend|✅ PASS|✅ PASS|
|Appropriate to audience|✅ PASS|✅ PASS|

Please get rid of using a bool argument in `capability_ambient_set_apply`.

poettering: let's not overcomplicate things.

hlovdal: In what way is making code easier to read "overcomplicate"?

yuwata: Simply, we do not use such coding style. If you'd like such, use it in your repository.

hlovdal: Notice that the following is a complaint about argument validity, **not code style**.

Make sure you frame yourself for that difference before continuing reading (and just to be clear, a valid argument is in no way the same thing as something you agree with).

----

> Simply, we do not use such coding style.

That argument sounds awfully similar to [that's the way we have always done it](https://html.duckduckgo.com/html?q=that%27s%20the%20way%20we%20have%20always%20done%20it). Am I wrong? Are there any differences?

There might very well exist good reasons for not using enums, and if so please use [valid arguments](https://www.thoughtco.com/validity-argument-1692577) to back it up.

vcaputo: It's important that coding style be consistent within a given project.  If you're proposing changing systemd's coding style then start contributing enough code to have a say and create a PR changing doc/CODING_STYLE.md accordingly.

As far as using enums in place of bools, personally I find it annoying vs. just having variants of the function encoding the true/false values in the function names.  Keeping the bool parameter version local to two public wrappers omitting the additional argument altogether.

i.e.

capability_set_apply_inherited(0);
capability_set_apply_uninherited(0);

would just be one-liner public wrappers to the existing bool-receiving capability_set_apply(0, {true,false}); made private.  That way the annoying bool-passing calls in the wrappers are adjacent to the function where it's obvious, and nobody ever has to write that again.

But that's just a personal preference, systemd has established practices and several developers who must agree on such things.

daiaji:
So actually, I can't use AmbientCapabilities in the user service, which sounds unfortunate, I do want to allow some programs in the service to use some privileges when running as an unprivileged user, specifically podman.

PR #23990, opened 2022-07-13
test-network: save and restore timezone

yuwata:
Several DHCP client tests change the system timezone.
Let's save the current timezone at the beginning, and restore it with
the saved value at the end.


keszybz:
Looks reasonable. The tests fail, but it's `test-bus-track` in the few cases I checked.

PR #23991, opened 2022-07-13
udev: reload rules and builtins only when mtime of a config changed

yuwata:
This makes udevd reload rules only when the timestamp is updated,
even on SIGHUP or `udevadm control --reload`.
So, we can call `udevadm control --reload` without huge performance
penalty when no rules, .link files, and so on are changed.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -276,6 +276,9 @@ def expectedFailureIfFQPIEIsNotAvailable():
 
     return f
 
+def udev_reload():
+    check_output('udevadm control --reload')
+
 def copy_network_unit(*units, copy_dropins=True):
     """
     Copy networkd unit files into the testbed.
@@ -287,17 +290,26 @@ def copy_network_unit(*units, copy_dropins=True):
 
     When a drop-in file is specified, its unit file is also copied in automatically.
     """
+    has_link = False
     mkdir_p(network_unit_dir)
     for unit in units:
         if copy_dropins and os.path.exists(os.path.join(networkd_ci_temp_dir, unit + '.d')):
             cp_r(os.path.join(networkd_ci_temp_dir, unit + '.d'), os.path.join(network_unit_dir, unit + '.d'))
+
         if unit.endswith('.conf'):
             dropin = unit
             unit = os.path.dirname(dropin).rstrip('.d')
             dropindir = os.path.join(network_unit_dir, unit + '.d')
             mkdir_p(dropindir)
             cp(os.path.join(networkd_ci_temp_dir, dropin), dropindir)
+
         cp(os.path.join(networkd_ci_temp_dir, unit), network_unit_dir)
+
+        if unit.endswith('.link'):
+            has_link = True
+
+    if has_link:
+        udev_reload()
 
 def remove_network_unit(*units):
     """
@@ -305,12 +317,29 @@ def remove_network_unit(*units):
 
     Drop-ins will be removed automatically.
     """
+    has_link = False
     for unit in units:
         rm_f(os.path.join(network_unit_dir, unit))
         rm_rf(os.path.join(network_unit_dir, unit + '.d'))
 
+        if unit.endswith('.link') or unit.endswith('.link.d'):
+            has_link = True
+
+    if has_link:
+        udev_reload()
+
 def clear_network_units():
+    has_link = False
+    if os.path.exists(network_unit_dir):
+        units = os.listdir(network_unit_dir)
+        for unit in units:
+            if unit.endswith('.link') or unit.endswith('.link.d'):
+                has_link = True
+
     rm_rf(network_unit_dir)
+
+    if has_link:
+        udev_reload()
 
 def copy_networkd_conf_dropin(*dropins):
     """Copy networkd.conf dropin files into the testbed."""
@@ -896,7 +925,6 @@ class NetworkctlTests(unittest.TestCase, Utilities):
     @expectedFailureIfAlternativeNameIsNotAvailable()
     def test_altname(self):
         copy_network_unit('26-netdev-link-local-addressing-yes.network', '12-dummy.netdev', '12-dummy.link')
-        check_output('udevadm control --reload')
         start_networkd()
         self.wait_online(['dummy98:degraded'])
 
@@ -3891,7 +3919,6 @@ class NetworkdSRIOVTests(unittest.TestCase, Utilities):
     @expectedFailureIfNetdevsimWithSRIOVIsNotAvailable()
     def test_sriov_udev(self):
         copy_network_unit('25-sriov.link', '25-sriov-udev.network')
-        call('udevadm control --reload')
 
         call('modprobe netdevsim')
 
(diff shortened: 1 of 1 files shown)


yuwata:
CentOS CI (CentOS Stream 8) failure should be unrelated:
```
======================================================================
ERROR: test_prefix_route (__main__.NetworkdNetworkTests)
----------------------------------------------------------------------
  File "./test/test-network/systemd-networkd-tests.py", line 2228, in test_prefix_route
    self.wait_online(['dummy98:routable', 'test1:routable', 'vrf99:carrier'])
  File "./test/test-network/systemd-networkd-tests.py", line 883, in wait_online
    check_output(*args, env=wait_online_env)
  File "./test/test-network/systemd-networkd-tests.py", line 98, in check_output
    return subprocess.run(command, check=True, universal_newlines=text, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs).stdout.rstrip()
  File "/usr/lib64/python3.6/subprocess.py", line 438, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/usr/lib/systemd/systemd-networkd-wait-online', '--timeout=20s', '--interface=dummy98:routable', '--interface=test1:routable', '--interface=vrf99:carrier']' returned non-zero exit status 1.
```
But, @mrc0mmand could you restart the CI?

src/udev/udev-rules.c
```diff
@@ -1313,10 +1321,29 @@ int udev_rules_load(UdevRules **ret_rules, ResolveNameTiming resolve_name_timing
...
-        return paths_check_timestamp(RULES_DIRS, &rules->dirs_ts_usec, true);
+        r = conf_files_list_strv(&files, ".rules", NULL, 0, RULES_DIRS);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to enumerate rules files: %m");
+
+        STRV_FOREACH(f, files) {
+                struct stat st;
+
+                if (stat(*f, &st) < 0)
+                        continue;
```
poettering: maybe add debug log, if any other error than ENOENT?

src/udev/udev-rules.c
```diff
@@ -1313,10 +1321,29 @@ int udev_rules_load(UdevRules **ret_rules, ResolveNameTiming resolve_name_timing
 }
 
 bool udev_rules_check_timestamp(UdevRules *rules) {
```
poettering: so i guess it is OK to make this change, but I think this is not as relevant as it might appear, since package managers and editors should change files atomically these days, i.e. create a new temporary inode, populate it, and then rename it over the old one. That way you'll always generate mtime updates on the parent dir...

Hence, in any realistic world I am pretty sure this should be a non-issue. But then again, checking this fully certainly shouldn't hurt, given that this is only done on request anyway, iiuc, i.e. not done frequently.

src/udev/udev-rules.c
```diff
@@ -1338,9 +1338,8 @@ bool udev_rules_check_timestamp(UdevRules *rules) {
                 if (stat(*f, &st) < 0)
                         continue;
 
-                r = hashmap_put_stats_by_path(&stats_by_path, *f, &st);
-                if (r < 0)
-                        return r;
+                if (hashmap_put_stats_by_path(&stats_by_path, *f, &st) < 0)
+                        return true;
```
poettering: maybe add a comment that errors are eaten up here and why that's safe. and maybe add log_debug() about unexpected erros.

src/udev/udevd.c
```diff
@@ -351,16 +351,42 @@ static void notify_ready(void) {
...
+                /* check for changed config, every 3 seconds at most */
+                return;
+        manager->last_usec = now_usec;
+
+        /* Reload SELinux label database, to make the child inherit the up-to-date database. */
+        mac_selinux_maybe_reload();
+
+        /* Nothing changed. It is not necessary to reload. */
+        if (!udev_rules_should_reload(manager->rules) && !udev_builtin_validate())
+                return;
```
poettering, line 372: So, this is probably not going to be liked by people who build reproducible OS images, and things like ostree systems. For those the mtimes carry little value, as reproducing the very same image x times successfully would mean all file contents and all metadata will match -- except of for the mtime which is different eahc time. Hence for example ostree sets the mtimes all to zero (i.e. jan 1st, 1970) in their images, because they want to share inodes between images (via hardlink farms) and they only can do that if the mtimes carry no value. rules files are often stored in /usr/ and hence have 1970 timestamps there, and stay that way even if updated by ostree.

So, I am sure that eating up explicit reload requests in ostree setups should continue to have an effect, as mtimes are useless there in /usr/ to detect changes.

to salvage that, two ideas:

1. honour the fact that ostree sets mtime to 0 for its inodes: i.e. don't suppress reloads if the dir mtime is 0. i.e. make mtime 0 magic
2. add an env var that allows overriding the logic.

src/udev/udevd.c
```diff
@@ -351,16 +351,42 @@ static void notify_ready(void) {
...
+
         udev_builtin_exit();
+        udev_builtin_init();
+
+        r = udev_rules_load(&rules, arg_resolve_name_timing);
+        if (r < 0)
+                log_warning_errno(r, "Failed to read udev rules, using the previously loaded rules, ignoring: %m");
+        else {
+                manager->rules = udev_rules_free(manager->rules);
+                manager->rules = rules;
```
poettering: shouldn't this use TAKE_PTR(), and `rules`  use some `_cleanup_` thing?

src/udev/udev-builtin-kmod.c
```diff
@@ -43,24 +43,29 @@ static int builtin_kmod_init(void) {
...
 
 /* called every couple of seconds during event activity; 'true' if config has changed */
 static bool builtin_kmod_validate(void) {
-        log_debug("Validate module index");
         if (!ctx)
                 return false;
-        return (kmod_validate_resources(ctx) != KMOD_RESOURCES_OK);
+
+        if (kmod_validate_resources(ctx) != KMOD_RESOURCES_OK) {
+                log_debug("kernel module index needs reloading.");
```
poettering: kernel → Kernel?

src/udev/udev-builtin-net_setup_link.c
```diff
@@ -75,11 +75,15 @@ static void builtin_net_setup_link_exit(void) {
 }
 
 static bool builtin_net_setup_link_validate(void) {
-        log_debug("Check if link configuration needs reloading.");
         if (!ctx)
                 return false;
 
-        return link_config_should_reload(ctx);
+        if (link_config_should_reload(ctx)) {
+                log_debug("link configuration context needs reloading.");
```
poettering: link → Link?

src/udev/udev-rules.c
yuwata: > But then again, checking this fully certainly shouldn't hurt, given that this is only done on request anyway, iiuc, i.e. not done frequently.

This is called at most once per 3 sec, unless `udevadm control --reload` is repeatedly called. Hence, this should not cause performance regression.

src/udev/udevd.c
yuwata: Interesting. But I'd like to keep the current behavior. Reasons:
- If files under /usr are updated, that means a package or os image is updated or replaced. In that case, udevd should be restarted (by e.g. rpm post script).
- I guess .rules or .link files under /etc or /run are managed by admin, instead of packager or image provider, and has non-zero mtime. Hence the changes on them can be handled by the current logic.
- The `stats_by_path` hashmap has full `struct stat`, not only mtime. So, of course that's not perfect, but when a .rules or .link file is updated, its file size is also changed in most cases, and we can detect the change even if mtime is not changed.

If the zero mtime introduces any real issues, then I'd like to address that in a later PR, as the current logic is already used by networkd (and IIRC, PID1 also use the logic partially).

yuwata:
@poettering Thank you for the review. All comments, except for the one about the zero mtime, are addressed. PTAL!

PR #23998, opened 2022-07-13
coredump: Try to write journald coredump metadata to the journal

daandemeyer:
Currently, if journald coredumps, the coredump is written to
/var/lib/systemd/coredump but the coredump metadata is not written
to  the journal meaning we can't find out about the coredump's
existence via the journal. This means that coredumpctl can't be
used to work with journald coredumps, as well as any other tools
that rely on journald to know about coredumps.

To solve the issue, let's have systemd-coredump try to write
systemd-journald coredump metadata to the journal. We have to be
careful though, since if journald coredumps, there's no active
reader on the receive end of the journal socket, so we have to make
sure we don't deadlock trying to write to the socket. To avoid the
deadlock, we put the socket in nonblocking mode before trying to
write to it.

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -29,6 +29,7 @@
 #include "fs-util.h"
 #include "io-util.h"
 #include "journal-importer.h"
+#include "journal-send.h"
 #include "log.h"
 #include "macro.h"
 #include "main-func.h"
@@ -842,12 +843,10 @@ log:
 
         core_message = strjoina(core_message, stacktrace ? "\n\n" : NULL, stacktrace);
 
-        if (context->is_journald) {
-                /* We cannot log to the journal, so just print the message.
-                 * The target was set previously to something safe. */
+        if (context->is_journald)
+                /* We might not be able to log to the journal, so let's always print the message to another
+                 * log target. The target was set previously to something safe. */
                 log_dispatch(LOG_ERR, 0, core_message);
-                return 0;
-        }
 
         (void) iovw_put_string_field(iovw, "MESSAGE=", core_message);
 
@@ -903,8 +902,29 @@ log:
                                  coredump_size, arg_journal_size_max);
         }
 
+        /* If journald is coredumping, we have to be careful that we don't deadlock when trying to write the
+         * coredump to the journal, so we put the journal socket in nonblocking mode before trying to write
+         * the coredump to the socket. */
+
+        if (context->is_journald) {
+                r = journal_fd_nonblock(true);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to make journal socket non-blocking: %m");
+        }
+
         r = sd_journal_sendv(iovw->iovec, iovw->count);
-        if (r < 0)
+
+        if (context->is_journald) {
+                int k;
+
+                k = journal_fd_nonblock(false);
+                if (k < 0)
+                        return log_error_errno(k, "Failed to make journal socket blocking: %m");
+        }
+
+        if (r == -EAGAIN && context->is_journald)
+                log_warning_errno(r, "Failed to log journal coredump, ignoring: %m");
+        else if (r < 0)
                 return log_error_errno(r, "Failed to log coredump: %m");
 
         return 0;
--- a/src/libsystemd/sd-journal/journal-send.c
+++ b/src/libsystemd/sd-journal/journal-send.c
@@ -66,6 +66,16 @@ retry:
         return fd;
 }
 
+int journal_fd_nonblock(bool nonblock) {
+        int r;
+
+        r = journal_fd();
+        if (r < 0)
+                return r;
+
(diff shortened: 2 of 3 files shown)


src/libsystemd/sd-journal/journal-send.c
```diff
@@ -318,7 +318,7 @@ _public_ int sd_journal_sendv(const struct iovec *iov, int n) {
         if (errno == ENOENT)
                 return 0;
 
-        if (!IN_SET(errno, EMSGSIZE, ENOBUFS))
+        if (!IN_SET(errno, EMSGSIZE, ENOBUFS, EAGAIN))
```
bluca, line 331: won't this have an effect on our public API?

daandemeyer: EAGAIN can only be returned if the socket is nonblocking, and there's no public API to make the socket nonblocking

vcaputo:
The last time I tried using `Storage=journal` in coredump.conf, my test coredump immediately purged all archived journals in a hopeless attempt to make space and it still failed to store the coredump.

I ended up immediately going back to the default of external in anger, and there's still an entry in my todo list surrounding this issue to see if journal storage of coredumps could be better informed about coredump size and estimating post-vacuum space trying to make room.

If this change, which I admit I haven't reviewed at all - only commenting based on the description, has the ability to trigger the same behavior, I'm totally against it.  Suddenly losing all your journals from previous boots because of a coredump can be very unexpected and annoying, especially when you don't even have the coredump you threw it all away for.

src/coredump/coredump.c
```diff
@@ -903,10 +902,32 @@ static int submit_coredump(
                                  coredump_size, arg_journal_size_max);
         }
 
+        /* If journald is coredumping, we have to be careful that we don't deadlock when trying to write the
+         * coredump to the journal, so we put the journal socket in nonblocking mode before trying to write
+         * the coredump to the socket. */
+
+        if (context->is_journald) {
+                jfd = journal_fd();
```
bluca: what about having a helper to change nonblock mode, instead of a helper to get the journal fd?

daandemeyer: I thought about that as well, I wonder if we should make that public API or not though

bluca: I'd say keep it private for now

bluca (approved):

bluca (approved):

src/coredump/coredump.c
```diff
@@ -903,10 +902,28 @@ static int submit_coredump(
...
+                if (r < 0)
+                        return log_error_errno(r, "Failed to make journal socket non-blocking: %m");
+        }
+
         r = sd_journal_sendv(iovw->iovec, iovw->count);
-        if (r < 0)
+        if (r == -EAGAIN && context->is_journald)
+                log_warning_errno(r, "Failed to log journal coredump, ignoring: %m");
+        else if (r < 0)
                 return log_error_errno(r, "Failed to log coredump: %m");
```
yuwata, line 928: On failure, is it not necessary to call `journal_fd_nonblock(false)`?

src/libsystemd/sd-journal/journal-send.c
```diff
@@ -66,6 +66,20 @@ static int journal_fd(void) {
...
+
+        r = journal_fd();
+        if (r < 0)
+                return r;
+
+        r = fd_nonblock(r, nonblock);
+        if (r < 0)
+                return r;
+
+        return 0;
```
yuwata: Not a matter, but you can shorten this by `return fd_nonblock(...);`

src/libsystemd/sd-journal/journal-send.c
```diff
@@ -318,7 +332,7 @@ _public_ int sd_journal_sendv(const struct iovec *iov, int n) {
         if (errno == ENOENT)
                 return 0;
 
-        if (!IN_SET(errno, EMSGSIZE, ENOBUFS))
+        if (!IN_SET(errno, EMSGSIZE, ENOBUFS, EAGAIN))
```
yuwata, line 331: Is it not necessary to handle `EAGAIN` from `send_one_fd_sa()` below?

If `EAGAIN` from the two `sendmsg()` are ignored, then it is not necessary to handle the error in `submit_coredump()`.

yuwata (commented):
Several questions.

src/libsystemd/sd-journal/journal-send.c
daandemeyer: We handle it here so that we go to the fallback code which uses an fd which might fit in the socket send buffer. If that fails, I think we want to fail with EAGAIN to inform the caller that we were unable to send the log to the journal. I don't think we should swallow the error in that case.

yuwata: Ah, I see. Thank you for the explanation.

PR #23999, opened 2022-07-13
Revert of recent changes in handling of background sessions

msekletar:


test/units/testsuite-35.sh
```diff
@@ -420,48 +420,6 @@ EOF
     fi
```
yuwata, line 430: Could you also drop crond from TEST-35/test.sh ?

msekletar: Fixed.

msekletar:
Rebased. @yuwata PTAL.

PR #24005, opened 2022-07-14
pid1: import credentials also from SMBIOS data

poettering:
Apparently this is the preferred way to pass stuff into VMs these days, as it is generic (unlike `fw_cfg` which is qemu specific) and faster.

### DIFF
--- a/TODO
+++ b/TODO
@@ -150,9 +150,6 @@ Features:
 * pam_systemd: on interactive logins, maybe show SUPPORT_END information at
   login time, á la motd
 
-* similar to the existing fw_cfg support, also support reading system
-  credentials from DMI vendor fields, as supported by qemu
-
 * sd-boot: instead of uncondtionally deriving the ESP to search boot loader
   spec entries in from the paths of sd-boot binary, let's optionally allow it
   to be configured on sd-boot cmdline + efi var. Usecase: embedd sd-boot in the


src/core/import-creds.c
```diff
@@ -446,26 +453,174 @@ static int import_credentials_qemu(ImportCredentialContext *c) {
...
+                dmi_field_header = data;
+                if (dmi_field_header->type != 11 ||
+                    dmi_field_header->length != 5)
+                        return log_error_errno(SYNTHETIC_ERRNO(EBADMSG), "Invalid DMI field header.");
+
+                r = parse_smbios_strings(c, dmi_field_header->contents, size - offsetof(struct dmi_field_header, contents));
+                if (r < 0)
+                        return r;
+
+                if (i == UINT_MAX) /* Prevent overflow */
```
medhefgo, line 599: Why not just `i < UINT_MAX` in the for condition to keep it compact?

poettering: well, because I wanted to count through the whole unsigned integer range. If you do as you suggest you miss one iteration. I mean, it's not that this matters, it's very unlikely there will be that many instances of this smbios object type, but I just couldn't stop myself from covering the range completely anyway. maybe I have OCD? ;-)

medhefgo: Maybe you should just limit this to UINT8_MAX or so. UINT_MAX sounds excessive.

And now you made me curious about whether there is a limit to how many DMI entries can possibly exist. *sigh*

poettering: Actually, the only reason we need to bound the loop is because we run out of bits in the counter. The loop will exit anyway once there is no object instance anymore by the current counter, which should be very quick.

Or to say this differently: if firmware/sysfs really exposes > 255 objects of the same type, then I doubt it's us in userspace who should catch that...

src/core/import-creds.c
```diff
@@ -446,26 +453,174 @@ static int import_credentials_qemu(ImportCredentialContext *c) {
...
 
+static int parse_smbios_strings(ImportCredentialContext *c, const char *data, size_t size) {
+        size_t left, skip;
+        const char *p;
+        int r;
+
+        assert(c);
+        assert(data || size == 0);
+
+        /* Unpacks a packed series of SMBIOS OEM vendor strings. These are a series of NUL termianted strings, one after the other. */
```
bluca: /termianted/terminated/

src/core/import-creds.c
```diff
@@ -446,26 +453,177 @@ static int import_credentials_qemu(ImportCredentialContext *c) {
...
+                        log_full_errno(r == -ENOENT ? LOG_DEBUG : LOG_WARNING, r, "Failed to open '%p', ignoring: %m", p);
+                        break;
+                }
+
+                if (size < offsetof(struct dmi_field_header, contents))
+                        return log_error_errno(SYNTHETIC_ERRNO(EBADMSG), "DMI field header of '%p' too short.", p);
+
+                dmi_field_header = data;
+                if (dmi_field_header->type != 11 ||
+                    dmi_field_header->length != 5)
```
yuwata: I think `offsetof(struct dmi_field_header, contents)` is better than 5.

docs/CREDENTIALS.md
```diff
@@ -50,10 +50,11 @@ purpose. Specifically, the following features are provided:
 6. Service credentials are placed in non-swappable memory. (If permissions
```
yuwata, line 50: If you will fix the above point,
Commit msg: siilar -> similar

yuwata (changes requested):

yuwata (approved):

yuwata:
(force-pushed to trigger CIs.)

PR #24006, opened 2022-07-14
growfs: don't actually resize on dry-run

Undef-a:
This causes systemd-growfs to exit before resizing the partition when
`--dry-run` is passed. Resizing during a dry run of a change breaks the
users expectations.

This fixes the following situation:
```
root@mobian:/home/mobian# /usr/lib/systemd/systemd-growfs --dry-run /
[  420.134647] EXT4-fs (dm-0): resizing filesystem from 847872 to 7368187 blocks
[  421.038260] EXT4-fs (dm-0): resized filesystem to 7368187
Successfully resized "/" to 28.1G bytes.
root@mobian:/home/mobian# df -h
Filesystem                   Size  Used Avail Use% Mounted on
udev                         1.3G     0  1.3G   0% /dev
tmpfs                        300M  1.6M  299M   1% /run
/dev/mapper/calamares_crypt   28G  2.7G   24G  11% /
tmpfs                        1.5G  348K  1.5G   1% /dev/shm
tmpfs                        5.0M   12K  5.0M   1% /run/lock
/dev/mmcblk0p1               731M   89M  589M  14% /boot
tmpfs                        300M   52K  300M   1% /run/user/1000
```

### DIFF
--- a/src/partition/growfs.c
+++ b/src/partition/growfs.c
@@ -241,6 +241,10 @@ static int run(int argc, char *argv[]) {
                 return log_error_errno(errno, "Failed to query size of \"%s\": %m", devpath);
 
         log_debug("Resizing \"%s\" to %"PRIu64" bytes...", arg_target, size);
+
+        if (arg_dry_run)
+                return 0;
+
         r = resize_fs(mountfd, size, &newsize);
         if (r < 0)
                 return log_error_errno(r, "Failed to resize \"%s\" to %"PRIu64" bytes: %m",


poettering:
Ouch. This was apparently broken in d6f1e66076929f071830334800fc380bd4954806. Thanks for the fix.

src/partition/growfs.c
```diff
@@ -241,6 +241,10 @@ static int run(int argc, char *argv[]) {
                 return log_error_errno(errno, "Failed to query size of \"%s\": %m", devpath);
 
         log_debug("Resizing \"%s\" to %"PRIu64" bytes...", arg_target, size);
+
+        if (arg_dry_run)
+                return 0;
```
medhefgo, line 246: I'm curious: do we do dry runs silently or log about things being skipped because of it?

Undef-a: On https://github.com/systemd/systemd/blob/2af239c20e024a77b800d9785643f2a0fe8c8cc9/src/partition/growfs.c#L71 it also silently continues, but it does make more sense to return a "stopping due to dry-run" message.

medhefgo (approved):

PR #24007, opened 2022-07-14
base-filesystem: pick more conservative access mode for /root/

poettering:
Let's not allow anyone to look into /root/ if we create it via the
base-filesystem logic. i.e. change 0755 → 0750 as default access mode
for /root/, in case we create it if it happens to be missing.

### DIFF
--- a/src/shared/base-filesystem.c
+++ b/src/shared/base-filesystem.c
@@ -31,7 +31,7 @@ typedef struct BaseFilesystem {
 static const BaseFilesystem table[] = {
         { "bin",      0, "usr/bin\0",                  NULL },
         { "lib",      0, "usr/lib\0",                  NULL },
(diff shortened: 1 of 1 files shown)


medhefgo (approved):

mbiebl:
typo: s/now/not/

poettering:
fixed the typo (usually we don't care too much about typos in commit msgs btw, unless they are grossly misleading).

bluca (approved):

bluca:
> fixed the typo (usually we don't care too much about typos in commit msgs btw, unless they are grossly misleading).

this one did make me double back, as I was like "wut why do we want everyone to be able to read /root now?" :-D

yuwata (approved):

PR #24010, opened 2022-07-14
tmpfiles: optionally allow configuring file contents for f/w lines to be encoded in base64

poettering:


src/tmpfiles/tmpfiles.c
```diff
@@ -1329,14 +1341,29 @@ static int path_set_attribute(Item *item, const char *path) {
...
+static int write_argument_data(Item *i, int fd, const char *path) {
+        int r;
+
+        assert(i);
+        assert(fd >= 0);
+        assert(path);
+
+        log_debug("Writing to \"%s\".", path);
+
+        r = loop_write(fd, item_binary_argument(i), item_binary_argument_size(i), /* do_poll= */ false);
```
bluca, line 1358: Before this change there was an assert on i->argument so NULL was never passed here, now it can happen. Should there be an assert on the return of `item_binary_argument`()?

src/tmpfiles/tmpfiles.c
```diff
@@ -461,6 +464,15 @@ static bool unix_socket_alive(const char *fn) {
         return set_contains(unix_sockets, fn);
 }
 
+/* Accessors for the argument in binary format */
+static const void* item_binary_argument(const Item *i) {
+        return i->binary_argument ?: i->argument;
```
bluca, line 470: `assert(i)` for peace of mind

src/tmpfiles/tmpfiles.c
bluca: (specifically referring to the calling path from write_one_file)

bluca:
`TEST-22-TMPFILES` is failing which seems related

man/tmpfiles.d.xml
```diff
@@ -489,6 +490,13 @@ w- /proc/sys/vm/swappiness - - - - 10</programlisting></para>
       be either directories or directory symlinks). For example, if there is a FIFO in place of one of the parent path
       components it will be replaced with a directory.</para>
 
+      <para>If the tilde character (<literal>~</literal>) is used the argument (i.e. 6th) column is <ulink
```
yuwata: I'd like to add a comma: "is used, the argument column is ..."

src/tmpfiles/tmpfiles.c
```diff
@@ -2687,7 +2703,8 @@ static bool item_compatible(const Item *a, const Item *b) {
 
         if (takes_ownership(a->type) && takes_ownership(b->type))
                 /* check if the items are the same */
-                return  streq_ptr(a->argument, b->argument) &&
+                return memcmp_nn(item_binary_argument(a), item_binary_argument_size(a),
+                                 item_binary_argument(b), item_binary_argument_size(b)) &&
```
yuwata: missing `== 0`.

yuwata (changes requested):

poettering:
Addressed all issues raised. And hopefully CI should pass now too. PTAL!

PR #24011, opened 2022-07-14
pid1: add new condition type "ConditionCredential=" for checking for credentials passed into the system

poettering:
This is mirroring ConditionKernelCommandLine= to some point, and should be quite useful for services that provision local functionality based on credentials that might or might not be passed into the system

### DIFF
--- a/man/systemd.link.xml
+++ b/man/systemd.link.xml
@@ -228,7 +228,7 @@
           <para>Matches against the hostname or machine ID of the host. See <varname>ConditionHost=</varname> in
           <citerefentry><refentrytitle>systemd.unit</refentrytitle><manvolnum>5</manvolnum></citerefentry>
           for details. When prefixed with an exclamation mark (<literal>!</literal>), the result is negated.
-          If an empty string is assigned, then previously assigned value is cleared.
+          If an empty string is assigned, the previously assigned value is cleared.
           </para>
         </listitem>
       </varlistentry>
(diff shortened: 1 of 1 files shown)


man/systemd.link.xml
```diff
@@ -269,6 +269,18 @@
...
       </varlistentry>
 
+      <varlistentry id='credential'>
+        <term><varname>Credential=</varname></term>
+        <listitem>
+          <para>Checks whether the specified credential was passed to the
+          <filename>systemd-networkd.service</filename> service. See <ulink
+          url="https://systemd.io/CREDENTIALS">System and Service Credentials</ulink> for details. When
+          prefixed with an exclamation mark (<literal>!</literal>), the result is negated.  If an empty
+          string is assigned, then previously assigned value is cleared.
```
bluca: "the previously assigned value"

bluca (approved):

src/shared/condition.c
```diff
@@ -140,6 +141,46 @@ static int condition_test_kernel_command_line(Condition *c, char **env) {
...
+
+                j = path_join(cd, c->parameter);
+                if (!j)
+                        return -ENOMEM;
+
+                if (laccess(j, F_OK) < 0) {
+                        if (errno == ENOENT) /* not found in this dir */
+                                continue;
+                        return -errno;
+                } else
```
yuwata: Unnecessary 'else'.

PR #24013, opened 2022-07-14
growfs: Expand FS even if underlying block expansion fails

Undef-a:
This allows growfs to expand the filesystem even when the underlying
block device cannot be expanded. This has been useful for example on
LUKS devices that have already been expanded using systemd-repart.

This works around the following error:
```
root@mobian:/home/mobian# /usr/lib/systemd/systemd-growfs /
crypt_resize() of /dev/block/179:2 failed: Operation not permitted
```

Tested on device successfully:
```
root@mobian:/mnt# LD_PRELOAD=/mnt/libsystemd-shared-251.so ./systemd-growfs /
crypt_resize() of /dev/block/179:2 failed: Operation not permitted
Unable to resize underlying device of "/", continuing on: Operation not permitted
Successfully resized "/" to 28.1G bytes.
```

I'm not certain this work around is the best solution to this particular problem, but noticed after finding https://github.com/systemd/systemd/pull/24006 that it does successfully solve the problem.

### DIFF
--- a/src/partition/growfs.c
+++ b/src/partition/growfs.c
@@ -222,8 +222,9 @@ static int run(int argc, char *argv[]) {
                 return log_error_errno(SYNTHETIC_ERRNO(ENODEV), "File system \"%s\" not backed by block device.", arg_target);
 
         r = maybe_resize_underlying_device(arg_target, devno);
+        /* This function can return -EPERM when growfs would otherwise succeed */
         if (r < 0)
-                return r;
+                log_warning_errno(r, "Unable to resize underlying device of \"%s\", proceeding anyway: %m", arg_target);
 
         mountfd = open(arg_target, O_RDONLY|O_CLOEXEC);
         if (mountfd < 0)


src/partition/growfs.c
```diff
@@ -222,7 +222,10 @@ static int run(int argc, char *argv[]) {
                 return log_error_errno(SYNTHETIC_ERRNO(ENODEV), "File system \"%s\" not backed by block device.", arg_target);
 
         r = maybe_resize_underlying_device(arg_target, devno);
-        if (r < 0)
+        /* This function can return -EPERM when growfs would otherwise succeed */
+        if (r == -EPERM)
+            log_warning_errno(r, "Unable to resize underlying device of \"%s\", continuing on: %m", arg_target);
+        else if (r < 0)
                 return r;
```
poettering: so I think we actually should continue in all cases here, not just EPERM.

we usually use the wording ", proceeding anyway" rather than ", continuing on"...

also, please follow coding style, i.e. indent by 8 spaces.

Undef-a: Thanks for the review. Those couple of things should be fixed now.

PR #24016, opened 2022-07-14
sysctl: also process sysctl requests via the "sysctl.extra" credential

poettering:


src/sysctl/sysctl.c
```diff
@@ -277,6 +278,25 @@ static int parse_file(OrderedHashmap **sysctl_options, const char *path, bool ig
         return r;
 }
 
+static int read_credential_lines(OrderedHashmap **sysctl_options) {
+        _cleanup_free_ char *j;
```
yuwata: missing `= NULL`.

yuwata (approved):
LGTM. But one point.

poettering:
added the "= NULL". Upgrading green label.

PR #24018, opened 2022-07-14
Make generators easier to test, fix various corner issues

keszybz:
I started working on the units for the initrd. The patches here are general cleanups that were generated on the way. I'll submit the "real" changes later on.

/cc @lnykryn

### DIFF
--- a/src/fstab-generator/fstab-generator.c
+++ b/src/fstab-generator/fstab-generator.c
@@ -631,13 +631,25 @@ static int parse_fstab(bool initrd) {
                          * mount units, but causes problems since it historically worked to have symlinks in e.g.
                          * /etc/fstab. So we canonicalize here. Note that we use CHASE_NONEXISTENT to handle the case
                          * where a symlink refers to another mount target; this works assuming the sub-mountpoint
-                         * target is the final directory. */
+                         * target is the final directory.
+                         *
+                         * FIXME: when chase_symlinks() learns to chase non-existent paths, use this here and
+                         *        drop the prefixing with /sysroot on error below.
+                         */
                         k = chase_symlinks(where, initrd ? "/sysroot" : NULL,
                                            CHASE_PREFIX_ROOT | CHASE_NONEXISTENT,
                                            &canonical_where, NULL);
-                        if (k < 0) /* If we can't canonicalize, continue as if it wasn't a symlink */
-                                log_debug_errno(k, "Failed to read symlink target for %s, ignoring: %m", where);
-                        else if (streq(canonical_where, where)) /* If it was fully canonicalized, suppress the change */
+                        if (k < 0) {
+                                /* If we can't canonicalize, continue as if it wasn't a symlink */
+                                log_debug_errno(k, "Failed to read symlink target for %s, using as-is: %m", where);
+
+                                if (initrd) {
+                                        canonical_where = path_join("/sysroot", where);
+                                        if (!canonical_where)
+                                                return log_oom();
+                                }
+
+                        } else if (streq(canonical_where, where)) /* If it was fully canonicalized, suppress the change */
                                 canonical_where = mfree(canonical_where);
                         else
                                 log_debug("Canonicalized what=%s where=%s to %s", what, where, canonical_where);


src/hibernate-resume/hibernate-resume-generator.c
```diff
@@ -109,17 +109,10 @@ static int process_resume(void) {
...
 
-        log_setup_generator();
-
-        if (argc > 1 && argc != 4)
-                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
-                                       "This program takes three or no arguments.");
-
-        if (argc > 1)
-                arg_dest = argv[1];
+        arg_dest = ASSERT_PTR(dest);
```
bluca, line 115: isn't this a change in behaviour? there was a default before, now there isn't - at least the commit message should explain the change?

src/boot/bless-boot-generator.c
```diff
@@ -10,62 +9,47 @@
...
 #include "virt.h"
 
-/* This generator pulls systemd-bless-boot.service into the initial transaction if the "LoaderBootCountPath" EFI
- * variable is set, i.e. the system boots up with boot counting in effect, which means we should mark the boot as
- * "good" if we manage to boot up far enough. */
+/* This generator pulls systemd-bless-boot.service into the initial transaction if the "LoaderBootCountPath"
+ * EFI variable is set, i.e. the system boots up with boot counting in effect, which means we should mark the
+ * boot as "good" if we manage to boot up far enough. */
 
-static const char *arg_dest = "/tmp";
```
bluca, line 20: same here

src/shared/generator.c
```diff
@@ -736,12 +736,16 @@ int generator_write_veritysetup_service_section(
         return 0;
 }
 
-void log_setup_generator(void) {
-        /* Disable talking to syslog/journal (i.e. the two IPC-based loggers) if we run in system context. */
-        if (cg_pid_get_owner_uid(0, NULL) == -ENXIO /* not running in a per-user slice */)
-                log_set_prohibit_ipc(true);
+void log_setup_generator(bool normal_invocation) {
```
bluca: 'normal' is a bit vague for a reader, maybe 'early_boot_invocation'?

bluca (approved):

src/hibernate-resume/hibernate-resume-generator.c
bluca: I guess there's a subsequent commit that explains, so looking it in order it wasn't clear at first

bluca: maybe postpone those changes to the later commit?

yuwata: `DEFINE_MAIN_GENERATOR_FUNCTION()` always sets `dest`. Hence the change in this commit looks OK for me.

src/boot/bless-boot-generator.c
yuwata: `DEFINE_MAIN_GENERATOR_FUNCTION()` always sets the destination. Hence the change is OK.

src/shared/generator.c
```diff
@@ -736,12 +736,16 @@ int generator_write_veritysetup_service_section(
...
+        if (normal_invocation) {
```
yuwata: It may be overkill, but how about to use the similar logic as `should_parse_proc_cmdline()` in log.c, instead of `argc == 4`?

yuwata (commented):
LGTM. But one suggestion.

src/hibernate-resume/hibernate-resume-generator.c
keszybz: @bluca I think you might be confused by github's reordering of commits. The commit is one of the first ones, and as @yuwata wrote, `DEFINE_MAIN_GENERATOR_FUNCTION()` always sets `dest`.

src/shared/generator.c
keszybz: The problem is that it's not just "early boot". It's also invoked this way during every deamon-reload…

bluca: `called_by_systemd` or something? For someone who comes across this files, 'normal' is going to be pretty confusing

src/hibernate-resume/hibernate-resume-generator.c
bluca: it is entirely possible, yes

src/shared/generator.c
poettering: maybe invert the flag, i.e. "manual_invocation" or so.

poettering: hm, can we remove this entirely, maybe, and just operate based on on_tty()? 

I think that'd be much better given that binding this to the number of arguments kinda sucks, if i want to test manually what gets generated into which dir...

src/fstab-generator/fstab-generator.c
```diff
@@ -635,9 +635,17 @@ static int parse_fstab(bool initrd) {
...
+                                /* If we can't canonicalize, continue as if it wasn't a symlink */
+                                log_debug_errno(k, "Failed to read symlink target for %s, using as-is: %m", where);
+
+                                if (initrd) {
+                                        canonical_where = path_join("/sysroot", where);
+                                        if (!canonical_where)
+                                                return log_oom();
+                                }
+
+                        } else if (streq(canonical_where, where)) /* If it was fully canonicalized, suppress the change */
```
poettering, line 652: hmm, so chase_symlinks() her already has CHASE_NONEXISTENT set, so the example you gave in the commit msg, since the case /foo/ should already be covered. 

CHASE_NONEXISTENT only allows the basename of the input path to be missing. Maybe we should add CHASE_NONEXISTENT_MANY which allows any suffix to be missing, including one consisting of a multi-component path.

the fix to chase_symlinks() should be pretty easy to allow that, and the overall result should be much nicer

poettering: (CHASE_NONEXISTANT should probably be renamed CHASE_NONEXISTANT_LAST or so)

src/hibernate-resume/hibernate-resume-generator.c
keszybz: I added a comment in the commit message.

src/boot/bless-boot-generator.c
keszybz: Also added a comment in the commit message.

src/fstab-generator/fstab-generator.c
keszybz: Yeah, that's an option. But I think we can merge those patches as they are, in particular this has the advantage that they are easy to backport. We can always simplify the code if somebody makes `chase_symlinks()` more powerful.

keszybz: Yeah, that's an option. But I think we can merge those patches as they are, in particular this has the advantage that they are easy to backport. We can always simplify the code if somebody makes `chase_symlinks()` more powerful.

bluca (approved):

yuwata (approved):
LGTM.

src/fstab-generator/fstab-generator.c
poettering: could you at least add a FIXME comment here, so that one can find this again later?

poettering:
what about the on_tty() thing I suggested?

src/basic/process-util.c
```diff
@@ -1579,6 +1579,27 @@ bool invoked_as(char *argv[], const char *token) {
...
+                return false;
+
+        if (streq(e, "*"))
+                /* For testing. */
+                return true;
+
+        pid_t p;
+        if (parse_pid(e, &p) < 0) {
+                /* We know that systemd sets the variable correctly. Something else must have set it. */
+                log_debug("Failed to parse \"SYSTEMD_EXEC_PID=%s\", ignoring.", e);
```
poettering: log_debug_errno(r, …);

i.e. show the precise error code parse_pid() returns?

keszybz:
> what about the on_tty() thing I suggested?

I implemented Yu's suggestion instead: the new `invoked_by_systemd()` function is used.

PR #24020, opened 2022-07-14
network: fix wlan interface handling when its RF is killed

### LINKED ISSUE #23649: DHCPv4 address dropped because the interface was once rfkill blocked
 > 251.2

**Used distribution**
 > Arch

 > `Linux ideapad 5.18.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Mon, 30 May 2022 17:53:11 +0000 x86_64 GNU/Linux`

 > x86_64

 > DHCPv4 address stays regardless of whether networkd has once failed to configure the interface due to rfkill block

 > DHCPv4 address get silently dropped because networkd has once failed to configure the interface due to rfkill block

 The network file for my `wlan0` is as simple as:
```
[Match]
Name=wlan0
[Network]
DHCP=ipv4
DNSSEC=no
```
After upgrading systemd from 250.5 to 251.2, I notice that `wlan0` no longer has DHCPv4 address assigned after boot. However when I check log:
```
Jun 07 08:21:50 ideapad systemd[1]: Starting Network Configuration...
Jun 07 08:21:51 ideapad systemd-networkd[460]: temp: netdev ready
Jun 07 08:21:51 ideapad systemd-networkd[460]: dhcp0: netdev ready
Jun 07 08:21:51 ideapad systemd-networkd[460]: bridge0: netdev ready
Jun 07 08:21:51 ideapad systemd-networkd[460]: lo: Link UP
Jun 07 08:21:51 ideapad systemd-networkd[460]: lo: Gained carrier
Jun 07 08:21:51 ideapad systemd-networkd[460]: Enumeration completed
Jun 07 08:21:51 ideapad systemd[1]: Started Network Configuration.
Jun 07 08:21:51 ideapad systemd-networkd[460]: dhcp0_tap: Configuring with /etc/systemd/network/dhcp0_tap.network.
Jun 07 08:21:51 ideapad systemd-networkd[460]: bridge0: Configuring with /etc/systemd/network/bridge0.network.
Jun 07 08:21:51 ideapad systemd-networkd[460]: bridge0: Link UP
Jun 07 08:21:51 ideapad systemd-networkd[460]: dhcp0_tap: Link UP
Jun 07 08:21:51 ideapad systemd-networkd[460]: temp_tap: Configuring with /etc/systemd/network/temp_tap.network.
Jun 07 08:21:51 ideapad systemd-networkd[460]: temp_tap: Link UP
Jun 07 08:21:51 ideapad systemd-networkd[460]: wlan0: found matching network '/etc/systemd/network/wlan0.network', based on potentially unpredictable interface name.
Jun 07 08:21:51 ideapad systemd-networkd[460]: wlan0: Configuring with /etc/systemd/network/wlan0.network.
Jun 07 08:21:51 ideapad systemd-networkd[460]: wlan0: Could not bring up interface: Operation not possible due to RF-kill
Jun 07 08:21:51 ideapad systemd-networkd[460]: dhcp0_tap: Gained carrier
Jun 07 08:21:51 ideapad systemd-networkd[460]: bridge0: Gained carrier
Jun 07 08:21:52 ideapad systemd-networkd[460]: wlan0: Link UP
Jun 07 08:21:52 ideapad systemd-networkd[460]: wlan0: Link DOWN
Jun 07 08:21:52 ideapad systemd-networkd[460]: bridge0: Gained IPv6LL
Jun 07 08:21:52 ideapad systemd-networkd[460]: wlan0: Link UP
Jun 07 08:21:53 ideapad systemd-networkd[460]: wlan0: Connected WiFi access point: 0yPgDo0B8yo6qz6z (c0:56:27:ca:21:9e)
Jun 07 08:21:53 ideapad systemd-networkd[460]: wlan0: Gained carrier
Jun 07 08:21:53 ideapad systemd-networkd[460]: wlan0: found matching network '/etc/systemd/network/wlan0.network', based on potentially unpredictable interface name.
Jun 07 08:21:53 ideapad systemd-networkd[460]: wlan0: DHCPv4 address 192.168.1.150/24, gateway 192.168.1.1 acquired from 192.168.1.1
Jun 07 08:21:54 ideapad systemd-networkd[460]: wlan0: Gained IPv6LL
```
It shows that the DHCPv4 client in networkd has worked. And after a restart of networkd, the DHCPv4 address is back:
```
Jun 07 08:22:09 ideapad systemd-networkd[460]: dhcp0_tap: Lost carrier
Jun 07 08:22:09 ideapad systemd-networkd[460]: bridge0: DHCPv6 lease lost
Jun 07 08:22:09 ideapad systemd[1]: Stopping Network Configuration...
Jun 07 08:22:09 ideapad systemd-networkd[460]: wlan0: DHCPv6 lease lost
Jun 07 08:22:09 ideapad systemd[1]: systemd-networkd.service: Deactivated successfully.
Jun 07 08:22:09 ideapad systemd[1]: Stopped Network Configuration.
Jun 07 08:22:09 ideapad systemd[1]: Starting Network Configuration...
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: Link UP
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: Gained carrier
Jun 07 08:22:09 ideapad systemd-networkd[844]: temp: netdev ready
Jun 07 08:22:09 ideapad systemd-networkd[844]: temp_tap: Link UP
Jun 07 08:22:09 ideapad systemd-networkd[844]: dhcp0: netdev ready
Jun 07 08:22:09 ideapad systemd-networkd[844]: dhcp0_tap: Link UP
Jun 07 08:22:09 ideapad systemd-networkd[844]: bridge0: netdev ready
Jun 07 08:22:09 ideapad systemd-networkd[844]: bridge0: Link UP
Jun 07 08:22:09 ideapad systemd-networkd[844]: bridge0: Gained carrier
Jun 07 08:22:09 ideapad systemd-networkd[844]: lo: Link UP
Jun 07 08:22:09 ideapad systemd-networkd[844]: lo: Gained carrier
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: Gained IPv6LL
Jun 07 08:22:09 ideapad systemd-networkd[844]: bridge0: Gained IPv6LL
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: Connected WiFi access point: 0yPgDo0B8yo6qz6z (c0:56:27:ca:21:9e)
Jun 07 08:22:09 ideapad systemd-networkd[844]: Enumeration completed
Jun 07 08:22:09 ideapad systemd[1]: Started Network Configuration.
Jun 07 08:22:09 ideapad systemd-networkd[844]: bridge0: netdev exists, using existing without changing its parameters
Jun 07 08:22:09 ideapad systemd-networkd[844]: dhcp0: netdev exists, using existing without changing its parameters
Jun 07 08:22:09 ideapad systemd-networkd[844]: temp: netdev exists, using existing without changing its parameters
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: found matching network '/etc/systemd/network/wlan0.network', based on potentially unpredictable interface name.
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: Configuring with /etc/systemd/network/wlan0.network.
Jun 07 08:22:09 ideapad systemd-networkd[844]: temp_tap: Configuring with /etc/systemd/network/temp_tap.network.
Jun 07 08:22:09 ideapad systemd-networkd[844]: dhcp0_tap: Configuring with /etc/systemd/network/dhcp0_tap.network.
Jun 07 08:22:09 ideapad systemd-networkd[844]: bridge0: Configuring with /etc/systemd/network/bridge0.network.
Jun 07 08:22:09 ideapad systemd-networkd[844]: wlan0: DHCPv4 address 192.168.1.150/24, gateway 192.168.1.1 acquired from 192.168.1.1
Jun 07 08:22:09 ideapad systemd-networkd[844]: dhcp0_tap: Gained carrier
```
~My hunch told me that it probably has something to do with the code in networkd that deals with IPv6LL, and  it was right: if I add `LinkLocalAddressing=no` to the `.network` for `wlan0`, the DHCPv4 address will not dropped after boot.~

~I can't tell whether the bug is more "racy" in nature or if it really only affects DHCPv4 address, but it doesn't seem to affect the static address assigned on e.g. `bridge0`.~

Actually it seems to *instead* have something to do with the fact that `wlan0` *was once* (note *was once*) "not operable" because it *was* rfkill-blocked. For now I have to disable systemd-networkd and pulls it with my rfkill-unblocking service (which is in turned pulled by `wlan0`):
```
[Unit]
Wants=iwd.service systemd-networkd.service
Before=iwd.service systemd-networkd.service

[Service]
Type=oneshot
ExecStart=/usr/bin/rfkill unblock wlan

[Install]
WantedBy=sys-subsystem-net-devices-wlan0.device
```

The thing is in my experience `rfkill` itself isn't really "synchronous". Nevertheless, this is obviously a regression in networkd anyway.

P.S. I'm almost 100% certain that it is due to the rfkill block *history*, since the issue does not show up on wired interface or wireless interface on my other laptop that does not have the interface rfkill-blocked on boot.

yuwata:
Fixes #23649.

### DIFF
--- a/src/network/networkd-setlink.c
+++ b/src/network/networkd-setlink.c
@@ -14,6 +14,7 @@
 #include "networkd-manager.h"
 #include "networkd-queue.h"
 #include "networkd-setlink.h"
+#include "networkd-wiphy.h"
 
 static int get_link_default_handler(sd_netlink *rtnl, sd_netlink_message *m, Link *link) {
         return link_getlink_handler_internal(rtnl, m, link, "Failed to sync link information");
@@ -998,13 +999,16 @@ static int link_up_or_down(Link *link, bool up, Request *req) {
         return request_call_netlink_async(link->manager->rtnl, m, req);
 }
 
-static bool link_is_ready_to_activate(Link *link) {
+static bool link_is_ready_to_activate(Link *link, bool up) {
         assert(link);
 
         if (!IN_SET(link->state, LINK_STATE_CONFIGURING, LINK_STATE_CONFIGURED))
                 return false;
 
         if (link->set_link_messages > 0)
+                return false;
+
+        if (up && link_rfkilled(link) > 0)
                 return false;
 
         return true;
@@ -1017,7 +1021,7 @@ static int link_process_activation(Request *req, Link *link, void *userdata) {
         assert(req);
         assert(link);
 
-        if (!link_is_ready_to_activate(link))
+        if (!link_is_ready_to_activate(link, up))
                 return 0;
 
         r = link_up_or_down(link, up, req);
@@ -1095,6 +1099,9 @@ static bool link_is_ready_to_bring_up_or_down(Link *link, bool up) {
                 return false;
 
         if (!link->activated)
+                return false;
+
+        if (up && link_rfkilled(link) > 0)
                 return false;
 
         return true;
--- a/src/network/networkd-wiphy.c
+++ b/src/network/networkd-wiphy.c
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
+#include <net/if_arp.h>
 #include <linux/nl80211.h>
 
 #include "device-private.h"
@@ -101,6 +102,129 @@ int wiphy_get_by_name(Manager *manager, const char *name, Wiphy **ret) {
                 *ret = w;
 
         return 0;
+}
+
+static int link_get_wiphy(Link *link, Wiphy **ret) {
+        _cleanup_(sd_device_unrefp) sd_device *phy = NULL;
+        const char *s;
+        int r;
+
+        assert(link);
+        assert(link->manager);
+
+        if (link->iftype != ARPHRD_ETHER)
+                return -EOPNOTSUPP;
+
+        if (!link->dev)
+                return -EOPNOTSUPP;
+
+        r = sd_device_get_devtype(link->dev, &s);
+        if (r < 0)
+                return r;
+
+        if (!streq_ptr(s, "wlan"))
+                return -EOPNOTSUPP;
+
+        r = sd_device_get_syspath(link->dev, &s);
+        if (r < 0)
+                return r;
+
+        s = strjoina(s, "/phy80211");
+        r = sd_device_new_from_syspath(&phy, s);
+        if (r < 0)
+                return r;
+
+        r = sd_device_get_sysname(phy, &s);
+        if (r < 0)
+                return r;
+
+        /* TODO:
+         * Maybe, it is better to cache the found Wiphy object in the Link object.
+         * To support that, we need to investigate what happens when the _phy_ is renamed. */
+
+        return wiphy_get_by_name(link->manager, s, ret);
+}
+
+static int rfkill_get_state(sd_device *dev) {
+        int r;
+
+        assert(dev);
+
+        /* The previous values may be outdated. Let's clear cache and re-read the values. */
+        device_clear_sysattr_cache(dev);
+
+        r = device_get_sysattr_bool(dev, "soft");
+        if (r < 0 && r != -ENOENT)
+                return r;
+        if (r > 0)
+                return RFKILL_SOFT;
+
+        r = device_get_sysattr_bool(dev, "hard");
+        if (r < 0 && r != -ENOENT)
+                return r;
+        if (r > 0)
+                return RFKILL_HARD;
+
+        return RFKILL_UNBLOCKED;
+}
+
+static int wiphy_rfkilled(Wiphy *w) {
+        int r;
+
+        assert(w);
+
+        if (!udev_available()) {
+                if (w->rfkill_state != RFKILL_UNBLOCKED) {
+                        log_wiphy_debug(w, "Running in container, assuming the radio transmitter is unblocked.");
+                        w->rfkill_state = RFKILL_UNBLOCKED; /* To suppress the above log message, cache the state. */
+                }
+                return false;
+        }
+
+        if (!w->rfkill) {
+                if (w->rfkill_state != RFKILL_UNBLOCKED) {
+                        log_wiphy_debug(w, "No rfkill device found, assuming the radio transmitter is unblocked.");
+                        w->rfkill_state = RFKILL_UNBLOCKED; /* To suppress the above log message, cache the state. */
+                }
+                return false;
+        }
+
+        r = rfkill_get_state(w->rfkill);
+        if (r < 0)
+                return log_wiphy_debug_errno(w, r, "Could not get rfkill state: %m");
+
+        if (w->rfkill_state != r)
+                switch (r) {
+                case RFKILL_UNBLOCKED:
+                        log_wiphy_debug(w, "The radio transmitter is unblocked.");
+                        break;
+                case RFKILL_SOFT:
+                        log_wiphy_debug(w, "The radio transmitter is turned off by software.");
+                        break;
+                case RFKILL_HARD:
+                        log_wiphy_debug(w, "The radio transmitter is forced off by something outside of the driver's control.");
+                        break;
+                default:
+                        assert_not_reached();
+                }
+
+        w->rfkill_state = r; /* Cache the state to suppress the above log messages. */
+        return r != RFKILL_UNBLOCKED;
+}
+
+int link_rfkilled(Link *link) {
+        Wiphy *w;
+        int r;
+
+        assert(link);
+
+        r = link_get_wiphy(link, &w);
+        if (IN_SET(r, -EOPNOTSUPP, -ENODEV))
+                return false; /* Typically, non-wifi interface or running in container */
+        if (r < 0)
+                return log_link_debug_errno(link, r, "Could not get phy: %m");
+
+        return wiphy_rfkilled(w);
 }
 
 static int wiphy_update_name(Wiphy *w, sd_netlink_message *message) {
--- a/src/network/networkd-wiphy.h
+++ b/src/network/networkd-wiphy.h
@@ -7,7 +7,18 @@
 
 #include "macro.h"
 
+typedef struct Link Link;
 typedef struct Manager Manager;
+
+/* The following values are different from the ones defined in linux/rfkill.h. */
+typedef enum RFKillState {
+        RFKILL_UNKNOWN,
+        RFKILL_UNBLOCKED,
+        RFKILL_SOFT,
+        RFKILL_HARD,
+        _RFKILL_STATE_MAX,
+        _RFKILL_STATE_INVALID = -EINVAL,
+} RFKillState;
 
 typedef struct Wiphy {
         Manager *manager;
@@ -17,6 +28,7 @@ typedef struct Wiphy {
 
         sd_device *dev;
         sd_device *rfkill;
+        RFKillState rfkill_state;
 } Wiphy;
 
 Wiphy *wiphy_free(Wiphy *w);
@@ -24,6 +36,8 @@ DEFINE_TRIVIAL_CLEANUP_FUNC(Wiphy*, wiphy_free);
 
 int wiphy_get_by_index(Manager *manager, uint32_t index, Wiphy **ret);
 int wiphy_get_by_name(Manager *manager, const char *name, Wiphy **ret);
+
+int link_rfkilled(Link *link);
 
 int manager_genl_process_nl80211_wiphy(sd_netlink *genl, sd_netlink_message *message, Manager *manager);
 int manager_udev_process_wiphy(Manager *m, sd_device *device, sd_device_action_t action);


tomty89:
Tested with 251.3 (made sure that rfkill is unblocked **after** networkd tried to configure wlan0). The problem is gone.

Thanks for the prompt reaction!

tomty89:
@yuwata Hmm, I just noticed that it seems like you would like to prevent `Could not bring up interface, ignoring: Operation not possible due to RF-kill` from showing at all, but I'm still seeing the error (well warning) if rfkill is still blocked, so in that regard something is still not right I suppose?

src/network/networkd-wifi.c
```diff
@@ -12,6 +14,68 @@
...
+                return r;
+
+        if (!streq_ptr(s, "wlan"))
+                return false;
+
+        r = sd_device_get_syspath(link->sd_device, &s);
+        if (r < 0)
+                return r;
+
+        s = strjoina(s, "/phy80211/rfkill1");
```
poettering: i think you need to search for the name here. i.e. iterate through all subdirs of the phy80211/ dir, then check every single one starting with "rfkill" in the name. I think there might be multiple rfkill devices associated with one phy80211 dir

yuwata:
@tomty89 and @poettering Thank you for your review and test. Now the correct rfkill device should be used. PTAL!

tomty89:
Tested with 251.3 again. Now everything works as expected (AFAICT).

yuwata:
@tomty89 Thank you!

src/network/networkd-wifi.c
```diff
@@ -12,6 +14,113 @@
...
+        }
+
+        r = device_get_sysattr_bool(dev, "soft");
+        if (r < 0)
+                return r;
+        if (r > 0) {
+                if (link->rfkill_state != RFKILL_SOFT) {
+                        log_link_debug(link,
+                                       "The radio transmitter is turned off by software. "
+                                       "Waiting for the transmitter being unblocked.");
```
daandemeyer: "being" => "to be"

src/network/networkd-wifi.c
```diff
@@ -12,6 +14,113 @@
...
+        }
+
+        r = device_get_sysattr_bool(dev, "hard");
+        if (r < 0)
+                return r;
+        if (r > 0) {
+                if (link->rfkill_state != RFKILL_HARD) {
+                        log_link_debug(link,
+                                       "The radio transmitter is forced off by something outside of the driver's control. "
+                                       "Waiting for the transmitter being turned on.");
```
daandemeyer: Ditto

src/network/networkd-wifi.c
```diff
@@ -12,6 +14,113 @@
...
+        r = device_get_sysattr_bool(dev, "soft");
+        if (r < 0)
+                return r;
+        if (r > 0) {
+                if (link->rfkill_state != RFKILL_SOFT) {
+                        log_link_debug(link,
+                                       "The radio transmitter is turned off by software. "
+                                       "Waiting for the transmitter being unblocked.");
+                        link->rfkill_state = RFKILL_SOFT;
+                }
```
daandemeyer: Don't we usually put a newline after the end of a block? (ditto elsewhere in this function)

src/network/networkd-setlink.c
```diff
@@ -998,7 +999,7 @@ static int link_up_or_down(Link *link, bool up, Request *req) {
         return request_call_netlink_async(link->manager->rtnl, m, req);
 }
 
-static bool link_is_ready_to_activate(Link *link) {
+static bool link_is_ready_to_activate(Link *link, bool up) {
```
daandemeyer, line 1002: Can we change this to return `int` and propagate the errors from `link_rfkilled()`?

src/network/networkd-setlink.c
```diff
@@ -1097,6 +1101,9 @@ static bool link_is_ready_to_bring_up_or_down(Link *link, bool up) {
         if (!link->activated)
                 return false;
 
+        if (up && link_rfkilled(link) > 0)
```
daandemeyer, line 1104: Ditto for properly propagating the errors from `link_rfkilled()` by making `link_is_ready_to_bring_up_or_down()` return `int`.

src/network/networkd-wifi.c
```diff
@@ -12,6 +14,113 @@
...
+                if (r < 0)
+                        return r;
+
+                r = sd_device_enumerator_add_match_parent(e, phy);
+                if (r < 0)
+                        return r;
+
+                rfkill = sd_device_enumerator_get_device_first(e);
+                if (!rfkill)
+                        return -ENOENT; /* No rfkill device found. */
```
daandemeyer: We should probably return `false` if this is the case? Unless it's expected for every wifi device to have an associated rfkill device.

DaanDeMeyer (changes requested):

src/network/networkd-setlink.c
tomty89: I'm not sure if that's a good idea. IMHO `link_rfkilled()` is somewhat "advisory", if not "cosmetic". Changing the "do that or not" functions (especially when they are not even wifi-specific) for that seems going too far, if not dangerous. (And if the point is to make it more verbose, why not just *do so*?)

daandemeyer: Sure, but we should still log about it if something goes wrong, we can still ignore the error higher up the stack

yuwata: The wifi interface may be handled by the old wext driver, instead of ieee80211. In that case, I do not know what is returned by the function. Hence, as @tomty89 commented, the function should be just a kind of advisory.

src/network/networkd-wifi.c
yuwata: Added a clear debug message.

src/network/networkd-setlink.c
daandemeyer: Can we add more debug logging to link_rfkilled itself then, in the error paths?

yuwata: In the revised version, the function logs in many failure path.

yuwata:
@tomty89 and @DaanDeMeyer Thank you for your comments.
I force-pushed a revised version. The logic is moved from networkd-wifi.c to networkd-wiphy.c as rfkill device in sysfs is under phy (e.g. phy0), not network interface (e.g. wlan0, wlp59s0 or so). The superficial code was changed a lot. But its actual logic should be mostly equivalent to the previous.
PTAL!

src/network/networkd-link.h
```diff
@@ -77,6 +78,7 @@ typedef struct Link {
         char *ssid;
         char *previous_ssid;
         struct ether_addr bssid;
+        RFKillState rfkill_state;
```
daandemeyer: Maybe move this to `Wiphy` as well?

src/network/networkd-wiphy.c
```diff
@@ -80,6 +84,188 @@ int wiphy_get_by_name(Manager *manager, const char *name, Wiphy **ret) {
...
+        Wiphy *w;
+        int r;
+
+        assert(link);
+
+        r = link_get_wiphy(link, &w);
+        if (IN_SET(r, -EOPNOTSUPP, -ENODEV))
+                return false; /* Typically, non-wifi interface. */
+        if (r < 0)
+                return log_link_debug_errno(link, r, "Could not get phy: %m");
```
daandemeyer: I would split this part off into `link_rfkilled()` and move the other part of this function to a new function `wiphy_rfkilled()`. `link_rfkilled()` and `link_get_wiphy()` can then remain in network-link.c. `link_rfkilled()` will then just call `link_get_wiphy()` followed by `wiphy_rfkilled()`

src/network/networkd-wiphy.c
```diff
@@ -80,6 +84,188 @@ int wiphy_get_by_name(Manager *manager, const char *name, Wiphy **ret) {
...
+        if (r < 0)
+                return r;
+
+        r = wiphy_get_by_name(link->manager, s, &w);
+        if (r < 0)
+                return r;
+
+        /* Optionally assign the sd-device object to the Wiphy object. */
+        if (!w->dev)
+                w->dev = TAKE_PTR(phy);
```
daandemeyer: This seems a bit weird, wouldn't it be easier to follow if we just attached the phy device object to the Wiphy object immediately after we initialized the wiphy object in `manager_genl_process_nl80211_wiphy()` and only remove it when we free the `Wiphy` object? That makes the lifetime a bit more clear.

DaanDeMeyer (changes requested):

src/network/networkd-wiphy.c
```diff
@@ -97,6 +283,13 @@ static int wiphy_update_name(Wiphy *w, sd_netlink_message *message) {
         if (streq_ptr(w->name, name))
                 return 0;
 
+        /* Both phy device and rfkill device depends on the name.
+         * E.g. phy device is /sys/devices/pci0000:00/0000:00:1c.2/0000:3b:00.0/ieee80211/phy0
+         * and its rfkill device is /devices/pci0000:00/0000:00:1c.2/0000:3b:00.0/ieee80211/phy0/rfkill1
+         * Hence, we need to recreate the sd-device objects about them after the phy is renamed. */
```
daandemeyer: Can we recreate these immediately in `wiphy_update_name()`?

yuwata:
@DaanDeMeyer Thank you for the comments. The PR now becomes much bigger than I initially expected. Hopefully all comments are addressed. PTAL.

yuwata:
The CentOS CI (CentOS Stream 8) failure is #24040, not related.

src/basic/alloc-util.h
```diff
@@ -63,6 +63,16 @@ typedef void* (*mfree_func_t)(void *p);
 #define free_and_replace(a, b)                  \
         free_and_replace_full(a, b, free)
 
+/* This is similar to free_and_replace_full(), but NULL is not assigned to 'b'. */
+#define unref_and_replace_full(a, b, unref_func) \
+        ({                                       \
+                typeof(a)* _a = &(a);            \
+                typeof(b) _b = (b);              \
+                unref_func(*_a);                 \
+                *_a = _b;                        \
```
daandemeyer, line 73: So I would either not ref `b` but set it to `NULL` or ref `b` but not set it to `NULL`. Both those approaches seem safe, but not reffing and not setting to `NULL` seems like it could easily introduce problems.

src/network/networkd-link.c
```diff
@@ -1488,20 +1488,17 @@ static int link_check_initialized(Link *link) {
...
         assert(m);
         assert(device);
 
         r = sd_device_get_action(device, &action);
-        if (r < 0) {
-                log_device_debug_errno(device, r, "Failed to get udev action, ignoring device: %m");
-                return 0;
-        }
+        if (r < 0)
+                return log_device_warning_errno(device, r, "Failed to get udev action, ignoring device: %m");
```
daandemeyer: Should we really have the "ignoring" in these error strings? Since this function doesn't really know whether the error is going to be ignored or not. I would remove the "ignoring" stuff here and do an extra log that we're ignore the device when calling manager_udev_process_link() in manager_process_uevent()

src/network/networkd-link.c
```diff
@@ -1510,28 +1507,26 @@ int manager_udev_process_link(sd_device_monitor *monitor, sd_device *device, voi
...
-                                       device_action_to_string(action));
-                return 0;
-        }
+        if (r < 0)
+                return log_device_warning_errno(device, r,
+                                                "Failed to determine the device is renamed or not, ignoring '%s' uevent: %m",
+                                                device_action_to_string(action));
         if (r > 0) {
-                log_device_debug(device, "Interface is under renaming, wait for the interface to be renamed.");
+                log_device_debug(device, "Device is under renaming, waiting for the interface to be renamed.");
```
daandemeyer: "Device is renaming" ?

src/network/networkd-wiphy.c
```diff
@@ -203,3 +238,36 @@ int manager_genl_process_nl80211_wiphy(sd_netlink *genl, sd_netlink_message *mes
...
+        const char *name;
+        Wiphy *w;
+        int r;
+
+        assert(m);
+        assert(device);
+
+        r = sd_device_get_action(device, &action);
+        if (r < 0)
+                return log_device_warning_errno(device, r, "Failed to get udev action, ignoring device: %m");
```
daandemeyer: Similar here, I would move the "ignoring device" up one level. But not sure how you'd choose the right log level one level up though

src/network/networkd-wiphy.c
```diff
@@ -203,3 +238,36 @@ int manager_genl_process_nl80211_wiphy(sd_netlink *genl, sd_netlink_message *mes
...
+                /* This error is not critical, as the corresponding genl message may be received later.
+                 * Let's log in the debug level. */
+                return log_device_debug_errno(device, r,
+                                              "Failed to get Wiphy object, ignoring '%s' uevent: %m",
+                                              device_action_to_string(action));
+
+        if (action == SD_DEVICE_REMOVE)
+                w->dev = sd_device_unref(w->dev);
+        else
+                device_unref_and_replace(w->dev, sd_device_ref(device));
```
daandemeyer: `device_unref_and_replace(w->dev, action == SD_DEVICE_REMOVE ? NULL : device)` ?

daandemeyer: Shouldn't this be similar to `manager_udev_process_link()` where we only remove the associated device when nl80211 tells us the device was removed? Or should `manager_udev_process_link()` be made similar to this, so that we also unref the device there?

It'd be great if we could be consistent across link, wiphy and rfkill here, but I'm not super familiar with the details so it might not be possible.

src/network/networkd-wiphy.c
```diff
@@ -271,3 +322,46 @@ int manager_udev_process_wiphy(Manager *m, sd_device *device) {
...
+
+        r = sd_device_get_action(device, &action);
+        if (r < 0)
+                return log_device_warning_errno(device, r, "Failed to get udev action, ignoring: %m");
+
+        r = sd_device_get_syspath(device, &s);
+        if (r < 0)
+                return log_device_warning_errno(device, r, "Failed to get syspath, ignoring: %m");
+
+        /* Do not use sd_device_get_parent() here, as this is a 'remove' uevent. */
```
daandemeyer: It might be any event no? Shouldn't this say, "as this might be a 'remove' uevent" ?

src/network/networkd-wiphy.c
```diff
@@ -86,6 +87,137 @@ int wiphy_get_by_name(Manager *manager, const char *name, Wiphy **ret) {
         return 0;
 }
 
+static int link_get_wiphy(Link *link, Wiphy **ret) {
```
daandemeyer, line 107: I would move this function and `link_rfkilled()` to network-link and keep the rest here.

DaanDeMeyer (changes requested):
I really like how this looks now, just a few more questions and suggestions

src/network/networkd-wiphy.c
yuwata: networkd-link.c is too large, I'd like to place them at networkd-wiphy.c.

yuwata:
@DaanDeMeyer Thank you for reviewing this PR so many times. All comments, except for the last one, are addressed. PTAL!

src/network/networkd-link.c
```diff
@@ -1490,48 +1488,39 @@ static int link_check_initialized(Link *link) {
...
-                return 0;
-
-        r = sd_device_get_ifindex(device, &ifindex);
-        if (r < 0) {
-                log_device_debug_errno(device, r, "Ignoring udev %s event for device without ifindex or with invalid ifindex: %m",
-                                       device_action_to_string(action));
+        /* Let's unref the sd-device object assigned to the corresponding Link object, but keep the Link
+         * object here. It will be removed only when rtnetlink says so. */
+        if (action == SD_DEVICE_REMOVE) {
+                link->sd_device = sd_device_unref(link->sd_device);
```
daandemeyer: Nit, but maybe add a commit to rename the `sd_device` field to `dev` just like in `Wiphy` ?

src/network/networkd-link.c
```diff
@@ -1490,48 +1488,39 @@ static int link_check_initialized(Link *link) {
...
         }
 
         r = device_is_renaming(device);
-        if (r < 0) {
-                log_device_debug_errno(device, r, "Failed to determine the device is renamed or not, ignoring '%s' uevent: %m",
-                                       device_action_to_string(action));
-                return 0;
-        }
+        if (r < 0)
+                return log_device_debug_errno(device, r, "Failed to determine the device is renamed or not: %m");
```
daandemeyer: "Failed to determine if the device is renaming or not" ?

src/network/networkd-wiphy.c
```diff
@@ -80,6 +87,137 @@ int wiphy_get_by_name(Manager *manager, const char *name, Wiphy **ret) {
...
+                return r;
+
+        /* TODO:
+         * Maybe, it is better to cache the found Wiphy object in the Link object.
+         * To support that, we need to investigate what happens when the _phy_ is renamed. */
+
+        if (ret)
+                *ret = w;
+
+        return 0;
```
daandemeyer: `return wiphy_get_by_name(link->manager, s, ret);` ?

src/network/networkd-wiphy.c
```diff
@@ -104,7 +242,89 @@ static int wiphy_update_name(Wiphy *w, sd_netlink_message *message) {
...
+        int r;
+
+        assert(w);
+
+        if (!udev_available())
+                return 0;
+
+        w->dev = sd_device_unref(w->dev);
+
+        if (!w->name)
```
daandemeyer: I wonder if we should simply not allocate a new Wiphy object if it doesn't have a name? I'd move `r = sd_netlink_message_read_string(message, NL80211_ATTR_WIPHY_NAME, &name);` out of `wiphy_update_name()` and into  `manager_genl_process_nl80211_wiphy()` and then if we can't read the wiphy name, ignore the message.

From nl80211.h:

```
 * @NL80211_CMD_NEW_WIPHY: Newly created wiphy, response to get request
 *	or rename notification. Has attributes %NL80211_ATTR_WIPHY and
 *	%NL80211_ATTR_WIPHY_NAME.
```

So I think we can assume that each new wiphy genl message will have a name attr included.

With this change, we can assume `name` is set everywhere else and remove all these `if` statements that check if `w->name` is set.

This can be done in a later PR though.

DaanDeMeyer (approved):
Just a few nits and a suggestion for a future follow up

DaanDeMeyer:
@tomty89 Can you give this one more test to make sure everything still works for you?

yuwata:
@DaanDeMeyer Thank you. All comments are addressed. Upgrading the green label.

yuwata:
Please backport b05e52000b4eee764b383cc3031da0a3739e996e.

keszybz:
Queued up for v251.5.

PR #24021, opened 2022-07-14
man: elaborate on the usefulness (and not-so-usefulness) of various process resource limits

poettering:


bluca (approved):

yuwata:
No CI is running for this PR, but I confirmed that the man page can be built. Merging.

keszybz:
This is opinionated, but I think it's the right thing to say.

PR #24022, opened 2022-07-14
Add systemd-hwdb to bug/RFE templates

daandemeyer:


yuwata:
Please also update advanced-issue-labeler.yml

PR #24024, opened 2022-07-14
coredump: Connect stdout/stderr to /dev/null before doing anything

daandemeyer:
When invoked as the coredump handler by the kernel, systemd-coredump's
stdout and stderr streams are closed. This is dangerous as this means
the fd's can get reallocated, leading to hard to debug errors such as
log messages ending up being appended to a compressed coredump file.

To avoid such issues in the future, let's bind stdout/stderr to
/dev/null so the file descriptors can't get used for anything else.

### DIFF
--- a/src/coredump/coredump.c
+++ b/src/coredump/coredump.c
@@ -1271,6 +1271,13 @@ static int process_kernel(int argc, char* argv[]) {
         struct iovec_wrapper *iovw;
         int r;
 
+        /* When we're invoked by the kernel, stdout/stderr are closed which is dangerous because the fds
+         * could get reallocated. To avoid hard to debug issues, let's instead bind stdout/stderr to
+         * /dev/null. */
+        r = rearrange_stdio(STDIN_FILENO, -1, -1);
+        if (r < 0)
+                return log_error_errno(r, "Failed to connect stdout/stderr to /dev/null: %m");
+
         log_debug("Processing coredump received from the kernel...");
 
         iovw = iovw_new();


DaanDeMeyer:
This makes sure parse_package_metadata() doesn't append log messages to coredump files anymore, but I still wonder if this isn't hiding the real issue which is that parse_package_metadata() is trying to log to stderr in the first place. We explicitly set the log target to LOG_TARGET_KMSG in systemd-coredump and I would expect that to get inherited over to the forked process that does the ELF parsing, so I don't get why the log macros are writing to stderr in the first place.

cc @bluca

DaanDeMeyer:
I guess this logic in log.c is the root cause:

```
                        k = write_to_kmsg(level, error, file, line, func, buffer);
                        if (k < 0) {
                                log_close_kmsg();
                                (void) log_open_console();
                        }
```

parse_elf() is sandboxed so I'm guessing it's not going to be allowed to write to kmsg, so we fall back to the console, which ended up writing to the coredump fd that got allocated to stderr.

Yikes what a bug

poettering:
wowza. what an embarassing bug! jeezuz

keszybz:
Yikes.

poettering:
I wonder if we have similar issues with other user mode helpers, i.e. kernel callouts?

At least systemd-cgroups-agent comes to mind here. I wonder if we have others?

poettering:
I prepped https://github.com/systemd/systemd/pull/24029 about the cgroups agent case. 

Anyone can think of any other user mode helpers we define?

Issue #24023, opened 2022-07-14
systemd-hwdb update takes 50s to complete when built with sanitizers


daandemeyer:
### systemd version the issue has been seen with

main

### Used distribution

Fedora 36

### Linux kernel version used

5.18.5-200.fc36.x86_64

### CPU architectures issue was seen on

x86_64

### Component

other

### Expected behaviour you didn't see

`systemd-hwdb update` completes quickly

### Unexpected behaviour you saw

`systemd-hwdb update` takes 50s to complete when built with sanitizers. This makes it a bit awkward to enable sanitizers by default in mkosi.

cc @mrc0mmand Have you dealt with this in the CI at all?

mrc0mmand:
I don't think this is a bug. Reading hwdb does _a lot_ of string ops, which are heavily sanitized (and thus slowed) by ASan, hence the slowdown overall (@evverx correct me please if I'm wrong). In CIs we've dealt with that by bumping the timeout:

https://github.com/systemd/systemd/blob/5eab88a569c1bb5f79093f38e872e588d64a2b26/test/test-functions#L882-L883

The `create_asan_wrapper()` function overall is a great example of what all needs to be done to make a full system run under sanitizers.

keszybz:
Yeah. This does not  sound like a bug, but just the expected thing with sanitizers. Valgrind is also slow.

evverx:
> The create_asan_wrapper() function overall is a great example of what all needs to be done to make a full system run under sanitizers.

I'm not sure it would be (fully) suitable for mkosi. The test suite kind of controls the environment where the tests are run so it can more or less get away with installing `systemd` and setting `LD_PRELOAD` where necessary while `mkosi` has to additionally run services provided by distributions. I think there it would make sense to redirect `systemd` and all the `systemd` services to the directory where `systemd` is built. This way they would pick up the libraries built with ASan/UBSan (thanks to `rpath`) and everything else would be run as usual. `mkosi` still should set `log_path` for PID1 and journald and look for ASan/UBsan reports once VMs/containers are stopped though.

> In CIs we've dealt with that by bumping the timeout

I'd also add that `systemd-hwdb-update.service` is usually masked and run under ASan/UBsan in a couple of tests only.

DaanDeMeyer:
For now I'm looking more at sanitizers just for development purposes. I figure we can do the same as CI and just disable systemd-hwdb-update.service by default in mkosi though, so that we can build with sanitizers by default

[...]

PR #24025, opened 2022-07-15
mkosi: Changes to allow booting with sanitizers in mkosi

daandemeyer:
mkosi: Changes to allow booting with sanitizers in mkosi
- Extra memory because ASAN needs it
- The environment variables to make the sanitizers more useful
- LD_PRELOAD because the ASAN DSO needs to be the first in the list
- The Fedora sanitizer library packages, I skipped the other distros
  for now because there's some issues to fix like centos stream 8 and
  9 shipping with different versions of libasan so we can't set
  LD_PRELOAD to the right value in the mkosi config because there's
  no libasan.so (only one suffixed with the version)

We don't yet build with sanitizers by default because of
https://github.com/systemd/systemd/issues/24023.

### DIFF
--- a/mkosi.default.d/10-systemd.conf
+++ b/mkosi.default.d/10-systemd.conf
@@ -9,7 +9,7 @@ HostonlyInitrd=yes
 # Prevent ASAN warnings when building the image
 Environment=ASAN_OPTIONS=verify_asan_link_order=false
 
-[Packages]
+[Content]
 BuildDirectory=mkosi.builddir
 Cache=mkosi.cache
 InstallDirectory=mkosi.installdir
@@ -17,7 +17,7 @@ SourceFileTransferFinal=copy-git-others
 
 [Host]
 QemuHeadless=yes
-NetworkVeth=yes
+Netdev=yes
 QemuMem=2G
 
 [Validation]
--- a/mkosi.default.d/debian/10-mkosi.debian
+++ b/mkosi.default.d/debian/10-mkosi.debian
@@ -7,7 +7,7 @@
 Distribution=debian
 Release=testing
 
-[Packages]
+[Content]
 BuildPackages=
         acl
         clang
--- a/mkosi.default.d/fedora/10-mkosi.fedora
+++ b/mkosi.default.d/fedora/10-mkosi.fedora
@@ -7,7 +7,7 @@
 Distribution=fedora
 Release=36
 
-[Packages]
+[Content]
 BuildPackages=
         diffutils
         docbook-style-xsl
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -66,7 +66,7 @@ reuse the host's package cache. To do this, create a mkosi override file in
 mkosi.default.d/ (e.g 20-local.conf) and add the following contents:
 
 ```
-[Packages]
+[Content]
 Cache=<full-path-to-package-manager-cache> # (e.g. /var/cache/dnf)
 ```
 
@@ -229,7 +229,7 @@ mkosi's config. The easiest way to set the option is to create a file 20-local.c
 add the following contents:
 
 ```
-[Packages]
+[Content]
 IncludeDirectory=mkosi.includedir
 ```
 
@@ -238,7 +238,7 @@ We already configured clangd to map any paths in /usr/include in the build image
 host in the mkosi-clangd.sh script.
 
 We also need to make sure clangd is installed in the build image. To have mkosi install clangd in the build
-image, edit the 20-local.conf file we created earlier and add the following contents under the `[Packages]`
+image, edit the 20-local.conf file we created earlier and add the following contents under the `[Content]`
 section:
 
 ```
--- a/mkosi.default.d/arch/10-mkosi.arch
+++ b/mkosi.default.d/arch/10-mkosi.arch
@@ -8,7 +8,7 @@
 [Distribution]
 Distribution=arch
 
-[Packages]
+[Content]
 BuildPackages=
         acl
         bzip2
--- a/mkosi.default.d/centos_epel/10-mkosi.centos_epel
+++ b/mkosi.default.d/centos_epel/10-mkosi.centos_epel
@@ -10,7 +10,7 @@ Distribution=centos_epel
 Format=gpt_xfs
 HostonlyInitrd=no
 
-[Packages]
+[Content]
 BuildPackages=
         diffutils
         docbook-style-xsl
--- a/mkosi.default.d/opensuse/10-mkosi.opensuse
+++ b/mkosi.default.d/opensuse/10-mkosi.opensuse
@@ -7,7 +7,7 @@
 Distribution=opensuse
 Release=tumbleweed
 
-[Packages]
+[Content]
 BuildPackages=
         docbook-xsl-stylesheets
         fdupes
--- a/mkosi.default.d/ubuntu/10-mkosi.ubuntu
+++ b/mkosi.default.d/ubuntu/10-mkosi.ubuntu
@@ -8,7 +8,7 @@ Distribution=ubuntu
 Release=jammy
 Repositories=main,universe
 
-[Packages]
+[Content]
 BuildPackages=
         acl
         docbook-xml
--- a/test/mkosi.default.networkd-test
+++ b/test/mkosi.default.networkd-test
@@ -21,7 +21,7 @@ Output=networkd-test.raw
 [Partitions]
 RootSize=3G
 
-[Packages]
+[Content]
 BuildPackages=
         audit-libs-devel
         bzip2-devel


bluca:
```
In /github/workspace/mkosi.postinst line 10:
    find / -name '*.service' -type f | xargs sed -i 's/^\(MemoryDeny\|SystemCall\)/# \1/'
    ^-- SC2038 (warning): Use -print0/-0 or -exec + to allow for non-alphanumeric filenames.
```

mkosi.default.d/10-systemd.conf
```diff
@@ -6,6 +6,8 @@
 Format=gpt_btrfs
 Bootable=yes
 HostonlyInitrd=yes
+KernelCommandLine=systemd.setenv=ASAN_OPTIONS=strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1:disable_coredump=0:use_madv_dontdump=1
+                  systemd.setenv=UBSAN_OPTIONS=print_stacktrace=1:print_summary=1:halt_on_error=1
```
evverx: I don't think `systemd.setenv` affects PID1 itself. To pass those variables to `systemd` they should be exported before running `/sbin/init`. Another option would be to set them using `ManagerEnvironment` but last time I checked it wasn't exactly reliable in the sense that variables set via that setting got reset somewhere and appeared on `daemon-reexec` only.

To make it easier to figure out why PID1 and journald fail under ASan/UBsan `log_path` should probably be set as well. It kind of works even without that as long as there are no issues there but it might come in handy when new bugs are introduced.

mkosi.default.d/fedora/10-mkosi.fedora
```diff
@@ -6,6 +6,7 @@
 [Distribution]
 Distribution=fedora
 Release=36
+KernelCommandLine=systemd.setenv=LD_PRELOAD=/usr/lib64/libasan.so.8
```
evverx: I'm not sure how the "postinst" hook works but if it could look for the library and add it to `DefaultEnvironment` by analogy with the test suite it wouldn't be necessary to hardcode the path. The test suite is compatible with `clang` though so if it isn't important here that part could probably be cut off.

daandemeyer: I agree this isn't ideal. If we do it via the config file in postinst, the changes won't end up in the initrd because the postinst script runs after the initrd has been generated. We could probably put the logic in the build script instead.

I'll see if I can rework this to do that.

evverx: Got it.

As far as I can tell arguments passed via `KernelCommandLine` aren't passed to systemd run in nspawn containers (at least with this PR applied `mkosi` failed to boot `systemd` built with ASan until `LD_PRELOAD` was added to `DefaultEnvironment` and `systemd` was reexecuted with `nsenter` manually) so it seems if there was a way to put it in config files in the build script it would automatically make it compatible with `nspawn` as well. (It could be that I screwed something up somewhere though)

daandemeyer: I have a PR open to make mkosi pass kernel cmdline arguments to nspawn. I'm also looking at adding systemd.setenv_manager to set manager environment variables from the kernel cmdline

evverx: > I have a PR open to make mkosi pass kernel cmdline arguments to nspawn

That should work with `mkosi` but as far as I understand for example `systemd-nspawn -i` wouldn't work with those images without passing those arguments manually. I don't know If they are supposed to be used like that though.

daandemeyer: That's true, I thought about having systemd in nspawn read from /etc/kernel/cmdline, but that seems like it'd be a bit more of an involved change

evverx: I think it would be possible to avoid reading `/etc/kernel/cmdline` if those images were self-contained with `LD_PRELOAD`, `ASAN_OPTIONS` and `UBSAN_OPTIONS` kept in `/etc/systemd/system.conf`. Having said that I agree that `KernelCommandLine` should be good enough at this point considering that `mkosi` is used to boot those images most of the time.

I'm wondering if there is a way to avoid setting `LD_PRELOAD` here when `systemd` is built without ASan though? My understanding is that an environment variable can be used to add `-Db_sanitize=address,undefined` and `--timeout-multiplier=...` to `mkosi.build` but I'm not sure how it can turn on/off `KernelCommandLine` conditionally.

daandemeyer: Yeah, that's the problem I'm running into now that prevents using KernelCommandLine. So I'm going to rework to use the config file instead to avoid that limitation since in the build script we can do an if check on the environment variable

Also, we don't need systemd.setenv_manager, the kernel can already do it for us (see https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html)

DaanDeMeyer:
Sanitizers can now be configured with the SANITIZERS environment variable. This PR depends on two mkosi PRs: https://github.com/systemd/mkosi/pull/1061 and https://github.com/systemd/mkosi/pull/1062

mkosi.build
```diff
@@ -49,18 +49,20 @@ if [ ! -f "$BUILDDIR"/build.ninja ] ; then
...
 
         meson "$BUILDDIR" \
                 -D "sysvinit-path=$sysvinit_path" \
                 -D "rootprefix=$rootprefix" \
                 -D man=false \
                 -D translations=false \
-                -D version-tag="${VERSION_TAG}"
+                -D version-tag="${VERSION_TAG}" \
+                -D mode=developer \
+                -D b_sanitize=${SANITIZERS:-none}
```
evverx: I think when `SANITIZERS` is set `ninja test` should be replaced with something like `meson test --timeout-multiplier=3`. Without that some tests time out every now and again. Also, it seems it would be somewhat more consistent if  `ASAN_OPTIONS` and `UBSAN_OPTIONS` were passed to the test suite as well.

evverx:
Given how useful this is I think it should certainly be mentioned somewhere in https://github.com/systemd/systemd/blob/main/docs/HACKING.md#hacking-on-systemd eventually.

I'd wait for those PRs to be merged to test it. I'll try to inject memory leaks, integer overflows and stuff like that to see where all the reports go, whether they are symbolized or not and how easy it is to find/make sense of them. Then again even if it misses some issues it should be fine because the test suite should catch them anyway. As far as I understand the idea is to make it easier to develop systemd and I'm pretty sure that goal has already been achieved here.

mkosi.default.d/10-systemd.conf
daandemeyer: Hmm, for interactive use I'd expect the journal to suffice really. It's great to have all the information in one place when debugging something

DaanDeMeyer:
> Given how useful this is I think it should certainly be mentioned somewhere in https://github.com/systemd/systemd/blob/main/docs/HACKING.md#hacking-on-systemd eventually.

Added a blurb to TESTING_WITH_SANITIZERS.md

mkosi.default.d/10-systemd.conf
evverx: > I'd expect the journal to suffice really

`systemd-journald` redirects its ASan/UBSan reports straight to `/dev/null` by default so if they aren't captured with `log_path` they end up there with no way to view them. `systemd` can send its ASan/UBSan reports to different places depending on where exactly it fails. They can end up in `/dev/kmsg` (if it isn't run in containers), they can be sent to the console bypassing `journald` altogether or to `/dev/null`.

> It's great to have all the information in one place when debugging something

It is but in practice it doesn't always work when `journald` itself fails or `systemd` crashes before `journald` is started or after it's stopped.

evverx: Granted, there are failure modes where `log_path` isn't helpful either but at least with it most backtraces are captured.

evverx:
> Added a blurb to TESTING_WITH_SANITIZERS.md

Thanks! As far as I can remember that page was added to document some stuff the CI did at the time and I don't think it's referred to as often as HACKING.md so in terms of reaching more contributors I still think it would be better to add it to HACKING.md. Having said that I agree  based on its content `TESTING_WITH_SANITIZERS.md` is probably where it belongs.

evverx:
Looks like `shellcheck` isn't happy:
```
2022-07-16T13:17:52.0978680Z ------
2022-07-16T13:17:52.0978846Z 
2022-07-16T13:17:52.0979116Z [1mIn /github/workspace/mkosi.build line 68:[0m
2022-07-16T13:17:52.0979554Z [0m                -D b_sanitize=${SANITIZERS:-none}[0m
2022-07-16T13:17:52.0980207Z [32m                              ^-----------------^ SC2086 (info): Double quote to prevent globbing and word splitting.[0m
2022-07-16T13:17:52.0980490Z 
2022-07-16T13:17:52.0980668Z [1mDid you mean: [0m
2022-07-16T13:17:52.0981127Z                 -D b_sanitize="${SANITIZERS:-none}"
2022-07-16T13:17:52.0981289Z 
2022-07-16T13:17:52.0981295Z 
2022-07-16T13:17:52.0981545Z [1mIn /github/workspace/mkosi.build line 78:[0m
2022-07-16T13:17:52.0981998Z [0m        if [ -n $SANITIZERS ]; then[0m
2022-07-16T13:17:52.0982560Z [31m                ^---------^ SC2070 (error): -n doesn't work with unquoted arguments. Quote or use [[ ]].[0m
2022-07-16T13:17:52.0983517Z [32m                ^---------^ SC2086 (info): Double quote to prevent globbing and word splitting.[0m
2022-07-16T13:17:52.0983892Z 
2022-07-16T13:17:52.0984099Z [1mDid you mean: [0m
2022-07-16T13:17:52.0984493Z         if [ -n "$SANITIZERS" ]; then
2022-07-16T13:17:52.0984683Z 
2022-07-16T13:17:52.0984819Z For more information:
2022-07-16T13:17:52.0985302Z   https://www.shellcheck.net/wiki/SC2070 -- -n doesn't work with unquoted arg...
2022-07-16T13:17:52.0985938Z   https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
2022-07-16T13:17:52.0986363Z ------[0m
```

docs/TESTING_WITH_SANITIZERS.md
```diff
@@ -13,6 +13,20 @@ This is mostly done automagically by various CI systems for each PR, but you may
...
+To build with sanitizers in mkosi, create a file 20-local.conf in mkosi.default.d/ and add the following
+contents:
+
+```
+[Content]
+Environment=SANITIZERS=address,undefined
+```
+
+The value of `SANITIZERS` is passed directly to meson's `b_sanitize` option, so you can set it to any value
+supported by `b_sanitize`. See https://mesonbuild.com/Builtin-options.html#base-options for a list of
```
evverx: It think it's a bit misleading in the sense that for example it isn't possible to just pass `memory` here and expect it to work because it would require `clang`, the dependencies built with `MSan`, `MSAN_OPTIONS`, `LD_LIBRARY_PATH` sprinkled here and there and `libc` more or less compatible with `MSan` and so on. I think it would be fair to say that `address` and `undefined` (shipped with `gcc`) is what it can work with and what is supported at this point.

daandemeyer: Updated to mention only the gcc sanitizers are supported

DaanDeMeyer:
Added a note to HACKING.md referring to TESTING_WITH_SANITIZERS.md. Should also have fixed the shellcheck options.

Once the mkosi PRs are merged we should probably enable this in CI as well

mkosi.default.d/10-systemd.conf
daandemeyer: > systemd-journald redirects its ASan/UBSan reports straight to /dev/null by default so if they aren't captured with log_path they end up there with no way to view them. systemd can send its ASan/UBSan reports to different places depending on where exactly it fails. They can end up in /dev/kmsg (if it isn't run in containers), they can be sent to the console bypassing journald altogether or to /dev/null.

This also depends on where systemd-journald is run. If it has access to /dev/kmsg, it will log there as well (and those messages) eventually end up in the journal.

I get the point that the journal might not always be accessible though. Ideally we could have asan log to both stderr and a log file. But log_path doesn't take multiple values. There's a log_to_syslog option as well, though if used for journald it might introduce a log cycle easily in journald

DaanDeMeyer:
Added the asan libraries for all distros instead of just Fedora

mkosi.default.d/10-systemd.conf
evverx: > This also depends on where systemd-journald is run. If it has access to /dev/kmsg, it will log there as well (and those messages) eventually end up in the journal.

ASan/UBsan reports aren't logged with `log_*` (used by `systemd`). They are logged to `stderr` by `ASan/UBsan` and if `stderr` points to `/dev/null` (which is set by default in `systemd-journald.service`) it is supposed to always go to `/dev/null`. If those reports somehow end up in `/dev/kmsg` it would be a bug in systemd because it would mean that it failed to set up `StandardOutput/StandardError` correctly.

> But log_path doesn't take multiple values

It doesn't but pids are added at the end.

evverx:
> Once the mkosi PRs are merged we should probably enable this in CI as well

For that to work `mkosi` would have to also look for asan/ubsan reports by analogy with https://github.com/systemd/systemd/blob/cbc2593eeaf35a42881319d7fa50b12fc5584bf9/test/test-functions#L1356.

mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -73,6 +73,9 @@ Packages=
         locales
         nano
         strace
+        xxd
+        libasan8
+        libubsan1
```
evverx: On Debian the `gcc` package depends on `libasan*` and `libubsan*` transitively via `gcc-*` and `libgcc-*-dev` so they are pulled automatically when `gcc` is installed. I think it would be better to leave out these packages and let `gcc` pull the right dependencies to avoid editing this file every time `gcc` gets updated.

Also, it seems `xxd` is commented out everywhere else.

mkosi.default.d/10-systemd.conf
```diff
@@ -6,6 +6,8 @@
 Format=gpt_btrfs
 Bootable=yes
 HostonlyInitrd=yes
+# Prevent ASAN warnings when building the image
+Environment=ASAN_OPTIONS=verify_asan_link_order=false
```
evverx, line 10: I wonder why it should be set to `false` here? If `mkosi` ends up using `systemd` libraries built and installed with ASan without setting `LD_PRELOAD=` it seems it should be fixed either by passing on `LD_PRELOAD=` somewhere or by avoiding using half-installed `systemd` to build images.

mkosi.default.d/debian/10-mkosi.debian
daandemeyer: The gcc package is not installed in the final image, only in the build image so we need to pull in the libasan/libubsan deps explicitly there.

xxd isn't commented out, it's a comment explaining what the package below that comment is providing, will clarify the comments

mkosi.default.d/10-systemd.conf
daandemeyer: mkosi itself is fine, it's running dracut that causes issues. We can't set LD_PRELOAD in mkosi itself because we'd leak context from systemd into mkosi. Ignoring the sanitizer errors during image build is probably fine. It's only set during the build process and everything from boot onwards is using LD_PRELOAD.

mkosi.default.d/debian/10-mkosi.debian
evverx: > we need to pull in the libasan/libubsan deps explicitly there

I wonder why `gcc` can't be installed instead? `libasan8` and `libubsan1` are going to get out of sync with `gcc` used to build `systemd`. If it can't be installed here for some reason I think `gcc-11` should be installed explicitly to catch the mismatch when `systemd` is built.

daandemeyer: We now install specific compiler versions on Debian/Ubuntu and Opensuse to avoid mismatch with libasan/libubsan libraries. It'd be nice if these distros would just provide unversion libasan/libubsan packages like fedora/centos

evverx: > It'd be nice if these distros would just provide unversion libasan/libubsan packages

Agreed. That would be ideal.

mkosi.default.d/10-systemd.conf
evverx: > Ignoring the sanitizer errors during image build is probably fine

Got it. I think as long as it works it should be fine.

mkosi.default.d/10-systemd.conf
daandemeyer: > ASan/UBsan reports aren't logged with log_* (used by systemd). They are logged to stderr by ASan/UBsan and if stderr points to /dev/null (which is set by default in systemd-journald.service) it is supposed to always go to /dev/null. If those reports somehow end up in /dev/kmsg it would be a bug in systemd because it would mean that it failed to set up StandardOutput/StandardError correctly.

You're right, I wrongly assumed everything in journald would go to kmsg

I would do it like this, when we're in CI mode, we set log_path, if we're in regular interactive mode, we have ASAN log to stdout, is that a good compromise? We already check the CI_BUILD environment variable in the build script so there's a precedence for this

mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -7,14 +7,15 @@
 Distribution=debian
 Release=testing
 
-[Packages]
+[Content]
+Environment=CC=gcc-12
```
evverx: Personally I'm kind of on the fence about this. I think what's great about the mkosi CI is that among other things it can be used to make sure that `systemd` is buildable with toolchains shipped by distributions by default. `gcc-12` is available on Debian but `gcc` still points to `gcc-11` so with this patch applied `mkosi` no longer builds `systemd` using the default compiler and `gcc` along with `libasan/libubsan` have to be kept in sync basically manually. I think it would be better if `gcc` could be added to `Packages` instead of `libasan*` and `libubsan*`.

mkosi.default.d/10-systemd.conf
evverx: > when we're in CI mode, we set log_path, if we're in regular interactive mode, we have ASAN log to stdout, is that a good compromise?

`StandardOutput` is set to `null` in `systemd-journald.service` so ASan/UBsan reports would be lost as well. It could work if `StandardOutput` was sent to a file (and the test suite does that too in addition to setting `log_path` to get around a bug in `libasan`) but in a way that would be the same as always setting `log_path`.

(it could be that I misunderstood what "we have ASAN log to stdout" means)

mkosi.default.d/debian/10-mkosi.debian
daandemeyer: Yeah, I give up, we'll just install gcc on these distros.

mkosi.default.d/10-systemd.conf
evverx: Just to clarify I think it can be done later in a PR where ASan is integrated into the mkosi CI (and where those scenarios should be handled automatically). In interactive mode it should be possible to spot failures triggered by ASan/UBSan and those backtraces can be restored interactively by tweaking the environment variables manually. It's not ideal but personally I can live with that.

daandemeyer: Sounds good.

I'm not sure how to solve the issue in interactive mode for journald though. Ideally we could tell asan to log to kmsg just like the rest of journald. We could connect stdout to a file but that requires the developer to explicitly look into that file to get the leaks.

Another idea is to connect stdout and stderr directly to /dev/kmsg, journal should automatically pick all those messages up and store them in the journal

daandemeyer:
```
Jul 18 14:27:35 fedora unknown: 
Jul 18 14:27:35 fedora unknown: 
Jul 18 14:27:35 fedora ==386==ERROR: LeakSanitizer: detected memory leaks
Jul 18 14:27:35 fedora unknown: 
Jul 18 14:27:35 fedora unknown: 
Jul 18 14:27:35 fedora unknown: 
Jul 18 14:27:35 fedora unknown: Direct leak of 100 byte(s) in 1 object(s) allocated from:
Jul 18 14:27:35 fedora unknown: 
Jul 18 14:27:35 fedora unknown:     #0 0x7fa36b6a168f in __interceptor_malloc (/lib64/libasan.so.8+0xba68f)
                            #1 0x4089cf in main ../src/src/journal/journald.c:31
                            #2 0x7fa368e8e54f in __libc_start_call_main (/lib64/libc.so.6+0x2954f)
Jul 18 14:27:35 fedora SUMMARY: AddressSanitizer: 100 byte(s) leaked in 1 allocation(s).
```

Not ideal but it does the trick.

evverx: I think the problem is that when `journald` fails under ASan/UBSan it tends to fail hard and get rate-limited by systemd eventually so it never picks up `/dev/kmsg`. Given that in interactive mode there should always be console probably I think setting `StandardOutput` to `kmsg+console` should suffice in the sense that those backtraces would be hard to miss.

daandemeyer: The problem with that is that `kmsg+console` is implemented by forwarding logs to the journald socket so it wouldn't really work either if journald breaks down.

We could try `StandardOutput=file:/dev/console` though I guess. And in CI we could use log_path instead.

daandemeyer: This seems to work even better:

```
[root@fedora ~]# systemctl kill -s TERM systemd-journald
Received SIGTERM from PID 1 (systemd).
[root@fedora ~]# 
==386==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 100 byte(s) in 1 object(s) allocated from:
    #0 0x7fed008c268f in __interceptor_malloc (/lib64/libasan.so.8+0xba68f)
    #1 0x4089cf in main ../src/src/journal/journald.c:31
    #2 0x7fecfe0af54f in __libc_start_call_main (/lib64/libc.so.6+0x2954f)

SUMMARY: AddressSanitizer: 100 byte(s) leaked in 1 allocation(s).
```

evverx: > This seems to work even better

Nice! Looks like `StandardOutput=file:/dev/console` is the way to go

daandemeyer: In systemd-nspawn the logs only appear on the console after a systemd-journald restart though.

```
[root@image ~]# systemctl kill -s TERM systemd-journald
[root@image ~]# systemd-journald.service: Scheduled restart job, restart counter is at 1.

[root@image ~]# systemctl kill -s TERM systemd-journald
Received SIGTERM from PID 1 (systemd).
[root@image ~]# 
==134==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 100 byte(s) in 1 object(s) allocated from:
    #0 0x7f11a0a3368f in __interceptor_malloc (/lib64/libasan.so.8+0xba68f)
    #1 0x4089cf in main ../src/src/journal/journald.c:31
    #2 0x7f119e22054f in __libc_start_call_main (/lib64/libc.so.6+0x2954f)

SUMMARY: AddressSanitizer: 100 byte(s) leaked in 1 allocation(s).
systemd-journald.service: Scheduled restart job, restart counter is at 2.
```

I switched to `StandardOutput=tty` which defaults to /dev/console

evverx: @DaanDeMeyer could you add a drop-in setting that? I think with that this PR should probably fully cover the interactive mode. It can be done later though. Just let me know if you are planning to add it so that I could postpone installing `mkosi` with those PRs merged and starting testing this PR.

docs/TESTING_WITH_SANITIZERS.md
```diff
@@ -13,6 +13,20 @@ This is mostly done automagically by various CI systems for each PR, but you may
 want to do it locally as well. The process slightly varies depending on the
 compiler you want to use and which part of the test suite you want to run.
 
+## mkosi
+
+To build with sanitizers in mkosi, create a file 20-local.conf in mkosi.default.d/ and add the following
```
evverx, line 18: Since this feature is new I think `mkosi` could be replaced with something like `mkosi >= 14 (or built from source)` to make it clear that older versions can't be used to make it work.

mkosi.default.d/10-systemd.conf
daandemeyer: I added the dropin to the PR

daandemeyer: I fixed the nspawn issue as well. Turns out both systemd and login were calling vhangup() on /dev/console, making ASAN unable to write the logs to /dev/console. Stopping both of those from calling vhangup() fixed the issue.

evverx: Thanks! I've just installed `mkosi` and built `systemd` but I got distracted by a bunch of skipped tests. Looks like some dependencies aren't installed there. It has nothing to do with this PR though so I'll get back to that later once all strategically injected integer overflows pop up.

evverx (approved):
It seems to be working. LGTM. Thanks!

Ubuntu CI has been failing to clone systemd since Saturday so it doesn't make much sense to wait for it. I'll go ahead and merge it.

PR #24030, opened 2022-07-15
Fix typo in systemd.network documentation in IPv6RoutePrefix section

lastkrick:
Hello! Looks like there is a typo in systemd.networkd documentation. There is no IPv6PrefixRoutes section, it should be IPv6RoutePrefix instead.


yuwata:
Thanks, but please do not wrap lines needlessly..

PR #24033, opened 2022-07-15
Add `systemctl list-automounts`

### LINKED ISSUE #6056: [RFE] systemctl list-automounts
### Submission type

 - Request for enhancement (RFE)

systemctl currently supports `list-{units,sockets,timers}`

The addition of `systemctl list-automounts` would help simplify reviewing existing automounts.

> 233

### Used distribution

> fedora

dtardon:
Fixes #6056.

### DIFF
--- a/shell-completion/bash/systemctl.in
+++ b/shell-completion/bash/systemctl.in
@@ -215,7 +215,7 @@ _systemctl () {
                              suspend-then-hibernate kexec list-jobs list-sockets
                              list-timers list-units list-unit-files poweroff
                              reboot rescue show-environment suspend get-default
-                             is-system-running preset-all'
+                             is-system-running preset-all list-automounts'
         [FILE]='link switch-root bind mount-image'
         [TARGETS]='set-default'
         [MACHINES]='list-machines'
--- a/shell-completion/zsh/_systemctl.in
+++ b/shell-completion/zsh/_systemctl.in
@@ -7,6 +7,7 @@
 
     local -a unit_commands=(
         # Unit Commands
+        "list-automounts:List automounts"
         "list-sockets:List sockets"
         "list-timers:List timers"
         "list-units:List units"


man/systemctl.xml
```diff
@@ -108,6 +108,25 @@ To show all installed unit files use 'systemctl list-unit-files'.</programlistin
...
+        <varlistentry>
+          <term><command>list-automounts</command> <optional><replaceable>PATTERN</replaceable>…</optional></term>
+
+          <listitem>
+            <para>List automount units currently in memory, ordered by mount path.  If one or more
+            <replaceable>PATTERN</replaceable>s are specified, only automount units matching one of them are shown.
+            Produces output similar to
+            <programlisting>
+WHAT        WHERE                    MOUNTED IDLE TIMEOUT UNIT
+-           /mnt/test                no      120s         mnt-test.automount
```
keszybz: Shouldn't we query the .mount unit and show WHAT for unmounted units too?

keszybz (commented):
Looks nice. Can you also add the new verb to shell completions? It's similar to `list-units`, so hopefully it won't be too much trouble.

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,205 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
...
+
+        r = bus_get_property_trivial(bus, &locator, "TimeoutIdleUSec", &error, 't', &timeout_idle_usec);
+        if (r < 0)
+                return log_error_errno(r, "Failed to get idle timeout: %s", bus_error_message(&error, r));
+
+        r = get_triggered_units(bus, info->unit_path, &triggered);
+        if (r < 0)
+                return r;
+
+        n = get_unit_list(bus, NULL, triggered, &triggered_infos, 0, &reply);
```
poettering: you don't need to list the triggered units here, because automounts only trigger a single automount unit. Hence, add a safety check that `strv_length(triggered)` equals 1, and then you can use the single unit name in the list directly. And then you can use `unit_dbus_path_from_name()` to generate the dbus path for the unit directly client side, without any roundtrip. Thus you can get rid of the for loop below.

man/systemctl.xml
poettering: I think we should show the mount unit belonging to the automount unit at the very least.

dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Right. I've already rewritten this code to address @keszybz's comment.

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,196 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
 
         return r;
 }
+
+struct automount_info {
+        const char* machine;
+        const char* id;
+        char* what;
+        char* where;
```
yuwata: We usually place `*` at the beginning of the variable name: `const char *machine`.

yuwata (approved):
LGTM. One minor comment.

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,196 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
...
+                .what = TAKE_PTR(what),
+                .where = TAKE_PTR(where),
+                .timeout_idle_usec = timeout_idle_usec,
+                .mounted = streq_ptr(state, "active"),
+        };
+
+        return 0;
+}
+
+static int output_automounts_list(struct automount_info *infos, unsigned n_infos) {
```
poettering: please use `size_t` for `n_infos`. Objects in memory should always be measured in `size_t`. (other code in systemctl is a bit sloppy about this, and should be fixed too, see #24105)

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,196 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
...
+                        printf("Pass --all to see loaded but inactive automounts, too.\n");
+        }
+
+        return 0;
+}
+
+int verb_list_automounts(int argc, char *argv[], void *userdata) {
+        _cleanup_(message_set_freep) Set *replies = NULL;
+        _cleanup_strv_free_ char **machines = NULL;
+        _cleanup_strv_free_ char **automounts = NULL;
```
poettering: merge these two lines?

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,196 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
...
+        _cleanup_free_ UnitInfo *unit_infos = NULL;
+        _cleanup_free_ struct automount_info *automount_infos = NULL;
+        unsigned cs = 0;
```
poettering: size_t as above

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,196 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
...
+        if (r < 0)
+                return log_error_errno(r, "Failed to get idle timeout: %s", bus_error_message(&error, r));
+
+        r = unit_name_from_path(where, ".mount", &mount);
+        if (r < 0)
+                return r;
+
+        mount_path = unit_dbus_path_from_name(mount);
+        if (!mount_path)
+                return -ENOMEM;
```
poettering: log_oom()

src/systemctl/systemctl-list-units.c
```diff
@@ -760,3 +761,196 @@ int verb_list_timers(int argc, char *argv[], void *userdata) {
...
+        if (r < 0)
+                return log_error_errno(r, "Failed to get automount target: %s", bus_error_message(&error, r));
+
+        r = bus_get_property_trivial(bus, &locator, "TimeoutIdleUSec", &error, 't', &timeout_idle_usec);
+        if (r < 0)
+                return log_error_errno(r, "Failed to get idle timeout: %s", bus_error_message(&error, r));
+
+        r = unit_name_from_path(where, ".mount", &mount);
+        if (r < 0)
+                return r;
```
poettering: needs to error log. 

(i.e. we want that every error is logged exactly once. This means either functions need to log about all errors themselves, or about none, and then propagate an error code that the caller can then handle/log about. Usually functions that are "main-program-like" do the former, and functions that are "library-like" do the latter, but things are blurry, and exceptions are OK. Key though is that you decide for each function what type of the two it is. This function otherwise logs about all errors, hence you should log here too, sticking to the rule)

poettering:
looks good, just some nitpicking on types and logging.

src/systemctl/systemctl-list-units.c
dtardon: Fixed.

src/systemctl/systemctl-list-units.c
dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Fixed.

src/systemctl/systemctl-list-units.c
dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Fixed.

src/systemctl/systemctl-list-units.c
```diff
@@ -352,6 +352,8 @@ static int output_sockets_list(struct socket_info *socket_infos, size_t cs) {
         _cleanup_(table_unrefp) Table *table = NULL;
         int r;
 
+        assert(socket_infos || cs == 0);
```
poettering, line 367: nice, this works too, around the static checker logic

bluca (approved):

PR #24038, opened 2022-07-15
udev: enable BLKID_SUBLKS_FSINFO tags

alberand:
libblkid gained new tags - FSSIZE, FSLASTBLOCK and FSBLOCKSIZE.
These tags are filesystem related properties probed from superblock.
All of them are enabled by BLKID_SUBLKS_FSINFO flag.

Set the flag to allow these tags to be cached in udev db.

CC: @karelzak

### DIFF
--- a/src/udev/udev-builtin-blkid.c
+++ b/src/udev/udev-builtin-blkid.c
@@ -23,6 +23,7 @@
 #include "gpt.h"
 #include "parse-util.h"
 #include "string-util.h"
+#include "strv.h"
 #include "strxcpyx.h"
 #include "udev-builtin.h"
 
@@ -57,6 +58,10 @@ static void print_property(sd_device *dev, bool test, const char *name, const ch
                 udev_builtin_add_property(dev, test, "ID_FS_LABEL", s);
                 blkid_encode_string(value, s, sizeof(s));
                 udev_builtin_add_property(dev, test, "ID_FS_LABEL_ENC", s);
+
+        } else if (STR_IN_SET(name, "FSSIZE", "FSLASTBLOCK", "FSBLOCKSIZE")) {
+                strscpyl(s, sizeof(s), "ID_FS_", name + 2, NULL);
+                udev_builtin_add_property(dev, test, s, value);
 
         } else if (streq(name, "PTTYPE")) {
                 udev_builtin_add_property(dev, test, "ID_PART_TABLE_TYPE", value);
@@ -293,6 +298,9 @@ static int builtin_blkid(sd_device *dev, sd_netlink **rtnl, int argc, char *argv
         blkid_probe_set_superblocks_flags(pr,
                 BLKID_SUBLKS_LABEL | BLKID_SUBLKS_UUID |
                 BLKID_SUBLKS_TYPE | BLKID_SUBLKS_SECTYPE |
+#ifdef BLKID_SUBLKS_FSINFO
+                BLKID_SUBLKS_FSINFO |
+#endif
                 BLKID_SUBLKS_USAGE | BLKID_SUBLKS_VERSION);
 
         if (noraid)


src/udev/udev-builtin-blkid.c
```diff
@@ -58,6 +58,10 @@ static void print_property(sd_device *dev, bool test, const char *name, const ch
                 blkid_encode_string(value, s, sizeof(s));
                 udev_builtin_add_property(dev, test, "ID_FS_LABEL_ENC", s);
 
+        } else if (startswith(name, "FS")) {
+                strscpyl(s, sizeof(s), "ID_FS_", name, NULL);
```
yuwata: Maybe, it is better to drop `FS` prefix: e.g. `ID_FS_FSSIZE` -> `ID_FS_SIZE`.

yuwata (commented):
Please drop Signed-off-by tag from the commit message.

src/udev/udev-builtin-blkid.c
alberand: Signed-off-by tag dropped. Changed to `name + 2`. Initially I thought about it but left it, to exactly correspond to libblkid names, but that looks better.

yuwata (approved):
Superficially LGTM. Let's wait for other reviewers.

meson.build
```diff
@@ -1214,6 +1214,8 @@ if want_blkid != 'false' and not skip_deps
 
         conf.set10('HAVE_BLKID_PROBE_SET_HINT',
                    have and cc.has_function('blkid_probe_set_hint', dependencies : libblkid))
+        conf.set10('HAVE_BLKID_SUBLKS_FSINFO',
+                   have and cc.has_header_symbol('blkid.h', 'BLKID_SUBLKS_FSINFO', dependencies : libblkid))
```
poettering: hm, BLKID_SUBLKS_FSINFO is a macro anyway, so you check for existance of a macro and then set a macro as result? why bother? why not check for the macro in the code directly, via `#ifdef BLKID_SUBLKS_FSINFO`?

what am I missing?

src/udev/udev-builtin-blkid.c
```diff
@@ -58,6 +58,10 @@ static void print_property(sd_device *dev, bool test, const char *name, const ch
                 blkid_encode_string(value, s, sizeof(s));
                 udev_builtin_add_property(dev, test, "ID_FS_LABEL_ENC", s);
 
+        } else if (startswith(name, "FS")) {
```
poettering: that looks like a pretty broad match... i.e. the word "FSALT" would match this, even though this is not really what is meant here...

The other case where we use a startswith() match here checks for a prefix ending in an underscore, which makes clear we check for a whole namespace (as that's how we do namespacing in udev props/env vars: via underscore). But this check here is super duper generic, just two alphabetic chars...

hence, given you know which fields you want to support here, why not list them here, i.e. use `STR_IN_SET(name, "FSSIZE", "FLSLASTBLOCK", "FSNLOCKSIZE")`?

poettering:
what's the usecase for this, btw?

(PR looks conceptually OK, I am just curious)

meson.build
alberand: Yes, you absolutely right. No reason for that. Haven't questioned myself on why to do it this way. Changed to just `#ifdef BLKID_SUBLKS_FSINFO`, revert changes in `meson.build`.

src/udev/udev-builtin-blkid.c
alberand: Sure. Initially, I decided to check only for `FS` in case any new tags pop-up. But I agree, it's too general. Changed to suggested macro.

alberand:
> what's the usecase for this, btw?
> 
> (PR looks conceptually OK, I am just curious)

Someone wanted to use those tags from the cache (not familiar with particular usecase; something with lvm). `blkid` has cache, but according to Karel it should not be used and is kinda deprecated. The other option is `libudev` which will also allow using those tags as non-root and expose them in `lsblk`.

yuwata (approved):

src/udev/udev-builtin-blkid.c
```diff
@@ -293,6 +298,9 @@ static int builtin_blkid(sd_device *dev, sd_netlink **rtnl, int argc, char *argv
         blkid_probe_set_superblocks_flags(pr,
                 BLKID_SUBLKS_LABEL | BLKID_SUBLKS_UUID |
                 BLKID_SUBLKS_TYPE | BLKID_SUBLKS_SECTYPE |
+#if BLKID_SUBLKS_FSINFO
```
yuwata: Some CIs seem to dislike this. How about to use `#if defined(...)`?

src/udev/udev-builtin-blkid.c
```diff
@@ -293,6 +298,9 @@ static int builtin_blkid(sd_device *dev, sd_netlink **rtnl, int argc, char *argv
         blkid_probe_set_superblocks_flags(pr,
                 BLKID_SUBLKS_LABEL | BLKID_SUBLKS_UUID |
                 BLKID_SUBLKS_TYPE | BLKID_SUBLKS_SECTYPE |
+#if defined(BLKID_SUBLKS_FSINFO)
```
poettering: we more often spell this as `#ifdef …` instead of `#if defined(…)`, but I guess this works too

alberand: No problem, will change to `#ifdef` to stay consistent

src/udev/udev-builtin-blkid.c
alberand: Thanks! Fixed with `#ifdef` as suggested below.

src/udev/udev-builtin-blkid.c
yuwata: > we more often spell this as `#ifdef …` instead of `#if defined(…)`, but I guess this works too

@poettering I tried to count `#ifdef` and `#if defined`, but our code seem to have much `#if defined` than `#ifdef`. What's the difference and why `#ifdef` is a preferable way? Though, I do not have strong opinion about that. That's just a pure question.

poettering: We usually use `#if defined(…)` if we have more complex expressions, no? (because you have to there...) otherwise it's more `#ifdef`...

poettering: but i don't think this really matters, at all

PR #24043, opened 2022-07-16
udev: make udev worker valgrind friendly

yuwata:


src/libsystemd/sd-event/sd-event.c
```diff
@@ -4306,7 +4309,7 @@ _public_ int sd_event_exit(sd_event *e, int code) {
         assert_return(e, -EINVAL);
         assert_return(e = event_resolve(e), -ENOPKG);
         assert_return(e->state != SD_EVENT_FINISHED, -ESTALE);
-        assert_return(!event_pid_changed(e), -ECHILD);
+        /* This can be called by a forked child. Hence, no event_pid_changed(e) check here. */
```
poettering, line 4312: I am not convinced this is really something we should do. 

fork() is a mess, there is simply no strategy to release all resources allocated in the parent once your fork()ed in the child. The general approach is to just accept the "leakage". I don't think we should try to be better.

Note that fork()ing a child that does more than an immediate execve() is kinda on life-support on Linux. In fact, much of the stuff we do in udev workers (such as NSS lookups, i.e. resolving users/groups) is actually not allowed by glibc, yet we do it anyway (we get away with it because we do NSS never in the parent, and only in the child, hence NSS state doesn'get get cloned at fork(), because there is none at that point.)

So I am pretty sure we shouldn't try to make fork()s "work" in context of resource leaks. The worker processes we fork off shouldn't live for a long time, and should conceptually be "throw-away", i.e. even if something isn't cleaned up it doesn't matter since the processes have a finite lifetime and then things are automatically cleaned up.

So, I added these `event_pid_changed()` checks to sd-event and other APis precisely to make any attempts to use or even just free an object inherited across a fork() in the child because that cannot work safely. This actually matters IRL: if the child would be able to destroy an sd-event object, including its epoll, and first deregister the event sources from the epoll, then it would break the parent's epoll fd, because the epoll fd is shared across the fork(), even if the memory of the sd-event is not!

Hence, let's not try to make the worker processes valgrind clean, it cannot really work. 

Sorry, if that's disappointing.

poettering: (if we wanted to do things "cleanly" and "correctly", then we'd split out the udev worker code into separate binaries, and then fork of worker processes that first execve() these binaries. That way the worker processes would be valgrind clean, we'd have no OOM trap, and not inherit a deep stack frame or half set up memory from the parent. I am wonder though what the performance impact of that would be)

mrc0mmand: I guess an alternative could be utilizing Valgrind's dynamic options: https://valgrind.org/docs/manual/manual-core.html#manual-core.dynopts

and calling something like
```c
#if HAVE_VALGRIND_VALGRIND_H
        if (RUNNING_ON_VALGRIND)
                VALGRIND_CLO_CHANGE("--leak-check=no");
#endif
```

in the worker process after fork. But I've never used this feature before, so it's all theoretical...

poettering: hmm, i wonder if there's a way to say in code "everything allocated before this point should not be considered a leak under any circumstances" or so. i.e. like a VALGRIND_RESET_LEAK_ACCOUNTING or so

mrc0mmand: Sadly, it doesn't seem like there's such option. I went through the `valgrind.h` and `memcheck.h` headers (which are very well documented, kudos for that), and there's nothing which would help us here. You can suppress errors (like an invalid access) via `VALGRIND_{ENABLE,DISABLE}_ADDR_ERROR_REPORTING_IN_RANGE(x, y);` but that doesn't apply here.

Using `VALGRIND_CLO_CHANGE("--leak-check=no");` appears to be the only solution so far, and a quick test shows that it seems to be doing the right thing (i.e. disabling Valgrind's leak check in a specific process). But it's definitely not ideal.

Another option would be a lot of if-def-ery to contain the "icky" changes to valgrind-enabled binaries only, but that sounds even worse.

PR #24044, opened 2022-07-16
Add a configuration option for setting default device timeout

dtardon:


src/core/main.c
```diff
@@ -627,6 +640,7 @@ static int parse_config_file(void) {
                 { "Manager", "DefaultTimeoutStartSec",       config_parse_sec,                   0,                        &arg_default_timeout_start_usec   },
                 { "Manager", "DefaultTimeoutStopSec",        config_parse_sec,                   0,                        &arg_default_timeout_stop_usec    },
                 { "Manager", "DefaultTimeoutAbortSec",       config_parse_default_timeout_abort, 0,                        NULL                              },
+                { "Manager", "DefaultDeviceTimeoutSec",      config_parse_sec, 0,                        &arg_default_device_timeout_usec  },
```
yuwata: Please align the columns.

yuwata (commented):
I think it is better to add a udev property to set timeout, if it works.

yuwata:
Applying longer timeout to all device units sounds not good to me. How about to create a drop-in config to set JobTimeoutSec=?

dtardon:
> Applying longer timeout to all device units sounds not good to me. How about to create a drop-in config to set JobTimeoutSec=?

That feels more like a workaround than a fix. Specifically:

1. It is not discoverable.
2. `JobTimeoutSec=` is IMHO an impl. detail (but `TimeoutStartSec=` should work too? That would be a better choice.)
3. I think the drop-in would likely be created for all devices anyway, because most admins would not bother constructing the right prefix to match just block devices.

man/kernel-command-line.xml
```diff
@@ -408,6 +408,16 @@
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>systemd.default_device_timeout_sec=</varname></term>
```
poettering: do we really this is as kernel cmdline arg? I mean, I am not totally opposed. But I am wondering: do you need this, do you want to actually make use of this? Because, if this is just for "completeness", I am not convinced, then we shouldn't have it, and maybe add later when somebody actually asks.

Having less knobs exposed is a good thing, hence let's try to be more minimal with these things.

poettering:
So this is redundant of course, given you could also just add a .device drop-in to change that. But then again it is a bit weird that so far we just copy the start time-out into the device time-out. 

So maybe, yeah, this might be ok to do.

man/kernel-command-line.xml
dtardon: Both have been asked for, but without any explanation why the cmdline option is also needed. I can think of only one reason for it myself: if one faces a system that doesn't boot because of block device timeouts, it's handy to have an easy way to get it to boot (and then configure it properly). I'm not sure if that's good enough reason though.

dtardon: I've dropped the cmdline option.

man/systemd-system.conf.xml
```diff
@@ -370,6 +370,18 @@
...
 
+      <varlistentry>
+        <term><varname>DefaultDeviceTimeoutSec=</varname></term>
+
+        <listitem><para>Configures the default timeout for waiting for devices. It can only be changed per a device via
+        the <varname>x-systemd.device-timeout=</varname> option in <filename>/etc/fstab</filename> and
+        <filename>/etc/crypttab</filename> (see
+        <citerefentry><refentrytitle>systemd.mount</refentrytitle><manvolnum>5</manvolnum></citerefentry>,
+        <citerefentry><refentrytitle>crypttab</refentrytitle><manvolnum>5</manvolnum></citerefentry>). It defaults to
+        90s.
```
poettering: that's not true. It can also be changed by doing a .device.d/ drop-in which sets JobTimeoutSec=

poettering: also, drop the "only". That's an unnecessary info that restricts what we can change later on.

poettering:
looks good, but docs need updating

PR #24051, opened 2022-07-17
json: use fpclassify() macro and refuse -ffinite-math-only

### LINKED ISSUE #23610: 251.2: test suite is failing in 3 units
Looks like 3 units of the test suite are failing.
Just in case I'm building and testing systemd inside LXC zone on x86/64.

meson setup:
```spec
%meson \
        -D acl=true \
        -D adm-gid=4 \
        -D apparmor=false \
        -D audio-gid=63 \
        -D audit=true \
        -D blkid=true \
        -D b_ndebug=false \
        -D bpf-framework=true \
        -D bzip2=false \
        -D cdrom-gid=11 \
        -D compat-mutable-uid-boundaries=true \
        -D cryptolib=openssl \
        -D default-dns-over-tls=no \
        -D default-dnssec=no \
        -D default-kill-user-processes=false \
        -D default-llmnr=resolve \
        -D default-locale=C \
        -D default-mdns=no \
        -D dev-kvm-mode=0666 \
        -D dialout-gid=18 \
        -D disk-gid=6 \
        -D dns-servers= \
        -D efi=true \
        -D elfutils=true \
        -D fallback-hostname=g2v \
        -D fdisk=true \
        -D gcrypt=true \
        -D gnu-efi=true \
        -D gnutls=false \
        -D hwdb=true \
        -D ima=true \
        -D input-gid=104 \
        -D install-tests=true \
        -D kmem-gid=9 \
        -D kmod=true \
        -D kvm-gid=36 \
        -D libcryptsetup=false \
        -D libcurl=true \
        -D libfido2=true \
        -D libidn2=true \
        -D libiptc=true \
        -D lp-gid=7 \
        -D lz4=true \
        -D man=true \
        -D microhttpd=true \
        -D mode=release \
        -D mode=release \
        -D nobody-group=nobody \
        -D nobody-user=nobody \
        -D ntp-servers='0.%{ntpvendor}.pool.ntp.org 1.%{ntpvendor}.pool.ntp.org 2.%{ntpvendor}.pool.ntp.org 3.%{ntpvendor}.pool.ntp.org' \
        -D oomd=true \
        -D openssl=true \
        -D p11kit=true \
        -D pam=true \
        -D pcre2=true \
        -D polkit=true \
        -D pwquality=true \
        -D qrencode=true \
        -D rc-local=/etc/rc.d/rc.local \
        -D render-gid=105 \
        -D rpmmacrosdir=no \
        -D seccomp=true \
        -D selinux=true \
        -D service-watchdog= \
        -D sgx-gid=106 \
        -D smack=true \
        -D split-bin=true \
        -D split-usr=false \
        -D standalone-binaries=true \
        -D status-unit-format-default=combined \
        -D systemd-journal-gid=190 \
        -D systemd-network-uid=192 \
        -D systemd-resolve-uid=193 \
        -D sysusers=true \
        -D sysvinit-path=/etc/rc.d/init.d \
        -D tape-gid=33 \
        -D tests=unsafe \
        -D tpm2=true \
        -D tpm=true \
        -D tty-gid=5 \
        -D tty-gid=5 \
        -D user-path="/usr/bin:/usr/sbin" \
        -D users-gid=100 \
        -D users-gid=100 \
        -D utmp-gid=22 \
        -D video-gid=39 \
        -D wheel-gid=10 \
        -D xkbcommon=true \
        -D xz=true \
        -D zlib=true \
        -D zstd=true \
```

yuwata:
Closes #23610.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -390,6 +390,11 @@ possible_common_cc_flags = [
 
 c_args = get_option('c_args')
 
+# Our json library does not support -ffinite-math-only, which is enabled by -Ofast or -ffast-math.
+if (('-Ofast' in c_args or '-ffast-math' in c_args or '-ffinite-math-only' in c_args) and not '-fno-finite-math-only' in c_args)
+        error('-Ofast, -ffast-math, or -ffinite-math-only is specified in c_args.')
+endif
+
 # Disable -Wmaybe-uninitialized when compiling with -Os/-O1/-O3/etc. There are
 # too many false positives with gcc >= 8. Effectively, we only test with -O0
 # and -O2; this should be enough to catch most important cases without too much


meson.build
```diff
@@ -390,6 +390,21 @@ possible_common_cc_flags = [
 
 c_args = get_option('c_args')
 
+# Our json library does not support -ffinite-math-only, which is enabled by -Ofast or -ffast-math.
+# Refuse to build in that case.
+have = false
+foreach a : c_args
+        if a == '-Ofast' or a == '-ffast-math' or a == '-ffinite-math-only'
```
medhefgo:
```suggestion
        if a in ['-Ofast', '-ffast-math', '-ffinite-math-only']
```

This also fails to detect a later override with `-O2` for example. (Personally, I think this loop is just overkill.)

src/shared/json.c
```diff
@@ -253,9 +253,7 @@ static JsonVariant *json_variant_formalize(JsonVariant *v) {
                 return json_variant_unsigned(v) == 0 ? JSON_VARIANT_MAGIC_ZERO_UNSIGNED : v;
 
         case JSON_VARIANT_REAL:
-                DISABLE_WARNING_FLOAT_EQUAL;
-                return json_variant_real(v) == 0.0 ? JSON_VARIANT_MAGIC_ZERO_REAL : v;
-                REENABLE_WARNING;
+                return iszero_relaxed(json_variant_real(v)) ? JSON_VARIANT_MAGIC_ZERO_REAL : v;
```
keszybz: No, this change is wrong. The fact that a floating-point number has subnormal representation is an internal detail of the implementation of the fp standard. It is exposed because it has effect on precision, but it is not something that we care about.

The warning is stupid, and we need to keep disabling it.

src/shared/json.c
```diff
@@ -352,16 +350,17 @@ int json_variant_new_real(JsonVariant **ret, double d) {
...
                 return 0;
-        }
-        REENABLE_WARNING;
 
-        /* JSON doesn't know NaN, +Infinity or -Infinity. Let's silently convert to 'null'. */
-        if (isnan(d) || isinf(d)) {
-                *ret = JSON_VARIANT_MAGIC_NULL;
+        case FP_ZERO:
+        case FP_SUBNORMAL:
+                *ret = JSON_VARIANT_MAGIC_ZERO_REAL;
```
keszybz: Here too.

meson.build
```diff
@@ -390,6 +390,21 @@ possible_common_cc_flags = [
...
+                have = true
+        endif
+        if a == '-fno-finite-math-only'
+                have = false
+        endif
+endforeach
+if have
+        error('-Ofast, -ffast-math, or -ffinite-math-only is specified in c_args.')
+endif
```
keszybz, line 396: Maybe write this as:
```python
if '-Ofast' in c_args or '-ffast-math' in c_args or '-ffinite-math-only' in c_args:
    error(…)
endif
```
I don't think we should spend more than three lines on this check.

keszybz (changes requested):

yuwata:
@medhefgo and @keszybz Thank you for your review.
- subnormal values are handled as is. as previously we did.
- simplified c_args check.

PTAL.

src/basic/math-util.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <math.h>
+
+#include "macro.h"
+
+/* On some optimization level, iszero(x) is converted to (x == 0.0), and emits warning -Wfloat-equal. */
+#define iszero_safe(x) (fpclassify(x) == FP_ZERO)
+
+/* To avoid x == y and triggering compile warning -Wfloat-equal. */
+#define fp_equal(x, y) iszero_safe(x - y)
```
poettering: is this safe? i.e. what if x minus y actually leaves the defined range for double? Does the comparison still work?

at the very least x and y should be enclosed in `()`

poettering:
looks generally ok – if fp_equal() actually works for the full double range... needs some test cases with DBL_MIN/DBL_MAX in various combinations. As well as tests when comparing huge values with values very close but not equal to zero.

src/basic/math-util.h
yuwata: `x - y` is always in double (of course when at least one is double), but may be NaN or infinity. But in that case, this should returns false, and actually.

The missing `()` are added.

yuwata:
@poettering Thank you for the comments. More test cases are added. PTAL.

src/basic/math-util.h
```diff
@@ -0,0 +1,14 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+#pragma once
+
+#include <math.h>
+
+#include "macro.h"
+
+/* On some optimization level, iszero(x) is converted to (x == 0.0), and emits warning -Wfloat-equal.
+ * The argument must be a floating point, i.e. one of float, double, or long double. */
+#define iszero_safe(x) (fpclassify(x) == FP_ZERO)
```
keszybz, line 10: Looking at the disassembled code, for numbers that are not infinity and not nan, gcc turns this into a normal comparison and there's no function call. So this seems reasonable.

keszybz (commented):
LGTM.

PR #24063, opened 2022-07-19
mkosi: Silence gdb debuginfo messages/prompts

daandemeyer:
Let's silence gdb asking about debuginfod and complaining about
missing debuginfo to reduce friction when using mkosi to work
on systemd.

### DIFF
--- a/mkosi.postinst
+++ b/mkosi.postinst
@@ -6,6 +6,11 @@ if [ "$1" = "final" ]; then
         bootctl install
     fi
 
+    cat >> /root/.gdbinit <<EOF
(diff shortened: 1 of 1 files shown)


bluca (approved):

mkosi.postinst
```diff
@@ -6,6 +6,11 @@ if [ "$1" = "final" ]; then
         bootctl install
     fi
 
+    cat > /root/.gdbinit <<EOF
```
bluca: maybe `>>` in case the user has something local in the skeleton defined?

PR #24067, opened 2022-07-20
tree-wide: Introduce free_and_replace_full()

daandemeyer:
Let's have all our free_and_replace() functions use a single
implementation.

### DIFF
--- a/src/basic/alloc-util.h
+++ b/src/basic/alloc-util.h
@@ -50,15 +50,18 @@ typedef void* (*mfree_func_t)(void *p);
 
 #define malloc0(n) (calloc(1, (n) ?: 1))
 
-#define free_and_replace(a, b)                  \
+#define free_and_replace_full(a, b, free_func)  \
         ({                                      \
                 typeof(a)* _a = &(a);           \
                 typeof(b)* _b = &(b);           \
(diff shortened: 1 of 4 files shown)


bluca (approved):

mrc0mmand:
Hmpf, looking through logs of other jobs, the C8S job might be a bit FUBAR, so I'll delve a bit deeper...

medhefgo:
Why do they return 0 instead of NULL?

DaanDeMeyer:
> Why do they return 0 instead of NULL?

No idea, we don't even use the return value anywhere, let's see if I can get rid of the statement expression altogether.

src/basic/alloc-util.h
```diff
@@ -50,7 +50,7 @@ typedef void* (*mfree_func_t)(void *p);
 
 #define malloc0(n) (calloc(1, (n) ?: 1))
 
-#define free_and_replace(a, b)                  \
+#define free_and_replace_full(a, b, free)       \
```
yuwata: for safety, I'd like to rename `free` -> `free_func` or `_free`.

yuwata (approved):
LGTM. One minor suggestion.

PR #24071, opened 2022-07-21
path/timer units: export env vars and D-Bus properties with reason that triggered the job

bluca:
Inform the triggered unit of what started it. The inotify path that woke us up in case of a path unit, or the name of the timer unit in case of a timer. All best-effort and documented as such, as there could be multiple triggers firing at the same time, and we'll always only report the one that happened to start the activation job.

### DIFF
--- a/man/systemd.exec.xml
+++ b/man/systemd.exec.xml
@@ -3698,8 +3698,10 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
         <varlistentry>
           <term><varname>$TRIGGER_UNIT</varname></term>
           <term><varname>$TRIGGER_PATH</varname></term>
+          <term><varname>$TRIGGER_TIMER_REALTIME_USEC</varname></term>
+          <term><varname>$TRIGGER_TIMER_MONOTONIC_USEC</varname></term>
 
-          <listitem><para>If the unit was activated dynamically (e.g.: a corresponding path unit), the
+          <listitem><para>If the unit was activated dynamically (e.g.: a corresponding path unit or timer unit), the
           unit that triggered it and other type-dependent information will be passed via these variables. Note that
           this information is provided in a best-effort way. For example, multiple triggers happening one after
           another will be coalesced and only one will be reported, with no guarantee as to which one it will be.
--- a/man/systemd.timer.xml
+++ b/man/systemd.timer.xml
@@ -366,6 +366,10 @@
 
   <refsect1>
       <title>See Also</title>
+      <para>Environment variables with details on the trigger will be set for triggered units. See the
+      <literal>Environment Variables Set on Triggered Units</literal> section in
+      <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>1</manvolnum></citerefentry>
+      for more details.</para>
       <para>
         <citerefentry><refentrytitle>systemd</refentrytitle><manvolnum>1</manvolnum></citerefentry>,
         <citerefentry><refentrytitle>systemctl</refentrytitle><manvolnum>1</manvolnum></citerefentry>,
--- a/src/core/timer.c
+++ b/src/core/timer.c
@@ -576,8 +576,10 @@ fail:
 }
 
 static void timer_enter_running(Timer *t) {
+        _cleanup_(activation_details_unrefp) ActivationDetails *details = NULL;
         _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
         Unit *trigger;
+        Job *job;
         int r;
 
         assert(t);
@@ -593,11 +595,20 @@ static void timer_enter_running(Timer *t) {
                 return;
         }
 
-        r = manager_add_job(UNIT(t)->manager, JOB_START, trigger, JOB_REPLACE, NULL, &error, NULL);
+        details = activation_details_new(UNIT(t));
+        if (!details) {
+                r = -ENOMEM;
+                goto fail;
+        }
+
+        r = manager_add_job(UNIT(t)->manager, JOB_START, trigger, JOB_REPLACE, NULL, &error, &job);
         if (r < 0)
                 goto fail;
 
         dual_timestamp_get(&t->last_trigger);
+        ACTIVATION_DETAILS_TIMER(details)->last_trigger = t->last_trigger;
+
+        job_set_activation_details(job, details);
 
         if (t->stamp_path)
                 touch_file(t->stamp_path, true, t->last_trigger.realtime, UID_INVALID, GID_INVALID, MODE_INVALID);
@@ -892,6 +903,91 @@ static int timer_can_start(Unit *u) {
         return 1;
 }
 
+static void activation_details_timer_serialize(ActivationDetails *details, FILE *f) {
+        ActivationDetailsTimer *t = ACTIVATION_DETAILS_TIMER(details);
+
+        assert(details);
+        assert(f);
+        assert(t);
+
+        (void) serialize_dual_timestamp(f, "activation-details-timer-last-trigger", &t->last_trigger);
+}
+
+static int activation_details_timer_deserialize(const char *key, const char *value, ActivationDetails **details) {
+        int r;
+
+        assert(key);
+        assert(value);
+
+        if (!details || !*details)
+                return -EINVAL;
+
+        ActivationDetailsTimer *t = ACTIVATION_DETAILS_TIMER(*details);
+        if (!t)
+                return -EINVAL;
+
+        if (!streq(key, "activation-details-timer-last-trigger"))
+                return -EINVAL;
+
+        r = deserialize_dual_timestamp(value, &t->last_trigger);
+        if (r < 0)
+                return r;
+
+        return 0;
+}
+
+static int activation_details_timer_append_env(ActivationDetails *details, char ***strv) {
+        ActivationDetailsTimer *t = ACTIVATION_DETAILS_TIMER(details);
+        int r;
+
+        assert(details);
+        assert(strv);
+        assert(t);
+
+        if (!dual_timestamp_is_set(&t->last_trigger))
+                return 0;
+
+        r = strv_extendf(strv, "TRIGGER_TIMER_REALTIME_USEC=%" USEC_FMT, t->last_trigger.realtime);
+        if (r < 0)
+                return r;
+
+        r = strv_extendf(strv, "TRIGGER_TIMER_MONOTONIC_USEC=%" USEC_FMT, t->last_trigger.monotonic);
+        if (r < 0)
+                return r;
+
+        return 2; /* Return the number of variables added to the env block */
+}
+
+static int activation_details_timer_append_pair(ActivationDetails *details, char ***strv) {
+        ActivationDetailsTimer *t = ACTIVATION_DETAILS_TIMER(details);
+        int r;
+
+        assert(details);
+        assert(strv);
+        assert(t);
+
+        if (!dual_timestamp_is_set(&t->last_trigger))
+                return 0;
+
+        r = strv_extend(strv, "trigger_timer_realtime_usec");
+        if (r < 0)
+                return r;
+
+        r = strv_extendf(strv, "%" USEC_FMT, t->last_trigger.realtime);
+        if (r < 0)
+                return r;
+
+        r = strv_extend(strv, "trigger_timer_monotonic_usec");
+        if (r < 0)
+                return r;
+
+        r = strv_extendf(strv, "%" USEC_FMT, t->last_trigger.monotonic);
+        if (r < 0)
+                return r;
+
+        return 2; /* Return the number of pairs added to the env block */
+}
+
 static const char* const timer_base_table[_TIMER_BASE_MAX] = {
         [TIMER_ACTIVE]        = "OnActiveSec",
         [TIMER_BOOT]          = "OnBootSec",
@@ -953,4 +1049,13 @@ const UnitVTable timer_vtable = {
         .bus_set_property = bus_timer_set_property,
 
         .can_start = timer_can_start,
+};
+
+const ActivationDetailsVTable activation_details_timer_vtable = {
+        .object_size = sizeof(ActivationDetailsTimer),
+
+        .serialize = activation_details_timer_serialize,
+        .deserialize = activation_details_timer_deserialize,
+        .append_env = activation_details_timer_append_env,
+        .append_pair = activation_details_timer_append_pair,
 };
--- a/src/core/unit.c
+++ b/src/core/unit.c
@@ -5931,6 +5931,7 @@ int unit_get_dependency_array(const Unit *u, UnitDependencyAtom atom, Unit ***re
 
 const ActivationDetailsVTable * const activation_details_vtable[_UNIT_TYPE_MAX] = {
         [UNIT_PATH] = &activation_details_path_vtable,
+        [UNIT_TIMER] = &activation_details_timer_vtable,
 };
 
 ActivationDetails *activation_details_new(Unit *trigger_unit) {
--- a/src/core/timer.h
+++ b/src/core/timer.h
@@ -2,6 +2,7 @@
 #pragma once
 
 typedef struct Timer Timer;
+typedef struct ActivationDetailsTimer ActivationDetailsTimer;
 
 #include "calendarspec.h"
 #include "unit.h"
@@ -64,11 +65,17 @@ struct Timer {
         char *stamp_path;
 };
 
+struct ActivationDetailsTimer {
+        ActivationDetails meta;
+        dual_timestamp last_trigger;
+};
+
 #define TIMER_MONOTONIC_CLOCK(t) ((t)->wake_system ? CLOCK_BOOTTIME_ALARM : CLOCK_MONOTONIC)
 
 void timer_free_values(Timer *t);
 
 extern const UnitVTable timer_vtable;
+extern const ActivationDetailsVTable activation_details_timer_vtable;
 
 const char *timer_base_to_string(TimerBase i) _const_;
 TimerBase timer_base_from_string(const char *s) _pure_;
@@ -77,3 +84,4 @@ const char* timer_result_to_string(TimerResult i) _const_;
 TimerResult timer_result_from_string(const char *s) _pure_;
 
 DEFINE_CAST(TIMER, Timer);
+DEFINE_ACTIVATION_DETAILS_CAST(ACTIVATION_DETAILS_TIMER, ActivationDetailsTimer, TIMER);
--- a/test/units/testsuite-07.sh
+++ b/test/units/testsuite-07.sh
@@ -8,6 +8,9 @@ set -o pipefail
 cat >/lib/systemd/system/my.service <<EOF
 [Service]
 Type=oneshot
+ExecStartPre=sh -c 'test "\$TRIGGER_UNIT" = my.timer'
+ExecStartPre=sh -c 'test -n "\$TRIGGER_TIMER_REALTIME_USEC"'
+ExecStartPre=sh -c 'test -n "\$TRIGGER_TIMER_MONOTONIC_USEC"'
 ExecStart=/bin/echo Timer runs me
 EOF
 


yuwata:
Why not about .socket units?

bluca:
> Why not about .socket units?

Those already pass down the socket file descriptor, which can be queried for the address, which is much more useful

man/systemd.exec.xml
```diff
@@ -3690,6 +3690,15 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
           system.</para></listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><varname>$TRIGGER_REASON</varname></term>
+
+          <listitem><para>If the unit was activated dynamically (e.g.: a corresponding path unit), the reason
```
poettering: "the" is a bit misleading, because there might not be "the" reason, but many. hence say, "a" reason?

man/systemd.exec.xml
```diff
@@ -3690,6 +3690,15 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
...
+          for the activation will be listed in this variable. Note that this is provided in a best-effort way.
```
poettering: "listed"? it's just one, so maybe "passed"?

man/systemd.exec.xml
```diff
@@ -3690,6 +3690,15 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
...
+          For example, there might be multiple triggers firing at any given time, and only one will be reported,
```
poettering: I think it also matters to say that all kinds of triggering events can be queued and colaesced in the queue. So it's not "firing at any given time", it's also "coalesced if happening one after the other". if you follow what I mean

man/systemd.exec.xml
```diff
@@ -3690,6 +3690,15 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
...
+          and there is no guarantee on which one it will be.</para></listitem>
+        </varlistentry>
```
poettering, line 3711: maybe emphasize that this is *unreliable*. i.e. say

"Because of this, this variable is primarily informational, i.e. useful for debugging purposes, is lossy, and should not be relied upon to propagate a comprehensive reason for activation."

src/core/path.c
```diff
@@ -532,17 +536,24 @@ static void path_enter_running(Path *p) {
...
-static bool path_check_good(Path *p, bool initial, bool from_trigger_notify) {
+static int path_check_good(Path *p, bool initial, bool from_trigger_notify, char **ret_trigger_path) {
         assert(p);
+        assert(ret_trigger_path);
 
         LIST_FOREACH(spec, s, p->specs)
-                if (path_spec_check_good(s, initial, from_trigger_notify))
-                        return true;
+                if (path_spec_check_good(s, initial, from_trigger_notify)) {
+                        *ret_trigger_path = strdup(s->path);
```
poettering: as per coding style: do not clobber return values on failure. Here, if the allocation fails, you will set the returb parameter to NULL...

man/systemd.path.xml
```diff
@@ -206,6 +206,23 @@
...
+    <variablelist class='environment-variables'>
+
+      <varlistentry>
+        <term><varname>$TRIGGER_REASON</varname></term>
+
+        <listitem><para>The path that caused the unit to be triggered will be set in this environment variable
+        for the associated service unit being triggered. Note that this is provided in a best-effort way.
+        For example, there might be multiple triggers firing at any given time, and only one will be reported,
+        and there is no guarantee on which one it will be.</para></listitem>
+      </varlistentry>
```
poettering: I think this should be moved to the service man page, as that's where it is set.

src/core/job.c
```diff
@@ -1160,6 +1162,8 @@ int job_serialize(Job *j, FILE *f) {
 
         bus_track_serialize(j->bus_track, f, "subscribed");
 
+        (void) serialize_item(f, "trigger_reason", j->trigger_reason);
```
poettering: all our serialization keys use `-` instead of `_` for separating words in the name so far

man/systemd.exec.xml
bluca: I'll also qualify it, because if there's only one trigger configured, then it will be reliable, so that only applies when there are multiple ones

man/systemd.exec.xml
```diff
@@ -3690,6 +3690,15 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
...
+
       </variablelist>
```
poettering, line 3713: so, what if I have a unit that can be triggered by both path and timer? I'll get one TRIGGER_REASON and I'll not have a clue how to interpret TRIGGER_REASON in that case.

bluca: well it's either a path or a foo.timer, should be pretty easy to tell apart no?

man/systemd.path.xml
bluca: it already is in systemd.exec - but I think it's good to have it in the trigger manpage, given that's what you look at when using trigger units

poettering:
So I think this should be more fine grained and extensible. The thing is that people who trigger by .path probably want to know more than just the path, i.e. the type of event (create, remove, rename, …) sooner or later. So we really should have a way we can reasonably propagate that info, even if we won't care initially.

Moreover I think you want to propagate the triggering unit name always, i.e. also in the path unit case.

Which kinda suggests you want $TRIGGER_UNIT= which is always set, including for socket units, and always contains the name of a triggering unit. And then separately for that we'd have for the inotify thing TRIGGER_PATH_FILENAME, TRIGGER_PATH_EVENTTYPE and so on. (well, TRIGGRE_PATH_EVENTTYPE doesn't really have to exist right away, but I think we should have th option to).

I also would prefer if this is done a bit more generically. i.e. define TriggerEventInfo as a structure somewhere that carries the information for these three things (and possibly more later), and then allocate one of those once the info is available, and set it into the Job object. I think this would be nicer, as it conceptually separates jobs from the triggering event metainfo. 

then, make this structure refcounted maybe, and make the triggering unit add/ref it to the job, and then from the job add/ref it into the activated unit, once it is activated.

Finally, I think this info should be exposed on both the Job and the Unit d-bus objects then, as this is useful for debugging.

poettering:
(there's a TODO list item somewhere, which asked for also reporting info about dbus clients that manually started a service. I think we should make it easy to add that too, simply by then extending the TriggerEventInfo structure, with some data about the dbus client)

man/systemd.exec.xml
poettering: nah, a service unit can always be started manually instead of relying on triggers. Thus you *always* must expect two ways of activation even if you only have one trigger configured: manual + whatever you added.

man/systemd.path.xml
poettering: I think the explanations of the env vars really belongs in systemd.exec and there only. But yes, there hsould be a pointer from the .path + .timer man pages to that.

man/systemd.exec.xml
bluca: Right, but in that case there will be nothing set. So the env var will still be reliable.

man/systemd.exec.xml
poettering: that would mean you could never add this logic for .socket units, too. i.e. think about a .socket unit that has two Listen* lines for two unix sockets. It might be interesting to propagate via this new infra the info which of the two sockets caused us to activate the service. Yes, the daemon could look at the two fds and guess, if only one of them is busy, but this would never answer you the question: which was triggered first...

So given AF_UNIX sockets are typically in the fs, and thus a path, you'd then create ambiguity too. i.e. were you activated because of an inotify event or because of traffic incoming on the AF_UNIX socket that the path refers to.

poettering:
So thinking more about this, I have more questions.

First, what's your specific usecase, can you explain?

I am a bit concerned that people will start using this in ways, they better shouldn't, i.e. write a single service that does one thing if activated at as normal service, and something very different when activated via a timer (think: a web service that when started at boot is a web service, but when run via a timer cleans up logs or so). While we can't stop them, and shouldn't, we should try to communicate that is is a bad idea more strongly, i.e. more than in just docs. For example, I think we should call these variables "TRACE_xyz" or so, i.e. indicate that this is really about tracing/debugging, and that if they implement behaviour the way I described they are using unreliable tracing info to carry workloads...

Then, I think this should be though to be even more generic than what I propose above, i.e. we should keep the door open so that we can communicate information about *all* forms of activation through this. i.e. if this was caused by IPC calls, if this was caused by the initial boot transaction, or a shutdown transaction and so on. And if this was the main unit of activation, or if this is just a dependency of another unit that was activated.

Hence, this should probably not just be called TriggerEventInfo, but ActivationEventInfo, so that we can extend this later with information about *all* kind of invocations.

I think the thought in the background should be that we design this generic enough so that eventually we can implement a call "systemd-analyze explain" or so, that looks at all running units, and tells you in tabular fashion for every single one in some reasonable detail how it came to be activated, i.e. which unit triggered it, which unit pulled it in, if it was pulled in by IPC, or by initial/shutdown transaction, which IPC client was it, and so on.

Sorry, you opened a can of worms here. ;-)

(But to clarify, I don't think this PR should implement all of this. All I am asking for is that it sets the ground work, to make things conceptually and in code easy to extend to eventually propagate all that info).

poettering:
Hmm, one more thing: I think it's correct to just store one of the activation reasons instead of all. But I think we should enforce a clear regime on which one we store: info about the *first* event that made us enqueue the job, or info about the *most recent* event that made us enqueue the job. This in a way would become API (though I'd document it in a code comment for now only, not in actual docs).

Codewise this would probably mean: whenever we merge to jobs, we either need to keep the info from the existing one around, and ignore the one from the new job. Or we must drop the old, and move ove the one from the new job. This includes invalidating the info if the job we think should win carries no info.

poettering:
And I think it should be the oldest event we know about for this activation that we keep (simply because for socket/path units and similar once we see one event we stop looking for further events, for efficiency reasons, hence it's the more useful one)

bluca:
> First, what's your specific usecase, can you explain?

A service triggered by a path unit that watches multiple files, they want to know which one was the trigger

poettering:
> > 
> 
> A service triggered by a path unit that watches multiple files, they want to know which one was the trigger

but then you cannot distinguish the case "triggered manually" from "triggered by path on older systemd"...

bluca:
> > > 
> > 
> > 
> > A service triggered by a path unit that watches multiple files, they want to know which one was the trigger
> 
> but then you cannot distinguish the case "triggered manually" from "triggered by path on older systemd"...

Not a problem for my use case

poettering:
> A service triggered by a path unit that watches multiple files, they want to know which one was the trigger

But why?

Why can't it check the files to see what needs to be done?

bluca:
> > A service triggered by a path unit that watches multiple files, they want to know which one was the trigger
> 
> But why?
> 
> Why can't it check the files to see what needs to be done?

The watch is over files created in a directory, so they'd need to manually keep track of what's there and what's not and so on.

src/core/job.c
```diff
@@ -180,9 +182,15 @@ static void job_merge_into_installed(Job *j, Job *other) {
...
         assert(j->unit == other->unit);
 
-        if (j->type != JOB_NOP)
+        if (j->type != JOB_NOP) {
                 assert_se(job_type_merge_and_collapse(&j->type, other->type, j->unit) == 0);
-        else
+
+                /* Keep the oldest ActivationEventInfo, if any */
+                if (j->activation_event_info)
+                        activation_event_info_unref(other->activation_event_info);
```
poettering: no need to destroy the object, is there? it would be destroyed anyway once the other job is freed..

src/core/job.c
```diff
@@ -180,9 +182,15 @@ static void job_merge_into_installed(Job *j, Job *other) {
...
+                else
+                        j->activation_event_info = TAKE_PTR(other->activation_event_info);
```
poettering: here we should unref `j->activation_event_info` first, before we fill in the new data

src/core/service.c
```diff
@@ -1641,6 +1641,12 @@ static int service_spawn_internal(
                 }
         }
 
+        Job *j = UNIT(s)->job;
+        if (j && j->activation_event_info) {
+                if (j->activation_event_info->trigger_unit_name && asprintf(our_env + n_env++, "TRIGGER_UNIT=%s", j->activation_event_info->trigger_unit_name) < 0)
```
poettering: that's a bit long... (i.e. line break plz)

src/core/unit.h
```diff
@@ -106,6 +106,62 @@ typedef union UnitDependencyInfo {
...
+
+        /* This should free all type-specific variables. It should be idempotent. */
+        void (*done)(ActivationEventInfo *a);
+
+        /* This should serialize all type-specific variables. */
+        void (*serialize)(ActivationEventInfo *a, FILE *f);
+
+        /* This should deserialize all type-specific variables, one at a time. */
+        int (*deserialize)(const char *key, const char *value, ActivationEventInfo **a);
+} ActivationEventInfoVTable;
```
poettering: oh, i would not bother with a vtable really, it's not that we need 100 fields or so in the info `ActivationEventInfo` structure. i'd just do separate fields in it that migh tbe unit type specific. let's keep this simple...

poettering:
So I'd rreally make the ActivationEventInfo structure simpler, i.e. no vtable, just add the fields you need directly, even if some of them won't be used in all unit types. I mean, it's not like we are going to have 100 fields in there (like we do have for Unit structs), so a bunch of unnecessary pointers shouldn't really matter that much...

And then just serialize the fields that are set in it, and that's all.

bluca:
> So I'd rreally make the ActivationEventInfo structure simpler, i.e. no vtable, just add the fields you need directly, even if some of them won't be used in all unit types. I mean, it's not like we are going to have 100 fields in there (like we do have for Unit structs), so a bunch of unnecessary pointers shouldn't really matter that much...
> 
> And then just serialize the fields that are set in it, and that's all.

This is still draft, I'll add a couple more, working on it

src/core/job.c
bluca: but it's only taken over if there is nothing, so there's nothing to unref? which is what we talked about, older info event wins

src/core/unit.h
bluca: I've just added a couple more, and with the plans you outlined to extend it further there will be even more, most likely. So I'd like to keep it. It looks much more elegant and allows to extend to other unit types with a lot less churn.

src/basic/glob-util.c
```diff
@@ -58,6 +58,14 @@ int glob_exists(const char *path) {
                 return false;
```
poettering, line 60: should do `if (ret_first) *ret_first = NULL;` here...

after all we have this rule that on success we should initialize all return vlaues, and on error leave them unset. But failure in this sense is only stuff < 0, success is everything >= 0.

src/core/service.c
```diff
@@ -1641,6 +1641,17 @@ static int service_spawn_internal(
                 }
         }
 
+        Job *j = UNIT(s)->job;
+        if (j && j->activation_event_info) {
+                r = activation_event_info_format(j->activation_event_info, &our_env);
```
poettering: maybe call this `format_env()` rather than just `format()`. After all this is formatting as env vars, and not for example formatting for serialization purposes.

poettering: maybe even activation_event_info_append_env() actually, to emphasize that the strv specified is extended, and not a simple return value

src/core/service.c
```diff
@@ -1641,6 +1641,17 @@ static int service_spawn_internal(
...
+                if (r < 0)
+                        return r;
+                /* The number of env vars added here can vary, rather than keeping the allocation block in
+                 * sync manually, these functions simply use the strv methods to append to it, so we need
+                 * to update n_env when we are done in case of future usage. */
+                n_env = strv_length(our_env) + 1;
```
poettering: maybe make activation_event_info_append_end() return the number of appended entries on success, so that here you can just do:

```c
n_env += r;
```

man/systemd.exec.xml
```diff
@@ -3694,6 +3694,20 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
...
 
+        <varlistentry>
+          <term><varname>$TRIGGER_UNIT</varname></term>
+          <term><varname>$TRIGGER_PATH</varname></term>
+
+          <listitem><para>If the unit was activated dynamically (e.g.: a corresponding path unit), the
+          unit that triggered it and other type-dependent information will be passed via these variable. Note that
+          this is provided in a best-effort way. For example, there might be multiple triggers coalesced if happening
+          one after the other, and only one will be reported, and there is no guarantee on which one it will be.
+          Because of this, if there are multiple triggers configured for the same unit, this variable will be primarily
```
poettering: "if there are multiple triggers"? There almost certainly are, after all users can always start services manually via "systemctl start". i'd thus be more vague here. maybe:

"Becuase of this, in most cases this variable will be primarily information, …"

src/core/path.c
```diff
@@ -518,10 +534,24 @@ static void path_enter_running(Path *p) {
...
+        if (free_and_strdup(&(ACTIVATION_EVENT_INFO_PATH(info))->trigger_path_filename, trigger_path) < 0) {
+                r = -ENOMEM;
+                goto fail;
+        }
+
+        r = manager_add_job(UNIT(p)->manager, JOB_START, trigger, JOB_REPLACE, NULL, &error, &job);
         if (r < 0)
                 goto fail;
 
+        if (job)
```
poettering: the added job might be result of a coalescing event, and thus might already have the `activation_event_info` field set

src/core/path.c
```diff
@@ -832,6 +864,67 @@ static int path_can_start(Unit *u) {
...
+        assert(key);
+        assert(value);
+
+        if (streq(key, "activation-event-info-path-filename")) {
+                if (!info || !*info)
+                        return -EINVAL;
+
+                ActivationEventInfoPath *p = ACTIVATION_EVENT_INFO_PATH(*info);
+                if (!p)
+                        return -EINVAL;
```
poettering: to simplify things I'd really do this things outside of the streq() if block. I mean, `*info` must be a valid ActivationEventInfo object either way, so it makes sense to test this generically beforehand, instead of specifically for the `activation-event-info-path-filename` field.

src/core/timer.c
```diff
@@ -594,12 +596,23 @@ static void timer_enter_running(Timer *t) {
...
+
+        r = manager_add_job(UNIT(t)->manager, JOB_START, trigger, JOB_REPLACE, NULL, &error, &job);
         if (r < 0)
                 goto fail;
 
         dual_timestamp_get(&t->last_trigger);
 
+        if (job) {
+                ACTIVATION_EVENT_INFO_TIMER(info)->last_trigger = t->last_trigger;
+                job->activation_event_info = activation_event_info_ref(info);
```
poettering: here too. if the job object wasn't allocated new here, but result of job coalescing, there might already be the activation_event_info field be initialized.

maybe add a helper: job_set_activation_event_info(), that does the unreffing+reffing, and is then also the place where the policy of "oldest activation info wins" is singularly implemented (with comment)

src/core/timer.c
```diff
@@ -893,6 +906,64 @@ static int timer_can_start(Unit *u) {
...
+        assert(key);
+        assert(value);
+
+        if (streq(key, "activation-event-info-timer-last-trigger")) {
+                if (!info || !*info)
+                        return -EINVAL;
+
+                ActivationEventInfoTimer *t = ACTIVATION_EVENT_INFO_TIMER(*info);
+                if (!t)
+                        return -EINVAL;
```
poettering: here, too, moe outside of the streq() if block

src/core/timer.c
```diff
@@ -893,6 +906,64 @@ static int timer_can_start(Unit *u) {
...
+        char *s;
+
+        assert(info);
+        assert(strv);
+        assert(t);
+
+        if (!dual_timestamp_is_set(&t->last_trigger))
+                return 0;
+
+        s = strjoin("TRIGGER_TIMER_REALTIME=", FORMAT_TIMESTAMP(t->last_trigger.realtime));
```
poettering: hmm, this would format the timestamps in our own format. I figure it might be nice to format it as integer simply, i.e. USEC_FMT. Should be simpler for people to consume I think.

Maybe rename TRIGGER_TIMER_REALTIME_USEC= also? i.e. make the unit clear, once it is just a formatted integer.

src/core/timer.c
```diff
@@ -893,6 +906,64 @@ static int timer_can_start(Unit *u) {
...
+        if (!s)
+                return -ENOMEM;
+
+        if (strv_consume(strv, TAKE_PTR(s)) < 0)
+                return -ENOMEM;
+
+        if (strv_extendf(strv, "TRIGGER_TIMER_MONOTONIC=%" PRIu64, t->last_trigger.monotonic) < 0)
```
poettering: following discussion above: rename TRIGGER_TIMER_MONOTONIC_USEC?

also, output via USEC_FMT, not PRIu64 please

bluca:
Implemented as requested:

- refcounted structure holds per-unit-type information
- structure passed from trigger to job, and from job to successfully activated unit
- structure published via D-Bus on both job and unit objects as array of key/value string pairs, with the keys corresponding to the env var (lowercase instead of upper case)

PTAL!

src/core/dbus-job.c
```diff
@@ -117,6 +117,40 @@ int bus_job_method_get_waiting_jobs(sd_bus_message *message, void *userdata, sd_
...
+                sd_bus *bus,
+                const char *path,
+                const char *interface,
+                const char *property,
+                sd_bus_message *reply,
+                void *userdata,
+                sd_bus_error *error) {
+
+        _cleanup_strv_free_ char **pairs = NULL;
+        Job *j = userdata;
```
dtardon: Use `ASSERT_PTR()` macro here?

src/core/dbus-unit.c
```diff
@@ -857,6 +857,40 @@ static int property_get_refs(
...
+                sd_bus *bus,
+                const char *path,
+                const char *interface,
+                const char *property,
+                sd_bus_message *reply,
+                void *userdata,
+                sd_bus_error *error) {
+
+        _cleanup_strv_free_ char **pairs = NULL;
+        Unit *u = userdata;
```
dtardon: Use ASSERT_PTR() macro here?

src/core/unit.h
```diff
@@ -110,6 +110,72 @@ typedef union UnitDependencyInfo {
...
+
+        /* This should append type-specific variables as key/value pairs for the D-Bus property of the job,
+         * and return the number of added pairs. */
+        int (*append_pair)(ActivationEventInfo *info, char ***strv);
+} ActivationEventInfoVTable;
+
+extern const ActivationEventInfoVTable * const activation_event_info_vtable[_UNIT_TYPE_MAX];
+
+static inline const ActivationEventInfoVTable* ACTIVATION_EVENT_INFO_VTABLE(const ActivationEventInfo *a) {
+        return activation_event_info_vtable[a->trigger_unit_type];
```
poettering: maybe add asserts that check the type here

man/org.freedesktop.systemd1.xml
```diff
@@ -1898,6 +1898,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
       readonly s CollectMode = '...';
       @org.freedesktop.DBus.Property.EmitsChangedSignal("false")
       readonly as Refs = ['...', ...];
+      @org.freedesktop.DBus.Property.EmitsChangedSignal("const")
```
poettering: "const" is wrong. A unit can be loaded and then be started/stopped multiple times and for each cycle the activation info should be able to change

src/core/dbus-unit.c
```diff
@@ -857,6 +857,39 @@ static int property_get_refs(
...
+        if (r < 0)
+                return r;
+
+        STRV_FOREACH_PAIR(key, value, pairs) {
+                r = sd_bus_message_append(reply, "(ss)", *key, *value);
+                if (r < 0)
+                        return r;
+        }
+
+        return sd_bus_message_close_container(reply);
```
poettering: dbus-job.c has pretty much the same function now, even under the same name. Please add a helper somewhere that unifies the main part of this function somewhere

man/org.freedesktop.systemd1.xml
```diff
@@ -1898,6 +1898,8 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
...
+      readonly a(ss) ActivationEventInfo = [...];
```
poettering: Somehow the "Info" suffix looks weird as a dbus property to me now. Maybe call this ActivationDetails? both in the dbus property and the structure behind it?

src/core/dbus-unit.c
poettering: also, there's sd_Bus_message_append_strv()

src/core/job.c
```diff
@@ -962,6 +968,10 @@ int job_finish_and_invalidate(Job *j, JobResult result, bool recursive, bool alr
 
         if (IN_SET(result, JOB_FAILED, JOB_INVALID))
                 j->manager->n_failed_jobs++;
+        else if (result == JOB_DONE && u->activation_event_info == NULL && IN_SET(t, JOB_START, JOB_RESTART))
+                u->activation_event_info = TAKE_PTR(j->activation_event_info); /* Older event wins */
+        else if (result == JOB_DONE && t == JOB_STOP)
+                u->activation_event_info = activation_event_info_unref(u->activation_event_info);
```
poettering: Hmm, so you are filling this in here after activation is complete, and invalidate it after deactivation is complete?

I think this isn#t right. I am sure this should be installed whenever we start activation, i.e. the `.start()` vtable functions of the unit types should be able to look in there, if they like.

Hence, I think unit_start() should take a pointer to the new struct as (optional) parameter, and install it already before invoking that vtable function `.start()`.

Moreover, I think we should leave the object in place when we stop the thing, so that (unless a unit is GC'ed) we can still see the info all the way until the next activation.

Or with other words, this info should have the life-time of most runtime bits we have, such as the invocation ID: we reset/initialize it in unit_start(), and then leave it around until basically the next unit_start().

src/core/service.c
```diff
@@ -1641,6 +1641,17 @@ static int service_spawn_internal(
                 }
         }
 
+        Job *j = UNIT(s)->job;
+        if (j && j->activation_event_info) {
```
poettering: so, this really should read the value off `u`, not of `u->job`...

It's indication that the copying from job to unit happens too late, as discussed above: plesae pass the info from job to unit as part of unit_start() parameters, so that it is in place here already, and you can just pick it up from `u->job`.

src/core/unit.h
```diff
@@ -110,6 +110,72 @@ typedef union UnitDependencyInfo {
...
+ActivationEventInfo *activation_event_info_new(Unit *trigger_unit);
+ActivationEventInfo *activation_event_info_ref(ActivationEventInfo *p);
+ActivationEventInfo *activation_event_info_unref(ActivationEventInfo *p);
+void activation_event_info_serialize(ActivationEventInfo *p, FILE *f);
+int activation_event_info_deserialize(const char *key, const char *value, ActivationEventInfo **info);
+int activation_event_info_append_env(ActivationEventInfo *info, char ***strv);
+int activation_event_info_append_pair(ActivationEventInfo *info, char ***strv);
+DEFINE_TRIVIAL_CLEANUP_FUNC(ActivationEventInfo*, activation_event_info_unref);
+
+typedef struct ActivationEventInfoVTable {
```
poettering: why have a separate vtable for this? why not use the unit vtable? you only have one of these per unit type, so it shoud just be the same?

poettering (changes requested):

src/core/unit.h
bluca: because the lifetime and management of the info object is independent from the lifetime and management of the unit object (given they are also owned by jobs), so it seems much cleaner and tidier to keep it separate, conceptually, even if in practice it wouldn't make a difference

src/core/dbus-unit.c
bluca: sd_bus_message_append_strv appends as `as`, but here I need key/value pair, `a(ss)`

src/core/service.c
bluca: ok, I've changed the rest, and kept u->job->activation_details here

man/org.freedesktop.systemd1.xml
```diff
@@ -2406,6 +2409,17 @@ node /org/freedesktop/systemd1/unit/avahi_2ddaemon_2eservice {
...
+      <para><varname>ActivationDetails</varname> contains a list of string pairs, key and value, that
+      describe the event that caused the unit to be activated, if any. The key describes the information
+      (e.g.: <varname>trigger_unit</varname>, with value <varname>foo.service</varname>). This is only filled
+      in if the unit was triggered by a <varname>Path</varname> or <varname>Timer</varname> unit, and it is
+      only provided in a best effort fashion: it is not guaranteed to be set, and it is not guaranteed to be
+      the only trigger. It is only guaranteed to be a valid trigger that caused the activation job to be
+      enqueued and complete successfully. The key value pairs correspond (in lowercase) to the environment
+      variables described in the <literal>Environment Variables Set on Triggered Units</literal> section in
+      <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>1</manvolnum></citerefentry>.
+      Note that new key value pair may be added at any time in future versions.</para>
```
keszybz: Are we saying that existing entries are API and will not be removed?

src/core/unit.h
```diff
@@ -110,6 +110,75 @@ typedef union UnitDependencyInfo {
         } _packed_;
 } UnitDependencyInfo;
 
+/* Store information about why a unit was activated.
+ * We start with trigger units (.path/.timer), eventually it will be expanded to include more metadata. */
+typedef struct ActivationDetails {
+        UnitType trigger_unit_type;
+        unsigned n_ref;
+        char *trigger_unit_name;
```
keszybz: It is strange to not have `n_ref` as the first entry.

man/systemd.exec.xml
```diff
@@ -3694,6 +3694,19 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
           system.</para></listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><varname>$TRIGGER_UNIT</varname></term>
+          <term><varname>$TRIGGER_PATH</varname></term>
+
+          <listitem><para>If the unit was activated dynamically (e.g.: a corresponding path unit), the
+          unit that triggered it and other type-dependent information will be passed via these variable. Note that
```
keszybz: variable**s**

man/systemd.exec.xml
```diff
@@ -3694,6 +3694,19 @@ StandardInputData=V2XigLJyZSBubyBzdHJhbmdlcnMgdG8gbG92ZQpZb3Uga25vdyB0aGUgcnVsZX
...
+          this is provided in a best-effort way. For example, there might be multiple triggers coalesced if happening
```
keszybz: this is → this information is

"For example, multiple triggers happening one after another will be coalesced and only one will be reported, with no guarantee as to which one it will be."

man/systemd.path.xml
```diff
@@ -206,6 +206,13 @@
     <xi:include href="systemd.service.xml" xpointer="shared-unit-options" />
   </refsect1>
 
+  <refsect1>
+    <para>Environment variables with details on the trigger will be set for triggered units. See the
+    <literal>Environment Variables Set on Triggered Units</literal> section in
+    <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>1</manvolnum></citerefentry>
+    for more details.</para>
+  </refsect1>
+
```
keszybz: Hmm, this adds a section without a title. Please move this paragraph to some other section instead.

src/core/path.c
```diff
@@ -229,6 +233,16 @@ static bool path_spec_check_good(PathSpec *s, bool initial, bool from_trigger_no
...
         }
 
+        if (good) {
+                if (!trigger) {
+                        trigger = strdup(s->path);
+                        if (!trigger)
+                                (void) log_oom_debug();
+                }
+                if (trigger)
+                        *ret_trigger_path = TAKE_PTR(trigger);
```
keszybz: The check at line 242 is always satisfied. I guess you want
```C
if (good && ret_trigger_path) {
   if (!trigger) {
      …
   }

   *ret_trigger_path = TAKE_PTR(trigger);
}
```

src/core/path.c
```diff
@@ -832,6 +863,85 @@ static int path_can_start(Unit *u) {
         return 1;
 }
 
+static void activation_details_path_done(ActivationDetails *details) {
+        ActivationDetailsPath *p = ACTIVATION_DETAILS_PATH(details);
```
keszybz: ASSERT_PTR() ?

man/systemd.timer.xml
```diff
@@ -364,6 +364,13 @@
     <xi:include href="systemd.service.xml" xpointer="shared-unit-options" />
   </refsect1>
 
+  <refsect1>
+    <para>Environment variables with details on the trigger will be set for triggered units. See the
+    <literal>Environment Variables Set on Triggered Units</literal> section in
+    <citerefentry><refentrytitle>systemd.exec</refentrytitle><manvolnum>1</manvolnum></citerefentry>
+    for more details.</para>
+  </refsect1>
```
keszybz: Also here, a section without a title.

keszybz (commented):
I didn't read the full discussion, so just some superficial comments on the patches.

keszybz:
Is the vtable really necessary? It adds a lot of complexity…

man/org.freedesktop.systemd1.xml
bluca: yes

src/core/unit.h
bluca: rearranged

bluca:
> Is the vtable really necessary? It adds a lot of complexity…

As per the long discussion, this is just the initial step and more usage/info will be added later, including more unit types. I very much prefer this way rather than having long if chains everywhere.

src/core/dbus-unit.c
```diff
@@ -951,6 +965,7 @@ const sd_bus_vtable bus_unit_vtable[] = {
         SD_BUS_PROPERTY("InvocationID", "ay", bus_property_get_id128, offsetof(Unit, invocation_id), SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE),
         SD_BUS_PROPERTY("CollectMode", "s", property_get_collect_mode, offsetof(Unit, collect_mode), SD_BUS_VTABLE_PROPERTY_CONST),
         SD_BUS_PROPERTY("Refs", "as", property_get_refs, 0, 0),
+        SD_BUS_PROPERTY("ActivationDetails", "a(ss)", property_get_activation_details, 0, SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE),
```
poettering: Use the offset parameter to the macro (i.e. 4th arg, where you currently pass 0), so that you can use the same implementation for the property_get_activation_details() of the unit and the job. If you pass the offset to the `activation_details` field in the Unit/Job structure there, then both just have to deref a pointer to a pointer to get to the struct.

i.e. here it would be:

```c
SD_BUS_PROPERTY("ActivationDetails", "a(ss)", busproperty_get_activation_details, offsetof(Unit, activation_details), SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE),
```

and over in the job dbus code:

```c
SD_BUS_PROPERTY("ActivationDetails", "a(ss)", busproperty_get_activation_details, offsetof(Job, activation_details), SD_BUS_VTABLE_PROPERTY_EMITS_CHANGE),
```

src/core/dbus-util.c
```diff
@@ -228,3 +228,26 @@ int bus_read_mount_options(
...
+                return r;
+
+        STRV_FOREACH_PAIR(key, value, pairs) {
+                r = sd_bus_message_append(reply, "(ss)", *key, *value);
+                if (r < 0)
+                        return r;
+        }
+
+        return sd_bus_message_close_container(reply);
+}
```
poettering, line 262: one of these days we should add sd_bus_message_append_strv_pairs(). I have the suspicion this is not the only place where we do this.

src/core/service.c
```diff
@@ -1641,6 +1641,17 @@ static int service_spawn_internal(
                 }
         }
 
+        Job *j = UNIT(s)->job;
+        if (j && j->activation_details) {
```
poettering: why do you use the job's activation details, instead of the unit's, here?

src/core/unit.c
```diff
@@ -1941,6 +1947,9 @@ int unit_start(Unit *u) {
         unit_add_to_dbus_queue(u);
         unit_cgroup_freezer_action(u, FREEZER_THAW);
 
+        if (u->activation_details == NULL) /* Older details object wins */
```
poettering: for pointers we rely on C's downgrade-to-bool syntax, i.e.:

```c
if (!u->activation_details) …
```

src/core/unit.c
```diff
@@ -5921,3 +5930,149 @@ int unit_get_dependency_array(const Unit *u, UnitDependencyAtom atom, Unit ***re
...
+                        return -ENOMEM;
+        }
+
+        if (ACTIVATION_DETAILS_VTABLE(details)->append_env) {
+                r = ACTIVATION_DETAILS_VTABLE(details)->append_env(details, strv);
+                if (r < 0)
+                        return r;
+        }
+
+        return r + (!isempty(details->trigger_unit_name) ? 1 : 0); /* Return the number of variables added to the env block */
```
poettering: the ternary expression is pointless... the `!isempty()` already returns 1 or 0 i the exact same cases

src/core/unit.c
```diff
@@ -5921,3 +5930,149 @@ int unit_get_dependency_array(const Unit *u, UnitDependencyAtom atom, Unit ***re
...
+int activation_details_append_pair(ActivationDetails *details, char ***strv) {
+        int r = 0;
+
+        assert(strv);
+
+        if (!details)
+                return 0;
+
+        if (!isempty(details->trigger_unit_name)) {
+                if (strv_extend(strv, "trigger_unit") < 0)
```
poettering: not that it matters, but given that strv_extend() already returns an error code, i'd just propagate that, instead of returning -ENOEMEM here..

src/core/unit.c
```diff
@@ -5921,3 +5930,149 @@ int unit_get_dependency_array(const Unit *u, UnitDependencyAtom atom, Unit ***re
...
+                        return -ENOMEM;
+        }
+
+        if (ACTIVATION_DETAILS_VTABLE(details)->append_env) {
+                r = ACTIVATION_DETAILS_VTABLE(details)->append_pair(details, strv);
+                if (r < 0)
+                        return r;
+        }
+
+        return r + (!isempty(details->trigger_unit_name) ? 1 : 0); /* Return the number of pairs added to the strv */
```
poettering: as above, the ternary op is pointless

src/core/unit.c
```diff
@@ -5921,3 +5930,149 @@ int unit_get_dependency_array(const Unit *u, UnitDependencyAtom atom, Unit ***re
...
+        assert(strv);
+
+        if (!details)
+                return 0;
+
+        if (!isempty(details->trigger_unit_name)) {
+                if (strv_extend(strv, "trigger_unit") < 0)
+                        return -ENOMEM;
+
+                if (strv_extend(strv, details->trigger_unit_name) < 0)
```
poettering: ditto

src/core/unit.h
```diff
@@ -110,6 +110,75 @@ typedef union UnitDependencyInfo {
...
+                return (MixedCase*) a;                                          \
+        }
+
+/* For casting the various unit types into a unit */
+#define ACTIVATION_DETAILS(u)                                      \
+        ({                                                            \
+                typeof(u) _u_ = (u);                                  \
+                ActivationDetails *_w_ = _u_ ? &(_u_)->meta : NULL; \
+                _w_;                                                  \
+        })
```
poettering, line 136: something is wrong with the alignment of the `\` on the rhs

src/core/path.c
```diff
@@ -518,10 +533,23 @@ static void path_enter_running(Path *p) {
...
 
-        r = manager_add_job(UNIT(p)->manager, JOB_START, trigger, JOB_REPLACE, NULL, &error, NULL);
+        details = activation_details_new(UNIT(p));
+        if (!details) {
+                r = -ENOMEM;
+                goto fail;
+        }
+
+        if (free_and_strdup(&(ACTIVATION_DETAILS_PATH(details))->trigger_path_filename, trigger_path) < 0) {
+                r = -ENOMEM;
```
poettering: as mentioned earlier, i'd just do:

```c
r = free_and_strdup();
if (r < 0)
        return r;
```

i.e. if we have error codes, always propagate them as we have them

src/core/path.c
```diff
@@ -832,6 +862,81 @@ static int path_can_start(Unit *u) {
...
+
+        ActivationDetailsPath *p = ACTIVATION_DETAILS_PATH(*details);
+        if (!p)
+                return -EINVAL;
+
+        if (!streq(key, "activation-details-path-filename"))
+                return -EINVAL;
+
+        if (free_and_strdup(&p->trigger_path_filename, value) < 0)
+                return -ENOMEM;
```
poettering: as above, here and everywhere

src/core/service.c
poettering: I mean, what I want to say: this should just use UNIT(s)->activation_details

src/core/unit.h
bluca: leftover from one of the requested big renames, done with search/replace

src/core/dbus-util.c
bluca: that would be good, but this is already large-ish PR and I'd rather not add more functionality

poettering:
(one of those days, .automount units should acquire this too, and expose this over the bus. i.e. info about process that triggered a mount, and similar)

(and socket units, too, of course)

poettering:
(and of course activation through bus calls, and of course activation through initial transaction)

src/core/timer.c
```diff
@@ -892,6 +903,91 @@ static int timer_can_start(Unit *u) {
...
+        if (!dual_timestamp_is_set(&t->last_trigger))
+                return 0;
+
+        r = strv_extendf(strv, "TRIGGER_TIMER_REALTIME_USEC=%" USEC_FMT, t->last_trigger.realtime);
+        if (r < 0)
+                return r;
+
+        r = strv_extendf(strv, "TRIGGER_TIMER_MONOTONIC_USEC=%" USEC_FMT, t->last_trigger.monotonic);
+        if (r < 0)
+                return -ENOMEM;
```
poettering: `return r`

bluca: I knew I'd forget one!

src/core/job.c
```diff
@@ -180,9 +182,13 @@ static void job_merge_into_installed(Job *j, Job *other) {
...
         assert(j->unit == other->unit);
 
-        if (j->type != JOB_NOP)
+        if (j->type != JOB_NOP) {
                 assert_se(job_type_merge_and_collapse(&j->type, other->type, j->unit) == 0);
-        else
+
+                /* Keep the oldest ActivationDetails, if any */
+                if (!j->activation_details)
+                        j->activation_details = TAKE_PTR(other->activation_details);
```
yuwata, line 190: Why not use `activation_details_ref()`??

src/core/unit.c
```diff
@@ -5919,3 +5928,154 @@ int unit_get_dependency_array(const Unit *u, UnitDependencyAtom atom, Unit ***re
...
+                **details = (ActivationDetails) {
+                        .n_ref = 1,
+                        .trigger_unit_type = t,
+                };
+
+                return 0;
+        }
+
+        if (streq(key, "activation-details-unit-name")) {
+                (*details)->trigger_unit_name = strdup(value);
```
yuwata, line 6014: `free_and_strdup()` is safer? Or, refuse or ignore the value when the unit name is already set.

PR #24072, opened 2022-07-21
decgroupsv1ification: first steps – remove from docs, and generate warnings

poettering:
It appears unlikely we can quickly remove cgroupsv1 support from our tree. But I guess we can start with removing any mentions of it from the docs, and generating warnings if people still use cgroupsv1-specific options.

This doesn't change anything code-wise really, it just simplifies the docs by removing a lot of cgroupsv1/cgroupsv2 conditionalizations.

### DIFF
--- a/src/core/load-fragment.c
+++ b/src/core/load-fragment.c
@@ -843,9 +843,9 @@ int config_parse_kill_mode(
 
         if (m == KILL_NONE)
                 log_syntax(unit, LOG_WARNING, filename, line, 0,
-                           "Unit configured to use KillMode=none. "
+                           "Unit uses KillMode=none. "
                            "This is unsafe, as it disables systemd's process lifecycle management for the service. "
-                           "Please update your service to use a safer KillMode=, such as 'mixed' or 'control-group'. "
+                           "Please update the service to use a safer KillMode=, such as 'mixed' or 'control-group'. "
                            "Support for KillMode=none is deprecated and will eventually be removed.");
 
         *k = m;


src/core/load-fragment.c
```diff
@@ -3913,9 +3936,14 @@ int config_parse_memory_limit(
                 c->memory_max = bytes;
         else if (streq(lvalue, "MemorySwapMax"))
                 c->memory_swap_max = bytes;
-        else if (streq(lvalue, "MemoryLimit"))
+        else if (streq(lvalue, "MemoryLimit")) {
+
+                log_syntax(unit, LOG_WARNING, filename, line, 0,
+                           "Unit configured to use MemoryLimit=. "
+                           "This is an obsolete cgroupv1 setting, please update to use MemoryMax= instead. Support for MemoryLimit= will be removed soon.");
```
keszybz: "Unit uses MemoryLimit=; please use MemoryMax= instead. Support for MemoryLimit= will be removed soon."

DaanDeMeyer (approved):

keszybz (changes requested):
I know that this is how we used to do things, but I think we should be more user-friendly and keep mentions of the obsolete settings in the docs. This is something we discussed during one of the meetings. Instead of simply dropping the docs, I'd very much prefer to add a section like "History" or "Obsolete Interfaces" and add a few short paragraphs there:

  BlockIOWeight= ,BlockIODeviceWeight=, …, have been obsoleted. Use IOWeight=, IODeviceWeight=…, instead. 

The motivation for this is that users will encounter those old settings in their units, and if we have this in the docs, it's much easier for users to figure out that the setting is deprecated (and later that is has been removed). People can also link to this section of the man page as  documentation. Only having this embedded in code and logs is not enough for many use cases. This transitions will go on for a long time, so easily-accessible and linkable docs for this will be very useful.

This would be a one time addition, so there wouldn't be any maintenance burden.

poettering:
> BlockIOWeight= ,BlockIODeviceWeight=, …, have been obsoleted. Use IOWeight=, IODeviceWeight=…, instead.

We have carried this for a long time. This PR gets rid of that. That's the whole point.

keszybz:
We can drop the detailed explanations of the old options. That's a lot of text that we don't need. If the choice is only between "drop everything" (like in the current version), and "keep things unchanged", then actually the latter is would be better. But I hope we can strike the middle (or 90-th centile) ground of removing everything but a brief mention to guide users in the right direction.

keszybz:
I did some small fixups to your commits (as discussed above) and add a commit that adds "History" sections with a list of removed settings. The PR needed a rebase anyway, so I didn't open a new pull request. I hope that's OK.

PR #24074, opened 2022-07-22
network: dhcp6: re-introduce RapidCommit= setting

### LINKED ISSUE #24065: Support 4-message dhcpv6 exchange (no rapid commit) and/or rebind-renew
### Component

systemd-networkd

### Is your feature request related to a problem? Please describe

Posting this following my comment here: https://github.com/systemd/systemd/issues/22633#issuecomment-1188297372.

My problem is two-fold:

- My ISP's router/modem supports PD with rapid-commit (as in, it sends a `reply` containing a valid IA_PD after a `solicit`), but doesn't route return packets to my router. The return route is only created properly when using 4-message exchange (without rapid-commit), or, as of the first `renew` message after a rapid-commit `solicit`/`reply`.
- The ISP's modem is stateless. Whenever they push firmware updates, it reboots and comes back up without return routes to my box, **and** it also regenerates its server identifier, which, annoyingly, is an LLT. This makes it so I can't blindly replay `renew` messages periodically to get it to add a return route. (Additionally, my modem is not directly plugged into my router, so it doesn't receive interface down/up events when the modem reboots)

Unsurprisingly, getting the ISP to fix this is futile. :smiling_face_with_tear: I doubt they have much control over the firmware to begin with. (this is a Compal router-modem)

### Describe the solution you'd like

For making this 100% reliable for me, I'd need two things: reintroduce `RapidCommit=false` with full 4-message exchange, and implement `networkctl renew` to kick off a `rebind`/`solicit`/`renew`. Or, even better, make rebinding occur periodically with an interval config setting in `[DHCPv6]`.

As requested, here is a pcap of a working 4-message exchange: [dhcpv6_pcap.zip](https://github.com/systemd/systemd/files/9147514/dhcpv6_pcap.zip).
And here's a pcap of a rebind/solicit/renew: [dhcpcd_pcap.zip](https://github.com/systemd/systemd/files/9147981/dhcpcd_pcap.zip).

### Describe alternatives you've considered

I've used other dhcp clients like `wide-dhcpv6` and `dhcpcd`. Invoking dhcpcd with the `-N` flag does a rebind-solicit-renew as expected, but doesn't include a feature to run it periodically.

For now, though, I run a small Go program that periodically runs the 4-message exchange from another LL address in the subnet.

### The systemd version you checked that didn't have the feature you are asking for

_No response_

yuwata:
Previous implementation of disabling rapid commit was broken, and dropped by 308d01f3c4decf3240ca0c6c7f91049953df7299.

Let's re-introduce the option in the correct way, as some servers seem to return borked message when the solicit message contain the rapid commit option.

Closes #24065.

### DIFF
--- a/test/test-network/systemd-networkd-tests.py
+++ b/test/test-network/systemd-networkd-tests.py
@@ -4170,6 +4170,44 @@ class NetworkdDHCPClientTests(unittest.TestCase, Utilities):
         print(output)
         self.assertRegex(output, 'token :: dev veth99')
 
+        print('## dnsmasq log')
+        output = read_dnsmasq_log_file()
+        print(output)
+        self.assertIn('DHCPSOLICIT(veth-peer)', output)
+        self.assertNotIn('DHCPADVERTISE(veth-peer)', output)
+        self.assertNotIn('DHCPREQUEST(veth-peer)', output)
+        self.assertIn('DHCPREPLY(veth-peer)', output)
+        self.assertIn('sent size:  0 option: 14 rapid-commit', output)
+
+        with open(os.path.join(network_unit_dir, '25-dhcp-client-ipv6-only.network'), mode='a', encoding='utf-8') as f:
+            f.write('\n[DHCPv6]\nRapidCommit=no\n')
+
+        stop_dnsmasq()
+        start_dnsmasq()
+
+        networkctl_reload()
+        self.wait_online(['veth99:routable', 'veth-peer:routable'])
+
+        # checking address
+        output = check_output('ip address show dev veth99 scope global')
+        print(output)
+        self.assertRegex(output, r'inet6 2600::[0-9a-f:]*/128 scope global dynamic noprefixroute')
+        self.assertNotIn('192.168.5', output)
+
+        # checking semi-static route
+        output = check_output('ip -6 route list dev veth99 2001:1234:5:9fff:ff:ff:ff:ff')
+        print(output)
+        self.assertRegex(output, 'via fe80::1034:56ff:fe78:9abd')
+
+        print('## dnsmasq log')
+        output = read_dnsmasq_log_file()
+        print(output)
+        self.assertIn('DHCPSOLICIT(veth-peer)', output)
+        self.assertIn('DHCPADVERTISE(veth-peer)', output)
+        self.assertIn('DHCPREQUEST(veth-peer)', output)
+        self.assertIn('DHCPREPLY(veth-peer)', output)
+        self.assertNotIn('rapid-commit', output)
+
     def test_dhcp_client_ipv4_only(self):
         copy_network_unit('25-veth.netdev', '25-dhcp-server-veth-peer.network', '25-dhcp-client-ipv4-only.network')
 


src/libsystemd-network/sd-dhcp6-client.c
```diff
@@ -491,6 +491,14 @@ int dhcp6_client_set_transaction_id(sd_dhcp6_client *client, uint32_t transactio
         return 0;
 }
 
+int sd_dhcp6_client_set_rapid_commit(sd_dhcp6_client *client, int enable) {
+        assert_return(client, -EINVAL);
+        assert_return(!sd_dhcp6_client_is_running(client), -EBUSY);
+
+        client->rapid_commit = enable;
```
bluca, line 498: implicit conversion int -> bool? maybe add a check that it's not being passed a negative number or so?

bluca (approved):

src/libsystemd-network/sd-dhcp6-client.c
yuwata: Hmm, we usually do not check such situation, that is, we only take into zero or non-zero.
Note that the function prefixed with `sd_`, but it is not exported yet.

yuwata: In other words, if we should check that, then that should be done complehensively in other later PR, I think.

bluca: then just a `!!enable` for an explicit conversion?

yuwata: I will check, but IIRC we have dropped `!!` previously e.g. from sd-bus functions which also take boolean argument as int.

yuwata: Hm, after grepping the code, we do not have consistent rule about that.
E.g. `sd_bus_set_bus_client()` or friends use `!!`. But, `sd_bus_slot_set_floating()` does not use `!!`.

yuwata: And in `sd_dhcp6_client_set_information_request()`, we do not use `!!`.
Hence, I'd like to keep the code now as is. And, if you or others would really like to use `!!`, then let's change that consistently later.

src/libsystemd-network/sd-dhcp6-client.c
```diff
@@ -1160,6 +1170,10 @@ static int client_process_advertise_or_rapid_commit_reply(
         if (message->type == DHCP6_MESSAGE_REPLY) {
                 bool rapid_commit;
 
+                if (!client->rapid_commit)
+                        return log_dhcp6_client_errno(client, SYNTHETIC_ERRNO(EINVAL),
+                                                      "Received unexpected reply message, ignoring.");
```
poettering: maybe mention why unexpected in the log msg

man/systemd.network.xml
```diff
@@ -2147,6 +2147,19 @@ Table=1234</programlisting></para>
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>RapidCommit=</varname></term>
+        <listitem>
+          <para>Takes a boolean. The DHCPv6 client can obtain configuration parameters from a DHCPv6 server
+          through a rapid two-message exchange (solicit and reply). When the rapid commit option is set by
+          both the DHCPv6 client and the DHCPv6 server, the two-message exchange is used. Otherwise, the
+          default four-message exchange (solicit, advertise, request, and reply) is used. The two-message
```
poettering: "default" is a bit misleading, given that our default actually is the two message exchange.

maybe "traditional" or "classic" or so is a better word?

yuwata: Simply dropped "default". Both two- and four-message exchange protocol is defined in RFC3315, which is one of the earliest RFC, if my knowledge is correct. At least, the rapid commit protocol is not new. Hence, "traditional" or "classic" sounds not good for me.

src/libsystemd-network/sd-dhcp6-client.c
yuwata: Updated.

yuwata:
@bluca and @poettering Thank you for the comments. Hopefully addressed @poettering's comments. And let's address @bluca's comment later in a consistent way if we have some agreement. Upgrading the green label.

ti-mo:
@yuwata Thanks for picking this up so quickly, works like a charm! These patches still apply cleanly to 251.3 that's currently in Debian backports. :+1:

PR #24077, opened 2022-07-22
tests: add a simple test that checks what fexecve_or_execve does

keszybz:
This can be easily used to show that the issue described in
ceedbf8185fc7593366679f02d31da63af8c4bd1 is still reproduced:

$ strace -e execveat build/test-execve /bin/grep Name /proc/self/status
execveat(3, "", ["/bin/grep", "Name", "/proc/self/status"], NULL, AT_EMPTY_PATH) = 0
Name:	3

### DIFF
--- a/src/test/test-execve.c
+++ b/src/test/test-execve.c
@@ -1,0 +1,40 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "exec-util.h"
+#include "fd-util.h"
+#include "log.h"
+#include "main-func.h"
+#include "strv.h"
+#include "tests.h"
+
+/* This program can be used to call programs through fexecve / execveat(…, "", …, AT_EMPTY_PATH),
+ * when compiled with -Dfexecve=true, and the fallback paths, when -Dfexecve=false.
+ *
+ * Example:
+ * $ strace -e execveat build/test-execve /bin/grep Name /proc/self/status
+ * execveat(3, "", ["/bin/grep", "Name", "/proc/self/status"], NULL, AT_EMPTY_PATH) = 0
+ * Name:   3
+ *
+ * FIXME: use the new kernel api to set COMM properly when the kernel makes that available.
+ * C.f. ceedbf8185fc7593366679f02d31da63af8c4bd1.
+ */
+
+static int run(int argc, char **argv) {
+        _cleanup_close_ int fd;
+        char **args = strv_skip(argv, 1);
+        int r;
+
+        test_setup_logging(LOG_DEBUG);
+
+        args = !strv_isempty(args) ? args : STRV_MAKE("/bin/true");
+
+        fd = open(args[0], O_RDONLY | O_CLOEXEC);
+        if (fd < 0)
+                return log_error_errno(errno, "open(%s) failed: %m", args[0]);
+
+        r = fexecve_or_execve(fd, args[0], args, NULL);
+        assert_se(r < 0);
+        return log_error_errno(r, "fexecve_or_execve(%s) failed: %m", args[0]);
+}
+
+DEFINE_MAIN_FUNCTION(run);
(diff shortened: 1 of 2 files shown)


src/test/test-execve.c
```diff
@@ -0,0 +1,24 @@
+#include "exec-util.h"
```
bluca, line 3: SPDX line is missing

daandemeyer: Should we write a test that checks that every file starts with an SPDX line?

bluca: ...but then we'll need a test that checks the test :-O
but yeah, that would be nice, although it will need some heuristics

bluca (approved):

src/test/test-execve.c
keszybz: Duh.

At least we don't use the copyright and license headers anymore.

src/test/test-execve.c
```diff
@@ -0,0 +1,26 @@
...
+        int r;
+
+        args = !strv_isempty(args) ? args : STRV_MAKE("/bin/true");
+
+        fd = open(path, O_RDONLY | O_CLOEXEC);
+        if (fd < 0)
+                return log_error_errno(errno, "open(%s) failed: %m", path);
+
+        r = fexecve_or_execve(fd, path, args, NULL);
+        assert(r < 0);
```
yuwata: `assert_se()` is better?

src/test/test-execve.c
```diff
@@ -0,0 +1,26 @@
...
+#include "log.h"
+#include "main-func.h"
+#include "strv.h"
+
+static int run(int argc, char **argv) {
+        _cleanup_close_ int fd;
+        const char *path = argv[1] ?: "/bin/true";
+        char **args = strv_skip(argv, 1);
+        int r;
+
```
yuwata, line 26: We usually call `test_setup_logging()` for tests at the beginning.

src/test/test-execve.c
```diff
@@ -0,0 +1,26 @@
...
+        args = !strv_isempty(args) ? args : STRV_MAKE("/bin/true");
```
yuwata, line 29: Not a matter, but we can drop `path`:
```
char **args = strg_skip(argv, 1);
int r;
...
args = !strv_isempty(args) ? args : STRV_MAKE("/bin/true");
...
r = fexecve_or_execve(fd, args[0], args, NULL);
```

yuwata (changes requested):

poettering:
I think this information from the commit msg should be included in the test as a comment, otherwise if you look at the test it might be very confusing what the point of it is...

> This can be easily used to show that the issue described in
> https://github.com/systemd/systemd/commit/ceedbf8185fc7593366679f02d31da63af8c4bd1 is still > reproduced:

> $ strace -e execveat build/test-execve /bin/grep Name /proc/self/status
>execveat(3, "", ["/bin/grep", "Name", "/proc/self/status"], NULL, AT_EMPTY_PATH) = 0
> Name: 3

src/test/test-execve.c
poettering: agree

keszybz:
Updated with all the suggestions.

keszybz:
centos ci: 16:38:26 TEST-13-NSPAWN-SMOKE

focal-pcc64el:
TEST-68-PROPAGATE-EXIT-STATUS:      SUCCESS  (  6 s)
TEST-69-SHUTDOWN:                   FAIL     ( 44 s)
TEST-70-TPM2:                       SUCCESS  (  1 s)
TEST-71-HOSTNAME:                   FAIL     (  0 s)
TEST-72-SYSUPDATE:                  FAIL     (  2 s)
TEST-73-LOCALE:                     FAIL     (  0 s)
TEST-75-RESOLVED:                   SUCCESS  (  0 s)

Both should be unrelated to this.

yuwata (approved):

PR #24078, opened 2022-07-22
manager: limit access to private dbus socket

keszybz:
For the system manager, /run/systemd/private is publicly accessible, because
/run/systemd is 0755, and /run/systemd/private is 0777. For the user manager,
/run/user/<uid> is 0700, and /run/user/<uid>/systemd/private is 0777. This
does not directly cause any security issue because we check the sender in
bus_check_peercred (ucred.uid != 0 && ucred.uid != geteuid()).

But it makes sense to limit access to the socket to avoid wasting time in PID1.
Somebody could send messages there that'd we'd reject anyway. It also makes
things more explicit.

### DIFF
--- a/src/core/dbus.c
+++ b/src/core/dbus.c
@@ -42,6 +42,7 @@
 #include "string-util.h"
 #include "strv.h"
 #include "strxcpyx.h"
+#include "umask-util.h"
 #include "user-util.h"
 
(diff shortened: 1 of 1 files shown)


bluca (approved):

poettering:
lgtm

yuwata (approved):

PR #24079, opened 2022-07-22
Document udevadm lock's exit code

OnkelUlla:
With respect to its exit code ``udevadm lock …`` is not different compared to similar commands which is what I expected. But to be sure I had to look up the [source code](https://github.com/systemd/systemd/blob/v251/src/udev/udevadm-lock.c#L396). So let's enhance the command's manpage accordingly.


poettering:
lgtm

yuwata (approved):
LGTM too.

Issue #9441, opened 2018-06-27
RFE: --force option for `machinectl copy-to/copy-from` 


shibumi:
**Describe the solution you'd like**
It would be nice to have the option to overwrite a file with the copy-to/copy-from feature. Currently we are not able to overwrite a file in a container:
```
 % sudo machinectl copy-to work .zshrc /home/chris/.zshrc                                                              
Failed to copy: File exists
```

What I suggest is something like this:
```
 % sudo machinectl copy-to work .zshrc /home/chris/.zshrc --force    
```

bluca:
https://github.com/systemd/systemd/pull/24080

PR #24080, opened 2022-07-22
Add --force flag to machinectl copy-[to|from]

rdtscp:
Relates to Issue #9441


man/org.freedesktop.machine1.xml
```diff
@@ -112,10 +112,12 @@ node /org/freedesktop/machine1 {
                        in  b mkdir);
       CopyFromMachine(in  s name,
                       in  s source,
-                      in  s destination);
+                      in  s destination,
+                      in  b force_copy);
```
poettering: hmm, so we probably should make this a flags param, i.e. `in t flags`. This smells as if we might want to add more sooner or later.

Flags in D-Bus suck, but it's the easiest still.

man/org.freedesktop.machine1.xml
```diff
@@ -394,7 +396,8 @@ node /org/freedesktop/machine1 {
 
       <para><function>CopyFromMachine()</function> copies files or directories from a container into the
       host. It takes a container name, a source directory in the container and a destination directory on the
-      host as arguments. <function>CopyToMachine()</function> does the opposite and copies files from a source
+      host as arguments. Optionally takes "--force" to force the copy if the destination already exists.
```
poettering: well, the dbus iface doesn't take `--force` really...

man/org.freedesktop.machine1.xml
```diff
@@ -112,10 +112,12 @@ node /org/freedesktop/machine1 {
...
       CopyFromMachine(in  s name,
                       in  s source,
-                      in  s destination);
+                      in  s destination,
+                      in  b force_copy);
       CopyToMachine(in  s name,
                     in  s source,
-                    in  s destination);
+                    in  s destination,
+                    in  b force_copy);
```
poettering: dbus methods are API, and hence you cannot just add or remove params, because that would break API.

Hence, instead of changing the existing method, please add CopyToMachineWithFlags() which is like CopyToMachine(), but takes the extra flags param

man/org.freedesktop.machine1.xml
poettering: Also, i think the internal flag should be called "replace", not "force". i.e. "force" could mean a lot, but this is specifically about replacing existing files, hence the naming should communicate that.

poettering: (i think it's fine if the switch on the cmdline is nonetheless called --force, since cp calls it that way)

src/shared/copy.c
```diff
@@ -711,7 +712,9 @@ static int fd_copy_regular(
                 if (r < 0)
                         return r;
         }
-        fdt = openat(dt, to, O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC|O_NOCTTY|O_NOFOLLOW, st->st_mode & 07777);
+        if (copy_flags & COPY_REPLACE)
+                dt_flags &= ~O_EXCL;
```
poettering: this doesn't work. Consider an existing file that is large, and you copy a small file over. If you just drop O_EXCL, you'd leave the file at its original size, and just override the beginning of it, but that's not what you want here.

So, do what 'cp' does: open for writing with O_EXCL, and if this yields EEXIST, then unlink the file, and try again.

poettering: also, this shouldn't just be done for regular files, but for any form of inodes we copy. i.e. the right place to put this unlink-and-retry logic is in fd_copy_directory() and copy_tree_at_full().

poettering:
also, please add a test to src/test/test-copy.c to test the COPY_REPLACE flag behaviour (i.e. does the same call with and without the flag and checks the right thing happens).

man/org.freedesktop.machine1.xml
rdtscp: Thank you for the review! I appreciate it as I am still learning about the repo.

> hmm, so we probably should make this a flags param, i.e. `in t flags`. This smells as if we might want to add more sooner or later.
> 
> Flags in D-Bus suck, but it's the easiest still.

Should we create an enum to store the possible flags here?

daandemeyer: > Should we create an enum to store the possible flags here?

Yes, create an enum, but it's not exposed in D-Bus itself. The D-Bus type for flags is uint64_t, or "t" in the D-Bus type system. You can then define an enum with all the values for use in the internal logic. See the `UnitFileFlags` enum for an example. It also defines a mask that's used to check that only defined flags are passed, and returns an error if there are undefined flags. This is so that we can add new flags in the future without breaking code that's setting undefined flags. You should do the same for the new enum.

src/shared/copy.c
rdtscp: Thanks for the catch!
Just so that I understand clearly, should we try to open the file and unlink if it returns EEXIST? Or is there an API we can use to check if the file exists (that is perhaps cheaper) and conditionally unlink?
Apologies if my question doesn't make sense, as I am new to using the Linux API

poettering: yes, try to open it with O_EXCL, as before, but then if EEXIST is seen and the COPY_REPLACE flag is set, unlink the file, and do the same thing again. If that then still fails with EEXIST something is off, just propagate the EEXIST.

man/org.freedesktop.machine1.xml
rdtscp: - [x] Need to fix the documentation

src/shared/copy.c
rdtscp: Thanks again for the review.
I have tried to address this by refactoring the "leaf" nodes of the file tree into a function, which I can then call twice in the case of EEXIST.
This is so that when copying a directory, we can selectively "unlink and retry" for only "leaf" nodes (i.e. not directory nodes).
Would appreciate your thoughts on this when you have time :)

src/machine/machine-dbus.c
```diff
@@ -924,9 +925,19 @@ int bus_machine_method_copy(sd_bus_message *message, void *userdata, sd_bus_erro
         if (m->class != MACHINE_CONTAINER)
                 return sd_bus_error_set(error, SD_BUS_ERROR_NOT_SUPPORTED, "Copying files is only supported on container machines.");
 
-        r = sd_bus_message_read(message, "ss", &src, &dest);
-        if (r < 0)
-                return r;
```
daandemeyer: You can keep this and only read the "t" type in the if block.

src/machine/machine-dbus.c
```diff
@@ -924,9 +925,19 @@ int bus_machine_method_copy(sd_bus_message *message, void *userdata, sd_bus_erro
...
+        if (sd_bus_message_is_method_call(message, NULL, "CopyFromMachineWithFlags") || sd_bus_message_is_method_call(message, NULL, "CopyToMachineWithFlags")) {
+                uint64_t raw_flags;
+                r = sd_bus_message_read(message, "sst", &src, &dest, &raw_flags);
+                if (r < 0)
+                        return r;
+
```
daandemeyer, line 937: You have to check here that there's no undefined flags in the flags read from the message and return an error if there are. If we ignored those we might break users that send undefined flags when we add support for a new flag in the future.

src/machine/machinectl.c
```diff
@@ -1119,20 +1119,38 @@ static int copy_files(int argc, char *argv[], void *userdata) {
...
         }
 
-        r = bus_message_new_method_call(
-                        bus,
-                        &m,
-                        bus_machine_mgr,
-                        copy_from ? "CopyFromMachine" : "CopyToMachine");
-        if (r < 0)
-                return bus_log_create_error(r);
+        if (arg_force) {
```
daandemeyer: We can just always use the "WithFlags" methods in machinectl. The command line tools don't need to support older versions of the daemons. If --force is not specified, you can just pas "0" as the flags argument.

src/machine/machined-dbus.c
```diff
@@ -1110,6 +1115,16 @@ const sd_bus_vtable manager_vtable[] = {
...
+        SD_BUS_METHOD_WITH_ARGS("CopyFromMachineWithFlags",
+                                SD_BUS_ARGS("s", name, "s", source, "s", destination, "t", flags),
+                                SD_BUS_NO_RESULT,
+                                method_copy_machine_with_flags,
+                                SD_BUS_VTABLE_UNPRIVILEGED),
+        SD_BUS_METHOD_WITH_ARGS("CopyToMachineWithFlags",
+                                SD_BUS_ARGS("s", name, "s", source, "s", destination, "t", flags),
+                                SD_BUS_NO_RESULT,
+                                method_copy_machine_with_flags,
+                                SD_BUS_VTABLE_UNPRIVILEGED),
```
daandemeyer: We should adapt CopyFrom/CopyTo in machine-dbus.c as well to have the WithFlags variants.

src/shared/install.h
```diff
@@ -48,6 +48,11 @@ enum UnitFileFlags {
         _UNIT_FILE_FLAGS_MASK_PUBLIC = UNIT_FILE_RUNTIME|UNIT_FILE_PORTABLE|UNIT_FILE_FORCE,
 };
 
+enum CopyFileFlags {
+        COPY_FILE_REPLACE = 1 << 0, /* Public API via DBUS, do not change */
+        _COPY_FILE_FLAGS_MASK_PUBLIC = COPY_FILE_REPLACE,
+};
```
daandemeyer: I'd put this in machine-dbus.h for now since it's specific to the machined D-Bus methods

src/shared/copy.h
```diff
@@ -27,9 +27,23 @@ typedef enum CopyFlags {
...
+         * mkdirat(parent_fd, subdir) — and are kept so that we can automatically remove the directory again
+         * when we are done. */
+        int parent_fd; /* Possibly AT_FDCWD */
+        char *subdir;
+} HardlinkContext;
+
 typedef int (*copy_progress_bytes_t)(uint64_t n_bytes, void *userdata);
 typedef int (*copy_progress_path_t)(const char *path, const struct stat *st, void *userdata);
 
+int fd_copy_tree_generic(int df, const char *from, const struct stat *st, int dt, const char *to, dev_t original_device, unsigned depth_left, uid_t override_uid, gid_t override_gid, CopyFlags copy_flags, HardlinkContext *hardlink_context, const char *display_path, copy_progress_path_t progress_path, copy_progress_bytes_t progress_bytes, void *userdata);
```
daandemeyer: This doesn't need to be a public method just yet, you can make it `static` in in copy.c since it's only used there. That means you can keep `HardlinkContext` internal as well.

src/shared/copy.c
```diff
@@ -1061,8 +1061,16 @@ int fd_copy_tree_generic(
...
                 r = fd_copy_directory(df, from, st, dt, to, original_device, depth_left-1, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_path, progress_bytes, userdata);
-        else
+        else {
                 r = fd_copy_leaf(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_bytes, userdata);
+                /* We just tried to copy a leaf node of the tree. If it failed because the node already exists *and* the COPY_REPLACE flag has been provided, we should unlink the node and re-copy. */
+                if (r == -EEXIST && (copy_flags & COPY_REPLACE)) {
+                        r = unlinkat(dt, to, 0);
+                        if (r < 0)
+                                return r;
+                        r = fd_copy_leaf(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_bytes, userdata);
```
daandemeyer: Missing newline before this statement

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+        assert_se(fd >= 0);
+        close(fd);
+
+        fd = mkostemp_safe(dest);
+        assert_se(fd >= 0);
+        close(fd);
+
+        assert_se(write_string_file(src, "bar bar", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(dest, "foo foo foo", WRITE_STRING_FILE_CREATE) == 0);
+
```
daandemeyer: Spurious newline here

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+
+        assert_se(read_full_file(dest, &buf, &sz) == 0);
+        assert_se(streq(buf, "foo foo foo\n"));
+        assert_se(sz == 12);
```
daandemeyer: No need to check the sizes here, you can just pass NULL as the sz argument to read_full_file() here and elsewhere

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+
+        /* Create the random source/destination directories */
+        r = mkdtemp(src_directory);
+        assert_se(r);
+        r = mkdtemp(dst_directory);
+        assert_se(r);
+
+        /* Construct the source/desitnation filepaths (should have different dir name, but same file names within) */
+        src_path1 = malloc(strlen(src_directory) + 2 + strlen(file1));
+        sprintf(src_path1, "%s/%s", src_directory, file1);
```
daandemeyer: We have `path_join()` for this

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+        size_t sz = 0;
+        int fd;
+
+        fd = mkostemp_safe(src);
+        assert_se(fd >= 0);
+        close(fd);
+
+        fd = mkostemp_safe(dest);
+        assert_se(fd >= 0);
+        close(fd);
```
daandemeyer: You can just use `tempfn_random()` here since we only need a path and no file descriptor

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+        assert_se(write_string_file(src_path1, "src file 1", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(src_path2, "src file 2", WRITE_STRING_FILE_CREATE) == 0);
+
+        assert_se(write_string_file(dst_path1, "dest file 1", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(dst_path2, "dest file 2", WRITE_STRING_FILE_CREATE) == 0);
+
+        {
+                assert_se(read_full_file(src_path1, &buf, &sz) == 0);
+                assert_se(streq(buf, "src file 1\n"));
+                assert_se(sz == 11);
```
daandemeyer: Ditto for no need to check sizes here.

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+        assert_se(r);
+        r = mkdtemp(dst_directory);
+        assert_se(r);
+
+        /* Construct the source/desitnation filepaths (should have different dir name, but same file names within) */
+        src_path1 = malloc(strlen(src_directory) + 2 + strlen(file1));
+        sprintf(src_path1, "%s/%s", src_directory, file1);
+
+        src_path2 = malloc(strlen(src_directory) + 2 + strlen(file2));
+        sprintf(src_path2, "%s/%s", src_directory, file2);
```
daandemeyer: We have `path_join()` for this

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+        char dst_directory[] = "/tmp/dirXXXXXX";
+
+        char file1[] = "foo_file";
+        char file2[] = "bar_file";
+
+        /* Create the random source/destination directories */
+        r = mkdtemp(src_directory);
+        assert_se(r);
+        r = mkdtemp(dst_directory);
+        assert_se(r);
```
daandemeyer: Check the `copy_atomic` test for how to do this. The idea is to use `mkdtemp_malloc()` in combination with `_cleanup_(rm_rf_physical_and_freep)`

src/test/test-copy.c
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+}
+
+TEST(copy_tree_replace_dirs) {
+        _cleanup_free_ char *buf = NULL;
+        size_t sz = 0;
+        char *r;
+        char *src_path1;
+        char *src_path2;
+        char *dst_path1;
+        char *dst_path2;
```
daandemeyer: We declare multiple declarations of the same type on the same line, so:

```
char *src_path1, *src_path2, ...
```
```

File: src/test/test-copy.c [hunk 20]
```diff
@@ -50,6 +51,146 @@ TEST(copy_file) {
...
+        sprintf(dst_path2, "%s/%s", dst_directory, file2);
+
+        /* Populate some data to differentiate the files. */
+        assert_se(write_string_file(src_path1, "src file 1", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(src_path2, "src file 2", WRITE_STRING_FILE_CREATE) == 0);
+
+        assert_se(write_string_file(dst_path1, "dest file 1", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(dst_path2, "dest file 2", WRITE_STRING_FILE_CREATE) == 0);
+
+        {
```
daandemeyer: Why use a nested block here?

src/shared/copy.h
rdtscp: The reason I declared this here is because `fd_copy_directory` calls this, and this calls `fd_copy_directory`.
I could have forward declared in `.c` but thought this would be better.
Shall I fwd declare?

src/test/test-copy.c
rdtscp: I left the size checks in because my original implementation mis-used `O_EXCL` which would have resulted in incorrect sizes (so my intention was for these checks to catch such an issue).
I copied it from a previous test as well so i assumed it was necessary, I can remove though

src/test/test-copy.c
rdtscp: Just to appease my eyes in separating the logic I am testing, and the validation of the state before/after. I can remove if it doesn't fit

src/shared/copy.h
daandemeyer: Forward declaring in .c should be fine

src/test/test-copy.c
daandemeyer: Ahh, that makes sense, in that case, let's keep the size checks.

Althoughn wouldn't that still get caught by the string comparison check?

src/test/test-copy.c
daandemeyer: I don't mind, let's see if someone else complains

src/test/test-copy.c
rdtscp: I thought it would be caught by streq :P but I assumed I was missing something and included it anyway

src/machine/machine-dbus.c
```diff
@@ -928,6 +930,22 @@ int bus_machine_method_copy(sd_bus_message *message, void *userdata, sd_bus_erro
         if (r < 0)
                 return r;
 
+        call_contains_flags = sd_bus_message_is_method_call(message, NULL, "CopyFromMachineWithFlags") || sd_bus_message_is_method_call(message, NULL, "CopyToMachineWithFlags")
+                || sd_bus_message_is_method_call(message, NULL, "CopyFromWithFlags") || sd_bus_message_is_method_call(message, NULL, "CopyToWithFlags");
+
+        if (call_contains_flags) {
+                uint64_t raw_flags;
+                r = sd_bus_message_read(message, "t", &raw_flags);
```
daandemeyer, line 934: We always put a newline between declarations and the first statement

src/machine/machinectl.c
```diff
@@ -1101,6 +1102,7 @@ static int copy_files(int argc, char *argv[], void *userdata) {
         sd_bus *bus = userdata;
         bool copy_from;
         int r;
+        CopyFileFlags copy_flags;
```
daandemeyer: This should be initialized to zero

src/machine/machined-dbus.c
```diff
@@ -19,6 +19,7 @@
 #include "format-util.h"
 #include "hostname-util.h"
 #include "image-dbus.h"
+#include "install.h"
```
daandemeyer: This can be dropped

src/machine/machine-dbus.c
```diff
@@ -928,6 +930,22 @@ int bus_machine_method_copy(sd_bus_message *message, void *userdata, sd_bus_erro
         if (r < 0)
                 return r;
 
+        call_contains_flags = sd_bus_message_is_method_call(message, NULL, "CopyFromMachineWithFlags") || sd_bus_message_is_method_call(message, NULL, "CopyToMachineWithFlags")
+                || sd_bus_message_is_method_call(message, NULL, "CopyFromWithFlags") || sd_bus_message_is_method_call(message, NULL, "CopyToWithFlags");
```
daandemeyer: My bad, it turns out there is a way to get the member from the D-Bus message (`sd_bus_message_get_member()`). You can use that and then check if that string ends with "WithFlags". We have the `endswith()` function for that.

src/machine/machine-dbus.c
```diff
@@ -25,6 +25,7 @@
 #include "format-util.h"
 #include "fs-util.h"
 #include "in-addr-util.h"
+#include "install.h"
```
daandemeyer: This can be dropped

src/shared/copy.c
```diff
@@ -44,6 +44,8 @@
  * case of bind mount cycles and suchlike. */
 #define COPY_DEPTH_MAX 2048U
 
+int fd_copy_tree_generic(int df, const char *from, const struct stat *st, int dt, const char *to, dev_t original_device, unsigned depth_left, uid_t override_uid, gid_t override_gid, CopyFlags copy_flags, HardlinkContext *hardlink_context, const char *display_path, copy_progress_path_t progress_path, copy_progress_bytes_t progress_bytes, void *userdata);
```
daandemeyer: This should be formatted the same as the definition

src/shared/copy.h
```diff
@@ -27,6 +27,18 @@ typedef enum CopyFlags {
...
+typedef struct HardlinkContext {
+        int dir_fd;    /* An fd to the directory we use as lookup table. Never AT_FDCWD. Lazily created, when
+                        * we add the first entry. */
+
+        /* These two fields are used to create the hardlink repository directory above — via
+         * mkdirat(parent_fd, subdir) — and are kept so that we can automatically remove the directory again
+         * when we are done. */
+        int parent_fd; /* Possibly AT_FDCWD */
+        char *subdir;
+} HardlinkContext;
```
daandemeyer: This should be moved back to copy.c

man/org.freedesktop.machine1.xml
```diff
@@ -394,8 +408,10 @@ node /org/freedesktop/machine1 {
 
       <para><function>CopyFromMachine()</function> copies files or directories from a container into the
       host. It takes a container name, a source directory in the container and a destination directory on the
-      host as arguments. <function>CopyToMachine()</function> does the opposite and copies files from a source
-      directory on the host into a destination directory in the container.</para>
+      host as arguments.
+      <function>CopyToMachine()</function> does the opposite and copies files from a source
+      directory on the host into a destination directory in the container.
+      <function>CopyFromMachineWithFlags()</function> and <function>CopyToMachineWithFlags</function> do the same but can pass flags</para>
```
daandemeyer: Missing '.' at the end of the sentence.

"but can pass flags" => "but take an additional flags argument".

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
         unlink(fn_copy);
 }
 
+TEST(copy_tree_replace_file) {
+        _cleanup_free_ char *buf = NULL;
+        char src[] = "/tmp/test-copy_file.XXXXXX";
+        char dest[] = "/tmp/test-copy_file.XXXXXX";
+
```
daandemeyer, line 62: We don't put newlines between consecutive declarations.

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
...
+        size_t sz = 0;
```
daandemeyer: No need to initialize `sz`, on a successful call to `read_full_file()` it will always be set.

DaanDeMeyer (changes requested):
Some more comments

src/machine/machinectl.c
poettering: Oh, actually, that's not really true, we *do* prefer to use the old methods, because people use -M or -H to access remote systems or container environments.

Hence, it's *good* to use the old calls when we can.

src/shared/copy.c
```diff
@@ -1034,6 +1014,61 @@ static int fd_copy_directory(
...
+                copy_progress_bytes_t progress_bytes,
+                void *userdata) {
+        int r;
+
+        if (S_ISDIR(st->st_mode))
+                r = fd_copy_directory(df, from, st, dt, to, original_device, depth_left-1, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_path, progress_bytes, userdata);
+        else
+                r = fd_copy_leaf(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_bytes, userdata);
+
+        return r;
```
poettering, line 1105: what about:

```c
assert(st);

if (S_ISDIR(st->st_mode))
        return fd_copy_directory(…);

return fd_copy_leaf(…);
```

shorter, and no extra local variable

src/shared/copy.c
```diff
@@ -1034,6 +1014,61 @@ static int fd_copy_directory(
...
+                r = fd_copy_fifo(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context);
+        else if (S_ISBLK(st->st_mode) || S_ISCHR(st->st_mode) || S_ISSOCK(st->st_mode))
+                r = fd_copy_node(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context);
+        else
+                r = -EOPNOTSUPP;
+
+        return r;
+}
+
+int fd_copy_tree_generic(
```
poettering: why not static?

src/shared/copy.c
```diff
@@ -502,18 +504,6 @@ static int fd_copy_symlink(
...
-typedef struct HardlinkContext {
-        int dir_fd;    /* An fd to the directory we use as lookup table. Never AT_FDCWD. Lazily created, when
-                        * we add the first entry. */
-
-        /* These two fields are used to create the hardlink repository directory above — via
-         * mkdirat(parent_fd, subdir) — and are kept so that we can automatically remove the directory again
-         * when we are done. */
-        int parent_fd; /* Possibly AT_FDCWD */
-        char *subdir;
-} HardlinkContext;
```
poettering: no need to move this?

src/shared/copy.c
```diff
@@ -1063,8 +1063,17 @@ int fd_copy_tree_generic(
...
         if (S_ISDIR(st->st_mode))
                 r = fd_copy_directory(df, from, st, dt, to, original_device, depth_left-1, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_path, progress_bytes, userdata);
-        else
+        else {
                 r = fd_copy_leaf(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_bytes, userdata);
+                /* We just tried to copy a leaf node of the tree. If it failed because the node already exists *and* the COPY_REPLACE flag has been provided, we should unlink the node and re-copy. */
+                if (r == -EEXIST && (copy_flags & COPY_REPLACE)) {
+                        r = unlinkat(dt, to, 0);
+                        if (r < 0)
+                                return r;
```
poettering: unlinkat() returns the error code in `errno`, and not as a negative return value.

(i.e. unlinkat() is a posix function, which uses `errno`, but in systemd we prefer negative error codes as return values instead. So you need to be careful, picking the right approach each time)

That said, I think if unlinkat() fails, it actually makes sense to propagate the error returned by fd_copy_leaf() instead of the failure code of unlinkat(), since we just tried to fix an error, here, but the original error is probably more interesting than our failed failure path...

hence:

```c
if (unlinkat(dt, to, 0) < 0)
        return -EEXIST; /* return original error */
```

src/test/test-copy.c
poettering: these files are not removed once the test completes, but they should.

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
...
+        int fd;
+
+        fd = mkostemp_safe(src);
+        assert_se(fd >= 0);
+        close(fd);
```
poettering: we generall prefer safe_close()

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
...
+
+        assert_se(write_string_file(src, "bar bar", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(dest, "foo foo foo", WRITE_STRING_FILE_CREATE) == 0);
+
+        assert_se(read_full_file(dest, &buf, &sz) == 0);
+        assert_se(streq(buf, "foo foo foo\n"));
+        assert_se(sz == 12);
+
+        // The file has been copied once - now overwite original contents, and test the COPY_REPLACE flag.
```
poettering: we intend to use // style comments for "FIXME" kind of comments only. Explanatory comments should really use /* */

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
...
+
+        assert_se(copy_tree(src, dest, UID_INVALID, GID_INVALID, COPY_REFLINK) != 0);
```
poettering: `!= 0`?

usually one would do `< 0` to check for errors or '>= 0` to check for success...

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
...
+        _cleanup_free_ char *src_path1, *src_path2, *dst_path1, *dst_path2;
+
+        char src_directory[] = "/tmp/dirXXXXXX";
+        char dst_directory[] = "/tmp/dirXXXXXX";
+
+        char file1[] = "foo_file";
+        char file2[] = "bar_file";
+
+        /* Create the random source/destination directories */
+        r = mkdtemp(src_directory);
```
poettering: maybe do:

```c
_cleanup_(rm_rf_physical_freep) char *src_directory = NULL;

assert_se(mkdtemp_malloc(&src_directory, "/tmp/dirXXXXXX") >= 0);
```

src/test/test-copy.c
```diff
@@ -50,6 +51,135 @@ TEST(copy_file) {
...
+        assert_se(r);
+        r = mkdtemp(dst_directory);
+        assert_se(r);
+
+        /* Construct the source/desitnation filepaths (should have different dir name, but same file names within) */
+        src_path1 = path_join(src_directory, file1);
+        src_path2 = path_join(src_directory, file2);
+        dst_path1 = path_join(dst_directory, file1);
+        dst_path2 = path_join(dst_directory, file2);
```
poettering: these can fail, due to oom. better catch that by surrounding this with `assert_se()`

poettering (changes requested):
Looks good, just some minor things

src/machine/machinectl.c
rdtscp: So to be explicit, is it preferred that I use the "CopyFromMachine" method when no flags are necessary, and use "...WithFlags" only when there are flags?
- [ ] Check

src/machine/machine-dbus.c
```diff
@@ -928,6 +929,22 @@ int bus_machine_method_copy(sd_bus_message *message, void *userdata, sd_bus_erro
         if (r < 0)
                 return r;
 
+        call_contains_flags = endswith(sd_bus_message_get_member(message), "WithFlags");
```
daandemeyer: I wouldn't bother with the local variable anymore now and just inline it into the `if`

src/machine/machinectl.c
```diff
@@ -1119,20 +1121,24 @@ static int copy_files(int argc, char *argv[], void *userdata) {
...
 
+        if (arg_force)
+                copy_flags |= COPY_FILE_REPLACE;
+
         r = bus_message_new_method_call(
                         bus,
                         &m,
                         bus_machine_mgr,
-                        copy_from ? "CopyFromMachine" : "CopyToMachine");
+                        copy_from ? "CopyFromMachineWithFlags" : "CopyToMachineWithFlags");
```
daandemeyer: You might have missed the comment but Lennart mentioned it's actually good to keep using the old methods if --force is not used, because machinectl can be used to call D-Bus methods on a remote host that's running an older version of systemd-machined. Sorry for the churn. Maybe add a function like this that returns the correct method name:

```
static const char *select_copy_method(bool copy_from, bool force) {
        ...
}
```

And then after adding the common arguments, check if `arg_force` is true and if so, append one extra argument with the COPY_REPLACE flag.

```
if (arg_force) {
        r = sd_bus_message_append(m, "t", COPY_REPLACE);
        ...       
}

File: src/shared/copy.c [hunk 42]
```diff
@@ -1034,6 +1041,68 @@ static int fd_copy_directory(
...
+        int r;
+
+        if (S_ISDIR(st->st_mode))
+                return fd_copy_directory(df, from, st, dt, to, original_device, depth_left-1, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_path, progress_bytes, userdata);
+
+        r = fd_copy_leaf(df, from, st, dt, to, override_uid, override_gid, copy_flags, hardlink_context, display_path, progress_bytes, userdata);
+        /* We just tried to copy a leaf node of the tree. If it failed because the node already exists *and* the COPY_REPLACE flag has been provided, we should unlink the node and re-copy. */
+        if (r == -EEXIST && (copy_flags & COPY_REPLACE)) {
+                if (unlinkat(dt, to, 0) < 0)
+                        return r;
```
daandemeyer, line 1100: I'd add a comment here explaining that we want to return the original error instead of `errno` if this fails.

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
         unlink(fn_copy);
 }
 
+TEST(copy_tree_replace_file) {
+        _cleanup_free_ char *buf = NULL;
+        char *src, *dest;
```
daandemeyer: Missing _cleanup_free_ for these (and they can then just be merged into the first declaration)

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+        size_t sz;
+        int fd;
```
daandemeyer: This can be dropped

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+        assert_se(write_string_file(src, "bar bar", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(dest, "foo foo foo", WRITE_STRING_FILE_CREATE) == 0);
+
+        assert_se(read_full_file(dest, &buf, &sz) == 0);
+        assert_se(streq(buf, "foo foo foo\n"));
+        assert_se(sz == 12);
+
+        /* The file exists- now overwite original contents, and test the COPY_REPLACE flag. */
+
+        assert_se(copy_tree(src, dest, UID_INVALID, GID_INVALID, COPY_REFLINK) < 0);
```
daandemeyer: I'd check specifically that this returns `-EEXIST`

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+
+        assert_se(read_full_file(dest, &buf, &sz) == 0);
+        assert_se(streq(buf, "bar bar\n"));
+        assert_se(sz == 8);
+}
+
+TEST(copy_tree_replace_dirs) {
+        _cleanup_free_ char *buf = NULL;
+        size_t sz;
+        _cleanup_free_ char *src_path1, *src_path2, *dst_path1, *dst_path2;
```
daandemeyer: _cleanup_free_ variables should always be initialized to `NULL`. Since the cleanup function is always invoked, and if the pointer still points to uninitialized memory, the cleanup function will try to free random memory, leading to a segmentation fault.

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+        _cleanup_(rm_rf_physical_and_freep) char *src_directory = NULL;
+        _cleanup_(rm_rf_physical_and_freep) char *dst_directory = NULL;
+        char file1[] = "foo_file";
+        char file2[] = "bar_file";
```
daandemeyer: I'd inline these two variables

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+                assert_se(streq(buf, "dest file 1\n"));
+                assert_se(sz == 12);
+
+                assert_se(read_full_file(dst_path2, &buf, &sz) == 0);
+                assert_se(streq(buf, "dest file 2\n"));
+                assert_se(sz == 12);
+        }
+
+        /* Copying without COPY_REPLACE should fail because the destination file already exists. */
+        assert_se(copy_tree(src_directory, dst_directory, UID_INVALID, GID_INVALID, COPY_REFLINK) < 0);
```
daandemeyer: Ditto for checking that this returns `-EEXIST`

src/test/test-copy.c
daandemeyer: Also, you use `dest` here but `dst` in the second test, I'd be consistent and use the same for both tests

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+                assert_se(sz == 11);
+
+                assert_se(read_full_file(dst_path1, &buf, &sz) == 0);
+                assert_se(streq(buf, "dest file 1\n"));
+                assert_se(sz == 12);
+
+                assert_se(read_full_file(dst_path2, &buf, &sz) == 0);
+                assert_se(streq(buf, "dest file 2\n"));
+                assert_se(sz == 12);
+        }
```
daandemeyer: I'd drop these assertions, we're testing that `copy_tree()` works, not `write_string_file()`

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
...
+
+        assert_se(tempfn_random("/tmp/test-copy_file.XXXXXX", NULL, &src) >= 0);
+        assert_se(tempfn_random("/tmp/test-copy_file.XXXXXX", NULL, &dest) >= 0);
+
+        assert_se(write_string_file(src, "bar bar", WRITE_STRING_FILE_CREATE) == 0);
+        assert_se(write_string_file(dest, "foo foo foo", WRITE_STRING_FILE_CREATE) == 0);
+
+        assert_se(read_full_file(dest, &buf, &sz) == 0);
+        assert_se(streq(buf, "foo foo foo\n"));
+        assert_se(sz == 12);
```
daandemeyer: Ditto, I'd drop this, we're testing `copy_tree()`, not `write_string_file()`

src/test/test-copy.c
daandemeyer: These can then also be merged with the first declaration.

src/test/test-copy.c
```diff
@@ -50,6 +51,123 @@ TEST(copy_file) {
         unlink(fn_copy);
 }
 
+TEST(copy_tree_replace_file) {
+        _cleanup_free_ char *buf = NULL;
+        char *src, *dest;
+        size_t sz;
```
daandemeyer: `streq()` will guarantee both strings are the same size so we can drop the size checks

DaanDeMeyer (changes requested):
Almost good to go! Just a few more things to address

src/test/test-copy.c
rdtscp: I opted not to here because I make use of the string multiple times (i.e. I'd have to inline the same raw string in multiple places).

daandemeyer: Makes sense, but then I think these can be made `const char *` instead of `char[]`

src/machine/machinectl.c
```diff
@@ -1119,14 +1127,16 @@ static int copy_files(int argc, char *argv[], void *userdata) {
...
         r = bus_message_new_method_call(
                         bus,
                         &m,
                         bus_machine_mgr,
-                        copy_from ? "CopyFromMachine" : "CopyToMachine");
+                        select_copy_method(copy_from, arg_force));
         if (r < 0)
                 return bus_log_create_error(r);
 
+
```
daandemeyer: Spurious newline here

src/machine/machined-dbus.c
```diff
@@ -539,6 +539,10 @@ static int method_copy_machine(sd_bus_message *message, void *userdata, sd_bus_e
         return redirect_method_to_machine(message, userdata, error, bus_machine_method_copy);
 }
 
+static int method_copy_machine_with_flags(sd_bus_message *message, void *userdata, sd_bus_error *error) {
+        return redirect_method_to_machine(message, userdata, error, bus_machine_method_copy);
+}
```
daandemeyer: Do we actually need this? I figure we can just reuse `method_copy_machine()` no?

src/shared/copy.c
```diff
@@ -992,18 +1009,8 @@ static int fd_copy_directory(
...
-                        q = fd_copy_regular(dirfd(d), de->d_name, &buf, fdt, de->d_name, override_uid, override_gid, copy_flags, hardlink_context, progress_bytes, userdata);
-                else if (S_ISLNK(buf.st_mode))
-                        q = fd_copy_symlink(dirfd(d), de->d_name, &buf, fdt, de->d_name, override_uid, override_gid, copy_flags);
-                else if (S_ISFIFO(buf.st_mode))
-                        q = fd_copy_fifo(dirfd(d), de->d_name, &buf, fdt, de->d_name, override_uid, override_gid, copy_flags, hardlink_context);
-                else if (S_ISBLK(buf.st_mode) || S_ISCHR(buf.st_mode) || S_ISSOCK(buf.st_mode))
-                        q = fd_copy_node(dirfd(d), de->d_name, &buf, fdt, de->d_name, override_uid, override_gid, copy_flags, hardlink_context);
-                else
-                        q = -EOPNOTSUPP;
+                }
```
daandemeyer, line 1012: Missing newline here

src/test/test-copy.c
```diff
@@ -1,5 +1,7 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
+#include <openssl/ossl_typ.h>
```
daandemeyer: Seems clangd added this header automatically.

src/test/test-copy.c
```diff
@@ -1,5 +1,7 @@
...
+#include <string.h>
```
daandemeyer: Do we need this include? I'm pretty sure it'll all still compile without adding this

src/test/test-copy.c
```diff
@@ -50,6 +52,86 @@ TEST(copy_file) {
...
+        assert_se(copy_tree(src, dst, UID_INVALID, GID_INVALID, COPY_REFLINK|COPY_REPLACE) == 0);
+
+        assert_se(read_full_file(dst, &buf, NULL) == 0);
+        assert_se(streq(buf, "bar bar\n"));
+}
+
+TEST(copy_tree_replace_dirs) {
+        _cleanup_free_ char *buf = NULL, *src_path1 = NULL, *src_path2 = NULL, *dst_path1 = NULL, *dst_path2 = NULL;
+        _cleanup_(rm_rf_physical_and_freep) char *src_directory = NULL;
+        _cleanup_(rm_rf_physical_and_freep) char *dst_directory = NULL;
```
daandemeyer: These can also be merged

src/test/test-copy.c
```diff
@@ -50,6 +52,86 @@ TEST(copy_file) {
...
+        const char *file1 = "foo_file", *file2 = "bar_file";
+
+        /* Create the random source/destination directories */
+        assert_se(mkdtemp_malloc("/tmp/dirXXXXXX", &src_directory) >= 0);
+        assert_se(mkdtemp_malloc("/tmp/dirXXXXXX", &dst_directory) >= 0);
+
+        /* Construct the source/desitnation filepaths (should have different dir name, but same file names within) */
```
daandemeyer: "desitnation" > "destination"

src/machine/machine-dbus.h
```diff
@@ -6,6 +6,11 @@
 #include "bus-util.h"
 #include "machine.h"
 
+typedef enum {
+        COPY_FILE_REPLACE = 1 << 0, /* Public API via DBUS, do not change */
+        _COPY_FILE_FLAGS_MASK_PUBLIC = COPY_FILE_REPLACE,
+} CopyFileFlags;
```
daandemeyer: I'd rename the enum to `MachineCopyFlags` (with `MACHINE_COPY_REPLACE`, ...)

DaanDeMeyer (changes requested):
Some final nits and then it's good to go!

src/machine/machinectl.c
```diff
@@ -1119,11 +1127,12 @@ static int copy_files(int argc, char *argv[], void *userdata) {
                 host_path = abs_host_path;
         }
 
+
```
daandemeyer: Spurious newline here? (in the first commit)

rdtscp: Ah perhaps I added it in wrong place from previous comment, let me check sorry about that one

src/test/test-copy.c
```diff
@@ -50,6 +50,85 @@ TEST(copy_file) {
         unlink(fn_copy);
 }
 
+TEST(copy_tree_replace_file) {
+        _cleanup_free_ char *buf = NULL;
+        char *src, *dst;
```
daandemeyer: Missing `_cleanup_free_` on these variables (and they should be merged into the above declaration)

DaanDeMeyer (changes requested):

src/test/test-copy.c
```diff
@@ -50,6 +50,76 @@ TEST(copy_file) {
         unlink(fn_copy);
 }
 
+static bool read_file_and_streq(const char* filepath, const char* expected_contents) {
+        _cleanup_free_ char *buf = NULL;
+        if (read_full_file(filepath, &buf, NULL) != 0)
```
daandemeyer: Missing newline between declarations and first statement

daandemeyer: Also, you can just use `assert_se()` here, that makes it easier to figure out what failed if a test fails.

DaanDeMeyer (approved):

PR #24081, opened 2022-07-22
journal: Move more pattern matching logic into pcre2-util

daandemeyer:
To avoid having "#if HAVE_PCRE2" all throughout the code, let's
confine the pcre2 header specific stuff to pcre2-util.c. Instead of
exposing all the individual symbols from pcre2, let's only expose
three high level functions that do all we need:

- pcre2_pattern_compile(): Compile the regex
- pcre2_pattern_matches(): Check if the compiled regex matches a message
- pcre2_pattern_free(): Free the compiled regex

We expose the compiled pcre2 pattern (which is of type pcre2_code *) as
a void pointer to avoid having to include pcre2.h in all code where we
work with compiled pcre2 patterns. For readability, we typedef void
to pcre2_pattern and use that as the type specifier for compiled pcre2
patterns.

### DIFF
--- a/src/shared/pcre2-util.c
+++ b/src/shared/pcre2-util.c
@@ -14,8 +14,10 @@ pcre2_code* (*sym_pcre2_compile)(PCRE2_SPTR, PCRE2_SIZE, uint32_t, int *, PCRE2_
 int (*sym_pcre2_get_error_message)(int, PCRE2_UCHAR *, PCRE2_SIZE);
 int (*sym_pcre2_match)(const pcre2_code *, PCRE2_SPTR, PCRE2_SIZE, PCRE2_SIZE, uint32_t, pcre2_match_data *, pcre2_match_context *);
 PCRE2_SIZE* (*sym_pcre2_get_ovector_pointer)(pcre2_match_data *);
+#endif
 
 int dlopen_pcre2(void) {
+#if HAVE_PCRE2
         /* So here's something weird: PCRE2 actually renames the symbols exported by the library via C
          * macros, so that the exported symbols carry a suffix "_8" but when used from C the suffix is
          * gone. In the argument list below we ignore this mangling. Surprisingly (at least to me), we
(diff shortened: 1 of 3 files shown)


src/shared/pcre2-util.h
```diff
@@ -3,21 +3,20 @@
 
 #include "macro.h"
 
-#if HAVE_PCRE2
+typedef void pcre2_pattern;
```
poettering: typedeffing this to void kinda sucks, since it disables all type-checking...

I think it would be nicer to define this as anonymous struct instead. This then means you have to cast, but that should be safe if you do it in a static/inline cast function.

poettering: or just define it either against the anonymous struct or against the real type, depending on HAVE_PCRE2. After all, you do that in a header file specific to pcre2-util, where it shouldn't matter...

poettering:
so in the rest of our codebase we tend to make the imported dlsym symbols public. Here you intend to hide them, but I am not sure that's really a good idea, after all the APIs *are* public anyway, because these are public functions.

Hence, i think the helper calls that allow us to remove a lot of ifdeffery from the consuming code makes a ton of sense, but the hiding of the sym_xyz variables (and thus not including the pcre2 headers into all consumers of that) I find less attractive...

src/shared/pcre2-util.c
```diff
@@ -33,12 +41,118 @@ int dlopen_pcre2(void) {
...
+        }
+
+        if (ret)
+                *ret = p;
+
+        return 0;
 #else
+        return -EOPNOTSUPP;
+#endif
+}
```
poettering, line 100: so this function logs about all errors, but is still library-like. Probably should be called …and_log() or so...

i.e. see CODING_STYLE regarding our rule that library like functions shouldn't log on their own, unless they make that clear in the name, or something to disable with a flags arg or so

src/shared/pcre2-util.c
```diff
@@ -33,12 +41,118 @@ int dlopen_pcre2(void) {
...
+
+        return 1;
+#else
+        return -EOPNOTSUPP;
+#endif
 }
+
+void *pcre2_pattern_free(pcre2_pattern *p) {
+#if HAVE_PCRE2
+        assert(pcre2_dl);
```
poettering: hmm, I think this function should be a NOP if p=NULL is passed, hence this assert should go, or be moved down before the sym_pcre2_code_free() calls

src/shared/pcre2-util.h
```diff
@@ -3,21 +3,20 @@
...
-extern pcre2_match_data* (*sym_pcre2_match_data_create)(uint32_t, pcre2_general_context *);
-extern void (*sym_pcre2_match_data_free)(pcre2_match_data *);
-extern void (*sym_pcre2_code_free)(pcre2_code *);
-extern pcre2_code* (*sym_pcre2_compile)(PCRE2_SPTR, PCRE2_SIZE, uint32_t, int *, PCRE2_SIZE *, pcre2_compile_context *);
-extern int (*sym_pcre2_get_error_message)(int, PCRE2_UCHAR *, PCRE2_SIZE);
-extern int (*sym_pcre2_match)(const pcre2_code *, PCRE2_SPTR, PCRE2_SIZE, PCRE2_SIZE, uint32_t, pcre2_match_data *, pcre2_match_context *);
-extern PCRE2_SIZE* (*sym_pcre2_get_ovector_pointer)(pcre2_match_data *);
+int pcre2_pattern_compile(const char *pattern, PatternCompileCase case_, pcre2_pattern **ret);
+int pcre2_pattern_matches(pcre2_pattern *compiled_pattern, const char *message, size_t size, size_t *ret_ovec);
+void *pcre2_pattern_free(pcre2_pattern *p);
```
poettering: we are kinda invading pcre2's own function naming namespace if we name these helper calls pcre2_xyz.

we do something like at variou splaces, but maybe we shouldn't...

Maybe drop the pcre2_ prefix, and just use pattern_ as prefix? after all the enum above you already callsed Pattern… hence this would match?

DaanDeMeyer:
Changed to keep the symbols public, and addressed all the rest of the comments, PTAL!

src/shared/pcre2-util.c
```diff
@@ -33,12 +35,118 @@ int dlopen_pcre2(void) {
...
+                        return log_oom();
+
+                r = pattern_compile_and_log("[[:upper:]]", PATTERN_COMPILE_CASE_SENSITIVE, &cs);
+                if (r < 0)
+                        return r;
+
+                r = sym_pcre2_match(cs, (PCRE2_SPTR8) pattern, PCRE2_ZERO_TERMINATED, 0, 0, md, NULL);
+                has_case = r >= 0;
+
+                flags = !has_case * PCRE2_CASELESS;
```
qdeslandes, line 74: This part should probably be extracted and the compiled pattern cached.

daandemeyer: pattern_compile_and_log() shouldn't be on the hot path, since we should only compile patterns once. If we'd need to cached this pattern, we'd either need to bother every user with taking care of freeing the pattern etc, or use thread local cache variables, at which point valgrind might start to complain because the value is never freed (which we could fix by calling a free function at the end of the process, but then we'd need to take care to do so everywhere, which would be a burden as well.

So if this was a hot path we should definitely do this, but since it isn't, I figure it's not a problem to compile this pattern every time we need to compile another pattern. Also, this is only done if the "smart" logic is used, which we'll probably only do in journalctl, and not in the journald log regex stuff I think?

poettering:
lgtm, but maybe @keszybz should give the final blessing as he did the original PCRE work

DaanDeMeyer:
Made one more change to call dlopen_pcre2() lazily in the compile and matches functions

src/shared/pcre2-util.c
```diff
@@ -33,12 +35,124 @@ int dlopen_pcre2(void) {
...
+                                       "Bad pattern \"%s\": %s", pattern,
+                                       r < 0 ? "unknown error" : (char *)buf);
+        }
+
+        if (ret)
+                *ret = p;
+
+        return 0;
 #else
+        return -EOPNOTSUPP;
```
poettering: should be `return log_error_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "Support for pcre2 not compiled in.")`

(i.e. this function should log all errors, and this is an error.

src/shared/pcre2-util.c
```diff
@@ -33,12 +35,124 @@ int dlopen_pcre2(void) {
                         DLSYM_ARG(pcre2_get_error_message),
                         DLSYM_ARG(pcre2_match),
                         DLSYM_ARG(pcre2_get_ovector_pointer));
+#else
+        return log_debug_errno(SYNTHETIC_ERRNO(EOPNOTSUPP), "PCRE2 support is not compiled in.");
```
poettering: the dlopen_many_sym_or_warn() call will log at LOG_ERR level, but here you only log at debug level. The function should log on the same level in either case.

src/shared/pcre2-util.c
```diff
@@ -33,12 +35,124 @@ int dlopen_pcre2(void) {
...
+#if HAVE_PCRE2
+        _cleanup_(sym_pcre2_match_data_freep) pcre2_match_data *md = NULL;
+        int r;
+
+        assert(compiled_pattern);
+        assert(message);
+
+        r = dlopen_pcre2();
+        if (r < 0)
+                return r;
```
poettering: replaces these three lines with (`assert(pcre2_dl);`). After all, if you have a compiled pattern object, then PCRE2 *must* have been successfully loaded.

poettering:
looks good, but there is still some confusion about logging, i.e. all branches of the functions should behave the same: i.e. log all errors or log none, as per coding style, see my comments above.

src/shared/pcre2-util.c
daandemeyer: It logs at LOG_ERR if a symbol cannot be found, but it only logs at LOG_DEBUG if the library is not installed, so the case is not as clear cut. Similarly, dlopen_bpf() also only logs at debug level if BPF support was not compiled in.

I think we shouldn't log at higher level than DEBUG when checking if a library is available or compiled in, especially since we might call a function like this to figure out if a particular library is available or not. That would lead to a lot of log spam overall. That's why here I think we should deviate from either always logging or never logging, since the logging when as symbol is missing in dlopen_many_sym_or_warn() is very useful and we should keep that.

DaanDeMeyer:
> looks good, but there is still some confusion about logging, i.e. all branches of the functions should behave the same: i.e. log all errors or log none, as per coding style, see my comments above.

I generally don't want these functions to log and behave like library functions, but getting the error message from pcre2 after one of these functions returns seems like it'd be more complex than we'd want so that's why I kept the logging inside the function. For stuff like returning EOPNOTSUPP I don't think we should log since someone might call these functions and want to gracefully handle EOPNOTSUPP themselves, so it doesn't seem to make sense to always error log since that would lead to log spam.

DaanDeMeyer:
Added back the verbose logging. I'll propose the logging changes in another PR. Setting the green label.

poettering:
> For stuff like returning EOPNOTSUPP I don't think we should log since someone might call these functions and want to gracefully handle EOPNOTSUPP themselves, so it doesn't seem to make sense to always error log since that would lead to log spam.

It is OK to special case some errors, and not log about them, but if we do this, there should be a big fat comment there, explaining that this is done specially, and we expect the caller do either handle it gracefully or log. We do this at other places already.

i.e. the rules about logging/error propagation may be bent, but if so, a comment must clarify that.

poettering:
lgtm

PR #24084, opened 2022-07-22
man: Sysext docs improvements

pothos:
- man: Document mask workaround for sysext images
    
    A read-only /usr may ship a sysext image by default and the user wants
    to opt out. Currently it's not clear how to do this.
    Document that a /dev/null symlink in /etc/extensions/ works to "mask" a
    sysext image in a folder with lower precedence.

- man: Do not recommend to overlay files with sysext even if possible
    
    While overlaying files with a sysext can be useful, it may lead to
    unexpected problems depending on when a process got started and which
    version of the file it gets.
    Call out that overlaying files is possible but don't recommend to make
    use of it.

### DIFF
--- a/man/systemd-sysext.xml
+++ b/man/systemd-sysext.xml
@@ -108,7 +108,9 @@
     accessed.</para>
 
     <para>Note that there is no concept of enabling/disabling installed system extension images: all
-    installed extension images are automatically activated at boot.</para>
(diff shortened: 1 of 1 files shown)


bluca (approved):

poettering:
lgtm

Issue #15273, opened 2020-03-30
homed: cannot create home on partition block devices (only whole disk block devices work)


pltrz:
**systemd version the issue has been seen with**
> ```
> $ systemctl --version
> systemd 245 (245.3-2-arch)
> +PAM +AUDIT -SELINUX -IMA -APPARMOR +SMACK -SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid
> ```

**Used distribution**
 > Arch Linux (2020-03-31)

**Expected behaviour you didn't see**
 > Create a user and home on a LUKS-encrypted block device

**Unexpected behaviour you saw**
 > `homectl` fails to create a user and home on a block device

**Steps to reproduce the problem**

I tried two ways, [1] with a vanilla partition, and [2] with an LVM logical volume.

--------------------------------------------

1. Vanilla partition

```
$ fdisk /dev/nvme0n1
  ...
```
Now the block device `/dev/nvme0n1p3` is available. It is a regular old partition under the GPT partition table.
```
$ homectl create foo --image-path=/dev/nvme0n1p3 --storage="luks"

Please enter new password for user foo: **********
Please enter new password for user foo (repeat): **********

Failed to create user home: Failed to execute operation: Block device required
```

In this case, homectl fails to create a home altogether - `homectl list` shows no homes were created.

---------------------------------------------

2. LVM logical volume

```
$ lvcreate -n home -L 5G volgroup
$ homectl create foo --image-path=/dev/volgroup/home --storage="luks"

Please enter new password for user foo: **********
Please enter new password for user foo (repeat): **********

$ homectl list
NAME UID   GID   STATE  REALNAME HOME      SHELL
foo  60484 60484 absent foo      /home/foo /bin/bash

$ homectl activate foo
Home of user foo is currently absent, please plug in the necessary stroage device or backing file system.
```

In this case, the home is created, but cannot be activated, and seems it cannot be activated by `homectl activate`. However, a user entry is definitely created, and can be shown with `homectl list` and `homectl inspect foo`.

poettering:
[...]

PR #24092, opened 2022-07-23
Use asynchronous discards with BTRFS on homed

AidanGG:
When using homed with `--luks-discard=true` and `--fs-type=btrfs`, prefer using asynchronous trim (https://btrfs.readthedocs.io/en/latest/Trim.html).

I need to write some tests but I figured I'd get some feedback first.

### DIFF
--- a/src/home/homework-mount.c
+++ b/src/home/homework-mount.c
@@ -5,6 +5,7 @@
 #include <linux/fs.h>
 
 #include "alloc-util.h"
+#include "condition.h"
 #include "fd-util.h"
 #include "format-util.h"
 #include "glyph-util.h"
@@ -61,7 +62,14 @@ int home_mount_node(
                         return log_oom();
         }
 
-        if (!strextend_with_separator(&joined, ",", discard ? "discard" : "nodiscard"))
+        const char *discard_mount_option = "discard";
+        if (streq(fstype, "btrfs")) {
+                Condition *c = condition_new(CONDITION_KERNEL_VERSION, ">= 5.6", false, false);
+                if (condition_test(c, NULL) > 0)
+                        discard_mount_option = "discard=async";
+                condition_free(c);
+        }
+        if (!strextend_with_separator(&joined, ",", discard ? discard_mount_option : "nodiscard"))
                 return log_oom();
 
         if (extra_mount_options) {


marcosfrm:
`discard=async` does not seem to work on pre-5.6 kernels.

https://kernelnewbies.org/Linux_5.6#Btrfs_asynchronous_SSD_trimming

```
# uname -a
Linux localhost 5.3.7-301.fc31.x86_64 #1 SMP Mon Oct 21 19:18:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

# mount /dev/sda4 /mnt/temp -o discard=async
mount: /mnt/temp: wrong fs type, bad option, bad superblock on /dev/sda4, missing codepage or helper program, or other error.

# dmesg | tail -2
[  444.429377] BTRFS info (device sda4): unrecognized mount option 'discard=async'
[  444.434370] BTRFS error (device sda4): open_ctree failed
```

src/home/homework-mount.c
```diff
@@ -61,7 +62,14 @@ int home_mount_node(
                         return log_oom();
         }
 
-        if (!strextend_with_separator(&joined, ",", discard ? "discard" : "nodiscard"))
+        const char *discard_mount_option = "discard";
+        if (streq(fstype, "btrfs")) {
+                Condition *c = condition_new(CONDITION_KERNEL_VERSION, ">= 5.6", false, false);
```
poettering, line 67: we frown on kernel version checks, since many distros (in particular commercial ones) are really keen on backporting features. That means version checks are not useful for checking feature availability.

Can we explicitly check if this feature is available already? i.e. btrfs has `/sys/fs/btrfs/features/`. Can we check availbility there? If not, is there maybe a closely related feature listed there that came at the same time, and unlikely to be skipped if async discard is backported?

What is the actual failure mode of the new discard option on old kernels? is it just ignored? or does the whole mount() syscall fail? If the latter, we could just try and fall back.

poettering:
Hmm, thinking about this: are you using homed with the LUKS backend on a loopback file? Or on a real SSD as backend?

if the loopback file stuff is used (which is the default), then the async discard shouldn't matter, as any trim on the file system would just translate to a hole punch on the backng file, and that should be super cheap. Hence, it appears to me that this option really should be enabled only if the backing storage is not a loopback file, but a real block device.

AidanGG:
I was using LUKS on a loopback file. Indeed I can't say that I'm using an external SSD or a separate internal SSD for my home directory.

If `--image-path` were extended to allow logical block devices (https://github.com/systemd/systemd/issues/15273), such as a dm-integrity volume, would async discards still be worthwhile (assuming the dm-integrity volume is set to pass discards down)? e.g. `/home/aidan` is a homed-managed home directory involving btrfs on LUKS directly on a dm-integrity volume (no loopback files).

poettering:
actually, I am wondering. Maybe the async discard logic might even make sense on loopback device backing store, to optimize fragmentation? i.e. maybe underlying file systems would benefit if they only have to punch holes a few times with big sizes than all the time with small sizes.

cmurf:
Tentatively `discard=async` will be enabled by default in a future kernel (it missed the 6.0 merge window so 6.1 at the earliest).
https://lore.kernel.org/linux-btrfs/20220727150158.GT13489@suse.cz/

poettering:
if we still need this, this can be hooked up with the new mount api logic which allows us to check if specific mount options are supported, see how mount_option_supported().

poettering:
ah, fuck, btrfs still hasn't been updated to the new mount api. so my suggestion wouldn't work.

so, hence did the default value of this setting change? if so, let's just close this I'd say.

marcosfrm:
Yes, `discard=async` is enabled by default since kernel 6.2:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=63a7cb13071842966c1ce931edacbc23573aada5

AidanGG:
Sure, closing. Thanks for the feedback.

PR #24094, opened 2022-07-23
systemctl: reduce variable scope and simplify

dtardon:


src/systemctl/systemctl-list-units.c
```diff
@@ -403,19 +402,18 @@ static int output_sockets_list(struct socket_info *socket_infos, unsigned cs) {
                                 return table_log_add_error(r);
 
                 }
-
```
yuwata, line 406: If you touch here, then could you also update `if (cs)` -> `if (cs > 0)` in the above?

src/systemctl/systemctl-list-units.c
```diff
@@ -403,19 +402,18 @@ static int output_sockets_list(struct socket_info *socket_infos, unsigned cs) {
...
 
         r = output_table(table);
         if (r < 0)
                 return r;
 
         if (arg_legend != 0) {
+                const char *on, *off;
+
+                on = cs > 0 ? ansi_highlight() : ansi_highlight_red();
+                off = ansi_normal();
```
yuwata: These are used only once. I think it is ok to inline them in the `printf()` below, if you'd like.

yuwata (approved):
LGTM. Several minor suggestions.

src/systemctl/systemctl-list-units.c
```diff
@@ -403,19 +402,18 @@ static int output_sockets_list(struct socket_info *socket_infos, unsigned cs) {
                                 return table_log_add_error(r);
 
                 }
-
-                on = ansi_highlight();
-                off = ansi_normal();
-        } else {
-                on = ansi_highlight_red();
-                off = ansi_normal();
         }
```
yuwata, line 410: You can drop this `}` now.

src/systemctl/systemctl-list-units.c
```diff
@@ -403,19 +402,18 @@ static int output_sockets_list(struct socket_info *socket_infos, unsigned cs) {
                                 return table_log_add_error(r);
```
yuwata, line 383: Please also drop this empty line.

src/systemctl/systemctl-list-units.c
yuwata: I noticed that the condition `if (cs)` can be dropped now.

dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Done.

src/systemctl/systemctl-list-units.c
dtardon: Okay, let's try that.

src/systemctl/systemctl-list-units.c
```diff
@@ -405,14 +415,8 @@ static int output_sockets_list(struct socket_info *socket_infos, unsigned cs) {
...
 
-        if (arg_legend != 0) {
-                printf("\n%s%u sockets listed.%s\n",
-                       cs > 0 ? ansi_highlight() : ansi_highlight_red(),
-                       cs,
-                       ansi_normal());
-                if (!arg_all)
-                        printf("Pass --all to see loaded but inactive sockets, too.\n");
-        }
+        if (arg_legend != 0)
```
medhefgo, line 416: Might as well move this into the function too…

dtardon: I wanted to keep it explicit that the output of legend is optional. But I don't care much either way...

src/systemctl/systemctl-list-units.c
```diff
@@ -641,20 +635,15 @@ static int output_timers_list(struct timer_info *timer_infos, unsigned n) {
...
         r = output_table(table);
         if (r < 0)
                 return r;
 
         if (arg_legend != 0) {
-                printf("\n%s%u timers listed.%s\n", on, n, off);
+                printf("\n%s%u timers listed.%s\n",
+                       n > 0 ? ansi_highlight() : ansi_highlight_red(),
+                       n,
+                       ansi_normal());
```
poettering: I am not sure these are really an improvement, no? Sure, it's a bit shorter, but I kinda liked the fact the coloring was separate from the actual string we output here. Now it's a more complex expression where it's harder to grok what is what...

src/systemctl/systemctl-list-units.c
```diff
@@ -369,38 +369,36 @@ static int output_sockets_list(struct socket_info *socket_infos, unsigned cs) {
 
         (void) table_set_empty_string(table, "-");
 
-        if (cs) {
-                for (struct socket_info *s = socket_infos; s < socket_infos + cs; s++) {
```
poettering, line 374: I have the suspicion some of the static checkers are not going to like this, given that socket_infos might be NULL here. i.e. the static checkers really don't like it if one does maths on NULL pointers, even if the value you add to them is 0 itself...

Hence, change the second part of the for expression from this:

```c
…; s < socket_infos + cs; …
```

to

```c
…; socket_infos && (s < socket_infos + cs); …
```

then the static checkers shouldn't mind

poettering:
looks mostly good, but the first commit i don#t find convincing, and the static checking thing should be fixed, before the static checker complains, even though it's not a real issue, just the checkers are too dumb.

src/systemctl/systemctl-list-units.c
dtardon: Well, my original patch just reduced scope of the two vars and shortened their definition. The inlining was @yuwata's suggestion. (And the extraction of the legend output to a separate function was an afterthought too.) I'm fine either way.

dtardon:
Btw, I've just realized it might make sense to merge this to #24033, because that could then use `output_legend()` too.

src/systemctl/systemctl-list-units.c
```diff
@@ -197,13 +197,11 @@ static int output_units_list(const UnitInfo *unit_infos, unsigned c) {
...
                                 puts("JOB    = Pending job for the unit.\n");
-                        on = ansi_highlight();
-                        off = ansi_normal();
-                } else {
-                        on = ansi_highlight_red();
-                        off = ansi_normal();
                 }
 
+                on = records > 0 ? ansi_highlight() : ansi_highlight_red();
+                off = ansi_normal();
```
poettering, line 203: I think the first commit should change the code like this, too. I think this is the most readable + relatively short way to spell this out.

poettering:
btw, while reviewing this, i noticed some issues that this PR fixes: #24105. Merging that will cause conflicts in this PR, and vice versa.

It might make sense to pull that PR into yours?

Either way, this PR should make similar fixes, i.e. in the second to last commit, the `n_items` parameter should also be switched from `unsigned` to `size_t`.

dtardon:
I've merged this to #24033 -> closing.

PR #24100, opened 2022-07-24
pid1: extend "ConditionFirmware=" for checking SMBIOS system identification information

d4nuu8:
add new condition type "ConditionDMI=" for checking SMBIOS/DMI system identification information.

### DIFF
--- a/man/systemd.unit.xml
+++ b/man/systemd.unit.xml
@@ -1235,10 +1235,24 @@
         <varlistentry>
           <term><varname>ConditionFirmware=</varname></term>
 
-          <listitem><para>Check whether the system's firmware is of a certain type. Possible values are:
-          <literal>uefi</literal> (for systems with EFI),
-          <literal>device-tree</literal> (for systems with a device tree) and
-          <literal>device-tree-compatible(xyz)</literal> (for systems with a device tree that is compatible to <literal>xyz</literal>).</para>
+          <listitem><para>Check whether the system's firmware is of a certain type. Multiple values are possible.</para>
+
+          <para><literal>uefi</literal> for systems with EFI.</para>
+
+          <para><literal>device-tree</literal> for systems with a device tree.</para>
+
+          <para><literal>device-tree-compatible(<replaceable>value</replaceable>)</literal> for systems with a device tree that is compatible to
+          <literal>value</literal>.</para>
+
+          <para><literal>smbios-field(<replaceable>field</replaceable> <replaceable>operator</replaceable> <replaceable>value</replaceable>)</literal>
+          for systems with a SMBIOS field containing a certain value.
+          <literal>field</literal> is the name of the SMBIOS field exposed as <literal>sysfs</literal> attribute file
+          below <filename>/sys/class/dmi/id/</filename>.
+          <literal>operator</literal> is one of <literal>&lt;</literal>, <literal>&lt;=</literal>,
+          <literal>&gt;=</literal>, <literal>&gt;</literal>, <literal>=</literal>, <literal>!=</literal> for version
+          comparison, or <literal>=$</literal>, <literal>!=$</literal> for string comparison.
+          <literal>value</literal> is the expected value of the SMBIOS field (shell-style globs are possible if
+          <literal>=$</literal> or<literal>!=$</literal> is used).</para>
           </listitem>
         </varlistentry>
 
--- a/src/shared/condition.c
+++ b/src/shared/condition.c
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
+#include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <fnmatch.h>
@@ -184,6 +185,10 @@ static int condition_test_credential(Condition *c, char **env) {
 typedef enum {
         /* Listed in order of checking. Note that some comparators are prefixes of others, hence the longest
          * should be listed first. */
+        _ORDER_FNMATCH_FIRST,
+        ORDER_FNMATCH_EQUAL = _ORDER_FNMATCH_FIRST,
+        ORDER_FNMATCH_UNEQUAL,
+        _ORDER_FNMATCH_LAST = ORDER_FNMATCH_UNEQUAL,
         ORDER_LOWER_OR_EQUAL,
         ORDER_GREATER_OR_EQUAL,
         ORDER_LOWER,
@@ -194,8 +199,10 @@ typedef enum {
         _ORDER_INVALID = -EINVAL,
 } OrderOperator;
 
-static OrderOperator parse_order(const char **s) {
+static OrderOperator parse_order(const char **s, bool allow_fnmatch) {
         static const char *const prefix[_ORDER_MAX] = {
+                [ORDER_FNMATCH_EQUAL] = "=$",
+                [ORDER_FNMATCH_UNEQUAL] = "!=$",
                 [ORDER_LOWER_OR_EQUAL] = "<=",
                 [ORDER_GREATER_OR_EQUAL] = ">=",
                 [ORDER_LOWER] = "<",
@@ -209,6 +216,8 @@ static OrderOperator parse_order(const char **s) {
 
                 e = startswith(*s, prefix[i]);
                 if (e) {
+                        if (!allow_fnmatch && (i >= _ORDER_FNMATCH_FIRST && i <= _ORDER_FNMATCH_LAST))
+                                break;
                         *s = e;
                         return i;
                 }
@@ -268,7 +277,7 @@ static int condition_test_kernel_version(Condition *c, char **env) {
                         break;
 
                 s = strstrip(word);
-                order = parse_order(&s);
+                order = parse_order(&s, /* allow_fnmatch= */ false);
                 if (order >= 0) {
                         s += strspn(s, WHITESPACE);
                         if (isempty(s)) {
@@ -329,7 +338,7 @@ static int condition_test_osrelease(Condition *c, char **env) {
                                         "Failed to parse parameter, key/value format expected: %m");
 
                 /* Do not allow whitespace after the separator, as that's not a valid os-release format */
-                order = parse_order(&word);
+                order = parse_order(&word, /* allow_fnmatch= */ false);
                 if (order < 0 || isempty(word) || strchr(WHITESPACE, *word) != NULL)
                         return log_debug_errno(SYNTHETIC_ERRNO(EINVAL),
                                         "Failed to parse parameter, key/value format expected: %m");
@@ -366,7 +375,7 @@ static int condition_test_memory(Condition *c, char **env) {
         m = physical_memory();
 
         p = c->parameter;
-        order = parse_order(&p);
+        order = parse_order(&p, /* allow_fnmatch= */ false);
         if (order < 0)
                 order = ORDER_GREATER_OR_EQUAL; /* default to >= check, if nothing is specified. */
 
@@ -392,7 +401,7 @@ static int condition_test_cpus(Condition *c, char **env) {
                 return log_debug_errno(n, "Failed to determine CPUs in affinity mask: %m");
 
         p = c->parameter;
-        order = parse_order(&p);
+        order = parse_order(&p, /* allow_fnmatch= */ false);
         if (order < 0)
                 order = ORDER_GREATER_OR_EQUAL; /* default to >= check, if nothing is specified. */
 
@@ -578,8 +587,62 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
         return strv_contains(dtcompatlist, dtcarg);
 }
 
+static int condition_test_firmware_smbios_field(const char *expression) {
+        _cleanup_free_ char *field = NULL, *expected_value = NULL, *actual_value = NULL;
+        OrderOperator operator;
+        int r;
+
+        assert(expression);
+
+        /* Parse SMBIOS field */
+        r = extract_first_word(&expression, &field, "!<=>$", EXTRACT_RETAIN_SEPARATORS);
+        if (r < 0)
+                return r;
+        if (r == 0 || isempty(expression))
+                return -EINVAL;
+
+        /* Remove trailing spaces from SMBIOS field */
+        delete_trailing_chars(field, WHITESPACE);
+
+        /* Parse operator */
+        operator = parse_order(&expression, /* allow_fnmatch= */ true);
+        if (operator < 0)
+                return operator;
+
+        /* Parse expected value */
+        r = extract_first_word(&expression, &expected_value, NULL, EXTRACT_UNQUOTE);
+        if (r < 0)
+                return r;
+        if (r == 0 || !isempty(expression))
+                return -EINVAL;
+
+        /* Read actual value from sysfs */
+        if (!filename_is_valid(field))
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Invalid SMBIOS field name");
+
+        const char *p = strjoina("/sys/class/dmi/id/", field);
+        r = read_virtual_file(p, SIZE_MAX, &actual_value, NULL);
+        if (r < 0) {
+                log_debug_errno(r, "Failed to read %s: %m", p);
+                if (r == -ENOENT)
+                        return false;
+                return r;
+        }
+
+        /* Remove trailing newline */
+        delete_trailing_chars(actual_value, WHITESPACE);
+
+        /* Finally compare actual and expected value */
+        if (operator == ORDER_FNMATCH_EQUAL)
+                return fnmatch(expected_value, actual_value, FNM_EXTMATCH) != FNM_NOMATCH;
+        if (operator == ORDER_FNMATCH_UNEQUAL)
+                return fnmatch(expected_value, actual_value, FNM_EXTMATCH) == FNM_NOMATCH;
+        return test_order(strverscmp_improved(actual_value, expected_value), operator);
+}
+
 static int condition_test_firmware(Condition *c, char **env) {
-        sd_char *dtc;
+        sd_char *arg;
+        int r;
 
         assert(c);
         assert(c->parameter);
@@ -592,24 +655,40 @@ static int condition_test_firmware(Condition *c, char **env) {
                         return false;
                 } else
                         return true;
-        } else if ((dtc = startswith(c->parameter, "device-tree-compatible("))) {
-                _cleanup_free_ char *dtcarg = NULL;
+        } else if ((arg = startswith(c->parameter, "device-tree-compatible("))) {
+                _cleanup_free_ char *dtc_arg = NULL;
                 char *end;
 
-                end = strchr(dtc, ')');
+                end = strchr(arg, ')');
                 if (!end || *(end + 1) != '\0') {
-                        log_debug("Malformed Firmware condition \"%s\"", c->parameter);
+                        log_debug("Malformed ConditionFirmware=%s", c->parameter);
                         return false;
                 }
 
-                dtcarg = strndup(dtc, end - dtc);
-                if (!dtcarg)
+                dtc_arg = strndup(arg, end - arg);
+                if (!dtc_arg)
                         return -ENOMEM;
 
-                return condition_test_firmware_devicetree_compatible(dtcarg);
+                return condition_test_firmware_devicetree_compatible(dtc_arg);
         } else if (streq(c->parameter, "uefi"))
                 return is_efi_boot();
-        else {
+        else if ((arg = startswith(c->parameter, "smbios-field("))) {
+                _cleanup_free_ char *smbios_arg = NULL;
+                char *end;
+
+                end = strchr(arg, ')');
+                if (!end || *(end + 1) != '\0')
+                        return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Malformed ConditionFirmware=%s: %m", c->parameter);
+
+                smbios_arg = strndup(arg, end - arg);
+                if (!smbios_arg)
+                        return log_oom_debug();
+
+                r = condition_test_firmware_smbios_field(smbios_arg);
+                if (r < 0)
+                        return log_debug_errno(r, "Malformed ConditionFirmware=%s: %m", c->parameter);
+                return r;
+        } else {
                 log_debug("Unsupported Firmware condition \"%s\"", c->parameter);
                 return false;
         }
--- a/src/test/test-condition.c
+++ b/src/test/test-condition.c
@@ -17,6 +17,7 @@
 #include "efi-loader.h"
 #include "env-util.h"
 #include "errno-util.h"
+#include "fileio.h"
 #include "fs-util.h"
 #include "hostname-util.h"
 #include "id128-util.h"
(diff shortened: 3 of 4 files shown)


poettering:
I would really prefer if we'd fold this into ConditionFirmware= somehow.

What's the usecase for this? Cna you elaborate? The thing is, we already allow to nicely match DMI from devices, and given that DMI checks are usually done for software that is associated with some device, it might make more sense to express checks as per-device. Hence, can you elaborate what your usecase is?

man/systemd.unit.xml
```diff
@@ -1361,6 +1361,20 @@
           </listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><varname>ConditionDMI=</varname></term>
```
poettering: So, ConditionFirmware= already knows `device-tree-compatible(…)`, which is a parameterized test. I think we should build this in a similar way.

(Also, I think "DMI" is actually a misnomer, the standard is actually called SMBIOS, and maybe it's time to use the right wording here for new stuff).

So what about this:

```
ConditionFirmware=smbios-field(xxxx == yyy)
```

src/shared/condition.c
```diff
@@ -353,6 +353,61 @@ static int condition_test_osrelease(Condition *c, char **env) {
         return true;
 }
 
+static int condition_test_dmi(Condition *c, char **env) {
+        int r;
+
+        assert(c);
+        assert(c->type == CONDITION_DMI);
+
+        for (const char *parameter = ASSERT_PTR(c->parameter);;) {
```
poettering: a for loop that isn't looping?

src/shared/condition.c
```diff
@@ -353,6 +353,61 @@ static int condition_test_osrelease(Condition *c, char **env) {
...
+                if (r < 0)
+                        return log_debug_errno(r, "Failed to read %s, ignoring: %m", p);
+
+                /* Might not be comparing versions, so do exact string matching */
+                if (order == ORDER_EQUAL)
+                        matches = streq_ptr(actual_value, word);
+                else if (order == ORDER_UNEQUAL)
+                        matches = !streq_ptr(actual_value, word);
+                else
+                        matches = test_order(strverscmp_improved(actual_value, word), order);
```
poettering: dmi fields are typically not version strings, hence, does this really make sense to use strverscmp_improved()?

src/shared/condition.c
```diff
@@ -353,6 +353,61 @@ static int condition_test_osrelease(Condition *c, char **env) {
...
+
+                if (!matches)
+                        return false;
```
poettering: So, does it really make sense to use ordering and equality operators on DMI fields? The stuff isn't usually something you would want to check order for. Instead, it's probably something to use matches like fnmatch() on, i.e. so that you can for example check for `*Lenovo*` or similar.

man/systemd.unit.xml
d4nuu8: > (Also, I think "DMI" is actually a misnomer, the standard is actually called SMBIOS, and maybe it's time to use the right wording here for new stuff).

I was struggling between DMI and SMBIOS, too. I've chosen DMI because that's was the kernel is using. But SMBIOS is also fine for me.

> So what about this:
> ```ConditionFirmware=smbios-field(xxxx == yyy)```

Sound reasonable for me. I'm going to implement this with the next update.

src/shared/condition.c
d4nuu8: I've chosen the `condition_test_osrelease` function as template and there it is done this way. Since this is my first contribution to systemd, I thought there would be something I'm missing right now...

This should be resolved with the integration into `condition_test_firmware`.

src/shared/condition.c
d4nuu8: I had the `xxx_version` fields in mind (e.g `bios_version`, `board_version`, `chassis_version`, `product_version`). Do you think this makes sense?

d4nuu8:
Changes since last version:

- Drop creation of new condition type `ConditionDMI=`
- Integrate functionality into already existing condition type `ConditionFirmware=`
- Use `fnmatch()`in case of `=`, or `!=` operator to allow shell-style globs

src/shared/condition.c
poettering: > I had the xxx_version fields in mind (e.g bios_version, board_version, chassis_version, product_version). Do you think this makes sense?

I have my doubts about that. They are unlikely to follow the versioning syntax we typically use in Linux packaging, hence I would really not go as far as assuming we could clearly define an ordering for them generically.

The DMI matching code we have so far (in udev) goes via fnmatch() too pretty much. Hence I'd really stick to that, and not do version compares here. I think it's quite a stretch to assume version strings read from DMI are sanely comparable...

poettering:
Do you actually have a real-life usecase for the version comapre stuff? otherwise, please drop, and just do fnmatch

d4nuu8:
Speaking for my company I can say, that we are indeed storing semantic versions in the `board_version` SMBIOS field. So for us this would be really useful.

src/shared/condition.c
```diff
@@ -578,8 +579,55 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
         return strv_contains(dtcompatlist, dtcarg);
 }
 
+static int condition_test_firmware_smbios_field(const char *expression) {
+        int r;
+        size_t len;
+        _cleanup_free_ char *field = NULL, *expected_value = NULL, *actual_value = NULL;
+        OrderOperator operator;
+
```
bluca, line 594: `assert(expression)`

src/test/test-condition.c
```diff
@@ -18,6 +18,7 @@
 #include "env-util.h"
 #include "errno-util.h"
 #include "fs-util.h"
+#include "fileio.h"
```
bluca, line 20: sort alphabetically

src/shared/condition.c
```diff
@@ -592,24 +640,43 @@ static int condition_test_firmware(Condition *c, char **env) {
...
+        } else if (streq(c->parameter, "uefi")) {
                 return is_efi_boot();
-        else {
+        } else if ((arg = startswith(c->parameter, "smbios-field("))) {
+                _cleanup_free_ char *smbios_arg = NULL;
+                char *end;
+
+                end = strchr(arg, ')');
+                if (!end || *(end + 1) != '\0') {
+                        log_debug(FIRMWARE_MALFORMED_MSG, c->parameter);
```
bluca: `return log_debug_errno(SYNTHETHIC_ERRNO(EINVAL), FIRMWARE_...`

src/shared/condition.c
```diff
@@ -592,24 +640,43 @@ static int condition_test_firmware(Condition *c, char **env) {
...
+
+                smbios_arg = strndup(arg, end - arg);
+                if (!smbios_arg)
+                        return -ENOMEM;
+
+                r = condition_test_firmware_smbios_field(smbios_arg);
+                if (r >= 0)
+                        return r;
+                log_debug(FIRMWARE_MALFORMED_MSG, c->parameter);
+                return -EINVAL;
```
bluca: same as above

src/shared/condition.c
d4nuu8: Changed

src/shared/condition.c
d4nuu8: Changed

src/test/test-condition.c
d4nuu8: Sorted

src/shared/condition.c
d4nuu8: Added

d4nuu8:
Changes since last version:

- Refactor logging like suggested in PR
- Sort includes alphabetically

d4nuu8:
> What's the usecase for this? Cna you elaborate? The thing is, we already allow to nicely match DMI from devices, and given that DMI checks are usually done for software that is associated with some device, it might make more sense to express checks as per-device. Hence, can you elaborate what your usecase is?

@poettering 
Sorry, seems like I missed this part of your comment.

We have some services which highly depend on the hardware they are running on. Because boot time is one critical requirement for us, we conditionalize those services to prevent every service from being started and check if it is really needed to be started at all.

Right now, we are achieving this by:

1. Reading the relevant SMBIOS fields within GRUB
2. Pass them via `systemd.setenv` to the kernel command line
3. Conditionalize the services by using `ConditionEnvironment=`

This has two major drawbacks for us:

1. We are heavily tied to GRUB and cannot switch to systemd-boot
2. This approach is not working for conditionalized services based on the board version we are running on

src/shared/condition.c
```diff
@@ -578,8 +579,57 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+
+        assert(expression);
+
+        /* Parse SMBIOS field */
+        r = extract_first_word(&expression, &field, "!<=>", EXTRACT_RETAIN_SEPARATORS);
+        if (r <= 0 || isempty(expression))
+                return -EINVAL;
+
+        /* Remove trailing spaces from SMBIOS field */
+        while ((len = strlen(field)) && isspace(field[len - 1]))
```
bluca: use `strstrip()` for this

src/shared/condition.c
```diff
@@ -578,8 +579,57 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+                field[len - 1] = '\0';
+
+        /* Parse operator */
+        operator = parse_order(&expression);
+        if (operator == -EINVAL)
+                return operator;
+
+        /* Remove leading spaces before parsing expected value */
+        while (isspace(*expression))
```
bluca: use `first_word()` for this

src/shared/condition.c
d4nuu8: Changed

src/shared/condition.c
d4nuu8: Replaced with `extract_first_word()`

d4nuu8:
Changes since last version:

- Use `ststrip` to remove trailing spaces from SMBIOS field name
- Use `extract_first_word()` to parse expected value
- Automatically add quotes in tests if any whitespace is contained in the expected value

poettering:
hmmm, ok, if you do control the hw then version compares are indeed ok. But in that case, I think we should probably be systematic it also do the regular equality/inequality matches with strverscmp_improved(). But I still think we should support fnmatch() based comparisons. hence maybe introduce a new operator for this, i.e. or `=$` or so?

That way, you'd have `<`, `<=`, `=`, `!=`, `>=`, `>` for version compares and `=$`and `!=$` for fnmatch compares?

or something like that? (other ideas?)

And I think we should probably then implement the same logic for the os release condition and share an implementation of the comparator code?

poettering:
and maybe == for simple string compares

src/shared/condition.c
```diff
@@ -578,8 +579,51 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+        int r;
+        _cleanup_free_ char *field = NULL, *expected_value = NULL, *actual_value = NULL;
+        OrderOperator operator;
+
+        assert(expression);
+
+        /* Parse SMBIOS field */
+        r = extract_first_word(&expression, &field, "!<=>", EXTRACT_RETAIN_SEPARATORS);
+        if (r <= 0 || isempty(expression))
+                return -EINVAL;
```
poettering, line 602: this really should propagate the original error if we have one

src/shared/condition.c
```diff
@@ -578,8 +579,51 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
 
+static int condition_test_firmware_smbios_field(const char *expression) {
+        int r;
+        _cleanup_free_ char *field = NULL, *expected_value = NULL, *actual_value = NULL;
+        OrderOperator operator;
+
+        assert(expression);
+
+        /* Parse SMBIOS field */
+        r = extract_first_word(&expression, &field, "!<=>", EXTRACT_RETAIN_SEPARATORS);
```
poettering: given the `"!<=>"` is now used at at least two places, please add a `#define` macro for it

src/shared/condition.c
```diff
@@ -578,8 +579,51 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+        if (r <= 0 || isempty(expression))
+                return -EINVAL;
+
+        /* Remove trailing spaces from SMBIOS field */
+        strstrip(field);
+
+        /* Parse operator */
+        operator = parse_order(&expression);
+        if (operator == -EINVAL)
```
poettering: why just check EINVAL? all parse errors should be propagated...

src/shared/condition.c
```diff
@@ -578,8 +579,51 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+                return operator;
+
+        /* Parse expected value */
+        r = extract_first_word(&expression, &expected_value, NULL, EXTRACT_UNQUOTE);
+        if (r <= 0)
+                return -EINVAL;
```
poettering, line 617: as above, propagate original error if you have one

src/shared/condition.c
```diff
@@ -578,8 +579,51 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+
+        /* Read actual value from sysfs */
+        const char *p = strjoina("/sys/class/dmi/id/", ascii_strlower(field));
```
poettering: Typically we operate in case-sensitive fashion, hence I don't think we should do ascii_strlower()

man/systemd.unit.xml
```diff
@@ -1235,10 +1235,24 @@
...
+
+          <para><literal>device-tree</literal> for systems with a device tree.</para>
+
+          <para><literal>device-tree-compatible(<replaceable>value</replaceable>)</literal> for systems with a device tree that is compatible to
+          <literal>value</literal>.</para>
+
+          <para><literal>smbios-field(<replaceable>field</replaceable> <replaceable>operator</replaceable> <replaceable>value</replaceable>)</literal>
+          for systems with a SMBIOS field containing a certain value.
+          <literal>field</literal> is the name of the SMBIOS field which is read from
+          <filename>/sys/class/dmi/id/<replaceable>field</replaceable></filename>.
```
poettering: unfortunately the sysfs attribute files are actually named differently from the fields in the SMBIOS spec. Hence maybe say `is the name of the SMBIOS field as exposed as <literal>sysfs</literal> attribute file below <filename>/sys/class/dmi/id/</filename>`

src/shared/condition.c
poettering: Also, please don't blindly append arbitrary strings to paths. Code hygiene demands you verify that the string is safe to append here. i.e. please check first if the string passes filename_is_valid(), and refuse things if not.

src/shared/condition.c
```diff
@@ -578,8 +579,51 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+        r = read_one_line_file(p, &actual_value);
+        if (r < 0) {
+                log_debug_errno(r, "Failed to read %s: %m", p);
+                return false;
```
poettering: please propagate unexpected errors as they are. That said, it might make sense to map ENOENT (which is kind of an expected error) more sensibly, i.e. maybe as `false`

dnicolodi:
> That way, you'd have `<`, `<=`, `=`, `!=`, `>=`, `>` for version compares and `=$`and `!=$` for fnmatch compares?

Perl uses `=~` and `!~` for positive and negative regular expression matches, thus maybe the same operator symbols could be used for fnmatch matches. They look a bit better than `=$` and `!=$`.

poettering:
> Perl uses `=~` and `!~` for positive and negative regular expression matches, thus maybe the same operator symbols could be used for fnmatch matches. They look a bit better than `=$` and `!=$`.

Yupp, in fact Perl came to my mind too, when I suggested `=$`. But precisely because Perl uses it for regex matches instead of fnmatches I figured we should not reuse `=~`...

d4nuu8:
> and maybe == for simple string compares

I used `ConditionMemory=`, `ConditionCPUs=` and `ConditionOSRelease=` as reference as they are using those operators. But for me this is fine.

> That way, you'd have <, <=, =, !=, >=, > for version compares and =$and !=$ for fnmatch compares?

Why do we need to use `strverscmp_improved()` for equality / inequality comparison of versions at all? Where is the difference to use `fnmatch()` in this case?

Shouldn't this result in the same result? 

And if we use `fnmatch()` for version comparison one could use something like `bios_version = 3.*`. Isn't this a valid use case?

Or am I missing something here?

poettering:
> Why do we need to use `strverscmp_improved()` for equality / inequality comparison of versions at all? Where is the difference to use `fnmatch()` in this case?

```
strverscmp_improved("4711*", "4711") → -1
fnmatch("4711*", "4711") → 0
```

poettering:
There are SMBIOS fields that are textual strings, and SMBIOS fields that can conceptually be considered version numbers. For the former fnmatch() makes sense, for the latter strverscmp_improved() matches. Since the two come to different equality/inequality results for the same pair of strings we should provide a way to pick explicitly which one the user wants.

dnicolodi:
> There are SMBIOS fields that are textual strings, and SMBIOS fields that can conceptually be considered version numbers.

Just brainstorming: what about then having different options for different type of SMBIOS fields? For example
```
ConditionFirmware=smbios-field-strver(foo >= 1.2.3)
ConditionFirmware=smbios-field(bar = baz*)
```
or something like this.

poettering:
hmm, not sure about that. the thing is that we probably should teach os-release condition checks and similar the same semantics, but there you can't just add a prefix, because it currently has none... thus i'd really like this to be folded into the operator

d4nuu8:
> hmm, not sure about that. the thing is that we probably should teach os-release condition checks and similar the same semantics, but there you can't just add a prefix, because it currently has none...

But right now `ConditionOSRelease=` is using `streq_ptr()` for `=` and `!=` operators, and `strverscmp_improved()` for the rest of the operators.

So if we are want to use `=` and `!=` for version comparison in `ConditionOSRelease=`  and here, this will be a breaking change anyways. Right?

Maybe we could use some kind of "function". For example
```
ConditionFirmware=smbios-field(version(foo = 1.2.3))
ConditionFirmware=smbios-field(bar = baz*)

ConditionOSRelease=PRETTY_NAME=Fedora*
ConditionOSRelease=version(VERSION_ID=17.1)
```

poettering:
please, let's not go overboard with this. nesting concepts is clearly too much.

So you have a point that `strverscmp_improved() == 0` is different from `streq()`. But I wonder if we can still save that, maybe by distuingishing the "=" operator from the "==" operator or so?

d4nuu8:
> But I wonder if we can still save that, maybe by distuingishing the "=" operator from the "==" operator or so?

Don't you think this makes it even more complicated if `=` and `==` are having a "slightly" different functionality?

I would prefer to use a operator which can be clearly distinguished.

d4nuu8:
Changes since last version:

- Update documentation
- Propagate unexpected errors as they are
- Validate SMBIOS field file name


Open issues:

- Clarify how to proceed with version / fnmatch comparison operators

src/shared/condition.c
d4nuu8: I think this can be closed

src/shared/condition.c
d4nuu8: I think this can be closed

src/shared/condition.c
d4nuu8: Adjusted in latest version

src/shared/condition.c
d4nuu8: Adjusted in latest version

src/shared/condition.c
d4nuu8: Adjusted in latest version

src/shared/condition.c
d4nuu8: Adjusted in latest version

man/systemd.unit.xml
d4nuu8: Adjusted in latest version

src/shared/condition.c
d4nuu8: Adjusted in latest version

poettering:
needs rebase

src/shared/condition.c
```diff
@@ -578,8 +579,60 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+                return r;
+        if (r == 0)
+                return -EINVAL;
+
+        /* Read actual value from sysfs */
+        if (!filename_is_valid(field))
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Invalid SMBIOS field name");
+
+        const char *p = strjoina("/sys/class/dmi/id/", field);
+        r = read_one_line_file(p, &actual_value);
```
poettering: maybe use read_virtual_file()

src/shared/condition.c
```diff
@@ -578,8 +579,60 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+        /* Remove trailing spaces from SMBIOS field */
+        strstrip(field);
+
+        /* Parse operator */
+        operator = parse_order(&expression);
+        if (operator < 0)
+                return operator;
+
+        /* Parse expected value */
+        r = extract_first_word(&expression, &expected_value, NULL, EXTRACT_UNQUOTE);
```
poettering, line 613: after this call you should check if `expression` is empty. If it isn't there's more garbage on the line, and we should refuse that.

src/test/test-condition.c
```diff
@@ -305,6 +306,107 @@ TEST(condition_test_architecture) {
...
+        condition_free(condition);
+
+        /* Test not existing SMBIOS field */
+        condition = condition_new(CONDITION_FIRMWARE, "smbios-field(not_existing=nothing)", false, false);
+        assert_se(condition);
+        assert_se(condition_test(condition, environ) == 0);
+        condition_free(condition);
+
+        /* Test with bios_vendor, if available */
+        if (read_one_line_file("/sys/class/dmi/id/bios_vendor", &bios_vendor) <= 0)
```
poettering: probably better to use read_virtual_file

poettering (changes requested):

poettering:
I still think we should do fnmatching only if `=$` or so is used, and otherwise do simple string compares on `=`

src/shared/condition.c
```diff
@@ -578,8 +579,60 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+        if (r < 0) {
+                log_debug_errno(r, "Failed to read %s: %m", p);
+                if (r == -ENOENT)
+                        return false;
+                return r;
+        }
+
+        /* Finally compare actual and expected value */
+        if (operator == ORDER_EQUAL)
+                return fnmatch(expected_value, actual_value, 0) != FNM_NOMATCH;
```
d4nuu8: @poettering Could we use `FNM_EXTMATCH` as flag here (GNU extension)?

d4nuu8:
Changes since last version:

- Rebase
- Introduce `=$` and `!=$` as new operators for `fnmatch()` comparison
- Use `FNM_EXTMATCH` as `fnmatch()` flag
- Replace `read_one_line_file()` with `read_virtual_file()`
- Validate if expression contains no garbage after parsing

src/test/test-condition.c
d4nuu8: Agree, replaced

src/shared/condition.c
d4nuu8: Added in latest version

src/shared/condition.c
d4nuu8: Agree, replaced in latest version

src/shared/condition.c
d4nuu8: This is not the case anymore until `condition_test_osrelease()` is extended with the new `fnmatch()` operator.

d4nuu8:
Changes since last version:

- Fix debug logging of `device-tree-compatible` path

d4nuu8:
May I ask for another round of review?

man/systemd.unit.xml
```diff
@@ -1235,13 +1235,29 @@
...
+          <literal>operator</literal> is one of <literal>&lt;</literal>, <literal>&lt;=</literal>,
+          <literal>&gt;=</literal>, <literal>&gt;</literal>, <literal>=</literal>, <literal>!=</literal> for version
+          comparison, or <literal>=$</literal>, <literal>!=$</literal> for string comparison.
+          <literal>value</literal> is the expected value of the SMBIOS field (shell-style globs are possible if
+          <literal>=$</literal> or<literal>!=$</literal> is used).</para>
           </listitem>
         </varlistentry>
 
+        is the name of the SMBIOS field as
+
```
keszybz: Leftover blob?

src/shared/condition.c
```diff
@@ -196,6 +199,8 @@ typedef enum {
 
 static OrderOperator parse_order(const char **s) {
         static const char *const prefix[_ORDER_MAX] = {
+                [ORDER_FNMATCH_EQUAL] = "=$",
+                [ORDER_FNMATCH_UNEQUAL] = "!=$",
```
keszybz: This adds the two new operators to every settting that users `parse_order`: ConditionKernelVersion, ConditionOSRelease, ConditionMemory, ConditionCPUs. I don't think this is appropriate.

Please add a flag here that specifies whether the two new operators are allowed, and only pass `true` in some callers.

I think ConditionOSRelease and ConditionKernelVersion *could* be extended to allow this too. But it should be a separate commit, and the documentation must be updated to match.

src/shared/condition.c
```diff
@@ -578,8 +583,63 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
         return strv_contains(dtcompatlist, dtcarg);
 }
 
+static int condition_test_firmware_smbios_field(const char *expression) {
+        int r;
```
keszybz: Please put `int r` at the end of the declaration list, see CODING_STYLE.

src/shared/condition.c
```diff
@@ -578,8 +583,63 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+
+        /* Parse SMBIOS field */
+        r = extract_first_word(&expression, &field, "!<=>$", EXTRACT_RETAIN_SEPARATORS);
+        if (r < 0)
+                return r;
+        if (r == 0 || isempty(expression))
+                return -EINVAL;
+
+        /* Remove trailing spaces from SMBIOS field */
+        strstrip(field);
```
keszybz: This potentially changes the pointer, which means that it cannot be used in `_cleanup_`. You need to either define a second variable to keep the original pointer unchanged, or use a function that only removes trailing whitespace, e.g. `delete_trailing_chars()`.

src/shared/condition.c
```diff
@@ -592,24 +652,40 @@ static int condition_test_firmware(Condition *c, char **env) {
...
-                _cleanup_free_ char *dtcarg = NULL;
+        } else if ((arg = startswith(c->parameter, "device-tree-compatible("))) {
+                _cleanup_free_ char *dtc_arg = NULL;
                 char *end;
 
-                end = strchr(dtc, ')');
+                end = strchr(arg, ')');
                 if (!end || *(end + 1) != '\0') {
-                        log_debug("Malformed Firmware condition \"%s\"", c->parameter);
+                        log_debug_errno(SYNTHETIC_ERRNO(EINVAL), FIRMWARE_MALFORMED_MSG, c->parameter);
```
keszybz: SYNTHETIC_ERRNO() is only for cases where `return log_…` is used.

src/shared/condition.c
```diff
@@ -578,8 +583,63 @@ static int condition_test_firmware_devicetree_compatible(const char *dtcarg) {
...
+
+        /* Finally compare actual and expected value */
+        if (operator == ORDER_FNMATCH_EQUAL)
+                return fnmatch(expected_value, actual_value, FNM_EXTMATCH) != FNM_NOMATCH;
+        if (operator == ORDER_FNMATCH_UNEQUAL)
+                return fnmatch(expected_value, actual_value, FNM_EXTMATCH) == FNM_NOMATCH;
+        return test_order(strverscmp_improved(actual_value, expected_value), operator);
+}
+
+#define FIRMWARE_MALFORMED_MSG "Malformed Firmware condition \"%s\": %m"
```
keszybz: No, please just copy the message in both places. `#define`s like this just make it hard to edit.

Also, "Malformed ConditionFirmware= settting: …".

src/shared/condition.c
```diff
@@ -592,24 +652,40 @@ static int condition_test_firmware(Condition *c, char **env) {
...
+                if (!dtc_arg)
                         return -ENOMEM;
 
-                return condition_test_firmware_devicetree_compatible(dtcarg);
-        } else if (streq(c->parameter, "uefi"))
+                return condition_test_firmware_devicetree_compatible(dtc_arg);
+        } else if (streq(c->parameter, "uefi")) {
                 return is_efi_boot();
-        else {
+        } else if ((arg = startswith(c->parameter, "smbios-field("))) {
```
keszybz: No {} around single-line blocks, see CODING_STYLE.

keszybz (changes requested):

src/shared/condition.c
d4nuu8: I'm going to change this.

> I think ConditionOSRelease and ConditionKernelVersion could be extended to allow this too. But it should be a separate commit, and the documentation must be updated to match.

Is it OK for you to do this in a following pull request?

man/systemd.unit.xml
d4nuu8: Removed

src/shared/condition.c
d4nuu8: Done

src/shared/condition.c
d4nuu8: Replaced with `delete_trailing_chars()`

src/shared/condition.c
d4nuu8: Adjusted

src/shared/condition.c
d4nuu8: Removed the define and adjusted the message

src/shared/condition.c
d4nuu8: Removed

d4nuu8:
Changes since last version:

- Removed left over in man page
- Changed declaration list order of `condition_test_firmware_smbios_field()`
- Replaced `strstrip()` with `delete_trailing_chars()`
- Removed `SYNTHETIC_ERRNO()` when not returning the error
- Removed error message define
- Adjusted error message
- Removed `{}` around single line block
- Extend `parse_order()` parameters to reject `fnmatch()` operators
- Extend `ConditionOSRelease=` test with `fnmatch()` operators

I'm not sure if my changes are causing the broken CI. Can anybody help me to check it?

src/shared/condition.c
```diff
@@ -184,6 +185,8 @@ static int condition_test_credential(Condition *c, char **env) {
 typedef enum {
         /* Listed in order of checking. Note that some comparators are prefixes of others, hence the longest
          * should be listed first. */
+        ORDER_FNMATCH_EQUAL,
```
poettering: maybe do it like this:

```c
_ORDER_FNMATCH_FIRST,
ORDER_FNMATCH_EQUAL = _ORDER_FNMATCH_FIRST,
ORDER_FNMATCH_UNEQUAL,
_ORDER_FNMATCH_LAST = ORDER_FNMATCH_UNEQUAL,
```

This then allows checking for the fnmatch operators generically like this:

```c
if (i >= _ORDER_FNMATCH_FIRST && i <= _ORDER_FNMATCH_LAST) …
```

Thus you don't have to hardcode which is the first or last, but have a generic way to reference it, and adding more is obvious.

d4nuu8:
Changes since last version:

- Add `_ORDER_FNMATCH_FIRST` and `_ORDER_FNMATCH_LAST`
- Add comments to `parse_order()` calls

src/shared/condition.c
d4nuu8: Very good idea! I've implemented this in the latest version.

src/shared/condition.c
```diff
@@ -592,24 +655,40 @@ static int condition_test_firmware(Condition *c, char **env) {
...
-                _cleanup_free_ char *dtcarg = NULL;
+        } else if ((arg = startswith(c->parameter, "device-tree-compatible("))) {
+                _cleanup_free_ char *dtc_arg = NULL;
                 char *end;
 
-                end = strchr(dtc, ')');
+                end = strchr(arg, ')');
                 if (!end || *(end + 1) != '\0') {
-                        log_debug("Malformed Firmware condition \"%s\"", c->parameter);
+                        log_debug_errno(-EINVAL, "Malformed ConditionFirmware=%s: %m", c->parameter);
```
poettering: log_xyz_errno() has two purposes:

1. you can pass an error in you received, that you then can print with %m
2. you can pass an error in you received or made up, that is then returned so that you can do: `return log_xyz_errno(…)`.

Neither is the case here though: you are making up an error, but not even returning it. So you are printing a made-up error via %m here, but that's pointless...

Just use plain `log_debug()` instead of `log_debug_errno()`, and do not output %m in the message

d4nuu8:
Changes since last version:

- Adjust logging

src/shared/condition.c
d4nuu8: Changed

src/shared/condition.c
```diff
@@ -592,24 +655,40 @@ static int condition_test_firmware(Condition *c, char **env) {
...
+                if (!end || *(end + 1) != '\0')
+                        return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Malformed ConditionFirmware=%s: %m", c->parameter);
+
+                smbios_arg = strndup(arg, end - arg);
+                if (!smbios_arg)
+                        return -ENOMEM;
+
+                r = condition_test_firmware_smbios_field(smbios_arg);
+                if (r >= 0)
+                        return r;
```
poettering: usually we want that the exceptional code path is indented, and that the main "clean" code path is not.

Thus, please change this around to read:

```c
r = condition_test_firmware_smbios_field(…);
if (r < 0)
        return log_debug_errno(r, "…");

return r;
```

src/shared/condition.c
```diff
@@ -592,24 +655,40 @@ static int condition_test_firmware(Condition *c, char **env) {
...
+                _cleanup_free_ char *smbios_arg = NULL;
+                char *end;
+
+                end = strchr(arg, ')');
+                if (!end || *(end + 1) != '\0')
+                        return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Malformed ConditionFirmware=%s: %m", c->parameter);
+
+                smbios_arg = strndup(arg, end - arg);
+                if (!smbios_arg)
+                        return -ENOMEM;
```
poettering:
```c
return log_oom_debug();
```

maybe

poettering (changes requested):
looks good to me. just some minor nitpicks

src/shared/condition.c
d4nuu8: Changed in latest version.

There a some more functions in `condition.c` which are using `return -ENOMEM;`.

Shall a add another commit replacing them, too?

src/shared/condition.c
d4nuu8: Changed

d4nuu8:
Changes since last version:

- Use `log_oom_debug()`
- Turn around `if` statement

bluca (approved):

bluca (approved):

d4nuu8:
Thank you for your help and time invested!

PR #24101, opened 2022-07-24
random-util: assume getrandom() family exists

zx2c4:
```
All stable kernels available on kernel.org now have getrandom(0) and
getrandom(GRND_INSECURE). So adjust random-util to assume it exists, and
freak out of it doesn't.

This allows for pretty vast simplifications throughout, but is also a
bit radical of a move. I don't expect this to actually be merged, at
least not right away, but I thought I'd still place it here for
reference for some point in the future, or just for general interest.
```

@poettering @bluca As the commit message mentions, I don't really know if this is appropriate to merge. I still think it's an interesting exercise, and a thing to keep filed away in the back of the mind as being theoretically possible. What's new that makes it so? The ≥4.9 backports of most of the neat recent kernel random.c work.

### DIFF
--- a/src/basic/random-util.c
+++ b/src/basic/random-util.c
@@ -30,142 +30,39 @@
 #include "missing_syscall.h"
 #include "parse-util.h"
 #include "random-util.h"
-#include "sha256.h"
-#include "time-util.h"
-
-/* This is a "best effort" kind of thing, but has no real security value.
- * So, this should only be used by random_bytes(), which is not meant for
- * crypto. This could be made better, but we're *not* trying to roll a
- * userspace prng here, or even have forward secrecy, but rather just do
- * the shortest thing that is at least better than libc rand(). */
-static void fallback_random_bytes(void *p, size_t n) {
-        static thread_local uint64_t fallback_counter = 0;
-        struct {
-                char label[32];
-                uint64_t call_id, block_id;
-                usec_t stamp_mono, stamp_real;
-                pid_t pid, tid;
-                uint8_t auxval[16];
-        } state = {
-                /* Arbitrary domain separation to prevent other usage of AT_RANDOM from clashing. */
-                .label = "systemd fallback random bytes v1",
-                .call_id = fallback_counter++,
-                .stamp_mono = now(CLOCK_MONOTONIC),
-                .stamp_real = now(CLOCK_REALTIME),
-                .pid = getpid(),
-                .tid = gettid()
-        };
-
-#if HAVE_SYS_AUXV_H
-        memcpy(state.auxval, ULONG_TO_PTR(getauxval(AT_RANDOM)), sizeof(state.auxval));
-#endif
-
-        while (n > 0) {
-                struct sha256_ctx ctx;
-
-                sha256_init_ctx(&ctx);
-                sha256_process_bytes(&state, sizeof(state), &ctx);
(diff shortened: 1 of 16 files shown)


medhefgo:
I think the argument was overly zealous seccomp containers blocking getrandom.

poettering:
So our kernel baseline for systemd is 3.15 right now. (See README). We intend to raise that to 4.3 soon though. Would that suffice?

poettering:
And yeah, overzealous seccomp is a problem here. It's kinda nice having a fallback for that, given this is about core functionality, and not leaf functionality... But if the situation is clear enough for that, I'd be fine dropping the old stuff.

poettering:
getrandom() was added in kernel 3.17 says the man page. And GRND_INSECURE is 5.6 Google tells me. So even if we bump the baseline to 4.3 this is not even closely enough to be able to rely on it.

zx2c4:
> getrandom() was added in kernel 3.17 says the man page. And GRND_INSECURE is 5.6 Google tells me. So even if we bump the baseline to 4.3 this is not even closely enough to be able to rely on it.

The "new" thing here is that GRND_INSECURE is now in 4.9, if you're using kernel.org stable kernels.

Anyway, I'm really not suggesting in earnest that we merge this now; I just wanted to put it on your radar for some time in the indeterminate future (if ever). By the time everybody is running 4.9+, I'd also expect for the current crop of kernel.org stable kernels that have those changes to be the dominant 4.9+.

poettering:
> The "new" thing here is that GRND_INSECURE is now in 4.9, if you're using kernel.org stable kernels.

Hmm, you mean it has been backported to some 4.9.x kernel?

zx2c4:
> > The "new" thing here is that GRND_INSECURE is now in 4.9, if you're using kernel.org stable kernels.
> 
> Hmm, you mean it has been backported to some 4.9.x kernel?

Right. If you go to kernel.org, every kernel that you can download on that webpage has GRND_INSECURE.

poettering:
> Right. If you go to kernel.org, every kernel that you can download on that webpage has GRND_INSECURE.

That's not enough for us. We have to care about what distros ship, and I am not sure they feel bound by kernel.org's frontpage list so much...

zx2c4:
> > Right. If you go to kernel.org, every kernel that you can download on that webpage has GRND_INSECURE.
> 
> That's not enough for us. We have to care about what distros ship, and I am not sure they feel bound by kernel.org's frontpage list so much...

Yea, I understand. It's just that distros tend to eventually pull in the kernel.org stable kernel patches, and the hope is that by the time 4.9 is a reasonable minimum kernel for systemd (or perhaps some time after), those patches will have percolated onward to various distros.

Again, I'm not seriously suggesting that this be merged now. But I think there's a decent enough chance it'll become possible at some point down the road, without having to wait for 5.6 as a minimum.

(BTW, if you'd like to close this PR since it's not intended for immediate merging, that's fine with me.)

poettering:
So, I added this to the new "deprecations and removals" section in our TODO list:

https://github.com/systemd/systemd/blob/main/TODO#L80

That way we should not forget about that, once we bump the kernel baseline far enough. We typically check the TODO list for stuff to do pretty regularly (at least I do).

poettering:
So, let's close this for now. Hope that's OK.

PR #24104, opened 2022-07-25
boot: a couple of tweaks recommended by Coccinelle

mrc0mmand:


medhefgo:
This cannot (shouldn't?) work. strempty operates on char16_t in sd-boot, but these are char-strings. You could define a strempty8/16 in efi-string.h for this…

medhefgo:
nvm, I wasn't looking at the source file names :D

PR #24109, opened 2022-07-25
sd-netlink: rename NLType and friends

yuwata:
In sd-netlink, we use 'type' for naming many variables with different type,
and worst of all, function like `type_get_type()` exists. That makes the
code hard to understood.

Let's rename
- NLType -> NLAPolicy
- NLTypeSystem -> NLAPolicySet
- NLTypeSystemUnion -> NLAPolicySetUnion
- type_get_type() -> policy_get_type()
- type_get_type_system() -> policy_get_policy_set(), and so on.

This also changes the type of attribute taken by e.g.
`sd_netlink_message_append_string()` from `unsigned short` -> `uint16_t`
to make the variable size clearer.

### DIFF
--- a/src/libsystemd/sd-netlink/netlink-genl.c
+++ b/src/libsystemd/sd-netlink/netlink-genl.c
@@ -12,7 +12,7 @@
 typedef struct GenericNetlinkFamily {
         sd_netlink *genl;
 
-        const NLTypeSystem *type_system;
+        const NLAPolicySet *policy_set;
 
         uint16_t id; /* a.k.a nlmsg_type */
         char *name;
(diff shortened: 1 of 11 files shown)


poettering:
I am curious, how does the kernel name this internally?

DaanDeMeyer:
Maybe "Attribute" or the shorter "Attr" is an alternative to "Policy"? The netlink man pages make reference to attributes as well so it'd be nice if we could reuse it.

yuwata:
The Linux kernel internally uses `struct nla_policy`: from kernel's include/net/netlink.h
```c
struct nla_policy {
        u8              type;  //  <-- NLA_U32 or so, which is corresponding to NETLINK_TYPE_U32 in our code.
        u8              validation_type;
        u16             len;
        ...
};
```
Hence, I renamed `NLType` -> `NLAPolicy`.

Also, from kernel's include/uapi/linux/netlink.h
```c
struct nlattr {
        __u16           nla_len;
        __u16           nla_type; // <-- IFLA_IFNAME or so.
};
```
So, I renamed the function arguments which takes attribute types (e.g. IFLA_IFNAME) as `attr_type`.

poettering:
lgtm then (PR marked as draft though)

yuwata:
I initially thought setting this ready for review after adding more commits. But this is already reviewed and approved. Let's merge this as is, and I will open another PR for remaining commits.

PR #12148, opened 2019-03-29
unit: drop initrd-udevadm-cleanup-db.service

yuwata:
Closes #11997.

PR #24113, opened 2022-07-25
units: drop RemainAfterExit=yes from systemd-udev-trigger.service

vivien:
The main system often has udev rules that the initramfs does not. In
order to apply them, udevadm trigger must be run again from the
main system.

With RemainAfterExit=yes, systemd-udev-trigger.service won't be started
again after transitioning from the initramfs, unless the problematic
initrd-udevadm-cleanup-db.service is present, which conflicts with
and stops the former service.

RemainAfterExit=yes was introduced to avoid running the
service again when it is pulled in as a dependency, but only
initrd-udevadm-cleanup-db.service and systemd-udev-settle.service
list the service.

Removing RemainAfterExit=yes fixes main systems lacking
initrd-udevadm-cleanup-db.service and systemd-udev-settle.service,
and is a step forward getting rid of these services.

Cc: @yuwata (#12148)

### DIFF
--- a/units/systemd-udev-trigger.service
+++ b/units/systemd-udev-trigger.service
@@ -18,5 +18,4 @@ ConditionPathIsReadWrite=/sys
 
 [Service]
 Type=oneshot
-RemainAfterExit=yes
 ExecStart=-udevadm trigger --type=all --action=add --prioritized-subsystem=module,block,tpmrm,tty,net


yuwata:
Hm? What initramfs generator do you use? The service should not be started in initramfs.

vivien:
> Hm? What initramfs generator do you use? The service should not be started in initramfs.
A simple OpenEmbedded image containing systemd plus the /etc/initrd-release file and /sbin/init symlink.

What do you mean by the service should not be started in initramfs? It is symlinked in /usr/lib/systemd/system/sysinit.target.wants/ and it only contains ConditionPathIsReadWrite=/sys as a condition. Do you mean that you expect something like ConditionPathExists=!/etc/initrd-release?

yuwata:
> Do you mean that you expect something like ConditionPathExists=!/etc/initrd-release instead?

At least I thought so at the time I commented in the above. But not sure. I will reconsider that.

yuwata:
No, my bad. Please forget the previous comment. The service should run both in initramfs and after switching root.

I guess your initramfs does not start initrd-udevadm-cleanup-db.service, which has `Conflicts=` settings for udev related units. See https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/units/initrd-udevadm-cleanup-db.service.

vivien:
Indeed I dropped initrd-udevadm-cleanup-db.service and systemd-udev-settle.service, sorry for the lack of clarity. If the initramfs cleans up the database, my luks+btrfs root partition doesn't show up as ready (SYSTEMD_READY=0) and subvolumes described in fstab get umounted during boot. Without initrd-udevadm-cleanup-db.service, my luks+btrfs root partition is seen as ready (SYSTEMD_READY=1) in the main system and mountpoints work as expected. I did not tackle this issue yet, but I figured that cleaning the database wasn't necessary in this case anyway.

That said, regardless the presence of initrd-udevadm-cleanup-db.service, I do not think that relying on Conflicts=systemd-udev-trigger.service is a safe way to unsure that the service is started again in the main system, rather than having the condition described in the service itself. And because there is no upstream unit that pull it in, I figured that removing RemainAfterExit=yes was the way to go, hence my patch.

Instead of this, I guess that I could edit initrd-udevadm-cleanup-db.service on my end to clear ExectStart=, but that would be ugly.

yuwata:
That sounds the udev database for the related disk or partition needs to have persistent flag. See udev(7):
https://www.freedesktop.org/software/systemd/man/udev.html#db_persist

The udev database files with the persistent flag are not removed by initrd-udevadm-cleanup-db.service.

yuwata:
Or the main system does not have necessary .rules files for the setup.

vivien:
> That sounds the udev database for the related disk or partition needs to have persistent flag.

But shouldn't that work seamlessly when transitioning from initramfs to main system already, especially for a device that systemd is already aware of in the initramfs like a provided root=/dev/mapper/root from the kernel command line? I don't understand why systemd would not persist that in the first place.

> Or the main system does not have necessary .rules files for the setup.

Except from removing initrd-udevadm-cleanup-db.service and systemd-udev-settle.service for investigating, I did not remove anything else. I reverted the removal and added a 83-foo.rules like this on a fresh install:

```
ACTION!="remove", SUBSYSTEM=="block", ENV{DM_UUID}=="CRYPT-*", ENV{DM_NAME}=="root", OPTIONS="db_persist", SYMLINK+="root"
```

But that did not work, I don't have a /dev/root symlink and the partition is still seen not seen as ready. I am not sure how one must set this db_persist option given the number of rules tweaking DM_UDEV_DISABLE_*_RULES_FLAGS and SYSTEMD_READY, but again this seems odd to have to do this for the root partition described on the kernel command line.

Thanks for the input on this SYSTEMD_READY issue. But regarding systemd-udev-trigger.service, I still do not think that it is correct to depend on a conflict with initrd-udevadm-cleanup-db.service to ensure that it is started again in the main system, especially if we wish to get rid of initrd-udevadm-cleanup-db.service and systemd-udev-settle.service eventually if I'm not mistaken.

yuwata:
> > That sounds the udev database for the related disk or partition needs to have persistent flag.
> 
> But shouldn't that work seamlessly when transitioning from initramfs to main system already, especially for a device that systemd is already aware of in the initramfs like a provided root=/dev/mapper/root from the kernel command line? I don't understand why systemd would not persist that in the first place.
>
> > Or the main system does not have necessary .rules files for the setup.
> 
> Except from removing initrd-udevadm-cleanup-db.service and systemd-udev-settle.service for investigating, I did not remove anything else. I reverted the removal and added a 83-foo.rules like this on a fresh install:
> 
> ```
> ACTION!="remove", SUBSYSTEM=="block", ENV{DM_UUID}=="CRYPT-*", ENV{DM_NAME}=="root", OPTIONS="db_persist", SYMLINK+="root"
> ```
> 
> But that did not work, I don't have a /dev/root symlink and the partition is still seen not seen as ready. I am not sure how one must set this db_persist option given the number of rules tweaking DM_UDEV_DISABLE_*_RULES_FLAGS and SYSTEMD_READY, but again this seems odd to have to do this for the root partition described on the kernel command line.
> 
> Thanks for the input on this SYSTEMD_READY issue. But regarding systemd-udev-trigger.service, I still do not think that it is correct to depend on a conflict with initrd-udevadm-cleanup-db.service to ensure that it is started again in the main system, especially if we wish to get rid of initrd-udevadm-cleanup-db.service and systemd-udev-settle.service eventually if I'm not mistaken.

In general, udev rules files in initrd and host system are different. So, we cannot rely udev database generated in initrd stage.
I understand that's inefficient, but we do not have any better way to do more efficient way.

Anyway, unfortunately, the PR in current form is not acceptable, sorry. (Actually, we suggested similar PR few years ago, but that's rejected.)

yuwata:
Sorry, but I cannot say anything for sure.

If your system luckily works well without initrd-udevadm-cleanup-db.service, then it maybe OK to disable the service, but not sure.
Please test the setting or configuration on your system with your responsibility.

Several suggestions:
- About triggering device, we recently introduced several options for `udevadm trigger`. Especially, `--initialized-nomatch` hopefully useful for embedded systems. Also, `--prioritized-subsystem=` may speed up the boot process. These were requested by and implemented for those who using embedded systems.

- About systemd-udev-settle.service. The service is mostly deprecated. It should not cause any issue, but is inefficient and slow down boot process in general. Please try to use `Want=/Requires=` and `After=` in services that require specific devices.
E.g. wpa_supplicant.service in my system needs to be started after the wlan device is ready. Hence, the following lines are added to the service unit provided by default.
```
[Unit]
Requires=sys-subsystem-net-devices-wlp59s0.device
After=sys-subsystem-net-devices-wlp59s0.device
```
- Or, `udevadm info --wait-for-initialization` may be also useful for waiting for specific unit to be processed by udevd.

If you have any ideas about improving boot performance. Please open a new RFE page. Thank you.

As I said in the previous comment, this is not acceptable, unfortunately. Closing.

PR #24121, opened 2022-07-26
mkosi: Build against Fedora rawhide instead of latest stable release

daandemeyer:
Might be more noisy but will also provide more signal compared to
building against the latest stable Fedora release.

### DIFF
--- a/.github/workflows/mkosi.yml
+++ b/.github/workflows/mkosi.yml
@@ -40,6 +40,8 @@ jobs:
             release: jammy
           - distro: fedora
             release: "36"
+          - distro: fedora
+            release: rawhide
           - distro: opensuse
             release: tumbleweed
           - distro: centos_epel


yuwata:
Packit tests for rawhide. Hence, I think mkosi should be for a released version.

yuwata:
Or, if resource allow that, how about to keep the latest release, and add fedora rawhide?

DaanDeMeyer:
What's the point of testing against a released version though? It would make sense for systemd-stable since backported PRs will actually end up in the current stable Fedora release, but most code we merge won't be backported and as such won't ever end up in the current stable Fedora release so why test against it?

bluca:
> What's the point of testing against a released version though? It would make sense for systemd-stable since backported PRs will actually end up in the current stable Fedora release, but most code we merge won't be backported and as such won't ever end up in the current stable Fedora release so why test against it?

Simply additional coverage and quality in general. If we break something on a stable Fedora release we should fix it, even if official releases likely won't make it there.

DaanDeMeyer:
Sounds good, We can do both

bluca (approved):

yuwata (approved):
LGTM.

Note, from README:
```
POLICY FOR SUPPORT OF DISTRIBUTIONS AND ARCHITECTURES:

        systemd main branch and latest major or stable releases are generally
        expected to compile on current versions of popular distributions (at
        least all non-EOL versions of Fedora,
```
Hence, I think it is better that we have at least one simple CI for non-EOL version of Fedora.

PR #24122, opened 2022-07-26
core/mount: ignore -EACCES from mkdir_p_label() on NFS

### LINKED ISSUE #24120: bind mounts for NFS no longer working

250.8

### Used distribution

Fedora 36

### Linux kernel version used

5.18.13-200.fc36.x86_64

x86_64

### Component

systemd

Successful bind mount for /home directory on NFS.

fstab entries:
fileserver.domain:/           /mnt/nfs4/fileserver         nfs4    defaults,vers=4.0        0 0
/mnt/nfs4/fileserver/home    /home                   none    bind,_netdev    0 0

systemd[1]: home.mount: Failed to make bind mount source '/mnt/nfs4/fileserver': Permission denied

The problem relates to this code:
https://github.com/systemd/systemd/blob/5f4ccb0705a54f18f2010b9481039918c448a524/src/core/mount.c#L1057

systemd tries to mkdir() the source directory of a bind mount.

If the kernel cache for a NFS bind mount source path is cold, the kernel will send out the actual mkdir() request to the NFS server. This will result in a -EACCES return code instead of the "expected" -EEXIST mkdir().

This commit changed the error code of mkdir() into a hard error "goto fail":
https://github.com/yuwata/systemd/commit/e4de58c8231e47509ffeb3aa47620ca42f22d7f6

A tested workaround is to stat() the source path before the mkdir() call, so the kernel knows if the directory or file is exists. Then the -EEXIST error code is also returned for NFS mounts and the bind mount proceeds as normal.

We encountered this issue after upgrading from Fedora 34 with systemd 248.10 to Federa 36 with systemd 250.8. Fedora 35 is also affected.

yuwata:
This fixes a bug introduced by e4de58c8231e47509ffeb3aa47620ca42f22d7f6.

Fixes #24120.

### DIFF
--- a/src/core/mount.c
+++ b/src/core/mount.c
@@ -1053,9 +1053,13 @@ static void mount_enter_mounting(Mount *m) {
         if (p && mount_is_bind(p)) {
                 r = mkdir_p_label(p->what, m->directory_mode);
                 /* mkdir_p_label() can return -EEXIST if the target path exists and is not a directory - which is
-                 * totally OK, in case the user wants us to overmount a non-directory inode. */
+                 * totally OK, in case the user wants us to overmount a non-directory inode. Also -EROFS can be
+                 * returned on read-only filesystem. Moreover, -EACCES (and also maybe -EPERM?) may be returned
+                 * when the path is on NFS. See issue #24120. All such errors will be logged in the debug level. */
                 if (r < 0 && r != -EEXIST)
-                        log_unit_error_errno(UNIT(m), r, "Failed to make bind mount source '%s': %m", p->what);
+                        log_unit_full_errno(UNIT(m),
+                                            (r == -EROFS || ERRNO_IS_PRIVILEGE(r)) ? LOG_DEBUG : LOG_WARNING,
+                                            r, "Failed to make bind mount source '%s', ignoring: %m", p->what);
         }
 
         if (p) {


bluca (approved):

poettering:
I don't grok the explanation. 

I am tempted to say, we should just log about any warning from mkdir, and the continue with the mount()? I mean, if we fail with the mkdir() then the worst that can happen is that the mount point inode won't exist, which will make the mount() later fail. But that's totally OK, given the failure happens either way.

hence, I'd probably rewrite this to be non-fatal, and then issue a log_full() with the log level either be LOG_DEBUG (in case of ERRNO_IS_PERMISSION or EEXIST), or LOG_WARNING (since something is off)

poettering:
EROFS also should be handled specially, too, see issue

yuwata:
@poettering Thank you for the suggestion. Updated. PTAL!

src/core/mount.c
```diff
@@ -1053,11 +1053,14 @@ static void mount_enter_mounting(Mount *m) {
...
-                        log_unit_error_errno(UNIT(m), r, "Failed to make bind mount source '%s': %m", p->what);
-                        goto fail;
-                }
+                 * totally OK, in case the user wants us to overmount a non-directory inode. Also -EROFS can be
+                 * returned on read-only filesystem. Moreover, -EACCES (and also maybe -EPERM?) may be returned
+                 * when the path is on NFS and the kernel cache about the path is cold. See issue #24120.
+                 * All such errors will be logged in the debug level. */
+                if (r < 0)
+                        log_unit_full_errno(UNIT(m),
+                                            (IN_SET(r, -EEXIST, -EROFS) || ERRNO_IS_PRIVILEGE(r)) ? LOG_DEBUG : LOG_WARNING,
```
bluca: I don't think this should log at all for EEXIST? It will just add noise to the debug logs, and it's not even an ignorable error, it's the expected final state

bluca: so maybe do `if (r < 0 && r != -EEXIST)` as the original check did?

poettering: yeah, probably

src/core/mount.c
```diff
@@ -1053,11 +1053,14 @@ static void mount_enter_mounting(Mount *m) {
...
                 r = mkdir_p_label(p->what, m->directory_mode);
                 /* mkdir_p_label() can return -EEXIST if the target path exists and is not a directory - which is
-                 * totally OK, in case the user wants us to overmount a non-directory inode. */
-                if (r < 0 && r != -EEXIST) {
-                        log_unit_error_errno(UNIT(m), r, "Failed to make bind mount source '%s': %m", p->what);
-                        goto fail;
-                }
+                 * totally OK, in case the user wants us to overmount a non-directory inode. Also -EROFS can be
+                 * returned on read-only filesystem. Moreover, -EACCES (and also maybe -EPERM?) may be returned
+                 * when the path is on NFS and the kernel cache about the path is cold. See issue #24120.
```
poettering: i'd drop the comment about the kernel cache. I think what matters is that on Linux if we try to create an inode that already exists and also lack the perms to do it, then EEXIST generally trumps EACCES. However, on NFS servers might do it the other way. Hence we need to handle both permission issues and EEXIST issues the same way.

src/core/mount.c
yuwata: OK. Done.

src/core/mount.c
yuwata: OK. Dropped.

yuwata:
@bluca and @poettering Thank you for the comments. The above points are addressed. Upgrading the green label.

PR #24127, opened 2022-07-26
gcrypt: prefer the OS RNG

crrodriguez:
by default, gcrypt defaults to an userspace RNG, this is
the wrong thing (tm) to do on linux.

Switch to the SYSTEM rng instead.

### DIFF
--- a/src/basic/gcrypt-util.c
+++ b/src/basic/gcrypt-util.c
@@ -15,6 +15,8 @@ void initialize_libgcrypt(bool secmem) {
          * feature should initialize the library manually */
         if (!secmem)
                 gcry_control(GCRYCTL_DISABLE_SECMEM);
+
+        gcry_control(GCRYCTL_SET_PREFERRED_RNG_TYPE, GCRY_RNG_TYPE_SYSTEM);
         gcry_control(GCRYCTL_INITIALIZATION_FINISHED, 0);
 }
 


crrodriguez:
@zx2c4 another userspace rng here .. maybe you got enough traction to convince gcrypt devs this is a bad default.

bluca (approved):
Of course the question is, why isn't the "right thing" also the "default thing"...

poettering:
actually userspace rng such as arc4random (which recently was added to glibc) and is seeded regularly by getrandom() should be more than fine. It increases throughput. Not sure what gcrypt is doing there, but if it does something like that I see no problem with it.

Do we actually rely on any gcrypt functionality that needs an RNG?

zx2c4:
`arc4random()` will now just call getrandom, without buffering. Safely buffering in userspace needs vDSO/kernel support, which is a work in progress. Until then, don't buffer in userspace. This conversation is going on elsewhere and I'm not going to duplicate it here. Suffice to say that this patch looks good to me.

poettering:
> `arc4random()` will now just call getrandom, without buffering.

Is that so? What do you mean by buffering?

https://sourceware.org/git/?p=glibc.git;a=commit;h=6f4e0fcfa2d2b0915816a3a3a1d48b4763a7dee2 ← that suggests getrandom() us used for reseeding every 16M only. In between it's something based on ChaCha20

zx2c4:
Yes, that is so. https://sourceware.org/git/?p=glibc.git;a=commit;h=eaad4f9e8f07fc43618f6c8635a7e82831a423dd

poettering:
oha, interesting. thanks for the pointer.

Issue #6036, opened 2017-05-26
systemd timers do not reset after a time/date change


zealws:
### Submission type

 - Bug report

### systemd version the issue has been seen with

> systemd 233

### Used distribution

> CentOS 7, Arch Linux

### In case of bug report: Expected behaviour you didn't see

> after a date change, `systemctl list-timers` should show times that are in the near future, regardless of the date before the change.

### In case of bug report: Unexpected behaviour you saw

> after a date change from 2020 to now, `systemctl list-timers` had times that were in 2020, and indicated the timers had `3 years left`

### In case of bug report: Steps to reproduce the problem

> create a timer that triggers periodically (e.g. `daily`)
> make sure the timer is started and shows up correctly in `systemctl list-timers`
> set the date into the past by 2 years
> `systemctl list-timers` will show a `NEXT` time that's 2 years in the future and `LEFT` will show `2 years left`

poettering:
what's the precise timer unit file you used?

ghost:
Same problem here - new motherboards clock defaulted to 2045-07-15.

    $ systemctl list-timers --no-pager                                                                                                                                       
    NEXT                         LEFT                    LAST                         PASSED                  UNIT                         ACTIVATES
    Sat 2017-07-29 18:55:49 PDT  20h left                Fri 2017-07-28 18:55:49 PDT  3h 15min ago            systemd-tmpfiles-clean.timer systemd-tmpfiles-clean.service
    Sun 2045-07-16 00:00:00 PDT  27 years 11 months left Sat 2045-07-15 13:18:15 PDT  27 years 11 months left logrotate.timer              logrotate.service
    Sun 2045-07-16 00:00:00 PDT  27 years 11 months left Sat 2045-07-15 13:18:15 PDT  27 years 11 months left man-db.timer                 man-db.service
    Sun 2045-07-16 00:00:00 PDT  27 years 11 months left Sat 2045-07-15 13:18:15 PDT  27 years 11 months left shadow.timer                 shadow.service
    Mon 2045-07-17 00:00:00 PDT  27 years 11 months left Sat 2045-07-15 13:18:15 PDT  27 years 11 months left fstrim.timer                 fstrim.service

systemd 234.11 on Arch Linux

smldis:
@poettering This is a serious bug since it is difficult to notice that your timers are not running until the user runs `systemctl list-timers` and look at the results driven by curiosity or by debugging an already happened issue.

Does someone have a workaround?

[...]

PR #24131, opened 2022-07-26
Fix issue with system time set back

joaofl:
As discussed in https://github.com/systemd/systemd/issues/6036, there is an issue with systemd timers in which the next trigger will get stuck with a future date (greater than the period), whenever the system time is set-back after the timer initialization.

In this PR I amended the testsuite 53 by introducing the failing scenario, and the potential fix for the problem.


joaofl:
@mrc0mmand 
Seems that all checks have passed this time. Do you know who could give a review to this PR?

src/core/timer.c
```diff
@@ -394,6 +394,15 @@ static void timer_enter_waiting(Timer *t, bool time_change) {
                 if (v->base == TIMER_CALENDAR) {
                         usec_t b, rebased;
 
+                        /* Update last_trigger to 'now' in case
+                         * the system time changes, so that
+                         * next_elapse is not stuck with a
+                         * future date. */
+                        if (time_change){
+                                t->last_trigger.realtime = ts.realtime;
```
yuwata: This sounds hackish for me. This intends to set `b = ts.realtime` in the below, right? How about to update the below condition like the following?
```c
if (time_change)
        b = ts.realtime;
else if (t->last_trigger.realtime > 0)
        b = t->last_trigger.realtime;
else if (state_translation_table[t->state] == UNIT_ACTIVE)
        b = UNIT(t)->inactive_exit_timestamp.realtime;
else
        b = ts.realtime;
```

src/core/timer.c
```diff
@@ -394,6 +394,15 @@ static void timer_enter_waiting(Timer *t, bool time_change) {
                 if (v->base == TIMER_CALENDAR) {
                         usec_t b, rebased;
 
+                        /* Update last_trigger to 'now' in case
+                         * the system time changes, so that
+                         * next_elapse is not stuck with a
+                         * future date. */
```
yuwata: Nowadays, we allow ~109ch per line. Please do not wrap lines so frequently.

yuwata (changes requested):
Sorry for late review.

src/core/timer.c
joaofl: Your suggestion seems correct to me, and indeed less hacky. It also pass the tests. Thus I am resolving this comment.

src/core/timer.c
joaofl: Changed

joaofl:
@yuwata I believe the patch is in good shape now for another check. Thanks

yuwata (approved):

joaofl:
@yuwata thanks for the review.
Any more actions required from my side?

Issue #23494, opened 2022-05-24
LLMNR resolution breaks some verification queries.


pemensik:
**systemd version the issue has been seen with**
 > systemd-251-1.fc37.x86_64

**Used distribution**
 > Fedora Rawhide (37)

**Linux kernel version used** (`uname -a`)

 > `Linux *.redhat.com 5.18.0-0.rc7.54.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Mon May 16 14:40:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**CPU architecture issue was seen on**
 > x86_64

**Expected behaviour you didn't see**
 > All dnssec enabled queries reach destination

**Unexpected behaviour you saw**
 > dig org. dnskey ended with NXDOMAIN, which is a nonsense

**Steps to reproduce the problem**
 > dnf install ldns-utils && drill -S org. dnskey

**Additional program output to the terminal or log subsystem illustrating the issue**
- LLMNR=yes
- DNSSEC=yes

```
# drill -S src.fedoraproject.org @127.0.0.53
;; Number of trusted keys: 1
;; Chasing: src.fedoraproject.org. A


DNSSEC Trust tree:
src.fedoraproject.org. (A)
|---fedoraproject.org. (DNSKEY keytag: 60624 alg: 14 flags: 256)
    |---fedoraproject.org. (DNSKEY keytag: 58125 alg: 14 flags: 257)
    |---fedoraproject.org. (DS keytag: 58125 digest type: 2)
No trusted keys found in tree: first error was: No DNSSEC public key(s)
;; Chase failed.
```

- LLMNR=no
- DNSSEC=yes
```
# drill -S src.fedoraproject.org @127.0.0.53
;; Number of trusted keys: 1
;; Chasing: src.fedoraproject.org. A


DNSSEC Trust tree:
src.fedoraproject.org. (A)
|---fedoraproject.org. (DNSKEY keytag: 60624 alg: 14 flags: 256)
    |---fedoraproject.org. (DNSKEY keytag: 58125 alg: 14 flags: 257)
    |---fedoraproject.org. (DS keytag: 58125 digest type: 2)
        |---org. (DNSKEY keytag: 15843 alg: 8 flags: 256)
            |---org. (DNSKEY keytag: 26974 alg: 8 flags: 257)
            |---org. (DS keytag: 26974 digest type: 2)
                |---. (DNSKEY keytag: 47671 alg: 8 flags: 256)
[...]

Issue #13763, opened 2019-10-12
re-assess the design decision to have all single-label DNS requests effectively get black-holed


fermulator:
> systemd --version
systemd 237
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD -IDN2 +IDN -PCRE2 default-hierarchy=hybrid

**Used distribution**
 > Ubuntu 18.04.3 LTS

**Expected behaviour you didn't see**
 > As per https://github.com/systemd/systemd/issues/10298, local single-label network name resolution MUST continue to work.  That is, in networks that have DNS pointing to a local server on the same subnet, we SHOULD continue to forward requests there, but NOT if it is external/public.

> Further, although perhaps confounded, in https://github.com/systemd/systemd/issues/2514, clearly some decisions were made for technical and purist reasons (or misinterpretation of the standard) - resulting in requirements/functionality breaking

**Unexpected behaviour you saw**
 > systemd-resolv is unable to resolve single label nameservers via the local DNS, but if we ask the local DNS directly via `nslookup` it works

Consider, the default for systemd-resolv, I will use "fermmy-git" as my example server, which SHOULD resolve to 10.0.0.54.
```
 nslookup fermmy-git
Server:		127.0.0.53
Address:	127.0.0.53#53

** server can't find fermmy-git: SERVFAIL
```

Compare with, how it SHOULD resolve if we ask the router directly:
```
$ nslookup fermmy-git 10.0.0.1
Server:		10.0.0.1
Address:	10.0.0.1#53

Name:	fermmy-git
Address: 10.0.0.54
```

**Steps to reproduce the problem**

Start with the very common Ubuntu distribution, <=16.04, that leverages resolvconf buy default.
Your network shall have a router, with static or DHCP where all systems point their primary DNS to the router.
The router has some internal nameservers (e.g. static DHCP entries), for systems within the network to do name-IP-resolution.
e.g. "my-server" should resolve to "192.168.0.10" (or whatever)

Upgrade to 18.04 (start using `systemd-resolv`).

If we use all the standard tools (`ping`, `nslookup`, `systemd-resolve`, ...) -but _especially_ from web browsers or applications that are hooked in by local name ... they will not resolve.  We have broken backwards compatibility for this very standard network design.

----
Other refs (as I find them)
[...]

PR #24133, opened 2022-07-26
resolved: fix single-label resolution over DNS

jacekmigacz:


poettering:
Hmm? You are aware of `ResolveUnicastSingleLabel=`?

jacekmigacz:
@poettering I'm, but I observed, that, when LLMNR claims itself authoritative for single-labeled domains, DNS scope is not even submitted as a candidate; LLMNR's `DNS_SCOPE_YES_BASE + 1` beats DNS' `DNS_SCOPE_MAYBE`.
Thus, even with `ResolveUnicastSingleLabel=yes` it is not possible to resolve TLD's root like `org.`.

jacekmigacz:
...also it resolves #23494, #13763 and couple others.

jacekmigacz:
That should do the trick.

Also, @poettering I wonder what is the meaning of `Unicast` in `ResolveUnicastSingleLabel`?
Is it suppose to be opposite to multicast? Is it about routing methodology or something else?
I'd reduce it to `ResolveSingleLabel` or change it to `ResolveNonMulticastSingleLabel`.

poettering:
We distuingish between "multicast" DNS (i.e. llmnr and mdns) and "unicast DNS" (i.e. classic port 53 DNS), hence the name.

jacekmigacz:
I'm asking, because if it is a routing term, then I find a bit awkward, because for classic DNS it is usually Anycast routing methodology, not Unicast (like Cloudflare[1.1.1.1] or Google[8.8.8.8]). Thus, unless a context is different it might be a bit confusing.

poettering:
Well, the fact that it's unicast or anycast is not relevant for the protocol, it's really just an implementation detail of the server side. But the fact that it is multicast matters a lot for the protocol, and changes almost everything about it.

jacekmigacz:
Fine then. Thanks for a merge!

PR #23213, opened 2022-04-28
Fixes for Jammy CI

bluca:

PR #24134, opened 2022-07-27
Meson fixups, mostly type safety

eli-schwartz:
Discovered when attempting to build systemd using muon.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2229,7 +2229,7 @@ exe = executable(
         dependencies : [versiondep,
                         libseccomp],
         install_rpath : rootpkglibdir,
-        install : conf.get('ENABLE_ANALYZE'))
+        install : conf.get('ENABLE_ANALYZE') == 1)
 public_programs += exe
 
 if want_tests != 'false'


eli-schwartz:
```
WARNING: Project targeting '>= 0.53.2' but tried to use feature introduced in '0.54.1': negative numbers in meson_options.txt.

meson_options.txt:221:0: ERROR: Fatal warnings enabled, aborting
```

That's pretty lol, I have to say. Negative numbers in meson_options.txt work if you rely on type unsafety converting it from string to int, but didn't work if it was an actual int. Sigh.

I think this depends on #23391 for at least the commit that bumps the requirement to 0.54 (available in Debian stable, latest Ubuntu LTS and non-LTS, all supported Fedora versions, etc.)

bluca:
> ```
> WARNING: Project targeting '>= 0.53.2' but tried to use feature introduced in '0.54.1': negative numbers in meson_options.txt.
> 
> meson_options.txt:221:0: ERROR: Fatal warnings enabled, aborting
> ```
> 
> That's pretty lol, I have to say. Negative numbers in meson_options.txt work if you rely on type unsafety converting it from string to int, but didn't work if it was an actual int. Sigh.
> 
> I think this depends on #23391 for at least the commit that bumps the requirement to 0.54 (available in Debian stable, latest Ubuntu LTS and non-LTS, all supported Fedora versions, etc.)

Just drop the first commit? The rest look fine to me

evverx:
> at least the commit that bumps the requirement to 0.54 (available in Debian stable, latest Ubuntu LTS and non-LTS, all supported Fedora versions, etc.)

As far as I understand the policy allows that so I'd go ahead and bump it. Having said that in reality regardless of what the policy says it can't be done until https://github.com/systemd/systemd/pull/23213 is addressed somehow and as far as I can see it was abandoned.

mbiebl:
> Having said that in reality regardless of what the policy says it can't be done until #23213 is addressed somehow and as far as I can see it was abandoned.

CI is indeed still using Focal. We could backport meson though for CI as we've done in the past via https://launchpad.net/~upstream-systemd-ci

bluca:
> CI is indeed still using Focal. We could backport meson though for CI as we've done in the past via https://launchpad.net/~upstream-systemd-ci

The CI is not an issue, creating needless friction and grief for users is. As far as I can see this is a cosmetic change, there's no bug or functional problem?

evverx:
> creating needless friction and grief for users is

I'm once again not sure what the point of the policy is if it can apparently be bent to keep distributions that aren't covered by it afloat.

> there's no bug or functional problem?

The first commit says what it's trying to accomplish
> other implementations of Meson such as muon choke on it.

eli-schwartz:
> The CI is not an issue, creating needless friction and grief for users is. As far as I can see this is a cosmetic change, there's no bug or functional problem?

Again, the functional problem is generally described by each patch, and again in the PR description: I discovered all these issues while trying to build systemd using [muon](https://sr.ht/~lattis/muon) (a c99 port of Meson). Each commit fixes something that caused muon to error out instead of successfully configuring.

I would call that a bug!

It's a really bad look for systemd to document that certain users and distros are supported, and certain ones are not -- then ignore that documentation and refuse to accept PRs that work perfectly and cause no grief for any (supported) users while proverbially wringing your hands and going "but think of the ~~children~~ users".

bluca:
> > The CI is not an issue, creating needless friction and grief for users is. As far as I can see this is a cosmetic change, there's no bug or functional problem?
> 
> Again, the functional problem is generally described by each patch, and again in the PR description: I discovered all these issues while trying to build systemd using [muon](https://sr.ht/~lattis/muon) (a c99 port of Meson). Each commit fixes something that caused muon to error out instead of successfully configuring.
> 
> I would call that a bug!
> 
> It's a really bad look for systemd to document that certain users and distros are supported, and certain ones are not -- then ignore that documentation and refuse to accept PRs that work perfectly and cause no grief for any (supported) users while proverbially wringing your hands and going "but think of the ~children~ users".

First of all, using loaded hyperboles is not going to help with anything. Secondly, as you can see from the CI list, this PR can't quite be described as "working perfectly" right now, given half the jobs are failing.

Finally and most importantly, the policy you mention is quite clear that on top of the minimum guaranteed, there will be judgement calls made on a case-by-case basis depending on circumstances:

>         systemd main branch and latest major or stable releases are generally
>         expected to compile on current versions of popular distributions (at
>         least all non-EOL versions of Fedora, Debian unstable/testing/stable,
>         latest Ubuntu LTS and non-LTS releases, openSUSE Tumbleweed/Leap,
>         CentOS Stream 8 and 9, up-to-date Arch, etc.)  We will generally
>         attempt to support also other non-EOL versions of various distros.
>         Features which would break compilation on slightly-older distributions
>         will only be introduced if there are significant reasons for this
>         (i.e. supporting them interferes with development or requires too many
>         resources to support). In some cases backports of specific libraries or
>         tools might be required

https://github.com/systemd/systemd/blob/main/README#L277

The first commit here clearly falls into the category described by the second half of the paragraph - it obviously breaks compilation on slightly-older but non-EOL distros, as the CI is showing.
Which means the costs-vs-benefits analysis needs to answer the question on whether there are "significant reasons" to do so.

These are the facts:

- meson, which is what we use and intend to continue to use for the foreseeable future, has no issue with the current handling of default options values, hence while obviously correct the change is only cosmetic
- the only functional benefit is to muon (of which I wasn't aware until now), which appears to be a new and experimental and separate reimplementation of meson, that seems to only be available on Alpine (they don't build systemd), Gentoo (systemd is not default) and some AUR repository: https://repology.org/project/muon-meson/versions
- focal is extremely popular and widely used, and will be for some time still

So, for the meson case, is a cosmetic change a "significant reason"? Answer seems obvious to me.

Which means the question essentially boils down to, is it more important to support muon or to support focal right now?

The answer seems obvious to me: focal is very popular and its replacement was out just a few months ago, which means it will be used (especially in enterprise environments) for quite some time yet, while muon is extremely new and available in a couple of bleeding-edge distros only, so I do not see how it could be a "significant reason".

So if you rebase and drop the first commit, the rest looks mergeable. The first change can then be added to the TODO list for when we will bump the minimum meson version - current tentative plan is second half of next year, after the next Debian release will be out. At that point most developers on focal should have moved on to jammy. We have quite a few things in the TODO list to apply once we can move to 0.60+, and the first commit here can certainly be added to that list.

evverx:
> the policy you mention is quite clear that on top of the minimum guaranteed

I don't think it's clear given that at least two people (including myself) can't say for sure whether Focal is supported or not. I'd say It's intentionally vague to I guess be able to bend it when it's convenient. Then again, it was obvious when it was introduced so I don't know what I expected here :-) It would be nice if the policy could be updated to be a bit more direct about what it actually covers.

> The answer seems obvious to me

It doesn't seem obvious to me.

> which means it will be used (especially in enterprise environments) for quite some time yet

I wonder why upstream projects should care about that?

Then again, I kind of get what is being said here so to avoid wasting contributors' time

yuwata:
@eli-schwartz Could you open a new PR with the last 4 commits? Then, we can merge them earlier.

Personally, I agree with bumping the required version of meson. But we need more discussion about that.

evverx:
> But we need more discussion about that

@yuwata given that is has been discussed several time I think it would be a waste of time at this point. It was pretty clear even before the "policy" was introduced that meson will be bumped next year.

yuwata:
> > But we need more discussion about that
> 
> @yuwata given that is has been discussed several times I think it would be a waste of time at this point.

We have discussed, but we do not have any consensus about that, if I understand correctly.
If someone doubt about that, then we need to consider that sincerely. Especially when issues are not critical. And, though the issues fixed by this PR are clearly bug I think, but the significance is extremely low. These are different from the runtime issue.

> It was pretty clear even before the "policy" was introduced that meson will be bumped next year.

I am not sure about that. We have some implied "policy" before, and bumped the required version based on that. But I have no idea about all members have the same "policy" in their brain. We should not discuss any more about the implied old policy.

--------
Note, again, I agree with bumping the required version of meson. From our policy:
```
In some cases backports of specific libraries or tools might be required.
```
and installing newer meson is not hard (compared with installing newer gcc, glibc, or so).
IIRC, CentOS requires EPEL for installing newer tools to build systemd. Hence, it is OK if bullseye-backports has sufficiently newer version for Debian case.
For Ubuntu case, the latest LTS 22.04 (jammy) was already released. Hence, So, we can cut LTS 20.04 (focal) support when our ubuntu CIs moved to jammy, that's only blocker for bumping that, I think.

mbiebl:
> and installing newer meson is not hard in general (compared with installing newer gcc, glibc, or so).
> IIRC, CentOS requires EPEL for installing newer tools to build systemd. Hence, it is OK if bullseye-backports has sufficiently newer version for Debian case.

Debian stable/bullseye has 0.56.2, so you don't actually need bullseye-backports

> For Ubuntu case, the latest LTS 22.04 (jammy) was already released. Hence, So, we can cut LTS 20.04 (focal) support when our ubuntu CIs moved to jammy, that's only blocker for bumping that, I think.

As outlined above, moving the CI to jammy is not really a blocker as we can provide a newer meson via the `~upstream-systemd-ci` repo. That said, moving to jammy at some point is certainly a good idea anyway.

I don't really have an opinion whether the meson requirement should be bumped for this change or not. I can see bluca's point of not bumping dependencies graciously. I do think build time dependencies are less problematic though.

yuwata:
> > and installing newer meson is not hard in general (compared with installing newer gcc, glibc, or so).
> > IIRC, CentOS requires EPEL for installing newer tools to build systemd. Hence, it is OK if bullseye-backports has sufficiently newer version for Debian case.
> 
> Debian stable/bullseye has 0.56.2, so you don't actually need bullseye-backports

Nice!

> > For Ubuntu case, the latest LTS 22.04 (jammy) was already released. Hence, So, we can cut LTS 20.04 (focal) support when our ubuntu CIs moved to jammy, that's only blocker for bumping that, I think.
> 
> As outlined above, moving the CI to jammy is not really a blocker as we can provide a newer meson via the `~upstream-systemd-ci` repo. That said, moving to jammy at some point is certainly a good idea anyway.

Thank you for the follow-up. That sounds easier for me, though I am not familiar with the internal of the ubuntu CI.

yuwata:
@eli-schwartz The first commit is cherry-picked into #23391. Could you drop the first commit now?

yuwata:
@bluca
> The first commit here clearly falls into the category described by the second half of the paragraph - it obviously breaks compilation on slightly-older but non-EOL distros, as the CI is showing.

Our policy says the latest LTS and non-LTS ubuntu releases are supported. And focal is not latest.
Do you think it is OK to bump at some level if the ubuntu CI is moved to jammy or has newer meson with the suggested way by @mbiebl ? Or, do you want to change the policy to all LTS releases (though we have dropped bionic support)??

bluca:
> Our policy says the latest LTS and non-LTS ubuntu releases are supported. And focal is not latest.
> Do you think it is OK to bump at some level if the ubuntu CI is moved to jammy or has newer meson with the suggested way by @mbiebl ? Or, do you want to change the policy to all LTS releases (though we have dropped bionic support)??

The CI is not the problem. Again, this is what the policy says:

>     We will generally
>     attempt to support also other non-EOL versions of various distros.
>     Features which would break compilation on slightly-older distributions
>     will only be introduced if there are significant reasons for this
>     (i.e. supporting them interferes with development or requires too many
>     resources to support). In some cases backports of specific libraries or
>     tools might be required

https://github.com/systemd/systemd/blob/main/README#L277

Does this break compilation on older non-EOL versions? Yes, clearly.
Are there significant reasons for doing so such as interfering with development or requiring too many resources to support? No, clearly: things are working fine with Meson, and the only functional benefit is for an experimental tool that's hardly available anywhere, and that we don't intend to use. Are non-functional clean-ups good to have? Yes, obviously, but when there's no functional difference they can't be described as significant.

Breaking backward compatibility willy-nilly is one of the greatest sources of grief, frustration and wasted time in the modern software engineering industry. That's why the policy was written in that way: because on top of the _bare minimum_ (eg: we don't break current releases, don't even try) we also don't want to cause needless grief without a good reason.

By the second half of next year, after Jammy has been out for more than one year and Bookworm is released, the damage done by breaking backward compatibility should be lower, and the benefit of the accumulated list of cleanups should be higher.

Today however, breaking Focal and causing grief for a cosmetic cleanup does not make any sense whatsoever.

yuwata:
OK. Thanks. I am happy that I can confirm that we do not need to support focal until its EOL, hehe :-)
Yeah, jammy may be too new now compared with the benefit of bumping the meson requirement.

But, as I said in the previous comment,
> installing newer meson is not hard in general (compared with installing newer gcc, glibc, or so),

let's not postpone to bump so longer. Maybe, about 2023 April?

Note, as @mbiebl commented, bullseye already has 0.56.2, and is not a blocker for this PR.

bluca:
> OK. Thanks. I am happy that I can confirm that we do not need to support focal until its EOL, hehe :-) Yeah, jammy may be too new now compared with the benefit of bumping the meson requirement.
> 
> But, as I said in the previous comment,
> 
> > installing newer meson is not hard in general (compared with installing newer gcc, glibc, or so),
> 
> let's not postpone to bump so longer. Maybe, about 2023 April?
> 
> Note, as @mbiebl commented, bullseye already has 0.56.2, and is not a blocker for this PR.

Whether it's easy or difficult is not the question, and policy doesn't speak about that. Because easy or difficult are subjective, for a power user it will be trivial, for an inexperienced developer using a platform for the first time it might mean days of grief. And for a user on a locked-down corporate development machine it might simply be impossible due to policy.

Also, we have a bunch of fixes and cleanups queued, it doesn't make much sense to do them one at a time. Better to break compatibility once and rip the bandaid, to avoid continuously causing grief again and again. Hence why waiting for Bookworm is best, which is tentatively scheduled for mid of next year.

evverx:
> but we do not have any consensus about that

@yuwata I don't think it matters these days. The date was set and the "policy" that can be bent to enforce that date was written. There is nothing inherently wrong with that. It's just that the point of that policy was to avoid arguing about stuff like that so I think the policy should be explicit about those timelines and where they come from to avoid wasting contributors' time.

evverx:
Speaking of "build" backward-compatibility it's interesting how it was thrown out of the window in https://github.com/systemd/systemd/pull/23204 (which broke build scripts that have been working for years). So basically the "policy" allows pushing stuff like that without having to bump `meson` and it has enough room to wiggle to keep supporting "slightly-older distributions" (whatever that means). All in all, I don't think that the policy that can't definitely settle stuff like that makes sense.

@eli-schwartz I'm sorry for misleading you. I actually for a second thought that the policy was somewhat enforceable. (It is enforceable when it's convenient though).

yuwata:
Possible mergeable replacement: #24154.

bluca:
> Possible mergeable replacement: #24154.

Thanks, let's close this and continue there

evverx:
@yuwata I wonder if the policy could be updated as well to be more explicit about where timelines come from? Based on my observations "judgement calls" are tied to a certain release cycle and I don't quite understand why it isn't mentioned anywhere explicitly. It would certainly help to avoid wasting contributors' time.

By the way I still don't understand why this was offloaded onto the systemd project and I think something like https://github.com/bus1/dbus-broker/pull/273 would be a better way to deal with "downstream" issues upstream.

yuwata:
@evverx I cannot find any strong reasons to update our policy more strictly. My understanding is that the policy is not a law we need to follow. But the policy should be a baseline of the discussion. And the current policy is I think has good balance for that. We need to discuss every time for each case when it breaks backward compatibility (or which emits warning like `install_tag`), based on its pros and cons.

About `install_tag`, it sounds to have much merit compared with its demerit. Then, we approve it.

About this PR, the merit is quite small compared with its demerit. Hence, we set 'ci-failds/needs-rework'. But, we do not refuse the changes. Moreover, we mostly approve it, and as you can see, I picked all commits in #24154 on top of my tiny commit. Why do you think we waste contributors' time??

> By the way I still don't understand why this was offloaded onto the systemd project and I think something like [bus1/dbus-broker#273](https://github.com/bus1/dbus-broker/pull/273) would be a better way to deal with "downstream" issues upstream.

That's completely different project. Simply, they have their own "policy", which I guess more radical than ours about supporting old distributions.

evverx:
> But the policy should be a baseline of the discussion

It should but in practice it boils down to following a certain release cycle. If it had been mentioned anywhere I'm pretty sure this PR wouldn't have been opened and it would have saved a lot of time.

> We need to discuss every time for each case when it breaks backward compatibility 

I don't know what the point of that is given that debatable stuff would be rejected if it didn't match certain implicit expectations. They should be explicit.

> About install_tag, it sounds to have much merit compared with its demerit

I wrote that reply mostly in response to all that stuff about how important backward compatibility is. Actually it isn't if it's not convenient. There are ways to make it backward-compatible but as discussed they would make all that stuff look like a kludge (which it is) and it would be hard to maintain.

> Why do you think we waste contributors' time??

Because it takes time to read vague policies, write patches following those policies, coming up with ideas and then getting lectured on how important backward-compatibility is. I'm surprised typosquatting hasn't come up here yet. It all can be avoided if implicit stuff would be explicit.

> That's completely different project.

It is and I'm not saying that systemd should be like `dbus-broker`. That particular PR just shows that sometimes it's fine to say "no" to patches putting the burden of maintaining outdated dependencies upstream. That's what distributions are for. I'm not saying that `systemd` should start breaking builds left and right but all that stuff about building the latest version of systemd on distributions with `meson-0.53` in corporate environments without access to PYPI because of typosquatting (but with access to GitHub) is getting ridiculous.

eli-schwartz:
Honestly, this has been a bit of an exasperating experience for me, quite frankly. For a bit of context here, I often try to contribute to projects for the sake of using better, more idiomatic Meson... I ***ALSO*** try to solve problems before they become "it doesn't build" style bugs, and have devoted considerable efforts to release candidate testing.

I submitted my PR because anything that breaks muon will also, likely, break Meson, if not now, then in the future. (Some parts of my PR are required for in-flight Meson PRs, in fact). More generally, it's an explicit goal of Meson to be type-safe, and I'd very much like that work to be completed before a Meson 1.0 (which has actually been floated as an idea to have for the 10th anniversary of Meson).

In response, I get wishy-washy "we have a policy, but we've carefully worded it to not actually mean anything at all, go home", "cleanups with no functional differences" (I literally promise you this is untrue), and "let me cherry-pick your entire PR into my own PR with a tiny tweak" (thanks, I guess? a review comment would have worked too)

Anyway, problem resolved, but it wasn't exactly a very comfortable way of doing it. This saps my energy to care about systemd in the future, quite frankly. You're welcome to learn about type safety bugs the standard way, next time -- when trivial refactorings in Meson break it and systemd no longer builds with the latest release.

> I cannot find any strong reasons to update our policy more strictly. My understanding is that the policy is not a law we need to follow.

I don't care if you update the policy more strictly. Just please, for the love of G-d, update your policy more ***clearly***. Currently, it misleads people into thinking it says something it does not say, then people spend the better part of a week and 30 comments engaging in catfights over what it actually means.

It boggles my mind that anyone could be happy with that state of affairs, and it drives off contributors who don't care what your minimum policy is, but do want you to communicate it clearly so they don't have to waste time trying to figure out what it is.

Issue #23295, opened 2022-05-07
systemd-cryptenroll: add cryptsetup like --key-file option to specify a key-file instead of password


crestfalln:
**Is your feature request related to a problem? Please describe.**
Right now it is not possible to use key-files for decryption purposes, I have been using the PASSWORD env variable but that's a subpar solution.

**Describe the solution you'd like**
A --key-file option like that in cryptsetup would be great.

**The systemd version you checked that didn't have the feature you are asking for**
250

josephlr:
I would also be interested in this feature, as it would allow for easier and more secure automated disk encryption setup.

Ideally, you would be able to do `cryptsetup luksFormat` using a random keyfile, and then use `systemd-cryptenroll` to enroll recovery / TPM2 / password credentials, without the user needing to type their password in multiple times. It could look something like:
```bash
# The keyfile already has high entropy, so set minimum PBKDF settings
cryptsetup luksFormat --key-file=/path/to//keyfile /dev/sda5
# Enroll a human-readable Recovery Key
systemd-cryptenroll --key-file=/path/to//keyfile --recovery-key /dev/sda5
# Enroll a TPM2-based protector
systemd-cryptenroll --key-file=/path/to//keyfile --tpm2-device=auto --tpm2-pcrs=7 /dev/sda5
# Only now, enroll as password and require user input.
# This might be later in the install process (during account creation). 
systemd-cryptenroll --key-file=/path/to//keyfile --password /dev/sda5
````
where `/path/to//keyfile` might be on a tmpfs or similar.

**Comment by crestfalln** [comment]
@poettering I am asking for the ability to use a key-file(binary or human-readable) to enroll other tokens/keys, one way of doing this would be via a `--key-file` option. 
Right now I am reading the key-file to the `PASSWORD` env variable to achieve this, but the discussion in issue #20955 discourages such use.
I also tried passing the key-file via stdin but it doesn't seem to work.
This doesn't work for me. 
```sh
systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=0+1+7 /dev/sda1 < path-to-keyfile
``` 
The above command just hangs without doing or printing anything and must be killed with C-c. Is this the intended functionality?

josephlr:
Should this be merged with https://github.com/systemd/systemd/issues/20961 ?

josephlr:
[...]

Issue #20961, opened 2021-10-07
RFE: Support reading passphrase from file when using systemd-cryptenroll


rittet:
**Is your feature request related to a problem? Please describe.**
systemd-cryptenroll does not support reading a passphrase from a file, instead it always needs to be entered manually. 

**Describe the solution you'd like**
A solution similar to `cryptsetup --key-file=` to read a passphrase  from a file.

**Describe alternatives you've considered**
Using $PASSWORD. See [#20955](https://github.com/systemd/systemd/issues/20955). (Which is not optimal).

**The systemd version you checked that didn't have the feature you are asking for**
systemd 249 (249.4-2-arch)

j5k:
I need also something like that. At the moment I have the following workaround for silent cryptenroll:

```
/bin/expect <(cat << EOF
spawn systemd-cryptenroll ${LUKS_PARTITION} --password
expect "current passphrase" { send "${CURRENT_PWD}\r" }
expect "new passphrase" { send "${PWD1}\r" }
expect "(repeat)" { send "${PWD2}\r" }
interact
EOF
) $> /dev/null
```

bluca:
https://github.com/systemd/systemd/pull/24138

PR #24138, opened 2022-07-27
adding the option to use a keyfile to unlock the device

Keksgesicht:
This should solve issue #20961 and #23295.

Does someone also need `--keyfile-offset` and `--keyfile-size`?
You could create a new file, fill it with random data, add it via `cryptsetup` to the volume and use this one instead.

For the case of stdin just use `--unlock-key-file=/dev/stdin`.

If I somehow created a memory leak or have to sanitize an input just let me know.

### DIFF
--- a/test/units/testsuite-70.sh
+++ b/test/units/testsuite-70.sh
@@ -11,6 +11,9 @@ dd if=/dev/zero of=$img bs=1024k count=20 status=none
 echo -n passphrase >/tmp/passphrase
 cryptsetup luksFormat -q --pbkdf pbkdf2 --pbkdf-force-iterations 1000 --use-urandom $img /tmp/passphrase
 
+# Unlocking via keyfile
+systemd-cryptenroll --unlock-key-file=/tmp/passphrase --tpm2-device=auto $img
+
 # Enroll unlock with default PCR policy
 env PASSWORD=passphrase systemd-cryptenroll --tpm2-device=auto $img
 /usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1


src/cryptenroll/cryptenroll.c
```diff
@@ -221,6 +226,10 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_enroll_type = ENROLL_RECOVERY;
                         break;
 
+                case ARG_UNLOCK_FILE:
+                        arg_unlock_file = strdup(optarg);
```
bluca: OOM check missing

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +495,30 @@ static int prepare_luks(
                                 strlen(envpw));
                 if (r < 0)
                         return log_error_errno(r, "Password from environment variable $PASSWORD did not work.");
+        } else if(arg_unlock_file) {
+                char *password;
```
bluca: `_cleanup_(erase_and_freep) char *password = NULL`

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +495,30 @@ static int prepare_luks(
...
+
+                r = crypt_volume_key_get(
+                                cd,
+                                CRYPT_ANY_SLOT,
+                                vk,
+                                &vks,
+                                password,
+                                passwordLen);
+
+                crypt_safe_free(password);
```
bluca: not needed

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +495,30 @@ static int prepare_luks(
...
+                if (r < 0)
+                        return log_error_errno(r, "Unlocking via keyfile did not work.");
```
bluca: `Unlocking via keyfile failed: %m`

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +495,30 @@ static int prepare_luks(
...
+                size_t passwordLen;
+
+                r = crypt_keyfile_device_read(
+                                cd,
+                                arg_unlock_file,
+                                &password,
+                                &passwordLen,
+                                0, 0, 0);
+                if (r < 0)
+                        return log_error_errno(r, "Unlocking via keyfile did not work.");
```
bluca: `Reading keyfile failed: %m`

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +495,30 @@ static int prepare_luks(
                                 strlen(envpw));
                 if (r < 0)
                         return log_error_errno(r, "Password from environment variable $PASSWORD did not work.");
+        } else if(arg_unlock_file) {
+                char *password;
+                size_t passwordLen;
```
bluca: nit: we don't do camel case

bluca (changes requested):
Please update the documentation, and also add a simple test case to test/units/testsuite-70.sh

src/cryptenroll/cryptenroll.c
```diff
@@ -28,6 +28,7 @@
 #include "tpm2-util.h"
 
 static EnrollType arg_enroll_type = _ENROLL_TYPE_INVALID;
+static char *arg_unlock_file = NULL;
```
bluca: needs a `STATIC_DESTRUCTOR_REGISTER` a bit below

src/cryptenroll/cryptenroll.c
bluca: still not addressed

Keksgesicht: changing this gives me:
```
free(): invalid pointer
Aborted
```
then `--unlock-file` parameter is provided.

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +499,28 @@ static int prepare_luks(
...
+                size_t password_len;
+
+                r = crypt_keyfile_device_read(
+                                cd,
+                                arg_unlock_file,
+                                &password,
+                                &password_len,
+                                0, 0, 0);
+                if (r < 0)
+                        return log_error_errno(r, "Unlocking via keyfile failed: %m");
```
bluca: this is not unlocking, it's reading, so as mentioned: `"Reading keyfile %s failed: %m", arg_unlock_file`

src/cryptenroll/cryptenroll.c
bluca: bah, it looks like it's not an actual string, it's a hidden special structure. You'll need to add a build time check for crypt_safe_free, as it's available in newer versions than the minimum requirement, and wrap the whole block in an ifdef and use crypt_safe_free

src/cryptenroll/cryptenroll.c
poettering: hmm, so crypttab and cryptsetup call the concept "keyfile". We should probably stick to the same naming, to minimize surprises.

src/cryptenroll/cryptenroll.c
```diff
@@ -221,6 +227,13 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_enroll_type = ENROLL_RECOVERY;
                         break;
 
+                case ARG_UNLOCK_FILE:
+                        arg_unlock_file = strdup(optarg);
```
poettering: please use parse_path_argument()

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +499,28 @@ static int prepare_luks(
                                 strlen(envpw));
                 if (r < 0)
                         return log_error_errno(r, "Password from environment variable $PASSWORD did not work.");
+        } else if(arg_unlock_file) {
```
poettering: space between `if` and `(`...

More importantly though. I am pretty sure if the user specified the file on the cmdline, it should take precedence over the PASSWORD env var. I.e. do this stuff first.

src/cryptenroll/cryptenroll.c
```diff
@@ -486,6 +499,28 @@ static int prepare_luks(
...
+                char *password;
+                size_t password_len;
+
+                r = crypt_keyfile_device_read(
+                                cd,
+                                arg_unlock_file,
+                                &password,
+                                &password_len,
+                                0, 0, 0);
```
poettering: please use read_full_file() or read_full_file_full() and use `READ_FULL_FILE_SECURE|READ_FULL_FILE_WARN_WORLD_READABLE|READ_FULL_FILE_CONNECT_SOCKET`

src/cryptenroll/cryptenroll.c
poettering: dont, just use read_full_file() see below, and then use `_cleanup_(erase_and_freep)`

poettering: also, given this is now binary (i.e. not a NUL-terminated string anymore), please use `void*`, and not `char*` (which suggests this was a string)

poettering:
(btw, i figure we should also allow enrolling key files with cryptenroll eventually, and then set a hint on the key slot indicating if it contains non-printable chars, as then it makes little sense to ask for the key interactively at boot)

src/cryptenroll/cryptenroll.c
Keksgesicht: Just to be sure. In https://github.com/systemd/systemd/issues/23295#issuecomment-1122119881 you said `--unlock-file=` make sense to you.
So are we using `--unlock-file=` and `--key-file=` or `--key-file=` and an unnamed parameter with the keyfile path like `cryptsetup`?

poettering: uh, oh. thanks for pointing me to my own earlier contradicting words ;-)

Hmm, so I think my current self would agree with my younger self there, i.e. that this probably should be called something with "unlock" to indicate that this is about unlocking for enrollment, not for enrolling. But my current self also has a point, it should use existing nomenclature. So maybe the combination of both make most sense, i.e. `--unlock-key-file=…`. That way we have the best of both worlds, and call it "key file" but also emphasize it is about "unlocking".

Does that make sense?

poettering:
> (btw, i figure we should also allow enrolling key files with cryptenroll eventually, and then set a hint on the key slot indicating if it contains non-printable chars, as then it makes little sense to ask for the key interactively at boot)

(And yes, apparently my younger self disagree with that opinion ;-))

src/cryptenroll/cryptenroll.c
medhefgo: So this isn't your usual keyfile (e.g. binary blob) we are used to from cryptsetup operations, but rather a file containing the password (which I think is restricted in content to some degree?), right? In that case it should really be called password_file or something like that to drive that point home.

poettering: Hmm, it isn't you usual key file? huh?

medhefgo: My cursory glance at this was that this option is just a means to pass the password to sd-cryptentroll (non-interactively). Whereas a keyfile would be a binary blob and something else entirely. I might be wrong here, though.

poettering: cryptsetup/LUKS do not distinguish passwords from binary blobs used to unlock stuff. It's always just some data with a size...

medhefgo: I would beg to differ (yes, the low-level stuff doesn't care, but the UI should/does): https://man.archlinux.org/man/cryptsetup.8#Passphrase_processing_for_LUKS

poettering: I don't grok what you are trying to say?

Keksgesicht: > So maybe the combination of both make most sense, i.e. `--unlock-key-file=…`. That way we have the best of both worlds, and call it "key file" but also emphasize it is about "unlocking".

This sounds like an good idea.

poettering:
lgtm

MrMarlin:
Hi. Thanks for adding this functionality. When will this be available in a release?

josephlr:
> Hi. Thanks for adding this functionality. When will this be available in a release?

Looking at https://github.com/systemd/systemd/commit/38db7a4ed388b032faae221067b77d553a54e6a6 it seems like this is in v252 (and later).

PR #24140, opened 2022-07-27
journalctl: allow statically linked build

jameshilliard:
The journalctl tool may be needed on cross compilation hosts in order to run --update-catalog against a target rootfs.

To avoid reliability issues caused by shared linking allow journalctl to be linked statically.


meson.build
```diff
@@ -4357,6 +4365,7 @@ foreach tuple : [
         ['link-systemctl-shared', get_option('link-systemctl-shared')],
         ['link-networkd-shared',  get_option('link-networkd-shared')],
         ['link-timesyncd-shared', get_option('link-timesyncd-shared')],
+        ['link-journalctl-shared',get_option('link-journalctl-shared')],
```
yuwata: Please add a white space after `,`, and shift all other entries.

yuwata (changes requested):

meson.build
jameshilliard: entries shifted by one space now

bluca (approved):

yuwata (approved):

PR #24141, opened 2022-07-28
dissect: Add systemd-dissect --umount

daandemeyer:
Still needs an integration test, and probably more graceful error handling as well, but putting up the PR already for discussion to make sure this is the right approach.

This improves on `umount` by trying to make sure the underlying loop device backing the mountpoint is fully removed before returning, hopefully reducing the chance of the annoying loop device race conditions popping up when using `systemd-dissect --mount`.

### DIFF
--- a/src/dissect/dissect.c
+++ b/src/dissect/dissect.c
@@ -8,7 +8,10 @@
 #include <sys/ioctl.h>
 #include <sys/mount.h>
 
+#include "sd-device.h"
+
 #include "architecture.h"
+#include "blockdev-util.h"
 #include "chase-symlinks.h"
 #include "copy.h"
 #include "dissect-image.h"
@@ -24,6 +27,7 @@
 #include "main-func.h"
 #include "mkdir.h"
 #include "mount-util.h"
+#include "mountpoint-util.h"
 #include "namespace-util.h"
 #include "parse-argument.h"
 #include "parse-util.h"
@@ -40,6 +44,7 @@
 static enum {
         ACTION_DISSECT,
         ACTION_MOUNT,
+        ACTION_UMOUNT,
         ACTION_COPY_FROM,
         ACTION_COPY_TO,
 } arg_action = ACTION_DISSECT;
@@ -58,6 +63,7 @@ static VeritySettings arg_verity_settings = VERITY_SETTINGS_DEFAULT;
 static JsonFormatFlags arg_json_format_flags = JSON_FORMAT_OFF;
 static PagerFlags arg_pager_flags = 0;
 static bool arg_legend = true;
+static bool arg_rmdir = false;
 
 STATIC_DESTRUCTOR_REGISTER(arg_verity_settings, verity_settings_done);
 
@@ -81,6 +87,7 @@ static int help(void) {
                "     --fsck=BOOL          Run fsck before mounting\n"
                "     --growfs=BOOL        Grow file system to partition size, if marked\n"
                "     --mkdir              Make mount directory before mounting, if missing\n"
+               "     --rmdir              Remove mount directory after unmounting\n"
                "     --discard=MODE       Choose 'discard' mode (disabled, loop, all, crypto)\n"
                "     --root-hash=HASH     Specify root hash for verity\n"
                "     --root-hash-sig=SIG  Specify pkcs7 signature of root hash for verity\n"
@@ -96,6 +103,8 @@ static int help(void) {
                "     --version            Show package version\n"
                "  -m --mount              Mount the image to the specified directory\n"
                "  -M                      Shortcut for --mount --mkdir\n"
+               "  -u --umount             Unmount the image from the specified directory\n"
+               "  -U                      Shortcut for --umount --rmdir\n"
                "  -x --copy-from          Copy files from image to host\n"
                "  -a --copy-to            Copy files from host to image\n"
                "\nSee the %2$s for details.\n",
@@ -122,6 +131,7 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_ROOT_HASH_SIG,
                 ARG_VERITY_DATA,
                 ARG_MKDIR,
+                ARG_RMDIR,
                 ARG_JSON,
         };
 
@@ -131,6 +141,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "no-pager",      no_argument,       NULL, ARG_NO_PAGER      },
                 { "no-legend",     no_argument,       NULL, ARG_NO_LEGEND     },
                 { "mount",         no_argument,       NULL, 'm'               },
+                { "umount",        no_argument,       NULL, 'u'               },
                 { "read-only",     no_argument,       NULL, 'r'               },
                 { "discard",       required_argument, NULL, ARG_DISCARD       },
                 { "fsck",          required_argument, NULL, ARG_FSCK          },
@@ -139,6 +150,7 @@ static int parse_argv(int argc, char *argv[]) {
                 { "root-hash-sig", required_argument, NULL, ARG_ROOT_HASH_SIG },
                 { "verity-data",   required_argument, NULL, ARG_VERITY_DATA   },
                 { "mkdir",         no_argument,       NULL, ARG_MKDIR         },
+                { "rmdir",         no_argument,       NULL, ARG_RMDIR         },
                 { "copy-from",     no_argument,       NULL, 'x'               },
                 { "copy-to",       no_argument,       NULL, 'a'               },
                 { "json",          required_argument, NULL, ARG_JSON          },
@@ -150,7 +162,7 @@ static int parse_argv(int argc, char *argv[]) {
         assert(argc >= 0);
         assert(argv);
 
-        while ((c = getopt_long(argc, argv, "hmrMxa", options, NULL)) >= 0) {
+        while ((c = getopt_long(argc, argv, "hmurMUxa", options, NULL)) >= 0) {
 
                 switch (c) {
 
@@ -180,6 +192,20 @@ static int parse_argv(int argc, char *argv[]) {
                         /* Shortcut combination of the above two */
                         arg_action = ACTION_MOUNT;
                         arg_flags |= DISSECT_IMAGE_MKDIR;
+                        break;
+
+                case 'u':
+                        arg_action = ACTION_UMOUNT;
+                        break;
+
+                case ARG_RMDIR:
+                        arg_rmdir = true;
+                        break;
+
+                case 'U':
+                        /* Shortcut combination of the above two */
+                        arg_action = ACTION_UMOUNT;
+                        arg_rmdir = true;
                         break;
 
                 case 'x':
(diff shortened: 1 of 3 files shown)


src/dissect/dissect.c
```diff
@@ -823,6 +823,8 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
         return 0;
 }
 
+
+
```
poettering, line 935: hmm?

src/dissect/dissect.c
```diff
@@ -823,7 +841,82 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
 
+static int action_umount(const char *path) {
+        _cleanup_free_ char *absolute = NULL;
+        dev_t devno;
+        const char *node;
+        _cleanup_(loop_device_unrefp) LoopDevice *d = NULL;
+        _cleanup_(sd_device_unrefp) sd_device *device = NULL;
+        _cleanup_(sd_device_enumerator_unrefp) sd_device_enumerator *e = NULL;
+        sd_device *part;
+        int r;
```
poettering: should we refuse early if path_is_mount_point() doesn't hold? I think so, no?

src/dissect/dissect.c
```diff
@@ -823,7 +841,82 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+
+        r = path_make_absolute_cwd(path, &absolute);
```
poettering: hmm, shouldn't this rather be chase_symlinks()? i.e. when operating with mounts that's usually the best approach.

And then maybe even use chase_symlinks_and_open() with O_DIRECTORY. and then use fd_is_mount_point() on the returned fd, as well as fd_get_whole_disk() (i.e. go via fd, rather than path more often then not). Of course, the fd needs be closed before any umount is attempted...

src/dissect/dissect.c
```diff
@@ -823,7 +841,82 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+
+        r = loop_device_flock(d, LOCK_EX);
+        if (r < 0)
+                return log_error_errno(r, "Failed to lock loop device '%s': %m", node);
+
+        r = umount_recursive(absolute, 0);
+        if (r < 0)
+                return log_error_errno(r, "Failed to unmount '%s': %m", absolute);
+
+        r = sd_device_enumerator_new(&e);
```
poettering: i'd split the rest from here down into a separate function delete_block_device_partitions() or so, that takes the sd_device().

And I wonder if this shouldn't rather be called from inside of the loop_device_unref() destructor, so that we can use it all over the place

poettering: Also, a comment would be good, that explains that we do this to gurantee synchronous deletion as much as reasonably possible.

poettering:
looks good, but some comments.

DaanDeMeyer:
I didn't make the change yet where we call block_device_remove_all_partitions() in loop_device_unrefp(). I figure it's better to leave that for another PR

src/basic/mountpoint-util.c
```diff
@@ -187,7 +187,7 @@ int fd_is_mount_point(int fd, const char *filename, int flags) {
 
         /* Insist that the specified filename is actually a filename, and not a path, i.e. some inode further
          * up or down the tree then immediately below the specified directory fd. */
-        if (!filename_possibly_with_slash_suffix(filename))
+        if (!FLAGS_SET(flags, AT_EMPTY_PATH) && !filename_possibly_with_slash_suffix(filename))
```
poettering: see: https://github.com/systemd/systemd/pull/24179

unfortunately I see no way how to make fd_is_mount_point() work on non-dir inodes on old kernels…

src/shared/loop-util.c
```diff
@@ -768,7 +768,12 @@ void loop_device_relinquish(LoopDevice *d) {
         d->relinquished = true;
 }
 
-int loop_device_open(const char *loop_path, int open_flags, LoopDevice **ret) {
+void loop_device_unrelinquish(LoopDevice *d) {
+        assert(d);
+        d->relinquished = false;
+}
+
+int loop_device_open(const char *loop_path, int open_flags, bool relinquish, LoopDevice **ret) {
```
poettering: btw, why add this param? seems exotic enough so that calling loop_device_unrelinquish()/loop_device_relinquish() immediately after loop_device_open() returns should suffice, no?

src/dissect/dissect.c
```diff
@@ -823,6 +857,75 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+
+        r = umount_recursive(canonical, 0);
+        if (r < 0)
+                return log_error_errno(r, "Failed to unmount '%s': %m", canonical);
+
+        if (arg_rmdir && rmdir(canonical) < 0)
+                return log_error_errno(errno, "Failed to remove mount directory '%s': %m", canonical);
+
+        /* We managed to lock and unmount successfully? That means we can try to remove the loop device.*/
+        (void) loop_device_unrelinquish(d);
```
poettering: i don't grok this lock? here you turn off automatic destruction of the loopback device? but that means nothing will destroy it? What am I not seeing?

daandemeyer: The opposite actually, relinquish means we release ownership, unrelinquish means we take ownership of the loop device. So this actually makes sure the automatic destruction runs, which we only want to do if we successfully managed to unmount the mounted directories

src/shared/loop-util.c
daandemeyer: Removed

src/shared/blockdev-util.c
```diff
@@ -476,3 +479,59 @@ int block_device_remove_partition(int fd, const char *name, int nr) {
...
+
+                r = sd_device_get_property_value(part, "PARTN", &v);
+                if (r < 0)
+                        return r;
+
+                r = safe_atoi(v, &nr);
+                if (r < 0)
+                        return r;
+
+                r = block_device_remove_partition(fd, devname, nr);
```
poettering, line 529: Hmm, shouldn't we handle gracefully if this returns ENODEV? (or whatever it returns when a partition is gone already?) That way it should be safe to call this when racing against the same code in the kernel?

src/dissect/dissect.c
```diff
@@ -823,6 +857,73 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+
+        r = umount_recursive(canonical, 0);
+        if (r < 0)
+                return log_error_errno(r, "Failed to unmount '%s': %m", canonical);
+
+        if (arg_rmdir && rmdir(canonical) < 0)
+                return log_error_errno(errno, "Failed to remove mount directory '%s': %m", canonical);
+
+        /* We managed to lock and unmount successfully? That means we can try to remove the loop device.*/
+        loop_device_unrelinquish(d);
```
poettering, line 912: shouldn't move this above the rmdir line above?

src/dissect/dissect.c
```diff
@@ -823,6 +857,73 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+
+        /* Before loop_device_unrefp() kicks in, let's explicitly remove all the partition subdevices of the
+         * loop device. We do this to ensure that all traces of the loop device are gone by the time this
+         * command exits. */
+        r = block_device_remove_all_partitions(d->fd);
+        if (r < 0)
+                return log_error_errno(r, "Failed to remove loop device '%s' partitions: %m", devname);
+
```
poettering, line 932: hmm, i think we should probably do block_device_remove_all_partitions() regardless if the rmdir worked or not. Right now we only do this when rmdir succeeded. hence maybe not fail immediately on rmdir() failures, but just remember them and propagate them only once the unrelinquish/block_device_remove_all_partitions() was done... i.e.

```c
loop_device_unrelinquish(d);

if (arg_rmdir) {
        k = rmdir(canonical);
        if (k < 0)
                log_error_errno(k, …);
} else
        k = 0;

r = block_device_remove_all_partitions(d->fd);
if (r < 0)
        log_error_errno(r, …);

return k < 0 ? k : r;
``

**Comment by poettering** [comment]
looks pretty good, just some minor things, and probably should be rebased on #24179

File: src/shared/blockdev-util.c (hunk 8 above)
Comment (daandemeyer) [reply src/shared/blockdev-util.c:529]: Should we do something with EBUSY as well here? `loop_device_unref()` already has some nasty EBUSY handling, not sure if we need that here as well

Comment (poettering) [reply src/shared/blockdev-util.c:529]: makes sense to me

File: src/dissect/dissect.c [hunk 11]
```diff
@@ -823,6 +857,77 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+        if (r < 0)
+                return log_error_errno(r, "Failed to unmount '%s': %m", canonical);
+
+        /* We managed to lock and unmount successfully? That means we can try to remove the loop device.*/
+        loop_device_unrelinquish(d);
+
+        if (arg_rmdir) {
+                k = RET_NERRNO(rmdir(canonical));
+                if (k < 0)
+                        log_error_errno(errno, "Failed to remove mount directory '%s': %m", canonical);
```
poettering: first param should be `k`,  not `errno

src/shared/blockdev-util.c
```diff
@@ -377,3 +383,159 @@ int path_is_encrypted(const char *path) {
...
+                if (r == -ENODEV) {
+                        log_debug("Kernel removed partition before us, ignoring");
```
poettering: maybe include devname in message output

src/shared/blockdev-util.c
daandemeyer: Updated to not fail on the first error, but keep track of it and still try to remove the remaining partitions before returning the error.

src/shared/blockdev-util.c
```diff
@@ -377,3 +383,162 @@ int path_is_encrypted(const char *path) {
...
+                r = sd_device_get_property_value(part, "PARTN", &v);
+                if (r < 0)
+                        return r;
+
+                r = safe_atoi(v, &nr);
+                if (r < 0)
+                        return r;
+
+                r = block_device_remove_partition(fd, devname, nr);
+                if (r == -ENODEV) {
```
poettering, line 530: you wanted to add EBUSY here too

daandemeyer: Do we just want to ignore EBUSY here though? I figure it makes more sense to ignore that higher up, since if we get EBUSY, that means we failed to remove a partition, and the function didn't actually do what its name advertises.

poettering: ok, but then action_umount() should handle the EBUSY gracefully?

daandemeyer: We now return EXIT_SUCCESS even if we get EBUSY in action_umount()

src/dissect/dissect.c
```diff
@@ -823,6 +857,82 @@ static int action_copy(DissectedImage *m, LoopDevice *d) {
...
+                        log_error_errno(k, "Failed to remove mount directory '%s': %m", canonical);
+        } else
+                k = 0;
+
+        /* Before loop_device_unrefp() kicks in, let's explicitly remove all the partition subdevices of the
+         * loop device. We do this to ensure that all traces of the loop device are gone by the time this
+         * command exits. */
+        r = block_device_remove_all_partitions(d->fd);
+        if (r == -EBUSY) {
+                log_error("One or more partitions of '%s' are busy, ignoring", devname);
```
poettering: i would still use `log_error_errno(r, …` here instead of `log_error(`. Note that the error number if specified is passed into ERRNO= in structured logging, and that is useful even if `%m` is not used in the format string.

bluca (approved):

PR #24143, opened 2022-07-28
virt: detect KubeVirt instance

ShirleyFei:
Kubevirt is currently technically based on KVM (but not xen yet[1]).
The systemd-detect-virt command, used to differentiate the current
virtualization environment, works fine on x86 relying on CPUID, while
fails to get the correct value (none instead of kvm) on aarch64.

Let's fix this by adding a new 'vendor[KubeVirt] = kvm' classification
considering the sys_vendor is always KubeVirt.

[1] https://groups.google.com/g/kubevirt-dev/c/C6cUgzTOsVg

Signed-off-by: Fei Li <lifei.shirley@bytedance.com>

### DIFF
--- a/src/basic/virt.c
+++ b/src/basic/virt.c
@@ -158,6 +158,7 @@ static Virtualization detect_vm_dmi_vendor(void) {
         } dmi_vendor_table[] = {
                 { "KVM",                 VIRTUALIZATION_KVM       },
                 { "OpenStack",           VIRTUALIZATION_KVM       }, /* Detect OpenStack instance as KVM in non x86 architecture */
+                { "KubeVirt",            VIRTUALIZATION_KVM       }, /* Detect KubeVirt instance as KVM in non x86 architecture */
                 { "Amazon EC2",          VIRTUALIZATION_AMAZON    },
                 { "QEMU",                VIRTUALIZATION_QEMU      },
                 { "VMware",              VIRTUALIZATION_VMWARE    }, /* https://kb.vmware.com/s/article/1009458 */


yuwata (approved):
Not familiar with that, but superficially LGTM. At least, this should not break other environments.

poettering:
is kubevirt qemu?

poettering:
(I am asking as our 'kvm' id is a bit of a misnomer, we only identify the qemu+kvm combo with that. If qemu is used without kvm, it's `qemu` and if kvm is used with some other userspace tool, the id reflects the name of that other userspace tool. hence, what does kubevirt do? is it just a wrapper around qemu that tells it to override its DMI/CPUID? or is it an independent userspace tool directly interfacing with kvm?)

ShirleyFei:
> (I am asking as our 'kvm' id is a bit of a misnomer, we only identify the qemu+kvm combo with that. If qemu is used without kvm, it's `qemu` and if kvm is used with some other userspace tool, the id reflects the name of that other userspace tool. hence, what does kubevirt do? is it just a wrapper around qemu that tells it to override its DMI/CPUID? or is it an independent userspace tool directly interfacing with kvm?)

Kubevirt architecture is designed to manage VM via libvirt => qemu => kvm on a kubernetes cluster (both libvirt and qemu lives in each pod). In this way, it is similar to openstack. PS: there was once a topic about integrating xen with kubevirt, but not accepted. I mentioned this in the commit message[1].
Thus we could conclude kubevirt uses kvm. :)

PR #24144, opened 2022-07-28
repart: Make boolean arguments optional

daandemeyer:
Allows e.g. "systemd-repart --dry-run" instead of
"systemd-repart --dry-run yes".

### DIFF
--- a/src/partition/repart.c
+++ b/src/partition/repart.c
@@ -4075,12 +4075,13 @@ static int help(void) {
                "     --version            Show package version\n"
                "     --no-pager           Do not pipe output into a pager\n"
                "     --no-legend          Do not show the headers and footers\n"
-               "     --dry-run=BOOL       Whether to run dry-run operation\n"
+               "     --dry-run[=BOOL]     Whether to run dry-run operation\n"
                "     --empty=MODE         One of refuse, allow, require, force, create; controls\n"
                "                          how to handle empty disks lacking partition tables\n"
-               "     --discard=BOOL       Whether to discard backing blocks for new partitions\n"
-               "     --pretty=BOOL        Whether to show pretty summary before doing changes\n"
-               "     --factory-reset=BOOL Whether to remove data partitions before recreating\n"
+               "     --discard[=BOOL]     Whether to discard backing blocks for new partitions\n"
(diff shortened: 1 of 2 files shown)


src/partition/repart.c
```diff
@@ -4131,15 +4132,15 @@ static int parse_argv(int argc, char *argv[]) {
...
                 { "empty",             required_argument, NULL, ARG_EMPTY             },
                 { "discard",           required_argument, NULL, ARG_DISCARD           },
-                { "factory-reset",     required_argument, NULL, ARG_FACTORY_RESET     },
+                { "factory-reset",     optional_argument, NULL, ARG_FACTORY_RESET     },
                 { "can-factory-reset", no_argument,       NULL, ARG_CAN_FACTORY_RESET },
                 { "root",              required_argument, NULL, ARG_ROOT              },
                 { "image",             required_argument, NULL, ARG_IMAGE             },
                 { "seed",              required_argument, NULL, ARG_SEED              },
-                { "pretty",            required_argument, NULL, ARG_PRETTY            },
+                { "pretty",            optional_argument, NULL, ARG_PRETTY            },
```
poettering, line 4143: hmpf. the `optional_argument` stuff in getop_long() is pretty crappy.

The thing is that for `no_argument` and `required_argument` it's very clear what this is:

```
foo --bar waldo
```

(i.e. for `no_argument` there's one switch `--bar` and one positional argument `waldo`. And for `required_argument` it's one switch with an argument `--bar=waldo`, with no positional argument.)

But for `optional_argument` it's weird. Can you answer correctly how getop_long() will interpret the line above without looking it up? 

(For short options it's even more confusing btw)

Because behaviour is so weird we actually ended up overriding getopt_long()'s own handling in some cases, for example journalctl's `-n` switch, where we manually modify optind in order to do something people are less confused about...

Because of that we kinda stopped using `optional_argument` switch in new code.

I am not sure we should start here again...

daandemeyer: I figured there'd be a reason behind this. Your explanation makes sense, so let's close this. It'd be nice if we could make this work but that seems like it would require more involved changes to argument parsing

PR #24146, opened 2022-07-28
stub: measure kernel/initrd/parameters into clean PCRs 11/12/13, and add "systemd-measure" tool to be able to pre-calculate values

poettering:
Currently PCR measurements on Linux are hard to pre-calculate. Let's do something about this: in sd-stub, let's cleanly measure the kernel/initrd/parameters we are about to invoke into clean, separate PCRs. That way, given any unified kernel image the expected value for PCR 12 
seen after boot can cleanly determined.

Then, let's add a tool to pre-calculate them, that takes similar parameters as the "objcopy" call that merges the blobs with the stub.

This is preparation to later being able to sign expect PCR values ahead of time, to allow cryptsetup TPM policies that bind to signatures of expected PCR values instead of the PCR values themselves, alleviating the PCR "brittleness" problem.

various other fixes, too.

### DIFF
--- a/TODO
+++ b/TODO
@@ -115,6 +115,14 @@ Features:
   on other disks. Always boot into them via NextBoot EFI variable, to not
   affect PCR values.
 
+* systemd-measure tool:
+  - pre-calculate PCR 12 (command line) + PCR 13 (sysext) the same way we can precalculate PCR 11
+  - sign pre-calculated hashes in a way compatible with TPM2 PCR hash signature
+    policies, in a way they can be included in unified PE kernel images, and
+    made available to userspace. There, this should be consumed by
+    systemd-cryptsetup to implement PCR signature based TPM volume unlock
+    policies.
+
 * in sd-boot: load EFI drivers from a new PE section. That way, one can have a
   "supercharged" sd-boot binary, that could carry ext4 drivers built-in.
 
@@ -381,12 +389,6 @@ Features:
   case the same wd is reused multiple times before we start processing
   IN_IGNORED again)
 
-* sd-stub: set efi var indicating stub features, i.e. whether they pick up
-  creds, sysexts and so on. similar to existing variable of sd-boot
-
-* sd-stub: set efi vars declaring TPM PCRs we measured creds/cmdline + sysext
-  into (even if we hardcode them)
-
 * systemd-fstab-generator: support addition mount specifications via kernel
   cmdline. Usecase: invoke a VM, and mount a host homedir into it via
   virtio-fs.
@@ -409,10 +411,6 @@ Features:
   - sd-stub: automatically pick up microcode from ESP (/loader/microcode/*)
     and synthesize initrd from it, and measure it. Signing is not necessary, as
     microcode does that on its own. Pass as first initrd to kernel.
-  - sd-stub should measure the kernel/initrd/… into a separate PCR, so that we
-    have one PCR we can bind the encrypted creds to that is not effected by
-    anything else but what we drop in via kernel-install, i.e. by earlier EFI
-    code running (i.e. like PCR 4)
 
 * Add a new service type very similar to Type=notify, that goes one step
   further and extends the protocol to cover reloads. Specifically, SIGHUP will
@@ -656,7 +654,7 @@ Features:
   dep in the base OS image)
 
 * sysext: automatically activate sysext images dropped in via new sd-stub
-  sysext pickup logic.
+  sysext pickup logic. (must insist on verity + signature on those though)
 
 * add concept for "exitrd" as inverse of "initrd", that we can transition to at
   shutdown, and has similar security semantics. This should then take the place
@@ -704,9 +702,9 @@ Features:
   what must be read-only, what requires encryption, and what requires
   authentication.
 
-* in uefi stub: query firmware regarding which PCRs are being used, store that
-  in EFI var. then use this when enrolling TPM2 in cryptsetup to verify that
-  the selected PCRs actually are used by firmware.
+* in uefi stub: query firmware regarding which PCR banks are being used, store
+  that in EFI var. then use this when enrolling TPM2 in cryptsetup to verify
+  that the selected PCRs actually are used by firmware.
 
 * rework recursive read-only remount to use new mount API
 
@@ -1602,7 +1600,6 @@ Features:
   - show whether UEFI audit mode is available
   - teach it to prepare an ESP wholesale, i.e. with mkfs.vfat invocation
   - teach it to copy in unified kernel images and maybe type #1 boot loader spec entries from host
-  - make it operate on loopback files, dissecting enough to find ESP to operate on
   - bootspec: properly support boot attempt counters when parsing entry file names
 
 * kernel-install:


poettering:
@Foxboron, @grawity This might be interesting to you?

(i.e. the systemd-measure tool is distantly related to https://github.com/grawity/tpm_futurepcr)

Foxboron:
Ah, this is lovely!

It would be nicer if we could also implement self selected PCR values in `systemd-cryptsetup` or in the future have some way to integrate between them. But this can come later. See https://github.com/systemd/systemd/issues/19204

My only concern is that the Authenticode checksum of the PE binary is tedious to get right. With signatures we need to ensure it's padded to 8 bytes, but without a signature it should not be padded. Should test this with a signed kernel and an unsiged kernel just to be sure.

poettering:
> It would be nicer if we could also implement self selected PCR values in `systemd-cryptsetup` or in the future have some way to integrate between them. But this can come later. See #19204

Would be happy to match a patch for that. maybe we should extend the syntax for PCR policies like this:

0+4:sha1=123456789abcdef…+4:sha256=123456…+9@signed-pcr-hash.p7+11=/some/file.bin

i.e. this would encode:

* PCR 0 should be bound to hash values taken from the running system (i.e. the current syntax)
* PCR 4 should be bound to hash values, with two for sha1 and sha256 banks specified literally (i.e. syntax is `<nr>:<bank>=<value`)
* PCR 9 is defined through signature of PCR hashes (syntax is `<nr>@<signaturefile>`)
* PCR 11 is defined through blob. The specified is file is opened, and hashed for all relevant banks, and then the result used.

poettering:
(but anyway, that discussion should probably followed elsewhere)

bluca:
```
../src/boot/efi/stub.c: In function ‘efi_main’:
../src/boot/efi/stub.c:211:57: error: pointer targets in passing argument 1 of ‘strlena’ differ in signedness [-Werror=pointer-sign]
                               strlena(unified_sections[section]) + 1, /* including NUL byte */
      |                                         ~~~~~~~~~~~~~~~~^~~~~~~~~
      |                                                         |
      |                                                         const char *
In file included from ../src/boot/efi/stub.c:4:
/usr/include/efi/efilib.h:396:24: note: expected ‘const CHAR8 *’ {aka ‘const unsigned char *’} but argument is of type ‘const char *’
  396 |     IN CONST CHAR8    *s1
```

src/boot/efi/measure.c
```diff
@@ -141,43 +141,79 @@ bool tpm_present(void) {
...
+
+        return ret;
 }
 
-EFI_STATUS tpm_log_load_options(const char16_t *load_options) {
+EFI_STATUS tpm_log_event_ascii(uint32_t pcrindex, EFI_PHYSICAL_ADDRESS buffer, UINTN buffer_size, const char *description, bool *ret_measured) {
+        _cleanup_free_ char16_t *c = NULL;
+
+        if (description)
+                c = xstra_to_str(description);
```
bluca, line 190: oom check?

man/systemd-stub.xml
```diff
@@ -239,6 +239,14 @@
         <citerefentry><refentrytitle>bootctl</refentrytitle><manvolnum>1</manvolnum></citerefentry> to view
         this data.</para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>StubPcrKernelParameters</varname></term>
+
+        <listitem><para>The PCR register index the kernel command line and credentials are measured into,
+        formatted as decimal ASCII string (i.e. <literal>12</literal>). This variable if a measurement was
```
bluca: "This variable is set if..."

man/systemd-stub.xml
```diff
@@ -240,6 +247,15 @@
         this data.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>StubPcrKernelImage</varname></term>
+
+        <listitem><para>The PCR register index the ELF kernel image/initial RAM disk image/boot
+        splash/devicetree database/embedded command line are measured into, formatted as decimal ASCII string
+        (i.e. <literal>11</literal>). This variable if a measurement was successfully completed, and remains
```
bluca: "This variable is set if..."

src/boot/efi/cpio.c
```diff
@@ -432,24 +434,40 @@ EFI_STATUS pack_cpio(
...
-                                NULL);
-                if (err != EFI_SUCCESS)
+                                &m);
+                if (err != EFI_SUCCESS) {
                         log_error_stall(L"Unable to add initrd TPM measurement for PCR %u (%s), ignoring: %r", tpm_pcr[i], tpm_description, err);
+                        measured = false;
+                        continue;
+                }
+
+                measured = measured < 0 ? m : measured && m;
```
bluca: `(measured && m)`?

src/boot/efi/measure.c
```diff
@@ -141,43 +141,79 @@ bool tpm_present(void) {
...
-                                    POINTER_TO_PHYSICAL_ADDRESS(load_options),
-                                    strsize16(load_options), load_options);
+                if (pcr == UINT32_MAX) /* Skip this one, if it's invalid, so that our 'measured' return value is not corrupted by it */
+                        continue;
+
+                err = tpm_log_event(pcr, POINTER_TO_PHYSICAL_ADDRESS(load_options), strsize16(load_options), load_options, &m);
                 if (err != EFI_SUCCESS)
                         return log_error_status_stall(err, L"Unable to add load options (i.e. kernel command) line measurement to PCR %u: %r", pcr, err);
+
+                measured = measured < 0 ? m : measured && m;
```
bluca: `(measured && m)`?

man/systemd-stub.xml
```diff
@@ -263,6 +263,15 @@
         formatted as decimal ASCII string (i.e. <literal>12</literal>). This variable if a measurement was
         successfully completed, and remains unset otherwise.</para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>StubPcrInitRDSysExts</varname></term>
+
+        <listitem><para>The PCR register index the systemd extensions for the initial RAM disk image, which
+        are picked up from the file system the kernel image is located on.  Formatted as decimal ASCII string
+        (i.e. <literal>13</literal>). This variable if a measurement was successfully completed, and remains
```
bluca: "This variable is set if..."

src/fundamental/tpm-pcr.h
```diff
@@ -22,3 +22,17 @@
 
 /* This TPM PCR is where we extend the initrd sysext images into which we pass to the booted kernel */
 #define TPM_PCR_INDEX_INITRD_SYSEXTS 13U
+
+/* List of PE sections that have special meaning for us in unified kernels. This is the canonical order in
+ * which we measure the sections into PCR 11 (see above)ye. PLEASE DO NOT REORDER! */
```
bluca: /above)ye/above)/

src/boot/efi/util.c
```diff
@@ -743,3 +743,22 @@ EFI_STATUS make_file_device_path(EFI_HANDLE device, const char16_t *file, EFI_DE
         SetDevicePathEndNode(dp);
         return EFI_SUCCESS;
 }
+
+void hexdump(const char16_t *prefix, const void *data, UINTN size) {
+        static const char hex[16] = "0123456789abcdef";
+        _cleanup_free_ char16_t *buf = NULL;
+        const uint8_t *d = data;
+
```
bluca: assert on prefix and data

src/boot/bootctl.c
```diff
@@ -613,22 +621,26 @@ static int status_variables(void) {
...
-                print_efi_option(options[i], false);
+                (void) print_efi_option(options[i], &n_printed, /* in_order= */ false);
 
         next_option:
                 continue;
         }
 
+        if (n_printed == 0) {
+                printf("No boot loaders listed in EFI Variables.\n\n");
+                return 0;
```
bluca: uneeded return

bluca (changes requested):
Superficial review, but looks great. How about a --json output switch for the measure tool, to facilitate further machine processing?

src/boot/efi/measure.c
poettering: in our efi code we have allocators that abort, given that we can't possibly recover from the inability to boot, given we are the only code running at that time, pretty much. The `x` in the name of `xstra_to_str()` is supposed to indicate that.

bluca: ah, got it

src/boot/efi/measure.c
```diff
@@ -141,43 +141,79 @@ bool tpm_present(void) {
...
-        EFI_TCG *tpm1;
+EFI_STATUS tpm_log_event(uint32_t pcrindex, EFI_PHYSICAL_ADDRESS buffer, UINTN buffer_size, const char16_t *description, bool *ret_measured) {
         EFI_TCG2 *tpm2;
+        EFI_STATUS ret;
 
         assert(description);
 
-        /* PCR disabled */
-        if (pcrindex == UINT32_MAX)
+        /* If EFI_SUCCESS is returned, will initialize ret_measured to true if we actually measured
```
medhefgo, line 151: This is just overcomplicating things. Why not just return EFI_NOT_FOUND and handle that where it's needed?

src/boot/efi/measure.c
```diff
@@ -141,43 +141,79 @@ bool tpm_present(void) {
...
+        }
+
+        if (ret == EFI_SUCCESS && ret_measured)
+                *ret_measured = true;
+
+        return ret;
 }
 
-EFI_STATUS tpm_log_load_options(const char16_t *load_options) {
+EFI_STATUS tpm_log_event_ascii(uint32_t pcrindex, EFI_PHYSICAL_ADDRESS buffer, UINTN buffer_size, const char *description, bool *ret_measured) {
```
medhefgo, line 186: Actually, the description can be any arbitrary binary data if I look at the spec correctly, so there really is no need to convert to char16_t. In fact it's probably nicer for userspace if we always log with regular char strings…

src/boot/efi/stub.c
```diff
@@ -204,6 +204,41 @@ EFI_STATUS efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *sys_table) {
...
+                m = false;
+
+                if (szs[section] == 0) /* not found */
+                        continue;
+
+                /* First measure the name of the section */
+                (void) tpm_log_event_ascii(
+                                TPM_PCR_INDEX_KERNEL_IMAGE,
+                                POINTER_TO_PHYSICAL_ADDRESS(sections[section]),
+                                strlena(sections[section]) + 1, /* including NUL byte */
```
medhefgo:
```suggestion
                                strsize8(sections[section]),
```

src/fundamental/tpm-pcr.h
```diff
@@ -0,0 +1,24 @@
...
+/* sd-stub used to write the kernel command line/credentials into PCR 8, in systemd <= 250. Let's provide for
+ * some compatibility. (Remove in 2023!) */
+#if EFI_TPM_PCR_COMPAT
+#define TPM_PCR_INDEX_KERNEL_PARAMETERS_COMPAT 8U
+#else
+#define TPM_PCR_INDEX_KERNEL_PARAMETERS_COMPAT UINT32_MAX
+#endif
+
+/* This TPM PCR is where we extend the initrd sysext images into which we pass to the booted kernel */
+#define TPM_PCR_INDEX_INITRD_SYSEXTS 13U
```
medhefgo, line 24: As an enum-enjoyer I have to ask: can we make these an enum? :D

src/fundamental/tpm-pcr.c
```diff
@@ -0,0 +1,14 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+
+#include "macro-fundamental.h"
```
medhefgo: What you need this for?

src/boot/efi/stub.c
```diff
@@ -103,6 +103,13 @@ static EFI_STATUS combine_initrd(
 }
 
 static void export_variables(EFI_LOADED_IMAGE_PROTOCOL *loaded_image) {
+        static const uint64_t stub_features =
+                EFI_STUB_FEATURE_REPORT_BOOT_PARTITION |    /* We set LoaderDevicePartUUID */
+                EFI_STUB_FEATURE_PICK_UP_CREDENTIALS |      /* We pick up credentials from the boot partition */
+                EFI_STUB_FEATURE_PICK_UP_SYSEXTS |          /* We pick up system extensions from the boot partition */
```
medhefgo, line 109: If you add feature flags for these, the dtb feature is recent too, no?

src/fundamental/efivars-fundamental.h
```diff
@@ -13,6 +14,12 @@
 #define EFI_LOADER_FEATURE_RANDOM_SEED             (UINT64_C(1) << 6)
 #define EFI_LOADER_FEATURE_LOAD_DRIVER             (UINT64_C(1) << 7)
 
+/* Features of the stub, i.e. systemd-stub */
+#define EFI_STUB_FEATURE_REPORT_BOOT_PARTITION     (UINT64_C(1) << 0)
+#define EFI_STUB_FEATURE_PICK_UP_CREDENTIALS       (UINT64_C(1) << 1)
+#define EFI_STUB_FEATURE_PICK_UP_SYSEXTS           (UINT64_C(1) << 2)
+#define EFI_STUB_FEATURE_THREE_PCRS                (UINT64_C(1) << 3)
```
medhefgo, line 24: `enum` feels left out :(

src/boot/efi/util.c
```diff
@@ -743,3 +743,22 @@ EFI_STATUS make_file_device_path(EFI_HANDLE device, const char16_t *file, EFI_DE
...
+        buf = xnew(char16_t, size*2+1);
+
+        for (UINTN i = 0; i < size; i++) {
+                buf[i*2] = hex[d[i] >> 4];
+                buf[i*2+1] = hex[d[i] & 0x0F];
+        }
+
+        buf[size*2] = 0;
+
+        log_error_stall(L"%s[%u]: %s", prefix, size, buf);
```
medhefgo: Needs `PRIuN` for size

man/systemd-cryptenroll.xml
```diff
@@ -296,11 +296,21 @@
                 <entry>The IMA project measures its runtime state into this PCR.</entry>
               </row>
 
+              <row>
+                <entry>11</entry>
+                <entry><citerefentry><refentrytitle>systemd-stub</refentrytitle><manvolnum>7</manvolnum></citerefentry> measures the ELF kernel image, embedded initrd and other payload of the PE image it is placed in in this PCR. Unlike PCR 4 (where the same data should be measured into), this PCR value should be easy to pre-calculate, as this only contains static parts of the PE binary. Use this PCR to bind TPM policies to a specific kernel image, possibly with an embedded initial RAM disk (initrd).</entry>
```
behrmann:
```suggestion
                <entry><citerefentry><refentrytitle>systemd-stub</refentrytitle><manvolnum>7</manvolnum></citerefentry> measures the ELF kernel image, embedded initrd and other payload of the PE image it is placed in into this PCR. Unlike PCR 4 (where the same data should be measured into), this PCR value should be easy to pre-calculate, as this only contains static parts of the PE binary. Use this PCR to bind TPM policies to a specific kernel image, possibly with an embedded initial RAM disk (initrd).</entry>
```

man/systemd-measure.xml
```diff
@@ -0,0 +1,154 @@
...
+    <citerefentry><refentrytitle>os-release</refentrytitle><manvolnum>5</manvolnum></citerefentry> file, and
+    boot splash file that make up the unified kernel image, and determines the PCR values expected to be in
+    place after booting the image. Calculation starts with a zero-initialized PCR 11, and is executed in a
+    fashion compatible with what <filename>systemd-stub</filename> does at boot.</para>
+  </refsect1>
+
+  <refsect1>
+    <title>Commands</title>
+
+    <para>The follow commands are understood:</para>
```
behrmann:
```suggestion
    <para>The following commands are understood:</para>
```

man/systemd-measure.xml
```diff
@@ -0,0 +1,154 @@
...
+      <xi:include href="standard-options.xml" xpointer="help" />
+      <xi:include href="standard-options.xml" xpointer="version" />
+    </variablelist>
+  </refsect1>
+
+  <refsect1>
+    <title>Examples</title>
+
+    <example>
+      <title>Generate a unified kernel image, and calculating the expected TPM PCR 11 value</title>
```
behrmann:
```suggestion
      <title>Generate a unified kernel image, and calculate the expected TPM PCR 11 value</title>
```
or
```suggestion
      <title>Generating a unified kernel image, and calculating the expected TPM PCR 11 value</title>
```

man/systemd-stub.xml
```diff
@@ -133,8 +140,8 @@
     core kernel, the embedded initrd and kernel command line (see above for a full list).</para>
 
     <para>Also note that the Linux kernel will measure all initrds it receives into TPM PCR 9. This means
-    every type of initrd will be measured twice: the initrd embedded in the kernel image will be measured to
-    both PCR 4 and PCR 9; the initrd synthesized from credentials will be measured to both PCR 12 and PCR 9;
+    every type of initrd will be measured twice or thrice: the initrd embedded in the kernel image will be measured to
```
behrmann: While I am a great fan of thrice, it is rather uncommon in most English vernaculars, at least if the intention is not to sound like some ancient tome of forbidden knowledge. :)
```suggestion
    every type of initrd will be measured twice or three times: the initrd embedded in the kernel image will be measured to
```
or
```suggestion
    every type of initrd will be measured two or three times: the initrd embedded in the kernel image will be measured to
```

bluca:
Some tests would also be good

src/fundamental/tpm-pcr.c
poettering: it was an easy way to get the definition of NULL. will change to stddef.h

src/boot/efi/stub.c
poettering: the feature flags are supposed to indicate stuff to userspace that it can make programmatical use of. i.e. it's more about reporting OS facing *API* features of the stub than just features of the stub that face elsewhere. I am not sure how an app running on the OS would take benefit of the dtb loading stuff, given the only way to get the dtb loaded is by linking it in via objcopy... it's a kernel builder facing feature, sure. But userspace API? i don't see it.

src/boot/efi/measure.c
poettering: makes calling code simpler. i.e. the idea is that pcrindex == UINT32_MAX magically disables the measurement without this need special handling in all callers...

i.e. conceptually if we want something not to be treated as an error, we shouldn't return it as an error. because that forces calles to manually handle the errors manually even for clean codepaths, which is doable, but not ideal I think.

i.e. I am generally of the opinion that it's better to hide conditionalizations inside the functions to be conditionalized, instead of doing that in the caller. it's a scheme of "cleaner abstractions", as it allows you to see the general codeflow easier in the calling process, and not get lost in details about specific exceptional conditions...

Does that make any sense?

src/boot/efi/measure.c
poettering: well, the efi tpm event log can be browsed with various tools, and they generally assume it contains a regular uefi string, i.e. utf16... the main reason for the decription is to make debugging easier, it is afte rall not part of the data that is measured. Hence it's helpful to stick to the format most events use, and that appears to be UTF-16...

poettering:
> Some tests would also be good

You are absolutely right, but this is a hard one, because we currently have no testsuite for sd-boot/sd-stub at all. We should though, but I am not sure this is the right PR to add that.

So, I am all for adding that, but let's maybe not block this PR for that.

bluca:
> > Some tests would also be good
> 
> You are absolutely right, but this is a hard one, because we currently have no testsuite for sd-boot/sd-stub at all. We should though, but I am not sure this is the right PR to add that.
> 
> So, I am all for adding that, but let's maybe not block this PR for that.

Maybe start with the `systemd-measure calculate` output? Some simple checks that say with an os-release and cmdline it spits out the expected values

poettering:
Force pushed a new version. Addressed the points raised. But not the enum-ification. I am not against converting the flags to an enum, but given the boot loader flags (which the stub flags take inspiration from) are macros right now, sounds like a clean-up to do separately, and then convert the whole thing.

Also, no test yet.

poettering:
Added test now, too.

bluca:
```
../src/src/boot/measure.c: In function ‘measure_pcr’:
../src/src/boot/measure.c:262:30: error: implicit declaration of function ‘read’; did you mean ‘readv’? [-Werror=implicit-function-declaration]
  262 |                         sz = read(fd, buffer, BUFFER_SIZE);
      |                              ^~~~
      |                              readv
../src/src/boot/measure.c:262:30: warning: nested extern declaration of ‘read’ [-Wnested-externs]
../src/src/boot/measure.c: In function ‘validate_stub’:
../src/src/boot/measure.c:429:21: error: implicit declaration of function ‘access’; did you mean ‘accept’? [-Werror=implicit-function-declaration]
  429 |                 if (access(p, F_OK) < 0) {
      |                     ^~~~~~
      |                     accept
../src/src/boot/measure.c:429:21: warning: nested extern declaration of ‘access’ [-Wnested-externs]
```

poettering:
Force pushed a new version. PTAL.

poettering:
(only fixes CI issues)

bluca:
> Superficial review, but looks great. How about a --json output switch for the measure tool, to facilitate further machine processing?

What about this?

poettering:
> > Superficial review, but looks great. How about a --json output switch for the measure tool, to facilitate further machine processing?
> 
> What about this?

Not against this. But maybe let's add this later. At this point, the whole tool is marked experimental in the docs, because I am not sure about the precise output. 

In particular I intend to add signing support, which makes the JSON stuff particularly interesting, since we need to output more complex structured data then. But I'd rather figure that out then, then doing half-assed JSON now that might need to be reworked later.

JSON means setting things in stone as an API, the current simply text based output much less so. Hence, I think JSON is what we should add once the main feature set we intend to provide here is in place and we can oversee what the API is going to be like.

poettering:
the CI failures appear unrelated.

src/boot/efi/measure.c
```diff
@@ -141,43 +142,79 @@ bool tpm_present(void) {
         return tcg2_interface_check() || tcg1_interface_check();
 }
 
-EFI_STATUS tpm_log_event(uint32_t pcrindex, EFI_PHYSICAL_ADDRESS buffer, UINTN buffer_size, const char16_t *description) {
-        EFI_TCG *tpm1;
+EFI_STATUS tpm_log_event(uint32_t pcrindex, EFI_PHYSICAL_ADDRESS buffer, UINTN buffer_size, const char16_t *description, bool *ret_measured) {
         EFI_TCG2 *tpm2;
+        EFI_STATUS ret;
```
medhefgo: Please rename to `err` for consistency everywhere.

src/boot/efi/util.c
```diff
@@ -743,3 +743,25 @@ EFI_STATUS make_file_device_path(EFI_HANDLE device, const char16_t *file, EFI_DE
...
+
+void hexdump(const char16_t *prefix, const void *data, UINTN size) {
+        static const char hex[16] = "0123456789abcdef";
+        _cleanup_free_ char16_t *buf = NULL;
+        const uint8_t *d = data;
+
+        assert(prefix);
+        assert(data || size == 0);
+
+        /* Debugging helper — please keep this around, even if not used */
```
medhefgo: This since is for debugging, this should be behind `#ifdef EFI_DEBUG` and maybe be defined close to `debug_break`.

src/boot/efi/measure.c
medhefgo: I maintain that this makes all the code being touched ugly. Doing a `IN_SET(measure(...), EFI_SUCCESS, EFI_NOT_FOUND)` is imho much nicer to read than the `bool` dance you're doing and works similar to how `open()` works in userspace.

You also don't have to use `EFI_NOT_FOUND`. You can define a custom warning code and just use that. `EFI_STATUS` explicitly makes anything that doesn't have the MSB set a warning, which fits perfectly for this. In that case using `EFI_ERROR` to only filter for errors would even makes sense and keep the consumer side clean.

You should at least make the ret bool param mandatory where it makes sense (`pack_cpio` for example) to keep things simpler.

src/boot/efi/cpio.c
```diff
@@ -432,23 +434,40 @@ EFI_STATUS pack_cpio(
                 return log_error_status_stall(err, L"Failed to pack cpio trailer: %r");
 
         for (UINTN i = 0; i < n_tpm_pcr; i++) {
+                bool m;
+
+                if (tpm_pcr[i] == UINT32_MAX) /* Disabled */
```
medhefgo, line 439: This is already done by tpm_log_event, why replicate?

bluca (approved):

src/boot/efi/measure.c
poettering: Also note that there's stuff such as pack_cpio() that will set a `ret_measured`. Here in particular it becomes clear why the fact whether we actually measured something should not be propagated as an error code — it is not an *error* after all — but as a side-parameter. After all, the cpio packing should not abort just because the PCR measurement doesn't work.

Thats conceptually at the core of the problem: failure generally implies abortion of the regular execution paths of functions. And bypassing the PCR measurements should not do that. It should have almost no effect..

src/boot/efi/cpio.c
poettering: if we measure multiple things to a PCR and one measurement isn't done we want to treat that as incompletely measured, i.e. set ret_measured = false. But if we measure the same thing to multiple PCRs, and one of that isn't done because explicitly disabled, then we still want to consider the measurement as having taken place.

i.e. measurement to multiple PCRs is redundant, and it's fine if one of them is disabled.

measurement of multiple blocks where one is skipped is *not* redundant though, and we should handle that differently.

poettering:
> You should at least make the ret bool param mandatory where it makes sense (pack_cpio for example) to keep things simpler.

it shouldn't be mandatory though. I intend to soon extend the whole logic so that we can embed signed PCR hash values in the PE image, and those should not be measured (because that would be cycle then), but still be packed up in a cpio.

(i.e. my goal here is that now that we have PCR hashes that can be pre-calculate people can use them to implement signature based PCR hash schemes, so that kernel images can be updated easily without this requiring re-enrollment of the hashes. These signatures should be included in the PE image, but unlike the rest of the sections which should be measured into PCRs, this signature section shouldn't be, because it already contains the expected results of the measurements, and making it input for the measurements is hence a bit useless)

Or in other words unified kernels would be built like this:

1. We prepare the components: kernel, initrd, boot splash, dtb, …
2. We pre-calculate the PCR hashes for these components
3. We sign the PCR hashes with a priv/pub key pair of our choosing
4. We now build the PE image from stub, from the mentioned components, and the signed expected hash values.
5. We sign the whole thing with SecureBoot certificates

On boot, the hash/sig data is taken by sd-stub, tunred into an initrd, and passed to the booted kernel. There it can be consumed by cryptsetup for example, to unlock a disk.

Net result: you can easily set up encrypted disks that can only be decrypted if a very specific series of kernels are booted. This is nice since SecureBoot certificates are typically excessively wide in focus (i.e, anyone who asks nicely will get a kernel/boot loader signed by MSFT), but with this we can tightly control with a very narrow focus which kernels actually get access to a disk.

And that's excellent for "opportunistic encryption", i.e. systems which are auto-encrypted by default, and whose keys are locked to TPM only and thus can be unlocked non-interactively at boot.

poettering:
Force pushed new version. Addressed some of @medhefgo's points, but not the `ret_measure` thing, nor the `enum` thing. (I disagree with the former, and the latter sounds good to me, but in a later commit, since it should probably converted together with other stuff not directly related to this PR)

poettering:
CI failures appear unrelated...

PR #24156, opened 2022-07-30
login: fixlets for state serialization

yuwata:


src/login/logind.c
```diff
@@ -1119,7 +1119,8 @@ static int manager_startup(Manager *m) {
                 (void) seat_start(seat);
 
         HASHMAP_FOREACH(user, m->users)
-                (void) user_start(user);
+                if (!user->stopping)
+                        (void) user_start(user);
```
poettering, line 1123: hmm, this change would mean that if you restart logind, before going down the user's state file would exist with 'stopping' set in it, and the `started` boolean would be set for the user. But when coming back from serialization, after restoring all state, we'd have 'stopping' set, but not 'started' anymore, because you skip `user_start()` here. So if anything then ends up callng `user_save()` the state file will be gone, and state is lost, and afte ranother restart cycle the user would just go away entirely.

This doesn't look right to me?

yuwata:
(I tried to fix an issue, but I cannot reproduce it anymore. Closing.)

PR #24162, opened 2022-07-31
gcrypt: switch to system rng before gcry_check_version

crrodriguez:
Current documentation claims this must be done before gcry_check_version.

¯\_(ツ)_/¯


yuwata (approved):
LGTM.

PR #24164, opened 2022-08-01
bpf: fix is_allow_list section

jameshilliard:
The llvm bpf compiler appears to place const volatile variables in a non-standard section which creates an incompatibility with the gcc bpf compiler.

To fix this force GCC to also use the rodata section.

Note this does emit an assembler warning:
```c
Generating src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.unstripped.o with a custom command
/tmp/ccM2b7jP.s: Assembler messages:
/tmp/ccM2b7jP.s:87: Warning: setting incorrect section attributes for .rodata
```

See:
https://github.com/llvm/llvm-project/issues/56468

Fixes:
```c
../src/core/restrict-ifaces.c:45:14: error: ‘struct
restrict_ifaces_bpf’ has no member named ‘rodata’; did you mean
‘data’?
   45 |         obj->rodata->is_allow_list = is_allow_list;
      |              ^~~~~~
      |              data
```

### DIFF
--- a/src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
+++ b/src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
@@ -6,7 +6,7 @@
 #include <linux/bpf.h>
 #include <bpf/bpf_helpers.h>
 
-const volatile __u8 is_allow_list = 0;
+const volatile __u8 is_allow_list SEC(".rodata") = 0;
 
 /* Map containing the network interfaces indexes.
  * The interpretation of the map depends on the value of is_allow_list.


yuwata (approved):
Not familiar, but superficially LGTM.

anakryiko:
The point of using const is to put variable in .rodata so that BPF verifier knows the value of this variable during verification. If you want non-read-only variable, just drop the `const`.

jameshilliard:
> The point of using const is to put variable in .rodata so that BPF verifier knows the value of this variable during verification.

So if one does want the variable in `.rodata` it should be declared with `const` and without `volatile` then?

For example declaring like this:
```c
volatile __u8 is_allow_list = 0;
```

Gives the following generated skeleton struct with gcc-bpf:
```c
struct restrict_ifaces_bpf {
	struct bpf_object_skeleton *skeleton;
	struct bpf_object *obj;
	struct {
		struct bpf_map *sd_restrictif;
		struct bpf_map *bss;
	} maps;
	struct {
		struct bpf_program *sd_restrictif_e;
		struct bpf_program *sd_restrictif_i;
	} progs;
	struct {
		struct bpf_link *sd_restrictif_e;
		struct bpf_link *sd_restrictif_i;
	} links;
	struct restrict_ifaces_bpf__bss {
		__u8 is_allow_list;
	} *bss;

#ifdef __cplusplus
	static inline struct restrict_ifaces_bpf *open(const struct bpf_object_open_opts *opts = nullptr);
	static inline struct restrict_ifaces_bpf *open_and_load();
	static inline int load(struct restrict_ifaces_bpf *skel);
	static inline int attach(struct restrict_ifaces_bpf *skel);
	static inline void detach(struct restrict_ifaces_bpf *skel);
	static inline void destroy(struct restrict_ifaces_bpf *skel);
	static inline const void *elf_bytes(size_t *sz);
#endif /* __cplusplus */
};
```

If I declare with `const` and without `volatile` like this:
```c
const __u8 is_allow_list = 0;
```

I get the following generated skeleton struct with gcc-bpf:
```c
struct restrict_ifaces_bpf {
	struct bpf_object_skeleton *skeleton;
	struct bpf_object *obj;
	struct {
		struct bpf_map *sd_restrictif;
		struct bpf_map *rodata;
	} maps;
	struct {
		struct bpf_program *sd_restrictif_e;
		struct bpf_program *sd_restrictif_i;
	} progs;
	struct {
		struct bpf_link *sd_restrictif_e;
		struct bpf_link *sd_restrictif_i;
	} links;
	struct restrict_ifaces_bpf__rodata {
		__u8 is_allow_list;
	} *rodata;

#ifdef __cplusplus
	static inline struct restrict_ifaces_bpf *open(const struct bpf_object_open_opts *opts = nullptr);
	static inline struct restrict_ifaces_bpf *open_and_load();
	static inline int load(struct restrict_ifaces_bpf *skel);
	static inline int attach(struct restrict_ifaces_bpf *skel);
	static inline void detach(struct restrict_ifaces_bpf *skel);
	static inline void destroy(struct restrict_ifaces_bpf *skel);
	static inline const void *elf_bytes(size_t *sz);
#endif /* __cplusplus */
};
```

For reference the existing declaration:
```c
const volatile __u8 is_allow_list = 0;
```

Gives the following generated skeleton struct with gcc-bpf:
```c
struct restrict_ifaces_bpf {
	struct bpf_object_skeleton *skeleton;
	struct bpf_object *obj;
	struct {
		struct bpf_map *sd_restrictif;
		struct bpf_map *data;
	} maps;
	struct {
		struct bpf_program *sd_restrictif_e;
		struct bpf_program *sd_restrictif_i;
	} progs;
	struct {
		struct bpf_link *sd_restrictif_e;
		struct bpf_link *sd_restrictif_i;
	} links;
	struct restrict_ifaces_bpf__data {
		__u8 is_allow_list;
	} *data;

#ifdef __cplusplus
	static inline struct restrict_ifaces_bpf *open(const struct bpf_object_open_opts *opts = nullptr);
	static inline struct restrict_ifaces_bpf *open_and_load();
	static inline int load(struct restrict_ifaces_bpf *skel);
	static inline int attach(struct restrict_ifaces_bpf *skel);
	static inline void detach(struct restrict_ifaces_bpf *skel);
	static inline void destroy(struct restrict_ifaces_bpf *skel);
	static inline const void *elf_bytes(size_t *sz);
#endif /* __cplusplus */
};
```

This is different from llvm and based on my understanding of [#56468](https://github.com/llvm/llvm-project/issues/56468) the gcc version is correct while llvm incorrectly puts `is_allow_list` in `rodata`.

The example in the docs I'm reading [here](https://facebookmicrosites.github.io/bpf/blog/2020/02/20/bcc-to-libbpf-howto-guide.html#application-configuration) appear to suggest it's expected that `const` `volatile` variables should end up in `rodata` even though that [appears to be incorrect](https://github.com/llvm/llvm-project/issues/56468). Should the example there be using `const` without `volatile` instead?

> If you want non-read-only variable, just drop the `const`.

Is that what we want in this case?

src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
```diff
@@ -6,7 +6,7 @@
 #include <linux/bpf.h>
 #include <bpf/bpf_helpers.h>
 
-const volatile __u8 is_allow_list = 0;
+const __u8 is_allow_list SEC(".rodata") = 0;
```
jameshilliard: The `SEC` macro also sets the [`used`](https://clang.llvm.org/docs/AttributeReference.html#used) attribute, is my understanding correct that this accomplishes the same purpose as `volatile` in regards to preventing the variable from being optimized away?

anakryiko:
> > The point of using const is to put variable in .rodata so that BPF verifier knows the value of this variable during verification.
> 
> So if one does want the variable in `.rodata` it should be declared with `const` and without `volatile` then?
> 

In such case compiler will assume it knows the final value of such variable and might just inline it in the code. But in reality this value can be changed and specified from user-space before BPF program is loaded. That's why we are specifying volatile, to avoid compiler assumptions about specific value at compilation time.

> 
> This is different from llvm and based on my understanding of [#56468](https://github.com/llvm/llvm-project/issues/56468) the gcc version is correct while llvm incorrectly puts `is_allow_list` in `rodata`.
> 
> The example in the docs I'm reading [here](https://facebookmicrosites.github.io/bpf/blog/2020/02/20/bcc-to-libbpf-howto-guide.html#application-configuration) appear to suggest it's expected that `const` `volatile` variables should end up in `rodata` even though that [appears to be incorrect](https://github.com/llvm/llvm-project/issues/56468). Should the example there be using `const` without `volatile` instead?
> 

I'm not language expert and not on C standards committee, but to me volatile is just a means to indicate that underlying memory might be changed through some means that compiler doesn't know about. The `const` part means it's read-only and thus it's logical that it maps to .rodata with read-only access.

> > If you want non-read-only variable, just drop the `const`.
> 
> Is that what we want in this case?

In this specific example it doesn't matter as we don't guard any BPF helper or some other kernel BPF feature that might not be unavailable on some kernel versions. But in general, .rodata and const volatile are very important for CO-RE and allows to have the code that will work both on old and new kernels and on new kernels will be able to utilize features that are not necessarily supported by old kernel. E.g., if there is some BPF helper that we want to use if it is supported, by guarding it with read-only variable it is possible to load the same BPF object file on old kernel by just setting read-only variable to such a value that it proves to BPF verifier that unsupported BPF helper will never be called (through dead code elimination logic in BPF verifier).

src/core/bpf/restrict_ifaces/restrict-ifaces.bpf.c
anakryiko: no, used attribute just prevents from compiler complaining that there is no explicit usage for given variable. We rely on used attribute more for BPF program declarations (and something for map declarations), less so for global variables. But we utilize the same SEC() macro for all of them.

anitazha (approved):
Latest change keeping `const volatile` and using `SEC` macro are good

anakryiko:
I don't understand this change. Is @jameshilliard now going to add SEC(".rodata") in every open-source BPF code he could find just to please GCC's BPF backend instead of trying to make GCC's BPF backend conformant to what Clang's BPF backend is doing?

As I repeated multiple times, @jameshilliard, drop the `const volatile` part for this specific case to unblock yourself, but ultimate GCC has to put `const volatile` variables into `.rodata` section. Please stop modifying everything around GCC but not GCC itself.

If I could, I'd reject this change. This change is ugly and unnecessary.

jameshilliard:
> As I repeated multiple times, @jameshilliard, drop the `const volatile` part for this specific case to unblock yourself, but ultimately GCC has to put `const volatile` variables into `.rodata` section.

I've opened #24881 to revert this as this behavior divergence issue is now fixed.

PR #24166, opened 2022-08-01
bpf: set gcc std and compile flags

jameshilliard:
This should make gcc bpf compilation more reliable.

### DIFF
--- a/src/core/bpf/meson.build
+++ b/src/core/bpf/meson.build
@@ -5,6 +5,7 @@ if conf.get('BPF_FRAMEWORK') != 1
 endif
 
 bpf_clang_flags = [
+        '-std=gnu11',
         '-Wno-compare-distinct-pointer-types',
         '-O2',
         '-target',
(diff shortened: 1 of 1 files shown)


src/core/bpf/meson.build
```diff
@@ -14,11 +14,13 @@ bpf_clang_flags = [
 ]
 
 bpf_gcc_flags = [
+        '-std=gnu11',
```
yuwata, line 18: How about to also set the same option for clang in the above??

yuwata (approved):
LGTM. But one suggestion.

src/core/bpf/meson.build
jameshilliard: AFAIU clang doesn't need this, gcc sometimes needs it for `typeof` support to work, there's an alternate `__typeof__` style which works with regular c11 in gcc as well but bpf headers uses mostly the `typeof` style.

jameshilliard: hmm, actually looks like might be helpful in clang(I guess defaults work however)

jameshilliard: added to clang flags as well

yuwata: > AFAIU clang doesn't need this, gcc sometimes needs it for `typeof` support to work, there's an alternate `__typeof__` style which works with regular c11 in gcc as well but bpf headers uses mostly the `typeof` style.

Even so, I think it is better to add that for consistency, if adding the option for clang does not introduce any issue. Then, all codes are compiled with `gnu11`.

yuwata: > added to clang flags as well

Thanks.

PR #24175, opened 2022-08-02
meson: Use feature options

medhefgo:


medhefgo:
The detection behavior should be the same, but there might be slight differences (most likely for bpf-framework, cryptsetup and libidn).

@evverx  would be nice if you could tell me if this works wrt fuzzers/sanitizers

@eli-schwartz would be nice if you could give this a quick glance to see if I am using the features correctly/as intended. Also would love if you could convert features to strings and also do something like `get_option('a') + get_option('b')` to combine them.

medhefgo:
Obviously, this does create some friction/incompatibility as features for whatever reason weren't designed to take `true/false`. This is also why semaphore ci (for example) is complaining.

This would create stronger incentive for meson 0.60.0 as that allows remapping `true/false` to `enabled/disabled`.

evverx:
I just took a quick look because I'm pretty sure this PR is doomed (in the sense that it's going to be rejected even though technically with this PR applied it's still possible to build systemd on all the distributions covered by the policy).

> @evverx would be nice if you could tell me if this works wrt fuzzers/sanitizers

As far as I can see the "want-oss" option no longer automatically turns off the dependencies so to make it work on OSS-Fuzz either `-Dauto_features=disabled/auto` should be passed to `meson` in `tools/oss-fuzz.sh` or `meson.build` should flip it itself if possible. I don't think feature options should affect sanitizers so they should probably keep working.

> This would create stronger incentive for meson 0.60.0 as that allows remapping true/false to enabled/disabled

I don't think it's going to happen any time soon.

eli-schwartz:
> Also would love if you could convert features to strings (right now it somethings something like `<mesonbuild.coredata.UserFeatureOption object at 0x7fde5a597610>` which isn't helpful)

Do you want the option name, or stringify its state?

> and also do something like `get_option('a') + get_option('b')` to combine them.
For e.g.

```meson
if want_libcryptsetup_plugins.enabled()
        if want_libcryptsetup.disabled()
                error('libcryptsetup-plugins can not be requested without libcryptsetup')
        else
                want_libcryptsetup = want_libcryptsetup_plugins
        endif
endif
```

I think what you want is:
```meson
if want_libcryptsetup_plugins.require(
            want_libcryptsetup.allowed(),
            error_message: 'without libcryptsetup'
        ).enabled()
    want_libcryptsetup = want_libcryptsetup_plugins
endif
```

The error message reported will be:

```
Feature libcryptsetup-plugins cannot be enabled: without libcryptsetup
```

Note you need 0.59.0 for this, at which point you're pretty close to 0.60.0 anyway and can use:
```
option('o7', type: 'feature', value: 'enabled', deprecated: {'true': 'enabled', 'false': 'disabled'})
```

eli-schwartz (approved):
From a quick look this seems to be correct.

medhefgo:
@eli-schwartz thx for the quick look. It's not possible to change meson options from `meson.build` by any chance? :D

> Do you want the option name, or stringify its state?

It should return the state as (`auto/enabled/disabled`) analogue to bools.

> I think what you want is:
> 
> ```meson
> if want_libcryptsetup_plugins.require(
>             want_libcryptsetup.allowed(),
>             error_message: 'without libcryptsetup'
>         ).enabled()
>     want_libcryptsetup = want_libcryptsetup_plugins
> endif
> ```

Ah yes, that would work nicely. If only we could use it… :(

> As far as I can see the "want-oss" option no longer automatically turns off the dependencies so to make it work on OSS-Fuzz either `-Dauto_features=disabled/auto` should be passed to `meson` in `tools/oss-fuzz.sh` or `meson.build` should flip it itself if possible. I don't think feature options should affect sanitizers so they should probably keep working.

@evverx  You're right, that eluded me somehow. Fixed.

eli-schwartz:
> It's not possible to change meson options from `meson.build` by any chance? :D

This is mentioned in https://mesonbuild.com/Contributing.html#no-lingering-state

evverx:
@medhefgo sorry I should have been a bit more specific about what I meant by passing `-Dauto_features=disabled/auto`. The script supports another mode where the dependencies should be auto-detected so if it's on `auto_features` should be set to `auto`. I think something like
```diff
diff --git a/tools/oss-fuzz.sh b/tools/oss-fuzz.sh
index 351e53321d..8db221c12a 100755
--- a/tools/oss-fuzz.sh
+++ b/tools/oss-fuzz.sh
@@ -29,8 +29,10 @@ mkdir -p "$build"

 if [ -z "$FUZZING_ENGINE" ]; then
     fuzzflag="llvm-fuzz=true"
+    auto_features="auto"
 else
     fuzzflag="oss-fuzz=true"
+    auto_features="disabled"

     apt-get update
     apt-get install -y gperf m4 gettext python3-pip \
@@ -73,7 +75,7 @@ else
     fi
 fi

-if ! meson "$build" "-D$fuzzflag" -Db_lundef=false -Dauto_features=disabled; then
+if ! meson "$build" "-D$fuzzflag" -Db_lundef=false -Dauto_features="$auto_features"; then
     cat "$build/meson-logs/meson-log.txt"
     exit 1
 fi
```
should make the script do what it currently does without the feature options.

evverx:
I'm not sure what to do about `fuzz-bcd`. It seems to be excluded on OSS-Fuzz due `auto_features=disabled`. As long as it isn't linked against `gnu-efi` (or any other external library) I think it should be possible to just pass `-Dgnu-efi=enabled` on OSS-Fuzz to start building it again.

medhefgo:
Fedora ci is failing because it passes `--auto-features=enabled`. It needs to disable `libidn` and `passwdqc`

meson.build
```diff
@@ -1349,23 +1349,12 @@ libp11kit = dependency('p11-kit-1',
...
-        else
-                have = false
-                libfido2 = []
-        endif
-else
-        have = false
-        libfido2 = []
-endif
-conf.set10('HAVE_LIBFIDO2', have)
+libfido2 = get_option('libfido2').require(
```
yuwata: I'd like to name this `required` or `req`. `libxyz` sounds it is a dependency.

meson.build
```diff
@@ -1572,43 +1555,25 @@ conf.set('DEFAULT_DNSSEC_MODE',
...
-        have = (conf.get('HAVE_OPENSSL') == 1 and
-                conf.get('HAVE_LIBFDISK') == 1)
-        if want_sysupdate == 'true' and not have
-                error('sysupdate support was requested, but dependencies are not available')
-        endif
-else
-        have = false
-endif
+have = get_option('sysupdate').require(
+        (conf.get('HAVE_OPENSSL') == 1 and
```
yuwata: Redundant `()` ?

meson.build
```diff
@@ -1572,43 +1555,25 @@ conf.set('DEFAULT_DNSSEC_MODE',
...
-                conf.get('HAVE_ZLIB') == 1 and
-                conf.get('HAVE_XZ') == 1)
-        if want_importd == 'true' and not have
-                error('importd support was requested, but dependencies are not available')
-        endif
-else
-        have = false
-endif
+have = get_option('importd').require(
+        (conf.get('HAVE_LIBCURL') == 1 and
```
yuwata: Also here.

meson.build
```diff
@@ -1572,43 +1555,25 @@ conf.set('DEFAULT_DNSSEC_MODE',
...
-                conf.get('HAVE_LIBFDISK') == 1 and
-                conf.get('HAVE_LIBCRYPTSETUP') == 1)
-        if want_homed == 'true' and not have
-                error('homed support was requested, but dependencies are not available')
-        endif
-else
-        have = false
-endif
+have = get_option('homed').require(
+        (conf.get('HAVE_OPENSSL') == 1 and
```
yuwata: Also here.

meson.build
```diff
@@ -1841,18 +1806,21 @@ efi_arch = {
...
 
-if get_option('bootloader') != 'false' and efi_arch != ''
-        conf.set_quoted('EFI_MACHINE_TYPE_NAME', efi_arch)
-elif get_option('bootloader') == 'false' and efi_arch != ''
-        # Ensure that if the option is explicitly set to false, then no EFI code is built, including tests
-        efi_arch = ''
-elif get_option('bootloader') == 'true' and efi_arch == ''
-        error('EFI not supported for this arch.')
-endif
+have = pymod.find_installation(
```
yuwata: I'd like to name `pyelf` or so, and drop `.found()`.

meson.build
```diff
@@ -1841,18 +1806,21 @@ efi_arch = {
...
+        'python3',
+        required : get_option('bootloader'),
+        modules : ['elftools']
+).found()
+
+have = get_option('bootloader').require(
+        have and get_option('efi') and efi_arch != '',
```
yuwata: Then, `pyelf.found()`.

test/fuzz/meson.build
```diff
@@ -36,18 +36,26 @@ fuzz_regression_tests += { 'fuzz-unit-file' : dict }
 
 ############################################################
 
+if get_option('auto_features').enabled()
+        sanitize_auto_features = 'enabled'
```
yuwata, line 41: Is it not possible to format auto_features?

medhefgo: Turns out it's possible… with meson >= 1.3.0 :D

(I added a TODO comment.)

yuwata:
> Fedora ci is failing because it passes `--auto-features=enabled`. It needs to disable `libidn` and `passwdqc`

I guess this causes the following error message, right?
```
meson.build:1125:8: ERROR: Problem encountered: passwdqc and pwquality cannot be requested simultaneously
```

So, how about to make them gracefully?
That is, if two conflicting libraries requested and found, how about to pick one of them gracefully?

medhefgo:
Now it's having the same error that mkosi had, requesting polkit support without it being installed: `
meson.build:1159:12: ERROR: Dependency "polkit-gobject-1" not found, tried pkgconfig
`

meson.build
```diff
@@ -1320,18 +1315,17 @@ libqrencode = dependency('libqrencode',
...
+                          required : feature.disabled() ? feature : false)
+if not libgpg_error.found()
+        # CentOS 8 does not provide the .pc file.
+        libgpg_error = cc.find_library('gpg-error', required : feature)
 endif
+
+have = libgcrypt.found() and libgpg_error.found()
 if not have
         # link to neither of the libs if one is not found
         libgcrypt = []
```
yuwata, line 1297: How about to use `dependency('', required : false)` for consistency with other places?

medhefgo: There is no value in that as long as no dep functions are used on it. So keeping it this way is shorter.

yuwata:
How about the following?
https://github.com/yuwata/systemd/commit/0ba5fffa91759e8b26a8678eedb8a7f9e540da7a
Feel free to pick it or squash with your commits, if you'd like.

meson.build
yuwata: OK. That makes sense.

meson.build
```diff
@@ -1119,32 +1119,16 @@ libfdisk = dependency('fdisk',
...
-else
-        have = false
-        libpwquality = []
+libpwquality = dependency('pwquality',
+                          version : '>= 1.4.1',
+                          required : get_option('pwquality'))
+have = libpwquality.found()
+if not have
+        libpwquality = dependency('passwdqc',
+                                  required : get_option('passwdqc'))
```
yuwata: This is problematic, as even if pwquality is auto and passwdqc is enabled, passwdqc will not be used when pwquality is installed.

meson.build
```diff
@@ -1275,30 +1257,16 @@ libcurl = dependency('libcurl',
...
-else
-        have = false
-        libidn = []
-endif
-conf.set10('HAVE_LIBIDN2', have)
-if not have and want_libidn != 'false' and not skip_deps
+libidn = dependency('libidn2',
+                    required : get_option('libidn2'))
+have = libidn.found()
+if not have
```
yuwata: Also in this case.

meson.build
medhefgo: I don't see an issue with that. If you request one, you should be disabling the other. In fact, the old code made requesting both an error so this question doesn't arise.

medhefgo: I just realized that this is the use case for `disable_auto_if()`. I've changed it to use that so that `-Dpwquality=auto -Dpasswdqc=enabled` should pick the latter, while `-Dpwquality=enabled -Dpasswdqc=enabled` will prefer the former. Same with libidn.

meson.build
```diff
@@ -1119,32 +1119,17 @@ libfdisk = dependency('fdisk',
...
-else
-        have = false
-        libpwquality = []
+feature = get_option('pwquality').disable_auto_if(get_option('passwdqc').enabled())
+libpwquality = dependency('pwquality',
+                          version : '>= 1.4.1',
+                          required : feature)
+have = libpwquality.found()
+if not have
+        libpwquality = dependency('passwdqc',
```
yuwata, line 1129: Please add a short comment that `libpwquality` is also used for `passwdqc`.
Though, we never pass it to any executables or libraries.

yuwata:
Still packit are in red:
```
Run-time dependency polkit-gobject-1 found: NO (tried pkgconfig)

meson.build:1160:12: ERROR: Dependency "polkit-gobject-1" not found, tried pkgconfig
```

meson.build
```diff
@@ -1119,32 +1119,17 @@ libfdisk = dependency('fdisk',
...
-
-if want_pwquality != 'false' and want_passwdqc != 'true' and not skip_deps
-        libpwquality = dependency('pwquality',
-                                  version : '>= 1.4.1',
-                                  required : want_pwquality == 'true')
-        have = libpwquality.found()
-else
-        have = false
-        libpwquality = []
+feature = get_option('pwquality').disable_auto_if(get_option('passwdqc').enabled())
```
yuwata, line 1122: Please also add short comment that pwquality has higher precedence when both enabled (or both auto).

meson.build
```diff
@@ -1205,141 +1157,88 @@ if have
...
-        install_polkit = true
-
-        libpolkit = dependency('polkit-gobject-1',
-                               required : false)
-        if libpolkit.found() and libpolkit.version().version_compare('< 0.106')
-                message('Old polkit detected, will install pkla files')
-                install_polkit_pkla = true
-        endif
+libpolkit = dependency('polkit-gobject-1',
+                       required : get_option('polkit'))
```
yuwata: The dependency object is not passed to executables or libraries, so, we can downgrade the required level like we did previously. Maybe, the below should work.
`required : get_option('polkit').allowed()`

meson.build
```diff
@@ -1205,141 +1157,88 @@ if have
...
+install_polkit = libpolkit.found()
```
yuwata: `install_polkit = get_option('polkit').enabled() or libpolkit.found()` ?

meson.build
medhefgo: I see now that the dependency is just used to check for old polkit. Behavior should now match the old one.

meson.build
```diff
@@ -1205,141 +1159,89 @@ if have
...
+           libblkid.found() and cc.has_function('blkid_probe_set_hint', dependencies : libblkid))
+
+libkmod = dependency('libkmod',
+                     version : '>= 15',
+                     required : get_option('kmod'))
+conf.set10('HAVE_KMOD', libkmod.found())
+
+libxenctrl = dependency('xencontrol',
+                        version : '>= 4.9',
+                        required : get_option('xenctrl'))
```
yuwata, line 1192: Could you downgrade the requirement except for x86_64 and arm64 ?
https://wiki.xenproject.org/wiki/Xen_Project_FAQ_Compatibility
Currently, only x86_64 and arm is supported, I am not sure the state of 32bit arm, but at least fedora does not package xen for 32bit arm.

yuwata (changes requested):
Looks mostly good.

Ubuntu CI failure should not be related, and most of them should be fixed by #28808, #28813, #28818, and #28819.

meson.build
medhefgo: I'd rather see packit pass `-Dxenctrl` explicitly based on what they want instead of adding workarounds just for that…

@keszybz Any chance you can do that?

yuwata: I think we should support auto_feature gracefully.

medhefgo: > I think we should support auto_feature gracefully.

That completely defeats the purpose of `auto-features=enabled`. If you pass it, you should be prepared for the fallout by toggling individual dependencies. And I really don't like the idea of encoding which dependency can work on whatever arch+distro combo people come up with.

idn/pwquality was different in that we can use only one anyway.

yuwata: -> https://src.fedoraproject.org/rpms/systemd/pull-request/115

Though, I am not familiar with rpm spec. I mostly forget the syntax...

yuwata:
Not tested, but fedora-ppc64le and s390x failure should be fixed by the following PR.
https://src.fedoraproject.org/rpms/systemd/pull-request/116

yuwata (approved):
All green now!! LGTM.

yuwata:
Setting the green label, and let's wait for @keszybz's comment for a while.

meson.build
keszybz: That PR was merged. I agree with Jan that the fix should be on the distro side in this case.

meson.build
```diff
@@ -1543,104 +1415,60 @@ else
...
                 compression = 'none'
         endif
-elif compression == 'zstd' and not have_zstd
+elif compression == 'zstd' and not libxz.found()
         error('default-compression=zstd requires zstd')
-elif compression == 'lz4' and not have_lz4
+elif compression == 'lz4' and not liblz4.found()
         error('default-compression=lz4 requires lz4')
-elif compression == 'xz' and not have_xz
+elif compression == 'xz' and not libzstd.found()
```
keszybz: Those are mixed up.

meson.build
```diff
@@ -1543,104 +1415,60 @@ else
...
 if compression == 'auto'
-        if have_zstd
+        if libxz.found()
                 compression = 'zstd'
-        elif have_lz4
+        elif liblz4.found()
                 compression = 'lz4'
-        elif have_xz
+        elif libzstd.found()
                 compression = 'xz'
```
keszybz: Here too.

meson_options.txt
```diff
@@ -350,11 +350,11 @@ option('www-target', type : 'string',
        description : 'the address and dir to upload docs too',
        value : 'www.freedesktop.org:/srv/www.freedesktop.org/www/software/systemd')
 
-option('seccomp', type : 'combo', choices : ['auto', 'true', 'false'],
+option('seccomp', type : 'feature', deprecated : { 'true' : 'enabled', 'false' : 'disabled' },
```
keszybz, line 353: Oh, this is nice. I didn't know that this is possible.

keszybz (changes requested):

keszybz:
This is very nice. I'm not 100% sure that the logic is correct in all cases, e.g. with some off the complex fallbacks to fallback on old systems, but if something's wrong, it should be easy enough to fix.

keszybz (approved):

PR #24178, opened 2022-08-02
test: install libgcc_s.so.1 explicitly if available

mrc0mmand:
Since the library is dlopen()ed by libpthread and required during
pthread_exit()/pthread_cancel(), let's install it explicitly if available to
avoid unexpected fails in tests. This also consolidates all related
workarounds for this library across the test scripts.

See: https://github.com/systemd/systemd/pull/23858

/cc @mbiebl @enr0n


bluca (approved):

mrc0mmand:
Ugh, `gcc --print-file-name "libgcc_s.so.1"` return the "unsymlinked" name, e.g. `/usr/lib64/libgcc_s-12-20220507.so.1` in this case, which means we don't install the underlying symlink (`/usr/lib64/libgcc_s.so.1`). I'll need to tweak this further.

PR #24179, opened 2022-08-02
mountpoint-util: make fd_is_mountpoint() work reasonably with AT_EMPT…

poettering:
…YPATH

Inspired by: https://github.com/systemd/systemd/pull/24141

Calling fd_is_mountpoint() with AT_EMPTYPATH and an empty filename can
only work if we have new statx() available. If we do not, we can still
make things work for directories, but not for other inodes (since there
we cannot query information about the parent inode to compare things.)

Hence, let's handle and test this explicitly, to support this to the
level this is possible.

### DIFF
--- a/TODO
+++ b/TODO
@@ -102,6 +102,9 @@ Deprecations and removals:
 * drop support for LOOP_CONFIGURE-less loopback block devices, once kernel
   baseline is 5.8.
 
+* drop fd_is_mount_point() fallback mess once we can rely on
+  STATX_ATTR_MOUNT_ROOT to exist i.e. kernel baseline 5.8
+
 * rework our PID tracking in services and so on, to be strictly based on pidfd,
   once kernel baseline is 5.13.
 


src/basic/mountpoint-util.c
```diff
@@ -182,12 +182,23 @@ int fd_is_mount_point(int fd, const char *filename, int flags) {
...
-        /* Insist that the specified filename is actually a filename, and not a path, i.e. some inode further
-         * up or down the tree then immediately below the specified directory fd. */
-        if (!filename_possibly_with_slash_suffix(filename))
+        /* Note that AT_EMPTY_PATH is only supported if the kernel supports statx(), or if the inode
+         * specified via 'fd' refers to a directory. Otherwise, we'll have to fail, because we have no kernel
+         * API to query the information we need.
+         *
+         * Also note that 'filename' being NULL will turn on AT_EMPTY_PATH implicitly. */
+
+        if (FLAGS_SET(flags, AT_EMPTY_PATH))
```
daandemeyer: Instead of all this normalization, why not just reduce the supported flags to only `AT_SYMLINK_FOLLOW` and infer `AT_EMPTY_PATH` solely from the `filename` argument?

src/test/test-mountpoint-util.c
```diff
@@ -290,6 +291,24 @@ TEST(fd_is_mount_point) {
...
+        assert_se(fd_is_mount_point(fd, NULL, 0) > 0);
+        assert_se(fd_is_mount_point(fd, "", AT_EMPTY_PATH) > 0);
+        assert_se(fd_is_mount_point(fd, "", 0) == -EINVAL);
+
+        safe_close(fd);
+        fd = open("/proc/version", O_RDONLY|O_CLOEXEC|O_NOCTTY);
+        assert_se(fd >= 0);
+
+        r = fd_is_mount_point(fd, NULL, 0);
+        assert_se(r == 0 || r == -ENOTDIR); /* on old kernels we can't determine if regular files are mount points if we have no directory fd */
```
daandemeyer: Use `IN_SET()` here and below?

src/basic/mountpoint-util.c
```diff
@@ -182,12 +182,23 @@ int fd_is_mount_point(int fd, const char *filename, int flags) {
...
+                filename = empty_to_null(filename); /* Normalize → NULL */
+        else if (!filename)
```
bluca: shouldn't this not be an `else` so that it catches the case where the above empty_to_null happens?

bluca:
> If we do not, we can still
> make things work for directories, but not for other inodes (since there
> we cannot query information about the parent inode to compare things.)

Does this mean this will introduce a regression for older kernels? Or simply things will improve only with new kernels and stay the same elsewhere?

DaanDeMeyer (approved):
Looks good, few nits.

Never figured it'd be this messy to figure out if something is a mount point.

DaanDeMeyer:
> Does this mean this will introduce a regression for older kernels? Or simply things will improve only with new kernels and stay the same elsewhere?

Currently, this function can't be used with AT_EMPTY_PATH at all. This PR makes it possible to use it with AT_EMPTY_PATH, so we only expand the use cases, we don't regress

poettering:
> Does this mean this will introduce a regression for older kernels? Or simply things will improve only with new kernels and stay the same elsewhere?

It makes stuff work under some circumstances that failed badly before. And where it can't make things work it will fail much less badly.

poettering:
> Never figured it'd be this messy to figure out if something is a mount point.

Once we can rely on STATX_ATTR_MOUNT_ROOT to always work things become *a* *lot* simpler...

I added another commit that adds an item to the TODO list: remove the terrible fallbacks once linux 5.8 becomes our baseline (i.e. the kernel STATX_ATTR_MOUNT_ROOT became available)

bluca (approved):

DaanDeMeyer (approved):

PR #24180, opened 2022-08-02
repart: unlock block device when finished

msizanoen1:
Unlock block device when finished to prevent systemd-repart from hanging
on EFI-variable-based factory reset.


DaanDeMeyer:
Can you strace systemd-repart and link the output here in a github gist to show exactly where repart hangs?

PR #24181, opened 2022-08-02
repart: unlock block device when finished

msizanoen1:
Unlock block device when finished to prevent systemd-repart from hanging
on EFI-variable-based factory reset.

PR #24183, opened 2022-08-02
repart: when keeping ref to backing inode/devnode, use fd_reopen() ra…

poettering:
…thern than F_DUPFD

Via the "backing_fd" variable we intend to pin the backing inode through
our entire code. So far we typically created the fd via F_DUPFD_CLOEXEC,
and thus any BSD lock taken one the original fd is shared with our
backing_fd reference. And if the origina fd is closed but our backing_fd
is not, we'll keep the BSD lock open, even if we then reopen the block
device through the backing_fd. If hit, this results in a deadlock.

Let's fix that by creating the backing_fd via fd_reopen(), so that the
locks are no longer shared, and if the original fd is closed all BSD
locks on it that are in effect are auto-released.

(Note the deadlock is only triggered if multiple operations on the same
backing inode are executed, i.e. factory reset, resize and applying of
partitions.)

Replaces: #24181

### DIFF
--- a/src/partition/repart.c
+++ b/src/partition/repart.c
@@ -1612,9 +1612,9 @@ static int context_load_partition_table(
 
         if (*backing_fd < 0) {
                 /* If we have no fd referencing the device yet, make a copy of the fd now, so that we have one */
-                *backing_fd = fcntl(fdisk_get_devfd(c), F_DUPFD_CLOEXEC, 3);
+                *backing_fd = fd_reopen(fdisk_get_devfd(c), O_RDONLY|O_CLOEXEC);
                 if (*backing_fd < 0)
-                        return log_error_errno(errno, "Failed to duplicate fdisk fd: %m");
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
On top of this, shouldn't we move the lock to be a single lock on the backing fd so that we keep the device locked for the entire duration of systemd-repart instead of unlocking/locking a few times as we move through the stages?

bluca (approved):

poettering:
> On top of this, shouldn't we move the lock to be a single lock on the backing fd so that we keep the device locked for the entire duration of systemd-repart instead of unlocking/locking a few times as we move through the stages?

I guess we could, but it's not strictly necessary, I see no reason why the changes we make need to appear atomic to the outside.

We can certainly do that in a separate PR though.

msizanoen1:
I have tested this and it seems to fix the issue.

poettering:
excellent, then let's merge.

PR #24187, opened 2022-08-03
repart: Only lock block device once

daandemeyer:
Let's lock the backing fd instead of locking/unlocking multiple
times when doing multiple operations with repart. It doesn't make
much sense for anything else to touch the block device while there
are still repart operations pending on it. By keeping the lock over
the full duration of repart, we avoid anything else from interfering
with the block device inbetween operations.


poettering:
lgtm

yuwata (approved):
LGTM.

bluca (approved):

PR #20255, opened 2021-07-20
Adding auto-enrollement of secure boot variables.

vdagonneau-anssi:
For now, it enrolls db, KEK and then PK as long as these
keys are available at the root on the ESP and Secure Boot
is in Setup Mode. The files should be named `db.auth`,
`KEK.auth` and `PK.auth`.

You also need to add the `auto-enroll yes` option to your
systemd-boot config file.

The files need to be prepared with `cert-to-efi-sig-list`
and then signed with `sign-efi-sig-list`.

Here is a short example to generate your own keys and the
right files for auto-enrollement.

```
keys="PK KEK DB"
for key in ${keys}; do
	openssl req -new -x509 -subj "/CN=${key}/ -keyout "${key}.key" -out "${key}.crt"
	openssl x509 -outform DER -in "${key}.crt" -out "${key}.cer"
	cert-to-efi-sig-list -g "${uuid}" "${key}.crt" "${key}.esl"
done

sign-efi-sig-list -c PK.crt -k PK.key PK PK.esl PK.auth
sign-efi-sig-list -c PK.crt -k PK.key KEK KEK.esl KEK.auth
sign-efi-sig-list -c KEK.crt -k KEK.key db DB.esl DB.auth
```

Once these keys are auto-enrolled, the UEFI binaries on the ESP
**NEED** to be signed in order to run. You can sign the binaries
with the `sbsign` tool, for example:

```
sbsign --key DB.key --cert DB.crt bzImage --output $ESP/bzImage
```

PR #24189, opened 2022-08-03
boot: Follow-up fixes for #20255

medhefgo:


bluca (approved):

man/loader.conf.xml
```diff
@@ -244,20 +244,29 @@
...
-  cert-to-efi-sig-list -g "${uuid}" "${key}.crt" "${key}.tmp"
+  cert-to-efi-sig-list -g "${uuid}" "${key}.crt" "${key}.esl"
 done
 
-sign-efi-sig-list -c PK.crt -k PK.key PK PK.tmp PK.esl
-sign-efi-sig-list -c PK.crt -k PK.key KEK KEK.tmp KEK.esl
-sign-efi-sig-list -c KEK.crt -k KEK.key db db.tmp db.esl
+# This step adds the Microsoft Windows and UEFI keys to the database and is optional, but recommended.
+# The first key is needed to be able to use the Windows boot loader, the second key is needed for
+# firmware drivers and third-party boot loaders signed by Microsoft.
```
poettering: drivers of the platform firmware don't have to be signed btw, they are trusted because loaded from a trusted storage source. 

I think we should mention the expression "option ROM" here, because that's what it really is about, i.e. stuff that is neither from the platform vendor nor from the OS vendor, but on some extension card.

man/loader.conf.xml
```diff
@@ -244,20 +244,29 @@
...
   openssl req -new -x509 -subj "/CN=${key}/" -keyout "${key}.key" -out "${key}.crt"
   openssl x509 -outform DER -in "${key}.crt" -out "${key}.cer"
-  cert-to-efi-sig-list -g "${uuid}" "${key}.crt" "${key}.tmp"
+  cert-to-efi-sig-list -g "${uuid}" "${key}.crt" "${key}.esl"
 done
 
-sign-efi-sig-list -c PK.crt -k PK.key PK PK.tmp PK.esl
-sign-efi-sig-list -c PK.crt -k PK.key KEK KEK.tmp KEK.esl
-sign-efi-sig-list -c KEK.crt -k KEK.key db db.tmp db.esl
+# This step adds the Microsoft Windows and UEFI keys to the database and is optional, but recommended.
```
poettering: I'd say "Microsoft Windows and Microsoft 3rd Party certificates"

poettering: I'd also conditionalize the "but recommended" a bit. i.e. say it's recommended for *general* *purpose* OSes.

After all, in environments such as VMs the Microsoft keys are pretty irrelevant, and even a bad idea. Hence, if you focus on VMs only, or on a specific platform, don't bother with the Microsoft keys. In fact, it might be wise to turn this around altogether, and mention outside of bare metal systems where dual boot with another OS or where option ROM devices are used it's totally unnecessary.

poettering:
looks good, some minor comments.

@vdagonneau-anssi could you have a look, too?

man/loader.conf.xml
vdagonneau-anssi: Agree, with @poettering here.

Should we maybe add an '(e.g. shim)' after third-party bootloader here? This would probably be the most common case.

vdagonneau-anssi: By the way, I think there has been multiple vulnerabilities discovered in signed version of the shim. Doesn't loading the microsoft key without loading any dbx for blacklisting these versions open you up to anyone booting the vulnerable shim and using it to do anything?

Maybe I'm thinking too far here.

medhefgo: @poettering I would not trust a vendor to always have their internal drivers be trusted. I did add option ROM (it's really a terrible name, considering they are less than optional for nvidia GPUs, even on laptops).

@vdagonneau-anssi That's a good point. The nice thing is that we have nothing to do here: :D
Microsoft provides a third key (KEK) which is used to sign the official dbx updates. So I just added that key to the instructioins and referred to fwupdmgr to update the dbx.

medhefgo:
I also added a another commit to skip the safety countdown in VMs.

poettering:
lgtm

PR #24195, opened 2022-08-03
meson: Bump libcryptsetup min version to 2.1.0

daandemeyer:
crypt_init_data_device() was introduced in libcryptsetup 2.1.0.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1283,7 +1283,7 @@ endif
 
 if want_libcryptsetup != 'false' and not skip_deps
         libcryptsetup = dependency('libcryptsetup',
-                                   version : want_libcryptsetup_plugins == 'true' ? '>= 2.4.0' : '>= 2.0.1',
+                                   version : want_libcryptsetup_plugins == 'true' ? '>= 2.4.0' : '>= 2.1.0',
                                    required : want_libcryptsetup == 'true' or want_libcryptsetup_plugins == 'true')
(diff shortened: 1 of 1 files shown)


mbiebl:
LGTM

bluca (changes requested):
it's used only for one optional argument of one optional tool, I can send a PR to ifdef instead

bluca:
See: https://github.com/systemd/systemd/pull/24196

mbiebl:
Hm, 2.1 is even available in Debian oldstable. Do we have distros, which don't ship 2.1?

bluca:
> Hm, 2.1 is even available in Debian oldstable. Do we have distros, which don't ship 2.1?

Yes there's Bionic, which is where Daan found the issue

mrc0mmand:
I mean, I am all for keeping backwards compat where it makes sense, but in this case it's an LTS release from two versions ago... but I guess that's just the meson version bump discussion all over again, not worth getting into.

bluca:
> I mean, I am all for keeping backwards compat where it makes sense, but in this case it's an LTS release from two versions ago... but I guess that's just the meson version bump discussion all over again, not worth getting into.

Well, good thing that we kept it working, so that Daan could find this issue, leading us to finding this little memory leak that nobody noticed :-) https://github.com/systemd/systemd/pull/24196#issuecomment-1204365895

PR #24196, opened 2022-08-03
integritysetup: do not use crypt_init_data_device after crypt_init

bluca:
crypt_init_data_device() replaces the crypt_device struct with a
new allocation, losing the old one, which we get from crypt_init().
Use crypt_set_data_device() instead.

Replaces https://github.com/systemd/systemd/pull/24195

### DIFF
--- a/src/integritysetup/integritysetup.c
+++ b/src/integritysetup/integritysetup.c
@@ -152,12 +152,6 @@ static int run(int argc, char *argv[]) {
                         return 0;
                 }
 
-                if (!isempty(arg_existing_data_device)) {
-                        r = crypt_init_data_device(&cd, device, arg_existing_data_device);
-                        if (r < 0)
-                                return log_error_errno(r, "Failed to add separate data device: %m");
-                }
-
                 r = crypt_load(cd,
                         CRYPT_INTEGRITY,
                         &(struct crypt_params_integrity) {
@@ -167,6 +161,12 @@ static int run(int argc, char *argv[]) {
                         });
                 if (r < 0)
                         return log_error_errno(r, "Failed to load integrity superblock: %m");
+
+                if (!isempty(arg_existing_data_device)) {
+                        r = crypt_set_data_device(cd, arg_existing_data_device);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to add separate data device: %m");
+                }
 
                 r = crypt_activate_by_volume_key(cd, volume, key_buf, key_buf_size, arg_activate_flags);
                 if (r < 0)
--- a/test/units/testsuite-67.sh
+++ b/test/units/testsuite-67.sh
@@ -41,6 +41,7 @@ if [ -z "${image_dir}" ] || [ ! -d "${image_dir}" ]; then
 fi
 
 dd if=/dev/zero of="${image_dir}/image" bs=1048576 count=64 || exit 1
+dd if=/dev/zero of="${image_dir}/data" bs=1048576 count=64 || exit 1
 loop="$(losetup --show -f "${image_dir}/image")"
 
 if [[ ! -e ${loop} ]]; then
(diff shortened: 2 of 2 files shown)


src/integritysetup/integritysetup.c
```diff
@@ -153,9 +153,13 @@ static int run(int argc, char *argv[]) {
                 }
 
                 if (!isempty(arg_existing_data_device)) {
+#if HAVE_CRYPT_INIT_DATA_DEVICE
                         r = crypt_init_data_device(&cd, device, arg_existing_data_device);
                         if (r < 0)
                                 return log_error_errno(r, "Failed to add separate data device: %m");
+#else
+                        log_warning("data-device= option not supported by current version of libcryptsetup, ignoring.");
```
poettering: is this really optional? i assumed if the data device is not specified when we are asked to the integrity stuff simply won't work. or what am I missing?

Hence, I would assume this here should be fatal?

bluca: it's an optional value, there's no fatal error if it's not provided

poettering: Hmm? so if i have set up my integrity volume with data device x and tags/integrity info on device y on my devel machine, and then want to use that on some other system via /etc/integritytab, and hence specify the two devices x and y in /etc/integritytab there and boot, then this will ignore the data device x and just use the device y, but since that contains no user data, but only integrity info it will fail with an ugly error, no?

I either don't understand what "data-device=" is actually about, or it's not optional at all if you already have a setup that splits up user data from integrity data storage and intend to mount that...

bluca: I'm just referring to how to the argument parsing is set up. The integritysetup manpage says:

> --data-device
>     Specify a separate data device that contains existing data. The <device> then will contain calculated
>     integrity tags and journal for this data device.

bluca: Yes even with the cryptsetup command line it's entirely optional, I guess it stores data inline or something:

```
$ sudo integritysetup format /tmp/integrity 

WARNING!
========
This will overwrite data on /tmp/integrity irrevocably.

Are you sure? (Type 'yes' in capital letters): YES
Formatted with tag size 4, internal integrity crc32c.
Wiping device to initialize integrity checksum.
You can interrupt this by pressing CTRL+c (rest of not wiped device will contain invalid checksum).
Finished, time 00:00.026,   23 MiB written, speed 914.4 MiB/s
$ sudo integritysetup open /tmp/integrity foo
$ sudo integritysetup status foo
/dev/mapper/foo is active.
  type:    INTEGRITY
  tag size: 4
  integrity: crc32c
  device:  /dev/loop5
  loop:    /tmp/integrity
  sector size:  512 bytes
  interleave sectors: 32768
  size:    49128 sectors
  mode:    read/write
  failures: 0
  journal size: 180224 bytes
  journal watermark: 50%
  journal commit time: 10000 ms
```

tasleson: If a user is actually using a device to store the checksums and a separate device for the data `--data-device`, the data device isn't optional.  Without it you have no data.  When you do a integrity format, the default is to interleave the checksums with the data on the same device.

bluca: so it sounds like the parameter is optional, but if it is passed then it needs to be used - I'll update to error out instead

DaanDeMeyer:
Actually, We already call `crypt_init()`, so we can just replace this function call with `crypt_set_data_device()`

bluca:
> Actually, We already call `crypt_init()`, so we can just replace this function call with `crypt_set_data_device()`

Ah, great catch, this is actually a bug, as it would then replace the crypt_device structure leaking it

src/integritysetup/integritysetup.c
bluca: never mind, it was wrong to call crypt_init_data_device here, will remove it

DaanDeMeyer (approved):

bluca:
> > Actually, We already call `crypt_init()`, so we can just replace this function call with `crypt_set_data_device()`
> 
> Ah, great catch, this is actually a bug, as it would then replace the crypt_device structure leaking it

More details: we have already called crypt_init() which allocates the crypt_device, but then crypt_init_data_device() calls it again:

https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/lib/setup.c#L680

replacing the existing pointer:

https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/lib/setup.c#L603

So the old crypt_device is leaked.

tasleson:
> @tasleson Does this look reasonable?

Yep, good catch!

bluca:
Added a test and found another issue (there was no coverage before), holding on for a moment while I solve it

bluca:
> Added a test and found another issue (there was no coverage before), holding on for a moment while I solve it

Ok sorted, the set_data_device needs to happen after crypt_data is initialized with the type. Added test case for separate data device so that it's covered.

DaanDeMeyer (approved):

PR #24199, opened 2022-08-04
Remove a few HAVE_BLKID ifdefs

daandemeyer:
The code inside these blocks doesn't actually depend on anything
from blkid so let's not gate on it.

### DIFF
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -125,7 +125,6 @@ not_found:
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


poettering:
> The code inside these blocks doesn't actually depend on anything
> from blkid so let's not gate on it.

hmm, but they are only used by code that is only enabled when blkid is enabled, no? thus, if people turn off blkid they'll get warnings about unused static functions being defined.

PR #24205, opened 2022-08-04
test-network: support the case that l2tp module is not supported

yuwata:
Hopefully fixes the issue reported at
https://github.com/systemd/systemd-centos-ci/pull/517#issuecomment-1205399798.


mrc0mmand (approved):
LGTM, thank you!

PR #24210, opened 2022-08-05
oomd: notify via dbus what have been killed

Mc-Pain:
This could be useful for notifying user about which applications/cgroups were killed by systemd-oomd.

Unfortunately, low-memory-monitor does not fit for this purpose. It tells whether oomd is about to kill something or not


src/oom/org.freedesktop.oom1.conf
```diff
@@ -41,6 +41,14 @@
                        send_interface="org.freedesktop.oom1.Manager"
                        send_member="DumpByFileDescriptor"/>
 
+                <allow send_destination="org.freedesktop.oom1"
+                       send_interface="org.freedesktop.oom1.Manager"
+                       send_member="KilledByMemoryUsed"/>
+
+                <allow send_destination="org.freedesktop.oom1"
+                       send_interface="org.freedesktop.oom1.Manager"
+                       send_member="KilledByMemoryPressure"/>
```
poettering: this is for methds, not signals.

src/oom/oomd-manager-bus.c
```diff
@@ -38,6 +38,14 @@ static const sd_bus_vtable manager_vtable[] = {
...
                                  bus_method_dump_by_fd,
                                  SD_BUS_VTABLE_UNPRIVILEGED),
+        SD_BUS_SIGNAL_WITH_NAMES("KilledForMemoryUsed",
+                                 "s",
+                                 SD_BUS_PARAM(name),
+                                 0),
+        SD_BUS_SIGNAL_WITH_NAMES("KilledForMemoryPressure",
+                                 "s",
+                                 SD_BUS_PARAM(name),
+                                 0),
```
poettering, line 45: this really should be the same signal i think, and carry a "reason" string field, or so. which contains "memory-used" or "memory-pressure"

"name" is not a good name, noone knows what that could be

src/oom/oomd-manager.c
```diff
@@ -410,14 +410,23 @@ static int monitor_swap_contexts_handler(sd_event_source *s, uint64_t usec, void
...
                                 log_notice("Killed %s due to memory used (%"PRIu64") / total (%"PRIu64") and "
                                            "swap used (%"PRIu64") / total (%"PRIu64") being more than "
                                            PERMYRIAD_AS_PERCENT_FORMAT_STR,
                                            selected,
                                            m->system_context.mem_used, m->system_context.mem_total,
                                            m->system_context.swap_used, m->system_context.swap_total,
                                            PERMYRIAD_AS_PERCENT_FORMAT_VAL(m->swap_used_limit_permyriad));
+
+                                /* send dbus signal */
+                                sd_bus_emit_signal(m->bus,
```
poettering: ignoring the return value is OK, but please cast the result to `(void)` please to tell static checkers you are knowingly ignoring the return value.

src/oom/org.freedesktop.oom1.conf
Mc-Pain: I think I could remove this hunk then

src/oom/oomd-manager-bus.c
Mc-Pain: Does "cgroup" fit instead of "name"?

bluca:
Please document this new signal. You can generate the skeleton to fill in with: `ninja -C <build dir> update-dbus-docs`

src/oom/oomd-manager-bus.c
```diff
@@ -38,6 +38,11 @@ static const sd_bus_vtable manager_vtable[] = {
                                  SD_BUS_PARAM(fd),
                                  bus_method_dump_by_fd,
                                  SD_BUS_VTABLE_UNPRIVILEGED),
+        SD_BUS_SIGNAL_WITH_NAMES("KilledForMemoryUsage",
```
poettering: signal should really be called just "Killed" or so?

man/org.freedesktop.oom1.xml
```diff
@@ -56,12 +59,16 @@ node /org/freedesktop/oom1 {
...
+    <variablelist class="dbus-signal" generated="True" extra-ref="Killed"/>
+
     <!--End of Autogenerated section-->
 
     <refsect2>
       <title>Methods</title>
 
       <para>...</para>
+
+      <para><function>Killed</function> signal is sent when any cgroup is killed by oomd</para>
```
bluca: Please be more precise here. Describe the two possible values of the 'reason' parameter and what they mean.

bluca:
Please squash the second commit in the first one. It shouldn't be separate as missing doc means failed tests.

man/org.freedesktop.oom1.xml
Mc-Pain:
```Possible reasons:
      <literal>memory-used</literal> - application took too much memory</para>
      <literal>memory-pressure</literal> - application took enough memory to cause sufficient slowdown of other applications</para>```

I suggest this (AFAIK what the PSI is)

**Review by bluca — APPROVED** [review]

File: man/org.freedesktop.oom1.xml [hunk 6]
```diff
@@ -56,12 +59,19 @@ node /org/freedesktop/oom1 {
...
     <variablelist class="dbus-method" generated="True" extra-ref="DumpByFileDescriptor()"/>
 
+    <variablelist class="dbus-signal" generated="True" extra-ref="Killed"/>
+
     <!--End of Autogenerated section-->
 
     <refsect2>
       <title>Methods</title>
 
       <para>...</para>
```
yuwata: Please drop this paragraph.

man/org.freedesktop.oom1.xml
```diff
@@ -56,12 +59,19 @@ node /org/freedesktop/oom1 {
...
+
+      <para><function>Killed</function> signal is sent when any cgroup is killed by oomd
```
yuwata: Missing period.

man/org.freedesktop.oom1.xml
```diff
@@ -56,12 +59,19 @@ node /org/freedesktop/oom1 {
...
+      Possible reasons:
+      <literal>memory-used</literal> - application took too much memory</para>
+      <literal>memory-pressure</literal> - application took enough memory to cause sufficient slowdown of other applications</para>
```
yuwata: Spurious paragraph tags. Maybe, it is better to use table?

yuwata (changes requested):

man/org.freedesktop.oom1.xml
yuwata: Not addressed? I mean, please drop `<para>...</para>` entirely.

Mc-Pain: Done. Sorry for dropping the wrong paragraph

yuwata (approved):
LGTM.

bluca:
docs are still not complete, to check run `ninja -C<builddir> update-dbus-docs` and see what's the diff

man/org.freedesktop.oom1.xml
```diff
@@ -56,12 +59,37 @@ node /org/freedesktop/oom1 {
...
+        <thead>
+          <row>
+            <entry>Reason</entry>
+            <entry>Description</entry>
+          </row>
+        </thead>
+        <tbody>
+          <row>
+            <entry>memory-used</entry>
+            <entry>application took too much memory</entry>
```
anitazha: nit: memory and swap.

Swap policy doesn't kill solely based on memory

anitazha (commented):
Thanks this looks good! This will also make it easy to expand on this in the future with pre-kill notifications.

man/org.freedesktop.oom1.xml
bluca: Also this should explicitly say that more reasons might added in the future

Mc-Pain:
Table:
- `s/memory/memory and swap/`
- Uppercase first letters in descriptions
- Add periods

man/org.freedesktop.oom1.xml
Mc-Pain: > Also this should explicitly say that more reasons might added in the future

any examples I should follow?

bluca: Something along those lines will suffice, it's just to ensure users now the list can and will be expanded in the future:

```
<para><function>Killed</function> signal is sent when any cgroup is killed by oomd.
Note that more reasons will be added in the future, and the table below will be expanded accordingly.</para>
```

bluca (approved):

PR #24211, opened 2022-08-05
mkosi: Add package libfdisk to Ubuntu dependencies

rphibel:
The libfdisk package is needed by systemd-repart.

### DIFF
--- a/mkosi.default.d/ubuntu/10-mkosi.ubuntu
+++ b/mkosi.default.d/ubuntu/10-mkosi.ubuntu
@@ -68,7 +68,7 @@ Packages=
         libqrencode4
         # We pull in the -dev package here, since the binary ones appear to change names too often, and the -dev package pulls the right deps in automatically
         libtss2-dev
-        libfdisk-dev
+        libfdisk1
         locales
         nano
         strace


mkosi.default.d/ubuntu/10-mkosi.ubuntu
```diff
@@ -68,6 +68,7 @@ Packages=
         libqrencode4
         # We pull in the -dev package here, since the binary ones appear to change names too often, and the -dev package pulls the right deps in automatically
         libtss2-dev
+        libfdisk-dev
```
bluca: mmh this is already in the BuildPackages= above so it should not be needed? -dev packages listed here are for dlopen libraries, and libfdisk is not dlopened

rphibel: Without this change, I had the following error when  launching `systemd-repart` in an image booted with `mkosi`:
```
systemd-repart: error while loading shared libraries: libfdisk.so.1: cannot open shared object file: No such file or directory
```

bluca: on which ubuntu version? which mkosi version?

rphibel: Host Ubuntu version: 20.04
Image version: 22.04

I use the mkosi in the source repository: https://github.com/systemd/mkosi/blob/main/bin/mkosi

bluca: ok, I will have a look

daandemeyer: This looks correct to me, stuff from BuildPackages is not included in the final image, so even if we pull in libfdisk-dev in `BuildPackages`; that doesn't mean it'll be available in the final image.

And using the -dev package is probably fine for the reason mentioned in the file, the name is constant unlike the actual library name.

bluca: It's not because the dependency should happen automatically because repart links to the library. The other ones are necessary because there's no linkage, they are dlopen'ed, so dh-shlibs doesn't see them.

daandemeyer: Yeah, but repart isn't shipped in ubuntu, remember :p ?

bluca: the build is not using the ubuntu specific-packaging, so that shouldn't matter

daandemeyer: No, but the only reason we don't need to do this for any of the other dependencies, is that we install systemd's package from ubuntu by default in the final image, which pulls in all the dependencies.

However, because ubuntu doesn't build systemd-repart by default, libfdisk is not installed into the final image, because it's not a dependency of ubuntu's systemd package, which is why @rphibel is seeing this issue. So that's why we're seeing this in the first place.

Remember, installing the systemd built from sources into the final is nothing more than a `meson install`. No dependencies are tracked at all that need to be installed into the final image, we just pick them up because ubuntu's systemd package is installed first into the image

daandemeyer: We should probably make this libfdisk1 though, to mirror the same thing we do in the debian config file

bluca: ahh got it - I got confused with the integration test where we rebuild the packages. We should check if anything else is missing. Yes, adding libfdisk1 would be better.

bluca (approved):

PR #18789, opened 2021-02-24
veritysetup: Add options for parity support with the cryptsetup's verity utility

gportay:
Dear Maintainers,

This PR adds support for the remaining `veritysetup` options for parity support with this implementation.

Regards,
Gaël

PR #24216, opened 2022-08-05
add ISPOWEROF2() macro and use it

poettering:
Inspired by discussion in #18789

### DIFF
--- a/src/fundamental/macro-fundamental.h
+++ b/src/fundamental/macro-fundamental.h
@@ -327,16 +327,7 @@
         })
 
 static inline size_t ALIGN_TO(size_t l, size_t ali) {
-        /* Check that alignment is exponent of 2 */
-#if SIZE_MAX == UINT_MAX
-        assert(__builtin_popcount(ali) == 1);
-#elif SIZE_MAX == ULONG_MAX
-        assert(__builtin_popcountl(ali) == 1);
-#elif SIZE_MAX == ULLONG_MAX
-        assert(__builtin_popcountll(ali) == 1);
-#else
-        #error "Unexpected size_t"
-#endif
+        assert(ISPOWEROF2(ali));
 
         if (l > SIZE_MAX - (ali - 1))
                 return SIZE_MAX; /* indicate overflow */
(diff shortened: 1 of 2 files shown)


bluca (approved):

src/fundamental/macro-fundamental.h
```diff
@@ -200,6 +200,19 @@
                 MIN(_c, z);                             \
         })
 
+/* Returns true if the passed integer is a positive power of two */
+#define CONST_ISPOWEROF2(x)                     \
+        ((x) > 0 && ((x) & ((x) - 1)) == 0)
+
+#define ISPOWEROF2(x)                                                  \
+        __builtin_choose_expr(                                         \
```
medhefgo, line 208: I doubt this does anything worthwhile. Either the compiler does evaluate ISPOWEROF2 at compile time or not. It's more interesting to do this for CONST_ISPOWEROF2 and return void in the non-const case to force compilation error.

src/fundamental/macro-fundamental.h
```diff
@@ -200,6 +200,19 @@
                 MIN(_c, z);                             \
         })
 
+/* Returns true if the passed integer is a positive power of two */
+#define CONST_ISPOWEROF2(x)                     \
+        ((x) > 0 && ((x) & ((x) - 1)) == 0)
```
medhefgo, line 205: I wonder if using popcount is faster (not that it probably matters). The ll variant could always be used as the compiler is intelligent enough to pick the smallest one.

src/fundamental/macro-fundamental.h
poettering: The only reason I use `__builtin_choose_expr()` here is so that the macro can be used inside constant and non-constant codepaths – but also guarantee side-effect free execution.

i.e. this C expression:

```c
uint8_t x[IS_POWEROF2(2)];
```

should really be a regular array, not a VLA.

src/fundamental/macro-fundamental.h
poettering: I think it's not that clear:

https://godbolt.org/z/5GjrTxjaT

popcount actually ends up doing a function call. The x&(x-1) thing does not

src/fundamental/macro-fundamental.h
medhefgo: I guess you have a point. That still allows using `CONST_ISPOWEROF2(x)` which imho, should be a compile-time error like with `ALIGN_TO/CONST_ALIGN_TO`.

PR #24222, opened 2022-08-05
network: wait for teql netdev before setting qdisc

yuwata:
And several test-network cleanups.


yuwata:
I hope the current form is good enough to merge, even though the issue reported in https://github.com/systemd/systemd-centos-ci/pull/517 is not fixed yet. PTAL!

mrc0mmand (approved):
LGTM, thank you!

PR #24224, opened 2022-08-06
shared/generator: Ensure growfs unit runs after repart

Undef-a:
When deploying an image using systemd-repart and systemd-growfs one
should have the image expanded entirely and ready to use after the first
boot. This ensures that growfs does not occur before repart, thus
requiring a second boot.

### DIFF
--- a/src/shared/generator.c
+++ b/src/shared/generator.c
@@ -537,7 +537,7 @@ int generator_hook_up_growfs(
                 "DefaultDependencies=no\n"
                 "BindsTo=%%i.mount\n"
                 "Conflicts=shutdown.target\n"
-                "After=%%i.mount\n"
+                "After=systemd-repart.service %%i.mount\n"
                 "Before=shutdown.target%s%s\n",
                 program_invocation_short_name,
                 target ? " " : "",


yuwata (approved):
LGTM. But I'll leave merging to other reviewers.

src/shared/generator.c
```diff
@@ -538,6 +538,7 @@ int generator_hook_up_growfs(
                 "BindsTo=%%i.mount\n"
                 "Conflicts=shutdown.target\n"
                 "After=%%i.mount\n"
+                "After=systemd-repart.service"
```
poettering: This cannot work, lacks `\n` at the end of the line.

That said please merge with previous line. i.e.

```
"After=systemd-repart.service %%i.mount\n"
```

poettering:
given the missing `\n` this cannot really have worked. Have you tested this? Please do!

Undef-a:
FIxed. Though interestingly I did test it in exactly the same way as I originally reproduced the issue. On the device I'm trying to get repartitioning to work reliably on the initial version worked, though I assume the `before=shutdown.target` got eaten/ignored.

bluca (approved):

Issue #24202, opened 2022-08-04
mkosi CIs get stuck in the Debian Testing and C9S jobs


mrc0mmand:
### systemd version the issue has been seen with

latest main

### Used distribution

Ubuntu Jammy

### Expected behaviour you didn't see

All mkosi CIs should pass.

### Unexpected behaviour you saw

Currently both Debian Testing and CentOS Stream 9 jobs get stuck for ~6 hours due to:

```
Run ./.github/workflows/run_mkosi.sh qemu
  ./.github/workflows/run_mkosi.sh qemu
  shell: /usr/bin/bash -e {0}
  env:
    KERNEL_CMDLINE: systemd.unit=mkosi-check-and-shutdown.service !quiet systemd.log_level=debug systemd.log_target=console udev.log_level=info systemd.default_standard_output=journal+console
‣ Running command:
‣ qemu-system-x86_64 -machine type=q35,accel=tcg,smm=on -smp 1 -m 2G -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0,id=rng-device0 -cpu max -nographic -nodefaults -serial mon:stdio -nic tap,script=no,downscript=no,ifname=vt-image,model=virtio-net-pci -drive if=pflash,format=raw,readonly=on,file=/usr/share/OVMF/OVMF_CODE.secboot.fd -global ICH9-LPC.disable_s3=1 -global driver=cfi.pflash01,property=secure,value=on -drive file=/var/tmp/.mkosi-6ww9kyhs,if=pflash,format=raw -drive if=none,id=hd,file=/home/runner/work/systemd/systemd/image.raw,format=raw -device virtio-scsi-pci,id=scsi -device scsi-hd,drive=hd,bootindex=1

3hBdsDxe: loading Boot0001 "UEFI QEMU QEMU HARDDISK " from PciRoot(0x0)/Pci(0x3,0x0)/Scsi(0x0,0x0)
BdsDxe: starting Boot0001 "UEFI QEMU QEMU HARDDISK " from PciRoot(0x0)/Pci(0x3,0x0)/Scsi(0x0,0x0)
systemd-boot@0x7D34B000,0x7D369000
!!!! X64 Exception Type - 0D(#GP - General Protection)  CPU Apic ID - 00000000 !!!!
ExceptionData - 0000000000000000
RIP  - 0000000072A7938C, CS  - 0000000000000038, RFLAGS - 0000000000000206
[...]

PR #24225, opened 2022-08-06
boot: Build with at least -O1 as workaround

medhefgo:
Fixes: #24202

### DIFF
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -199,6 +199,12 @@ efi_cflags = [
         ]
 )
 
+# On some distros, sd-boot/-stub may trigger some bug somewhere that will cause
+# kernel execution to fail. The cause seems to be purely based on code size and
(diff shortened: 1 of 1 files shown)


src/boot/efi/meson.build
```diff
@@ -199,6 +199,11 @@ efi_cflags = [
         ]
 )
 
+# On some distros, sd-boot/-stub may trigger some bug somewhere that will cause
+# kernel execution to fail. The cause seems to be purely based on code size and
+# always compiling with at least -O1 will work around that.
```
bluca, line 204: Given how strange this is, maybe link the relevant Github issue for reference here

bluca (approved):
I'll re trigger the mkosi builds a couple times to be sure

poettering:
hmm, wouldn't `-Os` be more appropriate for this, if it's about size?

Issue #21993, opened 2022-01-04
`systemd-cryptsetup`: option `keyfile-timeout=` is broken (possibly only in combination with raw device key material)


Alexander-Shukaev:
**systemd version the issue has been seen with**
 > `systemd 249 (249.7-2-arch)`

**Used distribution**
 > Arch Linux

**Linux kernel version used** (`uname -a`)

 > `Linux strix 5.15.8-arch1-1 #1 SMP PREEMPT Tue, 14 Dec 2021 12:28:02 +0000 x86_64 GNU/Linux
`

**Expected behaviour you didn't see**
 > Waiting for `/dev/disk/by-id/key-storage` should time out after `keyfile-timeout` (e.g. 5 seconds) and a fallback passphrase entry prompt should be presented.

**Unexpected behaviour you saw**
 > `/dev/disk/by-id/key-storage` received default timeout (90 seconds) instead and after it expired, the boot process is stuck forever because no fallback passphrase entry is offered.  Surfing through the web, there are examples where `keyfile-timeout` is claimed to be working as expected, all of them use true key file on some device (aka `/...:/...` or `/...:UUID=...` syntax) though.

**Steps to reproduce the problem**

Put something similar to `crypttab` for `initramfs`:

```
luks-xxx /dev/disk/by-id/luks-xxx /dev/disk/by-id/key-storage luks,header=/.../luks-xxx.header,keyfile-offset=1024,keyfile-size=1024,keyfile-timeout=5,timeout=0,password-echo=no
```

Have 2 key slots in LUKS2 header `/.../luks-xxx.header`: passphrase and key file that was written to `/dev/disk/by-id/key-storage`.  Try to boot with this setup, while `/dev/disk/by-id/key-storage` being unplugged.

Alexander-Shukaev:
Tagging potentially related stuff: #12194, #13004, #13615.

mrc0mmand:
Local reproducer (with `systemd 250 (250-292-g5f5131c)`):

```
# dd if=/dev/zero of=/crypt.img bs=1M count=32
# echo hello >/crypt.keyfile
# cryptsetup luksFormat --header /crypt.header --key-file /crypt.keyfile /crypt.img
## Add a passphrase to the next key slot
# cryptsetup luksAddKey --key-file /crypt.keyfile /crypt.header
<passphrase>
[...]

PR #24229, opened 2022-08-07
cryptsetup: support keyfile-timeout for using a device as the key file

yan12125:
Closes https://github.com/systemd/systemd/issues/21993

---

Didn't update tests as I cannot find existing tests for cryptsetup codes.


src/cryptsetup/cryptsetup-generator.c
```diff
@@ -227,7 +227,7 @@ static int generate_device_umount(const char *name,
         return 0;
 }
 
-static int print_dependencies(FILE *f, const char* device_path) {
+static int print_dependencies(FILE *f, const char* device_path, const char* timeout_value, bool canfail) {
         int r;
```
bluca, line 232: `assert(!canfail || timeout_value)`

poettering:
Looks OK. 

But looking at this, I really dislike the fact this so far has been passing around the timeout as string. Instead, this should parse the timeout with parse_usec(), then pass a usec_t around, and make a string from it as FORMAT_TIMESPAN()...

Taking unvalidated data as input and propagating it as-is to the output, sucks. We realy should parse/validate it, and then regenerate it in uniform way.

Anyway, this is an issue predating this PR. Hence I can't blame you for this, and the PR is good. But if you want extra points, it would be excellent to rework this to parse the timeout early on, and then format it back to a string where needed.

yan12125:
Yeah, indeed passing unparsed strings around is not that good. Looking into time-related functions in systemd, I feel that doing it correctly needs changes more than a few lines. I prefer such changes to be in another pull request.

bluca (approved):
That's fair, let's merge this

yan12125:
Thank you all for reviewing!

PR #24230, opened 2022-08-07
Check for multiple esp/xbootldr partitions

kreijack:
This was discussed previously in #22491.

This PR add a check for multiple esp/xbootldr partitions.
The problem is that the EFI firmware locates the ESP partition by its partition-uuid; instead the linux OS locates the ESP partition by its FS-UUID.
If a (e.g.) foreign hard disk (with a ESP partition) is mounted at boot time we don't have control about which one is selected by the EFI firmware.

The aim of this patch is to show a warning about possible duplicates in bootctl.

### DIFF
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -110,6 +110,24 @@ static const char *pick_efi_boot_option_description(void) {
         return arg_efi_boot_option_description ?: "Linux Boot Manager";
 }
 
+static void check_esp_path_duplicates(dev_t *devtptr) {
+        if (check_esp_duplicates(geteuid() != 0, devtptr) <= 1)
+                return;
+
+        printf("%s", ansi_highlight_red());
+        printf("WARNING: found more than one EFI SYSTEM partition !\n");
+        printf("%s", ansi_normal());
+}
+
+static void check_xbootldr_path_duplicates(dev_t *devtptr) {
+        if (check_xbootldr_duplicates(geteuid() != 0, devtptr) <= 1)
+                return;
+
+        printf("%s", ansi_highlight_red());
+        printf("WARNING: found more than one XBOOTLDR partition !\n");
+        printf("%s", ansi_normal());
+}
+
 static int acquire_esp(
                 bool unprivileged_mode,
                 bool graceful,
@@ -1779,6 +1797,9 @@ static int verb_status(int argc, char *argv[], void *userdata) {
 
         pager_open(arg_pager_flags);
 
+        check_esp_path_duplicates(&esp_devid);
+        check_xbootldr_path_duplicates(&xbootldr_devid);
+
         if (!arg_root && is_efi_boot()) {
                 static const struct {
                         uint64_t flag;
@@ -1972,6 +1993,10 @@ static int verb_list(int argc, char *argv[], void *userdata) {
         }
 
         pager_open(arg_pager_flags);
+
+        check_esp_path_duplicates(&esp_devid);
+        check_xbootldr_path_duplicates(&xbootldr_devid);
+
         return show_boot_entries(&config, arg_json_format_flags);
 }
 
@@ -2137,17 +2162,20 @@ static int verb_install(int argc, char *argv[], void *userdata) {
         uint32_t part = 0;
         bool install, graceful;
         int r;
+        dev_t esp_devid, xbootldr_devid;
 
         /* Invoked for both "update" and "install" */
 
         install = streq(argv[0], "install");
         graceful = !install && arg_graceful; /* support graceful mode for updates */
 
-        r = acquire_esp(/* unprivileged_mode= */ false, graceful, &part, &pstart, &psize, &uuid, NULL);
+        r = acquire_esp(/* unprivileged_mode= */ false, graceful, &part, &pstart, &psize, &uuid, &esp_devid);
         if (graceful && r == -ENOKEY)
                 return 0; /* If --graceful is specified and we can't find an ESP, handle this cleanly */
         if (r < 0)
                 return r;
+
+        check_esp_path_duplicates(&esp_devid);
 
         if (!install) {
                 /* If we are updating, don't do anything if sd-boot wasn't actually installed. */
@@ -2160,9 +2188,11 @@ static int verb_install(int argc, char *argv[], void *userdata) {
                 }
         }
 
-        r = acquire_xbootldr(/* unprivileged_mode= */ false, NULL, NULL);
+        r = acquire_xbootldr(/* unprivileged_mode= */ false, NULL, &xbootldr_devid);
         if (r < 0)
                 return r;
+
+        check_xbootldr_path_duplicates(&xbootldr_devid);
 
         r = settle_make_entry_directory();
         if (r < 0)
@@ -2228,14 +2258,19 @@ static int verb_install(int argc, char *argv[], void *userdata) {
 static int verb_remove(int argc, char *argv[], void *userdata) {
         sd_id128_t uuid = SD_ID128_NULL;
         int r, q;
+        dev_t esp_devid, xbootldr_devid;
 
-        r = acquire_esp(/* unprivileged_mode= */ false, /* graceful= */ false, NULL, NULL, NULL, &uuid, NULL);
+        r = acquire_esp(/* unprivileged_mode= */ false, /* graceful= */ false, NULL, NULL, NULL, &uuid, &esp_devid);
         if (r < 0)
                 return r;
 
-        r = acquire_xbootldr(/* unprivileged_mode= */ false, NULL, NULL);
+        check_esp_path_duplicates(&esp_devid);
+
+        r = acquire_xbootldr(/* unprivileged_mode= */ false, NULL, &xbootldr_devid);
         if (r < 0)
                 return r;
+
+        check_xbootldr_path_duplicates(&xbootldr_devid);
 
         r = settle_make_entry_directory();
         if (r < 0)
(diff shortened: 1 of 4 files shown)


poettering:
We really shouldn't check unrelated block devices. This would generate a confusing warning. i.e. just run `systemd-nspawn --image=…` in the background so that you have a loopback block device that possibly contains an ESP and you will get hammered with these warnings, event though exactly nothing is wrong.

It makes sense to warn if there's more than one ESP/XBOOTLDR partition on the "whole" block device we operate on, but we should not look at other block devices at all...

kreijack:
Good catch, I didn't thought about at the image; should we skip the /dev/loop* ?
Checking on the "whole" block what does means ? In my PC I have 5 hard disks; the problem against I want to warn is that the EFI firmware starts form a disk different than the one that hosts "/boot/efi".

poettering:
> Checking on the "whole" block what does means ? In my PC I have 5 hard disks; the problem against I want to warn is that the EFI firmware starts form a disk different than the one that hosts "/boot/efi".

I don't think we should reasonably warn about that. You cannot reasonably distinguish a disk connected "internally" form a disk connected "externally", hence we shouldn't warn about cases like this.

Consider the common case that you booted a bootable Fedora Installer image, to fix your installed OS. Suddenly this would generate warnings all over the place, because in that case you have two ESPs connected to the system.

Having multiple ESPs on *different* disks is not a reason to warn. Having multiple ESPs on the *same* disk is.

kreijack:
> You cannot reasonably distinguish a disk connected "internally" form a disk connected "externally", [...]

This is the reason why this is a **warning**: the system doesn't have enough information to classify this as error, but there is something that require the attention of the user.

> Having multiple ESPs on different disks is not a reason to warn. Having multiple ESPs on the same disk is.

Having multiple ESPs on the same disk is an **error** that the user should correct.
We can reconsidering the wording of the message in this case (not warning but error).

poettering:
Nah, I am pretty sure we should not warn about things that are quite likely not a problem at all. In my work I basically daily have more than one ESP "plugged in", and it's not a problem whatsoever.

Moreover, I think the case with multiple ESPs/XBOOTLDRs on one disk should also just be a warning, not an error (i.e. not be fatal), since it's pretty clear what usually should happen if this is the case: ignore all but the first. (I mean, I wouldn't bet that all firmware gets this right, but i am sure a good bunch will, hence warning should be good, an error not)

kreijack:
HI all,

I updated (and rebased to the latest main) the code to limit the check only on the ESP disk.
To do that, I iterate over all the partitions and compare the disk where the partition belongs to the one where ESP stays (same for XBOOTLDR).
The check is performing comparing the link sames below /sys/dev/block/<MAJOR>:<MINOR>
```
$ ls -l /sys/dev/block/8:3
lrwxrwxrwx 1 root root 0 Aug 24 19:02 /sys/dev/block/8:3 -> ../../devices/pci0000:00/0000:00:01.2/0000:01:00.1/ata2/host1/target1:0:0/1:0:0:0/block/sda/sda3
```
The partitions are on the same disks if the "second to last" in the link name are the same.

 I am not sure if it exists a bit more smart check. Suggests are welcome.

src/shared/bootspec.c
```diff
@@ -1046,6 +1052,189 @@ int boot_config_augment_from_loader(
...
+                return r;
+
+        FOREACH_DEVICE(e, d) {
+                const char *v;
+
+                r = sd_device_get_property_value(d, "ID_PART_ENTRY_TYPE", &v);
+                if (r < 0)
+                        continue;
+
+                if (devtptr) {
```
poettering, line 1197: hmm, so I take it you use a pointer here only to be able to check for NULL here?

just use a dev_t value of zero as way to indicate "unset". it's special in Linux and never assigned a real device (for that reason it is for example used as whiteout node for union mounts)

src/shared/bootspec.c
```diff
@@ -1046,6 +1052,189 @@ int boot_config_augment_from_loader(
...
+        int count = 0;
+        sd_device *d;
+
+        r = sd_device_enumerator_new(&e);
+        if (r < 0)
+                return r;
+
+        r = sd_device_enumerator_allow_uninitialized(e);
+        if (r < 0)
+                return r;
```
poettering, line 1184: why allow uninitialized devices? you check for the ID_PART_ENTRY_TYPE below which is only set for initialize devices anyway (since it comes from udev userspace), hence just drop this function call here: you enable something here you effectively ignore later on...

src/shared/bootspec.c
```diff
@@ -1046,6 +1052,189 @@ int boot_config_augment_from_loader(
...
+
+                r = sd_device_get_property_value(d, "ID_PART_ENTRY_TYPE", &v);
+                if (r < 0)
+                        continue;
+
+                if (devtptr) {
+                        dev_t devid;
+
+                        r = sd_device_get_devnum(d, &devid);
+                        if (!r && check_parent_by_devid(devid, *devtptr))
```
poettering, line 1201: !r is not really a useful check here... sd_device_get_devnum() returns < 0 on error

src/shared/bootspec.c
```diff
@@ -1046,6 +1052,189 @@ int boot_config_augment_from_loader(
...
+                if (dos_part_uuid && streq(v, dos_part_uuid))
+                        count++;
+        }
+
+        return count;
+#else
+        /* we don't have blkid, return a positive result */
+        return 1;
+#endif
+}
```
poettering, line 1170: quite frankly, drop the blkid based code path entirely. This is in a high-level, user-facing tool, i.e. udev, and is about generating a friendly warning. It's entirely fine if this only works once udev is up. 

We have blkid-based codepaths as alternative in some cases, but that's only for us in certainy very specific cases during early boot where we really need the functionality but udev is known not to be up yet. But this is noe one of those cases. Hence, please drop this function entirely, please, and use only udev, nothing else.

src/shared/bootspec.c
```diff
@@ -1046,6 +1052,189 @@ int boot_config_augment_from_loader(
...
+        if (r < 0)
+                return r;
+
+        /* bufb is in the form of '../../devices/pci0000:00/0000:00:01.2/0000:01:00.1/ata4/host3/target3:0:0/3:0:0:0/block/sde/sde3' */
+        split_last_three_by_slash(bufb, &preprelast, &prelast, &last);
+        if (strcmp(preprelast, "block"))
+                return -ENOENT;
+
+        return strcmp(prelast, aprelast);
+}
```
poettering, line 1104: Use our block_get_whole_disk(), we already have that.

src/shared/bootspec.c
```diff
@@ -1046,6 +1052,189 @@ int boot_config_augment_from_loader(
...
+
+                        r = sd_device_get_devnum(d, &devid);
+                        if (!r && check_parent_by_devid(devid, *devtptr))
+                                continue;
+                }
+
+                if (streq(v, part_uuid))
+                        count++;
+                if (dos_part_uuid && streq(v, dos_part_uuid))
+                        count++;
```
poettering, line 1208: so I am not a fan of comparing these things as strings. We generally prefer if we process parsed stuff everywhere. i.e. compare binary stuff, not string representations. Why? because string representations typically may vary for the same value, think uppercasing, prefixing with zeroes and such.

Hence, please pass around an sd_id128_t here for the gpt part type, and a uint8_t for the mbr part type. And parse the strings whenever we get them from udev via sd_id128_from_string() and safe_atou_full()

src/boot/bootctl.c
```diff
@@ -110,6 +110,24 @@ static const char *pick_efi_boot_option_description(void) {
         return arg_efi_boot_option_description ?: "Linux Boot Manager";
 }
 
+static void check_esp_path_duplicates(dev_t *devtptr) {
+        if (check_esp_duplicates(geteuid() != 0, devtptr) <= 1)
+                return;
+
+        printf("%s", ansi_highlight_red());
+        printf("WARNING: found more than one EFI SYSTEM partition !\n");
```
poettering, line 118: please output with log_warning(), this will color the usual way we color.

no space before `!`

please uppercase first char of sentence, i.e. found → Found.

And also say "on the same disk"

src/boot/bootctl.c
```diff
@@ -110,6 +110,24 @@ static const char *pick_efi_boot_option_description(void) {
...
+        printf("%s", ansi_normal());
+}
+
+static void check_xbootldr_path_duplicates(dev_t *devtptr) {
+        if (check_xbootldr_duplicates(geteuid() != 0, devtptr) <= 1)
+                return;
+
+        printf("%s", ansi_highlight_red());
+        printf("WARNING: found more than one XBOOTLDR partition !\n");
```
poettering, line 127: ditto

src/shared/bootspec.c
kreijack: Ok, this makes sense

src/shared/bootspec.c
kreijack: ok

src/shared/bootspec.c
kreijack: This means that when `!r` is true we can use `devid`, because `sd_device_get_devnum()`
returns a value less than `0` in case of error and `0` if everything is ok.

In any case I split the if in two separate check:
if (r < 0) -> error
if (check_parent_by_devid) -> ok, continue

In any case I renamed `check_parent_by_devid` as `are_the_same_disk` which is more clear.

src/shared/bootspec.c
kreijack: Ok

src/shared/bootspec.c
kreijack: Many thanks for the suggestion

src/boot/bootctl.c
kreijack: ok

src/shared/bootspec.c
kreijack: ok

PR #24234, opened 2022-08-08
hibernate-resume: The lnk pointer does not need to be initialized, an…

likunyur:
The lnk pointer has been assigned before use and does not need to be initialized, which can reduce one mov instruction.

Signed-off-by: Li kunyu <kunyu@nfschina.com>


yuwata:
No, please do not drop the initializations, otherwise the cleanup functions will free uninitialized pointers. Closing.

PR #24235, opened 2022-08-08
repart: Add support for drop-in overrides

rphibel:


src/partition/repart.c
```diff
@@ -1407,6 +1420,11 @@ static int context_read_definitions(
                 return log_error_errno(r, "Failed to enumerate *.conf files: %m");
 
         STRV_FOREACH(f, files) {
+                const char *drop_in_dir = strjoina(*f, ".d");
+                _cleanup_strv_free_ char **drop_in_files = NULL;
+                r = conf_files_list_strv(&drop_in_files, ".conf", NULL, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) STRV_MAKE(drop_in_dir));
+                if (r < 0)
+                        return log_error_errno(r, "Failed to enumerate drop-in files: %m");
```
daandemeyer: We generally don't mix declarations with statements, so first, we declare the variables, and then, we follow with the actual logic, so this stuff should move below all the variable declarations.

src/partition/repart.c
```diff
@@ -1407,6 +1420,11 @@ static int context_read_definitions(
                 return log_error_errno(r, "Failed to enumerate *.conf files: %m");
 
         STRV_FOREACH(f, files) {
+                const char *drop_in_dir = strjoina(*f, ".d");
```
daandemeyer: If you read the paragraphs on drop-in units in https://www.freedesktop.org/software/systemd/man/systemd.unit.html, you'll find that drop-ins are not simply searched in the same directory that the main config file is located in, but in all directories that config files are searched in.

So if systemd-repart is looking for configuration files in /usr/lib/repart.d/, etc/repart.d/, and /run/repart.d/, and it finds a config file /etc/repart.d/root-partition.conf, we should then look for drop-in files in /usr/lib/repart.d/root-partition.conf.d/, /etc/repart.d/root-partition.conf.d and /run/repart.d/root-partition.conf.d/.

In other words, for each file, you need to construct a strv of possible dropin directories by iterating over all possible config directories and then pass that list to `conf_files_list_strv()` to get the actual drop-in files.

src/partition/repart.c
```diff
@@ -1417,7 +1435,9 @@ static int context_read_definitions(
                 if (!p->definition_path)
                         return log_oom();
 
-                r = partition_read_definition(p, *f);
+                p->drop_in_files = strv_copy(drop_in_files);
```
daandemeyer: You can replace `strv_copy()` with `TAKE_PTR()`, this will assign `drop_in_files` to `p->drop_in_files` and then set `drop_in_files` to `NULL`. This makes sure that the strv is not freed by `_cleanup_strv_free_`.

src/partition/repart.c
```diff
@@ -1346,6 +1348,17 @@ static int partition_read_definition(Partition *p, const char *path) {
         if (r < 0)
                 return r;
 
+        STRV_FOREACH(drop_in_file, drop_in_files) {
+                r = config_parse(NULL, *drop_in_file, NULL,
```
poettering: nah, no need to hack this up manually, just use `config_parse_many_nulstr`

DaanDeMeyer (changes requested):

poettering:
config_parse_many_nulstr() will deal with the drop-on ordering logic for you. Just use that.

src/partition/repart.c
daandemeyer: That wouldn't work here right, since we don't use a fixed config file. Shouldn't it be `config_parse_many()` instead?

poettering: sure

src/shared/conf-parser.c
```diff
@@ -602,7 +603,11 @@ int config_parse_many(
         if (r < 0)
                 return r;
 
-        return config_parse_many_files(conf_files, files, sections, lookup, table, flags, userdata, ret_stats_by_path);
+        r = config_parse_many_files(conf_files, files, sections, lookup, table, flags, userdata, ret_stats_by_path);
```
daandemeyer, line 606: We only set return values on function success, so we should add an error check and return early here if `config_parse_many_files()` failed.

src/shared/conf-parser.c
```diff
@@ -602,7 +603,11 @@ int config_parse_many(
...
+
+        if (ret_dropin_files) *ret_dropin_files = TAKE_PTR(files);
```
daandemeyer: We never put the conditional and statement on the same line, so please add a newline here after the conditional.

src/shared/conf-parser.c
```diff
@@ -602,7 +603,11 @@ int config_parse_many(
...
+
+        return r;
```
daandemeyer: This can become `return 0` after the above comments are addressed.

DaanDeMeyer (changes requested):
The `config_parse_many()` change commit should also be ordered before the repart commit

src/partition/repart.c
```diff
@@ -1336,13 +1338,19 @@ static int partition_read_definition(Partition *p, const char *path) {
...
         };
         int r;
-
-        r = config_parse(NULL, path, NULL,
-                         "Partition\0",
-                         config_item_table_lookup, table,
-                         CONFIG_PARSE_WARN,
-                         p,
-                         NULL);
+        const char* filename = basename(path);
```
poettering: please do not use basename() in new code anymore. Use path_extract_filename()

src/partition/repart.c
```diff
@@ -1407,8 +1415,16 @@ static int context_read_definitions(
...
         STRV_FOREACH(f, files) {
+                char **conf_file_dirs;
+                _cleanup_strv_free_ char **drop_in_files = NULL;
                 _cleanup_(partition_freep) Partition *p = NULL;
 
+                if (directory)
+                        conf_file_dirs = STRV_MAKE(directory);
+                else {
+                        conf_file_dirs = CONF_PATHS_STRV("repart.d");
+                }
```
daandemeyer: We should reduce duplication by only calculating the list of directories to use once, since it's already done for the first call to `conf_files_list_strv()` earlier in this function as well. e.g. I'd do something like the following at the start of this function:

```
const char *const *dirs;

if (directory)
        dirs = STRV_MAKE_CONST(directory);
else
        dirs = (const char* const*) CONF_PATHS_STRV("repart.d");
```

Also remember that single line if statements should not have accompanying braces.

src/partition/repart.c
```diff
@@ -1310,7 +1312,7 @@ static int config_parse_gpt_flags(
         return 0;
 }
 
-static int partition_read_definition(Partition *p, const char *path) {
+static int partition_read_definition(Partition *p, const char *path, char **conf_file_dirs) {
```
daandemeyer: Type of `conf_file_dirs` here should already be `const char *const *`

src/partition/repart.c
```diff
@@ -1416,7 +1416,6 @@ static int context_read_definitions(
 
         STRV_FOREACH(f, files) {
                 char **conf_file_dirs;
-                _cleanup_strv_free_ char **drop_in_files = NULL;
```
daandemeyer: This should be merged into the second commit

src/partition/test-repart.sh
```diff
@@ -249,7 +249,7 @@ EOF
 
 rm -f abc
 
-systemd-repart --definitions="$D/definitions-overrides" --dry-run=yes --empty=create --size=100M --json=pretty abc >"$D/dropin-overrides"
```
daandemeyer: This should be merged into the 3rd commit

src/partition/test-repart.sh
```diff
@@ -222,3 +222,49 @@ echo "### Testing json output ###"
...
+Label=label1
+SizeMaxBytes=32M
+EOF
+
+cat >"$D/definitions-overrides/root.conf.d/override2.conf" <<EOF
+[Partition]
+Label=label2
+EOF
+
+rm -f abc
```
daandemeyer: I'd use `test-drop-in-image` as the name, that's a bit more descriptive than abc

src/partition/test-repart.sh
```diff
@@ -222,3 +222,49 @@ echo "### Testing json output ###"
...
+
+systemd-repart --definitions="$D/definitions-overrides" --dry-run=yes --empty=create --size=100M --json=pretty abc >"$D/dropin-overrides"
```
daandemeyer: You can just assign the output to a bash variable as follows instead of using a file

```
JSON_OUTPUT=$(systemd-repart --definitions="$D/definitions-overrides" --dry-run=yes --empty=create --size=100M --json=pretty abc)
```

This should only capture stdout and not stderr in the variable.

src/partition/test-repart.sh
```diff
@@ -222,3 +222,49 @@ echo "### Testing json output ###"
...
+
+cmp "$D/dropin-overrides" - <<EOF
```
daandemeyer: You can use `diff` instead of `cmp`, it provides better output when things don't match.

To pass strings to `diff`, you can do the following:

```
diff <(echo "$JSON_OUTPUT") - <<EOF
<json>
EOF
```

DaanDeMeyer (changes requested):
Looks good, a few more comments

src/partition/repart.c
```diff
@@ -1396,13 +1404,16 @@ static int context_read_definitions(
...
         assert(context);
 
         if (directory)
-                r = conf_files_list_strv(&files, ".conf", NULL, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) STRV_MAKE(directory));
+                dirs = STRV_MAKE_CONST(directory);
         else
-                r = conf_files_list_strv(&files, ".conf", root, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) CONF_PATHS_STRV("repart.d"));
+                dirs = (const char* const*) CONF_PATHS_STRV("repart.d");
+
+        r = conf_files_list_strv(&files, ".conf", NULL, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, dirs);
```
yuwata: `root` is dropped mistakenly(?).

DaanDeMeyer (approved):

yuwata (changes requested):

src/partition/repart.c
daandemeyer: I'd drop this variable since it's only used once

DaanDeMeyer (changes requested):
One more nit on top of @yuwata's comment about `root` being mistakenly dropped

src/partition/repart.c
```diff
@@ -1396,13 +1403,18 @@ static int context_read_definitions(
...
-        if (directory)
-                r = conf_files_list_strv(&files, ".conf", NULL, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) STRV_MAKE(directory));
-        else
-                r = conf_files_list_strv(&files, ".conf", root, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) CONF_PATHS_STRV("repart.d"));
+        if (directory) {
+                dirs = STRV_MAKE_CONST(directory);
+                r = conf_files_list_strv(&files, ".conf", NULL, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, dirs);
+        } else {
+                dirs = (const char* const*) CONF_PATHS_STRV("repart.d");
+                r = conf_files_list_strv(&files, ".conf", root, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, dirs);
```
daandemeyer: You can just do this outside of the `if` with a ternary operator to avoid duplication:

```
r = conf_files_list_strv(&files, ".conf", directory ? NULL : root, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, dirs);
```

DaanDeMeyer (changes requested):

src/partition/repart.c
rphibel: Yes, it was a mistake. It should be fixed now.

DaanDeMeyer (approved):

yuwata (approved):

yuwata:
```
stdout:
### Testing systemd-repart --empty=create ###
stderr:
+ set -o pipefail
+ [[ -e /dev/loop-control ]]
+ repart=/builddir/build/BUILD/systemd-251/redhat-linux-build/systemd-repart
+ udevadm=/builddir/build/BUILD/systemd-251/redhat-linux-build/udevadm
+ test -x /builddir/build/BUILD/systemd-251/redhat-linux-build/systemd-repart
+ test -x /builddir/build/BUILD/systemd-251/redhat-linux-build/udevadm
+ PATH=/builddir/.local/bin:/builddir/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin:/sbin:/usr/sbin
++ mktemp --tmpdir --directory test-repart.XXXXXXXXXX
+ D=/tmp/test-repart.51F9OxG74d
+ trap 'rm -rf '\''/tmp/test-repart.51F9OxG74d'\''' EXIT INT QUIT PIPE
+ mkdir -p /tmp/test-repart.51F9OxG74d/definitions
+ SEED=e2a40bf9-73f1-4278-9160-49c031e7aef8
+ echo '### Testing systemd-repart --empty=create ###'
+ /builddir/build/BUILD/systemd-251/redhat-linux-build/systemd-repart /tmp/test-repart.51F9OxG74d/zzz --empty=create --size=1G --seed=e2a40bf9-73f1-4278-9160-49c031e7aef8 --no-pager
/builddir/build/BUILD/systemd-251/src/partition/test-repart.sh: line 25: 60847 Segmentation fault      (core dumped) "$repart" "$D/zzz" --empty=create --size=1G --seed="$SEED" --no-pager
+ rm -rf /tmp/test-repart.51F9OxG74d
```

src/partition/repart.c
```diff
@@ -1396,13 +1403,16 @@ static int context_read_definitions(
...
+        const char *const *dirs;
 
         assert(context);
 
         if (directory)
-                r = conf_files_list_strv(&files, ".conf", NULL, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) STRV_MAKE(directory));
+                dirs = STRV_MAKE_CONST(directory);
         else
-                r = conf_files_list_strv(&files, ".conf", root, CONF_FILES_REGULAR|CONF_FILES_FILTER_MASKED, (const char**) CONF_PATHS_STRV("repart.d"));
+                dirs = (const char* const*) CONF_PATHS_STRV("repart.d");
```
yuwata: You cannot use `dirs` outside of the scope, like `alloca()`.

I think it is not necessary to optimize here. How about to simply use `strv_new()` and `strv_copy()`?

rphibel: Thanks for spotting the issue! I will try your suggestion.

daandemeyer: What about using a ternary instead of an if block? That would also fix the issue.

rphibel: Good idea. Thanks for the suggestion.

rphibel: Good idea. Thanks for the suggestion.

src/partition/repart.c
```diff
@@ -1336,13 +1338,18 @@ static int partition_read_definition(Partition *p, const char *path) {
...
         };
         int r;
-
-        r = config_parse(NULL, path, NULL,
-                         "Partition\0",
-                         config_item_table_lookup, table,
-                         CONFIG_PARSE_WARN,
-                         p,
-                         NULL);
+        const char* dropin_dirname = strjoina(basename(path), ".d");
```
poettering: please do not use basename() in new code anymore. Use path_extract_filename()

src/partition/repart.c
```diff
@@ -2058,7 +2067,8 @@ static int context_dump_partitions(Context *context, const char *node) {
                                 TABLE_UINT64, p->current_padding == UINT64_MAX ? 0 : p->current_padding,
                                 TABLE_UINT64, p->new_padding,
                                 TABLE_STRING, padding_change, TABLE_SET_COLOR, !p->partitions_next && sum_padding > 0 ? ansi_underline() : NULL,
-                                TABLE_STRING, activity ?: "unchanged");
+                                TABLE_STRING, activity ?: "unchanged",
+                                TABLE_STRV, p->drop_in_files);
```
poettering, line 2081: hmm, what's the point of displaying this? I mean, it's great we read it, but outside of debugging, I doubt this information should show up. Maybe in json output, but certainly not in the regular text output? (if you want to keep this in the json output, then hide it otherwise via table_hide_column_from_display())

poettering (changes requested):

src/partition/repart.c
daandemeyer: Does it really hurt to include it in the text output? I figure it's useful for users to be able to see all dropins that apply to some partition config

poettering: at the very least hide the column of not a single partition has drop-ins. That should suffice for me

src/partition/repart.c
```diff
@@ -1336,13 +1338,25 @@ static int partition_read_definition(Partition *p, const char *path) {
...
 
-        r = config_parse(NULL, path, NULL,
-                         "Partition\0",
-                         config_item_table_lookup, table,
-                         CONFIG_PARSE_WARN,
-                         p,
-                         NULL);
+        r = path_extract_filename(path, &filename);
+        if (r < 0)
+                return r;
```
poettering: please error log here.

src/partition/repart.c
```diff
@@ -2058,9 +2077,12 @@ static int context_dump_partitions(Context *context, const char *node) {
                                 TABLE_UINT64, p->current_padding == UINT64_MAX ? 0 : p->current_padding,
                                 TABLE_UINT64, p->new_padding,
                                 TABLE_STRING, padding_change, TABLE_SET_COLOR, !p->partitions_next && sum_padding > 0 ? ansi_underline() : NULL,
-                                TABLE_STRING, activity ?: "unchanged");
+                                TABLE_STRING, activity ?: "unchanged",
+                                TABLE_STRV, p->drop_in_files);
                 if (r < 0)
                         return table_log_add_error(r);
+
+                no_dropin_files &= strv_isempty(p->drop_in_files);
```
poettering: a bitwise `&` on a bool is weird. please use `n_d_f = n_d_f && …`

sooner or later a static analyzer will otherwise complain

poettering:
lgtm

DaanDeMeyer:
CI failure is unrelated

Issue #24161, opened 2022-07-31
SELinuxContextFromNet=yes + Accept=no get SELinux context from udevd


mikhailnov:
I have to machines with `netlabel.service` from netlabel_tools being on, SELinux context (s0, s1 etc.) is added into outgoing TCP/IP packets. I created a systemd socket unit with `SELinuxContextFromNet=yes` on machine 1. This socket unit ran a service that printed current SELinux context. I started this socket unit and ran `curl http://ip:port` on machine 2.

`curl` on machine 2 was ran under SELinux context `s0` and I expected my script to be run under `s0`, but it was run under `s0-s3`. Debugging by `strace` and `gdb` (see screenshot below) showed that **context was inherited from udevd**, I don't know why.

![2022-07-31_22-46](https://user-images.githubusercontent.com/15802528/182044124-eaf8ca91-9771-4be7-9ac4-dc816fa9d523.png)

Here is a minimal example to reproduce the issue: https://abf.io/mikhailnov/test-selinux-socket-cgi/tree/6128f96211ccda6ce3ed3a531f45aa8c60eeb7c0

Commit https://github.com/systemd/systemd/commit/16115b0a7b7cdf08fb38084d857d572d8a9088dc "socket: introduce SELinuxContextFromNet option" mentioned that SELinuxContextFromNet worked with `Accept=yes` only, but then commit https://github.com/systemd/systemd/commit/49590d67c9de85b651cac9645f090135e56d3074 "selinux: support infering SELinux label also from socket not connected to stdin" removed this.

I reworked my scripts and units to use `Accept=yes`  and `StandardInput=socket` (new tree: https://abf.io/mikhailnov/test-selinux-socket-cgi/tree/81acec51640c22b5036ff7a4000d91c883a7dca3), and now SELinux context of `curl` from machine 2 is inherited correctly (my script outputs `id` command output which prints the context).

So, the problem is that it imitates working with `Accept=no` by taking SELinux context from udev, which is obviously not expected bahaviour.

CC @msekletar , please have a look at this, both mentioned commits were yours))

systemd v243 (did not try newer)

PR #24240, opened 2022-08-08
selinux: fix SELinuxContext + SELinuxContextFromNet

mikhailnov:
It did not work correctly neither when both SELinuxContextFromNet in the socket unit
and SELinuxContext in the service unit were defined nor when only SELinuxContextFromNet
was set to true.

fcon was set only when exec_label (SELinuxContext=) was defined,
otherwise security_compute_create_raw() got null as fcon
and gave an error (Invalid arguement).

Also avoid heurestics when a specific context is set but cannot be applied
(security_compute_create_raw() works rather strange for me).

Initial implementation by commit 16115b0 made it work just like xinetd,
but xinetd does not have an alternative of SELinuxContext — a defined target context.
This patch adds propper support of it.

Fixes: 16115b0 ("socket: introduce SELinuxContextFromNet option")
See also: https://github.com/systemd/systemd/issues/24161

Tested on systemd v243.

### DIFF
--- a/src/shared/selinux-util.c
+++ b/src/shared/selinux-util.c
@@ -413,8 +413,18 @@ int mac_selinux_get_our_label(char **label) {
 
 int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *exec_label, char **label) {
 #if HAVE_SELINUX
-        _cleanup_freecon_ char *mycon = NULL, *peercon = NULL, *fcon = NULL;
-        _cleanup_context_free_ context_t pcon = NULL, bcon = NULL;
+        /* mycon — SELinux context under which systemd is running
+         * (e.g.: systemd_u:systemd_r:init_t:s0-s3.c0-c1023);
+         * peercon — SELinux context of the socket,
+         * from which we get the MLS level (the last part of the context)
+         * (e.g.: systemd_u:object_r:netlabel_peer_t:s0);
+         * targetcon — computed target context that should be set without correct MLS level;
+         * resultcon — computed target context that should be set with correct MLS level;
+         * tcon — internal target SELinux context;
+         * pcon — internal peercon SELinux context;
+         * bcon — internal SELinux context set as mycon, but with MLS level of peercon. */
+        _cleanup_freecon_ char *mycon = NULL, *peercon = NULL, *targetcon = NULL, *resultcon = NULL;
+        _cleanup_context_free_ context_t bcon = NULL, pcon = NULL, tcon = NULL;
         security_class_t sclass;
         const char *range = NULL;
         int r;
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
         if (r < 0)
                 return -errno;
 
-        if (!exec_label) {
+        /* If SELinuxContext= was set in the unit, use it */
+        if (exec_label) {
+                targetcon = strdup(exec_label);
+                if (!targetcon)
+                        return -ENOMEM;
+        } else {
                 /* If there is no context set for next exec let's use context
                    of target executable */
-                r = getfilecon_raw(exe, &fcon);
+                r = getfilecon_raw(exe, &targetcon);
                 if (r < 0)
                         return -errno;
         }
 
-        bcon = context_new(mycon);
-        if (!bcon)
+        /* If neither SELinuxContext= was set,
+         * nor a SELinux label exists on the executable file,
+         * use the context under which systemd is running */
+        if (!targetcon) {
+              targetcon = strdup(mycon);
+              if (!targetcon)
+                    return -ENOMEM;
+        }
+
+        tcon = context_new(targetcon);
+        if (!tcon)
                 return -ENOMEM;
 
+        /* Get context of the network socket */
         pcon = context_new(peercon);
         if (!pcon)
                 return -ENOMEM;
 
+        /* Extract last part of the network socket context
+         * (e.g.: systemd_u:object_r:netlabel_peer_t:s0 -> s0) */
         range = context_range_get(pcon);
         if (!range)
                 return -errno;
 
-        r = context_range_set(bcon, range);
+        /* Now change last part of previously computed targetcon to the
+         * one extracted from the network socket */
+        r = context_range_set(tcon, range);
         if (r)
                 return -errno;
-
-        freecon(mycon);
-        mycon = strdup(context_str(bcon));
-        if (!mycon)
-                return -ENOMEM;
 
         sclass = string_to_security_class("process");
         if (sclass == 0)
                 return -ENOSYS;
 
-        return RET_NERRNO(security_compute_create_raw(mycon, fcon, sclass, label));
+        resultcon = strdup(context_str(tcon));
+        if (!resultcon)
+                return -ENOMEM;
+
+        /* If SELinuxContext= was set explicitly, use it, replace MLS level
+         * with the one from socket and fail later if SELinux does not allow
+         * to transition to this context */
+        if (exec_label) {
+                *label = strdup(resultcon);
+                if (!*label)
+                      return -ENOMEM;
+        /* If heurestics were used, ask the Linux kernel which context it "recommends"
+         * (going here with explicitly set SELinuxContext= may result
+         * in its value silently being not applied) */
+        } else {
+                /* bcon is our current context, but with MLS level from socket;
+                 * required to calculate transition */
+                bcon = context_new(mycon);
+                if (!bcon)
+                        return -ENOMEM;
+                r = context_range_set(bcon, range);
+                if (r)
+                        return -errno;
+                freecon(mycon);
+                mycon = strdup(context_str(bcon));
+                if (!mycon)
+                        return -ENOMEM;
+                return RET_NERRNO(security_compute_create_raw(mycon, resultcon, sclass, label));
+        }
+
+        return 0;
 #else
         return -EOPNOTSUPP;
 #endif


src/shared/selinux-util.c
```diff
@@ -434,40 +444,82 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
         if (r < 0)
                 return -errno;
 
-        if (!exec_label) {
+        /* If SELinuxContext= was set in the unit, use it */
+        if (exec_label) {
+                targetcon = malloc(strlen(exec_label) + 1);
+                if (targetcon)
+                        strcpy(targetcon, exec_label);
```
poettering: strdup()...

also, should propagate allocation errors

src/shared/selinux-util.c
```diff
@@ -434,40 +444,82 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
 
-        bcon = context_new(mycon);
-        if (!bcon)
+        /* If neither SELinuxContext= was set,
+         * nor a SELinux label exists on the executable file,
+         * use the context under which systemd is running */
+        if (!targetcon) {
+              targetcon = malloc(strlen(mycon) + 1);
+              if (targetcon)
+                    strcpy(targetcon, mycon);
```
poettering: strdup

src/shared/selinux-util.c
```diff
@@ -434,40 +444,82 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
-        return RET_NERRNO(security_compute_create_raw(mycon, fcon, sclass, label));
+        resultcon = strdup(context_str(tcon));
+        if (!resultcon)
+                return -ENOMEM;
+
+        /* If SELinuxContext= was set explicitly, use it, replace MLS level
+         * with the one from socket and fail later if SELinux does not allow
+         * to transition to this context */
+        if (exec_label) {
+                r = asprintf(label, "%s", resultcon);
```
poettering: strdup()

mikhailnov:
@poettering Thanks, I've changed code to use strdup() and catch allocation errors where strdup was not used.

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
+        resultcon = strdup(context_str(tcon));
+        if (!resultcon)
+                return -ENOMEM;
+
+        /* If SELinuxContext= was set explicitly, use it, replace MLS level
+         * with the one from socket and fail later if SELinux does not allow
+         * to transition to this context */
+        if (exec_label) {
+                *label = strdup(resultcon);
+                if (!label)
```
mrc0mmand:
```suggestion
                if (!*label)
```

mikhailnov: thanks, fixed!

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
                 if (r < 0)
                         return -errno;
         }
 
-        bcon = context_new(mycon);
-        if (!bcon)
+        /* If neither SELinuxContext= was set,
+         * nor a SELinux label exists on the executable file,
+         * use the context under which systemd is running */
+        if (!targetcon) {
```
poettering, line 463: hmm, with your patch targetcon is always non-NULL, so why this check? what am I missing?

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
-        return RET_NERRNO(security_compute_create_raw(mycon, fcon, sclass, label));
+        resultcon = strdup(context_str(tcon));
+        if (!resultcon)
+                return -ENOMEM;
+
+        /* If SELinuxContext= was set explicitly, use it, replace MLS level
+         * with the one from socket and fail later if SELinux does not allow
+         * to transition to this context */
+        if (exec_label) {
+                *label = strdup(resultcon);
```
poettering, line 502: as per coding style: our functions should not modify return parameters on failure, and initialize all return parameters on sucess. Here, if the strdup() call fails, you have modified the return parameter `label` however.

hence strdup() into a temporary pointer first, and only on success write to `*label*

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
+         * (going here with explicitly set SELinuxContext= may result
+         * in its value silently being not applied) */
+        } else {
+                /* bcon is our current context, but with MLS level from socket;
+                 * required to calculate transition */
+                bcon = context_new(mycon);
+                if (!bcon)
+                        return -ENOMEM;
+                r = context_range_set(bcon, range);
+                if (r)
```
poettering, line 515: man page of context_range_set() says 0 or positive is returned on success. Thus this error check must be `r < 0`

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
+                        return -errno;
+                freecon(mycon);
+                mycon = strdup(context_str(bcon));
```
poettering, line 518: why make a copy of the string, if you just free it later?

poettering:
(i am reviewing just the code btw, i have not enough clue about selinux to comment anything about the ultimate merits of this work. we need someone with a better insight to review if this makes sense or not)

src/shared/selinux-util.c
mrc0mmand: Does it really? From the respective man page:

```
On failure context_*_set() functions return non-zero and 0 on success.
```

(but I read it as you as well the first couple of times, it's worded quite unfortunately)

poettering: hmm? non-zero and 0 kinda means *anything*? /me is totall confused now

mrc0mmand: As I understand it, it means != 0 on fail (*On failure context_\*_set() functions return non-zero*) and == 0 on success (*and 0 on success.*).

mikhailnov: It was like this before this patch, context range is set correctly

src/shared/selinux-util.c
mikhailnov: If exec_label is not-null, targetcon is set. Otherwise `getfilecon_raw(exe, &targetcon)` is called, but targetcon will remain null if there is no SELinux label on exe. I should probably add a comment explaining this?

Here is a small test program to make sure that it will remain null if getfilecon_raw() does not find a label:

```c
#include <selinux/selinux.h>
#include <stdio.h>
int main()
{
	char *context = NULL;
	getfilecon_raw("/tmp/t.c", &context);
	if (context) {
		printf("Context is not null\n");
	} else {
		printf("Context is null\n");
	}
}
```

I've saved it as /tmp/t.c and run it on a SELinux and a non-SELinux systems and it works as expected, printing "Context is not null" on a SELinux-enabled system and "Context is null" on a non-SELinux-enabled system.

mikhailnov: hm, there is already a comment: "nor a SELinux label exists on the executable file"

src/shared/selinux-util.c
mikhailnov: bcon is not a string (because context_range_set() works with context_t, not strings), it is context_t, but mycon is a string. A string is needed for security_compute_create_raw(), that is why I convert context_t bcon to char* mycon

src/shared/selinux-util.c
mikhailnov: Which strdup() call do you mean? There are no more strdup's after this one which writes into label. Am I missing something?

mikhailnov:
Are there any guidelines how to format comments in code? Not sure that my comments follow the required style.

yuwata:
> Are there any guidelines how to format comments in code? Not sure that my comments follow the required style.

All comments seem to be formatted in a good shape. Please see our coding style guide in docs/ dir.

src/shared/selinux-util.c
poettering: You are not just converting the context, you are then duplicating the result via strdup(), and there's really no reason to do so

src/shared/selinux-util.c
poettering: you do `*label = strdup(resultcon);` here. If this strdup() fails due OOM then you will have changed the `label` return parameter by setting it to NULL. And that's against our coding style given this is the a failure code path.

src/shared/selinux-util.c
poettering: getfilecon_raw() returns ENODATA in case no context is set. Look for that.

src/shared/selinux-util.c
mikhailnov: How do you suggest to do? If do the same in the end of the function, OOM may also happen and give the same result...

src/shared/selinux-util.c
mikhailnov: The only way how to avoid strdup() here that I see is the following:

```diff
diff --git a/src/basic/selinux-util.c b/src/basic/selinux-util.c
index 4ee3471640..23fd29ea20 100644
--- a/src/basic/selinux-util.c
+++ b/src/basic/selinux-util.c
@@ -348,11 +348,7 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
                 r = context_range_set(bcon, range);
                 if (r)
                         return -errno;
-                freecon(mycon);
-                mycon = strdup(context_str(bcon));
-                if (!mycon)
-                        return -ENOMEM;
-                r = security_compute_create_raw(mycon, resultcon, sclass, label);
+                r = security_compute_create_raw(context_str(bcon), resultcon, sclass, label);
                 if (r < 0)
                         return -errno;
         }
```
It will remove control if context_str() succeeds
Am I missing something?

src/shared/selinux-util.c
mrc0mmand: Do the `strdup()` into a temporary variable here and then assign it to the `*label`:

```c
...
if (exec_label) {
        _cleanup_free_ char *ret = NULL;

        ret = strdup(resultcon);
        if (!ret)
                return -ENOMEM;

        *label = TAKE_PTR(ret);
} ...
```

src/shared/selinux-util.c
mikhailnov: Oh, yes, this code is incorrect because "return -errno" will happen if there is no context set

src/shared/selinux-util.c
poettering: yeah, sounds ok to me.

poettering: or assigning the result to a temporary vairable `const char *bcon_str` or so, if context_str() can fail.

poettering:
Also needs a rebase

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
+                return -ENOMEM;
+
+        /* If SELinuxContext= was set explicitly, use it, replace MLS level
+         * with the one from socket and fail later if SELinux does not allow
+         * to transition to this context */
+        if (exec_label) {
+                *label = strdup(resultcon);
+                if (!*label)
+                      return -ENOMEM;
+        /* If heurestics were used, ask the Linux kernel which context it "recommends"
```
msekletar, line 505: This comment is confusing I don't understand what do you mean by heuristics, also it is misspelled.

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
+         * (going here with explicitly set SELinuxContext= may result
```
msekletar, line 506: How can we even reach the else branch here if SELinuxContext= isn't set?

src/shared/selinux-util.c
```diff
@@ -434,40 +444,84 @@ int mac_selinux_get_child_mls_label(int socket_fd, const char *exe, const char *
...
+         * in its value silently being not applied) */
+        } else {
+                /* bcon is our current context, but with MLS level from socket;
```
msekletar, line 509: Also I guess I missing the entire point of this else branch? Isn't this case already handed above by
```C
 targetcon = strdup(mycon);
```

src/shared/selinux-util.c
```diff
@@ -413,8 +413,18 @@ int mac_selinux_get_our_label(char **label) {
...
+        /* mycon — SELinux context under which systemd is running
+         * (e.g.: systemd_u:systemd_r:init_t:s0-s3.c0-c1023);
+         * peercon — SELinux context of the socket,
+         * from which we get the MLS level (the last part of the context)
+         * (e.g.: systemd_u:object_r:netlabel_peer_t:s0);
+         * targetcon — computed target context that should be set without correct MLS level;
+         * resultcon — computed target context that should be set with correct MLS level;
+         * tcon — internal target SELinux context;
+         * pcon — internal peercon SELinux context;
+         * bcon — internal SELinux context set as mycon, but with MLS level of peercon. */
```
msekletar, line 425: Need for such a comment means we are doing something wrong. What about having 3 functions like,

```C
static int mac_selinux_get_selinux_context_mls_label(const char *mls_range, char **ret);
static int mac_selinux_get_exe_mls_label(const char *mls_range, char **ret);
static int mac_selinux_get_manager_mls_label(const char *mls_range, char **ret);
```
Then we would get peer context and using that we would only call appropriate function (based on situation we are in) and return its result. This would simplify this function and would eliminated the need for this context variables.

msekletar (changes requested):

poettering:
ping?

mikhailnov:
Sorry for very big delay. I will try to deal with it.

PR #24242, opened 2022-08-08
Add option to stop idle sessions after specified timeout

msekletar:


src/login/logind-core.c
```diff
@@ -71,6 +71,8 @@ void manager_reset_config(Manager *m) {
 
         m->kill_only_users = strv_free(m->kill_only_users);
         m->kill_exclude_users = strv_free(m->kill_exclude_users);
+
+        m->stop_idle_session_usec = 0;
```
poettering: USEC_INFINITY sounds like the better value to mean "never" here

0 conceptually means "immediately" more.

src/login/logind.c
```diff
@@ -948,6 +949,71 @@ static void manager_gc(Manager *m, bool drop_not_started) {
...
+        Manager *m = userdata;
+        usec_t n, next;
+        int r;
+
+        assert(m);
+
+        if (m->stop_idle_session_usec <= 0)
+                return 0;
+
+        n = now(CLOCK_MONOTONIC);
```
poettering: use sd_event_now()

src/login/logind.h
```diff
@@ -96,6 +96,9 @@ struct Manager {
         HandleAction idle_action;
         bool was_idle;
 
+        sd_event_source *stop_idle_session_event_source;
```
poettering: I don't understand why this is a single event source on the Manager object? There should be one per Session object instead, stored in that object...

poettering (changes requested):

src/login/logind.h
msekletar: Hmm, I thought that because this is a global option for all sessions we don't really need per session evet source. I mean if we would later want to allow specifing different values for each session then we could move this down to session objects.

src/login/logind-core.c
msekletar: Ok, but I will use `config_parse_sec_fix_0` as a parser. I really want 0 to mean infinity, i.e. timeout disabled. Setting this to 0 doesn't make sense as you would be esseantially saying that you want session to be stop immediately after it started.

poettering: well, it's ok to use the parser that accepts 0, i don't mind, but I think we shold document "infinity" as a way to turn this off, and internally encode it like that...

src/login/logind.h
poettering: this really should be in the sessions. The whole reason we have sd_event_add_time() is so that we can efficiently implement a prioq of pending time events and run them all over the very same timerfd handle, which becomes very handy here. Basically you implemented a prioq yourself here in the original patch even though the one sd_event_add_time() is much nicer because more widely used.

src/shared/conf-parser.c
```diff
@@ -1875,5 +1875,36 @@ int config_parse_in_addr_non_null(
...
+
+        /* This is pretty much like config_parse_sec(), except that this treats a time of 0 as infinity, for
+         * compatibility with older versions of systemd where 0 instead of infinity was used as indicator to turn off a
+         * timeout. */
+
+        r = parse_sec_fix_0(rvalue, usec);
+        if (r < 0)
+                log_syntax(unit, LOG_WARNING, filename, line, r, "Failed to parse %s= parameter, ignoring: %s", lvalue, rvalue);
+
+        return 0;
```
poettering: I think you can replace this implementation by a single DEFINE_CONFIG_PARSE() macro invocation which generates the whole function for you for free.

man/logind.conf.xml
```diff
@@ -344,6 +344,15 @@
         are excluded from the effect of this setting. Defaults to <literal>yes</literal>.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>StopIdleSessionSec=</varname></term>
+
+        <listitem><para>Specifies the periodic timeout after which <filename>systemd-logind</filename> checks idle state
+        of all sessions. Every sessions that is idle for longer then the timeout will be stopped. Defaults to
+        <literal>infinity</literal>, i.e. <filename>systemd-logind</filename> is not checking the idle state of sessions.
```
poettering: We usually add a link to systemd.time man page for details on how to write times. WE usually also say that it takes an integer value which will be taken as seconds, or alternative any time specification that matches what systemd.time says with any time units of the user's choosing.

poettering:
Patch looks generally OK, but please, do per-session timer event sources, not a global one with a manual loop to figure out which the next idle session will be.

msekletar:
Updated. @poettering PTAL.

src/login/logind-session.c
```diff
@@ -697,6 +699,71 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+        if (idle) {
+                log_debug("Session \"%s\" of user \"%s\" is idle, stopping.", s->id, s->user->user_record->user_name);
+
+                r = session_stop(s, /* force */ true);
+                if (r < 0)
+                        log_error_errno(r, "Failed to stop idle session \"%s\": %m", s->id);
+                return r;
+        }
+
+        r = sd_event_source_set_time(source,  ts.monotonic + s->manager->stop_idle_session_usec);
```
poettering: sd_event_source_set_time_relative()

also double whitespace

src/login/logind-session.c
```diff
@@ -697,6 +699,71 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+                s->manager->event,
+                &s->stop_on_idle_event_source,
+                CLOCK_MONOTONIC,
+                s->manager->stop_idle_session_usec,
+                1,
+                session_dispatch_stop_on_idle, s);
+        if (r < 0)
+                return log_error_errno(r, "Failed to add stop on idle session event source: %m");
+
+        r = sd_event_source_set_priority(s->stop_on_idle_event_source, SD_EVENT_PRIORITY_IMPORTANT);
```
poettering: this should be done at a low priority, it's not important...

src/login/logind-session.c
```diff
@@ -697,6 +699,71 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+static int session_setup_stop_on_idle_timer(Session *s) {
+        int r;
+
+        assert(s);
+
+        if (s->manager->stop_idle_session_usec == USEC_INFINITY)
+                return 0;
+
+        r = sd_event_add_time_relative(
+                s->manager->event,
```
poettering: as per coding style, continuation lines indent at 16 spaces, please

src/login/logind-session.c
```diff
@@ -697,6 +699,71 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+                &s->stop_on_idle_event_source,
+                CLOCK_MONOTONIC,
+                s->manager->stop_idle_session_usec,
+                1,
```
poettering: this does not need high accuracy. just pass 0 here

src/login/logind-session.c
```diff
@@ -697,6 +699,71 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+                session_dispatch_stop_on_idle, s);
+        if (r < 0)
+                return log_error_errno(r, "Failed to add stop on idle session event source: %m");
+
+        r = sd_event_source_set_priority(s->stop_on_idle_event_source, SD_EVENT_PRIORITY_IMPORTANT);
+        if (r < 0)
+                return log_error_errno(r, "Failed to configure stop on idle session event source priority: %m");
+
+        r = sd_event_source_set_enabled(s->stop_on_idle_event_source, SD_EVENT_ONESHOT);
```
poettering: timer events are enabled in oneshot mode after creation anyway, no need to do this explicitly

poettering (changes requested):

src/login/logind-session.c
msekletar: I will fix double white-space. However, I will keep `sd_event_source_set_time()` because we want to schedule next wake up not in relation to the current time but to the idle hint timestamp of this session.

src/login/logind-session.c
msekletar: I suspect that feature will be used mostly it in order to be in compliance with various certification requirements. So I reckon that event priority is kind of important. Hence, let's just use the default priority instead of important.

msekletar:
Updated. @poettering PTAL.

src/login/logind-session.c
```diff
@@ -697,6 +699,58 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+        if (s->stopping)
+                return 0;
+
+        idle = session_get_idle_hint(s, &ts);
+        if (idle) {
+                log_debug("Session \"%s\" of user \"%s\" is idle, stopping.", s->id, s->user->user_record->user_name);
+
+                r = session_stop(s, /* force */ true);
+                if (r < 0)
+                        log_error_errno(r, "Failed to stop idle session \"%s\": %m", s->id);
```
poettering: acually, session_stop() already logs on all errors afaics, no need to log here again. please just drop the log message.

src/login/logind-session.c
```diff
@@ -697,6 +699,58 @@ static int session_start_scope(Session *s, sd_bus_message *properties, sd_bus_er
...
+                return r;
+        }
+
+        r = sd_event_source_set_time(source, ts.monotonic + s->manager->stop_idle_session_usec);
```
poettering: use usec_add() maybe here (i.e. saturated addition, instead of risking an overflow)

poettering:
very close.

msekletar:
Updated. @poettering PTAL, and of course thank you so much for reviewing and thoughtful comments, much appreciated!

poettering:
lgtm, thanks

poettering:
CI failures appear unrelated

PR #24244, opened 2022-08-08
sd-device-enumerator: fix sysattr match

yuwata:
Previously, if sd_device_enumerator_add_match_sysattr() is called for
the same sysattr with different values, then no device passed the filter.

Now, the accepted values (or patterns) are stored in strv, and if the
sysattr value of a device matches with the strv, then the device passes
the filter.

### DIFF
--- a/src/libsystemd/sd-device/test-sd-device.c
+++ b/src/libsystemd/sd-device/test-sd-device.c
@@ -310,6 +310,49 @@ TEST(sd_device_enumerator_filter_subsystem) {
         assert_se(n_new_dev + n_removed_dev <= 10);
 }
 
+TEST(sd_device_enumerator_add_match_sysattr) {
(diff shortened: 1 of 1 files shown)


src/libsystemd/sd-device/device-util.c
```diff
@@ -5,7 +5,68 @@
...
+                _cleanup_strv_free_ char **strv_alloc = NULL;
+                _cleanup_free_ char *key_alloc = NULL;
+
+                key_alloc = strdup(key);
+                if (!key_alloc)
+                        return -ENOMEM;
+
+                if (!value)
+                        strv_alloc = new0(char*, 1);
+                else
```
poettering: you can drop these two lines. strv_new() should be fine with a NULL argument.

Issue #24066, opened 2022-07-20
systemd-sysusers group creation logic gets confused by existing user


lucab:
### systemd version the issue has been seen with

251

### Component

systemd-sysusers

### Steps to reproduce the problem

I'm trying to port some existing systems to sysusers.d configuration files, starting from system groups entries.
In the process I seem to be reliably hitting some bug in `systemd-sysusers`, which results in groups being assigned an arbitrary dynamic GID instead of the configured one due to detected conflicts (even when `/etc/group` is wholly missing).

I haven't fully tracked down the logic responsible for this, but at least I managed to extract a minimal reproducer for the behavior:
```
# mkdir -p etc/sysusers.d
# echo 'g bin 1' > etc/sysusers.d/testcase.conf
# echo 'bin:x:1:1:bin:/bin:/sbin/nologin' > etc/passwd

# SYSTEMD_LOG_LEVEL=debug systemd-sysusers --root .
SELinux enabled state cached to: disabled
Reading config file "<...>/etc/sysusers.d/testcase.conf"…
Suggested group ID 1 for bin already used.
Creating group 'bin' with GID 999.

# cat etc/group 
bin:x:999:
```

This behavior looks buggy to me (i.e. breaking idempotence when `/etc/group` goes missing), but I'm happy to have a second pair of eyes on this.
I guess it could be related to a perceived conflict with the primary GID from the entry in the `passwd` file, but I am still reading through the code to understand what happens.

lucab:
After some debugging, I pinned down this behavior (of this specific testcase) to `gid_is_ok()` doing a **GID** existence check against the database of **UIDs**:
https://github.com/systemd/systemd/blob/8d0c34f3673e6c28eb0c5abaf639260260249099/src/sysusers/sysusers.c#L1186-L1187

Empirically, this can be confirmed by using `bin:x:1:20:bin:/bin:/sbin/nologin` as `passwd` content in the testcase above (i.e. a conflicting UID is a sufficient condition).

Overall it feels like either `gid_is_ok()` is implementing an overly aggressive logic, or that the consumer is wrongly expecting it to have a "GID presence check" semantic as mentioned in this comment:
https://github.com/systemd/systemd/blob/8d0c34f3673e6c28eb0c5abaf639260260249099/src/sysusers/sysusers.c#L1259

Related, similar logic mixing UIDs and GIDs exists in `uid_is_ok()`, with an additional layer of cross-name checking:
[...]

PR #24246, opened 2022-08-08
sysusers: only check whether the requested GID is available

lucab:
This relaxes the availability check when creating a group, if an
explicit GID has been requested.
It avoids mixing up users and groups entries with valid and unique
UIDs/GIDs, but each having the same ID number.

Fixes: https://github.com/systemd/systemd/issues/24066

### DIFF
--- a/src/sysusers/sysusers.c
+++ b/src/sysusers/sysusers.c
@@ -1169,7 +1169,7 @@ static int add_user(Item *i) {
         return 0;
 }
 
-static int gid_is_ok(gid_t gid) {
+static int gid_is_ok(gid_t gid, bool check_with_uid) {
         struct group *g;
         struct passwd *p;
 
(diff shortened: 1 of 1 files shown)


lucab:
I did cumulative manual testing of the rollup of this PR plus https://github.com/systemd/systemd/pull/24249 against the whole set of users/groups on a Fedora system (manually translated by me for now, hopefully without human errors), and I didn't spot any other apparent issues.

src/sysusers/sysusers.c
```diff
@@ -1169,6 +1169,28 @@ static int add_user(Item *i) {
         return 0;
 }
 
+static int gid_is_available(gid_t gid) {
```
poettering: hmm, so uid_is_ok() has this boolean parameter `check_with_gid`, which allows controlling whether the uid check should also imply a gid check. It appears we should do the inverse here, and thus modify gid_is_ok() to take a paramter `which_with_uid`.

lucab: Ah good point, indeed that keeps it symmetrical. I guess parameter name suggestion was a typo for `check_with_uid`.

poettering: yupp, it was. sorry for the confusion

lucab: No problem, it was clearly understandable from the context. Reworked the patch that way.

Issue #24217, opened 2022-08-05
sysusers: passwd entry gets created with the wrong primary group


lucab:
### systemd version the issue has been seen with

251

### Used distribution

Fedora 36

### Component

systemd-sysusers

### Unexpected behaviour you saw

I'm translating the default Fedora user/group entries to sysusers.d, and I hit a buggy case where the `games` user (see [allocation table](https://pagure.io/setup/blob/80eb4e6d332249616915424846f1a509f25fb7b7/f/uidgid#_21)) gets created with the wrong primary GID.
I haven't yet traced down the actual logic, but it looks like some mixup between users and groups name is involved (but not in the same way as https://github.com/systemd/systemd/issues/24066).

This is the minimal reproducer I managed to extract:
```
$ cat etc/sysusers.d/testcase.conf 
g users 100
g games 20
u games 12:100 "games" /usr/games -

$ systemd-sysusers --root .
Creating group 'users' with GID 100.
Creating group 'games' with GID 20.
Creating user 'games' (games) with UID 12 and GID 20.
```

The bug is that the passwd entry is created with a `12:20`, instead of the requested `12:100`.

As an interesting detail/workaround, changing the sysusers.d entry to use `12:users` does result in a different/correct  behavior:
```
$ cat etc/sysusers.d/testcase.conf
g users 100
g games 20
u games 12:users "games" /usr/games -

$ systemd-sysusers --root .
Creating group 'users' with GID 100.
Creating group 'games' with GID 20.
Creating user 'games' (games) with UID 12 and GID 100.
```

lucab:
I spent some time this morning following the flow of execution, and I found out that this groups mixup is coming from here:
https://github.com/systemd/systemd/blob/b622e95f2f59fcb58e23ddafed745eee26a0f52f/src/sysusers/sysusers.c#L1348-L1373

In particular, depending on the input entry this map lookup can be performed either by group-name or by user-name:
https://github.com/systemd/systemd/blob/b622e95f2f59fcb58e23ddafed745eee26a0f52f/src/sysusers/sysusers.c#L1351

In this case, we reach this point with the given `gid` but with `group_name == NULL`, and thus the map lookup logic falls back to use `i->name`. This is where the mixup between the username `games` and the separate entry for group `games` happens.

I believe it should be enough to put this logic behind a `if (!i->gid_set) { ... }` guard.
I'll try to manually test that, checking for other unexpected side-effects.

PR #24249, opened 2022-08-08
sysusers: properly process user entries with an explicit GID

lucab:
This tweaks user creation logic to properly take into consideration
an explicitly requested GID.
It fixes a bug where the creation flow would mistakenly fall back
to use the username instead, resulting in wrong lookups in case of
users and groups using the same name.

Fixes: https://github.com/systemd/systemd/issues/24217

### DIFF
--- a/src/sysusers/sysusers.c
+++ b/src/sysusers/sysusers.c
@@ -1346,9 +1346,11 @@ static int process_item(Item *i) {
         switch (i->type) {
 
         case ADD_USER: {
-                Item *j;
+                Item *j = NULL;
 
-                j = ordered_hashmap_get(groups, i->group_name ?: i->name);
+                if (!i->gid_set)
+                        j = ordered_hashmap_get(groups, i->group_name ?: i->name);
+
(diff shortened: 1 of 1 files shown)


lucab:
I did cumulative manual testing of the rollup of this PR plus https://github.com/systemd/systemd/pull/24246 against the whole set of users/groups on a Fedora system (manually translated by me for now, hopefully without human errors), and I didn't spot any other apparent issues.

src/sysusers/sysusers.c
```diff
@@ -1346,9 +1346,11 @@ static int process_item(Item *i) {
         switch (i->type) {
 
         case ADD_USER: {
-                Item *j;
+                Item *j = NULL;
+
+                if (!(i->gid_set))
```
poettering: redundant set of ()

lucab: Indeed, operator precedence is enough for this. Reworked and dropped the extra parenthesis.

PR #24251, opened 2022-08-08
Migrate cgroups to bus-locator.h

josh-gordon-fb:
Migrated cgroups.c to use new bus-locator.h bus_call_method call

### DIFF
--- a/src/core/cgroup.c
+++ b/src/core/cgroup.c
@@ -13,6 +13,7 @@
 #include "bpf-socket-bind.h"
 #include "btrfs-util.h"
 #include "bus-error.h"
+#include "bus-locator.h"
 #include "cgroup-setup.h"
 #include "cgroup-util.h"
 #include "cgroup.h"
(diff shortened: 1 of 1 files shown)


src/core/cgroup.c
```diff
@@ -2261,10 +2262,8 @@ static int unit_attach_pid_to_cgroup_via_bus(Unit *u, pid_t pid, const char *suf
         pp = strjoina("/", pp, suffix_path);
         path_simplify(pp);
 
-        r = sd_bus_call_method(u->manager->system_bus,
-                               "org.freedesktop.systemd1",
-                               "/org/freedesktop/systemd1",
-                               "org.freedesktop.systemd1.Manager",
+        r = bus_call_method(u->manager->system_bus,
+                               bus_systemd_mgr,
```
daandemeyer: Please align this line and the following ones just like it was before

src/core/cgroup.c
```diff
@@ -36,6 +36,7 @@
 #include "string-table.h"
 #include "string-util.h"
 #include "virt.h"
+#include "bus-locator.h"
```
daandemeyer: Please sort includes

DaanDeMeyer (changes requested):
We don't use the internal commit message template in systemd. Please modify the commit message to just a description of the change.

Also, the second commit needs to be merged into the first one. We want to be able to build systemd from each commit, that wouldn't work with the first commit

yuwata (approved):

PR #24252, opened 2022-08-08
Reduce sd_bus verbosity with bus functions

gtwang01:
Summary:
D-Bus functions are less verbose than sd_bus functions.
This modifies some sd_bus calls to equivalent bus calls.

Test Plan:
meson build -Dmode=developer
meson compile -C build
meson test -C build

### DIFF
--- a/src/sulogin-shell/sulogin-shell.c
+++ b/src/sulogin-shell/sulogin-shell.c
@@ -52,12 +52,12 @@ static int start_default_target(sd_bus *bus) {
 
         /* Start this unit only if we can replace basic.target with it */
         r = bus_call_method(
-                bus,
-                bus_systemd_mgr,
-                "StartUnit",
-                &error,
-                NULL,
-                "ss", SPECIAL_DEFAULT_TARGET, "isolate");
+                        bus,
+                        bus_systemd_mgr,
+                        "StartUnit",
+                        &error,
+                        NULL,
+                        "ss", SPECIAL_DEFAULT_TARGET, "isolate");
 
         if (r < 0)
                 return log_error_errno(r, "Failed to start "SPECIAL_DEFAULT_TARGET": %s", bus_error_message(&error, r));


bluca (approved):

evverx:
@gtwang01 I wonder why the "From" field in the commit message is `{{ <{{>`?

evverx:
@mrc0mmand on a somewhat related note I think https://github.com/systemd/systemd/blob/main/.github/workflows/coverity.yml#L31 should probably avoid using `--pretty=\"%aE\"` coming from random commit messages.

gtwang01:
> @gtwang01 I wonder why the "From" field in the commit message is `{{ <{{>`?

I... have no idea. Does this happen often?

evverx:
I don't think it happens very often. It should be possible to change that by running something like
```sh
git config --global user.name "firstname lastname"
git config --global user.email "realaddress@example.com"
git commit --reset-author --amend
```
and force-pushing the branch.

evverx:
@gtwang01 thanks! FWIW looking at https://github.com/systemd/systemd/pull/24251 it appears templates of some kind were used to set up `git` and `{{` and `}}` were copied verbatim even though they should have probably been removed somewhere.

src/sulogin-shell/sulogin-shell.c
```diff
@@ -52,14 +51,13 @@ static int start_default_target(sd_bus *bus) {
...
-                               &error,
-                               NULL,
-                               "ss", SPECIAL_DEFAULT_TARGET, "isolate");
+        r = bus_call_method(
+                bus,
+                bus_systemd_mgr,
+                "StartUnit",
+                &error,
+                NULL,
+                "ss", SPECIAL_DEFAULT_TARGET, "isolate");
```
yuwata: Please use 16ch indentation.

yuwata (changes requested):
Could you drop the unnecessary fields of the commit message, e.g. test plan, reviewers, and so on.

vcaputo:
It's surprising how much it pleases me to see PRs migrating code to the less verbose BusLocator-oriented API.

yuwata (approved):

DaanDeMeyer:
> It's surprising how much it pleases me to see PRs migrating code to the less verbose BusLocator-oriented API.

These make for amazing starter tasks to get new people familiar with contributing. I'm going to be rather sad once we run out of them.

Issue #24241, opened 2022-08-08
Build fails with LDFLAGS=-fuse-ld=mold and binutils 2.39


jpalus:
### systemd version the issue has been seen with

251.4

### Used distribution

PLD 3.0

### Linux kernel version used

5.19.0

### CPU architectures issue was seen on

aarch64

### Component

other

### Expected behaviour you didn't see

`systemd` builds successfully.

### Unexpected behaviour you saw

Build fails with:
```
/usr/bin/ld.bfd: warning: src/boot/efi/systemd-bootaa64.elf has a LOAD segment with RWX permissions
collect2: error: ld returned 1 exit status
```

Fix for `binutils 2.39` was already provided in #23996 however the issue with the fix is that it assumes default `cc` linker is the same as linker used for efi stub. That may not be true though -- for the most parts of `systemd` I'm using `mold` linker, while efi stub forces `bfd` in `efi_ldflags`:
https://github.com/systemd/systemd/blob/7b45d6b6f64e9f5c006bdf31559a77294dbe00ad/src/boot/efi/meson.build#L255-L256

Test for `--no-warn-rwx-segments` does not take `efi_ldflags` into account though:
https://github.com/systemd/systemd/blob/7b45d6b6f64e9f5c006bdf31559a77294dbe00ad/src/boot/efi/meson.build#L269-L273

and tests `mold` instead. In the end `bfd` fails linking because `mold` does not support `--no-warn-rwx-segments`.

### Steps to reproduce the problem

Build `systemd` with `LDFLAGS="-fuse-ld=mold"` on a system that supports efi stub and has `binutils 2.39` installed

medhefgo:
You should be able to work around this with `"-Defi-cflags=['-Wl,--no-warn-execstack','-Wl,--no-warn-rwx-segments']"`.

PR #24253, opened 2022-08-09
meson: Test correct efi linker for supported args

medhefgo:
Fixes: #24241

### DIFF
--- a/src/boot/efi/meson.build
+++ b/src/boot/efi/meson.build
@@ -266,11 +266,19 @@ efi_ldflags = [
         efi_crt0,
 ]
 
-possible_link_flags = [
-        '-Wl,--no-warn-execstack',
-        '-Wl,--no-warn-rwx-segments',
-]
-efi_ldflags += cc.get_supported_link_arguments(possible_link_flags)
+foreach arg : ['-Wl,--no-warn-execstack',
+               '-Wl,--no-warn-rwx-segments']
+        # We need to check the correct linker for supported args. This is what
+        # cc.has_multi_link_arguments() is for, but it helpfully overrides our
+        # choice of linker by putting its own -fuse-ld= arg after ours.
+        if run_command('bash', '-c',
+                       'exec "$@" -x c -o/dev/null <(echo "int main(void){return 0;}")' +
+                       ' -fuse-ld=' + efi_ld + ' -Wl,--fatal-warnings ' + arg,
+                       'bash', cc.cmd_array(),
+                       check : false).returncode() == 0
+                efi_ldflags += arg
+        endif
+endforeach
 
 if efi_arch[1] in ['aarch64', 'arm', 'riscv64']
         efi_ldflags += ['-shared']


medhefgo:
@jpalus Please give this a try.

src/boot/efi/meson.build
```diff
@@ -266,11 +266,19 @@ efi_ldflags = [
...
 
-possible_link_flags = [
-        '-Wl,--no-warn-execstack',
-        '-Wl,--no-warn-rwx-segments',
-]
-efi_ldflags += cc.get_supported_link_arguments(possible_link_flags)
+foreach arg : ['-Wl,--no-warn-execstack',
+               '-Wl,--no-warn-rwx-segments']
+        # We need to check the correct linker for supported args. This is what
+        # cc.has_multi_link_arguments() is for, but it helpfully overrides our
```
yuwata, line 272: Not a matter, but comment should be `cc.get_supported_link_arguments()`?

medhefgo: No, this is correct. `get_supported_link_arguments` tests each arg individually, but we want to test them together with `-fuse-ld=`, which `get_supported_link_arguments` does not support.

yuwata: I see. Thank you!

src/boot/efi/meson.build
```diff
@@ -266,11 +266,19 @@ efi_ldflags = [
...
+        # choice of linker by putting its own -fuse-ld= arg after ours.
+        if run_command('bash', '-c',
+                       'exec $1 -x c -o/dev/null <(echo "int main(void){return 0;}")' +
```
jpalus: Before I test I have two doubts regarding `exec $1` part:
1. `bash` man page states:
    ```
       -c        If the -c option is present, then commands are read from the first non-option argument command_string.  If
                 there are arguments after the command_string, the first argument is assigned to $0 and any remaining argu‐
                 ments  are  assigned to the positional parameters.  The assignment to $0 sets the name of the shell, which
                 is used in warning and error messages.
    ```
    so if ie `CC=gcc` isn't `gcc` end up as `$0` and `$1` will be empty?
2. What happens if ie `CC="ccache gcc"`?

medhefgo: Dunno why I wasn't using `$@`. I got lucky that meson was using `ccache cc`.

jpalus:
@medhefgo works fine. Thanks!

src/boot/efi/meson.build
bluca, line 272: Can you file an RFE for meson to support this natively? It looks like something generally useful I think?

bluca (approved):

src/boot/efi/meson.build
medhefgo: I doubt it would go trough. If I read the meson code correctly, this is because of msvc support.

/cc @eli-schwartz

bluca: maybe an optional flag or something?

src/boot/efi/meson.build
```diff
@@ -266,11 +266,19 @@ efi_ldflags = [
...
-        '-Wl,--no-warn-execstack',
-        '-Wl,--no-warn-rwx-segments',
-]
-efi_ldflags += cc.get_supported_link_arguments(possible_link_flags)
+foreach arg : ['-Wl,--no-warn-execstack',
+               '-Wl,--no-warn-rwx-segments']
+        # We need to check the correct linker for supported args. This is what
+        # cc.has_multi_link_arguments() is for, but it helpfully overrides our
+        # choice of linker by putting its own -fuse-ld= arg after ours.
+        if run_command('bash', '-c',
```
jpalus, line 274: Purely cosmetic remark: maybe it would be good idea to follow convention for standard tools and obtain path to `bash` first: `bash = find_program('bash')`?

Likely could be placed in a top-level `meson.build` next to other standard tools:
https://github.com/systemd/systemd/blob/ff71363c26008a322afe70221dd4b03364b62614/meson.build#L653-L662

Since `bash` is already implicitly required by ie `generate-*` scripts:
https://github.com/systemd/systemd/blob/ff71363c26008a322afe70221dd4b03364b62614/src/basic/generate-af-list.sh#L1
https://github.com/systemd/systemd/blob/ff71363c26008a322afe70221dd4b03364b62614/src/basic/generate-arphrd-list.sh#L1
https://github.com/systemd/systemd/blob/ff71363c26008a322afe70221dd4b03364b62614/src/basic/generate-cap-list.sh#L1
https://github.com/systemd/systemd/blob/ff71363c26008a322afe70221dd4b03364b62614/src/basic/generate-errno-list.sh#L1

medhefgo: It's not always done. And quite frankly, I fail to see the point of it. It amounts to the same thing with more work.

PR #24254, opened 2022-08-09
ci: Add mold to build tests

medhefgo:
This is on top of  #24253 to hopefully make ci happy.

### DIFF
--- a/.github/workflows/build_test.sh
+++ b/.github/workflows/build_test.sh
@@ -122,19 +122,10 @@ ninja --version
 for args in "${ARGS[@]}"; do
     SECONDS=0
 
-    # meson fails with
-    #   src/boot/efi/meson.build:52: WARNING: Not using lld as efi-ld, falling back to bfd
-    #   src/boot/efi/meson.build:52:16: ERROR: Fatal warnings enabled, aborting
-    # when LINKER is set to lld so let's just not turn meson warnings into errors with lld
-    # to make sure that the build systemd can pick up the correct efi-ld linker automatically.
-
     # The install_tag feature introduced in 0.60 causes meson to fail with fatal-meson-warnings
     # "Project targeting '>= 0.53.2' but tried to use feature introduced in '0.60.0': install_tag arg in custom_target"
     # It can be safely removed from the CI since it isn't actually used anywhere to test anything.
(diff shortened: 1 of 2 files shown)


.github/workflows/build_test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           - { COMPILER: "gcc",   COMPILER_VERSION: "12", LINKER: "gold", CRYPTOLIB: "openssl" }
           - { COMPILER: "clang", COMPILER_VERSION: "12", LINKER: "gold", CRYPTOLIB: "gcrypt"  }
           - { COMPILER: "clang", COMPILER_VERSION: "13", LINKER: "lld",  CRYPTOLIB: "openssl" }
-          - { COMPILER: "clang", COMPILER_VERSION: "14", LINKER: "bfd",  CRYPTOLIB: "auto"    }
+          - { COMPILER: "clang", COMPILER_VERSION: "14", LINKER: "mold", CRYPTOLIB: "auto"    }
```
keszybz: Hmm, I think it'd be better to drop one of the `gold` tests instead. `gold` is on life-support, while `bfd` is very commonly used, so I think we should test both compilers with `bfd`.

yuwata:
Maybe, this needs to be rebased?

medhefgo:
I think the issue is that ubuntu mold is too old. It's failing at the lto build but it works locally on my arch with 1.4.0.

```
mold: error: src/udev/ata_id.p/ata_id_ata_id.c.o: looks like this is an LLVM bitcode, but mold does not support LTO
mold: error: LLVM LTO is detected, so falling back to ld.lld
mold: execvp failed: ld.lld: No such file or directory
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

keszybz:
We get `mold amd64 1.0.3+dfsg-2`. Should this just wait until ubuntu is updated?

Also, maybe submit the second patch separately so we can merge it?

medhefgo:
> We get `mold amd64 1.0.3+dfsg-2`. Should this just wait until ubuntu is updated?

Yeah, think so.

> Also, maybe submit the second patch separately so we can merge it?

The second patch doesn't make too much sense without the first…

bluca (approved):

PR #24256, opened 2022-08-09
hwdb: Add Greaseweazle "drives" to the list of analyzers

hadess:
They're floppy disk flux readers and writers used in digital
preservation and can be broadly considered to be "analyzers" of magnetic
fluxes.

This will have the intended side-effect of giving access to the device
to users at the console, obsoleting:
https://github.com/keirf/greaseweazle/blob/master/scripts/49-greaseweazle.rules


hadess:
CC @keirf

Note that I haven't tested whether that's enough to access the functionality of the Greaseweazle just yet.

keirf:
Sounds good to me. Seems a shame that these config files can't match on usb strings. But v/p matching is clearly the order of the day here and greaseweazle has its own v/p so all good.

yuwata (approved):
Not familiar with the change, but superficially LGTM.

PR #24260, opened 2022-08-09
ci: fetch the latest Coverity build script

mrc0mmand:
Also, address https://github.com/systemd/systemd/pull/24252#issuecomment-1208747320
by using a pre-defined e-mail address stored in the GH Action secrets.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -38,8 +38,7 @@ jobs:
           #              missing shebangs)
           #   - .*\.(in|SKELETON) - all template/skeleton files
           #                         except kernel-install
-          #   - tools/coverity\.sh - external file (with some modifications)
-          FILTER_REGEX_EXCLUDE: .*/(man/.*|([^k]|k(k|ek)*([^ek]|e[^kr]))*(k(k|ek)*e?)?\.(in|SKELETON)|tools/coverity\.sh)$
+          FILTER_REGEX_EXCLUDE: .*/(man/.*|([^k]|k(k|ek)*([^ek]|e[^kr]))*(k(k|ek)*e?)?\.(in|SKELETON))$
           VALIDATE_ALL_CODEBASE: false
           VALIDATE_BASH: true
           VALIDATE_GITHUB_ACTIONS: true


mrc0mmand:
@evverx I pulled the latest version of the Coverity's Travis CI script and it consolidates the download & build scripts into one, which is nice. Also, I addressed https://github.com/systemd/systemd/pull/24252#issuecomment-1208747320 by using a pre-defined email for the Coverity notifications, which is stored in the GH Actions secrets (alongside the Coverity token) and is currently set to my email address, as I had no better idea (I wish we had some mailing list for the CI stuff).

I gave this a spin in https://github.com/systemd-ci-incubator/systemd/pull/63 with some modifications (since it doesn't like to run on PRs out-of-the-box) and it seems to be doing the right thing: https://github.com/systemd-ci-incubator/systemd/runs/7747353403?check_suite_focus=true (I fixed the fail after upload by re-applying a patch we had in the old version, which checks for HTTP code 200 instead of 201, but can't test it again, since I hit the Coverity quota: *Coverity Scan analysis NOT authorized until Tue, 09 Aug 2022 23:19:59 +0000 UTC.*).

PTAL!

.github/workflows/coverity.yml
```diff
@@ -17,27 +17,22 @@ jobs:
...
+      COVERITY_SCAN_BRANCH_PATTERN:     "${{ github.ref }}"
       COVERITY_SCAN_PROJECT_NAME:       "${{ github.repository }}"
-      # Set in repo settings -> secrets -> repository secrets
+      COVERITY_SCAN_BUILD_COMMAND:      "ninja -C cov-build"
+      # Set in repo settings -> secrets -> actions
       COVERITY_SCAN_TOKEN:              "${{ secrets.COVERITY_SCAN_TOKEN }}"
-      CURRENT_REF:                      "${{ github.ref }}"
+      COVERITY_SCAN_NOTIFICATION_EMAIL: "${{ secrets.COVERITY_SCAN_NOTIFICATION_EMAIL }}"
+      # Auxiliary variables utilized by the coverity script
+      TRAVIS_BRANCH:                    "${{ github.ref }}"
```
evverx: I'm not sure the script should keep carrying these variables. I'd just leave `COVERITY_SCAN_TOKEN` and `COVERITY_SCAN_NOTIFICATION_EMAIL` and drop everything else. More generally I'd just borrow https://github.com/dbus-fuzzer/dfuzzer/blob/master/.github/workflows/coverity.sh, replace `author_email` with `COVERITY_SCAN_NOTIFICATION_EMAIL` and also turn shellcheck on.

mrc0mmand: That sounds like a better approach, thanks. I wanted to use the Coverity script as vanilla as possible to make future updates easier, but I completely forgot you did most of the legwork already in dfuzzer regarding the script cleanup, so I'll go ahead and use that instead.

evverx: @mrc0mmand I think Travis is long gone (in terms of testing open source projects) so it's unlikely that that script will get updated updated very often (as far as I can tell it isn't even aware of the right status yet). I have been planning to replace it but I have to admit I was afraid to touch it here. Last time I tried something went wrong and I put it off :-) 

FWIW Coverity Scan will be updated on August 14 so this PR should probably be merged before or after that date to avoid weird side effects.

mrc0mmand: > @mrc0mmand I think Travis is long gone (in terms of testing open source projects) so it's unlikely that that script will get updated updated very often (as far as I can tell it isn't even aware of the right status yet). I have been planning to replace it but I have to admit I was afraid to touch it here. Last time I tried something went wrong and I put it off :-)

Hehe, yeah. And the fact you can't easily test it/debug it doesn't help either :-)

> 
> FWIW Coverity Scan will be updated on August 14 so this PR should probably be merged before or after that date to avoid weird side effects.

That's definitely my intention (as the update information was why I started digging into this).

mrc0mmand:
As discussed above, I pulled in @evverx's *enhanced and simplified™* Coverity script from dfuzzer, at it seems to be doing the right thing: https://github.com/systemd-ci-incubator/systemd/runs/7783159533?check_suite_focus=true

tools/coverity.sh
```diff
@@ -1,228 +1,62 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
evverx: Looking at ff12a7954c1 I think it would be better to keep `/usr/bin/env` since the other scripts use it.

tools/coverity.sh
```diff
@@ -1,228 +1,62 @@
...
-# These must be set by environment
-echo -e "\033[33;1mNote: COVERITY_SCAN_PROJECT_NAME and COVERITY_SCAN_TOKEN are available on Project Settings page on scan.coverity.com\033[0m"
-[ -z "$COVERITY_SCAN_PROJECT_NAME" ] && echo "ERROR: COVERITY_SCAN_PROJECT_NAME must be set" && exit 1
-[ -z "$COVERITY_SCAN_NOTIFICATION_EMAIL" ] && echo "ERROR: COVERITY_SCAN_NOTIFICATION_EMAIL must be set" && exit 1
-[ -z "$COVERITY_SCAN_BRANCH_PATTERN" ] && echo "ERROR: COVERITY_SCAN_BRANCH_PATTERN must be set" && exit 1
-[ -z "$COVERITY_SCAN_BUILD_COMMAND" ] && echo "ERROR: COVERITY_SCAN_BUILD_COMMAND must be set" && exit 1
-[ -z "$COVERITY_SCAN_TOKEN" ] && echo "ERROR: COVERITY_SCAN_TOKEN must be set" && exit 1
+    set +x # this is supposed to hide COVERITY_SCAN_TOKEN
+    echo -e "\033[33;1mDownloading Coverity Scan Analysis Tool...\033[0m"
+    wget -nv -O "$tool_archive" "$tool_url" --post-data "project=$COVERITY_SCAN_PROJECT_NAME&token=${COVERITY_SCAN_TOKEN:?}"
```
evverx, line 18: I think `:?` can be removed. `set -u` should take care of that
> If expansion is attempted on an unset variable or parameter, the shell prints an error message, and, if not interactive, exits with a non-zero status

Though "parameter null or not set" coming from `:?` seems to be a bit more clear than "unbound variable". I'm not sure :-)

.github/workflows/linter.yml
```diff
@@ -38,8 +38,7 @@ jobs:
           #              missing shebangs)
           #   - .*\.(in|SKELETON) - all template/skeleton files
           #                         except kernel-install
-          #   - tools/coverity\.sh - external file (with some modifications)
-          FILTER_REGEX_EXCLUDE: .*/(man/.*|([^k]|k(k|ek)*([^ek]|e[^kr]))*(k(k|ek)*e?)?\.(in|SKELETON)|tools/coverity\.sh)$
+          FILTER_REGEX_EXCLUDE: .*/(man/.*|([^k]|k(k|ek)*([^ek]|e[^kr]))*(k(k|ek)*e?)?\.(in|SKELETON))$
```
evverx, line 41: Nice!

tools/coverity.sh
mrc0mmand: Yeah, I added it mostly for the "if-empty" check, in case the variable is defined in the GH Action secrets, but without a value (and I guess the "better" error message is a nice side-effect).

evverx (approved):
LGTM. Thanks!

I'm not sure why both `curl` and `wget` are used in that script but it's on me and I can fix it later. I'll go ahead and merge it. Thanks!

Issue #24061, opened 2022-07-19
Allow distribution-independent sysext image matching and architecture matching


pothos:
### Component

systemd-sysext

### Is your feature request related to a problem? Please describe

A sysext image that merely contains static binaries has no dependency on the host distribution but needs to make sure it is used on the right CPU architecture.
Currently the matching rules are on the one hand too restricted and the `ID` for the distribution has to be set and `SYSEXT_LEVEL` has to be set on both, and on the other hand have no way to ensure that the right architecture is used.

### Describe the solution you'd like

The solution should support to have generic sysext images with static binaries that are not linked to a distribution and its version/`SYSEXT_LEVEL`. Architecture matching should become possible (but maybe not mandatory as it's too late already and it breaks backwards compatibility).

A new `ARCH` field should be supported in systext images to require a host CPU architecture (let's not consider to support binfmt binary translation for now). Also, it would be good to allow an empty `ID=` in sysext images and don't require `SYSEXT_LEVEL` either. One could say that in this case, at least an `ARCH` field should be required but since the sysext image could consist of shell scripts, I think it's not necessary to demand the `ARCH` in this case and it won't be mandatory elsewhere (see above).

I've created this single issue here to discuss both features at the same time instead of splitting the discussion.
If they sound good I can look into submitting patches.

### Describe alternatives you've considered

Use force-loading to skip the matching rules but it's not a good idea as it breaks those cases that rely on the matching rules because they supply multiple sysext images and only the matching one should be loaded.

bluca:
> don't require SYSEXT_LEVEL either

This is already the case, only the `ID` field is mandatory.

bluca:
> A new ARCH field should be supported in systext images to require a host CPU architecture

This makes sense to me as an optional field that if present needs to be satisfied - @poettering thoughts?

bluca:
> Also, it would be good to allow an empty ID=

I think we need to strike a balance between catching obvious errors early to provide useful feedback, and allowing flexibility. I think we should mandate the presence of an extension-release, and we should mandate that the ID field exists. Rather than allowing it to be empty, which could easily happen say if there's a build script that fails to fill in a placeholder, what about an explicit 'don't care' value? Maybe `any`? That of course works until a Linux distro called `any` appears...

pothos:
> This is already the case, only the ID field is mandatory.

That's only for the `isempty(host_os_release_version_id)` case where a rolling release host like ArchLinux has no `VERSION` set.

[...]

PR #24263, opened 2022-08-09
sysext: Support distribution-independent extensions with static binaries

pothos:
- sysext: support distribution-independent extensions using ID=any
    A sysext image that merely contains static binaries has no dependency
    on the host distribution and should be able to be used anywhere.
    Support the special 'any' value for the ID field in the extension to
    opt-out of ID and VERSION_ID/SYSEXT_LEVEL matching.
- sysext: introduce ARCH field to match host architecture
    When an extension image has binaries they should match the host
    architecture. Currently there is no way to specify this requirement.
    Introduce an ARCH field in the extension's release file that may be
    set to prevent loading on the wrong host architecture. Since this new
    field is introduced later, we don't want to make specifying it
    mandatory as it would break existing sysext images.

Closes https://github.com/systemd/systemd/issues/24061

### DIFF
--- a/man/os-release.xml
+++ b/man/os-release.xml
@@ -412,6 +412,18 @@
         </varlistentry>
 
         <varlistentry>
+          <term><varname>ARCHITECTURE=</varname></term>
+          <listitem><para>A string that specifies which CPU architecture the userspace binaries require.
+          The architecture identifiers are the same as for <varname>ConditionArchitecture=</varname>
+          described in <citerefentry><refentrytitle>systemd.unit</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
+          The field is optional and should only be used when just single architecture is supported.
+          It may provide redundant information when used in a GPT partition with a GUID type that already
+          encodes the architecture. If this is not the case, the architecture should be specified in
+          e.g., an extension image, to prevent an incompatible host from loading it.
+          </para></listitem>
+        </varlistentry>
+
+        <varlistentry>
           <term><varname>SYSEXT_LEVEL=</varname></term>
 
           <listitem><para>A lower-case string (mostly numeric, no spaces or other characters outside of 0–9,
--- a/man/systemd-sysext.xml
+++ b/man/systemd-sysext.xml
@@ -118,10 +118,14 @@
     file: the contained <varname>ID=</varname> fields have to match unless <literal>_any</literal> is set
     for the extension. If the extension <varname>ID=</varname> is not <literal>_any</literal>, the
     <varname>SYSEXT_LEVEL=</varname> field (if defined) has to match. If the latter is not defined, the
-    <varname>VERSION_ID=</varname> field has to match instead. System extensions should not ship a
-    <filename>/usr/lib/os-release</filename> file (as that would be merged into the host
-    <filename>/usr/</filename> tree, overriding the host OS version data, which is not desirable). The
-    <filename>extension-release</filename> file follows the same format and semantics, and carries the same
+    <varname>VERSION_ID=</varname> field has to match instead. If the extension defines the
+    <varname>ARCHITECTURE=</varname> field and the value is not <literal>_any</literal> it has to match the kernel's
+    architecture reported by <citerefentry><refentrytitle>uname</refentrytitle><manvolnum>2</manvolnum></citerefentry>
+    but the used architecture identifiers are the same as for <varname>ConditionArchitecture=</varname>
+    described in <citerefentry><refentrytitle>systemd.unit</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
+    System extensions should not ship a <filename>/usr/lib/os-release</filename> file (as that would be merged
+    into the host <filename>/usr/</filename> tree, overriding the host OS version data, which is not desirable).
+    The <filename>extension-release</filename> file follows the same format and semantics, and carries the same
     content, as the <filename>os-release</filename> file of the OS, but it describes the resources carried
     in the extension image.</para>
   </refsect1>
--- a/src/shared/extension-release.c
+++ b/src/shared/extension-release.c
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: LGPL-2.1-or-later */
 
 #include "alloc-util.h"
+#include "architecture.h"
 #include "env-util.h"
 #include "extension-release.h"
 #include "log.h"
@@ -15,7 +16,7 @@ int extension_release_validate(
                 const char *host_sysext_scope,
                 char **extension_release) {
 
-        const char *extension_release_id = NULL, *extension_release_sysext_level = NULL;
+        const char *extension_release_id = NULL, *extension_release_sysext_level = NULL, *extension_architecture = NULL;
 
         assert(name);
         assert(!isempty(host_os_release_id));
@@ -46,6 +47,16 @@ int extension_release_validate(
                         log_debug("Extension '%s' is not suitable for scope %s, ignoring extension.", name, host_sysext_scope);
                         return 0;
                 }
+        }
+
+        /* When the architecture field is present and not '_any' it must match the host - for now just look at uname but in
+         * the future we could check if the kernel also supports 32 bit or binfmt has a translator set up for the architecture */
+        extension_architecture = strv_env_pairs_get(extension_release, "ARCHITECTURE");
+        if (!isempty(extension_architecture) && !streq(extension_architecture, "_any") &&
+            !streq(architecture_to_string(uname_architecture()), extension_architecture)) {
+                log_debug("Extension '%s' is for architecture '%s', but deployed on top of '%s'.",
+                          name, extension_architecture, architecture_to_string(uname_architecture()));
+                return 0;
         }
 
         extension_release_id = strv_env_pairs_get(extension_release, "ID");
--- a/test/test-functions
+++ b/test/test-functions
@@ -715,7 +715,8 @@ EOF
 
         export initdir="$TESTDIR/app-nodistro"
         mkdir -p "$initdir/usr/lib/extension-release.d" "$initdir/usr/lib/systemd/system"
-        ( echo "ID=_any" ) >"$initdir/usr/lib/extension-release.d/extension-release.app-nodistro"
+        ( echo "ID=_any"
+          echo "ARCHITECTURE=_any" ) >"$initdir/usr/lib/extension-release.d/extension-release.app-nodistro"
         echo MARKER=1 >"$initdir/usr/lib/systemd/system/some_file"
         mksquashfs "$initdir" "$oldinitdir/usr/share/app-nodistro.raw" -noappend
     )


bluca:
Could you please add some very minimal test coverage for this?

The extension images for the tests are built here: https://github.com/systemd/systemd/blob/main/test/test-functions#L640
and tested here: https://github.com/systemd/systemd/blob/main/test/units/testsuite-50.sh#L302

src/shared/extension-release.c
```diff
@@ -50,17 +50,24 @@ int extension_release_validate(
...
-        if (!streq(host_os_release_id, extension_release_id)) {
+        if (!streq(host_os_release_id, extension_release_id) && !streq(extension_release_id, "any")) {
                 log_debug("Extension '%s' is for OS '%s', but deployed on top of '%s'.",
                           name, extension_release_id, host_os_release_id);
                 return 0;
         }
 
+        /* A sysext with no host OS dependency (static binaries or scripts) can match
+         * 'any' host OS, and VERSION_ID or SYSEXT_LEVEL are not required anywhere */
+        if (streq(extension_release_id, "any")) {
```
poettering: I am a bit concerned that `any` might be an ID people might use for an OS or app or so. Maybe pick something that is not as likely to collide with reguilar names, i.e. something that is not just a series of regular characters.

Maybe `_any` would suffice. or `*` or so?

bluca: `*` might be awkward to use in scripts (eg extensions build scripts)

src/shared/extension-release.c
```diff
@@ -48,6 +49,15 @@ int extension_release_validate(
                 }
         }
 
+        /* When the arch field is present and not 'any' it must match the host - for now just look at uname but in
+         * the future we could check if the kernel binfmt has something like qemu-static set up for the arch */
+        extension_arch = strv_env_pairs_get(extension_release, "ARCH");
```
poettering: hmm, this is a new field. Do we need that? What kind of images are you operating with here? Note that if you use GPT-style disk images with verity and stuff you have that same info already in the GPT partition type uuid....

I mean, I can see how this would be useful in directory/tarball images.

If we add this, I think this should be added to the full os-release primarily, and then be inherited into the extension-release thing from there.

poettering: Also, not really a fan of unnecessary abbreviations for things. There's nothing wrong with just calling the field $ARCHITECTURE

pothos: My use case are squashfs images because they are easy to build (no loop device setup needed, does even work from containers) but they don't have GPT partitions. Anyway, I don't think that the GPT is the best way to encode this and it's better to store the requirements in the release file at one place.
The os-release file of the host doesn't really need to contain this field from my perspective because what I am interested in is whether the kernel can execute these binaries, not whether the distribution uses has the libraries/binaries (e.g., a multiarch debian would have both x86 and x86-64). This reminds me that besides binfmt we can also handle the case of 32 bit x86 binaries being executable on a 64 bit kernel but that's not the most important thing right now.
Using `ARCHITECTURE` is good for me, too.

poettering: > My use case are squashfs images because they are easy to build (no loop device setup needed, does even work from containers) but they don't have GPT partitions.

Hmm, what? You are mounting squashfs images without loopback devices? How does that work?  Through some fuse mess?

poettering: > The os-release file of the host doesn't really need to contain this field from my perspective because what I am interested in is whether the kernel can execute these binaries, not whether the distribution uses has the libraries/binaries (e.g., a multiarch debian would have both x86 and x86-64). 

Well, it should be an optional field for environments where this is desirable. I am pretty sure extension-release and os-release should really be developed side-by-side, and hence have the same fields, and extension-release should "import" fields from os-release wherever that makes sense.

os-release is used for nspawn-style container images and for portable services too. It might make sense to encode the architecture there, if not operating on a GPT image.

I understand you are using raw squashfs without a gpt envelope. Do note that once you intend to support verity (i.e. signing your stuff sensibly) you'll notice that this will be too simple.

> This reminds me that besides binfmt we can also handle the case of 32 bit x86 binaries being executable on a 64 bit kernel but that's not the most important thing right now.

Sure, but that's a question orthogonal to the question if ARCHITECTURE= makes sense to declare what an image is for. I think it does make sense for images not doing GPT. And hence if this is added at all it should be added and defined for "os-release", and then imported into the spec for "extension-release". But it should be an optional field either way, since on GPT images it's kinda useless, and there are multi-arch images thinkable.

bluca: > > My use case are squashfs images because they are easy to build (no loop device setup needed, does even work from containers) but they don't have GPT partitions.
> 
> Hmm, what? You are mounting squashfs images without loopback devices? How does that work? Through some fuse mess?

To `build` images - we are doing the same. Building a squashfs is an entirely-userspace operation requiring no kernel involvement, root privileges, access to blockdev/loopdev, and so on, which is very very convenient.

bluca: > I understand you are using raw squashfs without a gpt envelope. Do note that once you intend to support verity (i.e. signing your stuff sensibly) you'll notice that this will be too simple.

We actually do verity with squashfs images just fine. Yes it's annoying to carry over 4 artifacts instead of one, but it is not a show-stopper.

poettering: > To `build` images - we are doing the same. Building a squashfs is an entirely-userspace operation requiring no kernel involvement, root privileges, access to blockdev/loopdev, and so on, which is very very convenient.

But so does wrapping them in GPT? What am I missing?

poettering: > We actually do verity with squashfs images just fine. Yes it's annoying to carry over 4 artifacts instead of one, but it is not a show-stopper.

For new stuff I'd advise maybe not making your life unnecessarily complex.

bluca: > > To `build` images - we are doing the same. Building a squashfs is an entirely-userspace operation requiring no kernel involvement, root privileges, access to blockdev/loopdev, and so on, which is very very convenient.
> 
> But so does wrapping them in GPT? What am I missing?

Without a loop device? How do you copy the squashfs to the partition?

> > We actually do verity with squashfs images just fine. Yes it's annoying to carry over 4 artifacts instead of one, but it is not a show-stopper.
> 
> For new stuff I'd advise maybe not making your life unnecessarily complex.

Sure, depending on requirements - but the point is that verity does fully work in that case too

poettering: > Without a loop device? How do you copy the squashfs to the partition?

You can either use a ready-made tool like systemd-repart which will do that for you. You can alternatively script something simple with sfdisk+dd if you like, it's not that hard: just define the minimal gpt table in sfdisk textual representation, with the right size, then call sfdisk to make an image file from it, then use sfdisk again, to dump the resulting full gpt table in text format, parse out the start offset, and dd the squashfs into that offset. dd/sfdisk operate just fine on regular files. Works without any privs.

bluca: Right - or you do `mksquashfs dir/` and then use the rest of the time to have a nap instead :-)

pothos: > Hmm, what? You are mounting squashfs images without loopback devices? How does that work? Through some fuse mess?

I meant creation of squashfs images with mksquashfs makes extensions easy to build because no mounting is involved. (As you both continued to discuss ;)

pothos: Does making `ARCHITECTURE` a field for `os-release` also mean that we should use it instead of `uname`'s arch to match against the extension's field? In this case it wouldn't be very useful for distro-independent images because each distro's `os-release` would need to have it then. I'm fine adding it to the spec if we don't use the host's `os-release` `ARCHITECTURE` in the sysext check but use the `uname` arch instead. Where does the spec for `os-release` live (I guess `man/os-release.xml` - and do I need to know/do something else than documenting it there?

poettering: No, I think ARCHITECTURE is decoration, not more. i.e. a property you can use if you have nothing better, but things like uname() should always win, because they *actually* say what is in effect.

src/shared/extension-release.c
pothos: I went with `_any`

src/shared/extension-release.c
pothos: Added docs for the field to the os-release man page

bluca (approved):

bluca:
test is failing on this:

```
17:44:21 [   58.911421] testsuite-50.sh[1220]: + systemd-run -P --property ExtensionImages=/usr/share/app-nodistro.raw --property RootImage=/tmp/tmp.ZLDBf2/minimal_0.raw cat /usr/lib/systemd/system/other_file
17:44:21 [   58.914980] testsuite-50.sh[1221]: + grep -q -F MARKER=1
```

src/shared/extension-release.c
poettering: > Right - or you do `mksquashfs dir/` and then use the rest of the time to have a nap instead :-)

Well, it's called technical debt, if you lock yourself into a world, where any extension means shipping/downloading more stuff. I mean, you are patching systemd here to support something you might as well achieve by just wrapping your stuff in a proper GPT envelope, that makes things auto-detectable, recognizable and self-descriptive. The fact that this PR was filed to basically add inside the image what we already support via the GPT envelope shows that you are making your changes at the wrong place.

I mean, adding more C code to get a crappier result where you could just add 4 lines of shell instead to give you a better result sounds like an awful choice to me.

@pothos is adding new infra to flatcaros, and I'd recommend to do that in way that ensures you keep things simple and extensible for the future, so that you have images that carry verity and pkcs7 inside them and whatever else needs to be added sooner or later. Doing this in sidecar files is just a mess, you create an artificial need to download many resources, sync them up and so on where you could just have a single blob instead, and that in an extensible way.

man/os-release.xml
```diff
@@ -411,6 +411,15 @@
           determines the fallback hostname.</para></listitem>
         </varlistentry>
 
+        <varlistentry>
+          <term><varname>ARCHITECTURE=</varname></term>
+          <listitem><para>A string that specifies which CPU architecture the binaries require. The architecture
+          is not exactly in the format reported by <citerefentry><refentrytitle>uname</refentrytitle><manvolnum>2</manvolnum></citerefentry>
```
poettering: "not exactly"? actually, they are quite different.

Please drop the references to hostname/uname. Just point the user to ConditionArchitecture= in systemd.unit(5) man page

man/os-release.xml
```diff
@@ -411,6 +411,15 @@
...
+          but uses the format as shown by <citerefentry><refentrytitle>hostnamectl</refentrytitle><manvolnum>1</manvolnum></citerefentry>.
+          An architecture should be specified in extensions images to prevent an incompatible host from loading it.
+          </para></listitem>
```
poettering, line 423: Please conditionalize this a bit, i.e. say that this is an optional field and only applies to systems that are intended for a single architecture only. (because debian intends to support true multi-arch, where an single image might be for multiple archs actually)

Also say that this is redundant information, as the on GPT disk and images the same information can likely be derived from the partition type UUID the OS is stored in.

Also say that this is supposed to be the userspace architecture. Sparc Linux was traditionally a 64bit kernel with a 32bit userspace, and this here is strictly about userspace.

src/shared/extension-release.c
```diff
@@ -50,17 +50,24 @@ int extension_release_validate(
...
                           name, extension_release_id, host_os_release_id);
                 return 0;
         }
 
+        /* A sysext with no host OS dependency (static binaries or scripts) can match
+         * '_any' host OS, and VERSION_ID or SYSEXT_LEVEL are not required anywhere */
+        if (streq(extension_release_id, "_any")) {
+                log_debug("Extension '%s' matches '_any' OS.", name);
+                return 1;
+        }
```
poettering, line 74: If you move this if block before the preceeding if block then you can simplify the condition in that one..

man/os-release.xml
pothos: Thanks, that's the perfect reference I wanted to have there but failed to come up with ;) Will do.

src/shared/extension-release.c
pothos: Maybe to add, I do this here with end users or third-party software distributors in mind. My main motivation for having ARCHITECTURE in the release file was to have everything in one place and also cover the squashfs/dir case. I agree, dm-verity is great, GPT partition GUIDs are also great (one could even bundle multiple architectures into a single image, each having its own partition - but it won't work right now as written below), and for the reasons you mentioned, both should be recommended over plain fs images. For beginners a quick squashfs is nice and I imagine people may prefer if over the verity dance - if we want to push users to verity, we have to improve the docs a lot from my point of view, even how to use GPT with sysext is not really documented (i.e., use the root fs GUID of the arch for the partition type and not something else).
Related to this: Currently the GPT GUID for the partition type is not checked against the host architecture for a loading the sysext image, so as of now it's not an alternative to the ARCHITECTURE field, should I open an issue about that?

man/os-release.xml
pothos: Thanks, done, please check the wording again

man/systemd-sysext.xml
```diff
@@ -115,8 +115,9 @@
     <para>A simple mechanism for version compatibility is enforced: a system extension image must carry a
     <filename>/usr/lib/extension-release.d/extension-release.<replaceable>$name</replaceable></filename>
     file, which must match its image name, that is compared with the host <filename>os-release</filename>
-    file: the contained <varname>ID=</varname> fields have to match, as well as the
-    <varname>SYSEXT_LEVEL=</varname> field (if defined). If the latter is not defined, the
+    file: the contained <varname>ID=</varname> fields have to match unless <literal>_any</literal> is set
+    for the extension. If the extension <varname>ID</varname> is not <literal>_any</literal>, the
```
yuwata: missing "=": `If the extension <varname>ID=</varname> is not`

src/shared/extension-release.c
```diff
@@ -48,6 +49,15 @@ int extension_release_validate(
                 }
         }
 
+        /* When the architecture field is present and not '_any' it must match the host - for now just look at uname but in
+         * the future we could check if the kernel also supports 32 bit or binfmt has a translator set up for the architecture */
+        extension_architecture = strv_env_pairs_get(extension_release, "ARCHITECTURE");
+        if (!isempty(extension_architecture) && !streq(extension_architecture, "_any") && !streq(architecture_to_string(uname_architecture()), extension_architecture)) {
```
yuwata: The line is slightly long. Could you wrap the line?

yuwata (changes requested):
LGTM. But two minor comments.

man/systemd-sysext.xml
pothos: Thanks, done

yuwata (approved):

src/shared/extension-release.c
pothos: Filed an issue for that here: https://github.com/systemd/systemd/issues/24318

poettering: > Related to this: Currently the GPT GUID for the partition type is not checked against the host architecture for a loading the sysext image, so as of now it's not an alternative to the ARCHITECTURE field, should I open an issue about that?

Well, it's used for searching the root fs in the image. We wouldn't even find the partition in the image if the gpt type uuid wouldn't exist.

PR #24264, opened 2022-08-09
mkosi settings tweaks

daandemeyer:


bluca (approved):

mkosi.default.d/10-systemd.conf
```diff
@@ -13,7 +13,6 @@ OutputDirectory=mkosi.output
 [Content]
 BuildDirectory=mkosi.builddir
 Cache=mkosi.cache
-InstallDirectory=mkosi.installdir
```
yuwata, line 15: Not familiar with mkosi, but is dropping the line safe? E.g. this does not break the host system, right?

daandemeyer: Definitely not, mkosi will never install things to the host system. This is just there to allow capturing the output of what's installed to the OS image

yuwata: Thanks.

PR #24265, opened 2022-08-09
udev: net_id: Use devicetree aliases when available

### LINKED ISSUE #17625: udevd net_id builtin does not identify platform eth on Raspberry Pi 4b
```
udevadm info /sys/class/net/eth0
P: /devices/platform/scb/fd580000.ethernet/net/eth0
L: 0
E: DEVPATH=/devices/platform/scb/fd580000.ethernet/net/eth0
E: INTERFACE=eth0
E: IFINDEX=2
E: SUBSYSTEM=net
E: USEC_INITIALIZED=2468495
E: ID_MM_CANDIDATE=1
E: ID_NET_NAMING_SCHEME=v245
E: ID_NET_NAME_MAC=enxdca63237faa6
E: ID_OUI_FROM_DATABASE=Raspberry Pi Trading Ltd
E: ID_PATH=platform-fd580000.ethernet
E: ID_PATH_TAG=platform-fd580000_ethernet
E: ID_NET_DRIVER=bcmgenet
E: ID_NET_LINK_FILE=/usr/lib/systemd/network/99-default.link
E: SYSTEMD_ALIAS=/sys/subsystem/net/devices/eth0
E: TAGS=:systemd:
```

Looking at names_platform, the syspath for the device neither matches _PLATFORM_PATTERN4 nor _PLATFORM_PATTERN3.

It would be nice if `/sys/devices/platform/scb/fd580000.ethernet/net/eth0` was recognized and something sensible got constructed out of it.

sean-anderson-seco:
Devicetree firmware contains an "aliases" node, containing various
aliases for devices described by the firmware. For ethernet devices,
these are named "ethernet0", "ethernet1", etc. They provide a convenient
means of numbering ethernet devices, especially on systems with no other
stable number other than the address. In particular, U-Boot already uses
these aliases to name its ethernet devices.

Previously, there have been attempts (such as [1]) to add support for
these aliases to Linux. However, these patches have been rejected
because it is the maintainers' view that naming policy be left to
userspace. Well, systemd is userspace, so here we are.

In terms of implementation, apparently there can be multiple device
trees at once. I have decided to dodge this problem for now, and just
use /proc/device-tree. If it is desired to support multiple device trees
later, then the scheme can be modified to something like enb2d3.

For the moment we only support "ethernetX" aliases. Future patches might
want to also handle "canX" and "wifiX".

It is common on boards with only one ethernet device to use an alias of
just "ethernet". In this case, the index is an implicit 0. In case the
author of the firmware made a mistake, we check to ensure that aliases
of "ethernet" and "ethernet0" do not both exist.

[1] https://patchwork.kernel.org/project/linux-arm-kernel/patch/1399390594-1409-1-git-send-email-boris.brezillon@free-electrons.com/

Signed-off-by: Sean Anderson <sean.anderson@seco.com>

Closes #17625.

### DIFF
--- a/src/udev/udev-builtin-net_id.c
+++ b/src/udev/udev-builtin-net_id.c
@@ -23,6 +23,7 @@
 #include <linux/pci_regs.h>
 
 #include "alloc-util.h"
+#include "chase-symlinks.h"
 #include "device-util.h"
 #include "dirent-util.h"
 #include "fd-util.h"
@@ -52,6 +53,7 @@ typedef enum NetNameType {
         NET_XENVIF,
         NET_PLATFORM,
         NET_NETDEVSIM,
+        NET_DEVICETREE,
 } NetNameType;
 
 typedef struct NetNames {
@@ -70,6 +72,7 @@ typedef struct NetNames {
         char xen_slot[ALTIFNAMSIZ];
         char platform_path[ALTIFNAMSIZ];
         char netdevsim_path[ALTIFNAMSIZ];
+        char devicetree_onboard[ALTIFNAMSIZ];
 } NetNames;
 
 /* skip intermediate virtio devices */
@@ -600,6 +603,90 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
         return 0;
 }
 
+static int dev_devicetree_onboard(sd_device *dev, NetNames *names) {
+        const char *alias, *ofnode_path, *ofnode_syspath;
+        _cleanup_free_ char *devicetree_syspath = NULL;
+        _cleanup_(sd_device_unrefp) sd_device *aliases_dev = NULL, *ofnode_dev = NULL;
+        sd_device *parent;
+        int r;
+
+        if (!naming_scheme_has(NAMING_DEVICETREE_ALIASES))
+                return 0;
+
+        /* check if our direct parent has an of_node */
+        r = sd_device_get_parent(dev, &parent);
+        if (r < 0)
+                return r;
+
+        r = sd_device_new_child(&ofnode_dev, parent, "of_node");
+        if (r < 0)
+                return r;
+
+        r = sd_device_get_syspath(ofnode_dev, &ofnode_syspath);
+        if (r < 0)
+                return r;
+
+        r = chase_symlinks("/proc/device-tree", NULL, 0, &devicetree_syspath, NULL);
+        if (r < 0)
+                return r;
+
+        /*
+         * Example paths:
+         * devicetree_syspath = /sys/firmware/devicetree/base
+         * ofnode_syspath = /sys/firmware/devicetree/base/soc/ethernet@deadbeef
+         * ofnode_path = soc/ethernet@deadbeef
+         */
+        ofnode_path = path_startswith(ofnode_syspath, devicetree_syspath);
+        if (!ofnode_path)
+                return -ENOENT;
+
+        /* Get back our leading / to match the contents of the aliases */
+        ofnode_path--;
+        assert(path_is_absolute(ofnode_path));
+
+        r = sd_device_new_from_syspath(&aliases_dev, strjoina(devicetree_syspath, "/aliases"));
+        if (r < 0)
+                return r;
+
+        FOREACH_DEVICE_SYSATTR(aliases_dev, alias) {
+                const char *alias_path, *alias_index, *conflict;
+                unsigned i;
+
+                alias_index = startswith(alias, "ethernet");
+                if (!alias_index)
+                        continue;
+
+                if (sd_device_get_sysattr_value(aliases_dev, alias, &alias_path) < 0)
+                        continue;
+
+                if (!path_equal(ofnode_path, alias_path))
+                        continue;
+
+                /* If there's no index, we default to 0... */
+                if (isempty(alias_index)) {
+                        i = 0;
+                        conflict = "ethernet0";
+                } else {
+                        r = safe_atou(alias_index, &i);
+                        if (r < 0)
+                                return log_device_debug_errno(dev, r,
+                                                "Could not get index of alias %s: %m", alias);
+                        conflict = "ethernet";
+                }
+
+                /* ...but make sure we don't have an alias conflict */
+                if (i == 0 && sd_device_get_sysattr_value(aliases_dev, conflict, NULL) >= 0)
+                        return log_device_debug_errno(dev, SYNTHETIC_ERRNO(EEXIST),
+                                        "Ethernet alias conflict: ethernet and ethernet0 both exist");
+
+                xsprintf(names->devicetree_onboard, "d%u", i);
+                names->type = NET_DEVICETREE;
+                return 0;
+        }
+
+        return -ENOENT;
+}
+
 static int names_pci(sd_device *dev, const LinkInfo *info, NetNames *names) {
         _cleanup_(sd_device_unrefp) sd_device *physfn_pcidev = NULL;
         _cleanup_free_ char *virtfn_suffix = NULL;
@@ -1034,6 +1121,15 @@ static int builtin_net_id(sd_device *dev, sd_netlink **rtnl, int argc, char *arg
                                  special_glyph(SPECIAL_GLYPH_ARROW_RIGHT), str + strlen(prefix));
 
                 ieee_oui(dev, &info, test);
+        }
+
+        /* get devicetree aliases; only ethernet supported for now  */
+        if (streq(prefix, "en") && dev_devicetree_onboard(dev, &names) >= 0 &&
+            names.type == NET_DEVICETREE) {
+                char str[ALTIFNAMSIZ];
+
+                if (snprintf_ok(str, sizeof str, "%s%s", prefix, names.devicetree_onboard))
+                        udev_builtin_add_property(dev, test, "ID_NET_NAME_ONBOARD", str);
         }
 
         /* get path names for Linux on System z network devices */
--- a/man/systemd.net-naming-scheme.xml
+++ b/man/systemd.net-naming-scheme.xml
@@ -98,10 +98,36 @@
     <variablelist>
         <varlistentry>
           <term><varname>ID_NET_NAME_ONBOARD=<replaceable>prefix</replaceable><constant>o</constant><replaceable>number</replaceable></varname></term>
+          <term><varname>ID_NET_NAME_ONBOARD=<replaceable>prefix</replaceable><constant>d</constant><replaceable>number</replaceable></varname></term>
 
           <listitem><para>This name is set based on the numeric ordering information given by the firmware
-          for on-board devices. The name consists of the prefix, letter <constant>o</constant>, and a number
-          specified by the firmware. This is only available for PCI devices.</para>
+          for on-board devices. Different schemes are used depending on the fiemware type, as described in the table below.</para>
+
+            <table>
+              <title>Onboard naming schemes</title>
+
+              <tgroup cols='2'>
+                <thead>
+                  <row>
+                    <entry>Format</entry>
+                    <entry>Description</entry>
+                  </row>
+                </thead>
+
+                <tbody>
+                  <row>
+                    <entry><replaceable>prefix</replaceable><constant>o</constant><replaceable>number</replaceable></entry>
+                    <entry>PCI onboard index</entry>
+                  </row>
+
+                  <row>
+                    <entry><replaceable>prefix</replaceable><constant>d</constant><replaceable>number</replaceable></entry>
+                    <entry>Devicetree alias index</entry>
+                  </row>
+
+                </tbody>
+              </tgroup>
+            </table>
           </listitem>
         </varlistentry>
 
@@ -412,6 +438,13 @@
           this is relaxed and we will use slot information to generate the name based on it but only if
           the PCI device has multiple functions. This is safe because distinct function number is a part of
           the device name for multifunction devices.</para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term><constant>v252</constant></term>
+
+          <listitem><para>Added naming scheme for platform devices with devicetree aliases.</para>
           </listitem>
         </varlistentry>
 
--- a/src/shared/netif-naming-scheme.c
+++ b/src/shared/netif-naming-scheme.c
@@ -24,6 +24,7 @@ static const NamingScheme naming_schemes[] = {
         { "v249", NAMING_V249 },
         { "v250", NAMING_V250 },
         { "v251", NAMING_V251 },
+        { "v252", NAMING_V252 },
         /* … add more schemes here, as the logic to name devices is updated … */
 
         EXTRA_NET_NAMING_MAP
--- a/src/shared/netif-naming-scheme.h
+++ b/src/shared/netif-naming-scheme.h
@@ -37,6 +37,7 @@ typedef enum NamingSchemeFlags {
         NAMING_REPLACE_STRICTLY          = 1 << 12, /* Use udev_replace_ifname() for NAME= rule */
         NAMING_XEN_VIF                   = 1 << 13, /* Generate names for Xen netfront devices */
         NAMING_BRIDGE_MULTIFUNCTION_SLOT = 1 << 14, /* Use PCI hotplug slot information associated with bridge, but only if PCI device is multifunction */
+        NAMING_DEVICETREE_ALIASES        = 1 << 15, /* Generate names from devicetree aliases */
 
         /* And now the masks that combine the features above */
         NAMING_V238 = 0,
@@ -49,6 +50,7 @@ typedef enum NamingSchemeFlags {
         NAMING_V249 = NAMING_V247 | NAMING_SLOT_FUNCTION_ID | NAMING_16BIT_INDEX | NAMING_REPLACE_STRICTLY,
         NAMING_V250 = NAMING_V249 | NAMING_XEN_VIF,
         NAMING_V251 = NAMING_V250 | NAMING_BRIDGE_MULTIFUNCTION_SLOT,
+        NAMING_V252 = NAMING_V251 | NAMING_DEVICETREE_ALIASES,
 
         EXTRA_NET_NAMING_SCHEMES
 


src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
 }
 
+#define DEVICETREE_ALIASES "/proc/device-tree/aliases"
+
+static int dev_devicetree_onboard(sd_device *dev, NetNames *names) {
+        char *ofnode;
+        const char *parent_syspath;
+        _cleanup_free_ char *devicetree_syspath = NULL;
+        _cleanup_free_ char *ofnode_syspath = NULL;
+        _cleanup_free_ char *ofnode_path = NULL;
```
yuwata: Please merge the three lines.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+        _cleanup_closedir_ DIR *aliases = NULL;
+        sd_device *parent = NULL;
```
yuwata: Unnecessary initialization for `parent`.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                return r;
+
+        r = sd_device_get_syspath(parent, &parent_syspath);
+        if (r < 0)
+                return r;
+
+        ofnode = strjoina(parent_syspath, "/of_node");
+        r = chase_symlinks(ofnode, NULL, 0, &ofnode_syspath, NULL);
+        if (r < 0)
+                return r;
```
yuwata: Maybe, you can use `sd_device_new_child()` and `sd_device_get_syspath()` for the child device?

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+
+        /*
+         * Example paths:
+         * ofnode_syspath = /sys/firmware/devicetree/base/soc/ethernet@deadbeef
+         * devicetree_syspath = /sys/firmware/devicetree/base
+         * ofnode_path = soc/ethernet@deadbeef
+         */
+        r = path_make_relative(devicetree_syspath, ofnode_syspath, &ofnode_path);
```
yuwata: Maybe, `path_startswith()` is enough?

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+        FOREACH_DIRENT_ALL(alias, aliases, break) {
+                const char *conflict = DEVICETREE_ALIASES "/ethernet";
+                _cleanup_free_ char *alias_path = NULL;
+                char alias_syspath[PATH_MAX];
+                size_t alias_pathlen;
+
+                if (!startswith(alias->d_name, "ethernet") ||
+                    !snprintf_ok(alias_syspath, sizeof(alias_syspath),
+                                 DEVICETREE_ALIASES "/%s", alias->d_name) ||
+                    !read_full_virtual_file(alias_syspath, &alias_path, &alias_pathlen))
```
yuwata: You can use `read_virtual_file_at()`, then it is not necessary to build `alias_syspath`.

Also, `read_full_virtual_file(...)` returns negative errno on error, hence please use `< 0`.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                if (ofnode_pathlen + 2 != alias_pathlen ||
+                    *alias_path != '/' ||
+                    !strneq(ofnode_path, alias_path + 1, ofnode_pathlen))
+                        continue;
+
+                /* If there's no index, we default to 0... */
+                if (streq(alias->d_name, "ethernet"))
+                        conflict = DEVICETREE_ALIASES "/ethernet0";
+                else if (sscanf(alias->d_name, "ethernet%u", &index) != 1)
+                        return -EINVAL;
```
yuwata: You can reuse the result of `startswith()` at the beginning of the loop:
```
const char *p;

p = startswith(alias->d_name, "ethernet");
if (!p)
        continue;

/* check the file contents */

if (isempty(p)) {
        index = 0;
        if (access(DEVICETREE_ALIASES "/ethernet0", F_OK) >= 0) {
                return log_device_debug_errno(dev, SYNTHETIC_ERRNO(EEXIST), "...");
else {
        r = safe_atou(p, &index);
        if (r < 0)
                return log_device_debug_errno(dev, r, "...");
}
```

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,101 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+        char *ofnode;
+        const char *parent_syspath;
+        _cleanup_free_ char *devicetree_syspath = NULL;
+        _cleanup_free_ char *ofnode_syspath = NULL;
+        _cleanup_free_ char *ofnode_path = NULL;
+        _cleanup_closedir_ DIR *aliases = NULL;
+        sd_device *parent = NULL;
+        int r;
+        size_t ofnode_pathlen;
+        unsigned int index = 0;
```
yuwata: Please use `unsigned`, instead of `unsigned int`.

src/udev/udev-builtin-net_id.c
```diff
@@ -1036,6 +1134,15 @@ static int builtin_net_id(sd_device *dev, sd_netlink **rtnl, int argc, char *arg
                 ieee_oui(dev, &info, test);
         }
 
+        /* get devicetree aliases; only ethernet supported for now  */
+        if (streq(prefix, "en") && dev_devicetree_onboard(dev, &names) >= 0 &&
+            names.type == NET_DEVICETREE) {
+                char str[ALTIFNAMSIZ];
+
+                if (snprintf_ok(str, sizeof str, "%s%s", prefix, names.devicetree_onboard))
+                        udev_builtin_add_property(dev, test, "ID_NET_LABEL_ONBOARD", str);
```
yuwata: `ID_NET_NAME_ONBOARD`

yuwata (changes requested):

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: Sorry, what is the child device here? This is a symlink to something in the devicetree firmware, outside of `/sys/devices`. Is it OK to do that? We really just want to get the relative path, which is what the alias will reference.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: That looks like it will work.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: Whoops, will fix.

src/udev/udev-builtin-net_id.c
yuwata: Not familiar with device tree. But if the resolved path (/sys/firmware/devicetree/base/soc/ethernet@deadbeef in the example below) is a directory, then new_child() should work fine.

yuwata:
BTW, is this for RFE #17625? If so, please add `Closes #17625.` in the commit message.
Also, please drop signed-off-by tag in the commit message. We do not use it.

yuwata:
> In terms of implementation, apparently there can be multiple device
> trees at once. I have decided to dodge this problem for now, and just
> use /proc/device-tree. If it is desired to support multiple device trees
> later, then the scheme can be modified to something like enb2d3.

What is enb2d3??

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK, I wasn't sure whether the `sd_device_*` stuff was for just `/sys/devices`, or for anything in `/sys`.

yuwata: You can easily test by calling `udevadm info /sys/firmware/devicetree/base/soc/ethernet@deadbeef`. If it works, sd-device also works for the path.

sean-anderson-seco:
Thank you for the quick review.

> BTW, is this for RFE #17625? If so, please add `Closes #17625.` in the commit message.

I did not write this specifically for that issue, but it looks like this will close that. I will add that tag.

>Also, please drop signed-off-by tag in the commit message. We do not use it.

OK (as it happened I quickly went through the history of this file and saw it used in d6eda677b32a0063a77cb639f37c9a454b180da9).

> > In terms of implementation, apparently there can be multiple device
> > trees at once. I have decided to dodge this problem for now, and just
> > use /proc/device-tree. If it is desired to support multiple device trees
> > later, then the scheme can be modified to something like enb2d3.
> 
> What is enb2d3??

According to the [openfirmware sysfs ABI documentation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/ABI/testing/sysfs-firmware-ofw), there can be multiple devicetrees. However, it does not document the naming scheme used. However, the actual attachment is done by [`__of_attach_node_sysfs`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/of/kobj.c#n115), which uses `safe_name` to create the new top-level tree. I believe this results in something like `base`, `base1`, ..., `base15`. So the idea is if we needed to support this we could use `b` for "base".

src/udev/udev-builtin-net_id.c
sean-anderson-seco: Thanks, that was very helpful.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: As it turns out, I can use sd_device_new_from_syspath on /aliases, which lets me use sd_device_get_sysattr_value to get the aliases instead of doing the reads directly.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                alias_index = startswith(alias, "ethernet");
+                if (!alias_index)
+                        continue;
+
+                r = sd_device_get_sysattr_value(aliases, alias, &alias_path);
+                if (r < 0)
+                        continue;
+
+                /* ofnode_path lacks a leading /, so we need to adjust */
+                if (*alias_path != '/' || !streq(ofnode_path, alias_path + 1))
```
poettering: path_equal()?

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                        continue;
+
+                /* If there's no index, we default to 0... */
+                if (streq(alias, "ethernet")) {
+                        conflict = "ethernet0";
+                } else {
```
poettering, line 669: no {} around single line if blocks, according to coding style

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+        /* check if our direct parent has an of_node */
+        r = sd_device_get_parent(dev, &parent);
+        if (r < 0)
+                return r;
+
+        r = sd_device_get_syspath(parent, &parent_syspath);
+        if (r < 0)
+                return r;
+
+        r = sd_device_new_child(&ofnode, parent, "/of_node");
```
poettering: ofnode is leaked. Must be defined with `_cleanup_(sd_device_unrefp)`

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+         * Example paths:
+         * devicetree_syspath = /sys/firmware/devicetree/base
+         * ofnode_syspath = /sys/firmware/devicetree/base/soc/ethernet@deadbeef
+         * ofnode_path = soc/ethernet@deadbeef
+         */
+        ofnode_path = path_startswith(ofnode_syspath, devicetree_syspath);
+        if (!ofnode_path)
+                return -ENOENT;
+
+        r = sd_device_new_from_syspath(&aliases, strjoina(devicetree_syspath, "/aliases"));
```
poettering: also leaked

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
 }
 
+#define DEVICETREE_ALIASES "/proc/device-tree/aliases"
+
+static int dev_devicetree_onboard(sd_device *dev, NetNames *names) {
+        const char *alias, *ofnode_path, *ofnode_syspath, *parent_syspath;
+        _cleanup_free_ char *devicetree_syspath = NULL;
+        sd_device *aliases, *ofnode, *parent;
+        int r;
+        unsigned index;
```
poettering: given libc has a funciton called "index" maybe pick a different name. old gcc used to warn about cases like this even

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                        conflict = "ethernet0";
+                } else {
+                        r = safe_atou(alias_index, &index);
+                        if (r < 0)
+                                return log_device_debug_errno(dev, r,
+                                        "could not get index of alias %s\n", alias);
+                }
+
+                /* ...but make sure we don't have an alias conflict */
+                if (!index && !sd_device_get_sysattr_value(aliases, conflict, &alias_path)) {
```
poettering: index might not be initialized here...

poettering: sd_device_get_sysattr_value() returns a negative error code, hence the success check is wrong here

poettering (changes requested):

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                return 0;
+
+        /* check if our direct parent has an of_node */
+        r = sd_device_get_parent(dev, &parent);
+        if (r < 0)
+                return r;
+
+        r = sd_device_get_syspath(parent, &parent_syspath);
+        if (r < 0)
+                return r;
```
yuwata: `parent_syspath` seems not used. Please drop it.

src/udev/udev-builtin-net_id.c
yuwata: Please drop `"/"`.

Also, not a big matter, but I prefer to prefix or suffix `dev` to the sd_device object: `dev_ofnode` of `ofnode_dev`. If you do not like it. Please ignore that point.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+
+        r = sd_device_get_syspath(ofnode, &ofnode_syspath);
+        if (r < 0)
+                return r;
+
+        r = chase_symlinks("/proc/device-tree", NULL, 0, &devicetree_syspath, NULL);
+        if (r < 0)
+                return r;
```
yuwata, line 631: I am not familiar with device-tree, but is this necessary?
If `devicetree_syspath` is always `/sys/firmware/devicetree/base`, then why not hardcode the path?
Or, the path may be `/sys/firmware/devicetree/baseX`??

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+        FOREACH_DEVICE_SYSATTR(aliases, alias) {
+                const char *alias_path, *alias_index, *conflict = "ethernet";
+
+                alias_index = startswith(alias, "ethernet");
+                if (!alias_index)
+                        continue;
+
+                r = sd_device_get_sysattr_value(aliases, alias, &alias_path);
+                if (r < 0)
+                        continue;
```
yuwata: If the return value is ignored, you can write this as follows:
```c
if (sd_device_get_sysattr_value(...) < 0)
        continue;
```

src/udev/udev-builtin-net_id.c
yuwata: How about the following?
```c
ofnode_path = path_startswith(ofnode_syspath, devicetree_syspath);
if (!ofnode_path || ofnode_path <= ofnode_syspath)
        return -ENOENT;

ofnode_path--;
assert(path_is_absolute(ofnode_path));

...

FOREACH_DEVICE_SYSATTR(aliases, alias) {
        ...

        if (!path_equal(alias_path, ofnode_path))
                continue;

File: src/udev/udev-builtin-net_id.c [hunk 18]
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+
+                /* If there's no index, we default to 0... */
+                if (streq(alias, "ethernet")) {
+                        conflict = "ethernet0";
+                } else {
+                        r = safe_atou(alias_index, &index);
+                        if (r < 0)
+                                return log_device_debug_errno(dev, r,
+                                        "could not get index of alias %s\n", alias);
+                }
```
yuwata: Please move the declaration of `index` in the loop. Also, you can use `alias_index` here. How about the following:
```c
const char *conflict;

...

if (isempty(alias_index)) {
        index = 0;
        conflict = "ethernet0";
} else {
        r = safe_atou(...);
        if (r < 0)
                return log_device_debug_errno(...);
        conflict = "ethernet";
}

if (index == 0 && sd_device_get_sysattr_value(aliases, conflict, NULL) >= 0)
        return log_device_debug_errno(dev, SYNTHETIC_ERRNO(EEXIST), "...");
```

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+
+                /* ...but make sure we don't have an alias conflict */
+                if (!index && !sd_device_get_sysattr_value(aliases, conflict, &alias_path)) {
+                        log_device_debug(dev, "ethernet alias conflict: %s and %s both exist\n",
+                                         alias, conflict);
+                        return -EINVAL;
```
yuwata: Please use `SYNTHETIC_ERRNO()`. See the previous comment. And I think `EEXIST` is better here.

src/udev/udev-builtin-net_id.c
yuwata: I'd like to also name the sd_device object as `dev_aliases` or `aliases_dev`. But that's super cosmetic, and feel free to ignore it.

yuwata (changes requested):
Looks much better now. One more round please.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: I could use path_equal for the second part, but alias_path is "absolute" (relative to devicetree_syspath), so they can't be compared directly. I could also add a leading / before ofnode_path to make it "absolute" in the same way.

edit: Resolved using yuwata's suggestion.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: > ofnode is leaked. Must be defined with _cleanup_(sd_device_unrefp)

What are the rules regarding this? `sd_device_get_parent` does not require this. Is it the `_new` vs `_get`?

> Please drop "/".

OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: > index might not be initialized here...

Whoops. Looks like I missed that while refactoring.

> sd_device_get_sysattr_value() returns a negative error code, hence the success check is wrong here

OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: Huh. I'll shorten it to `i`.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: > I am not familiar with device-tree, but is this necessary? If `devicetree_syspath` is always `/sys/firmware/devicetree/base`, then why not hardcode the path? Or, the path may be `/sys/firmware/devicetree/baseX`??

> Userspace must not use the /sys/firmware/devicetree/base
> path directly, but instead should follow /proc/device-tree
> symlink. It is possible that the absolute path will change
> in the future, but the symlink is the stable ABI.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/ABI/testing/sysfs-firmware-ofw

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: Nice. I think this will work.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: Sorry, I missed this last time. Will fix.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: That looks good.

src/udev/udev-builtin-net_id.c
yuwata: > > ofnode is leaked. Must be defined with _cleanup_(sd_device_unrefp)
> 
> What are the rules regarding this? `sd_device_get_parent` does not require this. Is it the `_new` vs `_get`?

Right. The parent device given by `sd_device_get_parent()` is owned by the child device, and will be freed when the child is freed. So, the caller must not free the parent device.
The child device created by `sd_device_new_child()` is owned by the caller, hence the caller must free it.

src/udev/udev-builtin-net_id.c
yuwata: Thanks.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                } else {
+                        r = safe_atou(alias_index, &i);
+                        if (r < 0)
+                                return log_device_debug_errno(dev, r,
+                                        "could not get index of alias %s\n", alias);
+                        conflict = "ethernet";
+                }
+
+                /* ...but make sure we don't have an alias conflict */
+                if (!i && sd_device_get_sysattr_value(aliases_dev, conflict, NULL) >= 0)
```
yuwata: Our coding style does not allow implicit cast from integer to boolean. Hence, please write as `i == 0`.

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+                        return log_device_debug_errno(dev, SYNTHETIC_ERRNO(EEXIST),
+                                "ethernet alias conflict: %s and %s both exist\n",
+                                alias, conflict);
```
yuwata: Please drop `\n` at the end, and capitalize the message.

Also, please use the following indentation:
- 16ch indentation
```
                        return log_device_debug_errno(
                                        dev, SYNTHETIC_ERRNO(EEXIST),
                                        "Ethernet alias conflict: both %s and %s exist.",
                                        alias, conflict);
```
- or align with '('
```
                        return log_device_debug_errno(dev, SYNTHETIC_ERRNO(EEXIST),
                                                      "Ethernet alias conflict: both %s and %s exist.",
                                                      alias, conflict);
```

src/udev/udev-builtin-net_id.c
```diff
@@ -600,6 +603,93 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
...
+
+                /* If there's no index, we default to 0... */
+                if (isempty(alias_index)) {
+                        i = 0;
+                        conflict = "ethernet0";
+                } else {
+                        r = safe_atou(alias_index, &i);
+                        if (r < 0)
+                                return log_device_debug_errno(dev, r,
+                                        "could not get index of alias %s\n", alias);
```
yuwata: Please drop `\n` at the end, capitalize the message, and append `: %m`,
```
return log_device_debug_errno(dev, r, "Could not get index of alias %s: %m", alias);
```

yuwata (changes requested):
Mostly LGTM. Only several coding style comments.

src/udev/udev-builtin-net_id.c
sean-anderson-seco: FWIW there is no conversion, since the result of ! is an int :)

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

src/udev/udev-builtin-net_id.c
sean-anderson-seco: OK

yuwata (approved):
LGTM. Thank you.

Issue #23557, opened 2022-05-30
systemd-oomd: Adding oom-kill status for scope units


nishalkulkarni:
**Is your feature request related to a problem? Please describe.**
Currently, there's no way to detect when an application gets killed by systemd-oomd as they are created as scope units and while previous work(#22629) added this result for .service units it would be helpful to have them for scope units too.

**Describe the solution you'd like**
For now, we can add oom-kill status only when systemd-oomd kills the whole unit, as having this behaviour for individual process OOM kills by the kernel is not feasible(and also out of scope for scope units according to current documentation)

From systemd.scope documentation
```
This also means that the exit statuses of these processes are not relevant for the scope unit failure state. Scope units may still enter a failure state, for example due to resource exhaustion or stop timeouts being reached...
```

**Describe alternatives you've considered**
None

**The systemd version you checked that didn't have the feature you are asking for**
v251

There's been some past discussion on this:
https://bugzilla.redhat.com/show_bug.cgi?id=2077496

There's also a notification mechanism in the works on GNOME's side (it works for service units but has no info on scope units)
https://gitlab.gnome.org/GNOME/gnome-settings-daemon/-/merge_requests/295

If someone's already working on this, this can be a tracker and if it's not being worked on, let me know I'd like to pick it up!

CC: @benzea @anitazha

anitazha:
Hey @nishalkulkarni! No one is working on it right now so happy to have a patch. There's ScopeResult in src/core/scope.h already that you can extend with a result for OOM kill. Otherwise it's the same flow as in service units that you added previously for checking and handling OOM kills.

PR #24269, opened 2022-08-09
core/scope: Adding oom-kill status for scope units

gtwang01:
From this [issue](https://github.com/systemd/systemd/issues/23557)

Changes:
Added new ScopeResult for oom-kills.
Added function for oom-kills based on respective function from service units.
Updated scope vtable with new function.

### DIFF
--- a/src/core/scope.c
+++ b/src/core/scope.c
@@ -627,6 +627,33 @@ static void scope_notify_cgroup_empty_event(Unit *u) {
                 unit_prune_cgroup(u);
 }
 
+static void scope_notify_cgroup_oom_event(Unit *u, bool managed_oom) {
+        Scope *s = SCOPE(u);
+
+        if (managed_oom)
+                log_unit_debug(u, "Process(es) of control group were killed by systemd-oomd.");
+        else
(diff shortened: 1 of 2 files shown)


gtwang01:
Please make sure the switch correctly sorts each state correctly. I wasn't entirely sure about that part.

anitazha (commented):
initial comments:

- please squash the 2 commits. i don't see any particular reason why they should be split up this way. and your first commit implies they should be all one commit. we also prefer to have one thesis per commit. force push the changes in order to update the PR.
- please update the commit title from `https://github.com/systemd/systemd/issues/23557` to something more descriptive (such as the title you are using in the PR).
- you can add `Closes #23557` so that GitHub will auto close the issue when the PR is merged

poettering:
lgtm.

yuwata:
Please relabel the commit message from "systemd-oomd" to "core/scope", as you can see by which file is changed by the commit.

poettering:
@anitazha does this look good to you as well?

anitazha (approved):
lgtm

Issue #23785, opened 2022-06-20
tests: varlink interfaces should probably be fuzzed


evverx:
By analogy with `dfuzzer` fuzzing DBus interfaces I think there should be `vfuzzer` fuzzing the varlink stuff. It should help to catch issues like https://github.com/systemd/systemd/issues/22480 and somewhat cover PRs like https://github.com/systemd/systemd/pull/22845 (where malformed varlink messages kick out all listeners) and https://github.com/systemd/systemd/pull/23750 (which introduces a memory leak at least)
```
[164210.524275] systemd-resolved[67]: =================================================================
[164210.525019] systemd-resolved[67]: ==67==ERROR: LeakSanitizer: detected memory leaks
[164210.525664] systemd-resolved[67]: Direct leak of 5 byte(s) in 5 object(s) allocated from:
[164210.526081] systemd-resolved[67]:     #0 0x7fcc079668f7 in strdup (/usr/lib64/libasan.so.6.0.0+0x598f7)
[164210.526817] systemd-resolved[67]:     #1 0x7fcc0686338a in free_and_strdup ../src/basic/string-util.c:940
[164210.527238] systemd-resolved[67]:     #2 0x7fcc065b80d0 in json_dispatch_string ../src/shared/json.c:4417
[164210.527725] systemd-resolved[67]:     #3 0x7fcc065b6a86 in json_dispatch ../src/shared/json.c:4260
[164210.528253] systemd-resolved[67]:     #4 0x5510ae in vl_method_resolve_service ../src/resolve/resolved-varlink.c:1021
[164210.528676] systemd-resolved[67]:     #5 0x7fcc066f8178 in varlink_dispatch_method ../src/shared/varlink.c:878
[164210.529082] systemd-resolved[67]:     #6 0x7fcc066f8b50 in varlink_process ../src/shared/varlink.c:953
[164210.529629] systemd-resolved[67]:     #7 0x7fcc067062dd in defer_callback ../src/shared/varlink.c:1889
[164210.530055] systemd-resolved[67]:     #8 0x7fcc06b809a0 in source_dispatch ../src/libsystemd/sd-event/sd-event.c:3623
[164210.530469] systemd-resolved[67]:     #9 0x7fcc06b888b3 in sd_event_dispatch ../src/libsystemd/sd-event/sd-event.c:4175
[...]

PR #24271, opened 2022-08-10
oom: drop invalid %m in the log message

yuwata:
Fixes https://github.com/systemd/systemd/issues/23785#issuecomment-1210030100.

### DIFF
--- a/src/oom/oomd-manager.c
+++ b/src/oom/oomd-manager.c
@@ -83,7 +83,7 @@ static int process_managed_oom_message(Manager *m, uid_t uid, JsonVariant *param
 
                         r = cg_path_get_owner_uid(message.path, &cg_uid);
                         if (r < 0) {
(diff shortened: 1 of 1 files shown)


yuwata:
cc @evverx and @anitazha

src/oom/oomd-manager.c
```diff
@@ -83,7 +83,7 @@ static int process_managed_oom_message(Manager *m, uid_t uid, JsonVariant *param
 
                         r = cg_path_get_owner_uid(message.path, &cg_uid);
                         if (r < 0) {
-                                log_debug("Failed to get cgroup %s owner uid: %m", message.path);
+                                log_debug("Failed to get cgroup %s owner uid.", message.path);
```
poettering: maybe instead use `log_debug_errno(r, …`?

yuwata: Yeah. I may need more caffeine...

anitazha (approved):
thanks!

PR #24272, opened 2022-08-10
Use ASSERT_PTR more

dtardon:


poettering:
ping?

dtardon:
Sorry, I've got distracted by other things. I've rebased this to current master now. I've also dropped the commit dding lots of additional asserts, because it's not ready.

src/core/job.c
```diff
@@ -1355,10 +1355,12 @@ void job_shutdown_magic(Job *j) {
...
 int job_get_timeout(Job *j, usec_t *timeout) {
         usec_t x = USEC_INFINITY, y = USEC_INFINITY;
-        Unit *u = j->unit;
+        Unit *u;
         int r;
 
-        assert(u);
+        assert(j);
+
+        u = ASSERT_PTR(j->unit);
```
poettering, line 1381:
```c
u = ASSERT_PTR(ASSERT_PTR(j)->unit);`
```

?

poettering:
lgtm. (i mean, i didn't check every single change, since there are so many. boot the ones I checked all looked ok.)

the second patch could be rewritten with a nested ASSERT_PTR() i guess. But my educated guess is that this doesn't work yet, because ASSERT_PTR() didn't recive the UNIQ treatment yet, and thus we can't actually nest ASSERT_PTR()...

Anyway, addressing that should not stop us from merging this as it is now.

src/core/job.c
dtardon: I considered this but I decided I don't like it. It just seems too opaque to me... (Whether it even works is another question.)

bluca:
```
../src/core/dbus-execute.c:460:22: error: unused variable 'c' [-Werror,-Wunused-variable]
        ExecContext *c = ASSERT_PTR(userdata);
                     ^
1 error generated.
```

dtardon:
> ../src/core/dbus-execute.c:460:22: error: unused variable 'c' [-Werror,-Wunused-variable]
>         ExecContext *c = ASSERT_PTR(userdata);

Addressed.

yuwata:
Not checked it is caused by this PR, but not a few tests are failed on focal-amd64...

dtardon:
> Not checked it is caused by this PR, but not a few tests are failed on focal-amd64...

I'd be surprised if it were. The PR essentially just reshuffles asserts a bit. And if I'd made I mistake, I'd expect a test to fail everywhere.

yuwata:
Please rebase and trigger CIs again.

Issue #24232, opened 2022-08-08
hwdb modalias matching issue on aarch64


leezu:
### systemd version the issue has been seen with

251

### Used distribution

Debian Testing

### Linux kernel version used

5.19.0

### CPU architectures issue was seen on

aarch64

### Component

hardware database files

### Expected behaviour you didn't see

In /etc/udev/hwdb.d/61-sensor-local.hwdb, I would expect the following rule to match on my system

```
sensor:modalias:platform:cros-ec-accel
 ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
```

### Unexpected behaviour you saw

Above rule did not take effect. Only the following catch-all rule would work:

```
sensor:modalias:platform:*
 ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
```

### Steps to reproduce the problem

Get a Chromebook with aarch64 chip. Install Linux in developer mode. Observe `sudo su -s /bin/sh geoclue -c monitor-sensor` shows the sensor reads "upside down". Try fixing with first rule that matches the sensor's modalias.

### Additional program output to the terminal or log subsystem illustrating the issue

Note that CONFIG_DMIID documented in 60-sensor.hwdb is not applicable to the aarch64 platform.

https://github.com/systemd/systemd/blob/2fb11652381c199ad19bb469e530543366d99dd4/hwdb.d/60-sensor.hwdb#L13-L18

I suspect there is some x86 related behavior hardcoded that breaks the modalias based matching on aarch64 systems.

Once this is fixed, all the rules for individual chromebooks at

https://github.com/systemd/systemd/blob/2fb11652381c199ad19bb469e530543366d99dd4/hwdb.d/60-sensor.hwdb#L406-L430

can be replaced by a single rule matching all `sensor:modalias:platform:cros-ec-accel` as the `ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1` on all Chromebooks.

poettering:
As we do not possess a device like this this is not really actionable for us. Happy to take a patch that addresses this issue though.

leezu:
Do you know which part of the code handles parsing and matching of the strings like `sensor:modalias:platform:cros-ec-accel` and `sensor:modalias:*`? The device I used to find this issue is CP513-1H-S60F (US) or CP513-1H-S6RG (DE)

alpernebbi:
[...]

PR #24276, opened 2022-08-10
hwdb: Apply existing accel orientation quirk to all Chromebooks

alpernebbi:
The cros-ec-accel and cros-ec-accel-legacy kernel modules internally correct for the board-specific accelerometer mounting orientations. Their sensor outputs are in a standard reference frame consistent across different boards, so the orientation matrix already added for a number of devices should apply to every device using cros-ec accelerometers. The different matrix for the 'Nocturne' board seems to be an error.

This replaces the existing hwdb rules for select Chromebooks with generic rules that apply to all Chromebooks.

Also see #24232. I only tested the first fragment in `/etc/udev/hwdb.d`, and only on one the board that I have.

### DIFF
--- a/hwdb.d/60-sensor.hwdb
+++ b/hwdb.d/60-sensor.hwdb
@@ -406,27 +406,13 @@ sensor:modalias:acpi:KXCJ9000*:dmi:*:bvrZY-8-BI-PX4S70VTR400-X423B-005-D:*:rvnAM
 #########################################
 # Google Chromebooks
 #########################################
-sensor:modalias:platform:cros-ec-accel:dmi:*:svnGOOGLE:*
+
+# CrOS EC & kernel drivers internally correct for per-board sensor orientations,
+# but they return values in the inverse direction (Android & W3C specs vs HID).
+sensor:modalias:platform:cros-ec-accel:*
  ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
 
-# caroline board (Samsung Chromebook Pro) reports itself as svnGoogle
-sensor:modalias:platform:cros-ec-accel:dmi:*:svnGoogle:pnCaroline*:*
- ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
-
-# Dell Inspiron Chromebook 14 2-in-1
-sensor:modalias:platform:cros-ec-accel:dmi:*svnGoogle:pnVayne*:*
- ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
-
-# nocturne board (Google Pixel Slate)
-sensor:modalias:platform:cros-ec-accel:dmi:*Google_Nocturne*:*
- ACCEL_MOUNT_MATRIX=1, 0, 0; 0, -1, 0; 0, 0, 1
-
-# rammus board (Asus Chromebook Flip C433)
-sensor:modalias:platform:cros-ec-accel:dmi:*svnGoogle:pnRammus*:*
- ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
-
-# Lenovo ThinkPad C13 Yoga
-sensor:modalias:platform:cros-ec-accel:dmi:*svnGoogle:pnMorphius*:*
+sensor:modalias:platform:cros-ec-accel-legacy:*
  ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
 
 #########################################


leezu:
Thank you, @alpernebbi! This PR will ensure Gnome Desktop Environment and other iio-sensor-proxy dependent software will work correctly out-of-the-box on Chromebooks.

amstan:
> and only on one the board that I have

I can confirm this is the correct transform matrix for all chromebooks and that this matrix is also consistent across chromebooks. See [W3C](https://www.w3.org/TR/orientation-sensor/#absoluteorientationsensor-model) and [Android docs](https://developer.android.com/reference/android/hardware/SensorEvent.html) for more details on this.

Note: the one nautilus board before this patch was probably wrong, it should have been the same as the rest.
Note 2: Seems like gnome and iio-sensor-proxy all expect the accelerometers to come in the HID format, which is exactly inversed from what W3C and android provide. Something something disagreement on convention about newton's third law and what the sensor actually measures. This is why this chromebook transform matrix is just full of `-1`s.

`Reviewed-by: Alexandru Stan <amstan@chromium.org>`

@gwendalcr might be into this patch, he's the person that taught me all about sensors for chromebooks.

poettering:
lgtm, but could you add a comment to the entry explaining why such a broad match makes sense here (i.e. say what you wrote above there in terse words)

alpernebbi:
Rebased and added this comment, hope it's good enough:

``` diff
+# CrOS EC & kernel drivers internally correct for per-board sensor orientations,
+# but they return values in the inverse direction (Android & W3C specs vs HID).
```

gwendalg:
The patch is correct for the lid accel (label:accel-display) (the one that matters for rotation).
For the base (label:accel-base), since gnome expects the 2 sensors to report the same data when the device is closed (lid angle is 0), but natural orientation/chrome expects the display and base sensors to report the same data when device is open (lid angle 180), we need an extra rotation along the X axis:
```
[[-1, 0, 0],    [[ 1, 0, 0],    [[-1, 0, 0],
 [0, -1, 0],  X  [0, -1, 0],  =  [ 0, 1, 0],
 [0, 0, -1]]     [0, 0, -1]]     [ 0, 0, 1]]
```

For the base accelerometer, the mount matrix is `ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, 1, 0; 0, 0, 1` - a quite common matrix for other base sensors.

I don't know how to reference the iio `label` sysfs entry in the sensor modalias to distinguish the base from the display sensor.

alpernebbi:
> chrome expects the display and base sensors to report the same data when device is open (lid angle 180), we need an extra rotation along the X axis:

I see this is the case with `cat /sys/bus/iio/devices/iio:device{1,3}/{location,in_accel_{x,y,z}_raw}` on rk3399-gru-kevin.

```
                | iio:device1 | iio:device3
----------------+-------------+------------
location        | base        | lid
in_accel_x_raw  | -677        | 128
in_accel_y_raw  | -429        | 240
in_accel_z_raw  | 16296       | 16672
```

> I don't know how to reference the iio `label` sysfs entry in the sensor modalias to distinguish the base from the display sensor.

I might be wrong, but it looks like what we can match in `hwdb.d` originates from `rules.d`, more specifically:

https://github.com/systemd/systemd/blob/b1358f03627210d4aab08afd8d9d3208d0404405/rules.d/60-sensor.rules#L5-L8

And maybe we can add the `location` sysfs file to that hwdb query, since that's what `$attr` does according to `udev(7)`

```
$attr{file}, %s{file}
           The value of a sysfs attribute found at the device where all keys
           of the rule have matched. If the matching device does not have such
           an attribute, and a previous KERNELS, SUBSYSTEMS, DRIVERS, or ATTRS
           test selected a parent device, then the attribute from that parent
           device is used.
```

But I don't know what happens when it finds no value for other devices. I'm assuming adding an extra `IMPORT` line would be easiest, have to test.

alpernebbi:
I got something that works, but not sure what's the best rule here. 

```
# /etc/udev/rules.d/61-cros-ec-accel.rules
ACTION=="remove", GOTO="sensor_end"

SUBSYSTEM=="iio", KERNEL=="iio*", SUBSYSTEMS=="usb|i2c|platform", \
  IMPORT{builtin}="hwdb 'sensor:modalias:$attr{modalias}:location:$attr{location}:$attr{[dmi/id]modalias}'", \
  GOTO="sensor_end"

LABEL="sensor_end"
```

```
# /etc/udev/hwdb.d/61-cros-ec-accel.hwdb
sensor:modalias:platform:cros-ec-accel:location:lid:*
 ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, -1, 0; 0, 0, -1
 ACCEL_LOCATION=display

sensor:modalias:platform:cros-ec-accel:location:base:*
 ACCEL_MOUNT_MATRIX=-1, 0, 0; 0, 1, 0; 0, 0, 1
 ACCEL_LOCATION=base
```

This one also works, without the hwdb entries:

```
# /etc/udev/rules.d/61-cros-ec-accel.rules
ACTION=="remove", GOTO="sensor_end"

SUBSYSTEM=="iio", KERNEL=="iio*", SUBSYSTEMS=="platform", \
  ATTRS{modalias}=="platform:cros-ec-accel", ATTR{location}=="base", \
  ENV{ACCEL_MOUNT_MATRIX}="-1, 0, 0; 0, 1, 0; 0, 0, 1", \
  ENV{ACCEL_LOCATION}="base", \
  GOTO="sensor_end"

SUBSYSTEM=="iio", KERNEL=="iio*", SUBSYSTEMS=="platform", \
  ATTRS{modalias}=="platform:cros-ec-accel", ATTR{location}=="lid", \
  ENV{ACCEL_MOUNT_MATRIX}="-1, 0, 0; 0, -1, 0; 0, 0, -1", \
  ENV{ACCEL_LOCATION}="display", \
  GOTO="sensor_end"

LABEL="sensor_end"
```

yuwata:
I do not have affected laptops, so I cannot follow the discussions. But feel free to open any follow-up PRs. Thank you.

alpernebbi:
Filed #24353 as a follow-up.

PR #24277, opened 2022-08-10
Migrate code to bus-locator.h in hostnamectl.c

ruopengw:
sd_bus_call_method is changed to bus_call_method

### DIFF
--- a/src/hostname/hostnamectl.c
+++ b/src/hostname/hostnamectl.c
@@ -13,6 +13,7 @@
 #include "architecture.h"
 #include "bus-common-errors.h"
 #include "bus-error.h"
+#include "bus-locator.h"
 #include "bus-map-properties.h"
 #include "format-table.h"
 #include "hostname-setup.h"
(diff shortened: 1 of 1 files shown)


anitazha:
We don't use the internal commit message template in systemd. Please modify the commit message to just a description of the change.

mrc0mmand:
CIs are definitely not happy:

```
[    9.230064] testsuite-71.sh[267]: + assert_rc 0 hostnamectl set-hostname testhost
[    9.105928] systemd[1]: Got message type=method_call sender=:1.5 destination=org.freedesktop.systemd1 path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=SetPrettyHostname cookie=2 reply_cookie=0 signature=sb error-name=n/a error-message=n/a
[    9.219888] systemd[1]: Sent message type=error sender=n/a destination=:1.5 path=n/a interface=n/a member=n/a cookie=56 reply_cookie=2 signature=s error-name=org.freedesktop.DBus.Error.UnknownMethod error-message=Unknown method SetPrettyHostname or interface org.freedesktop.systemd1.Manager.
[    9.226780] hostnamectl[283]: Could not set pretty hostname: Unknown method SetPrettyHostname or interface org.freedesktop.systemd1.Manager.
```

src/hostname/hostnamectl.c
```diff
@@ -370,11 +371,9 @@ static int show_status(int argc, char **argv, void *userdata) {
                 _cleanup_(json_variant_unrefp) JsonVariant *v = NULL;
                 const char *text = NULL;
 
-                r = sd_bus_call_method(
+                r = bus_call_method(
                                 bus,
-                                "org.freedesktop.hostname1",
-                                "/org/freedesktop/hostname1",
-                                "org.freedesktop.hostname1",
+                                bus_systemd_mgr,
```
daandemeyer, line 376: This should use the mgr object for /org/freedesktop/hostname1. Since that one doesn't exist yet, it needs to be added to bus-locator.h and bus-locator.c

PR #24280, opened 2022-08-11
A couple of Semaphore CI fixes for recent fails

mrc0mmand:


bluca (approved):
How did they manage to break this

PR #24285, opened 2022-08-11
sd-device: make sd_device_new_from_path() support symlink outside of /sys

yuwata:
Inspired by https://github.com/systemd/systemd/pull/24265#discussion_r942261094.

### DIFF
--- a/src/udev/udev-builtin-net_id.c
+++ b/src/udev/udev-builtin-net_id.c
@@ -604,9 +604,8 @@ static int names_platform(sd_device *dev, NetNames *names, bool test) {
 }
 
 static int dev_devicetree_onboard(sd_device *dev, NetNames *names) {
-        const char *alias, *ofnode_path, *ofnode_syspath;
-        _cleanup_free_ char *devicetree_syspath = NULL;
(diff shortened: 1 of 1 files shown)


src/libsystemd/sd-device/sd-device.c
```diff
@@ -510,13 +510,24 @@ _public_ int sd_device_new_from_devname(sd_device **ret, const char *devname) {
...
         assert_return(path, -EINVAL);
 
         if (path_startswith(path, "/dev"))
                 return sd_device_new_from_devname(ret, path);
 
-        return sd_device_new_from_syspath(ret, path);
+        if (path_startswith(path, "/sys"))
+                return sd_device_new_from_syspath(ret, path);
+
+        /* Maybe, a symlink located outside of /sys? Let's try to chase the symlink. */
```
poettering: mayb explicitly mention the /proc/device-tree usecase here, because this might raise eyebrows otherwise

poettering:
lgtm, but you marked it a draft

poettering:
hmm device_set_syspath() already can do chase_symlinks(). I wonder if it would be better to rely on that and make the checking of /sys/ prefix in device_set_syspath() optional via a flag.

yuwata:
@poettering Thank you for your review. PTAL!

PR #24294, opened 2022-08-12
repart: add support for list of definitions directories

rphibel:


src/partition/repart.c
```diff
@@ -1414,11 +1414,11 @@ static int context_read_definitions(
 
         assert(context);
 
-        dirs = directory ?
-                STRV_MAKE_CONST(directory) :
+        dirs = directories ?
+                directories :
```
poettering: use the gnu short form:

```c
dirs = directories ?: (const char* const*)CONF_PATHS_STRV("repart.d");
```

src/partition/repart.c
```diff
@@ -4281,11 +4281,15 @@ static int parse_argv(int argc, char *argv[]) {
...
 
-                case ARG_DEFINITIONS:
-                        r = parse_path_argument(optarg, false, &arg_definitions);
+                case ARG_DEFINITIONS: {
+                        _cleanup_free_ char *path = NULL;
+                        r = parse_path_argument(optarg, false, &path);
                         if (r < 0)
                                 return r;
+                        if (strv_extend(&arg_definitions, path) < 0)
+                                return log_oom();
```
poettering, line 4288: maybe also do strv_uniq() somewhere here

src/partition/repart.c
```diff
@@ -4281,11 +4281,15 @@ static int parse_argv(int argc, char *argv[]) {
...
                         break;
 
-                case ARG_DEFINITIONS:
-                        r = parse_path_argument(optarg, false, &arg_definitions);
+                case ARG_DEFINITIONS: {
+                        _cleanup_free_ char *path = NULL;
+                        r = parse_path_argument(optarg, false, &path);
                         if (r < 0)
                                 return r;
+                        if (strv_extend(&arg_definitions, path) < 0)
```
poettering: use strv_push() and TAKE_PTR afterwards to donate the allocated memory to the strv. i.e.

```c
if (strv_push(&arg_definitions, &path) < 0)
        return log_oom();

TAKE_PTR(path);
```

src/partition/repart.c
```diff
@@ -4904,7 +4908,7 @@ static int run(int argc, char *argv[]) {
         if (!context)
                 return log_oom();
 
-        r = context_read_definitions(context, arg_definitions, arg_root);
+        r = context_read_definitions(context, (const char * const*)arg_definitions, arg_root);
```
poettering: I'd probably just drop the 'const' from the context_read_definitions() parameter. constant arrays of strings are a mess, you have to cast all the time. Hence so far we said: don't bother with const if you deal with a `char**` `strv` array...

poettering:
looks good, just some minor comments

src/partition/repart.c
daandemeyer: What about 

```
if (strv_consume(&arg_definitions), TAKE_PTR(path)) < 0)
        return log_oom();
```

?

poettering: You are right, of course!

poettering:
lgtm, but you marked the PR as "draft" still?

PR #24300, opened 2022-08-12
network/bridge: fix UseBPDU= and AllowPortToBeRoot=

### LINKED ISSUE #24268: Inconsistent configuration key name

245.4-4ubuntu3.17

### Used distribution

Ubuntu 20.04

### Linux kernel version used

5.4.0-122-generic

x86_64

### Component

systemd-networkd

`AllowPortToBeRoot` configuration in `Bridge` section of `systemd.network(5)` implies as per its manual page (`systemd.network(5)`):

>        AllowPortToBeRoot=
>           Takes a boolean. Configures whether a given port is allowed to become a root port. Only used when STP is enabled on the bridge. When unset, the
>           kernel's default will be used.

It essentially controls the `root_block` option of the interface participating in the bridge, and its manual page (`bridge(8)`) reads:

>        root_block on or root_block off
>           Controls whether a given port is allowed to become root port or not. Only used when STP is enabled on the bridge. By default the flag is off.

If one sets `root_block` to `on`, the port will be prevented from becoming `root`. Yet the only way to set it to `on` using `systemd-networkd` configuration is to set `AllowPortToBeRoot` to `true`, which seems like opposite of expected behaviour, if I understand it correctly, i.e. when `AllowPortToBeRoot` is `true` it should result in `root_block` to `false` therefore enabling the port to become the `root`.

Please feel free to correct me, if I misunderstood.

Thanks!

_No response_

_No response_

_No response_

yuwata:
Fixes bugs caused by 7f9915f0de67f3a10a4b22810d119da65af8c84a.

Fixes #24268.

### DIFF
--- a/src/network/networkd-setlink.c
+++ b/src/network/networkd-setlink.c
@@ -232,7 +232,7 @@ static int link_configure_fill_message(
                         return r;
 
                 if (link->network->use_bpdu >= 0) {
-                        r = sd_netlink_message_append_u8(req, IFLA_BRPORT_GUARD, link->network->use_bpdu);
+                        r = sd_netlink_message_append_u8(req, IFLA_BRPORT_GUARD, !link->network->use_bpdu);
                         if (r < 0)
                                 return r;
                 }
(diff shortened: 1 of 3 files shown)


bluca (approved):

keszybz:
Pffff, the code is easy, but the tests conflict massively.

keszybz:
Queued up for v251-stable.

PR #24303, opened 2022-08-13
firstboot: fix can't overwrite timezone

bin456789:
Currently, it's failed to overwrite timezone symlink when using systemd-firstboot command
```
systemd-firstboot --force --timezone=Asia/Shanghai
Failed to create /etc/localtime symlink: File exists
```

Using a symlink-overwritable method like symlink_atomic can solve it, which exists in basic/fs-util

### DIFF
--- a/src/firstboot/firstboot.c
+++ b/src/firstboot/firstboot.c
@@ -479,8 +479,9 @@ static int process_timezone(void) {
                                 return log_error_errno(r, "Failed to read host timezone: %m");
 
                         (void) mkdir_parents(etc_localtime, 0755);
-                        if (symlink(p, etc_localtime) < 0)
-                                return log_error_errno(errno, "Failed to create %s symlink: %m", etc_localtime);
+                        r = symlink_atomic(p, etc_localtime);
+                        if (r < 0)
+                                return log_error_errno(r, "Failed to create %s symlink: %m", etc_localtime);
 
                         log_info("%s copied.", etc_localtime);
                         return 0;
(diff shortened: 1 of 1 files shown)


src/firstboot/firstboot.c
```diff
@@ -479,7 +479,7 @@ static int process_timezone(void) {
                                 return log_error_errno(r, "Failed to read host timezone: %m");
 
                         (void) mkdir_parents(etc_localtime, 0755);
-                        if (symlink(p, etc_localtime) < 0)
+                        if (symlink_atomic(p, etc_localtime) < 0)
                                 return log_error_errno(errno, "Failed to create %s symlink: %m", etc_localtime);
```
yuwata: Unlike `symlink()`, `symlink_atomic()` returns negative errno on error. Please pass the return value to the logger, like the following:
```
r = symlink_atomic(...);
if (r < 0)
        return log_error_errno(r, "...");
```

src/firstboot/firstboot.c
```diff
@@ -497,7 +497,7 @@ static int process_timezone(void) {
         e = strjoina("../usr/share/zoneinfo/", arg_timezone);
 
         (void) mkdir_parents(etc_localtime, 0755);
-        if (symlink(e, etc_localtime) < 0)
+        if (symlink_atomic(e, etc_localtime) < 0)
                 return log_error_errno(errno, "Failed to create %s symlink: %m", etc_localtime);
```
yuwata: Also here.

yuwata (changes requested):

yuwata (approved):
LGTM.

PR #24305, opened 2022-08-13
test-network: also set StartLimitIntervalSec=0 for systemd-networkd.socket

yuwata:
The socket unit is frequently restarted during the test.

c.f. https://github.com/systemd/systemd-centos-ci/pull/517#issuecomment-1214203242


mrc0mmand (approved):
LGTM, thanks! It doesn't fix the timeout, but the errors regarding frequent restart are gone.

PR #24307, opened 2022-08-13
Add --efi-boot-option-description argument to bootctl to control the name of the boot entry.

### LINKED ISSUE #17044: Bootctl: make EFI variable for sd-boot configurable, currently hard coded to "Linux Boot Manager"
When having multiple drives, each with a separate bootable Linux install, `bootctl` adds the undifferentiated name, _Linux Boot Manager_, to the EFI boot selection list. Having a selection list with multiple identically labeled entries makes the menu useless.  

Currently, the EFI variable for sd-boot is hard-coded to _Linux Boot Manager._ Please provide an `bootctl` option to allow the user to set the EFI variable for the name used in EFI boot selection list.

This has been discussed: https://askubuntu.com/questions/1125920/how-can-i-change-the-names-of-items-in-the-efi-uefi-boot-menu  (Ubuntu apparently ships a modified `bootctl` with _ubuntu_ hard-coded instead.)
Particularly, a work-around script: https://github.com/s-n-ushakov/rename-efi-entry

ReneHollander:
By default an entry named "Linux Boot Manager" is created (which is the
previous behavior). With the flag the name of the entry can be
controlled, which is useful when installing systemd-boot to multiple ESP
partitions and having uniquely named entries.

Fixes #17044.

Manual testing:

- Default behavior:
```
% sudo build/bootctl install
Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/boot/EFI/systemd/systemd-bootx64.efi".
Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/boot/EFI/BOOT/BOOTX64.EFI".
Random seed file /boot/loader/random-seed successfully written (32 bytes).
Created EFI boot entry "Linux Boot Manager".
% efibootmgr
BootCurrent: 0004
Timeout: 1 seconds
BootOrder: 0004,0005,0006,0001,0002,0003
Boot0001* UEFI:CD/DVD Drive     BBS(129,,0x0)
Boot0002* UEFI:Removable Device BBS(130,,0x0)
Boot0003* UEFI:Network Device   BBS(131,,0x0)
Boot0004* Linux Boot Manager    HD(1,GPT,XXXX,0x800,0xff800)/File(\EFI\systemd\systemd-bootx64.efi)
Boot0005  UEFI OS       HD(1,GPT,XXXX,0x800,0xff800)/File(\EFI\BOOT\BOOTX64.EFI)0000424f
Boot0006  Windows Boot Manager  HD(1,GPT,XXXX,0x800,0xff800)/File(\EFI\MICROSOFT\BOOT\BOOTMGFW.EFI)0000424f
```

- Custom label:
```
% sudo build/bootctl install --efi-label="Test Label"
Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/boot/EFI/systemd/systemd-bootx64.efi".
Copied "/usr/lib/systemd/boot/efi/systemd-bootx64.efi" to "/boot/EFI/BOOT/BOOTX64.EFI".
Random seed file /boot/loader/random-seed successfully written (32 bytes).
Created EFI boot entry "Test Label".
% efibootmgr                                         
BootCurrent: 0004
Timeout: 1 seconds
BootOrder: 0004,0005,0006,0001,0002,0003
Boot0001* UEFI:CD/DVD Drive     BBS(129,,0x0)
Boot0002* UEFI:Removable Device BBS(130,,0x0)
Boot0003* UEFI:Network Device   BBS(131,,0x0)
Boot0004* Test Label    HD(1,GPT,XXXX,0x800,0xff800)/File(\EFI\systemd\systemd-bootx64.efi)
Boot0005  UEFI OS       HD(1,GPT,XXXX,0x800,0xff800)/File(\EFI\BOOT\BOOTX64.EFI)0000424f
Boot0006  Windows Boot Manager  HD(1,GPT,XXXX,0x800,0xff800)/File(\EFI\MICROSOFT\BOOT\BOOTMGFW.EFI)0000424f
```
- Empty string:
```
% sudo build/bootctl install --efi-label=""
Invalid --efi-label=: 
```
- Invalid string:
```
% sudo build/bootctl install --efi-label="te\"st"
Invalid --efi-label=: te"st
```
- Too long:
```
% sudo build/bootctl install --efi-label="`python -c "print('x' * 1025)"`"
--efi-label= too long: 1025 > 1024
```

### DIFF
--- a/man/bootctl.xml
+++ b/man/bootctl.xml
@@ -377,6 +377,19 @@
         <listitem><para>Install binaries for all supported EFI architectures (this implies <option>--no-variables</option>).</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--efi-boot-option-description=</option></term>
+        <listitem><para>Description of the entry added to the firmware's boot option list. Defaults to <literal>Linux
+        Boot Manager</literal>.</para>
+
+        <para>Using the default entry name <literal>Linux Boot Manager</literal> is generally preferable as only
+        one bootloader installed to a single ESP partition should be used to boot any number of OS installations
+        found on the various disks installed in the system. Specifically distributions should not use this flag
+        to install a branded entry in the boot option list. However in situations with multiple disks, each with
+        their own ESP partition, it can be beneficial to make it easier to identify the bootloader being used in
+        the firmware's boot option menu.</para></listitem>
+      </varlistentry>
+
       <xi:include href="standard-options.xml" xpointer="no-pager"/>
       <xi:include href="standard-options.xml" xpointer="json" />
       <xi:include href="standard-options.xml" xpointer="help"/>
--- a/src/boot/bootctl.c
+++ b/src/boot/bootctl.c
@@ -57,6 +57,12 @@
 #include "verbs.h"
 #include "virt.h"
 
+/* EFI_BOOT_OPTION_DESCRIPTION_MAX sets the maximum length for the boot option description
+ * stored in NVRAM. The UEFI spec does not specify a minimum or maximum length for this
+ * string, but we limit the length to something reasonable to prevent from the firmware
+ * having to deal with a potentially too long string. */
+#define EFI_BOOT_OPTION_DESCRIPTION_MAX ((size_t) 255)
+
 static char *arg_esp_path = NULL;
 static char *arg_xbootldr_path = NULL;
 static bool arg_print_esp_path = false;
@@ -85,6 +91,7 @@ static enum {
         ARG_INSTALL_SOURCE_HOST,
         ARG_INSTALL_SOURCE_AUTO,
 } arg_install_source = ARG_INSTALL_SOURCE_AUTO;
+static char *arg_efi_boot_option_description = NULL;
 
 STATIC_DESTRUCTOR_REGISTER(arg_esp_path, freep);
 STATIC_DESTRUCTOR_REGISTER(arg_xbootldr_path, freep);
@@ -92,10 +99,15 @@ STATIC_DESTRUCTOR_REGISTER(arg_install_layout, freep);
 STATIC_DESTRUCTOR_REGISTER(arg_entry_token, freep);
 STATIC_DESTRUCTOR_REGISTER(arg_root, freep);
 STATIC_DESTRUCTOR_REGISTER(arg_image, freep);
+STATIC_DESTRUCTOR_REGISTER(arg_efi_boot_option_description, freep);
 
 static const char *arg_dollar_boot_path(void) {
         /* $BOOT shall be the XBOOTLDR partition if it exists, and otherwise the ESP */
         return arg_xbootldr_path ?: arg_esp_path;
+}
+
+static const char *pick_efi_boot_option_description(void) {
+        return arg_efi_boot_option_description ?: "Linux Boot Manager";
 }
 
 static int acquire_esp(
@@ -321,7 +333,7 @@ static int settle_entry_token(void) {
                 break;
         }
 
-        if (isempty(arg_entry_token) || !string_is_safe(arg_entry_token))
+        if (isempty(arg_entry_token) || !(utf8_is_valid(arg_entry_token) && string_is_safe(arg_entry_token)))
                 return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "Selected entry token not valid: %s", arg_entry_token);
 
         log_debug("Using entry token: %s", arg_entry_token);
@@ -1139,13 +1151,13 @@ static int install_variables(const char *esp_path,
                                        "Failed to determine current boot order: %m");
 
         if (first || r == 0) {
-                r = efi_add_boot_option(slot, "Linux Boot Manager",
+                r = efi_add_boot_option(slot, pick_efi_boot_option_description(),
                                         part, pstart, psize,
                                         uuid, path);
                 if (r < 0)
                         return log_error_errno(r, "Failed to create EFI Boot variable entry: %m");
 
-                log_info("Created EFI boot entry \"Linux Boot Manager\".");
+                log_info("Created EFI boot entry \"%s\".", pick_efi_boot_option_description());
         }
 
         return insert_into_order(slot, first);
@@ -1465,6 +1477,8 @@ static int help(int argc, char *argv[], void *userdata) {
                "                       Generate JSON output\n"
                "     --all-architectures\n"
                "                       Install all supported EFI architectures\n"
+               "     --efi-boot-option-description=DESCRIPTION\n"
+               "                       Description of the entry in the boot option list\n"
                "\nSee the %2$s for details.\n",
                program_invocation_short_name,
                link,
@@ -1491,29 +1505,31 @@ static int parse_argv(int argc, char *argv[]) {
                 ARG_ENTRY_TOKEN,
                 ARG_JSON,
                 ARG_ARCH_ALL,
+                ARG_EFI_BOOT_OPTION_DESCRIPTION,
         };
 
         static const struct option options[] = {
-                { "help",                      no_argument,       NULL, 'h'                           },
-                { "version",                   no_argument,       NULL, ARG_VERSION                   },
-                { "esp-path",                  required_argument, NULL, ARG_ESP_PATH                  },
-                { "path",                      required_argument, NULL, ARG_ESP_PATH                  }, /* Compatibility alias */
-                { "boot-path",                 required_argument, NULL, ARG_BOOT_PATH                 },
-                { "root",                      required_argument, NULL, ARG_ROOT                      },
-                { "image",                     required_argument, NULL, ARG_IMAGE                     },
-                { "install-source",            required_argument, NULL, ARG_INSTALL_SOURCE            },
-                { "print-esp-path",            no_argument,       NULL, 'p'                           },
-                { "print-path",                no_argument,       NULL, 'p'                           }, /* Compatibility alias */
-                { "print-boot-path",           no_argument,       NULL, 'x'                           },
-                { "no-variables",              no_argument,       NULL, ARG_NO_VARIABLES              },
-                { "no-pager",                  no_argument,       NULL, ARG_NO_PAGER                  },
-                { "graceful",                  no_argument,       NULL, ARG_GRACEFUL                  },
-                { "quiet",                     no_argument,       NULL, 'q'                           },
-                { "make-entry-directory",      required_argument, NULL, ARG_MAKE_ENTRY_DIRECTORY      },
-                { "make-machine-id-directory", required_argument, NULL, ARG_MAKE_ENTRY_DIRECTORY      }, /* Compatibility alias */
-                { "entry-token",               required_argument, NULL, ARG_ENTRY_TOKEN               },
-                { "json",                      required_argument, NULL, ARG_JSON                      },
-                { "all-architectures",         no_argument,       NULL, ARG_ARCH_ALL                  },
+                { "help",                        no_argument,       NULL, 'h'                             },
+                { "version",                     no_argument,       NULL, ARG_VERSION                     },
+                { "esp-path",                    required_argument, NULL, ARG_ESP_PATH                    },
+                { "path",                        required_argument, NULL, ARG_ESP_PATH                    }, /* Compatibility alias */
+                { "boot-path",                   required_argument, NULL, ARG_BOOT_PATH                   },
+                { "root",                        required_argument, NULL, ARG_ROOT                        },
+                { "image",                       required_argument, NULL, ARG_IMAGE                       },
+                { "install-source",              required_argument, NULL, ARG_INSTALL_SOURCE              },
+                { "print-esp-path",              no_argument,       NULL, 'p'                             },
+                { "print-path",                  no_argument,       NULL, 'p'                             }, /* Compatibility alias */
+                { "print-boot-path",             no_argument,       NULL, 'x'                             },
+                { "no-variables",                no_argument,       NULL, ARG_NO_VARIABLES                },
+                { "no-pager",                    no_argument,       NULL, ARG_NO_PAGER                    },
+                { "graceful",                    no_argument,       NULL, ARG_GRACEFUL                    },
+                { "quiet",                       no_argument,       NULL, 'q'                             },
+                { "make-entry-directory",        required_argument, NULL, ARG_MAKE_ENTRY_DIRECTORY        },
+                { "make-machine-id-directory",   required_argument, NULL, ARG_MAKE_ENTRY_DIRECTORY        }, /* Compatibility alias */
+                { "entry-token",                 required_argument, NULL, ARG_ENTRY_TOKEN                 },
+                { "json",                        required_argument, NULL, ARG_JSON                        },
+                { "all-architectures",           no_argument,       NULL, ARG_ARCH_ALL                    },
+                { "efi-boot-option-description", required_argument, NULL, ARG_EFI_BOOT_OPTION_DESCRIPTION },
                 {}
         };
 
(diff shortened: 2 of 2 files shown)


man/bootctl.xml
```diff
@@ -377,6 +377,12 @@
         <listitem><para>Install binaries for all supported EFI architectures (this implies <option>--no-variables</option>).</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--efi-label</option></term>
```
yuwata: Please append `=`.

man/bootctl.xml
```diff
@@ -377,6 +377,12 @@
...
+        <listitem><para>Name of the entry added to the firmware's boot loader list. Defaults to "Linux Boot
+        Manager".</para></listitem>
```
yuwata: Please use `<literal>` tag instead of using quotation.

src/boot/bootctl.c
```diff
@@ -1139,13 +1141,13 @@ static int install_variables(const char *esp_path,
                                        "Failed to determine current boot order: %m");
 
         if (first || r == 0) {
-                r = efi_add_boot_option(slot, "Linux Boot Manager",
+                r = efi_add_boot_option(slot, arg_efi_label ? arg_efi_label : "Linux Boot Manager",
```
yuwata: `arg_efi_label ?: "Linux Boot Manager"`

src/boot/bootctl.c
```diff
@@ -1139,13 +1141,13 @@ static int install_variables(const char *esp_path,
...
                                         part, pstart, psize,
                                         uuid, path);
                 if (r < 0)
                         return log_error_errno(r, "Failed to create EFI Boot variable entry: %m");
 
-                log_info("Created EFI boot entry \"Linux Boot Manager\".");
+                log_info("Created EFI boot entry \"%s\".", arg_efi_label ? arg_efi_label : "Linux Boot Manager");
```
yuwata: Also here.

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,12 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_arch_all = true;
                         break;
 
+                case ARG_EFI_LABEL:
+                        r = free_and_strdup(&arg_efi_label, optarg);
+                        if (r < 0)
+                                return log_oom();
+                        break;
```
yuwata: I am not sure what kind of string is supported, but I think it is better to check the string at least with `string_is_safe()`. Also, it is better to check the length.

yuwata (changes requested):

src/boot/bootctl.c
ReneHollander: Looking at the UEFI specification there is no mention of an upper limit for the name of the entry (referred to as "Description" in the spec): https://uefi.org/sites/default/files/resources/UEFI_Spec_2_9_2021_03_18.pdf (page 74)
Longest string I bothered testing was 16384 times the letter x as a label:
image

Nonetheless I added the checks as suggested. See manual tests in initiall PR message.

ReneHollander:
Thanks for taking a look! I implemented the suggestions.

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,18 @@ static int parse_argv(int argc, char *argv[]) {
...
+                case ARG_EFI_LABEL:
+                        r = free_and_strdup(&arg_efi_label, optarg);
+                        if (r < 0)
+                                return log_oom();
+                        if (isempty(arg_efi_label) || !string_is_safe(arg_efi_label))
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                                       "Invalid --efi-label=: %s", arg_efi_label);
+                        if (strlen(arg_efi_label) > 1024)
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                                       "--efi-label= too long: %ld > 1024", strlen(arg_efi_label));
```
yuwata: Please first check the string by `optarg`, then copy it.

Please use `%zu` instead of `%ld` for `strlen()`.

I do not know the validity of the limit 1024. Where does it come from?

src/boot/bootctl.c
```diff
@@ -1465,6 +1467,8 @@ static int help(int argc, char *argv[], void *userdata) {
                "                       Generate JSON output\n"
                "     --all-architectures\n"
                "                       Install all supported EFI architectures\n"
+               "     --efi-label=STRING\n"
```
yuwata: Not a matter, but maybe `STRING` -> `NAME` or `LABEL` is better?

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,18 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_arch_all = true;
                         break;
 
+                case ARG_EFI_LABEL:
+                        r = free_and_strdup(&arg_efi_label, optarg);
+                        if (r < 0)
+                                return log_oom();
+                        if (isempty(arg_efi_label) || !string_is_safe(arg_efi_label))
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                                       "Invalid --efi-label=: %s", arg_efi_label);
```
yuwata: If the string is not safe, the string should not be directly passed to the logger.
Please `cescape()` the string and log it.
E.g. see `config_parse_string()` in shared/conf-parser.c.

yuwata (changes requested):

src/boot/bootctl.c
ReneHollander: > I do not know the validity of the limit 1024. Where does it come from?

It was choosen arbitrarily by me. Having another thought I am not sure the code here should be in the business of length checking given the UEFI spec doesn't specify any upper limit (see resolved comment in previous round of review). So I removed it again...
I am assuming when communicating with the firmware it will fail gracefully and return a proper error in case the string is too long.

src/boot/bootctl.c
ReneHollander: Done.

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,18 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_arch_all = true;
                         break;
 
+                case ARG_EFI_LABEL:
+                        if (isempty(optarg) || !string_is_safe(optarg)) {
+                                _cleanup_free_ char *escaped = NULL;
+                                escaped = cescape(optarg);
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                                       "Invalid --efi-label=: %s", cescape(escaped));
```
yuwata: Please do not double escape the string. Please use `strna()`: `cescape(escaped)` -> `strna(escaped)`.

ReneHollander:
Thanks. I now left the review threads unresolved :) (I haven't used the code review flow in Github much so I am a bit unfamiliar with it...)

yuwata (changes requested):
Now you dropped the length check. I think it is better to set a limit. The previous comment means that simply I have no idea about why 1024, not requesting to drop the limit.

src/boot/bootctl.c
yuwata: I see. But 1024 is already too long, I guess. And even we do not know the exact limit, we should set a limit to not break firmware.

@medhefgo and @poettering WDYT?

src/boot/bootctl.c
yuwata: Wow...

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,18 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_arch_all = true;
                         break;
 
+                case ARG_EFI_LABEL:
+                        if (isempty(optarg) || !string_is_safe(optarg))
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                                       "Invalid --efi-label=: %s", strna(optarg));
```
yuwata: Please escape...
```
                        if (isempty(optarg) || !string_is_safe(optarg)) {
                                _cleanup_free_ char *escaped = NULL;

                                escaped = cescape(optarg);
                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
                                                       "Invalid --efi-label=: %s", strna(escaped));
                        }
```

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,18 @@ static int parse_argv(int argc, char *argv[]) {
...
+                        if (strlen(optarg) > 1024)
+                                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                                       "--efi-label= too long: %zu > 1024", strlen(optarg));
+                        r = free_and_strdup(&arg_efi_label, optarg);
+                        if (r < 0)
+                                return log_oom();
```
yuwata: Not a matter, but you can use `free_and_strdup_warn()`:
```
return free_and_strdup_warn(&arg_efi_label, optarg);
```

yuwata (changes requested):

src/boot/bootctl.c
ReneHollander: Sorry, I get it now... `strna` just prints out `n/a` if the string is NULL and doesn't handle escpaing.

src/boot/bootctl.c
ReneHollander: I can't return here because if there is no error during processing all other parameters would be skipped.

This would work, but not sure if worth?
```
                        r = free_and_strdup_warn(&arg_efi_label, optarg);
                        if (r < 0)
                                return r;
```

yuwata: Indeed. Yeah, your code should work.

src/boot/bootctl.c
medhefgo: You should be *very* careful testing this. You could end up bricking your device writing too much data into nvram. You should delete that var and do a cold boot or two to make sure the firmware had a chance to reclaim that space.

src/boot/bootctl.c
medhefgo: Afaik the kernel will enforce a limit to the var first, but it makes sense to limit this to something sensible anyway.

medhefgo:
I am not particularly convinced of the merits of this. The idea of sd-boot is to have one installed that picks up on the various boot entries from different OSes installed. Not to have several of them lying around…

src/boot/bootctl.c
ReneHollander: > limit this to something sensible anyway.

Lowered the limit to 250. Is this more sensible? (I don't even have a reference for how large NVRAM is on average and on average how much of it is used).

src/boot/bootctl.c
ReneHollander: I removed it again and my PC booted fine this morning :)

ReneHollander:
> I am not particularly convinced of the merits of this. The idea of sd-boot is to have one installed that picks up on the various boot entries from different OSes installed. Not to have several of them lying around…

My use case is having a RAID setup (specifically ZFS on Root) and ensuring that each disk has a boot loader, that in case one disk dies/disappears the system will still boot fine. Having the labels of the different boot menu entries be different would be helpful when having to deal with a bad disk and getting rid of the right entry quickly.

yuwata (approved):
LGTM. But I'd leave merging this to other reviewers.

man/bootctl.xml
```diff
@@ -377,6 +377,12 @@
         <listitem><para>Install binaries for all supported EFI architectures (this implies <option>--no-variables</option>).</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--efi-label=</option></term>
```
poettering: hmm, that is a bit generic a name. maybe `--efi-boot-option-description="`?

This uses the EFI spec terminology, where these things are called "boot options" and they have a `Description` field, for the human readable string.

There's really no need to be terse here, given this should normally not be an option you have to use daily.

man/bootctl.xml
```diff
@@ -377,6 +377,12 @@
...
+        <listitem><para>Name of the entry added to the firmware's boot loader list. Defaults to <literal>Linux
```
poettering: similar, here, call this "description", not "name" and "boot option list" instead of "boot loader list"

src/boot/bootctl.c
```diff
@@ -1139,13 +1141,13 @@ static int install_variables(const char *esp_path,
                                        "Failed to determine current boot order: %m");
 
         if (first || r == 0) {
-                r = efi_add_boot_option(slot, "Linux Boot Manager",
+                r = efi_add_boot_option(slot, arg_efi_label ?: "Linux Boot Manager",
```
poettering: let's not have this ternary op at multiple places. Maybe add a short helper that hides it internally:

```c
static const char *pick_efi_boot_option_description(void) {
        return arg_efi_boot_option_description ?: "Linux Boot Manager";
}
```

And then call this function whenever needed.

src/boot/bootctl.c
```diff
@@ -1647,6 +1653,22 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_arch_all = true;
                         break;
 
+                case ARG_EFI_LABEL:
+                        if (isempty(optarg) || !string_is_safe(optarg)) {
```
poettering: maybe also utf8 validate the string, i.e. use the utf8_is_valid() helper

src/boot/bootctl.c
poettering: For arbitrarily chosen numeric limits we prefer if macros are added. i.e. add at some appropriate place:

```c
#define EFI_BOOT_OPTION_DESCRIPTION_MAX 250
```

And maybe add a comment saying where the 250 comes from. Maybe something "UEFI Spec doesn't put a limit on this, but let's use something resonably conservative, that would fit into a Linux file name".

any reason to pick 250 instead of 255 btw?

poettering:
i agree with @medhefgo that this all is a bit icky to make configurable. But I do understand and accept @ReneHollander's case of having multiple disks where it indeed makes sense to be able to distinguish them in the firmware boot menu

poettering:
can we maybe add a comment to the man page, emphasizing that distros shouldn't make use of this to disinguish themselves from each other, but only to distinguish backing disks or so? After all the boot loader should be shared between distros and if distros put "FooLinux" in the boot option entry name, then this kinda defeats the point that this is shared between Linuxes...

src/boot/bootctl.c
ReneHollander: > any reason to pick 250 instead of 255 btw?

no. i put 255 again (i think i had it earlier, but when changing the numbers around last i put was 250)

man/bootctl.xml
ReneHollander: Done. +1 on the suggestion on following the UEFI spec here. Initially got "label" from efibootmgr `--label` flag.

ReneHollander:
> can we maybe add a comment to the man page, emphasizing that distros shouldn't make use of this to disinguish themselves from each other, but only to distinguish backing disks or so? After all the boot loader should be shared between distros and if distros put "FooLinux" in the boot option entry name, then this kinda defeats the point that this is shared between Linuxes...

I added an explanation in the man page. I hope the wording is alright :)

man/bootctl.xml
```diff
@@ -377,6 +377,19 @@
...
+      <varlistentry>
+        <term><option>--efi-boot-option-description=</option></term>
+        <listitem><para>Description of the entry added to the firmware's boot option list. Defaults to <literal>Linux
+        Boot Manager</literal>.</para>
+
+        <para>Using the default entry name <literal>Linux Boot Manager</literal> is generally perferable as only
+        one bootloader installed to a single ESP partition should be used to boot any number of OS installations
+        found on the various disks installed in the system. Specifically distributions should not use this flag
+        to install a branded entry in the boot option list. However in situations where there is a redundant array
+        of disks, each with their own ESP partition, it can be beneficial to make it easier to identify the
```
poettering: I'd drop the reference to RAID. Just say "However in situations whith multiple disks, each with their own ESP partition, …"

If RAID or not, if you have multiple ESPs for some reason (think: Fedora on built-in disk, and an Ubuntu on a stick), you want this.

src/boot/bootctl.c
poettering: still missing

poettering:
lgtm otherwise. once the reference to raid is dropped sounds good to merge.

src/boot/bootctl.c
ReneHollander: Not sure at which point this code change went missing... now its there

ReneHollander:
Hope thats everything. Thanks for the reviews and sorry for the many back and forths...

man/bootctl.xml
```diff
@@ -377,6 +377,19 @@
         <listitem><para>Install binaries for all supported EFI architectures (this implies <option>--no-variables</option>).</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><option>--efi-boot-option-description=</option></term>
+        <listitem><para>Description of the entry added to the firmware's boot option list. Defaults to <literal>Linux
+        Boot Manager</literal>.</para>
+
+        <para>Using the default entry name <literal>Linux Boot Manager</literal> is generally perferable as only
```
bluca: typo: perferable -> preferable

bluca (approved):

poettering:
@ReneHollander could you fix that typo that @bluca found? is good to merge then.

bluca:
> @ReneHollander could you fix that typo that @bluca found? is good to merge then.

It's a very trivial and obvious fix so I've fixed it directly. Setting green label.

ReneHollander:
Thanks for merging!

Issue #24293, opened 2022-08-12
sysext image filename check too strict with symlink usage


pothos:
### systemd version the issue has been seen with

250

### Used distribution

Fedora 36

### Component

systemd-sysext

### Expected behaviour you didn't see

The setup follows a scheme that is meant to be used with systemd-sysupdate, where more than one version of a sysext image is stored in a folder with a version marker like "NAME-VERSION.raw" and a "NAME-current.raw" symlink is managed by systemd-sysupdate. The final "/etc/extensions/NAME.raw" file is just a symlink to the "NAME-current.raw" symlink which itself is a symylink to "NAME-VERSION.raw".
Both systemd-sysext or systemd-dissect should load the sysext image and only consider the first symlink name for the filename check. This already works for sysext directories.

### Unexpected behaviour you saw

Both systemd-sysext and systemd-dissect fail to load the sysext image because they seems to take the backing filename into account instead of the first symlink's name.

### Steps to reproduce the problem

1. Create a working myext.raw sysext image (place in /etc/extensions/ and restart `systemd-sysext.service` to verify it works)
2. Now move the myext.raw image to `/tmp/myext-1.raw`
3. Let `/etc/extensions/myext.raw` be a symlink to `/tmp/myext-1.raw`
4. Restart `systemd-sysext.service` and see that it fails with `Failed to read metadata for image myext: No medium found`

pothos:
The problem comes from retrieving the image file name by looking up the backing file of the loop dev while the other output of `systemd-dissect` uses the name of the symlink. I think it makes sense to pass the name as additional parameter to skip the lookup of the backing file.

pothos:
Filed PR https://github.com/systemd/systemd/pull/24322 to fix this.

PR #24322, opened 2022-08-15
Use original filename for extension name check

pothos:
The loading of an extension image from a symlink "NAME.raw" to
"NAME-VERSION.raw" failed because the release file name check worked
with the backing file of the loop device which already resolves the
symlink and thus the found name "NAME-VERSION" mismatched "NAME".
Pass the original filename and use it instead of the backing file
when available. This fixes the loading of "NAME.raw" extensions which
are a symlink to "NAME-VERSION.raw" as, e.g., may be the case when
systemd-sysupdate manages multiple versions.

Fixes https://github.com/systemd/systemd/issues/24293

### DIFF
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -193,6 +193,7 @@ static int make_partition_devname(
 
 int dissect_image(
                 int fd,
+                const char *original_path,
                 const VeritySettings *verity,
                 const MountOptions *mount_options,
                 uint64_t diskseq,
@@ -307,15 +308,29 @@ int dissect_image(
         r = sd_device_get_sysname(d, &sysname);
         if (r < 0)
                 return log_debug_errno(r, "Failed to get device sysname: %m");
-        if (startswith(sysname, "loop")) {
-                _cleanup_free_ char *name_stripped = NULL;
+        if (original_path) {
+                _cleanup_free_ char *extracted_filename = NULL, *name_stripped = NULL;
+                r = path_extract_filename(original_path, &extracted_filename);
+                if (r < 0)
+                        return r;
+                r = raw_strip_suffixes(extracted_filename, &name_stripped);
+                if (r < 0)
+                        return r;
+
+                free_and_replace(m->image_name, name_stripped);
+        } else if (startswith(sysname, "loop")) {
+                _cleanup_free_ char *extracted_filename = NULL, *name_stripped = NULL;
                 const char *full_path;
 
+                /* Note that the backing_file reference resolves symlinks, while for sysext images we want the original path */
                 r = sd_device_get_sysattr_value(d, "loop/backing_file", &full_path);
                 if (r < 0)
                         log_debug_errno(r, "Failed to lookup image name via loop device backing file sysattr, ignoring: %m");
                 else {
-                        r = raw_strip_suffixes(basename(full_path), &name_stripped);
+                        r = path_extract_filename(full_path, &extracted_filename);
+                        if (r < 0)
+                                return r;
+                        r = raw_strip_suffixes(extracted_filename, &name_stripped);
                         if (r < 0)
                                 return r;
                 }
@@ -2812,7 +2827,7 @@ int dissect_loop_device_and_warn(
         if (!name)
                 name = ASSERT_PTR(loop->node);
 
-        r = dissect_loop_device(loop, verity, mount_options, flags, ret);
+        r = dissect_loop_device(loop, name, verity, mount_options, flags, ret);
         switch (r) {
 
         case -EOPNOTSUPP:
@@ -3078,6 +3093,7 @@ int verity_dissect_and_mount(
 
         r = dissect_loop_device(
                         loop_device,
+                        src,
                         &verity,
                         options,
                         dissect_image_flags,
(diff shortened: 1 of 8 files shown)


bluca (approved):
This looks fine to me, but IIRC @poettering specifically wanted the real name of the image to be used for the check, so deferring to him

poettering:
So I think we should go the other way. i.e. change the rules for finding the extension file in the image. Instead of stupidly using the original filename without suffix, only start with that, and if this results in ENOENT, chop off everything after the final "-" or "_" char and try again. And then do that again. And again. Until no "-" or "_" are in the extension name anymore.

This would mean if you have an image file `foo-bar-baz_47.raw`, then we'd first search for `/usr/lib/extension-release.d/extension-release.foo-bar-baz_47` in the image. And if that doesn't exist, then we'd search for `/usr/lib/extension-release.d/extension-release.foo-bar-baz`, and then `/usr/lib/extension-release.d/extension-release.foo-bar` and then `/usr/lib/extension-release.d/extension-release.foo`, and fail if that also doesn't exist.

That is more flexible I think, it allows you to add suffixes freely, but still enforces the idea that the name matters to some point at least.

pothos:
I found it a bit surprising that instead of working with the name of the symlink we work with the name of the final file because that's not what I expect as user.
The suggested alternative is also ok for me, I can have a look next week. (The sysext name would come from the extension release file. If more than one extension with the same name is in a folder, the first wins.)

src/shared/dissect-image.c
```diff
@@ -327,13 +328,20 @@ int dissect_image(
...
+                        /* The backing_file reference resolves symlinks but we need the original name for the sysext name check */
+                        r = raw_strip_suffixes(basename(original_filename), &name_stripped);
                         if (r < 0)
                                 return r;
+                } else {
+                        r = sd_device_get_sysattr_value(d, "loop/backing_file", &full_path);
+                        if (r < 0)
+                                log_debug_errno(r, "Failed to lookup image name via loop device backing file sysattr, ignoring: %m");
+                        else {
+                                r = raw_strip_suffixes(basename(full_path), &name_stripped);
```
poettering: when you touch this, please replace use of basename() with path_extract_filename(), which we prefer these days.

poettering:
I agree with your thinking. I guess if there are symlinks involved, we probably should understand them like UNIX usually does in this case, i.e. only look at the symlink name, ignore the target name...

That said, I also agree with my earlier thinking. ;-) i.e. I think if we have an image called "foo-4711.raw" it should also work with only a extension-release.foo file...

So maybe implement both schemes? Yours *and* the one I proposed?

(But I can understand if you only do the one you proposed for now, and leave the other thing for someone else/another time...)

Anyway, needs a rebase now, too

src/shared/dissect-image.c
```diff
@@ -204,6 +204,7 @@ static int make_partition_devname(
 
 int dissect_image(
                 int fd,
+                const char *original_filename,
```
poettering: this can be a fully path, no? hence name `original_path`?

usually if we call something "filename" we expect it to not include slashe,s i.e. really be just a filename, not a path

src/shared/dissect-image.c
```diff
@@ -327,13 +328,20 @@ int dissect_image(
...
                                 return r;
+                } else {
+                        r = sd_device_get_sysattr_value(d, "loop/backing_file", &full_path);
+                        if (r < 0)
+                                log_debug_errno(r, "Failed to lookup image name via loop device backing file sysattr, ignoring: %m");
+                        else {
+                                r = raw_strip_suffixes(basename(full_path), &name_stripped);
+                                if (r < 0)
+                                        return r;
+                        }
```
poettering: the conditioning here is wrong... right now this is done only if `startswith(sysname, "loop")` holds. I think if people specify `original_path` that should always take precedence.

pothos:
> So maybe implement both schemes? Yours _and_ the one I proposed?

I've just started to look into the suffix topic and without this PR here it would mean that we don't enforce a naming of the entry under the extension directories. One could try to enforce it even for both but I would just care about the names in the extension directories and leave the symlink target name unrestricted.

src/shared/dissect-image.c
pothos: Thanks for the review

src/shared/dissect-image.c
```diff
@@ -307,15 +308,27 @@ int dissect_image(
...
+                _cleanup_free_ char *extracted_filename = NULL, *name_stripped = NULL;
                 const char *full_path;
 
+                /* Note that the backing_file reference resolves symlinks, while for sysext images we want the original path */
                 r = sd_device_get_sysattr_value(d, "loop/backing_file", &full_path);
                 if (r < 0)
                         log_debug_errno(r, "Failed to lookup image name via loop device backing file sysattr, ignoring: %m");
                 else {
-                        r = raw_strip_suffixes(basename(full_path), &name_stripped);
+                        r = path_extract_filename(full_path, &extracted_filename);
```
poettering, line 330: missing failure handling

pothos: Thanks

PR #24326, opened 2022-08-16
man: Correct information on sysext masking

pothos:
While I had tested that a symlink to /dev/null works to "mask" a sysext
I must have gotten something wrong and thus the instructions in
519c2f0d6b343d140f7e08e3eb0f46708c023b4a don't work. What works,
at least at the moment, is to instead have an empty directory with the
extension name under /etc/extensions/.
Correct the info in the man page and add a test for it.


bluca (approved):

keszybz:
The failure for i386 was unrelated (buildroot setup issue with package-notes).

/packit build

PR #24327, opened 2022-08-16
mkosi: Install tests in final image

daandemeyer:


meson.build
```diff
@@ -3669,6 +3669,10 @@ if conf.get('ENABLE_REPART') == 1
                 test('test-repart',
                      test_repart_sh,
                      args : [exe.full_path(), udevadm.full_path()])
+
+                if install_tests
+                        install_data(test_repart_sh.full_path(), install_dir : testsdir)
```
yuwata: please use path() instead of full_path(), it seems to be supported since meson 0.55.0. See #23391.

yuwata (changes requested):

bluca:
This could be related?

```
11:00:35 ### Testing Format=/Encrypt=/CopyFiles=
11:00:35 + cat
11:00:35 + /usr/bin/systemd-repart /tmp/test-repart.3RXngoB9XC/zzz --size=auto --dry-run=no --seed=e2a40bf9-73f1-4278-9160-49c031e7aef8 --definitions=/tmp/test-repart.3RXngoB9XC/definitions --no-pager
11:00:35 /tmp/test-repart.3RXngoB9XC/definitions/home.conf:4: Failed to expand specifiers in Label=, ignoring: home-always-too-long-xxxxxxxxxxxxxx-%v
11:00:35 Automatically determined minimal disk image size as 3.0G, current image size is 3.0G.
11:00:35 File '/tmp/test-repart.3RXngoB9XC/zzz' grown from 3.0G to 3.0G by truncation.
11:00:35 Resized partition table.
11:00:35 Applying changes.
11:00:35 Successfully wiped file system signatures from future partition 6.
11:00:35 Successfully discarded data from future partition 6.
11:00:35 Encrypting future partition 6...
11:00:35 Not compatible PBKDF options.
11:00:35 Failed to add LUKS2 key: Cannot allocate memory
11:00:35 Failed to encrypt device: Cannot allocate memory
```

DaanDeMeyer:
I think we're triggering a repart bug by installing the repart test into the image?

bluca:
@mrc0mmand have you seen this before?

DaanDeMeyer:
test-oomd-util is also coredumping on Arch. The test expects memory.stat on the systemd cgroup controller to be 0 but that isn't the case anymore it seems

mrc0mmand:
> @mrc0mmand have you seen this before?

Nope. The OOM error would make sense if repart/cryptsetup would attempt to use Argon2 for the LUKS2 partition, but I have no idea where the `Not compatible PBKDF options.` stuff comes from (if it's related at all).

mrc0mmand:
Anyway, I'll restart both jobs, maybe it's all just a bad joke from the CI overlords.

DaanDeMeyer:
Seems this is related: https://linux.debian.bugs.dist.narkive.com/RtFDjedv/bug-923513-cryptsetup-bin-can-no-longer-luksformat-as-non-root-not-compatible-pbkdf-options

DaanDeMeyer:
I can't reproduce in mkosi unfortunately (also cancelled the reruns by pushing oops)

DaanDeMeyer:
I dropped all the test-repart.sh related changes since it's being moved into the overall test infra as part of #24373 anyway

mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -77,6 +77,7 @@ Packages=
         # Provides libasan/libubsan
         gcc
         diffutils
+        cryptsetup
```
bluca: for debian and ubuntu cryptsetup-bin is better I think, as that's where the binaries are, cryptsetup pulls in a lot of scripts and initramfs hooks and so on

bluca (approved):

mkosi.default.d/debian/10-mkosi.debian
daandemeyer: Dropped the two commits adding packages as I'm revamping the packages configuration for mkosi in a separate PR

PR #24328, opened 2022-08-16
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

I saw that your scripts are in great shape and that you already have shell linter in place, but I still think that you might find `differential-shellcheck` action useful. It is able to produce reports in SARIF format, GitHub understands this format and is able to display it nicely as PR comment and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

/cc @mrc0mmand

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,30 @@
+---
+# https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme
+
+name: Differential ShellCheck
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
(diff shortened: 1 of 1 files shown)


mrc0mmand (approved):
LGTM, thanks!

I played around this in https://github.com/systemd/systemd-centos-ci/pull/529 and the results are definitely prettier and easier to *consume*. I'll let both the SuperLinter's shellcheck and this one enabled for a while (to see if there are any differences in results), but eventually will just keep this one if no issues arise.

PR #24329, opened 2022-08-16
journald: Rate limit noisy logs in journald

rphibel:
In systemd-journald, we want to avoid excessive logging whenever we fail to log something. One way to go about this is to add a log_every_x_usec() macro that we can apply to noisy log messages to prevent them from triggering too often by only logging every x microseconds.

To implement the macro, we use the existing rate limiting features in systemd in src/basic/ratelimit.h

### DIFF
--- a/src/basic/log.h
+++ b/src/basic/log.h
@@ -8,6 +8,7 @@
 #include <syslog.h>
 
 #include "macro.h"
+#include "ratelimit.h"
 
 /* Some structures we reference but don't want to pull in headers for */
 struct iovec;
@@ -367,3 +368,41 @@ int log_syntax_invalid_utf8_internal(
 #define DEBUG_LOGGING _unlikely_(log_get_max_level() >= LOG_DEBUG)
 
 void log_setup(void);
+
+typedef struct LogRateLimit {
+        int error;
+        int level;
+        RateLimit ratelimit;
+} LogRateLimit;
+
+#define log_ratelimit_internal(_level, _error, _format, _file, _line, _func, ...)        \
+({                                                                              \
+        int _log_ratelimit_error = (_error);                                    \
+        int _log_ratelimit_level = (_level);                                    \
+        static LogRateLimit _log_ratelimit = {                                  \
+                .ratelimit = {                                                  \
+                        .interval = 1 * USEC_PER_SEC,                           \
+                        .burst = 1,                                             \
+                },                                                              \
+        };                                                                      \
+        unsigned _num_dropped_errors = ratelimit_num_dropped(&_log_ratelimit.ratelimit); \
+        if (_log_ratelimit_error != _log_ratelimit.error || _log_ratelimit_level != _log_ratelimit.level) { \
+                ratelimit_reset(&_log_ratelimit.ratelimit);                     \
+                _log_ratelimit.error = _log_ratelimit_error;                    \
+                _log_ratelimit.level = _log_ratelimit_level;                    \
+        }                                                                       \
+        if (ratelimit_below(&_log_ratelimit.ratelimit))                         \
+                _log_ratelimit_error = _num_dropped_errors > 0                  \
+                ? log_internal(_log_ratelimit_level, _log_ratelimit_error, _file, _line, _func, _format " (Dropped %u similar message(s))", __VA_ARGS__, _num_dropped_errors) \
+                : log_internal(_log_ratelimit_level, _log_ratelimit_error, _file, _line, _func, _format, __VA_ARGS__); \
+        _log_ratelimit_error;                                                   \
+})
+
+#define log_ratelimit_full_errno(level, error, format, ...)             \
+        ({                                                              \
+                int _level = (level), _e = (error);                     \
+                _e = (log_get_max_level() >= LOG_PRI(_level))           \
+                        ? log_ratelimit_internal(_level, _e, format, PROJECT_FILE, __LINE__, __func__, __VA_ARGS__) \
+                        : -ERRNO_VALUE(_e);                             \
+                _e < 0 ? _e : -ESTRPIPE;                                \
+        })
--- a/src/basic/ratelimit.c
+++ b/src/basic/ratelimit.c
@@ -30,9 +30,16 @@ bool ratelimit_below(RateLimit *r) {
         if (r->num < r->burst)
                 goto good;
 
+        r->num++;
         return false;
 
 good:
         r->num++;
         return true;
+}
+
+unsigned ratelimit_num_dropped(RateLimit *r) {
+        assert(r);
+
+        return r->num > r->burst ? r->num - r->burst : 0;
 }
--- a/src/basic/ratelimit.h
+++ b/src/basic/ratelimit.h
@@ -4,7 +4,6 @@
 #include <stdbool.h>
 
 #include "time-util.h"
-#include "util.h"
 
 typedef struct RateLimit {
         usec_t interval; /* Keep those two fields first so they can be initialized easily: */
@@ -22,3 +21,5 @@ static inline bool ratelimit_configured(RateLimit *rl) {
 }
 
 bool ratelimit_below(RateLimit *r);
+
+unsigned ratelimit_num_dropped(RateLimit *r);
--- a/src/journal/journald-server.c
+++ b/src/journal/journald-server.c
@@ -873,14 +873,14 @@ static void write_to_journal(Server *s, uid_t uid, struct iovec *iovec, size_t n
         }
 
         if (vacuumed || !shall_try_append_again(f->file, r)) {
-                log_error_errno(r, "Failed to write entry (%zu items, %zu bytes), ignoring: %m", n, IOVEC_TOTAL_SIZE(iovec, n));
+                log_ratelimit_full_errno(LOG_ERR, r, "Failed to write entry (%zu items, %zu bytes), ignoring: %m", n, IOVEC_TOTAL_SIZE(iovec, n));
                 return;
         }
 
         if (r == -E2BIG)
                 log_debug("Journal file %s is full, rotating to a new file", f->file->path);
         else
-                log_info_errno(r, "Failed to write entry to %s (%zu items, %zu bytes), rotating before retrying: %m", f->file->path, n, IOVEC_TOTAL_SIZE(iovec, n));
+                log_ratelimit_full_errno(LOG_INFO, r, "Failed to write entry to %s (%zu items, %zu bytes), rotating before retrying: %m", f->file->path, n, IOVEC_TOTAL_SIZE(iovec, n));
 
         server_rotate(s);
         server_vacuum(s, false);
@@ -892,7 +892,7 @@ static void write_to_journal(Server *s, uid_t uid, struct iovec *iovec, size_t n
         log_debug("Retrying write.");
         r = journal_file_append_entry(f->file, &ts, NULL, iovec, n, &s->seqnum, NULL, NULL);
         if (r < 0)
-                log_error_errno(r, "Failed to write entry to %s (%zu items, %zu bytes) despite vacuuming, ignoring: %m", f->file->path, n, IOVEC_TOTAL_SIZE(iovec, n));
+                log_ratelimit_full_errno(LOG_ERR, r, "Failed to write entry to %s (%zu items, %zu bytes) despite vacuuming, ignoring: %m", f->file->path, n, IOVEC_TOTAL_SIZE(iovec, n));
         else
                 server_schedule_sync(s, priority);
 }
--- a/src/test/test-ratelimit.c
+++ b/src/test/test-ratelimit.c
@@ -23,4 +23,21 @@ TEST(ratelimit_below) {
                 assert_se(ratelimit_below(&ratelimit));
 }
 
+TEST(ratelimit_num_dropped) {
+        int i;
+        RateLimit ratelimit = { 1 * USEC_PER_SEC, 10 };
+
+        for (i = 0; i < 10; i++) {
+                assert_se(ratelimit_below(&ratelimit));
+                assert_se(ratelimit_num_dropped(&ratelimit) == 0);
+        }
+        assert_se(!ratelimit_below(&ratelimit));
+        assert_se(ratelimit_num_dropped(&ratelimit) == 1);
+        assert_se(!ratelimit_below(&ratelimit));
+        assert_se(ratelimit_num_dropped(&ratelimit) == 2);
+        sleep(1);
+        assert_se(ratelimit_below(&ratelimit));
+        assert_se(ratelimit_num_dropped(&ratelimit) == 0);
+}
+
 DEFINE_TEST_MAIN(LOG_INFO);


src/basic/log.h
```diff
@@ -367,3 +368,23 @@ int log_syntax_invalid_utf8_internal(
...
+
+#define log_every_x_usec(usec, level, ...)              \
+        ({                                              \
+                static RateLimit rl = {                 \
+                        .interval = (usec),             \
+                        .burst = 1                      \
+                };                                      \
+                if (ratelimit_below(&rl))               \
+                        log_full(level, __VA_ARGS__);   \
+        })
```
daandemeyer: We don't need statement expressions here, so you can drop the parentheses around the block.

daandemeyer: Also, we might shadow variables from an outer scope by using a fixed variable name. I'd prefer if we used the infra for creating unique variable names here instead, then we can get rid of the block entirely.

Search for UNIQ_T() for how to use it to create unique variable names.

DaanDeMeyer (changes requested):

DaanDeMeyer:
Hmm, I thought about this a bit more, and for journald, we should probably go for a slightly different approach, since with the current approach, users might get the wrong idea about how much entries we are failing to write.

So we could do this instead: If we're below the ratelimit, we log as usual, if we're above the ratelimit, we increase a counter that tracks the number of failed writes instead. Then, once we're below the ratelimit again, we log again and include the number of entries we failed to write and reset the counter.

We should also keep track of the exact error that's causing writes to fail and if it changes, we should log immediately about the previous error and start tracking the new error.

This wouldn't really work with the `log_every_x_usec()` macro anymore though, so we should probably drop that from the PR.

src/journal/journald-server.c
```diff
@@ -818,6 +818,78 @@ static bool shall_try_append_again(JournalFile *f, int r) {
         }
 }
 
+typedef struct FailedEntry {
+        int r;
```
daandemeyer: We use `r` as the local variable name for errors, but in structs, we should use something more descriptive, so can we call this `error`?

src/journal/journald-server.c
```diff
@@ -818,6 +818,78 @@ static bool shall_try_append_again(JournalFile *f, int r) {
...
+        size_t nb_entries;
+        size_t nb_items;
```
daandemeyer: We tend to use `num` instead of `nb` as a shorthand

src/journal/journald-server.c
```diff
@@ -818,6 +818,78 @@ static bool shall_try_append_again(JournalFile *f, int r) {
...
+        size_t total_size;
+        const char *file;
+        const char *message;
+        RateLimit rl;
```
daandemeyer: We don't usually abbreviate in structs so I'd write `ratelimit` out in full as the variable name here

src/journal/journald-server.c
```diff
@@ -818,6 +818,78 @@ static bool shall_try_append_again(JournalFile *f, int r) {
         }
 }
 
+typedef struct FailedEntry {
```
daandemeyer: I'd name this `FailedEntryStats`

src/journal/journald-server.c
```diff
@@ -818,6 +818,78 @@ static bool shall_try_append_again(JournalFile *f, int r) {
...
+                                failed_entry->r,
+                                "Failed to write entry%s (%zu items, %zu bytes)%s: %m",
+                                dest,
+                                failed_entry->nb_items,
+                                failed_entry->total_size,
+                                failed_entry->message);
+
+        failed_entry->nb_entries = 0;
+        failed_entry->nb_items = 0;
+        failed_entry->total_size = 0;
```
daandemeyer: This function is only about writing the log messages so we should probably move these assignments out of the function and into the macro

src/journal/journald-server.c
```diff
@@ -873,14 +945,14 @@ static void write_to_journal(Server *s, uid_t uid, struct iovec *iovec, size_t n
         }
 
         if (vacuumed || !shall_try_append_again(f->file, r)) {
-                log_error_errno(r, "Failed to write entry (%zu items, %zu bytes), ignoring: %m", n, IOVEC_TOTAL_SIZE(iovec, n));
+                log_failed_entry(LOG_ERR, r, NULL, ", ignoring", n, IOVEC_TOTAL_SIZE(iovec, n));
```
daandemeyer: We can pass `f->file->path` here as well I think, it was probably just an oversight that we didn't include it in the log message before

src/journal/journald-server.c
```diff
@@ -818,6 +818,78 @@ static bool shall_try_append_again(JournalFile *f, int r) {
...
+                                failed_entry->nb_items,
+                                failed_entry->total_size,
+                                failed_entry->message);
+
+        failed_entry->nb_entries = 0;
+        failed_entry->nb_items = 0;
+        failed_entry->total_size = 0;
+}
+
+#define log_failed_entry(level, _r, _file, _message, n, size)   \
```
daandemeyer: What I would do is move the `failed_entry` out of this function and store it in the `Server` object instead, like follows:

```
struct Server {
        ...

        struct {
                FailedEntryStats without_rotate;
                FailedEntryStats before_rotate;
                FailedEntryStats after_rotate;
        } failed_entry_stats;
}
```

`log_failed_entry()` can then become a normal function instead of a macro. The relevant `FailedEntry` instance can simply be passed as an argument to `log_failed_entry()`.

DaanDeMeyer (changes requested):
Overall approach looks good, but we should probably try to get rid of the macro and replace it with a function, see comments

poettering:
So, what about this idea: let's add a new macro, log_full_ratelimit() that is pretty similar to log_full(), but internally also enforces ratelimiting using an implicit static RateLimit object.

something like this:

```c

#define log_internal_ratelimit(_level, _error, _file, _line, _func, ...) \
        ({ \
                static RateLimit rl = { 5 * USEC_PER_SEC, 5 }; \
                ratelimit_below(rl) ? \
                         log_internal(_level, _error, _file, _line, _func, __VA_ARGS__) \
                         : _error; \                    
        })

#define log_ratelimit_full_errno_zerook(level, error, ...)              \
        ({                                                              \
                int _level = (level), _e = (error);                     \
                _e = (log_get_max_level() >= LOG_PRI(_level))           \
                        ? log_ratelimit_internal(_level, _e, PROJECT_FILE, __LINE__, __func__, __VA_ARGS__) \
                        : -ERRNO_VALUE(_e);                             \
                _e < 0 ? _e : -ESTRPIPE;                                \
        })
```

Or something like that. You get the idea.

Maybe make it even a bit smarter, i.e. synthesize a log message whenever `rl->num` drops from > 0 back to 0, i.e. when we stopped suppressing events and say how many messages were dropped

We could then use this at various places.

src/journal/journald-server.c
```diff
@@ -2320,6 +2379,39 @@ int server_init(Server *s, const char *namespace) {
                         .interval = DEFAULT_KMSG_OWN_INTERVAL,
                         .burst = DEFAULT_KMSG_OWN_BURST,
                 },
+
+                .failed_entry_stats = {
+                        .without_rotate = {
+                                .error = 0,
+                                .num_entries = 0,
+                                .num_items = 0,
+                                .total_size = 0,
```
poettering: no need to spell out zero initialization, that's the implied default.

src/journal/journald-server.c
```diff
@@ -818,6 +818,65 @@ static bool shall_try_append_again(JournalFile *f, int r) {
...
+                failed_entry->num_entries = 0;
+                failed_entry->num_items = 0;
+                failed_entry->total_size = 0;
+                ratelimit_reset(&failed_entry->ratelimit);
+                failed_entry->error = _error;
+        }
+
+        failed_entry->num_entries += 1;
+        failed_entry->num_items += n;
+        failed_entry->total_size += size;
```
poettering: I am not sure the finegrained counting is really something we want or need. I think a message saying "Dropped 542 similar messages"  or so after we are leaving the ratelimit window is sufficient, and not saying what was in them

rphibel:
> So, what about this idea: let's add a new macro, log_full_ratelimit() that is pretty similar to log_full(), but internally also enforces ratelimiting using an implicit static RateLimit object.
> 
> something like this:
> 
> ```c
> #define log_internal_ratelimit(_level, _error, _file, _line, _func, ...) \
>         ({ \
>                 static RateLimit rl = { 5 * USEC_PER_SEC, 5 }; \
>                 ratelimit_below(rl) ? \
>                          log_internal(_level, _error, _file, _line, _func, __VA_ARGS__) \
>                          : _error; \                    
>         })
> 
> #define log_ratelimit_full_errno_zerook(level, error, ...)              \
>         ({                                                              \
>                 int _level = (level), _e = (error);                     \
>                 _e = (log_get_max_level() >= LOG_PRI(_level))           \
>                         ? log_ratelimit_internal(_level, _e, PROJECT_FILE, __LINE__, __func__, __VA_ARGS__) \
>                         : -ERRNO_VALUE(_e);                             \
>                 _e < 0 ? _e : -ESTRPIPE;                                \
>         })
> ```
> 
> Or something like that. You get the idea.
> 
> Maybe make it even a bit smarter, i.e. synthesize a log message whenever `rl->num` drops from > 0 back to 0, i.e. when we stopped suppressing events and say how many messages were dropped
> 
> We could then use this at various places.

Thanks for the review. We thought about doing something like that initially, but we wanted the rate limit to be reset if the error changed (See [comment from Daan above](https://github.com/systemd/systemd/pull/24329#issuecomment-1217728112)). Maybe we can still do something more generic while keeping track of changes in error type.

DaanDeMeyer:
> Thanks for the review. We thought about doing something like that initially, but we wanted the rate limit to be reset if the error changed (See https://github.com/systemd/systemd/pull/24329#issuecomment-1217728112). Maybe we can still do something more generic while keeping track of changes in error type.

Hmm, the error is passed as a variable to the macro though, so we could define another static variable to save the error and compare against it. If we do that, the generic approach should actually work (Sorry for the back and forth on implementations, given this is what you had first)

src/journal/journald-server.c
rphibel: Fixed in commit https://github.com/systemd/systemd/commit/d77fdd883eeadce347ca517d21b8cf2b1a2a6f04

src/journal/journald-server.c
rphibel: Fixed in commit https://github.com/systemd/systemd/commit/d77fdd883eeadce347ca517d21b8cf2b1a2a6f04

rphibel:
> > Thanks for the review. We thought about doing something like that initially, but we wanted the rate limit to be reset if the error changed (See [#24329 (comment)](https://github.com/systemd/systemd/pull/24329#issuecomment-1217728112)). Maybe we can still do something more generic while keeping track of changes in error type.
> 
> Hmm, the error is passed as a variable to the macro though, so we could define another static variable to save the error and compare against it. If we do that, the generic approach should actually work (Sorry for the back and forth on implementations, given this is what you had first)

You're right. I realized afterwards that it was possible to implement a generic approach. It's implemented in my last commit:  https://github.com/systemd/systemd/commit/d77fdd883eeadce347ca517d21b8cf2b1a2a6f04

src/basic/log.c
```diff
@@ -1478,3 +1478,18 @@ void log_setup(void) {
         if (log_on_console() && show_color < 0)
                 log_show_color(true);
 }
+
+void log_dropped_errors(size_t num_dropped_errors, int level) {
+        if (num_dropped_errors == 0)
+                return;
+
+        log_full(level, "Dropped %zu similar message(s)", num_dropped_errors);
```
poettering: this function should probably also reset the num_dropped_errors counter back to zero, instead of leaving that to the caller

src/basic/log.h
```diff
@@ -367,3 +368,41 @@ int log_syntax_invalid_utf8_internal(
...
+#define log_ratelimit_internal(_level, _error, _file, _line, _func, ...)        \
+({                                                                              \
+        int _log_ratelimit_error = (_error);                                    \
+        static ErrorRateLimit error_ratelimit = {                               \
+                .ratelimit = {                                                  \
+                        .interval = (1 * USEC_PER_SEC),                         \
+                        .burst = 1                                              \
+                },                                                              \
+        };                                                                      \
+        static size_t num_dropped_errors;                                       \
```
poettering: hmm, so the RateLimit object already counts hits. it currently stops doing so once the limit is reached though. I think we should change that hoewver, and then continue counting. That way we don't have to count hits while limited separately (like you do here), but instead get the count of dropped messages simply by subtracting the limit from `ratelimit->num`.

src/basic/log.h
```diff
@@ -367,3 +368,41 @@ int log_syntax_invalid_utf8_internal(
...
+        if (error_is_ratelimited(_error, &error_ratelimit)) {                   \
+                num_dropped_errors += 1;                                        \
+        } else {                                                                \
```
poettering: no {} around single-line if blocks btw, se coding style

src/basic/log.h
```diff
@@ -367,3 +368,41 @@ int log_syntax_invalid_utf8_internal(
...
+} ErrorRateLimit;
+
+void log_dropped_errors(size_t num_dropped_errors, int level);
+
+bool error_is_ratelimited(int error, ErrorRateLimit *error_ratelimit);
+
+#define log_ratelimit_internal(_level, _error, _file, _line, _func, ...)        \
+({                                                                              \
+        int _log_ratelimit_error = (_error);                                    \
+        static ErrorRateLimit error_ratelimit = {                               \
```
poettering: maybe name this with an underscore

src/basic/log.c
```diff
@@ -1478,3 +1478,18 @@ void log_setup(void) {
         if (log_on_console() && show_color < 0)
                 log_show_color(true);
 }
+
+void log_dropped_errors(size_t num_dropped_errors, int level) {
```
poettering: probably should take ErrorRateLimit object, and not just the counter

src/basic/log.h
```diff
@@ -367,3 +368,41 @@ int log_syntax_invalid_utf8_internal(
 #define DEBUG_LOGGING _unlikely_(log_get_max_level() >= LOG_DEBUG)
 
 void log_setup(void);
+
+typedef struct ErrorRateLimit {
```
poettering: I think LogRateLimit would be a better name, given this is not just about error messages per se, but any kind of log message really.

src/basic/log.h
```diff
@@ -367,3 +368,41 @@ int log_syntax_invalid_utf8_internal(
...
+        int error;
+        RateLimit ratelimit;
+} ErrorRateLimit;
+
+void log_dropped_errors(size_t num_dropped_errors, int level);
+
+bool error_is_ratelimited(int error, ErrorRateLimit *error_ratelimit);
```
poettering: we try to prefix all log related stuff with `log_`, hence this should carry a similar prefix

poettering (changes requested):

src/basic/log.c
rphibel: With the new approach (using the internal counter of the `RateLimit` object), this is not needed anymore

src/basic/log.h
rphibel: Thanks for the suggestion! I implemented it. It does simplify things :)

src/basic/log.c
rphibel: When we call `log_error_is_ratelimited`, the internal counter of the `RateLimit` object is reset if the error changes or if the error is not rate-limited anymore. So, we have to save the number of dropped errors in a local variable before calling `log_error_is_ratelimited`.

We cannot use the values in the `RateLimit` object after the rate limit has been lifted.

src/basic/log.h
```diff
@@ -367,3 +368,38 @@ int log_syntax_invalid_utf8_internal(
 #define DEBUG_LOGGING _unlikely_(log_get_max_level() >= LOG_DEBUG)
 
 void log_setup(void);
+
+typedef struct LogRateLimit {
+        int error;
+        RateLimit ratelimit;
+} LogRateLimit;
```
daandemeyer: We should save the level of the log message as well, it's not always guaranteed to be the same.

src/basic/log.h
```diff
@@ -367,3 +368,38 @@ int log_syntax_invalid_utf8_internal(
...
+                        .burst = 1                                              \
+                },                                                              \
+        };                                                                      \
+        size_t num_dropped_errors = ratelimit_num_dropped(&_log_ratelimit.ratelimit); \
+        if (!log_error_is_ratelimited(_error, &_log_ratelimit)) {               \
+                log_dropped_errors(num_dropped_errors, _level);                 \
+                _log_ratelimit_error = log_internal(_level, _error, _file, _line, _func, __VA_ARGS__);  \
+        }                                                                       \
+        _log_ratelimit_error;                                                   \
+})
```
daandemeyer: This doesn't seem entirely correct yet.

- Right now, the message about similar messages being dropped is logged separately from the original log message. This makes it harder to associate it to the original log message. Instead, what if we append this information to the original log message, so that we'd get something like the following:

```
Failed to write entry: Too Big (Dropped 56 similar messages)
```

You can easily concatenate two static format strings in the macro by putting their variables next to each other without a comma.

- Also, if we get a different error, we reset the rate limit but we don't log about the new error. This means that if the error is fatal and the program exits, we never log about the error. To avoid that, I think when we encounter a new error, we should first log about the ratelimited error, including the amount of similar messages dropped, followed by logging about the new error.

src/basic/log.c
```diff
@@ -1478,3 +1478,18 @@ void log_setup(void) {
...
+        if (num_dropped_errors == 0)
+                return;
+
+        log_full(level, "Dropped %zu similar message(s)", num_dropped_errors);
+}
+
+bool log_error_is_ratelimited(int error, LogRateLimit *log_ratelimit) {
```
daandemeyer: It's very unintuitive that this function is called `log_error_is_ratelimited()`, yet it still modifies the rate limit struct. I don't think this function adds much value, I'd remove it and inline it into the macro.

src/basic/log.h
rphibel: - Do you mean that we should keep track of the exact message not just the error type? It's possible for the error message to change between different calls to the macro. In that case, what should we display? What we could do is just add information about the type of error that was dropped: `Error: Too Big (Dropped 56 similar messages)`. WDYT?
- Actually we do log the new message immediately: after `log_dropped_errors`, we call `log_internal` with the new error.

src/basic/log.h
rphibel: Should we also reset the rate limit if the level changes?

src/basic/log.h
daandemeyer: Keeping track of the entire message is going to get too complicated. The level and error are the only things we can reasonably keep track of. So I'd apply the same logic to the level as we do to the error, but disregard the rest of the format arguments as we can't reasonably track them.

And thinking about it again, you're right, we already log the new message immediately, so we don't need to log twice. I'd still merge the "dropped similar messages" stuff into the log message though instead of doing a separate log for it.

So to summarize:
- Keep track of the level, reset the rate limit if it changes
- Don't track the full message or any of the other format arguments
- Merge the two log lines into a single log line

Let me know if something is unclear

src/basic/log.h
daandemeyer: See my other comment, I think we should

src/basic/log.h
```diff
@@ -367,3 +368,40 @@ int log_syntax_invalid_utf8_internal(
...
+        int level;
+        RateLimit ratelimit;
+} LogRateLimit;
+
+#define log_ratelimit_internal(_level, _error, _format, _file, _line, _func, ...)        \
+({                                                                              \
+        int _log_ratelimit_error = (_error);                                    \
+        static LogRateLimit _log_ratelimit = {                                  \
+                .ratelimit = {                                                  \
+                        .interval = (1 * USEC_PER_SEC),                         \
```
poettering: you can drop the () btw

src/basic/ratelimit.c
```diff
@@ -30,9 +30,14 @@ bool ratelimit_below(RateLimit *r) {
...
+        r->num++;
         return false;
 
 good:
         r->num++;
         return true;
 }
+
+size_t ratelimit_num_dropped(RateLimit *r) {
+        return (r->num > r->burst) ? (r->num - r->burst) : 0;
```
poettering: maybe do an `assert(r)` first.

and maybe drop the two pairs of (), they are redundant here

bluca: assert(r)

src/basic/ratelimit.h
```diff
@@ -22,3 +21,5 @@ static inline bool ratelimit_configured(RateLimit *rl) {
 }
 
 bool ratelimit_below(RateLimit *r);
+
+size_t ratelimit_num_dropped(RateLimit *r);
```
poettering: size_t is a weird type for this, given the count and burst are both just `unsigned`. Please stick to `unsigned` hence

src/basic/log.h
```diff
@@ -367,3 +368,40 @@ int log_syntax_invalid_utf8_internal(
...
+                        .burst = 1                                              \
```
poettering: not that it matters, but we typically put a trailing comma at structured initializations, to keep patches for later additions small...

src/basic/log.h
```diff
@@ -367,3 +368,40 @@ int log_syntax_invalid_utf8_internal(
...
+
+typedef struct LogRateLimit {
+        int error;
+        int level;
+        RateLimit ratelimit;
+} LogRateLimit;
+
+#define log_ratelimit_internal(_level, _error, _format, _file, _line, _func, ...)        \
+({                                                                              \
+        int _log_ratelimit_error = (_error);                                    \
```
poettering, line 380: _level needs the same treatment, to avoid double eval

src/basic/log.h
```diff
@@ -367,3 +368,40 @@ int log_syntax_invalid_utf8_internal(
...
+        static LogRateLimit _log_ratelimit = {                                  \
+                .ratelimit = {                                                  \
+                        .interval = (1 * USEC_PER_SEC),                         \
+                        .burst = 1                                              \
+                },                                                              \
+        };                                                                      \
+        size_t num_dropped_errors = ratelimit_num_dropped(&_log_ratelimit.ratelimit); \
```
poettering: please prefix name with `_` too

src/basic/log.h
```diff
@@ -367,3 +368,40 @@ int log_syntax_invalid_utf8_internal(
...
+        if (_error != _log_ratelimit.error || _level != _log_ratelimit.level) { \
+                ratelimit_reset(&_log_ratelimit.ratelimit);                     \
+                _log_ratelimit.error = _error;                                  \
+                _log_ratelimit.level = _level;                                  \
+        }                                                                       \
+        if (ratelimit_below(&_log_ratelimit.ratelimit))                         \
+                _log_ratelimit_error = (num_dropped_errors > 0)                 \
```
poettering: the () are redundant

poettering (changes requested):

PR #24334, opened 2022-08-16
bash-completion: autocomplete cgroup names in systemd-cgtop

bluca:


shell-completion/bash/systemd-cgtop
```diff
@@ -56,6 +56,9 @@ _systemd_cgtop() {
     fi
 
     COMPREPLY=( $(compgen -W '${OPTS[*]}' -- "$cur") )
+    if [ -d /sys/fs/cgroup/ ]; then
+        COMPREPLY+=( $(cd /sys/fs/cgroup/ && compgen -o nospace -o dirnames "$cur") )
```
keszybz, line 62: This does not work with cgroup v1. Do we care?

bluca: Speaking for myself, I really don't :-) I can add a quick check to ensure it's in unified mode?

keszybz: Maybe try to cd to /sys/fs/cgroup/systemd first, and then /sys/fs/cgroup otherwise?

bluca: ok, done

DaanDeMeyer (approved):

Issue #21165, opened 2021-10-28
Improve Suspend-Then-Hibernate functionality on Linux


axet:
**Is your feature request related to a problem? Please describe.**

suspend-then-resume is very noisy. Between suspend and hibernate it wake-up machine, starts display with few flickering, starts background music (if currently playing) and hibernate with few screen flickering.

**Describe the solution you'd like**

- [ ] ~~Create a new top-level cgroup and move everything below it, so that it can be frozen to avoid wasting power/resuming tasks, without affecting the wakeup handler which would be in a separate hierarchy - eg, /sys/fs/cgroup/root gets frozen and /sys/fs/cgroup/s2h keeps running with the handler~~
- [ ] ~~Hide the new top-level /sys/fs/cgroup/root/ so that existing hard-coded assumptions are not broken by creating a cgroup namespace early at boot where to move PID1 into, mounting /sys/fs/cgroup/s2h under /run/systemd/ so that it can be accessed~~
- [x] Estimate battery consumption and set up timed wakeup to assess again
- [x] Handle when to wake up for the first check (no previous data to do an estimation)
- [x] Record estimates/consumption on machine-specific (keyed by /etc/machine-id) storage location for reuse on future reboots
- [x] Add support for ACPI _BTP so that the bios/hw will trigger a wakeup when battery is low
- [x] Freeze/thaw `user.slice` on suspend/resume
- [ ] ~~Add qemu-based automated tests exercising as much of this feature as possible. Qemu/KVM support suspend/hibernate.~~

**Describe alternatives you've considered**

I guess, before suspend, machine should be prepared and moved in special state with blackscreen and no background sounds (suspended user apps), then after resume screen will be blank and hiberate process will continue in silince. Then after resume from hibernate it should restore black screen into normal screen and background sounds.

**The systemd version you checked that didn't have the feature you are asking for**

systemd-247.3-6

khansaUmatullah:
> We have funding for an Outreachy internship to work on exactly this issue. Looking for candidates at the moment: #20928 These are the high-level tasks: https://github.com/systemd/systemd/projects/2

@bluca I wanna work on this project. I'm working on repeatable tasks

poettering:
[...]

PR #24336, opened 2022-08-17
sleep: freeze and thaw user.slice to avoid wasting power/resuming tasks, without affecting the wakeup handler

cerebro1:
fix : https://github.com/systemd/systemd/issues/21165

### DIFF
--- a/src/sleep/sleep.c
+++ b/src/sleep/sleep.c
@@ -14,10 +14,13 @@
 #include <sys/timerfd.h>
 #include <unistd.h>
 
+#include "sd-bus.h"
 #include "sd-messages.h"
 
 #include "btrfs-util.h"
 #include "bus-error.h"
+#include "bus-locator.h"
+#include "bus-util.h"
 #include "def.h"
 #include "exec-util.h"
 #include "fd-util.h"
@@ -29,6 +32,7 @@
 #include "parse-util.h"
 #include "pretty-print.h"
 #include "sleep-config.h"
+#include "special.h"
 #include "stdio-util.h"
 #include "string-util.h"
 #include "strv.h"
@@ -349,10 +353,37 @@ static int custom_timer_suspend(const SleepConfig *sleep_config) {
         return 1;
 }
 
+/* Freeze when invoked and thaw on cleanup */
+static int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus = NULL;
+        int r;
+
+        if (!method || !*method)
+                return 0;
+
+        r = bus_connect_system_systemd(&bus);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to open connection to systemd: %m");
+
+        r = bus_call_method(bus, bus_systemd_mgr, *method, &error, NULL, "s", SPECIAL_USER_SLICE);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to execute operation: %s", bus_error_message(&error, r));
+
+        return 1;
+}
+
 static int execute_s2h(const SleepConfig *sleep_config) {
+        _unused_ _cleanup_(freeze_thaw_user_slice) const char *auto_method_thaw = NULL;
         int r, k;
 
         assert(sleep_config);
+
+        r = freeze_thaw_user_slice(&(const char*) { "FreezeUnit" });
+        if (r < 0)
+                log_debug_errno(r, "Failed to freeze unit user.slice, ignoring: %m");
+        else
+                auto_method_thaw = "ThawUnit"; /* from now on we want automatic thawing */;
 
         r = check_wakeup_type();
         if (r < 0)


src/shared/sleep-config.c
```diff
@@ -47,6 +52,11 @@
 #define DISCHARGE_RATE_FILEPATH "/var/lib/systemd/sleep/battery_discharge_percentage_rate_per_hour"
 #define BATTERY_DISCHARGE_RATE_HASH_KEY SD_ID128_MAKE(5f,9a,20,18,38,76,46,07,8d,36,58,0b,bb,c4,e0,63)
 #define SYS_ENTRY_RAW_FILE_TYPE1 "/sys/firmware/dmi/entries/1-0/raw"
+#define FREEZE_THAW_UNIT "user.slice"
+
+static BusTransport arg_transport = BUS_TRANSPORT_LOCAL;
```
bluca: These 3 are not needed, it's always local and password auth is unneeded

src/sleep/sleep.c
```diff
@@ -264,10 +264,16 @@ static int execute(
 
 static int execute_s2h(const SleepConfig *sleep_config) {
         _cleanup_hashmap_free_ Hashmap *last_capacity = NULL, *current_capacity = NULL;
+        _cleanup_(freeze_thaw_user_slice) const char *method = "thaw";
         int wakeup_type_bit = 0, r;
 
         assert(sleep_config);
 
+        method = "freeze";
```
bluca: This means the cleanup will freeze again

src/shared/sleep-config.c
```diff
@@ -127,6 +137,50 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+                return bus_log_connect_error(r, arg_transport);
+
+        (void) sd_bus_set_allow_interactive_authorization(*bus, arg_ask_password);
+
+        return 0;
+}
+
+/* Freeze or thaw user.slice */
+int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(bus_wait_for_units_freep) BusWaitForUnits *w = NULL;
```
bluca: Return immediately if method is null

src/sleep/sleep.c
```diff
@@ -264,10 +264,16 @@ static int execute(
...
+        r = freeze_thaw_user_slice(&method);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to freeze unit user.slice: %m");
```
bluca: This should be best effort - log but continue, and set method to null so that cleanup is skipped

src/sleep/sleep.c
cerebro1: ok. so initially to freeze, i can then directly pass `freeze` to function `freeze_thaw_user_slice`

src/shared/sleep-config.c
cerebro1: probably adding assert would be safe then ?

src/sleep/sleep.c
cerebro1: ok

src/shared/sleep-config.c
bluca: returning instead allows to set the cleanup var to null to skip it

src/shared/sleep-config.c
cerebro1: so we do not need to `acquire_bus` for this case?

src/shared/sleep-config.c
cerebro1: true

src/shared/sleep-config.c
```diff
@@ -127,6 +133,50 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+
+/* Freeze or thaw user.slice */
+int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *m = NULL;
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        sd_bus *bus;
+        int r;
+
+        if (*method == NULL)
+                return -1;
```
bluca: this is not an error, return 0 - also check `method`, not `*method`

src/shared/sleep-config.c
```diff
@@ -127,6 +133,50 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
         return 0;
 }
 
+static int acquire_bus(sd_bus **bus) {
```
bluca: there's no need for a separate function for this

src/shared/sleep-config.c
```diff
@@ -127,6 +133,50 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+                return bus_log_connect_error(r, BUS_TRANSPORT_LOCAL);
+
+        return 0;
+}
+
+/* Freeze or thaw user.slice */
+int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *m = NULL;
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        sd_bus *bus;
```
bluca: `_cleanup_(sd_bus_flush_close_unrefp) sd_bus`

src/shared/sleep-config.c
```diff
@@ -127,6 +133,50 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+        int r;
+
+        if (*method == NULL)
+                return -1;
+
+        r = acquire_bus(&bus);
+        if (r < 0)
+                return r;
+
+        r = bus_message_new_method_call(bus, &m, bus_systemd_mgr, *method);
```
bluca: just use `bus_call_method` which wraps all 3 calls in one

src/shared/sleep-config.c
```diff
@@ -16,10 +16,15 @@
 #include <unistd.h>
 
 #include "sd-device.h"
+#include "sd-bus.h"
```
bluca: sort alphabetically

src/shared/sleep-config.c
cerebro1: if i return 0, method will not be set to NULL to skip cleanup

bluca: how so?

src/shared/sleep-config.c
```diff
@@ -127,6 +133,50 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
 
+static int acquire_bus(sd_bus **bus) {
+        int r;
+
+        assert(bus);
+
+        if (*bus)
+                return 0;
+
+        r = bus_connect_transport(BUS_TRANSPORT_LOCAL, NULL, false, bus);
```
poettering: please go directly for `bus_connect_system_systemd` instead

src/shared/sleep-config.c
cerebro1: because it checks fir r<0, and not r <= 0.

src/shared/sleep-config.c
cerebro1: ok

src/shared/sleep-config.c
bluca: It doesn't matter, because it doesn't pass null at that point

src/shared/sleep-config.c
```diff
@@ -127,6 +133,27 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
         return 0;
 }
 
+/* Freeze or thaw user.slice */
+int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(sd_bus_message_unrefp) sd_bus_message *m = NULL;
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus;
```
bluca: Needs to be initialized to null since there's a cleanup

src/shared/sleep-config.c
```diff
@@ -127,6 +133,27 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+        int r;
+
+        if (method == NULL)
+                return 0;
+
+        r = bus_connect_transport(BUS_TRANSPORT_LOCAL, NULL, false, &bus);
+        if (r < 0)
+                return bus_log_connect_error(r, BUS_TRANSPORT_LOCAL);
+
+        r = bus_call_method(bus, bus_systemd_mgr, *method, &error, NULL, FREEZE_THAW_UNIT);
```
bluca: Type specifier is missing, "s"

src/shared/sleep-config.c
```diff
@@ -127,6 +133,27 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+        if (r < 0)
+                return log_error_errno(r, "Failed to execute operation: %s", bus_error_message(&error, r));
```
bluca: This should be debug level

src/shared/sleep-config.c
cerebro1: ok

src/shared/sleep-config.c
```diff
@@ -15,11 +15,16 @@
...
+#include "sd-bus.h"
 #include "sd-device.h"
 
 #include "alloc-util.h"
 #include "blockdev-util.h"
 #include "btrfs-util.h"
+#include "bus-error.h"
+#include "bus-locator.h"
+#include "bus-util.h"
+#include "bus-wait-for-units.h"
```
bluca: I don't think this header is used at the moment?

poettering (changes requested):

src/shared/sleep-config.c
cerebro1: hmm. corrected

src/shared/sleep-config.c
cerebro1: and return error in case opening the connection fails ?

bluca: yes

src/shared/sleep-config.c
cerebro1: yeah top 3 are used. 4th one is not.

src/sleep/sleep.c
```diff
@@ -264,10 +264,18 @@ static int execute(
...
 static int execute_s2h(const SleepConfig *sleep_config) {
         _cleanup_hashmap_free_ Hashmap *last_capacity = NULL, *current_capacity = NULL;
+        _cleanup_(freeze_thaw_user_slice) const char *method = "thaw";
+        const char *method_freeze = "freeze";
         int wakeup_type_bit = 0, r;
 
         assert(sleep_config);
 
+        r = freeze_thaw_user_slice(&method_freeze);
+        if (r <= 0) {
```
bluca: log_debug_errno will assert if `r` is 0. But it's not necessary to error out if it is, as that's an expected condition, so just check `r < 0`

src/shared/sleep-config.c
```diff
@@ -41,10 +45,13 @@
 #include "string-util.h"
 #include "strv.h"
 #include "time-util.h"
+#include "unaligned.h"
 
 #define BATTERY_LOW_CAPACITY_LEVEL 5
 #define DISCHARGE_RATE_FILEPATH "/var/lib/systemd/sleep/battery_discharge_percentage_rate_per_hour"
 #define BATTERY_DISCHARGE_RATE_HASH_KEY SD_ID128_MAKE(5f,9a,20,18,38,76,46,07,8d,36,58,0b,bb,c4,e0,63)
+#define SYS_ENTRY_RAW_FILE_TYPE1 "/sys/firmware/dmi/entries/1-0/raw"
+#define FREEZE_THAW_UNIT "user.slice"
```
keszybz: `"user.slice"` → `SPECIAL_USER_SLICE`

src/shared/sleep-config.c
```diff
@@ -526,6 +553,71 @@ int get_total_suspend_interval(Hashmap *last_capacity, usec_t *ret) {
...
+
+        assert(ret);
+
+        /* implementation via dmi/entries */
+        r = read_full_virtual_file(SYS_ENTRY_RAW_FILE_TYPE1, &s, &readsize);
+        if (r < 0)
+                return log_debug_errno(r, "Unable to read %s, ignoring: %m", SYS_ENTRY_RAW_FILE_TYPE1);
+
+        if (readsize < 25 || s[1] < 25)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Only read %zu bytes from %s (expected 24)", readsize, SYS_ENTRY_RAW_FILE_TYPE1);
```
keszybz: Hmm, this message could say "only read 24 bytes, expected 24", i.e. there's an off-by-one somewhere here.

src/sleep/sleep.c
```diff
@@ -264,78 +264,117 @@ static int execute(
...
-                        log_debug_errno(r, "Battery capacity percentage unavailable, cannot estimate discharge rate: %m");
-                        if (!woken_by_timer)
-                                return 0;
-                        break;
-                }
+                if (wakeup_type_bit != 3)
+                        /* For APM Timer wakeup, system should hibernate else wakeup */
+                        return 0;
+        } else
+                while (battery_is_low() == 0) {
```
keszybz: This `while` loop should be a separate function.

src/shared/sleep-config.c
cerebro1: yeah. fixed that here https://github.com/systemd/systemd/pull/23895
that commit is not updated.

src/sleep/sleep.c
cerebro1: Could be but avoided doing that since we need to iterate over suspend with the timer here based on different helper results and then executing SUSPEND or HIBERNATE so code seemed more readable and understandable to me that way....

src/shared/sleep-config.c
cerebro1: fixed

src/sleep/sleep.c
cerebro1: @keszybz I have fixed that and pushed in acpi_btp PR, here. https://github.com/systemd/systemd/pull/23895
Once that is merged after review, i will rebase here. that should reflect here then.

bluca:
needs a rebase

JeWe37:
system.slice/Non-user processes generally were not considered here? I assume it would be more complicated to untangle what could be safely frozen there? Freezing that might be interesting in other places in the future, say for safely using LUKS suspend when suspending the system.

bluca:
> system.slice/Non-user processes generally were not considered here? I assume it would be more complicated to untangle what could be safely frozen there? Freezing that might be interesting in other places in the future, say for safely using LUKS suspend when suspending the system.

It can be done later, no reason to tie the two together.

cerebro1:
> needs a rebase

done

src/shared/sleep-config.c
```diff
@@ -46,6 +50,7 @@
 #define DISCHARGE_RATE_FILEPATH "/var/lib/systemd/sleep/battery_discharge_percentage_rate_per_hour"
 #define BATTERY_DISCHARGE_RATE_HASH_KEY SD_ID128_MAKE(5f,9a,20,18,38,76,46,07,8d,36,58,0b,bb,c4,e0,63)
 #define SYS_ENTRY_RAW_FILE_TYPE1 "/sys/firmware/dmi/entries/1-0/raw"
+#define SPECIAL_USER_SLICE "user.slice"
```
poettering: SPECIAL_USER_SLICE is already declared exactly like that in `special.h`

src/shared/sleep-config.c
```diff
@@ -126,6 +131,26 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
         return 0;
 }
 
+/* Freeze or thaw user.slice */
+int freeze_thaw_user_slice(const char **method) {
```
poettering: why a ptr to a ptr? A simple ptr would suffice?

src/sleep/sleep.c
```diff
@@ -350,10 +350,18 @@ static int custom_timer_suspend(const SleepConfig *sleep_config) {
 }
 
 static int execute_s2h(const SleepConfig *sleep_config) {
+        _cleanup_(freeze_thaw_user_slice) const char *method = "thaw";
```
poettering: The method call is called FreezeUnit() and ThawUnit(), not thaw() and freeze()

src/shared/sleep-config.c
poettering: ah, i understand why you are doing this. but please add a comment about it.

src/shared/sleep-config.c
```diff
@@ -126,6 +131,26 @@ int parse_sleep_config(SleepConfig **ret_sleep_config) {
...
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus = NULL;
+        int r;
+
+        if (method == NULL)
```
poettering: we usually write this as `if (!method)`

src/sleep/sleep.c
```diff
@@ -350,10 +350,18 @@ static int custom_timer_suspend(const SleepConfig *sleep_config) {
...
+        const char *method_freeze = "freeze";
         int r, k;
 
         assert(sleep_config);
 
+        r = freeze_thaw_user_slice(&method_freeze);
+        if (r < 0) {
+                log_debug_errno(r, "Failed to freeze unit user.slice, ignoring: %m");
+                method = NULL;
```
poettering: it appears to me this is the wrong way round. instead of *disabling* the automatic thawing via `method = NULL` here in the error path I think we should only *enable* it in the success path, i.e. do:

```
r = freeze_thaw_user_slice(…);
if (r < 0)
        log_debug_errno(…);
else
        auto_method_thaw = "ThawUnit"; /* from now on we want automatic thawing */
```

src/shared/sleep-config.h
```diff
@@ -67,6 +67,7 @@ int estimate_battery_discharge_rate_per_hour(
                 usec_t after_timestamp);
 int check_wakeup_type(void);
 int battery_trip_point_alarm_exists(void);
+int freeze_thaw_user_slice(const char **method);
```
poettering: no need to export this, make the function static

src/sleep/sleep.c
poettering: I think we should give this a better name. i.e. `auto_method_thaw` or so?

src/shared/sleep-config.c
cerebro1: fixed

src/shared/sleep-config.c
cerebro1: Sure

src/sleep/sleep.c
cerebro1: Ahh. So sorry about that. Confused it with the verb in systemctl and then totally missed to set it. Fixing it now.

src/sleep/sleep.c
cerebro1: @poettering Yes automatic thawing is good but in case freeze was not done, don't we need to skip cleanup?

I will set it to NULL then where declaring the cleanup, makes sense right?

src/shared/sleep-config.h
cerebro1: ok.

src/sleep/sleep.c
```diff
@@ -349,11 +353,39 @@ static int custom_timer_suspend(const SleepConfig *sleep_config) {
...
+}
+
 static int execute_s2h(const SleepConfig *sleep_config) {
+        _cleanup_(freeze_thaw_user_slice) const char *auto_method_thaw = NULL;
+        const char *method_freeze = "FreezeUnit";
         int r, k;
 
         assert(sleep_config);
 
+        r = freeze_thaw_user_slice(&method_freeze);
```
bluca: @poettering is there some compound literal malarky we can do here to avoid the extra fixed variable?

poettering: yes. You should be able to do:

```c
r = free_that_user_slice(&(const char*) { "FreezeUnit" });
```

(untested)

i.e. this takes a pointer of a compound literal of a pointer, that is initialized to point to a literal string.

src/sleep/sleep.c
poettering: what I am saying is that `auto_method_thaw` should default to NULL initially, and then only be set to `"ThawUnit"` once we know the freezing succeeded.

src/sleep/sleep.c
bluca: ah nice, it works

src/sleep/sleep.c
```diff
@@ -349,11 +353,39 @@ static int custom_timer_suspend(const SleepConfig *sleep_config) {
         return 1;
 }
 
+/* Freeze when invoked and thaw on cleanup */
+static int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus = NULL;
+        int r;
+
+        if (!method)
```
bluca: on cleanup *method will be set, but the value it's pointing to will be null, so you need: `if (!method || !*method)`

src/sleep/sleep.c
cerebro1: yes. done the same. :)

src/sleep/sleep.c
cerebro1: yeah both should be set. nice pointed out.

src/sleep/sleep.c
cerebro1: Yes makes better sense to do it this way

src/sleep/sleep.c
```diff
@@ -349,11 +353,38 @@ static int custom_timer_suspend(const SleepConfig *sleep_config) {
         return 1;
 }
 
+/* Freeze when invoked and thaw on cleanup */
+static int freeze_thaw_user_slice(const char **method) {
+        _cleanup_(sd_bus_error_free) sd_bus_error error = SD_BUS_ERROR_NULL;
+        _cleanup_(sd_bus_flush_close_unrefp) sd_bus *bus = NULL;
+        int r;
+
+        if (!method || *method)
```
poettering: → `if (!method || !*method)`

cerebro1: my bad. fixed

bluca (approved):

poettering:
lgtm

bluca:
How do we bypass this false positive? Any idea @mrc0mmand @keszybz ?

```
../src/sleep/sleep.c:377:55: error: variable 'auto_method_thaw' set but not used [-Werror,-Wunused-but-set-variable]
        _cleanup_(freeze_thaw_user_slice) const char *auto_method_thaw = NULL;
```

mrc0mmand:
> How do we bypass this false positive? Any idea @mrc0mmand @keszybz ?
> 
> ```
> ../src/sleep/sleep.c:377:55: error: variable 'auto_method_thaw' set but not used [-Werror,-Wunused-but-set-variable]
>         _cleanup_(freeze_thaw_user_slice) const char *auto_method_thaw = NULL;
> ```


```c
_unused_ _cleanup_(freeze_thaw_user_slice) const char *auto_method_thaw = NULL
```

we do this in several places already with this "unusual" cleanup handlers.

PR #24338, opened 2022-08-17
fs-util: make touch() an inline function

poettering:


DaanDeMeyer:
Any specific reason? I don't expect this to be a very hot function

DaanDeMeyer (approved):

poettering:
> Any specific reason? I don't expect this to be a very hot function

It' just what we typically do when we have one function that is just a direct wrapper around another one with some arguments filled in. Such "single-line" wrapper functions should probably always be static inline

yuwata (approved):

Issue #23866, opened 2022-06-29
TEST-50-DISSECT is unstable


mrc0mmand:
### systemd version the issue has been seen with

latest main

### Used distribution

Arch Linux

### Linux kernel version used

5.18.7-arch1-1

### Expected behaviour you didn't see

TEST-50-DISSECT keeps failing sporadically with:

```
[   36.180263] H testsuite-50.sh[393]: + grep -q '{"rw":"ro","designator":"root","partition_uuid":"d0c470d6-cc53-34e4-2134-6067c0486fdf","partition_label":"Root Partition","fstype":"squashfs","architecture":"x86-64","verity":"yes",'
[   36.181339] H systemd[1]: dev-loop0p1.device: Changed dead -> plugged
[   36.181484] H testsuite-50.sh[392]: + systemd-dissect --json=short --root-hash d0c470d6cc5334e421346067c0486fdff46c9923d3f9ddc4fe7df7627ac86930 /tmp/tmp.S8oWPP/minimal_0.gpt
[   36.183917] H systemd[1]: sys-devices-virtual-block-loop0-loop0p1.device: Changed dead -> plugged
[   36.184271] H systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=UnitNew cookie=1431 reply_cookie=0 signature=so error-name=n/a error-message>
[   36.184505] H systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=UnitNew cookie=1432 reply_cookie=0 signature=so error-name=n/a error-message>
[   36.184663] H systemd[1]: dev-disk-by\x2ddiskseq-13.device: Job 608 dev-disk-by\x2ddiskseq-13.device/nop finished, result=done
[   36.184980] H systemd[1]: Sent message type=signal sender=n/a destination=n/a path=/org/freedesktop/systemd1 interface=org.freedesktop.systemd1.Manager member=JobRemoved cookie=1433 reply_cookie=0 signature=uoss error-name=n/a error-me>
[   36.185263] H systemd[1]: dev-loop0.device: Job 609 dev-loop0.device/nop finished, result=done
[...]

PR #24342, opened 2022-08-17
test: expect libdevmapper failure in TEST-50-DISSECT

bluca:
libdevmapper/device mapper driver can return semi-random failures when
opening verity devices, and we have fallback code to deal with it.
But the test was not expecting the fallback path, so it became unreliable.

Fixes https://github.com/systemd/systemd/issues/23866


mrc0mmand (approved):
LGTM, thanks!

PR #24350, opened 2022-08-18
docs: Recommend pkexec over using an askpass program with sudo

daandemeyer:
Less configuration required than downloading and configuring an
askpass program.


keszybz:
Looks reasonable, but since I don't use clangd, I'll let other people make the decision.

bluca (approved):
just tested this, it works, nice!

PR #24351, opened 2022-08-18
support for signed TPM2 PCR policies in cryptsetup/cryptenrolls/credentials

poettering:
This adds support for a new kinda of TPM2 policy to the TPM2 support in the credentials and cryptsetup subsystems. Specifically, it allows binding secrets to signatures of PCR hashes instead of the literal PCR hash values. This makes PCR-based policies a lot more useful, as it addresses the PCR "brittleness" problem, in particular when it comes to updating the OS and its kernel.

This builds on previous work to make PCR measurements deterministic: only if expected hashes can be pre-calculated they can also be signed by the vendor of an OS beforehand.

This adds the new concept to three areas:

1. The `systemd-measure` tool gained a new `sign` verb, that pre-calculates PCR measurements for unified kernel, and then signs it with a provided PEM private key. This is done for multiple banks at once, and the results are written to a JSON object, which is written to stdout.

2. The `systemd-cryptenroll`, `systemd-repart` and `systemd-creds` tools now optionally accept a PEM public key to lock encrypted objects to. This means these objects cannot be decrypted ("unsealed") unless a cryptographic signature matching the PEM public key can be provided that proves that the current PCR measurements are good.

3. The `systemd-cryptsetup` and `systemd-creds` tools now accept a signature JSON object (as generated in step 1), that match the PEM public key used in step 2 and can be used to unlock the protected secrets.

Intended flow of operations:

1. OS vendor uses `systemd-measure` to pre-calculate and sign expected PCR values at build-time of the OS, and ships signature and PEM public key along with the kernel image.

2. Whenever a disk or credential shall be encrypted on a final system, this locked to the PEM public key.

3. Whenever a disk or credential shall be unlocked on a final system the JSON signature shipped with the OS is used to unlock it.

4. Whenever the OS/kernel is updated a new signature file is shipped as well, which once booted will allow unlocking all objects previously encrypted.

### DIFF
--- a/test/TEST-70-TPM2/test.sh
+++ b/test/TEST-70-TPM2/test.sh
@@ -20,6 +20,7 @@ test_append_files() {
         install_dmevent
         generate_module_dependencies
         inst_binary tpm2_pcrextend
+        inst_binary openssl
 }
 
 TEST_70_TPM_DEVICE="tpm-tis"
--- a/test/units/testsuite-70.sh
+++ b/test/units/testsuite-70.sh
@@ -57,6 +57,8 @@ env PASSWORD=passphrase systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=0+7 $
 tpm2_pcrextend 0:sha256=0000000000000000000000000000000000000000000000000000000000000000
 /usr/lib/systemd/systemd-cryptsetup attach test-volume $img - tpm2-device=auto,headless=1 && exit 1
 
+rm $img
+
 if [[ -e /usr/lib/systemd/systemd-measure ]]; then
     echo HALLO > /tmp/tpmdata1
     echo foobar > /tmp/tpmdata2
@@ -69,8 +71,65 @@ if [[ -e /usr/lib/systemd/systemd-measure ]]; then
 EOF
 
     /usr/lib/systemd/systemd-measure calculate --linux=/tmp/tpmdata1 --initrd=/tmp/tpmdata2 --bank=sha1 --bank=sha256 --bank=sha384 --bank=sha512 | cmp - /tmp/result
+
+    cat >/tmp/result.json <<EOF
+{"sha1":[{"pcr":11,"hash":"5177e4ad69db92192c10e5f80402bf81bfec8a81"}],"sha256":[{"pcr":11,"hash":"37b48bd0b222394dbe3cceff2fca4660c4b0a90ae9369ec90b42f14489989c13"}],"sha384":[{"pcr":11,"hash":"5573f9b2caf55b1d0a6a701f890662d682af961899f0419cf1e2d5ea4a6a68c1f25bd4f5b8a0865eeee82af90f5cb087"}],"sha512":[{"pcr":11,"hash":"961305d7e9981d6606d1ce97b3a9a1f92610cac033e9c39064895f0e306abc1680463d55767bd98e751eae115bdef3675a9ee1d29ed37da7885b1db45bb2555b"}]}
+EOF
+
+    /usr/lib/systemd/systemd-measure calculate --linux=/tmp/tpmdata1 --initrd=/tmp/tpmdata2 --bank=sha1 --bank=sha256 --bank=sha384 --bank=sha512 -j | diff -u - /tmp/result.json
 else
-    echo "/usr/lib/systemd/systemd-measure not found, skipping the test case"
+    echo "/usr/lib/systemd/systemd-measure not found, skipping PCR policy test case"
+fi
+
+if [ -e /usr/lib/systemd/systemd-measure ] && \
+         [ -f /sys/class/tpm/tpm0/pcr-sha1/11 ] && \
+         [ -f /sys/class/tpm/tpm0/pcr-sha256/11 ]; then
+    # Generate key pair
+    openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 -out "/tmp/pcrsign-private.pem"
+    openssl rsa -pubout -in "/tmp/pcrsign-private.pem" -out "/tmp/pcrsign-public.pem"
+
+    # Sign current PCR state with it
+    /usr/lib/systemd/systemd-measure sign --current --bank=sha1 --bank=sha256 --private-key="/tmp/pcrsign-private.pem" --public-key="/tmp/pcrsign-public.pem" | tee "/tmp/pcrsign.sig"
+    dd if=/dev/urandom of=/tmp/pcrtestdata bs=1024 count=64
+    systemd-creds encrypt /tmp/pcrtestdata /tmp/pcrtestdata.encrypted --with-key=host+tpm2-with-public-key --tpm2-public-key="/tmp/pcrsign-public.pem"
+    systemd-creds decrypt /tmp/pcrtestdata.encrypted - --tpm2-signature="/tmp/pcrsign.sig" | cmp - /tmp/pcrtestdata
+
+    # Invalidate PCR, decrypting should fail now
+    tpm2_pcrextend 11:sha256=0000000000000000000000000000000000000000000000000000000000000000
+    systemd-creds decrypt /tmp/pcrtestdata.encrypted - --tpm2-signature="/tmp/pcrsign.sig" > /dev/null && { echo 'unexpected success'; exit 1; }
+
+    # Sign new PCR state, decrypting should work now.
+    /usr/lib/systemd/systemd-measure sign --current --bank=sha1 --bank=sha256 --private-key="/tmp/pcrsign-private.pem" --public-key="/tmp/pcrsign-public.pem" > "/tmp/pcrsign.sig2"
+    systemd-creds decrypt /tmp/pcrtestdata.encrypted - --tpm2-signature="/tmp/pcrsign.sig2" | cmp - /tmp/pcrtestdata
+
+    # Now, do the same, but with a cryptsetup binding
+    truncate -s 20M $img
+    cryptsetup luksFormat -q --pbkdf pbkdf2 --pbkdf-force-iterations 1000 --use-urandom $img /tmp/passphrase
+    systemd-cryptenroll --unlock-key-file=/tmp/passphrase --tpm2-device=auto --tpm2-public-key="/tmp/pcrsign-public.pem" --tpm2-signature="/tmp/pcrsign.sig2" $img
+
+    # Check if we can activate that (without the token module stuff)
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=0 /usr/lib/systemd/systemd-cryptsetup attach test-volume2 $img - tpm2-device=auto,tpm2-signature="/tmp/pcrsign.sig2",headless=1
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=0 /usr/lib/systemd/systemd-cryptsetup detach test-volume2
+
+    # Check if we can activate that (and a second time with the the token module stuff enabled)
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=1 /usr/lib/systemd/systemd-cryptsetup attach test-volume2 $img - tpm2-device=auto,tpm2-signature="/tmp/pcrsign.sig2",headless=1
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=1 /usr/lib/systemd/systemd-cryptsetup detach test-volume2
+
+    # After extending the PCR things should fail
+    tpm2_pcrextend 11:sha256=0000000000000000000000000000000000000000000000000000000000000000
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=0 /usr/lib/systemd/systemd-cryptsetup attach test-volume2 $img - tpm2-device=auto,tpm2-signature="/tmp/pcrsign.sig2",headless=1 && { echo 'unexpected success'; exit 1; }
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=1 /usr/lib/systemd/systemd-cryptsetup attach test-volume2 $img - tpm2-device=auto,tpm2-signature="/tmp/pcrsign.sig2",headless=1 && { echo 'unexpected success'; exit 1; }
+
+    # But once we sign the current PCRs, we should be able to unlock again
+    /usr/lib/systemd/systemd-measure sign --current --bank=sha1 --bank=sha256 --private-key="/tmp/pcrsign-private.pem" --public-key="/tmp/pcrsign-public.pem" > "/tmp/pcrsign.sig3"
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=0 /usr/lib/systemd/systemd-cryptsetup attach test-volume2 $img - tpm2-device=auto,tpm2-signature="/tmp/pcrsign.sig3",headless=1
+    /usr/lib/systemd/systemd-cryptsetup detach test-volume2
+    SYSTEMD_CRYPTSETUP_USE_TOKEN_MODULE=1 /usr/lib/systemd/systemd-cryptsetup attach test-volume2 $img - tpm2-device=auto,tpm2-signature="/tmp/pcrsign.sig3",headless=1
+    /usr/lib/systemd/systemd-cryptsetup detach test-volume2
+
+    rm $img
+else
+    echo "/usr/lib/systemd/systemd-measure or PCR sysfs files not found, skipping signed PCR policy test case"
 fi
 
 echo OK >/testok


poettering:
(still lacks docs + tests)

poettering:
Tests and docs added now. This should be ready for review now.

src/shared/openssl-util.c
```diff
@@ -109,6 +109,62 @@ int rsa_pkey_to_suitable_key_size(
...
+
+        sz = i2d_PublicKey(pk, NULL);
+        if (sz <= 0)
+                return log_debug_errno(SYNTHETIC_ERRNO(EINVAL), "Unable to convert public key to DER format");
+
+        dd = d = malloc(sz);
+        if (!d)
+                return log_oom_debug();
+
+        lsz = i2d_PublicKey(pk, &dd);
```
bluca, line 135: is there no error returned from these APIs?

man/systemd-measure.xml
```diff
@@ -84,29 +105,47 @@
...
+        <command>status</command> output. May be used more then once to specify multiple banks. If not
+        specified, defaults to the four banks <literal>sha1</literal>, <literal>sha256</literal>,
+        <literal>sha384</literal>, <literal>sha512</literal>.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><option>--private-key=PATH</option></term>
+        <term><option>--public-key=PATH</option></term>
+
+        <listitem><para>These switches take paths to a pair of PEM encoded RSA key files, for use with
```
bluca, line 137: This is great for local usage, development and testing, but for production usage usually you never sign with a plain text private key, and instead go through either an openssl engine (like pkcs11) or do a separate, discrete sign step.
How difficult would it be to implement engine support here? Alternatively, does the `calculate` verb output exactly what `sign` takes in to sign, so that offline signatures can be done?

src/shared/tpm2-util.c
```diff
@@ -670,13 +671,197 @@ static int tpm2_make_encryption_session(
...
+        e_bytes = BN_num_bytes(e);
+        assert_se(e_bytes > 0);
+        if ((size_t) e_bytes > sizeof_field(TPM2B_PUBLIC, publicArea.parameters.rsaDetail.exponent))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "RSA exponent too large for TPM2 public key object.");
+
+        *output = (TPM2B_PUBLIC) {
+                .size = sizeof(TPMT_PUBLIC),
+                .publicArea = {
+                        .type = TPM2_ALG_RSA,
+                        .nameAlg = TPM2_ALG_SHA256,
```
bluca, line 767: does the hash algo need to be configurable, rather than hard-coded? They have a tendency to eventually be deprecated

src/shared/tpm2-util.c
```diff
@@ -670,13 +671,197 @@ static int tpm2_make_encryption_session(
...
+                return log_error_errno(SYNTHETIC_ERRNO(ENXIO), "Signature lacks data for PCR bank '%s'.", k);
+
+        if (!json_variant_is_array(b))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL), "Bank data is not a JSON array.");
+
+        /* Now iterate through all signatures known for this bank */
+        JSON_VARIANT_ARRAY_FOREACH(i, b) {
+                _cleanup_free_ void *fpj_data = NULL, *polj_data = NULL;
+                JsonVariant *maskj, *fpj, *sigj, *polj;
+                size_t fpj_size, polj_size;
```
bluca, line 835: these variables names in this block are a bit of a mouthful to read, but I don't have better suggestions either

src/shared/tpm2-util.c
```diff
@@ -688,33 +873,44 @@ static int tpm2_make_pcr_session(
...
 
-                tpm2_pcr_mask_to_selection(pcr_mask, pcr_bank, &pcr_selection);
-        } else {
-                TPMI_ALG_HASH h;
+                if (pcr_bank != UINT16_MAX) {
+                        r = tpm2_pcr_mask_good(c, pcr_bank, hash_pcr_mask|pubkey_pcr_mask);
+                        if (r < 0)
+                                return r;
+                        if (r == 0)
+                                log_notice("Selected TPM2 PCRs are not initialized on this system, most likely due to a firmware issue. PCR policy is effectively not enforced. Proceeding anyway.");
```
bluca: This sounds bad, should it be maybe at least log_warning?

man/systemd-creds.xml
```diff
@@ -333,6 +333,40 @@
...
+        <filename>/etc/systemd/</filename>, <filename>/run/systemd/</filename>,
+        <filename>/usr/lib/systemd/</filename> (searched in this order), it is automatically used. The
+        <option>--tpm2-public-key-pcrs=</option> option takes a list of TPM2 PCR indexes to bind to (same
+        syntax as <option>--tpm2-pcrs=</option> described above). If not specified defaults to 11 (i.e. this
+        binds the policy to any unified kernel image for which a PCR signature can be provided).</para>
+
+        <para>Note the difference between <option>--tpm2-pcrs=</option> and
+        <option>--tpm2-public-key-pcrs=</option>: the former binds decryption to the current, specific PCR
+        values; the latter binds decryption to any set of PCR values for which a signature by the specified
+        public key can be provided. The latter is hence more useful in scenarios where software updates shell
```
bluca: /shell/shall/

man/systemd-creds.xml
```diff
@@ -333,6 +333,40 @@
...
+        be possible without losing access to all previously encrypted secrets.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><option>--tpm2-signature=</option><arg>PATH</arg></term>
+
+        <listitem><para>Takes a path to a TPM2 PCR signature file as generated by the
+        <citerefentry><refentrytitle>systemd-measure</refentrytitle><manvolnum>1</manvolnum></citerefentry>
+        tool and maybe be used to allow the <command>decrypt</command> command to decrypt credentials that
```
bluca: "and that may be used"

man/systemd-creds.xml
```diff
@@ -333,6 +333,40 @@
...
+        are bound to specific signed PCR values. If this this is not specified explicitly, and a credential
+        with a signed PCR policy is attempted to be decrypted a suitable signature file
```
bluca: "be decrypted,"

src/cryptenroll/cryptenroll-tpm2.c
```diff
@@ -157,10 +162,29 @@ int enroll_tpm2(struct crypt_device *cd,
...
+                pubkey_pcr_mask = 0;
+        } else {
+                /* Also try to load the signature JSON object, to verify that our enrollment will work. This is optional however. */
+
+                r = tpm2_load_pcr_signature(signature_path, &signature_json);
+                if (r < 0) {
+                        if (signature_path || r != -ENOENT)
+                                return log_debug_errno(r, "Failed to read TPM PCR signature: %m");
+
+                        log_debug_errno(r, "Failed to read TPM2 PCR signature, proceeding without: %m");
```
bluca, line 180: shouldn't this be info at least?

src/cryptsetup/cryptsetup.c
```diff
@@ -398,6 +400,16 @@ static int parse_one_option(const char *option) {
                 if (r < 0)
                         return r;
 
+        } else if ((val = startswith(option, "tpm2-signature="))) {
+
+                if (!path_is_absolute(val))
+                        return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                               "Header path \"%s\" is not absolute, refusing.", val);
```
bluca: Header?

src/shared/tpm2-util.c
```diff
@@ -1866,6 +1866,150 @@ int tpm2_make_luks2_json(
...
+                size_t *ret_blob_size,
+                void **ret_policy_hash,
+                size_t *ret_policy_hash_size,
+                TPM2Flags *ret_flags) {
+
+        _cleanup_free_ void *blob = NULL, *policy_hash = NULL, *pubkey = NULL;
+        size_t blob_size = 0, policy_hash_size = 0, pubkey_size = 0;
+        uint32_t hash_pcr_mask = 0, pubkey_pcr_mask = 0;
+        uint16_t primary_alg = TPM2_ALG_ECC; /* ECC was the only supported algorithm in systemd < 250, use that as implied default, for compatibility */
+        uint16_t pcr_bank = UINT16_MAX; /* default: pick automatically */
```
bluca, line 1903: shouldn't this be the sha256 bank, given the comment below about being a new option post v250 and defaulting to sha256?

src/shared/tpm2-util.c
poettering: this can be added later I guess. But given that tpm2 baseline says only sha1 and sha256 are a must and sha1 is crap there's very little reason to use anything else here.

Similar for RSA: it's the only thing the tpms have to support. We currently have code in other places for tpm2 code to prefer ECC in various cases (because key gen is so much faster with ecc), but there we have to add a lot of fallback code to deal with cases where ECC is not available. I'd rather not do that needlessly where we can avoid it, at least initially. Since here the performance doesn't matter (the rsa keys are generated on host here, not on TPM), i'd stick to RSA.

The thing here is that that OS vendors generate the keys and IRL this means they basicaly *have* to use RSA 2048 for that, since that's the only thing that is reasonably safe *and* universally supported. If they pick ECC keys instead then this would mean their PCR signatures are useless on many systems, which defeats the whole point...

src/shared/openssl-util.c
poettering: they return a negative size.

bluca: nothing useful for logging, like an errno or similar internal implementation?

man/systemd-measure.xml
poettering: > This is great for local usage, development and testing, but for production usage usually you never sign with a plain text private key, and instead go through either an openssl engine (like pkcs11) or do a separate, discrete sign step. How difficult would it be to implement engine support here? 

We should probably add `--private-key-pkcs11=` as alternative for `--private-key=` when the time comes, which takes a PCS11 uri to the module/device/key to use. Should be really easy to do actually. 

(Which reminds me, mkosi should learn PKCS11 for signing unified kernels in general. After all I have the suspicion the key to SecureBoot-sign the unified kernel and to sign the PCR values should be the same, and for both you want PKCS11)

> Alternatively, does the calculate verb output exactly what sign takes in to sign, so that offline signatures can be done?

Well, almost. `calculate` displays individual PCR values, while you `sign` a combination of PCR values. Given that we typically only want to sign PCR 11 (i.e. the one that corresponds to the unified kernel image) – i.e. a single PCR – the effective conceptual difference for us is minimial, but it does have an effect on the JSON structure since in the case of a PCR dump you have PCR indexes as numbers, while in case of the PCR signature you want them as a list of numbers.

The JSON objects are designed to be very similar for the PCR dump and for the PCR signature, but they are different due to this nature that signatures cover multiple PCR indexes, but dumps have one entry for each PCR.

poettering: (but yeah, teaching "systemd-measure sign" to take a JSON PCR dump generated via "calculate" or "status" as input should be easy, and is precisely what the JSON PCR dump format is good for.)

src/cryptenroll/cryptenroll-tpm2.c
poettering: This line is only hit if the user did not specify an explicit signature, and we then searched for one installed in the fs somewhere, and didn't find any. This is actually the common case on existing systems: they won't have a signature file installed, because simply noone builds systems with signed PCR values yet.

Hence, this really *is* a debugging thing: it will tell you explicitly that we failed to auto-discover a signature file then.

src/shared/tpm2-util.c
poettering: if the bank is not specified we communicate that as UINT16_MAX to the caller. Which tells the caller "figure it out yourself". The caller then checks which banks the TPM and the firmware support, and picks SHA256 if both support it, and SHA1 otherwise.

What's important is that during enrollment and during unlocking we use the same bank. If we auto-detect the bank we might come to different results though (for example: when volume was enrolled fimrware could only do SHA1, but a firmware update was done, which added SHA256 support and the TPM always could do it). Hence, in current versions we generally store away the bank we picked to avoid any discrepancies there. But then we need a way to communicate the "please audo-detect" mode which is what UINT16_MAX does here.

man/systemd-measure.xml
bluca: pkcs11 would cover many cases, but not all, in many cases really only an engine is available (think of the datacenter build system, where engines take a hash, ship it off to a secure box that checks some policy from the caller and returns the signature for the hash if it passes)

Can we at least have a flag for --calculate that results in the exact same payload that `sign` would sign, so that you can do discrete signing in absence of engine support?

poettering: well, there are pkcs11 remoting solutions, and I think that's actually a relatively nice box. But as mentioned, we can certainly add a mode where "sign" takes the output of "status" or "measure" as input. That way you can run "measure" on the build system, ship its result to a signing box, where it then does "sign", and returns the result.

bluca: We can't assume the remote box can use `systemd-measure` though, it might not even be running Linux

src/shared/tpm2-util.c
bluca: got it, thanks

man/systemd-measure.xml
poettering: Well, so, the actual signing operation is trivial once you know what you need to sign. We use a local TPM to generate the hash to sign currently (which can be swtpm though). We could certainly stop after calculating what to sign, and generate an "incomplete signature" basically, that contains the payload to sign, but not the actual signature. This could then be shipped off to some server which "completes" it by signing the data and returning it with the signature field filled in. The JSON format is ready for that.

But I'd cross the bridge once we get there. There are lot of details to figure out: i.e. you probably want the server side to validate some things before it signs it, so that you can't make the server sign stuff is isn't supposed to sign. i.e. it should probably check in some form that what we hand it actually is a PCR 11 unified kernel blob, and nothing else. But how such verification can work if the server is "dumb" and doesn't have systemd-measure is not clear to me.

bluca: I don't think we need to worry about policy and checks for the 'offline' mode (but we should for engine/pkcs11). In that case, it's up to the caller to do the right thing, we should just provide the content as-is.

src/shared/openssl-util.c
poettering: added

poettering:
CI failure appear unrelated

man/systemd-creds.xml
```diff
@@ -333,6 +333,40 @@
...
+        <option>--tpm2-public-key-pcrs=</option> option takes a list of TPM2 PCR indexes to bind to (same
+        syntax as <option>--tpm2-pcrs=</option> described above). If not specified defaults to 11 (i.e. this
+        binds the policy to any unified kernel image for which a PCR signature can be provided).</para>
+
+        <para>Note the difference between <option>--tpm2-pcrs=</option> and
+        <option>--tpm2-public-key-pcrs=</option>: the former binds decryption to the current, specific PCR
+        values; the latter binds decryption to any set of PCR values for which a signature by the specified
+        public key can be provided. The latter is hence more useful in scenarios where software updates shall
+        be possible without losing access to all previously encrypted secrets.</para></listitem>
+      </varlistentry>
```
williamcroberts, line 355: What does the actual update process look like to generate the new expected set of PCRs?

man/systemd-measure.xml
```diff
@@ -61,11 +63,30 @@
...
+        <option>--private-key=</option> and <option>--public-key=</option>. This will write a JSON object to
+        standard output that contains signatures for all specified PCR banks (see
+        <option>--pcr-bank=</option>) below, which may be used to unlock encrypted credentials (see
+        <citerefentry><refentrytitle>systemd-creds</refentrytitle><manvolnum>1</manvolnum></citerefentry>) or
+        LUKS volumes (see
+        <citerefentry><refentrytitle>systemd-cryptsetup@.service</refentrytitle><manvolnum>8</manvolnum></citerefentry>). This
+        allows binding secrets to a set of kernels for which such PCR 11 signatures can be provided.</para>
+
+        <para>Note that a TPM2 device must be available for this signing to take place, even though the
+        result is not tied to any TPM2 device or its state.</para></listitem>
```
williamcroberts, line 89: Why? Is this just for simplicity? The policy digests and signing can occur without a TPM2 device, do you just need examples?

man/systemd-measure.xml
williamcroberts: I would just have the ability to load an engine or a provider. OpenSSL has a PKCS11 Engine through p11 project. So you can go from openssl --> engine --> pkcs11 library. For OpenSSL3 supposedly OSSL is implementing a pkcs11 provider per https://github.com/OpenSC/libp11/issues/432

src/boot/measure.c
```diff
@@ -464,6 +525,235 @@ static int verb_calculate(int argc, char *argv[], void *userdata) {
...
+                                session_handle,
+                                ESYS_TR_NONE,
+                                ESYS_TR_NONE,
+                                ESYS_TR_NONE,
+                                &pcr_policy_digest);
+                if (rc != TSS2_RC_SUCCESS) {
+                        r = log_error_errno(SYNTHETIC_ERRNO(ENOTRECOVERABLE),
+                                            "Failed to get policy digest from TPM: %s", sym_Tss2_RC_Decode(rc));
+                        goto finish;
+                }
```
williamcroberts, line 669: The newer libpolicy on tpm2-tss will calculate all this without round tripping to the TPM. Probably too new for usage here, but calculating this by hand is pretty easy:
`policyDigestnew ≔ HpolicyAlg(policyDigestold || TPM_CC_PolicyPCR || pcrs || digestTPM)`

Where:
- policyDigestold is 32 0's
- TPM_CC_PolicyPCR is constant `TPM2_CC_PolicyPCR` from `tss2_tpm2_types.h` aka value 0x017f
- pcrs: libmu marshalled `TPML_PCR_SELECTION`.
- digestTPM: The expected digest of all the PCRs

Sample code can be found in this function:
  - https://github.com/tpm2-software/tpm2-tss/blob/cdeba44f2273873ed73f71c75dfed9738ab36691/src/tss2-fapi/ifapi_policy_calculate.c#L84

You'd be able to make setup TPM agnostic. Although you could just fire up a simulator for all of this.

src/boot/measure.c
```diff
@@ -464,6 +525,235 @@ static int verb_calculate(int argc, char *argv[], void *userdata) {
...
+
+                _cleanup_free_ void *sig = malloc(ss);
+                if (!ss) {
+                        r = log_oom();
+                        goto finish;
+                }
+
+                if (EVP_DigestSignFinal(mdctx, sig, &ss) != 1) {
```
williamcroberts, line 705: I'm confused Up here https://github.com/systemd/systemd/pull/24351/files#diff-087f16a06bf226a0017c61f4cdb346ae9a6895d4a18dc5587d30675d7686fafbR88 it says that TPM is needed for signing, but it looks like you sign the digest with OpenSSL.

src/creds/creds.c
```diff
@@ -832,13 +859,34 @@ static int parse_argv(int argc, char *argv[]) {
                         arg_tpm2_device = streq(optarg, "auto") ? NULL : optarg;
                         break;
 
-                case ARG_TPM2_PCRS:
+                case ARG_TPM2_PCRS: /* For fixed hash PCR policies only */
                         r = tpm2_parse_pcr_argument(optarg, &arg_tpm2_pcr_mask);
```
williamcroberts, line 863: Looks existing, but I was wondering what bank you select on? Does it default to the SHA256 bank. Perhaps in the future support tpm2-tools style with the bank selector would be a step towards making this algorithm agile.

Example: `sha1:3,4+sha256:all`

src/boot/measure.c
```diff
@@ -464,6 +525,235 @@ static int verb_calculate(int argc, char *argv[], void *userdata) {
...
+                tpm2_pcr_mask_to_selection(1 << TPM_PCR_INDEX_KERNEL_IMAGE, tpmalg, &pcr_selection);
+
+                rc = sym_Esys_PolicyPCR(
+                                c.esys_context,
+                                session_handle,
+                                ESYS_TR_NONE,
+                                ESYS_TR_NONE,
+                                ESYS_TR_NONE,
+                                &intermediate_digest,
+                                &pcr_selection);
```
williamcroberts, line 650: AFAICT `measure_pcr` and pcr_selection sent to `Esys_PolicyPCR` will be measured from the current running system, is that right? If so why? When I have done this I supply both the PCR selection and the PCR values to be expected, ie calculated from a known good golden machine.

Typically I do this somewhere and have 0 dependencies from anything on the TPM. But I am likely missing something here, as usual.

src/cryptenroll/cryptenroll.c
```diff
@@ -114,6 +120,13 @@ static int help(void) {
                "                       Enroll a TPM2 device\n"
                "     --tpm2-pcrs=PCR1+PCR2+PCR3+…\n"
                "                       Specify TPM2 PCRs to seal against\n"
+               "     --tpm2-public-key=PATH\n"
+               "                       Enroll signed TPM2 PCR policy against PEM public key\n"
+               "     --tpm2-public-key-pcrs=PCR1+PCR2+PCR3+…\n"
```
williamcroberts, line 125: Theirs two different formats of PCR values in the tools:
1. is just a binary list of PCR values with no extra metdata.
2. is a binary selection and values.

It would be nice to be able to take a file from a tool measuring the golden system with values and pass it back. Happy to add another format to tpm2-tools as well.

Without specifying the values, and just the selection, you have to go to the TPM and IIUC, `cryptenroll` would be done on the build of the OS.

src/cryptsetup/cryptsetup-tokens/cryptsetup-token-systemd-tpm2.c
```diff
@@ -156,45 +169,66 @@ _public_ void cryptsetup_token_dump(
...
+        int r;
 
         assert(json);
 
-        r = parse_luks2_tpm2_data(json, UINT32_MAX, &pcr_mask, &pcr_bank, &primary_alg, &base64_blob, &hex_policy_hash, &flags);
+        r = json_parse(json, 0, &v, NULL, NULL);
+        if (r < 0)
+                return (void) crypt_log_debug_errno(cd, r, "Failed to parse " TOKEN_NAME " JSON object: %m");
+
+        r = tpm2_parse_luks2_json(
```
williamcroberts, line 187: IIUC this generates a JSON that is shoved into the luks2 header so systemd can extract info later?

williamcroberts (commented):
Most of questions are around the flow, I looked at the test suite and that did help further my understanding. I think the setup needs to be accomplished without a TPM, ie Policy commands and PCR state. I would imagine building images on a machine that don't match the target machine but I could be misunderstanding. At that point, a quick call might be easier. We have a maintainers sync on teams, which works from a browser I could add you to do it in person.

src/cryptsetup/cryptsetup-tokens/cryptsetup-token-systemd-tpm2.c
poettering: This part *parses* the data from the LUKS2 header, and then uses it to unseal the LUKS key via the TPM2.

williamcroberts: Yeah I didn't mean to say generates (I was looking at A and commenting at B), but this is the data structure that gets generated and shoved into the LUKS header for systemd later? So when discussing verifying the session theirs an ESYS_TR I need to generate and storage would be via this JSON in the LUKS header?

src/creds/creds.c
poettering: The `systemd-creds` tool can be used to encrypt so called "credentials" that can be passed to system services, and are then decrypted only when those services are activated. This encryption/decryption combines a key we store in the file system with one we unseal via the TPM. Depending on the environment we might only use one of these keys. ("Credentials" are entirely independent from LUKS disk encryption btw, different concept)

Encryption/decryption is symmetric and supposed to happen on the system these credentials are stored and used. Thus, when we encrypt we pick the most suitable PCR bank. Specifically this means SHA256 if it is enabled in the TPM, and looks initialized. If SHA256 is disabled or doesn't look used (i.e. is all zeroes for the selected PCRs) we'll look at the SHA1 bank instead. If that one also doesn't look used we'll go for the SHA256 one after all, even if it doesn't look used. We won't consider the other banks if they exist.

We started doing this dance after getting reports from people who have systems that claim to support SHA256 banks but the firmware doesn't actually measure anything into them, and only measures to the SHA1 bank. 

The bank used is stored in the encrypted credential metadata. So when decrypting the automatic selection of the bank used is not repeated.

src/boot/measure.c
poettering: No. measure_pcr() is the function that calculates the expected PCR values from the components of a unified kernel image. The input of that function is a bunch of data files and it outputs a set of hashes for PCR 11, one for each selected bank.

I guess calculate_expected_pcr() might be a better name, if you want...

Anyway, measure_pcr() will generate data without talking to any TPM and independent of any system state. It operates solely on the kernel files you give it.

poettering: PCR 11 because that's the PCR I grabbed for measuring unified kernel images into. it apparently isn't used on Linux systems otherwise so far. By picking a PCR nothing else uses we can pre-calculate expected values easily.

src/boot/measure.c
poettering: Ah interesting. We can certainly look into calculating this without relying on a TPM. Doing this without any TPM would be excellent of course. For the testsuite we are currently using swtpm for this, and my thinking so far was that people can use that for systemd-measure too on their build systems.

But you are right, if it's easy to hash the values without the TPM we should really do this.

man/systemd-measure.xml
poettering: There's no technical reason to require a TPM here. It's mostly that I am not a TPM guru myself and what precisely the TPM hashes internally is a bit of a black box to me so far, so I just used the TPM as an implementation for the hashing, so that I don't have to figure out what precisely to hash. 

But your other comment should be enough to implement this without a TPM I guess. I'll put it on my TODO list.

man/systemd-creds.xml
poettering: 1. Build a linux kernel as ELF file
2. Build an initrd (cpio)
3. Pass these two files to `systemd-measure sign`, along with a RSA private key, which will then generate a JSON file for you basically containing a bunch of signatures for expected PCR 11 values once you booted this kernel+initrd, one for each PCR bank.
4. Use objcopy to generate a PE file from the combination of systemd-stub (an UEFI stub that does UEFI stuff and hands of to a linux kernel) + the linux kernel + the initrd cpio + the PCR signature JSON. i.e. the stuff from the first three steps above.
5. SecureBoot sign the resulting PE file

Do this for every single kernel update.

src/boot/measure.c
poettering: Well, we generate the policy hash via TPM, and then sign the resulting hash with openssl.

If this code is changed to implement the hashing of the policy client side without talking to a TPM we won't need a TPM anymore at all.

src/boot/measure.c
williamcroberts: > No. measure_pcr() is the function that calculates the expected PCR values from the components of a unified kernel image. The input of that function is a bunch of data files and it outputs a set of hashes for PCR 11, one for each selected bank.
> 
> I guess calculate_expected_pcr() might be a better name, if you want...
> 
> Anyway, measure_pcr() will generate data without talking to any TPM and independent of any system state. It operates solely on the kernel files you give it.

Ahh I see it's controlled by an argument for current:
```C
if (arg_current) {
 // do it through sysfs
else
  // do it by hand
```

src/cryptenroll/cryptenroll.c
poettering: So, there's a certain asymmetry here: the systemd-measure tool only pre-calculates expected PCR 11 value from a unified kernel image and optionally signs the result, that's really the only thing it can do. It hardcodes PCR 11. You (currently) can't use it to sign anything else than PCR 11, and anything else than a unified kernel image.

But it's only the signing side that is limited like this. The JSON format used to carry these signatures is generic, it works for any PCR index and bank, it is independent of any unified kernel stuff or so. Similar, the consumers for the JSON signature format are also fine with any PCR index for any data. Here you are looking at one of them: cryptenroll. It allows you to bind LUKS unlocking to any PCR you like. Of course if you pcik anything different than PCR 11, then tough luck but you currently have to find another tool than `systemd-measure` to pre-calculate/sign expected values for that. 

In the long run I figure `systemd-measure` should become more generic and allow generating expected hashes and signatures for any bank, any input data. i.e there should be a way where you specify a bunch of files as input for the PCR measurements, and there should be a way were you just literally specify the PCR values you want to sign. But right now I personally have no use for that, my focus was on getting the PCR 11 unified kernel stuff to work, so I only implemented that part.

The JSON format is super simple. It could be implemented in any tool really.

src/cryptsetup/cryptsetup-tokens/cryptsetup-token-systemd-tpm2.c
poettering: yupp

man/systemd-measure.xml
williamcroberts: I guess the comment, to be accurate but perhaps confusing to non-tpm savvy users, is :that its for calculating the policy digest that is signed". I can whip you up a function that given a TPML_PCR_SELECTION will give you the hash if you'd like?

src/cryptenroll/cryptenroll.c
williamcroberts: Can you provide a sample,  I could add support to tpm2-tools to dump this from a PCR read.

man/systemd-measure.xml
poettering: oh, i'd be delighted for a patch/snippet doing that!

src/cryptenroll/cryptenroll.c
poettering: The signature JSON output that `systemd-measure sign` generates, looks like this:

```
{
	"sha1" : [
		{
			"pcrs" : [
				11
			],
			"pkfp" : "b723bf7b1a14adf233d6e4781226588fd92585e975c0f7a743d2e5cdca9f6005",
			"pol" : "3025f59190d67a9881f4093c74da0f7d5ee3024980bcc4123bac0200122f3e9d",
			"sig" : "JnFGK7uxQgctDuQ+rY46R3SzfNaNMAfysd2H2L69zozwjR+OaTOMT4teoNCUq6Ok8rJq5ywjbKqN1Qe+XKaQQfMzOc7YnNJ3ML2paI4FXv3+XckSY7biuFyPCwG1l9ouerSCE2Tpy0stLXfHF/Y7iJYAdlRnlRJo8/lbUyaNARmwhX3aH2f4L2EqbO2T/Eku4NkXrSM0lC+cwTt300N5ioLPqdRujyl0nP2psTSVJ8y8lVKjmjEFLEYWXlHoEpowIysqeJX/4gAC0ZRDnCmRjq2uhMXIKIv55WSUbFZe8wdxGWFuiBtrhC2+/EUTQikywf6O5Q0t2UVjjh4lByFKPA=="
		}
	],
	"sha256" : [
		{
			"pcrs" : [
				11
			],
			"pkfp" : "b723bf7b1a14adf233d6e4781226588fd92585e975c0f7a743d2e5cdca9f6005",
			"pol" : "fd32fa22c52cfc8e1a0c29eb38519f87084cab0b04b0d8f020a4d38b2f4e223e",
			"sig" : "e56RREV9Z5RcV4wky/Aa9mjJbs3hk5IPw6T4qWtwLZM4CcOFy/jppCNY3o18k1U3tlinKjKDZF4udNNkZgZJOW74U8GeLGASw7Id0+nPlqOhPXXqm6A1tte+nmcZjkiKjDlsC8uFOWAOVZ/zZyH3MZsFbLaNjOrzuyO7HfkiKeZ/Wh141Y5oLGEMESNWz/+QX/jOOpGTH4nndRG2x2YcKJH/n+6ohxsbfO8btNc1WPerPVa9WX3QorG0ds663GQFSe/b+1cNYamJWbXsh6ghF0uNC295gqSRHef7Oo6Sn0Jl8tqKMiss4M44AqagQME9W6iAvjGjy2tKB5HrsryoYg=="
		}
	]
}

```

At the top-level there's an object, with one field for each bank. in the example above we carry signatures for the sha1 and sha256 bank. The fields are arrays of objects, each carrying one signature. These objects have the following fields:

1. `pcrs` this is an array of integers, with the PCR indexes this signature is for. `systemd-measure sign` as it stands right now hardcodes this to an array with a single element: 11.
2. `pkfp`: an hex sha256 fingerprint of the public key in DER format belonging to this signature
3. `pol`: the hash of the tpm2 policy in hex format
4. `sig`: base64 of the RSA signature

Note this object does not actually list the PCR values right now, though I guess we could add that. It only contains the policy hash.

poettering:
> Most of questions are around the flow, I looked at the test suite and that did help further my understanding. I think the setup needs to be accomplished without a TPM, ie Policy commands and PCR state. I would imagine building images on a machine that don't match the target machine but I could be misunderstanding. 

Right now, as mentioned, when pre-calculating PCRs+signing them, this PR uses the TPM as a way calculate the policy hash for me that I then sign via OpenSSL. My assumption so far would be that image builders would just run swtpm for this, which supports more banks than the real TPMs I have seen so far. But yeah, if we can easily pre-calc the policy hash even without talking to a TPM that would be even bette.

> At that point, a quick call might be easier. We have a maintainers sync on teams, which works from a browser I could add you to do it in person.

Sure, why not. Let's do this some time next week. central european works hours work for me. send invite to lpoettering at microsoft dot com.

src/cryptenroll/cryptenroll.c
poettering: and the idea is that a single of these JSON objects can carry unlimited signatures for any combination of pcr mask, public key fp and policy hash. The consumers will iterate through the array and use the signature by searching for an object with these three fields set as expected.

src/boot/measure.c
bluca: is this going to be addressed in this PR?

src/cryptenroll/cryptenroll.c
bluca: should this format be made into a spec for docs/ ?

bluca:
> > Most of questions are around the flow, I looked at the test suite and that did help further my understanding. I think the setup needs to be accomplished without a TPM, ie Policy commands and PCR state. I would imagine building images on a machine that don't match the target machine but I could be misunderstanding.
> 
> Right now, as mentioned, when pre-calculating PCRs+signing them, this PR uses the TPM as a way calculate the policy hash for me that I then sign via OpenSSL. My assumption so far would be that image builders would just run swtpm for this, which supports more banks than the real TPMs I have seen so far. But yeah, if we can easily pre-calc the policy hash even without talking to a TPM that would be even bette.

This is quite crucial, as often builds are done without any possibility of controlling devices (real or virtual), but just in a container with fakeroot

src/boot/measure.c
poettering: I'd do that later.

src/cryptenroll/cryptenroll.c
poettering: Later, yes. Now, not sure. I mean, systemd-measure itself is marked experimental and installed to /usr/lib/systemd/ and not /usr/bin/

I think we should wait a bit before doing a spec and make this an official thing.

poettering:
> This is quite crucial, as often builds are done without any possibility of controlling devices (real or virtual), but just in a container with fakeroot

swtpm is trivial to use, listens on an AF_UNIX socket. Requires not privs. It's actually extremely easy and straight-forward to run it in a container.

I mean, it would be nicer to not require it for pre-calculating PCR values, but that's a bit more convenient, that's all. Regardless if using swtpm or not, the privileges you require are the same.

poettering:
Switching to calculating the hashes in-process instead of via a real tpm or an swtpm is an implementation detail we can later make btw. It shouldn't really stop us from merging this as it is.

(Btw, earlier, before @williamcroberts suggested to simply do the policy hash calculation client-side I was actually thinking about simply forking off swtpm as a child process from systemd-measure, then ask it to do the calculation and kill it afterwards. That way the fact that swtpm was used would not even be visible to users. But of course, just hashing this ourselves would be a lot better, and remove the dep)

bluca (approved):
Forking swtpm adds a dependency on it and all its libraries, so I'd much rather the simple policy calculation was done locally instead. It can be added later.

This needs a rebase now. While there, please note in README the kernel requirements for this (pcr through /sys)

poettering:
> This needs a rebase now. While there, please note in README the kernel requirements for this (pcr through /sys)

I am not sure this really belongs in README, it's an experimental feature after all. However, I figured better even than docs is a helpful error message, hence in your case you should now see an explanatory message telling you exactly what might be the issue:

#24604

bluca:
Might be related?

```
13:08:53 FAILED TASKS:
13:08:53 -------------
13:08:53 TEST-70-TPM2
```

@mrc0mmand

mrc0mmand:
I guess it looks related:

```
[   62.010216] testsuite-70.sh[617]: + /usr/lib/systemd/systemd-measure calculate --linux=/tmp/tpmdata1 --initrd=/tmp/tpmdata2 --bank=sha1 --bank=sha256 --bank=sha384 --bank=sha512 -j
[   62.024991] testsuite-70.sh[618]: + diff -u - /tmp/result.json
[   62.170366] testsuite-70.sh[533]: + '[' -e /usr/lib/systemd/systemd-measure ']'
[   62.173086] testsuite-70.sh[533]: + '[' -f /sys/class/tpm/tpm0/pcr-sha1/11 ']'
[   62.175533] testsuite-70.sh[533]: + '[' -f /sys/class/tpm/tpm0/pcr-sha256/11 ']'
[   62.177644] testsuite-70.sh[533]: + openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 -out /tmp/pcrsign-private.pem
[   62.441699] testsuite-70.sh[619]: ..............+++++
[   62.873593] testsuite-70.sh[619]: ....................................................+++++
[   62.900379] testsuite-70.sh[533]: + openssl rsa -pubout -in /tmp/pcrsign-private.pem -out /tmp/pcrsign-public.pem
[   63.004602] testsuite-70.sh[620]: writing RSA key
[   63.026305] testsuite-70.sh[621]: + /usr/lib/systemd/systemd-measure sign --current --bank=sha1 --bank=sha256 --private-key=/tmp/pcrsign-private.pem --public-key=/tmp/pcrsign-public.pem
[   63.038697] testsuite-70.sh[622]: + tee /tmp/pcrsign.sig
[   63.206312] testsuite-70.sh[621]: TPM successfully started up.
[   63.349815] testsuite-70.sh[622]: {"sha1":[{"pcrs":[11],"pkfp":"23e7ceac0b938a850147d53675b26b08cb1a6261515b8e8b029e6abbea54a9e9","pol":"3025f59190d67a9881f4093c74da0f7d5ee3024980bcc4123bac0200122f3e9d","sig":"e9v6/jnqHtzHDIAqznbmwFQUsOZCpxbHSr9U0DFEoLMDHUvEydUoWOKx/NNqoK894EZSQVOqZ7KqDa19gsWvnHhqZKWuMNDgMYnM2xYX/UAKVyl4L/BZI701DGzw4x4KGCGSXvUJO/jKkNEfn/qn99uBZdOVF0MvnIDii1X0LOSOwcQDf/SXy3sxuaZ2Hfjv9uYCvmNJ0WJ4wVyOW7HDDz898CRFFGx0mw3fbsaiArBEgaV/RO49LXVWLXNIiJsjmTg/jrvQO3EGDrd9me6rCRdLVc7X7yObph5CFcKW22ixdXAjrQLExcBqjVXzKoAPrP2V7T2Cy2zO64NDzDexjw=="}],"sha256":[{"pcrs":[11],"pkfp":"23e7ceac0b938a850147d53675b26b08cb1a6261515b8e8b029e6abbea54a9e9","pol":"fd32fa22c52cfc8e1a0c29eb38519f87084cab0b04b0d8f020a4d38b2f4e223e","sig":"PjQNDrdl/TK4I8sJg6YEbtGqoV+FlI1CeJZ5U/MR+C35a0jwrQfzJTMoa8y1jNJuICmv+49BP20Cp6JCqj6J/Ld743tHlc9l7CgqXmSCjkOm3Qyt6YMMtxTZS2CviHFBttyhFdHkKhATviU4GWuQcEO7DuMShs5QjLCNfbsqOIAqO0yI4GoZrWG3OrYUH+XFYuX/uqTrt4boZ3AFqOB3WYsgLssVsIgCL5qlv/g9XNPGi527Npju6fC8wbVjZgWknRPStNvAa9fSxPdg77XjKXP3+sASlDMK0TN5yzNcjRzF9esyKLL5tRMCUcHrPEvaoXbUBfbXw3BdwT8Fvb3aKA=="}]}
[   63.377987] testsuite-70.sh[533]: + dd if=/dev/urandom of=/tmp/pcrtestdata bs=1024 count=64
[   63.445020] testsuite-70.sh[623]: 64+0 records in
[   63.446636] testsuite-70.sh[623]: 64+0 records out
[   63.456272] testsuite-70.sh[623]: 65536 bytes (66 kB, 64 KiB) copied, 0.0156266 s, 4.2 MB/s
[   63.461388] testsuite-70.sh[533]: + systemd-creds encrypt /tmp/pcrtestdata /tmp/pcrtestdata.encrypted --with-key=host+tpm2-with-public-key --tpm2-public-key=/tmp/pcrsign-public.pem
[   63.810756] testsuite-70.sh[624]: WARNING:esys:src/tss2-esys/api/Esys_LoadExternal.c:304:Esys_LoadExternal_Finish() Received TPM Error
[   63.813576] testsuite-70.sh[624]: ERROR:esys:src/tss2-esys/api/Esys_LoadExternal.c:108:Esys_LoadExternal() Esys Finish ErrorCode (0x000003c4)
```

```
[   63.461161] H testsuite-70.sh[533]: + systemd-creds encrypt /tmp/pcrtestdata /tmp/pcrtestdata.encrypted --with-key=host+tpm2-with-public-key --tpm2-public-key=/tmp/pcrsign-public.pem
[   63.570231] H systemd-creds[624]: Including credential name 'pcrtestdata.encrypted' in encrypted credential.
[   63.576315] H systemd-creds[624]: Including timestamp 'Thu 2022-09-08 11:54:58 UTC' in encrypted credential.
[   63.588860] H systemd-creds[624]: Unable to set file attribute 0x800000 on n/a, ignoring: Operation not supported
[   63.590816] H systemd-creds[624]: Failed to set file attributes for secrets file, ignoring: Operation not supported
[   63.597967] H systemd-creds[624]: Credential secret file '/var/lib/systemd/credential.secret' is not located on encrypted media, using anyway.
[   63.660325] H systemd-creds[624]: TPM successfully started up.
[   63.667380] H systemd-creds[624]: Creating primary key on TPM.
[   63.680770] H systemd-creds[624]: Successfully created ECC primary key on TPM.
[   63.682192] H systemd-creds[624]: Generating primary key on TPM2 took 21ms.
[   63.683394] H systemd-creds[624]: Starting HMAC encryption session.
[   63.728543] H systemd-creds[624]: Starting authentication session.
[   63.753420] H systemd-creds[624]: PCR 7 value: 3a3f780f11a4b49969fcaa80cd6e3957c33b2275
[   63.755488] H systemd-creds[624]: PCR 11 value: 0000000000000000000000000000000000000000
[   63.763990] H systemd-creds[624]: PCR 7 value: a58606a00cf970a353d116d6efeec33b2bd076641f3bda7c39a7483041b53601
[   63.765574] H systemd-creds[624]: PCR 11 value: 0000000000000000000000000000000000000000000000000000000000000000
[   63.766613] H systemd-creds[624]: TPM2 device supports SHA256 PCR bank and SHA256 PCRs are valid, yay!
[   63.796744] H systemd-creds[624]: Configuring public key based PCR policy.
[   63.810453] H testsuite-70.sh[624]: WARNING:esys:src/tss2-esys/api/Esys_LoadExternal.c:304:Esys_LoadExternal_Finish() Received TPM Error
[   63.813410] H testsuite-70.sh[624]: ERROR:esys:src/tss2-esys/api/Esys_LoadExternal.c:108:Esys_LoadExternal() Esys Finish ErrorCode (0x000003c4)
[   63.816058] H systemd-creds[624]: Failed to load public key into TPM: tpm:parameter(3):value is out of range or is not correct for the context
[   63.864219] H systemd[1]: systemd-journald.service: Received EPOLLHUP on stored fd 55 (stored), closing.
[   63.868723] H systemd[1]: Received SIGCHLD from PID 533 (bash).
[   63.869921] H systemd[1]: Child 533 (bash) died (code=exited, status=1/FAILURE)
[   63.871255] H systemd[1]: testsuite-70.service: Child 533 belongs to testsuite-70.service.
[   63.872629] H systemd[1]: testsuite-70.service: Main process exited, code=exited, status=1/FAILURE
[   63.874017] H systemd[1]: testsuite-70.service: Failed with result 'exit-code'.
```

poettering:
@williamcroberts hmm, any chance you can help me tracking down this CI issue? It's triggered by the rebase on your recent PR #24524 (which I merged earler). The LoadExternal tpm command now fails on the CI (but passed on my local system, which is much newer).

Relevant output is this:

```
Sep 08 13:54:58 H systemd-creds[624]: Configuring public key based PCR policy.
Sep 08 13:54:58 H testsuite-70.sh[624]: WARNING:esys:src/tss2-esys/api/Esys_LoadExternal.c:304:Esys_LoadExternal_Finish() Received TPM Error
Sep 08 13:54:58 H testsuite-70.sh[624]: ERROR:esys:src/tss2-esys/api/Esys_LoadExternal.c:108:Esys_LoadExternal() Esys Finish ErrorCode (0x000003c4)
Sep 08 13:54:58 H systemd-creds[624]: Failed to load public key into TPM: tpm:parameter(3):value is out of range or is not correct for the context
```

The code that this is seen in is this:

https://github.com/poettering/systemd/blob/7e07dbdbc2a6b5dadac22b4824f9ad750db4815d/src/shared/tpm2-util.c#L981

i.e. this where we load the public key that is used for signing PCR sets into the TPM. This worked fine on the CI before #24524 but fails now with that merged. 

Any idea what has changed here? do i have to change the LoadExternal call somehow to accomodate for the change from HMAC to bindkey? any help appreciated!

williamcroberts:
> Any idea what has changed here? do i have to change the LoadExternal call somehow to accomodate for the change from HMAC to bindkey? any help appreciated!

That shouldn't matter. We don't even use any session on the loadexternal call and the third parameter is the hierarchy (you can look at the command table [2] to see order). It might be the version of ESAPI used. In older versions the wrong type was used there, see commit [1] which is only available on ESAPI versions 3.0+. 

* What version of tpm2-tss is being used?
* if it's an older ESAPI and to be backwards compat **we should replace** `ESYS_TR_RH_OWNER` with  `TPM2_RH_OWNER`.

[1] https://github.com/tpm2-software/tpm2-tss/commit/9d48cc7922ad8dc3496444b04c94439c5bfbec8e
[2] https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf

mrc0mmand:
> That shouldn't matter. We don't even use any session on the loadexternal call and the third parameter is the hierarchy (you can look at the command table [2] to see order). It might be the version of ESAPI used. In older versions the wrong type was used there, see commit [1] which is only available on ESAPI versions 3.0+.
> 
>     * What version of tpm2-tss is being used?

tpm2-tss-2.3.2-4.el8.x86_64

> 
>     * if it's an older ESAPI and to be backwards compat **we should replace** `ESYS_TR_RH_OWNER` with  `TPM2_RH_OWNER`.

williamcroberts:
> tpm2-tss-2.3.2-4.el8.x86_64

Yep I am confident that is the problem:
Replace  `ESYS_TR_RH_OWNER` with  `TPM2_RH_OWNER` for LoadExternal command. That will provide backwards compat. Nasty little bug in our code, my apologies.

poettering:
> Replace `ESYS_TR_RH_OWNER` with `TPM2_RH_OWNER` for LoadExternal command. That will provide backwards compat. Nasty little bug in our code, my apologies.

Ah, I remember I actually found this issue once earlier with other code and worked around it just like that. I am a bit confused why this code passed CI though before the rebase. Weird.

Force pushed new version now, with the change applied. Let's see if CI loves it now

@williamcroberts thanks so much for the help!

poettering:
Setting green label again, given the trivial nature of this change and the fact this was approved before. Fingers crossed.

aplanas:
I am sure that I am missing something but I am having troubles to understand the vector attack that we are protecting with signed PCRs.

IIUC the generated policy is not bound to any TPM2's SK, meaning that the same signed policy can be used in any TPM2 to unseal different secrets, as far as the signature is valid (and that is the goal).

But the reverse should be true, any policy that can be validated with a provided public key will unseal the secret.

For UKI we will store the unified kernel, the JSON and the public key in the non-encrypted / non-validated ESP, so one can take the hard disk, replace the three elements and introduce a bogus initrd or kernel cmdline (evil maid attack). With SB enabled we still can protect the kernel and the modules, but eventually cryptsetup can unseal the LUKS2 TPM2 slot and mount rootfs.

What I am missing? Maybe the idea is to extend this policy to some other PCRs (sort of AND)?

aplanas:
> What I am missing? 

Ouch ... I know what I am missing. The secrets are sealed using the (SK, public key) tuple. So my logic is wrong. Only policy that can be validated by the same public part can be unsealed.

Sorry for the noise :) !

man/systemd-measure.xml
daandemeyer: @williamcroberts Any chance you might be willing to work on a patch to make systemd-measure work without a TPM device? Would be very useful for building images

williamcroberts: > @williamcroberts Any chance you might be willing to work on a patch to make systemd-measure work without a TPM device? Would be very useful for building images

Unfortunatley I got yanked off of all this work, so I won't have the cycles for major contributions, only support for previous work or guidance.

daandemeyer: @williamcroberts Actually I'm in luck, since https://github.com/systemd/systemd/pull/27627 was just opened, so the work has already been done!

akostadinov:
Could you explain what would be the difference between `--tpm2-pcrs` and `--tpm2-public-key-pcrs` when same values are used? For example value `12`. If one boots setting the kernel command line to something else, what would be the difference whether one or the other option was used?

Or maybe explain specifically a scenario where there will be a difference and what would be that difference.

In the man page I can only read:
> Note the difference between --tpm2-pcrs= and --tpm2-public-key-pcrs=:
>           the former binds decryption to the current, specific PCR values; the
>           latter binds decryption to any set of PCR values for which a signature
>           by the specified public key can be provided. The latter is hence more
>           useful in scenarios where software updates shell be possible without
>           losing access to all previously encrypted LUKS2 volumes.

I don't understand whether it is something that may help in the future or does help and for what specifically.

PR #24353, opened 2022-08-18
udev: hwdb: Add orientation quirk for base-mounted accelerometers on Chromebooks

alpernebbi:
This is a follow-up to #24276, which added generic hwdb entries for cros-ec-accel devices. Also see the post-merge comments to that. In short, the existing hwdb fragment is correct for the display-mounted accelerometer, but not for the base-mounted one. The only way to distinguish the two devices is a [`location`](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio-cros-ec) sysfs file before Linux v6.0, and a [`label`](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio) file after that version.

The iio device `label` looks like it would be useful for similar cases later on, so the first commit adds a rule to match it in the hwdb entries. The second commit adds entries in this style for cros-ec-accel. The third commit tries to import these rules based on the `location` file for older kernels, though I guess it's not essential and can be dropped if you want to keep driver-specific things out of the rules.

### DIFF
--- a/rules.d/60-sensor.rules
+++ b/rules.d/60-sensor.rules
@@ -7,6 +7,17 @@ SUBSYSTEM=="iio", KERNEL=="iio*", SUBSYSTEMS=="usb|i2c|platform", ATTR{label}!="
   IMPORT{builtin}="hwdb 'sensor:$attr{label}:modalias:$attr{modalias}:$attr{[dmi/id]modalias}'", \
   GOTO="sensor_end"
 
+# Before Linux v6.0, cros-ec-accel used a non-standard 'location' sysfs file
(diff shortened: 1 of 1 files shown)


alpernebbi:
`parse-hwdb` was failing, added currently standardized iio labels to its grammar.

amstan:
You are on a roll! Looks good to me.

`Reviewed-by: Alexandru Stan <amstan@chromium.org>`

@gwendalg, any concerns?

gwendalcr:
`Reviewed-by: Gwendal Grignou <gwendal@chromium.org>`

Thanks,
Gwendal.

yuwata (approved):

Issue #24258, opened 2022-08-09
sd-netlink API issues


poettering:
Here are a bunch of things we probably should figure out before doing a release with sd-netlink public.

1. What is the difference between sd_open_fd() and sd_new_with_fd()? shouldn't they be the same? They look like diverging reimplementations of each other.

2. The API .h file should mark return params with `ret_` in the name

3. sd_netlink_attach_filter()'s filter argument should probably be const, no? Also, isn't it a bit too low-level? hide for now?

4. hmm there's sd_netlink_message_open_container() + sd_netlink_message_close_container() and sd_netlink_message_enter_container() + sd_netlink_message_exit_container(). So far so good. But for arrays there's sd_netlink_message_open_array() + sd_netlink_message_cancel_array()? cancel and no exit? also, in the header files, these function should probably be grouped together.

5. Rename sd_netlink_message_request_dump() → sd_netlink_message_set_request_dump()?

6. Are the types of the various rtnl getter functions fully in order? i.e. i see a `unsigned *family` there, and it smells to me this should be `int*`?

7. sd_genl_add_match() takes a description string. Should it really? in sd-bus descriptions (being a debugging feature) can be set on the slot, but only after creation. i mean, i see the value to pass this in, but maybe do that in an internal helper?

8. similar, should the destroy cb be in the pimary function parameter list? it's overly long, and should be ok to be set in a second call afte rallocation

yuwata:
Partially related to #24124.

keszybz:
>     1. What is the difference between sd_open_fd() and sd_new_with_fd()? shouldn't they be the same? They look like diverging reimplementations of each other.

I couldn't make heads or tails of this. I think it's some artifact of how netlink and generic netlink we added in separate phases. sd_new_with_fd() was unused, so I just removed it.
 
>     2. The API .h file should mark return params with `ret_` in the name

Done.

>     3. sd_netlink_attach_filter()'s filter argument should probably be const, no? Also, isn't it a bit too low-level? hide for now?

Const done.

I'm not sure if it should be hidden. @yuwata?
 
>     4. hmm there's sd_netlink_message_open_container() + sd_netlink_message_close_container() and sd_netlink_message_enter_container() + sd_netlink_message_exit_container(). So far so good. But for arrays there's sd_netlink_message_open_array() + sd_netlink_message_cancel_array()? cancel and no exit? also, in the header files, these function should probably be grouped together.

I reordered functions to be in two groups: message construction and reading.
[...]

PR #24356, opened 2022-08-19
Small improvements to the sd-netlink api

keszybz:
For #24258.

### DIFF
--- a/src/libsystemd/sd-netlink/sd-netlink.c
+++ b/src/libsystemd/sd-netlink/sd-netlink.c
@@ -728,7 +728,8 @@ static int prepare_callback(sd_event_source *s, void *userdata) {
                         return r;
         }
 
-        r = sd_event_source_set_enabled(nl->time_event_source, enabled > 0);
+        r = sd_event_source_set_enabled(nl->time_event_source,
+                                        enabled > 0 ? SD_EVENT_ONESHOT : SD_EVENT_OFF);
         if (r < 0)
                 return r;
 


bluca (approved):

src/libsystemd/sd-netlink/sd-netlink.c
```diff
@@ -705,32 +705,30 @@ static int time_callback(sd_event_source *s, uint64_t usec, void *userdata) {
...
+        if (enabled < 0)
+                return enabled;
+        if (enabled > 0) {
+                r = sd_event_source_set_time(nl->time_event_source, until);
+                if (r < 0)
+                        return r;
         }
 
-        r = sd_event_source_set_enabled(nl->time_event_source, r > 0);
+        r = sd_event_source_set_enabled(nl->time_event_source, enabled > 0);
```
poettering: hmm, actually this would set the time event to SD_EVENT_ON, but I am pretty sure for time events SD_EVENT_ONESHOT is more appropriate

keszybz:
I added one more patch to use SD_EVENT_ONESHOT.

src/libsystemd/sd-netlink/sd-netlink.c
```diff
@@ -947,15 +947,15 @@ _public_ int sd_netlink_add_match(
...
+_public_ int sd_netlink_attach_filter(sd_netlink *nl, size_t len, const struct sock_filter *filter) {
         assert_return(nl, -EINVAL);
         assert_return(len == 0 || filter, -EINVAL);
 
         if (setsockopt(nl->fd, SOL_SOCKET,
                        len == 0 ? SO_DETACH_FILTER : SO_ATTACH_FILTER,
                        &(struct sock_fprog) {
                                .len = len,
-                               .filter = filter,
+                               .filter = (struct sock_filter*) filter,
```
yuwata, line 932: Maybe better to constify `struct sock_fprog`??

yuwata (changes requested):
Not a matter, but I'd like to merge the last commit into one of the previous commit.

src/libsystemd/sd-netlink/sd-netlink.c
keszybz: I tried that, but gcc still emits the error. I think the problem is that `filter` is a non-`const` pointer, so even if the structure *is* `const`, we get the warning.

keszybz:
> Not a matter, but I'd like to merge the last commit into one of the previous commit.

I don't think this is a good idea. It is a change in behaviour, unlike all the other commits.

src/libsystemd/sd-netlink/sd-netlink.c
yuwata: Hm, spurious. But OK.

PR #24367, opened 2022-08-19
measure: add json output

poettering:


bluca:
Can you add a test for the output in TEST-70? Just to ensure we don't somehow break it later on, given it would be an api-like output

poettering:
There's a test for this included in #24351, hence merging already.

src/boot/measure.c
```diff
@@ -47,15 +50,18 @@ static int help(int argc, char *argv[], void *userdata) {
                "  status             Show current PCR values\n"
                "  calculate          Calculate expected PCR values\n"
```
yuwata, line 51: Maybe better to align descriptions for verbs?

poettering: Actually, the #24351 PR adds more stuff to --help which makes this aligned properly again.

PR #24374, opened 2022-08-19
measure: store bank name in PcrState structure

poettering:
Let's not query/strdup()ascii_strlower() it all the time, but just cache
it once and use it.


poettering:
(just some minor refactoring)

yuwata (approved):

bluca (approved):

PR #24385, opened 2022-08-22
hostname: fix fallback chassis type

### LINKED ISSUE #24384: hostname: Chassis type through org.freedesktop.hostname1 but returned an emtpy string.

251.4-53

### Used distribution

Rawhide

### Linux kernel version used

6.0.0-0.rc1.20220819git4c2d0b039c5c.16.fc38.x86_64

x86_64

### Component

systemd-hostnamed

If we requested the chassis type through dbus org.freedesktop.hostname1 in a VM, we expected the chassis type should be "vm" rather than an empty string ''.

If we request the chassis type in qemu and virtual box, it always returns an empty string.
in_vm
Screen Shot 2022-08-22 at 11 40 14 AM

1. Install Fedora in a VM (QEMU or Virtual box)
2. Request chassis type through dbus interface "org.freedesktop.DBus.Properties"
3. the returned value should be an empty string.

_No response_

yuwata:
Fixes a bug introduced by 8c8b1800e90d4307397300ef32b0f6d95efad057 (v251).

Fixes #24384.

### DIFF
--- a/test/units/testsuite-71.sh
+++ b/test/units/testsuite-71.sh
@@ -44,9 +44,55 @@ test_hostname() {
     fi
 }
 
+restore_machine_info() {
+    if [[ -e /tmp/machine-info.bak ]]; then
+        mv /tmp/machine-info.bak /etc/machine-info
+    else
+        rm -f /etc/machine-info
+    fi
+}
+
+get_chassis() (
+    # shellcheck source=/dev/null
+    . /etc/machine-info
+
+    echo "$CHASSIS"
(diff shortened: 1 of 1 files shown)


poettering:
lgtm, but shellcheck/lint/code-scanning doesn't like it.

yuwata:
The CI failure should be fixed now. PTAL.

The first commit needs to be backported to v251-stable.

bluca (approved):

PR #24394, opened 2022-08-22
mkosi: Ensure we build all features/components in mkosi

daandemeyer:
Explicitly enable all features/components in the mkosi build to
ensure they all get built and we get an error if they can't be built.

We also rework the packages sections of all mkosi configs to reduce
duplication and cover all the dependencies necessary to build/use all
systemd features.

Note that for the final image, since systemd is installed by default
in base images, we rely on that to install the base library dependencies
and we only list extra optional dependencies and tools that aren't already
installed by default into the base image.

We also drop the centos stream 8 mkosi build as dependencies on that
distro are too out-of-date to be able to build all systemd features.
Since centos stream 9 has been out for a while, let's focus on that
and leave it to downstream to keep systemd building on centos stream 8.

Finally, there's a few additions to the mkosi scripts to make sure
services don't start by default on boot.

### DIFF
--- a/mkosi.default.d/debian/10-mkosi.debian
+++ b/mkosi.default.d/debian/10-mkosi.debian
@@ -8,17 +8,31 @@ Distribution=debian
 Release=testing
 
 [Content]
+Packages=
+        cryptsetup-bin
+        iproute2
+        isc-dhcp-server
+        libbpf0
+        libfido2-1
+        libglib2.0-0
+        libgnutls30
+        libidn2-0
+        libmicrohttpd12
+        libp11-kit0
+        libpam0g
+        libpwquality1
+        libqrencode4
+        libtss2-dev # Use the -dev package to avoid churn in updating version numbers
+        netcat-openbsd
+        policykit-1
+        procps
+        quota
+        xxd
+
 BuildPackages=
-        acl
-        clang
-        docbook-xml
+        bpftool
         docbook-xsl
-        gcc
         g++
-        gettext
-        git
-        gnu-efi
-        gperf
         libacl1-dev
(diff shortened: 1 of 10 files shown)


mkosi.default.d/debian/10-mkosi.debian
```diff
@@ -8,77 +8,63 @@ Distribution=debian
...
+        libfido2-1
+        libglib2.0-0
+        libgnutls30
+        libidn2-0
+        libmicrohttpd12
+        libp11-kit0
+        libpam0g
+        libpwquality1
+        libqrencode4
+        libtss2-esys-3.0.2-0
```
bluca: maybe use the -dev for all these libraries, so that it will require less maintenance when there are ABI breaks? extra cost should be minimal, some header files and symlinks, and maybe the odd static lib or two

daandemeyer: Do they break ABI that often? There were about 7 months between the last two releases. That don't seem like too much churn

bluca: it's really out of our control, and to my mind the less churn we need to worry about, the better

bluca (approved):

DaanDeMeyer:
opensuse is supposed to fail and will be fixed by my other mkosi PR once this one is merged

bluca:
This looks new?

```
‣   Generating initramfs images…
dracut: WARNING: running in hostonly mode in a container!!
findmnt: can't read (null): No such file or directory
Container mkosi-a03cedfb7d7841db9f1e24bf42b7a02d terminated by signal KILL.
```

PR #24396, opened 2022-08-22
stop using dirname()/dirname_malloc()

poettering:
The corner cases are weird, and we have a better call now in path_extract_directory(). Let's use that everywhere.

### DIFF
--- a/TODO
+++ b/TODO
@@ -77,6 +77,11 @@ Janitorial Clean-ups:
 * rework mount.c and swap.c to follow proper state enumeration/deserialization
   semantics, like we do for device.c now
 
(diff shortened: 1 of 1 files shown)


bluca (approved):

bluca:
```
Aug 22 15:07:36 H testsuite-65.sh[366]: + systemd-analyze verify --recursive-errors=no --root=/tmp/img/ testfile.service
Aug 22 15:07:36 H systemd-analyze[380]: Failed to generate unit load path: Destination address required
```

poettering:
Force pushed new version. Only change is a minor patch to address the CI issue. Since this was approved before, taking liberity to set green label.

PR #24399, opened 2022-08-22
measure: add --current switch for "systemd-measure calculate"

poettering:
This allows allows shortcutting measurements of the specified files and
use the information from /sys/ instead.

This is not too useful on its own given that "systemd-measure status"
already exists which displays the current, relevant PCR values. The main
difference is how "complete" the information is. "status" will detect if
the measurements make any sense, and show more than PCR 11. "calculate
--current" otoh only reads PCR 11 and uses that, and that's really it.

This is mainly preparation for later work to add PCR signing to the
tool, where usually it makes most sense to sign prepared kernel images,
but for testing it's really useful to shortcut signing to the current
PCR values instead

### DIFF
--- a/man/systemd-measure.xml
+++ b/man/systemd-measure.xml
@@ -92,6 +92,14 @@
       </varlistentry>
 
       <varlistentry>
+        <term><option>--current</option></term>
+        <listitem><para>When used with the <command>calculate</command> verb, takes the PCR 11 values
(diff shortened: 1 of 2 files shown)


poettering:
Yet another split-out from #24351 that makes some sense on its own.

bluca (approved):

PR #24402, opened 2022-08-22
bootctl: Allow uneven length BootXXXX variables

kreijack:
The BootXXXX variables may have an uneven length. Don't return error
in this case.

On my PC, some efi boot variables have an uneven length.
```
$ efivar -p -n 8be4df61-93ca-11d2-aa0d-00e098032b8c-Boot0000
GUID: 8be4df61-93ca-11d2-aa0d-00e098032b8c
Name: "Boot0000"
Attributes:
	Non-Volatile
	Boot Service Access
	Runtime Service Access
Value:
00000000  01 00 00 00 56 00 55 00  45 00 46 00 49 00 53 00  |....V.U.E.F.I.S.|
00000010  48 00 45 00 4c 00 4c 00  00 00 04 01 2a 00 01 00  |H.E.L.L.....*...|
00000020  00 00 00 08 00 00 00 00  00 00 00 00 20 00 00 00  |............ ...|
00000030  00 00 41 17 12 7e 7b aa  41 d0 ba ec 0f e2 33 7e  |..A..~{.A.....3~|
00000040  2c 50 02 02 04 04 28 00  5c 00 45 00 46 00 49 00  |,P....(.\.E.F.I.|
00000050  5c 00 53 00 48 00 45 00  4c 00 4c 00 58 00 36 00  |\.S.H.E.L.L.X.6.|
00000060  34 00 2e 00 45 00 46 00  49 00 00 00 7f ff 04 00  |4...E.F.I.......|
00000070  32                                                |2               |
```
unfortunately utf16_size() return an error in these cases which is printed in console.
```
$ ./build/bootctl
Failed to read boot option 0: Invalid argument
Failed to read boot option 5: Invalid argument
Failed to read boot option 1: Invalid argument
Failed to read boot option 6: Invalid argument
Failed to read boot option 7: Invalid argument
System:
      Firmware: UEFI 2.70 (American Megatrends 5.17)
 Firmware Arch: x64
   Secure Boot: disabled (setup)
  TPM2 Support: yes
  Boot into FW: supported

[...]
```
`efibootmgr` doesn't complaint so I assume that having an uneven length is fine.

It enough to remove the check inside utf16_size(), because even in case of an uneven
length the loop will not de-reference the last missing byte; this because the division
truncates the results to the last valid pair of bytes (ignoring the last byte).

### DIFF
--- a/src/shared/efi-api.c
+++ b/src/shared/efi-api.c
@@ -180,9 +180,6 @@ static ssize_t utf16_size(const uint16_t *s, size_t buf_len_bytes) {
 
         /* Returns the size of the string in bytes without the terminating two zero bytes */
 
-        if (buf_len_bytes % sizeof(uint16_t) != 0)
-                return -EINVAL;
-
         while (l < buf_len_bytes / sizeof(uint16_t)) {
                 if (s[l] == 0)
                         return (l + 1) * sizeof(uint16_t);


keszybz:
Looks reasonabl.e

poettering:
At the very least there should be a comment in place there explaining why we do not check for evenness. And that we ignore the extraneous byte in that case.

But I also think there should be a `log_debug()` message here, mentioning this case and that we ignore it.

PR #24403, opened 2022-08-22
test: install /etc/default/knot if available as well

mrc0mmand:
The knot.service on Ubuntu Jammy loads an env file which we didn't
install, causing the service to fail:

```
knot.service: Will spawn child (service_enter_start_pre): /usr/sbin/knotc
knot.service: Failed to load environment files: No such file or directory
knot.service: Failed to run 'start-pre' task: No such file or directory
knot.service: Failed with result 'resources'.
knot.service: Service will not restart (restart setting)
```

---

@surajkrishnan14 - could you please test if this fixes the issue you found?

### DIFF
--- a/test/TEST-75-RESOLVED/test.sh
+++ b/test/TEST-75-RESOLVED/test.sh
@@ -27,6 +27,7 @@ test_append_files() {
     image_install /lib/tmpfiles.d/knot.conf
     image_install "${ROOTLIBDIR:?}/system/knot.service"
     image_install -o /etc/dbus-1/system.d/cz.nic.knotd.conf
+    image_install -o /etc/default/knot
 
     # Copy over our configuration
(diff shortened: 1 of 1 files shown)


bluca (approved):

surajkrishnan14:
> The knot.service on Ubuntu Jammy loads an env file which we didn't install, causing the service to fail:
> 
> ```
> knot.service: Will spawn child (service_enter_start_pre): /usr/sbin/knotc
> knot.service: Failed to load environment files: No such file or directory
> knot.service: Failed to run 'start-pre' task: No such file or directory
> knot.service: Failed with result 'resources'.
> knot.service: Service will not restart (restart setting)
> ```
> 
> @surajkrishnan14 - could you please test if this fixes the issue you found?

Thanks @mrc0mmand ! It does appear to resolve the error I observed at first and addresses the error described in this PR.

I still see the test failing further along though. It may be unrelated, I've sent you the logs from the latest run with this fix

PR #24404, opened 2022-08-22
job: Don't discard propagated restart jobs when unit is activating

thatguystone:
When a service unit Requires= a socket, and the socket is restarted while the service is in state=activating, the propagated restart is being discarded. This is contrary to the documentation for Requires=, which states "this unit will be stopped (or restarted) if one of the other units is explicitly stopped (or restarted)".

### DIFF
--- a/src/core/job.c
+++ b/src/core/job.c
@@ -400,8 +400,22 @@ bool job_type_is_redundant(JobType a, UnitActiveState b) {
                         b == UNIT_RELOADING;
 
         case JOB_RESTART:
-                return
-                        b == UNIT_ACTIVATING;
+                /* Restart jobs must always be kept.
+                 *
+                 * For ACTIVE/RELOADING units, this is obvious.
+                 *
+                 * For ACTIVATING units, it's more subtle:
+                 *
+                 * Generally, if a service Requires= another unit, restarts of
+                 * the unit must be propagated to the service. If the service is
+                 * ACTIVATING, it must still be restarted since it might have
+                 * stale information regarding the other unit.
+                 *
+                 * For example, consider a service that Requires= a socket: if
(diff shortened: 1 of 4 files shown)


src/core/job.c
```diff
@@ -400,8 +400,7 @@ bool job_type_is_redundant(JobType a, UnitActiveState b) {
                         b == UNIT_RELOADING;
 
         case JOB_RESTART:
-                return
-                        b == UNIT_ACTIVATING;
+                return false;
```
poettering, line 418: Hmm, this should come with a comment explaining why this isn't redundant.

src/core/job.c
```diff
@@ -418,7 +417,7 @@ JobType job_type_collapse(JobType t, Unit *u) {
 
         case JOB_TRY_RESTART:
                 s = unit_active_state(u);
-                if (!UNIT_IS_ACTIVE_OR_RELOADING(s))
+                if (!UNIT_IS_ACTIVE_OR_ACTIVATING(s))
```
poettering, line 439: so, this will change that from now on JOB_TRY_RESTART will be collapsed into JOB_RESTART when UNIT_ACTIVATING state is in effect, as opposed to being suppressed to JOB_NOP.

I think a comment would be good here too, that clarifies that this is about not losing restart jobs enqueued while we are activating.

poettering:
So i think this OK, but this could use some explanatory comments in the code.

Also, a test case would be good (maybe extend an existing one in `test/TEST-??-*/` that is reasonably related)

thatguystone:
The failing tests don't appear to be related my changes:

jammy: mkosi failed with "ISC DHCP" errors.
tumbleweed: mkosi didn't install g++ for some reason
focal-amd64: failed because it's missing the `selinuxenabled` bin

Issue #24390, opened 2022-08-22
TEST-50-DISSECT: The Return of the Fail


mrc0mmand:
### systemd version the issue has been seen with

latest main

### Used distribution

Arch Linux

### Linux kernel version used

5.19.2-arch1-1

### Component

tests

### Unexpected behaviour you saw

A new (I guess) fail sprouted up in the recent CI runs on Arch Linux:

```
[   33.309930] testsuite-50.sh[522]: + systemd-dissect --json=short --root-hash ec88eab712bc6d647d1a5e7fc31ea5a2ff343578d55d2104f8a8247d63b303e0 /tmp/tmp.jekk57/minimal_0.gpt
[   33.314491] testsuite-50.sh[522]: Opened '/tmp/tmp.jekk57/minimal_0.gpt' in O_RDONLY access mode, with O_DIRECT disabled (O_DIRECT was requested but not supported).
[   33.314491] testsuite-50.sh[522]: Failed to enable O_DIRECT for backing file descriptor for loopback device. Continuing without.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/power/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/holders/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu11/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu9/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu7/uevent" does not exist.
[   33.315726] testsuite-50.sh[522]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu5/uevent" does not exist.
[...]

PR #24413, opened 2022-08-23
test: wait for the lodev to get properly initialized

mrc0mmand:
Otherwise we might start writing to one of its partition before the
respective node is created under /dev, resulting in... interesting
stuff.

Resolves: #24390

### DIFF
--- a/test/units/testsuite-50.sh
+++ b/test/units/testsuite-50.sh
@@ -187,6 +187,7 @@ if [ "${HAVE_OPENSSL}" -eq 1 ]; then
     sfdisk --part-label "${image}.gpt" 3 "Signature Partition"
 fi
 loop="$(losetup --show -P -f "${image}.gpt")"
+udevadm wait --timeout 60 --settle "${loop:?}"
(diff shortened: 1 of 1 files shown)


bluca (approved):

test/units/testsuite-50.sh
```diff
@@ -187,6 +187,7 @@ if [ "${HAVE_OPENSSL}" -eq 1 ]; then
     sfdisk --part-label "${image}.gpt" 3 "Signature Partition"
 fi
 loop="$(losetup --show -P -f "${image}.gpt")"
+udevadm wait --timeout 60 --settle "${loop:?}"
```
poettering, line 190: shouldn't this be `udevadm wait --timeout 60 --settle "${loop}p1" "${loop}p2"`

i.e. wait for the two partitions we actually intend to use next

mrc0mmand: According to @yuwata's findings that causes issues - see https://github.com/systemd/systemd/commit/1af39ef2f6ea7f92ea2cc35b3a5d0e720db6ea4a

PR #24417, opened 2022-08-23
core: allow disabling the check that verify

fbuihuu:
There might be (embedded) systems that get never updated (things like
e.g. entertainment systems of trains, for example) and where the adjustment of
the system clock (introduced by b10abe4bba61aebe4c667c412741193f11886298) would
do the wrong thing even if the difference between the systemd build time and
the rtc is 15 years or more.

Unfortunately the commit message of b10abe4bba61aebe4c667c412741193f11886298
explains neither which cases it attempted to address nor where the 15 years
limit comes from however it effectively put a limit on the duration before
systemd must be updated.

This patch allows disabling the adjustment by setting
'clock-valid-range-usec-max' meson option to 0 or to a negative value.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -223,7 +223,7 @@ option('status-unit-format-default', type : 'combo',
 option('time-epoch', type : 'integer', value : 0,
        description : 'time epoch for time clients')
 option('clock-valid-range-usec-max', type : 'integer', value : 473364000000000, # 15 years
-       description : 'maximum value in microseconds for the difference between RTC and epoch, exceeding which is considered an RTC error')
+       description : 'maximum value in microseconds for the difference between RTC and epoch, exceeding which is considered an RTC error ["0" disables]')
 option('default-user-shell', type : 'string', value : '/bin/bash',
        description : 'default interactive shell')
 
--- a/src/shared/clock-util.c
+++ b/src/shared/clock-util.c
@@ -153,7 +153,7 @@ int clock_apply_epoch(ClockChangeDirection *ret_attempted_change) {
         now_usec = now(CLOCK_REALTIME);
         if (now_usec < epoch_usec)
                 *ret_attempted_change = CLOCK_CHANGE_FORWARD;
-        else if (now_usec > usec_add(epoch_usec, CLOCK_VALID_RANGE_USEC_MAX))
+        else if (CLOCK_VALID_RANGE_USEC_MAX > 0 && now_usec > usec_add(epoch_usec, CLOCK_VALID_RANGE_USEC_MAX))
                 *ret_attempted_change = CLOCK_CHANGE_BACKWARD;
         else {
                 *ret_attempted_change = CLOCK_CHANGE_NOOP;


poettering:
something happened to the commit msg title, got truncated?

Can you update the description of the option() in meson_options.txt file, making clear that 0 now turns off the feature?

poettering:
> Unfortunately the commit message of [b10abe4](https://github.com/systemd/systemd/commit/b10abe4bba61aebe4c667c412741193f11886298)
> explains neither which cases it attempted to address nor where the 15 years
> limit comes from however it effectively put a limit on the duration before
> systemd must be updated.

The PR carries a rationale. It's simply that various RTCs where the battery fails might come up with time far in future rather than far in past...

poettering:
Note that strictly speaking you don't need this change, given you can as well just set the value to UINT64_MAX or so.

fbuihuu:
> The PR carries a rationale. It's simply that various RTCs where the battery fails might come up with time far in future rather than far in past...

Indeed but AFAICS the PR actually doesn't mention battery failures and if, in this case, the time read from the RTC is just a random value equally distributed in the future or in the past that can be different on each read. If so randomly fixing the system time at each reboot with systemd build epoch (potentially far in the past) doesn't seem really useful.

Also why it would be better to correct the system clock with a value potentially far in past instead of letting it even if it's far in future.

Wouldn't using a file as another "epoch"  that would be touched on each shutdown be slightly better ? At least from the system POV, the clock wouldn't be re-adjusted in the past on each reboot.

fbuihuu:
> Note that strictly speaking you don't need this change, given you can as well just set the value to UINT64_MAX or so.

Indeed but at least "0" is clearly a value that means "disables" unlike "UINT64_MAX" usec.

fbuihuu:
> The PR carries a rationale.

BTW it would be better if rationales are put in the commit messages rather than leaving them in the PRs IMHO.

poettering:
> > 
> 
> Indeed but AFAICS the PR actually doesn't mention battery failures and if, in this case, the time read from the RTC is just a random value equally distributed in the future or in the past that can be different on each read. If so randomly fixing the system time at each reboot with systemd build epoch (potentially far in the past) doesn't seem really useful.

Well, if the memory is indeed filled with random values then you are fucked, but it might be filled with 0xFF for example, in which case this logic here should work.

> Also why it would be better to correct the system clock with a value potentially far in past instead of letting it even if it's far in future.

Because time runs forward, not backward. If you are extremely far in the future and let time run even further, then you risk overflows. Thus it makes sense to get to values that are *known* to made have sense.

> Wouldn't using a file as another "epoch" that would be touched on each shutdown be slightly better ? At least from the system POV, the clock wouldn't be re-adjusted in the past on each reboot.

We implement that in systemd-timesyncd.

poettering:
> Indeed but at least "0" is clearly a value that means "disables" unlike "UINT64_MAX" usec.

Conceptually if you allow configuration of a range where times are valid, then it certainly makes sense to to set the range to a large value for a maximum of valid range rather htan a small value.

fbuihuu:
> Conceptually if you allow configuration of a range where times are valid, then it certainly makes sense to to set the range to a large value for a maximum of valid range rather htan a small value.

Agreed but in the case of passing the value to a meson option, which is set or read by humans, "0" is simpler to parse. Maybe using "infinity" instead and let meson converting this magic string into `USEC_INFINITY` is better ?

Issue #21592, opened 2021-12-01
systemd-growfs requires that udev ran first to work (because /dev/block/ symlinks must exist)


tewarid:
**systemd version the issue has been seen with**

v249.7

**Used distribution**
**Linux kernel version used** (`uname -a`)
**CPU architecture issue was seen on**

```text
sh-5.1# uname -a
Linux raspberrypi0-wifi 5.10.81 #1 Mon Nov 22 14:00:03 UTC 2021 armv6l
```

**Expected behaviour you didn't see**

root fs does not get resized.

The following works from shell

```
sh-5.1# /lib/systemd/systemd-growfs /
Successfully resized "/" to 348.3M bytes.
```

However, root fs does not get resized to fill all space available on sd card (e.g. 7.5GB of 7.9GB remains available).

**Unexpected behaviour you saw**

```text
sh-5.1# systemctl status -l systemd-growfs@-
x systemd-growfs@-.service - Grow File System on /
     Loaded: loaded (/run/systemd/generator/systemd-growfs@-.service; generated)
     Active: failed (Result: exit-code) since Fri 2021-11-19 17:19:34 UTC; 31s ago
       Docs: man:systemd-growfs@.service(8)
    Process: 111 ExecStart=/lib/systemd/systemd-growfs / (code=exited, status=1/FAILURE)
   Main PID: 111 (code=exited, status=1/FAILURE)

Nov 19 17:19:34 raspberrypi0-wifi systemd[1]: Starting Grow File System on /...
Nov 19 17:19:34 raspberrypi0-wifi systemd-growfs[111]: Failed to open "/dev/block/179:2": No such file or directory
Nov 19 17:19:34 raspberrypi0-wifi systemd[1]: systemd-growfs@-.service: Main process exited, code=exited, status=1/FAILURE
Nov 19 17:19:34 raspberrypi0-wifi systemd[1]: systemd-growfs@-.service: Failed with result 'exit-code'.
Nov 19 17:19:34 raspberrypi0-wifi systemd[1]: Failed to start Grow File System on /.
```

**Steps to reproduce the problem**

/etc/fstab is setup to grow root fs

```
/dev/root            /                    auto       defaults,x-systemd.growfs              1  1
```

poettering:
Note that systemd's growfs logic will only grow the file system itself, not the partition it is contained in. Use systemd-repart for that (or some other tool).

[...]

PR #24418, opened 2022-08-23
systemd-growfs: remove dependency on udev symlinks

tewarid:
systemd-growfs currently requires that udev ran first to work (because
/dev/block/ symlinks must exist) but that is not what happens when
we're not using initrd and systemd is PID1.

See #21592.

### DIFF
--- a/src/basic/devnum-util.c
+++ b/src/basic/devnum-util.c
@@ -5,6 +5,7 @@
 
 #include "chase-symlinks.h"
 #include "devnum-util.h"
+#include "fs-util.h"
 #include "parse-util.h"
 #include "path-util.h"
 #include "string-util.h"
@@ -53,6 +54,53 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
 
         if (asprintf(ret, "/dev/%s/" DEVNUM_FORMAT_STR, t, DEVNUM_FORMAT_VAL(devnum)) < 0)
                 return -ENOMEM;
+
+        return 0;
+}
+
+int device_path_make_major_minor_sysfs(mode_t mode, dev_t devnum, char **ret) {
+        _cleanup_free_ char *syspath = NULL, *link = NULL, *fname = NULL;
+        _cleanup_free_ char *devpath = NULL;
+        const char *t;
+        int r;
+
+        /* Generates the /dev/... path given a dev_t. What makes this different
+         * from device_path_make_major_minor is that it works even when udev 
+         * hasn't yet run */
+
+        if (S_ISCHR(mode))
+                t = "char";
+        else if (S_ISBLK(mode))
+                t = "block";
+        else
+                return -ENODEV;
+
+        if (asprintf(&syspath, "/sys/dev/%s/" DEVNUM_FORMAT_STR, t, DEVNUM_FORMAT_VAL(devnum)) < 0)
+                return -ENOMEM;
+
+        r = readlink_malloc(syspath, &link);
+        if (r < 0)
+                return r;
+
+        r = path_extract_filename(link, &fname);
+        if (r < 0)
+                return r;
+
+        devpath = path_join("/dev", fname);
+        if (!devpath)
+                return -ENOMEM;
+
+        struct stat st;
+        if (stat(devpath, &st) < 0)
+                return -errno;
+
+        if (((st.st_mode ^ mode) & S_IFMT) != 0)
+                return S_ISBLK(mode) ? -ENOTBLK : -ENODEV;
+
+        if (st.st_rdev != devnum)
+                return -ENXIO;
+
+        *ret = TAKE_PTR(devpath);
 
         return 0;
 }
--- a/src/basic/devnum-util.h
+++ b/src/basic/devnum-util.h
@@ -29,6 +29,7 @@ int parse_devnum(const char *s, dev_t *ret);
         })
 
 int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret);
+int device_path_make_major_minor_sysfs(mode_t mode, dev_t devnum, char **ret);
 int device_path_make_canonical(mode_t mode, dev_t devnum, char **ret);
 int device_path_parse_major_minor(const char *path, mode_t *ret_mode, dev_t *ret_devnum);
 
--- a/src/partition/growfs.c
+++ b/src/partition/growfs.c
@@ -41,7 +41,7 @@ static int resize_crypt_luks_device(dev_t devno, const char *fstype, dev_t main_
         if (r < 0)
                 return log_error_errno(r, "Cannot resize LUKS device: %m");
 
-        r = device_path_make_major_minor(S_IFBLK, main_devno, &main_devpath);
+        r = device_path_make_major_minor_sysfs(S_IFBLK, main_devno, &main_devpath);
         if (r < 0)
                 return log_error_errno(r, "Failed to format device major/minor path: %m");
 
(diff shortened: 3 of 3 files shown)


tewarid:
I'll be happy to rename `device_path_make_major_minor2` to something more interesting :)

src/basic/devnum-util.c
```diff
@@ -57,6 +58,42 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
         return 0;
 }
 
+int device_path_make_major_minor2(mode_t mode, dev_t devnum, char **ret) {
+        const char *t;
+        _cleanup_free_ char *syspath = NULL;
+        _cleanup_free_ char *link = NULL;
+        _cleanup_free_ char *fname = NULL;
```
poettering: please merge these three lines

src/basic/devnum-util.c
```diff
@@ -57,6 +58,42 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+                t = "char";
+        else if (S_ISBLK(mode))
+                t = "block";
+        else
+                return -ENODEV;
+
+        if (asprintf(&syspath, "/sys/dev/%s/" DEVNUM_FORMAT_STR, t, DEVNUM_FORMAT_VAL(devnum)) < 0)
+                return -ENOMEM;
+
+        int r = readlink_malloc(syspath, &link);
```
poettering: `r` is special for us, it's the *primary* variable we write function return error codes too, hence it should always be declared up top at the beginning of the function.

src/basic/devnum-util.c
```diff
@@ -57,6 +58,42 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+        if (r < 0)
+                return r;
+
+        r = path_extract_filename(link, &fname);
+        if (r < 0)
+                return r;
+
+        if (asprintf(ret, "/dev/%s", fname) < 0)
```
poettering: we concatenating strings use strjoin(), not asprintf(). In fact, in this case even use path_join(), since you are joining paths here.

poettering: also, as per coding style: our functions should initialize all return parameters on success, but none on failure. Here you are initializing ret already, even though the stat() below can still fail. If it does then you will have initialized ret even though the function overall returned an error. hence, please write the path to a local variable, and move it into `*ret` only after stat() succeeded.

src/basic/devnum-util.c
tewarid: Considered using a single variable. Does `_cleanup_free_` auto free when a new value is assigned?

src/basic/devnum-util.c
```diff
@@ -57,6 +58,42 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+                return -ENOMEM;
+
+        struct stat st;
+        if (stat(*ret, &st) < 0)
+                return -errno;
+
```
poettering, line 96: the reason why i suggested to have the stat() here is to verify the assumption that the the device node is named like the basename of the sysfs path of the device. i.e. after you stat() here, please check:

```c
if (((mode ^ st.st_mode) & S_IFMT) != 0)
        return S_ISBLK(mode) ? -ENOTBLK : -ENODEV;
if (st.st_rdev != devnum)
        return -ENXIO;
```

this will check that device node type (i.e. char vs. block) and major/minor actually match our expectations.

src/basic/devnum-util.c
```diff
@@ -57,6 +58,42 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
         return 0;
 }
 
+int device_path_make_major_minor2(mode_t mode, dev_t devnum, char **ret) {
```
poettering: "2"? that's chickening out while looking for a function name...

maybe `device_path_by_major_minor_sysfs()`? or so?

And maybe add a comment at the top, why this function exists, i.e. that it is supposed to work even without udev around, which device_path_make_major_minor() doesn't deliver.

poettering:
looks pretty good, but some comments, see above

src/basic/devnum-util.c
bluca: using 3 variables is fine, but declare them on the same line

tewarid: ah :)

src/basic/devnum-util.c
tewarid: :)

src/basic/devnum-util.c
```diff
@@ -57,6 +58,59 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+        if (r < 0)
+                return r;
+
+        devpath = path_join("/dev", fname);
+        if (!devpath)
+                return -ENOMEM;
+
+        struct stat st;
+        if (stat(devpath, &st) < 0) {
+                free(devpath);
```
poettering: please define devpath also with `_cleanup_free_`, so that you can drop the manual `free()` here.

src/basic/devnum-util.c
```diff
@@ -57,6 +58,59 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+                return -errno;
+        }
+
+        if ((st.st_mode & S_IFMT) != mode) {
+                free(devpath);
+                return -ENODEV;
```
poettering: i think it would make sense to return ENOTBLK if we expected a block device, but it is something else.

hence:

```c
return S_ISBLK(mode) ? -ENOTBLK : -ENODEV;
```

src/basic/devnum-util.c
```diff
@@ -57,6 +58,59 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+        if (!devpath)
+                return -ENOMEM;
+
+        struct stat st;
+        if (stat(devpath, &st) < 0) {
+                free(devpath);
+                return -errno;
+        }
+
+        if ((st.st_mode & S_IFMT) != mode) {
```
poettering: I think we should ignore the rest of the mode_t bits here (i.e. access mode and stuff). Thus, make the check like this:

```c
if (((st.st_mode ^ mode) & S_IFMT) != 0) {
```

src/basic/devnum-util.c
```diff
@@ -57,6 +58,59 @@ int device_path_make_major_minor(mode_t mode, dev_t devnum, char **ret) {
...
+                free(devpath);
+                return -ENODEV;
+        }
+
+        if (st.st_rdev != devnum) {
+                free(devpath);
+                return -ENXIO;
+        }
+
+        *ret = devpath;
```
poettering: once you changed `devpath` to use `_cleanup_free_` change this line to:

```c
*ret = TAKE_PTR(devpath);
```

`TAKE_PTR()` is a nice macro which would be equivalentt o:

```c
*ret = devpath;
devpath = NULL;
```

By setting `devpath` to NULL here, you disable the effect of `_cleanup_free_` at this point.

poettering (changes requested):

poettering:
lgtm

yuwata:
Sorry for late, but `sd_device_new_from_devnum()` and `sd_device_open()` should be used.

poettering:
oh yeah, at least the first and the last use of the new function would indeed be better done with sd_device_new_from_devnum() + sd_device_open() call

Any chance you can post a follow-up patch @tewarid?

tewarid:
Yep - I'll look into it.

tewarid:
> Sorry for late, but `sd_device_new_from_devnum()` and `sd_device_open()` should be used.

You mean, replace device_path_parse_major_minor_sysfs + open by sd_device_new_from_devnum + sd_device_open?

yuwata:
> > Sorry for late, but `sd_device_new_from_devnum()` and `sd_device_open()` should be used.
> 
> You mean, replace device_path_make_major_minor_sysfs + open by sd_device_new_from_devnum + sd_device_open?

Right.

Also, for other places, I'd like to replace `device_path_make_major_minor_sysfs()` with `sd_device_new_from_devnum()` and `sd_device_get_devname()`. It should be slightly slower, but safer I think.

PR #24420, opened 2022-08-23
Rework on_ac_power()

keszybz:


keszybz:
mkosi/ci (ubuntu,jammy): ● isc-dhcp-server.service  loaded failed failed ISC DHCP IPv4 server
mkosi/ci (opensuse): /root/mkosi.build: line 159: groupadd: command not found
CentOS CI (CentOS Stream 8):
23:47:32 TEST-64-UDEV-STORAGE
23:45:41 [   63.177025] testsuite-64.sh[774]: + udevadm wait --settle --timeout=30 /dev/sda
23:45:41 [   93.370584] testsuite-64.sh[3700]: Timed out for waiting devices being initialized.

So those don't look to be related.

src/shared/udev-util.c
```diff
@@ -704,6 +704,14 @@ int on_ac_power(void) {
 
                 bool is_battery = streq(val, "Battery");
                 if (is_battery) {
+                        r = sd_device_get_sysattr_value(d, "scope", &val);
+                        if (r < 0)
+                                log_device_debug_errno(d, r, "Failed to read 'scope' sysfs attribute, ignoring: %m");
+                        else if (streq(val, "Device")) {
```
poettering, line 710: i wonder if the inverted check would be safer, i.e. ignore all devices except those where we see "Unknown", "System" or the field is missing. Who knows what other kind of power supplies people will add later...

keszybz: This change is "minimal" in the sense that we only filter out devices that we explicitly know are "Device". I think it's OK: I expect that if this field grows further values, it'll be something like "Hybrid" or "Multi-function", which potentially includes also powering the system.

keszybz:
The CI failures are unrelated, the same as in other PRs… So we could merge this, unless there's further discussion.

src/shared/udev-util.c
poettering: my thinking that maybe "port-replicator" or so might be exposed, i.e. a set of devices instead of a single one.

i guess the summary is "we don't know the unknown", hence everything is wrong.

keszybz: The kernel docs say that "Device" means "a device or a group of devices"… Dunno. We should get some kernel folks to look at this.

src/libsystemd/sd-device/sd-device.c
```diff
@@ -2206,6 +2206,25 @@ _public_ int sd_device_get_sysattr_value(sd_device *device, const char *sysattr,
...
+                return r;
+
+        unsigned v;
+        r = safe_atou(value, &v);
+        if (r < 0)
+                return log_device_debug_errno(device, r, "Failed to parse '%s' attribute: %m", sysattr);
+
+        if (ret_value)
+                *ret_value = v;
+        /* We return "true" if the value is positive. */
```
dnicolodi, line 2224: It is unsigned, how can it be negative?

poettering: can be zero. "positive" in the sense of "> 0"

PR #24423, opened 2022-08-24
tree-wide: change --kill-who to --kill-whom

keszybz:
popt allows non-ambiguous abbreviations, so backwards-compat is maintained, and
people can use --kill-who (or even shorter abbreviations). English is flexible,
so in common speach people would use both forms, even if "whom" is technically
more correct. The advantage of using the longer form in the code is that we
effectively allow both forms, so we stop punishing people who DTGCT¹, but still
allow people to use the spoken form if they prefer.

1. Do the gramatically correct thing


poettering:
popt? getopt!

poettering:
btw, I hate this feature of getopt to allow non-ambiguous abbreviations. It's an API incompatibility time bomb.

But given we can't turn it off anyway, PR looks OK.

test/units/testsuite-57-short-lived.sh
```diff
@@ -12,7 +12,7 @@ fi
 echo "$counter" > /tmp/testsuite-57.counter
 
 if [ "$counter" -eq 5 ] ; then
-    systemctl kill --kill-who=main -sUSR1 testsuite-57.service
+    systemctl kill --kill-whom=main -sUSR1 testsuite-57.service
```
bluca, line 15: maybe leave one behind, so that we really ensure we don't break backward compat?

keszybz: Done. I also fixed the getopt/popt mixup.

keszybz:
> btw, I hate this feature of getopt to allow non-ambiguous abbreviations. It's an API incompatibility time bomb.

getopt has the advantage that it provides localized error messages. So I think we want to keep using it, but we might provide a wrapper that has a better interface and that also doesn't allow abbreviations. But that's obviously subject for another time.

bluca (approved):

medhefgo:
`kill-whom` sounds so yuck, even if it's technically correct. How about `kill-target` or `kill-victim` instead?

PR #24425, opened 2022-08-24
shutdown: lazily umount all API vfs

poettering:
Alternative for df48b430a4a85f923eaecb3fadf9c514692d2082.

@medhefgo does this work for you?

### DIFF
--- a/src/shutdown/umount.c
+++ b/src/shutdown/umount.c
@@ -529,11 +529,12 @@ static bool nonunmountable_path(const char *path) {
 }
 
 static void log_umount_blockers(const char *mnt) {
+        _cleanup_free_ char *blockers = NULL;
+        int r;
+
         _cleanup_closedir_ DIR *dir = opendir("/proc");
         if (!dir)
-                return (void) log_warning_errno(errno, "opendir(/proc) failed: %m");
-
-        _cleanup_free_ char *blockers = NULL;
+                return (void) log_warning_errno(errno, "Failed to open /proc/: %m");
 
         FOREACH_DIRENT_ALL(de, dir, break) {
                 if (!IN_SET(de->d_type, DT_DIR, DT_UNKNOWN))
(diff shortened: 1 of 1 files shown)


src/shutdown/umount.c
```diff
@@ -158,6 +158,10 @@ int mount_points_list_get(const char *mountinfo, MountPoint **head) {
                         .remount_options = TAKE_PTR(remount_options),
                         .remount_flags = remount_flags,
                         .try_remount_ro = try_remount_ro,
+
+                        /* Unmount sysfs/procfs/… lazily, since syncing doesn't matter there, and it's OK if
+                         * something keeps an fd open to it. */
+                        .umount_lazily = fstype_is_api_vfs(fstype),
```
medhefgo: You're calling this twice now in this function now. This should be cached.

src/shutdown/umount.c
```diff
@@ -632,23 +636,14 @@ static int umount_with_timeout(MountPoint *m, bool last_try) {
                  * -EIO rather than blocking indefinitely. If the filesysten is "busy", this may allow
                  * processes to die, thus making the filesystem less busy so the unmount might succeed
                  * (rather than return EBUSY). */
-                r = RET_NERRNO(umount2(m->path, MNT_FORCE));
+                r = RET_NERRNO(umount2(m->path,
+                                       UMOUNT_NOFOLLOW | /* Don't follow symlinks: this should never happen unless our mount list was wrong */
+                                       (m->umount_lazily ? MNT_DETACH : MNT_FORCE)));
```
medhefgo, line 659: Maybe make a `umount_flags` field instead, like `remount_flags`?

medhefgo (commented):
I was about to take a shot at this but you were faster. :D

I'll give this a try later.

src/shutdown/umount.c
poettering: I thought about this, but I think we probably should start mangling the flags in the loop here sooner or later, i.e. first try to unmount without MNT_FORCE, and then add this in in a later loop iteration. Thus pre-calculating the flags early doesn't really make too much sense, if you have to mangle them later anyway.

But I didn't want to add the conditional MNT_FORCE stuff into the current PR already, and do that some other time, as it is unrelated.

src/shutdown/umount.c
```diff
@@ -151,18 +151,20 @@ int mount_points_list_get(const char *mountinfo, MountPoint **head) {
                         remount_flags = (remount_flags|MS_REMOUNT|MS_RDONLY) & ~MS_BIND;
                 }
 
-                m = new0(MountPoint, 1);
+                m = new(MountPoint, 1);
                 if (!m)
                         return log_oom();
 
+                *m = (MountPoint) {
+                        .remount_options = TAKE_PTR(remount_options),
```
medhefgo: Not like that'll ever matter, but this gets leaked if the strdup futher down fails, no?

poettering: nope, the `m` variable is marked with `_cleanup_`

medhefgo: Yes, `m` is freed, but `m->remount_options` would leak.

medhefgo:
Seems like this works too.

src/shutdown/umount.c
poettering: oh, right. it's a simple `_cleanup_free_` indeed

PR #24427, opened 2022-08-24
check-os-release.py: properly close files

djasa:
Code now loops on lines of open file without making sure that file descriptors are closed eventually. This may be problem if the code is used as-is in long-lived programs. Let's use Python's "with open() as ...:" idiom to make sure that the FDs get properly closed. This change also allows more straightforward adaptation of parsing function to another source of data which may be useful in unit tests checking that consumers of read_os_release() work properly with various os-release files that can be found in the wild.

### DIFF
--- a/man/check-os-release.py
+++ b/man/check-os-release.py
@@ -8,12 +8,14 @@ import sys
 def read_os_release():
     try:
         filename = '/etc/os-release'
-        f = open(filename)
+        with open(filename, 'r') as f:
+            lines = f.readlines()
     except FileNotFoundError:
         filename = '/usr/lib/os-release'
-        f = open(filename)
+        with open(filename, 'r') as f:
+            lines = f.readlines()
 
-    for line_number, line in enumerate(f, start=1):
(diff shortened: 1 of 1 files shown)


keszybz:
No, this is a misunderstanding / cargo cult. It is not required for correctness of this program for the file to be closed at any specific point. In this way, the case of *reading* files is fundamentally different from *writing* of files. All we care about, is that the file object and the file descriptor at cleaned up at some point. What actually happens, at least with the default `cpython` implementation, is that file descriptor is closed when the file object is deallocated, which happens when the function returns. (`pypy` and other implementations might provide slightly different behaviour, but not in ways that matter here.)  So there is absolutely no problem in "long-lived programs". But even if we pretended that we don't know when exactly the file object will be closed, things would still be OK. The fd is a cheap resource, and it'd be totally OK to keep it open for some time. This is similar to various memory allocations that done for us, and for which we don't do explicit cleanup.

tl;dr: if you dont' do `del` on your variables, also don't bother with `with` for read-only file objects

PR #24434, opened 2022-08-25
boot: Small fixes

medhefgo:


src/boot/efi/util.c
```diff
@@ -760,12 +760,11 @@ EFI_STATUS make_file_device_path(EFI_HANDLE device, const char16_t *file, EFI_DE
...
-        FILEPATH_DEVICE_PATH *file_dp = (FILEPATH_DEVICE_PATH *) dp;
-        file_dp->Header.Type = MEDIA_DEVICE_PATH;
-        file_dp->Header.SubType = MEDIA_FILEPATH_DP;
-        memcpy(&file_dp->PathName, file, file_size);
-        SetDevicePathNodeLength(&file_dp->Header, SIZE_OF_FILEPATH_DEVICE_PATH + file_size);
+        /* Replace end node with file media device path. Use memcpy in case dp is unaligned. */
+        dp->Type = MEDIA_DEVICE_PATH;
+        dp->SubType = MEDIA_FILEPATH_DP;
+        memcpy((uint8_t *) dp + offsetof(FILEPATH_DEVICE_PATH, PathName), file, file_size);
+        SetDevicePathNodeLength(dp, SIZE_OF_FILEPATH_DEVICE_PATH + file_size);
```
poettering: not sure i follow here? if `dp` is aligned, then `file_dp` would be aligned too.

If things aren't aligned here, then you can't write to `dp->Type` either...

src/boot/efi/xbootldr.c
```diff
@@ -149,11 +133,18 @@ static EFI_STATUS try_gpt(
                 if (end < start) /* Bogus? */
                         continue;
 
-                ret_hd->PartitionNumber = i + 1;
-                ret_hd->PartitionStart = start;
-                ret_hd->PartitionSize = end - start + 1;
-                ret_hd->MBRType = MBR_TYPE_EFI_PARTITION_TABLE_HEADER;
-                ret_hd->SignatureType = SIGNATURE_TYPE_GUID;
+                *ret_hd = (HARDDRIVE_DEVICE_PATH) {
+                        .Header = (EFI_DEVICE_PATH) {
```
poettering: you can drop the `(EFI_DEVICE_PATH)` here. The type is clear. In fact, the fact you can specify it here at all is a GNU exension...

src/boot/efi/xbootldr.c
```diff
@@ -149,11 +133,18 @@ static EFI_STATUS try_gpt(
...
+                                .Type = MEDIA_DEVICE_PATH,
+                                .SubType = MEDIA_HARDDRIVE_DP,
+                                .Length = { 42 /* sizeof(HARDDRIVE_DEVICE_PATH) - padding */, 0 },
```
poettering: why not specify the field name where to assing the 42 and the 0 to as well?

src/boot/efi/xbootldr.c
```diff
@@ -243,8 +234,15 @@ static EFI_STATUS find_device(EFI_HANDLE *device, EFI_DEVICE_PATH **ret_device_p
                 }
 
                 /* Patch in the data we found */
-                EFI_DEVICE_PATH *xboot_path = path_dup(partition_path);
-                memcpy((uint8_t *) xboot_path + ((uint8_t *) part_node - (uint8_t *) partition_path), &hd, sizeof(hd));
+                size_t len = (uint8_t *) part_node - (uint8_t *) partition_path;
+                EFI_DEVICE_PATH *xboot_path = xmalloc(len + sizeof(hd) + END_DEVICE_PATH_LENGTH);
+                memcpy(xboot_path, partition_path, len);
+
+                part_node = (EFI_DEVICE_PATH *) ((uint8_t *) xboot_path + len);
```
poettering: btw, according to C, even such casts are illegal if the pointer is not aligned. Because the compiler automatically assumes alignment once a pointer has a certain type, and will generate optimized code from it. It's not legal to have an unaligned pointer of a specific type even if you never actually access it.

src/boot/efi/console.c
```diff
@@ -132,14 +132,15 @@ EFI_STATUS console_key_read(uint64_t *key, uint64_t timeout_usec) {
...
-                                shift |= EFI_SHIFT_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_CONTROL_PRESSED)
                                 shift |= EFI_CONTROL_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_ALT_PRESSED)
                                 shift |= EFI_ALT_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_LOGO_PRESSED)
                                 shift |= EFI_LOGO_PRESSED;
+
+                        /* Ideally we would handle shift here too, but that would prevent us from getting
+                         * uppercase letteres on some firmware. */
```
poettering: i am not sure i follow here? where precisely are the uppercase letters eaten? I mean, at this point you already have the key press info, and then just call KEYPRESS() to merge it, but that's a macro, and firmware is not involved anymore?

src/boot/efi/util.c
medhefgo: `EFI_DEVICE_PATH` has alignment of 1, so it's always fine (It's all bytes in it. The length field is technically uint16_t, but it is deliberately split into 2 bytes). But that of `FILEPATH_DEVICE_PATH` is 2 (and `dp` at this point is supposed to point to that). Mind you, device paths are a packed structure and we generally cannot just assume that the size leading up to `dp` gives the right alignment. We also cannot just create padding as at least OVMF didn't like that for HD device paths.

src/boot/efi/xbootldr.c
medhefgo: The spec has Length as a `uint16_t`, but both gnu-efi and edk2 define it as a `uint8_t[2]` to force `EFI_DEVICE_PATH` to have alignment of 1. (And this is what `DevicePathNodeLength` is for.

src/boot/efi/xbootldr.c
medhefgo: I know. `EFI_DEVICE_PATH` has alignment of 1 though. So it's fine.

src/boot/efi/util.c
poettering: packed as in gcc attribute packed?

medhefgo: > packed as in gcc attribute packed?

If only gnu-efi/edk2 would use that so we'd not have to care. One can dream… Some of their provided definitions even have tail padding, which at least OVMF doesn't like to see.

Quoting from the spec:
> Therefore all nodes are byte-packed data structures that may appear on any byte boundary. All code
references to device path notes [sic] must assume all fields are unaligned.

What's even more annoying that except for `EFI_DEVICE_PATH`, none of the device path structs are provide as struct definitions. They're all in (table-form) prose.

poettering: hmm, yuck. This really sounds to me as if we should put together the structs in heap or stack and then copy them into place, to deal with the alignment mess.

src/boot/efi/xbootldr.c
poettering: this deserves a comment. Or an assertion:

```c
assert_cc(alignof(EFI_DEVICE_PATH) == 1);
```

src/boot/efi/xbootldr.c
poettering: ah, this is an array, not a struct. go it. what a mess... maybe add a comment `/* NB: this is defined as array of uint8_t instead of single uint16-t to circumvent C's alignment rules */` or so

src/boot/efi/util.c
medhefgo: I fail to see the point in doing that here. You need to use memcpy regardles for FILEPATH_DEVICE_PATH, so I'm not going to do the same thing twice.

src/boot/efi/xbootldr.c
medhefgo: The definition is never gonna change, so I fail to see the reason. If the definition is ever gonna change, it will break code anyway (not just here but in initrd.c too).

keszybz: A comment helps explain it to the reader.

keszybz:
Looks all reasonable, but please add the two comments.

src/boot/efi/util.c
```diff
@@ -760,12 +760,12 @@ EFI_STATUS make_file_device_path(EFI_HANDLE device, const char16_t *file, EFI_DE
...
-        file_dp->Header.Type = MEDIA_DEVICE_PATH;
-        file_dp->Header.SubType = MEDIA_FILEPATH_DP;
-        memcpy(&file_dp->PathName, file, file_size);
-        SetDevicePathNodeLength(&file_dp->Header, SIZE_OF_FILEPATH_DEVICE_PATH + file_size);
+        /* Replace end node with file media device path. Use memcpy in case dp is unaligned (if accessed as
+         * FILEPATH_DEVICE_PATH). */
+        dp->Type = MEDIA_DEVICE_PATH;
+        dp->SubType = MEDIA_FILEPATH_DP;
+        memcpy((uint8_t *) dp + offsetof(FILEPATH_DEVICE_PATH, PathName), file, file_size);
+        SetDevicePathNodeLength(dp, SIZE_OF_FILEPATH_DEVICE_PATH + file_size);
```
poettering: looking in the headers, `SIZE_OF_FILEPATH_DEVICE_PATH` is defined as `EFI_FIELD_OFFSET(FILEPATH_DEVICE_PATH, PathName)`, hence equivalent to `offsetof(FILEPATH_DEVICE_PATH, PathName)`. The `memcpy()` call uses the latter, here you use the former. I'd stick to one way to write this. Probably use the offsetof() thing, simply because it's more readable.

src/boot/efi/util.c
```diff
@@ -760,12 +760,12 @@ EFI_STATUS make_file_device_path(EFI_HANDLE device, const char16_t *file, EFI_DE
         /* Point dp to the end node of the copied device path. */
         dp = (EFI_DEVICE_PATH *) ((uint8_t *) *ret_dp + dp_size - END_DEVICE_PATH_LENGTH);
 
-        /* Replace end node with file media device path. */
-        FILEPATH_DEVICE_PATH *file_dp = (FILEPATH_DEVICE_PATH *) dp;
-        file_dp->Header.Type = MEDIA_DEVICE_PATH;
-        file_dp->Header.SubType = MEDIA_FILEPATH_DP;
-        memcpy(&file_dp->PathName, file, file_size);
-        SetDevicePathNodeLength(&file_dp->Header, SIZE_OF_FILEPATH_DEVICE_PATH + file_size);
+        /* Replace end node with file media device path. Use memcpy in case dp is unaligned (if accessed as
```
poettering: maybe write `memcpy()` rather than `memcpy`

src/boot/efi/xbootldr.c
```diff
@@ -12,33 +12,30 @@ union GptHeaderBuffer {
...
 };
 
-static EFI_DEVICE_PATH *path_chop(EFI_DEVICE_PATH *path, EFI_DEVICE_PATH *node) {
+static EFI_DEVICE_PATH *path_replace(
+                const EFI_DEVICE_PATH *path,
+                const EFI_DEVICE_PATH *node,
+                const EFI_DEVICE_PATH *new_node) {
+
         assert(path);
         assert(node);
```
poettering, line 24: maybe a comment here that `path` and `node` must be part of the same path object, and that `new_node` must be a different object, would be good. Otherwise this is very hard to grok.

src/boot/efi/xbootldr.c
```diff
@@ -68,7 +68,7 @@ static bool verify_gpt(union GptHeaderBuffer *gpt_header_buffer, EFI_LBA lba_exp
         if (h->MyLBA != lba_expected)
                 return false;
 
-        if (h->SizeOfPartitionEntry < sizeof(EFI_PARTITION_ENTRY))
+        if (h->SizeOfPartitionEntry % sizeof(EFI_PARTITION_ENTRY))
```
poettering: as per coding style, please don't use C's downgrade-to-bool feature when not dealing with booleans, boolean-like stuff or pointers. For arithmetic values, please spell out what you are trying to say, i.e. write:

```c
if (h->SizeOfPartitionEntry % sizeof(EFI_PARTITION_ENTRY) != 0)
```

src/boot/efi/console.c
```diff
@@ -132,14 +132,18 @@ EFI_STATUS console_key_read(uint64_t *key, uint64_t timeout_usec) {
...
                                 shift |= EFI_CONTROL_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_ALT_PRESSED)
                                 shift |= EFI_ALT_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_LOGO_PRESSED)
                                 shift |= EFI_LOGO_PRESSED;
+
+                        /* Shift is not supposed to be reported for keys that can be represented as upperase
+                         * unicode chars (Shift+f is reported as F instead). Some firmware does it anyway, so
+                         * filter those out. */
+                        if (keydata.KeyState.KeyShiftState & EFI_SHIFT_PRESSED && keydata.Key.UnicodeChar == 0)
```
poettering: please enclose the binary `&` expression in `()`

src/boot/efi/console.c
```diff
@@ -132,14 +132,18 @@ EFI_STATUS console_key_read(uint64_t *key, uint64_t timeout_usec) {
...
-                        if (keydata.KeyState.KeyShiftState & EFI_SHIFT_PRESSED)
-                                shift |= EFI_SHIFT_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_CONTROL_PRESSED)
                                 shift |= EFI_CONTROL_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_ALT_PRESSED)
                                 shift |= EFI_ALT_PRESSED;
                         if (keydata.KeyState.KeyShiftState & EFI_LOGO_PRESSED)
                                 shift |= EFI_LOGO_PRESSED;
+
+                        /* Shift is not supposed to be reported for keys that can be represented as upperase
```
poettering: upperase → uppercase

man/systemd-boot.xml
```diff
@@ -257,12 +257,13 @@
...
       <varlistentry>
         <term><keycap>Esc</keycap></term>
+        <term><keycombo><keycap>Ctrl</keycap><keycap>c</keycap></keycombo></term>
         <listitem><para>Abort the edit and quit the editor</para></listitem>
       </varlistentry>
 
       <varlistentry>
         <term><keycombo><keycap>Ctrl</keycap><keycap>k</keycap></keycombo></term>
-        <listitem><para>Clear the command line</para></listitem>
+        <listitem><para>Clear the command line forwards</para></listitem>
```
poettering, line 266: why drop this? it only deletes to the end of line from the current location, no?

src/boot/bootctl.c
```diff
@@ -2331,29 +2331,28 @@ static int verb_is_installed(int argc, char *argv[], void *userdata) {
...
         assert(arg1);
         assert(ret_timeout);
         assert(ret_timeout_size);
 
-        if (streq(arg1, "menu-force"))
-                timeout = USEC_INFINITY;
-        else if (streq(arg1, "menu-hidden"))
-                timeout = 0;
+        if (STR_IN_SET(arg1, "menu-force", "menu-hidden"))
+                strcpy(utf8, arg1);
```
poettering: hmpf, strcpy(). that raises all alarm bells...

at the very least the utf8[] array should be declared as:

```c
char utf8[CONST_MAX3(DECIMAL_STR_MAX(usec_t), STRLEN("menu-force"), STRLEN("menu-hidden"));
```

(use CONST_MAX3() and STRLEN() instead of MAX3() and strlen() so this doesn't degrade into a VLA)

src/boot/bootctl.c
```diff
@@ -2331,29 +2331,28 @@ static int verb_is_installed(int argc, char *argv[], void *userdata) {
...
         else {
+                usec_t timeout;
                 r = parse_time(arg1, &timeout, USEC_PER_SEC);
                 if (r < 0)
                         return log_error_errno(r, "Failed to parse timeout '%s': %m", arg1);
-                if (timeout != USEC_INFINITY && timeout > UINT32_MAX * USEC_PER_SEC)
+                if (timeout != USEC_INFINITY && timeout > UINT32_MAX * USEC_PER_SEC) {
                         log_warning("Timeout is too long and will be treated as 'menu-force' instead.");
-        }
+                        timeout = USEC_INFINITY;
```
poettering: so why not write "menu-force" into the buffer in this case, too?

src/boot/bootctl.c
```diff
@@ -2331,29 +2331,28 @@ static int verb_is_installed(int argc, char *argv[], void *userdata) {
...
+                }
 
-        xsprintf(utf8, USEC_FMT, MIN(timeout / USEC_PER_SEC, UINT32_MAX));
+                xsprintf(utf8, USEC_FMT, MIN(timeout / USEC_PER_SEC, UINT32_MAX));
+        }
```
poettering: so, what about this: instead of having only utf8[] as an array, and then copying stuff into it that might be constant. Just rename utf8[] to buf[] and make utf8 a `const char*` pointer, that either points to buf[] (in ase of numeric value) or points to the literal string "menu-force" and "menu-hidden". 

That would mean: no ugly strcpy(), and everything is clean an properly sized.

src/boot/efi/boot.c
```diff
@@ -1532,6 +1541,34 @@ static void config_entry_add_type1(
...
+                return err;
+
+        if (streq16(value, u"menu-force")) {
+                *ret_value = TIMEOUT_MENU_FORCE;
+                return EFI_SUCCESS;
+        }
+        if (streq16(value, u"menu-hidden")) {
+                *ret_value = TIMEOUT_MENU_HIDDEN;
+                return EFI_SUCCESS;
+        }
```
poettering, line 1562: So, I am not convinced about this commit. This should be split out at the very least.

The thing is that "bootctl" and "sd-boot" might not be updated in lock-step (after all sd-boot is supposed to be shared between multiple OSes on the same host, and thus you might end up with sd-boot from fedora and a bootctl from suse in a very different version). So just setting these values from userspace and expecting them to work is not going to fly...

We communicate boot loader features from sd-boot to userspace via LoaderFeatures efi var. This is how we usually avoid such incompatibilities. But for this thing here it appears to be a bit too much.

Moroever, this stuff is actually defined as part of https://systemd.io/BOOT_LOADER_INTERFACE, i.e. is supposed to be stable...

man/systemd-boot.xml
medhefgo: I'm not dropping anything here, just clarifying. `Clear the command line` makes it sound like it deletes the whole thing, when it really just deletes forwards (just like your typical shell does).

src/boot/efi/boot.c
medhefgo: You know, I honestly think you're deluding yourself a bit on the BLS. I am actually unaware of any other boot loader implementing it after all these years (aside from some attempt by fedora for grub which afaik is bit-rotting and not upstreamed?).

Also, sd-boot and the BLS are already at odds with each other on this matter: `The time is specified in µs.`. sd-boot and bootctl always used seconds here.

man/systemd-boot.xml
poettering: oops, sorry, I inverted the patch in my head. sorry for the confusoin

src/boot/efi/boot.c
poettering: oh, the interface spec is IRL not so much about other boot loaders implementing it. It's more about other userspace tools making use of it. I happen to know some projects which set the efi vars in their own code and assume things just work. which is irrelevant for the issue at hand here, but it still makes the spec useful in general. And the fact it's officially independent of a specific implementation (though in reality it's about sd-boot) makes it easier politically for 3rd parties to support it without communicating so much they are binding things to systemd (even though they do)

src/boot/efi/xbootldr.c
```diff
@@ -243,9 +254,7 @@ static EFI_STATUS find_device(EFI_HANDLE *device, EFI_DEVICE_PATH **ret_device_p
                 }
 
                 /* Patch in the data we found */
-                EFI_DEVICE_PATH *xboot_path = path_dup(partition_path);
-                memcpy((uint8_t *) xboot_path + ((uint8_t *) part_node - (uint8_t *) partition_path), &hd, sizeof(hd));
-                *ret_device_path = xboot_path;
+                *ret_device_path = path_replace(partition_path, part_node, (EFI_DEVICE_PATH *) &hd);
```
poettering: hmm, using `&hd->header` as third arg should work too, no? and not require an ugly cast...

src/boot/efi/xbootldr.c
```diff
@@ -12,33 +12,33 @@ union GptHeaderBuffer {
...
+                new_node_len = DevicePathNodeLength(new_node);
 
-        memcpy(chopped, path, len);
-        SetDevicePathEndNode((EFI_DEVICE_PATH *) ((uint8_t *) chopped + len));
+        EFI_DEVICE_PATH *ret = xmalloc(len + new_node_len + END_DEVICE_PATH_LENGTH);
+        memcpy(ret, path, len);
 
-        return chopped;
-}
+        EFI_DEVICE_PATH *end = (EFI_DEVICE_PATH *) ((uint8_t *) ret + len);
```
poettering: maybe we should add GNU mempcpy() to our uefi apis, that would shorten this to:

```c
EFI_DEVICE_PATH *end = mempcpy(ret, path, len);
```
we use mempcpy() in userspace all the time

src/boot/efi/xbootldr.c
medhefgo: Should work, but that would make it look like it only operates on the header when it really works on the whole struct. That looks just wrong to me.

src/boot/efi/xbootldr.c
medhefgo: Yeah, there'd be quite a few users by now. Will do.

poettering:
So, PR looks good to me – except for the last commit. I think it's OK if sd-boot *reads* the textual values. I am also fine with sd-boot itself *sets* them. But I am pretty sure bootctl should stick to writing the values as numbers, for compat. We can revisit this in some years, but for now: we don't require userspace and boot loaders to be updated in lock step, and we really shouldn't.

src/boot/efi/boot.c
poettering: btw if you are talking about the boot loader spec (not the boot loader interface): fedora and others actually *do* implement that and use it by default in grub. I think they are working to get that upstream too.

src/boot/efi/xbootldr.c
poettering: but now it looks like you cast the pointer to something entirely unrelated...

poettering:
anyway, please also add the new textual values for those efi vars to the boot loader interface markdown texts

src/boot/efi/stub.c
```diff
@@ -79,20 +78,12 @@ static EFI_STATUS combine_initrd(
...
-        if (sysext_initrd) {
-                memcpy(p, sysext_initrd, sysext_initrd_size);
-                p += sysext_initrd_size;
-        }
+        if (credential_initrd)
+                p = mempcpy(p, credential_initrd, credential_initrd_size);
+        if (global_credential_initrd)
+                p = mempcpy(p, global_credential_initrd, global_credential_initrd_size);
+        if (sysext_initrd)
+                p = mempcpy(p, sysext_initrd, sysext_initrd_size);
```
poettering, line 86: ah, so much prettier

medhefgo: Btw, I found my very first gcc bug because of this :cat: : https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106868

poettering:
thanks!

lgtm!

PR #24438, opened 2022-08-25
mkosi: use the just built systemd-sysusers instead of groupadd

bluca:
It disappeared again from SUSE Tumbleweed. Just use sysusers since we
just built it.


bluca:
This gets a bit further, even though a gazillion unit tests fail at the next stage. Progress!

PR #24440, opened 2022-08-25
A couple of TEST-64 tweaks to make it usable without KVM

mrc0mmand:
Let's attempt to make TEST-64 usable with plain QEMU (TCG) by tweaking some limits and skipping only tests which require KVM.

~~This should, hopefully, make TEST-64 runnable in Ubuntu CI, since I tested it without KVM and with `QEMU_SMP=2` on an EC2 machine and it passed (albeit it took it around 18 minutes).~~ nope, Ubuntu CI is way too slow for this


test/test-functions
```diff
@@ -92,7 +92,7 @@ add_at_exit_handler() {
 # check if it's not explicitly disabled (TEST_NO_KVM) and we're not already
 # running under KVM. If these conditions are met, enable KVM (and possibly
 # nested KVM), otherwise disable it.
-if get_bool "${TEST_NESTED_KVM:=}" || (! get_bool "${TEST_NO_KVM:=}" && [[ "$(systemd-detect-virt -v)" != kvm ]]); then
+if get_bool "${TEST_NESTED_KVM:=}" || (! get_bool "${TEST_NO_KVM:=}" && ! systemd-detect-virt -v); then
```
yuwata, line 95: Not a matter, but maybe in such cases, calling `systemd-detect-virt` with `--quiet` is better for logging??
(Even if you agree with that, I think it is not necessary to be addressed in this PR. Let's consistently add the option later if you agree.)

yuwata (approved):

test/test-functions
mrc0mmand: That would definitely make sense. I'll add it to #24463 once this is merged.

PR #24445, opened 2022-08-25
mkdir: rewrite mkdir_p_root() with mkdir_p_safe()

yuwata:
There exist two similar functions. However, mkdir_p_root() has several
potential issues:
- if the path `p` is not absolute, e.g. `mkdir_p_root("/path/to/root", "foo")`,
  then `foo` is created under the current directory.
- the function is not tail recursive, hence may cause stack over flow.

Let's make mkdir_p_root() a tiny wrapper for mkdir_p_safe().

### DIFF
--- a/src/basic/mkdir.c
+++ b/src/basic/mkdir.c
@@ -167,6 +167,7 @@ int mkdir_p_internal(const char *prefix, const char *path, mode_t mode, uid_t ui
 
         /* Like mkdir -p */
 
+        assert(path);
         assert(_mkdirat != mkdirat);
 
         r = mkdir_parents_internal(prefix, path, mode, uid, gid, flags | MKDIR_FOLLOW_SYMLINK, _mkdirat);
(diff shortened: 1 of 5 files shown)


bluca:
Could you maybe add some tests in `src/test/test-mkdir.c`? To ensure that behaviour is really the same before/after, especially when there are symlinks escaping the root, etc

bluca (approved):

bluca:
```
stderr:
/* test_mkdir_p_safe */
/* test_mkdir_p_root */

==24507==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x555871656676 in __interceptor_realloc (/home/runner/work/systemd/systemd/build/test-mkdir+0xa4676) (BuildId: 7b06f094fee9c0f49262ccd2fd56195144bf724c)
    #1 0x7f9ba6dfc4d2 in path_extend_internal /home/runner/work/systemd/systemd/build/../src/basic/path-util.c:532:14
    #2 0x5558716927da in test_mkdir_p_root /home/runner/work/systemd/systemd/build/../src/test/test-mkdir.c:106:9
    #3 0x5558716936c7 in run_test_table /home/runner/work/systemd/systemd/build/../src/shared/tests.h:106:33
    #4 0x5558716936c7 in main /home/runner/work/systemd/systemd/build/../src/test/test-mkdir.c:118:1
    #5 0x7f9ba6314d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)

SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

src/basic/mkdir.c
```diff
@@ -19,7 +19,8 @@
 int mkdir_safe_internal(
                 const char *path,
                 mode_t mode,
-                uid_t uid, gid_t gid,
+                uid_t uid,
+                gid_t gid,
                 MkdirFlags flags,
                 mkdirat_func_t _mkdirat) {
```
poettering, line 26: (we should probably rework this whole function around open_mkdir_at(), i.e. create the dir, and immediately pin it via fd, and then adjust mode/ownership and verify it via the fd)

src/basic/mkdir.c
```diff
@@ -194,56 +195,16 @@ int mkdir_p_safe(const char *prefix, const char *path, mode_t mode, uid_t uid, g
...
-
-                nfd = openat(dfd, bn, O_RDONLY|O_CLOEXEC|O_DIRECTORY|O_NOFOLLOW);
-                if (nfd < 0)
-                        return -errno;
+        assert(path);
 
-                if (fchown(nfd, uid, gid) < 0)
-                        return -errno;
+        if (root) {
+                p = path_join(root, path);
```
poettering, line 204: mkdir_p_root() was about chase_symlinks(), i.e. that we mkdir a dir and don't leave the specified root while resolving the prefix to that dir. But here you just joine the path and then mkdir_p_safe() basically invokes mkdir() on the result, without understanding of the root parth thing

poettering (changes requested):

poettering:
So I figure mkdir_safe() should be reworked to first to chase_symlinks_and_open() with O_DIRECTORY on the parent dir of the specified path. It should then create the dir with open_mkdir_at(), and then tweak/check ownership/access mode via the returned fd.

That we mkdir_safe() becomes truly safe.

PR #18045, opened 2020-12-21
hostnamed: Filter out invalid DMI ids

felipeborges:
Values such as "To be filled by OEM", "TBD by OEM", "Default String",
"Board Manufacturer", "Board Vendor Name", "Board Product Name", "N/A".

See https://github.com/systemd/systemd/pull/15531#issuecomment-748167293

Filter taken from https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/sound/soc/soc-core.c#n1468

PR #24447, opened 2022-08-25
udev/rules: filter out mostly meaningless default strings

### LINKED ISSUE #24446: 'Hardware vendor' and 'Hardware model' produce useless identity information on non OEM system (Asus)

251

### Used distribution

Arch Linux

### Linux kernel version used

5.15.62

Referred from [https://gitlab.gnome.org/vstanek/gnome-info-collect/-/issues/23](https://gitlab.gnome.org/vstanek/gnome-info-collect/-/issues/23) @felipeborges 

My Asus motherboard doesn't have any useful identity info set under System Information in SMBIOS
```
System Information
	Manufacturer: System manufacturer
	Product Name: System Product Name
```

However, it does seem to have the right info set under Base Board Information
```
Base Board Information
	Manufacturer: ASUSTeK COMPUTER INC.
	Product Name: Z170 PRO GAMING
```

Looking at a few other boards over at [https://github.com/linuxhw/DMI](https://github.com/linuxhw/DMI), it appears to be a common occurrence (at least for Asus boards). Maybe the base board info should be used as a fallback if system information reports `System manufacturer`, `System Product Name` or `To be filled by O.E.M.`

[dmidecode](https://gist.github.com/dantob/9b476a0b53b42b2dfbef1d4cbe20e873)

yuwata:
The filter is generated based on the following results:
```shell
git clone git@github.com:linuxhw/DMI.git
cd DMI
git grep -h -A2 '^System Information$' | grep 'Manufacturer' | sort | uniq -c | sort -nr | less
git grep -h -A2 '^System Information$' | grep 'Product Name' | sort | uniq -c | sort -nr | less
```

Closes #24446.
Replaces #18045.

### DIFF
--- a/hwdb.d/20-dmi-id.hwdb
+++ b/hwdb.d/20-dmi-id.hwdb
@@ -1,5 +1,28 @@
 # This file is part of systemd
 
+# Filter out mostly meaningless names
+dmi:*:svnDefaultstring:*
+dmi:*:svnN/A:*
+dmi:*:svnO.E.M.:*
+dmi:*:svnOEM:*
+dmi:*:svnSystemmanufacturer:*
+dmi:*:svnSystemManufacturer:*
+dmi:*:svnTobefilledbyO.E.M.:*
+dmi:*:svnToBeFilledByO.E.M.:*
+    ID_SYS_VENDOR_IS_RUBBISH=1
+
+dmi:*:pnDefaultstring:*
+dmi:*:pnN/A:*
+dmi:*:pnO.E.M.:*
+dmi:*:pnOEM:*
+dmi:*:pnSystemproductname:*
+dmi:*:pnSystemProductName:*
+dmi:*:pnSystemname:*
+dmi:*:pnSystemName:*
+dmi:*:pnTobefilledbyO.E.M.:*
+dmi:*:pnToBeFilledByO.E.M.:*
+    ID_PRODUCT_NAME_IS_RUBBISH=1
+
 # Fix "Lenovo" capitalization in /sys/class/dmi/id/sys_vendor
 dmi:bvnLENOVO*
     ID_SYSFS_ATTRIBUTE_MODEL=product_version
--- a/rules.d/50-udev-default.rules.in
+++ b/rules.d/50-udev-default.rules.in
@@ -110,8 +110,9 @@ SUBSYSTEM=="ptp", ATTR{clock_name}=="KVM virtual PTP", SYMLINK += "ptp_kvm"
 SUBSYSTEM=="ptp", ATTR{clock_name}=="hyperv", SYMLINK += "ptp_hyperv"
 
 SUBSYSTEM!="dmi", GOTO="dmi_end"
-ENV{ID_VENDOR}="$attr{sys_vendor}"
-ENV{ID_SYSFS_ATTRIBUTE_MODEL}=="|product_name", ENV{ID_MODEL}="$attr{product_name}"
+ENV{ID_SYS_VENDOR_IS_RUBBISH}!="1", ENV{ID_VENDOR}="$attr{sys_vendor}"
+ENV{ID_SYSFS_ATTRIBUTE_MODEL}=="", ENV{ID_PRODUCT_NAME_IS_RUBBISH}!="1", ENV{ID_MODEL}="$attr{product_name}"
+ENV{ID_SYSFS_ATTRIBUTE_MODEL}=="product_name", ENV{ID_MODEL}="$attr{product_name}"
 ENV{ID_SYSFS_ATTRIBUTE_MODEL}=="product_version", ENV{ID_MODEL}="$attr{product_version}"
 # fallback to board information
 ENV{ID_VENDOR}=="", ENV{ID_VENDOR}="$attr{board_vendor}"


bluca (approved):

poettering:
This was proposed before and everyone agreed we should do this, but it should be done via hwdb, so that we don't have to list tons of strings in udev rules. See:

https://github.com/systemd/systemd/pull/18045

Would be good to warm that up again and get implemented.

yuwata:
The currently proposed PR #18045 filters out meaningless vendors and model from database. But this intends to filters out from sysfs.

Currently, hostnamed first try to get ID_VENDOR_FROM_DATABASE, then ID_VENDOR if the database entry is empty. Also a similar logic is used for the model. So, even if #18045 makes ID_VENDOR_FROM_DATABASE unset, still ID_VENDOR is set with e.g. "O.E.M" based on the sysfs attribute, and hostnamed will use it.

I do not agree with that we should check sysfs attribute something combined with hwdb, like you proposed `ID_SYS_VENDOR_IS_RUBBISH` in #18045. The database entry always has precedence over the sysfs entry. Hence, hwdb should only contain a meaningful value for each vendor, e.g. ASUS, DELL, and so on. Why we need to filter out sysfs attribute based on different attribute (ie modalias) and its matching hwdb entry??

In other words, even with this PR, each vendor or people can provide better string with appropriate modalias (if it has something specific to the product) through hwdb. Then hostnamed (and possible other applications) can use it through ID_VENDOR_FROM_DATABASE.

yuwata:
Or, or further in other words, hwdb is for genuine device specific values, not for default values. This provides a kind of default filtering. Hence, using hwdb is not suitable in this case, I think.

poettering:
the key point i am making is that list of strings like this PR contains really should be in hwdb. I am open to various different approaches how to precisely construct that, but we shouldn't have long string lists in udev rules...

yuwata:
Actually the concatenated string of the list is ugly long. But, is this only reason you do not like this PR??
Why longer hwdb is acceptable but not for udev rules?

Or, do you want to split the long line into shorter ones? (I can write the filter so, but I guess you neither want that...)

I usually agree with that device specific properties should be in hwdb. But I'd rephrase that in this case the filter provides a kind of 'default' behavior, not device specific.

poettering:
> Actually the concatenated string of the list is ugly long. But, is this only reason you do not like this PR?? Why longer hwdb is acceptable but not for udev rules?

hwdb is for static lookups, it's very efficient for that. It's "static data" if you so will, and not "code" which rules files are.

Distros are nowadays updating hwdb more often than the rest of systemd including the udev rules, and I think that's OK: the relatively static, rigid structure of hwdb allows us to decouple things a bit from the more actively developed udev rules.

The performance also matters to some point: udev rules are executed linearly, and so are it's match expressions. hwdb otoh is a lot more efficient, it's something like O(log(n)). Hence, let#s always prefer hwdb lookups for simple key-value stuff, over udev rules.

> Or, do you want to split the long line into shorter ones? (I can write the filter so, but I guess you neither want that...)
> 
> I usually agree with that device specific properties should be in hwdb. But I'd rephrase that in this case the filter provides a kind of 'default' behavior, not device specific.

hwdb is not just about device specific, it's also about device classes, i.e. stuff that applies to a bunch of devices at once. I'd see this as one of such occasions: whenever we see these "placeholder" strings, we should recognize them as such, regardles son which device we see it.

poettering:
the general goal here should be: if we maintain a list of literal strings for something, then it probably should be kept in hwdb.

yuwata:
> The performance also matters to some point: udev rules are executed linearly, and so are it's match expressions. hwdb otoh is a lot more efficient, it's something like O(log(n)). Hence, let#s always prefer hwdb lookups for simple key-value stuff, over udev rules.

Oh, that's an important point. OK. Now switched to hwdb. @poettering PTAL.

poettering:
CI failure:

```
 Invalid line rules.d/50-udev-default.rules:114: ENV{ID_SYSFS_ATTRIBUTE_MODEL}=="", ENV{ID_PRODUCT_NAME_IS_RUBBISH}!=1, ENV{ID_MODEL}="$attr{product_name}"
  clause: ENV{ID_PRODUCT_NAME_IS_RUBBISH}!=1
```
i.e. quotes missing

poettering:
lgtm otherwise

hwdb.d/20-dmi-id.hwdb
```diff
@@ -1,5 +1,26 @@
...
+dmi:*:pnDefaultstring:*
+dmi:*:pnO.E.M.:*
+dmi:*:pnOEM:*
+dmi:*:pnSystemproductname:*
+dmi:*:pnSystemProductName:*
+dmi:*:pnSystemname:*
+dmi:*:pnSystemName:*
+dmi:*:pnTobefilledbyO.E.M.:*
+dmi:*:pnToBeFilledByO.E.M.:*
+    ID_PRODUCT_NAME_IS_RUBBISH=1
```
poettering, line 24: hmm, what about "N/A"?

See @felipeborges's prior work: https://github.com/systemd/systemd/pull/18045/files

yuwata:
@poettering Thank you for the comments. Now, "N/A" is added to the list, and fixed the udev rule. PTAL.

PR #24458, opened 2022-08-26
optionally, embed PCR signature and public key in new sd-stub PE sections

poettering:
This is a companion for #24351 (which should be merged first).

This allows embedding PCR signatures and public keys in new PE sections in sd-stub based unified kernel images.

The two sections will be pushed as initrds into the booted kernel, where they can ultimately be picked up by cryptsetup/cryptenroll/systemd-creds.

With this it's relatively straight-forward to build kernels with a PCR signature policy that can be used to unlock a root fs.

### DIFF
--- a/TODO
+++ b/TODO
@@ -117,6 +117,11 @@ Deprecations and removals:
 
 Features:
 
+* during the initrd → host transition measure a fixed value into TPM PCR 11
+  (where we already measure the UKI into), so that unlock policies for disk
+  enryption/credential encryption can be put together that only work in the
+  initrd or only on the host (or both).
+
 * Add support for extra verity configuration options to systemd-reart (FEC, hash type, etc)
 
 * chase_symlinks(): take inspiraton from path_extract_filename() and return
@@ -160,11 +165,6 @@ Features:
 
 * systemd-measure tool:
   - pre-calculate PCR 12 (command line) + PCR 13 (sysext) the same way we can precalculate PCR 11
-  - sign pre-calculated hashes in a way compatible with TPM2 PCR hash signature
-    policies, in a way they can be included in unified PE kernel images, and
-    made available to userspace. There, this should be consumed by
-    systemd-cryptsetup to implement PCR signature based TPM volume unlock
-    policies.
 
 * in sd-boot: load EFI drivers from a new PE section. That way, one can have a
   "supercharged" sd-boot binary, that could carry ext4 drivers built-in.
@@ -249,8 +249,7 @@ Features:
 
 * repart: allow defining additional partitions via credential
 
-* tmpfiles: add snippet that provisions /etc/hosts, /etc/motd,
-  /root/.ssh/authorized_keys from credential
+* tmpfiles: add snippet that provisions /root/.ssh/authorized_keys from credential
 
 * timesyncd: pick NTP server info from credential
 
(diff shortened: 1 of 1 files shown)


poettering:
(this is not ready to commit, #24351 should be merged first. And this here lacks docs still which need to be added first)

poettering:
This should be ready for review. Docs added.

No tests (because we currently have no integration tests for the EFI stuff at all, which we should really do something about).

src/boot/efi/stub.c
```diff
@@ -305,6 +294,45 @@ EFI_STATUS efi_main(EFI_HANDLE image, EFI_SYSTEM_TABLE *sys_table) {
         if (sysext_measured)
                 (void) efivar_set_uint_string(LOADER_GUID, L"StubPcrInitRDSysExts", TPM_PCR_INDEX_INITRD_SYSEXTS, 0);
 
+        /* If the PCR signature was embedded in the PE image, then let's wrap it in a cpio and also pass it
+         * to the kernel, so that it can be read from /.extra/tpm2-pcr-signature.json. Note that this section
+         * is not measured, neither as raw section (see above), nor as cpio (here), because it is the
+         * signature of expected PCR values, i.e. it's input are PCR measurement, and hence it shouldn't
```
bluca, line 300: /it's/its/

man/systemd-measure.xml
```diff
@@ -104,6 +104,7 @@
         <term><option>--initrd=PATH</option></term>
         <term><option>--splash=PATH</option></term>
         <term><option>--dtb=PATH</option></term>
+        <term><option>--pcrpkey=PATH</option></term>
```
bluca, line 107: this is a bit of a mouthful - as a PE section name it might make sense to keep it short, but as a command line parameter can it be expanded to something readable? having both is also fine

TODO
```diff
@@ -406,14 +366,11 @@ Features:
...
-  generation counter into account that can only monotonically increase and can
-  be used to invalidate old PCR signatures. Otherwise people could downgrade to
-  old signed PCR sets whenever they want. Usecase: encrypt the rootfs with LUKS
-  with a key that can only be unlocked via a pristine pre-built Fedora
-  kernel+initrd.
+* TPM2: extend unlock policy to protect against version downgrades in signed
+  policies: policy probably must take some nvram based generation counter into
+  account that can only monotonically increase and can be used to invalidate
+  old PCR signatures. Otherwise people could downgrade to old signed PCR sets
+  whenever they want.
```
bluca, line 373: I have vague memories of a specific PCR that can take a policy that stops it from ever being reset, even by the kernel, is that something that could be useful for this?

bluca (approved):
This topic really ought to be covered in a new page for systemd.io I think? There's a lot of moving parts, so a single document summarizing everything, with use cases and examples, would be very useful

man/systemd-measure.xml
poettering: These map directly to the PE section name, just without the dot. Since these literally configure the PE section contents I think this should really stay that way.

In particular as we also have a --public-key= switch for selecting the public key to sign for. This can be quite confusing, i.e. the fact that there's one public key to embed and one to sign, so I think it's important to use the spelling we otherwise use for the PE section for the PE section stuff and the spelling we otherwise use for signing and stuff for signing and stuff.

Does that make sense?

I added an explicit paragraph to the docs of --public-key= btw which is supposed to highlight the relationship

bluca: It's really incomprehensible as a command line switch - can you at least add a readable one as an alias? So both can stay and be documented

TODO
poettering: Most PCRs cannot be reset unless you reboot. There are only a select few which can be reset by software, during system runtime. None of the PCRs are persistent, they all implicitly reset to zero on reboot.

But this is a different issue here: we want a counter that ensures that once you upgraded the OS you cannot go back with the OS and still retain access to your secrets. So the goal for that is a *persistent* counter we can increase on upgrades, and bind policy to when enrolling a disk against.

man/systemd-measure.xml
poettering: It sure is, but I still think its's easier to stick to the literal value here, since it directly maps to the objcopy command line that "system-measure sign" is usually paired with. i.e. from the example in the manpage:

```
 systemd-measure sign \
     --linux=vmlinux \
     --osrel=os-release.txt \
     --cmdline=cmdline.txt \
     --initrd=initrd.cpio \
     --splash=splash.bmp \
     --dtb=devicetree.dtb \
     --pcrpkey=tpm2-pcr-public.pem \
     --bank=sha1 \
     --bank=sha256 \
     --private-key=tpm2-pcr-private.pem \
     --public-key=tpm2-pcr-public.pem > tpm2-pcr-signature.json
```

and 

```
# objcopy \
    --add-section .linux=vmlinux --change-section-vma .linux=0x2000000 \
    --add-section .osrel=os-release.txt --change-section-vma .osrel=0x20000 \
    --add-section .cmdline=cmdline.txt --change-section-vma .cmdline=0x30000 \
    --add-section .initrd=initrd.cpio --change-section-vma .initrd=0x3000000 \
    --add-section .splash=splash.bmp --change-section-vma .splash=0x100000 \
    --add-section .dtb=devicetree.dtb --change-section-vma .dtb=0x40000 \
    --add-section .pcrsig=tpm2-pcr-signature.json --change-section-vma .splash=0x80000 \
    --add-section .pcrpkey=tpm2-pcr-public.pem --change-section-vma .splash=0x90000 \
    /usr/lib/systemd/boot/efi/linuxx64.efi.stub \
    foo.efi
```

i.e. the first 7 lines or so map to each other 1:1. it's very systematic in that. I think it would be a lot more confusing if we'd have to specify the sections under one name in systemd-measure and another one in objcopy. Simplest is to really reause the literal same spelling for both.

bluca: why not have an alias? then that still works, but the first impact on --help is way less confusing

poettering:
Will merge this now, given this has already been OK'ed pretty much. And will prep follow-up PR shortly with some more fixes.

poettering:
> This topic really ought to be covered in a new page for systemd.io I think? There's a lot of moving parts, so a single document summarizing everything, with use cases and examples, would be very useful

Yes, I will add that, it's high on my TODO list.

poettering:
> Will merge this now, given this has already been OK'ed pretty much. And will prep follow-up PR shortly with some more fixes.

→ #24620

medhefgo:
I feel like an accident is about to happen again if we keep adding more PE sections without programmatically deciding the VMA location: https://github.com/systemd/systemd/issues/24202#issuecomment-1236161697

PR #24459, opened 2022-08-26
tmpfiles: make absent source of "C" copy lines graceful

poettering:
This is split out of #24458, but I think makes a ton of sense on its own too.


keszybz:
LGTM.

PR #24460, opened 2022-08-26
sd-stub: some minor refactoring of the cpio/TPM code

poettering:
This is split out of #24458, but is a nice refactoring that makes sense on its own, too, and fixes a (minor) bug.

### DIFF
--- a/src/boot/efi/cpio.c
+++ b/src/boot/efi/cpio.c
@@ -488,7 +488,7 @@ nothing:
         *ret_buffer_size = 0;
 
         if (ret_measured)
-                *ret_measured = true;
+                *ret_measured = n_tpm_pcr > 0;
 
         return EFI_SUCCESS;
 }


src/boot/efi/cpio.c
```diff
@@ -304,6 +304,47 @@ static EFI_STATUS pack_cpio_trailer(
...
+                                buffer_size,
+                                tpm_description,
+                                &m);
+                if (err != EFI_SUCCESS) {
+                        log_error_stall(L"Unable to add initrd TPM measurement for PCR %u (%s), ignoring: %r", tpm_pcr[i], tpm_description, err);
+                        measured = false;
+                        continue;
+                }
+
+                measured = measured < 0 ? m : (measured && m);
```
keszybz: Maybe write this as
```C
if (measured != false)
     measured = m;
```
I think that's a bit easier to unwrap.

src/boot/efi/cpio.c
```diff
@@ -486,7 +486,7 @@ EFI_STATUS pack_cpio(
         *ret_buffer = NULL;
         *ret_buffer_size = 0;
 
-        if (ret_measured)
+        if (ret_measured && n_tpm_pcr > 0)
                 *ret_measured = true;
```
keszybz: Hmm, doesn't this violate the rule that the output variable should be always set?

keszybz (changes requested):

src/boot/efi/cpio.c
poettering: yes, i am an idiot ;-)

poettering:
pushed a new version, with the two points raised addressed. thanks!

keszybz (approved):
LGTM.

Issue #14155, opened 2019-11-26
how to unmask masked unit by systemd.mask= kernel command line option


yuwata:
**Is your feature request related to a problem? Please describe.**

When a system booted with `systemd.mask=foo.service`, then the service cannot be unmasked with `systemctl unmask foo.service`.

**Describe the solution you'd like**

Not sure...

**Describe alternatives you've considered**
```
chmod -x /usr/lib/systemd/system-generators/systemd-debug-generator
systemctl daemon-reload
chmod +x /usr/lib/systemd/system-generators/systemd-debug-generator
```

PR #24461, opened 2022-08-26
Better unmask message when unit is masked via kernel commandline

keszybz:


src/shared/install.h
```diff
@@ -95,88 +95,88 @@ struct UnitFileInstallInfo {
         bool auxiliary;
 };
 
-int unit_file_enable(
+int install_unit_enable(
```
poettering: I am not convinced this renaming is really a good idea. The D-Bus calls PID 1 exposes are all called UnitFileThis and ThatUnitFile, and that's API.

I think there's more value in keeping conistency between the dbus apis and the underlying C apis, than bikeshedding about the names...

src/shared/install.c
```diff
@@ -2302,6 +2307,25 @@ int install_unit_unmask(
                 if (!unit_name_is_valid(*name, UNIT_NAME_ANY))
                         return -EINVAL;
 
+                if (!root_dir) {
+                        /* If root_dir is set, we don't care about kernel commandline or generators */
+                        InstallInfo info = {
+                                .name = *name,
+                                .install_mode = _INSTALL_MODE_INVALID,
+                        };
```
poettering, line 2304: hmppf, having two ways how UnitFileInstallInfo can be allocated/freed is messy. In particular destruction, since who knows what unit_file_search() might initialize in future.

Why not just allocate it on the heap and use the regular destructor? This is not a performance sensitive inner loop, so the copies/allocations this costs really shouldn't matter...

src/shared/install.h
keszybz: The issue I was trying to solve is that the naming in the code is just misleading. E.g. you look at the code, and it says "files", but that's actually "unit names". And "unmask file", but it's really operating on a unit, which might have multiple files, etc. This creates a cognitive overhead in code which is very complicated anyway.

The D-Bus API cannot be changed, but the D-Bus API is something that is relatively simple and unchanging, so the amount of time spent looking at install.c is much much larger than the time spent looking at the D-bus wrappers which are just pass-through for the most part.

So I'd prefer to do the renaming, and maybe even follow it up with other renamings in the future.

src/shared/install.c
keszybz: Then in that case, it should probably be done the same way in the other place…

src/shared/install.h
poettering: i don't follow? this code this deals with files on disk, not with the runtime objects that these files create. hence unit files is exactly what this is about? I mean, sure, sometimes it's symlinks too, but that's effect how unix uses the word "files" for all types inodes usually.

keszybz: I'll paste the text from the commit message:

> The original names were conceived when a unit was equivalent to a unit file.
But the meaning has shifted over time, and while it is meanigful to say
"unit_file_enable", it is less so to say "unit_file_disable", since the
disablement can happen without a unit file being present. And we're not really
disabling the *file*, because the file is unchanged, and it's just the
surrounding state of that changes.

> Those functions are defined in install.[ch], so let's call them install_unit_*.
(This should be read as "install_(unit_enable)", not "(install_unit)_enable".)
And UnitFileChange → InstallChange.

poettering: Hmm? the symlinks to the unit file are removed. They string the symlinks point to reference the unit file, if it is missing or not. You cannot enable a unit really, only unit files, since the symlinks must point to something, which is a unit file, regardless if absent or not.

poettering: i mean if the symlinks wouldn't point to unit file paths i could follow your thinking, but they *do* point to unit file paths

keszybz: Let's start from other side. `unit_file_disable` does not operate on unit files. It operates on symlinks, which might point to the unit file, but also to other locations. And for the user, we say "disable unit", "could not disable unit", etc. This is particularly important in case of aliases, where we may operate on a single enablement symlink. If we were starting from scratch today, we certainly wouldn't have "file" in the name there. As I wrote above, this renaming is more relevant for some cases, less important for others. But it'd be ugly and confusing to have e.g. `install_unit_disable` and `unit_file_enable`, so I renamed them all to follow a common pattern.

keszybz:
I changed install.c to use `_cleanup_` for the InstallInfo struct everywhere. The result is indeed nicer.

src/shared/install.h
poettering: > but also to other locations

Like where?

> Let's start from other side. unit_file_disable does not operate on unit files. It operates on symlinks, which might point to the unit file, but also to other locations.

I still don't follow. For me "enabling" means something like "hooking in", "linking up", "setting up for activation", "empowering for doing", but all this words don't actually describe an operation on the thing you enable, but on auxiliary stuff you have to do for the thing to be enabled... So I am *really* not getting your point, at all.

I mean, I am not a native speaker. But I honestly, and truely don't follow what you are trying to say.

If I enable a car to got from A to B by building a road, then I build that road, and don't actually modify the car... If I fill up the car with petrol, then I move the petrol, I don't really modify the car. It's always the other thing you operate on to allow the thing you enable to do or be something.

Same here: to enable a unit file we create symlinks. But it's eally the unit file we enable, not the unit. We can't enable units at all.

Anyway, puzzled by this

src/shared/install.c
```diff
@@ -1040,22 +1040,28 @@ static int find_symlinks_in_scope(
         return 0;
 }
 
-static void install_info_free(InstallInfo *i) {
+static void install_info_clear(InstallInfo *i) {
```
poettering, line 1048: more often than not we call these "shallow destructors" `xyz_done()`. We probably should decide on a preferred naming and then sanction that in the coding style.

poettering: But then again we didn't even add to the coding style that _new()/_free()/_ref()/_unref() is what you are supposed to call constructors/destructors/reference counting calls...

src/shared/install.c
```diff
@@ -2287,12 +2292,32 @@ int install_unit_unmask(
...
+                                if (r != -ENOENT)
+                                        log_debug_errno(r, "Failed to look up unit %s, ignoring: %m", info.name);
+                        } else {
+                                if (info.install_mode == INSTALL_MODE_MASKED &&
+                                    path_is_generator(&lp, info.path))
+                                        install_changes_add(changes, n_changes,
+                                                            INSTALL_CHANGE_IS_MASKED_GENERATOR, info.name, info.path);
+                        }
+
+                        info.name = NULL;  /* … and give it back here */
```
poettering: `TAKE_PTR(info.name)`?

poettering:
Anyway, the two last commits look good to me. But I don't understand the renaming at all I must say. I don't understand the reasoning and I particularly dislike that the internal calls and the bus calls no widely deviate in naming even though one is just the backend for the other.

src/shared/install.c
keszybz: `git grep` says that `_clear()` is used more often than `_done()`…

keszybz:
I dropped the commit that changed the function names. The commits that changed or shortened the struc and field names remain. Also rebased and other comments are addressed. PTAL.

keszybz:
TEST-08-ISSUE-2730: [    0.008000] Kernel panic - not syncing: IO-APIC + timer doesn't work!  Boot with apic=debug and send a report.  Then try booting with the 'noapic' option.
Looks unrelated.

keszybz:
@poettering please take another look.

src/shared/install.c
```diff
@@ -2328,10 +2318,7 @@ int unit_file_unmask(
 
         r = 0;
         STRV_FOREACH(i, todo) {
-                _cleanup_free_ char *path = NULL;
-                const char *rp;
-
-                path = path_make_absolute(*i, config_path);
+                _cleanup_free_ char *path = path_make_absolute(*i, config_path);
```
yuwata: I am not a fan of mixing such changes into an unrelated commit... Of course, the change itself is ok.

src/shared/install.c
```diff
@@ -2347,7 +2334,7 @@ int unit_file_unmask(
 
                 install_changes_add(changes, n_changes, INSTALL_CHANGE_UNLINK, path, NULL);
 
-                rp = skip_root(lp.root_dir, path);
+                const char *rp = skip_root(lp.root_dir, path);
```
yuwata: Also here.

yuwata (approved):
LGTM.

yuwata:
I think it is OK to close #14155 by this PR.

src/shared/install.c
keszybz: I'll drop this from the commit. It's not really relevant.

keszybz:
Thank you for the review. Updated with just the two hunks removed.

keszybz:
centosci (archlinux + sanitzers) and jammy-ppc64el failed with timeouts. I don't think that's related, but I rebased to let it run again.

keszybz:
The tests pass now, so I assume the previous failure was unrelated.

src/shared/install.h
```diff
@@ -29,14 +29,14 @@ enum UnitFilePresetMode {
...
-        _UNIT_FILE_CHANGE_TYPE_MAX,
-        _UNIT_FILE_CHANGE_TYPE_INVALID = -EINVAL,
+        INSTALL_CHANGE_SYMLINK,
+        INSTALL_CHANGE_UNLINK,
+        INSTALL_CHANGE_IS_MASKED,
+        INSTALL_CHANGE_IS_DANGLING,
+        INSTALL_CHANGE_DESTINATION_NOT_PRESENT,
+        INSTALL_CHANGE_AUXILIARY_FAILED,
+        _INSTALL_CHANGE_MAX,
+        _INSTALL_CHANGE_INVALID = -EINVAL,
```
poettering, line 40: there probably should really `_INSTALL_ERRNO_MAX = -ERRNO_MAX,` here so that we cover the full error range. we do that at other places

PR #24463, opened 2022-08-26
(Not only) Ubuntu CI tweaks

mrc0mmand:
A couple of tweaks regarding timeouts and their handling in CIs.

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -92,7 +92,7 @@ add_at_exit_handler() {
 # check if it's not explicitly disabled (TEST_NO_KVM) and we're not already
 # running under KVM. If these conditions are met, enable KVM (and possibly
 # nested KVM), otherwise disable it.
-if get_bool "${TEST_NESTED_KVM:=}" || (! get_bool "${TEST_NO_KVM:=}" && ! systemd-detect-virt -v); then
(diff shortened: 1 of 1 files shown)


test/TEST-64-UDEV-STORAGE/test.sh
```diff
@@ -12,7 +12,6 @@ TEST_DESCRIPTION="systemd-udev storage tests"
 TEST_NO_NSPAWN=1
 # Save only journals of failing test cases by default (to conserve space)
 TEST_SAVE_JOURNAL="${TEST_SAVE_JOURNAL:-fail}"
-QEMU_TIMEOUT="${QEMU_TIMEOUT:-600}"
```
bluca, line 15: This might affect manual runs, and runs that we don't know about - what about keeping this and adding it to all other instances? So that if it's set by the CI we use it, but there's a default if not?

bluca: actually just now saw the next commit, that works just as fine

bluca (approved):

mrc0mmand:
focal-amd64 failed with a yet another timeout, but I'll push a fix for it later, once I finally debug TEST-13 lockups as well.

Issue #24453, opened 2022-08-25
DNSSEC test failures in TEST-75-RESOLVED


surajkrishnan14:
### systemd version the issue has been seen with

251.4

### Used distribution

Ubuntu 22.04

### Linux kernel version used

5.15.0-25-generic

### CPU architectures issue was seen on

x86_64

### Component

systemd-resolved

### Expected behaviour you didn't see

The test should pass when running:
`sudo make -C test/TEST-75-RESOLVED/ clean setup run `

### Unexpected behaviour you saw

The test fails with DNSSEC errors. I'll include a complete set of logs in the attachment, but a relavant snippet is:

```
testsuite-75.sh[31]: + grep -qF 'unsigned.test: 10.0.0.10' /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[31]: + grep -qF 'authenticated: no' /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[31]: + run dig @10.0.0.1 +short MX unsigned.test
testsuite-75.sh[161]: + dig @10.0.0.1 +short MX unsigned.test
testsuite-75.sh[162]: + tee /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[162]: 15 mail.unsigned.test.
testsuite-75.sh[31]: + grep -qF '15 mail.unsigned.test.' /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[31]: + run resolvectl query --legend=no -t MX unsigned.test
testsuite-75.sh[165]: + resolvectl query --legend=no -t MX unsigned.test
testsuite-75.sh[166]: + tee /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[166]: unsigned.test IN MX 15 mail.unsigned.test                   -- link: dns0
testsuite-75.sh[31]: + grep -qF 'unsigned.test IN MX 15 mail.unsigned.test' /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[31]: + : '--- ZONE: signed.systemd (static DNSSEC) ---'
testsuite-75.sh[31]: + run delv @10.0.0.1 signed.test
testsuite-75.sh[168]: + delv @10.0.0.1 signed.test
testsuite-75.sh[169]: + tee /tmp/tmp.PiS3oBFWnS
testsuite-75.sh[169]: ;; validating ./DNSKEY: no valid signature found (DS)
[...]

PR #24466, opened 2022-08-26
test: fix delv trust anchors location on Ubuntu

mrc0mmand:
delv on Ubuntu defaults to /etc/bind/bind.keys instead of /etc/bind.keys
when reading trust anchors, so let's create a symlink to make the test
work there as well.

Resolves: #24453

---

With this PR the test finally passes on Ubuntu Jammy (at least on my local VM). /cc @surajkrishnan14

### DIFF
--- a/test/units/testsuite-75.sh
+++ b/test/units/testsuite-75.sh
@@ -85,6 +85,8 @@ while read -ra line; do
 done < <(keymgr onlinesign.test. ds)
 knotc zone-commit test.
 
+knotc reload
+
 ### SETUP END ###
 
 : "--- nss-resolve/nss-myhostname tests"


yuwata (approved):
LGTM.

surajkrishnan14:
Thanks @mrc0mmand  - I tried this out, and ran into a different error. Attached logs:
[logs4.zip](https://github.com/systemd/systemd/files/9435170/logs4.zip)

mrc0mmand:
> Thanks @mrc0mmand - I tried this out, and ran into a different error. Attached logs: [logs4.zip](https://github.com/systemd/systemd/files/9435170/logs4.zip)

Ah, thank you. I managed to break the test on its first run, re-runs were working fine. I pushed an update which should, hopefully, fix this.

surajkrishnan14:
> > Thanks @mrc0mmand - I tried this out, and ran into a different error. Attached logs: [logs4.zip](https://github.com/systemd/systemd/files/9435170/logs4.zip)
> 
> Ah, thank you. I managed to break the test on its first run, re-runs were working fine. I pushed an update which should, hopefully, fix this.

Hmm, I think I'm still seeing the same error (can share logs if needed, LMK). I tried a clean build on commit eb1f34925e9c2952994f6bb5f0bcbcfe74838399

mrc0mmand:
> > > Thanks @mrc0mmand - I tried this out, and ran into a different error. Attached logs: [logs4.zip](https://github.com/systemd/systemd/files/9435170/logs4.zip)
> > 
> > 
> > Ah, thank you. I managed to break the test on its first run, re-runs were working fine. I pushed an update which should, hopefully, fix this.
> 
> Hmm, I think I'm still seeing the same error (can share logs if needed, LMK). I tried a clean build on commit [eb1f349](https://github.com/systemd/systemd/commit/eb1f34925e9c2952994f6bb5f0bcbcfe74838399)

Hmpf, you're right, interesting that it worked the ~5 times I tried it before pushing... let me dig a bit deeper.

mrc0mmand:
Interesting. There's a race condition where knotd sometimes ends up in a loop when waiting for DS RRs to get propagated to the respective parent zones. I've never seen this before, only on Ubuntu (afaik), however, reloading the zone configuration after committing all changes seems to mitigate this (or at least I can't reproduce the issue anymore). PTAL.

test/units/testsuite-75.sh
```diff
@@ -76,9 +79,7 @@ if knotc zone-get test. onlinesign.test. ds | grep .; then
     # Drop any old DS records, if present (e.g. on test re-run)
     knotc zone-unset test. onlinesign.test. ds
 fi
-knotc zone-commit test.
 # Propagate the new DS records
-knotc zone-begin test.
```
yuwata: The change in the third commit should be moved to the first commit.

mrc0mmand: Whoops, indeed, thanks.

surajkrishnan14:
> Interesting. There's a race condition where knotd sometimes ends up in a loop when waiting for DS RRs to get propagated to the respective parent zones. I've never seen this before on Arch/Fedora, only now on Ubuntu (afaik). However, reloading the zone configuration after committing all changes seems to mitigate this (or at least I can't reproduce the issue anymore). PTAL.

Thank you - I tried patch 8b1fd82222b6790ddce9c22b5251b473abe085b7 on this PR, and finally got the test suite passing on my ubuntu 22.04 VM! :-)

surajkrishnan14 (approved):

PR #24467, opened 2022-08-26
nspawn: add rootidmap as --bind option

qdeslandes:
`rootidmap` option for `--bind` will map the actual owner of the bind-mounted directory to the root user within the container.

Using `rootidmap`, files in a bind-mounted directory will be readable and writable by root in the container, while still being owned by the original user on the filesystem.

See https://github.com/systemd/mkosi/pull/1151 for example usage.

### DIFF
--- a/man/systemd-nspawn.xml
+++ b/man/systemd-nspawn.xml
@@ -1363,16 +1363,37 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
         multiple times for creating multiple independent bind mount points.</para>
 
         <para>Mount options are comma-separated. <option>rbind</option> and <option>norbind</option> control whether
-        to create a recursive or a regular bind mount. Defaults to "rbind". <option>idmap</option> and <option>noidmap</option>
-        control if the bind mount should use filesystem id mappings. Using this option requires support by the source filesystem
-        for id mappings. Defaults to "noidmap".</para>
+        to create a recursive or a regular bind mount. Defaults to "rbind". <option>noidmap</option>,
+        <option>idmap</option>, and <option>rootidmap</option> control ID mapping.</para>
+
+        <para>Using <option>idmap</option> or <option>rootidmap</option> requires support by the source filesystem
+        for user/group ID mapped mounts. Defaults to "noidmap". With <option>x</option> being the container's UID range
+        offset, <option>y</option> being the length of the container's UID range, and <option>p</option> being the
+        owner UID of the bind mount source inode on the host:
+
+        <itemizedlist>
+          <listitem><para>If <option>noidmap</option> is used, any user <option>z</option> in the range
+          <option>0 … y</option> seen from inside of the container is mapped to <option>x + z</option> in the
+          <option>x … x + y</option> range on the host. All host users outside of that range are mapped to
+          <option>nobody</option> inside the container.</para></listitem>
+          <listitem><para>If <option>idmap</option> is used, any user <option>z</option> in the UID range
+          <option>0 … y</option> as seen from inside the container is mapped to the same <option>z</option>
+          in the same <option>0 … y</option> range on the host. All host users outside of that range are
+          mapped to <option>nobody</option> inside the container.</para></listitem>
+          <listitem><para>If <option>rootidmap</option> is used, the user <option>0</option> seen from inside
+          of the container is mapped to <option>p</option> on the host. All host users outside of that range
+          are mapped to <option>nobody</option> inside the container.</para></listitem>
+        </itemizedlist></para>
+
+        <para>Whichever ID mapping option is used, the same mapping will be used for users and groups IDs. If
+        <option>rootidmap</option> is used, the group owning the bind mounted directory will have no effect</para>
 
         <para>Note that when this option is used in combination with <option>--private-users</option>, the resulting
         mount points will be owned by the <constant>nobody</constant> user. That's because the mount and its files and
         directories continue to be owned by the relevant host users and groups, which do not exist in the container,
         and thus show up under the wildcard UID 65534 (nobody). If such bind mounts are created, it is recommended to
         make them read-only, using <option>--bind-ro=</option>. Alternatively you can use the "idmap" mount option to
-        map the filesystem ids.</para></listitem>
+        map the filesystem IDs.</para></listitem>
       </varlistentry>
 
       <varlistentry>
--- a/src/nspawn/nspawn-mount.c
+++ b/src/nspawn/nspawn-mount.c
@@ -733,6 +733,8 @@ static int parse_mount_bind_options(const char *options, unsigned long *mount_fl
                         new_idmapping = REMOUNT_IDMAPPING_HOST_ROOT;
                 else if (streq(word, "noidmap"))
                         new_idmapping = REMOUNT_IDMAPPING_NONE;
+                else if (streq(word, "rootidmap"))
+                        new_idmapping = REMOUNT_IDMAPPING_HOST_OWNER;
                 else
                         return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
                                                "Invalid bind mount option: %s", word);
@@ -816,7 +818,7 @@ static int mount_bind(const char *dest, CustomMount *m, uid_t uid_shift, uid_t u
         }
 
         if (idmapping != REMOUNT_IDMAPPING_NONE) {
-                r = remount_idmap(where, uid_shift, uid_range, idmapping);
+                r = remount_idmap(where, uid_shift, uid_range, source_st.st_uid, idmapping);
                 if (r < 0)
                         return log_error_errno(r, "Failed to map ids for bind mount %s: %m", where);
         }
--- a/src/nspawn/nspawn.c
+++ b/src/nspawn/nspawn.c
@@ -3806,7 +3806,7 @@ static int outer_child(
             IN_SET(arg_userns_ownership, USER_NAMESPACE_OWNERSHIP_MAP, USER_NAMESPACE_OWNERSHIP_AUTO) &&
             arg_uid_shift != 0) {
 
-                r = remount_idmap(directory, arg_uid_shift, arg_uid_range, REMOUNT_IDMAPPING_HOST_ROOT);
+                r = remount_idmap(directory, arg_uid_shift, arg_uid_range, UID_INVALID, REMOUNT_IDMAPPING_HOST_ROOT);
                 if (r == -EINVAL || ERRNO_IS_NOT_SUPPORTED(r)) {
                         /* This might fail because the kernel or file system doesn't support idmapping. We
                          * can't really distinguish this nicely, nor do we have any guarantees about the
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -1380,7 +1380,7 @@ static int mount_partition(
                 (void) fs_grow(node, p);
 
         if (remap_uid_gid) {
-                r = remount_idmap(p, uid_shift, uid_range, REMOUNT_IDMAPPING_HOST_ROOT);
+                r = remount_idmap(p, uid_shift, uid_range, UID_INVALID, REMOUNT_IDMAPPING_HOST_ROOT);
                 if (r < 0)
                         return r;
         }
--- a/src/shared/mount-util.c
+++ b/src/shared/mount-util.c
@@ -1053,31 +1053,42 @@ int make_mount_point(const char *path) {
         return 1;
 }
 
-static int make_userns(uid_t uid_shift, uid_t uid_range, RemountIdmapping idmapping) {
+static int make_userns(uid_t uid_shift, uid_t uid_range, uid_t owner, RemountIdmapping idmapping) {
         _cleanup_close_ int userns_fd = -1;
         _cleanup_free_ char *line = NULL;
 
         /* Allocates a userns file descriptor with the mapping we need. For this we'll fork off a child
          * process whose only purpose is to give us a new user namespace. It's killed when we got it. */
 
-        if (asprintf(&line, UID_FMT " " UID_FMT " " UID_FMT "\n", 0u, uid_shift, uid_range) < 0)
-                return log_oom_debug();
-
-        /* If requested we'll include an entry in the mapping so that the host root user can make changes to
-         * the uidmapped mount like it normally would. Specifically, we'll map the user with UID_HOST_ROOT on
-         * the backing fs to UID 0. This is useful, since nspawn code wants to create various missing inodes
-         * in the OS tree before booting into it, and this becomes very easy and straightforward to do if it
-         * can just do it under its own regular UID. Note that in that case the container's runtime uidmap
-         * (i.e. the one the container payload processes run in) will leave this UID unmapped, i.e. if we
-         * accidentally leave files owned by host root in the already uidmapped tree around they'll show up
-         * as owned by 'nobody', which is safe. (Of course, we shouldn't leave such inodes around, but always
-         * chown() them to the container's own UID range, but it's good to have a safety net, in case we
-         * forget it.) */
-        if (idmapping == REMOUNT_IDMAPPING_HOST_ROOT)
-                if (strextendf(&line,
-                               UID_FMT " " UID_FMT " " UID_FMT "\n",
-                               UID_MAPPED_ROOT, 0u, 1u) < 0)
+        if (IN_SET(idmapping, REMOUNT_IDMAPPING_NONE, REMOUNT_IDMAPPING_HOST_ROOT)) {
+                if (asprintf(&line, UID_FMT " " UID_FMT " " UID_FMT "\n", 0u, uid_shift, uid_range) < 0)
                         return log_oom_debug();
+
+                /* If requested we'll include an entry in the mapping so that the host root user can make
+                 * changes to the uidmapped mount like it normally would. Specifically, we'll map the user
+                 * with UID_MAPPED_ROOT on the backing fs to UID 0. This is useful, since nspawn code wants
+                 * to create various missing inodes in the OS tree before booting into it, and this becomes
+                 * very easy and straightforward to do if it can just do it under its own regular UID. Note
+                 * that in that case the container's runtime uidmap (i.e. the one the container payload
+                 * processes run in) will leave this UID unmapped, i.e. if we accidentally leave files owned
+                 * by host root in the already uidmapped tree around they'll show up as owned by 'nobody',
+                 * which is safe. (Of course, we shouldn't leave such inodes around, but always chown() them
+                 * to the container's own UID range, but it's good to have a safety net, in case we
+                 * forget it.) */
+                if (idmapping == REMOUNT_IDMAPPING_HOST_ROOT)
+                        if (strextendf(&line,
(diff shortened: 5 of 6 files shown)


man/systemd-nspawn.xml
```diff
@@ -1363,9 +1363,11 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
...
 
         <para>Mount options are comma-separated. <option>rbind</option> and <option>norbind</option> control whether
-        to create a recursive or a regular bind mount. Defaults to "rbind". <option>idmap</option> and <option>noidmap</option>
-        control if the bind mount should use filesystem id mappings. Using this option requires support by the source filesystem
-        for id mappings. Defaults to "noidmap".</para>
+        to create a recursive or a regular bind mount. Defaults to "rbind". <option>idmap</option>, <option>noidmap</option>,
+        and <option>rootidmap</option> control if and how the bind mount should use filesystem id mappings. Using this option
+        requires support by the source filesystem for id mappings. <option>idmap</option> enables filesystem id mapping,
+        <option>noidmap</option> disables it, and <option>rootidmap</option> maps the owner of the bind-mounted directory to
+        the root user within the container. Defaults to "noidmap".</para>
```
poettering: if we add a third option here we should probably explain this better, i.e. say the net effect of each option, maybe something like this:

> With `x` being the container's UID range offset, and `y` being the length of the container's UID range, and `p` being the owner UID of the bind mount source inode on the host:
>
> 1. If `idmap` is used, any user `z` in the UID range `0 … y` as seen from inside the container is mapped to the same `z` in the same `0 … y` range on the host. All host users outside of that range are mapped to `nobody` inside the container.
>
> 2. If `noidmap` is used, any user `z` in the range `0 … y` seen from inside of the container is mapped to `x + z` in the `x … x + y` range on the host. All host users outside of that range are mapped to `nobody` inside the container.
>
> 3. If `rootidmap` is used, any user `z` in the range `1 … y` seen from inside of the container is mapped to `x + z` in the `x + 1 … x + y` range. The user `0` seen from inside of the container is mapped to `p` on the host. Al users outside of `x + 1 … x + y` and not equal to `p` on the host are mapped to `nobody` inside the container.

src/shared/dissect-image.c
```diff
@@ -1415,7 +1415,7 @@ static int mount_partition(
                 (void) fs_grow(node, p);
 
         if (remap_uid_gid) {
-                r = remount_idmap(p, uid_shift, uid_range, REMOUNT_IDMAP_HOST_ROOT);
+                r = remount_idmap(p, uid_shift, uid_range, 0 /* unused */, REMOUNT_IDMAPPING_HOST_ROOT);
```
poettering: pass UID_INVALID if it's unused

src/nspawn/nspawn.c
```diff
@@ -3806,7 +3806,7 @@ static int outer_child(
             IN_SET(arg_userns_ownership, USER_NAMESPACE_OWNERSHIP_MAP, USER_NAMESPACE_OWNERSHIP_AUTO) &&
             arg_uid_shift != 0) {
 
-                r = remount_idmap(directory, arg_uid_shift, arg_uid_range, REMOUNT_IDMAP_HOST_ROOT);
+                r = remount_idmap(directory, arg_uid_shift, arg_uid_range, 0 /* unused */, REMOUNT_IDMAPPING_HOST_ROOT);
```
poettering: pas UID_INVALID if its unused?

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,70 @@ int make_mount_point(const char *path) {
...
-         * (i.e. the one the container payload processes run in) will leave this UID unmapped, i.e. if we
-         * accidentally leave files owned by host root in the already uidmapped tree around they'll show up
-         * as owned by 'nobody', which is safe. (Of course, we shouldn't leave such inodes around, but always
-         * chown() them to the container's own UID range, but it's good to have a safety net, in case we
-         * forget it.) */
-        if (flags & REMOUNT_IDMAP_HOST_ROOT)
-                if (strextendf(&line,
-                               UID_FMT " " UID_FMT " " UID_FMT "\n",
-                               UID_MAPPED_ROOT, 0, 1) < 0)
+        if (idmapping == REMOUNT_IDMAPPING_NONE || idmapping == REMOUNT_IDMAPPING_HOST_ROOT) {
```
poettering: IN_SET()

src/shared/mount-util.h
```diff
@@ -118,18 +118,23 @@ int mount_image_in_namespace(pid_t target, const char *propagate_path, const cha
...
          * certain security implications defaults to off, and requires explicit opt-in. */
-        REMOUNT_IDMAP_HOST_ROOT = 1 << 0,
-} RemountIdmapFlags;
-
-int remount_idmap(const char *p, uid_t uid_shift, uid_t uid_range, RemountIdmapFlags flags);
+        REMOUNT_IDMAPPING_HOST_ROOT,
+        /* Include a mapping from root user within the container to the owner of the bind mounted directory.
+         * This ensure no root-owned files will be written in a bind-mounted directory owned by a different
+         * user. */
+        REMOUNT_IDMAPPING_HOST_OWNER,
```
poettering, line 134: For all our non-flag enums, we add two extra entries to the end:

```c
_REMOUNT_IDMAPPING_MAX,
_REMOUNT_IDMAPPING_INVALID = -EINVAL,
```

Simply as a matter of coding style. Please do so here too. It's useful once you intend to pass an invalid value in the enum, or want to know the valid rnage.

src/nspawn/nspawn-mount.c
```diff
@@ -815,8 +817,8 @@ static int mount_bind(const char *dest, CustomMount *m, uid_t uid_shift, uid_t u
                         return log_error_errno(r, "Read-only bind mount failed: %m");
         }
 
-        if (idmapped) {
-                r = remount_idmap(where, uid_shift, uid_range, REMOUNT_IDMAP_HOST_ROOT);
+        if (idmapping != REMOUNT_IDMAPPING_NONE) {
+                r = remount_idmap(where, uid_shift, uid_range, source_st.st_uid, idmapping);
```
poettering, line 821: the docs should say somewhere that the owning gid has no effect, we purely look at the uid, and gid is always copied from the uid

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,70 @@ int make_mount_point(const char *path) {
...
+                * accidentally leave files owned by host root in the already uidmapped tree around they'll show up
+                * as owned by 'nobody', which is safe. (Of course, we shouldn't leave such inodes around, but always
+                * chown() them to the container's own UID range, but it's good to have a safety net, in case we
+                * forget it.) */
+                if (idmapping == REMOUNT_IDMAPPING_HOST_ROOT) {
+                        if (strextendf(&line,
+                                       UID_FMT " " UID_FMT " " UID_FMT "\n",
+                                       UID_MAPPED_ROOT, 0, 1) < 0)
+                                return log_oom_debug();
+                }
```
poettering: no {} around single line if blocks

src/shared/mount-util.c
poettering: That said, I think this should probably be a `switch (idmapping) …` statement

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,70 @@ int make_mount_point(const char *path) {
...
+        }
+
+        if (idmapping == REMOUNT_IDMAPPING_HOST_OWNER) {
+                /* Remap the owner of the bind mounted directory to the root user within the container. This
+                 * way every file written by root within the container to bind-mounted directoryr will
```
poettering: typo

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,70 @@ int make_mount_point(const char *path) {
...
-        if (flags & REMOUNT_IDMAP_HOST_ROOT)
-                if (strextendf(&line,
-                               UID_FMT " " UID_FMT " " UID_FMT "\n",
-                               UID_MAPPED_ROOT, 0, 1) < 0)
+        if (idmapping == REMOUNT_IDMAPPING_NONE || idmapping == REMOUNT_IDMAPPING_HOST_ROOT) {
+                if (asprintf(&line, UID_FMT " " UID_FMT " " UID_FMT "\n", 0, uid_shift, uid_range) < 0)
                         return log_oom_debug();
 
+                /* If requested we'll include an entry in the mapping so that the host root user can make changes to
+                * the uidmapped mount like it normally would. Specifically, we'll map the user with UID_HOST_ROOT on
```
poettering: you might as well fix the typo here: UID_HOST_ROOT → UID_MAPPED_ROOT

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,70 @@ int make_mount_point(const char *path) {
...
+                                     UID_FMT " " UID_FMT " " UID_FMT "\n",
+                                     0, uid_shift + 1, uid_range - 1,
+                                     owner, uid_shift, 1) < 0)
+                                return log_oom_debug();
+                } else {
+                        if (asprintf(&line,
+                                     UID_FMT " " UID_FMT " " UID_FMT "\n"
+                                     UID_FMT " " UID_FMT " " UID_FMT "\n"
+                                     UID_FMT " " UID_FMT " " UID_FMT "\n",
+                                     0, uid_shift + 1, owner,
```
poettering: hmm, I am not sure I grok this: host uid 0 is mapped to uid 1 in the bind mount for the container? why that?

poettering:
Hmm, maybe I am a bit thick, but I don#t grok the new mapping...

What's the goal here anyway with rootidmap? Map everything with UID != 0 like `idmap`? Or map it like `noidmap`?  or leave it unmapped? I assumed the goal was to map that by `idmap`... But maybe leaving git unmapped might be better?

could really use a test case...

qdeslandes:
> Hmm, maybe I am a bit thick, but I don#t grok the new mapping...
> 
> What's the goal here anyway with rootidmap? Map everything with UID != 0 like `idmap`? Or map it like `noidmap`? or leave it unmapped? I assumed the goal was to map that by `idmap`... But maybe leaving git unmapped might be better?

Overall, `idmap` and `noidmap` are pretty similar, except that `idmap` has an extra mapping for host UID 0. The idea was to map `uid_shift` to host UID 0, but because we need to map the host directory owner to container UID 0, it creates a confusing non-linear mapping.

I would be ok only to map the directory owner to root and leave the remaining UIDs unmapped, but I wonder whether this could be an issue for some use cases. 

> could really use a test case...

Noted.

poettering:
> Overall, `idmap` and `noidmap` are pretty similar, except that `idmap` has an extra mapping for host UID 0. The idea was to map `uid_shift` to host UID 0, but because we need to map the host directory owner to container UID 0, it creates a confusing non-linear mapping.

You mean "rootidmap" somehwere here I presume?

src/nspawn/nspawn-mount.c
qdeslandes: Thinking about it, shouldn't we map the GID too? For the mkosi use case, for example, it would make sense for files to be owned by my UID and my GID on the host filesystem.

poettering:
> The idea was to make `rootidmap` similar to `idmap`: map all the UIDs and add a specific mapping for the use case (for `rootidmap` this would be mapping the directory owner). But we can't do this easily for `rootidmap` as the host UID we want to map is within the range `[host UID 0; host UID uid_range]`, so we end up with this very confusing mapping.

Hmm, but the owner might actually be outside of this range, no? i.e. some high UID > 0xFFFF?

src/shared/mount-util.h
poettering: not actually addressed?

src/shared/mount-util.h
```diff
@@ -128,9 +128,15 @@ typedef enum RemountIdmapping {
          * to add inodes to file systems mapped this way should set this flag, but given it comes with
          * certain security implications defaults to off, and requires explicit opt-in. */
         REMOUNT_IDMAPPING_HOST_ROOT,
+        /* Define a mapping from root user within the container to the owner of the bind mounted directory.
+         * This ensure no root-owned files will be written in a bind-mounted directory owned by a different
+         * user. No other users are mapped. */
+        REMOUNT_IDMAPPING_HOST_OWNER,
+        _REMOUNT_IDMAPPING_MAX,
+        _REMOUNT_IDMAPPING_INVALID = -EINVAL,
```
poettering, line 136: ah, you added this to the second commit. I figure these two lines should really be in the first commit

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,43 @@ int make_mount_point(const char *path) {
...
+                                return log_oom_debug();
+        }
+
+        if (idmapping == REMOUNT_IDMAPPING_HOST_OWNER) {
+                /* Remap the owner of the bind mounted directory to the root user within the container. This
+                 * way every file written by root within the container to the bind-mounted directory will
+                 * be owned by the original user. */
+                if (asprintf(&line, UID_FMT " " UID_FMT " " UID_FMT "\n", owner, uid_shift, 1u) < 0)
+                        return log_oom_debug();
+        }
```
poettering, line 1091: maybe add a comment emphasizing that all other users remain unmapped

poettering:
Love the simplicity of this version. This is looking good.

Looks pretty much ready to merge, just some minor nitpicks, see above.

src/shared/mount-util.h
qdeslandes: Definitely!

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,35 @@ int make_mount_point(const char *path) {
...
-         * (i.e. the one the container payload processes run in) will leave this UID unmapped, i.e. if we
-         * accidentally leave files owned by host root in the already uidmapped tree around they'll show up
-         * as owned by 'nobody', which is safe. (Of course, we shouldn't leave such inodes around, but always
-         * chown() them to the container's own UID range, but it's good to have a safety net, in case we
-         * forget it.) */
-        if (flags & REMOUNT_IDMAP_HOST_ROOT)
-                if (strextendf(&line,
-                               UID_FMT " " UID_FMT " " UID_FMT "\n",
-                               UID_MAPPED_ROOT, 0u, 1u) < 0)
+        if (IN_SET(idmapping, REMOUNT_IDMAPPING_NONE, REMOUNT_IDMAPPING_HOST_ROOT)) {
```
daandemeyer, line 1063: This seems like a behavior change on top of the rename, can we move this to the second commit?

man/systemd-nspawn.xml
```diff
@@ -1363,9 +1363,30 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
...
+          <listitem><para>If <option>idmap</option> is used, any user <option>z</option> in the UID range
+          <option>0 … y</option> as seen from inside the container is mapped to the same <option>z</option>
+          in the same <option>0 … y</option> range on the host. All host users outside of that range are
+          mapped to <option>nobody</option> inside the container.</para></listitem>
+          <listitem><para>If <option>rootidmap</option> is used, the user <option>0</option> seen from inside
+          of the container is mapped to <option>p</option> on the host. All host users outside of that range
+          are mapped to <option>nobody</option> inside the container.</para></listitem>
+        </itemizedlist></para>
+
+        <para>Whichever id mapping option is used, the same mapping will be used for users and groups id. If
```
daandemeyer: groups id => group ids

DaanDeMeyer (changes requested):
A few nits

man/systemd-nspawn.xml
```diff
@@ -1363,9 +1363,30 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
         multiple times for creating multiple independent bind mount points.</para>
 
         <para>Mount options are comma-separated. <option>rbind</option> and <option>norbind</option> control whether
-        to create a recursive or a regular bind mount. Defaults to "rbind". <option>idmap</option> and <option>noidmap</option>
-        control if the bind mount should use filesystem id mappings. Using this option requires support by the source filesystem
-        for id mappings. Defaults to "noidmap".</para>
+        to create a recursive or a regular bind mount. Defaults to "rbind". <option>noidmap</option>,
+        <option>idmap</option>, and <option>rootidmap</option> control id mapping.</para>
+
+        <para>Using <option>noidmap</option>, <option>idmap</option>, or <option>rootidmap</option> requires support
```
poettering: "noidmap" doesn't require that

man/systemd-nspawn.xml
```diff
@@ -1363,9 +1363,30 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
...
+        by the source filesystem for id mappings. Defaults to "noidmap". With <option>x</option> being the container's
```
poettering: "id mappings" → "user/group ID mapped mounts"

man/systemd-nspawn.xml
```diff
@@ -1363,9 +1363,30 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
...
+          <option>0 … y</option> as seen from inside the container is mapped to the same <option>z</option>
+          in the same <option>0 … y</option> range on the host. All host users outside of that range are
+          mapped to <option>nobody</option> inside the container.</para></listitem>
+          <listitem><para>If <option>rootidmap</option> is used, the user <option>0</option> seen from inside
+          of the container is mapped to <option>p</option> on the host. All host users outside of that range
+          are mapped to <option>nobody</option> inside the container.</para></listitem>
+        </itemizedlist></para>
+
+        <para>Whichever id mapping option is used, the same mapping will be used for users and groups ids. If
+        <option>rootidmap</option> is used, the group owning the bind mounted directory will have no effect</para>
```
poettering, line 1389: we usually upper case "id" in prose. Maybe do that here too. i.e. "ID mapping" and "… for user and group IDs"

and so on

src/shared/mount-util.c
```diff
@@ -1053,32 +1053,43 @@ int make_mount_point(const char *path) {
         return 1;
 }
 
-static int make_userns(uid_t uid_shift, uid_t uid_range, RemountIdmapping idmapping) {
+static int make_userns(uid_t uid_shift, uid_t uid_range,  uid_t owner, RemountIdmapping idmapping) {
```
poettering: spurious double space

man/systemd-nspawn.xml
```diff
@@ -1363,16 +1363,37 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
         multiple times for creating multiple independent bind mount points.</para>
 
         <para>Mount options are comma-separated. <option>rbind</option> and <option>norbind</option> control whether
-        to create a recursive or a regular bind mount. Defaults to "rbind". <option>idmap</option> and <option>noidmap</option>
-        control if the bind mount should use filesystem id mappings. Using this option requires support by the source filesystem
-        for id mappings. Defaults to "noidmap".</para>
+        to create a recursive or a regular bind mount. Defaults to "rbind". <option>noidmap</option>,
+        <option>idmap</option>, and <option>rootidmap</option> control ID mapping.</para>
+
+        <para>Using <option>idmap</option>, or <option>rootidmap</option> requires support by the source filesystem
```
poettering: ", or" → " or"

Issue #24360, opened 2022-08-19
TEST-60-MOUNT-RATELIMIT fails on the latest Arch


mrc0mmand:
### systemd version the issue has been seen with

latest main

### Used distribution

Arch Linux

### Linux kernel version used

5.19.2-arch1-1

### CPU architectures issue was seen on

x86_64

### Component

systemd-udevd, tests

### Unexpected behaviour you saw

Another week, another test fail.

Monday and yesterday cron job's attempts at updating the Arch Linux CI images were unsuccessful due to TEST-60-MOUNT-RATELIMIT consistently and reliably failing - specifically when waiting for the loop device via `udevadm wait` in this section:

https://github.com/systemd/systemd/blob/570fe6f12b96b47eff0a966fa589da77cbc5e0fe/test/units/testsuite-60.sh#L43-L44

I bumped the timeout multiple times without any change.

```
[  129.394248] testsuite-60.sh[815]: + test_dependencies
[  129.394455] testsuite-60.sh[815]: + systemd-detect-virt --quiet --container
[  129.402956] testsuite-60.sh[815]: + trap teardown_test_dependencies RETURN
[  129.402956] testsuite-60.sh[815]: + setup_loop 0
[  129.405494] testsuite-60.sh[815]: + truncate -s 30m /tmp/testsuite-60-dependencies-0.img
[  129.409310] testsuite-60.sh[815]: + sfdisk --wipe=always /tmp/testsuite-60-dependencies-0.img
[  129.697661] testsuite-60.sh[1309]: Checking that no-one is using this disk right now ... OK
[  129.697661] testsuite-60.sh[1309]: Disk /tmp/testsuite-60-dependencies-0.img: 30 MiB, 31457280 bytes, 61440 sectors
[  129.697661] testsuite-60.sh[1309]: Units: sectors of 1 * 512 = 512 bytes
[  129.697661] testsuite-60.sh[1309]: Sector size (logical/physical): 512 bytes / 512 bytes
[  129.697661] testsuite-60.sh[1309]: I/O size (minimum/optimal): 512 bytes / 512 bytes
[  129.697661] testsuite-60.sh[1309]: >>> Script header accepted.
[  129.697661] testsuite-60.sh[1309]: >>> Created a new GPT disklabel (GUID: E6B37A3A-1BD5-DE4C-AAED-1360C39F8046).
[...]

PR #24471, opened 2022-08-27
udevadm-wait: introduce periodic timer for checking devices

yuwata:


yuwata:
@mrc0mmand Still TEST-64 is too slow for non-kvm env??

mrc0mmand:
> @mrc0mmand Still TEST-64 is too slow for non-kvm env??

I was just looking at the log and wanted to ask you if the fail is expected, since there are too many events and the non-accelerated machine just can't keep up, or there's something else going on. If the former is the case, I guess bumping the timeout to 240s should, hopefully, suffice: https://github.com/systemd/systemd/blob/6dd3b818bbfc6a2c35e5a530186a8c0bc1e01118/test/units/testsuite-64.sh#L183

yuwata:
All events are processed after ~183sec. Hence, extending timeout to 240sec or so seems to work. I will extend it. Thank you for the suggestion.
```
[  314.231415] H testsuite-64.sh[658]: Timed out for waiting devices being initialized.
(snip)
[  317.120767] H systemd-udevd[604]: sda9: sd-device-monitor(worker): Passed 1272 byte to netlink monitor.
[  317.122004] H systemd-udevd[508]: No events are queued, removing /run/udev/queue.
```

src/udev/udevadm-wait.c
```diff
@@ -246,6 +254,59 @@ static int setup_timer(sd_event *event) {
...
+        _cleanup_(sd_event_source_unrefp) sd_event_source *s = NULL;
+        int r;
+
+        assert(event);
+
+        /* If --initialized=no is specified, it is not necessary to wait uevents for the specified devices to
+         * be processed by udevd. Let's periodically check the devices. Then, we may be able to finish this
+         * program earlier when udevd is very busy.
+         *
+         * This is useful for workarounding issues #24360 and #24450.
```
bluca: "workarounding" is a great idea for a verb, but unfortunately it is just a noun for now, so: "working around"

bluca (approved):

src/udev/udevadm-wait.c
```diff
@@ -367,6 +428,10 @@ int wait_main(int argc, char *argv[], void *userdata) {
         if (r < 0)
                 return log_error_errno(r, "Failed to set up device monitor: %m");
 
+        r = setup_periodic_timer(event);
+        if (r < 0)
+                return log_error_errno(r, "Failed to set up periodic timer: %m");
```
poettering, line 433: grmbl, this seesm very hacky. I'd be really curious to understand why these uevents go missing. There's some bug we should try to track down instead of taping over.

src/udev/udevadm-wait.c
```diff
@@ -246,6 +254,59 @@ static int setup_timer(sd_event *event) {
...
+        /* If --initialized=no is specified, it is not necessary to wait uevents for the specified devices to
+         * be processed by udevd. Let's periodically check the devices. Then, we may be able to finish this
+         * program earlier when udevd is very busy.
+         *
+         * This is useful for working around issues #24360 and #24450.
+         * For some reasons, the kernel sometimes does not emit uevents for loop block device on attach.
+         * Hence, without the periodic timer, no event source for this program will be triggered, and this
+         * will be timed out.
+         * Theoretically, inotify watch may be better, but this program typically expected to run in a short
+         * time. Hence, let's use the simpler periodic timer event source here. */
```
poettering, line 298: i am not sure it should be expected to run only a short time. think boot, where we wait for a LUKS block device, which might involve user interaction to unlock it. that might take a loooooooooooooooooooooong time...

src/udev/udevadm-wait.c
```diff
@@ -367,6 +428,10 @@ int wait_main(int argc, char *argv[], void *userdata) {
         if (r < 0)
                 return log_error_errno(r, "Failed to set up device monitor: %m");
 
+        r = setup_periodic_timer(event);
+        if (r < 0)
+                return log_error_errno(r, "Failed to set up periodic timer: %m");
+
```
poettering, line 434: i think a more interesting optimization would be to simply hook into the uevent stream from the kernel instead of the one to from udev if we don't need to wait for initialization. i.e. use the other netlink group, the one udev itself listens on. That way we'd still wait for the right event, but not be delayed by udev.

yuwata: Yeah, but the kernel sometimes does not send uevent for loopback block devices...

poettering: but we should figure out why. that sounds like a bug in the kernel? 

I mean, it appears uevents get lost on the CI sometimes, but are we even sure it's the kernel that loses them and not udev?

do we know if they are simply never generated? or maybe they are dropped because netlink buffer is full? But if the latter, isn't there some mechanism to detect such overruns, and then handle them somehow (maybe by rescanning all devices?)

yuwata: There are nothing I can say sure, but
> do we know if they are simply never generated?

At least from the log of the tests, uevents are not generated, at least udevd does not receive them.
See the logs in https://github.com/systemd/systemd/issues/24360.

> or maybe they are dropped because netlink buffer is full?

The possibility of this is quite small. udevadm-wait also increases the buffer size as same as the udevd uses.

> But if the latter, isn't there some mechanism to detect such overruns

This sounds nice, even though I guess the issue is not on the buffer size.

PR #24474, opened 2022-08-27
udevadm-settle: several cleanups

yuwata:


yuwata:
The TEST-64 failure in CentOS CI is unrelated, and fixed by #24471.

src/udev/udevadm-settle.c
```diff
@@ -167,11 +167,6 @@ int settle_main(int argc, char *argv[], void *userdata) {
         if (r <= 0)
                 return r;
 
-        if (running_in_chroot() > 0) {
```
bluca: why is this no longer necessary? please explaini in the commit message

bluca (approved):

src/udev/udevadm-settle.c
yuwata: Nah, I thought it is a uid check, but checking chroot. I overlooked 'ch'...

src/udev/udevadm-settle.c
```diff
@@ -194,6 +194,10 @@ int settle_main(int argc, char *argv[], void *userdata) {
                 r = udev_ctrl_wait(uctrl, MAX(5 * USEC_PER_SEC, arg_timeout));
                 if (r < 0)
                         return log_error_errno(r, "Failed to wait for daemon to reply: %m");
+        } else {
+                /* For non-privileged users, at least check if udevd is running. */
+                if (access("/run/udev/control", F_OK) < 0)
+                        return log_error_errno(errno, "udevd is not running.");
```
poettering, line 221: hmm, this might fail due to EACCES or so (think selinux), in which case the error message is quite confusing...

I'd prefer if we'd output the error string if unexpected error. i.e.

```c
if (access("/run/udev/control", F_OK) < 0) {
        if (errno == ENOENT)
                return log_error_errno(errno, "systemd-udevd is not running.");
        return log_error_errno(errno, "Failed to check if /run/udev/control exists: %m");
``

yuwata: Thanks. #24519.

Issue #24450, opened 2022-08-25
TEST-50-DISSECT: The Return of the Fail 2 (aka the sequel nobody wanted)


mrc0mmand:
### systemd version the issue has been seen with

latest main

### Used distribution

Arch Linux

### Linux kernel version used

5.19.2-arch1-2

### Unexpected behaviour you saw

Oh well, looks like only the second issue from #24390 was resolved, the first one is still present:

```
[   33.170687] testsuite-50.sh[535]: + systemd-dissect --json=short --root-hash e9633272e0fbb222e5fe59e542eb95723972fdba22cfdee1422ce8ebc1f2ce17 /tmp/tmp.d5Wfbk/minimal_0.gpt
[   33.171400] testsuite-50.sh[536]: + grep -q '{"rw":"ro","designator":"root","partition_uuid":"e9633272-e0fb-b222-e5fe-59e542eb9572","partition_label":"Root Partition","fstype":"squashfs","architecture":"x86-64","verity":"yes",'
[   33.171972] testsuite-50.sh[535]: Opened '/tmp/tmp.d5Wfbk/minimal_0.gpt' in O_RDONLY access mode, with O_DIRECT disabled (O_DIRECT was requested but not supported).
[   33.172623] testsuite-50.sh[535]: Failed to enable O_DIRECT for backing file descriptor for loopback device. Continuing without.
[   33.173404] testsuite-50.sh[535]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/power/uevent" does not exist.
[   33.174054] testsuite-50.sh[535]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/holders/uevent" does not exist.
[   33.174770] testsuite-50.sh[535]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/uevent" does not exist.
[   33.175469] testsuite-50.sh[535]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/uevent" does not exist.
[   33.176217] testsuite-50.sh[535]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu11/uevent" does not exist.
[   33.176949] testsuite-50.sh[535]: sd-device: the uevent file "/sys/devices/virtual/block/loop0/mq/0/cpu9/uevent" does not exist.
[...]

PR #24482, opened 2022-08-28
loop-util: attach an empty file before unlocking device

yuwata:


src/shared/loop-util.c
```diff
@@ -145,16 +145,17 @@ static int loop_configure(
          * block devices. */
         r = device_has_block_children(nr);
         if (r < 0)
-                return r;
+                return log_debug_errno(r, "Failed to determine if the loopback device has partitions: %m");
         if (r > 0) {
                 r = loop_is_bound(fd);
                 if (r < 0)
-                        return r;
+                        return log_debug_errno(r, "Failed to determine if a file descritor is attached to the loopback device: %m");
```
mrc0mmand: descritor -> descriptor

mrc0mmand (approved):
The loopback devices are truly an infinite source of _fun_.

bluca:
Why is the diskseq stuff being removed? We got it added to newer kernels to fix these races - of course won't help with older kernels, but at some point if should work out no?

yuwata:
> Why is the diskseq stuff being removed? We got it added to newer kernels to fix these races - of course won't help with older kernels, but at some point if should work out no?

As you can see, at least now it is not used, ie the argument is completely ignored. Please revert it when it is used.

yuwata:
Or, are there any places I dropped in this PR where the diskseq is really used?

bluca:
But it was used when it was introduced? Where did that stuff go?

yuwata:
I cannot dig the git history now from cellphone, but I guess these entries in LoopDevice were stopped to be used when Lennart stop to use sd_device_monitor for detecting loopback devices.

bluca:
@poettering do you remember what's the full story? It was quite hard to get the diskseq stuff in the kernel and it was explicitly to fix these issues, it's a pity if it was all for naught...

yuwata:
@bluca diskseq or friends are not used since 1b010ae7c96d953f11a9925d4c5aef29befc86d1.

yuwata:
BTW @mrc0mmand Does this help issue #24450?? Could you run a stress test for TEST-50-DISSECT with this PR?

mrc0mmand:
> BTW @mrc0mmand Does this help issue #24450?? Could you run a stress test for TEST-50-DISSECT with this PR?

I ran TEST-50-DISSECT 100 times in a row with this PR and it didn't fail a single time, so I'm inclined to say it might potentially fix that issue (but won't say it directly, otherwise something else breaks).

src/shared/loop-util.h
```diff
@@ -14,9 +14,6 @@ struct LoopDevice {
         dev_t devno;
         char *node;
         bool relinquished;
-        uint64_t diskseq; /* Block device sequence number, monothonically incremented by the kernel on create/attach, or 0 if we don't know */
```
poettering, line 17: I'd really keep the diskseq, it's how we can pin our use of the loopback device. Even if the current dissection stuff doesn#t use it, it probably should, and condition the partition removal to only execute if it still matches the original diskseq (enclosed in a BSD lock)

src/shared/loop-util.c
```diff
@@ -55,16 +55,23 @@ static int loop_is_bound(int fd) {
         return true; /* bound! */
 }
 
-static int device_has_block_children(sd_device *d) {
+static int device_has_block_children(int nr) {
```
poettering, line 58: should be renamed to `loop_block_device_has_children()` or so. It doesn't work for generic block devices anymore, after all, just for loopback devices

src/shared/loop-util.c
```diff
@@ -109,6 +109,44 @@ static int device_has_block_children(int nr) {
...
+static int attach_empty_file(int loop) {
+        _cleanup_close_ int fd = -1;
+
+        /* So here's the thing: on various kernels (5.8 at least) loop block devices might enter a state
+         * where they are detached but nonetheless have partitions, when used heavily. Accessing these
+         * partitions results in immediatey IO errors. There's no pretty way to get rid of them
+         * again. Neither LOOP_CLR_FD nor LOOP_CTL_REMOVE suffice (see above). What does work is to
+         * reassociate them with a new fd however. This is what we do here hence: we associate the devices
+         * with an empty file (i.e. an image that definitely has no partitions). We then immediately clear it
+         * again. This suffices to make the partitions go away. Ugly but appears to work. */
```
poettering, line 121: so, back when i added this code i wasn't aware of BLKPG_DEL_PARTITION. I wonder if a loop around that wouldn't have the same effect but be less ugly?

i.e. remove all partitions in a loop, until none are left?

poettering (changes requested):

poettering:
Inspired by this I prepped a PR in #24530, which fixes a bunch of related things. It also gets rid of the attach empty file logic entirely, in favour of explicit partition deletion, as I suggested above.

That new PR makes the last two of the commits in this PR redundant, I guess. (one because it drops the empty file attach stuff, the other because we actually add proper locking around clear+delete of block devices. at least in my tests this makes things entirely clean in test-loop-block even with 200 threads and 200 iterations)

yuwata:
@poettering Thanks. I will update or close this after your PR is merged.

PR #24487, opened 2022-08-28
scsi_id: retry inquiry ioctl if host_byte is DID_TRANSPORT_DISRUPTED

wenchao-hao:
The inquiry is issued to kernel via ioctl, kernelspace would set
this inquiry command's retry count to 0 which means the command
would not be retried in kernel space even if the LLDs returned
a status which need to be retried. So we should take the retry
in user space.

### DIFF
--- a/src/udev/scsi_id/scsi_serial.c
+++ b/src/udev/scsi_id/scsi_serial.c
@@ -69,6 +69,7 @@ static const char hex_str[]="0123456789abcdef";
 #define DID_NO_CONNECT               0x01        /* Unable to connect before timeout */
 #define DID_BUS_BUSY                 0x02        /* Bus remain busy until timeout */
 #define DID_TIME_OUT                 0x03        /* Timed out for some other reason */
+#define DID_TRANSPORT_DISRUPTED      0x0e        /* Transport disrupted and should retry */
 #define DRIVER_TIMEOUT               0x06
 #define DRIVER_SENSE                 0x08        /* Sense_buffer has been set */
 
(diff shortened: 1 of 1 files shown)


src/udev/scsi_id/scsi_serial.c
```diff
@@ -126,6 +128,8 @@ static int sg_err_category_new(int scsi_status, int msg_status, int
         if (host_status) {
                 if (IN_SET(host_status, DID_NO_CONNECT, DID_BUS_BUSY, DID_TIME_OUT))
                         return SG_ERR_CAT_TIMEOUT;
+                if (IN_SET(host_status, DID_TRANSPORT_DISRUPTED))
```
yuwata: Please use `==` if there exists only matching entry.

yuwata (changes requested):
Please drop the signed-off-by tag, unless if it is really requested by your organization or company.

wenchao-hao:
> Please drop the signed-off-by tag, unless if it is really requested by your organization or company.

Updated

src/udev/scsi_id/scsi_serial.c
yuwata: This is not addressed yet.

wenchao-hao: > This is not addressed yet.

I did not notice it, updated

yuwata:
This is superficially good. But I cannot judge if this is correct. Could you provide any reference about the change?

yuwata:
Ah, I found the source in kernel's include/scsi/scsi_status.h. LGTM.

PR #24491, opened 2022-08-29
condition: unify operator parsing handling

poettering:
A bunch of fixes to operator parsing (i.e. "<", ">", "=", "!=", "<=", ">=", …) with the goal of unifying behaviour wherever we currently parse them.

This is mostly refactoring, but also adds some useful features to condition parsing. For example ConditionOSRelease= and ConditionKernelVersion= gained the ability to match by shell glob.

It already changes the recently added operator `=$` (to do glob matches) to be `$=` instead. Something we should do before the upcoming release I am sure.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,7 +2,7 @@ systemd System and Service Manager
 
 CHANGES WITH 252 in spe:
 
-        Announcement of Future Feature Removal
+        Announcement of Future Feature Removal:
 
         * Please note that we intend to remove cgroupsv1 support from systemd
           release after EOY 2023. If you run services that make explicit use of
@@ -10,7 +10,20 @@ CHANGES WITH 252 in spe:
           sooner rather than later, if you haven't done so yet. Most of Linux
           userspace has been ported over already.
 
-        New features:
+        Compatibility Breaks:
+
+        * ConditionKernelVersion= checks that use the = or != operator will now
+          do simple string compares (as opposed to version compare – á la
+          stverscmp() — as before, which is still done for the ordering
+          operators <, >, <=, >=). Moreover, if no operator is specified a
+          shell-style glob match is now done. This creates a minor
+          incompatibility compared to older systemd versions, in case the *, ?,
+          [, ], characters have been used in such condition expressions before,
+          as these will now match per shell glob rules instead of
+          literally. Given that kernel version strings typically do not include
+          these characters we expect little breakage through this change.
+
+        New Features:
 
         * systemd-measure is a new helper to precalculate PCR measurements
           to make it easier to set TPM2 policies.


src/shared/compare-operator.c
```diff
@@ -0,0 +1,32 @@
...
+                [COMPARE_FNMATCH_UNEQUAL] = "!=$",
+
+                [COMPARE_LOWER_OR_EQUAL] = "<=",
+                [COMPARE_GREATER_OR_EQUAL] = ">=",
+                [COMPARE_LOWER] = "<",
+                [COMPARE_GREATER] = ">",
+                [COMPARE_EQUAL] = "=",
+                [COMPARE_UNEQUAL] = "!=",
+        };
+
```
bluca, line 36: `assert(s)`

src/shared/meson.build
```diff
@@ -71,6 +71,8 @@ shared_sources = files(
         'clean-ipc.h',
         'clock-util.c',
         'clock-util.h',
+        'compare-operator.c',
```
bluca, line 74: this new module could use a new unit test? just to encode expectations, which are in pid1 Condition and so on, to ensure we don't break anything accidentally

man/systemd.unit.xml
```diff
@@ -1694,10 +1694,13 @@
           <listitem><para>Verify that a specific <literal>key=value</literal> pair is set in the host's
           <citerefentry><refentrytitle>os-release</refentrytitle><manvolnum>5</manvolnum></citerefentry>.</para>
 
-          <para>Other than exact matching with <literal>=</literal>, and <literal>!=</literal>, relative
-          comparisons are supported for versioned parameters (e.g. <literal>VERSION_ID</literal>). The
-          comparator can be one of <literal>&lt;</literal>, <literal>&lt;=</literal>, <literal>=</literal>,
-          <literal>!=</literal>, <literal>&gt;=</literal> and <literal>&gt;</literal>.</para>
+          <para>Other than exact string matching with <literal>=</literal>, and <literal>!=</literal>,
+          relative comparisons are supported for versioned parameters (e.g. <literal>VERSION_ID</literal>),
+          and shell-style wildcard comparisons (<literal>*</literal>, <literal>?</literal>,
```
bluca: I think this minor compat break is fine, but it should go into NEWS

src/shared/condition.c
```diff
@@ -206,7 +206,7 @@ static int condition_test_kernel_version(Condition *c, char **env) {
                         break;
 
                 s = strstrip(word);
-                operator = parse_compare_operator(&s, COMPARE_ALLOW_FNMATCH);
+                operator = parse_compare_operator(&s, COMPARE_ALLOW_FNMATCH|COMPARE_EQUAL_BY_STRING);
```
bluca, line 209: this also should go into NEWS

bluca (approved):

src/shared/meson.build
poettering: the existing condition tests thoroughly test the compare stuff, I think we have roughly 1000% coverage already on this, not sure it's worth adding even more for that ;-)

PR #24499, opened 2022-08-30
meson: Compile with -Werror=format-signedness

medhefgo:


src/libudev/test-libudev.c
```diff
@@ -128,7 +128,7 @@ static void test_device_devnum(struct udev *udev) {
         dev_t devnum = makedev(1, 3);
         _cleanup_(udev_device_unrefp) struct udev_device *device;
 
-        log_info("/* %s, device %d:%d */", __func__, major(devnum), minor(devnum));
+        log_info("/* %s, device %u:%u */", __func__, major(devnum), minor(devnum));
```
yuwata, line 131: Not a matter, but `DEVNUM_FORMAT_STR/VAL()` can be used.

src/login/pam_systemd.c
```diff
@@ -278,7 +278,7 @@ static int get_seat_from_display(const char *display, const char **seat, uint32_
         if (r < 0)
                 return r;
 
-        if (asprintf(&sys_path, "/sys/dev/char/%d:%d", major(display_ctty), minor(display_ctty)) < 0)
+        if (asprintf(&sys_path, "/sys/dev/char/%u:%u", major(display_ctty), minor(display_ctty)) < 0)
```
yuwata, line 281: Also here.

yuwata (approved):
LGTM. The two comments below can be addressed later in another PR.

src/home/homework-mount.c
```diff
@@ -202,7 +202,7 @@ static int make_userns(uid_t stored_uid, uid_t exposed_uid) {
                 return log_oom();
 
         /* Now map the UID we are doing this for to the target UID. */
-        r = strextendf(&text, UID_FMT " " UID_FMT " " UID_FMT "\n", stored_uid, exposed_uid, 1);
+        r = strextendf(&text, UID_FMT " " UID_FMT " " UID_FMT "\n", stored_uid, exposed_uid, 1u);
```
poettering, line 205: maybe better would be `(uid_t) 1`? After all I am not sure POSIX says wether uids are signed or not?

src/journal/journald-server.c
```diff
@@ -1430,11 +1430,11 @@ static int dispatch_sigusr1(sd_event_source *es, const struct signalfd_siginfo *
         assert(s);
 
         if (s->namespace) {
-                log_error("Received SIGUSR1 signal from PID " PID_FMT ", but flushing runtime journals not supported for namespaced instances.", si->ssi_pid);
+                log_error("Received SIGUSR1 signal from PID %u, but flushing runtime journals not supported for namespaced instances.", si->ssi_pid);
```
poettering, line 1433: given this is actually a pid, we should probably cast the variable and use the right type.

src/journal/journald-server.c
```diff
@@ -1430,11 +1430,11 @@ static int dispatch_sigusr1(sd_event_source *es, const struct signalfd_siginfo *
...
                 return 0;
         }
 
-        log_info("Received SIGUSR1 signal from PID " PID_FMT ", as request to flush runtime journal.", si->ssi_pid);
+        log_info("Received SIGUSR1 signal from PID %u, as request to flush runtime journal.", si->ssi_pid);
```
poettering, line 1437: ditto

src/shared/bus-unit-procs.c
```diff
@@ -378,7 +378,7 @@ int unit_show_processes(
                 if (r == -ENOMEM)
                         goto finish;
                 if (r < 0)
-                        log_warning_errno(r, "Invalid process description in GetUnitProcesses reply: cgroup=\"%s\" pid="PID_FMT" command=\"%s\", ignoring: %m",
+                        log_warning_errno(r, "Invalid process description in GetUnitProcesses reply: cgroup=\"%s\" pid=%u command=\"%s\", ignoring: %m",
```
poettering, line 381: as above

src/test/test-path.c
```diff
@@ -94,7 +94,7 @@ static int _check_states(unsigned line,
                          UNIT(path)->id,
                          path_state_to_string(path->state),
                          path_result_to_string(path->result),
-                         end - n);
+                         (int64_t) (end - n));
```
poettering, line 97: this is kinda pointless... the subtraction is done unsigned, which might underrun, and it doesn't help to cast it to signed afterwards.

maybe just used usec_sub()

PR #24504, opened 2022-08-30
hwdb: Add accel orientation quirk for the Aya Neo Air

Maccraft123:
This PR adds quirk for it.
The device is identified by `AYANEO` as system vendor and `AIR`, `AIR Pro`, `AIR Lite`, `AIR Plus` and few more as product name.
To not duplicate entries this PR uses a wildcard at the end,
The Air Plus revision *might* have its display differently orientated as it uses a different, bit bigger, display.

I see two ways of doing this PR:
- Add entries for every product name DMI string, and add Air Plus in future
- Add single entry with a wildcard, and if Air Plus has different rotation than other Air revisions it'll either get split up into multiple entries, or Air Plus will be added so it's "more important" than the wildcard entry

I chose 2nd one, as it's likely that Air Plus will use the same display rotation, if it has different it will have wrong rotation anyway, but I leave the final say to the maintainers.

This PR has been tested on an Aya Neo Air with an `AIR Pro` in its DMI product name

### DIFF
--- a/hwdb.d/60-sensor.hwdb
+++ b/hwdb.d/60-sensor.hwdb
@@ -178,6 +178,10 @@ sensor:modalias:acpi:SMO8500*:dmi:*:svnStandard:pnWCBT1011::*
 # AYANEO
 #########################################
 
+# AYANEO AIR
+sensor:modalias:acpi:BMI0160*:dmi:*:svnAYANEO:pnAIR*:*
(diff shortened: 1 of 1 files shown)


Maccraft123:
Set as draft until it was tested by someone else and it turns out I accidentally pushed different revision.

Now it has been tested by a friend of mine and this version of PR works correctly on his Aya Neo Air unit

PR #24507, opened 2022-08-31
libbpf: Remove use of deprecated APIs

daandemeyer:
I fully expect this to break on a ton of distros due to out of date libbpf, but it's good to have the PR available at least.

### DIFF
--- a/src/shared/bpf-dlopen.c
+++ b/src/shared/bpf-dlopen.c
@@ -3,6 +3,7 @@
 #include "dlfcn-util.h"
 #include "bpf-dlopen.h"
 #include "log.h"
+#include "strv.h"
 
 #if HAVE_LIBBPF
 static void *bpf_dl = NULL;
@@ -45,8 +46,9 @@ static int bpf_print_func(enum libbpf_print_level level, const char *fmt, va_lis
 int dlopen_bpf(void) {
         int r;
 
-        r = dlopen_many_sym_or_warn(
-                        &bpf_dl, "libbpf.so.0", LOG_DEBUG,
+        FOREACH_STRING(f, "libbpf.so.1", "libbpf.so.0") {
+                r = dlopen_many_sym_or_warn(
+                        &bpf_dl, f, LOG_DEBUG,
                         DLSYM_ARG(bpf_link__destroy),
                         DLSYM_ARG(bpf_link__fd),
                         DLSYM_ARG(bpf_map__fd),
@@ -67,6 +69,9 @@ int dlopen_bpf(void) {
                         DLSYM_ARG(libbpf_probe_bpf_prog_type),
                         DLSYM_ARG(libbpf_set_print),
                         DLSYM_ARG(libbpf_get_error));
+                if (r >= 0)
+                        break;
+        }
         if (r < 0)
                 return r;
 


src/shared/bpf-dlopen.c
```diff
@@ -51,8 +51,8 @@ int dlopen_bpf(void) {
                         DLSYM_ARG(bpf_link__fd),
```
martinetd, line 53: (can't comment on lines out of diff range) should also look for libbpf.so.1 just above ^

It doesn't look like there is a way of providing multiple filenames, so will have to call this twice I guess...

src/shared/bpf-dlopen.c
```diff
@@ -51,8 +51,8 @@ int dlopen_bpf(void) {
...
                         DLSYM_ARG(bpf_map__fd),
                         DLSYM_ARG(bpf_map__name),
-                        DLSYM_ARG(bpf_create_map),
-                        DLSYM_ARG(bpf_map__resize),
+                        DLSYM_ARG(bpf_map_create),
```
martinetd, line 56: silly question: can this syntax support e.g. `bpf_map_create@LIBBPF_0.6.0` ? I don't see any '@' symbol in any DLSYM_* helper, but it'd be interesting for future proofing in theory.

src/shared/bpf-dlopen.c
```diff
@@ -61,10 +61,10 @@ int dlopen_bpf(void) {
                         DLSYM_ARG(bpf_object__attach_skeleton),
                         DLSYM_ARG(bpf_object__detach_skeleton),
                         DLSYM_ARG(bpf_object__destroy_skeleton),
-                        DLSYM_ARG(bpf_probe_prog_type),
                         DLSYM_ARG(bpf_program__attach_cgroup),
                         DLSYM_ARG(bpf_program__attach_lsm),
                         DLSYM_ARG(bpf_program__name),
+                        DLSYM_ARG(libbpf_probe_prog_type),
```
martinetd: `bpf_probe_prog_type` was indeed removed in 1.0.... But I do not see libbpf_probe_prog_type.
Did you mean `libbpf_probe_bpf_prog_type` (`@LIBBPF_0.7.0`) ?

I checked and all other symbols are present; here's the list with the version they appear in from nm -D:
```
0000000000024bc0 T bpf_link__destroy@@LIBBPF_0.0.4
0000000000024c50 T bpf_link__fd@@LIBBPF_0.0.8
000000000001d310 T bpf_map__fd@@LIBBPF_0.0.1
000000000001e060 T bpf_map__name@@LIBBPF_0.0.1
000000000000c4c0 T bpf_map_create@@LIBBPF_0.6.0
000000000001b080 T bpf_map__set_max_entries@@LIBBPF_0.1.0
000000000000d020 T bpf_map_update_elem@@LIBBPF_0.0.1
000000000000d2f0 T bpf_map_delete_elem@@LIBBPF_0.0.1
000000000001f6e0 T bpf_map__set_inner_map_fd@@LIBBPF_0.0.1
000000000002c690 T bpf_object__open_skeleton@@LIBBPF_0.0.7
000000000002cb40 T bpf_object__load_skeleton@@LIBBPF_0.0.7
000000000002ccf0 T bpf_object__attach_skeleton@@LIBBPF_0.0.7
000000000002cdc0 T bpf_object__detach_skeleton@@LIBBPF_0.0.7
000000000002ce10 T bpf_object__destroy_skeleton@@LIBBPF_0.0.7
0000000000027b20 T bpf_program__attach_cgroup@@LIBBPF_0.0.8
0000000000027af0 T bpf_program__attach_lsm@@LIBBPF_0.0.8
000000000001c680 T bpf_program__name@@LIBBPF_0.0.7
000000000002ef50 T libbpf_probe_bpf_prog_type@@LIBBPF_0.7.0
0000000000017a80 T libbpf_set_print@@LIBBPF_0.0.1
000000000001fe60 T libbpf_get_error@@LIBBPF_0.0.1
```

So there are two that'd need conditionals really: `bpf_map_create` and `libbpf_probe_bpf_prog_type` -- it's not THAT bad if you're ok with v0.1.0 (cutting at 2 years, it was released Aug 2020, 0.0.8 you were already using symbols from dates April 2020 so it's not too different)

Well, /shrug.

bluca:
Yeah this is really bad. We'll need to ifdef all of this. They should have put in a backward compatibility shim layer instead of pushing all the pain and friction to the users. Hopefully there is strong feedback going back to the maintainers to ensure this never happens again.

martinetd (commented):
Thanks for the brush up -- a few comments inline.

src/shared/bpf-dlopen.c
daandemeyer: My bad I got the symbol name wrong

src/shared/bpf-dlopen.c
poettering: As mentioned in the other PR: use separate dlopen() + dlsym_many_or_warn() instead of dlopen_many_sym_or_warn()

src/shared/bpf-dlopen.c
poettering: The macro generates both a variable name and a string from the argument. C doesn't really allows splitting strings in macros, hence what we could do is add `DLVSYM_ARG(symbol, version)` that encodes both, and uses one for the variable name, but passes both to the function call...

I am not entirely sure it's worth it though: the fact that libppf exposes versioned symbols doesn't really maean we need to use them, if they don't buy us anything. As I understand they broke compat quite massively between 0.6.0 and 1.0, so to me it appears the versioned symbols are not going to buy is anything...

poettering:
As mentioned in #24511: would love to see some analysis what the impact of this would be, regarding compat with downstreams and CIs...

martinetd:
> As mentioned in https://github.com/systemd/systemd/pull/24511: would love to see some analysis what the impact of this would be, regarding compat with downstreams and CIs...

Choices I can see:
 - my other PR works with 0.1.0+, but requires some compat cruft
 - require 0.7.0+ -- that only requires checking both filenames and we can share all the code/symbols, but it's only half a year old.
 - require 1.0.0 -- that's not realistic given it just got released a couple of weeks ago.

https://repology.org/project/libbpf/versions is not really optimistic, event rhel9 only ships 0.5.0; fedora35 0.4.0 (fedora36 is ok with 0.7.0), debian stable 0.3.0...
if you only care about bleeding edges distros we can have something fairly simple though.

bluca:
> if you only care about bleeding edges distros we can have something fairly simple though.

We keep compat with LTSes, see the readme for details.

poettering:
> We keep compat with LTSes, see the readme for details.

Well, I'd take that with a grain of salt. For optional, very new leaf features I think we could be more aggressive. Question is if libbpf stuff qualifies as that, or if it's already so widely use we can't be so aggressive. 

Or in other words, did a slow conservative distro like debian already enable libbpf stuff in systemd builds?

bluca:
> > We keep compat with LTSes, see the readme for details.
> 
> Well, I'd take that with a grain of salt. For optional, very new leaf features I think we could be more aggressive. Question is if libbpf stuff qualifies as that, or if it's already so widely use we can't be so aggressive.
> 
> Or in other words, did a slow conservative distro like debian already enable libbpf stuff in systemd builds?

It's not just about what ships, but it's also important to avoid adding friction and breaking developers who are actually running these systems. At this point there are several security-relevant features that use bpf (and perhaps it would have been better to avoid relying on that, given how unstable the ecosystem appears to be, but hindsight is 20/20), so not sure we can get away with saying that it's optional?

DaanDeMeyer:
Updated to bump the minimum libbpf version to 0.7 because we already depend on the skeleton APIs from libbpf 0.7. Having 0.7 as the minimum version makes sure all the new APIs that replace the deprecated APIs are available, so we can just switch to the new APIs and try to dlopen() both libbpf.so.0 and libbpf.so.1 to support both without ifdeffery.

bluca:
I'll double check later today

DaanDeMeyer:
> hmm, I could just build with libbpf 0.3 from debian? didn't try to run, but feature libbpf is enabled as per what meson reports. I also tried enabling bpf-framework as well and that also compiled fine.

On ubuntu I can also build with libbpf 0.5. If I inspect their libbpf so, I can see it includes symbols from libbpf 0.7. I guess there's backporting going? But that shouldn't affect our minimum version which should be determined by the minimum upstream version that supports all the symbols we use.

DaanDeMeyer:
Ah shit nvm, I just can't read, it's 0.0.7, not 0.7.0

martinetd:
> I'll double check later today

thanks!

> On ubuntu I can also build with libbpf 0.5. If I inspect their libbpf so, I can see it includes symbols from libbpf 0.7. I guess there's backporting going? But that shouldn't affect our minimum version which should be determined by the minimum upstream version that supports all the symbols we use.

No such backported symbol on debian:
```
# nm -D /usr/lib/x86_64-linux-gnu/libbpf.so | grep bpf_ | grep map | grep create
00000000000091d0 T bpf_create_map@@LIBBPF_0.0.1
0000000000009380 T bpf_create_map_in_map@@LIBBPF_0.0.1
00000000000092a0 T bpf_create_map_in_map_node@@LIBBPF_0.0.1
0000000000009240 T bpf_create_map_name@@LIBBPF_0.0.1
0000000000009150 T bpf_create_map_node@@LIBBPF_0.0.1
0000000000009060 T bpf_create_map_xattr@@LIBBPF_0.0.1
```

(note LIBBPF_0.0.7 and friends aren't libbpf 0.7; these would be in LIBBPF_0.7.0; I just went to https://packages.ubuntu.com/jammy/libbpf0 and downloaded the source packages there and see no such compat thing)

poettering:
so drop the please-review label again?

DaanDeMeyer:
Well, bumping to 0.7.0 still allows us to support libbpf v1.0, since all the replacements for the deprecated symbols are available in this version. The discussion would just be about whether we want to bump our minimum version to 0.7 or whether we want to keep our minimum version to 0.2 as before and use ifdeffery instead as is done in #24511.

I checked and none Debian, Ubuntu, Fedora, CentOS, Arch and opensuse are building systemd with libbpf support enabled. Latest stable/LTS releases of CentOS, Debian, Ubuntu and Opensuse ship a version of libbpf that's older than 0.7. So anyone backporting systemd to any of these distros with the bpf-framework option enabled would run into build issues if this PR is merged. The questions is whether we want to support those scenarios. I also wonder whether this is purely hypothetical or if someone is actually backporting systemd to one of these distros and is enabling the bpf-framework option.

poettering:
I think @bluca is using an antedeluvian Debian LTS from 1995 or so ;-) and he wants to be able to build systemd with the packages from that distro I think.

But as mentioned I have no problem with being more agressive on this here.

bluca:
> The questions is whether we want to support those scenarios. I also wonder whether this is purely hypothetical or if someone is actually backporting systemd to one of these distros and is enabling the bpf-framework option.

Given we don't have any telemetry, it can't really be proven - but it is a supported scenario, and on multiple officially supported distributions versions. Well, there's at least one user that we know of ;-)

The main point is that breaking compatibility always comes with a cost, the question is who pays for it. The libbpf maintainers obviously didn't want to pay for it or didn't care, and pushed it down to their users instead. I don't think we should do the same, as that's really bad practice, especially when it can be avoided. And if that means dealing with ifdeffery for a couple of years, then it just means that's the cost of dealing with libbpf, and next time we need to be more careful about it, until it's really stable and ready for production.

martinetd:
Since direction seems to be going that way, I'll test this with nixos when I can find time :)

keszybz:
> I checked and none Debian, Ubuntu, Fedora, CentOS, Arch and opensuse are building systemd with libbpf support enabled.

Hmm, this sentence has grammatical issues, so maybe I'm misreading it…
At least Fedora builds against libbpf.

> Well, bumping to 0.7.0 still allows us to support libbpf v1.0, since all the replacements for the deprecated symbols are available in this version. The discussion would just be about whether we want to bump our minimum version to 0.7

I'd vote for that. People who have very libbpf < 0.7 will not get this functionality, but I think this is reasonable. Systemd will still compile and work.

src/core/bpf-lsm.c
```diff
@@ -65,14 +66,14 @@ static int prepare_restrict_fs_bpf(struct restrict_fs_bpf **ret_obj) {
...
-        r = sym_bpf_map__resize(obj->maps.cgroup_hash, CGROUP_HASH_SIZE_MAX);
+        r = sym_bpf_map__set_max_entries(obj->maps.cgroup_hash, CGROUP_HASH_SIZE_MAX);
         assert(r <= 0);
         if (r < 0)
                 return log_error_errno(r, "bpf-lsm: Failed to resize BPF map '%s': %m",
                                        sym_bpf_map__name(obj->maps.cgroup_hash));
 
         /* Dummy map to satisfy the verifier */
-        inner_map_fd = sym_bpf_create_map(BPF_MAP_TYPE_HASH, sizeof(uint32_t), sizeof(uint32_t), 128, 0);
+        inner_map_fd = sym_bpf_map_create(BPF_MAP_TYPE_HASH, NULL, sizeof(uint32_t), sizeof(uint32_t), 128U, &opts);
```
anakryiko: given you don't set any options, you can just pass NULL here and you won't need LIBBPF_OPTS() declaration above

src/core/bpf-lsm.c
```diff
@@ -193,6 +194,7 @@ int lsm_bpf_unit_restrict_filesystems(Unit *u, const Set *filesystems, bool allo
         uint32_t dummy_value = 1, zero = 0;
         const char *fs;
         const statfs_f_type_t *magic;
+        LIBBPF_OPTS(bpf_map_create_opts, opts, .map_flags = 0);
```
anakryiko: same, NULL for opts is always support to mean "no non-default options"

bluca:
> > Well, bumping to 0.7.0 still allows us to support libbpf v1.0, since all the replacements for the deprecated symbols are available in this version. The discussion would just be about whether we want to bump our minimum version to 0.7
> 
> I'd vote for that. People who have very libbpf < 0.7 will not get this functionality, but I think this is reasonable. Systemd will still compile and work.

It's not that simple, we first need to get that version in the supported LTSes, which won't be simple. Which is yet another thing that the BPF maintainers should have done first.

keszybz:
>  I checked on pkgs.org and it wasn't listed there

pkg.org doesn't seem to know about weak dependencies (`Recommends`,`Suggests`,`Enhances`,`Supplements`), and doesn't seem to understand rich dependencies (`Requires:(foo if bar)`, etc.). That's the last 5 years of development of rpm ;)

> It's not that simple, we first need to get that version in the supported LTSes, which won't be simple. Which is yet another thing that the BPF maintainers should have done first.

LTS releases cannot be expected to have the latest versions of everything. Thus, **if** you're compiling latest systemd on something very old, and **if** you need bpf support, you might need to also provide a non-ancient version of libbpf. I think that's fair, since those are two very big "ifs".

(I agree that libbpf maintianers shouldn't have removed symbols and bumped the SOVERSION. That's just such an incredibly annoying and unnecessary thing to do. But even if the new libbpf versions were backwards compatible, we might still depend on new features, and wrt. to compiling systemd on very old distros, the result would be similar.)

keszybz:
```
Retrieving package libz-ng-compat1-2.0.6-4.2.x86_64 (21/225),  72.8 KiB (126.5 KiB unpacked)
Media source 'http://download.opensuse.org/tumbleweed/repo/oss/' does not contain the desired medium
Abort, retry, ignore? [a/r/i/...? shows all options] (a): a
```
Now I know why CI is so hard. The poor human who sits there and presses 'a'/'r/'i' must be so tired.

keszybz:
Hmm, the distinction between 0.0.7 and 0.7.0 was discussed above, but the commit message of the first commit is confused. In fact, the skeleton framework that we use was added in 0.0.7.

Before this PR:
```console
$ objdump -T /usr/lib64/libbpf.so.0.8.0|sed 's/^.*LIBBPF/LIBBPF/'|rg '(bpf_link__destroy|bpf_link__fd|bpf_map__fd|bpf_map__name|bpf_create_map|bpf_map__resize|bpf_map_update_elem|bpf_map_delete_elem|bpf_map__set_inner_map_fd|bpf_object__open_skeleton|bpf_object__load_skeleton|bpf_object__attach_skeleton|bpf_object__detach_skeleton|bpf_object__destroy_skeleton|bpf_probe_prog_type|bpf_program__attach_cgroup|bpf_program__attach_lsm|bpf_program__name|libbpf_set_print|libbpf_get_error)$'|sort
LIBBPF_0.0.1 bpf_create_map
LIBBPF_0.0.1 bpf_map_delete_elem
LIBBPF_0.0.1 bpf_map__fd
LIBBPF_0.0.1 bpf_map__name
LIBBPF_0.0.1 bpf_map__set_inner_map_fd
LIBBPF_0.0.1 bpf_map_update_elem
LIBBPF_0.0.1 libbpf_get_error
LIBBPF_0.0.1 libbpf_set_print
LIBBPF_0.0.2 bpf_map__resize
LIBBPF_0.0.2 bpf_probe_prog_type
LIBBPF_0.0.4 bpf_link__destroy
LIBBPF_0.0.7 bpf_object__attach_skeleton
LIBBPF_0.0.7 bpf_object__destroy_skeleton
LIBBPF_0.0.7 bpf_object__detach_skeleton
LIBBPF_0.0.7 bpf_object__load_skeleton
LIBBPF_0.0.7 bpf_object__open_skeleton
LIBBPF_0.0.7 bpf_program__name
LIBBPF_0.0.8 bpf_link__fd
LIBBPF_0.0.8 bpf_program__attach_cgroup
LIBBPF_0.0.8 bpf_program__attach_lsm
```
(This regexp is the list of symbols we import in `src/shared/bpf-dlopen.c`.)

After this PR:
```console
$ objdump -T /usr/lib64/libbpf.so.0.8.0|sed 's/^.*LIBBPF/LIBBPF/'|rg '(bpf_link__destroy|bpf_link__fd|bpf_map__fd|bpf_map__name|bpf_map_create|bpf_map__set_max_entries|bpf_map_update_elem|bpf_map_delete_elem|bpf_map__set_inner_map_fd|bpf_object__open_skeleton|bpf_object__load_skeleton|bpf_object__attach_skeleton|bpf_object__detach_skeleton|bpf_object__destroy_skeleton|bpf_program__attach_cgroup|bpf_program__attach_lsm|bpf_program__name|libbpf_probe_bpf_prog_type|libbpf_set_print|libbpf_get_error)$'|sort
LIBBPF_0.0.1 bpf_map_delete_elem
LIBBPF_0.0.1 bpf_map__fd
LIBBPF_0.0.1 bpf_map__name
LIBBPF_0.0.1 bpf_map__set_inner_map_fd
LIBBPF_0.0.1 bpf_map_update_elem
LIBBPF_0.0.1 libbpf_get_error
LIBBPF_0.0.1 libbpf_set_print
LIBBPF_0.0.4 bpf_link__destroy
LIBBPF_0.0.7 bpf_object__attach_skeleton
LIBBPF_0.0.7 bpf_object__destroy_skeleton
LIBBPF_0.0.7 bpf_object__detach_skeleton
LIBBPF_0.0.7 bpf_object__load_skeleton
LIBBPF_0.0.7 bpf_object__open_skeleton
LIBBPF_0.0.7 bpf_program__name
LIBBPF_0.0.8 bpf_link__fd
LIBBPF_0.0.8 bpf_program__attach_cgroup
LIBBPF_0.0.8 bpf_program__attach_lsm
LIBBPF_0.1.0 bpf_map__set_max_entries
LIBBPF_0.6.0 bpf_map_create
LIBBPF_0.7.0 libbpf_probe_bpf_prog_type
```
Please fix the commit message.

--

But this changes the considerations here a lot. My understanding was that our code wouldn't in fact work with old libbpf, but just fail during dlopen() and pretend to work during compilation. But it seems we were compatible with old libbpf versions, and this PR bumps the required version. I'm still leaning towards merging this, but it's less straightforward.

How much work would it be to add the iffdeffery to support older versions?

DaanDeMeyer:
See https://github.com/systemd/systemd/pull/24511 for the ifdeffery version

bluca:
> See #24511 for the ifdeffery version

Yes, that's the version we need - either that, or wait for 0.7 to make it into the supported releases. We can't break backward compatibility - this used to work, now it would be broken, and that's bad. Or more likely, we need both - and then we can remove the additional shim once we manage to get 0.7 where it's missing.

martinetd:
Note the ifdeffery version just got much simpler since we removed the bpf option type (passing NULL as parameter) - we can remove the struct and #ifndef #define from the compat header, leaving just the two static inlines to use one function or the other.
(The types all existed on older libbpf so we could still compile with old libbpf and have a newer libbpf at runtime or the other way around, too.)

Issue #24502, opened 2022-08-30
systemd can output invalid ANSI escape sequences to the serial console, breaking tmux


jlebon:
### systemd version the issue has been seen with

250, 251

### Used distribution

Fedora 36, 37, 38

### Linux kernel version used

5.18.16-200.fc36.x86_64 - 6.0.0-0.rc2.20220826git4c612826bec1.22.fc38.x86_64

### CPU architectures issue was seen on

x86_64

### Component

systemd

### Expected behaviour you didn't see

When booting Fedora CoreOS (stable or rawhide) via QEMU and connecting to the serial console, systemd writes valid ANSI escape sequences to the console when styling and colouring output.

### Unexpected behaviour you saw

When booting Fedora CoreOS (stable or rawhide) via QEMU and connecting to the serial console, systemd writes *invalid* ANSI escape sequences to the console when styling and colouring output. This can breaks at least tmux, which stops printing output following a particular bad sequence (clearly, there's a tmux bug to be filed there to strengthen its parser, but I think we should fix the systemd side too). GNOME Terminal does not break on this.

### Steps to reproduce the problem

Download and verify the Fedora CoreOS image from [getfedora.org](https://getfedora.org/en/coreos/download?tab=metal_virtualized&stream=stable&arch=x86_64). From within tmux, boot it up using QEMU while also `tee`ing its output to a file:

```
$ curl -L --remote-name-all https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/36.20220806.3.0/x86_64/fedora-coreos-36.20220806.3.0-qemu.x86_64.qcow2.xz{,.sig}
$ gpg --verify fedora-coreos-36.20220806.3.0-qemu.x86_64.qcow2.xz.sig fedora-coreos-36.20220806.3.0-qemu.x86_64.qcow2.xz
$ unxz fedora-coreos-36.20220806.3.0-qemu.x86_64.qcow2.xz
$ qemu-system-x86_64 -machine accel=kvm -cpu host -m 1G -smp 1 -nographic -drive if=virtio,file=fedora-coreos-36.20220806.3.0-qemu.x86_64.qcow2,format=qcow2 -snapshot -display none | tee console.log
...
[  OK  ] Listening on systemd-userd…0m - User Database Manager Socket.
[    6.158164] systemd[1]: Mounting dev-hugepages.mount - Huge Pages File System...
         Mounting dev-hugepages.mount - Huge Pages File System...
[...]

PR #24509, opened 2022-08-31
string-util: Do not cut off ANSI escape sequences

medhefgo:
Cutting off in the middle may leave the terminal in a bad state,
breaking further output.

Also, use memcpy_safe as we may end up copying zero bytes when
ellipsizing at the start/end of a string.

Fixes: #24502

### DIFF
--- a/src/basic/string-util.c
+++ b/src/basic/string-util.c
@@ -288,6 +288,36 @@ static int write_ellipsis(char *buf, bool unicode) {
         return 3;
 }
 
+static size_t skip_ansi_sequence(const char *s, size_t x, bool left) {
+        size_t i = x;
+
+        /* Find ASCII ESC. */
+        while (i > 0 && !IN_SET(s[i], 0, 0x1B))
+                i--;
+        if (s[i] != 0x1B)
+                return x;
+
+        size_t start = i++;
+
+        if (s[i] == 0x5B) { /* CSI sequence. */
+                i++;
+                while (s[i] && s[i] >= 0x30 && s[i] <= 0x3F) /* Parameter bytes. */
+                        i++;
+                while (s[i] && s[i] >= 0x20 && s[i] <= 0x2F) /* Intermediate bytes. */
+                        i++;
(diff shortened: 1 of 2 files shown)


medhefgo:
@jlebon Can you try this PR to see if it works for you?

poettering:
I figure the status output should always issue `\x1B[0m` (aka `ANSI_NORMAL`) after all strings that it potentially ellipsizes, to ensure that we'll reset the output attributes if the reset was dropped due to the ellipsis.

jlebon:
I guess if we want this low-level utility function to be as correct as possible for any input strings, we'd want to pick up and re-add every sequence contained in the ellipsized text. That would naturally ensure that e.g. `ANSI_NORMAL` is still emitted.

keszybz:
> I figure the status output should always issue `\x1B[0m` (aka `ANSI_NORMAL`) after all strings that it potentially ellipsizes, to ensure that we'll reset the output attributes if the reset was dropped due to the ellipsis.

Or at least do that if we dropped `ANSI_NORMAL` during ellipsization.

> I guess if we want this low-level utility function to be as correct as possible for any input strings, we'd want to pick up and re-add every sequence contained in the ellipsized text.

That might actually be easier to implement… Just pass through all escape sequences, that DTRT without doing anything special.

keszybz:
I'll try to rework this along the suggested lines.

keszybz:
> I'll try to rework this along the suggested lines.

Pfff, it turns out that this is a lot of work. IIUC, it's hard to figure out where ANSI sequences end in a generic way, so we'd need to implement a parser for specific types.

We definitely need to fix this somehow, but we don't need to block -rc1.

dustymabe:
> We definitely need to fix this somehow, but we don't need to block -rc1.

Thank you for continuing to consider this important!

dustymabe:
Hi. Do we still want to fix this issue?

poettering:
Let's close this one, and continue in #28146 which is based on this PR if I understand things correctly.

PR #24511, opened 2022-08-31
libbpf: Add libbpf 1.0.0 compat

martinetd:
I wanted to create a PR on top of #24507 but I can't seem to target https://github.com/DaanDeMeyer/systemd/tree/libbppf-1.0 ; github is such a mess for multiplayer contributions...

Anyway, rant aside this:
 - bases itself on top of #24507 for 1.0 compat
 - adds back old symbols for anything newer than 0.1.0 (Aug 2020); the logic is as simple as possible: if we're looking at libbpf.so.1 we have the new symbols and we use the new code, if we're looking at libbpf.so.0 it still has the old symbols even if the new ones might be available and we use the old code.
This WILL break systems with a version < 0.1.0 for bpf_map_create, but that one is old enough in my opinion -- if you want to play safe we can use the old one and add another compat helper.

I've successfully built this with libbpf 0.4.0, 0.7.0 and 1.0.0 -- didn't try running any though, but assuming the symbols that aren't found are left at NULL it should work. I'll assume tests will catch any breakage on older lib, and I can try 1.0.0 next weekend if nobody beats me to it. I need to look up how to actually test this first.

It should run fine with a different version of libbpf than the one it was compiled with for bonus points, as long as it can find either .so.

How does that sound?

cc @DaanDeMeyer @bluca @vcunat

### DIFF
--- a/src/shared/bpf-compat.h
+++ b/src/shared/bpf-compat.h
@@ -1,0 +1,54 @@
+/* SPDX-License-Identifier: LGPL-2.1-or-later */
+#pragma once
+
+/* libbpf has been moving quickly.
+ * They added new symbols in the 0.x versions and shortly after removed
+ * deprecated symbols in 1.0.
+ * We only need bpf_map_create and libbpf_probe_bpf_prog_type so we work
+ * around the incompatibility here by:
+ *  - declaring both symbols, and looking for either depending on the libbpf
+ *    so version we found
+ *  - having helpers that automatically use the appropriate version behind the
+ *    new API for easy cleanup later
+ *
+ * The advantage of doing this instead of only looking for the symbols declared at
+ * compile time is that we can then load either the old or the new symbols at runtime
+ * regardless of the version we were compiled with */
+
+
+/* declare the struct for libbpf <= 0.6.0 -- it causes no harm on newer versions */
+struct bpf_map_create_opts;
+
+/* new symbols available from 0.7.0.
+ * We need the symbols here:
+ *  - after bpf_map_create_opts struct has been defined for older libbpf
+ *  - before the compat static inline helpers that use them.
+ * When removing this file move these back to bpf-dlopen.h */
+extern int (*sym_bpf_map_create)(enum bpf_map_type,  const char *, __u32, __u32, __u32, const struct bpf_map_create_opts *);
+extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
+
+/* compat symbols removed in libbpf 1.0 */
+extern int (*sym_bpf_create_map)(enum bpf_map_type, int key_size, int value_size, int max_entries, __u32 map_flags);
+extern bool (*sym_bpf_probe_prog_type)(enum bpf_prog_type, __u32);
+
+/* helpers to use the available variant behind new API */
+static inline int compat_bpf_map_create(enum bpf_map_type map_type,
+                const char *map_name,
+                __u32 key_size,
+                __u32 value_size,
+                __u32 max_entries,
+                const struct bpf_map_create_opts *opts) {
+        if (sym_bpf_map_create)
+                return sym_bpf_map_create(map_type, map_name, key_size,
+                                          value_size, max_entries, opts);
+
+        return sym_bpf_create_map(map_type, key_size, value_size, max_entries,
+                                  0 /* opts->map_flags, but opts is always NULL for us so skip build dependency on the type */);
+}
+
+static inline int compat_libbpf_probe_bpf_prog_type(enum bpf_prog_type prog_type, const void *opts) {
+        if (sym_libbpf_probe_bpf_prog_type)
+                return sym_libbpf_probe_bpf_prog_type(prog_type, opts);
+
+        return sym_bpf_probe_prog_type(prog_type, 0);
+}
--- a/src/shared/bpf-dlopen.c
+++ b/src/shared/bpf-dlopen.c
@@ -6,8 +6,6 @@
 #include "strv.h"
 
 #if HAVE_LIBBPF
-static void *bpf_dl = NULL;
-
 struct bpf_link* (*sym_bpf_program__attach_cgroup)(struct bpf_program *, int);
 struct bpf_link* (*sym_bpf_program__attach_lsm)(struct bpf_program *);
 int (*sym_bpf_link__fd)(const struct bpf_link *);
@@ -29,6 +27,10 @@ const char* (*sym_bpf_program__name)(const struct bpf_program *);
 libbpf_print_fn_t (*sym_libbpf_set_print)(libbpf_print_fn_t);
 long (*sym_libbpf_get_error)(const void *);
 
+/* compat symbols removed in libbpf 1.0 */
+int (*sym_bpf_create_map)(enum bpf_map_type,  int key_size, int value_size, int max_entries, __u32 map_flags);
+bool (*sym_bpf_probe_prog_type)(enum bpf_prog_type, __u32);
+
 _printf_(2,0)
 static int bpf_print_func(enum libbpf_print_level level, const char *fmt, va_list ap) {
 #if !LOG_TRACE
@@ -44,16 +46,37 @@ static int bpf_print_func(enum libbpf_print_level level, const char *fmt, va_lis
 }
 
 int dlopen_bpf(void) {
+        void *dl;
         int r;
 
-        FOREACH_STRING(f, "libbpf.so.1", "libbpf.so.0") {
-                r = dlopen_many_sym_or_warn(
-                        &bpf_dl, f, LOG_DEBUG,
+        dl = dlopen("libbpf.so.1", RTLD_LAZY);
+        if (!dl) {
+                /* libbpf < 1.0.0 (we rely on 0.1.0+) provide most symbols we care about, but
+                 * unfortunately not all until 0.7.0. See bpf-compat.h for more details.
+                 * Once we consider we can assume 0.7+ is present we can just use the same symbol
+                 * list for both files, and when we assume 1.0+ is present we can remove this dlopen */
+                dl = dlopen("libbpf.so.0", RTLD_LAZY);
+                if (!dl)
+                        return log_debug_errno(SYNTHETIC_ERRNO(EOPNOTSUPP),
+                                               "neither libbpf.so.1 nor libbpf.so.0 are installed: %s", dlerror());
+
+                /* symbols deprecated in 1.0 we use as compat */
+                r = dlsym_many_or_warn(dl, LOG_DEBUG,
+                                DLSYM_ARG(bpf_create_map),
+                                DLSYM_ARG(bpf_probe_prog_type));
+        } else {
+                /* symbols available from 0.7.0 */
+                r = dlsym_many_or_warn(dl, LOG_DEBUG,
+                                DLSYM_ARG(bpf_map_create),
+                                DLSYM_ARG(libbpf_probe_bpf_prog_type));
+        }
+
+        r = dlsym_many_or_warn(
+                        dl, LOG_DEBUG,
                         DLSYM_ARG(bpf_link__destroy),
                         DLSYM_ARG(bpf_link__fd),
                         DLSYM_ARG(bpf_map__fd),
                         DLSYM_ARG(bpf_map__name),
-                        DLSYM_ARG(bpf_map_create),
                         DLSYM_ARG(bpf_map__set_max_entries),
                         DLSYM_ARG(bpf_map_update_elem),
                         DLSYM_ARG(bpf_map_delete_elem),
@@ -66,12 +89,8 @@ int dlopen_bpf(void) {
                         DLSYM_ARG(bpf_program__attach_cgroup),
                         DLSYM_ARG(bpf_program__attach_lsm),
                         DLSYM_ARG(bpf_program__name),
-                        DLSYM_ARG(libbpf_probe_bpf_prog_type),
                         DLSYM_ARG(libbpf_set_print),
                         DLSYM_ARG(libbpf_get_error));
-                if (r >= 0)
-                        break;
-        }
         if (r < 0)
                 return r;
 
--- a/src/shared/bpf-dlopen.h
+++ b/src/shared/bpf-dlopen.h
@@ -6,13 +6,14 @@
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
 
+#include "bpf-compat.h"
+
 extern struct bpf_link* (*sym_bpf_program__attach_cgroup)(struct bpf_program *, int);
 extern struct bpf_link* (*sym_bpf_program__attach_lsm)(struct bpf_program *);
 extern int (*sym_bpf_link__fd)(const struct bpf_link *);
 extern int (*sym_bpf_link__destroy)(struct bpf_link *);
 extern int (*sym_bpf_map__fd)(const struct bpf_map *);
 extern const char* (*sym_bpf_map__name)(const struct bpf_map *);
-extern int (*sym_bpf_map_create)(enum bpf_map_type,  const char *, __u32, __u32, __u32, const struct bpf_map_create_opts *);
 extern int (*sym_bpf_map__set_max_entries)(struct bpf_map *, __u32);
 extern int (*sym_bpf_map_update_elem)(int, const void *, const void *, __u64);
 extern int (*sym_bpf_map_delete_elem)(int, const void *);
@@ -25,7 +26,6 @@ extern int (*sym_bpf_object__attach_skeleton)(struct bpf_object_skeleton *);
 extern void (*sym_bpf_object__detach_skeleton)(struct bpf_object_skeleton *);
 extern void (*sym_bpf_object__destroy_skeleton)(struct bpf_object_skeleton *);
 extern const char* (*sym_bpf_program__name)(const struct bpf_program *);
-extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
 extern libbpf_print_fn_t (*sym_libbpf_set_print)(libbpf_print_fn_t);
 extern long (*sym_libbpf_get_error)(const void *);
 
--- a/meson.build
+++ b/meson.build
@@ -1050,7 +1050,7 @@ want_bpf_framework = get_option('bpf-framework')
 bpf_compiler = get_option('bpf-compiler')
 bpf_framework_required = want_bpf_framework == 'true'
 
-libbpf_version_requirement = '>= 0.7.0'
+libbpf_version_requirement = '>= 0.1.0'
 if bpf_compiler == 'gcc'
         libbpf_version_requirement = '>= 1.0.0'
 endif
--- a/src/core/bpf-lsm.c
+++ b/src/core/bpf-lsm.c
@@ -72,7 +72,7 @@ static int prepare_restrict_fs_bpf(struct restrict_fs_bpf **ret_obj) {
                                        sym_bpf_map__name(obj->maps.cgroup_hash));
 
         /* Dummy map to satisfy the verifier */
-        inner_map_fd = sym_bpf_map_create(BPF_MAP_TYPE_HASH, NULL, sizeof(uint32_t), sizeof(uint32_t), 128U, NULL);
+        inner_map_fd = compat_bpf_map_create(BPF_MAP_TYPE_HASH, NULL, sizeof(uint32_t), sizeof(uint32_t), 128U, NULL);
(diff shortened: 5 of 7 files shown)


src/shared/bpf-dlopen.h
```diff
@@ -6,14 +6,34 @@
 #include <bpf/bpf.h>
 #include <bpf/libbpf.h>
 
+/* define required struct to allow newer API with older bpf.h
+ * remove together with pre-1.0 compat helpers below */
+#ifndef bpf_map_create_opts__last_field
+struct bpf_map_create_opts {
```
bluca: It makes no sense whatsoever to have this stuff downstream. This needs to be provided upstream.

martinetd: This is upstream, from libbpf 0.6.0.
We need something that's called bpf_map_create_opts and has sz and map_flags fields to please the `LIBBPF_OPTS` call the previous patch added and the function definition just below.

Additionally it's magic and if the struct is close enough, using the libbpf 1.0.0 symbol when compiled with a very old libbpf will also just work so I went all the way with copying it.
Honestly since you're using dlopen it doesn't make much sense not to do that, you're not checking anything.
Either have something that gets proper functions automatically from the headers and I don't care about cross-version compatibility and this can be removed, or admit this system is a mess that doesn't provide any guarantee at which point I don't see the problem with this.

Anyway, I can just say I don't care -- I'll fix the LIBBPF_OPTS macro that I missed to pass checks and won't reply here again.

bluca: it's the same for all other "compat symbols removed in libbpf 1.0" - the point is that these should have been provided by upstream, it's absurd to force every single user to define them again and again and again

src/shared/bpf-dlopen.c
```diff
@@ -67,8 +72,32 @@ int dlopen_bpf(void) {
...
+                                DLSYM_ARG(bpf_object__load_skeleton),
+                                DLSYM_ARG(bpf_object__attach_skeleton),
+                                DLSYM_ARG(bpf_object__detach_skeleton),
+                                DLSYM_ARG(bpf_object__destroy_skeleton),
+                                DLSYM_ARG(bpf_probe_prog_type),
+                                DLSYM_ARG(bpf_program__attach_cgroup),
+                                DLSYM_ARG(bpf_program__attach_lsm),
+                                DLSYM_ARG(bpf_program__name),
+                                DLSYM_ARG(libbpf_set_print),
+                                DLSYM_ARG(libbpf_get_error));
```
poettering: please do not duplicate the whole shebang here. 

Instead, use dlopen() to open one or the other .so, and then use dlsym_many_or_warn() instead of dlopen_many_sym_or_warn() to resolve the symbols.

(I presume the list of symbols is exactly the same?)

src/shared/bpf-dlopen.h
poettering: i guess they knowingle broke compat there in libbpf1?

This stuff should move to libbpf-compat.h or so, and properly ifdeffed out

src/shared/bpf-dlopen.h
```diff
@@ -29,6 +49,38 @@ extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
...
+                return sym_libbpf_probe_bpf_prog_type(prog_type, opts);
+        }
+        return sym_bpf_probe_prog_type(prog_type, 0);
+}
+#ifndef LIBBPF_OPTS
+#define LIBBPF_OPTS DECLARE_LIBBPF_OPTS
+#endif
+
+
+
```
poettering: spurious empty lines

src/shared/bpf-dlopen.h
```diff
@@ -29,6 +49,38 @@ extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
 extern libbpf_print_fn_t (*sym_libbpf_set_print)(libbpf_print_fn_t);
 extern long (*sym_libbpf_get_error)(const void *);
 
+/* compat symbols removed in libbpf 1.0 */
+extern int (*sym_bpf_create_map)(enum bpf_map_type, int key_size, int value_size, int max_entries, __u32 map_flags);
+extern bool (*sym_bpf_probe_prog_type)(enum bpf_prog_type, __u32);
```
poettering: so, do the replacement symbols exist in both old *and* new libbpf? if so, why would we bother with the old symbols?

src/shared/bpf-dlopen.h
```diff
@@ -29,6 +49,38 @@ extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
...
+
+
```
poettering: spurious double empty line

src/shared/bpf-dlopen.h
```diff
@@ -29,6 +49,38 @@ extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
...
+/* helpers to use the right function */
+static inline int compat_bpf_map_create(enum bpf_map_type map_type,
+                              const char *map_name,
+                              __u32 key_size,
+                              __u32 value_size,
+                              __u32 max_entries,
+                              const struct bpf_map_create_opts *opts) {
```
poettering: we typically indent continuation lines to 16ch

src/shared/bpf-dlopen.h
```diff
@@ -29,6 +49,38 @@ extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
...
+        if (sym_bpf_map_create) {
+                return sym_bpf_map_create(map_type, map_name, key_size,
+                                          value_size, max_entries, opts);
+        }
```
poettering: as per coding style (https://systemd.io/CODING_STYLE/#formatting) no {} around single-line if blocks pls

src/shared/bpf-dlopen.h
```diff
@@ -29,6 +49,38 @@ extern bool (*sym_libbpf_probe_bpf_prog_type)(enum bpf_prog_type, const void *);
...
+        return sym_bpf_create_map(map_type, key_size, value_size, max_entries,
+                                  opts ? opts->map_flags : 0);
+}
+
+static inline int compat_libbpf_probe_bpf_prog_type(enum bpf_prog_type prog_type, const void *opts) {
+        if (sym_libbpf_probe_bpf_prog_type) {
+                return sym_libbpf_probe_bpf_prog_type(prog_type, opts);
+        }
```
poettering: ditto

src/shared/bpf-dlopen.c
```diff
@@ -28,6 +28,11 @@ const char* (*sym_bpf_program__name)(const struct bpf_program *);
 libbpf_print_fn_t (*sym_libbpf_set_print)(libbpf_print_fn_t);
 long (*sym_libbpf_get_error)(const void *);
 
+/* compat symbols removed in libbpf 1.0 */
+int (*sym_bpf_create_map)(enum bpf_map_type,  int key_size, int value_size, int max_entries, __u32 map_flags);
+bool (*sym_bpf_probe_prog_type)(enum bpf_prog_type, __u32);
+
+
```
poettering: spurious double empty lines

src/shared/bpf-dlopen.c
martinetd: Unfortunately, no; the list is not exactly the same.
`bpf_map_create` and `libbpf_probe_bpf_prog_type` got added in 0.6 and 0.7 respectively (released in Dec 2021 and Feb 2022), and many distros haven't updated yet to these so the direction was towards compat stuff.

If 0.7.0 is acceptable, then dlopen and dlsym_many_or_warn would be much easier, and all the compat cruft can be removed

bluca:
> So, what's the story here anyway? Can't we just hard switch to libbpf 1.0? What would the impact be for downstreams, CIs, and so on? If we can do without the compat kludges and just bump our deps, I'd be much happier.
> 
> @anitazha @DaanDeMeyer @mrc0mmand?

We can't, it breaks too many things, it was just released. The most important thing is to send the clear message that this must not happen again - it's the library that should have provided these compat kludges, forcing all users to duplicate them all over the place is absurd and not how libraries should behave.

In fact, they could still add these in 1.1 there's nothing stopping that from happening. If the answer is that the aesthetics of the internals of libbpf are more important to the maintainers than avoiding breaking users, then that would raise some extremely serious questions about the stability and viability of using libbpf in production, outside of toy projects. I would immediately disable it in Debian for example, and never enable it again until the situation changes for the better.

martinetd:
> The most important thing is to send the clear message that this must not happen again

(just FWIW I'm not a libbpf dev, and no-one from upstream have been involved in any of the discussions I could see here, so no message has been sent at this point. I highly doubt they'll add compat stuff back given the code lives in the kernel repository, but it can't hurt to ask.)

bluca:
> > The most important thing is to send the clear message that this must not happen again
> 
> (just FWIW I'm not a libbpf dev, and no-one from upstream have been involved in any of the discussions I could see here, so no message has been sent at this point. I highly doubt they'll add compat stuff back given the code lives in the kernel repository, but it can't hurt to ask.)

I am aware, the comments here are public and recorded, and were absolutely not directed at you personally. Sorry if I gave you that impression.

src/shared/bpf-dlopen.c
poettering: well, we still can have a single dlsym_many_or_warn() for the common parts, and then for the disjunct part have another one.

martinetd:
Sorry for the style, I meant it mostly as an example than actual submission but while I'm here might as well finish...

I've:
 - added bpf-compat.h to hide the ugly part
 - shared most symbols in a `dlsym_many_or_warn()`
 - fixed style remarks that still applied

I still don't understand what you mean by "proper ifdefs" -- do you want to limit yourself to the libbpf version that was available at compile time? I don't see any benefit from that, doing it properly woudl require adding meson checks and only looking for the symbols we found at that time. It's certainly possible but I don't think it's cleaner and it doesn't bring any benefit I can see (and won't work if libbpf gets upgraded, since systemd doesn't link against it directly it won't be noticed by packagers)

I also still haven't run any of this, I'm assuming bpf stuff are tested with older libbpf in the automatic tests and still intend to look at how to test this eventually™ (probably this weekend) with libbpf 1.0.

mbiebl:
On Debian we have `-BPF_FRAMEWORK`. This was apparently a wisely chosen decision on hindsight given the state of libbpf. Hopefully the library stabilizes/matures soon.

mbiebl:
Maybe systemd was just a bit too eager to pull that library in as a (optional) dependency given its immaturity

Anyway, just wanted to add that Debian/Ubuntu are not directly affected as those builds do not have bpf support enabled and it's probably prudent to wait a bit longer before doing so.

anitazha:
I've started a thread internally to see what direction the libbpf maintainers will take and if they would consider a compatibility layer. Likely won't get a response until next week since it's a long weekend for the company due to Labor Day.

bluca:
> I've started a thread internally to see what direction the libbpf maintainers will take and if they would consider a compatibility layer. Likely won't get a response until next week since it's a long weekend for the company due to Labor Day.

Thank you so much!

DaanDeMeyer:
I think we don't need compat declarations. We already use the skeleton APIs from libbpf 0.7, so I'm pretty sure we can raise our minimum required version to 0.7 (it was still 0.2 for some reason). If our minimum version is 0.7, we can depend on all the new APIs that replace the deprecated APIs, so we don't need any ifdeffery, and only need to try to load from both libbpf.so.1 and libbpf.so.0.

I've updated #24507 with these changes

anitazha:
After discussing with @anakryiko, it makes the most sense to have distributions bump to libbpf 0.8 which is both backwards compatible with 0.x releases and forwards compatible with 1.0. Building with the deprecated APIs will only print warnings in the 0.7-0.8 case. The distributions can remain on libbpf 0.8 as long as needed to allow ample time for projects to switch over from deprecated APIs and make changes. This path seems to be the most similar to what we take in systemd: print warnings for deprecation and eventually remove.

At this point it seems like the longest tail (on the distributions side) for getting off libbpf < 0.8 is Ubuntu 22.04 LTS which ends support in April 2027. Fedora 36 ends support in May 2023. All versions after that have libbpf 0.8 so projects can remove the deprecated APIs then.

martinetd:
Closing this to keep discussions focused -- let's continue in #24507

martinetd:
(reopened to update; I never remember if github refreshes the view of the branch if we push to pull branch with a PR closed but I think it doesn't...)

bluca:
needs a rebase

bluca:
apart from that, I think this is the way to go for now - if we manage to get some LTSes to ship 0.7 then we can slim this down again

martinetd:
rebased.

bluca:
```
In file included from ../src/shared/bpf-dlopen.h:9,
                 from ../src/shared/bpf-dlopen.c:4:
../src/shared/bpf-compat.h:24:102: error: ‘struct bpf_map_create_opts’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
   24 | extern int (*sym_bpf_map_create)(enum bpf_map_type,  const char *, __u32, __u32, __u32, const struct bpf_map_create_opts *);
      |                                                                                                      ^~~~~~~~~~~~~~~~~~~
../src/shared/bpf-compat.h:37:30: error: ‘struct bpf_map_create_opts’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
   37 |                 const struct bpf_map_create_opts *opts) {
      |                              ^~~~~~~~~~~~~~~~~~~
../src/shared/bpf-compat.h: In function ‘compat_bpf_map_create’:
../src/shared/bpf-compat.h:40:68: error: passing argument 6 of ‘sym_bpf_map_create’ from incompatible pointer type [-Werror=incompatible-pointer-types]
   40 |                                           value_size, max_entries, opts);
      |                                                                    ^~~~
      |                                                                    |
      |                                                                    const struct bpf_map_create_opts *
../src/shared/bpf-compat.h:40:68: note: expected ‘const struct bpf_map_create_opts *’ but argument is of type ‘const struct bpf_map_create_opts *’
../src/shared/bpf-compat.h:43:46: error: invalid use of undefined type ‘const struct bpf_map_create_opts’
   43 |                                   opts ? opts->map_flags : 0);
      |                                              ^~
../src/shared/bpf-dlopen.c: At top level:
../src/shared/bpf-dlopen.c:15:95: error: ‘struct bpf_map_create_opts’ declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
   15 | int (*sym_bpf_map_create)(enum bpf_map_type,  const char *, __u32, __u32, __u32, const struct bpf_map_create_opts *);
      |                                                                                               ^~~~~~~~~~~~~~~~~~~
../src/shared/bpf-dlopen.c:15:7: error: conflicting types for ‘sym_bpf_map_create’; have ‘int (*)(enum bpf_map_type,  const char *, __u32,  __u32,  __u32,  const struct bpf_map_create_opts *)’ {aka ‘int (*)(enum bpf_map_type,  const char *, unsigned int,  unsigned int,  unsigned int,  const struct bpf_map_create_opts *)’}
   15 | int (*sym_bpf_map_create)(enum bpf_map_type,  const char *, __u32, __u32, __u32, const struct bpf_map_create_opts *);
      |       ^~~~~~~~~~~~~~~~~~
In file included from ../src/shared/bpf-dlopen.h:9,
                 from ../src/shared/bpf-dlopen.c:4:
../src/shared/bpf-compat.h:24:14: note: previous declaration of ‘sym_bpf_map_create’ with type ‘int (*)(enum bpf_map_type,  const char *, __u32,  __u32,  __u32,  const struct bpf_map_create_opts *)’ {aka ‘int (*)(enum bpf_map_type,  const char *, unsigned int,  unsigned int,  unsigned int,  const struct bpf_map_create_opts *)’}
   24 | extern int (*sym_bpf_map_create)(enum bpf_map_type,  const char *, __u32, __u32, __u32, const struct bpf_map_create_opts *);
```

martinetd:
Ah... We only pass NULL pointers there now so we can probably alias it to void * (either properly checking via meson or just ugly macro). I'll do that tonight along with the actual version check.

martinetd:
Updated again:
 - my original libbpf 0.1.0 was correct; the third newest symbol we require is `bpf_map__set_max_entries@@LIBBPF_0.1.0` then it's all even older symbols we can rely on.
```
$ sed -ne 's/.*DLSYM_ARG(\([^)]*\)).*/\1/p' < code/systemd/src/shared/bpf-dlopen.c \
  | grep -Ff- <(nm -D /usr/lib64/libbpf.so.0 | grep @LIBBPF) | sort -t@ -k2
000000000000c4c0 T bpf_create_map_xattr@@LIBBPF_0.0.1
000000000000c570 T bpf_create_map_node@@LIBBPF_0.0.1
000000000000c5f0 T bpf_create_map@@LIBBPF_0.0.1
000000000000c660 T bpf_create_map_name@@LIBBPF_0.0.1
000000000000c6c0 T bpf_create_map_in_map_node@@LIBBPF_0.0.1
000000000000c750 T bpf_create_map_in_map@@LIBBPF_0.0.1
000000000000d5c0 T bpf_map_update_elem@@LIBBPF_0.0.1
000000000000d960 T bpf_map_delete_elem@@LIBBPF_0.0.1
00000000000184c0 T libbpf_set_print@@LIBBPF_0.0.1
000000000001f0c0 T bpf_map__fd@@LIBBPF_0.0.1
000000000001fc10 T bpf_map__name@@LIBBPF_0.0.1
00000000000207a0 T bpf_map__set_inner_map_fd@@LIBBPF_0.0.1
0000000000020ec0 T libbpf_get_error@@LIBBPF_0.0.1
000000000002ef90 T bpf_probe_prog_type@@LIBBPF_0.0.2
0000000000025920 T bpf_link__destroy@@LIBBPF_0.0.4
000000000001e440 T bpf_program__name@@LIBBPF_0.0.7
000000000002c640 T bpf_object__open_skeleton@@LIBBPF_0.0.7
000000000002c860 T bpf_object__load_skeleton@@LIBBPF_0.0.7
000000000002ca00 T bpf_object__attach_skeleton@@LIBBPF_0.0.7
000000000002cad0 T bpf_object__detach_skeleton@@LIBBPF_0.0.7
000000000002cb30 T bpf_object__destroy_skeleton@@LIBBPF_0.0.7
00000000000259a0 T bpf_link__fd@@LIBBPF_0.0.8
0000000000027450 T bpf_program__attach_lsm@@LIBBPF_0.0.8
0000000000027470 T bpf_program__attach_cgroup@@LIBBPF_0.0.8
000000000001c6d0 T bpf_map__set_max_entries@@LIBBPF_0.1.0
000000000000c130 T bpf_map_create@@LIBBPF_0.6.0
000000000002ef00 T libbpf_probe_bpf_prog_type@@LIBBPF_0.7.0
```
 - fixed the build issue:
   - struct declared in parameter type can be solved by just declaring `struct bpf_map_create_opts` unconditionally: that doesn't hurt with newer versions
   - we also need to avoid the `opts->map_flags` reference, but since we know opts is NULL (won't compile if someone tries to add opts somewhere, because `LIBBPF_OPTS` is missing in older versions among other things) we can just pass 0 instead.

It's still a bit more ugly than I'd have liked, so if you prefer checking version >= 0.7.0 and use one path or the other at compile time rather than dlopen time I guess I can rework that; but it should otherwise work.
Would appreciate guidance on how to do light tests if there are any for bpf, otherwise I'll try to spin it in a VM and use some IP filtering feature to verify bpf is loaded.

src/shared/bpf-compat.h
```diff
@@ -0,0 +1,54 @@
...
+        return sym_bpf_create_map(map_type, key_size, value_size, max_entries,
+                                  0 /* opts->map_flags, but opts is always NULL for us so skip build dependency on the type */);
+}
+
+static inline int compat_libbpf_probe_bpf_prog_type(enum bpf_prog_type prog_type, const void *opts) {
+        if (sym_libbpf_probe_bpf_prog_type)
+                return sym_libbpf_probe_bpf_prog_type(prog_type, opts);
+
+        return sym_bpf_probe_prog_type(prog_type, 0);
+}
```
keszybz, line 54: It's incredible that libbpf broke symbol version to avoid providing those two trivial wrappers.

keszybz (commented):
LGTM.

src/shared/bpf-compat.h
bluca: yes, it makes no sense

bluca (approved):

bluca:
Thanks for working on this!

keszybz:
We'll want to backport this, after it has baked a bit.

keszybz:
This is already in v251.6, dropping `needs-stable-backport`.

PR #24520, opened 2022-08-31
udevadm-wait: also listen kernel uevents

yuwata:
Suggested at https://github.com/systemd/systemd/pull/24471#discussion_r959703103.

### DIFF
--- a/src/udev/udevadm-wait.c
+++ b/src/udev/udevadm-wait.c
@@ -166,16 +166,9 @@ static int device_monitor_handler(sd_device_monitor *monitor, sd_device *device,
         if (path_strv_contains(arg_devices, name))
                 return check_and_exit(sd_device_monitor_get_event(monitor));
 
-        STRV_FOREACH(p, arg_devices) {
-                const char *link;
-
-                if (!path_startswith(*p, "/dev"))
-                        continue;
-
-                FOREACH_DEVICE_DEVLINK(device, link)
-                        if (path_equal(*p, link))
-                                return check_and_exit(sd_device_monitor_get_event(monitor));
-        }
+        FOREACH_DEVICE_DEVLINK(device, name)
+                if (path_strv_contains(arg_devices, name))
+                        return check_and_exit(sd_device_monitor_get_event(monitor));
 
         return 0;
 }


src/udev/udevadm-wait.c
```diff
@@ -410,16 +410,20 @@ int wait_main(int argc, char *argv[], void *userdata) {
...
         if (arg_wait_until == WAIT_UNTIL_ADDED) {
                 /* If --initialized=no is specified, it is not necessary to wait uevents for the specified
-                 * devices to be processed by udevd. Let's periodically check the devices. Then, we may be
-                 * able to finish this program earlier when udevd is very busy.
-                 *
-                 * This is useful for working around issues #24360 and #24450.
+                 * devices to be processed by udevd. Hence, let's listen on the kernel's uevent stream. Then,
+                 * we may be able to finish this program earlier when udevd is very busy.
+                 * Note, we still need to also setup udev monitor, as this may be invoked with a devlink. */
+                r = setup_monitor(event, MONITOR_GROUP_KERNEL, "kernel-uevent-monitor-event-source", &kernel_monitor);
```
poettering, line 431: why listen to both? why not just the kernel group if we don't need udev, and not bother with udev?

poettering:
if the time-based periodic stuff is left in, at the very least there should be some debug logging. i.e. only act if we see the device node in two periodic wake-ups and still haven't see a netlink event for it. and then debug log about this fact,

src/udev/udevadm-wait.c
yuwata: As explained in the comment, user may invoke with a devlink. In that case, we need to wait the uevent processed by udevd.

yuwata:
> if the time-based periodic stuff is left in, at the very least there should be some debug logging. i.e. only act if we see the device node in two periodic wake-ups and still haven't see a netlink event for it. and then debug log about this fact,
> oh, and the timer event should have low prio, so that the netlink event always is dispatched first, if both are queued

@poettering Thank you for the suggestion. PTAL.

src/udev/udevadm-wait.c
poettering: ahh, thats a good point

src/udev/udevadm-wait.c
```diff
@@ -179,14 +180,14 @@ static int device_monitor_handler(sd_device_monitor *monitor, sd_device *device,
...
-static int setup_monitor(sd_event *event, sd_device_monitor **ret) {
+static int setup_monitor(sd_event *event, MonitorNetlinkGroup group, const char *description, sd_device_monitor **ret) {
         _cleanup_(sd_device_monitor_unrefp) sd_device_monitor *monitor = NULL;
         int r;
 
         assert(event);
         assert(ret);
 
-        r = sd_device_monitor_new(&monitor);
+        r = device_monitor_new_full(&monitor, group, -1);
```
poettering: maybe `…, /* fd= */ -1);`?

Issue #23913, opened 2022-07-05
bootctl accepts relative paths for linux and initrd


ebsf1:
### systemd version the issue has been seen with

249

### Used distribution

Ubuntu 22.04

### Linux kernel version used

5.15.0-40-generic; 5.15.0-37-generic

### CPU architectures issue was seen on

x86_64

### Component

bootctl

### Expected behaviour you didn't see

Boot entries conforming with the [Boot Loader Specification](https://systemd.io/BOOT_LOADER_SPECIFICATION/) according to (a) its language; and (b) the output of `bootctl list`, which according to its [documentation](https://www.freedesktop.org/software/systemd/man/bootctl.html#list) returns conforming entries, would be observed and appear in the list on the systemd-boot screen.

### Unexpected behaviour you saw

Conforming entries do not appear in the systemd-boot list.

Specifically, systemd-boot, perhaps relying on systemd-bootx64.efi, does not recognize paths to kernel files on the file system of the EFI mount point or otherwise not on the ESP, _e.g._, ../vmlinuz or ../initrd.img.

It is elementary that the entire purpose of any boot loader (and therefore systemd-boot, perhaps in conjunction with systemd-bootx64.efi), after all, is simply to hand off execution to kernel files on another partition and file system.  systemd-boot seems incapable, however, of reading them.

Indeed, the entire construct of `kernel-install` and its related `bootctl` options, appears designed to compensate for the absence of this capability.  It attempts too much.  Das Gummiboot wird gesunken.  Establish this fundamental read capability and none of this is necessary.  Kernel files can remain encapsulated on the partitions they boot and kernel installation and update procedures can be left intact.

### Steps to reproduce the problem

This is a multiboot Linux system, chiefly with different versions and flavours of Ubuntu installed on discrete partitions, all on a single volume.  The ESP is /dev/sda1, /dev/sda2-7 are bootable, /dev/sda8 is swap, and /dev/sda9 is data.  The installations were competing for control over the GRUB boot loader, prompting an attempt to replace it with systemd-boot.

The base configuration is a conventional Ubuntu Desktop 22.04 installation:
- Kernel files are located in /boot, along with the symlinks vmlinuz and initrd.img, which reference the latest versions.
- The ESP is mounted to /boot/efi via fstab.

/boot/efi/loader/loader.conf is:

```
#	/boot/efi/loader/loader.conf
#	systemd-boot configuration file

timeout 5
#default [machine-id redacted]-*
[...]

PR #24521, opened 2022-08-31
boot loader spec fixes

### LINKED ISSUE #24247: boot loader spec is confused about $BOOT
the boot loader spec now defines $BOOT in a way that is different than before, and totally different from what is implemented in bootctl.

Previously $BOOT refers to XBOOTLDR if it exists, and ESP otherwise. 

Now it always points to the ESP?

But bootctl doesn't implement it that way.

I think the spec change should really be reverted, i.e.  #23504

poettering:
Fixes #24247

### DIFF
--- a/docs/BOOT_LOADER_SPECIFICATION.md
+++ b/docs/BOOT_LOADER_SPECIFICATION.md
@@ -8,12 +8,12 @@ SPDX-License-Identifier: LGPL-2.1-or-later
 # The Boot Loader Specification
 
 This document defines a set of file formats and naming conventions that allow
-the boot loader configuration to be shared between multiple operating systems
+the boot loader menu entries to be shared between multiple operating systems
 and boot loaders installed on one device.
 
-Operating systems cooperatively manage a boot loader configuration directory
-that contains drop-in files, making multi-boot scenarios easy to support. Boot
-menu items are defined via a simple format that can be understood by different
+Operating systems cooperatively manage boot loader menu entry directories that
+contain drop-in files, making multi-boot scenarios easy to support. Boot menu
+entries are defined via two simple formats that can be understood by different
 boot loader implementations, operating systems, and userspace programs. The
 same scheme can be used to prepare OS media for cases where the firmware
 includes a boot loader.
@@ -23,72 +23,105 @@ includes a boot loader.
 The target audience for this specification is:
 
 * Boot loader developers, to write a boot loader that directly reads its
-  configuration from these files
+  menu entries from these files
 * Firmware developers, to add generic boot loading support directly to the
   firmware itself
 * OS installer developers, to create appropriate partitions and set up the
-  initial boot loader configuration
-* Distribution developers, to create appropriate configuration snippets when
+  initial boot loader menu entries
+* Distribution developers, to create appropriate menu entry snippets when
   installing or updating kernel packages
 * UI developers, to implement user interfaces that list and select among the
   available boot options
 
-## The Boot Partition
+## The Partitions
 
 Everything described below is located on one or two partitions. The boot loader
-or user-space programs reading the boot loader configuration should locate them
+or user-space programs reading the boot loader menu entries should locate them
 in the following manner:
 
 * On disks with an MBR partition table:
 
-  * The boot partition — partition with the type ID of 0xEA — shall be used
-    for boot loader configuration and entries.
+  * The boot partition — a partition with the type ID of `0xEA` — shall be used
+    as the single location for boot loader menu entries.
 
 * On disks with GPT (GUID Partition Table)
 
-  * The EFI System Partition (ESP for short) — a partition with GPT type GUID
-    of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — should be used for boot loader
-    configuration and boot entries.
+  * The EFI System Partition (ESP for short) — a partition with a GPT type GUID
+    of `c12a7328-f81f-11d2-ba4b-00a0c93ec93b` — may be used as one of two locations for
+    boot loader menu entries.
 
   * Optionally, an Extended Boot Loader Partition (XBOOTLDR partition for
     short) — a partition with GPT type GUID of
-    `bc13c2ff-59e6-4262-a352-b275fd6f7172` — may be used as an additional
-    location for boot loader entries. This partition must be located on the
-    same disk as the ESP.
+    `bc13c2ff-59e6-4262-a352-b275fd6f7172` — may be used as the second of two
+    locations for boot loader menu entries. This partition must be located on
+    the same disk as the ESP.
 
-In the text below, `$BOOT` will be used to refer to (the root of) the first of
-the two partitions (the boot partition on MBR disks and the ESP on GPT disks),
-and `$XBOOTLDR` will be used to refer to (the root of) the optional second
-partition.
+There may be at most one partition of each of the types listed above on the
+same disk.
 
-An installer for the operating system should use this logic when selecting or
-creating partitions:
-
-  * If `$BOOT` is not found, a new suitably sized partition (let's say 500MB)
-    should be created, matching the characteristics described above. On disks
-    with GPT, only the ESP partition without the XBOOTLDR partition should be
-    created.
-
-  * If the OS is installed on a disk with GPT and the ESP partition is found
-    but is too small, a new suitably sized (let's say 500MB) XBOOTLDR partition
-    shall be created.
-
-Those file systems shall be determined during _installation time_, and an fstab
-entry may be created. If only one partition is used, it should be mounted on
-`/boot/`. If both XBOOTLDR partition and the ESP are used, they should be
-mounted on `/boot` and `/efi`, or on `/boot` and `/boot/efi`.
-
-**Note:** _Those file systems are **shared** among all OS installations on the
-system. Instead of maintaining one boot partition per installed OS (as `/boot/`
-was traditionally handled), all installed OSes use the same place for boot-time
-configuration._
+**Note:** _These partitions are **shared** among all OS installations on the
+same disk. Instead of maintaining one boot partition per installed OS (as
+`/boot/` was traditionally handled), all installed OSes use the same place for
+boot loader menu entries._
 
 For systems where the firmware is able to read file systems directly, the ESP
-must — and the XBOOTLDR partition should — be a file system readable by the
-firmware. For most systems this means VFAT (16 or 32 bit). Applications
-accessing both partitions should hence not assume that fancier file system
-features such as symlinks, hardlinks, access control or case sensitivity are
-supported.
+must — and the MBR boot and GPT XBOOTLDR partition should — be a file system
+readable by the firmware. For most systems this means VFAT (16 or 32
+bit). Applications accessing both partitions should hence not assume that
+fancier file system features such as symlinks, hardlinks, access control or
+case sensitivity are supported.
+
+### The `$BOOT` Partition Placeholder
+
+In the text below, the placeholder `$BOOT` will be used to refer to the
+partition determined as follows:
+
+ 1. On disks with an MBR partition table: → the boot partition, as described above
+
+ 2. On disks with a GPT partition table: → the XBOOTLDR partition if it exists
+
+ 3. Otherwise, on disks with a GPT partition table: → the ESP
+
+`$BOOT` is the *primary* place to put boot menu entry resources into, but
+typically not the only one. Most importantly, boot loaders should also pick up
+menu entries from the ESP, even if XBOOTLDR exists (for details see below).
+
+### Creating These Partitions
+
+An installer for an operating system should use this logic when selecting or
+creating partitions:
+
+  * If a boot partition (in case of MBR) or an XBOOTLDR partition (in case of
+    GPT) already exists it should be used as `$BOOT` and used as primary
+    location to place boot loader menu resources in.
+
+  * Otherwise, if on GPT and an ESP is found and it is large enough (let's say
+    at least 1G) it should be used as `$BOOT` and used as primary location to
+    place boot loader menu resources in.
+
+  * Otherwise, if on GPT and neither XBOOTLDR nor ESP exist, an ESP should be
+    created of the appropriate size and be used as `$BOOT`, and used as primary
+    location to place boot loader menu resources in.
+
+  * Otherwise, a boot partition (in case of MBR) or XBOOTLDR partition (in case
+    of GPT) should be created of an appropriate size, and be used as `$BOOT`,
+    and used as primary location to place boot loader menu resources in.
+
+These partitions shall be determined during _installation time_, and
+`/etc/fstab` entries may be created.
+
+### Mount Points
+
+It is recommended to mount `$BOOT` to `/boot/`, and the ESP to `/efi/`. If
+`$BOOT` and the ESP are the same, then either a bind mount or a symlink should
+be established making the partition available under both paths.
+
+(Mounting the ESP to `/boot/efi/`, as was traditionally done, is not
+recommended. Such a nested setup complicates an implementation via direct
+`autofs` mounts — as implemented by `systemd` for example —, as establishing
+the inner `autofs` will trigger the outer one. Mounting the two partitions via
+`autofs` is recommended because the simple VFAT file system has weak data
+integrity properties and should remain unmounted whenever possible.)
 
 ## Boot Loader Entries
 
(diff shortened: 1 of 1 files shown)


medhefgo:
So `$BOOT`/`/boot` refers to (in order of preference) `XBOOTLDR, ESP, MBRBOOT` and `/efi` always to the ESP (and may be a symlink to `/boot` if no XBOOTLDR partition was found on the same disk as the ESP, right?

This feels overcomplicated to me, but I also don't really care or have suggestions to make this simpler.

Also, while you're at it, might as well make sure the spec is [lawyer compatible](https://github.com/systemd/systemd/issues/23913#issuecomment-1176842284) kekw.

poettering:
> So `$BOOT`/`/boot` refers to (in order of preference) `XBOOTLDR, ESP, MBRBOOT` and `/efi` always to the ESP (and may be a symlink to `/boot` if no XBOOTLDR partition was found on the same disk as the ESP, right?

Correct. You can say it that way. (It should be fine to do the symlink the other way: /boot/ → /efi/. And it could also be a bind mount).

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -23,72 +23,106 @@ includes a boot loader.
...
 
-An installer for the operating system should use this logic when selecting or
+**Note:** _These partitions are **shared** among all OS installations on the
+same disk. Instead of maintaining one boot partition per installed OS (as
+`/boot/` was traditionally handled), all installed OSes use the same place for
+boot loader menu entries._
+
+For systems where the firmware is able to read file systems directly, the ESP
+must — and the MBR boot and GPT XBOOTLDR partition should — be a file system
+readable by the firmware. For most systems this means VFAT (16 or 32
```
cmurf, line 69: Would [efifs](https://github.com/pbatard/efifs) drivers satisfy this requirement? The pre-boot environment would thus be able to read most any file system, but of course it's limited since these fs drivers need to be signed for Secure Boot. That means we'd potentially get this explosion of distro signed drivers, having to encourage everyone to follow such a paradigm just to be certain they could read whatever volume format $BOOT happens to be. Permissive or strict?

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -23,72 +23,106 @@ includes a boot loader.
...
-mounted on `/boot` and `/efi`, or on `/boot` and `/boot/efi`.
+  * Otherwise, if on GPT and neither XBOOTLDR nor ESP exist, an ESP should be
+    created of the appropriate size and be used as `$BOOT`, and used as primary
+    location to place boot loader menu resources in.
 
-**Note:** _Those file systems are **shared** among all OS installations on the
-system. Instead of maintaining one boot partition per installed OS (as `/boot/`
-was traditionally handled), all installed OSes use the same place for boot-time
-configuration._
+  * Otherwise, a boot partition (in case of MBR) or XBOOTLDR partition (in case
```
cmurf, line 106: I'm confused. I thought $XBOOTLDR applies to MBR with 0xEA type code and GPT. What is a "boot partition" ? Seems vague, but also is not $XBOOTLDR.

docs/BOOT_LOADER_SPECIFICATION.md
medhefgo: Yes efifs works, as do any other filesystems supported by the firmware (ASUS boards can handle NTFS and I think exfat on newer ones).

But I always wonder what people have against fat for boot filesystems and would rather go through all kinds of hoops to avoid it. Fat is perfectly adequate for the job and mandated to be supported by the spec. You don't *need* fancy features on a filesystem for this.

poettering: I think using any other FS than FAT for XBOOTLDR is stupid. General purpose distros shouldn't do that. People who don't have to care about the general purpose and only focus on specific devices (i.e. devices without need for multi-boot) otoh can choose something else if they really like to.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: Nah, the spec calls the MBR 0xEA partition type just "boot partition". 

The spec calls the GPT partition type bc13c2ff-59e6-4262-a352-b275fd6f7172 "XBOOTLDR partition".

The spec calls the GPT partition type c12a7328-f81f-11d2-ba4b-00a0c93ec93b "EFI System Partition"

and $BOOT is a magic variable that in order of preference either points to MBR 0xEA, or GPT bc13c2ff-59e6-4262-a352-b275fd6f7172 or GPT c12a7328-f81f-11d2-ba4b-00a0c93ec93b.

docs/BOOT_LOADER_SPECIFICATION.md
cmurf: A significant advantage of using Btrfs (or ZFS) for this purpose is pooled space. You don't have to worry about the problem of $BOOT size, it's all shared space with $ROOT. I wonder if the spec should permit multiple XBOOTLDR, in effect pooling them? This means we can start with 1G $BOOT and if later on it's not big enough, just add another one instead of having to reprovision.

e.g. consider the case of https://github.com/systemd/systemd/pull/23841 pinning kernels+initramfs to $BOOT so that older snapshots can be booted. Estimating the size of $BOOT isn't exactly straightforward in advance, so in these cases we tend to over provision to avoid premature reprovision.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -108,37 +142,43 @@ from the user. Only entries matching the feature set of boot loader and system
...
-each boot menu item. Each OS may provide one or more such entries. The boot
-loader should enumerate both directories and provide a merged list.
-
-The file name is used for identification of the boot item but shall never be
-presented to the user in the UI. The file name may be chosen freely but should
-be unique enough to avoid clashes between OS installations. More specifically,
-it is suggested to include the `entry-token` (see
+`/loader/entries/` in `$BOOT` is the primary directory containing Type #1
+drop-in snippets defining boot entries, one `.conf` file for each boot menu
+item. Each OS may provide one or more such entries.
```
cmurf, line 163: As far as I can tell, this is the strongest language the spec gets when it comes to "ownership" of the snippet files themselves. I don't know for sure it needs to be made stronger, because it seems obvious that the purpose of snippet, kernel, initramfs, and defining $ROOT (an `/etc` and a `/usr` at least) is to boot a particular OS (or more specifically a particular root for a particular OS). We need all of these things to do that, so it seems really suspect on the face of it if a distro were to modify another distro's snippets.

However: https://bugzilla.redhat.com/show_bug.cgi?id=2120845 is an near example. Since Fedora doesn't set `/boot` as either 0xEA, or GPT partition type XBOOTLDR, it's not yet stepping on anything but other Fedora installations, and not by default either since the installer currently doesn't reuse `/boot` across installations. So right now it's just a bad look rather than a bad actor. I think.

But there is a sort of conflict raised between whether `kernel-install` intends to completely replace the entire kernel command line from `/etc/kernel/cmdline` into all snippets. Again, in the snapshotting example, the only way to define a root file system for LVM/Stratis/OSTree is a kernel boot parameter, so it has to be reliable. Maybe there should be different options lines with different owners? Like, options2 can be for things like video card params that could legit apply to every BLS snippet on that system, whether or not the same distro. Whereas option1 could be for root, rootflags, rd.md.uuid, rd.luks.uuid, and things that are necessary to find $ROOT.

I guess what I'm getting at is what seems obvious to me (don't do that) is not in the spec and others are saying well it's not in the spec so it's OK.

ebsf1:
> Also, while you're at it, might as well make sure the spec is [lawyer compatible](https://github.com/systemd/systemd/issues/23913#issuecomment-1176842284) kekw.


@medhefgo you miss the point, which has nothing to do with lawyering and everything to do with literacy.  kekw yourself.  Address the substance, if you can.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: @cmurf well, it am not convinced it's that hard to come up with a good estimation how much space you'll need and then double that or so. But sure, if this really really becomes an issue we can revisit this one day, but I am not convinced we are at that day yet. 

I mean, yes, when ESPs first appeared on the scene people didn't know how we'd like to use them today, so we added XBOOTLDR to address that. But we know pretty well what we want to use XBOOTLDR for for now, which makes it much easier to estimate a size that might work.

One day, when we realize that maybe we want to use it differently we can revisit things, but doing so ahead of time without knowing the problems the future brings is premature.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: I am not convinced that retro-updating old boot entries is a good thing. It's OK to update the current kernel, but the point of keeping the old stuff around is to keep the old stuff around, not to keep some hybrid of old and new around.

So, yes, it might make sense to invoke "kernel-install" for an existign kernel, to update things, but I am very strongly opposed to having a default codepaths of updating *all* kernels like this.

(I'd be fine with adding an option to kernel-install that *on* *request* allows rerunning the tool for all existing kernels (e.g. `kernel-install -a`), regenerating all snippets, but that should be an option for advanced admins/hackers, not part of the regular code flow)

poettering:
@keszybz could you bless this?

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -8,12 +8,12 @@ SPDX-License-Identifier: LGPL-2.1-or-later
...
 This document defines a set of file formats and naming conventions that allow
-the boot loader configuration to be shared between multiple operating systems
+the boot loader menu entries to be shared between multiple operating systems
 and boot loaders installed on one device.
 
-Operating systems cooperatively manage a boot loader configuration directory
-that contains drop-in files, making multi-boot scenarios easy to support. Boot
-menu items are defined via a simple format that can be understood by different
+Operating systems cooperatively manage boot loader menu entry directories that
+contains drop-in files, making multi-boot scenarios easy to support. Boot menu
```
keszybz: contains → contain

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -23,72 +23,106 @@ includes a boot loader.
...
-or user-space programs reading the boot loader configuration should locate them
+or user-space programs reading the boot loader menu entries should locate them
 in the following manner:
 
 * On disks with an MBR partition table:
 
-  * The boot partition — partition with the type ID of 0xEA — shall be used
-    for boot loader configuration and entries.
+  * The boot partition — a partition with the type ID of `0xEA` — shall be used
+    as single location for boot loader menu entries.
```
keszybz: article missing before "single location"

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -23,72 +23,106 @@ includes a boot loader.
...
+  * If a boot partition (in case of MBR) or an XBOOTLDR partition (in case of
+    GPT) already exists it should be used as `$BOOT` and used as primary
+    location to place boot loader menu resources in.
 
-  * If the OS is installed on a disk with GPT and the ESP partition is found
-    but is too small, a new suitably sized (let's say 500MB) XBOOTLDR partition
-    shall be created.
+  * Otherwise, if on GPT and an ESP is found and it is large enough (let's say
+    at least 1G) it should be used as `$BOOT` and used as primary location to
+    place boot loader menu resources in.
```
keszybz, line 100: s/boot loader menu resources/boot loader resources/
(below too)

"menu" here is strange… E.g. a sysext is a "boot loader resource", but has nothing to do with the menu.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -23,72 +23,106 @@ includes a boot loader.
...
+`$BOOT` and the ESP are the same, then either a bind mount or a symlink should
+be established making the partition available under both paths.
+
+(It's not recommended to mount the ESP to `/boot/efi/`, as was traditionally
+done. Such a nested setup complicates an implementation via direct `autofs`
+mounts — as implemented by `systemd` for example —, as establishing the inner
+`autofs` will trigger the outer one. Mounting the two partitions via `autofs`
+is a good idea though, since the simple VFAT file system they typically use,
+has weak data integrity properties and thus should be kept unmounted whenever
+possible.)
```
keszybz: (Mounting the ESP to `/boot/efi/`, as was traditionally done, is not recommended. Such a nested setup complicates an implementation via direct `autofs` mounts — as implemented by `systemd` for example —, as establishing the inner `autofs` will trigger the outer one. Mounting the two partitions via `autofs` is recommended because the simple VFAT file system has weak data integrity properties and should be kept unmounted if possible.)

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -108,37 +142,43 @@ from the user. Only entries matching the feature set of boot loader and system
...
-particularly common in the case of the ESP. Implementations of this specification
-must be able to operate correctly if files or directories other than `/loader/`
-are found in the top level directory. Implementations that add their own files
-or directories to the file systems should use well-named directories, to make
-name collisions between multiple users of the file system unlikely.
+Note that the three partitions described above are not supposed to be the
+exclusive territory of this specification. This specification only defines
+semantics of the `/loader/entries/` directory (along with the companion file
+`/loader/entries.srel`) and the `/EFI/Linux/` directory inside the file system,
+but it doesn't intend to define ownership of the rest of the file system. Boot
```
keszybz: s/ownership/contents/ ?

docs/BOOT_LOADER_SPECIFICATION.md
keszybz: I think don't think the spec needs to say more here. In general, the expectation is that normal updates just add new entries, and the spec describes how to pick file names to avoid conflicts. But there may be situations where all entries for a given OS should be edited, or maybe even a situation where editing entries for other OSes makes sense. So any questions about what to update and when are left to the implementation.

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -202,16 +242,17 @@ The following keys are recognized:
 
   Example: `sort-key fedora`
 
-* `linux` is the Linux kernel to spawn and as a path relative to file system
-  root. It is recommended that every distribution creates a machine id and
+* `linux` is the Linux kernel to spawn and is a path relative to the root of
+  the file system the boot entry snippet itself is located in. It is
```
keszybz: `linux` is the Linux kernel to spawn and is a path relative to the root of the file system containing the boot entry snippet itself

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -202,16 +242,17 @@ The following keys are recognized:
...
+  recommended that every distribution creates a entry token/machine id and
```
keszybz: an entry-token/machine-id

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -202,16 +242,17 @@ The following keys are recognized:
...
   version specific subdirectory and places its kernels and initial RAM disk
-  images there.
+  images there (see below).
```
keszybz: initrd images

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -220,7 +261,7 @@ The following keys are recognized:
 
 * `options` shall contain kernel parameters to pass to the Linux kernel to
   spawn. This key is optional and may appear more than once in which case all
-  specified parameters are used in the order they are listed.
+  specified parameters are joined in the order they are listed.
```
keszybz: joined → combined ?

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -344,57 +392,61 @@ On EFI, any such images shall be added to the list of valid boot entries.
...
-files `$BOOT/loader/entries/*.conf` and `$XBOOTLDR/loader/entries/*.conf`, and
-populates its boot menu. On EFI, it then extends this with any unified kernel
-images found in `$BOOT/EFI/Linux/*.efi` and `$XBOOTLDR/EFI/Linux/*.efi`. It may
-also add additional entries, for example a "Reboot into firmware" option.
-Optionally it may sort the menu based on the `sort-key`, `machine-id` and
-`version` fields, and possibly others. It uses the file name to identify
-specific items, for example in case it supports storing away default entry
-information somewhere. A boot loader should generally not modify these files.
+A _boot loader_ locates the `$BOOT` partition and the ESP, then simply reads
+all the files `/loader/entries/*.conf` in them, and populates its boot menu
```
keszybz: Hmm, so this is confusing. I think you need to say "locates XBOOTLDR and ESP".

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -344,57 +392,61 @@ On EFI, any such images shall be added to the list of valid boot entries.
...
+(and handle gracefully if one of the two partitions is missing, or both are the
+same). On EFI, it then extends this with any unified kernel images found in
```
keszybz: What does it mean "the same"? They have different UUIDs…

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -567,18 +619,18 @@ This specification brings the following advantages:
   interfering with the entries of other installed operating systems.
 
 * Drop-in directories are now pretty ubiquitous on Linux as an easy way to
-  extend configuration without having to edit, regenerate or manipulate
+  extend boot loader menus without having to edit, regenerate or manipulate
```
keszybz, line 621: this chunk should be reverted

docs/BOOT_LOADER_SPECIFICATION.md
```diff
@@ -567,18 +619,18 @@ This specification brings the following advantages:
...
+* Userspace code can sanely parse boot loader menu entries which is essential
   with modern firmware which does not necessarily initialize USB keyboards
   during boot, which makes boot menus hard to reach for the user. If userspace
-  code can parse the boot loader configuration too, UI can be written that
+  code can parse the boot loader menu entries too, UI can be written that
   select a boot menu item to boot into before rebooting the machine, thus not
   requiring interactivity during early boot.
 
-* To unify and thus simplify configuration of the various boot loaders, which
+* To unify and thus simplify menu entries of the various boot loaders, which
```
keszybz, line 632: Also here, the sentence made more sense before. The resulting menu entries are about the same, just how we derive them is changed.

keszybz (commented):
Sorry for the delay. Looks reasonable, apart from some minor issues.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: $BOOT can be the ESP... But OK, here we can change $BOOT to XBOOTLDR, so that they are distinct.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: So, the reason I wrote it like that is that resources associated directly with the boot loader, i.e. drivers and such should be in the ESP, but those that belong to the menu entries may be in XBOOTLDR, too.

So I wanted to clarify this: the stuff that is shipped with the boot loader itself (i.e. sd-boot.efi, loader.conf, maybe driver .efi and so on) is not meant here, but the stuff shipped for the boot loader menu entries is what we are talking about.

docs/BOOT_LOADER_SPECIFICATION.md
poettering: to me loader.conf is boot loader *configuration*, but this spec is not about loader.conf, that's not even mentioned. Instead here we focus on menu entry descriptions, which aren't really *configuration*

poettering:
Force pushed a new version. I addressed everything, except for the configuration/menu entry thing, see above for my explanation.

I'd be happy to change it to something else, but just saying "configuration" I think would be misleading, since there's other stuff boot loaders maintain that *actually* is configuration and this is much more static than that.

poettering:
So, given this is documentation only, let's just merge this. The only remaining thing (i.e. the configuration vs. menu entry question) can be dealt with in a later fix-up commit

PR #24522, opened 2022-08-31
core/device: drop nonexistent devlink units

### LINKED ISSUE #24518: Stale devlink device units are not cleaned up while dispatching UDEV change events

239-58.el8_6.4 (appears to affect `main`)

### Used distribution

AlmaLinux 8

### Linux kernel version used

4.18.0-372.19.1.el8_6.x86_64

x86_64

### Component

systemd

Systemd should remove stale devlinks device units while dispatching UDEV "change" events for the parent device (e.g. when renaming an LVM Logical Volume).

Currently systemd will create new device units for any new paths appearing in devlinks in a UDEV change event, but will not remove previously-created device units for paths which are no longer present in devlinks. This leaves device units for non-existent paths hanging around until the parent device is removed.

Further, if a daemon-reload is issued before the parent device unit is removed, the stale devlinks units lose their association with the parent device (through matching syspath) during device enumeration because the paths no longer exist, leaving them hanging around for ever (until rebooting).

Although I tested with 239-58.el8_6.4 on AlmaLinux 8, after checking over `src/core/device.c`, this still appears to affect `main`; there doesn't appear to be any special handling for the "change" event case.

Create a new Logical Volume:

```
$ lvcreate -n test-a-1 -L1G test
  Logical volume "test-a-1" created.
```

The following related units are created:

```
dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d1.device                                          loaded active plugged /dev/disk/by-id/dm-name-test-test--a--1
dev-disk-by\x2did-dm\x2duuid\x2dLVM\x2dvtt62H07Vx8I5nL9HB9qukE8AelCGphakqbY4UX2IOgOE8Da0yTpuolAtvHg6NL6.device loaded active plugged /dev/disk/by-id/dm-uuid-LVM-vtt62H07Vx8I5nL9HB9qukE8AelCGphakqbY4UX2IOgOE8Da0yTpuolAtvHg6NL6
dev-dm\x2d0.device                                                                                             loaded active plugged /dev/dm-0
dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d1.device                                                               loaded active plugged /dev/mapper/test-test--a--1
dev-test-test\x2da\x2d1.device                                                                                 loaded active plugged /dev/test/test-a-1
sys-devices-virtual-block-dm\x2d0.device                                                                       loaded active plugged /sys/devices/virtual/block/dm-0
```

Rename the previously-created LV:

```
$ lvrename /dev/test/test-a-1 test-a-2
  Renamed "test-a-1" to "test-a-2" in volume group "test"
```

The following paths no longer exist as a result:

```
$ ls -lh /dev/disk/by-id/dm-name-test-test--a--1 /dev/mapper/test-test--a--1 /dev/test/test-a-1
ls: cannot access '/dev/disk/by-id/dm-name-test-test--a--1': No such file or directory
ls: cannot access '/dev/mapper/test-test--a--1': No such file or directory
ls: cannot access '/dev/test/test-a-1': No such file or directory
```

Yet the following related device units now exist:

```
dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d1.device                                          loaded active plugged /dev/disk/by-id/dm-name-test-test--a--1
dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d2.device                                          loaded active plugged /dev/disk/by-id/dm-name-test-test--a--2
dev-disk-by\x2did-dm\x2duuid\x2dLVM\x2dvtt62H07Vx8I5nL9HB9qukE8AelCGphakqbY4UX2IOgOE8Da0yTpuolAtvHg6NL6.device loaded active plugged /dev/disk/by-id/dm-uuid-LVM-vtt62H07Vx8I5nL9HB9qukE8AelCGphakqbY4UX2IOgOE8Da0yTpuolAtvHg6NL6
dev-dm\x2d0.device                                                                                             loaded active plugged /dev/dm-0
dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d1.device                                                               loaded active plugged /dev/mapper/test-test--a--1
dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d2.device                                                               loaded active plugged /dev/mapper/test-test--a--2
dev-test-test\x2da\x2d1.device                                                                                 loaded active plugged /dev/test/test-a-1
dev-test-test\x2da\x2d2.device                                                                                 loaded active plugged /dev/test/test-a-2
sys-devices-virtual-block-dm\x2d0.device                                                                       loaded active plugged /sys/devices/virtual/block/dm-0
```

At this point, deleting the parent device (e.g. with `lvremove /dev/test/test-a-2`) will still result in all the associated devlinks device units being correctly removed (including the stale ones).

However, if a daemon-reload is issued before removing the parent device, the stale devlinks devices lose their association with the parent:

```
$ systemctl daemon-reload
```

```
dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d1.device                                          loaded active plugged dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d1.device
dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d2.device                                          loaded active plugged /dev/disk/by-id/dm-name-test-test--a--2
dev-disk-by\x2did-dm\x2duuid\x2dLVM\x2dvtt62H07Vx8I5nL9HB9qukE8AelCGphakqbY4UX2IOgOE8Da0yTpuolAtvHg6NL6.device loaded active plugged /dev/disk/by-id/dm-uuid-LVM-vtt62H07Vx8I5nL9HB9qukE8AelCGphakqbY4UX2IOgOE8Da0yTpuolAtvHg6NL6
dev-dm\x2d0.device                                                                                             loaded active plugged /dev/dm-0
dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d1.device                                                               loaded active plugged dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d1.device
dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d2.device                                                               loaded active plugged /dev/mapper/test-test--a--2
dev-test-test\x2da\x2d1.device                                                                                 loaded active plugged dev-test-test\x2da\x2d1.device
dev-test-test\x2da\x2d2.device                                                                                 loaded active plugged /dev/test/test-a-2
sys-devices-virtual-block-dm\x2d0.device                                                                       loaded active plugged /sys/devices/virtual/block/dm-0
```

Removing the parent device then leaves those dangling units behind:

```
$ lvremove /dev/test/test-a-2
Do you really want to remove active logical volume test/test-a-2? [y/n]: y
  Logical volume "test-a-2" successfully removed.
```

```
dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d1.device                                  loaded active plugged dev-disk-by\x2did-dm\x2dname\x2dtest\x2dtest\x2d\x2da\x2d\x2d1.device
dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d1.device                                                       loaded active plugged dev-mapper-test\x2dtest\x2d\x2da\x2d\x2d1.device
dev-test-test\x2da\x2d1.device                                                                         loaded active plugged dev-test-test\x2da\x2d1.device
```

The dangling units survive a daemon-reexec. Only a reboot will clean them up.

_No response_

yuwata:
Fixes #24518.

### DIFF
--- a/test/units/testsuite-17.08.sh
+++ b/test/units/testsuite-17.08.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
+set -o pipefail
+
+# shellcheck source=test/units/assert.sh
+. "$(dirname "$0")"/assert.sh
+
+# This is a test for issue #24518.
+
+mkdir -p /run/udev/rules.d/
+cat >/run/udev/rules.d/50-testsuite.rules <<EOF
+SUBSYSTEM=="mem", KERNEL=="null", OPTIONS="log_level=debug", TAG+="systemd"
+SUBSYSTEM=="mem", KERNEL=="null", ACTION=="add",    SYMLINK+="test/symlink-to-null-on-add",    ENV{SYSTEMD_ALIAS}+="/sys/test/alias-to-null-on-add"
+SUBSYSTEM=="mem", KERNEL=="null", ACTION=="change", SYMLINK+="test/symlink-to-null-on-change", ENV{SYSTEMD_ALIAS}+="/sys/test/alias-to-null-on-change"
+EOF
+
+udevadm control --reload
+
+udevadm trigger --settle --action add /dev/null
+for ((i = 0; i < 20; i++)); do
+    ((i == 0)) || sleep .5
+
+    (
+        systemctl -q is-active /dev/test/symlink-to-null-on-add
+        ! systemctl -q is-active /dev/test/symlink-to-null-on-change
+        systemctl -q is-active /sys/test/alias-to-null-on-add
+        ! systemctl -q is-active /sys/test/alias-to-null-on-change
+    ) && break
+done
+assert_rc 0 systemctl -q is-active /dev/test/symlink-to-null-on-add
+assert_rc 3 systemctl -q is-active /dev/test/symlink-to-null-on-change
+assert_rc 0 systemctl -q is-active /sys/test/alias-to-null-on-add
+assert_rc 3 systemctl -q is-active /sys/test/alias-to-null-on-change
+
+udevadm trigger --settle --action change /dev/null
+for ((i = 0; i < 20; i++)); do
+    ((i == 0)) || sleep .5
+
+    (
+        ! systemctl -q is-active /dev/test/symlink-to-null-on-add
+        systemctl -q is-active /dev/test/symlink-to-null-on-change
+        ! systemctl -q is-active /sys/test/alias-to-null-on-add
+        systemctl -q is-active /sys/test/alias-to-null-on-change
+    ) && break
+done
+assert_rc 3 systemctl -q is-active /dev/test/symlink-to-null-on-add
+assert_rc 0 systemctl -q is-active /dev/test/symlink-to-null-on-change
+assert_rc 3 systemctl -q is-active /sys/test/alias-to-null-on-add
+assert_rc 0 systemctl -q is-active /sys/test/alias-to-null-on-change
+
+udevadm trigger --settle --action add /dev/null
+for ((i = 0; i < 20; i++)); do
+    ((i == 0)) || sleep .5
+
+    (
+        systemctl -q is-active /dev/test/symlink-to-null-on-add
+        ! systemctl -q is-active /dev/test/symlink-to-null-on-change
+        systemctl -q is-active /sys/test/alias-to-null-on-add
+        ! systemctl -q is-active /sys/test/alias-to-null-on-change
+    ) && break
+done
+assert_rc 0 systemctl -q is-active /dev/test/symlink-to-null-on-add
+assert_rc 3 systemctl -q is-active /dev/test/symlink-to-null-on-change
+assert_rc 0 systemctl -q is-active /sys/test/alias-to-null-on-add
+assert_rc 3 systemctl -q is-active /sys/test/alias-to-null-on-change
+
+# cleanup
+rm -f /run/udev/rules.d/50-testsuite.rules
+udevadm control --reload
+
+exit 0


test/units/testsuite-17.08.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -ex
+set -o pipefail
+
+# shellcheck source=test/units/assert.sh
+. "$(dirname "$0")"/assert.sh
```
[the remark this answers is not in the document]
jamacku: Updating to `differential-shellcheck@v3` should resolve this kind of false positives.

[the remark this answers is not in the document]
yuwata: Thank you! #24533.

src/core/device.c
```diff
@@ -32,6 +32,27 @@ static const UnitActiveState state_translation_table[_DEVICE_STATE_MAX] = {
...
+        _cleanup_free_ char *e = NULL;
+        Unit *u;
+        int r;
+
+        assert(m);
+        assert(path);
+
+        r = unit_name_from_path(path, ".device", &e);
+        if (r < 0)
+                return log_debug_errno(r, "Failed to generate unit name from device path \"%s\", ignoring: %m", path);
```
keszybz: Hmm, this logs "ignoring", but it actually propagates the error. I think this inner function shouldn't do this, because if it'll be used in a different place, the message might be wrong.

Also we log here, but not from `manager_get_unit()` below…

src/core/device.c
```diff
@@ -1063,12 +1070,7 @@ static void device_propagate_reload_by_name(Manager *m, const char *path) {
         if (!u)
                 return;
```
keszybz: This is another place where `device_from_path()` could be used.

src/core/device.c
```diff
@@ -820,6 +805,12 @@ static int device_setup_devlink_units_on_remove(Manager *m, sd_device *dev, char
                 if (!path_startswith(d->path, "/dev/"))
                         continue;
 
+                if (devname && path_equal(d->path, devname))
+                        continue; /* This is not devlink, but real device node. */
+
+                if (device_has_devlink(dev, d->path))
+                        continue; /* The devlink is already processed in the above loop. */
```
keszybz: "is processed" → "was processed"

src/core/device.c
```diff
@@ -820,6 +805,12 @@ static int device_setup_devlink_units_on_remove(Manager *m, sd_device *dev, char
                 if (!path_startswith(d->path, "/dev/"))
                         continue;
 
+                if (devname && path_equal(d->path, devname))
+                        continue; /* This is not devlink, but real device node. */
```
keszybz: "not devlink …" → "not a devlink, but a real device node"

yuwata:
@keszybz Thank you for the comments. All comments are addressed. PTAL.

keszybz:
Looks reasonable. But needs a rebase now.

yuwata:
@keszybz Rebased. PTAL.

src/core/device.c
```diff
@@ -32,6 +32,27 @@ static const UnitActiveState state_translation_table[_DEVICE_STATE_MAX] = {
...
+        _cleanup_free_ char *e = NULL;
+        Unit *u;
+        int r;
+
+        assert(m);
+        assert(path);
+
+        r = unit_name_from_path(path, ".device", &e);
+        if (r < 0)
+                return r;
```
daandemeyer, line 45: Can we keep the debug log here that was removed from the two call sites?

DaanDeMeyer:
Can you moves the "Fixes" from the third commit to the fourth commit? I was extremely confused trying to figure out how the third commit could possibly fix the issue until I moved on to the fourth commit and noticed that that was the actual fix.

src/core/device.c
```diff
@@ -811,68 +875,65 @@ static int device_setup_devlink_units(Manager *m, sd_device *dev, char ***ret_re
...
-        if (device_setup_unit(m, dev, sysfs, /* main = */ true) < 0)
-                return;
+        r = sd_device_get_syspath(dev, &syspath);
+        if (r < 0)
+                return log_device_debug_errno(dev, r, "Couldn't get syspath from device, ignoring: %m");
 
-        /* Add an additional unit for the device node */
-        if (sd_device_get_devname(dev, &dn) >= 0)
-                (void) device_setup_unit(m, dev, dn, /* main = */ false);
+        if (device_for_action(dev, SD_DEVICE_REMOVE))
```
daandemeyer, line 864: Can we return early if this is the case? The rest of this function seems to only be adding more units, but are we ever going to add more units on a device remove?

src/core/device.c
```diff
@@ -697,6 +697,12 @@ static int device_setup_unit(Manager *m, sd_device *dev, const char *path, bool
         if (dev && device_is_bound_by_mounts(DEVICE(u), dev))
                 device_upgrade_mount_deps(u);
 
+        if (ready_units) {
+                r = set_ensure_put(ready_units, NULL, DEVICE(u));
+                if (r < 0)
+                        return log_unit_error_errno(u, r, "Failed to store unit: %m");
+        }
```
daandemeyer: Should we just call `ready_units` something more generic here, like `set`? We can pass any set to `device_setup_unit()`, so it seems weird to give it such a specific name here.

src/core/device.c
```diff
@@ -735,37 +741,112 @@ static bool device_is_ready(sd_device *dev) {
...
+                if (r >= 0) {
+                        r = strv_split_full(&aliases, s, NULL, EXTRACT_UNQUOTE);
+                        if (r < 0)
+                                log_device_warning_errno(dev, r, "Failed to parse SYSTEMD_ALIAS property, ignoring: %m");
+                }
+        }
+
+        STRV_FOREACH(alias, aliases) {
+                if (!path_is_absolute(*alias)) {
+                        log_device_warning(dev, "SYSTEMD_ALIAS is not an absolute path, ignoring: %s", *alias);
```
daandemeyer: This was already the case, but the warning message here seems wrong, this is not about SYSTEMD_ALIAS in its entirety, but a single alias in it.

src/core/device.c
```diff
@@ -735,37 +741,112 @@ static bool device_is_ready(sd_device *dev) {
...
+                        continue;
+                }
+
+                if (!path_is_normalized(*alias)) {
+                        log_device_warning(dev, "SYSTEMD_ALIAS is not a normalized path, ignoring: %s", *alias);
```
daandemeyer: Ditto

src/core/device.c
```diff
@@ -1109,45 +1135,38 @@ static int device_dispatch_io(sd_device_monitor *monitor, sd_device *dev, void *
...
-        }
-
-        if (ready || !strv_isempty(ready_devlinks))
+        if (!set_isempty(ready_units))
                 manager_dispatch_load_queue(m);
 
         if (action == SD_DEVICE_REMOVE)
                 /* If we get notified that a device was removed by udev, then it's completely gone, hence
                  * unset all found bits */
                 device_update_found_by_sysfs(m, sysfs, DEVICE_NOT_FOUND, DEVICE_FOUND_MASK);
```
daandemeyer: This only handles alias units right? Where's the code that does this for devlink units?

DaanDeMeyer (commented):
Some questions and suggestions. This one is hard to review because I lack knowledge about this part of systemd.

src/core/device.c
yuwata: No. That is one of the important part of this PR.

Even if a device is removed, a devlink previously points to the device may be still active, as now it may points to another device. Hence, we need to update the corresponding .device unit build from the devlink. That's done at the end of this function.

src/core/device.c
yuwata: No, this works for all (main, devnode, devlink, and alias) .device units still points to the removed device.

src/core/device.c
yuwata: I dropped the log in response to https://github.com/systemd/systemd/pull/24522#discussion_r962896850.
I will consider to add log message in caller side.

yuwata: I noticed all caller ignore the error.

src/core/device.c
```diff
@@ -735,37 +741,105 @@ static bool device_is_ready(sd_device *dev) {
...
+                (void) device_setup_unit(m, dev, *alias, /* main = */ false, ready_units);
+        }
+
+        l = hashmap_get(m->devices_by_sysfs, syspath);
+        LIST_FOREACH(same_sysfs, d, l) {
+                if (!d->path)
+                        continue;
+
+                if (path_startswith(d->path, "/dev/"))
+                        continue;
```
daandemeyer: Is this supposed to skip devlinks? If not, how do you make sure this loop skips devlinks and the same loop in device_setup_devlink_units() skips aliases? Because wouldn't you otherwise maybe end up adding devlinks to `not_ready_units` here and then add them to `ready_units` in the other loop again.

Should we have a sanity check somewhere that there's no unit in both `ready_units` and in `not_ready_units`?

yuwata: OK. Done. Please see the last three commits for core/device.c.

src/core/device.c
```diff
@@ -1103,54 +1143,42 @@ static int device_dispatch_io(sd_device_monitor *monitor, sd_device *dev, void *
...
-        else if (ready)
-                /* The device is found now, set the udev found bit */
-                device_update_found_by_sysfs(m, sysfs, DEVICE_FOUND_UDEV, DEVICE_FOUND_UDEV);
-        else
-                /* The device is nominally around, but not ready for us. Hence unset the udev bit, but leave
-                 * the rest around. */
-                device_update_found_by_sysfs(m, sysfs, DEVICE_NOT_FOUND, DEVICE_FOUND_UDEV);
+                 * unset all found bits. Note this affects all .device units still point to the removed
+                 * device. */
+                device_update_found_by_sysfs(m, sysfs, DEVICE_NOT_FOUND, DEVICE_FOUND_MASK);
```
daandemeyer, line 1138: If we get into this codepath, does it still make sense to execute the rest? If I understand correctly, if we get here, all the aliases and the devlinks will have all found bits unset, yet right after that we start changing them again for `ready` and `not_ready` units. If we get here, shouldn't all aliases/devlinks be "not found"?

yuwata: Multiple devices may request a same devlink. Hence, if one of the device is removed, the associated devlink may be still active. In that case, the syspath field of the devlink unit is updated, as now it is associated to another device, and the devlink unit is actually in ready_units.

src/core/device.c
```diff
@@ -718,153 +741,209 @@ static bool device_is_ready(sd_device *dev) {
...
+        assert(not_ready_units);
 
         r = sd_device_get_syspath(dev, &syspath);
         if (r < 0)
                 return r;
 
+        (void) sd_device_get_devname(dev, &devname);
+
+        FOREACH_DEVICE_DEVLINK(dev, devlink) {
+                if (PATH_STARTSWITH_SET(devlink, "/dev/block/", "/dev/char/"))
```
daandemeyer, line 784: A comment here on why we skip these would be great

test/units/testsuite-17.08.sh
```diff
@@ -0,0 +1,72 @@
...
+
+udevadm control --reload
+
+udevadm trigger --settle --action add /dev/null
+for ((i = 0; i < 20; i++)); do
+    ((i == 0)) || sleep .5
+
+    (
+        systemctl -q is-active /dev/test/symlink-to-null-on-add
+        ! systemctl -q is-active /dev/test/symlink-to-null-on-changel
```
daandemeyer: Typo here? ("/dev/test/symlink-to-null-on-changel")

test/units/testsuite-17.08.sh
```diff
@@ -0,0 +1,72 @@
...
+assert_rc 0 systemctl -q is-active /sys/test/alias-to-null-on-add
+assert_rc 3 systemctl -q is-active /sys/test/alias-to-null-on-change
+
+udevadm trigger --settle --action change /dev/null
+for ((i = 0; i < 20; i++)); do
+    ((i == 0)) || sleep .5
+
+    (
+        ! systemctl -q is-active /dev/test/symlink-to-null-on-add
+        systemctl -q is-active /dev/test/symlink-to-null-on-changel
```
daandemeyer: Ditto

test/units/testsuite-17.08.sh
```diff
@@ -0,0 +1,72 @@
...
+assert_rc 3 systemctl -q is-active /sys/test/alias-to-null-on-add
+assert_rc 0 systemctl -q is-active /sys/test/alias-to-null-on-change
+
+udevadm trigger --settle --action add /dev/null
+for ((i = 0; i < 20; i++)); do
+    ((i == 0)) || sleep .5
+
+    (
+        systemctl -q is-active /dev/test/symlink-to-null-on-add
+        ! systemctl -q is-active /dev/test/symlink-to-null-on-changel
```
daandemeyer: Ditto

test/units/testsuite-64.sh
```diff
@@ -83,6 +83,93 @@ helper_check_udev_watch() {(
...
+
+    if [[ ! -b "$path" && ! -c "$path" ]]; then
+        [[ "$log_level" == 1 ]] && echo >&2 "ERROR: invalid file type $path"
+        return 1
+    fi
+
+    read -r -a links < <(udevadm info -q symlink "$syspath" 2>/dev/null)
+    for link in "${links[@]}"; do
+        if [[ "/dev/$link" == "$path" ]]; then # DEVLINKS= given by -q symlink are relative to /dev
+            return 0
```
daandemeyer, line 117: Don't we need to check `SYSTEMD_ALIAS` as well before returning?

DaanDeMeyer (changes requested):
Found a few more things

test/units/testsuite-64.sh
yuwata: This function check if there exists matching devlink _or_ alias. If found here, then it is not necessary to search in aliases.

src/core/device.c
yuwata: Added.

yuwata:
Sad NEWS. 
LVM sets `ENV{SYSTEMD_ALIAS}="/dev/block/$major:$minor"` to forcibly creates the 'devlink' units.
So, we cannot judge where a virtual unit comes from, SYMLINK= or ENV{SYSTEMD_ALIAS}=.

Now, device_setup_devlink_units() and device_setup_alias_units() are merged into device_setup_extra_units().

PTAL.

src/core/device.c
```diff
@@ -718,153 +741,176 @@ static bool device_is_ready(sd_device *dev) {
...
+                return log_device_debug_errno(dev, r, "Couldn't get syspath from device, ignoring: %m");
 
-        /* Add an additional unit for the device node */
-        if (sd_device_get_devname(dev, &dn) >= 0)
-                (void) device_setup_unit(m, dev, dn, /* main = */ false);
+        /* First, process the main (that is, points to the syspath) and (real, not symlink) devnode units. */
+        if (device_for_action(dev, SD_DEVICE_REMOVE))
+                /* If the device is removed, the main and devnode units units will be removed by
+                 * device_update_found_by_sysfs() in device_dispatch_io(). Hence, it is not necessary to
+                 * store them to not_ready_units, and we have nothing to do here.
```
daandemeyer, line 867: `device_update_found_by_sysfs()` is not called in `device_enumerate()`, but I assume that this condition can only be triggered from `device_dispatch_io()` and not `device_enumerate()`?

DaanDeMeyer (approved):

keszybz:
focal-amd64 failed in "upstream": TEST-38-FREEZER:                    FAIL     ( 98 s), TEST-29-PORTABLE:                   FAIL     (217 s). The system seems to be very slow, so I hope it's some timeout issue, unrelated to this PR.

src/core/device.c
yuwata: No removed devices can be enumerated. Hence, that's no problem.

PR #24525, opened 2022-09-01
man: remove reference to mkinitrd

aafeijoo-suse:
Remove the only reference to `mkinitrd` present in the man pages.

Although there are more initramfs generators available, this reference is replaced with `dracut`, as it is the only one that appears elsewhere in the docs.


poettering:
This was added in dfc6109fcd7cd8987ca77326eef25c0deb25abd9, and the background was fedora, so I guess it's surprising to mention mkinitrd...

bluca (approved):

Issue #24431, opened 2022-08-24
systemd-gpt-auto-generator deletes probed partitions


msekletar:
### systemd version the issue has been seen with

251

### Used distribution

Fedora

### CPU architectures issue was seen on

x86_64

### Component

systemd-gpt-auto-generator

### Expected behaviour you didn't see

systemd-gpt-auto-generator doesn't delete probed partitions.

### Unexpected behaviour you saw

In 75d7e04eb46 we have introduced ability to relinquish partitions that were potentially probed while dissecting images. Relinquished partitions are later deleted before the underlying device that contains the image is closed (which would trigger asynchronous deletion of these partitions).

We have added call to `dissected_image_relinquish()` also to `gpt-auto-generator` and that causes generator to deleted the discovered partitions. On SELinux enabled systems this also causes following AVC message (deleting partitions requires CAP_SYS_ADMIN),

```
type=AVC msg=audit(08/24/2022 06:27:08.956:927) : avc:  denied  { sys_admin } for  pid=44801 comm=systemd-gpt-aut capability=sys_admin  scontext=system_u:system_r:systemd_gpt_generator_t:s0 tcontext=system_u:system_r:systemd_gpt_generator_t:s0 tclass=capability permissive=0 

```

It seems that current behaviour is not optimal because deleting these partitions seems pointless as they would need to be recreated anyway.

### Steps to reproduce the problem

1.  Run SELinux enabled Fedora and make use of systemd-gpt-auto-generator.

msekletar:
@poettering @DaanDeMeyer just following up on the issue that I've mentioned on the call 2 weeks ago.

poettering:
Fix in → #24530

QuintinHill:
This is causing https://bugzilla.redhat.com/show_bug.cgi?id=2141998 in Fedora 37.  In https://github.com/fedora-selinux/selinux-policy/pull/1468 there is a reluctance to allow sys_admin.  @keszybz can the commit fixing this be backported to 251-stable please.

PR #24530, opened 2022-09-01
a variety of loopback block device fixes

poettering:


poettering:
Inspired by #24482

src/shared/loop-util.c
```diff
@@ -440,6 +457,7 @@ static int loop_device_make_internal(
                         };
 
                         *ret = d;
+
```
daandemeyer: Spurious whitespace?

src/shared/loop-util.c
```diff
@@ -928,12 +978,29 @@ int loop_device_refresh_size(LoopDevice *d, uint64_t offset, uint64_t size) {
...
+        /* When unlocking just close the lock fd */
+        if ((operation & ~LOCK_NB) == LOCK_UN) {
+                d->lock_fd = safe_close(d->lock_fd);
+                return 0;
+        }
+
+        /* If we had no lock fd so far, create one and lock it right-away */
+        if (d->lock_fd < 0) {
+                if (d->fd < 0)
+                        return -EBADF;
```
daandemeyer: Shouldn't we just assert here? Calling this function with a loop device without an fd seems like a programming error

src/shared/loop-util.c
```diff
@@ -339,6 +339,9 @@ static int loop_configure(
         return 0;
 
 fail:
+        /* Close the lock fd explicitly before clearing the loopback block device, since an additional open
+         * fd would block the clearing to succeed */
+        lock_fd = safe_close(lock_fd);
```
daandemeyer, line 346: I'd merge this in the commit that introduces `lock_fd`

src/dissect/dissect.c
```diff
@@ -913,18 +913,6 @@ static int action_umount(const char *path) {
...
-         * command exits. */
-        r = block_device_remove_all_partitions(d->fd);
-        if (r == -EBUSY) {
-                log_error_errno(r, "One or more partitions of '%s' are busy, ignoring", devname);
-                r = 0;
-        }
-        if (r < 0)
-                log_error_errno(r, "Failed to remove one or more partitions of '%s': %m", devname);
-
-
```
daandemeyer: I think if we drop this we can simplify and remove the `k` variable as well and just use `r` everywhere

src/shared/loop-util.c
```diff
@@ -706,9 +708,21 @@ LoopDevice* loop_device_unref(LoopDevice *d) {
...
 
+                        if (flock(d->fd, LOCK_EX) < 0)
+                                log_debug_errno(errno, "Failed to lock loop block device, ignoring: %m");
+
+                        r = block_device_remove_all_partitions(d->fd);
+                        if (r < 0)
+                                log_debug_errno(r, "Failed to remove partitions of loopback block device, ignoring: %m");
+
+                        if (ioctl(d->fd, LOOP_CLR_FD) < 0)
+                                log_debug_errno(errno, "Failed to clear loop device, ignoring: %m");
```
daandemeyer: Shouldn't this commit be moved after the `lock_fd` commit, and then not do any extra locking here, but simply close `lock_fd` after removing all the partitions?

daandemeyer: (And maybe upgrade to exclusive lock in case `lock_fd` is a shared lock)

src/shared/loop-util.c
```diff
@@ -702,6 +738,8 @@ LoopDevice* loop_device_unref(LoopDevice *d) {
         if (!d)
                 return NULL;
 
+        d->lock_fd = safe_close(d->lock_fd);
```
yuwata, line 717: I think this should be moved to the end of this function, at least, after removing the partitions.

yuwata (changes requested):

src/shared/loop-util.c
poettering: Oh, that was added back in 738f29cb53b457ca0c17885f119de5bc1f10dead, two years ago. Can't fix up history ;-)

src/shared/loop-util.c
poettering: So the loopback allocation function takes the /dev/loop-control lock first, and the block device lock second. We have to follow the same order here, to avoid ABBA locking issues. Since we want to take the /dev/loop-control lock here (a few lines further down) we need to release the block device lock first.

Moreover, we intend to issue LOOP_CLR_FD here on the primary fd of the fd (as opposed to the locking fd). That ioctl can only succeed if it is called on the single open fd to the device. Hence we need to close the lock fd first, to ensure we only retain a single fd to the device at the moment we issue the clear ioctl.

I'll add a comment about this.

src/shared/loop-util.c
poettering: That cannot work, because LOOP_CLR_FD would fail if we had more than one fd open to the device (lock_fd is a second fd open after all).

Moreover: ABBA issues: we take two locks here: on the /dev/loop-control device and on the actual block devices. We do that during allocation since a long time. But the lock protocol is to take the control lock first, and the block lock second. We have to follow the same order here to avoid ABBA locking. Given that the locking fd might be open already we must close it hence before locking the control device. 

I'll add comments about that.

src/shared/loop-util.c
yuwata: Hm, how about to close `d->fd` first if `d->lock_fd >= 0`?? Then, there should be only fd we have.
IIUC `block_device_remove_all_partitions()` and `LOOP_CLR_FD` also works fine for `lock_fd`, right?
```c
        if (d->fd >= 0) {
                /* Implicitly sync the device, since otherwise in-flight blocks might not get written */
                if (fsync(d->fd) < 0)
                        log_debug_errno(errno, "Failed to sync loop block device, ignoring: %m");
                if (d->nr >= 0 && !d->relinquished) {
                        /* We are supposed to clear the loopback device. Let's do this synchronously: lock
                         * the device, manually remove all partitions and then clear it. This should ensure
                         * udev doesn't concurrently access the devices, and we can be reasonably sure that
                         * once we are done here the device is cleared and all its partition children
                         * removed. */

                        if (d->lock_fd >= 0) {
                                safe_close(d->fd);
                                d->fd = TAKE_FD(d->lock_fd);
                        } else if (flock(d->fd, LOCK_EX) < 0)
                                log_debug_errno(errno, "Failed to lock loop block device, ignoring: %m");

                        r = block_device_remove_all_partitions(d->fd);
                        if (r < 0)
                                log_debug_errno(r, "Failed to remove partitions of loopback block device, ignoring: %m");
                        if (ioctl(d->fd, LOOP_CLR_FD) < 0)
                                log_debug_errno(errno, "Failed to clear loop device, ignoring: %m");
                }
                safe_close(d->fd);
        }
```

src/shared/loop-util.c
```diff
@@ -706,9 +708,21 @@ LoopDevice* loop_device_unref(LoopDevice *d) {
                         log_debug_errno(errno, "Failed to sync loop block device, ignoring: %m");
 
                 if (d->nr >= 0 && !d->relinquished) {
```
yuwata: BTW, not a matter, but why this code block is conditionalized with `d->nr >= 0`, instead of `d->fd >= 0`??

src/shared/loop-util.c
```diff
@@ -699,13 +699,26 @@ int loop_device_make_by_path(
...
 
         d->lock_fd = safe_close(d->lock_fd);
 
+        /* Let's open the control device early, and lock it, so that we can release our block device and
+         * delete it in a synchronized fashion, and allocators won't needlessly see the block device as free
+         * while we are about to delete it. */
+        if (d->nr >= 0 && !d->relinquished) {
+                control = open("/dev/loop-control", O_RDWR|O_CLOEXEC|O_NOCTTY|O_NONBLOCK);
+                if (control < 0)
+                        log_debug_errno(errno, "Failed to open loop control device, cannot remove loop device %s: %m", strna(d->node));
```
bluca: is this missing a `return` or an `ignoring` in the log?

bluca:
How does this interact with verity? If I understand correctly, beforehand this was removing the partitions via dissect, but now it won't anymore, as when using verity there's no loopback right? The device is opened via the device-mapper driver instead in /dev/mapper?

src/shared/loop-util.c
poettering: to simplify things like `systemd-nspawn --image=/dev/sdc` we actually allow LoopDevice objects to be created from non-loopback devices too. i.e. if you point loop_device_make() to an existing regular block device, then we'll just use that as is, and not create a loopback device from the thing that is anyway a block device.

`d->nr` is set to the loopback device index (i.e. nr=4 for /dev/loop4), but when wrapping a block device that actually is not really a loopback bock device, we set it to -1.

I will add some comments about this.

poettering: (and when we wrap such a non-loopback block device, then we of course don't have to release it, it should just stay around, after all it's not managed by us then)

src/shared/loop-util.c
```diff
@@ -117,13 +117,26 @@ static int device_has_block_children(sd_device *d) {
         return !!sd_device_enumerator_get_device_first(e);
 }
 
+static int open_lock_fd(int primary_fd, int operation) {
+        int lock_fd;
+
+        lock_fd = fd_reopen(primary_fd, O_RDWR|O_CLOEXEC|O_NONBLOCK|O_NOCTTY);
```
yuwata, line 125: Missing assertion for `primary_fd`.

BTW, maybe it is better to move this to fd-util.c?

poettering:
> How does this interact with verity? If I understand correctly, beforehand this was removing the partitions via dissect, but now it won't anymore, as when using verity there's no loopback right? The device is opened via the device-mapper driver instead in /dev/mapper?

Hmm? Can't parse this?

The dissection logic can be used on any form of block device. We often use it on loopback block devices we allocate ourselves (for example, portable services, RootImage= or systemd-nspawn --image= do this), but we also use it on existing real block devices (most importantly, systemd-gpt-auto-generator does this for the device backing the root fs, to find the relevant partitions, but it is also used if you use `RootImage=/dev/sdb` or `systemd-nspawn --image=/dev/sdb`).

Since 75d7e04eb4662a814c26010d447eed8a862f5ec1 (i.e. very recently) the dissect code was deleting all partitions it didnt really want to use via BLKPG on any block device it was dissecting. This might be OK to do if dissecting a block device we ourselves just allocated. But it's not OK to do if we are dissecting a real, non-loopback block device, because we don't own those devices. Which lead to bugs like #24431. 

So this PR simply drops that part again: dissection won't ever delete partitions anymore. Instead destroying a loopback block device via loop_device_unref() might delete remaining partitions synchronously first. But that is safe, since in that case the loopback device belongs to us, and we won't delete any partitions on real disks.

verity doesn#t matter here, as dissection becomes long before verity is set up. i.e. the usual order is:

1. make loopback block device from regular file, if source is regular file.
2. dissect the loopback block device we just created, or the source block device if it is one
3. set up dm-verity + dm-crypt on relevant partitions
4. mount the file systems inside
5. unmount the file systems
6. destroy dm-verity/dm-crypt
7. detach loopback device (if there is one)

Previously, we'd end up deleting partitions as part of step 2. That's gone now. We'll now only delete them in step 7, if there are any.

Didn't grok your question but I hope this nonetheless answers your question?

src/shared/loop-util.c
poettering: We can't because of the mentioned ABBA issue:

1. We take a lock on /dev/loop-control
2. We take a lock on the actual loopback block device
3. We then remove all partitions + issue LOOP_CLR_FD
4. We release both locks

If we do things that way, and allocation works like this:

1. We take a lock on /dev/loop-control
2. We ask the control device for next available loopback device, and open and lock it
3. We delete any remaining partitions on it
4. We release the /dev/loop-control device, and possibly the loopback device lock

Then we safely avoided the race that between asking which device is available next, and actually allocating it and setting it up some other participant also takes it and then maybe releases it, or is still in progress of releasing it after having used it.

But this requires us to follow a strict locking protocol: to avoid ABBA deadlocks, we always must acquire /dev/loop-control first, and only the block device. Now, when loop_device_unref() is entered there might be a lock in effect on the loopback device, but there is not on /dev/loop-control. But we want one on that, and given the lock order constraints we must thus must give up the block device lock first, then acquire the control lock and only then acquire the block device lock again.

if we don't do it that way (and instead do as you suggest), we might deadlock: because a current owner which locked the block device might ask for the /dev/loop-control lock, while another process might just try to allocate it, thus already acquire /dev/loop-control and then goes on tries to lock the block device. Now each of the two processes has a lock on one device and tries to acquire one on the other, but can't because the other one has it. Deadlock...

So we fix this by giving up the lock temporarily here, and re-acquire it in the order our lock protocol dictates: get the /dev/loop-control lock first, then acquire the loopback block device. And never the other way round.

yuwata: Hm, interesting. Thank you for the explanation.

src/shared/loop-util.c
yuwata: Oh, so LoopDevice may not be related to a loopback block device... Ugh.

yuwata (approved):
LGTM.

bluca:
> Previously, we'd end up deleting partitions as part of step 2. That's gone now. We'll now only delete them in step 7, if there are any.
> 
> Didn't grok your question but I hope this nonetheless answers your question?

It does - before it did apply to non-loopback but it was unintentional, understood

bluca (approved):

PR #21316, opened 2021-11-11
ci: pin labeler

evverx:
Turns out GHActions where `pull_request_target` is used are capable
of pwning repositories: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

labeler doesn't check out the source code or build anything so
it's safe in its current form but to avoid surprises let's just pin
it to the latest version. It's annoying to manage dependencies like this
manually so additionally dependabot.yml is introduced to make it
easier to keep GHActions up to date more or less automatically:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot

@mrc0mmand could you take a look? Dependabot will create PRs like https://github.com/evverx/systemd/pull/10, https://github.com/evverx/systemd/pull/12, https://github.com/evverx/systemd/pull/11 once the PR is merged.

PR #24533, opened 2022-09-01
github: update differential shellcheck to v3

yuwata:
v3 supports external sources. Yey!

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -26,6 +26,6 @@ jobs:
           fetch-depth: 0
 
       - name: Differential ShellCheck
-        uses: redhat-plumbers-in-action/differential-shellcheck@60360c0fb283149ed03ad16b66257a967c3e5231
+        uses: redhat-plumbers-in-action/differential-shellcheck@a14889568f6210b361eb29e16f3b07f512fca846
         with:
           token: ${{ secrets.GITHUB_TOKEN }}


bluca (approved):

.github/workflows/differential-shellcheck.yml
```diff
@@ -26,6 +26,6 @@ jobs:
           fetch-depth: 0
 
       - name: Differential ShellCheck
-        uses: redhat-plumbers-in-action/differential-shellcheck@60360c0fb283149ed03ad16b66257a967c3e5231
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
```
jamacku: Please use `a14889568f6210b361eb29e16f3b07f512fca846` to comply with repository convention, where actions are pinned to SHA1.

https://github.com/redhat-plumbers-in-action/differential-shellcheck/commit/a14889568f6210b361eb29e16f3b07f512fca846

jamacku (changes requested):

.github/workflows/differential-shellcheck.yml
mrc0mmand: Could you pin this to the specific SHA instead of a tag - `a14889568f6210b361eb29e16f3b07f512fca846` appears to be the SHA for `v3.0.1` (see https://github.com/systemd/systemd/pull/21316 for reasoning).

yuwata:
@jamacku and @mrc0mmand Thank you for the suggestion. Updated. PTAL.

jamacku (approved):
LGTM

mrc0mmand (approved):

PR #24534, opened 2022-09-01
sysusers: make sp_lstchg shadow field reproducible

josch:
Make the sp_lstchg field respect the environment variable `SOURCE_DATE_EPOCH` as it is also done by shadow since https://github.com/shadow-maint/shadow/pull/146

### DIFF
--- a/docs/ENVIRONMENT.md
+++ b/docs/ENVIRONMENT.md
@@ -310,6 +310,16 @@ All tools:
   subvolumes if the backing filesystem supports them. If set to `0`, these
   lines will always create directories.
 
+`systemd-sysusers`
+
+* `SOURCE_DATE_EPOCH` — if unset, the field of the date of last password change
+  in `/etc/shadow` will be the number of days from Jan 1, 1970 00:00 UTC until
+  today. If SOURCE_DATE_EPOCH is set to a valid UNIX epoch value in seconds,
+  then the field will be the number of days until that time instead. This is to
+  support creating bit-by-bit reproducible system images by choosing a
+  reproducible value for the field of the date of last password change in
+  `/etc/shadow`. See: https://reproducible-builds.org/specs/source-date-epoch/
+
 `systemd-sysv-generator`:
 
 * `$SYSTEMD_SYSVINIT_PATH` — Controls where `systemd-sysv-generator` looks for
--- a/src/basic/env-util.c
+++ b/src/basic/env-util.c
@@ -776,6 +776,18 @@ int getenv_bool_secure(const char *p) {
         return parse_boolean(e);
 }
 
+int getenv_uint64_secure(const char *p, uint64_t *ret) {
+        const char *e;
+
+        assert(p);
+
+        e = secure_getenv(p);
+        if (!e)
+                return -ENXIO;
+
+        return safe_atou64(e, ret);
+}
+
 int set_unset_env(const char *name, const char *value, bool overwrite) {
         assert(name);
 
--- a/src/sysusers/sysusers.c
+++ b/src/sysusers/sysusers.c
@@ -10,6 +10,7 @@
 #include "creds-util.h"
 #include "def.h"
 #include "dissect-image.h"
+#include "env-util.h"
 #include "fd-util.h"
 #include "fileio.h"
 #include "format-util.h"
@@ -525,6 +526,18 @@ static int write_temporary_passwd(const char *passwd_path, FILE **tmpfile, char 
         return 0;
 }
 
+static usec_t epoch_or_now(void) {
+        uint64_t epoch;
+
+        if (getenv_uint64_secure("SOURCE_DATE_EPOCH", &epoch) >= 0) {
+                if (epoch > UINT64_MAX/USEC_PER_SEC) /* Overflow check */
+                        return USEC_INFINITY;
+                return (usec_t) epoch * USEC_PER_SEC;
+        }
+
+        return now(CLOCK_REALTIME);
+}
+
 static int write_temporary_shadow(const char *shadow_path, FILE **tmpfile, char **tmpfile_path) {
         _cleanup_fclose_ FILE *original = NULL, *shadow = NULL;
         _cleanup_(unlink_and_freep) char *shadow_tmp = NULL;
@@ -545,7 +558,7 @@ static int write_temporary_shadow(const char *shadow_path, FILE **tmpfile, char 
         if (r < 0)
                 return log_debug_errno(r, "Failed to open temporary copy of %s: %m", shadow_path);
 
-        lstchg = (long) (now(CLOCK_REALTIME) / USEC_PER_DAY);
+        lstchg = (long) (epoch_or_now() / USEC_PER_DAY);
 
         original = fopen(shadow_path, "re");
         if (original) {
--- a/src/basic/env-util.h
+++ b/src/basic/env-util.h
@@ -57,6 +57,8 @@ char *strv_env_pairs_get(char **l, const char *name) _pure_;
 int getenv_bool(const char *p);
 int getenv_bool_secure(const char *p);
 
+int getenv_uint64_secure(const char *p, uint64_t *ret);
+
 /* Like setenv, but calls unsetenv if value == NULL. */
 int set_unset_env(const char *name, const char *value, bool overwrite);
 


src/basic/env-util.c
```diff
@@ -776,6 +776,16 @@ int getenv_bool_secure(const char *p) {
         return parse_boolean(e);
 }
 
+int getenv_llu_secure(const char *p, unsigned base, unsigned long long *ret) {
+        const char *e;
+
```
bluca, line 781: `assert(p)`

src/basic/env-util.c
```diff
@@ -776,6 +776,17 @@ int getenv_bool_secure(const char *p) {
...
+        assert(p);
+        e = secure_getenv(p);
```
yuwata: Usually, we insert a empty line between assertion(s) and code:
```c
assert(p);

e = secure_getenv(p);
```

src/basic/time-util.c
```diff
@@ -55,6 +56,20 @@ usec_t now(clockid_t clock_id) {
         return timespec_load(&ts);
 }
 
+usec_t now_reproducible(clockid_t clock_id) {
```
yuwata: Not a big matter, but I'd like to rename this `epoch_or_now()` or so. "reproducible" here sounds strange for me.

src/basic/time-util.c
```diff
@@ -55,6 +56,20 @@ usec_t now(clockid_t clock_id) {
...
+        int r;
+        unsigned long long epoch;
+
+        r = getenv_llu_secure("SOURCE_DATE_EPOCH", 10, &epoch);
+        if (r == 0)
+                return (usec_t) epoch * USEC_PER_SEC;
+
+        assert_se(clock_gettime(map_clock_id(clock_id), &ts) == 0);
+
+        return timespec_load(&ts);
```
yuwata: Why not simply write `return now(clock_id);`??

yuwata (changes requested):

src/basic/time-util.c
josch: Thanks a lot for your very valuable input! I think I resolved all your points accordingly.

src/basic/time-util.c
```diff
@@ -55,6 +56,17 @@ usec_t now(clockid_t clock_id) {
         return timespec_load(&ts);
 }
 
+usec_t epoch_or_now(clockid_t clock_id) {
+        int r;
+        unsigned long long epoch;
+
+        r = getenv_llu_secure("SOURCE_DATE_EPOCH", 10, &epoch);
+        if (r == 0)
```
yuwata: Please use `>= 0` rather than `== 0`, and not necessary to assign a value to `r`:
```c
if (getenv_llu_secure(...) >= 0)
        return (usec_t) epoch * USEC_PER_SEC;
```

yuwata (commented):
Maybe better to mention the change to docs/ENVIRONMENT.md ??

src/basic/env-util.c
```diff
@@ -776,6 +776,18 @@ int getenv_bool_secure(const char *p) {
         return parse_boolean(e);
 }
 
+int getenv_llu_secure(const char *p, unsigned base, unsigned long long *ret) {
```
poettering: llu? That is such a weird type, given it's undefined length. I am pretty sure uint64_t is really the better choice here... i.e. getenv_uint64_secure()...

src/basic/time-util.c
```diff
@@ -55,6 +56,17 @@ usec_t now(clockid_t clock_id) {
...
+                return (usec_t) epoch * USEC_PER_SEC;
+
+        return now(clock_id);
+}
```
poettering: so this function sounds awfully specific to the sysusers usecase. I can't really see any other place we'd want to us it. Hence I think it should be added as static function to sysusers.c for now. We can still move it over if we find another use.

src/basic/time-util.c
```diff
@@ -55,6 +56,17 @@ usec_t now(clockid_t clock_id) {
...
 
+usec_t epoch_or_now(clockid_t clock_id) {
+        int r;
+        unsigned long long epoch;
+
+        r = getenv_llu_secure("SOURCE_DATE_EPOCH", 10, &epoch);
+        if (r == 0)
+                return (usec_t) epoch * USEC_PER_SEC;
+
+        return now(clock_id);
```
poettering: SOURCE_DATE_EPOCH is wallclock time. hence this function might either return time of the wallclock or time as specified in `clock_id`... That's broken. Just drop the parameter, and hard code CLOCK_REALTIME here. No other argument makes sense.

src/basic/time-util.c
```diff
@@ -55,6 +56,17 @@ usec_t now(clockid_t clock_id) {
         return timespec_load(&ts);
 }
 
+usec_t epoch_or_now(clockid_t clock_id) {
+        int r;
+        unsigned long long epoch;
+
+        r = getenv_llu_secure("SOURCE_DATE_EPOCH", 10, &epoch);
+        if (r == 0)
+                return (usec_t) epoch * USEC_PER_SEC;
```
poettering: please add an overflow check. i.e.:

```c
if (epoch > UINT64_MAX/USEC_PER_SEC) /* Overflow check */
        return USEC_INFINITY;
return (usec_t) epoch * USEC_PER_SEC;
```

src/basic/time-util.c
josch: Thanks, done.

docs/ENVIRONMENT.md
```diff
@@ -310,6 +310,15 @@ All tools:
...
   lines will always create directories.
 
+`systemd-sysusers`
+
+* `SOURCE_DATE_EPOCH` — if unset, the field of the date of last password change
+  in `/etc/shadow` will be the number of days from Jan 1, 1970 00:00 UTC until
+  today. If SOURCE_DATE_EPOCH is set to a valid UNIX epoch value, then the
+  field will be the number of days until that time instead. This is to support
+  creating bit-by-bit reproducible system images by choosing a reproducible
+  value for the field of the date of last password change in `/etc/shadow`.
```
poettering: should say unit is seconds.

And should carry link to → https://reproducible-builds.org/docs/source-date-epoch/

poettering: or maybe better → https://reproducible-builds.org/specs/source-date-epoch/

src/basic/env-util.c
josch: Thanks I changed the function name and the argument type to `uint64_t *`.

src/basic/time-util.c
josch: Done.

src/basic/time-util.c
josch: Good point, done.

src/basic/env-util.c
```diff
@@ -776,6 +776,18 @@ int getenv_bool_secure(const char *p) {
...
+int getenv_uint64_secure(const char *p, unsigned base, uint64_t *ret) {
+        const char *e;
+
+        assert(p);
+
+        e = secure_getenv(p);
+        if (!e)
+                return -ENXIO;
+
+        return safe_atollu_full(e, base, (unsigned long long *) ret);
```
poettering: actually, just use safe_atou64() here, and drop the `base` argument. It's highly unlikely we'll use anything but decimal here.

src/sysusers/sysusers.c
```diff
@@ -38,6 +38,7 @@
 #include "user-util.h"
 #include "utf8.h"
 #include "util.h"
+#include "env-util.h"
```
poettering: please keep include lines ordered alphabetically, as per coding style

poettering:
almost there...

josch:
@poettering thank you for your very valuable input -- I took care of your last few corrections. Thanks!

src/basic/env-util.c
```diff
@@ -776,6 +776,18 @@ int getenv_bool_secure(const char *p) {
         return parse_boolean(e);
 }
 
+int getenv_uint64_secure(const char *p, unsigned base, uint64_t *ret) {
```
bluca: `base` is unused now, you can drop it

bluca (approved):

PR #24536, opened 2022-09-02
dissect-image: introduce dissect_loop_device() which takes LoopDevice object

yuwata:
Currently, diskseq, uevent_seqnum_not_before, and timestamp_not_before
arguments are not used, hence, let's drop the explicit arguments for
them. But these may be used in future, and all of them are came from
LoopDevice object. So, let's make the function take LoopDevice object.

Replaces #24482.

### DIFF
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -2796,29 +2796,23 @@ finish:
         return r;
 }
 
-int dissect_image_and_warn(
-                int fd,
+int dissect_loop_device_and_warn(
                 const char *name,
+                const LoopDevice *loop,
                 const VeritySettings *verity,
                 const MountOptions *mount_options,
-                uint64_t diskseq,
-                uint64_t uevent_seqnum_not_before,
-                usec_t timestamp_not_before,
                 DissectImageFlags flags,
                 DissectedImage **ret) {
 
-        _cleanup_free_ char *buffer = NULL;
         int r;
 
-        if (!name) {
-                r = fd_get_path(fd, &buffer);
-                if (r < 0)
-                        return r;
+        assert(loop);
+        assert(loop->fd >= 0);
 
-                name = buffer;
-        }
+        if (!name)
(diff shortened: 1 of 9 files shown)


poettering:
hmm dissect_image() should really work on regular files too, no? I mean, we'll always set up a loopback device for it right now, but I think for testing and stuff it would be really good if we could use it on regular files too, and continue to do so.

LoopDevice always exposes a block device, hence always is privileged.

yuwata:
Yeah, so, it has dissect_image_fd().

yuwata:
OK. Now, `dissect_image()` is untouched, but this introduces a tiny wrapper of it which takes `LoopDevice` object. PTAL.

src/shared/dissect-image.c
```diff
@@ -2796,29 +2796,29 @@ int dissected_image_acquire_metadata(DissectedImage *m, DissectImageFlags extra_
...
 
         _cleanup_free_ char *buffer = NULL;
         int r;
 
+        assert(loop);
+        assert(loop->fd >= 0);
+
         if (!name) {
-                r = fd_get_path(fd, &buffer);
+                r = fd_get_path(loop->fd, &buffer);
```
poettering: hmm, if you have a LoopDevice object anyway, you could just use `loop->node` instead. Except thta there's one codepath right now where `loop->node` is not initialized, but we could fix that. i.e. in loop_device_make_internal() in the `if (S_ISBLK(st.st_mode))` branch at the top, in case we are not talking about a loopback block device, but another one. We could add a call to devpath_from_devnum() there, to get a pretty path from the stat data...

If you add that we would give the guarantee that `loop->node` is always valid, and thus you can drop the fd_get_path() stuff here

poettering:
looks good, but one suggestion to make it even better...

yuwata:
@poettering Thank you for the suggestion. PTAL.

poettering:
excellent! thanks!

Issue #24535, opened 2022-09-01
Can't spawn an nspawn container with `SYSTEMD_LOG_TARGET=journal`: Failed to duplicate stderr: Bad file descriptor


mrc0mmand:
### systemd version the issue has been seen with

latest main

### Used distribution

Arch Linux/Fedora 36

### Component

systemd-nspawn

### Unexpected behaviour you saw

I'm not sure if this is expected or not, but I came across this while debugging (and tweaking our tests) - I can't seem to spawn an nspawn container when `SYSTEMD_LOG_TARGET=journal`:

```
$ test/create-busybox-container test-cont

$ sudo -E systemd-nspawn --register=no -D test-cont /bin/busybox echo hello
Spawning container test-cont on /home/mrc0mmand/repos/systemd/test-cont.
Press ^] three times within 1s to kill container.
hello

$ SYSTEMD_LOG_TARGET=journal sudo -E systemd-nspawn --register=no -D test-cont /bin/busybox echo hello
$ echo $?
1
$ journalctl -e -o short-monotonic --no-hostname -t systemd-nspawn
...
[3349465.791055] systemd-nspawn[330761]: Spawning container test-cont on /home/mrc0mmand/repos/systemd/test-cont.
[3349465.791449] systemd-nspawn[330761]: Press ^] three times within 1s to kill container.
[3349465.808706] systemd-nspawn[330763]: Failed to duplicate stderr: Bad file descriptor
[3349465.809393] systemd-nspawn[330761]: Child died too early.
```

With debug:

```
$ SYSTEMD_LOG_LEVEL=debug SYSTEMD_LOG_TARGET=journal sudo -E systemd-nspawn --register=no -D test-cont /bin/busybox echo hello
[3349546.280273] systemd-nspawn[330908]: Setting RLIMIT_CPU to infinity.
[3349546.280567] systemd-nspawn[330908]: Setting RLIMIT_FSIZE to infinity.
[3349546.280627] systemd-nspawn[330908]: Setting RLIMIT_DATA to infinity.
[3349546.280677] systemd-nspawn[330908]: Setting RLIMIT_STACK to 8388608:infinity.
[3349546.280727] systemd-nspawn[330908]: Setting RLIMIT_CORE to 0:infinity.
[3349546.280770] systemd-nspawn[330908]: Setting RLIMIT_RSS to infinity.
[3349546.280816] systemd-nspawn[330908]: Setting RLIMIT_NPROC to 63535.
[...]

PR #24537, opened 2022-09-02
log: make log_dup_console() open console if it is not opened yet

yuwata:
Fixes #24535.

PR #24544, opened 2022-09-02
log: don't attempt to duplicate closed fd

poettering:
if the console fd is not open we shouldn#t try to move it out of the 0…2
range.

Fixes: #24535
Alternative-for: #24537


poettering:
@mrc0mmand can you check if this one works for you too?

yuwata (approved):

mrc0mmand (approved):
This seems to work as well, thanks!

```
$ SYSTEMD_LOG_LEVEL=debug SYSTEMD_LOG_TARGET=journal sudo -E build/systemd-nspawn --register=no -D test-cont /bin/busybox echo hello
hello

```

PR #24547, opened 2022-09-02
test: extend ConditionFirmware tests

d4nuu8:


src/test/test-condition.c
```diff
@@ -306,11 +306,82 @@ TEST(condition_test_architecture) {
...
+                condition_free(condition);
+        } else {
+                _cleanup_free_ char *dtcompat = NULL;
+                _cleanup_strv_free_ char **dtcompatlist = NULL;
+                size_t dtcompat_size;
+                const char *expression;
+
+                r = read_full_virtual_file("/proc/device-tree/compatible", &dtcompat, &dtcompat_size);
+                if (r < 0) {
+                        if ( r!= -ENOENT) {
```
yuwata: Unnecessary space after `(`.

src/test/test-condition.c
```diff
@@ -306,11 +306,82 @@ TEST(condition_test_architecture) {
...
+                                condition = condition_new(CONDITION_FIRMWARE, "device-tree-compatible()", false, false);
+                                assert_se(condition);
+                                assert_se(condition_test(condition, environ) == r);
```
yuwata: I think it is better here `< 0` instead of `== r`, even if it currently returns `r`, as the error code may be changed in the future.

src/test/test-condition.c
```diff
@@ -306,11 +306,82 @@ TEST(condition_test_architecture) {
...
+                                condition_free(condition);
+                                return
+                        }
+
+                        condition = condition_new(CONDITION_FIRMWARE, "device-tree-compatible()", false, false);
+                        assert_se(condition);
+                        assert_se(condition_test(condition, environ) == 0);
+                        condition_free(condition);
+                        return;
```
yuwata, line 358: You can merge this block like the following:
```c
                r = read_full_virtual_file("/proc/device-tree/compatible", &dtcompat, &dtcompat_size);
                if (r < 0) {
                        condition = condition_new(CONDITION_FIRMWARE, "device-tree-compatible()", false, false);
                        assert_se(condition);
                        if (r == -ENOENT)
                                assert_se(condition_test(condition, environ) == 0);
                        else
                                assert_se(condition_test(condition, environ) < 0);
                        condition_free(condition);
                        return;
                }
```

src/test/test-condition.c
```diff
@@ -306,11 +306,82 @@ TEST(condition_test_architecture) {
...
+                }
+
+                dtcompatlist = strv_parse_nulstr(dtcompat, dtcompat_size);
+
+                STRV_FOREACH(c, dtcompatlist) {
+                        expression = strjoina("device-tree-compatible(", *c, ")");
```
yuwata: Please do not use `strjoina()` in loop. Please use `strjoin()` instead.

src/test/test-condition.c
```diff
@@ -306,11 +306,82 @@ TEST(condition_test_architecture) {
...
+                        condition = condition_new(CONDITION_FIRMWARE, expression,  false, false);
```
yuwata: doubled space after `expression,`

yuwata (changes requested):
Looks nice. Several minor comments.

src/test/test-condition.c
d4nuu8: Removed

src/test/test-condition.c
d4nuu8: Done

src/test/test-condition.c
d4nuu8: Refactored

src/test/test-condition.c
d4nuu8: Done

src/test/test-condition.c
d4nuu8: Removed

d4nuu8:
Requested changes applied

src/test/test-condition.c
```diff
@@ -306,11 +306,77 @@ TEST(condition_test_architecture) {
...
+                        else
+                                assert_se(condition_test(condition, environ) < 0);
+                        condition_free(condition);
+                        return;
+                }
+
+                dtcompatlist = strv_parse_nulstr(dtcompat, dtcompat_size);
+
+                STRV_FOREACH(c, dtcompatlist) {
+                        expression = strjoin("device-tree-compatible(", *c, ")");
```
yuwata: Then, you need to free it. Also, missing assertion for it. Please declare `expression` in the loop and use `_cleanup_free_` attribute:
```c
STRV_FOREACH(...) {
        _cleanup_free_ char *expression = NULL;

        assert_se(expression = strjoin(...));
        ...
}
```

yuwata (changes requested):
Mostly good. But one more round please.

src/test/test-condition.c
d4nuu8: Done

d4nuu8:
Requested changes applied

yuwata (approved):
LGTM. Thank you!

d4nuu8:
Thanks for the review!

bluca (approved):

PR #24554, opened 2022-09-03
man: fix static bridge example

Fumon:
A NetDev is needed to create the bridge in order to match the example's description "This creates a bridge..."


yuwata (approved):

yuwata:
Thank you!

PR #24555, opened 2022-09-03
bootctl: Small improvements

medhefgo:


src/shared/bootspec.c
```diff
@@ -1288,7 +1288,11 @@ BootEntry* boot_config_find_entry(BootConfig *config, const char *id) {
 static void boot_entry_file_list(const char *field, const char *root, const char *p, int *ret_status) {
         int status = boot_entry_file_check(root, p);
 
-        printf("%13s%s ", strempty(field), field ? ":" : " ");
+        if (field)
+                printf("%13s: %s\n", field, root);
+
+        printf("%13s  %s", "", special_glyph(SPECIAL_GLYPH_TREE_RIGHT));
```
poettering: In the light of 84be9b63ebac1d320b1e39ccf6e0895e621af887 we shouldn't just use `└─` here, but use `├─` for all but the last item we display...

The code for this should really be simple to make that work, because in the STRV_FOREACH() loops that call this function it's super easy to figure out if we are at the last entry or not...

(might be worse adding a new helper function that displays the whole strv btw, instead of duplicating the STRV_FOREACH loops)

src/shared/bootspec.c
```diff
@@ -993,6 +993,9 @@ static int boot_entries_uniquify(BootEntry *entries, size_t n_entries) {
 static int boot_config_find(const BootConfig *config, const char *id) {
         assert(config);
 
+        if (strcaseeq_ptr(id, "@saved"))
+                id = config->entry_selected;
```
poettering: we probably should also add:

```
if (id[0] == '@') 
        return -1
```

here, to ensure `@` prefixed ids remain systematically special

keszybz:
```
       source: /boot/loader/entries/arch.conf
        linux: /boot
               └─/vmlinuz-linux
       initrd: /boot
               ├─/intel-ucode.img
               └─/initramfs-linux.img
```
I think `source` needs to be handles the same way too. After all, it's a a file on that partition, the same as the other files. BLS requires all files that comprise a boot entry to be on the same partition, so it's important that `/loader/entries/arch.conf`, `/vmlinuz-linux`, `/intel-ucode.img`, `/initramfs-linux.img` are all under `/boot`, and it'd be confusing to display one in a different way.

poettering:
i'd even be fine displaying it like this:

```
[/boot]/intel-ucode.img
```

And then additionally use grey ANI color for the `[/boot]` part.

medhefgo:
I've changed it to grey out the prefix or use `[]` with disabled colors

I don't think `source` should be treated this way, though. Mainly because the other fields *have* to be interpreted relative to the root, while `source` doesn't. The fact is that it's already in the correct place because we picked it up and (successfully) parsed it. Also, the path may point to stuff outside of a boot partition (the `/sys` efivar path in particular).

poettering:
> I've changed it to grey out the prefix or use `[]` with disabled colors

Hmm, so far we had the rule that turning off colors won't change anything but colors. i.e. that what you copy/paste from your terminal app is identical to what you#d get if you piped the output without colors somewhere...

I think that's a good rule, since poeple assume that piping output somewhere doesn't change things (and piping does turn off color). Hence I think we should either do the [] always, or never, but not conditionalize it on color output. (And I'd probably put it there)

keszybz:
/me likes.

> Hmm, so far we had the rule that turning off colors won't change anything but colors.

Yeah, that's a good point. By I think I'd prefer to remove `[]` altogether. Maybe we can add a note somewhere that the paths are prefix in this way? In particular in the docs, and maybe in the first line of output: `Boot Loader Entries (paths relative to file system root)` ?

keszybz:
The basic question is: what is the *purpose* of those paths? My main use is to select the path and use it in some command, e.g. to call `objdump` or `ls -l` or whatever. What else do you use the paths for?

medhefgo:
> Yeah, that's a good point. By I think I'd prefer to remove `[]` altogether. Maybe we can add a note somewhere that the paths are prefix in this way? In particular in the docs, and maybe in the first line of output: `Boot Loader Entries (paths relative to file system root)` ?

Nobody's gonna look at that. Just like they still wipe their ESP and put it somewhere else because they went out of space on it instead of making a XBOOTLDR partition.

> The basic question is: what is the _purpose_ of those paths? My main use is to select the path and use it in some command, e.g. to call `objdump` or `ls -l` or whatever. What else do you use the paths for?

To make sure that people can't say "look that file is right here in `/boot/somefile`, I have it open right now. Why is systemd-boot so dumb it cannot even find things!" when the file is actually looked up under `/efi`. And since they must be relative to their root when put in a `conf` snippet I would consider the interruption that the `[]` would cause when copy-pasting a benefit.

cmurf:
Another idea `//` which won't adversely affect copy/paste (in most shells anyway), e.g. `/boot//loader/entries/arch.conf`

poettering:
I'd be fine with the // idea, too

poettering:
@keszybz what's your take on this?

1. Use […] as separators
2. Use // as separators
3. Do not use a separator, just color

medhefgo:
Can we decide on the color of the bike shed and get this into v252?

keszybz:
2 + 3 (i.e. `<grey>/prefix/</grey>/path`) sounds like the best option.

keszybz:
Nice! Sorry for the delay.

PR #24557, opened 2022-09-04
repart: fix partition size calculation

### LINKED ISSUE #24553: systemd-repart dies with assertion
Assertion 'amount <= total' failed at src/partition/repart.c.533, function charge_size(). Aborting.

Config is:
```
[Partition]
Type=root
SizeMinBytes=10G
SizeMaxBytes=120G
GrowFileSystem=yes
```

yuwata:
Fixes #24553.

### DIFF
--- a/test/units/testsuite-58.sh
+++ b/test/units/testsuite-58.sh
@@ -570,6 +570,106 @@ EOF
     assert_in "$imgs/21817.img2 : start=      104448, size=      (100319| 98304)," "$output"
 }
 
+test_issue_24553() {
+    local defs imgs output
+
+    # testcase for #24553
+
+    defs="$(mktemp --directory "/tmp/test-repart.XXXXXXXXXX")"
+    imgs="$(mktemp --directory "/var/tmp/test-repart.XXXXXXXXXX")"
+    # shellcheck disable=SC2064
+    trap "rm -rf '$defs' '$imgs'" RETURN
+
+    cat >"$defs/root.conf" <<EOF
+[Partition]
+Type=root
+SizeMinBytes=10G
+SizeMaxBytes=120G
+EOF
+
+    cat >"$imgs/partscript" <<EOF
+label: gpt
+label-id: C9FFE979-A415-C449-B729-78C7AA664B10
+unit: sectors
+first-lba: 40
+
+start=40, size=524288, type=C12A7328-F81F-11D2-BA4B-00A0C93EC93B, uuid=F2E89C8A-DC5D-4C4C-A29C-6CFB643B74FD, name="ESP System Partition"
+start=524328, size=14848000, type=${root_guid}, uuid=${root_uuid}, name="root-${architecture}"
+EOF
+
+    # 1. Operate on a small image compared with SizeMinBytes=.
+    truncate -s 8g "$imgs/zzz"
+    sfdisk "$imgs/zzz" <"$imgs/partscript"
+
+    # This should fail, but not trigger assertions.
+    assert_rc 1 systemd-repart --definitions="$defs" \
+                               --seed="$seed" \
+                               --dry-run=no \
+                               "$imgs/zzz"
+
+    output=$(sfdisk --dump "$imgs/zzz")
+    assert_in "$imgs/zzz2 : start=      524328, size=    14848000, type=${root_guid}, uuid=${root_uuid}, name=\"root-${architecture}\"" "$output"
+
+    # 2. Operate on an larger image compared with SizeMinBytes=.
+    rm -f "$imgs/zzz"
+    truncate -s 12g "$imgs/zzz"
+    sfdisk "$imgs/zzz" <"$imgs/partscript"
+
+    # This should succeed.
+    systemd-repart --definitions="$defs" \
+                   --seed="$seed" \
+                   --dry-run=no \
+                   "$imgs/zzz"
+
+    output=$(sfdisk --dump "$imgs/zzz")
+    assert_in "$imgs/zzz2 : start=      524328, size=    24641456, type=${root_guid}, uuid=${root_uuid}, name=\"root-${architecture}\"" "$output"
+
+    # 3. Multiple partitions with Priority= (small disk)
+    cat >"$defs/root.conf" <<EOF
+[Partition]
+Type=root
+SizeMinBytes=10G
+SizeMaxBytes=120G
+Priority=100
+EOF
+
+    cat >"$defs/usr.conf" <<EOF
+[Partition]
+Type=usr
+SizeMinBytes=10M
+Priority=10
+EOF
+
+    rm -f "$imgs/zzz"
+    truncate -s 8g "$imgs/zzz"
+    sfdisk "$imgs/zzz" <"$imgs/partscript"
+
+    # This should also succeed, but root is not extended.
+    systemd-repart --definitions="$defs" \
+                   --seed="$seed" \
+                   --dry-run=no \
+                   "$imgs/zzz"
+
+    output=$(sfdisk --dump "$imgs/zzz")
+    assert_in "$imgs/zzz2 : start=      524328, size=    14848000, type=${root_guid}, uuid=${root_uuid}, name=\"root-${architecture}\"" "$output"
+    assert_in "$imgs/zzz3 : start=    15372328, size=     1404848, type=${usr_guid}, uuid=${usr_uuid}, name=\"usr-${architecture}\", attrs=\"GUID:59\"" "$output"
+
+    # 4. Multiple partitions with Priority= (large disk)
+    rm -f "$imgs/zzz"
+    truncate -s 12g "$imgs/zzz"
+    sfdisk "$imgs/zzz" <"$imgs/partscript"
+
+    # This should also succeed, and root is extended.
+    systemd-repart --definitions="$defs" \
+                   --seed="$seed" \
+                   --dry-run=no \
+                   "$imgs/zzz"
+
+    output=$(sfdisk --dump "$imgs/zzz")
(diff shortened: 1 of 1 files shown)


src/partition/repart.c
```diff
@@ -470,6 +470,9 @@ static uint64_t partition_max_size(const Context *context, const Partition *p) {
                 return p->current_size;
         }
 
+        if (p->size_max == UINT64_MAX)
+                return UINT64_MAX;
```
keszybz, line 551: Does this have any effect on the logic? After looking at the commit, I don't see anything…

src/partition/repart.c
```diff
@@ -713,7 +716,7 @@ static int context_grow_partitions_phase(
                                 p->new_size = rsz;
                                 charge = try_again = true;
 
-                        } else if (phase == PHASE_UNDERCHARGE && xsz != UINT64_MAX && xsz < share) {
+                        } else if (phase == PHASE_UNDERCHARGE && xsz < share) {
```
keszybz, line 833: This is a noop, because `xsz < share => xsz < UINT64_MAX` (`=>` in the sense of implication) …

src/partition/repart.c
```diff
@@ -821,7 +824,7 @@ static int context_grow_partitions_on_free_area(Context *context, FreeArea *a) {
                 m = MAX(a->after->new_size, round_down_size(a->after->new_size + span, context->grain_size));
 
                 xsz = partition_max_size(context, a->after);
-                if (xsz != UINT64_MAX && m > xsz)
+                if (m > xsz)
```
keszybz: Here too, this is a noop, because `m > xsz => xsz < UINT64_MAX`…

src/partition/repart.c
```diff
@@ -844,7 +847,7 @@ static int context_grow_partitions_on_free_area(Context *context, FreeArea *a) {
                         m = MAX(p->new_size, round_down_size(p->new_size + span, context->grain_size));
 
                         xsz = partition_max_size(context, p);
-                        if (xsz != UINT64_MAX && m > xsz)
+                        if (m > xsz)
```
keszybz: Here too, this is a noop, because `m > xsz => xsz < UINT64_MAX`.
So is the whole commit supposed to change anything, or is just cleanup? But then, what is the purpose of the first chunk?

src/partition/repart.c
```diff
@@ -750,24 +760,23 @@ static int context_grow_partitions_phase(
...
+                        uint64_t share, rsz, xsz;
 
                         r = scale_by_weight(*span, p->padding_weight, *weight_sum, &share);
                         if (r < 0)
                                 return r;
 
-                        if (phase == PHASE_OVERCHARGE && p->padding_min != UINT64_MAX && p->padding_min > share) {
-                                p->new_padding = p->padding_min;
+                        rsz = partition_min_padding(p);
+                        xsz = partition_max_padding(p);
```
keszybz, line 865: Hmm, why "x" and "r"? Maybe call this something more descriptive.

src/partition/repart.c
```diff
@@ -784,13 +783,10 @@ static bool context_grow_partitions_phase(
...
                                 *weight_sum = charge_weight(*weight_sum, p->padding_weight);
                         }
-
-                        if (try_again)
-                                return false; /* try again */
                 }
         }
 
-        return true; /* done */
+        return !try_again;
```
keszybz, line 886: Hmm, is the whole commit supposed to change behaviour?

src/partition/repart.c
```diff
@@ -753,7 +753,8 @@ static bool context_grow_partitions_phase(
                                  * assigning this shouldn't impact the shares of the other
                                  * partitions. */
 
-                                p->new_size = MAX(round_down_size(share, context->grain_size), rsz);
+                                assert(share >= rsz);
+                                p->new_size = MIN(MAX(round_down_size(share, context->grain_size), rsz), xsz);
```
keszybz: `CLAMP` ?

src/partition/repart.c
```diff
@@ -779,7 +780,8 @@ static bool context_grow_partitions_phase(
                                 p->new_padding = xsz;
                                 charge = try_again = true;
                         } else if (phase == PHASE_DISTRIBUTE) {
-                                p->new_padding = MAX(round_down_size(share, context->grain_size), rsz);
+                                assert(share >= rsz);
+                                p->new_padding = MIN(MAX(round_down_size(share, context->grain_size), rsz), xsz);
```
keszybz: Also here.

keszybz (commented):
Looks reasonable, but for a few commits I wasn't quite sure what the intent is.

One CI failed with `TEST-13-NSPAWN-SMOKE`, unrelated.

src/partition/repart.c
yuwata: Previously, this did not return UINT64_MAX, but returned a huge value. This fixes a theoretical issue, `xsz < share` may be true for sooooo huge disk. In realistic situations, this commit does not change anything.

src/partition/repart.c
yuwata: Yeah, so I dropped `xsz != UINT64_MAX`, as it is redundant.

src/partition/repart.c
yuwata: I do not know, I just followed the names used in the above block.

src/partition/repart.c
yuwata: This should not change anything.

yuwata:
@keszybz Thank you for the comments. I added several more commits to make actually fix your issue. Also, extends commit messages to explain the changes. PTAL.

poettering:
maybe include in all relevant commits "no actual code changes, just refactoring" or so, to emphasize the fixes and isolate it from the rest

src/partition/repart.c
```diff
@@ -523,35 +523,40 @@ static uint64_t free_area_available(const FreeArea *a) {
...
 
-        /* Similar to free_area_available(), but takes into account that the required size and padding of the
-         * preceding partition is honoured. */
-
-        avail = free_area_available(a);
-        if (a->after) {
-                uint64_t need, space_end, new_end;
+        /* Calculate where the free area ends, based on the offset of the partition preceding it. */
+        return round_up_size(a->after->offset + a->after->current_size, context->grain_size) + free_area_available(a);
+}
```
poettering, line 614: if this is now a helper of its own, maybe also add a code path dealing with the first free area, i.e. where `a->after` is NULL. In that case you can just return `free_area_available()` after all the offset is zero...

src/partition/repart.c
```diff
@@ -523,35 +523,40 @@ static uint64_t free_area_available(const FreeArea *a) {
...
+        assert(context);
+        assert(a);
+        assert(a->after);
+        assert(a->after->offset != UINT64_MAX);
+        assert(a->after->current_size != UINT64_MAX);
 
-                assert(a->after->offset != UINT64_MAX);
-                assert(a->after->current_size != UINT64_MAX);
+        /* Calculate where the partition would end when we give it as much as it needs. */
+        return round_up_size(a->after->offset + partition_min_size_with_padding(context, a->after), context->grain_size);
```
poettering, line 627: similar, return 0 here if `a->after` is NULL?

src/partition/repart.c
```diff
@@ -523,35 +523,40 @@ static uint64_t free_area_available(const FreeArea *a) {
...
 
-                /* Calculate saturated difference of the two: that's how much we have free for other partitions */
-                return LESS_BY(space_end, new_end);
-        }
+        /* Calculate saturated difference of the two: that's how much we have free for other partitions */
+        if (a->after)
+                return LESS_BY(free_area_current_end(context, a), free_area_min_end(context, a));
 
-        return avail;
+        return free_area_available(a);
```
poettering: and if you do the above, then this function becomes a lot simpler, it's just the LESS_BY() expression, nothing else.

src/partition/repart.c
```diff
@@ -578,6 +578,23 @@ static uint64_t charge_weight(uint64_t total, uint64_t amount) {
...
 
+static bool context_verify_free_areas(Context *context) {
+        assert(context);
+
+        /* Check that each existing partition can fit its area. */
+        for (size_t i = 0; i < context->n_free_areas; i++) {
+                FreeArea *a = context->free_areas[i];
+
+                if (!a->after)
+                        continue;
```
poettering: you can drop this too, if you make the suggested changes on th previous commit

src/partition/repart.c
```diff
@@ -5005,6 +5022,10 @@ static int run(int argc, char *argv[]) {
                         return r;
         }
 
+        if (!context_verify_free_areas(context))
+                return log_error_errno(SYNTHETIC_ERRNO(ENOSPC),
+                                       "Can't grow existing partitions into available free space, refusing.");
```
poettering: hmm, should this really be fatal?

I mean, this check will fail if PaddingMin= is larger than what is currently in effect and what space is available after the existing partition, right? I don't think we should fail in cases like this. i.e. PaddingMin= should be honoured by us for stuff we do, but if there's a violation of this constraint *already* in place before we even considered anything, then it should not matter to us.

Or am I misunderstanding what this commit does?

yuwata: PaddingMin= can also trigger the failure, but more important case is SizeMinBytes=, as reported #24553.

yuwata: BTW, we may simply set 'Partition.drop' instead of refuse the entire process.

yuwata:
@poettering Thank you for the suggestions and comments. All suggestions are applied. Also, now, the existing partition can be also dropped (though, it is handled as foreign, more precisely). PTAL.

poettering:
Hmm, needs rebase

poettering:
lgtm, but needs a rebase

PR #24560, opened 2022-09-04
repart: Add support for setting a partition's UUID to zero

daandemeyer:
This is useful when we need to fill in the UUID later, such as when
using verity partitions.

### DIFF
--- a/src/partition/repart.c
+++ b/src/partition/repart.c
@@ -138,6 +138,7 @@ struct Partition {
 
         sd_id128_t type_uuid;
         sd_id128_t current_uuid, new_uuid;
+        bool new_uuid_is_set;
         char *current_label, *new_label;
 
         bool dropped;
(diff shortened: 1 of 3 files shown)


src/partition/repart.c
```diff
@@ -2066,7 +2099,7 @@ static int context_dump_partitions(Context *context, const char *node) {
                                 t,
                                 TABLE_STRING, gpt_partition_type_uuid_to_string_harder(p->type_uuid, uuid_buffer),
                                 TABLE_STRING, empty_to_null(label) ?: "-", TABLE_SET_COLOR, empty_to_null(label) ? NULL : ansi_grey(),
-                                TABLE_UUID, sd_id128_is_null(p->new_uuid) ? p->current_uuid : p->new_uuid,
+                                TABLE_UUID, !sd_id128_is_null(p->new_uuid) || p->new_uuid_is_set ? p->new_uuid : p->current_uuid,
```
yuwata: Why not simply use `p->new_uuid_is_set`? I think `!sd_id128_is_null()` is completely redundant.

src/partition/repart.c
```diff
@@ -2184,7 +2217,7 @@ static int partition_hint(const Partition *p, const char *node, char **ret) {
                 goto done;
         }
 
-        if (!sd_id128_is_null(p->new_uuid))
+        if (!sd_id128_is_null(p->new_uuid) || p->new_uuid_is_set)
```
yuwata: Also here.

src/partition/repart.c
```diff
@@ -3249,7 +3282,7 @@ static int context_acquire_partition_uuids_and_labels(Context *context) {
 
                 if (!sd_id128_is_null(p->current_uuid))
                         p->new_uuid = p->current_uuid; /* Never change initialized UUIDs */
-                else if (sd_id128_is_null(p->new_uuid)) {
+                else if (sd_id128_is_null(p->new_uuid) && !p->new_uuid_is_set) {
```
yuwata: Also here.

src/partition/repart.c
```diff
@@ -4042,7 +4072,7 @@ static int context_open_copy_block_paths(
                 free_and_replace(p->copy_blocks_path, opened);
 
                 /* When copying from an existing partition copy that partitions UUID if none is configured explicitly */
-                if (sd_id128_is_null(p->new_uuid) && !sd_id128_is_null(uuid))
+                if (sd_id128_is_null(p->new_uuid) && !p->new_uuid_is_set && !sd_id128_is_null(uuid))
```
yuwata: And here.

src/partition/repart.c
```diff
@@ -1313,12 +1314,44 @@ static int config_parse_gpt_flags(
...
+        Partition *partition = ASSERT_PTR(data);
+        int r;
+
+        if (isempty(rvalue)) {
+                partition->new_uuid = SD_ID128_NULL;
+                partition->new_uuid_is_set = false;
+                return 0;
+        }
+
+        r = sd_id128_from_string(rvalue, &partition->new_uuid);
```
yuwata, line 1344: How about to accept a special value `"empty"`?

yuwata (changes requested):

src/partition/repart.c
```diff
@@ -1313,12 +1314,50 @@ static int config_parse_gpt_flags(
...
+        Partition *partition = ASSERT_PTR(data);
+        int r;
+
+        if (isempty(rvalue)) {
+                partition->new_uuid = SD_ID128_NULL;
+                partition->new_uuid_is_set = false;
+                return 0;
+        }
+
+        if (streq(rvalue, "empty")) {
```
poettering: Is "empty" the right word? we can't really set it to "empty". only to `00000000-0000-0000-0000-000000000000`, which maybe we should call "null" (after all, the C macro for it is called `SD_ID128_NULL`?)

Also, please mention this shortcut in the man page

poettering:
lgtm

src/partition/repart.c
```diff
@@ -4042,7 +4080,7 @@ static int context_open_copy_block_paths(
                 free_and_replace(p->copy_blocks_path, opened);
 
                 /* When copying from an existing partition copy that partitions UUID if none is configured explicitly */
-                if (sd_id128_is_null(p->new_uuid) && !sd_id128_is_null(uuid))
+                if (!p->new_uuid_is_set && !sd_id128_is_null(uuid))
                         p->new_uuid = uuid;
```
keszybz: I think ` p->new_uuid_is_set` should be set here, even if it does not matter for current logic, just so that we have a self-consistent state.

keszybz (changes requested):

yuwata (approved):
LGTM.

PR #24561, opened 2022-09-04
loop-util: several follow ups for recent changes

yuwata:
Several follow-ups for #24530.

~~Hopefully fixes the portablectl hang issue #24147.~~

### DIFF
--- a/src/shared/loop-util.c
+++ b/src/shared/loop-util.c
@@ -357,7 +357,7 @@ static int loop_device_make_internal(
                 int lock_op,
                 LoopDevice **ret) {
 
-        _cleanup_close_ int direct_io_fd = -1, lock_fd = -1;
+        _cleanup_close_ int direct_io_fd = -1;
         _cleanup_free_ char *node = NULL;
         bool try_loop_configure = true;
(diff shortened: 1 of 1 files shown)


yuwata:
@mrc0mmand Could you test the PR with https://github.com/systemd/systemd-centos-ci/pull/540?

yuwata:
Hm, it seems the issue is not fixed yet...

mrc0mmand:
> Hm, it seems the issue is not fixed yet...

Yeah :( I'm trying to fix the interactive debug stuff, since currently it's pretty much unusable with sanitizers (at least in TEST-29), so I can debug it a bit further.

src/shared/loop-util.c
```diff
@@ -851,7 +851,7 @@ int loop_device_open(
                 .nr = nr,
                 .node = TAKE_PTR(p),
                 .relinquished = true, /* It's not ours, don't try to destroy it when this object is freed */
-                .devno = st.st_dev,
+                .devno = st.st_rdev,
```
poettering, line 826: ouch

PR #24569, opened 2022-09-05
test: kill plymouthd after initrd transition if it's still running

mrc0mmand:
Until now using the INTERACTIVE_DEBUG=yes stuff together with sanitizers
was almost impossible, since the console kept eating up our inputs or
not responding at all. After a painful day of debugging I noticed that
if we use a shell script in the initrd -> root transition, we might end up
with a plymouthd still running, which kept screwing with the tty.

E.g. with initrd -> wrapper -> systemd transition, where the `wrapper`
is a simple script:

```
#!/bin/bash
exec -- /usr/lib/systemd/systemd "$@"
```

we'd end up with a stray plymouthd process after the bootup:

```
 1     0     440       2  20   0      0     0 worker I    ?          0:00 [kworker/5:2-ata_sff]
 1     0     453       2  20   0      0     0 worker I    ?          0:00 [kworker/9:2-rcu_gp]
 5     0     456       1  20   0   7252  1960 do_epo S    ?          0:00 @usr/sbin/plymouthd --mode=boot --pid-file=/run/plymouth/pid --attach-to-session
```

After killing it, the tty works finally as expected.

---

This took an embarrassing amount of time, but it seems to finally work. I suspect we end up with the stray plymouthd process because during the initrd -> asan_wrapper -> systemd transition we'll lose the internal systemd state including some plymouth services (like `plymouth-quit.service`), but that's just a wild guess.

/cc @evverx, @bluca

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -179,6 +179,7 @@ BASICTOOLS=(
     head
     ionice
     ip
+    killall
     ldd
     ln
     loadkeys
@@ -1794,6 +1795,26 @@ install_debug_tools() {
             image_install resize
             echo "resize" >>"$initdir/etc/profile"
         fi
+
+        # Sometimes we might end up with plymouthd still running (especially
+        # with the initrd -> asan_wrapper -> systemd transition), which will eat
+        # our inputs and make debugging via tty impossible. Let's fix this by
+        # killing plymouthd explicitly for the interactive sessions.
+        # Note: we can't use pkill/pidof/etc. here due to a bug in libasan, see:
+        #   - https://github.com/llvm/llvm-project/issues/49223
+        #   - https://bugzilla.redhat.com/show_bug.cgi?id=2098125
+        local plymouth_unit="${initdir:?}/etc/systemd/system/kill-plymouth.service"
+        cat >"$plymouth_unit" <<EOF
+[Unit]
+After=multi-user.target
+
+[Service]
+ExecStart=sh -c 'killall --verbose plymouthd || :'
+
+[Install]
+WantedBy=multi-user.target
+EOF
+        "${SYSTEMCTL:?}" enable --root "${initdir:?}" kill-plymouth.service
     fi
 }
 


test/test-functions
```diff
@@ -894,6 +895,15 @@ printf "[Unit]\nConditionVirtualization=container\n\n[Service]\nTimeoutSec=240s\
...
 printf "[Service]\nTimeoutSec=180s\n" >/etc/systemd/system/systemd-journal-flush.service.d/timeout.conf
 
+# With the initrd -> asan_wrapper -> systemd transition we might end up with
+# plymouthd still running, which will eat our inputs and make debugging via tty
+# impossible, so let's kill it before exec()'ing systemd.
+# Note: we can't use pkill/pidof/etc. here due to a bug in libasan, see:
+#   - https://github.com/llvm/llvm-project/issues/49223
+#   - https://bugzilla.redhat.com/show_bug.cgi?id=2098125
+# Thankfully, the standalone kill binary (not the bash's built-in) works too.
+/bin/kill --all --verbose plymouthd || :
```
bluca: on debian:

```
-bash-5.1# /bin/kill --all --verbose plymouthd || :
kill: invalid argument 

Usage:
 kill [options] <pid> [...]

Options:
 <pid> [...]            send signal to every <pid> listed
 -<signal>, -s, --signal <signal>
                        specify the <signal> to be sent
 -q, --queue <value>    integer value to be sent with the signal
 -l, --list=[<signal>]  list all signal names, or convert one to a name
 -L, --table            list all signal names in a nice table

 -h, --help     display this help and exit
 -V, --version  output version information and exit

For more details see kill(1).
```

bluca: good news is, killing plymouthd indeed fixes the problem I have, even without sanitizers! halleluja!

mrc0mmand: > on debian:
> 
> ```
> -bash-5.1# /bin/kill --all --verbose plymouthd || :
> kill: invalid argument 
> 
> Usage:
>  kill [options] <pid> [...]
> 
> Options:
>  <pid> [...]            send signal to every <pid> listed
>  -<signal>, -s, --signal <signal>
>                         specify the <signal> to be sent
>  -q, --queue <value>    integer value to be sent with the signal
>  -l, --list=[<signal>]  list all signal names, or convert one to a name
>  -L, --table            list all signal names in a nice table
> 
>  -h, --help     display this help and exit
>  -V, --version  output version information and exit
> 
> For more details see kill(1).
> ```

*[screams internally]*

Looks like `killall` works too, at least on Fedora.

bluca (approved):
```
-bash-5.1# killall --verbose plymouthd
Killed plymouthd(206) with signal 15
```

ship it!

bluca:
> ```
> -bash-5.1# killall --verbose plymouthd
> Killed plymouthd(206) with signal 15
> ```
> 
> ship it!

in fact, can we do this always, regardless of asan or not?

mrc0mmand:
> > ```
> > -bash-5.1# killall --verbose plymouthd
> > Killed plymouthd(206) with signal 15
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ship it!
> 
> in fact, can we do this always, regardless of asan or not?

I guess that would make sense, updated.

bluca (approved):

PR #24571, opened 2022-09-05
dissect: save image path to LoopDevice and use it when dissect loop device

yuwata:
Follow-up for #24322.

### DIFF
--- a/src/shared/dissect-image.c
+++ b/src/shared/dissect-image.c
@@ -723,7 +723,7 @@ int dissect_image(
                         }
 
                         if (designator != _PARTITION_DESIGNATOR_INVALID) {
-                                _cleanup_free_ char *t = NULL, *n = NULL, *o = NULL, *l = NULL;
+                                _cleanup_free_ char *t = NULL, *o = NULL, *l = NULL;
                                 const char *options = NULL;
 
                                 if (m->partitions[designator].found) {
@@ -745,10 +745,6 @@ int dissect_image(
                                                 return -ENOMEM;
                                 }
 
-                                n = strdup(node);
-                                if (!n)
-                                        return -ENOMEM;
-
                                 if (label) {
                                         l = strdup(label);
                                         if (!l)
@@ -768,7 +764,7 @@ int dissect_image(
                                         .rw = rw,
                                         .growfs = growfs,
                                         .architecture = architecture,
-                                        .node = TAKE_PTR(n),
+                                        .node = TAKE_PTR(node),
                                         .fstype = TAKE_PTR(t),
                                         .label = TAKE_PTR(l),
                                         .uuid = id,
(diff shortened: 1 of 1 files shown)


src/shared/blockdev-util.c
```diff
@@ -509,19 +509,26 @@ int block_device_resize_partition(
...
-                return r;
+        if (!dev) {
+                struct stat st;
+
+                if (fstat(fd, &st) < 0)
+                        return -errno;
+
+                r = sd_device_new_from_stat_rdev(&dev_unref, &st);
+                if (r < 0)
+                        return r;
```
poettering, line 575: maybe time for a sd_device_new_from_fd()?

src/shared/loop-util.c
```diff
@@ -332,6 +333,16 @@ static int loop_device_make_internal(
                         return r;
         }
 
+        if (path) {
+                backing_file = strdup(path);
+                if (!backing_file)
+                        return -ENOMEM;
```
poettering: maybe use path_make_absolute_cwd() and path_simplify()?

src/shared/dissect-image.c
```diff
@@ -191,14 +191,10 @@ static int make_partition_devname(
 }
 #endif
 
-int dissect_image(
-                int fd,
-                const char *image_path,
+int dissect_loop_device(
```
poettering: as i mentoined in that other PR: I am not convinced this is desirable: with this change we can only dissect stuff with privileges. Before we could analyze regular files too, without loopback devices, without privileges. I'd really like to keep that, and make this something we make use of in tests...

poettering:
looks mostly good, but I am not convinced we should go `LoopDevice` only for the dissection code: being able to dissect regular files and thus operate without privs is a good thing we should keep and make use of in tests at the very least.

src/shared/dissect-image.c
yuwata: Oh, was it possible to call this without privileges? I did not know that.

yuwata: At least on the current git HEAD, `dissect_image()` does not work for file descriptor for a regular file. Also, not work for block devices without privilege.
I guess we can modify the logic to support regular file or without privilege, but l'd like to do that later.

src/shared/blockdev-util.c
yuwata: Yeah, but I'd like to do that in another later PR.

yuwata:
@poettering Thank you for the comments. Again, dissect_image() is revived in the current version. PTAL.

src/shared/blockdev-util.c
```diff
@@ -539,6 +546,12 @@ int block_device_remove_all_partitions(int fd) {
         if (r < 0)
                 return r;
 
+        if (fd < 0) {
+                fd = sd_device_open(dev, O_CLOEXEC|O_NONBLOCK|O_NOCTTY|O_RDONLY);
+                if (fd < 0)
+                        return fd;
+        }
```
poettering, line 590: you are leaking the fd here

yuwata: Ouch. Fixed.

PR #24572, opened 2022-09-05
repart: Add support for formatting verity partitions

### LINKED ISSUE #24559: repart: Add support for formatting verity partitions
### Component

systemd-repart

### Is your feature request related to a problem? Please describe

Currently, users are required to manually run `veritysetup format` after running repart to format any verity partitions. Ideall,y repart should be able to do this for the user.

### Describe the solution you'd like

Add support for a new `Verity` section to repart definitions files, providing a way to configure and format a verity partition for the given partition. If the `Verity` section is provided, repart will generate a verity partition for the current partition. The settings in the `Verity` section can be used to configure the generated Verity partition. 

### Describe alternatives you've considered

_No response_

### The systemd version you checked that didn't have the feature you are asking for

main

daandemeyer:
This commit adds a new Verity= setting to repart definition files
with two possible values: "data" and "hash".

If Verity= is set to "data", repart works as before, and populates
the partition with the content from CopyBlocks= or CopyFiles=.

If Verity= is set to "hash", repart will try to find a matching
data partition with Verity=data and equal values for CopyBlocks=
or CopyFiles=, Format= and MakeDirectories=. If a matching data
partition is found, repart will generate verity hashes for that
data partition in the verity partition. The UUID of the data
partition is set to the first 128 bits of the verity root hash. The
UUID of the hashes partition is set to the final 128 bits of the
verity root hash.

Fixes https://github.com/systemd/systemd/issues/24559

### DIFF
--- a/man/repart.d.xml
+++ b/man/repart.d.xml
@@ -581,6 +581,38 @@
       </varlistentry>
 
       <varlistentry>
+        <term><varname>Verity=</varname></term>
+
+        <listitem><para>Takes one of <literal>off</literal>, <literal>data</literal> or
+        <literal>hash</literal>. Defaults to <literal>off</literal>. If set to <literal>off</literal> or
+        <literal>data</literal>, the partition is populated with content as specified by
+        <varname>CopyBlocks=</varname> or <varname>CopyFiles=</varname>. If set to <literal>hash</literal>,
+        the partition will be populated with verity hashes from a matching verity data partition. A matching
+        data partition is a partition with <varname>Verity=</varname> set to <literal>data</literal> and the
+        same verity match key (as configured with <varname>VerityMatchKey=</varname>). If not explicitly
+        configured, the data partition's UUID will be set to the first 128 bits of the verity root hash.
+        Similarly, if not configured, the hash partition's UUID will be set to the final 128 bits of the
+        verity root hash. The verity root hash itself will be included in the output of
+        <command>systemd-repart</command>.</para>
+
+        <para>This option has no effect if the partition already exists.</para>
+
+        <para>Usage of this option in combination with <varname>Encrypt=</varname> is not supported.</para>
+
+        <para>For each unique <varname>VerityMatchKey=</varname> value, a single verity data partition
+        (<literal>Verity=data</literal>) and a single verity hash partition (<literal>Verity=hash</literal>)
+        must be defined.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>VerityMatchKey=</varname></term>
+
+        <listitem><para>Takes a short, user-chosen identifier string. This setting is used to find sibling
+        verity partitions for the current verity partition. See the description for
+        <varname>Verity=</varname>.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><varname>FactoryReset=</varname></term>
 
         <listitem><para>Takes a boolean argument. If specified the partition is marked for removal during a
@@ -746,6 +778,28 @@ SizeMaxBytes=64M
 
 <para><programlisting># ln -s 50-root.conf /usr/lib/repart.d/70-root-b.conf
 # ln -s 60-root-verity.conf /usr/lib/repart.d/80-root-verity-b.conf
+</programlisting></para>
+    </example>
+
+     <example>
+      <title>Create a data and verity partition from a OS tree</title>
+
+      <para>Assuming we have an OS tree at /var/tmp/os-tree that we want to package in a root partition
+      together with a matching verity partition, we can do so as follows:</para>
+
+      <para><programlisting># 50-root.conf
+[Partition]
+Type=root
+CopyFiles=/var/tmp/os-tree
+Verity=data
+VerityMatchKey=root
+</programlisting></para>
+
+      <para><programlisting># 60-root-verity.conf
+[Partition]
+Type=root-verity
+Verity=hash
+VerityMatchKey=root
 </programlisting></para>
     </example>
 
--- a/src/partition/repart.c
+++ b/src/partition/repart.c
@@ -132,6 +132,14 @@ typedef enum EncryptMode {
         _ENCRYPT_MODE_INVALID = -EINVAL,
 } EncryptMode;
 
+typedef enum VerityMode {
+        VERITY_OFF,
+        VERITY_DATA,
+        VERITY_HASH,
+        _VERITY_MODE_MAX,
+        _VERITY_MODE_INVALID = -EINVAL,
+} VerityMode;
+
 struct Partition {
         char *definition_path;
         char **drop_in_files;
@@ -170,11 +178,18 @@ struct Partition {
         char **copy_files;
         char **make_directories;
         EncryptMode encrypt;
+        VerityMode verity;
+        char *verity_match_key;
 
         uint64_t gpt_flags;
         int no_auto;
         int read_only;
         int growfs;
+
+        uint8_t *roothash;
+        size_t roothash_size;
+
+        Partition *siblings[_VERITY_MODE_MAX];
 
         LIST_FIELDS(Partition, partitions);
 };
@@ -211,10 +226,18 @@ static const char *encrypt_mode_table[_ENCRYPT_MODE_MAX] = {
         [ENCRYPT_KEY_FILE_TPM2] = "key-file+tpm2",
 };
 
+static const char *verity_mode_table[_VERITY_MODE_MAX] = {
+        [VERITY_OFF]  = "off",
+        [VERITY_DATA] = "data",
+        [VERITY_HASH] = "hash",
+};
+
 #if HAVE_LIBCRYPTSETUP
 DEFINE_PRIVATE_STRING_TABLE_LOOKUP_WITH_BOOLEAN(encrypt_mode, EncryptMode, ENCRYPT_KEY_FILE);
+DEFINE_PRIVATE_STRING_TABLE_LOOKUP(verity_mode, VerityMode);
 #else
 DEFINE_PRIVATE_STRING_TABLE_LOOKUP_FROM_STRING_WITH_BOOLEAN(encrypt_mode, EncryptMode, ENCRYPT_KEY_FILE);
+DEFINE_PRIVATE_STRING_TABLE_LOOKUP_FROM_STRING(verity_mode, VerityMode);
 #endif
 
 
@@ -282,6 +305,9 @@ static Partition* partition_free(Partition *p) {
         free(p->format);
         strv_free(p->copy_files);
         strv_free(p->make_directories);
+        free(p->verity_match_key);
+
+        free(p->roothash);
 
         return mfree(p);
 }
@@ -407,6 +433,21 @@ static bool context_drop_one_priority(Context *context) {
 
                 p->dropped = true;
                 log_info("Can't fit partition %s of priority %" PRIi32 ", dropping.", p->definition_path, p->priority);
+
+                /* We ensure that all verity sibling partitions have the same priority, so it's safe
+                 * to drop all siblings here as well. */
+
+                for (VerityMode mode = VERITY_OFF + 1; mode < _VERITY_MODE_MAX; mode++) {
+                        if (!p->siblings[mode])
+                                continue;
+
+                        if (p->siblings[mode]->dropped)
+                                continue;
+
+                        p->siblings[mode]->dropped = true;
+                        log_info("Also dropping sibling verity %s partition %s",
+                                 verity_mode_to_string(mode), p->siblings[mode]->definition_path);
+                }
         }
 
         return true;
@@ -1352,6 +1393,8 @@ static int config_parse_uuid(
         return 0;
 }
 
+static DEFINE_CONFIG_PARSE_ENUM_WITH_DEFAULT(config_parse_verity, verity_mode, VerityMode, VERITY_OFF, "Invalid verity mode");
+
 static int partition_read_definition(Partition *p, const char *path, const char *const *conf_file_dirs) {
 
         ConfigTableItem table[] = {
@@ -1371,6 +1414,8 @@ static int partition_read_definition(Partition *p, const char *path, const char 
                 { "Partition", "CopyFiles",       config_parse_copy_files,  0, p                    },
                 { "Partition", "MakeDirectories", config_parse_make_dirs,   0, p                    },
                 { "Partition", "Encrypt",         config_parse_encrypt,     0, &p->encrypt          },
+                { "Partition", "Verity",          config_parse_verity,      0, &p->verity           },
+                { "Partition", "VerityMatchKey",  config_parse_string,      0, &p->verity_match_key },
                 { "Partition", "Flags",           config_parse_gpt_flags,   0, &p->gpt_flags        },
                 { "Partition", "ReadOnly",        config_parse_tristate,    0, &p->read_only        },
                 { "Partition", "NoAuto",          config_parse_tristate,    0, &p->no_auto          },
@@ -1428,6 +1473,31 @@ static int partition_read_definition(Partition *p, const char *path, const char 
                         return log_oom();
         }
 
+        if (p->verity != VERITY_OFF || p->encrypt != ENCRYPT_OFF) {
+                r = dlopen_cryptsetup();
+                if (r < 0)
+                        return log_syntax(NULL, LOG_ERR, path, 1, r,
+                                          "libcryptsetup not found, Verity=/Encrypt= are not supported: %m");
+        }
+
+        if (p->verity != VERITY_OFF && !p->verity_match_key)
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "VerityMatchKey= must be set if Verity=%s", verity_mode_to_string(p->verity));
+
+        if (p->verity == VERITY_OFF && p->verity_match_key)
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "VerityMatchKey= can only be set if Verity= is not \"%s\"",
+                                  verity_mode_to_string(p->verity));
+
+        if (p->verity == VERITY_HASH && (p->copy_files || p->copy_blocks_path || p->copy_blocks_auto || p->format || p->make_directories))
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "CopyBlocks=/CopyFiles=/Format=/MakeDirectories= cannot be used with Verity=%s",
+                                  verity_mode_to_string(p->verity));
+
+        if (p->verity != VERITY_OFF && p->encrypt != ENCRYPT_OFF)
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "Encrypting verity hash/data partitions is not supported");
+
         /* Verity partitions are read only, let's imply the RO flag hence, unless explicitly configured otherwise. */
         if ((gpt_partition_type_is_root_verity(p->type_uuid) ||
              gpt_partition_type_is_usr_verity(p->type_uuid)) &&
@@ -1438,6 +1508,47 @@ static int partition_read_definition(Partition *p, const char *path, const char 
         if (gpt_partition_type_knows_growfs(p->type_uuid) &&
             p->read_only <= 0)
                 p->growfs = true;
+
+        return 0;
+}
+
+static int find_verity_sibling(Context *context, Partition *p, VerityMode mode, Partition **ret) {
+        Partition *s = NULL;
+
+        assert(p);
+        assert(p->verity != VERITY_OFF);
+        assert(p->verity_match_key);
+        assert(mode != VERITY_OFF);
+        assert(p->verity != mode);
+        assert(ret);
+
+        /* Try to find the matching sibling partition of the given type for a verity partition. For a data
+         * partition, this is the corresponding hash partiton with the same verity name (and vice versa for
+         * the hash partition).
+         */
+
+        LIST_FOREACH(partitions, q, context->partitions) {
+                if (p == q)
+                        continue;
+
+                if (q->verity != mode)
+                        continue;
+
+                assert(q->verity_match_key);
+
+                if (!streq(p->verity_match_key, q->verity_match_key))
+                        continue;
+
+                if (s)
+                        return -ENOTUNIQ;
+
+                s = q;
+        }
+
+        if (!s)
+                return -ENXIO;
+
+        *ret = s;
 
         return 0;
 }
@@ -1478,6 +1589,42 @@ static int context_read_definitions(
                 LIST_INSERT_AFTER(partitions, context->partitions, last, p);
                 last = TAKE_PTR(p);
                 context->n_partitions++;
+        }
+
+        /* Check that each configured verity hash/data partition has a matching verity data/hash partition. */
+
+        LIST_FOREACH(partitions, p, context->partitions) {
+                if (p->verity == VERITY_OFF)
+                        continue;
+
+                for (VerityMode mode = VERITY_OFF + 1; mode < _VERITY_MODE_MAX; mode++) {
+                        Partition *q;
+
+                        if (p->verity == mode)
+                                continue;
+
+                        if (p->siblings[mode])
+                                continue;
+
+                        r = find_verity_sibling(context, p, mode, &q);
+                        if (r == -ENXIO)
+                                return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                                  "Missing verity %s partition for verity %s partition with VerityMatchKey=%s",
+                                                  verity_mode_to_string(mode), verity_mode_to_string(p->verity), p->verity_match_key);
+                        if (r == -ENOTUNIQ)
+                                return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                                  "Multiple verity %s partitions found for verity %s partition with VerityMatchKey=%s",
+                                                  verity_mode_to_string(mode), verity_mode_to_string(p->verity), p->verity_match_key);
+                        if (r < 0)
+                                return r;
+
+                        if (q->priority != p->priority)
+                                return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                                  "Priority mismatch (%i != %i) for verity sibling partitions with VerityMatchKey=%s",
+                                                  p->priority, q->priority, p->verity_match_key);
+
+                        p->siblings[mode] = q;
+                }
         }
 
         return 0;
@@ -2042,26 +2189,26 @@ static int context_dump_partitions(Context *context, const char *node) {
         _cleanup_(table_unrefp) Table *t = NULL;
         uint64_t sum_padding = 0, sum_size = 0;
         int r;
-        const size_t dropin_files_col = 13;
-        bool has_dropin_files = false;
+        const size_t roothash_col = 13, dropin_files_col = 14;
+        bool has_roothash = false, has_dropin_files = false;
 
         if ((arg_json_format_flags & JSON_FORMAT_OFF) && context->n_partitions == 0) {
                 log_info("Empty partition table.");
                 return 0;
         }
 
-        t = table_new("type", "label", "uuid", "file", "node", "offset", "old size", "raw size", "size", "old padding", "raw padding", "padding", "activity", "drop-in files");
+        t = table_new("type", "label", "uuid", "file", "node", "offset", "old size", "raw size", "size", "old padding", "raw padding", "padding", "activity", "roothash", "drop-in files");
         if (!t)
                 return log_oom();
 
         if (!DEBUG_LOGGING) {
                 if (arg_json_format_flags & JSON_FORMAT_OFF)
                         (void) table_set_display(t, (size_t) 0, (size_t) 1, (size_t) 2, (size_t) 3, (size_t) 4,
-                                                    (size_t) 8, (size_t) 11, dropin_files_col);
+                                                    (size_t) 8, (size_t) 11, roothash_col, dropin_files_col);
                 else
                         (void) table_set_display(t, (size_t) 0, (size_t) 1, (size_t) 2, (size_t) 3, (size_t) 4,
-                                                    (size_t) 5, (size_t) 6, (size_t) 7, (size_t) 9, (size_t) 10, (size_t) 12,
-                                                    dropin_files_col);
+                                                    (size_t) 5, (size_t) 6, (size_t) 7, (size_t) 9, (size_t) 10,
+                                                    (size_t) 12, roothash_col, dropin_files_col);
         }
 
         (void) table_set_align_percent(t, table_get_cell(t, 0, 5), 100);
@@ -2073,7 +2220,7 @@ static int context_dump_partitions(Context *context, const char *node) {
         (void) table_set_align_percent(t, table_get_cell(t, 0, 11), 100);
 
         LIST_FOREACH(partitions, p, context->partitions) {
-                _cleanup_free_ char *size_change = NULL, *padding_change = NULL, *partname = NULL;
+                _cleanup_free_ char *size_change = NULL, *padding_change = NULL, *partname = NULL, *rh = NULL;
                 char uuid_buffer[SD_ID128_UUID_STRING_MAX];
                 const char *label, *activity = NULL;
 
@@ -2101,6 +2248,12 @@ static int context_dump_partitions(Context *context, const char *node) {
                 if (p->new_padding != UINT64_MAX)
                         sum_padding += p->new_padding;
 
+                if (p->verity == VERITY_HASH) {
+                        rh = p->roothash ? hexmem(p->roothash, p->roothash_size) : strdup("TBD");
+                        if (!rh)
+                                return log_oom();
+                }
+
                 r = table_add_many(
                                 t,
                                 TABLE_STRING, gpt_partition_type_uuid_to_string_harder(p->type_uuid, uuid_buffer),
@@ -2116,10 +2269,12 @@ static int context_dump_partitions(Context *context, const char *node) {
                                 TABLE_UINT64, p->new_padding,
                                 TABLE_STRING, padding_change, TABLE_SET_COLOR, !p->partitions_next && sum_padding > 0 ? ansi_underline() : NULL,
                                 TABLE_STRING, activity ?: "unchanged",
+                                TABLE_STRING, rh,
                                 TABLE_STRV, p->drop_in_files);
                 if (r < 0)
                         return table_log_add_error(r);
 
+                has_roothash = has_roothash || !isempty(rh);
                 has_dropin_files = has_dropin_files || !strv_isempty(p->drop_in_files);
         }
 
@@ -2144,9 +2299,16 @@ static int context_dump_partitions(Context *context, const char *node) {
                                 TABLE_EMPTY,
                                 TABLE_STRING, b,
                                 TABLE_EMPTY,
+                                TABLE_EMPTY,
                                 TABLE_EMPTY);
                 if (r < 0)
                         return table_log_add_error(r);
+        }
+
+        if (!has_roothash) {
+                r = table_hide_column_from_display(t, roothash_col);
+                if (r < 0)
+                        return log_error_errno(r, "Failed to set columns to display: %m");
         }
 
         if (!has_dropin_files) {
@@ -2353,7 +2515,15 @@ static int context_dump_partition_bar(Context *context, const char *node) {
         return 0;
 }
 
-static int context_dump(Context *context, const char *node) {
+static bool context_has_roothash(Context *context) {
+        LIST_FOREACH(partitions, p, context->partitions)
+                if (p->roothash)
+                        return true;
+
+        return false;
+}
+
+static int context_dump(Context *context, const char *node, bool late) {
         int r;
 
         assert(context);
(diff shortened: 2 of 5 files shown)


DaanDeMeyer:
cc @poettering 

This "assigns" data partitions to a verity partition so that a data partition can only be assigned to a single verity partition. I did this because if you have two identical root partitions and two identical verity partitions, the two verity partitions will both match the first root partition and so there won't be any verity partition for the data in the second root partition (which won't have the same root hash as the first root partition due to mkfs magic).

What we could also do is ensure that those two root partitions have the same root hash by checking if there's an identical partition when populating partitions and copying from that partition if one is found instead of running mkfs again. As of now this PR skips running mkfs again when generating the verity partitions, but maybe we should do it for identical regular partitions as well. That would remove the need for matching verity partitions to unique data partitions since identical data partitions will have the same root hash.

Let me know which you prefer.

man/repart.d.xml
```diff
@@ -746,6 +763,26 @@ SizeMaxBytes=64M
...
+[Partition]
+Type=root
+CopyFiles=/var/tmp/os-tree
+</programlisting></para>
+
+      <para><programlisting># root-verity.conf
+[Partition]
+Type=root-verity
+CopyFiles=/var/tmp/os-tree
+Verity=yes
```
yuwata, line 785: `Verity=hash` ?

yuwata: `Verity=hash`

src/partition/repart.c
```diff
@@ -1428,6 +1453,15 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
         }
 
+        if (p->verity != VERITY_DATA && !p->copy_files && !p->copy_blocks_path && !p->copy_blocks_auto)
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "At least one of CopyFiles=/CopyBlocks= must be used when Verity=%s is used",
+                                  verity_mode_to_string(p->verity));
+
+        if (p->verity != VERITY_DATA && p->encrypt != ENCRYPT_OFF)
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "Verity hash or data partitions cannot be encrypted");
```
yuwata: Please also check that libcryptsetup is supported.

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
         return 0;
 }
 
+static Partition* find_verity_data_partition(Context *context, Partition *p) {
+        assert(context);
+        assert(p);
+        assert(p->copy_files || p->copy_blocks_path);
+        assert(p->verity != VERITY_DATA);
```
yuwata: `assert(p->verity == VERITY_HASH);` is better.

src/partition/repart.c
```diff
@@ -1480,6 +1562,27 @@ static int context_read_definitions(
                 context->n_partitions++;
         }
 
+        /* Check that each configured verity hash partition has a matching data partition. */
+
+        LIST_FOREACH(partitions, p, context->partitions) {
+                Partition *dp;
+
+                if (p->verity == VERITY_DATA)
```
yuwata: And, here should be `if (p->verity != VERITY_HASH)`.

src/partition/repart.c
```diff
@@ -1480,6 +1562,27 @@ static int context_read_definitions(
...
+
+                dp = find_verity_data_partition(context, p);
+                if (!dp)
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Missing verity data partition for verity partition from %s",
+                                          p->definition_path);
+
+                p->data_partition = dp;
+
+                if (dp->encrypt != ENCRYPT_OFF)
```
yuwata: This should be checked in the loop in find_verity_data_partition().

src/partition/repart.c
```diff
@@ -1480,6 +1562,27 @@ static int context_read_definitions(
...
+                if (p->verity == VERITY_DATA)
+                        continue;
+
+                dp = find_verity_data_partition(context, p);
+                if (!dp)
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Missing verity data partition for verity partition from %s",
+                                          p->definition_path);
+
+                p->data_partition = dp;
```
yuwata: How about to also cache the hash partition in the data partition?
`dp->hash_partition = p`

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+                /* Only use data partitions that haven't been assigned to another verity hash partition. */
+
+                LIST_FOREACH(partitions, r, context->partitions)
+                        if (r->data_partition == q) {
+                                used = true;
+                                break;
+                        }
+
+                if (used)
+                        continue;
```
yuwata: Then, this can be simply
```c
if (q->hash_partition)
         continue;
```

src/partition/repart.c
```diff
@@ -2042,26 +2145,26 @@ static int context_dump_partitions(Context *context, const char *node) {
         _cleanup_(table_unrefp) Table *t = NULL;
         uint64_t sum_padding = 0, sum_size = 0;
         int r;
-        const size_t dropin_files_col = 13;
-        bool no_dropin_files = true;
+        const size_t roothash_col = 13, dropin_files_col = 14;
+        bool no_roothash = true, no_dropin_files = true;
```
yuwata: I'd like to rename and invert these two booleans, e.g. has_roothash and `has_dropin_files`.

src/partition/repart.c
```diff
@@ -2042,26 +2145,26 @@ static int context_dump_partitions(Context *context, const char *node) {
...
-        t = table_new("type", "label", "uuid", "file", "node", "offset", "old size", "raw size", "size", "old padding", "raw padding", "padding", "activity", "drop-in files");
+        t = table_new("type", "label", "uuid", "file", "node", "offset", "old size", "raw size", "size", "old padding", "raw padding", "padding", "activity", "roothash", "drop-in files");
         if (!t)
                 return log_oom();
 
         if (!DEBUG_LOGGING) {
                 if (arg_json_format_flags & JSON_FORMAT_OFF)
                         (void) table_set_display(t, (size_t) 0, (size_t) 1, (size_t) 2, (size_t) 3, (size_t) 4,
-                                                    (size_t) 8, (size_t) 11, dropin_files_col);
+                                                    (size_t) 8, (size_t) 11, (size_t) roothash_col, dropin_files_col);
```
yuwata: The cast is not necessary.

src/partition/repart.c
```diff
@@ -2042,26 +2145,26 @@ static int context_dump_partitions(Context *context, const char *node) {
...
                 else
                         (void) table_set_display(t, (size_t) 0, (size_t) 1, (size_t) 2, (size_t) 3, (size_t) 4,
-                                                    (size_t) 5, (size_t) 6, (size_t) 7, (size_t) 9, (size_t) 10, (size_t) 12,
-                                                    dropin_files_col);
+                                                    (size_t) 5, (size_t) 6, (size_t) 7, (size_t) 9, (size_t) 10,
+                                                    (size_t) 12, (size_t) roothash_col, dropin_files_col);
```
yuwata: Here also.

src/partition/repart.c
```diff
@@ -2820,6 +2940,9 @@ static int context_copy_blocks(Context *context) {
                 if (PARTITION_EXISTS(p)) /* Never copy over existing partitions */
                         continue;
 
+                if (p->verity != VERITY_DATA)
```
yuwata: This is correct, but looks confusing. I prefer `if (p->verity == VERITY_HASH)`, as the following logic can be operated on non verity data partition.

src/partition/repart.c
```diff
@@ -3064,6 +3187,9 @@ static int context_mkfs(Context *context) {
                 if (!p->format)
                         continue;
 
+                if (p->verity != VERITY_DATA)
```
yuwata: Also here. `if (p->verity == VERITY_HASH)`.

src/partition/repart.c
```diff
@@ -3143,6 +3269,127 @@ static int context_mkfs(Context *context) {
...
+        assert(context);
+
+        LIST_FOREACH(partitions, p, context->partitions) {
+                Partition *dp;
+                _cleanup_(loop_device_unrefp) LoopDevice *hash_device = NULL;
+                LoopDevice *data_device;
+                _cleanup_free_ uint8_t *rh = NULL;
+                size_t rhs;
+
+                if (p->dropped)
```
yuwata, line 3431: Please also check `dp->dropped`.

src/partition/repart.c
```diff
@@ -3143,6 +3269,127 @@ static int context_mkfs(Context *context) {
...
+                        continue;
+
+                if (PARTITION_EXISTS(p)) /* Never format existing partitions */
```
yuwata, line 3434: This should be also checked for `dp`.

src/partition/repart.c
```diff
@@ -3143,6 +3269,127 @@ static int context_mkfs(Context *context) {
...
+                        return log_error_errno(r,
+                                               "Failed to make loopback device of verity hash partition %" PRIu64 ": %m",
+                                               p->partno);
+
+                r = do_verity_format(data_device, hash_device, context->sector_size, &rh, &rhs);
+                if (r < 0)
+                        return r;
+
+                assert(rhs >= sizeof(sd_id128_t) * 2);
+                assert(rhs % sizeof(sd_id128_t) == 0);
```
yuwata: Why the second assertion is necessary?

src/partition/repart.c
```diff
@@ -3143,6 +3269,127 @@ static int context_mkfs(Context *context) {
...
+
+                memcpy_safe(dp->new_uuid.bytes, rh, sizeof(sd_id128_t));
+                p->new_uuid_is_set = true;
```
yuwata: I guess this should be conditionalized when `if (!p->new_uuid_is_set)`, no?

src/partition/repart.c
```diff
@@ -3143,6 +3269,127 @@ static int context_mkfs(Context *context) {
...
+
+                memcpy_safe(p->new_uuid.bytes, rh + rhs - sizeof(sd_id128_t), sizeof(sd_id128_t));
+                dp->new_uuid_is_set = true;
```
yuwata: Here also.

test/units/testsuite-58.sh
```diff
@@ -598,6 +598,44 @@ EOF
...
+[Partition]
+Type=root-${architecture}
+CopyFiles=${defs}
+EOF
+
+    cat >"$defs/verity.conf" <<EOF
+[Partition]
+Type=root-${architecture}-verity
+CopyFiles=${defs}
+Verity=yes
```
yuwata: Verity=hash

test/units/testsuite-58.sh
```diff
@@ -598,6 +598,44 @@ EOF
...
+
+    output=$(systemd-repart --definitions="$defs" \
+                            --seed="$seed" \
+                            --dry-run=no \
+                            --empty=create \
+                            --size=auto \
+                            --json=pretty \
+                            "$imgs/verity" | tee proc/self/fd/2)
+
+    roothash=$(echo "$output" | jq -r '.[] | select(.type == "root-x86-64-verity") | .roothash')
```
yuwata: `root-x86-64-verity` -> `root-${architecture}-verity`

yuwata (changes requested):

man/repart.d.xml
```diff
@@ -580,6 +580,23 @@
         <para>This option has no effect if the partition already exists.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>Verity=</varname></term>
+
+        <listitem><para>Takes one of <literal>data</literal> or <literal>hash</literal>. Defaults to
+        <literal>data</literal>. If set to <literal>data</literal>, the partition will be populated with
```
yuwata: I'd like to also add `"no"` or `"off"`and defaults to that. For safety to make not partitions mistakenly detected as a data partition for a hash partition.

man/repart.d.xml
```diff
@@ -580,6 +580,23 @@
...
+        content as specified by <option>CopyBlocks=</option> or <option>CopyFiles=</option>. If set to
+        <literal>hash</literal>, the partition will be populated with verity hashes from a matching data
+        partition. A matching data partition is defined as a partition with the same values for either
+        <option>CopyBlocks=</option> or <option>CopyFiles=</option>, <option>Format=</option> and
+        <option>MakeDirectories=</option>. The data partition's UUID will be set to the first 128 bits of the
+        verity root hash. The hashes partition's UUID will be set to the final 128 bits of the verity root
+        hash. The verity root hash itself will be included in the output of <command>systemd-repart</command>.
+        </para>
+
+        <para>This option has no effect if the partition already exists.</para></listitem>
```
yuwata: Please mention that Encrypt= cannot be enabled.

src/partition/repart.c
```diff
@@ -3512,21 +3512,6 @@ static int context_write_partition_table(
...
-                (void) context_dump_partitions(context, node);
-
-                if (arg_json_format_flags & JSON_FORMAT_OFF) {
-                        putc('\n', stdout);
-                        (void) context_dump_partition_bar(context, node);
-                        putc('\n', stdout);
-                }
-
-                fflush(stdout);
-        }
```
poettering, line 3528: this doesn't look right. the idea was to show this before we do anything, and in particular show this if we are in dry run mode, so that you can verify everything is in order before you try again. Given that the dry-run check is now done *before* we dump the table the dry-run mode is half useless...

man/repart.d.xml
```diff
@@ -746,6 +763,26 @@ SizeMaxBytes=64M
...
+</programlisting></para>
+    </example>
+
+     <example>
+      <title>Create a data and verity partition from a OS tree</title>
+
+      <para>Assuming we have an OS tree at /var/tmp/os-tree that we want to package in a root partition
+      together with a matching verity partition, we can do so as follows:</para>
+
+      <para><programlisting># root.conf
```
poettering: please prefix this with some `50-` or so, to push people towards ordering their partitions that way.

man/repart.d.xml
```diff
@@ -746,6 +763,26 @@ SizeMaxBytes=64M
...
+[Partition]
+Type=root
+CopyFiles=/var/tmp/os-tree
+</programlisting></para>
+
+      <para><programlisting># root-verity.conf
```
poettering: similar here, prefix with `60-` or so?

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+                bool used = false;
+
+                /* Check that the partition is configured to be a verity data partition. */
+                if (q->verity != VERITY_DATA)
+                        continue;
+
+                /* The verity data partition must have the same CopyFiles= or CopyBlocks= setting. */
+                if (p->copy_files && !strv_equal(p->copy_files, q->copy_files))
+                        continue;
+                if (p->copy_blocks_path && !streq(p->copy_blocks_path, q->copy_blocks_path))
```
poettering: path_equal()

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+
+        LIST_FOREACH(partitions, q, context->partitions) {
+                bool used = false;
+
+                /* Check that the partition is configured to be a verity data partition. */
+                if (q->verity != VERITY_DATA)
+                        continue;
+
+                /* The verity data partition must have the same CopyFiles= or CopyBlocks= setting. */
+                if (p->copy_files && !strv_equal(p->copy_files, q->copy_files))
```
poettering: we probably should have strv_path_equal() or so, i.e. that is like strv_equal() but uses path_equal() instead of streq() for comparing the elements.

(or do we have that somewhere? a quick grep couldn't find it for me)

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+                if (p->copy_blocks_path && !streq(p->copy_blocks_path, q->copy_blocks_path))
+                        continue;
+
+                /* If a filesystem type is set for the verity hash partition, it must match the data
+                 * partition's filesystem type. */
+                if (p->format && !streq(p->format, q->format))
+                        continue;
+
+                /* Ditto for MakeDirectories=. */
+                if (p->make_directories && !strv_equal(p->make_directories, q->make_directories))
```
poettering: also strv_path_equal() here

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+                        continue;
```
poettering, line 1540: hmm, in all checks above you condition if the field is set at all in `p`. I think we shouldn't check that. If it isn't set in `p`, then it should not be set in `q` either. Hence, use just `streq_ptr(p->format, q->format)` as check instead of `p->format && streq(p->format, q->format)`. Or am I missing something?

src/partition/repart.c
```diff
@@ -1442,6 +1476,54 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+
+                /* Only use data partitions that haven't been assigned to another verity hash partition. */
+
+                LIST_FOREACH(partitions, r, context->partitions)
```
poettering: we reserver `r` pretty exclusively as variable to take the return code of function calls. Using it for anything else, and not having it an `int` is something we should probably not do.

src/partition/repart.c
```diff
@@ -1480,6 +1562,27 @@ static int context_read_definitions(
...
+                if (!dp)
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Missing verity data partition for verity partition from %s",
+                                          p->definition_path);
+
+                p->data_partition = dp;
+
+                if (dp->encrypt != ENCRYPT_OFF)
+                        return log_syntax(NULL, LOG_ERR, dp->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "A verity data partition cannot be encrypted");
```
poettering: i would say "is not supported" rather than "cannot". I mean, it obviously *can* be encrypted, we just decide not to bother with it, since it's pretty useless...

poettering (changes requested):
Looks excellent! Some minor things though.

The dump of what we are going to do needs some thinking. We definitely must output the data in dry-run mode already, i.e. before the arg_dry_run check that causes early exit. Question is how to output the root hash then. 

Maybe something like this:

if we are in dry-run mode, or in non-JSON mode, output the table early.

if we are in non-dry-run mode and in JSON mode, output the table late, with the root hash data included.

if we are in non-dry-run mode and in non-JSON mode, output the root hashes in a simple text list at the end of things.

That way we should get the behaviour we want: a single JSOn object in JSON mode, and we only delay output when we really need to.

yuwata:
Please also see the late suggestions:
https://github.com/systemd/systemd/issues/24559#issuecomment-1237804045
https://github.com/systemd/systemd/issues/24559#issuecomment-1237806934

src/partition/repart.c
daandemeyer: Why would the dry run check be useless now? If we're in dry run, we return 0 early from `context_write_partition_table()` and execute the table logic immediately after that in `run()`.

The only regression would be that we print the table after executing all operations in non dry-run mode. That doesn't seem like a problem to me though, since there's no prompt to ask the user if they want to continue after writing the table now, so even if we write the table early, it's not like the user can do anything with it because repart will immediately go on and write the partition table.

man/repart.d.xml
```diff
@@ -580,6 +580,34 @@
...
+      <varlistentry>
+        <term><varname>Verity=</varname></term>
+
+        <listitem><para>Takes one of <literal>off</literal>, <literal>data</literal> or
+        <literal>hash</literal>. Defaults to <literal>off</literal>. If set to <literal>off</literal> or
+        <literal>data</literal>, the partition is populated with content as specified by
+        <varname>CopyBlocks=</varname> or <varname>CopyFiles=</varname>. If set to <literal>hash</literal>,
+        the partition will be populated with verity hashes from a matching verity data partition. A matching
+        data partition is a partition with <varname>Verity=</varname> set to <literal>data</literal> and the
+        same verity name (as configured with <varname>VerityName=</varname>). If not explicitl configured,
```
yuwata: explicitl -> explicitly

src/partition/repart.c
```diff
@@ -1428,6 +1495,32 @@ static int partition_read_definition(Partition *p, const char *path, const char
                         return log_oom();
         }
 
+        if (p->verity != VERITY_OFF || p->encrypt != ENCRYPT_OFF) {
+                r = dlopen_cryptsetup();
+                if (ERRNO_IS_NOT_SUPPORTED(r))
+                        return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "libcryptsetup not found, Verity=/Encrypt= are not supported");
+                if (r < 0)
+                        return r;
```
yuwata: For safety, please check `r` is negative, as `ERRNO_IS_NOT_SUPPORTED()` works for positive values.
```
if (r < 0) {
        if (ERRNO_IS_NOT_SUPPORTED(r))
                return log_syntax(...);
        return r;
}
```

Also, please propagate the original issue instead of `EINVAL`.

src/partition/repart.c
```diff
@@ -1428,6 +1495,32 @@ static int partition_read_definition(Partition *p, const char *path, const char
...
+        }
+
+        if (p->verity != VERITY_OFF && !p->verity_name)
+                return log_syntax(NULL, LOG_ERR, path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                  "VerityName= must be set if Verity=%s", verity_mode_to_string(p->verity));
+        if (p->verity == VERITY_OFF && p->verity_name)
```
yuwata: Not a matter, but I'd like to insert a empty line, like the other checks.

src/partition/repart.c
```diff
@@ -1480,6 +1573,27 @@ static int context_read_definitions(
...
+                if (!q)
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Missing sibling partition for verity partition with VerityName=%s",
+                                          p->verity_name);
+
+                s = find_verity_sibling(q->partitions_next, p);
+                if (s)
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Multiple sibling partitions found for verity partition with VerityName=%s",
+                                          p->verity_name);
```
yuwata: Why not cache the result here?? I do not think it is necessary to search every time.

src/partition/repart.c
```diff
@@ -3143,6 +3274,136 @@ static int context_mkfs(Context *context) {
...
+                if (r < 0)
+                        return log_error_errno(r,
+                                               "Failed to make loopback device of verity hash partition %" PRIu64 ": %m",
+                                               p->partno);
+
+                r = do_verity_format(data_device, hash_device, context->sector_size, &rh, &rhs);
+                if (r < 0)
+                        return r;
+
+                assert(rhs >= sizeof(sd_id128_t) * 2);
```
yuwata, line 3462: Not a matter, but dissect_image() requires the length is longer or equals to 128bit.
https://github.com/systemd/systemd/blob/90782fde463ce2af95cc6fff3dcb631e67b2da81/src/shared/dissect-image.c#L243-L244

yuwata (changes requested):
Looks much better. Please also update json format in the test.

src/partition/repart.c
daandemeyer: Yes, but we need it to be at least twice the size to accomodate the boot loader spec, since the first 128 bits need to become the partition uuid for the data partition, and the final 128 bits are needed to become the uuid for the hash partition, that's why I added the assert for twice the size.

src/partition/repart.c
daandemeyer: There's never going to be thousands of partitions, so the search will always be instant, so I don't think we need to cache the result.

src/partition/repart.c
yuwata: So, should we also update the condition in dissect_image()??

src/partition/repart.c
yuwata: Even so, still I'd like to cache it, then the code should become much simpler, I guess.

src/partition/repart.c
daandemeyer: I think so?  Unless the verity and data UUID can be the same, not sure if that's intended to work

yuwata:
Please update the json output in `test_dropin()` and `test_multiple_definitions()`.

DaanDeMeyer:
> Please update the json output in test_dropin() and test_multiple_definitions().

The roothash field should not be included if it's not set for any of the partitions, so we shouldn't have to update any other json output. CentOS CI is currently failing because we still need jq to be installed on the runners

src/partition/repart.c
daandemeyer: How would the code become simpler? If I understand correctly, it would only allow us to replace subsequent calls to find_verity_sibling() with the variable instead, that doesn't seem like a major simplification.

src/partition/repart.c
```diff
@@ -3143,6 +3276,136 @@ static int context_mkfs(Context *context) {
...
+
+static int context_verity(Context *context) {
+        int fd = -1, r;
+
+        assert(context);
+
+        LIST_FOREACH(partitions, p, context->partitions) {
+                Partition *dp;
+                _cleanup_(loop_device_unrefp) LoopDevice *hash_device = NULL;
+                LoopDevice *data_device;
```
yuwata: `data_device` is leaked.

yuwata:
> > Please update the json output in test_dropin() and test_multiple_definitions().
> 
> The roothash field should not be included if it's not set for any of the partitions, so we shouldn't have to update any other json output. CentOS CI is currently failing because we still need jq to be installed on the runners

Indeed. Sorry for the noise.

src/partition/repart.c
poettering: You are right. But also, it's a bit strange if we first say "Applying changes" and only much later say which changes... Sounds better to say first what we are going to change, before we change it.

poettering: (I mean, we had various bugs where the application of the changes actually failed, so I think it's kinda nice to report first what are going to do, so that this is visible once we fail)

src/partition/repart.c
```diff
@@ -5026,6 +5011,21 @@ static int run(int argc, char *argv[]) {
...
+                (void) context_dump_partitions(context, node);
+
+                if (arg_json_format_flags & JSON_FORMAT_OFF) {
+                        putc('\n', stdout);
+                        (void) context_dump_partition_bar(context, node);
+                        putc('\n', stdout);
+                }
+
+                fflush(stdout);
+        }
```
poettering: should probably be a function of it's own actually, given that it's quite complex and main() mostly just dispatches complex stuff to other functions, and doesn#t do complex stuff on its own

man/repart.d.xml
```diff
@@ -580,6 +580,34 @@
...
+        </para>
+
+        <para>This option has no effect if the partition already exists.</para>
+
+        <para>Usage of this option in combination with <varname>Encrypt=</varname> is not supported.</para>
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>VerityName=</varname></term>
```
poettering: "name" sound too generic, it really sounds like a label or so that we write into the superblock.

Call it `VerityMatchKey=` or so?

man/repart.d.xml
```diff
@@ -580,6 +580,34 @@
...
+
+        <listitem><para>Takes a string. This setting is used to find sibling verity partitions for the
+        current verity partition. See the description for <varname>Verity=</varname>.</para></listitem>
```
poettering: maybe say "short, user-chosen identifier string"

man/repart.d.xml
```diff
@@ -580,6 +580,34 @@
...
+
+        <listitem><para>Takes one of <literal>off</literal>, <literal>data</literal> or
+        <literal>hash</literal>. Defaults to <literal>off</literal>. If set to <literal>off</literal> or
+        <literal>data</literal>, the partition is populated with content as specified by
+        <varname>CopyBlocks=</varname> or <varname>CopyFiles=</varname>. If set to <literal>hash</literal>,
+        the partition will be populated with verity hashes from a matching verity data partition. A matching
+        data partition is a partition with <varname>Verity=</varname> set to <literal>data</literal> and the
+        same verity name (as configured with <varname>VerityName=</varname>). If not explicitly configured,
+        the data partition's UUID will be set to the first 128 bits of the verity root hash. Similarly, if
+        not configured, The hash partition's UUID will be set to the final 128 bits of the verity root
```
poettering: The → the

man/repart.d.xml
```diff
@@ -580,6 +580,34 @@
...
+        hash. The verity root hash itself will be included in the output of <command>systemd-repart</command>.
+        </para>
+
+        <para>This option has no effect if the partition already exists.</para>
+
+        <para>Usage of this option in combination with <varname>Encrypt=</varname> is not supported.</para>
```
poettering, line 600: maybe also say (and enforce) that if there's one Verity=data partition there must also be a Verity=hash partition, and vice versa. And that we fail if either one is missing.

poettering:
also: if there's not enough disk space to fit all defined partitions in, and we start dropping partitions from the set that allow that, we must ensure to drop verity partitions in pairs or not at all.

src/partition/repart.c
```diff
@@ -375,6 +399,37 @@ static int context_add_free_area(
         return 0;
 }
 
+static Partition* find_verity_sibling(Partition *head, Partition *p) {
```
poettering: hmm, once we add the pkcs7 signature stuff there will not be a single sibling anymore...

maybe add a VerityMode parameter, and just search for that type? sounds more future proof

src/partition/repart.c
```diff
@@ -375,6 +399,37 @@ static int context_add_free_area(
...
+                assert(q->verity_name);
+
+                if (!streq(p->verity_name, q->verity_name))
+                        continue;
+
+                return q;
+        }
+
+        return NULL;
+}
```
poettering: needs some protection if people assign the same match key to multiuple partitions

src/partition/repart.c
```diff
@@ -375,6 +399,37 @@ static int context_add_free_area(
...
+
+
```
poettering: spurious double newline

man/repart.d.xml
```diff
@@ -580,6 +580,34 @@
...
+        </listitem>
+      </varlistentry>
+
+      <varlistentry>
+        <term><varname>VerityName=</varname></term>
+
+        <listitem><para>Takes a string. This setting is used to find sibling verity partitions for the
+        current verity partition. See the description for <varname>Verity=</varname>.</para></listitem>
+      </varlistentry>
```
poettering, line 613: I think it would make sense to default to the configuration snippet name for VerityMatchKey= if not specified but Verity= is set

src/partition/repart.c
```diff
@@ -407,6 +462,14 @@ static bool context_drop_one_priority(Context *context) {
...
                 p->dropped = true;
                 log_info("Can't fit partition %s of priority %" PRIi32 ", dropping.", p->definition_path, p->priority);
+
+                if (p->verity != VERITY_OFF) {
+                        Partition *dp;
+
+                        assert_se(dp = find_verity_sibling(context->partitions, p));
+                        dp->dropped = true;
+                        log_info("Also dropping sibling verity partition %s", dp->definition_path);
+                }
```
poettering, line 450: hmm, this should take the priority into consideration.

or maybe better: refuse a configuration where the verity data partition has a different priority than the hash partition already when parsing.

And then add comment here saying that it's safe to drop the sibling regardless of priority since we checked priority matching already before.

poettering:
> also: if there's not enough disk space to fit all defined partitions in, and we start dropping partitions from the set that allow that, we must ensure to drop verity partitions in pairs or not at all.

Ah, i see now this is already addressed.

src/partition/repart.c
```diff
@@ -1480,6 +1575,27 @@ static int context_read_definitions(
...
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Missing sibling partition for verity partition with VerityName=%s",
+                                          p->verity_name);
+
+                s = find_verity_sibling(q->partitions_next, p);
+                if (s)
+                        return log_syntax(NULL, LOG_ERR, p->definition_path, 1, SYNTHETIC_ERRNO(EINVAL),
+                                          "Multiple sibling partitions found for verity partition with VerityName=%s",
+                                          p->verity_name);
+        }
```
poettering, line 1628: i'd really hide the uniqueness check in find_verity_sibling() and make it return ENXIO if it can't find the sibling and ENOTUNIQ if not unique

And yeah, I think too it makes sense to cache the result in the partition structure (just add two pointer fields `verity_data_partition` and `verity_hash_partition` to it, no need to avoid wasting a few bytes here), and then use that for the actual verity operatoin instead of searching again

src/partition/repart.c
```diff
@@ -3143,6 +3276,135 @@ static int context_mkfs(Context *context) {
...
+
+                if (PARTITION_EXISTS(p)) /* Never format existing partitions */
+                        continue;
+
+                if (p->verity != VERITY_HASH)
+                        continue;
+
+                assert_se(dp = find_verity_sibling(context->partitions, p));
+
+                if (dp->dropped)
```
poettering: if this is set, then this would be a bug, since then `p` should have been dropped, too. hence, change to assert

src/partition/repart.c
```diff
@@ -3143,6 +3276,135 @@ static int context_mkfs(Context *context) {
...
+                        continue;
+
+                if (PARTITION_EXISTS(dp))
```
poettering: hmm, this is an interesting restriction. does this make sense? i mean, it sounds ok to take an image with a rootfs initialized and then use repart to add in a verity partition. we can just drop this restriction without ill effect, no?

man/repart.d.xml
daandemeyer: How would that work? Each of the verity sibling partitions will have a different filename

DaanDeMeyer:
We now output the partition table again at the end with the roothashes if we're not outputting JSON.

man/repart.d.xml
```diff
@@ -746,6 +778,28 @@ SizeMaxBytes=64M
...
+
+      <para>Assuming we have an OS tree at /var/tmp/os-tree that we want to package in a root partition
+      together with a matching verity partition, we can do so as follows:</para>
+
+      <para><programlisting># 50-root.conf
+[Partition]
+Type=root
+CopyFiles=/var/tmp/os-tree
+Verity=data
+VerityName=root
```
yuwata: VerityNmae= -> VerityMatchKey=

man/repart.d.xml
```diff
@@ -746,6 +778,28 @@ SizeMaxBytes=64M
...
+</programlisting></para>
+
+      <para><programlisting># 60-root-verity.conf
+[Partition]
+Type=root-verity
+Verity=hash
+VerityName=root
```
yuwata: Here also.

man/repart.d.xml
poettering: Yeah, but that way you have to specify VerityMatchKey= only in one of the two.

i.e. the idea would be that the data partition .conf would carry no verity-related setting at all, and the hash partition .conf could still reference it via VerityMatchKey= robustly

daandemeyer: At the moment that doesn't work because you still need to set `Verity=data` for the data partition.

Given that repart now has dropin support, I figure a simple dropin can be used to turn a non verity root partition into a verity data partition while still keeping everything nicely explicit.

yuwata: Yeah, I do not like the suggestion. We should needlessly complicate the matching logic, and should not do match two partitions automatically. So I suggested VerityName=/VerityMatchKey=.

src/partition/repart.c
```diff
@@ -2362,11 +2536,23 @@ static int context_dump(Context *context, const char *node) {
...
+                return 0;
+
         r = context_dump_partitions(context, node);
         if (r < 0)
                 return r;
 
-        if (arg_json_format_flags & JSON_FORMAT_OFF) {
+        /* Make sure we only write the partition bar once, even if we're writing the partition table twice to
+         * communicate roothashes. */
+        if (arg_json_format_flags & JSON_FORMAT_OFF && !late) {
```
poettering: if you mix logical and binary &/&& in one expression, please use (), it's easier on the eye even if not needed. FLAGS_SET() also works

src/partition/repart.c
```diff
@@ -1428,6 +1473,34 @@ static int partition_read_definition(Partition *p, const char *path, const char
                         return log_oom();
         }
 
+        if (p->verity != VERITY_OFF || p->encrypt != ENCRYPT_OFF) {
+                r = dlopen_cryptsetup();
+                if (r < 0) {
+                        if (ERRNO_IS_NOT_SUPPORTED(r))
+                                return log_syntax(NULL, LOG_ERR, path, 1, r,
+                                                "libcryptsetup not found, Verity=/Encrypt= are not supported");
+                        return r;
```
yuwata: Maybe, also better to log about the failure?

src/partition/repart.c
```diff
@@ -2353,7 +2518,16 @@ static int context_dump_partition_bar(Context *context, const char *node) {
         return 0;
 }
 
-static int context_dump(Context *context, const char *node) {
+static bool context_has_roothash(Context *context) {
+        bool has_roothash = false;
+
+        LIST_FOREACH(partitions, p, context->partitions)
+                has_roothash = has_roothash || p->roothash;
```
yuwata: Why not return eariler?
```
        LIST_FOREACH(partitions, p, context->partitions)
                if (p->roothash)
                        return true;
```

src/partition/repart.c
```diff
@@ -1371,6 +1414,8 @@ static int partition_read_definition(Partition *p, const char *path, const char
                 { "Partition", "CopyFiles",       config_parse_copy_files,  0, p                    },
                 { "Partition", "MakeDirectories", config_parse_make_dirs,   0, p                    },
                 { "Partition", "Encrypt",         config_parse_encrypt,     0, &p->encrypt          },
+                { "Partition", "Verity",          config_parse_verity,      0, &p->verity           },
+                { "Partition", "VerityMatchKey",      config_parse_string,  0, &p->verity_match_key },
```
bluca: indentation looks off?

src/partition/repart.c
```diff
@@ -3172,6 +3358,130 @@ static int context_mkfs(Context *context) {
...
+        r = sym_crypt_init(&cd, hash_device->node);
+        if (r < 0)
+                return log_error_errno(r, "Failed to allocate libcryptsetup context: %m");
+
+        r = sym_crypt_format(
+                        cd, CRYPT_VERITY, NULL, NULL, NULL, NULL, 0,
+                        &(struct crypt_params_verity){
+                                .data_device = data_device->node,
+                                .flags = CRYPT_VERITY_CREATE_HASH,
+                                .hash_name = "sha256",
```
bluca, line 3389: doesn't need to be done now, but there are a lot of parameters that can be decided by the caller when creating the verity device, including the hash type - should at least be an item in the TODO to add parameters for hash type, FEC, etc

DaanDeMeyer:
All remaining comments addressed, setting the green label

yuwata (approved):
LGTM.

poettering:
CI failures appear unrelated.

PR #24579, opened 2022-09-06
systemd-nspawn: support colon-separated pair of host / container network interface names

tmartin-git:
systemd-nspawn now optionally supports colon-separated pair of  host interface name and container interface name for --network-macvlan, --network-ipvlan and --network-interface options.
Also supported in .nspawn configuration files (i.e Interface=, MACVLAN=, IPVLAN= parameters).

Actually, follows the same pattern as --network-veth-extra= command line switch (and VirtualEthernetExtra= in .nspawn configuration files).

Tested manually with the help of mkosi framework.


src/nspawn/nspawn-network.c
```diff
@@ -743,3 +757,111 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+        if (p)
+                return -EINVAL;
+
+        r = strv_push_pair(l, a, b);
+        if (r < 0)
+                return -ENOMEM;
+
+        a = b = NULL;
+        return 0;
+}
```
poettering, line 789: you have the same function three times here? why?

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +757,111 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                if (!b)
+                        return -ENOMEM;
+        }
+
+        if (p)
+                return -EINVAL;
+
+        r = strv_push_pair(l, a, b);
+        if (r < 0)
+                return -ENOMEM;
```
poettering: propagate the original error please

src/nspawn/nspawn.c
```diff
@@ -5265,6 +5242,9 @@ static int run_container(
                                 _exit(EXIT_FAILURE);
                         }
 
+                        /* revert network interfaces pair list so that interfaces get their initial name back */
+                        arg_network_interfaces=strv_reverse(arg_network_interfaces);
```
poettering: follow coding style please, space before/after `=`

poettering: I don't grok this. can you explain?

also s/revert/reverse/

src/nspawn/nspawn-network.c
poettering, line 789: our functions should either log about all errors or about none (in the latter case the caller can log). Logging about some errors but not others is wrong, because this means we might get errors about which we don#t log at all, or log about twice.

Also see https://systemd.io/CODING_STYLE/#logging

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +757,111 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                return r;
+        if (r == 0 || !ifname_valid(a))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                        "IPVLAN network interface name not valid: %s", a);
+
+        r = test_network_interface_initialized(a);
+        if (r < 0)
+                return r;
+
+        r = extract_first_word(&p, &b, ":", EXTRACT_DONT_COALESCE_SEPARATORS);
```
poettering: no need to use extract_first_word() a second time: the whole remainder of the string must be the second ifname, and `p` points to that now.

src/nspawn/nspawn-network.c
```diff
@@ -561,16 +569,19 @@ int setup_macvlan(const char *machine_name, pid_t pid, char **ifaces) {
...
 
-                n = strjoin("mv-", *i);
-                if (!n)
-                        return log_oom();
-
-                r = shorten_ifname(n);
-                if (r > 0) {
-                        a = strjoin("mv-", *i);
-                        if (!a)
+                if (streq(*b,IFNAME_NOTSET)) {
```
poettering: coding style, please add space after `,`

poettering: (here and everywhere)

src/nspawn/nspawn-network.c
```diff
@@ -561,16 +569,19 @@ int setup_macvlan(const char *machine_name, pid_t pid, char **ifaces) {
...
-                }
+
+                        r = shorten_ifname(n);
+                        if (r > 0) {
+                                a = strjoin("mv-", *i);
+                                if (!a)
+                                        return log_oom();
+                        }
+                } else
+                        n = strdup(*b);
```
poettering: missing oom check

src/nspawn/nspawn-network.c
```diff
@@ -527,19 +535,19 @@ int move_network_interfaces(int netns_fd, char **ifaces) {
         return 0;
 }
 
-int setup_macvlan(const char *machine_name, pid_t pid, char **ifaces) {
+int setup_macvlan(const char *machine_name, pid_t pid, char **pairs) {
```
poettering: maybe call this ifaces_pairs(), it's still the list of interfaces after all

src/nspawn/nspawn-network.c
```diff
@@ -642,16 +653,19 @@ int setup_ipvlan(const char *machine_name, pid_t pid, char **ifaces) {
...
-                }
+
+                        r = shorten_ifname(n);
+                        if (r > 0) {
+                                a = strjoin("iv-", *i);
+                                if (!a)
+                                        return log_oom();
+                        }
+                } else
+                        n = strdup(*b);
```
poettering: missing oom check

src/nspawn/nspawn-network.c
```diff
@@ -642,16 +653,19 @@ int setup_ipvlan(const char *machine_name, pid_t pid, char **ifaces) {
...
 
-                n = strjoin("iv-", *i);
-                if (!n)
-                        return log_oom();
-
-                r = shorten_ifname(n);
-                if (r > 0) {
-                        a = strjoin("iv-", *i);
-                        if (!a)
+                if (streq(*b,IFNAME_NOTSET)) {
```
poettering: space after the comma

src/nspawn/nspawn-settings.c
```diff
@@ -470,6 +470,84 @@ int config_parse_veth_extra(
...
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+
+        r = network_ipvlan_parse_and_test(&settings->network_ipvlan, rvalue);
+        if (r < 0)
+                log_syntax(unit, LOG_WARNING, filename, line, r, "Invalid IPVLAN network interface link specification %s: %m", rvalue);
+
+        return 0;
+}
```
poettering, line 534: also same function three times. just have one function. and include the `lvalue` parameter in the log string to clarify which settin is having the invalid setting

man/systemd-nspawn.xml
```diff
@@ -841,7 +841,8 @@
       <varlistentry>
         <term><option>--network-interface=</option></term>
 
-        <listitem><para>Assign the specified network interface to the container. This will remove the
+        <listitem><para>Assign the specified network interface to the container. Alternatively, takes a
+        colon-separated pair of host interface name and container interface name. This will remove the
```
poettering: "container interface name" is too vague, could also mean interface that points to the container. Say "interface name inside the container". Say clearly this renames the interface.

Also: "Alternatively"? It still moves the interface to the container. either way

Hmm, so when the container dies the interface is moved back to the host, but the container name will stick to it.

This at the very least requires documentation, since it#s very confusing and surprising. i.e. if people use `--network-interface=enp0s1:eth0` and the container exits and now they suddenly have `eth0` on the host that's a *very* bad idea...

It kinda wants me not support this at all, this is just going to explode

man/systemd-nspawn.xml
```diff
@@ -869,8 +870,10 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
         <term><option>--network-macvlan=</option></term>
 
         <listitem><para>Create a <literal>macvlan</literal> interface of the specified Ethernet network
-        interface and add it to the container. A <literal>macvlan</literal> interface is a virtual interface
-        that adds a second MAC address to an existing physical Ethernet link. The interface in the container
+        interface and add it to the container. Alternatively, takes a colon-separated pair of host interface
```
poettering: "Alternatively" is wrong here too.

maybe say: "Either takes a single interface name, referencing the name on the host, or a colon-separated pair of interfaces, in which case the first one references the name on the host, and the second one the name in the container."

man/systemd.nspawn.xml
```diff
@@ -531,7 +531,9 @@
       <varlistentry>
         <term><varname>Interface=</varname></term>
 
-        <listitem><para>Takes a space-separated list of interfaces to
+        <listitem><para>Takes a space-separated list of interfaces
+        (alternatively a space-separated list of colon-separated pair
+        of host interface name and container interface name) to
```
poettering: here too

poettering (changes requested):

src/nspawn/nspawn-network.c
tmartin-git: ok - I will try and factorize these. (also taking into account your comments about logging)

src/nspawn/nspawn-network.c
tmartin-git: yes sure. 
(please note that this is inspired from the veth_extra_parse)

src/nspawn/nspawn.c
tmartin-git: arg_nework_interfaces contains a list of interface name pairs :
```
{ 
"enp0s1", "eth0", // first pair
"enp1s1", "eth1" // second pair
} 
```

In this context, we want to have network interfaces names restored to the main namespace with their initial names.


So reverting the list is just a way to do that without changing move_network_interfaces prototype / body.
 
```
{
"eth1", "enp1s1", // second pair 
"eth0","enp0s1",  // first pair
} 

```

src/nspawn/nspawn-network.c
tmartin-git: ok - will fix that.
(note that this code function is mainly inspired from the veth_extra_parse function.)

tmartin-git: Actually, I am not sure about removing the second _extract_first_word_ for two reasons:
- that's the way it was implemented for --network-veth-args,
-  _extract_first_word_  processes escaping and quotes. Using it a second time may help in keeping coherence between first and second interface name in the pair.

Of course, I will follow your advice if you are not convinced by those points.

src/nspawn/nspawn-network.c
tmartin-git: ok. will fix that

poettering: ah, please extend the comment and say explicitly this is about ensuring interfaces get their old name back when being moved back

tmartin-git:
Hello, just wanted to know if this PR is still on the right way.
 
Do you expect any action from my side ?

Thanks
/ Thierry

tmartin-git:
Hello, 
Just a kind reminder about this PR :). 

Please give me an indication about how to progress on this. 
Regards

tmartin-git:
Hello, I still have no news about this MR.

Please, just let me know how to have it go a step further ! :)

yuwata:
Sorry for late review. I added v253 milestone. Hence, we will not forget this PR, and will be reviewed at least before the next release.

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,40 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                                        "%s, interface name not valid: %s", iftype, a);
+
+        r = test_network_interface_initialized(a);
+        if (r < 0)
+                return r;
+
+        r = extract_first_word(&p, &b, ":", EXTRACT_DONT_COALESCE_SEPARATORS);
+        if (r < 0)
+                return log_error_errno(r, "Failed to extract second word in %s parameter: %m", iftype);
+        if (r == 0 || !ifname_valid(b)) {
```
poettering: this doesn't look right. if the second ifame is not valid this should generate and error, and not just silently be eaten up...

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,40 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                free(b);
+                b = strdup(IFNAME_NOTSET);
```
poettering: free_and_strdup()

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,40 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                return log_error_errno(r, "Failed to extract first word in %s parameter: %m", iftype);
+        if (r == 0 || !ifname_valid(a))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                        "%s, interface name not valid: %s", iftype, a);
+
+        r = test_network_interface_initialized(a);
+        if (r < 0)
+                return r;
+
+        r = extract_first_word(&p, &b, ":", EXTRACT_DONT_COALESCE_SEPARATORS);
```
poettering: no need to call extract_first_word() for the second part, you can as well just use `p` directly, after all it's going to point right after the `:` after the first call.

src/nspawn/nspawn-settings.c
```diff
@@ -470,6 +470,32 @@ int config_parse_veth_extra(
...
+        char*** l = data;
+        int r;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+
+        r = network_iface_pair_parse_and_test(lvalue, l, rvalue);
+        if (r < 0)
+                log_syntax(unit, LOG_WARNING, filename, line, r, "Invalid %s network interface link specification %s: %m", lvalue, rvalue);
```
poettering: this will generate two error messages for the same invalid specification

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,40 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                        return log_oom();
+        }
+
+        if (p)
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                        "%s, parameter not valid (garbage at the end: %s): %s", iftype, p, a);
+
+        r = strv_push_pair(l, a, b);
+        if (r < 0)
+                return log_error_errno(r, "Failed to push pair from %s parsed parameter: %m", iftype);
```
poettering: this can only fail in case of oom, hence maybe just use `log_oom()` here, instead of making up your own message

poettering (changes requested):

src/nspawn/nspawn-network.c
tmartin-git: I agree with that. But i wanted to be

src/nspawn/nspawn-network.c
tmartin-git: Ok, still requires strdup to get non const pointer.

src/nspawn/nspawn-network.c
tmartin-git: Ok.
The second parameter is optional - empty string means it was not provided by user.

src/nspawn/nspawn-network.c
tmartin-git: Ok.

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,39 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+        if (r == 0 || !ifname_valid(a))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                        "%s, interface name not valid: %s", iftype, a);
+
+        r = test_network_interface_initialized(a);
+        if (r < 0)
+                return r;
+
+        if (!p || isempty(p)) {
+                b=strdup(IFNAME_NOTSET);
```
daandemeyer: Coding style, spaces before and after '=' please

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,39 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                if (!b)
+                        return log_oom();
+        } else {
+                b=strdup(p);
```
daandemeyer: Ditto

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,39 @@ int remove_veth_links(const char *primary, char **pairs) {
...
 }
+
+int network_iface_pair_parse_and_test(const char* iftype, char ***l, const char *p) {
+        _cleanup_free_ char *a = NULL, *b = NULL;
+        int r;
+
+        r = extract_first_word(&p, &a, ":", EXTRACT_DONT_COALESCE_SEPARATORS);
+        if (r < 0)
+                return log_error_errno(r, "Failed to extract first word in %s parameter: %m", iftype);
+        if (r == 0 || !ifname_valid(a))
```
daandemeyer: Please log separately about `r == 0` saying the interface name is missing, not invalid

man/systemd-nspawn.xml
```diff
@@ -841,9 +841,11 @@
...
         <term><option>--network-interface=</option></term>
 
-        <listitem><para>Assign the specified network interface to the container. This will remove the
-        specified interface from the calling namespace and place it in the container. When the container
-        terminates, it is moved back to the calling namespace.  Note that
+        <listitem><para>Assign the specified network interface to the container. Either takes a single
+        interface name, referencing the name on the host, or a colon-separated pair of interfaces, in which
+        case the first one references the name on the host, and the second one the name in the container.
+        When the container
+        terminates, it is moved back to the calling namespace and renamed to its initial name.  Note that
```
daandemeyer: "it" => "the interface"
"initial" => "original"

man/systemd.nspawn.xml
```diff
@@ -531,8 +531,11 @@
       <varlistentry>
         <term><varname>Interface=</varname></term>
 
-        <listitem><para>Takes a space-separated list of interfaces to
-        add to the container. This option corresponds to the
+        <listitem><para>Takes a space-separated list of interface to add to the container.
```
daandemeyer: "of interfaces"

man/systemd.nspawn.xml
```diff
@@ -542,9 +545,11 @@
         <term><varname>MACVLAN=</varname></term>
         <term><varname>IPVLAN=</varname></term>
 
-        <listitem><para>Takes a space-separated list of interfaces to
+        <listitem><para>Takes a space-separated list of interface to
```
daandemeyer: "of interfaces"

src/nspawn/nspawn-settings.c
```diff
@@ -470,6 +470,29 @@ int config_parse_veth_extra(
...
+
+        char*** l = data;
+        int r;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+
+        r = network_iface_pair_parse_and_test(lvalue, l, rvalue);
+        return r;
```
daandemeyer: Please merge this into a single line and remove the `r` variable

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,42 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+int network_iface_pair_parse_and_test(const char* iftype, char ***l, const char *p) {
+        _cleanup_free_ char *a = NULL, *b = NULL;
+        int r;
+
+        r = extract_first_word(&p, &a, ":", EXTRACT_DONT_COALESCE_SEPARATORS);
+        if (r < 0)
+                return log_error_errno(r, "Failed to extract first word in %s parameter: %m", iftype);
+        if (r == 0)
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                        "Short read while reading %s parameter: %m", iftype);
```
daandemeyer: The indentation is off here. Please reduce by one whitespace

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,42 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+        if (!ifname_valid(a))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                        "%s, interface name not valid: %s", iftype, a);
```
daandemeyer: Ditto

src/nspawn/nspawn-settings.c
```diff
@@ -470,6 +470,27 @@ int config_parse_veth_extra(
...
+                void *data,
+                void *userdata) {
+
+        char*** l = data;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+
+        return  network_iface_pair_parse_and_test(lvalue, l, rvalue);
```
daandemeyer: Two spaces here, please only use one

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +778,42 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+
+        if (!p || isempty(p)) {
+                b = strdup(IFNAME_NOTSET);
+                if (!b)
+                        return log_oom();
+        } else {
+                b = strdup(p);
+                if (!b)
+                        return log_oom();
+                if (!ifname_valid(b))
```
daandemeyer: Please add a newline before this check

src/nspawn/nspawn-network.c
```diff
@@ -492,22 +494,37 @@ int test_network_interface_initialized(const char *name) {
...
+
+                /* if the first interface name in the pair is not set,
+                 * means we are restoring the interface back to its initial
+                 * namespace, and that no renaming was previously applied.
+                 * As a consequence, get the interface name from the second
+                 * interface name in the pair and skip useless renaming. */
+                if (streq(ifname, IFNAME_NOTSET)) {
+                        ifname = *b;
+                        do_rename = false;
+                }
```
daandemeyer: This can be simplified by just checking if either of the strings is `IFNAME_NOTSET`. Please remove the `do_rename` variable and just inline the check later in this function.

DaanDeMeyer (changes requested):
Maybe get rid of `IFNAME_NOTSET` and calculate the final name of the interface in the container when parsing the arguments? Then you get rid of all the `IFNAME_NOTSET` checking when doing the rename and you can just skip the rename if the name inside and outside of the container are the same.

DaanDeMeyer:
This still seems to be missing the suggestion to not use IFNAME_NOTSET at all

tmartin-git:
@DaanDeMeyer , in this last commit, the IFNAME_NOTSET is removed. 

That implied quite a lot of code changes, notably because the arg parser now has to be aware of the interface type.
Hope it is fine now. 

About code formating: do you have any code formating checker I could use ? 

Thanks for reviewing.

src/nspawn/nspawn-network.c
```diff
@@ -492,22 +494,32 @@ int test_network_interface_initialized(const char *name) {
...
+                char *ifname = *i;
+
+                /* if the first interface name in the pair is not set,
+                 * means we are restoring the interface back to its initial
+                 * namespace, and that no renaming was previously applied.
+                 * As a consequence, get the interface name from the second
+                 * interface name in the pair and skip useless renaming. */
+                if (streq(ifname, IFNAME_NOTSET)) {
+                        ifname = *b;
+                }
```
poettering: no {} around single-line if blocks, a sper coding style

poettering: also, why not just shorten this to:

```c
const char *ifname;

ifname = streq(*i, IFNAME_NOTSET) ? *b : *i;
```

src/nspawn/nspawn-network.c
```diff
@@ -519,27 +531,33 @@ int move_network_interfaces(int netns_fd, char **ifaces) {
                 if (r < 0)
                         return log_error_errno(r, "Failed to append namespace fd to netlink message: %m");
 
+                if (!(streq(*b, IFNAME_NOTSET) || streq(ifname, IFNAME_NOTSET))) {
```
poettering: this check doesn't make sense to me. if the second part holds then the first part will too, because of the assignment you added above?

poettering:
please don't add concepts in a earlier commit that you then remove in a later commit of the same PR. Please squash relevant commits, this is otherwise messy to review. We want "perfect" PRs, where individual commits are logical steps, not necessarily historical ones.

src/nspawn/nspawn-network.c
```diff
@@ -519,27 +520,33 @@ int move_network_interfaces(int netns_fd, char **ifaces) {
                 if (r < 0)
                         return log_error_errno(r, "Failed to append namespace fd to netlink message: %m");
 
+                if (!(streq(*b, ifname))) {
```
poettering: redundant ()

src/nspawn/nspawn-network.c
```diff
@@ -492,22 +492,23 @@ int test_network_interface_initialized(const char *name) {
...
 
         r = sd_netlink_open(&rtnl);
         if (r < 0)
                 return log_error_errno(r, "Failed to connect to netlink: %m");
 
-        STRV_FOREACH(i, ifaces) {
+        STRV_FOREACH_PAIR(i, b, iface_pairs) {
                 _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
                 int ifi;
+                char *ifname = *i;
```
poettering: what is this variable good for? might as well use *i in the loop everywhere, much the same as you use *b, no need to copy the ptr into a local var

src/nspawn/nspawn-network.c
```diff
@@ -561,13 +568,13 @@ int setup_macvlan(const char *machine_name, pid_t pid, char **ifaces) {
                 if (r < 0)
                         return log_error_errno(r, "Failed to add netlink interface index: %m");
 
-                n = strjoin("mv-", *i);
+                n = strdup(*b);
```
poettering, line 580: if these aren't modified anymore (i.e. no "mv-" prefix added here) then there's no need to allocate any memory. hence remove n and a from this loop, and use i/b directly here.

src/nspawn/nspawn-network.c
```diff
@@ -642,13 +649,13 @@ int setup_ipvlan(const char *machine_name, pid_t pid, char **ifaces) {
...
 
-                n = strjoin("iv-", *i);
+                n = strdup(*b);
                 if (!n)
                         return log_oom();
 
                 r = shorten_ifname(n);
                 if (r > 0) {
-                        a = strjoin("iv-", *i);
+                        a = strdup(*b);
```
poettering: as above

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +750,53 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+
+        r = test_network_interface_initialized(a);
+        if (r < 0)
+                return r;
+
+        if (!p || isempty(p)) {
+                b = strjoin(ifprefix, a);
+        } else {
+                b = strdup(p);
+        }
```
poettering: no {} around single line if blocks, as per coding style

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +750,53 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                                       "Short read while reading %s parameter: %m", iftype);
+        if (!ifname_valid(a))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                       "%s, interface name not valid: %s", iftype, a);
+
+        r = test_network_interface_initialized(a);
+        if (r < 0)
+                return r;
+
+        if (!p || isempty(p)) {
```
poettering: isempty() also checks for p == NULL, hence drop the left part of the expression, it's redundant

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +750,53 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+        r = strv_push_pair(l, a, b);
+        if (r < 0)
+                return log_oom();
+
+        a = b = NULL;
+        return 0;
+}
+
+int interface_pair_parse_and_test(char ***l, const char *p) {
+        return network_iface_pair_parse_and_test("Network interface", l, p, "");
```
poettering: you handle a NULL ifprefix just fine above, hence pass NULL as third arg here

src/nspawn/nspawn-settings.c
```diff
@@ -470,6 +470,69 @@ int config_parse_veth_extra(
...
+                void *data,
+                void *userdata) {
+
+        char*** l = data;
+
+        assert(filename);
+        assert(lvalue);
+        assert(rvalue);
+
+        return interface_pair_parse_and_test(l, rvalue);
```
poettering: i am not a fan of this. we generally prefer if log_syntax() is used for parse failures on config files, since it tells line numbers and such.

I am also a bit unhappy that we check system state from a config file parser. Parsers should parse, and not check system state. This should be done in a later step. 

Previously we didn't do such checks here and I doubt we should start now.

I think I would prefer if test_network_interface_initialized() would not be done when parsing but be called for all configured network interfaces in a separate step when are are actually about to actually start the container, after the parsing step. maybe in a new function, called from run(), right after we call verify_arguments().

src/nspawn/nspawn-network.c
tmartin-git: :+1:

src/nspawn/nspawn-network.c
tmartin-git: :+1:

src/nspawn/nspawn-network.c
tmartin-git: We still need to keep a copy for initial interface name in case it is shortened.
(see last modification)

src/nspawn/nspawn-network.c
tmartin-git: :+1:

src/nspawn/nspawn-network.c
tmartin-git: :+1:

src/nspawn/nspawn-network.c
tmartin-git: :+1:

src/nspawn/nspawn-network.c
tmartin-git: :+1: 
`network_iface_pair_parse_and_test` changed to support third arg=NULL.

src/nspawn/nspawn-settings.c
tmartin-git: Sorry, I missed this last comment -  but still I understand the point.

tmartin-git: Ok, Implemented

src/nspawn/nspawn-network.c
```diff
@@ -492,18 +492,28 @@ int test_network_interface_initialized(const char *name) {
         return 0;
 }
 
-int move_network_interfaces(int netns_fd, char **ifaces) {
+int test_network_interfaces_initialized(char **iface_pairs) {
+        int r;
+        STRV_FOREACH_PAIR(a, b, iface_pairs) {
+                r=test_network_interface_initialized(*a);
```
poettering: space before/after `=`

src/nspawn/nspawn-network.c
```diff
@@ -608,26 +620,28 @@ int setup_macvlan(const char *machine_name, pid_t pid, char **ifaces) {
                 if (r < 0)
                         return log_error_errno(r, "Failed to add new macvlan interfaces: %m");
 
-                (void) set_alternative_ifname(rtnl, n, a);
+                if (shortened>0)
```
poettering: space before/after `>`

src/nspawn/nspawn-network.c
```diff
@@ -608,26 +620,28 @@ int setup_macvlan(const char *machine_name, pid_t pid, char **ifaces) {
...
         if (r < 0)
                 return log_error_errno(r, "Failed to connect to netlink: %m");
 
-        STRV_FOREACH(i, ifaces) {
+        STRV_FOREACH_PAIR(i, b, iface_pairs) {
                 _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
-                _cleanup_free_ char *n = NULL, *a = NULL;
+                _cleanup_free_ char *n = NULL;
+                int shortened;
                 int ifi;
```
poettering: merge these two lines please

src/nspawn/nspawn-network.c
```diff
@@ -527,21 +543,22 @@ int move_network_interfaces(int netns_fd, char **ifaces) {
...
                 return log_error_errno(r, "Failed to connect to netlink: %m");
 
-        STRV_FOREACH(i, ifaces) {
+        STRV_FOREACH_PAIR(i, b, iface_pairs) {
                 _cleanup_(sd_netlink_message_unrefp) sd_netlink_message *m = NULL;
-                _cleanup_free_ char *n = NULL, *a = NULL;
+                _cleanup_free_ char *n = NULL;
+                int shortened;
                 struct ether_addr mac;
                 int ifi;
```
poettering: merge this line:

```c
int shortened, ifi;
```

src/nspawn/nspawn-network.c
```diff
@@ -685,7 +694,8 @@ int setup_ipvlan(const char *machine_name, pid_t pid, char **ifaces) {
                 if (r < 0)
                         return log_error_errno(r, "Failed to add new ipvlan interfaces: %m");
 
-                (void) set_alternative_ifname(rtnl, n, a);
+                if (shortened>0)
```
poettering: as above

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +753,52 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                return log_error_errno(r, "Failed to extract first word in %s parameter: %m", iftype);
+        if (r == 0)
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                       "Short read while reading %s parameter: %m", iftype);
+        if (!ifname_valid(a))
+                return log_error_errno(SYNTHETIC_ERRNO(EINVAL),
+                                       "%s, interface name not valid: %s", iftype, a);
+
+        if (isempty(p)) {
+                if(ifprefix)
```
poettering: space between `if` and `(`

src/nspawn/nspawn-network.c
```diff
@@ -743,3 +753,52 @@ int remove_veth_links(const char *primary, char **pairs) {
...
+                        b = strjoin(ifprefix, a);
+                else
+                        b = strdup(a);
+        } else
+                b = strdup(p);
+
```
poettering: we generally try to keep function calls and their immediate error handling together. hence maybe remove this empty line, since it separates the function call that might fail from the error handling that checks for this

src/nspawn/nspawn.c
```diff
@@ -1862,6 +1839,19 @@ static int verify_arguments(void) {
...
+        int r;
+        r = test_network_interfaces_initialized(arg_network_macvlan);
+        if (r < 0)
+                return r;
+
+        r = test_network_interfaces_initialized(arg_network_ipvlan);
+        if (r < 0)
+                return r;
+
+        return 0;
```
poettering, line 1856: what about covering `arg_network_interfaces` here?

man/systemd-nspawn.xml
```diff
@@ -841,9 +841,11 @@
...
         <term><option>--network-interface=</option></term>
 
-        <listitem><para>Assign the specified network interface to the container. This will remove the
-        specified interface from the calling namespace and place it in the container. When the container
-        terminates, it is moved back to the calling namespace.  Note that
+        <listitem><para>Assign the specified network interface to the container. Either takes a single
+        interface name, referencing the name on the host, or a colon-separated pair of interfaces, in which
+        case the first one references the name on the host, and the second one the name in the container.
+        When the container
+        terminates, the interface is moved back to the calling namespace and renamed to its original name.  Note that
```
poettering: rebreak lines?

man/systemd-nspawn.xml
```diff
@@ -869,8 +871,12 @@ After=sys-subsystem-net-devices-ens1.device</programlisting>
         <term><option>--network-macvlan=</option></term>
 
         <listitem><para>Create a <literal>macvlan</literal> interface of the specified Ethernet network
-        interface and add it to the container. A <literal>macvlan</literal> interface is a virtual interface
-        that adds a second MAC address to an existing physical Ethernet link. The interface in the container
+        interface and add it to the container. Either takes a single interface name, referencing the name
+        on the host, or a colon-separated pair of interfaces, in which case the first one references the name
+        on the host, and the second one the name in the container. A <literal>macvlan</literal> interface is
+        a virtual interface
```
poettering: ditto

keszybz:
LGTM. I also think all comments have been addressed.

